PR #16778, opened 2024-10-29
[ntuple] Adjust default cluster size

jblomer:
Following the experience with AGC testing, double the default compressed cluster size to 100 MB and also double the maximum uncompressed cluster size to 1 GiB.


silverweed (approved):

jblomer:
> > 100 MB and also double the maximum uncompressed cluster size to 1 GiB.
> 
> Are the 2 numbers meant to be both in MiB/GiB or both in MB/GB?

They are meant to be MB and GiB, since the first number refers to compressed bytes on disk, the second number is the sum of uncompressed in-memory objects.

pcanal:
>  They are meant to be MB and GiB, since the first number refers to compressed bytes on disk, the second number is the sum of uncompressed in-memory objects.

Why are we using 2 different units?

PR #16781, opened 2024-10-29
[ntuple] Reduce memory usage of `RPageSinkBuf`

### LINKED ISSUE #16752: [ntuple] Copy sealed page in RPageSinkBuf after compression
`RPageSinkBuf::CommitPage` currently allocates a sealed page buffer with the same size as the uncompressed page:
https://github.com/root-project/root/blob/226e4c00e5dc4b7aaf881bad810a0c78c8a14185/tree/ntuple/v7/src/RPageSinkBuf.cxx#L148

Instead the code should seal / compress the page into a temporary buffer, at which point the exact compressed size is known and the contents can be copied into an appropriately sized allocation. This saves (significant) memory depending on the compression factor, while the extra `memcpy` should not be visible compared to the compression itself.

### LINKED ISSUE #16753: [ntuple] Free uncompressed page buffers in RPageSinkBuf with IMT
With a task scheduler, `RPageSinkBuf::CommitPage` copies the uncompressed page
https://github.com/root-project/root/blob/226e4c00e5dc4b7aaf881bad810a0c78c8a14185/tree/ntuple/v7/src/RPageSinkBuf.cxx#L167-L170
for asynchronous compression. However, the uncompressed page is only released in `FlushClusterImpl` which leads to significant memory overhead for high compression factors (such as CMS miniAOD). In principle, the uncompressed buffer can be released directly after compression, which is only a bit tricky because it requires the allocator to be thread-safe.

hahnjo:
1. Shrink sealed page in `RPageSinkBuf`, https://github.com/root-project/root/issues/16752
2. Free uncompressed page after compression, https://github.com/root-project/root/issues/16753
3. Delay buffer allocation in `RPageSinkBuf`

For the non-IMT case, the memory overhead is reduced to around the zipped cluster size. With parallel compression using IMT, all uncompressed pages need to be buffered first. The actual memory overhead is a function of how quickly these are consumed in the background (the slower, the later they are freed after compression) on the one hand and on the other hand how many tasks are running concurrently (each of them needs a temporary compression buffer).

Closes #16752, closes #16753

### DIFF
--- a/tree/ntuple/v7/src/RPageSinkBuf.cxx
+++ b/tree/ntuple/v7/src/RPageSinkBuf.cxx
@@ -146,10 +146,13 @@ void ROOT::Experimental::Internal::RPageSinkBuf::CommitPage(ColumnHandle_t colum
    // elements are valid until DropBufferedPages().
    auto &zipItem = fBufferedColumns.at(colId).BufferPage(columnHandle);
    std::size_t maxSealedPageBytes = page.GetNBytes() + GetWriteOptions().GetEnablePageChecksums() * kNBytesPageChecksum;
-   zipItem.fBuf = std::make_unique<unsigned char[]>(maxSealedPageBytes);
-   R__ASSERT(zipItem.fBuf);
+   // Do not allocate the buffer yet, in case of IMT we only need it once the task is started.
    auto &sealedPage = fBufferedColumns.at(colId).RegisterSealedPage();
 
+   auto allocateBuf = [&zipItem, maxSealedPageBytes]() {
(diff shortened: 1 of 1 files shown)


pcanal:
> For the non-IMT case, the memory overhead is reduced to around the zipped cluster size. 

Is it because we buffer the compressed page before writing them?

hahnjo:
> > For the non-IMT case, the memory overhead is reduced to around the zipped cluster size.
> 
> Is it because we buffer the compressed page before writing them?

Yes obviously, that's the memory overhead by design and we cannot go lower than that.

Issue #16736, opened 2024-10-23
Please improve documentation and/or argument names for TH1::GetQuantiles()


AntoniMarcinek:
### Explain what you would like to see improved and how.

Currently there is:
```cpp
TH1::GetQuantiles(int nprobSum, double* q, const double* probSum);
```
where `q` is the quantile, while `probSum` is the probability associated with the given quantile. In the description, however, we have:
```
F(x_q) = q with 0 <= q <= 1.
```
where `x_q` is the quantile (i.e. argument `q` of `GetQuantiles`), while `q` is the probability (i.e. argument `probSum`). I propose to change the argument names to `int n`, `double* xq` and `const double* q` to avoid confusion and be in agreement with both the description in ROOT and e.g. Wikipedia.

Next, in the description we have
```
[in] probSum - array of positions where quantiles will be computed.
```
but this is confusing and, I think, wrong. It should rather be "array of probabilities associated with calculated quantiles". If the argument names were changed as I propose, then this text would not even be necessary, because the thing would be obvious from the equation above. The only thing about the arguments needing description would be `n` and the behaviour in case of `q == nullptr` (note that I mean the new `q`, not the current one).

### ROOT version

master

### Installation method

irrelevant

### Operating system

irrelevant

PR #16782, opened 2024-10-29
[hist] more consistent notation of GetQuantile function

ferdymercury:
Fixes https://github.com/root-project/root/issues/16736
following suggestions by @AntoniMarcinek
original code is from @eoffermann

### DIFF
--- a/hist/hist/src/TH1.cxx
+++ b/hist/hist/src/TH1.cxx
@@ -4552,6 +4552,7 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
 ///
 /// \param[in] n maximum size of array xp and size of array p (if given)
 /// \param[out] xp array to be filled with nq quantiles evaluated at (p)
+/// If p is null (default value), then xp is actually set to the histogram bin edges
 /// \param[in] p array of probabilities where quantiles should be evaluated.
 ///   - if p is null, the CDF of the histogram will be used instead as array, and will
 ///     have a size = number of bins + 1 in h. it will correspond to the
--- a/hist/hist/src/TH1.cxx
+++ b/hist/hist/src/TH1.cxx
@@ -4547,6 +4547,7 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
 ///      x_0.5 = Q(0.5)
 /// ~~~
 ///
+/// \author Eddy Offermann
 /// code from Eddy Offermann, Renaissance
 ///
 /// \param[in] n maximum size of array p and size of array xp (if given)
--- a/hist/hist/src/TH1.cxx
+++ b/hist/hist/src/TH1.cxx
@@ -4553,7 +4553,7 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
 /// \param[in] n maximum size of array xp and size of array p (if given)
 /// \param[out] xp array to be filled with nq quantiles evaluated at (p)
 /// \param[in] p array of probabilities where quantiles should be evaluated.
-///   - if p is null, p will be computed internally and will
+///   - if p is null, the CDF of the histogram will be used instead as array, and will
 ///     have a size = number of bins + 1 in h. it will correspond to the
 ///     quantiles calculated at the lowest edge of the histogram (quantile=0) and
 ///     all the upper edges of the bins.
--- a/hist/hist/src/TH1.cxx
+++ b/hist/hist/src/TH1.cxx
@@ -4531,30 +4531,33 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
 
 ////////////////////////////////////////////////////////////////////////////////
 /// Compute Quantiles for this histogram
-/// Quantile x_q of a probability distribution Function F is defined as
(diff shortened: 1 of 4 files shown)


hist/hist/src/TH1.cxx
```diff
@@ -4531,38 +4531,42 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
...
+///      x_0.5 = Q(0.5)
 /// ~~~
 ///
+/// \author Eddy Offermann
 /// code from Eddy Offermann, Renaissance
 ///
-/// \param[in] nprobSum maximum size of array q and size of array probSum (if given)
-/// \param[in] probSum array of positions where quantiles will be computed.
-///   - if probSum is null, probSum will be computed internally and will
+/// \param[in] n maximum size of array p and size of array xp (if given)
```
AntoniMarcinek, line 4553: Shouldn't it be "p (if given)" ? `xp` is not optional

AntoniMarcinek, line 4553: Shouldn't it be "p (if given)" ? `xp` is not optional

hist/hist/src/TH1.cxx
```diff
@@ -4531,38 +4531,42 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
...
+/// \param[out] xp array to be filled with nq quantiles evaluated at (p)
+/// \param[in] p array of probabilities where quantiles should be evaluated.
+///   - if p is null, p will be computed internally and will
 ///     have a size = number of bins + 1 in h. it will correspond to the
 ///     quantiles calculated at the lowest edge of the histogram (quantile=0) and
```
AntoniMarcinek, line 4558: This is not strictly to the changes asked and provided: if `p==nullptr`, then what is `in` and what is `out` is reversed, right? `xp` becomes just a list of edges already available otherwise, and `p` gets the CDF values at edges. This part is still not clear. Maybe some next ROOT version should remove default value for `p` and provide another function to return CDF values at edges?

AntoniMarcinek (changes requested):

hist/hist/src/TH1.cxx
ferdymercury, line 4553:
```suggestion
/// \param[in] n maximum size of array xp and size of array p (if given)
```

ferdymercury: Good catch thks

hist/hist/src/TH1.cxx
```diff
@@ -4531,38 +4531,42 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
...
+/// \author Eddy Offermann
 /// code from Eddy Offermann, Renaissance
 ///
-/// \param[in] nprobSum maximum size of array q and size of array probSum (if given)
-/// \param[in] probSum array of positions where quantiles will be computed.
-///   - if probSum is null, probSum will be computed internally and will
+/// \param[in] n maximum size of array xp and size of array p (if given)
+/// \param[out] xp array to be filled with nq quantiles evaluated at (p)
+/// \param[in] p array of probabilities where quantiles should be evaluated.
+///   - if p is null, p will be computed internally and will
```
ferdymercury, line 4556:
```suggestion
///   - if p is null, the CDF of the histogram will be used instead as array, and will
```

hist/hist/src/TH1.cxx
AntoniMarcinek: Sure, I know. I mean that the function does 2 very different things depending on the context. It either calculates xp given p or it calculates p. So it is either F^-1(p) or F(x)

hist/hist/src/TH1.cxx
```diff
@@ -4531,38 +4531,42 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
...
 /// ~~~
 ///
+/// \author Eddy Offermann
 /// code from Eddy Offermann, Renaissance
 ///
-/// \param[in] nprobSum maximum size of array q and size of array probSum (if given)
-/// \param[in] probSum array of positions where quantiles will be computed.
-///   - if probSum is null, probSum will be computed internally and will
+/// \param[in] n maximum size of array xp and size of array p (if given)
+/// \param[out] xp array to be filled with nq quantiles evaluated at (p)
```
ferdymercury, line 4554:
```suggestion
/// \param[out] xp array to be filled with nq quantiles evaluated at (p)
/// If p is null (default value), then xp is actually set to the histogram bin edges
```

hist/hist/src/TH1.cxx
```diff
@@ -4531,38 +4531,43 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
...
+/// If p is null (default value), then xp is actually set to the histogram bin edges
+/// \param[in] p array of probabilities where quantiles should be evaluated.
```
ferdymercury, line 4556:
```suggestion
/// \param[in] p array of cumulative probabilities where quantiles should be evaluated.
```

hist/hist/src/TH1.cxx
AntoniMarcinek: Added new issue. Sorry for the confusion

PR #13532, opened 2023-08-24
[PyROOT][9058] Implement TPython::ExecScript for Python 3

dpiparo:
Two implementations are provided, i.e. for Python versions older than 3.8, for which PEP 587 is not in force, and newer.

# This Pull request:

## Changes or fixes:
Implements TPython::ExecFile for Python 3

## Checklist:

- [v] tested changes locally
- [] updated the docs (if necessary)

This PR fixes #9058

PR #16785, opened 2024-10-29
[CPyCppyy] Don't use deprecated Py_GetProgramName

guitargeek:
The `Py_GetProgramName` function in the Python C API was deprecated in Python 3.13 and will be removed in Python 3.15.

However, it seems behavior is unchanged by just passing an empty string instead.

This avoids a warning when building CPyCppyy.

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/API.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/API.cxx
@@ -413,7 +413,21 @@ void CPyCppyy::ExecScript(const std::string& name, const std::vector<std::string
                           << std::endl;
             }
          };
-         WideStringListAppendHelper(&fConfig.argv, Py_GetProgramName());
+
+         // The first element of the PyWideStringList can be arbitray, which
+         // you can seen by running the following:
+         //
+         // Content of test.py
+         // ```python
+         // print(sys.executable) # will print the executable, e.g. "python" or "root"
+         // print(sys.argv)       # will print ['test.py', '1', '2', '3'] in this example
+         // ```
+         //
+         // ```c++
+         // CPyCppyy::ExecScript("test.py", sizeof(argv)/sizeof(argv[0]), argv);
+         // ```
+         WideStringListAppendHelper(&fConfig.argv, std::wstring{}.c_str());
+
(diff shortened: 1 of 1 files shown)


guitargeek:
@dpiparo, since you wrote this code last year:
https://github.com/root-project/root/pull/13532

You have any idea why `Py_GetProgramName` was necessary?

bindings/pyroot/cppyy/CPyCppyy/src/API.cxx
```diff
@@ -413,7 +413,21 @@ void CPyCppyy::ExecScript(const std::string& name, const std::vector<std::string
...
+         // Content of test.py
+         // ```python
+         // print(sys.executable) # will print the executable, e.g. "python" or "root"
+         // print(sys.argv)       # will print ['test.py', '1', '2', '3'] in this example
+         // ```
+         //
+         // ```c++
+         // CPyCppyy::ExecScript("test.py", sizeof(argv)/sizeof(argv[0]), argv);
+         // ```
+         WideStringListAppendHelper(&fConfig.argv, std::wstring{}.c_str());
```
vepadulano, line 429: Although I trust this is what happens, by reading the docs at https://docs.python.org/3/c-api/init.html#c.Py_GetProgramName  it looks to me that the more appropriate string to pass would be `fConfig.program_name` member

dpiparo:
> @dpiparo, since you wrote this code last year: #13532
> 
> You have any idea why `Py_GetProgramName` was necessary?

I think I was simply trying to evolve the existing Python 2 code existing at the time, there is no sophisticated strategy behind. If the usage of the deprecated function can be avoided, great.

bindings/pyroot/cppyy/CPyCppyy/src/API.cxx
guitargeek: Oh that's a pretty good idea!

guitargeek: Actually it doesn't work, because we never set `fConfig.program_name`. So it's a `nullptr`, and then there is a segfault down the line.

guitargeek:
Superseeded by #16851.

PR #16786, opened 2024-10-30
[ntuple] make CreateAnchor a public internal free function

silverweed:
This allows programs (such as 3rd party readers) to create a RNTuple anchor while keeping it clear that "regular" users are not meant to do that.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
@@ -72,11 +72,6 @@ private:
    /// Used when the file turns out to be a TFile container
    RResult<RNTuple> GetNTupleProper(std::string_view ntupleName);
 
-   RNTuple CreateAnchor(std::uint16_t versionEpoch, std::uint16_t versionMajor, std::uint16_t versionMinor,
-                        std::uint16_t versionPatch, std::uint64_t seekHeader, std::uint64_t nbytesHeader,
-                        std::uint64_t lenHeader, std::uint64_t seekFooter, std::uint64_t nbytesFooter,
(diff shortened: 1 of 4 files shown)


jblomer (approved):

vepadulano:
We usually have `Internal` just mean that it is internal to ROOT, i.e. other ROOT classes/namespaces can use it, but we don't advertise/support the usage from outside of ROOT. Why is that different in this case? Should we consider moving this to just a public interface, with docs explaining in which situations this function should be used?

silverweed:
@vepadulano the idea here is to separate "regular" users just using the RNTuple API from "power" users that want to reimplement parts thereof. I'm admittedly not sure if that's a clear enough distinction to warrant the use of the `Internal` namespace; we certainly want the public API to guide the regular user as much as possible to the correct usage of RNTuple and I think "power" users are fine with navigating the internals if they need to solve a very specific problem.  
I don't know if @jblomer has more to say about this, but I don't have a strong opinion either way.

jblomer:
I'd prefer keeping this in Internal for the time being. We can always make it more public (e.g., move to `Detail`) but it's hard to go the other way round.

pcanal:
> the idea here is to separate "regular" users just using the RNTuple API from "power" users 

Note that this is the definition of the `Detail` namespace.

The `Internal` namespace has interface that are **explicitly** only for internal purposes and (even if they are fully documented) have **explicitly** no guarantee on anything and the signature or even the semantic  can change at any time without any warning whatsoever.  Using those interface means that one takes on the burden of actively monitoring those changes (i.e. just 'us').

vepadulano (approved):

PR #16788, opened 2024-10-30
[RF] update MultiProcess/TestStatistics docs

### LINKED ISSUE #10967: [RF] MultiProcess README.md should be updated to include LikelihoodJob
#10966 fixes LikelihoodJob so that it can now be used for parallelizing non-gradient parts of the migrad minimization. This should be reflected in the relevant README.md section here: https://github.com/root-project/root/tree/master/roofit/roofitcore/src/TestStatistics#calculators

We have yet to thoroughly test the class on real life fits. We intend to do this in the coming weeks. After this, we will rewrite the docs, taking into account any possible fixes or API changes that may have been necessary after our tests.

egpbos:
# This Pull request:

## Changes or fixes:

Updates the RooFit::MultiProcess-enabled RooFit::TestStatistics classes docs to include all the most recent updates.

## Checklist:

- [x] ~tested changes locally~

This PR fixes #10967


roofit/doc/developers/test_statistics.md
```diff
@@ -150,8 +178,69 @@ ConstantTermsOptimizer::enableConstantTermsOptimization(&pdf, &normSet, dataset,
...
+Since tasks can vary in length, this means all workers simultaneously have something to do and they all finish at the same time.
+The following options using `RooFit::MultiProcess::Config` should all be set before creating the `RooMinimizer` that will perform the fit.
+
+
+### Number of workers
+The most basic setting to tweak is the number of workers $N_w$ using `RooFit::MultiProcess::Config::setDefaultNWorkers`.
+By default, the number of workers is set equal to the number of CPUs.
+Possibly, in some situations it may be better to decrease this number by one or two.
+This is because ``RooFit::MultiProcess`` uses two more processes: the master and the queue processes, giving a total of $N_w + 2$.
+This means two cores by default will have to host two processes, which may slow down the workers on those cores a bit.
```
guitargeek, line 195: Is there really such a strict mapping between processes and cores? Don't all N cores somehow dynamically share the N + 2 processes as scheduled by the OS?

roofit/doc/developers/test_statistics.md
```diff
@@ -150,8 +178,69 @@ ConstantTermsOptimizer::enableConstantTermsOptimization(&pdf, &normSet, dataset,
...
+To reduce the number of (groups of) components, set `RooFit::MultiProcess::Config::LikelihoodJob::defaultNComponentTasks` to a different desired number of tasks.
+Set `RooFit::MultiProcess::Config::LikelihoodJob::defaultNEventTasks` to anything higher than 1 to split into blocks of events.
+Both splitting strategies can be used simultaneously, resulting in worker tasks
+that calculate (groups of) components for partial event ranges.
+
+In the end all components will automatically be summed by the master process.
+
+
+### Manual task queue order
+If the above two options still do not yield the desired scaling performance an imperfect load balancing of tasks over workers could be to blame.
```
guitargeek, line 220:
```suggestion
If the above two options still do not yield the desired scaling performance, an imperfect load balancing of tasks over workers could be to blame.
```

guitargeek (approved):
Nice! I only have minor comments. Please merge at your convenience!

roofit/doc/developers/test_statistics.md
egpbos: It depends on the OS (and perhaps the hardware, I didn't test this on ARM CPUs for instance). On Linux, the processes get pinned to the cores. This boosts performance a lot, quite critical. On macOS, manual pinning is not possible, so one indeed would have to hope the OS schedules advantageously. The performance decreases if processes switch, because lowest level cache memory has to be reloaded.

egpbos:
Thanks for the review! I was recently removed from the ROOT core GitHub team, so cannot merge anymore. Perhaps a RooFit or ROOT::Math GitHub team makes sense for future commits like this? 🙂

roofit/doc/developers/test_statistics.md
guitargeek: Ok, thanks for the detailed explanation!

PR #16789, opened 2024-10-30
[ntuple] add a test for accessing a RNTupleView out of bounds

silverweed:
I realized we were missing a test for accessing a view out of bounds, so I added one.


jblomer (approved):

tree/ntuple/v7/test/ntuple_view.cxx
```diff
@@ -396,3 +396,28 @@ TEST(RNTuple, ViewFrameworkUse)
...
+      *foo = 2;
+      writer->Fill();
+   }
+
+   auto reader = RNTupleReader::Open("myNTuple", fileGuard.GetPath());
+   try {
+      auto viewJets = reader->GetView<std::int32_t>("foo");
+      viewJets(3);
+      FAIL() << "accessing an out-of-bounds entry with a view should throw";
+   } catch (RException ex) {
```
jblomer, line 420: Maybe here
```suggestion
   } catch (const RException &ex) {
```

Issue #16784, opened 2024-10-29
Remove default value of p from TH1::GetQuantiles() as is the case with TF1::GetQuantiles


AntoniMarcinek:
### Explain what you would like to see improved and how.

This is a follow-up to  #16736, see https://github.com/root-project/root/pull/16782#discussion_r1820807153

<details>
Depending on whether `p` argument is `nullptr` or not, the method actually does 2 different things.  It either calculates `xp` given `p` or it calculates `p`. So it is either F^-1(p) or F(x).
</details>

The default parameter of p = nullptr brings the method to return just the bin edges, so it is not very useful.

### ROOT version

master, although notation matches what #16782 brings

### Installation method

irrelevant

### Operating system

irrelevant

ferdymercury:
The wording is not strictly correct:

- If p is not null, it calculates F-1(p)
- If p is null, it calculates F-1(F(bin_edges)) = bin_edges

ferdymercury:
Maybe someone is using this second variant (p=null) for something, so changing this would break backward compatibility.

What is exactly your use case / what you want to achieve?

If what you want is to have a new method to get F(x), then that method already exists, it is TH1::GetCumulative, so not sure if it's useful to add a new function name for this.

AntoniMarcinek:
Oh. Missed GetCumulative. The thing that probably bothers me is the description of GetQuantiles or rather what is the use case of GetQuantiles with p==nullptr. When you say "If p is null, it calculates F-1(F(bin_edges)) = bin_edges", the bin_edges are already known (if it really calculates them inside, that is quite surprising). The really new thing (although maybe not a product, but side-effect) is the p.

I understand that removing the default would break backward compatibility, so I am not advocating just removing the  thing in the next release. It's just that if there is a default argument, normally it represents a frequent use case, so it brings attention while learning about a given method. Here it brings confusion (to me and to my student at least).

AntoniMarcinek:
Given what you wrote and what I wrote, I no longer have a firm opinion on what should be done with it unfortunately...

ferdymercury:
> he really new thing (although maybe not a product, but side-effect) is the p.

But p stays null. F(x) is stored to a temporary variable, p is not set to F(x).

> (if it really calculates them inside, that is quite surprising)

It does. It's a technical workaround to have both paths to follow the same code.

> It's just that if there is a default argument, normally it represents a frequent use case, so it brings attention while learning about a given method. Here it brings confusion (to me and to my student at least).

[...]

PR #16793, opened 2024-10-31
[hist] remove default parameter in TH1::GetQuantiles, as is the case with TF1::GetQuantiles

ferdymercury:
Fixes https://github.com/root-project/root/issues/16784

The currently default parameter of p = nullptr is a very weird use case, which calculates F-1(F(bin_edges)), ie it just returns the bin edges. So force user to really decide if he wants to pass that nullptr as argument.

Warning: this PR might break some preexisting scripts though relying on this weird feature.

### DIFF
--- a/hist/hist/src/TH1.cxx
+++ b/hist/hist/src/TH1.cxx
@@ -4552,7 +4552,7 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
 ///
 /// \param[in] n maximum size of array xp and size of array p (if given)
 /// \param[out] xp array to be filled with nq quantiles evaluated at (p). Memory has to be preallocated by caller.
-/// If p is null (default value), then xp is actually set to the (first n) histogram bin edges
+/// If p is null, then xp is actually set to the (first n) histogram bin edges
 /// \param[in] p array of cumulative probabilities where quantiles should be evaluated.
 ///   - if p is null, the CDF of the histogram will be used instead as array, and will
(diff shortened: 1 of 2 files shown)


lmoneta (commented):
Hi, 
I am not sure if somebody is using that functionality. I agree is weird, but it is not returning the bin edge, because in case of empty bins at the begining (and end) of the axis, it returns the first (last) non-empty bin edges. It could be useful in some cases.
In case we drop, we should probably remove alltogether the functionality.

hist/hist/src/TH1.cxx
```diff
@@ -4552,7 +4552,7 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
 ///
 /// \param[in] n maximum size of array xp and size of array p (if given)
 /// \param[out] xp array to be filled with nq quantiles evaluated at (p). Memory has to be preallocated by caller.
-/// If p is null (default value), then xp is actually set to the (first n) histogram bin edges
```
lmoneta, line 4555: The comment is maybe not correct. 
xp is always set to nbin+1 value, even if n is less than nbin+1

ferdymercury: This simple tests shows the opposite: https://github.com/root-project/root/pull/16782#discussion_r1820978143

hageboeck:
I would recommend not to merge this PR, since existing interfaces are broken. Instead, I would for a second time refine the documentation, see #18036. Happy for any suggestions on how to make the docs even clearer.

PR #16795, opened 2024-10-31
[CI] Add cern dns servers to container config.

hageboeck:
Due to a bug in podman + aardvark-dns, dns resolution frequently times out. The CERN dns servers are added to sidestep the container-local network.


hageboeck:
Reverted container backend to (still supported) cni, so hopefully no config changes necessary.

Issue #16796, opened 2024-10-31
RooBinSamplingPdf does not forward expectedEventsFunc creation calls


will-cern:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

I found that putting a RooBinSamplingPdf that wraps an extended pdf inside a RooAddPdf (that is in 'no-coefficient' mode, so that it is also extended), you cannot construct the NLL for this. The reason I think is that the RooBinSamplingPdf class does not forward the createExpectedEventsFunc call to the underlying pdf.

Will submit a fix in just a moment

### Reproducer

Here's a minimal reproducer:
```
   RooWorkspace w;
   w.factory("RooAddPdf::apdf(RooBinSamplingPdf::p(x,RooExtendPdf::epdf(EXPR::pdf('x',x[0,10]),e[100])))");
   RooDataSet d("data","data",*w.var("x"));
   w.pdf("apdf")->createNLL(d);
```
Produces:
```
[#1] INFO:Minimization -- p.d.f. provides expected number of events, including extended term in likelihood.
[#1] INFO:Fitting -- RooAbsPdf::fitTo(apdf) fixing normalization set for coefficient determination to observables in data
[#0] ERROR:InputArguments -- The pdf "p" of type RooBinSamplingPdf did not overload RooAbsPdf::createExpectedEventsFunc()!

 *** Break *** segmentation violation
```

### ROOT version

6.34

### Installation method

source

### Operating system

any

PR #16797, opened 2024-10-31
forward call to createExpectedEventsFunc to subpdf

will-cern:
## Changes or fixes:
https://github.com/root-project/root/issues/16796


guitargeek (approved):
Thank you very much for the fix!

PR #16800, opened 2024-10-31
[CMake] Disable TPython without development package

hahnjo:


guitargeek (approved):

guitargeek:
Thanks a lot! Makes sense

PR #16802, opened 2024-10-31
builtin_davix: Drop `PATCH_COMMAND`

hahnjo:
The build fix for FreeBSD is included already since DAVIX 0.8.6, so the `PATCH_COMMAND` is not needed (it was merged because the PR was older, opened in February, than the upgrade of the builtin in May). In fact the command breaks builtin_davix on Linux systems because `sed -i` is not portable between GNU sed and BSD sed. (Note that the same is true for the patch itself, see my comment in https://github.com/cern-fts/davix/pull/113#issuecomment-2449788349.)


guitargeek (approved):
LGTM, thanks!

PR #16810, opened 2024-11-04
[ntuple] fix IMT reading with type cast

### LINKED ISSUE #16815: [ntuple] Corruption reading `Double32_t` in IMT mode
### Check duplicate issues.

- [X] Checked for duplicates

### Description

Found in IMT reading of MiniAODs: when reading `Double32_t`, RNTuple crashes overreading the page buffer. This happens because the page is filled with floats but the column expects to see doubles.

### Reproducer

Read data from a `Double32_t` field with IMT turned on.

### ROOT version

master

### Installation method

n/a

### Operating system

n/a

### Additional context

_No response_

jblomer:
Fixes reading in IMT mode (parallel decompression) of columns that are unpacked into non-default in-memory types. This has been a conceptional issue: the parallel decompression and the page pool were not aware, so far, of the possibility of non-default in-memory types of pages. The issues was discovered in IMT reading of `Double32_t` in CMS MiniAODs.

Both, parallel decompression and the page pool are now fixed to handle non-default and multiple target in-memory types into which pages can be unsealed. The page pool will require more work. That's for a follow-up PR.

Fixes #16815

@Dr15Jones FYI.

### DIFF
--- a/tree/ntuple/v7/src/RColumnElement.cxx
+++ b/tree/ntuple/v7/src/RColumnElement.cxx
@@ -67,7 +67,7 @@ ROOT::Experimental::Internal::RColumnElementBase::GetValidBitRange(EColumnType t
    return std::make_pair(0, 0);
 }
 
-const char *ROOT::Experimental::Internal::RColumnElementBase::GetTypeName(EColumnType type)
+const char *ROOT::Experimental::Internal::RColumnElementBase::GetColumnTypeName(EColumnType type)
 {
    switch (type) {
    case EColumnType::kIndex64: return "Index64";
--- a/tree/ntuple/v7/src/RColumnElement.cxx
+++ b/tree/ntuple/v7/src/RColumnElement.cxx
@@ -108,9 +108,9 @@ const char *ROOT::Experimental::Internal::RColumnElementBase::GetColumnTypeName(
 
 template <>
 std::unique_ptr<ROOT::Experimental::Internal::RColumnElementBase>
-ROOT::Experimental::Internal::RColumnElementBase::Generate<void>(EColumnType type)
+ROOT::Experimental::Internal::RColumnElementBase::Generate<void>(EColumnType onDiskType)
 {
-   switch (type) {
+   switch (onDiskType) {
    case EColumnType::kIndex64: return std::make_unique<RColumnElement<ClusterSize_t, EColumnType::kIndex64>>();
    case EColumnType::kIndex32: return std::make_unique<RColumnElement<ClusterSize_t, EColumnType::kIndex32>>();
    case EColumnType::kSwitch: return std::make_unique<RColumnElement<RColumnSwitch, EColumnType::kSwitch>>();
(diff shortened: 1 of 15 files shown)


Dr15Jones:
@jblomer thanks for this. In your comment

>  The page pool will require more work.

Is that just an optimization or will it be needed to read back the values properly?

jblomer:
> @jblomer thanks for this. In your comment
> 
> > The page pool will require more work.
> 
> Is that just an optimization or will it be needed to read back the values properly?

The "more work" is only on optimization, not on correctness.

tree/ntuple/v7/src/RColumnElement.cxx
```diff
@@ -180,6 +180,12 @@ ROOT::Experimental::Internal::GenerateColumnElement(EColumnCppType cppType, ECol
    return nullptr;
 }
 
+std::unique_ptr<ROOT::Experimental::Internal::RColumnElementBase>
+ROOT::Experimental::Internal::GenerateColumnElement(const RColumnElementBase::RIdentifier &elementId)
+{
+   return GenerateColumnElement(elementId.fInMemoryType, elementId.fOnDiskType);
```
hahnjo, line 200: I believe this won't compile yet in this commit because this `GenerateColumnElement` doesn't exist yet... Maybe swap order with the next change?

tree/ntuple/v7/src/RPageStorage.cxx
```diff
@@ -84,37 +84,42 @@ ROOT::Experimental::RResult<std::uint64_t> ROOT::Experimental::Internal::RPageSt
...
-         return;
+   auto itr = fColumnInfos.find(physicalColumnId);
+   R__ASSERT(itr != fColumnInfos.end());
+   for (std::size_t i = 0; i < itr->second.size(); ++i) {
+      if (itr->second[i].fElementId != elementId)
+         continue;
+
+      itr->second[i].fRefCounter--;
+      if (itr->second[i].fRefCounter == 0) {
+         itr->second.erase(itr->second.begin() + i);
```
hahnjo, line 111: This may erase the last requested in-memory representation, resulting in `itr->second.empty()`. Do we then want to `fColumnInfos.erase(itr)`?

tree/ntuple/v7/src/RPageStorage.cxx
```diff
@@ -240,52 +240,57 @@ void ROOT::Experimental::Internal::RPageSource::UnzipClusterImpl(RCluster *clust
...
+            RSealedPage sealedPage;
+            sealedPage.SetNElements(pi.fNElements);
+            sealedPage.SetHasChecksum(pi.fHasChecksum);
+            sealedPage.SetBufferSize(pi.fLocator.fBytesOnStorage + pi.fHasChecksum * kNBytesPageChecksum);
+            sealedPage.SetBuffer(onDiskPage->GetAddress());
+            R__ASSERT(onDiskPage && (onDiskPage->GetSize() == sealedPage.GetBufferSize()));
+
+            auto taskFunc = [this, columnId, clusterId, firstInPage, sealedPage, elementId, &foundChecksumFailure,
+                             indexOffset = clusterDescriptor.GetColumnRange(columnId).fFirstElementIndex]() {
+               auto element = GenerateColumnElement(elementId);
```
hahnjo, line 272: This generates a column element per page, while we can share it across a column info. I think this (relatively simple) optimization should be retained from the current implementation (cf `allElements`)

tree/ntuple/v7/src/RColumnElement.cxx
jblomer: Thanks! I added another commit for this overload.

hahnjo (approved):
Changes look good to me (only a minor nit); we'll probably have to wait until the CI is green again...

PR #16814, opened 2024-11-04
Enable CppInterOp

aaronj0:
[CppInterOp](https://github.com/compiler-research/CppInterOp) exposes API from Clang and LLVM in a mostly backward compatibe way. The API support downstream tools that utilize interactive C++ by using the compiler as a service.

This PR is the first step in using pure clang based reflection API in meta, and part of eventually integrating the JITCall and DynamicLibraryManager infrastructure.

Adopting more of CppInterOp in ROOT will help simplify the LLVM migration process and allow us to upstream more patches to either CppInterOp or LLVM.

TODO:

- [x] Add github workflow to diffcheck with upstream
- [x] Improve CMake config for windows (symbol exports added upstream)
- [x] Propagate LLVM_BUILD_TYPE instead of ROOT's config to InterOp's build type
- [x] Test with `builtin_llvm=off` and `builtin_clang=off` 
- [x] reduce patches to 0


core/metacling/src/CMakeLists.txt
```diff
@@ -74,7 +75,8 @@ if(MSVC)
    set_source_files_properties(TCling.cxx COMPILE_FLAGS /bigobj)
 endif()
 
-add_dependencies(MetaCling CLING)
+add_dependencies(MetaCling CLING clangCppInterOp)
+# target_include_directories(MetaCling SYSTEM PUBLIC ${source_dir}/include/)
```
hahnjo: Can you move these modifications to the next commit, where CppInterOp is used?

interpreter/CMakeLists.txt
```diff
@@ -599,3 +599,29 @@ mark_as_advanced(FORCE BUG_REPORT_URL BUILD_CLANG_FORMAT_VS_PLUGIN BUILD_SHARED_
...
+
+
+##################### LIBINTEROP ###########################
+# Set the paths to Cling, Clang, and LLVM directories
+
+# if (cppinterop)
+
+  set(LLVM_DIR "${CMAKE_BINARY_DIR}/interpreter/llvm-project/llvm/lib/cmake/llvm")
+  set(Clang_DIR "${CMAKE_BINARY_DIR}/interpreter/llvm-project/llvm/tools/clang/lib/cmake/clang")
+  set(Cling_DIR "${CMAKE_BINARY_DIR}/interpreter/cling")
```
hahnjo, line 611: This will need adapting for externally built LLVM, Clang, and / or Cling. For LLVM and Clang, I believe this should already be taken care of above, no?

interpreter/CppInterOp/README.md
```diff
@@ -0,0 +1,372 @@
+<div align="center">
+
+# CppInterOp
```
hahnjo, line 3: As discussed offline, we probably want a similar workflow as `.github/workflows/llvm-diff.yml` to make sure our copies are not diverging

core/metacling/src/TClingCallFunc.cxx
```diff
@@ -67,6 +67,8 @@ C++ interpreter and the Clang C++ compiler, not CINT.
 
 #include "clang/Sema/SemaInternal.h"
 
+#include "clang/Interpreter/CppInterOp.h"
+
```
hahnjo, line 71: not needed yet

hahnjo (changes requested):
The direction looks good to me! In addition to the test failures that need fixing, some comments inline for consideration

core/metacling/src/CMakeLists.txt
```diff
@@ -113,10 +114,17 @@ ROOT_LINKER_LIBRARY(Cling
         $<TARGET_OBJECTS:ClingUtils>
         $<TARGET_OBJECTS:Dictgen>
         $<TARGET_OBJECTS:MetaCling>
+        $<TARGET_OBJECTS:clangCppInterOp>
```
vgvassilev, line 117: This needs to go in the `LIBRARIES` section.

core/metacling/src/TClingCallFunc.cxx
aaronj0: This change has been dropped!

interpreter/CMakeLists.txt
hahnjo: This comment still applies, even if the code was moved: variables related to this are already set above (and will change with https://github.com/root-project/root/pull/17354)

interpreter/CppInterOp/README.md
hahnjo: I would still strongly recommend implement this.

interpreter/CMakeLists.txt
```diff
@@ -591,3 +591,20 @@ mark_as_advanced(FORCE BUG_REPORT_URL BUILD_CLANG_FORMAT_VS_PLUGIN BUILD_SHARED_
...
+set(LLVM_DIR "${CMAKE_BINARY_DIR}/interpreter/llvm-project/llvm/lib/cmake/llvm")
+set(Clang_DIR "${CMAKE_BINARY_DIR}/interpreter/llvm-project/llvm/tools/clang/lib/cmake/clang")
+set(Cling_DIR "${CMAKE_BINARY_DIR}/interpreter/cling")
+set(CPPINTEROP_INCLUDE_DIRS
+${CMAKE_SOURCE_DIR}/interpreter/CppInterOp/include
+CACHE STRING "CppInterOp include directories.")
+
+
+set(USE_CLING ON CACHE BOOL "Use Cling as backend" FORCE)
+set(USE_REPL OFF CACHE BOOL "Use Clang-repl as backend" FORCE)
```
hahnjo, line 608: Uhm, can CppInterOp be changed to scope its configuration options? Setting this as a global `CACHE` variable is not great.

interpreter/CppInterOp/unittests/CppInterOp/FunctionReflectionTest.cpp
```diff
@@ -868,6 +868,10 @@ TEST(FunctionReflectionTest, JitCallAdvanced) {
 TEST(FunctionReflectionTest, GetFunctionCallWrapper) {
   if (llvm::sys::RunningOnValgrind())
     GTEST_SKIP() << "XFAIL due to Valgrind report";
+
+#ifdef _WIN32
+  GTEST_SKIP() << "Disabled on Windows (Fails with ROOT win-x64)";
+#endif
```
hahnjo, line 874: Is this done upstream in CppInterOp? We should not start by immediately diverging

interpreter/CppInterOp/unittests/CppInterOp/InterpreterTest.cpp
```diff
@@ -123,11 +123,8 @@ TEST(InterpreterTest, CreateInterpreter) {
 #endif
 }
 
-#ifdef LLVM_BINARY_DIR
-TEST(InterpreterTest, DetectResourceDir) {
-#else
+
 TEST(InterpreterTest, DISABLED_DetectResourceDir) {
-#endif // LLVM_BINARY_DIR
```
hahnjo: Is this done upstream in CppInterOp? We should not start by immediately diverging

aaronj0: This test is only required when shipping CppInterOp externally and doesn't work with ROOT since llvm is not externally built, which is why we opted to disable and track this as a patch to the tests..

interpreter/CppInterOp/README.md
aaronj0: Thanks! The workflow has been added

interpreter/CMakeLists.txt
aaronj0: Done! We can get away by just setting `USE_CLING ON`

.github/workflows/cppinterop-diff.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Check out ROOT
+        uses: actions/checkout@v4
+        with:
+          path: root
+      - name: Check out CppInterOp
+        uses: actions/checkout@v4
+        with:
+          repository: compiler-research/CppInterOp
+          path: CppInterOp
```
hahnjo, line 19: Hm, this probably checks out the default branch which is a moving target and we will not always want to update to the latest version, for example for release branches. I think we need a similar tag system as we have for LLVM

.github/workflows/cppinterop-diff.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Drop directories that are not added to ROOT
+        working-directory: CppInterOp
+        run: |
+          rm -rf .github
+      - name: Compare
+        continue-on-error: true
```
hahnjo, line 25: Ah, I was already wondering: With this line you just ignore the error and continue even if the two copies have a difference. That renders the check useless...

interpreter/CppInterOp/unittests/CppInterOp/InterpreterTest.cpp
vgvassilev: Here I think the binary dir is set but it is not what we expect it to be. I think that's part of the problem that LLVM's build system was not designed to be embedded in another build system...

interpreter/CppInterOp/unittests/CppInterOp/FunctionReflectionTest.cpp
vgvassilev: @aaronj0, do you remember how this used to fail?

vgvassilev (commented):
This practically looks good to me...

bellenot:
@vgvassilev about [your question](https://github.com/root-project/root/pull/16814#discussion_r1935455476), here is the error:
```
1: [ RUN      ] FunctionReflectionTest.GetFunctionCallWrapper
1: C:\root-dev\github\enable-interop\interpreter\CppInterOp\unittests\CppInterOp\FunctionReflectionTest.cpp(919): error: Expected equality of these values:
1:   output
1:     Which is: ""
1:   s
1:     Which is: "Hello World!\n"
1: C:\root-dev\github\enable-interop\interpreter\CppInterOp\unittests\CppInterOp\FunctionReflectionTest.cpp(944): error: Expected equality of these values:
1:   output
1:     Which is: ""
1:   "Default Ctor Called\n"
1: C:\root-dev\github\enable-interop\interpreter\CppInterOp\unittests\CppInterOp\FunctionReflectionTest.cpp(953): error: Expected equality of these values:
1:   output
1:     Which is: ""
1:   "Dtor Called\n"
1: [  FAILED  ] FunctionReflectionTest.GetFunctionCallWrapper (428 ms)
```

vgvassilev:
> @vgvassilev about [your question](https://github.com/root-project/root/pull/16814#discussion_r1935455476), here is the error:
> 
> ```
> 1: [ RUN      ] FunctionReflectionTest.GetFunctionCallWrapper
> 1: C:\root-dev\github\enable-interop\interpreter\CppInterOp\unittests\CppInterOp\FunctionReflectionTest.cpp(919): error: Expected equality of these values:
> 1:   output
> 1:     Which is: ""
> 1:   s
> 1:     Which is: "Hello World!\n"
> 1: C:\root-dev\github\enable-interop\interpreter\CppInterOp\unittests\CppInterOp\FunctionReflectionTest.cpp(944): error: Expected equality of these values:
> 1:   output
> 1:     Which is: ""
> 1:   "Default Ctor Called\n"
> 1: C:\root-dev\github\enable-interop\interpreter\CppInterOp\unittests\CppInterOp\FunctionReflectionTest.cpp(953): error: Expected equality of these values:
> 1:   output
> 1:     Which is: ""
> 1:   "Dtor Called\n"
> 1: [  FAILED  ] FunctionReflectionTest.GetFunctionCallWrapper (428 ms)
> ```

Any chance to debug this? Maybe it is the way we capture streams? I doubt this is as broken to not run the constructors on windows…

bellenot:
> Any chance to debug this? Maybe it is the way we capture streams? I doubt this is as broken to not run the constructors on windows…

Sure! I can take a look when I find some time for it

vgvassilev:
> > Any chance to debug this? Maybe it is the way we capture streams? I doubt this is as broken to not run the constructors on windows…
> 
> Sure! I can try to take a look when I find some time for it

In this case let's leave this pending and move forward with the PR.

.github/workflows/cppinterop-diff.yml
aaronj0: Yes, this is pending a release upstream, that includes patches required for ROOT. This will be updated once we have the release version

.github/workflows/cppinterop-diff.yml
aaronj0: Thanks, I will drop the continue-on-error setting. I will note that we may have to keep a patch disabling a unittest, assuming the other failing test on Windows is fixed and no longer disabled

interpreter/CMakeLists.txt
hahnjo: Note that the configuration options still have generic names, ie not prefixed with `CPPINTEROP_`...

interpreter/CMakeLists.txt
hahnjo: I gave this a short test after https://github.com/root-project/root/pull/17737 and both `builtin_llvm=OFF` and `builtin_llvm=OFF builtin_clang=OFF` seem to work correctly out-of-the-box :tada:

interpreter/CMakeLists.txt
```diff
@@ -594,3 +594,15 @@ mark_as_advanced(FORCE BUG_REPORT_URL BUILD_CLANG_FORMAT_VS_PLUGIN BUILD_SHARED_
                        C_INCLUDE_DIRS DEFAULT_SYSROOT FFI_INCLUDE_DIR FFI_LIBRARY_DIR
                        GCC_INSTALL_PREFIX LIBCLANG_BUILD_STATIC TOOL_INFO_PLIST)
 mark_as_advanced(CLEAR LLVM_ENABLE_ASSERTIONS LLVM_BUILD_TYPE)
+
+##################### LIBINTEROP ###########################
+
+#---Set InterOp include directories, for any ROOT component that requires the headers--------------------------------------------------------
+set(CPPINTEROP_INCLUDE_DIRS
+${CMAKE_SOURCE_DIR}/interpreter/CppInterOp/include
+CACHE STRING "CppInterOp include directories.")
```
hahnjo, line 603:
```suggestion
set(CPPINTEROP_INCLUDE_DIRS
  ${CMAKE_CURRENT_SOURCE_DIR}/CppInterOp/include
  CACHE STRING "CppInterOp include directories.")
```

.github/workflows/cppinterop-diff.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Check out ROOT
+        uses: actions/checkout@v4
+        with:
+          path: root
+      - name: Check out CppInterOp
+        uses: actions/checkout@v4
+        with:
+          repository: compiler-research/CppInterOp
+          ref: be5141cc071841b232e651f173a9759885f31ffb
```
hahnjo, line 19: As discussed offline, we may want to switch this to tags and potentially a tag file in `interpreter/CppInterOp` (similar to `interpreter/llvm-project/llvm-project.tag`) at some point

hahnjo (approved):

vgvassilev:
@smuzaffar, can you check if cmssw works after this pr?

hahnjo:
> can you check if cmssw works after this pr?

To be honest, I don't really see why this is relevant to check: In this PR, CppInterOp is just added and not actually used...

interpreter/CMakeLists.txt
aaronj0: Done!

vgvassilev:
> > can you check if cmssw works after this pr?
> 
> To be honest, I don't really see why this is relevant to check: In this PR, CppInterOp is just added and not actually used...

I understand, and nevertheless I'd like to get a green light on this.

smuzaffar:
cms tests are running here https://github.com/cms-sw/root/pull/218

aaronj0:
> cmssw tests passed

Thanks!

vgvassilev (approved):
Lgtm. Once we merge I will create a new release of CppInterOp.

PR #16748, opened 2024-10-26
[ci] Add Fedora 41 builds

dpiparo:

PR #16816, opened 2024-11-04
[CPyCppyy] Remove implicit conversion of any ctypes ptr type to `void*`

guitargeek:
The `IsCTypesArrayOrPointer` gives false positives in Python 3.13, resulting the void pointer converter to take the wrong code path and crash. See:
https://github.com/wlav/cppyy/issues/272

This code path is used for implicit conversion from other `ctypes` pointer types to `void*`, which is not strictly required. One can always do an explicit cast:  `ctypes.cast(my_ptr, ctypes.c_void_p )`.

Given that this a niche feature that broke Python 3.13 support for functions taking `void*`, which is quite common, it can be argued that it's better to remove this implicit conversion.

This commit fixes the following tests under Python 3.13:

```
roottest-python-basic-datatype
roottest-python-cpp-cpp
```

This reverts the following commit from upstream:
https://github.com/wlav/CPyCppyy/commit/80a0205f590394b88a583a296704356a9740606f

See also the Fedora 41 CI PR:
https://github.com/root-project/root/pull/16748

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
@@ -203,24 +203,6 @@ static bool IsPyCArgObject(PyObject* pyobject)
     return Py_TYPE(pyobject) == pycarg_type;
 }
 
-static bool IsCTypesArrayOrPointer(PyObject* pyobject)
-{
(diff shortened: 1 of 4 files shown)


vepadulano (approved):
LGTM and thanks for taking care of the synchronisation!

PR #16825, opened 2024-11-05
[ntuple] forbid creating a RNTupleWriter with an XML TFile

silverweed:
RMiniFile only knows how to write into proper ROOT files. 
I went with the blacklist approach because there are a bunch of valid cases that are not just TFiles that we want to support (e.g. TMemFile, TBufferMergerFile, ...).

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/src/RNTupleWriter.cxx
+++ b/tree/ntuple/v7/src/RNTupleWriter.cxx
@@ -25,6 +25,7 @@
 #include <ROOT/RPageStorage.hxx>
 #include <ROOT/RPageStorageFile.hxx>
 
+#include <TFile.h>
 #include <TROOT.h>
 
 #include <utility>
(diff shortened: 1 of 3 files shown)


tree/ntuple/v7/src/RNTupleWriter.cxx
```diff
@@ -95,6 +96,10 @@ std::unique_ptr<ROOT::Experimental::RNTupleWriter>
 ROOT::Experimental::RNTupleWriter::Append(std::unique_ptr<RNTupleModel> model, std::string_view ntupleName, TFile &file,
                                           const RNTupleWriteOptions &options)
 {
+   const auto *fileClassName = file.IsA()->GetName();
+   if (strcmp(fileClassName, "TXMLFile") == 0)
+      throw RException(R__FAIL("RNTupleWriter doesn't support writing to an XML file"));
```
pcanal, line 101:
```suggestion
   if (!file.IsBinary() == 0)
      throw RException(R__FAIL("RNTupleWriter only support writing to a binary file. Can not write into "+std::string(file.GetName()));
```
or if keeping the same logic
```suggestion
   static TClassRef XmlClass("TXMLFile");
   const auto *fileClassName = file.IsA()->GetName();
   if (XmlClass == file->IsA()) == 0)
      throw RException(R__FAIL("RNTupleWriter doesn't support writing to an XML file. Can not write into "+std::string(file.GetName()));
```

pcanal: Also a common text file is the JSON format.

silverweed: @pcanal but the TFile doesn't have special logic if you create it with a ".json" extension, does it?

pcanal: Right there is no  `TJsonFile`.

pcanal (approved):

PR #16813, opened 2024-11-04
[skip-ci] [ntuple] Add naming rules to the specification

vepadulano:

PR #16836, opened 2024-11-06
[ntuple] Enforce valid naming upon field creation

### LINKED ISSUE #16834: `RFieldBase::Create` does not enforce valid field names
### Check duplicate issues.

- [X] Checked for duplicates

### Description

`RFieldBase::Create` doesn't check whether the provided name of the created field is valid, this only happens in `RNTupleModel::[Add|Make]Field. This means that currently, it is possible to create field names containing dots in this way.

### Reproducer

```cpp
auto field = RFieldBase::Create("x.y", "float").Unwrap();
std::cout << field->GetFieldName() << std::endl;
```

### ROOT version

any

### Installation method

any

### Operating system

any

### Additional context

_No response_

enirolf:
Before, field name checking only took place in `RNTupleModel`. This PR performs the check upon construction of the `RFieldBase`. To still allow for special cases where fields can have an empty name, e.g. the zero field, this specific check has been moved out of `RFieldBase::EnsureValidFieldName` and to the places where this should actually be enforced.

Fixes #16834

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
@@ -214,8 +214,9 @@ private:
    /// Changed by Freeze() / Unfreeze() and by the RUpdater.
    bool fIsFrozen = false;
 
-   /// Checks that user-provided field names are valid in the context
-   /// of this NTuple model. Throws an RException for invalid names.
+   /// Checks that user-provided field names are valid in the context of this RNTuple model.
+   /// Throws an RException for invalid names, empty names (which is reserved for the zero field) and duplicate field
+   /// names.
    void EnsureValidFieldName(std::string_view fieldName);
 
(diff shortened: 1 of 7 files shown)


vepadulano (approved):
Thank you!

jblomer (approved):
Thanks! Looks in principle good to me. We can still clone though to the empty name, I think?

enirolf:
> We can still clone though to the empty name, I think? Probably that's ok. 

Yes, that's correct. In fact, that is necessary for (among others) `RNTupleDS` to work properly.

> The precise rule is that field names must not be empty _on disk_ but can be empty in memory.

Exactly, this responsibility is now moved to `RFieldDescriptorBuilder`.

vepadulano:
> The precise rule is that field names must not be empty on disk but can be empty in memory. 

I will add this to https://github.com/root-project/root/pull/16813

PR #16671, opened 2024-10-14
[Do not merge][PyROOT] Don't call `ClearProxiedObjects()` in hard shutdown mode

guitargeek:
The old PyROOT (pre 2019) had a slightly different `atexit` handler than the new one. The hard shutdown mode is different: `ClearProxiedObjects()` is called before the final `gROOT->EndOfProcessCleanups()`.

I had the impresstion that this is redundant, and indeed the old PyROOT didn't do it. This commit suggests to only call `ClearProxiedObjects()` in soft shutdown mode, in an attempt to avoid the frequent crashes seen at the end of PyROOT tutorials, especially on Ubuntu 24.10.

See also the original PRs where this functionality was introduced:

  * https://github.com/root-project/root/pull/4687
  * https://github.com/root-project/root/pull/4753

PR #16697, opened 2024-10-16
[hist] add Scale function to context menu

ferdymercury:
So that you can right click on the histogram and scale it.

sibling PR is in https://github.com/root-project/roottest/pull/1203

PR #16837, opened 2024-11-06
Simplify memory regulation logic

vepadulano:
With *all* these changes, I do not see anymore failures in a local run with the ubuntu 24.10 CI image. let's see if the rest of the CI agrees. Note that one of the changes is equivalent to https://github.com/root-project/root/pull/16671

### DIFF
--- a/.github/workflows/root-ci-config/buildconfig/ubuntu2410.txt
+++ b/.github/workflows/root-ci-config/buildconfig/ubuntu2410.txt
@@ -1,0 +1,2 @@
+pythia8=OFF
+tmva-cpu=OFF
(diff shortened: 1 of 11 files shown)


guitargeek (approved):
This is really great, and the CI also starts to be green! The failures are unrelated and caused by https://github.com/root-project/root/pull/16697. FYI, @ferdymercury.

ferdymercury:
> The failures are unrelated and caused by #16697. FYI, @ferdymercury.

Yep, it's because the sibling roottest PR was overlooked. (Should have been merged simultaneously)

vepadulano:
I have added more comments to the commit descriptions. Once the CI confirms these changes indeed work I will merge. Also, we can merge the CI for ubuntu24.10 directly with this PR.

guitargeek:
Nice! And cool to see that it also works on Fedora 41 with Python 3.13

PR #16838, opened 2024-11-06
[ntuple] allow for writing into TFile subdir

### LINKED ISSUE #14007: Cannot create a RNtuple into a TDirectory
### Check duplicate issues.

- [x] Checked for duplicates

### Description

I am naively trying to create a RNTuple inside a TDirectory, and I fail to do so. Moreover, when looking at the produced TFile with TBrowser, I see some weird recursive structure.

image

### Reproducer

I tried a few permutations of the following, all giving me weird results.

```C++
#include <TFile.h>
#include <TTree.h>
#include <TDirectory.h>
#include <ROOT/RNTuple.hxx>
#undef NDEBUG
#include <cassert>

int writeTree()
{
  TFile* f;
  int x = 0;
  f = new TFile("foo.root", "RECREATE");
  auto* dir = f->mkdir("dir");
  dir->cd();
  auto* tree = new TTree("T", "a Tree");
  tree->Branch("x", &x, "x/D");
  for (int i = 0; i < 10; i++) {
    x = i;
    tree->Fill();
  }
  tree->Write();

  using namespace ROOT::Experimental;
  // Create and fill an RNTuple
  auto wrPt = model->MakeField<float>("pt", 42.0);

  {
    RNTupleWriteOptions options;
    options.SetContainerFormat(ENTupleContainerFormat::kBare);
    auto* bar = f->mkdir("bar");
    bar->cd();
    
    auto ntuple = RNTupleWriter::Append(std::move(model), "f", *f, options);
    bool ok = bar->cd();

    ntuple->Fill();
    ntuple->CommitCluster();
    *wrPt = 24.0;
    ntuple->Fill();
    *wrPt = 12.0;
    ntuple->Fill();
    ntuple->CommitCluster();
  }

  return 0;
}
```

### ROOT version
```
  | Welcome to ROOT 6.28/04                        https://root.cern |
  | (c) 1995-2022, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for macosxarm64 on Nov 01 2023, 10:33:00                   |
  | From tags/v6-28-04-alice3@v6-28-04-3-g186ec88ba4                 |
  | With Apple clang version 15.0.0 (clang-1500.0.40.1)              |
```
### Installation method

aliBuild

### Operating system

macOS Ventura

### Additional context

_No response_

jblomer:
With the patch, `RNTupleWriter::Append()` can take a sub directory in a TFile. On the reading side, the RNTupleReader understands a ntuple name of the form `[/]foo/bar/ntpl` and opens it correctly. Opening from the anchor directly anyway worked.

Fixes #14007

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
@@ -69,8 +69,9 @@ private:
 
    /// Used when the file container turns out to be a bare file
    RResult<RNTuple> GetNTupleBare(std::string_view ntupleName);
-   /// Used when the file turns out to be a TFile container
-   RResult<RNTuple> GetNTupleProper(std::string_view ntupleName);
+   /// Used when the file turns out to be a TFile container. The ntuplePath variable is either the ntuple name
+   /// or an ntuple name preceded by a directory (`myNtuple` or `foo/bar/myNtuple` or `/foo/bar/myNtuple`)
+   RResult<RNTuple> GetNTupleProper(std::string_view ntuplePath);
 
    /// Searches for a key with the given name and type in the key index of the directory starting at offsetDir.
    /// The offset points to the start of the TDirectory DATA section, without the key and without the name and title
--- a/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
@@ -223,11 +223,9 @@ public:
    static std::unique_ptr<RNTupleFileWriter> Recreate(std::string_view ntupleName, std::string_view path,
                                                       EContainerFormat containerFormat,
                                                       const RNTupleWriteOptions &options);
-   /// Add a new RNTuple identified by ntupleName to the existing TFile.
-   static std::unique_ptr<RNTupleFileWriter> Append(std::string_view ntupleName, TFile &file, std::uint64_t maxKeySize);
-   /// Add a new RNTuple identified by ntupleName to the existing directory, which must be a directory in a file.
+   /// The directory parameter can also be a TFile object (TFile inherits from TDirectory).
    static std::unique_ptr<RNTupleFileWriter>
-   Append(std::string_view ntupleName, TDirectory &directory, std::uint64_t maxKeySize);
+   Append(std::string_view ntupleName, TDirectory &fileOrDirectory, std::uint64_t maxKeySize);
 
    RNTupleFileWriter(const RNTupleFileWriter &other) = delete;
    RNTupleFileWriter(RNTupleFileWriter &&other) = delete;
--- a/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
@@ -72,6 +72,12 @@ private:
    /// Used when the file turns out to be a TFile container
    RResult<RNTuple> GetNTupleProper(std::string_view ntupleName);
 
+   /// Searches for a key with the given name and type in the key index of the directory starting at offsetDir.
+   /// The offset points to the start of the TDirectory DATA section, without the key and without the name and title
+   /// of the TFile record (the root directory).
+   /// Return 0 if the key was not found. Otherwise returns the offset of found key.
+   std::uint64_t SearchInDirectory(std::uint64_t &offsetDir, std::string_view keyName, std::string_view typeName);
+
 public:
    RMiniFileReader() = default;
    /// Uses the given raw file to read byte ranges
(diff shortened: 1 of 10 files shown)


tree/ntuple/v7/src/RMiniFile.cxx
```diff
@@ -1105,12 +1105,27 @@ ROOT::Experimental::Internal::RNTupleFileWriter::Recreate(std::string_view ntupl
...
+
+   return std::make_pair(nameAndDir.substr(0, posLastSlash), nameAndDir.substr(posLastSlash + 1));
+}
+
 std::unique_ptr<ROOT::Experimental::Internal::RNTupleFileWriter>
 ROOT::Experimental::Internal::RNTupleFileWriter::Append(std::string_view ntupleName, TFile &file,
                                                         std::uint64_t maxKeySize)
 {
-   auto writer = std::unique_ptr<RNTupleFileWriter>(new RNTupleFileWriter(ntupleName, maxKeySize));
+   auto [dir, name] = SplitNTupleName(ntupleName);
```
pcanal, line 1125: Note that it would also be nice to have:
```
ROOT::Experimental::Internal::RNTupleFileWriter::Append(std::string_view ntupleName, TDIrectory &dir,
                                                        std::uint64_t maxKeySize) 
{
     if (!dir.GetFile())
        throw "We need a file";
   auto writer = std::unique_ptr<RNTupleFileWriter>(new RNTupleFileWriter(name, maxKeySize));
   writer->fFileProper.fFile = dir->GetFile();
   writer->fFileProper.fDirectory = &dir; 
```
(oups the type of `fDirectory` is wrong.

tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
```diff
@@ -104,6 +104,7 @@ class RNTupleFileWriter {
 private:
    struct RFileProper {
       TFile *fFile = nullptr;
+      std::string fDirectory;
```
pcanal, line 107: Either
```suggestion
      TDirectory* fDirectory;
```
or
```suggestion
      std::string fDirectoryPath;
```

jblomer:
@pcanal for the `RBlob` keys of the ntuple within a directory, the current code sets their parent to `kBEGIN`. This imitates what TTree does for the `TBasket` keys of a tree in a sub directory. Is this the expected behavior or should the parent of the `RBlob` / `TBasket` keys be rather the directory into which the tree/ntuple is stored?

tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
```diff
@@ -107,7 +114,10 @@ public:
 
 private:
    struct RFileProper {
+      /// If fDirectory is set, fFile is equal to fDirectory->GetFile()
       TFile *fFile = nullptr;
```
silverweed, line 118: Why do we need this? Can't we use `fDirectory` in either case?

jblomer: `TFile` has some file specific methods that we need, such as `Seek()` and `WriteBuffer()`.

jblomer: Ah, yes that makes sense.

tree/ntuple/v7/src/RNTupleWriter.cxx
```diff
@@ -92,14 +92,20 @@ ROOT::Experimental::RNTupleWriter::Recreate(std::initializer_list<std::pair<std:
...
+   auto file = fileOrDirectory.GetFile();
+   if (!file) {
+      throw RException(R__FAIL("RNTupleWriter only supports writing to a ROOT file. Cannot write into a directory "
+                               "that is not backed by a file"));
+   }
+   if (!file->IsBinary()) {
       throw RException(R__FAIL("RNTupleWriter only supports writing to a ROOT file. Cannot write into " +
-                               std::string(file.GetName())));
+                               std::string(file->GetName())));
+   }
```
hahnjo, line 106: unrelated to this PR, but we should probably have the same checks in `RNTupleParallelWriter::Append`...

hahnjo (commented):
In principle this LGTM, but it creates a bit of an asymmetry between reading and writing: Why do we need to support reading with the recursive directory traversal? Maybe we can just require the user to get the RNTuple anchor from the subdirectory?

jblomer:
> In principle this LGTM, but it creates a bit of an asymmetry between reading and writing: Why do we need to support reading with the recursive directory traversal? Maybe we can just require the user to get the RNTuple anchor from the subdirectory?

This is true, but we need the string-based API for RDataFrame. In the end, I'm not sure if the symmetry gap is an issue.

hahnjo (approved):
> This is true, but we need the string-based API for RDataFrame.

Oh, and also the `RNTupleProcessor`!

LGTM, I can take care of the `RNTupleParallelWriter` in a follow-up PR.

PR #16845, opened 2024-11-07
[v634][ntuple] Enforce valid naming upon field creation

enirolf:
Backport of #16836.


hahnjo (approved):
(maybe not needed anymore, but in any case...)

PR #16850, opened 2024-11-07
[ntuple] Enforce RNTuple naming rules

vepadulano:
RNTuple naming rules are enforced when creating a field or a new RNTuple for writing. The function responsible for doing this check is extracted into a utility header.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
@@ -475,8 +475,6 @@ public:
    /// Checks if the given type is supported by RNTuple. In case of success, the result vector is empty.
    /// Otherwise there is an error record for each failing sub field (sub type).
    static std::vector<RCheckResult> Check(const std::string &fieldName, const std::string &typeName);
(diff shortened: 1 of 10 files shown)


enirolf:
According to #16813 we disallow all control characters, so for example also carriage returns, null terminators, etc. Wikipedia has a good overview: https://en.wikipedia.org/wiki/Unicode_control_characters.

pcanal:
Should we consider using `https://en.cppreference.com/w/c/string/byte/isprint` ?

vepadulano:
> Should we consider using https://en.cppreference.com/w/c/string/byte/isprint and/or friends?

I believe we should not, since it depends on the locale.

pcanal:
Humm ... The current version already accept an arbitrary set of "illegal" characters.  I think we should make this set as small as practical.  Using [iscntrl](https://en.cppreference.com/w/cpp/string/byte/iscntrl) and/or the equivalent set of `<` comparison would be a good start.

jblomer (approved):
Thanks a lot! LGTM.

Please squash or rebase into one commit, so that we don't have a commit with failing tests.

PR #16857, opened 2024-11-07
[ntuple] Add zip timer to `RPageSinkBuf`

hahnjo:
There is a zip timer in `RPagePersistentSink`, used in `RPageSink{File,Daos}::CommitPageImpl`, but this code path not taken with buffered writing.

### DIFF
--- a/tree/ntuple/v7/src/RPageSinkBuf.cxx
+++ b/tree/ntuple/v7/src/RPageSinkBuf.cxx
@@ -37,8 +37,11 @@ ROOT::Experimental::Internal::RPageSinkBuf::RPageSinkBuf(std::unique_ptr<RPageSi
    fMetrics = Detail::RNTupleMetrics("RPageSinkBuf");
    fCounters = std::make_unique<RCounters>(RCounters{
       *fMetrics.MakeCounter<Detail::RNTuplePlainCounter *>("ParallelZip", "", "compressing pages in parallel"),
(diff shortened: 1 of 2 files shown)


tree/ntuple/v7/src/RPageSinkBuf.cxx
```diff
@@ -200,7 +206,10 @@ void ROOT::Experimental::Internal::RPageSinkBuf::CommitPage(ColumnHandle_t colum
       // Make sure the page buffer is not aliased so that we can free the uncompressed page.
       config.fAllowAlias = false;
       config.fBuffer = zipItem.fBuf.get();
-      sealedPage = SealPage(config);
+      {
+         Detail::RNTupleAtomicTimer timer(fCounters->fTimeWallZip, fCounters->fTimeCpuZip);
+         sealedPage = SealPage(config);
+      }
```
jblomer, line 212: Hm, at least for the wall time, this may be confusing because tasks can run in parallel.

hahnjo: Right, reading also only counts the time in `RPageSourceFile::LoadPageImpl` when executed in a blocking fashion... I removed the time from the IMT code path for now (I'm mostly interested in the non-IMT case for parallel writing), but it would be great to somehow get observability into background (de)compression as well.

jblomer (approved):

PR #16858, opened 2024-11-07
New interpCode=6 for PiecewiseInterpolation: Multiplicative Polynominal Interpolation with Linear Extrapolation

will-cern:
ATLAS Higgs group has a need for interpCode 4 (poly interp with linear extrap) but in a multiplicative form rather than additive. The new interpCode 6 adds this option. 

I also tried to update the documentation in PiecewiseInterpolation to explain all the interpcodes (I skipped interpCode 3 since it looks like its actually identical to interpCode 2 at the moment, and anyway I don't see it used in the wild atm).


guitargeek (approved):
LGTM, thank you very much for the new feature, the detailed documentation, and even providing a tests!

PR #16859, opened 2024-11-07
[ntuple] Improve RPagePool

jblomer:
Improve the lookup complexity for pages in the page pool from linear to constant in well-behaved cases, i.e. if there is a small number of pages per column and cluster. Some smaller cleanups around the RPage/RPagePool logic.

### DIFF
--- a/tree/ntuple/v7/src/RPagePool.cxx
+++ b/tree/ntuple/v7/src/RPagePool.cxx
@@ -26,12 +26,7 @@ void ROOT::Experimental::Internal::RPagePool::AddPage(const RPage &page, const R
 {
    assert(fLookupByBuffer.count(page.GetBuffer()) == 0);
    fLookupByBuffer[page.GetBuffer()] = index;
-   auto itr = fLookupByKey.find(key);
-   if (itr == fLookupByKey.end()) {
-      fLookupByKey.emplace(key, std::vector<size_t>({index}));
-   } else {
-      itr->second.emplace_back(index);
-   }
+   fLookupByKey[key].emplace_back(index);
 }
 
 ROOT::Experimental::Internal::RPageRef ROOT::Experimental::Internal::RPagePool::RegisterPage(RPage page, RKey key)
--- a/tree/ntuple/v7/src/RPagePool.cxx
+++ b/tree/ntuple/v7/src/RPagePool.cxx
@@ -18,22 +18,37 @@
 
 #include <TError.h>
 
+#include <algorithm>
 #include <cstdlib>
 #include <utility>
+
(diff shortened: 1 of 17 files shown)


tree/ntuple/v7/src/RPagePool.cxx
```diff
@@ -18,81 +18,100 @@
...
-         fPages.resize(N - 1);
-         fPageInfos.resize(N - 1);
-         fReferences.resize(N - 1);
+   const auto idx = fLookupByBuffer.at(page.GetBuffer());
+   const auto N = fEntries.size();
+
+   if (--fEntries[idx].fRefCounter == 0) {
+      fLookupByBuffer.erase(page.GetBuffer());
+
+      auto itrPageSet = fLookupByKey.find(fEntries[idx].fKey);
```
silverweed, line 65: We might want to assert that the ref counter is non-negative

hahnjo (commented):
This is great! Some stylistic comments inline.

It would be interesting to run the limits test, especially `Limits_ManyFields`, `Limits_ManyPages`, and `Limits_ManyPagesOneEntry`. It's possible that this PR addresses the quadratic complexity seen in there.

jblomer:
No changes to the "many pages" test. The "many fields" unit test got significantly faster. The overall complexity is still super-linear but much more benign.

hahnjo:
> No changes to the "many pages" test. The "many fields" unit test got significantly faster. The overall complexity is still super-linear but much more benign.

Ah right, now I remember that I had already profiled this before: The "many pages" test is actually bound by `RPageRange::Find`, which has a `TODO` to use binary search. The case we are speeding up here is many pages distributed over many fields, in which case performance was bound by the page pool.

tree/ntuple/v7/test/ntuple_limits.cxx
```diff
@@ -16,11 +16,11 @@
...
 TEST(RNTuple, DISABLED_Limits_ManyFields)
 {
-   // Writing and reading a model with 40k integer fields takes around 2s and seems to have more than linear
-   // complexity (80k fields take 9s).
+   // Writing and reading a model with 100k integer fields takes around 2s and seems to have more than linear
+   // complexity (200k fields take 7.5s).
    FileRaii fileGuard("test_ntuple_limits_manyFields.root");
 
-   static constexpr int NumFields = 40'000;
+   static constexpr int NumFields = 100'000;
```
hahnjo, line 23: This is fantastic that we can now process models with 100k fields in reasonable time!

hahnjo (approved):
LGTM, thanks!

PR #16861, opened 2024-11-07
[CPyCppyy] Fix regression in passing buffers through `void *`

guitargeek:
Credit goes to Wim Lavrijsen (@wlav).

See also: https://github.com/wlav/cppyy/issues/272.

Following up on #16816, hopefully fixing the problem without sacrificing any features.


bindings/pyroot/cppyy/cppyy/test/test_regression.py
```diff
@@ -9,8 +9,12 @@ class TestREGRESSION:
...
         import cppyy
 
-        def stringpager(text, cls=cls):
-            cls.helpout.append(text)
+        if sys.hexversion < 0x30d0000:
+            def stringpager(text, cls=cls):
+                cls.helpout.append(text)
+        else:
+            def stringpager(text, title='', cls=cls):
+                cls.helpout.append(text)
```
vepadulano, line 17: I'm just curious to understand why the extra `title` kwarg, which looks like it's unused in the function

vepadulano (approved):
Just a minor curiosity, rest looks good thanks!

bindings/pyroot/cppyy/cppyy/test/test_regression.py
guitargeek: That's more of a question to @wlav who wrote this. But for ROOT it makes no difference at this point.

But let's try to run all these tests also in our CI in the future!

wlav: The interface of the `pydoc` pager changed in 3.13 with the addition of that title argument, but I'd rather have the test capture the same information only, i.e. the help-text body, than introduce differences in test behavior between versions. (If anything needs changing, this test needs a teardown to restore the old pager.)

PR #16869, opened 2024-11-08
[PyROOT] Don't add gROOT and other globals manually to PyROOT via C++

guitargeek:
This is not necessary if we consider that `gROOT` is actually a
preprocessor macro that aliases to `ROOT::GetROOT()`, which we can use
directly in Python via cppyy. Same with the `gInterpreter`.

Also, a second commit refactors the PyROOT C++ code to avoid some boilerplate by having too granular translation units.


bindings/pyroot/pythonizations/python/ROOT/__init__.py
```diff
@@ -176,6 +176,8 @@ def find_spec(self, fullname: str, path, target=None) -> ModuleSpec:
 # Register cleanup
 import atexit
 
+# Private handle to gROOT for later cleanup
+_gROOTForCleanup = cppyy.gbl.ROOT.GetROOT()
```
vepadulano, line 180: Can't we initialize the variable directly within the `cleanup` function?

vepadulano (changes requested):
Cool thanks! One small comment from my side

bindings/pyroot/pythonizations/python/ROOT/__init__.py
guitargeek: Good idea! I will check if this is possible once the CI is green in a first pass

guitargeek: I realized we can get this directly from the facade, which also has it.

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/__init__.py
```diff
@@ -45,7 +43,7 @@ def inject_rbatchgenerator(ns):
 
 from ._gnn import RModel_GNN, RModel_GraphIndependent
 
-hasRDF = "dataframe" in gROOT.GetConfigFeatures()
+hasRDF = "dataframe" in cppyy.gbl.ROOT.GetROOT().GetConfigFeatures()
```
pcanal, line 46: Is this change necessary?  If it is, does it also imply that user code currently using `gROOT` would need to change?

bindings/pyroot/pythonizations/src/PyROOTWrapper.cxx
```diff
@@ -59,9 +39,4 @@ void PyROOT::Init()
 
    // Memory management
    gROOT->GetListOfCleanups()->Add(&GetRegulatorCleanup());
-
-   // Bind ROOT globals that will be needed in ROOT.py
-   AddToGlobalScope("gROOT", gROOT, gROOT->IsA()->GetName());
-   AddToGlobalScope("gSystem", gSystem, gSystem->IsA()->GetName());
-   AddToGlobalScope("gInterpreter", gInterpreter, gInterpreter->IsA()->GetName());
```
pcanal, line 66: `gInterpreter` is also a macro, after this can you still use `gInterpreter` on the `Python` prompt?

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/__init__.py
guitargeek: Yes, it's necessary because we don't import `gROOT` from `libROOTPythonizations` anymore (since I removed the reference in there). However, users don't have to change anything, because `gROOT` is still set in the ROOT facade. However, this code here is not using the ROOT modulde/facade. It's an implementation detail of the pythonizations, which also used the internal `libROOTPythonizations` before. Hence the change is required here but not in usercode, which does `ROOT.gROOT`.

bindings/pyroot/pythonizations/src/PyROOTWrapper.cxx
guitargeek: It does, even with vanilla cppyy! I guess there is some magic inside cppyy to make this work.

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/__init__.py
pcanal: I see
> from cppyy.gbl import gSystem
then I suspect
```
from cppyy.gbl import gROOT
```
would also work, right?
or this would also work, right?:
```
in cppyy.gbl.gROOT.GetConfigFeatures()
```
(not asking to change the code, just checking I understand correctly)

vepadulano (approved):
Thanks, LGTM! But let's wait for the CI to be fully green, and also it would be nice to make clang-format happy

guitargeek:
Thank you so much Vincenzo for fixing it up!

PR #16870, opened 2024-11-08
[gpad] Check gPad in bbox methods

linev:
1. Adjust TPad methods - check `gPad` as well
2. Remove double check in TEllipse
3. Replace `return (p)` with `return p`


couet (approved):
LGTM

PR #16599, opened 2024-10-04
[pyroot] fix clang warning due to changes in PyObject

silverweed:
[This commit](https://github.com/python/cpython/commit/ea2c0016502472aa8baa3149050ada776d17a009#diff-87272721a5cf1cd9915d6f503f6a7bbefa2f26c935c7ce83ca78706afd0ad05a) changed a PyObject field from a ssize_t to a union, so clang now wants parentheses around the object (this prevents building root with -Werror).
This was introduced in 3.12.0-beta1, aka PY_VERSION_HEX == 0x30c00b1.

PR #16873, opened 2024-11-08
[pyroot] fix clang warning due to changes in PyObject

guitargeek:
Re-applying a commit by @silverweed that was accidentally reverted.

Original PR: #16599.

This patch will now be upstreamed, so it won't be accidentally reverted again:
https://github.com/wlav/CPyCppyy/pull/43


vepadulano (approved):
Thanks for taking care of this!

PR #16877, opened 2024-11-10
[ntuple] update write option defaults

jblomer:
Consistently use MiB (units based on power of 2).


vepadulano (approved):
LGTM, but `tutorial-v7-ntuple-ntpl013_staged` is timing out on Windows, I would maybe restart the CI run and check that it was just a hiccup.

tree/ntuple/v7/doc/tuning.md
```diff
@@ -5,15 +5,15 @@ A cluster contains all the data of a given event range.
...
+The default cluster target size is 128 MiB of compressed data.
 The default can be changed by the `RNTupleWriteOptions`.
 The default should work well in the majority of cases.
 In general, larger clusters provide room for more and larger pages and should improve compression ratio and speed.
 However, clusters also need to be buffered during write and (partially) during read,
 so larger clusters increase the memory footprint.
 
 A second option in `RNTupleWriteOptions` specifies the maximum uncompressed cluster size.
-The default is 1 GiB.
+The default is 10x the default cluster target size, i.e. ~1.2 GiB.
```
pcanal, line 16: Why not
```suggestion
The default is 8x the default cluster target size, i.e. 1.0 GiB.
```
?
Or why the factor `10`?

pcanal: Was the reason for choosing `10` something else than 'get to a round number'?  If it is just to get to a round number we should switch to 8, if it is something else we should explain it here.

jblomer: Well, the largest compression factor we have seen in an experiment EDM so far was MiniAOD with a compression factor >9.

jblomer:
Since we have not yet updated measurements confirming that the new settings don't cause a performance degradation with the AGC benchmark, we remove the 6.34 milestone. We should settle for the final default settings before releasing the RNTuple v1 AGC dataset.

PR #16893, opened 2024-11-11
[RF] Added thorough parameter study for RooCBShape and RooJohnson

Patrik1van:


roofit/roofit/src/RooJohnson.cxx
```diff
@@ -13,31 +13,70 @@
...
-  \;\exp\left[-\frac{1}{2} \left(\gamma + \delta \sinh^{-1}\left(\frac{x-\mu}{\lambda}\right) \right)^2\right].
-\f]
-
-It is often used to fit a mass difference for charm decays, and therefore the variable \f$ x \f$ is called
-"mass" in the implementation. A mass threshold allows to set the PDF to zero to the left of the threshold.
-
-###References:
-Johnson, N. L. (1949). *Systems of Frequency Curves Generated by Methods of Translation*. Biometrika **36(1/2)**, 149–176. [doi:10.2307/2332539](https://doi.org/10.2307%2F2332539)
-
-\image html RooJohnson_plot.png
```
guitargeek, line 38: Why not use this plot anymore? I liked it, because it nicely illustrated the shape of the RooJohnson for different values of the parameter in one canvas:
https://root.cern/doc/v632/classRooJohnson.html

I think even with your added parameter studies, it's still a nice complementary plot.

guitargeek (changes requested):
Thank you very much @Patrik1van for the nice improvements to the docs!

If you still have time to work on this, I have a few change requests:

1. Could you please provide the code that is used to generate these plots? Maybe put it in a code block of the PR description. It would be good if we can easily reproduce these plots in case they need a different formatting, colors, etc.
2. Can you give the png files less generic names? Right now, `combined_sigmas.png` for example doesn't tell me at all that this is related to the RooCBShape. Maybe call it `RooCBShape_sigmas.png`? Same pattern for the other files.
3. The existing [RooJohnson documentation](https://root.cern/doc/v632/classRooJohnson.html) uses thicker line widths, which I think is very good. Can you do the same?


And when you create an updated branch with these changes, can you please [force-push](https://stackoverflow.com/questions/32346624/what-happens-if-you-force-push-to-a-branch-with-an-existing-pull-request) it to the branch with the same name `documentation_changes_cb_john` so that you don't have to open a new PR? Thank you very much!

guitargeek:
Closing stale PR because there was no reply to the change requests. Please feel free to open a new PR if you want to resume working on this!

PR #16900, opened 2024-11-12
[ntuple] Fixup `RNTupleIndex` error messages

enirolf:
The error message format was inconsistent with (most of) the other RNTuple error messages, which are typically not capitalized and don't end with a period.

In addition, the error message in the constructor when a join field is not present has been expanded to now also contain the name of the RNTuple.


vepadulano (approved):
LGTM thanks!

PR #16907, opened 2024-11-12
[ntuple] move RColumnElement out of anonymous namespace

silverweed:
to avoid spurious warnings by some compilers, such as:
```
‘void {anonymous}::RColumnElementQuantized<T>::SetValueRange(double, double) [with T = double]’ declared ‘static’ but never defined [-Wunused-function]
```

### DIFF
--- a/tree/ntuple/v7/src/RColumnElement.hxx
+++ b/tree/ntuple/v7/src/RColumnElement.hxx
@@ -311,10 +311,6 @@ inline void CastZigzagSplitUnpack(void *destination, const void *source, std::si
       dst[i] = sval;
    }
 }
-} // namespace
-
-// anonymous namespace because these definitions are not meant to be exported.
-namespace {
 
 using ROOT::Experimental::EColumnType;
 using ROOT::Experimental::Internal::RColumnElementBase;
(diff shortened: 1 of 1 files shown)


tree/ntuple/v7/src/RColumnElement.hxx
```diff
@@ -313,8 +313,7 @@ inline void CastZigzagSplitUnpack(void *destination, const void *source, std::si
 }
 } // namespace
 
-// anonymous namespace because these definitions are not meant to be exported.
-namespace {
+namespace ROOT::Experimental::Internal {
```
pcanal, line 316: Indeed, it is quite unusual to have an anonymous namespace in a header file.  I guess it would result in any symbol/function defined in there to be actually declared/defined multiple times (one per source file that include this header) with different actual names

pcanal: We should probably also review line 57 of this file and
```
bindings/pyroot/cppyy/CPyCppyy/src/DeclareConverters.h:namespace {
bindings/pyroot/cppyy/CPyCppyy/src/DeclareExecutors.h:namespace {
bindings/pyroot/cppyy/cppyy/test/stltypes.h:namespace {
io/io/test/io_test.hxx:namespace {
math/mathcore/src/MixMaxEngineImpl.h:namespace { 
tree/ntuple/v7/test/SimpleCollectionProxy.hxx:namespace {
```

PR #16911, opened 2024-11-12
[6.34] Hide RDF Numba Pythonizations

vepadulano:
Since they are only available for Define and Filter operations, the functionality is not ready yet for wider use.


dpiparo (approved):
Thanks, I would have missed this. Let's remember to document what is possible with Numba very clearly for users but also for ourselves.

PR #16913, opened 2024-11-12
Fix buffer overflow when opening the Fit Panel

langufo:
Fedora 40 build for 6.32.06 crashes when opening the parameter dialog of the Fit Panel
with a "buffer overflow detected" message.

Such behavior is caused by calls to `strlcpy` with an incorrect size of the destination buffer,
which this PR fixes.

## Checklist:

- [x] tested changes locally


bellenot (approved):
LGTM, thanks!

PR #16914, opened 2024-11-13
[PyROOT] Move interpreter initialization on the Python side

guitargeek:
Continue the effort to do less in our CPython extension and more in our Python module.


guitargeek:
What is so different with Ubuntu 24.04? I don't know... But this attempt didn't work out.

Issue #15230, opened 2024-04-14
[CI] clingtest=ON when cling touched


ferdymercury:
### Explain what you would like to see improved and how.

When a PR touches code within cling subdir, it would be great if the CI turns option clingtest ON automatically.

### ROOT version

Any

### Installation method

Any

### Operating system

Any

### Additional context

https://github.com/root-project/root/pull/9272#issuecomment-2053640480

Issue #6957, opened 2020-12-16
Does roottest run `cling-test`?


SimeonEhrig:
I have question. Does `roottest` runs `cmake --build . --target cling-test` when the cling interpreter code is changed? From time to time I see, that commits are merged regardless of whether the CI fails.
If I understand Cling's workflow correctly, changes are integrated and tested in ROOT and automatically streamed down to Cling after the merge in ROOT. So, if `roottest` does not run the integrated tests in cling, it is possible that the Cling CI is bypassed, or I'm wrong?

Axel-Naumann:
`cling-test` is part of the test suite run with `ctest` if ROOT is configured with `-Dclingtest=On`. You can trigger Jenkins to do that by saying `@phsft-bot build with flags -Dclingtest=On`. But I agree it would make sense to turn it on by default for PR and nightly builds.

SimeonEhrig:
@Axel-Naumann That sounds like a great idea.

dpiparo:
@hahnjo @devajithvs @vgvassilev I can enable these tests for 2-3 platforms in master: what do you think?

vgvassilev:
Sounds good to me.

hahnjo:
`clingtest` exports many more symbols from `libCling`, for testing purposes (see the code in `interpreter/CMakeLists.txt`). This means our (patched) LLVM and Clang will interfere with other LLVM installations used by dependent libraries. IIRC this can make tests fail!

dpiparo:
What then could be a good way to make progress about this issue?

vgvassilev:
> `clingtest` exports many more symbols from `libCling`, for testing purposes (see the code in `interpreter/CMakeLists.txt`). This means our (patched) LLVM and Clang will interfere with other LLVM installations used by dependent libraries. IIRC this can make tests fail!

We will build that version of ROOT only to run clingtest, right? We still will test ROOT the regular way.

PR #16917, opened 2024-11-13
[ci] build with clingtest flag active if interpreter folder was touched

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/15230
Fixes https://github.com/root-project/root/issues/6957

Sibling PR in https://github.com/root-project/roottest/pull/1232

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -247,6 +247,7 @@ jobs:
 
 
   build-windows:
+    needs: [ git_diff ]
     # For any event that is not a PR, the CI will always run. In PRs, the CI
     # can be skipped if the tag [skip-ci] or [skip ci] is written in the title.
     if: |
@@ -301,6 +302,7 @@ jobs:
         run: "C:\\setenv.bat ${{ matrix.target_arch }} &&
               python .github/workflows/root-ci-config/build_root.py
                     --buildtype       ${{ matrix.config }}
+                    --clingtest       ${{ needs.git_diff.outputs.interpreter_changed }}
                     --platform        windows10
                     --incremental     $INCREMENTAL
                     --base_ref        ${{ github.base_ref }}
@@ -344,6 +346,7 @@ jobs:
         run: "C:\\setenv.bat ${{ matrix.target_arch }} &&
               python .github/workflows/root-ci-config/build_root.py
                     --buildtype    ${{ matrix.config }}
+                    --clingtest    ${{ needs.git_diff.outputs.interpreter_changed }}
                     --platform     windows10
                     --incremental  false
                     --base_ref     ${{ github.ref_name }}
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -77,7 +77,36 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  git_diff:
+    name: Detect changes in Cling
+    # https://www.meziantou.net/executing-github-actions-jobs-or-steps-only-when-specific-files-change.htm
+    runs-on: ubuntu-latest
+    # Declare output for next jobs
+    outputs:
+      interpreter_changed: ${{ steps.check_file_changed.outputs.interpreter_changed }}
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        # Checkout as many commits as needed for the diff
+        fetch-depth: 2
+    - shell: pwsh
+      id: check_file_changed
+      run: |
+        # Diff HEAD with the latest commit of master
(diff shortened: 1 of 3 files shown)


.github/workflows/root-ci.yml
```diff
@@ -77,6 +77,29 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  conditional_job_check_files:
```
dpiparo, line 80: I do not understand why the CI jobs do not start: is there perhaps a mistake in the yaml?

.github/workflows/root-ci.yml
```diff
@@ -364,6 +388,9 @@ jobs:
             overrides: ["LLVM_ENABLE_ASSERTIONS=On"]
           - image: alma9
             overrides: ["LLVM_ENABLE_ASSERTIONS=On", "CMAKE_BUILD_TYPE=Debug"]
+          - image: alma9-clingtest
```
dpiparo, line 391: Where is this image coming from?

dpiparo (approved):
This PR is a great attempt to introduce a useful new feature. It seems some information is still needed before starting to test it thoroughly.

dpiparo:
@vgvassilev

.github/workflows/root-ci.yml
ferdymercury, line 391:
```suggestion
          - image: alma9
          - is_special: true
```

vgvassilev:
These changes are too advanced for me. Maybe @mcbarton could take a look...

.github/workflows/root-ci.yml
mcbarton: @dpiparo @ferdymercury If you click on checks from github you'll see that the one workflow has a warning flag next to it. If you click on the workflow you'll see the following error message
```
[Invalid workflow file: .github/workflows/root-ci.yml#L393](https://github.com/root-project/root/actions/runs/12096633306/workflow)
You have an error in your yaml syntax on line 393
```
You seem to have made an error at line 393 of the workflow. If you can't work out what it is I will investigate further.

interpreter/cling/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ else(WIN32)
   # support of earlier cmake versions will be removed soon
   cmake_minimum_required(VERSION 3.5)
 endif(WIN32)
-
+
```
ferdymercury, line 10:
```suggestion

```

ferdymercury:
It's working now.

hahnjo (changes requested):
This looks really (too) complicated and IMHO is not what we actually want:
 * `check-cling` doesn't seem to work in-tree on Windows.
 * On Apple Silicon, we (still) cannot catch exceptions from interpreted code and therefore many (all?) `NullDeref` tests fail. We probably want to start with an x86 machine.
 * For Linux, as I predicted in https://github.com/root-project/root/issues/6957#issuecomment-2076571672, exposing the LLVM symbols from `libCling` interferes with other libraries, in this case TensorFlow (but I think we have seen problems with Mesa in the past as well).
 * `roottest-root-meta-countIncludePaths` checks the number of include paths, which is larger with `clingtest=ON` to make the tests pass.

In general, as was discussed before, we still want to run all PR builds because this is what matters for ROOT's functionality. On top of this, we want *extra* builds with `clingtest` that only execute `check-cling`. This can be accomplished much easier with a separate workflow that triggers on `pull_request` with `paths`.

ferdymercury:
>     * On Apple Silicon, we (still) cannot catch exceptions from interpreted code and therefore many (all?) `NullDeref` tests fail. We probably want to start with an x86 machine.

Changed.

>     * For Linux, as I predicted in [Enable `cling-test` for nightly, PR builds #6957 (comment)](https://github.com/root-project/root/issues/6957#issuecomment-2076571672), exposing the LLVM symbols from `libCling` interferes with other libraries, in this case TensorFlow (but I think we have seen problems with Mesa in the past as well).
> 
>     * `roottest-root-meta-countIncludePaths` checks the number of include paths, which is larger with `clingtest=ON` to make the tests pass.

This should be fixed in my opinion in the CMakeLists using an `if(clingtest)` --> pass argument of extra include paths. (build_root.py should forward this flag to roottest as -Dclingtest when cmaking).
 
> This can be accomplished much easier with a separate workflow that triggers on `pull_request` with `paths`.

I hope someone can implement that variant soon ;) and I will close this PR then.

ferdymercury:
Closing in favour of https://github.com/root-project/root/pull/18127

PR #16920, opened 2024-11-13
[hist] Remove complex-arg functions from context menu

ferdymercury:
Fixes https://its.cern.ch/jira/browse/ROOT-10883

It does not make much sense to use a context menu and then copy from the terminal the hex pointer 0x123445 int-value from other histogram pointers, you would directly use the terminal then and tab-completion.

Sibling PR in https://github.com/root-project/roottest/pull/1216

### DIFF
--- a/hist/hist/inc/TH1.h
+++ b/hist/hist/inc/TH1.h
@@ -189,7 +189,7 @@ public:
 
    virtual Bool_t   Add(TF1 *h1, Double_t c1=1, Option_t *option="");
    virtual Bool_t   Add(const TH1 *h1, Double_t c1=1);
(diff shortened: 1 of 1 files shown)


bellenot (approved):
LGTM, thanks!

bellenot:
@ferdymercury FYI, if ROOT and rootest PRs have to be checked together, they should have the same branch name, so you can run the proper roottest with its matching changes in master

dpiparo:
yes I think the root PR was created before the roottest PR, but it's not a problem at all (thanks for these nice changes!). I am cancelling this build and restarting it in order for it to  pick the sibling PR.

PR #16923, opened 2024-11-13
[ntuple] Reduce includes of RColumnElement.hxx

hahnjo:
This internal header should only be included once into the `ROOTNTuple` library and as little as possible in the tests because it defines symbols in anonymous namespaces, to avoid ODR violations in the test `ntuple_endian.cxx`.

Achieve this by moving `kTestFutureType` to `RNTupleUtil.hxx`, next to other definitions used for compatibility tests. This leaves includes in `RColumnElement.cxx` (for the `ROOTNTuple` library), as well as the two tests `ntuple_endian.cxx` and `ntuple_packing.cxx` that unit test individual instantiations. This should also avoid compiler warnings seen recently.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
@@ -267,6 +267,9 @@ auto MakeAliasedSharedPtr(T *rawPtr)
    return std::shared_ptr<T>(fgRawPtrCtrlBlock, rawPtr);
 }
 
+inline constexpr EColumnType kTestFutureType =
+   static_cast<EColumnType>(std::numeric_limits<std::underlying_type_t<EColumnType>>::max() - 1);
+
 inline constexpr ENTupleStructure kTestFutureFieldStructure =
    static_cast<ENTupleStructure>(std::numeric_limits<std::underlying_type_t<ENTupleStructure>>::max() - 1);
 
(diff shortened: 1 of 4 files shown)


pcanal:
I think am guessing that in order for the warnings (unused static/local declaration), this would need that all 3 of the source files (`RColumnElement.cxx`, `ntuple_endian.cxx` and `ntuple_packing.cxx`) each uses *all* the declaration in the header files, isn't it?

hahnjo:
> I think am guessing that in order for the warnings (unused static/local declaration), this would need that all 3 of the source files (`RColumnElement.cxx`, `ntuple_endian.cxx` and `ntuple_packing.cxx`) each uses _all_ the declaration in the header files, isn't it?

Yes, that's currently the case and I think that's the *only* reason why a TU should include `RColumnElement.hxx`; all other uses should go via `RColumnElementBase`.

silverweed (approved):
LGTM, thanks

jblomer (approved):

pcanal:
> all other uses should go via RColumnElementBase.

@hahnjo I am confused because the above conflicts with my reading of

> This leaves includes ... as well as the two tests ntuple_endian.cxx and ntuple_packing.cxx. 

Which seems to say that those 2 tests still have a (unwanted) `#include "RColumnElement.hxx"`?
Shouldn't we also remove this includes?

PR #16924, opened 2024-11-13
[core] fix bug in tab completion with cv qualifiers

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-6636 reported by @olifre

Sibling PR in https://github.com/root-project/roottest/pull/1218

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/rint/src/TTabCom.cxx
+++ b/core/rint/src/TTabCom.cxx
@@ -874,11 +874,9 @@ TString TTabCom::DetermineClass(const char varName[])
    if (type.EndsWith(" *const"))
       type.Remove(type.Length() - 5);
    // we strip off left qualifiers
-   if (type.BeginsWith("const volatile "))
-      type.Remove(0, 15);
-   else if (type.BeginsWith("const "))
+   if (type.BeginsWith("const "))
       type.Remove(0, 6);
-   else if (type.BeginsWith("volatile "))
+   if (type.BeginsWith("volatile ")) // this automatically takes care of variant "const volatile"
       type.Remove(0, 9);
 
 cleanup:
--- a/core/rint/src/TTabCom.cxx
+++ b/core/rint/src/TTabCom.cxx
@@ -814,6 +814,7 @@ TString TTabCom::DetermineClass(const char varName[])
    assert(varName != nullptr);
    IfDebug(std::cerr << "DetermineClass(\"" << varName << "\");" << std::endl);
 
+   // @todo This is a long ugly workaround, replace with access via gInterpreter
    TString outf = ".TTabCom-";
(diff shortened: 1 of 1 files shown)


hahnjo:
This should definitely have (a) new test(s) added...

ferdymercury:
Just added a test.

ferdymercury:
Seems fixed now.

core/rint/src/TTabCom.cxx
```diff
@@ -868,10 +869,17 @@ TString TTabCom::DetermineClass(const char varName[])
...
+   // we start stripping off right "const"
+   if (type.EndsWith(" *const"))
       type.Remove(type.Length() - 5);
+   // we strip off left qualifiers
+   if (type.BeginsWith("const volatile "))
+      type.Remove(0, 15);
+   else if (type.BeginsWith("const "))
+      type.Remove(0, 6);
+   else if (type.BeginsWith("volatile "))
+      type.Remove(0, 9);
```
hahnjo, line 882: This can be simplified to first check for `const ` and then for `volatile `, without having to think about the combined case

core/rint/src/TTabCom.cxx
```diff
@@ -868,10 +869,17 @@ TString TTabCom::DetermineClass(const char varName[])
    type.ReadToDelim(file1, ')');
    IfDebug(std::cerr << type << std::endl);
 
-   // new version of CINT returns: "class TClassName*const)0x12345"
-   // so we have to strip off "const"
-   if (type.EndsWith("const"))
+   // new version of Cling returns: "qualifiers TClassName *const) 0x12345"
+   // we start stripping off right "const"
+   if (type.EndsWith(" *const"))
       type.Remove(type.Length() - 5);
```
hahnjo, line 875: Note that this is still only stripping off the last five characters, ie `const`. Why do we need to check ` *` in the beginning?

hahnjo (commented):
Some comments from my side. @dpiparo can I ask you to also have a look here?

core/rint/src/TTabCom.cxx
ferdymercury: To avoid problems if one wants to define a class name called e.g. "Reconst", so looking for the space ensures that it's a qualifier rather than the end of the classname.
Then, we do not want to remove the * from the type, since we want to know if it's a pointer or not, to verify if it should go with -> or .

ferdymercury: Probably one could also just check for `*const` but I thought it was more clean / understandable for the developers reading this code to also check for the space.

dpiparo (approved):
LGTM.

PR #16928, opened 2024-11-13
[metacling] Prevent nullptr access if compiler instance was not created

ferdymercury:
# This Pull request:

## Changes or fixes:

exit with error1 code instead of crashing

Fixes https://its.cern.ch/jira/browse/ROOT-10239

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/metacling/src/TCling.cxx
+++ b/core/metacling/src/TCling.cxx
@@ -1536,6 +1536,8 @@ TCling::TCling(const char *name, const char *title, const char* const argv[], vo
                                                        llvmResourceDir, extensions,
                                                        interpLibHandle);
 
+   if (!fInterpreter->getCI()) // Compiler instance could not be created. See https://its.cern.ch/jira/browse/ROOT-10239
+      exit(1);
    // Don't check whether modules' files exist.
    fInterpreter->getCI()->getPreprocessorOpts().DisablePCHOrModuleValidation =
       DisableValidationForModuleKind::All;
--- a/core/metacling/src/TCling.cxx
+++ b/core/metacling/src/TCling.cxx
@@ -1536,8 +1536,10 @@ TCling::TCling(const char *name, const char *title, const char* const argv[], vo
                                                        llvmResourceDir, extensions,
                                                        interpLibHandle);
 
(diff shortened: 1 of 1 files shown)


dpiparo:
this is good improvement: thanks! I was wondering whether exiting is the best course of actions or if an exception or a clear error message could be of help. Now that the location in the code has been identified, I can also re-open a PR including the commit included in this one and enhance the visibility of the problem for the user (which is a simple thing to do, at this point...)

ferdymercury:
I tried first throwing a runtime_exception, but the problem is that this class has an exception handler which auto-dumps a full stack trace onto stdout, which is exactly what we didn't want :S So exit(1) seemed cleaner since the error message was already printed before.

dpiparo:
Ah, I see. Thanks. Can we maybe go for a last "cerr" before the exit? Then for me the PR is ready to merge for me.

dpiparo:
Thanks a lot!

core/metacling/src/TCling.cxx
```diff
@@ -1536,6 +1536,10 @@ TCling::TCling(const char *name, const char *title, const char* const argv[], vo
                                                        llvmResourceDir, extensions,
                                                        interpLibHandle);
 
+   if (!fInterpreter->getCI()) { // Compiler instance could not be created. See https://its.cern.ch/jira/browse/ROOT-10239
+      std::cerr << "Exiting now since compiler instance is not available." << std::endl;
+      exit(EXIT_FAILURE);
```
vgvassilev, line 1541: This is a library and we should not use `exit`. What about early return here?

ferdymercury: Yep, I tried that in the beginning but then it crashed later on at another time point. So I thought that for the time being it was better a clean exit than a straight crash.

vgvassilev: We should check in the caller for nullptr in that case.

pcanal: Isn't this in a constructor?  return early from a constructor will return an invalid object (not fully setup).  Do we have the proper checks already in place to verify validity?

pcanal: Follow up PR and comments: https://github.com/root-project/root/pull/16966/files#r1849138933

PR #16930, opened 2024-11-13
build: The builtin TBB sets a debug suffix for its library on linux too

pcanal:
When build in debug mode the builtin TBB renames its library with `_debug` prefix.


dpiparo:
What is not completely clear to me is why ubu22, ubu2410, ubu2404 as well as alma9 are built in with "-DCMAKE_BUILD_TYPE=Debug" but we do not see any issue: do you know?

PR #16931, opened 2024-11-13
[ntuple] Fix race in cluster pool

jblomer:
TODO: better description / issue


jblomer:
Still something strange on mac14. Investigating.

jblomer:
Doesn't seem to be a leak but just excessive memory consumption with the access pattern.

hahnjo (commented):
Can you maybe add some explanation to the last commit's message why `fIsExpired` is not needed (anymore)? Is it because with the removal of the background unzipping earlier this year, the loaded clusters are directly given back to the main thread?

jblomer:
> Can you maybe add some explanation to the last commit's message why `fIsExpired` is not needed (anymore)? Is it because with the removal of the background unzipping earlier this year, the loaded clusters are directly given back to the main thread?

Sure, and yes. Updated the commit message to

```
The fIsExpired flag was used to indicate to the I/O thread that a loaded
cluster is not needed anymore. This was useful at the time when the I/O
thread also took care of the parallel decompression. In this way, we could
avoid useless decompression of clusters.

However, now the parallel decompression is anyway triggered by the main
thread. The loop over fInFlightCluster in RClusterPool::GetCluster()
already avoids decompression if a cluster was expired.

Hence, the out-of-band signal from the main thread to the I/O thread is
not needed anymore.
```

hahnjo (approved):
LGTM

PR #16938, opened 2024-11-14
[RF][HS3] Cleanup and minor fixes

cburgard:
# This Pull request:

Contains several bugfixes related to HS3:

- Performs code cleanup on some sections of the HS3 HistFactory implementation
- Improves constraint handling for ShapeFactor-type modifiers
- Enables proper tracking of Global Observables in import and export of HS3 JSON files

## Checklist:

- [x] tested changes locally

This PR fixes #


NickKyriacou:
Apologies, but is there an update on when this will be merged to an official release of ROOT? This is needed for some HEP-data validation for some research looking to finish publishing quite soon.

guitargeek:
Thanks for the ping @NickKyriacou. Let's get some new CI results, but as far as I remember this PR still has test failures that need to be ironed out. I'm working on it.

roofit/hs3/src/RooJSONFactoryWSTool.cxx
```diff
@@ -718,7 +686,7 @@ void importAnalysis(const JSONNode &rootnode, const JSONNode &analysisNode, cons
    for (const auto &p : pars) {
       if (mc->GetParametersOfInterest()->find(*p))
          continue;
-      if (p->isConstant() && !mainPars.find(*p)) {
+      if (p->isConstant() && !mainPars.find(*p) && domainPars.find(*p)) {
```
guitargeek, line 689: @cburgard, @NickKyriacou, If I can tell correctly, this is a backwards incompatible change.

RooFitHS3 versions prior to this PR are not exporting the global observables to "domains", and this new condition will therefore not cover them. Old JSON files will therefore give you ModelConfigs with missing global observables.

Am I missing something? Have you considered that?

@NickKyriacou the rest of the PR looks good to me, only the last commit seems problematic for backwards compatibility. Which part of this PR do you need for your work?

guitargeek (changes requested):
Something has to be done to ensure backwards compatibility.

roofit/hs3/src/RooJSONFactoryWSTool.cxx
cburgard: It's a bit long ago that I made these changes, but I'll try to recall my rationale here.

Old JSON files indeed do not contain any reference to the global observables. In fact, they were handled completely implicitly. In HistFactory scenarios, which were the majority of applications anyway, the constraint terms are never explicitly imported, and since the physics distribution does not depend on global observables by definition, they were missing from the model completely and thus also not be contained in the ModelConfig already before. So this was, in fact, a bug that is fixed by this change.
There might be individual models out there where people have been exporting constraint terms explicitly, where this might lead to a "breaking change" that is not a bugfix, but I would be a bit surprised if this would have happened more than a handful of times in total. I would deem this an acceptable loss.
If you are really concerned by this nevertheless, we could configure it such that it first tries to compose the list of global observables in the new way (with the risk of it running empty), and if the list of global observables does indeed turn out empty in the end in the new way, the old way is reinstantiated. It's not super clean, but it would catch the cases you are concerned about. This does, however, open the new construction site of "what about models that genuinely don't have global observables, and that would get some variables that are not global observables assigned to them in this way".
In the end, RooFit does have some logic of identifying global observables by itself anyway, which is typically only accessed during "fitTo" though. So what I would really like is if we could use the code from RooFit that identifies global observables and just call that in case the globs list turns out empty from the model, but this might require some refactoring on your side @guitargeek since I don't know how easily that functionality can be accessed from this part of the code. One of the main criticisms of this automatic determination of globs I think is that it only works if you know the dataset, but in the way we have configured the "analysis" object, we actually do have a pointer to the dataset at hand, so conceptually, there is no issue in using this functionality.

guitargeek: Thanks for the extensive answer! Yes, let's try not to do any automatic determination of global observables if we can avoid it.

I was worried about the already-published models [listed on the HS3 GitHub](https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard), but if for HistFactory models there is no problem and the published models are all HistFactory, it's all good. I have tested some of the published likelihood with and without this PR. and there is not difference in Global observables.

For example, I tested [this HEP data entry](https://www.hepdata.net/record/ins2832100) with this code snippet:
```
import ROOT

ws = ROOT.RooWorkspace()

tool = ROOT.RooJSONFactoryWSTool(ws)

tool.importJSON("likelihood.json")

mc = ws.obj("ModelConfig")

mc.Print()

nll = mc.createNLL(ws.data("obsData")) # will implicitly use global observables from ModelConfig

val = nll.getVal()

print(val)
```

Another check I did was to re-run the HistFactory closure tests with this PR, but using the JSON files exported by ROOT master without this PR (so the version that doesn't write the global observables). It works indeed, the global observables are implicitly created.

cburgard: Ok great, thanks for the quick check. In the end, we were saved by the fact that we put off the "proper" treatment of the constraint terms for so long.
I assume it's all okay now, and if it turns out to be a problem after all, we can look at the specific case to understand why it fails.
Let me know if you still need any input from my side.

guitargeek (approved):
I checked that the slight backwards incompatibility doesn't affect current usecases (HistFactory models published on HEP data), so OK from me now!

I also did some other improvements an cleanups. In particular, the `HistFactory::JSONTool` is gone now. It was already removed from the public interface but was kept for test coverage. But every time the "JSON -> RooWorkspace" code is updated like in this PR, closure tests start to fail without a corresponding update of the tool for "Measurement -> JSON". So maintaining this "dead" code became a bit too expensive.

guitargeek:
> Apologies, but is there an update on when this will be merged to an official release of ROOT? This is needed for some HEP-data validation for some research looking to finish publishing quite soon.

Hi @NickKyriacou, I'll backport Carstens commits for this PR to the next patch release 6.34.04, which will be released very soon.

PR #16939, opened 2024-11-14
[ntuple] Some fixes to the merger

silverweed:
- fix wrong check when deciding whether to resize the `fBuffers` array
- reconstruct projections when creating the output model


vepadulano (approved):
Nice thanks!

PR #16944, opened 2024-11-14
[ntuple] Merger: fix merging RNTuples with projected fields and handling of the output file compression

silverweed:
Depends on https://github.com/root-project/root/pull/16939.

More fixes to the merger, enough to successfully merge two RNTuples converted from CMS Open Data:
- use physical ids, not logical ids, in the API that require them
- don't try to merge aliased columns, reconstruct the projections instead
- fix the way we handle the output file's compression settings in case of fast merging (hadd's `-ff` flag). This is an important fix because right now we can **write corrupted data** when `-ff` is used (e.g. we can merge non-split columns as-is but write their type as split in the descriptor. This is a recoverable corruption using custom user code, but it makes the RNTuple return garbage data through the regular API and it's not easily identifiable).


tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -108,7 +108,10 @@ try {
                inFile->GetName());
          return -1;
       }
-      sources.push_back(RPageSourceFile::CreateFromAnchor(*anchor));
+      auto readOpts = RNTupleReadOptions();
+      // disable the cluster cache so we can catch the exception that happens on LoadEntry
+      readOpts.SetClusterCache(RNTupleReadOptions::EClusterCache::kOff);
```
jblomer, line 103: Should this stay as is? Background reading may be beneficial for merging.

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -195,7 +198,8 @@ struct RColumnMergeInfo {
    // the columns' parent fields' names plus the index of the column itself.
    // e.g. "Muon.pt.x._0"
    std::string fColumnName;
-   DescriptorId_t fInputId;
+   DescriptorId_t fInputLogicalId;
+   DescriptorId_t fInputPhysicalId;
```
jblomer, line 210: I think we may not need at all the logical columns.

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -82,26 +82,16 @@ try {
...
-
    // The remaining entries are the input files
    std::vector<std::unique_ptr<RPageSourceFile>> sources;
    std::vector<RPageSource *> sourcePtrs;
 
+   int firstFileCompression = kUnknownCompressionSettings;
    while (const auto &pitr = itr()) {
       TFile *inFile = dynamic_cast<TFile *>(pitr);
+      if (firstFileCompression == kUnknownCompressionSettings)
+         firstFileCompression = inFile->GetCompressionSettings();
```
jblomer, line 93: Not sure about this one: since we already diverged the RNTuple compression from the TFile compression (default 505 RNTuple vs default 101 TFile), should we interpret this option rather as: "look at the first _RNTuple_ as a reference for the compression settings".

jblomer (commented):
Nice! Some smaller comments and I think we should have tests for the encountered issues before merging.

tree/ntuple/v7/src/RNTupleMerger.cxx
silverweed: Sorry, I forgot to remove that line (it was for debugging)

tree/ntuple/v7/src/RNTupleMerger.cxx
silverweed: But technically there is no such thing as an "RNTuple compression settings", right? Each column range may in principle have a different compression setting, so which one should we pick?

silverweed: That said, I agree that if I do `-ff` when merging RNTuples, I wouldn't expect the output RNTuple to be compressed with 101 if the input was using the default 505 compression...

pcanal: >  Each column range may in principle have a different compression setting, so which one should we pick?

As a user I would expect each column of the output `RNTuple` to have the same compression setting as they have as in the first input file. (i.e. the output structure/meta-data should be a 'clone' of the structure/meta-data  of the first input `RNTuple`)

silverweed: We might decide that's what we want eventually; I propose that initially we just assume that the first column has the same compression as every other column (since we currently don't expose an API to use different compression).

pcanal: > We might decide that's what we want eventually

Humm ... unless we have a strong reason not to, we ought to be similar to `TTree` which already does the per column copy of the compression setting (implicitly).

silverweed: > What do you mean?

Even though the format supports different compressions per column range, currently from our API you can only set the compression for the entire RNTuple (through `RNTupleWriteOptions`)

main/src/hadd.cxx
hahnjo, line 1: The commit message seems to have two title lines, is this intentional?

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -78,22 +78,25 @@ try {
...
+      // user passed -ff or -fk: use the same compression as the first RNTuple we find in the sources.
+      // (do nothing here, the compression will be fetched below)
+   } else if (!defaultComp) {
+      // compression was explicitly passed by the user: use it.
+      compression = outFile->GetCompressionSettings();
+   } else {
+      // user passed no compression-related options: use default
+      compression = RCompressionSetting::EDefaults::kUseGeneralPurpose;
+      Info("RNTuple::Merge", "Using the default compression: %d", compression);
    }
```
hahnjo, line 100: Can you remind me what is the default if I just do `hadd out.root in1.root in2.root`? From a users perspective, I would not expect this to change the compression / recompress, but the code seems to suggest that I have to pass `-ff` or `-fk` to get "fast" merging?

tree/ntuple/v7/src/RNTupleMerger.cxx
hahnjo: Hm, I'm confused: Didn't we recently change this that alias columns (ie where logical column ids matter) are always higher than physical columns, which is what the merger probably should work on?

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -803,7 +803,10 @@ static void AddColumnsFromField(std::vector<RColumnMergeInfo> &columns, const RN
       const auto &srcColumn = srcDesc.GetColumnDescriptor(srcColumnId);
       RColumnMergeInfo info{};
       info.fColumnName = name + '.' + std::to_string(srcColumn.GetIndex());
-      info.fInputId = srcColumnId;
+      std::cerr << "Adding column " << info.fColumnName << "with id " << srcColumnId
+                << " (phys: " << srcColumn.GetPhysicalId() << ")\n";
```
hahnjo: Is this debug output?

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -799,12 +799,13 @@ static void AddColumnsFromField(std::vector<RColumnMergeInfo> &columns, const RN
    // NOTE: here we can match the src and dst columns by column index because we forbid merging fields with
    // different column representations.
    for (auto i = 0u; i < srcFieldDesc.GetLogicalColumnIds().size(); ++i) {
+      if (srcFieldDesc.IsProjectedField())
+         continue;
+
```
hahnjo, line 804: The commit message says to "add some extra verbose messages", but this is clearly a functional change (and should probably have a unit test, not just an integration test?)

hahnjo (changes requested):

tree/ntuple/v7/src/RNTupleMerger.cxx
silverweed: Yes, I split the changes in multiple commits after the fact and this must have slipped into the wrong one. I can put it into the proper commit and add a unit test for it

tree/ntuple/v7/src/RNTupleMerger.cxx
silverweed: Correct. I thought I removed it but apparently not from all git timelines. Thanks for spotting it

tree/ntuple/v7/src/RNTupleMerger.cxx
silverweed: Yes, the merger should only work on non-alias columns, so I will remove the distinction and just call it `fInputId`

silverweed:
I rebased this PR to not depend on https://github.com/root-project/root/pull/16949
Sorry for the confusion

tree/ntuple/v7/src/RNTupleMerger.cxx
pcanal: Does that prevent just the user from customizing the compression settings or does it also prevent the proper implementation inside the merger? [related issue: we are likely to 'forget' to update the Merger when/if we expand the API to allow the per column customization]

jblomer (approved):

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -476,13 +476,14 @@ static void ExtendDestinationModel(std::span<const RFieldDescriptor *> newFields
 // Merges all columns appearing both in the source and destination RNTuples, just copying them if their
 // compression matches ("fast merge") or by unsealing and resealing them with the proper compression.
 void RNTupleMerger::MergeCommonColumns(RClusterPool &clusterPool, DescriptorId_t clusterId,
-                                       std::span<RColumnMergeInfo> commonColumns, RCluster::ColumnSet_t commonColumnSet,
+                                       std::span<RColumnMergeInfo> commonColumns,
+                                       const RCluster::ColumnSet_t &commonColumnSet,
                                        RSealedPageMergeData &sealedPageData, const RNTupleMergeData &mergeData)
 {
-   assert(commonColumns.size() == commonColumnSet.size());
```
hahnjo, line 482: Is this `assert` intentionally dropped? In general, the diff of this commit doesn't seem to entirely match the message...

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -491,27 +492,24 @@ void RNTupleMerger::MergeCommonColumns(RClusterPool &clusterPool, DescriptorId_t
...
       RPageStorage::SealedPageSequence_t sealedPages;
       sealedPages.resize(pages.fPageInfos.size());
 
       // Each column range potentially has a distinct compression settings
-      const auto colRangeCompressionSettings = clusterDesc.GetColumnRange(columnId).fCompressionSettings;
-      const bool needsCompressionChange = mergeData.fMergeOpts.fCompressionSettings != kUnknownCompressionSettings &&
-                                          colRangeCompressionSettings != mergeData.fMergeOpts.fCompressionSettings;
-
+      const auto colRangeCompressionSettings = clusterDesc.GetColumnRange(column.fInputId).fCompressionSettings;
+      const bool needsCompressionChange = colRangeCompressionSettings != mergeData.fMergeOpts.fCompressionSettings;
```
hahnjo, line 512: For `needsCompressionChange`, you drop the condition `mergeData.fMergeOpts.fCompressionSettings != kUnknownCompressionSettings` - is this intentional? Again, not related to the column id changes

hahnjo (approved):

PR #16949, opened 2024-11-14
[ntuple] Merger: fix handling of compression-related options

silverweed:
Currently the RNTupleMerger has a pretty confusing and unintuitive handling of `hadd`'s compression options.
To simplify the situation, this PR implements these simpler rules:
1. the output RNTuple and its container file always have the same compression
2.  if the user explicitly passes a compression settings (`-f[0-9]`), use that compression
3. if the user doesn't pass any compression flag, use 505 as the output compression
4. if the user passes `-ff` or `-fk`, open the first source file, grab the RNTuple inside it, peek the first column range we can find and use its compression settings as the output compression. This is different from the current behavior of using the same compression as the first source *file*, as that behavior is probably not what the user wants.

This requires passing more information from hadd to the RNTupleMerger. I added a couple of TString merge options to do so, which won't impact the existing merging code as they only get interpreted by the RNTupleMerger.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/main/src/hadd.cxx
+++ b/main/src/hadd.cxx
@@ -427,8 +427,10 @@ int main( int argc, char **argv )
          else
             newcomp = ROOT::RCompressionSetting::EDefaults::kUseCompiledDefault;
          delete firstInput;
+         fileMerger.SetMergeOptions(TString("first_source_compression"));
       } else {
          newcomp = ROOT::RCompressionSetting::EDefaults::kUseCompiledDefault;
+         fileMerger.SetMergeOptions(TString("default_compression"));
       }
    }
    if (verbosity > 1) {
@@ -487,7 +489,7 @@ int main( int argc, char **argv )
          }
       }
       merger.SetNotrees(noTrees);
-      merger.SetMergeOptions(cacheSize);
+      merger.SetMergeOptions(TString(merger.GetMergeOptions()) + " " + cacheSize);
       merger.SetIOFeatures(features);
       Bool_t status;
       if (append)
--- a/tree/ntuple/v7/src/RNTupleMerger.cxx
+++ b/tree/ntuple/v7/src/RNTupleMerger.cxx
@@ -78,22 +78,25 @@ try {
       // pointer we just got.
    }
 
-   // The "fast" option is present if and only if we don't want to change compression.
-   const int compression =
-      mergeInfo->fOptions.Contains("fast") ? kUnknownCompressionSettings : outFile->GetCompressionSettings();
-
-   RNTupleWriteOptions writeOpts;
-   writeOpts.SetUseBufferedWrite(false);
-   if (compression != kUnknownCompressionSettings)
-      writeOpts.SetCompression(compression);
-   auto destination = std::make_unique<RPageSinkFile>(ntupleName, *outFile, writeOpts);
-
-   // If we already have an existing RNTuple, copy over its descriptor to support incremental merging
-   if (outNTuple) {
-      auto source = RPageSourceFile::CreateFromAnchor(*outNTuple);
-      source->Attach();
-      auto desc = source->GetSharedDescriptorGuard();
-      destination->InitFromDescriptor(desc.GetRef());
+   const bool defaultComp = mergeInfo->fOptions.Contains("default_compression");
+   const bool firstSrcComp = mergeInfo->fOptions.Contains("first_source_compression");
+   if (defaultComp && firstSrcComp) {
+      // this should never happen through hadd, but a user may call RNTuple::Merge() from custom code...
+      Warning(
+         "RNTuple::Merge",
+         "Passed both options \"default_compression\" and \"first_source_compression\": only the latter will apply.");
+   }
+   int compression = kUnknownCompressionSettings;
+   if (firstSrcComp) {
+      // user passed -ff or -fk: use the same compression as the first RNTuple we find in the sources.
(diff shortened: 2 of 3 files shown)


tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -78,22 +78,25 @@ try {
       // pointer we just got.
    }
 
-   // The "fast" option is present if and only if we don't want to change compression.
-   const int compression =
-      mergeInfo->fOptions.Contains("fast") ? kUnknownCompressionSettings : outFile->GetCompressionSettings();
-
-   RNTupleWriteOptions writeOpts;
-   writeOpts.SetUseBufferedWrite(false);
```
jblomer, line 86: This option got lost, didn't it?

jblomer (commented):
In principle looks good but we should add tests.

tree/ntuple/v7/src/RNTupleMerger.cxx
silverweed: Yes, a victim of the rebase 😅

silverweed:
@jblomer added a couple of tests for the new rules: https://github.com/root-project/roottest/pull/1220

main/src/hadd.cxx
hahnjo, line 1: The commit message seems to have two title lines, is this intentional?

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -78,22 +78,25 @@ try {
...
+   if (firstSrcComp) {
+      // user passed -ff or -fk: use the same compression as the first RNTuple we find in the sources.
+      // (do nothing here, the compression will be fetched below)
+   } else if (!defaultComp) {
+      // compression was explicitly passed by the user: use it.
+      compression = outFile->GetCompressionSettings();
+   } else {
+      // user passed no compression-related options: use default
+      compression = RCompressionSetting::EDefaults::kUseGeneralPurpose;
+      Info("RNTuple::Merge", "Using the default compression: %d", compression);
```
hahnjo, line 99: (cross-posting from https://github.com/root-project/root/pull/16944#discussion_r1843373630)
Can you remind me what is the default if I just do `hadd out.root in1.root in2.root`? From a users perspective, I would not expect this to change the compression / recompress, but the code seems to suggest that I have to pass `-ff` or `-fk` to get "fast" merging?

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -144,8 +144,12 @@ try {
       sources.push_back(std::move(source));
    }
 
+   // Always write the RNTuple and the file with the same compression.
+   outFile->SetCompressionSettings(compression);
```
hahnjo, line 148: Should we "always" do this, or only when there is exactly one RNTuple? What about a file that has one RNTuple (505) and one histogram (101)?

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -144,8 +144,12 @@ try {
...
+
    RNTupleWriteOptions writeOpts;
    assert(compression != kUnknownCompressionSettings);
+   writeOpts.SetUseBufferedWrite(false);
```
hahnjo, line 152: This should go to the previous commit... On the other hand it probably doesn't do anything since we are not using `RPageSinkBuf` in the first place, but construct a `RPageSinkFile` manually, so might as well just drop it (in a separate commit)

main/src/hadd.cxx
silverweed: It's a squash of two commits; I could remove one of the lines as it's not really that relevant

tree/ntuple/v7/src/RNTupleMerger.cxx
jblomer: I think I agree. At this point, we simply diverged and the default TFile compression is different from the default RNTuple compression. We can discuss if/how we want to address it but I don't think we need extra code in the merger. We have the same situation (different compression algorithms) if you write a new RNTuple.

jblomer (approved):
Thanks!
Some minor things and some open questions from @hahnjo 
Otherwise LGTM.

main/src/hadd.cxx
```diff
@@ -427,8 +427,10 @@ int main( int argc, char **argv )
          else
             newcomp = ROOT::RCompressionSetting::EDefaults::kUseCompiledDefault;
          delete firstInput;
+         fileMerger.SetMergeOptions(TString("first_source_compression"));
```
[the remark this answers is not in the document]
silverweed: Unfortunately if I don't explicitly construct it, the compiler gives a "call to member function is ambiguous" and doesn't compile

silverweed:
I added a battery of tests with all the relevant combinations of compression for input and output to verify that we can always read back the data. This amounts to 250 tests which run in about 10 seconds in debug mode. If you think that's too much I can reduce the tested combinations, but I think it doesn't hurt to overtest a bit if the cost in time is so modest.

main/src/hadd.cxx
hahnjo, line 1: The commit message is not what I think it should be...

hahnjo (commented):
LG, but the first commit message doesn't match the change...

main/src/hadd.cxx
silverweed: Sorry, I fixed it

PR #16950, opened 2024-11-14
[PyROOT] Retain C++ ownership for TColor

vepadulano:
Prevent crashes due to TColor memory ownership


guitargeek (approved):
Thanks!

dpiparo (approved):
The failure is unrelated. As far as I am concerned, the fix could be merged and backported to 6.34.

PR #16952, opened 2024-11-14
[RF] Correct broadcasting in PiecewiseInterpolation::doEval

guitargeek:
Before, we assumed that the "nominal" input always has the output size, but that's not generally true. Any of the inputs, including "nominal", might be a scalar that has to be broadcasted.


lmoneta (approved):
LGTM! 
Thank you Jonas for the fix

PR #16954, opened 2024-11-14
[RF] Change return type of `RooRealVar::format()` to std::string

guitargeek:
Returning a TString by raw pointer that the caller needs to delete is not memory safe and an outdated pattern, which merits a breaking change documented in the 6.36 release notes.


lmoneta (approved):
LGTM!
I agree is better to have this change!

PR #16956, opened 2024-11-15
[ntuple] Simplify `RNTuple::Streamer()`

hahnjo:
This makes the writing and reading paths symmetric.

### DIFF
--- a/tree/ntuple/v7/src/RNTuple.cxx
+++ b/tree/ntuple/v7/src/RNTuple.cxx
@@ -27,23 +27,12 @@
 void ROOT::RNTuple::Streamer(TBuffer &buf)
 {
    if (buf.IsReading()) {
-      UInt_t offClassBuf;
-      UInt_t bcnt;
-      auto classVersion = buf.ReadVersion(&offClassBuf, &bcnt);
-
-      // Strip class version from checksum calculation
-      UInt_t lenStrip = sizeof(Version_t);
-
-      if (bcnt < lenStrip)
-         throw Experimental::RException(R__FAIL("invalid anchor byte count: " + std::to_string(bcnt)));
-
-      auto lenCkData = bcnt - lenStrip;
       // Skip byte count and class version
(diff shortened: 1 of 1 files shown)


tree/ntuple/v7/src/RNTuple.cxx
```diff
@@ -27,23 +27,12 @@
...
-      if (bcnt < lenStrip)
-         throw Experimental::RException(R__FAIL("invalid anchor byte count: " + std::to_string(bcnt)));
-
-      auto lenCkData = bcnt - lenStrip;
       // Skip byte count and class version
-      auto offCkData = offClassBuf + sizeof(UInt_t) + sizeof(Version_t);
-      auto expectedChecksum = XXH3_64bits(buf.Buffer() + offCkData, lenCkData);
+      auto offCkData = buf.Length() + sizeof(UInt_t) + sizeof(Version_t);
+      buf.ReadClassBuffer(RNTuple::Class(), this);
+      std::uint64_t expectedChecksum = XXH3_64bits(buf.Buffer() + offCkData, buf.Length() - offCkData);
```
jblomer, line 33: So, we are assuming that (essentially) `bcnt` is the same as `buf.Length()`. Is this always true?

hahnjo: Yes, somewhat: From my understanding, `buf.Length()` is essentially the current offset in `buf`. `buf.Length() - offCkData` with `offCkData = buf.Length() + sizeof(UInt_t) + sizeof(Version_t)` is therefore the number of processed bytes after the version (which is according to `bcnt`).

For comparison with the old version, note that `lenStrip = sizeof(Version_t)`, which was the "relative offset" of the class buffer, while `offCkData` is the absolute offset in `buf`. See also the writing code path for the symmetry mentioned in the commit message.

pcanal (approved):

jblomer (approved):

PR #16957, opened 2024-11-15
[cmake] use the vendor-provided pkgconfig for libuuid

silverweed:
Our own FindLibuuid.cmake is flawed in several ways:
- it doesn't set the variables it advertises to set
- it doesn't always find libuuid in the system
- it improperly sets libuuid_FOUND if it only finds the include dir but not the libraries, which can lead to linking errors


bellenot:
And if it's only needed for UUID, should it be required globally? (Not sure if UUID is supported on Windows, but Davix for sure is not)

bellenot:
BTW, there is apparently a [FindLibUUID.cmake](https://github.com/Kitware/CMake/blob/9d439e235ce117fb24822e0cb1716c785e38db19/Source/Modules/FindLibUUID.cmake) in CMake

bellenot:
> Do we know which version of CMake it first appeared in?

2017 AFAICS, but let see this next week

CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ foreach(policy ${policy_old})
 endforeach()
 
 include(cmake/modules/CaptureCommandLine.cmake)
+find_package(PkgConfig REQUIRED)
```
bellenot, line 37:
```suggestion
if(NOT MSVC)
  find_package(PkgConfig REQUIRED)
endif()
```

bellenot (changes requested):

Issue #16946, opened 2024-11-14
Crash in RDF constructor with empty file list


TomasDado:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

When RDF is created with an ampty string and an emtpy file list, the code crashes with a segmetation fault.

This is probably a very weird piece of code, but I think there should be a proper error message and not a seg fault

### Reproducer

```c++
void tmp() {
  ROOT::RDataFrame df("", {});
}
```

### ROOT version

```
ROOT Version: 6.33.01
Built for linuxx8664gcc on Oct 11 2024, 11:11:55
From heads/master@9f8c9ca672
```

### Installation method

build from source

### Operating system

Ubuntu 24.10

TomasDado:
Just for completeness, this is the crash
```
tomas@tomas-IdeaPad-5-Pro-14IAP7:~$ root -l tmp.cxx 
root [0] 
Processing tmp.cxx...

 *** Break *** segmentation violation



===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================
#0  0x000071f492d1a127 in __GI___wait4 (pid=29072, stat_loc=stat_loc
entry=0x7ffe3a28bc18, options=options
entry=0, usage=usage
entry=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
#1  0x000071f492d1a26b in __GI___waitpid (pid=<optimized out>, stat_loc=stat_loc
entry=0x7ffe3a28bc18, options=options
entry=0) at ./posix/waitpid.c:38
#2  0x000071f492c5ad9b in do_system (line=<optimized out>) at ../sysdeps/posix/system.c:172
#3  0x000071f49372b769 in TUnixSystem::StackTrace() () from /home/tomas/ROOT/root_install/lib/libCore.so
#4  0x000071f49372b074 in TUnixSystem::DispatchSignals(ESignals) () from /home/tomas/ROOT/root_install/lib/libCore.so
#5  <signal handler called>
[...]

PR #16961, opened 2024-11-15
[df] Check for empty file list in constructor

vepadulano:
Fixes https://github.com/root-project/root/issues/16946


dpiparo (approved):
LGTM. Thanks for the fix. Does it perhaps qualify for a backport to 6.34?

PR #16969, opened 2024-11-18
[core] Fold AutoloadLibraryMU into AutoloadLibraryGenerator

hahnjo:
Use `AbsoluteSymbolsMaterializationUnit` to define the symbols.

### DIFF
--- a/core/metacling/src/TClingCallbacks.cxx
+++ b/core/metacling/src/TClingCallbacks.cxx
@@ -84,78 +84,6 @@ extern "C" {
    void TCling__UnlockCompilationDuringUserCodeExecution(void *state);
 }
 
-class AutoloadLibraryMU : public llvm::orc::MaterializationUnit {
-   std::string fLibrary;
-   llvm::orc::SymbolNameVector fSymbols;
-public:
-   AutoloadLibraryMU(const std::string &Library, const llvm::orc::SymbolNameVector &Symbols)
-      : MaterializationUnit({getSymbolFlagsMap(Symbols), nullptr}), fLibrary(Library), fSymbols(Symbols)
-   {
-   }
-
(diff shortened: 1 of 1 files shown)


hahnjo:
As far as I could understand, Lang recommends this simplification, so marking as ready for review

core/metacling/src/TClingCallbacks.cxx
```diff
@@ -200,10 +128,27 @@ class AutoloadLibraryGenerator : public llvm::orc::DefinitionGenerator {
...
       }
 
-      for (auto &&KV : found) {
-         auto MU = std::make_unique<AutoloadLibraryMU>(KV.first, std::move(KV.second));
-         if (auto Err = JD.define(MU))
-            return Err;
+      llvm::orc::SymbolMap loadedSymbols;
+      for (const auto &KV : found) {
+         // Try to load the library which should provide the symbol definition.
+         // TODO: Should this interface with the DynamicLibraryManager directly?
```
vgvassilev, line 134: Yes, ideally that'd be a good idea...

hahnjo:
> Do we have any performance numbers? Is the absolute symbol approach similar in terms of computational cost?

I'm not sure I understand, this PR is just moving the code from the MU to the Generator. `AbsoluteSymbolsMaterializationUnit` is literally just doing a `notifyResolved` of the symbols with their absolute pointers.

vgvassilev:
> > Do we have any performance numbers? Is the absolute symbol approach similar in terms of computational cost?
> 
> I'm not sure I understand, this PR is just moving the code from the MU to the Generator. `AbsoluteSymbolsMaterializationUnit` is literally just doing a `notifyResolved` of the symbols with their absolute pointers.

I think it answers my question.

vgvassilev (approved):
Lgtm!

Issue #16687, opened 2024-10-15
Loss of floating point precision when saving TCanvas as ROOT macro


guitargeek:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Floating point precision should be preserved when saving plots as ROOT macros with `TCanvas::SaveAs()`. This can probably be achieved with not too much effort using `std::setprecision`:
https://en.cppreference.com/w/cpp/io/manip/setprecision

### Reproducer

Run this code, which works fine:

```c++
TCanvas *c = new TCanvas("c", "c");
TH1D *histo = new TH1D("histo", "histo", 1, 595780500.110732, 595780531.878908);
histo->Draw();
c->SaveAs("plot.C");
```

The resulting macro will not preserve the precision on the floating point numbers:
```c++
   TH1D *histo__1 = new TH1D("histo__1","histo",1,5.957805e+08,5.957805e+08);
```
And you'll get an error when running it:
```txt
Processing plot.C...
Error in <TCanvas::Range>: illegal world coordinates range: x1=595780500.000000, y1=-0.131250, x2=595780500.000000, y2=1.181250
Error in <TCanvas::RangeAxis>: illegal axis coordinates range: xmin=595780500.000000, ymin=0.000000, xmax=595780500.000000, ymax=1.050000
root [1] Error in <TCanvas::Range>: illegal world coordinates range: x1=595780500.000000, y1=-0.131250, x2=595780500.000000, y2=1.181250
Error in <TCanvas::RangeAxis>: illegal axis coordinates range: xmin=595780500.000000, ymin=0.000000, xmax=595780500.000000, ymax=1.050000
```

### ROOT version

```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.32.06                        https://root.cern |
  | (c) 1995-2024, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Sep 21 2024, 18:21:53                 |
  | From tags/6-32-06@6-32-06                                        |
  | With g++ (GCC) 13.3.0                                            |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------
```

### Installation method

Nix

### Operating system

NixOS (Linux)

### Additional context

Can easily happen in the analysis of timeseries, where one often has large axis values because they represent time with respect to some reference point far back in time (e.g. astrophysics, finance, ...).

couet:
[...]

PR #16970, opened 2024-11-18
[hist] increase precision of TH1's and TGraph's SavePrimitive to double

ferdymercury:
# This Pull request:

## Changes or fixes:
Fixes https://github.com/root-project/root/issues/16687

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/hist/hist/src/TGraph.cxx
+++ b/hist/hist/src/TGraph.cxx
@@ -40,6 +40,8 @@
 #include <fstream>
 #include <cstring>
 #include <numeric>
+#include <limits>
+#include <iomanip>
 
 #include "HFitInterface.h"
 #include "Fit/DataRange.h"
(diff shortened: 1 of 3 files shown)


dpiparo:
Thanks! Does it make sense to restore the previous precision once done, i.e. not altering the state of `std::cout`?

ferdymercury:
> Thanks! Does it make sense to restore the previous precision once done, i.e. not altering the state of `std::cout`?

Hmmm, Thats already the case with the sstream, see last line in the code. Or you mean sth else?

dpiparo:
I really meant that and missed that line. Thanks for the clarification!

dpiparo:
The two failures on Windows are real. The test that fails is `gtest-hist-hist-testMapCppName`.

What I believe it should be done is to increase the tolerance from 200 to 211 here: https://github.com/root-project/root/blob/master/hist/hist/test/test_MapCppName.cxx#L48

pcanal:
> What I believe it should be done is to increase the tolerance from 200 to 211 here: [master/hist/hist/test/test_MapCppName.cxx#L48](https://github.com/root-project/root/blob/master/hist/hist/test/test_MapCppName.cxx?rgh-link-date=2024-11-18T14%3A02%3A38Z#L48)

Isn't it the base that changed rather than the tolerance?

ferdymercury:
Thanks, fixed

PR #16971, opened 2024-11-18
[cmake] replace custom FindLibuuid.cmake with upstream version

silverweed:
Taken from https://github.com/Kitware/CMake/blob/master/Source/Modules/FindLibUUID.cmake

Alternative for https://github.com/root-project/root/pull/16957. Needs to be validated on Windows and Mac, but it seems to work fine on Linux where the current FindLibuuid.cmake doesn't


cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -1122,7 +1122,7 @@ if(davix AND NOT builtin_davix)
     find_package(Davix 0.6.4)
     if(NOT DAVIX_FOUND)
       find_package(LibUUID)
-      if(NOT LibUUID_FOUND)
+      if(NOT TARGET LibUUID::LibUUID)
         message(STATUS "Davix dependency libuuid not found, switching OFF 'davix' option.")
```
hageboeck, line 1126: I'm wondering if we should error out here. Somebody seemed to have wanted davix, and it's surprising to not have enabled.

hageboeck (commented):
LGTM, with maybe one `message(STATUS` replace by an error. I don't see why ROOT should continue to configure when the requested configuration isn't achievable.

And we of course need to wait for the merge of the other PR ...

cmake/modules/SearchInstalledSoftware.cmake
pcanal: probably when "fail on missing" in ON but then it should be at the level of DAVIX ... (and that failure seems to be not implemented ...)

PR #16986, opened 2024-11-19
[ntuple] Several page pool improvements

jblomer:
- Improves the search complexity for pages in a given column from linear to logarithmic
- Prevents the same page from being added multiple times to the page pool
- Evicts unused pages from clusters before the currently active one

### DIFF
--- a/tree/ntuple/v7/src/RPagePool.cxx
+++ b/tree/ntuple/v7/src/RPagePool.cxx
@@ -106,11 +106,10 @@ void ROOT::Experimental::Internal::RPagePool::ReleasePage(const RPage &page)
 void ROOT::Experimental::Internal::RPagePool::RemoveFromUnusedPages(const RPage &page)
 {
    auto itr = fUnusedPages.find(page.GetClusterInfo().GetId());
-   if (itr != fUnusedPages.end()) {
-      itr->second.erase(page.GetBuffer());
-      if (itr->second.empty())
-         fUnusedPages.erase(itr);
-   }
+   assert(itr != fUnusedPages.end());
+   itr->second.erase(page.GetBuffer());
+   if (itr->second.empty())
+      fUnusedPages.erase(itr);
 }
 
 ROOT::Experimental::Internal::RPageRef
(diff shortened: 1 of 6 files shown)


tree/ntuple/v7/src/RPageStorage.cxx
```diff
@@ -330,14 +332,31 @@ void ROOT::Experimental::Internal::RPageSource::PrepareLoadCluster(
...
+{
+   if (fLastUsedCluster == clusterId)
+      return;
+
+   NTupleSize_t firstEntryIndex = GetSharedDescriptorGuard()->GetClusterDescriptor(clusterId).GetFirstEntryIndex();
+   auto itr = fPreloadedClusters.begin();
+   while ((itr != fPreloadedClusters.end()) && (itr->first < firstEntryIndex)) {
+      fPagePool.Evict(itr->second);
+      itr = fPreloadedClusters.erase(itr);
+   }
```
hahnjo, line 345: This evicts clusters with lower entry numbers, which is sufficient for sequential access and jumping ahead. But for truly random access, shouldn't the code also evict clusters past the cluster bunch size?

jblomer: Possibly, I'm not sure if we need extra code for this. The random access pattern is arguably a misuse, or at least users should turn of the read-ahead window in this case.

tree/ntuple/v7/src/RPagePool.cxx
```diff
@@ -68,30 +96,20 @@ void ROOT::Experimental::Internal::RPagePool::ReleasePage(const RPage &page)
...
+}
 
-      fEntries.resize(N - 1);
+void ROOT::Experimental::Internal::RPagePool::RemoveFromUnusedPages(const RPage &page)
+{
+   auto itr = fUnusedPages.find(page.GetClusterInfo().GetId());
+   if (itr != fUnusedPages.end()) {
+      itr->second.erase(page.GetBuffer());
+      if (itr->second.empty())
+         fUnusedPages.erase(itr);
```
hahnjo, line 112: Shouldn't we always find the unused page since we add all preloaded pages?

tree/ntuple/v7/src/RPageStorage.cxx
hahnjo: It wasn't clear to me if you wanted to implement it in this PR... I leave it up to you

hahnjo (approved):

tree/ntuple/v7/src/RPagePool.cxx
jblomer: We only do this in IMT mode.

PR #16987, opened 2024-11-19
[RF] Implement an initial version of comment annotations for codegen.

vgvassilev:
The produced code from the RooFit codegen backend is hard to follow. We add clarification comments with extra information from the RooFit grapn which can help the reader to correlate for which node the code was produced. That helps in understanding complex fits and should aid the adoption and debugging of AD in CMS Combine.

The implementation reworks the lexical scopes handling by implementing a stack to enable (mostly) correct indentations for both the generated code and the generated comments.


guitargeek (approved):
Thank you very much for these developments, LGTM!

PR #16988, opened 2024-11-20
[RF] MultiProcess additions to release notes

egpbos:
Hope this is still in time for the release.


egpbos:
@dpiparo done!

guitargeek (approved):
Thank you very much for adding to the release notes! I just removed the `[v634]` tag in the title, because we use this denote backports (which still needs to be done to the `[v634]` branch).

PR #16991, opened 2024-11-20
[RF] Add analytic integral for RooRealSumPdf (Codegen backend)

runtingt:
Reusing existing translate method since integration is linear


guitargeek (changes requested):
Thank you very much for this PR! It would be good to refactor it a bit to avoid code duplication, like we discussed offline.

guitargeek (approved):
Great! Thank you very much for this first contribution and for addressing the review comments! I'll take care of backporting this to 6.32 as well.

runtingt:
> Great! Thank you very much for this first contribution and for addressing the review comments! I'll take care of backporting this to 6.32 as well.

And to you for all your help, thanks again!

PR #16992, opened 2024-11-20
[tree] valid branch names or filenames does not ensure valid C++ variable names

ferdymercury:
# This Pull request:

## Changes or fixes:

So convert wherever need in produced selector code

Fixes https://its.cern.ch/jira/browse/ROOT-8775

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/treeplayer/src/TTreePlayer.cxx
+++ b/tree/treeplayer/src/TTreePlayer.cxx
@@ -733,18 +733,20 @@ Int_t TTreePlayer::MakeClass(const char *classname, const char *option)
    opt.ToLower();
 
    // Connect output files
-   if (!classname) classname = fTree->GetName();
-   const TString vclassname = ROOT::Internal::GetCppName(classname);
+   const TString fileNameStem = classname ? classname : fTree->GetName();
+   const TString cppClassName = ROOT::Internal::GetCppName(fileNameStem);
+   if (cppClassName != fileNameStem)
+      Warning("TTreePlayer::MakeClass", "The %s name provided ('%s') is not a valid C++ identifier and will be converted to '%s', the code produced will likely fail to compile.",(classname ? "class" : "tree"), fileNameStem.Data(), cppClassName.Data());
 
    TString thead;
-   thead.Form("%s.h", classname);
+   thead.Form("%s.h", fileNameStem.Data());
    FILE *fp = fopen(thead, "w");
    if (!fp) {
       Error("MakeClass","cannot open output file %s", thead.Data());
       return 3;
    }
    TString tcimp;
-   tcimp.Form("%s.C", classname);
+   tcimp.Form("%s.C", fileNameStem.Data());
    FILE *fpc = fopen(tcimp, "w");
    if (!fpc) {
       Error("MakeClass","cannot open output file %s", tcimp.Data());
@@ -781,8 +783,8 @@ Int_t TTreePlayer::MakeClass(const char *classname, const char *option)
    }
    fprintf(fp,"//////////////////////////////////////////////////////////\n");
    fprintf(fp,"\n");
-   fprintf(fp,"#ifndef %s_h\n",vclassname.Data());
-   fprintf(fp,"#define %s_h\n",vclassname.Data());
+   fprintf(fp,"#ifndef %s_h\n",cppClassName.Data());
+   fprintf(fp,"#define %s_h\n",cppClassName.Data());
    fprintf(fp,"\n");
    fprintf(fp,"#include <TROOT.h>\n");
    fprintf(fp,"#include <TChain.h>\n");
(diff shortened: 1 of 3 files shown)


dpiparo:
Hi @ferdymercury , this work is really good: thanks. I was wondering if we have a chance to re-use this function https://github.com/root-project/root/blob/master/core/clingutils/src/TClingUtils.cxx#L3269  via `TInterpreter::MapCppName`.
Example usage:
```
root [1] gInterpreter->MapCppName("my-strange<Branch*&")
(const char *) "mymIstrangelEBranchmUaN"
```
Clearly, the hard work here was to find what had to be done, the suggestion above is merely technical and meant to avoid duplication of code.

tree/treeplayer/src/TTreePlayer.cxx
```diff
@@ -734,6 +734,7 @@ Int_t TTreePlayer::MakeClass(const char *classname, const char *option)
 
    // Connect output files
    if (!classname) classname = fTree->GetName();
+   const TString vclassname = ROOT::Internal::GetCppName(classname).Data();
```
silverweed, line 737: You can omit the `.Data()` here

ferdymercury, line 737:
```suggestion
   const TString vclassname = ROOT::Internal::GetCppName(classname);
```

tree/treeplayer/src/TTreePlayer.cxx
```diff
@@ -734,6 +734,7 @@ Int_t TTreePlayer::MakeClass(const char *classname, const char *option)
 
    // Connect output files
    if (!classname) classname = fTree->GetName();
+   const TString vclassname = ROOT::Internal::GetCppName(classname);
```
pcanal, line 737:
```suggestion
   if (!classname) classname = ROOT::Internal::GetCppName(fTree->GetName());
```

tree/treeplayer/src/TTreeReaderGenerator.cxx
```diff
@@ -828,8 +838,8 @@ R"CODE(//////////////////////////////////////////////////////////
       ofs <<
 R"CODE(//////////////////////////////////////////////////////////
 
-#ifndef )CODE" << fClassname << R"CODE(_h
-#define )CODE" << fClassname << R"CODE(_h
+#ifndef )CODE" << ROOT::Internal::GetCppName(fClassname) << R"CODE(_h
```
pcanal, line 841: Should `fClassname` be updated instead?

tree/treeplayer/src/TTreePlayer.cxx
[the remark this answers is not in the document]
ferdymercury: I thought about that, but I opted not to change the filename because:

Imagine that I chose a treename called: "-myTree-". As user, I have no idea what the interpreter does behind the scenes to convert it. If I later call ls -myTree-* I will find the generated files. I don't care much if the internal code has changed the variable names, but I would like to find my files without knowing what conversion was applied. If the filename is changed to _myTree_.C I will not be able to find it (at least not so easily).

tree/treeplayer/src/TTreeReaderGenerator.cxx
ferdymercury: See comment above about user readability / findability of generated files.

tree/treeplayer/src/TTreePlayer.cxx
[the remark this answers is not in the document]
pcanal: This is a good point.  Another point of view in that in this example the file "-myTree-.C" now declares a class with a different name (`mImyTreemI`) but I guess this is 'easier' to find out.

Nonetheless we should clarify the purpose of the variable (i.e. `classname` is no longer the class name but something else). i.e. have something both `filestem` and `classname` set to their actual values.

With the current state of the PR we also have the 'issue' that if the user provide the `classname` but it is invalid, we produce invalid code.  So maybe we should use:
```
const TString filestem = classname;
if (classname) {
   auto validname( ROOT::Internal::GetCppName(classname));
   if (validname != classname)
      Error(functionname, "The class name provided ('%s') is no a valid C++ identifier, the code produced will likely fail to compile.", classname);
} else {
   filestem = fTree->GetName();
   classname = ROOT::Internal::GetCppName(fTree->GetName());
}
```

tree/treeplayer/src/TTreeReaderGenerator.cxx
pcanal: Here too, we should distinguish the variable name between filename and classname.

tree/treeplayer/src/TTreePlayer.cxx
```diff
@@ -733,18 +733,20 @@ Int_t TTreePlayer::MakeClass(const char *classname, const char *option)
    opt.ToLower();
 
    // Connect output files
-   if (!classname) classname = fTree->GetName();
-   const TString vclassname = ROOT::Internal::GetCppName(classname);
+   const TString fileNameStem = classname ? classname : fTree->GetName();
+   const TString cppClassName = ROOT::Internal::GetCppName(fileNameStem);
+   if (cppClassName != fileNameStem)
+      Warning("TTreePlayer::MakeClass", "The %s name provided ('%s') is not a valid C++ identifier and will be converted to '%s', the code produced will likely fail to compile.",(classname ? "class" : "tree"), fileNameStem.Data(), cppClassName.Data());
```
pcanal, line 739: Since we now convert both the case where the name comes from the `TTree` and from the argument the code should always compile, right?
```suggestion
      Warning("TTreePlayer::MakeClass", "The %s name provided ('%s') is not a valid C++ identifier and will be converted to '%s'.",(classname ? "class" : "tree"), fileNameStem.Data(), cppClassName.Data());
```

pcanal (approved):
LGTM.  Thanks.  We should Squash on Merge this PR.

PR #16996, opened 2024-11-20
[win] Fix potential test failure on Windows

bellenot:
Fixes the following potential failure on some Windows machines:
```
2: ======================================================================
2: FAIL: test_all_global_classes_prefix (pythonization_decorator.PythonizationDecorator.test_all_global_classes_prefix)
2: ----------------------------------------------------------------------
2: Traceback (most recent call last):
2:   File "C:\root-dev\git\master\bindings\pyroot\pythonizations\test\pythonization_decorator.py", line 269, in test_all_global_classes_prefix
2:     self.assertEqual(executed.pop(0), class_name)
2: AssertionError: 'TApplication' != 'Foo'
2: - TApplication
2: + Foo
```

### DIFF
--- a/bindings/pyroot/pythonizations/test/pythonization_decorator.py
+++ b/bindings/pyroot/pythonizations/test/pythonization_decorator.py
@@ -13,7 +13,7 @@ class PythonizationDecorator(unittest.TestCase):
     # in some tests, and because of immediate pythonization they will be
     # processed by the pythonizors. Just ignore them
     exclude = [ 'TClass', 'TSystem', 'TUnixSystem', 'TMacOSXSystem',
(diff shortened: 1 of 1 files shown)


vepadulano (approved):
We investigated the potential source of failure: on some Windows configuration the `TApplication` appears in the list of Pythonized classes within the test and was not written in the `exclude` list. In general, I believe the test should be refactored to only check whether the classes that are relevant for each Pythonization test are being properly Pythonized, so this `exclude` list should not be there in the first place imho. Given that this particular quirk does not appear on our CI and does not affect the functionality and test coverage, I think we can merge it as is.

PR #16997, opened 2024-11-20
[ntuple] Introduce and use new function `MakeUninitArray`

hahnjo:
With `std::make_unique<T[]>`, the array elements are value-initialized which is oftentimes not necessary for buffers. See also the message of commit 975989d6d2 ("Do not memset() a buffer that will be overwritten"). In particular, that commit changed the usage in `RPageSinkBuf` to allocate sealed page buffers, which I "simplified" in commit 80018cd764. This should become more clear with the explicit function name.


pcanal (approved):

silverweed:
> With std::make_unique<T[]>, the array elements are value-initialized 

TIL

silverweed (approved):

jblomer (approved):
Nice!

PR #16998, opened 2024-11-20
[core] Ensure non-repeated converted ids and valid return if empty input

ferdymercury:
# This Pull request:

## Changes or fixes:

Suppose a TTree with valid branch names: "1_coef", "2_coef". With TTree MakeSelector, the converted cpp names now would lead to a duplication of variables. Fix this by prepending a char instead of removing the digits. Also, for the corner case that one passes an empty string, return the "minimally valid variable name in C++", ie just a "_". int _ is a valid variable name.

Related to https://github.com/root-project/root/pull/16992

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/core/clingutils/src/TClingUtils.cxx
+++ b/core/clingutils/src/TClingUtils.cxx
@@ -3315,7 +3315,7 @@ void ROOT::TMetaUtils::GetCppName(std::string &out, const char *in)
    }
 
    // If out is empty, or if it starts with a number, it's not a valid C++ variable. Prepend a "_"
-   if (out.empty() || is_digit(out[0]))
+   if (out.empty() || isdigit(out[0]))
       out.insert(out.begin(), '_');
 }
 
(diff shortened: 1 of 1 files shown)


core/clingutils/src/TClingUtils.cxx
```diff
@@ -3313,10 +3314,9 @@ void ROOT::TMetaUtils::GetCppName(std::string &out, const char *in)
          out.push_back(c);
    }
 
-   // Remove initial numbers if any
-   auto firstNonNumber = out.find_first_not_of("0123456789");
-   if (firstNonNumber != std::string::npos)
-      out.replace(0,firstNonNumber,"");
+   // If out is empty, or if it starts with a number, it's not a valid C++ variable. Prepend a "_"
+   if (out.empty() || std::is_digit(out[0]))
+      out.insert(out.begin(), "_");
```
ferdymercury, line 3319:
```suggestion
      out.insert(out.begin(), '_');
```

pcanal (approved):
LGTM.  Thanks.

pcanal (approved):
This need to be "squashed and merged" when ready.

PR #16999, opened 2024-11-20
[RF] Implement channel masking for simultaneous likelihoods

guitargeek:
Upstreaming a feature from CMS combine.

Draft for now.

### DIFF
--- a/roofit/roofitcore/src/FitHelpers.cxx
+++ b/roofit/roofitcore/src/FitHelpers.cxx
@@ -51,6 +51,7 @@
 #endif
 
 using RooFit::Detail::RooNLLVarNew;
+using RooFit::Detail::RooSimNLL;
 
 namespace {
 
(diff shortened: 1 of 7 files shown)


roofit/roofitcore/src/FitHelpers.cxx
```diff
@@ -357,7 +358,7 @@ std::unique_ptr<RooAbsArg> createSimultaneousNLL(RooSimultaneous const &simPdf,
    }
 
    // Time to sum the NLLs
-   auto nll = std::make_unique<RooAddition>("mynll", "mynll", nllTerms);
+   auto nll = std::make_unique<RooSimNLL>("mynll", "mynll", nllTerms, simCat, true);
```
vgvassilev, line 361: Are the test failures due to this change?

guitargeek:
That was a prototype until I got reminded by ATLAS people that this feature should be implemented differently: by using ranges on the RooCategory objects that define the channels.

PR #17004, opened 2024-11-21
[ntuple] Fixes lookup & searching in the descriptor

jblomer:
Fixes several instances of lookups in the descriptor from linear to logarithmic complexity. As a result, many of the limit tests results improve significantly. So much so that I think we can turn on most of them on a regular basis.

Relies on #16986

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
@@ -427,6 +427,7 @@ class RClusterGroupDescriptor {
 private:
    DescriptorId_t fClusterGroupId = kInvalidDescriptorId;
    /// The cluster IDs can be empty if the corresponding page list is not loaded.
+   /// Otherwise, cluster ids are sorted by first entry number.
    std::vector<DescriptorId_t> fClusterIds;
    /// The page list that corresponds to the cluster group
    RNTupleLocator fPageListLocator;
@@ -1306,7 +1307,7 @@ public:
       fClusterGroup.fNClusters = nClusters;
       return *this;
    }
-   void AddClusters(const std::vector<DescriptorId_t> &clusterIds)
+   void AddSortedClusters(const std::vector<DescriptorId_t> &clusterIds)
    {
       if (clusterIds.size() != fClusterGroup.GetNClusters())
          throw RException(R__FAIL("mismatch of number of clusters"));
--- a/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
@@ -279,6 +279,10 @@ public:
       std::size_t ExtendToFitColumnRange(const RColumnRange &columnRange, const Internal::RColumnElementBase &element,
                                          std::size_t pageSize);
 
+      /// Has the same length than fPageInfos and stores the sum of the number of elements of all the pages
+      /// up to and including a given index. Used for binary search in Find().
+      std::vector<NTupleSize_t> fCumulativeNElements;
+
    public:
       /// We do not need to store the element size / uncompressed page size because we know to which column
       /// the page belongs
@@ -319,6 +323,7 @@ public:
          RPageRange clone;
          clone.fPhysicalColumnId = fPhysicalColumnId;
          clone.fPageInfos = fPageInfos;
+         clone.fCumulativeNElements = fCumulativeNElements;
          return clone;
       }
 
--- a/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
@@ -572,6 +572,9 @@ private:
    std::vector<RExtraTypeInfoDescriptor> fExtraTypeInfoDescriptors;
    std::unique_ptr<RHeaderExtension> fHeaderExtension;
 
+   // We don't expose this publicy because when we add sharded clusters, this interface does not make sense anymore
+   DescriptorId_t FindClusterId(NTupleSize_t entryIdx) const;
+
 public:
    static constexpr unsigned int kFeatureFlagTest = 137; // Bit reserved for forward-compatibility testing
 
--- a/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
@@ -577,7 +577,7 @@ private:
    std::vector<RExtraTypeInfoDescriptor> fExtraTypeInfoDescriptors;
    std::unique_ptr<RHeaderExtension> fHeaderExtension;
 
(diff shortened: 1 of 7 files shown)


tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -561,12 +567,19 @@ private:
...
+   /// Note that this list is empty during the descriptor building process and will only be
+   /// created when the final descriptor is extracted from the builder.
+   std::vector<DescriptorId_t> fSortedClusterGroupIds;
    /// May contain only a subset of all the available clusters, e.g. the clusters of the current file
    /// from a chain of files
    std::unordered_map<DescriptorId_t, RClusterDescriptor> fClusterDescriptors;
    std::vector<RExtraTypeInfoDescriptor> fExtraTypeInfoDescriptors;
    std::unique_ptr<RHeaderExtension> fHeaderExtension;
 
+   // We don't expose this publicy because when we add sharded clusters, this interface does not make sense anymore
```
silverweed, line 580: Typo: publicy
Also, it should probably be a TODO?

tree/ntuple/v7/src/RNTupleDescriptor.cxx
```diff
@@ -372,13 +372,58 @@ ROOT::Experimental::RNTupleDescriptor::FindPhysicalColumnId(DescriptorId_t field
...
+      return kInvalidDescriptorId;
+
+   // Binary search in the cluster group list, followed by a binary search in the clusters of that cluster group
+
+   std::size_t cgLeft = 0;
+   std::size_t cgRight = GetNClusterGroups() - 1;
+   while (cgLeft <= cgRight) {
+      const std::size_t cgMidpoint = (cgLeft + cgRight) / 2;
+      const auto &clusterIds = GetClusterGroupDescriptor(fSortedClusterGroupIds[cgMidpoint]).GetClusterIds();
+      R__ASSERT(!clusterIds.empty());
```
hahnjo, line 396: unrelated (for now): this requires deserializing all page lists to populate all cluster group descriptors. In the future, we may first want to search loaded cluster groups under the assumption that by loading the (global) entry first, we already have the necessary information...

tree/ntuple/v7/src/RNTupleDescriptor.cxx
```diff
@@ -428,30 +428,73 @@ ROOT::Experimental::RNTupleDescriptor::FindClusterId(DescriptorId_t physicalColu
...
    const auto &clusterDesc = GetClusterDescriptor(clusterId);
-   auto firstEntryInNextCluster = clusterDesc.GetFirstEntryIndex() + clusterDesc.GetNEntries();
-   // TODO(jblomer): binary search?
-   for (const auto &cd : fClusterDescriptors) {
-      if (cd.second.GetFirstEntryIndex() == firstEntryInNextCluster)
-         return cd.second.GetId();
-   }
-   return kInvalidDescriptorId;
+   const auto firstEntryInNextCluster = clusterDesc.GetFirstEntryIndex() + clusterDesc.GetNEntries();
+   return FindClusterId(firstEntryInNextCluster);
```
hahnjo, line 499: Is it worth shortcutting the common case here and check if `clusterId + 1` contains `firstEntryInNextCluster`?

tree/ntuple/v7/src/RNTupleDescriptor.cxx
```diff
@@ -428,30 +428,73 @@ ROOT::Experimental::RNTupleDescriptor::FindClusterId(DescriptorId_t physicalColu
...
    const auto &clusterDesc = GetClusterDescriptor(clusterId);
-   // TODO(jblomer): binary search?
-   for (const auto &cd : fClusterDescriptors) {
-      if (cd.second.GetFirstEntryIndex() + cd.second.GetNEntries() == clusterDesc.GetFirstEntryIndex())
-         return cd.second.GetId();
-   }
-   return kInvalidDescriptorId;
+   if (clusterDesc.GetFirstEntryIndex() == 0)
+      return kInvalidDescriptorId;
+   return FindClusterId(clusterDesc.GetFirstEntryIndex() - 1);
```
hahnjo, line 508: same question here

hahnjo (commented):
For the limits test, we should also check the memory consumption in addition to the time...

hahnjo:
`Limits_ManyPagesOneEntry` fails in the CI: it fills a `std::vector<int>` with 100 million elements and writes it as a single page. That probably means memory usage in the order of GBs, we probably should leave it disabled.

tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
jblomer: I don't think this is a TODO. We never plan to make this a public interface. Because with sharded clusters, this interface would be confusing (multiple possible clusters for the same entry number).

tree/ntuple/v7/src/RNTupleDescriptor.cxx
jblomer: Indeed. When we implement partial loading of page lists, we need to modify this, e.g. to first look into the available page lists and then load the remaining ones or so.

tree/ntuple/v7/src/RNTupleDescriptor.cxx
jblomer: That's a good question. The "problem" is that I think currently that shortcut will always trigger. However, I also don't want to rely on descriptor ID ordering... I need to think about it.

tree/ntuple/v7/test/ntuple_limits.cxx
```diff
@@ -160,13 +160,13 @@ TEST(RNTuple, DISABLED_Limits_ManyPages)
    }
 }
 
-TEST(RNTuple, DISABLED_Limits_ManyPagesOneEntry)
+TEST(RNTuple, Limits_ManyPagesOneEntry)
```
hahnjo, line 167: This test fails in the CI so we should not enable it either

tree/ntuple/v7/test/ntuple_limits.cxx
```diff
@@ -251,7 +251,7 @@ TEST(RNTuple, DISABLED_Limits_LargePage)
    }
 }
 
-TEST(RNTuple, DISABLED_Limits_LargePageOneEntry)
+TEST(RNTuple, Limits_LargePageOneEntry)
```
hahnjo: maybe don't enable it to disable it again two commits later...

tree/ntuple/v7/test/ntuple_limits.cxx
hahnjo: ping

hahnjo (commented):
Code changes LGTM, `Limits_ManyPagesOneEntry` still needs to be disabled for the CI to pass.

hahnjo (approved):

jblomer:
> Code changes LGTM, `Limits_ManyPagesOneEntry` still needs to be disabled for the CI to pass.

I think the problem is the recursive `ReadV` that can exhaust the stack...

PR #17006, opened 2024-11-21
[ntuple] Micro-optimize `RVectorField::AppendImpl`

hahnjo:
By allowing the compiler to use a single instruction for quotient and remainder, there is a significant improvement in the RNTuple serialization throughput (writing to `/dev/null`):

 * For my "standard" benchmark of writing random float values into a vector with a size determined by a Poisson distribution (mu = 5), this change reduces run time by around 2%.
 * For well tuned sequential writing of constant data (one fixed size vector), the reduction can be up to 7.5%.


pcanal (approved):
LGTM.  Would we gain anything by removing the `R__ASSERT`?

jblomer (approved):

hahnjo:
> Would we gain anything by removing the `R__ASSERT`?

According to my tests, the gain would be small on top of this optimization proposed here; less than 1%, interestingly it seems a bit more in the "standard" benchmark than in the tuned case... I think for now, we should keep it as a safeguard for the type punning that we do.

Issue #16976, opened 2024-11-19
Strange overflow bin bar when plotting TH1D with X1 option


karuboniru:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

ROOT produces strange result when plotting after setting excessive range via SetRange/SetRangeUser, namely the "bar" over X axis can overlap between different bins
![plot](https://github.com/user-attachments/assets/ab04df44-614e-4921-bb53-793835b0082f)

### Reproducer

```cxx
#include <TCanvas.h>
#include <TH1D.h>
#include <TRandom.h>
#include <TStyle.h>
#include <array>
#include <memory>

int main() {
  gStyle->SetOptStat(0);
  auto c1 = std::make_shared<TCanvas>("c1", "c1", 800, 600);
  auto x_bins = std::to_array<double>({0, 0.3, 0.5, 0.6, 1.0});
  auto h1 =
      std::make_shared<TH1D>("h1", "h1", x_bins.size() - 1, x_bins.data());
  for (int i = 0; i < 1000; i++) {
    h1->Fill(gRandom->Gaus(0.5, 1));
  }
  h1->GetXaxis()->SetRange(-1, 6);
  h1->Draw("X1 E0");
  c1->SaveAs("plot.eps");
}
```

It seems that ths issue is related to non-uniform bin width, maybe related to the code where the bar with of the overflow bin is determined. 

### ROOT version

```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.32.06                        https://root.cern |
  | (c) 1995-2024, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Sep 23 2024, 00:00:00                 |
  | From tags/6-32-06@6-32-06                                        |
  | With clang version 18.1.8 (Fedora 18.1.8-3.fc41)                 |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------
```

This reproduces on many versions.

### Installation method

Self built

### Operating system

Linux Fedora 41

dpiparo:
@couet , was this fixed in master perhaps? I get:
```
Warning in <TAxis::TAxis::SetRange>: first < 0, 0 is used
Warning in <TAxis::TAxis::SetRange>: last > fNbins+1, fNbins+1 is used
```
![image](https://github.com/user-attachments/assets/dca407f2-dd74-42f5-9f9a-9826f57f12fb)

[...]

PR #17007, opened 2024-11-21
Several fixes in TAxis

couet:
[This issue](https://github.com/root-project/root/issues/16976) revealed a graphics bug when displaying underflow and overflow bins for non-equidistant histogram bins with error bars. The problem arose from a mismatch in the bin width computation in TAxis::GetBinWidth. When the range included underflow or overflow bins, the bin width was calculated using the closest bins, unlike the approach in TAxis::GetBinLowEdge and TAxis::GetBinUpEdge, which apply a specific computation in such cases. This PR updates TAxis::GetBinWidth to follow the same logic as these methods.

This PR also updates the documentation for TAxis::SetRange and TAxis::SetRangeUser, clarifying the interpretation of ranges that extend beyond the normal bin limits. With these clarifications, the warning about bin numbers outside the valid range can now be suppressed.

### DIFF
--- a/hist/hist/src/TAxis.cxx
+++ b/hist/hist/src/TAxis.cxx
@@ -536,11 +536,14 @@ Double_t TAxis::GetBinUpEdge(Int_t bin) const
 
 ////////////////////////////////////////////////////////////////////////////////
 /// Return bin width
+///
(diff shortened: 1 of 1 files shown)


dpiparo:
If we approve this nice fix, we should definitively back-port it to 6.34 and even 6.32

dpiparo (approved):
LGTM, thanks! If you want, you could consider the formatting suggestions given by `clang-format`.

couet:
I'll check clang-format comments

Issue #17003, opened 2024-11-21
Fail to get TObject or TKey with one or more spaces at the end of its name from TFile


wxzhan:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

When I create a TObject with its name ends with space, and I write it into a TFile, its corresponding TKey's name will lose the space at the end. So I can't get it using neither `TFile::Get` nor `TFile::GetKey`.

### Reproducer

[testKey.zip](https://github.com/user-attachments/files/17846251/testKey.zip)
Here it is

In test.cxx I create two TH1D ended with and without space, written into save.root. Just run `root test.cxx`

To test the bug, run `root read.cxx`. You'll see first two lines indicate the failure of `Get()`. And another two lines to prove that the space in TKey is missing but in TObject it is correctly stored.

### ROOT version
```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.30/02                        https://root.cern |
  | (c) 1995-2023, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Aug 26 2024, 07:41:41                 |
  | From heads/master@tags/v6-30-02                                  |
  | With g++ (GCC) 13.1.0                                            |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------
```
### Installation method

build from source

### Operating system

CentOS Linux release 7.9.2009 (Core)

silverweed:
Hello @wxzhan,
it seems that stripping whitespaces is [the intended behavior](https://github.com/root-project/root/blob/v6-34-00-patches/io/io/src/TDirectoryFile.cxx#L1945-L1956) of TFile, but I'm not aware why that is the case. I don't know why the trailing spaces should be trimmed from the name but if that's the intention we should at least raise an error if an invalid name is given to an object, rather than silently failing to look it up.
Perhaps @pcanal can shed more light on this.

ferdymercury:
From 24y ago: https://root.cern.ch/root/roottalk/roottalk00/0292.html

pcanal:
It was likely a minor attempt to ensure that the name was useable as a C++ identifier on the command line for automatic search and read.

dpiparo:
Thanks to #17008 now a clear warning is prompted when writing objects with a name ending with a ' '.

PR #17008, opened 2024-11-21
[io] warn user about the trailing blanks being stripped from keyname upon writing

ferdymercury:
# This Pull request:

## Changes or fixes:
Applies the warning that @silverweed suggested.
See https://github.com/root-project/root/issues/17003


pcanal (approved):
LGTM.

PR #17011, opened 2024-11-22
[interpreter] Silence warnings raised by -Wnonnull

dpiparo:
This PR makes GHA reports more readable. In any case, we will not fix llvm, but rather update it to its most recent versions.


hahnjo:
IIRC the problem is not in `interpreter/llvm` where we ignore all warnings anyway, but from files in `core/metacling` (?) including Clang header files. However, we obviously don't want to suppress all warnings in our own code...

PR #17012, opened 2024-11-22
[ntuple] add missing include in RFieldMeta

silverweed:
Fixes a compiler error:
```
no matching conversion for functional-style cast from 'const TObjArray' to 'ROOT::Detail::TRangeStaticCast<TObjString>' (aka 'TRangeCast<TObjString, false>')
```


hahnjo (approved):
Sorry about that; I checked multiple times that I didn't need it in my compilation (I guess you need it with `dev=ON`?) and that the type was not explicitly mentioned in the file, so I thought I can drop it...

PR #17013, opened 2024-11-22
[ntuple] fix out-of-bounds in Real32Quant test

silverweed:
A wrong copypaste caused UB in a test for Real32Quant (allocating only 1 byte of storage when 4 were needed)

## Checklist:

- [x] tested changes locally


hahnjo (approved):
LGTM; let's see if this indeed make Windows green, but it looks promising :crossed_fingers:

PR #17014, opened 2024-11-22
[ntuple] Allow configuration of initial page size

hahnjo:
... instead of the initial number of elements. This allows the user to turn off the adaptive page size algorithm completely by specifying the same values for the initial and maximum page size.


pcanal (approved):
LGTM

jblomer (approved):

PR #17025, opened 2024-11-25
[ntuple] remove ArgsT template from MakeField

jblomer:
Remove the parameter pack from RNTupleModel::MakeField() that has been
used to set an initial value for a field. This API only makes sense with
a default model. E.g., it is confusing with bare models (initial value
ignored) or with the parallel writer or if multiple entries are created.

In practice, the functionality to set an initial value is only useful
for tests and can in many cases be replaced by immediately dereferencing
the shared pointer returned from MakeField().

Companion PR: https://github.com/root-project/roottest/pull/1228

### DIFF
--- a/tree/dataframe/test/dataframe_unified_constructor.cxx
+++ b/tree/dataframe/test/dataframe_unified_constructor.cxx
@@ -20,7 +20,7 @@ protected:
       for (const auto &fName : fFileNames) {
          {
             auto modelWrite = RNTupleModel::Create();
-            auto pt = modelWrite->MakeField<float>("ntuple_pt", 11.f);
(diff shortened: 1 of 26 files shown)


silverweed (approved):

jblomer:
> Does this change means that only classes with a C++ default constructors (as opposed to an I/O constructor) are supported?

The I/O constructor was supported by the type-erased `RFieldBase::Create()` API but not through `MakeField`. This is unrelated to this PR though.

Handling classes with an I/O constructor is now fixed, together with two more unrelated fixes that this change uncovered.

pcanal (approved):

PR #17026, opened 2024-11-25
Use strings for colors in Python tutorials

guitargeek:
This makes the code less verbose and more pythonic.

Sister PR in `roottest`:
https://github.com/root-project/roottest/pull/1227


linev:
I would provide return value for all `Set...Color` methods with color index.
One could check then if method really changed something.

guitargeek:
I know, that would be nice, but the existing versions of the `Set*Color(` methods are virtual, so we can't change their return types without causing incompatibilities :(
For example:
https://github.com/root-project/root/blob/master/core/base/inc/TAttLine.h#L40

tutorials/dataframe/df103_NanoAODHiggsAnalysis.py
```diff
@@ -206,23 +206,23 @@ def plot(sig, bkg, data, x_label, filename):
...
+    h_cmb.SetLineColor("kRed")
     h_cmb.SetLineWidth(2)
     h_cmb.SetMaximum(18)
     h_cmb.SetStats(False)
 
     h_bkg.SetLineWidth(2)
     h_bkg.SetFillStyle(1001)
-    h_bkg.SetLineColor(ROOT.kBlack)
-    h_bkg.SetFillColor(ROOT.kAzure - 9)
+    h_bkg.SetLineColor("kBlack")
```
pcanal, line 216: What is the typical Python behavior if the name is incorrect spelt?  I.e. should the new `SetLineColor` (and co) print an error message?

guitargeek: Right now, it will just silently do nothing. Not sure yet if this should be treated as an Error: also when you call the old `SetLineColor()` with an invalid integer, it doesn't complain, and when drawing it falls back to black.

guitargeek: You are right, the cases are not comparable. Would raising a `std::invalid_argument` be okay for you?

It would look like this:
```txt
  File "/home/rembserj/spaces/master/root/src/root/test.py", line 5, in <module>
    h.SetLineColor("kReed")
cppyy.gbl.std.invalid_argument: void TAttLine::SetLineColor(const string& lcolor) =>
    invalid_argument: "kReed" is not a valid color name
```

pcanal: Sounds fair.

pcanal (changes requested):
LGTM but we need to rebase ...

guitargeek:
Thanks for the ping @pcanal! Is it okay now? Could you approve together with the [roottest PR](https://github.com/root-project/roottest/pull/1227)?

By the way there was recently a forum discussion about color names in Python, which tells us again that this is a feature that our users will appreciate:
https://root-forum.cern.ch/t/color-names-in-python/62543/3

PR #17033, opened 2024-11-25
[RF] Use TRandom3 explicitly in RooFit

guitargeek:
This demonstrates the good practice of not relying of global variables.


aaronj0 (approved):
LGTM!

PR #17038, opened 2024-11-25
[RF] Remove `rf603_multicpu` tutorial

guitargeek:
The tutorial is not adding much information. It is a stripped down version of the `rf316` tutorial, existing just to show the `NumCPU()` option. This option is not even active anymore with the new vectorizing CPU backend, because it is faster than the old backend with any number of cores anyway (thanks to the relatively poor scaling of that one).

Compare also:
  * https://root.cern/doc/master/rf316__llratioplot_8C.html
  * https://root.cern/doc/master/rf603__multicpu_8C.html


dpiparo (approved):
LGTM!

PR #17041, opened 2024-11-25
[RF] Remove RooWorkspace tutorial for CINT to Cling migration

guitargeek:
Remove outdated tutorials and don't use second argument to RooWorkspace constructor that was only relevant in the CINT times.


dpiparo (approved):
LGTM. Thanks for this proactive nice improvement. I would even argue that the HLFactory should be deprecated in 6.36 and removed in 6.38.

PR #17042, opened 2024-11-25
[RF] Skip normalization over nEvents in RooSimultaneous if necessary

guitargeek:
If only one of the pdfs in the RooSimultaneous can't provide an expected
number of events, calculating the relative number of events is futile
and will result in exceptions when trying to ask these pdfs for the
expected number of events.

This fixes a bug reported by @gartrog from ATLAS (crash when printing
workspaces with such pdfs).


dpiparo (approved):
LGTM, thanks. Would it be easy to add a test for this fix?

guitargeek:
Sure, yes!

PR #17043, opened 2024-11-26
[core] Adjust kp6Violet to the correct Petroff value, #7a21dd

### LINKED ISSUE #17040: Small difference between kp6Violet implementation and official value from Petroff paper
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

We noticed that `kp6Violet` https://github.com/couet/root/blob/master/core/base/src/TColor.cxx#L1249 is implemented as `7021dd` and not `7a21dd` (see https://github.com/mpetroff/accessible-color-cycles/tree/master for reference). This is also visible in the graphics in the docs (https://root.cern.ch/doc/master/classTColor.html#C031)

FYI @lenzip @anigamova @oglez

### Reproducer

N/A

### ROOT version

From [v6.34.00-rc1](https://github.com/root-project/root/releases/tag/v6-34-00-rc1)

### Installation method

N/A

### Operating system

N/A

### Additional context

_No response_

dpiparo:
fixes #17040


couet (approved):
It looks good to me.

PR #17045, opened 2024-11-26
[ntuple] Add RNTuple[Parallel]Writer::CommitDataset()

jblomer:
Allows for explicit dataset commits. This is useful
- for the use in Python
- to extract the final metrics before destruction
- to properly handle exceptions during the final commit


hahnjo (approved):
LGTM

Issue #6607, opened 2020-10-11
TClassEdit::GetNormalizedName does not strip std::allocator on Windows


jblomer:
### Describe the bug
The type name normalization `TClassEdit::GetNormalizedName` does not strip the allocator off STL classes on Windows. On Linux, it works.

### Expected behavior
On both, Windows and Linux the following should be true

```
std::string n;
TClassEdit::GetNormalizedName(n, "std::vector<float, class std::allocator<float>>");
n == std::string("vector<float>");
```

Note that even on Windows the allocator _is_ stripped for `std::vector<float, std::allocator<float>>` (i.e. without prefixing `std::allocator` by `class`)

### Additional context
The issue was found while testing PR #6548

pcanal:
There is code to explicitly detect and remove the allocator.  Most likely `TClassEdit::IsDefAlloc` needs to be updated to take into account that the word `class` might be prefixed.  Albeit it is odd/unexpected that Clang would put the class keyword there (it is supposed be configured not to).

Issue #10749, opened 2022-06-13
RNTupleDS should check for column type mismatches


eguiraud:
This issue is just a reminder that we need to resolve this TODO:

https://github.com/root-project/root/blob/89776f40f6f9a0f1ea769ed86f8280696b85e01f/tree/dataframe/src/RNTupleDS.cxx#L288-L296

so that e.g. the following wrong usage errors out:

```cpp
#include <ROOT/RNTuple.hxx>
#include <ROOT/RNTupleDS.hxx>
#include <ROOT/RDataFrame.hxx>

int main() {
  {
    auto m = ROOT::Experimental::RNTupleModel::Create();
    auto x = m->MakeField<int>("x", 42);
    auto r = ROOT::Experimental::RNTupleWriter::Recreate(std::move(m), "n",
                                                         "f.root");
    r->Fill();
  }

  auto df = ROOT::Experimental::MakeNTupleDataFrame("n", "f.root");
  df.Filter([] (std::string &x) { return !x.empty(); }, {"x"}).Count().GetValue();
}
```

(on my machine the test above executes without crashes, but that's just luck)

jblomer:
We have a draft PR for fixing it it #6548 which is blocked by #6607
@pcanal Do you think that issue can be addressed in `TClassEdit`?

pcanal:
Yes, of course.  There is code to explicitly detect and remove the allocator.  Most likely `TClassEdit::IsDefAlloc` needs to be updated to take into account that the word `class` might be prefixed.  Albeit it is odd/unexpected that Clang would put the class keyword there (it is supposed be configured not to).

enirolf:
Fixed by #17749

PR #9284, opened 2021-11-12
TClassEdit memory hoarding

Axel-Naumann:
Addressed the valid-typename case of https://github.com/root-project/root/issues/9029

Issue #9029, opened 2021-09-24
Repeated calls to TClass::GetClass/TClass::ResolveTypedef hoarding memory


amete:
Hi everyone,

We came across this issue in ATLAS recently. There were similar reports in the past, such as [ROOT-7254](https://sft.its.cern.ch/jira/browse/ROOT-7254), but there is none that's still open as far as I could see. 

### Describe the bug
Repeatedly calling `TClass::GetClass`/`TClass::ResolveTypedef` w/ an input such as `std::vector<int>::value_size` seems to create memory hoarding.

### To Reproduce
Either of these python scripts should reproduce the issue:

```
import ROOT

while True:
    ROOT.TClass.GetClass("std::vector<int>::value_size")
```

```
import ROOT

while True:
    ROOT.TClassEdit.ResolveTypedef("std::vector<int>::value_size")
```

### Setup
Tested with both ROOT versions 6.24/00 and 6.24/06 in the ATLAS production environment.

Please let us know, many thanks.

Best,
Serhan

eguiraud:
At a quick glance it looks like `TClass::GetClass` calls the interpreter and jitting allocates a little bit every time, notoriously never giving that memory back until the end of the application. I can't tell whether `TClass::GetClass` could avoid any and all jitting on subsequent calls, but it seems reasonable to think so. @Axel-Naumann @pcanal ?

eguiraud:
In case it helps here's the output of massif for the first of the two reproducers: https://cernbox.cern.ch/index.php/s/aACGp2JdPzAQUfW (it should be readable with `ms_print`, the most relevant frame is the very last one)

akmorley:
I was wondering if there was a time scale for this issue being fixed as it impacts one of the calibration workflows in ATLAS?

Axel-Naumann:
Just FYI, `std::vector<int>::value_size` doesn't exist, which is why `TClass::GetClass()` tries again and again to see whether it has been made available in between the calls. This would not be an issue for an existing type, such as `std::vector<int>::value_type`

Axel-Naumann:
@amete with the additional info I provided, do you still see this as an issue that ROOT should be addressing?

amete:
Thanks @Axel-Naumann.

I'm afraid I still see the issue even if I ask for an existing type, e.g. `std::vector<int>::value_type`. I just tested the following snippet:

```
import ROOT

while True:
    ROOT.TClass.GetClass("std::vector<int>::value_type")
```

on:

```
[...]

PR #17048, opened 2024-11-26
[core] remove class prefix if any in IsDefAlloc

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/6607
Related: https://github.com/root-project/root/issues/10749

Sibling roottest PR: https://github.com/root-project/roottest/pull/1237

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -621,7 +621,7 @@ static size_t findNameEnd(const std::string &full, size_t pos)
 bool TClassEdit::IsDefAlloc(const char *allocname, const char *classname)
 {
    string_view a( allocname );
-   const static int clalloclen = strlen("class ");
+   constexpr static int clalloclen = strlen("class ");
    if (a.compare(0,clalloclen,"class ") == 0) {
       a.remove_prefix(clalloclen);
    }
@@ -631,7 +631,7 @@ bool TClassEdit::IsDefAlloc(const char *allocname, const char *classname)
    if (a=="__default_alloc_template<true,0>")   return true;
    if (a=="__malloc_alloc_template<0>")         return true;
 
-   const static int alloclen = strlen("allocator<");
+   constexpr static int alloclen = strlen("allocator<");
    if (a.compare(0,alloclen,"allocator<") != 0) {
       return false;
    }
@@ -680,7 +680,7 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
    string_view a( allocname );
    RemoveStd(a);
 
-   const static int alloclen = strlen("allocator<");
+   constexpr static int alloclen = strlen("allocator<");
    if (a.compare(0,alloclen,"allocator<") != 0) {
       return false;
    }
@@ -688,7 +688,7 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
 
    RemoveStd(a);
 
-   const static int pairlen = strlen("pair<");
+   constexpr static int pairlen = strlen("pair<");
    if (a.compare(0,pairlen,"pair<") != 0) {
       return false;
    }
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -621,6 +621,10 @@ static size_t findNameEnd(const std::string &full, size_t pos)
 bool TClassEdit::IsDefAlloc(const char *allocname, const char *classname)
 {
    string_view a( allocname );
+   const static int clalloclen = strlen("class ");
+   if (a.compare(0,clalloclen,"class ") == 0) {
+      a.remove_prefix(clalloclen);
+   }
    RemoveStd(a);
 
    if (a=="alloc")                              return true;
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -26,6 +26,7 @@
 #include <memory>
 #include <string_view>
 #include <algorithm>
+#include <string>
 
 #include "TSpinLockGuard.h"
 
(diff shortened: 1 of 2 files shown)


dpiparo:
Thanks for this PR. Would it make sense to also transform in a test (or add to the one written for this PR) the example reported in the original issue ?
```cpp
std::string n;
TClassEdit::GetNormalizedName(n, "std::vector<float, class std::allocator<float>>");
n == std::string("vector<float>"); 
```

ferdymercury:
@bellenot Do you understand by chance why the build fails on Windows after this change?

bellenot:
@ferdymercury I'll check after the hackathon (in a couple of days...)

ferdymercury:
> @ferdymercury I'll check after the hackathon (in a couple of days...)

Could it be maybe related to?:
https://github.com/root-project/root/issues/9029
https://github.com/root-project/root/pull/9284

bellenot:
@ferdymercury I'll debug it and let you know (need a full debug build)

bellenot:
I found the problem. Trying to find the proper fix.

bellenot:
@ferdymercury maybe not the correct and maybe not the final solution, but can you try this diff (including your change):
```
diff --git a/core/foundation/src/TClassEdit.cxx b/core/foundation/src/TClassEdit.cxx
index d3115602c3..759dc0a730 100644
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -621,6 +621,10 @@ static size_t findNameEnd(const std::string &full, size_t pos)
 bool TClassEdit::IsDefAlloc(const char *allocname, const char *classname)
 {
    string_view a( allocname );
+   const static int clalloclen = strlen("class ");
+   if (a.compare(0,clalloclen,"class ") == 0) {
+      a.remove_prefix(clalloclen);
+   }
    RemoveStd(a);

    if (a=="alloc")                              return true;
@@ -1538,9 +1542,8 @@ static void ResolveTypedefImpl(const char *tname,
       if (strncmp(tname+cursor,"const ",6) == 0) {
          cursor += 6;
          if (modified) result += "const ";
+         constprefix = true;
       }
-      constprefix = true;
-
    }

    if (len > 2 && strncmp(tname+cursor,"::",2) == 0) {
```
@pcanal can check/comment on this, but I think this code:
```
   if (tname[cursor]=='c' && (cursor+6<len)) {
      if (strncmp(tname+cursor,"const ",6) == 0) {
         cursor += 6;
         if (modified) result += "const ";
      }
      constprefix = true;

    }
    ```
is wrong. I.e. `if (tname[cursor]=='c' && (cursor+6<len)) {` is also true for `tname` beginning with `class `, and `constprefix = true;` should be moved inside the `if (strncmp(tname+cursor,"const ",6) == 0) {` condition. Or did I overlooked at this?

**Comment by ferdymercury** [comment]
It's solved now!

**Comment by bellenot** [comment]
> It's solved now!

Congrats!

**Review by bellenot — APPROVED** [review]
> It's solved now!

Congrats! Looks good to me, but I'll let @pcanal give the final green light

File: core/foundation/src/TClassEdit.cxx [hunk 1]
```diff
@@ -1533,14 +1537,17 @@ static void ResolveTypedefImpl(const char *tname,
       }
       while (tname[cursor]==' ') ++cursor;
    }
-
+   if (tname[cursor]=='c' && (cursor+6<len)) {
+      if (strncmp(tname+cursor,"class ",6) == 0) {
+         cursor += 6;
+      }
+   }
    if (tname[cursor]=='c' && (cursor+6<len)) {
```
bellenot, line 1545: BTW, you could simplify it like this:
```suggestion
```

bellenot: AFAICS, in your commit there is twice the same if statement so the `constprefix ` will be set in the second one anyway. Or did I overlook your commit?

ferdymercury: Mm it's almost the same, but in the second one the index cursor is cursor+6, and it's str-comparing against class instead of const, so we are not doing exactly the same.
Class int will go into the first if but not into the second since cursor has increased so constprefix will be false

bellenot: Well, as you prefer, it's only a repetition of identical code anyway, I guess the compiler will optimize it

core/foundation/src/TClassEdit.cxx
```diff
@@ -621,6 +621,10 @@ static size_t findNameEnd(const std::string &full, size_t pos)
 bool TClassEdit::IsDefAlloc(const char *allocname, const char *classname)
 {
    string_view a( allocname );
+   const static int clalloclen = strlen("class ");
```
pcanal, line 624:
```suggestion
   constexpr static int clalloclen = strlen("class ");
```

core/foundation/src/TClassEdit.cxx
pcanal: Actually, doesn't it depends on whether the name can sometimes have both `const ` and `class ` and in which order?

pcanal: Either way, the code should be augment with a comment expressing that we looked into that and what the answer was.

core/foundation/src/TClassEdit.cxx
ferdymercury: I've now tried with an alternative function, let's see if it works.

ferdymercury: Seems solved now

ferdymercury:
@pcanal do you know if the remaining test failures are related to this PR ? thanks 4 the review.

pcanal:
They shouldn't but they don't usually fail either.  Let's rerun the test and see.

ferdymercury:
> They shouldn't but they don't usually fail either. Let's rerun the test and see.

Yes, now it's others which fail, so it seemed random.

Issue #15842, opened 2024-06-13
Unintuitive behavior of `Math::PxPyPzMVector` in calculations


Zehvogel:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

Depending on the mixture of 4-vector types and the order of operations the result can change, e.g.
lv1 - lv2 != -lv2 + lv1
See reproducer for all combinations.

### Reproducer

```python
import ROOT

lv1 = ROOT.Math.PxPyPzEVector(0, 0, 1, 1)
lv2 = ROOT.Math.PxPyPzMVector(lv1)

print(lv1 - lv2)
# (0,0,0,0)
print(lv2 - lv1)
# (0,0,0,0)
print(-lv1 + lv2)
# (0,0,0,0)
print(-lv2 + lv1)
# (0,0,0,2)
```
but I am relatively sure that it will also happen in c++

### ROOT version

```
ROOT Version: 6.33.01
Built for linuxx8664gcc on Jun 12 2024, 23:54:24
From heads/master@v6-31-01-2329-g4fe33d2d91
```

### Installation method

LCG dev3

### Operating system

Alma9

PR #17049, opened 2024-11-26
[skip-ci] add note on coordinate system negation pitfalls

m-fila:
# This Pull request:
Adds a short note about an unintuitive behavior of 4-vector negation for coordinate systems based on mass instead of energy.
The behavior itself is documented in the affected coordinate systems ([`PtEtaPhiM4D::Negate`](https://root.cern/doc/v632/classROOT_1_1Math_1_1PtEtaPhiM4D.html#a3be54ee433d7f3821fb3067289668630), [`PxPyPzM4D::Negate`](https://root.cern/doc/v632/classROOT_1_1Math_1_1PxPyPzM4D.html#a735fc7329a1c390b601e0bb27afd64c0)) but the visibility of this information is a bit limited, hence issues like #15842

The doxygen should be able to translate markdown github-style admonitions to standard doxygen `\note` https://doxygen.nl/manual/markdown.html

@Zehvogel

### DIFF
--- a/documentation/users-guide/MathLibraries.md
+++ b/documentation/users-guide/MathLibraries.md
@@ -2083,6 +2083,9 @@ method, with any vector (`q`) implementing `x()`, `y()` and `z()`.
 Note that the multiplication between two vectors using the operator `*`
 is not supported because it is ambiguous.
 
+> [!note]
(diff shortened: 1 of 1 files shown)


vepadulano (approved):
Thank you for this improvement!

PR #17051, opened 2024-11-26
[RDF] allow setting column names from RDF::FromCSV

silverweed:
Adding a new overload of RDF::FromCSV that accepts a struct with the various options. Instead of having the `readHeaders` option, it has a vector of column names that may be given to set/override the column names. Passing a non-empty array implies `readHeaders = false` and vice-versa.

### DIFF
--- a/tree/dataframe/inc/ROOT/RCsvDS.hxx
+++ b/tree/dataframe/inc/ROOT/RCsvDS.hxx
@@ -80,7 +80,7 @@ protected:
 
 public:
    RCsvDS(std::string_view fileName, bool readHeaders = true, char delimiter = ',', Long64_t linesChunkSize = -1LL,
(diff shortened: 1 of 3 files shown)


pcanal (approved):
LGTM.

vepadulano (approved):

pcanal:
That code does not seems to have the protections that this PR had about different number of columns vs names.  Does the code in the master needs to be updated or are those checks unnecessary?

silverweed:
> That code does not seems to have the protections that this PR had about different number of columns vs names. Does the code in the master needs to be updated or are those checks unnecessary?

Yes, that is what #17312 is introducing (check on the number of elements in the columns)

pcanal:
Thanks.

PR #17053, opened 2024-11-26
[ntuple] Cache and expose field type name in REntry

hahnjo:
This makes entries standalone from the lifetime of the model and its fields, which will be important for the Python API.


jblomer (approved):

tree/ntuple/v7/inc/ROOT/REntry.hxx
```diff
@@ -139,9 +143,9 @@ private:
    {
       if constexpr (!std::is_void_v<T>) {
          const auto &v = fValues[token.fIndex];
-         if (v.GetField().GetTypeName() != RField<T>::TypeName()) {
+         if (fFieldTypes[token.fIndex] != RField<T>::TypeName()) {
             throw RException(R__FAIL("type mismatch for field " + v.GetField().GetQualifiedFieldName() + ": " +
```
pcanal, line 147: The declaration of `v` should probably move inside the `if` statement.

tree/ntuple/v7/inc/ROOT/REntry.hxx
```diff
@@ -79,6 +79,8 @@ private:
    std::vector<RFieldBase::RValue> fValues;
    /// For fast lookup of token IDs given a (sub)field name present in the entry
    std::unordered_map<std::string, std::size_t> fFieldName2Token;
+   /// To ensure that the entry is standalone, a copy of all field types
+   std::vector<std::string> fFieldTypes;
```
pcanal, line 83: This caches allows but does not ensure (in particular future code) to not be reaching out to the model.   Notably the code on (new) line 147 still uses `fValues[..].GetField()` (which uses the model indirectly, isn't it).  Albeit it is 'only' in the error case.  Am I missing something?

hahnjo: Yes indeed. In the future this will be dealt with by nulling the `RFieldBase` pointers when destructing the model. For the noted case, I added a function look up the field name in `fFieldName2Token` - since it's only in the error case, linear complexity is fine.

PR #17055, opened 2024-11-26
Allow passing numpy array to TH1::Fill()

maszyman:
This makes `Fill()` more pythonic as one can do:
```python
import numpy as np
data = np.array([1., 2, 2, 3, 3, 3, 4, 4, 5])
hist = ROOT.TH1F("hist", "hist", 5, 0, 10)
hist.Fill(data)
```
instead of
```py
hist.FillN(len(data), data, np.ones(len(data)))
```


guitargeek (approved):
Thank you very much for this!

Please follow up with a new PR adding documentation.

PR #17059, opened 2024-11-26
Support ipython tab completion when getting item TFile and RooWorkspace

guitargeek:
This makes it more efficient to explore files with RooFit workspaces in ipython and the jupyter notebook.


vepadulano (approved):
Thank you!

PR #17071, opened 2024-11-26
[tutorials] Create a new 'Visualisation tutorials' group

martamaja10:
During Hackathon, together with Monica we came up with an idea to organise the groups of tutorials better. We propose to move all the tutorials categories that are related to visualising data in various ways to be moved to a parent directory "Data visualisation tutorials". We plan to form similar parent directories for other categories, for example, Math and io. 

The parent directory will have a description with what is included inside (not done now as we first should discuss if such solution is acceptable).


couet (approved):

martamaja10:
the failing tests don't seem related (the CI was run quite a few times as well)

agheata (changes requested):
@martamaja10 the tutorials: `assembly.C` and `geodemo.C`  demonstrate different geometry features and their main purpose is not visualization so I would keep them in `tutorials/geom`

martamaja10:
> @martamaja10 the tutorials: `assembly.C` and `geodemo.C` demonstrate different geometry features and their main purpose is not visualization so I would keep them in `tutorials/geom`. No need to re-run the CI after moving them.

Hi Andrei! I moved those tutorials back in the Geometry category (`tutorials/geom`), the other geom tutorials stay in `tutorials/visualisation/geom`. Please let me know if this is good for you.

agheata:
> > @martamaja10 the tutorials: `assembly.C` and `geodemo.C` demonstrate different geometry features and their main purpose is not visualization so I would keep them in `tutorials/geom`. No need to re-run the CI after moving them.
> 
> Hi Andrei! I moved those tutorials back in the Geometry category (`tutorials/geom`), the other geom tutorials stay in `tutorials/visualisation/geom`. Please let me know if this is good for you.

Perfect, thanks!

PR #17075, opened 2024-11-26
[RF] Deprecate RooStats::HLFactory

guitargeek:
The `RooStats::HLFactory` class was deprecated and will be removed in
ROOT 6.38. It provided little advantage over using the RooWorkspace
directly or any of the other higher-level frameworks that exists in the
RooFit ecosystem.

As suggested by @dpiparo: https://github.com/root-project/root/pull/17041#pullrequestreview-2460312336


lmoneta (approved):
Good to remove it, if you did not have any complaints after deprecating the class

PR #17077, opened 2024-11-26
[math] Add vector filling overload for scalar distributions of TRandom

ericcano:
# [math] Add new array filling functions for distributions delivering scalars in TRandom 
## Changes or fixes:

Addition of `TRandom::<Distribution>N()` array filler functions.

```python
>>> vd = np.zeros(100, dtype=np.float64)
>>> r = ROOT.TRandom2(123)
>>> r.GausN(100, vd, 0, 1.0)
>>> vd
array([-0.51452618,  0.11320471, -0.2552159 , -0.9487521 ,  1.25042753,
        -0.34611829,  2.3636373 ,  1.52963734,  0.95990059,  0.21563747,
        -0.48241846,  1.51473977, -1.19058241, -1.62088594,  0.57052152,
         0.15026036,  1.58667136,  0.36062987, -1.23310079,  1.20035556,
        -0.42508447,  0.13440585,  0.63766076,  0.59627148, -0.46659047,
        -1.57966641, -0.23080729, -0.05128992,  1.27449527, -0.3358339 ,
        -2.53090467, -1.54203658, -0.84014657,  0.24656162,  0.08736865,
         0.66275005, -1.47442608, -0.48096595,  0.72262188,  1.07080308,
         0.24917973, -0.4976157 ,  0.47648086, -0.89377689,  0.82566228,
        -1.26325485, -0.17795292,  0.74986627,  1.16974563,  1.41039828,
        -0.01727918, -0.08254512,  0.39879297,  0.73254844,  0.80943978,
         1.79774612,  1.80436324, -1.13506493, -1.18507172, -0.63729355,
         1.08810798,  0.40591267,  0.03376047,  0.49296303,  2.89776279,
        -1.77849263, -0.3847957 , -0.15786366, -0.21491428, -0.9693155 ,
         1.50609877,  0.6414143 , -0.54620441,  1.363164  , -0.63699385,
         1.77694564,  0.62645766,  0.46607116,  1.12608175,  1.67122243,
        -0.40426394,  0.50792749,  0.49501287,  0.05635704,  1.66376808,
         0.06030797, -1.79686209,  1.9065689 , -0.60956203,  0.32817868,
        -1.14337713, -0.020879  ,  1.19166155, -1.55478791, -0.07289054,
         0.77477693, -0.93212149, -1.65628565,  0.29406467,  0.48126101])
```
This pythonisation is validated in unit tests.

We could not overload the existing functions as default values would create ambiguous cases.

## Checklist:

- [X] tested changes locally

### DIFF
--- a/bindings/pyroot/pythonizations/test/trandom_array.py
+++ b/bindings/pyroot/pythonizations/test/trandom_array.py
@@ -1,0 +1,73 @@
+import unittest
+
+import ROOT
+import numpy as np
+
+class TRandomNumpyArrays(unittest.TestCase):
+    """
+    Test for pythonizations thst allow filling numpy arrays with TRandom scalar
+    RNGs.
+    We only cover the technical aspect, ensuring the array if filled. With the seed
+    set, all executions should be repeatable.
+    """
+
+    # Tests
+    def test_binomial(self):
+        r = ROOT.TRandom2(123)
+        vi = np.zeros(100, dtype=np.int32)
+        r.BinomialN(100, vi, 123, 0.12345)
+        self.assertEqual(np.count_nonzero(vi), len(vi))
+
+    def test_breit_wigner(self):
+        r = ROOT.TRandom2(123)
+        vd = np.zeros(100, dtype=np.float64)
+        r.BreitWignerN(100, vd, 123, 0.12345)
+        self.assertEqual(np.count_nonzero(vd), len(vd))
+
+    def test_exp(self):
+        r = ROOT.TRandom2(123)
+        vd = np.zeros(100, dtype=np.float64)
+        r.ExpN(100, vd, 1.5)
+        self.assertEqual(np.count_nonzero(vd), len(vd))
+
+    def test_gaus(self):
+        r = ROOT.TRandom2(123)
+        vd = np.zeros(100, dtype=np.float64)
+        r.GausN(100, vd, 5, 1.5)
+        self.assertEqual(np.count_nonzero(vd), len(vd))
+
+    def test_integer(self):
+        r = ROOT.TRandom2(123)
+        vui = np.zeros(100, dtype=np.uint32)
+        r.IntegerN(100, vui, 15)
+        self.assertGreater(np.count_nonzero(vui), 0.9 * len(vui))
+
+    def test_landau(self):
+        r = ROOT.TRandom2(123)
+        vd = np.zeros(100, dtype=np.float64)
+        r.LandauN(100, vd, 5, 1.5)
+        self.assertEqual(np.count_nonzero(vd), len(vd))
+
+    def test_poisson(self):
+        r = ROOT.TRandom2(123)
+        vull = np.zeros(100, dtype=np.ulonglong)
+        r.PoissonN(100, vull, 15)
+        self.assertGreater(np.count_nonzero(vull), 0.9 * len(vull))
+
+    def test_poissond_d(self):
+        r = ROOT.TRandom2(123)
+        vd = np.zeros(100, dtype=np.float64)
+        r.PoissonDN(100, vd, 15)
+        self.assertGreater(np.count_nonzero(vd), 0.9 * len(vd))
+
+    def test_uniform(self):
+        r = ROOT.TRandom2(123)
+        vd = np.zeros(100, dtype=np.float64)
+        r.UniformN(100, vd, 5.0)
+        self.assertEqual(np.count_nonzero(vd), len(vd))
+
+    def test_uniform_2_params(self):
+        r = ROOT.TRandom2(123)
+        vd = np.zeros(100, dtype=np.float64)
+        r.UniformN(100, vd, 5, 1.5)
+        self.assertEqual(np.count_nonzero(vd), len(vd))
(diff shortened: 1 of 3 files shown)


bindings/pyroot/pythonizations/test/trandom_array.py
```diff
@@ -0,0 +1,73 @@
...
+    def test_uniform(self):
+        r = ROOT.TRandom2(123)
+        vd = np.zeros(100, dtype=np.float64)
+        r.UniformN(100, vd, 5.0)
+        self.assertEqual(np.count_nonzero(vd), len(vd))
+
+    def test_uniform_2_params(self):
+        r = ROOT.TRandom2(123)
+        vd = np.zeros(100, dtype=np.float64)
+        r.UniformN(100, vd, 5, 1.5)
```
aaronj0, line 72: Looks good!  We can use something like:
   
   ```python

    def _TRandom_Uniform(self, ...):

        import ROOT
        import numpy
    
       #  if user expects a vector:
            return self.UniformN(...)
       # if matches original Uniform implementation
           return self._Uniform(...)
       
      @pythonization('TRandom')
    def pythonize_trandom(klass):
        klass._Uniform = klass.Uniform
        klass.Uniform = _TRandom_Uniform
   ```
   
   To automatically forward to the loop wrappers when the user expects an array of random numbers, or to the original generation function when the input signatures follow the traditional format

aaronj0 (changes requested):

vepadulano:
I believe this is a great improvement! Although I'm thinking about the function signature itself, wouldn't it be more natural for a Python user to get directly the array from the function?
```python
my_array = r.GausN(100, 0, 1)
```

vepadulano:
After further discussion with @aaronj0 , I understand a bit better the reasoning behind the current changes. I agree that for the Python user we can always add a further pythonization for `Gaus` and others that doesn't need the `N`. I still believe that from the purely C++ point of view, it would be nicer to have a function signature with a return value rather than an out parameter, and the return value should ideally be a `std::array<T, N>`. This can be pythonized to be converted to a numpy.array afterwards

dpiparo:
very good job. I was wondering: if we can avoid to template on the size of the return type, should we do this for example returning an `std::vector`? I see at least two shortcomings related to the `std::array`: 
- a template jitting when invoking the function
- exposing to potential neophytes 2 templates parameters (type and size) instead of 1 (type) or zero (double by default)

ericcano:
We now have numpy-like interface:

```python
>>> import ROOT as RT
>>> import numpy
>>> r = RT.TRandom(123)
>>> aa=numpy.zeros([2,3,4])
>>> r.Gaus(4, 0.4, out=aa)
>>> aa
array([[[3.96642121, 4.08632025, 4.26221681, 3.96809844],
        [3.67775501, 4.22637515, 3.46672227, 4.34481581],
        [4.12859111, 3.57025717, 3.84007792, 3.82163463]],

       [[4.26388937, 4.01280784, 4.35675554, 3.52723725],
        [3.32696688, 4.19441549, 4.28081704, 3.72076935],
        [3.59817384, 3.42698926, 4.69339572, 3.97866244]]])
>>> r.Gaus(4, 0.4)
5.311091735110202
>>> r.Gaus(4, 0.4, size=[2,3,4])
array([[[3.89629527, 4.01775263, 3.51479162, 4.05398529],
        [4.12382313, 3.65945926, 4.02825931, 4.85850178],
        [3.99641824, 3.70762586, 3.9692221 , 4.15951614]],

       [[3.66456567, 3.68396422, 3.7158251 , 3.90612771],
        [4.17368496, 3.63361176, 3.92615234, 3.99501832],
        [3.72500961, 3.75392728, 4.00470541, 3.96299244]]])
>>> r.Binomial(size=[2,3], ntot=25, prob=0.123)
array([[4, 2, 5],
       [1, 3, 2]], dtype=int32)
>>> r.BreitWigner(size=[2,3], mean=3, gamma=0.789)
array([[2.10769386, 3.37180071, 2.75851619],
       [3.08555421, 2.77677075, 2.7986828 ]])
>>> r.Exp(size=[2,3], tau=1000)
array([[ 255.33683037, 2115.41171346,  243.69722096],
       [1552.28132836,  350.13630267,  509.41190404]])
>>> r.Integer(size=[2,3], imax=10)
array([[7, 4, 3],
       [6, 1, 1]], dtype=uint32)
>>> r.Landau(size=[2,3], mean=5, sigma=2)
array([[ 8.61497464,  5.11830138,  3.93617892],
       [ 2.08862153,  4.43892651, 13.53280511]])
>>> r.Poisson(size=[2,3], mean=5)
array([[ 5,  7,  5],
       [ 7,  2, 10]], dtype=uint64)
>>> r.PoissonD(size=[2,3], mean=5)
array([[2., 7., 6.],
       [5., 7., 4.]])
>>> r.Uniform(size=[2,3], x2=5)
array([[2.1122357 , 0.22695246, 1.04537933],
       [1.36725521, 4.89698287, 0.21730294]])
>>> r.Uniform(size=[2,3], x2=-10, x1=5)
array([[-0.73856891, -5.70391524, -9.68524487],
       [-5.63854465,  1.3690705 , -0.07928077]])
```

ericcano:
Regarding the C++ side of this, could we imagine that instead of getting a pointer from a numpy array in the pythonization we get a richer class that wraps in a C++ interface the [numpy array iteration](https://numpy.org/doc/2.1/reference/c-api/iterator.html)? This way we could simply pack any RNG and its parameters in a lambda and simply fill any iterable container (`std::vector` as well) using std::generate with something like:
```C++
auto r = ROOT::TRandom(123)
auto v=std::vector<double>(1000); 
double mean=1, sigma=2.0;
std::generate(v.begin(), v.end(), [=,&r](){ return r.Gaus(mean, sigma)});
```

This can be easily templated, but template + virtual functions will not work though.
That would be efficient for in-place filling of any kind of numpy array without the current reshaping trick (that potentially implies copies)...

PR #17083, opened 2024-11-27
[RF] Deprecate RooDataSet constructors other than universal constructor

guitargeek:
The `RooDataSet` constructors to construct a dataset from a part of an
existing dataset are deprecated and will be removed in ROOT 6.38. This
is to avoid interface duplication. One should use `RooAbsData::reduce()`
instead, or to change the weight column, the universal constructor with
the `Import()`, `Cut()`, and `WeightVar()` arguments.

The tutorials and tests were adjusted accordingly.

Also, fix a wrong comment in the `rf403` tutorials: they said that the
existing dataset is changed to interpret a different column as the
weight, but actually a new dataset was created. The comment is fixed in
this commit.


lmoneta (approved):
LGTM!
I agree with this deprecation

PR #17085, opened 2024-11-27
[python] Avoid interfering with gc at exit time

vepadulano:
Yet again attempting this to see what happens in our CI. In some preliminary tests in a development environment built against the latest LCG release, this helps avoiding segfaults.

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/__init__.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/__init__.py
@@ -184,15 +184,4 @@ def cleanup():
         facade.__dict__["app"].keep_polling = False
         facade.__dict__["app"].process_root_events.join()
 
-    if "libROOTPythonizations" in sys.modules:
-
-        from ROOT import PyConfig
-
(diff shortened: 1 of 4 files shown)


bindings/pyroot/pythonizations/python/ROOT/__init__.py
```diff
@@ -184,15 +184,4 @@ def cleanup():
         facade.__dict__["app"].keep_polling = False
         facade.__dict__["app"].process_root_events.join()
 
-    if "libROOTPythonizations" in sys.modules:
```
guitargeek, line 187: Why do we even do this `if "libROOTPythonizations" in sys.modules` check now that we don't use `gROOT` from `libROOTPythonizations` anymore? Maybe always running the code branch helps...

vepadulano: Yeah that should not be needed anymore, but probably it's orthogonal. I will check if it makes a difference anyway

PR #17092, opened 2024-11-27
[RF] Implement `RooFit::DataError()` pythonization in C++

guitargeek:
This reduces divergence between the Python and C++ interface.

Tested by the tutorials where this string-to-enum pythonization is used.


guitargeek:
Failure on `mac14` unrelated:
```
 1280:tutorial-dataframe-distrdf002_dask_connection-py
```

aaronj0 (approved):
LGTM!

PR #17094, opened 2024-11-27
[RF] Forbid calling `getVal()` with r-value references to norm set

guitargeek:
Calling RooAbsReal::getVal() with r-value references to the normalization set (i.e. with RooArgSets that are created in place) is a bad idea, because it breaks RooFits caching logic and potentially introduces significant overhead.

There was even a dedicated Pythonization to deal with the related memory problems.

Since to avoid the problems and the need for a workaround in Python, this commit simply suggests to forbid calling getVal() with r-value references at runtime by printing a clear exception.


lmoneta (approved):
I agree with this improvement. 
Will not be a problem for lot's of code where users do: pdf.getVal(x) ?

guitargeek:
It will cause compile errors, but before it most likely caused runtime crashes or severe performance problems, because the integrals are cached by pointer to the normSet. Better compiler errors than the latter class of issues!

PR #17098, opened 2024-11-27
[tutorials] Rename tmva category in tutorials.

martamaja10:


mdessole (approved):

lmoneta (commented):
Was this rename really needed? 
Also the main directory name in the sources is 'tmva`. This I think might confuse users and also break some existing links in existing  external documentation of TMVA.

lmoneta:
@martamaja10 can you please revert the PR? I think we can change just the tag name in Doxygen for the documentation but keeping the same directory structure

PR #17106, opened 2024-11-27
[RDF] add HTML display

silverweed:
# This Pull request:
builds on https://github.com/root-project/root/pull/17081 and adds the option to print a RDisplay as a HTML table.
Additionally, it uses the HTML display by default as a repr for notebooks.

This is a sample output:
image

## Checklist:

- [x] tested changes locally


bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdisplay.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    Parameters:
+    klass: class to be pythonized
+    """
+    def repr(klass):
+        import ROOT
+        opts = ROOT.RDF.RDisplay.RPrintOptions()
+        opts.fFormat = ROOT.RDF.RDisplay.EPrintFormat.kHtml
+        return klass.AsString(opts)
+    klass._repr_html_ = repr
```
vepadulano, line 24: Maybe we can add for the non-IPython interactive use case
```
klass.__repr__ = lambda self: self.AsString()
```

which will show the markdown table which is friendlier for terminal-based interaction

tree/dataframe/src/RDFDisplay.cxx
```diff
@@ -218,8 +218,70 @@ std::string RDisplay::DashesBetweenLines(size_t lastColToPrint, bool allColumnsF
...
+      return AsStringMarkdown(considerDots);
+   case EPrintFormat::kHtml:
+      return AsStringHtml();
+   default:
+      R__ASSERT(false);
+   }
+   return {};
+}
+
+std::string RDisplay::AsStringHtml() const
```
vepadulano, line 245: Can you also add vertical separators between different columns of the table?

vepadulano (changes requested):
Thank you, this is great! I have a couple of minor comments to improve the already very nice changes. As a side thought, these changes also change the behaviour of `df.Display` in the case of interactive use, from lazy to instant, i.e. now the RDataFrame will run the Display action immediately to print the output table to screen. I believe that's more than valid for interactive use, but we should probably document this. Can you add a line to the release notes (for 6.36 at this point)? Then we should also add some doc about this in the docstrings that can be found in `_rdataframe.py`

vepadulano (approved):
Fantastic!!!

vepadulano:
I suggest squashing the commits and applying clang-format suggestions if it makes sense before merging

PR #17107, opened 2024-11-27
[skip-ci] Add documentation with short example with TF1.SetParameters from numpy array

skkwan:
# This Pull request:

Adds a short code snippet where a TF1 is initialized in the ROOT constructor, and parameters are set from a np.array (also tested with array.array) using SetParameters.

## Changes or fixes:

This PR addresses the [Hackathon to-do item "Explore setting multiple parameters for a fit at once"](https://github.com/orgs/root-project/projects/18?pane=issue&itemId=88467009).

## Checklist:

- [x] tested changes locally

This PR fixes #

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tf1.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tf1.py
@@ -26,7 +26,7 @@ my_func = ROOT.TF1("my_func", func, -10, 10, npar=2, ndim=2)
 \endcode
 
 Second, after performing the initialisation with a Python functor, the TF1 instance can be evaluated using the Pythonized
-`TF1::EvalPar` function. The pythonization allows passing in 1D(single set of x variables) or 2D(a dataset) NumPy arrays.
+TF1.EvalPar function. The pythonization allows passing in 1D(single set of x variables) or 2D(a dataset) NumPy arrays.
 
 The following example shows how we can create a TF1 instance with a Python function and evaluate it on a dataset:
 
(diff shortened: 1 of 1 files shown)


bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tf1.py
```diff
@@ -59,6 +59,22 @@ def pyf_tf1_coulomb(x, p):
 res = rtf1_coulomb.EvalPar(x[:, ::2], params)
 \endcode
 
+The below example defines a TF1 instance using the ROOT constructor, and sets its parameters using the Pythonized TF1::SetParameters function (i.e. without evaluating).
```
vepadulano, line 62:
```suggestion
The below example defines a TF1 instance using the ROOT constructor, and sets its parameters using the Pythonized TF1.SetParameters function (i.e. without evaluating).
```

vepadulano (changes requested):
Thank you! just a tiny change for consistency with the Python syntax

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tf1.py
skkwan: Thanks for the catch, I also changed a ``TF1::EvalPar`` to TF1.EvalPar a bit earlier in the docstring as well to use this syntax if that's OK.

vepadulano: Yes that's good, thanks!

PR #17108, opened 2024-11-27
[hist] Add `TH1::FillRandom` overload that takes TF1 directly

guitargeek:
This makes it possible to use `TH1::FillRandom` without relying on ROOTs global state (`gROOT->ListOfFunctions()`).

The new overload is made the `virtual` method that is overridded in the derived classes, because looking up the function by name is the same step no matter the number of dimensions.

Some Python tutorials are updated to test and show this new interface.


guitargeek:
Force-push to resolve conflicts

vepadulano (approved):
LGTM! Please apply clang-format suggestions before merging

guitargeek:
I'd rather not, because it breaks the current formatting standard:
```diff
    virtual Int_t    Fill(const char *name, Double_t w);
    virtual void     FillN(Int_t ntimes, const Double_t *x, const Double_t *w, Int_t stride=1);
    virtual void     FillN(Int_t, const Double_t *, const Double_t *, const Double_t *, Int_t) {}
-   virtual void     FillRandom(TF1 *f1, Int_t ntimes=5000, TRandom * rng = nullptr);
-           void     FillRandom(const char *fname, Int_t ntimes=5000, TRandom * rng = nullptr);
+   virtual void FillRandom(TF1 *f1, Int_t ntimes = 5000, TRandom *rng = nullptr);
+   void FillRandom(const char *fname, Int_t ntimes = 5000, TRandom *rng = nullptr);
    virtual void     FillRandom(TH1 *h, Int_t ntimes=5000, TRandom * rng = nullptr);
    virtual Int_t    FindBin(Double_t x, Double_t y=0, Double_t z=0);
    virtual Int_t    FindFixBin(Double_t x, Double_t y=0, Double_t z=0) const;
```

We should discuss at a later point what we do with the formatting of older classes

pcanal:
> I'd rather not, because it breaks the current formatting standard:

I agree :)

> We should discuss at a later point what we do with the formatting of older classes

We should pester clang-format to support the layout we have :), at least for the current header file content, it looks to me much more readable than what clang-format proposed.

PR #17112, opened 2024-11-27
Add NumPy style Histogram creator

aaronj0:
This PR fixes #


bindings/pyroot/pythonizations/python/ROOT/_hist.py
```diff
@@ -0,0 +1,92 @@
+import numpy as np
```
guitargeek, line 1: Please don't import numpy globally here. You need to import it locally in the functions, otherwise you force always loading these math libraries that numpy loads (see failing tests).

Python is very good at caching imported modules, so moving the import inside the function bodies won't impact the performance (just mentioning in case this was the reason you were making this import global).

siliataider:
Hi @aaronj0 , thank you for the PR! One point I would like to raise is that now with the Unified Histogram Interface (UHI), we can already directly fill a histogram with numpy data, adapting your example:
```python
n = 10000
x = np.linspace(1, 10, n)
y = 2*np.log(x) + np.random.rand(n)
hist2 = ROOT.TH2D("hist2", "2D uniform binning", 100, 1, 10, 100, np.min(y), np.max(y))
hist2[...] = np.histogram2d(x, y, bins=(100, 100))[0]
```
Given this functionality maybe we can discuss further:
- Is this a simpler interface than UHI?
- Are there features this PR adds that UHI doesn't support?
- Maybe the implementation can be adapted to avoid the [for loop with the call to Fill for every entry](https://github.com/root-project/root/pull/17112/files#diff-21aa88892086d99b41ea3ab7648d4d5bc2a5b4815a5d6fd7fad126fb9ea07db3R94) by computing the bin contents with numpy directly and leveraging the UHI implementation which directly manipulates the`fArray` array of data?
- ...

PR #17117, opened 2024-11-27
[tutorials] Categorise tutorials by labels (proof of concept)

vkucera:
Adds a support for labelling tutorial files with custom documentation comments (`\classes`, `\keywords`) and for generating an index page with links to tutorial files grouped by categories and labels.
tutorial-index-categories

### DIFF
--- a/documentation/doxygen/Makefile
+++ b/documentation/doxygen/Makefile
@@ -61,6 +61,7 @@ doxygen: filter pyzdoc
 	$(call MkDir,$(DOXYGEN_EXAMPLE_PATH))
 	$(call MkDir,$(DOXYGEN_NOTEBOOK_PATH))
 	./makehtmlfooter.sh > htmlfooter.html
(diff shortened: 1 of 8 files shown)


guitargeek:
Closing this. Not because the code doesn't work, but because the feature it enables has no input and never gained traction. Sorry about this, and please feel free to contribute to ROOT again in the future!

**Why close**

`categorise_tutorials.py` groups tutorials by parsing `\keywords` and `\classes` fields out of their Doxygen headers. On current master, of 1134 tutorial `.C` and `.py` files, **zero** carry either field. This PR added them to 4 files as a demonstration, and `index_categories.md` here is the 37-line index generated from exactly those 4. Rebased and merged as-is, the Makefile rule would generate an effectively empty page and wire it into the Doxygen INPUT.

Worth noting that in the two years this sat open (including a full tutorials reorganization where essentially every file was touched anyway) no `\keywords` field was added anywhere. So the categorisation scheme has no real buy-in, rather than it having been overlooked.

**If someone wants to revive this**

The order should be: agree on the keyword/class vocabulary, annotate the corpus, *then* write the generator. The ~90 lines of Python here are the cheapest part to reproduce and the most likely to need reshaping around whatever scheme is actually chosen, so there's little point rebasing it ahead of the other two steps. It's in the history here if useful as a starting point.

PR #17122, opened 2024-11-28
[webcanvas] resuse same JSONs in canv_batch.htm

linev:
When many images created in webcanvas batch mode, similar JSONs does not parsed twice but reused.

This minimize size of created temporary files and improve speed of images generation with stressGraphics - 
where same canvas stored several times with different image format.

Reduce memory usage when executing `canv_batch.htm` - only current painting is active, cleanup and remove HTML
element once painting is finished.

All together reduce time execution of `stressGraphics --web` from 50 to 25 seconds.


bellenot (approved):
Thanks Sergey!

PR #17123, opened 2024-11-28
[ntuple] Support bulk-reading with views

hahnjo:
Some improvements to the `*Range` classes, adding a test for collection views, and then adding `RNTupleViewBase<T>::CreateBulk()` to support bulk-reading, which is most useful with `RNTupleCollectionView`s.

### DIFF
--- a/tree/ntuple/v7/test/ntuple_view.cxx
+++ b/tree/ntuple/v7/test/ntuple_view.cxx
@@ -100,6 +100,28 @@ TEST(RNTuple, CollectionView)
       EXPECT_EQ(expected, viewJetsItems(index));
       expected++;
    }
+
+   // The same items can be bulk-read.
+   auto bulk = viewJetsItems.CreateBulk();
+   auto mask = std::make_unique<bool[]>(range.size());
+   std::fill(mask.get(), mask.get() + range.size(), true);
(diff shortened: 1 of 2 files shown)


tree/ntuple/v7/test/ntuple_view.cxx
```diff
@@ -57,6 +57,73 @@ TEST(RNTuple, View)
...
+   for (auto &&index : range) {
+      EXPECT_EQ(expected, viewJetsItems(index));
+      expected++;
+   }
+
+   // The same items can be bulk-read.
+   auto bulk = viewJetsItems.CreateBulk();
+   auto mask = std::make_unique<bool[]>(range.size());
+   std::fill(mask.get(), mask.get() + range.size(), true);
+   std::int32_t *values = static_cast<std::int32_t *>(bulk.ReadBulk(*range.begin(), mask.get(), range.size()));
```
hahnjo, line 108: From a user's perspective, this is a bit annoying. Would it be useful to provide a `ReadBulk(RNTupleClusterRange)` overload that reads all elements in the range?

jblomer: Possibly. The range classes then need to move into a different header because the field should not depend on the view.

PR #17124, opened 2024-11-28
update xroofit for next ROOT release

will-cern:
This is a latest version of the xRooFit API. I have also made a first attempt at adding a python test for the API, which also provides an example of how to build a simple workspace and run a CLs limit.


guitargeek (approved):
Thank you very much for the update!

guitargeek:
The test is a bit on the long side though, with 17s runtime:
```
2024-11-29T12:19:05.3578825Z  775/2631 Test  #375: pyunittests-roofit-xroofit-xroofit-python .........................................................   Passed   17.02 sec
```
@will-cern, in future, please try to keep the runtime of tests as low as possible. I know that sometimes it can't be avoided, but I still wanted to let you know that this is important to us (Imagine each of the ~3000 tests in ROOT taking half a minute).

PR #17130, opened 2024-11-28
[PyROOT] Disable memory regulator

guitargeek:
This feature arguably causes more harm than good.

The idea of the memory regulator was to fix user errors like this:

```python
ROOT.gInterpreter.Declare("""
void consume(TTree & t) { delete &t; }
""")

t = ROOT.TTree("tree", "tree")
ROOT.consume(t)
print(t)
```
Instead of a segfault because `t` was deleted, currently ROOT will
print:
```
None
```

Basically, it detected if any object registered in `gROOT` was deleted somewhere on the C++ side to notify CPyCppyy, which reacts by swapping a cppyy-specific "None" type into the Python proxy. However, such proxies are in a state that causes the garbage collector to segfault in specific environments, and this is not the only problem the memory regulator has. It was also the source of many poorly understood crashes in the past.

One other problem is that it only works for types that are registered in `gROOT`, like TTree, TFile, TH1, etc. If a user tries to access any other C++ class via Python after it was deleted, it will still segfault. So the memory regulator only sporadically solves the problem and results in an inconsistent user experience: the "ROOT-tracked" classes behave even more different than they already do.

Therefore, this commit suggests to disable the PyROOT memory regulator.

**Disadvantage of doing this:** correct object ownership needs to be taken a little bit more seriously in Python code that uses ROOT (see required changes to the tutorials and pythonizations in this PR).


bindings/pyroot/pythonizations/python/ROOT/_pythonization/_th1.py
```diff
@@ -213,6 +213,45 @@ def _FillWithNumpyArray(self, *args):
...
+def pythonize_th1(klass):
+    klass._cpp_constructor = klass.__init__
+    klass.__init__ = _TH1_Constructor
+
+@pythonization('TH1F')
+def pythonize_th1(klass):
+    klass._cpp_constructor = klass.__init__
+    klass.__init__ = _TH1_Constructor
+
+@pythonization('THDF')
```
guitargeek, line 240:
```suggestion
@pythonization('TH2D')
```
Typo will be fixed when the PR is updated the next time.

dpiparo: thx

dpiparo (approved):
LGTM, thanks for trying to solve this very hard problem.

Issue #17132, opened 2024-11-29
Make `RNTupleProcessor` fully composable


enirolf:
### Explain what you would like to see improved and how.

This issue is meant to collect PRs working towards making the `RNTupleProcessor` fully composable. Here "composable" means that each type of processor can be create from other processors (e.g., an `RNTupleChainProcessor` can be created from several `RNTupleJoinProcessors` or vice-versa). Since this will involve significant changes and additions to the existing `RNTupleProcessor` code, it seems sensible to split this effort over multiple PRs to hopefully make it a bit easier to review and keep track of the changes.

Three main steps are foreseen:
- [ ] Add an `RNTupleBaseProcessor` for processing single RNTuples, using the same interface as `RNTuple[Chain|Join]Processor`.
- [ ] Refactor `RNTupleChainProcessor` to use `RNTuple[Base|Chain|Join]Processor` objects instead of single RNTuples directly.
- [ ] Refactor `RNTupleJoinProcessor` to use `RNTuple[Base|Chain|Join]Processor` objects instead of single RNTuples directly.

Additional code and test cleanups will also be tracked here.

### ROOT version

N/A

### Installation method

N/A

### Operating system

N/A

PR #17134, opened 2024-11-29
[ntuple] Add `RNTupleBaseProcessor`

enirolf:
The `RNTupleBaseProcessor` processes one single RNTuple, using the same interfaces as the `RNTupleChainProcessor` and `RNTupleJoinProcessor`. The point of adding a base processor is that this way, the other processors can use a unified interface for loading entries, without having to know whether they are dealing with a single RNTuple or a composed one.

This PR is part of a bigger set of (foreseen) changes, collected and tracked in #17132.


tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -105,6 +120,54 @@ void ROOT::Experimental::RNTupleProcessor::ConnectField(RFieldContext &fieldCont
...
+ROOT::Experimental::RNTupleBaseProcessor::RNTupleBaseProcessor(const RNTupleOpenSpec &ntuple, RNTupleModel &model)
+   : RNTupleProcessor({ntuple})
+{
+   fPageSource = Internal::RPageSource::Create(ntuple.fNTupleName, ntuple.fStorage);
+   fPageSource->Attach();
+
+   if (fPageSource->GetNEntries() == 0) {
+      throw RException(R__FAIL("specified RNTuple \"" + ntuple.fNTupleName + "\" at path \"" + ntuple.fStorage +
+                               "\" does not contain any entries"));
+   }
```
vepadulano, line 132: For the purposes of this PR I believe we can be restrictive and error out in case the input RNTuple is empty. In general, I would add a bit of information from experience here. With TTree-based processing, cases such as this are possible and fairly common. For example in distributed RDataFrame we create no-op tasks in case we detect an empty TTree in the current task. For RNTuple in general probably now this case is not available or not thoroughly designed, but I would suggest to add a comment here to remember this in the future

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -256,6 +260,30 @@ public:
...
 };
 
+// clang-format off
+/**
+\class ROOT::Experimental::RNTupleBaseProcessor
+\ingroup NTuple
+\brief Processor specializiation for processing a single RNTuple.
+*/
+// clang-format on
+class RNTupleBaseProcessor : public RNTupleProcessor {
```
vepadulano, line 270: imho the keyword `Base` sounds more like `RNTupleProcessor` should derive from this class and not the opposite. Just as a thought, we might want to consider alternative namings, for example:

1. One way would be two switch the two class names, the base class would be called `RNTupleBaseProcessor` and the new implementation in this PR would be the `RNTupleProcessor`.
2. Or maybe we could think about some other keyword for this one, such as `Unit`, `Single`, `Lone`, `One`

vepadulano (approved):
Very nice! I have a couple of minor remarks, but the code looks good already!

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
enirolf: This is a good point. I am not particularly in favor of swapping the two names, because (what is currently) the `RNTupleProcessor` is the main interface users interact with so IMO this name should be preserved like that. I think `Unit` is a nice alternative, though. I will change it to this for now, and then we can come back and have another critical look at the naming once the interface has stabilized more.

enirolf:
Indeed, this processor should also be able to handle empty ntuples. I have an implementation for this, but it required a bit more refactoring than I initially forsaw, so I propose to merge this PR as-is, and I will open a new one to address the empty ntuple case.

tree/ntuple/v7/src/RNTupleProcessor.cxx
pcanal: See  https://github.com/root-project/root/pull/17134#issuecomment-2538516397

PR #17137, opened 2024-11-29
[ntuple] Add a join processing tutorial

enirolf:
This PR adds a tutorial for using the `RNTupleProcessor` to join two RNTuples. Once RNTupleProcessors are fully composable (see #17132), we might consider merging it with `ntpl012_processor_chain.C`.

### DIFF
--- a/tutorials/v7/ntuple/ntpl012_processor_chain.C
+++ b/tutorials/v7/ntuple/ntpl012_processor_chain.C
@@ -1,7 +1,7 @@
 /// \file
 /// \ingroup tutorial_ntuple
 /// \notebook
-/// Demonstrate the RNTupleProcessor using multiple RNTuples
+/// Demonstrate the RNTupleProcessor for vertical compositions (chains) of RNTuples
 ///
 /// \macro_image
 /// \macro_code
@@ -93,7 +93,7 @@ void Read(const std::vector<RNTupleOpenSpec> &ntuples)
    hPx.DrawCopy();
 }
 
-void ntpl012_processor()
+void ntpl012_processor_chain()
 {
    // The ntuples to generate and subsequently process. The model of the first ntuple will be used to construct the
    // entry used by the processor.
--- a/tutorials/v7/ntuple/ntpl015_processor_join.C
+++ b/tutorials/v7/ntuple/ntpl015_processor_join.C
@@ -1,0 +1,113 @@
+/// \file
+/// \ingroup tutorial_ntuple
+/// \notebook
+/// Demonstrate the RNTupleProcessor for horizontal compositions (joins) of RNTuples
+///
+/// \macro_image
+/// \macro_code
+///
+/// \date November 2024
+/// \author The ROOT Team
+
+// NOTE: The RNTuple classes are experimental at this point.
+// Functionality and interface are still subject to changes.
+
+#include <ROOT/RNTupleModel.hxx>
+#include <ROOT/RNTupleWriter.hxx>
+#include <ROOT/RNTupleProcessor.hxx>
+
+#include <TCanvas.h>
+#include <TH1F.h>
+#include <TRandom.h>
+
+// Import classes from the `Experimental` namespace for the time being.
+using ROOT::Experimental::RNTupleModel;
+using ROOT::Experimental::RNTupleOpenSpec;
+using ROOT::Experimental::RNTupleProcessor;
+using ROOT::Experimental::RNTupleWriter;
+
+const std::string kMainNTupleName = "mainNTuple";
+const std::string kMainNTuplePath = "main_ntuple.root";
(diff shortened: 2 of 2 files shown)


enirolf:
The failing clang-format is not actually a formatting failure (not sure why the job failed..). I ran clang-format locally before pushing, so that should all be ok :woman_shrugging:.

tutorials/v7/ntuple/ntpl012_processor_chain.C
```diff
@@ -1,7 +1,7 @@
 /// \file
 /// \ingroup tutorial_ntuple
 /// \notebook
-/// Demonstrate the RNTupleProcessor using multiple RNTuples
+/// Demonstrate the RNTupleProcessor for chaining RNTuples
```
vepadulano, line 4:
```suggestion
/// Demonstrate the RNTupleProcessor for vertical concatenations (chains) of RNTuples
```

tutorials/v7/ntuple/ntpl015_processor_join.C
```diff
@@ -0,0 +1,112 @@
+/// \file
+/// \ingroup tutorial_ntuple
+/// \notebook
+/// Demonstrate the RNTupleProcessor for joining RNTuples
```
vepadulano, line 4:
```suggestion
/// Demonstrate the RNTupleProcessor for horizontal concatenations (joins) of RNTuples
```

tutorials/v7/ntuple/ntpl015_processor_join.C
```diff
@@ -0,0 +1,112 @@
...
+///
+/// \macro_image
+/// \macro_code
+///
+/// \date November 2024
+/// \author The ROOT Team
```
vepadulano, line 10: I would say your name is better suited here

tutorials/v7/ntuple/ntpl015_processor_join.C
```diff
@@ -0,0 +1,112 @@
...
+{
+   auto model = RNTupleModel::Create();
+
+   auto fldI = model->MakeField<std::uint32_t>("i");
+   auto fldVpx = model->MakeField<float>("vpx");
+
+   auto ntuple = RNTupleWriter::Recreate(std::move(model), ntupleName, ntupleFileName);
+
+   // The main ntuple only contains a subset of the entries present in the auxiliary ntuple.
+   for (int i = 0; i < kNEvents; i += gRandom->Integer(5)) {
```
vepadulano, line 46: I would say we don't want to have a random number of entries in a tutorial?

tutorials/v7/ntuple/ntpl015_processor_join.C
```diff
@@ -0,0 +1,112 @@
...
+   // field values is used. It is possible to specify up to 4 join fields. Providing an empty list of join fields
+   // signals to the processor that all entries are aligned.
+   auto processor = RNTupleProcessor::CreateJoin(ntuples, {"i"});
+
+   float px, py;
+   for (const auto &entry : *processor) {
+      // Fields from the main ntuple are accessed by their original name.
+      px = *entry.GetPtr<float>("vpx");
+      // Fields from auxiliary ntuples are accessed by prepending the name of the auxiliary ntuple.
+      py = *entry.GetPtr<float>(kAuxNTupleName + ".vpy");
```
vepadulano, line 96: But in this case it could also be accessed without the auxiliary name?

tutorials/v7/ntuple/ntpl015_processor_join.C
```diff
@@ -0,0 +1,112 @@
...
+
+      hPy.Fill(px + py);
+   }
+
+   std::cout << "Processed at total of " << processor->GetNEntriesProcessed() << " entries" << std::endl;
```
vepadulano, line 101:
```suggestion
   std::cout << "Processed a total of " << processor->GetNEntriesProcessed() << " entries" << std::endl;
```

vepadulano (changes requested):
Nice thanks! Some minor comments from my side

tutorials/v7/ntuple/ntpl015_processor_join.C
enirolf: All other tutorials also have "The ROOT team", in this case I would prefer to keep this consistency.

tutorials/v7/ntuple/ntpl015_processor_join.C
enirolf: No; in the current implementation fields in auxiliary fields always have to be prefixed even if they are unique. Apart from the fact that it would be non-trivial to change this I'm also not sure if we would even want that, I think it would only lead to unwanted ambiguities.

vepadulano (approved):
Thank you!

PR #17139, opened 2024-11-29
[ntuple] add streaming vector tutorial

jblomer:
Demonstrates how to use the RNTuple API to build a special purpose container that reads large vectors piece-wise.

Original idea & code from @gemmeren 

TODO: there are a few shortcomings in the current view API that should be addressed before merging.


vepadulano (commented):
This is a very nice addition thanks! A few considerations from my side

tutorials/v7/ntuple/ntpl015_streaming_vector.C
```diff
@@ -0,0 +1,161 @@
...
+   RNTupleCollectionView fVectorView;
+   RNTupleView<T> fItemView;
+   RNTupleClusterRange fRange;
+   NTupleSize_t fEntry{0};
+   NTupleSize_t fSize{0};
+
+public:
+   class iterator {
+      RNTupleClusterRange::RIterator fIndex;
+      RNTupleView<T> &fView;
```
hahnjo, line 93: should store a pointer to make the type copyable

silverweed: Should this be copyable though? To me the semantics would be a bit fuzzy (would the copy automatically get the same window of entries loaded in? Also, when you copy a std::vector you do a deep copy, but in this case to my understanding it wouldn't)

pcanal: Doesn't that hint that the class is more of a `RStreamingVectorView` and/or `RVectorStreamingView`?

jblomer:
Replaced by #19748

PR #17143, opened 2024-11-29
[skip-ci] Add a note in reference about the coordinate system negation pitfalls

m-fila:
# This Pull request:

Adds a short note about an unintuitive behavior of 4-vector negation for coordinate systems based on mass instead of energy (#15842)

This the same note as in #17049 but this time it will also appear in the Reference Documentation

image

## Checklist:

- [x] tested changes locally


vepadulano (approved):
LGTM

PR #17147, opened 2024-11-29
[TMVA] Fix warning on x86

bellenot:
Fix the following warning on 32 bit:
```
TMVA/ROperator_Reshape.hxx(82,74): warning GECC553F5: result of comparison of constant 9223372036854775807 with expression of type 'size_t' (aka 'unsigned int') is always false [-Wtautological-constant-out-of-range-compare]
            if ((output_length == 0 && fAllowZero == 0) || output_length > INT64_MAX) {
```

### DIFF
--- a/tmva/sofie/inc/TMVA/ROperator_Reshape.hxx
+++ b/tmva/sofie/inc/TMVA/ROperator_Reshape.hxx
@@ -79,7 +79,7 @@ public:
          size_t output_length = ConvertShapeToLength(output_shape);
          // (input_length == output_length) is the easy case : (2,3,4) -> (2,12)
          if (input_length != output_length) {
-            if ((output_length == 0 && fAllowZero == 0) || output_length > SIZE_MAX) {
+            if ((output_length == 0 && fAllowZero == 0) || static_cast<long>(output_length)  < 0) {
                // in this case value 0 or -1 in shape are automatically corrected
                bool replacementDone = false;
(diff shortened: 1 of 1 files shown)


lmoneta (changes requested):
This is not correct!
The idea here is that you get value larger than signed int because one of the inputs is negative. 
This line should  work if the sizes 
```
if ((output_length == 0 && fAllowZero == 0) || static_cast<long>(output_length)  < 0) {
```
 
since long is automatically 32 bits on 32 bit machines and 64 bits on 64 machines. In any case we don't expect to have arrays of size larger than 32 bits

guitargeek:
Thanks for saving the day Lorenzo, indeed I didn't review the PR carefully, since it's only one letter away from what I thought was the good solution (checking against the maximum of the corresponding signed type):
```c++
output_length > SSIZE_MAX
```

bellenot:
@lmoneta thanks for the clarification! But note that `long` is always 32 bit on Windows (both x64 and x86)... Is your test still valid in both cases?

lmoneta:
Thanks for pointing this out. It should still be fine as far the size of the array in question is not larger than a 32-bit integer.

bellenot:
@lmoneta I applied your suggestion. If you're sure it will work the way you want, feel free to approve this PR. Thanks!

lmoneta (approved):
LGTM!
Thank you Bertrand for fixing this warning!

PR #17148, opened 2024-11-29
Try to enable webgui tests again

linev:
1. webgui-ping now uses posix-spawn and must cleanup process correctly
2. stressGraphics --web is now optimized to use much less resources 
3. Disable 3D tests with firefox - one some platforms WebGL emulation does not work correctly in Firefox
4. Ensure normal tests runs without web graphics


couet (approved):
Seems ok, despite I cannot really judge if it will create problem with the CI or not.

linev:
>  despite I cannot really judge if it will create problem with the CI 

@couet 

Currently there are no browsers installed on problematic nodes.
Maybe we should try add chrome again to these nodes - to see if CI will complain or not

PR #17149, opened 2024-11-29
[Minuit2] Remove deleted header file reference

samford:
# This Pull request:

## Changes or fixes:

`math/minuit2/inc/Minuit2/FunctionMinimizer.h` was removed in 91bb4d7 (https://github.com/root-project/root/pull/16443) but there is still a reference to `FunctionMinimizer.h` in `math/minuit2/src/CMakeLists.txt` and this is causing the Minuit2 build to fail (e.g., https://github.com/Homebrew/homebrew-core/pull/199470#pullrequestreview-2470163494).

## Checklist:

- [x] tested changes locally

### DIFF
--- a/math/minuit2/src/CMakeLists.txt
+++ b/math/minuit2/src/CMakeLists.txt
@@ -33,7 +33,6 @@ set(MINUIT2_HEADERS
     FumiliStandardChi2FCN.h
     FumiliStandardMaximumLikelihoodFCN.h
     FunctionGradient.h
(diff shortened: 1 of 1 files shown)


guitargeek (approved):
Excellent, thank you so much for the fix!

samford:
Glad to help and thanks for the quick review. I've updated the aforementioned Homebrew PR to use this patch for the time being (removing it with the next version), so that will allow us to update to 6.34.00 in the interim time.

PR #17150, opened 2024-11-29
[v634][Minuit2] Remove deleted header file reference

guitargeek:
Backport of #17149.


guitargeek:
Should have indicated "skip-ci". We're not testing standalone Minuit2 compilation.

Issue #17145, opened 2024-11-29
Distributed RDataFrame cannot deal with same column name in different branches


vepadulano:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

In distributed RDataFrame, using the same column name in different branches leads to the following type of error

```
  File "/home/vpadulan/programs/rootproject/rootbuild/distrdf-debug/lib/DistRDF/Proxy.py", line 275, in get_proxy_for
    _update_internal_df_with_transformation(node, operation)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/vpadulan/programs/rootproject/rootbuild/distrdf-debug/lib/DistRDF/Proxy.py", line 63, in _update_internal_df_with_transformation
    node.get_head()._localdf = rdf_operation(*operation.args, **operation.kwargs)
                               ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vpadulan/programs/rootproject/rootbuild/distrdf-debug/lib/ROOT/_pythonization/_rdf_pyz.py", line 394, in _PyDefine
    return rdf._OriginalDefine(col_name, callable_or_str)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Template method resolution failed:
  ROOT::RDF::RInterface<ROOT::Detail::RDF::RLoopManager,void> ROOT::RDF::RInterface<ROOT::Detail::RDF::RLoopManager,void>::Define(string_view name, string_view expression) =>
    runtime_error: RDataFrame::Define: cannot define column "a". A column with that name has already been Define'd. Use Redefine to force redefinition.
  Failed to instantiate "Define(ROOT::RDF::RInterface<ROOT::Detail::RDF::RLoopManager,void>&,std::string,std::string)"
  Failed to instantiate "Define(ROOT::RDF::RInterface<ROOT::Detail::RDF::RLoopManager,void>*,std::string,std::string)"
  Failed to instantiate "Define(ROOT::RDF::RInterface<ROOT::Detail::RDF::RLoopManager,void>,std::string,std::string)"
```

@gpetruc FYI

### Reproducer

```python
from dask.distributed import Client, LocalCluster

import ROOT

RDataFrame = ROOT.RDF.Experimental.Distributed.Dask.RDataFrame


[...]

PR #17153, opened 2024-11-30
[634] Distributed RDataFrame cannot deal with same column name in different branches

dpiparo:
BP of https://github.com/root-project/root/issues/17145


vepadulano (approved):
Thank you!

PR #17160, opened 2024-12-01
[cling] fix error handling when parsing invaild code

Vipul-Cariappa:
`llvm::Error` needs to be consumed before destruction.


Vipul-Cariappa:
ping @vgvassilev

interpreter/cling/lib/Interpreter/IncrementalParser.cpp
```diff
@@ -911,8 +911,11 @@ namespace cling {
     PP.EnterSourceFile(FID, /*DirLookup*/nullptr, NewLoc);
     m_Consumer->getTransaction()->setBufferFID(FID);
 
-    if (!ParseOrWrapTopLevelDecl())
+    llvm::Expected<bool> res = ParseOrWrapTopLevelDecl();
```
vgvassilev, line 914: We should update the `ParseOrWrapTopLevelDecl` to return `llvm::Error` instead of `llvm::Expected<bool>`.

Vipul-Cariappa: Make the changes suggested. Please have a look.

vgvassilev (approved):
LGTM!

PR #17161, opened 2024-12-01
[PyROOT] Whitelist `libgcc_s` library in imported library test

guitargeek:
On some platforms, the low-level runtime library provided by GCC will also be loaded when importing ROOT. Therefore, it needs to be added to the whitelist such that the ROOT Python module tests don't fail because of it.

Fixes a test failure that I saw on my workstation with NixOS.


dpiparo (approved):
Thanks for fixing ROOT tests on Nix!

PR #17164, opened 2024-12-02
[RF] Clarify ownership of underlying TTree for RooDataSet

guitargeek:
Commit b0d96bf3 from 5 years ago invalidated some part of the documentation which needs to be updated accordingly: the TTrees underlying the RooTreeDataStore are not attached to any TFile anymore, therefore we should not write about how the tree-backed store could help to reduce RAM usage.

Inspired by this forum post:
https://root-forum.cern.ch/t/large-tree-based-roodataset/62366/4


lmoneta (approved):
LGTM!
Thank you Jonas for this doc fix!

PR #17165, opened 2024-12-02
[RF] Fix some design issues with the RooTreeDataStore

guitargeek:
1. The RooAbsDataStore has a virtual function `tree()` which is not meaningful for a general data store. This function should be removed.

2. The RooTreeDataStore has one `const` version on `tree()` and one non-const version. One returns a pointer and the other a reference, which is very confusing.

3. The RooTreeDataStore has some methods that just forward to the underlying TTree, but they were not used.

Changes to the data store classes can be made without impacting users because they are implementation details of the RooFit dataset classes.

### DIFF
--- a/roofit/roofitcore/inc/RooAbsDataStore.h
+++ b/roofit/roofitcore/inc/RooAbsDataStore.h
@@ -138,7 +138,6 @@ public:
 
   virtual bool hasFilledCache() const { return false ; }
 
-  virtual const TTree* tree() const { return nullptr ; }
   virtual void dump() {}
 
   virtual void loadValues(const RooAbsDataStore *tds, const RooFormulaVar* select=nullptr, const char* rangeName=nullptr,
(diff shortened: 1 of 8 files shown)


lmoneta (commented):
The RooAbsDataStore::addColumns is removed. What is the reason for this?

guitargeek:
> The RooAbsDataStore::addColumns is removed. What is the reason for this?

Dangerous interface:
```c++
RooArgSet* RooDataSet::addColumns(const RooArgList& varList)
{
  auto * holderSet = new RooArgSet{};
  for(RooAbsArg * var : varList) {
    holderSet->add(*addColumn(*var));
  }
  return holderSet;
}
```

How should the user guess that `addColumns()` is returning an owning pointer with a set? It's just too easy to get memory leaks with this function. To avoid this, one can either change the signature or remove the function all together. Both is backwards incompatible. But in this case, I think removal is the better solution because `addColumns` doesn't add much convenience compared to calling `addColumn` in a loop.

lmoneta (approved):
Thank you Jons for fixing these issues!
LGTM!

PR #17170, opened 2024-12-02
[RF] Fix creating yield histograms from extended pdfs

guitargeek:
When using the `Extended()` keyword argument in `createHistogram()` the
[docs](https://root.cern.ch/doc/master/classRooAbsReal.html#a9451168bb4159899fe1854f591f69814) say it has the following effect:

> Plot event yield instead of probability density (for extended pdfs only)

However, to get yields out of the probability densities, one has to not
only multiply with the total number of expected events, but also with
the bin volumes. I don't understand why the bin volume multiplication is
disable in *exactly* this case. This commit suggests to not do that
anymore, in order to be able to correctly produce predicted yields
histograms.

Thanks to the following forum post for noticing this:
https://root-forum.cern.ch/t/bin-content-of-pdf-and-generated-data/62370

This change of behavior is also mentioned in the release notes.

A unit test to validate that the yield histogram can now be correctly
created is also added.

### DIFF
--- a/README/ReleaseNotes/v636/index.md
+++ b/README/ReleaseNotes/v636/index.md
@@ -56,6 +56,14 @@ The following people have contributed to this new version:
     This pattern was not appropriate for a modern C++ library.
     If you absolutely need the old return type, wrap the call to `format()` inside `new TString{var.format(..)}`. However, this is not recommended.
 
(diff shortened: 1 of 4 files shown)


lmoneta (approved):
LGTM!
Thank you Jonas for fixing this!

PR #17173, opened 2024-12-03
[RDF] preserve order of messages in RDisplay::Print()

silverweed:
#17081 changed the order or printing in `RDisplay::Print()` so that all warnings are now printed before the actual message. This PR restores the original order of printing.


vepadulano:
My onest opinion: the previous behaviour was wrong and I sincerely believe that printing all warnings before the actual table just looks better and is less surprising for the user

PR #17176, opened 2024-12-03
[ntuple] `RMiniFile` improvements and preparations for parallel writing

hahnjo:
* Fix directory of `RBlob`s from `RFileProper`
 * Simplify methods and implementations for reserving / writing `RBlob` keys
 * Allow reserving blobs without writing key

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
@@ -155,6 +155,8 @@ private:
       std::uint64_t WriteKey(const void *buffer, std::size_t nbytes, std::size_t len, std::int64_t offset = -1,
                              std::uint64_t directoryOffset = 100, const std::string &className = "",
                              const std::string &objectName = "", const std::string &title = "");
+      /// Writes an RBlob opaque key with the provided buffer as data record and returns the offset of the record
+      std::uint64_t WriteBlobKey(const void *buffer, std::size_t nbytes, std::size_t len);
       operator bool() const { return fFile; }
    };
 
--- a/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
@@ -107,7 +107,7 @@ private:
       /// Low-level writing using a TFile
       void Write(const void *buffer, size_t nbytes, std::int64_t offset);
       /// Writes an RBlob opaque key with the provided buffer as data record and returns the offset of the record
-      std::uint64_t WriteKey(const void *buffer, size_t nbytes, size_t len);
+      std::uint64_t WriteBlobKey(const void *buffer, size_t nbytes, size_t len);
       operator bool() const { return fFile; }
    };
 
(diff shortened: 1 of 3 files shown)


tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
```diff
@@ -224,8 +228,9 @@ public:
    std::uint64_t WriteNTupleFooter(const void *data, size_t nbytes, size_t lenFooter);
    /// Writes a new record as an RBlob key into the file
    std::uint64_t WriteBlob(const void *data, size_t nbytes, size_t len);
-   /// Reserves a new record as an RBlob key in the file.
-   std::uint64_t ReserveBlob(size_t nbytes, size_t len);
+   /// Reserves a new record as an RBlob key in the file. If keyBuffer is specified, it must be written *before* the
+   /// returned offset. (Note that the array type is purely documentation, the argument is actually just a pointer.)
+   std::uint64_t ReserveBlob(size_t nbytes, size_t len, unsigned char keyBuffer[kBlobKeyLen] = nullptr);
```
hahnjo, line 233: This is an interface extension that is needed to implement parallel writing of the `RBlob` keys. I'm starting to experiment with this for distributed writing and I believe it's what we need, but I can also keep it out of this PR for now, just let me know.

tree/ntuple/v7/src/RMiniFile.cxx
```diff
@@ -828,6 +828,33 @@ void ROOT::Experimental::Internal::RMiniFileReader::ReadBuffer(void *buffer, siz
 
 ////////////////////////////////////////////////////////////////////////////////
 
+auto &kBlobKeyLen = ROOT::Experimental::Internal::RNTupleFileWriter::kBlobKeyLen;
```
silverweed, line 831: Why is this an `auto &`?

tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
jblomer: I think it's fine. Anyway internal API.

jblomer (approved):
LGTM

tree/ntuple/v7/src/RMiniFile.cxx
hahnjo: IIRC this was what the compiler suggested. Do we want to have this a `static constexpr auto kBlobKeyLen =`?

hahnjo: I changed the declaration accordingly, as discussed

pcanal: The new version seems to have replace the alias with a hard coded number.  Does it matter if they (inadvertently) get out of sync?

hahnjo: It's not clear to me what you are referring to: the new version is
```c++
static constexpr auto kBlobKeyLen = ROOT::Experimental::Internal::RNTupleFileWriter::kBlobKeyLen;
```
which itself is defined as
```c++
static constexpr std::size_t kBlobKeyLen = 42;
```
So the variables are `constexpr` and can never change its value. The value 42 itself is checked with two `R__ASSERT` in `PrepareBlobKey`.

pcanal: My bad. I kept not find the `static constexpr auto kBlobKeyLen` in the diff and mistakenly assumed that `kBlobKeyLen = 42` was new and that somewhere else there was another (pre-existing) `kBlobKeyLen = 42`.

So nevermind, this is resolved.

PR #17181, opened 2024-12-03
[skip-ci][meta] Update the doc of TClass::New focussing on the IO ctors

dpiparo:


core/meta/src/TClass.cxx
```diff
@@ -5010,33 +5010,44 @@ const void *TClass::DynamicCast(const TClass *cl, const void *obj, Bool_t up)
...
 /// Return a pointer to a newly allocated object of this class.
-/// The class must have a default constructor. For meaning of
-/// defConstructor, see TClass::IsCallingNew().
 ///
 /// If quiet is true, do no issue a message via Error on case
 /// of problems, just return 0.
 ///
-/// The constructor actually called here can be customized by
-/// using the rootcint pragma:
+/// This method is also used by the I/O subsystem to allocate the rigth memory
```
fwyzard, line 5017:
```suggestion
/// This method is also used by the I/O subsystem to allocate the right memory 
```

core/meta/src/TClass.cxx
```diff
@@ -5010,33 +5010,44 @@ const void *TClass::DynamicCast(const TClass *cl, const void *obj, Bool_t up)
...
-/// defConstructor, see TClass::IsCallingNew().
 ///
 /// If quiet is true, do no issue a message via Error on case
 /// of problems, just return 0.
 ///
-/// The constructor actually called here can be customized by
-/// using the rootcint pragma:
+/// This method is also used by the I/O subsystem to allocate the rigth memory 
+/// for the objects. If a default constructor is not defined for a certain
+/// class is not defined, some options are available.
```
fwyzard, line 5019:
```suggestion
/// for the objects. If the default constructor is not defined for a certain
/// class, some options are available.
```

core/meta/src/TClass.cxx
```diff
@@ -5010,33 +5010,44 @@ const void *TClass::DynamicCast(const TClass *cl, const void *obj, Bool_t up)
 
 ////////////////////////////////////////////////////////////////////////////////
 /// Return a pointer to a newly allocated object of this class.
-/// The class must have a default constructor. For meaning of
-/// defConstructor, see TClass::IsCallingNew().
 ///
 /// If quiet is true, do no issue a message via Error on case
 /// of problems, just return 0.
```
pcanal, line 5015: < /// of problems, just return 0.
> /// of problems, just return nullptr.

core/meta/src/TClass.cxx
```diff
@@ -5010,33 +5010,44 @@ const void *TClass::DynamicCast(const TClass *cl, const void *obj, Bool_t up)
...
 ///
-/// The constructor actually called here can be customized by
-/// using the rootcint pragma:
+/// This method is also used by the I/O subsystem to allocate the right amount 
+/// of memory for the objects. If a default constructor is not defined for a 
+/// certain class, some options are available.
+/// The simplest is to define an I/O constructor, for example
```
pcanal, line 5020:
```suggestion
/// The simplest is to define the default I/O constructor, for example
```

core/meta/src/TClass.cxx
```diff
@@ -5010,33 +5010,44 @@ const void *TClass::DynamicCast(const TClass *cl, const void *obj, Bool_t up)
...
 /// ~~~ {.cpp}
 ///    MyClass(UserClass*);
 ///    MyClass(TRootIOCtor*);
 ///    MyClass(); // Or a constructor with all its arguments defaulted.
 /// ~~~
 ///
-/// When more than one pragma ioctortype is used, the first seen as priority
-/// For example with:
+/// More than one `pragma ioctortype` is used, they are considered in 
+/// order of priority. For example with:
```
pcanal, line 5045:
```suggestion
/// When more than one `pragma ioctortype` is specified, the priority order is
/// defined as the definition order; the earliest definitions have higher priority.
/// For example with:
```

core/meta/src/TClass.cxx
```diff
@@ -5010,33 +5010,44 @@ const void *TClass::DynamicCast(const TClass *cl, const void *obj, Bool_t up)
...
+/// using a rootcling pragma as follows:
 /// ~~~ {.cpp}
 ///    #pragma link C++ ioctortype UserClass;
 /// ~~~
-/// For example, with this pragma and a class named MyClass,
-/// this method will called the first of the following 3
-/// constructors which exists and is public:
+/// If that is done, a constructor with an argument of type `UserClass*` 
+/// is specified. The effect of that syntax in the Linkdef file is that
+/// these constructors are looked for by ROOT, in this order:
```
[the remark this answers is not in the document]
pcanal: As a relevant aside, when looking at the code it looks like the order might actually be 
```
///    MyClass(TRootIOCtor*);
///    MyClass(UserClass*);
///    MyClass(); // Or a constructor with all its arguments defaulted.
```
but hopefully I misread the code.  One would need to read and/or expand the test `roottest/root/io/alloc:customConstructor`

dpiparo:
@pcanal, all comments addressed.

core/meta/src/TClass.cxx
```diff
@@ -5010,33 +5010,45 @@ const void *TClass::DynamicCast(const TClass *cl, const void *obj, Bool_t up)
...
+/// Moreover, the constructor called by TClass::New can be customized by
+/// using a rootcling pragma as follows:
 /// ~~~ {.cpp}
 ///    #pragma link C++ ioctortype UserClass;
 /// ~~~
-/// For example, with this pragma and a class named MyClass,
-/// this method will called the first of the following 3
-/// constructors which exists and is public:
+/// `TClass::New` will then look for a constructor (for a class `MyClass` in the 
+/// following example) in this order, constructing the object using the first one
```
pcanal, line 5036:
```suggestion
/// following example) in the following order, constructing the object using the first one
```

PR #17186, opened 2024-12-03
[ci] Try cloning 6 times and sleep in between attempts

dpiparo:
Try cloning 6 times and sleep in between attempts and, at each attempt, increase linearly the sleep time.


martamaja10 (approved):

dpiparo:
let's see this in action on the ci nodes

PR #15469, opened 2024-05-09
[core] Pass TString by value to TNamed ctor

ChristianTackeGSI:
# This Pull request:

Passing by const-reference can mean two constructions (once on the caller, once in the ctor using the TString copy ctor).

Instead passing by value and using move semantics in the ctor can reduce this to one construction on the caller and one move in the ctor.

## Checklist:

- [X] tested changes locally
- [ ] updated the docs (if necessary)

fixes: #15434

PR #17187, opened 2024-12-03
std::string_view for TNamed

flagarde:
# This Pull request: Move from char* to string_view for TNamed

## Changes or fixes:

As ROOT is now compiled using C++17, it is possible to use std::string_view in place of char*. For now only TNamed 'SetName' 'SetTitle' 'SetNameTitle' has been moved to std::string_view. Usage of std::string_view in other PR would allow to used std::string directly such as : 

```cpp
std::string name{"name"};
TH1F::TH1F th1(name,"title",1,0);
```
This PR is concentrating on TNamed as it quite low level in the class hierarchy.

## Checklist:

- [X] tested changes locally

### DIFF
--- a/core/base/inc/TNamed.h
+++ b/core/base/inc/TNamed.h
@@ -34,8 +34,7 @@ protected:
 
 public:
    TNamed(): fName(), fTitle() { } // NOLINT: not allowed to use = default because of TObject::kIsOnHeap detection, see ROOT-10300
-   TNamed(const char *name, const char *title) : fName(name), fTitle(title) { }
-   TNamed(const TString &name, const TString &title) : fName(name), fTitle(title) { }
+   TNamed(const std::string_view name, const std::string_view title) : fName(name), fTitle(title) { }
    TNamed(const TNamed &named);
    TNamed& operator=(const TNamed& rhs);
    virtual ~TNamed();
@@ -48,9 +47,9 @@ public:
             const char  *GetTitle() const override { return fTitle; }
             ULong_t  Hash() const override { return fName.Hash(); }
             Bool_t   IsSortable() const override { return kTRUE; }
-   virtual  void     SetName(const char *name); // *MENU*
-   virtual  void     SetNameTitle(const char *name, const char *title);
-   virtual  void     SetTitle(const char *title=""); // *MENU*
+   virtual  void     SetName(const std::string_view name); // *MENU*
+   virtual  void     SetNameTitle(const std::string_view name, const std::string_view title);
+   virtual  void     SetTitle(const std::string_view title=""); // *MENU*
             void     ls(Option_t *option="") const override;
             void     Print(Option_t *option="") const override;
    virtual  Int_t    Sizeof() const;
--- a/graf2d/asimage/src/TASImage.cxx
+++ b/graf2d/asimage/src/TASImage.cxx
@@ -6294,7 +6294,7 @@ const char *TASImage::GetTitle() const
 ////////////////////////////////////////////////////////////////////////////////
 /// Set a title for an image.
 
-void TASImage::SetTitle(const char *title)
+void TASImage::SetTitle(const std::string_view title)
 {
    if (fTitle.IsNull()) {
       CreateThumbnail();
@@ -6308,7 +6308,7 @@ void TASImage::SetTitle(const char *title)
    int stop = fTitle.Index("*/") - 1;
 
    if ((start > 0) && (stop - start > 0)) {
-      fTitle.Replace(start, stop - start, title);
+      fTitle.Replace(start, stop - start, title.data());
    }
 }
 
--- a/core/base/inc/TColor.h
+++ b/core/base/inc/TColor.h
@@ -69,7 +69,7 @@ public:
    void          Print(Option_t *option="") const override;
    virtual void  SetAlpha(Float_t a) { fAlpha = a; }
    virtual void  SetRGB(Float_t r, Float_t g, Float_t b);
-   void          SetName(const char* name) override;
+   void          SetName(const std::string_view name) override;
 
    static void    InitializeColors();
    static void    HLS2RGB(Float_t h, Float_t l, Float_t s, Float_t &r, Float_t &g, Float_t &b);
--- a/core/base/inc/TDirectory.h
+++ b/core/base/inc/TDirectory.h
@@ -256,7 +256,7 @@ public:
    virtual void        SetBufferSize(Int_t /* bufsize */) {}
    virtual void        SetModified() {}
    virtual void        SetMother(TObject *mother) {fMother = (TObject*)mother;}
-           void        SetName(const char* newname) override;
+           void        SetName(const std::string_view name) override;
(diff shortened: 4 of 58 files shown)


ferdymercury:
Related: https://github.com/root-project/root/pull/15469

graf2d/asimage/src/TASImage.cxx
```diff
@@ -6308,7 +6308,7 @@ void TASImage::SetTitle(const char *title)
    int stop = fTitle.Index("*/") - 1;
 
    if ((start > 0) && (stop - start > 0)) {
-      fTitle.Replace(start, stop - start, title);
+      fTitle.Replace(start, stop - start, title.data());
```
pcanal, line 6311: `std::string_view::data` is not 'null' terminated at the end of the 'view' and will return all the data from the string being view (starting from where the view starts):
```
root [0] std::string longname("this is a long name");
root [1] std::string_view v(longname.c_str()+5, 4);
root [2] v
(std::string_view &) "is a"[4]
root [3] v.data()
(std::basic_string_view<char, std::char_traits<char> >::const_pointer) "is a long name"
```

So any places inside the function where the `const char*` is being used needs to be reviewed to see:

1. if there is an overload of the function being called that takes a `std::string_view`
2. else consider adding that overload and/or change the signature to take a `std::string_view` (and recursively apply this source code change algorithm).
3. else (avoid if possible as it is much less performant) copy the `std::string_view` into a `std::string` before passing it.

core/base/inc/TNamed.h
```diff
@@ -48,9 +47,9 @@ class TNamed : public TObject {
             const char  *GetTitle() const override { return fTitle; }
             ULong_t  Hash() const override { return fName.Hash(); }
             Bool_t   IsSortable() const override { return kTRUE; }
-   virtual  void     SetName(const char *name); // *MENU*
-   virtual  void     SetNameTitle(const char *name, const char *title);
-   virtual  void     SetTitle(const char *title=""); // *MENU*
+   virtual  void     SetName(const std::string_view name); // *MENU*
```
pcanal, line 50: Those 3 functions are virtual functions that are overload in some places. (you already found many within `ROOT` itself).  It is known to also be overload in user code that have classes deriving (directly or indirectly) from `TNamed`.   This means that this change is *not* backward compatible for the user.  i.e. they need to change their code is order to use a version of ROOT containing this code.  Worse, **__If__** the user has not upgrade their code to use the `override` keyword (very very likely to be true in many cases), the code will still compile but start doing the "wrong" thing: in case where `SetName` is being called for their objects through the `TNamed` virtual interface, it will **no longer** call their override and instead revert to use the default.

A pattern would need to be devised to make sure that even if they don't change their code, their override is still be called.  One such solution (which will take some calendar time to complete and has significant drawbacks) is to (and this may and may not encounter some technical difficulties):

1. For some version of ROOT, let's call it N (where it would be 6.36 if we did this today)
	a.  Add only non virtual overload that takes a `std::string_view` then those the slow thing and create a `std::string` and pass the `c_str()` to the existing interface.
2. For version N+1 make the existing interface deprecated but for this to not cause problem, one also need to:
        b. Make sure that the `std::string_view` overload is prefered to the `const char*` overload even if the parameter is a `const char*` ... maybe (probably not) it can be done by making the interface `protected`.
3. For version N+2 make the `const char*` private so that we get (maybe/hopefully) an error message when user still overloads the `const char*` overload.

And (of course :)), we would have to test all those scenarii before even starting to merge the code.

Hopefully someone as a better idea to ease the transition (where the most important point is to not have a case where the user code silently does the wrong thing).

graf2d/asimage/src/TASImage.cxx
guitargeek: This is the problem with `std::string_view` and the reason why we didn't adopt it much yet. It's quite some work to avoid accidental copies, not sure how to argue it's worth it! Especially for our Python users there is no added benefit because any string-type class is pythonized to implicitly convert to `const char*`.

I was thinking about this years ago and realized there was not good backwards compatible solution without added boilerplate code, other than adding a new adapter class. This was done for RooFit:
https://github.com/root-project/root/blob/master/roofit/roofitcore/inc/RooStringView.h

Maybe this `RooStringView` can be generalized for the use in the rest of ROOT, like in TNamed? And maybe it should not have "StringView" in the name because it's not necessarily a view but more a `ROOT::StringAdapter`? But then these non-standard classes confuse the users..

Of course that doesn't help with `virtual` functions. We should keep the original virtual overloads there.

So that's why personally I concluded it was better to not change the status quo.

core/base/inc/TNamed.h
flagarde: Does the problem would only appeared if they have set the override keyword in their `SetName` ? 

In the case they did something like : 
```c++
class MyClass : public TNamed
{
public:
 void SetName(const char* name) override { /**...**/}
}
```

this code would not compile anymore as there is not void SetName(const char* name) in `TNamed`. While not backward compatible it does not create "wrong" behaviour. 

Does the compiler will not use their own function if they don't use the `override` keyword as before. 

If they did : 
```cpp
class Derived : public TNamed
{
public:
 void SetName(const char* name ){}
}

TNamed* t= new Derived();
t->SetName("name");
```
does actual behaviour would not be to call the TNamed::SetName() already anyway ? So to do the "wrong way" anyway as I imagine user would expect to call the Child function and not the on in Mother in this case.

graf2d/asimage/src/TASImage.cxx
flagarde: I guess option 1 can be easily done for the "Replace" function. For all the other changes in this PR at this end the std::string_view is passed to a TString at the end and as such is copied.

core/base/inc/TNamed.h
pcanal: > While not backward compatible it does not create "wrong" behaviour.

That is correct. Still the sudden compilation failures would only be acceptable if there is a very very clear advantage for the change ... in the eye of the users.

> Does the compiler will not use their own function if they don't use the override keyword as before.

The `override` keyword is optional and the current code (as in your 2nd example) will work perfectly (i.e. call `Derived::SetName`).

> does actual behaviour would not be to call the TNamed::SetName() already anyway ?

Nope. See previous answer.

flagarde: To my understanding the second example with call the TNamed one not the Child one if the override keyword is not here.  As the pointer is of TNamed class.

pcanal: I do not understand what you mean.  The execution seen in the example using CompilerExplorer: https://godbolt.org/z/4KE97nrfh shows that the Derived's class `SetName` is being called.  I.e. the `override` keywords is optional, the important part is that the base class function has the `virtual` keyword and that the derived function has the same signature.

flagarde: Yes, Sorry my bad :( I run the wrong test and was confused by the result

pcanal:
Let's close this until we have a workable plan forward.

PR #17191, opened 2024-12-04
Roo fit tutorial restructuring

DeclanGarvey:
# This Pull request:

## Changes or fixes:
- Restructuring of tutorials such that inside RooFit there contains RooFit Core, HistFactory, and RooStats as discussed.
- Also merging of rf407_latextables + rf206_treevistools → rf407_ComputationalGraphVisualization

## Checklist:

- [ ] tested changes locally
- [x] updated the docs (if necessary)

This PR fixes #

PR #17193, opened 2024-12-04
[tutorials] Roo fit tutorial restructuring

martamaja10:
This is a continuation of work from this PR: [root-project/root#17191](https://github.com/root-project/root/pull/17191) 

We create a Roofit category of tutorials in which we move Roofitcore (former roofit tutorials), roostats and histfactory.


guitargeek:
Hi @martamaja10, thank you very much for the initiative!

I see that right now, you named the categories like this:
```
  * RooFit Tutorials
     * RooFit Core
     * HistFactory
     * RooStats
```
And the directory structure:
```
roofit/roofitcore
roofit/histfactory
roofit/roostats
````

However, I wouldn't call the categories like that. The name "roofitcore" is already meaning something else in the context of this repository: it means the RooFit base classes for deriving pdfs and functions. Already "RooGaussian" is not a part of RooFit Core anymore, but the first tutorial is already about that class. In other words the name "RooFit Core" is an implementation detail on how we structure our libraries in C++ and I wouldn't reuse this term in the tutorials for a different context.

What about this suggestion, that I also discussed with Monica at the hackathon?
```
  * RooFit libraries (or RooFit Packages to sound more "pythonic")
     * RooFit
     * HistFactory
     * RooStats
```
And the directory structure:
```
roofit/
roofit/roofit
roofit/histfactory
roofit/roostats
```

guitargeek:
Actually, you don't have to move the files and the directories *at all* to get the desired changes in the tutorials. It's enough the rename the doxygen groups and introduce a new intermediate one for the RooFit tutorials that the three subgroups are part of.

I would also refrain from renaming the tutorials because there are many existing links to them all over the internet...

guitargeek (changes requested):
Please achieve the new grouping without renaming the files

guitargeek (changes requested):
Great, thanks a lot for wrapping this up!

Only two requests:

1. You are adding back the `rf509` tutorials in the first commit only to remove them again in a second commit. That's unnecessary. There should be only one commit doing the restructuring.
2. The `rf206_treevistools`tutorials were deleted in the first commit, but I think this should not be done in this PR because tutorial cleanups is different from the restructuring. Or maybe in the same PR, but in a different commit. Maybe better in a separate PR, because this change is not complete. My idea at the hackathon was to delete the `rf206` indeed because it only shows one new line of code, but on the other hand we should then show it "on the side" in some other tutorial or explain it better in the reference guide. But I can take care of that and you review the PR later, since I already have and idea :)

guitargeek (approved):
LGTM! Thanks for addressing the change requests

PR #17194, opened 2024-12-04
[win] use `std::rename` instead of `::rename`

bellenot:
Try to fix this kind of error when running `testUnfold7c`
```
Error in <TPostScript::Text>: Cannot open temporary file: exampleTR.eps_tmp_7812`
```


pcanal (approved):
I could not 'easily' find why this improves the behavior, but if it does fix it, there is no downside.

bellenot:
> I could not 'easily' find why this improves the behavior, but if it does fix it, there is no downside.

Well, just blindly trying, to see if that fixes the issue...

PR #17198, opened 2024-12-04
[json] correctly store/restore `TList`

linev:
There is principal difference between `lst->Add(obj)` and `lst->Add(obj, "")`
In first case `TObjLink` is created, in second `TObjOptLink`.
When such `TList` was stored in JSON and read back, all entries in the `TList` were `TObjOptLink`
Now such situation marked specially and reconstructed `TList` object is correct.

Modify storage and reading part.
Add correspondent test.

IMPORTANT!!!

Normal ROOT I/O also has problem with such `TList`. When read from the buffer, 
all entries with **empty** options converted into `TObjLink`. But `TObjLink::GetOption()` returns not empty string 
but `obj->GetOption()`. 

This can change drawing of TCanvas, when such canvas read from the file. 
There are many classes like `TH1` or `TPave` implementing `GetOption()` method and 
potentially returning non-empty string.

### DIFF
--- a/io/io/src/TBufferJSON.cxx
+++ b/io/io/src/TBufferJSON.cxx
@@ -1610,61 +1610,82 @@ void TBufferJSON::JsonWriteCollection(TCollection *col, const TClass *)
    // collection treated as JS Array
    AppendOutput("[");
 
-   bool islist = col->InheritsFrom(TList::Class());
    TMap *map = nullptr;
-   if (col->InheritsFrom(TMap::Class()))
+   TList *lst = nullptr;
+   if (col->InheritsFrom(TList::Class()))
+      lst = dynamic_cast<TList *>(col);
+   else if (col->InheritsFrom(TMap::Class()))
       map = dynamic_cast<TMap *>(col);
 
    TString sopt;
-   if (islist) {
+   Bool_t first = kTRUE;
+
+   if (lst) {
+      // handle TList with extra options
       sopt.Capacity(500);
       sopt = "[";
-   }
 
-   TIter iter(col);
-   TObject *obj;
-   Bool_t first = kTRUE;
-   while ((obj = iter()) != nullptr) {
-      if (!first)
-         AppendOutput(fArraySepar.Data());
+      auto lnk = lst->FirstLink();
+      while (lnk) {
+         if (!first) {
+            AppendOutput(fArraySepar.Data());
+            sopt.Append(fArraySepar.Data());
+         }
 
-      if (map) {
+         WriteObjectAny(lnk->GetObject(), TObject::Class());
+
+         if (dynamic_cast<TObjOptLink *>(lnk)) {
+            sopt.Append("\"");
+            sopt.Append(lnk->GetAddOption());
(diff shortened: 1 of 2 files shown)


linev:
Here modified test which fails with normal I/O:
```
TEST(TMemFile, TList_Add_Without_Option)
{
   TList lst0;
   lst0.Add(new TNamed("name0", "title0"));  // without option
   lst0.Add(new TNamed("name1", "title1"), ""); // with empty option
   lst0.Add(new TNamed("name2", "title2"), "option2"); // with non-empty option

   auto f = new TMemFile("test.root", "RECREATE");
   f->WriteTObject(&lst0, "list");

   lst0.Delete();

   auto lst1 = f->Get<TList>("list");
   delete f;

   EXPECT_NE(lst1, nullptr);

   auto link = lst1->FirstLink();

   EXPECT_STREQ("name0", link->GetObject()->GetName());
   EXPECT_STREQ("title0", link->GetObject()->GetTitle());
   EXPECT_STREQ("", link->GetAddOption()); // by default empty string returned
   EXPECT_EQ(dynamic_cast<TObjOptLink *>(link), nullptr);

   link = link->Next();

   EXPECT_STREQ("name1", link->GetObject()->GetName());
   EXPECT_STREQ("title1", link->GetObject()->GetTitle());
   EXPECT_STREQ("", link->GetAddOption());
   EXPECT_NE(dynamic_cast<TObjOptLink *>(link), nullptr); // this will fail in normal ROOT I/O

   link = link->Next();

   EXPECT_STREQ("name2", link->GetObject()->GetName());
   EXPECT_STREQ("title2", link->GetObject()->GetTitle());
   EXPECT_STREQ("option2", link->GetAddOption());
   EXPECT_NE(dynamic_cast<TObjOptLink *>(link), nullptr);

   link = link->Next();

   EXPECT_EQ(link, nullptr);
}
```

io/io/src/TBufferJSON.cxx
```diff
@@ -1610,61 +1610,82 @@ void TBufferJSON::JsonWriteCollection(TCollection *col, const TClass *)
    // collection treated as JS Array
    AppendOutput("[");
 
-   bool islist = col->InheritsFrom(TList::Class());
    TMap *map = nullptr;
-   if (col->InheritsFrom(TMap::Class()))
+   TList *lst = nullptr;
+   if (col->InheritsFrom(TList::Class()))
+      lst = dynamic_cast<TList *>(col);
+   else if (col->InheritsFrom(TMap::Class()))
```
pcanal, line 1617: Apriori we can simply the code with:
```suggestion
   TList *lst = dynamic_cast<TList *>(col);
   TMap *map = dynamic_cast<TMap*>(col);
```
i.e. I don't think that `InheritsFrom` is much faster than the `dynamic_cast`.

linev: I add commit with such changes

pcanal (approved):
LGTM.  Please open a new issue for the related binary I/O issue.

PR #17201, opened 2024-12-04
Remove dead code

dpiparo:


vepadulano (approved):
LGTM but I don't understand the `unused function 'GetPostscriptName' [-Wunused- function]` warning in the CI, is really this whole function unused?

dpiparo:
Good point. I think you are correct. This was something necessary only for macOS 10.9 Mavericks, which is not supported since 1 Dec 2016.

couet (approved):

PR #17205, opened 2024-12-05
strlen microoptimization: use compiletime-constant instead of runtime-constant

ferdymercury:
# This Pull request:

## Changes or fixes:

`strlen` cannot be a constexpr in all platforms otherwise.

This has probably no measurable performance gain, so feel free to close if you deem it's not worth.

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -624,7 +624,8 @@ bool TClassEdit::IsDefAlloc(const char *allocname, const char *classname)
    string_view a( allocname );
    // In Windows, allocname might be 'class const std::allocator<int>',
    // (never 'const class ...'), so we start by stripping the 'class ', if any
-   constexpr static int clalloclen = std::char_traits<char>::length("class ");
+   constexpr auto length = std::char_traits<char>::length;
+   constexpr static int clalloclen = length("class ");
    if (a.compare(0,clalloclen,"class ") == 0) {
       a.remove_prefix(clalloclen);
    }
@@ -634,7 +635,7 @@ bool TClassEdit::IsDefAlloc(const char *allocname, const char *classname)
    if (a=="__default_alloc_template<true,0>")   return true;
    if (a=="__malloc_alloc_template<0>")         return true;
 
-   constexpr static int alloclen = std::char_traits<char>::length("allocator<");
+   constexpr static int alloclen = length("allocator<");
    if (a.compare(0,alloclen,"allocator<") != 0) {
       return false;
    }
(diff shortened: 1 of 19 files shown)


core/foundation/src/TClassEdit.cxx
```diff
@@ -1419,30 +1419,30 @@ bool TClassEdit::IsStdClass(const char *classname)
 {
    classname += StdLen( classname );
    if ( strcmp(classname,"string")==0 ) return true;
-   if ( strncmp(classname,"bitset<",strlen("bitset<"))==0) return true;
+   if ( strncmp(classname,"bitset<",std::char_traits<char>::length("bitset<"))==0) return true;
```
guitargeek, line 1422:
```suggestion
   using std::char_traits<char>::length;

   if ( strncmp(classname,"bitset<",length("bitset<"))==0) return true;
```
Personally, I would prefer having a `using` statement at the beginning of `IsStdClass()` to make this less verbose.

guitargeek:
Hi @ferdymercury, thanks for this! It's also better to use the C++ standard library instead of holdovers from C.

> This has probably no measurable performance gain, so feel free to close if you deem it's not worth.

But there is also no disadvantage, right?

ferdymercury:
> But there is also no disadvantage, right?

Not that I know of.

guitargeek:
My bad, I didn't know that was a class member!

Maybe:
```c++
using ct = std::char_traits<char>;
ct::length(...); 
```
?

tree/treeplayer/src/TTreePlayer.cxx
```diff
@@ -2461,43 +2461,43 @@ Long64_t TTreePlayer::Scan(const char *varexp, const char *selection,
 
    if (opt.Contains("lenmax=")) {
       int start = opt.Index("lenmax=");
-      int numpos = start + strlen("lenmax=");
```
pcanal, line 2464: This function could also use the shortening of the name (and in general any function where there is more than one use of the function).

ferdymercury: Changed as suggested

core/foundation/src/TClassEdit.cxx
```diff
@@ -1083,7 +1085,7 @@ int TClassEdit::GetSplit(const char *type, vector<string>& output, int &nestedLo
                offset += 5;
             }
             static const char* char_traits_s = "char_traits<char>";
```
pcanal, line 1087:
```suggestion
            constexpr auto char_traits_s = "char_traits<char>";
            // or 
            // static constexpr char const* const char_traits_s = "char_traits<char>";
```

ferdymercury:
I guess the test failure is unrelated?

guitargeek:
Yes

guitargeek (approved):
Thanks a lot!

PR #17207, opened 2024-12-05
[ntuple] remove RPageSourceFriends

jblomer:
The functionality of the friend page source is replaced with a more robust design by the `RNTupleJoinProcessor`.


hahnjo (approved):
LG

PR #17209, opened 2024-12-05
[cling] Prototype SYCL support in ROOT

devajithvs:
# This Pull request:
Adds a prototype implementation of SYCL support in ROOT
## Checklist:
- [x] Clone AdaptiveCpp git repo instead of adding the whole AdaptiveCpp codebase to CLING (like clad)
- [x] Clean up CMakeLists:
  - Currently, AdaptiveCpp is installed in a temporary directory to get include directories and acpp binaries for SYCL - now follows what is done for CLAD
- [x] Building with ROOT works now
  - AdaptiveCpp requires LLVM be built with `-DLLVM_LINK_LLVM_DYLIB=ON`, which unfortunately ROOT doesn't support. Tried to get this working but this causes a cascade of issues. See: https://github.com/root-project/root/issues/12156#issuecomment-1660324761
  - Spent time trying various things, building AdaptiveCpp statically is what works. Some patches on top of adaptivecpp will be necessary
  - Cling builds and works independently, so it can be used in notebooks.
- [x] Add test cases for SYCL
- [x] Investigate DPC++: 
  - DPC++ does not currently support JIT

To build, you need the flag:
`-Dexperimental_adaptivecpp`

- [x] tested changes locally

Root's AdaptiveCpp fork (for static LLVM build and build-time usability - for genvectorx): https://github.com/root-project/AdaptiveCpp

Related: https://github.com/root-project/root/pull/18461

This PR fixes #


interpreter/CMakeLists.txt
```diff
@@ -53,7 +53,10 @@ set(LLVM_FORCE_USE_OLD_TOOLCHAIN ON CACHE BOOL "")
 # found before (because the user turned on cuda or tmva-gpu), this will not have
 # an effect, which is fine.
 # (Note that the option is very counter-intuitive: We turn *on* disabling it...)
-set(CMAKE_DISABLE_FIND_PACKAGE_CUDA ON)
+# Special case when we want to build cling with adaptiveCpp
```
hahnjo: Can you maybe a bit more comment here / integrate into the existing story. Something like
```suggestion
# Except when we want to build cling with adaptiveCpp, in which case adaptiveCpp needs to find CUDA in order to build the backend
```
(if that's the case?)

interpreter/cling/lib/Interpreter/CIFactory.cpp
```diff
@@ -1399,6 +1399,13 @@ namespace {
       argvCompile.push_back("-fno-omit-frame-pointer");
     }
 
+#ifdef CLING_WITH_ADAPTIVECPP
+    argvCompile.push_back("-D__ACPP_ENABLE_LLVM_SSCP_TARGET__");
+    argvCompile.push_back("-Xclang");
+    argvCompile.push_back("-disable-O0-optnone");
+    if (!debuggingEnabled)
+      argvCompile.push_back("-ffp-contract=fast");
```
hahnjo: This looks dangerous because `-ffp-contract=false` allows non-conforming floating-point optimizations. I think this will be confusing for users to get different results from their CPU code depending on whether Cling was built with SYCL support. Is this strictly needed?

interpreter/cling/lib/Interpreter/IncrementalParser.cpp
```diff
@@ -1033,9 +1033,11 @@ namespace cling {
       // Don't protect against crashes if we cannot run anything.
       // cling might also be in a PCH-generation mode; don't inject our Sema
       // pointer into the PCH.
+#ifndef CLING_WITH_ADAPTIVECPP
       if (!isCUDADevice && m_Interpreter->getOptions().PtrCheck)
         ASTTransformers.emplace_back(
             new NullDerefProtectionTransformer(m_Interpreter));
```
hahnjo, line 1039: Just to understand: this will entirely turn off pointer checks when built with SYCL?

hahnjo (commented):
Just commenting on the changes in `interpreter/`

interpreter/cling/lib/Interpreter/IncrementalParser.cpp
devajithvs: This transformer causes failures on CUDA and OpenMP backends. It should not be run for PTX code (as we already do), and I expect similar issues with HIP and other backends.

For OpenMP, it fails with an undefined `cling_runtime_internal_throwIfInvalidPointer` symbol:

Bash

```bash
[AdaptiveCpp Warning] [LoopSplitterInlining] cling_runtime_internal_throwIfInvalidPointer is not defined!
[AdaptiveCpp Warning] [omp_sscp_executable_object] Could not load library: jit-cache/7273508065178116436.12121653283064102631.jit.so (jit-cache/7273508065178116436.12121653283064102631.jit.so: undefined symbol: cling_runtime_internal_throwIfInvalidPointer)
[AdaptiveCpp Error] from /home/dvalapar/work/root/dev.adaptivecpp-proto/builddir/_deps/adaptivecpp-src/src/runtime/omp/omp_code_object.cpp:56 @ make_shared_library_from_blob(): omp_sscp_executable_object: could not load shared kernel library
[AdaptiveCpp Error] from /home/dvalapar/work/root/dev.adaptivecpp-proto/builddir/_deps/adaptivecpp-src/src/runtime/omp/omp_queue.cpp:524 @ submit_sscp_kernel_from_code_object(): omp_queue: Code object construction failed
```

I can look into this more and avoid disabling this check, but would be another patch on top of AdaptiveCpp (handling `cling_runtime_internal_throwIfInvalidPointer` separately) we can't potentially upstream.

interpreter/cling/lib/Interpreter/CIFactory.cpp
devajithvs: It's not strictly needed. I was trying to mimic AdaptiveCpp's behavior. We can remove it, as CUDA and HIP backends don't rely on the frontend for ffp-contract flags, and OpenMP users can pass it themselves if needed.

interpreter/CMakeLists.txt
devajithvs: Yes, that is right. Thanks.

interpreter/cling/lib/Interpreter/IncrementalParser.cpp
hahnjo: Ok. It would be nice if we can detect those backends the same way we have `isCUDADevice`, but I guess we can live with it for now.

hahnjo:
LGTM for the `interpreter/` changes. I didn't check the rest (CMake setup in particular)

dpiparo (approved):

PR #17211, opened 2024-12-05
[hist] Fix assertion failure (invalid comparator)

bellenot:
Fix the following error (visible with the `/RTC1` flag on Windows)
```
libHist: Debug Assertion Failed!
Expression: invalid comparator
```

### DIFF
--- a/hist/hist/src/TGraph.cxx
+++ b/hist/hist/src/TGraph.cxx
@@ -698,7 +698,7 @@ Bool_t TGraph::CompareArg(const TGraph* gr, Int_t left, Int_t right)
    Double_t xl = 0, yl = 0, xr = 0, yr = 0;
    gr->GetPoint(left, xl, yl);
    gr->GetPoint(right, xr, yr);
-   return (TMath::ATan2(yl, xl) > TMath::ATan2(yr, xr));
+   return (TMath::ATan2(yl, xl) >= TMath::ATan2(yr, xr));
 }
 
 ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


pcanal (changes requested):
If/Since this changes are needed we also need to update the function documentations and enhance the commit message to point/hint to the part of the standard indicating whether the comparators should strict (existing code) or not (this PR).

bellenot:
> If/Since this changes are needed we also need to update the function documentations and enhance the commit message to point/hint to the part of the standard indicating whether the comparators should strict (existing code) or not (this PR).

You mean this: https://en.cppreference.com/w/cpp/named_req/Compare

pcanal:
> You mean this: [en.cppreference.com/w/cpp/named_req/Compare](https://en.cppreference.com/w/cpp/named_req/Compare)

Yes something like that .. except that I am reading it as asking for strict comparison ... `For all a, comp(a, a) == false.`

bellenot:
> > You mean this: [en.cppreference.com/w/cpp/named_req/Compare](https://en.cppreference.com/w/cpp/named_req/Compare)
> 
> Yes something like that .. except that I am reading it as asking for strict comparison ... `For all a, comp(a, a) == false.`

Yes, I know, but the original code is from 2003 (see the [commit here](https://github.com/root-project/root/commit/10871c3c027666ef297b032672a59a6a75bb9d62) and was not using `std::stable_sort`, should I completely rewrite those methods?

guitargeek:
The solution of changing the comparator functions is not sustainable, because if one would do `TGraph::Sort()` with `ascending=false` the comparator is not inverted anymore and then it's not strict again.

I think the more sustainable and simpler solution is:
  * https://github.com/root-project/root/pull/18916

guitargeek:
Superseded by:
  * https://github.com/root-project/root/pull/18916

PR #4685, opened 2019-12-12
Fix the memory management of internal histograms in TMVA::ROCCalc class

lmoneta:
Some of the histograms cretaed internally by the TMVA ROCCalc class were deleted and others not.
For example the ROC curve histogram returned by ROCCalc::GetROCCurve was deleted in the destructor of
ROCCalc. This was causing a crash when exiting.

Some other histograms instead were not deleted

PR #17215, opened 2024-12-06
[TMVA] Specify ownership of histograms in the ROCCalc documentation

guitargeek:
Less risky alternative to #4685.


lmoneta (approved):
LGTM!
We can then probably close the PR #4685 and use this as solution

PR #17217, opened 2024-12-06
[cling] Always handle Error of `createObjectFile`

hahnjo:
This avoids test failures because the destructor of `llvm::Error` aborts he program "due to an unhandled Error", which can happen in case of race conditions with concurrent file operations in a directory.


vgvassilev (approved):
Good catch. @sahilpatidar, can you also include this change in our upstream PR?

hahnjo:
All nice and green :green_circle:

PR #6548, opened 2020-10-05
[DF] Check column types in RNTuple's GetColumnReadersImpl()

jblomer:
Fixes https://github.com/root-project/root/issues/10749

PR #17221, opened 2024-12-06
[DF] Check column types in GetColumnReadersImpl()

ferdymercury:
# This Pull request:

## Changes or fixes:

This is just an untested / rebased version of @jblomer 's PR https://github.com/root-project/root/pull/6548, just rebased here, now that the blocking issue has been solved in maser. Feel free to take over or close.

Fixes https://github.com/root-project/root/issues/10749

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/tree/dataframe/src/RNTupleDS.cxx
+++ b/tree/dataframe/src/RNTupleDS.cxx
@@ -277,7 +277,11 @@ void RNTupleDS::AddField(const RNTupleDescriptor &desc, std::string_view colName
          auto cardinalityField = std::make_unique<ROOT::Experimental::Internal::RRDFCardinalityField>();
          cardinalityField->SetOnDiskId(fieldId);
          fColumnNames.emplace_back("R_rdf_sizeof_" + std::string(colName));
-         fColumnTypes.emplace_back(cardinalityField->GetTypeName());
+         const auto typeName = cardinalityField->GetTypeName();
+         fColumnTypes.emplace_back(typeName);
+         std::string normalized;
+         TClassEdit::GetNormalizedName(normalized, typeName);
+         fNormalizedColumnTypes.emplace_back(normalized);
          fProtoFields.emplace_back(std::move(cardinalityField));
 
          for (const auto &f : desc.GetFieldIterable(fieldDesc.GetId())) {
@@ -359,13 +363,21 @@ void RNTupleDS::AddField(const RNTupleDescriptor &desc, std::string_view colName
 
    if (cardinalityField) {
       fColumnNames.emplace_back("R_rdf_sizeof_" + std::string(colName));
-      fColumnTypes.emplace_back(cardinalityField->GetTypeName());
+      const auto typeName = cardinalityField->GetTypeName();
+      fColumnTypes.emplace_back(typeName);
+      std::string normalized;
+      TClassEdit::GetNormalizedName(normalized, typeName);
+      fNormalizedColumnTypes.emplace_back(normalized);
       fProtoFields.emplace_back(std::move(cardinalityField));
    }
 
    fieldInfos.emplace_back(fieldId, nRepetitions);
    fColumnNames.emplace_back(colName);
-   fColumnTypes.emplace_back(valueField->GetTypeName());
+   const auto typeName = valueField->GetTypeName();
+   fColumnTypes.emplace_back(typeName);
+   std::string normalized;
+   TClassEdit::GetNormalizedName(normalized, typeName);
+   fNormalizedColumnTypes.emplace_back(normalized);
    fProtoFields.emplace_back(std::move(valueField));
 }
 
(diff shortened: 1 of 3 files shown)


dpiparo:
I stopped the builds since the code deserves a try, but after correcting a typo...

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -431,13 +443,30 @@ RDF::RDataSource::Record_t RNTupleDS::GetColumnReadersImpl(std::string_view /* n
...
 std::unique_ptr<ROOT::Detail::RDF::RColumnReaderBase>
-RNTupleDS::GetColumnReaders(unsigned int slot, std::string_view name, const std::type_info & /*tid*/)
+RNTupleDS::GetColumnReaders(unsigned int slot, std::string_view name, const std::type_info & tid)
 {
    // At this point we can assume that `name` will be found in fColumnNames
-   // TODO(jblomer): check incoming type
    const auto index = std::distance(fColumnNames.begin(), std::find(fColumnNames.begin(), fColumnNames.end(), name));
    auto field = fProtoFields[index].get();
 
+   std::string demangled = ROOT::Internal::RDF::DemangleTypeIdName(ti);
```
ferdymercury, line 452:
```suggestion
   std::string demangled = ROOT::Internal::RDF::DemangleTypeIdName(tid);
```

tree/dataframe/test/datasource_ntuple.cxx
```diff
@@ -51,6 +51,7 @@ class RNTupleDSTest : public ::testing::Test {
 
    void SetUp() override {
       auto model = RNTupleModel::Create();
+      *model->MakeField<std::uint32_t>("nevent", 1);
```
ferdymercury, line 54:
```suggestion
      *model->MakeField<std::uint32_t>("nevent") = 1;
```

ferdymercury:
Somehow, `EXPECT_THROW` is not working as expected.

dpiparo:
We are missing something here... The code seems to throw an exception with the right type, the log event shows that, but EXPECT_THROW does not catch the exception...

tree/dataframe/test/datasource_ntuple.cxx
```diff
@@ -132,6 +134,21 @@ TEST_F(RNTupleDSTest, CardinalityColumn)
...
+   auto df = ROOT::RDF::Experimental::FromRNTuple(fNtplName, fFileName);
+
+   // Allow use of float and Float_t interchangibly
+   EXPECT_DOUBLE_EQ(3.0, *df.Sum<std::vector<Float_t>>("jets"));
+   // Allow use of std int types and ROOT int types interchangibly
+   EXPECT_EQ(1U, df.Take<std::uint32_t>("nevent").GetValue()[0]);
+   EXPECT_EQ(1U, df.Take<UInt_t>("nevent").GetValue()[0]);
+   // jets is currently exposed as std::vector<float> and thus not usable as ROOT::RVec<float>
+   EXPECT_ANY_THROW(df.Sum<ROOT::RVec<float>>("jets"));
+   // EXPECT_THROW(df.Sum<ROOT::RVec<float>>("jets"), std::runtime_error); // This does not work directly, maybe due to jitting ?
```
jblomer, line 149: We should understand this before merging.

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -277,7 +277,11 @@ void RNTupleDS::AddField(const RNTupleDescriptor &desc, std::string_view colName
          auto cardinalityField = std::make_unique<ROOT::Experimental::Internal::RRDFCardinalityField>();
          cardinalityField->SetOnDiskId(fieldId);
          fColumnNames.emplace_back("R_rdf_sizeof_" + std::string(colName));
-         fColumnTypes.emplace_back(cardinalityField->GetTypeName());
+         const auto typeName = cardinalityField->GetTypeName();
+         fColumnTypes.emplace_back(typeName);
+         std::string normalized;
+         TClassEdit::GetNormalizedName(normalized, typeName);
+         fNormalizedColumnTypes.emplace_back(normalized);
```
jblomer, line 284: This and further down: I wonder if instead of adding the normalized type name everywhere where we append to fColumnTypes, perhaps we should loop once over all the column types when they are fully build, i.e. at the end of the constructor.

jblomer (commented):
Many thanks! Please let us know if you can continue to work on this, otherwise we can also take it from here.

tree/dataframe/test/datasource_ntuple.cxx
ferdymercury: Yep, the test is still failing even after that change, so the attempt was not useful

ferdymercury:
> Many thanks! Please let us know if you can continue to work on this, otherwise we can also take it from here.

Please take it from here ;) thanks!

vepadulano:
Note that this PR si superseded by https://github.com/root-project/root/pull/17749. Thanks @ferdymercury for pushing this!

PR #17224, opened 2024-12-09
[ci] Overwrite files when unzipping test results

hahnjo:
We always want to look at the latest test results; for example when rerunning a job with test failures in a workflow, its results should overwrite previous results.


hahnjo:
I got my spurious failure, but the job executed is still the old one: https://github.com/root-project/root/actions/runs/12234346440/job/34123356538#step:3:86 I guess it's taken from `master`, so we have to merge and check afterwards :neutral_face:

dpiparo (approved):

Issue #17222, opened 2024-12-08
Regression in Python ownership for histograms within subdirectories with ROOT 6.34.00


lahwaacz:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

I've just updated the [Arch Linux root package](https://archlinux.org/packages/extra-staging/x86_64/root/) to 6.34.00-1 and afterwards we noticed that several tests from the [uproot test suite](https://github.com/scikit-hep/uproot5/tree/main/tests) cause a segfault. The first such test is [tests/test_0345_bulk_copy_method.py::test_bulk_copy](https://github.com/scikit-hep/uproot5/blob/main/tests/test_0345_bulk_copy_method.py).

We have the details including a full backtrace in [this issue](https://gitlab.archlinux.org/archlinux/packaging/packages/python-uproot/-/issues/2) on Arch Linux GitLab. Can you take a look?

Note that the affected packages are still in the extra-staging repository where a rebuild for Python 3.13 is currently in progress.

### Reproducer

N/A

### ROOT version

6.34.00

### Installation method

build from source

### Operating system

Arch Linux

dpiparo:
Thanks for submitting this issue. Uproot is not part of ROOT: I would suggest to submit the issue to the uproot tracker repo https://github.com/scikit-hep/uproot5 .

lahwaacz:
@dpiparo The segfault comes from `libcppyy.so` which is part of root, though.

lahwaacz:
@dpiparo The following code is a simple reproducer for the issue:

```python
import os

import ROOT

tmp_path = "/tmp"
source_filename = os.path.join(tmp_path, "source.root")

f_histogram = ROOT.TFile(source_filename, "recreate")
f_histogram.mkdir("subdir")
f_histogram.cd("subdir")
x = ROOT.TH1F("hist", "title", 100, -5, 5)
x.Write()
f_histogram.Close()
```

```
$ test-env/bin/python bug.py
 *** Break *** segmentation violation



===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================
#0  0x00007a922249ca87 in wait4 () from /usr/lib/libc.so.6
#1  0x00007a92223eca9b in ?? () from /usr/lib/libc.so.6
#2  0x00007a9220f95094 in TUnixSystem::StackTrace() () from /usr/lib/root/libCore.so
#3  0x00007a9221521d53 in ?? () from /usr/lib/root/libcppyy_backend.so
[...]

PR #17229, opened 2024-12-09
Give ownership to C++ for objects with directory registration logic

vepadulano:
Fixes https://github.com/root-project/root/issues/17222


guitargeek (approved):
Thanks for addressing the ownership situation in a systematic way! I guess this needs to be backported. In the future we might want to look into avoiding to have to do all of these Pythonizations at the Python side, by adding new flags to TObject for example.

PR #17231, opened 2024-12-10
[tmva][pymva] Change default algorithm for AdaBoost to SAMME

lmoneta:
Since scikit version 1.4 the SAMME.R algorithm is deprecated and it has been removed since version 1.6. Change default to use SAMME
See https://scikit-learn.org/1.5/modules/generated/sklearn.ensemble.AdaBoostClassifier.html

This should fix test failures observed since upgrading to latest scikit-version


vepadulano (approved):
Thank you, LGTM! I believe this needs a backport all the way back to 6.30

PR #17233, opened 2024-12-10
[ntuple] create RNTupleExporter

silverweed:
This is an internal utility used to dump pages of an RNTuple. In the future it might be extended to also dump other interesting things or accept a wider range of options (e.g. customize the output file names).

Ideally it should supersede [ntuple_dump.C](https://github.com/jblomer/iotools/blob/master/ntuple_dump.C).

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntupleutil/CMakeLists.txt
+++ b/tree/ntupleutil/CMakeLists.txt
@@ -16,9 +16,11 @@ endif()
 ROOT_STANDARD_LIBRARY_PACKAGE(ROOTNTupleUtil
 HEADERS
   ROOT/RNTupleImporter.hxx
+  ROOT/RNTupleExporter.hxx
   ROOT/RNTupleInspector.hxx
 SOURCES
   v7/src/RNTupleImporter.cxx
+  v7/src/RNTupleExporter.cxx
   v7/src/RNTupleInspector.cxx
 LINKDEF
   LinkDef.h
--- a/tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
+++ b/tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
@@ -1,0 +1,54 @@
+/// \file ROOT/RNTupleExporter.hxx
+/// \ingroup NTuple ROOT7
+/// \author Giacomo Parolini <giacomo.parolini@cern.ch>
+/// \date 2024-12-10
+/// \warning This is part of the ROOT 7 prototype! It will change without notice. It might trigger earthquakes. Feedback
+/// is welcome!
+
+/*************************************************************************
+ * Copyright (C) 1995-2022, Rene Brun and Fons Rademakers.               *
+ * All rights reserved.                                                  *
+ *                                                                       *
+ * For the licensing terms see $ROOTSYS/LICENSE.                         *
+ * For the list of contributors see $ROOTSYS/README/CREDITS.             *
+ *************************************************************************/
+
+#ifndef ROOT7_RNTupleExporter
+#define ROOT7_RNTupleExporter
+
+#include <filesystem>
+#include <vector>
+
+namespace ROOT::Experimental::Internal {
+
+class RPageSource;
+
+struct RExportPagesOptions {
+   enum RExportPageFlags {
+      kNone = 0x0,
+      kIncludeChecksums = 0x1,   
+      kReportExportedFileNames = 0x2,
+
+      kDefaults = kReportExportedFileNames
+   };
+
+   std::filesystem::path fOutputPath = ".";
+   std::uint64_t fFlags = kDefaults;
+};
+
+struct RExportPagesResult {
+   int fNPagesExported;
+   // This only gets filled if kReportExportedFileNames was part of the options flags
+   std::vector<std::string> fExportedFileNames;
+};
+
+class RNTupleExporter {
+public:
+   /// Given a page source, writes all its pages to individual files (1 per page).
+   /// If the source is not already attached, it will be attached by this process.
+   static RExportPagesResult ExportPages(RPageSource &source, const RExportPagesOptions &options = {});
+};
+
+} // namespace ROOT::Experimental::Internal
+
+#endif
--- a/tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
+++ b/tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
@@ -16,7 +16,8 @@
 #ifndef ROOT7_RNTupleExporter
 #define ROOT7_RNTupleExporter
 
-#include <filesystem>
+#include <cstdint>
+#include <string>
 #include <vector>
 
 namespace ROOT::Experimental::Internal {
@@ -35,7 +36,7 @@ public:
          kDefaults = kShowProgressBar
       };
 
-      std::filesystem::path fOutputPath;
+      std::string fOutputPath;
       std::uint64_t fFlags;
 
       RPagesOptions() : fOutputPath("."), fFlags(kDefaults) {}
--- a/tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
+++ b/tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
@@ -6,7 +6,7 @@
 /// is welcome!
 
 /*************************************************************************
- * Copyright (C) 1995-2022, Rene Brun and Fons Rademakers.               *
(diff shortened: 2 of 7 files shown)


tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
```diff
@@ -0,0 +1,54 @@
...
+   };
+
+   std::filesystem::path fOutputPath = ".";
+   std::uint64_t fFlags = kDefaults;
+};
+
+struct RExportPagesResult {
+   int fNPagesExported;
+   // This only gets filled if kReportExportedFileNames was part of the options flags
+   std::vector<std::string> fExportedFileNames;
```
hahnjo, line 42: Does this really cost so much that we need it optional? Compared to reading and writing all pages, I would expect filling a `std::vector<std::string>` to be quite cheap...

tree/ntupleutil/v7/src/RNTupleExporter.cxx
```diff
@@ -0,0 +1,124 @@
...
+
+   RColumnExportInfo(const RNTupleDescriptor &desc, const RColumnDescriptor &colDesc, const RFieldDescriptor &fieldDesc)
+      : fColDesc(&colDesc),
+        fFieldDesc(&fieldDesc),
+        fQualifiedName(desc.GetQualifiedFieldName(fieldDesc.GetId()) + '-' + std::to_string(colDesc.GetIndex()))
+   {
+   }
+};
+
+static void AddColumnsFromField(std::vector<RColumnExportInfo> &vec, const RNTupleDescriptor &desc,
```
hahnjo, line 39: I believe this could move into an anonymous `namespace` to avoid it from being exported (even in `Internal`)

tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
```diff
@@ -0,0 +1,54 @@
...
+struct RExportPagesOptions {
+   enum RExportPageFlags {
+      kNone = 0x0,
+      kIncludeChecksums = 0x1,   
+      kReportExportedFileNames = 0x2,
+
+      kDefaults = kReportExportedFileNames
+   };
+
+   std::filesystem::path fOutputPath = ".";
```
hahnjo, line 35: AFAICT the implementation isn't using any of `std::filesystem::path` methods, it just converts to a string. Maybe the options should just take a `std::string` then?

tree/ntupleutil/v7/src/RNTupleExporter.cxx
```diff
@@ -0,0 +1,124 @@
...
+
+struct RColumnExportInfo {
+   const RColumnDescriptor *fColDesc;
+   const RFieldDescriptor *fFieldDesc;
+   std::string fQualifiedName;
+
+   RColumnExportInfo(const RNTupleDescriptor &desc, const RColumnDescriptor &colDesc, const RFieldDescriptor &fieldDesc)
+      : fColDesc(&colDesc),
+        fFieldDesc(&fieldDesc),
+        fQualifiedName(desc.GetQualifiedFieldName(fieldDesc.GetId()) + '-' + std::to_string(colDesc.GetIndex()))
```
hahnjo, line 34: Can you maybe add a comment why we don't need to take `colDes.GetRepresentationIndex()` into account? ie there's only one active representation and all other columns are suppressed.

tree/ntupleutil/v7/src/RNTupleExporter.cxx
```diff
@@ -0,0 +1,124 @@
...
+         const auto &pages = clusterDesc.GetPageRange(columnId);
+         const auto &colRange =clusterDesc.GetColumnRange(columnId);
+         std::uint64_t pageIdx = 0;
+         if (reportFilenames)
+            res.fExportedFileNames.reserve(res.fExportedFileNames.size() + pages.fPageInfos.size());
+
+         R__LOG_DEBUG(0) << "exporting column \"" << colInfo.fQualifiedName << "\" (" << pages.fPageInfos.size()
+                         << " pages)";
+
+         for (const auto &pageInfo : pages.fPageInfos) {
```
hahnjo, line 94: I believe this works automatically, but maybe add an explicit check for `colRange.fIsSuppressed` in which case `pages.fPageInfos` should be `empty()`...

tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
silverweed: For the internal usage std::string would suffice yes, but I think there is value to having a semantically-meaningful type here (since this is user-facing code). For example, a user might want to call some of the path-specific methods on this variable and it's convenient if the variable already is of the "right" type.

silverweed:
Converting to draft because I want to try an alternative way to export the pages: first do a pass to collect all pages informations and then export them (this may simplify some parts of the code and interface)

tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
silverweed: The solution of pre-counting the pages indeed solves the performance issues. There is still a matter of memory being potentially wasted, but it's in the order of 5 MiB for 100k pages. I will remove the flag for now to keep the code simpler; if this ever becomes a concern it's trivial to fix.

tree/ntupleutil/v7/src/RNTupleExporter.cxx
```diff
@@ -0,0 +1,167 @@
...
+         assert(!colRange.fIsSuppressed || pages.fPageInfos.empty());
+
+         for (const auto &pageInfo : pages.fPageInfos) {
+            ROnDiskPage::Key key{columnId, pageIdx};
+            const ROnDiskPage *onDiskPage = cluster->GetOnDiskPage(key);
+
+            // dump the page
+            const void *pageBuf = onDiskPage->GetAddress();
+            const auto maybeChecksumSize = (options.fFlags & RExportPagesOptions::kIncludeChecksums) ? 8 : 0;
+            const std::uint64_t pageBufSize = pageInfo.fLocator.fBytesOnStorage + maybeChecksumSize;
```
jblomer, line 135: Note that we also have to check if the page actually has a checksum (`pageInfo.fHasChecksum`). NB: perhaps we should amend the documentation of `ROnDiskPage` and clarify that this store the page data and the checksum (if applicable).

tree/ntupleutil/v7/test/ntuple_exporter.cxx
```diff
@@ -0,0 +1,269 @@
...
+   FileRaii fileGuard("ntuple_exporter.root");
+
+   // Create RNTuple to export
+   {
+      auto model = RNTupleModel::Create();
+      auto pFlt = model->MakeField<float>("flt");
+      auto pVec = model->MakeField<std::vector<int>>("vec");
+
+      auto opts = RNTupleWriteOptions();
+      opts.SetCompression(505);
```
jblomer, line 53: I don't think that zstd guarantees bit-by-bit identical output across versions. If we compare with reference bytes, I'd rather use uncompressed files.

tree/ntupleutil/v7/test/ntuple_exporter.cxx
jblomer, line 1: Perhaps add a test where the file is written without checksums.

jblomer (changes requested):
Very nice! Some minor comments.

jblomer (approved):
LGTM

tree/ntupleutil/CMakeLists.txt
```diff
@@ -34,4 +34,8 @@ if(MSVC)
   target_compile_definitions(ROOTNTupleUtil PRIVATE _USE_MATH_DEFINES)
 endif()
 
+if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
+    target_link_libraries(ROOTNTupleUtils stdc++fs)
+endif()
+
```
hahnjo, line 40: Hm, for me this would be a strong reason not to use `std::filesystem` in the interface definition, especially if we don't need it...

pcanal: this was removed.

tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
pcanal: The type was changed to std::string.

PR #17234, opened 2024-12-10
Fix dereference value-initialized vector iterator

bellenot:
Fix the following error on Windows:
```
Debug assertion Failed!
Can't dereference value-initialized vector iterator
```
As the standard says:
`std::vector::end()` returns an iterator to the element following the last element of the vector.
This element acts as a placeholder; attempting to access it results in undefined behavior.
See https://en.cppreference.com/w/cpp/container/vector/end 
And with empty vectors, vector.begin() == vector.end()

### DIFF
--- a/io/io/src/TEmulatedMapProxy.cxx
+++ b/io/io/src/TEmulatedMapProxy.cxx
@@ -71,7 +71,7 @@ void* TEmulatedMapProxy::At(UInt_t idx)
    // Return the address of the value at index 'idx'.
    if ( fEnv && fEnv->fObject )   {
       PCont_t c = PCont_t(fEnv->fObject);
-      return idx<(c->size()/fValDiff) ? ((char*)&(*c->begin())) + idx*fValDiff : 0;
+      return idx<(c->size() / fValDiff) ? c->data() + idx * fValDiff : 0;
    }
    Fatal("TEmulatedMapProxy","At> Logic error - no proxy object set.");
    return 0;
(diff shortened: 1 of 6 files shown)


io/io/src/TEmulatedMapProxy.cxx
```diff
@@ -71,7 +71,7 @@ void* TEmulatedMapProxy::At(UInt_t idx)
    // Return the address of the value at index 'idx'.
    if ( fEnv && fEnv->fObject )   {
       PCont_t c = PCont_t(fEnv->fObject);
-      return idx<(c->size()/fValDiff) ? ((char*)&(*c->begin())) + idx*fValDiff : 0;
+      return idx<(c->size() / fValDiff) ? c->data() + idx * fValDiff : 0;
```
pcanal, line 74: Minor readability improvement on the existing code:
```suggestion
      return (idx<(c->size() / fValDiff)) ? (c->data() + idx * fValDiff) : 0;
```

io/io/src/TGenCollectionProxy.cxx
```diff
@@ -1514,13 +1514,8 @@ void TGenCollectionProxy__VectorCreateIterators(void *obj, void **begin_arena, v
...
    }
-   *begin_arena = &(*vec->begin());
-#ifdef R__VISUAL_CPLUSPLUS
-   *end_arena = &(*(vec->end()-1)) + 1; // On windows we can not dererence the end iterator at all.
-#else
-   // coverity[past_the_end] Safe on other platforms
-   *end_arena = &(*vec->end());
-#endif
+   *begin_arena = vec->data();
+   *end_arena = vec->data() + vec->size(); // We can not dereference the end iterator at all.
```
pcanal, line 1518:
```suggestion
   *end_arena = vec->data() + vec->size();
```
The comment seems useful only in reference to the now gone code.

pcanal (approved):
LVGTM.  Thank you!

io/io/src/TEmulatedMapProxy.cxx
bellenot: right, I missed this one, thanks!

PR #17235, opened 2024-12-10
[ntuple] set proper UUID in simple mini file writer

jblomer:


tree/ntuple/v7/src/RMiniFile.cxx
```diff
@@ -512,9 +513,14 @@ struct RTFFile {
 /// A zero UUID stored at the end of the TFile record
 struct RTFUUID {
    RUInt16BE fVersionClass{1};
-   unsigned char fUUID[16] = {0};
+   unsigned char fUUID[16];
 
-   RTFUUID() = default;
+   RTFUUID()
+   {
+      TUUID uuid;
```
silverweed, line 520: TFile has the `kReproducible` flag which, if set, ensures the UUID is always 0. Should we provide the same mechanism for RMiniFile?

jblomer: I don't think we need to add it for the simple file writer. Users who wish to use reproducible files can provide their own `TFile` and the the proper file writer will take care of handling the UUID.

tree/ntuple/v7/src/RMiniFile.cxx
```diff
@@ -512,9 +513,14 @@ struct RTFFile {
 /// A zero UUID stored at the end of the TFile record
 struct RTFUUID {
    RUInt16BE fVersionClass{1};
-   unsigned char fUUID[16] = {0};
+   unsigned char fUUID[16];
```
pcanal, line 516: That seems undersized ...

tree/ntuple/v7/src/RMiniFile.cxx
```diff
@@ -512,9 +513,14 @@ struct RTFFile {
...
 
-   RTFUUID() = default;
+   RTFUUID()
+   {
+      TUUID uuid;
+      char *buffer = reinterpret_cast<char *>(this);
+      uuid.FillBuffer(buffer);
```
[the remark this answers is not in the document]
silverweed: `sizeof(*this)` is different from `sizeof(TUUID)` because `TUUID` has transient members (its size is 20, but the on-disk size is 16)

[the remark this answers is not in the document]
pcanal: Fair enough.  Implicit was the question of whether we wanted the life object or only the ondisk representation.
Eitherway, I would strongly recommend adding the corresponding assert in case anything changes (unlikely at it may be).

hahnjo (approved):
LGTM, but see Philippe's comment in https://github.com/root-project/root/pull/17235/files#r1878346443

tree/ntuple/v7/src/RMiniFile.cxx
jblomer: Clarified by the discussion with @silverweed, AFAICT

PR #17236, opened 2024-12-10
[io] Set empty TKey default title

jblomer:
Currently, the default title for a TKey is "object title". Objects that do not derive from TObject have no means to change this title. Change the default title to the empty string to save a few bytes per non-TObject object and to reduce the "noise" when listing directory contents.

Companion PR: https://github.com/root-project/roottest/pull/1242


enirolf (approved):
There is still a clang-format complaint about `TDirectory* motherDir` (pointer should attach to `motherDir`), otherwise LGTM, thanks!

pcanal (approved):
Thanks.

hahnjo (changes requested):
There's a crash in `roottest-root-io-TFile-make` that might be related (?), and `roottest-root-meta-runnamespace-auto` needs an update of the reference file

hahnjo (approved):

PR #17240, opened 2024-12-11
[6.32][tmva][pymva] Change default algorithm for AdaBoost to SAMME 

dpiparo:
BP of https://github.com/root-project/root/pull/17231


lmoneta (approved):
Thank you!

PR #17244, opened 2024-12-11
[tutorials] Move to legacy and reorganize

mdessole:
# This Pull request:

creates histogram legacy tutorial category, moving there `ratioplotOld.C` macro and its python version. It also improves some categories' naming.


vepadulano (approved):
LGTM, but consider applying clang-format suggestions before merging

PR #17245, opened 2024-12-11
[df] Early return if empty string in InterpreterCalc

vepadulano:
FYI @gpetruc


tree/dataframe/inc/ROOT/RDF/Utils.hxx
```diff
@@ -197,7 +197,7 @@ void InterpreterDeclare(const std::string &code);
 /// Jit code in the interpreter with TInterpreter::Calc, throw in case of errors.
 /// The optional `context` parameter, if present, is mentioned in the error message.
 /// The pointer returned by the call to TInterpreter::Calc is returned in case of success.
```
ferdymercury, line 199:
```suggestion
```

vepadulano: Thanks, this hasn't been the case since a long time anyway

dpiparo (approved):

PR #17246, opened 2024-12-11
[ntuple] Simplify bulk reading of entire ranges

hahnjo:


tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
```diff
@@ -740,7 +741,7 @@ public:
    /// Reads 'size' values from the associated field, starting from 'firstIndex'. Note that the index is given
    /// relative to a certain cluster. The return value points to the array of read objects.
    /// The 'maskReq' parameter is a bool array of at least 'size' elements. Only objects for which the mask is
-   /// true are guaranteed to be read in the returned value array.
+   /// true are guaranteed to be read in the returned value array. A 'nullptr' means to read all elements.
```
hahnjo, line 744: This is different than what we originally discussed: I first implemented that `maskReq` is mandatory and then omit it as an optimization in `ReadBulk(RNTupleClusterRange)`. However, this felt weird - why would we allow something for ourselves, but not for our users... But if wanted, I can easily change it back

tree/ntuple/v7/src/RFieldBase.cxx
```diff
@@ -620,7 +620,7 @@ std::size_t ROOT::Experimental::RFieldBase::ReadBulkImpl(const RBulkSpec &bulkSp
    std::size_t nRead = 0;
    for (std::size_t i = 0; i < bulkSpec.fCount; ++i) {
       // Value not needed
-      if (!bulkSpec.fMaskReq[i])
+      if (bulkSpec.fMaskReq && !bulkSpec.fMaskReq[i])
```
silverweed, line 623: I'd suggest caching `bulkSpec.fMaskReq` in a local variable. [Godbolt](https://godbolt.org/z/Y66vsqYs8) shows that it allows the compiler to optimize the loop better since it can know it's a loop invariant and basically compiles two different versions of the loop skipping the check for every iteration.

silverweed: Hm, I probably misread the assembly, but it stands to reason that the compiler will be able to create a more efficient version of the loop when having more information about the invariants, and I hardly consider an additional local variable as code complexity.

hahnjo: We might be able to help the compiler a bit by putting both `fMaskReq` and `fMaskAvail` in local variables, but before I see performance data I wouldn't do it.

silverweed: Fine for me

jblomer (approved):
LGTM

PR #17247, opened 2024-12-11
[RF] extend RooBinIntegrator to N-dimensions

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-9996

Currently it failed silently for more than 3 dimensions

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/roofit/roofitcore/src/RooBinIntegrator.cxx
+++ b/roofit/roofitcore/src/RooBinIntegrator.cxx
@@ -165,8 +165,7 @@ double RooBinIntegrator::integral(const double *)
    
    ROOT::Math::KahanSum<double> sum;
    
-   std::vector<double> center(_function->getDimension());
-   recursive_integration(0,1.,center.data(),sum);
+   recursive_integration(0,1.,sum);
    
    return sum.Sum();
 }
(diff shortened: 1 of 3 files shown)


roofit/roofitcore/src/RooBinIntegrator.cxx
```diff
@@ -160,69 +156,42 @@ bool RooBinIntegrator::checkLimits() const
...
 
-
+/**
+ * @brief It performs recursively for loops to perform N-dimensional integration
+ * @param d the current recursivity depth (dimension currently being for-looped)
+ * @param delta the (d-1)-dimensional bin width/area/volume/hypervolume...
+ * @param center the (d-1)-dimensional bin center of the current iteration
+ * @param sum the resulting integral where to accumulate the integral, passed by reference
+ */
+void RooBinIntegrator::recursive_integration(const UInt_t d, const double delta, const std::vector<double> center, ROOT::Math::KahanSum<double>& sum) {
```
guitargeek, line 180:
```suggestion
void RooBinIntegrator::recursive_integration(const UInt_t d, const double delta, std::span<const double> center, ROOT::Math::KahanSum<double>& sum) {
```
Vectors are expensive to copy because of manual memory allocation. Can we use a `std::span` here? Or at least a `const &`?

guitargeek (changes requested):
Thank you very much for the initiative, this is great! Please refactor the code a bit to not not introduce too many manual memory allocations by constructing `std::vector`. These are very expensive in performance. I guess just one allocation of some working memory that is passed around in the recursive function should work, if it can't be avoided entirely.

roofit/roofitcore/src/RooBinIntegrator.cxx
```diff
@@ -160,69 +156,42 @@ bool RooBinIntegrator::checkLimits() const
...
+   const std::vector<double>& binb = _binb[d];
+   const bool isLastDim = d+1 == _function->getDimension();
+   for (unsigned int ibin=0; ibin < binb.size() - 1; ++ibin) {
+      const double hi = binb[ibin + 1];
+      const double lo = binb[ibin];
+      const double mid = (hi+lo)/2.;
+      std::vector<double> bcenter = center;
```
guitargeek, line 187: This is what I mean: creating one `std::vector` for each bin in every dimension should be avoided 100 %.

guitargeek (approved):
Excellent, thank you so much for the improvement and addressing the review comments!

PR #16809, opened 2024-11-03
[tmva][sofie] Add Sin and Cos Operators to SOFIE

extint:
# This Pull request:
This Pull request adds the support for Sin and Cos operators in SOFIE
## Changes or fixes:
- Implemented the respective ROperator files
- Included respective tests and ONNX models

## Checklist:
- [x] tested changes locally
- [ ] updated the docs (if necessary)

This PR fixes #

PR #17248, opened 2024-12-11
[tmva][sofie] Add new operators and function to check input ONNX model

lmoneta:
This Pull request adds these new operators (including their tests): 
- Pad
- Where
- Sin and Cos (as new Unary operators). For these reuse tests implemented in #16809
- Einsum operator with optimal impelmentation using BLAS when possible (e.g for matrix multiplications)
- Split : include now implementation for all cases. 

In addition, implement: 
- a function to check the ONNX model for missing operators (not yet implmented in SOFIE) in the ONNX parser.
   Example of usage:

```
    TMVA::Experimental::SOFIE::RModelParser_ONNX p;
    bool ret = p.CheckModel("model.onnx");
 ```

- Fix retrieving the value data for the Constant operator
- Optimize Broadcast function to avoid unneeded allocations

With this PR SOFIE can now parse the CaloDiT distilled model for fast simulation, GNN2 and GNN3 from ATLAS


sanjibansg (approved):
LGTM! Thanks for this implementation.

PR #17249, opened 2024-12-11
[gpad] Fix Debug Assertion Failure in libGpad

bellenot:
Fix the following error in Debug mode with Runtime Check on Windows:
Debug Assertion Failed!
Expression: vector<bool> subscript out of range


couet (approved):
LGTM

PR #17250, opened 2024-12-11
[python] Factor memory utility functions in one module

vepadulano:


dpiparo (approved):
This is an elegant solution. LGTM.

guitargeek (approved):
Thanks a lot! Can this also be backported to the 6.34 branch?

I think it's to early to let the history diverge for this relatively new code where we might have to follow up on with more fixes. A diverging history makes it harder and more error prone to backport then.

PR #17253, opened 2024-12-11
Add C++23 support

CHN-beta:
# This Pull request:

## Changes or fixes:

C++23 support of ROOT (see https://root-forum.cern.ch/t/enable-c-23-in-root-6-34-00-rc1/62433).

Some changes are needed for code using `std::unique_ptr`. According to C++ standards, in the context that `T` is declared but not defined, `std::unique_ptr<T>` is constructible but not destructible, thus we need to move the definitions of some functions calling `std::unique_ptr<T>::~unique_ptr` from the header file into the source file. This rule is not new in C++ 23, but it causes compile failure when `-std=c++23` is specified, so we must fix it to support C++ 23.

## Checklist:

- [x] tested changes locally  -- but I only tested these changes on 6.34.00-rc1, not on master. If test on master is mandatory, please tell me and I will do it tomorrow.

### DIFF
--- a/graf3d/eve7/src/REveCaloData.cxx
+++ b/graf3d/eve7/src/REveCaloData.cxx
@@ -129,6 +129,9 @@ REveCaloData::REveCaloData(const char* n, const char* t):
    // Constructor.
 }
 
+REveCaloData::~REveCaloData() {}
+void REveCaloData::SetSelector(REveCaloDataSelector* iSelector) { fSelector.reset(iSelector); }
+
 ////////////////////////////////////////////////////////////////////////////////
 /// Process newly selected cells with given select-record.
 
--- a/roofit/histfactory/inc/RooStats/HistFactory/HistRef.h
+++ b/roofit/histfactory/inc/RooStats/HistFactory/HistRef.h
@@ -12,8 +12,7 @@
 #define HISTFACTORY_HISTREF_H
 
 #include <memory>
-
-class TH1;
+#include <TH1.h>
 
 namespace RooStats{
 namespace HistFactory {
--- a/roofit/roofitcore/inc/RooFit/Detail/NormalizationHelpers.h
+++ b/roofit/roofitcore/inc/RooFit/Detail/NormalizationHelpers.h
@@ -70,8 +70,7 @@ template <class T>
 std::unique_ptr<T> compileForNormSet(T const &arg, RooArgSet const &normSet)
 {
    RooFit::Detail::CompileContext ctx{normSet};
-   std::unique_ptr<RooAbsArg> head = arg.compileForNormSet(normSet, ctx);
-   return std::unique_ptr<T>{static_cast<T *>(head.release())};
+   return std::unique_ptr<T>{static_cast<T *>(arg.compileForNormSet(normSet, ctx).release())};
 }
 
 } // namespace Detail
--- a/graf3d/eve7/inc/ROOT/REveCaloData.hxx
+++ b/graf3d/eve7/inc/ROOT/REveCaloData.hxx
@@ -174,7 +174,7 @@ protected:
 
 public:
    REveCaloData(const char* n="REveCaloData", const char* t="");
-   ~REveCaloData() override {}
+   ~REveCaloData() override;
 
    void    FillImpliedSelectedSet(Set_t& impSelSet, const std::set<int>& sec_idcs) override;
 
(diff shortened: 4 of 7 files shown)


graf3d/eve7/src/REveCaloData.cxx
```diff
@@ -45,6 +45,9 @@ void REveCaloData::CellGeom_t::Dump() const
 
 ////////////////////////////////////////////////////////////////////////////////
 
+REveCaloData::~REveCaloData() {}
+void REveCaloData::SetSelector(REveCaloDataSelector* iSelector) { fSelector.reset(iSelector); }
```
osschar, line 49: Can you please move those two down below REveCaloData constructor, on line 118.

CHN-beta: No problem, I will do it now.

dpiparo:
Thanks for these changes. I started the builds and tests on the master branch.

osschar (approved):
OK for graf3d/eve7 -- thank you!

roofit/roofitcore/inc/RooFit/Detail/NormalizationHelpers.h
```diff
@@ -70,8 +70,7 @@ template <class T>
 std::unique_ptr<T> compileForNormSet(T const &arg, RooArgSet const &normSet)
 {
    RooFit::Detail::CompileContext ctx{normSet};
-   std::unique_ptr<RooAbsArg> head = arg.compileForNormSet(normSet, ctx);
-   return std::unique_ptr<T>{static_cast<T *>(head.release())};
+   return std::unique_ptr<T>{static_cast<T *>(arg.compileForNormSet(normSet, ctx).release())};
```
dpiparo, line 73: Could you please comment about the need of this change?

roofit/histfactory/inc/RooStats/HistFactory/HistRef.h
```diff
@@ -28,37 +28,23 @@ class HistRef {
 
 
    /// constructor - use gives away ownerhip of the given pointer
-   HistRef(TH1 * h = nullptr) : fHist(h) {}
+   HistRef(TH1 * h = nullptr);
```
dpiparo, line 31: Instead of moving to the cxx the code, would you mind including TH1.h in the header file?

dpiparo (changes requested):
Thanks for this nice PR. I left two comments in the code, which will be easily addressed. Would you mind also proposing the changes in 1 single commit? Once this is merged, I will enable 23 for a few builds.

CHN-beta:
> Would you mind also proposing the changes in 1 single commit?

No problem.

roofit/roofitcore/inc/RooFit/Detail/NormalizationHelpers.h
CHN-beta: Ok, let me explain the need of this change.

In the code without my patch, `std::unque_ptr<RooAbsArg> head` is constructed in the function, and it is needed to be destructed at the end of the function. However, `RooAbsArg` was still not defined, so it could not be destructed. In my test (gcc 13, with `-std=c++23`), this will cause a compile error.

So, generally speaking, we should move the function body to c++ source file. However, this is a function template, simply move the definition into source file will lead to failure in [template implicit instantiation](https://en.cppreference.com/w/cpp/language/function_template#Implicit_instantiation). Thus, I removed the definition of local variable `head` to avoid calling to its destrutor.

Of course, the return value of `arg.compileForNormSet(normSet, ctx)` is typed `std::unqie_ptr` and should also be destructed in this function, but this will not cause a compile error in my test. To my understand, this is because compiler did not know the type of the return value of `arg.compileForNormSet` before `T` is specified (i.e., before the function template is called and instantiated), so this statement will only be compiled when `T` is specified (i.e. when the function is actually called, where the template is implicitely instantiated). I am not sure this is a compiler specified behavior or in C++ standard.

roofit/histfactory/inc/RooStats/HistFactory/HistRef.h
CHN-beta: Sounds good. I will try it ASAP.

CHN-beta: It works.

dpiparo:
> > Would you mind also proposing the changes in 1 single commit?
> 
> No problem.

Thank you.

PR #17254, opened 2024-12-11
RooSimultaneous: fix bug with category fraction calculation. 

will-cern:
And ensure only current category state is considered, even in case where same pdf used for two states


guitargeek (approved):
Thank you very much for the fix! I have also added a test.

PR #17256, opened 2024-12-11
[textinput] add CTRL-backslash as keyboard shortcut for sigquit 

ferdymercury:
# This Pull request:

## Changes or fixes:

Finally, a way to hard-terminate ROOT with keyboard shortcut, without having to write .qqqqqqqqqq or closing the terminal.

Fixes https://its.cern.ch/jira/browse/ROOT-9906

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/base/src/TApplication.cxx
+++ b/core/base/src/TApplication.cxx
@@ -1339,6 +1339,7 @@ void TApplication::Help(const char *line)
          Printf("   Ctrl+D           : quit ROOT (if empty line)");
          Printf("   Ctrl+C           : send kSigInt interrupt signal");
          Printf("   Ctrl+Z           : send kSigStop pause job signal");
+         Printf("   Ctrl+\\           : send kSigQuit quit job signal");
 
          Printf("   Arrow_Down       : navigate downwards in command history [Ctrl+N]");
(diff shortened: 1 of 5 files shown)


core/base/src/TApplication.cxx
```diff
@@ -1339,6 +1339,7 @@ void TApplication::Help(const char *line)
          Printf("   Ctrl+D           : quit ROOT (if empty line)");
          Printf("   Ctrl+C           : send kSigInt interrupt signal");
          Printf("   Ctrl+Z           : send kSigStop pause job signal");
+         Printf("   Ctrl+\\           : send kSigQuit quit job signal");
```
silverweed, line 1342: If we advertise this shortcut we should either support it on Windows as well or (less ideal) hide this line on Windows.

ferdymercury: I've proposed an idea, but maybe @bellenot can advice on whether this is fine or it's better to call ::abort()

bellenot: At first glance I have no strong opinion. One can also simply use non Windows specific method like `std::abort`, `std::exit`, or `std::terminate`

PR #17259, opened 2024-12-12
Fix invalid read in TChain

vepadulano:
After merging #17301 , there is no need to create the logic to suppress the TChainIndex build error presently. The invalid read found by valgrind is still there, so this PR includes a test for the global join case and fixes the memory error.

### DIFF
--- a/tree/dataframe/test/dataframe_incomplete_entries.cxx
+++ b/tree/dataframe/test/dataframe_incomplete_entries.cxx
@@ -11,7 +11,10 @@
 #include <gtest/gtest.h>
 
 #include <numeric> // std::numeric_limits
+#include <random>
+#include <string>
 #include <thread>  // std::thread::hardware_concurrency
+#include <vector>
 
 template <typename T0, typename T1>
 void expect_vec_eq(const T0 &v1, const T1 &v2)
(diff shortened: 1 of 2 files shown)


tree/dataframe/test/dataframe_incomplete_entries.cxx
```diff
@@ -1093,16 +1093,116 @@ TEST_P(RDFMissingBranchesVec, DefaultValueForSimpleVariation)
...
+   }
+
+   static void TearDownTestSuite()
+   {
+      for (auto &&fileName : fFileNames)
+         std::remove(fileName);
+   }
+};
+
+TEST_P(RDFIndexedFriendsNotOrdered, SuppressIndexOrderingError)
```
pcanal, line 1155: The test seems to be failing.  Is it known why?

vepadulano: Not really, in fact it does not fail on all platforms. Needs to be investigated

vepadulano:
@pcanal investigating the test error led me to finding some faulty behaviour in the case of a chain moving to a new tree but not properly updating the information of the underlying TTreeIndex (specifically regarding the attached TTreeFormulas). You can find some description in the first commit message, and here is the full valgrind stacktrace 

[valgrind_tchain_friend_updateformulaleaves.txt](https://github.com/user-attachments/files/18168406/valgrind_tchain_friend_updateformulaleaves.txt)

Let's see if this was all or just part of the reason for the CI failures.

pcanal (approved):

PR #17260, opened 2024-12-12
Fix assertion failure in roofitcore

bellenot:
Fix Assertion failed: can't dereference out of range vector iterator


guitargeek (approved):
LGTM, thanks!

PR #17264, opened 2024-12-12
[RF] new pipe operators for JSONInterface

cburgard:
… TVectorD

# This Pull request:

adds new `<<` operators to the RooFit JSONInterface

## Changes or fixes:

- add a new operator that can pipe `std::vector<T>` directly into json without explicit use of `fill_seq`.
- add a new operator that can pipe `std::map<T_string,T>` directly into json (using the `set_map` method`
- add an implicit conversion for `TVectorT` to `std::vector<T>` to facilitate direct piping of `TVectorT` into json as well via the new `std::vector<T>` operator

## Checklist:

- [x] tested changes locally

This PR fixes #


cburgard:
@guitargeek Since the JSONInterface is quite useful, I was thinking it might be worthwhile moving it out of the `Detail` namespace.
What do you think of this?

guitargeek:
As discussed with Carsten in private, let's keep it in the detail namespace as long as Carsten is the sole developer and user (outside of the RooFit implementation details)

guitargeek (approved):

PR #17279, opened 2024-12-16
[rntuple] Fix assertion failure in RNtuple

bellenot:
Fix the following error on Windows:
```
Assertion failed: LeadingZeroes(q) >= unusedBits, file tree\ntuple\v7\src\RColumnElement.hxx, line 1038
```
`_BitScanReverse` returns the bit position of the first set bit (1) found,
so subtract it from the position of the last bit (31)


silverweed (approved):
thanks!

PR #16466, opened 2024-09-18
[ntuple] Mark missing fields for automatic schema evolution

hahnjo:
They are skipped while reading in the same way as a transient member.

PR #17281, opened 2024-12-16
Move RResult, RError, RException out of Experimental

jblomer:
Start of the PR series to move an initial set of RNTuple APIs out of the experimental namespace.


hahnjo (approved):
LGTM. This will need a rebase after https://github.com/root-project/root/pull/16466 and also https://github.com/root-project/root/pull/17285 (depending on sequencing) - sorry...

Issue #15520, opened 2024-05-15
Rewrite the RDataFrame JIT logic to avoid controlled leaks


vepadulano:
### Explain what you would like to see improved and how.

The RDataFrame JIT logic makes use of a series of controlled memory leaks between different functions to ensure lifetime of objects that are needed by the JIT until that actually happens.

This is an optimization but has certain fragilities, that were already analysed in a recent [bug report](https://github.com/root-project/root/issues/15399). The same topic surfaced in the past (see [this issue moved from JIRA](https://github.com/root-project/root/issues/15076)). And the fact we rely on leaking objects is subtly error-prone (see a recent [PR](https://github.com/root-project/root/pull/15477)).

We should investigate how to implement a different way to handle the JIT machinery so that it does not require leaking objects.

### ROOT version

Any

### Installation method

Any

### Operating system

Any

vepadulano:
Fixed by https://github.com/root-project/root/pull/17282

PR #17282, opened 2024-12-16
[df] [for discussion] JIT graph creation functions only once

gpetruc:
# This PR

This is some code, that could also be improved further, to reduce the per-sample JIT cost in RDataFrame, and potentially a way to avoid the "controlled memory leaks" #15520 (though this code currently does not yet address them in full).

I don't propose to merge this code as is, but I'm opening the PR as a reference for future discussion

### Before this PR

Currently JITed nodes of the computation graph, RDF JITs the function code just once, but JITs code lines to create the computational graph once per graph, with code like 
```cpp
ROOT::Internal::RDF::JitDefineHelper<ROOT::Internal::RDF::DefineTypes::RDefineTag>(
        R_rdf::func45, 
        new const char*[3]{"Jet_isSelected", "Jet_area", "Jet_pt"}, 3, 
        "SelJet_area",
        reinterpret_cast<ROOT::Detail::RDF::RLoopManager*>(0x562b52fe7cb0),
        reinterpret_cast<std::weak_ptr<ROOT::Detail::RDF::RJittedDefine>*>(0x562b5530d120),
        reinterpret_cast<ROOT::Internal::RDF::RColumnRegister*>(0x562b552e1900),
        reinterpret_cast<std::shared_ptr<ROOT::Detail::RDF::RNodeBase>*>(0x562b552e0d20));
```
This is done by methods like `BookDefineJit` that craft the code as a string and pass it to `RLoopManager::toJitExec`. Calls are accumulated, and executed in one go by `RLoopManager::Jit()`

### With this PR 

Methods like `BookDefineJit`  now call a new function `RLoopManager::RegisterJitHelperCall` and the whole operation is refactored in three steps

1. a definition via JIT of a registrator function, done once per function, kind of action (e.g. Define) and column name
```cpp
namespace R_rdf {
  void jitNodeRegistrator_70(
        ROOT::Detail::RDF::RLoopManager *lm, 
        std::shared_ptr<ROOT::Detail::RDF::RNodeBase> *prevNodeOnHeap,
        ROOT::Internal::RDF::RColumnRegister* colRegister, 
        const std::vector<std::string> & colNames, 
        void *wkJittedDefine, void *) {
                 ROOT::Internal::RDF::JitDefineHelper<ROOT::Internal::RDF::DefineTypes::RDefineTag>(R_rdf::func45,
                 colNames, 
                 "SelJet_area", 
                 lm, reinterpret_cast<std::weak_ptr<ROOT::Detail::RDF::RJittedDefine>*>(wkJittedDefine),
                 colRegister, 
                 prevNodeOnHeap);
        }
}
```
2. lookup of the function address in memory, via calls to `gInterpreter`
3. from C++ call the registrator function once per computation graph

Similarly to the current setup, the declarations from point (1) are accumulated in a single string and passed to the interpreter in one go by `RLoopManager::Jit()`, and the address lookup (2) is also done inside `RLoopManager::Jit()`

The deferred function calls (3) are done both in  `RLoopManager::Jit()` and also in `RLoopManager::Run` in order to also allow doing them in multiple threads when RunGraphs is used (RunGraphs calls Jit only on one of the loop managers, and then calls Run on all multithreaded)

## What next?
 * I tested the code on several RDF analyses, but it could be tested more extensively, and also not just on x86_64.
 * The column name could also be gathered in the registration, to reduce the number of registrator functions to JIT
 * The objects created on the heap by `BookDefineJit` and deleted in the JITed calls like `JitDefineHelper` could be instead owned by the deferred function call object via RAII
 * Code could be cleaned further, e.g. the registration key could become an integer and the map a concurrent container to avoid the read lock, a c++ std::variant could also be used instead of a `void *` for the weak pointer, the extra column names for variations could be in a smart pointer, ...


vepadulano:
Thank you @gpetruc for kickstarting this absolutely non-trivial change to a core part of RDataFrame! It is definitely worth discussing, I will take a look at this.

vepadulano:
I've taken the liberty of rebasing this PR on master. Locally, I have tried to exploit these changes as a baseline to remove the controlled leaks of https://github.com/root-project/root/issues/15520 , but I've hit a blocker because of `CallBuildAction` where not only the jitted action pointer, the column register pointer and the previous node pointer are deleted, but also the pointer to the `HelperArgType` instance which does not fit in the workflow of this PR (it would require some sort of template `DeferredJitCall` or another way to make it agnostic. Before proceeding, I'd like to see what's missing for this PR to be merged, then possibly tackle the leaks in another PR

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -766,24 +788,81 @@ void RLoopManager::Jit()
...
+            auto declid = gInterpreter->GetFunction(clinfo, codeAndName.second.c_str());
+            assert(declid);
+            auto minfo = gInterpreter->MethodInfo_Factory(declid);
+            assert(gInterpreter->MethodInfo_IsValid(minfo));
+            auto mname = gInterpreter->MethodInfo_GetMangledName(minfo);
+            addr = reinterpret_cast<JitHelperFunc_t>(gInterpreter->FindSym(mname));
+            gInterpreter->MethodInfo_Delete(minfo);
+         }
+      }
+      gInterpreter->ClassInfo_Delete(clinfo);
```
vepadulano: @pcanal @vgvassilev I would like your opinion in this section. What happens here is

* Declaration of the code that will eventually run the JITted functions (e.g. JitFilterHelper in ROOT/RDF/InterfaceUtils.hxx). 
* in the for loop, look at each of these "lazy JITters", and attempt a fast retrieval of the function pointer from the interpreter. This is then cached in the map

Does the current approach look good to you? Are there any further optimization opportunities here?

vgvassilev: This approach looks reasonable to me.

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -766,24 +788,81 @@ void RLoopManager::Jit()
...
 
    TStopwatch s;
    s.Start();
-   RDFInternal::InterpreterCalc(code, "RLoopManager::Run");
+   if (!codeToDeclare.empty()) {
+      // This takes care of autoloading when needed
+      // If instead we used gInterpreter->Declare, builds with runtime_cxxmodules=OFF would fail in
+      // jitted actions with custom helpers with errors like:
+      // error: 'MyHelperType' is an incomplete type
+      // return std::make_unique<Action_t>(Helper_t(std::move(*h)), bl, std::move(prevNode), colRegister);
```
vgvassilev: Does that result into a different performance between modules and non-modules? If so you can check if you have modules on using `HasPCMForLibrary` and do different things in both cases...

vepadulano:
@gpetruc I have squashed your proposed changes in one commit, the second commit introduces improvements that also lead to fixing #15520. Our CI is running well with the changes, let me know if you have time to check this PR again with your analysis, I'll be running some tests with valgrind locally as well.

gpetruc:
Hi @vepadulano,

Thanks for reviving and updating this PR, I've run a quick check with one of my analysis benchmarks and it looks ok.

I'm failing to do more tests because for reasons I don't yet understand the el9 docker containers I'm building with this root branch or the master one fail to access eos ( `Plugin version SecClnt v5.8.0 is incompatible with seckrb5 v5.9.1 (must be <= 5.8.x) in sec.protocol libXrdSeckrb5-5.so` )

gpetruc:
Okay, after fixing my docker builds now I managed to run also more tests with this PR, and confirm that results look ok.

vepadulano:
Tested the changes of this PR with our rootbench RDataFrame benchmarks at https://github.com/root-project/rootbench/blob/master/root/tree/dataframe/RDataFrameBenchmarks.cxx . On my machine, the overall execution time for all the benchmarks lowers:

master:
```
 1/21 Test  #1: rootbench-RDFBenchmarks .............................................   Passed   57.18 sec
```
patch:
```
 1/21 Test  #1: rootbench-RDFBenchmarks .............................................   Passed   41.73 sec
```

In particular, certain cases see a very large improvement. This example with JITting of a large graph (O(10K) JIT nodes with reusage of the same 2 functions)  sees a factor 2 reduction, going from 14 seconds to 7 seconds.

master
```
1: BM_RDataFrame_JitLargeGraphs/10000                              14413906 us     14411597 us            1
```
patch:
```
1: BM_RDataFrame_JitLargeGraphs/10000                                723228 us       723064 us            1
```

I tried offline and the difference would be even more dramatic if the computation graph functions would be reused not only within the same computation graph, but also across computation graphs, which is the main purpose of this PR.

I attach the CSV outputs from rootbench in case anyone is interested

[patch.csv](https://github.com/user-attachments/files/24522602/patch.csv)
[master.csv](https://github.com/user-attachments/files/24522603/master.csv)

I believe this PR is good for a final review, adding the other RDF devs @martamaja10 @hageboeck

tree/dataframe/inc/ROOT/RDF/RInterfaceBase.hxx
```diff
@@ -201,10 +201,11 @@ protected:
                                                                              fColRegister, proxiedPtr->GetVariations());
       auto jittedActionOnHeap = RDFInternal::MakeWeakOnHeap(jittedAction);
 
-      auto toJit = RDFInternal::JitBuildAction(validColumnNames, upcastNodeOnHeap, typeid(HelperArgType),
-                                               typeid(ActionTag), helperArgOnHeap, nullptr, nSlots, fColRegister,
-                                               GetDataSource(), jittedActionOnHeap, vector2RVec);
-      fLoopManager->ToJitExec(toJit);
+      auto definesCopy = new RDFInternal::RColumnRegister(fColRegister); // deleted in jitted call
```
hageboeck: If it's deleted in the call, could it possibly be passed as a unique_ptr by value?

*Edit*: I see now that it's also used in JITted code, so if it's too much work to do it, please ignore.

tree/dataframe/inc/ROOT/RDF/RLoopManager.hxx
```diff
@@ -240,6 +263,9 @@ public:
    void IncrChildrenCount() final { ++fNChildren; }
    void StopProcessing() final { ++fNStopsReceived; }
    void ToJitExec(const std::string &) const;
+   void RegisterJitHelperCall(const std::string &funcBody, std::shared_ptr<RNodeBase> *prevNodeOnHeap,
+                              ROOT::Internal::RDF::RColumnRegister *colRegister,
```
hageboeck: Could this be std::unique_ptr<ROOT::Internal::RDF::RColumnRegister> ?

tree/dataframe/src/RDFInterfaceUtils.cxx
```diff
@@ -764,50 +757,43 @@ BookVariationJit(const std::vector<std::string> &colNames, std::string_view vari
    const auto colRegisterAddr = PrettyPrintAddr(colRegisterCopy);
    auto jittedVariation = std::make_shared<RJittedVariation>(colNames, variationName, variationTags, type, colRegister,
                                                              lm, parsedExpr.fUsedCols);
+   auto variedColsOnHeap = new ColumnNames_t(colNames);
```
hageboeck: I'm not sure this is getting deleted, but please close if it's OK.

tree/dataframe/inc/ROOT/RDF/RLoopManager.hxx
```diff
@@ -204,24 +204,26 @@ class RLoopManager : public RNodeBase {
...
-           colRegister(cols),
-           colNames(colnames),
-           wkJittedNode(wkNodePtr),
-           argument(arg)
-      {
-      }
+      std::size_t fFunctionId{};
+      std::unique_ptr<ROOT::Internal::RDF::RColumnRegister> fColRegister;
+      std::vector<std::string> fColNames;
+      // Non-owning pointer to the jitted node.
```
hageboeck: This looks like it's (co-)owning, so maybe this comment can be removed?

tree/dataframe/src/RJittedAction.cxx
```diff
@@ -159,3 +160,8 @@ std::unique_ptr<ROOT::Internal::RDF::RActionBase> RJittedAction::CloneAction(voi
    assert(fConcreteAction != nullptr);
    return fConcreteAction->CloneAction(newResult);
 }
+
+std::shared_ptr<ROOT::Detail::RDF::RNodeBase> RJittedAction::MoveOutPrevNode()
+{
+   return std::move(fPrevNode);
```
hageboeck, line 166: This looks to be leaving the member in an empty state. Can this be a problem?

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -96,19 +96,61 @@ std::string &GetCodeToDeclare()
...
+void DeclareAndRetrieveDeferredJitCalls(const std::string &codeToDeclare)
+{
+   // Step 1: Declare the DeferredJitCall functions to the interpreter
+   // We use ProcessLine to ensure meta functionality (e.g. autoloading) is
+   // processed when needed.
+   // If instead we used Declare, builds with runtime_cxxmodules=OFF would fail
+   // in jitted actions with custom helpers with errors like:
+   // error: 'MyHelperType' is an incomplete type
+   // return std::make_unique<Action_t>(Helper_t(std::move(*h)), bl, std::move(prevNode), colRegister);
+   //                                   ^
```
hageboeck, line 127: Maybe the details of the error message can be removed.

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -96,19 +96,61 @@ std::string &GetCodeToDeclare()
...
+   gInterpreter->ProcessLine(codeToDeclare.c_str());
+
+   // Step 2: Retrieve the declared functions as function pointers, cache them
+   // for later use in RunDeferredCalls
+   auto &funcIdToFuncPointersMap = GetJitHelperFuncMap();
+   auto &funcBodyToFuncIdMap = GetJitFuncBodyToFuncIdMap();
```
hageboeck, line 133: Do we need a lock from here on?

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -837,19 +860,20 @@ void RLoopManager::Jit()
 void RLoopManager::RunDeferredCalls()
 {
    if (!fJitHelperCalls.empty()) {
-      R__READ_LOCKGUARD(ROOT::gCoreMutex); // methods are thread-safe but funcMap isn't (yet)
-      TStopwatch s2;
-      s2.Start();
-      auto &funcMap = GetJitHelperFuncMap();
+      // funcMap is not thread-safe (yet) and fJitHelperCalls is cleared at the end of this function
+      R__WRITE_LOCKGUARD(ROOT::gCoreMutex);
```
hageboeck: Is read enough? Given that `at` should be guaranteed to find the function, the map should stay invariant.

tree/dataframe/inc/ROOT/RDF/RInterfaceBase.hxx
hageboeck: This seems to have been resolved in the second commit.

tree/dataframe/src/RLoopManager.cxx
hageboeck: Looks like the lock is held in the calling function, but can we get there via another call path?

tree/dataframe/src/RJittedAction.cxx
vepadulano: This is the solution I adopted to avoid circular references with `std::shared_ptr`. We have
```
   struct DeferredJitCall {
[...]
      std::shared_ptr<void> fJittedNode;
[...]
 }
```

`fJittedNode` is a JITted node of any type (say e.g. `RJittedFilter`) which will hold internally a `shared_ptr` to the previous node. When the true node (i.e. the `RFilter` within this example) is created, that will also have a `shared_ptr` to the same previous node. The `DeferredJitCall` is owned by the `RLoopManager` it will never go out of scope until this RDataFrame is gone. The true node (i.e. `RFilter` within this example) is owned by the jitted node, so there's the circular reference.

This approach should not lead to any problem because any time a `Filter`, `Define` etc. is called we also call `RLoopManager::RegisterJitHelperCall` which will then lead to eventually calling the `MoveOutPrevNode`, and there is only one such call per JITted node.

tree/dataframe/src/RLoopManager.cxx
vepadulano: I would like to leave them there because it's such an obscure message and obscure situation that I want it to be clear why it's done.

tree/dataframe/src/RLoopManager.cxx
vepadulano: No I agree we need a lock in this whole function, it's both using the interpreter and writing data to the maps.

vepadulano:
@hageboeck I have addressed your comments in a separate commit, will squash later on :+1:

vepadulano:
Following offline discussion, here is an example showing the difference in terms of amount of code seen by the interpreter before/after this PR.

Take the following code
```cpp
cat jit_large_graphs.cpp
#include <ROOT/RLogger.hxx>
#include <iostream>
#include <vector>
#include <string>

auto verbosity = ROOT::RLogScopedVerbosity(ROOT::Detail::RDF::RDFLogChannel(), ROOT::ELogLevel::kDebug + 10);

void jit_large_graph()
{
    ROOT::RDataFrame root(1);
    std::vector<ROOT::RDF::RResultPtr<double>> dfs;
    const auto n = 2;
    for (auto i = 0u; i < n; i++) {
        const auto column = "x" + std::to_string(i);
        auto define = root.Define(column, "42.f");
        auto filter = define.Filter(column + " > 0.f");
        auto sum = filter.Sum(column);
        dfs.emplace_back(sum);
    }
    // Trigger event loop
    dfs[0].GetValue();
}

int main()
{
  // Run twice to see the benefit of reusing the same JITted node registrators
  jit_large_graph();
  jit_large_graph();
}
```

Executing the code will print, among other things, the code that is JITted. 

When running this program with ROOT master, we can see (just extracting interesting parts here)
```
// These lines refer to the actual function bodies that are JITted
namespace R_rdf {
auto func0(){return 42.f
;}
using func0_ret_t = typename ROOT::TypeTraits::CallableTraits<decltype(func0)>::ret_type;
}
namespace R_rdf {
auto func1(const float var0){return var0 > 0.f
;}
using func1_ret_t = typename ROOT::TypeTraits::CallableTraits<decltype(func1)>::ret_type;
}
```
These come from the strings passed to the `Define` and `Filter` in the application. Then, there are the lines used to actually instantiate the JITted nodes:
```
ROOT::Internal::RDF::JitDefineHelper<ROOT::Internal::RDF::DefineTypes::RDefineTag>(R_rdf::func0, new const char*[0]{}, 0, "x0", reinterpret_cast<ROOT::Detail::RDF::RLoopManager*>(0x33547580), reinterpret_cast<std::weak_ptr<ROOT::Detail::RDF::RJittedDefine>*>(0x346a0070), reinterpret_cast<ROOT::Internal::RDF::RColumnRegister*>(0x345225e0), reinterpret_cast<std::shared_ptr<ROOT::Detail::RDF::RNodeBase>*>(0x33547ce0));
ROOT::Internal::RDF::JitFilterHelper(R_rdf::func1, new const char*[1]{"x0"}, 1, "", reinterpret_cast<std::weak_ptr<ROOT::Detail::RDF::RJittedFilter>*>(0x3469a160), reinterpret_cast<std::shared_ptr<ROOT::Detail::RDF::RNodeBase>*>(0x346a03b0),reinterpret_cast<ROOT::Internal::RDF::RColumnRegister*>(0x346ab120));
ROOT::Internal::RDF::CallBuildAction<ROOT::Internal::RDF::ActionTags::Sum, float>(reinterpret_cast<std::shared_ptr<ROOT::Detail::RDF::RNodeBase>*>(0x345225c0), new const char*[1]{"x0"}, 1, 1, reinterpret_cast<shared_ptr<double>*>(0x3469a2a0), reinterpret_cast<std::weak_ptr<ROOT::Internal::RDF::RJittedAction>*>(0x3469cbf0), reinterpret_cast<ROOT::Internal::RDF::RColumnRegister*>(0x3469af60));ROOT::Internal::RDF::JitDefineHelper<ROOT::Internal::RDF::DefineTypes::RDefineTag>(R_rdf::func0, new const char*[0]{}, 0, "x1", reinterpret_cast<ROOT::Detail::RDF::RLoopManager*>(0x33547580), reinterpret_cast<std::weak_ptr<ROOT::Detail::RDF::RJittedDefine>*>(0x3469a4e0), reinterpret_cast<ROOT::Internal::RDF::RColumnRegister*>(0x346ab720), reinterpret_cast<std::shared_ptr<ROOT::Detail::RDF::RNodeBase>*>(0x3469a980));
ROOT::Internal::RDF::JitFilterHelper(R_rdf::func1, new const char*[1]{"x1"}, 1, "", reinterpret_cast<std::weak_ptr<ROOT::Detail::RDF::RJittedFilter>*>(0x3469a140), reinterpret_cast<std::shared_ptr<ROOT::Detail::RDF::RNodeBase>*>(0x346a57b0),reinterpret_cast<ROOT::Internal::RDF::RColumnRegister*>(0x346b47b0));
ROOT::Internal::RDF::CallBuildAction<ROOT::Internal::RDF::ActionTags::Sum, float>(reinterpret_cast<std::shared_ptr<ROOT::Detail::RDF::RNodeBase>*>(0x346a54a0), new const char*[1]{"x1"}, 1, 1, reinterpret_cast<shared_ptr<double>*>(0x346a56d0), reinterpret_cast<std::weak_ptr<ROOT::Internal::RDF::RJittedAction>*>(0x346a53f0), reinterpret_cast<ROOT::Internal::RDF::RColumnRegister*>(0x346b4aa0));
```
These lines are repeated for every new RDataFrame.

Instead, when running the same program after this PR, we have
```
namespace R_rdf {
auto func0(){return 42.f
;}
using func0_ret_t = typename ROOT::TypeTraits::CallableTraits<decltype(func0)>::ret_type;
}
namespace R_rdf {
auto func1(const float var0){return var0 > 0.f
;}
using func1_ret_t = typename ROOT::TypeTraits::CallableTraits<decltype(func1)>::ret_type;
}
```

These are the same as before. Then we have
```
namespace R_rdf {
  void jitNodeRegistrator_0(const std::vector<std::string> &colNames, ROOT::Internal::RDF::RColumnRegister &colRegister, ROOT::Detail::RDF::RLoopManager &lm, void *jittedDefine, std::shared_ptr<void> *) {
   ROOT::Internal::RDF::JitDefineHelper<ROOT::Internal::RDF::DefineTypes::RDefineTag>(R_rdf::func0, colNames, colRegister, lm, reinterpret_cast<ROOT::Detail::RDF::RJittedDefine *>(jittedDefine));
}
namespace R_rdf {
  void jitNodeRegistrator_1(const std::vector<std::string> &colNames, ROOT::Internal::RDF::RColumnRegister &colRegister, ROOT::Detail::RDF::RLoopManager &lm, void *jittedFilter, std::shared_ptr<void> *) {
   ROOT::Internal::RDF::JitFilterHelper(R_rdf::func1, colNames, colRegister, lm, reinterpret_cast<ROOT::Detail::RDF::RJittedFilter*>(jittedFilter));
}
namespace R_rdf {
  void jitNodeRegistrator_2(const std::vector<std::string> &colNames, ROOT::Internal::RDF::RColumnRegister &colRegister, ROOT::Detail::RDF::RLoopManager &lm, void *jittedAction, std::shared_ptr<void> *helperArg) {
   ROOT::Internal::RDF::CallBuildAction<ROOT::Internal::RDF::ActionTags::Sum, float>(colNames, colRegister, lm, reinterpret_cast<ROOT::Internal::RDF::RJittedAction *>(jittedAction), 1, reinterpret_cast<std::shared_ptr<double> *>(helperArg));
}
```

That is, there is one JIT node registrator line per function passed by the user (or per type of action, e.g. in the case of `Sum`). Afterwards, there is no other call to the interpreter for JITting, not even in the next computation graph.


I have repeated the test with `n=1000` instead, to show the difference more evidently. I attach the extracted lines of code from each case:
```
$: cat extracted_master.txt | wc -l
4020
$: cat extracted_patch.txt | wc -l
35
```

[extracted_patch.txt](https://github.com/user-attachments/files/24647631/extracted_patch.txt)
[extracted_master.txt](https://github.com/user-attachments/files/24647632/extracted_master.txt)


And just for completeness, here is the full list of commands to get to those results
```
g++ $(root-config --cflags --libs) -g -o jit_large_graphs_patch.out jit_large_graphs.cpp
./jit_large_graphs_patch.out 2>&1 | tee jit_large_graphs_patch.txt
grep -v "^Info in <\[ROOT.RDF\] " jit_large_graphs_patch.txt > extracted_patch.txt
cat extracted_patch.txt | wc -l
```

PR #17288, opened 2024-12-17
[ntuple] allow RNTuple to emulate fields of unknown types via RecordFields

silverweed:
# This Pull request:
adds the option to "emulate" fields of unknown user-defined types (i.e. types for which we lack dictionaries) as RecordFields whose structure is reconstructed from the on-disk information. This is off by default but can be toggled via `RCreateModelOptions`.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
@@ -69,8 +69,9 @@ This is and can only be partially enforced through C++.
 */
 // clang-format on
 class RFieldBase {
-   friend class ROOT::Experimental::RClassField;                             // to mark members as artificial
-   friend class ROOT::Experimental::RNTupleJoinProcessor;                    // needs ConstuctValue
+   friend class ROOT::Experimental::RClassField;          // to mark members as artificial
+   friend class ROOT::Experimental::RNTupleJoinProcessor; // needs ConstuctValue
+   friend class ROOT::Experimental::RFieldDescriptor;
    friend struct ROOT::Experimental::Internal::RFieldCallbackInjector;       // used for unit tests
    friend struct ROOT::Experimental::Internal::RFieldRepresentationModifier; // used for unit tests
    friend void Internal::CallFlushColumnsOnField(RFieldBase &);
@@ -473,9 +474,14 @@ protected:
    /// normalized type name and type alias
    /// TODO(jalopezg): this overload may eventually be removed leaving only the `RFieldBase::Create()` that takes a
    /// single type name
-   static RResult<std::unique_ptr<RFieldBase>> Create(const std::string &fieldName, const std::string &canonicalType,
-                                                      const std::string &typeAlias,
-                                                      const RCreateFieldOptions &options = {});
+   static RResult<std::unique_ptr<RFieldBase>>
+   Create(const std::string &fieldName, const std::string &canonicalType, const std::string &typeAlias,
+          const RCreateFieldOptions &options = {}, const RNTupleDescriptor *desc = nullptr,
+          DescriptorId_t fieldId = kInvalidDescriptorId);
+
+   static RResult<std::unique_ptr<RFieldBase>> Create(const std::string &fieldName, const std::string &typeAlias,
+                                                      const RCreateFieldOptions &options, const RNTupleDescriptor *desc,
+                                                      DescriptorId_t fieldId);
 
 public:
    template <bool IsConstT>
@@ -511,7 +517,8 @@ public:
    std::unique_ptr<RFieldBase> Clone(std::string_view newName) const;
 
    /// Factory method to resurrect a field from the stored on-disk type information
-   static RResult<std::unique_ptr<RFieldBase>> Create(const std::string &fieldName, const std::string &typeName);
+   static RResult<std::unique_ptr<RFieldBase>>
+   Create(const std::string &fieldName, const std::string &typeName, const RCreateFieldOptions &options = {});
 
    /// Checks if the given type is supported by RNTuple. In case of success, the result vector is empty.
    /// Otherwise there is an error record for each failing sub field (sub type).
--- a/tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
@@ -42,11 +42,15 @@ struct RFieldCallbackInjector;
 struct RFieldRepresentationModifier;
 class RPageSink;
 class RPageSource;
-// TODO(jblomer): find a better way to not have these four methods in the RFieldBase public API
+// TODO(jblomer): find a better way to not have these methods in the RFieldBase public API
 void CallFlushColumnsOnField(RFieldBase &);
 void CallCommitClusterOnField(RFieldBase &);
 void CallConnectPageSinkOnField(RFieldBase &, RPageSink &, NTupleSize_t firstEntry = 0);
 void CallConnectPageSourceOnField(RFieldBase &, RPageSource &);
+ROOT::RResult<std::unique_ptr<ROOT::Experimental::RFieldBase>>
+CallFieldBaseCreate(const std::string &fieldName, const std::string &canonicalType, const std::string &typeAlias,
+                    const RCreateFieldOptions &options, const RNTupleDescriptor *desc, DescriptorId_t fieldId);
+
 } // namespace Internal
 
 namespace Detail {
@@ -70,14 +74,18 @@ This is and can only be partially enforced through C++.
 // clang-format on
 class RFieldBase {
    friend class ROOT::Experimental::RClassField;                             // to mark members as artificial
-   friend class ROOT::Experimental::RNTupleJoinProcessor;                    // needs ConstuctValue
-   friend class ROOT::Experimental::RFieldDescriptor;
+   friend class ROOT::Experimental::RNTupleJoinProcessor;                    // needs ConstructValue
    friend struct ROOT::Experimental::Internal::RFieldCallbackInjector;       // used for unit tests
    friend struct ROOT::Experimental::Internal::RFieldRepresentationModifier; // used for unit tests
    friend void Internal::CallFlushColumnsOnField(RFieldBase &);
    friend void Internal::CallCommitClusterOnField(RFieldBase &);
    friend void Internal::CallConnectPageSinkOnField(RFieldBase &, Internal::RPageSink &, NTupleSize_t);
    friend void Internal::CallConnectPageSourceOnField(RFieldBase &, Internal::RPageSource &);
+   friend ROOT::RResult<std::unique_ptr<ROOT::Experimental::RFieldBase>>
+   Internal::CallFieldBaseCreate(const std::string &fieldName, const std::string &canonicalType,
+                                 const std::string &typeAlias, const RCreateFieldOptions &options,
+                                 const RNTupleDescriptor *desc, DescriptorId_t fieldId);
+
    using ReadCallback_t = std::function<void(void *)>;
 
 protected:
@@ -480,7 +488,8 @@ protected:
           const RCreateFieldOptions &options = {}, const RNTupleDescriptor *desc = nullptr,
           DescriptorId_t fieldId = kInvalidDescriptorId);
 
-   /// Same as the above overload of Create, but infers the normalized type name and the canonical type name from `typeName`.
+   /// Same as the above overload of Create, but infers the normalized type name and the canonical type name from
+   /// `typeName`.
    static RResult<std::unique_ptr<RFieldBase>> Create(const std::string &fieldName, const std::string &typeName,
                                                       const RCreateFieldOptions &options, const RNTupleDescriptor *desc,
                                                       DescriptorId_t fieldId);
--- a/tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
@@ -529,7 +529,7 @@ public:
 
    /// Factory method to resurrect a field from the stored on-disk type information
    static RResult<std::unique_ptr<RFieldBase>>
-   Create(const std::string &fieldName, const std::string &typeName, const RCreateFieldOptions &options = {});
+   Create(const std::string &fieldName, const std::string &typeName);
 
    /// Checks if the given type is supported by RNTuple. In case of success, the result vector is empty.
    /// Otherwise there is an error record for each failing sub field (sub type).
--- a/tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
@@ -17,6 +17,7 @@
 #define ROOT7_RFieldBase
 
 #include <ROOT/RColumn.hxx>
+#include <ROOT/RCreateFieldOptions.hxx>
 #include <ROOT/RNTupleRange.hxx>
 #include <ROOT/RNTupleUtil.hxx>
 
@@ -473,7 +474,8 @@ protected:
    /// TODO(jalopezg): this overload may eventually be removed leaving only the `RFieldBase::Create()` that takes a
    /// single type name
    static RResult<std::unique_ptr<RFieldBase>> Create(const std::string &fieldName, const std::string &canonicalType,
-                                                      const std::string &typeAlias, bool continueOnError = false);
+                                                      const std::string &typeAlias,
+                                                      const RCreateFieldOptions &options = {});
 
 public:
    template <bool IsConstT>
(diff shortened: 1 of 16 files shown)


silverweed:
Unfortunately this approach doesn't work in all cases.

When replacing the InvalidFields we would also need to patch up stuff like fItemSize on the parent, which is brittle to do.
Probably the only robust way to go is to construct the field hierarchy bottom-up. To do so, my and Jakob's idea is:
- change RFieldBase::Create() so that it returns a "FieldGenerator" rather than a field. This is a functor that accepts a Descriptor for the field and returns a RFieldBase
- recursively construct a tree of Generators from RFieldBase::Create()
- from RFieldDescriptor::CreateField(), visit the tree of generators and replace all RInvalidField generators with the respective record field generators, then call the generators depth-first to construct the field tree.

pcanal:
I am confused.  Why do we need to emulate the case?  What is the usage pattern we are trying to emulated?  Isn't it enough to write the `RNTuple` in one process with the dictionary and read it in a different process without the dictionary?

silverweed:
@pcanal The use case is reading a RNTuple as best as possible when the dictionaries for some of the types stored are not available. For example we still want to be able to print out the schema of the RNTuple even when we have no clue about the specifics of the types stored.
Maybe the word "emulate" is not the best choice here, but the idea is adding the capability for RNTuple to reconstruct the type hierarchy without actually recreating the original classes, but "emulating" them with Record fields instead.

tree/ntuple/v7/test/ntuple_fork.hxx
```diff
@@ -0,0 +1,35 @@
+#pragma once
```
hahnjo, line 1: We generally don't use `#pragma once` in ROOT and RNTuple

tree/ntuple/v7/test/ntuple_evolution.cxx
```diff
@@ -1,62 +1,14 @@
 #include "ntuple_test.hxx"
 
+#define NTUPLE_FORK_MACROS
```
hahnjo, line 3: Eww, this is ugly. Can we just have all implementations in a header file that we can include into multiple tests?

tree/ntuple/v7/test/ntuple_fork.hxx
```diff
@@ -0,0 +1,35 @@
...
+
+// These helpers are split into an *Impl function and a macro to check for and propagate fatal failures to the caller.
+
+void ExecInForkImpl(std::function<void(void)> fn);
+
+void DeclarePointerImpl(std::string_view type, std::string_view variable, void *ptr);
+
+void ProcessLineImpl(const char *line);
+
+#ifdef NTUPLE_FORK_MACROS
```
hahnjo, line 14: Why do we need to protect these macro definitions? But see also my other comment if we can just have all implementations in a header file

tree/ntuple/v7/test/ntuple_evolution.cxx
```diff
@@ -32,7 +32,7 @@ TEST(RNTupleEvolution, AddedMember)
 {
    FileRaii fileGuard("test_ntuple_evolution_added_member.root");
 
-   WriteOldInFork([&] {
+   ExecInFork([&] {
```
hahnjo, line 35: These changes have to go into the commit that renamed the functions!

tree/ntuple/v7/test/ntuple_emulated.cxx
```diff
@@ -61,7 +61,7 @@ struct Outer {
       try {
          auto model = desc.CreateModel(opts);
          FAIL() << "Creating a model without fEmulateUnknownTypes should fail";
-      } catch (RException ex) {
+      } catch (ROOT::RException ex) {
```
hahnjo, line 64: has to go into the commit that added the test

tree/ntuple/v7/test/ntuple_evolution.cxx
```diff
@@ -579,7 +579,7 @@ TEST(RNTupleEvolution, AddedIntermediateClass)
 {
    FileRaii fileGuard("test_ntuple_evolution_added_intermediate_class.root");
 
-   WriteOldInFork([&] {
+   ExecInFork([&] {
```
hahnjo, line 582: has to go into the commit that renamed the functions

tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
```diff
@@ -473,9 +481,15 @@ protected:
    /// normalized type name and type alias
    /// TODO(jalopezg): this overload may eventually be removed leaving only the `RFieldBase::Create()` that takes a
    /// single type name
-   static RResult<std::unique_ptr<RFieldBase>> Create(const std::string &fieldName, const std::string &canonicalType,
-                                                      const std::string &typeAlias, bool continueOnError = false,
-                                                      bool emulateMissingTypes = false);
+   static RResult<std::unique_ptr<RFieldBase>>
+   Create(const std::string &fieldName, const std::string &canonicalType, const std::string &typeAlias,
+          const RCreateFieldOptions &options = {}, const RNTupleDescriptor *desc = nullptr,
+          DescriptorId_t fieldId = kTraitInvalidField);
```
hahnjo, line 487: This looks odd to me: Is `kTraitInvalidField` really (semantically) a `DescriptorId_t`?

tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
```diff
@@ -244,6 +245,13 @@ private:
       }
    }
 
+   void SetEmulatedFromType(std::string_view type)
+   {
+      assert(fType.empty());
+      fTraits |= kTraitEmulatedField;
+      fType = type;
+   }
+
```
hahnjo, line 254: Do we need this generically in `RFieldBase` or can this move to `RRecordField`? potentially as a constructor option or with a factory method (was we have `RVectorField::CreateUntyped()`)

tree/ntuple/v7/src/RFieldSequenceContainer.cxx
```diff
@@ -586,6 +586,7 @@ void ROOT::Experimental::RVectorField::RVectorDeleter::operator()(void *objPtr,
 {
    auto vecPtr = static_cast<std::vector<char> *>(objPtr);
    if (fItemDeleter) {
+      R__ASSERT(fItemSize > 0);
```
hahnjo, line 589: This looks unrelated to the emulation itself

tree/ntuple/v7/test/ntuple_emulated.cxx
```diff
@@ -0,0 +1,90 @@
...
+   auto reader = RNTupleReader::Open("ntpl", fileGuard.GetPath());
+   const auto &desc = reader->GetDescriptor();
+
+   {
+      RNTupleDescriptor::RCreateModelOptions opts;
+      opts.fEmulateUnknownTypes = false;
+      try {
+         auto model = desc.CreateModel(opts);
+         FAIL() << "Creating a model without fEmulateUnknownTypes should fail";
+      } catch (RException ex) {
```
hahnjo: change to `ROOT::RException` needs to move here from another commit, and the compiler complains:
```
catching polymorphic type ‘class ROOT::RException’ by value
```

tree/ntuple/v7/test/ntuple_emulated.cxx
```diff
@@ -94,3 +94,107 @@ struct Outer {
...
+   auto reader = RNTupleReader::Open("ntpl", fileGuard.GetPath());
+   const auto &desc = reader->GetDescriptor();
+
+   {
+      RNTupleDescriptor::RCreateModelOptions opts;
+      opts.fEmulateUnknownTypes = false;
+      try {
+         auto model = desc.CreateModel(opts);
+         FAIL() << "Creating a model without fEmulateUnknownTypes should fail";
+      } catch (ROOT::RException ex) {
```
hahnjo, line 156: The compiler complains
```
catching polymorphic type ‘class ROOT::RException’ by value
```

hahnjo (changes requested):
I left some initial comments. In general, I find this PR hard to follow because some implementations get changed multiple times and some commits introduce options but only half implement them. Can you maybe try to clean this up a bit? Also many of the intermediate commits will probably not compile / pass the tests...

tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
silverweed: No, this was an oversight: it was supposed to be kInvalidDescriptorId. Thanks for spotting it.

tree/ntuple/v7/src/RFieldSequenceContainer.cxx
silverweed: I found a crash while testing it and it didn't look like it was worth a separate PR (or even commit)...

tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
```diff
@@ -460,8 +474,14 @@ protected:
...
-   static RResult<std::unique_ptr<RFieldBase>> Create(const std::string &fieldName, const std::string &canonicalType,
-                                                      const std::string &typeAlias, bool continueOnError = false);
+   static RResult<std::unique_ptr<RFieldBase>>
+   Create(const std::string &fieldName, const std::string &canonicalType, const std::string &typeAlias,
+          const RCreateFieldOptions &options = {}, const RNTupleDescriptor *desc = nullptr,
+          DescriptorId_t fieldId = kInvalidDescriptorId);
+
+   static RResult<std::unique_ptr<RFieldBase>> Create(const std::string &fieldName, const std::string &typeAlias,
+                                                      const RCreateFieldOptions &options, const RNTupleDescriptor *desc,
+                                                      DescriptorId_t fieldId);
```
jblomer, line 484: I think we should update the method documentation for the new parameters

tree/ntuple/v7/src/RFieldBase.cxx
```diff
@@ -278,32 +289,40 @@ ROOT::Experimental::RFieldBase::Check(const std::string &fieldName, const std::s
...
+   RCreateFieldOptions cfOpts{};
+   cfOpts.fReturnInvalidOnError = true;
+   cfOpts.fEmulateUnknownTypes = false;
+   fieldZero.Attach(RFieldBase::Create(fieldName, canonicalType, typeAlias, cfOpts).Unwrap());
 
    std::vector<RCheckResult> result;
    for (const auto &f : fieldZero) {
-      auto invalidField = dynamic_cast<const RInvalidField *>(&f);
-      if (!invalidField)
+      bool isInvalidField = f.GetTraits() & RFieldBase::kTraitInvalidField;
```
jblomer, line 299:
```suggestion
      const bool isInvalidField = f.GetTraits() & RFieldBase::kTraitInvalidField;
```

tree/ntuple/v7/src/RFieldBase.cxx
```diff
@@ -314,11 +333,20 @@ ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::
 
    std::unique_ptr<ROOT::Experimental::RFieldBase> result;
 
+   const auto maybeGetChildId = [desc, fieldId](int childIdx) {
+      if (desc) {
+         const auto &fieldDesc = desc->GetFieldDescriptor(fieldId);
+         return fieldDesc.GetLinkIds()[childIdx];
```
jblomer, line 339: Let's access with bounds check, just in case the descriptor is corrupted.
```suggestion
         return fieldDesc.GetLinkIds().at(childIdx);
```

tree/ntuple/v7/src/RFieldBase.cxx
```diff
@@ -560,13 +599,28 @@ ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::
                   result = std::make_unique<RClassField>(fieldName, canonicalType);
                }
             }
+         } else if (options.fEmulateUnknownTypes) {
+            assert(desc);
```
jblomer, line 603: This option can be set through the public API and then we don't have a descriptor, no?

jblomer (commented):
Very nice! We should also add and test emulated collection proxies (probably in another PR).

tree/ntuple/v7/test/ntuple_evolution.cxx
hahnjo: still needs to go into the commit that renamed the function

silverweed: I was sure I moved this, no idea how did this slip through.

tree/ntuple/v7/src/RFieldBase.cxx
silverweed: With [e2c3074](https://github.com/root-project/root/pull/17288/commits/e2c3074bf4a2a3af219060f45f24ceb2da346dfa) the user cannot directly call `RFieldBase::Create()` with `fEmulateUnknownTypes == true`, so this code path is only taken when called by `RFieldDescriptor::CreateField` (which does pass down the descriptor)

tree/ntuple/v7/src/RFieldBase.cxx
```diff
@@ -264,11 +264,22 @@ std::string ROOT::Experimental::RFieldBase::GetQualifiedFieldName() const
...
 
 ROOT::RResult<std::unique_ptr<ROOT::Experimental::RFieldBase>>
-ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::string &typeName)
+ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::string &typeName,
+                                       const RCreateFieldOptions &options)
 {
    auto typeAlias = Internal::GetNormalizedTypeName(typeName);
    auto canonicalType = Internal::GetNormalizedTypeName(GetCanonicalTypeName(typeAlias));
-   return R__FORWARD_RESULT(RFieldBase::Create(fieldName, canonicalType, typeAlias));
+   return R__FORWARD_RESULT(RFieldBase::Create(fieldName, canonicalType, typeAlias, options));
```
hahnjo: Should this just forward to `Create(fieldName, typeName, options, nullptr, kInvalidDescriptorId)`?

tree/ntuple/v7/test/ntuple_emulated.cxx
```diff
@@ -93,3 +94,108 @@ struct Outer {
...
+   // To achieve this "forgetting" we use fork() so that the parent process doesn't know about the class.
+
+   FileRaii fileGuard("test_ntuple_emulated_fields_vecs.root");
+
+   ExecInFork([&] {
+      // The child process writes the file and exits, but the file must be preserved to be read by the parent.
+      fileGuard.PreserveFile();
+
+      ASSERT_TRUE(gInterpreter->Declare(R"(
+struct Inner {
```
hahnjo, line 110: You should uniquely prefix the `struct` names to prevent any caching in the parent process when reading the file.

tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
```diff
@@ -69,15 +69,18 @@ This is and can only be partially enforced through C++.
...
    friend struct ROOT::Experimental::Internal::RFieldCallbackInjector;       // used for unit tests
    friend struct ROOT::Experimental::Internal::RFieldRepresentationModifier; // used for unit tests
    friend void Internal::CallFlushColumnsOnField(RFieldBase &);
    friend void Internal::CallCommitClusterOnField(RFieldBase &);
    friend void Internal::CallConnectPageSinkOnField(RFieldBase &, Internal::RPageSink &, NTupleSize_t);
    friend void Internal::CallConnectPageSourceOnField(RFieldBase &, Internal::RPageSource &);
+   // for RFieldBase::Create
+   friend std::unique_ptr<ROOT::Experimental::RFieldBase>
+   ROOT::Experimental::RFieldDescriptor::CreateField(const RNTupleDescriptor &ntplDesc,
+                                                     const RCreateFieldOptions &options) const;
```
hahnjo: I think this `friend` declaration becomes redundant with the next commit

tree/ntuple/v7/test/ntuple_emulated.cxx
```diff
@@ -199,3 +199,152 @@ struct Outer {
...
+   // Test struct containing an empty struct
+
+   FileRaii fileGuard("test_ntuple_emulated_emptystruct.root");
+
+   ExecInFork([&] {
+      // The child process writes the file and exits, but the file must be preserved to be read by the parent.
+      fileGuard.PreserveFile();
+
+      ASSERT_TRUE(gInterpreter->Declare(R"(
+struct Inner {
```
hahnjo, line 214: Should also be prefixed, see comment on a previous commit

tree/ntuple/v7/src/RNTupleDescriptor.cxx
```diff
@@ -125,7 +125,7 @@ ROOT::Experimental::RFieldDescriptor::CreateField(const RNTupleDescriptor &ntplD
       // NOTE: Unwrap() here may throw an exception, hence the try block.
       // If options.fReturnInvalidOnError is false we just rethrow it, otherwise we return an InvalidField wrapping the
       // error.
-      auto field = RFieldBase::Create(fieldName, typeName, typeName, options, &ntplDesc, fFieldId).Unwrap();
+      auto field = Internal::CallFieldBaseCreate(fieldName, typeName, typeName, options, &ntplDesc, fFieldId).Unwrap();
```
hahnjo: I'm not sure, but doesn't that mean that the previous commit doesn't compile?

hahnjo (approved):
LG from my side; there are some stray lines in commit messages that look like the result of squashing, and one comment about a non-compiling commit (not sure)

PR #17289, opened 2024-12-17
[ntuple] allow passing RCreateModelOptions to RNTupleReader::Open

silverweed:
# This Pull request:
adds two overloads to `RNTupleReader::Open` that allow passing `RCreateModelOptions` to it. These options will be used by the internal `CreateModel` calls.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/src/RNTupleReader.cxx
+++ b/tree/ntuple/v7/src/RNTupleReader.cxx
@@ -112,10 +112,32 @@ ROOT::Experimental::RNTupleReader::Open(std::unique_ptr<RNTupleModel> model, con
       new RNTupleReader(std::move(model), Internal::RPageSourceFile::CreateFromAnchor(ntuple, options), options));
 }
 
+std::unique_ptr<ROOT::Experimental::RNTupleReader>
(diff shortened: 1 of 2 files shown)


tree/ntuple/v7/src/RNTupleReader.cxx
```diff
@@ -140,7 +162,8 @@ void ROOT::Experimental::RNTupleReader::PrintInfo(const ENTupleInfo what, std::o
       {
          auto descriptorGuard = fSource->GetSharedDescriptorGuard();
          name = descriptorGuard->GetName();
-         fullModel = descriptorGuard->CreateModel();
+         fullModel =
+            descriptorGuard->CreateModel(fCreateModelOptions.value_or(RNTupleDescriptor::RCreateModelOptions{}));
```
jblomer, line 166: Perhaps we want to go even a step further here: For printing we can always use emulated classes, so that you can always look at the dataset.

jblomer (approved):
LGTM, one comment to consider.

tree/ntuple/v7/src/RNTupleReader.cxx
silverweed: Ok, I'll keep this in mind for the PR that adds the emulated classes

PR #17290, opened 2024-12-17
[ntuple] Improve `RNTupleProcessor` entry number bookkeeping

enirolf:
This PR improves the bookkeeping of entries in the `RNTupleProcessor`. Most importantly, a clearer distinction is made between the current entry number (relative to the currently open page source) and the number of entries processed. There was an incorrectness where this value was equal, whereas (assuming we linearly process one page source) `fNEntriesProcessed` should be higher than `fLocalEntryNumber` (e.g., after loading the first entry, `fNEntriesProcessed == 1`, and `fLocalEntryNumber == 0`).

In the process, some variables have been renamed to better capture their meaning.

As a consequence of this refactoring, the `RNTupleSingleProcessor` can now also handle empty ntuples.

Related to #17132.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
@@ -202,7 +202,7 @@ public:
       iterator operator++(int)
       {
          auto obj = *this;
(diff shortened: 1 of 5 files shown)


jblomer (approved):
LGTM

PR #17292, opened 2024-12-17
[RF] Fix parameter ordering bug in RooFormulaArgStreamer

### LINKED ISSUE #17291: [RF] Parameter ordering bug in RooFormulaArgStreamer
### Check duplicate issues.

- [X] Checked for duplicates

### Description

I noticed that if the RooFormulaArgStreamer is used with a RooFormulaVar that is built with the syntax like:
`RooFormulaVar("name","title","@1+@2+@3","RooArgList(arg1,arg2,arg2)")`
then in the case there are more than 10 arguments it can happen that the exported JSON has the wrong parameter names. I believe this is because the logic at 548-568 in https://root.cern.ch/doc/v634/JSONFactories__RooFitCore_8cxx_source.html assumes that the vector `paramsWithIndex` is ordered by the integer, while the sort actually orders on the pointers in the vector. In the case where the order of the pointers doesn't match to integer, this can cause the ReplaceAll at 567 to act more than once on the same parameter. Below is a minimal bit of code reproducing the problem.

I think it is fixed by removing the `vector<std::pair<RooAbsArg *, std::size_t>>` and just directly looping (in reverse order) on the parameters of the formula. I created a pull request to fix.

### Reproducer

Put the following code in a macro and run as `python macro.C`:

```
import ROOT
import random

# create a large number of variables
nvar = 67
vs = []
for i in range(0, nvar):
    v = ROOT.RooRealVar("v" + str(i), "v" + str(i), 0.0, -5.0, 5.0)
    vs.append(v)

# shuffle the order - if you comment this the code works fine
random.seed(871)
random.shuffle(vs)

# make a formula var that is product of the variables
thestring = ""
arglist = ROOT.RooArgList()
for i, v in enumerate(vs):
    thestring = thestring + "@" + str(i) + "*"
    arglist.add(v)

thestring = thestring.rstrip('*')
fv = ROOT.RooFormulaVar("fv", "fv", thestring, arglist)

# make gaussian with mean as product of the variables and import into workspace
x = ROOT.RooRealVar("x","x",0.0,-5.0,5.0)
g = ROOT.RooGaussian("g","g",x, fv, ROOT.RooRealConstant.value(1.0))
ws = ROOT.RooWorkspace("ws")
ws.Import(g)
print(fv.expression())

# export to json
t = ROOT.RooJSONFactoryWSTool(ws)
t.exportJSON('fv.json')

# try to import, crashes because the export was bad
newws = ROOT.RooWorkspace("newws")
t2 = ROOT.RooJSONFactoryWSTool(newws)
t2.importJSON('fv.json')
print(newws.obj('fv').expression())
```

### ROOT version

/cvmfs/atlas.cern.ch/repo/sw/software/0.4/StatAnalysis/0.4.1/InstallArea/x86_64-el9-gcc13-opt/bin/root

6.32.02

### Installation method

CVMFS installation

### Operating system

Linux

### Additional context

_No response_

mark-owen:
# This Pull request fixes:

The parameter ordering bug in RooFormulaArgStreamer. I changed the code to remove the `vector< pair<RooAbsArg*, size_t> >` and just directly loop in reverse order over the parameters of the formula.

## Checklist:

- [x] tested changes locally

This PR fixes #17291

### DIFF
--- a/roofit/hs3/src/JSONFactories_RooFitCore.cxx
+++ b/roofit/hs3/src/JSONFactories_RooFitCore.cxx
@@ -545,16 +545,16 @@ public:
       const RooArg_t *pdf = static_cast<const RooArg_t *>(func);
       elem["type"] << key();
       TString expression(pdf->expression());
-      std::vector<std::pair<RooAbsArg *, std::size_t>> paramsWithIndex;
+      std::vector<std::pair<std::size_t, const RooAbsArg*>> paramsWithIndex;
       paramsWithIndex.reserve(pdf->nParameters());
       for (size_t i = 0; i < pdf->nParameters(); ++i) {
-         paramsWithIndex.emplace_back(pdf->getParameter(i), i);
+	  paramsWithIndex.emplace_back(i, pdf->getParameter(i));
       }
       std::sort(paramsWithIndex.begin(), paramsWithIndex.end());
       // If the tokens follow the "x[#]" convention, the square braces enclosing each number
       // ensures that there is a unique mapping between the token and parameter name
-      for (auto [par, idx] : paramsWithIndex) {
-         expression.ReplaceAll(("x[" + std::to_string(idx) + "]").c_str(), par->GetName());
+      for (auto [idx, par] : paramsWithIndex) {
+	  expression.ReplaceAll(("x[" + std::to_string(idx) + "]").c_str(), par->GetName());
       }
       // If the tokens follow the "@#" convention, the numbers are not enclosed by braces.
       // So there may be tokens with numbers whose lower place value forms a subset string of ones with a higher place value,
@@ -562,8 +562,8 @@ public:
       // So the names of these parameters must be applied descending from the highest place value
       // in order to ensure each parameter name is uniquely applied to its token.
       for (auto it = paramsWithIndex.rbegin(); it != paramsWithIndex.rend(); ++it) {
-         RooAbsArg* par = it->first;
-         std::size_t idx = it->second;
+         const RooAbsArg* par = it->second;
+         std::size_t idx = it->first;
          expression.ReplaceAll(("@" + std::to_string(idx)).c_str(), par->GetName());
       }
       elem["expression"] << expression.Data();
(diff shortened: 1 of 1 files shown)


roofit/hs3/src/JSONFactories_RooFitCore.cxx
```diff
@@ -545,25 +545,25 @@ class RooFormulaArgStreamer : public RooFit::JSONIO::Exporter {
       const RooArg_t *pdf = static_cast<const RooArg_t *>(func);
       elem["type"] << key();
       TString expression(pdf->expression());
-      std::vector<std::pair<RooAbsArg *, std::size_t>> paramsWithIndex;
+      std::vector<std::pair<std::size_t, const RooAbsArg*>> paramsWithIndex;
```
guitargeek, line 548:
```suggestion
      std::vector<const RooAbsArg*> paramsWithIndex;
```
As you suggested it now, the vector is now sorted by the index (see one of the following lines):
```txt
std::sort(paramsWithIndex.begin(), paramsWithIndex.end());
```
However, in this case the information of the index is redundant. The sort does nothing at all. The original code was badly designed.

Could you please refactor the code such that this redundant index is not stored and there is no `std::sort`? That means that the loops will then look something like this:
```c++
      for (std::size_t i = 0; i < params.size(); ++i) {
          RooAbsArg const *par = params[idx];
	  expression.ReplaceAll(("x[" + std::to_string(idx) + "]").c_str(), par->GetName());
      }
```

guitargeek (changes requested):
Hi @mark-owen, thank you so much for this contribution!

Please make sure that the code is formatted with our clang-format style:
https://github.com/root-project/root/blob/master/.clang-format

Also, your changes made it apparent that the index information in the vector is redundant. Can you maybe change the code such that storing the index and sorting by it is avoided? Thanks.

mark-owen:
Thanks @guitargeek - I see the point, the code can be simplified a bit (actually not sure the vector is needed at all). I'll update this in the next days (and look at clang-format).

guitargeek:
Thank you!

roofit/hs3/src/JSONFactories_RooFitCore.cxx
mark-owen: I realised even the vector is not really needed, so everything is now simplified to run in one (reverse order) loop.

guitargeek (approved):
Thank you so much for the fix! The fixes that also simplify the code are my favorite kind of fixes :slightly_smiling_face:

PR #17293, opened 2024-12-17
[RF] Manually create hist2workspace documentation

guitargeek:
The documentation for `hist2workspace` is basically never changed, so the overhead of reproducing the man file and help string in the CMake configuration step is not really justified.

If particular, this causes problems when building RooFit standalone: the code generation requires a script that is internal to the ROOT build (argparse2help.py).


bellenot (approved):
LGTM

PR #17294, opened 2024-12-18
[RF] Follow up on RooDataSet constructor deprecation

guitargeek:
Follow up on cc99d62e9 by avoiding use of the deprecated constructor and avoiding an unnecessary warning that happens at ROOT build time.


guitargeek:
Commits merged together with #17264.

PR #17299, opened 2024-12-18
[ntuple] add TFileMerger options to mirror RNTupleMergeOptions

silverweed:
- [x] tested changes locally


tree/ntuple/v7/inc/ROOT/RNTupleMerger.hxx
```diff
@@ -55,6 +55,15 @@ struct RSealedPageMergeData;
 
 class RClusterPool;
 
+/// Set of merging options to pass to RNTupleMerger.
+/// If you're using the merger through TFileMerger you need to give it string-based options instead.
+/// Here is the mapping for the TFileMerger options:
```
pcanal, line 60: This information should also be added to the release notes.

main/src/hadd.cxx
```diff
@@ -427,10 +427,10 @@ int main( int argc, char **argv )
          else
             newcomp = ROOT::RCompressionSetting::EDefaults::kUseCompiledDefault;
          delete firstInput;
-         fileMerger.SetMergeOptions(TString("first_source_compression"));
+         fileMerger.SetMergeOptions(TString("rnt:FirstSrcCompression"));
```
pcanal, line 430: Why does it need to be handled differently than the 'same' option for `TTree`?

silverweed: I'm not sure how TTree handles the case, but my reasoning is that, since RNTuple has a different default compression than TFile/TTree, we want to distinguish the cases of "use default compression" (user didn't pass any flag to hadd) and that of "use first file compression" (user passed -ff or -fk).
As far as I know, we have no way of telling which case are we in from RNTuple::Merge, so we rely on explicit merge options for that.

Do you think there is a better way of handling it? I might be missing something that would let us avoid these options.

jblomer (commented):
In principle looks good to me, but see also Philippe's comments.

main/src/hadd.cxx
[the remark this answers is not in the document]
silverweed: I'm fine with the dot instead of colon, and I guess `rntuple` can be a bit more informative, but I like the capitalized first letter better because it makes for a cleaner mapping with the root naming style: when mapping options directly to fields of a struct you simply have to drop the `f`, and when you map to enum names you simply drop the `k` (e.g. `MergingMode=Union`).
That said, I have no strong opinion except that we should be coherent across all our API.

pcanal: In this case, there is no good reason to make this option being 'RNTuple' only.  The fact that `TTree` is currently cumbersomely doing an approximate implementation is interesting but not relevant (IMHO).

Currently `TTree` uses the first file's global compression setting to set the compression to use either:
* This is the right choice, then for symmetry reason it should be the same choice for RNTuple (and if RNTuple wants/needs more, we may need a new option 'use the compression of the first RNTuple/TTree').
* This is not the right choice, then TTree should be updated to switch behavior when seeing the flag to be like RNTUple and use the compression of the first TTree.

main/src/hadd.cxx
```diff
@@ -427,10 +427,10 @@ int main( int argc, char **argv )
...
       } else {
          newcomp = ROOT::RCompressionSetting::EDefaults::kUseCompiledDefault;
-         fileMerger.SetMergeOptions(TString("default_compression"));
+         fileMerger.SetMergeOptions(TString("rnt:DefaultCompression"));
```
pcanal, line 433: That options is also not `RNTuple` specific.  The user requested the 'default' compression, this is true through-out.  The fact that everything else, 'just' setting the request compression to the global default means that until now pushing this information down was redundant but now it is not redundant anymore (no way to know if `newcomp` is set to the default by happenstance or intentionally).  The fact that only `RNTuple` will (for now?) use this information does not make it `RNTuple` specific.

tree/ntuple/v7/test/ntuple_merger.cxx
```diff
@@ -1479,3 +1481,92 @@ INSTANTIATE_TEST_SUITE_P(Seq, RNTupleMergerCheckEncoding,
...
+      // Now Merge the inputs
+      // We expect this to fail in Filter and Strict mode since the fields between the sources do NOT match
+      {
+         auto nt1 = std::unique_ptr<TFile>(TFile::Open(fileGuard1.GetPath().c_str()));
+         auto nt2 = std::unique_ptr<TFile>(TFile::Open(fileGuard2.GetPath().c_str()));
+         TFileMerger fileMerger(kFALSE, kFALSE);
+         fileMerger.OutputFile(fileGuard3.GetPath().c_str(), "RECREATE");
+         fileMerger.AddFile(nt1.get());
+         fileMerger.AddFile(nt2.get());
+         fileMerger.SetMergeOptions(TString("rnt:MergingMode=Filter"));
```
pcanal, line 1533: For this option (`rntuple:MergingMode`), I would agree that it is specific to `RNTuple` has the list of `mode` would (likely?) be different for different types. (And even the selection might be different).

main/src/hadd.cxx
pcanal: Sounds good.  I think we can have the "update the TTree behavior to use the first TTree compression when the flag is present" to be a subsequent PR.

silverweed:
@pcanal @jblomer  is there still something left blocking this PR?

jblomer (approved):
LGTM

main/src/hadd.cxx
pcanal: Did we open an issue to keep track of the needed modification on the TTree side?

tree/ntuple/v7/test/ntuple_merger.cxx
pcanal: What was the resolution?

main/src/hadd.cxx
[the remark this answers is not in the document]
pcanal: What was the resolution?

silverweed: Thanks for the reminder -> https://github.com/root-project/root/issues/17438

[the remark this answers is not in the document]
silverweed: The prefix has been changed to `"rntuple"` and the colon is now a dot. The capitalization is unchanged.

tree/ntuple/v7/test/ntuple_merger.cxx
pcanal: > Did this need a resolution?

Yes, in my opinion, all conversation needs some sort of resolution.  Even it is "just" 'acknowledged'.  In this case the resolution is indeed:

> Currently the option has been added as "rntuple.MergingMode"

main/src/hadd.cxx
pcanal: Thanks.

PR #17300, opened 2024-12-18
[ntuple] Exporter: add a way to blacklist/whitelist column types

silverweed:
## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
+++ b/tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
@@ -30,10 +30,16 @@ class RPageSource;
 class RNTupleExporter {
 public:
    enum class EFilterType {
-      /// Skip all columns whose type is contained in `fFilter`
+      /// Don't export items contained in the filter's set
       kBlacklist,
-      /// Only include columns whose type is contained in `fFilter`
+      /// Export only items contained in the filter's set
       kWhitelist   
+   };
+
+   template <typename T>
+   struct RFilter {
+      std::unordered_set<T> fSet;
+      EFilterType fType = EFilterType::kBlacklist;
    };
    
    struct RPagesOptions {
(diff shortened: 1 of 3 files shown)


tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
```diff
@@ -39,6 +49,12 @@ public:
       std::string fOutputPath;
       std::uint64_t fFlags;
 
+      /// Optional filter that determines which columns are included or excluded from being exported.
+      /// By default, export all columns. If you only want to include certain column types, add them
+      /// to this set and change fFilterType to kWhitelist.
```
pcanal, line 54: I am guesstimating that this is *not* the common cases.  Rather than filtering on the data type, I expect the user to be want to filter/select in allmost all cases on the column names.

silverweed: Well, it is the only case I found so far (as the only user so far :) and for the ntupleutils I think our strategy is "implement things as they are requested". 
However I do think I should change the name to a more specific one (`fColumnTypeFilter`) to avoid name clashing in the future.

jblomer (approved):
LGTM

tree/ntupleutil/v7/inc/ROOT/RNTupleExporter.hxx
pcanal: ~~Trimming (and skimming) data files is an essential and extremely common operation in our field.  Despite the improvement in file size for `RNTuple`, I don't see changing any time soon and I assume users would assume the feature will will soon be present.~~

Nevermind, I was confused.  This code change is in the `Exporter` (which is an internal research/exploration tools for `RNTuple` and where this behavior actually makes more sense) instead of the `RNTupleMerger`

pcanal (approved):
LGTM

tree/ntupleutil/v7/src/RNTupleExporter.cxx
```diff
@@ -46,13 +46,27 @@ struct RColumnExportInfo {
...
+};
+
+RAddColumnsResult AddColumnsFromField(std::vector<RColumnExportInfo> &vec, const RNTupleDescriptor &desc,
+                                      const RFieldDescriptor &fieldDesc, const std::unordered_set<EColumnType> &filter,
+                                      RNTupleExporter::EFilterType filterType)
 {
    R__LOG_DEBUG(1, RNTupleExporterLog()) << "processing field \"" << desc.GetQualifiedFieldName(fieldDesc.GetId())
                                          << "\"";
 
+   RAddColumnsResult res{};
```
hahnjo, line 73: Took me a bit to realize that this actually zero-initializes the members (`RAddColumnsResult res;` would not!). Can we maybe add explicit member-initialization in `RAddColumnsResult`?

PR #17301, opened 2024-12-18
[tree] Avoid rebuilding the TChainIndex in every MT task

vepadulano:
Currently, in the creation of the friends to attach to the main chain to be processed in a MT task, if a friend was attached to a TVirtualIndex-derived instance, this needs to be rebuilt. Evidence in the past showed that just cloning was not enough because the cloned index could not appropriately probe the task-local friend chain afterwards due to a misconfigured connection with the branches. A call to `LoadTree(0)` before connecting the chain to the cloned index makes sure that the dataset schema is loaded and the probing can happen succesfully. In turn, this optimizes the task creation by avoiding the rebuilding of the chain index for every task.

FYI @TomasDado


pcanal (approved):
LGTM.

PR #17302, opened 2024-12-18
[core] fix non-proportional color resampling

ferdymercury:
# This Pull request:

## Changes or fixes:

The quotient of extremes 65535/255 was taken as calibration factor, but this is not a proper way of resampling integer variables, that only works well for doubles.

Related: https://root-forum.cern.ch/t/understanding-root-color-indexes/62464/30

There is this debug comment (still on master) that might be related:
https://github.com/root-project/root/commit/a09e2869bdec00bf1342b0c370e60316f9bb2da9#diff-479b410550ebac192a78a3adb39fa0c9b7bfdd162b92670b863219da062eeb80R1210

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/core/base/src/TColor.cxx
+++ b/core/base/src/TColor.cxx
@@ -2484,9 +2484,9 @@ ULong_t TColor::RGB2Pixel(Int_t r, Int_t g, Int_t b)
    if (b > 255) b = 255;
 
    ColorStruct_t color;
-   color.fRed   = UShort_t(r * 257);  // 65535/255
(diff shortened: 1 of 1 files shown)


couet:
@ferdymercury Thanks to have seen that. 257 is weird indeed. I am will check this methid history to see if there some reason for this value.

couet (commented):
It looks good to me.

couet (approved):

PR #17306, opened 2024-12-19
[ntuple] set NEntries in RPagePersistentSink::InitFromDescriptor

silverweed:
Also add a bunch of `reserve()` where appropriate

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/src/RNTupleDescriptor.cxx
+++ b/tree/ntuple/v7/src/RNTupleDescriptor.cxx
@@ -595,11 +595,9 @@ ROOT::Experimental::RNTupleDescriptor::AddClusterGroupDetails(DescriptorId_t clu
          return R__FAIL("invalid attempt to re-populate existing cluster");
       }
    }
-   std::sort(clusterIds.begin(), clusterIds.end(),
-             [this](DescriptorId_t a, DescriptorId_t b)
-             {
-                return fClusterDescriptors[a].GetFirstEntryIndex() < fClusterDescriptors[b].GetFirstEntryIndex();
-             });
+   std::sort(clusterIds.begin(), clusterIds.end(), [this](DescriptorId_t a, DescriptorId_t b) {
+      return fClusterDescriptors[a].GetFirstEntryIndex() < fClusterDescriptors[b].GetFirstEntryIndex();
+   });
    auto cgBuilder = Internal::RClusterGroupDescriptorBuilder::FromSummary(iter->second);
    cgBuilder.AddSortedClusters(clusterIds);
    iter->second = cgBuilder.MoveDescriptor().Unwrap();
@@ -949,10 +947,9 @@ ROOT::Experimental::RNTupleDescriptor ROOT::Experimental::Internal::RNTupleDescr
    for (const auto &[id, _] : fDescriptor.fClusterGroupDescriptors)
       fDescriptor.fSortedClusterGroupIds.emplace_back(id);
    std::sort(fDescriptor.fSortedClusterGroupIds.begin(), fDescriptor.fSortedClusterGroupIds.end(),
-             [this](DescriptorId_t a, DescriptorId_t b)
-             {
-               return fDescriptor.fClusterGroupDescriptors[a].GetMinEntry() <
-                      fDescriptor.fClusterGroupDescriptors[b].GetMinEntry();
+             [this](DescriptorId_t a, DescriptorId_t b) {
+                return fDescriptor.fClusterGroupDescriptors[a].GetMinEntry() <
+                       fDescriptor.fClusterGroupDescriptors[b].GetMinEntry();
              });
    RNTupleDescriptor result;
    std::swap(result, fDescriptor);
--- a/tree/ntuple/v7/src/RPageStorage.cxx
+++ b/tree/ntuple/v7/src/RPageStorage.cxx
@@ -257,6 +257,7 @@ void ROOT::Experimental::Internal::RPageSource::UnzipClusterImpl(RCluster *clust
          continue;
       const auto &columnInfos = fActivePhysicalColumns.GetColumnInfos(columnId);
 
+      allElements.reserve(allElements.size() + columnInfos.size());
       for (const auto &info : columnInfos) {
          allElements.emplace_back(GenerateColumnElement(info.fElementId));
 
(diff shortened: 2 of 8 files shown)


jblomer (approved):
LGTM

tree/ntuple/v7/src/RPageStorage.cxx
```diff
@@ -943,6 +948,9 @@ void ROOT::Experimental::Internal::RPagePersistentSink::InitFromDescriptor(const
       }
       fDescriptorBuilder.AddCluster(clusterBuilder.MoveDescriptor().Unwrap());
       fPrevClusterNEntries += nEntries;
+      // Even though the number of entries are set when we call AddClusterGroup, we want them to be correct already
+      // at the end of this function.
+      fDescriptorBuilder.AddNEntries(nEntries);
```
hahnjo, line 953: Can you comment on why we want this? This now makes `InitFromDescriptor` different from normal descriptor building where `NEntries` is only updated when adding a cluster group. This is also documented in the comment of `RNTupleDescriptor::fNEntries`, which would need at least some updating with this change. More generally, for knowing the number of "entries in flight", we can just look at `fPrevClusterNEntries`, no?

hahnjo (changes requested):
(also for the record, I'm getting less sure about all those `reserve`s sprinkled in; at least for me, they really decrease code readability for oftentimes questionable benefit...)

tree/ntuple/v7/src/RPageStorage.cxx
silverweed: The reason has to do with support for incremental merging with late model extension: currently if you open an existing RNTuple with the RNTupleMerger and you try to late-model extend it, it will call `RPageSink::UpdateSchema` passing the number of entries _processed so far_, which doesn't include the number of previously-existing entries in the destination. 

In theory this information is already stored in `fPrevClusterNEntries` when initting the Sink from the descriptor, but it's not accessible. I figured it would make more sense from a user standpoint to have `GetNEntries()` return the correct number of already-existing entries immediately after initting the descriptor, rather than creating a separate getter specifically for the Sink (which would only make sense for the PersistentSink at that).

hahnjo: As discussed offline, changing `fNEntries` in normal descriptor building earlier than the cluster group may give us problems later on with sharded clusters. We could also ignore them for now and worry later.

Looking at this again, I think we have two other options:
1. We could create a separate getter in the sink. After all, incremental merging was the use case I added `RPageSink::GetDescriptor()` for, which returns an empty descriptor for example for `RPageNullSink`. I think adding another one would be acceptable.
2. In `InitFromDescriptor`, we could in principle commit a cluster group at the end, which would update `fNEntries` in the descriptor. When I added support for incremental merging, we decided to fully flatten out all cluster groups in the input - we could potentially revisit this.

tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -1405,6 +1405,8 @@ public:
    RResult<void> AddClusterGroup(RClusterGroupDescriptor &&clusterGroup);
    RResult<void> AddCluster(RClusterDescriptor &&clusterDesc);
 
+   RResult<void> AddNEntries(std::uint64_t nEntries);
+
```
hahnjo, line 1409: This method is not needed

tree/ntuple/v7/src/RNTupleDescriptor.cxx
```diff
@@ -595,11 +595,9 @@ ROOT::Experimental::RNTupleDescriptor::AddClusterGroupDetails(DescriptorId_t clu
...
       }
    }
-   std::sort(clusterIds.begin(), clusterIds.end(),
-             [this](DescriptorId_t a, DescriptorId_t b)
-             {
-                return fClusterDescriptors[a].GetFirstEntryIndex() < fClusterDescriptors[b].GetFirstEntryIndex();
-             });
+   std::sort(clusterIds.begin(), clusterIds.end(), [this](DescriptorId_t a, DescriptorId_t b) {
+      return fClusterDescriptors[a].GetFirstEntryIndex() < fClusterDescriptors[b].GetFirstEntryIndex();
+   });
```
hahnjo, line 600: Can you please remove, or at least separate out, unrelated changes?

tree/ntuple/v7/src/RNTupleDescriptor.cxx
```diff
@@ -1173,6 +1170,12 @@ ROOT::Experimental::Internal::RNTupleDescriptorBuilder::AddClusterGroup(RCluster
    return RResult<void>::Success();
 }
 
+ROOT::RResult<void> ROOT::Experimental::Internal::RNTupleDescriptorBuilder::AddNEntries(std::uint64_t nEntries)
+{
+   fDescriptor.fNEntries += nEntries;
+   return RResult<void>::Success();
+}
+
```
hahnjo, line 1178: not needed

tree/ntuple/v7/src/RPageStorage.cxx
```diff
@@ -857,6 +858,8 @@ void ROOT::Experimental::Internal::RPagePersistentSink::UpdateSchema(const RNTup
    }
 
    const auto nColumns = descriptor.GetNPhysicalColumns();
+   fOpenColumnRanges.reserve(fOpenColumnRanges.size() + (nColumns - nColumnsBeforeUpdate));
+   fOpenPageRanges.reserve(fOpenPageRanges.size() + (nColumns - nColumnsBeforeUpdate));
```
hahnjo, line 862: Do these really provide any measurable improvement? IMHO it should at least be a separate commit, maybe even a separate PR

hahnjo (changes requested):

tree/ntuple/v7/src/RPageStorage.cxx
silverweed: Reserving an array when the size is known a priori is just the correct thing to do and there should be a very good reason *not* to do so, rather than having to justify doing it.  I don't agree with the fact that these hinder readability. Otherwise, for the same reasoning, we should provide measurements every time we choose to pass a `const std::string &` rather than a `std::string` and similar.

hahnjo: We can agree to disagree. It's an extra line with an argument that you need to use brain power to understand and hopefully get right, where in many cases it doesn't buy you anything measurable. I think the comparison to passing by reference doesn't make sense because that one a) is the same line, so it's not extra, and b) there is a semantic difference between passing a const-ref and creating a copy.

PR #17307, opened 2024-12-19
Fix compilation of TMVA SOFIE parser on s390x

ellert:
# This Pull request:

## Changes or fixes:

There were changes to the code but the big endian code was not updated to match the changes in the little endian code.

## Checklist:

- [x] tested changes locally


lmoneta (approved):
LGTM, Thanks for the fix

PR #17309, opened 2024-12-19
Disable run_h1analysis.C on Windows

bellenot:
Disable `run_h1analysis.C` because of Endpoint Security HTTP traffic scanning, which is corrupting the data on Windows

### DIFF
--- a/tutorials/CMakeLists.txt
+++ b/tutorials/CMakeLists.txt
@@ -421,6 +421,12 @@ set(extra_veto
   visualisation/eve7/*.C
   math/r/rootlogon.C)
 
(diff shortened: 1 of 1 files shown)


dpiparo (approved):

pcanal:
Is this a temporary measure or should we explicitly error out whenever the user is using `http` to read a ROOT file?

bellenot:
> Is this a temporary measure or should we explicitly error out whenever the user is using `http` to read a ROOT file?

If users have `Endpoint Security` , they can still disable the `HTTP traffic scanning` option on their machine...

PR #17312, opened 2024-12-19
[DF] add validation for number of column names imposed to CSV source

silverweed:
# This Pull request:
adds a check that the number of items in the imposed column vector matches the number of columns in the CSV file. This avoid accessing `fHeaders` out of bounds and increases the robustness of the interface.
Also adding a test to `datasource_csv` and making sure all tests are run when needed (a couple of tests were mistakenly under the IMT ifdef guards)

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/dataframe/test/datasource_csv.cxx
+++ b/tree/dataframe/test/datasource_csv.cxx
@@ -8,21 +8,21 @@
 
 using namespace ROOT::RDF;
 
-auto fileName0 = "RCsvDS_test_headers.csv";
-auto fileName1 = "RCsvDS_test_noheaders.csv";
-auto fileName2 = "RCsvDS_test_empty.csv";
-auto fileName3 = "RCsvDS_test_win.csv";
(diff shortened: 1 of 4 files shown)


tree/dataframe/test/datasource_csv.cxx
```diff
@@ -400,9 +401,11 @@ TEST(RCsvDS, NaNTypeIndentification)
 
 TEST(RCsvDS, NanWarningChecks)
 {
+#ifdef R__USE_IMT
    ROOT::DisableImplicitMT(); // to allow usage of display
+#endif
```
vepadulano, line 406: This was previously necessary because the IMT section was written before this section. Now when this test is being hit the implict MT has not been activated yet so this call is redundant

vepadulano (approved):
Changes LGTM, before merging make sure the CI is green, I believe the current failure is simply due to the rearrangement of the MT tests.

PR #17313, opened 2024-12-19
[ntuple][python] Generalise getting and setting values of an REntry

vepadulano:
The value held by an REntry for a particular field is retrieved as a shared_ptr to the underlying data. The actual type of the cppyy proxy depends on the type of the pointee. For most types, the Python proxy is exposed as a the type of the pointee: a `std::shared_ptr<std::vector>` is exposed as a `cppyy.gbl.std.vector`. For fundamental types, this does not happen: a `std::shared_ptr<int>` is exposed as `cppyy.gbl.std.shared_ptr[int]`. In the first case, we can retrieve the actual smart pointer via the `__smartptr__` method of the cppyy proxy. In the latter case, this method will return None.

For reading, it's sufficient to retrieve the smart pointer and return a read-only reference to the pointee via the `__deref__` method.

For writing, there is no automatic support for assignment to the pointee. Thus, a utility helper is introduced in C++. Since this is a very specific case, the helper is only introduced in the RNTuple library.

The test of the Python API is extended with a field of non-fundamental type.

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
@@ -24,14 +24,24 @@ def _REntry_GetPtr(self, key):
 
 
 def _REntry_getitem(self, key):
-    ptr = self.GetPtr(key)
-    return ptr.get()[0]
+    ptr_proxy = self.GetPtr(key)
+    # Most types held by a shared_ptr are exposed by cppyy as the type of the
+    # pointee, so we need to get the smart pointer before dereferencing. This
+    # does not happen for smart pointers to fundamental types.
+    real_smartptr = ptr_proxy.__smartptr__() if ptr_proxy.__smartptr__() is not None else ptr_proxy
+    return real_smartptr.__deref__()
 
 
 def _REntry_setitem(self, key, value):
-    ptr = self.GetPtr(key)
-    ptr.get()[0] = value
-
+    ptr_proxy = self.GetPtr(key)
+    # Most types held by a shared_ptr are exposed by cppyy as the type of the
+    # pointee, so we need to get the smart pointer before dereferencing. This
+    # does not happen for smart pointers to fundamental types.
+    real_smartptr = ptr_proxy.__smartptr__() if ptr_proxy.__smartptr__() is not None else ptr_proxy
+    # ref-assignment is not supported by cppyy (and it also does not apply well
+    # to Python in general), so we use a helper.
+    import ROOT
+    ROOT.Experimental.Internal.AssignValueToPointee(value, real_smartptr)
 
 @pythonization("REntry", ns="ROOT::Experimental")
 def pythonize_REntry(klass):
--- a/tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
@@ -20,6 +20,7 @@
 
 #include <string>
 #include <string_view>
+#include <type_traits>
 #include <variant>
 
 #include <ROOT/RError.hxx>
@@ -288,6 +289,14 @@ static_assert(kTestLocatorType < RNTupleLocator::ELocatorType::kLastSerializable
 
 /// Check whether a given string is a valid name according to the RNTuple specification
 RResult<void> EnsureValidNameForRNTuple(std::string_view name, std::string_view where);
+
+/// Helper for the Python interface to perform the assignment to the pointee of a shared_ptr.
+template <typename T, typename U>
+void AssignValueToPointee(T &&value, std::shared_ptr<U> &ptr)
+{
+   static_assert(std::is_convertible_v<T, U>, "Cannot convert type of input argument to type of the pointee.");
+   *ptr = std::forward<T>(value);
+}
 
 } // namespace Internal
 
--- a/tree/ntuple/v7/test/ntuple_basics.py
+++ b/tree/ntuple/v7/test/ntuple_basics.py
@@ -6,6 +6,7 @@ RNTupleModel = ROOT.Experimental.RNTupleModel
 RNTupleReader = ROOT.Experimental.RNTupleReader
 RNTupleWriter = ROOT.Experimental.RNTupleWriter
 
+
 class RNTupleBasics(unittest.TestCase):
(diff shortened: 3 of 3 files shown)


jblomer (approved):
Thank you! LGTM.

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
```diff
@@ -24,14 +24,24 @@ def _REntry_GetPtr(self, key):
 
 
 def _REntry_getitem(self, key):
-    ptr = self.GetPtr(key)
-    return ptr.get()[0]
+    ptr_proxy = self.GetPtr(key)
+    # Most types held by a shared_ptr are exposed by cppyy as the type of the
+    # pointee, so we need to get the smart pointer before dereferencing. This
+    # does not happen for smart pointers to fundamental types.
+    real_smartptr = ptr_proxy.__smartptr__() if ptr_proxy.__smartptr__() is not None else ptr_proxy
```
hahnjo, line 31: I think this should move to `_REntry_GetPtr`, we want this also for that lower level function.

hahnjo (commented):
> For writing, there is no automatic support for assignment to the pointee.

This is only partly true: using `ptr.get()[0]` allows assignment (once you have the smart pointer), and better yet (IMHO) is symmetric between reading and writing.

> Thus, a utility helper is introduced in C++. Since this is a very specific case, the helper is only introduced in the RNTuple library.

In my opinion this is not nice nor optimal, also because it moves the conversion from the Python interface to the C++ side. So now we have to hope that the first template type that PyROOT tries to instantiate `AssignValueToPointee` with is convertible on the C++ side. That's probably true in most cases, but I could easily imagine types that have special conversion before leaving the Python side. Is there a way we can make array / pointer assignment work in PyROOT, without the need of RNTuple specific code?

vepadulano:
> This is only partly true: using ptr.get()[0] allows assignment (once you have the smart pointer), and better yet (IMHO) is symmetric between reading and writing.

Unfortunately this pattern, present before this PR, is only valid for those types that cppyy exposes as `LowLevelView` (in practice mostly fundamental types), as I explained in the PR description. A simple test with the type `std::string`, the original problematic type as reported by https://mattermost.web.cern.ch/root/pl/sma1p3jhd3bubpds69h5mpgk7o, would show the issue

```
>>> import cppyy
>>> ptr = cppyy.gbl.std.make_shared[cppyy.gbl.std.string]("hello world")
>>> ptr
b'hello world'
>>> ptr.__smartptr__
<built-in method __smartptr__ of string object at 0x7fc36f064640>
>>> ptr.__smartptr__()
<cppyy.gbl.std.shared_ptr<string> object at 0x67c3960>
>>> ptr.__smartptr__().get()
b'hello world'
>>> ptr.__smartptr__().get()[0]
'h'
>>> ptr.__smartptr__().get()[0] = "new string"
  File "<stdin>", line 1, in <module>
TypeError: char& std::string::operator[](string::size_type __pos) =>
    TypeError: 'str' object cannot be interpreted as an integer
```


> In my opinion this is not nice nor optimal, also because it moves the conversion from the Python interface to the C++ side. So now we have to hope that the first template type that PyROOT tries to instantiate AssignValueToPointee with is convertible on the C++ side. That's probably true in most cases, but I could easily imagine types that have special conversion before leaving the Python side. Is there a way we can make array / pointer assignment work in PyROOT, without the need of RNTuple specific code?

Sure, there are ways to improve the situation. A straightforward one would be moving the same logic to our own CPython extension or cppyy, but as described above this was not done because of lack of concrete motivation. Suggestions are always welcome.

hahnjo:
Ok, the situation is not quite satisfactory from the point of view of writing pythonizations: We cannot use `smartptr` in `_REntry_GetPtr` directly because it seems returning it does not keep a reference count of the pointee, which is then garbage collected (at least that's how I interpret the crashes). The best I can offer is https://github.com/root-project/root/pull/17376, which does not need the C++ helper but is still not 100% symmetrical between reading and writing. Let's discuss on that PR which of the changes we want, if any...

vepadulano:
> it seems returning it does not keep a reference count of the pointee, which is then garbage collected (at least that's how I interpret the crashes)

I think your proposed alternative is valid, I went ahead already and approved the PR. But I wanted to ask you regarding the crashes you mention, is it the `TypeError` from the example I pasted above? From the surface I can't see how that is related to the smart pointer not keeping the correct reference count. Or maybe is it about something else?

hahnjo:
> But I wanted to ask you regarding the crashes you mention, is it the `TypeError` from the example I pasted above? From the surface I can't see how that is related to the smart pointer not keeping the correct reference count. Or maybe is it about something else?

Conceptually I tried moving the `__smartptr__()` code into `_REntry_GetPtr`. It would still not map in an ideal way to Python, but at least `GetPtr` would always return a `shared_ptr`, which might be interesting to a user as a low-level interface (in case they want to do something similar as our pythonization, who knows). However, that resulted in crashes and in trying to understand them I noticed that even the following diff (on current `master`) already crashes:
```diff
diff --git a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
index af165c7a7d..86ffe1fc68 100644
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
@@ -29,6 +29,7 @@ def _REntry_getitem(self, key):
     # pointee, so we need to get the smart pointer before dereferencing. This
     # does not happen for smart pointers to fundamental types.
     real_smartptr = ptr_proxy.__smartptr__() if ptr_proxy.__smartptr__() is not None else ptr_proxy
+    ptr_proxy = None
     return real_smartptr.__deref__()


```

This conceptually mimics the result of `self._GetPtr[fieldType](key)` going out-of-scope and therefore being garbage collected. Looking with the debugger, this fails when accessing `std::string` and given the one-line change it can only be during dereferencing. The only explanation that makes sense to me is that `real_smartptr` does not hold a reference count to the original proxy object...

vepadulano:
> The only explanation that makes sense to me is that real_smartptr does not hold a reference count to the original proxy object...

Ouch, might be yes. Thanks for the explanation!

Issue #17225, opened 2024-12-09
TFormula: Possibility of failure during dynamic compilation of predefined functions "gausn" and "landau"


AlexeyVKrasnoperov:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

It may happen that the program has a locale with a comma as a decimal separator 
("fr_FR.utf8", "de_DE.utf8", "ru_RU.utf8", .. ).
In this case, initialization of the TFormula objects with functions "gausn" and "landau"
ends with an error.  The following messages are displayed on the screen:
...
input_line_11:2:158: error: too many arguments to function call, expected single argument 'x', have 2 arguments
`Double_t TFormula____id3053186198674332598(Double_t *x,Double_t *p){ return p[0]*TMath::Exp(-0.5*(((x[0])-p[1])/p[2])*(((x[0])-p[1])/p[2]))/(TMath::Sqrt(2*3,141593)*p[2]) ; }
`
...

input_line_13:2:115: error: too many arguments to function call, expected at most 4, have 5
`Double_t TFormula____id3986067622224738083(Double_t *x,Double_t *p){ return p[0]*TMath::Landau((x[0]),p[1],p[2],0,000000) ; }`
...

### Reproducer

It is enough to execute this macro in ROOT with different locale names as a parameter.

```
#include <TFormula.h>
#include <locale.h>

void CheckTFormula(const char *locale = nullptr)
{  
  if( locale != nullptr )
    {
      const char * ptr = setlocale(LC_NUMERIC,locale);
      if( ptr == nullptr )
        {
          fprintf(stderr,"Cannot set locale: \"%s\"\n",locale);
          return;
        }
      printf("Numeric locale: \"%s\"\n",ptr);
    }
  lconv * lc = localeconv();
  if( lc != nullptr )
    printf("Decimal point: \"%s\"\n",lc->decimal_point);
  //
  double num = 12345.67890;
  printf("printf: \"%lf\"\n",num);  
  printf("TString::Format: \"%s\"\n\n",TString::Format("%lf",num).Data());
  //
  TFormula f0("f0","gausn(x)");
  if( f0.IsValid() )
    {
      f0.Print();
      printf("\n");
    }
  //
  TFormula f1("f1","landau(x)");
  if( f1.IsValid() )
    f1.Print();
}
```

### ROOT version

ROOT 6.32.08. Perhaps everywhere, starting from new TFromula ?

### Installation method

build from source

### Operating system

Linux, Debian 12.8

[...]

PR #17327, opened 2024-12-20
TFormula: support locale settings for which the decimal separator is ","

dpiparo:
Fixes https://github.com/root-project/root/issues/17225


lmoneta (approved):
LGTM, if it passes the tests should be fine. 
Thank you Danilo for the fix

hist/hist/src/TFormula.cxx
```diff
@@ -2263,7 +2263,9 @@ void TFormula::ProcessFormula(TString &formula)
          map<TString, Double_t>::iterator constIt = fConsts.find(fun.GetName());
          if (constIt != fConsts.end()) {
             TString pattern = TString::Format("{%s}", fun.GetName());
-            TString value = TString::Format("%lf", (*constIt).second);
+            // #17225: we take into account LC_LOCALE settings for which the decimal separator
+            // is , instead of ., e.g. de_AT.UTF-8
+            TString value = TString::Format("%lf", (*constIt).second).ReplaceAll(",", ".");
```
pcanal, line 2268: Shouldn't this be dependent on the value of LC_LOCALE?  (in some local we might have `1,200.00` for one thousand two hundreds).

dpiparo: actually it seems that there are only 3 decimal separators: comma, period and arabic decimal separator (https://randombits.dev/articles/number-localization/locale-list)

dpiparo: Thinking more about it... If we have an arabic decimal separator, it means that we also have arabic digits. At that point, I am unsure whether the decimal separator would be the first problem we hit...

pcanal: Actually, I think we need a larger conversation about local and how to handle them.  It seems odds/confusion to all support local for this single functions.  Until we have this conversation, I recommend we issue an explicit error when encountering a comma here.

dpiparo:
Superseded by https://github.com/root-project/root/pull/19734.

PR #17330, opened 2024-12-20
Text precision must be 2 in PaintText

couet:
In THistPainter::PaintText the text precision must be 2 .
This problem was mention here https://root-forum.cern.ch/t/th2-option-text-not-working-with-font-43/62513


bellenot (approved):
once all builds are green

PR #17332, opened 2024-12-21
Use legend entry's option instead of object's DrawOption when deciding endcaps

will-cern:
I have encountered the case where the legend is being drawn to a separate pad from the object, which means the `GetDrawOption()` returns a blank string no matter how it was drawn in the other pad.

It would surely make most sense to have the decision about endcaps be determined from the TLegendEntry option, like all the other aspects of the drawing of the legend entry. 

This PR makes that change.


couet (approved):
Thx. LGTM

PR #16285, opened 2024-08-22
[CMake] Only replace `find_package` if required

guitargeek:
**This is a replay of #16274 with the "clear build" label for the CI, to see if we can reproduce the build error that were observed after merging said PR**

Our CMake code modifies `find_package` to ignore any ROOT builtins, such that the builtin LLVM doesn't find and use system versions of the builtin dependencies.

This is a bit hacky, but fortunately this needs to be done only when builtins are used.

Therefore, this commit suggests to only do this redefinition of `find_package` if any builtins are used.

This closes #8633, where it was requested to avoid this hack for the sake of being compatible with the `vcpkg` package manager.

Although the hack is not completely removed, it is removed for the case where not builtins are used, which is probably what is done when using a C++ package manager to manage dependencies.

Issue #8633, opened 2021-07-08
Don't replace CMake's find_package


bernhardmgruber:
- [x] Checked for duplicates

### Describe the bug

ROOT overrides CMake's `find_package` in `cmake/modules/SearchInstalledSoftware.cmake`, which is not uncommon, but relies on undocumented behavior of CMake. See e.g.: https://crascit.com/2018/09/14/do-not-redefine-cmake-commands/

This however breaks other tools, that also rely on this behavior, resulting in an infinite recursion. My concrete example is vcpkg, a popular package manager (not only for Windows), that pulls itself into a CMake project via a toolchain file to inject dependencies. Unfortunately, vcpkg needs to hook `find_package` itself to point it to the installed dependencies.

### Expected behavior

ROOT can be configured with vcpkg and other tools sensitive to overriden `find_package`. This requires ROOT to not override `find_package`.

### To Reproduce

Clone ROOT, create `build_win` folder, run:
```
PS C:\dev\root\build_win> cmake -DCMAKE_TOOLCHAIN_FILE=D:\dev\vcpkg\scripts\buildsystems\vcpkg.cmake ..
-- Building for: Visual Studio 16 2019
-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19042.
-- The C compiler identification is MSVC 19.29.30037.0
-- The CXX compiler identification is MSVC 19.29.30037.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30037/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30037/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Git: C:/Program Files/Git/cmd/git.exe (found version "2.30.0.windows.2")
-- Detected ROOT_VERSION 6.25.01
-- Looking for Python
-- Could NOT find Python3 (missing: Python3_NumPy_INCLUDE_DIRS NumPy) (found version "3.9.2")
[...]

PR #17335, opened 2024-12-27
Revert "[CMake] Only replace `find_package` if required"

dpiparo:
Temporarily undo this PR https://github.com/root-project/root/pull/16285
since it broke the fedoras


guitargeek (approved):
Thanks! That PR of mine seems to have been merged by mistake

guitargeek:
Also, https://github.com/root-project/root/issues/8633 needs to be re-opened.

pcanal:
This reversion is necessary as the original PR was indeed not properly functioning.  See https://github.com/root-project/root/issues/8633#issuecomment-2563957177.

PR #17339, opened 2024-12-27
CI: proper escaping of parentheses

pcanal:
Fixes:
```
/__w/root/root/.github/workflows/root-ci-config/build_root.py:539: SyntaxWarning: invalid escape sequence '\('
  """, "Failed capture of lastest-stable underlying branch name")
```


pcanal:
The only failures is `tutorial-roostats-StandardHistFactoryPlotsWithCategories-py` on Fedora 41 and is unrelated.
Merging to see the actual effect of this PR which is seen in the execution of the push actions.

PR #17340, opened 2024-12-27
"io_ntuple" tutorials translated into Python.

PPaye:
# This Pull request:
Translates the the .C tutorials of "root/tutorials/v7/ntuple" folder into Python3. 
Each tutorials comes from its respective cppyy wrapper for root-classes or -functions when needed.

Remark : This is part of the 6.30 branch. Now, that folder doesn't exist anymore. 
So this should be on "root/tutorials/io/ntuple" not on "root/tutorials/experimental".

## Checklist:
- [ yes ] tested changes locally
- [ yes ] updated the docs (if necessary)


hahnjo (changes requested):
ROOT `master` now has properly working Python bindings for RNTuple. The wrappers proposed here are in my opinion not what we want to advertise to users. In general, I'm not sure we want to translate all RNTuple tutorials to Python...

jblomer:
Many thanks for your effort!

While this is an interesting exercise, the wrapper approach is not what we encourage users to follow. As @hahnjo  pointed out, RNTuple has now native Python support with a conscious selection of features and workflows to support in Python. We will follow this path and create Python tutorials on a case-by-base basis as we learn more about the intended Python usage of RNTuple.

PR #17345, opened 2025-01-02
[RF] Support also `createNLL` in case of partially-extended simul. fits 

guitargeek:
Support also `createNLL` in case of partially-extended simultaneous fits
with the new default CPU backend (and hence also "codegen").

This problem was reported by @gartrog in private communication.

The fix needs to be backported to the next 6.34 patch release.

Add also a corresponding test.

Some more commits in this PR improve the `RooArgSet` interface, fix a memory leak, and avoid manual memory management.


lmoneta (approved):
LGTM!

PR #17352, opened 2025-01-06
[ntuple] Add separate option for same page merging

hahnjo:


tree/ntuple/v7/src/RPageStorage.cxx
```diff
@@ -1014,7 +1014,7 @@ void ROOT::Experimental::Internal::RPagePersistentSink::CommitSealedPageV(
       locatorIndexes.reserve(locatorIndexes.size() + rangeSize);
 
       for (auto sealedPageIt = range.fFirst; sealedPageIt != range.fLast; ++sealedPageIt) {
-         if (!fFeatures.fCanMergePages || !sealedPageIt->GetHasChecksum()) {
+         if (!fFeatures.fCanMergePages || !fOptions->GetEnableSamePageMerging() || !sealedPageIt->GetHasChecksum()) {
```
jblomer, line 1017: Should we still check for the checksum here and not only for same page merging?

hahnjo: Yes right; I turned it into an explicit assertion that is checked before unwrapping the checksum.

jblomer (approved):
LGTM

PR #17354, opened 2025-01-06
[interpreter] Revert patch to move `ClangConfig.cmake`

hahnjo:
Commit https://github.com/root-project/root/commit/8a45c096f16a38d3990a7b2efc894145c178b992 already modified the setup of `ClingConfig.cmake`, this reverts the second part of commit https://github.com/root-project/root/commit/2b283ccf3a624f70dab3e8783d361d25c13e2c65 and removes one downstream patch from our Clang sources.

TODO:
- [x] More testing with `builtin_llvm=OFF` / `builtin_clang=OFF`, also for the individual commits.
- [x] Create new tag and update `interpreter/llvm-project.tag` (to be sequenced with other changes)


hahnjo:
I tested with various configurations, and it appears `builtin_clang=OFF` is broken since the upgrade to LLVM 16 (https://github.com/root-project/root/issues/17461). I checked with the diff pasted in the issue that this PR doesn't make things worse, and I will fix `builtin_clang=OFF` in a subsequent PR. @vgvassilev if the changes look ok to you, I will prepare a tag and update `interpreter/llvm-project.tag` so we can get rid of another of our downstream patches (and clean up our CMake code in the process)

vgvassilev (approved):
LGTM!

devajithvs (approved):

PR #17356, opened 2025-01-06
[RF] Fix inverted `gSystem.AccessPathName` in tutorial

hahnjo:
This caused the file `example_combined_GaussExample_model.root` to be written again while other tutorials were already running, potentially resulting in short reads and test failures.


hahnjo:
I marked this as "emergency" even though it's already in `master` since a while. Example failures, only in "my" PRs today:
 * https://github.com/root-project/root/runs/35196129425
 * https://github.com/root-project/root/runs/35193700405
 * https://github.com/root-project/root/runs/35191742902

guitargeek (approved):
Nice catch, thank you very much!

dpiparo:
Nice, I was looking for that since days!

PR #17358, opened 2025-01-07
[RF] Add parameter getters to RooCrystalBall

guitargeek:
The RooCrystalBall class didn't have getters to its parameters yet, which we generally need for all commonly-used PDFs for model inspection and IO (other than IO with ROOT files).

Requested by @VanyaBelyaev.


lmoneta (approved):
LGTM!
Thank you Jonas for the improvements

PR #17359, opened 2025-01-07
[ntuple] Clean up utility definitions

jblomer:
Preparation to move the common definitions from RNTupleUtil.h out of the experimental namespace. Some definitions are removed or moved to `ROOT::Internal`. Some API improvements for the remaining public types.

Part of the PR series to move an initial set of RNTuple APIs out of the experimental namespace.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
@@ -373,7 +373,7 @@ public:
    {
       return fColumnRanges.find(physicalId) != fColumnRanges.end();
    }
-   std::uint64_t GetBytesOnStorage() const;
+   std::uint64_t GetNBytesOnStorage() const;
 };
 
 class RClusterDescriptor::RColumnRangeIterable {
--- a/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
@@ -240,7 +240,7 @@ public:
       NTupleSize_t fNElements = kInvalidNTupleIndex;
       /// The usual format for ROOT compression settings (see Compression.h).
       /// The pages of a particular column in a particular cluster are all compressed with the same settings.
-      int fCompressionSettings = kUnknownCompressionSettings;
+      int fCompressionSettings = kNTupleUnknownCompression;
       /// Suppressed columns have an empty page range and unknown compression settings.
       /// Their element index range, however, is aligned with the corresponding column of the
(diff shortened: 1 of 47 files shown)


tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
```diff
@@ -173,8 +173,8 @@ public:
 /// Generic information about the physical location of data. Values depend on the concrete storage type.  E.g.,
 /// for a local file `fPosition` might be a 64bit file offset. Referenced objects on storage can be compressed
 /// and therefore we need to store their actual size.
-/// TODO(jblomer): consider moving this to `RNTupleDescriptor`
-struct RNTupleLocator {
+class RNTupleLocator {
+public:
```
hahnjo, line 176: Hm, what's the advantage of a `class` if we have getters and setters without further checks? Looking at the changes in this commit, a `struct` feels more ergonomic...

tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
```diff
@@ -189,7 +189,7 @@ public:
    };
 
 private:
-   std::uint64_t fBytesOnStorage = 0;
+   std::uint64_t fNBytesOnDisk = 0;
```
hahnjo: This gives up a bit of abstraction, for example when storing on object stores...

tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -373,7 +373,7 @@ public:
    {
       return fColumnRanges.find(physicalId) != fColumnRanges.end();
    }
-   std::uint64_t GetBytesOnStorage() const;
+   std::uint64_t GetNBytesOnDisk() const;
```
hahnjo, line 376: see also comment on previous commit; the binary format specification calls it "storage"...

hahnjo (commented):
Generally LGTM, but some comments especially regarding `NBytesOnDisk`. Not sure if giving up abstraction is an improvement...

tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
jblomer: It's to be more resilient to future changes. Maybe we want to rename something, or add a check (e.g., if the type used in `fPosition` fits the locator type).

hahnjo: Ok. There is still a forward declaration of `struct RNTupleLocator;` in `RPageStorageFile.hxx` that is now causing warnings...

hahnjo (approved):

PR #17364, opened 2025-01-07
[Python] Avoid using `std::regex` in CPyCppyy

guitargeek:
Using `std::regex` in CPyCppyy leads to problems in conjunction with XGBoost and PyTorch. We could previously work around these issues by importing these libraries before ROOT, but there still seem to be problems remaining as we see in the [CppInterop PR](https://github.com/root-project/root/pull/16814).

This commit tries to revert the introduction of `std::regex` in CPyCppyy and the import order hack, to see if this would also work.

https://github.com/wlav/CPyCppyy/commit/c437364950cafe06

For ROOTs CPyCppyy, `std::regex` was only introduced with the syncup in ROOT 6.32. Therefore, reverting back to the previous algorithm without `std::regex` should not affect our uses.

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
@@ -23,7 +23,6 @@
 #include <algorithm>
 #include <array>
 #include <locale>        // for wstring_convert
-#include <regex>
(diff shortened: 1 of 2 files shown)


guitargeek:
Didn't help in the CppInterOp PR, so not worth doing.

PR #17369, opened 2025-01-07
[ntuple] Improvements of Python interface and tests

hahnjo:
As discussed around the RNTuple workshop, forbid calling `RNTupleModel::CreateEntry()` directly from Python. This allows to clone the passed `RNTupleModel` when creating a reader or writer, and avoid destructively passing the user argument.

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
@@ -93,6 +93,18 @@ def pythonize_RNTupleModel(klass):
     klass.MakeField = MethodTemplateGetter(klass.MakeField, _RNTupleModel_MakeField)
 
 
+def _RNTupleReader_Open(maybe_model, *args):
+    if (
+        hasattr(type(maybe_model), "__cpp_name__")
+        and type(maybe_model).__cpp_name__ == "ROOT::Experimental::RNTupleModel"
+    ):
+        # In Python, the user cannot create REntries directly from a model, so we can safely clone it and avoid destructively passing the user argument.
+        maybe_model = maybe_model.Clone()
+    import ROOT
+
+    return ROOT.Experimental.RNTupleReader._Open(maybe_model, *args)
+
+
 def _RNTupleReader_LoadEntry(self, *args):
     if len(args) < 2:
         raise ValueError(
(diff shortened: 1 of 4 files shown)


bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
```diff
@@ -60,6 +60,12 @@ def _RNTupleModel_CreateBare(*args):
     return ROOT.Experimental.RNTupleModel._CreateBare()
 
 
+def _RNTupleModel_CreateEntry(self):
+    raise RuntimeError(
+        "creating entries from model is not supported in Python, create CreateEntry on the reader or writer"
```
vepadulano, line 65:
```suggestion
        "creating entries from model is not supported in Python, call CreateEntry on the reader or writer"
```

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
```diff
@@ -78,11 +84,25 @@ def pythonize_RNTupleModel(klass):
...
     klass.MakeField = MethodTemplateGetter(klass.MakeField, _RNTupleModel_MakeField)
 
 
+def _RNTupleReader_Open(maybe_model, *args):
+    if (
+        hasattr(type(maybe_model), "__cpp_name__")
+        and type(maybe_model).__cpp_name__ == "ROOT::Experimental::RNTupleModel"
+    ):
+        # In Python, the user cannot create REntries directly from a model, so we can safely clone it and avoid destructively passing the user argument.
+        maybe_model = maybe_model.Clone()
```
vepadulano, line 100: Here we might have a free opportunity for some extra error handling in an `else` branch letting the user know that only an `RNTupleModel` is accepted as input.

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
```diff
@@ -93,10 +113,33 @@ def _RNTupleReader_LoadEntry(self, *args):
...
+    klass._Open = klass.Open
+    klass.Open = _RNTupleReader_Open
+
     klass._LoadEntry = klass.LoadEntry
     klass.LoadEntry = _RNTupleReader_LoadEntry
 
 
+def _RNTupleWriter_Append(model, *args):
+    # In Python, the user cannot create REntries directly from a model, so we can safely clone it and avoid destructively passing the user argument.
+    model = model.Clone()
```
vepadulano, line 125: Maybe we want to have the same check as in `_RNTupleWriter_Recreate` here?

vepadulano (approved):
Very nice thanks! Some minor comments from my side.

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
hahnjo: Ew, that happens if I rewrite the error message too many times...

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
hahnjo: `RNTupleReader::Open` is overloaded, the first argument can also just be a string. In that case, the model is reconstructed from the on-disk header. I don't know if there is a better way to implement overloaded methods in Python...

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
hahnjo: We could, but for `RNTupleWriter::Append` the first `model` parameter is mandatory. For `Recreate`, the user can also pass a list of tuples (type, fieldname)`.

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
vepadulano: > I don't know if there is a better way to implement overloaded methods in Python...

The best approximation of it is https://docs.python.org/3/library/functools.html#functools.singledispatch , I don't know if it's worth using it for this particular PR.

jblomer (approved):

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
hahnjo: I also don't know how nicely this plays with cppyy. I'll leave the few extra cases as they are for the moment.

PR #17373, opened 2025-01-08
[ntuple] Remove redundant `RNTupleProcessor::LoadEntry` call

enirolf:
When iterating over entries in the `RNTupleProcessor`, the entry is currently loaded twice: once on `RNTupleProcessor::Advance`, which is called when the iterator is incremented, and once when the iterator is dereferenced. This second call is unnecessary and only slows down the processor, so we should remove it.


vepadulano (approved):
Thanks!

jblomer (approved):

PR #17376, opened 2025-01-09
[ntuple] Simplify `REntry` pythonizations

hahnjo:
For non-fundamental types, we can avoid getting the smartptr just to deference or assign to it. Also forbid calling `REntry::GetPtr` from Python, it does not map well.


vepadulano (approved):
Nice trick! I think we can merge if the CI agrees.

jblomer (approved):

PR #17379, opened 2025-01-09
[geom] Emulate Inside functionality to fix false positives in overlap checking

agheata:
# This Pull request:
This excludes solid mesh points used for overlap checking if these points are not on the surface, which can generate false positives in some Boolean cases.

## Changes or fixes:
Implemented an emulated `Inside` function in TGeoShape. This functionality exists in Geant4 and VecGeom, implemented natively by all solids, and detects points numerically on the solid surface within some tolerance. The emulation is implemented in `TGeoShape` by XOR-ing the return values given by the `Contains` function in a pushed and a pulled point along the surface normal vector. The push/pull is made using `TGeoShape::Tolerance`. 

The new `Inside` method discards mesh points on virtual surfaces inside joining Boolean solids. Previously, these points were considered on the solid surface and generated false-positive extrusions during overlap checking. I've made the `TGeoShape::ComputeNormal` interface const in the implementation process.

## Checklist:

- [x] tested changes locally
- [x] Added new google test inside `gtest-geom-geomTests`

This PR fixes #

### DIFF
--- a/geom/geom/inc/TGeoArb8.h
+++ b/geom/geom/inc/TGeoArb8.h
@@ -42,7 +42,7 @@ public:
    // methods
    Double_t Capacity() const override;
    void ComputeBBox() override;
-   void ComputeNormal(const Double_t *point, const Double_t *dir, Double_t *norm) override;
+   void ComputeNormal(const Double_t *point, const Double_t *dir, Double_t *norm) const override;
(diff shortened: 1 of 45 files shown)


bellenot (approved):
LGTM, thanks Andrei!

PR #17382, opened 2025-01-09
[df] Implement rule of five for RNodeBase and derived classes

vepadulano:


tree/dataframe/inc/ROOT/RDF/RFilterBase.hxx
```diff
@@ -53,9 +53,14 @@ public:
...
-   RFilterBase &operator=(const RFilterBase &) = delete;
 
-   ~RFilterBase() override;
+   // Rule of five
+
+   RFilterBase(const RFilterBase &) = delete;
+   RFilterBase &operator=(const RFilterBase &) = delete;
+   RFilterBase(RFilterBase &&) = delete;
+   RFilterBase &operator=(RFilterBase &&) = delete;
+   ~RFilterBase() override = default;
```
pcanal, line 63: Often the out-line destructor is used to 'anchor' the virtual table generation into a specific object file.  Is that not in play here?

tree/dataframe/src/RRangeBase.cxx
```diff
@@ -24,6 +24,3 @@ void RRangeBase::InitNode()
    fNProcessedEntries = 0;
    fHasStopped = false;
 }
-
-// outlined to pin virtual table
-RRangeBase::~RRangeBase() { }
```
pcanal, line 29: Is the pinning no longer needed?

tree/dataframe/inc/ROOT/RDF/RFilterBase.hxx
vepadulano: It was not done for every class, I understand it might have had its use, so I can write the destructor out of line for all the classes touched by this PR

pcanal (approved):
LGTM

PR #17389, opened 2025-01-09
[ntuple] remove kNTupleUnknownCompression

jblomer:
Instead represent undefined/unknown compression settings by std::optional<>. Also allows for representing compression settings as _unsigned_ integer.

Companion PR: https://github.com/root-project/roottest/pull/1249

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
@@ -240,7 +240,8 @@ public:
       NTupleSize_t fNElements = kInvalidNTupleIndex;
       /// The usual format for ROOT compression settings (see Compression.h).
       /// The pages of a particular column in a particular cluster are all compressed with the same settings.
-      int fCompressionSettings = kNTupleUnknownCompression;
+      /// If unset, the compression settings are undefined (for suppressed columns).
+      std::optional<std::uint32_t> fCompressionSettings;
       /// Suppressed columns have an empty page range and unknown compression settings.
       /// Their element index range, however, is aligned with the corresponding column of the
       /// primary column representation (see Section "Suppressed Columns" in the specification)
--- a/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
@@ -240,7 +240,7 @@ public:
       NTupleSize_t fNElements = kInvalidNTupleIndex;
       /// The usual format for ROOT compression settings (see Compression.h).
       /// The pages of a particular column in a particular cluster are all compressed with the same settings.
-      /// If unset, the compression settings are undefined (for suppressed columns).
+      /// If unset, the compression settings are undefined (deferred columns, suppressed columns).
       std::optional<std::uint32_t> fCompressionSettings;
       /// Suppressed columns have an empty page range and unknown compression settings.
       /// Their element index range, however, is aligned with the corresponding column of the
--- a/tree/ntuple/v7/src/RNTupleDescriptorFmt.cxx
+++ b/tree/ntuple/v7/src/RNTupleDescriptorFmt.cxx
@@ -124,7 +124,7 @@ void ROOT::Experimental::RNTupleDescriptor::PrintInfo(std::ostream &output) cons
 
          info.fNElements += columnRange.fNElements;
          if (compression == -1) {
-            compression = columnRange.fCompressionSettings;
+            compression = columnRange.fCompressionSettings.value();
          }
          const auto &pageRange = cluster.second.GetPageRange(column.second.GetPhysicalId());
          auto idx = cluster2Idx[cluster.first];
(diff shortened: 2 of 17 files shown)


tree/ntupleutil/v7/src/RNTupleInspector.cxx
```diff
@@ -166,8 +166,8 @@ ROOT::Experimental::RNTupleInspector::Create(std::string_view ntupleName, std::s
 
 std::string ROOT::Experimental::RNTupleInspector::GetCompressionSettingsAsString() const
 {
-   int algorithm = fCompressionSettings / 100;
-   int level = fCompressionSettings - (algorithm * 100);
+   int algorithm = fCompressionSettings.value() / 100;
```
silverweed, line 169: Shouldn't this properly handle the case where compression is unknown, rather than throwing?

tree/ntupleutil/v7/src/RNTupleExporter.cxx
```diff
@@ -162,7 +162,7 @@ RNTupleExporter::RPagesResult RNTupleExporter::ExportPages(RPageSource &source,
             const std::uint64_t pageBufSize = pageInfo.fLocator.GetNBytesOnStorage() + maybeChecksumSize;
             std::ostringstream ss{options.fOutputPath, std::ios_base::ate};
             ss << "/cluster_" << clusterDesc.GetId() << "_" << colInfo.fQualifiedName << "_page_" << pageIdx
-               << "_elems_" << pageInfo.fNElements << "_comp_" << colRange.fCompressionSettings << ".page";
+               << "_elems_" << pageInfo.fNElements << "_comp_" << colRange.fCompressionSettings.value() << ".page";
```
silverweed, line 165: Shouldn't this properly handle the case where compression is unknown, rather than throwing?

tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
```diff
@@ -212,7 +213,7 @@ public:
    ///
    /// \note Here, we assume that the compression settings are consistent across all clusters and columns. If this is
    /// not the case, an exception will be thrown when RNTupleInspector::Create is called.
-   int GetCompressionSettings() const { return fCompressionSettings; }
+   int GetCompressionSettings() const { return fCompressionSettings.value(); }
```
silverweed, line 216: Is there a reason why this shouldn't return an optional? 
The comment states that "we assume that the compression settings are consistent across all clusters and columns" - do we also assume the compression settings are not unknown? Perhaps we should add a brief explanation on why `fCompressionSettings` is an optional, it might not be obvious to a future reader.

tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -240,7 +240,8 @@ public:
       NTupleSize_t fNElements = kInvalidNTupleIndex;
       /// The usual format for ROOT compression settings (see Compression.h).
       /// The pages of a particular column in a particular cluster are all compressed with the same settings.
-      int fCompressionSettings = kNTupleUnknownCompression;
+      /// If unset, the compression settings are undefined (deferred columns, suppressed columns).
+      std::optional<std::uint32_t> fCompressionSettings;
```
pcanal, line 244: Does it change anything to the onfiles files in the columns meta data?

jblomer: No, there is no impact on the on disk data.

tree/ntupleutil/v7/src/RNTupleExporter.cxx
jblomer: At this point, we know that the column is neither suppressed nor in the deferred range. I added an assert and changed the access to the unchecked one.

tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
jblomer: Returns now an `std::optional`

tree/ntupleutil/v7/src/RNTupleInspector.cxx
jblomer: Fixed.

tree/ntuple/v7/src/RNTupleDescriptorFmt.cxx
```diff
@@ -124,7 +124,7 @@ void ROOT::Experimental::RNTupleDescriptor::PrintInfo(std::ostream &output) cons
 
          info.fNElements += columnRange.fNElements;
          if (compression == -1) {
-            compression = columnRange.fCompressionSettings;
+            compression = columnRange.fCompressionSettings.value();
```
silverweed, line 127: I think we should not throw here if the compression settings is unknown: even if we don't expect it to happen, this is a method used for informative reasons and it can handle the unexpected case in a clear manner (print "unknown"), so I think it should.

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -197,7 +197,7 @@ struct RChangeCompressionFunc {
       sealConf.fElement = &fDstColElement;
       sealConf.fPage = &page;
       sealConf.fBuffer = fBuffer;
-      sealConf.fCompressionSetting = fMergeOptions.fCompressionSettings;
+      sealConf.fCompressionSetting = fMergeOptions.fCompressionSettings.value();
```
silverweed, line 200: After the initial check at lines 889-896 the compression settings must be set to a valid value, so we can access it with a simple `*` during the whole `Merge` function

silverweed (approved):

PR #17393, opened 2025-01-09
[ntuple] Make `RNTupleChainProcessor` composable

enirolf:
This PR introduces the possibility to create `RNTupleChainProcessor`s from other processor objects. In turn, this makes it possible to, for example, create a chain of joined ntuples.

This PR is part of a bigger set of (foreseen) changes, collected and tracked in https://github.com/root-project/root/issues/17132.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
@@ -79,6 +79,9 @@ entry. Additional bookkeeping information can be obtained through the RNTuplePro
 // clang-format on
 class RNTupleProcessor {
    friend struct ROOT::Experimental::Internal::RNTupleProcessorEntryLoader; // for unit tests
+   friend class RNTupleSingleProcessor;
+   friend class RNTupleChainProcessor;
+   friend class RNTupleJoinProcessor;
 
 protected:
    // clang-format off
@@ -128,9 +131,9 @@ protected:
 
    std::unique_ptr<RNTupleModel> fModel;
 
-   NTupleSize_t fNEntriesProcessed = 0;  //< Total number of entries processed so far
-   std::size_t fCurrentNTupleNumber = 0; //< Index of the currently open RNTuple
-   NTupleSize_t fCurrentEntryNumber = 0; //< Current processor entry number
+   NTupleSize_t fNEntriesProcessed = 0;     //< Total number of entries processed so far
+   NTupleSize_t fCurrentEntryNumber = 0;    //< Current processor entry number
+   std::size_t fCurrentProcessorNumber = 0; //< Number of the currently open inner processor
 
    /////////////////////////////////////////////////////////////////////////////
    /// \brief Creates and connects a concrete field to the current page source, based on its proto field.
@@ -143,6 +146,16 @@ protected:
    ///
    /// \return `entryNumber` if the entry was successfully loaded, `kInvalidNTupleIndex` otherwise.
    virtual NTupleSize_t LoadEntry(NTupleSize_t entryNumber) = 0;
+
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Points processor entry's field values to those in the provided entry.
+   ///
+   /// \param[in] entry The entry whose field values to use.
+   virtual void SetEntryPointers(const REntry &entry) = 0;
+
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Get the total number of entries in this processor
+   virtual NTupleSize_t GetNEntries() = 0;
 
    /////////////////////////////////////////////////////////////////////////////
    /// \brief Create a new base RNTupleProcessor.
@@ -166,19 +179,17 @@ public:
 
    /////////////////////////////////////////////////////////////////////////////
    /// \brief Get the total number of entries processed so far.
-   ///
-   /// When only one RNTuple is present in the processor chain, the return value is equal to GetLocalEntryNumber.
    NTupleSize_t GetNEntriesProcessed() const { return fNEntriesProcessed; }
 
    /////////////////////////////////////////////////////////////////////////////
-   /// \brief Get the index to the RNTuple currently being processed, according to the sources specified upon creation.
-   std::size_t GetCurrentNTupleNumber() const { return fCurrentNTupleNumber; }
+   /// \brief Get the entry number that is currently being processed.
+   NTupleSize_t GetCurrentEntryNumber() const { return fCurrentEntryNumber; }
 
    /////////////////////////////////////////////////////////////////////////////
-   /// \brief Get the entry number local to the RNTuple that is currently being processed.
+   /// \brief Get the number of the inner processor currently being read.
    ///
-   /// When only one RNTuple is present in the processor chain, the return value is equal to GetGlobalEntryNumber.
-   NTupleSize_t GetCurrentEntryNumber() const { return fCurrentEntryNumber; }
+   /// This method is only relevant for the RNTupleChainProcessor. For the other processors, 0 is always returned.
+   std::size_t GetCurrentProcessorNumber() const { return fCurrentProcessorNumber; }
 
    const RNTupleModel &GetModel() const { return *fModel; }
 
@@ -186,7 +197,6 @@ public:
    /// \brief Returns a reference to the entry used by the processor.
    ///
    /// \return A reference to the entry used by the processor.
-   ///
    const REntry &GetEntry() const { return *fEntry; }
 
    // clang-format off
@@ -213,7 +223,7 @@ public:
          : fProcessor(processor), fCurrentEntryNumber(entryNumber)
       {
          // This constructor is called with kInvalidNTupleIndex for RNTupleProcessor::end(). In that case, we already
-         // know there is nothing to advance to.
+         // know there is nothing to load.
          if (fCurrentEntryNumber != kInvalidNTupleIndex) {
             fCurrentEntryNumber = fProcessor.LoadEntry(fCurrentEntryNumber);
          }
@@ -232,7 +242,7 @@ public:
          return obj;
       }
 
-      reference operator*() { return *fProcessor.fEntry; }
+      reference operator*() { return fProcessor.GetEntry(); }
 
       friend bool operator!=(const iterator &lh, const iterator &rh)
       {
@@ -260,6 +270,17 @@ public:
    /// \return A pointer to the newly created RNTupleProcessor.
    static std::unique_ptr<RNTupleProcessor>
    CreateChain(const std::vector<RNTupleOpenSpec> &ntuples, std::unique_ptr<RNTupleModel> model = nullptr);
+
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Create a new RNTuple processor chain for vertical concatenation of previously created processors.
+   ///
+   /// \param[in] innerProcessors A list with the processors to chain.
+   /// \param[in] model An RNTupleModel specifying which fields can be read by the processor. If no model is provided,
+   /// one will be created based on the descriptor of the first ntuple specified.
+   ///
+   /// \return A pointer to the newly created RNTupleProcessor.
+   static std::unique_ptr<RNTupleProcessor> CreateChain(std::vector<std::unique_ptr<RNTupleProcessor>> innerProcessors,
+                                                        std::unique_ptr<RNTupleModel> model = nullptr);
 
    /////////////////////////////////////////////////////////////////////////////
    /// \brief Create a new RNTuple processor for horizontallly concatenated RNTuples.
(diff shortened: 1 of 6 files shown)


tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -181,9 +181,13 @@ ROOT::Experimental::RNTupleChainProcessor::RNTupleChainProcessor(const std::vect
    if (fNTuples.empty())
       throw RException(R__FAIL("at least one RNTuple must be provided"));
 
+   fInnerNEntries.assign(ntuples.size(), kInvalidNTupleIndex);
+
    fPageSource = Internal::RPageSource::Create(fNTuples[0].fNTupleName, fNTuples[0].fStorage);
    fPageSource->Attach();
 
+   fInnerNEntries.at(0) = fPageSource->GetNEntries();
```
silverweed: This doesn't need to be a checked access since you are ensuring `ntuples` has at least 1 element a few lines above

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -231,26 +242,33 @@ ROOT::Experimental::NTupleSize_t ROOT::Experimental::RNTupleChainProcessor::Conn
...
 
-   if (fLocalEntryNumber >= fPageSource->GetNEntries()) {
-      do {
-         if (++fCurrentNTupleNumber >= fNTuples.size()) {
-            return kInvalidNTupleIndex;
-         }
-         // Skip over empty ntuples we might encounter.
-      } while (ConnectNTuple(fNTuples.at(fCurrentNTupleNumber)) == 0);
+   while (localEntryNumber >= fInnerNEntries.at(currentNTuple)) {
+      localEntryNumber -= fInnerNEntries.at(currentNTuple);
```
silverweed: I'm not a big fan of all these repeated `.at`s: I would argue it's better to check the condition once at the start of the function (throwing a helpful error message if the number is out of bounds) and then use the regular `[]` access throughout the rest of the function.

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -136,8 +138,12 @@ protected:
    /// \return `entryNumber` if the entry was successfully loaded, `kInvalidNTupleIndex` otherwise.
    virtual NTupleSize_t LoadEntry(NTupleSize_t entryNumber) = 0;
 
-   RNTupleProcessor(const std::vector<RNTupleOpenSpec> &ntuples)
-      : fNTuples(ntuples), fNEntriesProcessed(0), fCurrentNTupleNumber(0), fCurrentEntryNumber(0)
+   RNTupleProcessor(const std::vector<RNTupleOpenSpec> &ntuples, std::unique_ptr<RNTupleModel> model)
+      : fNTuples(ntuples),
+        fModel(std::move(model)),
+        fNEntriesProcessed(0),
```
silverweed: These default initializations should be inlined with the field declarations

EDIT: I see this is done in the following commit; consider squashing the changes

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -136,8 +138,12 @@ protected:
    /// \return `entryNumber` if the entry was successfully loaded, `kInvalidNTupleIndex` otherwise.
    virtual NTupleSize_t LoadEntry(NTupleSize_t entryNumber) = 0;
 
-   RNTupleProcessor(const std::vector<RNTupleOpenSpec> &ntuples)
-      : fNTuples(ntuples), fNEntriesProcessed(0), fCurrentNTupleNumber(0), fCurrentEntryNumber(0)
+   RNTupleProcessor(const std::vector<RNTupleOpenSpec> &ntuples, std::unique_ptr<RNTupleModel> model)
```
silverweed, line 155: I'd write a small comment to document that the model can be null and what happens in that case

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -171,33 +213,19 @@ void ROOT::Experimental::RNTupleSingleProcessor::Connect()
 
 //------------------------------------------------------------------------------
 
-ROOT::Experimental::RNTupleChainProcessor::RNTupleChainProcessor(const std::vector<RNTupleOpenSpec> &ntuples,
-                                                                 std::unique_ptr<RNTupleModel> model)
-   : RNTupleProcessor(ntuples, std::move(model))
+ROOT::Experimental::RNTupleChainProcessor::RNTupleChainProcessor(
+   std::vector<std::unique_ptr<RNTupleProcessor>> &processors, std::unique_ptr<RNTupleModel> model)
```
silverweed, line 217: `processors` should be passed by value since the caller is relinquishing its ownership

silverweed (changes requested):
I left a couple of, mostly minor, comments.
In general lgtm but there are some changes introduced in one commit that are changed by later commits, so it's a bit hard to review commit-wise.

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -207,8 +218,8 @@ public:
          : fProcessor(processor), fCurrentEntryNumber(entryNumber)
       {
          // This constructor is called with kInvalidNTupleIndex for RNTupleProcessor::end(). In that case, we already
-         // know there is nothing to advance to.
-         if (processor.GetCurrentEntryNumber() != kInvalidNTupleIndex) {
+         // know there is nothing to load.
+         if (fCurrentEntryNumber != kInvalidNTupleIndex) {
```
jblomer, line 222: Ah, here it is. Perhaps move to the previous commit.

silverweed (approved):

silverweed:
lgtm but better wait for other people's approval as well

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -138,6 +142,16 @@ protected:
...
    virtual NTupleSize_t LoadEntry(NTupleSize_t entryNumber) = 0;
 
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Points processor entry's field values to those in the provided entry.
+   ///
+   /// \param[in] entry The entry whose field values to use.
+   virtual void SetEntryPointers(const REntry &entry) = 0;
+
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Get the total number of entries in this processor
```
jblomer, line 152: Perhaps add a comment that this is costly for a chain processor (all underlying files need to be opened). I guess this expensive operation is only necessary if a chain is built from other chains..?

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -155,6 +188,16 @@ ROOT::Experimental::NTupleSize_t ROOT::Experimental::RNTupleSingleProcessor::Loa
    return entryNumber;
 }
 
+void ROOT::Experimental::RNTupleSingleProcessor::SetEntryPointers(const REntry &entry)
+{
+   for (const auto &value : *fEntry) {
+      auto &field = value.GetField();
+      auto valuePtr = entry.GetPtr<void>(field.GetQualifiedFieldName());
```
jblomer, line 195: Not for this PR: but we should implement/find a way to iterate over all the pointers of an entry without all the field lookups.

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -206,63 +235,65 @@ ROOT::Experimental::RNTupleChainProcessor::RNTupleChainProcessor(const std::vect
...
 ROOT::Experimental::NTupleSize_t ROOT::Experimental::RNTupleChainProcessor::LoadEntry(NTupleSize_t entryNumber)
 {
    NTupleSize_t localEntryNumber = entryNumber;
-   size_t currentNTuple = fCurrentNTupleNumber;
+   size_t currProcessor = 0;
 
-   while (localEntryNumber >= fInnerNEntries[currentNTuple]) {
-      localEntryNumber -= fInnerNEntries[currentNTuple];
+   // Determine to which inner processer and local entry number the provided global entry number belongs.
+   while (localEntryNumber >= fInnerNEntries[currProcessor]) {
```
jblomer, line 280: Maybe assert that `fInnerNEntries[0] != kInvalidNTupleIndex`

jblomer (commented):
That's great! I have some questions on details.

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -32,18 +32,14 @@ void EnsureUniqueNTupleNames(const std::vector<RNTupleOpenSpec> &ntuples)
...
-   auto model = pageSource->GetSharedDescriptorGuard()->CreateModel();
-   return RNTupleProcessor::Create(ntuple, *model);
-}
-
-std::unique_ptr<ROOT::Experimental::RNTupleProcessor>
-ROOT::Experimental::RNTupleProcessor::Create(const RNTupleOpenSpec &ntuple, RNTupleModel &model)
-{
-   return std::unique_ptr<RNTupleSingleProcessor>(new RNTupleSingleProcessor(ntuple, model));
+   if (!model) {
+      auto pageSource = Internal::RPageSource::Create(ntuple.fNTupleName, ntuple.fStorage);
```
hahnjo, line 38: This means creating the page source twice, can this be delayed until the processor is connected (as the chain and join processors do)?

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
hahnjo: Even then, you may be able to get around getting the number of entries in the inner chain by just trying to load an entry from the chain and have the inner `LoadEntry` fail in case it's past the last file.

tree/ntuple/v7/src/RNTupleProcessor.cxx
enirolf: In principle this is offered through `REntry::ConstIterator_t` (and used on `fEntry` in l.193), no? In this specific case the field lookup is needed because a pointer from the entry *not* being iterated over is needed.

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -223,34 +232,45 @@ ROOT::Experimental::NTupleSize_t ROOT::Experimental::RNTupleChainProcessor::Conn
...
+
+      if (++currentNTuple >= fNTuples.size())
+         return kInvalidNTupleIndex;
 
-      fLocalEntryNumber = 0;
+      if (fInnerNEntries[currentNTuple] == kInvalidNTupleIndex) {
+         auto tmpPageSource =
+            Internal::RPageSource::Create(fNTuples[currentNTuple].fNTupleName, fNTuples[currentNTuple].fStorage);
+         tmpPageSource->Attach();
+         fInnerNEntries[currentNTuple] = tmpPageSource->GetNEntries();
```
hahnjo: For future optimizations:
1. If we actually find the entry in this page source, we should not destroy it. On the other hand, we don't need the full `ConnectNTuple` call for intermediate ntuples that we just skip over.
2. If you store the cumulative number of entries, you can
a. check if `entryNumber` is smaller than the highest cached value in `fInnerNEntries`, only if not you have to linearly go through the ntuples.
b. If the `entryNumber` is contained, you can do binary search in the cached `fInnerNEntries`

(doesn't need to be in this PR)

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -122,35 +128,22 @@ protected:
 
    NTupleSize_t fNEntriesProcessed;  //< Total number of entries processed so far
    std::size_t fCurrentNTupleNumber; //< Index of the currently open RNTuple
-   NTupleSize_t fLocalEntryNumber;   //< Entry number within the current ntuple
+   NTupleSize_t fCurrentEntryNumber; //< Entry number within the current ntuple
```
hahnjo: (just for my understanding: this is already `Current processor entry number` as of this commit, right?)

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
hahnjo: I'm not sure I understand: If you have a join of a chain (or maybe better to discuss a chain of chains?) with 100 files of 100 entries and you need entry 150, it's not important to know that there are 10000 entries in total. You only need to know that you find entry 150 in the second processor of the chain. How I imagine this to work in my simplified view is:
1. Ask the chain to load entry 150.
2. The chain asks its first processor, which opens the ntuple and caches that it has 100 entries. It returns saying that it could not find the entry.
3. At that point, the chain asks the number of entries (which is cached), and proceeds to ask the second processor to load local entry 50, which succeeds.

jblomer: Not necessarily for this PR: I think eventually it would make sense to relax this method, e.g. to "entries encountered so far". It's good that the method is protected, so we have freedom to optimize it at a later point.

hahnjo: > But even then, we still need some interface to get this cached number of entries, no?

Yes, we do.

> And given (let's assume were dealing with a chain of chains) that in `RNTupleChainProcessor` `GetNEntries` is only called *after* calling `LoadEntry` on this inner processor (which is a chain), it will just return this cached value.

Yes, that was the idea. Because `LoadEntry` has to query all its inner processors to determine that it cannot load the entry. That's still not fully taken advantage of in the current code, we may want to do
```c++
   while (fInnerProcessors[currProcessor]->LoadEntry(localEntryNumber) == kInvalidNTupleIndex) {
      if (fInnerNEntries[currProcessor] == kInvalidNTupleIndex) {
         fInnerNEntries[currProcessor] = fInnerProcessors[currProcessor]->GetNEntries();
      }
      localEntryNumber -= fInnerNEntries[currProcessor];

      // The provided global entry number is larger than the number of available entries.
      if (++currProcessor >= fInnerProcessors.size())
         return kInvalidNTupleIndex;
   }
```

But as discussed, this goes into performance optimization territory and not needed for this PR. I'll resolve this thread.

hahnjo (approved):

PR #17399, opened 2025-01-10
[python] Pythonize `TObject::Clone()` overrides for correct ownership

guitargeek:
Following up on #17250, also pythonizing the `TObject::Clone()` overrides to automatically give up the ownership.

Thanks to the following forum post for noticing this problem: https://root-forum.cern.ch/t/a-question-about-root/62617


pcanal (approved):
LGTM

guitargeek:
Backported to 6.34 in #17484.

PR #17401, opened 2025-01-10
[df] Improve warning message when lazy Snapshot is not triggered

vepadulano:
The current message is not clear enough and may lead to confusing situations such as the one reported at https://root-forum.cern.ch/t/some-problem-with-lazy-snapshots-in-rdataframe/46740


martamaja10 (approved):
Very nice, thanks Vincenzo!

PR #17405, opened 2025-01-10
[windows] Fix root-config.bat for PowerShell

bellenot:
Fix an issue with `root-config.bat` when used in PowerShell like:
`Invoke-Expression "cl test.cpp $(root-config --cflags --libs)"`
as reported [on the Forum](https://root-forum.cern.ch/t/root-6-34-windows-compilation-w32pragma-issue-cea-43597/62630)


pcanal (approved):
LGTM

PR #17406, opened 2025-01-10
[cling] Move static init function renaming to BackendPasses.cpp

devajithvs:
# This Pull request:

## Checklist:

- [x] tested changes locally

This PR removes the need for the clang patch: https://github.com/root-project/llvm-project/commit/d84a0e3104c5 

This is already tested with `roottest-cling-staticinit-ROOT-7775`


interpreter/cling/lib/Interpreter/BackendPasses.cpp
```diff
@@ -420,6 +462,7 @@ void BackendPasses::CreatePasses(int OptLevel, llvm::ModulePassManager& MPM,
                                  StandardInstrumentations& SI) {
 
   // TODO: Remove this pass once we upgrade past LLVM 19 that includes the fix.
+  MPM.addPass(UniqueInitFunctionNamePass());
```
devajithvs, line 465: Oops. Didn't notice the comment. I'll move the comment down when I push to update LLVM tag.

dpiparo (approved):

interpreter/cling/lib/Interpreter/BackendPasses.cpp
```diff
@@ -40,6 +40,48 @@ using namespace cling;
...
+        if (!isPreprocessingNumberBody(NewFunctionName[i]))
+          NewFunctionName[i] = '_';
+      }
+
+      return NewFunctionName;
+    }
+
+    // make static initialization function names (__cxx_global_var_init) unique
+    bool runOnFunction(Function& F, const StringRef ModuleName) {
+      if (F.hasName() && F.getName().startswith("__cxx_global_var_init")) {
```
hahnjo, line 66: `llvm::StringRef::startswith` is deprecated, we should call `starts_with`.

(There are a couple of other uses of deprecated functions in our downstream patches and `core/metacling`. They are on my list of things to fix in January and also remove `-Wno-deprecated-declarations` to make the compiler warn about them.)

hahnjo (commented):
FWIW still needs a tag (we'll figure out a sequencing between all pending changes) and we should either reverse the order of commits or squash them so that the test passes at all points in the commit history.

devajithvs:
@hageboeck I've updated the tag here: https://github.com/root-project/llvm-project/releases/tag/ROOT-llvm18-20250116-01 to also backport https://github.com/llvm/llvm-project/pull/101761

interpreter/cling/lib/Interpreter/BackendPasses.cpp
hahnjo: Not addressed yet AFACIT; the comment applies directly to `WorkAroundConstructorPriorityBugPass` and should move there...

hahnjo (approved):
LGTM, after moving the comment as indicated.

> I've updated the tag here: https://github.com/root-project/llvm-project/releases/tag/ROOT-llvm18-20250116-01 to also backport https://github.com/llvm/llvm-project/pull/101761

Ideally I would have preferred this to be a separate PR because it has nothing to do with moving the renaming pass and dropping a Clang patch. Also in the future, please use `cherry-pick -x` for upstream commits in https://github.com/root-project/llvm-project/ so that we record the upstream commit hash. (no need to create another tag right now, we can fix it if we ever need to touch upstream commits again)

devajithvs:
> LGTM, after moving the comment as indicated.
> 
> > I've updated the tag here: https://github.com/root-project/llvm-project/releases/tag/ROOT-llvm18-20250116-01 to also backport [llvm/llvm-project#101761](https://github.com/llvm/llvm-project/pull/101761)
> 
> Ideally I would have preferred this to be a separate PR because it has nothing to do with moving the renaming pass and dropping a Clang patch. Also in the future, please use `cherry-pick -x` for upstream commits in https://github.com/root-project/llvm-project/ so that we record the upstream commit hash. (no need to create another tag right now, we can fix it if we ever need to touch upstream commits again)

Thanks Jonas, my bad, I've created a new tag with just the change needed for the PR.  I will  make a separate PR with the backport later.

PR #17413, opened 2025-01-13
[geom][mini] Disabling (temporarily) c++20 [[unlikely]] compiler hint.

agheata:
# This Pull request:
Disables a c++20 [[unlikely]] hint to prevent a `-Wattributes` compilation warning

## Changes or fixes:
I just commented, not removed entirely, to disable the warning on c++17 or less mode. We have `R__unlikely`, which cannot be used in a multiple-branch statement.

## Checklist:

- [x] tested changes locally

This PR fixes #

### DIFF
--- a/geom/geom/src/TGeoParallelWorld.cxx
+++ b/geom/geom/src/TGeoParallelWorld.cxx
@@ -40,10 +40,13 @@ method.
 #include <mutex>
 
 // this is for the bvh acceleration stuff
+#ifdef __GNUC__
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wall"
 #pragma GCC diagnostic ignored "-Wshadow"
 #pragma GCC diagnostic ignored "-Wunknown-pragmas"
+#pragma GCC diagnostic ignored "-Wattributes"
+#endif
 
 // V2 BVH
 #include <bvh/v2/bvh.h>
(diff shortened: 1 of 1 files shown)


geom/geom/inc/bvh/v2/bvh.h
```diff
@@ -148,8 +148,11 @@ void Bvh<Node>::traverse_top_down(Index start, Stack& stack, LeafFn&& leaf_fn, I
                 top = near_index;
             } else if (hit_right) {
                 top = right.index;
-            }
-            else [[unlikely]] {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wattributes"
+            // Prevent gcc warnings if the compiler is not c++20 compliant
+            } else [[unlikely]] {
+#pragma GCC diagnostic pop
```
bellenot, line 155: To prevent warnings on non-gcc compilers:
```suggestion
#if defined(__GNUC__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wattributes"
            // Prevent gcc warnings if the compiler is not c++20 compliant
            } else [[unlikely]] {
#pragma GCC diagnostic pop
#endif
```
or you can also use `#if !defined(_MSC_VER)`, but I don't know for other compilers...

bellenot (changes requested):

PR #17419, opened 2025-01-13
[webgui] let configure custom JSROOT `settings` with RWebWindow

linev:
There are different settings parameters for JSROOT, 
which normally changed from JS code. 
But some of these parameters need to be configured from C++ - like image resolution or usage of shortcut buttons.
Now it will be possible to done with method:
```
ROOT::RWebWindow::SetJSROOTSettings("{ ToolBar: false, CanEnlarge: false, CanvasScale: 2 }");
```


bellenot (approved):
LGTM

PR #17423, opened 2025-01-14
[cmake] Deprecate `mysql`, `odbc`, `pgsql` build options

hahnjo:
All of them require (optional) dependencies for very little use.

FYI @andresailer


cern-sft-spi:
Thanks for the heads up.
We still had mysql=ON.

It is enough to turn the options OFF, right?

jblomer:
I'm in favor of deprecating mysql, odbc, pgsql. But not sqlite. It is used, e.g., by cvmfs to use RDataFrame on sqlite data.

hahnjo:
Oh yes, my bad: I didn't realize `sqlite` was also used in places other than `sql`. I reverted the change to the `sqlite` option, we can decide independently if we want to deprecate `sql/sqlite` and the various `TSQL*` classes spread between `io/sql` and `net/net`. For the moment, I'm focusing on the "easy" deprecations...

README/ReleaseNotes/v636/index.md
```diff
@@ -37,6 +37,7 @@ The following people have contributed to this new version:
 * The `RooDataSet` constructors to construct a dataset from a part of an existing dataset are deprecated and will be removed in ROOT 6.38. This is to avoid interface duplication. Please use `RooAbsData::reduce()` instead, or if you need to change the weight column, use the universal constructor with the `Import()`, `Cut()`, and `WeightVar()` arguments.
 * The ROOT splash screen was removed for Linux and macOS
 * Proof support has been completely removed form RooFit and RooStats, after it was already not working anymore for several releases
+* The build options `mysql`, `odbc`, and `pgsql` have been deprecated. Please complain with root-dev@cern.ch should you still need one!
```
guitargeek, line 40: Should we explicitly mention the ROOT version where it will be removed here, to be consistent with the CMake deprecation warning? (which says "will be removed in the next release of ROOT")

hahnjo: In my personal opinion, we want to be flexible here: Users will only see the deprecation notice when they upgrade to 6.36. In order to give them some time to complain, we should not remove it right after the release and maybe not even for 6.38...

bellenot:
AFAIK, the `odbc` one is mostly for Windows and comes for free (i.e. It is available when installing Visual Studio). So we could maybe keep it? Or is it too much of a overhead?

README/ReleaseNotes/v636/index.md
pcanal: I agree with @hahnjo, those may be quietly used and we don't know and if there used we may be asked by the user to either keep them forever and remove them only within a certain version-frame (timeframe)

guitargeek: Ok makes sense and I agree! Just wanted to make sure you had the same thought process when making this PR

hahnjo:
> AFAIK, the `odbc` one is mostly for Windows and comes for free (i.e. It is available when installing Visual Studio). So we could maybe keep it? Or is it too much of a overhead?

I think the main question for me is whether it is used. Do we know of anybody relying on it? Eventually I'd personally like to see all SQL-related classes deprecated and removed from ROOT because I don't think we should encourage their use.

README/ReleaseNotes/v636/index.md
hahnjo: So, can we move ahead and deprecate these three options? I have more in the pipeline that I'd like to tackle next...

bellenot:
> > AFAIK, the `odbc` one is mostly for Windows and comes for free (i.e. It is available when installing Visual Studio). So we could maybe keep it? Or is it too much of a overhead?
> 
> I think the main question for me is whether it is used. Do we know of anybody relying on it? Eventually I'd personally like to see all SQL-related classes deprecated and removed from ROOT because I don't think we should encourage their use.

OK fair enough, let's deprecate it and we'll see

guitargeek (approved):
LGTM!

bellenot (approved):

dpiparo (approved):

PR #17425, opened 2025-01-14
[CMake] Fix a few small issues

hageboeck:
- Add a few missing labels to unittests
- Add warnings or errors when ROOT is configured in a way that cannot work (e.g. asking for webdisplay without http)

### DIFF
--- a/cmake/modules/RootMacros.cmake
+++ b/cmake/modules/RootMacros.cmake
@@ -1508,7 +1508,7 @@ set(ROOT_TEST_DRIVER ${CMAKE_CURRENT_LIST_DIR}/RootTestDriver.cmake)
 #                        [LABELS label1 label2]
 #                        [PYTHON_DEPS numpy numba keras torch ...] # List of python packages required to run this test.
 #                                                              A fixture will be added the tries to import them before the test starts.
-#                        [PROPERTIES prop1 value1 prop2 value2...] 
+#                        [PROPERTIES prop1 value1 prop2 value2...]
 #                       )
 #
 function(ROOT_ADD_TEST test)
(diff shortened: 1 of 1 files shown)


tutorials/CMakeLists.txt
```diff
@@ -871,10 +871,12 @@ if(ROOT_pyroot_FOUND)
       analysis/dataframe/df032_RDFFromNumpy.py
       math/fit/combinedFit.py
       math/fit/multifit.py
+      math/fit/NumericalMinimization.py
```
pcanal, line 874: Unrelated to this PR per se.  Do you know why we use `FILE(GLOB ....` here instead of just a hard coded list?

hageboeck: I guess the strategy is copied from a different location, and the advantage of `GLOB` is obviously the support for expressions like:
`math/fit/*_numpy.py`

hageboeck: Here is an example:
https://github.com/root-project/root/blob/d72ed64dfbbc38f4a5acec7003ac995a68e583d4/tutorials/CMakeLists.txt#L751

I guess people just use globs for all the vetos, so you can switch between list and globs.

bellenot (approved):
LGTM, thanks!

tutorials/CMakeLists.txt
pcanal: Glob are ... annoying in term of their state in regard to adding and removing files ... so we were moving away from them but this example looks new .. :(

hahnjo:
After getting this change from a `git pull`, I have to say that I find all the CMake messages quite annoying because they appear in every configuration step. Can they be shown only on the first invocation and suppressed afterwards?

hageboeck:
> After getting this change from a `git pull`, I have to say that I find all the CMake messages quite annoying because they appear in every configuration step. Can they be shown only on the first invocation and suppressed afterwards?

Let me check if we can save this knowledge.

PR #17426, opened 2025-01-14
[docu] Use CMake-generated Doxyfile instead of hard-coded one

ferdymercury:
# This Pull request:

## Changes or fixes:

It's quite hard to have an overview of a 3k-line-long Doxyfile, and it needs constant commits "update to version 1.9.8", etc. I propose here instead to use a CMake-generated Doxyfile, where you only declare those flags that are different from the default values. (as found out by command `doxygen -x_noenv` on the preexisting Doxyfile)

## Checklist:

- [x] tested changes locally


couet:
Hi @ferdymercury,

Thank you for this PR! We are currently in the process of transitioning the daily documentation build from Jenkins to GitHub Actions. As part of this effort, we have created a temporary folder named doxygen_s3. This setup is not yet finalized and remains unstable for now. I suggest we revisit your PR once this work is completed. Is that fine with you ?

ferdymercury:
Perfect, sounds good!

ferdymercury:
gentle ping @couet

ferdymercury:
@hageboeck this approach would be easier to maintain than keeping track of two Doxy files and their constant updates for each version of doxygen used

PR #17431, opened 2025-01-15
[ntuple] use MakeUninitArray when appropriate in tests

silverweed:
Minor optimization: use `MakeUninitArray` instead of `make_unique<T[]>` when the buffers don't need to be initialized.

## Checklist:

- [x] tested changes locally


hahnjo (approved):
LGTM; note that tests are not really performance critical (for the array sizes that are allocated here), so it's not really that important if we miss one occurrence

PR #17432, opened 2025-01-15
[windows] Update LZMA and use CMake to build it

bellenot:
Update LZMA to 5.6.3 and build from source also on Windows


hageboeck (approved):
LGTM!

PR #17441, opened 2025-01-17
[ntupleutil] exporter: check that the output file is writable

silverweed:
# This Pull request:
adds a check to RNTupleExporter that the given output path can be written to (currently it fails silently if not).

## Checklist:

- [x] tested changes locally


pcanal (approved):
LGTM.  The git log could be improve with "[ntupleutil] exporter: error out when output file is not writable"

PR #17443, opened 2025-01-17
Fix two includes to compile with gcc-15

hageboeck:
When fixed-size integer types are in use, gcc-15 requires the inclusion of `<cstdint>`.


lmoneta (approved):
LGTM!

PR #17446, opened 2025-01-17
Implement TGraph2D::RemoveDuplicates

couet:
The Delaunay triangulation algorithm assumes that each (x, y) coordinate corresponds to a unique z value,
meaning duplicate (x, y) points are not allowed. Consequently, when using drawing options that rely on this
algorithm (e.g., TRI, SURF, etc.), a warning may appear instructing you to remove duplicates.
This function provides a way to handle such duplicates.

### DIFF
--- a/hist/hist/src/TGraph2D.cxx
+++ b/hist/hist/src/TGraph2D.cxx
@@ -1447,26 +1447,12 @@ Int_t TGraph2D::RemovePoint(Int_t ipoint)
 {
    if (ipoint < 0) return -1;
    if (ipoint >= fNpoints) return -1;
-
-   fNpoints--;
-   Double_t *newX = new Double_t[fNpoints];
-   Double_t *newY = new Double_t[fNpoints];
-   Double_t *newZ = new Double_t[fNpoints];
-   Int_t j = -1;
-   for (Int_t i = 0; i < fNpoints + 1; i++) {
-      if (i == ipoint) continue;
-      j++;
-      newX[j] = fX[i];
-      newY[j] = fY[i];
-      newZ[j] = fZ[i];
+   for (Int_t i = ipoint; i < fNpoints - 1; i++) {
+      fX[i] = fX[i+1];
+      fY[i] = fY[i+1];
+      fZ[i] = fZ[i+1];
    }
-   delete [] fX;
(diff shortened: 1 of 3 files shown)


math/mathcore/src/Delaunay2D.cxx
```diff
@@ -179,6 +180,7 @@ void Delaunay2D::DoFindTriangles() {
    std::vector<CDT::V2d<double>> points(fNpoints);
    for (i = 0; i < fNpoints; ++i) points[i] = CDT::V2d<double>::make(fXN[i], fYN[i]);
    CDT::RemoveDuplicates(points);
+   if (fNpoints-points.size() > 0) Warning("DoFindTriangles","This TGraph2D has duplicated vertices. To remove them call RemoveDuplicates before drawing");
```
bellenot, line 183:
```suggestion
   if (fNpoints-points.size() > 0)
      Warning("DoFindTriangles","This TGraph2D has duplicated vertices. To remove them call RemoveDuplicates before drawing");
```

hist/hist/src/TGraph2D.cxx
```diff
@@ -1410,6 +1412,34 @@ TH1 *TGraph2D::Project(Option_t *option) const
...
+/// g->Draw("TRI1");
+/// ~~~
+
+Int_t TGraph2D::RemoveDuplicates()
+{
+   for (int i=0; i<fNpoints; i++) {
+      double x = fX[i];
+      double y = fY[i];
+      for (int j=i+1; j<fNpoints; j++) {
+         if (x==fX[j] && y==fY[j]) {RemovePoint(j); j--;}
```
dpiparo, line 1435: Won't this cause 3 deallocations, 3 allocations and 3 copies per point removed?
Can we:
- Avoid allocations and deallocations in `RemovePoint` by simply decrementing `fNPoints`?
- Add a private method, `RemovePoints`, that removes points in bulk, iterating on the indices to remove? `RemovePoint` can invoke the bulk method internally for 1 point.

couet: Ok, I will investigate the way `RemovePoint` is coded.

couet:
seems to me it can be merged

PR #17448, opened 2025-01-17
[tutorials] Move some graph and hist tutorials to better suiting categories  

martamaja10:
After major modernisation campaign regarding tutorials, it was noticed that some of them would better fit different categories.


dpiparo (approved):

martamaja10:
failing tests on ubuntu are io related and irrelevant to the changes from this PR. I will merge now.

PR #17450, opened 2025-01-17
[rbrowser] share single connection between all widgets

linev:
Up to now each widget as `TWebCanvas` or `RGeomViewer` was establishing separate websocket connection with the server. This consumes extra resources and takes longer time to reload RBrowser.

Now `RWebWindow` embedding will be used. For that one need to provide access to `RWebWindow` for each of concerned widgets.  

Improve handling of `catched` web window. For case of `TCanvas` and `RCanvas` find related object and handle it properly. This allows to handle properly canvas created from ROOT prompt or from ROOT macro.

Update JSROOT which also has to support such new feature


linev:
> Please check and fix the warnings

Thanks to point out it. I rebase last commit - now should be ok

Issue #17444, opened 2025-01-17
ROOT doesn't compile with gcc-15


hageboeck:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Starting from gcc-15, `<cstdint>` needs to be included when fixed-size integer types are used.
See also:
https://bugs.gentoo.org/942438

In ROOT, there is currently four places that create problems:
- [x] Two throughout ROOT (#17443)
- [ ] Two in the interpreter.
  LLVM fixed those [here 7e44305](https://github.com/llvm/llvm-project/commit/7e44305041d96b064c197216b931ae3917a34ac1) and as a side effect [here 2222fdd](https://github.com/llvm/llvm-project/commit/2222fddfc0a2ff02036542511597839856289094#diff-3a97dc53c7250e03c83d10c7bd769776dc47bdcc170fddcd49f92f8b12f1e5f7R16). The former applies cleanly, but the latter doesn't.
   **Update**: We can backport https://github.com/llvm/llvm-project/pull/123320 instead of 2222fdd if the PR gets merged. https://github.com/llvm/llvm-project/commit/7abf44069aec61eee147ca67a6333fc34583b524
  **Update 2**: This is now #17510 
- [x] One more location found by mellert: https://github.com/root-project/root/pull/17455

What could be the solutions to make ROOT compile cleanly?
1. Wait until the llvm upgrade?
2. Patch llvm ourselves until the upgrade?
3. Pull in already now the commits from llvm-main? (the latter doesn't apply, though)

@devajithvs @hahnjo @amadio 

### Reproducer

Compile ROOT with gcc-15

### ROOT version

Master, but also 6.34, 6.32 as can be seen in the gentoo bug report

### Installation method

Source

### Operating system

Gentoo

hahnjo:
> LLVM fixed those [here](https://github.com/llvm/llvm-project/commit/7e44305041d96b064c197216b931ae3917a34ac1) and as a side effect [here](https://github.com/llvm/llvm-project/commit/2222fddfc0a2ff02036542511597839856289094#diff-3a97dc53c7250e03c83d10c7bd769776dc47bdcc170fddcd49f92f8b12f1e5f7R16). The former applies cleanly, but the latter doesn't.

[...]

PR #17455, opened 2025-01-19
Fixes for compiling with GCC 15 (Fedora 42)

ellert:
# This Pull request:

These changes make it possible to compile root 6.34.02 using gcc 15 on Fedora 42.

Also the following commits need to be backported from master:

commit 2884f67f5a145ef5d2ba1f8050b2b1a543d97881
commit cfc922131c8a977e14c6611ed991f0889b0c47c8

... and the following commits from LLVM:

commit llvm/llvm-project@7e44305041d96b064c197216b931ae3917a34ac1
commit llvm/llvm-project@7abf44069aec61eee147ca67a6333fc34583b524 (see PR llvm/llvm-project#123320)

### DIFF
--- a/misc/minicern/src/cernlib.c
+++ b/misc/minicern/src/cernlib.c
@@ -172,8 +172,8 @@ int cfstati_(char *fname, int *info, int *lgname)
 #endif
 {
    struct stat buf;
-   char *ptname, *fchtak();
-   int istat=-1, stat();
+   char *ptname;
+   int istat = -1;
    ptname = fchtak(fname,*lgname);
    if (ptname == ((void *)0)) return -1;
    istat = stat(ptname, &buf);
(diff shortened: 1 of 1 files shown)


hageboeck:
Hi @ellert,

thanks for the initiative!

We have a few things to deal with:
1. Would you mind running git-clang-format on the "conflicting functions" commit?
2. Regarding the LLVM includes, we were already discussing in https://github.com/root-project/root/issues/17444 and https://github.com/llvm/llvm-project/pull/123320, but the situation is a bit complicated:
  - In order not to deviate from llvm (which allows for much faster future upgrades), we are not simply committing to the llvm tree.
  - Instead, we try to upstream patches to llvm, and backport afterwards. This will make them disappear during a rebase. (This is why there is a CI failure.)
  - For this reason, we have to wait for https://github.com/llvm/llvm-project/pull/123320 and backport.

Number 2 is a slow process, but we are working on it. So if you want to get out of this in a timely manner, I would propose to retain and clang-format the "conflicting functions" commit, and remove the LLVM commit. This way, we can merge this PR quickly, and we'll take care of the llvm business.

Let us know what you think.

hageboeck:
@devajithvs FYI

hageboeck (approved):
Great, thank you!

The LLVM commits are on their way, and we'll take care of putting all of these changes into the next release.
I'm just waiting for one last CI job before merging.

PR #17464, opened 2025-01-21
[cmake] deprecate qt5web option

linev:
Qt5 WebEngine uses too old `chromium`,
which is not working with openui5 library.
Therefore it make no big sense to support it in ROOT.

`qt5web` is only can be used for `TWebCanvas` with "noopenui" option, 
but it is not normal mode.

For any future developments `qt6web` with Qt6 WebEngine has to be used.


bellenot:
@andresailer FYI

bellenot (approved):

bellenot:
@cern-sft-spi FYI

PR #17471, opened 2025-01-21
[ntuple] Add deprecated EColumnType using directive

jblomer:


tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
```diff
@@ -313,6 +313,10 @@ RResult<void> EnsureValidNameForRNTuple(std::string_view name, std::string_view
 
 } // namespace Internal
 
+// TODO(jblomer): remove before branching ROOT v6.36
+using EColumnType [[deprecated("ROOT::Experimental::EColumnType moved to ROOT::Experimental::ENTupleColumnType")]] =
```
silverweed, line 317: Should probably use `R__DEPRECATED`

hahnjo: With `R__DEPRECATED`, we mark functions that are deprecated in a particular release and will be removed after. Here, we want to deprecate now and remove before 6.36. This would trigger `_R__DEPRECATED_636` expanding to `_R_DEPRECATED_REMOVE_NOW` and triggering an error. I think here it's ok to use the C++ attribute `deprecated` directly, it is standard since C++14...

hahnjo (approved):

PR #17473, opened 2025-01-21
[ntuple] Merger: pass the destination in the ctor rather than in Merge()

silverweed:
RNTupleMerger should share the lifetime of its destination file to better support incremental merging, e.g.

```c++
// open the destination file once
auto destination = CreatePageSink();
RNTupleMerger merger { *destination };
auto opts = RNTupleMergeOptions{};
opts.fMergingMode = ENTupleMergingMode::kUnion;
// keep merging stuff into it as it arrives (e.g. from the network)
while (auto result = AcquireSources()) {
  merger.Merge(result->sources, opts);
}
```

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleMerger.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleMerger.hxx
@@ -86,6 +86,7 @@ struct RNTupleMergeOptions {
  */
 // clang-format on
 class RNTupleMerger final {
+   std::unique_ptr<RPageSink> fDestination;
    std::unique_ptr<RPageAllocator> fPageAlloc;
    std::optional<TTaskGroup> fTaskGroup;
 
@@ -97,11 +98,11 @@ class RNTupleMerger final {
                             std::span<RColumnMergeInfo> extraDstColumns, RNTupleMergeData &mergeData);
 
 public:
-   RNTupleMerger();
+   /// Creates a RNTupleMerger with the given destination.
+   explicit RNTupleMerger(std::unique_ptr<RPageSink> destination);
 
    /// Merge a given set of sources into the destination.
-   RResult<void> Merge(std::span<RPageSource *> sources, RPageSink &destination,
-                       const RNTupleMergeOptions &mergeOpts = RNTupleMergeOptions());
+   RResult<void> Merge(std::span<RPageSource *> sources, const RNTupleMergeOptions &mergeOpts = RNTupleMergeOptions());
 
 }; // end of class RNTupleMerger
 
--- a/tree/ntuple/v7/src/RNTupleMerger.cxx
+++ b/tree/ntuple/v7/src/RNTupleMerger.cxx
@@ -215,7 +215,7 @@ try {
    }
 
    // Now merge
-   RNTupleMerger merger;
+   RNTupleMerger merger{std::move(destination)};
    RNTupleMergeOptions mergerOpts;
    mergerOpts.fCompressionSettings = *compression;
    mergerOpts.fExtraVerbose = extraVerbose;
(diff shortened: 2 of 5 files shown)


tree/ntuple/v7/inc/ROOT/RNTupleMerger.hxx
```diff
@@ -97,11 +98,12 @@ class RNTupleMerger final {
                             std::span<RColumnMergeInfo> extraDstColumns, RNTupleMergeData &mergeData);
 
 public:
-   RNTupleMerger();
+   /// Creates a RNTupleMerger with the given destination.
+   /// `destination` must outlive the RNTupleMerger.
+   explicit RNTupleMerger(RPageSink &destination);
```
pcanal, line 103: Wouldn't passing a `std::shared_ptr` be a better indication of the require lifetime?

silverweed: I don't think so, a) because it's clear that the merger should have a strictly smaller lifetime than the Sink it's operating on (whereas the `shared_ptr` would blur this concept) and, more importantly b) our API always returns PageSinks/Sources as `unique_ptr`.

tree/ntuple/v7/inc/ROOT/RNTupleMerger.hxx
```diff
@@ -97,11 +98,12 @@ class RNTupleMerger final {
...
 
    /// Merge a given set of sources into the destination.
-   RResult<void> Merge(std::span<RPageSource *> sources, RPageSink &destination,
```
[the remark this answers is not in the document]
silverweed: What would be the expected behavior if I create a Merger with destination `foo` and then call `Merge` with destination `bar`?

tree/ntuple/v7/inc/ROOT/RNTupleMerger.hxx
silverweed: Hm, but how would the user get back ownership if they need it when they are done with it? If we add a `RetrieveSink()` method (or whatever) we leave the Merger in an invalid state..

hahnjo: I don't think the user should ever get it back; as we discussed the merger will soon own the model, so I'm not sure the user could even do something useful with the sink except merging in more files

silverweed: Makes sense

tree/ntuple/v7/inc/ROOT/RNTupleMerger.hxx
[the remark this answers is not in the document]
pcanal: I appeared at first that this was adding complexity to the simple case but this was (apparently) just me misreading the changes in the test code.

tree/ntuple/v7/inc/ROOT/RNTupleMerger.hxx
pcanal: Using the `unique_ptr` solves the problem.

> because it's clear that the merger should have a strictly smaller lifetime than the Sink

Yes for a use case where the user still need access to the object (and thus unique_ptr is not an option), capturing the reference is 'silently' sharing the ownership/usage of the object and will lead to user error.  We are trying to move away from implicitness.  i.e. `should have a strictly smaller lifetime ` might be true but can not be enforced nor checked by the compiler (nor simple static analysis tools).

silverweed: @pcanal so we have an agreement on the unique_ptr right? Can this thread be resolved?

hahnjo (approved):
Looks good from my point of view

PR #17477, opened 2025-01-21
[CMake] Rewrite ROOT's FIND_PYTHON_MODULE logic

hageboeck:
ROOT's find_python_module function was clashing with llvm's function of the same name. Depending on where the function is called, one or the other could have been invoked.
Furthermore, the output variables of both functions were clashing as well, so llvm finding (or not) a package affects the behaviour of the corresponding function call in ROOT.

Therefore:
- Rename `find_python_module` --> `ROOT_FIND_PYTHON_MODULE` (in ROOT; llvm untouched)
- Rename the output variable from `PY_${module_name}_FOUND` to
  `ROOT_${module_name}_FOUND` to avoid clashes with llvm.
- Create an entry in the CMakeCache that will survive a reconfigure. This entry decides whether tests that depend on Python packages are enabled or disabled, and is called `ROOT_TEST_${module_name}`. Setting it to ON will enable the tests, irrespective of whether the packages are installed. If the user doesn't choose, ROOT will find out as explained above when the first configure runs.


cmake/modules/RootMacros.cmake
```diff
@@ -1987,38 +1986,38 @@ macro(ROOT_ADD_COMPILE_OPTIONS flags)
...
+
+  if(NOT DEFINED ${FOUND_VAR})
+    execute_process(COMMAND "${Python3_EXECUTABLE}" "-c" "import ${module}"
+      RESULT_VARIABLE status
+      ERROR_QUIET)
+
+    if(${status} EQUAL 0)
+      set(${FOUND_VAR} ON CACHE BOOL "Enable tests depending on '${module}'")
+    else()
+      set(${FOUND_VAR} OFF CACHE BOOL "Enable tests depending on '${module}'")
```
vepadulano, line 2006: I guess there's a typo?
```suggestion
      set(${FOUND_VAR} OFF CACHE BOOL "Disable tests depending on '${module}'")
```

hageboeck: It's not a typo. This comment is for your `CMakeCache`, where it will show as:
```
// Enable tests depending on numba
PY_NUMBA_FOUND:BOOL=OFF
```
So I think it makes sense to keep the comment the same, so users can configure them as ON/OFF if they would e.g. decide to install the python package and enable the tests.

hageboeck: But looking at it now, it should probably be called `ROOT_PY_NUMBA_ENABLED:BOOL=OFF`, what do you think?

vepadulano (approved):
I like this new approach and the changes look good to me, thanks! Let's also wait for @bellenot's opinion

bellenot (approved):
LGTM, thanks!

PR #17480, opened 2025-01-22
[hist] Make abstract TH1,2,3 methods truly abstract and add missing implementations

guitargeek:
The TH1K class implementation forgot to implement some abstract methods
from the TH1 class that need to be implemented.

As the TH1K is almost identical to TH1F (besides the KDE algorithm in
the bin lookup), the TH1F implementations are copy-pasted.

The incomplete implementation was noticed thanks to the test in #17399.

To make sure that these mistakes can't happen anymore, the virtual methods of `TH1` are made truly virtual in the sense of the C++ language standard. So far, the implementation was only telling you at runtime that the method should be overridden.

### DIFF
--- a/hist/hist/inc/TH1.h
+++ b/hist/hist/inc/TH1.h
@@ -200,8 +200,12 @@ public:
    virtual Bool_t   Add(TF1 *h1, Double_t c1=1, Option_t *option="");
    virtual Bool_t   Add(const TH1 *h1, Double_t c1=1);
    virtual Bool_t   Add(const TH1 *h, const TH1 *h2, Double_t c1=1, Double_t c2=1);
-   virtual void     AddBinContent(Int_t bin);
-   virtual void     AddBinContent(Int_t bin, Double_t w);
+   /// Increment bin content by 1.
+   /// Passing an out-of-range bin leads to undefined behavior
+   virtual void     AddBinContent(Int_t bin) = 0;
+   /// Increment bin content by a weight w.
+   /// Passing an out-of-range bin leads to undefined behavior
+   virtual void     AddBinContent(Int_t bin, Double_t w) = 0;
    static  void     AddDirectory(Bool_t add=kTRUE);
    static  Bool_t   AddDirectoryStatus();
            void     Browse(TBrowser *b) override;
@@ -456,8 +460,15 @@ public:
    ClassDefOverride(TH1,8)  //1-Dim histogram base class
 
 protected:
-   virtual Double_t RetrieveBinContent(Int_t bin) const;
-   virtual void     UpdateBinContent(Int_t bin, Double_t content);
+
+   /// Raw retrieval of bin content on internal data structure
+   /// see convention for numbering bins in TH1::GetBin
+   virtual Double_t RetrieveBinContent(Int_t bin) const = 0;
+
+   /// Raw update of bin content on internal data structure
+   /// see convention for numbering bins in TH1::GetBin
+   virtual void     UpdateBinContent(Int_t bin, Double_t content) = 0;
+
    virtual Double_t GetBinErrorSqUnchecked(Int_t bin) const { return fSumw2.fN ? fSumw2.fArray[bin] : RetrieveBinContent(bin); }
 };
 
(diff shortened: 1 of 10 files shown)


ferdymercury (approved):
Looks awesome, thanks.

Not sure if this change might break user code in some situations but thery were already getting a warning before so I'm up for it.

hist/hist/inc/TH2Poly.h
```diff
@@ -130,6 +130,9 @@ class TH2Poly : public TH2 {
 
    //functions not to be used for TH2Poly
 
+   void AddBinContent(Int_t) override { return; }           ///< NOT IMPLEMENTED for TH2Poly
+   void AddBinContent(Int_t, Double_t) override { return; } ///< NOT IMPLEMENTED for TH2Poly
```
vepadulano, line 134: I would consider adding a `static_assert` to inform the user

vepadulano (approved):
Looks like a clear improvement to me, thanks! One small comment for consideration

pcanal:
> Not sure if this change might break user code in some situations but thery were already getting a warning before so I'm up for it.

It might break user code and it might not have been noticed before.   The case is a class derived from TH1 where they both only overloaded a subset of the function **and** only use the overloaded functions.

hist/hist/inc/TH2Poly.h
pcanal: The `static_assert` would trigger whether the function are used or not.  Consider using a call `Fatal`.

guitargeek: I have now used `Error`, because this is already done for many of the methods at the end of `TH2Poly.cxx`.

guitargeek:
> Is there a plan to replace "AbstracMethod strings" in other Root classes into pure virtuals? Such as in Tunixsystem etc?

@ferdymercury, no. I just did it it this PR because I was doing it locally anyway to hunt for the cases where these methods were not overridden. There are maybe other cases like that though! But I guess sometimes it's actually better than a "dummy override" in case you actually don't want to implement the method.

> It might break user code and it might not have been noticed before. The case is a class derived from TH1 where they both only overloaded a subset of the function and only use the overloaded functions.

@pcanal, I can't see why anyone would like to do that, but of course you are right to point out this corner case. But does that mean you are against the suggested changes? It's not clear to me if you only made a note or requesting changes. Personally I think this corner case is so obscure that we don't need to consider it.

pcanal:
@guitargeek My comments was a general one and making a point that we can not "blindly" replace all the `AbstractMethod` with `C++ abstract method` and that some would require thoughts and/or communication to user.

In this case, I agree with the change as `AddBinContent` is fundamental to the functionality of the class (example of functions that are 'not' fundamental are things like `GetTitle`, `Print`, etc.)

hist/hist/inc/TH1.h
```diff
@@ -200,8 +200,12 @@ class TH1 : public TNamed, public TAttLine, public TAttFill, public TAttMarker {
    virtual Bool_t   Add(TF1 *h1, Double_t c1=1, Option_t *option="");
    virtual Bool_t   Add(const TH1 *h1, Double_t c1=1);
    virtual Bool_t   Add(const TH1 *h, const TH1 *h2, Double_t c1=1, Double_t c2=1);
-   virtual void     AddBinContent(Int_t bin);
-   virtual void     AddBinContent(Int_t bin, Double_t w);
+   /// Increment bin content by 1.
+   /// Passing an out-of-range bin leads to undefined behavior
+   virtual void     AddBinContent(Int_t bin) = 0;
+   /// Increment bin content by a weight w.
+   /// Passing an out-of-range bin leads to undefined behavior
```
pcanal, line 207: That is too strong of a statement.  For many derived classes, the behavior is specified (will increase the overflow/underflow bins).

pcanal: Nevermind, this is the function taking a bin number not a value.

guitargeek:
@pcanal, thanks for clarifying!

pcanal (approved):

ferdymercury:
@guitargeek Is it possible that this change produced a regression?

```
auto h2 = new TH2I("h2", "h2", 10, 0, 1, 2, 0, 2);
h2->AddBinContent(1,1,1);
```

works on 6.34.02 but fails on master. (We should add a test, too).

guitargeek:
Yes possible! I forgot some "using" statements.

PR #17484, opened 2025-01-22
[v634] Backport of recent fixes and improvements

guitargeek:
Backport of recent fixes and improvements.


hahnjo:
"Improvements" should not be backported to a patch branch... @dpiparo

guitargeek:
Has been like this for years by the way: new ROOT releases introduced new RooFit features for users to try out, but they didn't cover all possible usecases because it's a-priori not clear how relevant every code branch in RooFit is. Support for these cases is then added via patch releases following user request.

This time, there was a request to support non-uniform binnings in codegen by both ATLAS and CMS analyzers, and it would really help them if this is available to try in some kind of release ASAP.

I understand your concern, because there are components in ROOT like IO that have way stricter stability requirements. But the analysis world that e.g. RooFit caters to works differently. We risk losing our users if we apply a rigid backport policy: the one or two ROOT releases that we have each year are not frequent enough for evolving analysis software.

Probably you still disagree because policy is policy. But then we should have a serious discussion about other ways to mitigate the risks from building a monolithic software framework where the users of the different components have different expectations on how quickly changes make it to a release.

PR #17494, opened 2025-01-22
[tutorials] make eve and eve7 tutorials be on the same hierarchy level.

martamaja10:
During the Hackathon eve was made part of eve7 and it was not correct. Now both are on the same level within the tutorials/visualisation directory.


tutorials/visualisation/eve7/box.C
```diff
@@ -1,5 +1,5 @@
 /// \file
-/// \ingroup tutorial_eve
+/// \ingroup tutorial_eve_7
```
linev, line 2: Why some eve7 tutorials have group id `tutorial_eve_7`, other  `tutorial_eve_77`

linev (changes requested):

tutorials/visualisation/eve7/box.C
linev: Can you make all `eve7` tutorials with `tutorial_eve_7` group?

linev: I see - you already fix it.

martamaja10: I think this is the case now

hahnjo:
Note that the tutorials are still not run because `visualisation/eve/*.C` is unconditionally part of `guiveto`...

PR #17500, opened 2025-01-23
[ntuple] Make RNTupleDescriptor::Clone() return it by value

silverweed:
It doesn't make much sense to return the descriptor by unique_ptr, because it's not polymorphic and it's not copyable, so value+move semantics do a better job (and they are consistent with the other descriptors).

Also marked all descriptor classes final because they're not meant to be overridden by users.

## Checklist:

- [x] tested changes locally


jblomer (approved):
LGTM (but see compiler errors)

PR #17503, opened 2025-01-23
[RDF] Add option to not clone histograms in MT-RDF

hageboeck:
- Add a wrapper for objects which are not thread safe
- Write an RDF action that uses TH3D without cloning them

TODOs:
- [ ] Document the feature
- [ ] Converge on how to enable the interface
  - [ ] Extend to 1D, 2D, ND?
  - [ ] Provide clone = Auto option?


hageboeck:
Closing in favour of #17743

PR #17504, opened 2025-01-23
[tutorials] Fix the path of tree500_cernbuild.C

bellenot:


martamaja10 (approved):
Thanks for cleaning these up :)

PR #12114, opened 2023-01-25
[RF] Increase version number of `RooRealVar`

guitargeek:
In ecd98632, the class version of the `RooAbsArg` class was incremented.

As explained in #8791, there are sometimes warnings in the IO of derived classes if their version number is not increased as well.

Increasing the class version of RooRealVar indeed fixes this warning that one gets right now when reading old workspaces:

```
Warning in <TStreamerInfo::BuildCheck>:
   The StreamerInfo of class RooRealVar read from file toyws/WS-VHbb-STXS_mu_toy_new.root
   has the same version (=8) as the active class but a different checksum.
   You should update the version to ClassDef(RooRealVar,9).
   Do not try to write objects with the current class definition,
   the files will not be readable.

Warning in <TStreamerInfo::CompareContent>: The following data member of
the on-file layout version 8 of class 'RooRealVar' differs from
the in-memory layout version 8:
   RooAbsBinning _binning; //
vs
   unique_ptr<RooAbsBinning,default_delete<RooAbsBinning> > _binning;
```

PR #17506, opened 2025-01-24
[RF] Increase version number of RooCrystalBall

### LINKED ISSUE #17163: Warnings from  the deserialization of RooCrystalBall
### Check duplicate issues.

- [X] Checked for duplicates

### Description

From my colleague I have a ROOT file with RooWorkspace that contains the compound pdf  that uses `RooCrystalBall` for some components. 

I am trying to read this file using ROOT 6.23.02 and I get following warnings
```
TStreamerInfo::BuildCheck:0: RuntimeWarning: 
   The StreamerInfo of class RooCrystalBall read from file  <...> 
   has the same version (=1) as the active class but a different checksum.
   You should update the version to ClassDef(RooCrystalBall,2).
   Do not try to write objects with the current class definition,
   the files will not be readable.

TStreamerInfo::CompareContent:0: RuntimeWarning: The following data member of
the on-file layout version 1 of class 'RooCrystalBall' differs from 
the in-memory layout version 1:
   RooTemplateProxy<RooAbsReal> alphaR_; //
vs
   unique_ptr<RooTemplateProxy<RooAbsReal>,default_delete<RooTemplateProxy<RooAbsReal> > > alphaR_; //
TStreamerInfo::CompareContent:0: RuntimeWarning: The following data member of
the on-file layout version 1 of class 'RooCrystalBall' differs from 
the in-memory layout version 1:
   RooTemplateProxy<RooAbsReal> nR_; //
vs
   unique_ptr<RooTemplateProxy<RooAbsReal>,default_delete<RooTemplateProxy<RooAbsReal> > > nR_; //

``` 
Presumably the file has been create with some earlier version of ROOT.

  I have some suspicion that the warning is not harmless: 
  while the resulting PDF looks OK when I use it for the fit&draw,
  when I get the CDF from it,  for some entries in the range CDF (slightly) exceeds 1.
  Could it be a sequence of the problems wtih deserialization ?

   

### Reproducer

I am using  ROOT from LCG_106 build cvmfs 
(I get the same problem, when using LCG dev3 nightly build) 

### ROOT version

I am using  ROOT from LCG_106 build cvmfs 
(I get the same problem, when using LCG dev3 nightly build) 

### Installation method

LCG cvmfs 

### Operating system

lxplus 

### Additional context

_No response_

guitargeek:
Same reasoning as in commit bce142c from #12114.

Closes #17163.

### DIFF
--- a/roofit/roofit/inc/RooCrystalBall.h
+++ b/roofit/roofit/inc/RooCrystalBall.h
@@ -60,7 +60,7 @@ private:
    std::unique_ptr<RooRealProxy> alphaR_ = nullptr;
    std::unique_ptr<RooRealProxy> nR_ = nullptr;
 
-   ClassDefOverride(RooCrystalBall, 1)
+   ClassDefOverride(RooCrystalBall, 2)
 };
 
 #endif


pcanal (approved):
Indeed, when a base class version change, one need to also change the derived classes number :(.  Thanks for the update.

PR #17508, opened 2025-01-24
[RF] Consistently use universal `RooAbsData::reduce()` overload

guitargeek:
Consistently use `RooAbsData::reduce()` overload that takes RooFit command arguments in the RooFit code.

This allows us to deprecate all the other overloads one day, in the spirit of avoiding duplicate interfaces and user confusion.

Also, avoid some code repetition in setting up the cut expressions for slicing in RooAbsReal.cxx and RooSimultaneous.cxx.


lmoneta (approved):
LGTM!

PR #17518, opened 2025-01-24
[RF] Avoid unneeded cached projection integrals in RooAddPdf

### LINKED ISSUE #12645: [RF] Unneeded normalization integrals when fitting a RooAddPdf in a range
- [x] Checked for duplicates

### Describe the bug

In a `RooAddPdf`, the coefficients are tied to the normalization range. When passing a range name to the `RooAbsPdf::fitTo` call, this is visible because new integrals will be created over the fit range, to fix the definition of the coefficients. I will refer to these integrals as coefnorm integrals.
In ROOT 6.28, coefnorm integrals for ranged fits of `RooAddPdf`s were changed. This, however, added several integrals that with the old method would not be evaluated. In particular:
- the new implementation is unable to elide coefnorm integrals over ranges equivalents to the full range
- in the new implementation, to every coefnorm integral over the fit range correspond an integral over an empty named range. This is separate to the integral with no range, which is also done
- calling `addPdf.fixAddCoefRange(fit_range)` before the fit (a common solution if you wanted to avoid coefnorm integrals) removes the unnamed range coefnorm integrals, but not the named range ones

Note, however, that in all the test I performed these integral do not seem to affect fit results, which are the same before and after 6.28.
Instead, the problem is that those integrals may not be easy to compute: this issue was noticed because among the extra integrals was a complex numeric integral that had to be evaluated at each event in the dataset, was extremely slow, and in practice caused the fit to never converge.

### Expected behavior

No extra integrals over unnamed ranges, passing the full range to fitTo does not add integrals, calling `fixAddCoefRange` removes all coefnorm integrals.

### To Reproduce

```Python
import ROOT
from ROOT.RooFit import RooConst

# in this simple example the symptoms are only in the logs, in the names of the integrals evaluated
ROOT.RooMsgService.instance().addStream(ROOT.RooFit.INFO, Topic = ROOT.RooFit.Integration)

x = ROOT.RooRealVar('x', '', 0, 1)

g = ROOT.RooGaussian('g', '', x, RooConst(0.5), RooConst(0.2))
u = ROOT.RooUniform('u', '', x)

f = ROOT.RooRealVar('f', '', 0.5, 0, 1)
a = ROOT.RooAddPdf('a', '', [g, u], [f])

dt = a.generate({x}, NumEvents = 1000)

x.setRange('limited', 0.2, 0.8)
x.setRange('full', 0, 1)

a.fitTo(dt, Range = 'full', PrintLevel = -1)

a.fitTo(dt, Range = 'limited', PrintLevel = -1)

a.fixAddCoefRange('full')
a.fitTo(dt, Range = 'full', PrintLevel = -1)

a.fixAddCoefRange('limited')
a.fitTo(dt, Range = 'limited', PrintLevel = -1)

```
Compare 6.26 with master.

### Setup

ROOT 6.26 from LCG 102b
ROOT master from LCG dev3

guitargeek:
By already normalizing the pdf components integrals to the right fit range, one can elide some potentially expensive numeric correction factor integrals.

The code that calculates the predicted number of events was updated accordingly, since it containts also some integrals that depend on the components normalization range.

Closes #12645.

A unit test that validates that there are no superfluous integrals in the case of the GitHub reproducer was also added.

### DIFF
--- a/roofit/roofitcore/inc/RooFit/Detail/RooNormalizedPdf.h
+++ b/roofit/roofitcore/inc/RooFit/Detail/RooNormalizedPdf.h
@@ -32,6 +32,7 @@ public:
       auto name = std::string(pdf.GetName()) + "_over_" + _normIntegral->GetName();
       SetName(name.c_str());
       SetTitle(name.c_str());
+      _normRange = pdf.normRange(); // so that e.g. RooAddPdf can query over what we are normalized
    }
 
    RooNormalizedPdf(const RooNormalizedPdf &other, const char *name)
--- a/roofit/roofitcore/src/RooAddHelpers.cxx
+++ b/roofit/roofitcore/src/RooAddHelpers.cxx
@@ -56,14 +56,17 @@ AddCacheElem::AddCacheElem(RooAbsPdf const &addPdf, RooArgList const &pdfList, R
       nset2.add(refCoefNormSet);
    }
 
-   bool hasPdfWithCustomRange = false;
+   bool hasPdfWithDifferentRange = false;
 
    // Fill with dummy unit RRVs for now
    for (std::size_t i = 0; i < pdfList.size(); ++i) {
       auto pdf = static_cast<const RooAbsPdf *>(pdfList.at(i));
       auto coef = static_cast<const RooAbsReal *>(coefList.at(i));
 
-      hasPdfWithCustomRange |= pdf->normRange() != nullptr;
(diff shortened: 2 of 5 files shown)


dpiparo:
My apologies for stopping the builds. Before this is merged, https://github.com/root-project/root/pull/17519, everything would be red.

dpiparo (approved):

Issue #10774, opened 2022-06-16
[TTree] Inconcistency with the TTreeReader's SetEntries.


ikabadzhov:
I encountered the following inconcistent behaviour in the `TTreeReader`
```cpp
   auto dfWriter0 = RDataFrame(3).Define("x", [] { return 1; });
   dfWriter0.Snapshot<int>("treeA", "file0.root", {"x"});
   auto chain = std::make_shared<TChain>("treeA");
   chain->Add("file0.root");
   TTreeReader r(chain.get(), chain->GetEntryList());
   r.SetEntriesRange(4, 5);
```
this produces: `Error in <TTreeReader::SetEntriesRange()>: Error setting first entry 4: one of the readers was not successfully initialized`, which is good (as expected).

On the other hand, if the range becomes `r.SetEntriesRange(3, 5)` - start is now the last index + 1, there is no error produced or etc.

Using gdb, I saw that there is a call in `SetEntriesRange` to `SetEntry(beginEntry - 1)` (https://github.com/root-project/root/blob/f841b96e634b7f7136b0c80da0f9394e8e370314/tree/treeplayer/src/TTreeReader.cxx#L438)

`SetEntry` however, calls `SetEntryBase`, which modifies `fEntry`. Hence, the quickfix to call `SetEntry(beginEntry)` is obviously dangerous.

CC: @eguiraud @vepadulano

eguiraud:
I think the underlying problem is that `SetEntriesRange` has no input validation. We could (and should) error out at the point `SetEntriesRange` is called if `start > nEntries`.

ikabadzhov:
I can take care of making a check before calling SetEntriesRange. I was more concerned whether SetEntriesRange can become dangerous on its own.

eguiraud:
> SetEntriesRange can become dangerous on its own

what do you mean?

ikabadzhov:
> > SetEntriesRange can become dangerous on its own
> 
> what do you mean?

User not getting notified for specifying invalid range (e.g. start=last entry)

eguiraud:
Yes that's what input validation is for, what am I missing?

ikabadzhov:
Fyi I am currently doing my own sanity check (outside `SetEntriesRange`) by first calling `SetEntry` with the desired beginning. 

`SetEntry` returns an entry status, which is 0 upon success, and works in my case as expected, i.e. it properly catches that `SetEntry(lastEntry)` is invalid.

eguiraud:
[...]

PR #17520, opened 2025-01-24
Warn if specified entry range is larger than number of processed entries

### LINKED ISSUE #13800: [DF] RDatasetSpec allows a global range larger than the actual dataset

This is really up for discussion, but I wanted to at least raise the issue.

One can put an arbitrarily large number as the `endEntry` of the global entry range passed to `RDatasetSpec::WithGlobalRange`, and everything will run correctly: RDataFrame simply stops the event loop at the last actually available entry.

Diagnosing this case to warn the user or to error out is not straightforward in general: think of a dataset made of O(1k) files read over the network: hard to tell what the max number of entries is before we start processing. I think this is the reason for the current behavior.

Reproducer:

```cpp
#include "ROOT/RDF/RDatasetSpec.hxx"
#include "ROOT/RDF/RSample.hxx"

int main() {
  {
    TFile f("f.root", "recreate");
    TTree t("t", "t");
    t.Fill();
    t.Write();
  }
  
  ROOT::RDF::Experimental::RDatasetSpec spec;
  ROOT::RDF::Experimental::RSample sample("main", {{"t", "f.root"}});
  spec.AddSample(sample);
  spec.WithGlobalRange({0, 2}); // but the tree only has 1 entry
  ROOT::RDataFrame(spec).Count().GetValue(); // no error
}
```

### ROOT version

any that supports RDatasetSpec

### Installation method

any

### Operating system

any

### Additional context

_No response_

vepadulano:
This PR fixes #13800


tree/dataframe/src/RLoopManager.cxx
```diff
@@ -590,6 +590,16 @@ void RLoopManager::RunTreeProcessorMT()
...
       }
    });
+
+   auto &&processedEntries = entryCount.load();
+   if (fEndEntry != std::numeric_limits<Long64_t>::max() &&
+       static_cast<ULong64_t>(fEndEntry - fBeginEntry) > processedEntries) {
+      Warning(
+         "RDataFrame::Run",
+         "The specified global entry range (begin=%lld,end=%lld) would result in reading more entries than the actual "
+         "number of processed entries (%lld). Consider adjusting the end value of the entry range to a maximum of %lld",
```
martamaja10, line 600:
```suggestion
         "number of processed entries (%lld). Consider adjusting the end value of the entry range to the maximum entry number which is %lld",
```

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -629,6 +639,15 @@ void RLoopManager::RunTreeReader()
       throw std::runtime_error("An error was encountered while processing the data. TTreeReader status code is: " +
                                std::to_string(r.GetEntryStatus()));
    }
+
+   auto &&processedEntries = r.GetCurrentEntry();
+   if (fEndEntry != std::numeric_limits<Long64_t>::max() && (fEndEntry - fBeginEntry) > processedEntries) {
+      Warning(
+         "RDataFrame::Run",
+         "The specified global entry range (begin=%lld,end=%lld) would result in reading more entries than the actual "
+         "number of processed entries (%lld). Consider adjusting the end value of the entry range to a maximum of %lld",
```
martamaja10, line 648:
```suggestion
         "number of processed entries (%lld). Consider adjusting the end value of the entry range to the maximum entry number which is %lld",
```

tree/dataframe/src/RLoopManager.cxx
martamaja10: maybe the phrasing here is not perfect either but I would somehow try to make the noun `maximum` a bit more precise

vepadulano: Yes I agree we can improve the wording. I wanted to actually give further meaning to this sentence, i.e. that the user can specify any value for the end entry  between 0 and the maximum, that was my intent. We can think about how to improve the sentence with this sense

hageboeck: Currently, the logic doesn't include the case of end==0, or am I misreading this?

My suggestion for the sentence would be:
`The requested entry range (begin=%lld,end=%lld) is larger than the input dataset (%lld). Consider ..."

martamaja10: Regarding the consider part of this sentence maybe: 
"Consider adjusting the end value of the entry range so that it does not exceed the maximum entry number which is %lld"

vepadulano: >  is larger than the input dataset (%lld).

This is wrong unfortunately. In general we could have situations where the number of processed entries is less than the total number of entries in the input dataset, e.g. when a `Range` operation is used.

vepadulano: > the logic doesn't include the case of end==0, or am I misreading this?

You are partly right. To be precise, the user should specify any number between the value of `fBeginEntry` and the maximum number of entries that are processed. When `fBeginEntry == fEndEntry`, the event loop will just not run.

hageboeck: In a chat we came up with something like:
"The event loop finished at entry xxx, whereas a range of xxx -> yyy was selected." Something along these lines might be easier to understand for the users.

vepadulano:
This PR now also fixes https://github.com/root-project/root/issues/10774

hageboeck (approved):
LGTM!

martamaja10 (approved):
Thanks for considering all the options. The warnings are now also very clear.

PR #17523, opened 2025-01-25
io rules: Prevent dimensions from spilling.

pcanal:
This solves part 1 at https://github.com/cms-sw/cmsdist/pull/9627#issuecomment-2609571193.

without this path all inputs following an array input were wrongly marked as an array.


dpiparo (approved):
Thanks! would it be hard to add a test?

pcanal:
Not at all.  It is actually already written and test.  See https://github.com/root-project/roottest/pull/1252

dpiparo:
thanks a lot, I missed that.

jblomer (approved):

PR #17534, opened 2025-01-27
[ntuple] Fix handling of I/O rules' inputs

jblomer:
Fix execution of I/O customization rules that have on-disk data member sources (inputs). Previously, `TVirtualObject::fObject` pointed to the target object, which is plain wrong (but happens to work in some cases). Now, rule inputs are examined, fields are created according to the provided input's type, and the data is read in a staging area that follows the memory layout of the on-disk class given by TClass. This staging area is now what `TVirtualObject::fObject` points to. 

Note that the target still has to be a transient member. This restriction is going to be lifted in a second PR.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RField.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RField.hxx
@@ -157,7 +157,7 @@ protected:
    void ReadGlobalImpl(ROOT::NTupleSize_t globalIndex, void *to) final;
    void ReadInClusterImpl(RNTupleLocalIndex localIndex, void *to) final;
    void BeforeConnectPageSource(Internal::RPageSource &pageSource) final;
-   void OnConnectPageSource() final;
+   void AfterConnectPageSource() final;
 
 public:
    RClassField(std::string_view fieldName, std::string_view className);
@@ -448,7 +448,7 @@ protected:
    std::size_t AppendImpl(const void *from) final;
    void ReadGlobalImpl(ROOT::NTupleSize_t globalIndex, void *to) final;
 
-   void OnConnectPageSource() final;
+   void AfterConnectPageSource() final;
 
 public:
    static std::string TypeName() { return "TObject"; }
--- a/tree/ntuple/v7/inc/ROOT/RField.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RField.hxx
@@ -143,9 +143,13 @@ private:
    RClassField(std::string_view fieldName, const RClassField &source); ///< Used by CloneImpl
    RClassField(std::string_view fieldName, TClass *classp);
    void Attach(std::unique_ptr<RFieldBase> child, RSubFieldInfo info);
-   /// Register post-read callbacks corresponding to a list of ROOT I/O customization rules. `classp` is used to
-   /// fill the `TVirtualObject` instance passed to the user function.
-   void AddReadCallbacksFromIORules(const std::span<const TSchemaRule *> rules, TClass *classp = nullptr);
+   /// Register post-read callback corresponding to a ROOT I/O customization rules.
+   void AddReadCallbacksFromIORule(const TSchemaRule *rule);
+   /// Given the on-disk information from the page source, find all the I/O customization rules that apply
+   /// to the class field at hand, to which the fieldDesc descriptor, if provided, must correspond.
+   /// Fields may not have an on-disk representation (e.g., when inserted by schema evolution), in which case the passed
+   /// field descriptor is nullptr.
+   std::vector<const TSchemaRule *> FindRules(const RFieldDescriptor *fieldDesc);
 
 protected:
    std::unique_ptr<RFieldBase> CloneImpl(std::string_view newName) const final;
@@ -157,7 +161,6 @@ protected:
    void ReadGlobalImpl(ROOT::NTupleSize_t globalIndex, void *to) final;
    void ReadInClusterImpl(RNTupleLocalIndex localIndex, void *to) final;
    void BeforeConnectPageSource(Internal::RPageSource &pageSource) final;
-   void AfterConnectPageSource() final;
 
 public:
    RClassField(std::string_view fieldName, std::string_view className);
--- a/tree/ntuple/v7/inc/ROOT/RField.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RField.hxx
@@ -122,6 +122,13 @@ private:
       ESubFieldRole fRole;
       std::size_t fOffset;
    };
+   // Information to read into the staging area a field that is used as an input to an I/O customization rule
+   struct RStagingItem {
+      /// The field used to read the on-disk data. The fields type may be different from the on-disk type as long
+      /// as the on-disk type can be converted to the fields type (through type cast / schema evolution).
+      std::unique_ptr<RFieldBase> fField;
+      std::size_t fOffset; ///< offset in fStagingArea
+   };
    /// Prefix used in the subfield names generated for base classes
    static constexpr const char *kPrefixInherited{":"};
 
@@ -139,10 +146,24 @@ private:
    std::vector<RSubFieldInfo> fSubFieldsInfo;
    std::size_t fMaxAlignment = 1;
 
+   /// The staging area stores inputs to I/O rules according to the offsets given by the streamer info of
+   /// "TypeName@@Version". The area is allocated depending on I/O rules resp. the source members of the I/O rules.
+   std::unique_ptr<unsigned char[]> fStagingArea;
+   std::unordered_map<std::string, RStagingItem> fStagingItems; ///< Lookup staging items by member name
+
 private:
    RClassField(std::string_view fieldName, const RClassField &source); ///< Used by CloneImpl
    RClassField(std::string_view fieldName, TClass *classp);
    void Attach(std::unique_ptr<RFieldBase> child, RSubFieldInfo info);
+
+   /// Returns the id of member 'name' in the class field given by 'fieldId', or kInvalidDescriptorId if no such
+   /// member exist. Looks recursively in base classes.
+   ROOT::DescriptorId_t
+   LookupMember(const RNTupleDescriptor &desc, std::string_view memberName, ROOT::DescriptorId_t classFieldId);
+   /// If there are rules with inputs (source members), create the staging area according to the TClass instance
+   /// that corresponds to the on-disk field.
+   void PrepareStagingArea(const std::vector<const TSchemaRule *> &rules, const RNTupleDescriptor &desc,
+                           const RFieldDescriptor &classFieldId);
    /// Register post-read callback corresponding to a ROOT I/O customization rules.
    void AddReadCallbacksFromIORule(const TSchemaRule *rule);
    /// Given the on-disk information from the page source, find all the I/O customization rules that apply
--- a/tree/ntuple/v7/inc/ROOT/RField.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RField.hxx
@@ -149,6 +149,11 @@ private:
    /// The staging area stores inputs to I/O rules according to the offsets given by the streamer info of
    /// "TypeName@@Version". The area is allocated depending on I/O rules resp. the source members of the I/O rules.
    std::unique_ptr<unsigned char[]> fStagingArea;
+   /// The TClass instance that corresponds to the staging area.
+   /// The staging class exists as <class name>@@<on-disk version> if the on-disk version is different from the
+   /// current in-memory version, or it can be accessed by the first @@alloc streamer element of the current streamer
+   /// info.
+   TClass *fStagingClass = nullptr;
    std::unordered_map<std::string, RStagingItem> fStagingItems; ///< Lookup staging items by member name
 
 private:
@@ -160,6 +165,8 @@ private:
    /// member exist. Looks recursively in base classes.
    ROOT::DescriptorId_t
    LookupMember(const RNTupleDescriptor &desc, std::string_view memberName, ROOT::DescriptorId_t classFieldId);
+   /// Sets fStagingClass according to the given name and version
+   void SetStagingClass(const std::string &className, unsigned int classVersion);
    /// If there are rules with inputs (source members), create the staging area according to the TClass instance
    /// that corresponds to the on-disk field.
    void PrepareStagingArea(const std::vector<const TSchemaRule *> &rules, const RNTupleDescriptor &desc,
--- a/tree/ntuple/v7/src/RFieldMeta.cxx
+++ b/tree/ntuple/v7/src/RFieldMeta.cxx
@@ -208,7 +208,8 @@ std::vector<const ROOT::TSchemaRule *> ROOT::Experimental::RClassField::FindRule
    // Erase unknown rule types
    auto hasUnknownType = [](const ROOT::TSchemaRule *rule) {
       if (rule->GetRuleType() != ROOT::TSchemaRule::kReadRule) {
-         R__LOG_WARNING(ROOT::Internal::NTupleLog()) << "ignoring I/O customization rule with unsupported type";
+         R__LOG_WARNING(ROOT::Internal::NTupleLog())
+            << "ignoring I/O customization rule with unsupported type: " << rule->GetRuleType();
          return true;
       }
       return false;
--- a/tree/ntuple/v7/src/RFieldMeta.cxx
+++ b/tree/ntuple/v7/src/RFieldMeta.cxx
@@ -189,32 +189,45 @@ std::vector<const ROOT::TSchemaRule *> ROOT::Experimental::RClassField::FindRule
       }
    }
 
-   // For the time being, we only support rules targeting transient members
-   auto targetsNonTransientMembers = [classp = fClass](const ROOT::TSchemaRule *rule) {
-      if (rule->GetTarget() == nullptr)
-         return false;
-      for (auto target : ROOT::Detail::TRangeStaticCast<TObjString>(*rule->GetTarget())) {
-         const auto dataMember = classp->GetDataMember(target->GetString());
-         if (!dataMember || dataMember->IsPersistent()) {
-            R__LOG_WARNING(ROOT::Internal::NTupleLog())
-               << "ignoring I/O customization rule with non-transient member: " << dataMember->GetName();
-            return true;
-         }
-      }
-      return false;
-   };
-   rules.erase(std::remove_if(rules.begin(), rules.end(), targetsNonTransientMembers), rules.end());
+   // Cleanup and sort rules
+   std::size_t nskip = 0; // skip whole-object-rules that were moved to the end of the rules vector
+   for (auto itr = rules.begin(); itr != rules.end() - nskip;) {
+      const auto rule = *itr;
 
-   // Erase unknown rule types
-   auto hasUnknownType = [](const ROOT::TSchemaRule *rule) {
+      // Erase unknown rule types
       if (rule->GetRuleType() != ROOT::TSchemaRule::kReadRule) {
          R__LOG_WARNING(ROOT::Internal::NTupleLog())
             << "ignoring I/O customization rule with unsupported type: " << rule->GetRuleType();
-         return true;
+         itr = rules.erase(itr);
+         continue;
       }
-      return false;
-   };
-   rules.erase(std::remove_if(rules.begin(), rules.end(), hasUnknownType), rules.end());
+
+      // Rules targeting the entire object need to be executed at the end
+      if (rule->GetTarget() == nullptr) {
+         nskip++;
+         if (itr != rules.end() - nskip)
+            std::iter_swap(itr++, rules.end() - nskip);
+         continue;
+      }
+
+      // For the time being, we only support rules targeting transient members
+      bool hasPersistentTarget = false;
+      for (auto target : ROOT::Detail::TRangeStaticCast<TObjString>(*rule->GetTarget())) {
+         const auto dataMember = fClass->GetDataMember(target->GetString());
+         if (!dataMember || dataMember->IsPersistent()) {
+            R__LOG_WARNING(ROOT::Internal::NTupleLog())
+               << "ignoring I/O customization rule with non-transient member: " << dataMember->GetName();
+            hasPersistentTarget = true;
+            break;
+         }
+      }
+      if (hasPersistentTarget) {
+         itr = rules.erase(itr);
+         continue;
+      }
+
+      ++itr;
+   }
 
    return rules;
 }
--- a/tree/ntuple/v7/src/RFieldMeta.cxx
+++ b/tree/ntuple/v7/src/RFieldMeta.cxx
@@ -247,7 +247,7 @@ void ROOT::Experimental::RClassField::BeforeConnectPageSource(Internal::RPageSou
    }
 }
 
-void ROOT::Experimental::RClassField::OnConnectPageSource()
+void ROOT::Experimental::RClassField::AfterConnectPageSource()
 {
    // Add post-read callbacks for I/O customization rules; only rules that target transient members are allowed for now
    // TODO(jalopezg): revise after supporting schema evolution
@@ -844,7 +844,7 @@ void ROOT::Experimental::RField<TObject>::ReadGlobalImpl(ROOT::NTupleSize_t glob
    *reinterpret_cast<UInt_t *>(reinterpret_cast<unsigned char *>(to) + GetOffsetBits()) = bits;
 }
 
-void ROOT::Experimental::RField<TObject>::OnConnectPageSource()
+void ROOT::Experimental::RField<TObject>::AfterConnectPageSource()
 {
    if (GetTypeVersion() != 1) {
       throw RException(R__FAIL("unsupported on-disk version of TObject: " + std::to_string(GetTypeVersion())));
--- a/tree/ntuple/v7/src/RFieldMeta.cxx
+++ b/tree/ntuple/v7/src/RFieldMeta.cxx
@@ -235,6 +235,9 @@ std::size_t ROOT::Experimental::RClassField::AppendImpl(const void *from)
 
 void ROOT::Experimental::RClassField::ReadGlobalImpl(ROOT::NTupleSize_t globalIndex, void *to)
 {
+   for (const auto &[_, si] : fStagingItems) {
+      CallReadOn(*si.fField, globalIndex, fStagingArea.get() + si.fOffset);
+   }
    for (unsigned i = 0; i < fSubFields.size(); i++) {
       CallReadOn(*fSubFields[i], globalIndex, static_cast<unsigned char *>(to) + fSubFieldsInfo[i].fOffset);
    }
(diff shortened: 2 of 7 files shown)


tree/ntuple/v7/inc/ROOT/RField.hxx
```diff
@@ -123,13 +130,29 @@ private:
...
+   /// Returns the id of member 'name' in the class field given by 'fieldId', or kInvalidDescriptorId if no such
+   /// member exist. Looks recursively in base classes.
+   DescriptorId_t LookupMember(const RNTupleDescriptor &desc, std::string_view memberName, DescriptorId_t classFieldId);
+   /// If there are rules with inputs (source members), create the staging area according to the TClass instance
+   /// that corresponds to the on-disk field.
+   void PrepareStagingArea(const std::vector<const TSchemaRule *> &rules, const RNTupleDescriptor &desc,
+                           const RFieldDescriptor &classFieldId);
+   /// Register post-read callback corresponding to a ROOT I/O customization rules.
+   void AddReadCallbacksFromIORule(const TSchemaRule *rule);
+   /// Given the on-disk information from the page source, find all the I/O customization rules that apply.
```
pcanal, line 152:
```suggestion
   /// Given the on-disk information from the page source, find all the I/O customization rules that apply.
   /// Rules applies if the 'field' is considered a 'source' of the rule.
```

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -149,24 +149,55 @@ void ROOT::Experimental::RClassField::Attach(std::unique_ptr<RFieldBase> child,
    RFieldBase::Attach(std::move(child));
 }
 
-void ROOT::Experimental::RClassField::AddReadCallbacksFromIORules(const std::span<const ROOT::TSchemaRule *> rules,
-                                                                  TClass *classp)
+std::vector<const ROOT::TSchemaRule *> ROOT::Experimental::RClassField::FindRules(const RFieldDescriptor *fieldDesc)
```
pcanal, line 152:
```suggestion
std::vector<const ROOT::TSchemaRule *> ROOT::Experimental::RClassField::FindRules(const RFieldDescriptor *sourceFieldDesc)
```

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -149,24 +149,55 @@ void ROOT::Experimental::RClassField::Attach(std::unique_ptr<RFieldBase> child,
...
+         rules = ruleset->FindRules(fieldDesc->GetTypeName(), fieldDesc->GetTypeVersion(),
+                                    *fieldDesc->GetTypeChecksum());
+      } else {
+         rules = ruleset->FindRules(fieldDesc->GetTypeName(), fieldDesc->GetTypeVersion());
+      }
+   }
+
+   // For the time being, we only support rules targeting transient members
+   auto referencesNonTransientMembers = [classp = fClass](const ROOT::TSchemaRule *rule) {
+      if (rule->GetTarget() == nullptr)
```
pcanal, line 176: You might probably be able to also apply those rules as they implicitly applies to the whole object and thus need to be 'scheduled' after all other I/O as been performed.

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -149,24 +149,55 @@ void ROOT::Experimental::RClassField::Attach(std::unique_ptr<RFieldBase> child,
...
+      if (fieldDesc->GetTypeChecksum()) {
+         rules = ruleset->FindRules(fieldDesc->GetTypeName(), fieldDesc->GetTypeVersion(),
+                                    *fieldDesc->GetTypeChecksum());
+      } else {
+         rules = ruleset->FindRules(fieldDesc->GetTypeName(), fieldDesc->GetTypeVersion());
+      }
+   }
+
+   // For the time being, we only support rules targeting transient members
+   auto referencesNonTransientMembers = [classp = fClass](const ROOT::TSchemaRule *rule) {
```
pcanal, line 175:
```suggestion
   auto targetsNonTransientMembers = [classp = fClass](const ROOT::TSchemaRule *rule) {
```
'references' could also means 'listed as one of the source'

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -149,24 +149,55 @@ void ROOT::Experimental::RClassField::Attach(std::unique_ptr<RFieldBase> child,
...
+         }
+      }
+      return false;
+   };
+   rules.erase(std::remove_if(rules.begin(), rules.end(), referencesNonTransientMembers), rules.end());
+
+   // Erase unknown rule types
+   auto hasUnknownType = [](const ROOT::TSchemaRule *rule) {
       if (rule->GetRuleType() != ROOT::TSchemaRule::kReadRule) {
          R__LOG_WARNING(NTupleLog()) << "ignoring I/O customization rule with unsupported type";
```
pcanal, line 193:
```suggestion
         R__LOG_WARNING(NTupleLog()) << "ignoring I/O customization rule with unsupported type: " << rule->GetRuleType();
```

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -186,39 +217,134 @@ std::size_t ROOT::Experimental::RClassField::AppendImpl(const void *from)
...
+            throw RException(R__FAIL(std::string("cannot find on disk rule source member ") + GetTypeName() + "." +
+                                     source->GetName()));
+         }
+         const auto &memberFieldDesc = desc.GetFieldDescriptor(memberFieldId);
+
+         auto memberType = source->GetTypeForDeclaration() + source->GetDimensions();
+         stagingItem.fField = Create("" /* we don't need a field name */, std::string(memberType)).Unwrap();
+         stagingItem.fField->SetOnDiskId(memberFieldDesc.GetId());
+
+         stagingItem.fOffset = cl->GetDataMemberOffset(source->GetName());
```
pcanal, line 291: We should consider adding a check that `fOffset` is not `kMissing` (i.e. the source was not actually found in the class). [The code is not fully consistent yet and sometimes `0` will be returned when a member is not found :( )

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -186,39 +217,134 @@ std::size_t ROOT::Experimental::RClassField::AppendImpl(const void *from)
...
+void ROOT::Experimental::RClassField::PrepareStagingArea(const std::vector<const TSchemaRule *> &rules,
+                                                         const RNTupleDescriptor &desc,
+                                                         const RFieldDescriptor &classFieldDesc)
+{
+   auto clName = classFieldDesc.GetTypeName();
+   if (classFieldDesc.GetTypeVersion() != GetTypeVersion()) {
+      TClass::GetClass(classFieldDesc.GetTypeName().c_str())->GetStreamerInfo(classFieldDesc.GetTypeVersion());
+      clName += std::string("@@") + std::to_string(classFieldDesc.GetTypeVersion());
+   }
+   auto cl = TClass::GetClass(clName.c_str());
```
pcanal, line 267: Finding the `@@` class relies on the related `StreamerInfo` to have been build, is that a guarantee here?

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -186,39 +217,134 @@ std::size_t ROOT::Experimental::RClassField::AppendImpl(const void *from)
...
+
+   return kInvalidDescriptorId;
+}
+
+void ROOT::Experimental::RClassField::PrepareStagingArea(const std::vector<const TSchemaRule *> &rules,
+                                                         const RNTupleDescriptor &desc,
+                                                         const RFieldDescriptor &classFieldDesc)
+{
+   auto clName = classFieldDesc.GetTypeName();
+   if (classFieldDesc.GetTypeVersion() != GetTypeVersion()) {
```
pcanal, line 263: Technically this is not the right test. There can be staging even for the current version.

Inside TStreamerInfo the equivalent test are either (during building) whether there is a rule that applies for the pair (current version and onfile version) or whether first element of the TStreamerInfo being used start with an `@@alloc` element (i.e. an element with type 'TStreamerInfo::kCacheNew') [The element also give direct access to the staging/alloc TClass).


With the file sameversion.h, we get:
```
root [0] .L sameversion.h+
root [1] auto cl = TClass::GetClass("SameClassRule");
root [2] cl->GetStreamerInfo()->ls()

StreamerInfo for class: SameClassRule, version=3, checksum=0x1bf8119e
  SameClassRule@@3 @@alloc         offset=  0 type=1001                     
  int            fValue          offset=  8 type= 3  (cached,repeat)                     
  int            fValue          offset=  0 type= 3                     
  void           SameClassRule_rule0 offset=  0 type=1000  (wholeObject)                     
  SameClassRule@@3 @@dealloc       offset=  0 type=1002                     
   i= 0, @@alloc         type=1001, offset=  0, len=1, method=0
   i= 1, fValue          type=  3, offset=  8, len=1, method=0 [cached,repeat]
   i= 2, fValue          type=  3, offset=  0, len=1, method=0
   i= 3, SameClassRule_rule0 type=1000, offset=  0, len=1, method=0 [wholeObject]
   i= 4, @@dealloc       type=1002, offset=  0, len=1, method=0
```

```C++
// File sameversion.h
#pragma once

struct SameClassRule
{
   int fValue;
};


#ifdef __ROOTCLING__
#pragma link C++ options = version(3) class SameClassRule+;
#pragma read sourceClass="SameClassRule" targetClass="SameClassRule" versions="[1-]" \
   source="int fValue" target="" code="{ printf(\"inside rule\n\"); }"
#endif
```

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -186,39 +217,134 @@ std::size_t ROOT::Experimental::RClassField::AppendImpl(const void *from)
...
+         }
+         const auto &memberFieldDesc = desc.GetFieldDescriptor(memberFieldId);
+
+         auto memberType = source->GetTypeForDeclaration() + source->GetDimensions();
+         stagingItem.fField = Create("" /* we don't need a field name */, std::string(memberType)).Unwrap();
+         stagingItem.fField->SetOnDiskId(memberFieldDesc.GetId());
+
+         stagingItem.fOffset = cl->GetDataMemberOffset(source->GetName());
+         stagingAreaSize = std::max(stagingAreaSize, stagingItem.fOffset + stagingItem.fField->GetValueSize());
+         R__ASSERT(stagingAreaSize > 0);
```
pcanal, line 293: The `stagingAreaSize` is already calculated and we should have at the end `stagingAreaSize == cl->Size();`

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -186,39 +217,134 @@ std::size_t ROOT::Experimental::RClassField::AppendImpl(const void *from)
...
+   }
+}
+
+void ROOT::Experimental::RClassField::AddReadCallbacksFromIORule(const TSchemaRule *rule)
+{
+   auto func = rule->GetReadFunctionPointer();
+   R__ASSERT(func != nullptr);
+   fReadCallbacks.emplace_back([func, classp = fClass, stagingArea = fStagingArea.get()](void *target) {
+      TVirtualObject oldObj{nullptr};
+      oldObj.fClass = classp;
```
pcanal, line 308: This needs to be the '@@' class, is that the case?

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -835,7 +835,7 @@ void ROOT::Experimental::RField<TObject>::ReadGlobalImpl(NTupleSize_t globalInde
 
 void ROOT::Experimental::RField<TObject>::AfterConnectPageSource()
 {
-   if (GetTypeVersion() != 1) {
+   if (GetOnDiskTypeVersion() != 1) {
```
hahnjo, line 937: (note for future reviewers: the in-memory class version is `assert`ed in the constructor)

tree/ntuple/v7/inc/ROOT/RField.hxx
hahnjo: If I understand the implementation correctly, `fieldDesc` is the descriptor corresponding to the class, not of the rule source fields. In that case the proposed addition is incorrect.

(If the addition is correct, the grammar should be `Rules apply` or `A rule applies`)

tree/ntuple/v7/src/RFieldMeta.cxx
hahnjo: As far as I understand, the `fieldDesc` "correspond[s] to the current RClassField instance", so it's not the source of a specific rule

tree/ntuple/v7/src/RFieldMeta.cxx
hahnjo: (side question: can transient members ever be a source? They are not `onfile` after all...)

tree/ntuple/v7/src/RFieldMeta.cxx
hahnjo: A `return false` here means that the rule is currently applied, cf `std::remove_if`

tree/ntuple/v7/src/RFieldMeta.cxx
hahnjo: Not yet in this commit, this change only moves the registration of rules (edit: but I agree it should change in the commit after)

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -198,27 +229,49 @@ void ROOT::Experimental::RClassField::ReadInClusterImpl(RNTupleLocalIndex localI
...
       }
+
+      rules = FindRules(&fieldDesc);
+   }
+
+   for (const auto rule : rules) {
+      AddReadCallbacksFromIORule(rule);
    }
 
    // Iterate over all sub fields in memory and mark those as missing that are not in the descriptor.
```
hahnjo, line 350: Not sure if this is important, but before the refactoring `GetOnDiskId() == kInvalidDescriptorId` would early-exit while now we land here. I think there's no observable difference because `SetArtificial()` applies recursively to all sub-fields, so they should have already been marked by the parent...

tree/ntuple/v7/src/RFieldMeta.cxx
hahnjo: As far as I understand the code, the staging area *is* created regardless of this check, which is only there to determine which `clName` to ask from `TClass::GetClass`.

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -217,26 +217,96 @@ std::size_t ROOT::Experimental::RClassField::AppendImpl(const void *from)
...
+   }
+   auto cl = TClass::GetClass(clName.c_str());
+   R__ASSERT(cl);
+
+   std::size_t stagingAreaSize = 0;
+   for (const auto rule : rules) {
+      for (auto source : TRangeDynCast<TSchemaRule::TSources>(rule->GetSource())) {
+         auto [itr, isNew] = fStagingItems.emplace(source->GetName(), RStagingItem());
+         if (!isNew) {
+            // This source member has already been processed by another rule (and we only support one type per member)
```
hahnjo, line 275: Should the type be checked here?

tree/ntuple/v7/src/RFieldMeta.cxx
jblomer: Rearranged that code block and ensured that whole-object rules are executed last.

tree/ntuple/v7/src/RFieldMeta.cxx
jblomer: As far as I understand, transient member must not be a source; @pcanal ?

tree/ntuple/v7/src/RFieldMeta.cxx
jblomer: I thought that the line before, calling `TClass::GetClass(<class name>)->GetStreamerInfo(<version used in @@>)` is enough. Is another call necessary?

tree/ntuple/v7/src/RFieldMeta.cxx
jblomer: We may skip rules, so we'll check for `stagingAreaSize <= fStagingClass->Size()`

tree/ntuple/v7/src/RFieldMeta.cxx
jblomer: (Not here but in cleanup of the rules vector)

tree/ntuple/v7/src/RFieldMeta.cxx
jblomer: Isn't the point that if `GetOnDiskId() == kInvalidDescriptorId` then `knownSubFields` is empty, so the loop does nothing?

hahnjo: Yes, `knownSubFields` is empty so `knownSubFields.count(field->GetFieldName())` will always be 0, marking all sub-fields as artificial. As I said, this may be fine...

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -355,7 +361,10 @@ void ROOT::Experimental::RClassField::BeforeConnectPageSource(Internal::RPageSou
...
 
       rules = FindRules(&fieldDesc);
-      PrepareStagingArea(rules, desc, fieldDesc);
+      if (!rules.empty()) {
+         SetStagingClass(fieldDesc.GetTypeName(), fieldDesc.GetTypeVersion());
+         PrepareStagingArea(rules, desc, fieldDesc);
+      }
       for (auto &[_, si] : fStagingItems)
          Internal::CallConnectPageSourceOnField(*si.fField, pageSource);
    }
```
hahnjo: I guess this should also move into the `if (!rules.empty())`?

hahnjo (approved):
Looks good from my side. There are still two unresolved threads for @pcanal:
 * https://github.com/root-project/root/pull/17534#discussion_r1930653575
 * https://github.com/root-project/root/pull/17534#discussion_r1934360005

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -186,39 +251,145 @@ std::size_t ROOT::Experimental::RClassField::AppendImpl(const void *from)
...
+         idSourceMember = LookupMember(desc, memberName, subFieldDesc.GetId());
+         if (idSourceMember != kInvalidDescriptorId)
+            return idSourceMember;
+      }
+   }
+
+   return kInvalidDescriptorId;
+}
+
+void ROOT::Experimental::RClassField::SetStagingClass(const std::string &className, unsigned int classVersion)
```
pcanal, line 292: Is there another version of this routine for the case of unversioned classes (where the checksum needs to be used for the searches).

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -186,39 +251,145 @@ std::size_t ROOT::Experimental::RClassField::AppendImpl(const void *from)
...
+{
+   TClass::GetClass(className.c_str())->GetStreamerInfo(classVersion);
+   if (classVersion != GetTypeVersion()) {
+      fStagingClass = TClass::GetClass((className + std::string("@@") + std::to_string(classVersion)).c_str());
```
pcanal, line 296: For the search based on checksum search case, we don't know the version use for the name.  Instead we need to use something along the line of:
```
auto info = TClass::GetClass(className.c_str())->GetStreamerInfo(classVersion);
auto elem = info->GetElement(0);
if (elem && elem->GetType() == TStreamerInfo::kCacheNew)
   fStagingClass = elem->GetClass();
```
which will work for both.

tree/ntuple/v7/src/RFieldMeta.cxx
pcanal: Yes, it should be enough.

tree/ntuple/v7/src/RFieldMeta.cxx
pcanal: Correct.

tree/ntuple/v7/src/RFieldMeta.cxx
jblomer: I am not sure if we can end up in this situation. On write, we record the class version and class checksum and write the corresponding streamer info. On read, we use `FindRules(className, classVersion, classChecksum)` and then, in `SetStagingClass`, we use `TClass::GetClass(className)->GetStreamerInfo(classVersion)` and access the `@@` flavor of the class if the version is different from the current dictionaries.

Can you sketch a case where we would not have the version available?

pcanal: > On write, we record the class version and class checksum and write the corresponding streamer info. 

For unversioned class, the version number assigned (transiently) to the layout is on a first come/first serve basis.
Example:
* In process/build 1. Write unversioned class, checksum A - the assigned version is -1 (or 1 dependency on some code path) - output file A.root
* In process/build 2. Write unversioned class, checksum B -  the assigned version is -1 (or 1 dependency on some code path) - output file B.root
* In process/build 3. Write unversioned class, checksum C -  the assigned version is -1 (or 1 dependency on some code path) - output file C.root

* In process/build 4, read the file in the order A.root, B.root, C.root, then the version assigned are:
* * layout A : version 1
* * layout B : version 2
* * layout C : version 3
* In process/build 4, read the file in the order B.root, C.root, A.root, then the version assigned are:
* * layout A : version 3
* * layout B : version 1
* * layout C : version 2

pcanal: i.e. For unversioned class, there is no stable relationship between the version number and the checksum.  However in a given process, once the StreamerInfo if given a version number, it stays that way.

pcanal: > FindRules(className, classVersion, classChecksum) 

For unversioned classes, the user is consistent and uses only the `checksum` field (and maybe the version field to mark all apply, eg `versions="[1-]"`) and thus `FindRules` will de-facto ignore the `classVersion` argument.

> and then, in SetStagingClass, we use TClass::GetClass(className)->GetStreamerInfo(classVersion)

This works if and only if `classVersion` (or more exactly `fieldDesc.GetTypeVersion()`) was updated at read-time to match the version assigned (during the current process execution) to the `TStreamerInfo` corresponding to the checksum (`fieldDesc->GetTypeChecksum()`)

jblomer: Ok, thanks! I probably won't get to this immediately. I'd suggest to merge this PR as is and fix unversioned classes in a follow-up PR: https://github.com/root-project/root/issues/17713

tree/ntuple/v7/src/RFieldMeta.cxx
pcanal: Will be addressed in subsequent PR, see https://github.com/root-project/root/issues/17713

tree/ntuple/v7/src/RFieldMeta.cxx
pcanal: Fair enough.

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -187,39 +252,145 @@ std::size_t ROOT::Experimental::RClassField::AppendImpl(const void *from)
...
+void ROOT::Experimental::RClassField::AddReadCallbacksFromIORule(const TSchemaRule *rule)
+{
+   auto func = rule->GetReadFunctionPointer();
+   R__ASSERT(func != nullptr);
+   fReadCallbacks.emplace_back([func, stagingClass = fStagingClass, stagingArea = fStagingArea.get()](void *target) {
+      TVirtualObject oldObj{nullptr};
+      oldObj.fClass = stagingClass;
+      oldObj.fObject = stagingArea;
+      func(static_cast<char *>(target), &oldObj);
+      oldObj.fClass = nullptr; // TVirtualObject does not own the value
```
pcanal, line 352: Did you mean:
```suggestion
      oldObj.fObject = nullptr; // TVirtualObject does not own the value
```

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -187,39 +252,145 @@ std::size_t ROOT::Experimental::RClassField::AppendImpl(const void *from)
...
+      fStagingArea = Internal::MakeUninitArray<unsigned char>(stagingAreaSize);
+   }
+}
+
+void ROOT::Experimental::RClassField::AddReadCallbacksFromIORule(const TSchemaRule *rule)
+{
+   auto func = rule->GetReadFunctionPointer();
+   R__ASSERT(func != nullptr);
+   fReadCallbacks.emplace_back([func, stagingClass = fStagingClass, stagingArea = fStagingArea.get()](void *target) {
+      TVirtualObject oldObj{nullptr};
```
pcanal, line 348: Consider (optionally) 
```suggestion
      TVirtualObject onfileObj{nullptr};
```
as I have been migrating other places from the nomenclature `old/new` to `onfile/inmemory`

tree/ntuple/v7/test/CustomStructLinkDef.h
```diff
@@ -61,32 +61,32 @@
...
-//      - The streamer info returns the offset of the wrong data member
 #pragma read sourceClass = "StructWithTransientString" source = "char chars[4]" version = "[1-]" targetClass = \
-   "StructWithTransientString" target = "str" include = "string" code = "{ str = \"ROOT\"; }"
+   "StructWithTransientString" target = "str" include = "string" code = "{ str = std::string{onfile.chars, 4}; }"
 
-#pragma read sourceClass = "StructWithIORules" source = "float a;float b" version = "[1-]" targetClass = \
-   "StructWithIORules" target = "c" code = "{ c = onfile.a + onfile.b; }"
+// Whole object rule (without target member) listed first but should be executed last
+#pragma read sourceClass = "StructWithIORules" version = "[1-]" targetClass = "StructWithIORules" source = "" target = \
+   "" code = "{ newObj->cDerived = 2 * newObj->c; }"
```
pcanal, line 69: (optional)
```suggestion
#pragma read sourceClass="StructWithIORules" version="[1-]" targetClass="StructWithIORules" source="" \
    target ="" code="{ newObj->cDerived = 2 * newObj->c; }"
```
clang-format is mean ... mostly to be cutting of `target =` from its value.

tree/ntuple/v7/test/CustomStructLinkDef.h
```diff
@@ -61,32 +61,32 @@
...
+   "StructWithIORules" target = "" code = "{ }"
 
 // This rule uses a checksum to identify the source class
-#pragma read sourceClass = "StructWithIORules" source = "float checksumA" checksum = "[3494027874]" targetClass = \
+#pragma read sourceClass = "StructWithIORules" source = "" checksum = "[3494027874]" targetClass = \
    "StructWithIORules" target = "checksumA" code = "{ checksumA = 42.0; }"
 // This rule will be ignored due to a checksum mismatch
-#pragma read sourceClass = "StructWithIORules" source = "float checksumB" checksum = "[1]" targetClass = \
-   "StructWithIORules" target = "checksumB" code = "{ checksumB = 0.0; }"
+#pragma read sourceClass = "StructWithIORules" source = "" checksum = "[1]" targetClass = "StructWithIORules" target = \
```
pcanal, line 82: The checksum is a very unlikely value.  Did you mean `version=` or do we need to update the checksum for the correct value?

PR #17536, opened 2025-01-27
Avoid infinite recursion in implicit conversion 

### LINKED ISSUE #17497: Particular combination of function overload and std::runtime_error leads to segfault
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

I can see a segfault in a seemingly innocent situation with an overload set that includes `std::string_view` and `std::vector<std::string>` in which at least one of the overloads throws a runtime error. The gdb stacktrace appears to be recursive, see attachment

[gdb_stacktrace.txt](https://github.com/user-attachments/files/18508619/gdb_stacktrace.txt)

Note that this happens only if I pass an empty Python list `[]` that is then converted implicitly to the `std::vector`. If I pass manually an empty `cppyy.gbl.std.vector[cppyy.gbl.std.string]`, it works. See the reproducer.

This issue is the direct cause of part 1 of https://github.com/root-project/root/issues/17485

### Reproducer

```python
import cppyy

cppyy.cppdef(r"""

void fun(std::string_view, std::string_view){throw std::runtime_error("std::string_view overload");}
void fun(std::string_view, const std::vector<std::string> &){throw std::runtime_error("const std::vector<std::string> & overload");}

""")

# This throws the correct runtime error
cppyy.gbl.fun("", cppyy.gbl.std.vector[cppyy.gbl.std.string]())

# Uncomment the following line to get a segfault
# cppyy.gbl.fun("", [])
```

### ROOT version

master

### Installation method

Any

### Operating system

Linux Fedora 41

### Additional context

_No response_

vepadulano:
Fixes #17497 

@guitargeek These changes, taken from https://github.com/wlav/cppyy-backend/blob/master/clingwrapper/src/clingwrapper.cxx , fix the reproducer at the linked issue. Let's see what the rest of the CI says. Then we may want to discuss how to incorporate these changes into the repository.

### DIFF
--- a/bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx
+++ b/bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx
@@ -67,15 +67,33 @@ static const ClassRefs_t::size_type STD_HANDLE = GLOBAL_HANDLE + 1;
 
 typedef std::map<std::string, ClassRefs_t::size_type> Name2ClassRefIndex_t;
 static Name2ClassRefIndex_t g_name2classrefidx;
+static std::map<std::string, std::string> resolved_enum_types;
 
 namespace {
 
-static inline Cppyy::TCppType_t find_memoized(const std::string& name)
+static inline
+Cppyy::TCppType_t find_memoized_scope(const std::string& name)
 {
     auto icr = g_name2classrefidx.find(name);
     if (icr != g_name2classrefidx.end())
         return (Cppyy::TCppType_t)icr->second;
     return (Cppyy::TCppType_t)0;
+}
+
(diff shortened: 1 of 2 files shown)


bellenot:
So the failure on Windows is due to `throw std::runtime_error` which is not working on Windows x64!!!
```
C:\root-dev\build\x64\gh-17497>root -l
root [0] void fun(){throw std::runtime_error(""); }
root [1] fun()

C:\root-dev\build\x64\gh-17497>
```

vepadulano:
@bellenot thank you for your investigation! I will then disable this test on Windows pending a fix to a much larger issue on the platform.

guitargeek (approved):
Thank you! Good that you fixed the issue, and reducing the diff is always welcome :+1:

PR #17541, opened 2025-01-27
[ntuple] add different modes of descriptor loading to RNTupleSerializer

silverweed:
The default mode - the same we did so far - is "for reading": this means that we not only load the on-disk information into the in-memory descriptor, but we also fixup the suppressed column ranges and the clusters with the info coming from the header extension (deferred columns); this is a required step for reading the rntuple correctly.
We then add two additional modes: "for writing" and "raw". "for writing" is the same as "for reading" except it doesn't do the header extension fixup - this mode will be used by the RNTupleMerger to properly write out an incrementally-merged RNTuple. "raw" simply deserializes the on-disk information without doing any fixup.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleSerialize.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleSerialize.hxx
@@ -60,6 +60,10 @@ Deserialization errors throw exceptions. Only when indicated or when passed as a
 */
 // clang-format on
 class RNTupleSerializer {
+   static RResult<std::vector<RClusterDescriptorBuilder>>
+   DeserializePageListRaw(const void *buffer, std::uint64_t bufSize, DescriptorId_t clusterGroupId,
+                          const RNTupleDescriptor &desc);
+
 public:
    static constexpr std::uint16_t kEnvelopeTypeHeader = 0x01;
    static constexpr std::uint16_t kEnvelopeTypeFooter = 0x02;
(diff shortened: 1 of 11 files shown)


tree/ntuple/v7/inc/ROOT/RNTupleSerialize.hxx
```diff
@@ -275,9 +279,22 @@ public:
...
+      kForWriting,
+      /// Deserializes the descriptor and performs fixup on the suppressed column ranges and on clusters, taking
+      /// into account the header extension. This produces a descriptor that is suitable for reading.
+      kForReading,
+   };
    // The clusters vector must be initialized with the cluster summaries corresponding to the page list
    static RResult<void> DeserializePageList(const void *buffer, std::uint64_t bufSize, DescriptorId_t clusterGroupId,
-                                            RNTupleDescriptor &desc);
+                                            RNTupleDescriptor &desc,
+                                            EDescriptorDeserializeMode mode = EDescriptorDeserializeMode::kForReading);
```
jblomer, line 297: I would not set a default. Let's be specific in each invocation of the method.

tree/ntuple/v7/src/RNTupleSerialize.cxx
```diff
@@ -1891,13 +1890,47 @@ ROOT::RResult<void> ROOT::Experimental::Internal::RNTupleSerializer::Deserialize
...
+   auto clusterBuilders = clusterBuildersRes.Unwrap();
+
+   std::vector<RClusterDescriptor> clusters;
+   clusters.reserve(clusterBuilders.size());
+
+   // Conditionally fixup the clusters depending on the attach purpose
+   switch (mode) {
+   case EDescriptorDeserializeMode::kForReading:
+      for (auto &builder : clusterBuilders) {
+         builder.CommitSuppressedColumnRanges(desc).ThrowOnError();
```
jblomer, line 1917: We don't want to throw but to forward the error in `RResult`

tree/ntuple/v7/src/RNTupleSerialize.cxx
```diff
@@ -1891,13 +1890,47 @@ ROOT::RResult<void> ROOT::Experimental::Internal::RNTupleSerializer::Deserialize
...
+         builder.AddExtendedColumnRanges(desc);
+         clusters.emplace_back(builder.MoveDescriptor().Unwrap());
+      }
+      break;
+   case EDescriptorDeserializeMode::kForWriting:
+      for (auto &builder : clusterBuilders) {
+         builder.CommitSuppressedColumnRanges(desc).ThrowOnError();
```
jblomer, line 1924: see above

jblomer (commented):
In principle looks good to me! Except minor comments, I think we should add a unit test that constructs a descriptor with one suppressed column and one deferred column and exercises the three reconstruction modes.

tree/ntuple/v7/test/ntuple_serialize.cxx
```diff
@@ -1505,3 +1505,277 @@ TEST(RNTuple, SerializeMultiColumnRepresentationIncremental)
...
+         {
+            const auto &fieldDesc = desc.GetFieldDescriptor(desc.FindFieldId("deferred"));
+            const auto &columnIds = fieldDesc.GetLogicalColumnIds();
+            EXPECT_EQ(1u, columnIds.size());
+
+            auto &clusterDesc0 = desc.GetClusterDescriptor(0);
+            EXPECT_EQ(0u, clusterDesc0.GetFirstEntryIndex());
+            const auto columnRange0_0 = clusterDesc0.GetColumnRange(columnIds[0]);
+            RClusterDescriptor::RColumnRange expect0_0{2, 0, 1, {}, false};
+            EXPECT_EQ(expect0_0, columnRange0_0);
```
jblomer, line 1674: Here, we should also find the synthesized page in the page range.

jblomer (approved):
Thanks! LGTM.

PR #17544, opened 2025-01-28
Fix interaction between RSampleInfo and redirected EOS paths

### LINKED ISSUE #17449: [DF] Issues with RDataSpec paths using EOS
### Check duplicate issues.

- [x] Checked for duplicates

### Description

It seems there is some mismatch in the paths when creatind RDataFrame objects using RDataSpec when the input files are stored on EOS.

The corresponding method fills a SampleMap with the identifier filepath+"/"+treenamefor each file in line [443 of the RLoopManager class](https://github.com/root-project/root/blob/v6-34-00/tree/dataframe/src/RLoopManager.cxx#L443). 
Later, when running the TreeReader, the file linked with the TTree is compared to this metadata entry in line [781 of the same class](https://github.com/root-project/root/blob/v6-34-00/tree/dataframe/src/RLoopManager.cxx#L781). When the input file is available locally, this works fine but with files located on eos it causes issues.

The filepath of the format /eos/home-w/name/my/path/file.root is translated to the filepath associated with the TTree later but has the format root://eoshome-w.cern.ch//eos/user/name/my/path/file.root. In summary, the code looks for the key root://eoshome-w.cern.ch//eos/user/na,me/my/path/file.root/reco in a map containing the key /eos/home-w/name/my/path/file.root/reco which leads to a runtime error.

### Reproducer

The lines above indicate the problem

### ROOT version

ROOT Version: 6.35.01
Built for linuxx8664gcc on Dec 02 2024, 08:38:30
From heads/master@v6-35-01-366-g3dbc5f7f5d

### Installation method

build from source

### Operating system

Ubuntu 24.10

### Additional context

_No response_

vepadulano:
When the input files are paths to FUSE-mounted EOS files, during the event loop these paths will be redirected to the corresponding xroot EOS URL, in TFile::Open.

This was causing a bad interaction with the sample metadata and the subsequent usage in the event loop, e.g. through DefinePerSample. Specifically, we fill a map with the metadata at construction time, which includes the input file paths (without redirection). These will then be irretrievable during the event loop since the map key will not correspond to the redirected path. Fix this by also adding the redirected path during the filling of the map in ChangeSpec.

This PR fixes #17449 

@TomasDado FYI, this is an alternative to  https://github.com/root-project/root/pull/17537

### DIFF
--- a/tree/dataframe/src/RLoopManager.cxx
+++ b/tree/dataframe/src/RLoopManager.cxx
@@ -42,6 +42,24 @@
 #include "ROOT/RNTupleDS.hxx"
 #endif
 
+#ifdef R__UNIX
+// Functions needed to perform EOS XRootD redirection in ChangeSpec
+#include "TEnv.h"
+#include "TSystem.h"
+#ifndef R__FBSD
+#include <sys/xattr.h>
+#else
+#include <sys/extattr.h>
+#endif
+#ifdef R__MACOSX
+/* On macOS getxattr takes two extra arguments that should be set to 0 */
+#define getxattr(path, name, value, size) getxattr(path, name, value, size, 0u, 0)
+#endif
+#ifdef R__FBSD
+#define getxattr(path, name, value, size) extattr_get_file(path, EXTATTR_NAMESPACE_USER, name, value, size)
+#endif
+#endif
+
 #include <algorithm>
 #include <atomic>
 #include <cassert>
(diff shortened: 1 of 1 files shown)


dpiparo:
@amadio

amadio (approved):
Vincenzo and I had a chat about the potential solution. This is one of the solutions we discussed. Looks good to me.

PR #17547, opened 2025-01-28
[tree] Apply clang-format to TTreeReader[,Value,Array].[h,cxx]

vepadulano:
This is just  running `clang-format -i filename`, also done in preparation for more changes needed by TTree-based processing in RDataFrame.


pcanal (approved):
I am pondering about this NFC change.  On the one hand it gives a clean slate for further development but on the other hand its extensiveness means that future `git blame` tracking through this barrier will be a tad bit harder.

That said, I am hesitantly approving the merge as there change does not seem to include/touch any of the clang-format weakeness

Issue #14153, opened 2023-11-30
TH2 not drawn if only TH2::AddBinContent is called (Entries=0). For TH1 it works.


ferdymercury:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

Assume one only uses TH2::AddBinContent to fill a 2D histogram.
Then you draw it with COLZ.
Resulting canvas is empty.
In contrast, if you do the same with TH1::AddBinContent, it works without problems.

### Reproducer

```
auto h2 = new TH2I("h2", "h2", 10, 0, 1, 2, 0, 2);
h2->AddBinContent(16,5);
h2->Draw("COLZ");
```

Nothing happens now, canvas is empty.

You are forced to:

```
h2->SetEntries(1)
gPad->Modified();
gPad->Update();
```

In contrast, with TH1, all works fine without calling SetEntries:

```
auto h = new TH1I("h", "h", 10, 0, 1);
h->AddBinContent(3,5)
h->Draw()
```

### ROOT version

```
ROOT v6.30/02
Built for linuxx8664gcc on Nov 27 2023, 19:50:38
From tags/v6-30-02@
With c++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Binary directory: /opt/root/bin
```

### Installation method

Binary

### Operating system

Ubuntu 22.04

ferdymercury:
Good catch, thanks.

Alternatively, I would suggest adding in the Doxygen documentation of all the "SetBinContent" and "AddBinContent" functions, that you need to set the "entries" per hand later on, so that the result is drawn.

dpiparo:
@couet what is the outcome here?

PR #17548, opened 2025-01-28
[hist] paint TH2 even if nEntries==0, as already happens for TH1

ferdymercury:
# This Pull request:

## Changes or fixes:

If you fill a TH1 using only AddBinContent, the resulting nEntries=0, and call Draw(), it works. If you fill a TH2 using only AddBinContent, the resulting nEntries=0, and call Draw("COLZ"), it does not work. This commit tries to fix it so that it also works for the 2D case, for consistency.

Fixes https://github.com/root-project/root/issues/14153

## Checklist:

- [x] tested changes locally


couet (approved):
Yes it makes sense to me. I do not have in mind some possible side effects.

PR #17558, opened 2025-01-29
[math][minuit2] Add new version of Fumili algorithm using trust region method

lmoneta:
Implement a new version of the Fumili algorithm for minimization in Minuit2 using trust-region instead of line search for finding the next step. See this [presentation](https://indico.jlab.org/event/459/contributions/11597/attachments/9469/13862/Minuit2_CHEP2023.pdf). 

This PR implements also: 
-  the possibility to change the verbosity in fitting histograms. In addition to option `V`, the fitting options `VV` and `VVV` have been added
- a new version of gsl non-linear multidimensional fit algorithm. New classes were impelmented in GSL  versions 2, and now this PR impelments a new wrapper for the new GSL routines. Various different solver can be used for the `GSLMultiFit` minimization algorith, when fitting histograms.

### DIFF
--- a/math/fumili/src/TFumili.cxx
+++ b/math/fumili/src/TFumili.cxx
@@ -110,6 +110,8 @@ function argument.
 #include "TList.h"
 #include "TVirtualFitter.h"
 
+#include <iostream>
+
 
 extern void H1FitChisquareFumili(Int_t &npar, Double_t *gin, Double_t &f, Double_t *u, Int_t flag);
 extern void H1FitLikelihoodFumili(Int_t &npar, Double_t *gin, Double_t &f, Double_t *u, Int_t flag);
@@ -1067,6 +1069,27 @@ Bool_t TFumili::IsFixed(Int_t ipar) const
    else                return kFALSE;
 }
 
+void PrintVector(const char * name, int n, double * x) {
+   // print vector data
+   std::cout << name << " : {";
+   for (int i  = 0; i < n; i++)
+      std::cout << "  " << x[i];
+   std::cout << " }\n";
+}
+void PrintMatrix(const char * name, int n, double * x) {
+   // print matrix data
+   std::cout << name << " : \n";
+   int index = 0;
+   for (int i  = 0; i < n; i++) {
+      for (int j  = 0; j <= i; j++) {
+         std::cout << "  " << x[index];
+         index++;
+      }
+      std::cout << std::endl;
+   }
+   std::cout << "\n";
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 /// Main minimization procedure
 ///
@@ -1092,6 +1115,8 @@ Int_t TFumili::Minimize()
    //
    Int_t parn;
 
+   // I think this funciton call is needed for getting fA and fNpar
+   // and fAMX fAMN ?
    if(fFCN) {
       Eval(parn,fGr,fS,fA,9); fNfcn++;
    }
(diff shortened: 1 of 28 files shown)


math/minuit2/src/FumiliBuilder.cxx
```diff
@@ -263,19 +289,187 @@ FunctionMinimum FumiliBuilder::Minimum(const MnFcn &fcn, const GradientCalculato
...
+               // Cauchy point can be inside the trust region
+               double tau = std::min( (normGrad2*normGrad)/(gHg*delta), 1.0);
+               if (tau == 1.0) {
+                  // Cauchy is at the boundary
+                  step = - (delta/ normGrad) * s0.Gradient().Grad();
+                  if (scaleTR)
+                    step = Dinv2 * step;
+
+                  print.Debug("tau=1. - Use as new point the Cauchy  point - along gradient with norm=delta ", delta);
+               } else {
```
guitargeek, line 375: Are some of the code paths not redundant here? You have two code branches for where the Cauchy point is at the boundary: like 360 and line 368 (with `tau == 1.0`). Maybe the second branch is redundant? That would also avoid code repetition.

math/minuit2/src/MnPosDef.cxx
```diff
@@ -100,7 +101,9 @@ MinimumError MnPosDef::operator()(const MinimumError &e, const MnMachinePrecisio
 
    print.Warn("Matrix forced pos-def by adding to diagonal", pAdd);
 
-   return MinimumError(err, MinimumError::MnMadePosDef);
+   MinimumError epdf(err, MinimumError::MnMadePosDef);
+   std::cout << epdf.Hessian().size() << "  " << epdf.Hessian().Nrow() << std::endl;
```
guitargeek, line 105: Do we need this printout that looks it was added for debugging?

It comes from the 4th commit. Maybe that commit should be amended?

math/minuit2/inc/Minuit2/MinimumError.h
```diff
@@ -87,9 +91,10 @@ class MinimumError {
    bool HasReachedCallLimit() const { return GetStatus() == MnReachedCallLimit; }
    bool IsAvailable() const { return GetStatus() != MnUnset; }
 
-private:
+public:
```
guitargeek, line 94:
```suggestion
private:
```
Swtiching this `private` to `public` looks like a development artifact to me.

math/fumili/src/TFumili.cxx
```diff
@@ -1135,15 +1160,25 @@ Int_t TFumili::Minimize()
...
+   }
+
    // fS - objective function value - zero first
    fS = 0.;
    // n0 - number of variable parameters in fit
    n0 = 0;
    for( i = 0; i < n; i++) {
       fGr[i]=0.; // zero gradients
-      if (fPL0[i] > .0) {
+      if (fPL0[i] > .0) {   // fLO is negstive foa fixed parameters
```
guitargeek, line 1181:
```suggestion
      if (fPL0[i] > .0) {   // fLO is negative for fixed parameters
```

guitargeek (changes requested):
Some improvement suggestions inline. Also, a few comments about the individual commits:

3.  **[math][minuit2] Fix a bug in the assignment operator of LaSymMatrix**

    The commit is completely unrelated to the commit message. It only changes the logging in `MnLineSearch`. What is the motivation for this change? The commit should either be dropped or the commit message should be updated.

7. **[math][mathmore] Add new fitter algorithm from GSL under GSLMultiFit**

    Commit does not compile because of some artifacts from conflict resolution. It will be fixed up in the commit that follows, but it would be better if each commit compiles correctly. This helps when debugging problems with git bisect, and especially with Minuit2, the problems can often be quite subtle and bisecting is helpful.

9. **[math] Update builtin gsl version to 2.8 (from 2.5)**
    This commit is simply reverted again in the commit that follows. Can this commit be dropped?

guitargeek (approved):
Thank you very much for the contribution and addressing the review comments! Looks good to me now.

PR #17559, opened 2025-01-29
[ntuple] Remove RContext::fHeaderExtensionOffset

silverweed:
The offset can be easily derived without the need to store it.

## Checklist:

- [x] tested changes locally


enirolf (approved):
LGTM

PR #17563, opened 2025-01-29
[ntuple] properly support incremental merging with Union mode

silverweed:
Based on #17559. Opening as a draft because I'd like to add some more unit tests.

This is the last(*) PR of a series that adds proper support for incremental merging with Union mode, i.e. proper handling of deferred columns in the Merger.
With this change we should be able to support [ATLAS-like workflows](https://gitlab.cern.ch/amete/rootparallelmerger/-/blob/master/src/ParallelFileMerger.cpp?ref_type=heads) where the TFileMerger is used to incrementally construct an RNTuple from a sequence of files that may contain additional fields compared to the already-merged ones.

(*) at least for a minimal working case - more edge cases remain to be tested and likely fixed.

## A brief overview
When Union-merging files containing compatible-but-different models (e.g. the second file contains an additional field), we resort to Late Model Extension in the RNTupleMerger and therefore produce a merged RNTuple containing deferred columns in the header extension.
When we incrementally merge a new file, we need to read the metadata from the existing RNTuple, preserve the header extension as-is and possibly Late Model Extend again the new model created from this metadata. This requires some care because we need to explicitly keep all the extended fields and columns in the header extension when writing the file back (rather than merging them in the regular header).
This requires dropping the implicit assumption in our serializer API that assumes that no header extension is present when serializing the header the first time.

## Checklist:

- [x] tested changes locally

cc @amete who is probably interested, as he provided the ParallelFileMerger example linked above

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
@@ -548,15 +548,26 @@ private:
    /// Free text from the user
    std::string fDescription;
 
-   std::uint64_t fOnDiskHeaderXxHash3 = 0; ///< Set by the descriptor builder when deserialized
+   DescriptorId_t fFieldZeroId = kInvalidDescriptorId; ///< Set by the descriptor builder
+
+   std::uint64_t fNPhysicalColumns = 0; ///< Updated by the descriptor builder when columns are added
    std::uint64_t fOnDiskHeaderSize = 0;    ///< Set by the descriptor builder when deserialized
+
+   std::set<unsigned int> fFeatureFlags;
+   std::unordered_map<DescriptorId_t, RFieldDescriptor> fFieldDescriptors;
+   std::unordered_map<DescriptorId_t, RColumnDescriptor> fColumnDescriptors;
+
+   std::vector<RExtraTypeInfoDescriptor> fExtraTypeInfoDescriptors;
+   std::unique_ptr<RHeaderExtension> fHeaderExtension;
+
+   //// All fields above are part of the schema and are cloned when creating a new descriptor from a given one
+   //// (see CloneSchema())
+
+   std::uint64_t fOnDiskHeaderXxHash3 = 0; ///< Set by the descriptor builder when deserialized
    std::uint64_t fOnDiskFooterSize = 0; ///< Like fOnDiskHeaderSize, contains both cluster summaries and page locations
 
    std::uint64_t fNEntries = 0;         ///< Updated by the descriptor builder when the cluster groups are added
    std::uint64_t fNClusters = 0;        ///< Updated by the descriptor builder when the cluster groups are added
-   std::uint64_t fNPhysicalColumns = 0; ///< Updated by the descriptor builder when columns are added
-
-   DescriptorId_t fFieldZeroId = kInvalidDescriptorId; ///< Set by the descriptor builder
 
    /**
     * Once constructed by an RNTupleDescriptorBuilder, the descriptor is mostly immutable except for set of
@@ -567,9 +578,6 @@ private:
     */
    std::uint64_t fGeneration = 0;
 
-   std::set<unsigned int> fFeatureFlags;
-   std::unordered_map<DescriptorId_t, RFieldDescriptor> fFieldDescriptors;
-   std::unordered_map<DescriptorId_t, RColumnDescriptor> fColumnDescriptors;
    std::unordered_map<DescriptorId_t, RClusterGroupDescriptor> fClusterGroupDescriptors;
    /// References cluster groups sorted by entry range and thus allows for binary search.
    /// Note that this list is empty during the descriptor building process and will only be
@@ -578,11 +586,14 @@ private:
    /// May contain only a subset of all the available clusters, e.g. the clusters of the current file
    /// from a chain of files
    std::unordered_map<DescriptorId_t, RClusterDescriptor> fClusterDescriptors;
-   std::vector<RExtraTypeInfoDescriptor> fExtraTypeInfoDescriptors;
-   std::unique_ptr<RHeaderExtension> fHeaderExtension;
 
    // We don't expose this publicly because when we add sharded clusters, this interface does not make sense anymore
    DescriptorId_t FindClusterId(NTupleSize_t entryIdx) const;
+
+   /// Fills `into` with the schema information about this RNTuple, i.e. all the information needed to create
+   /// a new RNTuple with the same schema as this one but not necessarily the same clustering. This is used
+   /// when merging two RNTuples.
+   RNTupleDescriptor CloneSchema() const;
 
 public:
    static constexpr unsigned int kFeatureFlagTest = 137; // Bit reserved for forward-compatibility testing
@@ -1398,6 +1409,8 @@ public:
    RResult<void> EnsureValidDescriptor() const;
    const RNTupleDescriptor &GetDescriptor() const { return fDescriptor; }
    RNTupleDescriptor MoveDescriptor();
+
+   void CreateFromSchema(const RNTupleDescriptor &descriptor);
 
    void SetNTuple(const std::string_view name, const std::string_view description);
    void SetFeature(unsigned int flag);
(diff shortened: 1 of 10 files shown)


tree/ntuple/v7/src/RNTupleSerialize.cxx
```diff
@@ -247,13 +247,21 @@ std::uint32_t SerializePhysicalColumn(const ROOT::Experimental::RColumnDescripto
...
                                        std::span<const ROOT::Experimental::DescriptorId_t> fieldList,
                                        const ROOT::Experimental::Internal::RNTupleSerializer::RContext &context,
-                                       void *buffer)
+                                       void *buffer, bool forHeaderExtension)
 {
    auto base = reinterpret_cast<unsigned char *>(buffer);
    auto pos = base;
    void **where = (buffer == nullptr) ? &buffer : reinterpret_cast<void **>(&pos);
 
+   const auto *xHeader = !forHeaderExtension ? desc.GetHeaderExtension() : nullptr;
```
enirolf, line 256: `desc.GetHeaderExtension()` already returns `nullptr` if there is none. Is there any particular reason to explicitly use a boolean and set it to `nullptr` (same question for `SerializeAliasColumnsOfFields`)?

tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -1399,6 +1410,8 @@ public:
    const RNTupleDescriptor &GetDescriptor() const { return fDescriptor; }
    RNTupleDescriptor MoveDescriptor();
 
+   void CreateFromSchema(const RNTupleDescriptor &descriptor);
```
enirolf, line 1415: To me this method name is a bit vague, perhaps something like `SetSchemaFromExisting`?

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -232,7 +232,7 @@ try {
    *this = *outFile->Get<ROOT::RNTuple>(ntupleName.c_str());
 
    return 0;
-} catch (const RException &ex) {
+} catch (const std::exception &ex) {
```
enirolf, line 235: Why this change?

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -960,9 +947,14 @@ ROOT::RResult<void> RNTupleMerger::Merge(std::span<RPageSource *> sources, const
       }
    }
 
-   RNTupleMergeData mergeData{sources, *fDestination, mergeOpts};
+   // we should have a model if and only if the destination is initialized.
+   if (!!fModel != fDestination->IsInitialized()) {
+      return R__FAIL("Passing an already-initialized destination to RNTupleMerger (i.e. trying to do incremental "
+                     "merging) can only be done if you provided a valid RNTupleModel");
```
enirolf, line 953:
```suggestion
      return R__FAIL("passing an already-initialized destination to RNTupleMerger (i.e. trying to do incremental "
                     "merging) can only be done with a valid RNTupleModel");
```

enirolf (changes requested):
I have some minor comments but looks good in general! However I don't have the full RNTupleMerger picture so an extra set of eyes will be necessary.

tree/ntuple/v7/src/RNTupleSerialize.cxx
silverweed: The reason is that we can have a header extension even if `forHeaderExtension == false`, but in that case we want to ignore it

tree/ntuple/v7/src/RNTupleMerger.cxx
silverweed: I prefer the longer version because it might not be immediately clear where this "valid RNTupleModel" should be provided. In fact, I think I want to make it even more obvious, like "if you provided a valid RNTupleModel when constructing the RNTupleMerger" (because it's a bit of a long-distance relation, where you might have constructed the merger far away in the code from the place where you call Merge. I think it's helpful to make the user's life easier by being very verbose with this error)

hahnjo (commented):
Some comments inline. There are again many unrelated changes sprinkled into the various commits which makes it very hard to follow. As a suggestion, it may be worth splitting out the descriptor work into a separate PR and get the cloning and serialization of "nonstandard" header right before plugging it into the incremental merging (now that you have a working prototype and all individual pieces figured out)

silverweed:
Marking as draft because I'll split it into 2 PRs as suggested by @hahnjo

tree/ntuple/v7/inc/ROOT/RPageStorage.hxx
```diff
@@ -529,7 +530,7 @@ public:
    void UpdateExtraTypeInfo(const RExtraTypeInfoDescriptor &extraTypeInfo) final;
 
    /// Initialize sink based on an existing descriptor and fill into the descriptor builder.
-   void InitFromDescriptor(const RNTupleDescriptor &descriptor);
+   [[nodiscard]] std::unique_ptr<RNTupleModel> InitFromDescriptor(const RNTupleDescriptor &descriptor);
```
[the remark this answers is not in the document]
silverweed: > did you consider passing the model as a parameter?

How would this change the lifetime problem compared to the return value? In the end we have no control on when the model gets destroyed on the caller's site either way, no?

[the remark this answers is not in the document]
silverweed: Also, the lifetime must be at least as long as the sink, not necessarily longer (it's fine if they get dropped at the same time)

tree/ntuple/v7/src/RNTupleMerger.cxx
enirolf: My suggestion mainly is a stylistic one, stemming from the fact that all of the other error messages (as far as I'm aware) are written in the passive voice and non-capitalized. I agree with the verbosity! I would personally prefer to try to keep consistent with the style of phrasing in the errors, but also open to have my mind changed.

silverweed: I understand your point and agree with it to an extent: I'm certainly fixing the non-capitalization, but I wouldn't know how to write the error in a way that doesn't sound weird in a passive voice - and I don't believe it would add value in this case. I'd argue that for user-facing messages it's more important to be easy to read and act upon, rather than keeping a super-consistent formal style (and passive style sometimes hinders readability imho).
However if you have ideas on how to rephrase this message I'm open for suggestions!

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -726,7 +728,7 @@ static void GenerateExtraDstColumns(size_t nClusterEntries, std::span<RColumnMer
          const auto pageSize = (i < nPages - 1) ? kPageSizeLimit : nBytesOnStorage - kPageSizeLimit * (nPages - 1);
          const auto checksumSize = RPageStorage::kNBytesPageChecksum;
          const auto bufSize = pageSize + checksumSize;
-         auto &buffer = sealedPageData.fBuffers.emplace_back(new unsigned char[bufSize]);
+         auto &buffer = sealedPageData.fBuffers.emplace_back(new std::uint8_t[bufSize]);
```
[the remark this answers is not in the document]
silverweed: It was for consistency with the rest of the file, that uses std::uint8_t for all its buffers (which I personally like more than `unsigned char`).
However, if we want to keep consistency in all RNTuple code we should decide on that and change all the other uses (though I'd say in a separate PR)

silverweed:
Split into 3 separate PR, in order:

- removing a wrong assertion #17596
- adding CloneSchema to the descriptor and fixing InitFromDescriptor #17597
- integrating the changes into the Merger and properly handling deferred columns to support incremental merging #17598

tree/ntuple/v7/inc/ROOT/RPageStorage.hxx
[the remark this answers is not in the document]
hahnjo: > How would this change the lifetime problem compared to the return value?

That's true. IMO it makes it less likely to use wrongly than a unique ptr return value where you had to apply a `[[nodiscard]]` to avoid one source of problems already.

> the lifetime must be at least as long as the sink, not necessarily longer

Actually, I think we have to destroy the model before the sink because the fields are connected to the sink. We have to check `RNTupleFillContext` regarding the order of destruction.

tree/ntuple/v7/src/RNTupleMerger.cxx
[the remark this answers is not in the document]
hahnjo: Regardless of how we decide, I'm against "sneaking in" these changes in other commits

PR #17565, opened 2025-01-29
[ntuple] RMiniFile: write StreamerInfo with the correct compression

silverweed:
Currently RMiniFile doesn't honor the compression settings set in the RNTupleWriteOptions when writing out the StreamerInfo, but instead it always ZLIB-compresses it.
With this change, it will use the same compression as it was specified in the options (or 505 if unspecified).
This makes the behavior more consistent with the TFile, where the TFile compression determines also the StreamerInfo's compression (whereas now RNTuple may export a compressed StreamerInfo even if the TFile's compression is 0).

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/src/RMiniFile.cxx
+++ b/tree/ntuple/v7/src/RMiniFile.cxx
@@ -1218,10 +1218,10 @@ void ROOT::Experimental::Internal::RNTupleFileWriter::Commit(int compression)
       return;
    }
 
-   WriteTFileNTupleKey();
-   WriteTFileKeysList();
+   auto anchorSize = WriteTFileNTupleKey(compression);
+   WriteTFileKeysList(anchorSize); // NOTE: this is written uncompressed
    WriteTFileStreamerInfo(compression);
-   WriteTFileFreeList();
+   WriteTFileFreeList(); // NOTE: this is written uncompressed
 
    // Update header and TFile record
    memcpy(fFileSimple.fHeaderBlock, &fFileSimple.fControlBlock->fHeader, fFileSimple.fControlBlock->fHeader.GetSize());
@@ -1398,9 +1398,8 @@ void ROOT::Experimental::Internal::RNTupleFileWriter::WriteTFileStreamerInfo(int
    const auto bufPayload = buffer.Buffer() + keyLen;
    const auto lenPayload = buffer.Length() - keyLen;
 
-   RNTupleCompressor compressor;
    auto zipStreamerInfos = MakeUninitArray<unsigned char>(lenPayload);
-   auto szZipStreamerInfos = compressor.Zip(bufPayload, lenPayload, compression, zipStreamerInfos.get());
+   auto szZipStreamerInfos = RNTupleCompressor::Zip(bufPayload, lenPayload, compression, zipStreamerInfos.get());
 
    fFileSimple.WriteKey(zipStreamerInfos.get(), szZipStreamerInfos, lenPayload,
                         fFileSimple.fControlBlock->fHeader.GetSeekInfo(), RTFHeader::kBEGIN, "TList", "StreamerInfo",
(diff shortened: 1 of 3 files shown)


pcanal:
For context, the same PR (in spirit) was applied to `TFile` by commit b9a212ac5a9ee6da33b41446d2da50b66b320985 (i.e. it used to be the case for `TFile` that the `StreamerInfo` record was always compressed with zlib but this was removed)

tree/ntuple/v7/src/RMiniFile.cxx
```diff
@@ -1220,7 +1220,7 @@ void ROOT::Experimental::Internal::RNTupleFileWriter::Commit()
 
    WriteTFileNTupleKey();
    WriteTFileKeysList();
-   WriteTFileStreamerInfo();
+   WriteTFileStreamerInfo(compression);
```
pcanal, line 1223: It seems odd that the compression argument is only needed here.   What compression setting is used for the list of keys, free list and for the `RNTuple` key?

silverweed: I am under the impression that both of them are always uncompressed, even in regular TFiles.

jblomer (approved):

hahnjo:
So our first `RNTuple.root` is not fully uncompressed :cry:

tree/ntuple/v7/src/RMiniFile.cxx
silverweed: @pcanal let me know if you have other questions/comments or if this thread is resolved

pcanal: > I am under the impression that they are always uncompressed,

What lead you to this conclusion?

> the first two are uncompressed even in regular TFiles as far as I saw

Indeed the list of keys is not compressed (not sure why though) but I don't see a mechanism why the `RNTuple` anchor would not be compressed (besides maybe the compression failing to save space)

silverweed: > What lead you to this conclusion?

Reading the code in RMiniFile the only time we call `Zip` is in the WriteStreamerInfo function. Also, all RNTuple files I've inspected so far had both sections uncompressed, regardless of whether they have been written through RMiniFile or TFile.

> but I don't see a mechanism why the RNTuple anchor would not be compressed

I suppose it may be compressed if we manually wrote a RNTuple anchor through the TFile API, but if we pass through RMiniFile reader it will always be uncompressed (which is a behavior we may or may not want to change)

pcanal: >  Also, all RNTuple files I've inspected so far had both sections uncompressed, regardless of whether they have been written through RMiniFile or TFile.

That was really the question at hand.  Is it compressed in `TFile`?  It should have been but if not, why/how is it not compressed.

>  (which is a behavior we may or may not want to change - though not in this PR I'd argue)

As we are touching that code, I would include the change also in this PR.  The new code is:
```
   WriteTFileNTupleKey();
   WriteTFileKeysList();
   WriteTFileStreamerInfo(compression);
```
with `compression being a new variable.  This leaves 'ambiguous' whether the other two are 'not compressed' or 'compressed with an arbitrary value ignoring the user specified compression for undisclosed reasons'.
At the very least we would have:
```
   WriteTFileNTupleKey(); // Not compressed yet
   WriteTFileKeysList();     // Intentionally never compressed.
   WriteTFileStreamerInfo(compression);
```
or better
```
   WriteTFileNTupleKey(compress /* or 0 if one wants to keep the current erroneous behavior */ );
   WriteTFileKeysList();     // Intentionally never compressed.
   WriteTFileStreamerInfo(compression);
```

silverweed: I added a commit where the Anchor gets compressed as well. cc @jblomer

pcanal: Thanks.

pcanal (approved):
LGTM

PR #17568, opened 2025-01-29
[ntuple] RMiniFile: properly write the free slot's nbytes

silverweed:
# This Pull request:
depends on https://github.com/root-project/root/pull/17575.

Fixes the case in RMiniFileWriter where a RBlob is allocated inside a free slot that is bigger than the blob itself. In that case allocating the TKey also needs to update the header of the remaining free slot after the space for the blob was reserved (see [TKey::Create](https://github.com/root-project/root/blob/master/io/io/src/TKey.cxx#L510-L516)). This is normally done by `TKey::WriteFile`, but since RMiniFile bypasses that it needs to write the updated nbytes to disk manually.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/src/RMiniFile.cxx
+++ b/tree/ntuple/v7/src/RMiniFile.cxx
@@ -624,6 +624,8 @@ public:
       *seekKey = fSeekKey;
    }
 
+   bool WasAllocatedInAFreeSlot() const { return fLeft > 0; }
+
    ClassDefInlineOverride(RKeyBlob, 0)
 };
 
@@ -1097,6 +1099,13 @@ ROOT::Experimental::Internal::RNTupleFileWriter::RFileProper::ReserveBlobKey(siz
       unsigned char localKeyBuffer[kBlobKeyLen];
       PrepareBlobKey(offsetKey, nbytes, len, localKeyBuffer);
       Write(localKeyBuffer, kBlobKeyLen, offsetKey);
+   }
+
+   if (keyBlob.WasAllocatedInAFreeSlot()) {
+      // If the key was allocated in a free slot, the last 4 bytes of its buffer contain the new size
+      // of the remaining free slot and we need to write it to disk before the key gets destroyed at the end of the
+      // function.
+      Write(keyBlob.GetBuffer() + nbytes, sizeof(Int_t), offsetKey + kBlobKeyLen + nbytes);
    }
 
    auto offsetData = offsetKey + kBlobKeyLen;
--- a/tree/ntuple/v7/test/CustomStruct.hxx
+++ b/tree/ntuple/v7/test/CustomStruct.hxx
@@ -305,4 +305,8 @@ struct ThrowForVariant {
    ThrowForVariant &operator=(const ThrowForVariant &) = default;
 };
 
+struct RelativelyLargeStruct {
+   char fDummy[250];
+};
+
 #endif
--- a/tree/ntuple/v7/test/CustomStructLinkDef.h
+++ b/tree/ntuple/v7/test/CustomStructLinkDef.h
@@ -105,4 +105,6 @@
 
 #pragma link C++ class ThrowForVariant + ;
 
+#pragma link C++ class RelativelyLargeStruct + ;
+
 #endif
--- a/tree/ntuple/v7/test/ntuple_minifile.cxx
+++ b/tree/ntuple/v7/test/ntuple_minifile.cxx
@@ -5,6 +5,7 @@
 #include <TVector2.h>
 #include <TVector3.h>
 #include <TVirtualStreamerInfo.h>
+#include <TFree.h>
 
 #include <cstring>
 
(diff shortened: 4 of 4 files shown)


tree/ntuple/v7/src/RMiniFile.cxx
```diff
@@ -1099,6 +1101,14 @@ ROOT::Experimental::Internal::RNTupleFileWriter::RFileProper::ReserveBlobKey(siz
       Write(localKeyBuffer, kBlobKeyLen, offsetKey);
    }
 
+   if (keyBlob.WasAllocatedInAFreeSlot()) {
+      // If the key was allocated in a free slot, the last 4 bytes of its buffer contain the new size
+      // of the remaining free slot and we need to write it to disk before the key gets destroyed at the end of the
+      // function.
+      auto nbytesFree = *(Int_t *)(keyBlob.GetBuffer() + nbytes);
```
pcanal, line 1108: Do we need to turn it to a negative value?

silverweed: No, the value in the buffer is already the correct (negative) value (see [here](https://github.com/root-project/root/blob/master/io/io/src/TKey.cxx#L514-L515))

hahnjo, line 1108: The pointer resulting from `keyBlob.GetBuffer() + nbytes` may not be aligned, which could break on some architectures. I think we must use `memcpy` (which I think also avoids any endianess problems)

hahnjo (commented):
As discussed on Mattermost, we should have a test for this (there's already some for the key structure). Also can you amend the commit message to contain some of the details currently found in the PR summary?

tree/ntuple/v7/src/RMiniFile.cxx
pcanal: The value in the buffer is byte swapped.

tree/ntuple/v7/src/RMiniFile.cxx
```diff
@@ -1099,6 +1101,15 @@ ROOT::Experimental::Internal::RNTupleFileWriter::RFileProper::ReserveBlobKey(siz
       Write(localKeyBuffer, kBlobKeyLen, offsetKey);
    }
 
+   if (keyBlob.WasAllocatedInAFreeSlot()) {
+      // If the key was allocated in a free slot, the last 4 bytes of its buffer contain the new size
+      // of the remaining free slot and we need to write it to disk before the key gets destroyed at the end of the
+      // function.
+      char nbytesFree[sizeof(Int_t)];
+      memcpy(nbytesFree, keyBlob.GetBuffer() + nbytes, sizeof(nbytesFree));
+      Write(&nbytesFree, sizeof(nbytesFree), offsetKey + kBlobKeyLen + nbytes);
```
pcanal, line 1110: Why not?:
```suggestion
      Write(keyBlob.GetBuffer() + nbytes, sizeof(nbytesFree), offsetKey + kBlobKeyLen + nbytes);
```

tree/ntuple/v7/src/RMiniFile.cxx
pcanal: I see that https://github.com/root-project/root/pull/17568#discussion_r1935318321 raised the same point and was addressed.

silverweed:
@pcanal is there anything else blocking this PR?

io/io/inc/TFile.h
```diff
@@ -50,9 +52,86 @@ class TProcessID;
...
+   };
+
+   std::uint64_t fAddr = 0;
+   EType fType = kError;
+   std::uint32_t fLen = 0;
+
+   // these are only valid for Keys
+   Version_t fKeyVersion = 0;
+   Int_t fObjLen = 0;
+   Int_t fDatime = 0;
```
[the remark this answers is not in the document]
pcanal: but let's leave it as is for now.

io/io/inc/TFile.h
```diff
@@ -50,9 +52,86 @@ class TProcessID;
...
+   TIterator begin() const { return TIterator(fFile, 0); }
+   TIterator end() const { return TIterator(fFile, -1); }
+};
+
+} // namespace ROOT::Detail
+
 class TFile : public TDirectoryFile {
   friend class TDirectoryFile;
   friend class TFilePrefetch;
+  friend class ROOT::Detail::TKeyMapIterable::TIterator;
```
pcanal, line 134: What private information does the iterator access?

silverweed: `fBEGIN`

pcanal (approved):
LGTM.  Thanks.

io/io/inc/TFile.h
pcanal: Annoying but fair enough (The alternative would be to pass `fBEGIN` to the `TKeyMapIterable` constructor.

PR #17574, opened 2025-01-30
[ntuple] Fix-up read options

jblomer:
Preparation for moving read options out of experimental.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleReadOptions.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleReadOptions.hxx
@@ -19,6 +19,18 @@
 namespace ROOT {
 namespace Experimental {
 
+class RNTupleReadOptions;
+
+namespace Internal {
+
+class RNTupleReadOptionsManip final {
+public:
+   static unsigned int GetClusterBunchSize(const RNTupleReadOptions &options);
+   static void SetClusterBunchSize(RNTupleReadOptions &options, unsigned int val);
+};
+
+} // namespace Internal
+
 // clang-format off
 /**
 \class ROOT::Experimental::RNTupleReadOptions
@@ -29,6 +41,8 @@ All page source classes need to support the common options.
 */
 // clang-format on
 class RNTupleReadOptions {
+   friend class Internal::RNTupleReadOptionsManip;
+
 public:
    enum class EClusterCache {
       kOff,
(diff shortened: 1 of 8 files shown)


silverweed (approved):

tree/ntuple/v7/inc/ROOT/RNTupleReadOptions.hxx
```diff
@@ -75,8 +75,8 @@ public:
    EImplicitMT GetUseImplicitMT() const { return fUseImplicitMT; }
    void SetUseImplicitMT(EImplicitMT val) { fUseImplicitMT = val; }
 
-   bool HasMetricsEnabled() const { return fEnableMetrics; }
-   void SetMetricsEnabled(bool enable) { fEnableMetrics = enable; }
+   bool HasMetricsEnabled() const { return fMetricsEnabled; }
+   void SetMetricsEnabled(bool val) { fMetricsEnabled = val; }
```
hahnjo, line 79: Shouldn't it be `SetEnableMetrics()` to be in line with other verb + noun styles?

jblomer: I remember we discussed this but apparently I forgot what we settled on. I thought we want to carry the information that we don't enable metrics right away but we toggle the switch that eventually results in the metrics being enabled. But I may be wrong...

pcanal: Did we consider:
```
void EnableMetrics();
void DisableMetrics();
```
?
or
```
SetRecordMetrics(bool val);
```

hahnjo: I unfortunately also don't remember exactly what we settled on, so I'll just write down some thoughts:
1. We didn't want to call it `EnableMetrics()` directly because it's called on the options and we're not actually operating on the options, but just record what we set after.
2. Instead we liked the `Set` prefix because it conveys that we remember to later call `EnableMetrics` on the reader once created.
3. Between `SetMetricsEnabled()` and `SetEnableMetrics()`, my argument for the latter is that it is already spelt this way for all write options: `SetUseBufferedWrite()`, `SetUseDirectIO()`, `SetUseImplicitMT()`, `SetEnablePageChecksums()`, `SetEnableSamePageMerging()`. We can change, but then we should probably be consistent everywhere else...

jblomer: Thanks! I think that makes sense!

hahnjo (approved):

PR #17575, opened 2025-01-30
[io] Extract tkey walk logic from TFile::Map()

silverweed:
# This Pull request:
refactors `TFile::Map` into 2 methods: `Map` and `WalkTKeys`. The latter contains the logic of traversing the TKeys in the file and returns an array with information about keys, gaps and errors. `Map` now simply calls that method and prints out the relevant information, in the same format as before.

The main advantage of splitting `WalkTKeys` is that it can be used by other places (like unit tests or client code) that are interested in the internal TKey structure.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/io/io/inc/TFile.h
+++ b/io/io/inc/TFile.h
@@ -23,6 +23,7 @@
 
 #include <atomic>
 #include <string>
+#include <cstdint>
 
 #include "Compression.h"
 #include "TDirectoryFile.h"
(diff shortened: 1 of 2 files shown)


io/io/inc/TFile.h
```diff
@@ -50,6 +51,35 @@ class TProcessID;
...
+   std::uint32_t fLen;
+
+   // these are only valid for Keys
+   Version_t fKeyVersion;
+   Int_t fObjLen;
+   Int_t fDatime;
+   Short_t fKeyLen;
+   Short_t fCycle;
+   Long64_t fSeekKey;
+   Long64_t fSeekPdir;
```
jblomer, line 73: Default initialize

jblomer (commented):
I like the idea! In this approach, all keys are loaded in memory before printing the information. Do we need a piecewise / cursor based API?

silverweed:
>  In this approach, all keys are loaded in memory before printing the information. Do we need a piecewise / cursor based API?

This is possible, although a file containing 1 million keys would only occupy 68 MiB of memory for the TKeyMapNodes (for reference, the 3.8 GB ttjet_13tev benchmark dataset has 278470 keys - about 38 MiB of memory). 
Generating the array seems to be also quite fast on my machine (<1s in debug mode)

pcanal:
The current *observed* maximum number of baskets in TTree is 50 millions baskets ... and only because it reaches the 1Gb limit for the TTree object.  It will/can grow larger once we lift the 1Gb limit and can already reach larger size with RNTuple (probably not quite as easily due to page size being larger than basket sizes).

Nonetheless that is 3.1 GiB of memory for the TKeyMapNodes .... so indeed I would recommend some sort of iterators mechanism (other-wise the code simply 'crash/out-of-memory' for large files.

io/io/inc/TFile.h
```diff
@@ -50,9 +52,86 @@ class TProcessID;
...
+   TIterator begin() const { return TIterator(fFile, 0); }
+   TIterator end() const { return TIterator(fFile, -1); }
+};
+
+} // namespace ROOT::Detail
+
 class TFile : public TDirectoryFile {
   friend class TDirectoryFile;
   friend class TFilePrefetch;
+  friend class ROOT::Detail::TKeyMapIterable::TIterator;
```
pcanal, line 134: See conversation at https://github.com/root-project/root/pull/17568/files#r1943156869

silverweed: Is there any action to take about it?

pcanal (approved):
LGTM.  Thanks.

PR #508, opened 2017-04-18
Use proper dates in the copyright notice.

dpiparo:
The copyright notice was outdated.

PR #17579, opened 2025-01-31
[RDF] Add option to change default basket size in RDataFrame Snapshot

### LINKED ISSUE #17418: Add option to change default basket size in RDataFrame Snapshot

Following request from https://root-forum.cern.ch/t/changing-basket-size-in-rdataframe-snapshot/62628

We need a new option in https://github.com/root-project/root/blob/master/tree/dataframe/inc/ROOT/RSnapshotOptions.hxx , then it should be honored in https://github.com/root-project/root/blob/99dca16a302ec1118aed015684a03ee230fc923a/tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx#L1393

### ROOT version

Any

### Installation method

Any

### Operating system

Any

### Additional context

_No response_

Aditya-138-12:
# This Pull request:
Introduces a new option in RSnapshotOptions.hxx to allow users to configure the default basket size of new branches when using RDataFrame::Snapshot. Modify SetBranchesHelper in ActionHelpers.hxx to honor this option.

## Changes or fixes:
1. Add a fBasketSize field to RSnapshotOptions to store the user-defined basket size.
2. Update SetBranchesHelper to apply the specified basket size when creating output.
3. Ensure backward compatibility by maintaining the current default behavior if no custom size is set.
:)

## Checklist:

- [✅] tested changes locally
- [❌] updated the docs (if necessary)

This PR fixes #17418

### DIFF
--- a/tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
@@ -1335,7 +1335,7 @@ void *GetData(T & /*v*/)
 template <typename T>
 void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &inName, const std::string &name,
                        TBranch *&branch, void *&branchAddress, T *address, RBranchSet &outputBranches,
-                       bool /*isDefine*/)
+                       bool /*isDefine*/, const std::optional<int> &basketSize)
 {
    static TClassRef TBOClRef("TBranchObject");
 
@@ -1366,7 +1366,10 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
       // In particular, by keeping splitlevel equal to 0 if this was the case for `inputBranch`, we avoid
       // writing garbage when unsplit objects cannot be written as split objects (e.g. in case of a polymorphic
       // TObject branch, see https://bit.ly/2EjLMId ).
-      const auto bufSize = inputBranch->GetBasketSize();
+      const auto bufSize = basketSize.has_value()
+                              ? basketSize.value()
+                              : inputBranch->GetBasketSize(); // Use original basket size for existing branches
+                                                              // otherwise custom basket size for new branches
       const auto splitLevel = inputBranch->GetSplitLevel();
 
       if (inputBranch->IsA() == TBOClRef) {
@@ -1377,7 +1380,10 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
          outputBranch = outputTree.Branch(name.c_str(), address, bufSize, splitLevel);
       }
    } else {
-      outputBranch = outputTree.Branch(name.c_str(), address);
+      // Set Custom basket size for new branches.
+      const auto buffSize =
+         basketSize.has_value() ? basketSize.value() : (inputBranch ? inputBranch->GetBasketSize() : 32000);
+      outputBranch = outputTree.Branch(name.c_str(), address, buffSize);
    }
    outputBranches.Insert(name, outputBranch);
    // This is not an array branch, so we don't register the address of the output branch here
@@ -1397,7 +1403,8 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
 /// `branchAddress`) so we can intercept changes in the address of the input branch and tell the output branch.
 template <typename T>
 void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &inName, const std::string &outName,
-                       TBranch *&branch, void *&branchAddress, RVec<T> *ab, RBranchSet &outputBranches, bool isDefine)
+                       TBranch *&branch, void *&branchAddress, RVec<T> *ab, RBranchSet &outputBranches, bool isDefine,
+                       const std::optional<int> &basketSize)
 {
    TBranch *inputBranch = nullptr;
    if (inputTree) {
@@ -1434,7 +1441,10 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
          // needs to be SetObject (not SetAddress) to mimic what happens when this TBranchElement is constructed
          outputBranch->SetObject(ab);
       } else {
-         auto *b = outputTree.Branch(outName.c_str(), ab);
+         // Set Custom basket size for new branches if specified, otherwise get basket size from input branches
+         const auto buffSize =
+            basketSize.has_value() ? basketSize.value() : (inputBranch ? inputBranch->GetBasketSize() : 32000);
+         auto *b = outputTree.Branch(outName.c_str(), ab, buffSize);
          outputBranches.Insert(outName, b);
       }
       return;
@@ -1462,7 +1472,8 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
          // added to the output tree yet. However, the size leaf has to be available for the creation of the array
          // branch to be successful. So we create the size leaf here.
          const auto sizeTypeStr = TypeName2ROOTTypeName(sizeLeaf->GetTypeName());
-         const auto sizeBufSize = sizeLeaf->GetBranch()->GetBasketSize();
+         // Use Original basket size for Existing Branches otherwise use Custom basket Size.
+         const auto sizeBufSize = basketSize.has_value() ? basketSize.value() : sizeLeaf->GetBranch()->GetBasketSize();
          // The null branch address is a placeholder. It will be set when SetBranchesHelper is called for `sizeLeafName`
          auto *sizeBranch = outputTree.Branch(sizeLeafName.c_str(), (void *)nullptr,
                                               (sizeLeafName + '/' + sizeTypeStr).c_str(), sizeBufSize);
@@ -1478,7 +1489,9 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
                  bname);
       } else {
          const auto leaflist = std::string(bname) + "[" + sizeLeafName + "]/" + rootbtype;
-         outputBranch = outputTree.Branch(outName.c_str(), dataPtr, leaflist.c_str());
+         // Use original basket size for existing branches and new basket size for new branches
+         const auto branchBufSize = basketSize.has_value() ? basketSize.value() : inputBranch->GetBasketSize();
+         outputBranch = outputTree.Branch(outName.c_str(), dataPtr, leaflist.c_str(), branchBufSize);
          outputBranch->SetTitle(inputBranch->GetTitle());
          outputBranches.Insert(outName, outputBranch);
          branch = outputBranch;
@@ -1577,10 +1590,11 @@ public:
    void SetBranches(ColTypes &... values, std::index_sequence<S...> /*dummy*/)
    {
       // create branches in output tree
-      int expander[] = {(SetBranchesHelper(fInputTree, *fOutputTree, fInputBranchNames[S], fOutputBranchNames[S],
-                                           fBranches[S], fBranchAddresses[S], &values, fOutputBranches, fIsDefine[S]),
-                         0)...,
-                        0};
+      int expander[] = {
+         (SetBranchesHelper(fInputTree, *fOutputTree, fInputBranchNames[S], fOutputBranchNames[S], fBranches[S],
+                            fBranchAddresses[S], &values, fOutputBranches, fIsDefine[S], fOptions.fBasketSize),
+          0)...,
+         0};
       fOutputBranches.AssertNoNullBranchAddresses();
       (void)expander; // avoid unused variable warnings for older compilers such as gcc 4.9
    }
@@ -1779,7 +1793,7 @@ public:
       // hack to call TTree::Branch on all variadic template arguments
       int expander[] = {(SetBranchesHelper(fInputTrees[slot], *fOutputTrees[slot], fInputBranchNames[S],
                                            fOutputBranchNames[S], fBranches[slot][S], fBranchAddresses[slot][S],
-                                           &values, fOutputBranches[slot], fIsDefine[S]),
+                                           &values, fOutputBranches[slot], fIsDefine[S], fOptions.fBasketSize),
                          0)...,
                         0};
       fOutputBranches[slot].AssertNoNullBranchAddresses();
--- a/tree/dataframe/inc/ROOT/RSnapshotOptions.hxx
+++ b/tree/dataframe/inc/ROOT/RSnapshotOptions.hxx
@@ -12,6 +12,7 @@
 #define ROOT_RSNAPSHOTOPTIONS
 
 #include <Compression.h>
+#include <optional>
 #include <string_view>
 #include <string>
 
@@ -25,7 +26,8 @@ struct RSnapshotOptions {
    RSnapshotOptions(const RSnapshotOptions &) = default;
    RSnapshotOptions(RSnapshotOptions &&) = default;
    RSnapshotOptions(std::string_view mode, ECAlgo comprAlgo, int comprLevel, int autoFlush, int splitLevel, bool lazy,
-                    bool overwriteIfExists = false, bool vector2RVec = true)
+                    bool overwriteIfExists = false, bool vector2RVec = true,
+                    const std::optional<int> &basketSize = std::nullopt)
       : fMode(mode),
         fCompressionAlgorithm(comprAlgo),
         fCompressionLevel{comprLevel},
@@ -33,7 +35,8 @@ struct RSnapshotOptions {
         fSplitLevel(splitLevel),
         fLazy(lazy),
         fOverwriteIfExists(overwriteIfExists),
-        fVector2RVec(vector2RVec)
+        fVector2RVec(vector2RVec),
+        fBasketSize(basketSize)
    {
    }
    std::string fMode = "RECREATE"; ///< Mode of creation of output file
@@ -45,6 +48,8 @@ struct RSnapshotOptions {
    bool fLazy = false;                              ///< Do not start the event loop when Snapshot is called
    bool fOverwriteIfExists = false; ///< If fMode is "UPDATE", overwrite object in output file if it already exists
    bool fVector2RVec = true;        ///< If set to true will convert std::vector columns to RVec when saving to disk
+   std::optional<int> fBasketSize{}; ///< Set a custom basket size option. For more details, see
+                                     ///< https://root.cern/manual/trees/#baskets-clusters-and-the-tree-header
 };
 } // namespace RDF
 } // namespace ROOT
--- a/tree/dataframe/test/dataframe_snapshot.cxx
+++ b/tree/dataframe/test/dataframe_snapshot.cxx
@@ -61,6 +61,187 @@ protected:
 };
 #endif // R__USE_IMT
 
+// Test for custom basket size in Snapshot
+class SnapshotCustomBasketRAII {
+private:
+   std::string fInputFile;
+   std::string fOutputFileCustom;
+   std::string fOutputFileCollection;
+
+public:
+   SnapshotCustomBasketRAII()
+      : fInputFile("input_file.root"),
+        fOutputFileCustom("output_file_custom_basket.root"),
+        fOutputFileCollection("output_file_collection_basket.root")
+   {
+
+      // Create file and tree inside the constructor body
+      TFile fFile(fInputFile.c_str(), "RECREATE");
+      TTree fTree("tree", "Test Tree");
+
+      // Create scalar branch
+      float value = 0;
+      fTree.Branch("branch_x", &value);
+
+      // Create vector branch
+      std::vector<float> vec_values;
+      fTree.Branch("branch_vec", &vec_values);
+
+      // Fill the tree with some data
+      for (int i = 0; i < 1000; ++i) {
+         value = i;
+         vec_values.clear();
+         // Add some random number of elements to the vector
+         int vec_size = i % 10 + 1; // 1 to 10 elements
+         for (int j = 0; j < vec_size; ++j) {
+            vec_values.push_back(i + j * 0.1f);
+         }
+         fTree.Fill();
+      }
+      fFile.Write();
+   }
+
+   ~SnapshotCustomBasketRAII()
+   {
+      gSystem->Unlink(fInputFile.c_str());
+      gSystem->Unlink(fOutputFileCustom.c_str());
+      gSystem->Unlink(fOutputFileCollection.c_str());
+   }
+
+   const std::string &GetInputFile() const { return fInputFile; }
+   const std::string &GetOutputFileCustom() const { return fOutputFileCustom; }
+   const std::string &GetOutputFileCollection() const { return fOutputFileCollection; }
+};
+
+void TestCustomBasketSize()
+{
+   SnapshotCustomBasketRAII raii;
+
+   ROOT::RDataFrame df("tree", raii.GetInputFile());
+
+   auto df_with_new_columns = df.Define("branch_x_new", [](float x) { return x * 2; }, {"branch_x"})
+                                 .Define("branch_vec_new",
+                                         [](const std::vector<float> &vec) {
+                                            std::vector<float> result;
+                                            result.reserve(vec.size());
+                                            for (auto v : vec)
+                                               result.push_back(v * 2);
+                                            return result;
+                                         },
+                                         {"branch_vec"});
+
+   ROOT::RDF::RSnapshotOptions options;
+   options.fBasketSize = 2048;
+
+   df_with_new_columns.Snapshot<float, float>("tree", raii.GetOutputFileCustom(), {"branch_x", "branch_x_new"},
+                                              options);
+
+   df_with_new_columns.Snapshot<std::vector<float>, std::vector<float>>("tree", raii.GetOutputFileCollection(),
+                                                                        {"branch_vec", "branch_vec_new"}, options);
+
+   TFile output_file_custom(raii.GetOutputFileCustom().c_str());
+   auto output_tree_custom = output_file_custom.Get<TTree>("tree");
+
+   for (auto b : TRangeDynCast<TBranch>(*output_tree_custom->GetListOfBranches())) {
+      ASSERT_TRUE(b != nullptr);
+      EXPECT_EQ(b->GetBasketSize(), 2048) << "Incorrect basket size for scalar branch " << b->GetName();
+   }
+
+   TFile output_file_collection(raii.GetOutputFileCollection().c_str());
+   auto output_tree_collection = output_file_collection.Get<TTree>("tree");
+
+   for (auto b : TRangeDynCast<TBranch>(*output_tree_collection->GetListOfBranches())) {
+      ASSERT_TRUE(b != nullptr);
+      EXPECT_EQ(b->GetBasketSize(), 2048) << "Incorrect basket size for vector branch " << b->GetName();
+   }
+}
+
+void TestDefaultBasketSize()
+{
+   SnapshotCustomBasketRAII helper;
+   const Int_t defaultBasketSize = 32000;
+
+   ROOT::RDataFrame df("tree", helper.GetInputFile());
+
+   df.Snapshot("tree", helper.GetOutputFileCustom());
+
+   std::unique_ptr<TFile> f(TFile::Open(helper.GetOutputFileCustom().c_str()));
+   ASSERT_TRUE(f != nullptr);
+
+   auto tree = f->Get<TTree>("tree");
+   ASSERT_TRUE(tree != nullptr);
+
+   auto branchX = tree->GetBranch("branch_x");
+   ASSERT_TRUE(branchX != nullptr);
+   EXPECT_EQ(branchX->GetBasketSize(), defaultBasketSize) << "Scalar branch doesn't have default basket size";
+
+   auto branchVec = tree->GetBranch("branch_vec");
+   ASSERT_TRUE(branchVec != nullptr);
+   EXPECT_EQ(branchVec->GetBasketSize(), defaultBasketSize) << "Vector branch doesn't have default basket size";
+}
+
+void TestBasketSizePreservation()
+{
+   SnapshotCustomBasketRAII helper;
+
+   // Define the columns we want to snapshot
+   const std::vector<std::string> columns = {"branch_x", "branch_vec"};
+
+   // First create a tree with custom basket size
+   {
+      ROOT::RDataFrame df("tree", helper.GetInputFile());
+      ROOT::RDF::RSnapshotOptions options;
+      options.fBasketSize = 64000; // 64KB
+      df.Snapshot<float, std::vector<float>>("tree", helper.GetOutputFileCustom(), columns, options);
+   }
+
+   // Now read that tree and create new snapshot without specifying basket size
+   {
+      ROOT::RDataFrame df("tree", helper.GetOutputFileCustom());
+      df.Snapshot<float, std::vector<float>>("tree", helper.GetOutputFileCollection(), columns);
+   }
+
+   // Open both files and compare basket sizes
+   std::unique_ptr<TFile> f1(TFile::Open(helper.GetOutputFileCustom().c_str()));
+   std::unique_ptr<TFile> f2(TFile::Open(helper.GetOutputFileCollection().c_str()));
+   ASSERT_TRUE(f1 != nullptr);
+   ASSERT_TRUE(f2 != nullptr);
+
+   auto tree1 = f1->Get<TTree>("tree");
+   auto tree2 = f2->Get<TTree>("tree");
+   ASSERT_TRUE(tree1 != nullptr);
+   ASSERT_TRUE(tree2 != nullptr);
+
+   // Check both branches
+   for (const auto &branchName : columns) {
+      auto branch1 = tree1->GetBranch(branchName.c_str());
+      auto branch2 = tree2->GetBranch(branchName.c_str());
+      ASSERT_TRUE(branch1 != nullptr);
+      ASSERT_TRUE(branch2 != nullptr);
+
+      EXPECT_EQ(branch2->GetBasketSize(), branch1->GetBasketSize())
+         << "Branch '" << branchName << "' basket size not preserved";
+   }
+}
+
+// Test for custom basket size
+TEST(RDFSnapshotMore, CustomBasketSize)
+{
+   TestCustomBasketSize();
+}
+
+// Test for default basket size
+TEST(RDFSnapshotMore, DefaultBasketSize)
+{
+   TestDefaultBasketSize();
+}
+
+// Test for basket size preservation
+TEST(RDFSnapshotMore, BasketSizePreservation)
+{
+   TestBasketSizePreservation();
+}
+
 // fixture that provides fixed and variable sized arrays as RDF columns
 class RDFSnapshotArrays : public ::testing::Test {
 protected:
@@ -1453,6 +1634,29 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
    // TTree "t" should *not* be in there, differently from the single-thread case: see ROOT-10868
    EXPECT_NE(t, nullptr);
    gSystem->Unlink(fname);
+}
+
+TEST(RDFSnapshotMore, CustomBasketSizeMT)
+{
+   ROOT::EnableImplicitMT();
+   TestCustomBasketSize();
+   ROOT::DisableImplicitMT();
+}
+
+// Test for default basket size
+TEST(RDFSnapshotMore, DefaultBasketSizeMT)
+{
+   ROOT::EnableImplicitMT();
+   TestDefaultBasketSize();
+   ROOT::DisableImplicitMT();
+}
+
+// Test for basket size preservation
+TEST(RDFSnapshotMore, BasketSizePreservationMT)
+{
+   ROOT::EnableImplicitMT();
+   TestBasketSizePreservation();
+   ROOT::DisableImplicitMT();
 }
 
 #endif // R__USE_IMT


tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1397,7 +1397,7 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
 /// `branchAddress`) so we can intercept changes in the address of the input branch and tell the output branch.
 template <typename T>
 void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &inName, const std::string &outName,
-                       TBranch *&branch, void *&branchAddress, RVec<T> *ab, RBranchSet &outputBranches, bool isDefine)
+                       TBranch *&branch, void *&branchAddress, RVec<T> *ab, RBranchSet &outputBranches, bool isDefine, const RSnapshotOptions &options)
```
vepadulano, line 1400: Probably we can just pass the `basketSize` parameter directly without passing the entire options object.

vepadulano, line 1400: Also be careful, there is one more template for `SetBranchesHelper`, starting at line 1330. That should be adjusted as well.

vepadulano (changes requested):
Thank you so much @Aditya-138-12 for this contribution! I feel like we're already going in the right direction. I have a few introductory inline comments. Together with those, I would highly apppreciate if you could add a test for the new feature, checking that the output TTree after a Snapshot has the requested basket size. You can take inspiration from some of the tests already written in e.g. `dataframe_snapshot.cxx`.

tree/dataframe/inc/ROOT/RSnapshotOptions.hxx
```diff
@@ -45,6 +46,7 @@ struct RSnapshotOptions {
    bool fLazy = false;                              ///< Do not start the event loop when Snapshot is called
    bool fOverwriteIfExists = false; ///< If fMode is "UPDATE", overwrite object in output file if it already exists
    bool fVector2RVec = true;        ///< If set to true will convert std::vector columns to RVec when saving to disk
+   int fBasketSize = 32000;         /// Custom Basket Size option
```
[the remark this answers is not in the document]
Aditya-138-12: Yeah it would be better! I will update the code to use `std::optional<int>`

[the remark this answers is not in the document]
Aditya-138-12: Done!

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
Aditya-138-12: Done!

Aditya-138-12: Done!

tree/dataframe/inc/ROOT/RSnapshotOptions.hxx
```diff
@@ -25,15 +26,16 @@ struct RSnapshotOptions {
    RSnapshotOptions(const RSnapshotOptions &) = default;
    RSnapshotOptions(RSnapshotOptions &&) = default;
    RSnapshotOptions(std::string_view mode, ECAlgo comprAlgo, int comprLevel, int autoFlush, int splitLevel, bool lazy,
-                    bool overwriteIfExists = false, bool vector2RVec = true)
+                    bool overwriteIfExists = false, bool vector2RVec = true, std::optional<int> basketSize = 32000)
```
vepadulano, line 29:
```suggestion
                    bool overwriteIfExists = false, bool vector2RVec = true, const std::optional<int> &basketSize = std::nullopt)
```

tree/dataframe/inc/ROOT/RSnapshotOptions.hxx
```diff
@@ -45,6 +47,7 @@ struct RSnapshotOptions {
    bool fLazy = false;                              ///< Do not start the event loop when Snapshot is called
    bool fOverwriteIfExists = false; ///< If fMode is "UPDATE", overwrite object in output file if it already exists
    bool fVector2RVec = true;        ///< If set to true will convert std::vector columns to RVec when saving to disk
+   std::optional<int> fBasketSize = 32000;         /// Custom Basket Size option, for more details, see https://root.cern/manual/trees/#baskets-clusters-and-the-tree-header
```
vepadulano, line 50:
```suggestion
   std::optional<int> fBasketSize{};         /// Set a custom basket size. For more details, see https://root.cern/manual/trees/#baskets-clusters-and-the-tree-header
```

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1397,7 +1398,7 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
 /// `branchAddress`) so we can intercept changes in the address of the input branch and tell the output branch.
 template <typename T>
 void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &inName, const std::string &outName,
-                       TBranch *&branch, void *&branchAddress, RVec<T> *ab, RBranchSet &outputBranches, bool isDefine)
+                       TBranch *&branch, void *&branchAddress, RVec<T> *ab, RBranchSet &outputBranches, bool isDefine, const std::optional<int> basketSize)
```
vepadulano, line 1401:
```suggestion
                       TBranch *&branch, void *&branchAddress, RVec<T> *ab, RBranchSet &outputBranches, bool isDefine, const std::optional<int> &basketSize)
```

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1435,6 +1438,10 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
          outputBranch->SetObject(ab);
       } else {
          auto *b = outputTree.Branch(outName.c_str(), ab);
+         // Set Custom basket size for new branches.
+         if(isNewBranch && basketSize.value() > 0){
```
vepadulano, line 1442:
```suggestion
         if(isNewBranch && basketSize.has_value()){
```

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1478,7 +1486,9 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
                  bname);
       } else {
          const auto leaflist = std::string(bname) + "[" + sizeLeafName + "]/" + rootbtype;
-         outputBranch = outputTree.Branch(outName.c_str(), dataPtr, leaflist.c_str());
+         //Use original basket size for existing branches and new basket size for new branches
+         const auto branchBufSize = isNewBranch && basketSize.value() > 0 ? basketSize.value() : inputBranch->GetBasketSize();
```
vepadulano, line 1490:
```suggestion
         const auto branchBufSize = isNewBranch && basketSize.has_value() ? basketSize.value() : inputBranch->GetBasketSize();
```

vepadulano (changes requested):

Aditya-138-12:
> Thanks for the prompt interaction! We are still missing:
> 
> * Adapting the other overload of `SetBranchesHelper`
> * Some tests

Thank you so much for your guidance! I really appreciate your time and feedback. I’m currently working on the tests, but I wanted to make sure I fully understand your point about 'adapting the other overload of SetBranchesHelper.' Could you kindly clarify what exactly needs to be adapted? I’d be grateful for any additional details!

vepadulano:
> Could you kindly clarify what exactly needs to be adapted?

There is another function template of `SetBranchesHelper` starting at line 1336 of `ActionHelpers.hxx`. Currently in this PR you are only modifying the overload responsible for writing out branches storing collections instead, which is the one starting at 1399.

tree/dataframe/inc/ROOT/RSnapshotOptions.hxx
```diff
@@ -45,6 +47,7 @@ struct RSnapshotOptions {
    bool fLazy = false;                              ///< Do not start the event loop when Snapshot is called
    bool fOverwriteIfExists = false; ///< If fMode is "UPDATE", overwrite object in output file if it already exists
    bool fVector2RVec = true;        ///< If set to true will convert std::vector columns to RVec when saving to disk
+   std::optional<int> fBasketSize {};         /// Set a custom basket size option. For more details, see https://root.cern/manual/trees/#baskets-clusters-and-the-tree-header
```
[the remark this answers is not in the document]
Aditya-138-12: Done!

dpiparo:
Let's test this!

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1335,7 +1336,7 @@ void *GetData(T & /*v*/)
 template <typename T>
 void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &inName, const std::string &name,
                        TBranch *&branch, void *&branchAddress, T *address, RBranchSet &outputBranches,
-                       bool /*isDefine*/)
+                       bool /*isDefine*/, const std::optional<int> &basketSize)
```
vepadulano, line 1339: You have just added the function parameter, but not using it anywhere in the function. Also here, you should make use of this parameter to ensure that new branches are created with the correct basket size

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1455,67 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
    gSystem->Unlink(fname);
 }
 
+// Test for custom basket size in Snapshot, ROOT - #17418
+TEST(RDFSnapshotMore, CustomBasketSize)
+{
+    // Create a simple ROOT file with a TTree
+    auto input_file = TFile::Open("input_file.root", "RECREATE");
+    auto tree = new TTree("tree", "Test Tree");
```
vepadulano, line 1463: There is a `new` keyword without the corresponding `delete`. This is wrong and in general `new` should be avoided at all costs, prefer stack-based variables or constructs that automatically manage memory such as `std::unique_ptr`

vepadulano, line 1463: Do not create the dataset like this. Prefer RAII struct such as the ones you can find in many RDF tests. An example is https://github.com/root-project/root/blob/8b3840f73be1d892334ba7df6d2867cad470c196/tree/dataframe/test/dataframe_vary.cxx#L1626

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1455,67 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
...
+    tree->Branch("branch_x", &value);
+
+    // Fill the tree with some data
+    for (int i = 0; i < 1000; ++i) {
+        value = i;
+        tree->Fill();
+    }
+
+    input_file->Write();
+    input_file->Close();
```
vepadulano, line 1475: This should end up in the RAII constructor

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1455,67 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
...
+    for (auto branch : *output_tree_default->GetListOfBranches()) {
+        EXPECT_NE(branch->GetBasketSize(), 2048); // Verify the default basket size is not 2048
+    }
+
+    output_file_default_tfile->Close();
+
+    // Clean up
+    gSystem->Unlink("input_file.root");
+    gSystem->Unlink(output_file_custom.c_str());
+    gSystem->Unlink(output_file_default.c_str());
```
vepadulano, line 1518: This should end up in the RAII destructor

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1455,67 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
...
+    // Define a new column
+    auto df_with_new_column = df.Define("branch_x_new", [](float x) { return x; }, {"branch_x"});
+
+    // Create RSnapshotOptions with a custom basket size
+    ROOT::RDF::RSnapshotOptions options;
+    options.fBasketSize = 2048; // Set custom basket size (2KB)
+
+    // Snapshot with custom basket size
+    std::string output_file_custom = "output_file_custom_basket.root";
+    df_with_new_column.Snapshot("tree", output_file_custom, {"branch_x", "branch_x_new"}, options);
```
vepadulano, line 1489: This will for sure not work as you expect. All the branches involved in this computation graph are of type `float`. You are not modifying anything in `ActionHelpers.hxx` to act upon branches of this type. This is the other function template I was referring to. Currently you are just passing the `basketSize` as parameter, you need to actually make use of it in the function.

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1455,67 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
...
+
+    // Check the basket size in the output file
+    auto output_file_custom_tfile = TFile::Open(output_file_custom.c_str());
+    auto output_tree_custom = output_file_custom_tfile->Get<TTree>("tree");
+
+    for (auto branch : *output_tree_custom->GetListOfBranches()) {
+        EXPECT_EQ(branch->GetBasketSize(), 2048); // Verify the custom basket size
```
vepadulano, line 1496: `GetListOfBranches` is not returning a collection of `TBranch` objects, you need to cast it. See various places using `TRangeDynCast` for the right thing to do here. For example https://github.com/root-project/root/blob/8b3840f73be1d892334ba7df6d2867cad470c196/tree/ntupleutil/v7/src/RNTupleImporter.cxx#L168

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1455,67 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
    gSystem->Unlink(fname);
 }
 
+// Test for custom basket size in Snapshot, ROOT - #17418
+TEST(RDFSnapshotMore, CustomBasketSize)
```
vepadulano, line 1459: The test is only testing one branch of type `float`, i.e. a scalar. We need at least another test for collection types

vepadulano (changes requested):

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
Aditya-138-12: Done!

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1455,108 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
...
+
+public:
+   SnapshotCustomBasketRAII() : 
+      fInputFile("input_file.root"),
+      fOutputFileCustom("output_file_custom_basket.root"),
+      fOutputFileCollection("output_file_collection_basket.root")
+   {
+      // Create input file and tree
+      fFile.reset(TFile::Open(fInputFile.c_str(), "RECREATE"));
+      fTree.reset(new TTree("tree", "Test Tree"));
```
vepadulano, line 1475: No need to use `unique_ptr` in this case. Just create the TFile and TTree on the stack in the constructor body.

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1455,108 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
...
+         }
+         fTree->Fill();
+      }
+      fFile->Write();
+   }
+
+   ~SnapshotCustomBasketRAII() {
+      if (fFile) {
+         fFile->Close();
+      }
```
vepadulano, line 1502: If you create the TFile on the stack in the constructor there will be no need for this extra logic

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1455,108 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
...
+      gSystem->Unlink(fOutputFileCustom.c_str());
+      gSystem->Unlink(fOutputFileCollection.c_str());
+   }
+
+   std::string GetInputFile() const { return fInputFile; }
+   std::string GetOutputFileCustom() const { return fOutputFileCustom; }
+   std::string GetOutputFileCollection() const { return fOutputFileCollection; }
+};
+
+TEST_F(RDFSnapshotMore, CustomBasketSize)
```
vepadulano, line 1513: The `TEST_F` macro is not appropriate in this context, `TEST` will be enough

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1455,108 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
...
+      fFile.reset(TFile::Open(fInputFile.c_str(), "RECREATE"));
+      fTree.reset(new TTree("tree", "Test Tree"));
+
+      // Create scalar branch
+      float value = 0;
+      fTree->Branch("branch_x", &value);
+
+      // Create vector branch
+      std::vector<float> vec_values;
+      auto vec_branch = fTree->Branch("branch_vec", &vec_values);
```
vepadulano, line 1483: No need to store the result of `Branch` in a variable here

vepadulano (changes requested):

tree/dataframe/test/dataframe_snapshot.cxx
Aditya-138-12: Should we use `RDFSnapshotMT` instead of `RDFSnapshotMore`, because we are getting a error
---
ubuntu2404 LLVM_ENABLE_ASSERTIONS=On, CMAKE_BUILD_TYPE=Debug: tree/dataframe/test/dataframe_snapshot.cxx#L1513
‘RDFSnapshotMore’ was not declared in this scope; did you mean ‘RDFSnapshotMT’?

vepadulano: It's slightly more complicated than that. I believe it's fine to use `RDFSnapshotMore`, but the area of the file where you are writing this test is guarded by `#ifdef R__USE_IMT` starting at line 1056. That means this area contains tests that should use the multithreaded capabilities of ROOT. So we need to extra steps:
1. You can keep the test as is now, but you need to add `ROOT::EnableImplicitMT();` at the beginning of the test body and `ROOT::DisableImplicitMT();` at the end. See for example the `ReadWriteCarrayMT` test.
2. You then should copy-paste this test also in the area of the file *before* the `ifdef`. In that other test, you should not write `Enable/DisableImplicitMT`. Since the test will be the same, I suggest you write everything as a separate free-function, then call that free function in both tests.

Aditya-138-12: Got it!! :)

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1460,100 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
    gSystem->Unlink(fname);
 }
 
+// Test for custom basket size in Snapshot
```
vepadulano, line 1463: Both the RAII class and the free function should be placed before the first TEST, so you get proper definition of everything needed

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1460,100 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
...
+    TFile output_file_collection(raii.GetOutputFileCollection().c_str());
+    auto output_tree_collection = output_file_collection.Get<TTree>("tree");
+    
+    for (auto b : TRangeDynCast<TBranch>(*output_tree_collection->GetListOfBranches())) {
+        ASSERT_TRUE(b != nullptr);
+        EXPECT_EQ(b->GetBasketSize(), 2048) << "Incorrect basket size for collection branch " << b->GetName();
+    }
+}
+
+TEST(RDFSnapshotMore, CustomBasketSize){
```
vepadulano, line 1553: This test needs a different name, e.g. `CustomBasketSizeMT`.

vepadulano (changes requested):

tree/dataframe/test/dataframe_snapshot.cxx
Aditya-138-12: Done! 
Why do we need a different name?

tree/dataframe/test/dataframe_snapshot.cxx
Aditya-138-12: Done!

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -61,6 +61,102 @@ class RDFSnapshotMT : public ::testing::Test {
...
 
+
+// Test for custom basket size in Snapshot
+class SnapshotCustomBasketRAII {
+private:
+   std::string fInputFile;
+   std::string fOutputFileCustom;
+   std::string fOutputFileCollection;
+   TFile fFile;
+   TTree fTree;
```
vepadulano, line 72: Please move these data members inside the constructor body directly, no need to keep them around for the entire lifetime of the class instance

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -61,6 +61,102 @@ class RDFSnapshotMT : public ::testing::Test {
...
+
+   ~SnapshotCustomBasketRAII() {
+      gSystem->Unlink(fInputFile.c_str());
+      gSystem->Unlink(fOutputFileCustom.c_str());
+      gSystem->Unlink(fOutputFileCollection.c_str());
+   }
+
+   std::string GetInputFile() const { return fInputFile; }
+   std::string GetOutputFileCustom() const { return fOutputFileCustom; }
+   std::string GetOutputFileCollection() const { return fOutputFileCollection; }
```
vepadulano, line 112:
```suggestion
   const std::string &GetInputFile() const { return fInputFile; }
   const std::string &GetOutputFileCustom() const { return fOutputFileCustom; }
   const std::string &GetOutputFileCollection() const { return fOutputFileCollection; }
```

vepadulano (changes requested):

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -61,6 +61,102 @@ class RDFSnapshotMT : public ::testing::Test {
...
+    auto output_tree_collection = output_file_collection.Get<TTree>("tree");
+    
+    for (auto b : TRangeDynCast<TBranch>(*output_tree_collection->GetListOfBranches())) {
+        ASSERT_TRUE(b != nullptr);
+        EXPECT_EQ(b->GetBasketSize(), 2048) << "Incorrect basket size for collection branch " << b->GetName();
+    }
+}
+
+// Test for custom basket size
+TEST(RDFSnapshotMore, CustomBasketSizeMT){
```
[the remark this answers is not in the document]
Aditya-138-12: Done!

tree/dataframe/test/dataframe_snapshot.cxx
vepadulano: In this case it's mostly to ensure smoother lifetime management. There is really no need to keep the file open for the entire lifetime of the class instance, it would also possibly lead to erroneous interactions with the `gSystem->Unlink` calls in the destructor.

vepadulano:
Thanks for all the effort so far! It looks like we can start testing the changes now, let's see what the CI says about them.

vepadulano:
Hi @Aditya-138-12 ,

The CI job on Debian 12 has already finished and reports a failure for your newly introduced test, see [here](https://github.com/root-project/root/actions/runs/13117414753/job/36595788881?pr=17579):

```
   978/2625 Test  #508: gtest-tree-dataframe-dataframe-snapshot ...........................................................***Failed   46.62 sec
  Running main() from ./googletest/src/gtest_main.cc
  [==========] Running 48 tests from 4 test suites.
  [----------] Global test environment set-up.
  [----------] 30 tests from RDFSnapshotMore
  [ RUN      ] RDFSnapshotMore.CustomBasketSize
  /github/home/ROOT-CI/src/tree/dataframe/test/dataframe_snapshot.cxx:144: Failure
    b->GetBasketSize()
      Which is: 32000
    2048
  Incorrect basket size for scalar branch branch_x
  /github/home/ROOT-CI/src/tree/dataframe/test/dataframe_snapshot.cxx:152: Failure
    b->GetBasketSize()
      Which is: 32000
    2048
```

Can you please rerun the tests locally and make sure they pass before submitting the next commit? If you haven't done that yet, there are instructions [here](https://root.cern/for_developers/run_the_tests). You can use the syntax `ctest -R dataframe` to restrict your run to dataframe tests alone, or `gtest-tree-dataframe-dataframe-snapshot` to run only the `dataframe_snapshot.cxx` tests.

Aditya-138-12:
> Hi @Aditya-138-12 ,
> 
> The CI job on Debian 12 has already finished and reports a failure for your newly introduced test, see [here](https://github.com/root-project/root/actions/runs/13117414753/job/36595788881?pr=17579):
> 
> ```
>    978/2625 Test  #508: gtest-tree-dataframe-dataframe-snapshot ...........................................................***Failed   46.62 sec
>   Running main() from ./googletest/src/gtest_main.cc
>   [==========] Running 48 tests from 4 test suites.
>   [----------] Global test environment set-up.
>   [----------] 30 tests from RDFSnapshotMore
>   [ RUN      ] RDFSnapshotMore.CustomBasketSize
>   /github/home/ROOT-CI/src/tree/dataframe/test/dataframe_snapshot.cxx:144: Failure
>   Expected equality of these values:
>     b->GetBasketSize()
>       Which is: 32000
>     2048
>   Incorrect basket size for scalar branch branch_x
>   /github/home/ROOT-CI/src/tree/dataframe/test/dataframe_snapshot.cxx:152: Failure
>   Expected equality of these values:
>     b->GetBasketSize()
>       Which is: 32000
>     2048
> ```
> 
> Can you please rerun the tests locally and make sure they pass before submitting the next commit? If you haven't done that yet, there are instructions [here](https://root.cern/for_developers/run_the_tests). You can use the syntax `ctest -R dataframe` to restrict your run to dataframe tests alone, or `gtest-tree-dataframe-dataframe-snapshot` to run only the `dataframe_snapshot.cxx` tests.

Just to clarify, do you think there is a need to update any of the `project documentation` or perhaps add a `tutorial` to help users understand how to use this new functionality effectively? 
I will run the tests locally before making the next commit.

vepadulano:
> Just to clarify, do you think there is a need to update any of the project documentation or perhaps add a tutorial to help users understand how to use this new functionality effectively

We don't have specific examples for each and every option that Snapshot accepts. We rely on the documentation of RSnapshotOptions, which you have already modified inline in this PR. I believe that's fine.

> Also the clang-tools code analysis / clang-format is failing, should we worry about it?

Yes, you should apply the changes as suggested. You can run locally `git-clang-format` on your branch to apply the changes automatically.

vepadulano:
Ok the test is passing now, very nice! Although reading your last commit made me rethink about the functionality. Isn't it odd that by specifying the basket size option in `RDFSnapshotOptions` only new branches are changed? I believe if a user wants to set the basket size they want it to be set for all branches of the output tree, not just the new ones created via `Define`.

Aditya-138-12:
> Ok the test is passing now, very nice! Although reading your last commit made me rethink about the functionality. Isn't it odd that by specifying the basket size option in `RDFSnapshotOptions` only new branches are changed? I believe if a user wants to set the basket size they want it to be set for all branches of the output tree, not just the new ones created via `Define`.

Yeah, from the starting i had this in mind that why can't the user set the basket size for all the branches, Although i have tested it locally, the branch_x custom basket size can also be fixed....

I was also trying to make a new option for setting the custom basket size for other branches, how should we go about this.

vepadulano:
> Yeah, from the starting i had this in mind that why can't the user set the basket size for all the branches, Although i have tested it locally, the branch_x custom basket size can also be fixed....
I was also trying to make a new option for setting the custom basket size for other branches, how should we go about this?
Maybe removing the isDefine bool variable will do it?

I believe just setting all branches with the custom size indicated by the user is the way to go, irrespective of `isDefine`. Whenever there is a call to `Branch` and if the user specific a custom branch size, that branch should be modified accordingly.

Aditya-138-12:
> > Yeah, from the starting i had this in mind that why can't the user set the basket size for all the branches, Although i have tested it locally, the branch_x custom basket size can also be fixed....
> > I was also trying to make a new option for setting the custom basket size for other branches, how should we go about this?
> > Maybe removing the isDefine bool variable will do it?
> 
> I believe just setting all branches with the custom size indicated by the user is the way to go, irrespective of `isDefine`. Whenever there is a call to `Branch` and if the user specific a custom branch size, that branch should be modified accordingly.

Sounds great! This approach ensures consistency and gives users full control over branch sizes. I'll make the changes accordingly. 🚀

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1335,7 +1336,7 @@ void *GetData(T & /*v*/)
 template <typename T>
 void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &inName, const std::string &name,
                        TBranch *&branch, void *&branchAddress, T *address, RBranchSet &outputBranches,
-                       bool /*isDefine*/)
+                       bool isDefine, const std::optional<int> &basketSize)
```
vepadulano, line 1339: The warning you are seeing would be "fixed" by reverting this change and leaving the parameter as it was before: `/*isDefine*/`. A more generic fix would be to remove this parameter from the function signature, then taking care that all calls to `SetBranchesHelper` have the correct number of input parameters, i.e. they are not passing a boolean value that would be used as the `isDefine` parameter

Aditya-138-12: Got it, working on it! :)

Aditya-138-12: Hi @vepadulano ,
I believe the failing tests are not related to my changes. Would it be possible to proceed with the merge now? 😊

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1358,6 +1359,10 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
          outputBranch->SetAddress(address);
          branchAddress = address;
       }
+      // Set basket size specified by the user, regardless of branch types
+      if(basketSize.has_value()){
+         outputBranch->SetBasketSize(basketSize.value());
+      }
```
vepadulano, line 1365: This code is calling `SetBasketSize` potentially at every event. We want to call this method only once, whenever `outputree.Branch` is called. So this part can be removed.

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1378,6 +1383,10 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
       }
    } else {
       outputBranch = outputTree.Branch(name.c_str(), address);
+      // Set Custom basket size for new branches.
+      if(basketSize.has_value()){
+         outputBranch->SetBasketSize(basketSize.value());
+      }
```
vepadulano, line 1389: I would modify this section so that `SetBasketSize` is not needed, but the basket size is passed as an argument of `Branch`, like it happens in line 1380

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1433,8 +1444,16 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
       if (outputBranch) {
          // needs to be SetObject (not SetAddress) to mimic what happens when this TBranchElement is constructed
          outputBranch->SetObject(ab);
+         // Set basket size if specified by the user
+         if(basketSize.has_value()){
+            outputBranch->SetBasketSize(basketSize.value());
+         }
```
vepadulano, line 1450: Similar comments for the other `SetBranchesHelper` overload also apply here.

vepadulano (changes requested):
Let's let the CI finish running. But meanwhile I wanted to comment, we should only have to set the basket size once per branch, not once per event. Ideally, the basket size should be passed directly as an argument of the `Branch` method, i.e. the `SetBasketSize` method should not be needed in principle.

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
Aditya-138-12: Should I do something like this, 
```
const auto bufSize = basketSize.value_or(32000);
outputTree.Branch(name.c_str(), address, buffSize);
```

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
Aditya-138-12: Thank you for your review and valuable insights! I really appreciate your feedback. I'll go ahead and make the suggested changes accordingly. If there's anything specific you'd like me to refine further, please let me know.

Thank you for your Time and Support.

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
vepadulano: Yes, I like that!

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -61,6 +61,103 @@ class RDFSnapshotMT : public ::testing::Test {
 };
```
vepadulano, line 61: I believe we need two more test cases:

1. Check that a call to Snapshot that *does not* specify the custom basket size produces a TTree where branches have the default basket size of 32K. A simplified pseudo-code version would be
```cpp
ROOT::RDataFrame df{1000}; // Dataframe with empty sequential entries
df.Define("x", 42); // Define one column with a scalar value
df.Define("vec", std::vector<int>(...)); // Define column with collection
df.Snapshot(...);

// Check that output TTree has default basket size
```

2. Create a TTree with a custom basket size (e.g. using Snapshot or creating the TFile and writing the TTree manually). Then, create an RDF that reads that TTree, calls Snapshot *without* custom basket size. Then check that the output TTree has the same basket size as the input TTree. Example pseudo-code:
```cpp
void create_ttree_custom_basket_size(){
   // Create TTree with scalar and collection column with custom basket size
}
create_ttree_custom_basket_size();
ROOT::RDataFrame df{inputtree, inputfile};
df.Snapshot(outputtree, outputfile);
// Check that output TTree has the same basket size as the input
```


Remember for both test cases to check *both* scalar and collection columns, *and* to include the test case both in single threaded and multi-threaded mode.

vepadulano (changes requested):
The CI seems to be finishing successfully, that's great! I think we're very close to completion. The implementation should not be modified now. But I added a comment to add two more test cases.

tree/dataframe/test/dataframe_snapshot.cxx
Aditya-138-12: Got it, working on it!!
Thank you so much for your time and support!!

Aditya-138-12: Hi @vepadulano 
I think the implementation of basketSize.value_or(32000) at some places is not working as intended, as i was testing the above testcases, i see the basket size being set to 32000, instead of setting the basket size based on there input TTree.
Can you please guide me on this...

Thank you for your time and support.

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1407,6 +1410,8 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, const std::string &i
    }
    auto *outputBranch = outputBranches.Get(outName);
 
+   //bool isNewBranch = isDefine || !inputBranch;    // Determine if this is a new branch or not (Created via Define).
```
vepadulano, line 1413: Remove commented code

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1455,4 +1637,24 @@ TEST(RDFSnapshotMore, ZeroOutputEntriesMT)
...
+}
+
+// Test for basket size preservation
+TEST(RDFSnapshotMore, BasketSizePreservationMT){
+   ROOT::EnableImplicitMT();
+   TestBasketSizePreservation();
+   ROOT::DisableImplicitMT();
+}
+
+#endif // R__USE_IMT
```
vepadulano, line 1660: Add newline at end of file

vepadulano (changes requested):
Great work! We're converging towards the end, just a few last steps missing:

* Address final minor comments
* Run `git-clang-format` on the code changes for the whole PR
* Rebase all commits into just one commit. I suggest using `git rebase -i` and then `git push --force` to reupload the branch with only one commit, instead of the current 16.

tree/dataframe/test/dataframe_snapshot.cxx
Aditya-138-12: Hi @vepadulano 
I have added a new line at the EOF, but is there a specific reason to that, I'd love to know more about it....
Thank you for your time and support!

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -156,16 +157,20 @@ public:
...
 
-   void Initialize() { /* noop */}
+   void Initialize()
+   { /* noop */
+   }
 
-   void Finalize() { /* noop */}
+   void Finalize()
+   { /* noop */
+   }
```
vepadulano, line 173: are you sure that `git-clang-format` suggests this? The usage is 
```
$: git-clang-format HEAD~N
```

Where N is the number of commits that you want to apply the formatting to. This ensures only the changes in this PR are touched by clang-format. My suggestion is to remove this latest commit with formatting and repeat the formatting with `git-clang-format`

Aditya-138-12: Hi @vepadulano 
I did something like this
```$: clang-format -i <file_location>```

Aditya-138-12: ```$: git-clang-format HEAD~N```
Is suggesting this 
```void Initialize()
{
/* noop*/
}```

Instead of this ```void Initialize(){ /* noop*/}```

vepadulano: >  I thought applying the formatting in the entire file would be correct, please correct me if i am wrong:)

In general we avoid applying formatting to entire files, as often this leads to formatting parts of the file that are not touched by the PR and that leads to unnecessary confusion. `git-clang-format` is the correct tool to use in this case.

Now most of the changes in `ActionHelpers.hxx` are unfortunately very distracting. Before merging, I would like to ask you to revert anything that is not strictly related with the changes of this PR, retrying with `git-clang-format`. Otherwise you will always get those spurious suggestions. IIUC only the first 16 commits are necessary. Maybe try squashing them into one before retrying with `git-clang-format`.

vepadulano (approved):
Thank you @Aditya-138-12 for your contribution!

Aditya-138-12:
> Thank you @Aditya-138-12 for your contribution!

Thank you so much @vepadulano for your time and support!!!

Issue #14943, opened 2024-03-12
Web graphics does not work from PyROOT


jblomer:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

When using Python in interactive mode, the web window does not launch properly (see screenshot). It works from a C++ macro. One can also save canvas contents in ROOT batch mode.

![image](https://github.com/root-project/root/assets/1557360/d63c50fc-0b0f-45a3-8f6c-7cc4b2f65765)

### Reproducer

```python
import ROOT

f1 = ROOT.TF1("f", "sin(x)", 0, 10)
f1.Draw()
```
Output:
```bash
$ python3 test.py 
Info in <THttpEngine::Create>: Starting HTTP server on port 127.0.0.1:9577
Info in <TCanvas::MakeDefCanvas>:  created default TCanvas with name c1
[jakob@seaturtle demo]$ [101327:101345:0312/155756.062028:ERROR:object_proxy.cc(576)] Failed to call method: org.freedesktop.DBus.Properties.Get: object_path= /org/freedesktop/portal/desktop: org.freedesktop.DBus.Error.InvalidArgs: No such interface “org.freedesktop.portal.FileChooser”
[101327:101345:0312/155756.062084:ERROR:select_file_dialog_linux_portal.cc(285)] Failed to read portal version property
[101327:101327:0312/155756.162423:ERROR:policy_logger.cc(156)] :components/enterprise/browser/controller/chrome_browser_cloud_management_controller.cc(161) Cloud management controller initialization aborted as CBCM is not enabled. Please use the `--enable-chrome-browser-cloud-management` command line flag to enable it if you are not using the official Google Chrome build.
Fontconfig error: Cannot load default config file: No such file: (null)
```

### ROOT version

master

### Installation method

From sources

### Operating system

Linux

jblomer:
(Hold on, mixed up the reproducer, fixing...)

linev:
Can you add `ROOT.gROOT.SetWebDisplay("chrome")` to the macro?

For me it works ok.

jblomer:
Ok, now the reproducer is the correct one. Adding `SetWebDisplay("chrome")` doesn't help (also it is already on Chrome).

linev:
It also does not work with `SetWebDisplay("off")`. 

Canvas destroyed before it can be drawn.

I have idea to provide special `TCanvas::Draw()` handling for python - but never find time for this.

linev:
> This is a blocker for the web graphics.

[...]

Issue #15474, opened 2024-05-10
[Web Graphics] Two failures with one simple PyROOT plotter


dpiparo:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

On macOS, it is possible to demonstrate two bugs of the WebGraphics with one simple PyROOT script:
1. A crash, upon first execution
2. A system freeze, once the browser is loaded

## Crash
Upon first execution, a browser window is loaded, however the script exits, even if the `input` function should stop execution. The error message:
```
[70538:42503:0510/092746.969274:ERROR:trust_store_mac.cc(750)] Error parsing certificate:
ERROR: Failed parsing extensions
[70538:259:0510/092747.479455:ERROR:background_contents_service.cc(441)] No extension found for BackgroundContents - id = ejidjjhkpiekdjhfgtyshbnagldlkglhimk

$ 
-> Created TensorFlow Lite XNNPACK delegate for CPU.
```
The last line is not a mistake: it really appears on my terminal.

## Faulty behaviour: Freeze
Now that chrome is loaded, if I re-launch the same script, everything stop before quitting as expected, the webcanvas window is displayed, however it's not responsive: whatever is clicked has no effect.

### Reproducer

```py
import ROOT

ROOT.gROOT.SetWebDisplay()

c = ROOT.TCanvas()
h = ROOT.TH1D("","",64, -4,4)
h.FillRandom("gaus")
h.Draw()

input("Press Enter to exit")
```

### ROOT version

6.32 and master

### Installation method

Sources

### Operating system

MacOS 14

linev:
How you start python - with or without `-i` flag?

Graphics (normal or web) can normally works only in interactive python mode.
In such case canvas will be updated while `input()` is waiting for key.

In none-interactive mode events processing does not work.

linev:
I create separate issue https://github.com/root-project/root/issues/15498 concerning `gPad` in this script.
For me python creates second canvas - without real need.

dpiparo:
Thanks. I do not understand. I proposed a simple reproducer that illustrates a bad user experience, very bad, that makes the system unusable. What is the strategy to fix this bug?

linev:
Again, workaround with calling `input()` function in python does not work - neither with web not with normal ROOT graphics. Only when enabling interactive python mode `python - i` one gets ROOT events processing working.

This problem well known and I discussed it some time ago with @vepadulano. 
[...]

Issue #13744, opened 2023-09-28
[PyROOT, Graphics] Cannot pause pyROOT script with `input()` and see canvases


dpiparo:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

This script used to work, also on MacOS
```
import ROOT

h = ROOT.TH1F("foo", "bar", 64, -4, 4)
h.FillRandom("gaus")
c=ROOT.TCanvas()
h.Draw()
c.Update()

input()
```

and allowed to see and interact with the canvas to then end execution upon pressing a key. This results now in a deadlock in master.

### Reproducer

```
import ROOT

h = ROOT.TH1F("foo", "bar", 64, -4, 4)
h.FillRandom("gaus")
c=ROOT.TCanvas()
h.Draw()
c.Update()

input()
```

### ROOT version

master

### Installation method

built from sources

### Operating system

MacOS

linev:
I can reproduce problem with `input()` also on my Linux, 
And I see that MacOS handled differently.
I will provide PR soon to see if small changes could help

dpiparo:
Reopened b/c of this https://github.com/root-project/root/pull/13754

linev:
Seems to be it is not possible to run extra thread in python on Mac in script mode.
At the same time input hook is also not working in script mode.
Thus canvas - even shown - remains fully frozen. 
It is exactly a situation as was in previous ROOT releases.

In web case this leads just to empty web browser.
Probably I can improve this

Axel-Naumann:
Do I understand correctly that this is an old (i.e. <v6.28) issue? Can we then fix it in a patch release of 6.30 rather than blocking 6.30/00 for it?

dpiparo:
Yes, we can, and in my opinion, should.

linev:
After #17587 one can use block parameter in `canv.Draw(block = True)`
This includes waiting of user press `<space>` button on keyboard and running event loop for interactive graphics.
So macro will look like:

```
import ROOT

h = ROOT.TH1F("foo", "bar", 64, -4, 4)
h.FillRandom("gaus")
c=ROOT.TCanvas()
h.Draw()
c.Draw(block = True)
```

PR #17587, opened 2025-01-31
Support interactive graphics in python scripts

linev:
Currently it is problematic to use simple canvas (web or not-web) in non-interactive python.
Either macro exit very fast or created canvas is not possible to use interactively.

I propose to provide pythonization of `TCanvas.Update()` and `TCanvas.Draw()` methods which can block macro execution and wait until user press `<space>` key. While python waits for key press ROOT event loop continues to run and interactive canvas is fully functional. 

There is additional `block` parameter of `canvas.Update()` now which is False by default.
Without extra arguments it behaves as normal ROOT. Only when `block = True`, method blocks until <space> key pressed.

Simple example of usage:
```
import ROOT
ROOT.gROOT.SetWebDisplay("chrome")

c = ROOT.TCanvas()

h = ROOT.TH1I("h1", "h1", 100, -5, 5)
h.FillRandom("gaus", 10000)
h.Draw()

# block here until space is pressed
c.Update(True)

# continue macro execution
```

In batch mode or in python notebooks blocking will not be performed.

Tested on:
   * Linux, Mac and Windows, 
   * interactive/plain python mode, 
   * web/normal graphics

The only difference between Windows and other platforms - in interactive python mode one 
cannot use special hook to perform periodic actions while python prompt is waiting for next user command.
Therefore on Windows to work with canvas one always have to invoke `c.Update(True)`

Also `c.Draw(True)` behaves is the same way. It draw canvas, perform update and then run event loop until space button is pressed.

This PR resolves: #14943, #13744, #15474

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
@@ -111,7 +111,7 @@ def _TCanvas_Update(self, block = False):
       run_root_event_loop()
 
 
-def _TCanvas_Draw(self, block = False):
+def _TCanvas_Draw(self, option: str = "", block: bool = False):
    """
    Draw the canvas.
    Also blocks script execution and runs the ROOT graphics event loop until the <space> keyword is pressed,
@@ -121,14 +121,10 @@ def _TCanvas_Draw(self, block = False):
    * The script is running not in ipython notebooks.
    """
 
-   if isinstance(block, str):
-      self._Draw(block)
-      return
-
-   self._Draw()
+   self._Draw(option)
 
    # run loop if block flag is set
-   if block == True:
+   if block:
       self._Update()
       run_root_event_loop()
 


bellenot:
@linev and what about the browser?

linev:
@vepadulano @dpiparo 

Can you check PR? 
It finally allows to use graphics in python.
One may apply same approach also to `TCanvas::Draw()` method.

vepadulano:
@linev thank you for this PR! I believe it is improving a situation for a specific use case of our Python interface, that's good! Let me summarise my findings from a few quick tests I ran locally.

* The modified `canvas.Update` function allows to avoid having to specify the `-i` flag when running `python -i myscript.py`. As a side-note, this is more similar to what a script using matplotilib would do by default, with the extra functionality that the application can resume after the space key has been pressed. matplotlib by default blocks execution and you can just exit the Python process by pressing CTRL-C.
* Probably we want to have a similar functionality for `canvas.Draw`, because that's the first thing a user would do when writing their plotting script, it always starts with a `Draw` call, not with `Update`.
* This is a backwards-incompatible change in general. I could come up at least with one simple use case that would be broken by this change. That is, a live plot being updated. Consider this simple example which fills the contents of a histogram in a loop and refreshes the plot at every iteration:
```python
import ROOT

c = ROOT.TCanvas()
h = ROOT.TH1D("h","h",100,-10,10)
h.Draw()
c.Draw()
c.Update()
ROOT.gSystem.ProcessEvents()

for _ in range(1000):
    h.FillRandom("gaus",100)
    h.Draw()
    c.Update()
    ROOT.gSystem.ProcessEvents()
```
  which will be broken with the current changes. As discussed offline, this can be accommodated by adding a Python-only 
  optional argument to `Update`, `Draw` and anything else that will be modified, which switches to the blocking behaviour, 
  leaving by default the current behaviour which does not block.


Let me note that this example I came up with is just one, there could be many more. I wish there was a better way we could expedite this kind of testing. Notably, I have tried these changes at a Python prompt, at a IPython prompt and in a Jupyter notebook. Everything seems to work as intended there, but having to do this manually is unfortunate.

linev:
@vepadulano 

I add extra `block` argument to `canvas.Update(block = False)` pythonization.
By default it behaves as normal ROOT - means your macro will work as before.
Only when called `canvas.Update(True)` script execution will be blocked until `<space>` button is pressed

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
```diff
@@ -0,0 +1,83 @@
...
+
+   finally:
+      termios.tcsetattr(sys.stdin, termios.TCSADRAIN, old_settings)
+
+
+def _TCanvas_Update(self, block = False):
+   """
+   First invoke normal canvas Update.
+   If block == True specified, run ROOT event loop until <space> key is pressed
+   Event loop does not processed in batch mode or in ipython
```
vepadulano, line 47:
```suggestion
   Updates the canvas. Also blocks script execution and runs the ROOT graphics event loop until the <space> keyword is pressed, but only if the following conditions are met:
   * The `block` optional argument is set to `True`.
   * ROOT graphics are enabled, i.e. `ROOT.gROOT.IsBatch() == False`.
   * The script is being run in non-interactive mode, i.e. `python myscript.py`.
```

bindings/pyroot/pythonizations/python/ROOT/__init__.py
```diff
@@ -174,14 +174,13 @@ def find_spec(self, fullname: str, path, target=None) -> ModuleSpec:
...
 
-def cleanup():
-    # If spawned, stop thread which processes ROOT events
-    facade = sys.modules[__name__]
-    if "app" in facade.__dict__ and hasattr(facade.__dict__["app"], "process_root_events"):
-        facade.__dict__["app"].keep_polling = False
-        facade.__dict__["app"].process_root_events.join()
-
-atexit.register(cleanup)
+# atexit.register(cleanup)
```
vepadulano, line 186: Please remove all the commented code

bindings/pyroot/pythonizations/python/ROOT/_application.py
```diff
@@ -78,28 +78,29 @@ def init_graphics(self):
...
+            # def _process_root_events(self):
+            #     while self.keep_polling:
+            #         gSystem.ProcessEvents()
+            #         time.sleep(0.01)
+            # import threading
+            # self.keep_polling = True # Used to shut down the thread safely at teardown time
+            # update_thread = threading.Thread(None, _process_root_events, None, (self,))
+            # self.process_root_events = update_thread # The thread is joined at teardown time
+            # update_thread.daemon = True
+            # update_thread.start()
```
vepadulano, line 103: Also here please remove commented code

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
```diff
@@ -0,0 +1,83 @@
+from . import pythonization
+
```
vepadulano, line 2: We need the usual copyright header here and I would also add some documentation that will end up in the Python Interface section of the docs https://root.cern.ch/doc/master/group__Pythonizations.html. Take for example https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tfile.py

vepadulano (changes requested):
Thanks for the update, I left some comments. I still believe the `Draw` method sees more usage than `Update`.

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
linev: @vepadulano 

I apply your changes and rebase commits

vepadulano: We are still missing the documentation as I suggested, see the `\pythondoc` block in `_tfile.py`. I think it should describe the fact that the `Update` method has been pythonized and a simple example usage.

vepadulano (changes requested):

linev:
Yes, I add small docu and example macro

linev:
@vepadulano 

I add commit that also `c.Draw(True)` behaves is the same way. 
It draw canvas, perform update and then run event loop until space button is pressed.
I also checked that `jupyter lab` works as before

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
```diff
@@ -0,0 +1,145 @@
...
+   Also blocks script execution and runs the ROOT graphics event loop until the <space> keyword is pressed,
+   but only if the following conditions are met:
+   * The `block` optional argument is set to `True`.
+   * ROOT graphics are enabled, i.e. `ROOT.gROOT.IsBatch() == False`.
+   * The script is running not in ipython notebooks.
+   """
+
+   if isinstance(block, str):
+      self._Draw(block)
+      return
```
vepadulano: This is not good practice, we are injecting a different meaning *and* type into an existing argument, i.e. `TCanvas::Draw(Option_t *)`. Now, unfortunately the `Option_t *` argument in TCanvas::Draw is not even used, which is a poor design choice, but does not improve this one. The correct function call should be

```python
c = ROOT.TCanvas()
c.Draw("", True)
```

Or equivalently
```python
c = ROOT.TCanvas()
c.Draw(block=True)
```

This is achieved with the change

```python

def _TCanvas_Draw(self, option: str = "", block: bool = False):

   self._Draw(option)

   # run loop if block flag is set
   if block == True:
      self._Update()
      run_root_event_loop()
```

linev: I add commit with proposed changes

PR #17591, opened 2025-02-01
io: Allow implicit conversion from  vector<pair<A<B> > to map<A,B>

pcanal:
The roottest companion PR is https://github.com/root-project/roottest/pull/1255

### DIFF
--- a/io/io/src/TBufferJSON.cxx
+++ b/io/io/src/TBufferJSON.cxx
@@ -717,8 +717,10 @@ TString TBufferJSON::StoreObject(const void *obj, const TClass *cl)
 
 ////////////////////////////////////////////////////////////////////////////////
 /// Converts selected data member into json
-/// Parameter ptr specifies address in memory, where data member is located
-/// compact parameter defines compactness of produced JSON (from 0 to 3)
+/// Parameter ptr specifies address in memory, where data member is located.
+/// Note; if data member described by 'member'is pointer, `ptr` should be the
+/// value of the pointer, not the address of the pointer.
+/// compact parameter defines compactness of produced JSON (from 0 to 3).
 /// arraylen (when specified) is array length for this data member,  //[fN] case
 
 TString TBufferJSON::ConvertToJSON(const void *ptr, TDataMember *member, Int_t compact, Int_t arraylen)
@@ -1002,6 +1004,8 @@ void *TBufferJSON::ConvertFromJSONChecked(const char *str, const TClass *expecte
 
 ////////////////////////////////////////////////////////////////////////////////
 /// Convert single data member to JSON structures
+/// Note; if data member described by 'member'is pointer, `ptr` should be the
+/// value of the pointer, not the address of the pointer.
 /// Returns string with converted member
 
 TString TBufferJSON::JsonWriteMember(const void *ptr, TDataMember *member, TClass *memberClass, Int_t arraylen)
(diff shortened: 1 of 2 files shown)


jblomer (approved):
LGTM. Regarding the roottest PR: shouldn't it just test this conversion?

pcanal:
> Regarding the roottest PR: shouldn't it just test this conversion?

For once, it was the reverse and this PR is fixing an issue that was (re)discovered by the test reorganization and subsequent enabling of a test that had been forgotten.

pcanal:
The Windows failure is 'real' and was not noticed before as the test was not run on Windows (due to being part of a gmake tests)

io/io/src/TBufferJSON.cxx
```diff
@@ -1075,7 +1082,7 @@ TString TBufferJSON::JsonWriteMember(const void *ptr, TDataMember *member, TClas
          if (indx.IsArray())
             fOutBuffer.Append(indx.NextSeparator());
 
-         ppp += shift * member->GetUnitSize();
+         ppp += shift * unitSize;
```
pcanal, line 1085: @linev FYI

pcanal: This cause https://github.com/root-project/root/actions/runs/13082767761/job/36600813317 on Windows x64 because `sizeof(long)==32` vs `sizeof(void*)==64`

linev: With `UnitSize` change I agree - this was not clear for me.
And most probably Win64 was not there when code was introduced.

But in the PR you also changed pointer meaning which is provided to public static method:
```
static TString ConvertToJSON(const void *obj, TDataMember *member, Int_t compact = 0, Int_t arraylen = -1);
```
I do not see why we need this change.
If by chance users using this method - code will no longer working.

pcanal: The problem is that the function is was unclear and conflicted with the documentation:
```
/// Parameter ptr specifies address in memory, where data member is located
```
whereas the actual behavior added "but if the data member is a pointer the caller should de reference the pointer before calling this function".  As a consequence `TDataMember` was also only 'sort of' describing the memory pointed to by the parameter (i.e. everything but the 'is pointer' part) and made reasoning about the code harder.

> If by chance users using this method - code will no longer working.

That is indeed true. IMHO there is also a good chance they would be using it wrong (due to the documentation and due to the counter-intuitiveness of the previous behavior (the TDataMember is only sort-of describing the ptr).

linev: > That is indeed true. IMHO there is also a good chance they would be using it wrong (due to the documentation and due to the counter-intuitiveness of the previous behavior (the TDataMember is only sort-of describing the ptr).

If one tried to use this method - they just copy-and-paste existing code.

pcanal: Why would someone duplicate so much of the `TBufferJSON` code?   

You are right that we can not 'silently' change the behavior but I strongly propose with make this function private (and thus force this hypothetical user of a function that is really internal reach out to us), would that be okay with you?

linev: As I said - if we change visibility of the method and anybody uses it - this will be no longer possible.
If we change arguments meaning - user will be forced to add `#if  ROOT_VERSION` to handle such change.

I intentionally keep this signature public to let user work with class members when converting into JSON.
But have no idea if it really used or not.

pcanal: This is unfortunate as the function is 'hard' to use.  However, I will revert the change and replace it with an update of the documentation.

pcanal:
There was a timeout in `1735:roottest-python-distrdf-backends-test_all` which is unrelated/pre-existing.

PR #17595, opened 2025-02-02
Changing auto to double in ROOT::VecOps::Var definition

edfink234:
# This Pull request:

## Changes or fixes:

Changing auto to double in `ROOT::VecOps::Var` definition

## Checklist:

- [ ] tested changes locally

This PR fixes #

### DIFF
--- a/math/vecops/inc/ROOT/RVec.hxx
+++ b/math/vecops/inc/ROOT/RVec.hxx
@@ -2107,7 +2107,7 @@ double Var(const RVec<T> &v)
    auto pred = [&sum_squares, &squared_sum](const T& x) {sum_squares+=x*x; squared_sum+=x;};
    std::for_each(v.begin(), v.end(), pred);
    squared_sum *= squared_sum;
-   const auto dsize = (double) size;
+   const double dsize = (double) size;
    return 1. / (dsize - 1.) * (sum_squares - squared_sum / dsize );
 }
 


silverweed:
Hello,
why is this change needed?

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -2107,7 +2107,7 @@ double Var(const RVec<T> &v)
    auto pred = [&sum_squares, &squared_sum](const T& x) {sum_squares+=x*x; squared_sum+=x;};
    std::for_each(v.begin(), v.end(), pred);
    squared_sum *= squared_sum;
-   const auto dsize = (double) size;
+   const double dsize = (double) size;
```
ferdymercury, line 2110:
```suggestion
   const double dsize = size;
```

silverweed: This code is not really better than the original since it relies on implicit conversions. At that point one could forego the `dsize` variable entirely and rely on `(size - 1.)` in the next line to do the implicit conversion.
But I still don't see the point of changing this code like this unless there is a real reason for it.

ferdymercury: > and rely on `(size - 1.)` in the next line to do the implicit conversion.

I think the problem in that line is rather with `squared_sum / size` when `T = int`. But with a static_cast<double>(size) there that would be solved.

In any case I agree w you and I'd vote for closing.

PR #17596, opened 2025-02-03
[ntuple] remove wrong assert in RNTupleSerialize.cxx

silverweed:
This assert was introduced in da99545d831, but it is wrong (it may be the case that we call MapSchema when a header extension is already there). This code path will be exercised (and tested) in an upcoming PR about incremental merging (https://github.com/root-project/root/pull/17563/commits/9469b920c31be9fb03500b633946d83a1ea1c657)

## Checklist:

- [x] tested changes locally


hahnjo (approved):
IMHO would have made more sense together with https://github.com/root-project/root/pull/17597 which also tests this, but LGTM

PR #17597, opened 2025-02-03
[ntuple] Fix serialization when a header extension is already present

silverweed:
Based on #17596, derived from #17563.
Introduces a new method in RNTupleDescriptor/Builder to allow cloning only the "schema" part of a descriptor, then uses this new method to fix `RPagePersistentSink::InitFromDescriptor` in the general case where the given descriptor already contains a header extension.
The RNTupleSerializer also needs a couple of changes to cope with this case, because we don't want to move the fields and columns from the header extension to the regular header (which would hinder the Merger's capability of doing incremental merging).

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
@@ -1051,6 +1051,13 @@ public:
    /// We cannot create this vector when building the fFields because at the time when AddExtendedField is called,
    /// the field is not yet linked into the schema tree.
    std::vector<DescriptorId_t> GetTopLevelFields(const RNTupleDescriptor &desc) const;
+
+   bool ContainsField(DescriptorId_t fieldId) const { return fFieldIdsLookup.find(fieldId) != fFieldIdsLookup.end(); }
+   bool ContainsExtendedColumnRepresentation(DescriptorId_t columnId) const
+   {
+      return std::find(fExtendedColumnRepresentations.begin(), fExtendedColumnRepresentations.end(), columnId) !=
+             fExtendedColumnRepresentations.end();
+   }
 };
 
 namespace Internal {
--- a/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
@@ -56,6 +56,8 @@ class RClusterGroupDescriptorBuilder;
 class RExtraTypeInfoDescriptorBuilder;
 class RFieldDescriptorBuilder;
 class RNTupleDescriptorBuilder;
+
+RNTupleDescriptor CloneDescriptorSchema(const RNTupleDescriptor &desc);
 } // namespace Internal
 
 // clang-format off
@@ -538,6 +540,7 @@ writte struct. This allows for forward and backward compatibility when the meta-
 // clang-format on
 class RNTupleDescriptor final {
    friend class Internal::RNTupleDescriptorBuilder;
+   friend RNTupleDescriptor Internal::CloneDescriptorSchema(const RNTupleDescriptor &desc);
 
 public:
    class RHeaderExtension;
@@ -548,15 +551,26 @@ private:
    /// Free text from the user
    std::string fDescription;
 
-   std::uint64_t fOnDiskHeaderXxHash3 = 0; ///< Set by the descriptor builder when deserialized
-   std::uint64_t fOnDiskHeaderSize = 0;    ///< Set by the descriptor builder when deserialized
-   std::uint64_t fOnDiskFooterSize = 0; ///< Like fOnDiskHeaderSize, contains both cluster summaries and page locations
+   DescriptorId_t fFieldZeroId = kInvalidDescriptorId; ///< Set by the descriptor builder
 
-   std::uint64_t fNEntries = 0;         ///< Updated by the descriptor builder when the cluster groups are added
-   std::uint64_t fNClusters = 0;        ///< Updated by the descriptor builder when the cluster groups are added
    std::uint64_t fNPhysicalColumns = 0; ///< Updated by the descriptor builder when columns are added
 
-   DescriptorId_t fFieldZeroId = kInvalidDescriptorId; ///< Set by the descriptor builder
+   std::set<unsigned int> fFeatureFlags;
+   std::unordered_map<DescriptorId_t, RFieldDescriptor> fFieldDescriptors;
+   std::unordered_map<DescriptorId_t, RColumnDescriptor> fColumnDescriptors;
+
+   std::vector<RExtraTypeInfoDescriptor> fExtraTypeInfoDescriptors;
+   std::unique_ptr<RHeaderExtension> fHeaderExtension;
+
+   //// All fields above are part of the schema and are cloned when creating a new descriptor from a given one
+   //// (see CloneSchema())
+
+   std::uint64_t fOnDiskHeaderSize = 0;    ///< Set by the descriptor builder when deserialized
+   std::uint64_t fOnDiskHeaderXxHash3 = 0; ///< Set by the descriptor builder when deserialized
+   std::uint64_t fOnDiskFooterSize = 0; ///< Like fOnDiskHeaderSize, contains both cluster summaries and page locations
+
+   std::uint64_t fNEntries = 0;  ///< Updated by the descriptor builder when the cluster groups are added
+   std::uint64_t fNClusters = 0; ///< Updated by the descriptor builder when the cluster groups are added
 
    /**
     * Once constructed by an RNTupleDescriptorBuilder, the descriptor is mostly immutable except for set of
@@ -567,9 +581,6 @@ private:
     */
    std::uint64_t fGeneration = 0;
 
-   std::set<unsigned int> fFeatureFlags;
-   std::unordered_map<DescriptorId_t, RFieldDescriptor> fFieldDescriptors;
-   std::unordered_map<DescriptorId_t, RColumnDescriptor> fColumnDescriptors;
    std::unordered_map<DescriptorId_t, RClusterGroupDescriptor> fClusterGroupDescriptors;
    /// References cluster groups sorted by entry range and thus allows for binary search.
    /// Note that this list is empty during the descriptor building process and will only be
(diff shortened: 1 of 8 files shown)


tree/ntuple/v7/src/RNTupleDescriptor.cxx
```diff
@@ -663,32 +663,43 @@ ROOT::Experimental::RNTupleDescriptor::CreateModel(const RCreateModelOptions &op
...
-   clone.fOnDiskHeaderXxHash3 = fOnDiskHeaderXxHash3;
-   clone.fOnDiskHeaderSize = fOnDiskHeaderSize;
-   clone.fOnDiskFooterSize = fOnDiskFooterSize;
-   clone.fNEntries = fNEntries;
-   clone.fNClusters = fNClusters;
    clone.fNPhysicalColumns = fNPhysicalColumns;
    clone.fFieldZeroId = fFieldZeroId;
-   clone.fGeneration = fGeneration;
+   clone.fOnDiskHeaderSize = fOnDiskHeaderSize;
+   // OnDiskFooterSize not copied because it would not be correct (since we change the clustering)
```
hahnjo, line 674: This comment is specific to the merger. I think the proper explanation is
```suggestion
   // OnDiskFooterSize not copied because it contains information beyond the schema, for example the clustering.
```

tree/ntuple/v7/src/RNTupleDescriptor.cxx
```diff
@@ -663,32 +663,43 @@ ROOT::Experimental::RNTupleDescriptor::CreateModel(const RCreateModelOptions &op
...
+      clone.fHeaderExtension = std::make_unique<RHeaderExtension>(*fHeaderExtension);
+
+   return clone;
+}
+
+ROOT::Experimental::RNTupleDescriptor ROOT::Experimental::RNTupleDescriptor::Clone() const
+{
+   RNTupleDescriptor clone = CloneSchema();
+
+   clone.fOnDiskHeaderXxHash3 = fOnDiskHeaderXxHash3;
```
hahnjo, line 693: Shouldn't this move to `CloneSchema()`?

tree/ntuple/v7/inc/ROOT/RPageStorage.hxx
hahnjo, line 1: Can you please make the commit title more meaningful than "fix"? It's generally discouraged by various "writing good commit messages" guides...

tree/ntuple/v7/src/RPageStorage.cxx
```diff
@@ -919,39 +919,67 @@ void ROOT::Experimental::Internal::RPagePersistentSink::InitImpl(RNTupleModel &m
...
+      const auto &column = descriptor.GetColumnDescriptor(i);
+      RClusterDescriptor::RColumnRange columnRange;
+      columnRange.fPhysicalColumnId = i;
+      columnRange.fFirstElementIndex = column.GetFirstElementIndex();
+      columnRange.fNElements = 0;
+      columnRange.fCompressionSettings = GetWriteOptions().GetCompression();
+      fOpenColumnRanges.emplace_back(columnRange);
+      RClusterDescriptor::RPageRange pageRange;
+      pageRange.fPhysicalColumnId = i;
+      fOpenPageRanges.emplace_back(std::move(pageRange));
```
hahnjo, line 943: Is there a reason for moving one but not the other? I think both structures aren't big nor are we on a performance critical path, so even just `push_back` should be fine

tree/ntuple/v7/inc/ROOT/RNTupleDescriptor.hxx
hahnjo, line 1: Same here, can you write a more meaningful commit title?

tree/ntuple/v7/test/ntuple_test.hxx
```diff
@@ -114,6 +114,8 @@ using RPrepareVisitor = ROOT::Experimental::RPrepareVisitor;
 using RPrintSchemaVisitor = ROOT::Experimental::RPrintSchemaVisitor;
 using RRawFile = ROOT::Internal::RRawFile;
 using EContainerFormat = RNTupleFileWriter::EContainerFormat;
+template <typename T>
+using RNTupleView = ROOT::Experimental::RNTupleView<T>;
```
hahnjo, line 118: As Florine remarked in https://github.com/root-project/root/pull/17598#discussion_r1939166623, "This change seems unrelated to the rest of commit."

tree/ntuple/v7/src/RNTupleDescriptor.cxx
```diff
@@ -663,32 +663,43 @@ ROOT::Experimental::RNTupleDescriptor::CreateModel(const RCreateModelOptions &op
    return model;
 }
 
-ROOT::Experimental::RNTupleDescriptor ROOT::Experimental::RNTupleDescriptor::Clone() const
+ROOT::Experimental::RNTupleDescriptor ROOT::Experimental::RNTupleDescriptor::CloneSchema() const
```
hahnjo, line 666: `CloneSchema` should be exercised in a unit test, in particular which fields are not expected to be there

hahnjo (changes requested):

tree/ntuple/v7/inc/ROOT/RPageStorage.hxx
silverweed: What do you mean? The explanation is in the rest of the commit message, do you think it's not enough and I should add more information to it?

tree/ntuple/v7/src/RPageStorage.cxx
silverweed: `RPageRange` has a deleted destructor so it must be moved (see also line 876)

tree/ntuple/v7/inc/ROOT/RPageStorage.hxx
hahnjo: The reason I insist is that the commit title is what you see on GitHub in the list of commits and during blame, as well as locally with `log --oneline`. Having only "fix InitFromDescriptor" would make me wonder what is fixed, so if I'm searching a problem I have to go into the commit message and diff to see if it's relevant for my search. The new title "fix InitFromDescriptor handling of header extension" is much better in that regard :+1:

tree/ntuple/v7/src/RNTupleDescriptor.cxx
```diff
@@ -663,32 +663,43 @@ ROOT::Experimental::RNTupleDescriptor::CreateModel(const RCreateModelOptions &op
...
-   clone.fOnDiskHeaderSize = fOnDiskHeaderSize;
-   clone.fOnDiskFooterSize = fOnDiskFooterSize;
-   clone.fNEntries = fNEntries;
-   clone.fNClusters = fNClusters;
    clone.fNPhysicalColumns = fNPhysicalColumns;
    clone.fFieldZeroId = fFieldZeroId;
-   clone.fGeneration = fGeneration;
+   clone.fOnDiskHeaderSize = fOnDiskHeaderSize;
+   clone.fOnDiskHeaderXxHash3 = fOnDiskHeaderXxHash3;
+   // OnDiskFooterSize not copied because it contains information beyond the schema, for example the clustering.
```
jblomer, line 675: I think we cannot clone the header size and hash either because the extra type info (and the feature flags) gets merged from the extension header into the main header. We should perhaps clarify that header size, footer size, and xxhash3 are only valid when reading the descriptor from disk. Maybe they should also be optional.

tree/ntuple/v7/src/RNTupleDescriptor.cxx
jblomer, line 666: Not strictly needed now but we should clone the feature flags as well.

tree/ntuple/v7/src/RPageStorage.cxx
```diff
@@ -919,39 +919,67 @@ void ROOT::Experimental::Internal::RPagePersistentSink::InitImpl(RNTupleModel &m
...
-   }
+   // Create new descriptor
+   fDescriptorBuilder.SetSchemaFromExisting(srcDescriptor);
+   const auto &descriptor = fDescriptorBuilder.GetDescriptor();
 
-   auto clusterId = descriptor.FindClusterId(0, 0);
+   // Create column/page ranges
+   const auto nColumns = descriptor.GetNPhysicalColumns();
+   fOpenColumnRanges.reserve(fOpenColumnRanges.size() + nColumns);
+   fOpenPageRanges.reserve(fOpenPageRanges.size() + nColumns);
```
jblomer, line 932: Would `fOpenColumnRanges.size()` or `fOpenPageRanges.size()` ever not be 0?

tree/ntuple/v7/src/RPageStorage.cxx
```diff
@@ -919,39 +919,67 @@ void ROOT::Experimental::Internal::RPagePersistentSink::InitImpl(RNTupleModel &m
...
-
-      RClusterDescriptorBuilder clusterBuilder;
-      clusterBuilder.ClusterId(fDescriptorBuilder.GetDescriptor().GetNActiveClusters())
-         .FirstEntryIndex(fPrevClusterNEntries)
-         .NEntries(nEntries);
-
       for (unsigned int i = 0; i < fOpenColumnRanges.size(); ++i) {
          R__ASSERT(fOpenColumnRanges[i].fPhysicalColumnId == i);
+         if (!cluster.ContainsColumn(i))
+            break;
```
jblomer, line 954: Perhaps add a comment when this can happen

tree/ntuple/v7/src/RNTupleSerialize.cxx
```diff
@@ -247,13 +247,21 @@ std::uint32_t SerializePhysicalColumn(const ROOT::Experimental::RColumnDescripto
...
    auto pos = base;
    void **where = (buffer == nullptr) ? &buffer : reinterpret_cast<void **>(&pos);
 
+   const auto *xHeader = !forHeaderExtension ? desc.GetHeaderExtension() : nullptr;
+
    for (auto parentId : fieldList) {
+      // If we're serializing the non-extended header and we already have a header extension (which may happen if
+      // we load an RNTuple for incremental merging), we need to skip all the extended fields, as they need to be
+      // written in the header extension, not in the regular header.
+      if (xHeader && xHeader->ContainsField(parentId))
```
jblomer, line 262: Don't we need to check if the _column_ is part of the extension header? Then we would also cover extended column representations.

tree/ntuple/v7/src/RNTupleSerialize.cxx
```diff
@@ -476,13 +484,18 @@ std::uint32_t SerializeAliasColumn(const ROOT::Experimental::RColumnDescriptor &
...
+                                            void *buffer, bool forHeaderExtension)
 {
    auto base = reinterpret_cast<unsigned char *>(buffer);
    auto pos = base;
    void **where = (buffer == nullptr) ? &buffer : reinterpret_cast<void **>(&pos);
 
+   const auto *xHeader = !forHeaderExtension ? desc.GetHeaderExtension() : nullptr;
+
    for (auto parentId : fieldList) {
+      if (xHeader && xHeader->ContainsField(parentId))
```
jblomer, line 496: Same as above

jblomer (changes requested):

tree/ntuple/v7/src/RNTupleSerialize.cxx
silverweed: I need to check both, right? Since the header extension only contains information about the extended column ids..

jblomer: I thought that you'd anyway only process the first $k$ fields of the regular header if `forHeaderExtension` is false.

jblomer (approved):

PR #17600, opened 2025-02-03
[net] Allow users to specify binding options when opening sockets

dpiparo:
in a backwards compatible way. An enumerator, ESocketBindOption, is introduced to specify the option.

### DIFF
--- a/core/winnt/inc/TWinNTSystem.h
+++ b/core/winnt/inc/TWinNTSystem.h
@@ -232,8 +232,8 @@ public:
    int               GetServiceByName(const char *service) override;
    char              *GetServiceByPort(int port) override;
    int               OpenConnection(const char *server, int port, int tcpwindowsize = -1, const char *protocol = "tcp") override;
-   int               AnnounceTcpService(int port, Bool_t reuse, int backlog, int tcpwindowsize = -1) override;
-   int               AnnounceUdpService(int port, int backlog) override;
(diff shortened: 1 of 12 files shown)


core/winnt/inc/TWinNTSystem.h
```diff
@@ -232,8 +232,8 @@ class TWinNTSystem : public TSystem {
    int               GetServiceByName(const char *service) override;
    char              *GetServiceByPort(int port) override;
    int               OpenConnection(const char *server, int port, int tcpwindowsize = -1, const char *protocol = "tcp") override;
-   int               AnnounceTcpService(int port, Bool_t reuse, int backlog, int tcpwindowsize = -1) override;
-   int               AnnounceUdpService(int port, int backlog) override;
+   int               AnnounceTcpService(int port, Bool_t reuse, int backlog, int tcpwindowsize = -1, ESocketBindOption socketBindOption = = ESocketBindOption::kInaddrAny) override;
```
jblomer, line 235:
```suggestion
   int               AnnounceTcpService(int port, Bool_t reuse, int backlog, int tcpwindowsize = -1, ESocketBindOption socketBindOption = ESocketBindOption::kInaddrAny) override;
```

jblomer (approved):
LGTM. I think we can unit test the new functionality with `TServerSocket::GetLocalInetAddress().GetHostAddress() != "0.0.0.0"`

core/winnt/inc/TWinNTSystem.h
dpiparo: Thanks. Added the test and removed the typo on the file compiled only on windows.

Issue #14557, opened 2024-02-04
[ROOT-4550] TMessage doesn't honour kIsOwner bit when compression is used


vepadulano:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

From https://its.cern.ch/jira/browse/ROOT-4550

For TBuffer, one can set kIsOwner to false to ensure it doesn't delete the buffer passed to it. However, when using TMessage with compression, TMessage(void *buf, Int_t bufsize) moves the original fBuffer to fBufComp and decompresses that into fBuffer (I think).

When the TMessage is deleted, fBufComp will be deleted even if kIsOwner is false. Since it contains the original buffer passed to it, this is dangerous (and inconsistent), and in my case crashes.
fBuffer isn't deleted, but maybe it should be (allocated by TMessage).

Besides the destructor, there are a few other places in TMessage that also delete fBufComp, these would also be affected.

### Reproducer

As per the description

### ROOT version

Any

### Installation method

Any

### Operating system

Any

gganis:
Possible solution in a provate branch. Plan to create PR after some additional local tests

PR #17601, opened 2025-02-03
TMessage honour kIsOwner bit when compression is used

ferdymercury:
# This Pull request:

## Changes or fixes:

(Checking gganis solution)

Fixes https://github.com/root-project/root/issues/14557

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/net/net/inc/TMessage.h
+++ b/net/net/inc/TMessage.h
@@ -59,6 +59,9 @@ private:
    Bool_t TestBitNumber(UInt_t bitnumber) const { return fBitsPIDs.TestBitNumber(bitnumber); }
 
 protected:
+   enum EStatusBits {
+     kIsOwnerComp = BIT(18) // if TMessage owns fBufComp
+   };
    TMessage(void *buf, Int_t bufsize);   // only called by T(P)Socket::Recv()
    void SetLength() const;               // only called by T(P)Socket::Send()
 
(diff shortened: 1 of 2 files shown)


net/net/src/TMessage.cxx
```diff
@@ -82,12 +82,18 @@ TMessage::TMessage(void *buf, Int_t bufsize) : TBufferFile(TBuffer::kRead, bufsi
    fInfos      = nullptr;
    fEvolution  = kFALSE;
 
+   // The receiving socket owns the passed buffer
+   ResetBit(kIsOwner);
```
pcanal, line 86: This does not sounds consistent with the `TBufferFile` behavior.  We may need to extend the interface to match.
The corresponding (and called via the base class) constructor:
```
TBufferFile(TBuffer::EMode mode, Int_t bufsiz, void *buf, Bool_t adopt = kTRUE, ReAllocCharFun_t reallocfunc = nullptr);
```
where the default is to adopt/take-ownership of the passed buffer.

net/net/inc/TMessage.h
```diff
@@ -59,6 +59,9 @@ friend class TXSocket;
    Bool_t TestBitNumber(UInt_t bitnumber) const { return fBitsPIDs.TestBitNumber(bitnumber); }
 
 protected:
+   enum EStatusBits {
+     kIsOwnerComp = BIT(18) // if TMessage owns fBufComp
```
ferdymercury, line 63:
```suggestion
     kIsOwnerComp = BIT(19) // if TMessage owns fBufComp
```

dpiparo:
I stopped the build given the many failures. Happy to restart any time!

net/net/src/TMessage.cxx
ferdymercury: Moved to https://github.com/root-project/root/pull/17628

PR #17603, opened 2025-02-03
[tree] Clarify ownership of TBranchProxy data member

vepadulano:
The fCollection data member is an owning pointer to a TVirtualCollectionProxy. Its lifetime was not properly tracked in the class, leaving space for memory leaks. Change the data member to unique_ptr so the memory management is done automatically.

The memory leaks were found via asan, running the `imt_parTreeProcessing.C` tutorial which has recently started failing in the CI.

Note that this is only fixing *some* of the memory leaks I found, not all. The fix for the rest seems to be non-trivial, but I believe it is still worth checking these changes in isolation.

Related to https://github.com/root-project/root/pull/17594 . I still believe we should use TTreeReaderArray for the tutorial, but let's see how this looks.

### DIFF
--- a/tree/treeplayer/inc/TBranchProxy.h
+++ b/tree/treeplayer/inc/TBranchProxy.h
@@ -25,6 +25,7 @@
 #include <algorithm>
 #include <string>
 #include <iostream>
+#include <memory>
 
 class TBranch;
 class TStreamerElement;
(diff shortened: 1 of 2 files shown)


pcanal (approved):
LGTM. (on a side note I do not expect this to fix the failure in ` imt_parTreeProcessing.C`)

vepadulano:
> (on a side note I do not expect this to fix the failure in  imt_parTreeProcessing.C)

Not sure about that, but looking at all the CI failures they look like they're very much related with the not-so-trivial memleaks I mentioned in the description. This will need a deeper investigation, meanwhile I will close this PR as it is clearly not improving the situation.

PR #17542, opened 2025-01-27
[io] Apply TFile::k630forwardCompatibility when creating new file if set in rootrc

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/16189

PR #17606, opened 2025-02-04
[BP6.32] [io] Apply TFile::k630forwardCompatibility when creating new file if set in rootrc

ferdymercury:
https://github.com/root-project/root/pull/17542#issuecomment-2631958782


pcanal (approved):
Thanks.

PR #17608, opened 2025-02-04
xroofit update

will-cern:
Updates xroofit and slightly improves the coverage of the test/example


guitargeek (approved):
Thank you very much for the upgrades!

PR #17609, opened 2025-02-04
[ntuple] Merger: don't abort if a file doesn't contain a RNTuple

silverweed:
The file without the RNTuple will simply be skipped. This mirrors the merge behavior of other ROOT objects. 
Note that if the user gave the -ff or -fk options ("use first source compression") the compression will be taken from the first RNTuple found.

## Checklist:

- [x] tested changes locally


hahnjo:
In principle LGTM, we probably want a test for this, in `roottest`? Also again a very picky comment (sorry): our own [`CONTRIBUTING.md`](https://github.com/root-project/root/blob/master/CONTRIBUTING.md#your-commit-message) says we should use [present imperative mood](https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?id=HEAD#n239), so the commit message should be "Merger: skip files without RNTuple".

silverweed:
> our own [`CONTRIBUTING.md`](https://github.com/root-project/root/blob/master/CONTRIBUTING.md#your-commit-message) says we should use [present imperative mood](https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?id=HEAD#n239), so the commit message should be "Merger: skip files without RNTuple".

But "don't abort" is present imperative; am I missing something?

hahnjo:
> But "don't abort" is present imperative

You are right. I thought we also prefer positive message instead negations (well, at least I do), but then please disregard.

jblomer (approved):
LGTM; please add a test, I'd suggest a unit test in ntuple_merger using `TFileMerger`.

jblomer (approved):

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -158,9 +158,8 @@ try {
       TFile *inFile = dynamic_cast<TFile *>(pitr);
       ROOT::RNTuple *anchor = inFile ? inFile->Get<ROOT::RNTuple>(ntupleName.c_str()) : nullptr;
       if (!anchor) {
-         Error("RNTuple::Merge", "Failed to retrieve RNTuple anchor named '%s' from file '%s'", ntupleName.c_str(),
-               inFile->GetName());
-         return -1;
+         Info("RNTuple::Merge", "No RNTuple anchor named '%s' from file '%s'", ntupleName.c_str(), inFile->GetName());
```
pcanal, line 161: For the record, I don't think we issue any message for the similar situation for `TTree`.

PR #4615, opened 2019-11-05
[hadd] [WIP] Adding posibility to only hadd listed directories

Daniel-Dane:
hadd (and TFileMerger) has been missing the possibility to only hadd certain directories (and their content) of a bunch of root files. This is my attempt.

PR #17612, opened 2025-02-04
[hadd] add option to filter objects/directories to merge

silverweed:
# This Pull request:
depends on #17609. 
It's a new take on #4615, which adds two flags to hadd:
* `-L <FILE>`: specify a file that contains object or directory names, one per line, to be used as a filter. The filter type is specified by the second flag:
* `-Ltype <OnlyListed|SkipListed>`: determines whether the list of object names contained in FILE should be used as a whitelist or a blacklist to know which objects to merge. This internally uses `TFileMerger::AddObjectNames`, so it has the same restrictions, notably you cannot specify a subobject name such as `dir/object`. If this is a desired feature, it should be added to TFileMerger directly, in a separate PR.

Companion roottest PR: https://github.com/root-project/roottest/pull/1257

## Checklist:

- [x] tested changes locally

### DIFF
--- a/main/src/hadd-argparse.py
+++ b/main/src/hadd-argparse.py
@@ -34,6 +34,16 @@ def get_argparse():
         "compression setting after fk (for example 206 when using -fk206)"), action = 'store_true')
     parser.add_argument("-ff", help="The compression level use is the one specified in the first input", action = 'store_true')
     parser.add_argument("-k", help="Skip corrupt or non-existent files, do not exit", action = 'store_true')
+    parser.add_argument("-L", help=textwrap.fill(
+       "Read the list of objects from FILE and either only merge or skip those objects depending on "
+       "the value of \"-Ltype\". FILE must contain one object name per line, which cannot contain "
+       "whitespaces or '/'. You can also pass TDirectory names, which apply to the entire directory "
+       "content. Lines beginning with '#' are ignored. If this flag is passed, \"-Ltype\" MUST be "
+       "passed as well."), action = 'store_true')
+    parser.add_argument("-Ltype", help=textwrap.fill(
+        "Sets the type of operation performed on the objects listed in FILE given with the "
+        "\"-L\" flag. \"SkipListed\" will skip all the listed objects; \"OnlyListed\" will only merge those "
+        "objects. If this flag is passed, \"-L\" must be passed as well."), action = 'store_true')
     parser.add_argument("-O", help="Re-optimize basket size when merging TTree", action = 'store_true')
     parser.add_argument("-T", help="Do not merge Trees", action = 'store_true')
     parser.add_argument("-v", help=textwrap.fill(
--- a/main/src/hadd.cxx
+++ b/main/src/hadd.cxx
@@ -54,6 +54,14 @@
   \param -j [N_JOBS]   Parallelise the execution in `N_JOBS` processes. If the number of processes is not specified,
                        or is 0, use the system maximum.
   \param -k            Skip corrupt or non-existent files, do not exit
+  \param -L <FILE>     Read the list of objects from FILE and either only merge or skip those objects depending on
+                       the value of "-Ltype". FILE must contain one object name per line, which cannot contain
+                       whitespaces or '/'. You can also pass TDirectory names, which apply to the entire directory
+                       content. Lines beginning with '#' are ignored. If this flag is passed, "-Ltype" MUST be
+                       passed as well.
+  \param -Ltype <SkipListed|OnlyListed> Sets the type of operation performed on the objects listed in FILE given with the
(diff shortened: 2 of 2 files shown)


main/src/hadd.cxx
```diff
@@ -606,6 +674,10 @@ int main(int argc, char **argv)
    if (args.fCacheSize)
       Info() << "Using " << cacheSize << "\n";
 
+   ////////////////////////////// end flags processing /////////////////////////////////
+
+   gSystem->Load("libTreePlayer");
```
jblomer, line 679: Why do we need this library?

jblomer (commented):
LGTM but should have a corresponding test in roottest.

main/src/hadd.cxx
silverweed: Not sure, it was there already (I just moved it past the flags processing in this PR). Perhaps @pcanal knows it.

pcanal: The dependency was originally a explicit link added to "allow support for TTree with a build index" and then was moved to a run-time dependency because the usage is hidden being a layer of indirect and was dropped in some case by the remove unused dependency feature of the linker (or something like that).

The original commit ended with "until hadd can use the autoloader" which is likely the case now.

So it probably can be technically removed.

 A simple test would to create  a TTree, Create and index, save it to the file, merge the result a couple of times and check the results.

main/src/hadd-argparse.py
```diff
@@ -34,6 +34,16 @@ def get_argparse():
         "compression setting after fk (for example 206 when using -fk206)"), action = 'store_true')
     parser.add_argument("-ff", help="The compression level use is the one specified in the first input", action = 'store_true')
     parser.add_argument("-k", help="Skip corrupt or non-existent files, do not exit", action = 'store_true')
+    parser.add_argument("-L", help=textwrap.fill(
+       "Read the list of objects from FILE and either only merge or skip those objects depending on "
```
pcanal, line 38: This is a powerful new features but does not match the title of the PR.

The title talks about filtering which files are being merged together while the feature documented here and implemented below is about filtering which objects and `TDirectories` within the files are actually merged (but does not affect the list of files opened).

main/src/hadd-argparse.py
```diff
@@ -34,6 +34,16 @@ def get_argparse():
...
+       "the value of \"-Ltype\". FILE must contain one object name per line, which cannot contain "
+       "whitespaces or '/'. You can also pass directory names, which apply to the entire directory "
```
pcanal, line 40:
```suggestion
       "whitespaces or '/'. You can also pass TDirectory names, which apply to the entire directory "
```

main/src/hadd-argparse.py
```diff
@@ -34,6 +34,16 @@ def get_argparse():
...
+       "content. Lines beginning with '#' are ignored. If this flag is passed, \"-Ltype\" MUST be "
+       "passed as well."), action = 'store_true')
+    parser.add_argument("-Ltype", help=textwrap.fill(
+        "Sets the type of operation performed on the files listed in FILE given with the "
```
pcanal, line 44:
```suggestion
        "Sets the type of operation performed on the objects listed in FILE given with the "
```

main/src/hadd.cxx
```diff
@@ -54,6 +54,14 @@
   \param -j [N_JOBS]   Parallelise the execution in `N_JOBS` processes. If the number of processes is not specified,
                        or is 0, use the system maximum.
   \param -k            Skip corrupt or non-existent files, do not exit
+  \param -L <FILE>     Read the list of objects from FILE and either only merge or skip those objects depending on
+                       the value of "-Ltype". FILE must contain one object name per line, which cannot contain
+                       whitespaces or '/'. You can also pass directory names, which apply to the entire directory
+                       content. Lines beginning with '#' are ignored. If this flag is passed, "-Ltype" MUST be
+                       passed as well.
+  \param -Ltype <SkipListed|OnlyListed> Sets the type of operation performed on the files listed in FILE given with the
```
pcanal, line 62:
```suggestion
  \param -Ltype <SkipListed|OnlyListed> Sets the type of operation performed on the objects listed in FILE given with the
```

main/src/hadd-argparse.py
pcanal: See `tutorials/io/mergeSelective.C`

silverweed: @pcanal the PR title was indeed wrong; I changed "files" to "objects".

main/src/hadd.cxx
silverweed: I guess we want to do this in a separate PR

silverweed:
@pcanal anything else left to do in this PR?

pcanal (approved):
LGTM.

Issue #17611, opened 2025-02-04
Regression when using `TTreeProcessorMT` with ImplicitMT and Davix => 0.8.8


guitargeek:
### Description

When building ROOT against the recent [davix 0.8.8](https://github.com/cern-fts/davix/releases), processing a remotely-read TTree with `ROOT::TTreeProcessorMT`crashes.

This crash was first seen in the [imt_parTreeProcessing.C](https://github.com/root-project/root/blob/master/tutorials/io/tree/imt_parTreeProcessing.C) tutorial.

### Reproducer

```c++
#include <ROOT/TThreadedObject.hxx>
#include <ROOT/TTreeProcessorMT.hxx>

#include <TROOT.h>

void repro()
{
   ROOT::EnableImplicitMT(12); // the higher the number, the more likely to see the crash
   // There is no problem when you don't get the file from the web!
   ROOT::TTreeProcessorMT tp("http://root.cern/files/tp_process_imt.root", "events");
   tp.Process([](TTreeReader &myReader) {});
}
```

### ROOT version

ROOT `master`.

### Installation method

* reproduced when using davix 0.8.8 on the system.
* also reproduced with builtin davix after hacking the builtin davix version to 0.8.8

### Operating system

Linux

### Additional context

Can be seen [occasionally on the CI as well](https://github.com/root-project/root/actions/runs/13094677538/job/36535505575#step:13:28025).

Example stack trace of the crash (using builtin davix 0.8.8). Note that it might also crash at completely different places!
```gdb
Thread 6 "repro" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffdb8296c0 (LWP 319007)]
0x00007ffff46a6658 in _int_malloc () from /nix/store/h7zcxabfxa7v5xdna45y2hplj31ncf8a-glibc-2.40-36/lib/libc.so.6
(gdb) bt
#0  0x00007ffff46a6658 in _int_malloc ()
   from /nix/store/h7zcxabfxa7v5xdna45y2hplj31ncf8a-glibc-2.40-36/lib/libc.so.6
#1  0x00007ffff46a756d in malloc () from /nix/store/h7zcxabfxa7v5xdna45y2hplj31ncf8a-glibc-2.40-36/lib/libc.so.6
#2  0x00007fffdc95ce5a in ne_strdup (s=0x7fffcc00d00d "Partial Content")
[...]

PR #17614, opened 2025-02-04
[davix] Fix problems with locks and mutexes in `TDavixFile` classes

guitargeek:
See the commit descriptions for more detail.

Closes https://github.com/root-project/root/issues/17611, verified by running the reproducer 200 times in the
Fedora 40 container with this PR, without a single crash.

The new davix version seems to be more prone to concurrency issues.

### DIFF
--- a/net/davix/src/TDavixFile.cxx
+++ b/net/davix/src/TDavixFile.cxx
@@ -76,7 +76,10 @@ const char* open_mode_create = "CREATE";
 const char* open_mode_new = "NEW";
 const char* open_mode_update = "UPDATE";
 
-static TMutex createLock;
+static TMutex davixLockCreate;
+static TMutex davixLockPosition;
+static TMutex davixLockOpen;
+
 static Context* davix_context_s = NULL;
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -352,7 +355,7 @@ TDavixFileInternal::~TDavixFileInternal()
 Context *TDavixFileInternal::getDavixInstance()
 {
    if (davix_context_s == NULL) {
-      TLockGuard guard(&createLock);
+      TLockGuard guard(&davixLockCreate);
       if (davix_context_s == NULL) {
          davix_context_s = new Context();
       }
@@ -709,7 +712,7 @@ TString TDavixFile::GetNewUrl() {
 
 void TDavixFile::Seek(Long64_t offset, ERelativeTo pos)
 {
-   TLockGuard guard(&(d_ptr->positionLock));
+   TLockGuard guard(&davixLockPosition);
    switch (pos) {
       case kBeg:
          fOffset = offset + fArchiveOffset;
@@ -736,7 +739,7 @@ void TDavixFile::Seek(Long64_t offset, ERelativeTo pos)
 
 Bool_t TDavixFile::ReadBuffer(char *buf, Int_t len)
 {
-   TLockGuard guard(&(d_ptr->positionLock));
+   TLockGuard guard(&davixLockPosition);
    Davix_fd *fd;
    if ((fd = d_ptr->getDavixFileInstance()) == NULL)
       return kTRUE;
@@ -840,7 +843,7 @@ void TDavixFile::enableGridMode()
 
 bool TDavixFileInternal::isMyDird(void *fd)
 {
-   TLockGuard l(&(openLock));
+   TLockGuard l(&davixLockOpen);
    std::vector<void *>::iterator f = std::find(dirdVec.begin(), dirdVec.end(), fd);
    return (f != dirdVec.end());
 }
@@ -849,7 +852,7 @@ bool TDavixFileInternal::isMyDird(void *fd)
 
 void TDavixFileInternal::addDird(void *fd)
 {
-   TLockGuard l(&(openLock));
+   TLockGuard l(&davixLockOpen);
    dirdVec.push_back(fd);
 }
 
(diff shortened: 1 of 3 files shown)


net/davix/src/TDavixFileInternal.h
```diff
@@ -123,19 +101,16 @@ class TDavixFileInternal {
...
    // DAVIX
    Davix::Context *davixContext;
-   Davix::RequestParams *davixParam;
-   Davix::DavPosix *davixPosix;
-   Davix_fd *davixFd;
+   Davix::RequestParams *davixParam = nullptr;
+   Davix::DavPosix *davixPosix = nullptr;
+   Davix_fd *davixFd = nullptr;
    TUrl fUrl;
    Option_t* opt;
```
pcanal, line 112: Is this initialization left unchanged on purpose?

net/davix/src/TDavixFile.cxx
```diff
@@ -736,7 +737,7 @@ void TDavixFile::Seek(Long64_t offset, ERelativeTo pos)
 
 Bool_t TDavixFile::ReadBuffer(char *buf, Int_t len)
 {
-   TLockGuard guard(&(d_ptr->positionLock));
+   TLockGuard guard(&davixLock);
```
pcanal, line 740: Isn't this needed also in the next function line 759: `Bool_t TDavixFile::ReadBuffer(char *buf, Long64_t pos, Int_t len)` ?

pcanal: and/or why here and not every single other function :( ?

pcanal: Also the lock scope needs to be restricted.  The line 748/749:
```
Info("ReadBuffer", "%lld bytes of data read sequentially"
```
can take the ROOT global lock and thus can lead to dead lock if it is called within another locks.

net/davix/src/TDavixFileInternal.h
guitargeek: Yes, because this data member is initialized in the constructor anyway.

net/davix/src/TDavixFile.cxx
guitargeek: Thanks a lot for the comments! The last commit is addressing these.

> and/or why here and not every single other function :( ?

Well, I haven't looked into this in all detail. I just tried to add the minimal amount of locks to "never" hit the race condition in the unit test where we saw the crash.

net/davix/src/TDavixFile.cxx
```diff
@@ -736,11 +737,11 @@ void TDavixFile::Seek(Long64_t offset, ERelativeTo pos)
 
 Bool_t TDavixFile::ReadBuffer(char *buf, Int_t len)
 {
-   TLockGuard guard(&(d_ptr->positionLock));
    Davix_fd *fd;
    if ((fd = d_ptr->getDavixFileInstance()) == NULL)
       return kTRUE;
-   Long64_t ret = DavixReadBuffer(fd, buf, len);
+   Long64_t ret = DavixReadBuffer(fd, buf, fOffset, len);
+   fOffset += std::max((Long64_t)0, ret); // only shift offset on successful read
```
vepadulano, line 744: Doesn't this operation also need the lock?

net/davix/src/TDavixFile.cxx
```diff
@@ -908,21 +906,26 @@ void TDavixFile::eventStop(Double_t t_start, Long64_t len, bool read)
...
 {
    DavixError *davixErr = NULL;
    Double_t start_time = eventStart();
 
-   Long64_t ret = d_ptr->davixPosix->pread(fd, buf, len, fOffset, &davixErr);
+   Long64_t ret;
+   {
+      TLockGuard guard(&davixLock);
+      ret = d_ptr->davixPosix->pread(fd, buf, len, pos, &davixErr);
+   }
```
vepadulano, line 918: Consider the following code as a (arguable) stylistic improvement:
```suggestion
    const auto ret = [this] () -> Long64_t {
        TLockGuard guard(&this->davixLock);
        return this->davixPosix->pread(fd, buf, len, pos, &davixErr);
    }();
```

vepadulano (commented):
In principle the changes look good to me, just two minor comments.

net/davix/src/TDavixFile.cxx
vepadulano: Right, I was thinking about the modification of the offset data member, in case another thread is reading it from another method of the class at the same time.

pcanal: >  I was thinking about the modification of the offset data member

It is currently 'not allowed' (i.e. explicitly not thread safe) to use the same `TFile` instance from multiple threads.

guitargeek: Yes, that was also my assumption!

net/davix/src/TDavixFile.cxx
```diff
@@ -641,6 +642,7 @@ Int_t TDavixFileInternal::DavixStat(const char *url, struct stat *st)
 {
    DavixError *davixErr = NULL;
 
+   TLockGuard guard(&davixLock);
    if (davixPosix->stat(davixParam, url, st, &davixErr) < 0) {
```
pcanal, line 646:
```suggestion
   int stat_result = -1;
   { 
      TLockGuard guard(&davixLock);
      stat = davixPosix->stat(davixParam, url, st, &davixErr);
   }
   if ( stat_result < 0) {
```
because of the error message.

net/davix/src/TDavixFile.cxx
pcanal: Without additional information from the Davix internals, if the use of `dptr`/`dptr->davixPosix` needs a lock here, I have to conclude that it needs it every where :(

dpiparo:
We should wait for a few days: a thread safe patch of davix might be coming.

guitargeek:
Davix 0.8.10 was released with a fix

PR #17616, opened 2025-02-04
[DF] Check if column exists on `GetTypeName`

enirolf:
This PR adds a check to `RNTupleDS::GetTypeName` for the existence of the provided column name, and throws a `std::runtime_error` if this is not the case.

To simplify, we could also use `RNTupleDS::HasColumn`, but that would mean calling `std::find` on `fColumnNames` twice, so I opted for this slightly duplicated approach instead.

### DIFF
--- a/tree/dataframe/src/RNTupleDS.cxx
+++ b/tree/dataframe/src/RNTupleDS.cxx
@@ -684,7 +684,14 @@ void RNTupleDS::FinalizeSlot(unsigned int slot)
 
 std::string RNTupleDS::GetTypeName(std::string_view colName) const
 {
-   const auto index = std::distance(fColumnNames.begin(), std::find(fColumnNames.begin(), fColumnNames.end(), colName));
+   auto colNamePos = std::find(fColumnNames.begin(), fColumnNames.end(), colName);
+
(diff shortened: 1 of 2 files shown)


tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -684,7 +684,14 @@ void RNTupleDS::FinalizeSlot(unsigned int slot)
 
 std::string RNTupleDS::GetTypeName(std::string_view colName) const
 {
-   const auto index = std::distance(fColumnNames.begin(), std::find(fColumnNames.begin(), fColumnNames.end(), colName));
+   auto colNamePos = std::find(fColumnNames.begin(), fColumnNames.end(), colName);
+
+   if (colNamePos == fColumnNames.end()) {
+      auto msg = std::string("There is no column with name \"") + std::string(colName) + "\"";
```
vepadulano, line 690: Usually we provide some location info for this type of error messages, e.g.
```
RNTupleDS: There is no column with name...
```

vepadulano (approved):
Thanks, I just left a minor suggestion :+1:

PR #17627, opened 2025-02-05
[cppyy] Sync `cppyy_backend/loader.py` with upstream

guitargeek:
Reverts commit 834f8f346305a5b448a2c9681169c4b707d2cfa3 and move setting the hardcoded cppyy backend
path to the ROOT module, to get rid of a patch.

Sync the `load_cpp_backend` function with upsteam while at it.


vepadulano (approved):
Nice! Feel free to merge if CI agrees

bindings/pyroot/pythonizations/python/ROOT/__init__.py
```diff
@@ -8,16 +8,23 @@
...
 # Prevent cppyy's check for extra header directory
-environ["CPPYY_API_PATH"] = "none"
+os.environ["CPPYY_API_PATH"] = "none"
 
 # Prevent cppyy from filtering ROOT libraries
-environ["CPPYY_NO_ROOT_FILTER"] = "1"
+os.environ["CPPYY_NO_ROOT_FILTER"] = "1"
+
+# For macOS, need to load cppyy backend library with absolute path (see also PR #9725)
+if latform.system() == "Darwin":
```
vepadulano, line 24:
```suggestion
if platform.system() == "Darwin":
```

vepadulano (changes requested):

bindings/pyroot/pythonizations/python/ROOT/__init__.py
guitargeek: Ah, damn :laughing: Thanks for the comment

bindings/pyroot/pythonizations/test/rdataframe_misc.py
```diff
@@ -1,4 +1,3 @@
-import cppyy
```
vepadulano, line 1: Is this change related to the title of PR?

guitargeek: No, that was just something that I tried out for the CI.

But now I understand that this "macos12soversion" patch is still necessary.

I propose however to still sync the `loader.py` file now that we already have taken a look. The PR was updated to do that.

vepadulano (approved):
Even with the current changes, it's still an improvement, thanks!

PR #17628, opened 2025-02-05
TMessage honour kIsOwner bit when compression is used

ferdymercury:
# This Pull request:

## Changes or fixes:

Rebased version of https://github.com/root-project/root/pull/17601

Fixes https://github.com/root-project/root/issues/14557 reported by @cpulvermacher

### DIFF
--- a/net/net/src/TMessage.cxx
+++ b/net/net/src/TMessage.cxx
@@ -111,7 +111,8 @@ TMessage::TMessage(void *buf, Int_t bufsize, Bool_t adopt) : TBufferFile(TBuffer
 TMessage::~TMessage()
 {
    // We only own fBufComp when we explictly created it or adopt and kMESS_ZIP were true
-   if (TestBit(kIsOwnerComp)) delete [] fBufComp;
+   if (TestBit(kIsOwnerComp))
+      delete [] fBufComp;
    delete fInfos;
 }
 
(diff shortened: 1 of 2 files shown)


net/net/src/TMessage.cxx
```diff
@@ -88,6 +88,11 @@ TMessage::TMessage(void *buf, Int_t bufsize) : TBufferFile(TBuffer::kRead, bufsi
       fBufCompCur = fBuffer + bufsize;
       fBuffer     = nullptr;
       Uncompress();
+      if (adopt) {
+         SetBit(kIsOwnerComp); // bufcomp points to the original buf
+      }
+      // Force being owner of the newly created buffer (inside Uncompress)
+      SetBit(kIsOwner);
```
pcanal, line 95: This should be move closer to the allocation (technically that allocation is not quite unconditional and may be reached from other path), so inside `Uncompress`.

net/net/src/TMessage.cxx
```diff
@@ -105,7 +110,8 @@ TMessage::TMessage(void *buf, Int_t bufsize) : TBufferFile(TBuffer::kRead, bufsi
 
 TMessage::~TMessage()
 {
-   delete [] fBufComp;
+   // We only own fBufComp when we explictly created it or adopt and kMESS_ZIP were true
+   if (TestBit(kIsOwnerComp)) delete [] fBufComp;
```
pcanal, line 114:
```suggestion
   // We only own fBufComp when we explicitly created it or adopt and kMESS_ZIP were true
   if (TestBit(kIsOwnerComp))
      delete [] fBufComp;
```

pcanal (approved):
LGTM.  Thanks.

cpulvermacher:
You guys made my day, thank you! :slightly_smiling_face:

Issue #17470, opened 2025-01-21
Unwanted black line at y == 0 in TMultiGraph with TH1::SetDefaultSumw2()


Zehvogel:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

As the title says.

Probably just missed this case in this fix https://github.com/root-project/root/commit/f8f4760c147828cf0b6ef435556ea88918b6962a

### Reproducer

It is probably possible to build an easier reproducer, I just had this one already ready.
I would guess you can put whichever two (maybe even one?) `TGraphs` into a `TMultiGraph` and it will show up.

https://github.com/Zehvogel/enuW/blob/6770d0e578b729d1a05ed2b1012f7aa39dce6f0f/roo_ellipse.ipynb

### ROOT version

```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.32.04                        https://root.cern |
  | (c) 1995-2024, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Sep 22 2024, 09:03:08                 |
  | From heads/master@tags/v6-32-04                                  |
  | With g++ (Spack GCC) 14.2.0                                      |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------
```

### Installation method

`source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh  -r 2025-01-20`

### Operating system

Alma9

dpiparo:
Adding @couet in the loop, as he was the author of the commit you mention.

linev:
Here reproducer with plain macro.

[mg.cxx.txt](https://github.com/user-attachments/files/18673335/mg.cxx.txt)

With web canvas and ROOT master branch macro works now.
Also in 6.34 branch `RooEllipse` class will be working now in web canvas

PR #17631, opened 2025-02-05
Fix drawing of `TMultiGraph` axes

linev:
Reset Sumw2 array of histogram to avoid baseline drawing at position 0.

Solves #17470


couet (approved):
lgtm

linev:
Use of "AXIS" is not correct, title and grid is not drawn in this case.
I need to check again.

linev:
I rebase PR and now simply reset Sumw2 array of histogram

PR #17633, opened 2025-02-05
[cppyy] Update `GetFunctionAddress` in cppyy backend

guitargeek:
Sync the implementation of `GetFunctionAddress` with the latest version
in upstream cppyy-backend. This means we can remove the patch file with
the by-now outdated implementation, and revert the addition of some ROOT
meta interfaces.

The change is covered by the numba unit tests.

Partially reverts https://github.com/root-project/root/commit/9001fd29cb0088786b52303cac01feeda1b4a857 from https://github.com/root-project/root/pull/11096.


core/meta/inc/TInterpreter.h
```diff
@@ -559,9 +559,6 @@ class TInterpreter : public TNamed {
    virtual Bool_t IsPointerType(const void * /* QualTypePtr */) const {return 0;}
    virtual Bool_t IsVoidPointerType(const void * /* QualTypePtr */) const {return 0;}
 
-   // FunctionDecl interface
-   virtual Bool_t FunctionDeclId_IsMethod(DeclId_t /* fdeclid */) const {return 0;}
```
pcanal, line 563: This is in the public.  Technically (but unlikely) this might be used by some users.

guitargeek: You're right, but given that it was only introduced 2 years ago for the sake of cppyy, I doubt that. We still have the opportunity to remove this public interface again I'd say, and slimming the interface is always good

vepadulano (approved):
Nice thanks!

PR #17635, opened 2025-02-05
[core] Remove useless dictionaries

dpiparo:


core/unix/inc/TUnixSystem.h
```diff
@@ -216,7 +216,6 @@ class TUnixSystem : public TSystem {
    int               GetMemInfo(MemInfo_t *info) const override;
    int               GetProcInfo(ProcInfo_t *info) const override;
 
-   ClassDefOverride(TUnixSystem,0)  //Interface to Unix OS services
```
pcanal, line 219: I am confused.  This means that at run-time we can no longer distinguish between `TSystem` and `TUnixSystem` `(at least not via Core/Meta interfaces).   What is the motivation for this changes?

dpiparo:
I closed the PR since the improvement is minimal (1 dictionary less), the rationale was
1) If we check on what OS we are through meta, it's probably a rare occurrence since it's not tested anywhere
2) Since we do no IO for that class and we'll never do, see the 0 in the macro, the ClassDef is partially useful for the T*System case

Issue #16754, opened 2024-10-28
[ntuple] Support all template parameters of STL containers


hahnjo:
### Feature description

As pointed out by ATLAS, it's currently not possible to use a `std::vector` with a non-default allocator in RNTuple:
```c++
#include <ROOT/RNTupleModel.hxx>

#include <scoped_allocator>
#include <vector>

void ntuple_vector_allocator() {
  auto model = ROOT::Experimental::RNTupleModel::CreateBare();
  model->MakeField<std::vector<int, std::scoped_allocator_adaptor<std::allocator<int>>>>("v");
}
```
leads to
```
Processing ntuple_vector_allocator.C...
terminate called after throwing an instance of 'ROOT::Experimental::RException'
  what():  vector<int,scoped_allocator_adaptor<allocator<int> > > is not supported
At:
  ROOT::Experimental::RClassField::RClassField(std::string_view, std::string_view, TClass *) [/home/jhahnfel/ROOT/src/tree/ntuple/v7/src/RField.cxx:1807]

Aborted (core dumped)
```
This is because `RField` is only partially specialized for `std::vector<ItemT>`, so a non-default `Allocator` falls back to the default `RField` declaration (which assumes it's a class and checks that it's not in `std` namespace).

Note that in principle, this affects all STL containers. For many of them, there are other template parameters, for example `Hash` for `std::unordered_set`:
```c++
#include <ROOT/RNTupleModel.hxx>

#include <unordered_set>

struct IntHash : public std::hash<int> {};

void ntuple_unordered_set_hash() {
  auto model = ROOT::Experimental::RNTupleModel::CreateBare();
  model->MakeField<std::unordered_set<int, IntHash>>("s");
}
```
leads to
```
Processing ntuple_unordered_set_hash.C...
terminate called after throwing an instance of 'ROOT::Experimental::RException'
  what():  unordered_set<int,IntHash,equal_to<int>,allocator<int> > is not supported
At:
  ROOT::Experimental::RClassField::RClassField(std::string_view, std::string_view, TClass *) [/home/jhahnfel/ROOT/src/tree/ntuple/v7/src/RField.cxx:1807]

Aborted (core dumped)
```

hahnjo:
As discussed, for `std::vector` with a non-default allocator we should likely go through the collection proxy support. For maps and sets, we may already support (via the collection proxies) and it's just a matter of relaxing the partial specializations and maybe the logic in `RFieldBase::Create`.

[...]

PR #17637, opened 2025-02-05
[ntuple] fix type name normalization

### LINKED ISSUE #17570: [ntuple] Missing type name normalization for meta fields
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Fields of user-defined types in the "meta" category (RClassField, REnumField, RProxiedCollectionField, RStreamerField) don't normalize the provided type name. As a result, the on-disk type can differ depending on how the field is constructed (`MakeField<T>` or `RFieldBase::Create()`). Also, the `RFieldBase::Create()` API does now include default template parameters.

### Reproducer

For a type
```
template <typename FirstT, typename SecondT = double>
class DataVector {
public:
   FirstT fFirst;
   SecondT fSecond;
};
```

the on-disk type of
```
model->MakeField<DataVector<int>>("f1");
```
differs from
```
model->AddField(RFieldBase::Create("f2", "DataVector<int>").Unwrap());
```

### ROOT version

master, 6.34

### Installation method

n/a

### Operating system

n/a

### Additional context

_No response_

jblomer:
Fixes type name and type alias normalization, especially for types that involve meta (TClass etc.). Introduces patch version 1.0.0.1 of the specification.

Fixes #17570

### DIFF
--- a/tree/ntuple/v7/doc/BinaryFormatSpecification.md
+++ b/tree/ntuple/v7/doc/BinaryFormatSpecification.md
@@ -1,4 +1,4 @@
-# RNTuple Binary Format Specification 1.0.0.0
+# RNTuple Binary Format Specification 1.0.0.1
 
 ## Versioning Notes
 
@@ -150,9 +150,9 @@ Note that this type system is independent (and different) from the regular ROOT 
 _Integer_: Integers are encoded in two's complement, little-endian format.
 They can be signed or unsigned and have lengths up to 64bit.
 
-_String_: A string is stored as a 32bit unsigned integer indicating the length of the string
+_String_: A string is stored as a 32bit unsigned integer indicating the number of bytes of the string
 followed by the characters.
-Strings are ASCII encoded; every character is a signed 8bit integer.
+String data are UTF-8 encoded.
 
 _Compression settings_: A 32bit integer containing both a compression algorithm and the compression level.
 The compression settings are encoded according to this formula: $settings = algorithm * 100 + level$.
@@ -802,6 +802,30 @@ This section is a comprehensive list of the C++ types with RNTuple I/O support.
 Within the supported type system complex types can be freely composed,
 e.g. `std::vector<MyEvent>` or `std::vector<std::vector<float>>`.
 
+### Type Name Normalization
+
+Type names are stored according to the following normalization rules
+  - The type name of a field has typedefs and usings fully resolved (except for the following rule).
+  - The integer types `signed char`, `unsigned char`, and `[signed|unsigned](short|int|long[ long])`
+    are replaced by the corresponding (at the time of writing) `std::[u]int(8|16|32|64)_t` standard integer typedef.
+  - Qualifiers `volatile` and `const` that do not appear in template arguments are removed.
+  - The `class`, `struct`, and `enum` keywords are removed.
+  - Type names are fully qualified by the namespace in which they are declared;
+    the root namespace ('::' prefix) is stripped.
+  - Type names used as template arguments are themselves normalized.
+  - For user-defined types, default-initialized template arguments are explicitly spelled out;
+    optional template arguments of stdlib types such as allocators are omitted.
+  - Integer template arguments are written in standard decimal notation;
+    only if integer values are larger than 2^63, the suffix `u` (`unsigned long long`) is added.
+  - Only whitespaces strictly required by C++ semantic are kept.
+
+Both the field type name and the field alias type name are normalized,
+except that the field alias type name does not apply any type name resolution.
+
+For example, the type name `const pair<size_t, array<class ::Event, 2>>` will be normalized on 64bit architectures to
+`std::pair<std::uint64_t,std::array<Event,2>>` with an alias type name
+`std::pair<size_t,std::array<Event,2>>`.
+
 ### Fundamental Types
 
 The following fundamental types are stored as `leaf` fields with a single column each.
@@ -829,7 +853,6 @@ Such cases are marked as `R` in the table.
 | Real32Trunc                        |        |             |        |          |           |           |            |           |            |           |            |    W    |    W     |
 | Real32Quant                        |        |             |        |          |           |           |            |           |            |           |            |    W    |    W     |
 
-Possibly available `const` and `volatile` qualifiers of the C++ types are ignored for serialization.
 The default column for serialization is denoted with an asterisk.
 If the ntuple is stored uncompressed, the default changes from split encoding to non-split encoding where applicable.
 
@@ -954,6 +977,8 @@ User defined C++ classes are supported with the following limitations
   - There is no support for polymorphism,
     i.e. a field of class `A` cannot store class `B` that derives from `A`.
   - Virtual inheritance is unsupported.
+  - Bit-field declarations (data member with an explicit size in bits) are unsupported.
+  - Template arguments of templated classes are restricted to integers and types that have RNTuple type support.
 
 User classes are stored as a record parent field with no attached columns.
 Direct base classes and persistent members are stored as subfields with their respective types.
@@ -994,6 +1019,8 @@ The valid types are `std::uint32_t` and `std::uint64_t`.
 A field with the structural role 0x04 ("streamer") represents an object serialized by the ROOT streamer
 into a single `Byte` column.
 It can have any type supported by `TClass` (even types that are not available in the native RNTuple type system).
+For templated types, however,
+the restrictions on template arguments (only RNTuple supported types and integers) still apply.
 The first (principal) column is of type `(Split)Index[32|64]`.
 The second column is of type `Byte`.
 In effect, the column representation is identical to a collection of `std::byte`.
--- a/tree/ntuple/v7/src/RFieldBase.cxx
+++ b/tree/ntuple/v7/src/RFieldBase.cxx
@@ -98,11 +98,11 @@ void ROOT::Experimental::Internal::CallConnectPageSourceOnField(RFieldBase &fiel
 }
 
 ROOT::RResult<std::unique_ptr<ROOT::Experimental::RFieldBase>>
-ROOT::Experimental::Internal::CallFieldBaseCreate(const std::string &fieldName, const std::string &canonicalType,
-                                                  const std::string &typeAlias, const RCreateFieldOptions &options,
-                                                  const RNTupleDescriptor *desc, ROOT::DescriptorId_t fieldId)
+ROOT::Experimental::Internal::CallFieldBaseCreate(const std::string &fieldName, const std::string &typeName,
+                                                  const RCreateFieldOptions &options, const RNTupleDescriptor *desc,
+                                                  ROOT::DescriptorId_t fieldId)
 {
-   return RFieldBase::Create(fieldName, canonicalType, typeAlias, options, desc, fieldId);
+   return RFieldBase::Create(fieldName, typeName, options, desc, fieldId);
 }
 
 //------------------------------------------------------------------------------
@@ -274,32 +274,18 @@ std::string ROOT::Experimental::RFieldBase::GetQualifiedFieldName() const
 ROOT::RResult<std::unique_ptr<ROOT::Experimental::RFieldBase>>
 ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::string &typeName)
 {
-   auto typeAlias = Internal::GetNormalizedTypeName(typeName);
-   auto canonicalType = Internal::GetNormalizedTypeName(GetCanonicalTypeName(typeAlias));
-   return R__FORWARD_RESULT(RFieldBase::Create(fieldName, canonicalType, typeAlias, RCreateFieldOptions{}));
-}
-
-ROOT::RResult<std::unique_ptr<ROOT::Experimental::RFieldBase>>
-ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::string &typeName,
-                                       const RCreateFieldOptions &options, const RNTupleDescriptor *desc,
-                                       ROOT::DescriptorId_t fieldId)
-{
-   auto typeAlias = Internal::GetNormalizedTypeName(typeName);
-   auto canonicalType = Internal::GetNormalizedTypeName(GetCanonicalTypeName(typeAlias));
-   return R__FORWARD_RESULT(RFieldBase::Create(fieldName, canonicalType, typeAlias, options, desc, fieldId));
+   return R__FORWARD_RESULT(
+      RFieldBase::Create(fieldName, typeName, RCreateFieldOptions{}, nullptr, ROOT::kInvalidDescriptorId));
 }
 
 std::vector<ROOT::Experimental::RFieldBase::RCheckResult>
 ROOT::Experimental::RFieldBase::Check(const std::string &fieldName, const std::string &typeName)
 {
-   auto typeAlias = Internal::GetNormalizedTypeName(typeName);
-   auto canonicalType = Internal::GetNormalizedTypeName(GetCanonicalTypeName(typeAlias));
-
    RFieldZero fieldZero;
    RCreateFieldOptions cfOpts{};
    cfOpts.fReturnInvalidOnError = true;
    cfOpts.fEmulateUnknownTypes = false;
-   fieldZero.Attach(RFieldBase::Create(fieldName, canonicalType, typeAlias, cfOpts).Unwrap());
+   fieldZero.Attach(RFieldBase::Create(fieldName, typeName, cfOpts, nullptr, kInvalidDescriptorId).Unwrap());
 
    std::vector<RCheckResult> result;
    for (const auto &f : fieldZero) {
@@ -315,10 +301,13 @@ ROOT::Experimental::RFieldBase::Check(const std::string &fieldName, const std::s
 }
 
 ROOT::RResult<std::unique_ptr<ROOT::Experimental::RFieldBase>>
-ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::string &canonicalType,
-                                       const std::string &typeAlias, const RCreateFieldOptions &options,
-                                       const RNTupleDescriptor *desc, ROOT::DescriptorId_t fieldId)
+ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::string &typeName,
+                                       const RCreateFieldOptions &options, const RNTupleDescriptor *desc,
+                                       ROOT::DescriptorId_t fieldId)
 {
+   const auto typeAlias = Internal::GetNormalizedTypeName(typeName);
+   const auto canonicalType = Internal::GetNormalizedTypeName(GetCanonicalTypeName(typeAlias));
+
    thread_local CreateContext createContext;
    CreateContextGuard createContextGuard(createContext);
    if (options.fReturnInvalidOnError)
--- a/tree/ntuple/v7/src/RFieldBase.cxx
+++ b/tree/ntuple/v7/src/RFieldBase.cxx
@@ -20,16 +20,6 @@
 
 namespace {
 
-/// Resolve usings and typedefs
-std::string GetResolvedTypeName(const std::string &typeName)
-{
-   // The following types are asummed to be resolved; thus, do not perform `typedef` resolution on those
-   if (typeName.substr(0, 5) == "std::" || typeName.substr(0, 25) == "ROOT::RNTupleCardinality<")
-      return typeName;
-
-   return TClassEdit::ResolveTypedef(typeName.c_str());
-}
-
 /// Used as a thread local context storage for Create(); steers the behavior of the Create() call stack
 class CreateContextGuard;
 class CreateContext {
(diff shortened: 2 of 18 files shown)


tree/ntuple/v7/doc/BinaryFormatSpecification.md
```diff
@@ -802,6 +802,27 @@ This section is a comprehensive list of the C++ types with RNTuple I/O support.
 Within the supported type system complex types can be freely composed,
 e.g. `std::vector<MyEvent>` or `std::vector<std::vector<float>>`.
 
+### Type Name Normalization
+
+Type names are stored according to the following normalization rules
+  - All whitespaces are removed.
+  - CV qualifiers (`const`, `volatile`) are removed.
```
pcanal, line 809: Only the outer ones can be safely removed: `usertemplate<const int>` might have a complete different layout than `usertemplate<int>`

tree/ntuple/v7/doc/BinaryFormatSpecification.md
```diff
@@ -802,6 +802,27 @@ This section is a comprehensive list of the C++ types with RNTuple I/O support.
 Within the supported type system complex types can be freely composed,
 e.g. `std::vector<MyEvent>` or `std::vector<std::vector<float>>`.
 
+### Type Name Normalization
+
+Type names are stored according to the following normalization rules
+  - All whitespaces are removed.
```
pcanal, line 808: Not exact phrasing, in part because of https://github.com/root-project/root/pull/17637/files#r1943506781

tree/ntuple/v7/doc/BinaryFormatSpecification.md
```diff
@@ -802,6 +802,27 @@ This section is a comprehensive list of the C++ types with RNTuple I/O support.
...
+  - CV qualifiers (`const`, `volatile`) are removed.
+  - The `class`, `struct`, and `enum` keywords are removed.
+  - Type names are fully qualified by the namespace in which they are declared.
+  - The integer types `signed char`, `unsigned char`, and `[signed|unsigned](short|int|long[ long])` are replaced by
```
pcanal, line 812: What about `cont char*`?

tree/ntuple/v7/doc/BinaryFormatSpecification.md
```diff
@@ -802,6 +802,27 @@ This section is a comprehensive list of the C++ types with RNTuple I/O support.
...
+    the corresponding (at the time of writing) `std::[u]int(8|16|32|64)_t` standard integer typedef
+  - Type names used as template arguments are themselves normalized.
+  - For user-defined types, default-initialized template arguments are explicitly spelled out;
+    optional template arguments of stdlib types such as allocators are omitted.
+  - Integer template arguments are written in standard decimal notation; only if necessary (long integer types),
+    a suffix `l` or `ul` is added according to the template parameter type.
+
+The type name of a field has typedefs and usings fully resolved (except for standard integer typedefs).
```
pcanal, line 820: This may belong sooner in the document as the above mentioned rules are applying to the name with fully resolved typedefs.

tree/ntuple/v7/src/RFieldBase.cxx
```diff
@@ -617,26 +575,39 @@ ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::
...
+      const TClassEdit::EModType modType = static_cast<TClassEdit::EModType>(
+         TClassEdit::kDropStlDefault | TClassEdit::kDropComparator | TClassEdit::kDropHash);
+      std::string normOrigType{typeName};
+      TClassEdit::TSplitType splitname(normOrigType.c_str(), modType);
+      splitname.ShortType(normOrigType, modType);
+      // See TClassEdit::GetNormalizedName
+      if (normOrigType.length() > 2 && normOrigType[0] == ':' && normOrigType[1] == ':') {
+         normOrigType.erase(0, 2);
+      }
+      normOrigType = Internal::GetRenormalizedTypeName(normOrigType);
```
pcanal, line 604: Side note: both `TClassEdit::TSplitType` and `Internal::GetRenormalizedTypeName` are tokenizing the template name.

tree/ntuple/v7/doc/BinaryFormatSpecification.md
jblomer: `char *` and by extension `const char *` are not supported by RNTuple. Users need to use `std::string`.

tree/ntuple/v7/doc/BinaryFormatSpecification.md
jblomer: Hm, I'm not sure I understand. The normalization rules, as far as I understand, should apply to both the original type name and the resolved type name.

tree/ntuple/v7/doc/BinaryFormatSpecification.md
jblomer: Clarified.

tree/ntuple/v7/doc/BinaryFormatSpecification.md
jblomer: Thanks, good point! Fixed.

tree/ntuple/v7/src/RFieldBase.cxx
jblomer: Added a TODO to `TokenizeTypeList`

tree/ntuple/v7/doc/BinaryFormatSpecification.md
pcanal: The replacement of the typedef is an integral type of the 'normalization' of the name.  Without that part, the name are not unique and thus not normalized.  I.e. if you have both `const vector<vector<Int_t> >` and `vector<vector<int32_t, typedef_to_regular_allocator>>` "only" pretty printing the name does not help much.  Similar if the typedef are not replaced, neither would/should the Integer template arguments be calculated/compacted.  In the second example, without resolving the typedef we can not even remove the default template argument.

tree/ntuple/v7/doc/BinaryFormatSpecification.md
```diff
@@ -802,6 +802,27 @@ This section is a comprehensive list of the C++ types with RNTuple I/O support.
 Within the supported type system complex types can be freely composed,
 e.g. `std::vector<MyEvent>` or `std::vector<std::vector<float>>`.
 
+### Type Name Normalization
+
+Type names are stored according to the following normalization rules
+  - All `volatile` qualifiers are removed; `const` qualifiers that do not appear in template arguments are removed.
```
pcanal, line 808: According to this [example](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:13,endLineNumber:14,positionColumn:13,positionLineNumber:14,selectionStartColumn:13,selectionStartLineNumber:14,startColumn:13,startLineNumber:14),source:'%23include+%3Coptional%3E%0A%0Atemplate+%3Ctypename+T%3E%0Astruct+Device+%7B%0A++++int+stack_size+%3D+0%3B%0A++++int+heap_size+%3D+0%3B%0A%7D%3B%0A%0Aint+main()+%7B%0A++++Device%3Cint%3E+d0%3B%0A++++Device%3Cvolatile+int%3E+d1%3B%0A++++Device%3Cconst+int%3E+d2%3B%0A++++//+d1+%3D+d2%3B%0A++++d0+%3D+d1%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:51.145194274028626,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:executor,i:(argsPanelShown:'1',compilationPanelShown:'0',compiler:g132,compilerName:'',compilerOutShown:'0',execArgs:'',execStdin:'',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B17+-Wall+-Wextra+-Wpedantic',overrides:!(),runtimeTools:!(),source:1,stdinPanelShown:'1',tree:0,wrap:'1'),l:'5',n:'0',o:'Executor+x86-64+gcc+13.2+(C%2B%2B,+Editor+%231)',t:'0')),header:(),k:48.85480572597136,l:'4',m:100,n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4). 
the types are also different for volatile being part of the template arguments.

tree/ntuple/v7/doc/BinaryFormatSpecification.md
```diff
@@ -802,6 +802,27 @@ This section is a comprehensive list of the C++ types with RNTuple I/O support.
...
+  - All whitespaces except between `const` and a template argument are removed.
```
pcanal, line 809:
```suggestion
  - All whitespaces not required by C++ semantic are removed.
```

tree/ntuple/v7/doc/BinaryFormatSpecification.md
```diff
@@ -802,6 +802,27 @@ This section is a comprehensive list of the C++ types with RNTuple I/O support.
...
+  - Type names are fully qualified by the namespace in which they are declared.
+  - The integer types `signed char`, `unsigned char`, and `[signed|unsigned](short|int|long[ long])` are replaced by
+    the corresponding (at the time of writing) `std::[u]int(8|16|32|64)_t` standard integer typedef.
+  - Type names used as template arguments are themselves normalized.
+  - For user-defined types, default-initialized template arguments are explicitly spelled out;
+    optional template arguments of stdlib types such as allocators are omitted.
+  - Integer template arguments are written in standard decimal notation; only if necessary (long integer types),
+    a suffix `l` or `ul` is added according to the template parameter type.
+
+The type name of a field has typedefs and usings fully resolved (except for standard integer typedefs).
```
pcanal, line 820:
```suggestion
  - The type name of a field has typedefs and usings fully resolved (except for the following rule).
  - The integer types `signed char`, `unsigned char`, and `[signed|unsigned](short|int|long[ long])` are replaced by
    the corresponding (at the time of writing) `std::[u]int(8|16|32|64)_t` standard integer typedef.
  -`volatile` and `const` qualifiers that do not appear in template arguments are removed.
  - The `class`, `struct`, and `enum` keywords are removed.
  - Type names are fully qualified by the namespace in which they are declared.
  - Type names used as template arguments are themselves normalized.
  - For user-defined types, default-initialized template arguments are explicitly spelled out;
    optional template arguments of stdlib types such as allocators are omitted.
  - Integer template arguments are written in standard decimal notation; only if necessary (long integer types),
    a suffix `l` or `ul` is added according to the template parameter type.
  - Only whitespaces strictly required by C++ semantic are kept.
```

tree/ntuple/v7/doc/BinaryFormatSpecification.md
```diff
@@ -802,6 +802,27 @@ This section is a comprehensive list of the C++ types with RNTuple I/O support.
...
+Type names are stored according to the following normalization rules
+  - All `volatile` qualifiers are removed; `const` qualifiers that do not appear in template arguments are removed.
+  - All whitespaces except between `const` and a template argument are removed.
+  - The `class`, `struct`, and `enum` keywords are removed.
+  - Type names are fully qualified by the namespace in which they are declared.
+  - The integer types `signed char`, `unsigned char`, and `[signed|unsigned](short|int|long[ long])` are replaced by
+    the corresponding (at the time of writing) `std::[u]int(8|16|32|64)_t` standard integer typedef.
+  - Type names used as template arguments are themselves normalized.
+  - For user-defined types, default-initialized template arguments are explicitly spelled out;
+    optional template arguments of stdlib types such as allocators are omitted.
```
pcanal, line 816: Note that the I/O normalization we have elected to keep the stdlib optional argument whose value is not the default value.  ie.:
```
TClass::GetClass("vector<int, other_alloc>")->GetName()
(const char *) "vector<int,other_alloc>"
```

tree/ntuple/v7/doc/BinaryFormatSpecification.md
jblomer: If the user passes us `vector<vector<int32_t, typedef_to_regular_allocator>>`, its "pretty printed" version will be registered as the field's alias type, i.e. the "original spelling" from the user. Otherwise (`vector<vector<int>>`), the alias type is empty.

tree/ntuple/v7/doc/BinaryFormatSpecification.md
jblomer: Ok, I think that should be solved when fixing #16754

tree/ntuple/v7/doc/BinaryFormatSpecification.md
pcanal: For the original spelling, I guess we ought to decide to either really put it as-is or we need to spell out, separately from the normalization, what applies there.
For example, it is clear that the 'whitespace' rule would apply,  maybe the `const` rule and what else?  (Not the `typedef` for sure, what about the `std::[u]int(8|16|32|64)_t` or even the fully qualified)

tree/ntuple/v7/src/RFieldBase.cxx
```diff
@@ -20,16 +20,6 @@
...
-/// Resolve usings and typedefs
-std::string GetResolvedTypeName(const std::string &typeName)
-{
-   // The following types are asummed to be resolved; thus, do not perform `typedef` resolution on those
-   if (typeName.substr(0, 5) == "std::" || typeName.substr(0, 25) == "ROOT::RNTupleCardinality<")
-      return typeName;
-
-   return TClassEdit::ResolveTypedef(typeName.c_str());
-}
-
```
hahnjo: Can you add some more details to the commit message what was wrong / what is fixed? It's probably in some of the discussion threads of this PR, but I can't find it...

hahnjo (commented):
Some comments inline

tree/ntuple/v7/test/ntuple_types.cxx
```diff
@@ -32,15 +32,38 @@ TEST(RNTuple, TypeName) {
...
+   EXPECT_EQ("", RFieldBase::Create("f", "class CustomStruct").Unwrap()->GetTypeAlias());
+
    EXPECT_EQ("CustomStruct", RFieldBase::Create("f", "struct CustomStruct").Unwrap()->GetTypeName());
+   EXPECT_EQ("", RFieldBase::Create("f", "struct CustomStruct").Unwrap()->GetTypeAlias());
+
    EXPECT_EQ("CustomEnum", RFieldBase::Create("f", "enum CustomEnum").Unwrap()->GetTypeName());
+   EXPECT_EQ("", RFieldBase::Create("f", "enum CustomEnum").Unwrap()->GetTypeAlias());
+
    EXPECT_EQ("std::int32_t", RFieldBase::Create("f", "signed").Unwrap()->GetTypeName());
+   EXPECT_EQ("", RFieldBase::Create("f", "std::int32_t").Unwrap()->GetTypeAlias());
```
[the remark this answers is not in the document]
jblomer: Ah, yes, it should test `signed`. The result is the same (normalized original == normalized resolved, therefore alias type is empty).

hahnjo:
I'm still trying to wrap my head around the type alias commit. I think we should test it more with class templates, like
```diff
diff --git a/tree/ntuple/v7/test/rfield_class.cxx b/tree/ntuple/v7/test/rfield_class.cxx
index 02cbc93ed6..52e1bc3ba4 100644
--- a/tree/ntuple/v7/test/rfield_class.cxx
+++ b/tree/ntuple/v7/test/rfield_class.cxx
@@ -278,6 +278,8 @@ TEST(RNTuple, TClassDefaultTemplateParameter)
       model->MakeField<DataVector<int, float>>("f2");
       model->AddField(RFieldBase::Create("f3", "DataVector<int>").Unwrap());
       model->AddField(RFieldBase::Create("f4", "struct DataVector<bool,vector<unsigned>>").Unwrap());
+      model->AddField(RFieldBase::Create("f5", "DataVector<Double32_t>").Unwrap());
+      model->AddField(RFieldBase::Create("f6", "DataVector<int, double>").Unwrap());
       auto writer = RNTupleWriter::Recreate(std::move(model), "ntpl", fileGuard.GetPath());
    }
 
@@ -286,10 +288,18 @@ TEST(RNTuple, TClassDefaultTemplateParameter)
 
    const auto &desc = reader->GetDescriptor();
    EXPECT_EQ("DataVector<std::int32_t,double>", desc.GetFieldDescriptor(desc.FindFieldId("f1")).GetTypeName());
+   EXPECT_EQ("", desc.GetFieldDescriptor(desc.FindFieldId("f1")).GetTypeAlias());
    EXPECT_EQ("DataVector<std::int32_t,float>", desc.GetFieldDescriptor(desc.FindFieldId("f2")).GetTypeName());
+   EXPECT_EQ("", desc.GetFieldDescriptor(desc.FindFieldId("f2")).GetTypeAlias());
    EXPECT_EQ("DataVector<std::int32_t,double>", desc.GetFieldDescriptor(desc.FindFieldId("f3")).GetTypeName());
+   EXPECT_EQ("", desc.GetFieldDescriptor(desc.FindFieldId("f3")).GetTypeAlias());
    EXPECT_EQ("DataVector<bool,std::vector<std::uint32_t>>",
              desc.GetFieldDescriptor(desc.FindFieldId("f4")).GetTypeName());
+   EXPECT_EQ("", desc.GetFieldDescriptor(desc.FindFieldId("f4")).GetTypeAlias());
+   EXPECT_EQ("DataVector<double,double>", desc.GetFieldDescriptor(desc.FindFieldId("f5")).GetTypeName());
+   EXPECT_EQ("DataVector<Double32_t,double>", desc.GetFieldDescriptor(desc.FindFieldId("f5")).GetTypeAlias());
+   EXPECT_EQ("DataVector<std::int32_t,double>", desc.GetFieldDescriptor(desc.FindFieldId("f6")).GetTypeName());
+   EXPECT_EQ("", desc.GetFieldDescriptor(desc.FindFieldId("f6")).GetTypeAlias());
 
    auto v1 = reader->GetView<DataVector<int>>("f1");
    auto v3 = reader->GetView<DataVector<int>>("f3");
```

However, that doesn't seem to work right now:
```
553: [ RUN      ] RNTuple.TClassDefaultTemplateParameter
553: /home/jhahnfel/ROOT/src/tree/ntuple/v7/test/rfield_class.cxx:295: Failure
553: Expected equality of these values:
553:   ""
553:   desc.GetFieldDescriptor(desc.FindFieldId("f3")).GetTypeAlias()
553:     Which is: "DataVector<std::int32_t>"
553: /home/jhahnfel/ROOT/src/tree/ntuple/v7/test/rfield_class.cxx:299: Failure
553: Expected equality of these values:
553:   "DataVector<double,double>"
553:   desc.GetFieldDescriptor(desc.FindFieldId("f5")).GetTypeName()
553:     Which is: "DataVector<Double32_t,double>"
553: /home/jhahnfel/ROOT/src/tree/ntuple/v7/test/rfield_class.cxx:300: Failure
553: Expected equality of these values:
553:   "DataVector<Double32_t,double>"
553:   desc.GetFieldDescriptor(desc.FindFieldId("f5")).GetTypeAlias()
553:     Which is: "DataVector<Double32_t>"
553: [  FAILED  ] RNTuple.TClassDefaultTemplateParameter (10 ms)
```

I believe `Double32_t` as template argument was already wrong before, but the type alias for `f3` looks wrong to me...

tree/ntuple/v7/test/ntuple_types.cxx
```diff
@@ -37,7 +37,10 @@ TEST(RNTuple, TypeNameNormalization)
    EXPECT_EQ("std::int32_t", RFieldBase::Create("f", "signed").Unwrap()->GetTypeName());
    EXPECT_EQ("std::map<std::int32_t,std::int32_t>", RFieldBase::Create("f", "map<int, int>").Unwrap()->GetTypeName());
 
-   EXPECT_TRUE(RFieldBase::Create("f", "std::int32_t").Unwrap()->GetTypeAlias().empty());
```
hahnjo: You probably want to keep this newly added test that `std::int32_t` has an empty type alias

tree/ntuple/v7/src/RFieldBase.cxx
```diff
@@ -388,7 +388,7 @@ ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::
          if (arrayDef.size() != 2) {
             return R__FORWARD_RESULT(fnFail("the template list for std::array must have exactly two elements"));
          }
-         auto arrayLength = std::stoi(arrayDef[1]);
+         auto arrayLength = std::stoull(arrayDef[1]);
```
hahnjo, line 391: We probably want more error checking here, for example `std::stoull("123junk") = 123` without error reporting. `std::stoull` takes a second `pos` argument that stores the number of processed characters, or we can use the underlying `std::strtoull` where `str_end` should point to `NULL` after successful conversion.

tree/ntuple/v7/src/RFieldUtils.cxx
```diff
@@ -237,6 +238,31 @@ std::string ROOT::Experimental::Internal::GetNormalizedUnresolvedTypeName(const
...
+      return std::to_string(val) + "ll";
+   return std::to_string(val);
+}
+
+std::string ROOT::Experimental::Internal::GetNormalizedInteger(const std::string &intTemplateArg)
+{
+   R__ASSERT(!intTemplateArg.empty());
+   if (intTemplateArg[0] == '-')
+      return GetNormalizedInteger(std::stoll(intTemplateArg));
+   return GetNormalizedInteger(std::stoull(intTemplateArg));
```
hahnjo, line 263: same here

hahnjo (approved):
LGTM, some last minor comments from my side

tree/ntuple/v7/src/RFieldBase.cxx
jblomer: While looking again into the integer parsing, I realized that as of C++11, unsuffixed integers are `long long`, and the suffix `u` supports ints up to `unsigned long long`. So I think we should simplify the spec and the implementation and only add `u` if the literal is larger than 2^63.

PR #17638, opened 2025-02-05
[RF] Make it clear when to use RooSimultaneous::generateSimGlobal

guitargeek:
More details in the comments in the code added in this commit.

To make it short: the user probably wants to use `generateSimGlobal`
when not generating the index category value (or taking it from a proto
dataset), and `RooAbsPdf::generate()` otherwise.

To avoid confusion, there are now new errors when one does the opposite,
which is certainly a user error.

Revert also the commit where I deprecated `generateSimGlobal` for a lack of a real understanding of its purpose.

Thanks to @Tomas-Dado for pointing out that it's still needed, and for privately providing a code example where it's necessary.


lmoneta (approved):
LGTM!

PR #17639, opened 2025-02-05
[RF] Remove deprecated `ClassImp` macro calls in RooFit

guitargeek:
Since `ClassImp` is a no-op now and it will be deprecated soon.


dpiparo (approved):
Thanks for exploring this ahead of time for ROOT. LGTM!

PR #17640, opened 2025-02-06
[tut] Avoid races caused by Davix 0.8.8

dpiparo:
by using xrootd and reading from EOS Opendata


guitargeek (approved):
Very good, thanks!

PR #17643, opened 2025-02-06
Always remove Sumw2 in object painting

linev:
In classes like `THStack`, `TGraph2D`, `TF1`, `TF2` created histogram may have Sumw2 array - 
if  `TH1::SetDefaultSumw2();` was activated before. Always remove such array to save memory and 
to avoid paint artifacts as in #17470


linev:
All these histograms created for painting of objects where Sumw2 is not used.

PR #17645, opened 2025-02-06
[skip-ci] True and False are not defined

couet:
avoid the error:
```
/github/home/ROOT-CI/src/tutorials/roofit/roostats/rs401d_FeldmanCousins.C:139:49: error: use of undeclared identifier 'False'
   RooMsgService::instance().setStreamStatus(0, False);
                                                ^
```


guitargeek (approved):
Thank you very much for the fix!

PR #17650, opened 2025-02-06
[TTree] merge also branches when merging TTrees

ferdymercury:
Fixes https://github.com/root-project/root/issues/14558
Fixes https://github.com/root-project/root/issues/12510

## Checklist:

- [x] tested changes locally
- [x] updated the docs (if necessary)

PR #17652, opened 2025-02-06
 [TTree] when merging, ignore trees without branches

### LINKED ISSUE #12510: Issue with `hadd` when first file has empty tree
### Describe the bug
It seems `hadd` is confused by the zero length TTree as a first file. Pls find dummy root files in the attached ZIP to reproduce the issue. In `dummyFile0.root`, there're 1000 events in the `dummyTree`; While in `dummyFile1.root`, the tree is empty. So when `hadd` 2 files, like
```
hadd dummyFile01.root dummyFile0.root dummyFile1.root
``` 
Everything is okay. While if you do like
```
hadd dummyFile10.root dummyFile1.root dummyFile0.root
```
And you open the merged `dummyFile10.root` to get entries like
```
root [1] dummyTree->GetEntries()
(long long) 1000
```
It looks okay. But if you show the first event, like
```
root [1] dummyTree->Show(0)
======> EVENT:0
 index           = 0
 dummyVariable   = 2.13654e-306
```
The `dummyVariable` is definitely non-sense. It should be `0.261287` as it is in the `dummyFile0.root`. I guess if put the empty tree at the first in merging, ROOT is trying get a variable from an empty pointer? 

### Expected behavior
```
root [1] dummyTree->Show(0)
======> EVENT:0
 index           = 0
 dummyVariable   = 0.261287
```

### Setup
1. ROOT version: 6.26/06
2. Operating system: macosx64

[dummyFiles.zip](https://github.com/root-project/root/files/11020163/dummyFiles.zip)

### LINKED ISSUE #14558: [ROOT-4716] TTree merging problems when including empty trees
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

From https://its.cern.ch/jira/browse/ROOT-4716

Dear Developers,

This problem that I'll describe now was first discussed in

https://savannah.cern.ch/bugs/?98501

But since by now the issue doesn't seem to have anything to do with PROOF and TProofOutputFile, I thought I'd rather continue the discussion in a new thread.

What I originally observed was that in the latest 5.34.XX version some of my PROOF jobs would return me an empty TTree, even though the job reported that it selected some events. For a while I was not able to find a handle on what triggered this problem, as it seemed quite random. (The same job would not always return empty trees.)

I think I now understand what's going on. It has to do with the issue in TFileMerger that was discovered in ATLAS production jobs a long time ago. When TFileMerger is given multiple files, the first of which doesn't have any branches in it, the merged output file will not have any branches either. Even if the rest of the files given to TFileMerger did have some structure.

In my PROOF jobs this happens because my code only sets up the format of its output TTree when it opens its first file. The output TTree format depends a bit on what branches are available on the input, so I can't do this any sooner.

But PROOF in some cases will not process any events in some of its workers. When this happens, it's actually these workers that return their output files to the PROOF master first. Files which hold completely empty TTree-s.

This made my jobs random, as PROOF would not always do this, sometimes (depending on the state of our cluster) it would actually process some events on all of the workers in this job.

The problem can be demonstrated very easily. Attached is a macro that does this. (The macro was first written to demonstrate a problem with merging a file into a file on XRootD, but this issue doesn't depend on where the file is put.)

While I may be able to circumvent the problem in my analysis code for now (by initializing the TTree format earlier in the code), this merging problem would be very good to solve. As I wrote at the beginning, the ATLAS production system also has to have extra code in place which knows not to give files with empty TTrees to hadd as the first file.

Cheers,
Attila

### Reproducer

NA

### ROOT version

Any

### Installation method

Any

### Operating system

Any

### Additional context

_No response_

ferdymercury:
# This Pull request:

Alternative to https://github.com/root-project/root/pull/17650 as suggested by pcanal

## Changes or fixes:

Fixes #14558
Fixes #12510

### DIFF
--- a/io/io/test/TFileMergerTests.cxx
+++ b/io/io/test/TFileMergerTests.cxx
@@ -1,6 +1,7 @@
 #include "ROOT/TestSupport.hxx"
 
 #include "TFileMerger.h"
+#include "TFileMergeInfo.h"
 
 #include "TMemFile.h"
 #include "TTree.h"
@@ -84,12 +85,12 @@ TEST(TFileMerger, MergeSingleOnlyListed)
    hist3->Fill(1);   hist3->Fill(1);   hist3->Fill(1);
    hist4->Fill(1);   hist4->Fill(1);   hist4->Fill(1);   hist4->Fill(1);
    a.Write();
-   
+
    TFileMerger merger;
    auto output = std::unique_ptr<TFile>(new TFile("SingleOnlyListed.root", "RECREATE"));
    bool success = merger.OutputFile(std::move(output));
    ASSERT_TRUE(success);
-   
+
    merger.AddObjectNames("hist1");
    merger.AddObjectNames("hist2");
    merger.AddFile(&a, false);
@@ -100,4 +101,61 @@ TEST(TFileMerger, MergeSingleOnlyListed)
    output = std::unique_ptr<TFile>(TFile::Open("SingleOnlyListed.root"));
    ASSERT_TRUE(output.get() && output->GetListOfKeys());
    EXPECT_EQ(output->GetListOfKeys()->GetSize(), 2);
+}
+
+// https://github.com/root-project/root/issues/14558 aka https://its.cern.ch/jira/browse/ROOT-4716
+TEST(TFileMerger, MergeBranches)
+{
+   TTree atree("atree", "atitle");
+   int value;
+   atree.Branch("a", &value);
+
+   TTree abtree("abtree", "abtitle");
+   abtree.Branch("a", &value);
+   abtree.Branch("b", &value);
+   value = 11;
+   abtree.Fill();
+   value = 42;
+   abtree.Fill();
+
+   TTree dummy("emptytree", "emptytitle");
+   TList treelist;
+
+   // Case 1 - Static: NoBranch + NoEntries + 2 entries
+   treelist.Add(&dummy);
+   treelist.Add(&atree);
+   treelist.Add(&abtree);
+   std::unique_ptr<TFile> file1(TFile::Open("b_4716.root", "RECREATE"));
+   auto rtree = TTree::MergeTrees(&treelist);
+   ASSERT_TRUE(rtree->FindBranch("a") != nullptr);
+   EXPECT_EQ(rtree->FindBranch("a")->GetEntries(), 2);
+   ASSERT_TRUE(rtree->FindBranch("b") == nullptr);
+   file1->Write();
+
+   // Case 2 - This (NoBranch) + NoEntries + 2 entries
+   treelist.Clear();
+   treelist.Add(&atree);
+   treelist.Add(&abtree);
+   std::unique_ptr<TFile> file2(TFile::Open("c_4716.root", "RECREATE"));
+   TFileMergeInfo info2(file2.get());
+   dummy.Merge(&treelist, &info2);
+   ASSERT_TRUE(dummy.FindBranch("a") != nullptr);
+   ASSERT_TRUE(dummy.FindBranch("b") == nullptr);
+   EXPECT_EQ(dummy.FindBranch("a")->GetEntries(), 2);
+   EXPECT_EQ(atree.FindBranch("a")->GetEntries(), 2);
+   // atree has now 2 entries instead of zero since it was used as skeleton for dummy
+   file2->Write();
+
+   // Case 3 - This (NoEntries) + 2 entries
+   TTree a0tree("a0tree", "a0title"); // We cannot reuse atree since it was cannibalized by dummy
+   a0tree.Branch("a", &value);
+   treelist.Clear();
+   treelist.Add(&abtree);
+   std::unique_ptr<TFile> file3(TFile::Open("d_4716.root", "RECREATE"));
+   TFileMergeInfo info3(file3.get());
+   a0tree.Merge(&treelist, &info3);
+   ASSERT_TRUE(a0tree.FindBranch("a") != nullptr);
+   ASSERT_TRUE(a0tree.FindBranch("b") == nullptr);
+   EXPECT_EQ(a0tree.FindBranch("a")->GetEntries(), 2);
+   file3->Write();
 }
--- a/tree/tree/src/TTree.cxx
+++ b/tree/tree/src/TTree.cxx
@@ -30,7 +30,7 @@
 /** \class TTree
 \ingroup tree
 
-A TTree represents a columnar dataset. Any C++ type can be stored in its columns. The modern 
+A TTree represents a columnar dataset. Any C++ type can be stored in its columns. The modern
 version of TTree is RNTuple: please consider using it before opting for TTree.
 
 A TTree, often called in jargon *tree*, consists of a list of independent columns or *branches*,
@@ -141,19 +141,19 @@ It is strongly recommended to persistify those as objects rather than lists of l
 ~~~ {.cpp}
     auto branch = tree.Branch( branchname, STLcollection, buffsize, splitlevel);
 ~~~
-`STLcollection` is the address of a pointer to a container of the standard 
-library such as `std::vector`, `std::list`, containing pointers, fundamental types 
+`STLcollection` is the address of a pointer to a container of the standard
+library such as `std::vector`, `std::list`, containing pointers, fundamental types
 or objects.
 If the splitlevel is a value bigger than 100 (`TTree::kSplitCollectionOfPointers`)
-then the collection will be written in split mode, i.e. transparently storing 
+then the collection will be written in split mode, i.e. transparently storing
 individual data members as arrays, therewith potentially increasing compression ratio.
 
-### Note 
-In case of dynamic structures changing with each entry, see e.g. 
+### Note
+In case of dynamic structures changing with each entry, see e.g.
 ~~~ {.cpp}
     branch->SetAddress(void *address)
 ~~~
-one must redefine the branch address before filling the branch 
+one must redefine the branch address before filling the branch
 again. This is done via the `TBranch::SetAddress` member function.
 
 \anchor addingacolumnofobjs
@@ -186,8 +186,8 @@ Another available syntax is the following:
 - `p_object` is a pointer to an object.
 - If `className` is not specified, the `Branch` method uses the type of `p_object`
   to determine the type of the object.
-- If `className` is used to specify explicitly the object type, the `className` 
-  must be of a type related to the one pointed to by the pointer. It should be 
+- If `className` is used to specify explicitly the object type, the `className`
+  must be of a type related to the one pointed to by the pointer. It should be
   either a parent or derived class.
 
 Note: The pointer whose address is passed to `TTree::Branch` must not
@@ -205,13 +205,13 @@ or not
     tree.Branch(branchname, &p_object);
 ~~~
 In either case, the ownership of the object is not taken over by the `TTree`.
-Even though in the first case an object is be allocated by `TTree::Branch`, 
+Even though in the first case an object is be allocated by `TTree::Branch`,
 the object will <b>not</b> be deleted when the `TTree` is deleted.
 
 \anchor addingacolumnoftclonesarray
 ## Add a column holding TClonesArray instances
 
-*The usage of `TClonesArray` should be abandoned in favour of `std::vector`, 
+*The usage of `TClonesArray` should be abandoned in favour of `std::vector`,
 for which `TTree` has been heavily optimised, as well as `RNTuple`.*
 
 ~~~ {.cpp}
@@ -347,7 +347,7 @@ int main()
 ~~~
 ## TTree Diagram
 
-The following diagram shows the organisation of the federation of classes related to TTree. 
+The following diagram shows the organisation of the federation of classes related to TTree.
 
 Begin_Macro
 ../../../tutorials/legacy/tree/tree.C
@@ -6841,6 +6841,8 @@ bool TTree::MemoryFull(Int_t nbytes)
 ///
 /// Trees in the list can be memory or disk-resident trees.
 /// The new tree is created in the current directory (memory if gROOT).
+/// Trees with no branches will be skipped, the branch structure
+/// will be taken from the first non-zero-branch Tree of {li}
 
 TTree* TTree::MergeTrees(TList* li, Option_t* options)
 {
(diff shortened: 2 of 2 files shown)


io/io/test/TFileMergerTests.cxx
```diff
@@ -101,3 +102,36 @@ TEST(TFileMerger, MergeSingleOnlyListed)
    ASSERT_TRUE(output.get() && output->GetListOfKeys());
    EXPECT_EQ(output->GetListOfKeys()->GetSize(), 2);
 }
+
+// https://github.com/root-project/root/issues/14558 aka https://its.cern.ch/jira/browse/ROOT-4716
+TEST(TFileMerger, MergeBranches)
+{
+   TTree atree("atree", "atitle");
```
pcanal, line 109: Does this test fail before the fix?

tree/tree/src/TTree.cxx
```diff
@@ -6929,7 +6928,8 @@ Long64_t TTree::Merge(TCollection* li, TFileMergeInfo *info)
       } else {
          TDirectory::TContext ctxt(info->fOutputDirectory);
          TIOFeatures saved_features = fIOFeatures;
-         TTree *newtree = CloneTree(-1, options);
+         Long64_t nentries = tree->GetEntries();
```
ferdymercury, line 6931:
```suggestion
         Long64_t nentries = GetEntries();
```

io/io/test/TFileMergerTests.cxx
ferdymercury: Yes, I get:

```
dummy.FindBranch("a")
(TBranch *) nullptr
```

pcanal: I don't see why.  Both TTree have a `a` branch and an entry ....

ferdymercury: Dummy has zero branches. We are merging dummy + two extra trees

io/io/test/TFileMergerTests.cxx
```diff
@@ -101,3 +102,36 @@ TEST(TFileMerger, MergeSingleOnlyListed)
...
+   atree.Branch("a", &value);
+   value = 42;
+   atree.Fill();
+
+   TTree abtree("abtree", "abtitle");
+   abtree.Branch("a", &value);
+   value = 42;
+   abtree.Fill();
+  
+   TTree dummy;
```
pcanal, line 120: This does not create a fully functional tree and thus is not representative of user code.  A fully functional tree needs a name and title.

io/io/test/TFileMergerTests.cxx
pcanal: Note: the issue (in the original problem description) is *not* zero branches but zero entries.

I am also not sure how this patch change the behavior in the test (with the dummy/non-functional starter tree) change (i.e. I am not sure which change in TTree.cxx triggeres the modification nor why).

ferdymercury: > Note: I think that the issue (in the original problem description) is _not_ zero branches but zero entries.

I don't think so, but maybe @krasznaa can confirm what he meant by:

`When TFileMerger is given multiple files, the first of which doesn't have any branches in it, the merged output file will not have any branches either.`

Does this mean zero branches or zero entries?

If I create an empty tree with branches but zero entries, then I do not see any problem in master.

```
TTree atree("atree", "atitle");
int value;
atree.Branch("a", &value);
value = 11;
atree.Fill();

TTree abtree("abtree", "abtitle");
abtree.Branch("a", &value);
value = 42;
abtree.Fill();
TTree dummy("emptytree", "emptytitle");
bool zeroBranches = false;
if(!zeroBranches) dummy.Branch("a", &value);
TList treelist;
treelist.Add(&atree);
treelist.Add(&abtree);
std::unique_ptr<TFile> file(TFile::Open("c4716.root", "RECREATE"));
TFileMergeInfo info(file.get());
dummy.Merge(&treelist, &info);
dummy.Scan()
************************
*    Row   *       a.a *
************************
*        0 *        11 *
*        1 *        42 *
************************
```


That's also what @hahnjo identified here: https://github.com/root-project/root/issues/14558#issuecomment-1941277951

On master, the problem only appears if I put zeroBranches=true; entries are somehow copied but not the branches.

I get:

```
************
*    Row   *
************
*        0 *
*        1 *
************
```

io/io/test/TFileMergerTests.cxx
ferdymercury, line 120:
```suggestion
   TTree dummy("emptytree","emptytitle");
```

io/io/test/TFileMergerTests.cxx
ferdymercury: Which is why I had gone into the importBranch mechanism of https://github.com/root-project/root/pull/17650

pcanal: There is a lot of confusion (including in some of my memories).  

> I don't think so, but maybe @krasznaa can confirm what he meant by:

Indeed you are right, it is zero branches and the change in this PR are not quite right.
The change in https://github.com/root-project/root/pull/17650 are however 'too much' to implement the intent.

Arguably the 'right' solution is for the merging to skip/ignore trees that are completely empty.
I.e. For merging:
```
TTree t1("tree","1st tree");
TTree t2("tree","2nd tree");
t2.Branch("a", &a_value);
TTree t3("tree","3rd tree");
t3.Branch("a", &a_value);
t3.Branch("b", &b_value);
```
You should end up with a tree that has a single branch `a` and its number of entries the sum of the entries in t2 and t3.

For the static function `TTree::MergeTrees`, this is 'easy', just replace:
```
      if (nentries == 0) continue;
```
with
```
     if (tree->GetListOfBranches()->IsEmpty())
        continue; // Completely ignore the empty trees.
     if (newtree && nentries == 0)
        continue; // If we already have the structure and we have no entry, save time and skip
```
The `-1` in `CloneTree` do no need to change they are implied (the `-1` gets replaced by the actual number of entries, so what I had proposed for those calls what essential a no-op)

For `TTree::Merge(TCollection* li, TFileMergeInfo *info)`, this is a bit more annoying to implement if we want (as we should) to avoid too many recursion.  The obvious but slightly (i.e. missing an essential part) wrong change is:
```
Long64_t  TTree::Merge(TCollection* li, TFileMergeInfo *info) {
    if (fBranches.IsEmpty()) {
       if (!li || li->IsEmpty())
          return 0; // Nothing to do ....
      // Let's find the first non-empty
      TIter next(li);
      TTree *tree;
      while ((tree = (TTree*)next())) {
         if (tree==this || tree->GetListOfBranches()->IsEmpty())
            continue;
         return tree->Merge(li, info);
      }
   }
```
The problem is that `TFileMerge` is expecting the result to be in the original `this` and then will delete the other element of the list (including the one we used for the merging).

Replacing in the code above:
```
         return tree->Merge(li, info);
```
by
```
         // We could come from a list made up of different names, the first one still wins
         tree->SetName(this->GetName());
         auto prevEntries = tree->GetEntries();
         auto result = tree->Merge(li, info);
         if (result != prevEntries) {
            // If there is no additional entries, the first write was enough.
            tree->Write();
         }
         // Make sure things are really written out to disk before attempting any reading.
         info->fOutputDirectory->GetFile()->Flush();
         // Read back the complete info in this TTree, so that TFileMerge does not 
         // inadvertently write the empty tree.
         info->fOutputDirectory->ReadTObject(this, this->GetName());
         return result;
```


<details>
<summary>Other details on the test</summary>


The case:
```
...
TTree dummy;
...
std::unique_ptr<TFile> file(TFile::Open("c4716.root", "RECREATE"));
TFileMergeInfo info(file.get());
```
leads to broken file:
```
root.exe -b -l -q c4716.root -e '_file0->Print();'
TFile**         c4716.root
 TFile*         c4716.root
  KEY: TTree    ;1
```
Note the empty name record on the file.  If we tweaks the 'name' of the `dummy` variable, one will see that the name given on file is determine by the one given to `dummy` (So `emptytree` in your last example).

The usual case for the merging of trees is to merge trees that have the same name and thus the first name is the one being selected.

The code that ends with:
```
std::unique_ptr<TFile> file(TFile::Open("c4716.root", "RECREATE"));
TFileMergeInfo info(file.get());
dummy.Merge(&treelist, &info);
dummy.Scan()
```
is missing a call to:
```
file->Write();
```
without it the produced file does not contains the final `TTree` object.  For example without it you get:
```
root.exe -b -l -q missingEntries.C ; root.exe -b -l -q c4716.root -e 'dummytree->Print();'

Processing missingEntries.C...
************************
*    Row   *       a.a *
************************
*        0 *        22 *
*        1 *        11 *
*        2 *        42 *
************************

Attaching file c4716.root as _file0...
(TFile *) 0x2ad23c0
******************************************************************************
*Tree    :dummytree : emptytitle                                             *
*Entries :        1 : Total =             905 bytes  File  Size =        441 *
*        :          : Tree compression factor =   1.00                       *
******************************************************************************
*Br    0 :a         : a/I                                                    *
*Entries :        1 : Total  Size=        546 bytes  File Size  =         77 *
*Baskets :        1 : Basket Size=      32000 bytes  Compression=   1.00     *
*............................................................................*
```

</details>

pcanal:
Part of my confusion is that https://github.com/root-project/root/issues/12510 is indeed about a file with branches and no entries.
In this case, the problem is not that there is missing branches or missing entries but rather that the first entry (number 0) in the output file is corrupted (has the wrong values).

With the files https://github.com/root-project/root/files/11333378/dummyFiles.zip, we get:

```
$ root.exe -b -l -q dummyFiles/dummyFile0.root -e 'dummyTree->Scan("*","","",2,0);'

Attaching file dummyFiles/dummyFile0.root as _file0...
(TFile *) 0x22986b0
************************************
*    Row   * index.ind * dummyVari *
************************************
*        0 *         0 * 0.2612867 *
*        1 *         1 * 0.7484587 *
************************************
```
```
$ hadd -f dummyFile10.root dummyFiles/dummyFile1.root dummyFiles/dummyFile0.root 
Info in <hadd>: target file: dummyFile10.root
Info in <hadd>: compression setting for all output: 101
hadd Source file 1: dummyFiles/dummyFile1.root
hadd Source file 2: dummyFiles/dummyFile0.root
hadd Target path: dummyFile10.root:/
```
```
$ root.exe -b -l -q dummyFile10.root -e 'dummyTree->Scan("*","","",2,0);'

Attaching file dummyFile10.root as _file0...
(TFile *) 0x2f39310
************************************
*    Row   * index.ind * dummyVari *
************************************
*        0 *         0 * 2.13e-306 *
*        1 *         1 * 0.7484587 *
************************************
```
I.e. Row 0 is wrong.

Nonetheless, it might (or might not) be fixed by this PR, in particular in my last proposal the part:
```
     if (newtree && nentries == 0)
        continue; // If we already have the structure and we have no entry, save time and skip
```
i.e. cloning the structure even if there is no entries, might solve the issue.

tree/tree/src/TTree.cxx
```diff
@@ -6929,7 +6932,8 @@ Long64_t TTree::Merge(TCollection* li, TFileMergeInfo *info)
       } else {
          TDirectory::TContext ctxt(info->fOutputDirectory);
          TIOFeatures saved_features = fIOFeatures;
-         TTree *newtree = CloneTree(-1, options);
+         Long64_t nentries = GetEntries();
+         TTree *newtree = CloneTree(nentries == 0 ? 0 : -1, options);
```
ferdymercury, line 6936:
```suggestion
       TTree *newtree = CloneTree(-1, options);
```

io/io/test/TFileMergerTests.cxx
```diff
@@ -101,3 +102,56 @@ TEST(TFileMerger, MergeSingleOnlyListed)
...
+   treelist.Add(&atree);
+   treelist.Add(&abtree);
+   std::unique_ptr<TFile> file1(TFile::Open("b_4716.root", "RECREATE"));
+   auto rtree = TTree::MergeTrees(&treelist);
+   ASSERT_TRUE(rtree->FindBranch("a") != nullptr);
+   EXPECT_EQ(rtree->FindBranch("a")->GetEntries(), 2);
+   ASSERT_TRUE(rtree->FindBranch("b") == nullptr);
+   file1->Write();
+   
+   // Case 2 - This - NoBranch + NoEntries + 2 entries
```
ferdymercury, line 135:
```suggestion
   // Case 2 - This (NoBranch) + NoEntries + 2 entries
```

tree/tree/src/TTree.cxx
```diff
@@ -6877,9 +6882,38 @@ TTree* TTree::MergeTrees(TList* li, Option_t* options)
...
+      // Let's find the first non-empty
+      TIter next(li);
+      TTree *tree;
+      while ((tree = (TTree*)next())) {
+         if (tree==this || tree->GetListOfBranches()->IsEmpty())
+            continue;
+         // We could come from a list made up of different names, the first one still wins
+         tree->SetName(this->GetName());
+         auto prevEntries = tree->GetEntries();
+         auto result = tree->Merge(li, info);
```
ferdymercury, line 6902:
```suggestion
         auto result = tree->Merge(li, options);
```

tree/tree/src/TTree.cxx
```diff
@@ -6877,9 +6882,38 @@ TTree* TTree::MergeTrees(TList* li, Option_t* options)
...
+         // Make sure things are really written out to disk before attempting any reading.
+         if (tree->GetCurrentFile()) {
+            tree->GetCurrentFile()->Flush();
+            // Read back the complete info in this TTree, so that caller does not 
+            // inadvertently write the empty tree.
+            tree->GetDirectory()->ReadTObject(this, this->GetName());
+         }
+         return result;
+      }
+   }
```
ferdymercury, line 6916:
```suggestion
      }
      return 0; // All trees have empty branches
   }
```

tree/tree/src/TTree.cxx
```diff
@@ -6912,11 +6946,38 @@ Long64_t TTree::Merge(TCollection* li, Option_t *options)
...
+            tree->Write();
+         }
+         // Make sure things are really written out to disk before attempting any reading.
+         info->fOutputDirectory->GetFile()->Flush();
+         // Read back the complete info in this TTree, so that TFileMerge does not 
+         // inadvertently write the empty tree.
+         info->fOutputDirectory->ReadTObject(this, this->GetName());
+         return result;
+      }
+   }
```
ferdymercury, line 6980:
```suggestion
      }
      return 0; // All trees have empty branches
   }
```

io/io/test/TFileMergerTests.cxx
ferdymercury: Should we print a "Warning" message saying that "this tree has no branches" ? (or maybe if gDebug)

pcanal: Yes, we `gDebug > 2`

ferdymercury: Done!

tree/tree/src/TTree.cxx
```diff
@@ -6852,8 +6854,15 @@ TTree* TTree::MergeTrees(TList* li, Option_t* options)
    while ((obj=next())) {
       if (!obj->InheritsFrom(TTree::Class())) continue;
       TTree *tree = (TTree*)obj;
+      if (tree->GetListOfBranches()->IsEmpty()) {
+         if (gDebug > 2) {
+            Warning("MergeTrees","TTree %s has no branches, skipping.", tree->GetName());
```
ferdymercury, line 6859:
```suggestion
            tree->Warning("MergeTrees","TTree %s has no branches, skipping.", tree->GetName());
```

pcanal (approved):
Thanks.

PR #17654, opened 2025-02-06
[RF][HF] Re-use any type of RooAbsArgs in ParamHistFunc::createParamSet

### LINKED ISSUE #12225: Allowing to use AddPreprocessFunction for shape factors in HistFactory
We would like to have an option to reparametrise shape-factors, i.e. the per-bin normalisation factors in HistFactory. This would be useful for template fits where different histograms represent different values of some parameter, e.g. mass. Currently, it seems that only NormFactors can be reparametrised and the code crashes with a seg fault when trying to reparametrise the shape factors. See the example attached.

[shapeFactorCrash.zip](https://github.com/root-project/root/files/10581134/shapeFactorCrash.zip)

### Describe the solution you'd like
We would like to be able to use  HistFactory::Measurement::AddPreprocessFunction for shapefactors

### Describe alternatives you've considered
There is a workaround by doing this reparametrisation manually, but it cumbersome and errorprone

### Additional context
Note that even if this is not supperted, the code should probably not crash with a seg fault (tested with ROOT Version: 6.27/01)

guitargeek:
When connecting preprocessing functions to nuisance parameters via `Measurement::AddPreprocessFunction`, the nuisance parameter is replaced by a RooAbsReal that is not a RooRealVar, and the implementation of `ParamHistFunc::createParamSet` should be able to deal with that.

So far, `createParamSet` assumed that any pre-existing argument in the workspace with the nuisance parameter name is of type RooRealVar, which is not valid in that case. Therefore, using preprocessing functions in place of these parameters resulted in a segfault.

To fix this, the code is generalized to any RooAbsArg type.

Closes #12225.

### DIFF
--- a/roofit/histfactory/src/ParamHistFunc.cxx
+++ b/roofit/histfactory/src/ParamHistFunc.cxx
@@ -317,10 +317,8 @@ RooArgList ParamHistFunc::createParamSet(RooWorkspace& w, const std::string& Pre
       gamma.setConstant( false );
 
       w.import( gamma, RooFit::RecycleConflictNodes() );
-      RooRealVar* gamma_wspace = (RooRealVar*) w.var( VarName );
 
-      paramSet.add( *gamma_wspace );
-
+      paramSet.add(*w.arg(VarName));
     }
   }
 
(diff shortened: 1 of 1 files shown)


lmoneta (approved):
LGTM

PR #17661, opened 2025-02-07
[ntuple] don't try to construct an Emulated TClass as a RClassField

### LINKED ISSUE #17648: [ntuple] Record field emulation not working with TFile
### Description

When opening a `ROOT::RNTuple` with `TFile`, record field emulation does not work.

### Reproducer

Observed in the wild, but can be reproduced with the related test:

```diff
diff --git a/tree/ntuple/v7/test/ntuple_emulated.cxx b/tree/ntuple/v7/test/ntuple_emulated.cxx
index 69627705ba..c29bf0f89f 100644
--- a/tree/ntuple/v7/test/ntuple_emulated.cxx
+++ b/tree/ntuple/v7/test/ntuple_emulated.cxx
@@ -91,7 +91,12 @@ struct Outer_Simple {
    // Now test loading entries with a reader
    RNTupleDescriptor::RCreateModelOptions cmOpts;
    cmOpts.fEmulateUnknownTypes = true;
-   reader = RNTupleReader::Open(cmOpts, "ntpl", fileGuard.GetPath());
+   ROOT::TestSupport::CheckDiagsRAII diagRAII;
+   diagRAII.optionalDiag(kWarning, "TClass::Init", "no dictionary for class",
+                         /*matchFullMessage=*/false);
+   std::unique_ptr<TFile> file(TFile::Open(fileGuard.GetPath().c_str()));
+   std::unique_ptr<ROOT::RNTuple> ntpl(file->Get<ROOT::RNTuple>("ntpl"));
+   reader = RNTupleReader::Open(cmOpts, *ntpl);
    reader->LoadEntry(0);
 }

```

### ROOT version

`master`

silverweed:
When constructing a RClassField we now require that the TClass is at least Interpreted. If not, we follow the same logic as when we have no TClass at all (i.e. either create an emulated field or fail, depending on the user options).

## Checklist:

- [x] tested changes locally

This PR fixes #17648

### DIFF
--- a/tree/ntuple/v7/src/RFieldBase.cxx
+++ b/tree/ntuple/v7/src/RFieldBase.cxx
@@ -552,7 +552,11 @@ ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::
 
       if (!result) {
          auto cl = TClass::GetClass(typeName.c_str());
-         if (cl != nullptr) {
+         // NOTE: if the class is not at least "Interpreted" we currently don't try to construct
+         // the RClassField, as in that case we'd need to fetch the information from the StreamerInfo
+         // rather than from TClass. This might be desirable in the future, but for now in this
+         // situation we rely on field emulation instead.
+         if (cl != nullptr && cl->GetState() >= TClass::kInterpreted) {
             createContextGuard.AddClassToStack(resolvedType);
             if (cl->GetCollectionProxy()) {
                result = std::make_unique<RProxiedCollectionField>(fieldName, typeName);
--- a/tree/ntuple/v7/src/RFieldMeta.cxx
+++ b/tree/ntuple/v7/src/RFieldMeta.cxx
@@ -95,6 +95,10 @@ ROOT::Experimental::RClassField::RClassField(std::string_view fieldName, TClass 
                                     ROOT::ENTupleStructure::kRecord, false /* isSimple */),
      fClass(classp)
 {
+   if (fClass->GetState() < TClass::kInterpreted) {
+      throw RException(R__FAIL(std::string("RField: RClassField \"") + classp->GetName() +
+                               " cannot be constructed from a class that's not at least Interpreted"));
+   }
    // Avoid accidentally supporting std types through TClass.
    if (fClass->Property() & kIsDefinedInStd) {
       throw RException(R__FAIL(std::string(GetTypeName()) + " is not supported"));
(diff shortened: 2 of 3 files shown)


jblomer (approved):
Nice fix!

hahnjo (approved):
Thanks!

tree/ntuple/v7/src/RFieldBase.cxx
```diff
@@ -594,7 +594,10 @@ ROOT::Experimental::RFieldBase::Create(const std::string &fieldName, const std::
 
       if (!result) {
          auto cl = TClass::GetClass(canonicalType.c_str());
-         if (cl != nullptr) {
+         // NOTE: if the class is not at least "Interpreted" we don't have enough information to
+         // properly construct the RClassField (e.g. we are missing the list of bases), so in that
```
pcanal, line 598: This is not precise.  The list of base class is available in the `TStreamerInfo` and thus available for an Emulated class.

Is this PR a temporary fix/workaround or am I missing something?

silverweed: How do you access that information? When we load a class from file without dictionary we observe that `TClass::GetListOfBases()` returns `nullptr`. Do we need to take some explicit steps to have that list filled? Or do we need to call a different method?

pcanal: You have to look at the TStreamerInfo, eg.
```
auto info = cl->GetStreamerInfo(); // This return the current StreamerInfo describing the in memory layout
for(auto elem : TRangeStaticCast<TStreamerElement>( *info ) {
   if (elem->IsBase()) {
       ...
   }
}
```

silverweed: @pcanal  I guess you meant `TRangeStaticCast<TStreamerElement>(info->GetElements())`; 
but this gives me `TStreamerElement`s and `TStreamerBase`s: how do I get to the `TBaseClass` from there?
Or can I still access stuff like `GetDelta()` from a `TStreamerBase`?

pcanal: >  I guess you meant TRangeStaticCast<TStreamerElement>(info->GetElements());

Indeed, sorry :)

> how do I get to the TBaseClass from there?

You can't, it is a different interface.

> Or can I still access stuff like GetDelta() from a TStreamerBase?

Yes, through `GetOffset`.  Note that the list of `TStreamerElement` will only contains the direct base classes.

silverweed: @pcanal for now we decided to go through field emulation when the TClass is emulated, we might change this in the future if we see the need for it

pcanal: You might still need to update the PR summary.

I still find it confusing to use the same adjective for to similar but distinct concept.  I.e. 'emulated [T]Class' vs 'emulated Field'.

silverweed: We can rediscuss the terminology, but I don't believe this PR is the best place to do so.
I personally don't have a problem with it, considering it's an analogous concept done in a different context. As long as both concepts are well documented, I don't think there is much room for confusion.

pcanal: Fair enough.

PR #17662, opened 2025-02-07
[skip-ci] replace False by false

couet:
Correct an error where `kFALSE` was changed to `false`. This instance was missed and remains as `False`, but it should be `false`. This issue doesn't appear in the Doxygen log because the macro isn't executed because no image is generated for it since there's no `\macro_image` directive; only the code is displayed. The error was discovered using a `grep` search.


guitargeek (approved):
Thank you!

PR #17665, opened 2025-02-07
[ntuple] Move public utility types out of experimental

jblomer:
Companion PR: https://github.com/root-project/roottest/pull/1258

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RColumn.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RColumn.hxx
@@ -61,7 +61,7 @@ private:
    /// The currently mapped page for reading
    RPageRef fReadPageRef;
    /// The column id in the column descriptor, once connected to a sink or source
-   DescriptorId_t fOnDiskId = kInvalidDescriptorId;
+   ROOT::DescriptorId_t fOnDiskId = ROOT::kInvalidDescriptorId;
    /// Global index of the first element in this column; usually == 0, unless it is a deferred column
    ROOT::NTupleSize_t fFirstElementIndex = 0;
    /// Used to pack and unpack pages on writing/reading
@@ -120,9 +120,9 @@ public:
    /// Connect the column to a page sink.  `firstElementIndex` can be used to specify the first column element index
    /// with backing storage for this column.  On read back, elements before `firstElementIndex` will cause the zero page
    /// to be mapped.
-   void ConnectPageSink(DescriptorId_t fieldId, RPageSink &pageSink, ROOT::NTupleSize_t firstElementIndex = 0U);
+   void ConnectPageSink(ROOT::DescriptorId_t fieldId, RPageSink &pageSink, ROOT::NTupleSize_t firstElementIndex = 0U);
    /// Connect the column to a page source.
-   void ConnectPageSource(DescriptorId_t fieldId, RPageSource &pageSource);
+   void ConnectPageSource(ROOT::DescriptorId_t fieldId, RPageSource &pageSource);
 
    void Append(const void *from)
    {
@@ -351,7 +351,7 @@ public:
    }
    std::uint32_t GetIndex() const { return fIndex; }
    std::uint16_t GetRepresentationIndex() const { return fRepresentationIndex; }
-   DescriptorId_t GetOnDiskId() const { return fOnDiskId; }
+   ROOT::DescriptorId_t GetOnDiskId() const { return fOnDiskId; }
    ROOT::NTupleSize_t GetFirstElementIndex() const { return fFirstElementIndex; }
    RPageSource *GetPageSource() const { return fPageSource; }
    RPageSink *GetPageSink() const { return fPageSink; }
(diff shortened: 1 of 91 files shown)


tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -344,29 +344,29 @@ std::ostream &operator<<(std::ostream &os, const std::optional<RColumnDescriptor
...
-   if (a == ENTupleColumnType::kSplitUInt64 && b == ENTupleColumnType::kUInt64) return true;
-   if (a == ENTupleColumnType::kIndex32 && b == ENTupleColumnType::kSplitIndex32) return true;
-   if (a == ENTupleColumnType::kSplitIndex32 && b == ENTupleColumnType::kIndex32) return true;
-   if (a == ENTupleColumnType::kIndex64 && b == ENTupleColumnType::kSplitIndex64) return true;
-   if (a == ENTupleColumnType::kSplitIndex64 && b == ENTupleColumnType::kIndex64) return true;
-   if (a == ENTupleColumnType::kReal32 && b == ENTupleColumnType::kSplitReal32) return true;
-   if (a == ENTupleColumnType::kSplitReal32 && b == ENTupleColumnType::kReal32) return true;
-   if (a == ENTupleColumnType::kReal64 && b == ENTupleColumnType::kSplitReal64) return true;
-   if (a == ENTupleColumnType::kSplitReal64 && b == ENTupleColumnType::kReal64) return true;
+   if (a == ROOT::ENTupleColumnType::kInt16 && b == ROOT::ENTupleColumnType::kSplitInt16) return true;
```
silverweed, line 350: At least in this function - but even file-wide - I'd like to have a `using ENTupleColumnType = ROOT::ENTupleColumnType` to avoid all the namespace noise...

tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
```diff
@@ -27,6 +27,10 @@
 
 namespace ROOT {
 
+class RLogChannel;
+/// Log channel for RNTuple diagnostics.
+ROOT::RLogChannel &NTupleLog();
+
```
hahnjo, line 33: Actually, do we want this in the public interface? Would it be sufficient to be in `ROOT::Internal`?

tree/ntuple/v7/inc/ROOT/RColumn.hxx
```diff
@@ -39,7 +39,7 @@ namespace ROOT::Experimental::Internal {
 // clang-format on
 class RColumn {
 private:
-   ENTupleColumnType fType;
+   ROOT::ENTupleColumnType fType;
```
hahnjo, line 42: Question here and everywhere else: Do we want to qualify the types in our own classes? I think it should not be needed due to `namespace` nesting, and it becomes redundant when we move other classes. Will we remove them later on?

tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
```diff
@@ -126,8 +126,6 @@ constexpr NTupleSize_t kInvalidNTupleIndex = std::uint64_t(-1);
 using DescriptorId_t = std::uint64_t;
 constexpr DescriptorId_t kInvalidDescriptorId = std::uint64_t(-1);
 
-namespace Experimental {
-
 /// Addresses a column element or field item relative to a particular cluster, instead of a global NTupleSize_t index
 class RNTupleLocalIndex {
```
hahnjo, line 130: We are missing a `using` declaration - or is this not needed because we recently renamed the class anyway?

tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
```diff
@@ -175,8 +175,6 @@ public:
    ROOT::NTupleSize_t GetIndexInCluster() const { return fIndexInCluster; }
 };
 
-namespace Experimental {
-
 /// RNTupleLocator payload that is common for object stores using 64bit location information.
```
hahnjo, line 178: There should be a `using` declaration for `RNTupleLocator`

hahnjo (commented):
some questions inline

tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
jblomer: I didn't include usings for this and the following class because while public, it should be very rare use. I was planning to add usings only for the common types. We anyway don't cover everything (e.g. constants like `kInvalidDescriptorId`). But this could be handled differently, too.

In general, the problem with usings is that we clutter compilation with deprecation warnings from all classes that are still in experimental, because they pick up the `ROOT::Experimental::Xyz` version then. For the typedefs, I felt it makes anyway sense to universally prefix them by `ROOT::`. But I wouldn't necessarily want that for all classes.

tree/ntuple/v7/inc/ROOT/RColumn.hxx
hahnjo: As discussed in the IO meeting, it's needed because otherwise it will pick up the (deprecated) `using` declarations in `ROOT::Experimental`. The approach may be to remove the deprecation message for now and only add it back once all classes are moved until we remove the `using` declarations before branching.

hahnjo (approved):
Please see my question about `NTupleLog` in https://github.com/root-project/root/pull/17665#discussion_r1946628952, but I leave that up to you.

PR #17668, opened 2025-02-07
[DF][ntuple] Only create `RVec`-type columns for vectors, arrays and `RVec` fields

enirolf:
Not all collection-type fields should be read as RVecs, given that some collection types have different semantic meaning (e.g., sets). With this change, only `std::vector`, `std::array` and `ROOT::RVec` collection fields are added to the data source as RVecs.

The implemented change corresponds to the way collection column types are determined in TTree-based RDFs.

### DIFF
--- a/tree/dataframe/src/RNTupleDS.cxx
+++ b/tree/dataframe/src/RNTupleDS.cxx
@@ -68,9 +68,8 @@ protected:
    void ConstructValue(void *where) const final { *static_cast<std::size_t *>(where) = 0; }
 
 public:
-   static std::string TypeName() { return "std::size_t"; }
    RRDFCardinalityField()
-      : ROOT::Experimental::RFieldBase("", TypeName(), ROOT::ENTupleStructure::kLeaf, false /* isSimple */)
+      : ROOT::Experimental::RFieldBase("", "std::size_t", ROOT::ENTupleStructure::kLeaf, false /* isSimple */)
    {
    }
    RRDFCardinalityField(RRDFCardinalityField &&other) = default;
--- a/tree/dataframe/src/RNTupleDS.cxx
+++ b/tree/dataframe/src/RNTupleDS.cxx
@@ -231,7 +231,7 @@ public:
 RNTupleDS::~RNTupleDS() = default;
 
 void RNTupleDS::AddField(const RNTupleDescriptor &desc, std::string_view colName, ROOT::DescriptorId_t fieldId,
-                         std::vector<RNTupleDS::RFieldInfo> fieldInfos)
+                         std::vector<RNTupleDS::RFieldInfo> fieldInfos, bool convertToRVec)
 {
    // As an example for the mapping of RNTuple fields to RDF columns, let's consider an RNTuple
    // using the following types and with a top-level field named "event" of type Event:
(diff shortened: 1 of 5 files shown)


tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -337,8 +340,14 @@ void RNTupleDS::AddField(const RNTupleDescriptor &desc, std::string_view colName
          valueField =
             std::make_unique<ROOT::Experimental::RArrayAsRVecField>("", std::move(valueField), fieldInfo.fNRepetitions);
       } else {
-         // Actual ROOT::RVec
-         valueField = std::make_unique<ROOT::Experimental::RRVecField>("", std::move(valueField));
+         // Actual collection. std::vector, std::array and RVecs get turned into RVecs. All other collection types keep
```
vepadulano, line 343: IIUC the `std::array` conversion part is still handled by the previous `if` branch through the `RArrayAsRVecField` right?

vepadulano (approved):
LGTM, thanks! Just a minor comment

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -135,6 +135,7 @@ class RArraySizeField final : public ROOT::Experimental::RFieldBase {
    }
 
 public:
+   static std::string TypeName() { return "std::size_t"; }
```
jblomer, line 138: I think here and for `RRDFCardinalityField`, we can try to remove `TypeName()`. This static method is only used for compile-time type names in the `RField<T>` constructor.

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -283,12 +284,14 @@ void RNTupleDS::AddField(const RNTupleDescriptor &desc, std::string_view colName
          fProtoFields.emplace_back(std::move(cardinalityField));
 
          for (const auto &f : desc.GetFieldIterable(fieldDesc.GetId())) {
-            AddField(desc, std::string(colName) + "." + f.GetFieldName(), f.GetId(), fieldInfos);
+            AddField(desc, std::string(colName) + "." + f.GetFieldName(), f.GetId(), fieldInfos, true);
```
jblomer, line 287: Unrelated to this change: I think with the revised untyped collections, this `if` branch can be merged with the `else` branch.

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -337,8 +340,14 @@ void RNTupleDS::AddField(const RNTupleDescriptor &desc, std::string_view colName
...
       } else {
-         // Actual ROOT::RVec
-         valueField = std::make_unique<ROOT::Experimental::RRVecField>("", std::move(valueField));
+         // Actual collection. A std::vector or ROOT::RVec gets added as a ROOT::RVec. All other collection types keep
+         // their original type.
+         if (convertToRVec) {
+            valueField = std::make_unique<ROOT::Experimental::RRVecField>("", std::move(valueField));
+         } else {
+            auto outerFieldType = desc.GetFieldDescriptor(fieldInfo.fFieldId).GetTypeName();
+            valueField = RFieldBase::Create("", outerFieldType).Unwrap();
```
jblomer, line 349: I think this may break the projection of inner collection members as individual collections. So, if you have
an `std::set<Point>` with `struct Point { float x,y; };`, are the columns for `x`, and `y` correctly exposed as `std::set<float>`?

hahnjo: I think we cannot expose the individual columns as `std::set`: Even if the complete `Point` is unique in the set, the individual `x` and `y` may not. I think we have to expose them as sequence collection.

jblomer (approved):

PR #17669, opened 2025-02-07
[tutorials] Don't rely on conversion of Python tuple to TColorNumber

guitargeek:
The conversion of Python tuples to TColorNumber doesn't work for older Python versions. Therefore, `TColor::GetColor()` is brought back to the tutorials, following up on 8eefabd.

The underlying problem is a wrong overload resolution because of implicit conversion from double to integer, see also the error that one gets for example with Python 3.9:
```txt
/Users/couet/git/couet-root/tutorials/analysis/dataframe/df105_WBosonAnalysis.py:176: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  h.SetFillColor(color)
```

Comments are added to clarify that this would not be needed with Python 3.11 onwards, reminding us that we have the option to revert this commit once the minimum Python version is 3.11.


couet (approved):
LGTM

Issue #14556, opened 2024-02-04
[ROOT-3452] Suggestions for Minimizer documentation


vepadulano:
### Explain what you would like to see improved and how.

From https://its.cern.ch/jira/browse/ROOT-3452

I have some specific suggestions on improving the documentation for the minimizer classes:

* TFitterMinuit should have a deprecation warning and point to the new style classes

* It would be useful to have the ROOT::Math::Minimizer page give a simple example of its use, or at least point to the tutorial NumericalMinimization.C

* ROOT::Math::Minimizer::SetLimitedVariable is undocumented, and the arguments are unnamed. Just putting names on the arguments would help.

* The page for ROOT::Math::IBaseFunctionMultiDim could specifically mention ROOT::Math::Functor, since that seems the most appropriate way to use one's own function with the fitting framework.

I hope these can be added \- they would make the documentation a lot easier for users new to the whole fitting framework.

### ROOT version

Any

### Installation method

Any

### Operating system

Any

ferdymercury:
> * TFitterMinuit should have a deprecation warning

Does it refer to `etc/plugins/TVirtualFitter/P030_TFitterMinuit.C` or to what class exactly?

PR #17679, opened 2025-02-10
[docu] improve Minimizer documentation and do not mention TFitterMinuit

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/14556

## Checklist:

- [ ] tested changes locally


etc/plugins/TVirtualFitter/P030_TFitterMinuit.C
```diff
@@ -1,5 +1,5 @@
 void P030_TFitterMinuit()
 {
-   gPluginMgr->AddHandler("TVirtualFitter", "Minuit2", "TFitterMinuit",
-      "Minuit2", "TFitterMinuit(Int_t)");
+   //gPluginMgr->AddHandler("TVirtualFitter", "Minuit2", "TFitterMinuit",
```
guitargeek, line 3: You can remove the whole file, and also the corresponding `TFitterFumili` plugin.

The implementations are long gone, removed in this commit:
465e8ee5589349e91a736e5f56bc9ff43780db2f

By the way, `math/minuit2/doc/index.txt` also has some comments about `TFitterMinuit` and `TFitterFumili` that should be removed.

math/mathcore/inc/Math/IFunction.h
```diff
@@ -54,6 +54,8 @@ namespace ROOT {
           Derived classes must implement the pure private virtual method DoEval(const double *) for the
           function evaluation in addition to NDim() and Clone().
 
+          @see ROOT::Math::Functor is an ppropriate way to use one's own function with the fitting framework.
```
guitargeek, line 57:
```suggestion
          @see ROOT::Math::Functor is the recommended way to use one's own function with the fitting framework.
```
Typo, and maybe say "recommended" instead of "appropriate" (which is clearer IMHO)

math/mathcore/inc/Math/Minimizer.h
```diff
@@ -171,18 +171,30 @@ class Minimizer {
    /// set initial covariance matrix
    virtual bool SetCovariance(std::span<const double> cov, unsigned int nrow);
 
-   /// set a new lower limit variable  (override if minimizer supports them )
-   virtual bool SetLowerLimitedVariable(unsigned int  ivar , const std::string & name , double val , double step , double lower ) {
-      return SetLimitedVariable(ivar, name, val, step, lower, std::numeric_limits<double>::infinity() );
+   /// set a new lower limit variable  (override if minimizer supports them), leave upper bound unlimited
```
guitargeek, line 174:
```suggestion
   /// Set a new lower limit variable  (override if minimizer supports them), leave upper bound unlimited.
```
If you already change the line, you can also follow the convention of starting the doxygen brief summaries with a capital letter and ending them with a period.

roofit/roofit/test/testFitPerf.cxx
```diff
@@ -34,7 +34,6 @@
 #include "TStopwatch.h"
 
 #include "TVirtualFitter.h"
-// #include "TFitterMinuit.h"
```
guitargeek, line 37: Why not remove also the other commented-out includes in this file?

guitargeek (changes requested):
Thank you very much! Just a few comments.

And maybe let's not skip the CI, as we're also changing plugin code

guitargeek (approved):
Thank you so much!

PR #17684, opened 2025-02-10
[ntuple] Store number of entries in the `RNTupleProcessor`

enirolf:
Calling `GetNEntries` on the page source each time is expensive, so instead we store this as a member of the processor.

Mostly relevant for the `RNTupleSingleProcesor`, because the chain processor already is composed of other processors, and the join processor will do the same soon.


vepadulano (approved):
LGTM!

jblomer (approved):
LGTM but I think the `fIsConnected` member can be removed altogether.

enirolf:
> LGTM but I think the `fIsConnected` member can be removed altogether.

It is, right? 🤔

PR #17687, opened 2025-02-10
[ntuple] Assign names to `RNTupleProcessor`s 

enirolf:
Naming `RNTupleProcessors` is necessary to make the `RNTupleJoinProcessor` composable. Currently, we use the name of the auxiliary ntuple as prefix to the field names, but when this is abstracted to another processor, we need something else to identify these fields. Adding names to processors also fixes the current issue where two RNTuples with the same name cannot be joined. 

Providing a name is optional; by default, for single processors the name of the ntuple is used. For chain processors, the name of the first ntuple/inner processor in the chain is used. For join processors, the name of the primary ntuple is used.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
@@ -271,6 +271,14 @@ public:
    RIterator begin() { return RIterator(*this, 0); }
    RIterator end() { return RIterator(*this, ROOT::kInvalidNTupleIndex); }
 
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Create an `RNTupleProcessor` for a single RNTuple.
+   ///
+   /// \param[in] ntuple The name and storage location of the RNTuple to process.
+   /// \param[in] model An RNTupleModel specifying which fields can be read by the processor. If no model is provided,
+   /// one will be created based on the descriptor of the first ntuple specified.
+   ///
+   /// \return A pointer to the newly created RNTupleProcessor.
    static std::unique_ptr<RNTupleProcessor>
    Create(const RNTupleOpenSpec &ntuple, std::unique_ptr<RNTupleModel> model = nullptr);
 
--- a/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
@@ -168,10 +168,7 @@ protected:
    ///
    /// \note Before processing, a model *must* exist. However, this is handled downstream by the RNTupleProcessor's
    /// factory functions (CreateSingle, CreateChain and CreateJoin) and constructors.
-   RNTupleProcessor(const std::vector<RNTupleOpenSpec> &ntuples, std::unique_ptr<RNTupleModel> model)
-      : fNTuples(ntuples), fModel(std::move(model))
-   {
-   }
+   RNTupleProcessor(std::unique_ptr<RNTupleModel> model) : fModel(std::move(model)) {}
 
 public:
    RNTupleProcessor(const RNTupleProcessor &) = delete;
--- a/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
@@ -122,6 +122,7 @@ protected:
       bool IsAuxiliary() const { return fNTupleIdx > 0; }
    };
 
+   std::string fProcessorName;
    std::vector<RNTupleOpenSpec> fNTuples;
    std::unique_ptr<REntry> fEntry;
    std::unique_ptr<Internal::RPageSource> fPageSource;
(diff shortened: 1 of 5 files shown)


tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -270,6 +303,21 @@ public:
    static std::unique_ptr<RNTupleProcessor>
    CreateChain(const std::vector<RNTupleOpenSpec> &ntuples, std::unique_ptr<RNTupleModel> model = nullptr);
 
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Create a new RNTuple processor chain for vertical combinations of RNTuples.
```
vepadulano, line 307: The docstring here and in the `Create` overloads for the single RNTupleProcessor could be made a bit more symmetric, i.e. if for the single processor we have

> Create an `RNTupleProcessor` for a single RNTuple.

for the chain processor could be

> Create an `RNTupleProcessor` that vertically combines multiple RNTuples in a chain.

Or something similar

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -300,6 +363,29 @@ public:
    static std::unique_ptr<RNTupleProcessor> CreateJoin(const std::vector<RNTupleOpenSpec> &ntuples,
                                                        const std::vector<std::string> &joinFields,
                                                        std::vector<std::unique_ptr<RNTupleModel>> models = {});
+
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Create a new RNTuple processor for horizontally combined RNTuples.
```
vepadulano, line 368: Similar comment here w.r.t. aligning the docs with other RNTupleProcessor derived classes factories.

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -300,6 +363,29 @@ public:
...
+   ///
+   /// \param[in] ntuples A list specifying the names and locations of the ntuples to process. The first ntuple in the
+   /// list will be considered the primary ntuple and drives the processor iteration loop. Subsequent ntuples are
+   /// considered auxiliary, whose entries to be read are determined by the primary ntuple (which does not necessarily
+   /// have to be sequential).
```
vepadulano, line 373:
```suggestion
   /// considered auxiliary, whose entries to be read are determined by the primary ntuple (which do not necessarily
   /// have to be sequential).
```

*iff* the verb refers to the entries of the auxiliary ntuples.

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -300,6 +363,29 @@ public:
...
+   /// \param[in] joinFields The names of the fields on which to join, in case the specified ntuples are unaligned.
+   /// The join is made based on the combined join field values, and therefore each field has to be present in each
+   /// specified RNTuple. If an empty list is provided, it is assumed that the specified ntuple are fully aligned, and
+   /// `RNTupleIndex` will not be used.
+   /// \param[in] processorName The name to give to the processor. Use
+   /// CreateJoin(const std::vector<RNTupleOpenSpec> &, const std::vector<std::string> &, std::unique_ptr<RNTupleModel>)
+   /// to automatically use the name of the input RNTuple instead.
+   /// \param[in] models A list of models for the ntuples. This list must either contain a model for each ntuple in
+   /// `ntuples` (following the specification order), or be empty. When the list is empty, the default model (i.e.
+   //// Scontaining all fields) will be used for each ntuple.
```
vepadulano, line 383: Maybe a typo here?
```suggestion
   /// `ntuples` (following the specification order), or be empty. When the list is empty, the default model (i.e.
   //// containing all fields) will be used for each ntuple.
```

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -300,6 +363,29 @@ public:
...
+   ///
+   /// \return A pointer to the newly created RNTupleProcessor.
+   static std::unique_ptr<RNTupleProcessor>
+   CreateJoin(const std::vector<RNTupleOpenSpec> &ntuples, const std::vector<std::string> &joinFields,
+              std::string_view processorName, std::vector<std::unique_ptr<RNTupleModel>> models = {});
```
vepadulano, line 388: Looking again at this factory, one possibility to make the interface more prescriptive could be having two overloads, this one could just become
```suggestion
   static std::unique_ptr<RNTupleProcessor>
   CreateJoin(const std::vector<RNTupleOpenSpec> &ntuples, const std::vector<std::string> &joinFields,
              std::string_view processorName);
```
which would only use the default model, then we could have a new one
```cpp
   static std::unique_ptr<RNTupleProcessor>
   CreateJoin(
  const std::vector<std::pair<RNTupleOpenSpec, std::unique_ptr<RNTupleModel>> &ntuplesWithModels,
  const std::vector<std::string> &joinFields,
  std::string_view processorName);
```

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -68,18 +85,40 @@ ROOT::Experimental::RNTupleProcessor::CreateChain(std::vector<std::unique_ptr<RN
...
-      new RNTupleChainProcessor(std::move(innerProcessors), std::move(model)));
+      new RNTupleChainProcessor(std::move(innerProcessors), processorName, std::move(model)));
+}
+
+std::unique_ptr<ROOT::Experimental::RNTupleProcessor>
+ROOT::Experimental::RNTupleProcessor::CreateJoin(const std::vector<RNTupleOpenSpec> &ntuples,
+                                                 const std::vector<std::string> &joinFields,
+                                                 std::vector<std::unique_ptr<RNTupleModel>> models)
+{
+   if (ntuples.size() < 1)
```
vepadulano, line 113: Is there a reason not to use `empty` here?

vepadulano (changes requested):
Thanks! Some misc comments after a first read.

vepadulano (approved):
LGTM!

jblomer (approved):
LGTM!

PR #17689, opened 2025-02-10
[hist] in TTree::Draw, last bin should include vmax values

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://root-forum.cern.ch/t/bug-or-feature-in-ttree-draw/62862

Rounding issue reproducer::

`TH1F h("h","h", 100, -1000, 1e31*std::nextafter(float(0),INFINITY));  h.Fill(-999); h.Fill(0);  h.GetBinContent(100)`

changing to 1e32 makes it work.

It comes down to this line in TAxis:

`bin = 1 + int (fNbins*(x-fXmin)/(fXmax-fXmin) );`

so:

`int(100*1000./(1000.+1e-13)) is 99`
whereas
`int(100*1000./(1000.+1e-14)) is 100`

## Checklist:

- [x] tested changes locally

### DIFF
--- a/hist/hist/src/THLimitsFinder.cxx
+++ b/hist/hist/src/THLimitsFinder.cxx
@@ -401,5 +401,12 @@ void THLimitsFinder::OptimizeLimits(Int_t nbins, Int_t &newbins, Double_t &xmin,
       if (xmin +nbins*bw < umax) {nbins++; xmax = xmin +nbins*bw;}
       if (xmin > umin)           {nbins++; xmin = xmax -nbins*bw;}
    }
+   else {
+      xmax = std::max(xmax + 1e-12, std::nextafter(xmax,INFINITY));
+      // If we put the upper bin limit directly at xmax, then all values at xmax will go into the overflow and will be invisible.
+      // So shift it slightly to the right, by at least 1e-12.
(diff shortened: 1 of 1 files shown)


ferdymercury:
I would propose merging this soonish, since it's silently hiding data. (Hiding them in the overflow)

hageboeck:
> I would propose merging this soonish, since it's silently hiding data. (Hiding them in the overflow)

Hello @ferdymercury,

Thanks for your persistence on this issue! 🙂

I agree that it's an annoying "feature", so it needs to be fixed. I looked again today, and believe I found the underlying cause:
`OptimizeLimits` removes the first/last bin from the range based on a slightly arbitrary distance criterion.
By adding a check that min and max are within the axis range, I managed to fix the issue. You can see the relevant change in #19506. There is a good argument to doing it in that function, because the check I added uses the same floating-point arithmetic as the computation of the axis range, so it's guaranteed that min/max fall within the valid range (unless I didn't get < vs <= right, please have a look!).
#19605 is in draft for the moment, because I plan to add a test also.

ferdymercury:
Thanks Stephan for looking into it and proposing an alternative PR!

PR #17607, opened 2025-02-04
[CMake] Speed up copying of headers

hageboeck:
- ~~Move the copy of artifacts (e.g. tutorials) from build time to configure time, and use a faster copy function.~~
- At build time, replace the one-by-one copy of headers by copy in batches.

Here are some results:
| `make -j6 move_headers`  | New | Old |
|--------|--------|--------|
| 1st invocation | 10.7 s | 34.5 s |
| 2nd invocation | 2.7 s | 3.8 s |
| Invocation after touching a header in core/ | 3.1 s | 4.2 s | 
| Invocation after touching a header in RooFit/hs3 | 2.9 s | 3.9 s |

For the copy of artifacts, there is no agreement yet since it's more complicated. I will factor this part out into a second PR.

Fix #14953

PR #17692, opened 2025-02-11
Speed up the copy of artifacts

### LINKED ISSUE #14953: [cmake] Glob when copying headers and tutorials

We can copy all the tutorials and header files at once. We do not need to do these operations individually causing things like:

```bash
[ 45%] Copying /Users/sftnight/vvassilev/root-src/tutorials/proof/ProofSimple.h
[ 45%] Copying /Users/sftnight/vvassilev/root-src/tutorials/proof/ProofSimpleFile.C
[ 45%] Copying /Users/sftnight/vvassilev/root-src/tutorials/proof/ProofSimpleFile.h
[ 45%] Copying /Users/sftnight/vvassilev/root-src/tutorials/proof/ProofStdVect.C
```

When using the Makefile Generator that's quite slow for no good reason...

### ROOT version

master

### Installation method

N/A

### Operating system

N/A

### Additional context

_No response_

hageboeck:
This was factored out of #17607.

In contrast to the initial proposal in #17607, I managed to keep the artifact copy at build time. This should alleviate all problems where an out-of-date file might surprise developers.
I used `cmake -E copy_directory` instead of going file-by-file, which again significantly reduces build targets and makes the initial copy much faster. We don't gain on the second invocation of make, and when tutorials are touched, one pays a 5s penalty, but I think the 15s saved on first invocation and the reduced number of targets make this worthwhile.

| `make -j6 move_artifacts` | Old | New |
|--------|--------|--------|
| 1st invocation | 25.5 | 10.4 |
| 2nd invocation | 2.5 | 2.6 |
| Touch a RooFit tutorial | 2.5 | 7.8 | 

### For the people who wonder why we do this at all
- Although all tutorials that run as part of ctest are invoked with a path to the source directory, 77 tutorials break because they call `gROOT->GetTutorialDir()`, which points to the path where the `root.exe` is.
- Users might be used to finding the tutorials next to their `root.exe`, and we didn't yet have a discussion on whether or not and how this should be changed. 

Fix #14953.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -310,29 +310,33 @@ add_custom_command(OUTPUT ${stamp_file}
                    COMMAND ${CMAKE_COMMAND} -E touch ${stamp_file}
                    COMMENT "Copying directories such as etc, icons, fonts, js, ui5, etc. to build area")
 
+
+#---Copy files to the build area at build time---------------------------------
+set(directories_to_copy etc test icons fonts macros tutorials)
 if(http)
-set(jsroot_files js/*)
+  list(APPEND directories_to_copy js)
 endif()
 if(webgui)
-set(openui5_files ui5/*)
+  list(APPEND directories_to_copy ui5)
 endif()
 
-#---Copy files to the build area, with dependency---------------------------------
-file(GLOB_RECURSE artifact_files RELATIVE ${CMAKE_SOURCE_DIR} tutorials/* etc/* test/* icons/* fonts/* macros/* ${jsroot_files} ${openui5_files})
-set(artifact_files_builddir)
-foreach(artifact_file ${artifact_files})
-  # Filter out hsimple.root; someone might have created it in the src dir, and the hsimple.root
-  # target below will interfere.
-  if (NOT (artifact_file STREQUAL "tutorials/hsimple.root"))
-    add_custom_command(OUTPUT ${CMAKE_BINARY_DIR}/${artifact_file}
-      COMMAND ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_SOURCE_DIR}/${artifact_file} ${CMAKE_BINARY_DIR}/${artifact_file}
-      COMMENT "Copying ${CMAKE_SOURCE_DIR}/${artifact_file}"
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -310,29 +310,34 @@ add_custom_command(OUTPUT ${stamp_file}
...
+  # For target dependencies:
+  file(GLOB_RECURSE artifacts_in ${dir_to_copy}/*)
+  set(artifacts_out ${artifacts_in})
+  list(TRANSFORM artifacts_out REPLACE ${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR})
+  string(REGEX REPLACE ".*/" "" destination ${dir_to_copy})
+  list(APPEND all_artifacts ${artifacts_out})
+
+  add_custom_command(OUTPUT ${artifacts_out}
+      COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_SOURCE_DIR}/${dir_to_copy} ${CMAKE_BINARY_DIR}/${destination}
+      COMMENT "Copy ${DIR_TO_COPY}"
```
pcanal, line 333:
```suggestion
      COMMENT "Copy ${dir_to_copy}"
```

hageboeck: Sorry, I hadn't pushed the last version. This wasn't supposed to be there.

CMakeLists.txt
```diff
@@ -310,29 +310,34 @@ add_custom_command(OUTPUT ${stamp_file}
...
-  endif()
+  list(APPEND directories_to_copy ui5)
+endif()
+
+foreach(dir_to_copy ${directories_to_copy})
+  # For target dependencies:
+  file(GLOB_RECURSE artifacts_in ${dir_to_copy}/*)
+  set(artifacts_out ${artifacts_in})
+  list(TRANSFORM artifacts_out REPLACE ${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR})
+  string(REGEX REPLACE ".*/" "" destination ${dir_to_copy})
```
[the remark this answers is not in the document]
hageboeck: > Why is this substitution needed?

Not needed, any more. I'll remove it.

PR #17699, opened 2025-02-12
[ntuple] Properly late-model extend projected fields in the merger

silverweed:
The RNTupleMerger needs to bypass RUpdater and directly use RNTupleModelChangeset because it doesn't interface with a RNTupleWriter. So far it was duplicating the functionality of RUpdater::AddField() and RUpdater::AddProjectedField(), but the latter was incomplete and therefore late-model extending a projected field was broken. This PR moved the functionality of Add*Field() from RUpdater to RNTupleModelChangeset and has the Merger call those methods instead of duplicating their functionality. This fixes the bug. In addition, two unit tests are added related to projected fields, one of which covers this feature.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleMerger.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleMerger.hxx
@@ -34,7 +34,8 @@ namespace Experimental::Internal {
 
 enum class ENTupleMergingMode {
    /// The merger will discard all columns that aren't present in the prototype model (i.e. the model of the first
-   /// source)
+   /// source); also all subsequent RNTuples must contain at least all the columns that are present in the prototype
+   /// model
    kFilter,
    /// The merger will refuse to merge any 2 RNTuples whose schema doesn't match exactly
    kStrict,
--- a/tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
@@ -94,28 +94,6 @@ public:
    bool IsEmpty() const { return fFieldZero->begin() == fFieldZero->end(); }
 };
 
-// clang-format off
-/**
-\class ROOT::Experimental::Internal::RNTupleModelChangeset
-\ingroup NTuple
-\brief The incremental changes to a `RNTupleModel`
-
-Represents a set of alterations to a `RNTupleModel` that happened after the model is used to initialize a `RPageSink`
-instance. This object can be used to communicate metadata updates to a `RPageSink`.
-You will not normally use this directly; see `RNTupleModel::RUpdater` instead.
-*/
-// clang-format on
-struct RNTupleModelChangeset {
-   RNTupleModel &fModel;
-   /// Points to the fields in fModel that were added as part of an updater transaction
-   std::vector<RFieldBase *> fAddedFields;
-   /// Points to the projected fields in fModel that were added as part of an updater transaction
-   std::vector<RFieldBase *> fAddedProjectedFields;
-
-   RNTupleModelChangeset(RNTupleModel &model) : fModel(model) {}
-   bool IsEmpty() const { return fAddedFields.empty() && fAddedProjectedFields.empty(); }
-};
-
 } // namespace Internal
 
 // clang-format off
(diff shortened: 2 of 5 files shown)


tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -246,8 +187,8 @@ private:
    RNTupleModel(std::unique_ptr<RFieldZero> fieldZero);
 
 public:
-   RNTupleModel(const RNTupleModel&) = delete;
-   RNTupleModel& operator =(const RNTupleModel&) = delete;
+   RNTupleModel(const RNTupleModel &) = delete;
+   RNTupleModel &operator=(const RNTupleModel &) = delete;
```
enirolf, line 191: These are unrelated to the rest of the commit

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -402,6 +343,74 @@ public:
    std::size_t EstimateWriteMemoryUsage(const RNTupleWriteOptions &options = RNTupleWriteOptions()) const;
 };
 
+namespace Internal {
```
enirolf, line 346: Why move the classes? This makes it kind of difficult to compare the changes you've made inside these classes..

tree/ntuple/v7/src/RNTupleModel.cxx
```diff
@@ -240,7 +252,8 @@ void ROOT::Experimental::RNTupleModel::EnsureNotBare() const
 
 ROOT::Experimental::RNTupleModel::RNTupleModel(std::unique_ptr<RFieldZero> fieldZero)
    : fFieldZero(std::move(fieldZero)), fModelId(GetNewModelId()), fSchemaId(fModelId)
-{}
+{
+}
```
enirolf, line 256: unrelated change

tree/ntuple/v7/inc/ROOT/RNTupleMerger.hxx
```diff
@@ -33,7 +33,7 @@ namespace ROOT::Experimental::Internal {
 
 enum class ENTupleMergingMode {
    /// The merger will discard all columns that aren't present in the prototype model (i.e. the model of the first
-   /// source)
+   /// source); also all subsequent RNTuples must have at least all the columns that are present in the prototype model
```
enirolf, line 36:
```suggestion
   /// source); also, all subsequent RNTuples must contain at least all the columns that are present in the prototype model
```

tree/ntuple/v7/test/ntuple_merger.cxx
```diff
@@ -1423,6 +1423,190 @@ TEST(RNTupleMerger, MergeProjectedFields)
...
+      // Gather the input sources
+      std::vector<std::unique_ptr<RPageSource>> sources;
+      sources.push_back(RPageSource::Create("ntuple", fileGuard1.GetPath(), RNTupleReadOptions()));
+      sources.push_back(RPageSource::Create("ntuple", fileGuard2.GetPath(), RNTupleReadOptions()));
+      std::vector<RPageSource *> sourcePtrs;
+      for (const auto &s : sources) {
+         sourcePtrs.push_back(s.get());
+      }
+
+      // Now Merge the inputs
```
enirolf, line 1545:
```suggestion
      // Now merge the inputs
```

enirolf (approved):
LGMT in principle, some minor comments

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
silverweed: Because otherwise it cannot compile due to needing the fwd-decl in one class but also the complete type in the other

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -560,11 +560,14 @@ static void ExtendDestinationModel(std::span<const RFieldDescriptor *> newFields
...
-      if (fieldDesc->IsProjectedField())
-         changeset.fAddedProjectedFields.emplace_back(field.get());
-      else
-         changeset.fAddedFields.emplace_back(field.get());
-      changeset.fModel.AddField(std::move(field));
+      if (fieldDesc->IsProjectedField()) {
+         const auto sourceId = fieldDesc->GetProjectionSourceId();
+         const auto &sourceField = mergeData.fSrcDescriptor->GetFieldDescriptor(sourceId);
+         changeset.AddProjectedField(std::move(field),
+                                     [dstName = sourceField.GetFieldName()](const std::string &) { return dstName; });
```
hahnjo, line 567: I think this is insufficient for field hierarchies with subfields. This function maps all projected fields to the same source field (I was a bit confused by `dstName`, should the capture be called `srcName`?). I think you have to map the entire field hierarchy as we do in `RPageSinkBuf`. There we use an `Internal` function to add a `FieldMap_t` directly, probably as a short-cut to not create a map of field names first to use in a lambda...

hahnjo (changes requested):

jblomer (commented):
In principle looks good to me but see @hahnjo's comment

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -558,13 +558,34 @@ static void ExtendDestinationModel(std::span<const RFieldDescriptor *> newFields
...
-      if (fieldDesc->IsProjectedField())
+      if (!fieldDesc->IsProjectedField()) {
+         auto field = fieldDesc->CreateField(*mergeData.fSrcDescriptor);
+         changeset.AddField(std::move(field));
+      }
+   }
+   // ...then add all projected fields.
+   Internal::RProjectedFields::FieldMap_t fieldMap;
+   for (const auto *fieldDesc : newFields) {
+      if (fieldDesc->IsProjectedField()) {
```
jblomer, line 571: Tiny style point: In such a case (long `if` block as the only element of a `for` block), I'd find
```
    if (!fieldDesc->IsProjectedField())
      continue;

   ...
```
slightly more readable.  Feel free to ignore though.

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -558,13 +558,34 @@ static void ExtendDestinationModel(std::span<const RFieldDescriptor *> newFields
...
+         auto field = fieldDesc->CreateField(*mergeData.fSrcDescriptor);
+         const auto sourceId = fieldDesc->GetProjectionSourceId();
+         const auto &sourceField = dstModel.GetConstField(mergeData.fSrcDescriptor->GetQualifiedFieldName(sourceId));
+         fieldMap[field.get()] = &sourceField;
+         const auto &subFieldIds = fieldDesc->GetLinkIds();
+         int i = 0;
+         for (const auto &subfield : *field) {
+            const auto subFieldId = subFieldIds[i++];
```
jblomer, line 579: I don't think that's correct. The `*field` iterator walks through the entire subfield hierarchy, but the `GetLinkIds()` only gets the direct children. I think we should test this with an alias field of a `vector<CustomStruct>` field.

jblomer (approved):
LGTM

tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -558,13 +558,33 @@ static void ExtendDestinationModel(std::span<const RFieldDescriptor *> newFields
...
    changeset.fAddedFields.reserve(newFields.size());
+   // First add all non-projected fields...
    for (const auto *fieldDesc : newFields) {
+      if (!fieldDesc->IsProjectedField()) {
+         auto field = fieldDesc->CreateField(*mergeData.fSrcDescriptor);
+         changeset.AddField(std::move(field));
+      }
+   }
+   // ...then add all projected fields.
+   Internal::RProjectedFields::FieldMap_t fieldMap;
```
hahnjo, line 569: This does not look right, we need to have one `fieldMap` per projected field! I guess the variable should move into the loop, you probably need a second source with two projections as a test.

hahnjo (changes requested):

PR #17701, opened 2025-02-12
[ntuple] Add compression settings to RStagedCluster::RColumnInfo

silverweed:
The compression of a column range is fixed when we call StageCluster(), so we should use the same one when committing the cluster, rather than reading it from `fOpenColumnRanges` (since it may in principle have changed in the meantime).

## Checklist:

- [x] tested changes locally


hahnjo (approved):

enirolf:
FYI clang-format wasn't happy...

PR #17705, opened 2025-02-12
[core] Delete pre-C++17 code in RRangeCast

guitargeek:
Now that the minimum C++ standard is C++17, we can delete some code in RRangeCast.hxx that was written only for C++14 support.


pcanal (approved):
LGTM.  Thanks.

PR #17707, opened 2025-02-12
[tree] TTree::SetAlias() properly interpret the constant

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-4012
fyi @thiemonagel

Regression since https://github.com/root-project/root/commit/bfbfe8783d0a216e8efff5f4aaff991bdba47563

## Checklist:

- [x] tested changes locally
- [x] updated the docs (if necessary)

PR #17708, opened 2025-02-13
[tree] reserve use of default TTreeFormula constructor only for ROOT I/O

ferdymercury:
since fManager is null and could lead to crashes if object is used later on

See https://github.com/root-project/root/pull/17707#discussion_r1953453947

## Checklist:

- [x] tested changes locally


pcanal (approved):
Thanks.

pcanal:
@ferdymercury Can you also add this change to the release notes?

PR #17709, opened 2025-02-13
[RF][HF] Remove LinkDef entry for HistFactory ConfigParser

guitargeek:
The ConfigParser parses XML to HistFactory measurements and is only built when `xml=ON`. To avoid unused class rule warnings for `xml=OFF`, this commit suggests to remove the LinkDef entry, which is not necessary anyway. The `ConfigParser` is a helper class for this `hist2workspace` tool.


vepadulano (approved):
And the class does not really support I/O anyway so I believe this is a needed change.

PR #17711, opened 2025-02-13
[ntuple] fix namespaces after move out of ROOT::Experimental

silverweed:


enirolf (approved):

hahnjo (approved):
You may want to merge this early as soon as the first build confirmed ROOT is fully building again

PR #17712, opened 2025-02-13
Remove temporary and swap files from dependency tracking for tutorials

hageboeck:
This is a fix for ROOT-9688 (JIRA).

Since globbing is used to list artifacts, CMake can accidentally pick up files like `.<filename>.swp`, and start tracking dependencies for them. When the original gets deleted, CMake will generate an error.
Here, file endings such as `.d`, `.so`, `.bak` and similar are removed from dependency tracking.


CMakeLists.txt
```diff
@@ -321,8 +321,11 @@ if(webgui)
 endif()
 
 foreach(dir_to_copy ${directories_to_copy})
-  # For target dependencies:
+  # Create a list of files for which we want dependency tracking.
+  # Exclude certain endings, so CMake doesn't complain when they are deleted (JIRA#9688)
   file(GLOB_RECURSE artifacts_in ${dir_to_copy}/*)
+  list(FILTER artifacts_in EXCLUDE REGEX "\.(d$|swp|so|dylib|bak|old)$")
```
ferdymercury, line 327: Thanks a lot! Maybe add things like CMakeLists.txt.user or others listed in the .gitignore of ROOT, that generated when you open them in an IDE.
As well as .pcm files in case dictionaries are generated?

bellenot, line 327: OK, but what about `dll`, `exp`, `pcm` , `lib` , ... ?

hageboeck: I looked in .gitignore, and `xxx.user` doesn't seem to be listed. pcm is a good idea, I will add that!

hageboeck: As we just discussed, it's mostly for cosmetics to keep the dependencies cleaner. If you don't `dir` into the source tree and run from there, you should not see this issue.

hageboeck: I had a second look, and I don't think that `.pcm` needs to be added. We are not suppressing it in `.gitignore`, and there's not a single pcm in the source tree.

ferdymercury: > and `xxx.user` doesn't seem to be listed

I see here this specific one: https://github.com/root-project/root/blob/master/.gitignore#L22

pcanal: On the other hand if you list `|so|dylib` there is no reason not to list `|lib|dll`.  I.e. they are either all needed or all not needed :).

CMakeLists.txt
```diff
@@ -321,8 +321,11 @@ if(webgui)
 endif()
 
 foreach(dir_to_copy ${directories_to_copy})
-  # For target dependencies:
+  # Create a list of files for which we want dependency tracking.
+  # Exclude certain endings, e.g. ACLiC artifacts, so CMake doesn't complain when they are deleted (JIRA#9688)
   file(GLOB_RECURSE artifacts_in ${dir_to_copy}/*)
+  list(FILTER artifacts_in EXCLUDE REGEX "\.(d|swp|so|dylib|pcm|bak)$")
```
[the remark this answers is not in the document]
hageboeck: I concur that these maybe created, but we won't be able to cover all cases. Specifically the ones where you can choose the extension.
We now have ACliC covered and the case mentioned in the bug report, so I hope that we took it far enough now. Certainly, it's an improvement with regards to the previous situation where none of this was covered.

Let me also point out that it's good to keep the source directory clean, and that a simple `cmake .` fixes the problem when you delete the file.

Good to merge?

CMakeLists.txt
hageboeck: I will set an arbitrary cutoff now 🙂: Not enough root developers use QTCreator to warrant the inclusion of this pattern.

ferdymercury (approved):

CMakeLists.txt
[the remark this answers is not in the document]
pcanal: > Good to merge?

Yes.

> it's an improvement with regards to the previous situation where none of this was covered.

Indeed (but not as good as not using a glob and using an explicit registration).

> Let me also point out that it's good to keep the source directory clean

Not 'that' easy to avoid the files created by editors (the current filter addresses the `.bak` and `.swp`).  [Not easy mostly because neither the user nor git cares that they are there].

PR #17724, opened 2025-02-14
[Tree] Account correctly for top branches

dpiparo:
fixes [ROOT-7926](https://its.cern.ch/jira/browse/ROOT-7926)

### DIFF
--- a/tree/tree/src/TTree.cxx
+++ b/tree/tree/src/TTree.cxx
@@ -7307,7 +7307,9 @@ void TTree::Print(Option_t* option) const
       for (l=0;l<nl;l++) {
          leaf = (TLeaf *)const_cast<TTree*>(this)->GetListOfLeaves()->At(l);
          br   = leaf->GetBranch();
-         if (strchr(br->GetName(),'.')) {
+         auto bname = br->GetName();
+         auto pDot = strchr(bname,'.');
+         if ( pDot && pDot!=&bname[strlen(bname)-1] ) {
             count[l] = -1;
             count[keep] += br->GetZipBytes();
          } else {


tree/tree/src/TTree.cxx
```diff
@@ -7307,7 +7307,9 @@ void TTree::Print(Option_t* option) const
       for (l=0;l<nl;l++) {
          leaf = (TLeaf *)const_cast<TTree*>(this)->GetListOfLeaves()->At(l);
          br   = leaf->GetBranch();
-         if (strchr(br->GetName(),'.')) {
+         auto bname = br->GetName();
```
pcanal, line 7310: I wonder if `br == br->GetMother()` would work as well?

pcanal:
Can we add a test for this? (maybe in `tree/tree/test/TTreeRegressions.cxx `)

tree/tree/src/TTree.cxx
dpiparo: Maybe: in what cases a branch has the same of its mother?

dpiparo:
> Can we add a test for this? (maybe in `tree/tree/test/TTreeRegressions.cxx `)

definitively.

tree/tree/src/TTree.cxx
pcanal: `br == br->GetMother()` is not checking the name but only the pointer.  It is essentially checking 'is this branch its own (top level) mother' which is 'true' only for the top level branches.

ferdymercury: I tried here: https://github.com/root-project/root/pull/17777

dpiparo:
superseded by https://github.com/root-project/root/pull/17777

PR #17726, opened 2025-02-14
[ntuple] Rename `RNTupleIndex` to `RNTupleJoinTable`

enirolf:
The term "index" is used in various other contexts. The functionality of the class is also inherently different from what is considered an "index" in database terms (where the name choice initially originated from). Hence, this rename should remove any naming ambiguity and better describe the actual functionality of the class.


jblomer (approved):
LGTM

hahnjo (approved):

PR #17730, opened 2025-02-15
[math] Fix warnings related to `snprintf`

guitargeek:
Fix warnigs related to `snprintf`, like for example in TSplot.cxx:

```txt
/home/rembserj/code/root/math/splot/src/TSPlot.cxx: In member function ‘void TSPlot::FillXvarHists(Int_t)’:
/home/rembserj/code/root/math/splot/src/TSPlot.cxx:535:39: error: ‘__builtin___snprintf_chk’ output may be truncated before the last format character [-Werror=format-truncation=]
  535 |       snprintf(name,sizeof(name), "x%d", i);
      |
```

### DIFF
--- a/math/splot/src/TSPlot.cxx
+++ b/math/splot/src/TSPlot.cxx
@@ -19,6 +19,9 @@
 #include "TMath.h"
 #include "snprintf.h"
 
+#include <sstream>
+#include <string>
+
 extern void Yields(Int_t &, Double_t *, Double_t &f, Double_t *x, Int_t iflag);
 
 ClassImp(TSPlot);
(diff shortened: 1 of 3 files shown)


math/splot/src/TSPlot.cxx
```diff
@@ -639,12 +632,13 @@ void TSPlot::FillYpdfHists(Int_t nbins)
          return;
    }
 
-   char name[34];
    for (ispecies=0; ispecies<fNSpecies; ispecies++){
       for (i=0; i<fNy; i++){
-         snprintf(name,sizeof(name), "pdf_species%d_y%d", ispecies, i);
+         std::stringstream nameStream;
+         nameStream << "pdf_species" << ispecies << "_y" << i;
```
pcanal, line 638: Out of curiosity.  Why a `std::stringsream` instead of `operator+ for std::string` here or vice-et-versa?

pcanal (approved):
LGTM

math/splot/src/TSPlot.cxx
guitargeek: I use `operator+` when it's not too many elements that need to be concatenated, and otherwise `stringstream`.

PR #17736, opened 2025-02-17
[tmva][sofie] Support for SYCL in TMVA SOFIE

sanjibansg:
This PR adds the support for calling inference in SOFIE with SYCL. This is based on the work previously done by Ioanna-Maria Panagou (@joannapng).

### DIFF
--- a/tmva/sofie/CMakeLists.txt
+++ b/tmva/sofie/CMakeLists.txt
@@ -83,8 +83,52 @@ ROOT_STANDARD_LIBRARY_PACKAGE(ROOTTMVASofie
 
 target_include_directories(ROOTTMVASofie PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>)
+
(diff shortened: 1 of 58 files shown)


lmoneta:
> Hi @sanjibansg ,
> 
> doesn't this deserve a bit of discussion (PPP meeting for example)? The model by which SYCL is used together with ROOT has not really been discussed so far.

Yes, The integration of SYCL in ROOT would require a wider discussion. However, this PR is not meant to be integrated now in ROOT, but to have a working prototype backend of SOFIE in SYCL which can be used for benchmarking withg other GPU solutions we plan to develop

guitargeek:
I'm closing this PR (which also has significant merge conflicts), since we decided that the inference code emit for heterogeneous architectures will not be implemented in ROOT, but in a standalone repository:
  * https://github.com/ML4EP/SOFIE

SOFIE in ROOT will stick with only the ONNX parser and regular CPU code emit capabilities.

Issue #17461, opened 2025-01-21
`builtin_clang=OFF` broken because of missing transitive dependencies


hahnjo:
Since the upgrade to LLVM 16, building ROOT with `builtin_clang=OFF` fails to link. It can be "fixed" with the following diff:
```diff
diff --git a/interpreter/cling/lib/Interpreter/CMakeLists.txt b/interpreter/cling/lib/Interpreter/CMakeLists.txt
index e397da97f6..d0819c4667 100644
--- a/interpreter/cling/lib/Interpreter/CMakeLists.txt
+++ b/interpreter/cling/lib/Interpreter/CMakeLists.txt
@@ -14,13 +14,16 @@ set(LIBS
   clangParse
   clangSema
   clangAnalysis
+  clangASTMatchers
   clangEdit
   clangRewrite
   clangRewriteFrontend
   clangSerialization
+  clangAPINotes
   clangAST
   clangBasic
   clangLex
+  clangSupport
 )

 set(LLVM_LINK_COMPONENTS
@@ -29,6 +32,8 @@ set(LLVM_LINK_COMPONENTS
   coroutines
   coverage
   executionengine
+  FrontendDriver
+  FrontendHLSL
   ipo
   jitlink
   lto
```

However, this is incorrect because it just specifies the missing transitive dependencies. The proper fix is to correctly load `ClangConfig.cmake` which specifies the library targets *and their dependencies*.

PR #17737, opened 2025-02-17
[interpreter] Use `find_package({LLVM,Clang})`

hahnjo:
This gives CMake full information about (transitive) library dependencies.

Fixes https://github.com/root-project/root/issues/17461


devajithvs (approved):
LGTM

vgvassilev (approved):
Great to see that’s finally possible. Did you make sure it works with -Dbuiltin_clang,llvm=Off?

PR #17738, opened 2025-02-17
Some cleanups of CMake options

hahnjo:
`cxxmodules` was deprecated at https://github.com/root-project/root/pull/12746

### DIFF
--- a/cmake/modules/RootBuildOptions.cmake
+++ b/cmake/modules/RootBuildOptions.cmake
@@ -386,7 +386,7 @@ endif()
 #---Removed options------------------------------------------------------------
 # Please notify SPI when adding to this list
 foreach(opt afdsmgrd afs alien bonjour builtin_afterimage castor chirp cxx11 cxx14 cxx17
-        exceptions geocad gfal glite globus gsl_shared hdfs html ios jemalloc krb5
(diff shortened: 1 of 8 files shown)


cmake/modules/RootBuildOptions.cmake
```diff
@@ -409,11 +409,6 @@ foreach(opt minuit2_omp minuit2_mpi)
   endif()
 endforeach()
 
-#---Replaced options--------------------------------------------------------------------------
-if(python)
```
pcanal, line 413: Beside saving 3 lines of code, what is the advantage of removing this warning?  Users might very well 'upgrade' from a version that had the `python` option directly to `v6.38` without first testing the intermediaries.

hahnjo: Please see the commit message (that's why I write them...):
> It was modified in commit 5410bb44676ee7bd87581a073b6644513f74aac9, but that never worked because it was also added to the list of removed options, which will issue a `FATAL_ERROR` if enabled.

In other words, such user would never see the warning anyway.

pcanal: Fair enough.  Thanks.

pcanal (approved):
LGTM.

PR #15750, opened 2024-06-04
[DF] Add initial implementation for snapshotting to RNTuple

enirolf:
This PR adds a first iteration of snapshotting to RNTuple from an RDataFrame. It uses the existing `Snapshot` interface, with an addition to `RSnapshotOptions`, `kOutputFormat`. This option can be set to write to either TTree, RNTuple, or take the default choice. Currently the default choice means that TTree will be used as the output format, also if the input format is RNTuple. In this case, a warning will be issued to the user. With the changes in this PR, snapshotting from TTree to RNTuple is not yet possible, because it requires additional conversion for certain branch types considered out of scope for this PR, similar to what is currently done with the RNTupleImport. Using this functionality a follow-up PR will add support for Snapshots from TTree to RNTuple.

## Implementation
As mentioned, the existing `Snapshot` interface is used. A new `SnapshotRNTupleHelper` has been created to handle the creation and writing of the RNTuple, akin to the existing `SnapshotHelper` (which has been renamed to `SnapshotTTreeHelper` for consistency).

### RLoopManager data source initialization (rev bbf221f)
The snapshot action creates a new loop manager which manages the snapshotted data set. The loop manager gets initialized before the actual snapshotting takes place. Originally, the pointer to the data source owned by the loop manager was marked as `const`. Because the RNTuple's data source _has_ to be created after the loop manager, for this PR the `const` qualifier has been dropped.

## Current limitations and follow-ups
This PR adds the minimal functionality for (single-threaded) snapshotting to RNTuple. A number of follow-ups are foreseen:

### RNTuple write options
Currently no RNTuple-specific write options have been added to `RSnapshotOptions` yet, except for compression settings which were already present as an option. Adding (a subset) of the other `RNTupleWriteOptions` is trivial.

### Default compression settings
`RSnapshotOptions`' default compression setting is 101 (Zlib). However, RNTuple's default compression setting is 505 (zstd). We could change the default compression setting to `kInherit` and decide which settings to use according to the target data format (unless explicitly set by the user, of course).

### Multithreaded snapshotting
This PR only adds single-threaded RNTuple snapshotting. Multithreaded (and parallel) snapshotting will be addressed in a follow-up PR.

## Tests

Corresponding `roottest` PR: https://github.com/root-project/roottest/pull/1178

Tests for Windows have been disabled, due to permission denied-errors related to trying to recreate currently open TFiles. The regular snapshot tests have also been disabled for Windows, presumably for the same reason.

PR #17739, opened 2025-02-17
[df] Allow late data source attachment

vepadulano:
Extra logic in RLoopManager helping https://github.com/root-project/root/pull/15750


tree/dataframe/src/RLoopManager.cxx
```diff
@@ -1387,3 +1390,11 @@ ROOT::Detail::RDF::CreateLMFromFile(std::string_view datasetName, const std::vec
 
 // outlined to pin virtual table
 ROOT::Detail::RDF::RLoopManager::~RLoopManager() = default;
+
+void ROOT::Detail::RDF::RLoopManager::SetDataSource(std::unique_ptr<ROOT::RDF::RDataSource> dataSource)
+{
+   if (dataSource) {
+      fDataSource = std::move(dataSource);
```
enirolf, line 1397: `fNSlots` has to be propagated to the data source as well:
```suggestion
      fDataSource = std::move(dataSource);
      fDataSource->SetNSlots(fNSlots);
```

tree/dataframe/inc/ROOT/RDF/RLoopManager.hxx
```diff
@@ -113,7 +113,15 @@ using ROOT::RDF::RDataSource;
 /// This class is responsible of running the event loop.
 class RLoopManager : public RNodeBase {
    using ColumnNames_t = std::vector<std::string>;
-   enum class ELoopType { kROOTFiles, kROOTFilesMT, kNoFiles, kNoFilesMT, kDataSource, kDataSourceMT };
+   enum class ELoopType {
+      kNoData,
```
enirolf, line 117: The difference between `kNoData` and `kNoFiles` might become ambiguous/confusing. What about `fInvalid` or `fNotRunnable` or something explicit like this to clearly signal the implication of this loop type?

enirolf (changes requested):
Thank you! Looks good to me in principle, but I have two comments before merging; one needs to be addressed before merging and the other one is more of a suggestion to clarify things.

enirolf (approved):
LGTM!

PR #17740, opened 2025-02-17
[ntuple] Merger: improve merging of deferred columns

silverweed:
# This Pull request:
fixes the previously-misbehaving case of merging 2 RNTuples with a common column where the column in the second source is deferred.
This requires synthesizing zero pages for the missing elements in order to fill the "holes" between the two sources. Since we already had code to do that when dealing with extra destination columns, I just reused it for this case.

A set of unit tests was added to test this case.

In addition, since this will be required in general, this PR changes the required type of the merger destination to `RPagePersistentSink` rather than the more abstract `RPageSink`.

## Checklist:

- [x] tested changes locally


tree/ntuple/v7/test/ntuple_merger.cxx
```diff
@@ -834,11 +834,21 @@ static bool VerifyPageCompression(const std::string_view fileName, std::uint32_t
...
+   if (compAlgo == 0) {
+      // This page might be uncompressed because compressing it wouldn't have saved space: check if that's the case.
+      const auto nbytesComp =
+         RNTupleCompressor::Zip(sealedPage.GetBuffer(), sealedPage.GetDataSize(), expectedComp, buffer.get());
+      if (nbytesComp == sealedPage.GetDataSize()) {
+         // Yep, the page was uncompressible. Accept that and return true.
+         return true;
+      }
+      // Not good, the page should have been compressed!
+      return false;
```
hahnjo, line 850: I think this should fall through to the existing code and print the error

hahnjo (approved):
One minor comment; otherwise LGTM (needs a second review as discussed)

PR #17743, opened 2025-02-17
[RDF] Add an option to run without cloned TH3Ds

hageboeck:
Cloning 3-dimensional histograms in multithreaded RDF can create significant memory pressure. To alleviate this pressure, this PR gives users the option share histograms across threads.

- Add a new function for TH3D that fills it atomically (only available in C++20)
- Add an action helper to RDF that either calls the atomic fill function (C++20) or falls back to locking and filling
- Add `ROOT::RDF::Experimental::ThreadsPerTH3(unsigned int)` to adjust the number of cloned histograms
- Add corresponding tests

A test with the usual rdf102 tutorial that's filling a single TH3 from 16 threads yields:
|  | Cloning | Atomics in TH3 (c++20) | Locks for TH3 (c++17) |
|--------|--------|--------|--------|
| Event throughput | 8 M / s | 6.6 M / s | 3.5M / s | 

I just saw that one thing is missing:
- [x] Remove doubling of helper functions such as `ScalarConstIterator`. I moved them out of the standard FillHelper, but sill redefined them in the ThreadSafeFillHelper ...
- [x] run clang-format
  **Edit**: Note that my clang-format (v 19.1.7) really produces the code that was pushed. The CI doesn't like it, but that's probably a different version.

Updates:
- It was suggested to hide the thread-safe fill function in TH3D, so it is now accessible only as `ROOT::Internal::FillThreadSafe(histogram, x, y, z, w)`.
- Move from completely disabling cloning to the `ThreadsPerTH3()` way outlined above.


hist/hist/inc/TH3.h
```diff
@@ -28,6 +28,14 @@
 class TH2D;
 class TProfile2D;
 
+namespace ROOT::Internal {
+/// Entrypoint for thread-safe filling from RDataFrame.
+template<typename T, typename... Args>
+auto FillThreadSafe(T & histo, Args ... args) -> decltype(histo.FillThreadSafe(args...), void()) {
+   histo.FillThreadSafe(args...);
+}
```
vepadulano, line 36: Probably nitpick since we're passing integer values anyway
```suggestion
auto FillThreadSafe(T & histo, Args &&... args) -> decltype(histo.FillThreadSafe(args...), void()) {
   histo.FillThreadSafe(std::forward<Args>(args)...);
}
```

hist/hist/test/test_TH3.cxx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+TEST(TH3D, FillAtomicNoSumW2)
+{
+   TH3D h0("h0", "h0", 50, 0, 50, 50, 0, 50, 50, 0, 50);
+   // Forgetting to call SumW2, we cannot fill with weights:
+   EXPECT_THROW(ROOT::Internal::FillThreadSafe(h0, 1., 2., 3., 4.), std::logic_error);
+}
+
+#endif
```
vepadulano, line 67: Missing newline at end of file

vepadulano (approved):
Super, thanks! I believe this is good for a first round of benchmarking with users! Only two very minor comments.

PR #15031, opened 2024-03-21
[RF] Re-implement `MakeNumpyDataFrame` with no interpreter calls

guitargeek:
Re-implement `MakeNumpyDataFrame` with no interpreter calls, and avoid a
explicit lifetime management of Python objects in the RDNumpyDS helper.

Like this, the former `RNumpyDS.hxx` helper could be moved to
`tree/dataframe`, as it doesn't import `Python.h` anymore.

The `RNumpyDS` also got renamed to `RVecDS`, because it takes RVecs and
not NumPy arrays directly.

PR #17745, opened 2025-02-17
[df] Extend lifetime of Python-bound TTree dataset

### LINKED ISSUE #17691: Crash when building dataframe from TChain
### Check duplicate issues.

- [x] Checked for duplicates

### Description

When I try to make dataframes from TChains I see a crash. The chains do not seem to be owned by the dataframe and when the chains go out of scope, accessing the dataframe will cause the crash.

### Reproducer

```python
import numpy
from ROOT import RDataFrame, RDF, TChain

# ----------------------------------
class Data:
    l_file = ['a.root', 'b.root', 'c.root']
# ----------------------------------
def _make_file(name : str) -> None:
    arr_x = numpy.random.uniform(0, 1, 1000)
    arr_y = numpy.random.uniform(0, 1, 1000)

    rdf   = RDF.FromNumpy({'x' : arr_x, 'y' : arr_y})
    rdf.Snapshot('tree', name)
# ----------------------------------
def _make_files():
    for name in Data.l_file:
        _make_file(name=name)
# ----------------------------------
def _get_rdf() -> RDataFrame:
    chain = TChain('tree')
    for file in Data.l_file:
        chain.Add(file)

    rdf = RDataFrame(chain)
    #rdf.chain = chain  # Uncomment this line to get the crash

    return rdf
# ----------------------------------
def main():
    _make_files()
    rdf = _get_rdf()
    rdf.Display().Print()

main()
```

### ROOT version

6.34.02

### Installation method

conda

### Operating system

Alma9

### Additional context

_No response_

vepadulano:
If a TTree dataset is created in a Python application and the RDataFrame constructor that takes in a `TTree &` parameter is used, the ownership of the TTree might result ambiguous for the user. Specifically, a Python user might expect the following to work out of the box:
```py
def foo():
    ch = ROOT.TChain()
    return ROOT.RDataFrame(ch)
```
Whereas it won't since the chain will become a dangling reference by the end of the function scope.

To cope with this unfortunate interaction of the C++ API and Python, we introduce an internal mechanism to keep track of the shared ownership of the *Python* proxy object. The implementation features a C++ wrapper in the ROOT CPython extension, which increases/decreases the Python reference count of a `PyObject *` data member. If this Python object is carefully given, i.e. if it is always pointing to the same underlying C++ object, then the reference counting will work correctly and destroy the C++ object only when there are no more Python objects referencing it. In order to keep the wrapper alive, a `std::any` data member is added to RLoopManager which represents the C++ counterpart of the Python lifeline. Within the same computation graph, different nodes share the same RLoopManager, so the Python object will stay alive as long as at least one node of the computation graph is still alive. If the same TTree is used (via a reference) by multiple computation graphs, then the wrapper mechanism will ensure that reference counting is correct across the different RDataFrame objects.

This PR fixes #17691

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdf_namespace.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdf_namespace.py
@@ -77,9 +77,20 @@ def _rdataframe(local_rdf, distributed_rdf):
     """
 
     def rdataframe(*args, **kwargs):
+        import ROOT
+        from libROOTPythonizations import PyObjRefCounterAsStdAny
+
         if kwargs.get("executor", None) is not None:
-            return distributed_rdf(*args, **kwargs)
+            rdf = distributed_rdf(*args, **kwargs)
+            rnode = ROOT.RDF.AsRNode(rdf._headnode.rdf_node)
         else:
-            return local_rdf(*args, **kwargs)
+            rdf = local_rdf(*args, **kwargs)
+            rnode = ROOT.RDF.AsRNode(rdf)
+
+        if args and isinstance(args[0], ROOT.TTree):
+            ROOT.Internal.RDF.SetTTreeLifeline(
+                rnode, PyObjRefCounterAsStdAny(args[0]))
+
+        return rdf
 
     return rdataframe
--- a/bindings/pyroot/pythonizations/src/PyROOTModule.cxx
+++ b/bindings/pyroot/pythonizations/src/PyROOTModule.cxx
@@ -25,6 +25,7 @@
 #include "RConfigure.h"
 
 // Standard
+#include <any>
 #include <string>
 #include <sstream>
 #include <utility>
@@ -58,6 +59,67 @@ PyObject *RegisterExecutorAlias(PyObject * /*self*/, PyObject *args)
    CPyCppyy::RegisterExecutorAlias(PyUnicode_AsUTF8(name), PyUnicode_AsUTF8(target));
 
    Py_RETURN_NONE;
+}
+
+/// \brief A PyObject wrapper to track reference counting of external objects
+///
+/// This wrapper can be useful in shared ownership scenarios when a C++ object
+/// is created on the Python side and there is no easy way to track its ownership
+/// on the C++ side. If multiple instances of this class are given the same
+/// Python proxy, they will increase/decrease its reference counting following
+/// Python rules and ensure proper destruction of the underlying C++ object when
+/// no other Python objects are referencing it.
+class PyObjRefCounter final {
+   PyObject *fObject{nullptr};
+
+   void Reset(PyObject *object)
+   {
+      if (fObject) {
+         Py_DECREF(fObject);
+         fObject = nullptr;
+      }
+      if (object) {
+         Py_INCREF(object);
+         fObject = object;
+      }
+   }
+
+public:
+   PyObjRefCounter(PyObject *object) { Reset(object); }
+
+   ~PyObjRefCounter() { Reset(nullptr); }
+
+   PyObjRefCounter(const PyObjRefCounter &other) { Reset(other.fObject); }
+
+   PyObjRefCounter &operator=(const PyObjRefCounter &other)
+   {
+      Reset(other.fObject);
+      return *this;
+   }
+
+   PyObjRefCounter(PyObjRefCounter &&other)
+   {
+      fObject = other.fObject;
+      other.fObject = nullptr;
+   }
+
+   PyObjRefCounter &operator=(PyObjRefCounter &&other)
+   {
+      fObject = other.fObject;
+      other.fObject = nullptr;
+      return *this;
+   }
+};
+
+PyObject *PyObjRefCounterAsStdAny(PyObject * /*self*/, PyObject *args)
+{
+   PyObject *object = nullptr;
+
+   PyArg_ParseTuple(args, "O:PyObjRefCounterAsStdAny", &object);
+
+   // The std::any is managed by Python
+   return CPyCppyy::Instance_FromVoidPtr(new std::any{std::in_place_type<PyObjRefCounter>, object}, "std::any",
+                                         /*python_owns=*/true);
 }
 
 } // namespace PyROOT
@@ -104,6 +166,8 @@ static PyMethodDef gPyROOTMethods[] = {
     (char *)"Register a custom converter that is a reference to an existing converter"},
    {(char *)"CPyCppyyRegisterExecutorAlias", (PyCFunction)PyROOT::RegisterExecutorAlias, METH_VARARGS,
     (char *)"Register a custom executor that is a reference to an existing executor"},
+   {(char *)"PyObjRefCounterAsStdAny", (PyCFunction)PyROOT::PyObjRefCounterAsStdAny, METH_VARARGS,
+    (char *)"Wrap a reference count to any Python object in a std::any for resource management in C++"},
    {NULL, NULL, 0, NULL}};
 
 struct module_state {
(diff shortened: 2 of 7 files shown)


tree/dataframe/src/RInterface.cxx
```diff
@@ -56,3 +56,8 @@ std::string ROOT::Internal::RDF::GetDataSourceLabel(const ROOT::RDF::RNode &node
       return "EmptyDS";
    }
 }
+
+void ROOT::Internal::RDF::SetTTreeLifeline(ROOT::RDF::RNode &node, const std::any &lifeline)
+{
+   node.GetLoopManager()->SetPythonTTreeLifeline(lifeline);
+}
```
guitargeek, line 63:
```suggestion
void ROOT::Internal::RDF::SetTTreeLifeline(ROOT::RDF::RNode &node, const std::any &lifeline)
{
   node.GetLoopManager()->SetTTreeLifeline(lifeline);
}
```

Mentioning Python in the names, or lifeline, is maybe too specific? This is a general way to manage resources used for the RLoopManager. Maybe reflect this generality in the names? Otherwise, the reader of the code might think it's actually coupled to C Python.

I mean you could even use this from C++ and it would also make sense:
```c++
ROOT::Internal::RDF::SetTTreeLifeline(rdf, std::make_any<std::shared_ptr<TTree>>(tree)); // or something like that
```

Another way to see the `std::any` here is as a smart and general alternative to `void*` and `std::shared_ptr` for managing resources:
https://devblogs.microsoft.com/cppblog/stdany-how-when-and-why/

bindings/pyroot/pythonizations/src/PyROOTModule.cxx
```diff
@@ -60,6 +61,69 @@ PyObject *RegisterExecutorAlias(PyObject * /*self*/, PyObject *args)
    Py_RETURN_NONE;
 }
 
+namespace Internal {
```
guitargeek, line 64:
```suggestion
```
Just a nitpick: but, `.cxx` sources are always internal anyway so `Internal`/`Detail` namespaces have no meaning IMHO

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -93,6 +94,7 @@ void ChangeBeginAndEndEntries(const RNode &node, Long64_t begin, Long64_t end);
 void ChangeSpec(const ROOT::RDF::RNode &node, ROOT::RDF::Experimental::RDatasetSpec &&spec);
 void TriggerRun(ROOT::RDF::RNode node);
 std::string GetDataSourceLabel(const ROOT::RDF::RNode &node);
+void SetTTreeLifeline(ROOT::RDF::RNode &node, const std::any &lifeline);
```
guitargeek, line 97:
```suggestion
void SetTTreeLifeline(ROOT::RDF::RNode &node, std::any lifeline);
```
...and then `std::move()` the `lifeline` inside the function to elide a copy. Also makes the signature more compact and highlights the fact that the wrapped type should be cheaply copyable.

Also, maybe consider a more general name like `SetSharedResource`? I'm thinking that we will also re-use this interface to manage other resources from the Python side, like NumPy arrays and so on

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdf_namespace.py
```diff
@@ -77,9 +77,20 @@ def _rdataframe(local_rdf, distributed_rdf):
     """
 
     def rdataframe(*args, **kwargs):
+        import ROOT
+        from libROOTPythonizations import CreatePyObjRefCountWrapper
```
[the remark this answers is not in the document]
guitargeek: ...on the other hand, the `std::any` is the wrapper. Well the name just feels a bit long

guitargeek (changes requested):
A few suggestions inline.

Also, we should consider to implement the RVecDS that was [refactored last year](https://github.com/root-project/root/pull/15031) with the same `std::any` pattern. It uses the `std::function<void()>` pattern, which is less general and I also think we should solve similar problems with the same pattern for a sustainable codebase.

bindings/pyroot/pythonizations/src/PyROOTModule.cxx
```diff
@@ -60,6 +61,69 @@ PyObject *RegisterExecutorAlias(PyObject * /*self*/, PyObject *args)
...
+   {
+      Reset(other.fObject);
+      other.Reset(nullptr);
+   }
+
+   PyObjRefCountWrapper &operator=(PyObjRefCountWrapper &&other)
+   {
+      Reset(other.fObject);
+      other.Reset(nullptr);
+      return *this;
```
guitargeek, line 111:
```suggestion
      fObject = other.fObject);
      other.fObject = nullptr;
      return *this;
```
To avoid incrementing and decrementing the reference counter, making moving truly cheap (can also be done for the move constructor).

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdf_namespace.py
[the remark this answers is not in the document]
guitargeek: Also, maybe we should use a name that is "invertible", because maybe we'll need one day a function to do the inverse, unpacking the Python type from the `std::any`? Not that I have a usecase now, it's just something so consider for naming.

Maybe `PyObjectToStdAny`, and the other would be `StdAnyToPyObject`?

bindings/pyroot/pythonizations/src/PyROOTModule.cxx
```diff
@@ -60,6 +61,69 @@ PyObject *RegisterExecutorAlias(PyObject * /*self*/, PyObject *args)
...
+namespace Internal {
+/// \brief A PyObject wrapper to track reference counting of external objects
+///
+/// This wrapper can be useful in shared ownership scenarios when a C++ object
+/// is created on the Python side and there is no easy way to track its ownership
+/// on the C++ side. If multiple instances of this class are given the same
+/// Python proxy, they will increase/decrease its reference counting following
+/// Python rules and ensure proper destruction of the underlying C++ object when
+/// no other Python objects are referencing it.
+class PyObjRefCountWrapper final {
```
guitargeek, line 73:
```suggestion
class PyObjRefCount final {
```
As said in the other comment: the `std::any` is the wrapper, so maybe not use "Wrapper" in this class name?

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdf_namespace.py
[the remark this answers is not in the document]
vepadulano: There is no notion of `std::any` in the implementation of `PyObjRefCountWrapper`, so I'm unsure on linking the two names. On the other hand, without `std::any` the class doesn't make sense, so...

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
vepadulano: > Also, maybe consider a more general name like SetSharedResource? I

This function is specifically meant for setting the value of the data member of the RLoopManager. Before making it more generic, I would like to see how we would use the same pattern for another class (e.g. the RVecDS). I would say that's for another PR

guitargeek (approved):
Thank you for considering my comments! For me it's good now

bindings/pyroot/pythonizations/src/PyROOTModule.cxx
pcanal: The equivalent for a source file is to use the anonymous namespace.  Note that both solution (anonymous namespace or Internal namespace) had the advantage of hiding (mangling) the `PyObjRefCounter` symbols from the linker.

guitargeek: Yes I have thought about that as well with the linker, it's a good point. But it's already in the `PyROOT` namespace, which I double will collide with anything (`PyROOT::Internal` is not reducing the probability that much further)

PR #17747, opened 2025-02-18
Implement `TBrowser::Draw()` method

linev:
Now dummy canvas is drawn. 
Instead create browser implementation again if it was deleted before

Correctly reset `TBrowser::fImp` pointer when 'normal' TRootBrowser is closed and deleted.
Up to now TBrowser was using no-longer-valid pointer

Small fix in `RBrowser::Show()` method - which is invoked by `TBrowser::Draw()` now


bellenot (approved):
LGTM, thanks!

PR #17749, opened 2025-02-18
[ntuple] Allow for alternative column types in `RNTupleDS`

### LINKED ISSUE #16034: [ntuple] Unable to use `std::vector` in RDF
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Reading vector fields in an RNTuple through RDF is currently only possible when they are specified as `ROOT::RVec`. This breaks compatibility with TTree-based RDF event loops, where both `std::vector` and `ROOT::RVec` can be used for branches that hold collections.

### Reproducer

```cpp
using ROOT::Experimental::RNTupleModel;
using ROOT::Experimental::RNTupleWriter;

const std::string kTreeName = "T";
const std::string kTreeFileName = "ttree_rdf_stdvector.root";

const std::string kNTupleName = "N";
const std::string kNTupleFileName = "ntuple_rdf_stdvector.root";

void write_tree() {
  auto file = std::unique_ptr<TFile>(TFile::Open(kTreeFileName.c_str(), "RECREATE"));
  auto tree = std::make_unique<TTree>(kTreeName.c_str(), kTreeName.c_str());

  std::vector<int> brVector;
  tree->Branch("V", &brVector);

  brVector = {1, 2, 3};
  tree->Fill();
  tree->Write();
}

void write_ntuple() {
  auto fldVector = model->MakeField<std::vector<int>>("V");

  auto ntuple =
      RNTupleWriter::Recreate(std::move(model), kNTupleName, kNTupleFileName);

  *fldVector = {1, 2, 3};
  ntuple->Fill();
}

void read(std::string_view name, std::string_view path) {
  ROOT::RDataFrame df(name, path);
  auto dfRes = df.Define("V_size", [](std::vector<int> vs) { return vs.size(); }, {"V"})
                 .Display();
  dfRes->Print();
}

void rdf_stdvector() {
  write_tree();
  read(kTreeName, kTreeFileName);

  write_ntuple();
  read(kNTupleName, kNTupleFileName);
}

```

Running this as a ROOT macro will show that it works for TTree, but for RNTuple it results in a segmentation fault. Changing the type signature of the `Define` lambda to `ROOT::RVec<int>` will run and produce the expected output for both TTree and RNTuple.

### ROOT version

master, any

### Installation method

any

### Operating system

any

### Additional context

_No response_

enirolf:
This PR adds the possibility of using different (valid) types for RDF columns in the `RNTupleDS`. A notable example is the interchangeable use of `std::vector<T>` and `ROOT::RVec<T>`. We enable this by introducing *alternative prototype fields*, which are created on-demand when a column reader is requested for a type other than the original prototype field that was added when the datasource was constructed. 

An error is thrown in `RNTupleColumnReader` in case the alternative column type is incompatible with the on-disk type of the corresponding field.

Fixes #16034 and #10749.

### DIFF
--- a/tree/dataframe/src/RNTupleDS.cxx
+++ b/tree/dataframe/src/RNTupleDS.cxx
@@ -194,7 +194,15 @@ public:
          }
       }
 
-      ROOT::Experimental::Internal::CallConnectPageSourceOnField(*fField, source);
+      try {
+         ROOT::Experimental::Internal::CallConnectPageSourceOnField(*fField, source);
+      } catch (const ROOT::RException &err) {
+         auto onDiskType = source.GetSharedDescriptorGuard()->GetFieldDescriptor(fField->GetOnDiskId()).GetTypeName();
+         std::string msg = "RNTupleDS: invalid type \"" + fField->GetTypeName() + "\" for column \"" +
+                           fDataSource->fFieldId2QualifiedName[fField->GetOnDiskId()] + "\" with on-disk type \"" +
+                           onDiskType + "\"";
+         throw std::runtime_error(msg);
+      }
 
       if (fValuePtr) {
          // When the reader reconnects to a new file, the fValuePtr is already set
--- a/tree/dataframe/src/RNTupleDS.cxx
+++ b/tree/dataframe/src/RNTupleDS.cxx
@@ -16,6 +16,7 @@
 
 #include <ROOT/RDF/RColumnReaderBase.hxx>
 #include <ROOT/RField.hxx>
+#include <ROOT/RFieldUtils.hxx>
 #include <ROOT/RPageStorageFile.hxx>
 #include <ROOT/RNTupleDescriptor.hxx>
 #include <ROOT/RNTupleDS.hxx>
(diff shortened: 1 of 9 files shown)


tree/dataframe/test/datasource_ntuple.cxx
```diff
@@ -267,6 +267,67 @@ TEST_F(RNTupleDSTest, ChainMT)
    ChainTest(fNtplName, fFileName);
 }
 
+TEST_F(RNTupleDSTest, AlternativeColumnTypes)
+{
+   auto df = ROOT::RDF::Experimental::FromRNTuple(fNtplName, fFileName);
+
+   auto usingDouble = df.Define("nJets", [](ROOT::RVec<double> jets) { return jets.size(); }, {"jets"})
```
vepadulano, line 274:
```suggestion
   auto usingDouble = df.Define("nJets", [](const ROOT::RVec<double> &jets) { return jets.size(); }, {"jets"})
```

I would try to use `const &` everywhere in function arguments passed to RDF API. That being said, we may want to add an extra ad hoc test case for when the user wants to modify the values of an input column. i.e. with a first `Define` call that takes the input value as `ROOT::RVec<double> &`, modfies it, then in a second `Define` or `Filter` perform some computation or selection based on the modified values. I'll let you decide if you want to add it for this PR, but it can also be added separately

tree/dataframe/test/datasource_ntuple.cxx
```diff
@@ -267,6 +267,67 @@ TEST_F(RNTupleDSTest, ChainMT)
...
+
+   auto jitted = df.Define("jetsType", "ROOT::Internal::RDF::TypeID2TypeName(typeid(jets))")
+                    .Take<std::string>("jetsType")
+                    .GetValue();
+   EXPECT_EQ("ROOT::VecOps::RVec<float>", jitted[0]);
+
+   auto nested = df.Define("nNnlo", [](std::vector<std::vector<float>> nnlo) { return nnlo.size(); }, {"nnlo"})
+                    .Take<std::size_t, ROOT::RVec<std::size_t>>("nNnlo")
+                    .GetValue();
+   EXPECT_EQ(3ull, ROOT::VecOps::Sum(nested));
```
vepadulano, line 304: Very good that we're testing this! Can we also add a test for `ROOT::RVec<std::vector<float>>` and the opposite combination and see if those work?

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -193,7 +194,13 @@ class RNTupleColumnReader : public ROOT::Detail::RDF::RColumnReaderBase {
          }
       }
 
-      ROOT::Experimental::Internal::CallConnectPageSourceOnField(*fField, source);
+      try {
+         ROOT::Experimental::Internal::CallConnectPageSourceOnField(*fField, source);
+      } catch (const ROOT::RException &err) {
+         std::string msg = "RNTupleDS: invalid type \"" + fField->GetTypeName() + "\" for column \"" +
+                           fDataSource->fFieldId2QualifiedName[fField->GetOnDiskId()] + "\"";
```
vepadulano, line 201: Does it make sense to also report the on-disk type for the given column?

vepadulano (changes requested):
Thanks a lot for working on this! See some comments inline.

tree/dataframe/test/datasource_ntuple.cxx
enirolf: Noted! I will address this in a separate PR.

vepadulano (approved):
Very nice, thanks!

ferdymercury (commented):
Thanks!!
One question, shouldn't the Tests also include the cases of Float_t vs float and uint32_t vs UInt_t ?
Like done in https://github.com/root-project/root/pull/17221/files#diff-4e81901cc89256ef932186f77d6d7df37f2ddf6b4665c9ccd958159bbca1af57

enirolf:
> Thanks!! One question, shouldn't the Tests also include the cases of Float_t vs float and uint32_t vs UInt_t (being interchangable) ? Like done in https://github.com/root-project/root/pull/17221/files#diff-4e81901cc89256ef932186f77d6d7df37f2ddf6b4665c9ccd958159bbca1af57

Added!

ferdymercury (approved):

PR #17764, opened 2025-02-19
fix broken ninja builds with cppinterop

aaronj0:
No longer force `CppInterOpUnitTests` to build in metacling by adding a dependency to libCling (a bad hack). This target is not built since it is excluded from the build tree. Setting the property to off, triggers the `check-cppinterop` target, so ninja no longer complains about not finding the executables while building dictionaries


silverweed (approved):

aaronj0:
Thanks @silverweed Though the test failures can be attributed to files that can't be retrieved from eos, I'd prefer to rerun the CI once it is up and then merge.

dpiparo:
thanks for this fix.

PR #12745, opened 2023-05-01
Simplify CallFunc. NFC

vgvassilev:
As part of our efforts in the area of cppyy and libInterOp we need to port CallFunc. These parts help code reuse and I'd prefer to keep in sync the to the two versions.

PR #17767, opened 2025-02-19
Simplify TClingCallFunc code based on InterOp's JitCall

aaronj0:
These improvements are based on https://github.com/vgvassilev/root/tree/simplify-cf-tors and latest changes in JitCall

### DIFF
--- a/core/metacling/src/TClingCallFunc.cxx
+++ b/core/metacling/src/TClingCallFunc.cxx
@@ -196,9 +196,7 @@ void TClingCallFunc::collect_type_info(QualType &QT, ostringstream &typedefbuf, 
          QT.print(OS, Policy, type_name);
          OS.flush();
       }
-      for (int i = 0; i < indent_level; ++i) {
-         typedefbuf << kIndentString;
-      }
+      indent(typedefbuf, indent_level);
       typedefbuf << "typedef " << fp_typedef_name << ";\n";
       return;
    } else if (QT->isMemberPointerType()) {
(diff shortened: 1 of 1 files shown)


hahnjo:
If the changes are based on https://github.com/root-project/root/pull/12745, I honestly don't understand why we need a new PR. As far as I can see, that PR was accepted in 2023 with one minor comment but never pushed. What is the reason for not addressing that and getting it merged? Just copying over changes feels like bad practice.

aaronj0:
> If the changes are based on https://github.com/root-project/root/pull/12745, I honestly don't understand why we need a new PR. As far as I can see, that PR was accepted in 2023 with one minor comment but never pushed. What is the reason for not addressing that and getting it merged? Just copying over changes feels like bad practice.

The implementation of the structor code preparation is based on InterOp https://github.com/root-project/root/blob/bfeb0a10add47ff5ba09dfc9ea723aa8d0665106/interpreter/CppInterOp/lib/Interpreter/CppInterOp.cpp#L2536 and has some minor changes from the one from 2023, and since I could not rebase that PR to test, I chose to raise a new one. I can close this one if you prefer

vgvassilev:
Is that PR a subset or a superset of #12745?

vgvassilev:
> > If the changes are based on #12745, I honestly don't understand why we need a new PR. As far as I can see, that PR was accepted in 2023 with one minor comment but never pushed. What is the reason for not addressing that and getting it merged? Just copying over changes feels like bad practice.
> 
> The implementation of the structor code preparation is based on latest InterOp
> 

We need to be careful when merging. Both codes departed from each other and both had important fixes...

aaronj0:
> Is that PR a subset or a superset of #12745?

It's a superset, also handles typedefs to anonymous types, and further sinks code duplication. I am pushing to the earlier PR and will close this one.

vgvassilev (approved):
LGTM!

PR #17770, opened 2025-02-19
[gui] Update the browser(s) when opening a file

bellenot:


gui/gui/src/TRootCanvas.cxx
```diff
@@ -840,6 +841,17 @@ Bool_t TRootCanvas::ProcessMessage(Longptr_t msg, Longptr_t parm1, Longptr_t)
                         if (!fi.fFilename) return kTRUE;
                         dir = fi.fIniDir;
                         new TFile(fi.fFilename, "update");
```
linev, line 843: Probably here one should use `TFile::Open()` - it is more robust

linev (approved):

gui/gui/src/TRootCanvas.cxx
bellenot: done

linev: Thanks!

PR #17773, opened 2025-02-19
[net] Improve RRawDavixFile, do not check the SSL certificate when reading https with Davix, and move http read in a tutorial to EOS

dpiparo:
This PR accomplishes three things, all necessary to streamline the way in which files are accessed or not through http(s) in ROOT's test suite.
1) The `ROOT::Internal::RRawDavixFile` class now honours the `Davix.GSI.CACheck` parameter
2) The default value of the `Davix.GSI.CACheck` parameter is now `n` instead of `y`, therewith we do not check the peer certificate any more when using Davix bacause the functionality is broken for example on macOS (see https://github.com/cern-fts/davix/issues/135)
3) Some heavy reading that happened from a web server during tests is now happening through xrootd on EOS Public at CERN


tutorials/analysis/mp_processSelector.C
```diff
@@ -19,11 +19,11 @@
 
 #include <memory>
 
-const auto file0 = "http://root.cern/files/h1/dstarmb.root";
+const auto file0 = "root://eospublic.cern.ch//eos/root-eos/h1/dstarmb.root";
 const std::vector<std::string> files = {file0,
-                                        "http://root.cern/files/h1/dstarp1a.root",
-                                        "http://root.cern/files/h1/dstarp1b.root",
-                                        "http://root.cern/files/h1/dstarp2.root"};
+                                        "root://eospublic.cern.ch//eos/root-eos/h1/dstarp1a.root",
```
silverweed, line 24: Are the double slashes after `cern.ch` intentional?

dpiparo: yes! It's how we instruct our plugin manager to load the right library.

net/davix/src/utils.h
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright (C) 1995-2025, Rene Brun and Fons Rademakers.               *
+ * All rights reserved.                                                  *
+ *                                                                       *
+ * For the licensing terms see $ROOTSYS/LICENSE.                         *
+ * For the list of contributors see $ROOTSYS/README/CREDITS.             *
+ *************************************************************************/
+
+#include <string>
+
+inline bool isno(const char *str)
```
silverweed, line 11: Any specific reason for moving this into its own file?

silverweed (approved):

tutorials/legacy/multicore/mp104_processH1.C
```diff
@@ -30,10 +30,10 @@ static std::string tutname = "mp104_processH1: ";
 static std::string logfile = "mp104_processH1.log";
 static RedirectHandle_t gRH;
 
-std::vector<std::string> files {"http://root.cern/files/h1/dstarmb.root",
-                                "http://root.cern/files/h1/dstarp1a.root",
-                                "http://root.cern/files/h1/dstarp1b.root",
-                                "http://root.cern/files/h1/dstarp2.root"};
+std::vector<std::string> files {"root://eospublic.cern.ch//eos/root-eos/h1/dstarmb.root",
```
pcanal, line 33: Note that until the xrootd over eos issue with vector reads are solved, this is a disaster for not-at-cern cases.

Issue #10454, opened 2022-04-26
[cling] Memory hogging when checking if type is an enum


etejedor:
Initially reported here:

https://root-forum.cern.ch/t/pyroot-and-std-vector-dramatic-ram-usage/49711

when a user was getting a `vector<int>` branch from a tree in a loop in PyROOT.

After a bit of digging, it seems the culprit is a call that is made from cppyy to `TInterpreter::ClassInfo_IsEnum`. The following is a C++ reproducer (tested with current master):

```python
float getMem() {
        auto info = ProcInfo_t();
        gSystem->GetProcInfo(&info);
        float mem = (float)info.fMemResident;
        return mem*1e-3;
}

void test()
{
    for (int i=0; i < 100000; ++i) {
        gInterpreter->ClassInfo_IsEnum("vector<int>");
        if (i%1000 == 0)
            printf("RSS at iteration %d is %f\n", i, getMem());
    }
}
```

When running the macro above, it can be seen how the memory usage increases after each iteration.

eguiraud:
Related (or maybe the exact same issue): https://github.com/root-project/root/issues/9029

pcanal:
I don't know if it is an option at that place in the code but calling a combination of 
```TClass::GetClass(name_of_type);```
and
```TEnum::GetEnum(name_of_type);```
should be more memory efficient.

**Comment by etejedor** [comment]
Dear Alberto,

I'm afraid this is not yet fixed, but @Axel-Naumann can confirm.

**Comment by Axel-Naumann** [comment]
@etejedor https://github.com/root-project/root/issues/10454#issuecomment-1110193565 suggests a change in PyROOT. Can you confirm whether that helps?

**Comment by etejedor** [comment]
So iiuc the suggestion is to replace [this line](https://github.com/root-project/root/blob/master/bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx#L949) by something like `return TEnum::GetEnum(tn_short.c_str());`? Where should `TClass::GetClass` be used here and for what purpose?

**Comment by Axel-Naumann** [comment]
Something along the lines of
```
    if (TClass::GetClass(tn_short.c_str()) return false;
    return gInterpreter->ClassInfo_IsEnum(tn_short.c_str());
```
maybe?

etejedor:
[...]

PR #17775, opened 2025-02-20
Fix `IsEnum` check in cppyy/metacling

aaronj0:
Fixes https://github.com/root-project/root/issues/10454, and the following reproducer no longer seg faults:

```python
import ROOT

ROOT.gInterpreter.Declare('''
class UponCopyPrinter {
public:
   ~UponCopyPrinter() = default;          
   UponCopyPrinter() = default;      
   UponCopyPrinter(const UponCopyPrinter &) { std::cout << "Invoking copy c'tor!" << std::endl; }
   UponCopyPrinter(UponCopyPrinter &&) = default;
   UponCopyPrinter &operator=(const UponCopyPrinter &) = default;
   UponCopyPrinter &operator=(UponCopyPrinter &&) = default;
};
''')

RVec_UponCopyPrinter = ROOT.ROOT.VecOps.RVec(ROOT.UponCopyPrinter)

v = RVec_UponCopyPrinter(3)
v2 = RVec_UponCopyPrinter(v.data(), v.size())
print("%s and %s" %(v.data(), v2.data()))

v2.resize(4)
print("%s and %s" %(v.data(), v2.data()))
```
This PR also enables InterOp API in clingwrapper,  which is the first step towards performing clang-based reflection in cppyy, bypassing the limitations of the type system.

I am currently testing in TCling whether this is ready to use for `ClassInfo_IsEnum`. Ideally we would patch `Cppyy::IsEnum` to fix this issue, which I will test in a following commit after the CI run, hence the changes in clingwrapper's cmake.


hahnjo:
Note that by relying on CppInterOp will not allow to backport to 6.32 and 6.34 releases...

bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx
```diff
@@ -1078,9 +1081,13 @@ bool Cppyy::IsAbstract(TCppType_t klass)
 bool Cppyy::IsEnum(const std::string& type_name)
 {
     if (type_name.empty()) return false;
-    std::string tn_short = TClassEdit::ShortType(type_name.c_str(), 1);
-    if (tn_short.empty()) return false;
-    return gInterpreter->ClassInfo_IsEnum(tn_short.c_str());
```
pcanal, line 1083: Note: this solve the problem 'only' for Python.  It would be even better to fix `TCling::ClassInfo_IsEnum`.

aaronj0: Agreed! The initial fix was in TCling(https://github.com/root-project/root/actions/runs/13424841558), that caused a certain set of test failures. I was curious to know which of them happened only via the python interface leading me to temporarily patch clingwrapper instead

PR #17782, opened 2025-02-20
[ntuple] Decouple `RPageSource` from `RNTupleJoinTable`

enirolf:
This is the first PR (out of a yet unknown number) related to the refactoring and extension of the `RNTupleJoinTable` to enable more complex join table structures which may contain entries from more than one RNTuple. This is necessary to make the `RNTupleJoinProcessor` fully composable (see also #17132).

The changes in this PR are twofold:
1. Join tables have to now be build explicitly through a call to `RNTupleJoinTable::Build`.
2. The `RNTupleJoinTable` doesn't store the page source as a data member anymore. Instead, this page source is passed to `RNTupleJoinTable::Build`.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
@@ -70,10 +70,11 @@ private:
    /// fields) to their respective entry numbers.
    std::unordered_map<RCombinedJoinFieldValue, std::vector<ROOT::NTupleSize_t>, RCombinedJoinFieldValueHash> fJoinTable;
 
-   /// The page source belonging to the RNTuple for which to build the join table.
-   std::unique_ptr<RPageSource> fPageSource;
+   /// Names of the join fields used for the mapping to their respective entry numbers.
+   std::vector<std::string> fJoinFieldNames;
 
-   /// The fields for which the join table is built. Used to compute the hashes for each entry value.
+   /// The fields for which the join table is built, corresponding to `fJoinFieldNames`. Used to compute the hashes for
+   /// each entry value.
    std::vector<std::unique_ptr<RFieldBase>> fJoinFields;
 
    /// Only built join tables can be queried.
@@ -84,8 +85,7 @@ private:
    ///
    /// \param[in] fieldNames The names of the join fields to use for the join table. Only integral-type fields are
    /// allowed.
-   //  \param[in] pageSource The page source.
-   RNTupleJoinTable(const std::vector<std::string> &fieldNames, const RPageSource &pageSource);
+   RNTupleJoinTable(const std::vector<std::string> &fieldNames) : fJoinFieldNames(fieldNames) {}
 
    /////////////////////////////////////////////////////////////////////////////
    /// \brief Ensure the RNTupleJoinTable has been built.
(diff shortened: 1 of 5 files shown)


tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -70,10 +70,11 @@ private:
...
 
-   /// The page source belonging to the RNTuple for which to build the join table.
-   std::unique_ptr<RPageSource> fPageSource;
+   /// Names of the join fields used for the mapping to their respective entry numbers.
+   std::vector<std::string> fJoinFieldNames;
 
-   /// The fields for which the join table is built. Used to compute the hashes for each entry value.
+   /// The fields for which the join table is built, corresponding to `fJoinFieldNames`. Used to compute the hashes for
+   /// each entry value.
    std::vector<std::unique_ptr<RFieldBase>> fJoinFields;
```
hahnjo, line 78: Do we still need the fields as a member variable or can it become local to `Build()`?

enirolf: In `GetAllEntryNumbers`, we need to know the size of the field to be able to properly cast the incoming value pointers. This currently uses `RFieldBase::GetValueSize`. We can of course also only store the sizes themselves.

enirolf:
> Just curious for the future direction: Will `RNTupleJoinTable::BuildIndex` be changed to eventually accept a processor, for full composability?

I would like to have the `RNTupleJoinTable` not depend on the `RNTupleProcessor` because I think that would become a bit messy dependency-wise. I can explain and discuss my current idea in more detail offline, but briefly, what I'm currently thinking about is restructuring the join table such that it can be partitioned according to some key into smaller mappings. These mappings are all coupled to a single page source, but the partition table can contain mappings from multiple page sources. On the `RNTupleProcessor` side, there would then be a method `CreateJoinTable` method (name pending) for each processor that takes care of building the mapping (for `RNTupleSingleProcessor`) and accumulating them into these partitions (for `RNTupleChainProcessor`, with one partition per inner processor and `RNTupleJoinProcessor`, probably just a single partition).

tree/ntuple/v7/src/RNTupleJoinTable.cxx
```diff
@@ -84,28 +56,47 @@ void ROOT::Experimental::Internal::RNTupleJoinTable::Build()
...
+   std::vector<std::unique_ptr<RFieldBase>> fields;
    std::vector<RFieldBase::RValue> fieldValues;
-   fieldValues.reserve(fJoinFields.size());
+   fieldValues.reserve(fJoinFieldNames.size());
+
+   for (const auto &fieldName : fJoinFieldNames) {
+      auto fieldId = desc->FindFieldId(fieldName);
+      if (fieldId == ROOT::kInvalidDescriptorId)
+         throw RException(R__FAIL("could not find join field \"" + std::string(fieldName) + "\" in RNTuple \"" +
+                                  pageSource.GetNTupleName() + "\""));
```
vepadulano, line 71: Add information on the storage location of the RNTuple perhaps? e.g. if this is building from a chain of ntuples the name could be the same for every one

vepadulano (approved):
LGTM, thanks! See a minor comment inline

tree/ntuple/v7/src/RNTupleJoinTable.cxx
enirolf: Unfortunately `RPageSource` does not provide this information, so this is something that will need to be handled by the upstream application (i.e., the `RNTupleProcessor`). `Build` could return a `RResult` instead to propagate the error instead of throwing it directly, but since the `RNTupleJoinTable` as well as its integration into the `RNTupleProcessor` will still change significantly I would propose to address this at a later point.

PR #17783, opened 2025-02-20
[ntuple] Fix RClassField not resolving context-dependent typedefs

### LINKED ISSUE #17774: [ntuple] ATLAS RNTuple Writing Issue (2025-02-19)
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Hi everyone,

We've noticed that RNTuple writing seems to be broken for ATLAS in the most recent `dev3LCG` nightlies. I have a standalone reproducer here:

https://gitlab.cern.ch/amete/rntuple-bug-report-20250219

When running this against the head of master (see the README), I get:

```
  what():  unknown type: Variable_t<int> (unchecked RResult access!) (unchecked RResult access!)
At:
  ROOT::Experimental::RFieldBase::Create(const std::string&, const std::string&, const ROOT::Experimental::RCreateFieldOptions&, const ROOT::Experimental::RNTupleDescriptor*, ROOT::DescriptorId_t)::<lambda(const std::string&, ROOT::Experimental::RInvalidField::RCategory)> [/build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-HEAD/src/ROOT/HEAD/tree/ntuple/v7/src/RFieldBase.cxx:311]
  static ROOT::RResult<std::unique_ptr<ROOT::Experimental::RFieldBase> > ROOT::Experimental::RFieldBase::Create(const std::string&, const std::string&, const ROOT::Experimental::RCreateFieldOptions&, const ROOT::Experimental::RNTupleDescriptor*, ROOT::DescriptorId_t) [/build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-HEAD/src/ROOT/HEAD/tree/ntuple/v7/src/RFieldBase.cxx:610]
  static ROOT::RResult<std::unique_ptr<ROOT::Experimental::RFieldBase> > ROOT::Experimental::RFieldBase::Create(const std::string&, const std::string&) [/build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-HEAD/src/ROOT/HEAD/tree/ntuple/v7/src/RFieldBase.cxx:266]
  static ROOT::RResult<std::unique_ptr<ROOT::Experimental::RFieldBase> > ROOT::Experimental::RFieldBase::Create(const std::string&, const std::string&) [/build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-HEAD/src/ROOT/HEAD/tree/ntuple/v7/src/RFieldBase.cxx:266]
```

whereas, running it against ROOT `v6.34` all seems to be OK.

I suspect this might be a fallout of the most recent changes, e.g., [root/pull/17637](https://github.com/root-project/root/pull/17637), but I haven't explicitly confirmed this.

Can someone take a look at this, please? Many thanks in advance.

Best,
Serhan 

### Reproducer

Provided in the aforementioned Git repo.

### ROOT version

Head of master.

### Installation method

LCG

### Operating system

AlmaLinux9

### Additional context

_No response_

silverweed:
# This Pull request:
With #17637 we started using `TDataMember::GetFullTypeName()` for the type name of RClassField's subfields, relying on later type renormalization. However, this approach fails when the type in question cannot be looked up by TClass::GetClass because it's context dependent (e.g. a templated typedef inside a class).
This PR goes back to use `TDataMember::GetTrueTypeName()` instead, and assigns the FullTypeName to the type alias of the created subfield. This allows handling cases of context-dependent typedefs, as we use the already-resolved type when creating the subfield.

## Checklist:

- [x] tested changes locally

This PR fixes #17774


tree/ntuple/v7/test/ntuple_types.cxx
```diff
@@ -2233,3 +2229,30 @@ TEST(RNTuple, RColumnRepresentations)
...
+         auto ptr = std::make_unique<DerivedWithTypedef>();
+         ptr->m.push_back(i);
+         entry->BindRawPtr("foo", ptr.get());
+         ntuple->Fill(*entry);
+      }
+   }
+
+   {
+      auto reader = RNTupleReader::Open("ntpl", fileGuard.GetPath());
+      EXPECT_EQ(reader->GetNEntries(), 10);
```
pcanal, line 2256: Should we also check that the types are spelt the way we expects?

pcanal (approved):
LGTM.

tree/ntuple/v7/test/ntuple_types.cxx
```diff
@@ -2233,3 +2229,30 @@ TEST(RNTuple, RColumnRepresentations)
...
+      auto fieldBase = RFieldBase::Create("foo", "DerivedWithTypedef").Unwrap();
+      model->AddField(std::move(fieldBase));
+      auto ntuple = RNTupleWriter::Recreate(std::move(model), "ntpl", fileGuard.GetPath());
+      for (auto i = 0; i < 10; ++i) {
+         auto entry = ntuple->GetModel().CreateBareEntry();
+         auto ptr = std::make_unique<DerivedWithTypedef>();
+         ptr->m.push_back(i);
+         entry->BindRawPtr("foo", ptr.get());
+         ntuple->Fill(*entry);
+      }
```
pcanal, line 2251: Random question.  Is this code also valid (and lead to the same output)?:
```suggestion
      auto entry = ntuple->GetModel().CreateBareEntry();
      auto ptr = std::make_unique<DerivedWithTypedef>();
      entry->BindRawPtr("foo", ptr.get());
      for (auto i = 0; i < 10; ++i) {
         ptr->m.clear();
         ptr->m.push_back(i);
         ntuple->Fill(*entry);
      }
```

hahnjo: Yes, entries can be reused and `BindRawPtr` called infinitely often. In the end it's just a container / staging area for one call to `Fill()`.

tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -161,6 +164,8 @@ ROOT::Experimental::RClassField::RClassField(std::string_view fieldName, TClass
       std::unique_ptr<RFieldBase> subField;
 
       subField = RFieldBase::Create(dataMember->GetName(), typeName).Unwrap();
+      if (subField->fTypeAlias.empty() && typeName != subField->GetTypeName())
+         subField->fTypeAlias = typeName;
```
hahnjo, line 168: I think we have to ignore the set `subField->fTypeAlias` entirely: It is based on the assumption that the user specified `dataMember->GetTrueTypeName()` which is not the case. Also I think we need to call `Internal::GetNormalizedUnresolvedTypeName` before comparing, something like:
```suggestion
       // RFieldBase::Create() set subField->fTypeAlias based on the assumption that the user specified typeName, which already went through one round of type resolution.
       std::string origTypeName{dataMember->GetFullTypeName()};
       // TODO: or move up, together with the other loop
       // For C-style arrays, complete the type name with the size for each dimension, e.g. `int[4][2]`
      if (dataMember->Property() & kIsArray) {
         for (int dim = 0, n = dataMember->GetArrayDim(); dim < n; ++dim)
            origTypeName += "[" + std::to_string(dataMember->GetMaxIndex(dim)) + "]";
      }
      const auto normTypeName = Internal::GetNormalizedUnresolvedTypeName(origTypeName);
      if (normTypeName == subField->GetTypeName())
         subField->fTypeAlias = "";
      } else {
         subField->fTypeAlias = normTypeName;
     }
```

hahnjo (commented):
I agree with Philippe, we should verify that the fields (including class member fields) have the expected type names and aliases. Maybe this should go to `tree/ntuple/v7/test/ntuple_type_name.cxx` though and we keep `ntuple_types.cxx` about being able to write and read the data.

tree/ntuple/v7/test/ntuple_types.cxx
```diff
@@ -2027,6 +2027,8 @@ TEST(RNTuple, TClassStlDerived)
       RNTupleWriteOptions options;
       auto ntuple = RNTupleWriter::Recreate(std::move(model), "f", fileGuard.GetPath(), options);
       for (int i = 0; i < 10000; i++) {
+         if (fieldKlass) 
+            fieldKlass->~PackedContainer<int>();
```
hahnjo, line 2031: Ok, if you fix this here, please do it in a separate commit

hahnjo (approved):
LGTM in principle. However now we lost the test case that we can actually reconstruct the model and read the data. I think we should have this in `ntuple_types.cxx`

silverweed:
@hahnjo 
> However now we lost the test case that we can actually reconstruct the model and read the data. I think we should have this in ntuple_types.cxx

What do you mean exactly? The updated test that I moved in `ntuple_type_name.cxx` is a superset of the previous one, so I'm not sure what are we missing..

hahnjo:
> What do you mean exactly? The updated test that I moved in `ntuple_type_name.cxx` is a superset of the previous one, so I'm not sure what are we missing..

Ok, maybe I missed that in the previous version of the PR: I think we should test the model reconstruction and actual reading of entries.

jblomer (approved):
Thanks!

hahnjo (approved):
LGTM, thanks! You may want to squash the last two commits, the first still says WIP

PR #17784, opened 2025-02-20
[ntuple] use raw strings in ntuple_show for ease of reading

silverweed:
Aesthetic-only change: use raw strings instead of string concatenation for the expected values in ntuple_show.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/test/ntuple_show.cxx
+++ b/tree/ntuple/test/ntuple_show.cxx
@@ -84,44 +84,46 @@ TEST(RNTupleShow, BasicTypes)
    std::ostringstream os;
    ntuple2->Show(0, os);
    // clang-format off
-   std::string fString{ std::string("")
-      + "{\n"
-      + "  \"pt\": 5,\n"
-      + "  \"db\": 9.99,\n"
-      + "  \"byte\": 0x89,\n"
-      + "  \"int\": -4,\n"
-      + "  \"uint\": 3,\n"
-      + "  \"uint64\": 44444444444,\n"
-      + "  \"string\": \"TestString\",\n"
-      + "  \"boolean\": true,\n"
-      + "  \"uint8\": 97,\n"
-      + "  \"int8\": 97,\n"
-      + "  \"bitset\": \"10000000000000000000000000000010000000000000000000000000000010010\",\n"
-      + "  \"pstring\": \"abc\",\n"
-      + "  \"atomic\": false\n"
-      + "}\n" };
+   std::string fString{
+R"({
+  "pt": 5,
+  "db": 9.99,
+  "byte": 0x89,
+  "int": -4,
+  "uint": 3,
+  "uint64": 44444444444,
+  "string": "TestString",
+  "boolean": true,
+  "uint8": 97,
+  "int8": 97,
+  "bitset": "10000000000000000000000000000010000000000000000000000000000010010",
+  "pstring": "abc",
+  "atomic": false
+}
+)" };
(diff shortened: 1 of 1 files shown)


tree/ntuple/v7/test/ntuple_show.cxx
```diff
@@ -374,26 +383,27 @@ TEST(RNTupleShow, Collections)
...
-      + "* N-Tuple : Collections                                                        *\n"
-      + "* Entries : 1                                                                  *\n"
-      + "********************************************************************************\n"
-      + "* Field 1           : myCollection                                             *\n"
-      + "*   Field 1.1       : _0                                                       *\n"
-      + "*     Field 1.1.1   : myShort (std::int16_t)                                   *\n"
-      + "*     Field 1.1.2   : myFloat (float)                                          *\n"
-      + "********************************************************************************\n" };
+   std::string outputFields{
+      "************************************ NTUPLE ************************************\n"
```
pcanal: This portion is matching better the surrounding indentation than in the case of the raw strings usage (edited for clarity).

tree/ntuple/v7/test/ntuple_show.cxx
```diff
@@ -84,44 +84,46 @@ TEST(RNTupleShow, BasicTypes)
...
-      + "  \"string\": \"TestString\",\n"
-      + "  \"boolean\": true,\n"
-      + "  \"uint8\": 97,\n"
-      + "  \"int8\": 97,\n"
-      + "  \"bitset\": \"10000000000000000000000000000010000000000000000000000000000010010\",\n"
-      + "  \"pstring\": \"abc\",\n"
-      + "  \"atomic\": false\n"
-      + "}\n" };
+   std::string fString{
+R"({
```
pcanal, line 88: This PR both this syntax and compile time string concatenation (https://github.com/root-project/root/pull/17784#discussion_r1963744923).

The main disadvantage (not fatal) is that this section is not indented like the surrounding code (whereas the other syntax is).

hahnjo (commented):
I don't really mind either way, but I think there are again formatting-only changes in the commit that make it hard to determine which lines just have whitespace-changes and which are actually modified...

tree/ntuple/v7/test/ntuple_show.cxx
silverweed: True, but I'd argue the advantage in having a "heredoc-style" string where you see the expected output 1:1 outweighs the misindentation

tree/ntuple/v7/test/ntuple_show.cxx
silverweed: Not sure if I interpret this comment correctly: you mean you like this more than the raw strings? In this case I could do without a raw string because there are no tons of `"` to escape, so it works; in all other cases I personally find the raw string much more readable.
However, as it's a subjective choice, the only way to decide is by majority vote I think.

silverweed:
> there are again formatting-only changes in the commit that make it hard to determine which lines just have whitespace-changes and which are actually modified...

I think it's safe to say that nothing was modified in such a way that its meaning changed, otherwise the related test would have failed.

tree/ntuple/v7/test/ntuple_show.cxx
pcanal: At the time, I must have missed the escaped quotes in the original ....

pcanal (approved):
LGETM

PR #17786, opened 2025-02-20
[RF] Compatibility fix with Clad master in getting gradient function

guitargeek:
Clad `master` provides different overloads for the gradient, and we need to resolve to the one that we want. Without the `static_cast` introduced in this commig, getting the function pointer to the gradient function would be ambiguous.


vgvassilev:
That probably means that we can bump to clad v1.9 as the windows issue of v1.8 seems gone…

vgvassilev (approved):
Lgtm!

PR #17787, opened 2025-02-20
[ntuple] Skip columns without pages in RNTupleInspector

hahnjo:


tree/ntupleutil/v7/test/ntuple_inspector.cxx
```diff
@@ -622,6 +624,10 @@ TEST(RNTupleInspector, PageSizeDistribution)
    // Requesting a histogram for a column type not present in the given RNTuple should give an empty histogram
    auto nonExistingTypeHisto = inspector->GetPageSizeDistribution(ENTupleColumnType::kReal32);
    EXPECT_EQ(0, nonExistingTypeHisto->Integral());
+
+   // Requesting a histogram for a column type without columns in the given RNTuple should give an empty histogram
```
silverweed, line 628: You mean `for a column type without pages`, right?

enirolf (changes requested):
Thanks for this catch! LGTM, but the same check will have to be added before line 369 for completeness.

enirolf (approved):
Thanks! Perhaps squash the two commits before merging.

hahnjo:
> Thanks! Perhaps squash the two commits before merging.

It's two individual fixes though, no?

silverweed (approved):

enirolf:
> > Thanks! Perhaps squash the two commits before merging.
> 
> It's two individual fixes though, no?

Ah wait sorry, yes you're right.

Issue #11603, opened 2022-10-18
Disable automatic 'call home' in cmake when not needed


wdconinc:
### Is your feature request related to a problem? Please describe.
1. We would like to compile ROOT without triggering a 'call home' in the CMakeLists.txt, at https://github.com/root-project/root/blob/master/CMakeLists.txt#L124, so we can install ROOT on network-isolated nodes without incurring artificial build delays, or on network-connected nodes which we would prefer to remain private.
2. When desired, we would like  the ROOT internet connection check to work regardless of captive portals and other web screening portals, since currently those sites may appear as connected because the content of the downloaded file is not checked for accuracy.
3. We would like to be notified in advance of the privacy policy that governs the data that is (inevitably) collected by the root.cern webservers without the user's consent or knowledge, since there exist user privacy expectations here. [GDPR principles](https://gdpr.eu/what-is-gdpr/) that may apply: lack of transparency, collection beyond the limited purpose, lack of data minimization, and probably some others (I am not a laywer).

### Describe the solution you'd like
The 'call home' in https://github.com/root-project/root/blob/master/CMakeLists.txt#L124 should default to off, and should only be called when explicitly requested by a user, and only when the features (`builtin_gsl` and `clad`) that depend on it are enabled. When it is used, the internet connection check should check a checksum of the downloaded file.

### Describe alternatives you've considered
Alternatively, the user can be alerted upon downloading the ROOT source code from any of its various locations that compiling this software may trigger data collection on the root.cern server, e.g. https://root.cern/install/build_from_source/, https://github.com/root-project/root/releases, etc.

bellenot:
So we could introduce such option (e.g `check_connectivity`), set it ON by default (to keep the current behavior), and set the `NO_CONNECTION` to `TRUE` if it is disabled. Would that fit your need?
And maybe @Axel-Naumann can also comment on this

guitargeek:
Hi @wdconinc and @bellenot! It would be nice if we could fix the issue without the added complexity of an additional flag.

Wouter, you're probably using the `fail-on-missing` option to build ROOT, right? The flag to make sure that the features don't get quietly disabled at configuration time.

The connectivity check doesn't make sense with `fail-on-missing=ON`. More on this in the description of the PR that I opened:
  * https://github.com/root-project/root/pull/15467

Would that PR fix your issue?

wdconinc:
The fix seems fine (we are indeed using `fail-on-missing` by default through the spack package).

[...]

PR #17791, opened 2025-02-21
[interop] Fix build errors with `-Dtesting=off`

aaronj0:
This was due to a bug in CppInterOp where `option(CPPINTEROP_ENABLE_TESTING "Enables the testing infrastructure." BOOL)`
was defined twice conditionally. This is incorrect and does not allow users to override the option, leading to testing always being on. Fixed with https://github.com/compiler-research/CppInterOp/pull/506

This also requires ROOT to set the same flag in the interpreter CMake (which did not work earlier due to the underlying bug in InterOp)

cc @bellenot


interpreter/CppInterOp/CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ set(CMAKE_CXX_EXTENSIONS NO)
 include(GNUInstallDirs)
   option(CPPINTEROP_USE_CLING "Use Cling as backend" OFF)
   option(CPPINTEROP_USE_REPL "Use clang-repl as backend" ON)
+  option(CPPINTEROP_ENABLE_TESTING "Enable the CppInterOp testing infrastructure." ON)
```
bellenot, line 62: Wrong place, you set `set(CPPINTEROP_ENABLE_TESTING OFF)` in `interpreter/CMakeLists.txt` but add the option here... Please see my diff again

interpreter/CMakeLists.txt
```diff
@@ -525,4 +525,9 @@ set(CPPINTEROP_INCLUDE_DIRS
 #---Set InterOp to build on Cling, this can be toggled to use Clang-REPL-------------------------------------------------------
 set(CPPINTEROP_USE_CLING ON BOOL "Use Cling as backend")
 
+#---Disable testing on InterOp if ROOT's testing is OFF (InterOp tests are enabled by default) -------------------------------------------------------
+if(NOT testing)
+set(CPPINTEROP_ENABLE_TESTING OFF)
```
bellenot, line 530:
```suggestion
  set(CPPINTEROP_ENABLE_TESTING OFF)
```

bellenot (dismissed):

interpreter/CppInterOp/CMakeLists.txt
aaronj0: These options are meant to live inside CppInterOp. See `CPPINTEROP_USE_CLING`. We configure them from within the Interpreter CMake that propagates the flags to InterOp.

aaronj0: We cannot define the option in ROOT, since InterOp is also a standalone library.

bellenot: Well, anyway, you cannot set a value before defining it...

interpreter/CMakeLists.txt
vgvassilev: Don’t we just need set in the cache with force?

aaronj0:
> Why do we need this PR I thought everything was already in CppInterOp?

Yes the fix from https://github.com/compiler-research/CppInterOp/pull/506 is pulled in by using the later commit, but  the testing flags are not correctly propagated and CppInterOp does not configure or check the flags when it is not built standalone. For eg: currently with ROOT you can set both `USE_CLING` and `USE_REPL` ON. This is fixed by the patch [here](https://github.com/root-project/root/pull/17791/commits/fb11b367e81304bdf43e25774b1ab2ebc63286e8) which will be opened upstream. The changes in ROOT's interpreter cmake is to set the flags in a better way.

guitargeek (approved):
LGTM! Thanks for the PR, and it also fixed the usecase of building ROOT with `-Dfail-on-missin=ON -Dtesting=OFF -Dbuiltin_gtest=OFF` without internet connection, which was broken by the initial interop PR. Building ROOT without internet connection is important for some experiments need to build ROOT on network-isolated nodes (see also #11603). I also tested that it works again with this PR.

@aaronj0, please merge when you're ready.

aaronj0:
The failures look unrelated to this PR, merging

PR #17792, opened 2025-02-21
Allow untyped reading of TTree values

vepadulano:


tree/dataframe/inc/ROOT/RDF/RTreeColumnReader.hxx
```diff
@@ -107,10 +110,17 @@ class R__CLING_PTRCHECK(off) RTreeColumnReader<RVec<T>> final : public ROOT::Det
...
          if (readerArraySize > 0) {
-            RVec<T> rvec(readerArray.begin(), readerArray.end());
-            swap(fRVec, rvec);
+            // Array is not contiguous, make a full copy of it.
+            fRVec = RVec<Value_t>();
+            fRVec.reserve(readerArraySize * sizeof(T));
+            for (std::size_t i{0}; i < readerArraySize; i++)
+            {
+               auto val = readerArray.At(i);
+               std::copy(val, val + sizeof(T), std::back_inserter(fRVec));
```
vepadulano, line 119: I would like to remove the need for `sizeof(T)` using `TTreeReaderUntypedArray::GetValueSize()`
```suggestion
               std::copy(val, val + arrayReader.GetValueSize(), std::back_inserter(fRVec));
```

But unfortunately it is currently not working as expected. For example, in case the TTreeReaderUntypedArray is reading from a `float` data member of the class which is the element type of the branch type (e.g. `std::vector<MyClass>`), then the `GetValueSize()` function will return `sizeof(MyClass)` instead of `sizeof(float)`, directly resulting from `collectionProxy->SizeOf()` for the reader.

PR #17794, opened 2025-02-21
[ntuple] Move RPage/RPagePool/RPageAllocator out of Experimental

silverweed:
Moving the following classes from `Experimental::Internal` to `Internal`:

* `RPage`
* `RPagePool`
* `RPageAllocator`


tree/ntuple/v7/inc/ROOT/RPageStorage.hxx
```diff
@@ -45,13 +45,17 @@
...
+
 namespace Experimental {
 
 class RNTupleModel;
 
 namespace Internal {
 class RColumn;
-class RNTupleCompressor;
 struct RNTupleModelChangeset;
 class RPageAllocator;
```
hahnjo, line 60: This should be removed

hahnjo (commented):
Not sure if this was "ready for review" yet, but this seems to be a self-contained move. Maybe update the PR summary to reflect the latest changes.

silverweed:
@hahnjo I'm splitting this PR into various sub-PRs, so in the end I'll either close this or rebase it to become significantly smaller

hahnjo (approved):
LGTM, should have a second pair of eyes

tree/ntuple/v7/inc/ROOT/RPageStorage.hxx
```diff
@@ -782,11 +786,11 @@ public:
    REntryRange GetEntryRange() const { return fEntryRange; }
 
    /// Allocates and fills a page that contains the index-th element. The default implementation searches
-   /// the page and calls LoadPageImpl(). Returns a default-constructed RPage for suppressed columns.
-   virtual RPageRef LoadPage(ColumnHandle_t columnHandle, ROOT::NTupleSize_t globalIndex);
+   /// the page and calls LoadPageImpl(). Returns a default-constructed ROOT::Internal::RPage for suppressed columns.
```
enirolf, line 789: FYI, since we don't have any other instances of classes named `RPage`, as far as doxygen is concerned adding the namespace here isn't necessary (it will create the correct link automatically). On the other hand, writing out the full namespace is maybe also not a bad thing. I don't really have an opinion on either, just wanted to point out!

enirolf (approved):
LGTM! One minor comment but I'm fine with merging things as is.

PR #17795, opened 2025-02-21
[tmva] Use local input files instead of downloading them from root.cern.ch for tmva tests and tutorials

lmoneta:
Use as input to TMVA tutorials and tests local files uploaded in github in the tutorials/machine_learning/data directory instead of getting them from root.cern.ch


dpiparo (approved):

dpiparo:
it seems we have some Python in C++ and that does not compile
```c++
TString infileName = ROOT.gROOT.GetTutorialDir() ....
```

PR #17802, opened 2025-02-24
[ntuple] Extend RNTupleFileWriter for external writing

hahnjo:
Both interfaces are useful for external clients that want to write blobs independently and only use the `RNTupleFileWriter` for metadata.


tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
```diff
@@ -112,6 +112,10 @@ public:
    /// The key length of a blob. It is always a big key (version > 1000) with class name RBlob.
    static constexpr std::size_t kBlobKeyLen = 42;
 
+   /// Prepares buffer for a new record as an RBlob key at offset.
+   /// (Note that the array type is purely documentation, the argument is actually just a pointer.)
+   static void PrepareBlobKey(std::int64_t offset, size_t nbytes, size_t len, unsigned char buffer[kBlobKeyLen]);
```
enirolf, line 117: This method should be declared further down, after the private methods.

tree/ntuple/v7/src/RMiniFile.cxx
```diff
@@ -1191,6 +1190,16 @@ ROOT::Experimental::Internal::RNTupleFileWriter::Append(std::string_view ntupleN
...
 }
 
+void ROOT::Experimental::Internal::RNTupleFileWriter::Seek(std::uint64_t offset)
+{
+   if (!fFileSimple)
+      throw RException(R__FAIL("invalid attempt to seek non-simple writer"));
+
+   fFileSimple.fFilePos = offset;
+   fFileSimple.fKeyOffset = offset;
+   // The next Write() will Flush() if necessary.
```
enirolf, line 1200: Perhaps add to the method docstring as well.

enirolf (changes requested):
LGTM in principle, just one comment to address before merging and a suggestion I will leave to you judgement.

tree/ntuple/v7/src/RMiniFile.cxx
```diff
@@ -1190,6 +1190,16 @@ ROOT::Experimental::Internal::RNTupleFileWriter::Append(std::string_view ntupleN
    return writer;
 }
 
+void ROOT::Experimental::Internal::RNTupleFileWriter::Seek(std::uint64_t offset)
+{
+   if (!fFileSimple)
+      throw RException(R__FAIL("invalid attempt to seek non-simple writer"));
+
+   fFileSimple.fFilePos = offset;
```
silverweed, line 1198: - should we do bounds checking on the offset?
- should we return a RResult<void> instead of throwing?

tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
hahnjo: Hm, not sure: `PrepareBlobKey` is some sort of `static` utility method, while the other two `public` functions `Recreate` and `Append` are factory methods that actually create a `RNTupleFileWriter`. I think I would prefer to keep the declaration close to that of `kBlobKeyLen`.

tree/ntuple/v7/src/RMiniFile.cxx
hahnjo: I'm not sure, which bounds should we compare `offset` to? The idea is that external clients can use the method to skip over potentially a big amount of data, which the `RNTupleFileWriter` doesn't know about.

Regarding `RResult`, my understanding was that we use it for error conditions because of user input. That's not really the case here, it's a logic error in the program to call `Seek` on a non-simple writer. A good analogy would be passing a non-matching entry to `RNTupleWriter::Fill()` or `RNTupleReader::LoadEntry` - in both cases we also throw an exception.

silverweed (approved):
LGTM

enirolf (approved):

PR #17803, opened 2025-02-24
[jsroot] dev 24/02/2025

linev:
- Implement 'pol' draw option for `TH2`
- Let edit histograms and graphs title via context menu
- Support `TF12` drawing - projection of `TF2`
- Fix - properly handle image sizes in svg2pdf
- [Internals] use private members in JSROOT classes like `#draw_object`
- [Internals] use negative indexes in array and String

### DIFF
--- a/js/changes.md
+++ b/js/changes.md
@@ -1,19 +1,24 @@
 # JSROOT changelog
 
 ## Changes in dev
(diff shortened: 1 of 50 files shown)


couet (approved):
lgtm

PR #17807, opened 2025-02-24
[ci] Point Davix to the curl CA bundle on macOS

dpiparo:
in order for Davix to be able to read through https


bellenot (approved):
Fine with me if it works

Issue #17809, opened 2025-02-24
TGraph2D doc (TGraph2D::GetFunction does not work)


akira-okumura:
### Explain what you would like to see improved and how.

https://root.cern/doc/v612/classTGraph2D.html
In the "Associated functions" section, it says `*myfunc = gr->GetFunction("myfunc");` but `FindObject` is correct.

### ROOT version

6.12

### Installation method

Doc issue

### Operating system

all

ferdymercury:
Hello, could you clarify what you mean?

`gr->GetFunction` is implemented in such a way that it calls `gr->GetListOfFunctions->FindObject`.

Have you found a case where this is not working? You propose to instead use `gROOT->FindObject` when you are using an external function that is not added to the list of graph functions while fitting?

ferdymercury:
Oh, I see what you mean, it works for TGraph but not for TGraph2D

ferdymercury:
I proposed a fix here: https://github.com/root-project/root/pull/17812

akira-okumura:
Thank you very much. I first thought that `TGraph2D::GetFunction` was not implemented for some technical reason. But if the new implementation does not have any side effect, the consistency between TGraph and TGraph2D is better than changing the doc.

PR #17812, opened 2025-02-24
[hist] Add TGraph2D::GetFunction as with TGraph

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/17809

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/hist/hist/src/TGraph2D.cxx
+++ b/hist/hist/src/TGraph2D.cxx
@@ -790,8 +790,7 @@ void TGraph2D::ExecuteEvent(Int_t event, Int_t px, Int_t py)
 
 TObject *TGraph2D::FindObject(const char *name) const
 {
-   if (fFunctions) return fFunctions->FindObject(name);
-   return nullptr;
+   return fFunctions ? fFunctions->FindObject(name) : nullptr;
 }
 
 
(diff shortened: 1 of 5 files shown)


hist/hist/src/TGraph2D.cxx
```diff
@@ -986,6 +986,18 @@ void TGraph2D::CreateInterpolator(Bool_t oldInterp)
...
+////////////////////////////////////////////////////////////////////////////////
+/// Return pointer to function with name.
+///
+/// Functions such as TGraph2D::Fit store the fitted function in the list of
+/// functions of this graph.
+
+TF2 *TGraph2D::GetFunction(const char *name) const
+{
+   if (!fFunctions) return nullptr;
+   return (TF2*)fFunctions->FindObject(name);
```
hageboeck, line 998:
```suggestion
   return dynamic_cast<TF2*>(FindObject(name));
```

hageboeck: I would do the safe thing here and use a dynamic_cast to `TF2`.

Also, we don't need to duplicate the find logic, since there is already `FindObject`. What do you think?

hageboeck (commented):
Hi @ferdymercury,

I would shorten the code and move to dynamic_cast, what do you think?

ferdymercury:
> I would shorten the code and move to dynamic_cast, what do you think?

Good idea, thanks for the review!
I have applied the same idea in TGraph.cxx so that they look equal.

hageboeck (approved):
LGTM!

hageboeck:
@ferdymercury OK if I squash the commits when merging?

ferdymercury:
> @ferdymercury OK if I squash the commits when merging?

Of course, sounds perfect. I always avoid squashing locally to avoid rerunning the CI again, but I can do it locally if you think it's better.

PR #17813, opened 2025-02-24
[TCling] Use `clang::Module::HeaderKind` to avoid hardcoded index

devajithvs:
Use HK_* enums instead. Will reduce the diff when upgrading to LLVM20

Ref:
https://github.com/llvm/llvm-project/commit/61946687bc68ccba763571cb420049b9a3749dfe

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes #


pcanal (approved):
LGTM.  Thanks.

PR #17814, opened 2025-02-24
meta: Disable TClass creation during tear down.

pcanal:
Since code assume that the creation will succeed rather than returning `nullptr`, we create a trivial/empty `TClass` instead.


core/meta/src/TGenericClassInfo.cxx
```diff
@@ -247,6 +247,13 @@ namespace Internal {
       if (!gROOT)
          ::Fatal("TClass::TClass", "ROOT system not initialized");
 
+      if (gROOT->TestBit(kInvalidObject)) {
+         if (gDebug > 1)
+           ::Warning("TClass::TClass", "ROOT system is in the process of being tear down");
```
hageboeck, line 252:
```suggestion
           ::Warning("TClass::TClass", "ROOT system is being teared down");
```

hageboeck (approved):
LGTM

PR #17815, opened 2025-02-24
Use appropriate constructor in test

vepadulano:
Small followup to https://github.com/root-project/root/pull/17739

Use the RLoopManager constructor without a data source in the simple tests that just check for the validity of the constructed RLoopManager.


guitargeek (approved):
LGTM!

PR #17816, opened 2025-02-25
[RF] Avoid using `std::pow` in interpolation code

guitargeek:
The interpolation code for HistFactory excessively used `std::pow` since the code was written ages ago (see here in [ROOT 5](https://github.com/root-project/root/blob/v5-34-00-patches/roofit/histfactory/src/FlexibleInterpVar.cxx#L329)).

Replacing it with simple multiplications and in one place trading it for the faster `std::exp` results in significant speedups: roughly 30 % faster in the ATLAS Higgs combination benchmark (even 50 % faster in the case of AD, further increasing the benefit you get from using AD in that benchmark. Also the legacy evaluation backend is affected by this improvement thanks to the shared code in `MathFuncs.h`, so any HistFactory user will benefit from this change.

### DIFF
--- a/roofit/roofitcore/inc/RooFit/Detail/MathFuncs.h
+++ b/roofit/roofitcore/inc/RooFit/Detail/MathFuncs.h
@@ -256,7 +256,7 @@ inline double flexibleInterpSingle(unsigned int code, double low, double high, d
       } else if (paramVal < -1) {
          return -1 * (2 * a - b) * (paramVal + 1) + low - nominal;
       } else {
-         return a * std::pow(paramVal, 2) + b * paramVal + c;
+         return a * paramVal * paramVal + b * paramVal + c;
       }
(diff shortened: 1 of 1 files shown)


vgvassilev (approved):
LGTM!

lmoneta (commented):
LGTM, 
Thank you Jonas for the improvement. 
It is clear if we are taking already the log(base), makes sense using std::exp instead of std::pow, otherwise I thing std::pow(a,b) is implemented as std::exp(log(a),b).
Also for the std::pow of integers exponent, you can profit from re-using the values. 
Maybe you can also cache `x0Sq*x0Sq` and be slightly faster?

guitargeek:
>  Maybe you can also cache x0Sq*x0Sq and be slightly faster?

That's a good idea, I will do this in a next update.

PR #17817, opened 2025-02-25
[Minuit2] Cache transformed parameter values in MnHesse

guitargeek:
It was figured out with `perf` and `flamegraph.pl` that the main
performance bottleneck when using Minuits Hesse on RooFit likelihoods is
the transformation to internal Minuit parameters in MnHesse, which is a
relatively expensive trigonometric operation. It is done for all
parameters at every function operation, even if only one parameter is
changed. We need to do order n-squared function calls for the Hessian.
The total runtime of calling the RooFit function itself scales only
linearly with the number of parameters, thanks to the caching in RooFit.
However, the parameter transformation in MnHesse implements no caching
and therefore has quadratic cost.

This PR implements caching for the transformed parameters to make
this bottleneck go away completely.

With the changes in this PR, the plan-of-work item of "Speedup the
computation of the Hessian for big Higgs combinations at least by factor
of 2" is completed. Our ATLAS benchmark is now doing the Hesse step in
100 s instead of 120 s. But as the addressed bottleneck grows with the
number of fit parameters squared, this optimization will have a much
stronger impact on some reported user workflows, where computing the
Hessian takes hours right now. In any case, considering also the
performance improvements in other PRs in this development cycle, one
gets a 2 x speedup in our benchmark too (see #17816).

This change in Minuit indirectly affects all RooFit users. I saw that
this parameter transformation is also the main bottleneck in evaluating
Hessians with likelihoods from CMS combine.

### With ROOT master

Total runtime of `minimize()` and `hesse()`: 160 s.

atlas_old

### With this PR and #17816

Total runtime of `minimize()` and `hesse()`: 105 s (34 % faster).

atlas_new

The new bottlenecks are again in RooFit:

  * dirty flag propagation with `RooAbsArg::setValueDirty()` (about 10 s runtime, we can get rid of it easily because the new CPU evaluation backend doesn't use the dirty flag information anyway)
  * Other overhead in `RooFit::Evaluator::run()` that is not related to actual computation, again 10 more seconds. I don't know what to do about it.

In particular, the `setValueDirty()` is responsible for most of the runtime in the line search. If we get rid of it, the line search will bottleneck fits with AD much less, where the gradient step is very fast and the line search is the bottleneck of the overall minimization.

### DIFF
--- a/math/minuit2/src/MnHesse.cxx
+++ b/math/minuit2/src/MnHesse.cxx
@@ -27,6 +27,19 @@ namespace ROOT {
 
 namespace Minuit2 {
 
+namespace {
+
+/// Internal function to compute the Hessian using numerical derivative
+/// computation.
+MinimumState ComputeNumerical(const MnFcn &, const MinimumState &, const MnUserTransformation &, unsigned int maxcalls,
+                              MnStrategy const &strat);
+
+/// Internal function to compute the Hessian using an analytical computation or
+/// externally provided in the FCNBase class.
+MinimumState ComputeAnalytical(const FCNBase &, const MinimumState &, const MnUserTransformation &);
+
+} // namespace
+
 MnUserParameterState
 MnHesse::operator()(const FCNBase &fcn, const MnUserParameterState &state, unsigned int maxcalls) const
 {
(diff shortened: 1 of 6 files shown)


lmoneta (commented):
Thank Jonas for the PR. It is great you have found this and achieved this speedup!
Can you maybe split the PR in 2, one fixing the MnHesse, the other applying other changes in the code (renaming files, etc....)? 
The second one can then applied later after we merged the pending PR in Minuit2

guitargeek:
Hi @lmoneta, of course you're right. I was abusing this PR as a development branch to get CI feedback, I'll remove all the commits unrelated to the functional change in MnHesse and create a different PR from that.

math/minuit2/src/MnHesse.cxx
```diff
@@ -147,17 +164,65 @@ MinimumState MnHesse::ComputeAnalytical(const FCNBase & fcn, const MinimumState
...
+            _vpar[transform.ExtOfInt(i)] = transform.Int2ext(i, v(i));
+            _lastInput[i] = v(i);
+         }
+      }
+
+      return static_cast<MnUserFcn const &>(_mfcn).callWithTransformedParams(_vpar);
+   }
+
+private:
+   MnFcn const &_mfcn;
```
lmoneta, line 206: To be consistent, in Minuit we use the ROOT coding convention, data members should start with f + capital letters instead of "_". 
Ex:
_mfcn; -> fMfcn

lmoneta (approved):
LGTM!
Thank you Jonas for this optimisation. I have just a small comment on the naming of the data members for the new class caching the internal parameter values

math/minuit2/src/MnHesse.cxx
guitargeek: Ah yes, I was thinking about RooFit. No problem, I'll update that

PR #17821, opened 2025-02-25
[ntuple] Simplify `RNTupleJoinTable` probe interface

enirolf:
With this change, `RNTupleJoinTable` returns all matching entries in all cases, and the responsibility of deciding which entry to use is more explicitly put on the downstream application. This is fine, because these decisions need to be made downstream anyways. The templated interface is removed as well, as its practical use is currently not clear enough.

The `RNTupleJoinTable` itself is in the process of a redesign, where it will store the mappings of join values to entry indexes in partitions. With this change, it won't be possible to directly return the pointer to the list of entry indexes stored in the join table anymore. The changes in this commit already anticipate this redesign by changing the return type of `GetEntryIndexes`.

### DIFF
--- a/tree/ntuple/v7/src/RNTupleJoinTable.cxx
+++ b/tree/ntuple/v7/src/RNTupleJoinTable.cxx
@@ -104,17 +104,8 @@ void ROOT::Experimental::Internal::RNTupleJoinTable::Build(RPageSource &pageSour
    fIsBuilt = true;
 }
 
-ROOT::NTupleSize_t
-ROOT::Experimental::Internal::RNTupleJoinTable::GetFirstEntryNumber(const std::vector<void *> &valuePtrs) const
-{
-   const auto entryIndices = GetAllEntryNumbers(valuePtrs);
-   if (!entryIndices)
-      return ROOT::kInvalidNTupleIndex;
-   return entryIndices->front();
-}
-
-const std::vector<ROOT::NTupleSize_t> *
-ROOT::Experimental::Internal::RNTupleJoinTable::GetAllEntryNumbers(const std::vector<void *> &valuePtrs) const
+std::vector<ROOT::NTupleSize_t>
+ROOT::Experimental::Internal::RNTupleJoinTable::GetEntryIndexes(const std::vector<void *> &valuePtrs) const
 {
    EnsureBuilt();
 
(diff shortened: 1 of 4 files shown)


tree/ntuple/v7/src/RNTupleJoinTable.cxx
```diff
@@ -104,17 +104,8 @@ void ROOT::Experimental::Internal::RNTupleJoinTable::Build(RPageSource &pageSour
...
-   const auto entryIndices = GetAllEntryNumbers(valuePtrs);
-   if (!entryIndices)
-      return ROOT::kInvalidNTupleIndex;
-   return entryIndices->front();
-}
-
-const std::vector<ROOT::NTupleSize_t> *
-ROOT::Experimental::Internal::RNTupleJoinTable::GetAllEntryNumbers(const std::vector<void *> &valuePtrs) const
+std::vector<ROOT::NTupleSize_t>
+ROOT::Experimental::Internal::RNTupleJoinTable::GetEntryIndexes(const std::vector<void *> &valuePtrs) const
```
silverweed, line 108: I'm not getting it from the commit message: why are we now returning a copy of the array rather than a pointer?

enirolf: Ah right, I will add it to the commit message. It's anticipatory right now for some larger changes to the join table I am working on. See [this comment](https://github.com/root-project/root/pull/17782#issuecomment-2673956270) for the gist of it. With these partitions, it may be the case that join values are present in more than one partition, and in this case their corresponding entry indexes will need to be accumulated somewhere, so it won't be possible to directly return a pointer anymore.

I can postpone changing the signature to the point where this is actually needed if you prefer, but I figured that since we're refactoring this method anyways, might as well do it here anyways.

tree/ntuple/v7/test/ntuple_join_table.cxx
```diff
@@ -21,7 +21,7 @@ TEST(RNTupleJoinTable, Basic)
    EXPECT_FALSE(joinTable->IsBuilt());
 
    try {
-      joinTable->GetFirstEntryNumber<std::uint64_t>(0);
+      joinTable->GetEntryIndexes({0});
```
hahnjo, line 24: That's now a pointer, right? Should it be
```suggestion
      std::uint64_t value = 0;
      joinTable->GetEntryIndexes({&value});
```
or explicitly
```suggestion
      joinTable->GetEntryIndexes({nullptr});
```

tree/ntuple/v7/test/ntuple_join_table.cxx
```diff
@@ -136,12 +136,13 @@ TEST(RNTupleJoinTable, SparseSecondary)
...
       if (i % 2 == 1) {
-         EXPECT_EQ(joinTable->GetFirstEntryNumber<std::uint64_t>(event), ROOT::kInvalidNTupleIndex)
+         EXPECT_EQ(joinTable->GetEntryIndexes({&event}), std::vector<ROOT::NTupleSize_t>{})
             << "entry should not be present in the join table";
       } else {
-         auto idx = joinTable->GetFirstEntryNumber<std::uint64_t>(event);
-         EXPECT_EQ(idx, i / 2);
-         EXPECT_FLOAT_EQ(fldX(idx), static_cast<float>(idx) / 3.14);
+         auto entryIdxs = joinTable->GetEntryIndexes({&event});
+         EXPECT_EQ(1ul, entryIdxs.size());
```
hahnjo, line 143: You may want to use `ASSERT_EQ` here (and elsewhere) to immediately fail the test instead of accessing out-of-bounds below

hahnjo (approved):
LGTM, mostly minor comments about the test changes. Should have a second approval before merging.

silverweed (approved):

Issue #17820, opened 2025-02-25
Wrong interaction between indexed TTree friend, GetEntries, GetListOfFriends


vepadulano:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

In a scenario with TTree friends for which an index was built, if the `TTree::GetEntries` function is called, then the list of friends for the main tree is not properly populated. This in turn leads to bad interactions with other tools. For example, the list of all branches of the dataset cannot be properly created since the branches of the friend trees will simply be absent due to the list of friends being unavailable.

### Reproducer

```cpp
#include <TChain.h>
#include <TFile.h>
#include <TTree.h>

#include <filesystem>
#include <iostream>
#include <memory>

void write_data(std::string_view treename, std::string_view filename)
{
    TFile f{filename.data(), "update"};
    TTree t{treename.data(), treename.data()};
    int runNumber{};
    int eventNumber{};
    float val{};
    t.Branch("runNumber", &runNumber, "runNumber/I");
    t.Branch("eventNumber", &eventNumber, "eventNumber/I");
    t.Branch("val", &val, "val/F");
    if (treename == "main")
    {
        for (auto rn = 0; rn < 3; rn++)
        {
            runNumber = rn;
            for (auto en = 0; en < 5; en++)
            {
                eventNumber = en;
                val = en * rn;
                t.Fill();
            }
        }
    }
    else
    {
        for (auto rn = 0; rn < 3; rn++)
        {
            runNumber = rn;
            for (auto en = 4; en >= 0; en--)
            {
                eventNumber = en;
                val = en * rn;
                t.Fill();
            }
        }
    }

    f.Write();
}

void simple()
{
    if (!std::filesystem::exists("test_file.root"))
    {
        write_data("main", "test_file.root");
        write_data("friend", "test_file.root");
    }

    auto main = std::make_unique<TChain>("main");
    main->AddFile("test_file.root");
    auto frChain = std::make_unique<TChain>("friend");
    frChain->AddFile("test_file.root");
    frChain->BuildIndex("runNumber", "eventNumber");
    main->AddFriend(frChain.get());

    // Calling GetEntries() will make the following GetListOfFriends call
    // return a nullptr
    main->GetEntries();

    // If we don't call GetEntries we still need to call LoadTree for the first
    // entry so that the chain populates the fTree data member and the GetTree
    // method can return the correct pointer. In this case, the list of friends
    // is connected.
    // main->LoadTree(0);

[...]

PR #17822, opened 2025-02-25
[tree] Fix interaction between GetEntries and GetListOfFriends

vepadulano:
Fixes https://github.com/root-project/root/issues/17820

Full analysis of the issue follows:

1. TChain::GetEntries calls TChain::LoadTree(TTree::kMaxEntries - 1)

2. Internally, this will set fReadEntry == -2, due to how LoadTree works and the fact it's trying to read beyond the last entry in the chain

3. This has an effect here https://github.com/root-project/root/blob/f33985dca2d6bc505e83498f7b4561ddd969be17/tree/tree/src/TChain.cxx#L1675-L1679 where `t` is the current friend being traversed in the list of friends of the *chain*. We call `t->LoadTree(TTree::kMaxEntries - 1)`.

4. Inside the call to LoadTree on the friend, we reach this line https://github.com/root-project/root/blob/f33985dca2d6bc505e83498f7b4561ddd969be17/tree/tree/src/TChain.cxx#L1484 where the fTree data member of the friend is invalidated.

5. Back to the LoadTree of the main chain, these lines https://github.com/root-project/root/blob/f33985dca2d6bc505e83498f7b4561ddd969be17/tree/tree/src/TChain.cxx#L1681-L1684 would normally connect the fTree data member of the main chain to the current friend, so the list of friends is properly populated. But due to 4., `friend->GetTree()` will return nullptr so the connection won't happen.

### DIFF
--- a/tree/tree/src/TChain.cxx
+++ b/tree/tree/src/TChain.cxx
@@ -985,7 +985,14 @@ Long64_t TChain::GetEntries() const
       return fProofChain->GetEntries();
    }
    if (fEntries == TTree::kMaxEntries) {
-      const_cast<TChain*>(this)->LoadTree(TTree::kMaxEntries-1);
+      const auto readEntry = fReadEntry;
+      auto *thisChain = const_cast<TChain *>(this);
+      thisChain->LoadTree(TTree::kMaxEntries - 1);
+      thisChain->InvalidateCurrentTree();
+      if (readEntry >= 0)
+         thisChain->LoadTree(readEntry);
+      else
+         thisChain->fReadEntry = readEntry;
    }
    return fEntries;
 }
--- a/tree/tree/src/TChain.cxx
+++ b/tree/tree/src/TChain.cxx
@@ -985,6 +985,10 @@ Long64_t TChain::GetEntries() const
       return fProofChain->GetEntries();
    }
    if (fEntries == TTree::kMaxEntries) {
+      // If the following is true, we are within a recursion about friend,
+      // and `LoadTree` will be no-op.
+      if (kLoadTree & fFriendLockStatus)
+         return fEntries;
       const auto readEntry = fReadEntry;
       auto *thisChain = const_cast<TChain *>(this);
       thisChain->LoadTree(TTree::kMaxEntries - 1);
--- a/tree/treeplayer/test/ttreeindex_getlistoffriends.cxx
+++ b/tree/treeplayer/test/ttreeindex_getlistoffriends.cxx
@@ -1,0 +1,114 @@
+#include "TBranch.h"
+#include "TChain.h"
+#include "TCollection.h" // TRangeDynCast
+#include "TFile.h"
+#include "TFriendElement.h"
+#include "TObjArray.h"
+#include "TTree.h"
+
+#include "gtest/gtest.h"
+
+#include <string>
+#include <string_view>
+#include <vector>
+
+void write_data(std::string_view treename, std::string_view filename)
+{
+   TFile f{filename.data(), "update"};
+   TTree t{treename.data(), treename.data()};
+   int runNumber{};
+   int eventNumber{};
+   float val{};
+   t.Branch("runNumber", &runNumber, "runNumber/I");
+   t.Branch("eventNumber", &eventNumber, "eventNumber/I");
+   t.Branch("val", &val, "val/F");
+   if (treename == "main") {
+      for (auto rn = 0; rn < 3; rn++) {
+         runNumber = rn;
+         for (auto en = 0; en < 5; en++) {
+            eventNumber = en;
+            val = en * rn;
+            t.Fill();
+         }
+      }
+   } else {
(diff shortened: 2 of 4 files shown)


vepadulano:
Note that the current change apparently fixed the attached reproducer on my local machine, but evidently something is still missing. Suggestions on how to proceed are welcome!

tree/tree/src/TChain.cxx
```diff
@@ -1373,13 +1375,21 @@ Long64_t TChain::LoadTree(Long64_t entry)
          TIter next(fFriends);
          TFriendLock lock(this, kLoadTree);
          TFriendElement* fe = nullptr;
+         const bool reconstructLocalTreeListOfFriends = fTree->GetListOfFriends() ? false : true;
```
pcanal, line 1378: This test might not be sufficient.  As the internal tree (within the TChain) may have its own (persistent) friends.

tree/tree/src/TChain.cxx
```diff
@@ -985,7 +985,9 @@ Long64_t TChain::GetEntries() const
       return fProofChain->GetEntries();
    }
    if (fEntries == TTree::kMaxEntries) {
+      const auto readEntry = GetReadEntry();
       const_cast<TChain*>(this)->LoadTree(TTree::kMaxEntries-1);
+      const_cast<TChain *>(this)->LoadTree(std::max(readEntry, 0ll));
```
pcanal, line 990:
```suggestion
      if  (readEntry < 0) {
          const_cast<TChain *>(this)->fReadEntry = readEntry;
          const_cast<TChain *>(this)->fTreeNumber = -1
      } else 
         const_cast<TChain *>(this)->LoadTree(std::max(readEntry, 0ll));
```

tree/tree/src/TChain.cxx
vepadulano: If the provided unit test will pass on the CI, maybe we could leave this case for a followup PR? At the moment I don't know about use cases with chain-wide friends and tree-local file-loaded friends at the same time.

tree/treeplayer/test/ttreeindex_getlistoffriends.cxx
```diff
@@ -0,0 +1,86 @@
...
+   friendChain.BuildIndex("runNumber", "eventNumber");
+
+   mainChain.AddFriend(&friendChain);
+
+   // Calling GetEntries used to mess with the fTree data member of the main
+   // chain, not connecting it to the friend chain and thus losing the list
+   // of friends
+   mainChain.GetEntries();
+
+   const auto *listOfFriends = mainChain.GetTree()->GetListOfFriends();
```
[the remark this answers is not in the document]
pcanal: Furthermore, a different/better test?
```suggestion
   EXPECT_EQ( mainChain.GetTree(), nullptr);
   mainChain->GetListOfBranches();   //  This is used in RDF and has the side-effect of calling LoadTree(0);
   EXPECT_NE( mainChain.GetTree(), nullptr);
   const auto *listOfFriends = mainChain.GetTree()->GetListOfFriends();
```
Side note, it might be good thing for `RDF` to **not** rely on `GetListOfBranches` to set the initial cursor and to call either `LoadTree(0)` explicitly or better yet (if the information is available soon enough) `LoadTree(lower_bound_of_entry_range)`

pcanal:
This should also fix https://github.com/root-project/root/issues/17843

tree/treeplayer/test/ttreeindex_getlistoffriends.cxx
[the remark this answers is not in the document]
vepadulano: Done! The improvement on the RDF side is left for another PR.

ferdymercury:
> This should also fix #17843

Thanks. I rebuilt ROOT in Debug mode using master + this patch but I still see the same crash.

pcanal (approved):
LGTM.

ferdymercury (approved):

wdconinc:
FYI, this PR caused a regression in one of our analysis validation scripts when running validation for 6.34.

We were using:
```
    auto tree = new TChain("events"); 
    tree->Add(rec_file.c_str()); 
    TTreeReader tree_reader(tree); // !the tree reader

    //...

    tree_reader.SetEntriesRange(0, tree->GetEntries());
```
This worked before but now segfaults and requires a `LoadTree(0)` to succeed.

https://github.com/eic/physics_benchmarks/pull/82

vepadulano:
Hi @wdconinc ,

Thanks for reaching out! I am sorry that this fix introduced an unexpected side-effect in your code. The changes in this PR are crucial for other analysis tools and they cannot be avoided, but I am sure we can find a way to avoid a segfault in your program. Could you open an issue on this repository with a standalone reproducer or a series of steps to get to one?

Thanks!

wdconinc:
> Could you open an issue on this repository with a standalone reproducer or a series of steps to get to one?

See https://github.com/root-project/root/issues/19814.

PR #16585, opened 2024-10-02
[ntuple] Remove collection writer

jblomer:
Similar functionality can be accessed through the RRecordField (untyped records) and RVectorField::CreateUntyped() (untyped collections).

Fixes #14642

PR #17827, opened 2025-02-25
[ntuple][DF] Update handling of untyped collections in `RNTupleDS`

enirolf:
With the removal of the `RCollectionField` (#16585), all collections can now be treated the same in `RNTupleDS::AddField`. This change removes the addition of a redundant cardinality field, and fixes the column name of untyped collections that contain untyped records where before, it would include the field name of the inner record field in the collection.


vepadulano (approved):
Thank you!

PR #17828, opened 2025-02-25
[skip-ci][ntuple] Fixup `RNTupleProcessor` docs

enirolf:
The PR fixes typos, makes sure the tenses are consistent, adds missing documentation, etc.


hahnjo (approved):
LGTM. I did a quick `grep` and the RNTuple documentation uses a wild mixture of tenses, but now at least the processor uses only one :+1:

enirolf:
> LGTM. I did a quick `grep` and the RNTuple documentation uses a wild mixture of tenses, but now at least the processor uses only one 👍

Yes indeed, it would be nice if we could get it fully consistent but that would require some more effort..

PR #17829, opened 2025-02-25
[ntuple] Remove blanks between angle brackets earlier

hahnjo:
... in `GetCanonicalTypePrefix`: This is required so substring operations work correctly in `RFieldBase::Create`, for example to get the item type of containers. It is hard to observe from the outside since template parameters are further processed in `GetRenormalizedTypeName` and `GetNormalizedUnresolvedTypeName` respectively. As a test, check field emulation of a templated type inside a vector.


enirolf (approved):

silverweed (approved):
LGTM, thanks

PR #17623, opened 2025-02-05
[TFormula] TFormula: Pol functions do not accept variable name as arguments 

Aditya-138-12:
# This Pull request:
Extend TFormula's syntax for polynomials
1. Modified HandlePolN to accept named parameters and variables
2. Modified the HandlePolN to use modern c++ functions like std::getline
3. Added tests for the same in test_TFormula.cxx
## Changes or fixes:
Allow variables in TFormula Pol functions like
`TF1 f1("f1", "pol1(x, 0)");`
`TF1 f1("f1", "pol1(x, [A], [B])");`

## Checklist:

- [✅] tested changes locally
- [❌] updated the docs (if necessary)

This PR fixes #16794

PR #17833, opened 2025-02-26
[hist] Fix invalid int comparison and printf usage in TFormula

silverweed:
#17623 introduced a warning (that is, a compile error in `dev=ON`) due to improper int comparisons.
This PR fixes it. I chose to cast to `Int_t` rather than `size_t` to automatically handle the case where `degree < 0` (which I assume is "impossible" but you never know, given that `degree` is derived from string parsing above...)


hageboeck (approved):
Thanks, LGTM!

PR #17835, opened 2025-02-26
[DF] Add test for modified input values

enirolf:
Adds a test that defines a new column by modifying the values of an existing value in-place (i.e., by passing a non-const ref to `Define`) and subsequently using this new column. To mirror the behaviour of TTree-based dataframes, we expect the original column to be modified iff the modified column is also (indirectly) accessed in the action of the event loop. If only the original column is used in the action, its values shouldn't be modified. Vice-versa, if only the columns defined from this original column are used, their values (of course) should be modified.

Originally suggested by @vepadulano in https://github.com/root-project/root/pull/17749#discussion_r1967328154.

### DIFF
--- a/tree/dataframe/test/datasource_ntuple.cxx
+++ b/tree/dataframe/test/datasource_ntuple.cxx
@@ -310,6 +310,58 @@ TEST_F(RNTupleDSTest, ChainTailScheduling)
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


vepadulano (approved):
Fantastic, thanks!

PR #17837, opened 2025-02-26
[CMake] Fix range splitting logic in ROOT_CREATE_HEADER_COPY_TARGETS

guitargeek:
In CMake, `foreach(.. RANGE start stop)` is inclusive for both start and stop, so we need to decrement the LIST_LENGTH to get the desired logic.

I noticed this logic error when I changed something in ROOT that resulted one package to have *exactly* 100 header files, triggering the following error:

```txt
CMake Error at cmake/modules/RootMacros.cmake:1293 (list):
  list begin index: 100 is out of range 0 - 99
Call Stack (most recent call first):
  CMakeLists.txt:391 (ROOT_CREATE_HEADER_COPY_TARGETS)
```

Follow-up on 4d0f60386ce.


hageboeck (approved):
LGTM, thanks!

PR #17842, opened 2025-02-27
[ntuple] Consistently use RResult to communicate failure in RNTupleSerialize

silverweed:
RNTupleSerialize is currently using an unclear mix of `RResult` and `RException` to communicate failure, without a consistent logic behind it.
Since we decided that `RResult` is the way we want to handle errors due to incorrect user input, this PR changes all exceptions that were thrown in case of wrong input to RResult (this includes invalid values, invalid buffer sizes and such).

Furthermore, all functions that internally call fallible sub-functions now propagate the error to the caller instead of throwing in place. 
This has the benefit of making the error handling uniform and explicitly communicate through the functions' interface their fallibility.

## Checklist:

- [x] tested changes locally


tree/ntuple/v7/inc/ROOT/RNTupleSerialize.hxx
```diff
@@ -223,17 +223,17 @@ public:
...
-   static std::uint32_t SerializeExtraTypeInfoId(ROOT::Experimental::EExtraTypeInfoIds id, void *buffer);
+   static RResult<std::uint32_t> SerializeFieldStructure(ROOT::ENTupleStructure structure, void *buffer);
+   static RResult<std::uint32_t> SerializeColumnType(ROOT::ENTupleColumnType type, void *buffer);
+   static RResult<std::uint32_t> SerializeExtraTypeInfoId(ROOT::Experimental::EExtraTypeInfoIds id, void *buffer);
    static RResult<std::uint32_t> DeserializeFieldStructure(const void *buffer, ROOT::ENTupleStructure &structure);
    static RResult<std::uint32_t> DeserializeColumnType(const void *buffer, ROOT::ENTupleColumnType &type);
    static RResult<std::uint32_t>
    DeserializeExtraTypeInfoId(const void *buffer, ROOT::Experimental::EExtraTypeInfoIds &id);
 
    static std::uint32_t SerializeEnvelopePreamble(std::uint16_t envelopeType, void *buffer);
```
enirolf, line 234: What about this method?

tree/ntuple/v7/inc/ROOT/RNTupleSerialize.hxx
```diff
@@ -243,25 +243,25 @@ public:
 
    static std::uint32_t SerializeRecordFramePreamble(void *buffer);
    static std::uint32_t SerializeListFramePreamble(std::uint32_t nitems, void *buffer);
```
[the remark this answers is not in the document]
silverweed: As of now they cannot fail (in the sense that they don't call any function that return a RResult - only basic serializing routines that don't do any check on the input)

silverweed:
> It also seems to me like the two current commits should be squashed into one (or at least, the second commit contains changes to some return types that should belong in the first one).

I wanted to keep them separate because they're already quite big changes as they are - I made sure that the first commit is self-contained and doesn't depend on the second, so even if they are logically grouped they may stay separate without issues.

tree/ntuple/v7/inc/ROOT/RNTupleSerialize.hxx
silverweed: Same answer as above

tree/ntuple/v7/inc/ROOT/RNTupleSerialize.hxx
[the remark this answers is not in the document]
enirolf: I would be in favor for full consistency in this regard. Especially since this effort is made now anyways, plus there's is a small risk that if ever these methods are changed to have the possibility to fail, the return value is left unchanged and the inconsistency is reintroduced.

Alternatively, it could be more clearly documented if and how each method could fail, but that seems like more effort.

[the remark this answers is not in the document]
silverweed: Thinking about it a bit more, I would rather keep the signatures as-is, for two reasons:
1- I think it's valuable to preserve the real fallibility information in the function's signature, rather than somewhat arbitrarily pretend they may fail; if in the future the implementation of the functions change so that they start have a failing path, the interface should change to reflect that.
> there's is a small risk that if ever these methods are changed to have the possibility to fail, the return value is left unchanged and the inconsistency is reintroduced.

I think we can pretty easily catch this in a code review, and besides even changing the signature wouldn't prevent someone from adding a failure as an exception. If they are careful enough and take a look at how the error handling is made throughout the entire file, they should have no problem doing the right thing regardless of the signature.

2- changing those couple of functions to return RResult has a small cascade effect that requires several other currently unfallible file-local functions to change *their* signatures as well, which adds unneeded noise due to the added error handling and propagation that needs to happen at all stages.

I am open to counter-arguments, but my preferred solution right now would be to go with the current set of signatures.

tree/ntuple/v7/src/RNTupleSerialize.cxx
```diff
@@ -40,8 +40,17 @@
...
 
-std::uint32_t SerializeField(const ROOT::Experimental::RFieldDescriptor &fieldDesc, ROOT::DescriptorId_t onDiskParentId,
-                             ROOT::DescriptorId_t onDiskProjectionSourceId, void *buffer)
+#define GLUE_(a, b) a##b
+#define GLUE(a, b) GLUE_(a, b)
+#define UNWRAP_OR_RETURN(out, x)                    \
+   auto GLUE(res, __LINE__) = (x);                  \
+   if (!GLUE(res, __LINE__))                        \
+      return R__FORWARD_ERROR(GLUE(res, __LINE__)); \
+   out = GLUE(res, __LINE__).Unwrap()
```
hahnjo, line 49: This preprocessor macro isn't very nice. As far as I can tell, it saves only one or maybe two lines per usage. I think I would be in favor of just spelling it out every time, as the existing code already did in many places.

tree/ntuple/v7/src/RNTupleSerialize.cxx
```diff
@@ -53,7 +62,9 @@ std::uint32_t SerializeField(const ROOT::Experimental::RFieldDescriptor &fieldDe
    pos += RNTupleSerializer::SerializeUInt32(fieldDesc.GetFieldVersion(), *where);
    pos += RNTupleSerializer::SerializeUInt32(fieldDesc.GetTypeVersion(), *where);
    pos += RNTupleSerializer::SerializeUInt32(onDiskParentId, *where);
-   pos += RNTupleSerializer::SerializeFieldStructure(fieldDesc.GetStructure(), *where);
+   std::uint32_t result;
+   UNWRAP_OR_RETURN(result, RNTupleSerializer::SerializeFieldStructure(fieldDesc.GetStructure(), *where));
+   pos += result;
```
hahnjo, line 67: For example
```suggestion
   auto result = RNTupleSerializer::SerializeFieldStructure(fieldDesc.GetStructure(), *where)
   if (!result)
      return R__FORWARD_ERROR(result);
   pos += result.Unwrap();
```
is not much longer and IMHO much more readable.

tree/ntuple/v7/inc/ROOT/RNTupleSerialize.hxx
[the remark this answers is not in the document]
enirolf: Alright, I see your arguments. Let's leave it like this for now then :+1:

tree/ntuple/v7/src/RNTupleSerialize.cxx
silverweed: I actually find the macro much "cleaner" since it neatly wraps a lot of repeated line noise and it slightly reduces the chance of copypaste errors. It makes it very clear that in that place there is error forwarding involved and lets the reader split up easily the `if`s that are some more specific logic from those that are just error propagation.

I'd like to hear the other reviewers' opinion as well so we can better come up to an agreement.

tree/ntuple/v7/src/RNTupleSerialize.cxx
silverweed: Could you be a little more specific about what's "not very nice" about it?

tree/ntuple/v7/src/RNTupleSerialize.cxx
```diff
@@ -40,8 +40,17 @@
 namespace {
 using RNTupleSerializer = ROOT::Experimental::Internal::RNTupleSerializer;
 
-std::uint32_t SerializeField(const ROOT::Experimental::RFieldDescriptor &fieldDesc, ROOT::DescriptorId_t onDiskParentId,
-                             ROOT::DescriptorId_t onDiskProjectionSourceId, void *buffer)
+#define GLUE_(a, b) a##b
```
pcanal, line 43: This is a duplicate of existing macros in `RConfig.cxx` (eg. `_R__JOIN_`)

tree/ntuple/v7/src/RNTupleSerialize.cxx
pcanal: In the general sense, macro are 'quite' annoying when it comes to debugging code, in particular stepping through).

tree/ntuple/v7/src/RNTupleSerialize.cxx
pcanal:
```
   if (auto result = RNTupleSerializer::SerializeFieldStructure(fieldDesc.GetStructure(), *where))
       pos += result.Unwrap();
   else
      return R__FORWARD_ERROR(result);
```
might be another option.

tree/ntuple/v7/src/RNTupleSerialize.cxx
silverweed: I suspected we had one but I couldn't find it, thanks

tree/ntuple/v7/src/RNTupleSerialize.cxx
hahnjo: Debugging is one thing, but there are many other things for example documented here: https://gcc.gnu.org/onlinedocs/cpp/Macros.html (especially the pitfalls)

In particular: Local variable definitions are awkward. You are using one approach here which appends the line number, that is by no means a guarantee that there is no clash. Furthermore, the variable "pollutes" the calling site because the whole macro is not wrapped in the usual `do { } while (0)` idiom, to also swallow the semicolon. Then I personally find the usage of the `out` variable not ideal because the type is unclear (yes, the compiler will check it, but the error messages are not great because of the macro) and the operation could either be setting or summing. IMHO all of these go naturally away without the macro, at only one additional line of code per instance.

silverweed: > that is by no means a guarantee that there is no clash

I'm curious to know the circumstances where there may be a clash, considering it's only used in one file.

> the whole macro is not wrapped in the usual do { } while (0) idiom, to also swallow the semicolon

With the latest push declaring the variable in the `if` directly this is no longer the case.

> Then I personally find the usage of the out variable not ideal because the type is unclear

True, it's not great, but unfortunately I can't think of a better way of doing it.

While I recognize the shortcomings of the macro, I still find it a bit better than the alternative, but since we are now 2 votes against 1 I'll respect the majority's opinion (unless someone else votes for the macro within the next hours).

pcanal: > I'm curious to know the circumstances where there may be a clash, considering it's only used in one file.

Not a realistic scenario but this would fail:
```
   UNWRAP_OR_RETURN(out, expression1); UNWRAP_OR_RETURN(out, expression2);
```

silverweed: It is indeed not a realistic scenario.

silverweed: (That said, I will remove the macro because the majority prefers the other version)

tree/ntuple/v7/src/RNTupleSerialize.cxx
```diff
@@ -124,51 +134,62 @@ ROOT::RResult<std::uint32_t> DeserializeField(const void *buffer, std::uint64_t
...
    auto fnFrameSizeLeft = [&]() { return frameSize - (bytes - base); };
-   auto result = RNTupleSerializer::DeserializeFrameHeader(bytes, bufSize, frameSize);
-   if (!result)
-      return R__FORWARD_ERROR(result);
-   bytes += result.Unwrap();
+   if (auto res = RNTupleSerializer::DeserializeFrameHeader(bytes, bufSize, frameSize)) {
+      bytes += res.Unwrap();
+   } else {
+      return R__FORWARD_ERROR(res);
+   }
```
pcanal, line 141: This part (and similar below) seem to be non-functional changes, is that correct?  If so, what is the advantages of the new formulation?

silverweed: The advantage is to reduce the scope of the `res`/`result` variable to only the `if` statement.

silverweed: The actual reason is that there were like 50 usages of `UNWRAP_OR_RETURN` and I didn't check case-by-case. I hardly believe this kind of minor syntactical details are worth the time we are spending on them, as in the end it's all equivalent.

silverweed: Is this a blocker? Or can this conversation be resolved?

hahnjo: No, let's do it uniformly everywhere, resolving from my side.

hahnjo (approved):
LG from my side. Note that there is now a conflict in `tree/ntuple/v7/src/RNTupleSerialize.cxx` that you need to resolve...

PR #17844, opened 2025-02-27
[hist] optimization of polymarker save as c macro

ferdymercury:
# This Pull request:

## Changes or fixes:

avoid calling many times the same function. For 50000 points in tutorials/hist/hist056_TPolyMarker_contour.C, when you save the second canvas as .C, the resulting running time is (old) 30s vs 3s (new)

## Checklist:

- [x] tested changes locally


linev (changes requested):
Because of fixed `pmx` and `pmy` names this will not work for several `TPolyMarker` objects.

linev:
> `Double_t *dum = 0;`

This I removing in my PR.

linev:
In https://github.com/root-project/root/pull/17851 performance problem in `TPolyMarker` will be resolved.

PR #17845, opened 2025-02-27
Introduce `TObject::SavePrimitiveConstructor()` method

linev:
Store invocation of object constructor in the created macro.
Replaces several lines in typical `SavePrimitive()` implementation.

PR also includes `TNamed::SavePrimitiveNameTitle` method as small extension.

And several examples for classes like TGraph, THStack, TPolyMarker how introduced methods can be used.

If approach is fine, one can simplify most of existing `SavePrimitive()` implementations.
At the same time one can revise code there.

### DIFF
--- a/core/base/inc/TNamed.h
+++ b/core/base/inc/TNamed.h
@@ -32,6 +32,8 @@ protected:
    TString   fName;            //object identifier
    TString   fTitle;           //object title
 
+   void SavePrimitiveNameTitle(std::ostream &out, const char *variable_name);
+
 public:
    TNamed(): fName(), fTitle() { } // NOLINT: not allowed to use = default because of TObject::kIsOnHeap detection, see ROOT-10300
    TNamed(const char *name, const char *title) : fName(name), fTitle(title) { }
--- a/core/base/inc/TObject.h
+++ b/core/base/inc/TObject.h
@@ -53,7 +53,7 @@ protected:
    void MakeZombie() { fBits |= kZombie; }
    virtual void DoError(int level, const char *location, const char *fmt, va_list va) const;
 
-   void SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char *variable_name, const char *constructor_agrs = "", Bool_t empty_line = kTRUE);
+   static void SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char *variable_name, const char *constructor_agrs = "", Bool_t empty_line = kTRUE);
 
 public:
    //----- Global bits (can be set for any object and should not be reused).
--- a/core/base/inc/TObject.h
+++ b/core/base/inc/TObject.h
@@ -53,6 +53,8 @@ protected:
    void MakeZombie() { fBits |= kZombie; }
    virtual void DoError(int level, const char *location, const char *fmt, va_list va) const;
 
+   void SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char *variable_name, const char *constructor_agrs = "", Bool_t empty_line = kTRUE);
+
 public:
    //----- Global bits (can be set for any object and should not be reused).
    //----- Bits 0 - 13 are reserved as global bits. Bits 14 - 23 can be used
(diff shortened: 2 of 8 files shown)


core/base/inc/TNamed.h
```diff
@@ -32,6 +32,8 @@ class TNamed : public TObject {
    TString   fName;            //object identifier
    TString   fTitle;           //object title
 
+   void SavePrimitiveNameTitle(std::ostream &out, const char *variable_name);
```
pcanal, line 35:
```suggestion
   static void SavePrimitiveNameTitle(std::ostream &out, const char *variable_name);
```

hist/hist/src/THStack.cxx
```diff
@@ -1071,12 +1063,12 @@ void THStack::SavePrimitive(std::ostream &out, Option_t *option /*= ""*/)
          TString hname = h->GetName();
```
pcanal, line 1063: Doesn't this need to be transformed into a valid variable name?

core/base/inc/TNamed.h
linev: static does not work here

hist/hist/src/THStack.cxx
linev: I not yet touch this code.

Maybe one need other method to create unique and valid identifier for such cases

core/base/inc/TNamed.h
pcanal: why not?

linev: Then one need extra argument for `this` pointer

hist/hist/src/THStack.cxx
pcanal: We do have `ROOT::TMetaUtils::GetCppName` that could/should be used for all `Name to variablename` transformation.  (This change could indeed be in separate PR).

core/base/inc/TNamed.h
pcanal: My bad.  I commented on the wrong routine. :(

core/base/inc/TObject.h
```diff
@@ -53,6 +53,8 @@ class TObject {
    void MakeZombie() { fBits |= kZombie; }
    virtual void DoError(int level, const char *location, const char *fmt, va_list va) const;
 
+   void SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char *variable_name, const char *constructor_agrs = "", Bool_t empty_line = kTRUE);
```
pcanal, line 56:
```suggestion
   static void SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char *variable_name, const char *constructor_agrs = "", Bool_t empty_line = kTRUE);
```

hist/hist/src/THStack.cxx
linev: it should be not only valid name, but also unique name. 
I prefer to change it like for other classes.

hist/hist/src/TGraph.cxx
```diff
@@ -2146,26 +2146,17 @@ void TGraph::SaveAs(const char *filename, Option_t *option) const
...
    frameNumber++;
 
-   TString fXName, fYName;
-
+   TString args;
    if (fNpoints >= 1) {
-      fXName = SaveArray(out, "fx", frameNumber, fX);
-      fYName = SaveArray(out, "fy", frameNumber, fY);
+      auto xname = SaveArray(out, "fx", frameNumber, fX);
+      auto yname = SaveArray(out, "fy", frameNumber, fY);
```
pcanal, line 2155:
```suggestion
      TString xname = SaveArray(out, "fx", frameNumber, fX);
      TString yname = SaveArray(out, "fy", frameNumber, fY);
```
The next line require `xname` and `yname` to be `TString`, it is probably better to fail compilation here if `SaveArray` changes return type.

pcanal (approved):
LGTM.

PR #17846, opened 2025-02-28
[core] AddBefore/After/At option variant, as with Add/First/Last

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-9288 (the first part)

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/core/cont/inc/TList.h
+++ b/core/cont/inc/TList.h
@@ -85,10 +85,15 @@ public:
    void              AddLast(TObject *obj) override;
    virtual void      AddLast(TObject *obj, Option_t *opt);
    void              AddAt(TObject *obj, Int_t idx) override;
+   void              AddAt(TObject *obj, Int_t idx, Option_t *option);
    void              AddAfter(const TObject *after, TObject *obj) override;
    virtual void      AddAfter(TObjLink *after, TObject *obj);
+   void              AddAfter(const TObject *after, TObject *obj, Option_t *option);
+   void              AddAfter(TObjLink *after, TObject *obj, Option_t *option);
    void              AddBefore(const TObject *before, TObject *obj) override;
    virtual void      AddBefore(TObjLink *before, TObject *obj);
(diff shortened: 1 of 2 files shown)


silverweed:
Do we really want to add support for a class that users are not supposed to ever use directly anymore? Is there a compelling current use case that requires this?

ferdymercury:
> Do we really want to add support for a class that users are not supposed to ever use directly anymore? Is there a compelling current use case that requires this?

Maybe it's better to ask the reporter of the JIRA ticket, who requested it, if he really needs it.
I am happy either way, if not, in that case https://its.cern.ch/jira/browse/ROOT-9288 can be closed as won't fix.

silverweed:
Let's wait for @dpiparo and @pcanal's opinion on this.
The jira ticket is almost 7 years old at this point, so I doubt it's still relevant (but if we manage to contact the user we can hear what they have to say)

dpiparo:
Even if this is a legacy class, since this nice work has been done, I would be in favour of merging and finally address the JIRA item.

dpiparo:
@pcanal ?

core/cont/inc/TList.h
```diff
@@ -85,10 +85,15 @@ friend  class TListIter;
    void              AddLast(TObject *obj) override;
    virtual void      AddLast(TObject *obj, Option_t *opt);
    void              AddAt(TObject *obj, Int_t idx) override;
+   void              AddAt(TObject *obj, Int_t idx, Option_t *option);
```
pcanal, line 88: A challenge is that those routines are non-virtual while their counterpart without option are virtual.  This virtuality is used by `THashList` and possibly (maybe) by other user classes.  As-is, if one call this signature on a `THashList`, the state of the container will be invalid/incorrect.

dpiparo:
Are you suggesting to close as "won't fix" because the cost-benefit ratio of the full solution would not be too favourable?

PR #17851, opened 2025-02-28
Introduce TObject::SavePrimitiveArray

linev:
1. Let store array of doubles as plain C array in output macro
2. Returns unique variable name which used then in other operation
3. Use in `TPolyMarker` to speedup creation of such objects, for 50K markers speedup from 37s to 0.5s
4. Use method in `TGraph` classes, removing `TGraph::SaveArray` 
5. Improve `SavePrimitive` methods of `TGraph` - based classes

### DIFF
--- a/core/base/src/TObject.cxx
+++ b/core/base/src/TObject.cxx
@@ -797,7 +797,7 @@ TString TObject::SavePrimitiveArray(std::ostream &out, const char *prefix, Int_t
       for (Int_t i = 0; i < len-1; i++) {
          out << arr[i] << ",";
          if (i && (i % 16 == 0))
-            out << std::endl << "   ";
+            out << "\n   ";
          else
             out << " ";
       }
--- a/core/base/src/TObject.cxx
+++ b/core/base/src/TObject.cxx
@@ -46,6 +46,7 @@ class hierarchies (watch out for overlaps).
 #include <sstream>
 #include <fstream>
 #include <iostream>
+#include <iomanip>
 
 #include "Varargs.h"
 #include "snprintf.h"
@@ -777,6 +778,36 @@ void TObject::SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char
       out << cl->GetName() << " *";
    out << variable_name << " = new " << cl->GetName() << "(" << constructor_agrs << ");" << std::endl;
 }
+
+////////////////////////////////////////////////////////////////////////////////
+/// Save array in the output stream "out".
+/// Create unique variable name based on prefix value
+
+TString TObject::SavePrimitiveArray(std::ostream &out, const char *prefix, Int_t len, Double_t *arr)
+{
+   static int arrid = 0;
+
+   TString arrname = TString::Format("%s_darr%d", prefix, arrid++);
+
+   out << "   Double_t " << arrname << "[" << len << "] = { ";
+   if (len > 0) {
+      const auto old_precision{out.precision()};
+      constexpr auto max_precision{std::numeric_limits<double>::digits10 + 1};
+      out << std::setprecision(max_precision);
+      for (Int_t i = 0; i < len-1; i++) {
+         out << arr[i] << ",";
+         if (i && (i % 16 == 0))
+            out << std::endl << "   ";
+         else
+            out << " ";
+      }
+      out << arr[len - 1];
+      out << std::setprecision(old_precision);
+   }
+   out << " };" << std::endl;
+   return arrname;
+}
+
 
 ////////////////////////////////////////////////////////////////////////////////
 /// Save a primitive as a C++ statement(s) on output stream "out".
--- a/core/base/src/TObject.cxx
+++ b/core/base/src/TObject.cxx
@@ -785,7 +785,7 @@ void TObject::SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char
 
 TString TObject::SavePrimitiveArray(std::ostream &out, const char *prefix, Int_t len, Double_t *arr)
 {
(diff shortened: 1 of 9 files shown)


core/base/src/TObject.cxx
```diff
@@ -778,6 +779,36 @@ void TObject::SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char
    out << variable_name << " = new " << cl->GetName() << "(" << constructor_agrs << ");" << std::endl;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// Save array in the output stream "out".
+/// Create unique variable name based on prefix value
+
+TString TObject::SavePrimitiveArray(std::ostream &out, const char *prefix, Int_t len, Double_t *arr)
+{
+   static int arrid = 0;
```
pcanal, line 788:
```suggestion
   thread_local int arrid = 0;
```
to work toward `SavePrimitive` begin callable from 2 distinct threads on 2 distinct objects.

hist/hist/src/TGraph.cxx
```diff
@@ -2145,66 +2145,35 @@ void TGraph::SaveAs(const char *filename, Option_t *option) const
...
 
 ////////////////////////////////////////////////////////////////////////////////
 /// Save histogram and list of functions of TGraph as C++ statement
 /// Used in all TGraph-derived classes
 
-void TGraph::SaveHistogramAndFunctions(std::ostream &out, const char *varname, Int_t &frameNumber, Option_t *option)
+void TGraph::SaveHistogramAndFunctions(std::ostream &out, const char *varname, Option_t *option)
 {
-   char quote = '"';
+   static Int_t frameNumber = 0;
```
pcanal, line 2166:
```suggestion
   thread_local Int_t frameNumber = 0;
```

pcanal (approved):
Thanks.

core/base/src/TObject.cxx
```diff
@@ -778,6 +779,36 @@ void TObject::SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char
...
+
+   out << "   Double_t " << arrname << "[" << len << "] = { ";
+   if (len > 0) {
+      const auto old_precision{out.precision()};
+      constexpr auto max_precision{std::numeric_limits<double>::digits10 + 1};
+      out << std::setprecision(max_precision);
+      for (Int_t i = 0; i < len-1; i++) {
+         out << arr[i] << ",";
+         if (i && (i % 16 == 0))
+            out << std::endl << "   ";
```
silverweed, line 800: This should use `"\n"` instead of `std::endl` (we don't need to flush the stream every time)

core/base/src/TObject.cxx
```diff
@@ -778,6 +779,36 @@ void TObject::SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char
    out << variable_name << " = new " << cl->GetName() << "(" << constructor_agrs << ");" << std::endl;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// Save array in the output stream "out".
+/// Create unique variable name based on prefix value
+
+TString TObject::SavePrimitiveArray(std::ostream &out, const char *prefix, Int_t len, Double_t *arr)
```
silverweed, line 786: Why is this a `TObject` method? What `TObject` member field does it need?

silverweed (dismissed):
Left a couple of comments

core/base/src/TObject.cxx
pcanal: It is solely used for implementation of override of `TObject::SavePrimitive` (And as such is marked as `protected`, a free function would be, for better or worth, more accessible .. here it is left as an implementation detail of sort)

linev: As Philippe comment, free function has no big sense while saving array only should be used in `SavePrimitive` implementations. Like in several places of TGraph-derived classes.

core/base/src/TObject.cxx
linev: I do not think that performance here is a major concern. 
For me `std::endl` is preferable while it is standard line ending marker.

silverweed: > For me std::endl is preferable while it is standard line ending marker.

Please read the [std::endl documentation](https://en.cppreference.com/w/cpp/io/manip/endl) before dismissing my point: `std::endl` has nothing to do with being a "standard line ending marker" (whatever that means) but it specifically puts a `\n` followed by a `ostream.flush()`. As such, the documentation itself suggests not using it unless you specifically want to flush the output stream (which only makes sense at the very end of the output - if at all)

silverweed (changes requested):
Please don't dismiss other people's reviews before they even had a chance to answer.

core/base/src/TObject.cxx
linev: Performance here not a problem at all.
Even writing 1^10 array entries here will be done in reasonable time.
Real problem is in many other ROOT classes - like TPolyMarker from this PR.
Without usage of this method macro execution for just 50K lines takes 37s or even 110s if compiled with gcc.
And in ROOT exactly 13987 places where `std::endl` is used. 
To be consequent - one need to review all these places and replace approx 13800 or maybe 13900 of them by "\n"

silverweed (approved):

core/base/src/TObject.cxx
pcanal: This particular instance (and most of the 13987 others) are not particularly important in themselves.  However when writing new code or updating existing code, I do recommend to use `\n` to introduce new lines and `std::endl` only when the intent is to flush the output.

PR #17852, opened 2025-02-28
[Minuit2] Modernize linear algebra code

guitargeek:
Part of the modernization campain to make the Minuit2 code more accessible to Minuit2 users and developers.

This commit is about modernizing and isolating the linear algebra code that deals with symmetric matricies and vectors.

  * Translate Fortran-style C++ to more idiomatic C++. This should have a positive effect on performance, because the typical manual loop unrolling from Fortran only gets in the way of the optimizations in modern C++ compilers.

  * Greatly simplify class sturcture of `ABObj` and friends helper classes for the linear algebra operators, thanks to what is possible with C++ now

  * Merge all headers relevant to linear algebra in Minuit into `MnMatrix.h`

  * Merge all `.cxx` sources to one translation unit (`MnMatrix.cxx`)

  * Avoid using preprocessor macros

  * Various other refactorings without affecting the functionality


lmoneta (approved):
LGTM!
Thank you Jonas for re-arrenging and modernize the linear algebra code

PR #17853, opened 2025-03-01
Cppyyfix

DingXuefeng:
# This Pull request:

## Changes or fixes:

fix: forgot to add r for r-string.

## Checklist:

- [x] tested changes locally

This PR fixes # 

fix: forgot to add r for r-string.

### DIFF
--- a/bindings/pyroot/cppyy/patches/multi_python_cppyy.patch
+++ b/bindings/pyroot/cppyy/patches/multi_python_cppyy.patch
@@ -74,7 +74,7 @@ index b3b1b84..0d2fa35 100644
 +# tell cppyy that libcppyy_backend is versioned
 +def _check_py_version(lib_name, cbl_var):
 +    import re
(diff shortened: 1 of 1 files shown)


guitargeek (changes requested):
Hi @DingXuefeng, thanks for your PR! However, this change is not needed at this point. You are only changing a patch file to track changes with respect to upstream `cppyy`. If you want your change to have an effect, you need to change the actual Python file, in this case that one:

https://github.com/root-project/root/blob/v6-30-00-patches/bindings/pyroot/cppyy/cppyy/python/cppyy/__init__.py#L72

Of course it's also important to update the patch accordingly (like you already did :+1:), but updating the patch is not enough :slightly_smiling_face:

guitargeek:
Closing since @DingXuefeng didn't follow up. Please feel free to open a new PR if you should have new developments! Thanks.

PR #17854, opened 2025-03-01
[ntuple] Add alignment calculation in RClassField::Attach method

### LINKED ISSUE #16765: [ntuple] Wrong alignment with transient class member
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

`RClassField::Attach` only takes into account persistent members to determine alignment:
https://github.com/root-project/root/blob/30917203c35b075cd9f0f87165596ee0c63f17db/tree/ntuple/v7/src/RField.cxx#L1874-L1879

### Reproducer

```c++
#include <ROOT/RField.hxx>

class ClassWithTransient {
  char a;
  int t; //!
  char b;
};

void ntuple_class_align() {
  ROOT::Experimental::RField<ClassWithTransient> f("f");
  std::cout << "value size: " << f.GetValueSize() << ", alignment: " << f.GetAlignment() << "\n";
}
```
output:
```
Processing ntuple_class_align.C...
value size: 12, alignment: 1
```
This can lead to problems if such class is embedded into a record, for example a `std::pair`.

### ROOT version

`master`

### Installation method

from source

### Operating system

any

### Additional context

_No response_

Aditya-138-12:
# This Pull request:
Add alignment calculation in RClassField::Attach method

## Changes or fixes:
1. Get the TypeName of the class
2. Loop over all the data members, and find maximum alignment requirement
3. Update the maximum alignment considering both the class members and child fields
4. This takes into account the transient members along with the persistent members.

Now this 
```
#include <ROOT/RField.hxx>

class ClassWithTransient {
  char a;
  int t; //!
  char b;
};

void ntuple_class_align() {
  ROOT::Experimental::RField<ClassWithTransient> f("f");
  std::cout << "value size: " << f.GetValueSize() << ", alignment: " << f.GetAlignment() << "\n";
}
```
Return this
```value size: 12, alignment: 4```

## Checklist:

- [x] tested changes locally

This PR fixes #16765

### DIFF
--- a/tree/ntuple/v7/src/RFieldMeta.cxx
+++ b/tree/ntuple/v7/src/RFieldMeta.cxx
@@ -185,6 +185,15 @@ ROOT::Experimental::RClassField::RClassField(std::string_view fieldName, TClass 
 
 void ROOT::Experimental::RClassField::Attach(std::unique_ptr<RFieldBase> child, RSubFieldInfo info)
 {
+   auto *cls = TClass::GetClass(GetTypeName().c_str());
+   if(cls) {
+      TIter memberIter(cls->GetListOfDataMembers());
+      TDataMember *dataMember;
+      while ((dataMember = (TDataMember*)memberIter())) {
+         std::size_t maxAlign = dataMember->GetUnitSize();
+         fMaxAlignment = std::max(fMaxAlignment, maxAlign);
+      }
+   }
    fMaxAlignment = std::max(fMaxAlignment, child->GetAlignment());
    fSubFieldsInfo.push_back(info);
    RFieldBase::Attach(std::move(child));


hahnjo (changes requested):
Hi @Aditya-138-12, thanks for the PR! Some thoughts:
1. `RClassField::Attach` is called for every (persistent) member, which means the added code is executed multiple times. The result is probably still correct, but this can get expensive for very big classes with many members.
2. Using `TDataMember::GetUnitSize()` returns the size of the member, not its alignment. That works ok for fundamental types, but if the member is itself an object, it will lead to severe overalignment.

I think the proper way to fix this problem is exposing alignment information from ROOT Meta. And once we have that, I think we should not try to determine the alignment ourselves but call a (to-be-implemented) `TClass::GetAlignment` directly.

Aditya-138-12:
Hi @hahnjo,
I was thinking for the comlex classes, maybe we can try recursive approach but it would be expensive too...
I will try to explore ROOT's Meta for this...
Thank you for your time!

Aditya-138-12:
Hi @hahnjo,
I gave it a try but I was overwhelmed by the complexity of the code base, maybe I should leave this to the relevant ROOT experts. 
Thank you for time!!

PR #17856, opened 2025-03-02
[Python] Use `magic` submodule from `IPython.Core`

guitargeek:
Relying on the `IPython.core.magic` module instead of `metakernel.magic`
avoids the metakernel dependency, making sure that failures to import
`metakernel` are not resulting in an error.

https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.magic.html

Also, remove the `@options` decorators with additions to the docstrings
only, because the `metakernel.magic` API was not used correctly there
anyway. If options are registered via the decorators, they need to be
retrieved with `Magic.get_args()` in the implementation of the magic,
which didn't happen. Instead, the input strings were just parsed
manually anyway.

Sister PR in roottest:
  * https://github.com/root-project/roottest/pull/1266

### DIFF
--- a/bindings/jupyroot/python/JupyROOT/kernel/magics/cppmagic.py
+++ b/bindings/jupyroot/python/JupyROOT/kernel/magics/cppmagic.py
@@ -13,22 +13,21 @@
 # For the list of contributors see $ROOTSYS/README/CREDITS.                    #
 ################################################################################
 
-from metakernel import Magic, option
+from IPython.core.magic import Magics, cell_magic, magics_class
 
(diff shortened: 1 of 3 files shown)


guitargeek:
Replacing the magics is not enough, because we still use `MetaKernel` as as class itself.

Needs to be fixed upstream, an issue is already opened:
https://github.com/Calysto/metakernel/issues/288

guitargeek:
I think the best we can do for now is to work around the problem by constraining the ipython version.

PR #17861, opened 2025-03-03
[io] TFileMergerTest: add missing TBranch include

silverweed:
A missing include in #17652 causes a compiler warning (aka an error in dev=ON)


ferdymercury (approved):
Thks!

PR #15853, opened 2024-06-14
[CI] enable dev=on on debian12

silverweed:
# This Pull request:
adds the `-Ddev=on` flag on the debian CI machine. It's a "test run" to see if we can later enable it on more (potentially all) build machines.

Here's what the flag does:
- enables asserts
- (Linux|BSD) enables -Werror
- (Linux|BSD) does not relink if a dependent .so has changed (CMAKE_LINK_DEPENDS_NO_SHARED On)
- (Linux|BSD) splits debug info (-gsplit-dwarf)
- (Linux|BSD) uses lld if available
- defines USE_LESS_INCLUDES

PR #17863, opened 2025-03-03
[skip-ci] Suggest using dev=ON in contributing.txt

silverweed:
Using `dev=ON` is a good practice that should ideally be adopted by most, if not all, ROOT developers by default. We should encourage contributors to enable it.


ferdymercury:
Why not enabling one of the CI builds as dev=on ?

ferdymercury (approved):

silverweed:
@ferdymercury it's unfortunately not easy to do and we don't want to enable it on the entire CI because we still want to test the default build.
I opened a PR (#15853) a while ago that enabled dev=ON on one of the CI machines (which should be enough to catch warnings), but unfortunately I just left it there for a while without much follow-up. Last week I managed to get it working, so I believe it can now be merged :)
We just need @dpiparo's and others' approval, but it should be good to go.

PR #17865, opened 2025-03-03
[NOT TO MERGE] Review clang patches that can be removed

devajithvs:
# This Pull request:

Remove patches that offer no significant benefit. All tests pass on my local machine. ROOT startup times are almost the same. 

TODO: Do thorough testing and performance evaluations.

## Checklist:

- [ ] tested changes locally

This PR fixes #


hahnjo:
Partially now resubmitted as https://github.com/root-project/root/pull/23165; `Expose all files included in PCH/PCM to FileManager.` is still needed to avoid the reproducer from ROOT-6794 to fail.

hahnjo:
`Store token literal data in PCH` as well, see https://github.com/root-project/root/pull/23235

PR #17867, opened 2025-03-03
[cling] Ensure correct template argument printing for large integers

devajithvs:
# This Pull request:
We do not need a patch to set `ULL` suffix manually on top of Clang.

`Policy.AlwaysIncludeTypeForTemplateArgument = true` does the job.

This will get rid of the patch: https://github.com/root-project/llvm-project/commit/5ae13042588e6ab98eefd33c8b43bfe7f52cb6ee

I started the process of upstreaming (https://github.com/llvm/llvm-project/pull/129235).
But on further digging, I realized we can simply use `AlwaysIncludeTypeForTemplateArgument = true`

## Checklist:

- [x] tested changes locally

This PR fixes #


pcanal:
> Policy.AlwaysIncludeTypeForTemplateArgument = true does the job.

Does that also change the way smaller numbers are printed?

devajithvs:
> Does that also change the way smaller numbers are printed?

Yes. I realized that only after the tests failed. I'll try to figure out something.

devajithvs:
Closing this as it will be a breaking change (can reopen this at a later stage when we really want to get rid of all the clang  patches)

Issue #17843, opened 2025-02-27
[TTreeReader] Crash in TNotifyLink with reused TChain with friend


ferdymercury:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

```
0 0 0
1 -1 1
2 -2 2
3 -3 3
4 -4 4
5 -5 5
6 -6 6
7 -7 7
8 -8 8
9 -9 9

 *** Break *** segmentation violation

#11 0x00007fc3be9c866a in TNotifyLink<TTreeReader>::Notify (this=0x7ffe3bf64ed8) at /opt/root_src/core/base/inc/TNotifyLink.h:139
#12 0x00007fc3be922a28 in TNotifyLink<ROOT::Detail::TBranchProxy>::Notify (this=0x55718d59b5d8) at /opt/root_src/core/base/inc/TNotifyLink.h:141
#13 0x00007fc3be5c7abe in TChain::LoadTree (this=0x55718cd92860, entry=10) at /opt/root_src/tree/tree/src/TChain.cxx:1800
#14 0x00007fc3be9c1179 in TTreeReader::SetEntryBase (this=0x7ffe3bf64eb0, entry=10, local=false) at /opt/root_src/tree/treeplayer/src/TTreeReader.cxx:668
#15 0x00007fc3be9c25c6 in TTreeReader::SetEntry (this=0x7ffe3bf64eb0, entry=10) at /opt/root_src/tree/treeplayer/inc/TTreeReader.h:217
#16 0x00007fc3cf0b3345 in TTreeReader::Next (this=0x7ffe3bf64eb0) at /home/user/builds/build-root_src-Desktop-Debug/include/TTreeReader.h:210
#17 0x00007fc3cf0b2bf9 in TTreeReaderFriend () at /tmp/TTreeReaderFriend.C:71
#18 0x00007fc3d16283b8 in cling::IncrementalExecutor::executeWrapper(llvm::StringRef, cling::Value*) const () from /home/user/builds/build-root_src-Desktop-Debug/lib/libCling.so
#19 0x00007fc3d15aa41c in cling::Interpreter::RunFunction(clang::FunctionDecl const*, cling::Value*) () from /home/user/builds/build-root_src-Desktop-Debug/lib/libCling.so
[...]

PR #17823, opened 2025-02-25
[test] for TTreeReader loads wrong entry from chain friend

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes/Tests https://its.cern.ch/jira/browse/ROOT-9886

## Checklist:

- [x] tested changes locally
- [x] updated the docs (if necessary)

PR #17868, opened 2025-03-03
[test] implement test reusing chain

ferdymercury:
To demonstrate https://github.com/root-project/root/issues/17843


ferdymercury:
I tried with master and it seems it doesn't crash, so maybe the problem is only in the latest stable binaries?

vepadulano:
Dear @ferdymercury , thanks for opening this PR to check the reproducer! I took the liberty of starting the CI also on the other platforms. This is indeed a very curious situation, I would be surprised by an issue with TChain that can only be reproduced with the published binaries. Let's see when the CI finishes and then we proceed with the investigation

ferdymercury:
Continued on https://github.com/root-project/root/pull/17823 and https://github.com/root-project/root/issues/17843

PR #17870, opened 2025-03-03
[eve7] Add single view option

alja:
# This Pull request:
Add option to have a single client view using parameter` Single=ViewName`. This can be extend to undock/dock window manager feature when the  multi-client option is enabled.

## Changes or fixes:
* Unnecessary client long messages. 
* Cleanup of lego.C demo.
* Improves axis font
* Updates RenderCore to version 1.7


alja:
### Example of single client view and the standard view 

Screenshot of collection_proxies.C demo with multiple client options `REveManager::AllowMultipleRemoteConnections(false, fase)`

The single view uses URL parameter `Single=RhoZ%20View
`
Screenshot from 2025-03-03 10-36-20

alja:
@linev @osschar Please check.

alja:
@linev 

We got into the project when we needed multiple single views together with the 'standard' view for CMS control room.
FireworksWeb has multiple clients enabled by the default and there I added GUI options to undock/dock  windows  with use of a simple BroadcastChannel.
For ROOT demo I could add server action like in event_demo.C[1] and call xdg_-open. Here I'm a bit concerned about what will happen when one runs the demo remotely.
What do you think?

Alternatively I could make different toolbar actions in EveViews for multi client and single client mode. This may need to be implemented in separate PR.

[1]https://github.com/root-project/root/blob/master/tutorials/visualisation/eve7/event_demo.C#L333

linev:
@alja 

> For ROOT demo I could add server action like in event_demo.C[1] and call xdg_-open. Here I'm a bit concerned about what will happen when one runs the demo remotely.

It is not necessary to use `xdg-open` tool to start new browser window. One should use `RWebWindow::Show()` functionality where display arguments in form of `RWebDisplayArgs` object can be provided. And it can include extra URL parameters like in your case. 

May be not in this PR, but for follow up PR one can provide possibility to specify such extra URL parameters in  `event_demo.C` example (or any other). Just to clearly see that functionality exists and how it can be used. It can be extra arguments of `REveManager::Show()` method.

linev (approved):

PR #17871, opened 2025-03-03
[Minuit2] Also cache transformed parameter values in gradient evaluation

guitargeek:
Just like it was already done in MnHesse, the
Numerical2PGradientCalculator now caches the transformed parameter
values.

This speeds up the minimization in the ATLAS Higgs combination benchmark
by 10 % from 30.7 seconds down to 27.6 s.

In more optimized RooFit-based likelihoods, the relative speedup is even
more significant. Minimizing the CMS Higgs discovery likelihood is 50 %
faster, for example.


lmoneta (approved):
LGTM!
Thank you Jonas for applying this improvement also in the gradient calculation

PR #17872, opened 2025-03-03
[RF] Remove irrelevant allocation in RooRealVar IO constructor

guitargeek:
The IO constructor should not allocate memory for the smart pointer members like `_binning`, because it will be reset anyway by the IO.

Not doing this avoids some overhead when reading a large RooWorkspace.

### DIFF
--- a/roofit/roofitcore/src/RooRealVar.cxx
+++ b/roofit/roofitcore/src/RooRealVar.cxx
@@ -91,7 +91,7 @@ RooRealVarSharedProperties& RooRealVar::_nullProp()
 ////////////////////////////////////////////////////////////////////////////////
 /// Default constructor.
 
-RooRealVar::RooRealVar()  :  _error(0), _asymErrLo(0), _asymErrHi(0), _binning(new RooUniformBinning())
+RooRealVar::RooRealVar()
 {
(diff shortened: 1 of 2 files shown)


roofit/roofitcore/src/RooRealVar.cxx
```diff
@@ -91,7 +91,7 @@ RooRealVarSharedProperties& RooRealVar::_nullProp()
 ////////////////////////////////////////////////////////////////////////////////
 /// Default constructor.
 
-RooRealVar::RooRealVar()  :  _error(0), _asymErrLo(0), _asymErrHi(0), _binning(new RooUniformBinning())
+RooRealVar::RooRealVar()
```
pcanal, line 94: The major challenge is whether the relevant code handles this 'new' state (i.e. `_binning` can be null if the user uses the default constructor directly or indirectly (creation done by I/O but reading not 'yet' done).

A cursory look indicates a few places where there might be a problem.

guitargeek: Yes you are right, and the failing unit test suggests that someone was expecting the default constructor to work outside of IO, and hence a test was added. Need to think more about it

PR #17873, opened 2025-03-03
[rntuple] Fix minor typo in the spec

ariostas:
This PR fixes a minor typo in the RNTuple spec. It showed `SplitInt32` as having 64 bits instead of 32.


hahnjo (approved):
Absolutely, thanks for catching! For future changes, note that we usually put `[ntuple]` in the commit message because that's the path in the repository (the `R` is a conventional prefix for class names). I will fix this when merging.

PR #17875, opened 2025-03-04
[ntuple] Merger: handle RNTuples with empty schemas

silverweed:
# This Pull request:
fixes an edge case in the Merger where the input source(s) have an empty schema. Adds unit tests for the various cases.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/test/ntuple_merger.cxx
+++ b/tree/ntuple/v7/test/ntuple_merger.cxx
@@ -133,7 +133,6 @@ TEST(RNTupleMerger, MergeSymmetric)
          sourcePtrs.push_back(s.get());
       }
 
-      // Now merge the inputs
       RNTupleMergeOptions opts;
       {
          auto destination = std::make_unique<RPageSinkFile>("ntuple", fileGuard3.GetPath(), RNTupleWriteOptions());
@@ -239,7 +238,6 @@ TEST(RNTupleMerger, MergeAsymmetric1)
          sourcePtrs.push_back(s.get());
       }
 
-      // Now merge the inputs
       // We expect this to fail in Filter and Strict mode since the fields between the sources do NOT match
       RNTupleMergeOptions opts;
       {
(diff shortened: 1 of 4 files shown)


tree/ntuple/v7/test/ntuple_merger.cxx
```diff
@@ -2787,3 +2787,227 @@ TEST(RNTupleMerger, MergeIncrementalLMExtMemFile)
...
+      {
+         auto destination = std::make_unique<RPageSinkFile>("ntuple", fileGuardOut.GetPath(), RNTupleWriteOptions());
+         opts.fMergingMode = ENTupleMergingMode::kStrict;
+         RNTupleMerger merger{std::move(destination)};
+         auto res = merger.Merge(sourcePtrs, opts);
+         EXPECT_TRUE(bool(res));
+      }
+   }
+
+   // We expect the output ntuple to have no entries
```
enirolf, line 2827: We should somewhere document the conditions under which entries are preserved with the different merge modes when empty schemas are involved.

enirolf (approved):
LGTM! Just one comment/thought regarding documentation.

tree/ntuple/v7/test/ntuple_merger.cxx
```diff
@@ -2787,3 +2775,224 @@ TEST(RNTupleMerger, MergeIncrementalLMExtMemFile)
...
+         sourcePtrs.push_back(s.get());
+      }
+
+      RNTupleMergeOptions opts;
+      {
+         auto destination = std::make_unique<RPageSinkFile>("ntuple", fileGuardOut.GetPath(), RNTupleWriteOptions());
+         opts.fMergingMode = ENTupleMergingMode::kFilter;
+         RNTupleMerger merger{std::move(destination)};
+         auto res = merger.Merge(sourcePtrs, opts);
+         EXPECT_TRUE(bool(res));
```
pcanal, line 2880: Is this silently creating an empty `RNTuple`?

tree/ntuple/v7/test/ntuple_merger.cxx
silverweed: I wasn't sure where to add it, so I added it to the Merge() method documentation.

tree/ntuple/v7/test/ntuple_merger.cxx
pcanal: Can we add a `Warning` message (maybe only if `gDebug > 1`) when that happens?  (i.e. I would assume it is unlikely to be the intended user behavior).

PR #17880, opened 2025-03-04
[CMake] Only build Clad on a single core if the workaround if needed

guitargeek:
The problem around for which the added `-j 1` was a workaround was reported and solved upstream in CMake:

https://gitlab.kitware.com/cmake/cmake/-/issues/26398

It's better to not do the workaround for the fixed CMake versions anymore, because actually the increase in build time can be quite significant.

On my machine, building ROOT `master` with a warm ccache takes 5m33s with 32 threads, which reduces to 4m6s without the workaround. I did the measurements twice each, so they are reproducible!

Follows up on #16682.


bellenot (approved):

hahnjo:
There was a failure in the nightlies with the exact same pattern described in #16682: https://github.com/root-project/root/actions/runs/13667097289/job/38210410533 I think we will need to revert this PR...

PR #17881, opened 2025-03-04
[ntuple] Distinguish primary and auxiliary ntuples in `RNTupleProcessor::CreateJoin`

enirolf:
Grouping them all in one collection doesn't really make sense from a semantic point of view. This separation should make the roles of the different ntuples more clear.

A next consideration is the representation of the model(s). Ideally (IMO) there should just be one model passed to this factory that already represents the schema after the join. However, this would put a lot of burden on the user to create a "correct" model, so this requires a bit more thought and will be followed up on later.


tree/ntuple/v7/test/ntuple_processor_join.cxx
```diff
@@ -117,16 +108,14 @@ TEST_F(RNTupleJoinProcessorTest, Basic)
...
 {
    try {
-      std::vector<RNTupleOpenSpec> ntuples = {{fNTupleNames[0], fFileNames[0]}, {fNTupleNames[0], fFileNames[0]}};
-      auto proc = RNTupleProcessor::CreateJoin(ntuples, {});
+      auto proc =
+         RNTupleProcessor::CreateJoin({fNTupleNames[0], fFileNames[0]}, {{fNTupleNames[0], fFileNames[0]}}, {});
       FAIL() << "ntuples with the same name cannot be joined horizontally";
    } catch (const ROOT::RException &err) {
-      EXPECT_THAT(err.what(), testing::HasSubstr("horizontal joining of RNTuples with the same name is not allowed"));
+      EXPECT_THAT(err.what(), testing::HasSubstr("joining RNTuples with the same name is not allowed"));
```
pcanal, line 115: Is it planned to lift this limitation (and/or is there already an alternative)?  For example in TTree, there is a  syntax to allow setting an alias to allow registration (and distinction) of same name friends.

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -106,25 +106,22 @@ ROOT::Experimental::RNTupleProcessor::CreateChain(std::vector<std::unique_ptr<RN
...
+                                                 const std::vector<RNTupleOpenSpec> &auxNTuples,
                                                  const std::vector<std::string> &joinFields,
                                                  std::string_view processorName,
                                                  std::vector<std::unique_ptr<RNTupleModel>> models)
 {
-   if (ntuples.size() < 1)
-      throw RException(R__FAIL("at least one RNTuple must be provided"));
-
-   if (models.size() > 0 && models.size() != ntuples.size()) {
+   if (models.size() > 0 && models.size() != (auxNTuples.size() + 1)) {
```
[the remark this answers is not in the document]
enirolf: Done: https://github.com/root-project/root/blob/0ce1f45a07e56af739df831d30b644d89dcf3455/tree/ntuple/v7/src/RNTupleProcessor.cxx#L124

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -359,29 +352,37 @@ ROOT::Experimental::RNTupleJoinProcessor::RNTupleJoinProcessor(const RNTupleOpen
...
       const auto &[fieldContext, _] =
          fFieldContexts.try_emplace(fieldName, field.Clone(fieldName), fEntry->GetToken(fieldName));
       ConnectField(fieldContext->second, *fPageSource, *fEntry);
    }
+
+   for (unsigned i = 0; i < auxNTuples.size(); ++i) {
+      if (models.empty())
+         AddAuxiliary(auxNTuples[i], joinFields);
+      else
+         AddAuxiliary(auxNTuples[i], joinFields, std::move(models[i + 1]));
```
pcanal, line 384: This requires `models.size() > auxNTuples.size()` is it checked somewhere in the routine and/or is it already guaranteed by the control flow (`models` and `auxNTuples` are parameters of this function)?

enirolf: It is checked in `CreateJoin` (l. 124) and this constructor is private, so unless I'm overlooking something it should be guaranteed by the workflow.

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -353,43 +353,46 @@ public:
...
    /// \brief Create an RNTupleProcessor for a *join* (i.e., a horizontal combination) of RNTuples.
    ///
-   /// \param[in] ntuples A list specifying the names and locations of the RNTuples to process. The first RNTuple in the
-   /// list will be considered the primary RNTuple and drives the processor iteration loop. Subsequent RNTuples are
-   /// considered auxiliary, whose entries to be read are determined by the primary RNTuple (and do not necessarily
-   /// have to be in sequential order).
+   /// \param[in] primaryNTuple The name and location of the primary RNTuple, which will drive the processor iteration
+   /// loop.
+   /// \param[in] auxNTuples The names and locations of the RNTuples to join the primary RNTuple with. Which entries of
+   /// these auxiliary RNTuples are read is determined by the primary RNTuple, which does not necessarily happens in
```
vepadulano, line 359: There is a small typo, but I actually wanted to ask a clarification regarding this sentence. What does not necessarily happen in sequential order?
```suggestion
   /// these auxiliary RNTuples are read is determined by the primary RNTuple, which does not necessarily happen in
```

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -521,13 +524,19 @@ private:
    /// \brief Construct a new RNTupleJoinProcessor.
    ///
    /// \param[in] mainNTuple The source specification (name and storage location) of the primary RNTuple.
+   /// \param[in] auxNTUples The source specifications (name and storage location) of the auxiliary RNTuples.
+   /// \param[in] joinFields The names of the fields on which to join, in case the specified RNTuples are unaligned.
+   /// The join is made based on the combined join field values, and therefore each field has to be present in each
+   /// specified RNTuple. If an empty list is provided, it is assumed that the specified ntuple are fully aligned.
```
vepadulano, line 530:
```suggestion
   /// specified RNTuple. If an empty list is provided, it is assumed that the specified ntuples are fully aligned.
```

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -521,13 +524,19 @@ private:
...
    /// \param[in] processorName Name of the processor. Unless specified otherwise in RNTupleProcessor::CreateJoin, this
    /// is the name of the main RNTuple.
-   /// \param[in] model The model that specifies which fields should be read by the processor. The pointer returned by
-   /// RNTupleModel::MakeField can be used to access a field's value during the processor iteration. When no model is
-   /// specified, it is created from the RNTuple's descriptor.
-   RNTupleJoinProcessor(const RNTupleOpenSpec &mainNTuple, std::string_view processorName,
-                        std::unique_ptr<RNTupleModel> model = nullptr);
+   /// \param[in] models The models that specify which fields should be read by the processor, ordered according to
+   /// {mainNTuple, auxNTuple[0], ...} The pointer returned by RNTupleModel::MakeField can be used to access a field's
```
vepadulano, line 534:
```suggestion
   /// {mainNTuple, auxNTuple[0], ...}. The pointer returned by RNTupleModel::MakeField can be used to access a field's
```

vepadulano, line 534: This description of the `models` parameter is slightly different from other descriptions of the same parameter in other places. Would it be better to homogenise them perhaps?

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -19,13 +19,13 @@
 
 namespace {
 using ROOT::Experimental::RNTupleOpenSpec;
-void EnsureUniqueNTupleNames(const std::vector<RNTupleOpenSpec> &ntuples)
+void EnsureUniqueNTupleNames(const RNTupleOpenSpec primaryNTuple, const std::vector<RNTupleOpenSpec> &auxNTuples)
```
vepadulano, line 22:
```suggestion
void EnsureUniqueNTupleNames(const RNTupleOpenSpec &primaryNTuple, const std::vector<RNTupleOpenSpec> &auxNTuples)
```

tree/ntuple/v7/test/ntuple_processor.cxx
```diff
@@ -165,11 +165,11 @@ TEST_F(RNTupleProcessorTest, ChainedChain)
...
 {
-   std::vector<RNTupleOpenSpec> ntuples{{fNTupleNames[0], fFileNames[0]}, {fNTupleNames[1], fFileNames[1]}};
-
    std::vector<std::unique_ptr<RNTupleProcessor>> innerProcs;
-   innerProcs.push_back(RNTupleProcessor::CreateJoin(ntuples, {}));
-   innerProcs.push_back(RNTupleProcessor::CreateJoin(ntuples, {"i"}));
+   innerProcs.push_back(
+      RNTupleProcessor::CreateJoin({fNTupleNames[0], fFileNames[0]}, {{fNTupleNames[1], fFileNames[1]}}, {}));
+   innerProcs.push_back(
+      RNTupleProcessor::CreateJoin({fNTupleNames[0], fFileNames[0]}, {{fNTupleNames[1], fFileNames[1]}}, {"i"}));
```
vepadulano, line 172: Is this practically chaining together a join composition made of two files that are aligned with another join composition with unaligned events?

tree/ntuple/v7/test/ntuple_processor_join.cxx
```diff
@@ -81,23 +81,14 @@ class RNTupleJoinProcessorTest : public testing::Test {
...
-      FAIL() << "creating a processor without at least one RNTuple should throw";
-   } catch (const ROOT::RException &err) {
-      EXPECT_THAT(err.what(), testing::HasSubstr("at least one RNTuple must be provided"));
-   }
-
-   ntuples = {{fNTupleNames[0], fFileNames[0]}};
-
-   auto proc = RNTupleProcessor::CreateJoin(ntuples, {});
+   // Primary ntuple only
+   auto proc = RNTupleProcessor::CreateJoin({fNTupleNames[0], fFileNames[0]}, {}, {});
```
vepadulano, line 87: I know this test was already here before, but looking at it again made me wonder. Is this just a sanity check or do we expect this syntax to be used? On the surface it looks weird that a user calls `CreateJoin` if there is only one RNTuple.

vepadulano (changes requested):
Thank you! I left some comments inline for discussion.

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
enirolf: I've reworden the last sentence a bit, but I'm not sure if that's what you meant. The description is different because here we're talking about a list of models, whereas in the other descriptions a single model is assumed.

tree/ntuple/v7/test/ntuple_processor.cxx
enirolf: Sort of but not really. Both joins use the same (aligned) RNTuples so they are aligned in both cases. However, in the second case the join table is used (i.e. we pretend they are unaligned). Tests for the truly unaligned case are covered already in `ntuple_processor_join.cxx`, this is more concerned with the composition itself.

tree/ntuple/v7/test/ntuple_processor_join.cxx
enirolf: I don't know if we *expect* this syntax to be used, but I don't see why we shouldn't allow this use case (we're also not disallowing creating a chain with only one ntuple). I could think of a use case where users more or less automatize the creation of the processor and then having this possibility could be useful. I don't think it matters much for performance because practically all `auxNTuple` stuff is just skipped, but this is of course something we can benchmark to make sure.

vepadulano (approved):
Thanks! Just a leftover typo.

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -353,43 +353,46 @@ public:
...
+   /// \param[in] models A list of models for the RNTuples. This list must either contain a model for the primary
+   /// RNTuple and each auxiliary RNTuple (following the specification order), or be empty. When the list is empty, the
+   /// default model (i.e. containing all fields) will be used for each RNTuple.
    ///
    /// \return A pointer to the newly created RNTupleProcessor.
-   static std::unique_ptr<RNTupleProcessor> CreateJoin(const std::vector<RNTupleOpenSpec> &ntuples,
-                                                       const std::vector<std::string> &joinFields,
-                                                       std::vector<std::unique_ptr<RNTupleModel>> models = {});
+   static std::unique_ptr<RNTupleProcessor>
+   CreateJoin(const RNTupleOpenSpec &primaryNTuple, const std::vector<RNTupleOpenSpec> &auxNTuples,
```
pcanal, line 370: Is there a plan to allow creating join that involve in-memory `RNTuple` (for example one that is stored inside a `TMemFile`)?

pcanal (approved):
LGTM.  Thanks.

tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
enirolf: I don't see why not! Let me try it out and see how much for this is already supported out of the box.

PR #17883, opened 2025-03-04
[io] properly stream TClonesArray with empty slots

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-6788 by @olifre

This fixes the crash, but not sure if it is correct.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/io/io/src/TStreamerInfoActions.cxx
+++ b/io/io/src/TStreamerInfoActions.cxx
@@ -2497,7 +2497,8 @@ namespace TStreamerInfoActions
       static INLINE_TEMPLATE_ARGS Int_t LoopOverCollection(TBuffer &buf, void *start, const void *end, const TConfiguration *config)
       {
          for(void *iter = start; iter != end; iter = (char*)iter + sizeof(void*) ) {
-            action(buf, *(void**)iter, config);
+            if (*(void**)iter)
+               action(buf, *(void**)iter, config);
          }
          return 0;
       }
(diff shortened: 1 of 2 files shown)


ferdymercury:
gentle ping @pcanal

io/io/src/TStreamerInfoActions.cxx
```diff
@@ -2491,7 +2491,8 @@ namespace TStreamerInfoActions
       static INLINE_TEMPLATE_ARGS Int_t LoopOverCollection(TBuffer &buf, void *start, const void *end, const TConfiguration *config)
       {
          for(void *iter = start; iter != end; iter = (char*)iter + sizeof(void*) ) {
-            action(buf, *(void**)iter, config);
+            if (*(void**)iter)
+               action(buf, *(void**)iter, config);
```
pcanal, line 2501: The issue with this solution is that it change the index of the element when they are read back (what is written *_is_* different from what is read).

pcanal: Instead we should consider 
```
            if (*(void**)iter)
               action(buf, *(void**)iter, config);
            else 
               buf << nullptr;
```
and have the converse in the reading part.

ferdymercury: I added a commit, but not sure if that's what you meant.

There are 10 more occurences of void** iter so maybe those should be modified, too?

pcanal: I guess so.  Thanks.

io/io/src/TStreamerInfoActions.cxx
```diff
@@ -2491,7 +2491,9 @@ namespace TStreamerInfoActions
...
+            buf << static_cast<const Char_t *>(nullptr);
```
pcanal:
```suggestion
            if (*(void**)iter)
               action(buf, *(void**)iter, config);
            else 
                buf << static_cast<const Char_t *>(nullptr);
```
I miss typed my previous recommendation (i.e. the `nullptr` pointer should only be stored if needed :) )

PR #17886, opened 2025-03-05
[tree] Avoid stack-use-after-scope by also resetting friends branch addresses

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/17843

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/tree/src/TChain.cxx
+++ b/tree/tree/src/TChain.cxx
@@ -2524,10 +2524,10 @@ void TChain::ResetBranchAddresses()
    if (fTree) {
       fTree->ResetBranchAddresses();
    }
-   for (TObject *frel : *fFriends) {
-      TTree *frtree = ((TFriendElement *)frel)->GetTree();
-      if (frtree) {
-         frtree->ResetBranchAddresses();
+   for (auto *frEl: TRangeDynCast<TFriendElement>(fFriends)) {
+      auto *frTree = frEl->GetTree();
+      if (frTree) {
+         frTree->ResetBranchAddresses();
       }
    }
 }
--- a/tree/tree/src/TChain.cxx
+++ b/tree/tree/src/TChain.cxx
@@ -2524,7 +2524,7 @@ void TChain::ResetBranchAddresses()
    if (fTree) {
       fTree->ResetBranchAddresses();
    }
-   for (auto *frEl: TRangeDynCast<TFriendElement>(fFriends)) {
+   for (auto *frEl : TRangeDynCast<TFriendElement>(fFriends)) {
       auto *frTree = frEl->GetTree();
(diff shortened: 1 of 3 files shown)


tree/tree/src/TTree.cxx
```diff
@@ -8129,9 +8129,9 @@ void TTree::ResetAfterMerge(TFileMergeInfo *info)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-/// Tell all of our branches to set their addresses to zero.
+/// Tell a branch to set their address to zero.
```
ferdymercury, line 8132:
```suggestion
/// Tell a branch to set its address to zero.
```

vepadulano (approved):
This looks sensible to do to me! Let's also wait for @pcanal's opinion. Thank you for such a thorough investigation!

tree/tree/src/TChain.cxx
```diff
@@ -2524,6 +2524,12 @@ void TChain::ResetBranchAddresses()
    if (fTree) {
       fTree->ResetBranchAddresses();
    }
+   for (auto *frEl: TRangeDynCast<TFriendElement>(fFriends)) {
```
ferdymercury, line 2527:
```suggestion
   for (auto *frEl : TRangeDynCast<TFriendElement>(fFriends)) {
```

tree/tree/src/TChain.cxx
```diff
@@ -2524,6 +2524,12 @@ void TChain::ResetBranchAddresses()
    if (fTree) {
       fTree->ResetBranchAddresses();
    }
+   for (auto *frEl : TRangeDynCast<TFriendElement>(fFriends)) {
```
pcanal, line 2527: As is, this will infinitely recurse in case there is a circular dependency in friendship (trivial reproducer `chain->AddFriend(friendchain); friendchain->AddFriend(chain);`)

And example of what needs to be done can be found in `TTree::SetBranchStatus` and the usage of `fFriendLockStatus` and `TFriendLock`. We will also need to add a new enum value in `TTree::ELockStatusBits`.

Also `TTree::ResetBranchAddresses()` needs to be similarly updated.  (The same example that fails with `TChain` will fail if the chains are replaced by trees)

ferdymercury: Thanks for the help and hints. I've tried now to follow that example.

vepadulano:
Hi @ferdymercury , I believe you can merge this PR, or I will go ahead if you prefer!

ferdymercury:
Thanks Vincenzo! I have no merge rights, so please go ahead;) an remember squash on merge.

PR #17888, opened 2025-03-05
[CMake] Compare to the right CMake version for `-j 1` workaround

guitargeek:
Follows up on e3fbf24ea1ed8c8, fixing a typo in a number that was committed there and not detected in the PR because of no full rebuild.


bellenot (approved):
Thanks Jonas!

PR #17031, opened 2024-11-25
[tmva] fix crash in addEvent if no addVariable was called before

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-10018

## Checklist:

- [x] tested changes locally
- [ ] updated the docs (if necessary)

PR #17894, opened 2025-03-05
[tmva] fix spacing in error message and add test

ferdymercury:
# This Pull request:

## Changes or fixes:

Tests https://its.cern.ch/jira/browse/ROOT-9833
 
That issue was inadvertently fixed by https://github.com/root-project/root/pull/17031

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tmva/tmva/src/DataLoader.cxx
+++ b/tmva/tmva/src/DataLoader.cxx
@@ -299,7 +299,8 @@ void TMVA::DataLoader::AddEvent( const TString& className, Types::ETreeType tt,
    fATreeType   = clIndex;
    fATreeWeight = weight;
    if (event.size() > fATreeEvent.size()) {
-      Error("TMVA::DataLoader::AddEvent", "Number of variables defined through DataLoader::AddVariable (%zu) is inconsistent"
+      Error("AddEvent",
+       "Number of variables defined through DataLoader::AddVariable (%zu) is inconsistent"
        " with number of variables given to DataLoader::Add*Event (%zu)."
        " Please check your variable definitions and statement ordering."
        " This event will not be added.", fATreeEvent.size(), event.size());
--- a/tmva/tmva/src/DataLoader.cxx
+++ b/tmva/tmva/src/DataLoader.cxx
@@ -7,7 +7,7 @@
  * Project: TMVA - a Root-integrated toolkit for multivariate data analysis       *
  * Package: TMVA                                                                  *
  * Class  : DataLoader                                                            *
- *                                             *
+ *                                                                                *
  *                                                                                *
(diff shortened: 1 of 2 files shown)


pcanal (approved):

pcanal:
> Fixes [its.cern.ch/jira/browse/ROOT-9833](https://its.cern.ch/jira/browse/ROOT-9833)

How does this PR fixes the issue?

ferdymercury:
> How does this PR fixes the issue?

That issue was inadvertently fixed by https://github.com/root-project/root/pull/17031
So the JIRA Issue could be already closed.

This PR only adds a test for it and fixes some formatting.

pcanal (approved):

ferdymercury:
@pcanal I am not sure how to properly catch the raised Error. Is try-catch sufficient? Or do I need to register an error handler?
Thks

ferdymercury:
The catching of the error seems fixed now.

pcanal:
> I am not sure how to properly catch the raised Error. I

Which error?

ferdymercury:
> > I am not sure how to properly catch the raised Error. I
> 
> Which error?

Adding

> > I am not sure how to properly catch the raised Error. I
> 
> Which error?

The one raised now by the function instead of crashing, see below. But I managed to sort it out using a handler.

```
root [0] std::vector<double> evData = {1, 2};
root [1]    TMVA::DataLoader dl;
root [2] dl.AddEvent("class1", TMVA::Types::kTraining, evData, 1.0)
DataSetInfo              : [default] : Added class "class1"
Error in <TMVA::DataLoader::TMVA::DataLoader::AddEvent>: Number of variables defined through DataLoader::AddVariable (0) is inconsistentwith number of variables given to DataLoader::Add*Event (2).Please check your variable definitions and statement ordering.This event will not be added.
```

pcanal:
> But I managed to sort it out using a handler.

Why was capturing `stderr` not sufficient?

pcanal:
This is because the test already have their custom error handler and already have a mechanism to check the values.  So we might need here something like:
```
   const auto errmsg = "file %s/doesnotexist.root does not exist";
   TString expecteddiag;
   expecteddiag.Form(errmsg, gSystem->pwd());
   ROOT::TestSupport::CheckDiagsRAII diagRAII{kError, "TFile::TFile", expecteddiag.Data()};
```

ferdymercury:
> This is because the test already have their custom error handler and already have a mechanism to check the values. So we might need here something like:

Thanks a lot for the help. Changed now as suggested.

pcanal (approved):

ferdymercury:
Thanks a lot! Please close the associated JIRA issue.

PR #17898, opened 2025-03-05
Remove unused code about autoloading from LinkDef files

guitargeek:
The `CLING_USE_AUTLOADING` is not defined anywhere, and hence the code is presumably not needed.

### DIFF
--- a/roofit/roofit/inc/LinkDef1.h
+++ b/roofit/roofit/inc/LinkDef1.h
@@ -131,10 +131,6 @@
 #pragma link C++ class RooCFunction4Binding<double,double,double,double,bool>+ ;
 #pragma link C++ class RooCFunction4PdfBinding<double,double,double,double,bool>+ ;
 
-//#ifdef USE_FOR_AUTLOADING
-//#pragma link C++ class RooFit ;
-//#else
 #pragma link C++ namespace RooFit ;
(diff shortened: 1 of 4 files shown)


roofit/roofit/inc/LinkDef1.h
```diff
@@ -131,10 +131,4 @@
 #pragma link C++ class RooCFunction4Binding<double,double,double,double,bool>+ ;
 #pragma link C++ class RooCFunction4PdfBinding<double,double,double,double,bool>+ ;
 
-//#ifdef USE_FOR_AUTLOADING
-//#pragma link C++ class RooFit ;
-//#else
-#pragma link C++ namespace RooFit ;
```
pcanal, line 137: This was the only active one.  It may still be useful for build without the `runtime_cxxmodule`.

dpiparo: the modules-off build worked. There are other failures though.

guitargeek: Ah thanks for noticing this oversight

guitargeek:
@pcanal, I have addressed your comment. Can you make review again? Thanks!

pcanal (approved):
Thanks.

PR #14105, opened 2023-11-24
[hist] Improve calculation of `TAxis::FindBin(x)` when x is at the bin edges

lmoneta:
When x is at the bin edge values numerical error can cause the computation of the bin to be the one before (or after). Correct for this case assuming the bin edge are computed as in `TH1::getBinLowEdge` and `TH1::GetBinUpEdge`. It is clear that the bin edge values are represented as floating point, so depending on how they are computed they can be sometime different. However, it is better to have teh consistency to return the correct value when computed as internally in TAxis.

This should fix teh problem reported in https://root-forum.cern.ch/t/bug-in-taxis-findbin/57210 and fixes #14091

PR #17899, opened 2025-03-06
[skip-ci] clarify behavior and rounding precision limitations of FindBin

ferdymercury:
# This Pull request:

## Changes or fixes:

See https://root-forum.cern.ch/t/bug-or-feature-in-ttree-draw/62862/16

Related: https://github.com/root-project/root/pull/17689

### DIFF
--- a/hist/hist/src/TAxis.cxx
+++ b/hist/hist/src/TAxis.cxx
@@ -280,10 +280,18 @@ void TAxis::ExecuteEvent(Int_t event, Int_t px, Int_t py)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-/// Find bin number corresponding to abscissa x. NOTE: this method does not work with alphanumeric bins !!!
+/// Find bin number corresponding to abscissa `x`. NOTE: this method does not work with alphanumeric bins !!!
 ///
-/// If x is underflow or overflow, attempt to extend the axis if TAxis::kCanExtend is true.
-/// Otherwise, return 0 or fNbins+1.
+/// If `x` is underflow or overflow, attempt to extend the axis if TAxis::kCanExtend is true.
+/// Otherwise, return `0` or `fNbins+1`.
+///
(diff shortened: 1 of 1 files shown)


guitargeek:
Does this maybe interfere with #14105, where I have made some suggestions to improve the numerical stability of `FindBin()`? That PR is kind of stuck, because then we would also have to improve other functions like `GetBinLowEdge` and `GetBinHighEdge` to be consistent. And changing the numeric behavior is always a risk for backwards compatibility, even if it goes in the right direction :|

We need to check whether the example `TAxis(1, -1., 0.).FindBin(-1e-17)` still wrongly returns the overflow bin even with the other PR merged. Thank we can merge this PR without having to worry that it gets invalidated by #14105.

ferdymercury:
Thanks for the feedback!

> We need to check whether the example `TAxis(1, -1., 0.).FindBin(-1e-17)` still wrongly returns the overflow bin even with the other PR merged.

I checked with that PR you have mentioned, and it still returns the wrong bin.
So it does not really interfere with the other pull request. This PR would be just documenting the current behaviour.

See reproducer with the other PR:

```
      double x = -1e-17;
      const double fXmin = -1000.;
      const double fXmax = 0.;
      const int fNbins = 1000;
      // shift x and fXmax by fXmin:
      x = x - fXmin;
      const double b = fXmax - fXmin;
      const double s = fNbins * x; // scaled version of x
      double m = std::floor(s / b);
      // iterative correction in case of floating point errors due to floating point division
      while (m * b > s) // if left bin boundary is greater then x, decrement
         m = m - 1;
      while ((m + 1) * b <= s) //if right bin boundary is smaller or equal, increment
         m = m + 1;
      cout << 1 + Int_t(m) << endl;
```
Same issue if you cast everything to long-double.

guitargeek (approved):
Thanks for the PR and following up to my comment! The PR looks good to me.

PR #17902, opened 2025-03-06
[ntuple] Use actual unaligned ntuples in test for chained joins

enirolf:
We previously only pretended the join was unaligned by providing a join key. To further strengthen the test, now a join between two actual unaligned ntuples is used.


vepadulano (approved):
Very nice test case, thanks!

PR #17904, opened 2025-03-06
[Docs] TTree example is missing the critical lines

### LINKED ISSUE #17864: [Docs] TTree example is missing the critical lines
### Check duplicate issues.

- [x] Checked for duplicates

### Description

In the TTree documentation, the section "A complete example" shows a code example. Due to a doxygen parsing bug, the critical lines after
```
// Create a ROOT Tree
```
are missing, though.

### Reproducer

Look at https://root.cern.ch/doc/master/classTTree.html

### ROOT version

Master

### Installation method

Doesn't apply

### Operating system

Doesn't apply

### Additional context

_No response_

Aditya-138-12:
# This Pull request:
1. Modified the example to be in multiple code blocks
2. Added some comments

## Changes or fixes:
Doxygen version - 1.10.0 (ec....)
Screenshot of the changes
Screenshot from 2025-03-06 17-35-13

## Checklist:

- [x] tested changes locally

This PR fixes #17864

### DIFF
--- a/tree/tree/src/TTree.cxx
+++ b/tree/tree/src/TTree.cxx
@@ -270,6 +270,7 @@ For this reasons, ROOT offers the concept of friends for TTree and TChain.
 \anchor fullexample
 ## A Complete Example
 
+Start by creating a file and histogram:
 ~~~ {.cpp}
 // A simple example creating a tree
 // Compile it with: `g++ myTreeExample.cpp -o myTreeExample `root-config --cflags --libs`
@@ -288,7 +289,11 @@ int main()
 
    // Define a histogram and some simple structures
    TH1D hpx("hpx", "This is the px distribution", 100, -4, 4);
-
+   
+   ~~~
+   
+   Next, define the data structures:
+   ~~~ {.cpp}
    typedef struct {
       Float_t x, y, z;
    } Point;
@@ -300,13 +305,18 @@ int main()
    } Event;
    Point point;
    Event event;
+   
+   ~~~
+   
+   Finally, create  a tree, and branches, and fill it with data:
+   ~~~ {.cpp}
 
    // Create a ROOT Tree
    TTree tree("T", "An example of ROOT tree with a few branches");
    tree.Branch("point", &point, "x:y:z");
    tree.Branch("event", &event, "ntrack/I:nseg:nvertex:flag/i:temperature/F");
    tree.Branch("hpx", &hpx);
-
+   
    float px, py;
 
    TRandom3 myGenerator;


vepadulano:
Dear @Aditya-138-12 ,

Thank you for your help with addressing this issue! It's good to know that we can split the code of the example in multiple blocks to make all of them visible on their own!

On the other hand, the paragraph title is `A complete example`, and I believe it is actually nice to have a single code block to show that can be even copy-pasted as a tutorial. In the end, the absence of parts of this code block from the rendered documentation on the website looks more like a bug with Doxygen (or with our usage of it), so probably we want to fix it somewhere higher in the call chain rather than providing a temporary solution which might actually hide the real problem.

I would propose to close this PR, pending a more complete solution, I will also let @couet comment since it's his area of expertise.

ferdymercury:
Thanks a lot for trying this! I agree that we should try to find out what's happening with doxygen rather than finding a workaround. It might affect other parts of the codebase, too, so it is not worth to fix them one by one. As a suggestion to @Aditya-138-12 try to rebuild the documentation with doxygen 1.13.2 (clone git repo and build from source) to see if the problem is still there. If yes, try to create a minimum example reproducer with just that file in INPUT, and post it in the Doxygen GitHub repo Issues.

ferdymercury:
Maybe related: https://github.com/doxygen/doxygen/issues/8569 https://github.com/doxygen/doxygen/pull/8578

Aditya-138-12:
> Thanks a lot for trying this! I agree that we should try to find out what's happening with doxygen rather than finding a workaround. It might affect other parts of the codebase, too, so it is not worth to fix them one by one. As a suggestion to @Aditya-138-12 try to rebuild the documentation with doxygen 1.13.2 (clone git repo and build from source) to see if the problem is still there. If yes, try to create a minimum example reproducer with just that file in INPUT, and post it in the Doxygen GitHub repo Issues.

Ahh, I used the older versions of the doxygen, I will once try out with `1.13.2`,

Aditya-138-12:
@ferdymercury @vepadulano @couet 
After generating the html file with doxygen version `1.13.2`, i was able to see the  lines after `Create a ROOT tree`, 
so maybe it's somewhere high in the call chain as @vepadulano has suggested.
Thank you for your time!
Screenshot from 2025-03-12 02-25-36

ferdymercury:
Thanks to you for finding out. I would thus suggest @couet to upgrade the doxygen version in the CI machine and hopefully that fixes the issue without any modification in the code.

couet:
Sorry if I missed something but I compared line by line the doxygen page https://root.cern/doc/master/classTTree.html#fullexample with the source code. And I do not see any missing lines in the example in the doxygen documentation.

ferdymercury:
> And I do not see any missing lines in the example in the doxygen documentation.

It was missing during a couple of days (I checked), but it's true that now it seems to work again. So maybe it was an interplay with the failing TMVA tutorial or so and now it's restored.

couet:
Indeed it does not only work "again", it works "better" because before the link to "tree" in the example was pointing to a completely wrong TMVA link https://root.cern/doc/v632/classTTree.html#fullexample

PR #17895, opened 2025-03-05
[WIP] Add RDataSource for TTree

vepadulano:
All tests are working on my machine with these changes, so I want to start testing on all platforms. Changes will be later split in multiple commits/PRs.

PR #17905, opened 2025-03-06
[NFC] Do not rely on string representation in test

vepadulano:
The string representation of the dataframe might change and it is not the purpose of this test anyway.

Part 2 of N for https://github.com/root-project/root/pull/17895


pcanal (approved):
LGTM (albeit it is a functional change :) ).

PR #17909, opened 2025-03-06
Prevent LLVM cmake from finding builtin zstd. 

pcanal:
* Fix removal of ROOTSYS from CMAKE_PREFIX_PATH.
Current code was, if CMAKE_PREFIX_PATH had more than one element, essentially
inadvertently replacing ROOTSYS by the current path. (eg. if `CMAKE_PREFIX_PATH`
contains just `:` or `::` or `:somethingelse`, it will add the current directory (top level
of the build directory to the prefix path).

* Fix find_package disabling for zstd.
LLVM uses the lower case spelling for 'zstd' and thus find_package was still being called.

In practice this lead my build: `-Dbuiltin_zstd=ON -Dbuiltin_xxhash=ON -Dminimal=ON -Dbuiltin_zlib=ON` on `AlmaLinux release 9.5 (Teal Serval)` to pick up the `lib/libstd.a` for the LLVM libraries and fails to link `rootcling_stage1` because of missing `xxHash` symbols.

This PR also fixes: [its.cern.ch/jira/browse/ROOT-10537](https://its.cern.ch/jira/browse/ROOT-10537)

This is somewhat related to #16285 
and https://github.com/root-project/root/issues/8633


ferdymercury:
Somewhat related: https://its.cern.ch/jira/browse/ROOT-10537

pcanal:
You are right.  I will fix that issue as part of this PR.

pcanal:
Sigh ... But sometimes we seem to want to use the builtin libzstd:
```
  input_line_1:1:2: fatal error: malformed or corrupted AST file: 'LLVM was not built with LLVM_ENABLE_ZSTD or did not find zstd at build time'
  #include <new>
   ^
  input_line_1:1:2: note: after modifying system headers, please delete the module cache at '/github/home/ROOT-CI/build/lib'
```
i.e. either libzstd is not installed locally or it is no longer found.

hageboeck:
The cleanup of the prefix path looks reasonable.

Just one clarification regarding the pcms:
> so it is a bit too bad that we can't have the builtins used by LLVM. (and the builds are fixed by deleting the existing pcm files)

When LLVM wasn't configured with compression libraries, this still results in functioning pcms, just uncompressed? In that case I think we can go ahead.

CMakeLists.txt
```diff
@@ -58,14 +58,22 @@ set(DIR_PERMISSIONS DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE G
...
-  string(REPLACE "$ENV{ROOTSYS}/lib" "" ENV_PYTHONPATH "$ENV{PYTHONPATH}")
-  string(REPLACE "$ENV{ROOTSYS}" "" ENV_CMAKE_PREFIX_PATH "$ENV{CMAKE_PREFIX_PATH}")
-  set(ENV{PATH} "${ENV_PATH}")
-  set(ENV{LD_LIBRARY_PATH} "${ENV_LD_LIBRARY_PATH}")
-  set(ENV{PYTHONPATH} "${ENV_PYTHONPATH}")
-  set(ENV{CMAKE_PREFIX_PATH} "${ENV_CMAKE_PREFIX_PATH}")
+  function(strip_path ENV_VAR _path)
+    string(REPLACE ":${_path}:" ":" _temp_envvar "${_temp_envvar}")
+    string(REGEX REPLACE "^${_path}:" "" _temp_envvar "$ENV{${ENV_VAR}}")
+    string(REGEX REPLACE ":${_path}$" "" _temp_envvar "${_temp_envvar}")
```
hageboeck, line 64: It looks like the lines were reordered, so the last argument needs to change.
And after the first line dealt with `:A:`, you should get away with a single regex:
```suggestion
    string(REPLACE ":${_path}:" ":" _temp_envvar "$ENV{${ENV_VAR}}")
    string(REGEX REPLACE ":?${_path}:?" "" _temp_envvar "${_temp_envvar}")
```

CMakeLists.txt
```diff
@@ -58,14 +58,22 @@ set(DIR_PERMISSIONS DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE G
...
+    if ("${_temp_envvar}" STREQUAL "${_path}")
+      set(ENV{ENV_VAR} "")
```
hageboeck, line 66: This should be unnecessary if the above suggestion works as I think it works. The replacement should take care of all of
- `A:`
- `:A`
- `A`

CMakeLists.txt
```diff
@@ -58,14 +58,22 @@ set(DIR_PERMISSIONS DIRECTORY_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE G
...
+    else()
+      set(ENV{ENV_VAR} "${_temp_envvar}")
+    endif()
+  endfunction()
+  strip_path("PATH" "$ENV{ROOTSYS}/bin")
+  strip_path("LD_LIBRARY_PATH" "$ENV{ROOTSYS}/lib")
+  strip_path("PYTHONPATH" "$ENV{ROOTSYS}/lib")
+  # if we leave the ':' it will result in an empty entry in the CMAKE_PREFIX_PATH
+  # which will interpreted as the current directory.
```
hageboeck, line 75: This comment seems outdated now.

hageboeck (changes requested):
> I updated to consistent apply the fix to all 3 variable 

Great, I was going to ask that! 🙂
I think I found a mistake (the replacements should start from the original ENV, not from the temporary variable, which is always empty.
And using a larger regex, one can probably save a few lines.

CMakeLists.txt
pcanal: The comment is maybe misplaced but still relevant.  For the other paths, there is no side effect to having `::` whereas it is disastrous for `CMAKE_PREFIX_PATH`.

CMakeLists.txt
pcanal: > It looks like the lines were reordered, so the last argument needs to change.

Indeed, exactly what happeed :(

pcanal: > And after the first line dealt with :A:, you should get away with a single regex:
```
string(REGEX REPLACE ":?${_path}:?" "" _temp_envvar "${_temp_envvar}")
```
I think they would still wrong partially replace
```
/some/other/path/${_path}
${_path}64
...:${_path}64:...
etc.
```

hageboeck: I see. If `A` is not a full path, that could indeed create a problem

hageboeck (approved):
LGTM!

pcanal:
The failure is unrelated (something about numba).

PR #17914, opened 2025-03-07
[ntuple] Move Internal RNTupleUtils out of Experimental

silverweed:
Moving the following out of Experimental:

- `MakeAliasedSharedPtr`
- `MakeUninitArray`
- `EnsureValidNameForRNTuple`
- Some internal test values (e.g. `kTestFutureType`)

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
@@ -297,6 +297,12 @@ public:
    std::uint32_t GetTag() const { return fTag; }
 };
 
+} // namespace Internal
+
+} // namespace Experimental
+
+namespace Internal {
+
 template <typename T>
 auto MakeAliasedSharedPtr(T *rawPtr)
 {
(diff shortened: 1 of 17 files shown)


tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
```diff
@@ -328,6 +334,8 @@ RResult<void> EnsureValidNameForRNTuple(std::string_view name, std::string_view
 
 } // namespace Internal
 
+namespace Experimental {
```
pcanal, line 337:
```suggestion
namespace Experimental {

namespace Internal {
```
isn't it? (to be consistent with previous version)

pcanal (approved):

tree/ntuple/v7/inc/ROOT/RNTupleUtil.hxx
silverweed: The `EColumnType` typedef was not in `Internal`, it should be correct as-is

hahnjo (approved):
LGTM. We may want to (independently) rename `kTestFutureType` to `kTestFutureColumnType`...

PR #17915, opened 2025-03-07
[ntuple] Initialize variable in ntuple_zip test

hahnjo:
GCC complains (with -O0):
```
warning: ‘x’ may be used uninitialized [-Wmaybe-uninitialized]
```

### DIFF
--- a/tree/ntuple/v7/test/ntuple_zip.cxx
+++ b/tree/ntuple/v7/test/ntuple_zip.cxx
@@ -13,7 +13,7 @@ TEST(RNTupleZip, Basics)
 
 TEST(RNTupleZip, Empty)
 {
-   char x;
+   char x = 0;
    char z;
    EXPECT_EQ(0U, RNTupleCompressor::Zip(&x, 0, 0, &z));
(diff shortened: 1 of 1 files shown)


tree/ntuple/v7/test/ntuple_zip.cxx
```diff
@@ -13,7 +13,7 @@ TEST(RNTupleZip, Basics)
 
 TEST(RNTupleZip, Empty)
 {
-   char x;
+   char x = 0;
    char z;
```
silverweed, line 17: So it complains for `x` but not for `z`? Interesting...

silverweed (approved):

tree/ntuple/v7/test/ntuple_zip.cxx
hahnjo: Yes, I think because the first parameter is a `const` pointer, so you are not allowed to write. The only reason to pass in a `const` pointer is to read from it (which we don't do here because the size is 0).

enirolf (approved):

PR #17919, opened 2025-03-07
[ntuple] Add partioning to `RNTupleJoinTable`

enirolf:
With this addition, the `RNTupleJoinTable` can be split up into several mappings from join values to entry numbers, according to some (numeric) partition key. Using a partition key is optional; a default partition key is used when none has been specified by the user.

The `RNTupleJoinTable` now has an inner class, `REntryMapping`, which in practice now represents what the `RNTupleJoinTable` represented previously, i.e., a mapping from join field values to entry numbers:
```
entryMapping
|- {x_1, y_1} -> 42
|- {x_2, y_2} -> 99
|- {x_3, y_3} -> 12
|- ...
```
The join table itself now instead provides a mapping from partition keys to (a collection of) these entry mappings:
```
joinTable
|- 0
   |- entryMapping_1
|- 4
   |- entryMapping_0
   |- entryMapping_1
|- ...
|- kDefaultPartitionKey
   |- entryMapping_0
```

This has several use cases, but the most immediate one is that with this approach, the `RNTupleJoinTable` itself is not restricted to one page source anymore (this restriction is now in `REntryMapping` instead. This is useful for the integration into the `RNTupleProcessor`, where we want to be able to create joins of chains of RNTuples and have to deal with more than one page source.

### TODO
- [ ] Test and properly handle passing non-present partition keys to `GetEntryIndexes` and `GetMappingsForKey`.


tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -35,63 +36,113 @@ namespace Internal {
...
+      //////////////////////////////////////////////////////////////////////////
+      /// Container for the combined hashes of the join fields.
+      class RCombinedJoinFieldValue {
+      public:
+         std::vector<NTupleJoinFieldValue_t> fFieldValues;
+         RCombinedJoinFieldValue(const std::vector<NTupleJoinFieldValue_t> &fieldValues)
+         {
+            fFieldValues.reserve(fieldValues.size());
+            fFieldValues = fieldValues;
+         }
```
vepadulano, line 63: Shouldn't this be equivalent?
```suggestion
         RCombinedJoinFieldValue(const std::vector<NTupleJoinFieldValue_t> &fieldValues): fFieldValues(fieldValues)
         {
         }
```

tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -35,63 +36,113 @@ namespace Internal {
...
+   */
+   // clang-format on
+   class REntryMapping {
+   private:
+      using NTupleJoinFieldValue_t = std::uint64_t;
+
+      //////////////////////////////////////////////////////////////////////////
+      /// Container for the combined hashes of the join fields.
+      class RCombinedJoinFieldValue {
+      public:
```
vepadulano, line 57: Maybe `struct` better represents the class layout in this case?

tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -35,63 +36,113 @@ namespace Internal {
...
+         }
+      };
 
-   /////////////////////////////////////////////////////////////////////////////
-   /// Hash combining the individual join field value hashes from RCombinedJoinFieldValue. Uses the implementation from
-   /// `boost::hash_combine` (see
-   /// https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine).
-   struct RCombinedJoinFieldValueHash {
-      inline std::size_t operator()(const RCombinedJoinFieldValue &joinFieldValue) const
+      /// The mapping itself. Maps field values (or combinations thereof in case the join key is a composed of multiple
```
vepadulano, line 85:
```suggestion
      /// The mapping itself. Maps field values (or combinations thereof in case the join key is composed of multiple
```

tree/ntuple/v7/test/ntuple_join_table.cxx
```diff
@@ -246,3 +242,80 @@ TEST(RNTupleJoinTable, MultipleMatches)
...
+      // Without explicitly providing the partition key we expect multiple resulting entry indexes
+      auto entryIdxs = joinTable->GetEntryIndexes({i.get()});
+      EXPECT_EQ(entryIdxs.size(), 4ul);
+
+      entryIdxs = joinTable->GetEntryIndexes({i.get()}, *run);
+
+      switch (*run) {
+      // For run == 0, no entry indexes are expected because they were added using the default partition key
+      case 0: EXPECT_EQ(entryIdxs.size(), 0ul); break;
+      // For run == 0, we expect exactly one entry index
```
vepadulano, line 311:
```suggestion
      // For run == 1, we expect exactly one entry index
```

vepadulano (changes requested):
Nice improvement! Some minor comments for now

tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -35,63 +36,114 @@ namespace Internal {
...
+   */
+   // clang-format on
+   class REntryMapping {
+   private:
+      using NTupleJoinFieldValue_t = std::uint64_t;
+
+      //////////////////////////////////////////////////////////////////////////
+      /// Container for the combined hashes of join field values.
+      struct RCombinedJoinFieldValue {
+      public:
```
silverweed, line 57: This `public` is superfluous

tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -35,63 +36,114 @@ namespace Internal {
...
-   RNTupleJoinTable(const std::vector<std::string> &fieldNames) : fJoinFieldNames(fieldNames) {}
+   RNTupleJoinTable(const std::vector<std::string> &joinFieldNames) : fJoinFieldNames(joinFieldNames) {}
 
    /////////////////////////////////////////////////////////////////////////////
-   /// \brief Ensure the RNTupleJoinTable has been built.
-   ///
-   /// \throws RException If the join table has not been built, and can therefore not be used yet.
-   void EnsureBuilt() const;
+   /// \brief Get all entry mappings in the join table.
+   const std::vector<REntryMapping *> GetAllMappings() const;
```
silverweed, line 142: Since you're returning a value, it's better to return it non-const (same thing for the next function)

silverweed (commented):
A couple of minor comments

tree/ntuple/v7/src/RNTupleJoinTable.cxx
```diff
@@ -78,51 +61,140 @@ void ROOT::Experimental::Internal::RNTupleJoinTable::Build(RPageSource &pageSour
...
+}
+
+void ROOT::Experimental::Internal::RNTupleJoinTable::Build()
+{
+   if (fIsBuilt)
+      return;
+
+   for (auto &mapping : GetAllMappings()) {
+      mapping->Build();
+   }
```
hahnjo, line 177: Given that `GetAllMappings` creates a copy of the pointers stored in `fPartitions`, should this just iterate over it directly?
```suggestion
   for (const auto &partition : fPartitions) {
      for (const auto &joinMapping : partition) {
         joinMapping->Build();
      }
   }
```

tree/ntuple/v7/test/ntuple_join_table.cxx
```diff
@@ -246,3 +242,74 @@ TEST(RNTupleJoinTable, MultipleMatches)
    entryIdxs = joinTable->GetEntryIndexes({&(++run)});
    EXPECT_EQ(std::vector<ROOT::NTupleSize_t>{}, entryIdxs);
 }
+
+TEST(RNTupleJoinTable, Partitions)
+{
+   auto fnWriteNTuple = [](const FileRaii &fileGuard, std::uint32_t run) {
```
hahnjo, line 248:
```suggestion
   auto fnWriteNTuple = [](const FileRaii &fileGuard, std::int16_t run) {
```
to make the field type below?

tree/ntuple/v7/test/ntuple_join_table.cxx
```diff
@@ -246,3 +242,74 @@ TEST(RNTupleJoinTable, MultipleMatches)
...
+      auto model = RNTupleModel::Create();
+      *model->MakeField<std::int16_t>("run") = run;
```
hahnjo, line 250: or rather
```suggestion
      *model->MakeField<std::uint32_t>("run") = run;
```
here?

tree/ntuple/v7/test/ntuple_join_table.cxx
```diff
@@ -246,3 +242,74 @@ TEST(RNTupleJoinTable, MultipleMatches)
...
+   std::vector<FileRaii> fileGuards;
+   fileGuards.emplace_back("test_ntuple_join_partition1.root");
+   fileGuards.emplace_back("test_ntuple_join_partition2.root");
+   fileGuards.emplace_back("test_ntuple_join_partition3.root");
+   fileGuards.emplace_back("test_ntuple_join_partition4.root");
+
+   std::int16_t runNumbers[4] = {1, 2, 3, 3};
+   std::vector<std::unique_ptr<RPageSource>> pageSources;
+
+   // Create three ntuples where with their corresponding run numbers and add them to the join table using this run
```
hahnjo, line 272:
```suggestion
   // Create four ntuples where with their corresponding run numbers and add them to the join table using this run
```

tree/ntuple/v7/test/ntuple_join_table.cxx
```diff
@@ -246,3 +242,74 @@ TEST(RNTupleJoinTable, MultipleMatches)
...
+   auto run = proc->GetEntry().GetPtr<int16_t>("run");
+
+   // Calling GetEntryIndexes with a partition key not present in the join table shouldn't fail; it should just return
+   // an empty vector.
+   EXPECT_EQ(std::vector<ROOT::NTupleSize_t>{}, joinTable->GetEntryIndexes({i.get()}, 4));
+
+   *i = 99;
+   // Similarly, calling GetEntryIndexes with a partition key that is present in the join table but a join value that
+   // isn't shouldn't fail; it should just return an empty vector.
+   EXPECT_EQ(std::vector<ROOT::NTupleSize_t>{}, joinTable->GetEntryIndexes({i.get()}, 4));
```
hahnjo, line 298: But 4 isn't present, see the comment for the previous check. Did you mean to write
```suggestion
   *i = 99;
   // Similarly, calling GetEntryIndexes with a partition key that is present in the join table but a join value that
   // isn't shouldn't fail; it should just return an empty vector.
   EXPECT_EQ(std::vector<ROOT::NTupleSize_t>{}, joinTable->GetEntryIndexes({i.get()}, 3));
```
?

tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -138,14 +191,46 @@ public:
...
    ///
-   /// \return The entry indexes that correspond to `valuePtrs`. An empty vector is returned when there are no matching
-   /// indexes.
-   std::vector<ROOT::NTupleSize_t> GetEntryIndexes(const std::vector<void *> &valuePtrs) const;
+   /// \return The entry numbers that correspond to `valuePtrs`. When there are no corresponding entries, an empty
+   /// vector is returned.
+   std::vector<ROOT::NTupleSize_t> GetEntryIndexes(const std::vector<void *> &valuePtrs) const
+   {
+      return GetEntryIndexes(valuePtrs, fPartitionKeys);
+   }
```
hahnjo, line 227: Is this useful in case of multiple partitions since the return value loses the information which mapping the indexes are relative to? Maybe what we want instead is `GetEntryIndexes(const std::vector<void *> &valuePtrs, PartitionKey_t partitionKey = kDefaultPartitionKey)`?

tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -35,63 +36,113 @@ namespace Internal {
...
    std::vector<std::string> fJoinFieldNames;
 
-   /// The size (in bytes) for each join field, corresponding to `fJoinFieldNames`. This information is stored to be
-   /// able to properly cast incoming void pointers to the join field values in `GetEntryIndexes`.
-   std::vector<std::size_t> fJoinFieldValueSizes;
+   /// Partition keys used in the join table.
+   std::vector<PartitionKey_t> fPartitionKeys;
+
+   /// Partitions of one or multiple entry mappings.
+   std::vector<std::vector<std::unique_ptr<REntryMapping>>> fPartitions;
```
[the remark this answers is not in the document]
enirolf: Done (and updated `Build` and `GetEntryIndexes` to iterate directly over the partitions) . This change has made `GetAllMappings` and `GetMappingsForKey` redundant; I've removed them.

tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -103,32 +154,28 @@ public:
...
-   /// \param[in] pageSource The page source of the RNTuple for which to build the join table.
    ///
-   /// Only a built join table can be queried (with RNTupleJoinTable::GetEntryIndexes).
-   void Build(RPageSource &pageSource);
+   /// \param[in] pageSource The page source of the RNTuple with the entries to map.
+   /// \param[in] partitionKey Which partition to add the mapping to. If not provided, it will be added to the default
+   /// partition.
+   ///
+   /// \return A reference to the updated join table.
+   RNTupleJoinTable &Add(RPageSource &pageSource, PartitionKey_t partitionKey = kDefaultPartitionKey);
```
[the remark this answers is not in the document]
hahnjo: Two more options:
4. (variation of 2.) continue to have a single `RNTupleJoinTable::Build`, but pass it all page sources at once. This still requires the `RNTupleJoinTable` to figure out which entry mapping a page source belonged to - not sure I see how to do this.
5. (variation of 3.) have `RNTupleJoinTable::Add` also build the mapping, but lazily construct the entire join table in `RNTupleJoinProcessor`. This may also simplify the code a bit since a join table is always built, but it's still deferred from the user perspective.

I currently favor 5. but that's because I thought of it last, so I probably haven't discovered its shortcomings yet :smiley:

tree/ntuple/v7/src/RNTupleJoinTable.cxx
```diff
@@ -78,51 +61,115 @@ void ROOT::Experimental::Internal::RNTupleJoinTable::Build(RPageSource &pageSour
...
+std::unordered_map<ROOT::Experimental::Internal::RNTupleJoinTable::PartitionKey_t, std::vector<ROOT::NTupleSize_t>>
+ROOT::Experimental::Internal::RNTupleJoinTable::GetPartitionedEntryIndexes(
+   const std::vector<void *> &valuePtrs, const std::vector<PartitionKey_t> &partitionKeys) const
+{
+   std::unordered_map<PartitionKey_t, std::vector<ROOT::NTupleSize_t>> entryIdxs{};
+
+   for (const auto &partitionKey : partitionKeys) {
+      auto entriesForPartition = GetEntryIndexes(valuePtrs, partitionKey);
+      entryIdxs[partitionKey].insert(entryIdxs[partitionKey].end(), entriesForPartition.begin(),
+                                     entriesForPartition.end());
```
hahnjo, line 154: Should this check
```suggestion
      auto entriesForPartition = GetEntryIndexes(valuePtrs, partitionKey);
      if (!entriesForPartition.empty())
         entryIdxs[partitionKey].insert(entryIdxs[partitionKey].end(), entriesForPartition.begin(),
                                        entriesForPartition.end());
```
? The question is what do we want to return if there is no index in a particular partition: an empty `vector` or omit the key from the map. The other overload seems to do the latter, and that's also what I would semantically expect. (we should probably add a test for this overload, if I'm not mistaken)

hahnjo (commented):
LGTM in principle, thanks for making all the changes. I have one final question about the semantics of `GetPartitionedEntryIndexes`, and what it should return in case there is no index in a partition.

tree/ntuple/v7/src/RNTupleJoinTable.cxx
enirolf: Oh yeah, thanks for catching this! I think for now indeed it's fine to use the latter behavior and let time and future developments tell whether whether this was the right intuition. I've expanded the tests as well.

vepadulano (approved):
Nice improvements, thanks!

hahnjo (approved):

PR #17921, opened 2025-03-07
[Python] Reduce time before asyncronous run in JupyROOT unit test

guitargeek:
Only waiting 0.25 seconds before printing "Asyncronous" should increase the probability to have this string printed by the time the main thread waits one second, which is what the unit test expects.

Hopefully avoids test failures like these:
https://github.com/root-project/root/actions/runs/13712379279/job/38351177011#step:13:22736


dpiparo (commented):
lgtm and thanks.

dpiparo (approved):

PR #17927, opened 2025-03-10
Improve semantics of data member

vepadulano:
Change the data member holding the column names which should be ignored in case of TTreeReader errors regarding missing branches, from std::vector to std::set. The column names should not be repeated to avoid having multiple times the same column name left in the data member, which results in leftover names even after the corresponding operation of the computation graph has finished. Once a column is found, it will be available for all nodes of the computation graph anyway.

Part 4 of N for https://github.com/root-project/root/pull/17895

### DIFF
--- a/tree/dataframe/inc/ROOT/RDF/RLoopManager.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/RLoopManager.hxx
@@ -196,7 +196,7 @@ class RLoopManager : public RNodeBase {
    // List of branches for which we want to suppress the printed error about
    // missing branch when switching to a new tree. This is modified by readers,
    // so must be declared before them in this class.
-   std::vector<std::string> fSuppressErrorsForMissingBranches{};
+   std::set<std::string> fSuppressErrorsForMissingBranches{};
    ROOT::Internal::RDF::RStringCache fCachedColNames;
    std::set<std::pair<std::string_view, std::unique_ptr<ROOT::Internal::RDF::RDefinesWithReaders>>>
       fUniqueDefinesWithReaders;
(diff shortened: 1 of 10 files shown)


tree/dataframe/inc/ROOT/RDF/RLoopManager.hxx
```diff
@@ -196,7 +196,7 @@ class RLoopManager : public RNodeBase {
    // List of branches for which we want to suppress the printed error about
    // missing branch when switching to a new tree. This is modified by readers,
    // so must be declared before them in this class.
-   std::vector<std::string> fSuppressErrorsForMissingBranches{};
+   std::set<std::string> fSuppressErrorsForMissingBranches{};
```
silverweed, line 199: Is there a specific reason for using an ordered set?

vepadulano: Not in general. But in `GetValidatedColumnNames` it comes handy to have a `std::set` rather than a `std::unordered_set` so that the column names are already sorted for the `std::set_intersection` call. If I change to `std::unordered_set` I would need to define a custom comparator inline. Could be done, but given the size of this set is going to be at worst O(100), I would prefer to leave things as they are.

tree/dataframe/src/RDFInterfaceUtils.cxx
```diff
@@ -931,10 +931,9 @@ ColumnNames_t GetValidatedColumnNames(RLoopManager &lm, const unsigned int nColu
       // Look for a possible overlap between the unknown columns and the
       // columns we should ignore for the purpose of the following exception
       std::set<std::string> intersection;
-      auto colsToIgnore = lm.GetSuppressErrorsForMissingBranches();
+      auto &&colsToIgnore = lm.GetSuppressErrorsForMissingBranches();
```
pcanal, line 934: Why 
```suggestion
      auto &&colsToIgnore = lm.GetSuppressErrorsForMissingBranches();
```
rather than:
```suggestion
      const auto &colsToIgnore = lm.GetSuppressErrorsForMissingBranches();
```

vepadulano: No particular reason, it won't matter much for the purpose of passing it to `std::set_intersection`, I will change it in the next commit

pcanal: Related post: https://stackoverflow.com/questions/13130708/what-is-the-advantage-of-using-forwarding-references-in-range-based-for-loops/13130795#13130795

pcanal (approved):
LGTM.

PR #17933, opened 2025-03-10
[RF] Remove RooFitResult methods that use `gMinuit`

guitargeek:
These methods were specific to the old Fortran Minuit that provides the `gMinuit` pointer, and therefore don't work anymore nowadays now that RooFit uses Minuit2 by default.


lmoneta (approved):
LGTM!

PR #17934, opened 2025-03-10
[ntuple] Extend `RNTupleOpenSpec` to allow creating processors from a `TDirectory`

enirolf:
This PR changes `RNTupleOpenSpec` to accept a `TDirectory` (or one of its descendants) instead of a plain file name, and create a `RNTupleProcessor` from there. This provides more flexibility for creating `RNTupleProcessor` objects. Among others, it enables processing RNTuples stored in subdirectories and `TMemFile`s.

One caveat is that the use of `std::variant` is not very ergonomic in Python. After discussing different solutions with @vepadulano, we concluded that the specification would/should not be used for anything beyond creating `RNTupleProcessor`s, and that public access to its data members is not strictly required. It has therefore been changed to a write-only class, and the `RNTupleProcessor` creation factory arguments have been updated accordingly.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
@@ -36,14 +36,32 @@ namespace Internal {
 struct RNTupleProcessorEntryLoader;
 } // namespace Internal
 
-/// Used to specify the underlying RNTuples in RNTupleProcessor
-struct RNTupleOpenSpec {
+// clang-format off
+/**
+\class ROOT::Experimental::RNTupleOpenSpec
+\ingroup NTuple
+\brief Specification of the name and location of an RNTuple, used for creating a new RNTupleProcessor.
+
+An RNTupleOpenSpec can be created by providing either a string with a path to the ROOT file or a pointer to the
+TDirectory (or any of its subclasses) that contains the RNTuple.
+
+Note that the RNTupleOpenSpec is *write-only*, to prevent usability issues with Python.
+*/
+// clang-format on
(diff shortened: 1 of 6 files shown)


tree/ntuple/v7/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -39,9 +39,11 @@ struct RNTupleProcessorEntryLoader;
 /// Used to specify the underlying RNTuples in RNTupleProcessor
 struct RNTupleOpenSpec {
    std::string fNTupleName;
-   std::string fStorage;
+   std::variant<std::string, TDirectory *> fStorage;
```
vepadulano, line 42: I think we need to discuss a bit the public interface of this. I am not sure it's going to be easily usable from Python, which might or might not be what we want.

enirolf:
> As an alternative interface, maybe we can accept a `RNTuple` anchor directly (only for `RNTupleSingleProcessor`?) and require the user to compose that?

I'm not very fond of that idea because 1) it would put more burden on the user and 2) I don't know to what extent that would make things nicer in Python. Of course when it turns out that this is a desired interface by users we can add it, but for now I would prefer not to extend the interface.

I discussed possible solutions with @vepadulano, and ultimately we settled on making `RNTupleOpenSpec` write-only (and making them pass-by-value in the factories). We anyways don't expect users do anything with the spec except create processors and this removes the possibility of having a not-easy-to-use-in-Python variant exposed. Again, only time will tell if this is actually what we want, but it at least keeps the current interface consistent.

vepadulano (approved):
LGTM thanks!

PR #17936, opened 2025-03-10
[ntuple] Split primary and auxiliary models in  `RNTupleJoinProcessor::CreateJoin`

enirolf:
To reflect the interface changes introduced in #17964, instead of a single list of models (one per primary/auxiliary ntuple), a distinction is made between the models of the primary and auxiliary ntuples.

Same as before, internally they are combined into one singular model used by the processor. This model has to have the schemas of auxiliary RNTuples as untyped record fields, where the name of the record field is the name of the RNTuple.

A new addition is the added support for _partial_ model specification in the list of auxiliary models. This means that users can pass a null pointer in the position of the corresponding ntuple, and then only for that ntuple, the processor will infer the model from the on-disk descriptor.

Helps move forward #17132.

### DIFF
--- a/tree/ntuple/v7/src/RNTupleProcessor.cxx
+++ b/tree/ntuple/v7/src/RNTupleProcessor.cxx
@@ -398,6 +398,13 @@ ROOT::Experimental::RNTupleJoinProcessor::RNTupleJoinProcessor(const RNTupleOpen
          return fieldName.substr(0, n.fNTupleName.size()) == n.fNTupleName;
       });
 
+      // If the current field name does not begin with the name of one of the auxiliary ntuples, we are dealing with a
+      // field from the primary ntuple, so it can be added as a field context. Otherwise, if it does begin with the
+      // name, but is not equal to just the name (e.g. it is a subfield of `auxNTupleName`, which means it is a proper
+      // field in the corresponding auxiliary ntuple) we also need to add it as a field context. If it is exactly equal
+      // to an auxiliary ntuple name, it is the untyped record field containing the auxiliary fields itself. This one we
+      // don't want to add as a field context, because there is nothing to read from.
+      // TODO(fdegeus) handle the case where a primary field has the name of an auxiliary ntuple.
       if (auxNTupleName == auxNTuples.end()) {
          fFieldContexts.try_emplace(fieldName, field.Clone(field.GetFieldName()), fEntry->GetToken(fieldName));
       } else if (fieldName != auxNTupleName->fNTupleName) {
--- a/tree/ntuple/v7/src/RNTupleProcessor.cxx
+++ b/tree/ntuple/v7/src/RNTupleProcessor.cxx
@@ -109,23 +109,23 @@ ROOT::Experimental::RNTupleProcessor::CreateChain(std::vector<std::unique_ptr<RN
 }
 
 std::unique_ptr<ROOT::Experimental::RNTupleProcessor>
-ROOT::Experimental::RNTupleProcessor::CreateJoin(RNTupleOpenSpec primaryNTuple, std::vector<RNTupleOpenSpec> auxNTuples,
+ROOT::Experimental::RNTupleProcessor::CreateJoin(const RNTupleOpenSpec &primaryNTuple,
+                                                 const std::vector<RNTupleOpenSpec> &auxNTuples,
                                                  const std::vector<std::string> &joinFields,
-                                                 std::vector<std::unique_ptr<RNTupleModel>> models)
+                                                 std::unique_ptr<RNTupleModel> primaryModel,
+                                                 std::vector<std::unique_ptr<RNTupleModel>> auxModels)
 {
-   auto processorName = primaryNTuple.fNTupleName;
-   return CreateJoin(std::move(primaryNTuple), std::move(auxNTuples), joinFields, processorName, std::move(models));
+   return CreateJoin(primaryNTuple, auxNTuples, joinFields, primaryNTuple.fNTupleName, std::move(primaryModel),
+                     std::move(auxModels));
 }
 
-std::unique_ptr<ROOT::Experimental::RNTupleProcessor>
-ROOT::Experimental::RNTupleProcessor::CreateJoin(RNTupleOpenSpec primaryNTuple, std::vector<RNTupleOpenSpec> auxNTuples,
-                                                 const std::vector<std::string> &joinFields,
-                                                 std::string_view processorName,
-                                                 std::vector<std::unique_ptr<RNTupleModel>> models)
+std::unique_ptr<ROOT::Experimental::RNTupleProcessor> ROOT::Experimental::RNTupleProcessor::CreateJoin(
+   const RNTupleOpenSpec &primaryNTuple, const std::vector<RNTupleOpenSpec> &auxNTuples,
+   const std::vector<std::string> &joinFields, std::string_view processorName,
+   std::unique_ptr<RNTupleModel> primaryModel, std::vector<std::unique_ptr<RNTupleModel>> auxModels)
 {
-   if (!models.empty() && models.size() != (auxNTuples.size() + 1)) {
-      throw RException(R__FAIL("number of provided models must match number of specified ntuples"));
-   }
+   if (!auxModels.empty() && auxModels.size() != auxNTuples.size())
+      throw RException(R__FAIL("number of auxiliary models and auxiliary RNTuples does not match"));
 
    if (joinFields.size() > 4) {
       throw RException(R__FAIL("a maximum of four join fields is allowed"));
@@ -146,14 +146,8 @@ ROOT::Experimental::RNTupleProcessor::CreateJoin(RNTupleOpenSpec primaryNTuple, 
       }
    }
 
-   std::unique_ptr<RNTupleJoinProcessor> processor;
-   if (!models.empty()) {
-      processor = std::unique_ptr<RNTupleJoinProcessor>(new RNTupleJoinProcessor(
-         std::move(primaryNTuple), std::move(auxNTuples), joinFields, processorName, std::move(models)));
-   } else {
-      processor = std::unique_ptr<RNTupleJoinProcessor>(
-         new RNTupleJoinProcessor(std::move(primaryNTuple), std::move(auxNTuples), joinFields, processorName));
-   }
+   std::unique_ptr<RNTupleJoinProcessor> processor = std::unique_ptr<RNTupleJoinProcessor>(new RNTupleJoinProcessor(
+      primaryNTuple, auxNTuples, joinFields, processorName, std::move(primaryModel), std::move(auxModels)));
 
    processor->SetJoinFieldTokens(joinFields);
    processor->ConnectFields();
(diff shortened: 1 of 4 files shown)


tree/ntuple/v7/test/ntuple_processor_join.cxx
```diff
@@ -241,24 +241,28 @@ TEST_F(RNTupleJoinProcessorTest, MissingEntries)
...
+   std::vector<RNTupleOpenSpec> auxNTuples{{fNTupleNames[1], fFileNames[1]}, {fNTupleNames[2], fFileNames[2]}};
+
+   auto joinModel = RNTupleProcessor::CreateJoinModel(std::move(primaryModel), std::move(auxModels), auxNTuples);
+
+   auto i = joinModel->GetDefaultEntry().GetPtr<int>("i");
+   auto x = joinModel->GetDefaultEntry().GetPtr<float>("x");
+   auto y = joinModel->GetDefaultEntry().GetPtr<std::vector<float>>(fNTupleNames[1] + ".y");
+   auto z = joinModel->GetDefaultEntry().GetPtr<float>(fNTupleNames[2] + ".z");
+
+   auto proc = RNTupleProcessor::CreateJoin({fNTupleNames[0], fFileNames[0]}, auxNTuples, {"i"}, std::move(joinModel));
```
pcanal, line 265: The interface boils down to:
```
auto primaryModel = RNTupleModel::Create();
std::vector<std::unique_ptr<RNTupleModel>> auxModels;
... fill auxModels ...
std::vector<RNTupleOpenSpec> auxNTuples;
... fill auxNTuples ...
auto joinModel = RNTupleProcessor::CreateJoinModel(std::move(primaryModel), std::move(auxModels), auxNTuples);
auto proc = RNTupleProcessor::CreateJoin({fNTupleNames[0], fFileNames[0]}, auxNTuples, {"i"}, std::move(joinModel));
```

Feels a bit redundant and/or disjointed (pun not really intended).  In particular in the last 2 lines `auxNTuples` is repeated and (I assume) must be in the exact same state (ie. be the same) during both calls.  What happens if you pass 2 completely different vector to both?  Couldn't the `CreateJoin` extract the `auxNTuples` from the result of `CreateJoinModel` (in which case the risk of wrong call would be eliminated).

silverweed: I agree with Philippe that if we can reduce the redundancy on the user side we should go for it, even if it means making the function signature longer (maybe we should pass an `Options` struct to it?)

enirolf: I think in this particular case using an `Options` struct will be just as if not more cumbersome, but I will change the signature to take `primaryModel, {auxModels...}` then :+1:.

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -352,135 +349,122 @@ ROOT::Experimental::RNTupleJoinProcessor::RNTupleJoinProcessor(const RNTupleOpen
...
-      model = pageSource->GetSharedDescriptorGuard()->CreateModel();
+   fModel = std::move(primaryModel);
+   fModel->Unfreeze();
 
-   model->Freeze();
-   auto entry = model->CreateBareEntry();
+   // Create an anonymous record field for each auxiliary ntuple, containing their top-level fields. These original
+   // top-level fields are registered as subfields in the join model, such that they can be accessed as
+   // `auxNTupleName.fieldName`.
+   // TODO(fdegeus) Support projected fields
```
hahnjo, line 408: I'm not sure what this means. IIRC projected fields are generally not "supported" for reading, ie they are duplicated in the entry

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -352,135 +349,122 @@ ROOT::Experimental::RNTupleJoinProcessor::RNTupleJoinProcessor(const RNTupleOpen
...
+   for (unsigned i = 0; i < auxModels.size(); ++i) {
+      std::vector<std::unique_ptr<RFieldBase>> auxFields;
 
-   // Append the auxiliary fields to the join model
-   fModel->Unfreeze();
+      if (!auxModels[i]) {
+         fAuxiliaryPageSources[i]->Attach();
+         auxModels[i] = fAuxiliaryPageSources[i]->GetSharedDescriptorGuard()->CreateModel();
+      }
```
hahnjo, line 415: Maybe this can be moved to the `RNTupleJoinProcessor` constructor? In case of an empty vector, `resize` with `nullptr` and then generally replace all `nullptr` with a reconstructed model?

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -352,135 +349,122 @@ ROOT::Experimental::RNTupleJoinProcessor::RNTupleJoinProcessor(const RNTupleOpen
...
 
-      auxFields.emplace_back(field.Clone(field.GetQualifiedFieldName()));
-   }
-   std::unique_ptr<RFieldBase> auxParentField =
-      std::make_unique<RRecordField>(auxNTuple.fNTupleName, std::move(auxFields));
+      auto auxParentField = std::make_unique<RRecordField>(fNTuples[i + 1].fNTupleName, std::move(auxFields));
+      if (!auxParentField) {
+         throw RException(R__FAIL("cannot add fields of auxiliary RNTuple \"" + fNTuples[i + 1].fNTupleName + "\""));
+      }
```
hahnjo, line 424: Under which conditions can this happen, other than out-of-memory?

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -367,120 +376,72 @@ ROOT::Experimental::RNTupleJoinProcessor::RNTupleJoinProcessor(const RNTupleOpen
...
-      // If the field belongs to the auxiliary ntuple currently being added, apart from assigning its entry value the
-      // correct pointer, we also have to create a field context for it.
-      if (fieldContext == fFieldContexts.end()) {
-         // If the model has a default entry, use the value pointers from the entry in the entry managed by the
-         // processor. This way, the pointers returned by RNTupleModel::MakeField can be used in the processor loop to
-         // access the corresponding field values.
-         if (!model->IsBare()) {
-            auto valuePtr = model->GetDefaultEntry().GetPtr<void>(field.GetFieldName());
-            newEntry->BindValue(field.GetQualifiedFieldName(), valuePtr);
+      // If the model has a default entry, adopt its value pointers.This way, the pointers returned by
```
hahnjo, line 432:
```suggestion
      // If the model has a default entry, adopt its value pointers. This way, the pointers returned by
```

tree/ntuple/v7/src/RNTupleProcessor.cxx
enirolf: You're right, honestly I don't remember my train of thought when adding this comment and it works all fine. I'll delete it!

tree/ntuple/v7/src/RNTupleProcessor.cxx
enirolf: Done, see lines 358-366.

tree/ntuple/v7/src/RNTupleProcessor.cxx
enirolf: None that I can think of right now. I've wrapped it in an `R__unlikely` now, but I'm also fine with removing it altogether.

hahnjo: I believe if we don't check other return values of `std::make_unique` it should be removed here. If the constructor of `RRecordField` throws, the exception will be propagated on its own.

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -352,13 +349,28 @@ ROOT::Experimental::RNTupleJoinProcessor::RNTupleJoinProcessor(const RNTupleOpen
...
+         fAuxiliaryPageSources[i]->Attach();
+         auxModels[i] = fAuxiliaryPageSources[i]->GetSharedDescriptorGuard()->CreateModel();
+      }
+   }
+
+   SetModel(std::move(primaryModel), std::move(auxModels));
 
    fModel->Freeze();
-   fEntry = fModel->CreateEntry();
+   fEntry = fModel->CreateBareEntry();
```
hahnjo, line 373: Can you test this with a bare model? I think in that case this change creates a bare entry without `BindValue` below.

A possible improvement could be to just use the default entry in case of a non-bare model (without carrying over the value pointers below), and only call `CreateEntry()` for a bare model.

hahnjo (commented):
One final comment when looking at the code again; we probably want to test a mixture of bare and non-bare models...

tree/ntuple/v7/src/RNTupleProcessor.cxx
enirolf: Done, see latest commit. I have actually been considering to use the default entry instead to simplify things, but since this should apply to the other processors then as well I will address this in a separate PR.

enirolf: Done (and indeed have to use a regular entry here). I was thinking the same for the default entry actually, that would simplify a lot of things. This would also require changes to the other processors to keep it consistent, so I will open a separate PR for this.

hahnjo (approved):
Thanks, LGTM! (should have a second approval)

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -364,123 +374,91 @@ ROOT::Experimental::RNTupleJoinProcessor::RNTupleJoinProcessor(const RNTupleOpen
       auto &field = value.GetField();
       const auto &fieldName = field.GetQualifiedFieldName();
 
+      auto isAuxParent = std::find_if(auxNTuples.cbegin(), auxNTuples.cend(), [&fieldName](const RNTupleOpenSpec &n) {
```
silverweed: I think this part could use a comment to explain why we want to skip the loop when this condition is true...it's not immediately obvious to me.

tree/ntuple/v7/src/RNTupleProcessor.cxx
```diff
@@ -364,123 +374,91 @@ ROOT::Experimental::RNTupleJoinProcessor::RNTupleJoinProcessor(const RNTupleOpen
...
-   // ntuples.
-   std::vector<std::unique_ptr<ROOT::RFieldBase>> auxFields;
-   auxFields.reserve(entry->fValues.size());
-   for (const auto &val : *entry) {
-      auto &field = val.GetField();
+   // Create an anonymous record field for each auxiliary ntuple, containing their top-level fields. These original
+   // top-level fields are registered as subfields in the join model, such that they can be accessed as
+   // `auxNTupleName.fieldName`.
+   for (unsigned i = 0; i < auxModels.size(); ++i) {
+      std::vector<std::unique_ptr<ROOT::RFieldBase>> auxFields;
```
silverweed, line 412: You should probably `reserve` the space in the vector here

silverweed (approved):
LGTM

PR #17939, opened 2025-03-10
[ntuple] Make the fields of RPageRange private

silverweed:
## Checklist:

- [x] tested changes locally


hahnjo (commented):
Generally LGTM, but requires an update in `roottest`...

enirolf (approved):
Thanks!

PR #17942, opened 2025-03-10
[Minuit2] Rule of five for Minuit2 linear algebra classes

guitargeek:
Avoid potential double free that would happen if the LASymMatrix or LAVector is move-constructed.

The subsequent undefined behavior might be resulting in the sporadic build failures that we currently have.

The problem has been there since forever, but my suspicion is that the undefined behavior only got visible now by chance with other changes to Minuit2.


lmoneta (approved):
Thanks for adding these extra constructors and assignment operators

PR #17943, opened 2025-03-11
[Minuit2] Bring back old implementation of `mndspr` and `Mndspmv`

guitargeek:
I must have made a mistake in the translation to more idiomatic C++ code for these functions in the previous PR:
   * https://github.com/root-project/root/pull/17852

The mistake led to sporadic failures on Windows when using Minuit 2.

This PR suggests to use the old implementation again.


lmoneta (approved):
LGTM!
I think it is better to keep the orginal code here

PR #17947, opened 2025-03-11
Add GetVoidArray for access to TArrayC data through a void*

siliataider:
# This Pull request:

## Changes:

Introduce the `GetVoidArray` methods to the `TArrayC` class, enabling an alternative way to access to the underlying data through a void*. This is particularly needed for the UHI use case of the `TH1C` class, bypassing the cppyy interpretation of `char*` as `str`.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/cont/inc/TArrayC.h
+++ b/core/cont/inc/TArrayC.h
@@ -42,6 +42,8 @@ public:
    void          Copy(TArrayC &array) const {array.Set(fN,fArray);}
    const Char_t *GetArray() const { return fArray; }
    Char_t       *GetArray() { return fArray; }
+   const void *GetVoidArray() const { return static_cast<const void *>(fArray); }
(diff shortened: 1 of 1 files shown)


core/cont/inc/TArrayC.h
```diff
@@ -42,6 +42,8 @@ class TArrayC : public TArray {
    void          Copy(TArrayC &array) const {array.Set(fN,fArray);}
    const Char_t *GetArray() const { return fArray; }
    Char_t       *GetArray() { return fArray; }
+   const void *GetVoidArray() const { return static_cast<const void *>(fArray); }
+   void *GetVoidArray() { return static_cast<void *>(fArray); }
```
pcanal, line 46: If we can get away with it, I think it would be better to have only the `const` overload. Also the name `GetDataPtr` may or may not be a clearer name.

PR #17952, opened 2025-03-11
[interop] Update CppInterOp to latest

aaronj0:
This version bump pulls in fixes to the CMake, upstreamed in https://github.com/compiler-research/CppInterOp/pull/517. This PR also gets rid of the temporary diff with upstream induced by https://github.com/root-project/root/pull/17791


aaronj0:
Merging, as the failure on Windows is unrelated

PR #8709, opened 2021-07-21
[CMake] Move towards target-based CMake and partly fix picking up headers from an installed ROOT

hageboeck:
Partial fix of #8708.
In a setup where ROOT was installed in a system directory, ROOT was picking up headers from that directory instead of its own. 

## How to reproduce:
1. `echo '#error This is the wrong header' > /my/include/directory/RooSpan.h` (or a few other headers).
2. Install some builtins into that directory, e.g. VDT
3. `cmake -DCMAKE_PREFIX_PATH=/my/include/directory/ <root>` to create a dependency to that include directory.
4. Build.

The problem only becomes visible when `A` depends on `B and C`, and `B` depends on something in `/system/include/`, and `C` is installed in those system includes as well. This generates a compile command such as:
```
-I.../core/x -I.../core/y -I.../core/... -I.../A/include -I.../B/include -I/system/include/ -I.../C/include ...
```

## In this PR:
- Includes for VDT and XROOTD are fixed by making them `IMPORTED SYSTEM` targets, so their includes have lowest precedence.
- Some cheating where include directories are copied around between targets is removed. CMake should handle this.
- Some dependency and target management is simplified (or rather modernised with target-based cmake)
- A broken dependency in RooFit is fixed, which was previously hidden by the cheating with include directories.
- Teach cling to deal with `-isystem` (https://its.cern.ch/jira/browse/ROOT-10414)

## What remains to be done:
It is likely that more builtins (or rather `FindXXX` have to be converted to `IMPORTED` targets, so they don't provoke this error again. A broken configuration can be detected by
1. Having CMake pick up a dependency in some common directory
2. Either
  - Placing a lot of `#error`-ROOT headers in there or
  - Searching `compile_commands.json` for `-I/my/include/directory/`
3. Fixing the `FindXXX` for this dependency.

PR #17956, opened 2025-03-11
[CI] Post install tests

hageboeck:
~~Note, this contains the commits from #8709, so that should be merged first.~~ ✅ 

After the usual CI builds:
- Install root
- Configure and build a small CMake project that links against ROOT
- Run a ROOT macro and run a compiled executable using the installed ROOT


guitargeek (approved):
Thanks! This is a really great idea :+1:

PR #17957, opened 2025-03-11
[Minuit2] Add missing operator to add algebraic vector objects

guitargeek:
This is to accommodate the mathematical operations in recent Fumili developments:
   * https://github.com/root-project/root/pull/17558


lmoneta (approved):
LGTM!
Thanks for re-adding this operator

PR #17951, opened 2025-03-11
Add TF1::SavePrimitive testing in stressGraphics

linev:
1. Use array to store points in TF1
2. Correctly store all line/fill/axis attributes in TF1
3. Enable test of C-macro storage for test 45 in stressGraphics

PR #17958, opened 2025-03-11
Implement `TF2::SavePrimitive` and testing in `stressGraphics`

linev:
1. Provide special TF2 constructor signature (as for TF1)
2. Save large array of points instead of function (as TF1)
3. Implement storage of custom color palette which used in `waves()` 
4. Enable C file creation for test32 in `stressGraphics`, making exactly same output

Follow up for https://github.com/root-project/root/pull/17951

Here for testing of stressGraphics results


pcanal (approved):
Looks fine, thanks.

couet (approved):

PR #17960, opened 2025-03-12
[python] Remove fragile test

vepadulano:
This test is fragile because we cannot be sure that RSS is a stable metric to evaluate the absence of a memory leak. The test itself has already demonstrated to be limited, in fact it was already modified or relaxed twice before:

https://github.com/root-project/root/commit/038054b3e5fdec0a499f0e25f571f88016565557 https://github.com/root-project/root/commit/015f44025d607d793922ba1eaae4aa9297339b07

It now starts also causing intermittent failures on MacOS which are not representative of a new memory leak, so it is better to just remove it.

### DIFF
--- a/bindings/pyroot/pythonizations/test/memory.py
+++ b/bindings/pyroot/pythonizations/test/memory.py
@@ -13,37 +13,6 @@ def _leak(obj):
 
 class MemoryStlString(unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


dpiparo (approved):
Thanks for the thorough investigation.

PR #17950, opened 2025-03-11
[python] Add UHI PlottableHistogram protocol Support to 1D TH1 Derived Classes

siliataider:
# This Pull request:

## Adds:

This PR introduces the [Unified Histogram Interface](https://uhi.readthedocs.io/en/latest/index.html) plottable histogram functionality to the `TH1`-derived classes in ROOT. The changes include the addition of the methods described by the UHI `PlottableHistogram` protocol, ensuring compatibility with third party plotting libraries that support the protocol.

**Note:**  Currently, support is limited to 1D histograms.

## Checklist:

- [x] tested changes locally

PR #17967, opened 2025-03-12
Add option for generic Python tests

vepadulano:
Follow up to https://github.com/root-project/root/pull/17950

In the linked PR we found ourselves in need to create a Python ctest test that uses `pytest` as the test driver. The currently available `ROOT_ADD_PYUNITTEST` cmake function is limited because it only allows Python tests written with `unittest`, so we can introduce a new one more generic.


dpiparo (approved):
Thanks for this work. For me the PR can be merged as is, nevertheless let me ask a question: have you considered perhaps unifying the two functions not to add another one to the CMake infra?

vepadulano:
>  have you considered perhaps unifying the two functions not to add another one to the CMake infra?

Thanks for the idea, I implemented and rebased :+1:

PR #17968, opened 2025-03-12
[pyroot] Move JupyROOT files at build time.

### LINKED ISSUE #17859: [CMake] JupyROOT sources not correctly tracked by CMake
When changing `JupyROOT` sources in incremental builds, the updated Python sources are not copied into the build directory and the install directory. This makes it impossible to test changes in the CI without triggering a full rebuild, which is unnecessarily expensive.

See in the relevant CMake file, where there is only a copy triggered in the initial configuration:
https://github.com/root-project/root/blob/master/bindings/jupyroot/CMakeLists.txt#L31

I guess one would have to do something with custom commands and targets, similar to the CMakeLists.txt of the pythonization sources:
https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/CMakeLists.txt#L151

hageboeck:
@guitargeek could you check if this fixes the problem?

Fix #17859.

### DIFF
--- a/bindings/jupyroot/CMakeLists.txt
+++ b/bindings/jupyroot/CMakeLists.txt
@@ -27,8 +27,17 @@ set(py_sources
   JupyROOT/magics/jsrootmagic.py
 )
 
-set(JupyROOTPySrcDir python/JupyROOT)
-file(COPY ${JupyROOTPySrcDir} DESTINATION ${localruntimedir})
+# Create a target that keeps JupyROOT source in the build tree up to date:
+set(JupyROOTPySrcDir ${CMAKE_CURRENT_SOURCE_DIR}/python/JupyROOT/)
+file(GLOB_RECURSE input_files RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} ${JupyROOTPySrcDir}/**)
+set(output_files ${input_files})
+list(TRANSFORM output_files REPLACE "python/" ${localruntimedir}/)
+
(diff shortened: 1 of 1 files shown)


ferdymercury:
Tangentially related: https://its.cern.ch/jira/browse/ROOT-10986

guitargeek:
> @guitargeek could you check if this fixes the problem?

Yes it does, thank you so much! If I change a source file in `bindings/jupyroot/python/JupyROOT/` with your PR and do an incremental build, the corresponding sources in the build and in the install tree are both synced. This wasn't the case before.

@ferdymercury, thanks for the ping on that ticket. I think what is suggested there doesn't scale for the reasons that Serguei explained. I'll close that ticket, also because I know that the author of that ticket is not a ROOT user anymore.

hageboeck:
Trying again with a clean build

hageboeck:
@guitargeek do you think that the failure on fedora is related to the PR? Does it have to do with `-m py_compile` happening too late?

guitargeek:
Strange, looks like it picked up the docker images I updated yesterday with this commit...
* https://github.com/root-project/root-ci-images/commit/3f07f6c0ce64ac848ddeb86587401aed9a4d04bd

...but didn't pick up Vincenzos patch for Python 3.13 support:
* https://github.com/root-project/root/commit/65cca1211b97b2ec32b754842e1894d970451ecc

At least it's not related to the changes suggested in this PR :)

Issue #17969, opened 2025-03-12
Failure of thisroot.sh on freebsd


eamjensen:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

What is the intention of the `cut` command in line 172 in `thisroot.sh` exactly?

```shell
trueExe=$(cut -d '' -f1 /proc/$$/cmdline) || return 1
```

It does not make sense to use `cut` with an empty delimiter, and doing so produces the following error on FreeBSD (I maintain the ROOT port on FreeBSD: https://www.freshports.org/devel/root):

```shell
cut: bad delimiter
```

As a result, the shell variable SHELLNAME cannot be defined, when `thisroot.sh` is sourced, and the following message is then printed

```shell
WARNING: shell name was not found. Assuming 'bash'.
```

On FreeBSD, the contents of `/proc/$$/cmdline` could for example be

```shell
-bash^@
```

where `^@` is the null character.

I would suggest to pipe whatever output one wishes to get from `/proc/$$/cmdline` through `tr -d '\0'` in order to get rid of any potential null characters, e.g.

```shell
trueExe=$(cut -d ' ' -f1 /proc/$$/cmdline | tr -d '\0') || return 1
```

Note the space, `' '`, now given as the delimiter to `cut`.

### Reproducer

`source <path to root dir>/bin/thisroot.sh`

### ROOT version

```
ROOT v6.34.04
Built for freebsdamd64 on Mar 12 2025, 11:51:50
From tags/6-34-04@6-34-04
With 
Binary directory: /usr/local/bin
```

### Installation method

Build from source

### Operating system

FreeBSD 14.2-RELEASE-p2

ferdymercury:
> What is the intention of the `cut` command in line 172 in `thisroot.sh` exactly?

I've asked the author: https://stackoverflow.com/a/23011530/7471760

eamjensen:
Hi @ferdymercury, I see the author just replied, maybe you saw it too.

The idea apparently is to have a null character as the delimiter, so my question then is, would a suitable solution on all platforms be to do

```shell
trueExe=$(cut -d '\0' -f1 /proc/$$/cmdline) || return 1
```

?

ferdymercury:
Thanks! The solution you proposed does not work on bash/Ubuntu. I have updated now the PR to contain:
$'\000'
as delimiter, which seems to work well on Ubuntu. Does that work for you, too?

eamjensen:
Unfortunately not.
But I may have found another solution:

[...]

PR #17970, opened 2025-03-12
cut command splitting by null bytes fails on MacOS and BSD

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/17969

This prevents errors in freebsd and macos, with the cut command being implemented differently vs Linux and complaining about a bad delimiter (empty) when using `-d ''`. While this might be a 'bug' in the freebsd and macos implementation of "cut", it makes sense to find an alternative here that is more cross-platform compliant.

All credit for the solution goes to eamjensen and mkelement0

Related: https://apple.stackexchange.com/questions/268352/how-to-use-cut-with-null-delimiter
https://stackoverflow.com/questions/23011370/how-to-recognize-whether-bash-or-dash-is-being-used-within-a-script/23011530?noredirect=1#comment140210695_23011530

## Checklist:

- [x] tested changes locally

### DIFF
--- a/config/thisroot.sh
+++ b/config/thisroot.sh
@@ -169,7 +169,7 @@ getTrueShellExeName() { # mklement0 https://stackoverflow.com/a/23011530/7471760
    local trueExe nextTarget 2>/dev/null # ignore error in shells without `local`
    # Determine the shell executable filename.
    if [ -r "/proc/$$/cmdline" ]; then
-      trueExe=$(cut -d $'\000' -f1 /proc/$$/cmdline) || return 1
+      trueExe=$(xargs -0 < /proc/$$/cmdline | awk NR==1'{ printf "%s",$1 }') || return 1
       # Qemu emulation has cmdline start with the emulator
       if [ "${trueExe##*qemu*}" != "${trueExe}" ]; then
          # but qemu sets comm to the emulated command
--- a/config/thisroot.sh
+++ b/config/thisroot.sh
@@ -169,7 +169,7 @@ getTrueShellExeName() { # mklement0 https://stackoverflow.com/a/23011530/7471760
    local trueExe nextTarget 2>/dev/null # ignore error in shells without `local`
    # Determine the shell executable filename.
    if [ -r "/proc/$$/cmdline" ]; then
-      trueExe=$(cut -d '' -f1 /proc/$$/cmdline) || return 1
+      trueExe=$(cut -d ' ' -f1 /proc/$$/cmdline | tr -d '\0') || return 1
       # Qemu emulation has cmdline start with the emulator
       if [ "${trueExe##*qemu*}" != "${trueExe}" ]; then
          # but qemu sets comm to the emulated command
--- a/config/thisroot.sh
+++ b/config/thisroot.sh
@@ -169,7 +169,7 @@ getTrueShellExeName() { # mklement0 https://stackoverflow.com/a/23011530/7471760
    local trueExe nextTarget 2>/dev/null # ignore error in shells without `local`
    # Determine the shell executable filename.
    if [ -r "/proc/$$/cmdline" ]; then
-      trueExe=$(cut -d ' ' -f1 /proc/$$/cmdline | tr -d '\0') || return 1
+      trueExe=$(cut -d $'\000' -f1 /proc/$$/cmdline) || return 1
       # Qemu emulation has cmdline start with the emulator
       if [ "${trueExe##*qemu*}" != "${trueExe}" ]; then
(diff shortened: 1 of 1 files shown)


config/thisroot.sh
```diff
@@ -169,7 +169,7 @@ getTrueShellExeName() { # mklement0 https://stackoverflow.com/a/23011530/7471760
    local trueExe nextTarget 2>/dev/null # ignore error in shells without `local`
    # Determine the shell executable filename.
    if [ -r "/proc/$$/cmdline" ]; then
-      trueExe=$(cut -d '' -f1 /proc/$$/cmdline) || return 1
+      trueExe=$(cut -d $'\000' -f1 /proc/$$/cmdline) || return 1
```
ferdymercury, line 172:
```suggestion
      trueExe=$(xargs -0 < /proc/$$/cmdline | awk NR==1'{ printf "%s",$1 }') || return 1
```

hageboeck:
Hello @ferdymercury, what is the reason for the changes? Did separating by space not work?

ferdymercury:
Hi, separating by space fails if path contains a space. See discussion in the stack overflow link of the issue.

hageboeck:
Hello @ferdymercury,

looks good now! Do you want to squash your commits now?

ferdymercury:
Thanks! Is it ok if you do that yourself via the option squash on merge? (I am without computer this week)

ferdymercury:
Thanks! Hehe makes sense :). I hereby give you global permission for any of my PRs to always squash.

ferdymercury:
Btw i think these two could be closed: https://its.cern.ch/jira/browse/ROOT-10537
https://its.cern.ch/jira/browse/ROOT-9484

couet:
Hi, this PR is from last week, and since then the build of doc master does not work anymore. I do not say that this is due to this PR but [I put some debugging statements in the CI building the ROOT master doc](https://github.com/root-project/root/actions/runs/13972412743/job/39117403803) and it seems to show `source thisroot.sh` is not working anymore. Is it just a coincidence or is this PR the faulty one ?

hageboeck:
> Yes, it's surely this one the faulty one, sorry about that.
> 
> What surprises me is that in https://github.com/root-project/root/actions/runs/13972412743/job/39117403803#step:11:86 it reports: `trueExe=bash`, which I guess is correct in the CI ? but then `SHELLNAME=`
> 
> so we might need to revert unless someone understand why this happens.
> 
> Or alternatively do something like:
> 
> ```
> 
> try cut -d ''
> if it fails, then fall back to xargs variant.
> ```

I tried it manually in an alma9 and ubuntu container, and it works fine.
For some reason, when the documentation builds, the exit code of the subshell is non-zero, so it assigns `trueExe=bash` correctly, but then exits from the function. I don't see why this should be the case, but it didn't make `thisroot.sh` more robust.
We could indeed try a hybrid solution using trial and error until we have the shell name. What do you think @ferdymercury ?

couet:
May beI will revert to see if it fixes the problem the doc cannot be build because of that.

ferdymercury:
> We could indeed try a hybrid solution using trial and error 

Maybe something like:

`trueExe=$(cut -d '' -f1 /proc/$$/cmdline) 2>/dev/null || $(xargs -0 -n 1 < /proc/$$/cmdline 2>/dev/null | head -n 1) || return 1`

hageboeck:
> Maybe something like:
> 
> `trueExe=$(cut -d '' -f1 /proc/$$/cmdline 2>/dev/null) || $(xargs -0 -n 1 < /proc/$$/cmdline | head -n 1) || return 1`

More like `trueExe=$(cut -d '' -f1 /proc/$$/cmdline 2>/dev/null) || trueExe=$(xargs -0 -n 1 < /proc/$$/cmdline | head -n 1) || return 1`, but that indeed seems to work now.

ferdymercury:
> More like `trueExe=$(cut -d '' -f1 /proc/$$/cmdline 2>/dev/null) || trueExe=$(xargs -0 -n 1 < /proc/$$/cmdline | head -n 1) || return 1`, but that indeed seems to work now.

@eamjensen can you verify on your side?

eamjensen:
@ferdymercury sorry for the delay!
This new solution works on FreeBSD (and hence, probably on Mac as well - do we know someone who can verify this?)

hageboeck:
> @ferdymercury sorry for the delay! This new solution works on FreeBSD (and hence, probably on Mac as well - do we know someone who can verify this?)

It should be OK for mac, since one of the commands in the chain always works.

PR #17972, opened 2025-03-12
Fix numba tests with Python 3.13+

vepadulano:
See commits descriptions


guitargeek (approved):
Thank you so much for the quick fix! Great work

PR #17975, opened 2025-03-12
Improve exception safety of RDataSource range processing

vepadulano:
Create an internal RAII struct to manage the calls to the initialization and finalization routines for the nodes of the computation graphs and the data source.

Part 7 of N for https://github.com/root-project/root/pull/17895


martamaja10 (approved):
LGTM, thanks Vincenzo!

PR #17977, opened 2025-03-12
[CMake] Use flexiblas/cblas.h if FlexiBLAS is found

guitargeek:
Consistently use `cblas` interface.


guitargeek:
Doesn't help the Alma 9 and Fedora situation.

PR #17984, opened 2025-03-13
[ci] Use Python 3.12+ tarfile data extraction filter

guitargeek:
Fixes the following deprecation warning:
```txt
  /__w/root/root/.github/workflows/root-ci-config/build_root.py:293: DeprecationWarning: Python 3.14 will, by default, filter extracted tar archives and reject files or modify their metadata. Use the filter argument to control this behavior.
```

Analogous to:
  * https://github.com/scikit-hep/pyhf/pull/2455

See also:
  * https://docs.python.org/3.12/library/tarfile.html#tarfile-extraction-filter


dpiparo (approved):
lgtm!

PR #17986, opened 2025-03-13
[ntuple] Ensure type name given by RField<T> is renormalized

vepadulano:
To avoid inconsistencies between the full type name of a custom class reported by the RNTuple IO and the type name reported by ROOT meta. The commit adds a unittest. Previously, the test would fail with exceptions such as:

```
321: unknown file: Failure
321: C++ exception with description "type mismatch for field f2: DataVector<std::int32_t,std::vector<CustomStruct>> vs. DataVector<int,vector<CustomStruct> >
321: At:
321:   void ROOT::Experimental::REntry::EnsureMatchingType(RFieldToken) const [with T = DataVector<int, std::vector<CustomStruct> >]
321: " thrown in the test body.
```

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RField.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RField.hxx
@@ -18,6 +18,7 @@
 
 #include <ROOT/RError.hxx>
 #include <ROOT/RFieldBase.hxx>
+#include <ROOT/RFieldUtils.hxx>
 #include <ROOT/RNTupleSerialize.hxx>
 #include <ROOT/RNTupleUtil.hxx>
 #include <ROOT/RSpan.hxx>
@@ -291,7 +292,7 @@ public:
 template <typename T, typename = void>
 class RField final : public RClassField {
 public:
-   static std::string TypeName() { return ROOT::Internal::GetDemangledTypeName(typeid(T)); }
+   static std::string TypeName() { return ROOT::Internal::GetRenormalizedDemangledTypeName(typeid(T)); }
    RField(std::string_view name) : RClassField(name, TypeName())
    {
       static_assert(std::is_class_v<T>, "no I/O support for this basic C++ type");
--- a/tree/ntuple/v7/test/ntuple_type_name.cxx
+++ b/tree/ntuple/v7/test/ntuple_type_name.cxx
@@ -70,6 +70,7 @@ TEST(RNTuple, TClassDefaultTemplateParameter)
       model->AddField(RFieldBase::Create("f4", "struct DataVector<bool,vector<unsigned>>").Unwrap());
       model->AddField(RFieldBase::Create("f5", "DataVector<Double32_t>").Unwrap());
       model->AddField(RFieldBase::Create("f6", "DataVector<int, double>").Unwrap());
+      model->MakeField<DataVector<StructUsingCollectionProxy<int>>>("f7");
       auto writer = RNTupleWriter::Recreate(std::move(model), "ntpl", fileGuard.GetPath());
    }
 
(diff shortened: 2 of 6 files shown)


tree/ntuple/v7/inc/ROOT/RField.hxx
```diff
@@ -286,7 +287,7 @@ public:
 template <typename T, typename = void>
 class RField final : public RClassField {
 public:
-   static std::string TypeName() { return ROOT::Internal::GetDemangledTypeName(typeid(T)); }
+   static std::string TypeName() { return ROOT::Experimental::Internal::GetDemangledRenormalizedTypeName(typeid(T)); }
```
pcanal, line 290: Shouldn't `GetDemangledRenormalizedTypeName` be simply in `ROOT::Internal`?

tree/ntuple/v7/test/ntuple_type_name.cxx
```diff
@@ -281,3 +281,26 @@ TEST(RNTuple, ContextDependentTypeNames)
...
+   // Test two cases for type name renormalization:
+   // - class with template argument that is normalized by the RNTuple IO (e.g. int to std::int32_t)
+   // - class with template argument which is another template class (e.g. myclass<std::vector<int>>) which by default
+   //   is stripped of the `std::` prefix and has a trailing blank space.
+   FileRaii fileGuard("test_ntuple_atlas_like_datavector.root");
+
+   {
+      auto model = RNTupleModel::Create();
+      auto f1 = model->MakeField<DataVector<int>>("f1");
+      auto f2 = model->MakeField<DataVector<int, std::vector<CustomStruct>>>("f2");
```
pcanal, line 296: Do we need to add 'new' dictionaries request to this test in order to fully test the user case?

tree/ntuple/v7/test/ntuple_type_name.cxx
```diff
@@ -281,3 +281,26 @@ TEST(RNTuple, ContextDependentTypeNames)
...
+      auto writer = RNTupleWriter::Recreate(std::move(model), "ntpl", fileGuard.GetPath());
+   }
+
+   auto reader = RNTupleReader::Open("ntpl", fileGuard.GetPath());
+   EXPECT_EQ(0u, reader->GetNEntries());
+
+   // The test would fail because of an exception thrown in REntry::EnsureMatchingType
+   auto f1 = reader->GetModel().GetDefaultEntry().GetPtr<DataVector<int>>("f1");
+   auto f2 = reader->GetModel().GetDefaultEntry().GetPtr<DataVector<int, std::vector<CustomStruct>>>("f2");
```
hahnjo, line 305: This can go into the existing `TClassDefaultTemplateParameter` which already creates the exact fields. It already tests views, which work because we allow casting and only require that the on-disk column types can be connected...

hahnjo (changes requested):
I believe we need the same treatment for `RField` inheriting from `RProxiedCollectionField`

tree/ntuple/v7/test/ntuple_type_name.cxx
vepadulano: I have changed the second field to use `DataVector<bool, std::vector<unsigned int>>` which already has a dictionary request at https://github.com/root-project/root/blob/cb10c61bef3546cfbd24d12f4288eeb2c3655fd0/tree/ntuple/v7/test/CustomStructLinkDef.h#L30 and for the first field `DataVector<int>` has already a dictionary request at https://github.com/root-project/root/blob/cb10c61bef3546cfbd24d12f4288eeb2c3655fd0/tree/ntuple/v7/test/CustomStructLinkDef.h#L28 IIUC since the second template argument is defaulted to `double`.

hahnjo (approved):
LG from my side

tree/ntuple/v7/test/ntuple_type_name.cxx
```diff
@@ -70,6 +70,7 @@ TEST(RNTuple, TClassDefaultTemplateParameter)
       model->AddField(RFieldBase::Create("f4", "struct DataVector<bool,vector<unsigned>>").Unwrap());
       model->AddField(RFieldBase::Create("f5", "DataVector<Double32_t>").Unwrap());
       model->AddField(RFieldBase::Create("f6", "DataVector<int, double>").Unwrap());
+      model->MakeField<DataVector<StructUsingCollectionProxy<int>>>("f7");
```
pcanal, line 73: I am still not finding a dictionary for `DataVector<StructUsingCollectionProxy<int>>` ... am I missing something or is that intentional?

vepadulano: Missed that again, apologies :)

PR #17994, opened 2025-03-13
Fix spurrious auto parsing when looking up a TClass

### LINKED ISSUE #17992: Spurrious auto parsing when looking up a TClass
### Check duplicate issues.

- [x] Checked for duplicates

### Description

As seen in https://github.com/cms-sw/cmssw/issues/47470 in some circumstances calling `TClass::GetClass` may lead to an unnecessary/unwanted loading of the header files.

### Reproducer

In the example, there is a lookup for `vector<edm::Ref<vector<l1t::TkElectron>,l1t::TkElectron,edm::refhelper::FindUsingAdvance<vector<l1t::TkElectron>,l1t::TkElectron> > >`
which, because it is a normalized class name lead to a code path where we do
1. auloading the library needed for this class and it dependencies
2. attempt to normalize the name and resolve potential typedefs
3. check for the dictionary.
However in this case `edm::refhelper::FindUsingAdvance<vector<l1t::TkElectron>,l1t::TkElectron> >` is intentionally not known to core/meta/TClass and thus the name normalization must check if it is a typedef and this step triggers the (auto) parsing of the headers (for `l1t::TkElectron`).

This can be solved by executing between 1 and 2:
1a. check for dictionary with the name as is.

### ROOT version

at least 6.32 up to master (probably more)

### Installation method

any

### Operating system

any

### Additional context

_No response_

pcanal:
This fixes #17992

Technically the fix is in commit 3812e5e8a3a0852b6e95b9516752f0ae0d73c59f

However analysis of the surrounding code showed that 
* one of the code path was missing a call to PostLoadCheck (74a7abbb77de51e9479d7bbb69e6213082fdbc2b)
* the code could be simplified a tad bit.

### DIFF
--- a/core/meta/src/TClass.cxx
+++ b/core/meta/src/TClass.cxx
@@ -3169,7 +3169,6 @@ TClass *TClass::GetClass(const char *name, Bool_t load, Bool_t silent, size_t hi
      }
    } else {
       normalizedName = cl->GetName(); // Use the fact that all TClass names are normalized.
-      checkTable = load && (normalizedName != name);
    }
 
    if (!load) return nullptr;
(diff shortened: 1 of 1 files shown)


dpiparo (approved):
Thanks Philippe for this important fix. LGTM.

PR #17999, opened 2025-03-14
[DF][ntuple] Enable RNTuple Snapshotting to `TDirectory`

enirolf:
This PR lifts the restriction of not being able to Snapshot to RNTuple in a TDirectory (i.e., `dir/ntuple`). I naively/wrongly assumed that it wasn't possible to read back RNTuples in sub-directories, but it actually just works fine.


vepadulano (approved):
Nice, thanks!

PR #18016, opened 2025-03-16
[ci] Use C++17 for mac14 to speedup building and testing

dpiparo:


.github/workflows/root-ci.yml
```diff
@@ -100,7 +100,7 @@ jobs:
             overrides: ["LLVM_ENABLE_ASSERTIONS=On", "builtin_zlib=ON"]
           - platform: mac14
             arch: X64
-            overrides: ["LLVM_ENABLE_ASSERTIONS=On", "CMAKE_CXX_STANDARD=20"]
+            overrides: ["LLVM_ENABLE_ASSERTIONS=On"]
```
pcanal, line 103: What's the difference in speed between C++20 and C++17?

guitargeek (approved):
Great idea!

.github/workflows/root-ci.yml
guitargeek: The difference is huge. Timing measurement still running, but I'll report on what I see on my system.

ferdymercury: https://www.reddit.com/r/cpp/comments/o94gvz/what_happened_with_compilation_times_in_c20/?rdt=59968

silverweed: @guitargeek did you manage to get any measurements yet?

guitargeek: Hi! Yes I have it now. Building on my Linux system with `-j32` and GCC 14.2.1.

```txt
C++20
real	22m36.392s
user	384m57.664s
sys	37m38.654s
```
```
C++17
real	20m9.631s
user	339m7.843s
sys	33m55.512s
```

So the "huge" difference I remembered is not that huge after all :laughing: Maybe the compiler was improved since the last time I compared, or my perception of time was not linear.

With `ccache` enabled, doing a fresh build after deleting the previous build directory I even see a smaller difference (6m50s vs. 6m39s real time).

guitargeek: @ferdymercury, your linked reddit thread is also 4 years old. Maybe the situation has really improved

ferdymercury: thanks! Did you check with latest master? Note that, since yesterday, LLVM is built with C++17 even if you specify C++20. So it would be interesting to check with LLVM also in C++20.

PR #18018, opened 2025-03-16
[Minuit2] Performance and thread safety improvements plus refactoring

guitargeek:
See commit descriptions.


math/minuit2/src/HessianGradientCalculator.cxx
```diff
@@ -83,6 +83,8 @@ HessianGradientCalculator::DeltaGradient(const MinimumParameters &par, const Fun
    unsigned int startElementIndex = mpiproc.StartElementIndex();
    unsigned int endElementIndex = mpiproc.EndElementIndex();
 
+   MnFcnCaller fcnCaller{fFcn};
```
guitargeek, line 86: This will be a performance improvement when using `strategy==2`, where the HessianGradientCalculator is used.

dpiparo (approved):

PR #18019, opened 2025-03-16
[RF] Avoid setting out-of-range value in NaNPacker tests

guitargeek:
This would clip to the nearest in-range value anyway, so we should pass the lower limit directly.


guitargeek:
Right now it will silently clip to the limit, bit I plan to change this to a warning or even an error. This PR was to avoid that this eventual error will be triggered in the RooFit tests.

PR #18020, opened 2025-03-16
[RF][RS] Fix StandardHistFactoryPlotsWithCategories tutorial

guitargeek:
The StandardHistFactoryPlotsWithCategories tutorial assumed that the nominal and plus-minus one sigma values of all nuisance parameters was zero, plus one, and minus one respectively. This is not true, especially for `gamma` parameters, where the nominal value is one.

The actual values and uncertainties should be taken from the parameters with `getVal()` and `getError()`. This also fixes the output plot in the tutorial for the gamma parameters:

https://root.cern/doc/master/StandardHistFactoryPlotsWithCategories_8C.html


lmoneta (approved):
LGTM!
Thank you Jonas for fixing this tutorial!

PR #18021, opened 2025-03-16
[Math] Move TimingScope to `ROOT::Math::Util`

guitargeek:
Move TimingScope to `ROOT::Math::Util` in MathCore to avoid code duplication while at the same time not introducing new dependencies (MathCore is the only dependency that Minuit2 is allowed to have because of the standalone build possibility).

This follows up on previous comments by @lmoneta.


math/mathcore/src/Util.cxx
```diff
@@ -0,0 +1,69 @@
+#include "Math/Util.h"
+
+#include <chrono>
+
+ROOT::Math::Util::TimingScope::TimingScope(std::function<void(std::string const &)> printer, std::string const &message)
+   : fPrinter{printer}, fMessage{message}
+{
+   using std::chrono::steady_clock;
+   fBegin = new steady_clock::time_point{steady_clock::now()};
```
silverweed, line 9: Why are we 
a) heap-allocating a `time_point`
b) storing it as a `void*` and
c) reinterpret-casting it back?

This seems weird and inefficient...

silverweed: Why is `<chrono>` less acceptable than, say `<functional>`? It seems a pretty innocent include to me

guitargeek: `<functional>` is used in other mathcore headers already. It's just that recently I got some flak for being the first to include `<any>` in some other part of ROOT and I wanted to avoid discussions about that. I agree with you that's not a well-founded argument. Let me change this PR to include `<chrono>` in the header and if @lmoneta has a problem with that he can still comment.

lmoneta (approved):
LGTM!
Thank you Jonas for this nice improvements and moving the TimingScope class to ROOT::Math::Util so it can be used by other ROOT packages

PR #18022, opened 2025-03-16
[CMake] Use `clangCppInterOp` directly as input for Cling target

guitargeek:
Use `clangCppInterOp` directly as input for Cling target, just like the other objects that are used (like `MetaCling`).

This fixes the following CMake generation error on my system:
```txt
CMake Error: install(EXPORT "ROOTExports" ...) includes target "Cling" which requires target "clangCppInterOp" that is not in any export set.
CMake Error in CMakeLists.txt:
  export called with target "Cling" which requires target "clangCppInterOp"
  that is not in any export set.
```


aaronj0 (approved):
LGTM, Thanks for this fix!

PR #18029, opened 2025-03-18
[RF] Fix tutorial category of `rf206_treevistools`

guitargeek:
I think this was an oversight due to conflict resolution.

That will fixup the RooFit tutorial page:
https://root.cern/doc/master/group__tutorial__roofit.html


martamaja10 (approved):
Thanks for the fix!

PR #18033, opened 2025-03-18
Provide default implementation of virtual method

vepadulano:
The SetNSlots method can now have a default implementation since the recent upstream of the fNSlots data member to the base class RDataSource.

Part 8 of N for https://github.com/root-project/root/pull/17895


martamaja10 (approved):
LGTM, thank you!

PR #18034, opened 2025-03-18
[df] Avoid creating readers for unavailable columns

vepadulano:
DefaultValueFor, FilterAvailable and FilterMissing register the column which might have missing values during the creation of a RDefaultValueFor or RFilterWithMissingValues object. Move the creation of the data source column readers after this is done, to avoid creating a reader for a column that might potentially be missing at that time.

Part 9 of N for https://github.com/root-project/root/pull/17895


martamaja10 (approved):
Thanks!

PR #18036, opened 2025-03-18
[hist] Improve TH1::GetQuantiles

hageboeck:
- Improve documentation of TH1::GetQuantiles and TF1::GetQuantiles
- Remove manual allocation and a const_cast in `TH1::GetQuantiles`.

This PR is an attempt to fix #16793, which proposed to remove the default parameter of `TH1::GetQuantiles`. As discussed there, the removal is not a good idea.
The existing documentation also created confusion about what the function does when the last parameter is left at its default value, which is clarified here.

So TL;DR: Don't change the default parameter, but clarify what it does.

Fix #16793

### DIFF
--- a/hist/hist/src/TH1.cxx
+++ b/hist/hist/src/TH1.cxx
@@ -4507,8 +4507,8 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-/// Compute Quantiles for this histogram
-/// Quantile x_p := Q(p) is defined as the value x_p such that the cumulative
+/// Compute Quantiles for this histogram.
+/// A quantile x_p := Q(p) is defined as the value x_p such that the cumulative
 /// probability distribution Function F of variable X yields:
 ///
 /// ~~~ {.cpp}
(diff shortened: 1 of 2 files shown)


hist/hist/src/TH1.cxx
```diff
@@ -4527,16 +4528,18 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
 /// \author Eddy Offermann
 /// code from Eddy Offermann, Renaissance
 ///
-/// \param[in] n maximum size of array xp and size of array p (if given)
+/// \param[in] n maximum size of the arrays xp and p (if given)
 /// \param[out] xp array to be filled with nq quantiles evaluated at (p). Memory has to be preallocated by caller.
-/// If p is null (default value), then xp is actually set to the (first n) histogram bin edges
+///   - If `p == nullptr`, the quantiles are computed at the probabilities p given by the CDF of the histogram.
```
ferdymercury, line 4533:
```suggestion
///   - If `p == nullptr`, the quantiles are computed at the (first `n`) probabilities p given by the CDF of the histogram.
```

hist/hist/src/TH1.cxx
```diff
@@ -4527,16 +4528,18 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
...
-///   - if p is null, the CDF of the histogram will be used instead as array, and will
-///     have a size = number of bins + 1 in h. It will correspond to the
-///     quantiles calculated at the lowest edge of the histogram (quantile=0) and
-///     all the upper edges of the bins. (nbins might be > n).
-///   - if p is not null, it is assumed to contain at least n values.
-/// \return value nq (<=n) with the number of quantiles computed
+///   - if `p == nullptr`, the CDF of the histogram will be used to compute the quantiles, and will
+///     have a size of N_bins + 1.
+///   - Otherwise, it is assumed to contain at least n values.
+/// \return number of quantiles computed
```
ferdymercury, line 4542:
```suggestion
/// \return number of quantiles computed
/// \note Unlike in TF1::GetQuantiles, p is here is an optional argument
```

hist/hist/src/TH1.cxx
```diff
@@ -4527,16 +4528,18 @@ TVirtualHistPainter *TH1::GetPainter(Option_t *option)
...
+///     at 3.
 /// \param[in] p array of cumulative probabilities where quantiles should be evaluated.
-///   - if p is null, the CDF of the histogram will be used instead as array, and will
-///     have a size = number of bins + 1 in h. It will correspond to the
-///     quantiles calculated at the lowest edge of the histogram (quantile=0) and
-///     all the upper edges of the bins. (nbins might be > n).
-///   - if p is not null, it is assumed to contain at least n values.
-/// \return value nq (<=n) with the number of quantiles computed
+///   - if `p == nullptr`, the CDF of the histogram will be used to compute the quantiles, and will
+///     have a size of N_bins + 1.
```
ferdymercury, line 4540:
```suggestion
///     have a size of n.
```
See proof here https://github.com/root-project/root/pull/16793#discussion_r1829739084

ferdymercury (changes requested):

hist/hist/src/TH1.cxx
hageboeck: Makes total sense!

hageboeck:
Thanks @ferdymercury,
your suggestions are already squashed into the second commit.

PR #18038, opened 2025-03-18
Improve SavePrimitive in TG classes

linev:
1. Introduce `TGFrame::SaveCrtrArgs` method to prepare options and custom color for use in saved constructor. There are many places in code where such combination handled with many if statements
2. Use new method where it make sence
3. Replace std::endl by `\n` and quote by `\"`
4. ~~Fix several problems where with custom colors also font and `uGC->GetGC()` must be initialized. Otherwise when saved into the macro arguments will remain empty~~, not necessary
5. Add to TGMainFrame::SavePrimitive several commands to popup window
6. Always use TString::ReplaceSpecialCppChars when saving text or tooltips
7. Simplify implementation of `TGFrame::GetOptionString()`

Tested with several macros from tutorials like calendar.C or listBox.C or statusBar.C

### DIFF
--- a/core/base/src/TString.cxx
+++ b/core/base/src/TString.cxx
@@ -1113,7 +1113,7 @@ TString& TString::ReplaceAll(const char *s1, Ssiz_t ls1, const char *s2,
 
 TString &TString::ReplaceSpecialCppChars()
 {
-   return ReplaceAll("\\","\\\\").ReplaceAll("\"","\\\"");
+   return ReplaceAll("\\","\\\\").ReplaceAll("\"","\\\"").ReplaceAll("\n","\\n").ReplaceAll("\t","\\t");
 }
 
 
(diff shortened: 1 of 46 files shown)


bellenot (approved):
LGTM, thanks. FYI, one way of testing if it works properly would be to run `stressgui` before and after these changes

bellenot:
> stressGUI creates macros but not execute them.
> And created size is not checked.

One should execute `stressgui -ref` to generate the `.ref` file and then `stressgui` to run the test

linev:
> One should execute stressgui -ref to generate the .ref file and then stressgui to run the test

Is all about PNG image sizes.
But created ROOT macro is not checked and not executed.
I already find more bugs and will fix them with following PR

bellenot:
Oh yes, sorry, you're right...

PR #18040, opened 2025-03-19
[ntuple] Move Descriptor classes out of Experimental

silverweed:
# This Pull request:
Moves the following classes out of Experimental:
- `RNTupleDescriptor`
- `R(Field|Column|Cluster|ClusterGroup|ExtraTypeInfo)Descriptor`
- Their Builders

## Checklist:

- [x] tested changes locally

This PR fixes #

### DIFF
--- a/tree/dataframe/inc/ROOT/RNTupleDS.hxx
+++ b/tree/dataframe/inc/ROOT/RNTupleDS.hxx
@@ -56,7 +56,7 @@ class RNTupleDS final : public ROOT::RDF::RDataSource {
    };
 
    /// A clone of the first pages source's descriptor.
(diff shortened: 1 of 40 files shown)


hahnjo (approved):
I believe the dependency tree is `RNTupleDescriptor` -> `RNTupleModel` -> `REntry`, so in principle they should have moved first. But LGTM as we have it right now. Do we want `using` declarations or do we assume that the classes are not (widely) used at the moment?

silverweed:
> Do we want using declarations or do we assume that the classes are not (widely) used at the moment?

I think we want them, but I forgot to add them. I will do it.

PR #18042, opened 2025-03-19
[CMake] Raise minimum CMake version for Cling to 3.10

guitargeek:
The recent cmake version 3.31.5 gives the following warning:

```txt
CMake Deprecation Warning at interpreter/cling/CMakeLists.txt:8 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.
```

Raise the minimum version to CMake 3.10, which was released in 2017: https://www.kitware.com/cmake-3-10-0-available-for-download/

Whoever is reviewing this, please also take a look at this related `roottest` PR with the same branch name so they are tested together:
https://github.com/root-project/roottest/pull/1284

### DIFF
--- a/interpreter/cling/CMakeLists.txt
+++ b/interpreter/cling/CMakeLists.txt
@@ -1,12 +1,4 @@
-if(WIN32)
-  # We need cmake to support exporting of symbols not only from libraries but
-  # from executables too. This way cling can find symbols from its own
-  # executable during runtime.
-  cmake_minimum_required(VERSION 3.6.2)
-else(WIN32)
-  # support of earlier cmake versions will be removed soon
-  cmake_minimum_required(VERSION 3.5)
-endif(WIN32)
+cmake_minimum_required(VERSION 3.10)
 
 # If we are not building as a part of LLVM, build Cling as an
 # standalone project, using LLVM as an external library:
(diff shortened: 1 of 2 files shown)


bellenot (approved):
Fine with me once the builds are green

interpreter/cling/CMakeLists.txt
```diff
@@ -1,12 +1,4 @@
-if(WIN32)
-  # We need cmake to support exporting of symbols not only from libraries but
-  # from executables too. This way cling can find symbols from its own
-  # executable during runtime.
-  cmake_minimum_required(VERSION 3.6.2)
-else(WIN32)
-  # support of earlier cmake versions will be removed soon
-  cmake_minimum_required(VERSION 3.5)
-endif(WIN32)
+cmake_minimum_required(VERSION 3.10)
```
pcanal, line 1: What is the default CMake version for the oldest distribution we support?

guitargeek: It's 3.26.5 on `alma8` and `3.22.1` on Ubuntu 22.04, so we're safe there. Note that Ubuntu 20.04 has version 3.16.3, but that distribution will be EOL by the time ROOT 6.36 comes out

PR #18046, opened 2025-03-19
[webgui] exclude old headless mode for newest chrome browser

linev:
From versions 120 to 132 chrome supports two different kinds of headless modes, but finally remove old in version 132.

https://developer.chrome.com/blog/removing-headless-old-from-chrome

Therefore use plain headless mode in newest Chrome. 
Do it by default on Windows where version is not tested.


bellenot (approved):
Thanks Sergey! LGTM, once all builds are green 😉

PR #18049, opened 2025-03-19
[ntuple] Implement `Detail::RRawPtrWriteEntry`

### LINKED ISSUE #17900: [ntuple] Add write API with const pointers
During the HEP-CCE review and when implementing the RNTuple prototype in CMSSW, it became clear that the current `REntry` interface doesn't work for writing `const` objects (via raw pointers). This was confirmed by ATLAS in yesterday's RNTuple planning discussion. As `REntry` and `RFieldBase::RValue` heavily rely on shared ownership of the objects, a possibility would be to introduce a separate `RConstEntry` (name to be discussed) that only works for writing and acts as a sort of container for raw pointers only.

hahnjo:
It is a container of `const` raw pointers that can be used to write constant data products in frameworks.

Closes #17900

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RFieldToken.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RFieldToken.hxx
@@ -24,6 +24,12 @@ namespace ROOT {
 class REntry;
 class RNTupleModel;
 
+namespace Experimental {
+namespace Detail {
+class RRawPtrWriteEntry;
+} // namespace Detail
+} // namespace Experimental
+
 // clang-format off
 /**
 \class ROOT::RFieldToken
@@ -36,8 +42,9 @@ It can be used for fast indexing in REntry's methods, e.g. REntry::BindValue(). 
 class RFieldToken {
    friend class REntry;
    friend class RNTupleModel;
+   friend class Experimental::Detail::RRawPtrWriteEntry;
 
-   std::size_t fIndex = 0;                      ///< The index in `fValues` that belongs to the field
+   std::size_t fIndex = 0;                      ///< The index of the field (top-level or registered subfield)
    std::uint64_t fSchemaId = std::uint64_t(-1); ///< Safety check to prevent tokens from other models being used
    RFieldToken(std::size_t index, std::uint64_t schemaId) : fIndex(index), fSchemaId(schemaId) {}
 
--- a/tree/ntuple/v7/inc/ROOT/RFieldToken.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RFieldToken.hxx
@@ -1,0 +1,50 @@
+/// \file ROOT/RFieldToken.hxx
+/// \ingroup NTuple ROOT7
+/// \author Jonas Hahnfeld <jonas.hahnfeld@cern.ch>
+/// \date 2025-03-19
+/// \warning This is part of the ROOT 7 prototype! It will change without notice. It might trigger earthquakes. Feedback
+/// is welcome!
+
+/*************************************************************************
+ * Copyright (C) 1995-2025, Rene Brun and Fons Rademakers.               *
+ * All rights reserved.                                                  *
+ *                                                                       *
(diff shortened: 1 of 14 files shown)


tree/ntuple/v7/inc/ROOT/RFieldToken.hxx
```diff
@@ -0,0 +1,56 @@
...
+class REntry;
+class RNTupleModel;
+
+namespace Detail {
+class RRawPtrWriteEntry;
+} // namespace Detail
+
+// clang-format off
+/**
+\class ROOT::Experimental::RFieldToken
```
hahnjo, line 33: Shall I move the class out of `Experimental` directly? @silverweed

silverweed: I think so. But maybe we can double check with @jblomer to see if he has any objection to this.

hahnjo: I rebased the PR, resolved the conflicts, and moved to `ROOT::RFieldToken`.

jblomer (approved):
Thanks! LGTM!

tree/ntuple/v7/inc/ROOT/RFieldToken.hxx
```diff
@@ -36,8 +42,9 @@ It can be used for fast indexing in REntry's methods, e.g. REntry::BindValue().
 class RFieldToken {
    friend class REntry;
    friend class RNTupleModel;
+   friend class Experimental::Detail::RRawPtrWriteEntry;
 
-   std::size_t fIndex = 0;                      ///< The index in `fValues` that belongs to the field
+   std::size_t fIndex = 0;                      ///< The index of the top-level field
```
enirolf, line 47: With the addition of `REntry::RegisterSubfield`, this doesn't necessarily only have to be a top-level field.

tree/ntuple/v7/inc/ROOT/RNTupleWriter.hxx
```diff
@@ -110,6 +111,18 @@ public:
    /// Fill an entry into this ntuple, but don't commit the cluster. The calling code must pass an RNTupleFillStatus
    /// and check RNTupleFillStatus::ShouldFlushCluster.
    void FillNoFlush(ROOT::REntry &entry, RNTupleFillStatus &status) { fFillContext.FillNoFlush(entry, status); }
+
+   /// Fill an RRawPtrWriteEntry into this ntuple.  This method will perform
+   /// a light check whether the entry comes from the ntuple's own model
```
enirolf, line 116: What exactly does "light check" mean? Is there something that's not checked that the caller should take of themselves?

enirolf (approved):
LGTM! I added two minor documentation-related comments but otherwise good to go.

tree/ntuple/v7/inc/ROOT/RNTupleWriter.hxx
silverweed: I think it just means that the check itself is not computationally expensive. The caller should not have to do specific checks in the code (unless they are weirdly mixing up entries from different models).

tree/ntuple/v7/inc/ROOT/RFieldToken.hxx
```diff
@@ -0,0 +1,57 @@
+/// \file ROOT/RFieldToken.hxx
+/// \ingroup NTuple ROOT7
+/// \author Jonas Hahnfeld <jonas.hahnfeld@cern.ch>
+/// \date 2015-03-19
```
vepadulano, line 4:
```suggestion
/// \date 2025-03-19
```

vepadulano (approved):
Thanks!

tree/ntuple/v7/inc/ROOT/RNTupleWriter.hxx
hahnjo: I amended the documentation to talk about the model ID, and mention that the method throws an exception otherwise.

PR #18054, opened 2025-03-19
Improve stressGUI, fix TG SavePrimitive methods

linev:
In stressGUI disable web mode and with `-keep` flag also preserve created macro files

Running all these macro, found and fix several old and new errors.
Now all these macros are usable


bellenot (approved):
Thanks Sergey!

PR #18057, opened 2025-03-19
Fix/improve SavePrimitive in g3d classes

linev:
Improve `TPolyLine3D` and `TMarker3DBox`. 

Replace `THelix::SavePrimitive`. 
Implementation is broken and not usable.
For real implementation many changes are required.

### DIFF
--- a/graf3d/g3d/src/THelix.cxx
+++ b/graf3d/g3d/src/THelix.cxx
@@ -175,28 +175,6 @@ THelix::THelix(Double_t const* xyz, Double_t const* v, Double_t w,
 }
 
 
-#if 0
-////////////////////////////////////////////////////////////////////////////////
-/// Helix copy constructor.
-
-THelix::THelix(const THelix &h) : TPolyLine3D()
-{
-   fX0   = h.fX0;
-   fY0   = h.fY0;
-   fZ0   = h.fZ0;
-   fVt   = h.fVt;
-   fPhi0 = h.fPhi0;
-   fVz   = h.fVz;
-   fW    = h.fW;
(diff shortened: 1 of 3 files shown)


graf3d/g3d/src/THelix.cxx
```diff
@@ -293,30 +263,6 @@ void THelix::Print(Option_t *option) const
 }
 
 
-////////////////////////////////////////////////////////////////////////////////
-/// Save primitive as a C++ statement(s) on output stream out.
-
-void THelix::SavePrimitive(std::ostream &out, Option_t * /*= ""*/)
```
pcanal, line 299: What happens if someone calls `SavePrimitives` on a `THelix` after that?  Is it a silent failure or should be add an Error message here instead?

linev: User will get message in the produced C macro that SavePrimitive is not implemented for that class

couet (approved):

graf3d/g3d/src/THelix.cxx
pcanal: By what mechanism?  Isn't `THelix` inheriting from `TPolyLine3D` and thus `TPolyLine3D::SavePrimitives` will be called?

linev: Ah yes. Then I will put dummy here.

linev: I add alternative implementation. It works somehow, but is not fully cover functionality of THelix.
At least random instance like `THelix *helix = new THelix(5.2, 4.8, 5, 1, 1, 1, 5);` is reproduced.

PR #18061, opened 2025-03-19
[ci][6.32] Remove Ubuntu 20 from the battery of platforms

dpiparo:


dpiparo:
Thanks for the meta-approval.

PR #18063, opened 2025-03-19
[RF] Dump debug macro from codegen on compilation failure

guitargeek:
Dumping the generated code in case it fails to compile greatly helps in development and debugging of the new RooFit code generation backend.

This will be useful during the upcoming CMS CAT hackathon, where we work on supporting the models from CMS Combine with the codegen backend.

Also, run parts of the `stressRooFit` test suite with the codegen backend, which increases test coverage and also allows us to remove a redundant test.


Michealinc (approved):

lcanal00 (dismissed):

lmoneta (approved):
LGTM!

PR #11721, opened 2022-11-16
[tmva][sofie] Exclude SOFIE test when blas is not found

lmoneta:
This Pull request exclude running of SOFIE tests when Blas is not found (this is when `tmva-cpu=Off`)

PR #18064, opened 2025-03-19
[tmva][sofie] Don't make sofie tests conditional on `tmva-cpu`

guitargeek:
In principle, `tmva-cpu` and `tmva-sofie` are separate features, and they should not be entangled.

Since PR #11721, the `tmva-cpu` variable is used as a proxy to detect whether BLAS is on the system or not, to conditionally disable some SOFIE tests. This means one would lose SOFIE test coverage when disabling `tmva-cpu`, which is not necessary.

This PR suggests to always build the SOFIE tests that require BLAS.

Hopefully, the new CI has BLAS installed anyway  in all environments that test `tmva-sofie`, then we don't even need an alternative mechanism to disable the tests that require it (the alternative mechanism would be a separate check if BLAS is available, independently of `tmva-cpu`).

### DIFF
--- a/tmva/sofie/test/CMakeLists.txt
+++ b/tmva/sofie/test/CMakeLists.txt
@@ -60,7 +60,6 @@ add_custom_command(TARGET SofieCompileModels_ONNX POST_BUILD
 
 
 # Creating a Google Test
-if (tmva-cpu)  # we need BLAS for compiling the models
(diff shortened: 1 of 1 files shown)


guitargeek:
I think I have a better solution:
   * https://github.com/root-project/root/pull/18065

PR #15439, opened 2024-05-07
[TMVA] Enable again all tests on `alma9` and `fedora39`

guitargeek:
Now that we're using a system-compatible version of NumPy on the CI images for these platforms, the TMVA tests should all pass again.

Issue #16720, opened 2024-10-19
4 TMVA test requires BLAS library but run even when it was not found.


pcanal:
See https://github.com/root-project/root/issues/16553

The following 3 tests fails consistently with missing BLAS symbols
```
984:tutorial-tmva-TMVA_SOFIE_GNN_Application
988:tutorial-tmva-TMVA_SOFIE_RDataFrame
990:tutorial-tmva-TMVA_SOFIE_RSofieReader
353-gtest-tmva-pymva-TestRModelParserKeras
```

dpiparo:
Unassigning and reassigning to @lmoneta

dpiparo:
I am not sure I understand the problem.
If I look at `984:tutorial-tmva-TMVA_SOFIE_GNN_Application`, the tests run fine in the CI for alma9: the reason for the failure could be due to the local setup.

As for
```
988:tutorial-tmva-TMVA_SOFIE_RDataFrame
990:tutorial-tmva-TMVA_SOFIE_RSofieReader
353-gtest-tmva-pymva-TestRModelParserKeras
```
If I understand correctly, they do not run in any of platforms of the CI since PyKeras is not available anywhere, e.g. in the `requirements.txt` file.

dpiparo:
Apologies, I said something in error. All the aforementioned tests can be found successfully running in the CI. I think we are dealing with a difference between the container setup of the CI and the local setup.
@pcanal could you double-check your setup against the one of the container (see [here](https://github.com/root-project/root-ci-images/blob/main/alma9/packages.txt) and the requirements.txt file)

pcanal:
Yes, it is clearly a difference is setup.  In the failing one the `BLAS` library is not installed (or not found) and the real problem is that this result in a very obscure and hard to understand failure:
```
IncrementalExecutor::executeFunction: symbol 'saxpy_' unresolved while linking [cling interface function]!
IncrementalExecutor::executeFunction: symbol 'sgemm_' unresolved while linking [cling interface function]!
```
Instead we should either
* fail to configure if BLAS is missing
or 
* disable the components that require BLAS when configuring whe BLAS is missing
or
* disable the test that require BLAS when BLAS is not there.

and at the very least we should list somewhere (if not already done so, but I don't see it) that the `BLAS` library is required to successfully run the tests and some of the features.

Issue #16719, opened 2024-10-19
PyTorch test/tutorials are (likely) using the same model files.


pcanal:
Doing:
```
ctest -R tmva -j 32
```
will result in an arbitrary result (sometimes pass sometime fail) for
```
gtest-tmva-pymva-TestRModelParserKeras
gtest-tmva-pymva-TestRModelParserPyTorch 
```
re-running just those tests (whether they succeeded or not) will lead to both of them failing.
The failure report is indicate that they 'now' need the BLAS library (which is not available on the system).

As a possible clue (or not), the following 3 test fails systemically on the system due to the missing BLAS library:
```
        996 - tutorial-tmva-TMVA_SOFIE_GNN_Application (Failed)
        1000 - tutorial-tmva-TMVA_SOFIE_RDataFrame (Failed)
        1002 - tutorial-tmva-TMVA_SOFIE_RSofieReader (Failed)
```

pcanal:
It is confirms that one of those files:
```
-rw-r--r--. 1 pcanal us_cms   8962 Oct 19 17:56 ./tmva/pymva/test/PyTorchModelSequential.pt
-rw-r--r--. 1 pcanal us_cms  11913 Oct 19 17:56 ./runtutorials/modelClassification.pt
-rw-r--r--. 1 pcanal us_cms  10564 Oct 19 17:56 ./runtutorials/PyTorchModel.pt
-rw-r--r--. 1 pcanal us_cms  10941 Oct 19 17:56 ./runtutorials/modelMultiClass.pt
-rw-r--r--. 1 pcanal us_cms  11330 Oct 19 17:56 ./runtutorials/trainedModelMultiClass.pt
-rw-r--r--. 1 pcanal us_cms  12110 Oct 19 17:56 ./runtutorials/trainedModelClassification.pt
-rw-r--r--. 1 pcanal us_cms   7853 Oct 19 17:57 ./runtutorials/modelRegression.pt
-rw-r--r--. 1 pcanal us_cms   7972 Oct 19 17:57 ./runtutorials/trainedModelRegression.pt
-rw-r--r--. 1 pcanal us_cms  11044 Oct 19 18:02 ./tmva/pymva/test/PyTorchModelModule.pt
-rw-r--r--. 1 pcanal us_cms   8337 Oct 19 18:02 ./tmva/pymva/test/PyTorchModelConvolution.pt
[...]

Issue #16553, opened 2024-09-27
[tmva] Missing dependency or clean up in TMVA test/tutorials


pcanal:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

On a large node (127 cores, 128 GB), I ran:
1. ctest -j 32
2. ctest --rerun-failed
3. ctest -j 32

After 1. many test failes due to lack of resources (running out of threads, see #16552 ):
```
47:PyMVA-Keras-Classification
348:PyMVA-Keras-Regression
349:PyMVA-Keras-Multiclass
350:gtest-tmva-pymva-test-TestRModelParserKeras
984:tutorial-tmva-TMVA_SOFIE_GNN_Application
985:tutorial-tmva-TMVA_SOFIE_Keras
986:tutorial-tmva-TMVA_SOFIE_Keras_HiggsModel
988:tutorial-tmva-TMVA_SOFIE_RDataFrame
990:tutorial-tmva-TMVA_SOFIE_RSofieReader
1238:tutorial-tmva-RBatchGenerator_PyTorch-py
1239:tutorial-tmva-RBatchGenerator_TensorFlow-py
1246:tutorial-tmva-TMVA_SOFIE_Models-py
1247:tutorial-tmva-TMVA_SOFIE_RDataFrame-py
1252:tutorial-tmva-keras-GenerateModel-py
1253:tutorial-tmva-keras-MulticlassKeras-py
```
However in 2., several tests still failed (even-though resources where no longer an issue):
```
50:gtest-tmva-pymva-test-TestRModelParserKeras
984:tutorial-tmva-TMVA_SOFIE_GNN_Application
986:tutorial-tmva-TMVA_SOFIE_Keras_HiggsModel
988:tutorial-tmva-TMVA_SOFIE_RDataFrame
990:tutorial-tmva-TMVA_SOFIE_RSofieReader
1247:tutorial-tmva-TMVA_SOFIE_RDataFrame-py
```
The errors listed there included:
```
IncrementalExecutor::executeFunction: symbol 'saxpy_' unresolved while linking [cling interface function]!
IncrementalExecutor::executeFunction: symbol 'sgemm_' unresolved while linking [cling interface function]!
tutorials/tmva/TMVA_SOFIE_RDataFrame.C:29:10: fatal error: 'Higgs_trained_model.hxx' file not found
/tutorials/tmva/TMVA_SOFIE_GNN_Application.C:10:10: fatal error: 'encoder.hxx' file not found
```

[...]

PR #18065, opened 2025-03-19
[TMVA] Only consider GSL BLAS for `tmva-cpu`

### LINKED ISSUE #18055: misterious dependency on `mkl` for conda+ROOT 6.34
### Check duplicate issues.

- [x] Checked for duplicates

### Description

For my (Py)ROOT-based project for the last couple of months I actively use  ROOT via conda.

```
conda create --name test-env root python ipython cmake ninja
conda activate tets-env 
... build my project... 
```
Very recently conda provides ROOT 6.34 and when I' try to use usin absolutely the same machinery 
with no changes in my project, my project builds, but I see the failure in manual loading the library:  
```
In [7]: import ROOT
In [8]: ROOT.gSystem.Load('./INSTALL/lib/libostap.so')
cling::DynamicLibraryManager::loadLibrary(): libmkl_intel_lp64.so.2: cannot open shared object file: No such file or directory
Out[8]: -1

In [9]: ROOT.gSystem.Load('./INSTALL/lib/libostap.so')
cling::DynamicLibraryManager::loadLibrary(): libmkl_intel_lp64.so.2: cannot open shared object file: No such file or directory
Out[9]: -1


```
Note that my project has no dependency on `mkl`. (I even have no idea what it is  where it comef from)

If I roolback to 'old' ROOT 6.32 I can build and successfully use my project:

```
conda create --name test-env root=6.32  python ipython cmake ninja
conda activate tets-env 
... build my project... 
```

It looks like the strange dependency on `` comes through ROOT.

I see no problems when  use ROOT from LCG cvmfs nightlues - both for released versions and for dev3-nightly builds.  The problem comes only from 'ROOT 6.32+conda' bundle.

What should be the soltuion ?

    Thank you veyr much in advance
  

conda create --name test-env root python ipython cmake ninja

### Reproducer

My project is too heavy., I do nto know how to make a simple reproducer
I'll make a try, but it cudl take a time. 

### ROOT version

ROOT 6.34 from conda.

There are no problems with ROOT from LCG cvmfs area (reliease and dev3-nightlies
and there are no problems for conda+ROOT 6.32 

 

### Installation method

conda 

### Operating system

Linux

### Additional context

_No response_

guitargeek:
Using the GSL BLAS should avoid collisions with other loaded BLAS libraries, so we can hopefully enable `tmva-cpu` again for all platforms and we don't have to deal with random BLAS related problems anymore. GSL is already a dependency of mathmore, and so far it has been an unproblematic dependency, also when using ROOT from Python.

### What we gain

* No dependency of ROOT on specialized BLAS implementations anymore, avoiding problems like:
  * symbol clashes with different BLAS libraries, e.g. incompatibility between NumPy and flexiblas that has to be [worked around by setting the `BLA_VENDOR` eplicitly](https://github.com/root-project/root/pull/15439)
  * [unexpected MKL dependencies](https://github.com/root-project/root/issues/18055)
  * Problems when [importing TensorFlow and ROOT from Conda](https://github.com/conda-forge/root-feedstock/issues/253)
  * I have had other personal conversations with users reporting similar issues when using ROOT in Python
* More predictable builds, because it's not up to CMake anymore which BLAS to link against
* Less dependencies overall: GSL is already a dependency for a typical ROOT build with `mathmore=ON`
* We can now enable `tmva-cpu` on all CI platforms, because when using the GSL CBLAS there are no problems
* More of a side effect from disentangling `tmva-sofie` from `tmva-cpu`: we get increased test coverage on TMVA SOFIE as well (previously, some tests were not run when `tmva-cpu` was disabled)

### What we lose

* Performance in the TMVA deep learning CPU backend. But given that people nowadays use mostly GPUs to train neural networks, this can be acceptable

Note that we *don't* lose any test coverage, because we're just changing the BLAS library used under the hood! That means no code will rot and it would also be easy change back to an optimized BLAS if desired.

### Neutral

* The `tmva-cpu` feature depends on GSL, but that should be acceptable given that `mathmore` already depends on GSL, and we always want to enable `mathmore`.

Closes #18055.

Note that this is basically a follow up to 582583e5dfa from 2018, where the possibility to use GSL BLAS was added. Now it would just be the only possible choice.


cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -1605,6 +1605,13 @@ endif()
 #---Check for protobuf-------------------------------------------------------------------
 
 if(tmva-sofie)
+  if(testing)
+    message(STATUS "Looking for BLAS as an optional testing dependency of TMVA-SOFIE")
+    find_package(BLAS)
```
vepadulano, line 1610: For SOFIE it is still fine to look for any BLAS backend?

ferdymercury: A bit related:
https://github.com/root-project/root/issues/16720
https://github.com/root-project/root/issues/16719
https://github.com/root-project/root/issues/16553

lmoneta (changes requested):
Will it still be possible to set the BLAS version? I understand for convenience it is better to use as default GSL BLAS, but it would be nice to tell to cmake to use a specific BLAS version (of available), given the poor performances of the GSL BLAS implementation

guitargeek:
Hi @lmoneta! Thanks for you comment, you're right it's good to be bale to switch to an optimized BLAS.

I have suggested a new option `use_gsl_cblas` for that, which is `ON` by default. What do you think?

dpiparo (approved):

guitargeek:
It failed on `mac15` because `onnx` the Python package was not installed there. I logged in to it and installed it with `pip`, so it should work in the next test run.

lmoneta (approved):
LGTM!
Thank you Jonas for fixing this issue and implementing the option to still use optimised BLAS

PR #18069, opened 2025-03-20
[ntuple] Use newer `boost_combine` implementation in `RNTupleJoinTable`

enirolf:
The previous implementation came from Boost v1.33. As explained in https://www.boost.org/doc/libs/1_81_0/libs/container_hash/doc/html/hash.html#notes_hash_combine, several improvements have been introduced until v1.81 to improve the distribution of the values and reduce the chance of collisions.

To see the impact of this change, I've created a small benchmark comparing the v1.33 implementation against the v1.81 implementation on the combination of the `run`, `lumiBlock` and `event` fields of one of the AGC NanoAODs (find the code [here](https://github.com/enirolf/ntuple_join_field_hashing)):

hash_distributionv1_33

hash_distributionv1_81

Clearly, the new implementation gives a much better hash distribution. In addition, in this particular case, the old implementation lead to 508 hash collisions (for ~1.2M, so not dramatic, but still..), whereas the new implementation has none.


tree/ntuple/v7/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -54,15 +54,62 @@ private:
    /////////////////////////////////////////////////////////////////////////////
    /// Hash combining the individual join field value hashes from RCombinedJoinFieldValue. Uses the implementation from
    /// `boost::hash_combine` (see
-   /// https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine).
+   /// https://www.boost.org/doc/libs/1_87_0/libs/container_hash/doc/html/hash.html#ref_hash_combine).
+   ///
```
pcanal, line 58:
```suggestion
   /// And in particular it mentions "Several improvements of the 64 bit function have been subsequently proposed, by [David Stafford](https://zimbry.blogspot.com/2011/09/better-bit-mixing-improving-on.html), [Pelle Evensen](https://mostlymangling.blogspot.com/2019/12/stronger-better-morer-moremur-better.html), and [Jon Maiga](http://jonkagstrom.com/mx3/mx3_rev2.html)." We currently use Jon Maiga’s function.
```

Michealinc (dismissed):

hahnjo (approved):
LGTM

Issue #18013, opened 2025-03-15
[RF] RooFormula doesn't complain if you use input called `x` but don't supply it


guitargeek:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

See reproducer.

### Reproducer

```c++
RooRealVar B("B", "", 0.516952);
RooRealVar r("r", "", 0.214107);
RooRealVar x("x", "", 0.2, 1);
RooRealVar y("y", "", 0.2, 1);

// should not work but does!
RooFormulaVar f1("f1", "r + B + x", {r, B});

// correctly complains
RooFormulaVar f2("f2", "r + B + y", {r, B});
```

### ROOT version

any

### Installation method

any

### Operating system

any

### Additional context

If this bug would be not present, this user would not have been confused why his code doesn't work:
https://root-forum.cern.ch/t/non-convergence-of-fit/63117/5

PR #18070, opened 2025-03-20
[RF] correctly detect undefined variables in RooArgList

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/18013

## Checklist:

- [x] tested changes locally

### DIFF
--- a/roofit/roofitcore/src/RooFormula.cxx
+++ b/roofit/roofitcore/src/RooFormula.cxx
@@ -184,6 +184,41 @@ void replaceVarNamesWithIndexStyle(std::string &formula, RooArgList const &varLi
    }
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// From the internal representation, construct a formula by replacing all index place holders
+/// with the names of the variables that are being used to evaluate it, and return it as string.
+std::string reconstructFormula(std::string internalRepr, RooArgList const& args) {
+  const auto nArgs = args.size();
+  for (unsigned int i = 0; i < nArgs; ++i) {
+    const auto& var = args[i];
+    std::stringstream regexStr;
+    regexStr << "x\\[" << i << "\\]|@" << i;
+    std::regex regex(regexStr.str());
+
+    std::string replacement = std::string("[") + var.GetName() + "]";
+    internalRepr = std::regex_replace(internalRepr, regex, replacement);
+  }
+
+  return internalRepr;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+/// From the internal representation, construct a null-formula by replacing all
+/// index place holders with zeroes, and return it as string
+std::string reconstructNullFormula(std::string internalRepr, RooArgList const& args) {
+  const auto nArgs = args.size();
+  for (unsigned int i = 0; i < nArgs; ++i) {
+     std::stringstream regexStr;
+     regexStr << "x\\[" << i << "\\]|@" << i;
+     std::regex regex(regexStr.str());
+
+     std::string replacement = "1e-18";
+     internalRepr = std::regex_replace(internalRepr, regex, replacement);
+  }
+
+  return internalRepr;
+}
+
 }
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -218,7 +253,6 @@ RooFormula::RooFormula(const RooFormula& other, const char* name) :
 
   _tFormula = std::move(newTF);
 }
-
 
 ////////////////////////////////////////////////////////////////////////////////
 /// Process given formula by replacing all ordinal and name references by
@@ -284,7 +318,6 @@ std::string RooFormula::processFormula(std::string formula) const {
   return formula;
 }
 
-
 ////////////////////////////////////////////////////////////////////////////////
 /// Analyse internal formula to find out which variables are actually in use.
 RooArgList RooFormula::usedVariables() const {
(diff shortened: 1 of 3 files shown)


roofit/roofitcore/src/RooFormula.cxx
```diff
@@ -331,6 +331,22 @@ std::string RooFormula::reconstructFormula(std::string internalRepr) const {
   return internalRepr;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// From the internal representation, construct a null-formula by replacing all
+//  index place holders with zeroes
+std::string RooFormula::reconstructNullFormula(std::string internalRepr) const {
+   for (unsigned int i = 0; i < _origList.size(); ++i) {
+      const auto& var = _origList[i];
```
ferdymercury, line 339:
```suggestion
```

roofit/roofitcore/test/testRooFormula.cxx
```diff
@@ -77,3 +77,15 @@ TEST(RooFormula, UnusedVariables)
...
+TEST(RooFormula, UndefinedVariables)
+{
+   RooRealVar B("B", "", 0.516952);
+   RooRealVar r("r", "", 0.214107);
+   RooRealVar x("x", "", 0.2, 1);
+   RooRealVar y("y", "", 0.2, 1);
+   
+   ASSERT_ANY_THROW(RooFormulaVar f1("f1", "r + B + x", {r, B}))  << "Formulae with missing x in arg list cannot work.";
+   ASSERT_ANY_THROW(RooFormulaVar f2("f2", "r + B + y", {r, B}))  << "Formulae with missing (x,)y in arg list cannot work.";
+   ASSERT_ANY_THROW(RooFormulaVar f2("f2", "r + B + y", {r, B, y})) << "Formulae with missing implicit (x,) in arg list cannot work since TFormula assumes that when y is specified, x exists too."; // Very similar to TestInvalidFormulae
```
ferdymercury, line 90:
```suggestion
   ASSERT_NO_THROW(RooFormulaVar f2("f2", "r + B + y", {r, B, y})) << "Formula with specified y must work.";
```

roofit/roofitcore/src/RooFormula.cxx
```diff
@@ -331,6 +331,22 @@ std::string RooFormula::reconstructFormula(std::string internalRepr) const {
...
+      std::stringstream regexStr;
+      regexStr << "x\\[" << i << "\\]|@" << i;
+      std::regex regex(regexStr.str());
+
+      std::string replacement = "0";
```
ferdymercury, line 344:
```suggestion
      std::string replacement = "1e-18";
```

Michealinc (approved):

roofit/roofitcore/src/RooFormula.cxx
```diff
@@ -512,21 +527,19 @@ void RooFormula::installFormulaOrThrow(const std::string& formula) {
...
-      if (varName.BeginsWith("x[") && varName[varName.Length()-1] == ']')
-        continue;
-
-      haveProblem = true;
-      msg << theFormula->GetVarName(i) << ",";
-    }
-    if (haveProblem) {
+  if (theFormula && theFormula->GetNdim() != 0) {
+    auto theNullFormula = std::make_unique<TFormula>((std::string(GetName()) + "0").c_str(), reconstructNullFormula(processedFormula).c_str());
+    if (theNullFormula && theNullFormula->GetNdim() != 0) {
```
guitargeek, line 532:
```suggestion
    TFormula nullFormula{"nullFormula", reconstructNullFormula(processedFormula, /*addToGlobList=*/false).c_str()};
    if (nullFormula.GetNdim() != 0) {
```
1. We generally don't use this `the*` prefix for local variables anymore
2. Can be on the stack
3. `nullptr` check redundant because `make_unique` always constructs something (or throws an exception)
4. Better set `addToGlobList=false` to avoid interacting with `gROOT` for this local object
5. Following from 3, the name now doesn't matter and you can avoid some string manipulation

roofit/roofitcore/src/RooFormula.cxx
```diff
@@ -331,6 +331,21 @@ std::string RooFormula::reconstructFormula(std::string internalRepr) const {
   return internalRepr;
 }
 
+////////////////////////////////////////////////////////////////////////////////
```
guitargeek, line 334:
```suggestion
namespace {

////////////////////////////////////////////////////////////////////////////////
```

Doesn't need to be a private member of `RooFormula`, no? It can be a free function in the anonymous namespace, you just need a `RooArgList const& args` argument for the `_origList`. Like this, we don't clutter the `RooFormula` interface with implementation details.

guitargeek (changes requested):
Thank you so much for your help here @ferdymercury, and this is a really nice trick! However, I have some suggestions to improve the implementation.

roofit/roofitcore/src/RooFormula.cxx
ferdymercury: sounds good. The original "reconstructFormula" we keep it though in the header, right? (I had just put it there for parallelism).

roofit/roofitcore/src/RooFormula.cxx
ferdymercury: Thanks for the review :)

Question:
1) I copy-pasted from theFormula. Do you want me to replace also theFormula with formula, or just use nullFormula and leave the previous variable untouched?

guitargeek: Personally I would keep `theFormula`, just for the sake of unique Ctrl+F-ability in the file :slightly_smiling_face:  `nullFormula` is already unique, so extending it to `theNullFormula` doesn't help.

roofit/roofitcore/src/RooFormula.cxx
guitargeek: If it can also be converted to a free function easily, it is better to move it too for consistency. I think static helper functions should be preferred over private member functions, because it makes it immediately clear what data members are used in the call, also also you can be immediately sure that no private data members are used. And also there are no exposed symbols and the user can't access these functions with `#define private public`.

guitargeek (approved):
LGTM!

PR #18071, opened 2025-03-20
[RF] average should interpolate if xFirst==xLast

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-9204


guitargeek (approved):
Thank you so much, that makes total sense!

PR #18074, opened 2025-03-20
[CMake] Don't force builtin libpng/gif/tiff/jpeg on macOS

guitargeek:
There is probably no reason to do this anymore. These libraries should be installed on the system, and if not, the automatic detection a few lines above the removed lines will take care of it.

Both the ROOT conda package and the nix package have patches to circumvent this already, and it would be better to do this upstream too.

### DIFF
--- a/cmake/modules/SearchInstalledSoftware.cmake
+++ b/cmake/modules/SearchInstalledSoftware.cmake
@@ -499,7 +499,7 @@ if(asimage)
     if(GIF_FOUND)
        set(_gifinclude  --with-gif --with-gif-includes=${GIF_INCLUDE_DIR} --without-builtin-gif)
     else()
-       set(_gifinclude)
+       set(_gifinclude  --with-builtin-ungif)
     endif()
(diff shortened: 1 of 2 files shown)


cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -511,11 +511,10 @@ if(asimage)
     else()
       set(_tiffinclude --with-tiff=no)
     endif()
-    if(cocoa)
-      set(_jpeginclude --without-x --with-builtin-jpeg)
-      set(_gifinclude  --with-builtin-ungif)
-      set(_pnginclude  --with-builtin-png)
-      set(_tiffinclude --with-tiff=no)
+    if(x11)
```
guitargeek, line 514: It's true that `cocoa=ON` usually implies `x11=OFF`, but it's better to check the `x11` flag directly.

bellenot (approved):
LGTM, thanks Jonas!

PR #18083, opened 2025-03-20
[Root7] A path to reducing global object registration and management.

hageboeck:
This is the start of a longer-term effort to transform ROOT into working with less globals, making ownership of objects clearer, and going more towards standard C++ behaviour instead of implicit ownership. One could call this the "ROOT 7 mode". In order to allow for early testing, this mode is introduced as an opt-in feature in ROOT 6.

## What this PR brings

`TROOT.h` gets three new functions (names up for discussion):
```c++
   namespace ROOT::Experimental {
   void EnableObjectAutoRegistration();
   void DisableObjectAutoRegistration();
   bool ObjectAutoRegistrationEnabled();
   }
```
https://github.com/hageboeck/root/blob/040980428a842a3314c75bde1ba9e6ae208fe8b1/core/base/inc/TROOT.h#L100-L104

The naming and API is inspired by `EnableImplicitMT()` and its colleagues. These determine whether TObject-derived classes auto-register themselves to `gDirectory` when created. A well-known example is TH1 and all its derived classes, but what's covered by these functions will be extended further on the path to ROOT 7. Currently, this mode also covers RooFit objects, TGraph, and TEfficiency.
An extensive list of objects with auto registration and their state (maybe not fully complete) can be found below.

A lot of preparatory work was done to prepare ROOT to work both with auto registration on and off. A detailed list of this work can be found at the end. Currently, there is one commit that switches auto registration off across the entire CI to demonstrate that ROOT already today works in this mode. Once the PR is ready to merge, this commit will be removed to not break the workflows of ROOT's users who did not opt in. A single (or a few; TBD) builds with auto registration off will be added to the CI to keep test coverage for this feature.

### Difference to `TH1::AddDirectoryStatus()`
The ROOT 7 mode with auto registration off might look similar to `TH1::AddDirectoryStatus()`, but there are important differences:
- DirectoryStatus is a static (global) bool while `ObjectAutoRegistrationEnabled()` is `thread_local`
- DirectoryStatus only affects histogram-like objects (see table at the end). AutoRegistrationEnabled should eventually extend to more cases, e.g. RooFit objects, TEventList (planned), functions (planned).
- DirectoryStatus affects the behaviour of `TFile::Get`. `file->Get("name"); file->Get("name")` returns the same object when directory status is `true`, and it reads a new object every time when it's `false`. I consider it best to not interfere here, because one can use TFile if such behaviour is desired, and RFile if not. `AutoRegistrationEnabled()` should only decide what happens to new objects, not objects read from a file.
- DirectoryStatus cannot be set from the environment, and its default cannot be changed. It is a non-thread-safe runtime feature. `AutoRegistrationEnabled()` can be initialised from the environment (thread safe), and its state can be changed per thread, which might be vital for using it in MT frameworks.

### Setting defaults 
Some users or frameworks might be interested in operating in ROOT 7 mode already today, but they don't want to add calls to the above functions in each thread of their code. Especially for testing, they should be able to set a default without recompiling. Defaults can be set in a .rootrc:
```
Root.ObjectAutoRegistration: 0
```
and also using an environment variable:
```
ROOT_OBJECT_AUTO_REGISTRATION=0 ./executable --args
```

When any object queries `ObjectAutoRegistrationEnabled()`, a thread-safe initialisation is performed, taking defaults from the environment above. If no defaults are set, registration defaults to "on" in ROOT 6 and "off" in ROOT 7.

This, however, might lead to an "it worked on my machine" problem: Some users might have altered their environment, while others might have chosen the opposite. To make it obvious when a ROOT 6 runs in ROOT 7 preview mode, info messages such as the following are printed if the environment modifies the default:
```shell
$ ./executable --args
Info in <TROOT>: Object auto registration disabled using the environment variable ROOT_OBJECT_AUTO_REGISTRATION
```
(A similar message exists for rootrc)

## List of classes covered (or intended to be covered) by "DisableObjectAutoRegistration"

|                                           | Auto registration disabled with `DisableObjectAutoRegistration()` | How could it be disabled before?      | Comments                                                                                |
| --------------------------------------------- | --------------------------------- | ----------------------------- | ----------------------------------------------------------------------------------------------------- |
| TH1 and derived                                           | Yes                                 | TH1::AddDirectoryStatus()     |                                                                                                       |
| TGraph2D                                      | Yes                                 | TH1::AddDirectoryStatus()     |                                                                                                       |
| RooPlot                                       | Yes                                 | RooPlot::addDirectoryStatus() |                                                                                                       |
| TEfficiency                                   | Yes (New)                           | Was inconsistent     |                                                                                                       |
| TProfile2D                         | Yes                                 | TH1::AddDirectoryStatus()             |                                                                                                       |
| TEntryList                                    | Yes (in the future)                                 |  Not possible                             | To be done later                                                                                                     |
| TEventList                                    | Yes (in the future)                                 |  Not possible                             |  To be done later                                                                                                     |
| TTree                                         | Probably no (to be decided)                                 |                               | Unclear if it should be changed. It would be broken without being attached to a file                           |
| TBranch                                       | No                                 |                               | It needs to know what its tree is                                                   |
| TChain, TNTuple                               | No                                 |                               | Same as for TTree                                                                                    |
| TFunction | Probably yes, in the future. To be decided  |    | This is a registration not to `gDirectory` but to a global list of functions. Way forward to be decided. | 

### TODOs:
- [x] Add tests for implicit / explicit registration of TH1s
- [x] Deprecate `TH1::AddDirectory()` as suggested in #21523 ?
  - In Doxygen, it is already marked as deprecated, but a C++ deprecation should probably happen a bit later on the way to ROOT 7.
- [x] Merge the PRs below to make ROOT ready for running both with and without auto registration.
- [x] Remove the commit that globally switches auto registration off in the CI. Instead, create a dedicated build for this feature.

### Preparatory work
The following PRs were part of this branch, and have been split off over time to separate the introduction of this feature from preparing ROOT to work both with auto registration off and on:
- #20818
- #20830
- #20832
- #20833 
- #20886
- #20888 
- #20889
- #20919
- #20920
- #20921
- #20922 
- #20978
- #21224
- #21225
- #21226
- #21228
- #21229
- #21231
- #21248 
- #21287
- #21296 
- #21319
- #21341
- #21523
- #21587 
- #21646
- #21648 
- #21679
- #21708

### DIFF
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -413,7 +413,7 @@ jobs:
           - image: alma10
             platform_config: alma10-clang_ninja
             is_special: true
-            property: "clang Ninja builtin"
+            property: "clang Ninja builtins auto-registration off"
             overrides: ["CMAKE_CXX_STANDARD=20"]
           # Fedora Rawhide with Python debug build
           - image: rawhide
@@ -474,6 +474,10 @@ jobs:
           ccache -s || true
           if [ -d /py-venv/ROOT-CI/bin/ ]; then
             . /py-venv/ROOT-CI/bin/activate && echo PATH=$PATH >> $GITHUB_ENV;
+          fi
+          if ${{ contains(matrix.property, 'auto-registration off') }}; then
+            echo "Disabling ROOT's object auto registration for this job"
+            echo ROOT_OBJECT_AUTO_REGISTRATION=0 >> $GITHUB_ENV;
           fi
 
       - name: Checkout
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -468,17 +468,13 @@ jobs:
       POST_INSTALL_DIR: /github/home/ROOT-CI/PostInstall
 
     steps:
-      - name: Configure ccache
+      - name: Set up environment
         run: |
           ccache -o max_size=${{ matrix.is_special && '5G' || '1.5G' }}
-          ccache -p || true
           ccache -s || true
-
-      - name: Set up Python Virtual Env
-        # if the `if` expr is false, `if` still has exit code 0.
-        # if the `if` block is entered, the block's exit code becomes the exit
-        # code of the `if`.
-        run: 'if [ -d /py-venv/ROOT-CI/bin/ ]; then . /py-venv/ROOT-CI/bin/activate && echo PATH=$PATH >> $GITHUB_ENV; fi'
+          if [ -d /py-venv/ROOT-CI/bin/ ]; then
+            . /py-venv/ROOT-CI/bin/activate && echo PATH=$PATH >> $GITHUB_ENV;
+          fi
 
       - name: Checkout
         uses: actions/checkout@v6
@@ -499,7 +495,8 @@ jobs:
 
       - name: Print debug info
         run:  'printf "%s@%s\\n" "$(whoami)" "$(hostname)";
-               ls -la
+               ls -la;
+               echo "Path is: $PATH"
               '
 
       - uses: root-project/gcc-problem-matcher-improved@main
--- a/core/base/src/TROOT.cxx
+++ b/core/base/src/TROOT.cxx
@@ -71,9 +71,11 @@ of a main program creating an interactive version is shown below:
 #include <ROOT/RVersion.hxx>
 #include "RConfigure.h"
 #include "RConfigOptions.h"
+#include <atomic>
 #include <filesystem>
 #include <string>
 #include <map>
+#include <sstream>
 #include <set>
 #include <cstdlib>
 #ifdef WIN32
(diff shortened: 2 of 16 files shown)


hist/hist/src/TH1.cxx
```diff
@@ -741,7 +741,7 @@ TH1::TH1(const char *name,const char *title,Int_t nbins,const Double_t *xbins)
 
 Bool_t TH1::AddDirectoryStatus()
 {
-   return fgAddDirectory;
+   return !ROOT::ROOT7MemoryManagement() && fgAddDirectory;
```
pcanal, line 744: Note that this flag is primarily about file constructions (and secondarily about shared ownership).  Beside the (intentional) memory leak it introduces in existing scripts, it leads most of them to not longer write the histograms to the file (because they are no longer attached to the File).  It is not clear whether this kind of approach is worth it compared to a more user active change (ie. Switching explicitly from TFile to RFile).

Michealinc (dismissed):

.github/workflows/root-ci.yml
```diff
@@ -157,10 +157,11 @@ jobs:
           INCREMENTAL: ${{ !contains(github.event.pull_request.labels.*.name, 'clean build') && !matrix.platform == 'mac15' && !matrix.platform == 'mac26'}}
           GITHUB_PR_ORIGIN: ${{ github.event.pull_request.head.repo.clone_url }}
           OVERRIDES: ${{ join( matrix.overrides, ' ') }}
+          ROOT_OBJECT_AUTO_REGISTRATION: 0
         run: |
           [ -d "${VIRTUAL_ENV_DIR}" ] && source ${VIRTUAL_ENV_DIR}/bin/activate
           echo "Python is now $(which python3) $(python3 --version)"
-          src/.github/workflows/root-ci-config/build_root.py  \
+          src/.github/workflows/root-ci-config/build_root.py \
```
pcanal, line 164:
```suggestion
          src/.github/workflows/root-ci-config/build_root.py  \
```

core/base/src/TROOT.cxx
```diff
@@ -293,6 +295,78 @@ namespace {
       static std::vector<ModuleHeaderInfo_t> moduleHeaderInfoBuffer;
       return moduleHeaderInfoBuffer;
    }
+
+   enum class AutoReg : unsigned char {
+      NotInitialised = 0,
+      On,
+      Off,
```
pcanal, line 302:
```suggestion
      kNotInitialised = 0,
      kOn,
      kOff,
```

core/base/src/TROOT.cxx
```diff
@@ -293,6 +295,78 @@ namespace {
...
+            }
+         }
+      }
+
+      if (tlsState == AutoReg::NotInitialised) {
+         assert(defaultState != AutoReg::NotInitialised);
+         tlsState = defaultState;
+      }
+
+      return tlsState;
```
pcanal, line 368: ~Where is it documented that the run-time turning on and off is only a thread-local operation?~  Found it.

core/base/src/TROOT.cxx
```diff
@@ -293,6 +295,78 @@ namespace {
...
+   /// A default can be set in a .rootrc using e.g. "Root.ObjectAutoRegistration: 1" or setting
+   /// the environment variable "ROOT_OBJECT_AUTO_REGISTRATION=0".
+   AutoReg &ObjectAutoRegistrationEnabledImpl()
+   {
+      static std::atomic<AutoReg> defaultState = AutoReg::NotInitialised;
+      thread_local static AutoReg tlsState = AutoReg::NotInitialised;
+      constexpr auto rcName = "Root.ObjectAutoRegistration";    // Update the docs if this is changed
+      constexpr auto envName = "ROOT_OBJECT_AUTO_REGISTRATION"; // Update the docs if this is changed
+
+      if (defaultState.load(std::memory_order_relaxed) == AutoReg::NotInitialised) {
```
pcanal, line 317: Did you consider (instead of the atomic):
```suggestion
      static const AutoReg defaultState = []() {
      ....
      } ();
```

core/base/src/TROOT.cxx
```diff
@@ -616,6 +689,68 @@ namespace Internal {
...
+   /// TH1x* histogram = new TH1x(...);
+   /// histogram->SetDirectory(&file);
+   /// \endcode
+   ///
+   /// A default can be set in a .rootrc using e.g. `Root.ObjectAutoRegistration: 1` or setting
+   /// the environment variable `ROOT_OBJECT_AUTO_REGISTRATION=0`.
+   ///
+   /// \note For classes deriving from TH1, both ObjectAutoRegistrationEnabled() and TH1::AddDirectoryStatus()
+   /// need to be true for auto-registration to take effect. The former should be preferred over the latter, because
+   /// it is thread safe and extends to more objects such as TGraph2D, TEfficiency, RooPlot.
```
pcanal, line 730:
```suggestion
   /// it is thread local and extends to more objects such as TGraph2D, TEfficiency, RooPlot.
```

core/base/src/TROOT.cxx
hageboeck: This should work, thanks for the idea!

.github/workflows/root-ci.yml
hageboeck: This commit will be removed before merging. It is here to demonstrate that all tests work today also when object auto registration is off.

core/base/src/TROOT.cxx
```diff
@@ -616,6 +686,68 @@ namespace Internal {
       return 0;
 #endif
    }
+
+   /// Namespace for ROOT features in testing.
+   /// The API might change without notice until moved out of this namespace.
+   namespace Experimental {
+   ////////////////////////////////////////////////////////////////////////////////
+   /// \brief Enable automatic registration of objects for the current thread (ROOT 6 default).
```
dpiparo, line 699: Sorry to be picky here: does it make sense to have a table of the classes the instances of which are affected by the switch in this documentation?
That table can be also updated when new classes are added, such as`TE*List`.

dpiparo (approved):
Thanks for this PR and the massive preparatory work. I am proud to see ROOT managed to put some order in this very core functionality, to modernise itself and offer an even more rewarding user-experience.

core/base/src/TROOT.cxx
hageboeck: Yes, the table appeared now after the latest rebase. It's, however, not in the function you marked (which cannot be called outside of `TROOT.cxx`), it's in the function `ROOT::Experimental::DisableObjectAutoRegistration()` and its siblings.

Once merged, the docs should look roughly like this:
image

PR #18084, opened 2025-03-20
[tmva][sofie] Add support for ScatterElements operator

lmoneta:
Add support for a new operator, ScatterElements, needed for parsing the GNN models such as the tracking one used by ATLAS

Add also test for ScatterElements


sanjibansg (approved):
LGTM! Thanks for the implementation.

Issue #15252, opened 2024-04-17
Wrong, missing, and misleading documentation in RooBarlowBeestonLL.cxx


HDembinski:
### Explain what you would like to see improved and how.

There are several issues with the class RooBarlowBeestonLL on the master.

Short version:

- [ ] The name suggests it implements the Barlow-Beeston likelihood, while in fact it implements Conway's approximation. The class name is misleading, since it does not match the implementation. The difference between the two matters, they are not equivalent.

- [ ] The code documentation, shown here https://root.cern/doc/master/classRooStats_1_1HistFactory_1_1RooBarlowBeestonLL.html, says it is the Barlow-Beeston likelihood, which is incorrect.

- [ ] References to Conway in code documentation and user guide (https://cds.cern.ch/record/1456844/files/CERN-OPEN-2012-016.pdf, page 5) are missing. The reference is https://inspirehep.net/literature/891252. Note that the user guide under that link contains no bibliography at all.

- [ ] Clarification wrt usage of weighted samples. It should be clarified both in the code documentation and in the user guide, if the implementation can handle weighted templates in the asymptotic limit, by applying the SPD approximation discussed by Bohm and Zech, 2014. This is possible (as we show in our paper https://inspirehep.net/literature/2512593), but only if the implementations handles this special case. Conway's paper does not discuss weighted templates and does not explain how to use the SPD approximation in this context.

Long version:

The code documentation of RooBarlowBeestonLL on the master is incorrect. It is stated that this class implements the Barlow-Beeston likelihood computation, but it actually implements an approximation introduced by Conway in 2011. The approximation is only asymptotically equivalent to Barlow-Beeston, for small samples it is inferior. We recently published a paper about a new approximation to Barlow-Beeston (https://inspirehep.net/literature/2512593), which is fast to compute and can be used with weighted templates and weighted data. In this paper, we compare the performance of Barlow-Beeston and several approximations including ours. It is clearly shown there, that Conway's approximation is inferior to the exact computation by Barlow-Beeston.

I was pointed to this class by the LHCb collaboration. There is uncertainty in the community what this class actually computes and whether it is asymptotically correct for weighted templates. I kindly ask to correct and extend the documentation.

**I think the best solution is to revert the implementation of BarlowBeestonLL to actually implement the original BarlowBeeston.** Then make a new class called ConwayLL which implements Conway's approximation. I am happy to provide another class which implements our method in ROOT. So far, the reference implementation is in the iminuit Python package.

### ROOT version

[...]

PR #18090, opened 2025-03-21
[RF] copy-paste barlow-beeston to new class with explicit Conway name

ferdymercury:
Fixes https://github.com/root-project/root/issues/15252

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes #

### DIFF
--- a/roofit/histfactory/CMakeLists.txt
+++ b/roofit/histfactory/CMakeLists.txt
@@ -41,6 +41,7 @@ ROOT_STANDARD_LIBRARY_PACKAGE(HistFactory
     RooStats/HistFactory/PiecewiseInterpolation.h
     RooStats/HistFactory/PreprocessFunction.h
     RooStats/HistFactory/RooBarlowBeestonLL.h
+    RooStats/HistFactory/RooBarlowBeestonLLConway.h
     RooStats/HistFactory/Sample.h
     RooStats/HistFactory/Systematics.h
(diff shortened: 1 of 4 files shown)


ferdymercury:
@HDembinski would you like to propose two things into this PR based on your reported issue?

- [ ] Replace current BarlowBeestonLL with exact computation
- [ ] Provide a third class using the approximation from your paper

HDembinski:
Yes, I can do that. Would be good to have a common base class for the Conway approximation and my implementation, because they are very similar, except for the actual computation of the likelihood.

ferdymercury:
> Yes, I can do that. Would be good to have a common base class for the Conway approximation and my implementation, because they are very similar, except for the actual computation of the likelihood.

Thanks! Makes sense. Or maybe even a private enum member that chooses which approx to choose.

Michealinc (approved):

ferdymercury:
> Thanks! Makes sense. Or maybe even a single class RooBarlowBeeston with a private enum member that chooses which approx to choose: 0 = Exact, 1 = Conway, 2 = Dembinski ? (And erasing RooBarlowBeestonLLConway)

@guitargeek what alternative would be the best way to go in your opinion?

ferdymercury:
> @guitargeek what alternative would be the best way to go in your opinion?

gentle ping @guitargeek

PR #18092, opened 2025-03-21
[DF] Retrieve RDatasetSpec for RDF built from JSON

martamaja10:
Separating FromSpec function into two - so that we can retrieve the Spec when we create the an RDataFrame using `FromSpec(jsonFile)`. This is needed to implement FromSpec in distRDF.


tree/dataframe/inc/ROOT/RDF/Utils.hxx
```diff
@@ -13,12 +13,14 @@
...
 #include "ROOT/RSpan.hxx"
 #include <string_view>
+#include "ROOT/RDF/RDatasetSpec.hxx"
 #include "ROOT/RVec.hxx"
 #include "ROOT/TypeTraits.hxx"
 #include "Rtypes.h"
 
 #include <array>
 #include <deque>
+#include <fstream>
```
vepadulano, line 23: I think these two headers are not necessary, I would remove them. For `RDatasetSpec`, a forward declaration should be enough, i.e. writing the following somewhere after the `#include`s:

```
namespace ROOT::RDF::Experimental{
class RDatasetSpec;
}
````

File: tree/dataframe/src/RDataFrame.cxx [hunk 2]
```diff
@@ -1962,78 +1962,7 @@ namespace Experimental {
...
-   if (fullData.contains("range")) {
-      std::vector<int> range = fullData["range"];
-
-      if (range.size() == 1)
-         spec.WithGlobalRange({range[0]});
-      else if (range.size() == 2)
-         spec.WithGlobalRange({range[0], range[1]});
-   }
+   RDatasetSpec spec = ROOT::Internal::RDF::RetrieveSpecFromJson(jsonFile);
    return ROOT::RDataFrame(spec);
```
vepadulano, line 1966: Avoids the extra copy
```suggestion
   return ROOT::RDataFrame(ROOT::Internal::RDF::RetrieveSpecFromJson(jsonFile));
```

vepadulano (changes requested):
Thank you! Some minor comments before merging

tree/dataframe/inc/ROOT/RDF/Utils.hxx
martamaja10: thank you for the comments, fstream is needed to parse the JSON file though --> `const nlohmann::ordered_json fullData = nlohmann::ordered_json::parse(std::ifstream(jsonFile));`

Michealinc (dismissed):

vepadulano (approved):
LGTM thanks! Please squash the commits before merging 👍

PR #18098, opened 2025-03-21
[tmva] Fix using  TString in Python

lmoneta:
# This Pull request:

-  Avoid using the automatic conversions from TString to Python string which causes a lot of errors like  
 `Error in <TString::AssertElement>: out of bounds`.  Use now `str(TString)`

- Fix warning in ROperator_Conv


sanjibansg (approved):
LGTM! Thanks, I think there are some formatting issues with Ruff which we might fix before merging?

PR #18105, opened 2025-03-24
[skip-ci][ci] Remove Fedora40 builds

dpiparo:
since its EOL is coinciding with the foreseen release date of ROOT 6.36.00


guitargeek (approved):
Makes sense, thanks!

PR #18110, opened 2025-03-24
[ntuple] update RNTupleModel's documentation

silverweed:


tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -49,13 +49,14 @@ RProjectedFields &GetProjectedFieldsOfModel(RNTupleModel &model);
 /**
 \class ROOT::Experimental::Internal::RProjectedFields
 \ingroup NTuple
-\brief The projected fields of a `RNTupleModel`
+\brief Container for the projected fields of a RNTupleModel
```
enirolf, line 52:
```suggestion
\brief Container for the projected fields of an RNTupleModel
```

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -49,13 +49,14 @@ RProjectedFields &GetProjectedFieldsOfModel(RNTupleModel &model);
...
 
 Projected fields are fields whose columns are reused from existing fields. Projected fields are not attached
-to the models zero field.  Only the real source fields are written to, projected fields are stored as meta-data
-(header) information only.  Only top-level projected fields are supported because otherwise the layout of types
+to the Model's zero field but form a separate hierarchy with their own zero field (which is stored in this class).
```
enirolf, line 55:
```suggestion
to the model's zero field but form a separate hierarchy with their own zero field (which is stored in this class).
```
(or "RNTupleModel's"

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -49,13 +49,14 @@ RProjectedFields &GetProjectedFieldsOfModel(RNTupleModel &model);
...
+Only the real source fields are written to: projected fields are stored as meta-data
```
enirolf, line 56:
```suggestion
Only the real source fields are written to: projected fields are stored as metadata
```
Also applies to any other occurrences (see #18107)

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -105,28 +107,42 @@ public:
 /**
 \class ROOT::Experimental::RNTupleModel
 \ingroup NTuple
-\brief The RNTupleModel encapulates the schema of an ntuple.
+\brief The RNTupleModel encapulates the schema of a RNTuple.
```
enirolf, line 110:
```suggestion
\brief The RNTupleModel encapulates the schema of an RNTuple.
```

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -105,28 +107,42 @@ public:
...
 
 The ntuple model comprises a collection of hierarchically organized fields. From a model, "entries"
```
enirolf, line 112:
```suggestion
The RNTupleModel comprises a collection of hierarchically organized fields. From a model, REntries
```

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -105,28 +107,42 @@ public:
...
+can be extracted or created. For convenience, the RNTupleModel provides a default entry unless it is created as a "bare model".
+Models have a unique model identifier that facilitates checking whether entries are compatible with it
 (i.e.: have been extracted from that model).
 
-A model is subject to a state transition during its lifetime: it starts in a building state, in which fields can be
-added and modified.  Once the schema is finalized, the model gets frozen.  Only frozen models can create entries.
-From frozen, models move into a expired state.  In this state, the model is only partially usable: it can be cloned
+A model is subject to state transitions during its lifetime: it starts in a Building state, in which fields can be
+added and modified.  Once the schema is finalized, the model gets Frozen.  Only frozen models can create entries.
+From frozen, models move into an Expired state. In this state, the model is only partially usable: it can be cloned
```
enirolf, line 119: Not sure if I agree with the capitalization here, I think together with the diagram below it reads quite naturally as is. Perhaps use something else to emphasize the different states:
```suggestion
A model is subject to state transitions during its lifetime: it starts in a *building* state, in which fields can be
added and modified.  Once the schema is finalized, the model gets *frozen*.  Only frozen models can create entries.
From frozen, models move into an *expired* state. In this state, the model is only partially usable: it can be cloned
```

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -307,37 +326,67 @@ public:
    /// Creating projections for fields containing `std::variant` or fixed-size arrays is unsupported.
    RResult<void> AddProjectedField(std::unique_ptr<ROOT::RFieldBase> field, FieldMappingFunc_t mapping);
 
+   /// Transitions a RNTupleModel from the Building state to the Frozen state, disabling adding additional fields and enabling
```
enirolf, line 329:
```suggestion
   /// Transitions an RNTupleModel from the *building* state to the *frozen* state, disabling adding additional fields and enabling
```
same applies below

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -307,37 +326,67 @@ public:
...
+   /// creating entries from it.
+   /// Freezing an already-frozen model is a no-op.
+   /// Throws an RException if the model is in the Expired state.
```
enirolf, line 332:
```suggestion
   /// Throws an RException if the model is in the *expired* state.
```
same applies below

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -307,37 +326,67 @@ public:
...
+   /// See RNTupleModel for more explanations on the state transitions.
```
enirolf, line 333:
```suggestion
   /// See RNTupleModel for more explanation on the state transitions.
```

enirolf, line 333:
```suggestion
   /// See RNTupleModel for more explanation on the state transitions.
```
same applies below

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -307,37 +326,67 @@ public:
...
    bool IsBare() const { return !fDefaultEntry; }
    std::uint64_t GetModelId() const { return fModelId; }
    std::uint64_t GetSchemaId() const { return fSchemaId; }
 
+   /// Creates a new entry with default values for each field.
    std::unique_ptr<REntry> CreateEntry() const;
-   /// In a bare entry, all values point to nullptr. The resulting entry shall use BindValue() in order
-   /// set memory addresses to be serialized / deserialized
+   /// Created a "bare entry", i.e. a entry with all null values. The user needs to explicitly call BindValue() or
+   /// BindRawPtr to set memory addresses before serializing / deserializing the entry.
```
enirolf, line 359:
```suggestion
   /// BindRawPtr() to set memory addresses before serializing / deserializing the entry.
```

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -372,6 +421,8 @@ struct RNTupleModelChangeset {
    bool IsEmpty() const { return fAddedFields.empty() && fAddedProjectedFields.empty(); }
 
    void AddField(std::unique_ptr<ROOT::RFieldBase> field);
+
+   /// \see RNTupleModel::AddProjectedField
```
enirolf, line 425:
```suggestion
   /// \see RNTupleModel::AddProjectedField()
```

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -414,6 +465,7 @@ public:
 
    void AddField(std::unique_ptr<ROOT::RFieldBase> field);
 
+   /// \see RNTupleModel::AddProjectedField
```
enirolf, line 468:
```suggestion
   /// \see RNTupleModel::AddProjectedField()
```

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
silverweed: Whereas I can agree with the emphasis/non-capitalization in the class comment (as it's the first time the states are introduced), I'd argue that here the capitalization fits better because we are referring to a precise, named concept (corresponding to the nodes of the ascii graph above and to the values of the EState enum - which, sure, are private, but are still visible in the documentation)

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
silverweed: Is "explanation" uncountable? In which case, shouldn't we say "for a more detailed explanation" or something? ("more explanation" sounds wrong to me..)

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
enirolf: Ok fair enough, I'm personally not a big fan of lots of capitalization but I won't stay awake over it :)

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
enirolf: As far as I'm aware in this context it is, because we're explaining one single topic (state transtitions, it doesn't matter that there are multiple aspects to it). "More explanation" might be a Dutchism, so changing it to "for a more detailed explanation" sounds good to me as well.

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
enirolf: Just one thing I noticed now, in line 114-116 is is written italicized and uncapitalized

tree/ntuple/v7/inc/ROOT/RNTupleModel.hxx
```diff
@@ -291,31 +310,60 @@ public:
    /// Creating projections for fields containing `std::variant` or fixed-size arrays is unsupported.
    RResult<void> AddProjectedField(std::unique_ptr<ROOT::RFieldBase> field, FieldMappingFunc_t mapping);
 
+   /// Transitions a RNTupleModel from the Building state to the Frozen state, disabling adding additional fields and enabling
+   /// creating entries from it.
+   /// Freezing an already-frozen model is a no-op.
+   /// Throws an RException if the model is in the Expired state.
+   /// See RNTupleModel for more detailed explanatio on the state transitions.
```
enirolf, line 317:
```suggestion
   /// See RNTupleModel for more detailed explanation on the state transitions.
```

enirolf (approved):

PR #18112, opened 2025-03-24
[skip-ci][ntuple] Fix-up `RFieldBase` documentation

enirolf:


tree/ntuple/v7/inc/ROOT/RFieldBase.hxx
```diff
@@ -560,7 +559,9 @@ public:
    /// Generates an object of the field type and wraps the created object in a shared pointer and returns it an RValue
    /// connected to the field.
    RValue CreateValue();
-   /// The returned bulk is initially empty; RBulk::ReadBulk will construct the array of values
+   /// Creates a new, initially empty bulk
```
hahnjo, line 564: We may want a full stop or `\brief` here, otherwise I think Doxygen will just concatenate the next line as well

hahnjo (approved):
LGTM, one minor comment

PR #18116, opened 2025-03-24
[skip-ci][ntuple] Fix-up `RField<T>` documentation

enirolf:


tree/ntuple/v7/inc/ROOT/RField/RFieldProxiedCollection.hxx
```diff
@@ -207,48 +207,50 @@ struct HasCollectionProxyMemberType<
...
-///    using IsCollectionProxy = std::true_type;
-/// };
-/// ```
-///
-/// Of course, there is another possible solution which is to have a single `RClassField` that implements both
-/// the regular-class and the collection-proxy behaviors, and always chooses appropriately at run time.
-/// We found that less clean and probably has more overhead, as most probably it involves an additional branch + call
-/// in each of the member functions.
+/// Helper type trait for marking classes as a collection proxy.
+/// This type trait must be set to for collection proxy-based RNTuple fields created through MakeField<T>.
```
silverweed, line 211:
```suggestion
/// This type trait must be set for collection proxy-based RNTuple fields created through MakeField<T>.
```

tree/ntuple/v7/inc/ROOT/RField/RFieldProxiedCollection.hxx
```diff
@@ -207,48 +207,50 @@ struct HasCollectionProxyMemberType<
...
+// ```
+// auto field = std::make_unique<RField<MyClass>>("klass");
+// // vs
+// auto otherField = std::make_unique<RField<MyClass, ROOT::Experimental::TagIsCollectionProxy>>("klass");
+// ```
+//
+// That is convenient only for non-nested types, i.e. it doesn't work with, e.g. `RField<std::vector<MyClass>,
+// ROOT::Experimental::TagIsCollectionProxy>`, as the tag is not forwarded to the instantiation of the inner RField
+// (that for the value type of the vector).  The following two possible solutions were considered:
+// - A wrapper type (much like `ntuple/v7/inc/ROOT/RNTupleUtil.hxx:49`), that helps to differentiate both cases.
```
silverweed, line 225:
```suggestion
// - A wrapper type that helps to differentiate both cases.
```

silverweed (approved):
LGTM, thanks

tree/ntuple/v7/inc/ROOT/RField/RFieldProxiedCollection.hxx
```diff
@@ -207,48 +207,50 @@ struct HasCollectionProxyMemberType<
...
-/// public:
-///    using IsCollectionProxy = std::true_type;
-/// };
-/// ```
-///
-/// Of course, there is another possible solution which is to have a single `RClassField` that implements both
-/// the regular-class and the collection-proxy behaviors, and always chooses appropriately at run time.
-/// We found that less clean and probably has more overhead, as most probably it involves an additional branch + call
-/// in each of the member functions.
+/// Helper type trait for marking classes as a collection proxy.
```
vepadulano, line 210: Consider moving this inline doc statement right above the struct declaration

tree/ntuple/v7/inc/ROOT/RField/RFieldProxiedCollection.hxx
```diff
@@ -207,48 +207,50 @@ struct HasCollectionProxyMemberType<
...
+// class MyClass {
+// public:
+//    using IsCollectionProxy = std::true_type;
+// };
+// ```
+//
+// Of course, there is another possible solution which is to have a single RClassField that implements both
+// the regular-class and the collection-proxy behaviors, and always chooses appropriately at run time.
+// We found that less clean and probably has more overhead, as most probably it involves an additional branch + call
+// in each of the member functions.
```
vepadulano, line 248: Every other comment in this block looks to me like it's more a doc for developers than a description of the struct itself. Given that this is in the public namespace, we could consider having a single `/** ... */` doxygen block, probably wrapped by `\cond HIDDEN_SYMBOLS` so it doesn't show up on the reference. Note that much of the wording here is colloquial/technical, see https://root.cern.ch/doc/master/structROOT_1_1IsCollectionProxy.html


At the same time, since this PR is merely changing the formatting of this section, which preexisted the PR itself, this can be done in a separate PR.

enirolf: This why I changed the comment delimiter from `///` to `//`. IIUC the latter won't be picked up by doxygen but if this is not the case I will change it to your suggestion!

PR #18122, opened 2025-03-24
[hist] For consistency, provide also GetParError(name), and constify some functions

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-7322

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/hist/hist/inc/TF1.h
+++ b/hist/hist/inc/TF1.h
@@ -303,9 +303,9 @@ protected:
 private:
    // NSUM parsing helper functions
    void DefineNSUMTerm(TObjArray *newFuncs, TObjArray *coeffNames,
-		       TString &fullFormula,
-		       TString &formula, int termStart, int termEnd,
-		       Double_t xmin, Double_t xmax);
+               TString &fullFormula,
+               TString &formula, int termStart, int termEnd,
+               Double_t xmin, Double_t xmax);
    int TermCoeffLength(TString &term);
 
 protected:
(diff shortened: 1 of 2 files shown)


hist/hist/inc/TF1.h
```diff
@@ -596,17 +600,17 @@ class TF1 : public TNamed, public TAttLine, public TAttFill, public TAttMarker {
    {
       return (fFormula) ? fFormula->GetVariable(name) : 0;
    }
-   virtual Double_t GradientPar(Int_t ipar, const Double_t *x, Double_t eps = 0.01);
+   virtual Double_t GradientPar(Int_t ipar, const Double_t *x, Double_t eps = 0.01) const;
    template <class T>
-   T GradientPar(Int_t ipar, const T *x, Double_t eps = 0.01);
+   T GradientPar(Int_t ipar, const T *x, Double_t eps = 0.01 const;
```
ferdymercury, line 605:
```suggestion
   T GradientPar(Int_t ipar, const T *x, Double_t eps = 0.01) const;
```

hageboeck (commented):
Hi @ferdymercury,

thanks, it looks good! Could you interactive rebase your commits, so the fix and the changes to constness are separated? If I squash, everything would be in one commit.
The best would probably be to do the constness as first commit, and add the "fix" in the second one.

ferdymercury:
> Hi @ferdymercury,
> 
> thanks, it looks good! Could you interactive rebase your commits, so the fix and the changes to constness are separated? If I squash, everything would be in one commit. The best would probably be to do the constness as first commit, and add the "fix" in the second one.

Thanks for the review Stephan, sure, rebased.

dpiparo:
@lmoneta

hageboeck (approved):
Thanks @ferdymercury!

PR #18124, opened 2025-03-25
Split long-running test

vepadulano:
Splits `ntuple_extended` test into its subtests in different files, allowing for better parallelization in ctest runs.


vepadulano:
I have re-instated the original `ntuple_extended` test, so we can get a runtime comparison at the same time between the integral test and the split subtests

vepadulano:
Anecdotally, here is what I see from the latest CI run

https://github.com/root-project/root/actions/runs/14086677685/job/39452494192?pr=18124

sequential test:
ntuple-extended 55s

parallel tests:
ntuple-double32imt 0.6s
ntuple-largefile1 17s
ntuple-largefile2 19.3s
ntuple-largepages 8s
ntuple-multicolumnexpansion 1.2s
ntuple-randomaccess 21.6s
ntuple-realworld1 5.4s

sum of runtimes for all sub-tests that have run in parallel: 73s

hahnjo (commented):
Not sure we want to inflate the number of tests with a single unit test per binary. In particular `ntuple-double32imt`, `ntuple-largepages`, `ntuple-multicolumnexpansion`, and `ntuple-realworld1` maybe look a bit questionable to me. Maybe leave them in `ntuple-extended` and just split off the others?

hahnjo (approved):
Thanks, LG!

pcanal (approved):

PR #18127, opened 2025-03-25
[CI] Run cling tests on CI

devajithvs:
# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes #


ferdymercury:
Thanks for this! Should I close then https://github.com/root-project/root/pull/16917 ? The main difference was that the special build was only running if 'cling' was touched.

.github/workflows/cling-test.yml
```diff
@@ -0,0 +1,75 @@
+name: 'Cling Test CI'
+
+on:
+  pull_request:
+    paths:
+      - 'interpreter/cling/**'
+      - '.github/workflows/cling-check.yml'
```
ferdymercury, line 7:
```suggestion
      - '.github/workflows/cling-test.yml'
```
?

PR #18131, opened 2025-03-25
[metacling] Use GetSizeOfType from CppInterOp

guitargeek:
Use GetSizeOfType from CppInterOp.


aaronj0 (approved):
LGTM, Thanks for this PR!

PR #18136, opened 2025-03-25
[ntuple] Fix unnecessary join table builds

enirolf:


bellenot (approved):
LGTM, (once all is green). Thanks Florine!

hahnjo (changes requested):
Tests are red - I think we either need one boolean flag per join table or build all together

enirolf:
> where/when is `fJoinTablesAreBuilt` reset to false?

Never, but this is intended because building the join table is a (lazy) one-time only action (i.e., it is not possible to "unbuild" a join table). `fJoinTablesAreBuilt` is initialized as `false` when an `RNTupleJoinProcessor` gets created.

hahnjo (approved):

PR #18142, opened 2025-03-26
[skip-ci][tutorials] Show categories of tutorials in an alphabetical order.

martamaja10:
On the main page of tutorials, we first have a description in which we link to the most often used categories. Because of that, the table below can now be ordered in an alphabetical order instead of "priority" order.


mdessole (approved):

mdessole:
LGTM, thank you Marta!

PR #18149, opened 2025-03-26
[CPyCppyy] Don't use deprecated `PyErr_Fetch` and `PyErr_Restore` API

guitargeek:
Replace API functions deprecated since Python 3.12:

  * https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Fetch
  * https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Restore

See also:

  * https://stackoverflow.com/questions/78926819/c-python-c-api-is-exception-returned-by-pyerr-getraisedexception-normalized


vepadulano (approved):
Thank you! It's good to keep track of the CPython developments

PR #18152, opened 2025-03-26
[tutorials] Rename `ntuple` to `reader` and `writer`

enirolf:
...distinguishing more explicitly between the use of `RNTupleReader` and `RNTupleWriter`.


silverweed (approved):
Thanks!

PR #18151, opened 2025-03-26
[tutorials][skip-ci] Fix paths to analysis/parallel in CMake

jolly-chen:
# This Pull request:
Adds some missing changes to the tutorials `CMakeLists.txt`  that are required after moving the parallel analysis tutorials into `tutorial/analysis/parallel` #18145

PR #18164, opened 2025-03-27
[tutorials] Fix veto of `mp*.C`

hahnjo:
They were moved in commit 9a126e1773.


hahnjo:
https://github.com/root-project/root/pull/18151 was already open, just not merged...

PR #18166, opened 2025-03-27
[RDF] Provide access to the shared_ptr in RResultPtr.

### LINKED ISSUE #14766: RResulPtr could better convey/share ownership of pointee

This forum post https://root-forum.cern.ch/t/too-many-operators-for-the-ttree-project-conditions/58009/5 triggered a discussion that probably `RResultPtr` could better behave with respect to ownership of its pointee. For example, this function
```cpp
TH1D *DrawStuff(const char *columnname) {
   auto file = TFile::Open("tutorials/hsimple.root", "READ");
   auto ntuple = file->Get<TTree>("ntuple");
   ROOT::RDataFrame df(*ntuple);
   auto ht = df.Histo1D(*h, columname);
   return ht.GetPtr();
}
```

Will return a dangling pointer since the `RResultPtr` going out of scope will destroy the pointee. The forum user presents a pattern such as

```cpp
    THStack* histStack = new THStack("histStack", "Bs_MCORR");
    for(int i=0; i<sampNumb; i++) {
        TFile* file = new TFile(rootList[i]);
        TTree* tree = (TTree*)file->Get("DecayTree");
        ROOT::RDataFrame df(*tree);
        auto ht = df.Filter(conds[i].Data()).Histo1D(h, "Bs_MCORR");
        histStack->Add(ht.GetPtr(), , "HIST && E1 && EX0");
   }
   histStack->Draw();
```

Which also doesn't work with the current API. Admittedly, this example would not work in general since `THStack` does not take ownership of the passed histograms and assumes they will be kept alive by ROOT global objects. In general, for this kind of use case we could convey the message to users they can store the histograms in a vector of shared pointers which is the clearest and most standard data structure to represent the situation correctly. So we need:

* `std::shared_ptr<T> RResultPtr<T>::GetSharedPtr()`
* More documentation on the RDataFrame user guide about these use cases
* A tutorial starting from the forum post reproducer once we have the API

### ROOT version

Any

### Installation method

Any

### Operating system

Any

### Additional context

_No response_

hageboeck:
Users reported problems with losing the underlying object when the RResultPtr goes out of scope. By adding GetSharedPtr(), this can be avoided.

Fix #14766.

### DIFF
--- a/tree/dataframe/inc/ROOT/RResultPtr.hxx
+++ b/tree/dataframe/inc/ROOT/RResultPtr.hxx
@@ -96,15 +96,19 @@ namespace RDFInternal = ROOT::Internal::RDF;
 namespace RDFDetail = ROOT::Detail::RDF;
 namespace TTraits = ROOT::TypeTraits;
 
-/// Smart pointer for the return type of actions
+/// Smart pointer for the return type of actions.
 /**
 \class ROOT::RDF::RResultPtr
 \ingroup dataframe
 \brief A wrapper around the result of RDataFrame actions able to trigger calculations lazily.
 \tparam T Type of the action result
 
-A smart pointer which allows to access the result of a RDataFrame action. The
-methods of the encapsulated object can be accessed via the arrow operator.
+A wrapper around a shared_ptr which allows to access the result of RDataFrame actions.
+The underlying object can be accessed by dereferencing the RResultPtr:
+~~~{.cpp}
+ROOT::RDF::RResultPtr<TH1D> histo = rdf.Histo1D(...);
+histo->Draw(); // Starts running the event loop
+~~~
 Upon invocation of the arrow operator or dereferencing (`operator*`), the
 loop on the events and calculations of all scheduled actions are executed
 if needed.
@@ -114,12 +118,17 @@ for (auto& myItem : myResultProxy) { ... };
 ~~~
 If iteration is not supported by the type of the proxied object, a compilation error is thrown.
 
+When shared ownership to the result is desired, a copy of the underlying shared_ptr can be obtained:
+~~~{.cpp}
+std::shared_ptr<TH1D> ProduceResult(const char *columnname) {
+   auto ht = rdf.Histo1D(*h, columname);
+   return ht.GetSharedPtr();
+}
+~~~
+Note that this will run the event loop. If this is not desired, the RResultPtr can be copied.
 */
 template <typename T>
 class RResultPtr {
-   // private using declarations
-   using SPT_t = std::shared_ptr<T>;
-
    // friend declarations
    template <typename T1>
    friend class RResultPtr;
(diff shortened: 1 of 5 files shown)


tree/dataframe/inc/ROOT/RResultPtr.hxx
```diff
@@ -223,33 +222,45 @@ public:
...
    T &operator*()
    {
       ThrowIfNull();
-      return *Get();
+      return *GetSharedPtr();
    }
 
    /// Get a pointer to the encapsulated object.
-   /// Ownership is not transferred to the caller.
+   /// \note: Ownership is not transferred to the caller.
```
ferdymercury, line 258:
```suggestion
   /// \note Ownership is not transferred to the caller.
```

guitargeek:
How does this address the usecase in the related issue?
```c++
    THStack* histStack = new THStack("histStack", "Bs_MCORR");
    for(int i=0; i<sampNumb; i++) {
        TFile* file = new TFile(rootList[i]);
        TTree* tree = (TTree*)file->Get("DecayTree");
        ROOT::RDataFrame df(*tree);
        auto ht = df.Filter(conds[i].Data()).Histo1D(h, "Bs_MCORR");
        histStack->Add(ht.GetPtr(), , "HIST && E1 && EX0");
   }
   histStack->Draw();
```

vepadulano (changes requested):
Thank you for addressing this issue! I believe the code changes are fine, but before merging it would be good to address the other two points mentioned in the linked issue:

* Documentation in the user guide. This is such a tricky part of the interaction with the API that I believe it's not enough to document it only in the RResultPtr class but it deserves a paragraph in the user guide, maybe after [this paragraph](https://root.cern/doc/master/classROOT_1_1RDataFrame.html#missing-values). It should have a description of the scenario with a somewhat realistic example, maybe a simplified version of the THStack example or something similar that shows the shared_ptr leaving some function scope so the RResultPtr would be destroyed.
* A new tutorial using either the forum post or the snippet I showed in the related issue

tutorials/analysis/dataframe/df039_RResultPtr_basics.C
```diff
@@ -0,0 +1,52 @@
...
+///
+/// \date 2025
+/// \author Stephan Hageboeck (CERN)
+
+#include <ROOT/RDataFrame.hxx>
+#include <ROOT/RDFHelpers.hxx>
+
+#include <iostream>
+
+// A function that adds a "lazy" histogram to a computation graph.
```
martamaja10, line 19: maybe one more short explanation what "lazy" is here just to be extra clear - something like the execution will not happen immediately

martamaja10 (changes requested):
Thank you Stephan! I mainly focused on the Tutorials which generally look very nice, I left a few comments to clarify a few things in the comments of the tutorials.

tutorials/analysis/dataframe/df039_RResultPtr_basics.C
```diff
@@ -0,0 +1,52 @@
...
+   std::cout << "Declared histo1\n";
+
+   // When RDF results are declared in functions, one has to choose if one wants run it to immediately or lazily:
+   auto rNode = ROOT::RDF::AsRNode(rdf);
+   ROOT::RDF::RResultPtr<TH1D> histo2 = histoLater(rNode);
+   std::cout << "Declared histo2\n";
+
+   std::shared_ptr<TH1D> histo3 = histoNow(rNode);
+   std::cout << "Declared histo3\n";
+}
```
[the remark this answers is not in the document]
hageboeck: Good point. Nothing is really different. I'll add a comment pointing this out.

martamaja10 (approved):
LGTM, thank you Stephan!

tree/dataframe/inc/ROOT/RResultPtr.hxx
```diff
@@ -223,33 +222,45 @@ public:
...
    {
       ThrowIfNull();
-      return *Get();
+      return *GetSharedPtr();
    }
 
    /// Get the pointer to the encapsulated object.
    /// Triggers event loop and execution of all actions booked in the associated RLoopManager.
-   T *GetPtr() { return Get(); }
+   /// \note: Ownership is not transferred to the caller.
```
ferdymercury, line 246:
```suggestion
   /// \note Ownership is not transferred to the caller.
```

Issue #9261, opened 2021-11-08
[cling][DF] `auto firstTwo = Take(v, 2)` causes error


ShamrockLee:
- [X] Checked for duplicates

### Describe the bug
```C++
ROOT::VecOps::RVec v {2., 3., 1.};
auto firstTwo = Take(v, 2);
```
causes error in the ROOT interactive environment for ROOT version 6.22/08 (released) and 6.25/01 (master).

### Expected behavior
The above expressions work in the interactive environment.
It should be guaranteed to work since it's documented as the example code in the [manual](https://root.cern/doc/master/group__vecops.html#gab314e8e57fba6274d8d3c652a24b2c52)

### To Reproduce
On ROOT version 6.22/08
```
$ root -l
root [0] ROOT::VecOps::RVec<double> v {2., 3., 1.};
root [1] Take(v, 2)
(ROOT::VecOps::RVec<double>) { 2.0000000, 3.0000000 }
root [2] ROOT::VecOps::RVec<double> firstTwo = Take(v, 2);
root [3] firstTwo
(ROOT::VecOps::RVec<double> &) { 2.0000000, 3.0000000 }
root [4] auto firstTwo2 = Take(v, 2);
ROOT_prompt_4:1:1: error: Syntax error
auto firstTwo2 = Take(v, 2);
^
FunctionDecl 0x3882650 <input_line_19:1:1, ROOT_prompt_4:3:1> input_line_19:1:6 __cling_Un1Qu38 'void (void *)'
|-ParmVarDecl 0x38825a0 <col:22, col:28> col:28 vpClingValue 'void *'
|-CompoundStmt 0x38829e0 <col:42, ROOT_prompt_4:3:1>
| |-DeclStmt 0x38829b8 <line:1:1, col:28>
| | `-VarDecl 0x3882728 <col:1, col:27> col:6 firstTwo2 'auto' cinit
| |   `-CallExpr 0x3882980 <col:18, col:27> '<dependent type>'
| |     |-DeclRefExpr 0x38828f0 <col:18> '<dependent type>' lvalue Var 0x3882860 'Take' '<dependent type>'
| |     |-DeclRefExpr 0x3882938 <col:23> 'ROOT::VecOps::RVec<double>':'class ROOT::VecOps::RVec<double>' lvalue Var 0x27808f8 'v' 'ROOT::VecOps::RVec<double>':'class ROOT::VecOps::RVec<double>'
[...]

PR #18170, opened 2025-03-27
[cling] Emit proper diagnostics in DynamicLookup for `AutoType`

devajithvs:
Replace AST dumps with proper error messages in dynamic lookup when dealing with `auto`, which is unsupported for lookup

Changes this:
```
root [0] auto var = someexpr();
ROOT_prompt_0:1:1: error: Syntax error
auto var = someexpr();
^
FunctionDecl 0x60cb530887f0 <input_line_8:1:1, ROOT_prompt_0:3:1> input_line_8:1:6 __cling_Un1Qu30 'void (void *)'
|-ParmVarDecl 0x60cb53088738 <col:22, col:28> col:28 vpClingValue 'void *'
|-CompoundStmt 0x60cb53088c30 <col:42, ROOT_prompt_0:3:1>
| |-DeclStmt 0x60cb53088c10 <line:1:1, col:22>
| | `-VarDecl 0x60cb53088910 <col:1, col:21> col:6 var 'auto' cinit
| |   `-CallExpr 0x60cb53088bc0 <col:12, col:21> '<dependent type>'
| |     `-DeclRefExpr 0x60cb53088b80 <col:12> '<dependent type>' lvalue Var 0x60cb53088aa8 'someexpr' '<dependent type>'
| `-NullStmt 0x60cb53088c28 <line:2:1>
|-AnnotateAttr 0x60cb53088a00 <<invalid sloc>> Implicit R"ATTRDUMP(__ResolveAtRuntime)ATTRDUMP"
`-AnnotateAttr 0x60cb53088b10 <<invalid sloc>> Implicit R"ATTRDUMP(__ResolveAtRuntime)ATTRDUMP"
<<<NULL>>>
```
to

```
root [0] auto var = someexpr();
ROOT_prompt_0:1:12: error: Cannot deduce 'auto' from unknown expression
auto var = someexpr();
           ^
```

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes # https://github.com/root-project/root/issues/9261


dpiparo:
CC: @silverweed

vgvassilev (approved):
Lgtm, if we add a test in clingtest.

interpreter/cling/test/Extensions/Lookup/StillError.C
```diff
@@ -28,4 +28,10 @@ std::unique_ptr<cling::test::SymbolResolverCallback> SRC;
 SRC.reset(new cling::test::SymbolResolverCallback(gCling, false))
 gCling->setCallbacks(std::move(SRC));
 p.q // CHECK: {{input_line_.*: error: use of undeclared identifier 'p'}}
+
+// `auto` is not supported
+auto x = unknownexpression()
+// CHECK: {{input_line_[0-9]+:[0-9]+:[0-9]+}}: error: cannot deduce 'auto' from unknown expression
```
vgvassilev, line 34: What about adding `-verify` and `//expected-error`?

devajithvs: If we have -verify, we would have to change that for p.q too, I tried to use something like
```
p.q // expected-error {{use of undeclared identifier 'p'}}

// `auto` is not supported
auto x = unknownexpression(); // expected-error {{cannot deduce 'auto' from unknown expression}}
```

But I get this (the string is the exact same, but the line number is different)
``
error: 'expected-error' diagnostics expected but not seen: 
  File input_line_27 Line 2: use of undeclared identifier 'p'
error: 'expected-error' diagnostics seen but not expected: 
  File input_line_28 Line 2: use of undeclared identifier 'p'
``

dpiparo: Is there a combination that guarantees coverage and works? We need pragmatism... It would be nice to have this fixed asap :)

hahnjo: As far as I understand, the file used `FileCheck` before. I agree it would be nice to use `-verify`, but I've seen other cases in the past where it was hard because of Cling line errors... Maybe this is one of the cases where you have to give the exact line number as `// expected-error@input_line_28 {{cannot deduce 'auto' from unknown expression}}` (and then sometimes suffer updating the line numbers when changing the test)

dpiparo:
@devajithvs do you think we can follow @hahnjo suggestion to move on, merge and close the associated issue?

devajithvs:
Done! The test is indeed very brittle and can break based on where // expected-error goes.

eg,
```
p.q // expected-error@input_line_28:2 {{use of undeclared identifier 'p'}}
```
will fail.

PR #18171, opened 2025-03-27
[ci] Format only changed lines with ruff

### LINKED ISSUE #18167: [ci] Apply Ruff Formatting to Changed Lines Only

Currently, Ruff operates at the file level when applying linting and formatting changes. However, in many use cases, especially when working on old large files, we may only want to target specific changed lines rather than reformatting the entire file and potentially changing code unrelated to our PR.

**Proposed Improvement:**
Introduce support for line-range-based formatting, allowing Ruff to process only the modified lines within a file.  This can be done with the `ruff format` range option:
```
Editor options:
      --range <RANGE>  When specified, Ruff will try to only format the code in the given range.
                       It might be necessary to extend the start backwards or the end forwards, to fully enclose a logical line.
                       The `<RANGE>` uses the format `<start_line>:<start_column>-<end_line>:<end_column>`.
```

### ROOT version

Any

### Installation method

Any

### Operating system

Any

### Additional context

_No response_

siliataider:
# This Pull request:

## Adds:
This PR updates the ruff formatting workflow to detect and act on the changed lines instead of the whole file.
On failure, the command to run is highlighted so that it can directly be copied and applied locally
image

**Note**: linting still applies to the full file.

Could be improved once astral-sh/ruff#12800 is implemented
This PR closes #18167

### DIFF
--- a/.github/workflows/code_analysis.yml
+++ b/.github/workflows/code_analysis.yml
@@ -67,7 +67,28 @@ jobs:
       run: |
         files=$(cat changed_files.txt | grep '\.py$' || echo "")
         if [ -n "$files" ]; then
-          echo "$files" | xargs ruff format --diff
+          diff_command=""
+          apply_command=""
+          for file in $files; do
+            while IFS=- read -r start length; do
+              [ -z "$start" ] && continue
+              length=${length:-1}
+              end=$((start + length))
+              diff_command+="ruff format --diff --range $start-$end $file && "
+              apply_command+="ruff format --range $start-$end $file && "
+            done < <(git diff --unified=0 origin/$GITHUB_BASE_REF "$file" | grep '^@@' | sed -E 's/^@@ -[0-9]+(,[0-9]+)? \+([0-9]+)(,([0-9]+))? @@.*/\2-\4/')
+          done
+
+          if [ -n "$diff_command" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/code_analysis.yml
```diff
@@ -67,7 +67,28 @@ jobs:
...
+              diff_command+="ruff format --diff --range $start-$end $file && "
+              apply_command+="ruff format --range $start-$end $file && "
+            done < <(git diff --unified=0 origin/$GITHUB_BASE_REF "$file" | grep '^@@' | sed -E 's/^@@ -[0-9]+(,[0-9]+)? \+([0-9]+)(,([0-9]+))? @@.*/\2-\4/')
+          done
+
+          if [ -n "$diff_command" ]; then
+            diff_command=${diff_command% && }
+            if ! eval "$diff_command"; then
+              apply_command=${apply_command% && }
+              echo "::error::Formatting failed. To apply the changes locally, run the following command: $apply_command"
```
silverweed, line 86: I would add a newline before `$apply_command` so it's more easily distinguishable from the error message

vepadulano (approved):
LGTM, thanks!

silverweed (commented):
Thank you! I left a small comment

silverweed (approved):

PR #18173, opened 2025-03-27
[cppyy] Enable upstream test suite

aaronj0:
This PR enables the upstream cppyy test suite, allowing ROOT to test C++ features supported in its Python bindings with minimal C++ test cases. This also lets us track where ROOT regresses from upstream, e.g  multi inheritance, and currently crashing tests, marked `skip` with pytest

TODO

- [ ] Add CMake that builds the test dictionaries, and registers python tests
- [ ] Drop early versions of these tests from`roottest/python/` 
- [ ] Add patch files tracking these changes over upstream


guitargeek:
Hi! How can I run the tests if this PR would be merged?

aaronj0:
> Hi! How can I run the tests if this PR would be merged?

Currently it follows the cppyy way: 

```
cd root/bindings/pyroot/cppyy/cppyy/test/
make all
python -m pytest
```

The next step would be to move to CMake and run as a part of ctest, and then drop the legacy tests

guitargeek (approved):
Great, thank you very much! I have tested that I can run the tests as you describe. Looking forward to the follow-up PRs :)

PR #18175, opened 2025-03-27
[ntuple] Move REntry out of Experimental

silverweed:
# This Pull request:
Moves `REntry` out of Experimental.

## Checklist:

- [x] tested changes locally


tree/ntuple/v7/inc/ROOT/RField.hxx
```diff
@@ -44,14 +44,14 @@ class TVirtualStreamerInfo;
 namespace ROOT {
 
 class TSchemaRule;
+class REntry;
```
hahnjo, line 47: probably not needed and can be removed (now or in follow-up PR)

hahnjo (approved):
LGTM

PR #18178, opened 2025-03-27
Make basket size option a serializable type

vepadulano:
The type given to the fBasketSize member of the RSnapshotOptions, std::optional<int>, is not serializable by ROOT. This can lead to warnings and potentially faulty behaviour in distributed RDataFrame (and similar scenarios that require objects to be serializable). For example, this can be seen in the test output:

TStreamerInfo::Build:0: RuntimeWarning: ROOT::RDF::RSnapshotOptions: optional<int> has no streamer or dictionary, data member "fBasketSize" will not be saved
TStreamerInfo::Build:0: RuntimeWarning: _Optional_base<int,true,true>: base class _Optional_base_impl<int,_Optional_base<int,true,true> > has no streamer or dictionary it will not be saved
TStreamerInfo::Build:0: RuntimeWarning: _Optional_base<int,true,true>: _Optional_payload<int,true,true,true> has no streamer or dictionary, data member "_M_payload" will not be saved


martamaja10 (approved):
LGTM, thanks Vincenzo!

PR #18179, opened 2025-03-27
[CPyCppyy] Use macros from the C Python API for better compatibility

guitargeek:
Use macros from the C Python API for better compatibility between Python versions and also to compile with Python builds with free threading (aka. without the GIL).


vepadulano (approved):
Thanks!

PR #18182, opened 2025-03-28
[HistFactory] Disable global histogram registration for HistFactory.

### LINKED ISSUE #18172: [RF] Inefficient un-registering of histograms in HistFactory

While doing some profiling check with VTune with @guitargeek we noticed that in some extreme cases the bottleneck can come from the un-registering of histograms during the creation of the WS in the HistFactory `MakeModelAndMeasurementFast()` method.

Please, see the attached result of VTune.

Image

Image

### ROOT version

ROOT Version: 6.35.01
Built for linuxx8664gcc on Mar 27 2025, 14:27:43
From heads/master@1c94b446

### Installation method

build from source with debug symbols on

### Operating system

Ubuntu 24.10

### Additional context

_No response_

hageboeck:
The HistFactory histograms are owned by the HistFactory models, and they are explicitly written to files. Therefore, they should not participate in directory registration.

Fix #18172

### DIFF
--- a/roofit/histfactory/src/MakeModelAndMeasurementsFast.cxx
+++ b/roofit/histfactory/src/MakeModelAndMeasurementsFast.cxx
@@ -83,6 +83,7 @@ RooFit::OwningPtr<RooWorkspace>
 RooStats::HistFactory::MakeModelAndMeasurementFast(RooStats::HistFactory::Measurement &measurement,
                                                    HistoToWorkspaceFactoryFast::Configuration const &cfg)
 {
+  TDirectory::TContext dirContext{nullptr};
   std::unique_ptr<TFile> outFile;
 
   auto& msgSvc = RooMsgService::instance();
@@ -133,7 +134,8 @@ RooStats::HistFactory::MakeModelAndMeasurementFast(RooStats::HistFactory::Measur
     // This holds the TGraphs that are created during the fit
     std::string outputFileName = measurement.GetOutputFilePrefix() + "_" + measurement.GetName() + ".root";
     cxcoutIHF << "Creating the output file: " << outputFileName << std::endl;
-    outFile = std::make_unique<TFile>(outputFileName.c_str(), "recreate");
+    outFile = std::make_unique<TFile>(outputFileName.c_str(), "RECREATE_WITHOUT_GLOBALREGISTRATION");
+    gDirectory = nullptr; // Prevent global registration of histograms. They are owned by HF.
 
     cxcoutIHF << "Creating the HistoToWorkspaceFactoryFast factory" << std::endl;
     HistoToWorkspaceFactoryFast factory{measurement, cfg};
--- a/roofit/histfactory/src/Channel.cxx
+++ b/roofit/histfactory/src/Channel.cxx
@@ -464,8 +464,8 @@ TH1* RooStats::HistFactory::Channel::GetHistogram(std::string InputFile, std::st
     throw hf_exc();
   }
 
-
-  TH1 * ptr = static_cast<TH1 *>(hist->Clone());
+  TDirectory::TContext ctx{nullptr}; // Suppress any kind of registration
+  TH1 *ptr = static_cast<TH1 *>(hist->Clone());
 
   if(!ptr){
     std::cerr << "Not all necessary info are set to access the input file. Check your config" << std::endl;
@@ -475,6 +475,7 @@ TH1* RooStats::HistFactory::Channel::GetHistogram(std::string InputFile, std::st
     throw hf_exc();
   }
 
+  ptr->SetBit(kMustCleanup, false); // Don't try to clean up this histogram
   ptr->SetDirectory(nullptr);
 
 


roofit/histfactory/src/MakeModelAndMeasurementsFast.cxx
```diff
@@ -83,6 +83,8 @@ RooFit::OwningPtr<RooWorkspace>
 RooStats::HistFactory::MakeModelAndMeasurementFast(RooStats::HistFactory::Measurement &measurement,
                                                    HistoToWorkspaceFactoryFast::Configuration const &cfg)
 {
+  const bool oldAddDirectoryStatus = TH1::AddDirectoryStatus();
```
pcanal, line 86: This is thread unfriendly and we should find a better solution: either 'TDirectory::TContext ctxt(nullptr);` or calling `SetDirectory(nullptr)` on each directory.

hageboeck: Thanks Philippe, but I don't think that's the solution. First of all, creating a HistFactory model is not thread safe. It was never part of the design.

If I understand correctly, you suggest to set `gDirectory` to nullptr, and restore it to the value before the instance was created on destruction. That's actually not what's needed, because HistFactory is based on the following workflow:
```c++
dirA->cd();
histA->Write();

dirB->cd();
histB1->Write();
histB2->Write();
// ...
```
It therefore relies on `gDirectory` having some value, which is set just before the writes. Setting it to `nullptr` using the context wouldn't have any effect, as it's overwritten in line 1 anyway.

Your second suggestion to manually set all TH1::SetDirectory is possible, but there are a lot of places to be touched. Also, there is overhead, because each TH1 will register itself on construction, and we will deregister it immediately after that.

What I actually want is that the TH1s don't register themselves. At all. The solution should be `TH1::AddDirectory(false)`. Given the code above, we would have:
```c++
TH1::AddDirectory(false);

dirA->cd();
histA->Write();

dirB->cd();
histB1->Write();
histB2->Write();
// ...
TH1::AddDirectory(previous);
```
If you want, I can make the restoration of the flag into a RAII, but I think `gDirectory = nullptr` has no effect.

hageboeck: Also to give context where the function would typically be called:
- From `hist2workspace`, which runs in isolation, independent of all other ROOT tasks.
- Or by user code that is trying to output a HistFactory model. It's highly unlikely that the user would do anything else concurrently.

pcanal:
```
dirA->cd();
histA->Write();

dirB->cd();
histB1->Write();
histB2->Write();
// ...
```
Actually this is exactly what `TDirectory::TContext` supports.  It "only" set gDirectory to nullptr at 'that' moment.
Then you might also need to use the pattern:
```
{
  TDirectory::TContext ctxt();
  auto f = TFile::Open(...);
}
```
or simply
```
  auto f = TFile::Open(...);
  gDirectory = nullptr;
```
(the later still require one starting `TContext` to return `gDirectory` to what it was at the start.)

> First of all, creating a HistFactory model is not meant thread safe. It was never part of the design.

Yes, but do you actively prevent completely unrelated ROOT code to be run in another thread?

roofit/histfactory/src/MakeModelAndMeasurementsFast.cxx
```diff
@@ -83,7 +83,9 @@ RooFit::OwningPtr<RooWorkspace>
 RooStats::HistFactory::MakeModelAndMeasurementFast(RooStats::HistFactory::Measurement &measurement,
                                                    HistoToWorkspaceFactoryFast::Configuration const &cfg)
 {
+  TDirectory::TContext dirContext;
   std::unique_ptr<TFile> outFile;
+  gDirectory = nullptr; // Disable global registration of histograms this file. HistFactory places them explicitly later.
```
pcanal, line 88:
```suggestion
  TDirectory::TContext dirContext{nullptr} // Disable global registration of histograms this file. HistFactory places them explicitly later.
  std::unique_ptr<TFile> outFile;
```
Since the next line does not allocated a `TFile` and thus does not change `gDirectory`

pcanal: However line 138:
```
outFile = std::make_unique<TFile>(outputFileName.c_str(), "recreate");
```
should probably be:
```
outFile = std::make_unique<TFile>(outputFileName.c_str(), "RECREATE_WITHOUT_GLOBALREGISTRATION") // note: casing does not matter.
gDirectory = nullptr;  // Make sure the files does not become the current directory.  We do not need TContext here since we have one on the outer scope.
```
Side note the `_WITHOUT_GLOBALREGISTRATION` is probably the only thing strictly necessary for the goal of this PR (i.e. since it is not registered, it is not on the list of Cleanups and thus will not be traversed on delete of the histograms).

TomasDado:
Hello, the execution is slightl faster from ~260s to about 230s but it seems most of the time is stil; spent in the destructors
```
tomas@tomas-IdeaPad-5-Pro-14IAP7:~/TRExFitter$ root --version
ROOT Version: 6.35.01
Built for linuxx8664gcc on Mar 31 2025, 07:22:16
From remotes/hageboeck/histFactoryCleanups@7b0297d5e9
```
Screenshot From 2025-03-31 09-47-28

roofit/histfactory/src/MakeModelAndMeasurementsFast.cxx
hageboeck: Oh, I had no idea this option existed! That's really good to know!

hageboeck:
> Hello, the execution is slightly faster from ~260s to about 230s but it seems most of the time is still spent in the destructors
> 
Hello @TomasDado, my attempt at changing it was mostly without effect as pointed out by Philippe. I force-pushed a new version of the commit that should have larger impact. Could you try again, please?

TomasDado:
> > Hello, the execution is slightly faster from ~260s to about 230s but it seems most of the time is still spent in the destructors
> 
> Hello @TomasDado, my attempt at changing it was mostly without effect as pointed out by Philippe. I force-pushed a new version of the commit that should have larger impact. Could you try again, please?

Unfortunately, it still looks very similar, runnig time of about 230s:
```
tomas@tomas-IdeaPad-5-Pro-14IAP7:~/TRExFitter$ root --version
ROOT Version: 6.35.01
Built for linuxx8664gcc on Mar 31 2025, 11:12:47
From remotes/hageboeck/histFactoryCleanups@18a27e3303
```

hageboeck:
Closing in favour of #18566

PR #18184, opened 2025-03-28
Implement TGraph2D::RemoveDuplicates

couet:
#The Delaunay triangulation algorithm assumes that each (x, y) coordinate corresponds to a unique z value,
meaning duplicate (x, y) points are not allowed. Consequently, when using drawing options that rely on this
algorithm (e.g., TRI, SURF, etc.), a warning may appear instructing you to remove duplicates.
This function provides a way to handle such duplicates.


linev (approved):
Looks ok

PR #18185, opened 2025-03-28
[ntuple] Add `GetView<void>` with type name string or `std::type_info` argument

enirolf:
Enables passing a type name string to `GetView<void>` to signal that the pointer has a different underlying type than the on-disk field's type. The field used by `RNTupleView` will be constructed with this type, in turn benefitting from RNTuples built-in type- and bounds-checking.

To illustrate further, assuming `myInt` is a `std::uint64_t` on-disk:
```cpp
std::int32_t myInt;
void* myIntPtr = &myInt;
auto myIntView = reader->GetView<void>("myInt", myIntPtr);
myIntView(0);
doSomething(myInt);
```
`GetView<void>` will currently always reconstruct the field used for loading the values from the on-disk descriptor. However, in the example above this can lead to undefined behavior when the actual values on disk don't fit in `std::int32_t`. The change proposed in this PR enables passing the type name string to `GetView<void>`, which is then used for the construction of the field instead, leveraging RNTuple's internal type- and bounds-checking:
```cpp
std::int32_t myInt;
void* myIntPtr = &myInt;
auto myIntView = reader->GetView<void>("myInt", myIntPtr, "std::int32_t");
myIntView(0);
doSomething(myInt);
```

This feature was requested by and discussed with ATLAS.

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
@@ -296,6 +296,12 @@ public:
       return GetView<T>(RetrieveFieldId(fieldName), objPtr);
    }
 
+   template <typename T>
+   ROOT::RNTupleView<T> GetView(std::string_view fieldName, std::shared_ptr<T> objPtr, std::string_view typeName)
+   {
+      return GetView<T>(RetrieveFieldId(fieldName), objPtr, typeName);
+   }
+
    /// Provides access to an individual (sub)field, reading its values into `rawPtr`.
    ///
    /// \sa GetView(std::string_view, std::shared_ptr<T>)
@@ -303,6 +309,12 @@ public:
    ROOT::RNTupleView<T> GetView(std::string_view fieldName, T *rawPtr)
    {
       return GetView<T>(RetrieveFieldId(fieldName), rawPtr);
+   }
+
+   template <typename T>
+   ROOT::RNTupleView<T> GetView(std::string_view fieldName, T *rawPtr, std::string_view typeName)
+   {
+      return GetView<T>(RetrieveFieldId(fieldName), rawPtr, typeName);
    }
 
    /// Provides access to an individual (sub)field from its on-disk ID.
@@ -327,6 +339,15 @@ public:
       return ROOT::RNTupleView<T>(std::move(field), range, objPtr);
    }
 
+   template <typename T>
+   ROOT::RNTupleView<T> GetView(ROOT::DescriptorId_t fieldId, std::shared_ptr<T> objPtr, std::string_view typeName)
+   {
+      static_assert(std::is_void_v<T>, "calling GetView with a type name string is only allowed for [T = void]");
+      auto field = RNTupleView<T>::CreateField(fieldId, *fSource, typeName);
+      auto range = ROOT::Internal::GetFieldRange(*field, *fSource);
+      return RNTupleView<T>(std::move(field), range, objPtr);
+   }
+
    /// Provides access to an individual (sub)field from its on-disk ID, reading its values into `rawPtr`.
    ///
    /// \sa GetView(std::string_view, std::shared_ptr<T>)
@@ -336,6 +357,15 @@ public:
       auto field = ROOT::RNTupleView<T>::CreateField(fieldId, *fSource);
       auto range = ROOT::Internal::GetFieldRange(*field, *fSource);
       return ROOT::RNTupleView<T>(std::move(field), range, rawPtr);
+   }
+
+   template <typename T>
+   ROOT::RNTupleView<T> GetView(ROOT::DescriptorId_t fieldId, T *rawPtr, std::string_view typeName)
+   {
+      static_assert(std::is_void_v<T>, "calling GetView with a type name string is only allowed for [T = void]");
+      auto field = RNTupleView<T>::CreateField(fieldId, *fSource, typeName);
+      auto range = ROOT::Internal::GetFieldRange(*field, *fSource);
+      return RNTupleView<T>(std::move(field), range, rawPtr);
    }
 
    /// Provides direct access to the I/O buffers of a **mappable** (sub)field.
--- a/tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
@@ -308,19 +308,17 @@ public:
    /// Provides access to an individual (sub)field, reading its values into `rawPtr` as the type provided by `typeName`.
    ///
    /// \sa GetView(std::string_view, std::shared_ptr<T>)
-   template <typename T>
-   ROOT::RNTupleView<T> GetView(std::string_view fieldName, T *rawPtr, std::string_view typeName)
+   ROOT::RNTupleView<void> GetView(std::string_view fieldName, void *rawPtr, std::string_view typeName)
    {
-      return GetView<T>(RetrieveFieldId(fieldName), rawPtr, typeName);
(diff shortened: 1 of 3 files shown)


tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
```diff
@@ -291,12 +291,24 @@ public:
       return GetView<T>(RetrieveFieldId(fieldName), objPtr);
    }
 
+   template <typename T>
+   RNTupleView<T> GetView(std::string_view fieldName, std::shared_ptr<T> objPtr, std::string_view typeName)
```
silverweed, line 295: Any specific reason for not making `typeName` a defaulted param in the existing `GetView`?

silverweed (approved):
LGTM, just one question that probably doesn't belong to this PR

tree/ntuple/v7/inc/ROOT/RNTupleView.hxx
```diff
@@ -64,14 +64,18 @@ protected:
...
       std::unique_ptr<ROOT::RFieldBase> field;
       {
          const auto &desc = pageSource.GetSharedDescriptorGuard().GetRef();
          const auto &fieldDesc = desc.GetFieldDescriptor(fieldId);
          if constexpr (std::is_void_v<T>) {
-            field = fieldDesc.CreateField(desc);
+            if (typeName.empty())
+               field = fieldDesc.CreateField(desc);
+            else
+               field = ROOT::RFieldBase::Create(fieldDesc.GetFieldName(), std::string(typeName)).Unwrap();
```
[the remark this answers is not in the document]
enirolf: Probably, yes. I'll open a PR.

hahnjo (approved):
LG, thanks for the clear implementation!

tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
```diff
@@ -313,6 +325,15 @@ public:
       return RNTupleView<T>(std::move(field), range, objPtr);
    }
 
+   template <typename T>
+   RNTupleView<T> GetView(ROOT::DescriptorId_t fieldId, std::shared_ptr<T> objPtr, std::string_view typeName)
+   {
+      static_assert(std::is_void_v<T>, "calling GetView with an type name string is only allowed for [T = void]");
```
pcanal, line 331:
```suggestion
      static_assert(std::is_void_v<T>, "calling GetView with a type name string is only allowed for [T = void]");
```

tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
```diff
@@ -321,6 +342,15 @@ public:
       return RNTupleView<T>(std::move(field), range, rawPtr);
    }
 
+   template <typename T>
+   RNTupleView<T> GetView(ROOT::DescriptorId_t fieldId, T *rawPtr, std::string_view typeName)
+   {
+      static_assert(std::is_void_v<T>, "calling GetView with an type name string is only allowed for [T = void]");
```
pcanal, line 348:
```suggestion
      static_assert(std::is_void_v<T>, "calling GetView with a type name string is only allowed for [T = void]");
```

tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
```diff
@@ -293,12 +293,36 @@ public:
       return GetView<T>(RetrieveFieldId(fieldName), objPtr);
    }
 
+   template <typename T>
+   ROOT::RNTupleView<T> GetView(std::string_view fieldName, std::shared_ptr<T> objPtr, const std::type_info &ti)
```
silverweed, line 297: I just noticed we don't have any documentation on the `GetView` methods that accept a pointer...maybe we should add it (even just to one of them and point the other methods' doc to it)

jblomer (approved):
Many thanks! LGTM!

tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
enirolf: See #18261 for the already existing ones, will do the same for these new additions afterwards!

tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
```diff
@@ -324,7 +363,22 @@ public:
...
+   ROOT::RNTupleView<T> GetView(ROOT::DescriptorId_t fieldId, T *rawPtr, std::string_view typeName)
+   {
+      static_assert(std::is_void_v<T>, "calling GetView with a type name string is only allowed for [T = void]");
+      auto field = RNTupleView<T>::CreateField(fieldId, *fSource, typeName);
+      auto range = ROOT::Internal::GetFieldRange(*field, *fSource);
+      return RNTupleView<T>(std::move(field), range, rawPtr);
+   }
+
+   template <typename T>
+   RNTupleDirectAccessView<T> GetDirectAccessView(std::string_view fieldName)
```
vepadulano, line 381: I've noticed this compiler failure on Debian, probably we need `ROOT::RNtupleDirectAccessView`?

vepadulano (approved):
Thank you! But see comment before merging

tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
```diff
@@ -300,6 +336,29 @@ public:
...
    }
 
+   /// Provides access to an individual (sub)field from its on-disk ID, reading its values into `objPtr` as the type
+   /// provided by `typeName`.
+   ///
+   /// \sa GetView(std::string_view, std::shared_ptr<T>)
+   template <typename T>
+   ROOT::RNTupleView<T> GetView(ROOT::DescriptorId_t fieldId, std::shared_ptr<T> objPtr, std::string_view typeName)
+   {
+      static_assert(std::is_void_v<T>, "calling GetView with a type name string is only allowed for [T = void]");
```
hahnjo, line 346: Actually, if we only allow `T = void`, does it make sense to have this function templated at all? It would deviate from other overloads, but I find it less surprising than the documentation telling me it's a template function and then being presented a `static_assert`...

tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
```diff
@@ -300,6 +336,29 @@ public:
       return ROOT::RNTupleView<T>(std::move(field), range, objPtr);
    }
 
+   /// Provides access to an individual (sub)field from its on-disk ID, reading its values into `objPtr` as the type
+   /// provided by `typeName`.
+   ///
+   /// \sa GetView(std::string_view, std::shared_ptr<T>)
+   template <typename T>
+   ROOT::RNTupleView<T> GetView(ROOT::DescriptorId_t fieldId, std::shared_ptr<T> objPtr, std::string_view typeName)
```
hahnjo, line 344: ... and in that light, do we expect usage of `std::shared_ptr<void>` from framework code?

tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
enirolf: Yeah this is actually something I was asking myself as well when I was writing the docs. Hooking into your other comment, I don't think it is used in framework code so until it is I'm also fine removing it here.

hahnjo:
As discussed on Friday, I removed the `shared_ptr` overload and un-templated `GetView` instead of the `static_assert` that `T = void`

vepadulano (approved):
Changed as discussed, LGTM

PR #18186, opened 2025-03-28
Replace `SavePrimitiveArray` by `SavePrimitiveVector`

linev:
Plain array is allocated in stack and therefore depending on platform maximal array size is limited.
Therefore instead of storing array store vector like:
```
   std::vector<Double_t> hp0__1_vect0{
      1, 0, 0, 0, 0, 0, 0, 0, 0, 0,
      0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
      0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
      0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
      0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
      0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
      0, 0
   };
```
Adjust all places where array was used and remove `SavePrimitiveArray` - it was introduced several weeks ago and was never released.

### DIFF
--- a/core/base/src/TObject.cxx
+++ b/core/base/src/TObject.cxx
@@ -779,37 +779,6 @@ void TObject::SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char
    out << variable_name << " = new " << cl->GetName() << "(" << constructor_agrs << ");\n";
 }
 
-////////////////////////////////////////////////////////////////////////////////
-/// Save array in the output stream "out".
-/// Create unique variable name based on prefix value
-
-TString TObject::SavePrimitiveArray(std::ostream &out, const char *prefix, Int_t len, Double_t *arr, Bool_t empty_line)
-{
-   thread_local int arrid = 0;
-
-   TString arrname = TString::Format("%s_darr%d", prefix, arrid++);
-
-   if (empty_line)
-      out << "   \n";
-
-   out << "   Double_t " << arrname << "[" << len << "] = { ";
-   if (len > 0) {
-      const auto old_precision{out.precision()};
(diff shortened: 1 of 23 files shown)


core/base/src/TObject.cxx
```diff
@@ -779,36 +779,40 @@ void TObject::SavePrimitiveConstructor(std::ostream &out, TClass *cl, const char
...
+   thread_local int vectid = 0;
 
-   TString arrname = TString::Format("%s_darr%d", prefix, arrid++);
+   TString vectame = TString::Format("%s_vect%d", prefix, vectid++);
 
    if (empty_line)
       out << "   \n";
 
-   out << "   Double_t " << arrname << "[" << len << "] = { ";
+   out << "   std::vector<Double_t> " << vectame;
```
ferdymercury, line 797: why not std::array<Double_t, len> instead of std vector?

linev: Yes, std::array allocated in stack, which is much smaller than heap.
My doubt - if proposed syntax also has some side-effects. If interim array allocated in stack - one also will get same problems. I will try to test it.

linev: `TGraph` with 500K points works  - at least creation and storage of such object in macro.
Painting is other story.

pcanal: >  If interim array allocated in stack 

It should not, it should store directly in the object file data segment.  

Another alternative would have been to use a function static std::array (or regular array).

pcanal (approved):

PR #18187, opened 2025-03-28
[ntuple] Merger: fix improper output column id assignment in incremental merging

silverweed:
We were creating new output column ids even when doing incremental merging, which may end up associating the wrong output column to an input column that is already present in the destination. By prefilling the column id map with the destination columns we avoid this problem.

## Checklist:

- [x] tested changes locally

This PR fixes [this use case](https://gitlab.cern.ch/amete/rootparallelmerger/-/blob/master-rntuple-prototype) by @amete

### DIFF
--- a/tree/ntuple/v7/src/RNTupleMerger.cxx
+++ b/tree/ntuple/v7/src/RNTupleMerger.cxx
@@ -996,6 +996,25 @@ static RColumnInfoGroup GatherColumnInfos(const RDescriptorsComparison &descCmp,
    return res;
 }
 
+static void PrefillColumnMap(const ROOT::RNTupleDescriptor &desc, const ROOT::RFieldDescriptor &fieldDesc,
+                             ColumnIdMap_t &colIdMap, const std::string &prefix = "")
+{
+   std::string name = prefix + '.' + fieldDesc.GetFieldName();
+   for (const auto &colId : fieldDesc.GetLogicalColumnIds()) {
+      const auto &colDesc = desc.GetColumnDescriptor(colId);
+      RColumnOutInfo info{};
+      const auto colName = name + '.' + std::to_string(colDesc.GetIndex());
+      info.fColumnId = colDesc.GetLogicalId();
+      info.fColumnType = colDesc.GetType();
+      colIdMap[colName] = info;
+   }
+
+   for (const auto &subId : fieldDesc.GetLinkIds()) {
+      const auto &subfield = desc.GetFieldDescriptor(subId);
+      PrefillColumnMap(desc, subfield, colIdMap, name);
+   }
+}
+
 RNTupleMerger::RNTupleMerger(std::unique_ptr<RPagePersistentSink> destination,
                              std::unique_ptr<ROOT::RNTupleModel> model)
    // TODO(gparolini): consider using an arena allocator instead, since we know the precise lifetime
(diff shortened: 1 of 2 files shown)


tree/ntuple/v7/src/RNTupleMerger.cxx
```diff
@@ -996,6 +996,25 @@ static RColumnInfoGroup GatherColumnInfos(const RDescriptorsComparison &descCmp,
    return res;
 }
 
+static void PrefillColumnMap(const ROOT::RNTupleDescriptor &desc, const ROOT::RFieldDescriptor &fieldDesc,
+                             ColumnIdMap_t &colIdMap, const std::string &prefix = "")
+{
+   std::string name = prefix + '.' + fieldDesc.GetFieldName();
```
pcanal, line 1002: Is that the correct behavior if the prefix has the default value?  i.e. the name 'starting' with a period?

silverweed: Yes, it's the same behavior as [line 909](https://github.com/root-project/root/pull/18187/files#diff-f47e9d804f32081f46022555af6dd04fd77b8f6a389a175d953e693aeabc90bbL909). (since it's just an internal name it's fine if it looks a bit odd)

tree/ntuple/v7/test/ntuple_merger.cxx
```diff
@@ -2997,3 +3000,219 @@ TEST(RNTupleMerger, MergeSecondEmptySchema)
...
+         EXPECT_FLOAT_EQ(*pnFoo, expFoo);
+         EXPECT_FLOAT_EQ(*pnBar, expBar);
+      }
+   }
+
+   fileGuardMerged.PreserveFile();
+}
+
+namespace {
+// Adapted from ATLAS example: https://gitlab.cern.ch/amete/rootparallelmerger/-/tree/master-rntuple-prototype
```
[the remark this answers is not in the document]
silverweed: The idea was to test the case of merging RNTuples alongside other objects, though I guess that can be an integration test.

tree/ntuple/v7/src/RNTupleMerger.cxx
silverweed: Well, column names are not a thing in RNTuple, so in respect to what are they "invalid"?

hahnjo: Oh yeah, that's right. I was thinking about field names, but in the end a field can have multiple columns etc. Fine from my side.

tree/ntuple/v7/test/ntuple_merger.cxx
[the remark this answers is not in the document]
hahnjo: Yes, sounds like an integration test. To be considered though, the test compares RNTuple and TTree - I'm not sure we want this in the long run...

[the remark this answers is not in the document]
pcanal: I think the `TTree` test is actually useful (and missing) so I would definitively keep it.  The test involving both `TTree` and `RNTuple` should probably be moved to `roottest/root/rntuple`

tree/ntuple/v7/src/RNTupleMerger.cxx
pcanal: > (since it's just an internal name it's fine if it looks a bit odd)

Are we guaranteed those names won't leak anywhere (eg. displaying meta-data information about the `RNTuple`)?

silverweed: They are printed if you run the Merger in ExtraVerbose mode, but that's mostly for our own debugging. In the future we may want to improve the ExtraVerbose mode messages if there is the need for that.

PR #18191, opened 2025-03-28
Remove current RHist prototype

hahnjo:
For a number of reasons, it is not what we want for the future generation of histogram classes in ROOT:

 * It is templated on the `DIMENSIONS`, the `PRECISION`, and optional bin statistics. This makes it hard(er) to use for IO.
 * The `RHist` interface has a pointer to an abstract, polymorphic `RHistImplBase`.
 * The concrete `RHistImpl` is templated on the processed axis types.
 * The polymorphic pointer makes the current prototype incompatible with RNTuple (unless using streamer fields).
 * It also takes minutes to compile for higher dimensions because concrete `RHistImpl` have to be generated for all combinations of axis types, which grows exponentially.

The plan is to remove the current prototype now, in time for the next ROOT release v6.36, and then add a new implementation after.


linev:
There is significant peace of code in JSROOT. Do we want to keep it?

hahnjo:
> There is significant peace of code in JSROOT. Do we want to keep it?

Yes, I saw it but I'm not expert enough in that code. We may want to remove it as well, I'm not sure how much could (or even should) be reused with a future prototype

tutorials/experimental/rcanvas/rcanvas_mt.cxx
```diff
@@ -1,123 +0,0 @@
...
-#include "TRandom3.h"
-#include "TEnv.h"
-#include "TROOT.h"
-
-#include <thread>
-#include <iostream>
-
-using namespace ROOT::Experimental;
-
-void draw_canvas(const std::string &title, RColor col)
```
pcanal, line 38: @linev This tutorials seems to be about using `RCanvas` from multiple threads.  Rather than being removed, should be update to use the old `TH1D` class?

gui/browsable/src/RProvider.cxx
```diff
@@ -509,10 +509,6 @@ class RDefaultProvider : public RProvider {
    RDefaultProvider()
    {
       // TODO: let read from rootrc or any other files
-      RegisterClass("ROOT::Experimental::RH1D", "sap-icon://bar-chart", "", "", "libROOTHistDrawProvider");
-      RegisterClass("ROOT::Experimental::RH2D", "sap-icon://pixelate", "", "", "libROOTHistDrawProvider");
-      RegisterClass("ROOT::Experimental::RH3D", "sap-icon://product", "", "", "libROOTHistDrawProvider");
-      RegisterClass("ROOT::Experimental::RCanvas", "sap-icon://business-objects-experience", "", "", "libROOTHistDrawProvider");
```
pcanal, line 515: That smell odd to remove a line about `RCanvas` (but it also smell odd-ish that it ask for `libROOTHistDrawProvider`)

tutorials/experimental/rcanvas/rcanvas_mt.cxx
linev: Yes, one can create such tutorial with `TH1` class. But one can remove old one as first step.

gui/browsable/src/RProvider.cxx
linev: Yes, this is automatic loading of libs to handle `RH1D`, `RH2D`, `RH3D` classes. 
This should be removed from RBrowser as well, but this can be done after merging this PR

linev:
> Yes, I saw it but I'm not expert enough in that code. We may want to remove it as well, I'm not sure how much could (or even should) be reused with a future prototype

It is very much depends how future prototype will looks like. 
Probably I will remove code from repository for next ROOT/JSROOT releases, 
but one can reuse it later again. Many draw options (like lego) reuse same code anyway for TH2 and RH2.

linev (approved):

hageboeck:
> It is very much depends how future prototype will looks like.
> Probably I will remove code from repository for next ROOT/JSROOT releases,
> but one can reuse it later again.

Indeed. The code can still be found and checked out and adapted after removal. We would only lose the ability to `git log` the files. Is that OK @linev ?

hahnjo:
Build failure on `mac-beta` is related to latest SDK update

PR #15183, opened 2024-04-09
[CI] Add `xgboost` and `scikit-learn` to requirements

guitargeek:
This should be added so we can test RBDT in the CI.

PR #18192, opened 2025-03-28
[Python] Fix RDF Pythonization tests with `builtin_llvm=OFF`

guitargeek:
On my machine, the RDF pythonization tests don't run when I build ROOT with `builtin_llvm=OFF`. There is a crash unless I do `import numba` in the beginning.

I guess it's not a priority to understand the underlying problem because not many people build ROOT like this and the workaround is easy, but at least the tests should be green.

We do the same "magically ordered imports" also in other Python tests where ROOT doesn't work together with xgboost because of `std::regexp` symbol clashes, unless you do the import in a certain order:

https://github.com/root-project/root/pull/15183

### DIFF
--- a/bindings/pyroot/pythonizations/test/rdf_define_pyz.py
+++ b/bindings/pyroot/pythonizations/test/rdf_define_pyz.py
@@ -1,6 +1,12 @@
 import unittest
-import ROOT
+
+import numba  # noqa: F401
 import numpy as np
+import ROOT
+
+# numba is not used directly, but tests can crash when ROOT is built with
+# builtin_llvm=OFF and numba is not imported at the beginning
+
 
 class PyDefine(unittest.TestCase):
     """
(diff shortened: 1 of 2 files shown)


bindings/pyroot/pythonizations/test/rdf_define_pyz.py
```diff
@@ -1,6 +1,13 @@
 import unittest
-import ROOT
+
+import numba
 import numpy as np
+import ROOT
+
+# numba is not used directly, but tests can crash when ROOT is built with
+# builtin_llvm=OFF and numba is not imported at the beginning:
+numba.__version__  # this is to avoid unused import warnings
```
vepadulano, line 9: This should be enough to avoid the unused import warning
```suggestion
import numba # noqa: F401
import numpy as np
import ROOT

# numba is not used directly, but tests can crash when ROOT is built with
# builtin_llvm=OFF and numba is not imported at the beginning:
```

vepadulano (changes requested):
That's unfortunate :( but I agree it wouldn't be the first time we see something like this and changing the test is ok. I left a minor comment

vepadulano (approved):
Thanks! Let's still wait for the CI to finish before merging

Issue #17855, opened 2025-03-02
Add `linspace`, `regspace`, and `logspace` to `ROOT::VecOps`


edfink234:
#### **Description**  
Currently, `ROOT::VecOps` provides `Range`, which is useful for generating sequences of integers like Python’s `range()`. However, no direct equivalent exists for generating evenly spaced floating-point sequences like `linspace`, `regspace`, or `logspace`, which exist in NumPy and other numerical computing libraries like Armadillo.  

#### **Motivation**  
- Many scientific and data analysis workflows require generating sequences of floating-point numbers, especially for defining analysis binning, parameter sweeps, or interpolation grids.  
- `ROOT::VecOps::Range` only works with integers, which limits its applicability in cases where floating-point precision is needed. Unlike Python’s `range()`, which produces a lazy range object, `ROOT::VecOps::Range` produces an actual `RVec`, making it natural to extend this functionality to floating-point sequences.  
- Implementing `linspace`, `regspace`, and `logspace` would align `ROOT::VecOps` with common numerical computing conventions, improving usability and reducing the need for external workarounds.  

#### **Proposed Feature**  
- **`ROOT::VecOps::linspace(start, stop, num_points)`**: Generate `num_points` evenly spaced values between `start` and `stop` (inclusive).  
- **`ROOT::VecOps::regspace(start, stop, step)`**: Generate values from `start` to `stop` with step size `step`.  
- **`ROOT::VecOps::logspace(start_exp, stop_exp, num_points, base=10)`**: Generate `num_points` logarithmically spaced values between `base^start_exp` and `base^stop_exp`.  

This addition would significantly improve ROOT’s numerical capabilities and enhance user experience for scientific computing.

#### **Additional Context**  
Current workarounds involve using `std::vector<double>` and manually populating sequences, which is inefficient and less readable compared to built-in ROOT `RVec` operations. Another place where it would be useful would be for TGraph, i.e. with the range function one can simply do something like the following analogous to the numpy and mpl API

```c++
ROOT::VecOps::RVec<int> x = ROOT::VecOps::Range(1, 6);
ROOT::VecOps::RVec<int> y = ROOT::VecOps::Range(1, 6);
int n = x.size();
auto g = new TGraph(n, x.data(), y.data());
g->Draw("AC*");
```

but for floating point operations, there is no equivalent API for this convenient plotting mechanism in ROOT, hence why it could be useful for these functions to be added.

[...]

PR #18200, opened 2025-03-30
Adding linSpace, logSpace, and arange prototypes

edfink234:
# This Pull request:

## Changes or fixes:

Adds `linSpace`, `logSpace`, and `arange` functions to `ROOT::VecOps` namespace

## Checklist:

- [x] tested changes locally

This PR address [issue #17855](https://github.com/root-project/root/issues/17855)


dpiparo:
Ah, and also, the equality of double precision floating point is not working too well because of the very meaning of equality of floating point ;-)

edfink234:
Hi,

For this next commit, I’ve updated the functions as follows:

 - LinSpace, LogSpace, and Arange are now templated to deduce a common arithmetic type that defaults to double when necessary
 - I’ve adjusted parameter naming to match VecOps conventions by using a lowercase 'n' with a default power-of-two value (128). 
 - The Doxygen documentation has been expanded to fully detail the template parameters, function behavior, usage examples, and to clarify key implementation details such as the check against LLONG_MAX. 
 - I’ve updated the tests to use approximate (tolerance-based) comparisons for floating-point values with a `CheckNear` function

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,203 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * cout << LinSpace(-1, 5, 5) << "\n";
+ * // { -1, 0.5, 2, 3.5, 5 }
+ * cout << LinSpace(3, 12, 5) << "\n";
+ * // { 3, 5.25, 7.5, 9.75, 12 }
+ * cout << LinSpace(1.4, 13.66, 5) << "\n";
+ * // { 1.4, 4.465, 7.53, 10.595, 13.66 }
+ * ~~~
+ */
+template <typename T1 = double, typename T2 = double, typename T3 = double, typename Common_t = std::conditional_t<std::is_floating_point_v<std::common_type_t<T1, T2, T3>>, std::common_type_t<T1, T2, T3>, double>>
+inline RVec<Common_t> LinSpace(T1 start, T2 end, unsigned long long n = 128)
```
dpiparo, line 3287: This template is really nice. I would like to propose to use a simpler one, though: `template<typename T=double> RVec<T> LinSpace`. It is *great* to have code as generic as possible, but perhaps in this case 3 template parameters when realistically the cases will be float and double might seem a bit overcomplicated.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,203 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * cout << Arange(0, 0, 5) << '\n';
+ * // {  }
+ * cout << Arange(-7, 20, 4) << '\n';
+ * // { -7, -3, 1, 5, 9, 13, 17 }
+ * cout << Arange(1, 13, 5) << '\n';
+ * // { 1, 6, 11 }
+ * ~~~
+ */
+template <typename T1 = double, typename T2 = double, typename T3 = double, typename Common_t = std::conditional_t<std::is_floating_point_v<std::common_type_t<T1, T2, T3>>, std::common_type_t<T1, T2, T3>, double>>
+inline RVec<Common_t> Arange(T1 start, T2 end, T3 step)
```
silverweed, line 3421: What does the "A" in "Arange" stand for? I personally find this function name pretty obscure...

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,203 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * cout << LinSpace(1.4, 13.66, 5) << "\n";
+ * // { 1.4, 4.465, 7.53, 10.595, 13.66 }
+ * ~~~
+ */
+template <typename T1 = double, typename T2 = double, typename T3 = double, typename Common_t = std::conditional_t<std::is_floating_point_v<std::common_type_t<T1, T2, T3>>, std::common_type_t<T1, T2, T3>, double>>
+inline RVec<Common_t> LinSpace(T1 start, T2 end, unsigned long long n = 128)
+{
+    RVec<Common_t> temp;
+    
+    if (!n || (n > LLONG_MAX)) // Check for invalid or absurd n.
```
dpiparo, line 3291: Thanks for the clarification in the comment. Can we perhaps use std::numeric_limits instead of LLONG_MAX?

math/vecops/inc/ROOT/RVec.hxx
edfink234: Idk, it was the name for the function given in the edited title of the PR so I just used that...

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,203 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
    return ret;
 }
 
+/**
```
dpiparo, line 3245: Very good documentation.

dpiparo (changes requested):
Good job. Once the changes are addressed and the tests pass, we are ready to merge.

edfink234:
Thanks for the quick response @dpiparo!  I will get to these hopefully today (Tuesday PST) if not at some point this week. It's almost 2 am where I am atm 😅...

math/vecops/inc/ROOT/RVec.hxx
silverweed: @dpiparo do you know this?

dpiparo: There is something fishy going on. The original title of the issue was `Add linspace, regspace, and logspace to ROOT::VecOps`, should this be Logspace?

edfink234:
@dpiparo I made another commit to address the new comments and modified the tests a bit to hopefully have them pass.

math/vecops/inc/ROOT/RVec.hxx
ferdymercury: It's because of numpy I guess:
https://stackoverflow.com/questions/75487100/what-is-the-a-in-numpy-arange
https://stackoverflow.com/questions/55102806/why-was-the-name-arange-chosen-for-the-numpy-function

silverweed: Thanks @ferdymercury. I would spell it out entirely to be honest, it's not like it saves much typing anyway.

ferdymercury: Yep, quite obscure. ArrayRange is not superhelpful either, considering that you already know that the return value is an RVec. So what you gain from fully spelling it, you lose by diverging from what people are already aware / remember from numpy.
So my two cents would be: Either depart completely and do sth like GetLinearlySpacedValues instead of aRange that could be even merged with overloads with LinSpace, since they are very similar (https://stackoverflow.com/questions/62106028/what-is-the-difference-between-np-linspace-and-np-arange)
or alternatively just follow exactly numpy's obscurity for consistency.

dpiparo:
Thanks @edfink234 . Let's see how this goes.

ferdymercury (approved):

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,203 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * cout << Linspace(-1, 5, 5) << "\n";
+ * // { -1, 0.5, 2, 3.5, 5 }
+ * cout << Linspace(3, 12, 5) << "\n";
+ * // { 3, 5.25, 7.5, 9.75, 12 }
+ * cout << Linspace(1.4, 13.66, 5) << "\n";
+ * // { 1.4, 4.465, 7.53, 10.595, 13.66 }
+ * ~~~
+ */
+template <typename T1 = double, typename T2 = double, typename T3 = double, typename Common_t = std::conditional_t<std::is_floating_point_v<std::common_type_t<T1, T2, T3>>, std::common_type_t<T1, T2, T3>, double>>
+inline RVec<Common_t> Linspace(T1 start, T2 end, unsigned long long n = 128)
```
vepadulano, line 3287: The template parameter `T3` is not used anywhere in the template, so I guess it can be removed. Also, from what I understand the return type will always be an RVec of floating point type. I personally do not see the reason to not always just have `RVec<double>`.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,203 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
    return ret;
 }
 
+/**
+ * @brief Produce RVec with N evenly-spaced entries from start to end inclusive.
+ *
+ * This function generates a vector of evenly spaced values, starting at @p start and ending at @p end,
+ * with exactly @p n elements. The spacing is computed as
+ * \f$\text{step} = \frac{\text{end} - \text{start}}{n-1}\f$, so that the vector includes both endpoints.
```
vepadulano, line 3250: Throughout the ROOT codebase, the Doxygen token used as sequence beginner is `\` rather than `@`. See for example the docs of `Construct` from above: `\begin`, `\tparam`, `\p` etc. Please move all the `@` instances to `\` instead.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,203 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * cout << Linspace(1.4, 13.66, 5) << "\n";
+ * // { 1.4, 4.465, 7.53, 10.595, 13.66 }
+ * ~~~
+ */
+template <typename T1 = double, typename T2 = double, typename T3 = double, typename Common_t = std::conditional_t<std::is_floating_point_v<std::common_type_t<T1, T2, T3>>, std::common_type_t<T1, T2, T3>, double>>
+inline RVec<Common_t> Linspace(T1 start, T2 end, unsigned long long n = 128)
+{
+    RVec<Common_t> temp;
+    
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
```
vepadulano, line 3291: What is the reason to not use the same type as the input `n` parameter?
```suggestion
    if (n == 0 || (n > std::numeric_limits<unsigned long long>::max())) // Check for invalid or absurd n.
```

If I read the doc, this check is also there to "prevent reserving an absurd amount of memory". This definition is quite subjective imho. Probably already `n == 1e9` would be absurd for some definition of the word.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,203 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ */
+template <typename T1 = double, typename T2 = double, typename T3 = double, typename Common_t = std::conditional_t<std::is_floating_point_v<std::common_type_t<T1, T2, T3>>, std::common_type_t<T1, T2, T3>, double>>
+inline RVec<Common_t> Logspace(T1 start, T2 end, unsigned long long n = 128, T3 base = 10.0)
+{
+    RVec<Common_t> temp;
+    
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
+        return temp;
+    }
```
vepadulano, line 3359: Similar comments that I made for `LinSpace` also apply here

vepadulano (changes requested):
This is a quite useful contribution, thanks a lot! I have left some comments for discussion

math/vecops/inc/ROOT/RVec.hxx
edfink234: When I run the following:

```cpp
std::cout << std::boolalpha << (static_cast<unsigned long long>(-1) > std::numeric_limits<long long>::max()) << '\n';
```

I get `true`, but if I run the following:

```cpp
std::cout << std::boolalpha << (static_cast<unsigned long long>(-1) > std::numeric_limits<unsigned long long>::max()) << '\n';
``` 

I get `false`, so it seems to me like keeping it as `long long` could check and prevent catastrophe if the user entered a negative value by mistake, or 0 for step in `Arange`. For example, when I run the following problematic cases with how its currently implemented, I get a quick return of an empty `RVec` because of that check, which at least on my machine doesn't work if I change it to `std::numeric_limits<unsigned long long>::max()`

```cpp
Linspace(1, 10, -1000) //(ROOT::VecOps::RVec<double>) {}
Logspace(1, 4, -10) //(ROOT::VecOps::RVec<double>) {}
Arange(1, 10, 0) //(ROOT::VecOps::RVec<double>) {}
```

math/vecops/inc/ROOT/RVec.hxx
edfink234: I think the benefit could be if the user wants float or some other floating point type, they have some options on how they call these functions (`Linspace`, `Logspace`, `Arange`) to yield the desired floating point type. For example (after the changes from the most recent commit to this PR), to illustrate some of this flexibility below:

```cpp
root [1] Linspace(1, 10, 10)
(ROOT::VecOps::RVec<double>) { 1.0000000, 2.0000000, 3.0000000, 4.0000000, 5.0000000, 6.0000000, 7.0000000, 8.0000000, 9.0000000, 10.000000 }
root [2] Linspace(1.0f, 10.0f, 10)
(ROOT::VecOps::RVec<float>) { 1.00000f, 2.00000f, 3.00000f, 4.00000f, 5.00000f, 6.00000f, 7.00000f, 8.00000f, 9.00000f, 10.0000f }
root [3] Linspace<float, float>(1, 10, 10)
(ROOT::VecOps::RVec<float>) { 1.00000f, 2.00000f, 3.00000f, 4.00000f, 5.00000f, 6.00000f, 7.00000f, 8.00000f, 9.00000f, 10.0000f }
root [4] Linspace<long double, long double>(1, 10, 10)
(ROOT::VecOps::RVec<long double>) { 1.0000000L, 2.0000000L, 3.0000000L, 4.0000000L, 5.0000000L, 6.0000000L, 7.0000000L, 8.0000000L, 9.0000000L, 10.000000L }
root [5] Logspace(1.0f, 5.0f, 5, 10.0f)
(ROOT::VecOps::RVec<float>) { 10.0000f, 100.000f, 1000.00f, 10000.0f, 100000.f }
root [6] Logspace<float, float, float>(1, 5, 20)
(ROOT::VecOps::RVec<float>) { 10.0000f, 16.2378f, 26.3665f, 42.8133f, 69.5193f, 112.884f, 183.298f, 297.635f, 483.293f, 784.760f, 1274.28f, 2069.14f, 3359.82f, 5455.60f, 8858.67f, 14384.5f, 23357.2f, 37926.9f, 61584.8f, 100000.f }
root [7] Arange<float, float, float>(1, 10, 3)
(ROOT::VecOps::RVec<float>) { 1.00000f, 4.00000f, 7.00000f }
root [8] Arange<long double, long double, long double>(3, 15, 3.1)
(ROOT::VecOps::RVec<long double>) { 3.0000000L, 6.1000000L, 9.2000000L, 12.300000L }
```

ferdymercury:
Thanks a lot for this!

I have some suggestions, since I think it would be good to mimick python numpy more:
- Linspace/Logspace should accept function argument `const bool endpoint = false/true; // for lin/log`, see https://numpy.org/doc/2.1/reference/generated/numpy.linspace.html https://numpy.org/doc/2.1/reference/generated/numpy.logspace.html
- when `dtype=int`, we should template-specialize, ie not static_cast, but rather "round towards -inf", as explained in Numpy docu.
- Could Lin/Arange's default value of Common_t be automatically inferred based on the input args, as done in Numpy?. This would make it easier to do `for(unsigned int i : Linspace<unsigned int>(0, 10, 11))` instead of currently `for(unsigned int i : Linspace<unsigned int, unsigned int, unsigned int>(0, 10, 11))`. If this is not possible, I would add a doxygen `\note` saying that this is different from Numpy.
- Related to the previous point, in the Doxy docu, we could say that with C++23, you could just do a python-like enumerate for loop:
```
for (auto const [index, val] : std::views::enumerate(ROOT::VecOps::Linspace(6,10,16)))
```

ferdymercury:
> I tried my best to address these, though for the second and third points, I don't know how to cleanly add a specialization without incurring an ambiguous compiler error wrt being unable to deduce the correct overload.

Not sure, maybe it's easier to use just a ternary operator such as:

double result = std::is_floating_point ? static_cast : floor ;

ferdymercury (changes requested):

math/vecops/inc/ROOT/RVec.hxx
ferdymercury: I think we could get the desired functionality by simplifying this as follows:
```
template <typename T, typename Ret = T> Ret Linspace(T start, T end, unsigned long long n = 50, const bool endpoint=true) {
   const long double step = (end - start) / (n - endpoint);
   tmp.push_back(std::is_floating_point_v<Ret> ? start + i * step : std::round(start + i * step));
}
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,242 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+inline RVec<Common_t> Linspace(T1 start, T2 end, unsigned long long n = 128, const bool endpoint = true)
+{
+    RVec<Common_t> temp;
+    
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
+        return temp;
+    }
+    
+    Common_t step = (static_cast<Common_t>(end) - static_cast<Common_t>(start)) / static_cast<Common_t>(n - endpoint);
```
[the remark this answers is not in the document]
ferdymercury: see https://github.com/numpy/numpy/blob/734e2cd9467b08524e7f7c96ee6ad876d804d3ba/numpy/_core/function_base.py#L181

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,242 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * cout << Linspace(3, 12, 5) << "\n";
+ * // { 3, 5.25, 7.5, 9.75, 12 }
+ * cout << Linspace(3, 12, 5, false) << "\n";
+ * // { 3, 4.8, 6.6, 8.4, 10.2 }
+ * Linspace<int, int, int>(1, 10, 3) << "\n";
+ * // { 1, 5, 9 }
+ * ~~~
+ */
+template <typename T1 = double, typename T2 = double, typename Common_t = std::conditional_t<std::is_floating_point_v<std::common_type_t<T1, T2>>, std::common_type_t<T1, T2>, double>>
+inline RVec<Common_t> Linspace(T1 start, T2 end, unsigned long long n = 128, const bool endpoint = true)
```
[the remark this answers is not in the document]
ferdymercury: Ok, thanks, I missed that, makes sense. (Maybe specify in the docu that this default is different from traditional numpy)

math/vecops/test/vecops_rvec.cxx
```diff
@@ -885,6 +903,53 @@ TEST(VecOps, RangeBeginEndStride)
...
+   CheckNear(Linspace(1.4, 13.66, 5), RVecD{ 1.4, 4.465, 7.53, 10.595, 13.66 });
+   CheckNear(Linspace(3, 12, 5, false), RVecD{ 3, 4.8, 6.6, 8.4, 10.2 });
+   CheckNear(Linspace(3, 12, 1, false), RVecD{ 3 });
+   CheckNear(Linspace(3, 12, 1, true), RVecD{ 3 });
+   CheckEqual(Linspace<int, int, int>(1, 10, 3), RVecI{ 1, 5, 9 });
+}
+
+TEST(VecOps, Logspace)
+{
+   CheckNear(Logspace(4, 10, 12), RVecD{ 10000.0, 35111.917, 123284.67, 432876.13, 1519911.1, 5336699.2, 18738174., 65793322., 2.3101297e+08, 8.1113083e+08, 2.8480359e+09, 1e+10 });
```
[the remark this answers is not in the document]
edfink234: I'm unsure what you mean; I look at numpy logspace and the signature's [`numpy.logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None, axis=0)`](https://numpy.org/doc/stable/reference/generated/numpy.logspace.html#numpy.logspace)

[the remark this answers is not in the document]
edfink234: Ah, you meant the 3rd, 4th, and 5th parameters, respectively, my bad.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,242 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * // {  }
+ * cout << Arange(-7, 20, 4) << '\n';
+ * // { -7, -3, 1, 5, 9, 13, 17 }
+ * cout << Arange(1, 13, 5) << '\n';
+ * // { 1, 6, 11 }
+ * cout << Arange<unsigned int, unsigned int, unsigned int, unsigned int>(5, 9, 1) << '\n';
+ * // { 5, 6, 7, 8 }
+ * ~~~
+ */
+template <typename T1 = double, typename T2 = double, typename T3 = double, typename Common_t = std::conditional_t<std::is_floating_point_v<std::common_type_t<T1, T2, T3>>, std::common_type_t<T1, T2, T3>, double>>
```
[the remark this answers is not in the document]
edfink234: I think it's because to make it flexible and avoid problems like below:

```cpp
#include <iostream>
#include <typeinfo>
using namespace std;

template<typename T1, typename T2>
void func(T1 x, T1 y)
{
    cout << x << ' ' << y << '\n';
    cout << typeid(T1).name() << '\n';
}

int main() {
    func(1.3f, 1);
    return 0;
}
```

Which gives me an error:

```
./Playground/file0.cpp: In function 'int main()':
./Playground/file0.cpp:13:9: error: no matching function for call to 'func(float, int)'
   13 |     func(1.3f, 1);
      |     ~~~~^~~~~~~~~
./Playground/file0.cpp:6:6: note: candidate: 'template<class T1, class T2> void func(T1, T1)'
    6 | void func(T1 x, T1 y)
      |      ^~~~
./Playground/file0.cpp:6:6: note:   template argument deduction/substitution failed:
./Playground/file0.cpp:13:9: note:   deduced conflicting types for parameter 'T1' ('float' and 'int')
   13 |     func(1.3f, 1);
      |     ~~~~^~~~~~~~~
```

So I think something like the above could happen if the user does something like `Logspace(1.3f, 5)`, for example.

math/vecops/test/vecops_rvec.cxx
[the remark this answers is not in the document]
ferdymercury: Yep. Also, for lin/logspace, I would suggest adding argument long double *retstep = nullptr; and inside the function_ if(retstep) *retstep = step;
(numpy has sth like that for linspace)

math/vecops/inc/ROOT/RVec.hxx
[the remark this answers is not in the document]
ferdymercury: I think it's preferable to force the user to write 1.3f, 1.f ie to force them to write the start and end with the same precision / data type. But that's just my opinion hehe.

math/vecops/inc/ROOT/RVec.hxx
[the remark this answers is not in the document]
edfink234: Well, this I'm not so sure what to say. In the way I have it, if I implement your suggestion, i.e. (I think?)

```cpp
Common_t step = std::is_floating_point_v<Common_t> ? (end - start) / (n - endpoint) : static_cast<Common_t>(end - start) / (n - endpoint);
```

then e.g. for `Linspace(-1, 5, 5)` this gives `{ -1, 0, 1, 2, 3 }` instead of the expected `{-1, 0.5, 2, 3.5, 5}`. But again, I don't know if this is what you had in mind.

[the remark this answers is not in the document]
ferdymercury: I was proposing removing Common_t typename, and just using long double for step. Then you have a Ret typename which is the RVec contents. Ret tparam could be set to double by default.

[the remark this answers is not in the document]
ferdymercury: see https://github.com/root-project/root/pull/18200#discussion_r2037158383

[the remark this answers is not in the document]
edfink234: But at the end of the day, I'd like to get this merged asap, so if you think it will get merged iff I change it, I will do it lol!

tutorials/experimental/rcanvas/rcanvas_mt.cxx
```diff
@@ -92,7 +92,7 @@ void draw_canvas(const std::string &title, Int_t col1)
    canvas->Remove();
 }
 
-void rcanvas_mt(bool block_main_thread = true)
+void rcanvas_mt(bool block_main_thread = false)
```
ferdymercury, line 95: You are (accidentally) modifying three files not related to the PR. Something went wrong with the 'rebasing'.
Try updating your fork's master branch first to be like root's one, and then:
git rebase --interactive master
being sure to edit the conflicts by hand. (If you find issues, I can solve it for you and force-push if you prefer.)

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,243 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * cout << Linspace(3, 12, 5) << "\n";
+ * // { 3, 5.25, 7.5, 9.75, 12 }
+ * cout << Linspace(3, 12, 5, false) << "\n";
+ * // { 3, 4.8, 6.6, 8.4, 10.2 }
+ * Linspace<int, int, int>(1, 10, 3) << "\n";
+ * // { 1, 5, 9 }
+ * ~~~
+ */
+template <typename T1 = double, typename T2 = double, typename Common_t = std::conditional_t<std::is_floating_point_v<std::common_type_t<T1, T2>>, std::common_type_t<T1, T2>, double>>
+inline RVec<Common_t> Linspace(T1 start, T2 end, unsigned long long n = 128, const bool endpoint = true)
```
vepadulano, line 3302: My biggest concern for this PR are the complications added for the templates of all the three newly introduced functions. IMHO, I don't see why we should allow something like

```
auto arr = [Linspace/Logspace](1.f, 33,...);
```
I guess it should be more than enough to always have only one `T` parameter for all functions, and then change the return type to simply be
* RVec<T> if T is a floating point type
* RVec<double> otherwise
Thoughts @dpiparo ?

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,243 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * cout << Logspace(0, 0, 0) << '\n';
+ * // {  }
+ * cout << Logspace(4, 10, 12, 10.0, false) << '\n';
+ * // { 10000, 31622.8, 100000, 316228, 1e+06, 3.16228e+06, 1e+07, 3.16228e+07, 1e+08, 3.16228e+08, 1e+09, 3.16228e+09 }
+ * cout << Logspace<int, int, int, int>(1, 5, 3) << '\n';
+ * // { 10, 1000, 100000 }
+ * ~~~
+ */
+template <typename T1 = double, typename T2 = double, typename T3 = double, typename Common_t = std::conditional_t<std::is_floating_point_v<std::common_type_t<T1, T2, T3>>, std::common_type_t<T1, T2, T3>, double>>
+inline RVec<Common_t> Logspace(T1 start, T2 end, unsigned long long n = 128, const bool endpoint = true, T3 base = 10.0)
```
vepadulano, line 3380: I'm unsure whether the spelling is fine as is or we should have

* `LinSpace`
* `LogSpace`
* `ARange`

Thoughts @dpiparo ?

tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
```diff
@@ -305,6 +305,39 @@ public:
       return GetView<T>(RetrieveFieldId(fieldName), rawPtr);
    }
 
+   /// Provides access to an individual (sub)field from its on-disk ID.
```
vepadulano, line 308: There are many spurious extra changes that somehow ended up in this PR, make sure to remove them

vepadulano (changes requested):
This is a good idea, but I think we still have design decisions to make. I left some comments to invite further discussion

math/vecops/inc/ROOT/RVec.hxx
ferdymercury: In my opinion, we should also allow to return RVec<int> as numpy does if you specify `dtype=int`.
So two template parameters:
- T for start/end
- Ret = double (by default), but can be changed to int.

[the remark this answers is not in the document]
dpiparo: fair enough, it's passing by value after all...

math/vecops/inc/ROOT/RVec.hxx
dpiparo: I do not have strong opinions here. numpy has all lowercase names. In ROOT method names start with upper case letters.

tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
dpiparo: it's actually a full file...

math/vecops/inc/ROOT/RVec.hxx
ferdymercury: > For what concerns returning `RVec<int>`: what would be the use case? Can this be covered by casting a `RVec<double>`?

```
double myVector[100];
double myDecimatedVector[10];
for (auto const [idx, val] : std::views::enumerate(ROOT::VecOps::Linspace<auto, int>(0, 90, 10)))
   myDecimatedVector[idx] = myVector[val];
```

Yes, one could cast `val` to an int instead, true. But I just thought it's more convenient to follow the same conventions as numpy that allows you to specify dtype=int. And working with integers is safer than working with floating points wrt precision issues.

See silly reproducer below, the result is not the same with floats as with ints.

```
#include "ROOT/RVec.hxx"
#include <iostream>

template <typename T, typename Ret = double> ROOT::VecOps::RVec<Ret> Linspace(const T start, const T end, const unsigned long long n = 128, const bool endpoint = true, long double *const retstep = nullptr)
{
   ROOT::VecOps::RVec<Ret> temp;
   long double step = std::is_floating_point_v<T> ? (end - start) / (n - endpoint) : static_cast<long double>(end - start) / (n - endpoint);
   if (retstep)
      *retstep = step;
   temp.reserve(n);
   temp.push_back(static_cast<Ret>(start));
   for (unsigned long long i = 1; i < n; i++)
      temp.push_back(std::is_floating_point_v<T> ? static_cast<Ret>(start + i * step) : std::floor(start + i * step));
   return temp;
}

void linspace()
{
   double myVector[100000000];
   double myDecimatedVector[10];
   // with C++23: for (auto const [idx, val] : std::views::enumerate(Linspace<auto, int>(0, 90, 10)))
   auto idx = 0;
   for (auto val : Linspace<int, int>(100000000-100, 100000000-10, 10)) {
      myDecimatedVector[idx++] = myVector[val];
      std::cout << val << std::endl;
   }
   for (auto val : Linspace<int, float>(100000000-100, 100000000-10, 10)) {
      myDecimatedVector[idx++] = myVector[int(val)];
      std::cout << int(val) << std::endl;
   }
}

```

tutorials/experimental/rcanvas/rcanvas_mt.cxx
edfink234: I was trying to do this but was very unsure whether to keep head or the commit with the long hash so I think I chose incorrectly at some point. If you could solve it for me and let me know how I can do update my local branch I would greatly appreciate it

ferdymercury: Hi, it's fixed now. You can now overwrite your branch in your local PC following this link: https://stackoverflow.com/a/61490618/7471760 (be sure first to checkout the branch)

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,242 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * Linspace<int, int, int>(1, 10, 3) << "\n";
+ * // { 1, 5, 9 }
+ * ~~~
+ */
+template <typename T1 = double, typename T2 = double, typename Common_t = std::conditional_t<std::is_floating_point_v<std::common_type_t<T1, T2>>, std::common_type_t<T1, T2>, double>>
+inline RVec<Common_t> Linspace(T1 start, T2 end, unsigned long long n = 128, const bool endpoint = true)
+{
+    RVec<Common_t> temp;
+    
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
```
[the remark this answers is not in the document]
edfink234: @ferdymercury When I add this and do 

```cpp
Linspace(3, 12, 1, true)
```

It gives me an empty vector even though numpy gives me `{3}`, see numpy below:

```python
>>> np.linspace(3, 12, 1, endpoint=True)
array([3.])
```

So should I change it to what I had before?

```cpp
if (!n || (n > std::numeric_limits<long long>::max()))
```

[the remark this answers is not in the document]
ferdymercury: Oh, right. I think it's cleaner if you write if yourCondition return {} elseif endpoint and n==1 return Rvec{start};

math/vecops/inc/ROOT/RVec.hxx
edfink234: @dpiparo @vepadulano @ferdymercury  So, concretely, what changes should I make now? Personally, I have no issue merging the different template types into one.

edfink234:
@dpiparo @vepadulano @ferdymercury I made two commits. The first one was changing `temp.reserve(n); temp.push_back(val)` in `Linspace/Logspace/Arange` `temp(n); temp[i] = val;` as I ran numerous tests to confirm that the latter is more efficient and easier for the compiler to optimise. The second commit is a separate commit that merges `T1`, `T2`, `T3` into one template type `T` in case you prefer this version (which I gather may be the case); of course for this separate commit the test cases have been modified accordingly.

ferdymercury:
Thanks a lot!! From my side: this would be the only change left to do:

```
long double step = std::is_floating_point_v<CommonT> ? (end - start) / (n - endpoint) : static_cast<long double>(end - start) / (n - endpoint);
temp[i]=(std::is_floating_point_v<CommonT> ? static_cast<Ret>(start + i * step) : std::floor(start + i * step));
```

Since step must be floating even if CommonT is int. And for high N, we avoid extrapolation errors by always choosing doubles (or long doubles) even if CommonT is float.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,235 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * cout << Linspace(3, 12, 5) << "\n";
+ * // { 3, 5.25, 7.5, 9.75, 12 }
+ * cout << Linspace(3, 12, 5, false) << "\n";
+ * // { 3, 4.8, 6.6, 8.4, 10.2 }
+ * Linspace<int, int>(1, 10, 3) << "\n";
+ * // { 1, 5, 9 }
+ * ~~~
+ */
+template <typename T = double, typename Common_t = std::conditional_t<std::is_floating_point_v<T>, T, double>>
+inline RVec<Common_t> Linspace(T start, T end, unsigned long long n = 128, const bool endpoint = true)
```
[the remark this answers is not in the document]
ferdymercury: Thanks Vincenzo for the feedback!

> Now suddenly the common type is not the common type at all!

From my point of view:
- Common_t should be renamed to Ret_t (return type) as I wrote in my suggested snippet. (and fixing step https://github.com/root-project/root/pull/18200#issuecomment-2817750795)
- The doxygen comment is wrong and should be fixed accordingly

> I still do not understand why the template needs to be anything more complicated than `template <typename T>`

In my opinion, it's essential that the user can separately specify the auto-deduced input type and the (if-not-default) output type.

Example with your suggestion:

`<typename Input> RVec<Input> Linspace(Input start, Input end, ...)`

I write `Linspace(3,5, 64)`

So it's clear that I want double as returned values. But auto-deduction in this case will cast the return value to int's, hence returning nonsense. Of course, you could say it's the user fault, it should have specified `Linspace(3.,5., 64)`, but that's not super-user-friendly. And it would go against the default behavior of np.linspace, so translatability of scripts to/from Python would be risked.

Now you would argue, ok, let's always force that the return values are doubles even if inputs arguments are passed as ints.

`<typename Input> RVec<double> Linspace(Input start, Input end, ...)`

But that strategy would invalidate this other valid use case:

`Linspace(1,10,10)` where here I really want integers returned, not doubles, so that I can do for-range based loops using integers.

So to me, the best compromise is really:
`<typename Input, typename Output = double> RVec<Output> Linspace(Input start, Input end, ...)`

edfink234:
Hi @dpiparo @vepadulano @ferdymercury

I added the changes suggested by @vepadulano and @ferdymercury. I'm not sure how I feel about the suggestion of @ferdymercury for step if I understand the need for all these casts and what the floor does exactly, but ok, I changed it and checked that it yields the expected output.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * cout << Arange(1, 13, 5) << '\n';
+ * // { 1, 6, 11 }
+ * cout << Arange<unsigned int, unsigned int>(5, 9, 1) << '\n';
+ * // { 5, 6, 7, 8 }
+ * ~~~
+ */
+template <typename T = double, typename Ret_t = std::conditional_t<std::is_floating_point_v<T>, T, double>>
+inline RVec<Ret_t> Arange(T start, T end, T step)
+{
+    unsigned long long n = std::ceil(static_cast<Ret_t>(end-start)/static_cast<Ret_t>(step)); // Ensure floating-point division.
```
ferdymercury, line 3483:
```suggestion
    unsigned long long n = std::ceil(static_cast<Ret_t>(end-start)/static_cast<long double>(step)); // Ensure floating-point division.
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+    {
+        return RVec<Ret_t>{};
+    }
+    
+    long double step = std::is_floating_point_v<Ret_t> ?
+    (static_cast<Ret_t>(end) - static_cast<Ret_t>(start)) / static_cast<Ret_t>(n - endpoint) :
+    static_cast<long double>(end - start) / (n - endpoint);
+        
+    RVec<Ret_t> temp(n);
+    temp[0] = static_cast<Ret_t>(start);
```
ferdymercury, line 3313:
```suggestion
    temp[0] = std::is_floating_point_v<Ret_t> ? static_cast<Ret_t>(start) : std::floor(start);
```
Needed for consistency, since static_cast<int> rounds towards zero whereas std::floor rounds towards negative.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+    
+    Ret_t start_c = static_cast<Ret_t>(start);
+    Ret_t end_c   = static_cast<Ret_t>(end);
+    Ret_t base_c  = static_cast<Ret_t>(base);
+    
+    long double step = std::is_floating_point_v<Ret_t> ?
+    (static_cast<Ret_t>(end_c - start_c) / static_cast<Ret_t>(n - endpoint)) :
+    static_cast<long double>(end - start) / (n - endpoint);
+
+    temp[0] = static_cast<Ret_t>(std::pow(base_c, start_c));
```
ferdymercury, line 3403: Same as above. Or start loop at i=0

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+    unsigned long long n = std::ceil(static_cast<Ret_t>(end-start)/static_cast<Ret_t>(step)); // Ensure floating-point division.
+
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
+        return RVec<Ret_t>{};
+    }
+    
+    RVec<Ret_t> temp(n);
+    
+    Ret_t start_c = static_cast<Ret_t>(start);
```
ferdymercury, line 3492: Same as above, or move inside loop.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+inline RVec<Ret_t> Linspace(T start, T end, unsigned long long n = 128, const bool endpoint = true)
+{
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
+        return RVec<Ret_t>{};
+    }
+    
+    long double step = std::is_floating_point_v<Ret_t> ?
+    (static_cast<Ret_t>(end) - static_cast<Ret_t>(start)) / static_cast<Ret_t>(n - endpoint) :
+    static_cast<long double>(end - start) / (n - endpoint);
```
ferdymercury, line 3310:
```suggestion
    end >= start ? static_cast<long double>(end - start) / (n - endpoint) : (static_cast<long double>(end)- start)) / (n - endpoint);
```

This is need to avoid an underflow if user calls Linspace(5UL, 1UL);

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+        
+    RVec<Ret_t> temp(n);
+    temp[0] = static_cast<Ret_t>(start);
+    if (std::is_floating_point_v<Ret_t>)
+    {
+        for (unsigned long long i = 1; i < n; i++)
+        {
+            temp[i] = static_cast<Ret_t>(start) + static_cast<Ret_t>(i) * step;
```
ferdymercury, line 3318: maybe it can be simplified as follows?

```suggestion
            temp[i] = static_cast<Ret_t>(start + i * step);
```
same in all below.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+    }
+    RVec<Ret_t> temp(n);
+    
+    Ret_t start_c = static_cast<Ret_t>(start);
+    Ret_t end_c   = static_cast<Ret_t>(end);
+    Ret_t base_c  = static_cast<Ret_t>(base);
+    
+    long double step = std::is_floating_point_v<Ret_t> ?
+    (static_cast<Ret_t>(end_c - start_c) / static_cast<Ret_t>(n - endpoint)) :
+    static_cast<long double>(end - start) / (n - endpoint);
```
ferdymercury, line 3401:
```suggestion
    end >= start ? static_cast<long double>(end - start) / (n - endpoint) : (static_cast<long double>(end)- start)) / (n - endpoint);
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+        return RVec<Ret_t>{};
+    }
+    
+    RVec<Ret_t> temp(n);
+    
+    Ret_t start_c = static_cast<Ret_t>(start);
+    
+    long double step_c = std::is_floating_point_v<Ret_t> ?
+    static_cast<Ret_t>(step) :
+    static_cast<long double>(step);
```
ferdymercury, line 3496:
```suggestion
```
I would suggest casting only the final result, not the intermediate variables.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+{
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
+        return RVec<Ret_t>{};
+    }
+    RVec<Ret_t> temp(n);
+    
+    Ret_t start_c = static_cast<Ret_t>(start);
+    Ret_t end_c   = static_cast<Ret_t>(end);
+    Ret_t base_c  = static_cast<Ret_t>(base);
```
ferdymercury, line 3397: Maybe just cast to ret when you store things in temp vector, not the intermediate variables.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+template <typename T = double, typename Ret_t = std::conditional_t<std::is_floating_point_v<T>, T, double>>
+inline RVec<Ret_t> Linspace(T start, T end, unsigned long long n = 128, const bool endpoint = true)
+{
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
+        return RVec<Ret_t>{};
+    }
+    
+    long double step = std::is_floating_point_v<Ret_t> ?
+    (static_cast<Ret_t>(end) - static_cast<Ret_t>(start)) / static_cast<Ret_t>(n - endpoint) :
```
ferdymercury, line 3309:
```suggestion
    (end - start) / static_cast<long double>(n - endpoint) :
```
no need for Ret cast of intermediate variables?

edfink234:
Hi @ferdymercury 

I just pushed the new commit with these changes!

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+{
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
+        return RVec<Ret_t>{};
+    }
+    RVec<Ret_t> temp(n);
+    
+    Ret_t start_c = static_cast<Ret_t>(start);
+    Ret_t end_c   = static_cast<Ret_t>(end);
+    Ret_t base_c  = static_cast<Ret_t>(base);
```
ferdymercury, line 3397:
```suggestion
    long double start_c = start;
    long double end_c   = end;
    long double base_c  = base;
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+    
+    long double step = std::is_floating_point_v<Ret_t> ?
+    (end_c - start_c) / static_cast<long double>(n - endpoint) :
+    (end >= start ? static_cast<long double>(end - start) / (n - endpoint) : (static_cast<long double>(end) - start) / (n - endpoint));
```
ferdymercury, line 3401:
```suggestion
    long double step = (end_c - start_c) / (n - endpoint);
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+    temp[0] = std::is_floating_point_v<Ret_t> ?
+    static_cast<Ret_t>(std::pow(base_c, start_c)) :
+    std::floor(std::pow(base_c, start_c));
+     
+    if (std::is_floating_point_v<Ret_t>)
+    {
+        for (unsigned long long i = 1; i < n; i++)
+        {
+            Ret_t exponent = start_c + i * step;
+            temp[i] = std::pow(base_c, exponent);
```
ferdymercury, line 3412:
```suggestion
            auto exponent = start_c + i * step;
            temp[i] = static_cast<Ret_t>(std::pow(base_c, exponent));
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * ~~~
+ */
+template <typename T = double, typename Ret_t = std::conditional_t<std::is_floating_point_v<T>, T, double>>
+inline RVec<Ret_t> Arange(T start, T end, T step)
+{
+    unsigned long long n = std::ceil(static_cast<Ret_t>(end-start)/static_cast<long double>(step)); // Ensure floating-point division.
+
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
+        return RVec<Ret_t>{};
```
ferdymercury, line 3489:
```suggestion
        return {};
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * cout << Logspace<int, int>(1, 5, 3) << '\n';
+ * // { 10, 1000, 100000 }
+ * ~~~
+ */
+template <typename T = double, typename Ret_t = std::conditional_t<std::is_floating_point_v<T>, T, double>>
+inline RVec<Ret_t> Logspace(T start, T end, unsigned long long n = 128, const bool endpoint = true, T base = 10.0)
+{
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
+        return RVec<Ret_t>{};
```
ferdymercury, line 3391:
```suggestion
        return {};
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * Linspace<int, int>(1, 10, 3) << "\n";
+ * // { 1, 5, 9 }
+ * ~~~
+ */
+template <typename T = double, typename Ret_t = std::conditional_t<std::is_floating_point_v<T>, T, double>>
+inline RVec<Ret_t> Linspace(T start, T end, unsigned long long n = 128, const bool endpoint = true)
+{
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
+        return RVec<Ret_t>{};
```
ferdymercury, line 3305:
```suggestion
        return {};
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+        {
+            Ret_t exponent = start_c + i * step;
+            temp[i] = std::pow(base_c, exponent);
+        }
+    }
+    else
+    {
+        for (unsigned long long i = 1; i < n; i++)
+        {
+            Ret_t exponent = start_c + i * step;
```
ferdymercury, line 3419:
```suggestion
            auto exponent = start_c + i * step;
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+    unsigned long long n = std::ceil(static_cast<Ret_t>(end-start)/static_cast<long double>(step)); // Ensure floating-point division.
+
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
+        return RVec<Ret_t>{};
+    }
+    
+    RVec<Ret_t> temp(n);
+    
+    Ret_t start_c = std::is_floating_point_v<Ret_t> ? static_cast<Ret_t>(start) : std::floor(start);
```
ferdymercury, line 3494:
```suggestion
    long double start_c = start;
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+        
+    long double step_c = step;
+
+    temp[0] = start_c;
```
ferdymercury, line 3498:
```suggestion
    temp[0] = std::is_floating_point_v<Ret_t> ? static_cast<Ret_t>(start) : std::floor(start);
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,277 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+    if (std::is_floating_point_v<Ret_t>)
+    {
+        for (unsigned long long i = 1; i < n; i++)
+        {
+            temp[i] = start_c + i * step_c;
```
ferdymercury, line 3503:
```suggestion
            temp[i] = static_cast<Ret_t>(start_c + i * step_c);
```

ferdymercury (changes requested):

edfink234:
Hi @ferdymercury @dpiparo @vepadulano! I made a commit with the most recent changes suggested by @ferdymercury.

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,274 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * \brief Produce RVec with n log-spaced entries from base^{start} to base^{end}.
+ *
+ * This function generates a vector of values where the exponents are evenly spaced, and then returns the
+ * corresponding values of base raised to these exponents. If \p endpoint is true (default), the vector
+ * contains \p n values with the last element equal to \f$base^{end}\f$. If \p endpoint is false, the
+ * sequence is computed as if there were n+1 evenly spaced samples over the interval in the exponent space,
+ * and the final value (\f$base^{end}\f$) is excluded, resulting in a sequence of n values.
+ *
+ * The function is templated to allow for different arithmetic types. The return type \c Ret_t, if not explicitly specified,
+ * iis determined as follows: if \p T is a floating point type, that type is used; otherwise, the arithmetic is performed using \c double.
```
ferdymercury, line 3341:
```suggestion
 * The function is templated to allow for different return types. The return type \c Ret_t, if not explicitly specified,
 * is determined as follows: if \p T is a floating point type, that type is used; otherwise, the return type is \c double.
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,274 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ *
+ * \tparam T Type of the start and end exponents and the base. Default is double.
+ * \tparam Ret_t Deduced type used for arithmetic, which, if not explicitly specified, is \p T if that is a floating point type, or double otherwise.
```
ferdymercury, line 3344:
```suggestion
 * \tparam Ret_t Deduced type used for return type, which, if not explicitly specified, is \p T if that is a floating point type, or double otherwise.
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,274 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * the sequence consists of n values computed as if there were n+1 evenly spaced samples, with the final
+ * value (\p end) omitted; in this case, \f$\text{step} = \frac{\text{end} - \text{start}}{n}\f$.
+ *
+ * The function is templated to allow for different arithmetic types. The return type \c Ret_t, if
+ * not explicitly specified, is determined as follows: if \p T is a floating point type, that type is used;
+ * otherwise, the arithmetic is performed using \c double.
+ *
+ * \tparam T Type of the start and end value. Default is double.
+ * \tparam Ret_t Return type used for arithmetic, which, if not explicitly specified
+ * in the template, is \p T if that is a floating point type, or double otherwise.
```
ferdymercury, line 3261:
```suggestion
 * The function is templated to allow for different return types. The return type \c Ret_t, if
 * not explicitly specified, is determined as follows: if \p T is a floating point type, that type is used;
 * otherwise, the return type is \c double.
 *
 * \tparam T Type of the start and end value. Default is double.
 * \tparam Ret_t Return type used, which, if not explicitly specified
 * in the template, is \p T if that is a floating point type, or double otherwise.
```

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,274 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * \f]
+ * ensuring that the arithmetic is performed in a floating-point context when needed.
+ *
+ * The function is templated to allow for different arithmetic types. The deduced type \c Ret_t, if not
+ * explicitly specified, is determined as follows: if \p T is a floating point type, that type is used;
+ * otherwise, the arithmetic is performed using \c double.
+ *
+ * \tparam T Type of the start, end, and step values. Default is double.
+ * \tparam Ret_t Deduced type used for arithmetic, which, if not explicitly
+ * specified, is \p T if that is a floating point type, or double otherwise.
```
ferdymercury, line 3442:
```suggestion
 * The function is templated to allow for different return types. The return type \c Ret_t, if not
 * explicitly specified, is determined as follows: if \p T is a floating point type, that type is used;
 * otherwise, the return type is \c double.
 *
 * \tparam T Type of the start, end, and step values. Default is double.
 * \tparam Ret_t Return type, which, if not explicitly
 * specified, is \p T if that is a floating point type, or double otherwise.
```

math/vecops/test/vecops_rvec.cxx
```diff
@@ -885,6 +903,53 @@ TEST(VecOps, RangeBeginEndStride)
...
+   CheckNear(Linspace(0, 0, 5), RVecD{ 0, 0, 0, 0, 0 });
+   CheckNear(Linspace(-7, 20, 4), RVecD{ -7, 2, 11, 20 });
+   CheckNear(Linspace(1, 13, 5), RVecD{ 1, 4, 7, 10, 13 });
+   CheckNear(Linspace(1, 4, 0), RVecD{});
+   CheckNear(Linspace(1, 14, 0), RVecD{});
+   CheckNear(Linspace(1.4, 13.66, 5), RVecD{ 1.4, 4.465, 7.53, 10.595, 13.66 });
+   CheckNear(Linspace(3, 12, 5, false), RVecD{ 3, 4.8, 6.6, 8.4, 10.2 });
+   CheckNear(Linspace(3, 12, 1, false), RVecD{ 3 });
+   CheckNear(Linspace(3, 12, 1, true), RVecD{ 3 });
+   CheckEqual(Linspace<int, int>(1, 10, 3), RVecI{ 1, 5, 9 });
```
ferdymercury, line 921:
```suggestion
   CheckEqual(Linspace<int, int>(1, 10, 3), RVecI{ 1, 5, 10 });
```

ferdymercury (changes requested):
Thanks a lot for the changes!

I have now some corrections for the documentation,
and one issue with the int-Linspace, I am getting locally different results than yours, please cross-check.

math/vecops/inc/ROOT/RVec.hxx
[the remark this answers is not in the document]
vepadulano: Thank you @ferdymercury , I definitely see your point and I agree it is a valid use case. I like it with the new naming that makes it clear it is a user choice to have a different return type than the type of the inputs.

vepadulano (changes requested):
Some more minor comments, I feel like we're getting very close to the conclusion, thanks for bearing through!

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,274 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * ~~~{.cpp}
+ * using namespace ROOT::VecOps;
+ * cout << Linspace(-1, 5, 5) << "\n";
+ * // { -1, 0.5, 2, 3.5, 5 }
+ * cout << Linspace(3, 12, 5) << "\n";
+ * // { 3, 5.25, 7.5, 9.75, 12 }
+ * cout << Linspace(3, 12, 5, false) << "\n";
+ * // { 3, 4.8, 6.6, 8.4, 10.2 }
+ * Linspace<int, int>(1, 10, 3) << "\n";
+ * // { 1, 5, 9 }
```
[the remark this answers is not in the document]
edfink234: No, you are right, that is indeed a typo. Thanks for the catch!

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,274 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+    long double end_c   = end;
+    long double base_c  = base;
+    
+    long double step = (end_c - start_c) / (n - endpoint);
+    
+    temp[0] = std::is_floating_point_v<Ret_t> ?
+    static_cast<Ret_t>(std::pow(base_c, start_c)) :
+    std::floor(std::pow(base_c, start_c));
+     
+    if (std::is_floating_point_v<Ret_t>)
```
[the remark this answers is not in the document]
edfink234: Does this turn it into a compile-time instruction (I've never seen it before 😅)

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,274 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * \note If the template parameter \c Ret_t is explicitly overridden with an integral type, the arithmetic may cause rounding issues. Consequently, the resulting sequence may not end exactly at \p end. To ensure that the sequence ends exactly at \p end, consider casting the result as follows: `RVec<integral_type>(Linspace(...))` (which is equivalent in numpy to `np.linspace(...).astype(integral_type)`). This behavior is different than NumPy in Python.
+ *
+ * \par C++23 Enumerate Support:
+ * With C++23, you can use the range-based enumerate view to iterate over the resulting vector with both the index
+ * and the value, similar to Python's `enumerate`. For example:
+ * ~~~{.cpp}
+ * for (auto const [index, val] : std::views::enumerate(ROOT::VecOps::Linspace(6, 10, 16))) {
+ *     // Process index and val.
+ * }
+ * ~~~
```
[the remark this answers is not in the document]
edfink234: I think @ferdymercury recommended this... @ferdymercury?

edfink234:
@ferdymercury @vepadulano @dpiparo I made the updates!

I also just wanted to share an updated benchmark using the actual `ROOT RVec` type instead of std::vector. Running the attached cpp file with the compilation directive `g++ -std=c++20 -o linspaceTestAlternateRVec linspaceTestAlternateRVec.cpp -O2 -fno-inline-functions $(root-config --libs --cflags)` 

RVec
```
Linspace benchmark Time elapsed = 0.108864
Logspace benchmark time elapsed = 0.853611
Arange benchmark time elapsed = 0.0218328
```

This is better than I get with `std::vector`:

```
Linspace benchmark Time elapsed = 0.541573
Logspace benchmark time elapsed = 1.34327
Arange benchmark time elapsed = 0.0794047
```

vs Python numpy (`python linspaceTest.py`):

```
Linspace time elapsed = 2.097444 seconds
Logspace time elapsed = 4.336540 seconds
Arange time elapsed = 0.135273 seconds
```

[LinspaceLogspaceArangeNumpyVsRVec.zip](https://github.com/user-attachments/files/20275603/LinspaceLogspaceArangeNumpyVsRVec.zip)

math/vecops/inc/ROOT/RVec.hxx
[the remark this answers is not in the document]
ferdymercury: I thought to put it here since we are dealing with Python equivalent functions in C++, but yes, it's not really related, I agree. My vote is still to keep this information, since many people are unaware of it. But I am fine with removing it, too.

edfink234:
@vepadulano @dpiparo Any updates? I wasn't sure from the discussion about `std::views` if it was wanted to delete this documentation about it...

edfink234:
@vepadulano @dpiparo Any updates? I wasn't sure from the discussion about std::views if it was wanted to delete this documentation about it...

edfink234:
@vepadulano @dpiparo Any updates? I wasn't sure from the discussion about std::views if it was wanted to delete this documentation about it.

edfink234:
@vepadulano @dpiparo Any updates? I wasn't sure from the discussion about std::views if it was wanted to delete this documentation about it.

edfink234:
@vepadulano @dpiparo Any updates? I wasn't sure from the discussion about std::views if it was wanted to delete this documentation about it.

edfink234:
@vepadulano @dpiparo Any updates? I wasn't sure from the discussion about std::views if it was wanted to delete this documentation about it.

edfink234:
@vepadulano @dpiparo Any updates? I wasn't sure from the discussion about std::views if it was wanted to delete this documentation about it.

edfink234:
@vepadulano @dpiparo Any updates? I wasn't sure from the discussion about std::views if it was wanted to delete this documentation about it.

edfink234:
@vepadulano @dpiparo Any updates? I wasn't sure from the discussion about std::views if it was wanted to delete this documentation about it.

edfink234:
Is this PR still being considered for merging in the near future? It seems to me that this PR has gone a bit stale...

math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -3242,6 +3242,274 @@ RVec<typename RVec<T>::size_type> Enumerate(const RVec<T> &v)
...
+ * // { 3, 4.8, 6.6, 8.4, 10.2 }
+ * cout << Linspace<int, int>(1, 10, 3) << "\n";
+ * // { 1, 5, 10 }
+ * ~~~
+ */
+template <typename T = double, typename Ret_t = std::conditional_t<std::is_floating_point_v<T>, T, double>>
+inline RVec<Ret_t> Linspace(T start, T end, unsigned long long n = 128, const bool endpoint = true)
+{
+    if (!n || (n > std::numeric_limits<long long>::max())) // Check for invalid or absurd n.
+    {
```
vepadulano, line 3301: Since this is an "absurd" case, shouldn't we throw an exception?

math/vecops/test/vecops_rvec.cxx
```diff
@@ -37,6 +37,24 @@ void CheckEqual(const ROOT::RVecD &a, const ROOT::RVecD &b, std::string_view msg
    }
 }
 
+// Checks if all the elements in `v1` are at most `percentage` percent different
+// from `v2`
+void CheckNear(const RVecD &v1, const RVecD &v2, double percentage = 0.01)
```
vepadulano: Ideally I think the `percentage` should be higher precision, at least 10^-4.

math/vecops/inc/ROOT/RVec.hxx
edfink234: I'm not sure it is needed since this absurd case I would reckon is most likely to be a negative-integer passed in for size, so I feel like returning an empty vector is sufficiently benign?

edfink234:
Ok, I believe I've addressed this comment, thanks @vepadulano

dpiparo:
thanks again for this great functionality!

PR #18204, opened 2025-03-31
[ntuple] Use `std::string_view` instead of `const std::string &` in parameter types

enirolf:
This PR unifies the type signature for string-based parameters in the RNTuple interface, using `std::string_view` instead of `const std::string &`.

The only place where the type signature hasn't been changed is in `RNTupleModel::FieldMappingFunc_t`, because that would introduce a breaking interface change so we should discuss whether we want to change it here still.

### DIFF
--- a/tree/ntuple/v7/src/RFieldUtils.cxx
+++ b/tree/ntuple/v7/src/RFieldUtils.cxx
@@ -62,7 +62,7 @@ const std::unordered_map<std::string_view, std::string_view> typeTranslationMap{
 
 // Recursively normalizes a template argument using the regular type name normalizer F as a helper.
 template <typename F>
-std::string GetNormalizedTemplateArg(const std::string &arg, F fnTypeNormalizer)
+std::string GetNormalizedTemplateArg(std::string_view arg, F fnTypeNormalizer)
 {
    R__ASSERT(!arg.empty());
 
@@ -81,7 +81,7 @@ std::string GetNormalizedTemplateArg(const std::string &arg, F fnTypeNormalizer)
 }
 
 using AnglePos = std::pair<std::string::size_type, std::string::size_type>;
-std::vector<AnglePos> FindTemplateAngleBrackets(const std::string &typeName)
+std::vector<AnglePos> FindTemplateAngleBrackets(std::string_view typeName)
 {
    std::vector<AnglePos> result;
    std::string::size_type currentPos = 0;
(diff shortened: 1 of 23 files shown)


tree/ntuple/v7/src/RFieldUtils.cxx
```diff
@@ -122,9 +122,9 @@ std::vector<AnglePos> FindTemplateAngleBrackets(const std::string &typeName)
 
 } // namespace
 
-std::string ROOT::Internal::GetCanonicalTypePrefix(const std::string &typeName)
+std::string ROOT::Internal::GetCanonicalTypePrefix(std::string_view typeName)
 {
-   std::string canonicalType{TClassEdit::CleanType(typeName.c_str(), /*mode=*/1)};
+   std::string canonicalType{TClassEdit::CleanType(typeName.data(), /*mode=*/1)};
```
hahnjo, line 127: `string_view::data()` is not necessarily NULL-terminated. If the function accepts a `std::string_view`, it must create a `std::string` copy in order to then get a correct `c_str()`.

tree/ntuple/v7/src/RFieldUtils.cxx
```diff
@@ -264,11 +264,11 @@ std::string ROOT::Internal::GetRenormalizedTypeName(const std::string &metaNorma
    return normName;
 }
 
-std::string ROOT::Internal::GetNormalizedUnresolvedTypeName(const std::string &origName)
+std::string ROOT::Internal::GetNormalizedUnresolvedTypeName(std::string_view origName)
 {
    const TClassEdit::EModType modType = static_cast<TClassEdit::EModType>(
       TClassEdit::kDropStlDefault | TClassEdit::kDropComparator | TClassEdit::kDropHash);
-   TClassEdit::TSplitType splitname(origName.c_str(), modType);
+   TClassEdit::TSplitType splitname(origName.data(), modType);
```
hahnjo, line 271: same

tree/ntuple/v7/src/RFieldUtils.cxx
```diff
@@ -344,54 +344,54 @@ std::string ROOT::Internal::GetNormalizedInteger(unsigned long long val)
...
       return GetNormalizedInteger(ParseIntTypeToken(intTemplateArg));
    return GetNormalizedInteger(ParseUIntTypeToken(intTemplateArg));
 }
 
-long long ROOT::Internal::ParseIntTypeToken(const std::string &intToken)
+long long ROOT::Internal::ParseIntTypeToken(std::string_view intToken)
 {
    std::size_t nChars = 0;
-   long long res = std::stoll(intToken, &nChars);
+   long long res = std::stoll(intToken.data(), &nChars);
```
hahnjo, line 358: same

hahnjo (changes requested):
There's a problem that `std::string_view::data()` isn't guaranteed to return a NULL-terminated buffer. I marked the relevant places in the first commit, but I think this requires some more thought. Maybe it is actually more efficient to accept `const std::string &` in some places?

pcanal:
In the general case, it make sense to other an `std::string_view` interface if and only if all the functions called inside are already expecting `std::string_view` (also it does not make sense to take an `std::string_view` if the intent is to copy the content into an `std::string).   The general wisdom is that offering `std::string_view` interface is to be done/implemented from the inside out (innermost function first).  `TClassEdit` could be upgraded to accept `std::string_view` instead a `const char*`.

Issue #12935, opened 2023-06-02
[RF] Global correlation coefficients after SumW2Error


goi42:
### Is your feature request related to a problem? Please describe.

As of 6.28/00, `RooFitResult::globalCorr` causes a seg fault if `SumW2Error` was used in the fit. First reported in the forum [here](https://root-forum.cern.ch/t/extracting-global-correlation-causes-a-seg-fault-with-sumw2error/).

### Describe the solution you'd like

`RooFitResult::globalCorr` should return the global correlation coefficient using the corrected error matrix. As stated in the [forum](https://root-forum.cern.ch/t/extracting-global-correlation-causes-a-seg-fault-with-sumw2error/55138/3), the global correlation coefficients are defined as
$`\rho_k^2 = 1 - [V_{kk} * (V^{-1})_{kk}]^{-1}`$,
so the corrected correlation coefficients could be defined as
$`(\rho_{corr})_k^2 = 1 - [(V_{corr})_{kk} * (V_{corr}^{-1})_{kk}]^{-1}`$.

### Describe alternatives you've considered

At the very least, calling `RooFitResult::globalCorr` should raise a sensible error instead of causing a seg fault.

### Additional context

More generally, it would be nice if RooFit printed the corrected matrix instead of (or after) the uncorrected one at the end of the fit. At present, one sees a long list of parameter updates printed, but the errors and correlations in the final one printed differ from those stored in the `RooFitResult`.

dpiparo:
I unfortunately spoke too soon. The issue is indeed still there.

PR #18207, opened 2025-03-31
[RF] fix segfault in fillLegacyCorrMatrix

ferdymercury:
# This Pull request:

## Changes or fixes:

segfault shown in https://root-forum.cern.ch/t/extracting-global-correlation-causes-a-seg-fault-with-sumw2error/55138

This does not fully solve https://github.com/root-project/root/issues/12935, it's just adding safety.

## Checklist:

- [x] tested changes locally


guitargeek (approved):
Thanks! That's already a step in the right direction.

Issue #17729, opened 2025-02-15
[Python Interface] Regression: can't properly convert a vector to 2D/3D numpy array


dpiparo:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Issue opened fom a [Forum post](https://root-forum.cern.ch/t/wrong-nd-vector-conversion-to-numpy-array-in-pyroot/62947)

### Reproducer

```python
import ROOT
import numpy as np

a = ROOT.vector("vector<vector<float>>")() # 3D
a1 = ROOT.vector("vector<float>")() # 2D
a2 = ROOT.vector("float")()
a2.push_back(1.)
a2.push_back(2.)
a2.push_back(3.)
a1.push_back(a2)
a1.push_back(a2)
a1.push_back(a2)
a.push_back(a1)
a.push_back(a1)
a.push_back(a1)

print(a)

print(np.array(a).shape) # 3D
print(np.array(a1).shape) # 2D
```

The output should be, as given by 6.30.06:
```
{ { { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f } }, { { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f } }, { { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f } } }
(3, 3, 3)
(3, 3)
```

However, for 6.34.04 it is:
```
{ { { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f } }, { { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f } }, { { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f }, { 1.00000f, 2.00000f, 3.00000f } } }
(3,)
(3,)
```

### ROOT version

6.34.04

### Installation method

all

### Operating system

all

lwpiotr:
Hi. Is anything happening regarding this regression? The inability to convert vectors to numpy arrays probably makes most PyROOT code that reads vectors unusable with recent ROOT versions.

[...]

PR #18208, opened 2025-03-31
[CPyCppyy] Drop `__array__` from std::vector pythonizations

aaronj0:
The addition of the `__array__` utility to std::vector Python proxies causes a bug where the resulting array is a single dimension, causing loss of data when converting to numpy arrays, for >1dim vectors. The recursive nature of this
function `VectorArray`, passes each subarray (pydata) to the next call and only the final buffer is cast to a lowlevel view and resized (in VectorData), resulting in only the first 1D array to be returned.

Since this C++ pythonization was added with the upgrade in 6.32, and is only defined and used recursively, the safe option is to disable this function and no longer add it. It is temporarily removed to prevent errors due to `-Wunused-function`

Fixes: https://github.com/root-project/root/issues/17729

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/Pythonize.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/Pythonize.cxx
@@ -527,6 +527,13 @@ PyObject* VectorData(PyObject* self, PyObject*)
 }
 
 
+// This function implements __array__, added to std::vector python proxies and causes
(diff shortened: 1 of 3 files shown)


guitargeek (approved):
Great, thank you very much for the investigation and finding just the right fix!

I'll also backport it to the 6.34 branch, so if there is a patch release of 6.34 before the upcoming 6.36, our users will also benefit from this fix.

PR #18212, opened 2025-03-31
[ci] Disable TMVA on Alma8

dpiparo:


.github/workflows/root-ci-config/buildconfig/alma8.txt
```diff
@@ -3,5 +3,10 @@ builtin_nlohmannjson=ON
 builtin_tbb=ON
 builtin_vdt=On
 fortran=OFF
```
lmoneta, line 5: No need to disable all tmva
Just do
`tmva-pymva=OFF`

PR #18215, opened 2025-04-01
[TMVA] Exclude RBatchGenerator tests if TMVA is disabled

martinfoell:
# This Pull request:

## Changes or fixes:
Check if TMVA is enabled for the RBatchGenerator tests

### DIFF
--- a/bindings/pyroot/pythonizations/test/CMakeLists.txt
+++ b/bindings/pyroot/pythonizations/test/CMakeLists.txt
@@ -189,8 +189,6 @@ ROOT_ADD_PYUNITTEST(pyroot_tcomplex tcomplex_operators.py)
 ROOT_ADD_PYUNITTEST(pyroot_memory memory.py)
 
 # rbatchgenerator tests
-# TODO: We currently do not support TensorFlow for Python >= 3.12 (see requirements.txt)
-# Update here once that is fixed.
-if (NOT MSVC AND Python3_VERSION VERSION_LESS 3.12 AND tmva)
+if (NOT MSVC AND tmva)
   ROOT_ADD_PYUNITTEST(batchgen rbatchgenerator_completeness.py PYTHON_DEPS numpy tensorflow torch)
 endif()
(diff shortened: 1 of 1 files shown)


dpiparo:
Thanks. Let's watch closely Alma8 builds.

lmoneta (commented):
The check should be independent of TMVA, but only on Python version or eventually a check on importing tensor flow. 
 See my changes in https://github.com/root-project/root/pull/18217

lmoneta (approved):
Sorry, I misunderstood. There is already a check on importing tensor flow with the dependency in the ROOT_ADD_PYUNITTEST. 
So it is correct to have the check also on tmva since the pythonization is built only when tmva is ON.
Thanks for the fix!

bindings/pyroot/pythonizations/test/CMakeLists.txt
```diff
@@ -191,6 +191,6 @@ ROOT_ADD_PYUNITTEST(pyroot_memory memory.py)
 # rbatchgenerator tests
 # TODO: We currently do not support TensorFlow for Python >= 3.12 (see requirements.txt)
 # Update here once that is fixed.
-if (NOT MSVC AND Python3_VERSION VERSION_LESS 3.12)
+if (NOT MSVC AND Python3_VERSION VERSION_LESS 3.12 AND tmva)
```
guitargeek, line 194:
```suggestion
if (NOT MSVC AND tmva)
```
As discussed already by @lmoneta: the tensorflow dependency is already encoded in `PYTHON_DEPS`, and if TensorFlow is not installed as per our `requirements.txt`, then the test will be disabled anyway. So any Python version check is redundant.

I would suggest to not check any Python version and also remove the TODO above. In fact, as it is now, we will probably forget to re-enable the test once we support TensorFlow for Python >= 3.12... If this is only encoded in the `requirements.txt`, things are easier.

martinfoell: Thanks for the suggestion @guitargeek, I have removed the Python version check now.

dpiparo:
Good job @martinfoell . This PR though seems to have superseded the functionality provided by this emergency [PR](https://github.com/root-project/root/pull/18217). Do you think this one can be closed?

Issue #18214, opened 2025-04-01
Project based on (Py)ROOT fails with loading `ROOT::v5::TFormula::GetExpFormula(char const*) const`


VanyaBelyaev:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

I have a project baed on ROOT/PyROOT.  
I am using ROOT from LCG dev3 nighly slots on cvmfs.
And this week my project fails with: 

```
cling::DynamicLibraryManager::loadLibrary(): /afs/cern.ch/user/i/ibelyaev/cmtuser/RELEASE/ostap/build/INSTALL/LCGdev3/x86_64-el9-clang19-opt/lib/libostap.so: undefined symbol: _ZNK4ROOT2v58TFormula13GetExpFormulaEPKc
```
The missing symbol correswponds to 
```
$ c++filt _ZNK4ROOT2v58TFormula13GetExpFormulaEPKc
ROOT::v5::TFormula::GetExpFormula(char const*) const
```
I have no changes in my porject for more then a aweek. 
The last successfun run was the last Thursday. Monday thjoisweek and today I see this problem. 
 

### Reproducer

I do not know if it is possible  to make a simple reproducer - my project is rather large.

### ROOT version

ROOT from cvmfs LCG dev3 nightly builds. 
The problme appears somewhen between last Thursday (the last successful tet o fmy projects) and this Monday (where I saw the problem)

### Installation method

cvmfs LCG dev3 nightly builds 

### Operating system

Linux

### Additional context

The problme appears somewhen between last Thursday (the last successful tet o fmy projects) and this Monday (where I saw the problem)

guitargeek:
Probably related to:

  * https://github.com/root-project/root/pull/18194

@ferdymercury, I think you need to revert the `v5` part of your PR, as I mentioned in the comment to that PR.

VanyaBelyaev:
Dear @guitargeek @ferdymercury 
Thank you very much for the prompt fix!

dpiparo:
My bad, not seeing issues in our CI, I thought it was time to merge. Thanks everybody!

pcanal:
Note that this error seems to indicates that:
```
 /afs/cern.ch/user/i/ibelyaev/cmtuser/RELEASE/ostap/build/INSTALL/LCGdev3/x86_64-el9-clang19-opt/lib/libostap.so
```
[...]

PR #18194, opened 2025-03-29
[hist] allow specifying the floating point precision in GetExpFormula

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/10769

## Checklist:

- [ ] tested changes locally
- [x] updated the docs (if necessary)

PR #18216, opened 2025-04-01
[hist] revert changes to v5 GetExpFormula fl_format

ferdymercury:
Revert part of recent changes (https://github.com/root-project/root/pull/18194) since it breaks backward compatibility as the function was virtual.

Fixes https://github.com/root-project/root/issues/18214

thanks guitargeek for finding out

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes #


guitargeek (approved):
Thank you very much for the quick fixup!

PR #18217, opened 2025-04-01
 Fix for a failure in importing tensorflow 

lmoneta:
- Re-enable tmva on alma8
- fix for ewrror in importing tensorflow in TMVA tutorial
- fix batchgenerator test for lower python versions


bindings/pyroot/pythonizations/test/CMakeLists.txt
```diff
@@ -191,6 +191,6 @@ ROOT_ADD_PYUNITTEST(pyroot_memory memory.py)
 # rbatchgenerator tests
 # TODO: We currently do not support TensorFlow for Python >= 3.12 (see requirements.txt)
 # Update here once that is fixed.
-if (NOT MSVC AND Python3_VERSION VERSION_LESS 3.12)
+if (NOT MSVC AND tmva AND Python3_VERSION VERSION_LESS 3.12 AND Python3_VERSION VERSION_GREATER 3.9)
```
guitargeek, line 194: The information becomes a bit scattered here, because we're also adding such constraints to the requirements.txt:
https://github.com/root-project/root/blob/master/requirements.txt#L14

What about completely dropping the Python version constraints here and only using the `requirements.txt` for that, as I also suggested here? https://github.com/root-project/root/pull/18215#discussion_r2022422703

sanjibansg (approved):

dpiparo:
Thanks @lmoneta !

PR #18218, opened 2025-04-01
[Python] Correctly manage NumPy dependencey in pythonization tests

guitargeek:
The dependency on Python libraries like NumPy needs to be declared in the CMakeLists.txt, such that the test gets disabled when the library is not available.

In some cases, the dependency on NumPy could be avoided by using the builtin "array" library.


vepadulano (approved):
LGTM, test failures seem unrelated

PR #18220, opened 2025-04-01
[rfile] Remove current RFile prototype

silverweed:
At the ROOT Retreat it was decided that the current RFile prototype will be replaced by a new prototype after 6.36.
The new prototype will go in a different direction by removing the shared ownership between RFile and the application and it won't have RFile inherit from RDirectory.


hageboeck (approved):
LGTM!

hahnjo:
Do we also want to remove `RDirectory`?

silverweed:
@linev I noticed that `RGeomViewer` is using RDirectory in `SaveAsMacro`: can this be removed or replaced with something else? (Also, why is this non-experimental class relying on an experimental one?)

linev:
>  I noticed that RGeomViewer is using RDirectory in SaveAsMacro: can this be removed or replaced with something else? (Also, why is this non-experimental class relying on an experimental one?)

It only used when geometry viewer saved as macro.
I will provide PR.

But we definitely need some replacement for `RDirectory::Heap()` functionality - if we aim to use `shared_ptr` and `unique_ptr` in our interfaces.

silverweed:
> But we definitely need some replacement for RDirectory::Heap() functionality - if we aim to use shared_ptr and unique_ptr in our interfaces.

Currently we don't plan to provide a globally-accessible directory and I would very much like to avoid it if we can. Why do you say it's necessary? I also don't see how it is related to the use of smart pointers

linev:
To fix usage of `RDirectory::Heap()` I create https://github.com/root-project/root/pull/18228

linev:
> Currently we don't plan to provide a globally-accessible directory and I would very much like to avoid it if we can.

`shared_ptr` or `unique_ptr` impose ownership. 
If we create objects like `RGeomViewer` or `RBrowser` as `shared_ptr` in macro -  object just destroyed when leaving function scope. But we want to preserve these objects until ROOT macro is running. `RDirectory::Heap()` was a possible solution for such problem. And we need to have some generic solution in the future  and do not create every time new global lists - like `RCanvas::GetCanvases()`.

silverweed:
@linev thanks for your quick PR. Let's rediscuss the `Heap` feature in the future, when the new RFile prototype is in a more mature state.

PR #18222, opened 2025-04-01
[RFC][CMake] Add alias for new root-merge tool

hahnjo:
Originally `hadd` was used to add histograms, which explains the name. Today it can do much more, for example concatenate TTrees or RNTuples while potentially recompressing the data. This would be better described by a new name: root-merge, or `rm` for short.

### DIFF
--- a/main/CMakeLists.txt
+++ b/main/CMakeLists.txt
@@ -38,6 +38,11 @@ else()
   if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
     target_link_libraries(hadd stdc++fs)
   endif()
+  # New name: root-merge, `rm` for short
+  # TODO: Implement clean command line interface
+  add_custom_command(TARGET hadd POST_BUILD
+                     COMMAND ln -f hadd rm
+                     WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
 endif()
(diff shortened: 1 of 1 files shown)


silverweed:
Are you really suggesting to call a tool `rm`? :eyes:

guitargeek (approved):
LGTM! PR failures seem unrelated

main/CMakeLists.txt
```diff
@@ -38,6 +38,11 @@ else()
   if (CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
     target_link_libraries(hadd stdc++fs)
   endif()
+  # New name: root-merge, `rm` for short
+  # TODO: Implement clean command line interface
+  add_custom_command(TARGET hadd POST_BUILD
+                     COMMAND ln -f hadd rm
```
ferdymercury, line 44: doesn't rm remind someone of remove command in linux ?

enirolf (commented):
Probably not for this PR, but we could maybe also start thinking of adding an `-r` flag for differentiating between TTrees and RNTuples?

guitargeek:
Good point @enirolf, but what about friend trees?

hahnjo:
Thanks everyone for entertaining the idea! I was made aware that it clashes with an insignificant command line tool, so we cannot go this route for the moment. I will come back with a new proposal in the next 365 days.

Issue #13880, opened 2023-10-17
Stats box content changes after zoom/unzoom with mouse wheel


linev:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

If display histogram, zoom in with mouse wheel an then zoom out, content of stats box changes.
Most probably due to the fact that `TAxis::kAxisRange` flag is not reset after unzooming.

### Reproducer

Load and display histogram with the command:
```
root --web=off -l https://root.cern/js/files/hsimple.root -e 'hpxpy->Draw()' 
```
And then zoom in/out with mouse wheel. 

Before:
![before](https://github.com/root-project/root/assets/4936580/82dba545-b9ed-4a2d-b59a-aff1a72a5ece)

After:
![after](https://github.com/root-project/root/assets/4936580/6659d04d-65c2-4e0f-b872-564ce3b34b01)

### ROOT version

master

### Installation method

build from source

### Operating system

OpenSUSE Linux

PR #18226, opened 2025-04-01
[hist] reset kAxisRange bit when fully unzooming

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/13880

## Checklist:

- [x] tested changes locally

### DIFF
--- a/hist/hist/src/TAxis.cxx
+++ b/hist/hist/src/TAxis.cxx
@@ -1039,8 +1039,9 @@ void TAxis::ChangeLabelByValue(Double_t labValue, Double_t labAngle, Double_t la
 ///  If `last > fNbins`, the overflow bin is included in the range and will be displayed during drawing.
 ///  If `first < 1`, the underflow bin is included in the range and will be displayed during drawing.
 ///
-///  In the special case that `first == 1` and `last == fNbins`, the kAxisRange bit is also reset,
-///  and the functions TAxis::GetFirst() and TAxis::GetLast() will return 1 and fNbins.
+///  In the special case that `first == 1` and `last == fNbins` and `kAxisRange` bit is set,
+///  and `fFirst` and `fLast` were already those values, then the `kAxisRange` bit is also reset. This
+///  happens when calling this function twice with the same arguments eg when unzooming with the mouse.
 ///
 ///  \note For historical reasons, SetRange(0,0) resets the range even though bin 0 is
 ///       technically reserved for the underflow; in order to set the range of the axis
@@ -1054,13 +1055,15 @@ void TAxis::SetRange(Int_t first, Int_t last)
 
    // special reset range cases
    if (last < first || (first < 0 && last < 0) ||
-         (first > nCells && last > nCells) || (first == 0 && last == 0) ||
-         (first == 1 && last == fNbins)
+         (first > nCells && last > nCells) || (first == 0 && last == 0)
    ) {
       fFirst = 1;
       fLast = fNbins;
       SetBit(kAxisRange, false);
-   } else {
+   } else if (first == 1 && last == fNbins && fFirst == 1 && fLast == fNbins && TestBit(kAxisRange)) {
+      SetBit(kAxisRange, false);
+   }
+   else {
       fFirst = std::max(first, 0);
       fLast = std::min(last, nCells);
       SetBit(kAxisRange, true);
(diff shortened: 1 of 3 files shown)


linev:
Can one do it in two steps?
1. `first == 1` and `last == fNbins` and `SetBit(kAxisRange, true)`
2. `first == 1` and `last == fNbins` and `SetBit(kAxisRange, false)`

ferdymercury:
>     2. `first == 1` and `last == fNbins` and `SetBit(kAxisRange, false)`

You mean TestBit?

linev:
> You mean TestBit?

I mean at first step range set to default, but `kAxisRange` bit remains set
And only by second attempt to set default range bit will be reset.

ferdymercury:
> I mean at first step range set to default, but `kAxisRange` bit remains set
> And only by second attempt to set default range bit will be reset.

Implemented, thks.

linev:
I tried your PR. It is still does not work as intended.
Just try:  
```
root -l --web=off tutorials/hsimple.root -e 'hpx->Draw("")'       
```
If one do zoom-in and then zooming-out with mouse wheel, one see always toggling two states.
Seems to be, while zooming-out `kAxisRange` bit toggling between on/off, changing output in stats box.

ferdymercury:
> If one do zoom-in and then zooming-out with mouse wheel, one see always toggling two states. Seems to be, while zooming-out `kAxisRange` bit toggling between on/off, changing output in stats box.

Yep, after the latest change, it stopped working. I tried now to fix it.

linev:
Yes, now zoom-out with mouse wheel works reasonable.
I let @couet approve PR.

couet (approved):

ferdymercury:
> I let @couet approve PR.

However, this change breaks the roottest/runProjections test

linev:
> this change breaks the roottest/runProjections test

One should check if `runProjections` relies on old behavior of Zoom?

linev:
As I said - one should check why runProjections fails.
Is it because output in stats box is different?

ferdymercury:
> As I said - one should check why runProjections fails. Is it because output in stats box is different?

the script calls SetRange so the resulting 1D or 2D projections have changed as a consequence (overflow included or not).

This sibling PR fixes it: https://github.com/root-project/roottest/pull/1295

linev (changes requested):
We are changing code in the wrong place.
And failure shown by roottest is real - our change may affect existing code.

Problem was not `TAxis::SetRange` itself but interactive mouse wheel zooming.
I propose to move check of `kAxisRange` flag in mouse wheel handler.

hist/hist/src/TAxis.cxx
```diff
@@ -1039,6 +1038,13 @@ void TAxis::ChangeLabelByValue(Double_t labValue, Double_t labAngle, Double_t la
 ///  If `last > fNbins`, the overflow bin is included in the range and will be displayed during drawing.
 ///  If `first < 1`, the underflow bin is included in the range and will be displayed during drawing.
 ///
+///  In the special case that `first == 1` and `last == fNbins`,
+///  and `fFirst` and `fLast` were already those values, then the `kAxisRange` bit is also reset. This
+///  happens when calling this function twice with the same arguments eg when (fully) unzooming with the mouse wheel.
+///  From a script, to fully unset the axis range, either call function twice with (1, fNbins) or call ResetBit(kAxisRange) by hand;
+///  to set the range to exactly 1 to fNbins without overflows, call SetRange(2,fNbins-1) and then SetRange(1,fNbins)
+///  or call SetRange(1,fNbins) and SetBit(kAxisRange) by hand.
+///
```
ferdymercury, line 1047:
```suggestion
```

hist/hist/src/TAxis.cxx
```diff
@@ -1056,15 +1062,15 @@ void TAxis::SetRange(Int_t first, Int_t last)
       fFirst = 1;
       fLast = fNbins;
       SetBit(kAxisRange, false);
+   } else if (first == 1 && last == fNbins && fFirst == 1 && fLast == fNbins) {
+      SetBit(kAxisRange, false);
```
ferdymercury, line 1066:
```suggestion
```

linev (approved):

ferdymercury:
Thanks for the hint Sergey! It's working fine now in my local build, hopefully tests are ok.

(Please "squash on merge")

linev (approved):
Yes, it works for me as well.
Tested with TH1 and TH2 zooming.
Thanks, @ferdymercury!

Issue #14155, opened 2023-11-30
roottest.root.hist.h2root Fails always on march=native builds and sporadically on all other platforms


pcanal:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

As seen in https://lcgapp-services.cern.ch/root-jenkins/job/root-pullrequests-build/190460/testReport/projectroot.roottest.root/hist/roottest_root_hist_h2root, `h2root` fails from time to time on Linux and always if the *march=native* option is active.

### Reproducer

```
-- TEST COMMAND -- 
cd /home/sftnight/build/workspace/root-pullrequests-build/build/roottest/root/hist
/usr/bin/timeout -s USR2 270s h2root mb4i1.hbook
-- BEGIN TEST OUTPUT --

 *** Break *** segmentation violation



===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================
#0  0x00007f55ea9f53ea in __GI___wait4 (pid=1686062, stat_loc=stat_loc
entry=0x7ffffe689758, options=options
entry=0, usage=usage
entry=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
#1  0x00007f55ea9f53ab in __GI___waitpid (pid=<optimized out>, stat_loc=stat_loc
entry=0x7ffffe689758, options=options
entry=0) at ./posix/waitpid.c:38
#2  0x00007f55ea95bbdb in do_system (line=<optimized out>) at ../sysdeps/posix/system.c:171
#3  0x00007f55eb242bc4 in TUnixSystem::StackTrace() () from /home/sftnight/build/workspace/root-pullrequests-build/build/lib/libCore.so.6.31
#4  0x00007f55eb23fef5 in TUnixSystem::DispatchSignals(ESignals) () from /home/sftnight/build/workspace/root-pullrequests-build/build/lib/libCore.so.6.31
#5  <signal handler called>
#6  0x0000556fff8ce440 in mzlink_ ()
#7  0x0000556fff8be829 in hlimit_ ()
#8  0x0000556fff8ad586 in main ()
#9  0x00007f55ea934d90 in __libc_start_call_main (main=main
entry=0x556fff8ad4d0 <main>, argc=argc
entry=2, argv=argv
entry=0x7ffffe68cf78) at ../sysdeps/nptl/libc_start_call_main.h:58
[...]

PR #15915, opened 2024-06-24
Hopefully, fix the long-standing problem with h2root

couet:
Using the C option (meaning C I.O) in hropen seems to fix the issue on Mac.

Issue #14682, opened 2024-02-13
MIsbehaviour of ROOT with ASAN enabled: some tests are failing


dpiparo:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

ROOT misbehaves when ASAN is enabled. Below different incarnation of the problem are illustrated:
- [ ] root/io/bigevent, root/io/event, root/io/compression, root/io/datamodelevolution, root/io/perf, root/treeformula/event, root/treeformula/retobj, 
root/treeformula/schemaEvolution, root/treeformula/sync, root/tree/friend, root/tree/evolution, root/tree/treeproblem, root/treeproxy, root/io/compression : bad linkage to asan functions
- [ ] root/io/references:  badly setup dependency on libEvent
- [ ] root/tree/addresses : `AddressSanitizer: heap-buffer-overflow`
- [ ] root/io/newClassDef: Real failure with ASAN enabled
- [ ] root/io/cint: bad linkage to ASAN libraries. Should also check if the tests are needed at all

Moreover, the above (and root/tree/addresses) are automatic Makefile driven tests. They could be translated to CMake to gain granularity.

### Reproducer

Run with Asan

### ROOT version

master

### Installation method

From sources (need asan)

### Operating system

Linux, mac

ferdymercury:
Cross-post from https://github.com/root-project/root/issues/19329

Adding -fsanitize undefined / address c++ flags to main/CMakeLists.txt (h2root) target, we get:
```

 /github/home/ROOT-CI/src/misc/minicern/src/cernlib.c:82:18: runtime error: left shift of negative value -2012982016
  /github/home/ROOT-CI/src/misc/minicern/src/cernlib.c:83:18: runtime error: left shift of negative value -2012982016
  /github/home/ROOT-CI/src/misc/minicern/src/cernlib.c:100:12: runtime error: left shift of 1212370737 by 8 places cannot be represented in type 'int'
  /github/home/ROOT-CI/src/misc/minicern/src/cernlib.c:100:39: runtime error: left shift of 1212370737 by 24 places cannot be represented in type 'int'
  /github/home/ROOT-CI/src/misc/minicern/src/zebra.f:6746:72: runtime error: load of address 0x5560461bae2c with insufficient space for an object of type 'integer(kind=4)'
  0x5560461bae2c: note: pointer points here
[...]

Issue #11657, opened 2022-11-03
ROOT does not compile with link-time optimizations (LTO) enabled


amadio:
Full bug report is at https://bugs.gentoo.org/879323.

Important part of the build log is below.
```

/usr/bin/cmake -E copy_if_different /var/tmp/portage/sci-physics/root-6.26.08/work/root-6.26.08/tutorials/tree/jets.C /var/tmp/portage/sci-physics/root-6.26.08/work/root-6.26.08_build/tutorials/tree/jets.C
/var/tmp/portage/sci-physics/root-6.26.08/work/root-6.26.08/misc/minicern/src/zebra.f:767:72: error: type of <E2><80>-Werror=lto-type-mismatch][]]
  767 |      + ISTAT)
      |                                                                        ^
/var/tmp/portage/sci-physics/root-6.26.08/work/root-6.26.08/misc/minicern/src/cernlib.c:107:6: note: ‘cfget_’ was previously declared here
  107 | void cfget_(int *lundes, int *medium, int *nwrec, int *nwtak, char *mbuf,
      |      ^
/var/tmp/portage/sci-physics/root-6.26.08/work/root-6.26.08/misc/minicern/src/cernlib.c:107:6: note: code may be misoptimized unless ‘-fno-strict-aliasing’ is used
/var/tmp/portage/sci-physics/root-6.26.08/work/root-6.26.08/misc/minicern/src/kernlib.f:398:72: error: type of <E2><80>-Werror=lto-type-mismatch][]]
  398 |       CFSTAT = CFSTATI (NAME, INFO, LGN)
      |                                                                        ^
/var/tmp/portage/sci-physics/root-6.26.08/work/root-6.26.08/misc/minicern/src/cernlib.c:171:5: note: type mismatch in parameter 4
  171 | int cfstati_(char *fname, int *info, int *lgname)
      |     ^
/var/tmp/portage/sci-physics/root-6.26.08/work/root-6.26.08/misc/minicern/src/cernlib.c:171:5: note: type ‘void’ should match type ‘long int’
/var/tmp/portage/sci-physics/root-6.26.08/work/root-6.26.08/misc/minicern/src/cernlib.c:171:5: note: ‘cfstati_’ was previously declared here
/var/tmp/portage/sci-physics/root-6.26.08/work/root-6.26.08/misc/minicern/src/kernlib.f:386:72: error: type of <E2><80>-Werror=lto-type-mismatch][]]
[...]

PR #18234, opened 2025-04-02
[h2root] explore/circumvent if occassional memory corruption issue stems from substring replacement

ferdymercury:
# This Pull request:

## Changes or fixes:

By adding a space, we create an extra char that, on Fortran's side will be substituted by c, rather than appending to the (temporary passed as argument C-)string.

Related discussion: https://github.com/root-project/root/pull/15915/files

Fixes https://github.com/root-project/root/issues/14155

### DIFF
--- a/main/src/h2root.cxx
+++ b/main/src/h2root.cxx
@@ -329,6 +329,8 @@ int main(int argc, char **argv)
 
    if (!hfile) {
       printf("Error: can't open output file: %s \n",file_out);
+      if (argc <= 2)
+         delete[] file_out;
       return 1;
    }
 
@@ -339,7 +341,9 @@ int main(int argc, char **argv)
    hfile->ls();
    hfile->Close();
    delete hfile;
-   return(0);
+   if (argc <= 2)
+         delete[] file_out;
+   return 0;
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/main/src/h2root.cxx
+++ b/main/src/h2root.cxx
@@ -312,7 +312,11 @@ int main(int argc, char **argv)
 
    int lun = 10;
 #ifndef WIN32
-   hropen(lun,PASSCHAR("example"),PASSCHAR(file_in),PASSCHAR("px"),record_size,ier,7,strlen(file_in),2);
+   // "px" is a string being changed to "pxc" in the fortran side; since it's a temporary on C's side,
+   // we need a buffer of at least 3 chars on Cs side, too. Predefine it as a extra variable "px " since that way
+   // the space will be replaced by 'c' on the preallocated memory, instead of appending a new char (new memory) to "px".
+   auto opt = PASSCHAR("px ");
+   hropen(lun,PASSCHAR("example"),PASSCHAR(file_in),opt,record_size,ier,7,strlen(file_in),2);
 #else
    hropen(lun,PASSCHAR("example"),PASSCHAR(file_in),PASSCHAR("px"),record_size,ier);
 #endif
--- a/misc/minicern/CMakeLists.txt
+++ b/misc/minicern/CMakeLists.txt
@@ -16,3 +16,5 @@ target_link_libraries(minicern ${CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES})
 # Disable warnings, since what has worked for 40 years...
 # (see https://sft.its.cern.ch/jira/browse/ROOT-9179 for the warnings)
 set_target_properties(minicern PROPERTIES COMPILE_FLAGS "-O0 -w")
+# set_target_properties(minicern PROPERTIES COMPILE_FLAGS "-fsanitize=undefined -fsanitize=address")
+# target_link_options(minicern BEFORE PUBLIC -fsanitize=undefined PUBLIC -fsanitize=address)
--- a/.github/workflows/root-ci-config/buildconfig/alma8.txt
+++ b/.github/workflows/root-ci-config/buildconfig/alma8.txt
@@ -3,5 +3,4 @@ builtin_nlohmannjson=ON
 builtin_tbb=ON
 builtin_vdt=ON
 ccache=ON
-fortran=OFF
 tmva-sofie=ON
--- a/.github/workflows/root-ci-config/buildconfig/ubuntu2404-cuda.txt
+++ b/.github/workflows/root-ci-config/buildconfig/ubuntu2404-cuda.txt
@@ -7,7 +7,6 @@ tmva-cpu=ON
 tmva-gpu=ON
 tmva-cudnn=ON
 tmva-pymva=OFF
-fortran=OFF
 gdml=OFF
 spectrum=OFF
 sqlite=OFF
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -387,7 +387,7 @@ jobs:
           - image: alma9
             is_special: true
             property: march_native
-            overrides: ["CMAKE_BUILD_TYPE=RelWithDebInfo", "CMAKE_CXX_FLAGS=-march=native", "CMAKE_C_FLAGS=-march=native", "fortran=OFF", "builtin_zlib=ON", "builtin_zstd=ON"]
+            overrides: ["CMAKE_BUILD_TYPE=RelWithDebInfo", "CMAKE_CXX_FLAGS=-march=native", "CMAKE_C_FLAGS=-march=native", "builtin_zlib=ON", "builtin_zstd=ON"]
           - image: alma9
             is_special: true
             property: arm64


main/src/h2root.cxx
```diff
@@ -312,7 +312,7 @@ int main(int argc, char **argv)
 
    int lun = 10;
 #ifndef WIN32
-   hropen(lun,PASSCHAR("example"),PASSCHAR(file_in),PASSCHAR("px"),record_size,ier,7,strlen(file_in),2);
+   hropen(lun,PASSCHAR("example"),PASSCHAR(file_in),PASSCHAR("px     "),record_size,ier,7,strlen(file_in),2);
```
ferdymercury, line 315:
```suggestion
   hropen(lun,PASSCHAR("example"),PASSCHAR(file_in),PASSCHAR("px"),record_size,ier,7,strlen(file_in),2);
```
Does not seem to make much difference.

misc/minicern/CMakeLists.txt
```diff
@@ -15,4 +15,6 @@ target_link_libraries(minicern ${CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES})
 # Disable optimization since it some cases was causing crashes.
 # Disable warnings, since what has worked for 40 years...
 # (see https://sft.its.cern.ch/jira/browse/ROOT-9179 for the warnings)
-set_target_properties(minicern PROPERTIES COMPILE_FLAGS "-O0 -w")
+# set_target_properties(minicern PROPERTIES COMPILE_FLAGS "-O0 -w")
+set_target_properties(minicern PROPERTIES COMPILE_FLAGS "-fsanitize=undefined -fsanitize=address")
+target_link_options(minicern BEFORE PUBLIC -fsanitize=undefined PUBLIC -fsanitize=address)
```
ferdymercury:
```suggestion
set_target_properties(minicern PROPERTIES COMPILE_FLAGS "-O0 -w")
# set_target_properties(minicern PROPERTIES COMPILE_FLAGS "-fsanitize=undefined -fsanitize=address")
# target_link_options(minicern BEFORE PUBLIC -fsanitize=undefined PUBLIC -fsanitize=address)
```

main/src/h2root.cxx
```diff
@@ -329,6 +330,8 @@ int main(int argc, char **argv)
 
    if (!hfile) {
       printf("Error: can't open output file: %s \n",file_out);
+      if (argc <= 2)
+         delete[] file_out;
```
silverweed, line 337: What about changing `file_out` to a `std::string` instead?

main/src/h2root.cxx
```diff
@@ -339,7 +342,9 @@ int main(int argc, char **argv)
    hfile->ls();
    hfile->Close();
    delete hfile;
-   return(0);
+   if (argc <= 2)
+         delete[] file_out;
```
silverweed, line 349: Same comment as above

main/src/h2root.cxx
```diff
@@ -312,7 +312,8 @@ int main(int argc, char **argv)
 
    int lun = 10;
 #ifndef WIN32
-   hropen(lun,PASSCHAR("example"),PASSCHAR(file_in),PASSCHAR("px"),record_size,ier,7,strlen(file_in),2);
+   auto opt = PASSCHAR("px ");
+   hropen(lun,PASSCHAR("example"),PASSCHAR(file_in),opt,record_size,ier,7,strlen(file_in),2);
```
pcanal, line 319: Why increase the size of 'only' the 3rd string?  What was the result of the test?

ferdymercury: px is being changed to pxc in the fortran side, and it was leading to a crash. So i explored whether we need a buffer of at least 3 chars on Cs side, too.
Not sure if the other strings are being modified on Fortrans side.

ferdymercury:
@pcanal @hageboeck @couet @guitargeek
In MacOS, I am seeing `Library not loaded: @rpath/libgfortran.5.dylib` and `ld: warning: ignoring duplicate libraries: '-lemutls_w', '-lgcc', '-lgfortran', '-lquadmath'`
Do you know if this has something to do with incremental building?
Or is it an issue with CMake not properly setting the rpath?
Maybe related: https://github.com/scipy/scipy/issues/19387#issuecomment-1763399481

main/src/h2root.cxx
ferdymercury: Thanks, sounds good, I'd like to avoid changing this until finding out if the memory corruption with "px" disappears.

(Since I do not want to make an extra mistake by translating to C++ these four obscure lines:)
```
      strlcpy(file_out,file_in,nchf);
      char *dot = strrchr(file_out,'.');
      if (dot) strcpy(dot+1,"root");
      else     strlcat(file_out,".root",nchf);
```

main/src/h2root.cxx
pcanal: Fair enough.  If this is the case, we will need to add this explanation as a explanatory comment.

main/src/h2root.cxx
```diff
@@ -312,7 +312,8 @@ int main(int argc, char **argv)
 
    int lun = 10;
 #ifndef WIN32
-   hropen(lun,PASSCHAR("example"),PASSCHAR(file_in),PASSCHAR("px"),record_size,ier,7,strlen(file_in),2);
+   auto opt = PASSCHAR("px ");
```
ferdymercury, line 318:
```suggestion
   // "px" is a string being changed to "pxc" in the fortran side; since it's a temporary on C's side, we need a buffer of at least 3 chars on Cs side, too. Predefine it as a extra variable "px " since that way the space will be replace by c instead of appending a new char.
   auto opt = PASSCHAR("px ");
```

main/src/h2root.cxx
pcanal: Let's have this in a separate PR.

pcanal:
> Outstanding asan diagnostics:

We should open a issue to record those.

pcanal (approved):
LGTM.

ferdymercury:
> LGTM.

What about the libgfortran dyld failure on mac?

ferdymercury:
> > Outstanding asan diagnostics:
> 
> We should open a issue to record those.

added to https://github.com/root-project/root/issues/14682

pcanal:
> What about the libgfortran dyld failure on mac?

We indeed need to 'fix' that part before merging.

couet:
@ferdymercury 

> @pcanal @hageboeck @couet @guitargeek In MacOS, I am seeing `Library not loaded: @rpath/libgfortran.5.dylib` 

I see this message on the CI build for instance on macphsft31. I checked on that one and gfortran seems properlly installed:

```
sftnight@macphsft31 ~ % which gfortran                                      
/usr/local/bin/gfortran
sftnight@macphsft31 ~ % ls -l /usr/local/lib/*for*
lrwxr-xr-x  1 root  wheel  43 Mar 20  2024 /usr/local/lib/libgfortran.5.dylib -> /usr/local/gfortran/lib/libgfortran.5.dylib
sftnight@macphsft31 ~ % ls -l /usr/local/gfortran/lib/libgfortran.5.dylib
-rwxr-xr-x  1 root  wheel  1847232 Nov 18  2022 /usr/local/gfortran/lib/libgfortran.5.dylib
sftnight@macphsft31 ~ % 
```

I checked the other Macs : It is the same

ferdymercury:
> 
> > @pcanal @hageboeck @couet @guitargeek In MacOS, I am seeing `Library not loaded: @rpath/libgfortran.5.dylib`
> 
> I see this message on the CI build for instance on macphsft31. I checked on that one and gfortran seems properlly installed:
> 
> ```
> sftnight@macphsft31 ~ % which gfortran                                      
> /usr/local/bin/gfortran
> sftnight@macphsft31 ~ % ls -l /usr/local/lib/*for*
> lrwxr-xr-x  1 root  wheel  43 Mar 20  2024 /usr/local/lib/libgfortran.5.dylib -> /usr/local/gfortran/lib/libgfortran.5.dylib
> sftnight@macphsft31 ~ % ls -l /usr/local/gfortran/lib/libgfortran.5.dylib
> -rwxr-xr-x  1 root  wheel  1847232 Nov 18  2022 /usr/local/gfortran/lib/libgfortran.5.dylib
> sftnight@macphsft31 ~ % 
> ```
> 
> I checked the other Macs : It is the same

Could it be that this related to https://github.com/root-project/root/issues/19134 ?

@guitargeek should I try to apply your https://github.com/root-project/root/pull/19135/ here?

guitargeek:
I don't think adding something hardcoded to the RPATH is a good idea here, because on the users system the `libfortran.so` might be in a different place. I think the solution is to just set the paths for the test environment correctly:

https://github.com/root-project/root/pull/19174

But let's see if that actually works :laughing:

main/src/h2root.cxx
```diff
@@ -312,7 +312,11 @@ int main(int argc, char **argv)
 
    int lun = 10;
 #ifndef WIN32
-   hropen(lun,PASSCHAR("example"),PASSCHAR(file_in),PASSCHAR("px"),record_size,ier,7,strlen(file_in),2);
+   // "px" is a string being changed to "pxc" in the fortran side; since it's a temporary on C's side,
+   // we need a buffer of at least 3 chars on Cs side, too. Predefine it as a extra variable "px " since that way
+   // the space will be replaced by 'c' on the preallocated memory, instead of appending a new char (new memory) to "px".
+   auto opt = PASSCHAR("px ");
+   hropen(lun,PASSCHAR("example"),PASSCHAR(file_in),opt,record_size,ier,7,strlen(file_in),2);
```
hahnjo, line 319: This explanation does not make sense to me because, as far as I can see, the string passed into `HROPEN` is copied into a character array of length 8:
https://github.com/root-project/root/blob/cc14662a25876cf755e1e1a1d7bfa88e5fb13500/misc/minicern/src/hbook.f#L250-L251
Appending a space should not change things; if it does, it's either moving or hiding another problem...

ferdymercury: Yes, it's surprising that march_native no longer crashes with this fix.
I have no idea about Fortran, but, maybe the problem is that the assumed length of CHOPTT in Fortran is off by one, in the sense that it also counts the null-terminator as part of the length, so adding a space forces the copy to work correctly?

ferdymercury: https://community.intel.com/t5/Intel-Fortran-Compiler/Inconsistent-behavior-with-null-terminated-strings/m-p/808339#M41364

It could also be some of these other issues:
https://github.com/root-project/root/issues/11657
https://github.com/root-project/root/issues/14682#issuecomment-2985345816

ferdymercury: @hahnjo it's confirmed. I added this debug line:

```
      CHOPT=CHOPTT
      print*, 'HROPEN0: "',CHOPT,'"',LEN(CHOPTT)
```

And I get: 

On mac15 and mac-beta:
`HROPEN0: "px      "           2`

On mac13 and mac14:
`HROPEN0: "px exam"    2`

which is the sum of "px" and "example" separate arguments in C.

The question now is if this is a gfortran-version-compiler-bug, or if it's an undefined behavior, so we need to fix it by hand.

If I add: 
```
      DO 11 I=LEN(CHOPTT)+1,8
         CHOPT(I:I) = ' '
  11  CONTINUE
```
Then it seems to work fine in mac13, but not on mac14. See https://github.com/root-project/root/actions/runs/15984458534/job/45085965269?pr=19191

Maybe related: https://www.tek-tips.com/threads/string-copy.1609892/post-6335242

Issue #18213, opened 2025-04-01
macos sequoia 15.4 homebrew - Failed to Open ROOT after today's update


yansonhuang:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

On my end, after today's sequoia 15.4 update. Homebrew failed to correctly build root (I tried to reinstall but no luck either.) I suspect it is a change of library or some C++ patches in the new MAC OS system. Also please see the following command line when I tried to open root after the installation:

/opt/homebrew/Cellar/root/6.34.06/etc/root/cling/std_darwin.modulemap:540:12: error: header 'experimental/__memory' not found
    header "experimental/__memory"
           ^
input_line_1:1:10: note: submodule of top-level module 'std' implicitly imported here
#include <new>
         ^
Warning in cling::IncrementalParser::CheckABICompatibility():
  Failed to extract C++ standard library version.
Warning in cling::IncrementalParser::CheckABICompatibility():
  Possible C++ standard library mismatch, compiled with _LIBCPP_ABI_VERSION '1'
  Extraction of runtime standard library version was: ''
input_line_2:1:2: fatal error: module file '/tmp/root-20250327-18543-ll7y48/root-6.34.06/builddir/lib/_mach_port_t.pcm' not found: module file not found
#include <cling/Interpreter/RuntimeUniverse.h>
 ^
input_line_2:1:2: note: imported by module 'pthread' in '/opt/homebrew/Cellar/root/6.34.06/lib/root/pthread.pcm'
input_line_2:1:2: note: imported by module 'std' in '/opt/homebrew/Cellar/root/6.34.06/lib/root/std.pcm'
input_line_2:1:2: note: imported by module 'Cling_Runtime_Extra' in '/opt/homebrew/Cellar/root/6.34.06/lib/root/Cling_Runtime_Extra.pcm'
input_line_2:1:2: note: imported by module 'Cling_Runtime' in '/opt/homebrew/Cellar/root/6.34.06/lib/root/Cling_Runtime.pcm'
<<< cling interactive line includer >>>: fatal error: module file '/opt/homebrew/Cellar/root/6.34.06/lib/root/std.pcm' is out of date and needs to be rebuilt
<<< cling interactive line includer >>>: note: imported by module 'Darwin' in '/opt/homebrew/Cellar/root/6.34.06/lib/root/Darwin.pcm'
Failed to load module Darwin
Failed to load module std
Failed to load module ROOT_Foundation_C
Failed to load module ROOT_Rtypes
[...]

PR #18235, opened 2025-04-02
std_darwin.modulemap: Wrap modulemap dynamically

devajithvs:
# This Pull request:
Fixes https://github.com/root-project/root/issues/18213 for good, in a future proof way.

## Checklist:

- [x] tested changes locally

This PR fixes #


devajithvs:
@vgvassilev @hahnjo I was wondering if we should still consider this approach or try to find a different solution.

hahnjo:
I would still like to see an automatic solution because at least that gives us a chance that a changed modulemap in an SDK release will work automatically, while currently it is almost guaranteed to break (because of private headers) and we need to rush to fix it and then maintain multiple versions. This is a no-risk improvement over the current situation because if nothing changes in the SDK, the script will just generate the same modifications every time.

core/clingutils/CMakeLists.txt
```diff
@@ -129,10 +129,19 @@ if (runtime_cxxmodules)
   if (APPLE)
     if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 15.0.0.15000309)
       set(custom_modulemaps ${custom_modulemaps} std_darwin.MacOSX14.2.sdk.modulemap)
-    elseif(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 17.0.0.17000013)
-      set(custom_modulemaps ${custom_modulemaps} std_darwin.MacOSX15.2.sdk.modulemap)
     else()
-      set(custom_modulemaps ${custom_modulemaps} std_darwin.modulemap)
+      set(GENERATED_MODULEMAP ${CMAKE_BINARY_DIR}/std_darwin.modulemap)
```
hahnjo:
```suggestion
      set(GENERATED_MODULEMAP ${CMAKE_PROJECT_BINARY_DIR}/std_darwin.modulemap)
```
in case ROOT ever gets added as subproject, or maybe even
```suggestion
      set(GENERATED_MODULEMAP ${CMAKE_CURRENT_BINARY_DIR}/std_darwin.modulemap)
```
?

core/clingutils/CMakeLists.txt
```diff
@@ -129,10 +129,19 @@ if (runtime_cxxmodules)
...
+
+      add_custom_command(
+        OUTPUT ${GENERATED_MODULEMAP}
+        COMMAND ${Python3_EXECUTABLE} ${CMAKE_SOURCE_DIR}/cmake/scripts/std_modulemap_darwin_fix.py ${GENERATED_MODULEMAP}
```
hahnjo:
```suggestion
        COMMAND ${Python3_EXECUTABLE} ${CMAKE_PROJECT_SOURCE_DIR}/cmake/scripts/std_modulemap_darwin_fix.py ${GENERATED_MODULEMAP}
```

core/clingutils/CMakeLists.txt
```diff
@@ -129,10 +129,19 @@ if (runtime_cxxmodules)
...
+        DEPENDS ${CMAKE_SOURCE_DIR}/cmake/scripts/std_modulemap_darwin_fix.py
+        COMMENT "Generating wrapped std modulemap for macOS SDK"
+      )
+      add_custom_target(generate_std_modulemap DEPENDS ${GENERATED_MODULEMAP})
+      list(APPEND copy_commands COMMAND ${CMAKE_COMMAND} -E copy ${GENERATED_MODULEMAP} ${CMAKE_BINARY_DIR}/etc/cling/std_darwin.modulemap)
```
hahnjo, line 141: (just to note: here `CMAKE_BINARY_DIR` is likely correct because we want it relative to the top-level build directory)

hahnjo (approved):
LGTM! We can probably get rid of `std_darwin.MacOSX14.2.sdk.modulemap` as well as we dropped support for `mac13` a while ago...

PR #18238, opened 2025-04-02
Fix build issue on macOS 15.4 / XCode 16.3

ktf:
`less` is defined in functional according to the C++ standard.


ktf:
I think the issue with the mac-beta is not in my PR.

PR #18239, opened 2025-04-03
[ntuple] apply I/O rules also to persistent members

jblomer:


tree/ntuple/v7/test/CustomStruct.hxx
```diff
@@ -266,6 +266,11 @@ struct CoordinatesWithIORules {
    float fPhi; //!
 };
 
+struct LowPrecsionFloatWithIORules {
```
silverweed, line 269: Typo: `LowPrecision`

tree/ntuple/v7/test/CustomStructLinkDef.h
```diff
@@ -106,6 +106,14 @@
    "CoordinatesWithIORules" target = "fPhi,fR" include = "cmath" code =                                         \
       "{ fR = sqrt(onfile.fX * onfile.fX + onfile.fY * onfile.fY); fPhi = atan2(onfile.fY, onfile.fX); }"
 
+#pragma link C++ options = version(3) class LowPrecisionFloatWithIORules + ;
+
+#pragma read sourceClass = "LowPrecisionFloatWithIORules" source = "float fLast8BitsZero;" version =       \
+   "[3]" targetClass = "LowPrecisionFloatWithIORules" target = "fLast8BitsZero" include = "cstring" code = \
+      "{ unsigned int bits; std::memcpy(&bits, &onfile.fLast8BitsZero, sizeof(bits)); \
```
hahnjo, line 113: Maybe better to use `std::uint32_t` instead of `unsigned int` as well?

hahnjo (approved):
LGTM, one suggestion

PR #18242, opened 2025-04-03
[tmva][sofie]  Optimize LayerNormalization operator

lmoneta:
For square operations, do not use std::pow with base 2 since it is slower than simple multiplication This speeds up evaluation on ATLAS GNN tracking model by 40%

Also avoids allocating a small shape vector in LayerNorm and improve the indentation of the sigmoid operator

### DIFF
--- a/tmva/sofie/inc/TMVA/ROperator_LayerNormalization.hxx
+++ b/tmva/sofie/inc/TMVA/ROperator_LayerNormalization.hxx
@@ -155,11 +155,11 @@ public:
       return out.str();
    }
 
-   std::string Generate(std::string OpName) override
+   std::string Generate(std::string opName) override
    {
-      OpName = "op_" + OpName;
+      opName = "op_" + opName;
       if (fShapeX.empty()) {
-         throw std::runtime_error("TMVA::SOFIE LayerNormalization operator " + OpName +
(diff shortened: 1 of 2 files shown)


tmva/sofie/inc/TMVA/ROperator_LayerNormalization.hxx
```diff
@@ -169,18 +169,20 @@ public:
...
+         inputShape[i] = fShapeX[i].GetVal();
+         //out << fShapeX[i].GetVal();
+         //if (i + 1 < fSize) {
+         //   out << ",";
+         //}
       }
-      out << "});\n";
-      std::string inputShape = OpName + "_InputShape";
+      //out << "});\n";
+      //std::string inputShape = opName + "_InputShape";
```
sanjibansg, line 185: we can maybe remove the commented code here

sanjibansg (approved):
LGTM! Thanks for the contribution. Just a small comment on removing the commented code before we merge. Thanks!

tmva/sofie/inc/TMVA/ROperator_LayerNormalization.hxx
lmoneta: Done

PR #18244, opened 2025-04-03
[core] improve TDirectory::DecodeNameCycle

silverweed:
- avoid scanning the string twice
- allow `name` to be null if one only wants the cycle
- return the actual name length so the caller can know whether the name was truncated or not
- improve function documentation

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/base/inc/TDirectory.h
+++ b/core/base/inc/TDirectory.h
@@ -272,8 +272,8 @@ public:
    /// \brief Write an object with proper type checking.
    /// \param[in] obj Pointer to an object to be written.
    /// \param[in] name Name of the object in the file.
-   /// \param[in] option Options. See TDirectory::WriteTObject.
-   /// \param[in] bufsize Buffer size. See TDirectory::WriteTObject.
+   /// \param[in] option Options. See TDirectoryFile::WriteTObject.
+   /// \param[in] bufsize Buffer size. See TDirectoryFile::WriteTObject.
    ///
    /// This overload takes care of instances of classes that are not derived
    /// from TObject. The method redirects to TDirectory::WriteObjectAny.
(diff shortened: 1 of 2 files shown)


hageboeck (approved):
LGTM!

core/base/inc/TDirectory.h
```diff
@@ -303,7 +303,7 @@ can be replaced with the simpler and exception safe:
    virtual void        WriteKeys() {}
 
    static Bool_t       Cd(const char *path);
-   static void         DecodeNameCycle(const char *namecycle, char *name, Short_t &cycle, const size_t namesize = 0);
+   static size_t DecodeNameCycle(const char *namecycle, char *name, Short_t &cycle, const size_t namesize = 0);
```
pcanal, line 306:
```suggestion
   static size_t       DecodeNameCycle(const char *namecycle, char *name, Short_t &cycle, const size_t namesize = 0);
```
please align `Decode` and `Cd`

core/base/src/TDirectory.cxx
```diff
@@ -1294,55 +1294,78 @@ void TDirectory::SetName(const char* newname)
...
-      // No ';' found
-      len = strlen(buffer);
-      ni = &buffer[len];
+   if (name) {
+      size_t truncatedNameLen = nameLen;
+      if (namesize) {
+         // accommodate string terminator
+         truncatedNameLen = std::min(truncatedNameLen, namesize - 1);
+      } else {
+         ::Warning("TDirectory::DecodeNameCycle",
```
pcanal, line 1339:
```suggestion
         ::Error("TDirectory::DecodeNameCycle",
```

core/base/src/TDirectory.cxx
```diff
@@ -1294,55 +1294,78 @@ void TDirectory::SetName(const char* newname)
...
+      while (ch) {
+         if (ch == ';') {
+            semicolonIdx = nameLen;
+            break;
+         }
+         ++nameLen;
+         ch = buffer[nameLen];
+      }
+   }
+   assert(semicolonIdx <= static_cast<int>(nameLen));
```
pcanal, line 1331: Isn't it:
```suggestion
   assert(semicolonIdx >= static_cast<int>(nameLen));
```
or even
```suggestion
   assert(semicolonIdx == static_cast<int>(nameLen));
```
?

pcanal: oups, I forgot about the case where there is no semi-colon.  So really it is:
```suggestion
   assert(semicolonIdx == -1 || semicolonIdx == static_cast<int>(nameLen));
```

pcanal (approved):
Thanks.

core/base/src/TDirectory.cxx
silverweed: Isn't this potentially a breaking change? (if the user set up a custom error handler)

PR #18250, opened 2025-04-03
[cling][JIT] Inject compiler-rt complex division symbols on  macOS

devajithvs:
# This Pull request:

Clang may emit calls to compiler-rt symbols such as __divdc3 for
std::complex<T> division. These may not be resolved during JIT'ing
on macOS when using Cling.

This patch forward-declares the relevant symbols and explicitly
injects them into the JIT symbol table

Similar to: https://github.com/root-project/root/commit/4b6075b

## Checklist:

- [x] tested changes locally

This PR fixes #

### DIFF
--- a/interpreter/cling/lib/Interpreter/IncrementalJIT.cpp
+++ b/interpreter/cling/lib/Interpreter/IncrementalJIT.cpp
@@ -408,6 +408,28 @@ CreateTargetMachine(const clang::CompilerInstance& CI, bool JITLink) {
   return cantFail(JTMB.createTargetMachine());
 }
 
+#ifdef __APPLE__
+// Forward-declare compiler-rt complex division helpers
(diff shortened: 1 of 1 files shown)


interpreter/cling/lib/Interpreter/IncrementalJIT.cpp
```diff
@@ -37,6 +37,10 @@
 #include <sys/stat.h>
 #endif
 
+#ifdef __APPLE__
+#include <complex>
+#endif
+
```
hahnjo, line 43: This can probably be removed

devajithvs: Done!

hahnjo (approved):
LGTM

PR #18257, opened 2025-04-04
[TMVA] Disable relevant PyMVA tests if BLAS is not found

guitargeek:
This commit makes sure that the tests of PyMVA that require BLAS (indirectly via SOFIE) are not run when BLAS is not available.

Following up on this PR, which did the same change already for the SOFIE tests:

https://github.com/root-project/root/pull/18065


ferdymercury:
Thanks! Does this maybe fix https://github.com/root-project/root/issues/16720 ?

guitargeek:
At least partially. This test is correctly vetoed now:
```txt
353-gtest-tmva-pymva-TestRModelParserKeras
```

ferdymercury:
Thanks! The remaining ones could be maybe vetoed here? https://github.com/guitargeek/root/blob/pymva_tests/tutorials/CMakeLists.txt#L321

guitargeek:
You're right. I'll add a commit to this PR that will do the veto for the tutorials

pcanal (approved):
Thanks

PR #18259, opened 2025-04-04
[tree] Avoid reading from nullptr in TTreeReader.

hageboeck:
When a TTreeReaderValue is not initialised correctly, it generates an error message. It could, however, inadvertently read a nullptr in doing that.

### DIFF
--- a/tree/treeplayer/src/TTreeReaderValue.cxx
+++ b/tree/treeplayer/src/TTreeReaderValue.cxx
@@ -832,8 +832,8 @@ void ROOT::Internal::TTreeReaderValueBase::ErrorAboutMissingProxyIfNeeded()
 {
    // Print the error only if the branch name does not appear in the list of
    // missing proxies that the user explicitly requested not to error about
-   if (std::find(fTreeReader->fMissingProxies.cbegin(), fTreeReader->fMissingProxies.cend(), fBranchName.Data()) ==
-       fTreeReader->fMissingProxies.cend())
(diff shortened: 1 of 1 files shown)


tree/treeplayer/src/TTreeReaderValue.cxx
```diff
@@ -832,8 +832,8 @@ void ROOT::Internal::TTreeReaderValueBase::ErrorAboutMissingProxyIfNeeded()
 {
    // Print the error only if the branch name does not appear in the list of
    // missing proxies that the user explicitly requested not to error about
-   if (std::find(fTreeReader->fMissingProxies.cbegin(), fTreeReader->fMissingProxies.cend(), fBranchName.Data()) ==
-       fTreeReader->fMissingProxies.cend())
+   if (!fTreeReader || std::find(fTreeReader->fMissingProxies.cbegin(), fTreeReader->fMissingProxies.cend(),
+                                 fBranchName.Data()) == fTreeReader->fMissingProxies.cend())
       Error("TTreeReaderValue::Get()", "Value reader not properly initialized, did you call "
```
pcanal, line 837: We may wan to the branchname in the error message.

pcanal (approved):
Thanks.

tree/treeplayer/src/TTreeReaderValue.cxx
hageboeck: Good point! I will add that in PR #18265 so the CI on this one doesn't need to rerun

vepadulano (approved):
Thank you!

PR #18264, opened 2025-04-04
Fix a warning when compiling quartz on macos 15.4 with clt 16.3

couet:
Fix a warning when compiling quartz on macos 15.4 with clt 16.3


pcanal (approved):

dpiparo:
@couet we'll need this on 6.34 and 6.32 as well.

Issue #18195, opened 2025-03-29
TBufferJSON not serializing std::map


rlalik:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

I'm trying to run example from the `TBufferJSON` doc, but getting nothing but it workd for vectors.
```
root [0] std::map<int,string> m;
root [1] m[1] = "number 1";
root [2] m[2] = "number 2";
root [3] TBufferJSON::ToJSON(&m)
(TString) "[]"[2]
root [4] std::vector<int> vect = {1,4,7};
root [5] TBufferJSON::ToJSON(&vect)
(TString) "[1, 4, 7]"[9]
``` 
The expected results as per documentation should be:
```
[
  {"$pair" : "pair<int,string>", "first" : 1, "second" : "number 1"},
  {"$pair" : "pair<int,string>", "first" : 2, "second" : "number 2"}
]
```

### Reproducer

Run the example from `TBufferJSON` documentation for maps.

### ROOT version

```
ROOT v6.34.04
Built for linuxx8664gcc on Feb 14 2025, 10:28:48
From tags/6-34-04@6-34-04
With 
Binary directory: /usr/bin
```

### Installation method

from source

### Operating system

Linux - Gentoo

rlalik:
I also have even more confusing and inconsistent behaviour. Depend on the vector length and `compact` value, I have different results:
```
root [22] std::vector<int> vect = {1, 4, 7, 0, 1};

root [23] TBufferJSON::ToJSON(&vect, 23)
(TString) "[1,4,7,0,1]"[11]

root [24] TBufferJSON::ToJSON(&vect, 0)
(TString) "[1, 4, 7, 0, 1]"[15]

root [25] TBufferJSON::ToJSON(&vect)
(TString) "[1, 4, 7, 0, 1]"[15]


root [26] std::vector<int> vect = {1, 4, 7, 0, 1, 2};

root [27] TBufferJSON::ToJSON(&vect, 23)
(TString) "{"$arr":"Int32","len":6,"v":[1,4,7,0,1,2]}"[42]

root [28] TBufferJSON::ToJSON(&vect, 0)
(TString) "[1, 4, 7, 0, 1, 2]"[18]

root [29] TBufferJSON::ToJSON(&vect)
(TString) "[1, 4, 7, 0, 1, 2]"[18]
```
This cannot be so inconsistent.

linev:
> Depend on the vector length and compact value, I have different results:

[...]

PR #18266, opened 2025-04-04
[json] Check dictionary for std::map object in streaming

linev:
When stream directly `std::map` object in macro like:
```
   std::map<int,string> m;
   m[1] = "number 1";
   m[2] = "number 2";
   auto json = TBufferJSON::ToJSON(&m);
```
Real dictionary class for `std::map` is required.
It checked automatically when map is member of user class,
but was not checked when map object itself streamed

### DIFF
--- a/io/io/src/TBufferJSON.cxx
+++ b/io/io/src/TBufferJSON.cxx
@@ -1364,6 +1364,12 @@ void TBufferJSON::JsonWriteObject(const void *obj, const TClass *cl, Bool_t chec
          map_convert = 2; // mapped into normal object
       else
          map_convert = 1;
+
+      if (!cl->HasDictionary()) {
+         Error("JsonWriteObject", "Cannot stream class %s without dictionary", cl->GetName());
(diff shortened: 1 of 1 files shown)


linev:
It prevents crash of ROOT in situation reported in #18195 - which is just our documentation.

But it is still open question if code was working before without dictionary.

dpiparo:
It is my understanding that this never worked, right @rlalik?

linev:
Code taken from here: https://github.com/root-project/root/blob/master/io/io/src/TBufferJSON.cxx#L50-L64

I put it in comments in ROOT version 6.18. 
But I cannot guarantee that I did it without dictionary.

One probably can try to start ROOT 6.18 and test it

rlalik:
> It is my understanding that this never worked, right @rlalik?

It was the very first time I used this feature so cannot say about the past.

linev:
I check `6.20.08-ubuntu20.04` docker image - it does not work without dictionary.
I will adjust documentation

pcanal (approved):

PR #18269, opened 2025-04-05
[cppyy] Disable warnings for invalid function casts

vepadulano:
Some C++ functions defined in CPyCppyy are passed to CPython API, e.g. to build Python methods that use some C++ functionality. This is done by creating a instance of a PyMethodDef struct
(https://docs.python.org/3/c-api/structures.html#c.PyMethodDef). One data member of this struct is of type PyCFunction
(https://docs.python.org/3/c-api/structures.html#c.PyCFunction), a typedef of a function with signature
```
PyObject *PyCFunction(PyObject *self,
                      PyObject *args);
```
In many cases, the C++ functions that are used as the data member of the PyMethodDef are not directly implemented as PyCFunction, thus need to be cast. This cast is often invalid, since many of such functions do not really respect the function signature prescribed by the API. This pattern is actually encouraged for CPython extension implementations by the official CPython docs
(https://docs.python.org/3/extending/extending.html#keyword-parameters-for-extension-functions). As such, the compiler warnings that are generated becausee of the invalid function casts, for example
```
root/bindings/pyroot/cppyy/CPyCppyy/src/Pythonize.cxx:1949:50: warning: cast from 'PyObject *(*)(PyObject *)' (aka '_object *(*)(_object *)') to 'PyCFunction' (aka '_object *(*)(_object *, _object *)') converts to incompatible function type [-Wcast-function-type-mismatch]
 1949 |         Utility::AddToClass(pyclass, "__repr__", (PyCFunction)ComplexRepr, METH_NOARGS);
```
cannot be avoided and have to be suppressed. This is also done upstream, see https://github.com/wlav/CPyCppyy/commit/33d4a6ebe8ddd2ea62f6c422ca76ea98dfeda6b3

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/CMakeLists.txt
+++ b/bindings/pyroot/cppyy/CPyCppyy/CMakeLists.txt
@@ -73,9 +73,13 @@ if(NOT "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang" AND NOT MSVC)
     -Wno-unused-but-set-parameter)
 endif()
 
-# Disables warnings coming from PyCFunction casts
-if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND ${CMAKE_CXX_COMPILER_VERSION} VERSION_GREATER_EQUAL 8)
-  target_compile_options(${libname} PRIVATE -Wno-cast-function-type)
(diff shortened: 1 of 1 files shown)


guitargeek (approved):
Thank you very much for fixing the warnings and adding so much detail!

PR #18270, opened 2025-04-05
[df] Move RNTupleDS out of experimental

vepadulano:
And into the ROOT::RDF namespace, aligning it with all other RDataSource implementations.


silverweed (approved):
LGTM, thanks!

PR #18273, opened 2025-04-05
[math] Use std::vector instead of C array with run time sizes

lmoneta:
This should fix a warning seen on a new mac version

### DIFF
--- a/math/mathcore/inc/Fit/FitUtil.h
+++ b/math/mathcore/inc/Fit/FitUtil.h
@@ -232,7 +232,7 @@ namespace FitUtil {
         // Figure out the size of the SIMD vectors.
         constexpr static int vecSize = sizeof(ROOT::Double_v) / sizeof(double);
         double xBuffer[vecSize];
-        ROOT::Double_v xx[fDim];
+        std::vector<ROOT::Double_v> xx(fDim);
         for (unsigned int i = 0; i < fDim; ++i) {
            // The Load() function reads multiple values from the pointed-to
            // memory into xx. This is why we have to copy the input values from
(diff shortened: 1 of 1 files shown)


guitargeek (changes requested):
This introduces a manual memory allocation in `ExecFunc`, which is quite expensive and will probably kill all the performance benefit that you get from using `veccore`.

IMHO, an alternative solution needs to be found.

dpiparo:
Would this allow us to release macos 15.4 + xcode 16.3 compatible 6.32 and 6.34 versions while we think about a good and performant fix?
```c++
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wvla-cxx-extension"

     inline double ExecFunc(const IModelFunctionTempl<ROOT::Double_v> *f, const double *x, const double *p) const
     {
        // Figure out the size of the SIMD vectors.
        constexpr static int vecSize = sizeof(ROOT::Double_v) / sizeof(double);
        double xBuffer[vecSize];
        ROOT::Double_v xx[fDim];
        for (unsigned int i = 0; i < fDim; ++i) {
           // The Load() function reads multiple values from the pointed-to
           // memory into xx. This is why we have to copy the input values from
           // the x array into a zero-padded buffer to read from. Otherwise,
           // Load() would access the x array out of bounds.
           *xBuffer = x[i];
           for(int j = 1; j < vecSize; ++j) {
              xBuffer[j] = 0.0;
           }
           vecCore::Load<ROOT::Double_v>(xx[i], xBuffer);
        }
        auto res = (*f)(xx, p);
        return vecCore::Get<ROOT::Double_v>(res, 0);
     }
#pragma clang diagnostic pop
```

guitargeek:
Yes, if this really suppresses the warning, then great!

lmoneta:
I agree, we can suppress the warning as suggested by Danilo.

silverweed:
What is the maximum reasonable value that `fDim` may have? If we got by until now by allocating it on the stack we can probably just set a maximum size and assert that the dimension is less than that size...

guitargeek:
Yes, that's what I also discussed with @lmoneta in private

guitargeek:
Superseded by:
  * https://github.com/root-project/root/commit/c2def6e0cf8163e6725181f62f4a3f33db63d19d

PR #18276, opened 2025-04-06
[RF] Fix likelihood fits with constant probabliliy

guitargeek:
Fix likelihood fits with constant probabliliy, which is an edge case but it should still work with the new vectorizing CPU and CUDA backends.

If the probability is a scalar, the assumption that the number of events can be retrieved by `probas.size()` is wrong. The proposed solution is to always fill the weight buffer, even in case of unity weights, because then the number of events can be retrieved like `weights.size()`.

A new unit test to cover this case was implemented, and I verified locally that it also works with the CUDA backend.

Thanks to @Moelf for noticing this!


lmoneta (approved):
LGTM!
Thank you Jonas for this fix!

PR #18280, opened 2025-04-06
[RF] Use smart pointers in `RooStats::MCMCInterval`

guitargeek:
Use smart pointers in `RooStats::MCMCInterval` for more memory safety. Also avoid a leaking `RooArgList` and specify in the documentation that one of the setter functions takes ownership of the argument.


guitargeek:
Note that ROOT's schema evolution automatically deals with the transition from raw to smart pointer, and also the transition from C-style array to `std::vector` (not that I think that the `MCMCInterval` class is ever used for IO, the non-zero `ClassDef` version was probably just for proof...).

lmoneta (approved):
Thank you Jonas for these improvements in the code!

PR #18281, opened 2025-04-06
[v632][TMVA] Disable relevant PyMVA tests if BLAS is not found

dpiparo:
BP of https://github.com/root-project/root/pull/18258


guitargeek (approved):
LGTM! We don't have the situation that BLAS is missing on the CI, but this is still an improvement for other build configurations that I think should be merged.

PR #18282, opened 2025-04-06
[RF] Consistent collection proxy names between constructors

guitargeek:
We want consistent collection proxy names, because it doesn't make sense
to set different names in the different constructors of the same class.
This only breaks code that assumes a certain proxy name, like the RooFit
HS3 JSON I/O.

Also implement a new collection proxy constructor that just copies the
name. This constructor is now the recommended one.


lmoneta (approved):
LGTM!

PR #18283, opened 2025-04-06
[RF] Set default name argument value in `RooAbsArg::clone()` overrides

guitargeek:
The `RooAbsArg::clone()` method has a default `nullptr` value for the `newname` argument.

Not setting the default value in the derived classes brings us in the awkwared situation where the default empty newname works only when calling the function via a `RooAbsArg` base class pointer or reference, and one has to do `clone(nullptr)` explicitly when creating a clone with the same name. Or even having to pass the original name explicitly in the case of PyROOT, because passing `ROOT.nullptr` for a string is not supported.


lmoneta (approved):
LGTM!
Thank you for fixing this in the constructor of almost all RooFit classes

PR #18284, opened 2025-04-06
[cppyy] Do not use non-C++ GCC warning

vepadulano:
Although `bad-function-cast` is present for both Clang and GCC, on the latter it is only used for C and Objective-C code. Using `no-cast-function-type` seems to already disable the PyCFunction-cast-related warnings seen with AppleClang.


guitargeek (approved):
Thank you very much!

Issue #13200, opened 2023-07-06
[FreeBSD] roofit tests segfault


mkrzewic:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

roofit tests (test-stressroofit-batchmode-*) segfault at very end, maybe it rings a bell with the experts I cannot find the culprit (yet):
```
304: Test 52 : B Physics p.d.f.s......................................OK
304: Test 53 : Automated MC studies...................................OK
304: Test 54 : MC Study with chi^2 calculator.........................OK
304: Test 55 : MC Study with param rand. and Z calc...................OK
304: Test 56 : MC Studies with aux. obs. constraints..................OK
304:
304:  *** Break *** segmentation violation
304: ******************************************************************
304: *  SYS: FreeBSD hack 13.2-RELEASE-p1 FreeBSD 13.2-RELEASE-p1 GENERIC
304: ******************************************************************
304: ******************************************************************
304: *  ROOTMARKS =4262.5   *  Root6.29/01   20221221/1150
304: ******************************************************************
304: Time at the end of job = 18.882812 seconds
304:  Generating stack trace...
304:  0x0000000826fefe68 in _ZNSt3__112__hash_tableINS_12basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEEENS_4hashIS6_EENS_8equal_toIS6_EENS4_IS6_EEE25__emplace_unique_key_argsIS6_JRKS6_EEENS_4pairINS_15__hash_iteratorIPNS_1 at /usr/include/c++/v1/__hash_table:2070 from /home/mkrzewic/build/root/lib/libTree.so
304:  0x0000000830314f3e in TClass::TDeclNameRegistry::AddQualifiedName(char const*) at /usr/include/c++/v1/stdexcept:153 from /home/mkrzewic/build/root/lib/libCore.so
304:  0x0000000830315169 in TClass::InsertTClassInRegistryRAII::~InsertTClassInRegistryRAII() at /usr/include/c++/v1/stdexcept:153 from /home/mkrzewic/build/root/lib/libCore.so
304:  0x000000083032618c in TClass::SetUnloaded() at /usr/include/c++/v1/stdexcept:153 from /home/mkrzewic/build/root/lib/libCore.so
304:  0x00000008302d88cb in ROOT::RemoveClass(char const*, TClass*) at /usr/include/c++/v1/stdexcept:153 from /home/mkrzewic/build/root/lib/libCore.so
[...]

PR #18288, opened 2025-04-06
Fix build on FreeBSD

### LINKED ISSUE #18285: [FreeBSD] root build fails in master
### Check duplicate issues.

- [x] Checked for duplicates

### Description

make fails in multiple places in master on FreeBSD, a PR will follow to fix (#18288 ). This is just a placeholder bug.

### Reproducer

make

### ROOT version

master

### Installation method

source

### Operating system

FreeBSD

### Additional context

_No response_

mkrzewic:
# This Pull request:

This fixes the build on FreeBSD14.2 and 15

## Changes or fixes:

Includes fixes from the FreeBSD port maintainer @eamjensen plus some more recent fixes to deal with master

## Checklist:

- [ ] tested changes locally

This PR fixes #18285


ferdymercury:
Thanks a lot for this! Could you also check if this crash (https://github.com/root-project/root/issues/13200) is still present?

interpreter/cling/lib/Interpreter/Interpreter.cpp
```diff
@@ -455,8 +455,17 @@ namespace cling {
     const char* Attr = LangOpts.CPlusPlus ? " throw () " : "";
 #else
     const char* LinkageCxx = Linkage;
+#ifdef __FreeBSD__
```
dpiparo, line 458: Could we perhaps have a regular R__FREEBSD variable for consistency?

dpiparo (changes requested):
Thanks a lot for these changes! We are discussing internally how to proceed. Unfortunately we do not have FreeBSD builds in our CI, and the support is sort of "best effort" at the moment

interpreter/cling/lib/Interpreter/Interpreter.cpp
mkrzewic: there seems to be a R__FBSD defined in RConfig.hxx, could switch to that, but it also gets defined if __FreeBSD__ is there which also is used in a lot of places...

ferdymercury: Side note, there is also: `__386BSD__`, `__OpenBSD__`, `__NetBSD__`

vgvassilev (commented):
The interpreter part looks good to me.

PR #18302, opened 2025-04-07
[TMVA][SOFIE] Use `uint8_t` instead of `bool` in return types

guitargeek:
The `std::vector<bool>` is a special class that doesn't wrap contiguous
memory. It limits the usability of the return object (e.g. the memory
can't be adopted copy-free by a NumPy array). It prevents us
from using C-style arrays under the hood and differentiating through the
inference function.

Also, further simplify the code that deals with the return objects.


sanjibansg (approved):
LGTM! Thanks for this contribution Jonas!

PR #18303, opened 2025-04-07
[core] Allow to skip arguments by TApplication.

rlalik:
# This Pull request:

Current `TApplication` constructor will process all known input options and also positional arguments if they are valid files. With this commit the static function `TApplication::SkipParsingArguments()` tells `TApplication` to ignore arguments (but not options) and let to handle them by users, e.g. it can be file name which is not root file but must be converted from binary format to something else.

Currently I don't think there is possibility to do that beside doing some nasty manual operations on arguments.

```c++
int main(int argc, char* argv[])
{
    std::vector<std::string> v_argv;
    v_argv.reserve(argc + 5);
    for (int i = 0; i < argc; ++i) { v_argv.push_back(argv[i]); }
    spdlog::info("argc = {}  argv = {}", argc, v_argv);

    // TApplication app("app", &argc, argv, nullptr, -1);
    // TApplication::SkipParsingArguments();
    TApplication app("app", &argc, argv);

    v_argv.clear();
    for (int i = 0; i < argc; ++i) { v_argv.push_back(argv[i]); }
    spdlog::info("argc = {}  argv = {}", argc, v_argv);

    v_argv.clear();
    for (int i = 0; i < app.Argc(); ++i) { v_argv.push_back(app.Argv()[i]); }
    spdlog::info("Argc = {}  Argv = {}", app.Argc(), v_argv);
}
```
gives me:
```
argc = 5  argv = ["./bin/smxwebgui", "--aaa", "-x", "-b", "/home/rafal/cbm/data/data_1sec.bin"]
argc = 2  argv = ["./bin/smxwebgui", "--aaa"]
Argc = 5  Argv = ["./bin/smxwebgui", "--aaa", "-x", "-b", "/home/rafal/cbm/data/data_1sec.bin"]
```
With `TApplication::SkipParsingArguments();`, the output would look like:
```
argc = 5  argv = ["./bin/smxwebgui", "--aaa", "-x", "-b", "/home/rafal/cbm/data/data_1sec.bin"]
argc = 3  argv = ["./bin/smxwebgui", "--aaa", "/home/rafal/cbm/data/data_1sec.bin"] 
Argc = 5  Argv = ["./bin/smxwebgui", "--aaa", "-x", "-b", "/home/rafal/cbm/data/data_1sec.bin"] 
```

## Changes or fixes:

* Adds static function `TApplication::SkipArguments();`

## Checklist:

- [x] tested changes locally

This PR fixes #

### DIFF
--- a/core/base/inc/TApplication.h
+++ b/core/base/inc/TApplication.h
@@ -72,6 +72,7 @@ private:
 
    static Bool_t      fgGraphNeeded;    // True if graphics libs need to be initialized
    static Bool_t      fgGraphInit;      // True if graphics libs initialized
+   static Bool_t      fgSkipArguments;  // True if GetOptions() should skip processing positional arguments
 
    TApplication(const TApplication&) = delete;
    TApplication& operator=(const TApplication&) = delete;
@@ -163,6 +164,7 @@ public:
    static TList   *GetApplications();
    static void     CreateApplication();
    static void     NeedGraphicsLibs();
+   static void     SkipParsingArguments();
 
    ClassDefOverride(TApplication,0)  //GUI application singleton
 };
--- a/core/base/src/TApplication.cxx
+++ b/core/base/src/TApplication.cxx
@@ -53,6 +53,7 @@ TApplication (see TRint).
 TApplication *gApplication = nullptr;
 Bool_t TApplication::fgGraphNeeded = kFALSE;
 Bool_t TApplication::fgGraphInit = kFALSE;
+Bool_t TApplication::fgSkipArguments = kFALSE;
 TList *TApplication::fgApplications = nullptr;  // List of available applications
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -235,6 +236,15 @@ TApplication::~TApplication()
 void TApplication::NeedGraphicsLibs()
 {
    fgGraphNeeded = kTRUE;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+/// Static method. This method should be called if TApplication should skip
+/// processing positional arguments leaving this to user.
+
+void TApplication::SkipParsingArguments()
+{
+      fgSkipArguments = kTRUE;
 }
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -472,6 +482,7 @@ void TApplication::GetOptions(Int_t *argc, char **argv)
                argv[i] = null;
          }
       } else if (argv[i][0] != '-' && argv[i][0] != '+') {
+         if (fgSkipArguments) continue;
          Long64_t size;
          Long_t id, flags, modtime;
          char *arg = strchr(argv[i], '(');


guitargeek:
Hi, thanks a lot for the PR! We generally do PRs to the `master` branch first, so I changed the target branch of your PR before launching the tests.

guitargeek:
That didn't go well... Can you please rebase your PR on top of `master`? Thanks!

rlalik:
Yes, I did it. I though if you accept for the `v6-34-00-patches` you can later cherry pick to the `master`. With the rebase I have to recompile almost all of the root :-)

guitargeek:
Thanks! No, we don't do forward ports, only backports. And the `TApplication` class is rather stable, it probably doesn't matter that you did your developments against 6.34. The CI will tell :slightly_smiling_face:

rlalik:
I see that linter failed, but what should I do about these changes it want to do:
```diff
diff --git a/core/base/inc/TApplication.h b/core/base/inc/TApplication.h
index b0f6befe4e..dc7fc52c36 100644
--- a/core/base/inc/TApplication.h
+++ b/core/base/inc/TApplication.h
@@ -72,7 +72,7 @@ private:
 
    static Bool_t      fgGraphNeeded;    // True if graphics libs need to be initialized
    static Bool_t      fgGraphInit;      // True if graphics libs initialized
-   static Bool_t      fgSkipArguments;  // True if GetOptions() should skip processing positional arguments
+   static Bool_t fgSkipArguments;       // True if GetOptions() should skip processing positional arguments
 
    TApplication(const TApplication&) = delete;
    TApplication& operator=(const TApplication&) = delete;
@@ -164,7 +164,7 @@ public:
    static TList   *GetApplications();
    static void     CreateApplication();
    static void     NeedGraphicsLibs();
-   static void     SkipParsingArguments();
+   static void SkipParsingArguments();
 
    ClassDefOverride(TApplication,0)  //GUI application singleton
 };
diff --git a/core/base/src/TApplication.cxx b/core/base/src/TApplication.cxx
index 680550c4e8..a325ece8ac 100644
--- a/core/base/src/TApplication.cxx
+++ b/core/base/src/TApplication.cxx
@@ -244,7 +244,7 @@ void TApplication::NeedGraphicsLibs()
 
 void TApplication::SkipParsingArguments()
 {
-      fgSkipArguments = kTRUE;
+   fgSkipArguments = kTRUE;
 }
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -482,7 +482,8 @@ void TApplication::GetOptions(Int_t *argc, char **argv)
                argv[i] = null;
          }
       } else if (argv[i][0] != '-' && argv[i][0] != '+') {
-         if (fgSkipArguments) continue;
+         if (fgSkipArguments)
+            continue;
          Long64_t size;
          Long_t id, flags, modtime;
          char *arg = strchr(argv[i], '(');
```
The one in cxx are reasonable, but in the header, out of the line with rest of the file. I could reformat whole file to be consistent. I am not sure which command `git-clang-format` executes, is it 18, right? Because different clang version produce different outputs and my default is 19.

rlalik:
> Thanks! No, we don't do forward ports, only backports. And the `TApplication` class is rather stable, it probably doesn't matter that you did your developments against 6.34. The CI will tell 🙂
> 
> ** closed and reopened the PR because the CI didn't pick it up anymore after the rebase **

It is due to another 500 changes between master and v6-34-00-patches :-)

pcanal:
`clang-format` is indeed wrong for the header file and you can ignore it.

silverweed:
Just to better understand what you mean by "nasty manual operations on arguments": doesn't this accomplish the same thing?

```c++
int main(int argc, char **argv) {
  std::vector<std::string> argsCopies { argv, argv + argc };
  // ...
}
```

rlalik:
Maybe I exaggerate a bit :) but I meant to take (or better to reintroduce) the list of positional arguments from the unparsed list of arguments after `TApplication::GetOption()` removes them.
And your example, what is does it only creates a vector from the inputs. If this is done after creating `TApplication` one need to reintroduce the arguments into the vector at correct position (I could have custom option taking one argument which is or isn't the file). So why I called this nasty.

And there is more to the `GetOption()`. I think that it tries to do something with the files, like adding to the list of files. So there is some side effect to the `GetOption()` you may not want. If I am wrong then tell me.

linev:
In my mind this PR follows too simple approach - just ignore file arguments.
It may work in several use-cases, but by far not in all situations.
To be consequent - one should be able to configure filter which able to sort-out all user-specific args. 
Which can ends in extremely complex regex expression - which I never liked.

There is much simpler approach. If some command-line arguments should not be handled by `TApplication` - these arguments simply should not be provided to it. And this easily can be done with plain array or `std::vector<char *>`. 

I would not merge this PR.

silverweed:
> . If this is done after creating TApplication one need to reintroduce the arguments into the vector at correct position (I could have custom option taking one argument which is or isn't the file).

Sure, but if an application knows it has to handle the command line arguments I don't see why it cannot, as the first thing, just clone them before instancing TApplication. It only takes one line, it's straightforward and it doesn't require any specific knowledge about TApplication.
I do agree that TApplication consuming the arguments is weird and unexpected, but to me the proposed solution sounds more complicated than it needs to be.

rlalik:
@linev My approach is take it all or nothing. If you decided to skip arguments, you know what to do with it later.
> If some command-line arguments should not be handled by TApplication - these arguments simply should not be provided to it. And this easily can be done with plain array or std::vector<char *>.

If user provides a cmd line arguments, one need to manually pick them, is that what you mean? And that is tricky and nasty I would say. Requires extra work.

And you mention your use-case. Well in my use-cases whenever I need to write an app which must use `TApplication` usually I don't need and don't want root to parse any of the arguments. They are mostly useful when running just `root ...` command. But as I am forced to provide them to `TApplication`, they may be filter out - I might want to have custom `-x` or `-b` or anything else which this prevents me from.

Perhaps the simplest solution would be to have `TApplication` constructor which doesn't take any argv/argc? I know I can perhaps simulate it by calling with dummy args but as argv is taken by address, I need to create dummy.

silverweed:
> but as argv is taken by address, I need to create dummy.
Passing `0, nullptr` for argc and argv should also work

linev:
> If user provides a cmd line arguments, one need to manually pick them, is that what you mean? And that is tricky and nasty I would say. Requires extra work.

But is user want to use it - user MUST parse and analyze such arguments and pick interesting ones. 

In your particular application you can by definition say - file name is always last argument. 
Therefore TApplication will get `argc-1`. 
That's it - no extra work, no extra copies.

rlalik:
> > If user provides a cmd line arguments, one need to manually pick them, is that what you mean? And that is tricky and nasty I would say. Requires extra work.
> 
> But is user want to use it - user MUST parse and analyze such arguments and pick interesting ones.
> 
> In your particular application you can by definition say - file name is always last argument. Therefore TApplication will get `argc-1`. That's it - no extra work, no extra copies.

No, it's very bad assumption. I might have varying number of arguments, I don't want to be restricted by their number, order. I may want to have options with arguments like `-p param_file -c calib_file` and none of the ideas above will easily work if the arguments will be eaten by `GetOption()` leaving me out with `'-p', '-c'`...

silverweed:
> none of the ideas above will easily work 

I still don't get it: what does not easily work in my proposed solution? Of course you need to copy the arguments before instancing TApplication, but that qualifies as "easy" in my book, unless I'm missing something.

linev:
> No, it's very bad assumption. I might have varying number of arguments, I don't want to be restricted by their number, order. I may want to have options with arguments like -p param_file -c calib_file and none of the ideas above will easily work if the arguments will be eaten by GetOption() leaving me out with '-p', '-c'...

But it is exactly my point - pick your arguments and redirect only non-recognized ones to TApplication.
Why ROOT need to be instructed about your arguments which it does not used at all?

rlalik:
And here we come to the manual juggling with params. Picking up my arguments means I need to go through all values, compare it (so long list of if-else) put roots parameters in one vector, handle other in another way. I want to use some automatic tool for parsing arguments like getopt but it is not possible because of mess in root options, look below:

> OPTIONS:
>   -b                                   Run in batch mode without graphics
>   -x                                   Exit on exceptions
>   -e                                   Execute the command passed between single quotes
>   -n                                   Do not execute logon and logoff macros as specified in .rootrc
>   -t                                   Enable thread-safety and implicit multi-threading (IMT)
>   -q                                   Exit after processing command line macro files
>   -l                                   Do not show the ROOT banner
>   -a                                   Show the ROOT splash screen
>   **-config                              print ./configure options**
>   -h, -?, --help                       Show summary of options
>   --version                            Show the ROOT version
>   --notebook                           Execute ROOT notebook
>   --web                                Use web-based display for graphics, browser, geometry
>   --web=<type>                         Use the specified web-based display such as chrome, firefox, qt6
>                                        For more options see the documentation of TROOT::SetWebDisplay()
>   --web=off                            Disable any kind of web-based display

There are various standards but at least people try to be consistent, like in GNU, single dash is a short option, double dash is a long option, short dash with string is interpreted as gluing of short option, but you among all short and long options have just one option called `-config` which is a long option and goes out of the pattern or consistency with other long options.

`TApplication` is also a mess with parsing the options. It will eat up every file, whether root knows what to do with it or not. It will eat up root files, txt files, jpeg and whatever valid file on the disk is. It will leave out anything what isn't file. So if I pass a string which is a name of file to be written/created, it will work in one run because it will remain in the list of not parsed arguments, but in the second execution this file will be already eaten and program will behave in different way.

So with this commit at least I want to give to myself a chance to have control over what is root doing. It is fully backward compatible, so will not hurt any old code. Is **documented** what is does and if anyone decide to use it at least know what to expect. With current root behavior it's  lottery because what will survive `TApplication` execution depends on what is physically on the disk.

rlalik:
And to add to this, from the documentation of `TApplication` constructor:

> Create an application environment.
> 
> The application environment provides an interface to the graphics system and eventloop (be it X, Windows, macOS or BeOS). After creating the application object start the eventloop by calling its Run() method. The command line options recognized by TApplication are described in the GetOptions() method. The recognized options are removed from the argument array. The original list of argument options can be retrieved via the Argc() and Argv() methods. The appClassName "proofserv" is reserved for the PROOF system. The "options" and "numOptions" arguments are not used, except if you want to by-pass the argv processing by GetOptions() in which case you should specify numOptions<0. All options will still be available via the Argv() method for later use.

Says nothing about eating up arguments, only options, and there is clear distinction between options (starting with dash/double dash, also having optional or required arguments) and positional arguments. `TApplication` should care only about options which control its behavior, should not process the input arguments which should be of concern of user. `TApplication` just does to much, it should have one task of parsing options not handling arguments.

linev:
> So why any other file is also handled? I am passing bin file. So another discrepancy between official doc and real behavior.
> Sorry, but you will not convince me that everything is fine and I am making up problems.

This is another story. `TApplication` handle each argument which points to existing file as ROOT file and calls `TFile::Open()`. Disregard of its extension.

For me solution is clear - just describe this clearly in documentation and in `rootCommandLineOptionsHelp.h`.

rlalik:
> > I meant to take (or better to reintroduce) the list of positional arguments from the unparsed list of arguments after TApplication::GetOption() removes them.
> 
> Sorry if I come back to this, but isn't this functionality already available through `TApplication::Argc()` / `TApplication::Argv()` ?
> 
> I'm not sure I'm really getting what the problem is. I'm confused by this statement:
> 
> > Picking up my arguments means I need to go through all values, compare it (so long list of if-else) put roots parameters in one vector, handle other in another way.
> 
> Isn't this exactly what you would have to do anyway, regardless of TApplication? Why would you need TApplication to specifically skip the non-option arguments to avoid doing that?

The `Argc()` and `Argv()` return the original list of arguments while the `main`'s `argc` and `argv` are modified by `GetOption()`. What you propose is to parse the original list but even this is complicated due to inconsistency in long option names in root - the `-config` option.

And honestly I am not satisfied with all the propositions of workarounds, because their either do not work, do not work properly or are nasty pieces of code I don't want to write. You perhaps some use cases where this may work, but your use cases are not the only one and not the only valid.

By looking on what `--help' gives us:
```
[dir] if dir is a valid directory cd to it before executing
[data1.root...dataN.root] Open the given ROOT files; remote protocols (such as http://) are supported
[file1.C...fileN.C] Execute the ROOT macro file1.C ... fileN.C
Compilation flags as well as macro arguments can be passed, see format in https://root.cern/manual/root_macros_and_shared_libraries/
[file1_C.so...fileN_C.so] Load and execute file1_C.so ... fileN_C.so (or .dll if on Windows)
They should be already-compiled ROOT macros (shared libraries) or:
regular user shared libraries e.g. userlib.so with a function userlib(args)
```
it should only:
* change directory
* take root file
* run macro
* load library

and it makes sens only when running `root` - to write interpreted code, run macro, open browser.

But what is worse, root is inconsistent with above because actually **takes every valid file** (that exists on disk) even if it doesn't know what to do with it! Like my bin file.

And most likely, if someone writes standalone application, he:
* doesn't want to change directory
* doesn't want to run macro
* doesn't want to add root file to list of files
* doesn't want to load library (he can link against it!)

and even if he want to do some of it, it can do it easily in the app.

So in this commit I propose to make a possible to:
* let root parse all options it recognizes
* let the user parse remaining options by the parsing system of his choice
* let the user process the positional arguments the way he want, not the trivial and incorrect way `root` does now.

And to be also honest, `TApplication` is badly designed, it does to much that should. It should parse only options to setup the app (not `root`) environment:
```
  -b                                   Run in batch mode without graphics
  -x                                   Exit on exceptions
  -n                                   Do not execute logon and logoff macros as specified in .rootrc
  -t                                   Enable thread-safety and implicit multi-threading (IMT)
  -l                                   Do not show the ROOT banner
  -a                                   Show the ROOT splash screen
  -config                              print ./configure options
  --version                            Show the ROOT version
  --web                                Use web-based display for graphics, browser, geometry
  --web=<type>                         Use the specified web-based display such as chrome, firefox, qt6
                                       For more options see the documentation of TROOT::SetWebDisplay()
  --web=off                            Disable any kind of web-based display
  ```
(I removed `-e`, `-q`, `-h`, `--notebook` from the list, it my opinion it belongs to the interpreter functionality, not application). Parsing positional argument should be separate part, perhaps in `TRint` or other part which is strictly related to interpreted mode started with `root`, and I would really appreciate if these things were separated. This would then invalidate this PR but only if this separation is done.

So, I presented a lot of arguments why the current behavior is bad and problematic. So, just make your mind whether you accept this PR on not. I am not interested in discussing workaround to this obvious misbehavior.

PR #18305, opened 2025-04-07
[skip-ci] Minor documentation improvement for --web=server

2xB:
# This Pull request:

Minorly improves documentation for `--web=server` since I didn't find it immediately and also the environment variable `ROOT_WEBDISPLAY` was not documented at all.

Note that I believe that the `embed`/`embedded` option should also be documented in `RWebDisplayArgs` but I don't understand it enough to do so.

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/gui/webdisplay/src/RWebDisplayArgs.cxx
+++ b/gui/webdisplay/src/RWebDisplayArgs.cxx
@@ -153,6 +153,7 @@ bool RWebDisplayArgs::SetPosAsStr(const std::string &str)
 ///       local - either cef or qt6
 ///         off - disable web display
 ///          on - first try "local", then "native", then "default" (default option)
+///      server - run as web server without display (URL printed to command line; also available via e.g. RWebWindow::GetUrl)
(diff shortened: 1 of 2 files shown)


linev:
> Note that I believe that the embed/embedded option should also be documented in RWebDisplayArgs but I don't understand it enough to do so.

One should not use "embed" qualifier directly - code `window->Show("embed")` will does nothing.
It is handled inside static `RWebWindow::ShowWindow()` method to embed one window inside 'main' window.
Such mode used in `RBrowser` where all other widgets embeded into main `RWebWindow` - sharing same WebSocket connection. Therefore `root --web=embed` normally has no effect. 

I wrote _normally_, while in some special situations it can make sense. Like calling web_canv->ShowWebWindow("embed") will force creation of RWebWindow instance - which can be used later. But such use-cases are very special and described at the place where they used. Like:

https://github.com/linev/bootstrap-example/blob/v634/webwindow.cxx#L106

PR #18308, opened 2025-04-08
[proof] Deprecate PROOF and PROOFLITE with warnings

dpiparo:
since the tools will be removed in 6.38.00. RDataFrame and DistRDF can be used instead.

### DIFF
--- a/proof/proof/inc/TProof.h
+++ b/proof/proof/inc/TProof.h
@@ -1071,7 +1071,7 @@ public:
    static Int_t         GetParameter(TCollection *c, const char *par, Double_t &value);
 
    ClassDefOverride(TProof,0)  //PROOF control class
-};
+} R__DEPRECATED(6,38, "Legacy interactive distributed processing engine. Use RDataFrame instead.");
 
 // Global object with default PROOF session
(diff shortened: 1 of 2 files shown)


proof/proof/inc/TProof.h
```diff
@@ -1071,7 +1071,7 @@ friend class TXProofServ;     // to access EUrgent
    static Int_t         GetParameter(TCollection *c, const char *par, Double_t &value);
 
    ClassDefOverride(TProof,0)  //PROOF control class
-};
+} R__DEPRECATED(6,38, "Legacy interactive distributed processing engine. Use RDataFrame and DistRDF instead.");
```
vepadulano, line 1074:
```suggestion
} R__DEPRECATED(6,38, "Legacy interactive distributed processing engine. Use RDataFrame instead.");
```

`DistRDF` is an internal/colloquial name. I would not use it in a public-facing warning. Alternatively, we could write `RDataFrame (local or distributed) instead`.

dpiparo: makes sense. Thanks for the comment.

PR #18310, opened 2025-04-08
[ntuple] Use std::variant in RNTupleFileWriter for fileSimple/Proper

silverweed:
# This Pull request:
fixes an old TODO. Using a std::variant for fileProper/ fileSimple should help keeping the code more maintainable especially if we are going to add more variants to it (like a RFile-based backend).

## Checklist:

- [x] tested changes locally


tree/ntuple/v7/src/RMiniFile.cxx
```diff
@@ -1198,17 +1200,18 @@ ROOT::Experimental::Internal::RNTupleFileWriter::Append(std::string_view ntupleN
    assert(file->IsBinary());
 
    auto writer = std::unique_ptr<RNTupleFileWriter>(new RNTupleFileWriter(ntupleName, maxKeySize));
-   writer->fFileProper.fDirectory = &fileOrDirectory;
+   writer->fFile.emplace<RFileProper>(&fileOrDirectory);
```
hahnjo, line 1203: I think I would prefer simplicity and have `RFileProper` also default-constructible:
```suggestion
   auto &fileProper = writer->fFile.emplace<RFileProper>();
   fileProper.fDirectory = &fileOrDirectory;
```

tree/ntuple/v7/inc/ROOT/RMiniFile.hxx
```diff
@@ -123,6 +123,7 @@ private:
       /// argument is actually just a pointer.)
       std::uint64_t ReserveBlobKey(size_t nbytes, size_t len, unsigned char keyBuffer[kBlobKeyLen] = nullptr);
       operator bool() const { return fDirectory; }
+      RFileProper(TDirectory *dir) : fDirectory(dir) {}
```
hahnjo, line 126: ... then this constructor shouldn't be needed

hahnjo (commented):
Thanks for addressing this long-standing TODO!

PR #18311, opened 2025-04-08
[webgui] testing /snap/bin/chromium

linev:
Partially resolves https://root-forum.cern.ch/t/63182/ 
Let use snap installation of chromium on the platform where pre-compiled ROOT version is used

Also add special handling of temporary directory. 
snap version of chromium has limitation that files from `/tmp`
cannot be accessed. Therefore if TMPDIR is not set, try to use directly home directory.

Adjust batch code to use new functionality with temp directory

Delete temporary HTML file as soon as connection is established


couet (approved):

linev:
@hahnjo 

Yes, I applied clang-format and didnot check result.
Thanks for fixing it

Issue #18316, opened 2025-04-08
[ntuple] RValue::GetRef<T> should type-check


hahnjo:
Using `RNTupleView<void>`, it is possible to write the following code:
```c++
#include <ROOT/RNTupleModel.hxx>
#include <ROOT/RNTupleReader.hxx>
#include <ROOT/RNTupleWriter.hxx>
#include <ROOT/RNTupleView.hxx>

#include <iostream>

void ntuple_view_void() {
  {
    auto model = ROOT::RNTupleModel::Create();
    *model->MakeField<float>("px") = 42.0f;
    auto writer = ROOT::RNTupleWriter::Recreate(std::move(model), "ntpl", "void.root");
    writer->Fill();
  }

  auto reader = ROOT::RNTupleReader::Open("ntpl", "void.root");
  auto view = reader->GetView<void>("px");
  view(0);
  auto value = view.GetValue().GetRef<double>();
  std::cout << "value = " << value << "\n";
}
```

This should throw when trying to get a `double` reference to a `float` object.

hahnjo:
same for `RValue::GetPtr<T>` obviously

PR #18317, opened 2025-04-08
[ntuple] Remove `GetView<void>` with external address

hahnjo:
This interface is potentially dangerous since the underlying field is constructed based on the on-disk type name, without being able to verify that the passed pointer is compatible.

---

I left `GetView<void>("f")` without external address because in principle we can do type checking there (https://github.com/root-project/root/issues/18316).

### DIFF
--- a/tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
+++ b/tree/ntuple/v7/inc/ROOT/RNTupleReader.hxx
@@ -279,9 +279,21 @@ public:
    }
 
    template <typename T>
+   ROOT::RNTupleView<T> GetView(std::string_view fieldName, std::shared_ptr<T> objPtr, std::string_view typeName)
+   {
+      return GetView<T>(RetrieveFieldId(fieldName), objPtr, typeName);
+   }
+
+   template <typename T>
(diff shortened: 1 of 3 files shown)


hahnjo:
@amete @Nowakus @Dr15Jones @makortel while discussing https://github.com/root-project/root/pull/18185, we realized that the current interface of `GetView<void>(std::string_view fieldName, void *rawPtr)` is potentially dangerous because the underlying field is created based on the on-disk type name without a possibility to check that the passed pointer is compatible. I realize we are quite late here, but how bad would it be if we required passing the type name, effectively mandating https://github.com/root-project/root/pull/18185?

Nowakus:
Hi,
Once we get the new interface, the one with the typeinfo/typename, we will
not need the 'old' one. That's basically what we wanted to have for schema
evolution support, as discussed a few weeks ago.
Just please keep in mind we also do:

ptr = dsc.view->GetField().CreateObject<void>().release();

and
view->BindRawPtr( ptr );

and this should probably create the object of "our" type (and bind an
object of our type).
That's something I just realized - maybe it needs another discussion...
Cheers, Marcin

Issue #9924, opened 2022-02-17
I/O customization rule not run for a all-transient nested objects in a split branch.


pcanal:
With the patch at the end of this report, the test `roottest-root-io-issue-9899-read_cmssw_class_v3` fails with (partial output):
```
Error in <readfile>: For entries 0 we have result=19200
```
This is due to the fact that there is no branch for `w.fPersValue.fPersLowData.fProblems.fContent.fHolder` because the typeof `fHolder` : `RefVectorMemberPointersHolder` only has one transient member.   Consequently, the line in `TBranchElement`'s `GatherArtificialElements` function:
```
         if (!foundRelatedSplit) {
            continue;
         }
```
which is designed to avoid drilling through data members that are "new" in the current (in-memory) class layout.   `fContent` also has no explicit branch (it is intentionally elided) but `fContent.fID` has a branch.  Since the object (`fContent` is updated during the read, we should be running the I/O rules on its sub-objects ... but we don't

The issue can be senen by applying this patch to `roottest` and running `ctest -R issue-9899`

```
diff --git a/root/io/issue-9899/read_cmssw_class_v3.cxx b/root/io/issue-9899/read_cmssw_class_v3.cxx
index 58b50a93..e0aef7e1 100644
--- a/root/io/issue-9899/read_cmssw_class_v3.cxx
+++ b/root/io/issue-9899/read_cmssw_class_v3.cxx
@@ -129,10 +129,8 @@ struct PersHighData {
    }
    int check(int slide = 1) {
       int res = 0;
-#ifdef FIXED_ISSUE_99xx
       for(auto &d : fPersLowData)
          res += d.check(slide, 0);
-#endif
       for(auto &d : fLowData)
          res += d.check(slide * BIT(6), 10);
       return res;
```

PR #18320, opened 2025-04-08
tree: Fix address for nested split obj with io rule.

pcanal:
In the case a nested split object has a rule applying to it, we need to read its component (the sub-branches) into the 'fOnfileObject' rather than the user in-memory object

The example usage (and skeleton for test) can be seen at https://github.com/jblomer/auto_ptr_demo

### DIFF
--- a/tree/tree/inc/TBranch.h
+++ b/tree/tree/inc/TBranch.h
@@ -176,7 +176,7 @@ protected:
 
    TString  GetRealFileName() const;
 
-   virtual void SetAddressImpl(void *addr, bool /* implied */) { SetAddress(addr); }
+   virtual void SetAddressImpl(void *addr, bool /* implied */, Int_t /* offset */) { SetAddress(addr); }
 
 private:
    Int_t    GetBasketAndFirst(TBasket*& basket, Long64_t& first, TBuffer* user_buffer);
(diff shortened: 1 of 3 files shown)


ferdymercury:
Thanks! Maybe a bit related:
https://github.com/root-project/root/issues/9924
https://its.cern.ch/jira/browse/ROOT-10271
https://its.cern.ch/jira/browse/ROOT-118

jblomer (commented):
In principle looks good to me but in my auto_ptr example I still get
```
free(): double free detected in tcache 2
```

This may be an issue in the example though...

EDIT: with a fixed example and this patch, auto pointer migration works!

jblomer (approved):

jblomer:
I will turn the example into a `roottest` as soon as I can, perhaps before Monday or early next week.

dpiparo:
thanks @jblomer for this contribution to this very useful PR!

pcanal:
> Thanks! Maybe a bit related: ...

@ferdymercury Could you prepare a roottest PR with (failing) test for those?

pcanal:
>  > This may be an issue in the example though...
> EDIT: with a fixed example and this patch, auto pointer migration works!

Indeed. The example did not work without the required change I proposed in a PR (mostly, the resetting of the value in the old auto_ptr after use).

pcanal:
> I am not sure if you want me to put all those files from the zip in roottest PR?

Not really :).   But the sources inside `io-rules-regression` and re-write of the build and execution steps in our own CMakeLIsts' "language".

pcanal:
> Here it is: https://github.com/root-project/roottest/pull/1300 (for https://github.com/root-project/root/issues/9924)

Thanks. Let's if this PR fixes it.

PR #18324, opened 2025-04-09
[TMVA][SOFIE] Fix RModel move constructor and assignment

guitargeek:
The move constructor and assignment were defined manually, leading to the classic oversight that it was not updated when data members and base classes were added.

It is safer to let it get defined by the compiler, since also all data members are trivially movable and there is no manual memory management.

The only other channge that was required to make the default move work was to avoid a `const std::vector` data member of the RModel_Base class.


lmoneta (approved):
LGTM!

vepadulano (commented):
I believe this is a good improvement! But check the compiler warning on Mac ARM
```
explicitly defaulted move assignment operator is implicitly deleted [-Wdefaulted-function-deleted]
```

guitargeek:
Ah yes, I missed another `const` member that should be static

Issue #18322, opened 2025-04-09
[RF] "Definitely lost" reported by Valgrind in RooAbsCollection


TomasDado:
### Explain what you would like to see improved and how.

Running valgrind reports this memory leak:
```
==41074== 5,528 (72 direct, 5,456 indirect) bytes in 1 blocks are definitely lost in loss record 16,046 of 16,872
==41074==    at 0x4848F95: operator new(unsigned long) (vg_replace_malloc.c:487)
==41074==    by 0x5AE6A13: RooAbsCollection::find(char const*) const (RooAbsCollection.cxx:932)
==41074==    by 0x5D71B86: RooWorkspace::import(RooAbsArg const&, RooCmdArg const&, RooCmdArg const&, RooCmdArg const&, RooCmdArg const&, RooCmdArg const&, RooCmdArg const&, RooCmdArg const&, RooCmdArg const&, RooCmdArg const&) (RooWorkspace.cxx:665)
==41074==    by 0x7553342: RooStats::HistFactory::HistoToWorkspaceFactoryFast::MakeCombinedModel(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, std::vector<std::unique_ptr<RooWorkspace, std::default_delete<RooWorkspace> >, std::allocator<std::unique_ptr<RooWorkspace, std::default_delete<RooWorkspace> > > >&) (HistoToWorkspaceFactoryFast.cxx:1565)
==41074==    by 0x4B72C77: TRExFit::PerformWorkspaceCombination(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&) const (TRExFit.cc:5014)
==41074==    by 0x4BF32E3: TRExFit::PrepareMixedDataset(TRExFit::WorkspaceType) (TRExFit.cc:10421)
==41074==    by 0x4C01732: TRExFit::Fit(bool) (TRExFit.cc:4338)
[...]

PR #16202, opened 2024-08-08
TProtoClass: do not complain innards of transients.

pcanal:
In `GetRealData`, do not complain about missing information about data members that are directly or indirectly within a transient member of the top level class.

This solves the underlying problem from https://github.com/root-project/root/pull/15733

PR #15733, opened 2024-06-04
[RF] Avoid using transient `std::unique_ptr`

guitargeek:
Harmless change to circumvent errors like these when building RooFit standalone:

```
Error in <TProtoClass::FindDataMember>: data member with index 0 is not found in class tuple<TreeReadBuffer*,default_delete<TreeReadBuffer> >
Error in <CreateRealData>: Cannot find data member # 0 of class tuple<TreeReadBuffer*,default_delete<TreeReadBuffer> > for parent RooAddPdf!
Error in <TProtoClass::FindDataMember>: data member with index 1 is not found in class tuple<TreeReadBuffer*,default_delete<TreeReadBuffer> >
Error in <CreateRealData>: Cannot find data member # 1 of class tuple<TreeReadBuffer*,default_delete<TreeReadBuffer> > for parent RooAddPdf!
Error in <TProtoClass::FindDataMember>: data member with index 0 is not found in class tuple<TreeReadBuffer*,default_delete<TreeReadBuffer> >
Error in <CreateRealData>: Cannot find data member # 0 of class tuple<TreeReadBuffer*,default_delete<TreeReadBuffer> > for parent RooPolynomial!
```

PR #18327, opened 2025-04-09
[RF] fix memleak in rooabscollection::find

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/18322

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/roofit/roofitcore/src/RooAbsCollection.cxx
+++ b/roofit/roofitcore/src/RooAbsCollection.cxx
@@ -929,7 +929,8 @@ RooAbsArg * RooAbsCollection::find(const char *name) const
 
   if (_hashAssistedFind || _list.size() >= _sizeThresholdForMapSearch) {
     if (!_hashAssistedFind || !_hashAssistedFind->isValid()) {
-      delete _hashAssistedFind; // needed for the case !_hashAssistedFind->isValid(), does not harm for the case of !_hashAssistedFind
+      if (_hashAssistedFind)
+        delete _hashAssistedFind;
       _hashAssistedFind = new HashAssistedFind{_list.begin(), _list.end()};
     }
 
--- a/roofit/roofitcore/src/RooAbsCollection.cxx
+++ b/roofit/roofitcore/src/RooAbsCollection.cxx
@@ -930,7 +930,7 @@ RooAbsArg * RooAbsCollection::find(const char *name) const
   if (_hashAssistedFind || _list.size() >= _sizeThresholdForMapSearch) {
     if (!_hashAssistedFind || !_hashAssistedFind->isValid()) {
       if (_hashAssistedFind)
-        delete _hashAssistedFind;
+         delete _hashAssistedFind;
       _hashAssistedFind = new HashAssistedFind{_list.begin(), _list.end()};
     }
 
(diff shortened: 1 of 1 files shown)


roofit/roofitcore/src/RooAbsCollection.cxx
```diff
@@ -929,6 +929,7 @@ RooAbsArg * RooAbsCollection::find(const char *name) const
 
   if (_hashAssistedFind || _list.size() >= _sizeThresholdForMapSearch) {
     if (!_hashAssistedFind || !_hashAssistedFind->isValid()) {
+      delete _hashAssistedFind; // needed for the case !_hashAssistedFind->isValid(), does not harm for the case of !_hashAssistedFind
```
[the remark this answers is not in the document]
guitargeek: No, we can't. See:
  * https://github.com/root-project/root/pull/15733

guitargeek:
What is `_hashAssistedFind` is a nullptr? Then this code branch will activate, and calling delete on a nullptr is not alllowed. Also, it would be better to set the pointer to `nullptr` after the delete, so there won't be a problem with dereferencing garbage pointers later

ferdymercury:
> What is `_hashAssistedFind` is a nullptr? Then this code branch will activate, and calling delete on a nullptr is not alllowed. Also, it would be better to set the pointer to `nullptr` after the delete, so there won't be a problem with dereferencing garbage pointers later

Delete on nullptr is okish, right?
According to cppreference: If ptr is a null pointer value, no destructors are called, and the deallocation function may or may not be called (it's unspecified), but the default deallocation functions are guaranteed to do nothing when passed a null pointer.

No need to set it to null after delete, right?
That's being done just below. In line 933.

Anyway I will change as suggested, does not harm either.

roofit/roofitcore/src/RooAbsCollection.cxx
ferdymercury, line 932:
```suggestion
      if (_hashAssistedFind)
        delete _hashAssistedFind;
```

roofit/roofitcore/src/RooAbsCollection.cxx
```diff
@@ -929,6 +929,8 @@ RooAbsArg * RooAbsCollection::find(const char *name) const
 
   if (_hashAssistedFind || _list.size() >= _sizeThresholdForMapSearch) {
     if (!_hashAssistedFind || !_hashAssistedFind->isValid()) {
+      if (_hashAssistedFind)
+        delete _hashAssistedFind;
```
ferdymercury, line 933:
```suggestion
         delete _hashAssistedFind;
```

silverweed:
Indeed, calling `delete` on a nullptr is perfectly valid and harmless, so the check is superfluous.
I'll ask here as well if we should take the occasion to change that variable to a unique_ptr though...

guitargeek:
Hi thanks @silverweed, I didn't know that! Sorry Ferdy for the confusion.

Unfortunately we can't migrate to `std::unique_ptr` until this PR is merged:
  * https://github.com/root-project/root/pull/16202

guitargeek (approved):
Thanks you very much for this PR! Sorry for making meaningless comments earlier, I really didn't know calling `delete` on a `nullptr` is okay :)

roofit/roofitcore/src/RooAbsCollection.cxx
```diff
@@ -929,6 +929,8 @@ RooAbsArg * RooAbsCollection::find(const char *name) const
 
   if (_hashAssistedFind || _list.size() >= _sizeThresholdForMapSearch) {
     if (!_hashAssistedFind || !_hashAssistedFind->isValid()) {
+      if (_hashAssistedFind)
+         delete _hashAssistedFind;
```
pcanal, line 933: The if statement is redundant, the exact same if statement is inside the operator delete implementation.

guitargeek: Yes, we have removed that in Tomas' follow up PR

PR #18332, opened 2025-04-09
[tmva][sofie] Restructure emitted code to be differentiable with Clad

guitargeek:
Restructure emitted code to be differentiable with Clad.

### DIFF
--- a/tmva/sofie/inc/TMVA/SOFIE_common.hxx
+++ b/tmva/sofie/inc/TMVA/SOFIE_common.hxx
@@ -616,16 +616,6 @@ void col2im(const Dtype* data_col, const int channels,
   //std::cout << "finishing col2imp" << std::endl;
 }
 
-// Used at the end of infer() to fill the return object.
-template <class T>
-void FillOutput(T const *arr, std::vector<T> &out, std::size_t n)
-{
-   out.resize(n);
-   for (std::size_t i = 0; i < n; ++i) {
-      out[i] = arr[i];
-   }
-}
-
 }  // end namespace UTILITY
 
 namespace BLAS {
--- a/tmva/sofie/inc/TMVA/SOFIE_common.hxx
+++ b/tmva/sofie/inc/TMVA/SOFIE_common.hxx
@@ -628,11 +628,21 @@ void FillOutput(T const *arr, std::vector<T> &out, std::size_t n)
 
 }  // end namespace UTILITY
 
-namespace BLAS{
-extern "C" void sgemm_(const char * transa, const char * transb, const int * m, const int * n, const int * k,
-                       const float * alpha, const float * A, const int * lda, const float * B, const int * ldb,
-                       const float * beta, float * C, const int * ldc);
-}//BLAS
+namespace BLAS {
+
+extern "C" void saxpy_(const int *n, const float *alpha, const float *x, const int *incx, float *y, const int *incy);
+
+extern "C" void scopy_(const int *n, const float *x, const int *incx, float *y, const int *incy);
+
+extern "C" void sgemm_(const char *transa, const char *transb, const int *m, const int *n, const int *k,
+                       const float *alpha, const float *A, const int *lda, const float *B, const int *ldb,
+                       const float *beta, float *C, const int *ldc);
+
+extern "C" void sgemv_(const char *trans, const int *m, const int *n, const float *alpha, const float *A,
+                       const int *lda, const float *X, const int *incx, const float *beta, const float *Y,
+                       const int *incy);
+
+} // namespace BLAS
 
 
 struct GNN_Data {
--- a/tmva/sofie/src/RModel.cxx
+++ b/tmva/sofie/src/RModel.cxx
@@ -542,7 +542,7 @@ std::string GenerateConstantTensorCode(const std::pair<std::string, InitializedT
       else {
          strs << ConvertValuesToString(length, data) << ";\n";
       }
-      strs << "const " << type << " * tensor_" + t.first + " = fTensor_" + t.first + ".data();\n";
+      strs << type << " * tensor_" + t.first + " = fTensor_" + t.first + ".data();\n";
    }
    return strs.str();
 }
@@ -744,13 +744,16 @@ void RModel::GenerateOutput()
    if (!doInferArgs.empty())
       doInferArgs += ",";
    for (std::string const &name : fOutputTensorNames) {
-      fGC += SP + "std::vector<" + typeForOutput(GetTensorType(name)) + " > output_tensor_" + name + ";\n";
-      doInferArgs += " output_tensor_" + name + ",";
+      bool isIntermediate = fIntermediateTensorInfos.count(name) > 0;
+      std::string n = isIntermediate ? std::to_string(ConvertShapeToLength(GetTensorShape(name)))
+                                     : ConvertDynamicShapeToLength(GetDynamicTensorShape(name));
+      fGC += SP + "std::vector<" + typeForOutput(GetTensorType(name)) + " > output_tensor_" + name + "(" + n + ");\n";
+      doInferArgs += " output_tensor_" + name + ".data(),";
    }
    if (!doInferArgs.empty())
       doInferArgs.back() = ' ';
 
-   fGC += SP + "doInfer(" + doInferArgs + ");\n";
+   fGC += SP + "doInfer(this, " + doInferArgs + ");\n";
 
    fGC += SP + "return {";
    for (size_t i = 0; i < fOutputTensorNames.size(); i++) {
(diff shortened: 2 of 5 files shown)


tmva/sofie/inc/TMVA/SOFIE_common.hxx
```diff
@@ -705,6 +705,13 @@ inline GNN_Data Copy(const GNN_Data & data) {
    return out;
 }
 
+inline void Copy(float const *b, float const *e, float *o)
```
vgvassilev, line 708: Does providing a pullback for std::copy not work?

tmva/sofie/test/TestCustomModelsFromONNX.cxx
```diff
@@ -2515,7 +2515,7 @@ TEST(ONNX, Equal){
    });
 
    TMVA_SOFIE_Equal::Session s("Equal_FromONNX.dat");
-   std::vector<bool> output = s.infer(input1.data(),input2.data());
```
vgvassilev, line 2518: Did that fail to differentiate?

tmva/sofie/inc/TMVA/SOFIE_common.hxx
guitargeek: No, I tried a bit but then gave up. This was my approach:
```c++
#include <Math/CladDerivator.h>

namespace std {

void copy_pullback(double const *first, double const *last, double *out_first, double *_d_out, double *_d_first,
                   double *_d_last, double *_d_out_first)
{
   // Implementation doesn't matter yet, it doesn't compile anyway
}

} // namespace std

void fooImpl(double const *x, double *y)
{
   std::copy(x, x + 1, y);
}

void foo(double const *x, double *y)
{
   fooImpl(x, y);
}

double g(double *variables)
{
   double out;
   foo(variables, &out);
   return out * variables[1];
}

void clademo()
{

   // Call clad to generate the gradient of g.
   auto g_grad = clad::gradient(g, "variables");

   // Execute the generated gradient function.
   double variables[]{3., 4.};
   double grad_output[]{0., 0.};
   g_grad.execute(variables, grad_output);
   std::cout << "grad_output[0]: " << grad_output[0] << std::endl;
   std::cout << "grad_output[1]: " << grad_output[1] << std::endl;

   // Dump the generated gradient code to standard output.
   g_grad.dump();
}
```
It segfaults. I think Clad just doesn't play well with the STL algos that take iterators, it's better to avoid this, no?

In any case, supporting this is not crucial for this PR. I was refactoring things to avoid this copy call in the generated code anyway.

Once this PR is functional for our usecase (actually now, but I also want to make the ROOT CI pass again), I'll write up what was not perfect in Clad for this and open issues

tmva/sofie/test/TestCustomModelsFromONNX.cxx
guitargeek: No, I didn't even try to differentiate the models in the test. I'm solely focusing on the SBI usecase that we implement with LHCb. The reason why I changed this is because `std::vector<bool>` is not a good output type parameter. See:
  * https://github.com/root-project/root/pull/18302

tmva/sofie/inc/TMVA/SOFIE_common.hxx
vgvassilev: Ah, I see. Probably worth opening an issue in clad…

guitargeek:
### Proof of concept test for this PR

Take this ONNX file (remove the `.txt` suffix after downloading):

[VRlL_real_500k_evts_model.onnx.txt](https://github.com/user-attachments/files/19854120/VRlL_real_500k_evts_model.onnx.txt)

Here are the scripts to convert the model to C++ and then to differentiate it with Clad:
```C++
// onnx_to_cpp.C

void onnx_to_cpp()
{
   using namespace TMVA::Experimental;
   SOFIE::RModelParser_ONNX parser;
   SOFIE::RModel model = parser.Parse("./VRlL_real_500k_evts_model.onnx");
   model.SetOptimizationLevel(SOFIE::OptimizationLevel::kBasic);
   model.Generate();
   model.PrintRequiredInputTensors();

   model.OutputGenerated("./VRlL_real_500k_evts_model.hxx");
}
```
```C++
// sofie_ad.C

#include "VRlL_real_500k_evts_model.hxx"

#include <Math/CladDerivator.h>

using Sess = TMVA_SOFIE_VRlL_real_500k_evts_model::Session;

// Wrapper functions for Clad
float my_func(Sess const &session, float const *tensor_x, float *tensor_theory_params)
{
   float out = 0.;
   TMVA_SOFIE_VRlL_real_500k_evts_model::doInfer(session, tensor_x, tensor_theory_params, &out);
   return out;
}

float my_func_wrapper(Sess const &session, float const *tensor_x, float *tensor_theory_params)
{
   return my_func(session, tensor_x, tensor_theory_params);
}

void sofie_ad()
{
   // Let's go Clad!
   clad::gradient(my_func_wrapper, "tensor_theory_params");

   // Get a function pointer to the pullback. If you are unsure what the
   // signature is, try to cast the pullback to some function pointer, like
   // static_cast<void (*)(float)>(my_func_pullback) in the interpreter, and
   // the compiler will tell you what the real signature is.
   using Grad_t = void (*)(const Sess &, const float *, float *, float, Sess *, float *);

   // Get the functions from the interpreter (remove semicolor to get the code printed)
   auto grad = reinterpret_cast<Grad_t>(gInterpreter->ProcessLine("my_func_pullback;"));

   std::vector<float> input1{5.0, 2.0, 1.0, -1.0, 1.0};
   std::vector<float> input2{0.0};

   // A trick: pre-allocate session struct both for the forward pass and
   // backward pass, to that no memory allocation of intermediate tensors has
   // to happend in the gradient.
   TMVA_SOFIE_VRlL_real_500k_evts_model::Session s("VRlL_real_500k_evts_model.dat");
   TMVA_SOFIE_VRlL_real_500k_evts_model::Session d_s("VRlL_real_500k_evts_model.dat");

   // Calculate numerical gradient
   auto numDiff = [&](int i) {
      const float eps = 1e-4;
      std::vector<float> p{input2};
      p[i] = input2[i] - eps;
      float funcValDown = my_func(s, input1.data(), p.data());
      p[i] = input2[i] + eps;
      float funcValUp = my_func(s, input1.data(), p.data());
      return (funcValUp - funcValDown) / (2 * eps);
   };

   for (std::size_t i = 0; i < input2.size(); ++i) {
      std::cout << i << ":" << std::endl;
      std::cout << "  numr : " << numDiff(i) << std::endl;
   }

   // Calculate gradient with Clad
   float grad_output[]{0., 0., 0., 0., 0.};
   grad(s, input1.data(), input2.data(), 1.0, &d_s, grad_output);

   std::cout << "  clad : " << grad_output[0] << std::endl;
}
```

Usage with expected output (replace `libblas.so` location with relevant path for your system):
```
root [1] .x onnx_to_cpp.C
Model requires following inputs:
Fully Specified Tensor name: theory_params	type: float	shape: [1]
Fully Specified Tensor name: x	type: float	shape: [5]

root [2] .x sofie_ad.C
0:
  numr : -0.531077
  clad : -0.532437
root [3] .q
```

vgvassilev:
Why did we decide to not pursue this?

guitargeek:
@vgvassilev, sorry that was totally an accident. Maybe I confused it with another PR, or I wanted to close and re-open the PR to run the tests, but apparently I missed the "reopen" button.

guitargeek:
So it works now, and even includes a test to validate the gradient of a fully-connected multi-layer network! I'll organize the changes a bit better before marking this as not a draft.

vgvassilev:
Is there any way we can easily compare the performance and memory footprint against say PyTorch?

guitargeek:
Yes, I'm working on that. So far, the generated gradient is not competitive because it's not optimized. I'll have to throw a few more `pragma clad checkpoint loop` in the generated code, for example.

But these optimizations can be better done in a separate PR. I also need to follow up with a `RooONNXFunction` class for easy use in RooFit. Performance optimizations only comes after that.

tmva/sofie/inc/TMVA/ROperator_Gemm.hxx
```diff
@@ -393,9 +393,12 @@ namespace SOFIE{
              << (fAttrTransB ? "true, " : "false, ")
              << (fAttrTransA ? "true, " : "false, ")
              << n << ", " << m << ", " << k << ", ";
-            out << std::setprecision(std::numeric_limits<float>::max_digits10) << fAttrAlpha << ", tensor_" << fNB;
+            // TODO: the cast to (float *) is not needed here from the C++ language perspective (the arguments to
+            // Gemm_Call are const already), but Clad bug https://github.com/vgvassilev/clad/issues/1721 is requiring
```
vgvassilev, line 397: We are in a bit chicken and egg problem here. That seems resolved in the master of clad -- do we release before merging and removing these workarounds?

guitargeek: Ah yes you are right, I forgot about that! Okay I'll just test it locally then with Clad `master` and tell you how it goes. Releasing before merging this PR is better then.

guitargeek: I confirm this PR works with Clad `master` even without the workaround!

guitargeek: We discussed in private that the next Clad release will only happen later when the Hessian problem is fixed, so it's probably better to merge this PR as soon as possible with the workaround, so it doesn't collide with other SOFIE developments, according to @lmoneta

tmva/sofie/src/RModel.cxx
```diff
@@ -1194,32 +1345,47 @@ void RModel::GenerateSessionCode()
...
 
-   for (std::string const &name : fOutputTensorNames) {
-      // need to check is size is the same (don't want to return a vector with
-      // larger size) in that case better to copy
-      bool isIntermediate = fIntermediateTensorInfos.count(name) > 0;
-      std::string n = isIntermediate ? std::to_string(ConvertShapeToLength(GetTensorShape(name)))
-                                     : ConvertDimShapeToLength(GetDimTensorShape(name));
-      fGC += SP + "FillOutput(tensor_" + name + ", output_tensor_" + name + ", " + n + ");\n";
+   if (fUseSession && !fIsGNNComponent) {
+      // Collect all "tensor_*" data members that are not inpt or output tensors
```
lmoneta: inpt -> input

tmva/sofie/test/TestCustomModelsFromONNX.cxx
```diff
@@ -113,6 +113,86 @@ TEST(ONNX, Linear16)
 
    ASSERT_INCLUDE_AND_RUN(std::vector<float>, "Linear_16", input);
 
+   // Use this test to also test differentiating a fully-connected neural network with Clad.
```
lmoneta: Would not be better maybe to have an independent test  for Clad? Or is it easier to keep it inside TestCustomModelsFromONNX?

guitargeek: No, you're right it's better to move it for now. Maybe I'll revisit it later when all the models in `TestCustomModelsFromONNX` can also be tested with Clad systematically.

guitargeek:
CI failure is unrelated. The SOFIE Keras parser tests fail in all PRs on `alma8` today, almost certainly because of the update from NumPy 2.4.3 to 2.4.4

Issue #18329, opened 2025-04-09
[core] "Definitely lost" from Valgrind in TUnixSystem::UnixOpendir


TomasDado:
### Explain what you would like to see improved and how.

Running with
```
valgrind --leak-check=full          --show-leak-kinds=all          --track-origins=yes          --verbose          --log-file=valgrind-out.txt --suppressions=$ROOTSYS/etc/valgrind-root.supp
```

Gives:
```
==18986== 32,816 bytes in 1 blocks are definitely lost in loss record 15,685 of 15,939
==18986==    at 0x484880F: malloc (vg_replace_malloc.c:446)
==18986==    by 0x64F948B: __alloc_dir (opendir.c:115)
==18986==    by 0x64F948B: opendir_tail (opendir.c:63)
==18986==    by 0x64F948B: opendir (opendir.c:86)
==18986==    by 0x500B6B6: TUnixSystem::UnixOpendir(char const*) (TUnixSystem.cxx:3996)
==18986==    by 0x756085F: RooStats::HistFactory::MakeModelAndMeasurementFast(RooStats::HistFactory::Measurement&, RooStats::HistFactory::HistoToWorkspaceFactoryFast::Configuration const&) (MakeModelAndMeasurementsFast.cxx:122)
==18986==    by 0x4BCA202: TRExFit::ToRooStats(bool, bool) const (TRExFit.cc:3519)
==18986==    by 0x114EEB: FitExample(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (trex-fitter.cc:448)
==18986==    by 0x10FE1E: main (trex-fitter.cc:696)
```

### ROOT version

```
ROOT Version: 6.35.01
Built for linuxx8664gcc on Apr 09 2025, 09:00:09
From remotes/ferdymercury/findhash2@c4cace9023c
```

### Installation method

build from source

### Operating system

Ubuntu 24.10

ferdymercury:
Thanks for the report! Could you try rebuilding with https://github.com/root-project/root/pull/18331 to see if it solves the issue?

ferdymercury:
> Which I assume is a false positive?

I'd say it's a true error. Please on a separate one ;) Thanks.

TomasDado:
[...]

PR #18333, opened 2025-04-09
[io] fix memleak in generateinfoforpair

ferdymercury:
# This Pull request:

## Changes or fixes:

Issue reported in https://github.com/root-project/root/issues/18329#issuecomment-2789064824

## Checklist:

- [ ] tested changes locally


core/meta/src/TClass.cxx
```diff
@@ -3261,8 +3261,11 @@ TClass *TClass::GetClass(const char *name, Bool_t load, Bool_t silent, size_t hi
          auto pairinfo = TVirtualStreamerInfo::Factory()->GenerateInfoForPair(normalizedName, silent, hint_pair_offset, hint_pair_size);
          // Fall-through to allow TClass to be created when known by the interpreter
          // This is used in the case where TStreamerInfo can not handle them.
-         if (pairinfo)
-            return pairinfo->GetClass();
+         if (pairinfo) {
+            auto pairclass = pairinfo->GetClass();
+            delete pairinfo;
```
pcanal, line 3266: That is definitively incorrect.  In most case the StreamerInfo will be adopted by the TClass (possibly generated as a side effect of the call to 3265)

PR #18344, opened 2025-04-10
[cmake] abort when daos is requested and not found

silverweed:
...rather than silently disabling it.

EDIT: making this a draft because I noticed we already have a `fail-on-missing`. I argue it should then be the default option.

## Checklist:

- [x] tested changes locally


guitargeek:
I think this was a good change. The `daos` feature is so highly specific that the people who set `daos=ON` are the same experts who always do `fail-on-missing=ON` anyway :slightly_smiling_face:

PR #18345, opened 2025-04-10
[ntuple] Move RPageStorage classes out of Experimental

silverweed:
# This Pull request:
moves `RPage(Sink|Source)` and related classes out of Experimental.

Note: I split this into multiple commits, but please let me know if it would be better off as a single one.

## Checklist:

- [x] tested changes locally


hahnjo (changes requested):
needs an update in `roottest`

hahnjo (approved):
LGTM, question doesn't need to be addressed in this PR (we probably want to review `using` in all tests at a later point)

PR #18347, opened 2025-04-10
[core] rootcling: avoid UB if ROOTSYS not defined

ferdymercury:
# This Pull request:

## Changes or fixes:

std::string(nullptr) is ill-defined. Found out by clang-tidy

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/core/rootcling_stage1/src/rootcling_stage1.cxx
+++ b/core/rootcling_stage1/src/rootcling_stage1.cxx
@@ -24,12 +24,18 @@ static void (*dlsymaddr)() = &usedToIdentifyRootClingByDlSym;
 ROOT::Internal::RootCling::TROOTSYSSetter gROOTSYSSetter;
 
 static const char *GetIncludeDir() {
-   static std::string incdir = std::string(getenv("ROOTSYS")) + "/include";
+   auto renv = getenv("ROOTSYS");
+   if (!renv)
+      return "";
+   static std::string incdir = std::string(renv) + "/include";
    return incdir.c_str();
 }
 
 static const char *GetEtcDir() {
-   static std::string etcdir = std::string(getenv("ROOTSYS")) + "/etc";
+   auto renv = getenv("ROOTSYS");
+   if (!renv)
+      return "";
+   static std::string etcdir = std::string(renv) + "/etc";
    return etcdir.c_str();
 }
 
--- a/core/rootcling_stage1/src/rootcling_stage1.cxx
+++ b/core/rootcling_stage1/src/rootcling_stage1.cxx
@@ -26,7 +26,7 @@ ROOT::Internal::RootCling::TROOTSYSSetter gROOTSYSSetter;
 static const char *GetIncludeDir() {
    auto renv = getenv("ROOTSYS");
    if (!renv)
-      return "";
+      return nullptr;
    static std::string incdir = std::string(renv) + "/include";
    return incdir.c_str();
 }
@@ -34,7 +34,7 @@ static const char *GetIncludeDir() {
 static const char *GetEtcDir() {
    auto renv = getenv("ROOTSYS");
    if (!renv)
-      return "";
+      return nullptr;
    static std::string etcdir = std::string(renv) + "/etc";
    return etcdir.c_str();
 }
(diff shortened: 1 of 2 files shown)


core/rootcling_stage1/src/rootcling_stage1.cxx
```diff
@@ -24,12 +24,18 @@ static void (*dlsymaddr)() = &usedToIdentifyRootClingByDlSym;
 ROOT::Internal::RootCling::TROOTSYSSetter gROOTSYSSetter;
 
 static const char *GetIncludeDir() {
-   static std::string incdir = std::string(getenv("ROOTSYS")) + "/include";
+   auto renv = getenv("ROOTSYS");
+   if (!renv)
+      return "";
```
guitargeek, line 29: What made you choose `return ""` instead of `return nullptr` here? Not saying that one is more correct than the other, both are better than UB. But depending on how `GetIncludeDir()` is used, one might be better than the other.

Maybe @pcanal has a preference here? I would have gone with `nullptr`, because otherwise you might break code like `if(!GetIncludeDir()) { throw std::runtime_error("include dir not set");  }`.

ferdymercury: Good point. I just thought it was better not to propagate the issue on the caller-side.
If someone was calling std::string myIncludeDir = GetIncludeDir(...); then he would have to deal with UB again.
This way, he can just check if myIncludeDir.empty()
rather than:
auto iDir = GetIncludeDir();
if(iDir)
std::string myIncludeDir(iDir);

but yeah, it's matter of taste. I am happy either way.

guitargeek: Ok I was taking a look in how this function is used, in only `rootcling_impl.cxx` actually.

If we patch that file to avoid adding using empty paths, we need to patch it like this (assuming `nullptr` returns):
```diff
@@ -4218,8 +4218,10 @@ int RootClingMain(int argc,
       clingArgs.push_back(FullWDiag);
    }

-   std::string includeDir = llvm::sys::path::convert_to_slash(gDriverConfig->fTROOT__GetIncludeDir());
-   clingArgs.push_back(std::string("-I") + includeDir);
+   const char *includeDir = gDriverConfig->fTROOT__GetIncludeDir();
+   if (includeDir) {
+      clingArgs.push_back("-I" + llvm::sys::path::convert_to_slash(includeDir));
+   }

    std::vector<std::string> pcmArgs;
    for (size_t parg = 0, n = clingArgs.size(); parg < n; ++parg) {
@@ -4300,9 +4302,9 @@ int RootClingMain(int argc,
       for (const std::string &modulemap : gOptModuleMapFiles)
          clingArgsInterpreter.push_back("-fmodule-map-file=" + modulemap);

-      clingArgsInterpreter.push_back("-fmodule-map-file=" +
-                                     std::string(gDriverConfig->fTROOT__GetIncludeDir()) +
-                                     "/ROOT.modulemap");
+      if (includeDir) {
+         clingArgsInterpreter.push_back("-fmodule-map-file=" + std::string(includeDir) + "/ROOT.modulemap");
+      }
       std::string ModuleMapCWD = ROOT::FoundationUtils::GetCurrentDir() + "/module.modulemap";
       if (llvm::sys::fs::exists(ModuleMapCWD))
          clingArgsInterpreter.push_back("-fmodule-map-file=" + ModuleMapCWD);
```
Maybe we can do this in this PR too? If you do the `return ""` variant that's also fine, but then you need to use `strlen()` or something to check the length of the string. But I think `return nullptr` is simpler here.

Whatever you choose, can we also improve the implementation in `rootcling_stage1.cxx` maybe along the lines I suggested?

ferdymercury: Thanks for the review! Sure, changes applied.
Should a Warning or throw be also issued ?
Side note: the implementation logic of GetRootsys in FoundationUtils.cxx is slightly different, with a fallback value.

guitargeek (approved):
Thanks a lot! You missed a `convert_to_slash` for Windows, but I fixed that up.

ferdymercury:
> Thanks a lot! You missed a `convert_to_slash` for Windows, but I fixed that up.

Thanks for finding out! My bad, I should have copy-pasted from your post instead of eye-pasting xD

PR #18348, opened 2025-04-10
[tmva][sofie] Add new operators with feedback from 1st NGT Hackathon

sanjibansg:
This PR adds new operators to SOFIE based on the feedback we received from the 1st NGT Hackathon. Following are the missing operators that will be added:

- CastLike
- Round
- Not


dpiparo:
this is a draft, of course, but a comment about the commit messages: https://github.com/root-project/root/blob/master/CONTRIBUTING.md - please follow the guidelines. I propose to start with "[sofie]" when it's about SOFIE

lmoneta (changes requested):
Thanks Sanjiban for the PR. 
I see that the NOT operator is implemented as Neg, but it is not exactly the same. The NOT acts only on bool types. Currently Neg supports only Float. We should change this and have support for Boolean in ParseNeg in case it is a NOT

guitargeek:
Hi @sanjibansg, can you please resolve the merge conflicts?

PR #18349, opened 2025-04-10
[TMVA][SOFIE] Implement wrapper for `BLAS::sgemm_` call

guitargeek:
Clad can't do reverse mode AD if you use functions that take parameters that are at the same time an input and an output, like the matrix C in `sgemm_`.

This can be avoided by creating a wrapper where the input and output arrays are separate, and then there is an internal copy.

Having a wrapper like this improves the SOFIE code anyway, because the copying code is now written out by SOFIE before the `sgemm_` call. So by having a wrapper, one also reduces the verbosity of the generated code.


sanjibansg (approved):
LGTM! Thanks for the implementation Jonas!

PR #18326, opened 2025-04-09
fixes for cmake 4.0.0

couet:
cmake 4.0.0 (the official current cmake version) needs some fixes

Issue #18312, opened 2025-04-08
Problem with cmake 4.0.0


couet:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

On macOS (but it may well be the same on other platforms).  with cmake version 4.0.0 we get get a problem with TBB. with older version of cmake it works fine.

### Reproducer

```
$ git clone https://github.com/root-project/root.git
$ mkdir build
$ cd build
$ cmake ../root
$ make
[  0%] Creating directories for 'FREETYPE'
[  0%] Performing download step (verify and extract) for 'FREETYPE'
-- FREETYPE download command succeeded.  See also /Users/sftnight/couet/build/FREETYPE-prefix/src/FREETYPE-stamp/FREETYPE-download-*.log
[  0%] No update step for 'FREETYPE'
[  0%] No patch step for 'FREETYPE'
[  0%] Performing configure step for 'FREETYPE'
-- FREETYPE configure command succeeded.  See also /Users/sftnight/couet/build/FREETYPE-prefix/src/FREETYPE-stamp/FREETYPE-configure-*.log
[  0%] Performing build step for 'FREETYPE'
-- FREETYPE build command succeeded.  See also /Users/sftnight/couet/build/FREETYPE-prefix/src/FREETYPE-stamp/FREETYPE-build-*.log
[  0%] No install step for 'FREETYPE'
[  0%] Completed 'FREETYPE'
[  0%] Built target FREETYPE
[  0%] Creating directories for 'AFTERIMAGE'
[  0%] Performing download step for 'AFTERIMAGE'
-- AFTERIMAGE download command succeeded.  See also /Users/sftnight/couet/build/AFTERIMAGE-prefix/src/AFTERIMAGE-stamp/AFTERIMAGE-download-*.log
[  0%] No update step for 'AFTERIMAGE'
[  0%] No patch step for 'AFTERIMAGE'
[  0%] Performing configure step for 'AFTERIMAGE'
-- AFTERIMAGE configure command succeeded.  See also /Users/sftnight/couet/build/AFTERIMAGE-prefix/src/AFTERIMAGE-stamp/AFTERIMAGE-configure-*.log
[  0%] Performing build step for 'AFTERIMAGE'
-- AFTERIMAGE build command succeeded.  See also /Users/sftnight/couet/build/AFTERIMAGE-prefix/src/AFTERIMAGE-stamp/AFTERIMAGE-build-*.log
[  0%] Performing install step for 'AFTERIMAGE'
[...]

PR #18350, opened 2025-04-10
[BP 6.34][cmake] fixes for cmake version 4.0.0

couet:
cmake 4.0.0 (the official current cmake version) needs some fixes for DAVIX VDT etc ..(see https://github.com/root-project/root/issues/18312)

BP of https://github.com/root-project/root/pull/18326


pcanal:
It appears that VC also need update.

couet:
the backport should be done form here: https://github.com/root-project/root/pull/18533

PR #18352, opened 2025-04-10
[BP 6.32][cmake] fixes for cmake version 4.0.0

couet:
cmake 4.0.0 (the official current cmake version) needs some fixes for DAVIX VDT etc ..(see https://github.com/root-project/root/issues/18312)

BP of https://github.com/root-project/root/pull/18326


couet:
The backport should be done from here: https://github.com/root-project/root/pull/18533

PR #18353, opened 2025-04-10
[core] AddBefore/After/At option variant, as with Add/First/Last v2

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-9288

New version addressing this comment from pcanal: https://github.com/root-project/root/pull/17846#discussion_r2031731219

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/cont/inc/TList.h
+++ b/core/cont/inc/TList.h
@@ -85,10 +85,15 @@ public:
    void              AddLast(TObject *obj) override;
    virtual void      AddLast(TObject *obj, Option_t *opt);
    void              AddAt(TObject *obj, Int_t idx) override;
+   virtual void      AddAt(TObject *obj, Int_t idx, Option_t *opt);
    void              AddAfter(const TObject *after, TObject *obj) override;
    virtual void      AddAfter(TObjLink *after, TObject *obj);
+   virtual void      AddAfter(const TObject *after, TObject *obj, Option_t *opt);
+   virtual void      AddAfter(TObjLink *after, TObject *obj, Option_t *opt);
    void              AddBefore(const TObject *before, TObject *obj) override;
    virtual void      AddBefore(TObjLink *before, TObject *obj);
+   virtual void      AddBefore(const TObject *before, TObject *obj, Option_t *opt);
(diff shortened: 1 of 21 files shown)


core/cont/inc/TList.h
```diff
@@ -85,10 +85,15 @@ friend  class TListIter;
    void              AddLast(TObject *obj) override;
    virtual void      AddLast(TObject *obj, Option_t *opt);
    void              AddAt(TObject *obj, Int_t idx) override;
+   virtual void      AddAt(TObject *obj, Int_t idx, Option_t *opt);
```
pcanal, line 88: Rather than declaration this `virtual` we need to also update to `TSeqCollection.h`

ferdymercury: Thanks. I was just copy pasting the structure from lines 85-86. AddLast overrides from TSeqCollection, AddLast with option is virtual. Otherwise it's not consistent?

ferdymercury: And if we also moved AddLast and AddFirst to TSeqCollection to be consistent, that would break user-scripts deriving from that classes since they would become purely virtual (if one wants to keep consistent style).

ferdymercury: And if we made them non-purely-virtual, then users might start getting warnings such as warning: ‘virtual void ::AddAt(TObject*, Int_t, Option_t*)’ was hidden ?

pcanal: > Otherwise it's not consistent?

Right.  I missed that.

> start getting warnings such as warning: ‘virtual void ::AddAt(TObject*, Int_t, Option_t*)’ was hidden ?

To extent that is better than use that:
* Inherits from TList
* Does not write/override `AddAt(TObject*, Int_t, Option_t*`
* Somehow starts using it ...
* It would 'silently' fail (depending what the overload does - but for example it would/was the case for `THashList`)

ferdymercury: v3 version here: https://github.com/root-project/root/pull/18389

PR #18355, opened 2025-04-10
[tmva][sofie] Add optimization modes in SOFIE

sanjibansg:
This PR adds two separate optimization modes in SOFIE so as to control the level  of optimization the user prefers. As of now, we have `OptimizationLevel:kBasic` and  `OptimizationLevel:kExtended`. By default, ` `OptimizationLevel:kExtended` is enabled which can be changed by:

```
using namespace TMVA::Experimental;
SOFIE::RModelParser_ONNX parser;
SOFIE::RModel model = parser.Parse("model.onnx");
model.SetOptimizationLevel(SOFIE::OptimizationLevel::kBasic);
model.Generate();
```

Optimization levels with their corresponding flags:

- kBasic: Operator fusion (for example: fusing ReLU with a preceeding GEMM)
- kExtended: Reuse memory for intermediate tensors

### DIFF
--- a/tmva/sofie/inc/TMVA/RModel.hxx
+++ b/tmva/sofie/inc/TMVA/RModel.hxx
@@ -18,6 +18,8 @@ private:
    int fBatchSize = -1;
    long fReadPos = 0;  // reading file position
 
+   OptimizationLevel fOptimizationLevel = OptimizationLevel::kExtended;
+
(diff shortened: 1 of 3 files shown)


guitargeek (approved):
Thank you very much for this PR!

I have validated that this works now for our use case of doing reverse mode AD with Clad on the emitted code.

Maybe consider squashing the commits on merge, since the first commit is not a state where the tests would pass and the second commit is a fixup to the first commit (let's try to make the tests pass at every commit).

PR #18357, opened 2025-04-10
[ntuple] support rules with source class != dest class

jblomer:
Including class rename rules.

### DIFF
--- a/tree/ntuple/v7/src/RFieldMeta.cxx
+++ b/tree/ntuple/v7/src/RFieldMeta.cxx
@@ -26,6 +26,7 @@
 #include <TBaseClass.h>
 #include <TBufferFile.h>
 #include <TClass.h>
+#include <TClassEdit.h>
 #include <TDataMember.h>
 #include <TEnum.h>
 #include <TObject.h>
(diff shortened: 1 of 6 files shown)


tree/ntuple/v7/src/RFieldMeta.cxx
```diff
@@ -394,19 +405,20 @@ void ROOT::RClassField::BeforeConnectPageSource(ROOT::Experimental::Internal::RP
...
-
       for (auto linkId : fieldDesc.GetLinkIds()) {
          const auto &subFieldDesc = desc.GetFieldDescriptor(linkId);
          regularSubfields.insert(subFieldDesc.GetFieldName());
       }
 
       rules = FindRules(&fieldDesc);
+
+      // If the field's type name is not the on-disk name but we found a rule, we know it is valid to read
+      // on-disk data because we found the rule according to the on-disk (source) type name and version/checsksum.
```
silverweed, line 416:
```suggestion
      // on-disk data because we found the rule according to the on-disk (source) type name and version/checksum.
```

hahnjo (approved):
LGTM, see Giacomo's comment about a typo.

Note that, if I remember correctly, it is also possible to write entire class rename rules with only `sourceClass` and `targetClass`, but no `source` and `target`. I believe then all members are subject to automatic evolution, but should be checked. In a first implementation, we could require explicit copying of all members by the rule.

pcanal:
> Note that, if I remember correctly, it is also possible to write entire class rename rules with only sourceClass and targetClass, but no source and target.

That is accurate and nicked named 'renaming rule'

> I believe then all members are subject to automatic evolution, but should be checked. In a first implementation, we could require explicit copying of all members by the rule.

This would be sub-optimal as this would mean that for 'renamed' classes (using that type of rule), individual members/column would not be readable on their own; you would have to read the whole object 'all' the time.

PR #18358, opened 2025-04-10
Speedup retrieval of branch names

vepadulano:
When the input `TTree` has a large dataset schema `O(10k)` and when the branch types are non trivial (i.e. many of the branches store some kind of nested data structure with possibly multiple levels of sub-branches representing the data members), retrieving the full list of available branches in the tree may become a costly operation. Anecdotally, a `TTree` used by an ATLAS analysis with >30k total leaves results in a retrieval time of around 2 minutes on my machine. With the changes in this PR, this time goes down **to around 6 seconds**. Main two contributors:
* a new transient map data member in `TTree`. When the `TTree` is being read, it already needs to traverse all its branch hierarchy. Take this opportunity to fill a map that connects the string with the full name of the leaf to the `TBranch` pointer. Then use this map when retrieving the branch.
* A fix in the logic in `RDF` that retrieves the full list of branches, such that in cases where the parent branch name is overlapping with the name of one of its sub branches, the concatenated string is first deduplicated before being passed to `TTree::GetBranch.` This ensures that a proper match in the aforementioned map can be found.

### DIFF
--- a/tree/dataframe/src/RLoopManager.cxx
+++ b/tree/dataframe/src/RLoopManager.cxx
@@ -133,10 +133,29 @@ void InsertBranchName(std::set<std::string> &bNamesReg, ColumnNames_t &bNames, c
 void ExploreBranch(TTree &t, std::set<std::string> &bNamesReg, ColumnNames_t &bNames, TBranch *b,
                           std::string prefix, std::string &friendName, bool allowDuplicates)
 {
+   // We want to avoid situations of overlap between the prefix and the
+   // sub-branch name that might happen when the branch is composite, e.g.
+   // prefix=reco_Wdecay2_from_tbar_4vect_NOSYS.fCoordinates.
+   // subBranchName=fCoordinates.fPt
+   // which would lead to a repetition of fCoordinates in the output branch name
+   // Boundary to search for the token before the last dot
+   auto prefixEndingDot = std::string::npos;
+   if (!prefix.empty() && prefix.back() == '.')
+      prefixEndingDot = prefix.size() - 2;
+   std::string lastPrefixToken{};
+   if (auto prefixLastRealDot = prefix.find_last_of('.', prefixEndingDot); prefixLastRealDot != std::string::npos)
+      lastPrefixToken = prefix.substr(prefixLastRealDot + 1, prefixEndingDot - prefixLastRealDot);
+
    for (auto sb : *b->GetListOfBranches()) {
       TBranch *subBranch = static_cast<TBranch *>(sb);
       auto subBranchName = std::string(subBranch->GetName());
       auto fullName = prefix + subBranchName;
+
+      if (auto subNameFirstDot = subBranchName.find_first_of('.'); subNameFirstDot != std::string::npos) {
+         // Concatenate the prefix to the sub-branch name without overlaps
+         if (!lastPrefixToken.empty() && lastPrefixToken == subBranchName.substr(0, subNameFirstDot))
+            fullName = prefix + subBranchName.substr(subNameFirstDot + 1);
+      }
 
       std::string newPrefix;
       if (!prefix.empty())
--- a/tree/tree/inc/TTree.h
+++ b/tree/tree/inc/TTree.h
@@ -71,10 +71,15 @@ class TVirtualIndex;
 class TBranchRef;
 class TBasket;
 class TStreamerInfo;
+class TTree;
 class TTreeCache;
 class TTreeCloner;
 class TFileMergeInfo;
 class TVirtualPerfStats;
+
+namespace ROOT::Internal::TreeUtils {
+void TBranch__SetTree(TTree *tree, TObjArray &branches);
+}
 
 class TTree : public TNamed, public TAttLine, public TAttFill, public TAttMarker {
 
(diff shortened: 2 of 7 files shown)


tree/dataframe/src/RLoopManager.cxx
```diff
@@ -139,7 +141,16 @@ void ExploreBranch(TTree &t, std::set<std::string> &bNamesReg, ColumnNames_t &bN
       TBranch *subBranch = static_cast<TBranch *>(sb);
       auto subBranchName = std::string(subBranch->GetName());
       auto fullName = prefix + subBranchName;
-
+      auto prefixTokens = ROOT::Split(prefix, ".", true /*skipEmpty*/);
```
silverweed, line 144: This may be done once outside of the `for`

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -139,7 +141,16 @@ void ExploreBranch(TTree &t, std::set<std::string> &bNamesReg, ColumnNames_t &bN
...
+      auto subBranchNameTokens = ROOT::Split(subBranchName, ".", true /*skipEmpty*/);
+      if (!prefixTokens.empty() && !subBranchNameTokens.empty() && prefixTokens.back() == subBranchNameTokens.front()) {
+         fullName = std::string{};
+         for (const auto &s : prefixTokens)
+            fullName += s + ".";
+         for (decltype(subBranchNameTokens.size()) i = 1; i < subBranchNameTokens.size(); i++)
```
silverweed, line 150: If I understand this code correctly you don't really need to Split the strings; can't you instead find the last occurrence of `.` in `prefix`, the first occurrence of `.` in `subBranchName` and then, after matching the substrings, do
```
fullName = prefix + subBranchName.substr(firstDotIdx);
```
?

TomasDado:
Just to add that with these changes I see about a factor of 5 improvement in the runnig speed for the initialisation part of an RDF setup with a file with 30k branches, thanks!

tree/tree/src/TTree.cxx
```diff
@@ -5279,6 +5281,9 @@ TBranch *R__GetBranch(const TObjArray &branches, const char *name)
       if (!strcmp(b->GetName(), name)) {
          return b;
       }
+      const std::string prefix{b->GetName()};
+      if (prefix.compare(0, prefix.size(), name, prefix.size()) != 0)
+         continue;
```
pcanal, line 5286: This is seem odd.  Which cases does it support?  Does it work correctly if the top level branch is elided from the name search or the branch name (i.e. 'no trailing dot' case)?

tree/tree/inc/TTree.h
```diff
@@ -226,6 +229,8 @@ class TTree : public TNamed, public TAttLine, public TAttFill, public TAttMarker
    };
 
 public:
+   void InsertNameAndBranch(std::pair<std::string, TBranch *> &&kv) { fNamesToBranches.insert(kv); }
```
pcanal, line 232: This probably should not be public.  We may need to have `TBranch__SetTree` become a friend of TTree.
```suggestion
   void RegisterBranchFullName(std::pair<std::string, TBranch *> &&kv) { fNamesToBranches.insert(kv); }
```

tree/tree/src/TTree.cxx
vepadulano: It was a smaller optimization I was playing with before the unordered_map. It probably won't be necessary so I think I will remove it.

tree/tree/inc/TTree.h
vepadulano: Done in latest iteration

tree/tree/src/TTree.cxx
pcanal: Should we update the summary (and git log) with the latest results? :)

pcanal: The git log was already updated.  Thanks.  I updated the summary.

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -133,11 +133,30 @@ void InsertBranchName(std::set<std::string> &bNamesReg, ColumnNames_t &bNames, c
 void ExploreBranch(TTree &t, std::set<std::string> &bNamesReg, ColumnNames_t &bNames, TBranch *b,
                           std::string prefix, std::string &friendName, bool allowDuplicates)
 {
+   // We want to avoid situations of overlap between the prefix and the
```
pcanal, line 136: Does the algorithm handle properly the case where (for better or worse), the valid branch name is something like `toplevel.fInner.fInner.fPt` (The top level class has a data member named `fInner` which is of a type that has a data member also named `fInner`)?  

Related but different:  The case where the user did not put the trailing dot and thus the toplevel name is `toplevel`, the branch name is `fInner.fInner.fPt` and the full branch name is stil `toplevel.fInner.fInner.fPt`.

vepadulano: Just to clarify, this part of the code is never run based on user input. This is part of the routine that just populates a list of strings of all the valid branch names in the TTree, i.e. based on on-disk information alone. Thus, I believe the second point is not relevant here.

Regarding the first point, no, the algorithm now does not handle that properly. Unfortunately, I don't think there is a way to make the distinction. Crucially, without the current logic in the algorithm, the changes in TTree have no effect at all, i.e. we still have very poor performance in retrieving the full list of branch names. I need to create a reproducer first, then see what can be done on the TTree side to expose enough information to distinguish the two cases.

vepadulano: After investigation, I can actually confirm that the logic is still working for the case of the nested data member with the same name. The reason is, it was faulty before already. I have created a test that is added to the latest commit. Before the logic is applied, the `fullName` variable is actually of the form `"toplevel.fInner.fInner.fInner.fPt"`, note the 3 repetitions of `fInner` instead of 2. So the logic removes the extra repetition. The real question is why TTree happily digests what looks to me like a wrong leaf name, but that is out of scope for this PR I believe.

pcanal (approved):
LGTM.  Thanks.

PR #18364, opened 2025-04-11
[TMVA][SOFIE] Implement custom pullback for Gemm operator and corresponding test

GoodingJamie:
This PR introduces the pullback for the Gemm operator (`Gemm_Call `, introduced in #18349), such that this can be differentiated by Clad. A test is added to ensure that a function incorporating `Gemm_Call` is differentiated correctly.

## Checklist:

- [x] tested changes locally
 
Cc @guitargeek

### DIFF
--- a/tmva/sofie/test/TestGemmDerivative.cxx
+++ b/tmva/sofie/test/TestGemmDerivative.cxx
@@ -1,0 +1,132 @@
+#include <TInterpreter.h>
+#include <TInterpreterValue.h>
+
+#include "gtest/gtest.h"
+
+#include <stdexcept>
+
+class GemmTest : public testing::TestWithParam<std::tuple<int, int, int>> {
+public:
+   void SetUp() override
+   {
+      m = std::get<0>(GetParam());
+      n = std::get<1>(GetParam());
+      k = std::get<2>(GetParam());
+   }
+
+protected:
+   int m = 0;
+   int n = 0;
+   int k = 0;
+};
+
+void validator(float *variables, int m, int n, int k, float *target)
+{
+   // Offset iteration by number of parameters before matA
+   int offset = 2;
+
+   // Check derivative with respect to alpha
+   for (int _m = 1; _m < m + 1; ++_m) {
+      for (int _n = 1; _n < n + 1; ++_n) {
+         for (int _k = 1; _k < k + 1; ++_k) {
+            int _a_idx = (_k - 1) * m + (_m - 1) % m;
+            int _b_idx = (_n - 1) * k + (_k - 1) % k;
+            // int _c_idx = (_n - 1) * m + (_m - 1) % m;
+
+            target[0] += variables[0] * variables[_a_idx + offset] * variables[_b_idx + m * k + offset];
+         }
+      }
+   }
+
+   // Check derivative with respect to beta
+   // target[1] = 0;
+   for (int i = offset + m * k + k * n; i < offset + m * k + k * n + m * n; ++i) {
+      target[1] += variables[i];
+   }
+
+   // Check derivatives with respect to matA, matB, matC
+   for (int i = offset; i < m * k + k * n + m * n + offset; ++i) {
+      target[i] = 0;
+      if ((i - offset) < m * k) {
+         // Check derivative with respect to matA
+         int _k = ((i - offset) - (i - offset) % m + 1) / m + 1;
+         for (int _n = 1; _n < n + 1; ++_n) {
+            int index = m * k + (_k - 1) % k + (_n - 1) * k;
+            target[i] += variables[index + offset];
+         }
+      } else if ((i - offset) < m * k + k * n) {
+         // Check derivative with respect to matB
+         int _k = ((i - offset) - m * k) % k + 1;
+         for (int _m = 1; _m < m + 1; ++_m) {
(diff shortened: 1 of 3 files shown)


tmva/sofie/test/TestGemmDerivative.cxx
```diff
@@ -0,0 +1,115 @@
...
+   {
+      m = std::get<0>(GetParam());
+      n = std::get<1>(GetParam());
+      k = std::get<2>(GetParam());
+   }
+
+protected:
+   int m;
+   int n;
+   int k;
```
guitargeek, line 20:
```suggestion
   int m = 0;
   int n = 0;
   int k = 0;
```
Uninitialized variables make me fee uneasy :slightly_smiling_face:

tmva/sofie/test/TestGemmDerivative.cxx
```diff
@@ -0,0 +1,115 @@
...
+                    ret += output[i];
+                }
+                return ret;
+            }
+
+            #pragma clad ON
+            void clad_request () {
+                clad::gradient(gemm_function, "variables");
+            }
+            #pragma clad OFF
```
guitargeek, line 96:
```suggestion
            #pragma clad ON
            clad::gradient(gemm_function, "variables");
            #pragma clad OFF
```
I think this should suffice actually

tmva/sofie/test/TestGemmDerivative.cxx
```diff
@@ -0,0 +1,115 @@
...
+      gInterpreter->Declare(R"cpp(
+            #include <Math/CladDerivator.h>
+
+            float gemm_function(float *variables, int m, int n, int k) {
+                // variable is assumed to pack Amk and Bkn,
+                // so it's of length m*k + k*n
+
+                float alpha = variables[0];
+                float *matA = &variables[1];
+                float *matB = &variables[m*k+1];
```
guitargeek, line 75:
```suggestion
                float alpha = variables[0];
                float *matA = variables + 1;
                float *matB = variables + m*k+1;
```
You can just offset the pointer instead of dereferencing an element and getting back a pointer to it

tmva/sofie/test/TestGemmDerivative.cxx
```diff
@@ -0,0 +1,115 @@
...
+         for (int _m = 1; _m < m + 1; ++_m) {
+            for (int _n = 1; _n < n + 1; ++_n) {
+               for (int _k = 1; _k < k + 1; ++_k) {
+                  _a = (_k - 1) * m + (_m - 1) % m + 1;
+                  _b = (_n - 1) * k + (_k - 1) % k + 1;
+                  target[i] += variables[_a] * variables[_b + m * k];
+               }
+            }
+         }
+      } else if ((i - 1) < m * k) {
```
guitargeek, line 40:
```suggestion
   // Check derivative with alpha
   for (int _m = 1; _m < m + 1; ++_m) {
      for (int _n = 1; _n < n + 1; ++_n) {
         for (int _k = 1; _k < k + 1; ++_k) {
            int _a = (_k - 1) * m + (_m - 1) % m + 1;
            int _b = (_n - 1) * k + (_k - 1) % k + 1;
            target[0] += variables[_a] * variables[_b + m * k];
         }
      }
   }

   for (int i = 1; i < m * k + k * n + 1; ++i) {
      if ((i - 1) < m * k) {
```
Personally, I would avoid some indentation here and make the code more readable by putting the `alpha` section outside the loop.

tmva/sofie/test/TestGemmDerivative.cxx
```diff
@@ -0,0 +1,115 @@
...
+
+                float alpha = variables[0];
+                float *matA = &variables[1];
+                float *matB = &variables[m*k+1];
+
+                constexpr int n_out = 16; // Allocate a maximum length of arrays here
+                assert (m*n+1 <= n_out);
+
+                float output[n_out];
+                float matC[n_out]{};
```
guitargeek, line 81:
```suggestion
                float matC[n_out]{1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.};
```
If we make `matC` non-zero, the derivative with respect to `beta` will be non-zero and we can also validate it by adding `beta` to the variables (the reference result would be the sum of all elements of C).

guitargeek (changes requested):
Thank you very much! I have made some suggestions, and I think we can also easily extend the test to validate also `_d_beta`.

tmva/sofie/test/TestGemmDerivative.cxx
GoodingJamie: Fixed in https://github.com/root-project/root/pull/18364/commits/92e8bcba77ebec1aced9ca59ef307e8247a42736

tmva/sofie/test/TestGemmDerivative.cxx
GoodingJamie: I've tried this but I was getting segfaults with the change, so I've not included it at the moment

guitargeek (approved):
Thank you very much for opening this PR and addressing the review comments!

I was just rebasing it on master, squashing the commits, and avoiding a warning about unused variables. Good to merge now once the CI passes!

Issue #15091, opened 2024-03-28
Clad is an optional module but TFormula (and tests of) fails without it.


pcanal:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

```
Processing /Users/pcanal/root_working/code/master/tutorials/fit/exampleFit3D.C...
In file included from input_line_38:1:
/opt/build/root_builds/master-pr/include/Math/CladDerivator.h:25:10: fatal error: 'plugins/include/clad/Differentiator/Differentiator.h' file not found
#include <plugins/include/clad/Differentiator/Differentiator.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
input_line_39:4:1: error: use of undeclared identifier 'clad'
clad::hessian(TFormula____id13317627456646018083, "p[0:2]" );
^
```

This is due to `TFormula` seemingly unconditionally running:
```
IncludeCladRuntime
from TFormula::GenerateHessianPar(
from TFormula::HessianPar
from GeneralHessianCalc<T>::Hessian
from GeneralHessianCalc<T>::IsAvailable(fFunc)
from WrappedMultiTF1Templ<T>::HasParameterHessian()
from BasicFCNL::HasHessian
from ROOT::Fit::Fitter::DoInitMinimizer
```

 Note: The tutorials seems unrelated to Clad.

### Reproducer

disabled and run `ctest -R tutorial-fit-exampleFit3D`.

With clad disabled, we get:
```
Command: "/Applications/CMake.app/Contents/bin/cmake" "-DCMD=/Users/pcanal/root_working/builds/master-pr/bin/root.exe^-b^-l^-q^/Users/pcanal/root_working/code/master/tutorials/fit/exampleFit3D.C" "-DSYS=/Users/pcanal/root_working/bu
ilds/master-pr" "-DENV=PATH=/Users/pcanal/root_working/builds/master-pr/bin::/Library/Frameworks/Python.framework/Versions/3.9/bin:/Library/Frameworks/Python.framework/Versions/3.9/bin:/Library/Frameworks/Python.framework/Versions/3
.8/bin:/Users/pcanal/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstra
[...]

PR #18371, opened 2025-04-11
[hist,RF] Do not include Clad runtime if clad is disabled in CMake

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/15091

## Checklist:

- [ ] tested changes locally


guitargeek (approved):
Thanks! Indeed this `target_compile_definitions` is a good tool :wink:

PR #18373, opened 2025-04-11
Prevent type sufix to appear in ROOT normalized name

### LINKED ISSUE #18363: Incorrect name normalization for template with size_t template argument
### Check duplicate issues.

- [x] Checked for duplicates

### Description

This is issue is reported indirectly in https://github.com/cms-sw/cmssw/issues/47697 where this bug lead to an unwanted and failing auto-parsing, leading to a crash as a consequence of the failed auto-parsing.

The root cause is that the name normalization of:
```
template <size_t I> struct Outer { template <size_t II> struct View {}; };
Outer<128>
Outer<128>::View<128>
```
is not consistently handled.  During dictionary generation we end up with the following:
```
Outer<128>
Outer<128UL>::View<128>
```
Note the type suffix in 1 out of the 3 possibility ... when none is (currently) desired.

In the reported case this lead the `TProtoClass` for `Outer<128>::View<128>` to have the (normalized) name `Outer<128UL>::View<128>` while at run-time we look up `Outer<128>::View<128>` fail to find it in the rootpcm when using the normalized name interface and thus go through a phase of name normalization which (due to the same/similar failed lookup) leads to the need to load the class information in the interpreter (aka auto-parse its headers).

### Reproducer

```
$ cat Objects.hxx 
#pragma once

namespace reco { 
template <size_t I, bool B> struct PFRecHitSoALayout {
  template <size_t II, bool B1, bool B2, bool B3> struct ViewTemplateFreeParams {};
  using View = ViewTemplateFreeParams<I, B, true, true>; 
}; 
using PFRecHitSoA = PFRecHitSoALayout<128UL,false>; // try with 128ul
}

static reco::PFRecHitSoA abc;
static reco::PFRecHitSoALayout<128UL,false> a2;
```
```
$ cat LinkDef.h 
#pragma link C++ class reco::PFRecHitSoA+;
#pragma link C++ class reco::PFRecHitSoA::View+;
```
```
$ cat Makefile 
all: libCmsSoa.so DataFormatsParticleFlowReco_x2r.cxx

DataFormatsParticleFlowReco_x2r.cxx:  Objects.hxx selection.xml
	rootcling --reflex -failOnWarnings -f -rml libCmsSoa.so -rmf libCmsSoa.rootmap  DataFormatsParticleFlowReco_x2r.cxx Objects.hxx selection.xml

libCmsSoa.so: DataFormatsParticleFlowReco_x2r.cxx
	g++ `root-config --cflags --libs` -o libCmsSoa.so -fPIC --shared DataFormatsParticleFlowReco_x2r.cxx

clean:
	rm -f libCmsSoa.so DataFormatsParticleFlowReco_x2r.cxx DataFormatsParticleFlowReco_x2r_rdict.pcm libCmsSoa.rootmap
```
```
$ root.exe -b -l DataFormatsParticleFlowReco_x2r_rdict.pcm -e 'TObjArray *arr = nullptr; arr = gFile->Get<TObjArray>("__ProtoClasses"); arr->ls()' -q
```

### ROOT version

up to v6.34.08 

### Installation method

any

### Operating system

any

### Additional context

@makortel

pcanal:
This fixes #18363 and thus https://github.com/cms-sw/cmssw/issues/47697.

The companion PR is https://github.com/root-project/roottest/pull/1302

This PR actually 2 related parts
a - an update to LLVM type printer and its use within ROOT
b - an improvement in the detection of unwanted auto-parsing

**a. LLVM update**

 The update to LLVM allows to completely prevent the addition of type suffixes when printing names rather than the current default (there for some types - and in addition, at least how we use it in ROOT, even for those types not consistently there).

In practice, in the failing case, we were getting this normalized name:
```
class reco::PFRecHitSoALayout<128UL,false>::ViewTemplateFreeParams<128,false,true,true>
```
instead of 
```
class reco::PFRecHitSoALayout<128,false>::ViewTemplateFreeParams<128,false,true,true>
```
Note that in the failing case only **one** of the two integers is suffixed by "UL".  I did not tracked down the cause of this discrepancy.
In addition, we also produce alternate spelling of names to ease lookup (by preventing auto-parsing in some trivial case of alternative spelling).

As a side note, another source of confusion is the following.
In the successful run of the example with get:
```
70:      instance.AdoptAlternate(::ROOT::AddClassAlternate("reco::PFRecHitSoALayout<128,false>","reco::PFRecHitSoA"));
72:      instance.AdoptAlternate(::ROOT::AddClassAlternate("reco::PFRecHitSoALayout<128,false>","reco::PFRecHitSoALayout<128ul, false>"));
119:      instance.AdoptAlternate(::ROOT::AddClassAlternate("reco::PFRecHitSoALayout<128,false>::ViewTemplateFreeParams<128,false,true,true>","reco::PFRecHitSoA::View"));
121:      instance.AdoptAlternate(::ROOT::AddClassAlternate("reco::PFRecHitSoALayout<128,false>::ViewTemplateFreeParams<128,false,true,true>","reco::PFRecHitSoALayout<128ul, false>::ViewTemplateFreeParams<128ul, false, true, true>"));
```
where the alternate have the suffix ... but in lower case.  The alternates here are produced by `abi::__cxa_demangle`. (the exact same spelling appears for the alternate in the failing case).

### DIFF
--- a/interpreter/llvm-project/clang/include/clang/AST/PrettyPrinter.h
+++ b/interpreter/llvm-project/clang/include/clang/AST/PrettyPrinter.h
@@ -76,6 +76,7 @@ struct PrintingPolicy {
         SuppressImplicitBase(false), FullyQualifiedName(false),
         PrintCanonicalTypes(false), PrintInjectedClassNameWithArguments(true),
         UsePreferredNames(true), AlwaysIncludeTypeForTemplateArgument(false),
+        NeverIncludeTypeForTemplateArgument(false),
         CleanUglifiedParameters(false), EntireContentsOfLargeArray(true),
         UseEnumerators(true) {}
 
@@ -325,6 +326,13 @@ struct PrintingPolicy {
   /// parameters.
   LLVM_PREFERRED_TYPE(bool)
   unsigned AlwaysIncludeTypeForTemplateArgument : 1;
+
+  /// Whether to never use type suffixes (eg: 1U) on integral non-type template
+  /// parameters.  This is useful to cancel the behavior of
+  /// TemplateParameterList::shouldIncludeTypeForArgument sometimes request the
+  /// type suffix even if AlwaysIncludeTypeForTemplateArgument is not true.
+  LLVM_PREFERRED_TYPE(bool)
+  unsigned NeverIncludeTypeForTemplateArgument : 1;
 
   /// Whether to strip underscores when printing reserved parameter names.
   /// e.g. std::vector<class _Tp> becomes std::vector<class Tp>.
--- a/core/clingutils/src/TClingUtils.cxx
+++ b/core/clingutils/src/TClingUtils.cxx
@@ -635,6 +635,8 @@ bool TClingLookupHelper::GetPartiallyDesugaredNameWithScopeHandling(const std::s
          // TMetaUtils::GetNormalizedName, we just do the 'strip leading std' and fix
          // white space.
          clang::PrintingPolicy policy(fInterpreter->getCI()->getASTContext().getPrintingPolicy());
+         // For now we do not want the integral type suffixes in the normalized name.
+         policy.NeverIncludeTypeForTemplateArgument = true;
          policy.SuppressTagKeyword = true; // Never get the class or struct keyword
          policy.SuppressScope = true;      // Force the scope to be coming from a clang::ElaboratedType.
          // The scope suppression is required for getting rid of the anonymous part of the name of a class defined in an anonymous namespace.
@@ -1411,6 +1413,8 @@ void ROOT::TMetaUtils::GetQualifiedName(std::string &qual_name, const clang::Nam
 {
    llvm::raw_string_ostream stream(qual_name);
    clang::PrintingPolicy policy( cl.getASTContext().getPrintingPolicy() );
+   // For now we do not want the integral type suffixes in the normalized name.
+   policy.NeverIncludeTypeForTemplateArgument = true;
    policy.SuppressTagKeyword = true; // Never get the class or struct keyword
    policy.SuppressUnwrittenScope = true; // Don't write the inline or anonymous namespace names.
 
@@ -4161,6 +4165,8 @@ void ROOT::TMetaUtils::GetNormalizedName(std::string &norm_name, const clang::Qu
 
    clang::ASTContext &ctxt = interpreter.getCI()->getASTContext();
    clang::PrintingPolicy policy(ctxt.getPrintingPolicy());
+   // For now we do not want the integral type suffixes in the normalized name.
+   policy.NeverIncludeTypeForTemplateArgument = true;
    policy.SuppressTagKeyword = true; // Never get the class or struct keyword
    policy.SuppressScope = true;      // Force the scope to be coming from a clang::ElaboratedType.
    policy.AnonymousTagLocations = false; // Do not extract file name + line number for anonymous types.
--- a/interpreter/llvm-project/clang/lib/AST/DeclTemplate.cpp
+++ b/interpreter/llvm-project/clang/lib/AST/DeclTemplate.cpp
@@ -240,6 +240,8 @@ bool TemplateParameterList::hasAssociatedConstraints() const {
 bool TemplateParameterList::shouldIncludeTypeForArgument(
     const PrintingPolicy &Policy, const TemplateParameterList *TPL,
     unsigned Idx) {
+  if (Policy.NeverIncludeTypeForTemplateArgument)
+    return false;
   if (!TPL || Idx >= TPL->size() || Policy.AlwaysIncludeTypeForTemplateArgument)
     return true;
   const NamedDecl *TemplParam = TPL->getParam(Idx);


pcanal:
@makortel @Dr15Jones Any opinions on the interface to see the auto-parsed classes and to disable auto-parsing in TClass::GetClass.

vgvassilev:
Can we find which commit broke that in llvm?

interpreter/llvm-project/clang/include/clang/AST/PrettyPrinter.h
```diff
@@ -326,6 +327,13 @@ struct PrintingPolicy {
   LLVM_PREFERRED_TYPE(bool)
   unsigned AlwaysIncludeTypeForTemplateArgument : 1;
 
+  /// Whether to never use type suffixes (eg: 1U) on integral non-type template
+  /// parameters.  This is useful to cancel the bahevior of TemplateParameterList::shouldIncludeTypeForArgument
```
ferdymercury, line 331:
```suggestion
  /// parameters.  This is useful to cancel the behavior of TemplateParameterList::shouldIncludeTypeForArgument
```

pcanal: Thanks.

pcanal:
> Can we find which commit broke that in llvm?

'Technically' it is `8dff24d0b687e3b2781e00697d32983a9d4d7b52` but it was completely extended and reverting it now requires a longer diff:

<details>
<summary>Details</summary>

```
commit 6e5397bbb20381b93d120bcb8bbd11e985f84102
Author: Philippe Canal <pcanal@fnal.gov>
Date:   Thu Apr 10 14:35:54 2025 -0500

    Revert in Spirit 8dff24d0b687e3b2781e00697d32983a9d4d7b52

diff --git a/interpreter/llvm-project/clang/lib/AST/TemplateBase.cpp b/interpreter/llvm-project/clang/lib/AST/TemplateBase.cpp
index 130bbe8893..80bcaf262d 100644
--- a/interpreter/llvm-project/clang/lib/AST/TemplateBase.cpp
+++ b/interpreter/llvm-project/clang/lib/AST/TemplateBase.cpp
@@ -108,20 +108,20 @@ static void printIntegral(const TemplateArgument &TemplArg, raw_ostream &Out,
     if (const auto *BT = T->getAs<BuiltinType>()) {
       switch (BT->getKind()) {
       case BuiltinType::ULongLong:
-        Out << Val << "ULL";
-        break;
+//        Out << Val << "ULL";
+//        break;
       case BuiltinType::LongLong:
-        Out << Val << "LL";
-        break;
+//        Out << Val << "LL";
+//        break;
       case BuiltinType::ULong:
-        Out << Val << "UL";
-        break;
+//        Out << Val << "UL";
+//        break;
       case BuiltinType::Long:
-        Out << Val << "L";
-        break;
+//        Out << Val << "L";
+//        break;
       case BuiltinType::UInt:
-        Out << Val << "U";
-        break;
+//        Out << Val << "U";
+//        break;
       case BuiltinType::Int:
         Out << Val;
         break;
```

</details>

However, the improvement in itself is not bad and is actually necessary in other cases when producing code, it just need to be made optional (which is what the commit in this PR does).

hahnjo:
> 'Technically' it is `8dff24d0b687e3b2781e00697d32983a9d4d7b52`

Are you sure this is the right commit? 8dff24d0b687e3b2781e00697d32983a9d4d7b52 is a downstream commit by Axel.

> However, the improvement in itself is not bad and is actually necessary in other cases when producing code, it just need to be made optional (which is what the commit in this PR does).

The code you were touching in your revert is in fact optional, for `IncludeType = false`. The question is just which code path leads there, see also https://github.com/root-project/root/pull/17867

vgvassilev:
Why this cannot go upstream?

pcanal:
> Why this cannot go upstream?

I personally do not see any reason why not (It is an opt-in behavior change)

pcanal:
> Are you sure this is the right commit? 8dff24d is a downstream commit by Axel.

Yes, it is the one I meant as it 'originally' (or at least that more or what I thought ... I could wrong) started to introduce the suffixes.

> The code you were touching in your revert is in fact optional, for IncludeType = false

There is indeed 100% correct ;) and the patch I proposing forces `IncludeType` to be false; the fact that it is sometimes not false during our calls is the cause of our problems.

> Edit: Ah, now I see this is where TemplateParameterList::shouldIncludeTypeForArgument comes into play...

:)

vgvassilev:
Probably the part that we will need to explain is why we need a separate switch and should not be the default behavior.

hahnjo:
> > Are you sure this is the right commit? [8dff24d](https://github.com/root-project/root/commit/8dff24d0b687e3b2781e00697d32983a9d4d7b52) is a downstream commit by Axel.
> 
> Yes, it is the one I meant as it 'originally' (or at least that more or what I thought ... I could wrong) started to introduce the suffixes.

Our downstream version yes, but not the code that is causing problems / that you were modifying in your "revert": If you look at https://github.com/root-project/root/blame/b41de52bae91bc44a24f132f80a0e973f05b0c35/interpreter/llvm-project/clang/lib/AST/TemplateBase.cpp#L107-L135, that code comes from upstream LLVM 13. Our downstream lines are in https://github.com/root-project/root/blame/b41de52bae91bc44a24f132f80a0e973f05b0c35/interpreter/llvm-project/clang/lib/AST/TemplateBase.cpp#L137-L142 and there you can also see that it went through many revisions and doesn't look anymore like the first version in 8dff24d0b687e3b2781e00697d32983a9d4d7b52 (but only adds it if required).

If you then look at the upstream blame https://github.com/llvm/llvm-project/blame/028429ac452acde227ae0bfafbfe8579c127e1ea/clang/lib/AST/TemplateBase.cpp#L104-L132 you see that it was introduced by https://github.com/llvm/llvm-project/commit/99d63ccff04b672694f8a2b3eed024b873dc163d (committed by @vgvassilev). That commit also adds `TemplateParameterList::shouldIncludeTypeForArgument`, which AFAICT was introduced during review in https://reviews.llvm.org/D77598. So what really needs investigating is why the CMS template parameter has dependent type, why upstream thinks this requires printing the type, and whether this is potentially correct and we need to adapt to that behavior in ROOT instead of the other way around.

(And why am I spending so much time explaining to you how `git-blame` and archeology works...)

hahnjo (changes requested):
Can you please split off the new `TClass` feature into a separate PR? It should be independent of the template argument printing in Clang.

pcanal:
> Our downstream version yes, but not the code that is causing problems / that you were modifying in your "revert": 

I have not look at the upstream LLVM and the end result is unchanged:
* there is a 'new' behavior in LLVM vaguely relate to Axel's commit
* there failure (unwanted suffix) is related to that and to everything you mentioned
* adding the option, solve the problem (until we change the name normalization).

>  So what really needs investigating is why the CMS template parameter has dependent type, 

If you want you can explore the details with the standalone reproducer https://github.com/root-project/roottest/pull/1302

> and whether this is potentially correct and we need to adapt to that behavior in ROOT instead of the other way around.

Sure but if it is the 'correct' behavior, we still can *not* adapt ROOT in a short time.  We need the name without the suffix (for now).  Hence the opt-in proposed behavior which allows to create a consistent no-suffix name.

vgvassilev:
Okay, I think I start understanding where we are going. So in some places we lost the spelling as written in the code and that resulted the suffix to be printed only sometimes. The current approach of dropping the suffix will likely not work for literals that are larger than in and really need the suffix to make it work. We currently have such cases in the dictionaries iirc (mersenne_twister). 

Can we find out why the suffix was not printed when the user typed it in the code instead?

hahnjo:
A change that fixes one use case still doesn't make a correct solution. I maintain that more care and deeper investigation is required.

> [...] the end result is unchanged:
>  * there is a 'new' behavior in LLVM vaguely relate to Axel's commit

I disagree: Axel's downstream patch adds the type for 64-bit unsigned integers unconditionally *if* required to represent the literal. https://github.com/llvm/llvm-project/commit/99d63ccff04b672694f8a2b3eed024b873dc163d adds a suffix for all widths and regardless of the value if `IncludeType = true`. For me, that's two very different behaviors (admittedly in the same function).

vgvassilev:
I guess this is one of the tricky situations where we have discussed in the past as a theoretical setup.

An experiment needs a fix which ought to be quick. We have a quick fix, which I am pretty sure it is not acceptable upstream. We can't really dedicate a lot of time to turn things upside down in clang due to @pcanal time commitments I suspect.

We have several suboptimal options:
1) Ask cms to keep a patched version of a patched version of llvm
2) Merge this as it is, and have more patches of llvm in ROOT
3) Figure out why we lost the type sugar in that one case, submit a patch upstream, merge it and backport it.

In the past we did not dare with 1); we mostly did 2) and that's why we are where we are now. We rarely tried 3) which arguably suboptimal but a reasonable way forward due to the constrain various people have...

pcanal:
Note: the PR has been trimmed to only the 2 commits related to the name normalization including 0a0e7a97be38d03dc16f1a8e30d2f432bafe5250 which would need to make its way to the mono repo (i.e. 'fix' https://github.com/root-project/root/actions/runs/14448578670/job/40515239847?pr=18373)

vgvassilev:
> > Ask cms to keep a patched version of a patched version of llvm
> 
> This is not an option. The problem would affect any code where the class with a dictionary has a template parameter which is a `size_t` and result is a name normalization that is inconsistent.
> 

I agree.

> > Figure out why we lost the type sugar in that one case, submit a patch upstream, merge it and backport it.
> 
> I am not sure what you have in mind but the code paths here are the one for 'name normalization' where we don't want the type sugar (except in some cases like `Double32_t`).

I do not think we understand what this fixes in practice. I want to reproduce the error with clang and normal c++ code first. That would mean taking a look at the patch that @hahnjo sent with his git blame and see how the tests triggered the codepaths in question.

Another easy way to see the impact of your patch is to apply it on clang, make it on by default and run its test suite. I think the current patch would break quite a few things.

pcanal:
> I think the current patch would break quite a few things.

That's "impossible".  The patch is completely opt-in.  The code would need to somehow (I assume using UB) set the new data member `Policy.NeverIncludeTypeForTemplateArgument` ... i.e. If I read my patch correctly *and* it change the behavior of `Clang` (beside required a rebuild) then there is something seriously wrong with Clang :) (which as I prefaced is "impossible" :) ).

vgvassilev:
> > I think the current patch would break quite a few things.
> 
> That's "impossible". The patch is completely opt-in. The code would need to somehow (I assume using UB) set the new data member `Policy.NeverIncludeTypeForTemplateArgument` ... i.e. If I read my patch correctly _and_ it change the behavior of `Clang` (beside required a rebuild) then there is something seriously wrong with Clang :) (which as I prefaced is "impossible" :) ).

I wrote "Another easy way to see the impact of your patch is to apply it on clang, **make it on by default** and run its test suite."

vgvassilev:
> > > I think the current patch would break quite a few things.
> > 
> > 
> > That's "impossible". The patch is completely opt-in. The code would need to somehow (I assume using UB) set the new data member `Policy.NeverIncludeTypeForTemplateArgument` ... i.e. If I read my patch correctly _and_ it change the behavior of `Clang` (beside required a rebuild) then there is something seriously wrong with Clang :) (which as I prefaced is "impossible" :) ).
> 
> I wrote "Another easy way to see the impact of your patch is to apply it on clang, **make it on by default** and run its test suite."

EDIT: The impossible part is technically right for the current patch and the tests. I can tweak my statement to "I think the current patch would silently break quite a few things" :)

pcanal:
> I wrote "Another easy way to see the impact of your patch is to apply it on clang, make it on by default and run its test suite."

Sorry, I missed that :)

vgvassilev:
> > I wrote "Another easy way to see the impact of your patch is to apply it on clang, make it on by default and run its test suite."
> 
> Sorry, I missed that :)

I am not intending to be annoying here and it may not be obvious from outside but we have spent a lot of blood, sweat and tears in trying to reduce patches that are born in the same way as this PR is going... Let's try to do something else here -- once you get the setup and the commit -- I expect that invested effort won't be significantly larger to fix the issue upstream.

pcanal:
> I want to reproduce the error with clang and normal c++ code first. 
> Another easy way to see the impact of your patch is to apply it on clang, make it on by default and run its test suite. I think the current patch would break quite a few things.

That is fair and would be an investigation into whether or not the change @hahnjo mentioned are functioning as intended or not.  Can someone volunteer to do this investigation while we merge this maybe temporary patch so that CMSSW can move forward? (Note: As long as `TemplateParameterList::shouldIncludeTypeForArgument` is anything more that just `return Policy.AlwaysIncludeTypeForTemplateArgument` then we will still the patch)

pcanal:
> Let's try to do something else here -- once you get the setup and the commit -- I expect that invested effort won't be significantly larger to fix the issue upstream.

What is the issue you are suspecting?  See the example `https://github.com/root-project/roottest/pull/1302/files#diff-9a7c6f13a755a6c9a2b0cef17215a82022ffa407cbe43fe60e6b714f18293ee2`

@devajithvs @aaronj0 Do you have time to look into this?

vgvassilev:
> > I want to reproduce the error with clang and normal c++ code first.
> > Another easy way to see the impact of your patch is to apply it on clang, make it on by default and run its test suite. I think the current patch would break quite a few things.
> 
> That is fair and would be an investigation into whether or not the change @hahnjo mentioned are functioning as intended or not. Can someone volunteer to do this investigation while we merge this maybe temporary patch so that CMSSW can move forward? (Note: As long as `TemplateParameterList::shouldIncludeTypeForArgument` is anything more that just `return Policy.AlwaysIncludeTypeForTemplateArgument` then we will still the patch)

If we merge this pull request we essentially move the responsibility from the author of the PR to somebody else to fix it and that somebody has no incentive to do anything. That's how we got here in a first place. To be constructive, if you feed me a reproducer with standalone clang I can take a look.

pcanal:
> I do not think we understand what this fixes in practice. 

It fixes that when `libCore` or `rootcling` ask for a type to be printed, it can ask **explicitly and locally* for the type to be printed without any integer suffix.

hahnjo:
I had a quick look at one of the Clang tests and `shouldIncludeTypeForArgument` is there for deduced `auto` template parameters:
https://github.com/llvm/llvm-project/blob/81499edb30665d377a680990ef3c5129f9b54261/clang/test/AST/ast-dump-templates.cpp#L79-L103

In this case, the suffix is absolutely required because all of `foo<0,0>`, `foo<0,0u>`, `foo<0,0LL>` are different types. Yes, ROOT Meta currently doesn't treat them correctly, but for me that means just hiding the type is incorrect. The question is why `size_t` is treated the same way...

devajithvs:
>     * adding the option, solve the problem (until we change the name normalization).
What would be required to modify the name normalization? (I’m trying to understand).

I assume you mean setting `shouldIncludeTypeForArgument` to `true`, so that the behavior is always consistent and we can also get rid of one of our LLVM patches.

pcanal:
> What would be required to change the name normalization? (I’m trying to understand)

The problem is not to change it (albeit due to the many code paths, it might also be not straightforward), it is the **consequences** of changing it.  

The change in name normalization will change the checksum of those classes **and** all the classes that contain them.  We would need to upgrade all the algorithm that are use to match CheckSum values to Class layout schema to handle the transition from the old to the new scheme (to be able to still properly read 'old' file).  We may have to adjust older releases (patches) to do the converse to allow forward compatibility.   In addition some experiments (CMS for example) have code to make sure that their code is not changed without update class version and so they record the checksum of the current version ... those would need to be updated.
And finally in user written I/O customization rule that depend on CheckSum would need to be reviewed and possibly updated to reflect the new CheckSum.
And we probably also need for backward compatibility to add code to match the old names to the new names.

hahnjo:
I think this PR can be closed after merging https://github.com/root-project/root/pull/18464

pcanal:
Thanks.

PR #18383, opened 2025-04-12
[clad] Bump clad to v1.10.

guitargeek:
This release introduces significant updates, including support for clang versions 10 through 20 and improved diagnostics for unsupported features. Forward mode now handles variadic functions, while reverse mode sees major enhancements like marking const methods as non-differentiable, enabling differentiation of global variables and static member functions, generating constructor pullbacks automatically, and reducing tape usage through optimizations. CUDA support has been expanded with better handling of device pullbacks and kernel variables. For details, please see the release [notes](https://github.com/vgvassilev/clad/releases/tag/v1.10).

### DIFF
--- a/interpreter/cling/tools/plugins/clad/CMakeLists.txt
+++ b/interpreter/cling/tools/plugins/clad/CMakeLists.txt
@@ -74,7 +74,7 @@ if (DEFINED CLAD_SOURCE_DIR)
   list(APPEND _clad_extra_settings SOURCE_DIR ${CLAD_SOURCE_DIR})
 else()
   list(APPEND _clad_extra_settings GIT_REPOSITORY https://github.com/vgvassilev/clad.git)
-  list(APPEND _clad_extra_settings GIT_TAG v1.9)
+  list(APPEND _clad_extra_settings GIT_TAG v1.10)
 endif()
 
 #list(APPEND _clad_patches_list "patch1.patch" "patch2.patch")


guitargeek:
As one can see from the `CladDericatorTests`, the gradient generation with Clad doesn't seem to work anymore. There is no function call happening then calling `exec()` on the gradient, and `dump()` also doesn't give you any code output.

Here an example macro:
```c++
#include <Math/CladDerivator.h>

double g(double *variables) { return variables[0] * variables[0] * variables[1]; }

void simpleclad() {

    // Call clad to generate the gradient of g.
    auto g_grad = clad::gradient(g, "variables");

    // Execute the generated gradient function.
    double variables[]{3., 4.};
    double grad_output[]{0., 0.};
    g_grad.execute(variables, grad_output);
    std::cout << "grad_output[0]: " << grad_output[0] << std::endl;
    std::cout << "grad_output[1]: " << grad_output[1] << std::endl;

    // Dump the generated gradient code to standard output.
    g_grad.dump();

}
```

The output with ROOT compiled with Clad `master` is empty:
```txt
root [0]
Processing simpleclad.C...
grad_output[0]: 0
grad_output[1]: 0
The code is:

root [1]
```

guitargeek:
The regression originated from this commit, @vgvassilev:

https://github.com/vgvassilev/clad/commit/612eeb8875be2e196ff664e47add9b4417f19673

vgvassilev:
> The regression originated from this commit, @vgvassilev:
> 
> [vgvassilev/clad@612eeb8](https://github.com/vgvassilev/clad/commit/612eeb8875be2e196ff664e47add9b4417f19673)

Can you try this patch:

```diff
diff --git a/tools/ClangPlugin.cpp b/tools/ClangPlugin.cpp
index 8f5a2550..919ccb79 100644
--- a/tools/ClangPlugin.cpp
+++ b/tools/ClangPlugin.cpp
@@ -138,7 +138,7 @@ namespace clad {
         if (!isa<FunctionDecl>(D))
           continue;
         FunctionDecl* FD = cast<FunctionDecl>(D);
-        if (FD->isConstexpr()) {
+        if (FD->isConstexpr() || !m_Multiplexer) {
           DiffCollector collector(DGR, CladEnabledRange, m_DiffRequestGraph, S,
                                   opts);
           break;
```

guitargeek:
Indeed, that fixes it! Thanks for your quick help

vgvassilev:
> Indeed, that fixes it! Thanks for your quick help

Ok, in that case I can land that on the other side and release.

vgvassilev:
@guitargeek, do we need somehow a clean build or there is something else to be fixed?

guitargeek:
The problem will be fixed by this PR:
https://github.com/root-project/root/pull/18478

vgvassilev:
@dpiparo this seems that it can go forward for the 6.36 release.

guitargeek:
Hi @vgvassilev yes I think we can do that. Can you approve this PR so it can me merged? Then we'll see about the 6.36 backport

vgvassilev (approved):
Lgtm.

guitargeek:
Unfortunately, this upgrade has introduced a significant performance hits to our Higgs combination benchmarks.

I ran the same workflow as described in this PR:
https://github.com/root-project/root/pull/17926

The with current ROOT master, I get this result:
plot_clad_1_10

This problem seems to be independent of TBR, because when I disable it in the [relevant place](https://github.com/root-project/root/blob/master/roofit/roofitcore/src/RooFuncWrapper.cxx#L176) with `<clad::opts::disable_tbr>`, there is still the regression but it's also a bit slower because of no TBR:

plot_clad_1_10_no_tbr

If I use ROOT `master` but change the Clad tag back to 1.9, I get back the "old" results:

plot_clad_1_9

Can I help somehow to address this regression?

vgvassilev:
Maybe https://github.com/vgvassilev/clad/pull/1373 with setting an env variable `CLAD_ENABLE_TIMING` can help us figure out where we waste most time.

PR #18384, opened 2025-04-12
[ci] Remove duplicate env variables

hahnjo:
They are already defined globally.


hahnjo:
Unfortunately will have to test after merging, the authentication is relevant for uploading the artifacts...

dpiparo (approved):

PR #18386, opened 2025-04-13
[CMake] Fixes and imporvments for buildng with `builtin_clang=OFF`.

guitargeek:
1. Make it possible to find Clang outside of the LLVM prefix
2. Correctly reset `LLVM_LINK_LLVM_DYLIB` after finding Clang

More detail in the commit descriptions.


dpiparo (approved):

dpiparo:
we may even think to a build exercising this functionality at this point.

guitargeek:
> we may even think to a build exercising this functionality at this point.

Thanks for the review! I migrated my personal workflows to always build with `builtin_llvm=OFF` and `builting_clang=OFF` now, so save some time when rebuilding from scratch. So we'll get some almost daily test coverage for this :slightly_smiling_face:

core/clingutils/CMakeLists.txt
```diff
@@ -80,31 +80,19 @@ foreach(dict ${stldicts})
...
-    file(GLOB CHILDREN RELATIVE ${CLANG_RESOURCE_DIR_STEM} ${CLANG_RESOURCE_DIR_STEM}/*)
-    list(LENGTH CHILDREN CHILDREN_LENGTH)
-    if (${CHILDREN_LENGTH} GREATER 1)
-      message(FATAL_ERROR "Found more than one version of clang. CLANG_RESOURCE_DIR_VERSION contains: '${CHILDREN}'." )
-    endif()
-
-    list(GET CHILDREN 0 CLANG_RESOURCE_DIR_VERSION)
+  # If no explicit Clang prefix is set, assume it lives inside LLVM prefix.
+  if(NOT DEFINED CLANG_INSTALL_PREFIX)
+    set(CLANG_INSTALL_PREFIX ${LLVM_DIR})
```
hahnjo, line 88: `LLVM_DIR` is the wrong variable, it is only a hint specified by the user. The right approach is to take one of the variables set by `LLVMConfig.cmake`, see also commit bb4fde0b89a03242213079844527079d2c244cd1

hahnjo:
I would appreciate if at least one person involved in the interpreter / LLVM business is added as reviewer for this kind of PRs...

interpreter/CMakeLists.txt
```diff
@@ -407,6 +407,12 @@ if (builtin_clang)
 else()
   find_package(Clang REQUIRED CONFIG)
   message(STATUS "Found Clang ${CLANG_PACKAGE_VERSION} in ${CLANG_CMAKE_DIR}")
+
+  # Disable linking against shared LLVM.
+  # We have to do this after find_package(Clang). Finding Clang internally
+  # calls find_package(LLVM), which overwrites LLVM_LINK_LLVM_DYLIB to what it
+  # was set when building LLVM.
+  set(LLVM_LINK_LLVM_DYLIB FALSE)
```
hahnjo, line 415: Revisiting the code, this is also not ideal: The right solution that we have chosen in Cling is to pass `DISABLE_LLVM_LINK_LLVM_DYLIB` to not use the shared library even if present. @guitargeek can you please revert this patch and implement the proper solutions?

guitargeek: It's not as easy, as we have discussed in private. Also, the exact same thing is done in the `builtin_clang` code path:
https://github.com/root-project/root/blob/master/interpreter/CMakeLists.txt#L380

So this PR here only made the configuration code less inconsistent, which is always good IMHO. So I don't see the reason to revert if the supposedly "right" solution is not implemented in our main configuration code path either.

core/clingutils/CMakeLists.txt
guitargeek: You're right, thanks! I'll fix this up, going back to the `${LLVM_LIBRARY_DIR}` that was used before

interpreter/CMakeLists.txt
hahnjo: > It's not as easy, as we have discussed in private.

I disagree with this summary of our discussions. For me it is clear, and always has been, that the right solution is implemented on the linking side for `libCling`.

> Also, the exact same thing is done in the `builtin_clang` code path: https://github.com/root-project/root/blob/master/interpreter/CMakeLists.txt#L380

Yes, though it is slightly different because there we are actually building Clang, not just modifying a variable after `find_package(Clang)`. But I agree that it should also be reviewed.

> So this PR here only made the configuration code less inconsistent, which is always good IMHO.

(for a bit more consistency in the file, it should have been `OFF` instead of `FALSE`...)

> So I don't see the reason to revert if the supposedly "right" solution is not implemented in our main configuration code path either.

This PR was proposed on a Sunday afternoon, on the day before locking the branch, and merged within 2 hours, without review from any of the developers recently involved in this area before. I have pointed out one immediate mistake, and I maintain that fiddling with `LLVM_LINK_LLVM_DYLIB` is not the correct solution. The statement about "main configuration code path" is incorrect since the default, and only supported configuration for production use as far as I'm concerned, is `builtin_llvm=ON builtin_clang=ON`.

As changes to the build system this late in the cycle can have a significant blast area, I maintain that this PR should be reverted.

core/clingutils/CMakeLists.txt
guitargeek: PR opened: https://github.com/root-project/root/pull/18407

interpreter/CMakeLists.txt
guitargeek: > For me it is clear, and always has been, that the right solution is implemented on the linking side for libCling.

You seemed to have no problem approving and merging the precedent for the `builtin_clang=ON` case: https://github.com/root-project/root/pull/13420#pullrequestreview-1571176267. What's *really* different now?

What you say seems contradictory to me. In the issue I opened you said that `builtin_llvm=OFF && builtin_clang=OFF` are maintained on a [best effort basis](https://github.com/root-project/root/issues/18387#issuecomment-2800639450), and now you argue that changes in these configuration paths can have a significant blast area. Why don't you just let me do the best effort then even though you think it's not perfect :slightly_smiling_face: Better than having something that doesn't work. And it's only natural that these things are discovered late in the release cycle, because that's when we start thinking about how the new release can be built and packaged.

> The statement about "main configuration code path" is incorrect since the default, and only supported configuration for production use as far as I'm concerned, is builtin_llvm=ON builtin_clang=ON

You're right, I apologize. I forgot that the precedent was for `builtin_llvm=ON` `builtin_clang=OFF`

hahnjo: 1. In that code path we are building Clang, which I assumed is influenced by setting this variable. If I understand your upstream PR correctly, setting it before building Clang actually doesn't make a difference. So maybe it doesn't serve a purpose in the end, to be tested.
2. The PR added `DISABLE_LLVM_LINK_LLVM_DYLIB` for linking, which I think is the right solution. The setting of `LLVM_LINK_LLVM_DYLIB` may have been an oversight among the other changes in the PR.
3. I delved deeper into `interpreter/CMakeLists.txt` since then, in particular last December / January when I cleaned it up quite a bit. This required some reverse engineering to understand what is going on, when I also discovered that some things are not working as I thought.
4. After realizing that one path is not the right solution, I think it is bad practice to approve merging more code. If code is already there, it should be reworked instead of added to.

Beyond that, I don't consider it good style to link a PR from almost two years ago and say "you accepted it back then"...

> now you argue that changes in these configuration paths can have a significant blast area

What I argue is that build system changes in general can have a significant blast radius. This PR contains changes on the main configuration code path in `core/clingutils/CMakeLists.txt`.

> Why don't you just let me do the best effort then even though you think it's not perfect

Because "best effort" doesn't mean that I don't care, and it doesn't mean that the use cases can be broken at will.

> Better than having something that doesn't work.

YMMV.

> And it's only natural that these things are discovered late in the dev cycle, because that's when we start thinking about how the new release can be built and packaged.

Then this should be changed, there will be a next cycle.

hahnjo: I've submitted a revert in https://github.com/root-project/root/pull/18446

PR #18391, opened 2025-04-14
[ci] Remove Ubuntu 24.10 builds

dpiparo:
6.36 will be released at the end of May. EOL of that distro is scheduled for July 11th.
This will also free one linux build slot for each and every PR to the main and future v6-36-00-patches branch.


guitargeek (approved):
Thanks!

PR #14747, opened 2024-02-16
[core] Do not skip libCore when finding out to what library the class belongs

ferdymercury:
# This Pull request:

## Changes or fixes:

Bug introduced in commit from 2019 Jun 2nd, where gInterpreter->GetClassSharedLibs() returns nullptr for any class inside the ROOT core base classes.

## Checklist:

- [x] tested changes locally
- [ ] updated the docs (if necessary)

Fixes https://github.com/root-project/root/issues/11667

PR #18396, opened 2025-04-14
Remove RIndexIter.hxx

hahnjo:
I missed this header in commit 9aa2fc5bb2 when removing the current RHist prototype.


dpiparo (approved):
Thanks!

hahnjo:
The CI shows a couple of failures that I can actually reproduce locally:
 * 3 PyMVA and 1 TMVA test crash with `munmap_chunk(): invalid pointer` (in CI) or segmentation violations (locally). AFAICT all of them involve PyTorch and it looks like the `std::regex` problem, though it's unclear how this is triggered by removing an unrelated header. @guitargeek @lmoneta @sanjibansg can you maybe have a look?
 * `roottest-root-tree-fastcloning-execCheckClusterRange` shows changed file sizes (-2 bytes, +5 bytes). Again unclear how this is triggered by an unrelated header - @pcanal any idea?

In the end, I don't really care if we remove this unused header. I didn't expect issues (that's why I didn't run the full test suite locally), but in the end it's worrying if such easy change can trigger this amount of failures...

pcanal:
> roottest-root-tree-fastcloning-execCheckClusterRange shows changed file sizes (-2 bytes, +5 bytes). Again unclear how this is triggered by an unrelated header - @pcanal any idea?

That is very surprising :(

ferdymercury:
> > roottest-root-tree-fastcloning-execCheckClusterRange shows changed file sizes (-2 bytes, +5 bytes). Again unclear how this is triggered by an unrelated header - @pcanal any idea?
> 
> That is very surprising :(

We had some surprises here in the past: https://github.com/root-project/root/pull/14747#issuecomment-1949014653

pcanal:
I can reproduce `roottest-root-tree-fastcloning-execCheckClusterRange` in the docker `alma9 march=native` but can not reproduce locally on my alma9 node.

ferdymercury:
> I can reproduce `roottest-root-tree-fastcloning-execCheckClusterRange` in the docker `alma9 march=native` but can not reproduce locally on my alma9 node.

I can reproduce it locally on Ubu22 (with runtime_modules enabled before/after).

When grepping, I see that before the change, RIndexIter appears here:

```
build/core/G__Core.cxx:#include "ROOT/RIndexIter.hxx"
build/include/ROOT.modulemap:  module "ROOT/RIndexIter.hxx" { header "ROOT/RIndexIter.hxx" export * }
build/include/module.modulemap.extra:  module "ROOT/RIndexIter.hxx" { header "ROOT/RIndexIter.hxx" export * }
build/include/module.modulemap.extra.tmp:  module "ROOT/RIndexIter.hxx" { header "ROOT/RIndexIter.hxx" export * }
grep: build/lib/modules.idx: binary file matches
grep: build/lib/Core.pcm: binary file matches
```

so somehow, that ends up affecting the BuildStreamerInfo. Maybe the total number of headers in Core's PCM plays a role? (We could check by putting back RIndexIter in the branch, but leaving it fully empty. Or by adding an RIndexIter2 and RIndexIter3 files).

pcanal:
The issue seems to be that the order in which StreamerInfo are build is different which affect the compression ration of the StreamerInfo record.  In turn this affect the location of baskets in the file which affects the compression ration of the `TTree` which affect the printout.  Since this is auxiliary to the test, the solution is to disable compression for the file in that test.  Nonetheless, I am still debugging/investigating why/how this header's presence affects the TStreamerInfo creation order.

pcanal:
The failure CheckClusterRange test should work once https://github.com/root-project/root/pull/18452 is merged.

hahnjo:
> The failure in the CheckClusterRange test should work once #18452 is merged.

Thanks @pcanal, this also fixes the PyTorch failures. Maybe in those cases, library load order is also different?

@dpiparo I think the PR is ready now. The failure on Windows is due to "not enough space on the disk" ...

pcanal:
> this also fixes the PyTorch failures. Maybe in those cases, library load order is also different?

This is indeed likely (with this PR and without the fix, any tickling of TBranch I/O could/would load an arbitrary unwanted library) ... Now, I would say this technically would mean that the bug is uncovering a (known) real issue, i.e. that our PyTorch is sensitive to order library loading (but I guess that's just the regex issue ...)

Issue #13300, opened 2023-07-21
Test SIGTERM vs ~TFile


Axel-Naumann:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Rumors have it that SIGTERM doesn't trigger writing in-memory objects to open `TFile`s. We should have a test for this.

### Reproducer

Likely this should be enough:
```C++
TFile file("out.root", "RECREATE");
TH1F hist("h", "h", 10, 0., 1.);
raise(SIGTERM);
```

### ROOT version

6.29/01

### Installation method

From source

### Operating system

Some Linux

Axel-Naumann:
It might be that this bug requires the object to be in a sub-`TDirectory`:
```C++
TFile file("out.root", "RECREATE");
file.mkdir("subdir")->cd();
TH1F hist("h", "h", 10, 0., 1.);
raise(SIGTERM);
```

ferdymercury:
If we add this in TerminalConfigUnix.cpp, then the first example is fixed, but the second one is still failing.

```
#include "TROOT.h"
#include "TMethodCall.h"

if (signum==SIGTERM) {
// TO-DO this could be moved to gROOT->CleanupOnSigterm();
if (gROOT) {
     R__LOCKGUARD(gROOTMutex);

     if (gROOT->GetListOfFiles()) {
        TIter next(gROOT->GetListOfFiles());
        TObject* obj = nullptr;
        while((obj = (TObject*)next())) {
           if (obj) {
              TMethodCall call1(obj->IsA(), "Write", "");
              TMethodCall call2(obj->IsA(), "Close", "");
              call1.Execute((void *)(obj));
              call2.Execute((void *)(obj));
           }
        }
     }
  gROOT->GetListOfFiles()->Delete("slow");
  }
      if (gROOT->GetListOfSockets())
         gROOT->GetListOfSockets()->Delete();
      if (gROOT->GetListOfMappedFiles())
         gROOT->GetListOfMappedFiles()->Delete("slow");
      if (gROOT->GetListOfClosedObjects())
         gROOT->GetListOfClosedObjects()->Delete("slow");
}
  signal(signum, SIG_DFL); // unregister ourselves
```

pcanal:
> It might be that this bug requires the object to be in a sub-TDirectory:

Is there really a difference between the 2 cases?

> Rumors have it that SIGTERM doesn't trigger writing in-memory objects to open TFiles.

[...]

PR #18403, opened 2025-04-14
[io] Prevent losing data due to SIGTERM (under Unix)

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/13300

- I tested this locally on Ubu22 and it works fine.
- I do not know how to properly test it using GTEST, I only found test_polling.cxx doing something like that but it looked too complicated, so I ended up simplifying.
- I do not know if Windows must catch SIGTERM somewhere else @bellenot ?

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/base/src/TROOT.cxx
+++ b/core/base/src/TROOT.cxx
@@ -119,6 +119,7 @@ FARPROC dlsym(void *library, const char *function_name)
 #include "TClass.h"
 #include "TClassEdit.h"
 #include "TClassGenerator.h"
+#include "TDirectory.h"
 #include "TDataType.h"
 #include "TStyle.h"
 #include "TObjectTable.h"
@@ -238,14 +239,10 @@ void TROOT::WriteCloseAllFiles()
          TIter next(gROOT->GetListOfFiles());
          while (TObject *obj = next()) {
             if (obj && obj->InheritsFrom(TClass::GetClass("TFile", kFALSE, kTRUE))) {
-               TMethodCall callIsWritable(obj->IsA(), "IsWritable", "");
-               Longptr_t retLong = 0;
-               callIsWritable.Execute((void *)(obj), retLong);
-               if (retLong == 1) {
-                  TMethodCall callWrite(obj->IsA(), "Write", "");
-                  callWrite.Execute((void *)(obj));
-                  TMethodCall callClose(obj->IsA(), "Close", "");
-                  callClose.Execute((void *)(obj));
+               auto fobj = static_cast<TDirectory *>(obj);
+               if (fobj->IsWritable()) {
+                  fobj->Write();
+                  fobj->Close();
                }
             }
          }
--- a/core/base/src/TROOT.cxx
+++ b/core/base/src/TROOT.cxx
@@ -227,9 +227,31 @@ static Int_t ITIMQQ(const char *time)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+/// Write and Close all open Files, useful to be called when SIGTERM is caught.
+
+void TROOT::WriteCloseAllFiles()
+{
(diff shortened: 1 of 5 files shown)


bellenot:
@ferdymercury The SIGILL and SIGTERM signals aren't generated under Windows. They're included for ANSI compatibility. See the [signal reference](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=msvc-170)

dpiparo:
Let's have a look to the tests once again

core/base/src/TROOT.cxx
```diff
@@ -226,10 +226,37 @@ static Int_t ITIMQQ(const char *time)
...
+void TROOT::WriteCloseAllFiles()
+{
+   if (gROOT) {
+      R__LOCKGUARD(gROOTMutex);
+
+      if (gROOT->GetListOfFiles()) {
+         TIter next(gROOT->GetListOfFiles());
+         while (TObject *obj = next()) {
+            if (obj && obj->InheritsFrom(TClass::GetClass("TFile", kFALSE, kTRUE))) {
+               TMethodCall callIsWritable(obj->IsA(), "IsWritable", "");
```
dpiparo, line 241: Why is `TMethodCall` needed and a cast is not enough?

ferdymercury: We would need to cast to TFile and thus `#include "TFile.h"`, which would introduce a dependency of libCore from libRIO, which I guess is not ideal? The only other solution would be to create a baseclass in Core from which T(Directory)File derives that implements IsWritable and is overridden in the derived.

pcanal: It is already in `TDirectory` ...

core/textinput/src/textinput/TerminalConfigUnix.cpp
```diff
@@ -104,6 +106,11 @@ TerminalConfigUnix::HandleSignal(int signum) {
     }
   }
 
+  // gentle save and close if SIGTERM
+  if (signum == SIGTERM) {
```
pcanal, line 110: If I understood correctly, this would mean that after code similar to:
```
root [] TFile f(filename, RECREATE);
root [] do_stuff();
root [] if (all_is_well) f.Write();
```
We have a different behavior if the user does:
```
root [] .q
```
and
```
send SIGTERM
```
Where the former does not write the file while the later would now write the file.

ferdymercury: Yes.

dpiparo: As I tried to explain here, https://github.com/root-project/root/issues/13300#issuecomment-3998719876, I think this is a logical behaviour. The prompt experience does not change. The behaviour if SIGKILL is used does not change. However, if SIGTERM is used, some kind of "soft termination" that gives the process a way to wrap up and end gracefully, is the right thing to do.

dpiparo:
I would be in favour of approving, @pcanal , but only if you also agree with the point I made above.

core/textinput/src/textinput/TerminalConfigUnix.cpp
ferdymercury: We could create a ".wq" ROOT command (one char longer than ".q") to do the equivalent of save and exit as if it were vim, if one wants to rely on this without having to look how to send SIGTERM. Or just for symmetry with SIGNALS.

pcanal:
See https://github.com/root-project/root/issues/13300#issuecomment-4014769133 for details on why this underlying feature can not be implemented without undefined behavior.

Thanks for proposing this PR.

PR #18406, opened 2025-04-14
Fix the module map for GCC 15 and bump the C++ standard to 23 for the Fedora 42 build

dpiparo:
Thanks @vgvassilev for the commit fixing the module map.

### DIFF
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -362,7 +362,7 @@ jobs:
         include:
           - image: fedora41
           - image: fedora42
-            overrides: ["CMAKE_CXX_STANDARD=20"]
+            overrides: ["CMAKE_CXX_STANDARD=23"]
           - image: alma8
           - image: alma9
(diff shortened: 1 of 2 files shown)


vgvassilev:
Do we have a build with gcc-8?

hahnjo:
> Do we have a build with gcc-8?

Yes, `alma8` is with GCC 8.5.0. It's not using C++20 though, so there shouldn't be any difference for that (famous last words)

vgvassilev:
> > Do we have a build with gcc-8?
> 
> Yes, `alma8` is with GCC 8.5.0. It's not using C++20 though, so there shouldn't be any difference for that (famous last words)

I am interested in the other than 20 builds so should be fine (another set of famous last words).

hahnjo (approved):
Changes LGTM

dpiparo:
Thanks both. Let's wait for 25.04 and then I'll merge. Please contact me if you think the matrix of builds can be better organised, e.g. in terms of C++ standards.

PR #18410, opened 2025-04-15
[ntuple] Move RNTuple out of ROOT 7

silverweed:
# This Pull request:
- moves all RNTuple code and doc from `tree/ntuple/v7` to `tree/ntuple` 
- Removes the ROOT 7 experimental warnings in the header comments of all classes that have been moved out of Experimental
- makes RNTuple build by default with ROOT - not gated anymore by the `root7` flag.

Note that `ntupleutil` stays in v7 for the time being.

## Checklist:

- [x] tested changes locally


tree/ntuple/Testing/Temporary/LastTest.log
hahnjo, line 1: These `Testing` files look unintended :monocle_face:

hahnjo (approved):
LGTM after removing two unintended files in the last commit. Thanks for concluding the move!

jblomer (commented):
I think we should also
  - Change the header guards `ROOT7_...` --> `ROOT_...`
  - Remove `\ingroup ROOT7`
  - Probably move the RNTuple data source in the RDF package out of root7

jblomer (commented):
And perhaps also: 
  - Update (remove) the last section of the README.md
  - Move the ntuple tutorials out of root7

silverweed:
>  Update (remove) the last section of the README.md

I think we can do this in a future PR where we do another pass on the doc (removing those lines would remove the ROOT7 doxygen group and we probably want to keep it for the few classes that still belong to it)

>  Move the ntuple tutorials out of root7

Those were already moved during the hackathon :)

vepadulano (approved):
LGTM for the RDF part, thank you!

PR #18411, opened 2025-04-15
[ntuple] Remove dependency from TError.h

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-10805

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/core/foundation/inc/ROOT/RError.hxx
+++ b/core/foundation/inc/ROOT/RError.hxx
@@ -27,6 +27,16 @@
 #include <utility>
 #include <vector>
 
+/*! Checks condition `e` and throws an exception if it's false.
+ * \warning this check is NOT stripped in release mode, so it should not be used for hot paths.
+ * For those cases, prefer a regular `assert()`;
+ */
+#define R7__ASSERT(e)                                             \
+   do {                                                           \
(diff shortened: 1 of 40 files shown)


ferdymercury:
Related PR fixing RNTupleMerger part by @silverweed: https://github.com/root-project/root/pull/18411

jblomer:
Many thanks for the patch and apologies for the late reply! This patch gives us a clearer picture of the `assert` situation.

I think that the current state is actually better than ROOT-10805 may indicate. Contrary to early ideas in the RNTuple development, we don't need or want to entirely remove the dependency to `TError.h`. We do want to distinguish the error cases
  - exceptional event from which user code can recover, such as I/O failure (the case for `RException`) from
  - fatal, unrecoverable error because of an unexpected condition (case for `assert` or `R__ASSERT` depending on whether we check only in debug builds or release builds)

So I think over time we came to the conclusion that `R__ASSERT` has a valid where we want to fail early to help debugging.

My suggestion is to close this PR and to review the use of `R__ASSERT` in the RNTuple code, making sure that the error handling is consistent. I'll update ROOT-10805.

PR #16410, opened 2024-09-11
[cling] Remove support for declarations without 'auto' keyword

devajithvs:
# This Pull request:
Removes the implicit `auto` keyword support added in: https://github.com/root-project/root/commit/071d08487ceea4f8f28231ea182ba350e6836e30 and https://github.com/root-project/root/commit/1f4a6dc0539acfc0ec58cea7443fe08a0a4479d4
## Changes or fixes:
This feature was deprecated in: https://github.com/root-project/root/pull/14645

Will update LLVM and squash the commits here before the final merge/decision.

## Checklist:

- [ ] tested changes locally
- [ ] updated the docs (if necessary)

This PR fixes #

PR #14311, opened 2024-01-08
[cmake] Set `fail-on-missing=ON` by default

guitargeek:
Most ROOT developers and users seem to agree that autonomously toggling features at configuration time based on the environment is not good. The feature set that ROOT is built with is then not deterministic.

In past cases, this already resulted in accidentally missing test coverage because features were switched off after environment changes. That's why for the CI, we are always building with the `fail-on-missing-option`.

However, this is not only a problem in testing, but everytime ROOT is built. That's why this commit suggests to make `fail-on-missing` the default, and warn the users of potential future deprecation of this flag in the release notes.

An interesting point is also that the `fail-on-missing` code path in `SearchInstalledSoftware.cmake` is much simpler, which also helps to reduce the margin for error.

See also:
https://github.com/root-project/root/issues/14188#issuecomment-1844965943

PR #18413, opened 2025-04-15
[cmake] feature=OFF should have hierarchical priority over (cached, or not) builtin_feature=ON

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-10743

feature=ON can force-enable builtin_feature=OFF, but not the other way round, a warning is issued in that case now instead of silently enabling.

This prevents automatic enablings of features that the user can not disable unless he writes feature=OFF builtin_feature=OFF.

I've been often annoyed about using -Dfftw3=OFF and cmake overriding it anyway.
If this new behavior is not wanted, then I would suggest closing the associated JIRA Issue.

Related:

- https://its.cern.ch/jira/browse/ROOT-9385

## Checklist:

- [x] tested changes locally


guitargeek (approved):
I'm all for it, thank you very much! And I think emitting a warning is a very good idea

guitargeek:
I think this has to be mentioned in the release notes though

ferdymercury:
Thanks Jonas for the review! 6.36 relnotes or 6.38? I will do that in a separate PR.

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -642,7 +642,9 @@ if(mathmore OR builtin_gsl OR (tmva-cpu AND use_gsl_cblas))
     # FIXME: one need to find better way to extract path with GSL include files
     set(GSL_INCLUDE_DIR ${CMAKE_BINARY_DIR}/GSL-prefix/src/GSL-build)
     set(GSL_FOUND ON)
-    set(mathmore ON CACHE BOOL "Enabled because builtin_gls requested (${mathmore_description})" FORCE)
+    if(NOT mathmore)
+      message(WARNING "builtin_gls was enabled but mathmore wasn't")
+    endif()
```
pcanal, line 647: Wouldn't that be 'better' or is `mathmore` already set on by default somewhere else?
```suggestion
    set(mathmore ON CACHE BOOL "Enabled because builtin_gls requested (${mathmore_description})")
    if(NOT mathmore)
      message(WARNING "builtin_gls was enabled but mathmore wasn't")
    endif()
```

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -924,24 +926,27 @@ if(fftw3)
...
-    BUILD_IN_SOURCE 1
-    BUILD_BYPRODUCTS ${FFTW_LIBRARIES}
-    TIMEOUT 600
-  )
-  set(FFTW_INCLUDE_DIR ${CMAKE_BINARY_DIR}/include)
-  set(FFTW3_TARGET FFTW3)
-  set(fftw3 ON CACHE BOOL "Enabled because builtin_fftw3 requested (${fftw3_description})" FORCE)
+  if(NOT fftw3)
+    message(WARNING "fftw3 is OFF, hence ignoring the activated setting 'builtin_fftw3'")
+  else()
```
pcanal, line 931: This does not feel quite right.  This seems to now require *both*
```
-Dbuiltin_fftw3=ON -Dffw3=ON
```
to enable `ffw3` with builtin or is some other part of the `CMakeList` allowing the simpler syntax?
I.e. like https://github.com/root-project/root/pull/18413/files#r2044449280 maybe we 
need to move old line 944 here-ish, remove the `FORCE` and test afterwards?

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -950,9 +955,12 @@ if(fitsio OR builtin_cfitsio)
     ROOT_CHECK_CONNECTION_AND_DISABLE_OPTION("builtin_cfitsio")
   endif()
   if(builtin_cfitsio)
-    add_library(CFITSIO::CFITSIO STATIC IMPORTED GLOBAL)
-    add_subdirectory(builtins/cfitsio)
-    set(fitsio ON CACHE BOOL "Enabled because builtin_cfitsio requested (${fitsio_description})" FORCE)
+    if(NOT fitsio)
```
pcanal, line 958: Same comment :)

pcanal (dismissed):
Unless I am mistaken, the patch does not seem to accomplish exactly what the title hints at.

ferdymercury:
> Unless I am mistaken, the patch does not seem to accomplish exactly what the title hints at.

Thanks for the review!

The proposed behavior here is as follows:
Calling just -Dbuiltin_fftw3=ON, one will get a warning saying that fftw3=OFF, so it will be ignored. The user needs to call both.
On the other hand, if one sets fftw3=ON, CMake can auto-set internally builtin_fftw3=ON if it does not find the system-wide package.
Finally, if I want to turn it back of, I just need to set fftw3=OFF, and not like before fftw3=OFF builtin_fftw3=OFF.

So the debate is, do we want 2 flags for turning off vs 2 flags for turning on.

I do not know if there is a third way allowing for "remembering" if builtin_fftw3 was set by a previous call to CMake by the user, or by auto-finding when not getting system packages, or actively by the user.

ferdymercury:
> Note that this sounds very risk to do one week before branching. There is simply not enough time to test all possible combinations and make sure they work correctly.

Yes, I fully agree, 6.38 seems more reasonable.

cmake/modules/SearchInstalledSoftware.cmake
ferdymercury: we have `ROOT_BUILD_OPTION(mathmore OFF "Build libMathMore extended math library (requires GSL) [GPL]")`
unless -Dall=ON then we have set(mathmore_defvalue ON)

But: tmva-cpu AND use_gsl_cblas could require builtin_gsl also.
So, it's not 100% legit to turn on mathmore just because builtin_gsl is ON ?

This goes with the general philosophy of this PR. Builtins are hierarchically lower than "features". You turn on or off features, which may or not use builtins. Turning ON builtins cannot turn ON features automatically, you will just get a warning (or a cmake error if you prefer), whereas turning ON features can force-enable builtins without further thoughts.

pcanal:
>  I do not know if there is a third way allowing for "remembering" if builtin_fftw3 was set by a previous call to CMake by the user

There is :). (not necessarily easy code but there is)

pcanal:
> So the debate is, do we want 2 flags for turning off vs 2 flags for turning on.

An important side note is that the proposed behavior will (likely) break existing build scripts (i.e. this is a breaking change).

In attempt to clarify my mind on this, I started this table to list the current behavior, proposed behavior and comments.

| CMakeCache State | fftw3 | builtin_fftw3 | Resulting State | Proposed State | Comment
| --- | --- | --- | --- | --- | --- |
| (empty) | ON | (none) |  fftw3=ON builtin_fftw3=ON | (same) | |
| (empty) | (none) | ON |  fftw3=ON builtin_fftw3=ON | fftw3=OFF builtin_fftw3=ON | Why not keep turning on fftw3 in that case? | 
| fftw3=ON builtin_fftw3=ON | OFF | (none) | fftw3=ON builtin_fftw3=ON | fftw3=OFF builtin_fftw3=ON | The existing situation is surprising. |
| fftw3=OFF builtin_fftw3=ON (currently impossible) | ON | (none) | fftw3=ON builtin_fftw3=ON | fftw3=ON builtin_fftw3=ON |  |

I am not sure it helps or if we should flush it out.

I does denotes in the 2nd row, the breaking change ... which I believe is not necessary and could be kept as is while still addressing the bug shown in the 3rd row.

guitargeek:
But the change For row 2 is the whole point of this PR as Ferdy said, and personally I agree with his opinion. Why you don't think that the proposed behavior is better, besides backwards compatibility concerns? Fixing row 3 is more of a niche case, not sure if it's worth to do something about it...

pcanal:
> Why you don't think that the proposed behavior is better, besides backwards compatibility concerns? 

Because it unnecessarily (as far I can tell) breaks backward compatibility (and the proposed code in the PR does not quite introduce "hierarchical priority" (which would be backward compatible)). (And in row 2, the state `fftw3=OFF builtin_fftw3=ON` seems like an unintentional of 'waste' of time - why would a user request the builtin fftw3 ... just for it to sit unused? -- if the user say `fftw3=OFF` explicitly then fine, this is a what they asked for but if they did not specify it, it makes no sense (to me))

cmake/modules/SearchInstalledSoftware.cmake
pcanal: > This goes with the general philosophy of this PR. 

Indeed, I agree with the first part (`disabling the feature should disable the feature`) but not with the second part (`enabling builtin without *also* enabling the feature fails to do what the user likely meant` [there is editorial comment in the 2nd description :) :) ] )

guitargeek:
> proposed code in the PR does not quite introduce "hierarchical priority"

As far as I understand, the right hierarchical dependency that we want is the following. So far: `xrootd=ON` implies `xrootd_fftw3=ON` (if not found on the system), and `builtin_xrootd=ON` implied `xrootd=ON`. So they are "symmetric". We want that `xrootd_fftw3=ON` does not imply anything anymore, but instead `xrootd=OFF` implies `xrootd_fftw3=OFF`. This introduces a clear hierarchy on which options value implies which other option.

> And in row 2, the state fftw3=OFF builtin_fftw3=ON seems like an unintentional of 'waste' of time - why would a user request the builtin fftw3 ... just for it to sit unused? -- if the user say fftw3=OFF explicitly then fine, this is a what they asked for but if they did not specify it, it makes no sense (to me)

This is exactly what this PR is supposed to address, no @ferdymercury? Maybe what makes it so confusing that so far, there is no clear policy on how `option` and `builtin_option` interact, they all do it differently. For example, `xrootd=ON` implies `xrootd_fftw3=ON` if xrootd is not found on the system, like I mentioned before. However, `fftw3=ON` does **not** imply `builtin_fftw3=ON`, but still `builtin_fftw3=ON` implies `fftw3=ON`, which is actually the inverted hierarchy of what this PR proposes! Doesn't make sense to me.

> Because it unnecessarily (as far I can tell) breaks backward compatibility

The configuration is so random right now that I would not prioritize backwards compatibility but instead consistency and clear documented policies from now on.

ferdymercury:
> This introduces a clear hierarchy on which options value implies which other option.

Yes, exactly, that was my proposal. And independently on whether the option was cached from a previous run or not.

> still `builtin_fftw3=ON` implies `fftw3=ON`

Are you sure? I just thought that this snippet prevented that:

```
if(builtin_fftw3)
  if(NOT fftw3)
    message(WARNING "fftw3 is OFF, hence ignoring the activated setting 'builtin_fftw3'")
```

> However, `fftw3=ON` does **not** imply `builtin_fftw3=ON`

I have fixed this issue in the latest commit, thanks for noticing. (I overlooked this).

> I would not prioritize backwards compatibility but instead consistency and clear documented policies from now on.

In my opinion, if keeping backward compatibility can be done with let's say 20 lines of extra total CMake code (I am not sure how to do that), then I agree with @pcanal's assessment. If however it requires very complicated CMake code that is hard to maintain and makes the automatic ROOT-CMake procedures less hard to disentangle / reproduce, then I agree with @guitargeek that having a simplified decision hygiene, a bit simple and dumb, but giving you a bit pedantic warnings so that it's supereasy to follow, would be better even at the price of breaking backward compatibility (with a clear warning of what to fix when calling your build script). And it's easier to maintain if someone else adds in the future a new ROOT build option.
Related discussion: https://github.com/root-project/root/pull/14311#issue-2071310206

guitargeek:
> Are you sure? I just thought that this snippet prevented that:

Sorry I was talking about the inconsistent behavior of ROOT `master`, not this PR

pcanal:
> In my opinion, if keeping backward compatibility can be done with let's say 20 lines of extra total CMake code (I am not sure how to do that)

It looks like it can be done with a couple of lines (per case of builtin).  For example: https://github.com/root-project/root/pull/18413/files#r2044449280 might be sufficient (and if not it can still be done with a couple more lines).

pcanal:
> but still builtin_fftw3=ON implies fftw3=ON

I don't think that is a problem.  The 'only' problem I see is that the current code does the wrong thing in that it does more than that and actually does `builtin_fftw3=ON **forces** fftw3=ON` (and this is a serious problem).

I extend the table to include all configurations for a from scratch configuration and in the proposed state column is now included 'my' recommendation:

| # | CMakeCache State | fftw3 | builtin_fftw3 | Resulting State | Proposed State | Comment |
| --- | --- | --- | --- | --- | --- | --- |
| 1 | (empty) | ON | (none) |  fftw3=ON builtin_fftw3=ON | (same) | |
| 2 | (empty) | ON | OFF |  fftw3=ON builtin_fftw3=OFF | (same) | | 
| 3 | (empty) | ON | ON |  fftw3=ON builtin_fftw3=ON | (same) |  | 
| 4 | (empty) | (none) | (none) |  fftw3=OFF builtin_fftw3=OFF | (same)  |  | 
| 5 | (empty) | (none) | OFF |  fftw3=OFF builtin_fftw3=OFF | (same) |  | 
| 6 | (empty) | (none) | ON |  fftw3=ON builtin_fftw3=ON | (same) | the debated configuration  | 
| 7 | (empty) | OFF | (none) |  fftw3=OFF builtin_fftw3=OFF | (same) |  | 
| 8 | (empty) | OFF | OFF |  fftw3=OFF builtin_fftw3=OFF | (same) |  | 
| 9 | (empty) | OFF | ON |  fftw3=ON builtin_fftw3=ON | fftw3=OFF builtin_fftw3=ON | the problem  | 
| 10 | fftw3=ON builtin_fftw3=ON | OFF | (none) | fftw3=ON builtin_fftw3=ON | fftw3=OFF builtin_fftw3=ON | The existing situation is surprising. |
| 11 | fftw3=OFF builtin_fftw3=ON (currently impossible) | ON | (none) | fftw3=ON builtin_fftw3=ON | fftw3=ON builtin_fftw3=ON |  |

As a side note, if we follow the recommendation from the PR, the existing configuration:
```
cmake -Dbuiltin_fftw3=ON $root_src
```
will 'silently' no longer build the `fftw3` library (i.e. an experiment could build the entire stack release it, only to have their user discover the miss configuration sometime later).
So if we go that route we actually need to detect the case and error out.
Consequently the amount of code (times the number of builtin) is the same order of magnitude for both solution.


>  Maybe what makes it so confusing that so far, there is no clear policy on how option and builtin_option interact, they all do it differently. 

That is also bad and need to either be 'fixed' or documented ....

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -642,7 +642,9 @@ if(mathmore OR builtin_gsl OR (tmva-cpu AND use_gsl_cblas))
     # FIXME: one need to find better way to extract path with GSL include files
     set(GSL_INCLUDE_DIR ${CMAKE_BINARY_DIR}/GSL-prefix/src/GSL-build)
     set(GSL_FOUND ON)
-    set(mathmore ON CACHE BOOL "Enabled because builtin_gls requested (${mathmore_description})" FORCE)
+    if(NOT mathmore)
+      message(WARNING "builtin_gls was enabled but mathmore wasn't")
```
ferdymercury, line 646:
```suggestion
      message(FATAL_ERROR "builtin_gls was enabled, conflict with mathmore=OFF")
```

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -912,36 +914,43 @@ if(fftw3)
...
-    LOG_DOWNLOAD 1 LOG_CONFIGURE 1 LOG_BUILD 1 LOG_INSTALL 1 LOG_OUTPUT_ON_FAILURE 1
-    BUILD_IN_SOURCE 1
-    BUILD_BYPRODUCTS ${FFTW_LIBRARIES}
-    TIMEOUT 600
-  )
-  set(FFTW_INCLUDE_DIR ${CMAKE_BINARY_DIR}/include)
-  set(FFTW3_TARGET FFTW3)
-  set(fftw3 ON CACHE BOOL "Enabled because builtin_fftw3 requested (${fftw3_description})" FORCE)
+  if(NOT fftw3)
+    message(WARNING "fftw3 is OFF, hence ignoring the activated setting 'builtin_fftw3'")
```
ferdymercury, line 934:
```suggestion
    message(FATAL_ERROR "fftw3 is OFF, conflict with the activated setting 'builtin_fftw3'")
```

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -950,9 +959,12 @@ if(fitsio OR builtin_cfitsio)
     ROOT_CHECK_CONNECTION_AND_DISABLE_OPTION("builtin_cfitsio")
   endif()
   if(builtin_cfitsio)
-    add_library(CFITSIO::CFITSIO STATIC IMPORTED GLOBAL)
-    add_subdirectory(builtins/cfitsio)
-    set(fitsio ON CACHE BOOL "Enabled because builtin_cfitsio requested (${fitsio_description})" FORCE)
+    if(NOT fitsio)
+       message(WARNING "fitsio is OFF, hence ignoring the activated setting 'builtin_cfitsio'")
```
ferdymercury, line 963:
```suggestion
       message(FATAL_ERROR "fitsio is OFF, conflict with the activated setting 'builtin_cfitsio'")
```

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -1013,19 +1025,22 @@ if(xrootd AND NOT builtin_xrootd)
 endif()
 
 if(builtin_xrootd)
-  ROOT_CHECK_CONNECTION("builtin_xrootd=OFF")
-  if(NO_CONNECTION)
-    message(FATAL_ERROR "No internet connection. Please check your connection, or either disable the 'builtin_xrootd'"
-      " option or the 'fail-on-missing' to automatically disable options requiring internet access")
+  if(NOT xrootd)
+    message(WARNING "xrootd is OFF, hence ignoring the activated setting 'builtin_xrootd'")
```
ferdymercury, line 1029:
```suggestion
    message(FATAL_ERROR "xrootd is OFF, conflict with the activated setting 'builtin_xrootd'")
```

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -1184,15 +1199,18 @@ if(davix AND NOT builtin_davix)
 endif()
 
 if(builtin_davix)
-  ROOT_CHECK_CONNECTION("builtin_davix=OFF")
-  if(NO_CONNECTION)
-    message(STATUS "No internet connection, disabling the 'davix' and 'builtin_davix' options")
-    set(builtin_davix OFF CACHE BOOL "Disabled because there is no internet connection" FORCE)
-    set(davix OFF CACHE BOOL "Disabled because there is no internet connection" FORCE)
+  if(NOT davix)
+    message(WARNING "davix is OFF, hence ignoring the activated setting 'builtin_davix'")
```
ferdymercury, line 1203:
```suggestion
    message(FATAL_ERROR "davix is OFF, conflict with the activated setting 'builtin_davix'")
```

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -1286,15 +1304,19 @@ int main() { return 0; }" tbb_exception_result)
 endif()
 
 if(builtin_tbb)
-  ROOT_CHECK_CONNECTION("builtin_tbb=OFF")
-  if(NO_CONNECTION)
-    message(STATUS "No internet connection, disabling 'builtin_tbb' and 'imt' options")
-    set(builtin_tbb OFF CACHE BOOL "Disabled because there is no internet connection" FORCE)
-    set(imt OFF CACHE BOOL "Disabled because 'builtin_tbb' was set but there is no internet connection" FORCE)
+  if(NOT imt)
+    message(WARNING "imt is OFF, hence ignoring the activated setting 'builtin_tbb'")
```
ferdymercury, line 1308:
```suggestion
    message(FATAL_ERROR "imt is OFF, conflict with the activated setting 'builtin_tbb'")
```

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -1367,9 +1389,12 @@ endif()
 
 #---Check for Vc---------------------------------------------------------------------
 if(builtin_vc)
-  unset(Vc_FOUND)
-  unset(Vc_FOUND CACHE)
-  set(vc ON CACHE BOOL "Enabled because builtin_vc requested (${vc_description})" FORCE)
+  if(NOT vc)
+    message(WARNING "vc is OFF, hence ignoring the activated setting 'builtin_vc'")
```
ferdymercury, line 1393:
```suggestion
    message(FATAL_ERROR "vc is OFF, conflict with the activated setting 'builtin_vc'")
```

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -1483,72 +1507,76 @@ elseif(veccore)
...
-    set(VecCore_LIBRARIES ${VecCore_LIBRARIES} ${Vc_LIBRARIES})
-    target_link_libraries(VecCore INTERFACE ${Vc_LIBRARIES})
-  endif()
-
-  find_package_handle_standard_args(VecCore
-    FOUND_VAR VecCore_FOUND
-    REQUIRED_VARS VecCore_INCLUDE_DIRS VecCore_LIBRARIES
-    VERSION_VAR VecCore_VERSION)
+  if(NOT veccore)
+    message(WARNING "veccore is OFF, hence ignoring the activated setting 'builtin_veccore'")
```
ferdymercury, line 1516:
```suggestion
    message(FATAL_ERROR "veccore is OFF, conflict with the activated setting 'builtin_veccore'")
```

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -1576,46 +1604,49 @@ if(vdt OR builtin_vdt)
...
-    set(vdt ON CACHE BOOL "Enabled because builtin_vdt enabled (${vdt_description})" FORCE)
-    set_property(GLOBAL APPEND PROPERTY ROOT_BUILTIN_TARGETS VDT)
-    add_library(VDT::VDT STATIC IMPORTED GLOBAL)
-    set_target_properties(VDT::VDT
-      PROPERTIES
-        IMPORTED_LOCATION "${VDT_LIBRARIES}"
-        INTERFACE_INCLUDE_DIRECTORIES "${VDT_INCLUDE_DIRS}"
-    )
+    if(NOT vdt)
+      message(WARNING "vdt is OFF, hence ignoring the activated setting 'builtin_vdt'")
```
ferdymercury, line 1608:
```suggestion
      message(FATAL_ERROR "vdt is OFF, conflict with the activated setting 'builtin_vdt'")
```

ferdymercury:
> discover the miss configuration sometime later).

Side note, this also happens if they have a missing Internet connection and fail-on-missing is off.

> So if we go that route we actually need to detect the case and error out.
> Consequently the amount of code (times the number of builtin) is the same order of magnitude for both solution.

Agreed. I have transformed the warnings into errors now. So I guess this would be plan A, and plan B would be remove the FATAL_ERROR message and soft-enable the feature via the cache as you mentioned.

ferdymercury:
Related issue: https://its.cern.ch/jira/projects/ROOT/issues/ROOT-9385

guitargeek (changes requested):
Taking back the approval because superseded by:
  * https://github.com/root-project/root/pull/18467

silverweed:
@ferdymercury should this PR be closed, considering that there is a v2 of it, or is it still needed for something?

ferdymercury:
> @ferdymercury should this PR be closed, considering that there is a v2 of it, or is it still needed for something?

I think that first, the ROOT team should decide what's the best way to follow of these three options.

a) Leave things as they are and close the JIRA issue as won't fix and document this limitation.
b) Change behavior by simplifying things a lot as done in this PR (v1). Very clear to follow and easy to maintain in the future, it forces though the user to be a bit pedantic when passing some options since some of the automation logic is removed. Even if it might interfere with some build scripts, it's very easy to fix them and people might like the new clear pedantic logic rather than the current magic OFF does not turn OFF. Note: the current behavior is undocumented, broken and unconsistent. For example, builtin_unuran does not turn unuran ON. So does this PR really break behavior? Since there are two simultaneous philosophies, we are choosing one to be consistent, so it could be seen more as a fix of UndefinedBehavior rather than as a breaking change.
c) Do not compromise the complex automatic logic. Try at the same time to fix the bug ad suggested by pcanal. This needs advanced complex Cmake tricks. That's what I attempted in v2 of this PR. But I did not succeed, see discussion https://github.com/root-project/root/pull/18467#issuecomment-2823616362 Maybe https://stackoverflow.com/questions/43542381/set-a-cmake-variable-if-it-is-not-changed-by-the-user helps but I think an expert needs to take over here.
d) Like b) but adding first a deprecation warning and doing the change in 6.40. I don't really like this option since the current behavior is still broken or inconsistent or undocumented, so why deprecating it, if the solution is rather easy if a user relied on the auto-turn-on.

So before closing v1 or v2 or both, a decision or analysis of cost benefit ratio is needed I guess.

This issue reminds me a bit of https://github.com/root-project/root/pull/16410

In my view in both issues, if the code complexity/maintenance/clarity is reduced, then it's worth to nudge the users to modify their scripts to add `auto ` or `feature=ON`.

ferdymercury:
Closing: superseded by https://github.com/root-project/root/pull/23092

> a) Leave things as they are and close the JIRA issue as won't fix and document this limitation.
> b) Change behavior by simplifying things a lot as done in this PR (v1). Very clear to follow and easy to maintain in the future, it forces though the user to be a bit pedantic when passing some options since some of the automation logic is removed. Even if it might interfere with some build scripts, it's very easy to fix them and people might like the new clear pedantic logic rather than the current magic OFF does not turn OFF. Note: the current behavior is undocumented, broken and unconsistent. For example, builtin_unuran does not turn unuran ON. So does this PR really break behavior? Since there are two simultaneous philosophies, we are choosing one to be consistent, so it could be seen more as a fix of UndefinedBehavior rather than as a breaking change.
> c) Do not compromise the complex automatic logic. Try at the same time to fix the bug ad suggested by pcanal. This needs advanced complex Cmake tricks. That's what I attempted in v2 of this PR. But I did not succeed, see discussion [#18467 (comment)](https://github.com/root-project/root/pull/18467#issuecomment-2823616362) Maybe https://stackoverflow.com/questions/43542381/set-a-cmake-variable-if-it-is-not-changed-by-the-user helps but I think an expert needs to take over here.
> d) Like b) but adding first a deprecation warning and doing the change in 6.40. I don't really like this option since the current behavior is still broken or inconsistent or undocumented, so why deprecating it, if the solution is rather easy if a user relied on the auto-turn-on.

In summary, the new PR chooses option b) but with a philosophy of giving the user a helpful best-educated-guess copy-pastable error message so that, everything that was happening before behind the scenes automatically is converted instead to a visible error message that the user can then decide to copy-paste or not into his command line. That way, responsibility is delegated, behind-the-scenes is avoided, and code maintenance is minimized by avoiding buggy issues with the cache. The user-mediated second-pass input is our human-cache :)
Since 'what the cache was doing before' is copy-pastable into the own terminal, the developer does not lose much time in figuring what builtins have to be enabled or not, etc, compared to having to parse the error messages scattered in many places.
So: there is no hierarchy at all now, user build options are 'sacred' and no one can turn them ON or OFF. Only suggestions will be printed to terminal.

PR #18414, opened 2025-04-15
[ci] Skip invalid ranges with ruff format

siliataider:
When the first lines of a file are removed (as we can see in [18412](https://github.com/root-project/root/actions/runs/14465867825/job/40567985665?pr=18412)), `git diff` hunks produce a range of `0-0`. This range is invalid to pass to ruff format and results in an error.
This addresses the issue by skipping such ranges.


vepadulano (approved):
Thank you for the fix!

PR #18418, opened 2025-04-15
[df] Use heap TTree and TFile in test

vepadulano:
Similar to what was done in https://github.com/root-project/root/commit/5f56b95832d0f31d0d85dc0a6cc93aa9cb375159

Might help fixing the sporadic failures seen on the CI such as https://github.com/root-project/root/actions/runs/14436281697/job/40477834555#step:13:23422

### DIFF
--- a/tree/dataframe/test/dataframe_interface.cxx
+++ b/tree/dataframe/test/dataframe_interface.cxx
@@ -48,14 +48,21 @@ TEST(RDataFrameInterface, CreateFromStrings)
 class TreeInFileRAII {
 private:
    std::string fPath;
-   TFile fFile;
 
 public:
-   explicit TreeInFileRAII(const std::string &path) : fPath(path), fFile(path.c_str(), "recreate")
+   explicit TreeInFileRAII(const std::string &path) : fPath(path)
    {
-      TTree t("t", "t");
-      fFile.WriteObject(&t, "t");
-      fFile.Close();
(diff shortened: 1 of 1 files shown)


pcanal (approved):
LGTM.

tree/dataframe/test/dataframe_interface.cxx
```diff
@@ -48,14 +48,21 @@ TEST(RDataFrameInterface, CreateFromStrings)
...
-   explicit TreeInFileRAII(const std::string &path) : fPath(path), fFile(path.c_str(), "recreate")
+   explicit TreeInFileRAII(const std::string &path) : fPath(path)
    {
-      TTree t("t", "t");
-      fFile.WriteObject(&t, "t");
-      fFile.Close();
+      // NOTE(vpadulan): these TFile and TTree are created on the heap to work around a know bug that can
+      // cause a TObject to be incorrectly marked as "on heap" and attempted to be freed despite
+      // living on the stack.
+      // The bug is caused by the magic bit pattern `kObjectAllocMemValue` used by TStorage to
```
pcanal, line 58: On some platform, the mechanism is wholly disable due to the OS introducing its own memory poising that conflicts with the mechanism used by to attempt to detect stack object.  The wording `caused by the magic bit pattern` is a bit odd unless it is  meant to imply that the random failure is caused by a 'random' situation leading to the magic pattern to happen to be present when it should not have been (i.e. by random bad luck).

Issue #16327, opened 2024-08-29
Improve header file encapsulation


vgvassilev:
### Explain what you would like to see improved and how.

We should move all headers from $ROOTSYS/include to $ROOTSYS/ROOT/{component} and we should automatically generate wrapping headers that `#include` the relevant ROOT/ header with a deprecation warning.
Eg.

```bash
mv ROOTSYS/include/TLish.h ROOTSYS/include/ROOT/core/TLish.h
cat ROOTSYS/include/TLish.h
```

```cpp
#warn "This legacy header will go away in X"
#include "ROOT/Base/TLish.h"
```

The trampoline header file can be generated by our build system by adding a -DROOT_COMPATIBILITY switch that's on by default for few releases and then off.

### ROOT version

master

### Installation method

N/A

### Operating system

N/A

dpiparo:
Hi @vgvassilev could you please perhaps share a bit of the motivation you see behind this proposal?

vgvassilev:
Pretty much the same motivation of having new headers in ROOTSYS/include/ROOT. The upshot is that the third party codebase will become clearer of that `TList.h` is a ROOT thing; and more importantly we will define away a class of problems we have when installing ROOT. Right now on some systems we install everything in `/usr/include`, which is not a good practice unless one is libc.

vgvassilev:
I remember @linev has made some progress in the past with this but then we were blocked by something wrt to modules that I do not remember...

linev:
Some time ago we adjust cmake files so that when building ROOT libraries we using headers from source directories and not headers from `$ROOTSYS/include`. This makes dependencies between ROOT components more clear.

I just checked $ROOTSYS/include - there are 1476 files. And going this way we will double number of files.

vgvassilev:
Yes, that is correct but only doubling them in count and during the deprecation phase.

ferdymercury:
This seems a duplicate (or very similar to) of: https://its.cern.ch/jira/browse/ROOT-9403 so I'd suggest closing one or the other.

I am copying here the relevant discussion there:

When installing ROOT, we currently install all headers into `/usr/include`, e.g. `/usr/include/TFile.h`. However, as modules requires us to provide a modulemap for our headers which has the be in the same directory, we would have to place our ROOT modulemap in `/usr/include/modulemodulemap`. This leads to conflicts with all other system libraries, which also need to place a modulemap in `/usr/include` (libc being the prime example).

[...]

PR #6014, opened 2020-07-09
Including headers for all libraries based on ROOT_STANDARD_LIBRARY_PACKAGE

bernhardmgruber:
Partially solves: [ROOT-10915](https://its.cern.ch/jira/browse/ROOT-10915)

Reasoning from the ticket:
> It seems most header files are not added to the library targets that compose ROOT. This is most imminent to me when I create a Visual Studio project by CMake. Almost all header files are missing in the solution explorer.

> While this still allows to build ROOT, the IDE experience is severly limited. Automatic code refactoring and static analysis fails to a large degree, because the headers are not considered part of the project. E.g. the clang-tidy integration in Visual Studio cannot fix any header files.

PR #18419, opened 2025-04-15
[RFC] [cmake] Add core/base headers to target HEADERS file_set

ferdymercury:
# This Pull request:

## Changes or fixes:

Gives a taste on how it would look like to properly add HEADERS to targets in CMake.

This does not just help with the IDE workflow, it gives an idea on future paths:

One could make them PUBLIC instead of PRIVATE, then you would not need the target_include_directories line, and installing and cpacking would be easier.

Related:
- https://github.com/root-project/root/issues/16327
- https://github.com/root-project/root/pull/8709
- https://github.com/root-project/root/pull/6014
- https://its.cern.ch/jira/browse/ROOT-6428
- https://its.cern.ch/jira/browse/ROOT-10915
- https://github.com/root-project/root/pull/17607

## Checklist:

- [ ] tested changes locally

This PR fixes #

### DIFF
--- a/core/base/CMakeLists.txt
+++ b/core/base/CMakeLists.txt
@@ -198,10 +198,26 @@ set(BASE_SOURCES
   src/TVirtualX.cxx
 )
 
+set(FULL_BASE_HEADERS ${BASE_HEADERS})
+list(TRANSFORM FULL_BASE_HEADERS PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/inc/)
+
 # only here complete list of headers can be propogated to parent cmake file
 set_property(TARGET Core APPEND PROPERTY DICT_HEADERS ${BASE_HEADERS})
 
 target_sources(Core PRIVATE ${BASE_SOURCES})
+
+if(NOT CMAKE_VERSION VERSION_LESS "3.23.0") # https://discourse.cmake.org/t/file-set-xyz-is-listed-in-interface-file-sets-of-w-but-has-not-been-exported/9131/3
+  target_sources(
+      Core
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/ src/
+      FILES
+          ${CMAKE_CURRENT_SOURCE_DIR}/src/TListOfTypes.h
+          ${FULL_BASE_HEADERS}
+  )
+endif()
 
 target_include_directories(Core PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>


pcanal:
Also potentially related: https://github.com/root-project/root/pull/17607

core/base/CMakeLists.txt
```diff
@@ -209,6 +212,21 @@ set_property(TARGET Core APPEND PROPERTY DICT_HEADERS ${BASE_HEADERS})
...
+      Core
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/ src/ v7/inc/
+      FILES
+          ${CMAKE_CURRENT_SOURCE_DIR}/src/TListOfTypes.h
+          ${FULL_BASE_HEADERS}
+          ${CMAKE_CURRENT_SOURCE_DIR}/v7/inc/ROOT/RFloat16.hxx
+          ${CMAKE_CURRENT_SOURCE_DIR}/v7/inc/ROOT/RIndexIter.hxx
```
ferdymercury, line 226:
```suggestion
```

core/base/CMakeLists.txt
```diff
@@ -209,6 +212,21 @@ set_property(TARGET Core APPEND PROPERTY DICT_HEADERS ${BASE_HEADERS})
 
 target_sources(Core PRIVATE ${BASE_SOURCES})
 
+if(NOT CMAKE_VERSION VERSION_LESS "3.23.0") # https://discourse.cmake.org/t/file-set-xyz-is-listed-in-interface-file-sets-of-w-but-has-not-been-exported/9131/3
+  target_sources(
+      Core
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/ src/ v7/inc/
```
ferdymercury, line 221:
```suggestion
      BASE_DIRS inc/ src/
```

core/base/CMakeLists.txt
```diff
@@ -198,11 +198,27 @@ set(BASE_SOURCES
   src/TVirtualX.cxx
 )
 
+set(FULL_BASE_HEADERS ${BASE_HEADERS})
+list(TRANSFORM FULL_BASE_HEADERS PREPEND ${CMAKE_CURRENT_SOURCE_DIR}/inc/)
+
```
hageboeck: Reading the documentation in CMake, I believe this is unnecessary, since CMake will expand the paths relative to the current directory (+ a possible BASE_DIR), see below.
```suggestion
```

core/base/CMakeLists.txt
```diff
@@ -198,11 +198,27 @@ set(BASE_SOURCES
...
+if(NOT CMAKE_VERSION VERSION_LESS "3.23.0") # https://discourse.cmake.org/t/file-set-xyz-is-listed-in-interface-file-sets-of-w-but-has-not-been-exported/9131/3
+  target_sources(
+      Core
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/ src/
+      FILES
+          ${CMAKE_CURRENT_SOURCE_DIR}/src/TListOfTypes.h
+          ${FULL_BASE_HEADERS}
```
hageboeck: BASE_DIRS should make the headers findable, and there's nothing in `src` that needs to be picked up using a relative path, I believe:
```suggestion
      BASE_DIRS inc/
      FILES
          ${BASE_HEADERS}
          ${CMAKE_CURRENT_SOURCE_DIR}/src/TListOfTypes.h
```

hageboeck (commented):
@ferdymercury, would you agree that the same can be achieved with less code, since CMake already does the expansion to absolute paths for us?

core/base/CMakeLists.txt
ferdymercury: I hoped so, but it does not work the 'simple way'.

If I do what you say, then I get these errors at configure time:

```
/home/user/root_src/core/base/TVirtualX.h
[cmake] 
[cmake]   must be in one of the file set's base directories:
[cmake] 
[cmake]     /home/user/root_src/core/base/inc
[cmake]     /home/user/root_src/core/base/src
```

It's because BASE_DIR does not mean "add this relative prefix" to every single header, but rather something different. Bad choice of name from CMake side I guess. See https://gitlab.kitware.com/cmake/cmake/-/issues/25402#note_1443892

However, it's true that I can slightly simplify by using instead:
`list(TRANSFORM FULL_BASE_HEADERS PREPEND inc/)`
ie not need to prepend the full path there.

core/base/CMakeLists.txt
ferdymercury: Does not work correctly like this, see explanation above.
BASE_DIRS is not to make the header findable by adding a relative path, but to force the headers to be in those dirs. (Weird CMake convention). Ie the FILES must be already written each of them 'relative to the current cmake-source-dir', not 'relative to base-dir'

ferdymercury:
Simplification done!

Btw this is the result in the GUI / IDE after this change:

image

Unfolded:

image

core/base/CMakeLists.txt
ferdymercury: > there's nothing in `src` that needs to be picked up

Btw, we could split things into two file-sets, one called private_headers and the other public_headers.

Later on, we would only "INSTALL" the public ones I guess, using sth like:
https://discourse.cmake.org/t/cmake-file-set-for-header-when-using-destination/5692/2

Here is a full example of 'migration' using also a VERIFY target. https://github.com/bemanproject/exemplar/pull/30/files

In principle, we could then also get rid of the copy-headers-to-build-dir, but that's orthogonal to this PR. This is an independent first step that makes the IDE work better.
That's the reason that I did not remove (yet) the `target_include_directories(build_dir/inc)`

pcanal: > In principle, we could then also get rid of the copy-headers-to-build-dir,

That part is tightly related to how the dictionary are produced and indeed should definitively be its own PR.

core/base/CMakeLists.txt
```diff
@@ -198,11 +198,27 @@ set(BASE_SOURCES
...
+
 # only here complete list of headers can be propogated to parent cmake file
 set_property(TARGET Core APPEND PROPERTY DICT_HEADERS ${BASE_HEADERS})
 
 target_sources(Core PRIVATE ${BASE_SOURCES})
 
+if(NOT CMAKE_VERSION VERSION_LESS "3.23.0") # https://discourse.cmake.org/t/file-set-xyz-is-listed-in-interface-file-sets-of-w-but-has-not-been-exported/9131/3
+  target_sources(
+      Core
+      PRIVATE
```
ferdymercury, line 212: @guitargeek thanks for the discussion. Followup: see this keyword here, one can easily separate private from public headers by doing:

```
target_sources(Core
  PUBLIC
  FILE_SET public_headersOrWhateverNameYouWant
  TYPE HEADERS
...


target_sources(Core
  PRIVATE
  FILE_SET private_headersOrWhateverNameYouWant
...
```

Right now, it might make more sense to first define everything is private headers without touching target_include_directories, ie we are just adding the headers for IDE purposes.

In a separate MR one could then remove all the target_include_directories by correctly defining which headers are public and which are private.

guitargeek: Ok, thank you for the clarification! Then I'm looking forward to your follow-ups :slightly_smiling_face:

PR #18422, opened 2025-04-16
Initial changes for different snapshot behaviours in distRDF

### LINKED ISSUE #17136: Allow different Snapshot behaviours in distributed RDataFrame
### Feature description

The Snapshot operation, when called locally, will produce one output file to the path provided by the user. Currently, the distributed version of Snapshot produces one output file per task, with the result of running the computation graph on the event range processed by that task. This difference in behaviour was introduced for performance reasons, but for the end users it can be sometimes confusing. We should introduce new behaviours to the distributed RDataFrame function:

1. Allow the user to request one final merged output file, understanding that this will incur in performance costs.
2. Allow the user to get back the list of partial output files, so they can harvest them according to further workflow requirements. Also, we should check that it's possible to write the files to a remote storage location where the user has write access (e.g. via an xrootd path).

### Alternatives considered

_No response_

### Additional context

_No response_

Aditya-138-12:
# This Pull request:
Allows different snapshot behaviours in dist RDF

## Changes or fixes:
1. Made 3 new functions within the class, `_get_base_filename`, `_merge_snapshot_files`, `_cleanup_partial_files`
2. added a default option for merging the files -> `MERGE_OUTPUT`

## Checklist:

- [ ] tested changes locally

This PR fixes #17136

### DIFF
--- a/bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
+++ b/bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
@@ -50,7 +50,7 @@ class SnapshotResult(object):
         return another distributed RDataFrame build from a Spark backend
         instance. And so on for all other DistRDF backends.
 
-        If MERGE_OUTPUT is True, it will merge all partial fils into a single ouptut file
+        If MERGE_OUTPUT is True, it will merge all partial files into a single output file
         """
         snapshot_chain = ROOT.TChain(self.treename)
         # Add partial snapshot files to the chain
@@ -98,7 +98,7 @@ class SnapshotResult(object):
         output_tree = Chai.CloneTree(-1, "fast")
         if not output_tree:
             print("Error: Failed to clone tree")
-            #output_file.close()
+            output_file.close()
             return ""
         print(f"Writing tree with {output_tree.GetEntries()} entries")
         output_tree.Write()
--- a/bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
+++ b/bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import Union, List, TYPE_CHECKING
 
 import ROOT
+import os
 from ROOT._pythonization._rdataframe import AsNumpyResult
 
 if TYPE_CHECKING:
@@ -14,6 +15,8 @@ class SnapshotResult(object):
     Encapsulate information coming from a Snapshot operation and know how to
     merge it with other objects of this type.
     """
+
+    MERGE_OUTPUT = True
 
     def __init__(self, treename: str, filenames: List[str], resultptr: ROOT.RDF.RResultPtr = None) -> None:
         self.treename = treename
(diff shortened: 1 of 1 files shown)


bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
```diff
@@ -46,13 +49,73 @@ def GetValue(self, backend: BaseBackend):
         computation was created via a Spark backend, then this function will
         return another distributed RDataFrame build from a Spark backend
         instance. And so on for all other DistRDF backends.
+
+        If MERGE_OUTPUT is True, it will merge all partial fils into a single ouptut file
```
ferdymercury, line 53:
```suggestion
        If MERGE_OUTPUT is True, it will merge all partial files into a single output file
```

bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
```diff
@@ -0,0 +1,122 @@
...
+    
+    def _get_base_filename(self) -> str:
+        first_file = self.filenames[0]
+        if not first_file.endswith(".root"):
+            return first_file
+        basename = os.path.splitext(first_file)[0] # basically it will remote .root
+        parts = basename.split('_')
+        if parts[-1].isdigit():
+            return '_'.join(parts[:-1]) + '.root'
+        return first_file
```
vepadulano, line 81: I don't understand what is the goal of this function:

* If you want to extract the base name of the file path, there is https://docs.python.org/3/library/os.path.html#os.path.basename
* If you want just the root prefix of the file name, without the extension, you can use https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem

bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
```diff
@@ -0,0 +1,122 @@
...
+    from DistRDF.Backends.Base import BaseBackend
+
+
+class SnapshotResult(object):
+    """
+    Encapsulate information coming from a Snapshot operation and know how to
+    merge it with other objects of this type.
+    """
+
+    MERGE_OUTPUT = True
```
vepadulano, line 19: This is not acceptable for merging, we cannot have a global setting for all distributed Snapshot calls

bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
```diff
@@ -0,0 +1,122 @@
...
+        This is done by constructing a TChain with the name and the list of
+        paths stored in this object. The chain is then passed to the
+        `make_dataframe` function that changes depending on the backend.
+
+        For example, if the original RDataFrame that triggered the distributed
+        computation was created via a Spark backend, then this function will
+        return another distributed RDataFrame build from a Spark backend
+        instance. And so on for all other DistRDF backends.
+
+        If MERGE_OUTPUT is True, it will merge all partial files into a single output file
```
vepadulano, line 53: This is not an acceptable option for users. One would need to probably pythonize the `Snapshot` function so that for distributed RDataFrame only an extra keyword argument can be passed.

bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
```diff
@@ -0,0 +1,122 @@
...
+        if not first_file.endswith(".root"):
+            return first_file
+        basename = os.path.splitext(first_file)[0] # basically it will remote .root
+        parts = basename.split('_')
+        if parts[-1].isdigit():
+            return '_'.join(parts[:-1]) + '.root'
+        return first_file
+    
+    def _merge_snapshot_files(self, output_path: str) -> str:
+        print(f"Merging {len(self.filenames)} files into {output_path}")
```
vepadulano, line 84: print statements like this are not acceptable in final production code. In this situation logging is not even warranted.

bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
```diff
@@ -0,0 +1,122 @@
...
+                print(f"Adding file: {filename}")
+                chain.Add(filename)
+            else:
+                print(f"Warning: file {filename} does not exist")
+        output_file = ROOT.TFile(output_path, "RECREATE")
+        if not output_file or output_file.IsZombie():
+            print(f"Error: could not create output file {output_path}")
+            return ""
+        print(f"Cloning tree to {output_path}")
+        output_tree = Chai.CloneTree(-1, "fast")
```
vepadulano, line 98: What is `Chai`?

vepadulano, line 98: CloneTree is not the right way to go here. Most probably we would need to call `hadd` in a tree-reduction pattern in the distributed workers.

bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
```diff
@@ -0,0 +1,122 @@
...
+
+        if SnapshotResult.MERGE_OUTPUT and len(self.filenames) > 1:
+            output_path = self._get_base_filename()
+            merged_file_path = self._merge_snapshot_files(output_path)
+            if os.path.exists(merged_file_path):
+                self._cleanup_partial_files()
+            merged_chain = ROOT.TChain(self.treename)
+            merged_chain.Add(merged_file_path)
+            self.filenames = [merged_file_path]
+            return backend.make_dataframe(merged_chain)
```
vepadulano, line 68: This logic is not enough, it does not take into account tree reduction and it forces the creation of the output TChain at creation time of the SnapshotResult object. Furthermore, it does not even take into account the possibility of RNTuple Snapshot

vepadulano (changes requested):
Dear @Aditya-138-12 ,

Thanks for your contribution! At the moment these changes are very far from being ready, I have added a few comments to try to guide the developments in a better direction.

vepadulano:
Dear @Aditya-138-12 ,

Thank you for the interest shown in the ROOT project and thanks for your contribution! Given that my previous review comments haven't been addressed in a while, I was wondering if you're still interested in continuing this PR? No worries if not, we'll just close it for the moment and it could be revisited/reopened at a later stage.

Aditya-138-12:
Dear @vepadulano 
I am really interested to work on this PR and I am actively doing too. But due to some reasons the progress is really slow. Shall I for sometime convert this PR into a draft and then open again.
I forgot to address your comments, will do it asap. Thank you!

bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
Aditya-138-12: it was supposed to be `chain`, I missed it

bindings/experimental/distrdf/python/DistRDF/PythonMergeables.py
Aditya-138-12: But isn't it a good practice to let the user know what is going on ...
Would love to know more, does it affect the performance of the software?

PR #18423, opened 2025-04-16
[python] Improve CPyCppyy heuristic for Clone methods

### LINKED ISSUE #16725: Pyroot crashes reading TClonesArray in a TTree
### Check duplicate issues.

- [X] Checked for duplicates

### Description

A pyroot python program crashes when reading a TFile containing a TTree containing a TClonesArray. It appears to crash the second time it accesses the TClonesArray. This has started happening since moving to ROOT 6.32 on AlmaLinux 9.4. It worked fine with 6.20.04 running on Centos7. The library for accessing the TTree information was created using MakeProject. 

### Reproducer

This is the python code that crashes
```
#!/usr/bin/env python3
import os
import ROOT
import argparse

parser = argparse.ArgumentParser(description='Analyse events.')
parser.add_argument("inputfile")
parser.add_argument("-o","--outputfile",default="Output.root",help="Name of output file")
args = parser.parse_args()

READEVENTANALYSISLIBRARY=ROOT.gSystem.Getenv("READEVENTANALYSISLIBRARY");
ROOT.gSystem.Load(READEVENTANALYSISLIBRARY);

globalRecon = ROOT.TChain("ReconDir/Global")
globalRecon.AddFile(args.inputfile)

OutputFile = ROOT.TFile(args.outputfile,"RECREATE","ND280 Analysis")
hTrackMomenta = ROOT.TH1F("TrackMomenta", "Charged Track Momentum", 100, 0, 1000.0)

entries =  globalRecon.GetEntries()
for entry in range(entries):
	print(" entry = "+str(entry))
	ientry = globalRecon.LoadTree(entry)
	if ientry < 0:
		break
	nb = globalRecon.GetEntry(entry)
	if nb<=0:
		continue
	
	for	pid in globalRecon.PIDs :
		if(pid.Charge != 0) :
			print(" charge is "+str(pid.Charge))
		#	hTrackMomenta.Fill(pid.FrontMomentum)
OutputFile.Write()
OutputFile.Close()
```

The crash traceback suggest this as the line causing the crash:

#7  0x00007fe963a68e9a in TBranchElement::ReadLeavesClones(TBuffer&) () from /home/aleph/ajf/t2k/root/lib/libTree.so
#8  0x00007fe963a60899 in TBranch::GetEntry(long long, int) () from /home/aleph/ajf/t2k/root/lib/libTree.so
#9  0x00007fe963a7326b in TBranchElement::GetEntry(long long, int) () from /home/aleph/ajf/t2k/root/lib/libTree.so
#10 0x00007fe963ad7300 in TTree::GetEntry(long long, int) () from /home/aleph/ajf/t2k/root/lib/libTree.so

### ROOT version

6.32.00  

### Installation method

prebuilt binary root_v6.32.00.Linux-almalinux9.4-x86_64-gcc11.4.tar.gz

### Operating system

AlmaLinux 9.4

### Additional context

_No response_

vepadulano:
The heuristic checks if the token "Clone" is present in the method name being called and gives Python the ownership of the returned object. This is valid behaviour for methods such as `TObject::Clone` where the ownership of the returned value is given to the caller. But the `CPPOverload::Set` method gets as the method name being called the actual function name concatenated with the full template argument list. For example, given the function

```
template <typename T>
void foo();
```

When it passes through this method its name will be `foo<T>`. This behaviour has an undesired side-effect. When `T` is a type with the token `Clone` in its name, then the heuristic will trigger and Python will take ownership of the return value. This is the root cause of the problem seen at https://github.com/root-project/root/issues/16725, where `T` is `TClonesArray`.

This commit improves on the heuristic by stripping the template argument list from the method name before checking for the presence of the token.

Fixes #16725 
Hopefully supersedes #18416

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/CPPOverload.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/CPPOverload.cxx
@@ -1072,10 +1072,19 @@ void CPyCppyy::CPPOverload::Set(const std::string& name, std::vector<PyCallable*
     if (name == "__init__")
         fMethodInfo->fFlags |= (CallContext::kIsCreator | CallContext::kIsConstructor);
 
-// special case, in heuristics mode also tag *Clone* methods as creators
-    if (CallContext::sMemoryPolicy == CallContext::kUseHeuristics && \
-            name.find("Clone") != std::string::npos)
-        fMethodInfo->fFlags |= CallContext::kIsCreator;
+// special case, in heuristics mode also tag *Clone* methods as creators. Only
+// check that Clone is present in the method name, not in the template argument
+// list.
+    if (CallContext::sMemoryPolicy == CallContext::kUseHeuristics) {
+        std::string_view name_maybe_template = name;
+        auto begin_template = name_maybe_template.find_first_of('<');
+        if (begin_template <= name_maybe_template.size()) {
+            name_maybe_template = name_maybe_template.substr(0, begin_template);
+        }
+        if (name_maybe_template.find("Clone") != std::string_view::npos) {
+            fMethodInfo->fFlags |= CallContext::kIsCreator;
+        }
+    }
 
 #if PY_VERSION_HEX >= 0x03080000
     fVectorCall = (vectorcallfunc)mp_vectorcall;


bindings/pyroot/cppyy/CPyCppyy/src/CPPOverload.cxx
```diff
@@ -1072,10 +1072,19 @@ void CPyCppyy::CPPOverload::Set(const std::string& name, std::vector<PyCallable*
...
 
-// special case, in heuristics mode also tag *Clone* methods as creators
-    if (CallContext::sMemoryPolicy == CallContext::kUseHeuristics && \
-            name.find("Clone") != std::string::npos)
-        fMethodInfo->fFlags |= CallContext::kIsCreator;
+    // special case, in heuristics mode also tag *Clone* methods as creators. Only
+    // check that Clone is present in the method name, not in the template argument
+    // list.
+    if (CallContext::sMemoryPolicy == CallContext::kUseHeuristics) {
+       auto name_maybe_template = name;
```
guitargeek, line 1079: Is there a way to search in the substring without copying the string? Algorithmically, that should be possible. Just for the sake of performance

vepadulano: I can make it a `string_view`, but a copy needs to happen anyway to pass the right string to the `find` method

guitargeek: Well, now you're using `std::string_view::find`, which is what I meant. Sorry for not being clear before!

guitargeek (approved):
Thank you very much for making this PR and explaining the problem in the commit message!

PR #18424, opened 2025-04-16
[tmva][sofie] shape and type inference methods need not be pure virtual

sanjibansg:
This PR modifies the ShapeInference method of the base ROperators class that is not needed to be pure virtual since not every ROperator will not be implementing or using it, like the Shape Operator.


lmoneta (approved):
LGTM!
Thank you Sanjiban for the fix!

PR #18435, opened 2025-04-17
[gui] Some fixes to TGNumberEntry

### LINKED ISSUE #17334: TGNumberEntry string length checks are inaccurate/dangerous.
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

As seen in https://github.com/root-project/root/pull/16913, there is some attempt to guarantee that there will be no write past the end of buffer in the string manipulations.  However in several places, it fall short (literally or maybe missing documentation).

We should consider replacing the fixed size buffer or improving the bound checks.  

Namely the routines seems to assume that the buffer has a fixed length of 256 but in several place, the buffer is offset compare to its actual beginning.

`StrInt(char *text, Long_t i, Int_t digits)` hard codes the length 250 for its input buffer, we should pass the actual length left there.  In particular line `TGNumberEntry.cxx:310` and `TGNumberEntry.cxx:316` needs to be updated.

We should also review the rest `TGNumberEntry.cxx` for similar problematic patterns. 

### Reproducer

.

### ROOT version

master

### Installation method

.

### Operating system

.

### Additional context

_No response_

silverweed:
An attempt to fix and improve some old code in `TGNumberEntry` by:

- fixing some broken logic
- simplifying some functions
- improving safety by adding bounds checks where appropriate

This is not an exhaustive pass on the file, as there are many other places that could use a pass or a rewrite. However it should address some of the more glaring safety concerns - particularly, it removes all uses of `strcpy` and the coverity annotations.

I haven't tested this code yet so I'm opening this as a draft PR. Feel free to comment on it already.

## Checklist:

- [x] tested changes locally

This PR fixes #17334

### DIFF
--- a/gui/gui/src/TGNumberEntry.cxx
+++ b/gui/gui/src/TGNumberEntry.cxx
@@ -233,19 +233,20 @@ static Bool_t IsGoodChar(char c, TGNumberFormat::EStyle style,
 
 ////////////////////////////////////////////////////////////////////////////////
 
-static char *EliminateGarbage(char *text,
-                              TGNumberFormat::EStyle style,
-                              TGNumberFormat::EAttribute attr)
+static void CopyAndEliminateGarbage(char *dst,
+                                    std::size_t dstCap,
+                                    const char *src,
+                                    TGNumberFormat::EStyle style,
+                                    TGNumberFormat::EAttribute attr)
 {
-   if (text == 0) {
-      return 0;
-   }
-   for (Int_t i = strlen(text) - 1; i >= 0; i--) {
-      if (!IsGoodChar(text[i], style, attr)) {
-         memmove(text + i, text + i + 1, strlen(text) - i);
+   std::size_t dstIdx = 0;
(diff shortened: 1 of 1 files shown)


ferdymercury:
Thanks a lot for this initiative. If you want a reproducer to test against: https://igit.ific.uv.es/ferhue/pulse-surfer

silverweed:
I tested on @ferdymercury 's codebase and it seems to work fine.

dpiparo (approved):
thanks for this. LGTM.

ferdymercury:
@silverweed I am seeing a regression in TGNumberEntry, that might be connected with this PR.

In 6.36.00
root --web=off $ROOTSYS/tutorials/geom/building.C

then click on Clipping, Plane, numbers look fine.

In master:
`root --web=off $ROOTSYS/tutorials/visualisation/geom/building.C`

both in first tab and in Clipping, Plane, numbers look weird.
(I am rebuilding to see if sth went wrong with rebasing or so).

PR #18437, opened 2025-04-17
Move distributed RDataFrame out of experimental

vepadulano:
The distributed RDataFrame Python package has been moved out of experimental
mode and it is now called `ROOT.RDF.Distributed`. The functionality has already
been merged with the traditional dataframe constructor so that users can type
`ROOT.RDataFrame(..., executor=...)` to get a distributed RDataFrame. For the
moment, we keep the old naming with the Experimental keyword and raise a warning
to the user when they access the module through the old name. This will be
completely removed in ROOT 6.38.

Sibling roottest PR at https://github.com/root-project/roottest/pull/1307


martamaja10 (approved):
LGTM! Thank you Vincenzo! Very nice to see RDF out of the Experimental namespace :)

PR #18438, opened 2025-04-17
[ntuple] add test for merging projected fields with diff sources

silverweed:
Add a test to verify that the merger properly refuses to merge projected fields with a different source field.

## Checklist:

- [x] tested changes locally


enirolf (approved):
LGTM! Out of curiosity, was there a particular reason to add this test other than covering more cases?

silverweed:
> as there a particular reason to add this test other than covering more cases?

It was in my backlog for a while but I never got to it until now. It should hopefully conclude the list of "reasonable use cases" related to projected fields for the merger.

PR #18440, opened 2025-04-17
[base] Move RFloat16.hxx out of ROOT7

silverweed:
Since RFloat16 is required by RNTuple, it too must be now included unconditionally after https://github.com/root-project/root/pull/18410

## Checklist:

- [x] tested changes locally


pcanal (approved):
Thanks.

Issue #17698, opened 2025-02-12
build: Allow building ROOT using the Clang shared library


jakeforster:
### Explain what you would like to see improved and how.

Hello gentle maintainers

My understanding is PR https://github.com/root-project/root/pull/15563 added support for building Cling as a standalone package using shared libraries from LLVM and Clang.  This was facilitated by adopting the new style for embedding LLVM in a project (see https://llvm.org/docs/CMake.html#embedding-llvm-in-your-project).

Similarly, does ROOT's build system allow ROOT to be built using shared libraries from LLVM and Clang?  If not, would you be interested in modifying it to support this?

As discussed in that previous PR, there is an advantage to allowing these builds to be done separately, and I assume this modification could be done such that ROOT can still be built with the bundled LLVM and Clang when the builtin options are passed.

For reference, my attempt to build ROOT with external LLVM, Clang and Cling fails at the 'rootcling_stage1' target, which I'm assuming is due to the expectation of building using static libraries (though I'd be happy to be wrong):

```
ld: /gnu/store/vy8c11s0svpy3hy6g31a704f23cllhj1-cling-1.1/lib/libclingInterpreter.a(LookupHelper.cpp.o):(.text._ZN5clingL18ParseWithShortcutsEPN5clang11DeclContextERNS0_10ASTContextEN4llvm9StringRefEPNS_11InterpreterERNS0_13UnqualifiedIdENS_12LookupHelper11DiagSettingERNS_15ParserStateRAIIE+0x4e5): undefined reference to `clang::Sema::RequireCompleteDeclContext(clang::CXXScopeSpec&, clang::DeclContext*)'
ld: /gnu/store/vy8c11s0svpy3hy6g31a704f23cllhj1-cling-1.1/lib/libclingInterpreter.a(LookupHelper.cpp.o): in function `cling::LookupHelper::findScope(llvm::StringRef, cling::LookupHelper::DiagSetting, clang::Type const**, bool) const':
(.text._ZNK5cling12LookupHelper9findScopeEN4llvm9StringRefENS0_11DiagSettingEPPKN5clang4TypeEb+0x12b9): undefined reference to `clang::Sema::RequireCompleteDeclContext(clang::CXXScopeSpec&, clang::DeclContext*)'
collect2: error: ld returned 1 exit status
make[2]: *** [core/rootcling_stage1/CMakeFiles/rootcling_stage1.dir/build.make:167: core/rootcling_stage1/src/rootcling_stage1] Error 1
```

[...]

PR #18446, opened 2025-04-18
Revert PR #18386

hahnjo:
This reverts commit https://github.com/root-project/root/commit/02b0dc129d0c62e6bf7252bb89e8ec0777a40640. It was partially incorrect and should not use LLVM_DIR, which is only a hint specified by the user.   We need more time to decide on and test a better solution.

This reverts commit https://github.com/root-project/root/commit/3ef1193b0bd3d937e87815d0dfdc12370f8d190d. The
taken approach is not the right direction, this should be handled on the linking side of libCling as discussed post-merge in the pull request:  https://github.com/root-project/root/pull/18386

### DIFF
--- a/core/clingutils/CMakeLists.txt
+++ b/core/clingutils/CMakeLists.txt
@@ -80,19 +80,31 @@ foreach(dict ${stldicts})
   target_include_directories(${dict}Dict PRIVATE ${CMAKE_SOURCE_DIR}/interpreter/cling/include/cling/cint)
 endforeach()
 
+set(CLANG_RESOURCE_DIR_STEM)
 if (builtin_clang)
-    set(CLANG_INSTALL_PREFIX ${CMAKE_BINARY_DIR}/interpreter/llvm-project/llvm/${CMAKE_CFG_INTDIR})
+  set(CLANG_RESOURCE_DIR_STEM ${CMAKE_BINARY_DIR}/interpreter/llvm-project/llvm/${CMAKE_CFG_INTDIR}/lib/clang)
+  set(CLANG_RESOURCE_DIR_VERSION ${LLVM_VERSION_MAJOR})
 else ()
-  # If no explicit Clang prefix is set, assume it lives inside LLVM prefix.
-  if(NOT DEFINED CLANG_INSTALL_PREFIX)
-    set(CLANG_INSTALL_PREFIX ${LLVM_DIR})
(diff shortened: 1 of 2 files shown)


pcanal (approved):

dpiparo:
Let's integrate this, and then we can discuss with calm in May how to proceed: we have all the elements on the table.

guitargeek:
Please give me an alternative then how I can build with `builtin_clang=off`, using LLVM from the packager and without changing the ROOT code. The saved build time was so convenient that I don't want to go back.

Also, what was wrong with: https://github.com/root-project/root/pull/18407

hahnjo:
@pcanal while I appreciate you editing the PR summary and adding more details (which were already available as commit messages), please don't put statements that I did not make.

hahnjo:
> For when we discuss this in May: please give an alternative then how I can build with `builtin_clang=off`, using LLVM from the packager and without changing the ROOT code. The saved build time was so convenient that I don't want to go back....

This works today, if you build without `LLVM_LINK_LLVM_DYLIB`.

> Also, what was wrong with: #18407

If you want others to review a PR, you should add them as reviewers... Fundamentally, the additional PR doesn't address the issue that we are changing the build system just before branching.

guitargeek:
> This works today, if you build without LLVM_LINK_LLVM_DYLIB


I don't want to rebuild LLVM, I want to use the LLVM from my Linux distro which is built with LLVM_DYLIB!

pcanal:
> which were already available as commit messages

Some reviewer do not have a commit focused workflow and those cases it is much more convenient to also express the change in the PR in the summary even if it is as simple as a copy/pasting of the commit message.

In my opinion, ideally the the summary should also include also a higher level view, in particular for reverts, a general statement of purpose for the reversion and upcoming plan is helpful (i.e. it is useful to know difference between not needed, needs small revision, needs large redesign, needs discussion or just wait and resubmit later as is)

> please don't put statements that I did not make.

Absolutely; Thank you for correcting the text.
I thought I was paraphrasing something I understood you meant in related comments
But I agree, it would have been much better for me to add those are separate comments.

guitargeek:
My fix that was reverted was also needed because of the added libInterOp, which is new in 6.36.

See:
https://github.com/root-project/root/issues/17698#issuecomment-2750290854

I think this revert has left the release branch in a worse state.

PR #18451, opened 2025-04-21
[ntuple] fix schema evolution with streamer fields

jblomer:


pcanal:
What do you mean by "fix de-s11n of streamer info records " ?

tree/ntuple/test/rfield_streamer.cxx
```diff
@@ -305,3 +305,53 @@ TEST(RField, StreamerMergeIncremental)
...
+
+   ASSERT_EQ(1U, reader->GetNEntries());
+   auto ptrF = reader->GetModel().GetDefaultEntry().GetPtr<TemperatureKelvin>("f");
+
+   // TODO(jblomer): this should fail with an exception when we connect the page source
+   ROOT::TestSupport::CheckDiagsRAII diagRAII;
+   diagRAII.requiredDiag(kError, "TBufferFile::ReadVersion", "Could not find the StreamerInfo with a checksum of",
+                         false /* matchFullMessage */);
+   diagRAII.requiredDiag(kError, "TBufferFile::CheckByteCount", "object of class TemperatureKelvin read too few bytes",
+                         false /* matchFullMessage */);
```
pcanal, line 355: I don't understand those error message.  What is the case being tested here?

tree/ntuple/test/ntuple_evolution.cxx
```diff
@@ -863,3 +863,61 @@ struct RenamedIntermediateDerived : public RenamedIntermediate2 {
...
+      writer->Fill();
+
+      // Reset / close the writer and flush the file.
+      writer.reset();
+   });
+
+   ASSERT_TRUE(gInterpreter->Declare(R"(
+struct StreamerField {
+   int fInt = 0;
+   int fAdded = 137;
```
pcanal, line 903: To test the indirect collection of StreamerInfo, there should be here a pointer to another class (possibly initialized with an instance of a derived class).

tree/ntuple/test/rfield_streamer.cxx
jblomer: I write a `TemperatureCelsius` and then try to read it as `TemperatureKelvin`.

tree/ntuple/test/ntuple_evolution.cxx
jblomer: I think we already test that with https://github.com/root-project/root/blob/031f6bd8789ec363ed5552531657865b6296d2fe/tree/ntuple/test/rfield_streamer.cxx#L126

jblomer:
> What do you mean by "fix de-s11n of streamer info records " ?

Two bugs are fixed in `RNTupleSerializer::DeserializeStreamerInfos`

tree/ntuple/test/ntuple_evolution.cxx
pcanal: Technically the test linked will succeed whether or not the `StreamerInfo` are stored ... since they are still in memory from the already loaded dictionary.

tree/ntuple/test/rfield_streamer.cxx
pcanal: Humm ... why isn't the error message closer to "Can not convert (or read) a `TemperatureCelsius` (in)to a `TemperatureKelvin`" ?   Without the corresponding checks  the failure could be 'serious'.  If I understood correctly, the user could have a class `A` with `version 2` which contains a `long` then request the reading a class `B` with also `version 2` which contains a pointer the the I/O would succeed but not apply any check nor conversion and thus have a random number stored in the pointer.  Did I miss something?

tree/ntuple/inc/ROOT/RPageStorage.hxx
```diff
@@ -612,6 +612,7 @@ private:
    REntryRange fEntryRange;    ///< Used by the cluster pool to prevent reading beyond the given range
    bool fHasStructure = false; ///< Set to true once `LoadStructure()` is called
    bool fIsAttached = false;   ///< Set to true once `Attach()` is called
+   bool fHasStreamerInfosRegistered = false; ///< Set to true when ReisterStreamerInfos() is called.
```
hahnjo, line 615:
```suggestion
   bool fHasStreamerInfosRegistered = false; ///< Set to true when RegisterStreamerInfos() is called.
```

tree/ntuple/inc/ROOT/RPageStorage.hxx
```diff
@@ -818,6 +819,10 @@ public:
    // TODO(gparolini): for symmetry with SealPage(), we should either make this private or SealPage() public.
    RResult<ROOT::Internal::RPage>
    UnsealPage(const RSealedPage &sealedPage, const ROOT::Internal::RColumnElementBase &element);
+
+   // Builds the streamer info records from the descriptor's extra type info section. This is necessary when
+   // connecting streamer fields so that emulated classes can be read.
```
hahnjo, line 824:
```suggestion
   /// Builds the streamer info records from the descriptor's extra type info section. This is necessary when
   /// connecting streamer fields so that emulated classes can be read.
```

tree/ntuple/test/ntuple_evolution.cxx
```diff
@@ -863,3 +863,61 @@ struct RenamedIntermediateDerived : public RenamedIntermediate2 {
       EXPECT_THAT(err.what(), testing::HasSubstr("incompatible type name for field"));
    }
 }
+
+TEST(RNTupleEvolution, StreamerField)
+{
+   FileRaii fileGuard("test_ntuple_evolution_streamer_field.root");
+   fileGuard.PreserveFile();
```
hahnjo, line 870: This should not be needed, probably a left-over from debugging

tree/ntuple/test/rfield_streamer.cxx
hahnjo: I think a proper exception is part of the `TODO` written above. I don't understand your example, how is this different from the case tested here that even has compatible class layouts?

jblomer: I guess then we are not using the `TClass` / `TBufferFile` optimally. Currently, when we read the data from a streamer field, we do
```
TBufferFile buffer(TBuffer::kRead, nbytes);
// Fill buffer.Buffer() with nbytes of data
fClass->Streamer(to, buffer);
```

Now, if the `TClass` instance pointed to by `fClass` is not compatible with the streamed data on disk, we get the error messages tested for in the test. How can I make `TClass` report a better error message?

pcanal: > Currently, when we read the data from a streamer field, we do

The check is meant to be done before hand.   RNTuple/TTree know what is stored in the column/branch and what the user is passing and needs to check compatibility of the two.  For TTree, this is done in `Tree::CheckBranchAddressType`, in particular the lines:
```
       ptrClass->GetSchemaRules()->HasRuleWithSourceClass( expectedClass->GetName() ) ) {
```
and (for collection)
```
             inmemValueClass->GetSchemaRules()->HasRuleWithSourceClass(onfileValueClass->GetName() ) )
```

jblomer: Ok. This doesn't seem to be a one-liner and it is a different topic than the one addressed in this PR. This PR is supposed to make the streamer field work if the types _are_ compatible. For better handling of incompatible types, I created https://github.com/root-project/root/issues/18611

jblomer:
Is there anything else missing for this PR to be merged?

tree/ntuple/test/rfield_streamer.cxx
pcanal: Fair enough.  This PR is a step forward.   As a side note, in case the class are versions and the on-file class and the unrelated class (in case of user error) have the same class versions, we would not see the message `Could not find the StreamerInfo with a checksum of` but instead the streaming will proceed with the wrong StreamerInfo with a set of possible outcomes: best case scenario we see the `read too few bytes` or `read too many bytes` output, in other cases (if one class layout has pointers in different places (and/or missing/added) than the other class layout) then we could get a crashes and in the 'worst' case scenario the 2 layout have the same number of bytes in their onfile representations and we get a silent failure (wrong data).

Issue #18453, opened 2025-04-21
Some test fails on linux with Davix disabled


pcanal:
The following tests:
```
62:gtest-io-io-TFile
1503:roottest-python-cmdLineUtils-WebRootls1
1504:roottest-python-cmdLineUtils-WebRootls2
1963:roottest-root-io-webfile-make
```
fail (for me) on alma9 with Davix support disable.  The error are similar to:
```
+Error in <TWebFile::TWebFile>: http://root.cern.ch:80/files/na49.root? does not exist
+failed to open http://root.cern.ch/files/na49.root
```

Issue #8890, opened 2021-08-24
Error 400 while creating a `TWebFile`


vepadulano:
### Describe the bug
With both 6.24/02 and master the following snippet outputs an error:
```cpp
#include <TWebFile.h>

int main(){
    TWebFile f{"https://github.com/root-project/root/raw/master/tutorials/dataframe/df017_vecOpsHEP.root"};    
}
```
```bash
$: g++ -g -o twebfile_open_github twebfile_open_github.cpp `root-config --cflags --glibs`
$: ./twebfile_open_github 
Error in <TWebFile::GetFromWeb10>: https://github.com:443/root-project/root/raw/master/tutorials/dataframe/df017_vecOpsHEP.root?: Bad Request (400)
```
The error is triggered [here](https://github.com/root-project/root/blob/5d536b29d7ce437aad79ff046788735a33fb1dbe/net/net/src/TWebFile.cxx#L866). Curiously, with 6.24 the equivalent `TFile::Open` doesn't error
```cpp
$: root
   ------------------------------------------------------------------
  | Welcome to ROOT 6.24/02                        https://root.cern |
  | (c) 1995-2021, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Jul 03 2021, 08:02:00                 |
  | From tag , 28 June 2021                                          |
  | With                                                             |
  | Try '.help', '.demo', '.license', '.credits', '.quit'/'.q'       |
   ------------------------------------------------------------------

root [0] auto f = TFile::Open("https://github.com/root-project/root/raw/master/tutorials/dataframe/df017_vecOpsHEP.root");
root [1] f->GetName()
(const char *) "https://github.com/root-project/root/raw/master/tutorials/dataframe/df017_vecOpsHEP.root"
```

But it still errors with the latest master@59e754a
```cpp
$: root
   ------------------------------------------------------------------
  | Welcome to ROOT 6.25/01                        https://root.cern |
  | (c) 1995-2021, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Aug 23 2021, 13:49:23                 |
  | From heads/master@v6-25-01-1831-g61575a3f47                      |
  | With c++ (GCC) 11.2.1 20210728 (Red Hat 11.2.1-1)                |
  | Try '.help', '.demo', '.license', '.credits', '.quit'/'.q'       |
   ------------------------------------------------------------------

[...]

PR #18458, opened 2025-04-22
[test] disable web tests if davix is OFF

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/18453

Sibling PR: https://github.com/root-project/roottest/pull/1308

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/io/io/test/CMakeLists.txt
+++ b/io/io/test/CMakeLists.txt
@@ -6,9 +6,6 @@
 
 ROOT_ADD_GTEST(RRawFile RRawFile.cxx LIBRARIES RIO)
 ROOT_ADD_GTEST(TFile TFileTests.cxx LIBRARIES RIO)
-if (davix)
-  target_compile_definitions(TFile PRIVATE R__DAVIX_SUPPORT)
-endif()
 ROOT_ADD_GTEST(TBufferFile TBufferFileTests.cxx LIBRARIES RIO)
 ROOT_ADD_GTEST(TBufferMerger TBufferMerger.cxx LIBRARIES RIO Imt Tree)
 ROOT_ADD_GTEST(TBufferJSON TBufferJSONTests.cxx LIBRARIES RIO)
--- a/io/io/test/CMakeLists.txt
+++ b/io/io/test/CMakeLists.txt
@@ -6,6 +6,9 @@
 
 ROOT_ADD_GTEST(RRawFile RRawFile.cxx LIBRARIES RIO)
 ROOT_ADD_GTEST(TFile TFileTests.cxx LIBRARIES RIO)
+if (davix)
+  target_compile_definitions(TFile PRIVATE R__DAVIX_SUPPORT)
+endif()
 ROOT_ADD_GTEST(TBufferFile TBufferFileTests.cxx LIBRARIES RIO)
 ROOT_ADD_GTEST(TBufferMerger TBufferMerger.cxx LIBRARIES RIO Imt Tree)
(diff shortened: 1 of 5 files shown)


io/io/test/CMakeLists.txt
```diff
@@ -6,6 +6,9 @@
 
 ROOT_ADD_GTEST(RRawFile RRawFile.cxx LIBRARIES RIO)
 ROOT_ADD_GTEST(TFile TFileTests.cxx LIBRARIES RIO)
+if (davix)
```
dpiparo, line 9: This is not strictly necessary: the preprocessor macro `R__HAS_DAVIX` is already available

io/io/test/TFileTests.cxx
```diff
@@ -130,6 +130,7 @@ void TestReadWithoutGlobalRegistrationIfPossible(const char *fname)
 // https://github.com/root-project/root/issues/10742
 #ifndef R__WIN32
 // We prefer not to read remotely files from Windows, if possible
+#ifdef R__DAVIX_SUPPORT
```
dpiparo, line 133: This is indeed the solution, however I would propose we go for the existing macro `R__HAS_DAVIX`

dpiparo (changes requested):
Thanks a lot for this PR! It is correct logically, however one can trim down the changes by using the already available preprocessor macro `R__HAS_DAVIX` in the C++ code and be done with it.

io/io/test/TFileTests.cxx
```diff
@@ -130,6 +130,7 @@ void TestReadWithoutGlobalRegistrationIfPossible(const char *fname)
 // https://github.com/root-project/root/issues/10742
 #ifndef R__WIN32
 // We prefer not to read remotely files from Windows, if possible
+#ifdef R__HAS_DAVIX
```
pcanal, line 133: This is effective.  But do we understand why the test fail?  For example, is there a missing feature in the `TWebFile` or are the test "really" needing `Davix`?

linev: `TWebFile` does not support HTTP protocol 2.0, only 1.1.

See my comment here:

https://github.com/root-project/root/issues/8890#issuecomment-911515928

pcanal: This may or may not be the root cause here ... The error message I see:
```
Error in <TWebFile::TWebFile>: http://root.cern.ch/files/na49.root? does not exist
```
where the trailing interrogation mark is suspicious.

ferdymercury: The interrogation mark comes from:

```
fBasicUrl += "?";
fBasicUrl += fUrl.GetOptions();
```

I have added the following guard:

```
      if (strlen(fUrl.GetOptions())) {
         fBasicUrl += "?";
         fBasicUrl += fUrl.GetOptions();
      }
```

and now I get with `TWebFile f{"https://root.cern/files/na49.root"};`:

`Error in <TWebFile::TWebFile>: http://root.cern:443/files/na49.root does not exist`

without the question mark.

The error stems from the ::GetHead function. When calling with message:

`HEAD https://root.cern:443/files/na49.root HTTP/1.0\r\nUser-Agent: ROOT-TWebFile/1.1\r\n\r\n`

one gets as return string: `HTTP/1.0 404 Not Found`

and the 404 return code is forwarded as "-2" to the constructor.

pcanal: Thanks.

dpiparo:
great PR. Thanks.

PR #18460, opened 2025-04-22
[CMake] Get Clang directory from CLANG_INSTALL_PREFIX

guitargeek:
If Clang was installed standalone (in particular, not in the LLVM tree),
use the CLANG_INSTALL_PREFIX variable to indicate where it is.

Actually, we can always use `CLANG_INSTALL_PREFIX`. It's defined by
`ClangConfig.cmake`. We should also need to take into account
`LLVM_LIBDIR_SUFFIX`.

Follows up on:
  * https://github.com/root-project/root/pull/18386
  * https://github.com/root-project/root/pull/18446


hahnjo:
I disagree with merging any build system related work at this point.

guitargeek:
Sure, this can also be done after the branching  of 6.36!

guitargeek:
@hahnjo, what do you think about this for `master`?

core/clingutils/CMakeLists.txt
```diff
@@ -85,7 +85,13 @@ if (builtin_clang)
...
   set(CLANG_RESOURCE_DIR_VERSION ${LLVM_VERSION_MAJOR})
 else ()
-  set(CLANG_RESOURCE_DIR_STEM ${LLVM_LIBRARY_DIR}/clang)
+  # If Clang was installed standalone (in particular, not in the LLVM tree),
+  # use the CLANG_INSTALL_PREFIX variable to indicate where it is.
+  if(DEFINED CLANG_INSTALL_PREFIX)
+    set(CLANG_RESOURCE_DIR_STEM ${CLANG_INSTALL_PREFIX}/lib/clang)
+  else()
+    set(CLANG_RESOURCE_DIR_STEM ${LLVM_LIBRARY_DIR}/clang)
+  endif()
```
hahnjo, line 94: I think we can always use `CLANG_INSTALL_PREFIX`, it's defined by `ClangConfig.cmake`. We should also take into account `LLVM_LIBDIR_SUFFIX` (untested):
```suggestion
  set(CLANG_RESOURCE_DIR_STEM ${CLANG_INSTALL_PREFIX}/lib${LLVM_LIBDIR_SUFFIX}/clang)
```

hahnjo (changes requested):

core/clingutils/CMakeLists.txt
guitargeek: Thanks for the suggestion! I'll test it

hahnjo (changes requested):
Hm wait, `CLANG_INSTALL_PREFIX` is not set for me with `builtin_clang`... Could be that we have to forward it

hahnjo (commented):
I think we need
```diff
diff --git a/interpreter/CMakeLists.txt b/interpreter/CMakeLists.txt
index e6512a88828..014be83dc28 100644
--- a/interpreter/CMakeLists.txt
+++ b/interpreter/CMakeLists.txt
@@ -407,6 +407,9 @@ if (builtin_clang)
 else()
   find_package(Clang REQUIRED CONFIG)
   message(STATUS "Found Clang ${CLANG_PACKAGE_VERSION} in ${CLANG_CMAKE_DIR}")
+
+  # Forward CLANG_INSTALL_PREFIX so it can be used by core/clingutils/CMakeLists.txt
+  set(CLANG_INSTALL_PREFIX ${CLANG_INSTALL_PREFIX} PARENT_SCOPE)
 endif()
 
 # Reset the compiler flags after compiling LLVM and Clang
```
It might have worked by accident before if `/lib/clang/` happened to contain files for the right Clang version...

guitargeek:
> Hm wait, `CLANG_INSTALL_PREFIX` is not set for me with `builtin_clang`... Could be that we have to forward it

But the change is in the `else` branch of the `if (builtin_clang)` so it should be alright, no?

I have tested it for my setup: system LLVM without Clang and then standalone Clang in different directory, and it works

guitargeek:
Ok, I'll check. But I will also make some test with a full LLVM build including patched Clang so I'm more sure to understand all cases. So for me the PR is on hold anyway

core/clingutils/CMakeLists.txt
hahnjo: For `LLVM_LIBDIR_SUFFIX`, we'll also need https://github.com/root-project/root/pull/18567

guitargeek:
Hi @hahnjo! I now tested this PR with two configurations:

1. System LLVM with no Clang on the system, plus standalone build of patched Clang
2. No LLVM on the system, but custom LLVM build with `LLVM_ENABLE_PROJECTS=clang`

Both works! By the way, what do you think about doing `find_package(Clang)` also in `core`, instead of the `set(CLANG_INSTALL_PREFIX ${CLANG_INSTALL_PREFIX} PARENT_SCOPE)`? Setting variables in the parent scope is more confusing IMO :slightly_smiling_face:

hahnjo (approved):

hahnjo:
> By the way, what do you think about doing `find_package(Clang)` also in `core`, instead of the `set(CLANG_INSTALL_PREFIX ${CLANG_INSTALL_PREFIX} PARENT_SCOPE)`? Setting variables in the parent scope is more confusing IMO

Potentially, but we also compute many other variables in `interpreter/CMakeLists.txt` and having them all in one place might be the better choice

PR #18461, opened 2025-04-22
Genvectorx 2.0

mdessole:
# This Pull request:

Adds GenVectorX, a prototype implementation of SYCL support for 4-vectors operations. It is work in progress and is not ready for merge.

## Checklist:
To build, you need the flags:
-Dexperimental_adaptivecpp
-Dexperimental_genvectorx

- [x] tested changes locally


math/experimental/genvectorx/inc/MathX/GenVectorX/AxisAngle.h
```diff
@@ -0,0 +1,371 @@
...
+namespace ROOT_MATH_ARCH {
+
+//__________________________________________________________________________________________
+/**
+   AxisAngle class describing rotation represented with direction axis (3D Vector) and an
+   angle of rotation around that axis.
+
+   @ingroup GenVector
+
+   @sa Overview of the @ref GenVector "physics vector library"
```
ferdymercury: see https://github.com/root-project/root/pull/19110#pullrequestreview-2945154935

hahnjo (commented):
The change in `interpreter/CMakeLists.txt` is a very small subset of https://github.com/root-project/root/pull/17209, so I assume that PR will be merged first

math/experimental/genvectorx/inc/MathX/GenVectorX/AxisAngle.h
mdessole: Thank you @ferdymercury, I totally missed this
Should be all fixed now

cmake/modules/SetupAdaptiveCpp.cmake
```diff
@@ -0,0 +1,116 @@
+message(STATUS "Building AdaptiveCpp for SYCL support.")
+
+include(FetchContent)
+
+if(NOT DEFINED ADAPTIVE_CPP_SOURCE_DIR)
+  FetchContent_Declare(
+    AdaptiveCpp
+    GIT_REPOSITORY https://github.com/root-project/AdaptiveCpp.git
```
vgvassilev: I am a bit confused. ~~If that PR is merged does it mean that ROOT will be fetching ROOT?~~

Replying to self: AdaptiveCpp seems some runtime that we need. Any clue what would it take to depend on llvm only and what would it take to move the minimal set of infrastructure we need of AdaptiveCpp in llvm?

devajithvs: Root will be fetching `AdaptiveCpp` at configure time if `-Dexperimental_adaptivecpp` is enabled. `AdaptiveCpp` upstream uses dynamic LLVM (we need a few patches to get it building statically, so a separate fork. This can potentially be up-streamed)

vgvassilev:
@mdessole, could you update the PR description to elaborate on why this work cannot be done in the GenVector package which already seems to be the ROOT component that's widely used by the community.

mdessole:
> @mdessole, could you update the PR description to elaborate on why this work cannot be done in the GenVector package which already seems to be the ROOT component that's widely used by the community.

Done, hope it's more clear now.

vgvassilev:
> > @mdessole, could you update the PR description to elaborate on why this work cannot be done in the GenVector package which already seems to be the ROOT component that's widely used by the community.
> 
> Done, hope it's more clear now.

Can we include an answer to the question why we decided to not added the support in the GenVector instead? Historically, we managed to retrofit things like explicit simd (vectorization) to it. That is, how is this change different from; why we can't add a bunch of template specializations that enable the feature.

math/experimental/genvectorx/inc/MathX/GenVectorX/3DConversions.h
```diff
@@ -0,0 +1,124 @@
...
+{
+   ERROR_This_Rotation_Conversion_is_NOT_Supported();
+}
+
+// ----------------------------------------------------------------------
+// conversions from Rotation3D
+/**
+   conversion functions from 3D rotation.
+   For converting to Euler Angles see
+   <A><HREF="http://www.cern.ch/mathlibs/documents/eulerAngleComputation.pdf">here</A>
```
ferdymercury: This link is broken

math/experimental/genvectorx/inc/MathX/GenVectorX/3DConversions.h
```diff
@@ -0,0 +1,124 @@
...
+void convert(AxisAngle const &from, EulerAngles &to);
+void convert(AxisAngle const &from, Quaternion &to);
+void convert(AxisAngle const &from, RotationZYX &to);
+
+// ----------------------------------------------------------------------
+// conversions from EulerAngles
+
+void convert(EulerAngles const &from, Rotation3D &to);
+void convert(EulerAngles const &from, AxisAngle &to);
+void convert(EulerAngles const &from, Quaternion &to);
```
ferdymercury, line 72: From my point of view, it would be great that this conversion allows for non-default order of Euler Angles. Ie to mimick this function in MATLAB:
https://www.mathworks.com/help/aerotbx/ug/angle2quat.html
where rotationSequence can be specified if not equal to default.
See also https://github.com/bminortx/Buckshot/blob/master/matlabComponents/Matlab_functions/angle2quat.m

ferdymercury (changes requested):
Thanks!! Some docu-related observations from my side, and a suggestion wrt Euler angles interface

dpiparo (approved):

dpiparo:
> Can we include an answer to the question why we decided to not added the support in the GenVector instead? Historically, we managed to retrofit things like explicit simd (vectorization) to it. That is, how is this change different from; why we can't add a bunch of template specializations that enable the feature.

Of course: this is an experimental addition to ROOT, which can be enabled with the `experimental_genvectorx` CMake option. For now we'll expose this to experiments and users, get feedback and based on adoption, we'll decide how to continue develop and maintain the code, which will be then integrated in GenVector. I hope this helps.

PR #18464, opened 2025-04-22
[clang] Forward TPL of NestedNameSpecifier

### LINKED ISSUE #18363: Incorrect name normalization for template with size_t template argument

hahnjo:
This avoids type suffixes for integer constants when the type can be inferred from the template parameter.

Closes #18363


vgvassilev (approved):
Now that I see the solution I remember that passing the TPL also helped in a few other occasions. I wished I have remembered that sooner. LGTM!

vgvassilev:
I'd prefer if both commits are squashed makes the history clearer in terms of what was fixed and what was broken.

hahnjo:
> I'd prefer if both commits are squashed makes the history clearer in terms of what was fixed and what was broken.

I don't want to do this because we want to backport to at least 6.36 and we still have the separate `roottest` repository there.

vgvassilev:
Probably we can adjust the backport instead?

hahnjo:
Hrm, it's extra work and probably requires me to edit `.patch` files, but fine...

PR #18369, opened 2025-04-11
[ci] Test cpp23 for fedora42, ubu25 and mac-beta

dpiparo:
Revamp of https://github.com/root-project/root/pull/17273

PR #18465, opened 2025-04-22
[core] Add missing includes

devajithvs:
This will fix build errors like:
`error: declaration of 'type_info' must be imported from module 'std.std_typeinfo' before it is required`

in https://github.com/root-project/root/pull/18369 when building with C++23

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes #


pcanal (approved):
Thanks.

PR #18466, opened 2025-04-22
Add check for targets incdirs

rlalik:
# This Pull request:

Adds additional variable check for not being `NOTFOUND` when building dictionary for targets specified with `MODULE`.

Without this, using `ROOT_GENERATE_DICTIONARY` with `MODULE` value results sometimes with:

> CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
> Please set them or make sure they are set and tested correctly in the CMake files:
> /\<some path\>/target_incdirs
>    used as include directory in directory /\<some path\> 

## Checklist:

- [x ] tested changes locally
- [x ] updated the docs (if necessary)


pcanal (approved):
Thanks.

PR #18467, opened 2025-04-22
[cmake] explicit feature=OFF should have hierarchical priority over cached or explicit builtin_feature=ON v2

ferdymercury:
# This Pull request:

## Changes or fixes:

This is v2 of https://github.com/root-project/root/pull/18413

Fixes https://its.cern.ch/jira/browse/ROOT-10743

Related:

https://its.cern.ch/jira/browse/ROOT-9385

## Checklist:

- [ ] tested changes locally


ferdymercury:
@pcanal I have tried to follow your suggestion.

My question now is, how do I prevent enabling fftw3=ON if I passed explicitly -Dfftw3=OFF ?

Right now the sequence
```
cmake -Dbuiltin_fftw3=ON ../root_src/
cmake -Dfftw3=OFF ../root_src/
```

still leaves fftw3 ON.

Do I need to do something like https://stackoverflow.com/a/54302498/7471760 or rather https://stackoverflow.com/a/61986176/7471760?

Sth like: 

```
if (NOT fftw3 and NOT DEFINED CACHE{builtin_fftw3})
   set(fftw3 CACHE BOOL ON FORCE)
endif()
```

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -912,18 +914,25 @@ if(fftw3)
...
     find_package(FFTW)
     if(NOT FFTW_FOUND)
-      if(fail-on-missing)
-        message(FATAL_ERROR "FFTW3 libraries not found and they are required (fftw3 option enabled)")
-      else()
-        message(STATUS "FFTW3 not found. Set [environment] variable FFTW_DIR to point to your FFTW3 installation")
-        message(STATUS "                 Alternatively, you can also enable the option 'builtin_fftw3' to build FFTW3 internally'")
-        message(STATUS "                 For the time being switching OFF 'fftw3' option")
-        set(fftw3 OFF CACHE BOOL "Disabled because FFTW3 not found and builtin_fftw3 disabled (${fftw3_description})" FORCE)
+      set(builtin_fftw3 ON CACHE BOOL "Enabled because FFTW3 not found (${fftw3_description})" FORCE)
```
pcanal, line 917:
```suggestion
      set(builtin_fftw3 ON CACHE BOOL "Enabled because FFTW3 not found (${fftw3_description})")
```

ferdymercury:
> Try without the `FORCE` ...

I tried and it does not solve the issue. Probably because the default option fftw3=OFF, so that has priority over the cache?

So without force -Dbuiltin_fftw3=ON does not turn fftw3=ON.

I tried also NOT DEFINED CACHE, to no avail.

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -642,7 +642,9 @@ if(mathmore OR builtin_gsl OR (tmva-cpu AND use_gsl_cblas))
     # FIXME: one need to find better way to extract path with GSL include files
     set(GSL_INCLUDE_DIR ${CMAKE_BINARY_DIR}/GSL-prefix/src/GSL-build)
     set(GSL_FOUND ON)
-    set(mathmore ON CACHE BOOL "Enabled because builtin_gls requested (${mathmore_description})" FORCE)
+    if(NOT mathmore)
+      set(mathmore ON CACHE BOOL "Enabled because builtin_gls requested (${mathmore_description})")
```
ferdymercury, line 646:
```suggestion
      set(mathmore ON CACHE BOOL "Enabled because builtin_gsl requested (${mathmore_description})")
```

hageboeck:
@ferdymercury, with yours and our recent work on the builtins, can we close this one?

ferdymercury:
> , can we close this one?

Hmmm not really.

This still fails:

cmake -Dbuiltin_fftw3=ON ../root_src/
cmake -Dfftw3=OFF ../root_src/
-- Enabled support for:  asimage asimage_tiff builtin_clang builtin_cling builtin_civetweb builtin_fftw3 builtin_llvm builtin_openui5 builtin_vdt builtin_xxhash check_connection clad curl dataframe **fftw3** gdml geom http imt opengl pyroot roofit root7 runtime_cxxmodules shared spectrum ssl thisroot_scripts tmva tmva-cpu tmva-cudnn tpython use_gsl_cblas vdt webgui x11 xml xrootd


The solution is easy, if the ROOT team agrees to this change of behavior:

-Dbuiltin_feature=ON is ignored if -Dfeature=OFF or if -Dfeature=UNSET (a warning will be issued)

Otherwise a more complex solution is needed. Considering that builtins are supposed to be the 'last exception' and that one should use system packages, I vote for the more simple solution. Less CMake lines of codes to maintain, not worth the complexity for a corner case scenario, that is anyway buggy in the current state.

ferdymercury:
Closing: superseded by https://github.com/root-project/root/pull/23092

In summary, the new PR chooses the philosophy of giving the user a helpful best-educated-guess copy-pastable error message so that, everything that was happening before behind the scenes automatically is converted instead to a visible error message that the user can then decide to copy-paste or not into his command line. That way, responsibility is delegated, behind-the-scenes is avoided, and code maintenance is minimized by avoiding buggy issues with the cache. The user-mediated second-pass input is our human-cache :)
Since 'what the cache was doing before' is copy-pastable into the own terminal, the developer does not lose much time in figuring what builtins have to be enabled or not, etc, compared to having to parse the error messages scattered in many places.
So: there is no hierarchy at all now, user build options are 'sacred' and no one can turn them ON or OFF. Only suggestions will be printed to terminal.

PR #18478, opened 2025-04-23
[hist] Mark input `x` of the declared TFormula function as `const`

guitargeek:
Under the hood, a TFormula declares a function to the interpreter. This function is then used also for differentiating through the TFormula with Clad. However, only the gradient with respect to the parameters `p` is requested, and the upcoming Clad release expects the other input arguments to be marked `const` for optimization purposes.

This will fix the problems that we saw in the draft PR that tries to update the Clad version inside ROOT to the current `master`:
https://github.com/root-project/root/pull/18383


vgvassilev (approved):
Looks like the failures are unrelated. LGTM!

PR #18479, opened 2025-04-23
[CMake] Indicate NumPy dependency for tutorials that make use of it

guitargeek:
This ensures that the tutorials that require NumPy are not run as unit tests if NumPy is not available in the environment.


dpiparo (approved):
Thanks. Should it be backported to 6.36 too perhaps?

PR #18462, opened 2025-04-22
[math] implement genvector custom axis rotation

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://root-forum.cern.ch/t/tvector3-rotate-to-arbitrary-rotation-using-xyzvector/63244

## Checklist:

- [ ] tested changes locally
- [ ] updated the docs (if necessary)

PR #18480, opened 2025-04-23
[math][GenVector] Add access operator for Rotation3D

mdessole:
# This Pull request:

Fixes https://root-forum.cern.ch/t/tvector3-rotate-to-arbitrary-rotation-using-xyzvector/63244

Alternative to https://github.com/root-project/root/pull/18462

## Checklist:

- [x] tested changes locally

### DIFF
--- a/math/genvector/inc/Math/GenVector/Rotation3D.h
+++ b/math/genvector/inc/Math/GenVector/Rotation3D.h
@@ -34,6 +34,7 @@
 #include "Math/GenVector/RotationYfwd.h"
 #include "Math/GenVector/RotationZfwd.h"
 
+#include "Math/GenVector/GenVector_exception.h"
 
 #include <algorithm>
 #include <cassert>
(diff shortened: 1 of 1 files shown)


ferdymercury:
Thanks a lot Monica for this alternative suggestion! If we decide to go this way, could you also implement the helper function:

```
/** copy-doxygen-docu from other pull request, too*/
         template <class Vector>
         Vector Rotate(const Vector & v, double alpha, const Vector & axis) {
            Vector a = v;
            ROOT::Math::VectorUtil::Rotate(a, ROOT::Math::Rotation3D(ROOT::Math::AxisAngle(axis, alpha)));
            return a;
         }
```

as well as 'rescue' the TRotation alpha fmod fix, and the fixed typoes in the documentation from the linked PR, that could be then closed ?

ferdymercury (approved):
Thanks!!

Maybe just one minor comment, you could add \param i, j doxygen docu, to let user know what is rows and what is columns.

PR #18473, opened 2025-04-23
[ntuple] Move RCluster/RClusterPool out of Experimental

silverweed:
- [x] tested changes locally
- [x] updated the docs (if necessary)

PR #18482, opened 2025-04-23
 [6.36][ntuple] Move RCluster/RClusterPool out of Experimental

silverweed:
BP of https://github.com/root-project/root/pull/18473


hahnjo:
Note that `cherry picked from commit` is dangerous here since the hashes will change when rebasing. https://github.com/root-project/root/pull/18473 must be merged to `master` first...

silverweed:
@hahnjo right, I updated the hashes.

PR #18483, opened 2025-04-23
[json] improve roottest/io/json testing

linev:
1. Generate dictionary for custom tested classes
2. Load this library when running test macros
3. Avoid multiple cling workarounds
4. All ref files remains as is

### DIFF
--- a/roottest/root/io/json/CMakeLists.txt
+++ b/roottest/root/io/json/CMakeLists.txt
@@ -1,3 +1,58 @@
 
-ROOTTEST_COMPILE_MACRO(test_classes.h FIXTURES_SETUP json-test-classes-build)
-ROOTTEST_ADD_AUTOMACROS(FIXTURES_REQUIRED json-test-classes-build)
+ROOTTEST_ADD_TEST(ArrayCompress
+                  MACRO runArrayCompress.C
+                  OUTREF ArrayCompress.ref)
+
+ROOTTEST_ADD_TEST(PolyMarker
+                  MACRO runPolyMarker.C
+                  OUTREF PolyMarker.ref)
+
+ROOTTEST_GENERATE_DICTIONARY(libJsonTestClasses
+                  test_classes.h
+                  LINKDEF LinkDef.h
+                  NO_ROOTMAP NO_CXXMODULE)
+
+ROOTTEST_ADD_TEST(BasicTypes
+                  MACRO runBasicTypes.C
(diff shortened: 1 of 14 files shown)


roottest/root/io/json/CMakeLists.txt
```diff
@@ -1,3 +1,58 @@
...
+
+ROOTTEST_GENERATE_DICTIONARY(libJsonTestClasses
+                  test_classes.h
+                  LINKDEF LinkDef.h
+                  NO_ROOTMAP NO_CXXMODULE)
+
+ROOTTEST_ADD_TEST(BasicTypes
+                  MACRO runBasicTypes.C
+                  OUTREF BasicTypes.ref
+                  DEPENDS ${GENERATE_DICTIONARY_TEST})
```
pcanal, line 18: Rather than `depends` we are migrating to using `FIXTURES_SETUP/REQUIRED` which have the added benefit of running the `SETUP` even if you only request to run the actual test.

roottest/root/io/json/runSTL0.C
```diff
@@ -1,41 +1,29 @@
...
 
-#if defined(ClingWorkAroundMissingDynamicScope) && !defined(SECOND_RUN)
-#define SECOND_RUN
-   gROOT->ProcessLine(".x runSTL0.C");
-#else
+void runSTL0()
+{
+   gSystem->Load("libJsonTestClasses");
 
    TJsonEx7 ex7;
```
pcanal, line 7: This is odd.  The compilation by Cling of `runSTL0` happens before the `gSystem->Load` is executed, so it looks like the test is relying on auto-loading of the library instead.  Is that the intent?  Does it still work when ROOT is compiled with `runtime_cxxmodule=OFF`?

linev: I guess, we will see it in CI

linev:
Trying with updates libs for roottest

roottest/root/io/json/runSTL0.C
pcanal: > Library only needed to get generated streamer infos - 

Fair enough, I now understand what you mean.  I see that all the the `TJSon...` classes are 'header' only classes and the library only contains the dictionary (source) files, thus the function's code can be 'compiled' and linked without the library.

pcanal (approved):
LGTM and thanks for your patience.

PR #18459, opened 2025-04-22
[hist] Implement AutoZoom of TH1/TH2/TH3/TAxis to filled range

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/14538

## Checklist:

- [x] tested changes locally
- [x] updated the docs (if necessary)

PR #18487, opened 2025-04-24
[doNOTmerge] check if filesize increase depends on comment title

ferdymercury:
# This Pull request:

## Changes or fixes:

See https://github.com/root-project/root/pull/18459#discussion_r2056563805

### DIFF
--- a/hist/hist/inc/TH1.h
+++ b/hist/hist/inc/TH1.h
@@ -97,7 +97,7 @@ public:
    friend class TH1Merger;
 
 protected:
-    Int_t         fNcells;          ///<  Number of bins (1D) / total of cells (2D) / voxels (3D) +Under/Overflows
+    Int_t         fNcells;          ///<  Number of bins(1D), cells (2D) +U/Overflows
     TAxis         fXaxis;           ///<  X axis descriptor
     TAxis         fYaxis;           ///<  Y axis descriptor
     TAxis         fZaxis;           ///<  Z axis descriptor
--- a/hist/hist/inc/TH1.h
+++ b/hist/hist/inc/TH1.h
@@ -97,7 +97,7 @@ public:
    friend class TH1Merger;
 
 protected:
-    Int_t         fNcells;          ///<  Number of bins(1D), cells (2D) +U/Overflows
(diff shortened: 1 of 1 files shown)


ferdymercury:
No, it does not depend on the number of functions.
Trying now removing things.

hist/hist/inc/TH1.h
```diff
@@ -323,6 +323,7 @@ class TH1 : public TNamed, public TAttLine, public TAttFill, public TAttMarker {
 
    virtual Int_t    GetQuantiles(Int_t n, Double_t *xp, const Double_t *p = nullptr);
    virtual Double_t GetRandom(TRandom * rng = nullptr) const;
+   void             GetRangeOfFilledWeights(const Int_t, Int_t, Int_t, bool) const {};
```
ferdymercury, line 326:
```suggestion
   void             GetRangeOfFilledWeights(Int_t, Int_t, Int_t, bool) const {};
```

ferdymercury:
@linev @pcanal this little change is the one that leads to a change in the compressed file size:

https://github.com/root-project/root/pull/18487/commits/27ae4384d7753a33a2807aa86feb301e81ce1344

linev:
@pcanal 

Is it by chance class checksum which changes after changing in class methods?

ferdymercury:
> @pcanal
> 
> Is it by chance class checksum which changes after changing in class methods?

Note that if I add many other functions the checksum does not change. Only if a parameter "const int" instead of "int" appears.

linev:
For me checksum remains.

linev:
@ferdymercury 
Yes, changed comments can be a reason. They stored with streamer infos in the file.
Probably just modify your original PR and do not touch comments there .

ferdymercury:
> @ferdymercury Yes, changed comments can be a reason. They stored with streamer infos in the file. Probably just modify your original PR and do not touch comments there .

But I need there *MENU* comments anyway, so I will need to still modify the roottests ?

linev:
You can check [here](https://jsroot.gsi.de/latest/?file=https://jsroot.gsi.de/files/hsimple.root&item=StreamerInfo) that is stored in streamer infos of hsimple.root file.

Issue #18477, opened 2025-04-23
TPad::ClipPolygon illegal access vector.at(-1)


mimott:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Hi, plotting into the off leads to an exception. Running the following code with 6.28/00 works like expected. 

But using 6.32.06 the run stops with the following exception:

`/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = double; _Alloc = std::allocator<double>; std::vector<_Tp, _Alloc>::reference = double&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.`

### Reproducer

```
#include "TCanvas.h"
#include "TPolyLine.h"

int main(int argc, char **argv)
{
   TCanvas* c = new TCanvas("c", "Something", 0, 0, 800, 600);

   double x[5], y[5];

   x[0] = x[3] = x[4] = -0.5;
   x[1] = x[2] = -0.1;
   y[0] = y[1] = y[4] = 0.2;
   y[2] = y[3] = 0.3;

   auto* line = new TPolyLine(5, x, y);
   line->Draw("f");

   c->Print("demo.png");
   return 0;
}
```

### ROOT version

ROOT 6.32.06

### Installation method

yum

### Operating system

Rocky Linux 9.4

mimott:
Hi,
thank you for your answer.

With "plotting into off" I mean, that the line is not drawn to the canvas since some or all of its coordinates are outside the domain. In my example the x coordinate starts at -0.5 and ends at -0.1. If the line is patially drawn to the canvas (ends at 0.25) the exception does not happen.

I compiled the code with gcc 11.4.1 in Release (-O3 -DNDEBUG) same for clang 17.0.6

If I load the sample in ROOT interpreter mode I get the same assertion.

Now I downloaded 6.32.12 and the problem disappears. The same if I use a 6.32.06 downloaded by myselfe.

The only difference is that both downloaded packages are in my user context while the one with the problem is installed into the system. Is it needed to have some variables to be set (e.g. ROOTSYS, LD_LIBRARY_PATH) if ROOT is installed to the system? I thought that wouldn't be needed.

ferdymercury:
[...]

PR #3955, opened 2019-06-12
Add protection in ClipPolygon

couet:
In some cases when the polygons are just a few points some of the internal arrays indices got negative values. For instance this was the case with the example $ROOTSYS/test/delaunayTriangulation.cxx

PR #18488, opened 2025-04-24
[gpad] Fix illegal access to vector.at(-1)

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/18477

The reproducer is:

```
#include "TCanvas.h"

void poly()
{
   TCanvas* c = new TCanvas("c", "Something", 0, 0, 800, 600);

   double x[5], y[5];

   x[0] = x[3] = x[4] = -0.5;
   x[1] = x[2] = -0.1;
   y[0] = y[1] = y[4] = 0.2;
   y[2] = y[3] = 0.3;

   Int_t n = 5;
   Int_t nc = 2*n+1;
   std::vector<Double_t> vx(nc, 0.);
   std::vector<Double_t> vy(nc, 0.);

   Int_t res = c->ClipPolygon(n, x, y, nc, &vx.front(), &vy.front(),0,1,0,1);
}
```

but it only crashes in "yum" version of RockyLinux, probably because it's undefined behavior.

I checked though with the debugger that there was this line 984 in TPad.cxx:

x1 = xc2[nc2-1]; y1 = yc2[nc2-1];

that was being called with `nc2 = 0`.

This is an analogous fix to the one done by @couet in https://github.com/root-project/root/pull/3955, but for the other 'axes'. 

## Checklist:

- [x] tested changes locally

### DIFF
--- a/graf2d/gpad/src/TPad.cxx
+++ b/graf2d/gpad/src/TPad.cxx
@@ -955,10 +955,13 @@ Int_t TPad::ClipPolygon(Int_t n, Double_t *x, Double_t *y, Int_t nn, Double_t *x
    std::vector<Double_t> xc2(nn), yc2(nn);
 
    // Clip against the left boundary
-   x1 = x[n-1]; y1 = y[n-1];
+   if (n > 0) {
+      x1 = x[n - 1];
+      y1 = y[n - 1];
+   }
    nc2 = 0;
    Int_t i;
-   for (i=0; i<n; i++) {
+   for (i = 0; i < n; i++) {
       x2 = x[i]; y2 = y[i];
       if (x1 == x2) {
          slope = 0;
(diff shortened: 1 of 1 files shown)


graf2d/gpad/src/TPad.cxx
```diff
@@ -1005,60 +1011,64 @@ Int_t TPad::ClipPolygon(Int_t n, Double_t *x, Double_t *y, Int_t nn, Double_t *x
...
-      nc2 = 0;
-      for (i=0; i<nc; i++) {
-         x2 = xc[i]; y2 = yc[i];
-         if (x1 == x2) {
-            slope = 0;
+   // Clip against the right boundary
+   if (nc > 0) {
+      x1 = xc[nc - 1];
+      y1 = yc[nc - 1];
+   }
```
linev, line 1018: You let execute now complete branch which was before checked with `if (nc > 0)`.
Is it really necessary?

ferdymercury: It's not executed since `for ( ; i < 0 ; )` never enters the loop. This way, we set nc2 = 0 before going to the next dimension.

linev: But it completely useless. 
If `nc == 0` there is nothing to do, why you want to check this several more times?

ferdymercury: Fair enough.

linev (approved):
Thanks!

PR #18491, opened 2025-04-24
[cmake] do not check none-existing variables

linev:
Then checking ROOTSYS, do not try to modify
none-existing variables. 
Like LD_LIBRARY_PATH on windows.
cmake on Windows simply fails when ROOTSYS set.

Also no need to change ROOTSYS if it set correctly


pcanal (approved):
Thanks.

Issue #17841, opened 2025-02-27
PyROOT multidimenstional int16_t arrays stopped working in 6.34


cozzyd:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

We serialize some objects containing 2D int16_t arrays (these are waveforms). For our PyROOT interface, we convert them to numpy arrays using numpy.frombuffer.

Starting in ROOT 6.34, at least as distributed by Fedora and by EL9 EPEL, this no longer works.  One-dimensional arrays seem fine.

### Reproducer

Here is a standalone example exhibiting the problem 

```
import cppyy
import cppyy.ll
import numpy

cppyy.cppdef("""
 struct Foo {int16_t bar[24][2] = {};};
 Foo foo;
 foo.bar[0][0]=1;
 foo.bar[10][1]=1;
 for(int i = 0; i < 24; i++) { std::cout << foo.bar[i][0] << " " << foo.bar[i][1] << std::endl; }
 """)

arr = cppyy.gbl.foo.bar
print(arr.shape)
print(arr.typecode)
print(arr.format)
# this cast was necessary for some reason in older ROOT versions, though it doesn't change anything here
a = numpy.frombuffer(cppyy.ll.cast['int16_t*'](arr), dtype='int16', count=24*2).reshape((24,2))
print(a)
```

This gives something like
```
1 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 1
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
0 0
(268435455, -1)
L
L
[[  7424  -5267]
 [-14155 -14397]
 [-11728 -28243]
 [ 32764      0]
 [  2288 -32302]
 [ 32521      0]
 [ 25856 -32290]
 [ 32521      0]
 [  1856 -32275]
 [ 32521      0]
 [     0      0]
 [     0      0]
 [  7424  -5267]
 [-14155 -14397]
 [-11712 -28243]
 [ 32764      0]
 [-29904  25230]
 [ 32521      0]
 [     1      0]
 [     0      0]
 [-13552  -8813]
 [ 32520      0]
 [     0      0]
 [     0      0]]

```

Obviously the shape and typecode are wrong, and the array is garbage, when this used to work fine. 

### ROOT version

ROOT 6.34.04 

### Installation method

dnf

### Operating system

Fedora and also EPEL9

### Additional context

[...]

Issue #9809, opened 2022-02-04
[prompt, win32] Read-after-write of variable of type `double` does not yield the correct result


linev:
Typing in ROOT prompt one gets:
```
root [1] double z = 1234;
root [2] z = 1
(double) 1.00000
root [3] z
(double) 1234.0000
```
Last line is wrong - expected of course `1.0000`. 
Works ok on Linux and on Win64 (according to @bellenot)

bellenot:
Windows 32 bit

bellenot:
It's on my local computer, but I can give you access to a VM (tomorrow)
```
C:\Users\bellenot>root -l
root [0] gClingOpts->AllowRedefinition = 0;
root [1] double z = 1234;
root [2] z = 1
(double) 1.0000000
root [3] z
(double) 1234.0000
root [4]
```

jalopezg-r00t:
> It's on my local computer, but I can give you access to a VM (tomorrow)

Yes, let's do that tomorrow!  :+1:

jalopezg-r00t:
The issue only occurs in 32-bit Windows.  As discussed, it is related to an optimization pass (`-O1` is enabled for ROOT, but not for standalone cling).  Both workarounds below yield the correct result; the exact pass triggering this is to be determined.

```c++
root [0] .O 0
root [1] double z = 1234;
root [2] z = 1
(double) 1.0000000
root [3] z
(double) 1.0000000
```
or
```c++
root [0] volatile double z = 1234;
// ...
```
As discussed with Axel, we'll probably defer this after the llvm upgrade.

bellenot:
Thanks @jalopezg-r00t for the investigation and for finding this!

bellenot:
And interestingly enough, if I comment out the `clingArgsStorage.push_back("-O1");` in `TCling.cxx`, the test failures on Windows `x64` decrease from about 30 down to 12

jalopezg-r00t:
FYI (to be seen after the llvm upgrade), via `-opt-bisect-limit=` I have tracked this down to an issue in the optimization pass "X86 DAG->DAG Instruction Selection" during codegen for the assignment statement:
```c++
root [1] z = 1
```

Specifically, the instructions emitted for the `-O0` version that works are
```
  renamable $xmm0 = MOVSDrm_alt $noreg, 1, $noreg, %const.0, $noreg
[...]

PR #18492, opened 2025-04-24
[cppyy] Add converters and low-level views for fixed width integers

aaronj0:
fixes: https://github.com/root-project/root/issues/17841

The support for the fixed-width types `int16_t` and `int32_t` indeed diverges from upstream cppyy and uses `unsigned long` which triggers the incorrect numpy view. Despite not having converters for the same, upstream relies on patches (or older behaviour) along the `Cppyy::ResolveName -> TClassEdit::ResolveTypedef` code path (see [ResolveTypeDefImpl](https://github.com/root-project/root/blob/976bd38b3d229a7006dfad9501ba5ac5ce2df3c2/core/foundation/src/TClassEdit.cxx#L1528)) that resolved fixed width integer types to their standard type counterparts (`int16_t[][]` to `short[][]` and `int32_t[][]` to `int[][]`)  when creating converters. Trying to follow the same code branch in ROOT can have a large blast radius in the type system. Two non-intrusive ways to fix this issue are either aliasing to existing short and int converters, or adding the converters and low-level views for both types. This PR performs the latter


guitargeek (approved):
Thank you very much for implementing better support for fixed-width integers! Looks good to me, but please also add a test :slightly_smiling_face:

guitargeek:
Oh and don't forget to fix also the compiler warnings, otherwise the `debian125` build that treats warnings as errors because of `dev=ON` will fail

aaronj0:
The following issue: https://github.com/root-project/root/issues/9809 causes failures on windows x86, where the arrays assigned on the C++ side itself are indexed/read wrongly:

```cpp
root [0] int test[]={2,4,6,8,10};
root [1] test[4]
(int) 2
root [2] test[4]=20;
root [3] test[4]
(int) 20
root [4] test[0]
(int) 20
root [5] test
(int [5]) { 20, 4, 6, 8, 10 }
root [6]
```

Hence validating those buffers fail and the test will have to be disabled on win86.

aaronj0:
Looks like the failures on win 64 are unrelated: `roottest-root-io-stdpair`

PR #18494, opened 2025-04-24
[gpad] TPad fix clang-tidy warnings

ferdymercury:
# This Pull request:

## Changes or fixes:

applies fixes to clang-tidy diagnostics in class TPad

### DIFF
--- a/graf2d/gpad/src/TPad.cxx
+++ b/graf2d/gpad/src/TPad.cxx
@@ -7025,7 +7025,7 @@ TObject *TPad::WaitPrimitive(const char *pname, const char *emode)
    TObject *obj = nullptr;
    Bool_t testlast = kFALSE;
    Bool_t hasname = pname && (strlen(pname) > 0);
-   if (!pname[0] && !emode[0]) testlast = kTRUE;
+   if ((!pname || !pname[0]) && (!emode || !emode[0])) testlast = kTRUE;
    if (testlast) gROOT->SetEditorMode();
    while (!gSystem->ProcessEvents() && gROOT->GetSelectedPad() && gPad) {
       if (gROOT->GetEditorMode() == 0) {


graf2d/gpad/src/TPad.cxx
```diff
@@ -4220,8 +4220,10 @@ void TPad::PaintHatches(Double_t dy, Double_t angle,
       if (yrot < ymin) ymin = yrot;
    }
    ymax = (Double_t)((Int_t)(ymax/dy))*dy;
-
-   for (ycur=ymax; ycur>=ymin; ycur=ycur-dy) {
+   
+   const Int_t nYsteps = static_cast<Int_t>((ymax - ymin)/dy) + 1;
```
linev, line 4224: Not sure if it does exactly the same as before.
I do not see why last value of `ycur` must be exactly the `ymin`.

linev (dismissed):
I prefer not to change code of `TPad::PaintHatches`

graf2d/gpad/src/TPad.cxx
ferdymercury: Last value is not exactly ymin because of the static_cast<Int_t>.

`lastYcur = ymax - dy * static_cast<Int_t>((ymax - ymin)/dy)`

which is equivalent to `ymax - dy -dy -dy -dy ...`

linev: But why it should be better than before?

linev: I do not see such usecase. 
`dy` always reasonable value compared with `y`.
Otherwise `y/dy` also will be abnormal.
Just see how `PaintHatches` method called from `TPad::PaintFillAreaHatches`.
I would not touch this code.

ferdymercury: I am sorry to insist: did you check e.g. this:
https://pvs-studio.com/en/docs/warnings/v3521/

For example, when compiled with GCC or Microsoft Visual Studio 2013 and executed on an x86 processor, the loop is evaluated only nine times; whereas other compilers may execute it ten times. static-cast to Int prevents that undeterministic behavior.

https://wiki.sei.cmu.edu/confluence/display/c/FLP30-C.+Do+not+use+floating-point+variables+as+loop+counters

Also, even if in this case it's less of a concern since doubles and not floats are being used, if we have many 'warnings' left when running code-coverage tools or clang-tidy, it's hard to find the real warnings as we are swamped with many others.

I think my proposal goes a bit in the direction that sprintf *coverity* warnings that e.g. @silverweed has been fixing in his recent PR https://github.com/root-project/root/pull/18435, even if for that use case the sprintf things worked just fine, it's good to solve them so to concentrate on real true positives when reading coverity or clang-tidy reports.

linev: Again - my doubt that you change calculation of steps and rounding.
Like you wrote `const Int_t nYsteps = static_cast<Int_t>((ymax - ymin)/dy) + 1;`. 
But in original code steps has nothing to do with `ymin` value at all. 

We do not have reliable test to check that after such changes all fill styles will work as before.
And having idea how such styles working and why they have such patterns - small mistake here can have big influence on the result.

In JSROOT I re-wrote this code from scratch and have very precise test - there I checking position of every line segments before and after each change in the code. But in JSROOT I use SVG to let fill with pattern arbitrary area - hoping SVG engine of browser does it correct. But in ROOT this logic combined together - filling by hatches and handling arbitrary fill area - making all story to test such changes.

You can introduce integer for steps and calculation of `ycur` based on such steps. 
But I really want to see final condition `ycur>=ymin` in the code.

ferdymercury: > But in original code steps has nothing to do with `ymin` value at all.

It does, since the stopping condition in the original code is `for( ; ycur>=ymin; )`
so it stops when `ycur < ymin`. 
And at a generic iteration we have `ycur = ymax - dy -dy -dy -dy - ....`

So in the original code, the stopping condition is 
` ymax - dy -dy -dy -dy - ... < ymin`
 
 which equals
 
` ymax - nIteratedSteps *dy < ymin`

so

nIteratedSteps > (ymax - ymin)/dy

so by doing the static_cast() + 1, we are fulfilling the exact same stop condition.

Do you want me to create a self-contained reproducer comparing both code implementations for a variety of cases checking that the stopping condition is equivalent?

linev: Here my alternative https://github.com/root-project/root/pull/18510

Keep integer conversion - but then stay with such integer.
Does not change logic.

guitargeek (approved):
Thank you!

PR #18495, opened 2025-04-24
Avoid sprintf deprecation warning in test

vepadulano:

PR #18502, opened 2025-04-25
[Windows] Fix warning C4477 + code formatting

bellenot:
Fix the following compilation warning on Windows:
```
IoBigEventGeneration.cxx(355): warning C4477: 'printf' : format string '%ld' requires an argument of type 'long', but variadic argument 1 has type 'unsigned __int64'
IoBigEventGeneration.cxx(355): note: consider using '%zd' in the format string
```

### DIFF
--- a/roottest/root/io/bigevent/IoBigEventGeneration.cxx
+++ b/roottest/root/io/bigevent/IoBigEventGeneration.cxx
@@ -141,16 +141,18 @@ Event::Event()
    fDequePair.clear();
    
    fVectorshort.clear();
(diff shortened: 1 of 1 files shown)


vepadulano:
I merged this related PR https://github.com/root-project/root/pull/18495 so that we don't use the deprecated sprintf function anymore. I guess the error about the `'%ld'` format string will still be there, let's see after a rebase 👍 .

bellenot:
@vepadulano thanks! The PR has been rebased

pcanal (approved):

PR #18510, opened 2025-04-25
[gpad] avoid potential float precision lost in PaintHatches

linev:
Use integer conversion once and then just check boundary with ymin

Alternative to #18494


ferdymercury:
Side note: an integer for loop is faster than doing a double comparison on every loop (while) step for stopping.

couet (approved):

PR #18512, opened 2025-04-25
[RF] allow obs to be specified in extended binding

will-cern:
I noticed that we cannot set which obs are used in the RooExtendedBinding. This should be fixed.


roofit/roofitcore/inc/RooExtendedBinding.h
```diff
@@ -14,19 +14,22 @@
...
   RooExtendedBinding() {} ;
   RooExtendedBinding(const char *name, const char *title, RooAbsPdf& _pdf);
+  RooExtendedBinding(const char *name, const char *title, RooAbsPdf& _pdf, const RooArgSet& _obs);
   RooExtendedBinding(const RooExtendedBinding& other, const char* name=nullptr) ;
   TObject* clone(const char* newname=nullptr) const override { return new RooExtendedBinding(*this,newname); }
 
 protected:
 
   RooRealProxy pdf ;
+  RooSetProxy _obsList;
```
ferdymercury, line 32: classdef version should be increased? or this member marked as transient ?

guitargeek (approved):
Thank you! I also added a test.

PR #18516, opened 2025-04-25
[json] modify json roottest

linev:
Alternative for https://github.com/root-project/root/pull/18483

Use extra LIBS argument to `ROOTTEST_GENERATE_DICTIONARY`

### DIFF
--- a/roottest/cmake/modules/RoottestMacros.cmake
+++ b/roottest/cmake/modules/RoottestMacros.cmake
@@ -307,7 +307,7 @@ endmacro(ROOTTEST_COMPILE_MACRO)
 #
 #-------------------------------------------------------------------------------
 macro(ROOTTEST_GENERATE_DICTIONARY dictname)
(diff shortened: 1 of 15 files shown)


linev:
> Do we understand 'why' the extra library is needed 

No. I hope Bertrand @bellenot  can help.
Problem that I cannot reproduce wired behavior when building of dictionary failed, but than reported as ok.
Therefore I just trigger CI jobs with different approach

linev:
Problem will be fixed, no need for extra libs

pcanal:
> Problem will be fixed, no need for extra libs

For the record how/where will the problem be fixed?  Thanks.

linev:
Bertrand found mismatch how ROOT libraries configured for roottest.
See: https://github.com/root-project/root/blob/master/roottest/CMakeLists.txt#L70-L74
`Graf` is not included there. 
If I understand correctly - Bertrnad want to sync list here with default list

PR #18517, opened 2025-04-25
Use jsroot 7.9.0 in jupyter and http server docu [6.36]

linev:
JSROOT 7.9.0 is base version for ROOT 6.36 and has to be used as fallback for jupyter


dpiparo:
Can we have another PR updating the Release Notes too?

Issue #18519, opened 2025-04-25
Concurrency issue with TClassEdit::ResolveTypedef and TClass::GetListOfMethods


Dr15Jones:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

While testing RNTuple code with CMS release CMSSW_15_1_RNTUPLE_X_2025-04-24-2300 we are seeing lots of failures when running the system with multiple threads. The tracebacks all involve one thread calling RNTuple routines which include `TClassEdit::ResolveTypedef ` in the traceback and the others involve code that is asking TClass about member functions of classes.

### Reproducer

setup work area using CMSSW_15_1_RNTUPLE_X_2025-04-24-2300.

In the `/src/` directory do
```
mkdir Test/Test/bin
```

In that directory add the file main.cc containing
```C++
#include "TClassEdit.h"

#include "FWCore/Reflection/interface/TypeWithDict.h"

#include "DataFormats/Candidate/interface/LeafCandidate.h"

#include <thread>
#include <atomic>


int main() {

  std::atomic<int> countDown{2};

  std::thread t1([](std::atomic<int>& count) {
                   --count;
                   while(count != 0);
                   TClassEdit::ResolveTypedef("std::vector<unsigned int>", false);
                 }, std::ref(countDown));
  std::thread t2([](std::atomic<int>& count) {
                   edm::TypeWithDict t(typeid(reco::LeafCandidate));
                   --count;
                   while(count!=0);
                   t.functionMemberByName("eta");
                 }, std::ref(countDown));

  t1.join();
  t2.join();
  return 0;
}
```
also create the file BuildFile.xml containing
```xml
<bin name="testRoot" file="test.cc">
  <use name="root"/>
  <use name="DataFormats/Candidate"/>
  <use name="FWCore/Reflection"/>
</bin>
```

build using `scram b` then from the src directory run the program `../bin/el8_amd64_gcc12/testRoot`

Several tries are normally necessary before a failure is seen. One such failure while running in the debugger gave an assertion failure with the message

```
testRoot: /data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/BUILD/el8_amd64_gcc12/lcg/root/6.35.1-6c0d80f51d83322604af89ae12687a2c/root-6.35.1/interpreter/llvm-project/clang/include/clang/Lex/Preprocessor.h:1877: const clang::Token& 
[...]

PR #18522, opened 2025-04-25
[cling] The LookupHelper routines need the ROOT lock.

### LINKED ISSUE #18520: Missing lock deep inside TClassEdit::ResolveTypedef
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

This is related but different from https://github.com/root-project/root/issues/18442 and https://github.com/root-project/root/issues/18236.

At https://github.com/cms-sw/cmssw/issues/47953#issuecomment-2830633844, noticed that in the stack:
```
#19 0x000014f132954a8d in cling::LookupHelper::findScope(llvm::StringRef, cling::LookupHelper::DiagSetting, clang::Type const**, bool) const () from /cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02886/el8_amd64_gcc12/cms/cmssw/CMSSW_15_1_RNTUPLE_X_2025-04-24-2300/external/el8_amd64_gcc12/lib/libCling.so
#20 0x000014f13250a5e5 in ROOT::TMetaUtils::TClingLookupHelper::IsDeclaredScope (this=0x14f144cdba80, base=..., isInlined=@0x7ffe8500027f: false) at /data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/BUILD/el8_amd64_gcc12/lcg/root/6.35.1-6c0d80f51d83322604af89ae12687a2c/root-6.35.1/core/clingutils/src/TClingUtils.cxx:588
#21 0x000014f14e89172e in ResolveTypedefImpl (tname=0x14f0f9ce38c0 "std::pair<edm::Hash<1>,edm::ParameterSetBlob>", len=45, cursor=@0x7ffe850005fc: 3, modified=@0x7ffe850005fb: false, result=...) at /data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/BUILD/el8_amd64_gcc12/lcg/root/6.35.1-6c0d80f51d83322604af89ae12687a2c/root-6.35.1/core/foundation/src/TClassEdit.cxx:1605
#22 0x000014f14e892b63 in TClassEdit::ResolveTypedef[abi:cxx11](char const*, bool) (tname=0x14f0f9ce38c0 "std::pair<edm::Hash<1>,edm::ParameterSetBlob>") at /data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/BUILD/el8_amd64_gcc12/lcg/root/6.35.1-6c0d80f51d83322604af89ae12687a2c/root-6.35.1/core/foundation/src/TClassEdit.cxx:1788
#23 0x000014f0f275a06b in ROOT::RFieldBase::Create (fieldName=..., typeName=..., options=..., desc=0x14f0ee1afc90, fieldId=0) at /data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/BUILD/el8_amd64_gcc12/lcg/root/6.35.1-6c0d80f51d83322604af89ae12687a2c/root-6.35.1/tree/ntuple/src/RFieldBase.cxx:299
```

The ROOT global is not taken (until it is too late) leading to a crash.
Questions to investigate:
* Was/is `TClassEdit::ResolveTypedef` meant to use with the global lock or is the user meant to take the lock
* If it is meant to be thread safe, should it take the lock (this is technical issue as `core/foundation` does not have access to the lock yet)
* If it is meant to be thread safe but not take a lock where should be taken?
* * `findScope` is the obvious place from the stack trace
* * but `ResolveTypedefImpl` is also calling `GetPartiallyDesugaredNameWithScopeHandling` which need to protected.

### Reproducer

.

### ROOT version

.

### Installation method

.

### Operating system

.

### Additional context

.

pcanal:
Those routines are at the very least looking up information into Clang and thus need to prevent concurrent updates.  Some of the routines can also sometimes induces changes in Clang (eg. template instantiation).

This fixes #18520 and #18519

### DIFF
--- a/core/clingutils/src/TClingUtils.cxx
+++ b/core/clingutils/src/TClingUtils.cxx
@@ -54,7 +54,7 @@
 #include "cling/Interpreter/Transaction.h"
 #include "cling/Interpreter/Interpreter.h"
 #include "cling/Utils/AST.h"
-#include "cling/Interpreter/EnterUserCodeRAII.h"
+#include "cling/Interpreter/InterpreterAccessRAII.h"
 
 #include "llvm/Support/Path.h"
 #include "llvm/Support/FileSystem.h"
@@ -571,7 +571,7 @@ bool TClingLookupHelper::IsAlreadyPartiallyDesugaredName(const std::string &nond
                                                          const std::string &nameLong)
 {
    // We are going to use and possibly update the interpreter information.
-   LockCompilationDuringUserCodeExecutionRAII LCDUCER(*interp);
+   cling::InterpreterAccessRAII LockAccess(*fInterpreter);
 
    const cling::LookupHelper& lh = fInterpreter->getLookupHelper();
    clang::QualType t = lh.findType(nondef.c_str(), ToLHDS(WantDiags()));
@@ -589,7 +589,7 @@ bool TClingLookupHelper::IsAlreadyPartiallyDesugaredName(const std::string &nond
 bool TClingLookupHelper::IsDeclaredScope(const std::string &base, bool &isInlined)
 {
    // We are going to use and possibly update the interpreter information.
-   LockCompilationDuringUserCodeExecutionRAII LCDUCER(*interp);
+   cling::InterpreterAccessRAII LockAccess(*fInterpreter);
 
    const cling::LookupHelper& lh = fInterpreter->getLookupHelper();
    const clang::Decl *scope = lh.findScope(base.c_str(), ToLHDS(WantDiags()), nullptr);
@@ -626,7 +626,7 @@ bool TClingLookupHelper::GetPartiallyDesugaredNameWithScopeHandling(const std::s
    if (fAutoParse) fAutoParse(tname.c_str());
 
    // We are going to use and possibly update the interpreter information.
-   LockCompilationDuringUserCodeExecutionRAII LCDUCER(*interp);
+   cling::InterpreterAccessRAII LockAccess(*fInterpreter);
 
    // Since we already check via other means (TClassTable which is populated by
    // the dictonary loading, and the gROOT list of classes and enums, which are
--- a/core/clingutils/src/TClingUtils.cxx
+++ b/core/clingutils/src/TClingUtils.cxx
@@ -54,6 +54,7 @@
 #include "cling/Interpreter/Transaction.h"
 #include "cling/Interpreter/Interpreter.h"
 #include "cling/Utils/AST.h"
+#include "cling/Interpreter/EnterUserCodeRAII.h"
 
 #include "llvm/Support/Path.h"
 #include "llvm/Support/FileSystem.h"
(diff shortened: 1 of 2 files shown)


dpiparo (approved):
The change looks fine to me, thanks, provided that I understood correctly what was the role of the RAII (the header cannot be found)

vgvassilev:
Why not locking the underlying interface in the LookupHelper?

pcanal:
> Why not locking the underlying interface in the LookupHelper?

Because one need to lock both the (potential) write access and the read accesses.

For example in `GetPartiallyDesugaredNameWithScopeHandling` we have:
```
   clang::QualType t = lh.findType(tname.c_str(), ToLHDS(WantDiags()));
   // Technically we ought to try:
   //   if (t.isNull()) t =  lh.findType(TClassEdit::InsertStd(tname), ToLHDS(WantDiags()));
   // at least until the 'normalized name' contains the std:: prefix.
   if (!t.isNull()) {
```
The lock needs to be held during both the 1st (because of potential modification by this thread) and the last line (because of potential modification by another thread)

We could improve the scaling further by leveraging the fact that the ROOT lock is a read-write lock but that level of granularity is not available from Cling yet.

pcanal:
>  the role of the RAII (the header cannot be found)

The header is `interpreter/cling/lib/Interpreter/EnterUserCodeRAII.h` and the `RAII` does the lock taking and releasing as needed.  The name of the `RAII` is nowadays a misnomer as it is now used past its original usage intent.

vgvassilev:
> > Why not locking the underlying interface in the LookupHelper?
> 
> Because one need to lock both the (potential) write access and the read accesses.

Why the read access? Do you mean because read access sometimes mutates the AST?

pcanal:
> Why the read access? Do you mean because read access sometimes mutates the AST?

No.  Because another thread might mutate the area you are reading for example:
```
thread 1 time 1: Take lock
thread 1 time 2:   Load header file with class A
thread 1 time 3:   Find class A
thread 1 time 4: Release lock
thread 1 time 7: Iterate through class A's content
```
and
```
thread 2 time 4: Take lock
thread 2 time 5:    Unload header file class A (revert Transaction)
thread 2 time 6: Release lock
```
Then during ` time 7`, `thread 1` is using data/memory that is either deleted or least no longer what it was meant to be.

As long as there is 'destructive' or `changing` operation left in Cling (eg. unloading transaction) then we do need to have at least a Read Lock when reading data.

The general rule we have been trying (and only incompletely succeeding) to always hold the ROOT lock when accessing Clang/Cling information.

dpiparo:
I think this can be merged, I do not believe fedora42 error is real.

hahnjo:
Hm, so the first commit of this PR will not compile on its own? That's not great :disappointed:

pcanal:
> Hm, so the first commit of this PR will not compile on its own? That's not great

Nope :( I remember right after click Merge than I was meant to click 'Squash' :(  Sorry.

hahnjo:
We must definitely get rid of duplicated code because it will be a nightmare to figure this out in some years from now, after the two copies diverged...

pcanal:
See https://github.com/root-project/root/pull/18551

PR #18532, opened 2025-04-28
[ci] Enable `unuran` on Fedora 42

guitargeek:
The builtin unuran version doesn't compile with gcc 15, so let's try using the system version.


pcanal:
> The builtin unuran version doesn't compile with gcc 15

If this is the case, shouldn't we consider upgrading the version of the builtin unuran or at the very least at an explicit configuration failure (i.e. something like `if (gcc15 and builtin_unuran) message(FATAL_ERROR "Unsupported combination ... ") endif() ` )?

guitargeek:
Sure, you can do that too if you want! I think it's a good idea with the explicit error. Like this, we're encouraging users to go into the direction of using fewer builtins, and upgrading/fixing the builtin `unuran` would not go towards that.

> shouldn't we consider upgrading the version of the builtin unuran
I don't know if the latest version (from 2023) works with g++15 :slightly_smiling_face: to be checked.

I'll take care of that, but IMO this is independent of this PR, which is about about updating the CI config to use less builtins and enable `unuran=ON` on Fedora 42.

pcanal (approved):

PR #18538, opened 2025-04-28
[roottest] Delete outdated copy of unuran tests from roottest directory

guitargeek:
The up-to-date version of the tests is contained in `math/unuran/test`. The outdated copy in the `roottest` directory was actually completely inactive, as the declaration of the tests was "commented out" in the CMakeLists.txt with an `if(OFF)`.


pcanal (approved):
Thanks

PR #18541, opened 2025-04-29
[HF] Prevent a crash when an empty measurement is created.

hageboeck:
The following code provided by Tomas Dado lead to a crash:
```
  RooStats::HistFactory::Measurement meas("meas", "meas");
  meas.SetOutputFilePrefix("test");
  meas.SetPOI("pt0_m0_y0_XSec");
  meas.CollectHistograms();

  std::unique_ptr<RooWorkspace> work(RooStats::HistFactory::MakeModelAndMeasurementFast(meas));
```

This is caused by a nullptr being passed to a HistFactory function. By returning early, the crash is avoided, and the following error message is produced:
`Error in <RooStats::HistFactory::HistoToWorkspaceFactoryFast::MakeCombinedModel>: Input vectors are empty - return a nullptr`


guitargeek (approved):
Nice! Thanks a lot

PR #18543, opened 2025-04-29
[RF] Use the actual proxy list as input for RooFormula::doEval()

guitargeek:
This fixes problems when the variable ist in the RooFormula gets out of
sync with the proxies in the RooFormulaVar or the RooGenericPdf, which
are using the RooFormula under the hood. This can happen when the
variables are renamed using `SetName()`, for example in the NLL creation
for simultaneous pdfs with the new RooFit CPU backend.

This fixes a bug reported by ATLAS.

Also some general `RooFormula` improvements in separate PRs.


lmoneta (approved):
LGTM!
Thank you Jonas for the fix!

Issue #18542, opened 2025-04-29
TMinuit warning about array out of bounds access


pcanal:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

```
/uscms/home/pcanal/root_working/code/quick-devel/math/minuit/src/TMinuit.cxx: In member function 'virtual void TMinuit::mnline(Double_t*, Double_t, Double_t*, Double_t, Double_t)':
/uscms/home/pcanal/root_working/code/quick-devel/math/minuit/src/TMinuit.cxx:5966:4: warning: array subscript -1 is outside array bounds of 'Double_t [3]' {aka 'double [3]'} [-Warray-bounds]
 5966 |    --parx2p;
      |    ^~~~~~~~
/uscms/home/pcanal/root_working/code/quick-devel/math/minuit/src/TMinuit.cxx:4740:23: note: while referencing 'xvals'
 4740 |    Double_t fvals[3], xvals[3], f1, fvmin, xvmin, ratio, f2, f3 = 0., fvmax;
      |                       ^~~~~
/uscms/home/pcanal/root_working/code/quick-devel/math/minuit/src/TMinuit.cxx:5966:4: warning: array subscript -1 is outside array bounds of 'Double_t [3]' {aka 'double [3]'} [-Warray-bounds]
 5966 |    --parx2p;
      |    ^~~~~~~~
/uscms/home/pcanal/root_working/code/quick-devel/math/minuit/src/TMinuit.cxx:4740:23: note: while referencing 'xvals'
 4740 |    Double_t fvals[3], xvals[3], f1, fvmin, xvmin, ratio, f2, f3 = 0., fvmax;
      |                       ^~~~~
/uscms/home/pcanal/root_working/code/quick-devel/math/minuit/src/TMinuit.cxx: In member function 'virtual void TMinuit::mncros(Double_t&, Int_t&)':
/uscms/home/pcanal/root_working/code/quick-devel/math/minuit/src/TMinuit.cxx:5966:4: warning: array subscript -1 is outside array bounds of 'Double_t [3]' {aka 'double [3]'} [-Warray-bounds]
 5966 |    --parx2p;
      |    ^~~~~~~~
/uscms/home/pcanal/root_working/code/quick-devel/math/minuit/src/TMinuit.cxx:1801:13: note: while referencing 'alsb'
 1801 |    Double_t alsb[3], flsb[3], bmin, bmax, zmid, sdev, zdir, zlim;
      |             ^~~~
[...]

PR #18549, opened 2025-04-29
[math] fix out of bounds array

ferdymercury:
Fixes https://github.com/root-project/root/issues/18542


guitargeek (approved):
Ah yes that's how it is when automatically translating Fortran code. Thanks!

PR #18550, opened 2025-04-29
[v636][cling] The LookupHelper routines need the ROOT lock

pcanal:
Back-port of https://github.com/root-project/root/pull/18522

Those routines are at the very least looking up information into Clang and thus need to prevent concurrent updates. Some of the routines can also sometimes induces changes in Clang (eg. template instantiation).

This fixes https://github.com/root-project/root/issues/18520 and https://github.com/root-project/root/issues/18519


pcanal:
The Fedora failures are unrelated.

PR #18551, opened 2025-04-29
cling: Deduplicate lock code. Keep semantic distinction.

pcanal:
LockCompilationDuringUserCodeExecutionRAII: to be used right before compiling/executing user code
InterpreterAccessRAII: to be used during loookups


hahnjo:
I don't really understand the semantic distinction between the two types: both access the interpreter and need to lock during compilation.

pcanal:
The 'difference' is that the user-code related one is intended to be later temporarily suspended (via `EnterUserCodeRAII`) during the actual user code execution while the new usages are meant to be use in places where there is no user code being called and thus the lock will actually be held continuously by the thread until it is released by the RAII.

hahnjo (approved):
ok

dpiparo (approved):

PR #18560, opened 2025-04-30
[cmake] collect all errors when fail-on-missing rather than stopping at first one

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes second-to-last item of https://its.cern.ch/jira/browse/ROOT-9385


hageboeck (approved):
Thanks, that's a nice improvement!

ferdymercury:
Thanks Stefan! I would suggest closing the associated JIRA issue since the missing fifth item is probably quite nightmarish to maintain.

PR #18562, opened 2025-04-30
[cmake][roottest] Add libGraf in ROOT_LIBRARIES

bellenot:


linev (approved):
Should fix problem in https://github.com/root-project/root/pull/18483

PR #18504, opened 2025-04-25
[DF][Win] Fix compiler error C3493

bellenot:
Fix the following compiler error on Windows:
```
dataframe_nodes.cxx(86,40): error C3493: 'N' cannot be implicitly captured because no default capture mode has been specified
```

PR #18568, opened 2025-04-30
[RDF] Fix a warning in an RSlotStack test.

hageboeck:


guitargeek (commented):
What :laughing: Can't make all compilers happy at the same time it seems!

   * https://github.com/root-project/root/pull/18504

Maybe we just make it not `constexpr unsigned int N = 8`? That probably confuses the compiler. Just `const` is probably fine for a test?

hageboeck:
> What 😆 Can't make all compilers happy at the same time it seems!
> 
> * [[DF][Win] Fix compiler error C3493 #18504](https://github.com/root-project/root/pull/18504)
> 
> Maybe we just make it not `constexpr unsigned int N = 8`? That probably confuses the compiler. Just `const` is probably fine for a test?

Wow, that was why I added that capture ... 🤷‍♂️
OK ... The constexpr is actually better than const, but I will try providing a default capture to make Windows happy ...

PR #18570, opened 2025-04-30
[metacling] copy-paste from RTypes ClassDefBase into interpreted version

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-9415

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/core/metacling/src/TCling.cxx
+++ b/core/metacling/src/TCling.cxx
@@ -304,7 +304,21 @@ private: \
 public: \
    static TClass *Class() { static TClass* sIsA = 0; if (!sIsA) sIsA = TClass::GetClass(#name); return sIsA; } \
    static const char *Class_Name() { return #name; } \
-   virtual_keyword Bool_t CheckTObjectHashConsistency() const overrd { return true; } \
+   virtual_keyword Bool_t CheckTObjectHashConsistency() const overrd {                         \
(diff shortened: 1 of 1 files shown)


core/metacling/src/TCling.cxx
```diff
@@ -304,7 +304,21 @@ private: \
 public: \
    static TClass *Class() { static TClass* sIsA = 0; if (!sIsA) sIsA = TClass::GetClass(#name); return sIsA; } \
    static const char *Class_Name() { return #name; } \
-   virtual_keyword Bool_t CheckTObjectHashConsistency() const overrd { return true; } \
+   virtual_keyword Bool_t CheckTObjectHashConsistency() const overrd {                         \
+      static std::atomic<UChar_t> recurseBlocker(0);                                           \
```
pcanal, line 308: Having a static in an inline function often causes problem on `Windows`  (@bellenot is that still the case).  Albeit this is a virtual inline function since it is virtual and should/would only be seen by the interpreter.

bellenot: It should be fine, the issue with static inline comes when linking and trying to use it from outside. And apparently the builds are green (let me restart the failing one...)

bellenot (approved):

PR #18572, opened 2025-04-30
[v636] Improve CI Performance 

### LINKED ISSUE #18535: Running an invariant reconfigure lead to spurious rebuild
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Just running `cmake .` (in cases where it leads to **no** changes in `CMakeCache.txt`), leads to a lots of spurious recompilation (there should be none).  For example in a minimal build, there 128 targets rebuild (it includes all dictionaries).

This leads to the CI 'wasting' a lot of time doing to spurious rebuilds.

~~The first cause of the issue (there might or might not be other causes) is the unconditional regeneration of `cling-compiledata.h`.~~

### Reproducer

`ninja; cmake .; ninja` should have no target rebuild (except for the harmless `etc/gitinfo.txt`)

### ROOT version

master, v6-36-00-patches (and many older releases

### Installation method

local build

### Operating system

At least Linux and MacOS

### Additional context

_No response_

pcanal:
This sets of tweaks and improve re-running cmake or the buld without any changes leads to no compilation at all.

This will speed up CI operation quite a bit.

The change regarding `rtti` means that after the first build configuration and compilation `rootcling_stage1.cxx.o` was incorrectly build.  Gladfully, this did not seem to have had any practical consequences.

This fixes #18535


hahnjo:
As discussed at multiple other occasions, I don't think we should backport build system changes at this point.

pcanal:
Those changes are limited in scope/reach but will reduce the load on the CI significantly.   In this case I do believe the gain is very much worth the small risk.

hahnjo:
> will reduce the load on the CI significantly

I don't agree: By definition, there are fewer pushes and PRs for `v6-36-00-patches`, the vast majority of CI is spent on `master`.

pcanal:
And on the the 3rd hand 90% of the changes can be considered bug fixes (unintended things recorded, wrong flags used) and the last bit is only used by the CI.

dpiparo:
now that we also released 6.38 perhaps we can close this PR?

pcanal:
> now that we also released 6.38 perhaps we can close this PR?

On the other hand v6.36 is a LTS release ...

PR #18573, opened 2025-04-30
Improve CI Performance

pcanal:
This sets of tweaks and improve re-running cmake or the buld without any changes leads to no compilation at all.

This will speed up CI operation quite a bit.

The change regarding rtti means that after the first build configuration and compilation rootcling_stage1.cxx.o was incorrectly build. Gladfully, this did not seem to have had any practical consequences.

This fixes https://github.com/root-project/root/issues/18535

Forward port of https://github.com/root-project/root/pull/18572


dpiparo (approved):

hahnjo:
Is it possible that one of these changes causes `Core.pcm is out of date` which I feel we see more frequently now?

pcanal:
Per se, yes.  Prior to this `Core.pm` was always regenerated.

PR #18580, opened 2025-05-01
[cling] Add a debug facility to dump the object file and IR from the JIT

vgvassilev:
This would help when debugging ABI mismatches. This patch stores the llvm ir and the object files from the jit in files named cling-module-X.Y.{o,ll} where X is the number of the incremental input compiled and Y is the number of dumped module. X and Y are different because some transactions of Cling do not produce IR.

This functionality can be enabled by:
`EXTRA_CLING_ARGS=" --debug-only=orc " root.exe -l -b -q`

### DIFF
--- a/interpreter/cling/lib/Interpreter/IncrementalJIT.cpp
+++ b/interpreter/cling/lib/Interpreter/IncrementalJIT.cpp
@@ -554,6 +554,40 @@ IncrementalJIT::~IncrementalJIT() {
   }
 }
 
+class DebuggingSimpleCompiler : public llvm::orc::SimpleCompiler {
+public:
+  DebuggingSimpleCompiler(llvm::TargetMachine& TM,
+                          llvm::ObjectCache* ObjCache = nullptr)
+      : llvm::orc::SimpleCompiler(TM, ObjCache) {}
+  llvm::Expected<CompileResult> operator()(llvm::Module& M) override {
+    auto R = llvm::orc::SimpleCompiler::operator()(M);
+
+#ifndef NDEBUG
+#define DEBUG_TYPE "orc"
+    bool Debugging = DebugFlag && isCurrentDebugType(DEBUG_TYPE);
+    if (R && Debugging) {
+      std::string FileStem;
+      {
+        static unsigned counter = 0;
+        llvm::raw_string_ostream FileStemStream(FileStem);
+        FileStemStream << M.getModuleIdentifier() << "." << ++counter;
+      }
+      std::error_code EC;
+      llvm::raw_fd_ostream ModuleStream(FileStem + ".ll", EC);
+      ModuleStream << M;
+      llvm::raw_fd_ostream ObjectStream(FileStem + ".o", EC);
+      ObjectStream << (*R)->getBuffer();
+      LLVM_DEBUG(dbgs() << "Created file '" << FileStem + ".ll"
+                        << "'\n");
+      LLVM_DEBUG(dbgs() << "Created file '" << FileStem + ".o"
+                        << "'\n");
+    }
+#undef DEBUG_TYPE
+#endif // NDEBUG
+    return R;
+  }
+};
+
 IncrementalJIT::IncrementalJIT(
     IncrementalExecutor& Executor, const clang::CompilerInstance &CI,
     std::unique_ptr<llvm::orc::ExecutorProcessControl> EPC, Error& Err,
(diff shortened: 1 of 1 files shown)


guitargeek (changes requested):
Hi @vgvassilev! This looks like something that could be useful for the AI so it can debug better, so I got this PR reviewed by the AI. Below is what it thinks. I think at least the first point should be addressed before this can get merged.

--------------

Thanks for this — the ability to get the exact IR and object bytes out of the JIT is
something I've wanted more than once when chasing ABI mismatches.

I applied the patch on top of current master (it still applies cleanly) and tested it
properly rather than just reading it:

- `clingInterpreter` builds clean, 0 errors and 0 warnings.
- Recompiling `IncrementalJIT.cpp` with `-UNDEBUG` so the guarded block is actually
  live: also clean, 0 warnings.
- Running it end to end, I get e.g. `cling-module-6.1.{ll,o}`, `cling-module-0.2.{ll,o}`
  and `__standard_lib.3.{ll,o}`. The `.o` is a valid ELF relocatable containing the
  expected `__cling_N50::__cling_Un1Qu30` wrapper and the variable from my test line,
  and `llvm-as` accepts the `.ll`. So the feature genuinely works and the artifacts are
  exactly what you'd want.

The idea and the implementation are sound, and it's risk-free for release builds since
it compiles out entirely. A few things I'd like addressed before I approve.

---

## 1. In a default ROOT build this is silently a no-op

This is the one that worries me most, because the failure mode is silence rather than
an error.

`CMakeLists.txt` strips `-[UD]NDEBUG` from the build-type flags **only when
`asserts=ON`**:

```cmake
#---Enable asserts---
if(asserts)
  string(REGEX REPLACE "-[UD]NDEBUG(=.*)?" "" "CMAKE_CXX_FLAGS_${_BUILD_TYPE_UPPER}" ...)
```

`asserts` defaults to `OFF` (ON only for `CMAKE_BUILD_TYPE=Debug` and/or `dev=ON`), and
`RelWithDebInfo` — what most of us build — carries `-O2 -g -DNDEBUG`. I confirmed the
whole block is absent from the resulting object file: the `"Created file"` string count
in `IncrementalJIT.cpp.o` is 0 in a default build and 1 with `-UNDEBUG`.

LLVM registers the `-debug-only` option under `#ifndef NDEBUG` as well, so on such a
build the invocation from the PR description behaves like this:

```
$ EXTRA_CLING_ARGS=" --this-option-does-not-exist " root.exe -l -b -q
error: unknown argument: '--this-option-does-not-exist'

$ EXTRA_CLING_ARGS=" --debug-only=orc " root.exe -l -b -q
        # accepted, no error, no output, no files
```

Someone following the instructions on a normal build gets nothing and no hint why.
Could the PR description (and ideally a line in the commit message) say that this
requires `-Dasserts=ON` or `-Ddev=ON`?

## 2. `-debug` on its own will dump files

`isCurrentDebugType()` returns true when *no* `-debug-only` was given at all:

```cpp
bool isCurrentDebugType(const char *DebugType, int Level) {
  if (CurrentDebugType->empty())
    return true;
```

So `bool Debugging = DebugFlag && isCurrentDebugType(DEBUG_TYPE);` is effectively just
`DebugFlag`, and a plain `-debug` — which is a natural first thing to try — writes two
files per compiled module into the current directory. On a real session that is a lot
of files appearing unannounced in whatever directory the user happened to be in.

## 3. Please use a cling-specific debug type

`DEBUG_TYPE "orc"` is LLVM's own ORC debug type, so `-debug-only=orc` also turns on all
of upstream ORC's tracing, which is very verbose. Something like `"cling-jit"` would
make this independently selectable and wouldn't collide with upstream.

Related and minor: the manual guard uses `isCurrentDebugType(DEBUG_TYPE)`, i.e. level 0,
while `LLVM_DEBUG` expands to `isCurrentDebugType(TYPE, 1)`, level 1. With
`-debug-only=orc:0` the files are written but the "Created file" messages are
suppressed. Worth making the two consistent.

## 4. The `std::error_code` is never checked

```cpp
std::error_code EC;
llvm::raw_fd_ostream ModuleStream(FileStem + ".ll", EC);
ModuleStream << M;
llvm::raw_fd_ostream ObjectStream(FileStem + ".o", EC);
```

`EC` is passed to both constructors and never inspected (and reused across the two). If
the working directory isn't writable, both writes silently do nothing and the
`LLVM_DEBUG` line still claims the files were created. A check and an early warning
would save someone a confusing debugging session.

## 5. The class should be in the anonymous namespace

As placed, `DebuggingSimpleCompiler` lands inside `namespace cling` (it goes in right
after `IncrementalJIT::~IncrementalJIT()`, which is within the `cling` block), so a
file-local implementation detail ends up with external linkage — `nm` shows
`cling::DebuggingSimpleCompiler::operator()` and its vtable as weak symbols. The file
already has an unnamed namespace higher up; moving the class there would keep it
internal.

## 6. Smaller points

- The output files always land in the current working directory with no way to redirect
  them. An env var or an `-mllvm` option would make this much more pleasant for a long
  session inside a source tree.
- `#define DEBUG_TYPE` / `#undef DEBUG_TYPE` inside the member function body works, but
  the LLVM convention is to define it once near the top of the file. No conflict today
  since `IncrementalJIT.cpp` doesn't define `DEBUG_TYPE` anywhere else — just
  surprising to read.
- The description says the files are named `cling-module-X.Y.{o,ll}`; in practice other
  module identifiers show up too, e.g. `__standard_lib.3.ll`. Harmless, but worth
  wording it as "the module identifier" instead.
- Not a problem today, just noting it: the `static unsigned counter` is incremented
  without synchronization. Since cling never calls `setNumCompileThreads` the compile
  callback runs on the calling thread, so this is fine — but it would quietly become a
  race if compile threads were ever enabled.

vgvassilev:
Yes, we can implement a trigger that we are comfortable with...

Options are: either `gDebug` level but that variable is too overloaded; or through a new env variable that enables it... Ideally ROOT if ROOT had a more flexible debugging system (as the one of llvm) we could do something else.

PR #18587, opened 2025-05-02
[meta] Correct TClass::LoadClassInfo.

### LINKED ISSUE #18556: Class methods not available to python in cases of transient data members of private nested class type
### Check duplicate issues.

- [x] Checked for duplicates

### Description

hi -

With recent versions on the ROOT trunk, we've been seeing a new
dictionary issue.  If a class has a (transient) data member the type of which
is a private nested class, then depending on the order in which dictionaries
are created, the class methods may not be visible from Python.

A standalone reproducer is provided in the attached zip file.
This was tested using lcg dev3 nightly build from Apr. 26
(/cvmfs/sft-nightlies.cern.ch/lcg/nightlies/dev3/Sat).

```
$ ./setup   # set up ROOT
$ ./build   # build dictionary shared library.
$ python write.py # write a TTree using those classes.
$ python read.py # Open it and test for pythonization of method.
  File "/afs/cern.ch/user/s/ssnyder/atlas-work3g/dicttest/read.py", line 7, in <module>
    s = cl.testMethod
        ^^^^^^^^^^^^^
AttributeError: <class cppyy.gbl.Base at 0x797a510> has no attribute 'testMethod'. Full details:
  type object 'Base' has no attribute 'testMethod'
  'Base::testMethod' is not a known C++ class
  'testMethod' is not a known C++ template
  'testMethod' is not a known C++ enum
```

So we cannot find the methods of the class Base from python.

The error goes away if i uncomment the third line of read.py to create
the dictionary information earlier.

This was discussed in this ATLAS issue:
https://its.cern.ch/jira/browse/ATEAM-1071

We worked around it for now by adding a dictionary for the private class.

This appeared in our nightly builds against dev3 LCG as of about Apr. 15.
Bisecting, the first ROOT commit from which i see this issue is this:

```
commit 87676c81c50cf7d85ded6f42a28f4abf445ff63f
Author: Philippe Canal <pcanal@fnal.gov>
Date:   Thu Aug 8 18:07:49 2024 -0500

    TProtoClass: do not complaint about innards of transients.
    
    Do not complain about missing information about data members that directly or indirectly within
    a transient members of the top level class in GetRealData.
    
    This solves the underlying problem from https://github.com/root-project/root/pull/15733
```

A further note from the ATLAS issue:

Not completely sure yet, but this seems to be related to
TClass::LoadClassInfo:

```
   ClassInfo_t       *GetClassInfo() const {
      if (fCanLoadClassInfo)
         LoadClassInfo();
      return fClassInfo;
   }

void TClass::LoadClassInfo() const
{
   R__LOCKGUARD(gInterpreterMutex);

   // Return if another thread already loaded the info
   // while we were waiting for the lock
   if (!fCanLoadClassInfo || TestBit(kLoading))
      return;

   bool autoParse = !gInterpreter->IsAutoParsingSuspended();

   if (autoParse)
      gInterpreter->AutoParse(GetName());

   if (!fClassInfo)
      gInterpreter->SetClassInfo(const_cast<TClass *>(this));

   if (autoParse && !fClassInfo) {
      if (fImplFileLine == -1 && fClassVersion == 0) {
         // We have a 'transient' class with a ClassDefInline and apparently no interpreter
         // information. Since it is transient, it is more than likely that the lack
         // will be harmles.
      } else {
         ::Error("TClass::LoadClassInfo", "no interpreter information for class %s is available"
                                          " even though it has a TClass initialization routine.",
                 fName.Data());
      }
      return;
   }

   fCanLoadClassInfo = false;
}
```

When this fails, this is getting called with autoloading disabled.
In that case, the classinfo is not loaded, but further fCanLoadClassInfo
is set to false, so it won't be retried later when autoloading is
available.

### Reproducer

[dicttest.zip](https://github.com/user-attachments/files/19966896/dicttest.zip)

### ROOT version

lcg dev3

### Installation method

lcg

### Operating system

lxplus

### Additional context

_No response_

pcanal:
This fixes #18556.

We should keep trying to try to load a `ClassInfo` until we try the auto-parsing.  `TClass::LoadClassInfo` should set `fCanLoadClassInfo` to `false` when the `ClassInfo` is found *and* set it to false also when we tried auto parsing but did not find the `ClassInfo`.

In the later case, the `ClassInfo` might still be loaded (via a different routine/mechanims) if information about the class is later loaded/parsed.

### DIFF
--- a/core/meta/src/TClass.cxx
+++ b/core/meta/src/TClass.cxx
@@ -5971,10 +5971,10 @@ void TClass::LoadClassInfo() const
 
    bool autoParse = !gInterpreter->IsAutoParsingSuspended();
 
-   if (autoParse)
+   if (autoParse && !fClassInfo)
       gInterpreter->AutoParse(GetName());
 
-   if (!fClassInfo)
+   if (!fClassInfo) // Could be indirectly set by the parsing
       gInterpreter->SetClassInfo(const_cast<TClass *>(this));
 
    if (autoParse && !fClassInfo) {
@@ -5987,10 +5987,13 @@ void TClass::LoadClassInfo() const
                                           " even though it has a TClass initialization routine.",
                  fName.Data());
       }
-      return;
    }
 
-   fCanLoadClassInfo = false;
+   // Keep trying to load the ClassInfo, since we have no ClassInfo yet,
+   // we will get an update even when there is an explicit load.  So whether
+   // or not the autoparsing is on, we will need to keep trying to load
+   // the ClassInfo.
+   fCanLoadClassInfo = !fClassInfo;
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/roottest/root/meta/tclass/CMakeLists.txt
+++ b/roottest/root/meta/tclass/CMakeLists.txt
@@ -93,18 +93,20 @@ ROOTTEST_ADD_TEST(datamemberload
 ROOTTEST_GENERATE_DICTIONARY(libInitOrderDups
                              InitOrderDups.h
                              LINKDEF InitOrderDupsLinkDef.h
-                             NO_ROOTMAP NO_CXXMODULE)
+                             NO_ROOTMAP NO_CXXMODULE
+                             FIXTURES_SETUP InitOrder)
 
 ROOTTEST_GENERATE_DICTIONARY(libInitOrder
                              InitOrder.h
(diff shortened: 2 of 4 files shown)


dpiparo (approved):
Thanks for these changes. I propose to have the fix in 6.36 too.

PR #18591, opened 2025-05-03
[ci] Add AlmaLinux 10

guitargeek:
The AlmaLinux 10 stable preview is now available:

https://wiki.almalinux.org/development/almalinux-os-kitten-10.html#mirrors-isos-and-update-frequency


dpiparo (approved):
Thanks for this work: awesome. 
I added a minor change commit to make sure the alma10 builds are for the moment considered "special".

dpiparo:
I think we should backport this to the 6.36 branch.

PR #18592, opened 2025-05-03
[tmva][sofie] Mark input arrays to interence function as `const`

guitargeek:
This is good practice, as it makes it clear to the reader and compiler that the data will not be modified. This also helps Clad in generating optimized gradients. Actually, it is a requirement for differentiation through the emitted code with the new Clad v1.10, which is now part of ROOT.

In a separate commit, refactor the reading of tensors from an input stream to avoid duplicating boilerplate code in the generated inference session code.


lmoneta (approved):
LGTM!
Thanks Jonas

PR #18593, opened 2025-05-03
[ci] Set `builtin_nlohmannjson=OFF` also for AlmaLinux and Fedora

guitargeek:
When these configurations were created, it was thought that AlmaLinux and Fedora don't have any package for nlohmann-json. I guess that was an understandable oversight, because the package doesn't have "nlohmann" in the name. It's just called `json-devel`, and was added to all relevant CI images earlier today.

### DIFF
--- a/.github/workflows/root-ci-config/buildconfig/alma9-clang.txt
+++ b/.github/workflows/root-ci-config/buildconfig/alma9-clang.txt
@@ -1,4 +1,3 @@
-builtin_nlohmannjson=ON
 builtin_vdt=On
 tmva-sofie=On
 ccache=On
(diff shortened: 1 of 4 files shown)


dpiparo (approved):
Thanks for these changes. I think we should also have them in 6.36, if possible.

pcanal:
Side question, do we still have at least one CI build that are using this (and other now less often tested) builtins?

guitargeek:
All the macs and Windowses still use it.

By the way, especially for `nlohmann_json` it's important to use the system library if available, because it's a dependency at the interface level. So to avoid mixing different `nlohmann_json` versions, one should avoid the builtin, especially in the release binaries that we're shipping

PR #18594, opened 2025-05-03
[v626][ci] Also run on `alma9`

guitargeek:
AlmaLinux 9 is the main platform supported by ROOT, so if the 6.26 build can also succeed for this platform, it would be good to add it to the CI matrix. Like this we can be sure that there is still some way to build older ROOT releases.


guitargeek:
Didn't work as easily as I expected.

PR #18595, opened 2025-05-03
[ci] Set `builtin_vdt=OFF` also for AlmaLinuxes and Fedora 41

guitargeek:
Most likely, there is also no need to use the builtin VDT anymore. For example, Fedora 42 already disables it.


guitargeek:
Ah, my bad. Vdt is not available on EPEL.

PR #18598, opened 2025-05-04
[CMake] Update `builtin_cling=OFF` configuration code

guitargeek:
In an attempt to further reduce build times, I attempted to build ROOT with `builtin_cling=OFF`, using a custom cling build that I install on my system.

However, I ran into several errors with cling header files not being found, as well as linker errors.

The simplest fix was to set the `GLOBAL` argument in `find_package(Cling)`, to make all the Cling targets available globally. Then, one also doesn't need all this code for the transient dependencies anymore (this is all handled correctly in the ClingConfig.cmake).

The only caveat of this is that the `GLOBAL` argument is only available from CMake 3.24 onwards, so the configuration with `builtin_cling=OFF` will now error out if the CMake version is lower. That should be acceptable, since users that are adventurous enough to build ROOT like this probably also have a CMake version that is new enough (RHEL 9 has CMake version 3.26, and Ubuntu 24.04 has CMake version 3.28).

Also, don't add any custom search paths to `find_package(Cling)` and `find_package(Clang)`, so that one is forced to use the `Clang_DIR` and `Cling_DIR` hints correctly.

### DIFF
--- a/core/metacling/src/CMakeLists.txt
+++ b/core/metacling/src/CMakeLists.txt
@@ -85,18 +85,6 @@ add_dependencies(MetaCling CLING clangCppInterOp)
 
 ##### libCling #############################################################
 
-if(NOT builtin_clang)
-  set(prefixed_link_libraries)
-  foreach(dep ${CLING_DEPEND_LIBS})
-    if("${dep}" MATCHES "^clang")
-      set(dep "${LLVM_LIBRARY_DIR}/lib${dep}.a")
-     endif()
-     list(APPEND prefixed_link_libraries "${dep}")
-   endforeach()
-  set(LINK_LIBS "${prefixed_link_libraries}")
-  link_directories("${LLVM_LIBRARY_DIR}")
-endif()
-
 # We need to paste the content of the cling plugins disabling link symbol optimizations.
 set(CLING_PLUGIN_LINK_LIBS)
 if (clad)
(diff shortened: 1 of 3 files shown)


vgvassilev:
Can we improve the Cling cmake infrastructure when it is being discovered as a installed package. I do not think we ever invested time in this and it's probably worth it.

guitargeek:
Which part of the CMake do you mean exactly? With this PR, building ROOT with `builtin_cling=OFF` seems to work, at least for the case with `clad=OFF`. For `clad=ON`, I still get linker errors, even if I build Cling with `CLING_BUILD_PLUGINS=ON` and `clad=ON`. So that's something that could be fixed. You see other problems?

vgvassilev:
> Which part of the CMake do you mean exactly? With this PR, building ROOT with `builtin_cling=OFF` seems to work, at least for the case with `clad=OFF`. For `clad=ON`, I still get linker errors, even if I build Cling with `CLING_BUILD_PLUGINS=ON` and `clad=ON`. So that's something that could be fixed. You see other problems?

I meant to avoid needing `GLOBAL`.

hahnjo:
@guitargeek I would appreciate if you could start adding me as reviewer for this kind of PRs.......

guitargeek:
There are merge conflicts, and I have managed to reduce build times by other means to I don't plan anymore to use `builtin_cling=OFF`.

PR #18601, opened 2025-05-05
[core] Allow using RError with non-default-constructible types

silverweed:
We now wrap fValue in a std::optional.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/foundation/inc/ROOT/RError.hxx
+++ b/core/foundation/inc/ROOT/RError.hxx
@@ -189,15 +189,15 @@ RResult<T>::Unwrap() can be used as a short hand for
 int value = FuncThatReturnsRResultOfInt().Unwrap();  // may throw
 ~~~
 
-There is no implict operator that converts RResult<T> to T. This is intentional to make it clear in the calling code
+There is no implicit operator that converts RResult<T> to T. This is intentional to make it clear in the calling code
 where an exception may be thrown.
 */
 // clang-format on
 template <typename T>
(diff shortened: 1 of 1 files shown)


core/foundation/inc/ROOT/RError.hxx
```diff
@@ -189,15 +189,16 @@ RResult<T>::Unwrap() can be used as a short hand for
...
 */
 // clang-format on
 template <typename T>
 class RResult : public RResultBase {
 private:
-   /// The result value in case of successful execution
-   T fValue;
+   /// The result value in case of successful execution.
+   // Note that if T is not default constructible we wrap it in an optional to allow for an "empty" value.
+   std::conditional_t<std::is_default_constructible_v<T>, T, std::optional<T>> fValue;
```
hahnjo, line 201: Do we maybe want to just always wrap it in a `std::optional`? That would simplify the code and convey the right semantics...

silverweed: That would be suboptimal for the majority of cases (default constructible `T`) and imho not really justified given how little code is necessary to support the conditional `optional`.

pcanal: Indeed, even a default constructor can be expensive. +1 for just always use `std::optional`.

hahnjo (approved):
LGTM; I restarted the one failing job, but this may also be related to incremental PR builds...

PR #18607, opened 2025-05-05
[treeformula] Properly process submembers when param is not single char

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-6741

Note: this is an empirical fix, seems to work now, but not sure if it breaks other use cases.

Maybe related: https://its.cern.ch/jira/browse/ROOT-7686

## Checklist:

- [x] tested changes locally

This PR fixes #


tree/tree/test/TTreeRegressions.cxx
```diff
@@ -230,3 +232,62 @@ TEST(TTreeRegressions, EmptyLeafObject)
...
+   {
+      for (size_t i = 0; i < sub.size(); ++i)
+         if (sub[i].id == id)
+            return &sub[i];
+      return 0;
+   }
+};
+TEST(TTreeRegressions, TTreeFormulaMemberIndex)
+{
+   gInterpreter->Declare("class MySubClass { public: MySubClass(int Id = 0, double X = 0) : id(Id), x(X) {} virtual ~MySubClass() {} int id; double x; };");
```
pcanal, line 259: The code duplication here (and the next line) is sub-optimal.  Would it be possible to use a (private) header file instead?

tree/treeplayer/src/TTreeFormula.cxx
```diff
@@ -2352,7 +2352,7 @@ Int_t TTreeFormula::FindLeafForExpression(const char* expression, TLeaf*& leaf,
             final = true;
 
             // Record in 'i' what we consumed
-            i += strlen(params);
+            i += 2; // open and close parentheses
```
pcanal, line 2355: That does not 'sound' sufficient unless 'i' is incremented by the size of the params content somewhere else, is it?

tree/tree/test/TTreeRegressions.cxx
ferdymercury: I guess so, (though I copy-pasted the structure from the test by Enrico in the same file above), so not sure if it's worth to change all nor to change just this one (inconsistency) ? Also if someone deletes the test later on he might forget to delete the header file, this way it's more self-contained?

Maybe a trick with a preprocessor stringification could be a good compromise to avoid duplication? (not fully sure if it's possible)

tree/treeplayer/src/TTreeFormula.cxx
ferdymercury: I don't really follow the logic in this function, but this was needed in order to make the check at line 2619 work. So it was an ad-hoc solution.

Another way of solving this would be to record strlen(params) in an auxiliary variable and then do:

`if (i < nchname + paramLength - 2)`

would that be better?

ferdymercury: I checked to do this before my patch, and I get:

`mc.Get(1)->x` and I get 9 < 10
`mc.Get(12)->x` and I get 12 < 11
`mc.Get(123)->x` and I get 14 < 12

so the difference is increasing a factor of 2 per char. So it seems to me that without my patch, length(params) is being summed twice?

pcanal: Humm .. you are right, `i` is already pointing to the closing parenthesis at that point, so actually it feels like:
```
i += 0; // nothing to do really
```
seems like the 'right' number (i.e. removing the line entirely) ... but the fact that it sort of work sometime with 2 or some other random number is puzzling :)

tree/tree/test/TTreeRegressions.cxx
```diff
@@ -234,24 +234,17 @@ TEST(TTreeRegressions, EmptyLeafObject)
...
+MYSUBCLASS
+MYCLASS
+#define STRINGIFY(string) #string
+#define TO_LITERAL(string) STRINGIFY(string)
+
 TEST(TTreeRegressions, TTreeFormulaMemberIndex)
 {
-   gInterpreter->Declare("struct MySubClass { int id; double x; };");
-   gInterpreter->Declare("struct MyClass { std::vector<MySubClass> sub; MySubClass *Get(int id) { for (size_t i = 0; i < sub.size(); ++i) if (sub[i].id == id) return &sub[i]; return nullptr; } };");
+   gInterpreter->Declare(TO_LITERAL(MYSUBCLASS));
```
pcanal, line 246:
```suggestion
   gInterpreter->Declare(_QUOTE_(MYSUBCLASS));
```
if that does not work:
```
#define TO_LITERAL(string) _QUOTE_(string)
```

tree/tree/test/TTreeRegressions.cxx
```diff
@@ -230,3 +232,51 @@ TEST(TTreeRegressions, EmptyLeafObject)
    TLeafObject tlo;
    EXPECT_EQ(tlo.GetObject(), nullptr);
 }
+
+// https://its.cern.ch/jira/browse/ROOT-6741
+#define MYSUBCLASS struct MySubClass { int id; double x; };
+#define MYCLASS struct MyClass { std::vector<MySubClass> sub; MySubClass *Get(int id) { for (size_t i = 0; i < sub.size(); ++i) if (sub[i].id == id) return &sub[i]; return nullptr; } };
+MYSUBCLASS
+MYCLASS
+#define STRINGIFY(string) #string
```
ferdymercury, line 241:
```suggestion
```

tree/tree/test/TTreeRegressions.cxx
```diff
@@ -230,3 +232,51 @@ TEST(TTreeRegressions, EmptyLeafObject)
...
+#define TO_LITERAL(string) STRINGIFY(string)
```
ferdymercury, line 242:
```suggestion
#define TO_LITERAL(string) _QUOTE_(string)
```

pcanal (approved):
Thanks.

PR #18609, opened 2025-05-05
[ntuple] RNTupleMerger: use RLogger instead of TError-style logging

silverweed:
For consistency with the rest of the RNTuple codebase, use the RLogger macros rather than the TError ones.
Notably, this changes the log category from `RNTuple::Merge` to `ROOT.NTuple`.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/test/ntuple_merger.cxx
+++ b/tree/ntuple/test/ntuple_merger.cxx
@@ -742,7 +742,7 @@ TEST(RNTupleMerger, MergeThroughTFileMergerKey)
 {
    ROOT::TestSupport::CheckDiagsRAII diags;
    diags.requiredDiag(kWarning, "TFileMerger", "Merging RNTuples is experimental");
-   diags.requiredDiag(kError, "RNTuple::Merge", "Output file already has key, but not of type RNTuple!");
+   diags.requiredDiag(kError, "ROOT.NTuple", "Output file already has key, but not of type RNTuple!");
    diags.requiredDiag(kError, "TFileMerger", "Could NOT merge RNTuples!");
    diags.requiredDiag(kError, "TFileMerger", "error during merge of your ROOT files");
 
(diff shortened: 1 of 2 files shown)


hahnjo (approved):
LGTM

tree/ntuple/test/ntuple_merger.cxx
```diff
@@ -1949,7 +1949,7 @@ TEST(RNTupleMerger, MergeAsymmetric1TFileMerger)
          fileMerger.SetMergeOptions(TString("rntuple.MergingMode=Strict"));
          CheckDiagsRAII diags;
          diags.requiredDiag(kError, "TFileMerger::Merge", "error during merge", false);
-         diags.requiredDiag(kError, "RNTuple::Merge", "missing the following field", false);
+         diags.requiredDiag(kError, "ROOT.NTuple", "missing the following field", false);
```
pcanal, line 1952: I fully support the move from `TError` to `RLogger`.  It seems we need to add more details to the messages though.  The old messages are, i.e. something like:
```
"Error in RNTuple::Merge: missing the following field ..."
```
has more information (namely the activity that failed) than the new message, something like (assuming I am reading correctly):
```
Error ROOT.NTuple: missing the following field ..."
```

Unless the user provides additional context, this message could 'a priori' come from any operation that expects/needs a field.  

In the output of a large framework run, this might make the difference between knowing immediately where the problem is and not knowing until having instrumented the (user) code and/or put a breakpoint in the error logger

pcanal (commented):
This likely need a follow up.

tree/ntuple/test/ntuple_merger.cxx
silverweed: That makes sense. In this case, would it be fine to use a different log channel, like `ROOT.NTuple.Merge`?

pcanal: I think so.

PR #18612, opened 2025-05-06
[ntuple] Save a memcpy when reading a compressed Anchor in RMiniFile

silverweed:
Do:
  (read into unzipBuf) -> (unzip into finalBuf)

instead of:
  (read into finalBuf) -> (unzip into unzipBuf) -> (copy back to finalBuf)

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/src/RMiniFile.cxx
+++ b/tree/ntuple/src/RMiniFile.cxx
@@ -763,25 +763,27 @@ ROOT::RResult<ROOT::RNTuple> ROOT::Internal::RMiniFileReader::GetNTupleProper(st
    if (key.fObjLen < kMinNTupleSize) {
       return R__FAIL("invalid anchor size: " + std::to_string(key.fObjLen) + " < " + std::to_string(sizeof(RTFNTuple)));
    }
+
    // The object length can be smaller than the size of RTFNTuple if it comes from a past RNTuple class version,
    // or larger than it if it comes from a future RNTuple class version.
    auto bufAnchor = MakeUninitArray<unsigned char>(std::max<size_t>(key.fObjLen, sizeof(RTFNTuple)));
    RTFNTuple *ntuple = new (bufAnchor.get()) RTFNTuple;
 
-   auto objNbytes = key.GetSize() - key.fKeyLen;
-   ReadBuffer(ntuple, objNbytes, offset);
+   const auto objNbytes = key.GetSize() - key.fKeyLen;
    if (objNbytes != key.fObjLen) {
-      // Decompress into a temporary buffer
-      auto unzipBuf = MakeUninitArray<unsigned char>(key.fObjLen);
-      RNTupleDecompressor::Unzip(bufAnchor.get(), objNbytes, key.fObjLen, unzipBuf.get());
-      // Then copy back to bufAnchor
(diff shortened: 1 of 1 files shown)


tree/ntuple/src/RMiniFile.cxx
```diff
@@ -763,25 +763,27 @@ ROOT::RResult<ROOT::RNTuple> ROOT::Internal::RMiniFileReader::GetNTupleProper(st
...
-      RNTupleDecompressor::Unzip(bufAnchor.get(), objNbytes, key.fObjLen, unzipBuf.get());
-      // Then copy back to bufAnchor
-      memcpy(bufAnchor.get(), unzipBuf.get(), key.fObjLen);
+      // Read into a temporary buffer
+      auto unzipBuf = MakeUninitArray<unsigned char>(std::max<size_t>(key.fObjLen, sizeof(RTFNTuple)));
+      ReadBuffer(unzipBuf.get(), objNbytes, offset);
+      // Unzip into the final buffer
+      RNTupleDecompressor::Unzip(unzipBuf.get(), objNbytes, key.fObjLen, ntuple);
+   } else {
+      ReadBuffer(ntuple, objNbytes, offset);
```
hahnjo, line 780: I don't really think this is on the critical path, and now we have two calls to `ReadBuffer`...

silverweed: What do you mean two calls? They're mutually exclusive so we only have one; we're simply doing one less operation in one of the branches.

hahnjo: There are two lines of code where we call `ReadBuffer`, which we will have to make sure updated in sync. That current code is more maintainable.

silverweed: Why is `ReadBuffer` harder to maintain than `memcpy`? That too needs to be synchronized, and both are trivial things to do

pcanal: Skipping a useless mempcy can be a significant performance difference, for example in case of tight looping over a single branch and is worth the slight extra complexity.

hahnjo: This code is about the anchor which is read once when opening the file. There is simply no performance implication in that.

hahnjo: which btw also means that the amount of memory copied is 78 bytes...

pcanal: I missed that ... indeed for such a low size ... even doing the compression is debatable ....

hahnjo: For `RMiniFile`, this was explicitly implemented in https://github.com/root-project/root/commit/7b03bccd5737843fa8049b81e704b4a009e9c419. I believe `TFile` always attempts compression and goes with it if beneficial, ie compressed size smaller than uncompressed.

silverweed: This change is not about performance at all, it's about code simplicity and straightforwardness. This code does less work and therefore is simpler to understand. However if I'm the only one who thinks so, this PR can be closed and not much of value would be lost, being such a small change.

PR #18619, opened 2025-05-06
[ntuple] Fix setting of trivial traits for `RStreamerField`

hahnjo:
Before, classes without explicit constructor / destructor but containing a non-trivial member would still be flagged as trivial.

This fixes the first of two issues causing crashes found by CMS: https://github.com/cms-sw/cmssw/issues/48005

### DIFF
--- a/tree/ntuple/src/RFieldMeta.cxx
+++ b/tree/ntuple/src/RFieldMeta.cxx
@@ -828,9 +828,12 @@ ROOT::RStreamerField::RStreamerField(std::string_view fieldName, TClass *classp)
      fIndex(0)
 {
    fTraits |= kTraitTypeChecksum;
-   if (!(fClass->ClassProperty() & kClassHasExplicitCtor))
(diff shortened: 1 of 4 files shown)


tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -828,9 +828,12 @@ ROOT::RStreamerField::RStreamerField(std::string_view fieldName, TClass *classp)
      fIndex(0)
 {
    fTraits |= kTraitTypeChecksum;
-   if (!(fClass->ClassProperty() & kClassHasExplicitCtor))
+   // For RClassField, we only check for explicit constructors and destructors and then recursively combine traits from
+   // all member subfields. For RStreamerField, we treat the class as black box and additionally need to check for
```
enirolf, line 832:
```suggestion
   // all member subfields. For RStreamerField, we treat the class as a black box and additionally need to check for
```

enirolf (approved):
LGTM, thanks!

jblomer (approved):
Thanks!

PR #18620, opened 2025-05-06
[ntuple] Correctly construct values in staging area

hahnjo:
... and delete them in the destructor of `RClassField`. This is important for non-trivial types where zero-initialization is not sufficient or that need to free resources in their destructor (for example a `std::vector` constructed by an `RVectorField`).

This fixes the second issue causing crashes found by CMS: https://github.com/cms-sw/cmssw/issues/48005

### DIFF
--- a/tree/ntuple/src/RFieldMeta.cxx
+++ b/tree/ntuple/src/RFieldMeta.cxx
@@ -366,7 +366,8 @@ void ROOT::RClassField::PrepareStagingArea(const std::vector<const TSchemaRule *
 
    if (stagingAreaSize) {
       R__ASSERT(static_cast<Int_t>(stagingAreaSize) <= fStagingClass->Size()); // we may have removed rules
-      fStagingArea = ROOT::Internal::MakeUninitArray<unsigned char>(stagingAreaSize);
+      // We use std::make_unique instead of MakeUninitArray to zero-initialize the staging area.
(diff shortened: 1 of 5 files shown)


dpiparo (approved):

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -366,7 +378,14 @@ void ROOT::RClassField::PrepareStagingArea(const std::vector<const TSchemaRule *
 
    if (stagingAreaSize) {
       R__ASSERT(static_cast<Int_t>(stagingAreaSize) <= fStagingClass->Size()); // we may have removed rules
-      fStagingArea = ROOT::Internal::MakeUninitArray<unsigned char>(stagingAreaSize);
+      // We use std::make_unique instead of MakeUninitArray to zero-initialize the staging area.
```
pcanal, line 381: Is the zero-initialize redundant with the constructor calls or do we also need it for the trivial cases?

hahnjo: Strictly speaking it may not be necessary for trivial default constructors, but I think it's better to start with a defined value than uninitialized memory. It should not cost since this code is run once when connecting `RClassField` to a page source.

jblomer (approved):
Thanks!

Issue #18401, opened 2025-04-14
valgrind reports conditional jump on uninitialized value in llvm::ConstantExpr::getGetElementPtr


Dr15Jones:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

While running valgrind on a job using https://github.com/hep-cce2/root_serialization with a CMS AOD file, the job reported

```
==1738875==    at 0xCC74F90: llvm::ConstantExpr::getGetElementPtr(llvm::Type*, llvm::Constant*, llvm::ArrayRef<llvm::Value*>, bool, std::optional<unsigned int>, llvm::Type*) (in /cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02884/el8_amd64_gcc12/lcg/root/6.35.1-10a0985d87af2d932b7c4640d485bb50/lib/libCling.so)
==1738875==    by 0x8A5BA2B: (anonymous namespace)::ItaniumRTTIBuilder::BuildVTablePointer(clang::Type const*) (in /cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02884/el8_amd64_gcc12/lcg/root/6.35.1-10a0985d87af2d932b7c4640d485bb50/lib/libCling.so)                             
==1738875==    by 0x8A5C23F: (anonymous namespace)::ItaniumRTTIBuilder::BuildTypeInfo(clang::QualType) (in /cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02884/el8_amd64_gcc12/lcg/root/6.35.1-10a0985d87af2d932b7c4640d485bb50/lib/libCling.so)                                     
==1738875==    by 0x8A5DF12: (anonymous namespace)::ItaniumCXXABI::getAddrOfRTTIDescriptor(clang::QualType) (in /cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02884/el8_amd64_gcc12/lcg/root/6.35.1-10a0985d87af2d932b7c4640d485bb50/lib/libCling.so)                                
==1738875==    by 0x8A637A7: (anonymous namespace)::ItaniumCXXABI::emitVTableDefinitions(clang::CodeGen::CodeGenVTables&, clang::CXXRecordDecl const*) (in /cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02884/el8_amd64_gcc12/lcg/root/6.35.1-10a0985d87af2d932b7c4640d485bb50/lib/libCling.so)
[...]

Issue #15996, opened 2024-07-05
cppyy.ll.cast() does not understand the type of a branch (invocation through GetBranchAttr)


Nowakus:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

The new cppyy produces an error when we try to call GetBranchAttr on one of our DataVector branches. 

### Reproducer

# reproducer using only the ATLAS environment, no files necessary
export ATLAS_LOCAL_ROOT_BASE=/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase 
source $ATLAS_LOCAL_ROOT_BASE/user/atlasLocalSetup.sh
asetup main--dev4LCG,latest,Athena
python
    import ROOT, cppyy.ll
    ROOT.TClass.GetClass('DataVector<xAOD::PixelCluster_v1>').HasDictionary()
True
    cppyy.ll.cast['DataVector<xAOD::PixelCluster_v1>'](0)
TypeError: Could not find "cppyy_cast<DataVector<xAOD::PixelCluster_v1>>" (set cppyy.set_debug() for C++ errors):
  Failed to instantiate "cppyy_cast<DataVector<xAOD::PixelCluster_v1>>(int)"

# but using a typedef to the DataVector works (note the real type of the returned object):
    cppyy.ll.cast['xAOD::PixelClusterContainer_v1>'](0)
<cppyy.gbl.DataVector<xAOD::PixelCluster_v1> object at 0x11cbd2c0>
# and after the previous call autoloaded some libraries, the real type starts to work as well:
    cppyy.ll.cast['DataVector<xAOD::PixelCluster_v1>'](0)
<cppyy.gbl.DataVector<xAOD::PixelCluster_v1> object at 0x1f542770>

### ROOT version

6.32+, possibly since 6.30

### Installation method

SFT dev nightlies

### Operating system

alma9

amete:
Hi @guitargeek, has there been any progress on this? Many thanks.

guitargeek:
As discussed offline, the underlying reason seems to be that the forward declaration of `DataVector` is inconsistent with the definition.

Here is the [forward declaration](https://gitlab.cern.ch/atlas/athena/-/blob/main/Control/AthContainers/AthContainers/DataVector.h#L581):
```c++
template <class T, class BASE> class DataVector;
```

And here is the [definition](https://gitlab.cern.ch/atlas/athena/-/blob/main/Control/AthContainers/AthContainers/DataVector.h#L793):
```c++
[...]

Issue #10166, opened 2022-03-18
`std::pair` return by CollectionProxy for std::map/multilmap is missing the `const` 


pcanal:
Currently we have:
```
root [1] TClass::GetClass("map<int,double>")->GetCollectionProxy()->GetValueClass()->GetName()
(const char *) "pair<int,double>"
```
where as the C++ standard requires:
```
root [1] TClass::GetClass("map<int,double>")->GetCollectionProxy()->GetValueClass()->GetName()
(const char *) "pair<const int,double>"
```
As the binary format for both `pairs` is exactly the same this is a minor issue.

However a consequence is that the auto-generation of the `TClass` for the `const` version will likely fail in some circumstances.  See https://github.com/root-project/root/pull/10132/files#r829931867

ferdymercury:
Maybe related: https://github.com/root-project/root/issues/10512

still the same in 6.38

Issue #8470, opened 2021-06-17
Desugaring involving unknown type name sometimes replace that name with the text "int"


pcanal:
```
TClass::GetClass("vector<int(const Zyx*)>")->GetName()
```
returns
```
(const char *) "vector<int(const int*)>"
```
rather than either failing or returning the input as-is.

Note that the function parameter in the example must have the `const` keyword to trigger the problem.

```
root [6] TClass::GetClass("vector<int(Zyx*)>")->GetName()
...
(const char *) "vector<int(Zyx*)>"
```
Having the template parameter being a function (signature) is also essential, without it the result is correct.

```
root [7] TClass::GetClass("vector<const Zyx*)>")->GetName()
(const char *) "vector<const Zyx*)>"
```

Debugging:  `root.exe -b -l -e 'TClass::GetClass("vector<function<int(const Zyx*,const std::string&)> >")->GetName()' -q`
We can see that the trouble appears to be triggers inside `ROOT::TMetaUtils::TClingLookupHelper::GetPartiallyDesugaredNameWithScopeHandling `:

```
reakpoint 2, ROOT::TMetaUtils::TClingLookupHelper::GetPartiallyDesugaredNameWithScopeHandling (this=0x55555616cc10, tname=..., result=..., dropstd=true)
    at /home/pcanal/root_working/code/root/core/clingutils/src/TClingUtils.cxx:615
615     {
(gdb) p tname.c_str()
$33 = 0x555556b67030 "function<int(const Zyx*,const std::string&)>"
(gdb) bt
#0  ROOT::TMetaUtils::TClingLookupHelper::GetPartiallyDesugaredNameWithScopeHandling (this=0x55555616cc10, tname=..., result=..., dropstd=true)
    at /home/pcanal/root_working/code/root/core/clingutils/src/TClingUtils.cxx:615
#1  0x00007ffff7ce30ea in TClassEdit::TSplitType::ShortType (this=0x7fffffffa780, answ=..., mode=3618) at /home/pcanal/root_working/code/root/core/foundation/src/TClassEdit.cxx:437
#2  0x00007ffff7ce48e9 in TClassEdit::GetNormalizedName (norm_name=..., name=...) at /home/pcanal/root_working/code/root/core/foundation/src/TClassEdit.cxx:854
[...]

PR #18625, opened 2025-05-06
[core] Improve class name normalisation | fix crash in tutorial-analysis-dataframe-df014_CSVDataSource-py

hageboeck:
When looking up a TClass via name vs via typeid, the lookup proceeded
without default template arguments in the former, and with default
arguments in the latter case.
This lead to a situation where
`tclass->GetActualClass(std::map<x,y>*)`
could delete itself while performing the lookup, since a new instance
was created that replaced the existing instance.

In the df014 tutorial, cppyy was essentially calling
```c++
   auto first = TClass::GetClass("unordered_map<string,char>", true, true);
   std::unordered_map<std::string,char> map;
   auto second = first->GetActualClass(&map);
```
Before this PR, `first != second`, and `first` was being deleted while `first->GetActualClass` was still running.

Now, the class lookup in such a case is retried without the default STL arguments, which should lead to `first == second` in the above example. A test for this behaviour was added.

Thanks to @pcanal for advice.

### DIFF
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -33,40 +33,40 @@
 using std::string, std::string_view, std::vector, std::set;
 
 namespace {
-   static TClassEdit::TInterpreterLookupHelper *gInterpreterHelper = nullptr;
+static TClassEdit::TInterpreterLookupHelper *gInterpreterHelper = nullptr;
 
-   template <typename T>
-   struct ShuttingDownSignaler : public T {
-      using T::T;
+template <typename T>
+struct ShuttingDownSignaler : public T {
+   using T::T;
 
-      ShuttingDownSignaler() = default;
-      ShuttingDownSignaler(T &&in) : T(std::move(in)) {}
+   ShuttingDownSignaler() = default;
+   ShuttingDownSignaler(T &&in) : T(std::move(in)) {}
 
-      ~ShuttingDownSignaler()
-      {
-         if (gInterpreterHelper)
-            gInterpreterHelper->ShuttingDownSignal();
-      }
-   };
-}
+   ~ShuttingDownSignaler()
+   {
+      if (gInterpreterHelper)
+         gInterpreterHelper->ShuttingDownSignal();
+   }
+};
 
 ////////////////////////////////////////////////////////////////////////////////
 /// Return the length, if any, taken by std:: and any
 /// potential inline namespace (well compiler detail namespace).
 
-static size_t StdLen(const std::string_view name)
+size_t StdLen(const std::string_view name)
 {
    size_t len = 0;
-   if (name.compare(0,5,"std::")==0) {
+   if (name.compare(0, 5, "std::") == 0) {
       len = 5;
 
       // TODO: This is likely to induce unwanted autoparsing, those are reduced
       // by the caching of the result.
       if (gInterpreterHelper) {
-         for(size_t i = 5; i < name.length(); ++i) {
-            if (name[i] == '<') break;
+         for (size_t i = 5; i < name.length(); ++i) {
+            if (name[i] == '<')
+               break;
             if (name[i] == ':') {
-               std::string scope(name.data(),i);
+               std::string scope(name.data(), i);
 
                // We assume that we are called in already serialized code.
                // Note: should we also cache the negative answers?
@@ -81,21 +81,20 @@ static size_t StdLen(const std::string_view name)
 
                if (isInlined) {
                   len = i;
-                  if (i+1<name.length() && name[i+1]==':') {
+                  if (i + 1 < name.length() && name[i + 1] == ':') {
                      len += 2;
                   }
                } else {
                   std::string scoperesult;
-                  if (!gInterpreterHelper->ExistingTypeCheck(scope, scoperesult)
-                     && gInterpreterHelper->IsDeclaredScope(scope, isInlined))
-                  {
+                  if (!gInterpreterHelper->ExistingTypeCheck(scope, scoperesult) &&
+                      gInterpreterHelper->IsDeclaredScope(scope, isInlined)) {
                      if (isInlined) {
                         {
                            ROOT::Internal::TSpinLockGuard lock(spinFlag);
                            gInlined.insert(scope);
                         }
                         len = i;
-                        if (i+1<name.length() && name[i+1]==':') {
+                        if (i + 1 < name.length() && name[i + 1] == ':') {
                            len += 2;
                         }
                      }
@@ -113,11 +112,11 @@ static size_t StdLen(const std::string_view name)
 /// Remove std:: and any potential inline namespace (well compiler detail
 /// namespace.
 
-static void RemoveStd(std::string &name, size_t pos = 0)
+void RemoveStd(std::string &name, size_t pos = 0)
 {
-   size_t len = StdLen({name.data()+pos,name.length()-pos});
+   size_t len = StdLen({name.data() + pos, name.length() - pos});
    if (len) {
-      name.erase(pos,len);
+      name.erase(pos, len);
    }
 }
 
@@ -132,6 +131,8 @@ static void RemoveStd(std::string_view &name)
       name.remove_prefix(len);
    }
 }
+
+} // namespace
 
 ////////////////////////////////////////////////////////////////////////////////
 
@@ -1087,7 +1088,7 @@ int TClassEdit::GetSplit(const char *type, vector<string>& output, int &nestedLo
                offset += 5;
             }
             constexpr auto char_traits_s = "char_traits<char>";
-            // or 
+            // or
             // static constexpr char const* const char_traits_s = "char_traits<char>";
             static constexpr unsigned int char_traits_len = std::char_traits<char>::length(char_traits_s);
             if (full.compare(offset, char_traits_len, char_traits_s) == 0) {
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -50,6 +50,14 @@ struct ShuttingDownSignaler : public T {
 };
 
 ////////////////////////////////////////////////////////////////////////////////
+/// Remove the next spaces.
+void RemoveSpace(std::string_view &s)
+{
+   while (!s.empty() && s[0] == ' ')
+      s.remove_prefix(1);
+}
+
+////////////////////////////////////////////////////////////////////////////////
 /// Return the length, if any, taken by std:: and any
 /// potential inline namespace (well compiler detail namespace).
 
@@ -124,11 +132,21 @@ void RemoveStd(std::string &name, size_t pos = 0)
 /// Remove std:: and any potential inline namespace (well compiler detail
 /// namespace.
 
-static void RemoveStd(std::string_view &name)
+void RemoveStd(std::string_view &name)
 {
    size_t len = StdLen(name);
    if (len) {
       name.remove_prefix(len);
+   }
+}
+
+////////////////////////////////////////////////////////////////////////////////
+/// Remove instances of "::".
+
+void RemoveScopeResolution(std::string &name)
+{
+   if (name.length() > 2 && name[0] == ':' && name[1] == ':') {
+      name.erase(0, 2);
    }
 }
 
@@ -454,7 +472,10 @@ void TClassEdit::TSplitType::ShortType(std::string &answ, int mode)
          if (gInterpreterHelper &&
              (gInterpreterHelper->ExistingTypeCheck(fElements[i], typeresult)
               || gInterpreterHelper->GetPartiallyDesugaredNameWithScopeHandling(fElements[i], typeresult))) {
-            if (!typeresult.empty()) fElements[i] = typeresult;
+            if (!typeresult.empty() && typeresult != fElements[i]) {
+               // the interpreter helper keeps the default template arguments, so shorten again
+               fElements[i] = TClassEdit::ShortType(typeresult.c_str(), mode | TClassEdit::kKeepOuterConst);
+            }
          }
       }
    }
@@ -642,6 +663,7 @@ bool TClassEdit::IsDefAlloc(const char *allocname, const char *classname)
    }
    a.remove_prefix(alloclen);
 
+   RemoveSpace(a);
    RemoveStd(a);
 
    string_view k = classname;
@@ -665,7 +687,9 @@ bool TClassEdit::IsDefAlloc(const char *allocname, const char *classname)
       a.remove_prefix(k.length());
    }
 
-   if (a.compare(0,1,">")!=0 && a.compare(0,2," >")!=0) {
+   RemoveSpace(a);
+
+   if (a.compare(0, 1, ">") != 0) {
       return false;
    }
 
@@ -683,6 +707,7 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
    if (IsDefAlloc(allocname,keyclassname)) return true;
 
    string_view a( allocname );
+   RemoveSpace(a);
    RemoveStd(a);
 
    constexpr auto length = std::char_traits<char>::length;
@@ -692,6 +717,7 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
    }
    a.remove_prefix(alloclen);
 
+   RemoveSpace(a);
    RemoveStd(a);
 
    constexpr static int pairlen = length("pair<");
@@ -705,6 +731,7 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
       a.remove_prefix(constlen+1);
    }
 
+   RemoveSpace(a);
    RemoveStd(a);
 
    string_view k = keyclassname;
@@ -736,14 +763,12 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
       }
       a.remove_prefix(end);
    } else {
-      size_t end = k.length();
-      if ( (a[end-1] == '*') || a[end]==' ' ) {
-         size_t skipSpace = (a[end] == ' ');
-         if (a.compare(end+skipSpace,constlen,"const") == 0) {
-            end += constlen+skipSpace;
-         }
+      // Deal with a trailing const of the allocated type
+      a.remove_prefix(k.length());
+      RemoveSpace(a);
+      if (a.compare(0, constlen, "const") == 0) {
+         a.remove_prefix(constlen);
       }
-      a.remove_prefix(end);
    }
 
    if (a[0] != ',') {
@@ -773,7 +798,9 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
       a.remove_prefix(v.length());
    }
 
-   if (a.compare(0,1,">")!=0 && a.compare(0,2," >")!=0) {
+   RemoveSpace(a);
+
+   if (a.compare(0, 1, ">") != 0) {
       return false;
    }
 
@@ -899,9 +926,7 @@ void TClassEdit::GetNormalizedName(std::string &norm_name, std::string_view name
    // Depending on how the user typed their code, in particular typedef
    // declarations, we may end up with an explicit '::' being
    // part of the result string.  For consistency, we must remove it.
-   if (norm_name.length()>2 && norm_name[0]==':' && norm_name[1]==':') {
-      norm_name.erase(0,2);
-   }
+   RemoveScopeResolution(norm_name);
 
    if (gInterpreterHelper) {
       // See if the expanded name itself is a typedef.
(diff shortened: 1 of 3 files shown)


core/meta/src/TClass.cxx
```diff
@@ -3181,6 +3181,14 @@ TClass *TClass::GetClass(const char *name, Bool_t load, Bool_t silent, size_t hi
       if (normalizedName != name) {
          cl = (TClass*)gROOT->GetListOfClasses()->FindObject(normalizedName.c_str());
 
+         if (!cl) {
+            // The normalized name above might still contain the default template arguments
```
pcanal, line 3185: That part is odd.  Do we know why `TClassEdit::GetNormalizedName` does not strip the default parameter in this case.

ferdymercury: Side note, default-Parameter-related issues:
https://github.com/root-project/root/issues/7470
https://github.com/root-project/root/issues/15996
https://its.cern.ch/jira/browse/ROOT-5862

pcanal: #7470 Likely different as it is about the forward declaration written in the rootmap file
#15996 Is about the customization of which default argument to drop from the normalization name
#8470 Should not play a role here (it is about handling of unknown types)

[its.cern.ch/jira/browse/ROOT-10311](https://its.cern.ch/jira/browse/ROOT-10311)   is unrelated (it is about look for an non-existing type) and may have already been solved

[its.cern.ch/jira/browse/ROOT-5862](https://its.cern.ch/jira/browse/ROOT-5862)
#18401 
might or might not be related.

ferdymercury: Thanks for the insight!

Side note:
ROOT-10311 is still an issue in master.

TTree::TTree constructor function gets auto-stripped into TTree (it thinks first one might be a namespace) and then GetClass("TTree") returns a nullptr.

dpiparo: puzzling.
```
root [0] string s="unordered_map<string,char>"
(std::string &) "unordered_map<string,char>"
root [1] string n
(std::string &) ""
root [2]  TClassEdit::GetNormalizedName(n, s);
root [3] n
(std::string &) "unordered_map<string,char,hash<string>,equal_to<string>,allocator<pair<const string,char> > >"

Comment (dpiparo) [reply core/meta/src/TClass.cxx:3185]: Digging deeper. There is a difference in interpreted and compiled code. This snippet:
```
std::string s = "unordered_map<string,char>";
std::string n;
TClassEdit::GetNormalizedName(n, s);
std::cout << s << " --> " << n << std::endl;
```
if compiled in an executable prints `unordered_map<string,char> --> unordered_map<string,char>`.
If interpreted, e.g. at the prompt, returns `unordered_map<string,char> --> unordered_map<string,char,hash<string>,equal_to<string>,allocator<pair<const string,char> > >`

Comment (hageboeck) [reply core/meta/src/TClass.cxx:3185]: > Digging deeper. There is a difference in interpreted and compiled code. This snippet:
> 
> ```
> std::string s = "unordered_map<string,char>";
> std::string n;
> TClassEdit::GetNormalizedName(n, s);
> std::cout << s << " --> " << n << std::endl;
> ```
> 
> if compiled in an executable prints `unordered_map<string,char> --> unordered_map<string,char>`. If interpreted, e.g. at the prompt, prints `unordered_map<string,char> --> unordered_map<string,char,hash<string>,equal_to<string>,allocator<pair<const string,char> > >`

I see this, too, in a test, but the difference here is whether the test runs in isolation
```
./testClassEdit --gtest_filter=*TestName*
```
or whether it runs after the other tests:
```
./testClassEdit
```
I have a fix for when it runs in isolation, but I'm digging now for the case where other queries to the type system have happened before.

Comment (hageboeck) [reply core/meta/src/TClass.cxx:3185]: > That part is odd. Do we know why `TClassEdit::GetNormalizedName` does not strip the default parameter in this case.

@pcanal Now we know, indeed:
- The trailing ` const` in the allocator template was throwing off the type normalisation.
- And the `__cxx11::basic_string` was preventing the normalisation to `string`.

Those two are solved now, and I get a consistent TClass now in this test (part of the PR now):
```c++
auto first = TClass::GetClass("unordered_map<string,char>", true, true);
    std::unordered_map<std::string, char> map;
    auto second = first->GetActualClass(&map);
    EXPECT_EQ(first, second);
```

hageboeck: Now there seems to be only one remaining issue, and I'm trying to find out if we can live with it. It's about the stripping (or not) of the default template arguments:

> Digging deeper. There is a difference in interpreted and compiled code. This snippet:
> 
> ```
> std::string s = "unordered_map<string,char>";
> std::string n;
> TClassEdit::GetNormalizedName(n, s);
> std::cout << s << " --> " << n << std::endl;
> ```
> 
> if compiled in an executable prints `unordered_map<string,char> --> unordered_map<string,char>`. If interpreted, e.g. at the prompt, prints `unordered_map<string,char> --> unordered_map<string,char,hash<string>,equal_to<string>,allocator<pair<const string,char> > >`

The code responsible is this:
https://github.com/root-project/root/blob/b30a5f2c7ece0544e2d17e14dee1f3cf146c5c42/core/foundation/src/TClassEdit.cxx#L903-L911

Before these lines, and with the changes of the PR applied, the `unordered_map<very long list>` collapses to `unordered_map<string,char>`, as the documentation of the functions claims:
https://github.com/root-project/root/blob/b30a5f2c7ece0544e2d17e14dee1f3cf146c5c42/core/foundation/src/TClassEdit.cxx#L853-L861

However, when `gInterpreterHelper != nullptr`, the `norm_name` runs through TCling, and comes back as `unordered_map<string,char,hash<string>,equal_to<string>,allocator<pair<const string,char> > >`. This contradicts the docstring, but it explains the difference between interpreted and compiled code that Danilo observed.
This also explains why my test `./testClassEdit --gtest_filter=*TestName*` yields the short name, and `./testClassEdit` yields the long name with default arguments. Interestingly, the TClass lookup seems to work now in both cases, so the origin of the `df104_CSVDataSource.py` crash seems to be fixed nevertheless.

That being said, I would like to ask @pcanal and @dpiparo to weigh in what the normalised name should be (see next conversation).

core/foundation/test/testClassEdit.cxx
```diff
@@ -292,3 +292,20 @@ TEST(TClassEdit, GetNormalizedName)
...
+   std::string in{"std::unordered_map<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, char, std::hash<std::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::equal_to<std::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::basic_string<char, std::char_traits<char>, std::allocator<char> > const, char> > >"};
+   const auto target = "unordered_map<string,char>";
+
+   std::string out;
+   TClassEdit::GetNormalizedName(out, in);
+   EXPECT_STREQ(target, out.c_str());
+
+   TClassEdit::GetNormalizedName(out, in_cxx11);
+   EXPECT_STREQ(target, out.c_str());
+}
```
hageboeck, line 311: @pcanal @dpiparo here is the test for the normalised name, which passes when the interpreterHelper isn't instantiated yet, but fails when it is.

In the former case, we get
`unordered_map<string,char>`
and in the latter:
`unordered_map<string,char,hash<string>,equal_to<string>,allocator<pair<const string,char> > >`

What should be the correct result?

dpiparo: `unordered_map<string,char>` is the normalized name.

core/foundation/src/TClassEdit.cxx
```diff
@@ -48,6 +48,11 @@ namespace {
             gInterpreterHelper->ShuttingDownSignal();
       }
    };
+
+   void RemoveSpace(std::string_view & s)
+   {
+      while (!s.empty() && s[0] == ' ') s.remove_prefix(1);
```
ferdymercury, line 54: Potential optimization:
```suggestion
      s.remove_prefix(std::min(s.find_first_not_of(" "), s.length()));
```

core/foundation/src/TClassEdit.cxx
```diff
@@ -48,6 +48,11 @@ namespace {
             gInterpreterHelper->ShuttingDownSignal();
       }
    };
+
+   void RemoveSpace(std::string_view & s)
```
ferdymercury, line 52:
```suggestion
   void RemoveLeadingSpace(std::string_view & s)
```

core/foundation/src/TClassEdit.cxx
```diff
@@ -682,6 +691,7 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
    if (IsDefAlloc(allocname,keyclassname)) return true;
 
    string_view a( allocname );
+   RemoveSpace(a);
```
ferdymercury, line 694:
```suggestion
   RemoveLeadingSpaces(a);
```

core/foundation/src/TClassEdit.cxx
```diff
@@ -691,6 +701,7 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
    }
    a.remove_prefix(alloclen);
 
+   RemoveSpace(a);
```
ferdymercury, line 704:
```suggestion
   RemoveLeadingSpaces(a);
```

core/foundation/src/TClassEdit.cxx
```diff
@@ -704,6 +715,7 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
       a.remove_prefix(constlen+1);
    }
 
+   RemoveSpace(a);
```
ferdymercury, line 718:
```suggestion
   RemoveLeadingSpaces(a);
```

core/foundation/src/TClassEdit.cxx
```diff
@@ -735,14 +747,12 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
...
    } else {
-      size_t end = k.length();
-      if ( (a[end-1] == '*') || a[end]==' ' ) {
-         size_t skipSpace = (a[end] == ' ');
-         if (a.compare(end+skipSpace,constlen,"const") == 0) {
-            end += constlen+skipSpace;
-         }
+      // Deal with a trailing const of the allocated type
+      a.remove_prefix(k.length());
+      RemoveSpace(a);
```
ferdymercury, line 752:
```suggestion
      RemoveLeadingSpaces(a);
```

core/foundation/src/TClassEdit.cxx
```diff
@@ -772,7 +782,9 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
       a.remove_prefix(v.length());
    }
 
-   if (a.compare(0,1,">")!=0 && a.compare(0,2," >")!=0) {
+   RemoveSpace(a);
```
ferdymercury, line 785:
```suggestion
   RemoveLeadingSpaces(a);
```

core/foundation/src/TClassEdit.cxx
```diff
@@ -735,14 +747,12 @@ bool TClassEdit::IsDefAlloc(const char *allocname,
       }
       a.remove_prefix(end);
    } else {
-      size_t end = k.length();
-      if ( (a[end-1] == '*') || a[end]==' ' ) {
-         size_t skipSpace = (a[end] == ' ');
-         if (a.compare(end+skipSpace,constlen,"const") == 0) {
-            end += constlen+skipSpace;
-         }
+      // Deal with a trailing const of the allocated type
```
ferdymercury, line 750: If it's a "trailing" const, shouldn't it be remove_suffix rather than remove_prefix? (And then removeTrailingSpaces)

core/foundation/test/testClassEdit.cxx
pcanal: Right.  This does not return the normalized name but the type which indeed might/should still have the default template arguments.  If I remember correctly the are meant to be remove lexicographicaly in a higher level routine.

core/meta/src/TClass.cxx
pcanal:
```
std::string s = "unordered_map<string,char>";
std::string n;
TClassEdit::GetNormalizedName(n, s);
std::cout << s << " --> " << n << std::endl;
```
> if compiled in an executable prints 1unordered_map<string,char> --> unordered_map<string,char>`. 
> If interpreted, e.g. at the prompt, prints 1unordered_map<string,char> --> unordered_map<string,char,hash<string>,equal_to<string>,allocator<pair<const string,char> > >1

Since this is `GetNormalizedName` I would expect that the default template parameter to be dropped for `unordered_map`.

core/foundation/test/testClassEdit.cxx
pcanal: I.e `GetNormalizedType` can/should return with the default parameter
but in the end `GetNormalizedName` needs to drop them (for STL collections) both compiled and interpreted code.

hageboeck: > I.e `GetNormalizedType` can/should return with the default parameter
> but in the end `GetNormalizedName` needs to drop them (for STL collections) both compiled and interpreted code.

I will implement this now.

core/foundation/src/TClassEdit.cxx
hageboeck: Thank you, but since `find_first_of` most likely goes character-by-character, too, it's unlikely that it saves instructions. For fun, I ran the two through godbolt, and mine had 16 lines of assembly, yours has 22. This doesn't tell us which one is faster, but I will keep the function is it is now.

core/foundation/src/TClassEdit.cxx
hageboeck: I think that would be overdoing it slightly. If you look in the surroundings, there is already `RemoveStd`, and this function adds to the list of local helpers in the same spirit.
The reason it's in the .cxx in an anonymous namespace is to allow for inlining, but not affecting anything outside of this source file. It's local, and targeted to its job, but it shouldn't affect the rest of the world.

core/foundation/src/TClassEdit.cxx
hageboeck: No, the parsing logic compares allocator and allocated value by "eating up" common parts starting from the front. If the allocated value is empty up to `const>`, it's a default allocator. That's why we always `remove_prefix`.

ferdymercury: Thanks. Maybe the comment could be clarified as "Deal with an initially trailing (now leading)..."

core/meta/test/testTClass.cxx
```diff
@@ -62,4 +62,15 @@ TEST(TClass, TypeNameDouble)
...
+// https://github.com/root-project/root/issues/18654
+TEST(TClass, ConsistentSTLLookup)
+{
+   // The lookup via normalised shortened name yielded a different class
+   // than lookup via typeid. This lead to crashes in cppyy.
+   auto first = TClass::GetClass("unordered_map<string,char>", true, true);
+   std::unordered_map<std::string, char> map;
+   auto second = first->GetActualClass(&map);
+   EXPECT_EQ(first, second);
+}
```
ferdymercury, line 76:
```suggestion
}

```

core/foundation/src/TClassEdit.cxx
```diff
@@ -125,14 +132,26 @@ static void RemoveStd(std::string &name, size_t pos = 0)
...
    size_t len = StdLen(name);
    if (len) {
       name.remove_prefix(len);
    }
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// Remove instances of "::".
+
+void RemoveScopeResolutionOp(std::string &name)
```
pcanal, line 146: What is the semantic of the trailing `Op`?

core/foundation/src/TClassEdit.cxx
```diff
@@ -48,13 +48,20 @@ namespace {
...
+
+   ////////////////////////////////////////////////////////////////////////////////
+   /// Remove the next spaces.
+   void RemoveSpace(std::string_view &s)
+   {
+      while (!s.empty() && s[0] == ' ')
+         s.remove_prefix(1);
+   }
 
 ////////////////////////////////////////////////////////////////////////////////
```
pcanal, line 60: If those are brough into the (unnamed) namespace they probably should also follow the indentation of the other members (and/or vice et versa) for consistency.

core/foundation/src/TClassEdit.cxx
```diff
@@ -641,9 +660,11 @@ bool TClassEdit::IsDefAlloc(const char *allocname, const char *classname)
    }
    a.remove_prefix(alloclen);
 
+   RemoveSpace(a);
```
pcanal, line 663: Those may or may not be right but they don't _seem_ to have a description/raison-d'etre in the commit log/PR description.  How do they help?

pcanal:
The usage of the same code pattern (`ExistingTypeCheck/GetPartiallyDesugaredNameWithScopeHandling` around `TClassEdit.cxx:447` also need an update.  Likely just moving the call to `ShortType` after (rather than before) those calls.

core/foundation/src/TClassEdit.cxx
hageboeck: I was thinking about it, but ultimately, it created a lot of noise in the diffs. I wasn't sure if that diff noise wouldn't hide the real changes.

core/foundation/src/TClassEdit.cxx
hageboeck: They are supposed to make the function more robust. One of the errors I had to fix was `> const` vs `>const`, and that made me realise that a lot more spellings of the allocator that would be valid in C++ would fail the check in TClassEdit. For example, if an allocator was passed as
a) `std::allocator<std::pair<K,V>>`
b) `std::allocator<   std::pair<K,V>  >`
c) `std::allocator<std::  pair<K,V>>`,
d) `std::allocator<std::pair<K,V>  const  >`
they would all be valid default allocators, but TClassEdit would recognise only the first one. Admittedly, c) wouldn't be recognised with the current implementation either, but being able to b) and d) should improve things. Another advantage of the function is that it removes all spaces instead of only one.

Maybe I'm misunderstanding what you asked, and you only wanted this explanation to show up in the commit log?

pcanal: Both :).  This explanation makes sense and indeed should be part of the commit log.

pcanal:
> In contrast, GetNormalizedName was trying to strip the default template arguments, but not succeeding because they came back in the interpreter call:

Humm ... since the call to `ShortType` as a generic `mode` I assumed (I think correctly) that this mode could included the dropping of the defaults.

What we (re?) discovered is that `GetPartiallyDesugaredNameWithScopeHandling` (intentionally) doesn't produce a normalized name and the calling routine should handle it as such.  In this case, this means applying the options.

However, I think I was hasty in the solution because we probably actually want to be passing the more normalized name to `ExistingTypeCheck`.  So unfortunately, I think we have to use:
```
fElements[i] = TClassEdit::ShortType(fElements[i].c_str(),mode | TClassEdit::kKeepOuterConst); 
 if (mode&kResolveTypedef) { 
    // We 'just' need to check whether the outer type is a typedef or not; 
    // this also will add the default template parameter if any needs to 
    // be added. 
    string typeresult; 
    if (gInterpreterHelper && 
        (gInterpreterHelper->ExistingTypeCheck(fElements[i], typeresult) 
         || gInterpreterHelper->GetPartiallyDesugaredNameWithScopeHandling(fElements[i], typeresult))) { 
       if (!typeresult.empty() && typeresult != fElements[i]) {
          fElements[i] = TClassEdit::ShortType(typeresult, mode | TClassEdit::kKeepOuterConst); ;
       }
....
```

hageboeck:
> However, I think I was hasty in the solution because we probably actually want to be passing the more normalized name to `ExistingTypeCheck`. So unfortunately, I think we have to use:
> 
> ```
> fElements[i] = TClassEdit::ShortType(fElements[i].c_str(),mode | TClassEdit::kKeepOuterConst); 
>  if (mode&kResolveTypedef) { 
>     // We 'just' need to check whether the outer type is a typedef or not; 
>     // this also will add the default template parameter if any needs to 
>     // be added. 
>     string typeresult; 
>     if (gInterpreterHelper && 
>         (gInterpreterHelper->ExistingTypeCheck(fElements[i], typeresult) 
>          || gInterpreterHelper->GetPartiallyDesugaredNameWithScopeHandling(fElements[i], typeresult))) { 
>        if (!typeresult.empty() && typeresult != fElements[i]) {
>           fElements[i] = TClassEdit::ShortType(typeresult, mode | TClassEdit::kKeepOuterConst); ;
>        }
> ....
> ```

I agree, and I'm pushing this change soon.

core/foundation/src/TClassEdit.cxx
hageboeck: Added it now as an [NFC] commit

ferdymercury:
Loosely related: https://github.com/root-project/root/issues/10166#issue-1173670182

pcanal (approved):
Thank you very much.

core/foundation/src/TClassEdit.cxx
pcanal: As a side note, I suppose that `IsDefAlloc` was meant to (originally?) receive an input that was already normalized from the spacing point of view ...

PR #18629, opened 2025-05-07
[ntuple] Cast min/max to the actual type when quantizing/unquantizing in Real32Quant

silverweed:
See the commit message for more details.

This PR should be backported, probably even to 6.34 to make it work with the rntuple validation suite.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/src/RColumnElement.hxx
+++ b/tree/ntuple/src/RColumnElement.hxx
@@ -1028,6 +1028,11 @@ int QuantizeReals(Quantized_t *dst, const T *src, std::size_t count, double min,
    static_assert(sizeof(T) <= sizeof(double));
    assert(1 <= nQuantBits && nQuantBits <= 8 * sizeof(Quantized_t));
 
+   // `min` and `max` are supposed to be exactly representable by type `T` because we cast them to `T` in
+   // SetQuantized().
+   assert(min == static_cast<double>(static_cast<T>(min)));
+   assert(max == static_cast<double>(static_cast<T>(max)));
+
    const std::size_t quantMax = (1ull << nQuantBits) - 1;
    const double scale = quantMax / (max - min);
    const std::size_t unusedBits = sizeof(Quantized_t) * 8 - nQuantBits;
@@ -1067,8 +1072,7 @@ int UnquantizeReals(T *dst, const Quantized_t *src, std::size_t count, double mi
    const double scale = (max - min) / quantMax;
    const std::size_t unusedBits = sizeof(Quantized_t) * 8 - nQuantBits;
    const double eps = std::numeric_limits<double>::epsilon();
-   const double emin = min - std::abs(min) * eps;
-   const double emax = max + std::abs(max) * eps;
+   const double emax = max + std::max(1.0, std::abs(max)) * eps;
 
    int nOutOfRange = 0;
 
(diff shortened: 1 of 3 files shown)


hahnjo:
> probably even to 6.34 to make it work with the rntuple validation suite.

The goal is to use exactly 6.34.00, the first public version to produce the final RNTuple format. If that version has implementation bugs, I think I would rather skip `Real32Quant` for the first set of reference files.

silverweed:
Right, we cannot magically change 6.34.00...then I'd just backport it to 6.36.

tree/ntuple/src/RColumnElement.hxx
```diff
@@ -1057,18 +1059,20 @@ int QuantizeReals(Quantized_t *dst, const T *src, std::size_t count, double min,
...
 
+   const T min = minD;
+   const T max = maxD;
    const std::size_t quantMax = (1ull << nQuantBits) - 1;
    const double scale = (max - min) / quantMax;
    const std::size_t unusedBits = sizeof(Quantized_t) * 8 - nQuantBits;
-   const double eps = std::numeric_limits<double>::epsilon();
-   const double emin = min - std::abs(min) * eps;
-   const double emax = max + std::abs(max) * eps;
+   const T eps = std::numeric_limits<double>::epsilon();
```
pcanal, line 1073: Would this make sense?
```suggestion
   const T eps = std::numeric_limits<T>::epsilon();
```

silverweed:
As it's apparent from the unit test failures (which I'm not sure why I didn't see when I ran them locally the first time), either this change is incorrect or some of our tests are (in regards to the max error they allow in respect to the bit width).
I need to think more about this and what really is the correct solution to this conundrum.

tree/ntuple/src/RColumnElement.hxx
```diff
@@ -1080,10 +1083,20 @@ int UnquantizeReals(T *dst, const Quantized_t *src, std::size_t count, double mi
...
+      // by a quantized value of 0, meaning there is no floating point error accumulation coming from
+      // `e = fq * scale + min` (as it just becomes `e = min`).
+      // For the max value, however, some error is introduced by the operation, therefore we allow for
+      // some leeway in the out of range check.
+      nOutOfRange += !(min <= e && e <= emax);
 
-      nOutOfRange += !(emin <= dst[i] && dst[i] <= emax);
+      // Since we want to guarantee that the out value is always within `min` and `max`,
+      // after the bounds check we clamp the value to the stored `max`.
+      e = std::min(e, max);
```
pcanal, line 1097: Should there be an assert at the start of the function `assert(min <= max);` ?

tree/ntuple/src/RColumnElement.hxx
```diff
@@ -1028,6 +1028,10 @@ int QuantizeReals(Quantized_t *dst, const T *src, std::size_t count, double min,
    static_assert(sizeof(T) <= sizeof(double));
    assert(1 <= nQuantBits && nQuantBits <= 8 * sizeof(Quantized_t));
 
+   // `min` and `max` are supposed to be representable by type `T` because we cast them to `T` in SetQuantized().
```
pcanal, line 1031: Do you mean?:
```suggestion
   // `min` and `max` are supposed to be exactly representable by type `T` because we cast them to `T` in SetQuantized().
```
and/or
```suggestion
   // `min` and `max` are supposed to be perfectly representable by type `T` because we cast them to `T` in SetQuantized().
```

silverweed: I guess so, although it might be a bit misleading: we don't require the user to provide an exactly representable number, but we do cast it to the in-memory type of the field. Therefore we by definition end up with a number exactly representable by `T`.

hahnjo (approved):
LGTM, but a second approval would be good since I was involved in the offline discussions around this PR.

enirolf (approved):
LGTM!

PR #18630, opened 2025-05-07
[ntuple] Accept a single auxiliary processor in `RNTupleProcessor::CreateJoin`

enirolf:
Instead of having the `RNTupleJoinProcessor` keep track of multiple auxiliary processors per instance, it now only has one at a time. In case a multiway join is desired (i.e. with multiple auxiliary processors), the now-recommended way is to compose it using multiple `RNTupleJoinProcessor`s.

The main rationale behind this change is the fact that this removes some overhead when creating the join processors (i.e., no need anymore to first create a vector of `RNTupleOpenSpecs` or processors), and that we currently foresee the majority of joins to only involve one auxiliary RNTuple (or combination thereof) anyways.

As a semi-related change, name conflicts are now handled differently. Fields from the auxiliary processor get added to the model of the primary processor wrapped in an untyped collection with the name of the auxiliary processor. Therefore, instead of checking that the primary and auxiliary processors don't share the same name, we should check that the primary processor doesn't already have a field with the name of the auxiliary processor.

### DIFF
--- a/tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
@@ -299,50 +299,44 @@ public:
    ///
    /// \param[in] primaryNTuple The name and location of the primary RNTuple. Its entries are processed in sequential
    /// order.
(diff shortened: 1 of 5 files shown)


hahnjo (approved):
LGTM, thanks for the detailed explanations in the commit messages!

PR #18632, opened 2025-05-07
Move some functionality away from RLoopManager

vepadulano:
Some code is not necessary anymore since the functionalities have been moved to RTTreeDS.

### DIFF
--- a/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
@@ -1411,18 +1411,15 @@ public:
                                                  const RSnapshotOptions &options = RSnapshotOptions())
    {
       const auto definedColumns = fColRegister.GenerateColumnNames();
-      auto *tree = fLoopManager->GetTree();
 
-      const auto treeBranchNames = tree != nullptr ? ROOT::Internal::TreeUtils::GetTopLevelBranchNames(*tree) : ColumnNames_t{};
       const auto dsColumns = GetDataSource() ? ROOT::Internal::RDF::GetTopLevelFieldNames(*GetDataSource()) : ColumnNames_t{};
       // Ignore R_rdf_sizeof_* columns coming from datasources: we don't want to Snapshot those
       ColumnNames_t dsColumnsWithoutSizeColumns;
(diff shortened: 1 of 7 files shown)


tree/dataframe/src/RLoopManager.cxx
```diff
@@ -595,110 +560,6 @@ bool validTTreeReaderRead(TTreeReader &treeReader)
 }
 } // namespace
 
-/// Run event loop over one or multiple ROOT files, in parallel.
```
hageboeck, line 598: Ubuntu 22 is issuing a warning that the function above is unused.

hageboeck (approved):
LGTM, but an unused function warning should be fixed.

tree/dataframe/src/RLoopManager.cxx
vepadulano: Thanks! That wasn't caught on my machine

vepadulano: Also found out why, it's because the function is now only used in the MT case, and the method calling it is guarded by `#ifdef R__USE_IMT`, adding the guard also to this function.

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -1560,18 +1557,14 @@ public:
...
-      const auto treeBranchNames =
-         tree != nullptr ? ROOT::Internal::TreeUtils::GetTopLevelBranchNames(*tree) : ColumnNames_t{};
       const auto dsColumns = GetDataSource() ? GetDataSource()->GetColumnNames() : ColumnNames_t{};
       // Ignore R_rdf_sizeof_* columns coming from datasources: we don't want to Snapshot those
       ColumnNames_t dsColumnsWithoutSizeColumns;
       std::copy_if(dsColumns.begin(), dsColumns.end(), std::back_inserter(dsColumnsWithoutSizeColumns),
                    [](const std::string &name) { return name.size() < 13 || name.substr(0, 13) != "R_rdf_sizeof_"; });
       ColumnNames_t columnNames;
-      columnNames.reserve(definedColumns.size() + treeBranchNames.size() + dsColumns.size());
+      columnNames.reserve(definedColumns.size() + +dsColumns.size());
```
ferdymercury, line 1566:
```suggestion
      columnNames.reserve(definedColumns.size() + dsColumns.size());
```

PR #18633, opened 2025-05-07
[tmva][sofie] Implement support for dynamic tensor shapes 

lmoneta:
Add support for dynamic tensor shapes in several operators and introduce the managing of a new type of tensors (shape tensors) representing shapes which are the output of some operators (e.g. Shape) and input of others (e.g ConstantOfShape, Reshape, ...), see specific commit for more details. 

With this PR we can parse the Atlas GNN dev tracking model, specifying when creating the Session class the maximum number of nodes and edges, in order to allocate the tensors during Session initialization and then using at run time the given edges and nodes sizes.


tmva/sofie/inc/TMVA/ROperator_Concat.hxx
```diff
@@ -98,40 +101,75 @@
...
+               //    if (pos == concat_dim.length())
+               //       ret[fAxis] = Dim{i_concat_dim}; // dimension is integer
+               //    else {
+               //       // check if a composite expression
+               //       ret[fAxis] = Dim{concat_dim};
+               // }
+               // catch (std::invalid_argument const& ex) {
+
+               // }
+
```
sanjibansg: maybe we can remove the commented code

tmva/sofie/src/SOFIE_common.cxx
```diff
@@ -338,17 +359,24 @@ std::vector<size_t>  UTILITY::MultidirectionalBroadcastShape(std::vector<std::ve
    }
 }
 
-std::vector<size_t>  UTILITY::UnidirectionalBroadcastShape(std::vector<size_t> shapeA, std::vector<size_t> shapeB)
+// check multi-directional broadcasting of two shapes (need to pass inputs by non const ref. since we might prepends with one's
+// return a pair of integer flag and new broadcasted shape
+// if flag = 0: shape are identical
+//    flag = 1: return shape is equal to A, we broadcast B
+//    flag = 2: return shape is equal to B we broadcast A
+//    flag = 3: return shape is common of two we broadcast A and B to output
```
sanjibansg, line 367: probably we have a enum for this? this might be confusing or error-prone in future developments.

lmoneta: The flag is also used as a bit set , first bit indicates first input tensor is broadcasted, second bit that second input tensor is broadcasted. I would then leave it as an integer flag. This is however an internal API, not exposed to users

PR #18634, opened 2025-05-07
[sql] remove deprecated options and code

ferdymercury:
# This Pull request:

## Changes or fixes:

Removes deprecated build options `mysql odbc pgsql`. It also does some cleanup of leftovers from the already deleted `oracle`option.

@cern-sft-spi FYI

## Checklist:

- [ ] tested changes locally


linev (changes requested):
Plugin for TXNetFile can remain

Also all `tutorials/io/sql` can remain - they show usage of `TSQLServer` API 
and can be useful for understanding how it can be used. 
Even when examples refer to mysql or oracle DB.

andresailer:
The LCGCMake recipes have been updated. Thanks for the heads-up!

linev (approved):

guitargeek:
@ferdymercury, please let us know when this PR is ready from your side so it can be merged.

hahnjo:
Note that this merge was very early and does not match what I thought we agreed: We haven't even released 6.36.00 yet, so users cannot have hit the deprecation warning in a released version. Sure, if they complain now we can still restore it from `git`, but I'm not sure what we gain by immediately removing it...

devajithvs:
Someone asked about using `ODBC` : https://root-forum.cern.ch/t/odbc-and-mysql-connectivity-with-root-6-36-and-beyond/64065

PR #18639, opened 2025-05-07
Do not execute the test case that assumes geometry is there if there is no geometry

dpiparo:
Noticed while doing unrelated work.

### DIFF
--- a/cmake/modules/RootConfiguration.cmake
+++ b/cmake/modules/RootConfiguration.cmake
@@ -496,6 +496,12 @@ if (uring)
 else()
   set(hasuring undef)
 endif()
+if (geom)
+  set(hasgeom define)
+else()
+  set(hasgeom undef)
+endif()
+
 
 CHECK_CXX_SOURCE_COMPILES("
 inline __attribute__((always_inline)) bool TestBit(unsigned long f) { return f != 0; };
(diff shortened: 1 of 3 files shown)


ferdymercury:
Thanks!!

There is this other file using unguarded geom:

main/src/g2root.f

Maybe the easiest way could be to change:

`gSystem->Load("libGeom");`

with 

`R_ASSERT(gSystem->Load("libGeom") >= 0); // g2root needs ROOT built with geom active`

dpiparo:
Hi @ferdymercury : thanks a lot. I would maybe build the executable only if the geom is active, would that maybe be a solution too?

ferdymercury:
> Hi @ferdymercury : thanks a lot. I would maybe build the executable only if the geom is active, would that maybe be a solution too?

Thanks to you. Mmm not sure, according to https://linux.die.net/man/1/g2root
g2root is only creating a text C++ macro, not really running ROOT. So even if geom is not installed, g2root should be able to work without problems. Now if someone gets that macro, sends it over SSH to another computer, and that other computer has no geom, it should get the error. That's why I was suggesting to edit the created C++ test macro with a more sensible error message. (Otherwise you would get gGeoManager error not defined.)

ferdymercury:
Integrated in https://github.com/root-project/root/pull/18737, thanks!

PR #18640, opened 2025-05-07
[v636] Backport fixes for building with standalone Clang

guitargeek:
Backport the PRs that improve support for the case of building with `builtin_clang=OFF`:


hahnjo (changes requested):
For the same reason as two weeks ago, I'm against backporting build system changes!

hahnjo:
Quoting from https://github.com/root-project/root/pull/18386#discussion_r2042841433:
> As changes to the build system this late in the cycle can have a significant blast area, I maintain that this PR should be reverted.

That continues to hold.

from https://github.com/root-project/root/pull/18386#discussion_r2044577968:
> What I argue is that build system changes in general can have a significant blast radius. This PR contains changes on the main configuration code path in `core/clingutils/CMakeLists.txt`.

That is the case for my commit that you backport while I never mean this to go into `v6-36-00-patches`.

In https://github.com/root-project/root/pull/18460#issuecomment-2821066489 you said:
> Sure, this can also be done after the branching of 6.36!

and https://github.com/root-project/root/pull/18460#issuecomment-2830659313

> what do you think about this for `master`?

For me, this was also never meant to go into `v6-36-00-patches`.

I have pointed this out a number of times in the past: in my opinion, we backport too many commits. A branch is created and frozen so it can stabilize. That cannot happen if changes keep getting added at a higher than necessary pace.

Also pinging @dpiparo, I think we need a fundamental decision about our strategy here.

guitargeek:
Whatever, it's not worth it.

Issue #18624, opened 2025-05-06
[cling] UNREACHABLE executed when unloading deduction guide


hahnjo:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

When built with assertions, Cling crashes when trying to unload deduction guides.

### Reproducer

```
root [0] .rawInput
Using raw input
root [1] template <typename T> struct A { A(int) {} };
root [2] A(int) -> A<int>;
root [3] .rawInput
Not using raw input
root [4] .undo
Can't mangle a deduction guide name!
UNREACHABLE executed at /path/to/src/interpreter/llvm-project/clang/lib/AST/ItaniumMangle.cpp:1736!
Aborted (core dumped)
```

### ROOT version

`master`, also reproduced with `v6-32-00-patches`, and probably happens with all ROOT versions built with C++17

### Installation method

from source

### Operating system

Linux

### Additional context

Side remark: we may want to turn off `LLVM_UNREACHABLE_OPTIMIZE`, this would clearly error out for such cases in all configurations.

dpiparo:
A somewhat more visible crash with llvm20:
```
lxplus:~$ root
   ------------------------------------------------------------------
  | Welcome to ROOT 6.38.00                        https://root.cern |
  | (c) 1995-2025, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Dec 18 2025, 00:00:00                 |
  | From tags/6-38-00@6-38-00                                        |
  | With g++ (GCC) 11.5.0 20240719 (Red Hat 11.5.0-11) std201703     |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------

root [0]  .rawInput
Using raw input
root [1] template <typename T> struct A { A(int) {} };
root [2] A(int) -> A<int>;
root [3] .rawInput
Not using raw input
root [4] .undo

 *** Break *** illegal instruction


===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================
#0  0x00007f0d8dcd9fba in wait4 () from /lib64/libc.so.6
#1  0x00007f0d8dc4f483 in do_system () from /lib64/libc.so.6
#2  0x00007f0d8e6aace6 in TUnixSystem::StackTrace() () from /usr/lib64/root/libCore.so.6.38
[...]

PR #18642, opened 2025-05-07
[interpreter] Disable `LLVM_UNREACHABLE_OPTIMIZE`

hahnjo:
We don't want to "optimize `llvm_unreachable()` as undefined behavior" but trap if we hit this in production builds.

I think this would have found https://github.com/root-project/root/issues/18624 in CMSSW production environment.

### DIFF
--- a/interpreter/CMakeLists.txt
+++ b/interpreter/CMakeLists.txt
@@ -181,6 +181,11 @@ if(builtin_llvm)
     endif()
   endif()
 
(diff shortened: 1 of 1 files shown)


vgvassilev (approved):
Ideally, I'd think that in interactive systems we should rewire the unreachable and asserts into ROOT errors. 

LGTM!

devajithvs (approved):
LGTM

PR #18645, opened 2025-05-07
[tmva][sofie] Correctly add intermediate tensors for ROperator GRU

guitargeek:
Consistently using the `RModel::AddIntermediateTensor()` interface will
make it easier to restructure the generated code for AD support with
Clad.


lmoneta (approved):
LGTM!
Thank you Jonas for fixing this. Using intermediate tensors for the temporary tensors in GRU is better also for exploiting the memory optimization we are using for the intermediate tensors.

PR #18646, opened 2025-05-07
[tmva][sofie] Avoid `std::vector<bool>` for tensor types

guitargeek:
The `std::vector<bool>` is a special type that doesn't wrap contiguous memory. For compatibility with basic C interfaces and to be differentiable with Clad, it would be better to not use it.


lmoneta (approved):
LGTM!
If the tests pass is good to be merged!
Thank you Jonas for this change!

PR #18469, opened 2025-04-22
Testing code coverage workflow 

pcanal:
# This Pull request:

## Checklist:

- [ ] tested changes locally
- [ ] updated the docs (if necessary)

This PR fixes #

PR #18652, opened 2025-05-08
[ci] Try to fix retry logic

hahnjo:
When there is a problem during upload, the exception is correctly caught but then during handling, another exception occurs:
```
File "/__w/root/root/.github/workflows/root-ci-config/build_utils.py", line 261, in upload_file
  build_utils.print_warning(f"""Attempt {attempt} to upload {src_file} to {dest_object} failed. Retrying in {sleep_time} seconds...""")
NameError: name 'build_utils' is not defined
```

---

Note: I didn't test if this fully fixes the try logic...


pcanal:
This change has been tested as part of https://github.com/root-project/root/pull/18469

PR #18668, opened 2025-05-09
[ntuple] Fix handling of composed auxiliary processors

enirolf:
Instead of using a plain untyped record field for auxiliary processors, a lightweight field type specialization only meant for reading by the `RNTupleProcessor`, the `RAuxiliaryProcessorField`, is introduced. This way, we can recognise if the auxiliary processor to be added to the model contains nested auxiliary processors. When this is the case, their subfields are added as well.

### DIFF
--- a/tree/ntuple/src/RNTupleProcessor.cxx
+++ b/tree/ntuple/src/RNTupleProcessor.cxx
@@ -336,6 +336,28 @@ void ROOT::Experimental::RNTupleChainProcessor::AddEntriesToJoinTable(Internal::
 
 //------------------------------------------------------------------------------
 
+namespace ROOT::Experimental::Internal {
+class RAuxiliaryProcessorField final : public ROOT::RRecordField {
+private:
+   using RFieldBase::GenerateColumns;
+   void GenerateColumns() final
+   {
+      throw RException(R__FAIL("RAuxiliaryProcessorField fields must only be used for reading"));
+   }
+
+public:
+   RAuxiliaryProcessorField(std::string_view fieldName, std::vector<std::unique_ptr<RFieldBase>> itemFields)
+      : ROOT::RRecordField(fieldName, "RAuxiliaryProcessorField")
+   {
+      fOffsets.reserve(itemFields.size());
+      for (auto &item : itemFields) {
+         fOffsets.push_back(GetItemPadding(fSize, item->GetAlignment()));
+      }
+      AttachItemFields(std::move(itemFields));
+   }
+};
+} // namespace ROOT::Experimental::Internal
+
 ROOT::Experimental::RNTupleJoinProcessor::RNTupleJoinProcessor(std::unique_ptr<RNTupleProcessor> primaryProcessor,
                                                                std::unique_ptr<RNTupleProcessor> auxProcessor,
                                                                const std::vector<std::string> &joinFields,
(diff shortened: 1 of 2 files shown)


tree/ntuple/src/RNTupleProcessor.cxx
```diff
@@ -336,6 +336,28 @@ void ROOT::Experimental::RNTupleChainProcessor::AddEntriesToJoinTable(Internal::
...
+private:
+   void GenerateColumns() final { assert(false && "RAuxiliaryProcessorField fields must only be used for reading"); }
+
+public:
+   RAuxiliaryProcessorField(std::string_view fieldName, std::vector<std::unique_ptr<RFieldBase>> itemFields)
+      : ROOT::RRecordField(fieldName, "RAuxiliaryProcessorField")
+   {
+      fOffsets.reserve(itemFields.size());
+      for (auto &item : itemFields) {
+         // fSize += GetItemPadding(fSize, item->GetAlignment());
```
silverweed, line 350: leftover comment

tree/ntuple/src/RNTupleProcessor.cxx
```diff
@@ -336,6 +336,28 @@ void ROOT::Experimental::RNTupleChainProcessor::AddEntriesToJoinTable(Internal::
...
+         fOffsets.push_back(GetItemPadding(fSize, item->GetAlignment()));
+      }
+      AttachItemFields(std::move(itemFields));
+   }
+   RAuxiliaryProcessorField(RAuxiliaryProcessorField &&other) = default;
+   RAuxiliaryProcessorField &operator=(RAuxiliaryProcessorField &&other) = default;
+   ~RAuxiliaryProcessorField() final = default;
+};
```
silverweed, line 358: Aren't these implicitly defined by the compiler already?

tree/ntuple/src/RNTupleProcessor.cxx
```diff
@@ -336,6 +336,21 @@ void ROOT::Experimental::RNTupleChainProcessor::AddEntriesToJoinTable(Internal::
 
 //------------------------------------------------------------------------------
 
+namespace ROOT::Experimental::Internal {
+class RAuxiliaryProcessorField final : public ROOT::RRecordField {
```
hahnjo, line 340: As far as I understand, this field is only meant for reading. You can add
```c++
   void GenerateColumns() final { throw RException(R__FAIL("RAuxiliaryProcessorField must only be used for reading")); }
```
as an additional safeguard.

vepadulano (approved):
Nice improvement, thanks!

tree/ntuple/src/RNTupleProcessor.cxx
enirolf: I had this additionally, but the debian build didn't like it because of it being a hidden virtual overloaded function. The now-added `using` statement solves this though!

tree/ntuple/src/RNTupleProcessor.cxx
```diff
@@ -336,6 +336,25 @@ void ROOT::Experimental::RNTupleChainProcessor::AddEntriesToJoinTable(Internal::
...
+private:
+   using RFieldBase::GenerateColumns;
+   void GenerateColumns() final { assert(false && "RAuxiliaryProcessorField fields must only be used for reading"); }
```
hahnjo, line 343: If we then go with the argument of https://github.com/root-project/root/pull/18672#discussion_r2081624920, it should be `RException` as well ;-)

hahnjo (approved):
LGTM, with one comment

hahnjo (approved):
LGTM, thanks.

PR #18670, opened 2025-05-09
[df] Support DefinePerSample for RNTuple

### LINKED ISSUE #15086: DefinePerSample is not supported for RDataSource
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

[`RLoopManager::RunDataSource`](https://github.com/root-project/root/blob/727181ec7612c44aa3d6f04fcccfc5b23834fec2/tree/dataframe/src/RLoopManager.cxx#L614) is lacking the logic to update the sample information which is present in the equivalent runners for the empty data source or the [TTree data source](https://github.com/root-project/root/blob/727181ec7612c44aa3d6f04fcccfc5b23834fec2/tree/dataframe/src/RLoopManager.cxx#L581-L583). This notably prevents using `DefinePerSample` effectively when the input data source is an `RNTuple`.

```
$: ./repro.out
+-----+-----------+
| Row | file_name |
+-----+-----------+
| 0   | ""        |
+-----+-----------+
| 1   | ""        |
+-----+-----------+
| 2   | ""        |
+-----+-----------+
| 3   | ""        |
+-----+-----------+
| 4   | ""        |
+-----+-----------+
| 5   | ""        |
+-----+-----------+
| 6   | ""        |
+-----+-----------+
| 7   | ""        |
+-----+-----------+
| 8   | ""        |
+-----+-----------+
| 9   | ""        |
+-----+-----------+
| 10  | ""        |
+-----+-----------+
| 11  | ""        |
+-----+-----------+
| 12  | ""        |
+-----+-----------+
| 13  | ""        |
+-----+-----------+
| 14  | ""        |
+-----+-----------+
```

### Reproducer

```cpp
#include <ROOT/RNTupleDS.hxx>
#include <TFile.h>
#include <ROOT/RNTuple.hxx>
#include <ROOT/RVec.hxx>

#include <iostream>

using RNTupleModel = ROOT::Experimental::RNTupleModel;
using RNTupleWriter = ROOT::Experimental::RNTupleWriter;

void create_definepersample(const std::string &ntpl_name, const std::vector<std::string> &filenames)
{
    for (const auto &fn : filenames)
    {
        auto fldX = model->MakeField<ULong64_t>("x");
        auto ntpl = RNTupleWriter::Recreate(std::move(model), ntpl_name, fn);
        for (ULong64_t entry = 0; entry < 10; entry++)
        {
            *fldX = entry;
            ntpl->Fill();
        }
    }
}

int main()
{
    std::vector<std::string> filenames{
        "sample1.root",
        "sample2.root",
        "sample3.root"};
    std::string ntpl_name = "Events";
    create_definepersample(ntpl_name, filenames);
    ROOT::RDataFrame df{ntpl_name, filenames};
    auto df1 = df.DefinePerSample("file_name", [](unsigned int, const ROOT::RDF::RSampleInfo &si) {
      return si.AsString();
    });
    df1.Display<std::string>({"file_name"}, 30)->Print();
}
```

### ROOT version

Any

### Installation method

Any

### Operating system

Any

### Additional context

_No response_

vepadulano:
The information of each sample is populated at every new `GetEntryRanges` call. For the single thread case, every call corresponds to a new file. For the MT case, different ranges might be accessing the same file. Extra logic is needed to expose the storage location of the RNTuple of the current range, since this is not available directly from the RPageSource.

Fixes #15086

### DIFF
--- a/tree/dataframe/src/RNTupleDS.cxx
+++ b/tree/dataframe/src/RNTupleDS.cxx
@@ -13,6 +13,7 @@
 
 #include <ROOT/RDF/RColumnReaderBase.hxx>
 #include <ROOT/RDataFrame.hxx>
+#include <ROOT/RDF/Utils.hxx>
 #include <ROOT/RField.hxx>
 #include <ROOT/RFieldUtils.hxx>
 #include <ROOT/RPageStorageFile.hxx>
@@ -411,6 +412,7 @@ std::unique_ptr<ROOT::Internal::RPageSource> CreatePageSource(std::string_view n
 ROOT::RDF::RNTupleDS::RNTupleDS(std::string_view ntupleName, std::string_view fileName)
    : RNTupleDS(CreatePageSource(ntupleName, fileName))
 {
+   fFileNames = std::vector<std::string>{std::string{fileName}};
 }
 
 ROOT::RDF::RNTupleDS::RNTupleDS(RNTuple *ntuple) : RNTupleDS(ROOT::Internal::RPageSourceFile::CreateFromAnchor(*ntuple))
@@ -534,6 +536,7 @@ void ROOT::RDF::RNTupleDS::PrepareNextRanges()
             // to open and attach files here.
             range.fSource = CreatePageSource(fNTupleName, fFileNames[fNextFileIndex]);
          }
+         range.fFileName = fFileNames[fNextFileIndex];
          range.fSource->Attach();
          fNextFileIndex++;
 
@@ -553,6 +556,8 @@ void ROOT::RDF::RNTupleDS::PrepareNextRanges()
    unsigned int nSlotsPerFile = fNSlots / nRemainingFiles;
    for (std::size_t i = 0; (fNextRanges.size() < fNSlots) && (fNextFileIndex < nFiles); ++i) {
       std::unique_ptr<ROOT::Internal::RPageSource> source;
+      // Need to look for the file name to populate the sample info later
+      const auto &sourceFileName = fFileNames[fNextFileIndex];
       std::swap(fStagingArea[fNextFileIndex], source);
       if (!source) {
          // Empty files trigger this condition
@@ -595,6 +600,7 @@ void ROOT::RDF::RNTupleDS::PrepareNextRanges()
          auto end = rangesByCluster[iRange - 1].second;
 
          REntryRangeDS range;
+         range.fFileName = sourceFileName;
          // The last range for this file just takes the already opened page source. All previous ranges clone.
          if (iSlot == N - 1) {
             range.fSource = std::move(source);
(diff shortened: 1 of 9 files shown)


tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -788,3 +805,34 @@ ROOT::RDataFrame ROOT::RDF::FromRNTuple(std::string_view ntupleName, const std::
...
+
+   // TODO: There is no support for RNTuple in RDatasetSpec, thus the sample map
+   // is always empty at the moment.
+   if (sampleMap.empty())
+      return ROOT::RDF::RSampleInfo(
+         ntupleID, std::make_pair(fCurrentRanges[rangeIdx].fFirstEntry, fCurrentRanges[rangeIdx].fLastEntry));
+
+   if (sampleMap.find(ntupleID) == sampleMap.end())
+      throw std::runtime_error("Full sample identifier '" + ntupleID + "' cannot be found in the available samples.");
+
```
hageboeck, line 834: Could we offer a suggestion on how to avoid this error on the next run?

hageboeck: The word "Full" confuses me. It makes me think that there is also a partial sample identifier, so I would think that the solution might have to do with this.

hageboeck (approved):
LGTM, I have comments only regarding a missing include and maybe to make an error message more precise.

tree/dataframe/src/RNTupleDS.cxx
vepadulano: This is the same error as the TTree-based DefinePerSample, so I will leave any change for a follow-up PR. 

In regards to your confusion, I agree. The word "full" is meant to specify that this sample identifier is made of two parts: the storage location and the name of the dataset, e.g. `/data/myfile.root/mytree` (yes, the usage of `/` token is suboptimal, but not easily avoidable at this point). So far there hasn't been a real situation where the sample identifier of a TTree dataset was not found, except for the issue we had with the EOS xroot url redirect https://github.com/root-project/root/pull/17544, which was not a user error anyway.

So, if I had to think about something to suggest to the user in case of error, I would maybe mention they could double-check the input list of filenames/dataset names in the dataset specification.

tree/dataframe/test/dataframe_definepersample.cxx
```diff
@@ -170,20 +173,183 @@ TEST(DefinePerSample, TwoExecutions)
...
+         auto ntpl = ROOT::RNTupleWriter::Recreate(std::move(model), "ntuple", fn);
+         *fldX = i;
+         ntpl->Fill();
+      }
+   }
+
+   ~InputRNTuplesRAII()
+   {
+      for (auto i = 0u; i < fNFiles; ++i)
+         std::remove((fPrefix + std::to_string(i) + ".root").c_str());
```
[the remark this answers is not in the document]
vepadulano: Well it's in TClass.h, so it probably trickles down to everything. I agree we should not rely on it though, I'm adding it to the test directly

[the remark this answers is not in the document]
vepadulano: update: std::remove is actually declared in cstdio https://en.cppreference.com/w/cpp/io/c/remove

PR #18672, opened 2025-05-09
[ntuple] Improve error reporting consistency

hahnjo:
`RException` for field misuse, `R__ASSERT(false)` for unhandled column types.

### DIFF
--- a/tree/dataframe/src/RNTupleDS.cxx
+++ b/tree/dataframe/src/RNTupleDS.cxx
@@ -79,7 +79,7 @@ public:
       return representations;
    }
    // Field is only used for reading
-   void GenerateColumns() final { assert(false && "Cardinality fields must only be used for reading"); }
+   void GenerateColumns() final { throw RException(R__FAIL("Cardinality fields must only be used for reading")); }
    void GenerateColumns(const ROOT::RNTupleDescriptor &desc) final
    {
       GenerateColumnsImpl<ROOT::Internal::RColumnIndex>(desc);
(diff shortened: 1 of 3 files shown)


tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -79,7 +79,7 @@ class RRDFCardinalityField final : public ROOT::RFieldBase {
       return representations;
    }
    // Field is only used for reading
-   void GenerateColumns() final { assert(false && "Cardinality fields must only be used for reading"); }
+   void GenerateColumns() final { throw RException(R__FAIL("Cardinality fields must only be used for reading")); }
```
enirolf, line 82: Since this is an internal field and defined only in the implementation, I'm wondering whether in this case it actually makes more sense to keep it as an assertation :thinking:, same for `RArraySizeField`.

tree/ntuple/src/RColumnElement.cxx
```diff
@@ -61,7 +61,7 @@ std::pair<std::uint16_t, std::uint16_t> ROOT::Internal::RColumnElementBase::GetV
    default:
       if (type == kTestFutureColumnType)
          return std::make_pair(32, 32);
-      assert(false);
```
silverweed, line 64: I would not make these into `R__ASSERT`s: a crash here would mean that the developers (us) forgot to add this `case`, which is supposed to be "impossible" (it would be found during development / unit testing very quickly) and for such things I think an `assert` that disappears in release is a better fit.

tree/dataframe/src/RNTupleDS.cxx
silverweed: I tend to agree. Since technically these might be called by a user we might use `R__ASSERT` to make sure they hit it even in release builds (unless they disable assertions, in which case they would just get a no-op which is not catastrophic)

tree/ntuple/src/RColumnElement.cxx
hahnjo: Yes, but if this "impossible" condition is somehow met in production, we want to know instead of silently ignoring it. The column type also (indirectly) comes from user input, so I think being extra careful is useful. If the "impossible" never happens, then there should be no difference between the two.

tree/dataframe/src/RNTupleDS.cxx
enirolf: That's fair enough

enirolf (approved):
LGTM :)

silverweed (approved):

PR #18674, opened 2025-05-09
[meta] Properly handle _Atomic type specifier

### LINKED ISSUE #18643: [C++23][macOS] Failing test `roottest-root-io-tclass-make` with C++23
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Enabling C++23 resulted in the failure of the following test: https://github.com/root-project/root/actions/runs/14597245841/job/40946206187#step:6:14473

### Reproducer

A minimal reproduce with C++23 in ROOT - killed (SIGKILL):

```C++
void reproducer() {
   TClass::GetClass("TClass")->BuildRealData();
}
```

Running this on mac-beta outputs:
`zsh: killed     bin/root.exe reproducer.C`

There is no issue with C++20

### ROOT version

ROOT master

### Installation method

build from source

### Operating system

MacOS Beta XCode 16.4beta1 and 15.4+XCode 16.3

### Additional context

Upon preliminary investigation, it seems to be an issue with the data member in TClass - `std::atomic<TClass**> fPersistentRef`

dpiparo:
# This Pull request:

## Changes or fixes:
Updates the name normalisation routines in order to properly take into accoung the C11 atomic type specifier `_Atomic`. In absence of this treatment, ill formed type names end up in the lexer of Clang, causing an infinite recursion.

## Checklist:

- [v] tested changes locally
- [v] updated the docs (if necessary)

This PR fixes #18643

### DIFF
--- a/core/clingutils/src/TClingUtils.cxx
+++ b/core/clingutils/src/TClingUtils.cxx
@@ -4184,10 +4184,18 @@ void ROOT::TMetaUtils::GetNormalizedName(std::string &norm_name, const clang::Qu
    cling::Interpreter::PushTransactionRAII clingRAII(const_cast<cling::Interpreter*>(&interpreter));
    normalizedType.getAsStringInternal(normalizedNameStep1,policy);
 
+   // Remove the _Atomic type specifyier if present before normalising
+   TClassEdit::AtomicTypeNameHandlerRAII atomicTypeNameHandler_step1(
+      normalizedNameStep1, TClassEdit::AtomicTypeNameHandlerRAII::EBehavior::kDetectStrip);
+
    // Still remove the std:: and default template argument for STL container and
    // normalize the location and amount of white spaces.
    TClassEdit::TSplitType splitname(normalizedNameStep1.c_str(),(TClassEdit::EModType)(TClassEdit::kLong64 | TClassEdit::kDropStd | TClassEdit::kDropStlDefault | TClassEdit::kKeepOuterConst));
    splitname.ShortType(norm_name,TClassEdit::kDropStd | TClassEdit::kDropStlDefault );
+
+   TClassEdit::AtomicTypeNameHandlerRAII atomicTypeNameHandler_norm_name(
+      norm_name, atomicTypeNameHandler_step1.IsAtomic() ? TClassEdit::AtomicTypeNameHandlerRAII::EBehavior::kReadd
+                                                        : TClassEdit::AtomicTypeNameHandlerRAII::EBehavior::kNoOp);
 
    // The result of this routine is by definition a fully qualified name.  There is an implicit starting '::' at the beginning of the name.
    // Depending on how the user typed their code, in particular typedef declarations, we may end up with an explicit '::' being
--- a/core/foundation/inc/TClassEdit.h
+++ b/core/foundation/inc/TClassEdit.h
@@ -175,7 +175,8 @@ namespace TClassEdit {
       AtomicTypeNameHandlerRAII(std::string &typeName, EBehavior behavior = EBehavior::kDetectStripReadd)
          : fTypeName(typeName), fBehavior(behavior)
       {
-         if (fBehavior == EBehavior::kReadd || fBehavior == EBehavior::kNoOp) return;
+         if (fBehavior == EBehavior::kReadd || fBehavior == EBehavior::kNoOp)
+            return;
          if (0 == fTypeName.find(fgPrefix)) {
             fIsAtomic = true;
             if (fBehavior == EBehavior::kDetectStrip || fBehavior == EBehavior::kDetectStripReadd) {
--- a/core/foundation/inc/TClassEdit.h
+++ b/core/foundation/inc/TClassEdit.h
@@ -16,6 +16,7 @@
 #include <ROOT/RConfig.hxx>
 #include "RConfigure.h"
 #include <stdlib.h>
+#include <stdexcept>
 #ifdef R__WIN32
 #ifndef UNDNAME_COMPLETE
 #define UNDNAME_COMPLETE 0x0000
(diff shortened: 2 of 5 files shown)


core/foundation/inc/TClassEdit.h
```diff
@@ -153,6 +153,30 @@ namespace TClassEdit {
...
+   /// It expects no spaces at the beginning or end of the string
+   class AtomicTypeNameHandlerRAII {
+      std::string &fTypeName;
+      bool fIsAtomic = false;
+      const char *fgPrefix = "_Atomic(";
+
+   public:
+      AtomicTypeNameHandlerRAII(std::string &typeName) : fTypeName(typeName)
+      {
+         if (0 == fTypeName.find(fgPrefix)) {
```
pcanal, line 166: We should check and/or assert that `typeName.back() == ')'`

core/clingutils/src/TClingUtils.cxx
```diff
@@ -4184,6 +4184,8 @@ void ROOT::TMetaUtils::GetNormalizedName(std::string &norm_name, const clang::Qu
    cling::Interpreter::PushTransactionRAII clingRAII(const_cast<cling::Interpreter*>(&interpreter));
    normalizedType.getAsStringInternal(normalizedNameStep1,policy);
 
+   TClassEdit::AtomicTypeNameHandlerRAII atomicTypeNameHandler(normalizedNameStep1);
```
pcanal, line 4187: This may or may not be doing what is intended.  `normalizedNameStep1` is used only as input to the next step.  The output is `norm_name` which after this addition will **not** contain the `_Atomic(...)` part.

core/foundation/inc/TClassEdit.h
dpiparo: excellent idea.

core/clingutils/src/TClingUtils.cxx
dpiparo: Addressed.

core/foundation/inc/TClassEdit.h
```diff
@@ -153,6 +154,33 @@ namespace TClassEdit {
...
+            fIsAtomic = true;
+            fTypeName.erase(0, 8);
+            if (0 == fTypeName.size() || ')' != fTypeName[fTypeName.size() - 1]) {
+               throw std::runtime_error("Cannot remove substring \")\" at the end of typename \""+typeName+"\"");
```
pcanal, line 171: Note that the calling code is very unlikely to be exception safe.

core/clingutils/src/TClingUtils.cxx
```diff
@@ -4184,6 +4184,10 @@ void ROOT::TMetaUtils::GetNormalizedName(std::string &norm_name, const clang::Qu
    cling::Interpreter::PushTransactionRAII clingRAII(const_cast<cling::Interpreter*>(&interpreter));
    normalizedType.getAsStringInternal(normalizedNameStep1,policy);
 
+   norm_name = normalizedNameStep1;
+   TClassEdit::AtomicTypeNameHandlerRAII atomicTypeNameHandler_step1(normalizedNameStep1);
+   TClassEdit::AtomicTypeNameHandlerRAII atomicTypeNameHandler_norm_name(norm_name);
```
pcanal, line 4189: The first line is a fully copy of the (possibly long type).  We might want to extend AtomicTypeNameHandlerRAII so that we can write 'just':
```suggestion
   TClassEdit::AtomicTypeNameHandlerRAII atomicTypeNameHandler_step1(normalizedNameStep1);
   TClassEdit::AtomicTypeNameHandlerRAII atomicTypeNameHandler_norm_name(norm_name, atomicTypeNameHandler_step1.IsAtomic());
```
where the second one just record the reference and trust the requested state.

core/foundation/inc/TClassEdit.h
dpiparo: Yes. The exception is expressive enough to explain why ROOT probably crashed though perhaps?

core/clingutils/src/TClingUtils.cxx
dpiparo: Thanks. I do not fully understand why you mention a "copy". The RAAI stores a reference to the string (see https://github.com/root-project/root/pull/18674/files#diff-bc1b5fbce8a4116a51b5d348e8510df04cc97eac8a74e75889bf86e561c59212R160 ) but maybe you are referring to something else - sorry to be thick.

pcanal: line 4187 makes a copy of the intermediary string.

core/foundation/inc/TClassEdit.h
pcanal: It is probably fine in practice but still iffy.  i.e. it is unknown what state catching this exception would leave ROOT in; we should assume 'unusable' and if we do, then this might as well be a call to `Fatal`.

core/clingutils/src/TClingUtils.cxx
dpiparo: but isn't the suggested change require a copy anyway?

dpiparo: ah maybe you mean checking if norm_name uses "_Atomic" as a type specifier, and if yes go through the expensive path and if not do as we do now, right?

core/foundation/inc/TClassEdit.h
```diff
@@ -153,6 +154,36 @@ namespace TClassEdit {
...
+            fIsAtomic = isAtomic;
+         } else if (0 == fTypeName.find(fgPrefix)) {
+            fIsAtomic = true;
+            fTypeName.erase(0, 8);
+            if (0 == fTypeName.size() || ')' != fTypeName[fTypeName.size() - 1]) {
+               throw std::runtime_error("Cannot remove substring \")\" at the end of typename \"" + typeName + "\"");
+            }
+            fTypeName.pop_back();
+         }
+      }
```
pcanal, line 177:
```suggestion
      AtomicTypeNameHandlerRAII(std::string &typeName, bool isAtomic) : fTypeName(typeName), fIsAtomic(isAtomic) {}

      AtomicTypeNameHandlerRAII(std::string &typeName) : fTypeName(typeName)
      {
         if (0 == fTypeName.find(fgPrefix)) {
            fIsAtomic = true;
            fTypeName.erase(0, 8);
            if (0 == fTypeName.size() || ')' != fTypeName[fTypeName.size() - 1]) {
               throw std::runtime_error("Cannot remove substring \")\" at the end of typename \"" + typeName + "\"");
            }
            fTypeName.pop_back();
         }
      }
```
seems clearer.

dpiparo:
Thanks for the suggestion. Now it's part of the changes.

core/foundation/inc/TClassEdit.h
```diff
@@ -153,6 +154,50 @@ namespace TClassEdit {
...
+      std::string &fTypeName;
+      bool fIsAtomic = false;
+      EBehavior fBehavior;
+      const char *fgPrefix = "_Atomic(";
+
+   public:
+      AtomicTypeNameHandlerRAII(std::string &typeName, EBehavior behavior = EBehavior::kDetectStripReadd)
+         : fTypeName(typeName), fBehavior(behavior)
+      {
+         if (fBehavior == EBehavior::kReadd || fBehavior == EBehavior::kNoOp) return;
```
pcanal, line 178:
```suggestion
         if (fBehavior == EBehavior::kReadd || fBehavior == EBehavior::kNoOp)
            return;
```

pcanal (approved):
LGTM.  Thanks.

PR #18675, opened 2025-05-09
[ntuple] Remove unused private `REntry` methods

enirolf:
Both `UpdateValue` and `GetValue` were originally added to be used by the `RNTupleProcessor`, but are not used there anymore. Since they are private methods and have no uses anywhere else in the code base, there's no reason to keep them around.

### DIFF
--- a/tree/ntuple/inc/ROOT/REntry.hxx
+++ b/tree/ntuple/inc/ROOT/REntry.hxx
@@ -94,10 +94,6 @@ private:
       return value.template GetPtr<T>();
    }
 
-   /// Return the RValue currently bound to the provided field.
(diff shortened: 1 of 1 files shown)


pcanal (approved):
LTGM

vepadulano (approved):

hahnjo (approved):
LGTM. I think we can also trim the list of `friend`s to only `RNTupleSingleProcessor` which is now the only class that calls `REntry::Read` (can be in a follow-up PR).

enirolf:
> LGTM. I think we can also trim the list of `friend`s to only `RNTupleSingleProcessor` which is now the only class that calls `REntry::Read` (can be in a follow-up PR).

Already have this ready in a related PR coming up soon :+1:.

Issue #18667, opened 2025-05-09
Add setting to disable file writing in MakeModelAndMeasurementsFast


Superharz:
### Explain what you would like to see improved and how.

Add an option to [`MakeModelAndMeasurementsFast`](https://github.com/root-project/root/blob/ccb4914297de3110d0e4e47a0ef50526a64ced55/roofit/histfactory/src/MakeModelAndMeasurementsFast.cxx#L83) to not create any files.

Right now, there is a setting to disable writing the per-channel workspaces to disk.

A similar setting would be convenient to add for the combined workspace. Basically having `MakeModelAndMeasurementsFast` only return the combined workspace but not writing anything to disk.

### ROOT version

ROOT 6.37.01 from heads/master@v6-37-01-6587-gc7b62f1e9c

### Installation method

Build from source

### Operating system

AlmaLinux 9

PR #18676, opened 2025-05-09
[RF][HF] Add configuration parameter to disable workspace writing

guitargeek:
It should be possible to disable writing the combined RooWorkspace to
disk, because often the workspace is created and used in the same
script. In that case, writing the workspace to a file is unnecessary
overhead.

Tested by running the `hf001` tutorial with this flag, because the
tutorial is one of these cases where the workspace doesn't need to be
written to a file.

Closes https://github.com/root-project/root/issues/18667.

Also, ensure that the logging is correctly reset and that no output directory is created if not needed (there are separate commits for that).

### DIFF
--- a/roofit/histfactory/inc/RooStats/HistFactory/HistoToWorkspaceFactoryFast.h
+++ b/roofit/histfactory/inc/RooStats/HistFactory/HistoToWorkspaceFactoryFast.h
@@ -11,45 +11,49 @@
 #ifndef ROOSTATS_HISTOTOWORKSPACEFACTORYFAST
 #define ROOSTATS_HISTOTOWORKSPACEFACTORYFAST
 
+#include <RooStats/HistFactory/Systematics.h>
+
+#include <RooArgSet.h>
+#include <RooDataSet.h>
+#include <RooFitResult.h>
+#include <RooPlot.h>
+#include <RooWorkspace.h>
+
+#include <TH1.h>
+#include <TObject.h>
+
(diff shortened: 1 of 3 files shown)


vgvassilev:
Any chance for a test?

roofit/histfactory/inc/RooStats/HistFactory/HistoToWorkspaceFactoryFast.h
```diff
@@ -11,45 +11,49 @@
...
+class Measurement;
+class Channel;
+class Sample;
+
+class HistoToWorkspaceFactoryFast : public TObject {
+
+public:
+   /// \brief Configuration settings for HistFactory behavior.
+   struct Configuration {
+      /// \brief Enable or disable optimization for binned likelihood fits (default `true`).
```
hageboeck, line 48: By the way, if you want to save a few characters per docstring, our doxygen is set to take the first sentence up to the `.` as `\brief` if you don't specify any (`AUTOBRIEF`).

hageboeck (approved):
LGTM

roofit/histfactory/inc/RooStats/HistFactory/HistoToWorkspaceFactoryFast.h
guitargeek: Thanks! I'll keep this in mind for next time

Issue #9022, opened 2021-09-22
hadd with -n option misses some content


will-cern:
I haven't had a chance to look into the hadd code (TFileMerger) but in 6.24/04 I the following problem. Create two files like this:

```
{
  TFile f1("f1.root","RECREATE");
  TH1D h("h1","h1",10,0,10);
  h.Write();
  f1.Close();
  TFile f2("f2.root","RECREATE");
  TH1D h2("h2","h2",10,0,10);
  h2.Write();
  f2.Close();
}
```
now if we hadd these as normal (`hadd merge.root f1.root f2.root`) we get both histograms:

```
TFile**         merge.root      
 TFile*         merge.root      
  KEY: TH1D     h1;1    h1
  KEY: TH1D     h2;1    h2
```
but do this with the `-n 1` option I only get h1:

```
TFile**         merge.root      
 TFile*         merge.root      
  KEY: TH1D     h1;1    h1
```

jblomer:
Additional observation: `-n1` is different from `-n 1`. I can reproduce the problem with `-n 1` on master.

jblomer:
Option parsing in `hadd` demands a space between `-n` and the value. The value is passed on to `TFileMerger::SetMaxOpenedFiles()` which in turn sets it to a minimum value of `fMaxOpenedFiles = 2`.

From my tests, it appears that `TFileMerger` merges the first `fMaxOpenedFiles - 1` input files, which is probably not the desired behavior.

jblomer:
I'm reassigning to Philippe. I think it boils down to this: when processing the excess files, `TFileMerger` switches to the incremental mode. In this mode, `TFileMerger::MergeOne` sets `canBeFound = true`, which in turn at the end of `MergeOne` skips the if block commented with `// Don't write the partial result for TTree and TH1`. Perhaps `canBeFound` should not be true for a histogram that is not (yet) found in the target file?

pcanal:
@silverweed Can you check this issue?

PR #18679, opened 2025-05-09
[io] don't miss writing a histogram that is only in a last file with option -n 2

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/9022

Explores solution suggested by @jblomer

fyi @will-cern

## Checklist:

- [x] tested changes locally


io/io/test/TFileMergerTests.cxx
```diff
@@ -219,3 +219,41 @@ TEST(TFileMerger, ChangeFile)
...
+   }
+   {
+      TFileMerger filemerger{false, false};
+      filemerger.SetMaxOpenedFiles(2);
+      filemerger.OutputFile(std::unique_ptr<TFile>{TFile::Open(outname, "RECREATE")});
+
+      filemerger.AddFile(filename1);
+      filemerger.AddFile(filename2);
+
+      // Before the fix, TTree::ChangeFile was called during Merge
```
silverweed, line 247: This comment only makes sense in the context of this PR: once we merge it it won't have the context anymore and readers will be left to wonder "which fix?"
Therefore I'd either reference the commit hash or remove the comment

silverweed (commented):
Thank you for tackling this long-standing issue!
I left a small comment.

io/io/test/TFileMergerTests.cxx
ferdymercury: That was a typo, sorry, removed now :) thanks for noticing.

silverweed (approved):
LGTM but I'd wait for @pcanal 's approval as well

io/io/src/TFileMerger.cxx
```diff
@@ -542,7 +542,8 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type,
             keyname, keytitle);
       return kTRUE;
    }
-   Bool_t canBeFound = (type & kIncremental) && (current_sourcedir->GetList()->FindObject(keyname) != nullptr);
+   Bool_t canBeFound = (type & kIncremental) && (current_sourcedir->GetList()->FindObject(keyname) != nullptr) &&
```
pcanal, line 545: Why is it `&&` and not `||` ?

pcanal: Or more exactly, I don't understand yet why the fact the histogram can be found means that it is not written in the end ...

ferdymercury: Check out:

https://github.com/ferdymercury/root/blob/1531153ea11a7b54f4eb3c170bbd28e9bc46447f/io/io/src/TFileMerger.cxx#L808-L817

so if canBeFound is true, there is an optimization that spare some write cycles.

We use && to avoid it being true, ie to force writing to file. Using || would go in the different direction.

Do you want me to rename canBeFound to skipPartialWriting ?

pcanal: > Do you want me to rename canBeFound to skipPartialWriting ?

Not yet. I am still confused.

The original `canBeFound` meant (in the context of incremental merge) 'histogram can be found in the source' while the new version is `histogram can be found in the source and in the target`.

The optimization is indeed 'skip partial writing if we can found the histogram again'.

So I don't understand (yet) the semantic of the change.  ie. Why is the new criteria the right choice?  Is the new criteria instead 'just' making `canBeFound` always false?

Another avenue of inquiry is 'the original code assume that if `canBeFound` is true then there will be another change to write the histogram.  Why is it no true anymore (it does not seem to be realted to 'can not be found in target')?  Is there other variation of the example that also fails (how does the `-n X` value relates to the number of files in the input list and how many are 'missing' the histograms).

Related: is it possible that the alternative if that at the refresh boundary there needs to be a flush/write as if we were at the end?

ferdymercury: Thanks for the review. I am not sure about these questions; I just followed jblomer's suggestion. My (limited) understanding is that this change just forces an extra partial write the first time that a new histogram appear in any file. So it does not really harm, but is suboptimal since, if all files have exactly the same histograms, then we could have waited until the end. But it makes it work if there are some files with and some without, independently of the chosen N.

pcanal: >  this change just forces an extra partial write the first time that a new histogram appear in any file.

That would be fair enough.  Can we verify that it just one and not number_of_input files?  (Related but probably unavoidable is that for the case of just 2 files with all the same histograms the number of write is doubled .... actually the number of double for each histogram that is in more than one file (i.e. usual case)).

io/io/src/TFileMerger.cxx
```diff
@@ -806,7 +807,8 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type,
          delete ndir;
       }
    } else if (!canBeFound) { // Don't write the partial result for TTree and TH1
-
+      if (gDebug > 1)
```
ferdymercury, line 810:
```suggestion
      if (gDebug > 0)
```

io/io/src/TFileMerger.cxx
pcanal: My guess is that with:
```
   {
      TFile f1(filename1, "RECREATE");
      TH1F h("h1", "h1", 1, 0, 1);
      h.Write();
      TH1F h("h3", "h3", 1, 0, 1);
      h3.Write();
      f1.Close();
      TFile f2(filename2, "RECREATE");
      TH1F h2("h2", "h2", 1, 0, 1);
      h2.Write();
      TH1F h("h3", "h3", 1, 0, 1);
      h3.Write();
      f2.Close();
   }
```
we ended up also with:
```
Info in TFileMerger::MergeOne: Writing partial result of h3 into target
```
which would/should indicate a 'spurrious' write for h3 (i.e. the normal case of a histogram in all files).

ferdymercury: I don't follow, sorry. If h3 is in all files, when merging file2, target->FindObject will evaluate to true, so canBeFound will be true, thus no partial result will be written? Only file1 will trigger partial result write?

pcanal: Yes but when going through the first file, it would not be (yet) in both the input and output, so it would be (spuriously) written, or am I still missing something?

ferdymercury: Hmm well, this is the result I get in debug mode:

```
Info in TFileMerger::MergeOne: Writing partial result of h1 into target
Info in TFileMerger::MergeOne: Writing partial result of h3 into target
Info in TFileMerger::MergeOne: Writing partial result of h2 into target
```

So... are you referring to the second line `Writing partial result of h3 into target` when you say spurious write?

pcanal: I think so.

pcanal: I think so.  Since h3 is in every file, shouldn't it not need any partial write?

ferdymercury: Behaviour prior to this PR:

```
Info in <TFileMerger::MergeOne>: Writing partial result of h1 into target
Info in <TFileMerger::MergeOne>: Writing partial result of h3 into target
```
This happens because for the first file, (type & kIncremental) is false.

Behavior after this PR:

```
Info in <TFileMerger::MergeOne>: Writing partial result of h1 into target
Info in <TFileMerger::MergeOne>: Writing partial result of h3 into target
Info in <TFileMerger::MergeOne>: Writing partial result of h2 into target
```

So, this PR is only improving things, right?

If I understand you correctly, on top of this improvement, you would like to additionally see h3 not partially written at all:

```
Info in <TFileMerger::MergeOne>: Writing partial result of h1 into target
Info in <TFileMerger::MergeOne>: Writing partial result of h2 into target
```

to be even more performant, but what you suggest would require to traverse all the files in advance to know which key should be not partially written at all if it's in all files. I think that would be less efficient than one extra partial write in the first occurrence of a key throughout the list of files; right?

ferdymercury: Ok, let's do a more elaborate example:

```
auto filename0 = "f0_9022.root";
auto filename1 = "f1_9022.root";
auto filename2 = "f2_9022.root";
auto outname = "file9022mergeroutput.root";
TFile f0(filename0, "RECREATE");
TH1F h0("h0", "h0", 1, 0, 1);
h0.Write();
f0.Close();
TFile f1(filename1, "RECREATE");
TH1F h("h1", "h1", 1, 0, 1);
h.Write();
TH1F h3("h3", "h3", 1, 0, 1);
h3.Write();
f1.Close();
TFile f2(filename2, "RECREATE");
TH1F h2("h2", "h2", 1, 0, 1);
h2.Write();
h3.Write();
f2.Close();
TFileMerger filemerger{false, false};
filemerger.SetMaxOpenedFiles(2);
filemerger.OutputFile(std::unique_ptr<TFile>{TFile::Open(outname, "RECREATE")});
filemerger.AddFile(filename0);
filemerger.AddFile(filename1);
filemerger.AddFile(filename2);
gDebug = 1;
filemerger.Merge();
```

So before the PR we get:
```
Info in <TFileMerger::MergeOne>: Writing partial result of h0 into target
```
(because for first type, type&incremental is false).

And the output file only contains histogram h0.

After the PR:
```
Info in <TFileMerger::MergeOne>: Writing partial result of h0 into target
Info in <TFileMerger::MergeOne>: Writing partial result of h1 into target
Info in <TFileMerger::MergeOne>: Writing partial result of h3 into target
Info in <TH1Merger::ExamineHistogram>: Examine histogram h3 - labels 1 - same limits 1 - axis found 1
Info in <TH1Merger::ExamineHistogram>: Examine histogram h3 - labels 1 - same limits 1 - axis found 1
Info in <TH1Merger::LabelMerge>: Merging histogram h3 into h3
Info in <TFileMerger::MergeOne>: Writing partial result of h2 into target
```

And the output file contains all histograms.

So the question is: why do you think it is a problem to have a first write whenever it appears? It's the only way to make it work so that the next appearance can call Merge. It's true that with a higher value of SetMaxOpenedFiles and with all files containing exactly the same hists, it's less optimum with this change, but still, this version is safer I guess?...

io/io/test/TFileMergerTests.cxx
```diff
@@ -219,3 +219,39 @@ TEST(TFileMerger, ChangeFile)
...
+   }
+   {
+      TFile file(outname, "READ");
+      EXPECT_NE(file.Get<TH1>("h1"), nullptr);
+      EXPECT_NE(file.Get<TH1>("h2"), nullptr);
+   }
+   gSystem->Unlink(filename1);
+   gSystem->Unlink(filename2);
+   gSystem->Unlink(outname);
+}
```
hageboeck, line 321: Hello @ferdymercury,

you seem to be working on a similar problem (I have some changes in #19331), so maybe you could cherry-pick the following test to see if your PR fixes it:
https://github.com/root-project/root/commit/46148c75da52151861e92e17325113cb81bc6787
[branch](https://github.com/hageboeck/root/tree/TFileMerger_mergeSelective)

ferdymercury: Done :)

ferdymercury:
@hageboeck your test looks green now ;)

io/io/test/TFileMergerTests.cxx
```diff
@@ -221,3 +225,119 @@ TEST(TFileMerger, ChangeFile)
...
+   const auto file0 = baseDir / "tomerge00.root";
+   const auto file1 = baseDir / "tomerge01.root";
+   try {
+      #ifdef WIN32
+      copy(baseDir / "hsimple.root", file0);
+      copy(baseDir / "hsimple.root", file1);
+      #else
+      create_symlink(baseDir / "hsimple.root", file0);
+      create_symlink(baseDir / "hsimple.root", file1);
+      #endif
```
hageboeck: Given that both files go anyway into the cleanup set, I guess that best would be to always copy. Less code, same functionality:
```suggestion
      copy(baseDir / "hsimple.root", file0);
      copy(baseDir / "hsimple.root", file1);
```

io/io/src/TFileMerger.cxx
pcanal: After investigating further, I think the 'right' solution is to only test on the target:
```suggestion
   Bool_t canBeFound = (type & kIncremental) &&
```
with the current state of the code , the main purpose (left ; (in previous incarnation it has a wider role) for 'canBeFound' is to detect that the object is an object owned by the file (`Hist` and `TTree`) and does not need to be written to the output and for `Hist` and `TTree` the 'write' to the output is associated with making it known to the output (have its own copy of the object).  As such it actually the target we need to look at and not the source.  (also the `type & kIncremental` makes it that it was skipping the writing only for either (the first if in append mode) or (the second and subsequent file if in regular mode).

io/io/src/TFileMerger.cxx
```diff
@@ -542,7 +542,8 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type,
             keyname, keytitle);
       return kTRUE;
    }
-   Bool_t canBeFound = (type & kIncremental) && (current_sourcedir->GetList()->FindObject(keyname) != nullptr);
+   Bool_t canBeFound = (type & kIncremental) && (current_sourcedir->GetList()->FindObject(keyname) != nullptr) &&
+                       (target->GetList()->FindObject(keyname) != nullptr);
```
pcanal:
```suggestion
   Bool_t canBeFound = (type & kIncremental) && (target->GetList()->FindObject(keyname) != nullptr);
```
See https://github.com/root-project/root/pull/18679/files#r2243572945

io/io/src/TFileMerger.cxx
ferdymercury: Thanks, it looks good to me.

Just to complete the discussion:

originally, there were two booleans, canBeFound and writeAnyway, see https://github.com/root-project/root/commit/c4d94efa7a3e44c25f21369807eb2ee59bf43f07#diff-765b2bfb5aed033baddbf7036ddda5a103b3b82fa5456b462b74b31cd79dc51b

and canBeFound was for the source-list and writeAnyway was for the target. The latter had the comment:
```
               // For now the code require a key (that might be ignored see search through current_sourcedir->GetList())
               // in the output file in order for it to be even considered.
```

and there were two separate if blocks:

```
 else if (!canBeFound) { // Don't write the partial result for TTree and TH1
               // Don't overwrite, if the object were not merged.
               // NOTE: this is probably wrong for emulated objects.
...
            } else if (needWriteAnyWay) {
...
```

Later, a different commit simplified this splitting and removed the needWriteAnyway which led to the current issue. Right now we are merging canBeFound condition with needWriteAnyway by relating canBeFound with the target instead of the source.

io/io/src/TFileMerger.cxx
```diff
@@ -829,7 +829,8 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type,
          delete ndir;
       }
    } else if (!canBeFound) { // Don't write the partial result for TTree and TH1
```
ferdymercury:
```suggestion
   } else if (!canBeFound) { // object (TTree, TH1) is not yet owned by the target, thus write it
```

pcanal (approved):
LGTM.

PR #18687, opened 2025-05-12
[skip-ci] Drop PyROOT from the README

aaronj0:
In the direction of updating references to "PyROOT" to ROOT's Python Interface/Bindings


vepadulano (approved):
Thanks!

PR #18690, opened 2025-05-12
[RF] Use `std::erf` and `std::erfc` in RooFit math functions

guitargeek:
This makes the code more portable, and the `TMath` functions call the standard library functions internally anyway.

Also gives test coverage for https://github.com/vgvassilev/clad/issues/1289.


vgvassilev (approved):
Nice. Looks like a good improvement to me.

PR #18692, opened 2025-05-12
[roottest] Remove an unstable test.

hageboeck:
This test was checking the amount of VMem before and after a graphics call to figure out if the interpreter uses memory. Unfortunately, the amount of VMem is not as stable as anticipated. This lead to spurious failures on Mac14.

We discussed with Danilo and Vincenzo that it might be better to remove this test.


hageboeck:
One example:
https://github.com/root-project/root/actions/runs/14941045537/job/41985202760

vepadulano (approved):

PR #18695, opened 2025-05-12
[skip-ci][ntuple] Add notes about disabled type checks

hahnjo:
Also add method documentations in `REntry`.

### DIFF
--- a/tree/ntuple/inc/ROOT/REntry.hxx
+++ b/tree/ntuple/inc/ROOT/REntry.hxx
@@ -186,6 +186,9 @@ public:
    ///
    /// Ownership is shared with the caller and the object will be kept alive until it is replaced (by a call to
    /// EmplaceNewValue, BindValue, or BindRawPtr) or the entry is destructed.
+   ///
+   /// **Note**: if `T = void`, type checks are disabled. It is the caller's responsibility to match the field and
+   /// object types.
    template <typename T>
    void BindValue(std::string_view fieldName, std::shared_ptr<T> objPtr)
    {
@@ -207,6 +210,9 @@ public:
    ///
    /// The caller retains ownership of the object and must ensure it is kept alive when reading or writing using the
    /// entry.
+   ///
+   /// **Note**: if `T = void`, type checks are disabled. It is the caller's responsibility to match the field and
+   /// object types.
    template <typename T>
    void BindRawPtr(std::string_view fieldName, T *rawPtr)
    {
@@ -227,6 +233,9 @@ public:
    /// Get the (typed) pointer to the value for the field referenced by `token`.
    ///
    /// Ownership is shared and the caller can continue to use the object after the entry is destructed.
+   ///
+   /// **Note**: if `T = void`, type checks are disabled. It is the caller's responsibility to use the returned pointer
+   /// according to the field type.
    template <typename T>
    std::shared_ptr<T> GetPtr(std::string_view fieldName) const
    {
--- a/tree/ntuple/inc/ROOT/REntry.hxx
+++ b/tree/ntuple/inc/ROOT/REntry.hxx
@@ -45,7 +45,7 @@ class RNTupleJoinProcessor;
 /**
 \class ROOT::REntry
 \ingroup NTuple
-\brief The REntry is a collection of values in an ntuple corresponding to a complete row in the data set.
+\brief The REntry is a collection of values in an RNTuple corresponding to a complete row in the data set.
 
 The entry provides a memory-managed binder for a set of values read from fields in an RNTuple. The memory locations that are associated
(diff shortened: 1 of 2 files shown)


tree/ntuple/inc/ROOT/REntry.hxx
```diff
@@ -45,7 +45,7 @@ class RNTupleJoinProcessor;
 /**
 \class ROOT::REntry
 \ingroup NTuple
-\brief The REntry is a collection of values in an ntuple corresponding to a complete row in the data set.
+\brief The REntry is a collection of values in an RNTuple corresponding to a complete row in the data set.
```
pcanal, line 48: Does the next line (`The memory locations that are associated with values are managed through shared pointers.`) need to be updated when taking in consideration the raw bindings?

Is clear from other context (am I even right ;)) that 'data set` means the data in the (potentially sub-)set of fields described by the model used to create the `REntry` (I assume it is clear).

tree/ntuple/inc/ROOT/REntry.hxx
```diff
@@ -177,12 +182,22 @@ public:
       fValues[token.fIndex].Bind(objPtr);
    }
 
+   /// Bind the value for the field, referenced by its name, to `objPtr`.
+   ///
+   /// Ownership is shared with the caller and the object will be kept alive until it is replaced (by a call to
+   /// EmplaceNewValue, BindValue, or BindRawPtr) or the entry is destructed.
+   ///
+   /// **Note**: if `T = void`, type checks are disabled. It is the caller's responsibility to match the field and
+   /// object types.
```
pcanal, line 191: I am not sure but maybe either
```suggestion
   /// object's type.
```
or
```suggestion
   /// objects' type.
```

tree/ntuple/inc/ROOT/REntry.hxx
```diff
@@ -191,12 +206,22 @@ public:
       fValues[token.fIndex].BindRawPtr(rawPtr);
    }
 
+   /// Bind the value for the field, referenced by its name, to `rawPtr`.
+   ///
+   /// The caller retains ownership of the object and must ensure it is kept alive when reading or writing using the
+   /// entry.
+   ///
+   /// **Note**: if `T = void`, type checks are disabled. It is the caller's responsibility to match the field and
+   /// object types.
```
pcanal, line 215: Same as https://github.com/root-project/root/pull/18695/files#r2087320742

tree/ntuple/inc/ROOT/RNTupleReader.hxx
```diff
@@ -287,6 +291,11 @@ public:
    ///    std::cout << i << ": " << *pt << "\n";
    /// }
    /// ~~~
+   ///
+   /// **Note**: if `T = void`, type checks are disabled. It is the caller's responsibility to match the field and
+   /// object types. It is strongly recommended to use an overload that allows passing the `typeName`, such as
```
pcanal, line 296: See https://github.com/root-project/root/pull/18695/files#r2087320742

tree/ntuple/inc/ROOT/REntry.hxx
hahnjo: I thought about this and my feeling (not a native speaker) was that "field and object types" conveys the message the best. See for example also https://www.w3schools.com/typescript/typescript_object_types.php for the term "object type".

tree/ntuple/inc/ROOT/REntry.hxx
hahnjo: Raw pointers are internally encapsulated in a (non-owning) shared pointer.

I don't fully understand your second point, is it about imposing a model to read only a subset of the dataset and the nuance that the `REntry` is then not a complete row of the dataset on disk? We can maybe word this differently, but I don't see this as related to the notes about type checking.

tree/ntuple/inc/ROOT/REntry.hxx
silverweed: Pretty sure the original formulation is ok.

jblomer (approved):

tree/ntuple/inc/ROOT/REntry.hxx
pcanal: >  but I don't see this as related to the notes about type checking.

It is not.  It is indeed 'just' about the semantic of `complete` and/or `data set` in `complete row in the data set` (i.e. yes it is "about imposing a model to read only a subset of the dataset and the nuance that the REntry is then not a complete row of the dataset on disk").

pcanal: Yes since I can not think of succinct way of making the statement clearer ....

PR #18696, opened 2025-05-12
build: Actually record dep from RooFitJSONInterface to Core. 

### LINKED ISSUE #18669: Incremental builds with `dev=ON` break when modifying Core header
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Locally testing https://github.com/root-project/root/pull/18601 in a build configured with `dev=ON`, I can reproduce the crashes when generating `RooFitHS3.pcm`.

### Reproducer

1. Configure a build with `dev=ON`.
2. Fully build `master`.
3. Apply the changes of https://github.com/root-project/root/pull/18601 (or revert them, if they were merged in the meantime) - it appears that just `touch`ing the header is not sufficient to trigger the problem.
4. Try to rebuild `RooFitHS3.pcm` incrementally, without other prior cleanups.

```
<<< cling interactive line includer >>>: fatal error: module file '/path/to/build/lib/Core.pcm' is out of date and needs to be rebuilt: module file out of date
rootcling: /path/to/src/interpreter/llvm-project/clang/include/clang/Basic/SourceLocation.h:340: unsigned int clang::PresumedLoc::getLine() const: Assertion `isValid()' failed.
```

### ROOT version

`master`

### Installation method

from source

### Operating system

Linux

### Additional context

Probably caused by / uncovered by #18573

pcanal:
Also mark explicitly the dependency of a dictionary source file and its pcm on the pcm of the libraries it depends on.

This fixes #18669.

#18669 was due to the missing dependency of `[G__]RooFitJSONInterface` on `Core`.

See more detail in #18669

### DIFF
--- a/cmake/modules/RootMacros.cmake
+++ b/cmake/modules/RootMacros.cmake
@@ -592,6 +592,12 @@ function(ROOT_GENERATE_DICTIONARY dictionary)
       set(dependent_pcm ${libprefix}${dep}_rdict.pcm)
       if (runtime_cxxmodules AND NOT dep IN_LIST local_no_cxxmodules)
         set(dependent_pcm ${dep}.pcm)
+        if(TARGET ${dep})
+          get_target_property(_dep_pcm_filename ${dep} ROOT_PCM_FILENAME)
+          if(_dep_pcm_filename)
+            list(APPEND pcm_dependencies ${_dep_pcm_filename})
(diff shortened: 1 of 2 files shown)


hageboeck (approved):
This looks like the right thing to do.

pcanal:
The failures are unrelated.

PR #18699, opened 2025-05-13
[RF] Implement codegen support for the RooWrapperPdf

guitargeek:
The implementation is trivial, because it only has to forward to the
function that is wrapped. Needed for the RooFit SBI workflows.

Also implement a unit test for it and avoid some code duplication in the testing code.


vgvassilev (approved):
Reducing the overall code size is probably a good idea. LGTM!

PR #18702, opened 2025-05-13
[roottest] use cmake for `io-xml` tests, modernize code

linev:
* Like in https://github.com/root-project/root/pull/18483, create library with used tested classes and load id.
* Only for SAX parser testing check if xml is build
* Use nullptr, auto, std::cout
* Add missing includes
* Keep all ref files as is


roottest/root/io/xml/LinkDef.h
```diff
@@ -0,0 +1,17 @@
+#ifdef __CLING__
+
+#pragma link off all    globals;
+#pragma link off all    classes;
+#pragma link off all    functions;
```
pcanal, line 5: Really minor item.  Those are no longer needed (they are off by default).

roottest/root/io/xml/LinkDef.h
```diff
@@ -0,0 +1,17 @@
+#ifdef __CLING__
```
pcanal, line 1: Another minor item
```suggestion
#ifdef __ROOTCLING__
```

roottest/root/io/xml/CMakeLists.txt
```diff
@@ -1,9 +1,37 @@
...
+ROOTTEST_ADD_TEST(enginexml
+                  MACRO runenginexml.C
+                  WORKING_DIR ${CMAKE_CURRENT_SOURCE_DIR}
+                  OUTREF enginexml.ref)
+
+ROOTTEST_ADD_TEST(XmlDir
+                  MACRO runXmlDir.C
+                  OUTREF XmlDir.ref)
+
+ROOTTEST_GENERATE_DICTIONARY(libXmlTestClasses
```
pcanal, line 22:
```suggestion
ROOTTEST_GENERATE_DICTIONARY(libXmlTestDictionaries
```
Part of my (slight) original confusion is  that library name is a slight misnomer (it does not contain any of the implementation of the TestClasses, just their dictionaries).

pcanal (approved):
LGTM.

PR #18705, opened 2025-05-13
[roottest] use cmake for some root/io tests

linev:
Making clear dependency
Generate dictionary where required
Avoid temporary files in source directory
`root-io-tuple` test disabled on Windows, was never tested before there


pcanal:
What step did you take to ensure that all the implicitly generated Makefile test have been ported?

linev:
> implicitly generated Makefile test have been ported

To my understanding there are no interim Makefiles where generated by touched tests.

roottest/root/io/TFile/.rootrc
```diff
@@ -1 +0,0 @@
-Rint.History:            .root_hist
```
pcanal, line 1: What is the advantage of removing the `.rootrc`?

linev: It was there to avoid putting roottest invocations in global ROOT history.
But now it managed centrally and not need to be done individually in each subdirectory.

pcanal: What do you mean by 'managed' centrally?

linev: There is `.rootrc` already in `roottest/root` directory - actually no idea why.

As I can see all `gROOT->ProcessLine` invocations not landing in `.root_hist` file and therefore there is no need to redefine it in custom `.rootrc`.

pcanal: Maybe because there is still the instance that was copied over before your remove it.  Either way adding or removing them is not related to the migration from Makefile to CMakeLists.  Please consider leaving them for now.

pcanal: ( in addition, some `.rootrc` are more that just the setting of the root history files )

pcanal: Why? `ROOTTEST_ADD_TESTDIRS` is supposed to do it automatically.

linev: Moreover - `ROOTTEST_ADD_TESTDIRS` automatically create `.rootrc` for every sub-directory.
Therefore there is no need to have such generic `.rootrc` file in each test.

roottest/root/io/TFile/CMakeLists.txt
```diff
@@ -1,9 +1,30 @@
...
-#
-#-------------------------------------------------------------------------------
-if(geom)
-  ROOTTEST_ADD_OLDTEST()
+ROOTTEST_ADD_TEST(assert_tcontext_thread_hoping
+                  MACRO assert_tcontext_thread_hoping.cxx
+                  OUTREF assert_tcontext_thread_hoping.ref)
+
+
+if(CMAKE_SIZEOF_VOID_P EQUAL 8)
```
pcanal, line 6: The existing test was not conditional.   Why remove the 32 bit test?

linev: I forgot that we still have Windows 32 bit testing. 
I will add it back now

linev: Now I remember - test was doing nothing in case of 32 bit. Check - it starts with  `#ifdef R__B64`
One also can see in 32-bit ref file - it is empty.

Therefore there is no any advantage to run dummy macro in Win32.

pcanal: Indeed.  That test does not make sense to run on 32bits.  Thanks.

roottest/root/io/tuple/CMakeLists.txt
```diff
@@ -1,7 +1,6 @@
-#-------------------------------------------------------------------------------
-#
-# Placeholder file to translate the tests to the new CTest system. Meanwhile we
-# define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
-#
-#-------------------------------------------------------------------------------
-ROOTTEST_ADD_OLDTEST()
+if(NOT MSVC)
```
pcanal, line 1: As far as I can tell the test was run on MSVC.  It is meant to check that all the files are readable and all the platforms.

roottest/root/io/webfile/CMakeLists.txt
```diff
@@ -1,9 +1,5 @@
-#-------------------------------------------------------------------------------
-#
-# Placeholder file to translate the tests to the new CTest system. Meanwhile we
-# define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
-#
-#-------------------------------------------------------------------------------
-if(geom AND davix)
-  ROOTTEST_ADD_OLDTEST()
+if(davix)
```
pcanal, line 1:
```suggestion
if(geom AND davix)
```

roottest/root/io/tuple/CMakeLists.txt
linev: According to @bellenot - none of Makefile-based tests were running on Windows.
And this test failing now.
Therefore disable it and Bertrand may try to debug it later.

pcanal: Let's try this one :)

pcanal: I.e. you are right the Makefile test were not running because there is no 'gmake' on Windows (and thus not necessarily because they don't work).

pcanal: I had not noticed you already tried.  What was the error?

linev: I create PR https://github.com/root-project/root/pull/18815 to reproduce problem.
The output is:
```
Writing tuple.root
  Reading tuple.root
  Error in <read>: Component 0 incorrect read 0 instead of 33
  Error in <read>: Component 1 incorrect read 0.000000 instead of 66.666667
  Reading tuple.macos.root
  Error in <read>: Component 0 incorrect read 0 instead of 33
  Error in <read>: Component 1 incorrect read 0.000000 instead of 66.666667
  Error in <read>: Component 2 incorrect read -85 instead of 13
  Reading tuple.x86_64.root
  Error in <read>: Component 0 incorrect read 0 instead of 33
  Error in <read>: Component 1 incorrect read 0.000000 instead of 66.666667
  Error in <read>: Component 2 incorrect read -85 instead of 13
```

pcanal: That's sad but clear :(

pcanal (approved):
Thanks.

PR #18708, opened 2025-05-13
[v636] build: Actually record dep from RooFitJSONInterface to Core.

### LINKED ISSUE #18669: Incremental builds with `dev=ON` break when modifying Core header

pcanal:
backport of https://github.com/root-project/root/pull/18696

Also mark explicitly the dependency of a dictionary source file and its pcm on the pcm of the libraries it depends on.

This fixes #18669.

#18669 was due to the missing dependency of `[G__]RooFitJSONInterface` on `Core`.

See more detail in #18669

### DIFF
--- a/cmake/modules/RootMacros.cmake
+++ b/cmake/modules/RootMacros.cmake
@@ -592,6 +592,12 @@ function(ROOT_GENERATE_DICTIONARY dictionary)
       set(dependent_pcm ${libprefix}${dep}_rdict.pcm)
       if (runtime_cxxmodules AND NOT dep IN_LIST local_no_cxxmodules)
         set(dependent_pcm ${dep}.pcm)
+        if(TARGET ${dep})
+          get_target_property(_dep_pcm_filename ${dep} ROOT_PCM_FILENAME)
+          if(_dep_pcm_filename)
+            list(APPEND pcm_dependencies ${_dep_pcm_filename})
+          endif()
+        endif()
       endif()
       set(newargs ${newargs} -m  ${dependent_pcm})
     endforeach()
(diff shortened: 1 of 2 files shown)


guitargeek:
Maybe I can use this as a precedent to reopen my PR :)
 https://github.com/root-project/root/pull/18640

hahnjo:
This cannot fix https://github.com/root-project/root/issues/18669 which is not present in 6.36...

pcanal:
> This cannot fix https://github.com/root-project/root/issues/18669 which is not present in 6.36...

Why/how is it not in 6.36?

hahnjo:
> > This cannot fix #18669 which is not present in 6.36...
> 
> Why/how is it not in 6.36?

Hm, I was under the impression that this only happened since https://github.com/root-project/root/pull/18573, but it's indeed possible to reproduce in `v6-36-00-patches`. Which leaves me confused why we didn't see this before... but fair enough

pcanal:
> Which leaves me confused why we didn't see this before...

In the CI, it was not seen (recently at least) because the build where never incremental (all dictionary rebuild all the time).  For local build, I suppose 'we' just worked around it by rebuilding/cleaning.

pcanal:
> Maybe I can use this as a precedent to reopen my PR :)

@guitargeek One major difference is this PR is strictly a bug fix.  The PR you mentioned is more similar to https://github.com/root-project/root/pull/18572 which can not be merge as is since it is a mix of minor bug fix and slight behavior changes.  Even-though #18572 appeared harmless, it uncovered a pre-existing issue and thus lead to unexpected (seemingly unrelated) failures.

PR #18723, opened 2025-05-14
[v636] [core] Fix a TClass lookup for STL containers / fix the crash in df104_CSVDataSource

### LINKED ISSUE #18654: TClassEdit name normalisation fails for unordered_map, leading to a crash in df104_CSVDataSource
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

The Python version of df104_CSV frequently crashes on Ubuntu when Cppyy queries types `unordered_map<string,char>`. Although a TClass for this might exist, a query for
```
std::unordered_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, char, std::hash<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::equal_to<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, char> > >
```
might not normalise to the same class, so another TClass instance is created. During its creation, it deletes the instance for `unordered_map<string,char>`, leading to a dangling pointer and a use-after-free crash.

### Reproducer

```
   auto first = TClass::GetClass("unordered_map<string,char>", true, true);
   std::unordered_map<std::string, char> map;
   auto second = first->GetActualClass(&map);
   EXPECT_EQ(first, second);
```

### ROOT version

Master and v6.36

### Installation method

Source

### Operating system

All, but the crash manifests on Ubuntu

### Additional context

_No response_

hageboeck:
Backport of #18625 

Fix #18654

### DIFF
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -33,40 +33,40 @@
 using std::string, std::string_view, std::vector, std::set;
 
 namespace {
-   static TClassEdit::TInterpreterLookupHelper *gInterpreterHelper = nullptr;
+static TClassEdit::TInterpreterLookupHelper *gInterpreterHelper = nullptr;
 
-   template <typename T>
-   struct ShuttingDownSignaler : public T {
-      using T::T;
+template <typename T>
+struct ShuttingDownSignaler : public T {
+   using T::T;
 
-      ShuttingDownSignaler() = default;
-      ShuttingDownSignaler(T &&in) : T(std::move(in)) {}
+   ShuttingDownSignaler() = default;
+   ShuttingDownSignaler(T &&in) : T(std::move(in)) {}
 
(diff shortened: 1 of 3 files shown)


pcanal (approved):

ferdymercury:
This probably has nothing to do with this PR, but just in case: in https://github.com/root-project/root/actions/runs/15023476682/job/42218306209?pr=18719 line 18883 I see that in Windows, the type name includes the keyword "class" for some default allocators:

std::vector float expected, but Windows says it is std::vector<float, class std::allocator<float>>

Which reminds me of this line: https://github.com/root-project/root/pull/17048/files#diff-70dbabe20d3c49cf037fb2117892e3a611719d799abaae22b8a6563cb7f5b6e8R625

PR #18726, opened 2025-05-14
[roottest] fix leak when using TSystem::ConcatFileName

linev:
Return value must be released.

### DIFF
--- a/roottest/root/io/newstl/TestHelpers.C
+++ b/roottest/root/io/newstl/TestHelpers.C
@@ -7,7 +7,7 @@
 #include <utility>
 
 void fillListOfDir(const TString &directory, TList &l) {
-   
+
    void *dir = gSystem->OpenDirectory(directory);
 
    const char *file = 0;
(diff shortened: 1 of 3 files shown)


ferdymercury:
Thanks for noticing! I think there are two more leaks with the same function in RooWorkspace.cxx ? Since the returned value is "copied" to a string rather than "moved" ?

roottest/root/io/newstl/TestHelpers.C
```diff
@@ -25,11 +25,17 @@ void fillListOfDir(const TString &directory, TList &l) {
          TString s = file;
          if ( (basename!=file) && s.Index(re) == kNPOS) continue;
 
-	 TString dirname = file;
-	 if (directory != ".")
-	    dirname = gSystem->ConcatFileName(directory, file);
+	      TString dirname = file;
```
pcanal, line 28: There seems to be (preexisiting) tab characters in the source file.  On github the result look oddly aligned.

linev: I replace `tabs` by normal spaces here

PR #18730, opened 2025-05-15
[RF] Fix memory leak of ConcatFileName return value in RooWorkspace

guitargeek:
Fix a memory leak of the `TSystem::ConcatFileName()` return value in RooWorkspace, and also refactor the code to avoid repetition.

Thanks a lot to @ferdymercury for noticing this!
https://github.com/root-project/root/pull/18726#issuecomment-2880508421

### DIFF
--- a/roofit/roofitcore/src/RooWorkspace.cxx
+++ b/roofit/roofitcore/src/RooWorkspace.cxx
@@ -1481,6 +1481,25 @@ std::list<TObject*> RooWorkspace::allGenericObjects() const
 }
 
 
+namespace {
+
+std::string findFileInPath(std::string const &file, std::list<std::string> const &dirList)
+{
+   // Check list of additional paths
+   for (std::string const &diter : dirList) {
+
+      char *cpath = gSystem->ConcatFileName(diter.c_str(), file.c_str());
+      std::string path = cpath;
+      delete[] cpath;
+      if (!gSystem->AccessPathName(path.c_str())) {
+         // found file
+         return path;
+      }
+   }
+   return "";
+}
+
+} // namespace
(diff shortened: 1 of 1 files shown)


ferdymercury (approved):
Thanks for the swift fix!

roofit/roofitcore/src/RooWorkspace.cxx
```diff
@@ -1481,6 +1481,25 @@ std::list<TObject*> RooWorkspace::allGenericObjects() const
...
 
 
+namespace {
+
+std::string findFileInPath(std::string const &file, std::list<std::string> const &dirList)
+{
+   // Check list of additional paths
+   for (std::string const &diter : dirList) {
+
+      char *cpath = gSystem->ConcatFileName(diter.c_str(), file.c_str());
```
silverweed, line 1491: This `ConcatFileName` function is horrible: any chance we could deprecate it and provide a new one that returns a `std::string`? (Looking at its implementation it doesn't even make sense that it returns a `char *` rather than a `TString`...)

guitargeek: Of course it is horrible, haha! But can you maybe still approve the fix I have now, so I can merge and backport it? For the future, I agree we should absolutely get rid of this function.

silverweed (approved):

roofit/roofitcore/src/RooWorkspace.cxx
silverweed: Yes, sorry, I wasn't implying that it should be done in this PR, just something to think of for the future :)

Issue #11712, opened 2022-11-16
1-1 not the same as (1-1) in TTree arrays indices


couet:
### Describe the bug
When accessing arrays in TTree::Scan() (or Draw()) if the array index is a formula like 1-1 it should be enclosed by "()" otherwise it does not work as expected. See:

```
root [3] t1->Scan("energy[1-1]:energy[(1-1)]:energy[0]")
************************************************
*    Row   * energy[1- * energy[(1 * energy[0] *
************************************************
*        0 * 275.16714 * 326.72375 * 326.72375 *
*        1 * 213.84567 * 261.90139 * 261.90139 *
*        2 * 413.62146 * 415.01107 * 415.01107 *
```
The problem was [posted on the forum](https://root-forum.cern.ch/t/strange-operations-with-array-indices/52384).

### Expected behavior
`1-1` should be the same as `(1-1)`

### To Reproduce
Use any tree with an array `x` and access` x[1-1]` and `x[(1-1)]`

### Setup
Any ROOT version.

JIRA: https://its.cern.ch/jira/browse/ROOT-5946

PR #18579, opened 2025-05-01
[hist] Do not parse string to int if it's not a digit

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/11712

## Checklist:

- [x] tested changes locally
- [ ] updated the docs (if necessary)

PR #18739, opened 2025-05-15
[hist, tree] Do not parse string to int if it's not a digit

dpiparo:
A re-organisation of the commits by @ferdymercury in this PR https://github.com/root-project/root/pull/18579 . All the credit goes to the original author - here the original commits are just reshuffled.

Fixes https://github.com/root-project/root/issues/11712


pcanal (approved):
Thanks.

PR #18742, opened 2025-05-16
[RF][HF] Avoid unnecessary creation of Asimov datasets

guitargeek:
Creating the per-channel asimov dataset is only meaningful if we
actually create the files with the stored per-channel workspaces.
Otherwise, we just spend time calculating something that gets thrown
away anyway (for the combined workspace, we'll create a new Asimov).

This improves the performance for large workspace creation in TRexFitter
by about 20 %, which can be significant if the total runtime is in the
order of hours.

FYI @TomasDado, @superharz

### DIFF
--- a/roofit/roofitcore/inc/RooFit/ModelConfig.h
+++ b/roofit/roofitcore/inc/RooFit/ModelConfig.h
@@ -309,7 +309,9 @@ public:
    /// alias for GetWS()
    RooWorkspace *GetWorkspace() const { return GetWS(); }
 
-   void GuessObsAndNuisance(const RooAbsData &data, bool printModelConfig = true);
+   void GuessObsAndNuisance(const RooArgSet &obsSet, bool printModelConfig = true);
+
+   inline void GuessObsAndNuisance(const RooDataSet &data, bool printModelConfig = true) { return GuessObsAndNuisance(data, printModelConfig); }
 
    /// overload the print method
    void Print(Option_t *option = "") const override;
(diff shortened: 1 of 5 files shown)


dpiparo:
once integrated, could you please backport to 6.32 @guitargeek ?

guitargeek:
You mean 6.36? Yes I'll do! Can you approve this PR please?

dpiparo:
I indeed meant 6.36.

dpiparo (approved):
Thanks!

roofit/roofitcore/inc/RooFit/ModelConfig.h
```diff
@@ -309,7 +309,9 @@ class ModelConfig final : public TNamed, public RooWorkspaceHandle {
    /// alias for GetWS()
    RooWorkspace *GetWorkspace() const { return GetWS(); }
 
-   void GuessObsAndNuisance(const RooAbsData &data, bool printModelConfig = true);
+   void GuessObsAndNuisance(const RooArgSet &obsSet, bool printModelConfig = true);
+
+   inline void GuessObsAndNuisance(const RooDataSet &data, bool printModelConfig = true) { return GuessObsAndNuisance(data, printModelConfig); }
```
hahnjo, line 314: There is a Clang warning for this function definition:
```
roofit/roofitcore/inc/RooFit/ModelConfig.h:314:90: warning: all paths through this function will call itself [-Winfinite-recursion]
  314 |    inline void GuessObsAndNuisance(const RooDataSet &data, bool printModelConfig = true) { return GuessObsAndNuisance(data, printModelConfig); }
      |                                                                                          ^
```

To me, it seems the compiler is correct - is this missing something? Quickly looking at the Doxygen, maybe `*data.get()` to get a `RooArgSet`?

guitargeek: Ah yes, sure! Thanks a lot for the catch :+1: I kept this signature for backwards compatibility but didn't use it, hence I missed this.

PR #18748, opened 2025-05-16
[core] Slightly improve TSystem::ConcatFileName()

silverweed:
Imho this function should be marked as deprecated and replaced with one that returns a `std::string` (or at least a `TString`). It's unclear to me why this wasn't done in the first place, considering the function internally allocates a TString anyway, but I assume it was created when TString wasn't a thing yet and later changed.

In the meantime, we can at least improve a bit its implementation. This PR does the following:

- save a couple of memory allocations per call (now guaranteed to only allocate once) and a bunch of memmoves;
- don't do the extra work of replacing '/' with '\\' in the entire string (that is not specified by the function's documentation);
- the function is not virtual anymore (although it internally uses a virtual method to get the correct directory separator; this could be avoided if had a platform-dependent `#define` or `constexpr` public constant for it, but this for now seems like the simplest solution).

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/base/inc/TSystem.h
+++ b/core/base/inc/TSystem.h
@@ -434,12 +434,13 @@ public:
    const char             *pwd() { return WorkingDirectory(); }
    virtual const char     *TempDirectory() const;
    virtual FILE           *TempFileName(TString &base, const char *dir = nullptr, const char *suffix = nullptr);
+   virtual char            DirectorySeparator() const { return '/'; }
 
    //---- Paths & Files
    virtual const char     *BaseName(const char *pathname);
    virtual const char     *DirName(const char *pathname);
    virtual TString         GetDirName(const char *pathname);
-   virtual char           *ConcatFileName(const char *dir, const char *name);
+   char                   *ConcatFileName(const char *dir, const char *name);
    virtual Bool_t          IsAbsoluteFileName(const char *dir);
    virtual Bool_t          IsFileInIncludePath(const char *name, char **fullpath = nullptr);
    virtual const char     *PrependPathName(const char *dir, TString& name);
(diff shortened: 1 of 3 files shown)


ferdymercury:
> It's unclear to me why this wasn't done in the first place, considering the function internally allocates a TString anyway, but I assume it was created when TString wasn't a thing yet and later changed.

Yep, it's because of https://github.com/root-project/root/commit/8f232bef23df9a2b3c4a6649a1493d8b34ac71f3

> * don't do the extra work of replacing '/' with '\' in the entire string (that is not specified by the function's documentation);

I saw some Windows comments here: https://github.com/root-project/root/commit/d3e9f54c0057f498ecc6a641237720fa93be2981

silverweed:
> I saw some Windows comments here: https://github.com/root-project/root/commit/d3e9f54c0057f498ecc6a641237720fa93be2981

Judging from the commit's comments, it seems this function really needs to do more than it says. In that case it might not be worth doing this change and risk breaking some obscure use case (which unfortunately doesn't seem to be covered by the tests, given they all pass even with the changed code)...

(I also noticed I didn't take into account the fact that `dir` and `name` can be null)

silverweed:
Closing because the performance improvement is not worth the extra code necessary to consider all corner cases.

pcanal:
> this function should be marked as deprecated and replaced 

It is already the case per se.  The replacement is `TSystem::PrependPathName`.   We had lost track of it but all the recent change replace to `ConcatFileName` should actually have been much better off by using `PrependPathName`.

pcanal:
If updating/replacing the code in `PrependPathName`, we should consider using `std::filesystem::path`

PR #18749, opened 2025-05-16
[ntuple] Adapt hashing to 32-bit

enirolf:
On 32-bit platforms or platforms where `std::size_t` is 32-bit instead of 64-bit, the current `mix` function doesn't work. The alternative added with this change also originates from the [`boost::hash_combine` writeup](https://www.boost.org/doc/libs/1_87_0/libs/container_hash/doc/html/hash.html#notes_hash_combine) linked in the `RCombinedJoinFieldValueHash` class documentation.

### DIFF
--- a/tree/ntuple/inc/ROOT/RNTupleJoinTable.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleJoinTable.hxx
@@ -118,11 +118,19 @@ private:
 
          inline std::size_t mix(std::size_t init) const
          {
+#ifdef R__B64
             init ^= init >> 32;
             init *= 0xe9846af9b1a615d;
             init ^= init >> 32;
             init *= 0xe9846af9b1a615d;
             init ^= init >> 28;
+#else
+            init ^= init >> 16;
+            init *= 0x21f0aaad;
+            init ^= init >> 15;
+            init *= 0x735a2d97;
+            init ^= init >> 15;
+#endif
             return init;
          }
       };


tree/ntuple/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -118,11 +118,19 @@ private:
...
+               init ^= init >> 15;
+               init *= 0x735a2d97;
+               init ^= init >> 15;
+            } else { // 64-bit
+               init ^= init >> 32;
+               init *= 0xe9846af9b1a615d;
+               init ^= init >> 32;
+               init *= 0xe9846af9b1a615d;
+               init ^= init >> 28;
+            }
```
bellenot, line 133:
```suggestion
#if defined(__i386__) // 32-bit
               init ^= init >> 16;
               init *= 0x21f0aaad;
               init ^= init >> 15;
               init *= 0x735a2d97;
               init ^= init >> 15;
#else
               init ^= init >> 32;
               init *= 0xe9846af9b1a615d;
               init ^= init >> 32;
               init *= 0xe9846af9b1a615d;
               init ^= init >> 28;
#endif
```
to prevent compilation warning on x86

bellenot (changes requested):

tree/ntuple/inc/ROOT/RNTupleJoinTable.hxx
bellenot, line 133: To prevent compilation warnings:
```suggestion
#if defined(__x86_64__) || defined(_WIN64)
               init ^= init >> 32;
               init *= 0xe9846af9b1a615d;
               init ^= init >> 32;
               init *= 0xe9846af9b1a615d;
               init ^= init >> 28;
#else // 32-bit
               init ^= init >> 16;
               init *= 0x21f0aaad;
               init ^= init >> 15;
               init *= 0x735a2d97;
               init ^= init >> 15;
#endif
```

bellenot (changes requested):

tree/ntuple/inc/ROOT/RNTupleJoinTable.hxx
```diff
@@ -118,11 +118,19 @@ private:
...
+               init *= 0x21f0aaad;
+               init ^= init >> 15;
+               init *= 0x735a2d97;
+               init ^= init >> 15;
+            } else { // 64-bit
+               init ^= init >> 32;
+               init *= 0xe9846af9b1a615d;
+               init ^= init >> 32;
+               init *= 0xe9846af9b1a615d;
+               init ^= init >> 28;
```
bellenot, line 132: To disable compilation warning on Windows 32 bit:
```suggestion
#ifdef _MSC_VER
#pragma warning(push)
#pragma warning(disable:4293)
#endif
               init ^= init >> 32;
               init *= 0xe9846af9b1a615d;
               init ^= init >> 32;
               init *= 0xe9846af9b1a615d;
               init ^= init >> 28;
#ifdef _MSC_VER
#pragma warning(pop)
#endif
```

enirolf: Rather than just suppressing the warning I would prefer to actually address the issue itself, which I think the latest push does. Please let me know if that works!

bellenot: Well, your latest push make the code use the 16 bit code on both 32 and 64 bit Windows (#ifdef _MSC_VER is generic for Visual Studio), and does not solve the problem on other platforms (e.g. 32 bit Linux), and is using the 16 bit solution on 64 bit architecture good enough?
I think using `if constexpr (sizeof(std::size_t) == 4) { // 32-bit` is the proper solution, one has simply to silence the compiler warning...

enirolf: For 32-bit Linux systems we would get the same compiler warnings with the `if constexpr`, and since we only release for 64-bit Linux I personally don't think it is worth manually checking for any other platform (especially considering this is still experimental code). The 32-bit solution should be good enough also for 64-bit systems given the expected values we'll be hashing with this.

bellenot: OK, then fine with me. But note that even if we don't release for 32 bit Linux, we are still supposed to support it, unless I missed something @dpiparo ?
And especially if the code produce an undefined behavior

bellenot: OK, so we don't support other 32 bit platforms than Windows, so all good

silverweed: I think we can just use the `R__B64` preprocessor macro defined in `ROOT/RConfig.hxx` to determine if we're in 32 bit or 64.

enirolf: Aha, I was looking for a macro like this but couldn't find it, thank you!

Issue #18751, opened 2025-05-16
Better thread safety in TClassEdit::CleanType


Dr15Jones:
### Explain what you would like to see improved and how.

The two `static` function variables at 
https://github.com/root-project/root/blob/600573f0d93ef302b81ffd2a12f31794b03d521c/core/foundation/src/TClassEdit.cxx#L1277
and
https://github.com/root-project/root/blob/600573f0d93ef302b81ffd2a12f31794b03d521c/core/foundation/src/TClassEdit.cxx#L1285

are not presently being modified in the function, but it would be better if they were both explicitly made const to guarantee they could not be modified. I.e.

```C++
static const char* const remove[] = {"class", "const", "volatile", nullptr};
```

```C++
static const std::vector<size_t> lengths{ initLengthsVector() };
```

### ROOT version

Present root master branch

### Installation method

build from source

### Operating system

all

PR #18756, opened 2025-05-16
[core] make static vars constexpr for thread safety v2

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/18751

All credit goes to @guitargeek

Is an optimized alternative to https://github.com/root-project/root/pull/18753

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -151,6 +151,8 @@ void RemoveScopeResolution(std::string &name)
    }
 }
 
+/// Get an array with the respective lengths of an array of strings, e.g. {"ab", "abc"} will return {2,3}
+/// The first parameter is a variadic parameter, pass std::make_index_sequence<std::size(stringArray)>()
 template <std::size_t N, std::size_t... Is>
 constexpr std::array<std::size_t, N>
 getLengthsArray(std::index_sequence<Is...>, std::array<const char *, N> const &stringArray)
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -27,6 +27,7 @@
 #include <string_view>
 #include <algorithm>
 #include <string>
+#include <utility>
 
 #include "TSpinLockGuard.h"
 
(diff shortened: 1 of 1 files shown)


pcanal:
This looks fine too ... Albeit is the slight extra complexity (plus churn) worth it?  Maybe.

core/foundation/src/TClassEdit.cxx
```diff
@@ -1274,15 +1284,8 @@ int TClassEdit::GetSplit(const char *type, vector<string>& output, int &nestedLo
...
-   static const char* remove[] = {"class", "const", "volatile", nullptr};
-   auto initLengthsVector = []() {
-      std::vector<size_t> create_lengths;
-      for (int k=0; remove[k]; ++k) {
-         create_lengths.push_back(strlen(remove[k]));
-      }
-      return create_lengths;
-   };
-   static std::vector<size_t> lengths{ initLengthsVector() };
+   constexpr std::array<const char *, 3> remove{"class", "const", "volatile"};
```
vepadulano, line 1287: constexpr does not imply static, so that keyword is still needed

core/foundation/src/TClassEdit.cxx
```diff
@@ -1274,15 +1284,8 @@ int TClassEdit::GetSplit(const char *type, vector<string>& output, int &nestedLo
...
+   constexpr auto lengths = getLengthsArray(std::make_index_sequence<std::size(remove)>(), remove);
```
vepadulano, line 1288: This should also work
```cpp
constexpr static auto lengths = []{
    constexpr std::array<std::size_t, 3> ret{};
    for(std::size_t i = 0; i < remove.size(); i++)
        ret[i] = remove[i];
    return ret;
}();
```

vepadulano (changes requested):

core/foundation/src/TClassEdit.cxx
ferdymercury: Explored here: https://github.com/root-project/root/pull/18757

core/foundation/src/TClassEdit.cxx
guitargeek: That's true, but if it's done at compile time it doesn't matter if it's static or not I think. But yeah why not!

ferdymercury: I guess it depends on `-O`
https://stackoverflow.com/a/76696023/7471760
related:
https://stackoverflow.com/questions/54466591/replacing-constants-when-to-use-static-constexpr-and-inline-constexpr

pcanal:
See https://github.com/root-project/root/pull/18757

PR #18757, opened 2025-05-16
[core] make static vars constexpr for thread safety v3

ferdymercury:
Fixes https://github.com/root-project/root/issues/18751

Please squash.

All credit goes to vepadulano and guitargeek

Alternatives to https://github.com/root-project/root/pull/18753 and https://github.com/root-project/root/pull/18756/

Of course, I could also make a version 4 which just does `lengths {5,5,8}` which is even more compact :) and this array did not change in 22 years, so...

### DIFF
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -1274,15 +1274,13 @@ int TClassEdit::GetSplit(const char *type, vector<string>& output, int &nestedLo
 
 string TClassEdit::CleanType(const char *typeDesc, int mode, const char **tail)
 {
-   static const char* remove[] = {"class", "const", "volatile", nullptr};
-   auto initLengthsVector = []() {
-      std::vector<size_t> create_lengths;
-      for (int k=0; remove[k]; ++k) {
-         create_lengths.push_back(strlen(remove[k]));
-      }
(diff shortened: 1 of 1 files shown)


core/foundation/src/TClassEdit.cxx
```diff
@@ -1274,15 +1274,13 @@ int TClassEdit::GetSplit(const char *type, vector<string>& output, int &nestedLo
...
-      std::vector<size_t> create_lengths;
-      for (int k=0; remove[k]; ++k) {
-         create_lengths.push_back(strlen(remove[k]));
-      }
-      return create_lengths;
-   };
-   static std::vector<size_t> lengths{ initLengthsVector() };
+   constexpr static std::array<const char *, 3> remove{"class", "const", "volatile"};
+   constexpr static auto lengths = []() constexpr {
+      std::array<std::size_t, 3> ret{};
```
guitargeek, line 1279:
```suggestion
      std::array<std::size_t, std::size(remove)> ret{};
```
To avoid the magic number three, if one wants, but yeah not a big deal! Super nice @vepadulano, I didn't know that for loops can be `constexpr` :+1::+1:

pcanal (approved):
Nice.  Thanks.

PR #18426, opened 2025-04-16
[math] Optimize CrystalBallIntegral to avoid issues with high N

ferdymercury:
# This Pull request:

## Changes or fixes:

Follow up of https://github.com/root-project/root/pull/17660 and https://github.com/root-project/root/pull/17893

## Checklist:

- [ ] tested changes locally
- [ ] updated the docs (if necessary)

PR #18761, opened 2025-05-17
[RDF] fix copy-paste typo in test and potential race condition

ferdymercury:
# This Pull request:

## Changes or fixes:

No idea whether it would solve the test failure found in https://github.com/root-project/root/actions/runs/15083048614/job/42402160290?pr=18426

See https://github.com/root-project/root/pull/18426

### DIFF
--- a/tree/dataframe/test/dataframe_friends.cxx
+++ b/tree/dataframe/test/dataframe_friends.cxx
@@ -305,10 +305,10 @@ void FillIndexedFriend(const char *mainfile, const char *auxfile)
    f2.Close();
 }
 
-void TestIndexedFriendChain()
+void TestIndexedFriendChain(bool useIMT = false)
 {
-   auto mainFile = "IndexedFriendChain_main.root";
-   auto auxFile = "IndexedFriendChain_aux.root";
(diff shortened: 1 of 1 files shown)


vepadulano:
Thank you @ferdymercury , I would first merge the other PR as I believe that the cause of the CI failure is more likely the issue with stack-based TTree and TFile. We can keep this open in case the error still resurfaces, although there are many test that use the same file name for the MT/non-MT versions.

ferdymercury:
> Thank you @ferdymercury , I would first merge the other PR as I believe that the cause of the CI failure is more likely the issue with stack-based TTree and TFile. We can keep this open in case the error still resurfaces, although there are many test that use the same file name for the MT/non-MT versions.

Thanks! Indeed the renaming is not relevant of MT vs non-MT.

But the important part is the "fix typo", since we are now calling twice TestIndexedFriendChain, from different tests, both from non-multithreaded mode, which could lead to a race.

ferdymercury:
Ok, thanks, I think I understand now, GoogleTests do not run in parallel within the same file, so I will close this...

PR #18770, opened 2025-05-19
[core,pyroot] add ROOT_BATCH environment variable

### LINKED ISSUE #18744: Add environment variable to enable batch mode
### Feature description

I suggest to introduce a new environment variable (e.g. `$ROOT_BATCH`) to configure batch mode as an alternative to the `-b` command line switch.

This is mainly relevant for pyROOT to ensure it launches in batch mode once imported (but without forcing an import, see below).

See https://github.com/root-project/root/compare/master...fwinkl:root:batch_env for a proposed implementation, which I am happy to turn into a PR if agreed that this is a useful feature.

### Alternatives considered

A review of current workarounds and their shortcomings: 

1.  Add `-b` to `sys.argv`:
     This may conflict with already existing user arguments.
2. `ROOT.gROOT.SetBatch()`:
    In experiment frameworks (e.g. athena) we usually do not want to force `import ROOT` unless it's actually needed by the workflow. But we also do not have control over every single place where `import ROOT` might be used.
3. `unset DISPLAY`:
    This effectively enables batch mode but it's specific to Linux and may affect other applications.

### Additional context

_No response_

fwinkl:
Add new `ROOT_BATCH` environment variable to initialize the batch mode as an alternative to using the `-b` command line switch.

Strictly speaking only the changes in `TApplication` are necessary (as that one always gets instantiated before `TROOT`). But logically it makes sense to also check the variable in `TROOT`.

Fixes #18744.


fwinkl:
> Thanks @fwinkl . Before starting the CI, would it be perhaps possible to steer the bahviour of ROOT through the value of the `ROOT_BATCH` environment variable (1 batch, 0 normal) and not its mere existence?

I considered that but then one would also need to decide what happens if someone uses the `root -b` command line switch and sets `ROOT_BATCH=0`. Which one should have precedence?

fwinkl:
Not clear to me if these test failure are related to my changes. I also added the changes to the tutorial tests as suggested by @guitargeek on #18744.

fwinkl:
@dpiparo, is there anything that needs to be done from my side to move this forward?

fwinkl:
I resolved the conflicts. I would appreciate if this could be merged soon. It's been sitting here since 4 months.

fwinkl:
Let me try to ping here again. @dpiparo or @guitargeek, could we please move this PR forward? I would really like to see this in the next ROOT release as it would make setting batch mode in ATLAS a lot easier than what we have now.

guitargeek:
> Let me try to ping here again. @dpiparo or @guitargeek, could we please move this PR forward? I would really like to see this in the next ROOT release as it would make setting batch mode in ATLAS a lot easier than what we have now.

Yes, sorry for missing this. I launched the tests to merge is ASAP. Thanks for the ping!

jmcarcell:
I'm also interested in this and will test it once merged. I don't see this variable being documented, and more in general, are these documented anywhere? For example, `ROOT_WEBENV`.

guitargeek:
> I'm also interested in this and will test it once merged. I don't see this variable being documented, and more in general, are these documented anywhere? For example, `ROOT_WEBENV`.

I think most of these variables are documented in the default `.rootrc` file. But given that we are close to the release, I'd suggest we first merge this PR and I follow up with the documentation later, to make sure that this gets into 6.38.

PR #18773, opened 2025-05-19
[core] Do not load MathCore lib at the prompt if this is a module build

dpiparo:


core/rint/src/TRint.cxx
ferdymercury, line 1: Thanks! There is a grammar problem in line 198

ferdymercury: Amd makes maybe more sense to move the ifndef before 198

core/rint/src/TRint.cxx
```diff
@@ -195,11 +195,13 @@ TRint::TRint(const char *appClassName, Int_t *argc, char **argv, void *options,
       PrintLogo(lite);
    }
 
-   // Explicitly load libMathCore it cannot be auto-loaded it when using one
-   // of its freestanding functions. Once functions can trigger autoloading we
-   // can get rid of this.
+#ifndef R__USE_CXXMODULES
+   // Explicitly load libMathCore. It cannot be auto-loaded it when using one
```
hageboeck, line 199:
```suggestion
   // Explicitly load libMathCore. It cannot be auto-loaded when using one
```

hageboeck (commented):
Could you add the reason for not loading it? It seems that this would prevent using the free-standing functions?
Or are these usable because they are in the module?

core/rint/src/TRint.cxx
dpiparo: Maybe one can be even more explicit. Autoloading through autoload keys in the rootmap files never worked for functions. This is why libMathCore has always been loaded automatically, given the widespread use of its free functions. However, modules fix this and we can load the library on demand, saving some time when starting ROOT (reading less files on CVMFS, for example) and saving some memory.

dpiparo: done.

ferdymercury (approved):

hageboeck (approved):
Thanks, now it makes a lot of sense

PR #18774, opened 2025-05-19
[roottest] Remove builtin googletest version from roottest

guitargeek:
Since `roottest` is now part of the main ROOT build, we don't need dedicated `builtin_gtest` logic for roottest.


pcanal (approved):
Thanks.

PR #18776, opened 2025-05-19
[RF] Avoid infinite recursion in `ModelConfig::GuessObsAndNuisance`

guitargeek:
This follows up on commit f15e44c26cb3, which missed to get the `RooArgSet` from the `RooDataSet` in the backwards-compatibility overload of `ModelConfig::GuessObsAndNuisance`.

Thanks to @hahnjo for noticing this!


dpiparo (approved):
thanks.

PR #18780, opened 2025-05-19
[interop] Upgrade CppInterOp to v1.7.0

aaronj0:
This is a pre-release PR to ensure the latest upgrades to CppInterOp are stable on all platforms for ROOT, and that newly added tests in `CppInterOpUnitTests` passes.

Some of the latest features and fixes are motivated by bugs in interpreter creation, llvm 20 support, wrapper generation in JitCall, and webassembly support.


aaronj0:
> LGTM assuming it's an update of the tag we had with no other changes that I cannot verify due to the large diff.

For that we have a workflow `Diff CppInterOp against upstream` that compares the hash/tag, similar to LLVM. If we have any patches, that job will fail and dump the diff

PR #18781, opened 2025-05-19
[build] Specify minimum required version for gtest

vepadulano:
Some of our tests use the macro GTEST_SKIP, which was added by https://github.com/google/googletest/pull/1544 and was only made available after gtest version 1.8. Thus, practically, we have an implicit minimum required version strictly greater than 1.8, i.e. 1.10.

This was in practice already known, whether explicitly or not. In fact, a system install of gtest on Alma 8 provides version 1.8 and our CI config enforces builtin_gtest=ON to get a later version of the package: https://github.com/root-project/root/blob/v6-37-01/.github/workflows/root-ci-config/buildconfig/alma8.txt

This commit simply solidifies the already established practice by requiring a minimum gtest version 1.10 when looking for the package.

Note: I found out about this because I was compiling ROOT on our Alma 8 CI image but without the `builtin_gtest=ON` option, which results in a failed build.


guitargeek (approved):
Thanks!

PR #18795, opened 2025-05-19
[roottest] Fixes to tree/cloning/CMakeLists.txt

guitargeek:
Remove unused variables and code branches.

Updated version of the following `roottest` PR:

  * https://github.com/root-project/roottest/pull/669

### DIFF
--- a/roottest/root/tree/cloning/CMakeLists.txt
+++ b/roottest/root/tree/cloning/CMakeLists.txt
@@ -1,76 +1,24 @@
-if(ROOTTEST_DIR)
-      set(ROOT_EVENT_DIR ${ROOTTEST_DIR}/root/treeformula/event/)
-else()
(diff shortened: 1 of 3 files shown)


roottest/root/tree/cloning/CMakeLists.txt
```diff
@@ -1,75 +1,23 @@
...
-      add_custom_target(event-generation ALL DEPENDS Event.root)
-
-      add_custom_command(
-        TARGET event-generation
-        COMMAND ${CMAKE_COMMAND} -E copy
-                Event.root
-                event1.root
-        COMMAND ${CMAKE_COMMAND} -E copy
-                Event.root
-                event2.root)
```
pcanal, line 42: Since `eventexe` is now always target, we could also have replace the `generate-eventfile.sh` with those (adapted to the current/new location) and thus enabling the test on Windows.

guitargeek: Great idea! I'll get there step by step :slightly_smiling_face:

PR #18797, opened 2025-05-20
[df] Remove unused constructor in RNTupleDS

vepadulano:
Since the removal of the corresponding factory function in https://github.com/root-project/root/commit/a85fc79a51cc9415397cad40c3e068d7831ff889, this constructor is not necessary anymore. Remove it for good measure.


dpiparo (approved):
thanks.

PR #18799, opened 2025-05-20
Revert "[cmake] Add rootcling dependencies for Windows"

bellenot:
This reverts commit 5cb988860f06ea39e51327386aa4bf6b950c46c4.
It has side effects when building the tests

### DIFF
--- a/cmake/modules/RootMacros.cmake
+++ b/cmake/modules/RootMacros.cmake
@@ -618,13 +618,13 @@ function(ROOT_GENERATE_DICTIONARY dictionary)
     set(pcm_name)
   else()
     if(CMAKE_PROJECT_NAME STREQUAL ROOT)
-      # Modules need RConfigure.h copied into include/.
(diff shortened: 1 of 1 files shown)


pcanal:
What is/was the negative side effect?

bellenot:
> What is/was the negative side effect?

rebuilding rootcling and re-generating the dictionaries when building the tests, hence making the whole testing process longer...

pcanal:
> rebuilding rootcling and re-generating the dictionaries when building the tests, hence making the whole testing process longer...

That's pointing to an/the underlying (known) problem about timestamps ....

pcanal (approved):
Fair enough but there is problem more to do ...

bellenot:
> > rebuilding rootcling and re-generating the dictionaries when building the tests, hence making the whole testing process longer...
> 
> That's pointing to an/the underlying (known) problem about timestamps ....

Right...

PR #18800, opened 2025-05-20
[io] TFileMerger: fail by default on invalid keys

### LINKED ISSUE #18665: hadd: Error in header does not cause hadd failure (as I'd expect)
### Check duplicate issues.

- [x] Checked for duplicates

### Description

File H4_v34b_1GeV_-27.7_1_20250422T165958Z_009584.root has some error in its header. This can be seen when trying to access the TTree VirtualDetector/BeforeTarget:

<pre>
root [3] VirtualDetector->cd()
(bool) true
root [4] BeforeTarget
Error R__unzip_header: error in header.  Values: 00
Error R__unzip_header: error in header.  Values: 00
Error R__unzip_header: error in header.  Values: 00
input_line_32:2:3: error: use of undeclared identifier 'BeforeTarget'
 (BeforeTarget)
  ^
Error in <HandleInterpreterException>: Error evaluating expression (BeforeTarget)
Execution of your code was aborted.
</pre>

When I hadd this with another similar file (H4_v34b_1GeV_-27.7_1_20250421T185009Z_001069.root), hadd does not explicitly fail -- I think it should explicitly fail if one does not provide the `-k` option

### Reproducer

Download input files: 
[H4_v34b_1GeV_-27.7_1_20250422T165958Z_009584.root ](https://www.dropbox.com/scl/fi/vsaxwsgmlq5o7xn6myvol/H4_v34b_1GeV_-27.7_1_20250422T165958Z_009584.root?rlkey=jt7e1yxs0lxjvrveg9iyamv5l&st=t8xlqlpo&dl=0)

[H4_v34b_1GeV_-27.7_1_20250421T185009Z_001069.root](https://www.dropbox.com/scl/fi/94cflt24ja4xgxpwyftv6/H4_v34b_1GeV_-27.7_1_20250421T185009Z_001069.root?rlkey=0x32dalvu8l5ve3kkvon21c8p&st=5m0focph&dl=0)

`hadd output.root H4_v34b_1GeV_-27.7_1_20250422T165958Z_009584.root H4_v34b_1GeV_-27.7_1_20250421T185009Z_001069.root`

### ROOT version

ROOT Version: 6.28/12
Built for linuxx8664gcc on Jan 30 2024, 08:17:35
From tags/v6-28-12@v6-28-12

### Installation method

pre-built binary

### Operating system

AL9 within SL7 container

### Additional context

_No response_

silverweed:
This commit introduces a `Error Behavior` flag in TFileMerger that determines what should happen upon encountering errors. The only place where this flag is currently honored is when reading an object's key inside MergeOne(): up until now if an error happened when retrieving an object from a file, the file would just be skipped. With the introduction of this flag, you can decide whether the file is skipped or if the merging process should abort.
For consistency with hadd, the default behavior has been set to Fail. Using `hadd -k` will set the behavior to Skip.

**NOTE that this is a potentially-breaking change**: should we keep Skip by default? But then how would we keep consistency with `hadd`?

This PR fixes #18665

### DIFF
--- a/io/io/src/TFileMerger.cxx
+++ b/io/io/src/TFileMerger.cxx
@@ -657,10 +657,18 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type, 
                   if (key2) {
                      hobj = key2->ReadObj();
                      if (!hobj) {
-                        Info("MergeRecursive", "could not read object for key {%s, %s}; skipping file %s",
-                           keyname, keytitle, nextsource->GetName());
-                              nextsource = (TFile*)sourcelist->After(nextsource);
-                              return kTRUE;
+                        switch (fErrBehavior) {
+                        case EErrorBehavior::kFailOnError:
+                           Error("MergeRecursive", "could not read object for key {%s, %s}; in file %s",
+                              keyname, keytitle, nextsource->GetName());
+                           nextsource = (TFile*)sourcelist->After(nextsource);
+                           return kFALSE;
+                        case EErrorBehavior::kSkipOnError:
+                           Info("MergeRecursive", "could not read object for key {%s, %s}; skipping file %s",
+                              keyname, keytitle, nextsource->GetName());
+                           nextsource = (TFile*)sourcelist->After(nextsource);
+                           return kTRUE;
+                        }
                      }
                      todelete.Add(hobj);
                   }
@@ -735,10 +743,18 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type, 
                   if (key2) {
                      TObject *hobj = key2->ReadObj();
                      if (!hobj) {
-                        Info("MergeRecursive", "could not read object for key {%s, %s}; skipping file %s",
+                        switch (fErrBehavior) {
+                        case EErrorBehavior::kFailOnError:
+                           Error("MergeRecursive", "could not read object for key {%s, %s}; in file %s",
                               keyname, keytitle, nextsource->GetName());
-                        nextsource = (TFile*)sourcelist->After(nextsource);
-                        return kTRUE;
+                           nextsource = (TFile*)sourcelist->After(nextsource);
+                           return kFALSE;
+                        case EErrorBehavior::kSkipOnError:
+                           Info("MergeRecursive", "could not read object for key {%s, %s}; skipping file %s",
+                              keyname, keytitle, nextsource->GetName());
+                           nextsource = (TFile*)sourcelist->After(nextsource);
+                           return kTRUE;
+                        }
                      }
                      // Set ownership for collections
                      if (hobj->InheritsFrom(TCollection::Class())) {
--- a/README/ReleaseNotes/v636/index.md
+++ b/README/ReleaseNotes/v636/index.md
@@ -181,6 +181,9 @@ This release changes that behavior, meaning the `Scale(bool)` command argument i
 * The logic to retrieve branch names of a TTree has been improved, leading to O(10) speedup. For example, an RDataFrame
   wrapping a TTree with O(10K) columns could be spending multiple minutes just to retrieve the full dataset schema,
   whereas now it will take seconds.
+* `TFileMerger`'s (and `hadd`'s) default behavior on encountering invalid/corrupt objects while merging has changed: previously the corrupt object would be skipped without aborting the merging process; now the merging process will abort by default. The "skip" behavior can be restored by:
+  * calling the new method `TFileMerger::SetErrorBehavior()` (if using `TFileMerger` directly), or
+  * using the `-k` flag if using `hadd`.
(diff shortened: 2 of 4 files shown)


io/io/src/TFileMerger.cxx
```diff
@@ -657,10 +657,18 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type,
                   if (key2) {
                      hobj = key2->ReadObj();
                      if (!hobj) {
-                        Info("MergeRecursive", "could not read object for key {%s, %s}; skipping file %s",
-                           keyname, keytitle, nextsource->GetName());
-                              nextsource = (TFile*)sourcelist->After(nextsource);
-                              return kTRUE;
+                        switch (fErrBehavior) {
```
pcanal, line 660: For consistent, we probably also need to add this logic on line 738+ (and if not add a documentation of why not).

pcanal:
Below is a set of description to make sure I understood the situations.

Oddly the current default can be seen as both consistent and inconsistent.  The case of corrupt object is currently handled `as if` the object was not the file in the first place.  On the other hand if there is corruption at the file level we fail but if there is corruption at the object level we don't fail.

I am leaning towards agreeing with this PR as a present-but-broken is still "meant" to be there and thus is 'surprising' to be not used in the result.

--- Current situation

Current default `hadd` behavior:
* fail on corrupt/missing files
* skip file with corrupt objects
* ignore  'missing' objects (i.e when there is no key for the name)

Current `hadd` behavior with `-k`
* skip on corrupt/missing files
* skip files with corrupt objects
* ignore 'missing' objects

---- This PR ---- 

New default `hadd` behavior:
* fail on corrupt/missing files
* fail on seeing any corrupt objects
* ignore  'missing' objects (i.e when there is no key for the name)

New `hadd` behavior with `-k`
** same as before **


---- If we want to keep current behavior  ---

We would need yet another flag/state so we could have:

Default `hadd` behavior:
** same as before **

`hadd` behavior with `-k`
** same as before **

`hadd` behavior with new flag: `fail on corrupt objects` (and/or `always fail`)
* fail on corrupt/missing files
* fail on seeing any corrupt objects
* ignore  'missing' objects (i.e when there is no key for the name)

silverweed:
@pcanal correct. 
Adding a new flag to hadd (and new enum value for the err behavior) is less than ideal, but it is indeed the only way to preserve the current behavior. 
It's hard for me to predict if the current behavior is relied upon by some users, so it would probably be the safest option (meaning that we keep the default behavior as-is and introduce this new "always fail" capability as opt-in)

io/io/src/TFileMerger.cxx
```diff
@@ -657,10 +657,18 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type,
...
+                        case EErrorBehavior::kFailOnError:
+                           Error("MergeRecursive", "could not read object for key {%s, %s}; in file %s",
+                              keyname, keytitle, nextsource->GetName());
+                                 nextsource = (TFile*)sourcelist->After(nextsource);
+                                 return kFALSE;
```
pcanal, line 665:
```suggestion
                           Error("MergeRecursive", "could not read object for key {%s, %s}; in file %s",
                                     keyname, keytitle, nextsource->GetName());
                           nextsource = (TFile*)sourcelist->After(nextsource);
                           return kFALSE;
```
same in 3 other similar new code fragment.

pcanal (approved):
LGTM

silverweed:
@pcanal so we're ok changing the default behavior and not introducing the new flag to preserve the old one?

pcanal:
Yes.  However we should indeed announced it in the Releases Notes.  My thinking is still/more:

> I am leaning towards agreeing with this PR as a present-but-broken object is still "meant" to be there and thus is 'surprising' to be not used in the result.

io/io/src/TFileMerger.cxx
```diff
@@ -735,10 +743,18 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type,
...
                      if (!hobj) {
-                        Info("MergeRecursive", "could not read object for key {%s, %s}; skipping file %s",
+                        switch (fErrBehavior) {
+                        case EErrorBehavior::kFailOnError:
+                           Error("MergeRecursive", "could not read object for key {%s, %s}; in file %s",
+                              keyname, keytitle, nextsource->GetName());
+                           nextsource = (TFile*)sourcelist->After(nextsource);
+                           return kFALSE;
+                        case EErrorBehavior::kSkipOnError:
+                           Info("MergeRecursive", "could not read object for key {%s, %s}; skipping file %s",
```
ferdymercury, line 753: Thanks! Maybe Warning instead of Info ?

pcanal:
Minor note: the clang-format recommendation seems to be correct.

PR #18789, opened 2025-05-19
[ci] Restrict clang-format to PR changes

pcanal:
Related github discussion: https://github.com/orgs/community/discussions/59677

For a given PR, the value of `github.event.pull_request.base.sha` always corresponds to the SHA of the head/tip of the base branch at the moment of either creation of the PR or the last force-push (regardless of whether that force-push actually rebased the PR branch against base or was just modifying an existing commit in it) to the PR branch.

In addition the merge commit is updated whenever a new commit is push onto the PR branch but this merge commit is not against the (fixed) `github.event.pull_request.base.sha` but is instead against the (moving) tip of the target branch.

Currently (before this commit), our clang-format action will check the differences between that commit and the merge commit.

In practice this means that if a PR languish and one or more PR is actually merged into the master branch then the comparison will now involved not only the commit introduced the PR but also all the commits added to master since the PR creation (or its last force push).

There is 2 general solutions to replace the `github.event.pull_request.base.sha`:

1. Find and use the non PR branch parent of the merge commit
2. Find and use the common ancestor of the master and the PR

This commit implement 2nd options by using `git merge-base` with the arguments 'tip of the PR branch' and `github.event.pull_request.base.sha` which is guaranteed to be ahead of/have a commong ancestor with the PR branch.

Until there is a resolution to the above mentioned github discussion this is more complex than before but more stable.

NOTE: This matters in our case since most existing files and some PRs are merge without passing clang-format (in part because clang-format still gives some significantly sub-optimal result in a few important cases)

As a concrete example, the PR https://github.com/root-project/root/pull/18587 was at some point in a state where the merge commit was:
```
*   354e754e771 - (3 hours ago) Merge e54a1eac28e364e3d6130f914e2f2b34786c79b7 into 59c28d595ae433dc7a6b607f859f489bf2859319 — Philippe Canal
```
The tip of master was:
```
* | 59c28d595ae - (5 hours ago) [nfc] mention deprecated ConcatFileName (#18755) — ferdymercury
```
but the `github.event.pull_request.base.sha` was pointing to:
```
* | bda0b8490a1 - (3 days ago) [cmake] Check if cplusplus is 23 by default — ferdymercury
```
As seen below this meant that the clang-format stage was reporting about a lot more than what that PR had actually changed.
```
*   354e754e771 - (3 hours ago) Merge e54a1eac28e364e3d6130f914e2f2b34786c79b7 into 59c28d595ae433dc7a6b607f859f489bf2859319 — Philippe Canal
|\
[...]

PR #18801, opened 2025-05-20
[CI] Simplify clang-format check

hageboeck:
An attempt at further simplifying the CI based on Philippe's work in https://github.com/root-project/root/pull/18789.

The `format_script.sh` becomes unnecessary. All information to reproduce is in the workflow logs:
```
Run git clang-format --diff $MERGE_BASE HEAD
  git clang-format --diff $MERGE_BASE HEAD
  shell: /usr/bin/bash -e {0}
  env:
    MERGE_BASE: e9b38023f809f80ef63053ef3cadc04e52f3ee13
diff --git a/tree/dataframe/test/dataframe_cache.cxx b/tree/dataframe/test/dataframe_cache.cxx
index 4c300f7e2..5b3019058 100644
--- a/tree/dataframe/test/dataframe_cache.cxx
+++ b/tree/dataframe/test/dataframe_cache.cxx
@@ -15,7 +15,7 @@ using namespace ROOT::VecOps;
 
 TEST(Cache, FundType)
 {
-      ROOT::RDataFrame tdf(5);
+   ROOT::RDataFrame tdf(5);
    int i = 1;
    auto cached =
       tdf.Define("c0", [&i]() { return i++; }).Define("c1", []() { return 1.; }).Cache<int, double>({"c0", "c1"});
Error: Process completed with exit code 1.
```

## TODOs:
- [x] Remove deliberate formatting errors before merging


.github/workflows/code_analysis.yml
```diff
@@ -18,18 +18,19 @@ jobs:
...
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 1024
+        ref: ${{ github.event.pull_request.head.sha }}
+    - name: Fetch base sha
+      run: git fetch --depth=1024 origin +${{github.event.pull_request.base.sha}}:origin/base_sha
+    - name: Compute merge-base
+      run: echo "MERGE_BASE=$(git merge-base origin/base_sha HEAD)" >> "$GITHUB_ENV"
+    - name: run clang-format
+      run: git clang-format --diff $MERGE_BASE HEAD
```
pcanal, line 33: One active part of the script is missing though:
```
| grep -i -v '.mjs$' | grep -i -v LinkDef | grep -v -E '^D +' | sed -E 's,^.[[:space:]]+,,'
```
which filters:
* `.mjs` : javascript
* LinkDef : clang-format does an horrible job on `#pragma` lines
* '^D +' : avoid deleted files
* `| sed -E 's,^.[[:space:]]+,,'`: also related to delete files (5bc00921db08106e44196585d89615e99b7e7153)

pcanal: On a related note, as seen at https://github.com/root-project/root/actions/runs/15039480948/job/42267590977?pr=18713, the current filter for delete file might not be working well ... 

And similarly, it probably should not be looking at `.rootrc` files.

hageboeck: > * `.mjs` : javascript

This doesn't pose a problem, since clang-format doesn't format this extension.
> * `LinkDef` : clang-format does an horrible job on `#pragma` lines

LinkDef is indeed checked, but here, maybe it would pay off to use `//@clang-format off`. This would have the benefit of not bothering you with suggestions or formatting when you run `git clang-format master` or similar on your machine.

> * `'^D +'` : avoid deleted files
> * `| sed -E 's,^.[[:space:]]+,,'`: also related to delete files ([5bc0092](https://github.com/root-project/root/commit/5bc00921db08106e44196585d89615e99b7e7153))

Deleted files are ignored by `git clang-format`, so this one also isn't a problem. I added a commit to my branch to show the result.

hageboeck: I added one more commit to test `// clang-format off` for LinkDefs, and it works as expected. So maybe the best solution is to add this to all relevant LinkDefs, and then use the default `git clang-format`.

> And similarly, it probably should not be looking at .rootrc files.

These are also ignored by `git clang-format`.

pcanal: Then I wonder which part failed in https://github.com/root-project/root/actions/runs/15039480948/job/42267590977?pr=18713 ?

pcanal: > when you run git clang-format master or similar on your machine.

That would be a "disaster" :) ... at least until at clang-format 21.

>  // clang-format off for LinkDefs, and it works as expected. So maybe the best solution 

This is indeed the clearest ... but of course need a large-ish PR (of its own) to make the update.

hageboeck: > This is indeed the clearest ... but of course needs a large-ish PR (of its own) to make the update.

This is now #18836

hageboeck: @pcanal ready to approve?

pcanal: Mainly yes ... I would like to understood whether the failure seen at https://github.com/root-project/root/pull/18801#discussion_r2100295803 could affect this code (i.e. for that we need to understand/know what really failed there).

hageboeck: > Mainly yes ... I would like to understood whether the failure seen at [#18801 (comment)](https://github.com/root-project/root/pull/18801#discussion_r2100295803) could affect this code (i.e. for that we need to understand/know what really failed there).

The failure in this PR is the script being too complicated:
If I pull Sergey's commit and get a `--name-status` from it, I get:
```
D<tab>roottest/root/io/abstractclass/.rootrc
M<tab>roottest/root/io/abstractclass/CMakeLists.txt
```
The script tries to suppress deleted files with `^D +` (space at least once or more). Therefore, we are explicitly invoking clang-format on a deleted file, and that will fail.

The good news is that all this is unnecessary with git-clang-format, since it automatically ignores deleted files.

vgvassilev:
Note that the script can be run locally and also tells people how to run clang format on multi-commit history so that the formatting changes do not bloat the history. Is this somewhere in the new changes?

hageboeck:
> Note that the script can be run locally and also tells people how to run clang format on multi-commit history so that the formatting changes do not bloat the history. Is this somewhere in the new changes?

I admit that I had never read that far, because I only checked for possible clang-format complaints.
I'm happy to add the instructions how to rebase with clang-format, but I fear that the instructions are brittle. For example, many people don't update `master`, and don't use `origin/master` as remote. Maybe that deserves some thought.

vgvassilev:
> > Note that the script can be run locally and also tells people how to run clang format on multi-commit history so that the formatting changes do not bloat the history. Is this somewhere in the new changes?
> 
> I admit that I had never read that far, because I only checked for possible clang-format complaints. I'm happy to add the instructions how to rebase with clang-format, but I fear that the instructions are brittle. For example, many people don't update `master`, and don't use `origin/master` as remote. Maybe that deserves some thought.

One can take these variables from the PR information and add instructions how to add a remote, too.

PR #18803, opened 2025-05-20
[v636][ntuple] Rename `RBulk` to `RBulkValues`

enirolf:
Backport of #18769


silverweed:
We have an unwritten habit of using `cherry-pick -x` when backporting so the original commit hash is stored in the backported commit's message; could you please add it?

PR #18805, opened 2025-05-20
[RF] Build RooFit without the legacy evaluation backend

guitargeek:
WIP PR to see where we stand.


guitargeek:
It's apparent that the new vectorizing CPU backend is not feature complete, as support for chi-square fits is still missing.

PR #18810, opened 2025-05-20
[v626][ci] Add AlmaLinux 10 to builds

guitargeek:
Add AlmaLinux 10 to 6.26 builds.

Sister PR in roottest:
https://github.com/root-project/roottest/pull/1320


guitargeek:
There is a problem with modules that I won't be able to solve :slightly_smiling_face: Maybe let's just stop with 6.28, to which the backport has been successful

PR #18819, opened 2025-05-21
[df] Add parsing of metadata for distributed RDataFrame 

martamaja10:
Adding the missing part of the functionality of FromSpec (and RDatasetSpec in general) - processing of the metadata information. A number of use cases are being tested - same trees in all samples, different trees in samples, files with multiple tress (when each tree constitutes a separate sample). Adding friend information should work as well (simple test added) but if more convoluted tests are needed, happy to add such.


hageboeck (commented):
Hi, thanks for the work!

I found a few places where one could beautify the assignment of lines of code to commits. It looks like there was a small mistake during a rebase.

And then I want to ask if one would also need to test that event ranges are correct. Of course, the metadata is tested in this PR, but do we also have a test that covers the actual branches, making sure that every event has been read exactly once?
(If it's already there, sorry, I missed it 🙂 )

roottest/python/distrdf/data/ttree/spec_differenttrees.json
```diff
@@ -0,0 +1,31 @@
...
+                "lum": 200.0
+            }
+        },
+        "sample2": {
+            "trees": [
+                "tree"
+            ],
+            "files": [
+                "../data/ttree/distrdf_roottest_check_fromspec.root",
+                "../data/ttree/distrdf_roottest_check_rungraphs.root"
```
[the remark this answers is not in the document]
martamaja10: but I can also add the case in which both trees belong to the same sample

tree/dataframe/inc/ROOT/RDF/RMetaData.hxx
```diff
@@ -71,6 +71,9 @@ public:
    double GetD(const std::string &key, double defaultVal) const;
    const std::string GetS(const std::string &key, const std::string &defaultVal) const;
 
+   void ImportJSON(const std::string &jsonString); // move to private and provide friend class
+   std::string ExportJSON() const; // move to private and provide friend class
+
```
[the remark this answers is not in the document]
martamaja10: I think we still want to move it to the private namespace (but yes, actually comments stayed there because I forgot about these two functions)

roottest/python/distrdf/data/ttree/spec_differenttrees.json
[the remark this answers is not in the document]
hageboeck: That should indeed be sufficient, because the trees are part of the same distributed run. No need to make an extra sample.

hageboeck (approved):
Looks good to me after these items:
- Move the last commit to the front? It looks like the current first commit uses it.
- Ruff had some complaints
- And as we discussed, a test that computes some kind of checksum over a branch in a file that actually changes would test that the partitioning is correct. (Maybe this is already part of a test that wasn't touched in this PR. In that case, disregard this item)

roottest/python/distrdf/backends/check_fromspec.py
```diff
@@ -1,27 +1,85 @@
...
+        jsonfile_three = "../data/ttree/spec_withfriends.json"
+        rdf_friends = ROOT.RDF.Experimental.FromSpec(jsonfile_three, executor = connection)
+        
+        rdf_friends_new = rdf_friends.DefinePerSample("lumi", 'rdfsampleinfo_.GetD("lumi")')
+        
+        rdf_friends_filtered = rdf_friends_new.Filter("lumi == 1.")
+        rdf_friends_filtered_two = rdf_friends_new.Filter("lumi == 0.5")
+    
+        assert rdf_friends_filtered.Count().GetValue() == 4
+        assert rdf_friends_filtered_two.Count().GetValue() == 1
```
vepadulano, line 81: Maybe I would add to this test some check on the friend trees values

tree/dataframe/inc/ROOT/RDF/RMetaData.hxx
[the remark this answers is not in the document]
vepadulano: I agree I would move the two functions to Internal namespace before merging. Although, these two look like they could be moved to the public interface at some point, I just wouldn't do it in the same PR.

bindings/distrdf/python/DistRDF/HeadNode.py
```diff
@@ -636,23 +636,40 @@ def __init__(self, backend: BaseBackend, npartitions: Optional[int], localdf: RO
...
+        
+        # A map to map file id (filename/tree) to SerializableRSamples
+        self.sampleMap = {}
+        
+        for sample in self.samples:
+            trees = sample.GetTreeNames()
+            files = sample.GetFileNameGlobs()
+            for i in range (len(files)):
+                sampleId = files[i] + "/" + trees[i]
+                self.sampleMap.update({sampleId : Ranges.SerializableRSample(sample)})
```
vepadulano, line 672: Since we're anyway just storing `self.sampleMap`, shouldn't we remove it `self.samples` and just use it as a local variable? In that case, it wouldn't even be necessary to convert it to a list, just the vector returned from `MoveOutSamples` should be good.

bindings/distrdf/python/DistRDF/HeadNode.py
```diff
@@ -715,33 +731,69 @@ def attach_friend_info_if_present(current_range: Ranges.TreeRange,
...
-            ds.AddSample(
-                ("", clustered_range.treenames, clustered_range.filenames))
+        
+            filenames_cluster = clustered_range.filenames
+            treenames_cluster = clustered_range.treenames
+            
+            treesInFileMap = {}
+            fileids_cluster = []
+                        
+            for i in range(len(filenames_cluster)):
```
vepadulano, line 757: This would be more Pythonic
```suggestion
            for filename, treename in zip(filenames_cluster, treenames_cluster):
```

bindings/distrdf/python/DistRDF/HeadNode.py
```diff
@@ -715,33 +731,69 @@ def attach_friend_info_if_present(current_range: Ranges.TreeRange,
...
-            # add a sample with no name to represent the whole dataset
-            ds.AddSample(
-                ("", clustered_range.treenames, clustered_range.filenames))
+        
+            filenames_cluster = clustered_range.filenames
+            treenames_cluster = clustered_range.treenames
+            
+            treesInFileMap = {}
+            fileids_cluster = []
+
```
vepadulano, line 756: Here I would add a few lines of comments, explaining the logic below that reconnects the files and trees of this task back to the correct sample from the spec

bindings/distrdf/python/DistRDF/HeadNode.py
```diff
@@ -715,33 +731,69 @@ def attach_friend_info_if_present(current_range: Ranges.TreeRange,
...
-                ("", clustered_range.treenames, clustered_range.filenames))
+        
+            filenames_cluster = clustered_range.filenames
+            treenames_cluster = clustered_range.treenames
+            
+            treesInFileMap = {}
+            fileids_cluster = []
+                        
+            for i in range(len(filenames_cluster)):
+                fileid = filenames_cluster[i] + "/" + treenames_cluster[i] # needed for cases of more than one tree in a rootfile
```
vepadulano, line 758: > needed for cases of more than one tree in a rootfile

Isn't the full ID (filename + / + treename) always needed?

bindings/distrdf/python/DistRDF/HeadNode.py
```diff
@@ -715,33 +731,69 @@ def attach_friend_info_if_present(current_range: Ranges.TreeRange,
...
+                fileids_cluster.append(fileid)
+                treesInFileMap.update({fileid : treenames_cluster[i]})
+                        
+            unique_samples = [] 
+            samplenames = []
+            
+            for sample in clustered_range.samples:
+                mysample = pickle.loads(pickle.dumps(sample))._sample
```
vepadulano, line 766: If the `sample` variable has made it this far, it means it was already serialized (from the local user process) and deserialized (in the current distributed worker process)
```suggestion
                mysample = sample._sample
```

bindings/distrdf/python/DistRDF/HeadNode.py
```diff
@@ -715,33 +731,69 @@ def attach_friend_info_if_present(current_range: Ranges.TreeRange,
...
+                samplename = mysample.GetSampleName()
+                if samplename not in samplenames:
+                    unique_samples.append(mysample)
+                    samplenames.append(samplename)
+                    
+            for mysample in unique_samples:
+                sample_fileids = []
+                for i in range(len(mysample.GetFileNameGlobs())):
+                    mysample_fileid = mysample.GetFileNameGlobs()[i] + "/" + mysample.GetTreeNames()[i]
+                    sample_fileids.append(mysample_fileid)
```
vepadulano, line 776:
```suggestion
                sample_fileids = [
                    filenameglob + "/" + treename
                    for filenameglob, treename in zip(mysample.GetFileNameGlobs(), mysample.GetTreeNames())
                ]
```

bindings/distrdf/python/DistRDF/HeadNode.py
```diff
@@ -715,33 +731,69 @@ def attach_friend_info_if_present(current_range: Ranges.TreeRange,
...
+                    
+                good_files = list((Counter(filenames_cluster) & Counter(mysample.GetFileNameGlobs())).elements())
+                good_fileids = list((Counter(fileids_cluster) & Counter(sample_fileids)).elements()) # to match treenames correctly 
+
+                good_trees = []
+                for fileid in good_fileids:
+                    goodtree = treesInFileMap.get(fileid)
+                    good_trees.append(goodtree)
```
vepadulano, line 784:
```suggestion
good_trees =[
    treesInFileMap.get(fileid) for fileid in good_fileids
]
```

bindings/distrdf/python/DistRDF/Ranges.py
```diff
@@ -407,3 +486,182 @@ def get_clustered_range_from_percs(percrange: TreeRangePerc) -> Tuple[Optional[T
                           globalstart, globalend, percrange.friendinfo)
 
     return treerange, entries_in_trees
+
+def get_percentage_ranges_sample(treenames: List[str], filenames: List[str], npartitions: int,
+                          sampleMap: Dict[str, SerializableRSample], 
+                          friendinfo: Optional[ROOT.Internal.TreeUtils.RFriendInfo],
+                          exec_id: ExecutionIdentifier) -> List[TreeRangeSamplePerc]:
```
vepadulano, line 493: I would try to think about whether this function and the next could be simplified, or removed. I could be missing something, but from what I can see right now, they are fairly similar if not practically the same of their counterparts without the `_sample` part of the name, except for the fact that we need to pass the `sample` down to the new `TreeRangeSamplePerc` or `TreeRangeSample`.

As an idea, I can imagine we could add `samples: Optional[List[SerializableRSample]]` to the already existing `TreeRangePerc` and `TreeRange`, then optionally inject the sample infomation in `get_percentage_ranges`.

vepadulano (changes requested):
Thank you @martamaja10 , this is great! Some comments for consideration

bindings/distrdf/python/DistRDF/HeadNode.py
```diff
@@ -729,19 +744,66 @@ def build_rdf_from_range(current_range: Ranges.TreeRangePerc) -> TaskObjects:
...
+            # Making sure we don't double count samples  
+            unique_samples = [] 
+            samplenames = []
+            
+            for sample in clustered_range.samples:
+                mysample = sample._sample
+                samplename = mysample.GetSampleName()
+                if samplename not in samplenames:
+                    unique_samples.append(mysample)
+                    samplenames.append(samplename)
```
vepadulano, line 780: I don't see the `samplenames` list being used anywhere else, is it necessary?

bindings/distrdf/python/DistRDF/Ranges.py
```diff
@@ -238,30 +248,69 @@ def get_percentage_ranges(treenames: List[str], filenames: List[str], npartition
...
+                first_tree_start_perc_tasks[rangeid], last_tree_end_perc_tasks[rangeid], friendinfo, samples)
             for rangeid in range(npartitions)
-        ]
+            ]
+
+        else:   
+            return [
+                TreeRangePerc(
+                    exec_id, rangeid, treenames, filenames, start_sample_idxs[rangeid], end_sample_idxs[rangeid],
+                    first_tree_start_perc_tasks[rangeid], last_tree_end_perc_tasks[rangeid], friendinfo, [])
```
vepadulano, line 270: This would be more semantically correct
```suggestion
                    first_tree_start_perc_tasks[rangeid], last_tree_end_perc_tasks[rangeid], friendinfo, None)
```

bindings/distrdf/python/DistRDF/Ranges.py
```diff
@@ -238,30 +248,69 @@ def get_percentage_ranges(treenames: List[str], filenames: List[str], npartition
...
-                first_tree_start_perc_tasks[rangeid], last_tree_end_perc_tasks[rangeid], friendinfo
-            )
-            for rangeid in range(npartitions)
-        ]
+        else:
+        
+            return [
+                TreeRangePerc(
+                    exec_id, rangeid, tasktreenames[rangeid], taskfilenames[rangeid], 0, len(taskfilenames[rangeid]),
+                    first_tree_start_perc_tasks[rangeid], last_tree_end_perc_tasks[rangeid], friendinfo, []
```
vepadulano, line 310:
```suggestion
                    first_tree_start_perc_tasks[rangeid], last_tree_end_perc_tasks[rangeid], friendinfo,None
```

vepadulano (changes requested):
Thank you! Very minor changes before merging.

Issue #18816, opened 2025-05-21
[Minuit2] Run time switch for multithreading in minuit2


SilentAssassinMa:
### Feature description

In current version of minuit2, it's possible to build with openMP support (with warnings). However if such a flag is turned on, the minuit2 will run in the multithread mode by default and the only way to switch that off is setting the environment variable `OMP_NUM_THREADS` to 1, which is not an ideal solution.

The only part uses the openMP parallel is when calculating the gradient in [Numerical2PGradientCalculator.cxx](https://github.com/root-project/root/blob/master/math/minuit2/src/Numerical2PGradientCalculator.cxx#L100) line 100-104:
```c++
#pragma omp parallel
#pragma omp for
   //#pragma omp for schedule (static, N_PARALLEL_PAR)

   for (int i = 0; i < int(n); i++) {
```
which should be easy change the omp part to 
```c++
#pragma omp parallel for if(dopara)
```
where `dopara` is a runtime bool variable which can be set from outside. Such feature can be added in the parent class [`GradientCalculator`](https://github.com/root-project/root/blob/master/math/minuit2/inc/Minuit2/GradientCalculator.h):
```c++
public:
    static void DoParallel(bool doParallel = true) { fDoParallel = doParallel; }
protected:
    static inline bool fDoParallel = false; // flag to indicate if parallel processing is used
```
In this way, the multithreading is switched off by default and can be manually switched on by the users in their programs. This will also helps to resolve [this issue](https://github.com/root-project/root/issues/12940).

### Additional context

I just did a local test by adding such a function to control the multithreading, and it worked as expected. However I'm not familar with the workflow on how to merge such changes with the master.

PR #18820, opened 2025-05-21
[minuit2] disable parallelization by default even if openMP is active, allow switch

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/18816

Credit for the solution goes to @SilentAssassinMa

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/README/ReleaseNotes/v638/index.md
+++ b/README/ReleaseNotes/v638/index.md
@@ -4,6 +4,10 @@
 * The `RooStats::HLFactory` class that was deprecated in ROOT 6.36 is now removed. It provided little advantage over using the RooWorkspace directly or any of the other higher-level frameworks that exist in the RooFit ecosystem.
 * The build options `mysql`, `odbc` and `pgsql`, that were deprecated in ROOT 6.36, are now removed.
 
+## Minuit2
+
+* Behavior change: building ROOT using `minuit2_omp=ON` option no longer enables OpenMP parallelization by default. One has to call now additionaly GradientCalculator::DoParallelOMP().
+
 ## RooFit
 
 ### Error out when setting out-of-range variable value instead of silent clipping
--- a/README/ReleaseNotes/v638/index.md
+++ b/README/ReleaseNotes/v638/index.md
@@ -4,7 +4,9 @@
 * The `RooStats::HLFactory` class that was deprecated in ROOT 6.36 is now removed. It provided little advantage over using the RooWorkspace directly or any of the other higher-level frameworks that exist in the RooFit ecosystem.
 * The build options `mysql`, `odbc` and `pgsql`, that were deprecated in ROOT 6.36, are now removed.
 
-## Minuit2
+## Math
+
+### Minuit2
 
 * Behavior change: building ROOT using `minuit2_omp=ON` option no longer enables OpenMP parallelization by default. One has to call now additionaly GradientCalculator::DoParallelOMP().
 
--- a/README/ReleaseNotes/v638/index.md
+++ b/README/ReleaseNotes/v638/index.md
@@ -8,7 +8,7 @@
 
 ### Minuit2
 
-* Behavior change: building ROOT using `minuit2_omp=ON` option no longer enables OpenMP parallelization by default. One has to call now additionaly GradientCalculator::DoParallelOMP().
+* Behavior change: building ROOT using `minuit2_omp=ON` option no longer enables OpenMP parallelization by default. One has to call now additionaly GradientCalculator::SetParallelOMP().
 
 ## RooFit
 
--- a/math/minuit2/inc/Minuit2/GradientCalculator.h
+++ b/math/minuit2/inc/Minuit2/GradientCalculator.h
@@ -35,7 +35,7 @@ public:
 
    virtual bool G2(const MinimumParameters &, MnAlgebraicVector &) const { return false;}
 
-   static bool DoParallelOMP(bool doParallel = true);
+   static bool SetParallelOMP(bool doParallel = true);
 
 protected:
    static inline bool fDoParallelOMP = false; ///< flag to indicate if parallel OpenMP processing is used
--- a/math/minuit2/inc/Minuit2/GradientCalculator.h
+++ b/math/minuit2/inc/Minuit2/GradientCalculator.h
@@ -37,13 +37,14 @@ public:
 
    /**
     * Enable parallelization of gradient calculation using OpenMP.
+    * This is different from the default parallel mechanism elsewhere (IMT, threads, TBB, ...)
     * \param doParallel true to enable, false to disable
     * \note If OPENMP is not available, this function has no effect
     */
-   static void DoParallel(bool doParallel = true) { fDoParallel = doParallel; }
+   static void DoParallelOMP(bool doParallel = true) { fDoParallelOMP = doParallel; }
 
 protected:
-   static inline bool fDoParallel = false; ///< flag to indicate if parallel processing is used
+   static inline bool fDoParallelOMP = false; ///< flag to indicate if parallel OpenMP processing is used
 };
 
 } // namespace Minuit2
--- a/math/minuit2/inc/Minuit2/GradientCalculator.h
+++ b/math/minuit2/inc/Minuit2/GradientCalculator.h
@@ -37,7 +37,8 @@ public:
 
    /**
     * Enable parallelization of gradient calculation using OpenMP.
-    * This is different from the default parallel mechanism elsewhere (IMT, threads, TBB, ...)
+    * This is different from the default parallel mechanism elsewhere (IMT, threads, TBB, ...).
+    * It can only be used to minimise thread-safe functions in Minuit2.
     * \param doParallel true to enable, false to disable
     * \note If OPENMP is not available, this function has no effect
     */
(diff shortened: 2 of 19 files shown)


guitargeek:
This is great! I was thinking about it years ago, but there was no strong reason to actually do it, lacking concrete user requests.

But if the parallelization can be toggled, what is the point of actually building `minuit2` without omp support inside ROOT?

I would suggest the following:

* change your new flag such that it's `false` by default
* set the `minuit2_omp=true` by default in the ROOT builds
* if the CI shows that it works well on all platforms, we deprecate the `minuit2_omp` flag and always go for building Minuit 2 with OpenMP support

Does that make sense to you?

ferdymercury:
Thanks! What about the warning:

`      message(WARNING "The option '${opt}' can only be used to minimise thread-safe functions in Minuit2. It cannot be used for Histogram/Graph fitting and for RooFit. If you want to use Minuit2 with OpenMP or MPI support, it is better to build Minuit2 as a standalone library.")`

pcanal:
> and always go for building Minuit 2 with OpenMP support

`OpenMP` migth not always be available :)

guitargeek:
That's true, as confirmed by the CI. It was still interesting to try out and see where it is available!

Okay, so since it's clearly not available everywhere, we need to keep the option. But setting it to `true` by default in the `global` config, and only to `false` on macOS is something we can do. The warning can be disabled for the case of `minuit2_omp` because if the parallelization is not enabled by default for builds with `minuit2_omp`, there is no problem

math/minuit2/inc/Minuit2/GradientCalculator.h
```diff
@@ -35,6 +35,15 @@ class GradientCalculator {
 
    virtual bool G2(const MinimumParameters &, MnAlgebraicVector &) const { return false;}
 
+   /**
+    * Enable parallelization of gradient calculation using OpenMP.
+    * \param doParallel true to enable, false to disable
+    * \note If OPENMP is not available, this function has no effect
+    */
+   static void DoParallel(bool doParallel = true) { fDoParallel = doParallel; }
```
pcanal, line 43: Should this indicate in the signature that it is OMP/process as opposed to the default parallel mechanism elsewhere (i.e. usually threads/TBB)?

ferdymercury:
> Pretty nice that it seems to be already available on Windows though!

Nope, it does not, since the Windows build-config has an explicit minuit2_omp=OFF

@bellenot do you think I could remove this flag?

math/minuit2/src/Numerical2PGradientCalculator.cxx
```diff
@@ -97,8 +97,7 @@ operator()(const MinimumParameters &par, const FunctionGradient &Gradient) const
 
    // parallelize this loop using OpenMP
 //#define N_PARALLEL_PAR 5
-#pragma omp parallel
-#pragma omp for
+#pragma omp parallel for if (fDoParallelOMP)
    //#pragma omp for schedule (static, N_PARALLEL_PAR)
 
    for (int i = 0; i < int(n); i++) {
```
ferdymercury, line 103:
```suggestion
   for (unsigned int i = 0; i < n; i++) {
```

math/minuit2/inc/Minuit2/GradientCalculator.h
```diff
@@ -35,6 +35,17 @@ class GradientCalculator {
...
    virtual bool G2(const MinimumParameters &, MnAlgebraicVector &) const { return false;}
 
+   /**
+    * Enable parallelization of gradient calculation using OpenMP.
+    * This is different from the default parallel mechanism elsewhere (IMT, threads, TBB, ...).
+    * It can only be used to minimise thread-safe functions in Minuit2.
+    * \param doParallel true to enable, false to disable
+    * \note If OPENMP is not available, this function has no effect
+    */
+   static void DoParallelOMP(bool doParallel = true) { fDoParallelOMP = doParallel; }
```
guitargeek, line 45: I can already see the reports by people who try `DoParallelOMP()`, and then complain because it's not faster because they didn't make sure that Minuit 2 was built with OpenMP support :slightly_smiling_face:

Maybe we can move the implementation of `DoParallelOMP` in the `.cxx` file, and then implement it like this, with an exception?

```c++
void GradientCalculator::DoParallelOMP(bool doParallel) {
#ifdef _OPENMP
   fDoParallelOMP = doParallel; 
#else
   if (doParallel) {
      throw std::runtime_error("Minuit 2 was built without OpenMP support! Can't enable parallel gradients with OpenMP");
   }
#endif
}
```

math/minuit2/inc/Minuit2/GradientCalculator.h
```diff
@@ -35,6 +35,17 @@ class GradientCalculator {
...
+
+protected:
```
guitargeek, line 47:
```suggestion
private:
```
I don't think this needs to be protected

SilentAssassinMa: I made that protected because this variable need to be accessible to the child class `Numerical2PGradientCalculator` where the omp is actually implemented.

guitargeek: Ah yes! Thanks.

math/minuit2/inc/Minuit2/GradientCalculator.h
ferdymercury: There is no .cxx file, should I create a new one?

ferdymercury: On the other hand, this could be problematic, imagine running the same script on two different machines, one with, another without. Maybe a warning rather than an exception is less intrusive?

SilentAssassinMa: I mean, maybe the parent class is not the best place for this functionality. I chose this class just because it was convienent for the local test.

guitargeek: That's true. @SilentAssassinMa, would it be okay for you if me move this into the `Numerical2PGradientCalculator` class? Because indeed, as it is right now it's not very precise. There is also the `HessianGradientCalculator` that inherits from the same base class, and it doesn't support OpenMP for example.

guitargeek: @ferdymercury, that's a fair point with the exception. What about at least an error, like:
```c++
      MnPrint print("Numerical2PGradientCalculator");
      MnPrint.Error(...);
```
?

SilentAssassinMa: For me it's okay. However we also have to consider the user friendly of this feature, as `Numerical2PGradientCalculator.h` is not included by any other header files as far as I see, so one has to manually inclued this to activate the parallelization. Or we can update the documentation to instruct the user about this feature.

pcanal: And the very least the routine should return a bool (i.e. `false` it did not do what it said it would :) )

ferdymercury:
In Linux unit tests, should we enable somewhere a call to DoParallelOMP? Otherwise we do not really test it.

guitargeek:
Yes, sure it would be good to have a test for the parallel gradients! But it's not really the scope of this PR and the GitHub issue, so we might also do it later. The only reason why this PR is still not merged is because I still want to do some profiling in big fits. I want to convince myself that `#pragma omp parallel for if (fDoParallelOMP)` really has no overhead :slightly_smiling_face:

README/ReleaseNotes/v638/index.md
```diff
@@ -4,6 +4,10 @@
 * The `RooStats::HLFactory` class that was deprecated in ROOT 6.36 is now removed. It provided little advantage over using the RooWorkspace directly or any of the other higher-level frameworks that exist in the RooFit ecosystem.
 * The build options `mysql`, `odbc` and `pgsql`, that were deprecated in ROOT 6.36, are now removed.
 
+## Minuit2
```
guitargeek, line 7: Like in the [6.36 release notes](https://root.cern/doc/v636/release-notes.html#math), Minuit 2 related developments should be a sub-section of Math
```suggestion
## Math

### Minuit2
```

math/minuit2/inc/Minuit2/GradientCalculator.h
```diff
@@ -35,6 +35,10 @@ class GradientCalculator {
 
    virtual bool G2(const MinimumParameters &, MnAlgebraicVector &) const { return false;}
 
+   static bool DoParallelOMP(bool doParallel = true);
```
guitargeek, line 38:
```suggestion
   static bool SetParallelOMP(bool doParallel = true);
```
...or `EnableParallelOMP()` maybe...

The problem with the `Do*` prefix is that it's not clear from the name whether this is a setter or a getter, and more importantly this name breaks a convention that is everywhere in the math code: functions starting with `DoFoo` are usually implementation details that are called by the public `Foo` function. For example here in the fitter:

https://github.com/root-project/root/blob/master/math/mathcore/inc/Fit/Fitter.h#L174

So maybe it's not too late to change the name?

guitargeek (approved):
Thank you very much! The performance is unaffected, even for the ATLAS Higgs combination benchmark.

PR #18832, opened 2025-05-21
[RDF] fix typo in called test function

ferdymercury:
Since title of test and called function do not match. It seems a copy-paste typo from the function just above?


vepadulano (approved):
Nice catch, thanks!

PR #18852, opened 2025-05-23
[tree] fix duplicated anchor in filename

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-5002

## Checklist:

- [x] tested changes locally


roottest/root/tree/entrylist/execTEntryListZip.cxx
```diff
@@ -0,0 +1,24 @@
...
+// First run: root $ROOTSYS/tutorials/tree/cernbuild.C
+// Then run: zip -n root cernstaff cernstaff.root
+
+int execTEntryListZip()
+{
+    TChain chain("T");
+    chain.Add("cernstaff.zip#cernstaff.root");
+    chain.Draw(">> elist0", "Age > 40", "entrylist");
+
+    gDirectory->ls();
```
ferdymercury, line 14:
```suggestion
    // gDirectory->ls();
```

roottest/root/tree/entrylist/execTEntryListZip.cxx
```diff
@@ -0,0 +1,24 @@
...
+// Then run: zip -n root cernstaff cernstaff.root
+
+int execTEntryListZip()
+{
+   TChain chain("T");
+   chain.Add("cernstaff.zip#cernstaff.root");
+   chain.Draw(">> elist0", "Age > 40", "entrylist");
+
+   // gDirectory->ls();
+   auto el = static_cast<TEntryList*>(gDirectory->Get("elist0"));
```
ferdymercury, line 15:
```suggestion
   auto el = static_cast<TEntryList *>(gDirectory->Get("elist0"));
```

roottest/root/tree/entrylist/execTEntryListZip.ref
```diff
@@ -0,0 +1,3 @@
+
+Processing execTEntryListZip.cxx+...
+(int)0
```
ferdymercury, line 3:
```suggestion
(int) 0
```

vepadulano:
Apologies, I'm trying to understand better what this PR is solving. I can't see from the original issue or from the new reproducer what was the previously broken situation, e.g. what was the file name? In the original issue I read
```
The method TEntryList::GetEntryList called inside TChain::SetEntryList doesn't find any match between with the TChain 
```
Does the match happen via the file name?

ferdymercury:
Hi, if I recall correctly, the problem was:

`chain.Add("cernstaff.zip#cernstaff.root");`

which later led to in TEntryList to:

```
filename = tree->GetTree()->GetCurrentFile()->GetName();
TUrl url(filename.Data(), true);
```
url copies the "hash part" as anchor as an internal variable of the class.

but then, we also call:

`url.SetFile(filename);`

where filename still contains the hash part.

so we ended up with

`filename = url.GetUrl();`

which appends the previous filename with the internal anchor, so we end up having two times the hash part.

so `cernstaff.zip#filename.root#filename.root`

Which means that then, the TEntryList does find the "filename-filename" file inside that zip container.

dpiparo:
revamping this nice PR.

tree/tree/src/TEntryList.cxx
```diff
@@ -1317,6 +1317,7 @@ void TEntryList::SetTree(const TTree *tree)
          if (!gSystem->IsAbsoluteFileName(filename))
             gSystem->PrependPathName(gSystem->pwd(), filename);
          filename = gSystem->UnixPathName(filename);
+         url.SetAnchor(""); // Avoid potentially duplicated # when calling SetFile below
```
pcanal: This fell oddly place.  Did we investigate whether this actually belongs inside `SetFile` per se (i.e. which use case  is `SetFile` appending the anchors serving?)

ferdymercury: Wouldn't that potentially break preexisting user scripts?
```
url.SetAnchor(...);
url.SetFile(...); // Erases the anchor set above?
```

pcanal: So the question is whether:
```
url.SetAnchor(...);
url.SetFile(...);
```
and
```
url.SetFile(...);
url.SetAnchor(...);
```
are meant to be semantically the same. (And in addition whether we want to deal the backward incompatibility if not). 

For my own clarity, the issue at hand is that we have in this code:
```
url.SetFile(filename); // file name with possibly some anchor 
if (some_condition) {
   // tweak the filename
   url.SetFile(new_filename); 
}
```

and you observation is that in this case the anchor are additive rather than being a substitution,right?

ferdymercury: Not really, what we have is the following within TEntryList.cxx:

```
TUrl url("cernstaff.zip#cernstaff.root", true);
// This internally calls SetUrl, which calls FindFile, which sets fFile = "cernstaff.zip" and fAnchor = "cernstaff.root"
// But then:
if (url.GetProtocol() is file)
   ExpandPathName("cernstaff.zip#cernstaff.root");
   url.SetFile(/full/path/cernstaff.zip#cernstaff.root"); 
    // This modifies fFile and sets it as fFile = /full/path/cernstaff.zip#cernstaff.root but leaves fAnchor untouched ie it's still cernstaff.root.
```

So the possible solutions are:
- set Anchor=""
- call SetUrl instead of SetFile
- ExpandPathName of fFile instead of the full fUrl

I will change to the 3rd option since it looks the cleanest.

ferdymercury:
@vepadulano one Windows still failing but I guess it's unrelated.

PR #18858, opened 2025-05-23
[df] Add missing dict dependency for datasource_tree

pcanal:
Fix runtime failures (seen on Debian notes):
```
<<< cling interactive line includer >>>: fatal error: module file '/github/home/ROOT-CI/build/lib/RIO.pcm' is out of date and needs to be rebuilt: module file out of date
it means to be say
```
but this is a lie, it is meant to say:
```
<<< cling interactive line includer >>>: fatal error: module file '/github/home/ROOT-CI/build/tree/dataframe/test/datasource_tree.pcm' is out of date and needs to be rebuilt: module file out of date as it depends on the new file '/github/home/ROOT-CI/build/lib/RIO.pcm'
```


silverweed:
Why is cling lying? Can we fix it?

vepadulano (approved):

PR #18671, opened 2025-05-09
[cppyy] Add CMakeLists.txt for cppyy/test

Vipul-Cariappa:
Enabling running tests using `ctest`

PR #18867, opened 2025-05-26
[cppyy] Use ROOT dictionary generation macro in cppyy tests

guitargeek:
This fixes build errors with `builtin_clang=OFF` that appeared after https://github.com/root-project/root/pull/18671.


aaronj0 (approved):
LGTM!

Issue #18861, opened 2025-05-25
Turn soversion ON by default


lmoureaux:
### Explain what you would like to see improved and how.

ROOT binary distributions don't always enable the `soversion` option even though it has always been recommended.
https://github.com/root-project/root/blob/15f0275211ba50d36449d0be82d8f470dba834a1/cmake/modules/RootBuildOptions.cmake#L161

For instance, Conda enables it explicitly and LCG doesn't.

Since this option is recommended and can help prevent subtle bugs caused by version mismatches, it should be enabled by default. Versioning libraries is in any case good practice.

### ROOT version

ROOT 6.34.04 from `/cvmfs/sft.cern.ch/lcg/releases/ROOT/6.34.04-2bf94/x86_64-el9-gcc14-opt/`.
ROOT 6.32.02 from conda-forge.
Own build with last shared commit at cd805272f42e2992da10a1e204afafbfa64cd411.

### Installation method

conda, LCG, source build

### Operating system

Linux (all)

dpiparo:
thanks @lmoureaux for this report.

dpiparo:
@andresailer

lmoureaux:
Thanks for the quick action! Wouldn't it be better to change the default in CMake though?

ferdymercury:
Maybe related: 09501f794eda08886db9b48c4353f42a36d3aade

guitargeek:
> Since this option is recommended and can help prevent subtle bugs caused by version mismatches, it should be enabled by default. Versioning libraries is in any case good practice.

Thanks to discussing with @bellenot, I understand now that the "recommendation" was misleading, because as `soversion=ON` is implemented now, version mismatches would still happen in case on multiple installed ROOT versions, because the rootmap files don't follow the same versioning pattern. So `soversion=ON` can actually result in more intransparent errors in case of version mismatches, that can be even less helpful to the user than duplicate filenames. We therefore suggest to address this issue by un-recommending this flag:

https://github.com/root-project/root/pull/18876

Maybe we can one day recommend it if the rootmap files etc. follow the same versioning convention too as well, ensuring consistency. But for now there is not a usecase that motivates working on this.

bellenot:
Right, and AFAIK, the `.pcm` files are also not versioned (if that matters at all)...

lmoureaux:
Sad. I guess this is the same class of issues that forces the use of LD_LIBRARY_PATH over RUNPATH everywhere.

guitargeek:
[...]

PR #18870, opened 2025-05-26
[cmake] enable soversion by default

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/18861


bellenot:
I don't know why it was off by default... I'll check and maybe we'll have to discuss the impact that can have

cholmcc:
> Relevant commit on "why" it was disabled 25 years ago by @cholmcc:
> https://github.com/root-project/root/commit/d1263cb4680e9d96d0047a9a5177a3daf6dddb6b
> ```This option is only supported on some platforms. Default is false.```
> https://github.com/root-project/root/commit/8fe865546e3900231c79b91343d174f0a748a610

Those comments related to the old (hand-crafted, pre-CMake) build system, and may or may not be relevant to the CMake build system.   Presumably, CMake knows how to handle this on various platforms  - if it doesn't, then what is the point of CMake? :smile: 

Yours,
_Christian_

Issue #16805, opened 2024-10-31
A TChain whose trees have friend TChains stops updating


lmoureaux:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

Consider the following tree loading structure:
```
TChain c2 ---> TTree ---(friend)---> TChain c1 ---> TTree
```
When iterating over `c2`, branches coming from `c1` stop updating after the first file boundary is reached in `c1`. They start updating again once a file boundary is reached in `c2`.

I think this is distinct from #16804 because in this case the issues lie in `TChain::LoadTree`, while for #16804 `TChain::LoadTree` is the sole culprit.

As in #16804, the root cause of the issue is that the following code is not invoked often enough:
https://github.com/root-project/root/blob/2cc4b70d2e2bd89d9b639c894b4e0558be47c22b/tree/tree/src/TChain.cxx#L1407-L1434
I will call the operation above an "address refresh".

TChain has code triggering an address refresh when needed for friends of the current tree:
https://github.com/root-project/root/blob/2cc4b70d2e2bd89d9b639c894b4e0558be47c22b/tree/tree/src/TChain.cxx#L1384-L1393

Unfortunately this doesn't work for two reasons:
1. The code cannot possibly be reached unless `c2` itself has friends because of the following condition:
   https://github.com/root-project/root/blob/2cc4b70d2e2bd89d9b639c894b4e0558be47c22b/tree/tree/src/TChain.cxx#L1370
  In our case `c2` doesn't have friends, but the `fTree` does. This is easy to fix (or work around).
2. Even if `fFriends` is made non-null, `fetree->IsUpdated()` always returns `false` because `TTree::LoadTree` resets it:
   https://github.com/root-project/root/blob/18b4f317b389fa7931d7eb8e00525f4dca44be0f/tree/tree/src/TTree.cxx#L6512
  Fixing this is much more annoying. For [my application](https://gitlab.cern.ch/Proto/Darwin/-/merge_requests/44), I ended up subscribing to the `Notify` signal and calling `TFriendElement::MarkUpdated` again. This is only a partial solution because when `c2` loads a new file, it doesn't reset friend elements that were marked updated, which would happen somewhere along this line:
  https://github.com/root-project/root/blob/2cc4b70d2e2bd89d9b639c894b4e0558be47c22b/tree/tree/src/TChain.cxx#L1666

### Reproducer

[...]

PR #18871, opened 2025-05-26
[tree] Delegate SetBranchAddress to friends if necessary

### LINKED ISSUE #16804: TChain as a friend of TTree gets stuck after the first file
### Check duplicate issues.

- [X] Checked for duplicates

### Description

When a `TChain` is added as a friend to a `TTree` , branches coming from the `TChain` are no longer updated after switching to the second file in the chain.

### Reproducer

The [attached macro](https://github.com/user-attachments/files/17592119/bugDemo.txt) reproduces the issue. Save it `bugDemo.C` and run it. The correct behavior is to always print `true`, but we get some `false`.

The code first constructs two trees with an `index` branch. The index is the entry number, going from 0 to 10 in the first file and 10 to 20 in the second one. Then, a longer tree is constructed with 20 empty entries. The two files containing the entry number as merged into a chain and attached as a friend to the empty tree.

Iterating over the long tree yields the correct value of `index` until the 10th entry, than a constant 9 (the last value in the first file). Correct behavior is 0 to 20. This can be obtained by wrapping the long tree in a `TChain`.

The underlying issue is that `TTree::LoadTree` lacks the following code present in `TChain::LoadTree`:
https://github.com/root-project/root/blob/2cc4b70d2e2bd89d9b639c894b4e0558be47c22b/tree/tree/src/TChain.cxx#L1407-L1434

### ROOT version

Seen in:
* ROOT 6.30/04 (anaconda)
* ROOT 6.28/04 (LCG 104)

I couldn't find any code change in  `master` in the relevant areas.

### Installation method

anaconda and LCG 104

### Operating system

Linux (CentOS 9, Ubuntu 22.04, Ubuntu 24.04)

### Additional context

Possibly related to the following Jira tickets:
* https://its.cern.ch/jira/browse/ROOT-10778 (open)
* https://its.cern.ch/jira/browse/ROOT-2935 (reportedly fixed in 6.04.00)

vepadulano:
In the case of a TTree with a friend TChain, the call to `GetEntry` that needs to move to a new tree in the friend TChain was not correctly updating the branch addresses. SetBranchAddress now makes a distinction between the case of the branch belonging to the main tree versus the case of the branch belonging to one of the friends. In order for this to work, GetBranch functionality has been split in two: GetBranchFromSelf checks whether the branch name corresponds to a branch of this tree and returns it if so; GetBranchFromFriends checks whether the branch name corresponds to a branch found in one of the trees and returns it if so.

Fixes #16804

### DIFF
--- a/tree/tree/inc/TChainElement.h
+++ b/tree/tree/inc/TChainElement.h
@@ -46,6 +46,7 @@ protected:
    char         *fPackets;           ///<! Packet descriptor string
    TBranch     **fBranchPtr;         ///<! Address of user branch pointer (to updated upon loading a file)
    Int_t         fLoadResult;        ///<! Return value of TChain::LoadTree(); 0 means success
+   bool          fIsDelayed{false};  ///<! Postpone detection of missing branch in certain contexts (e.g. TTree::SetBranchStatus)
 
 public:
    TChainElement();
@@ -78,6 +79,8 @@ public:
    virtual void        SetNumberEntries(Long64_t n) {fEntries=n;}
    virtual void        SetPacketSize(Int_t size = 100);
    virtual void        SetStatus(Int_t status) {fStatus = status;}
+           void        IsDelayed(bool isDelayed) { fIsDelayed = isDelayed; }
+           bool        IsDelayed() const { return fIsDelayed; }
 
    ClassDefOverride(TChainElement,2);  //A chain element
 };
--- a/tree/tree/src/TTree.cxx
+++ b/tree/tree/src/TTree.cxx
@@ -5338,10 +5338,96 @@ TBranch *R__GetBranch(const TObjArray &branches, const char *name)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+/// Returns a pointer to the branch with the given name, if it can be found in
+/// this tree. Otherwise, returns nullptr.
+TBranch *TTree::GetBranchFromSelf(const char *branchName)
+{
+   // Look for an exact match in the list of top level
+   // branches.
+   if (auto *br = static_cast<TBranch *>(fBranches.FindObject(branchName)))
+      return br;
+
+   // Look for an exact match in the mapping from branch name to TBranch *
+   // gathered when first reading the TTree from disk.
+   if (auto it = fNamesToBranches.find(branchName); it != fNamesToBranches.end())
+      return it->second;
+
+   // Search using branches, breadth first.
+   if (auto *br = R__GetBranch(fBranches, branchName))
+      return br;
+
+   // Search using leaves.
+   TObjArray *leaves = GetListOfLeaves();
+   Int_t nleaves = leaves->GetEntriesFast();
+   for (Int_t i = 0; i < nleaves; i++) {
+      TLeaf *leaf = (TLeaf *)leaves->UncheckedAt(i);
+      TBranch *branch = leaf->GetBranch();
+      if (!strcmp(branch->GetName(), branchName)) {
+         return branch;
+      }
+      if (!strcmp(branch->GetFullName(), branchName)) {
+         return branch;
+      }
+   }
+
+   return nullptr;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+/// Returns a pointer to the branch with the given name, if it can be found in
+/// the list of friends of this tree. Otherwise, returns nullptr.
+TBranch *TTree::GetBranchFromFriends(const char *branchName)
+{
+   if (!fFriends) {
+      return nullptr;
+   }
+
+   // Search in list of friends.
+   TFriendLock lock(this, kGetBranch);
+   TIter next(fFriends);
+   TFriendElement *fe = nullptr;
+   while ((fe = (TFriendElement *)next())) {
+      TTree *t = fe->GetTree();
+      if (t) {
+         TBranch *branch = t->GetBranch(branchName);
+         if (branch) {
+            return branch;
+         }
+      }
+   }
+
+   // Second pass in the list of friends when
+   // the branch name is prefixed by the tree name.
+   next.Reset();
+   while ((fe = (TFriendElement *)next())) {
+      TTree *t = fe->GetTree();
+      if (!t) {
+         continue;
+      }
+      const char *subname = strstr(branchName, fe->GetName());
+      if (subname != branchName) {
+         continue;
+      }
+      Int_t l = strlen(fe->GetName());
+      subname += l;
+      if (*subname != '.') {
+         continue;
+      }
+      subname++;
+      TBranch *branch = t->GetBranch(subname);
+      if (branch) {
+         return branch;
+      }
+   }
+
+   return nullptr;
+}
+
+////////////////////////////////////////////////////////////////////////////////
 /// Return pointer to the branch with the given name in this tree or its friends.
 /// The search is done breadth first.
 
-TBranch* TTree::GetBranch(const char* name)
+TBranch *TTree::GetBranch(const char *name)
 {
    // We already have been visited while recursively
    // looking through the friends tree, let's return.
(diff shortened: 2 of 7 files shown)


tree/tree/src/TTree.cxx
```diff
@@ -8524,27 +8553,42 @@ Int_t TTree::SetBranchAddress(const char* bname, void* addr, TClass* ptrClass, E
...
-      if ((res & kNeedEnableDecomposedObj) && !branch->GetMakeClass())
-         branch->SetMakeClass(true);
-      SetBranchAddressImp(branch,addr,ptr);
-   } else {
-      if (ptr) *ptr = nullptr;
+   // Check friends
+   if (fFriends) {
+      for (auto *fe : TRangeDynCast<TFriendElement>(fFriends)) {
+         if (auto *tree = fe->GetTree()) {
+            int status = tree->SetBranchAddress(bname, addr, ptr);
```
pcanal, line 8579:
```suggestion
            int status = tree->SetBranchAddress(bname, addr, ptr, ptrClass, datatype, isptr);
```
Otherwise we inadvertently disable the type check for the branches in the friends, isn't it?

tree/tree/src/TTree.cxx
```diff
@@ -8524,27 +8553,42 @@ Int_t TTree::SetBranchAddress(const char* bname, void* addr, TClass* ptrClass, E
...
+            if (status != kMissingBranch) {
+               return status;
+            }
```
pcanal, line 8582:
```suggestion
            return status;
```
Now, that we use the full `SetBranchAddress`, the error handling is already done, isn't it?

tree/treeplayer/test/gh16804.cxx
```diff
@@ -0,0 +1,162 @@
...
+      friendChain.Add(fn);
+
+   auto mainFile = std::make_unique<TFile>(fMainFileName);
+   auto mainTree = mainFile->Get<TTree>(fMainTreeName);
+   mainTree->AddFriend(&friendChain);
+
+   int index = -1, value = -1, mainBranch = -1;
+   auto mainBranchRet = mainTree->SetBranchAddress("mainBranch", &mainBranch);
+   auto indexRet = mainTree->SetBranchAddress("index", &index);
+   auto valueRet = mainTree->SetBranchAddress("value", &value);
```
pcanal, line 93: We should consider adding some error case testing ...

tree/treeplayer/test/gh16804.cxx
```diff
@@ -0,0 +1,162 @@
...
+
+TEST_F(RegresssionGH16804, MainTTreeFriendTChain)
+{
+   TChain friendChain{fFriendTreeName};
+   for (const auto &fn : fFriendFileNames)
+      friendChain.Add(fn);
+
+   auto mainFile = std::make_unique<TFile>(fMainFileName);
+   auto mainTree = mainFile->Get<TTree>(fMainTreeName);
+   mainTree->AddFriend(&friendChain);
```
pcanal, line 88: We should also test here-ish the case of a friend chain of a chain (in particular to double check we did not break it)

pcanal (changes requested):

vepadulano:
Dear @lmoureaux ,

Thank you for confirming this fixes the linked issue. I would not tie fixing both issues together in the same PR. Once the review process is done for this one I will merge it, then move to fixing #16805

lmoureaux:
Thanks @vepadulano for the clarification, this sounds reasonable!

tree/treeplayer/test/gh16804.cxx
```diff
@@ -0,0 +1,388 @@
...
+   // for friend 2. This happens at every file boundary, so the errors are printed
+   // 2 times for friend 1 and 4 times for friend 2. It is yet unclear how this
+   // could be prevented. Note that the values are properly found and loaded for
+   // all branches of all friends and in fact the EXPECT_EQ calls in the
+   // following for loop all pass.
+   ROOT::TestSupport::CheckDiagsRAII diagRAII;
+   diagRAII.requiredDiag(kError, "TTree::SetBranchStatus", "unknown branch -> a");
+   diagRAII.requiredDiag(kError, "TTree::SetBranchStatus", "unknown branch -> b");
+   diagRAII.requiredDiag(kError, "TTree::SetBranchStatus", "unknown branch -> index");
+   diagRAII.requiredDiag(kError, "TTree::SetBranchStatus", "unknown branch -> value");
```
vepadulano, line 378: See explanation on current issue with spurious error messages

vepadulano (commented):
I have addressed comments and added the two tests suggested by @pcanal . I wanted to add a third test, i.e. a main TTree with two top-level friend TChains. While the loading of the entries works and all values are properly retrieved, there is currently an issue with spurious error messages being printed by `TTree::SetBranchStatus` see the test file for a full description

tree/treeplayer/test/gh16804.cxx
lmoureaux: Hmm this is indeed annoying. From the description, I think we'll also want a test where both friend chains share a branch. It should always load values from the first chain.

vepadulano: Indeed by only using `SetBranchAddress` the branch read is the one from the first TTree who has the branch "x", any other TTree who might have the same branch name is disregarded.

lmoureaux: But does this work with chains? Since the full branch list isn't known before the first GetEntry...

vepadulano: Just tested, works with TChain the same way

pcanal: Aren't those message directly linked to https://github.com/root-project/root/pull/18871#discussion_r2109406331 (Even-though the propose solution is too simplistic).

tree/treeplayer/test/gh16804.cxx
```diff
@@ -0,0 +1,485 @@
...
+   std::vector<int> expectedB(20);
+   std::iota(expectedMainBranch.begin(), expectedMainBranch.end(), 0);
+   std::reverse(expectedMainBranch.begin(), expectedMainBranch.end());
+   std::iota(expectedIndex.begin(), expectedIndex.end(), 0);
+   std::iota(expectedValue.begin(), expectedValue.end(), 100);
+   std::iota(expectedA.begin(), expectedA.end(), 200);
+   std::iota(expectedB.begin(), expectedB.end(), 300);
+   auto nEntries = mainTree->GetEntriesFast();
+
+   // FIXME: The following diagnostics should not happen. This is the situation:
```
pcanal, line 356: This is, in my humble opinion, a blocker.

pcanal:
Side note: this PR was arbitrarily picked to check the test coverage feature on a PR.

tree/treeplayer/test/gh16804.cxx
lmoureaux: Or we force the chain to load, like `TChain` already does in some cases: `FindBranch()`, `FindLeaf()`, `GetAlias()`, etc:
https://github.com/root-project/root/blob/40d2a622d3fe1b876c8c31a04edd37c25423a6ef/tree/tree/src/TChain.cxx#L930-L933

pcanal:
Does the new code 'properly' warning if a branch is actually missing?

vepadulano:
> Does the new code 'properly' warning if a branch is actually missing?

Yes, the errors still appear properly within `SetBranchAddress`. I have added more tests to demonstrate this, see all the `WrongBranchName*` test cases.

~The error printout disappears, for this very specific case, from `SetBranchStatus`. I believe this is actually desired, as the extra output is at most redundant and in the case highlighted with this PR simply wrong. There is one roottest test named `roottest-root-tree-branches-make ` which was actually testing for this extra printout. My current understanding is we could simply change the reference file there and remove the extra error statements, I will double check that it makes sense~

EDIT: after investigation, the issue with the roottest test was simply that the new boolean data member in TChainElement wasn't properly initialised, now that is done and the test should be working again.

tree/tree/inc/TChainElement.h
```diff
@@ -78,6 +78,8 @@ class TChainElement : public TNamed {
    virtual void        SetNumberEntries(Long64_t n) {fEntries=n;}
    virtual void        SetPacketSize(Int_t size = 100);
    virtual void        SetStatus(Int_t status) {fStatus = status;}
+   virtual void IsDelayed(bool isDelayed) { fIsDelayed = isDelayed; }
```
pcanal: formatting; try to align with the other functions.

tree/tree/inc/TChainElement.h
```diff
@@ -42,11 +42,11 @@ class TChainElement : public TNamed {
    UInt_t        fBaddressType;      ///<! Type of the value pointed to by fBaddress
    bool          fBaddressIsPtr : 1; ///<! True if the address is a pointer to an address
    bool          fDecomposedObj : 1; ///<! True if the address needs the branch in MakeClass/DecomposedObj mode.
-   bool          fCheckedType : 1;   ///<! True if the branch type and the address type have been checked.
+   bool fCheckedType : 1;            ///<! True if the branch type and the address type have been checked.
    char         *fPackets;           ///<! Packet descriptor string
    TBranch     **fBranchPtr;         ///<! Address of user branch pointer (to updated upon loading a file)
    Int_t         fLoadResult;        ///<! Return value of TChain::LoadTree(); 0 means success
-
+   bool fIsDelayed{false}; ///<! Postpone detection of missing branch in certain contexts (e.g. TTree::SetBranchStatus)
```
pcanal: formatting; try to align with the other data members.

vepadulano:
The latest commit highlighted a pre-existing bug that was relying on a side-effect of the status of master before this PR. Now it is tracked at https://github.com/root-project/root/issues/19220 . Once that is fixed we can resume the work of this PR.

tree/treeplayer/test/gh16804.cxx
```diff
@@ -0,0 +1,1111 @@
...
+   for (const auto &fn : fFriendFileNames)
+      friendChain.Add(fn);
+   TChain otherFriendChain{fOtherFriendTreeName};
+   for (const auto &fn : fOtherFriendFileNames)
+      otherFriendChain.Add(fn);
+
+   auto mainFile = std::make_unique<TFile>(fMainFileName);
+   auto mainTree = mainFile->Get<TTree>(fMainTreeName);
+   mainTree->AddFriend(&friendChain);
+   mainTree->AddFriend(&otherFriendChain);
```
pcanal, line 577: As a side note, it looks like we are not testing (in this file), the case(s):
```
   mainTree->AddFriend(&friendChain);
   friendChain->AddFriend(&otherFriendChain);
```

tree/treeplayer/test/gh16804.cxx
pcanal: This is essentially what we ended up with.   Thanks.

tree/treeplayer/test/gh16804.cxx
pcanal: This was address and those cases now issue an error.

tree/treeplayer/test/gh16804.cxx
vepadulano: Indeed, this PR does not involve changes nor testing regarding the semantics and functionality of "friends of friends". I will leave that work for another PR as I believe it may be related to #16805 anyway

PR #18872, opened 2025-05-26
[cppyy] Enable tests on windows

aaronj0:
Trigger a verbose CI run with Windows to examine the cppyy test status


guitargeek:
```txt
2025-05-26T19:53:28.0200811Z 24:         if sc == -1:
2025-05-26T19:53:28.0200984Z 24: >           raise RuntimeError("Unable to load reflection library "+name)
2025-05-26T19:53:28.0201385Z 24: E           RuntimeError: Unable to load reflection library C:\ROOT-CI\build\bindings\pyroot\cppyy\cppyy\test/stltypesDict
```

Looks like we have to use a proper `os.path.join` in Python...

roottest/python/stl/Makefile
```diff
@@ -1,29 +0,0 @@
-# This is a template for all makefile.
```
pcanal: Please consider submitted that commit in its own (NFC) PR.

guitargeek: Sure! Would you mind reviewing it?

 * https://github.com/root-project/root/pull/20859

guitargeek (approved):
Thank you very much! I made some additional changes to bring the PR over the finish line:

  * use ACLiC for dictionary generation, which worked the easiest on all platforms (as was done by the old `roottest/python`)
  * remove some tests in `roottest/python` that are redundant, now that we also run the cppyy tests on all platforms
  * consistently use the pytest markers to steer disabling of tests. This gives us a better overview of what is actually disabled, and lets us do more granular disabling
  * temporarily circumvent a crash on Windows 64 bit by replacing a `Py_DECREF` with `Py_XDECREF`, which includes a `nullptr` protection (FYI, @bellenot).

Note that some incremental builds might fail in the next hours, because some declarations will be available both in the old test dictionaries generated by `ROOT_GENERATE_DICTIONARY`, and the new ones generated by ACLiC. The declarations might clash and cause test failures.

PR #18876, opened 2025-05-27
[CMake] Discourage the use of `soversion=ON`

guitargeek:
Discourage the use of `soversion=ON` because the rootmap files don't
support it. Therefore, this option has no clear benefit. It seemingly
allows you to have multiple ROOT versions in parallel on the system, by
things will break anyway since e.g. the rootmap files are unversioned.


cholmcc:
Just to be clear - numbers in so-names makes the runtime environment much more stable.  An application linked to ROOT X.Y.Z will not use libraries from ROOT X.Y.W.   As for the `rootmap` files - yes, they are not versioned (but probably should be).   A simple fix is that you install `.rootmap` files in a versioned sub-directory of `lib` - e.g., `lib/X.Y.Z` - along with any plugin "libraries".   This is the _only_ way to make a clean system that can accommodate multiple run-time environments and which adhere to the FHS.

ferdymercury:
> which adhere to the FHS.

Cross-link: https://its.cern.ch/jira/browse/ROOT-10151

guitargeek:
Nice idea! If we have a way to proper version the rootmap files and others, we can definitely recommend the `soversion=ON`. But as long as this is not done by our build system, we can't recommend it.

cmake/modules/RootBuildOptions.cmake
```diff
@@ -158,7 +158,7 @@ ROOT_BUILD_OPTION(rpath ON "Link libraries with built-in RPATH (run-time search
 ROOT_BUILD_OPTION(runtime_cxxmodules ON "Enable runtime support for C++ modules")
 ROOT_BUILD_OPTION(shadowpw OFF "Enable support for shadow passwords")
 ROOT_BUILD_OPTION(shared ON "Use shared 3rd party libraries if possible")
-ROOT_BUILD_OPTION(soversion OFF "Set version number in sonames (recommended)")
+ROOT_BUILD_OPTION(soversion OFF "Set version number in sonames for shared libraries. Not recommended, as the pcm and rootmap files are unversioned and always point to the non-versioned shared libraries.")
```
ferdymercury:
```suggestion
ROOT_BUILD_OPTION(soversion OFF "Set version number in sonames for shared libraries. Not recommended, as the pcm and rootmap files do not (yet) support versioning and always point to the non-versioned shared libraries.")
```

cholmcc: > Nice idea! If we have a way to proper version the rootmap files and others, we can definitely recommend the soversion=ON. But as long as this is not done by our build system, we can't recommend it.

It shouldn't be all that difficult to do with the current build system.  Just install `.rootmap` in `lib`/<version>, and include that directory in the `system.rootrc` as part of the `DynamicPath` setting.  I believe ROOT looks for `.rootmap` files in that path and those set by some environment variable (not `LD_LIBRARY_PATH`, if I remember correctly).   

Point is, you can do it tomorrow if you want.  Better fix this in a proper way right away than keep pushing down the line. 

Yours,
_Christian_

cholmcc: > in `lib/`,

Should have been `lib/`&lt;_version_&gt;

cholmcc: Of course, even better if it is in the code to load the `.rootmap` files.

guitargeek: You're right with all of that, but even if conceptually the solution is mapped out someone still needs to do it :slightly_smiling_face: Until then, let's put the suggested note in the build options (thanks @bellenot and @ferdymercury) and continue the discussion in the GitHub issue.

PR #18885, opened 2025-05-28
[ntuple] Merger: properly handle untyped records

silverweed:
With this change, we explicitly refuse to merge any 2 untyped records that don't have compatible subfields (in the proper order). This was de facto already the case before this change, but instead of failing with an error the merger would simply assert.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/src/RNTupleMerger.cxx
+++ b/tree/ntuple/src/RNTupleMerger.cxx
@@ -285,7 +285,7 @@ struct RCommonField {
    const ROOT::RFieldDescriptor *fSrc;
    const ROOT::RFieldDescriptor *fDst;
 
-   RCommonField(const ROOT::RFieldDescriptor *src, const ROOT::RFieldDescriptor *dst) : fSrc(src), fDst(dst) {}
+   RCommonField(const ROOT::RFieldDescriptor &src, const ROOT::RFieldDescriptor &dst) : fSrc(&src), fDst(&dst) {}
 };
 
 struct RDescriptorsComparison {
@@ -411,7 +411,7 @@ CompareDescriptorStructure(const ROOT::RNTupleDescriptor &dst, const ROOT::RNTup
       const auto srcFieldId = src.FindFieldId(dstField.GetFieldName());
       if (srcFieldId != ROOT::kInvalidDescriptorId) {
          const auto &srcField = src.GetFieldDescriptor(srcFieldId);
-         commonFields.push_back({&srcField, &dstField});
+         commonFields.push_back({srcField, dstField});
       } else {
          res.fExtraDstFields.emplace_back(&dstField);
       }
@@ -423,7 +423,11 @@ CompareDescriptorStructure(const ROOT::RNTupleDescriptor &dst, const ROOT::RNTup
    }
 
    // Check compatibility of common fields
-   for (const auto &field : commonFields) {
+   auto fieldsToCheck = commonFields;
+   // NOTE: using index-based for loop because the collection may get extended by the iteration
+   for (std::size_t fieldIdx = 0; fieldIdx < fieldsToCheck.size(); ++fieldIdx) {
+      const auto &field = fieldsToCheck[fieldIdx];
+
       // NOTE: field.fSrc and field.fDst have the same name by construction
       const auto &fieldName = field.fSrc->GetFieldName();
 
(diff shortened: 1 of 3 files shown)


pcanal:
Why is shuffling the data member not supported?

pcanal:
> Because untyped records store their members consecutively with the same layout as C/C++ would, so they are not necessarily compatible.

So is regular ROOT I/O.  Nonetheless supporting the schema evolution of shuffling/changing the order is as straightforward as updating the (in memory) offset based on the field name rather than 'just' relying on their ordering ...

silverweed:
We might want to support this in the future but at the moment there is no explicit interaction between the Merger and schema evolution, so I think we want to discuss the matter first.

silverweed:
I'm actually reconsidering my conclusion: the fact that the record fields have different in-memory layout should not impact our ability to merge them, as on disk their subfields they should be treated as if they were top-level fields in different order, which we can merge. I'll see if I can easily support that.

EDIT: I tried, and I realized that, while doable, it's not that easy. The reason is that currently some parts of the merger rely on the fact that we can 1:1 map column indices in all common fields. Supporting scrambled record subfields would require introducing a mapping between the source and the destination's common columns. Again, this is doable but I'd take some time to discuss it and possibly introduce it in a future PR, not this one.

silverweed:
@pcanal is it fine if we merge this and then discuss the unordered untyped records when @jblomer is back?

tree/ntuple/src/RNTupleMerger.cxx
```diff
@@ -533,6 +547,22 @@ CompareDescriptorStructure(const ROOT::RNTupleDescriptor &dst, const ROOT::RNTup
...
+      // Require that subfields are compatible
+      const auto &srcLinks = field.fSrc->GetLinkIds();
+      const auto &dstLinks = field.fDst->GetLinkIds();
+      if (srcLinks.size() != dstLinks.size()) {
+         std::stringstream ss;
+         ss << "Field `" << field.fSrc->GetFieldName()
+            << "` has a different number of children than previously-seen field with the same name (old: "
+            << dstLinks.size() << ", new: " << srcLinks.size() << ")";
+         errors.push_back(ss.str());
+      } else
```
hahnjo, line 560: This is quite hard to read, I would suggest to use curly braces consistently

tree/ntuple/src/RNTupleMerger.cxx
```diff
@@ -533,6 +547,22 @@ CompareDescriptorStructure(const ROOT::RNTupleDescriptor &dst, const ROOT::RNTup
...
+         for (std::size_t linkIdx = 0, linkNum = srcLinks.size(); linkIdx < linkNum; ++linkIdx) {
+            const auto &srcSubfield = src.GetFieldDescriptor(srcLinks[linkIdx]);
+            const auto &dstSubfield = dst.GetFieldDescriptor(dstLinks[linkIdx]);
+            fieldsToCheck.push_back(RCommonField{srcSubfield, dstSubfield});
```
hahnjo, line 564: It looks like you are implementing a workqueue algorithm, so `std::deque` might be more efficient and avoid collecting all fields.

hahnjo (commented):
I only had time to review now (the PR was only approved by Philippe); some considerations inline. Also the commit message doesn't seem to reflect well the latest changes: AFAICT it's now checking the entire field hierarchy and also the structure and number of subfields...

tree/ntuple/src/RNTupleMerger.cxx
silverweed: It is indeed a work queue algorithm, but `vector` works just fine in this case since I don't have to pop elements (incrementing the index is sufficient and faster)

tree/ntuple/src/RNTupleMerger.cxx
silverweed: https://github.com/root-project/root/pull/19018

tree/ntuple/src/RNTupleMerger.cxx
hahnjo: Well yes, any workqueue algorithm can be implemented with a `std::vector`, but it requires memory to store all elements. A `std::deque` internally uses a circular buffer, so it is equally efficient to iterate but you only need memory for the maximum number of elements ever in the queue.

silverweed: Sure, but we're talking about queues that hardly ever have more than ~1000 elements or so (usually far less), each of whom is 2 pointers, so I don't think memory is much of a concern here. However if you disagree I can open a PR to change it.

PR #18886, opened 2025-05-28
[RF] Don't use TRef for pointer to workspace in ModelConfig

guitargeek:
The behavior of the TRef is not very in the context of the RooWorkspace, leading to inconsistencies when copying workspace while a file is open.

Using raw transient pointers that are set when streaming or copying the workspace is simpler and more robust.


roofit/roofitcore/inc/RooFit/ModelConfig.h
```diff
@@ -369,8 +368,7 @@ class ModelConfig final : public TNamed, public RooWorkspaceHandle {
    std::unique_ptr<RooAbsReal> createNLLImpl(RooAbsData &data, const RooLinkedList &cmdList) const;
    std::unique_ptr<RooFitResult> fitToImpl(RooAbsData &data, const RooLinkedList &cmdList) const;
 
-   ClassDefOverride(ModelConfig,
-                    6); ///< A class that holds configuration information for a model using a workspace as a store
+   ClassDefOverride(ModelConfig, 7);
```
pcanal, line 371: What is the plan to read file containing version 6?

guitargeek: Automatic schema evolution: a member was made transient from version 6 to 7. The transient member will be filled now in the `RooWorkspace::Streamer()` when reading, since any ModelConfig is always stored inside a RooWorkspace, or the workspace pointer `fWSName` is not set anyway.

pcanal: Fair enough I had missed that `fRefWS` is now 'transient'.   Nonetheless we should test reading the old file to make sure there is not a spurrious message about `TRef` not being readable into a `RooWorkspace*`.

ferdymercury: Maybe related/affected by: https://its.cern.ch/jira/browse/ROOT-8837

lmoneta (approved):
Thank you Joans for this improvement. 
If there are no I/O compatibility issues reading old files, it is fine for me. 
One question, would not be better to use a transient shared_ptr to the Workspace, in case a workspace i deleted before the ModelConfig?

roofit/roofitcore/inc/RooFit/ModelConfig.h
guitargeek: Good point. I have added a unit test for reading an old file with a ModelConfig version 6.

Issue #18847, opened 2025-05-23
[RF] Memory issue in destructor of RooStats::HypoTestInverterResult


TomasDado:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

The attached code causes random crashes but it can be reliably identified as an issue with valgrind.
The example stack trace is:
```
===========================================================
There was a crash.
This is the entire stack trace of all threads:
===========================================================
#0  __syscall_cancel (a1=<optimized out>, a2=a2
entry=140730861103080, a3=a3
entry=0, a4=a4
entry=0, a5=a5
entry=0, a6=a6
entry=0, nr=61) at ./nptl/cancellation.c:77
#1  0x00007291be31b79f in __GI___wait4 (pid=<optimized out>, stat_loc=stat_loc
entry=0x7ffe74fc27e8, options=options
entry=0, usage=usage
entry=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
#2  0x00007291be31b7eb in __GI___waitpid (pid=<optimized out>, stat_loc=stat_loc
entry=0x7ffe74fc27e8, options=options
entry=0) at ./posix/waitpid.c:38
#3  0x00007291be25bebb in do_system (line=<optimized out>) at ../sysdeps/posix/system.c:172
#4  0x00007291bff1a6fc in TUnixSystem::Exec (this=0x5d47e70e21c0, shellcmd=<optimized out>) at /home/tomas/ROOT_DEBUG/root_src/core/unix/src/TUnixSystem.cxx:2160
#5  TUnixSystem::StackTrace (this=0x5d47e70e21c0) at /home/tomas/ROOT_DEBUG/root_src/core/unix/src/TUnixSystem.cxx:2451
#6  0x00007291bff1a004 in TUnixSystem::DispatchSignals (this=0x5d47e70e21c0, sig=kSigSegmentationViolation) at /home/tomas/ROOT_DEBUG/root_src/core/unix/src/TUnixSystem.cxx:3671
#7  <signal handler called>
#8  TObjArray::Expand (this=this
entry=0x5d47eb599048, newSize=0) at /home/tomas/ROOT_DEBUG/root_src/core/cont/src/TObjArray.cxx:400
#9  0x00007291beeca9b9 in RooAbsArg::ioStreamerPass2Finalize () at /home/tomas/ROOT_DEBUG/root_src/roofit/roofitcore/src/RooAbsArg.cxx:2433
[...]

PR #18888, opened 2025-05-28
[RF] Don't schema evolve TRefArray to proxy list if there are no proxies

guitargeek:
If there are no proxies for a given RooAbsArg, we don't need to
schedeule the schema evolution of it's proxies. This saves both CPU and
memory usage, since the old proxy TRefArrays are kept around in a
temporary map until the full workspace is read.

This also prevents spurious crashes in the schema evolution of large
workspaces, that can happen for example when the schema evolution of the RooRealVars in a RooFit dataset are interfering with the schema evolution of the variables in the main workspace.

Closes https://github.com/root-project/root/issues/18847.

Also, refactor the RooAbsArg schema evolution to not use global variables.

### DIFF
--- a/roofit/roofitcore/src/RooAbsArg.cxx
+++ b/roofit/roofitcore/src/RooAbsArg.cxx
@@ -103,9 +103,23 @@ bool RooAbsArg::_verboseDirty(false) ;
 bool RooAbsArg::_inhibitDirty(false) ;
 bool RooAbsArg::inhibitDirty() const { return _inhibitDirty && !_localNoInhibitDirty; }
 
-std::map<RooAbsArg*,std::unique_ptr<TRefArray>> RooAbsArg::_ioEvoList;
-std::stack<RooAbsArg*> RooAbsArg::_ioReadStack ;
+namespace {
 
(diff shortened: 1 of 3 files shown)


dpiparo (approved):
Thanks!

roofit/roofitcore/src/RooAbsArg.cxx
```diff
@@ -2374,16 +2388,21 @@ void RooAbsArg::SetNameTitle(const char *name, const char *title)
 void RooAbsArg::Streamer(TBuffer &R__b)
 {
    if (R__b.IsReading()) {
-     _ioReadStack.push(this) ;
+     ioReadStack().push(this) ;
```
pcanal, line 2391: This is not thread safe.  2 threads reading a RooAbsArg will trample on each other.

pcanal: But the 'failure' mode may simply be data put fine in the wrong place leading to inaccurate results that the user may or may not notice and may or may not guess where it is form ... i.e. a silent failure :(.

PR #18895, opened 2025-05-28
Merge requirements.txt from roottest and main repo

guitargeek:
Follows up on the merge of roottest, centralizing also the list of Python dependencies. Increases the size of the implied virtual environment by only 0.05 % (adding 3.6 kB to the 6.9 GB before this PR).

This will make it easier to install the environment for running and testing all features of ROOT, also for the docker images.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,8 +20,10 @@ numba>=0.48
 cffi>=1.9.1
 
 # Notebooks: ROOT C++ kernel
-notebook>=4.4.1
+IPython
+jupyter
 metakernel>=0.20.0
+notebook>=4.4.1
 
 # Distributed RDataFrame
 pyspark>=2.4 # Spark backend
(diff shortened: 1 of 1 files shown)


pcanal (approved):
Thanks.  Should `roottest/requirements.txt` be removed then?

guitargeek:
> Thanks. Should `roottest/requirements.txt` be removed then?

I don't think that's necessary. The master branch of the roottest repo is not used anywhere anymore, so there is no good reason to update it I think

pcanal:
I meant the file `https://github.com/root-project/root/blob/master/roottest/requirements.txt` which is now part of the ROOT repository but is unused :)

guitargeek:
Ah, damn yes of course :laughing: I'll do this together with some next updates, not worth to open a PR for that :slightly_smiling_face:

PR #18904, opened 2025-05-29
Misc ci fixes

dpiparo:
1) Turn mySQL off for alma10
2) Re-enable Windows because in 1 month, 6.32 and 6.36 will be the active branches


pcanal (approved):
LGTM

PR #18905, opened 2025-05-29
[ci] Use builtin compression libraries for special alma9 builds

guitargeek:
This should resolve the new test failues on `alma9 march_native` and `alma9 arm64` that showed up the the transition from AlmaLinux 9.5 to AlmaLinux 9.6.


pcanal (approved):
Thanks.

Issue #18900, opened 2025-05-29
Allow batch mode (gROOT::SetBatch) to be changed from true to false in middle of session


will-cern:
### Feature description

Hello,
I have noticed that if I start a python session in batch mode (`ROOT.gROOT.SetBatch(True)`) and run some lines that would draw some things to a canvas, if I subsequently set `ROOT.gROOT.SetBatch(False)` in the same session, I am unable to visualize new canvases. 
Is there a way to switch between batch mode in the same session? It would be great if this was supported!
Thanks
Will

ferdymercury:
@will-cern Could you try applying this change? https://github.com/root-project/root/pull/18911
(Rebuilding is immediate).

PR #18911, opened 2025-05-29
[pyroot] allow toggling batch mode

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/18900

But seems to break other things.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_facade.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_facade.py
@@ -183,10 +183,10 @@ class ROOTFacade(types.ModuleType):
 
     def _finalSetup(self, setBatch = False):
         if setBatch:
-           if self.__dict__["gROOT"] != cppyy.gbl.ROOT.GetROOT():
-              return
+            if self.__dict__["gROOT"] != cppyy.gbl.ROOT.GetROOT():
+                return
             if not self.gROOT.IsBatch() and self.PyConfig.StartGUIThread:
-               self.app.init_graphics()
+                self.app.init_graphics()
             return
         # Prevent this method from being re-entered through the gROOT wrapper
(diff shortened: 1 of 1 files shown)


bindings/pyroot/pythonizations/python/ROOT/_facade.py
```diff
@@ -181,7 +181,11 @@ def _register_converters_and_executors(self):
         for name, target in executor_aliases.items():
             CPyCppyyRegisterExecutorAlias(name, target)
 
-    def _finalSetup(self):
+    def _finalSetup(self, setBatch = False):
+        if setBatch:
```
ferdymercury, line 185:
```suggestion
        if setBatch and self.__dict__["gROOT"] == cppyy.gbl.ROOT.GetROOT():
```

bindings/pyroot/pythonizations/python/ROOT/_facade.py
```diff
@@ -181,7 +181,13 @@ def _register_converters_and_executors(self):
...
+            if self.__dict__["gROOT"] != cppyy.gbl.ROOT.GetROOT():
+                return
+            if not self.gROOT.IsBatch() and self.PyConfig.StartGUIThread:
+                self.app.init_graphics()
```
vepadulano, line 189: To be honest I don't really understand these changes. the code branch is within `if setBatch`, so I would expect this part to call `SetBatch` somehow, but it doesn't happen. Could you add a comment to better explain the changes?

ferdymercury: Hi!
Before: finalSetup was never called if SetBatch was the parent caller.
The idea now: finalSetup is always called and does something different depending on if the parent caller was SetBatch or someone else.

PR #18916, opened 2025-05-30
[hist] Ensure strict comparison in `std::stable_sort`

guitargeek:
According to the standard, comparison functions require `comp(a, a) == false` for all `a`:
https://en.cppreference.com/w/cpp/named_req/Compare

The `TGraph::Sort()` method is implemented with `std::stable_sort`, which requires such a comparator. However, it's also possible to flip the comparator with the `ascending` flag, meaning it's very eary to end up in a situation where the comparison is not strict anymore. To avoid this, it's better to shortcut the comparison with an additional `a != b` check.

Superseeds:
  * https://github.com/root-project/root/pull/17211


guitargeek:
@bellenot, is this PR also working for you?

lmoneta (approved):
Thank you for the fix!

PR #18922, opened 2025-05-30
[cmake] skip some tests in roottest that require geom when geom is OFF

silverweed:
## Checklist:

- [x] tested changes locally


roottest/root/multicore/CMakeLists.txt
```diff
@@ -127,10 +127,12 @@ ROOTTEST_ADD_TEST(exectsautoparse
                   MACRO exectsautoparse.C
                   OUTREF exectsautoparse.ref)
 
-ROOTTEST_ADD_TEST(exectsinclude
-                 MACRO  exectsinclude.C
-                 OUTCNVCMD grep -v -e "Line processed"
-                 OUTREF exectsinclude.ref)
+if(geom)
+  ROOTTEST_ADD_TEST(exectsinclude
```
pcanal, line 131: This test does not fundamentally depends on geom and already has internal code to support other optional feature missing.  How does it fail without geom?

roottest/root/multicore/exectsinclude.C
```diff
@@ -22,7 +22,9 @@ void exectsinclude(){
                    "TBrowser.h",
                    "TMatrixD.h",
                    "TF1.h",
+#if __has_include("TGeoSphere.h")
```
ferdymercury, line 25:
```suggestion
#ifdef R__HAS_GEOM
```

pcanal: Why? The rest of the file use `__has_include`.

ferdymercury: Just a suggestion, maybe not worth here. In theory, it makes sense if you have 10 headers of the same library to only put one ifdef rather than 10 if has include. (No idea if it's the case here)

ferdymercury (approved):

roottest/root/multicore/exectsinclude.C
silverweed: It would make sense if we had multiple geometry headers, but since here we only have 1 let's stay consistent with the current code.

silverweed:
@pcanal is this good to be merged now?

PR #18927, opened 2025-06-01
[roottest][CMake] Sync minimum CMake version with main ROOT config

guitargeek:
Sync minimum CMake version with main ROOT config.


bellenot (approved):
Thanks Jonas!

PR #18932, opened 2025-06-02
[ntuple] Handle missing values in RNTupleProcessor

enirolf:
When an entry managed by the `RNTupleProcessor` contains missing data for certain fields, for example because they are missing in a subsequent chain or auxiliary ntuple, we need a way to signal to the user that the resulting entry has missing data and is therefore invalid.

The `RNTupleProcessorValuePtr`, together with the `RNTupleProcessorEntry` (which is a thin wrapper around `REntry`) capture this requirement by only returning actual data if the entry is marked as "valid" by the processor. Its use can be seen in action in the `MissingEntries` test in [`ntuple_processor_join.cxx`](https://github.com/enirolf/root/blob/38e21aaabf3f9d64248d466647035a94ffa8c02e/tree/ntuple/test/ntuple_processor_join.cxx).

With the addition of `RNTupleValuePtr`, pointers to the values of an RNTupleModel's default entry, returned by e.g., `MakeField` should not be used together with the RNTupleProcessor anymore, because they may contain incorrect data. More specifically, when a particular field's
data could not be loaded for a certain entry, it will retain its previous value. The `RNTupleValuePtr` adds appropriate checks for this, but to prevent users from using pointers returned from a model's default entry, we only allow RNTupleProcessors to be created from bare models.

Finally, this PR changes the iterator provided by the `RNTupleProcessor` from the full entry to only the (global) entry index. Providing the full entry was already a bit questionable, because it would allow users to call `GetPtr` or related functions *inside* of the loop, which we want to avoid. Other than that, there is (currently) no additional useful information one can obtain from the full entry.

### DIFF
--- a/tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
@@ -336,10 +336,10 @@ public:
    public:
       using iterator_category = std::forward_iterator_tag;
       using iterator = RIterator;
-      using value_type = RNTupleProcessorEntry;
+      using value_type = ROOT::NTupleSize_t;
       using difference_type = std::ptrdiff_t;
-      using pointer = RNTupleProcessorEntry *;
-      using reference = const RNTupleProcessorEntry &;
+      using pointer = ROOT::NTupleSize_t *;
+      using reference = ROOT::NTupleSize_t &;
 
       RIterator(RNTupleProcessor &processor, ROOT::NTupleSize_t entryNumber)
          : fProcessor(processor), fCurrentEntryNumber(entryNumber)
@@ -364,7 +364,7 @@ public:
          return obj;
       }
 
-      reference operator*() { return *fProcessor.fEntry; }
+      reference operator*() { return fCurrentEntryNumber; }
 
       friend bool operator!=(const iterator &lh, const iterator &rh)
       {
--- a/tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
@@ -32,6 +32,8 @@
 
 namespace ROOT {
 namespace Experimental {
+template <typename T>
+class RNTupleProcessorValuePtr;
 
 namespace Internal {
 struct RNTupleProcessorEntryLoader;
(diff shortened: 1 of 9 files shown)


tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -65,6 +67,130 @@ public:
...
+   template <typename T>
+   friend class RNTupleProcessorValuePtr;
+
+private:
+   std::unique_ptr<ROOT::REntry> fEntry;
+   bool fIsValid = true;
+
+   RNTupleProcessorEntry(std::unique_ptr<ROOT::REntry> entry) : fEntry(std::move(entry)) {}
+
+   void SetValid() { fIsValid = true; }
```
silverweed: Any reason not to prefer a `SetValid(bool valid)`? I'd say that generally it's a better API as it composes better (e.g. avoids situations where one has to write `if (valid) { SetValid() } else { SetInvalid() }`).

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -65,6 +67,130 @@ public:
...
+   const RNTupleProcessorEntry *fProcessorEntry;
+   ROOT::RFieldToken fToken;
+
+   RNTupleProcessorValuePtr(std::string_view fieldName, const RNTupleProcessorEntry *processorEntry)
+      : fProcessorEntry(processorEntry), fToken(processorEntry->GetToken(fieldName))
+   {
+   }
+
+public:
+   bool HasValue() { return fProcessorEntry->IsValid(); }
```
silverweed: Probably should be `const`

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -65,6 +67,130 @@ public:
...
+
+template <>
+class RNTupleProcessorValuePtr<void> {
+   friend class RNTupleProcessor;
+
+private:
+   const RNTupleProcessorEntry *fProcessorEntry;
+   ROOT::RFieldToken fToken;
+
+   RNTupleProcessorValuePtr(std::string_view fieldName, const RNTupleProcessorEntry *processorEntry)
```
silverweed: Can `processorEntry` be null here? If not, I'd pass it by reference (but it's fine to store it by pointer ofc)

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -116,6 +242,19 @@ protected:
...
    std::size_t fCurrentProcessorNumber = 0;    //< Number of the currently open inner processor
 
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Check that the model provided for constructing the processor is bare.
+   ///
+   /// The use of bare models in the RNTupleProcessor is enforced because it adds additional field value pointer
+   /// management through RNTupleProcessorValuePtr. The use of pointers returned from a model's default entry (e.g.,
+   /// through RNTupleModel::AddField) may contain incorrect information, specifically when the corresponding field is
+   /// missing from an entry.
+   static void EnsureModelIsBare(const RNTupleModel &model)
```
silverweed: If I'm not missing something, this is probably better off as a static function in the cxx file directly.

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -65,6 +67,130 @@ public:
...
+template <typename T>
+class RNTupleProcessorValuePtr {
+   friend class RNTupleProcessor;
+   friend class RNTupleProcessorEntry;
+
+private:
+   const RNTupleProcessorEntry *fProcessorEntry;
+   ROOT::RFieldToken fToken;
+
+   RNTupleProcessorValuePtr(std::string_view fieldName, const RNTupleProcessorEntry *processorEntry)
```
vepadulano: Depending on the interplay between `RNTupleProcessorEntry` and the dependent type `RNTupleProcessorValuePtr`, the current class layout might not represent a very user-friendly interface

tree/ntuple/src/RNTupleProcessor.cxx
```diff
@@ -493,15 +456,16 @@ ROOT::NTupleSize_t ROOT::Experimental::RNTupleJoinProcessor::LoadEntry(ROOT::NTu
...
-                                  fAuxiliaryProcessor->GetProcessorName() + "\""));
-      }
+      if (fAuxiliaryProcessor->LoadEntry(entryNumber) == kInvalidNTupleIndex)
+         fEntry->SetInvalid();
 
       return entryNumber;
    }
 
+   // We need to validate the entry again in order to read the values of the join fields from the primary processor --
+   // which is guaranteed to have valid values.
```
vepadulano, line 478: Is it always the case? What happens for example if the primary processor holds a chain of files and we are currently reading at least one field from the main processor which is not present in the current file of the chain?

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -65,6 +67,130 @@ public:
...
+   bool IsValid() const { return fIsValid; }
+
+   ROOT::RFieldToken GetToken(std::string_view fieldName) const { return fEntry->GetToken(fieldName); }
+
+   ROOT::REntry::ConstIterator_t begin() const { return fEntry->begin(); }
+   ROOT::REntry::ConstIterator_t end() const { return fEntry->end(); }
+};
+
+template <typename T>
+class RNTupleProcessorValuePtr {
```
vepadulano, line 136: Since we're introducing this new part of the API, I'm wondering. Does it make sense to think about a naming using terms borrowed from computer science concepts which might be relevant here? I'm thinking about something like `RNTupleProcessor[Future,Promise,Delayed]`.

tree/ntuple/test/ntuple_processor.cxx
```diff
@@ -143,46 +142,32 @@ TEST_F(RNTupleProcessorTest, Base)
 {
    auto proc = RNTupleProcessor::Create({fNTupleNames[0], fFileNames[0]});
 
-   int nEntries = 0;
+   auto x = proc->GetValuePtr<float>("x");
+   auto y = proc->GetValuePtr<void>("y");
```
vepadulano, line 147: Just for curiosity, why not `GetValuePtr<std::vector<float>>` here?

tree/ntuple/test/ntuple_processor_chain.cxx
```diff
@@ -151,34 +143,36 @@ TEST_F(RNTupleChainProcessorTest, WithBareModel)
...
-      entry++;
+      iter++;
    }
 
    try {
-      entry++;
+      iter++;
       FAIL() << "having missing fields in subsequent ntuples should throw";
    } catch (const ROOT::RException &err) {
       EXPECT_THAT(err.what(), testing::HasSubstr("field \"y\" not found in the current RNTuple"));
```
vepadulano, line 178: It would be extremely useful to point out that field `y` is missing from RNTuple named `xxx` in file `yyy`.

tree/ntuple/test/ntuple_processor_join.cxx
```diff
@@ -187,61 +192,42 @@ TEST_F(RNTupleJoinProcessorTest, MissingEntries)
...
+   EXPECT_EQ(i.GetPtr(), nullptr);
    try {
-      ++procIter;
+      *i;
+      FAIL() << "dereferencing a value from an invalid entry should throw";
    } catch (const ROOT::RException &err) {
-      EXPECT_THAT(err.what(),
-                  testing::HasSubstr(
-                     "entry 3 in the primary processor has no corresponding entry in auxiliary processor \"ntuple4\""));
+      EXPECT_THAT(err.what(), testing::HasSubstr("cannot read \"i\" because the entry it belongs to is invalid"));
```
vepadulano, line 211: Ideally, all the information from the previous error and the current error messages would appear together.
* which entry did not match
* in which processor name
* which field cannot be read because of this

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
enirolf: I see your point, but I don't think `[Future,Promise,Delayed]` is the most accurate here since there is no asynchronicity involved. I would say semantically this is more similar to a `std::optional` (or Haskell's `Maybe` for that matter ;)). In that case it could/would be `RNTupleProcessor[Optional,Maybe]Value` or something like that.

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
enirolf: Could you elaborate? For users the only way to get access to an `RNTupleProcessorValuePtr` is through `RNTupleProcessor::GetValuePtr`. The `RNTupleProcessorEntry` is completely hidden.

tree/ntuple/test/ntuple_processor.cxx
enirolf: Because I want to test that it also works with `void`. Let me add a comment to clarify.

tree/ntuple/test/ntuple_processor_join.cxx
enirolf: The field name I can add now, but for the other two it's tricky because then either the `RNTupleProcessorValuePtr` or `RNTupleProcessorEntry` class needs to be become dependent on the `RNTupleProcessor` itself. Of course this can be done, but perhaps this is something for an error message further upstream (e.g., in the RNTuple data source).

tree/ntuple/src/RNTupleProcessor.cxx
enirolf: This is a really good point, thanks for pointing out! With the current implementation this will actually be caught in `RNTupleSingleProcessor::Connect`, so even before the entry is loaded. It would be nice however to have the same behavior as for invalid joins, where the entry just gets marked as invalid.

I have started to implement this but it turns out to be a bit trickier than I thought, due to (without going into to much detail here) the fact that there is currently no way to "synchronize" models between existing processors.

We already have discussed a revamp of the way models are handled here, and initially I wanted to address this in a separate, follow-up PR, but since this appears to be more coupled to this current PR I think I will start to directly include them to this PR, instead, so I'll change it into a draft for the time being.

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
vepadulano: I like RNTupleProcessorOptionalValue or just RNTupleProcessorOptional 👍

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
vepadulano: Indeed as long as it is a private constructor then we're good, thanks!

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -65,6 +67,129 @@ public:
...
+   template <typename T>
+   friend class RNTupleProcessorValuePtr;
+
+private:
+   std::unique_ptr<ROOT::REntry> fEntry;
+   bool fIsValid = true;
+
+   RNTupleProcessorEntry(std::unique_ptr<ROOT::REntry> entry) : fEntry(std::move(entry)) {}
+
+   void SetValid(bool valid) { fIsValid = valid; }
```
vepadulano, line 83: Just as a heads-up, down the line we may need to add more semantic information to whether the entry is valid or not, see e.g. https://github.com/root-project/root/blob/b4955fc7fb18b15ee2ab96358571a9de1a0e495b/tree/treeplayer/inc/TTreeReader.h#L149-L162

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -65,6 +67,129 @@ public:
...
+
+   std::shared_ptr<T> GetPtr() const
+   {
+      if (fProcessorEntry->IsValid())
+         return fProcessorEntry->GetPtr<T>(fToken);
+
+      return nullptr;
+   }
+
+   const T &operator*() const
```
hahnjo, line 160: You may also want to implement `operator->` so the user can directly call functions.

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
hahnjo: I also like the analogy to `std::optional`, though I'll throw `RNTupleProcessorOptionalPtr` into the mix because it's not actually a copy of the value as far as I understand. Unfortunately all class names are quite long, it might be a possibility to nest them into `RNTupleProcessor`, but not sure how much this would actually help...

enirolf:
Closing this as it has been superseded by #19111.

PR #18950, opened 2025-06-03
[roottest] using cmake in io tests

linev:
Extend functionality of ROOTTEST_GENERATE_DICTIONARY macro,
adding optional sources and compiler options for created library with dictionary

Use together with ROOTTEST_GENERATE_EXECUTABLE in several io tests.

Crucial is library names on Windows - one have to use `lib` in the beginning.
Also for Windows add copying of `libName.lib` file to build directory to let linker working.


roottest/root/io/clones/CMakeLists.txt
```diff
@@ -1,7 +1,24 @@
...
+    RtObj.h
+    LINKDEF RtObjLinkDef.h
+    NO_ROOTMAP NO_CXXMODULE
+    FIXTURES_SETUP RtObjDict_fixture)
+
+  ROOTTEST_GENERATE_EXECUTABLE(tester
+    tester.C
+    LIBRARIES Core libRtObjDictlibgen
+    FIXTURES_REQUIRED RtObjDict_fixture
+    FIXTURES_SETUP RtObj_tester)
```
pcanal, line 17: The test is now slightly different.  Loading from libraries and from executable is extremely similar but not identical ... (did not investigate yet whether it matters to this test or not, it would be best to document whether it is ...)

roottest/root/io/compression/CMakeLists.txt
```diff
@@ -1,7 +1,30 @@
-#-------------------------------------------------------------------------------
-#
-# Placeholder file to translate the tests to the new CTest system. Meanwhile we
-# define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
-#
-#-------------------------------------------------------------------------------
-ROOTTEST_ADD_OLDTEST()
+ROOTTEST_GENERATE_DICTIONARY(libEventDict
```
pcanal, line 1: Now that the name are in the 'global' CMake namespace, do we need to disambiguate them?

roottest/root/io/const/CMakeLists.txt
```diff
@@ -1,7 +1,3 @@
-#-------------------------------------------------------------------------------
-#
-# Placeholder file to translate the tests to the new CTest system. Meanwhile we
-# define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
-#
-#-------------------------------------------------------------------------------
-ROOTTEST_ADD_OLDTEST()
+ROOTTEST_ADD_TEST(test
```
pcanal, line 1: The `rootlogon.C` should either be copied into the binary directory ... or it looks like we no longer need it and it could be removed.

roottest/root/io/constmembers/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
-#-------------------------------------------------------------------------------
-#
-# Placeholder file to translate the tests to the new CTest system. Meanwhile we
-# define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
-#
-#-------------------------------------------------------------------------------
-ROOTTEST_ADD_OLDTEST()
+if(broken_tests)
```
pcanal, line 1: With this type of migration from `Makefile` to `CMake` we lose the existing warning/reminder (which to be fair was being overlooked).  We should probably open a generic issue point to the existence of this tests that will be since only with `-Dbroken_tests=ON` (ideally we would have one issue per failing test/directory).   

My apologies if those issues already exists.

roottest/root/io/customCollection/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
-#-------------------------------------------------------------------------------
-#
-# Placeholder file to translate the tests to the new CTest system. Meanwhile we
-# define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
-#
-#-------------------------------------------------------------------------------
-ROOTTEST_ADD_OLDTEST()
+ROOTTEST_ADD_TEST(write
+  MACRO execWriteCustomCollection.cxx+
```
pcanal, line 2:
```
OUTREF execWriteCustomCollection.ref
```
might be missing.

roottest/root/io/clones/CMakeLists.txt
linev: In old `Makefile` also executable was created and run.
And it was run only when failing tests were enabled.

roottest/root/io/compression/CMakeLists.txt
linev: Yes, I just recognize that library name also create **global** target of same name.
I propose to add `LIBNAME` parameter to `ROOTTEST_GENERATE_DICTIONARY` - 
and then use more unique target names

roottest/root/io/constmembers/CMakeLists.txt
linev: We have `win_broken_tests` variable which can be used to enable broken tests on Windows.
I guess we can add  `broken_tests` variable for tests which known to be broken - and cannot be handled by `WILLFAIL` option

roottest/root/io/customCollection/CMakeLists.txt
linev: Yes, I will add it

linev: If I add ref file - I see failure because of extra warning:
```
Warning: Attempt to select a class DataVectorTmplt<Content> with two rules which have incompatible attributes. The attributes such as transiency might not be correctly propagated to the typesystem of ROOT.
		Selected (line -1): Yes
		Attributes: 
		file_name = /home/linev/git/webgui/roottest/root/io/customCollection/execWriteCustomCollection.cxx
		pattern = *
		No field sel rules
		No method sel rules
Conflicting rule already matched:
		Selected (line -1): Yes
		Attributes: 
		name = DataVectorTmplt<Content>
		No field sel rules
		No method sel rules
```
It appears only when ACLiC generates `execWriteCustomCollection.so` file.
Therefore if this test runs second time - warning disappear. 
No idea how to workaround this?

roottest/root/io/clones/CMakeLists.txt
pcanal: Yes, so we need to look into the history/intend to the test to know if it matters whether the dictionary is in a library or in an executable.

roottest/root/io/compression/CMakeLists.txt
pcanal: Isn't that already the first argument?

roottest/root/io/constmembers/CMakeLists.txt
pcanal: The windows one is more systemic choice.  The known to fail every where test are more focused and pre-dated the github issue system.  At the time it was thought that 'recording' their failure by a warning was sufficient, nowadays, I believe it is not and that we need to record it (either coarse grain by having an issue mentioning `broken_tests` or pin-pointed by having an issue by test.  The advantage is that contributor will then have some visibility on the fact that those outstanding issue exist and can be picked up to work on.

roottest/root/io/customCollection/CMakeLists.txt
pcanal:
```
diff --git a/roottest/root/io/customCollection/execWriteCustomCollection.cxx b/roottest/root/io/customCollection/execWriteCustomCollection.cxx
index 060fee42f3c..ddc6b1327e9 100644
--- a/roottest/root/io/customCollection/execWriteCustomCollection.cxx
+++ b/roottest/root/io/customCollection/execWriteCustomCollection.cxx
@@ -63,7 +63,9 @@ public:
 #pragma link C++ class vector<Content>+;
 #endif
 
+#ifndef __ROOTCLING__
 typedef DataVectorTmplt<Content> DataVector;
+#endif
 
 #include "TFile.h"
 #include "TClass.h"
```

linev:
```
+#ifndef __ROOTCLING__
 typedef DataVectorTmplt<Content> DataVector;
+#endif
```
It even worse - ROOT crashes after such change

linev: Looks like following solution works:
```
typedef DataVectorTmplt<Content> DataVector;


#ifdef __MAKECINT__
#pragma link C++ class DataVectorTmplt<Content>+;
#pragma link C++ class vector<Content>+;
#pragma link C++ class DataVector+;
#endif
```

roottest/root/io/clones/CMakeLists.txt
linev: In old `Makefile` dictionary object was directly linked to executable - 

https://github.com/root-project/roottest/blob/master/root/io/clones/Makefile#L30

Now it is done via shared library.

pcanal: Yes.  That is a description of the change that I noticed.   The question is whether this difference is important or not important to the test.

roottest/root/io/compression/CMakeLists.txt
linev: Global target name also name of library in current directory.
My idea that target name can be any while library name can be just `libEvent`. 
But as usual - Windows does not like such idea. 
I cannot convince it to use custom name for `libEvent.lib` file - it always using target name. 
Maybe @bellenot has idea how I can change it in `ROOTTEST_GENERATE_DICTIONARY` macro.

roottest/root/io/clones/CMakeLists.txt
linev: ROOT fails already when trying to generate dictionary with errors:
```
Generating libRtObjDict.cxx, libRtObjDict_rdict.pcm
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
Warning: FindScope got an invalid tag decl
ERROR: FindScope about to return an invalid decl
```

linev: Seems to be I found a reason for this failure - it missing TNamed.h includes

linev: After include `TNamed.h` test works: https://github.com/root-project/root/pull/18973/commits/ee3c6a9f12e8f18b77f7cb54500b1e169c2ba51c

I also linking now directly dictionary source - not library

pcanal: > ROOT fails already when trying to generate dictionary with errors:

This is a real bug (that the error message is not understandable) and should be reported so that the Cling team can fix it.  This is independent of the test that indeed needs to be fixed by itself.

Issue #18951, opened 2025-06-03
[gui] regression in TGNumberEntry


ferdymercury:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

After (probably) https://github.com/root-project/root/pull/18435 it seems that there are weird numbers in TGNumberEntry, such as:

1.00003433334848.33434803330030333333333333375555555250555345555573333331003333333333365155155-541000001033333300333333-0333333,6900333333000.18-15595255555555555055558333333000800.008000-03,-.,233333388033333355000,,,-733333333333357633333352233333313133

### Reproducer

` root --web=off $ROOTSYS/tutorials/visualisation/geom/building.C`

In 6.36.000 it works fine:
`root --web=off $ROOTSYS/tutorials/geom/building.C`

### ROOT version

```
ROOT v6.37.01
Built for linuxx8664gcc on Jun 03 2025, 11:02:14
From heads/master@v6-29-01-15845-gb690e5c5f91
With g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Binary directory: /home/ferhue/builds/build-root_src-Desktop-Debug/bin
```

### Installation method

From source

### Operating system

Ubuntu 22

PR #18952, opened 2025-06-03
[gui] stop copying when finding a null terminator char

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/18951

## Checklist:

- [x] tested changes locally

### DIFF
--- a/gui/gui/src/TGNumberEntry.cxx
+++ b/gui/gui/src/TGNumberEntry.cxx
@@ -233,14 +233,15 @@ static Bool_t IsGoodChar(char c, TGNumberFormat::EStyle style,
 
 ////////////////////////////////////////////////////////////////////////////////
 
-static void CopyAndEliminateGarbage(char *dst,
+static void CopyAndSkipGarbage(char *dst,
                                     std::size_t dstCap,
                                     const char *src,
                                     TGNumberFormat::EStyle style,
                                     TGNumberFormat::EAttribute attr)
 {
+   assert(dstCap > 0); 
    std::size_t dstIdx = 0;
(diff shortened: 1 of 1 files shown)


gui/gui/src/TGNumberEntry.cxx
```diff
@@ -232,15 +232,26 @@ static Bool_t IsGoodChar(char c, TGNumberFormat::EStyle style,
...
+static void CopyAndSkipGarbage(char *dst,
                                     std::size_t dstCap,
                                     const char *src,
                                     TGNumberFormat::EStyle style,
                                     TGNumberFormat::EAttribute attr)
 {
+   assert(dstCap > 0); 
    std::size_t dstIdx = 0;
-   while (dstIdx < dstCap - 1 && src) {
+   while (dstIdx < dstCap - 1 && src && ((*src) != 0)) {
```
silverweed: This `&& src` was supposed to be a `&& *src`:
```suggestion
   assert(src);
   while (dstIdx < dstCap - 1 && *src) {
```

silverweed (changes requested):
Thanks for spotting this.

gui/gui/src/TGNumberEntry.cxx
silverweed: Not 100% sure about `assert(src)`... maybe it could be `if (!src) return;` (not sure it `text` can be null in the caller)

ferdymercury: Good point! Thanks.

silverweed (approved):

PR #18954, opened 2025-06-03
[core] do not append objects to TFiles opened in READ mode

ferdymercury:
# This Pull request:

## Changes or fixes:

see https://root-forum.cern.ch/t/how-to-get-the-string-branch-created-in-geant4/63662/4?u=ferhue


linev:
It is very fundamental 'feature' of ROOT.
I do not think that we can change ROOT behavior  this way.

ferdymercury:
Thanks, makes sense, I'll close it

PR #18959, opened 2025-06-04
[ntuple] Add `RNTupleProcessor::PrintStructure`

enirolf:
`PrintStructure` graphically shows how different RNTuples are composed, which can be useful for, among others, debugging purposes.

For example, a processor representing a join between a single primary ntuple and a chain of two auxiliary ntuples will be represented as follows:

```
+-----------------------------+ +-----------------------------+
| ntuple                      | | ntuple_aux                  |
| ntuple.root                 | | ntuple_aux1.root            |
+-----------------------------+ +-----------------------------+
                                +-----------------------------+
                                | ntuple_aux                  |
                                | ntuple_aux2.root            |
                                +-----------------------------+
```
More example outputs can be found in the unit tests.

### DIFF
--- a/tree/ntuple/src/RNTupleProcessor.cxx
+++ b/tree/ntuple/src/RNTupleProcessor.cxx
@@ -18,8 +18,11 @@
 #include <ROOT/RFieldBase.hxx>
 #include <ROOT/RNTuple.hxx>
 #include <ROOT/RPageStorageFile.hxx>
+#include <ROOT/StringUtils.hxx>
 
 #include <TDirectory.h>
+
+#include <iomanip>
 
 std::unique_ptr<ROOT::Internal::RPageSource> ROOT::Experimental::RNTupleOpenSpec::CreatePageSource() const
 {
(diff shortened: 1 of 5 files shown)


tree/ntuple/src/RNTupleProcessor.cxx
```diff
@@ -541,3 +581,27 @@ void ROOT::Experimental::RNTupleJoinProcessor::AddEntriesToJoinTable(Internal::R
...
    fPrimaryProcessor->AddEntriesToJoinTable(joinTable, entryOffset);
 }
+
+std::string ROOT::Experimental::RNTupleJoinProcessor::GetStructureString() const
+{
+   const auto primaryStructure = ROOT::Split(fPrimaryProcessor->GetStructureString(), "\n", /*skipEmpty=*/true);
+   const auto primaryStructureWidth = primaryStructure.front().size();
+   const auto auxStructure = ROOT::Split(fAuxiliaryProcessor->GetStructureString(), "\n", /*skipEmpty=*/true);
+
+   std::ostringstream structure;
```
silverweed: Since inside here we are building the string through a stream, only to pass it back to another stream in `PrintStructure`, I'm wondering if the reversed structure would be a bit cleaner:

```c++
public:
   virtual void PrintStructure(std::ostream &output = std::cout) const {
       // ... do what GetStructureString() is doing now but using `output` instead of the `ostringstream` ...
   }

   std::string GetStructureString() const {
     std::ostringstream structure;
     PrintStructure(structure);
     return structure.str();
   }
```

Not 100% sure it's better, but to be considered.

enirolf: So I'm not sure I 100% understood what you mean, but it was enough of an inspiration to refactor everything a bit and make smarter use of the stream :-). Let me know if you think this is an improvement or not!

silverweed: Thanks, that's basically what I meant :)
I just kept the `GetStructureString()` as a helper, but that's probably not needed as it only saves one or two lines.

PR #18966, opened 2025-06-04
[ntuple] Support creating ntuples within directories in RNTupleImporter

### LINKED ISSUE #18965: Support creating an RNTuple inside a directory with the RNTupleImporter

Currently, the RNTupleImporter does not allow creating an RNTuple inside a directory within the destination file. This makes it less straightforward to preserve the structure of a file containing multiple directories with multiple TTrees during the conversion.

If the specified fNTupleName is "dir/subdir/TupleName", the importer should create the necessary directories and place the RNTuple inside.

### ROOT version

6.36

### Installation method

all

### Operating system

all

### Additional context

_No response_

siliataider:
# This Pull request:

## Changes or fixes:
This PR extends the functionality of the `RNTupleImporter` to allow creating ntuples within directories in the destination file during the conversion. 

```
auto importer = RNTupleImporter::Create(treeFileName, "Tuple/DecayTree", ntupleFileName);
```
-> This would create the directory `Tuple` if it doesn't already exists in the destination file and place the ntuple `DecayTree` inside it.

## Checklist:

- [x] tested changes locally

This PR fixes #18965

### DIFF
--- a/tree/ntupleutil/v7/test/ntuple_importer.cxx
+++ b/tree/ntupleutil/v7/test/ntuple_importer.cxx
@@ -35,6 +35,28 @@ TEST(RNTupleImporter, Empty)
    EXPECT_THROW(importer->Import(), ROOT::RException);
 }
 
+TEST(RNTupleImporter, CreateInDirectory)
+{
+   FileRaii fileGuard("test_ntuple_importer_create_in_directory.root");
+   {
+      std::unique_ptr<TFile> file(TFile::Open(fileGuard.GetPath().c_str(), "RECREATE"));
+      TDirectory *dir = file->mkdir("dir1");
+      ASSERT_NE(dir, nullptr);
+      file->cd("dir1");
+      auto tree = std::make_unique<TTree>("tree", "");
+      tree->Write();
+   }
+
+   auto importer = RNTupleImporter::Create(fileGuard.GetPath(), "dir1/tree", fileGuard.GetPath());
+   importer->SetIsQuiet(true);
+   EXPECT_THROW(importer->Import(), ROOT::RException);
+   importer->SetNTupleName("dir2/ntuple");
+   importer->Import();
+   auto reader = RNTupleReader::Open("dir2/ntuple", fileGuard.GetPath());
+   EXPECT_EQ(0U, reader->GetNEntries());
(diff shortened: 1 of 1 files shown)


dpiparo:
thanks for these changes. Is it easy to add a test?

dpiparo:
very nice. Thanks.

tree/ntupleutil/v7/src/RNTupleImporter.cxx
```diff
@@ -371,13 +372,29 @@ ROOT::RResult<void> ROOT::Experimental::RNTupleImporter::PrepareSchema()
 
 void ROOT::Experimental::RNTupleImporter::Import()
 {
-   if (fDestFile->FindKey(fNTupleName.c_str()) != nullptr)
+   TDirectory *targetDir = fDestFile.get();
+   auto parts = ROOT::Split(fNTupleName, "/", true);
+   if (parts.empty()) {
+      throw RException(R__FAIL("treeName cannot be empty"));
+   }
+   std::string treeName = parts.back();
```
silverweed: Rather than splitting and then rejoining the string you should be able to simply get the last index of `'/'` (using `strrchr`) and operate on the substrings you get from it, avoiding string copies.

silverweed (changes requested):

tree/ntupleutil/v7/src/RNTupleImporter.cxx
```diff
@@ -371,13 +371,27 @@ ROOT::RResult<void> ROOT::Experimental::RNTupleImporter::PrepareSchema()
...
+   std::string treeName = fNTupleName;
+   if (lastSlash != std::string::npos) {
+      // Create the directory if it does not exist
+      auto dirName = fNTupleName.substr(0, lastSlash);
+      treeName = fNTupleName.substr(lastSlash + 1);
+      targetDir = fDestFile->mkdir(dirName.c_str(), "", true);
+      if (!targetDir) {
+         throw RException(R__FAIL("Failed to create directory " + dirName + " in file " + fDestFileName));
+      }
+   }
```
enirolf: It would be nice to have a bit of documentation explaining what exactly this snippet of code is doing and why we need it.

PR #18925, opened 2025-06-01
[meta] Correctly treat data member types containing functions

dpiparo:
This commit fixes #18833, by reverting 0f8670. 
It also honours #18842, without breaking DD4HEP.
Moreover it adds back a test and enhances it.

PR #18968, opened 2025-06-04
[636][meta] Correctly treat data member types containing functions

dpiparo:
BP of https://github.com/root-project/root/pull/18925 but for the root part only (roottest in the sibling PR https://github.com/root-project/roottest/pull/1328)


pcanal:
Something seems odd with the merging/rebasing .... (all CI jobs failed)

PR #18973, opened 2025-06-05
[roottest] adjust tests modified in PR #18950

linev:
Follow up of https://github.com/root-project/root/pull/18950

Resolve problem with `io/customCollection`. 

Fix and enable `io/clones` test

Use cmake for ` io/customStreamer `


roottest/root/io/clones/RtObj.h
```diff
@@ -1,4 +1,4 @@
-#include <TObject.h>
+#include <TNamed.h>
```
linev, line 1: @pcanal 
I create issue https://github.com/root-project/root/issues/18982 reporting problem with dictionary generation

pcanal: Thanks.

pcanal (approved):
Thanks.

PR #18976, opened 2025-06-05
[roottest] using cmake for `io/datamodelevolution` tests

linev:
There are 5 sub-dirs, each handled by separate commit.

1. `00` got minimal changes

2. `cms-00` and `01` using generic object dump and compare content when object created and after object restored from the TTree. `compare.sh` shell script was used before to show difference between produced dump files. While `diff` is not directly available on Windows - I just analyze content difference directly in macro and print extra output if content will differ. These tests with ref files were significantly changed because of that.

3. `misc` test I used `${CMAKE_SOURCE_DIR}/roottest/scripts/build.C` script to generate class library for 
specified source - as in old `Makefile` was done. To use created libs, I add extra argument to ROOTTEST_ADD_TEST
```
ROOTEXE_OPTS -e "gROOT->ProcessLine(\".L MyClass.cxx+\")" -e "gROOT->ProcessLine(\".L SomeClass.cxx+\")"
```
It does similar as `CALLROOTEXE += -e 'gROOT->ProcessLine(".L MyClass.cxx+");'` in old `Makefile`.
The only problem - semicolon `;` at the end of ProcessLine does not work while `cmake` uses it also for list separation. 

Solution - call `-e "(void)gROOT->ProcessLine(\".L MyClass.cxx+\")"`!

4. In `stl` directory `libcond` build first. Several warnings are fixed there. And then writing and reading executed. Ref file were minimally adjusted.

Most interesting if all works on Windows


roottest/root/io/datamodelevolution/cms-00/CMakeLists.txt
```diff
@@ -1,7 +1,107 @@
...
+
+ROOTTEST_ADD_TEST(read_reflex_test1
+                  MACRO read_test1.cxx+
+                  MACROARG "\"r\""
+                  OUTREF read_reflex_test1.ref
+                  LABELS longtest
+                  FIXTURES_REQUIRED io-datamodelevolution-cms-00-reflex-test1 io-datamodelevolution-cms-00-read_test1)
+
+# test2 and test3 does not work at all
+if(broken_tests)
```
pcanal: The failure marker was clearer in the Makefile:
```
ifneq ($(ClingWorkAroundNoPrivateClassIO),)
```

roottest/root/io/datamodelevolution/cms-00/CMakeLists.txt
```diff
@@ -1,7 +1,107 @@
...
+ROOTTEST_ADD_TEST(reflex_test2
+                  MACRO test2.cxx+
+                  MACROARG "\"r\""
+                  OUTREF reflex_test2.ref
+                  WILLFAIL
+                  LABELS longtest
+                  FIXTURES_REQUIRED libDataModelV2_dictrflx_fixture io-datamodelevolution-cms-00-reflex-test1 io-datamodelevolution-cms-00-test2)
+
+ROOTTEST_ADD_TEST(test3
+                  MACRO test3.cxx+
```
pcanal: Note: when/if needed, we can split the run and compile time by using the `ROOTTEST_COMPILE_MACRO`  (This has nothing to do with this line in particular, just a random recollection :) ).

roottest/root/io/datamodelevolution/cms-00/test1.cxx
```diff
@@ -30,105 +32,99 @@ int test1(const char *mode = "") {
...
-   ofstream o7( TString::Format("../logs/cms-00/%stest07_wv1.log",prefix) );
-   ofstream o8( TString::Format("../logs/cms-00/%stest08_wv1.log",prefix) );
-   ofstream o9( TString::Format("../logs/cms-00/%stest09_wv1.log",prefix) );
-   ofstream o10( TString::Format("../logs/cms-00/%stest10_wv1.log",prefix) );
-   ofstream o11( TString::Format("../logs/cms-00/%stest11_wv1.log",prefix) );
-   ofstream o12( TString::Format("../logs/cms-00/%stest12_wv1.log",prefix) );
-   ofstream o13( TString::Format("../logs/cms-00/%stest13_wv1.log",prefix) );
-   ofstream o14( TString::Format("../logs/cms-00/%stest14_wv1.log",prefix) );
-   //ofstream o15( TString::Format("../logs/cms-00/%stest11_wv1.log",prefix) );
+   //ofstream o1( TString::Format("%stest01_wv1.log",prefix) );
```
pcanal, line 52: Is this replaced by a loop somewhere?

linev: No, I modify `dump` function - replacing by `test_dump` which creates output file inside. 
Plus `test_dump` can directly compare created file with output file created by other macro - and print normal message or error output if compare fails.

roottest/root/io/datamodelevolution/cms-00/CMakeLists.txt
linev: It was missing in cmake-based tests, now I add it with correspondent comments.
By the way - there are several such defines in `RoottestCTest.cmake` which are disabled in makefiles?

roottest/root/io/datamodelevolution/stl/CMakeLists.txt
```diff
@@ -1,7 +1,32 @@
...
+                  FIXTURES_REQUIRED libDataModelCond_fixture)
+
+ROOTTEST_ADD_TEST(WriteFile
+                  MACRO execWriteFile.C
+                  OUTREF execWriteFile.ref
+                  LABELS longtest
+                  FIXTURES_SETUP WriteFile_fixture)
+
+ROOTTEST_ADD_TEST(ReadFile
+                  MACRO execReadFile.C
```
pcanal, line 36: Consider using `ROOTTEST_COMPILE_MACRO` to make the compilation of `readFile.C` and `writeFile.C` a separate test (it was in the Makefile).

roottest/root/io/datamodelevolution/01/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+                  OUTREF test3.ref
+                  LABELS longtest
+                  FIXTURES_REQUIRED libDatModelV2_dictcint_fixture io-datamodelevolution-01-test1
+                  FIXTURES_SETUP io-datamodelevolution-01-test3)
+
+ROOTTEST_ADD_TEST(rtest3
+                  MACRO test3.cxx+
+                  MACROARG "\"r\""
+                  OUTREF rtest3.ref
+                  WILLFAIL
```
pcanal: If I read correctly the Makefile, this used to work.  What changed?

roottest/root/io/datamodelevolution/01/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+                  LABELS longtest
+                  FIXTURES_REQUIRED libDatModelV2_dictrflx_fixture io-datamodelevolution-01-rtest1 io-datamodelevolution-01-test3)
```
pcanal:
```
FIXTURES_REQUIRED libDatModelV2_dictrflx_fixture io-datamodelevolution-01-rtest2 io-datamodelevolution-01-test3)
```

In the Makefile the test depends on `rtest2` rather than `rtest1`.

pcanal: Also having the compilation of the script as a separate tests allowed test3 and rtest3 to run in parallel.

roottest/root/io/datamodelevolution/misc/readTree.C
```diff
@@ -3,9 +3,9 @@ void readTree(const char* filename = "treeTest.root", bool loadSomeClass = true)
 {
 #ifndef ClingWorkAroundMissingDynamicScope
    gROOT->ProcessLine(".L MyClass.cxx+");
-#endif
 	if (loadSomeClass) gROOT->ProcessLine(".L SomeClass.cxx+");
-	
+#endif
```
pcanal, line 6: The previous #endif location was intentional.  The script does not use `SomeClass` explicitly in its code and thus the library can be loaded during the script without issues.

roottest/root/io/datamodelevolution/misc/CMakeLists.txt
```diff
@@ -1,7 +1,54 @@
...
+                  MACRO writeTree.C
+                  ROOTEXE_OPTS -e "(void)gROOT->ProcessLine(\".L MyClass.cxx+\")" -e "(void)gROOT->ProcessLine(\".L SomeClass.cxx+\")"
+                  OUTREF writeTree.ref
+                  LABELS longtest
+                  FIXTURES_REQUIRED BuildMyClass-fixture BuildSomeClass-fixture
+                  FIXTURES_SETUP writeTree)
+
+ROOTTEST_ADD_TEST(postobj
+                  MACRO runpostobj.C
+                  ROOTEXE_OPTS -e "(void)gROOT->ProcessLine(\".L MyClass.cxx+\")" -e "(void)gROOT->ProcessLine(\".L SomeClass.cxx+\")"
```
pcanal: The second -e is not necessary (with previous code) especially

roottest/root/io/datamodelevolution/misc/CMakeLists.txt
```diff
@@ -1,7 +1,54 @@
...
+                  LABELS longtest
+                  FIXTURES_REQUIRED WriteOrderClass)
+
+ROOTTEST_ADD_TEST(BuildMyClass
+                  MACRO ${CMAKE_SOURCE_DIR}/roottest/scripts/build.C
+                  MACROARG "\"MyClass.cxx\""
+                  COPY_TO_BUILDDIR MyClass.h MyClass.cxx MyClass_linkdef.h
+                  FIXTURES_SETUP BuildMyClass-fixture)
+
+ROOTTEST_ADD_TEST(BuildSomeClass
```
pcanal: Is this better than `ROOTTEST_COMPILE_MACRO` ?

roottest/root/io/datamodelevolution/cms-00/CMakeLists.txt
pcanal: Yes, the `ClingWorkAroundXYZ` are meant to be help test work around deficiencies we were hoping would be temporary when we migrated from CINT to Cling.  The same definition should be in both the CMake and Makefile systems.

roottest/root/io/datamodelevolution/01/CMakeLists.txt
linev: First of  all - `01` dir was excluded from testing. It simply not working with Makefile - because of many reasons
I trying to reintegrate it. And only test3 in reflex mode fails.

Compare to cling-generated dictionary, with reflex there are extra error messages when trying to read file. These are:
```
Warning in <TClass::Init>: no dictionary for class ClassAIns is available
Warning in <TClass::Init>: no dictionary for class ClassA is available
Warning in <TClass::Init>: no dictionary for class ClassABase is available
Error in <TTree::SetBranchAddress>: The pointer type given (ClassAIns2) does not correspond to the class needed (ClassAIns) by the branch: TestAIns
Warning in <TStreamerInfo::BuildOld>: Cannot convert ClassD::m_d from type: ClassAIns to type: ClassAIns2, skip element
Error in <TBufferFile::ReadClass>: The on-file class is "ClassB" which is not compatible with the requested class: "ClassA"
Error in <TBufferFile::ReadObject>: got object of wrong class! requested ClassA but got ClassB
Error in <TBufferFile::ReadClass>: The on-file class is "ClassB" which is not compatible with the requested class: "ClassA"
Error in <TBufferFile::ReadObject>: got object of wrong class! requested ClassA but got ClassB
Error in <TBufferFile::CheckByteCount>: object of class vector<ClassB> read too few bytes: 104 instead of 332
Error in <TBufferFile::CheckByteCount>: object of class vector<ClassC> read too few bytes: 208 instead of 288
```
As a result, read data does not match to original. 

I attach log file from `test3.cxx+("r")` run.
[rtest3.log](https://github.com/user-attachments/files/20668081/rtest3.log)

roottest/root/io/datamodelevolution/01/CMakeLists.txt
linev: `rtest2` and `rtest3` reading data produced by `rtest1`.
One can do also `rtest3` -> `rtest2` -> `rtest1` - but why?

That should happen with `gROOT->ProcessLine("ROOT :: Cintex :: Cintex :: Enable();");`?
There is no `Cintex`  namespace or class in ROOT sources.

roottest/root/io/datamodelevolution/misc/CMakeLists.txt
linev: It is required. One only can run root as:
```
root -e 'gROOT->ProcessLine("int i = 1")' -e 'gROOT->ProcessLine("i += 3")'                                                                                                                                                                            
```

Without second `-e` it does not work.

roottest/root/io/datamodelevolution/misc/CMakeLists.txt
linev: Yes, I replace it here

roottest/root/io/datamodelevolution/misc/readTree.C
linev: I move `#endif` on previous place and add extra line to postobj.ref file:
```
Info in <ACLiC>: unmodified script has already been compiled and loaded
```

roottest/root/io/datamodelevolution/stl/CMakeLists.txt
linev: Yes, I did this.

roottest/root/io/datamodelevolution/misc/CMakeLists.txt
pcanal: The previous version of the code did not need to have it.  What changed?

roottest/root/io/datamodelevolution/01/CMakeLists.txt
pcanal: The `DataModelV2_selection.xml` is missing the translation of the I/O customization rules:
```
#pragma read sourceClass="ClassA" version="[2]" targetClass="ClassA2" source="ClassABase" target="ClassABase2"
#pragma read sourceClass="ClassAIns" version="[2]" targetClass="ClassAIns2"
#pragma read sourceClass="ClassABase" version="[2]" targetClass="ClassABase2"
#pragma read sourceClass="ClassC" version="[2]" targetClass="ClassC" source="ClassABase" target="ClassABase2"
#pragma read sourceClass="ClassB" version="[2]" targetClass="ClassB" source="ClassA" target="ClassA2"
```

roottest/root/io/datamodelevolution/01/CMakeLists.txt
pcanal: Cintex was removed when we move to Cling.  This is now a no-op.

pcanal: > rtest2 and rtest3 reading data produced by rtest1.

I missed that :)

roottest/root/io/datamodelevolution/misc/readTree.C
pcanal: Once you move the `#endif`, you can remove the equivalent `-e` (so that the library is loaded only once)

roottest/root/io/datamodelevolution/misc/CMakeLists.txt
linev: https://github.com/root-project/roottest/blob/master/root/io/datamodelevolution/misc/Makefile#L69

Makefile has such feature, but maybe I need to add `if(ClingWorkAroundMissingDynamicScope)`

roottest/root/io/datamodelevolution/misc/readTree.C
linev: Will do

roottest/root/io/datamodelevolution/01/CMakeLists.txt
linev: Yes, adding such rules fixes rtest3!

linev (commented):
I add missing rules to `DataModelV2_selection.xml` and
provide `if(ClingWorkAroundMissingDynamicScope)` to
load macros with ROOTEXE_OPTS.

roottest/root/io/datamodelevolution/misc/CMakeLists.txt
```diff
@@ -1,7 +1,59 @@
...
+
+ROOTTEST_COMPILE_MACRO(MyClass.cxx
+                  FIXTURES_SETUP root-io-datamodelevolution-misc-BuildMyClass-fixture)
+
+ROOTTEST_COMPILE_MACRO(SomeClass.cxx
+                  FIXTURES_SETUP root-io-datamodelevolution-misc-BuildSomeClass-fixture)
+
+if(ClingWorkAroundMissingDynamicScope)
+  set(_myclass_load -e "(void)gROOT->ProcessLine(\".L MyClass.cxx+\")")
+  set(_someclass_load -e "(void)gROOT->ProcessLine(\".L SomeClass.cxx+\")")
```
pcanal, line 43:
```suggestion
```
This (`_someclass_load) shouldn't be needed now that the script (as before) does it unconditionally.

Ref: https://github.com/root-project/root/pull/18976#discussion_r2132775088
Ref: https://github.com/root-project/root/pull/18976/files#r2132770117

roottest/root/io/datamodelevolution/misc/CMakeLists.txt
pcanal: This is moved to https://github.com/root-project/root/pull/18976/files#r2146039074

roottest/root/io/datamodelevolution/common/Dumper.h
```diff
@@ -275,7 +253,118 @@ void dump( void* obj, TClass* cl, std::ostream& out, std :: string prefix = "" )
...
+//------------------------------------------------------------------------------
+template <typename Type>
+void test_dump( Type* obj, const char *prefix, int testid, const char *suffix0, const char *suffix = nullptr, int correct = 0)
+{
+   const std::type_info& ti( typeid( Type ) );
+   TClass*  cl         = TClass :: GetClass( ti );
+
+   if( !cl )
+   {
+      std :: cout << "[!] Could not get dictionry for class: ";
```
pcanal:
```suggestion
      std :: cout << "[!] Could not get dictionary for class: ";
```

roottest/root/io/datamodelevolution/misc/CMakeLists.txt
linev: `writeTree.C` is not changed and still have such ifdef:
```
#ifndef ClingWorkAroundMissingDynamicScope
   gROOT->ProcessLine(".L MyClass.cxx+");
   gROOT->ProcessLine(".L SomeClass.cxx+");
#endif
```
So `MyClass.cxx` and `SomeClass.cxx` handled in the same way

roottest/root/io/datamodelevolution/common/Dumper.h
linev: There are several such places, I will add commit directly

roottest/root/io/datamodelevolution/misc/CMakeLists.txt
pcanal: You are right.  I had mis-read the Makefile and missed some inner code in the read script.

roottest/root/io/datamodelevolution/misc/CMakeLists.txt
```diff
@@ -1,7 +1,59 @@
...
+                  COMPILE_OPTIONS ${warnopt}
+                  NO_ROOTMAP)
+
+ROOTTEST_ADD_TEST(Conversions
+                  MACRO execConversions.cxx+
+                  OUTREF execConversions.ref
+                  LABELS longtest)
+
+ROOTTEST_ADD_TEST(WriteOrderClass
+                  MACRO execWriteOrderClass.cxx+
```
pcanal, line 24: In part because the build is part of the run, the error seen at https://github.com/root-project/root/pull/18976/checks?check_run_id=44212289416 is related.

We can solve the issue with any of:
1. Use `-Wsuggest-override`
2. Split the compilation from the execution
3. Update the `OrderClass_v1.*h` to use the `override` keyword and `ClassDefOverride`

Probably the best is to do both 2 and 3.

linev: I use override syntax

roottest/root/io/datamodelevolution/01/DataModelV1_selection.xml
```diff
@@ -5,15 +5,15 @@
...
-  <class name="std::vector<std::pair<int,double> >" />
-  <class name="std::vector<ClassA>" />
-  <class name="std::vector<ClassA*>" />
-  <class name="std::vector<ClassB>" />
-  <class name="std::vector<ClassB*>" />
-  <class name="std::vector<ClassC>" />
-  <class name="std::vector<ClassC*>" />
-  <class name="std::vector<ClassD>" />
-  <class name="std::vector<ClassD*>" />
+  <class name="std::vector&lt;double&gt;" />
```
pcanal, line 8: I missed this change.  Do we really need to escape special character within the quotes?

linev: There was Windows nodes which complained about XML parsing.
And I had suspect that it caused by such special characters.

After many other changes XML parsing failure disappeared - one can try to revert these changes.

pcanal: > one can try to revert these changes.

Please and thanks :)

linev: https://github.com/root-project/root/pull/19074

Yes, it works now also on Windows

PR #18978, opened 2025-06-05
[Minuit2] Only compute global correlation matrix when needed

guitargeek:
The function `double ROOT::Math::Minimizer::GlobalCC(unsigned int ivar)` was changed to `std::vector<double> ROOT::Math::Minimizer::GlobalCC()`, always returning the full list of global correlations.

This change was needed because global correlations are not unconditionally computed and cached in the minimizer anymore. Only computing them when calling `GlobalCC()` avoids unneeded overhead when global correlations are not required.

### DIFF
--- a/README/ReleaseNotes/v638/index.md
+++ b/README/ReleaseNotes/v638/index.md
@@ -62,6 +62,11 @@ The following people have contributed to this new version:
 
 ## Math
 
+### Minimizer interface
+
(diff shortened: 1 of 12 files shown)


lmoneta (approved):
LGTM!
The changes seems correct, this would require the computation of the GlobalCC, which requires inverting a matrix, only when requested. It makes then sense to change also the interface to return the full vector of global correlations.

PR #18990, opened 2025-06-07
[cmake][pyroot] Fix install directory on Windows

bellenot:


dpiparo:
Could a reference to this commit be added in the commit message and PR comment so that we remember what happened? https://github.com/root-project/root/commit/9bb2f5076762c4171bfa71f44da3ee74fbd27b5b
@guitargeek is there a way in which we can preserve the nice improvements of github.com/root-project/root/pull/18969 and not break install on macos and win perhaps?

guitargeek:
We have zero test coverage for the install tree, so I guess any changes affect it will have these incremental fixups :(

But good that you fixed it for Windows now! I have not seen the problem on macOS though, do you have an example error message and build log?

I think eventually we really need some post install tests, like suggested by Stephan:
https://github.com/root-project/root/pull/17956

Just checking if some basic ROOT macro and `import ROOT` in Python works would have already helped here!

dpiparo:
You really have a point. Something relatively inexpensive could be done for PRs too...

PR #18991, opened 2025-06-07
Save memory. Do not link to LLVM libraries in parallel

ellert:
# This Pull request:

Linking to the built-in static LLVM libraries uses a lot ov memory. Linking more than one such target in parallel can exhaust the memory. This PR adds dependencies to prevent this.

Compare

https://github.com/root-project/root/blob/a32e7d0eb6ff583c3420e32b4198d20cd7e8ebd4/core/metacling/src/CMakeLists.txt#L126-L127

### DIFF
--- a/core/clingutils/test/CMakeLists.txt
+++ b/core/clingutils/test/CMakeLists.txt
@@ -31,3 +31,6 @@ if(NOT builtin_clang)
 endif()
 
 ROOT_ADD_UNITTEST_DIR(Core RIO ${CLING_LIBRARIES} $<TARGET_OBJECTS:ClingUtils>)
+
+# Save memory. Do not link to LLVM libraries in parallel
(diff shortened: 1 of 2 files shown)


dpiparo:
Thanks for these changes. How much memory is consumed in the worst case? Is some potential for parallelism lost at the risk of some potential reduction of build time?

hahnjo:
These dependency hacks are not really great for targeted rebuilds; for example the current hack already results in `ninja Cling` (which I do quite often) rebuilding also `rootcling_stage1`. If it's possible to use `ninja`, we could look into [`JOB_POOLS`](https://cmake.org/cmake/help/latest/prop_gbl/JOB_POOLS.html) in the way that LLVM provides `LLVM_PARALLEL_LINK_JOBS`.

PR #18993, opened 2025-06-07
Fix test for 32 bit architectures

ellert:
The size of the struct isdifferent on 32 bit architectures.
```
[ RUN      ] TypeReflectionTest.GetSizeOfType
/builddir/build/BUILD/root-6.36.00-build/root-6.36.00/interpreter/CppInterOp/unittests/CppInterOp/TypeReflectionTest.cpp:79: Failure
  Cpp::GetSizeOfType(Cpp::GetVariableType(Decls[4]))
    Which is: 12
  16
[  FAILED  ] TypeReflectionTest.GetSizeOfType (23 ms)

[ RUN      ] ScopeReflectionTest.SizeOf
/builddir/build/BUILD/root-6.36.00-build/root-6.36.00/interpreter/CppInterOp/unittests/CppInterOp/ScopeReflectionTest.cpp:161: Failure
  Cpp::SizeOf(Decls[7])
    Which is: 12
  (size_t)16
    Which is: 16
[  FAILED  ] ScopeReflectionTest.SizeOf (26 ms)
```

### DIFF
--- a/interpreter/CppInterOp/unittests/CppInterOp/ScopeReflectionTest.cpp
+++ b/interpreter/CppInterOp/unittests/CppInterOp/ScopeReflectionTest.cpp
@@ -158,7 +158,8 @@ TEST(ScopeReflectionTest, SizeOf) {
   EXPECT_EQ(Cpp::SizeOf(Decls[4]), (size_t)0);
   EXPECT_EQ(Cpp::SizeOf(Decls[5]), (size_t)1);
   EXPECT_EQ(Cpp::SizeOf(Decls[6]), (size_t)4);
-  EXPECT_EQ(Cpp::SizeOf(Decls[7]), (size_t)16);
+  struct B {short a; double b;};
+  EXPECT_EQ(Cpp::SizeOf(Decls[7]), sizeof(B));
 }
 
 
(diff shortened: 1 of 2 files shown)


guitargeek:
This needs to be reviewed by @Vipul-Cariappa or @aaronj0. Maybe the same change should be applied upstream?

interpreter/CppInterOp/unittests/CppInterOp/ScopeReflectionTest.cpp
```diff
@@ -158,7 +158,7 @@ TEST(ScopeReflectionTest, SizeOf) {
   EXPECT_EQ(Cpp::SizeOf(Decls[4]), (size_t)0);
   EXPECT_EQ(Cpp::SizeOf(Decls[5]), (size_t)1);
   EXPECT_EQ(Cpp::SizeOf(Decls[6]), (size_t)4);
-  EXPECT_EQ(Cpp::SizeOf(Decls[7]), (size_t)16);
+  EXPECT_EQ(Cpp::SizeOf(Decls[7]), (size_t) (sizeof(size_t) == 4 ? 12 : 16));
```
aaronj0: Based on the CI failure, I believe we also need to account for 32-bit MSVC platforms since the double is 8-byte aligned.

Perhaps an approach like: 
```suggestion
  struct B {short a; double b;};
  EXPECT_EQ(Cpp::SizeOf(Decls[7]), sizeof(B));
```
should always reflect the correct platform ABI

ellert: I have implemented this changes (for both cases).

aaronj0:
Hi @ellert, thanks again for this nice improvement! This commit has been upstreamed to CppInterOp: https://github.com/compiler-research/CppInterOp/pull/649 and will be pulled into ROOT along with other patches when we move to the latest tag/release. Hence, I will close this PR here and link it when we upgrade.

Issue #18768, opened 2025-05-18
Regression in i-adding array.array to std::vector in PyROOT


lwpiotr:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

In (Py)ROOT 6.30.06 one can properly += (iadd) an array.array to an std.vector. In the current master, this operation changes the type from (in my test case) cppyy.gbl.std.vector<unsigned short> to cppyy.gbl.__gnu_cxx.__normal_iterator<unsigned short*,vector<unsigned short> >, which is not compatible with the standard vector.

+= with nested list of arrays was the only way known to me to assign values to a nested std::vector with ROOT 6.30.06. It seems that now .assign(nested pure list) works. So, detecting the user's ROTO version, I can easily make my code work (however dirty the hack with the ROTO version is). Still, I wanted to report the regression and also the fact, that += nested list doesn't work (while the assign() does).

### Reproducer

```python
import ROOT
import array
d = ROOT.vector("unsigned short")()
a = array.array('H', [0, 1])
d+=a
print(type(d))
```

### ROOT version

heads/master@v6-37-01-6541-g9a9e7697e4

### Installation method

Self compilation

### Operating system

Fedora 42

PR #18999, opened 2025-06-09
[cppyy] fix iadd pythonization for np/array buffers

aaronj0:
Backport of https://github.com/wlav/CPyCppyy/pull/51
Fixes https://github.com/root-project/root/issues/18768

Currently, the `iadd` implementation for `std::vector` in `Pythonize.cxx` works perfectly with Python lists and iterables for which we have an `ItemGetter`. But the code path for numpy/array buffers do not. The difference between `x+=y` and `x.__iadd__(y)` leads to unintuitive behaviour: if the intent is to obtain the resulting iterator when calling insert, the user can do `d.insert(d.end(), a)` which makes more sense than expecting it from `d+=a`. When the user does call `d+=a`, since `__iadd__` is overriden, operation does not end with the call to `VectorIAdd` which should just return the iterator, and reflect the change by insertion on `d`. Instead, it ends with `d = d.__iadd__(a)` which reassigns `d` to the returned iterator, losing the vector in the process. This PR fixes this by returning self in our pythonization.


dpiparo:
Thanks for these changes. Can we add a (some) test(s)?

PR #19004, opened 2025-06-10
Miscellaneous changes for RDF Snapshot test

vepadulano:


vepadulano:
Well, the `dataframe_snapshot` failure on Mac ARM64 is due to throwing an exception when JITting. The reason it worked on my laptop is because I'm testing on my branch with Snapshot that does not need JITting :) That will soon get its own PR, I guess at least we know it should improve also this one 👍

vepadulano:
https://github.com/root-project/root/pull/19006 is a requirement for this PR

PR #19005, opened 2025-06-10
[RDF] Fix a crash when passing an invalid glob expression

hageboeck:
The old code was unconditionally accessing the first element of the list of files that the glob expression expands to. When an invalid glob is used, this was crashing root.

I'm also sneaking in an update of the ttreereader documentation that was sitting around in one of my local branches.


tree/dataframe/src/RLoopManager.cxx
```diff
@@ -1211,16 +1211,15 @@ std::unique_ptr<TFile> OpenFileWithSanityChecks(std::string_view fileNameGlob)
...
-                             ? ROOT::Internal::TreeUtils::ExpandGlob(std::string{baseName})[0] + std::string{query}
-                             : fileNameGlob};
+   std::string fileToOpen{fileNameGlob};
+   if (baseName.find_first_of("[]*?") != std::string_view::npos) { // Wildcards accepted by TChain::Add
+      const auto expanded = ROOT::Internal::TreeUtils::ExpandGlob(std::string{baseName});
+      if (expanded.empty())
+         throw std::invalid_argument{"RDataFrame: The glob expression '" + std::string{baseName} +
+                                     "' did not match any files."};
+
+      fileToOpen = expanded.front() + std::string{query};
```
vepadulano, line 1221: I see two test cases that need to be added:
* The case described in the PR description, i.e. when a wrong glob is passed
* The case when the new exception is thrown

vepadulano (changes requested):
Nice thanks! Can you also add tests?

tree/dataframe/src/RLoopManager.cxx
hageboeck: Sure, if you want tests, I can add them. It is, however, both the same test, because it will test that an exception is thrown instead of the program crashing.

vepadulano (approved):
Thank you!

PR #19006, opened 2025-06-10
Lift template requirement for TTree Snapshot

vepadulano:
These are the changes needed to avoid the requirement to call Snapshot to a TTree dataset with template arguments for the column types, or to JIT the call in case the template arguments are not given.

Still missing:
* Doing the same for RNTuple
* Deprecating the Snapshot call with the argument type list


tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -2228,6 +2247,136 @@ public:
    }
 };
 
+class R__CLING_PTRCHECK(off) UntypedSnapshotTTreeHelper final : public RActionImpl<UntypedSnapshotTTreeHelper> {
```
hageboeck, line 2250: Is there possibly a benefit in making all snapshot helpers start with "Snapshot"? This one could be SnapshotHelperTTreeUntyped, for example. Or could this in the long run just be SnapshotHelperTTree, because the others can be discontinued?
Maybe it's too much work to rename them for no benefit ...

tree/dataframe/inc/ROOT/RDF/RActionSnapshot.hxx
```diff
@@ -0,0 +1,185 @@
...
+      fValues[slot].clear();
+      fHelper.CallFinalizeTask(slot);
+   }
+
+   /// Clean-up and finalize the action result (e.g. merging slot-local results).
+   /// It invokes the helper's Finalize method.
+   void Finalize() final
+   {
+      fHelper.Finalize();
+      SetHasRun();
```
hageboeck, line 139: Would it make sense to clear also `fValues`, or is `FinalizeSlot` called also before Finalize?
In that case, only the empty vectors would remain, which is probably OK.

tree/dataframe/inc/ROOT/RDF/RLazyDSImpl.hxx
```diff
@@ -52,34 +62,10 @@ class RLazyDS final : public ROOT::RDF::RDataSource {
...
-      const auto namesIt = std::find(colBegin, colEnd, colName);
-      const auto index = std::distance(colBegin, namesIt);
-
-      Record_t ret(fNSlots);
-      for (auto slot : ROOT::TSeqU(fNSlots)) {
-         ret[slot] = fPointerHolders[index][slot]->GetPointerAddr();
-      }
-      return ret;
+      // We use the new API
+      return {};
```
hageboeck, line 68: Will this lead to surprising results? Maybe hard to trace down what's causing the error?
One could e.g.
- Remove the function altogether. This would entail rethinking the data source interface ...
- Throw a `logic_error`

tree/dataframe/src/RDFActionHelpers.cxx
```diff
@@ -364,49 +377,522 @@ void CreateCStyleArrayBranch(TTree *inputTree, TTree &outputTree, ROOT::Internal
       const auto leaflist = std::string(bname) + "[" + sizeLeafName + "]/" + rootbtype;
       // Use original basket size for existing branches and new basket size for new branches
       const auto bufSize = (basketSize > 0) ? basketSize : inputBranch->GetBasketSize();
-      auto *outputBranch =
-         outputTree.Branch(outputBranchName.c_str(), static_cast<void *>(nullptr), leaflist.c_str(), bufSize);
+      void *addressForBranch = [address]() -> void * {
+         if (address) {
+            // Address here points to a ROOT::RVec<std::byte> coming from RTreeUntypedArrayColumnReader. We know we need
+            // its buffer, so we cast it and extract the address of the buffer
+            auto *rawRVec = reinterpret_cast<ROOT::RVec<std::byte> *>(address);
```
hageboeck, line 384: It looks somewhat dangerous. Is there a chance that it will get reallocated or does some kind of max array size save us?

tree/dataframe/src/RDFActionHelpers.cxx
```diff
@@ -364,49 +377,522 @@ void CreateCStyleArrayBranch(TTree *inputTree, TTree &outputTree, ROOT::Internal
...
+      else
+         outputBranch = outputTree.Branch(outputBranchName.c_str(), classPtr->GetName(), nullptr, bufSize);
       outputBranches.Insert(outputBranchName, outputBranch);
       return;
    }
 
-   // Find if there is an input branch, check for cases where we need a leaflist (e.g. C-style arrays)
-   CreateCStyleArrayBranch(inputTree, outputTree, outputBranches, inputBranchName, outputBranchName, basketSize);
+   // We are not aware of other cases
+   throw std::runtime_error(
```
hageboeck:
```suggestion
   throw std::logic_error(
```
This might be a slightly better fit, as it signifies an error in the program logic (not necessarily user's fault).

tree/dataframe/src/RDFActionHelpers.cxx
```diff
@@ -364,49 +377,522 @@ void CreateCStyleArrayBranch(TTree *inputTree, TTree &outputTree, ROOT::Internal
...
+
+   auto *inputBranch = SearchForBranch(inputTree, inputBranchName);
+   if (auto *outputBranch = outputBranches.Get(outputBranchName); outputBranch && valueAddress) {
+      // The output branch was already created, we just need to (re)set its address
+      SetBranchAddress(inputBranch, *outputBranch, actionHelperBranchPtrAddress,
+                       outputBranches.IsCArray(outputBranchName), valueAddress);
+      return;
+   }
+
+   auto *dictionary = TDictionary::GetDictionary(valueTypeID);
```
hageboeck, line 456: To understand the performance implications: This code and below runs once at the beginning, but the above runs every time that branch addresses have to be reset, correct?

tree/dataframe/src/RDFActionHelpers.cxx
```diff
@@ -364,49 +377,522 @@ void CreateCStyleArrayBranch(TTree *inputTree, TTree &outputTree, ROOT::Internal
...
+   fOutputFile = outFile.get();
+   fMerger = std::make_unique<ROOT::TBufferMerger>(std::move(outFile));
+}
+
+void ROOT::Internal::RDF::UntypedSnapshotTTreeHelperMT::Finalize()
+{
+   // This assertion is wrong because there is a very important case where fOutputFiles is never filled: if the
+   // input dataset has no entries, the computation graph will simply not run and our InitTask function will never
+   // be called
+   // assert(std::any_of(fOutputFiles.begin(), fOutputFiles.end(), [](const auto &ptr) { return ptr != nullptr; }));
```
hageboeck: Maybe this can be removed?

tree/dataframe/src/RDFInterfaceUtils.cxx
```diff
@@ -1067,3 +1067,42 @@ void RemoveRNTupleSubFields(ColumnNames_t &columnNames)
...
+
+   if (!ds.HasColumn(colName) &&
+       lm.GetSuppressErrorsForMissingBranches().find(colName) == lm.GetSuppressErrorsForMissingBranches().end())
+      return;
+
+   const auto nSlots = lm.GetNSlots();
+   std::vector<std::unique_ptr<ROOT::Detail::RDF::RColumnReaderBase>> colReaders;
+   colReaders.reserve(nSlots);
+   // TODO consider changing the interface so we return all of these for all slots in one go
+   for (auto slot = 0u; slot < lm.GetNSlots(); ++slot)
```
hageboeck:
```suggestion
   for (auto slot = 0u; slot < nSlots; ++slot)
```

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1008,6 +1008,7 @@ TEST(RDFSnapshotMore, TClonesArray)
 // ROOT-10702
 TEST(RDFSnapshotMore, CompositeTypeWithNameClash)
 {
+   GTEST_SKIP() << "Skipping test with interpreted class without std::type_info.";
```
hageboeck: What will happen with this test? Will it be removed?

hageboeck (approved):
That was a big piece of work, thanks!

I don't see showstoppers, I added mostly questions for better understanding.

tree/dataframe/src/RDFActionHelpers.cxx
vepadulano: Yes it runs when a branch hasn't been created yet in the output TTree

tree/dataframe/test/dataframe_snapshot.cxx
vepadulano: Need to update it to have dictionary info

tree/dataframe/inc/ROOT/RDF/RLazyDSImpl.hxx
vepadulano: The idea is to completely deprecate the old `GetColumnReaders` API, but I want to leave that for a followup PR.

tree/dataframe/src/RDFActionHelpers.cxx
vepadulano: This will be stable for all nodes within the same event. I don't know about the max array size part.

PR #19007, opened 2025-06-11
[ntuple] Add RMiniFile::GetNTupleProperAtOffset

silverweed:
# This Pull request:
refactors a bit `RMIniFile::GetNTupleProper` adding a `GetNTupleProperAtLocation` subroutine to load the anchor once its location in the file is known. This will be useful for the RNTuple Attributes prototype as it will need to load the Attributes' anchor from the same file from a known location.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/inc/ROOT/RMiniFile.hxx
+++ b/tree/ntuple/inc/ROOT/RMiniFile.hxx
@@ -68,6 +68,8 @@ private:
    /// Used when the file turns out to be a TFile container. The ntuplePath variable is either the ntuple name
    /// or an ntuple name preceded by a directory (`myNtuple` or `foo/bar/myNtuple` or `/foo/bar/myNtuple`)
    RResult<RNTuple> GetNTupleProper(std::string_view ntuplePath);
+   /// Loads an RNTuple anchor from a TFile at the given file offset (unzipping it if necessary).
+   RResult<RNTuple> GetNTupleProperAtOffset(std::uint64_t keyOffset);
 
    /// Searches for a key with the given name and type in the key index of the directory starting at offsetDir.
    /// The offset points to the start of the TDirectory DATA section, without the key and without the name and title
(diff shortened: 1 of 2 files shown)


tree/ntuple/inc/ROOT/RMiniFile.hxx
```diff
@@ -68,6 +68,8 @@ private:
    /// Used when the file turns out to be a TFile container. The ntuplePath variable is either the ntuple name
    /// or an ntuple name preceded by a directory (`myNtuple` or `foo/bar/myNtuple` or `/foo/bar/myNtuple`)
    RResult<RNTuple> GetNTupleProper(std::string_view ntuplePath);
+   /// Loads an RNTuple anchor from a TFile at the given location (unzipping it if necessary).
+   RResult<RNTuple> GetNTupleProperAtLocation(std::uint64_t keyLocation);
```
enirolf: For me, "location" could also very well refer to a path. Perhaps `GetNTupleProperAtKey` instead would make the distinction more clear?

silverweed: Well, the ntuple is always "at a key" as well, so I don't think it's much clearer...but maybe using `AtSeekPos` or similar would make it more obvious that it is a seek position in the file..?

hahnjo: Why not just call it `offset`, as it's called in the specification and the previous code was already doing?

tree/ntuple/src/RMiniFile.cxx
```diff
@@ -714,46 +714,10 @@ std::uint64_t ROOT::Internal::RMiniFileReader::SearchInDirectory(std::uint64_t &
    return 0;
 }
 
-ROOT::RResult<ROOT::RNTuple> ROOT::Internal::RMiniFileReader::GetNTupleProper(std::string_view ntuplePath)
+ROOT::RResult<ROOT::RNTuple> ROOT::Internal::RMiniFileReader::GetNTupleProperAtSeekPos(std::uint64_t keySeekPos)
```
hahnjo: Maybe move the new function after `GetNTupleProper`, as in the header? (this will also reduce the diff significantly because the existing function is just split)

tree/ntuple/inc/ROOT/RMiniFile.hxx
pcanal: Indeed, in that type of context we have been using `offset`

PR #19008, opened 2025-06-11
[skip-ci][ci] Add cling and textinput sync action

dpiparo:


hahnjo:
Just to document what I wrote on Mattermost: we need the sync to be incremental over what is already there, so `git pull` works and tags are correctly present in the history.

PR #19009, opened 2025-06-11
New Method TPad::DivideRatios

couet:
This method divides the current canvas in nx by ny according to the width and height ratios.
Was inspired by the CMS style subplots script.

### DIFF
--- a/graf2d/gpad/inc/TPad.h
+++ b/graf2d/gpad/inc/TPad.h
@@ -186,6 +186,7 @@ public:
    void              CopyPixmaps() override;
    void              DeleteExec(const char *name) override;
    void              Divide(Int_t nx=1, Int_t ny=1, Float_t xmargin=0.01, Float_t ymargin=0.01, Int_t color=0) override; // *MENU*
+   void              DivideRatios(Int_t nrows, Int_t ncolumns, const std::vector<double>& width_ratios={}, const std::vector<double>& height_ratios={}, const double canvas_top_margin=0., const double canvas_left_margin=0.);
    virtual void      DivideSquare(Int_t n, Float_t xmargin=0.01, Float_t ymargin=0.01, Int_t color=0);
    void              Draw(Option_t *option="") override;
    void              DrawClassObject(const TObject *obj, Option_t *option="") override;
--- a/graf2d/gpad/src/TPad.cxx
+++ b/graf2d/gpad/src/TPad.cxx
@@ -1346,6 +1346,109 @@ void TPad::Divide(Int_t nx, Int_t ny, Float_t xmargin, Float_t ymargin, Int_t co
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+/// Divide the canvas according to ratios.
+///
+/// The current canvas is divided in nx by ny according to the width and height ratios.
+/// If the ratios are not specified they are assumed to be equal.
+///
+/// Pads are automatically named `canvasname_n` where `n` is the division number
+/// starting from top left pad.
+///
+/// Top and left margins can be defined.
+
+void TPad::DivideRatios(Int_t nx, Int_t ny,
+                        const std::vector<double>& width_ratios,
+                        const std::vector<double>& height_ratios,
+                        const double canvas_top_margin,
+                        const double canvas_left_margin
+                        )
+{
+   cd();
+
+   int wrs = width_ratios.size();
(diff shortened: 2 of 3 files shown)


graf2d/gpad/inc/TPad.h
```diff
@@ -186,6 +186,7 @@ friend class TWebCanvas;
    void              CopyPixmaps() override;
    void              DeleteExec(const char *name) override;
    void              Divide(Int_t nx=1, Int_t ny=1, Float_t xmargin=0.01, Float_t ymargin=0.01, Int_t color=0) override; // *MENU*
+   void              DivideRatios(Int_t nrows, Int_t ncolumns, const std::vector<double>& width_ratios={}, const std::vector<double>& height_ratios={}, const double canvas_top_margin=0., const double canvas_left_margin=0.);
```
silverweed: I think the default parameters should be passed in a struct so that one doesn't have to specify all of them if they only want to set `canvas_left_margin`.
Also the names should use the usual naming convention (`widthRatios`, `heightRatios`, ...).

graf2d/gpad/src/TPad.cxx
```diff
@@ -1345,6 +1345,109 @@ void TPad::Divide(Int_t nx, Int_t ny, Float_t xmargin, Float_t ymargin, Int_t co
...
+   double xr = 0.;
+   double yr = 0.;
+   double x  = 0.;
+   double y  = 1.;
+   double x1, y1, x2, y2;
+
+   // Check the validity of the margins
+   if (canvas_top_margin <0 || canvas_top_margin >1 ) {
+      Error("DivideRatios", "The canvas top margin must be >= 0 and <= 1");
+      return;
```
silverweed, line 1385: It would be better to either return `false` or throw an exception when an error occurs, otherwise there is no easy way to programmatically know whether the function failed or not.

graf2d/gpad/inc/TPad.h
couet: This way simply mimic what is done for the other Divide methods:
```  
 void              Divide(Int_t nx=1, Int_t ny=1, Float_t xmargin=0.01, Float_t ymargin=0.01, Int_t color=0) override; // *MENU*

```

graf2d/gpad/src/TPad.cxx
couet: Same comment as before, the other Divide method (Divide Square for instance) behave that way. I thought a consistent behaviour with what was done is the past is better than inventing a new way which can be applied only to this new method as we can not change the others for backward compatibility.

graf2d/gpad/src/TPad.cxx
```diff
@@ -1345,6 +1345,108 @@ void TPad::Divide(Int_t nx, Int_t ny, Float_t xmargin, Float_t ymargin, Int_t co
...
+      else     yr = 1./nyl;
+      for (int j=0; j<nxl; j++) {
+         if (wrs) xr = widthRatios[j];
+         else     xr = 1./nxl;
+         x1 = TMath::Max(0., x);
+         y1 = TMath::Max(0., y - yr);
+         x2 = TMath::Min(1., x + xr);
+         y2 = TMath::Min(1., y);
+         auto pad = new TPad(Form("%s_%d\n", GetName(), pn),
+                             Form("%s_%d\n", GetName(), pn),
```
linev: Use of multiple `Form()` invocations in single call is dangerous.
Please use `TString::Format()` instead.

graf2d/gpad/src/TPad.cxx
```diff
@@ -1435,8 +1435,8 @@ void TPad::DivideRatios(Int_t nx, Int_t ny,
          y1 = TMath::Max(0., y - yr);
          x2 = TMath::Min(1., x + xr);
          y2 = TMath::Min(1., y);
-         auto pad = new TPad(Form("%s_%d\n", GetName(), pn),
-                             Form("%s_%d\n", GetName(), pn),
+         auto pad = new TPad(TString::Format("%s_%d\n", GetName(), pn),
+                             TString::Format("%s_%d\n", GetName(), pn),
```
linev: @couet 
Only now I see - there is new line symbol "\n" in the name and title. Why?

couet: you are right ...

linev (approved):

PR #19010, opened 2025-06-11
[core] Disable R__SIZEDDELETE definition for macOS 26 beta

devajithvs:
MacOS 26.0 beta fails to build without this

This along with https://github.com/root-project/root/pull/18235 should fix MacOs 26.0 builds

This snippet fails to compile with the latest beta (unless we pass `-fsized-deallocation`):

```C++
#include <new>
int main() {
    void* ptr = nullptr;
    size_t size = 10;
    ::operator delete(ptr, size);
    return 0;
}
```

# This Pull request:

## Checklist:

- [x] tested changes locally

This PR fixes #

### DIFF
--- a/core/foundation/inc/ROOT/RConfig.hxx
+++ b/core/foundation/inc/ROOT/RConfig.hxx
@@ -341,11 +341,8 @@
 #   endif
 #endif
 
-#if defined(__GLIBCXX__) && !defined(__cpp_sized_deallocation)
-   // Sized global deallocation functions in libstc++ are only enabled if
-   // __cpp_sized_deallocation is defined, which Clang only does if explicitly
-   // passed -fsized-deallocation.
-#else
+#ifdef __cpp_sized_deallocation
+// Sized global deallocation functions are only enabled if __cpp_sized_deallocation is defined.
 #   define R__SIZEDDELETE
 #endif
 
--- a/roottest/root/meta/CMakeLists.txt
+++ b/roottest/root/meta/CMakeLists.txt
@@ -52,9 +52,16 @@ configure_file(typelist_win32.v6.cxx17.txt . COPYONLY)
 ROOTTEST_ADD_AUTOMACROS(DEPENDS ANSTmpltInt.C TmpltInt0.C TmpltInt1.C TmpltFloat.C
                                 TmpltNoSpec.C Event.cxx ${COMPILE_MACRO_TEST}
                                 EXCLUDE runMemberComments runautoload)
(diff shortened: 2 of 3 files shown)


core/foundation/inc/ROOT/RConfig.hxx
```diff
@@ -341,10 +341,11 @@
 #   endif
 #endif
 
-#if defined(__GLIBCXX__) && !defined(__cpp_sized_deallocation)
-   // Sized global deallocation functions in libstc++ are only enabled if
-   // __cpp_sized_deallocation is defined, which Clang only does if explicitly
-   // passed -fsized-deallocation.
+#if (defined(__GLIBCXX__) && !defined(__cpp_sized_deallocation)) || \
+   (defined(R__MACOSX) && __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 260000)
```
hahnjo: Is that needed because of an upstream change in libc++? Then we should probably check its version...

hahnjo (commented):
Can you try just
```c++
#ifdef __cpp_sized_deallocation
#   define R__SIZEDDELETE
#endif
```
As far as I understand https://github.com/llvm/llvm-project/commit/ef804d8f9b4ef4ff39e873d5910e94887519bdb6, libc++ will now also honor feature flag so my change from commit b5ad658677431041817b8a7ef10cf7056eb1f03f can be simplified even more.

devajithvs:
> Can you try just
> 
> ```c++
> #ifdef __cpp_sized_deallocation
> #   define R__SIZEDDELETE
> #endif
> ```
> 
> As far as I understand [llvm/llvm-project@ef804d8](https://github.com/llvm/llvm-project/commit/ef804d8f9b4ef4ff39e873d5910e94887519bdb6), libc++ will now also honor feature flag so my change from commit [b5ad658](https://github.com/root-project/root/commit/b5ad658677431041817b8a7ef10cf7056eb1f03f) can be simplified even more.

Thanks Jonas!

hahnjo:
Looks like we need a new reference file for macOS... Though it's curious that the functions are still universally there on Linux, I thought that `-fsized-deallocation` is not the default

hahnjo (approved):
LGTM. If needed, we can tweak choosing the reference file in the future, for example if some configuration on Linux doesn't enable `__cpp_sized_deallocation` either

hahnjo:
I'm seeing a local failure of `roottest-root-meta-runMemberComments` with Clang 18. I'm proposing to globally enable sized deallocation in https://github.com/root-project/root/pull/19053.

PR #19011, opened 2025-06-11
[io] Fix WalkTKeys mishandling strings longer than 127 characters

silverweed:
Using a signed `char` is incorrect, as strings in root files can be up to 255 characters long (as far as I understand).

## Checklist:

- [x] tested changes locally


io/io/src/TFile.cxx
```diff
@@ -1684,19 +1684,18 @@ std::optional<ROOT::Detail::TKeyMapNode> ROOT::Detail::TKeyMapIterable::TIterato
    }
 
    const auto readString = [&buffer, &header](bool skipCheck = false) {
-      char stringLen;
+      std::uint8_t stringLen;
       char str[256];
       if (!skipCheck && ((buffer - header) >= headerSize)) {
          stringLen = 0;
       } else {
          frombuf(buffer, &stringLen);
```
pcanal: Note that the 'authoritative' converse of this storage in actually (indirectly) `TKey::Streamer`.  The strings are stored using `TString::Streamer`  and thus by `TBufferFile::WriteTString` which indeed confirm that the size is stored as a `UChar_t` and in addition it actually supports/allows string longer than 255 characters.

silverweed: >  in addition it actually supports/allows string longer than 255 characters.

TIL. So the value 255 in a serialized string length means "the actual length is the next 4 bytes", as far as I understand?

pcanal: indeed.

silverweed:
@pcanal can this be merged?

PR #19012, opened 2025-06-11
[roottest] use cmake for `io/evolution` tests

linev:
There are several subdirs, which are handled fully independent now.

Several common scripts like `compile.C` or `run.C` moved to directories where they used - `numeric_conversion` and `stl_conversions`

Split several reference files - according to independent tests running

Remove no longer necessary `.rootrc`, `rootlogon.C`, `rootalias.C` files

Finally split large makefile test on 171 separate cmake tests


linev:
I see sporadic errors of `roottest-root-io-evolution-fixarr2` on Mac platforms.
Also on my Mac I see in 10% this test failing. 
Not sure if problem cased by changing from Makefile to cmake

pcanal:
> I see sporadic errors of roottest-root-io-evolution-fixarr2 on Mac platforms.

What is the error?

roottest/root/io/evolution/CMakeLists.txt
```diff
@@ -1,8 +1,205 @@
-#-------------------------------------------------------------------------------
-#
-# Placeholder file to translate the tests to the new CTest system. Meanwhile we
-# define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
-#
-#-------------------------------------------------------------------------------
-ROOTTEST_ADD_OLDTEST(LABELS longtest io)
```
pcanal, line 7: The test seems to have lost the io label ...

roottest/cmake/modules/RoottestMacros.cmake
```diff
@@ -683,7 +683,7 @@ macro(ROOTTEST_SETUP_MACROTEST)
                -e "gROOT->SetMacroPath(\"${CMAKE_CURRENT_SOURCE_DIR}\")"
                -e "gInterpreter->AddIncludePath(\"-I${CMAKE_CURRENT_BINARY_DIR}\")"
                -e "gSystem->AddIncludePath(\"-I${CMAKE_CURRENT_BINARY_DIR}\")"
-               ${RootExternalIncludes} ${RootExeOptions}
+               ${RootExternalIncludes} ${RootExeOptions} ${ARG_ROOTEXE_OPTS}
```
pcanal, line 686: The variable `RootExeOptions` and the argument `ROOTEXE_OPTS` now seems to have the exact same semantic.  Any reason to keep both?

linev:
>>    I see sporadic errors of roottest-root-io-evolution-fixarr2 on Mac platforms.

> What is the error?

See result of last CI on mac15:

https://github.com/root-project/root/actions/runs/15714091698/job/44279370611?pr=19012#step:6:8703

It produces stack trace when file is deleted.
In rare cases same failure appears on my Mac and once or twice on my Linux.

Old Makefile-based tests may miss such problem while only stdout was analyzed - which is ok at this moment

linev:
> The variable RootExeOptions and the argument ROOTEXE_OPTS now seems to have the exact same semantic. Any reason to keep both?

For me `ROOTEXE_OPTS` is much clear compared to `RootExeOptions`.
While `RootExeOptions` affects all tests in current directory - and subdirs.
I rather prefer replace use of `RootExeOptions` - by `ROOTEXE_OPTS`.
There are very few places in ROOT

roottest/root/io/evolution/CMakeLists.txt
linev: > The test seems to have lost the io label ...

I will add labels to all tests.

pcanal:
> For me ROOTEXE_OPTS is much clear compared to RootExeOptions.
> I rather prefer replace use of RootExeOptions - by ROOTEXE_OPTS.

I totally agree.  Can you make a separate PR with just that change?  Thanks.

roottest/root/io/evolution/CMakeLists.txt
pcanal: Thanks (albeit no all of the new test should be marked as long :) ).

pcanal:
> Old Makefile-based tests may miss such problem while only stdout was analyzed - which is ok at this moment

The old makefile would not (should not) missed any failure that include a stack trace or crash.

linev:
> The old makefile would not (should not) missed any failure that include a stack trace or crash.

I just checked roottest repository - Makefile not activate `fixarr` target.
So test was not running at all.

pcanal:
> I just checked roottest repository - Makefile not activate fixarr target.
> So test was not running at all.

Fair enough.  We should probably open an issue to remind us to look at it later.

linev:
> > For me ROOTEXE_OPTS is much clear compared to RootExeOptions.
> > I rather prefer replace use of RootExeOptions - by ROOTEXE_OPTS.

> I totally agree. Can you make a separate PR with just that change? Thanks.

@pcanal 
I prepared such branch: https://github.com/linev/root/tree/roottest_exe_opts

But first this PR must be merged.

roottest/cmake/modules/RoottestMacros.cmake
linev: This will be handled in follow-up PR

PR #19015, opened 2025-06-11
[Python] Check for identity and not equality where appropriate

guitargeek:
There are several unit tests that assert if different variables actually
point to the same histogram object, using the `==` operator.

Until now, this was correct by accident, because the default equality
check for TObject-derived classes in PyROOT is to call
`TObject::IsEqual()`, which does checks the pointer identity (if not
overridden is not implemented). See:
https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/src/TObjectPyz.cxx#L49

Since PR https://github.com/root-project/root/pull/17950 (introduction of UHI interface), an equality operator is
implemented for histograms on the Python side. Therefore, we really need
to correctly use Pythons `is` function to explicitly check for identity
(meaning both variables point to the same object).

### DIFF
--- a/bindings/pyroot/pythonizations/test/tdirectory_attrsyntax.py
+++ b/bindings/pyroot/pythonizations/test/tdirectory_attrsyntax.py
@@ -49,7 +49,7 @@ class TDirectoryReadWrite(unittest.TestCase):
         self.dir0["h"]
         # check that the cached value in is actually the object
         # inside the directory
-        self.assertEqual(self.dir0._cached_items["h"], self.dir0["h"])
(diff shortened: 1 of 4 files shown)


vepadulano (approved):
Thanks, but I would argue the commit description is slightly incorrect

> Therefore, the tests are not correct anymore.

From my understanding, the tests have never been correct. If the purpose of the tests was not to check the equality of the histograms (for whatever definition of equality), but they were meant to test that in fact the two objects were pointing to the same thing, then the tests should have always used the `is` builtin rather than equality.

PR #19019, opened 2025-06-12
[roottest] remove files from source dir after macro compilation [win]

linev:
On Windows `ROOTTEST_COMPILE_MACRO` creates number of files in source directory. 
Copy necessary to build directory and then remove all files from source dir

### DIFF
--- a/roottest/cmake/modules/RoottestMacros.cmake
+++ b/roottest/cmake/modules/RoottestMacros.cmake
@@ -286,7 +286,13 @@ macro(ROOTTEST_COMPILE_MACRO filename)
        COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/${dll_name}.dll
                                         ${CMAKE_CURRENT_BINARY_DIR}/
        COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/${dll_name}_ACLiC_dict_rdict.pcm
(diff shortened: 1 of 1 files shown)


bellenot (approved):

roottest/cmake/modules/RoottestMacros.cmake
```diff
@@ -286,7 +286,13 @@ macro(ROOTTEST_COMPILE_MACRO filename)
        COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/${dll_name}.dll
                                         ${CMAKE_CURRENT_BINARY_DIR}/
        COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/${dll_name}_ACLiC_dict_rdict.pcm
-                                        ${CMAKE_CURRENT_BINARY_DIR}/)
+                                        ${CMAKE_CURRENT_BINARY_DIR}/
+       COMMAND ${CMAKE_COMMAND} -E remove -f ${CMAKE_CURRENT_SOURCE_DIR}/${dll_name}.d
```
pcanal, line 290: For both Linux and Windows, we might be able to use `gSystem->SetBuildDir` to tell `ACLiC` where to put the build products ....

linev: Yes, seems to be it is possible and works on my windows

https://github.com/root-project/root/pull/19054

PR #19023, opened 2025-06-12
[tmva][sofie] fix compute strides for output tensor in topK

sanjibansg:
This PR fixes the computation of strides for the output tensor in the topK operator.


tmva/sofie/inc/TMVA/ROperator_Gemm.hxx
```diff
@@ -64,6 +64,7 @@ namespace SOFIE{
          fActivation = activation;
          fType = "float";
 
+         fInputTensorNames  = { fNA, fNB, fNC };
```
lmoneta: Is this extra lines needed? Was not there before?

sanjibansg: no, we missed this, because of this the input tensors here were not correctly flushed by the memory optimization mechanism

PR #19025, opened 2025-06-12
[reve] Send data to clients only to active connections

alja:
Here are two corrections needed for the case of event display instance periodically looping over hundred thousand events. E.g. live event display visualization.


alja:
@linev Please check

linev (approved):

PR #19028, opened 2025-06-13
[roottest] Disable tests that require NumPy if NumPy is not available

guitargeek:
Disable tests that require NumPy if NumPy is not available by factorizing the relevant tests out into a separate tests, where the NumPy dependency is correctly registered.

This helps when testing ROOT with newer Python versions where NumPy is not readily available yet.


guitargeek:
Thanks for the review! `ruff` has so much to complain about this file, and I think we should not change it too much at this point do it doesn't deviate as much from upstream cppyy

PR #19029, opened 2025-06-13
Upgrade to latest CppInterOp

aaronj0:
This upgrade brings in various developments that improve JitCall, add support for multiple interpreters and add new reflection interfaces for cppyy

The library was also reorganised, leading to the large diff


guitargeek (approved):
Thanks!

PR #19030, opened 2025-06-13
[CMake] Fix compile definitions being passed to rootcling as -D-Dxxx.

hageboeck:
When passing compile definitions to cling, ROOT unconditionally adds -D. Some projects such as XRootD already have -D in their definition, leading to an error such as:
While building module 'NetxNG':
<command line>:4:9: error: macro name must be an identifier \#define -D_FILE_OFFSET_BITS 64

The "canonical" way in CMake would be to not have explicit `-D`s in the COMPILE_DEFINITIONS property, so we can work around the issue https://github.com/xrootd/xrootd/issues/2543 (fixed in 5.8.4) by overwriting XRootD's COMPILE_DEFINITIONS property.

### DIFF
--- a/cmake/modules/SearchInstalledSoftware.cmake
+++ b/cmake/modules/SearchInstalledSoftware.cmake
@@ -968,6 +968,17 @@ if(xrootd AND NOT builtin_xrootd)
       endif()
     endif()
   endif()
+
+  if(XRootD_VERSION VERSION_LESS 5.8.4)
+    # Remove -D from XRootD's exported compile definitions. https://github.com/xrootd/xrootd/issues/2543
+    foreach(XRDTarget XRootD::XrdCl XRootD::XrdUtils)
+      if(TARGET ${XRDTarget})
+        get_target_property(PROP ${XRDTarget} INTERFACE_COMPILE_DEFINITIONS)
+        list(TRANSFORM PROP REPLACE "^-D" "")
+        set_property(TARGET ${XRDTarget} PROPERTY INTERFACE_COMPILE_DEFINITIONS ${PROP})
+      endif()
+    endforeach()
+  endif()
 endif()
 
 if(builtin_xrootd)


cmake/modules/RootMacros.cmake
```diff
@@ -661,7 +661,8 @@ function(ROOT_GENERATE_DICTIONARY dictionary)
       # The COMPILE_DEFINITIONS list might contain empty elements. These are
       # removed with the FILTER generator expression, excluding elements that
       # match the ^$ regexp (only matches empty strings).
-      set(module_defs "$<FILTER:$<TARGET_PROPERTY:${ARG_MODULE},COMPILE_DEFINITIONS>,EXCLUDE,^$>")
+      # Some projects (e.g. XRootD) export the defs including -D, so we have to remove them here
```
pcanal: Isn't it actually a bug in the `XRootD` projects?  E.g. https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_DEFINITIONS.html  seem to be explicit that the `-D` part is not expected ...

hageboeck: I wasn't sure, but I did already report it:
https://github.com/xrootd/xrootd/issues/2543

hageboeck: > https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_DEFINITIONS.html seem to be explicit that the -D part is not expected ...

I read this page slightly differently. CMake from v3.26 tries to be helpful in that it removes the `-D`, but it may mean that the original string still has the `-D`, don't you think?

I wrote a test that I posted to https://github.com/xrootd/xrootd/issues/2543 to confirm this, and that shows:
1. CMake stores no `-D` in the property, when the property is set implicitly via `target_compile_definitions`.
2. CMake doesn't do anything when the property is set directly, so `-D`s might slip in.
3. CMake generates compile commands correctly, so it somehow prevents `-D-D`.
4. *But* When you read the property via a generator expression, the `-D` from 2. are visible.

So it looks like it is "preferred" to not have `-D` in this property, and CMake is set up to deal with them if they are present, but our problem is that we were directly reading the property without any postprocessing.

So, to get out of this situation, I suggest that we filter leading `-D`. Either forever, or until the minimum supported XRootD version becomes 5.8.4. What do you think @pcanal ?

hageboeck: But the CI convinced me otherwise now. The genex-feature I used isn't available in our oldest supported CMake version.
Instead, I will apply a fix local to XRootD, overwriting the compile definitions property of XRootD until we only support >= 5.8.4

pcanal (approved):
Thanks.

PR #19032, opened 2025-06-13
[core] Fix TClassEdit not handling combination of <> and ()

### LINKED ISSUE #16119: Handling of std::conditional by Cling in ROOT 6.32.02.
### Check duplicate issues.

- [X] Checked for duplicates

### Description

When running with 6.32.02 the ALICE reconstruction we were plagued by errors like the following:

./itstpcMatch.log[16548:itstpc-track-writer]: Error in <TStreamerInfo::Build>: o2::dataformats::AbstractRef<25,5,2>, unknown type: typename std::conditional<(32> mRef

which were not there when using 6.30.0. We have since been able to workaround this by doing:

https://github.com/AliceO2Group/AliceO2/pull/13310/files

however I was wondering if this is not some more serious issue. In particular I am puzzled by the "(32" template argument, which seems to indicate some weird memory overwrite / macro expansion of NBIT.

### Reproducer

I do not have a simple reproducer.

### ROOT version

6.32.02

### Installation method

aliBuild

### Operating system

Linux

### Additional context

_No response_

silverweed:
This PR fixes #16119

### DIFF
--- a/core/foundation/src/TClassEdit.cxx
+++ b/core/foundation/src/TClassEdit.cxx
@@ -1204,10 +1204,10 @@ int TClassEdit::GetSplit(const char *type, vector<string>& output, int &nestedLo
       //we have 'something<'
       output.push_back(string(full,0,c - full.c_str()));
 
-      const char *cursor;
+      const char *cursor = c + 1;
       int level = 0;
       int parenthesis = 0;
-      for(cursor = c + 1; *cursor != '\0' && !(level==0 && *cursor == '>'); ++cursor) {
+      for ( ; *cursor != '\0' && !(level == 0 && parenthesis == 0 && *cursor == '>'); ++cursor) {
          if (*cursor == '(') {
             ++parenthesis;
             continue;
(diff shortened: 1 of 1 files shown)


core/foundation/test/file_16199.C
```diff
@@ -0,0 +1,31 @@
+#ifdef __ROOTCLING__
+#pragma link C++ class o2::dataformats::AbstractRef<25,5,2>+;
+#endif
```
pcanal: We usually put this after the declarations.

core/foundation/test/testClassEdit.cxx
```diff
@@ -336,3 +338,19 @@ TEST(TClassEdit, UnorderedMapNameNormalization)
...
+   TClassEdit::TSplitType t1("std::conditional<(1 < 32), int, float>");
+   EXPECT_STREQ("std::conditional", t1.fElements[0].c_str());
+   EXPECT_STREQ("(1<32)", t1.fElements[1].c_str());
+   EXPECT_STREQ("int", t1.fElements[2].c_str());
+   EXPECT_STREQ("float", t1.fElements[3].c_str());
+
+   gInterpreter->ProcessLine(".L file_16199.C+");
+   auto c = TClass::GetClass("o2::dataformats::AbstractRef<25,5,2>");
+   auto si = (TStreamerInfo*) c->GetStreamerInfo();
+   si->Build();
```
pcanal: That should be redundant/nop (i.e. already done in the `GetStreamerInfo`)

pcanal: Should we use `si->ls("noaddr");` instead?

Issue #14091, opened 2023-11-22
Bug in TAxis::FindBin (Double_t x) ?


cussol:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

Hello all,
It seems there is a bug in the TAxis::FindBin (Double_t x) method. When the joint script is executed, one can see that the bin number returned from ax->FindBin (ax->GetBinLowEdge (i)) is not always i as expected.

When looking at the code of TAxis::FindBin ([ROOT: hist/hist/src/TAxis.cxx Source File](https://root.cern/doc/v628/TAxis_8cxx_source.html#l00293) line 314), the formula used to compute the bin number is subject to rounding errors. It should be corrected as propose in the joint script.

Daniel CUSSOL

### Reproducer

File TAxisBug.C

```
#include "TAxis.h"
#include "Riostream.h"
#include "TROOT.h"

Int_t GetRightBin (TAxis *ax, Double_t x)
{
//
//  Get correct bin number from x value for TAxis ax
//   
Int_t retVal = 1+Int_t(ax->GetNbins()*(x-ax->GetXmin())/(ax->GetXmax()-ax->GetXmin()) + 0.5);
return retVal;
}

void TAxisBug (Int_t nBins = 30, Double_t xMin=3.0, Double_t xMax=6.0)
{
//
// Test TAxis bug 
// 
TAxis *ax=new TAxis (nBins, xMin, xMax);
cout << "Root version : " << gROOT->GetVersion() << endl; 
for (Int_t i=1;i<=ax->GetNbins();i++)
 {
 cout << i << " => " << ax->FindBin (ax->GetBinLowEdge (i)) 
       << " : " << 
       GetRightBin (ax, ax->GetBinLowEdge (i)); 
 if (i != ax->FindBin (ax->GetBinLowEdge (i)))
  {
  cout << " <= Bug!!! ";
  }
 cout << endl;
 }
cout << endl;
}

```

root TAxisBug.C+

### ROOT version

from ROOT v6.24.00

### Installation method

pre-built and binaries

### Operating system

Ubuntu 22.04.3 LTS (Jammy Jellyfish)

cussol:
As said in the forum, this is not really a bug but rather a general rounding issue. I close this bug report. Sorry for the disturbance !

chenhx6:
Please pay attention to this problem, as it may cause some pictures using special bins to become usable. Thanks, mala

hageboeck:
[...]

PR #19033, opened 2025-06-13
[hist] Improve precision of TAxis::FindFixBin / FindBin. 

hageboeck:
The bin computation in TAxis can suffer from floating-point uncertainties, returning bins such that this assertion breaks:
```c++
int bin = axis.FindBin(x);
assert(axis.GetBinLowEdge(bin) <= x && x < axis.GetBinUpEdge(bin));
```
This is of course surprising for users, and sometimes brings data into the overflow, although they fall into the valid range of the axis. Here, two terms are added, which fix these floating-point errors.

On a recent AMD CPU, there is no runtime difference, since the latency of the virtual call overshadows the additional computations for the correction. For 10M calls, running 100 times with perf, we arrive at:
| | Uncorrected | Corrected |
|---|---|---|
| Instructions | 250M | 360M |
| cycles | 216M | 218M |
| Run time | 0.0495 +- 0.0003 s | 0.0493 +- 0.0002 s |

Note:
Several attempts at reordering the equations for better precision failed. There were always a few test cases in the included tests that didn't pass, so applying the correction looks to be the best solution.

This PR fixes the issues brought up here:
- https://root-forum.cern.ch/t/floating-point-rounding-error-when-filling-the-histogram/35189
- https://root-forum.cern.ch/t/bug-or-feature-in-ttree-draw/62862/16
- 1703c54
- #14091

### DIFF
--- a/hist/hist/test/test_TH1.cxx
+++ b/hist/hist/test/test_TH1.cxx
@@ -149,3 +149,133 @@ TEST(TH1, Normalize)
    EXPECT_FLOAT_EQ(v2.Integral("width"), 1.);
    EXPECT_FLOAT_EQ(v2.GetMaximum(), 7.9999990);
 }
+
+TEST(TAxis, BinComputation_FPAccuracy)
+{
+   // Example from 1703c54
+   EXPECT_EQ(TAxis(1, -1., 0.).FindBin(-1e-17), 1);
+
+   {
+      // https://root-forum.cern.ch/t/floating-point-rounding-error-when-filling-the-histogram/35189
+      TAxis axis(128, -0.352, 0.352);
+      constexpr double x = -0.0220;
+      EXPECT_EQ(axis.FindBin(x), 61);
+      EXPECT_EQ(axis.FindFixBin(x), 61);
+      EXPECT_LE(axis.GetBinLowEdge(61), x);
+      EXPECT_GT(axis.GetBinUpEdge(61), x);
+   }
+
+   {
+      // https://github.com/root-project/root/issues/14091
+      constexpr int nBins = 30;
+      constexpr double xMin = 3.0, xMax = 6.0;
+      TAxis ax(nBins, xMin, xMax);
+
+      for (Int_t i = 1; i <= ax.GetNbins(); i++) {
+         EXPECT_EQ(i, ax.FindBin(ax.GetBinLowEdge(i)));
+      }
+   }
+
+   {
+      TAxis axis(2000, -1000., 1000.);
+      EXPECT_EQ(axis.FindFixBin(std::nextafter(-1000., -2000.)), 0);
+      EXPECT_EQ(axis.FindFixBin(-1000.), 1);
+      EXPECT_EQ(axis.FindFixBin(std::nextafter(-1000., 0)), 1);
+
+      EXPECT_EQ(axis.FindFixBin(-500.00000000001), 500);
+      EXPECT_EQ(axis.FindFixBin(-500.), 501);
+      EXPECT_EQ(axis.FindFixBin(-499.9), 501);
+
+      EXPECT_EQ(axis.FindFixBin(-1.E-20), 1000);
+      EXPECT_EQ(axis.FindFixBin(std::nextafter(-0., -1)), 1000);
+      EXPECT_EQ(axis.FindFixBin(-0.), 1001);
+      EXPECT_EQ(axis.FindFixBin(0.), 1001);
+
+      EXPECT_EQ(axis.FindFixBin(499.9), 1500);
+      EXPECT_EQ(axis.FindFixBin(500.), 1501);
+
+      EXPECT_EQ(axis.FindFixBin(1000. - 1.E-13), 2000);
+      EXPECT_EQ(axis.FindFixBin(std::nextafter(1000., 0.)), 2000);
+      EXPECT_EQ(axis.FindFixBin(1000.), 2001);
+
+      EXPECT_EQ(axis.FindBin(std::nextafter(-1000., -2000.)), 0);
+      EXPECT_EQ(axis.FindBin(-1000.), 1);
+      EXPECT_EQ(axis.FindBin(std::nextafter(-1000., 0)), 1);
+
+      EXPECT_EQ(axis.FindBin(-500.00000000001), 500);
+      EXPECT_EQ(axis.FindBin(-500.), 501);
+      EXPECT_EQ(axis.FindBin(-499.9), 501);
+
+      EXPECT_EQ(axis.FindBin(-1.E-20), 1000);
+      EXPECT_EQ(axis.FindBin(std::nextafter(-0., -1)), 1000);
+      EXPECT_EQ(axis.FindBin(-0.), 1001);
+      EXPECT_EQ(axis.FindBin(0.), 1001);
+
+      EXPECT_EQ(axis.FindBin(499.9), 1500);
+      EXPECT_EQ(axis.FindBin(500.), 1501);
+
+      EXPECT_EQ(axis.FindBin(1000. - 1.E-13), 2000);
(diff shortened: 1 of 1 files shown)


ferdymercury:
Thanks a lot Stephan!
Some questions from my side:
- Does this supersede https://github.com/root-project/root/pull/14105 ? Maybe it can be closed.
- Does this fix https://github.com/root-project/root/issues/14091 ? Maybe it can be linked.
- Does this new implementation affect runtime performance when filling a histogram? If yes, maybe it's better to have a separate function FindBinHighRes or so. One could even think of using long-double.
- Is it sure that the resulting bin is only off by one? I can imagine of (probably very weird cases) where the floating point limited precision might be two?

hageboeck:
> Thanks a lot Stephan! Some questions from my side:
> 
> * Does this supersede [[hist] Improve calculation of `TAxis::FindBin(x)` when x is at the bin edges #14105](https://github.com/root-project/root/pull/14105) ? Maybe it can be closed.
> * Does this fix [Bug in TAxis::FindBin (Double_t x) ? #14091](https://github.com/root-project/root/issues/14091) ? Maybe it can be linked.
> * Does this new implementation affect runtime performance when filling a histogram? If yes, maybe it's better to have a separate function FindBinHighRes or so. One could even think of using long-double.
> * Is it sure that the resulting bin is only off by one? I can imagine of (probably very weird cases) where the floating point limited precision might be two?

Good questions, I'm checking! 🙂

hageboeck:
> * Does this supersede [[hist] Improve calculation of `TAxis::FindBin(x)` when x is at the bin edges #14105](https://github.com/root-project/root/pull/14105) ? Maybe it can be closed.

It can be closed, indeed.

> * Does this fix [Bug in TAxis::FindBin (Double_t x) ? #14091](https://github.com/root-project/root/issues/14091) ? Maybe it can be linked.

It fixes the problem described there. I linked the issue.

> * Does this new implementation affect runtime performance when filling a histogram? If yes, maybe it's better to have a separate function FindBinHighRes or so. One could even think of using long-double.

It does not affect run time on a modern CPU, even though more instructions are issued. When you run the bare code to find the bin (not as part of a virtual function), you have the following behaviour:

| | Uncorrected | Corrected |
|---|---|---|
| Instructions | 8 | 19 |
| Cycles | 33 | 31 |

It doesn't matter how many instructions are issued. What's interesting is how many cycles they need to complete. Although that depends on the CPU in use, you see that the two versions are very close.

The killer argument, however, is that all of this happens inside a virtual function, and that of course has its own latencies. When plugging the codes into a virtual function, and benchmarking 10M calls, we arrive at:
| | Uncorrected | Corrected |
|---|---|---|
| Instructions | 250M | 360M |
| cycles | 216M | 218M |
| Run time | 0.0495 +- 0.0003 s | 0.0493 +- 0.0002 s |

So you see that more instructions are issued, but due to the CPU being a superscalar CPU, these additional instructions have no noteworthy effect on the run time, because the correction terms can be evaluated *in parallel* to the original computation.

**This means, that in TAxis, where it is part of a virtual function, we should not see a run time difference.**

> * Is it sure that the resulting bin is only off by one? I can imagine of (probably very weird cases) where the floating point limited precision might be two?

I did not manage to provoke a more-than-one difference. I didn't set out to prove it mathematically, but I had lots of difficulties crafting a case that made an "off by -1" error.

ferdymercury:
Awesome analysis, thanks a lot!!
Does this mean that I can close https://github.com/root-project/root/pull/17689 if you implement the same strategy in that part of the code ?
And not sure if you want to review this alleviation strategy: https://github.com/root-project/root/pull/17896

ferdymercury:
Btw, this PR reminds me of this issue: https://its.cern.ch/jira/browse/ROOT-10179

hageboeck:
> Awesome analysis, thanks a lot!! Does this mean that I can close #17689 if you implement the same strategy in that part of the code ?

@ferdymercury I think for this PR, the `x == max` --> overflow problem remains. So you could probably just use:
```c++
- max = currentMax;
+ max = std::nextafter(currentMax, inf);
```
but that's to be tested.

> Btw, this PR reminds me of this issue: https://its.cern.ch/jira/browse/ROOT-10179

Was this the correct link?

ferdymercury:
> Was this the correct link?

It was about the question/fact of whether performance was impacted. Maybe more the subissue https://its.cern.ch/jira/browse/ROOT-10185 which would have been easier for you if rootbench could be triggered from a PR.


> ```c++
>  max = std::nextafter(currentMax, inf);
>     
> ```

I had tried that and it didn't work. I really had to add 
`xmax = std::max(xmax + 1e-15*(xmax - xmin), std::nextafter(xmax,INFINITY));`

For most cases, the 1e-15*max_width is what really counts.
Or do you mean that that should be fine with just nextafter, 'after' your PR gets merged?

Related: https://github.com/root-project/root/pull/17896

dpiparo:
> Btw, this PR reminds me of this issue: https://its.cern.ch/jira/browse/ROOT-10179

Yes - well remembered @ferdymercury . `RootBench` should first of all be revived and regularly built and run as part of the battery of actions, as a start, nightly. After that is achieved, we can decide whether the timing resolution we have on the current machines used for the CI is enough.

dpiparo (approved):
Thanks for these changes and the spectacular performance analysis. As far as I am concerned, this could be merged now.

hageboeck:
> Awesome analysis, thanks a lot!! Does this mean that I can close #17689 if you implement the same strategy in that part of the code ?

@ferdymercury it looks like your change is still required. See this test (without your PR):
```c++
TH1F histo("histo", "Histo", 100, 0, 0);
histo.Fill(-999); histo.Fill(0);
histo.GetXaxis()->GetBinLowEdge(101) == 0.; // Overflow bin starts at 0.!
```

guitargeek:
Right, that's because as I said in [Lorenzos PR](https://github.com/root-project/root/pull/14105#pullrequestreview-2568781881), the functions like `GetBinLowEdge()` have separate implementations:

https://github.com/root-project/root/blob/master/hist/hist/src/TAxis.cxx#L522

Now we are in the situation that some functions (`TAxis::FindBin`) have this precision optimization with this PR, and others don't. So ROOT is a bit self consistent now :slightly_smiling_face: That's the reason why I didn't merge Lorenzos PR. I thought letting the precision of of `TAxis::FindBin()` get out of sync with other algorithms in ROOT was too risky.

But it's some time I looked at this and I don't remember the details now :slightly_smiling_face: I just wanted to point out the possible problems as good as I could with the memory I had.

hageboeck:
> Right, that's because as I said in [Lorenzos PR](https://github.com/root-project/root/pull/14105#pullrequestreview-2568781881), the functions like `GetBinLowEdge()` have separate implementations:
> 
> https://github.com/root-project/root/blob/master/hist/hist/src/TAxis.cxx#L522
> 
> Now we are in the situation that some functions (`TAxis::FindBin`) have this precision optimization with this PR, and others don't. So ROOT is a bit self consistent now 🙂 That's the reason why I didn't merge Lorenzos PR. I thought letting the precision of of `TAxis::FindBin()` get out of sync with other algorithms in ROOT was too risky.

No, actually I don't think that's what happened in this PR. **Only now**, the `GetBinLowEdge` and `FindBin` are in sync! 😄
The trick about this change is that we compute the bin in the same old way (with fp uncertainties), and the new terms *ensure consistency with what GetBinLowEdge returns*. So I would argue that now, for the first time in 25 years, TAxis is consistent.
The test that I included is supposed to reflect this. The hand-written parts are engineered using `GetBinLowEdge`, and the externally provided examples frequently compare to `GetBinLowEdge`.

What the initial comment referred to is a deficiency of the auto-binning feature when you don't give any axis limits. This needs a bit more work, and ferdy was on to it.

PR #19036, opened 2025-06-13
[tree] properly copy entry list, also queried indices

### LINKED ISSUE #13338: (Major?) bug : TEntryList and TTree inconsitency of entry number returned by tree if entry list has been used previously
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Hello, 

I am creating this issue because there is a bug when setting an entry list that has already been used to a `TTree`.
The entry number returned by `tree->GetEntryNumber(index)` are messed up until that function is called 2 times in a row for the same index.... 

You can find a standalone simple reproducer of the bug below. 

In that code  by setting `numTest` to `0`, `1` or `2` you will obtain the different print below. 
With that example by calling `tree->GetEntryNumber(index)` it should return the value `index` 
e.g. `tree->GetEntryNumber(2)` should return `2`. However until the `GetEntryNumber()` function has not been called 2 times for the same index then the returned entry number is NOT  the expected one. 

Here it is simple to solve the issue i.e. not call  `entryList->GetEntry(15);` before setting the entry list to the tree 
but I find this behaviour really error prone as one would need to be sure the entry list has never been used basically.    

Tagging @pcanal and @Axel-Naumann 

For `numTest=0`
```
tree->GetEntryNumber(0)=16
tree->GetEntryNumber(0)=0
tree->GetEntryNumber(0)=0
tree->GetEntryNumber(0)=0
```
For  `numTest=1`
```
tree->GetEntryNumber(0)=16
tree->GetEntryNumber(1)=17
tree->GetEntryNumber(1)=1
tree->GetEntryNumber(1)=1
```
For `numTresr=2`
```
tree->GetEntryNumber(0)=16
tree->GetEntryNumber(1)=17
tree->GetEntryNumber(2)=18
tree->GetEntryNumber(2)=2
tree->GetEntryNumber(2)=2
```

### Reproducer

```
#include "TFile.h"
#include "TTree.h"
#include "TError.h"
#include "TEntryList.h"

#include<string>
#include<vector>
#include<iostream>


// function defined after main 
void createTree(const std::string &outputFile);

int main(){
  
  // number of the test 
  // set to 0, 1 or 2
  int numTest = 2; 

  // Define an entry list 
  TEntryList *entryList = new TEntryList(); 
  std::vector<long long int> entries={
    0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19
  };

  // fill the entry list 
  for (long long int entry : entries){
    entryList->Enter(entry);
  }

  // Create tree 
  std::string inFileName = "myFile.root"; 
  createTree(inFileName);

  // Read tree 
  TFile *file = nullptr; 
  TTree *tree = nullptr;

  file = TFile::Open(inFileName.c_str());
  tree = (TTree*) file->Get("myTree");

  // On purpose get entry 15 
  // before setting entry list for tree 
  entryList->GetEntry(15);
  tree->SetEntryList(entryList);
  
  if (numTest == 0){
    std::cout << "tree->GetEntryNumber(0)=" << tree->GetEntryNumber(0) << std::endl; 
    std::cout << "tree->GetEntryNumber(0)=" << tree->GetEntryNumber(0) << std::endl;
    std::cout << "tree->GetEntryNumber(0)=" << tree->GetEntryNumber(0) << std::endl;
    std::cout << "tree->GetEntryNumber(0)=" << tree->GetEntryNumber(0) << std::endl;
  }
  
  if (numTest == 1){
    std::cout << "tree->GetEntryNumber(0)=" << tree->GetEntryNumber(0) << std::endl; 
    std::cout << "tree->GetEntryNumber(1)=" << tree->GetEntryNumber(1) << std::endl;
    std::cout << "tree->GetEntryNumber(1)=" << tree->GetEntryNumber(1) << std::endl;
    std::cout << "tree->GetEntryNumber(1)=" << tree->GetEntryNumber(1) << std::endl;
  }

  if (numTest == 2){
    std::cout << "tree->GetEntryNumber(0)=" << tree->GetEntryNumber(0) << std::endl; 
    std::cout << "tree->GetEntryNumber(1)=" << tree->GetEntryNumber(1) << std::endl;
    std::cout << "tree->GetEntryNumber(2)=" << tree->GetEntryNumber(2) << std::endl;
    std::cout << "tree->GetEntryNumber(2)=" << tree->GetEntryNumber(2) << std::endl;
    std::cout << "tree->GetEntryNumber(2)=" << tree->GetEntryNumber(2) << std::endl;
  }

  return 0; 
}

void createTree(const std::string &outputFile){

  unsigned int NEntries = 100;
  TTree *tree = nullptr;
  double x1;

  // create file 
  TFile *file = TFile::Open(outputFile.c_str(),"recreate"); 
  // init tree 
  tree = new TTree("myTree","myTree"); 
  tree->Branch("x1", &x1);

  for (unsigned int iEntry = 0 ; iEntry < NEntries; iEntry++){
    x1 = iEntry;
    tree->Fill();
  }

  file->Write(); 
  delete file;
}
```

### ROOT version

ROOT Version: 6.26/08 
I believe all versions are affected by that bug even the most recent ones 

### Installation method

Built from source

### Operating system

Linux

### Additional context

_No response_

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/13338

fyi @romainbqt

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/tree/src/TEntryList.cxx
+++ b/tree/tree/src/TEntryList.cxx
@@ -1264,6 +1264,8 @@ void TEntryList::SetTree(const char *treename, const char *filename)
             elist->fBlocks = fBlocks;
             fBlocks = nullptr;
             elist->fNBlocks = fNBlocks;
+            elist->fLastIndexQueried = fLastIndexQueried;
+            elist->fLastIndexReturned = fLastIndexReturned;
             fLists->Add(elist);
             elist = new TEntryList("", "", treename, fn.Data());
             if (elist->GetDirectory()) {
--- a/tree/tree/test/chain_setentrylist.cxx
+++ b/tree/tree/test/chain_setentrylist.cxx
@@ -162,9 +162,9 @@ TEST(TChain, SetEntryListSyncWrongNumberOfSubLists) {
 TEST(TTree, SetEntryListPreUsed)
 {
    // https://github.com/root-project/root/issues/13338
-   // Define and fill an entry list 
+   // Define and fill an entry list
    TEntryList entryList;
-   const std::vector<long long int> entries {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19};
+   const std::vector<long long int> entries{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19};
    for (long long int entry : entries)
       entryList.Enter(entry);
    // Read tree
@@ -176,7 +176,7 @@ TEST(TTree, SetEntryListPreUsed)
       x1 = iEntry;
       tree.Fill();
    }
-   // On purpose get entry 15 before setting entry list for tree 
+   // On purpose get entry 15 before setting entry list for tree
    entryList.GetEntry(15);
    tree.SetEntryList(&entryList);
    // Check results
--- a/tree/tree/test/chain_setentrylist.cxx
+++ b/tree/tree/test/chain_setentrylist.cxx
@@ -158,3 +158,30 @@ TEST(TChain, SetEntryListSyncWrongNumberOfSubLists) {
    gSystem->Unlink(filename1);
    gSystem->Unlink(filename2);
 }
+
+TEST(TTree, SetEntryListPreUsed)
+{
+   // https://github.com/root-project/root/issues/13338
+   // Define and fill an entry list 
+   TEntryList entryList;
+   const std::vector<long long int> entries {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19};
+   for (long long int entry : entries)
+      entryList.Enter(entry);
+   // Read tree
+   TTree tree("t", "t");
+   unsigned int x1;
+   tree.Branch("x1", &x1);
+   unsigned int NEntries = 100;
+   for (unsigned int iEntry = 0; iEntry < NEntries; iEntry++) {
+      x1 = iEntry;
+      tree.Fill();
+   }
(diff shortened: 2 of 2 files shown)


tree/tree/src/TEntryList.cxx
```diff
@@ -1264,6 +1264,8 @@ void TEntryList::SetTree(const char *treename, const char *filename)
             elist->fBlocks = fBlocks;
             fBlocks = nullptr;
             elist->fNBlocks = fNBlocks;
+            elist->fLastIndexQueried = fLastIndexQueried;
+            elist->fLastIndexReturned = fLastIndexReturned;
```
pcanal, line 1268: Most other places, we seem to be using:
```suggestion
            elist->fLastIndexQueried = -1;
            elist->fLastIndexReturned = 0;
```
why is it different here?  (especially since I am not sure whether LastIndex returns is likely to need to be expressed in local entry number)

ferdymercury: I am not sure. It just fixed the issue (empirically)

pcanal: I think we do need to understand in order to be able to add a comment with the explanation (so that future maintainer do not spend time 'pondering' the differences).  In particular in this case, I don't know whether it is assigning (as opposed to not assignment) or assigning to this particular values that fixes the problem (In particular since the later may or may not work for case of larger chains).

pcanal: Did we resolve this question?

ferdymercury: Nope

dpiparo:
moving to 6.42 Milestone since we still need to converge.

PR #19039, opened 2025-06-13
[tree] avoid resetting index in entry list copy constructor

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-10807

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/tree/src/TEntryList.cxx
+++ b/tree/tree/src/TEntryList.cxx
@@ -287,8 +287,8 @@ TEntryList::TEntryList(const TEntryList &elist) : TNamed(elist)
    fFileName = elist.fFileName;
    fStringHash = elist.fStringHash;
    fTreeNumber = elist.fTreeNumber;
-   fLastIndexQueried = -1;
-   fLastIndexReturned = 0;
+   fLastIndexQueried = elist.fLastIndexQueried;
+   fLastIndexReturned = elist.fLastIndexReturned;
    fN = elist.fN;
    fShift = elist.fShift;
    fLists = nullptr;


pcanal:
Could you add a test?

ferdymercury:
Test added. I am not sure if the failures are race condition or if it's breaking sth.

tree/tree/test/entrylist_addsublist.cxx
```diff
@@ -66,3 +67,61 @@ TEST(TEntryList, addSubList) {
...
+      value = lcopy.GetEntryAndTree(1, treenum);
+      EXPECT_EQ(value, 1);
+      EXPECT_EQ(treenum, 0);
+      value = lcopy.GetEntryAndTree(2, treenum);
+      EXPECT_EQ(value, 1);
+      EXPECT_EQ(treenum, 1);
+   }
+
+   for (auto filename : {filename1, filename2})
+      gSystem->Unlink(filename1);
```
silverweed:
```suggestion
      gSystem->Unlink(filename);
```

silverweed (changes requested):

tree/tree/src/TEntryList.cxx
```diff
@@ -287,8 +287,8 @@ TEntryList::TEntryList(const TEntryList &elist) : TNamed(elist)
    fFileName = elist.fFileName;
    fStringHash = elist.fStringHash;
    fTreeNumber = elist.fTreeNumber;
-   fLastIndexQueried = -1;
-   fLastIndexReturned = 0;
+   fLastIndexQueried = elist.fLastIndexQueried;
+   fLastIndexReturned = elist.fLastIndexReturned;
```
pcanal, line 290: It is a bit surprising changing for the usual starting point to the previously visited entry fixes the problem.  Those 2 data members are meant to be accelerator and so the default value should have been fine. ...  What is the mechanism leading to the failure?

ferdymercury: I guess it was an empirical fix. Without this, the returned value is PreviousVisitedEntry + 1, so 16 even if Index is -1
It seems as if the variable index is -1 but the entry pointer is still at 15.

PR #12484, opened 2023-03-14
JsMVA: use latest JSROOT, use `/static` path to load code

linev:
Fixes https://github.com/root-project/root/issues/12441

1. Move all JS/CSS files to `js/mva/` folder which can be accessed via `/static/mva/` URL from Jupyter Notebook. This allows to use JsMVA code directly from ROOT without need to rely on `https://root.cern/js/jsmva/latest` 
2. Remove minified scripts - no real gain in performance, only extra step in deployment
3. Adjust `JsMVA.js` code to JSROOT v7. Use bundle provided in `build/jsroot.js`. 
4. Adjust `OutputTransformer.py` to work with python3, probably other scripts should be improved as well
5. Adjust `NetworkDesigner.js` to use d3 v6, load d3 from external source

Still to do: 
1. Fix more problems with `python3` - see attached image from my notebook
python3_error
2. Fix `NeuralNetwork.js` to be usable with latest d3.js. I need working example for it.

PR #19040, opened 2025-06-13
[Python] Don't include JsMVA in builds

### LINKED ISSUE #12441: JsMVA should be updated to use the same jsroot version as the rest of ROOT
### Explain what you would like to see improved

The JsMVA code `bindings/jsmva/` and `etc/notebook/JsMVA/` uses a different and older version of jsroot than the rest of ROOT.
 
### Optional: share how it could be improved

It should either be updated to use the same version or dropped.

guitargeek:
The JsMVA package is undocumented, [untested](https://github.com/root-project/root/blob/master/roottest/python/JsMVA/CMakeLists.txt), and lagging behind in terms
of the [JsROOT version that it uses](https://github.com/root-project/root/issues/12441).

At this point, there is no clear reason for including it in the ROOT
builds, so this commit suggests to remove it from the build and install
tree temporarily, until we are sure that it works (also with the newest
JsROOT version).

Alternative to #12484.

Closes #12441.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -345,6 +345,9 @@ file(GLOB_RECURSE artifact_files RELATIVE ${CMAKE_SOURCE_DIR} etc/*)
 # Exclude the *.in files, just like in the install tree
 list(FILTER artifact_files EXCLUDE REGEX "\.(in)$")
 
+# Exclude etc/notebook/JsMVA
+list(FILTER artifact_files EXCLUDE REGEX "^etc/notebook/JsMVA/")
+
 foreach(artifact_file ${artifact_files})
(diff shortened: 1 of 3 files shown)


dpiparo (approved):
Thanks. I agree with this course of actions. 
Should this become an app external to ROOT too?

guitargeek:
That's hard to say. The only information I could find about this package are these GSoC presentations (it was a [GSoC project](https://github.com/qati/GSOC16) mentored by @etejedor):
  * https://indico.cern.ch/event/565201/#3-integrating-machine-learning
  * https://indico.cern.ch/event/569308/#7-jupyter-and-tmva

The maintenance inside ROOT is challenging because it's the only JavaScript code in ROOT that is not maintained by @linev. Indeed, it might be easier to ship it as a small Python package on PyPI.

Issue #11431, opened 2022-09-25
Installing with CMAKE_INSTALL_PYTHONDIR different from CMAKE_INSTALL_LIBDIR leads to incorrect runpath configuration


guicale:
- [x] Checked for duplicates

### Describe the bug
Installing ROOT with a path for `CMAKE_INSTALL_PYTHONDIR` different than the one for `CMAKE_INSTALL_LIBDIR` leads to problems trying to use the Python modules provided by ROOT, such as the example that follows.

```
$ rootbrowse
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/cppyy/__init__.py", line 60, in <module>
    importlib.import_module(libcppyy_mod_name)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 674, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 571, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 1176, in create_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
ImportError: libcppyy_backend3_10.so.6.26: cannot open shared object file: No such file or directory

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/rootbrowse", line 10, in <module>
    import cmdLineUtils
  File "/usr/local/lib/python3.10/dist-packages/cmdLineUtils.py", line 82, in <module>
    import ROOT
  File "/usr/local/lib/python3.10/dist-packages/ROOT/__init__.py", line 22, in <module>
    import cppyy
  File "/usr/local/lib/python3.10/dist-packages/cppyy/__init__.py", line 62, in <module>
    raise ImportError(
ImportError: Failed to import libcppyy3_10. Please check that ROOT has been built for Python 3.10
```

A check with `readelf` reveals the runpath of the libraries installed to `CMAKE_INSTALL_PYTHONDIR` is the reason for this problem.

```
$ readelf -a /usr/local/lib/root/libcppyy3_10.so | grep -i runpath
[...]

PR #19044, opened 2025-06-14
[CMake] Avoid unneeded artifacts in the library installation directory

guitargeek:
Avoid unneeded artifacts in the library installation directory, in particular empty directories if the Python interfaces are installed in a directory other than the library installation directory.

Motivated by:
  * https://github.com/root-project/root/issues/11431

More detail in the commit description.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -121,10 +121,17 @@ target_compile_definitions(Core PRIVATE
 if (runtime_cxxmodules)
   list(APPEND core_implicit_modules "-mSystemByproducts")
   # Force generation of _Builtin_intrinsics from Core.
-  list(APPEND core_implicit_modules "-m" "_Builtin_intrinsics" "-mByproduct" "_Builtin_intrinsics")
-  list(APPEND core_implicit_modules "-mByproduct" "ROOT_Foundation_Stage1_NoRTTI")
-  list(APPEND core_implicit_modules "-mByproduct" "ROOT_Foundation_C")
-  list(APPEND core_implicit_modules "-mByproduct" "ROOT_Rtypes")
+  list(APPEND core_implicit_modules "-m" "_Builtin_intrinsics")
+
+  set(byproducts _Builtin_intrinsics ROOT_Foundation_Stage1_NoRTTI ROOT_Foundation_C ROOT_Rtypes)
+  foreach(byproduct ${byproducts})
+    list(APPEND core_implicit_modules "-mByproduct" "${byproduct}")
+    install(FILES ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/${byproduct}.pcm DESTINATION ${CMAKE_INSTALL_LIBDIR})
+  endforeach()
+
+  list(APPEND core_implicit_modules "-mByproduct" "")
+  list(APPEND core_implicit_modules "-mByproduct" "")
+  list(APPEND core_implicit_modules "-mByproduct" "")
 endif(runtime_cxxmodules)
 
 get_target_property(CORE_DICT_HEADERS Core DICT_HEADERS)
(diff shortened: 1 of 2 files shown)


core/CMakeLists.txt
```diff
@@ -117,10 +117,17 @@ target_compile_definitions(Core PRIVATE
...
-  list(APPEND core_implicit_modules "-mByproduct" "ROOT_Rtypes")
+  list(APPEND core_implicit_modules "-m" "_Builtin_intrinsics")
+
+  set(byproducts _Builtin_intrinsics ROOT_Foundation_Stage1_NoRTTI ROOT_Foundation_C ROOT_Rtypes)
+  foreach(byproduct ${byproducts})
+    list(APPEND core_implicit_modules "-mByproduct" "${byproduct}")
+    install(FILES ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/${byproduct}.pcm DESTINATION ${CMAKE_INSTALL_LIBDIR})
+  endforeach()
+
+  list(APPEND core_implicit_modules "-mByproduct" "")
```
pcanal: What is the semantic of those additions?

core/CMakeLists.txt
```diff
@@ -117,10 +117,17 @@ target_compile_definitions(Core PRIVATE
 if (runtime_cxxmodules)
   list(APPEND core_implicit_modules "-mSystemByproducts")
   # Force generation of _Builtin_intrinsics from Core.
-  list(APPEND core_implicit_modules "-m" "_Builtin_intrinsics" "-mByproduct" "_Builtin_intrinsics")
-  list(APPEND core_implicit_modules "-mByproduct" "ROOT_Foundation_Stage1_NoRTTI")
-  list(APPEND core_implicit_modules "-mByproduct" "ROOT_Foundation_C")
-  list(APPEND core_implicit_modules "-mByproduct" "ROOT_Rtypes")
+  list(APPEND core_implicit_modules "-m" "_Builtin_intrinsics")
```
pcanal, line 124: This looks like it would now be listed twice in `core_implicit_modules`, i.e. once more inside the new foreach loop.

bindings/pyroot/cppyy/CPyCppyy/CMakeLists.txt
```diff
@@ -106,15 +104,9 @@ set_property(GLOBAL APPEND PROPERTY ROOT_EXPORTED_TARGETS ${libname})
...
-                            RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT libraries
-                            LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT libraries
-                            ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT libraries)
-if (NOT MSVC AND NOT CMAKE_INSTALL_LIBDIR STREQUAL CMAKE_INSTALL_PYTHONDIR)
-  # add a symlink to ${libname} in CMAKE_INSTALL_PYTHONDIR
-  set(LIB_FILE_NAME ${CMAKE_SHARED_LIBRARY_PREFIX}${libname}.so)
-  install(CODE "file(CREATE_LINK ${PYTHONDIR_TO_LIBDIR}/${LIB_FILE_NAME}
-    \$ENV{DESTDIR}${CMAKE_INSTALL_FULL_PYTHONDIR}/${LIB_FILE_NAME} SYMBOLIC)")
-endif()
+                            RUNTIME DESTINATION ${CMAKE_INSTALL_PYTHONDIR} COMPONENT libraries
```
pcanal: The original reason for the symlinks was:
```
    libcppyy.so and libcppyy_backend.so are special, as they are needed
    by both python modules (namely, by cppyy) and C++ libraries (e.g. by
    TPython), so if CMAKE_INSTALL_PYTHONDIR is different from
    CMAKE_INSTALL_LIBDIR we install these files in CMAKE_INSTALL_LIBDIR
    but also add symlinks to them in CMAKE_INSTALL_PYTHONDIR.
```
Is this no longer applicable? (eg. maybe there is nothing less 'like' `TPython`?)

guitargeek: You are right, I'll drop that commit. I still don't like the symlinks though, because some ways of shipping ROOT don't allow for them. But I'll work on that in other PRs.

Issue #19045, opened 2025-06-14
The 3D visualization tutorial glViewerExercise.C doesn't work with web graphics


guitargeek:
### Description

While you're going to see the initial visualization, the tutorial will error out before finishing.

### Reproducer

```
ssh -X -Y rembserj@lxplus9.cern.ch # you need trusted X11 forwarding

# download the tutorial files
wget "https://raw.githubusercontent.com/root-project/root/refs/heads/master/tutorials/visualisation/gl/nucleus.C"
wget "https://raw.githubusercontent.com/root-project/root/refs/heads/master/tutorials/visualisation/gl/glViewerExercise.C"

# source ROOT 6.36.00, for example
source /cvmfs/sft.cern.ch/lcg/app/releases/ROOT/6.36.00/x86_64-almalinux9.5-gcc115-opt/bin/thisroot.sh

# run the tutorial
root glViewerExercise.C
```

While you're going to see the initial visualization, the tutorial will quit prematurely with the following error:
```txt
   ------------------------------------------------------------------
  | Welcome to ROOT 6.36.000                       https://root.cern |
  | (c) 1995-2025, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on May 25 2025, 16:48:27                 |
  | From tags/v6-36-00@v6-36-00                                      |
  | With c++ (GCC) 11.5.0 20240719 (Red Hat 11.5.0-5)                |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------

root [0]
Processing glViewerExercise.C...
Info in <TGeoManager::TGeoManager>: Geometry nucleus, Model of a nucleus created
Info in <TGeoManager::SetTopVolume>: Top volume is WORLD. Master volume is WORLD
Info in <TGeoNavigator::BuildCache>: --- Maximum geometry depth set to 100
Info in <TGeoManager::CheckGeometry>: Fixing runtime shapes...
Info in <TGeoManager::CheckGeometry>: ...Nothing to fix
Info in <TGeoManager::CloseGeometry>: Counting nodes...
Info in <TGeoManager::Voxelize>: Voxelizing...
Info in <TGeoManager::CloseGeometry>: Building cache...
Info in <TGeoManager::CountLevels>: max level = 1, max placements = 100
Info in <TGeoManager::CloseGeometry>: 101 nodes/ 3 volume UID's in Model of a nucleus
[...]

PR #19047, opened 2025-06-16
[tutorials] check gPad in gl tutorials

linev:
If run in web mode canvas not automatically created and therefore `gPad` can be `nullptr`. 
Add simple check to avoid crash.

Disable web mode for viewer3DLocal.C and viewer3DMaster.C while features with TBuffer3D not supported by web graphics yet.

Fixes https://github.com/root-project/root/issues/19045


dpiparo (approved):
Thanks a lot for these changes!

PR #19053, opened 2025-06-16
Globally enable sized deallocation functions

hahnjo:
They are part of C++14 and are turned on by default starting with Clang 19. Pass the explicit compiler option `-fsized-deallocation` to always enable it even for older compilers, also fixing the test `roottest-root-meta-runMemberComments` when compiling with Clang 18 (or older) on Linux.


interpreter/cling/lib/Interpreter/CIFactory.cpp
```diff
@@ -1382,11 +1382,6 @@ namespace {
       argvCompile.push_back("-fno-omit-frame-pointer");
 #endif
 
-#ifdef __cpp_sized_deallocation
-      // Propagate the setting of the compiler to the interpreter
-      argvCompile.push_back("-fsized-deallocation");
```
pcanal, line 1387: Isn't it not yet the default for the Clang we are using?

pcanal (approved):
LGTM.  Thanks.

core/base/inc/TObject.h
```diff
@@ -190,11 +190,8 @@ class TObject {
    void    *operator new[](size_t sz, void *vp) { return TStorage::ObjectAlloc(sz, vp); }
    void     operator delete(void *ptr);
    void     operator delete[](void *ptr);
-#ifdef R__SIZEDDELETE
```
pcanal, line 193: One slight concern is how does this affect user code compilation if/when they do not also specify the `-fsized-allocation` flag?

interpreter/cling/lib/Interpreter/CIFactory.cpp
hahnjo: No, it's not yet the default. But I believe enabling it belongs in ROOT's `TCling` constructor.

core/base/inc/TObject.h
hahnjo: If https://en.cppreference.com/w/cpp/memory/new/operator_delete is anything to go by, the "Class-specific usual deallocation functions" are not "since C++14". I guess they might not be called, but defining them seems fine judging by my local test and the CI on macOS.

devajithvs (approved):

PR #19054, opened 2025-06-16
[roottest] set build dir when compile macro

linev:
Try to avoid creation of many temporary files in source directory where macro is situated. 
Most significant drawback has on Windows when several files has to be moved or deleted from source directory

Need to adjust several existing tests to correctly configure fixtures (dependencies) between tests.
Re-enable `cling/templateMembers/runtemplateMembers` test which was not included in cmake before, but exists in Makefile

Several failures was due to fact that both compiled and interpreted version of macro need to be running.
To exclude interference - I add deletion of compiled code and configure strict dependency:
```
      PRECMD ${CMAKE_COMMAND} -E remove -f ${CMAKE_CURRENT_BINARY_DIR}/runvbase_C${CMAKE_SHARED_LIBRARY_SUFFIX} ${CMAKE_CURRENT_BINARY_DIR}/runvbase_C_ACLiC_dict_rdict.pcm
```


pcanal:
I am exactly sure why it works before and not after.  The test failures seems to indicates that the ACLiCed library is being loaded both explicitly (as intended by the test) and implicitly (autoloading).  It might be related to (before the rootmap file would not be copied, after the rootmap file would be built in the run-time/binary directory)

```
# Select whether and how ACLiC tracks the dependency of the libraries.
#  0 - No tracking
#  1 - [Default] New libraries are explicitly linked to all currently loaded libraries
#  2 - Generate and use rootmap file for each library
#  3 - Generate rootmap file for each library _and_ explicitly link to the
#      needed libraries.
# On Windows, the default is 3
#ACLiC.LinkLibs:      1
```

linev:
Most (all?) failures due to fact that both compiled and interpreted version of macro need to be running.

To exclude interference - I will try to delete all libs first, then run interpreted version and finally run ACLiC version.

linev:
@pcanal After fixing dependency new approach works

pcanal:
> Several failures was due to fact that both compiled and interpreted version of macro need to be running.

It should not be a problem and is not on Linux ...  What appended when using the `.rootrc`:
```
Rint.History:  .root_hist
ACLiC.LinkLibs:  1
```
(i.e. the one created by roottest's cmake)?

linev:
> It should not be a problem and is not on Linux ... What appended when using the .rootrc:

Main problem was Windows. 
And use of `-e "gSystem->SetBuildDir(\"${CMAKE_CURRENT_BINARY_DIR}\", true)"` only has real effect on Windows.
And only use of `FIXTURES_SETUP/FIXTURES_REQUIRED` helps to solve the problem

pcanal:
> And only use of FIXTURES_SETUP/FIXTURES_REQUIRED helps to solve the problem

Yes but make the test serial ....

What was the result of not using those fixture and using the 'correct' `.rootrc` to prevent the generation of the rootmap file on Windows?

linev:
> Yes but make the test serial ....

But I see no other chance to run same macro in interpreted and in compiled mode in same directory.
First delete all compiled artifacts, then run macro in interpreted mode, and then run macro with ACLiC.
And `.rootrc` had no special entries compared to default one.

pcanal:
> And .rootrc had no special entries compared to default one.

If I am not missing something, the removed `.rootrc` had only:
```
Rint.History:  .root_hist
```
The one auto generated have:
```
Rint.History:  .root_hist
ACLiC.LinkLibs:  1
```
The default are expressed here:
```
# Select whether and how ACLiC tracks the dependency of the libraries.
#  0 - No tracking
#  1 - [Default] New libraries are explicitly linked to all currently loaded libraries
#  2 - Generate and use rootmap file for each library
#  3 - Generate rootmap file for each library _and_ explicitly link to the
#      needed libraries.
# On Windows, the default is 3
#ACLiC.LinkLibs:      1
```
So with the removed `.rootrc` on Linux we had effectively:
```
Rint.History:  .root_hist
ACLiC.LinkLibs:  1
```
so no change ... but on Windows we had:
```
Rint.History:  .root_hist
ACLiC.LinkLibs:  3
```
so on Windows it is changed to no longer produce the rootmap file .... which should have helped ... but isn't :(  ... so case closed.

PR #19056, opened 2025-06-16
Added support for e2k (Elbrus) arch

zubankov:
# This Pull request:
Added support for e2k (Elbrus) arch
## Changes or fixes:
The changes consist of:
1. e2k arch support for llvm-project with root patches.
These changes are based on [llvm-project-e-2-k](https://gitflic.ru/project/e2khome/llvm-project-e-2-k/)
It can be omitted, but it is required for successful build and work.
2. e2k arch support for cling and root.
For build it is necessary to use some cmake flags due to specificity of e2k arch:
    -DBUILD_SHARED_LIBS:BOOL=ON
    -DCMAKE_BUILD_TYPE=MinSizeRel
    -DLLVM_TARGET_ARCH=e2k
    -DLLVM_DEFAULT_TARGET_TRIPLE=e2k64-unknown-linux-gnu
    -DLLVM_WITH_LCCRT=/usr/lib64/lccrt
    -DLLVM_LCCRT_RELEASE=True
    -Dclad:BOOL=OFF

## Checklist:

- [x] tested changes locally on Elbrus-8c2 and Elbrus-2c3 processors

image


dpiparo:
Thanks for these changes. However, ROOT does not seem the right place to integrate them.
I think the llvm main repo is a more correct target. Once there, these changes will be usable for ROOT.

silverweed:
As an addendum, this kind of huge PRs should be discussed with the ROOT team beforehand, to avoid lots of effort going wasted.
The PR itself is not in a good reviewable state (only 2 commits for 190 files, dead code left in such as `#if 1` or commented-out parts, [seemingly-arbitrary decisions](https://github.com/root-project/root/pull/19056/files#diff-a70ef53467b2a7d7a69e07269174f3d24765a34fe56deafbf785a3075695f700R351) and so on).
If support for e2k is required the best course of action is opening an issue and starting a discussion on how and if we should try and add it.

zubankov:
Thank you for your comments! Yes, absolutely agree that it's right to add e2k support to the main llvm repo.
However, I am not the developer of the [llvm port on e2k arch](https://gitflic.ru/project/e2khome/llvm-project-e-2-k/), and I don't know why developers of llvm-e2k don't add their pacthes to the main llvm repo.

Perhaps I should do a pull request with changes concerning only adding e2k arch support to root and cling with no changes to llvm (see second commit)?

pcanal:
> and I don't know why developers of llvm-e2k don't add their pacthes to the main llvm repo.

This is a serious problem.  The change in the ROOT code per se are limited and could be reviewed/improved. However incorporating them without the changes in LLVM seems unhelpful.  The change in LLVM are way too large for us to have the effort to maintain them (or even to deal with them any time we upgrade our version of LLVM).    Please get back to us as soon as the llvm port to e2k has been successfully incorporated in LLVM.

zubankov:
Thank you! I hope that e2k support will be added to the main llvm repo someday.

PR #19058, opened 2025-06-17
[cmake] add -Wno-error to mathcore and minuit

silverweed:
In MathCore and Minuit there are warnings stemming from -Winconsistent-missing-override (like:
 'Gradient' overrides a member function but is not marked 'override').

Applying the suggested fix doesn't work because the functions in question are only conditionally overriding a base class function, depending on the type used as their CRTP argument.

This should probably be fixed in a better way, but in the meantime this commit allows building ROOT with dev=ON, which now fails.

### DIFF
--- a/math/mathcore/CMakeLists.txt
+++ b/math/mathcore/CMakeLists.txt
@@ -203,6 +203,10 @@ ROOT_STANDARD_LIBRARY_PACKAGE(MathCore
     ${MATHCORE_BUILTINS}
 )
 
+if(NOT MSVC)
+  target_compile_options(MathCore PRIVATE -Wno-error)
+endif()
+
 target_include_directories(MathCore PRIVATE ${Vc_INCLUDE_DIR})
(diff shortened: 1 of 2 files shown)


ferdymercury:
Alternative suggestion:

use in the five involved places:

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Winconsistent-missing-override"
  virtual Gradient()...;
#pragma GCC diagnostic pop
```

silverweed:
The proper solution would be to not make that function conditionally virtual, but that requires some thought and knowledge of the code which I don't have at the moment..

Silencing the warning should be done only if no good alternative is found.

dpiparo:
@lmoneta

ferdymercury:
Alternative PR: https://github.com/root-project/root/pull/19070

silverweed:
Superseded by https://github.com/root-project/root/pull/19070

PR #19060, opened 2025-06-17
[ntuple] Add some Internal helpers needed for Attributes

silverweed:
- const version of GetProjectedFieldsOfModel()
- RPageSink::GetUnderlyingDirectory()

These functionalities are needed in the RNTuple Attribute prototype (they are both Internal so there is no long-term implication of adding them).

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/inc/ROOT/RPageStorage.hxx
+++ b/tree/ntuple/inc/ROOT/RPageStorage.hxx
@@ -313,6 +313,8 @@ public:
 
    virtual ROOT::NTupleSize_t GetNEntries() const = 0;
 
+   virtual TDirectory *GetUnderlyingDirectory() const { return nullptr; }
+
    /// Physically creates the storage container to hold the ntuple (e.g., a keys a TFile or an S3 bucket)
    /// Init() associates column handles to the columns referenced by the model
    void Init(RNTupleModel &model)
--- a/tree/ntuple/inc/ROOT/RMiniFile.hxx
+++ b/tree/ntuple/inc/ROOT/RMiniFile.hxx
@@ -33,7 +33,10 @@ class TVirtualStreamerInfo;
 namespace ROOT {
 
 namespace Internal {
+class RNTupleFileWriter;
 class RRawFile;
+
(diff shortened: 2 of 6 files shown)


tree/ntuple/inc/ROOT/RPageStorage.hxx
```diff
@@ -313,6 +313,8 @@ public:
 
    virtual ROOT::NTupleSize_t GetNEntries() const = 0;
 
+   virtual TDirectory *GetUnderlyingDirectory() const { return nullptr; }
```
hahnjo, line 316: This breaks the abstraction of `RPageSink` quite heavily, since not all sinks are a file and even of those not all have a `TDirectory`. I'm not sure this makes sense to ask a `TDirectory` from all page sinks...

hahnjo: I guess the wanted functionality is to open a new page sink that ends up in the same storage? Maybe we can have an interface `virtual std::unique_ptr<RPageSink> CreateAtSameLocation(std::string_view ntupleName)` that throws if not implemented by a sink, or if the sink is unable to provide that functionality (simple `RMiniFile`, for the moment)?

silverweed: True, but for the prototype Attributes only work with TFile-based RNTuples and breaking the abstraction makes it much easier to deal with...at this point maybe the prototype should not be merged in master yet, until there is a more finalized design?
Either that or we merge it and later change it (making sure everything is in Internal/Experimental so we can do that easily)

hahnjo: For the prototype itself, I think we cannot merge until we are sure the on-disk format is right because it requires bumping (at least) the minor version. In the past, we did merge interface extensions to make work-in-progress prototypes easier (my work on distributed parallel writing, the `RNTupleProcessor` while it was / is still evolving), but I think we should still strive to get the interfaces right (my 2 cents at least).

silverweed: Good point on the binary format...we definitely cannot merge that until we're sure.

PR #19061, opened 2025-06-17
[Python][UHI] Extend TH1 equality operator to consider additional properties

### LINKED ISSUE #19038: [Python] TH1 equality operator pythonization not appropriate for ROOT histograms
### Description

The UHI implementation of the [equality operator](https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_uhi.py#L382) only compares the histogram type, binning and counts.

ROOT histograms have way more properties, like drawing options, and even more crucially they store the sum-of-weights-squared, where different values make a big difference in the statistical interpretation (the errorbars are directly inferred from it).

Therefore, I think the currently-implemented equality check is oversimplifying things and therefore potentially dangerous.

Furthermore, the new equality operator marks a significant behavior change that is [undocumented in the release notes](https://root.cern/doc/v636/release-notes.html), since previously it was doing a comparison by pointer. This should be documented in the release notes if we decide to keep the UHI implementation.

We have examples in the ROOT code base itself, where the equality operator wrongly assumed that it checks for pointer equality:
  * https://github.com/root-project/root/pull/19015
  * https://github.com/root-project/root/issues/19035

It's likely that there is user code that there is user code that makes the same assumption, which would be broken in 6.36 with the UHI.

### Reproducer

```python
import ROOT

hist1 = ROOT.TH1D("hist1", "hist1", 1, -1., 1.)
hist1.Fill(0., 2.)
hist1.Fill(0., 2.)
print(hist1.GetSumw2()[1])

hist2 = ROOT.TH1D("hist2", "hist2", 1, -1., 1.)
hist2.Fill(0., 4.)
print(hist2.GetSumw2()[1])

print(hist1 == hist2) # ROOT will tell you the histograms are identical, while they have different sumw2!
```
The output will be:
```
8.0
16.0
True
```

siliataider:
# This Pull request:

## Changes or fixes:
The equality operator for ROOT histograms now compares the following properties:

- Histogram type (including dimension).
- Histogram kind.
- Shape (number of bins per axis).
- Bin contents array.
- Sum-of-weights-squared array.
- Axes properties.

## Checklist:

- [x] tested changes locally

This PR fixes #19038

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_uhi.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_uhi.py
@@ -386,6 +386,9 @@ def _eq(self, other):
         isinstance(other, type(self))
         and _shape(self) == _shape(other)
         and np.array_equal(_values_default(self), _values_default(other))
+        and self.kind == other.kind
+        and np.array_equal(_get_sum_of_weights_squared(self), _get_sum_of_weights_squared(other))
+        and all(a == b for a, b in zip(self.axes, other.axes))
     )
 
 
@@ -559,11 +562,12 @@ def _get_sum_of_weights_squared(self) -> np.typing.NDArray[Any]:  # noqa: F821
     import numpy as np
 
     shape = _shape(self, include_flow_bins=False)
-    return np.frombuffer(
+    sumw2_arr = np.frombuffer(
         self.GetSumw2().GetArray(),
         dtype=self.GetSumw2().GetArray().typecode,
         count=self.GetSumw2().GetSize(),
-    ).reshape(shape, order="F")[tuple([slice(1, -1)] * len(shape))]
+    )
+    return sumw2_arr[tuple([slice(1, -1)] * len(shape))].reshape(shape, order="F") if sumw2_arr.size > 0 else sumw2_arr
 
 
 values_func_dict: dict[str, Callable] = {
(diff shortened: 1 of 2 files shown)


vepadulano (approved):
LGTM, but let's wait for discussion with @guitargeek

siliataider:
@guitargeek I did not add checks for properties related to the GUI. We should discuss if those are part of determining whether 2 ROOT histograms are equal or not

dpiparo:
A discussion might be needed. I personally lean towards not considering aspects such as colours or marker type part of the equality: they are not related to the statistical properties of the histos and in any case those will not be part of the new histograms...

vepadulano:
> I think we should keep this new equality operator only if there is a clear motivation from a user perspective,

While I see your point and I understand this new behaviour might lead to potential confusion for downstream users in the wild, let me stress that the pre-existing behaviour in master was simply wrong and a bug. If we decide against merging this PR, imagining that we find a way to accommodate the requirements for the UHI testing suite without using this equality operator, I will request that the equality operator is simply disabled for TH1 and derived classes, i.e. `h1 == h2` should throw an exception.

guitargeek:
Yes, okay then we're on the same page here!

> I will request that the equality operator is simply disabled for TH1 and derived classes, i.e. `h1 == h2` should throw an exception.

I think that would mean kicking out this pythonization:
https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/src/TObjectPyz.cxx

Then we don't even need to throw an exception but it's simply not implemented.

siliataider:
Closed in favor of #19209

PR #19070, opened 2025-06-17
[Math] Move gradient methods to IBaseFunction classes

guitargeek:
Move the gradient-related methods from the
`IGradientFunctionMultiDimTempl` to the `IBaseFunctionMultiDimTempl` base class, and the same for the 1D version.

This makes the `IGradientFunction` classes pure "indicator" classes that indicate whether the gradient is implemented by the user or not. The actual function interface is unified in the IBaseFunctions.

This fixes the problem that the override keywords could not be used consistently in I*Function-derived classes where the base class was templated to be either `IBaseFunction` or `IGradientFunction`.

Alternative to:
  * https://github.com/root-project/root/pull/19058

### DIFF
--- a/math/mathcore/inc/Fit/Chi2FCN.h
+++ b/math/mathcore/inc/Fit/Chi2FCN.h
@@ -120,7 +120,7 @@ public:
    }
 
    // need to be virtual to be instantiated
-   virtual void Gradient(const double *x, double *g) const {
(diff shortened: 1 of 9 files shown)


ferdymercury:
Thanks for tackling this!

Besides the failing test, roofit/roofit/test/GaussFunction.h needs an override, too

PR #19071, opened 2025-06-17
[ci] Add Fedora 43 to the CI

guitargeek:
Fedora 43 is not released yet, so it's marked as a special build.

The reason why it's added to the CI already early is that it already provides Python 3.14, which will be released in 3 months. So it's very useful for testing the ROOT Python interfaces with the upcoming Python version.


dpiparo:
Thanks for this addition! We have some reports about Python 14 already, and it's useful to be able to test it.

dpiparo (approved):

PR #19074, opened 2025-06-18
[roottest] try to use XML files without escape symbols

linev:
On some Windows nodes there were problems to parse original files, try to restore them


bellenot (approved):

pcanal:
Thanks.

PR #19086, opened 2025-06-18
[roottest] dismiss `RootExeOptions` and `RootExternalIncludes` cmake variables

linev:
Both variables adding custom arguments to `root.exe` invocation in the roottest macros.
Main disadvantage of such variables - all consequent tests will use such variables - 
which was not necessary in several cases, especially in `root/tree/cloning` tests.

All this can be replaced by `ROOTEXE_OPTS` argument of `ROOTTEST_ADD_TEST` macro.
It assigned per-test and therefore more precise.

In the tests where such options were replaced also fix dependencies using fixtures - 
which is more reliable than use of `DEPENDS` arguments.


roottest/cling/staticinit/ROOT-10426/CMakeLists.txt
```diff
@@ -4,14 +4,22 @@
...
 
-add_library(ROOT10426 SHARED
-   ROOT10426.cxx)
+add_library(ROOT10426 SHARED ROOT10426.cxx)
 target_link_libraries(ROOT10426 PUBLIC ROOT::Core)
 
-ROOTTEST_ADD_TEST(ROOT-10426
-   MACRO  execROOT10426.C
-   OUTREF execROOT10426.ref)
+add_test(NAME roottest-cling-staticinit-ROOT-10426-lib
```
pcanal: Is this needed?  Isn't the library build during the main ROOT build?

linev: But idea not to depend from ROOT build.

roottest/root/io/stdpair/CMakeLists.txt
```diff
@@ -1,28 +1,22 @@
...
 
-ROOT_GENERATE_DICTIONARY(G__CmsPairCollection ${CMAKE_CURRENT_SOURCE_DIR}/cmspair.h LINKDEF FullLinkDef.h)
-ROOTTEST_LINKER_LIBRARY(CmsPairCollection TEST G__CmsPairCollection.cxx LIBRARIES ${ROOT_LIBRARIES})
+ROOTTEST_COMPILE_MACRO(cmspair.h
+                       FIXTURES_SETUP root-io-stdpair-compile-fixture)
 
-ROOT_GENERATE_DICTIONARY(G__CmsPair ${CMAKE_CURRENT_SOURCE_DIR}/cmspair.h LINKDEF PairLinkDef.h)
-ROOTTEST_LINKER_LIBRARY(CmsPair TEST G__CmsPair.cxx LIBRARIES ${ROOT_LIBRARIES})
-
-ROOT_ADD_TEST(roottest-root-io-stdpair-collection-build
```
pcanal, line 12: Changing this may or may not be the right direction.  However it 'looks' like a noticeable change (see also https://github.com/root-project/roottest/pull/858) that does not appear directly link to `RootExeOptions`.   Can you open a separate PR for this?  Thanks.

linev: Changing `RootExeOptions` is trivial in most cases. 
But now split them will be too much work.

roottest/root/io/transient/base/CMakeLists.txt
```diff
@@ -1,25 +1,27 @@
...
-set(CMAKE_ROOTTEST_NOROOTMAP OFF)
-ROOT_GENERATE_DICTIONARY(base_cling ${CMAKE_CURRENT_SOURCE_DIR}/base.h MODULE base LINKDEF baseLinkDef.h)
-ROOTTEST_LINKER_LIBRARY(base base_cling.cxx testobject.cpp testobjectderived.cpp TEST LIBRARIES ROOT::Core ROOT::RIO)
-
-set(RootExeOptions -e "gSystem->Load(\"libbase\")")   # Adding a ; to avoid the output is complicated
+ROOTTEST_GENERATE_DICTIONARY(libRoottestIoTransientBase
+                             testobject.h testobjectderived.h
+                             LINKDEF baseLinkDef.h
+                             SOURCES testobject.cpp testobjectderived.cpp
+                             NO_CXXMODULE
```
pcanal: This seems to be a behavior change (vs `MODULE base`).  Why is it advantageous?

linev: I had failures on some platforms - because of auto loading with hadd

roottest/root/io/transient/base/execWriteFile.ref
```diff
@@ -1,5 +1,4 @@
 
-(int) 0
```
pcanal, line 2: I am vaguely surprised those disappeared ... (this seems related to the comment at line 7 of roottest/root/io/transient/base/CMakeLists.txt)

roottest/root/tree/cloning/CMakeLists.txt
```diff
@@ -12,97 +12,81 @@ else()
    set(test_bin_dir "${CMAKE_BINARY_DIR}/test")
 endif()
 
-set(RootExeOptions -e "gSystem->Load(\"${test_bin_dir}/libEvent\")")
+ROOTTEST_ADD_TEST(hsimple-create
+                  MACRO ${CMAKE_SOURCE_DIR}/tutorials/hsimple.C
+                  PASSRC 255
+                  FIXTURES_SETUP root-tree-cloning-hsimple-fixture)
```
pcanal, line 18: This is now technically "redundant" with the last step of the ROOT build (i.e. we are now always in the else branch (line 40 of the previous version of this file.

roottest/root/tree/cloning/CMakeLists.txt
```diff
@@ -12,97 +12,81 @@ else()
...
+                  COMMAND ${CMAKE_COMMAND} -E copy hsimple.root hsimple1.root)
+
+ROOTTEST_ADD_TEST(hsimple-copy2
+                  FIXTURES_REQUIRED root-tree-cloning-hsimple-fixture
+                  FIXTURES_SETUP root-tree-cloning-hsimple2-fixture
+                  COMMAND ${CMAKE_COMMAND} -E copy hsimple.root hsimple2.root)
+
+ROOTTEST_ADD_TEST(hsimple-clone
+                  MACRO run.C
+                  FIXTURES_REQUIRED root-tree-cloning-hsimple1-fixture root-tree-cloning-hsimple2-fixture)
```
pcanal, line 32: Thanks!

roottest/root/io/stdpair/CMakeLists.txt
pcanal: Ok. Bear with me while I look at this one directory a bit longer.

roottest/root/io/transient/base/CMakeLists.txt
pcanal: We should add a comment here of what those failures are and why they require the no cxxmodule.  However I also don't see immediately why it worked before and now fails.

roottest/root/io/transient/base/execWriteFile.ref
linev: Yes, I was using similar trick in different places with:
```
     ROOTEXE_OPTS -e "(void)gSystem->Load(\"libRoottestIoTransientBase\")"
```

roottest/root/tree/cloning/CMakeLists.txt
linev: This really ensure that `hsimple.root` in **CURRENT** directory is created.
Instead copying it from `tutorials/` sub-directory.

roottest/root/io/transient/base/CMakeLists.txt
linev: To my understanding - building standard ROOT library also creates entry in the modules - which then used by `hadd` for autoloading. And somehow modules does not work with `ROOTTEST_GENERATE_DICTIONARY`.

linev: I remove commit from the PR where `io/transient/base` should use `ROOTTEST_GENERATE_DICTIONARY` macro. Seems to be it is not flexible enough to create dictionary, library and correct module for them. 
Therefore I dismiss it. Later one can try again.

roottest/root/tree/cloning/CMakeLists.txt
pcanal: Fair enough.  Thanks.

roottest/root/io/transient/base/execWriteFile.ref
pcanal: Ah. I see.  I had missed the additional `(void)`.  Thanks.

roottest/cling/staticinit/ROOT-10426/CMakeLists.txt
pcanal: This is a double edge sword.  On the hand it does what you want, on the other want it does what you ask :) .... The literal consequence (seen in particular in the windows build) is that this can trigger a (almost) full rebuild of ROOT during the testing (something to do with the granularity of the timestamp) ... and so instead we had consider going the other direction (all builds at build time) but have neither made a decision not did the investigation to really look into the related issues.  

For now, I would not add this kind of test unless they are necessary.

linev: Idea was to formulate clear dependency between building library and running test.

But I will revert this change.

roottest/root/io/stdpair/CMakeLists.txt
```diff
@@ -1,28 +1,22 @@
...
-              FIXTURES_SETUP InputFiles)
-
-ROOTTEST_ADD_TEST(CopyRootFiles
+ROOTTEST_ADD_TEST(copyfiles
                   COMMAND ${CMAKE_COMMAND} -E echo "Copying root files"
                   COPY_TO_BUILDDIR aa02.root ab02.root
-                  FIXTURES_SETUP InputFiles)
-
-ROOTTEST_COMPILE_MACRO(cmspair.h)
+                  FIXTURES_REQUIRED root-io-stdpair-cmspaircoll-fixture root-io-stdpair-cmspair-fixture root-io-stdpair-compile-fixture
```
pcanal: This could be move to new line 31, right? (i.e. those requirement are not use in 'this' copy?)

pcanal (approved):
LGTM.

roottest/root/io/stdpair/CMakeLists.txt
linev: Yes, I moved fixtures to tests itself.

PR #19069, opened 2025-06-17
[cppyy] disable tests failing with mac-beta ARM64

Vipul-Cariappa:
Disabling tests that fail on mac-beta ARM64

PR #19088, opened 2025-06-18
[roottest] move reflex test from `io/evolution` to subdir

linev:
Generated reflex library may be incomplete when other tests running,
causing failure messages. Instead creating dependency for each tests
just move reflex dictionary generation to separate dir


pcanal (approved):
LGTM

pcanal:
> Generated reflex library may be incomplete when other tests running,
causing failure messages.

For the record what was the error message?

pcanal:
The `pyunittests-bindings-pyroot-cppyy-cppyy-test-datatypes` error seem unrelated.  It seems that the dictionary step is generating (unwanted) rootmap file ~but it is 'strange' it only fails on one platform~

Nevermind the last observation.  The Warning about the duplicate rootmap file are on other platforms but do not lead to a failure (I assume the output of the test is being ignored).

So those failures are likely to 'just' be not-yet-addressed new incompatibility between cppyy and MacOS 26/xcode

pcanal:
(and the test was disable in this PR: https://github.com/root-project/root/pull/19069 )

PR #19092, opened 2025-06-19
[ci] Fail CI if missing override instead of just warning

ferdymercury:
Currently, many override warnings remain unnoticed since the gcc-problem-matcher of the GitHub CI does not work correctly, and it's also easy to forget those annotations if you do not visit the "Files changed" tab.

If some devs later use dev=ON and testing=ON, this leads to failures for compilers that auto-enable -Winconsistent-missing-override.

To prevent this, we enable the suggest-override warning in the build of debian with dev=ON instead of in the one from alma, this way we are sure of always failing if some does a PR that forgets an override, forcing pull-requesters to fix their issues before merging.

Within this PR: also fix missing overrides uncovered by turning ON this flag in the dev=ON build.

### DIFF
--- a/roottest/root/io/clones/RtObj.h
+++ b/roottest/root/io/clones/RtObj.h
@@ -6,10 +6,5 @@ class RtObj : public T
   public:
     RtObj() {}
     RtObj( const T & val ) : T(val) {}
-    ClassDefT(RtObj,1)
+    ClassDefOverride(RtObj,1)
  } ;
-
-ClassDefT2(RtObj,T)
-
-ClassImpT(RtObj,T)
-
--- a/roottest/root/io/evolution/equivalent/ioReadRuleEquivalentCode.h
+++ b/roottest/root/io/evolution/equivalent/ioReadRuleEquivalentCode.h
@@ -4,7 +4,7 @@ class A:public TObject{
 
 public:
    A(){}
-   ClassDef(A,1)
+   ClassDefOverride(A,1)
 
 };
 
@@ -13,7 +13,7 @@ class B:public TObject{
 public:
    B(){}
    void initializeTransients(){transient_=2;}
-   ClassDef(B,1)
+   ClassDefOverride(B,1)
 private:
    double transient_; //!
 
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -369,7 +369,7 @@ jobs:
             is_special: true
           - image: alma8
           - image: alma9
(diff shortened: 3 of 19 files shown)


ferdymercury:
@pcanal I see a `fatal error: module file '/github/home/ROOT-CI/build/lib/MultiProc.pcm' is out of date and needs to be rebuilt: module file out of date` in https://github.com/root-project/root/actions/runs/15753057412/job/44402174612?pr=19092

linev:
Yes, this is also my observation.
But normally such error appears only when converting from old `Makefiles` to `cmake`-based testing. 
Then normal compiled options are applied - and on some CI platforms warnings automatically causing errors.

ferdymercury:
Thanks @linev for the feedback.

Btw, do you approve the changes in roottest/io ?

roottest/root/io/clones/RtObj.h
```diff
@@ -6,10 +6,5 @@ class RtObj : public T
   public:
     RtObj() {}
     RtObj( const T & val ) : T(val) {}
-    ClassDefT(RtObj,1)
+    ClassDefOverride(RtObj,1)
```
linev, line 9: Why using override here?
None of virtual method is override.
Also not sure if `ClassDefT` is equivalent to `ClassDefOverride`

roottest/root/io/evolution/equivalent/ioReadRuleEquivalentCode.h
```diff
@@ -4,7 +4,7 @@ class A:public TObject{
 
 public:
    A(){}
-   ClassDef(A,1)
+   ClassDefOverride(A,1)
```
linev, line 7: Same here - as long there are no new virtual methods override, there is no sense to use `ClassDefOverride`

linev:
> Btw, do you approve the changes in roottest/io ?

For this part clearly Philippe @pcanal is responsible

roottest/root/io/evolution/equivalent/ioReadRuleEquivalentCode.h
ferdymercury: It derives from TObject, which has itself a ClassDef with virtual functions.

within the derived class, if you define ClassDef again, it redefines those virtual.
So you need ClassDefOverride.

roottest/root/io/clones/RtObj.h
ferdymercury: ClassDefT is deprecated. It contains exactly the same definition as ClassDef.

ferdymercury: Wrt override: same explanation, if you derive from TObject, you need to use ClassDefOverride instead of ClassDef.

PR #19097, opened 2025-06-19
[test] Correct typo in linkdef pragma

dpiparo:
the class std::pair<int,std::vector<int>> was not selected correctly.

This macro, aclic'd in a test, sporadically failed on macbeta.


dpiparo:
This is a very sporadic failure. Thanks to @devajithvs we now know the fix is not enough (ROOT is anyway able to write pairs of T,V if T,V have a dictionary without a dictionary). Debugging further...

dpiparo:
even if not enough to fix the problem, something to be done anyway.

PR #19099, opened 2025-06-19
[GenVector] Make most vector and point classes nothrow move constructible.

### LINKED ISSUE #19042: Ensure basic data classes are nothrow_move_constructible

Consider the following struct:

```c++
struct mylovelyDataType{
   std::vector<std::vector<double>> massivedata;
  TLorentzVector roottype;
}
```

The presence of TLorentzVector here caused `(std::is_nothrow_move_constructible_v<mylovelyDataType>)` to equal `false` this means that various cases cannot be optimized such as the expansion of the vector `std::vector<mylovelyDataType>`

This can be rectified by adding `noexcept` to the copy constructors of `TLorentzVector` in this example. But should be expanded to all basic data types.

### ROOT version

Latest version

### Installation method

Any

### Operating system

Any

### Additional context

_No response_

hageboeck:
- Make genvector classes nothrow move constructible. Old user-defined copy constructors and assignment operators were preventing the auto-generated ones.
- Add a test that statically asserts their moveability / trivially copyable
- ~~Fix a doxygen link [here](https://root.cern.ch/doc/master/classROOT_1_1Math_1_1Cartesian2D.html) The "see also" generates a dead link because doxygen tries to link to "Overview" and "the".~~

Fixes #19042

### DIFF
--- a/math/genvector/inc/Math/GenVector/Cartesian2D.h
+++ b/math/genvector/inc/Math/GenVector/Cartesian2D.h
@@ -47,7 +47,7 @@ public :
    /**
       Default constructor  with x=y=0
    */
-   Cartesian2D() : fX(0.0), fY(0.0)  {  }
+   constexpr Cartesian2D() noexcept = default;
 
    /**
       Constructor from x,y  coordinates
@@ -61,24 +61,6 @@ public :
    template <class CoordSystem>
    explicit constexpr Cartesian2D(const CoordSystem & v)
       : fX(v.X()), fY(v.Y()) {  }
-
-
-   // for g++  3.2 and 3.4 on 32 bits found that the compiler generated copy ctor and assignment are much slower
-   // re-implement them ( there is no no need to have them with g++4)
-   /**
-      copy constructor
-    */
-   Cartesian2D(const Cartesian2D & v) :
-      fX(v.X()), fY(v.Y())  {  }
-
-   /**
-      assignment operator
-    */
-   Cartesian2D & operator= (const Cartesian2D & v) {
-      fX = v.X();
-      fY = v.Y();
-      return *this;
-   }
 
    /**
       Set internal data based on 2 Scalar numbers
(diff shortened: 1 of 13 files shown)


math/genvector/inc/Math/GenVector/VectorUtil.h
```diff
@@ -39,7 +39,7 @@ namespace ROOT {
 
        @ingroup GenVector
 
-       @sa Overview of the @ref GenVector "physics vector library"
+       @sa GenVector "physics vector library"
```
ferdymercury:
```suggestion
       @note See also the overview of the @ref GenVector "physics vector library"
```

The \sa command is only guaranteed to work well if all what is inside are classes or function names / files separated by commas. Otherwise, it might or might not render correctly the output.

math/genvector/inc/Math/GenVector/Translation3D.h
```diff
@@ -45,7 +45,7 @@ namespace Impl {
 
     @ingroup GenVector
 
-    @sa Overview of the @ref GenVector "physics vector library"
+    @sa GenVector "physics vector library"
```
ferdymercury: Same comment for all other occurences

math/genvector/inc/Math/GenVector/VectorUtil.h
hageboeck: I'm not sure that's correct. Please check the current docs with two broken links: "Overview" and "the" go to `TRangeDynCast`.
image

Compare that to something that I generated *locally* from this PR:
image

Unless I'm using a different version of Doxygen, the changes proposed in the PR are better. I would only remove the extra "physics vector library" now.

ferdymercury: Yep, but I suggested changing \sa to \note to avoid that. See more info here: https://github.com/doxygen/doxygen/issues/11434

But yeah, removing "physics vector library" and leaving the \sa also works

hageboeck: Alright, I'm moving this out to #19110, so the CI doesn't need to rerun in case we want to iterate on the doxygen.

Issue #15922, opened 2024-06-25
TScatter palette settings


ferdymercury:
### Explain what you would like to see improved and how.

In the new TScatter class, it was found here https://root-forum.cern.ch/t/scatter-plot-with-colorbar/59562/14 that:

- There is not an easy way to control the color bar range
- Save TScatter object as .C and reload it, then the palette axis title is lost. The function SavePrimitive should be amended.

### ROOT version

```
ROOT v6.32.00
Built for linuxx8664gcc on May 28 2024, 05:02:02
From tags/v6-32-00@v6-32-00
With c++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Binary directory: /opt/root/bin
```

### Installation method

Binary Release

### Operating system

Ubuntu 22.04

ferdymercury:
Thanks for the fix @couet!

I would also suggest adding an alternative way (commented) in the tutorial to "zoom" rather than "setting user range", namely to use:

```
   scatter->GetHistogram()->SetMinimum(10);
   scatter->GetHistogram()->SetMaximum(200);
```

This way, the palette is also zoomed, unlike with SetUserRange, which just acts as a filter on what points to plot or not.

Also: there is a bug:
- If one calls SetMinimum only, then the plot is broken. If that's intended, then maybe we should create a function scatter->SetMinMax() so that people don't forget to put both.

- the behavior of the TH2F when zooming Z-axis is different from TScatter. In TH2F, colors outside the range are clipped to max or min color. Here in TScatter, instead of clipping, they are filtered out of the plot. I don't know if that's a good solution, but if it is intended, it should be documented in the Doxygen docu. If not inteded, then the color should be clipped to the max/min of the palette.

ferdymercury:
Thanks a lot for addressing one of the points! I am reopening to keep track of two minor bugs:

> Also: there is a bug:
> 
>     * If one calls SetMinimum only, then the plot is broken. If that's intended, then maybe we should create a function scatter->SetMinMax() so that people don't forget to put both.
> 
>     * the behavior of the TH2F when zooming Z-axis is different from TScatter. In TH2F, colors outside the range are clipped to max or min color. Here in TScatter, instead of clipping, they are filtered out of the plot. I don't know if that's a good solution, but if it is intended, it should be documented in the Doxygen docu. If not inteded, then the color should be clipped to the max/min of the palette.

couet:
[...]

PR #19101, opened 2025-06-19
Improvements in TScatter drawing.

couet:
If one calls scat->GetHistogram()->SetMinimum only, then the scatter plot drawing was broken.This PR protect this issue.
It was mentioned here: https://github.com/root-project/root/issues/15922. Add the new option SKIPCOL.


ferdymercury (approved):
Thanks!!

hist/hist/src/TScatter.cxx
```diff
@@ -45,9 +45,10 @@ End_Macro
 ### TScatter's plotting options
 TScatter can be drawn with the following options:
 
-| Option   | Description                                                       |
-|----------|-------------------------------------------------------------------|
-| "A"      | Produce a new plot with Axis around the graph |
+| Option    | Description                                                       |
+|-----------|-------------------------------------------------------------------|
+| "A"       | Produce a new plot with Axis around the graph |
+| "CLIPCOL" | Do not draw the points outside the color range. By default, such points' color is set to the minimum or maximum color, depending on whether the color is smaller or bigger than the color range |
```
ferdymercury:
```suggestion
| "SKIPCOL" | Do not draw the points outside the color range. By default, such points' color is clipped to the minimum or maximum color, depending on whether the color is smaller or bigger than the color range |
```

ferdymercury (changes requested):
Thanks a lot for this Olivier! Just one tiny change suggested.

ferdymercury (approved):

linev (approved):

ferdymercury:
To fix the failure, I believe you'll need to add option "SKIPCOL" to scatter->Draw("A"); in stressGraphics.cxx

test/stressGraphics_web.ref
```diff
@@ -35,7 +35,7 @@ Test#   SVG1Ref#  SVG1Err#  PDFRef#   PDFErr#   JPGRef#   JPGErr#   PNGRef#   PN
    34   1111600       100   3933582        50    248290     61958    298343    100544   1124318       700
    35      3851       100      9012       300     40000     11000     37000      9000      3851       100
    36     11750       100     41000       300     73043     25000     90000     37000     11725       100
-   37     29299       100     86156      3000     77792     25000     72945     24000     29299       100
+   37     26771       100     82724      3000     77792     25000     72945     24000     26771       100
```
ferdymercury, line 38: Thanks! Does this mean that at the moment, the web ignores the option "SKIPCOL" ?

linev: Yes, one need to implement such option

PR #19106, opened 2025-06-20
[reve] New demo for dark openui5 theme and single/detached view

alja:
This PR adds new demo and minor correction for single view type when view is not mandatory (e.g. its scenes are not streamed or subscribed to client by the default)

The screenshots is here of server windows in multi client mode:

Screenshot from 2025-06-19 18-10-27


alja:
@linev Please check.

PR #19109, opened 2025-06-20
[metacling] Destroy and release llvm::ManagedStatics

### LINKED ISSUE #18933: Make sure we call llvm_shutdown in the TCling destruction
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Currently when we destroy ROOT and TCling we do not properly shut down llvm causing the llvm::ManagedStatics to not be handled properly.

### Reproducer

N/A

### ROOT version

master

### Installation method

N/A

### Operating system

all

### Additional context

_No response_

aaronj0:
Fixes #18933 

Following up on this issue, CppInterOp now uses ManagedStatic to support a stack of multiple interpreters. Although we do not use this feature, this shutdown call is now required in TCling.


core/metacling/src/TCling.cxx
```diff
@@ -1365,6 +1365,9 @@ TCling::TCling(const char *name, const char *title, const char* const argv[], vo
    fCxxModulesEnabled = true;
 #endif
 
+   // Destroy llvm::ManagedStatic variables CppInterOp uses ManagedStatic to
+   // maintain a stack of multiple interpreters.
+   llvm::llvm_shutdown_obj shutdownTrigger;
```
vgvassilev: This will shut down llvm when tcling is constructed. We need to do this somewhere around `DestroyInterpreter`.

aaronj0: Isn't calling `llvm::llvm_shutdown()` what actually shuts down llvm? Of course I did not intend to shutdown in the constructor. If I understand correctly, this helper class does that for us when it is destroyed

vgvassilev: Yes, but when the helper class is destroyed? At the end of the ctor function, right?

aaronj0: Oh yes, I thought I added that as a member...

vgvassilev:
Can you check if the clad timing system works through setting the env variable?

guitargeek:
It doesn't work yet. Here is how you can test:
```diff
git a/interpreter/cling/tools/plugins/clad/CMakeLists.txt b/interpreter/cling/tools/plugins/clad/CMakeLists.txt
index c0316e679e0..bbebc2ab69a 100644
--- a/interpreter/cling/tools/plugins/clad/CMakeLists.txt
+++ b/interpreter/cling/tools/plugins/clad/CMakeLists.txt
@@ -74,7 +74,8 @@ if (DEFINED CLAD_SOURCE_DIR)
   list(APPEND _clad_extra_settings SOURCE_DIR ${CLAD_SOURCE_DIR})
 else()
   list(APPEND _clad_extra_settings GIT_REPOSITORY https://github.com/vgvassilev/clad.git)
-  list(APPEND _clad_extra_settings GIT_TAG v1.10)
+  list(APPEND _clad_extra_settings GIT_TAG master)
 endif()

 #list(APPEND _clad_patches_list "patch1.patch" "patch2.patch")
diff --git a/tutorials/roofit/roofit/rf101_basics.C b/tutorials/roofit/roofit/rf101_basics.C
index 969fed87b0c..dfb0be244fd 100644
--- a/tutorials/roofit/roofit/rf101_basics.C
+++ b/tutorials/roofit/roofit/rf101_basics.C
@@ -64,7 +64,7 @@ void rf101_basics()
    // -----------------------------

    // Fit pdf to data
-   gauss.fitTo(*data, PrintLevel(-1));
+   gauss.fitTo(*data, PrintLevel(-1), EvalBackend("codegen"));

    // Print values of mean and sigma (that now reflect fitted values and errors)
    mean.Print();
```

Then run inside `tutorials/roofit/roofit`:
```
CLAD_ENABLE_TIMING=1 EXTRA_CLING_ARGS="-Xclang -print-stats -ftime-report" root -b -q rf101_basics.C
```

If you were to change the managed static to a regular static in clad, it will work:
```diff
diff --git a/lib/Differentiator/Timers.cpp b/lib/Differentiator/Timers.cpp
index 67f3ec5..74de19d 100644
--- a/lib/Differentiator/Timers.cpp
+++ b/lib/Differentiator/Timers.cpp
@@ -100,8 +100,8 @@ public:
 CladTimeInfo* CladTimeInfo::TheTimingInfo;
 void InitTimers() {
   assert(!CladTimeInfo::TheTimingInfo);
-  static llvm::ManagedStatic<CladTimeInfo> CTI;
-  CladTimeInfo::TheTimingInfo = &*CTI;
+  static CladTimeInfo CTI;
+  CladTimeInfo::TheTimingInfo = &CTI;
 }

 TimedAnalysisRegion::TimedAnalysisRegion(llvm::StringRef Name) {
```

dpiparo:
Can we decide what to do about this item?

vgvassilev:
We need to understand why the `llvm::ManagedStatic`s are not destroyed.

PR #19117, opened 2025-06-20
Fix hardware_destructive_interference_size across TUs.

hageboeck:
ROOT might currently change the layout of classes by making
use of hardware_interference_size in publicly visible types. gcc
rightfully warns about that when the feature is used in a header. ROOT
works around the warning by assigning the value to another variable, but
this doesn't solve the issue.

Since the value of std::hardware_destructive_interference_size can vary
with compiler versions, standard library, or tune settings, it might
change across translation units in ROOT, breaking the ABI of RDF/RVec.
This, for example, happens when building with address sanitizer:
JITted code was trying to use the standard C++ value, whereas precompiled
code didn't even have the feature (it is suppressed when asan is used).
ASAN builds couldn't even be compiled, because ROOT unconditionally
tried to read the interference size for RVec.
Moreover, when users use an installed ROOT, they don't necessarily use
it on the same CPU / with the same tune settings that were used to
compile ROOT, think e.g. of ROOT on cvmfs.

For these reasons, the interference size is determined once at the CMake
configure stage, saved in RConfigure.h, and kept fixed afterwards.

### DIFF
--- a/cmake/modules/RootConfiguration.cmake
+++ b/cmake/modules/RootConfiguration.cmake
@@ -501,20 +501,26 @@ else()
    set(has_found_attribute_noinline undef)
 endif()
 
-# We could just check `#ifdef __cpp_lib_hardware_interference_size`, but on at least Mac 11
-# libc++ defines that macro but is missing the actual feature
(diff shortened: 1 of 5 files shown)


vepadulano (approved):
Nice thanks!

PR #19118, opened 2025-06-21
[test] Disable treeformula mapvector test for macos26

dpiparo:
until a solution for the underlying problem is identified.


pcanal:
Did we open an issue to keep track of the underlying problem?

PR #15790, opened 2024-06-07
[tmva][pymva] Improve evaluations in pymva and add support for Keras3 (TF >= 2.16)

lmoneta:
This Pull request starts adding support for new TF version (2.16) coming with Keras3 API. 

In addition, it improves the evaluation of models for regression and multi-class by using all events instead of the single event as before. With TF 2.16 s not possible anymore to disable eager execution and have a decent speed i evaluating the model

PR #19124, opened 2025-06-22
[CI] Disable PyMVA so we can support the newest TensorFlow versions

guitargeek:
Right now we don't support the newest TensorFlow releases after the 2.15 series, initially released in November 2023 [1].

Support for TensorFlow>=2.16 is worked on [2], but it's difficult to make progress if we don't have TensorFlow>=2.16 in our CI images. And we veto it from our CI image via the `requirements.txt` mechanism.

To continue, we need to temporarily disable PyMVA in the CI, such that we can include the newest TensorFlow version in our images, and then re-enable it in the PR that fixes support for the newest TensorFlow. This is what is done in this commit.

Supporting the newest TensorFlow became a bit more important now that we also have another feature besides PyMVA that uses TensorFlow: the RBatchGenerator. It would not good to hold off testing the actively-developed RBatchGenerator on newer TensorFlow versions, just because PyMVA prevents us from installing these on our CI.

[1] https://github.com/tensorflow/tensorflow/releases
[2] https://github.com/root-project/root/pull/15790


guitargeek:
The changes suggested by this PR should only be merged after the tutorial selection system is fixed by this other PR:
  * https://github.com/root-project/root/pull/19123

Like this, we avoid bad surprises when we eventually enable `tmva-pymva` again.

dpiparo:
@lmoneta @sanjibansg could you make sure the tests are correctly running on all platforms once this PR is merged and then re-enable them?

guitargeek:
Now that the tutorial tests were are restored by merging #19123, it became apparent that the `tmva-pymva` build option is not correctly considered in the tutorial vetos. So another PR is required before:
  * https://github.com/root-project/root/pull/19131

dpiparo (approved):
Thanks!

PR #19128, opened 2025-06-23
Speed up test compilation when Ninja is in use

hageboeck:
- Roottest tests were compiled sequentially when Ninja was in use, to avoid a problem (the commit doesn't mention details). Since that was 8 years ago, we should give parallel Ninja another try.

Difference on my machine (8 physical cores) running `ctest -R meta -j16`:
| Old | New |
|---|---|
| 760.82user 54.57system | 654.64user 52.71system |
| 6:31.81elapsed 208%CPU | 1:52.29elapsed 629%CPU |

**Note**:
- [x] I tested this locally on Gentoo, but our CI is kind of blind to running with Ninja. Maybe it's worth to take one of the distros and make it compile with Ninja?


dpiparo (approved):

dpiparo:
That is a good idea. We need to 1) Add ninja to one of the container images 2) build with ninja for that platform - this can be one of the experimental builds (clang? nomodules? marchnative?)

PR #19129, opened 2025-06-23
[v636][TMVA] Re-enable many tutorials for which we lost test coverage (PyMVA, SOFIE, RBatchGenerator)

guitargeek:
Backport of https://github.com/root-project/root/pull/19123.


guitargeek:
Pythia8 test failures unrelated.

PR #19130, opened 2025-06-23
[CMake] Three small updates

hageboeck:
- Mark two tutorials that are using numpy
- Make a CMake condition not depend on testing, but on the actual target that it's setting properties for
- Fix CMake with asan=On


guitargeek (commented):
Thank you!

guitargeek (approved):

PR #11506, opened 2022-10-06
cmake: Replace ever expanding modules_idx_deps cached variable

pcanal:
We replace it with a combination of a global property and an additional custom target.
The global property is used to collect the dependencies created before the modules.idx target is created.
The custom target is used as a direct dependency of the custom_command (add_dependencies can not be used on the target of a custom_command).  Then we can use
this custom target to collect the dependencies created after the modules.idx target.

This fixes #10510.

To test:

```
ls -rtl lib/modules.idx lib/libEve.so
rm lib/*Eve*
ninja # or make or cmake --build ; this should update modules.idx
ls -rtl lib/modules.idx lib/libEve.so
```

PR #19135, opened 2025-06-23
[CMake] Append to `CMAKE_INSTALL_RPATH` with `CACHE FORCE`

### LINKED ISSUE #19134: Installation issue for ROOT 6.36 RPATHs through Spack on MacOS
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Attempting to install ROOT 6.36 via the Spack package manager on MacOS (multiple versions and architectures on different machines) builds without error, but the installation step fails to properly set the RPATH and ROOT is unable to find core libraries when executing or linking other ROOT-based packages.

<details><summary>Runtime error</summary>
<pre>
```
dyld[79397]: Library not loaded: @rpath/libRint.6.36.so
  Referenced from: <27660B66-7EE2-3FB4-9855-E43EF17D9259> /Users/alfaci/jwilkins/spack/opt/spack/darwin-m1/root-6.36.00-s4iedayuyeec6kyav2zahhnvnepvy2gi/bin/root.exe
  Reason: tried: '/private/var/folders/8c/3pr52n2x4ng0j9p3rmqtfyvc0000gn/T/alfaci/spack-stage/spack-stage-root-6.36.00-s4iedayuyeec6kyav2zahhnvnepvy2gi/spack-build-s4ieday/lib/libRint.6.36.so' (no such file) [...]
```
</pre></details>

There are also several errors appearing in the install step from `install_name_tool`, complaining about duplicate entries in the RPATH:
<details><summary>install_name_tool errors</summary>
<pre>
```
error: /Library/Developer/CommandLineTools/usr/bin/install_name_tool: for: /Users/alfaci/jwilkins/FairSoft_spack_splitrepos/spack/opt/spack/darwin-m1/root-6.36.00-bceszqc7of6ksgxwrbcwya3gjw4vtmwo/bin/root (for architecture arm64) option "-add_rpath /Users/alfaci/jwilkins/FairSoft_spack_splitrepos/spack/opt/spack/darwin-m1/root-6.36.00-bceszqc7of6ksgxwrbcwya3gjw4vtmwo/lib" would duplicate path, file already has LC_RPATH for: /Users/alfaci/jwilkins/FairSoft_spack_splitrepos/spack/opt/spack/darwin-m1/root-6.36.00-bceszqc7of6ksgxwrbcwya3gjw4vtmwo/lib
error: /Library/Developer/CommandLineTools/usr/bin/install_name_tool: for: /Users/alfaci/jwilkins/FairSoft_spack_splitrepos/spack/opt/spack/darwin-m1/root-6.36.00-bceszqc7of6ksgxwrbcwya3gjw4vtmwo/bin/root (for architecture arm64) option "-add_rpath /Users/alfaci/jwilkins/FairSoft_spack_splitrepos/spack/opt/spack/darwin-m1/root-6.36.00-bceszqc7of6ksgxwrbcwya3gjw4vtmwo/lib64" would duplicate path, file already has LC_RPATH for: /Users/alfaci/jwilkins/FairSoft_spack_splitrepos/spack/opt/spack/darwin-m1/root-6.36.00-bceszqc7of6ksgxwrbcwya3gjw4vtmwo/lib64
```
</pre></details>

I'm able to fix the issue on my side by patching RootBuildOptions.cmake to revert https://github.com/root-project/root/pull/18277/ which appears for the first time in 6.36 - from some debugging (thanks also to @wdconinc) the issue appears to be some interaction between the `gnuinstall` flag in ROOT (which is applied by default in Spack), the explicitly set `CMAKE_INSTALL_RPATH`s from Spack, and the `list(APPEND CAKE_INSTALL_RPATH..)` in the current version of RootBuildOptions. I've tried adding various different flags to the recipe to fix the issue (e.g. explicitly setting `CMAKE_SKIP_INSTALL_RPATH` to false), but nothing seems to have helped.
The CMakeCache including the full set of CMake flags invoked by Spack is included under "Additional context"

See https://github.com/spack/spack-packages/issues/215 for more context (Bringing it upstream as there's a chance this also interacts with source building under some combination of flags).

### Reproducer

```sh
git clone https://github.com/spack/spack   # check out spack package manager
cd spack
source share/spack/setup-env.sh # Set up paths for spack
spack env create testenv  # Create and activate test environment
spacktivate testenv
spack add root@6.36.00 # Add ROOT version to environment spec
spack install # Concretise and install environment
root
```

### ROOT version

6.36.000

### Installation method

Package manager (Spack)

### Operating system

MacOS: 14.7.6 x86_64 ; 15.5 x86_64 ; 15.5 ARM64

### Additional context

[CMakeCache.txt](https://github.com/user-attachments/files/20862542/CMakeCache_arm64.txt)

guitargeek:
Correctly append the additional RPATHs set by the ROOT build system to the `CMAKE_INSTALL_RPATH` variable set by the user, which is cached in `CMAKE_INSTALL_RPATH-CACHED` by the ROOT build system.

Follows up on c77bd87.

Closes #19134.

Needs a backport to 6.36.

### DIFF
--- a/cmake/modules/RootBuildOptions.cmake
+++ b/cmake/modules/RootBuildOptions.cmake
@@ -429,10 +429,13 @@ if(rpath)
   if(APPLE)
     set(CMAKE_MACOSX_RPATH TRUE)
     set(CMAKE_INSTALL_NAME_DIR "@rpath")
-    list(APPEND CMAKE_INSTALL_RPATH @loader_path @loader_path/${BINDIR_TO_LIBDIR})
+
+    set(_rpath_values "@loader_path" "@loader_path/${BINDIR_TO_LIBDIR}")
   else()
-    list(APPEND CMAKE_INSTALL_RPATH $ORIGIN $ORIGIN/${BINDIR_TO_LIBDIR})
+    set(_rpath_values "$ORIGIN" "$ORIGIN/${BINDIR_TO_LIBDIR}")
   endif()
+
+  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH-CACHED};${_rpath_values}" CACHE STRING "Install RPATH" FORCE)
 
   unset(BINDIR_TO_LIBDIR)
 endif()


dpiparo:
Thanks for these changes!

cmake/modules/RootBuildOptions.cmake
```diff
@@ -429,11 +429,14 @@ if(rpath)
...
     set(CMAKE_INSTALL_NAME_DIR "@rpath")
-    list(APPEND CMAKE_INSTALL_RPATH @loader_path @loader_path/${BINDIR_TO_LIBDIR})
+
+    set(_rpath_values "@loader_path" "@loader_path/${BINDIR_TO_LIBDIR}")
   else()
-    list(APPEND CMAKE_INSTALL_RPATH $ORIGIN $ORIGIN/${BINDIR_TO_LIBDIR})
+    set(_rpath_values "$ORIGIN" "$ORIGIN/${BINDIR_TO_LIBDIR}")
   endif()
 
+  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_RPATH-CACHED};${_rpath_values}" CACHE STRING "Install RPATH" FORCE)
```
pcanal, line 438: Doesn't this increase the length of the `CMAKE_INSTALL_RPATH` for every run of cmake? (I.e. see `grep CMAKE_INSTALL_RPATH CMakeCache.txt; cmake .; grep CMAKE_INSTALL_RPATH CMakeCache.txt`)

pcanal: We had a similar pattern for modules_idx_deps (see #11506 ; the solution does not apply here but the challenge may (or may not) be the same).

guitargeek: It won't, because `CMAKE_INSTALL_RPATH-CACHED` is a custom ROOT mechanism that will explicitly only cache what is passed as CMake arguments:

https://github.com/root-project/root/blob/master/cmake/modules/CaptureCommandLine.cmake

PR #19136, opened 2025-06-23
[http] upgrade civetweb, enable more CORS headers

linev:
Unfortunately, since 2 years `civetweb` does not making official releases,
but there are many updates and fixes which are provided to master branch - also from us.
Therefore:

* update `civetweb` to recent master branch as of 23.06.2025 - https://github.com/civetweb/civetweb/tree/5864b55a94f4b5238155cbf2baec707f0fa2ba6d
* apply our small patch - exclude `openssl/engine.h` include
* add more CORS headers with `cred_cors` URL option, enables partial access to ROOT files with CORS and enabled authentication.

[Use-case from the forum](https://root-forum.cern.ch/t/thttpserver-with-cors-and-htdigest-auth/63769/) is not directly affected by this PR while there not ROOT but JSON files are requested - which do not use partial 206 requests.

### DIFF
--- a/net/http/civetweb/CREDITS.md
+++ b/net/http/civetweb/CREDITS.md
@@ -4,6 +4,7 @@
 * Abramo Bagnara
 * Adam Bailey
 * Adam Hunyadi
+* Adam Rowell
 * Alan Somers
 * Alberto Bignotti
 * Alex Kozlov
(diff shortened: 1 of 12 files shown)


linev:
Failure on mac-beta not related to PR, will be fixed by https://github.com/root-project/root/pull/19086/commits/bdd587aa6dd9a26a366d9c09724748c00d5833a1

Issue #19104, opened 2025-06-19
RDataFrame Reads Garbage Data instead of File


davidkowalk:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

This bug report is the result of this forum thread:

https://root-forum.cern.ch/t/rdataframe-reading-garbage-data/63763

When loading root files created by older versions RDataFrame reads arbitrary data instead of actual file.

### Reproducer

The following Root File 

```cpp
using namespace ROOT;

void check_photon_data() {
    RDataFrame df("t", "data/merged_20250610.root");

    auto calculate_photon_pair_mass = [](const RVec<TVector3> Ps, const RVec<double> Es) -> RVec<double> {
        //Calculates the invariant mass of each photon pair in hit
        //Later the hits with only etas can be counted.

        RVec<double> masses;
        for (size_t i = 0; i < Es.size(); ++i) {
                for (size_t j = i + 1; j < Es.size(); ++j) {
                    TLorentzVector p4_1, p4_2;
                    p4_1.SetVect(Ps[i]);
                    p4_1.SetE(Es[i]);
                    p4_2.SetVect(Ps[j]);
                    p4_2.SetE(Es[j]);
                    masses.push_back((p4_1 + p4_2).M());
                }
        }
        return masses;
    };

    auto df_out = df.Define("mass", calculate_photon_pair_mass, {"lP4BGONeutral.fP", "lP4BGONeutral.fE"})
                    .Define("p2",
                        [](const RVec<TVector3>& Ps){
                            RVec<double> p2;
                            for (TVector3 p:Ps) {
                                p2.push_back(p*p);
                            }
                            return p2;
                        },
                        {"lP4BGONeutral.fP"})
                    .Define("E2",
                        [](const RVec<double>& Es) {
                            RVec<double> E2;
                            for (double E:Es) {
                                E2.push_back(E*E);
                            }
                            return E2;
                        },
                        {"lP4BGONeutral.fE"})
                    .Define("nPhotonsP", [](const RVec<TVector3>& Ps){return Ps.size();}, {"lP4BGONeutral.fP"})
                    .Define("nPhotonsE", [](const RVec<double>& Es){return Es.size();}, {"lP4BGONeutral.fE"})
                    ;

    df_out.Display({"mass", "E2", "p2", "lP4Tagger.fE"}, 50) -> Print();

    df.Histo1D("lP4BGONeutral.fE") -> DrawClone();
[...]

PR #19140, opened 2025-06-23
Read correct size of TClonesArray collection value type

vepadulano:
Fxies #19104


ferdymercury:
Thanks!
Maybe related:
https://its.cern.ch/jira/browse/ROOT-8952
https://its.cern.ch/jira/browse/ROOT-7823

Issue #14012, opened 2023-11-02
Reloading sparsehist.C triggers an assert in LLVM


pcanal:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

```
root [0] .L sparsehist.C 
root [1] .x sparsehist.C 
Assertion failed: (isa<X>(Val) && "cast<Ty>() argument of incompatible type!"), function cast, file Casting.h, line 255.
Abort trap: 6
```

### Reproducer

See description and https://root-forum.cern.ch/t/break-segmentation-violation-when-executing-sparsehist-c/56948/5

### ROOT version

```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.31/01                        https://root.cern |
  | (c) 1995-2023, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for macosxarm64 on Nov 02 2023, 22:29:51                   |
  | From heads/master-13574@v6-31-01-108-g1cf66addc0                 |
  | With Apple clang version 15.0.0 (clang-1500.0.40.1)              |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------
```

### Installation method

local build

### Operating system

MacOS but likely any

### Additional context

Full trace:                                                                                                                                    
                                                                                                                        
```
1  __pthread_kill_implementation                                                             pthread_kill.c          44   0x7f2d9b3d29fc 
2  __pthread_kill_internal                                                                   pthread_kill.c          78   0x7f2d9b3d29fc 
3  __GI___pthread_kill                                                                       pthread_kill.c          89   0x7f2d9b3d29fc 
4  __GI_raise                                                                                raise.c                 26   0x7f2d9b37e476 
5  __GI_abort                                                                                abort.c                 79   0x7f2d9b3647f3 
6  __assert_fail_base                                                                        assert.c                92   0x7f2d9b36471b 
7  __GI___assert_fail                                                                        assert.c                101  0x7f2d9b375e96 
8  llvm::cast<clang::RecordType, clang::QualType>                                            Casting.h               566  0x7f2d8c3ad48d 
9  clang::Type::castAs<clang::RecordType>                                                    TypeNodes.inc           91   0x7f2d8cb47c4c 
10 (anonymous namespace)::CXXNameMangler::mangleUnqualifiedName                              ItaniumMangle.cpp       1568 0x7f2d920b02b8 
[...]

PR #19141, opened 2025-06-23
[cling] Handle unloading unnamed `VarDecl`s in catch statements

devajithvs:
Fixes the assertion error with the following:

```
[cling]$ #include <exception>
[cling]$ try {} catch (std::exception &) {}
[cling]$ .undo 1
```

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes https://github.com/root-project/root/issues/14012

### DIFF
--- a/interpreter/cling/lib/Interpreter/DeclUnloader.cpp
+++ b/interpreter/cling/lib/Interpreter/DeclUnloader.cpp
@@ -603,6 +603,15 @@ namespace cling {
     // * variables and parameters with dependent context;
     // * mangled names for parameters;
     if (!isa<ParmVarDecl>(VD) && !VD->getDeclContext()->isDependentContext()) {
+      // Exception variables without identifiers are not added to scope and will
+      // fail in the steps after the `if` block.
+      // Assuming this rule extends to non-exception variables too.
+      if (!VD->getIdentifier()) {
+        DeclContext* DC = VD->getLexicalDeclContext();
+        if (DC->containsDecl(VD))
+          DC->removeDecl(VD);
+        return true;
+      }
       // Cleanup the module if the transaction was committed and code was
       // generated. This has to go first, because it may need the AST
       // information which we will remove soon. (Eg. mangleDeclName iterates the
--- a/interpreter/cling/test/CodeUnloading/Exception.C
+++ b/interpreter/cling/test/CodeUnloading/Exception.C
@@ -1,0 +1,15 @@
+//------------------------------------------------------------------------------
+// CLING - the C++ LLVM-based InterpreterG :)
+//
+// This file is dual-licensed: you can choose to license it under the University
+// of Illinois Open Source License or the GNU Lesser General Public License. See
+// LICENSE.TXT for details.
+//------------------------------------------------------------------------------
+
+// RUN: cat %s | %cling 2>&1 | FileCheck %s
+try {} catch (int &) {}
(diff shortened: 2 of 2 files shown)


pcanal:
I am confused on 'why' this is the right fix?  The 'Skip' part of the description seems to indicate that the 'undo'/'unload' is being ignored for this example/case ... and thus (of course) no longer has negative consequence (i.e it may or may not just 'hide' the problem rather than solve it).  Could expand the commit log to explain why this is the right choice?

interpreter/cling/lib/Interpreter/DeclUnloader.cpp
```diff
@@ -602,7 +602,8 @@ namespace cling {
     // llvm::Module cannot contain:
     // * variables and parameters with dependent context;
     // * mangled names for parameters;
-    if (!isa<ParmVarDecl>(VD) && !VD->getDeclContext()->isDependentContext()) {
+    if (!isa<ParmVarDecl>(VD) && !VD->getDeclContext()->isDependentContext() &&
+        VD->getIdentifier()) {
```
vgvassilev: I imagine that the anonymous variables are given some internal name. Can we undo that instead of not undoing anything?

devajithvs:
> I am confused on 'why' this is the right fix? The 'Skip' part of the description seems to indicate that the 'undo'/'unload' is being ignored for this example/case ... and thus (of course) no longer has negative consequence (i.e it may or may not just 'hide' the problem rather than solve it). Could expand the commit log to explain why this is the right choice?

Thanks, that might have been an oversight. I was assuming that this would extend to all unnamed `VarDecls` 

In this snippet:

https://github.com/root-project/root/blob/master/interpreter/llvm-project/clang/lib/Sema/SemaDeclCXX.cpp#L17067


```
  // Add the exception declaration into this scope.
  if (II)
    PushOnScopeChains(ExDecl, S);
  else
    CurContext->addDecl(ExDecl);
```

The declaration is only pushed to the scope when it is named (i.e., when II is true). This means we don't have to go through the remaining steps when unloading an unnamed exception declaration.

I have updated the PR to separately handle unnamed VarDecls in catch statements.

interpreter/cling/lib/Interpreter/DeclUnloader.cpp
```diff
@@ -603,6 +603,14 @@ namespace cling {
     // * variables and parameters with dependent context;
     // * mangled names for parameters;
     if (!isa<ParmVarDecl>(VD) && !VD->getDeclContext()->isDependentContext()) {
+      // Exception variables without identifiers are not added to scope and will
+      // fail when tried to unload.
+      if (!VD->getIdentifier() && VD->isExceptionVariable()) {
```
vgvassilev: Isn’t that rule true for non-exception variables, too?

devajithvs: I can't find or imagine a case where non-exception `VarDecl`s are unnamed. So, I am not sure.

vgvassilev: Perhaps we can drop the second check or make it an assert?

devajithvs: Done!

pcanal (approved):
Thanks.

PR #17179, opened 2024-12-03
[math] Fix debug assertion failure

bellenot:
Fix the following error (visible in TMVA):
```
libTMVA: Debug Assertion Failed!
         Expression: cannot seek vector iterator before begin
```

PR #19143, opened 2025-06-23
[Core] Deprecate `TMath::BinarySearch` overload for iterators

guitargeek:
PR https://github.com/root-project/root/pull/17179 got stuck because there were three different opinions on how to resolve undefined behavior in case the search value is smaller than the smallest element in the collection. In particular, the problem was that in TMVA, the undefined behavior was used, because in practice for a `std::vector()`, the expression `std::distance(v.begin(), v.begin() - 1)` evaluates to -1, even though it's actually UB to take the iterator before the beginning of a vector. So there is no backwards compatible way to resolve the undefined behavior.

The only way out that I see is to deprecate the `TMath::BinarySearch` overload for iterators, and tell people to use the standard library instead. This is implemented in this commit.

Alternative to #17179.

One can use this code snipped to validate that the code changes in TMVA are correct:
```c++
std::vector<double> v{0., 1., 2};

auto check = [&](double x) {
   double r1 = std::distance(v.begin(), TMath::BinarySearch(v.begin(), v.end(), x));
   double r2 = std::distance(std::lower_bound(v.rbegin(), v.rend(), x, std::greater{}), v.rend()) - 1;
   std::cout << "x : " << x << std::endl;
   std::cout << r1 << std::endl;
   std::cout << r2 << std::endl;
   std::cout << std::endl;
};

check(-0.5);
for (double x : v) {
    check(x);
    check(x + 0.5);
}
```

### DIFF
--- a/core/base/inc/TMathBase.h
+++ b/core/base/inc/TMathBase.h
@@ -97,7 +97,8 @@ namespace TMath {
    // Binary search
    template <typename T> Long64_t BinarySearch(Long64_t n, const T  *array, T value);
    template <typename T> Long64_t BinarySearch(Long64_t n, const T **array, T value);
-   template <typename Iterator, typename Element> Iterator BinarySearch(Iterator first, Iterator last, Element value);
+   template <typename Iterator, typename Element> Iterator BinarySearch(Iterator first, Iterator last, Element value)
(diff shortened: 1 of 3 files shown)


pcanal (approved):
I think this fine.  The downside is the deprecation of a routine we have no idea how often it is used.  On the other hand, the standard replacement is pretty straightforward.

Issue #19035, opened 2025-06-13
The `rootcp --replace` logic behaves counterintuitively


guitargeek:
When using `rootcp --replace`, objects are only replaced if they are *not* equal to the existing object by value, which doesn't seem to makes sense to me.

### Reproducer

Write some histograms:
```c++
// write_histo_files.C
void write_histo_file(double w1, double w2, std::string const &name)
{
   std::string filename = name + ".root";
   std::unique_ptr<TFile> file{TFile::Open(filename.c_str(), "RECREATE")};

   TH1 *histo = new TH1D{"histo", "histo", 20, -4, 4};
   histo->Fill(1.0, w1);
   histo->Fill(1.0, w2);

   histo->Write();
}

void write_histo_files()
{
    // Same histograms
    write_histo_file(1, 1., "source_same");
    write_histo_file(1, 1, "target_same");

    // Different histograms
    write_histo_file(1.0, 1.0, "source_different");
    write_histo_file(0.5, 0.5, "target_different");
}
```

Now run this script and use rootcp to merge the source with target files, and check the target files:
```bash
root -b -q write_histo_files.C

rootcp --replace source_same.root:histo target_same.root:histo
rootcp --replace source_different.root:histo target_different.root:histo

echo "Same histos:"
rootls -t target_same.root
echo "Different histos:"
rootls -t target_different.root
```

The output will be:

```txt
Same histos:
TH1D  Jun 13 19:55 2025 histo;2 "histo" [current cycle]
TH1D  Jun 13 19:55 2025 histo;1 "histo" [backup cycle]
Different histos:
TH1D  Jun 13 19:55 2025 histo;1 "histo"
```

How does it make sense to write only a backup if the histograms were the same anyway, and otherwise not? This logic needs to be revisited I think:
https://github.com/root-project/root/blob/master/main/python/cmdLineUtils.py#L700

I think with the `--replace` option, the object should always be replaced without no backup cycle, and that's actually how it was at some point. It is not clear to me why this behavior was changed in f25b6e561a2b7 ten years ago.

@dpiparo, do you maybe have a hint since you merged that commit?

This logical flaw becomes apparent now that histogram classes implement an equality operator via UHI. Before, the equality operator was falling back to pointer comparison, which always failed since the objects in the source and target files are different. Hence, the histogram object was always overwritten with no backup cycle.

PR #19145, opened 2025-06-23
[Python] Remove TObject `__eq__` pythonization

guitargeek:
The TObject `__eq__` pythonization calls `TObject::Equals()`, which by default does a pointer comparison. That can be very confusing for Python users.

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tobject.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tobject.py
@@ -8,7 +8,6 @@
 # For the list of contributors see $ROOTSYS/README/CREDITS.                    #
 ################################################################################
 
-from ROOT.libROOTPythonizations import AddTObjectEqNePyz
 import cppyy
 
 # Searching
@@ -56,7 +55,6 @@ def pythonize_tobject():
     klass.__contains__ = _contains
 
     # Inject comparison operators
-    AddTObjectEqNePyz(klass)
     klass.__lt__ = _lt
     klass.__le__ = _le
     klass.__gt__ = _gt


vepadulano:
Just a fly-by thought, this pythonization has been around for so many years that, while I still strongly believe its current implementation is buggy, I am afraid this will introduce a backwards-incompatible behaviour far, far worse than https://github.com/root-project/root/pull/19061. One possible way to mitigate this is to have a check that if there is no `operator ==`  implementation (or equivalent) for a class, then we can raise an informative exception. Actually, we should probably just raise a warning and declare a deprecation phase for this. Also, note that while the discussion at #19061 is regarding `TH1` classes and derived only, this PR is going to have effects for every `TObject` derived class.

guitargeek:
That's exactly my intention! I wanted to see what the CI says, and if there are no problems deprecate this Pythonization.

> I am afraid this will introduce a backwards-incompatible behaviour far, far worse than https://github.com/root-project/root/pull/19061.

In my opinion, the silent behavior changes that are far worse. Removing a method like suggested by this PR at least gives you an error (or warning after this PR is updated).

You also have to see the picture with ROOT 6.34 before the UHI introduction in 6.36. What you guys did with the new `__eq__` was to introduce a change of behavior without documenting it, and as you can see from our tests and #19035 this even confused our own code. So it definitely has the potential to create bugs in user code. Doesn't matter how wrong or "buggy" you think the old behavior of comparing by pointer was: it has been there for many years as you said and people might rely on it.

> Also, note that while the discussion at https://github.com/root-project/root/pull/19061 is regarding TH1 classes and derived only, this PR is going to have effects for every TObject derived class.

Yes, and this is a bonus :slightly_smiling_face:

vepadulano:
> Yes, and this is a bonus 🙂

Only if we make sure this won't disrupt users' code. As things stand, this PR cannot be merged. My suggestion is to start a deprecation phase with a warning, then make it an exception or not implemented in a later ROOT release like 6.40 or 6.42

guitargeek:
> My suggestion is to start a deprecation phase with a warning, then make it an exception or not implemented in a later ROOT release like 6.40 or 6.42

Good idea! I have implemented what you suggested, and also added an entry in the release notes.

vepadulano (approved):
Great thank you! Let's see if it doesn't break any existing test in the CI, then this can be merged 👍

PR #19146, opened 2025-06-23
[CPyCppyy] Fix logic error in CPPMethod::Call with keyword arguments

### LINKED ISSUE #16406: cppyy passing arguments by name to C++ binding
### Feature description

Is it possible to support passing arguments by name to a C++ method in cppyy, including passing out of order and not passing all arguments if some are defaulted? 

Here is an example:

```
import ROOT

ROOT.gInterpreter.ProcessLine("""
class MyClass {
public:
    class MyObj {
    public:
        MyObj(const char* in) : s(in) { }
        TString s;
    };
    void MyMethod(const MyObj& x="hello", bool opt=false, bool opt2=true) {
      std::cout << x.s << " " << opt << " " << opt2 << std::endl;
    }

};
""")

ROOT.MyClass().MyMethod(opt2=False,x="hi") # I'd like this to output "hi 0 0"
```
At the moment this produces:

```
File cppyy_default_compiler:1, in <module>

NameError: name 'false' is not defined
```

Thanks!

### Alternatives considered

_No response_

### Additional context

_No response_

dpiparo:
This PR fixes #16406

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/CPPMethod.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/CPPMethod.cxx
@@ -1013,7 +1013,7 @@ PyObject* CPyCppyy::CPPMethod::Call(CPPInstance*& self,
         ctxt->fPyContext = (PyObject*)cargs.fSelf;    // no Py_INCREF as no ownership
 
 // translate the arguments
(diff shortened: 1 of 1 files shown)


guitargeek:
It took some time to dig into result for the test failures, but in hindsight the problem is obvious. I have added a commit with a fix, it's a one-liner. Hopefully it works now!

dpiparo:
nice.

PR #19149, opened 2025-06-24
[roottest] use cmake for io tests [part4]

linev:
Use cmake in several `roottest/root/io` tests:
* `io/newdelete`
* `io/pointers`
* `io/prefetching`
* `io/recover`
* `io/treeForeign`

Resolve different compiler warnings for CI platform with `-Ddev=On`


roottest/root/io/prefetching/CMakeLists.txt
```diff
@@ -1,9 +1,13 @@
-#-------------------------------------------------------------------------------
-#
-# Placeholder file to translate the tests to the new CTest system. Meanwhile we
-# define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
-#
-#-------------------------------------------------------------------------------
 if(xrootd)
-  ROOTTEST_ADD_OLDTEST(LABELS longtest)
+  ROOTTEST_GENERATE_DICTIONARY(libRoottestIoPrefetching
+                               atlasFlushed/atlasFlushedProjectHeaders.h
```
pcanal, line 3: The previous incarnation was split the dictionary in 2 part (allowing parallel build and reduce memory usage).  Did you compare both (i.e. just running the makefile and just running the new version)?

linev: Dictionary build takes about 50 seconds.
But running of the test after - 200 seconds. 
So it will be not that big difference if dictionary can be done in parallel

pcanal (approved):
Thanks.

PR #19151, opened 2025-06-24
[do not merge] RFile Prototype 

silverweed:
# This Pull request:
Draft PR for the new RFile prototype, using a non-RDirectory-based approach (an alternative PR using a RDirectory-based approach will follow soon).

Introduces `ROOT::Experimental::RFile` which presents a new, minimal interface for reading and writing data to a ROOT file. 

NOTE: the commit history is a bit messy at the moment, but it's not meant to be "review-ready"; it's more to showcase the current progress of the prototype.

### DIFF
--- a/io/v7/rfile/test/rfile.cxx
+++ b/io/v7/rfile/test/rfile.cxx
@@ -566,3 +566,28 @@ TEST(RFile, LongKeyName)
    auto it = keys.begin();
    EXPECT_EQ(it->fName, kKeyLong);
 }
+
+TEST(RFile, NormalizedPaths)
+{
+   FileRaii fileGuard("test_rfile_normalizedpaths.root");
+
+   auto file = RFile::Recreate(fileGuard.GetPath());
+   std::string obj = "obj";
+   file->Put("/s", obj);
+   // "a" and "/a" are equivalent paths, so we cannot overwrite it using Put()...
+   EXPECT_THROW(file->Put("s", obj), ROOT::RException);
+   // ...and this is true no matter how many leading slashes we have.
+   EXPECT_THROW(file->Put("////s", obj), ROOT::RException);
+   EXPECT_EQ(*file->Get<std::string>("s"), obj);
+   EXPECT_EQ(*file->Get<std::string>("//s"), obj);
+
+   TH1D h("h", "h", 10, -10, 10);
+   // Cannot write directory 's': already taken by `obj`.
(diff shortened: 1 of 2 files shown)


dpiparo:
Just to put this on the table https://its.cern.ch/jira/browse/ROOT-7400
Maybe to further increase the value of the new RFile.

io/v7/rfile/test/rfile.cxx
```diff
@@ -0,0 +1,568 @@
...
+      EXPECT_EQ(hist, nullptr);
+      // ...but any cycle before the latest should still be there!
+      hist = file->Get<TH1D>("hist;1");
+      EXPECT_NE(hist, nullptr);
+   }
+}
+
+TEST(RFile, WrongExtension)
+{
+   FileRaii fileGuard("test_rfile_wrong.xml");
```
pcanal, line 204: Until now we have not enforced the convention for the extension.

io/v7/rfile/test/rfile.cxx
```diff
@@ -0,0 +1,568 @@
...
+   try {
+      file->Put("a/b/c", s);
+      FAIL() << "creating a directory over a leaf path should fail.";
+   } catch (const ROOT::RException &ex) {
+      EXPECT_THAT(ex.what(), testing::HasSubstr("'a/b'"));
+      EXPECT_THAT(ex.what(), testing::HasSubstr("name already taken"));
+   }
+}
+
+TEST(RFile, IterateKeysRecursive)
```
pcanal, line 332: What does 'recursing' over the directories look like for 'complex' operation?  For example creating a graph representation of the file's content (text or in TBrowser nodes); Another example could be create Canvas and subCanvas display the histo of each directory (in an organization mimicking the directory organization).

io/v7/rfile/test/rfile.cxx
silverweed: But if I'm not mistaken TFile will try to open a TXMLFile if it detects this extension (at least with TFile::Open) and I wanted to avoid that

pcanal: That is a slightly different interface per say.  `TFile::Open` (as oppose to the TFile constructor) is supposed to find the right plugin and it indeed bases that search on pattern matching (`protocol` ... filename `.extension`) but if some code is creating the pluging class directly then there is less checks (in part because they are redundant, in part because `new TXMLFile(filename)` is an explicit request to open that file as xml)

PR #19152, opened 2025-06-24
[roottest] use fixtures in io/filemerger

linev:
Avoid DEPENDS while single tests cannot be run by such method 
Split many tests on two parts - first generate files with different compressions using hadd and then check sizes.
By such technique much clear dependency can be configured and many files can be generated/checked in parallel.
Use more clear fixtures names to be able correctly follow dependencies 
Adjust testSimpleFile.C macro


roottest/root/io/filemerger/CMakeLists.txt
```diff
@@ -1,372 +1,407 @@
 ROOTTEST_ADD_TEST(execKeyOrder
-	          MACRO execKeyOrder.C
-		  OUTREF execKeyOrder.ref)
-
+	               MACRO execKeyOrder.C
```
pcanal: Both the original and the new version contains tab characters, can you update to use space?

pcanal (approved):
Thanks.

roottest/root/io/filemerger/CMakeLists.txt
linev: Done.

linev:
Failure `tutorial-machine_learning-RBatchGenerator_TensorFlow-py` not related to PR

PR #19155, opened 2025-06-24
Make CppInterOp work with new gtest target names.

hageboeck:
Starting from CMake v3.23, one can use GTest::gtest and similar in CMake. Since the root-project wants to move to these targets, Interop should work with both the new or the old target names in order not to start building its own gtest.

This is a replay of the same commit added to interop [here](https://github.com/compiler-research/CppInterOp/pull/643).


hageboeck:
Will do this in conjunction with other interOp updates.

PR #19156, opened 2025-06-24
[roottest] use cmake for io/references

linev:
Do not use label `longtest` while each single test is relatively fast
Split refs file which were produced by concatenation of output of several tests
Put under `if(broken_tests)` generation of dictionary which fails, was also so in Makefile
Directly compile macro without using `create.C` which just deleted

### DIFF
--- a/roottest/root/io/references/A.C
+++ b/roottest/root/io/references/A.C
@@ -10,8 +10,4 @@ public:
   ClassDef(A, 1)
 };
 
-#ifdef __ROOTCLING__
-#pragma link C++ class A+;
-#endif
-
 #endif
--- a/roottest/root/io/references/CMakeLists.txt
+++ b/roottest/root/io/references/CMakeLists.txt
@@ -9,6 +9,7 @@ ROOTTEST_COMPILE_MACRO(A.C
                        FIXTURES_SETUP root-io-references-A-fixture)
 
 ROOTTEST_COMPILE_MACRO(lotsRef.C
+                       BUILDLIB A_C
                        FIXTURES_REQUIRED root-io-references-A-fixture
                        FIXTURES_SETUP root-io-references-lotsRef-fixture)
 


roottest/root/io/references/RefArrayCompress.hh
```diff
@@ -1,31 +1,33 @@
+#ifndef RefArrazCompress_hh
+#define RefArrazCompress_hh
```
pcanal:
```suggestion
#ifndef RefArrayCompress_hh
#define RefArrayCompress_hh
```
?

roottest/root/io/references/CMakeLists.txt
```diff
@@ -1,7 +1,70 @@
...
+ROOTTEST_ADD_TEST(assertManyRefs
+                  MACRO assertManyRefs.C
+                  OUTREF assertManyRefs.ref)
+
+ROOTTEST_COMPILE_MACRO(A.C
+                       FIXTURES_SETUP root-io-references-A-fixture)
+
+ROOTTEST_COMPILE_MACRO(lotsRef.C
+                       FIXTURES_REQUIRED root-io-references-A-fixture
+                       FIXTURES_SETUP root-io-references-lotsRef-fixture)
```
pcanal: One difference between the previous implementation and this implementation is that `lotsRef_C.so` was linked against `A_C.so` (because it was loaded when `lotsRef.C` was compiled by ACLiC.

roottest/root/io/references/A.C
```diff
@@ -1,17 +1,17 @@
...
+class A : public TObject {
+public:
+  A(A* other = nullptr) { _o = other; }
+  TRef _o;
+  ClassDef(A, 1)
 };
 
+#ifdef __ROOTCLING__
+#pragma link C++ class A+;
+#endif
```
pcanal: The need for this is directly linked to the change indicated at: https://github.com/root-project/root/pull/19156/files#r2167236044.

In particular due to the fact that `lotsRef.C` includes `A.C`, this line means that the dictionary for 'A' is generated for/in both `A_C.so` and `lotsRefs_C.so` ...

roottest/root/io/references/CMakeLists.txt
linev: But how it can be done now in `ROOTTEST_COMPILE_MACRO`.
Should I also add here `ROOTEXE_OPTS` arguments?

roottest/root/io/references/A.C
linev: Yes, you right!

I add `ROOTEXE_OPTS` to `ROOTTEST_COMPILE_MACRO` and remove such extra entry

roottest/root/io/references/CMakeLists.txt
linev: See my new commits

roottest/root/io/references/RefArrayCompress.hh
pcanal: There is (I think) a typo where a 'z' is used at the end of 'array' :)

linev: Fixed by https://github.com/root-project/root/pull/19156/commits/04b1c7462be843dde86538d24b5a72ca08314dee

German keyboard swap y/z keys

roottest/root/io/references/CMakeLists.txt
```diff
@@ -1,7 +1,71 @@
...
+
+ROOTTEST_ADD_TEST(assertManyRefs
+                  MACRO assertManyRefs.C
+                  OUTREF assertManyRefs.ref)
+
+ROOTTEST_COMPILE_MACRO(A.C
+                       FIXTURES_SETUP root-io-references-A-fixture)
+
+ROOTTEST_COMPILE_MACRO(lotsRef.C
+                       ROOTEXE_OPTS -e ".L A.C+"
```
pcanal: This works and is fine.
The following should also work:
```suggestion
                       BUILDLIB A_C
```
(although it is not actually used yet ... so who knows it works as 'advertised' :) )

pcanal:
The windows failures in roottest-root-io-bigevent-IoBigEventGenerator-build seems to indicate a potential 'race condition' but it is not clear whether it is related to this PR or not.

roottest/root/io/references/CMakeLists.txt
linev: Yes, seems to be works

PR #19163, opened 2025-06-24
[CI] Enable PyMVA

guitargeek:
Now that TensorFlow 2.19 made it to the CI images, we should re-enable PyMVA for a check and see if things work with the commits from #15790.


guitargeek:
Doesn't pass the tests in the CI, so now we know that #15790 still needs some work

PR #19165, opened 2025-06-24
[CMake] Also print version of found Python module

guitargeek:
This does the same thing as CMakes find package now, printing the version of the Python module in the same style if available.

This will make it easier to debug compatibility problems from the build logs.


bellenot (approved):
Cool! Thanks Jonas!

Issue #19166, opened 2025-06-25
RooSimultaneous cannot accept ConditionalObservables properly.


BaseHardware:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

`RooSimultaneous` is a PDF class that performs the simultaneous fitting of multiple pairs of a PDF and a dataset.

`RooFit::ConditionalObservables(RooAbsArgs...)` is a command argument to the `fitTo()` method of the `RooAbsPdf` classes, which specifies variables that should be excluded from the normalization set of a PDF when fitting.

If I use the command argument for fitting using the class, the class does not deal with the conditional variable well, and the conditional variables become floating parameters. This should be fixed.

I already implemented the solution and will make a pull request for the patch. This issue is merely for mentioning in the pull request.

### Reproducer

I've included the code below for illustrative purposes only. It may be incorrect.
```
RooWorkspace* ws = new RooWorkspace();
ws.factory("cat[A=1, B=2]");
ws.factory("x[0, -10, 10]");
ws.factory("sigmaA[1, 0.01, 10]");
ws.factory("sigmaB[5, 0.01, 10]");

RooRealVar* condVar = ws.factory("condVar[5, 2, 10]");
ws.factory("comMean[3, -5, 5]");

ws.factory("expr::offsetMean(\"@0 / @1\", comMean, condVar)");

ws.factory("RooGaussian::pdfA(x, paramA, offsetMean)");
ws.factory("RooGaussian::pdfB(x, paramB, offsetMean)");

RooAbsPdf* simPdf = ws.factory("SIMUL::[cat, A=pdfA, B=pdfB]");

RooDataSet* combined_ds = // I will skip creating the categorized RooDataSet. This RDS instance should contain axis named "x" and "condVar" and be categorized with RooAbsCategory named "cat".

simPdf.fitTo(combined_ds, RooFit::ConditionalVariables(*condVar)); // condVar becomes floating parameters. Not even a "variable"
simPdf.fitTo(combined_ds) // Now condVar becomes "variable" in the fitting, not a conditional one. RooFit attempts normalization over the range of condVar.
```

### ROOT version

6.36.00

### Installation method

build from source

### Operating system

MacOS, Linux

BaseHardware:
PR at https://github.com/root-project/root/pull/19167 addresses this issue.

ferdymercury:
Maybe related:
https://github.com/root-project/root/issues/18718
[...]

PR #19167, opened 2025-06-25
Fixing issues in RooSimultaneous with ConditionalObservables

BaseHardware:
# This Pull request:

## Changes or fixes:

`RooSimultaneous` is a PDF class that performs the simultaneous fitting of multiple pairs of a PDF and a dataset.

`RooFit::ConditionalObservables(RooAbsArgs...)` is a command argument to the `fitTo()` method of the `RooAbsPdf` classes, which specifies variables that should be excluded from the normalization set of a PDF when fitting.

If I use the command argument for fitting using the class, the class does not deal with the conditional variable well, and the conditional variables become floating parameters. This should be fixed.

## Checklist:

- [x] tested changes locally

This PR addresses the bug. Please see the details of the issue in the link: https://github.com/root-project/root/issues/19166

### DIFF
--- a/roofit/roofitcore/src/FitHelpers.cxx
+++ b/roofit/roofitcore/src/FitHelpers.cxx
@@ -745,7 +745,17 @@ std::unique_ptr<RooAbsReal> createNLL(RooAbsPdf &pdf, RooAbsData &data, const Ro
 
       RooArgSet normSet;
       pdf.getObservables(data.get(), normSet);
-      normSet.remove(projDeps, true, true);
+
+      if (dynamic_cast<RooSimultaneous const*>(&pdf)) {
+         for (auto i : projDeps) {
+            auto res = normSet.find(i->GetName());
+            if (res != nullptr) {
+               res->setAttribute("__conditional__");
+            }
+         }
+      } else {
+         normSet.remove(projDeps);
+      }
 
       pdf.setAttribute("SplitRange", splitRange);
       pdf.setStringAttribute("RangeName", rangeName);
--- a/roofit/roofitcore/test/testRooSimultaneous.cxx
+++ b/roofit/roofitcore/test/testRooSimultaneous.cxx
@@ -456,3 +456,54 @@ TEST(RooSimultaneous, DuplicateExtendedPdfs)
    EXPECT_FLOAT_EQ(simPdfVal, 0.05);
    EXPECT_DOUBLE_EQ(simPdfVal, simPdfRef.getVal(normSet));
 }
+
+/// GitHub issue #19166.
+/// RooSimultaneous::fitTo() should work well with ConditionalOvservables.
+/// We test this by performing the similar test to a test case in
+/// RooSimultaneousSingleChannelCrossCheck.
+TEST_P(TestStatisticTest, RooSimultaneousSingleChannelCrossCheckWithCondVar)
+{
+   using namespace RooFit;
+
+   // silence log output
+   RooHelpers::LocalChangeMsgLevel changeMsgLvl(RooFit::WARNING);
+
+   RooWorkspace ws;
+   ws.factory("Uniform::uniform(width[1, 3])");
+   ws.factory("expr::sigma(\"@0*@1\", width, width_scale[0.8, 0.5, 1.5])");
+   ws.factory("Gaussian::model(x[-20, 20], mean[3, -20., 20.], sigma)");
+
+   RooRealVar &width = *ws.var("width");
+   RooAbsPdf &widthModel = *ws.pdf("uniform");
+   std::unique_ptr<RooDataSet> protoData{widthModel.generate(width, 100)};
+
+   RooRealVar &x = *ws.var("x");
+   RooAbsPdf &model = *ws.pdf("model");
+
+   std::unique_ptr<RooDataSet> data{model.generate(x, *protoData)};
+
+   RooCategory cat("cat", "cat");
+   cat.defineType("physics");
+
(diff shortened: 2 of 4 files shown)


dpiparo:
Let's start the CI waiting for @guitargeek

dpiparo:
and thanks for choosing ROOT for your first PR!

guitargeek:
Hi @BaseHardware, thank you very much for this nice PR!

Could you please write a unit test for this and put it inside [roofit/roofitcore/test/testRooSimultaneous.cxx](https://github.com/root-project/root/blob/master/roofit/roofitcore/test/testRooSimultaneous.cxx)? What we need is a closure check that a likelihood with conditional observables gives the same fit result with the same free parameters, also when it's wrapped in a RooSimultaneousPdf.

Thanks a lot in advance and congrats to your first ROOT contribution!

BaseHardware:
@guitargeek 

Okay, as this is my first time creating a unit test, let me examine the source code and learn how to do it properly. Thanks for your comment. By the way, I have a question regarding this commit:

From the source code of `RooSimultaneous::compileForNormSet()`, it is evident that the arguments for invoking `fixAddCoefNormalization()` should not contain the conditional variables.
However, I am not sure this is also true for `pdfContext.compile()` (https://github.com/root-project/root/blob/30aff7079c/roofit/roofitcore/src/RooSimultaneous.cxx#L1227), as the current modification also removes the conditional variables from the invocation.

What is the behavior of the function? Should the conditional variables also be excluded from that function?

BaseHardware:
@guitargeek 

Also, I have a question also for your request of the unit test.
I understood the request from your comment that there is a wrapper class named `RooSimultaneousPdf` in RooFit, and the behavior in fitting of the class should be the same as that of `RooSimultaneous`, but I couldn't find such a class in the source code tree of ROOT. (I searched that by `grep RooSimultaneousPdf * -ri` in my linux environment.)

Could you explain exactly what test should be added to the unit test file?

ferdymercury:
Maybe he meant to compare that:
```
RooWhateverLikelihood w;
// set free params
```
behaves the same as
```
RooSimultaneous rs;
rs->addPdf(w);
// set same params
// Compare both results
```

ferdymercury:
Or rather check structure in: https://github.com/root-project/root/blob/master/roofit/roofitcore/test/testRooSimultaneous.cxx#L190

BaseHardware:
@ferdymercury 
Ah, you mean, he meant a test by comparing the fit result with the conditional variable from a standalone PDF (i.e., RooGaussian) and RooSimultaneous that wraps the same PDF with a single category.

The test clause you've mentioned is a case without a conditional variable, and what he wants me to create is a case with a conditional variable, right? I will make that. Thanks!

guitargeek:
You got it, thanks!

BaseHardware:
Done.

dpiparo:
and starting the ci!

BaseHardware:
Sorry, the style formatting failed in the previous commit. I updated the latest commit with the formatted version.

roofit/roofitcore/src/FitHelpers.cxx
```diff
@@ -745,7 +745,17 @@ std::unique_ptr<RooAbsReal> createNLL(RooAbsPdf &pdf, RooAbsData &data, const Ro
 
       RooArgSet normSet;
       pdf.getObservables(data.get(), normSet);
-      normSet.remove(projDeps, true, true);
+
+      if (pdf.InheritsFrom("RooSimultaneous")) {
```
guitargeek:
```suggestion
      if (dynamic_cast<RooSimultaneous const*>(&pdf)) {
```
No need to use ROOTs reflection here, you can just `dynamic_cast`.

roofit/roofitcore/test/testRooSimultaneous.cxx
```diff
@@ -456,3 +456,52 @@ TEST(RooSimultaneous, DuplicateExtendedPdfs)
...
+
+   RooDataSet combData("combData", "combData", {x, width}, Index(cat), Import({{"physics", data.get()}}));
+
+   params.assign(initialParams);
+   std::unique_ptr<RooFitResult> resDirect{
+      model.fitTo(*data, RooFit::ConditionalObservables(width), RooFit::Save(), RooFit::PrintLevel(-1))};
+
+   params.assign(initialParams);
+   std::unique_ptr<RooFitResult> resSimWrapped{
+      modelSim.fitTo(combData, RooFit::ConditionalObservables(width), RooFit::Save(), RooFit::PrintLevel(-1))};
```
guitargeek:
```suggestion
   using namespace RooFit;

   params.assign(initialParams);
   std::unique_ptr<RooFitResult> resDirect{
      model.fitTo(*data, ConditionalObservables(width), Save(), PrintLevel(-1), _evalBackend)};

   params.assign(initialParams);
   std::unique_ptr<RooFitResult> resSimWrapped{
      modelSim.fitTo(combData, ConditionalObservables(width), Save(), PrintLevel(-1), _evalBackend)};
```

This is a parametrized test with different choices of the NLL evaluation backend, so you need to use this `_evalBackend` argument like in the other tests. Like this, the test will also run with the CUDA and legacy backends.

guitargeek (changes requested):
Excellent first PR, thank you very much @BaseHardware! Only two small change requests :slightly_smiling_face:

guitargeek (approved):
I have implemented the suggestions to the PR can be merged quickly :slightly_smiling_face:

PR #8688, opened 2021-07-16
[ntuple] Add tutorial for multi-threaded filling

jblomer:
# This Pull request:
Add `RNTupleWriter::CreateEntry()` and uses it in a demonstration on multi-threaded writing.

PR #19171, opened 2025-06-25
[skip-ci][ntuple] Improve RNTupleWriter doc

### LINKED ISSUE #19168: Documentation of RNTupleWriter / RNTuple could be improved a bit
### RNTupleWriter
I tried to find out what `RNTupleWriter::Append` is supposed to do. Please see the current state that would need more information:
https://root.cern.ch/doc/master/classROOT_1_1RNTupleWriter.html#a7cbb027b0733b166da1b9c818fd90e2e

A similar extension could be done for `Recreate`.

Finally, the overview of the class at the top of the doxygen page starts by talking about RNTuple, but not about the writer. One could start maybe by saying something like:
> RNTupleWriter is a user interface for writing RNTuples to disc. Users can instantiate writers using the functions Append and Recreate (ensure that doxygen links) while providing an RNTupleModel that defines the fields/columns to be written.

And then go on to explain how to assign values to fields and trigger the writes. This could be finished off with a minimal code example and/or a link to a tutorial.

### RNTuple
Looking at the top of https://root.cern.ch/doc/master/classROOT_1_1RNTuple.html, I believe this is not what users want to see if they look for "RNTuple". Although what is described there is formally correct, they probably wanted to know how to read or write an RNT. So maybe one could extend the documentation to:

> `ROOT::RNTuple` is the representation of an [RNTuple](https://root.cern.ch/doc/master/classROOT_1_1RNTuple.html) data set in a [ROOT](https://root.cern.ch/doc/master/namespaceROOT.html) file.
> - To read an RNTuple, see RNTupleReader or use RDataFrame
> - To write an RNTuple, see RNTupleWriter

and from here continue the documentation of `RNTuple`.

### ROOT version

6.36, master

### Installation method

--

### Operating system

--

### Additional context

_No response_

silverweed:
# This Pull request:
Improves the documentation of RNTupleWriter, in particular Recreate and Append.

This PR fixes #19168

### DIFF
--- a/tree/ntuple/inc/ROOT/RNTupleWriter.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleWriter.hxx
@@ -49,10 +49,53 @@ CreateRNTupleWriter(std::unique_ptr<ROOT::RNTupleModel> model, std::unique_ptr<I
 \ingroup NTuple
 \brief An RNTuple that gets filled with entries (data) and writes them to storage
 
-An output ntuple can be filled with entries. The caller has to make sure that the data that gets filled into an ntuple
-is not modified for the time of the Fill() call. The fill call serializes the C++ object into the column format and
-writes data into the corresponding column page buffers.  Writing of the buffers to storage is deferred and can be
-triggered by FlushCluster() or by destructing the writer.  On I/O errors, an exception is thrown.
+RNTupleWriter is an interface for writing RNTuples to storage. It can be instantiated using the static functions
+Append() and Recreate(), providing an RNTupleModel that defines the schema of the data to be written.
+
+An RNTuple can be thought of as a table, whose columns are defined by its schema (i.e. by its associated RNTupleModel,
+whose Fields map to 0 or more columns). 
+Writing into an RNTuple happens by filling *entries* into the RNTupleWriter, which make up the rows of the table.
+The simplest way to do so is by:
+
+- retrieving a (shared) pointer to each Field's value;
+- writing a value into each pointer;
+- calling `writer->Fill()` to commit the entry with all the current pointer values.
+
+~~~ {.cpp}
+#include <ROOT/RNTupleWriter.hxx>
+
+/// 1. Create the model.
+auto model = ROOT::RNTupleModel::Create();
+// Define the schema by adding Fields to the model.
+// MakeField returns a shared_ptr to the value to be written (in this case, a shared_ptr<int>)
+auto pFoo = model->MakeField<int>("foo");
+
+/// 2. Create writer from the model.
+auto writer = ROOT::RNTupleReader::Recreate(std::move(model), "myNTuple", "some/file.root");
+
+/// 3. Write into it.
+for (int i = 0; i < 10; ++i) {
+   // Assign the value you want to each RNTuple Field (in this case there is only one Field "foo").
+   *pFoo = i;
+
+   // Fill() writes the entire entry to the RNTuple.
+   // After calling Fill() you can safely write another value into `pFoo` knowing that the previous one was
+   // already saved.
+   writer->Fill();
+}
+
+// On destruction, the writer will flush the written data to disk.
+~~~
+
+The caller has to make sure that the data that gets filled into an RNTuple is not modified for the time of the
+Fill() call. The Fill call serializes the C++ object into the column format and
+writes data into the corresponding column page buffers.
+
+The actual writing of the buffers to storage is deferred and can be triggered by FlushCluster() or by
+destructing the writer.
+
+On I/O errors, a ROOT::RException is thrown.
+
 */
 // clang-format on
 class RNTupleWriter {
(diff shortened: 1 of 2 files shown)


tree/ntuple/inc/ROOT/RNTupleWriter.hxx
```diff
@@ -83,14 +118,33 @@ private:
                                                 const ROOT::RNTupleWriteOptions &options);
 
 public:
+   /// Creates an RNTupleWriter backed by `storage`, overwriting it if one with the same URI exists.
+   /// The format of the backing storage is determined by `storage`: in the simplest case it will be a local file, but
+   /// a different backend may be selected via the URI prefix. The "daos://" prefix will use the experimental DAOS
+   /// backend.
```
hahnjo: I'm not sure if we want to mention DAOS here since it might only be available if enabled.

tree/ntuple/inc/ROOT/RNTupleWriter.hxx
```diff
@@ -83,14 +118,33 @@ private:
...
+   ///
    /// Throws an exception if the model is null.
    static std::unique_ptr<RNTupleWriter>
    Recreate(std::unique_ptr<ROOT::RNTupleModel> model, std::string_view ntupleName, std::string_view storage,
             const ROOT::RNTupleWriteOptions &options = ROOT::RNTupleWriteOptions());
+
+   /// Same as Recreate(std::unique_ptr<ROOT::RNTupleModel>, std::string_view, std::string_view, const
+   /// ROOT::RNTupleWriteOptions &) but the RNTupleModel will be created with the dynamically-typed fields specified in
+   /// `fields`. This overload may be used when the RNTuple schema is created at runtime with types unknown at compile
+   /// time.
```
hahnjo: The recommended way should be `RFieldBase::Create` and `RNTupleModel::AddField`. This `Recreate` overload is simply a convenience function...

silverweed: I actually didn't even know we had this overload. What is the use case we implemented it for?

hahnjo: I think it was implemented during a ROOT hackathon, mostly with usage from Python in mind (that was before we were able to fully use `RNTupleModel` from Python)

silverweed: Do we still have a reason to keep it? Should we deprecate it?

tree/ntuple/inc/ROOT/RNTupleWriter.hxx
```diff
@@ -83,14 +118,33 @@ private:
...
    static std::unique_ptr<RNTupleWriter>
    Recreate(std::initializer_list<std::pair<std::string_view, std::string_view>> fields, std::string_view ntupleName,
             std::string_view storage, const ROOT::RNTupleWriteOptions &options = ROOT::RNTupleWriteOptions());
-   /// Throws an exception if the model is null.
+
+   /// Creates an RNTupleWriter that writes into an existing TFile or TDirectory, without overwriting its content.
+   /// `fileOrDirectory` may be an empty TFile.
+   /// \see Recreate() for more information.
```
hageboeck: Doxygen tries to autolink everything you mention after `\see`, so it would try to create links for "for" "more" and "information" as well.
```suggestion
   /// \see Recreate()
```

tree/ntuple/inc/ROOT/RNTupleWriter.hxx
```diff
@@ -83,14 +118,33 @@ private:
...
+   /// The format of the backing storage is determined by `storage`: in the simplest case it will be a local file, but
+   /// a different backend may be selected via the URI prefix. The "daos://" prefix will use the experimental DAOS
+   /// backend.
+   ///
+   /// The RNTupleWriter will create an RNTuple with the schema determined by `model` (which must not be null) and
+   /// with name `ntupleName`. This same name can later be used to read back the RNTuple via RNTupleReader.
+   ///
+   /// `options` may be passed to this function to customize the behavior of the RNTupleWriter (\see
+   /// RNTupleWriteOptions).
+   ///
```
hageboeck, line 140: Would it make sense to structure the docstring?
```c++
\param model ...
\param ntupleName ...
\param storage ...
\param options ...
```
A random example for how it shows up in doxygen:
https://root.cern/doc/master/classROOT_1_1RDF_1_1RInterface.html#a9ce937f8ab033be390d7209df594669d

silverweed:
> What about the part on RNTuple, would you be willing to consider this, too?

Yes, I will append another commit to address that as well

hageboeck (approved):
LGTM. Thanks!

tree/ntuple/inc/ROOT/RNTupleWriter.hxx
maszyman: FWIW, it was originally suggested [here](https://github.com/root-project/root/pull/8688#issuecomment-883396155).

silverweed: Thanks. Then I suppose we don't want to deprecate it, but at the same time it seems like the intended use is not the one I expected and it's just to have an inline model...I'll amend the doc comment.

Issue #14579, opened 2024-02-04
[ROOT-10190] Excluded fastcloningeventtree test in roottest.git should be re-enabled back


vepadulano:
### Explain what you would like to see improved and how.

From https://its.cern.ch/jira/browse/ROOT-10190

Files for fastcloningeventtree were moved from filemerger directory during its migration to CMake. During migration to CMake, it was found that test was not running for some time and as a result it is not compiling. 

Fastcloningeventtree is temporary disabled for this reason.

### ROOT version

Any

### Installation method

Any

### Operating system

Any

ferdymercury:
Fixed by https://github.com/root-project/root/pull/19173

PR #19173, opened 2025-06-25
[roottest] build `libEvent` and dependent tests with cmake

linev:
1. Extend `ROOTTEST_ADD_TEST` to support `cc` and `cpp` extensions for macros. Alternatively one can rename several old files
2. Add `STOREOUT` argument to `ROOTTEST_ADD_TEST` to store output of some executable running - like it always done for macro. Let use such output as reference for other tests
3. Convert `root/io/event` to cmake. 
4. Reuse library and executable in other tests (via fixtures):
   * `root/io/compression`
   * `root/tree/cache`
   * `root/tree/friends`
   * `root/treeformula/retobj`
   * `root/treeformula/sync`
   * `root/treeformula/casting`
5. Use drawing test with `libEvent`  in `root/treeformula/event`:
    * Replace `dt_runDrawTest.sh` logic by cmake loop. 
    * Testing exactly same files but all individual tests "visible" as cmake target
    * Enable such drawing test in the Windows
6. Enable `root/tree/fastcloningeventree` with drawing loop:
   * test was not running with Makefile
   * similar drawing loop as with `root/treeformula/event`
   * in-between `ctrans.C` with fastcloning create new files
   * there are failing tests with old splitting, one can disable such tests completely
8. Special handling of `root/treeformula/schemaEvolution`
   * Keep `Event.h` and `Event.cxx` while two versions of Event class is used
   * Create reference file using `STOREOUT`
   * Via `foreach` loop run tests with different condition and compare with reference file
9. Fully remove Event.h and drawing scripts from `root/meta` while it fully replicates `root/treeformula/event`.
10. Fully remove not used `root/io/newClassDef/showmember` subdirectory with duplication of many tests
11. While many Makefiles depends on libEvent target just because of sub-directories, remove this target from Rules.mk as no longer needed


roottest/root/io/event/CMakeLists.txt
```diff
@@ -1,12 +1,22 @@
...
+                             FIXTURES_SETUP root-io-event-libTestIoEvent-fixture)
+
+ROOTTEST_GENERATE_EXECUTABLE(TestIoEvent
+                             MainEvent.cxx
+                             LIBRARIES Core RIO libTestIoEventlibgen
+                             FIXTURES_REQUIRED root-io-event-libTestIoEvent-fixture
+                             FIXTURES_SETUP root-io-event-TestIoEvent-fixture)
+
+ROOTTEST_ADD_TEST(write
+                  COMMAND ./TestIoEvent 10 1 1 1
```
pcanal, line 16: As the makefile did, we should check the output.  The goal there was to detect unexpected Warning or Error messages (that did not lead to error exit code).

roottest/root/tree/fastcloningeventtree/dt_DrawTest.C
```diff
@@ -71,85 +69,95 @@ TDirectory* GenerateDrawHist(TTree *tree, int quietLevel = 0, int level = 3)
...
-   DrawSkippable(tree,"fYfirst","fEvtHdr.fEvtNum%10 == 2","hYfirst");
-   DrawSkippable(tree,"fZfirst","fEvtHdr.fEvtNum%10 == 2","hZfirst");
-   DrawSkippable(tree,"fXlast",  "fEvtHdr.fEvtNum%10 == 3","hXlast");
-   DrawSkippable(tree,"fYlast",  "fEvtHdr.fEvtNum%10 == 3","hYlast");
-   DrawSkippable(tree,"fZlast",  "fEvtHdr.fEvtNum%10 == 3","hZlast");
-   DrawSkippable(tree,"fCharge","fPx < 0","hCharge");
-   DrawSkippable(tree,"fNpoint","fPx < 0","hNpoint");
-   DrawSkippable(tree,"fValid",  "fPx < 0","hValid");
-   DrawSkippable(tree,"fPointValue","hPointValue", gBranchStyle!=0);
+   tree->Draw("fTracks.fPx>>hPx","fEvtHdr.fEvtNum%10 == 0","goff");
```
pcanal, line 92: The change made in this file seems to be a reversion of the local change/improvement made to this version.  Are they intentional?

roottest/root/tree/cache/Event.h
```diff
@@ -1,204 +0,0 @@
...
-class TDirectory;
-
-class Track : public TObject {
-
-private:
-   Float_t      fPx;           //X component of the momentum
-   Float_t      fPy;           //Y component of the momentum
-   Float_t      fPz;           //Z component of the momentum
-   Float_t      fRandom;       //A random track quantity
-   Float16_t    fMass2;        //[0,0,8] The mass square of this particle
```
pcanal, line 30: All 4 removed Event.h have significant differences (for example here `fMass2` has a different type).  How were those differences handled/preserved?

roottest/root/tree/friend/CMakeLists.txt
```diff
@@ -1,7 +1,117 @@
...
+                  OUTREF Unaligned.ref)
+
+ROOTTEST_ADD_TEST(MissingParentValue
+                  MACRO execMissingParentValue.C
+                  OUTREF execMissingParentValue.ref)
+
+ROOTTEST_ADD_TEST(NonExistingFriend
+                  MACRO execNonExistingFriend.C
+                  # produced error output includes full file path
+                  WILLFAIL
```
pcanal: The test is meant to succeed.  The Makefile had the filter `| sed -e 's?: file [A-Za-z/\].*[/\]?: file ?' ` which might explain the difference.

roottest/root/tree/friend/CMakeLists.txt
```diff
@@ -1,7 +1,117 @@
...
+ROOTTEST_ADD_TEST(testFriend
+                  MACRO testFriend.cc
+                  FIXTURES_REQUIRED root-tree-friend-file-fixture
+                                    root-tree-friend-mcfiles-fixture)
+ROOTTEST_ADD_TEST(testFriend2
+                  MACRO testFriend2.cc
+                  FIXTURES_REQUIRED root-tree-friend-file-fixture
+                                    root-tree-friend-mcfiles-fixture)
+ROOTTEST_ADD_TEST(testFriend3
+                  MACRO testFriend2.cc
```
pcanal:
```suggestion
                  MACRO testFriend3.cc
```

roottest/root/treeformula/casting/Makefile
```diff
@@ -1,30 +0,0 @@
...
-   endif
-endif
-
-include $(ROOTTEST_HOME)/scripts/Rules.mk
-
-Event.root: copiedEvent$(ExeSuf)
-	$(CMDECHO) ./copiedEvent 20 1 9 1
-
-Simple.log: Simple_cxx.$(DllSuf) Read.C Create.C Run.C
-ifneq ($(ClingWorkAroundCallfuncAndInline),)
```
pcanal, line 21: This does not seem to be needed anymore here.  We should investigate whether (as it is likely) this work-around is no longer needed at all.

roottest/root/treeformula/schemaEvolution/schemaRun1.C
```diff
@@ -1,26 +1,12 @@
+int schemaRun1()
 {
-#if  defined(ClingWorkAroundIncorrectTearDownOrder)
```
pcanal, line 2: Those were intentionally left as marker of known issues to be resolved/investigated later.  If we are removing them from the test they should be replaced by github issue with a reproducer(s) (or if they are resolved already we should remove the workarounds everywhere).

roottest/root/treeformula/sync/CMakeLists.txt
```diff
@@ -1,9 +1,40 @@
...
+                  FIXTURES_SETUP root-treeformula-sync-writefile-fixture)
+
+ROOTTEST_ADD_TEST(readfile-0
+                  MACRO sync.C
+                  MACROARG 0
+                  ROOTEXE_OPTS -e "gInterpreter->AddIncludePath(\"-I${CMAKE_SOURCE_DIR}/roottest/root/io/event\")"
+                               -e "gSystem->AddDynamicPath(\"${CMAKE_BINARY_DIR}/roottest/root/io/event\")"
+                               -e "(void) gSystem->Load(\"libTestIoEvent\")"
+                  OUTREF readfile-0.ref
+                  WILLFAIL
```
pcanal, line 30: The test does fail but this is (if I recall correctly) a bug and thus not a desirable failure.  We should open a github issue to record that this needs investigation.

roottest/root/treeformula/event/CMakeLists.txt
```diff
@@ -1,7 +1,49 @@
...
+  elseif(${split} EQUAL -2)
+    set(_smode old)
+    set(_sname split)
+  endif()
+
+  set(_file_name  Event.${_smode}.split${_sname}.root)
+  ROOTTEST_ADD_TEST(write-${_smode}-${_sname}
+                    COMMAND ${CMAKE_BINARY_DIR}/roottest/root/io/event/TestIoEvent ${_size} ${_comp} ${split} ${_action} ${_tracks} ${_file_name}
+                    FIXTURES_REQUIRED root-io-event-TestIoEvent-fixture
+                    FIXTURES_SETUP root-treeformula-event-write-${_sname}-fixture)
```
pcanal, line 24:
```suggestion
                    FIXTURES_SETUP root-treeformula-event-write-${_smode}-${_sname}-fixture)
```
might be better (allow more parallelism).

roottest/root/treeformula/event/CMakeLists.txt
```diff
@@ -1,7 +1,49 @@
...
+                                 -e "gSystem->AddDynamicPath(\"${CMAKE_BINARY_DIR}/roottest/root/io/event\")"
+                                 -e "gSystem->Load(\"libTestIoEvent\")"
+                    FIXTURES_REQUIRED root-io-event-libTestIoEvent-fixture
+                                      root-treeformula-event-write-${_sname}-fixture
+                    FIXTURES_SETUP root-treeformula-event-reference-fixture)
+  endif()
+
+  foreach(mode 0 1 2 3 4)
+    ROOTTEST_ADD_TEST(draw-${_smode}-${_sname}-${mode}
+                    MACRO dt_RunDrawTest.C+
```
pcanal, line 40: The Makefile was using `dt_wrap.C` which seems to negate/invert the return value of `dt_RunDrawTest`.   Why is it no longer needed (and/or where is it now done)?

roottest/root/io/event/CMakeLists.txt
linev: I did not saw any reference files which were checked during dictionary build for the classes in `Event.h` 

And when running write and read command - it was just extra `echo` output after normal return of executable:
```
bigeventTest.write.log: ./Event$(ExeSuf) ./libEvent.$(DllSuf)
	$(CMDECHO) $(call locked_execution,bigeventTest.write, sleep 10; (./Event 10 1 1 1;) > Event.write.log 2>&1 && echo "write success" > bigeventTest.write.log,test, echo "write success" > bigeventTest.write.log)

bigeventTest.read.log: ./Event$(ExeSuf) ./libEvent.$(DllSuf) bigeventTest.write.log
	$(CMDECHO)./Event 10 1 1 20 > Event.write.log 2>bigeventTest.read.log && echo "read success" > bigeventTest.read.log
```
So checking return value here should be enough

roottest/root/tree/fastcloningeventtree/dt_DrawTest.C
linev: Here I copied tests which are used in `roottest/root/treeformula/event` directory.
And tiny difference here - use of "fastclone" in-between.
So all macros can be the same.

roottest/root/tree/cache/Event.h
linev: If this was significant for tree/cache test - I will reintroduce custom Event.h here. I was think that in such case it does not affect testing of tree cache.

roottest/root/tree/friend/CMakeLists.txt
linev: Yes, you are right. I add commit with:
```
OUTCNVCMD sed -e "s: file [A-Za-z/\].*[/\]: file :g"
```

roottest/root/tree/friend/CMakeLists.txt
linev: Yes, will be fixed

roottest/root/treeformula/casting/Makefile
linev: That kind of workaround?
Here I see dependence from source code and from compiled macro.

roottest/root/treeformula/schemaEvolution/schemaRun1.C
linev: But it is very generic question - should one mix testing of very different issues in one macro?
`ClingWorkAroundIncorrectTearDownOrder` is about cling cleanup problem in unnamed macros.
But should we test/handle this problem together with schema evolution in TTreeFormula?
Why not simplify scripts and concentrate only on one concrete problem?

roottest/root/treeformula/sync/CMakeLists.txt
linev: In Makefile `Run.C(0)` was executed only when failing tests were selected - `ifneq ($(FAIL),)`
Here is issue: https://github.com/root-project/root/issues/19238

roottest/root/treeformula/event/CMakeLists.txt
linev: It is not necessary. `_sname` is unique for all kind of split args. `_smode` is just extra variable which can be `new` or `old` and only used in the file names. Or in test names.

roottest/root/treeformula/event/CMakeLists.txt
linev: While I modify `dt_RunDrawTest.C` to return expected value.

roottest/root/treeformula/sync/CMakeLists.txt
pcanal: Thanks.

roottest/root/treeformula/schemaEvolution/schemaRun1.C
pcanal: > Why not simplify scripts and concentrate only on one concrete problem?

This would be ideal.  However the consequence is that one need to ensure that the removed issues are indeed tested elsewhere and/or reported with enough detail to reproduce.

roottest/root/io/event/CMakeLists.txt
pcanal: The logs were concatenated and checked together ... ( no real reason to keep the concatenation part).
```
bigeventTest.log: bigeventTest.write.log bigeventTest.read.log
        $(CMDECHO) cat bigeventTest.write.log bigeventTest.read.log > bigeventTest.log

bigeventTest.success: bigeventTest.log
        $(CMDECHO) diff -u -b bigeventTest.ref bigeventTest.log && touch bigeventTest.success

bigeventTest: bigeventTest.log bigeventTest.success
        $(TestDiff)
```

roottest/root/tree/fastcloningeventtree/dt_DrawTest.C
pcanal: > And tiny difference here - use of "fastclone" in-between.

What do you mean?

roottest/root/treeformula/casting/Makefile
pcanal: Cling/JIT uses to not handle well inlined function.  This PR is removing ClingWorkAroundCallfuncAndInline locally.  My question was whether there was other place it was used and if it could be removed from there too.

roottest/root/tree/cache/Event.h
pcanal: All 4 seemed to have each their own quirk/changes.  Did you catalog and verify that each of those changes were 'innocuous'?

roottest/root/treeformula/schemaEvolution/schemaRun1.C
linev: `ClingWorkAroundIncorrectTearDownOrder` used in many other places.
And by using this `#define` we do not test some feature - but workaround problem.

roottest/root/io/event/CMakeLists.txt
linev: But logs were produce by `echo "write success"`  and `echo "read success"` in the Makefile - not by application itself. Why we want to repeat this logic one-to-one? I guess checking return code is enough.

roottest/root/tree/fastcloningeventtree/dt_DrawTest.C
linev: I mean - I copied drawing macros from other test. 
And difference compere to other test - adding `CloneTree(-1,"fast");` between data generation and drawing.
So many steps in such tests remains identical.

roottest/root/treeformula/casting/Makefile
linev: `ClingWorkAroundCallfuncAndInline` used in several other sources, but not activated in cmake-based tests.

roottest/root/tree/cache/Event.h
linev: Actually there is change in ref file. 

https://github.com/root-project/root/pull/19173/commits/9096ad25e24b118ab03bdada07880a1ceaad6792#diff-3ddd6718ef7ba5b8d82ebc929c7a4585e359df318971c763ac2de90b51f86a96

Now with LZ4 and ZLIB same number of I/O operations are performed.

My guess was that we testing reproducibility of TTree caching mechanism.
With slightly different event class we may have different number of I/O actions.
And now we just will test that new numbers stays constant.

roottest/root/treeformula/schemaEvolution/schemaRun1.C
pcanal: Indeed but we had/have a large set of example to use when time comes to investigate the issue (it is also indeed unknown how many of those examples are redundant for the purpose of each work around).

roottest/root/io/event/CMakeLists.txt
pcanal: > not by application itself. 

Except in case of failure.  I.e we had:
```
Event ...    2>bigeventTest.read.log       && echo "read success" > bigeventTest.read.log
```
So in case of failure the error output was diffed.

So net effect was to display/give 'easy' access to the error output in case of error.

As long as this is still somewhat the case (i.e. that the output can be seen in case of error) then we are fine.

roottest/root/treeformula/casting/Makefile
pcanal: Fair enough, I had forgotten the note on this (f932ba45b93a8c91f498f81bad8fcd689c3063d0)
```
Disable ClingWorkAroundCallfuncAndInline but keep it around just in case we have to take one step back when introducing the module
```
Now that module have been throughly tested we can remove `ClingWorkAroundCallfuncAndInline` (and associated code) entirely.

roottest/root/io/compression/CMakeLists.txt
```diff
@@ -1,30 +1,13 @@
...
-
-set(values 0 101 208 301 404 505 5)
-
-foreach(comp ${values})
   ROOTTEST_ADD_TEST(write-${comp}
-    EXEC ./MainEvent
-    OPTS 20 ${comp} 1 1
-    FIXTURES_REQUIRED io-compression-exec
-    FIXTURES_SETUP io-compression-write-${comp})
+                    COMMAND ${CMAKE_BINARY_DIR}/roottest/root/io/event/TestIoEvent 20 ${comp} 1 1 600 ${file_name}
```
pcanal:
```suggestion
                    COMMAND ${CMAKE_BINARY_DIR}/roottest/root/io/bigevent/TestIoEvent 20 ${comp} 1 1 600 ${file_name}
```
`roottest/root/io/compression/Event.h` (and co.) are copies of `roottest/root/io/bigevent/Event.h`

roottest/root/io/compression/CMakeLists.txt
```diff
@@ -1,30 +1,13 @@
...
+                    FIXTURES_REQUIRED root-io-event-TestIoEvent-fixture
+                    FIXTURES_SETUP root-io-compression-write-${comp}-fixture)
 
   ROOTTEST_ADD_TEST(read-${comp}
-    EXEC ./MainEvent
-    OPTS 20 ${comp} 1 20
-    FIXTURES_REQUIRED io-compression-write-${comp})
+                    COMMAND ${CMAKE_BINARY_DIR}/roottest/root/io/event/TestIoEvent 20 ${comp} 1 20 600 ${file_name}
```
pcanal:
```suggestion
                    COMMAND ${CMAKE_BINARY_DIR}/roottest/root/io/bigevent/TestIoEvent 20 ${comp} 1 20 600 ${file_name}
```

roottest/root/treeformula/event/Event.h
```diff
@@ -1,203 +0,0 @@
-#ifndef ROOT_Event
```
pcanal, line 1: Fine to replace with `roottest/root/io/event`.  The only significant difference are:
```
40,41c43,44
<    Double32_t   fCharge;       //Charge of this track
<    Double32_t   fVertex[3];    //Track vertex position
---
>    Double32_t   fCharge;       //[-1,1,2] Charge of this track
>    Double32_t   fVertex[3];    //[-30,30,16] Track vertex position
45c48
<    Double32_t*  fPointValue;   //[fNsp] a special quantity for some point.
---
>    Double32_t*  fPointValue;   //[fNsp][0,3] a special quantity for some point.
```
where the customized `Double32_t` are actually improve test coverage per se.

roottest/root/treeformula/retobj/Event.h
```diff
@@ -1,221 +0,0 @@
-#ifndef ROOT_Event
```
pcanal, line 1: Fine to replace with roottest/root/io/event.  The main differences are improvements in particular the implementation of a correct copy constructor for the Track class. (the other major change is using and customizing `Double32_t` a few more member:
```
40,41c43,44
<    Float_t      fCharge;       //Charge of this track
<    Float_t      fVertex[3];    //Track vertex position
---
>    Double32_t   fCharge;       //[-1,1,2] Charge of this track
>    Double32_t   fVertex[3];    //[-30,30,16] Track vertex position
45,46c48,52
<    Float_t*     fPointValue;   //[fNsp] a special quantity for some point.
<    TBits        fTrigBits;     // 
---
>    Double32_t*  fPointValue;   //[fNsp][0,3] a special quantity for some point.
>    TBits        fTriggerBits;  // Bits triggered by this track.
> 
```

roottest/root/meta/Event.h
```diff
@@ -1,203 +0,0 @@
-#ifndef ROOT_Event
```
pcanal, line 1: Fine to replace with `roottest/root/io/event`.  The only significant difference are:
```
40,41c43,44
<    Double32_t   fCharge;       //Charge of this track
<    Double32_t   fVertex[3];    //Track vertex position
---
>    Double32_t   fCharge;       //[-1,1,2] Charge of this track
>    Double32_t   fVertex[3];    //[-30,30,16] Track vertex position
45c48
<    Double32_t*  fPointValue;   //[fNsp] a special quantity for some point.
---
>    Double32_t*  fPointValue;   //[fNsp][0,3] a special quantity for some point.
```
where the customized `Double32_t` are actually improve test coverage per se.

roottest/root/io/compression/Event.h
```diff
@@ -1,292 +0,0 @@
-#ifndef ROOT_Event
```
pcanal, line 1: It is fine to replace this Event with the one in `roottest/root/io/bigevent`

roottest/root/tree/cache/Event.h
pcanal: After checking, the difference indeed result in basket size differences that lead to a different layout on the file and thus difference fetching by the cache.  I agree that the difference is spurious in regard to this test.

roottest/root/io/event/CMakeLists.txt
linev: One will see failing test if application returns value other than 0.

roottest/root/io/compression/CMakeLists.txt
linev: I will provide my own commit - while `bigevent` executable need to be adjusted as well to let use custom file name

roottest/root/io/compression/CMakeLists.txt
linev: Will be done in my own commit

roottest/root/io/compression/Event.h
linev: Will be done

roottest/root/tree/fastcloningeventtree/dt_DrawTest.C
pcanal: I'll have to review which of the version is 'better' than the others ...

roottest/root/treeformula/casting/Makefile
linev: As far as I can see `ClingWorkAroundCallfuncAndInline` is not used in the macros touched by this PR.
One can make extra PR which exclude testing of this no-longer-used defines

roottest/root/treeformula/schemaEvolution/schemaRun1.C
pcanal: For the purpose of this PR and to keep the spirit of `ClingWorkAroundIncorrectTearDownOrder`, please do one of the following:
* Revert change to this file to keep this reproducer.
or
* Open a PR which a standalone reproducer showing  why we needed `ClingWorkAroundIncorrectTearDownOrder` (and/or link here an existing issue).

thanks.

linev: `ClingWorkAroundIncorrectTearDownOrder` still appears in 7 other tests.
This defines is not set - neither in cmake not in old `FixCling.mk` file.
It is associated with issue https://its.cern.ch/jira/browse/ROOT-4768, which was fixed before ROOT 5.28.
Is it really important to restore it?

pcanal: > This defines is not set - neither in cmake not in old FixCling.mk file.

I missed that. So the point is moot.  Thanks.

roottest/root/treeformula/event/dt_DrawTest.C
```diff
@@ -128,31 +128,30 @@ TDirectory* GenerateDrawHist(TTree *tree,int level = 3, int quietLevel = 0)
...
    // Test on vector of numerical type
-   tree->Draw("fVClosestDistance>>hVClosestDistance","","goff");
-   tree->Draw("fVClosestDistance[2]>>hVClosestDistance2","","goff");
-   tree->Draw("fVClosestDistance[9]>>hVClosestDistance9","","goff");
-   DrawSkippable(tree,"fVClosestDistance[fNvertex/2]","hVClosestDistanceIndex",
-                 (level>0));
+   // tree->Draw("fVClosestDistance>>hVClosestDistance","","goff");
+   // tree->Draw("fVClosestDistance[2]>>hVClosestDistance2","","goff");
+   // tree->Draw("fVClosestDistance[9]>>hVClosestDistance9","","goff");
+   // DrawSkippable(tree,"fVClosestDistance[fNvertex/2]","hVClosestDistanceIndex", (level>0));
```
pcanal: Why do we need to stop testing those?

roottest/root/tree/fastcloningeventtree/dt_RunDrawTest.C
```diff
@@ -182,22 +172,25 @@ bool dt_RunDrawTest(const char* from, Int_t mode = 0, Int_t verboseLevel = 0) {
...
+     return 1;
    }
    DrawMarks();
 
-   if (gQuietLevel<2) cout << "DrawTest: Comparison was successfull" << endl;
+   if (gQuietLevel<2)
+      std::cout << "DrawTest: Comparison was successfull" << std::endl;
    if (hfile) delete hfile;
-   else delete tree;
+         else delete tree;
```
pcanal, line 193: (minor) The indentation looks wrong on 'my' github display (it should the `else` starting at the same column as the 'i' in `hfile`.

roottest/root/treeformula/event/dt_DrawTest.C
linev: This member not exists in generic libEvent - it was appearing only in copy inside `roottest/root/treeformula/event` directory

pcanal: That's a bummer ... we probably need to add them to the generic libEvent then ... they are there to test specific features of the I/O and `TTree::Draw` ....

ferdymercury:
Fixes https://github.com/root-project/root/issues/14579

roottest/root/treeformula/event/dt_DrawTest.C
linev: I add vector-based members into basic `root/io/event/Event.h` class and enable drawing test with these memebrs again. I had to rebase PR to be able run CI

roottest/root/treeformula/event/dt_DrawTest.C
```diff
@@ -138,21 +138,20 @@ TDirectory* GenerateDrawHist(TTree *tree,int level = 3, int quietLevel = 0)
    tree->Draw("fVClosestDistance>>hVClosestDistance","","goff");
    tree->Draw("fVClosestDistance[2]>>hVClosestDistance2","","goff");
    tree->Draw("fVClosestDistance[9]>>hVClosestDistance9","","goff");
-   DrawSkippable(tree,"fVClosestDistance[fNvertex/2]","hVClosestDistanceIndex",
-                 (level>0));
+   DrawSkippable(tree,"fVClosestDistance[fNvertex/2]","hVClosestDistanceIndex", (level>0));
 
    // Test on vector of objects:
-   DrawSkippable(tree,"fVEvtHdr.fRun","hVRun",(level>2));
+   DrawSkippable(tree,"fEvtHdr.fRun","hRun",(level>2));
```
pcanal: Now that the vector data member were added, this change should not bee needed any more, right?
```suggestion
   DrawSkippable(tree,"fVEvtHdr.fRun","hVRun",(level>2));
```

roottest/root/treeformula/event/dt_DrawTest.C
```diff
@@ -138,21 +138,20 @@ TDirectory* GenerateDrawHist(TTree *tree,int level = 3, int quietLevel = 0)
...
    DrawSkippable(tree,"fVEvtHdr.fRun[2]","hVRun2",(level>2));
    DrawSkippable(tree,"fVEvtHdr.fRun[fNvertex/2]","hVRunIndex",(level>2));
-   DrawSkippable(tree,"fVEvtHdr.GetRun()","hVRunFunc",(level>2 && gHasLibrary));
-   
+   DrawSkippable(tree,"fEvtHdr.GetRun()","hRunFunc",(level>2 && gHasLibrary));
```
pcanal: See https://github.com/root-project/root/pull/19173/files#r2267140991

roottest/root/treeformula/event/dt_DrawTest.C
```diff
@@ -138,21 +138,20 @@ TDirectory* GenerateDrawHist(TTree *tree,int level = 3, int quietLevel = 0)
...
+
    // Test on vector of pointer
-   DrawSkippable(tree,"fVTracks.fPx","hVPx",(level>2));
-   DrawSkippable(tree,"fVTracks.fCharge","fVTracks.fPx < 0","hVCharge",(level>2));
+   // DrawSkippable(tree,"fVTracks.fPx","hVPx",(level>2));
+   // DrawSkippable(tree,"fVTracks.fCharge","fVTracks.fPx < 0","hVCharge",(level>2));
```
pcanal: See https://github.com/root-project/root/pull/19173/files#r2267140991

linev: Yes, you are right

linev: I enable these tests too now

pcanal: Thanks.

PR #19175, opened 2025-06-25
[roottest] Fix typo in required fixture

hahnjo:
This could cause `datamodelevolution-cms-00-read_reflex_test1` to fail because "reflex_testv1.root does not exist"


hahnjo:
Sorry, didn't see https://github.com/root-project/root/pull/19086 because it was already merged

Issue #19125, opened 2025-06-22
Tutorial `RBatchGenerator_TensorFlow.py` sporadically fails


guitargeek:
### Description

The `RBatchGenerator_TensorFlow.py` tutorial sporadically fails on `alma8` reproducibly with the following error:

```txt
2025-06-22T09:55:16.5679633Z 2985/2996 Test #1366: tutorial-machine_learning-RBatchGenerator_TensorFlow-py ...........................................***Timeout 1500.10 sec
2025-06-22T09:55:16.5681333Z 2025-06-22 11:30:16.779682: I tensorflow/core/util/port.cc:110] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
2025-06-22T09:55:16.5688011Z 2025-06-22 11:30:16.781562: I tensorflow/tsl/cuda/cudart_stub.cc:28] Could not find cuda drivers on your machine, GPU will not be used.
2025-06-22T09:55:16.5688914Z 2025-06-22 11:30:16.848932: I tensorflow/tsl/cuda/cudart_stub.cc:28] Could not find cuda drivers on your machine, GPU will not be used.
2025-06-22T09:55:16.5690059Z 2025-06-22 11:30:16.849429: I tensorflow/core/platform/cpu_feature_guard.cc:182] This TensorFlow binary is optimized to use available CPU instructions in performance-critical operations.
2025-06-22T09:55:16.5691336Z To enable the following instructions: AVX2 AVX512F AVX512_VNNI AVX512_BF16 FMA, in other operations, rebuild TensorFlow with the appropriate compiler flags.
2025-06-22T09:55:16.5692492Z 2025-06-22 11:30:17.567088: W tensorflow/compiler/tf2tensorrt/utils/py_utils.cc:38] TF-TRT Warning: Could not find TensorRT
2025-06-22T09:55:16.5693076Z Epoch 1/2
2025-06-22T09:55:16.5693195Z 
2025-06-22T09:55:16.5693857Z  1/54 [..............................] - ETA: 1:47 - loss: 0.7541 - accuracy: 0.3516
2025-06-22T09:55:16.5694651Z  7/54 [==>...........................] - ETA: 0s - loss: 0.1297 - accuracy: 0.9074  
[...]

PR #19176, opened 2025-06-25
[CMake] CMake / include fixes for upcoming target-based CMake

hageboeck:
These are commits cherry-picked from #8709. These should be able to stand on their own, but they will be required for the upcoming changes in #8709.

Concretely, we have here:
- Moving a few include directives to where they are used.
- Move setting of `OSX_SYSROOT` before the `project()` call as required in the CMake documentation
- Make an implicit dependency of sofie explicit


vepadulano (approved):
Thanks!

guitargeek:
Good that the failure on `alma8` happened, it's only the second time I see it!

So new sporadic confirmed:
https://github.com/root-project/root/issues/19125

Issue #18949, opened 2025-06-03
`Vc.pcm` not found when running root after building with external Vc with runtime_cxxmodules ON


jmcarcell:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

I have been trying to turn `runtime_cxxmodules` ON in Spack (https://github.com/spack/spack/pull/50432) and I have found that when building with a standalone Vc and then running root I get the following error (reproduced on Alma 9 with GCC 14, on Ubuntu 24.04 with GCC 13 and in the Spack CI).
```
<<< cling interactive line includer >>>: fatal error: module file '/tmp/root/spack-stage/spack-stage-root-6.36.00-3zoafy6mjx64vkgps6eq5nhuamqu57ri/spack-build-3zoafy6/lib/Vc.pcm' not found: module file not found
<<< cling interactive line includer >>>: note: imported by module 'MathCore' in '/cvmfs/sw-nightlies.hsf.org/key4hep/releases/2025-05-31/x86_64-almalinux9-gcc14.2.0-opt/root/6.36.00-3zoafy/lib/root/MathCore.pcm'
input_line_4:2:2: error: use of undeclared identifier 'include'
```
followed by a bunch of errors that depend on exactly what I'm running with root. The `/tmp/root...` folder is the build directory.

### Reproducer

Build Vc standalone, and then build ROOT with `-Dbuiltin_vc=OFF` and `-Druntime_cxxmodules=ON` and then delete the build directory. If the build directory is not deleted, the error is not triggered since `Vc.pcm` can be found. This problem is also fixed by changing to `-Dbuiltin_vc=ON`. 

### ROOT version

I only tested this on v6.36.00

### Installation method

Spack, source

### Operating system

Linux: Alma 9 and Ubuntu 24.04

guitargeek:
Hi! Thanks for the report. I think @LukasBreitwieser is already looking at the equivalent problem in a different environment, so we're on it.

jmcarcell:
This is not fixed after https://github.com/root-project/root/pull/19179 and 3f8c86e509bf3af48c7ef8ed39147e7883a2f9ae, I keep getting the same error about `Vc.pcm` not found in the build directory building today's master (3f8c86e509bf3af48c7ef8ed39147e7883a2f9ae). Build options for completeness from Spack (same as before but I think I had to add `builtin_civetweb`:
```
[...]

PR #19179, opened 2025-06-25
Make non-system include paths accessible at runtime

LukasBreitwieser:
The current implementation faces issues with accessing header files from external libraries stored in non-system paths when the C++ interpreter tries to resolve them at runtime. This access is necessary under certain conditions even if a corresponding module file exists. Related issues: [SPI-2794](https://its.cern.ch/jira/browse/SPI-2794), https://github.com/root-project/root/issues/18949

To address this issue, we need to ensure that non-system path includes, which are known at build time, remain accessible at runtime. The proposed solution leverages the existing `ROOT_INCLUDE_PATH` environment variable.

This commit introduces a new CMake variable `DEFAULT_ROOT_INCLUDE_PATH` and a helper function `IF_NOT_SYSTEM_PATH_APPEND`. This function allows us to conditionally append include paths of external libraries to `DEFAULT_ROOT_INCLUDE_PATH`, for example `IF_NOT_SYSTEM_PATH_APPEND("${Vc_INCLUDE_DIR}" DEFAULT_ROOT_INCLUDE_PATH)`

The `DEFAULT_ROOT_INCLUDE_PATH` variable is then utilized to populate the `ROOT_INCLUDE_PATH` environment variable in the configuration files (`thisroot.{sh,csh,fish,bat}`).

If the library's include path is dynamically modified between build and run time, the configuration can be updated by patching the `thisroot.{sh,csh,fish,bat}` files.


cmake/modules/RootMacros.cmake
```diff
@@ -2135,3 +2135,35 @@ function(generateManual name input output)
...
+# The 1st argument is the path that should be checked
+# The 2nd argument is the return value
+#----------------------------------------------------------------------------
+function (IS_SYSTEM_PATH path is_system_path)
+  foreach (dir ${CMAKE_SYSTEM_PATH})
+    if ("${path}" STREQUALS "$dir")
+      set(${is_system_path} true PARENT_SCOPE)
+    endif()
+  endforeach()
+  set(${is_system_path} false PARENT_SCOPE)
```
hageboeck: The canonical bools are all caps, and we can break out early:
```suggestion
      set(${is_system_path} TRUE PARENT_SCOPE)
      return()
    endif()
  endforeach()
  set(${is_system_path} FALSE PARENT_SCOPE)
```

config/thisroot.sh
```diff
@@ -89,6 +89,10 @@ clean_environment()
          default_manpath=""
       fi
    fi
+   if [ -n "${ROOT_INCLUDE_PATH-}" ]; then
+      drop_from_path "$ROOT_INCLUDE_PATH" "@DEFAULT_ROOT_INCLUDE_PATH@"
+      ROOT_INCLUDE_PATH=$newpath
```
hageboeck, line 94: `newpath` seems to be undefined here, or did I miss something?

cmake/modules/RootMacros.cmake
```diff
@@ -2135,3 +2135,35 @@ function(generateManual name input output)
...
+endfunction()
+
+#----------------------------------------------------------------------------
+# If path is not a system path, append it to the given variable using 
+# CMAKE_PATH_SEPARATOR
+# The 1st argument is the path to be checked
+# The 2nd argument is the variable that the path gets appended to if it is 
+# not a system path.
+#----------------------------------------------------------------------------
+function (IF_NOT_SYSTEM_PATH_APPEND path variable)
```
hageboeck: Is this function ever going to be called with another variable than the default root include path? If yes, fair enough, otherwise I would call it something like `BUILD_ROOT_INCLUDE_PATH("${Vc_INCLUDE_DIR}")`, and remove the second argument, since any negations in names are hard to parse.

config/thisroot.sh
```diff
@@ -162,6 +166,14 @@ set_environment()
    else
       JUPYTER_CONFIG_DIR=$ROOTSYS/etc/notebook:$JUPYTER_CONFIG_DIR; export JUPYTER_CONFIG_DIR
    fi
+
+   if [ -z "${ROOT_INCLUDE_PATH-}" ]; then
+      ROOT_INCLUDE_PATH=@DEFAULT_ROOT_INCLUDE_PATH@
+      export ROOT_INCLUDE_PATH
+   else
+      ROOT_INCLUDE_PATH=@DEFAULT_ROOT_INCLUDE_PATH@:$ROOT_INCLUDE_PATH
+      export ROOT_INCLUDE_PATH
```
hageboeck, line 175:
```suggestion
      export ROOT_INCLUDE_PATH=@DEFAULT_ROOT_INCLUDE_PATH@
   else
      export ROOT_INCLUDE_PATH=@DEFAULT_ROOT_INCLUDE_PATH@:$ROOT_INCLUDE_PATH
```

hageboeck (changes requested):
Nice! I added a few comments below

hageboeck:
@LukasBreitwieser, what happens when you install (and therefore relocate) ROOT? Given that these paths are system paths, things should continue to work, won't they?

cmake/modules/RootMacros.cmake
```diff
@@ -2135,3 +2135,35 @@ function(generateManual name input output)
...
+#----------------------------------------------------------------------------
+# If path is a system path, set the return variable is_system_path to true.
+# The 1st argument is the path that should be checked
+# The 2nd argument is the return value
+#----------------------------------------------------------------------------
+function (IS_SYSTEM_PATH path is_system_path)
+  foreach (dir ${CMAKE_SYSTEM_PATH})
+    if ("${path}" STREQUALS "$dir")
+      set(${is_system_path} true PARENT_SCOPE)
+    endif()
```
andresailer, line 2128:
```suggestion
    return()
    endif()
```

Maybe my cmake is too rusty, but shouldn't we not always set is_system_path to false at the end of the function?

config/thisroot.sh
```diff
@@ -162,6 +166,14 @@ set_environment()
    else
       JUPYTER_CONFIG_DIR=$ROOTSYS/etc/notebook:$JUPYTER_CONFIG_DIR; export JUPYTER_CONFIG_DIR
    fi
+
+   if [ -z "${ROOT_INCLUDE_PATH-}" ]; then
+      ROOT_INCLUDE_PATH=@DEFAULT_ROOT_INCLUDE_PATH@
+      export ROOT_INCLUDE_PATH
+   else
+      ROOT_INCLUDE_PATH=@DEFAULT_ROOT_INCLUDE_PATH@:$ROOT_INCLUDE_PATH
```
andresailer, line 174: Do you consider this an issue if DEFAULT_ROOT_INCLUDE_PATH is empty? Not just here but in the other shell setups?

cmake/modules/RootMacros.cmake
hageboeck: Indeed, I think the return is needed

config/thisroot.sh
```diff
@@ -162,6 +166,14 @@ set_environment()
    else
       JUPYTER_CONFIG_DIR=$ROOTSYS/etc/notebook:$JUPYTER_CONFIG_DIR; export JUPYTER_CONFIG_DIR
    fi
+
+   if [ -z "${ROOT_INCLUDE_PATH-}" ]; then
```
hageboeck, line 170: The variable name contains an extra `-` if I'm not mistaken, so this would always be zero.

andresailer:
I am getting an error in the lcg build with this

```
[100%] Generating tutorials/hsimple.root
root.exe: /build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-default-root-include-path/src/ROOT/default-root-include-path/interpreter/llvm-project/clang/lib/Serialization/ASTReader.cpp:4332: clang::ASTReader::ASTReadResult clang::ASTReader::ReadModuleMapFileBlock(RecordData&, ModuleFile&, const ModuleFile*, unsigned int): Assertion `M && M->Name == F.ModuleName && "found module with different name"' failed.
```
https://lcgapp-services.cern.ch/cdash/build/98193/files

config/thisroot.sh
andresailer: This is bash syntax for using a default value (of nothing) when the variable is not set.

hageboeck: OK, I had to read up on the subtle difference between `${VAR:-}` vs `${VAR-}` ...

pcanal:
What do you practically mean by:
```
If the library's include path is dynamically modified between build and run time, the configuration can be updated by patching the thisroot.{sh,csh,fish,bat} files.
```
?

config/thisroot.sh
LukasBreitwieser: `newpath` is set by the function drop_from_path.

config/thisroot.sh
LukasBreitwieser: I don't think this should be an issue. This case is also covered by all GHA runners.

config/thisroot.sh
LukasBreitwieser: For the sake of consistency, I followed the style of all other environment variables in `thisroot.sh`

LukasBreitwieser:
Thank you all for your comments!

@hageboeck 
> @LukasBreitwieser, what happens when you install (and therefore relocate) ROOT? Given that these paths are system paths, things should continue to work, won't they?

Yes this works. Only third party include paths (i.e., include paths outside the root src/build/install directories) should be added. Perhaps a good idea to add additional checks for these conditions.

@pcanal
If the include path (to e.g. Vc) at runtime is different to the one at build time, which might happen for spack installations using its binary cache.
In that case the `thisroot.*` files must be patched.

@andresailer 
Thank you for running the lcg build. Will investigate the failure.

cmake/modules/RootMacros.cmake
```diff
@@ -2135,3 +2135,46 @@ function(generateManual name input output)
...
+# If path is a system path, set the return variable is_system_path to true.
+# The 1st argument is the path that should be checked
+# The 2nd argument is the return value
+#----------------------------------------------------------------------------
+function (IS_SYSTEM_PATH path is_system_path)
+  foreach (dir ${CMAKE_SYSTEM_PATH})
+    if ("${path}" STREQUALS "$dir")
+      set(${is_system_path} TRUE PARENT_SCOPE)
+    endif()
+    return()
```
andresailer: If at all, the return is needed before the endif -- like I proposed -- not after.

LukasBreitwieser:
Reported CI test failures also present on yesterdays ROOT Main nightly.

@andresailer I sent a PR for [lcgcmake](https://gitlab.cern.ch/sft/lcgcmake/-/merge_requests/3202).
I ran the dev3 lcgcmake build on alma9 and root-tests  passed. Do you want to check the other stacks too?

andresailer:
I think for me this is fine as it is. Thanks!

andresailer:
What is the status here? Is this going to get merged soon?

andresailer:
The tests added in https://github.com/root-project/root/pull/19667
are not working in the nightly tests because
https://github.com/root-project/root/blob/c8faaaf223455e6683c53ce2c805312113677443/tutorials/CMakeLists.txt#L23
overwrites our existing ROOT_INCLUDE_PATH. So would be nice if this here is merged, or we need a fix in that other line.

cmake/modules/RootMacros.cmake
```diff
@@ -2114,6 +2114,48 @@ function(ROOT_APPEND_LIBDIR_TO_INSTALL_RPATH target install_dir)
   set_property(TARGET ${target} APPEND PROPERTY INSTALL_RPATH "${new_rpath}")
 endfunction()
 
+#----------------------------------------------------------------------------
+# If path is a system path, set the return variable is_system_path to true.
+# The 1st argument is the path that should be checked
+# The 2nd argument is the return value
+#----------------------------------------------------------------------------
+function (IS_SYSTEM_PATH path is_system_path)
+  foreach (dir ${CMAKE_SYSTEM_PATH})
```
hageboeck: I don't think this variable is part of (the official) CMake, so this might always be false.

cmake/modules/RootMacros.cmake
```diff
@@ -2114,6 +2114,51 @@ function(ROOT_APPEND_LIBDIR_TO_INSTALL_RPATH target install_dir)
...
 endfunction()
 
+#----------------------------------------------------------------------------
+# If path is a system include path, set the return variable
+# is_system_include_path to true.
+# The 1st argument is the path that should be checked
+# The 2nd argument is the return value
+#----------------------------------------------------------------------------
+function (IS_SYSTEM_INCLUDE_PATH path is_system_include_path)
+  message("CMAKE_SYSTEM_INCLUDE_PATH: ${CMAKE_SYSTEM_INCLUDE_PATH}")
```
hageboeck: Let's not forget to remove this before merging.
```suggestion
```

CMakeLists.txt
```diff
@@ -594,7 +591,8 @@ if(WIN32)
   set(hsimple_cmd COMMAND ${CMAKE_COMMAND} -E env PATH="${CMAKE_RUNTIME_OUTPUT_DIRECTORY}\\\;%PATH%"
                           ROOTIGNOREPREFIX=1 ROOT_HIST=0 $<TARGET_FILE:root.exe> -l -q -b -n -x ${CMAKE_SOURCE_DIR}/tutorials/hsimple.C -e return)
 else()
-  set(hsimple_cmd COMMAND ${MODULES_ROOT_INCPATH} ${ld_library_path}=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}:$ENV{${ld_library_path}}
+  set(hsimple_cmd COMMAND ROOT_INCLUDE_PATH="${DEFAULT_ROOT_INCLUDE_PATH}" ${ld_library_path}=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}:$ENV{${ld_library_path}}
+                          ROOT_INCLUDE_PATH=${DEFAULT_ROOT_INCLUDE_PATH}
```
hageboeck: This looks like a small oversight:
```suggestion
  set(hsimple_cmd COMMAND ROOT_INCLUDE_PATH="${DEFAULT_ROOT_INCLUDE_PATH}" ${ld_library_path}=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}:$ENV{${ld_library_path}}
```

hageboeck (approved):
Please see a possible oversight below, but otherwise, LGTM!

+don't forget to squash, please. 🙂

CMakeLists.txt
andresailer: I think the additional unquoted `ROOT_INCLUDE_PATH=${DEFAULT_ROOT_INCLUDE_PATH}` actually made things work. Because the quotes get escaped by cmake and then things no longer work.

hageboeck: This might well be the case. @LukasBreitwieser ?

andresailer: I can tell you things definitely do not work (https://lcgapp-services.cern.ch/cdash/build/119422/files) and if I drop the quotes things are working.

PR #19181, opened 2025-06-25
Ensure that hadd python tests are run only if pytest is available

LukasBreitwieser:
Change CMake function from ROOTTEST_ADD_TEST to ROOT_ADD_PYUNITTEST


vepadulano (approved):

linev:
Hi,

There are one problem using `ROOT_ADD_PYUNITTEST` test.
If python script runs in source directory - it also creates `__pycache__` subdirectory deep in root sources.
To avoid this - one should copy first python script to build directory.
Therefore one need to modify `ROOT_ADD_PYUNITTEST` macro.

PR #19182, opened 2025-06-25
[RDF] Move Snapshot-related functions / types to their own files.

hageboeck:
Split all Snapshot-related things from the rest of the RDF action helpers.
Used include-what-you-use to minimise the amount of dead includes in the affected files.

This should be pretty much an NFC, since all code was just moved and clang-formatted, but since I also thinned down the includes, we of course need the full CI run.


vepadulano:
Can you apply clang-format suggestions before merging?

PR #19183, opened 2025-06-25
[v636] Backports of recent bugfixes in h2root and Python interface

guitargeek:
Backport of todays bugfixes to the release branch.


pcanal:
[This comment actually apply equally to the original and to the backport]

>  enable Fortran on all Linux 

We should consider disabling it on at least one Linux to ensure that we test what user will usually use (i.e. (even if it is unlikely) we could later on mistakenly introduce code in the 'if fortran is on' CMake code path that is necessary for the other CMake code path).

guitargeek:
That would be the GPU config if finally enabled :) Until we have that, maybe we use the Clang build? Should have no effect on Fortran. The current march native is not the right choice, because as we saw it's h2root is more sensitive

PR #19185, opened 2025-06-26
[DEBUG] change mac-beta build config to debug

aaronj0:
Investigating https://github.com/root-project/root/actions/runs/15865243002/job/44730775303#step:8:30482,  seg fault in `roottest-pyhton-basic-datatype`

### DIFF
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -108,7 +108,7 @@ jobs:
           - platform: mac-beta
             is_special: true
             arch: ARM64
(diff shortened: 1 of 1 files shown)


guitargeek:
Independent of the investigation: should be maybe always run `mac-beta` with Debug? You see any drawbacks?

dpiparo:
no, it should become debug. I would do this as a customisation in the yaml, not in the build options files, so to have release binaries.

guitargeek:
Closing, as this was only for a temporary investigation.

PR #19184, opened 2025-06-26
[roottest] examine the status of h2root on Windows

ferdymercury:
# This Pull request:

## Changes or fixes:

Check if h2root works on Windows.

if yes, @bellenot can you run the h2root test in a Debug build of Windows?

PR #19188, opened 2025-06-26
[CMake] Also consider fail-on-missing for `fortran` features

guitargeek:
Motivated by https://github.com/root-project/root/pull/19184

FYI, @ferdymercury.

The first CI run will fail because not all platforms have Fortran installed. In a second pass, `fortran` will then be disabled for all platforms where it failed (probably only Windows).


ferdymercury (approved):
Thks!

PR #19196, opened 2025-06-26
[root.exe] Use non-zero exit status when invalid arguments are passed

hageboeck:
Followup to #19195:
RDF IMT tutorials didn't run since 10 months, since the following passed with a 0 exit status:
```
$ root '-e "ROOT::EnableImplicitMT(4)"' ...
root: unrecognized option '-e "ROOT::EnableImplicitMT(4)"'
Try 'root --help' for more information.
  2965/2993 Test tutorial-analysis-dataframe-df102_NanoAODDimuonAnalysis ............. Passed 0.15 sec
```

To avoid similar problem in the future, I suggest here to exit with a non-zero exit status. I chose 2 because 1 usually means that the script/macro/command that was running *in the interpreter* failed:
```
bin/root -b -q -e "invalid"; echo $?
1
```

Whereas a failure to parse the argument would now be:
```
$ root '-e "ROOT::EnableImplicitMT(4)"'; echo $?
2
```

### DIFF
--- a/core/rint/src/TRint.cxx
+++ b/core/rint/src/TRint.cxx
@@ -181,7 +181,7 @@ TRint::TRint(const char *appClassName, Int_t *argc, char **argv, void *options, 
          std::cerr << "root: unrecognized option '" << argv[n] << "'\n";
       }
       std::cerr << "Try 'root --help' for more information.\n";
-      TApplication::Terminate(0);
+      TApplication::Terminate(2);
    }
 
    fNcmd          = 0;
--- a/core/rint/test/TRintTests.cxx
+++ b/core/rint/test/TRintTests.cxx
@@ -8,34 +8,6 @@
 using testing::internal::CaptureStderr;
 using testing::internal::GetCapturedStderr;
 
-TEST(TRint, ExitOnUnknownArgs)
-{
-   // Create array of options.
-   // We need to create it as a dynamic array for the following reasons:
-   // - TRint constructor accepts a char** so we construct directly that type
-   // - TRint will modify this array, removing recognized options and leaving
-   //   only unrecognized ones, so we can't create an std::vector and pass its
-   //   data to TRint directly.
-   int argc{4};
-   char e1[]{"-q"};
-   char e2[]{"-z"};
-   char e3[]{"--nonexistingoption"};
-   char e4[]{"-b"};
-   char *argv[]{e1, e2, e3, e4};
-
-   CaptureStderr();
-   // Unrecognized options will be printed to stderr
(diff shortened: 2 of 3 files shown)


core/rint/test/TRintTests.cxx
```diff
@@ -8,34 +8,6 @@
 using testing::internal::CaptureStderr;
 using testing::internal::GetCapturedStderr;
 
-TEST(TRint, ExitOnUnknownArgs)
```
vepadulano, line 11: Why is this test removed?

hageboeck: Because I cannot make it pass. Now that TApplication exits with status 2, the end of this test cannot be reached. It will always fail.

It is replaced by
```
 add_test(NAME root_exitStatus COMMAND $<TARGET_FILE:root.exe> "-e invalid")
 set_property(TEST root_exitStatus PROPERTY WILL_FAIL TRUE)
```
which admittedly doesn't test the error message, but it does test that root exits with != 0

hageboeck:
> LGTM.
> 
> Side note: The RUFF does not seem to handle well deleted files, see: https://github.com/root-project/root/actions/runs/15901873603/job/44846574695?pr=19196 which says:
> 
> ```
> Run files=$(cat changed_files.txt | grep '\.py$' || echo "")
> Error: tutorials/.enableImplicitMTWrapper.py:1:1: E902 No such file or directory (os error 2)
> Error: Process completed with exit code 123.
> ```

@pcanal I had noticed the same, but this problem was fixed in the mean time here: #19197

core/rint/src/TRint.cxx
```diff
@@ -181,7 +181,7 @@ TRint::TRint(const char *appClassName, Int_t *argc, char **argv, void *options,
          std::cerr << "root: unrecognized option '" << argv[n] << "'\n";
       }
       std::cerr << "Try 'root --help' for more information.\n";
-      TApplication::Terminate(0);
+      TApplication::Terminate(2);
```
silverweed, line 184: Just for curiosity: why 2? (and not, say, 1)

hageboeck: Please see description at the top:
https://github.com/root-project/root/pull/19196#issue-3178852646

core/rint/test/TRintTests.cxx
hageboeck: OK, now done @vepadulano. It's easy to test for both exit status and error message if the test is split into two.

core/rint/src/TRint.cxx
silverweed: Ah sorry, I missed that sentence.

PR #19197, opened 2025-06-26
[skip-ci][Ruff] Only include files that are added, copied, modified.

hageboeck:
Ruff currently fails to lint when a python file is deleted from the project. To fix this, I added a status filter to `git diff` which masks all files except for ones that have been **A**dded **M**odified **R**enamed

An example of this failure is in #19196.
In #19195, I already pushed the commit from this PR, and Ruff passes there.
So I think we can "smuggle" this fix in here, bypassing the full CI run and reducing the pressure on our builders.

### DIFF
--- a/.github/workflows/code_analysis.yml
+++ b/.github/workflows/code_analysis.yml
@@ -51,7 +51,7 @@ jobs:
       id: diff
       run: |
         git fetch --depth=1 origin $GITHUB_BASE_REF 
-        git diff --name-only origin/$GITHUB_BASE_REF > changed_files.txt
+        git diff --diff-filter=AMR --name-only origin/$GITHUB_BASE_REF > changed_files.txt
 
     - name: Install ruff
       uses: astral-sh/ruff-action@v3


hageboeck:
@siliataider @dpiparo if you are more comfortable running the CI, I can merge it as part of #19195, and close this PR.

siliataider (approved):
Looks good, thanks for the fix! I don't think running the full CI is necessary

.github/workflows/code_analysis.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       id: diff
       run: |
         git fetch --depth=1 origin $GITHUB_BASE_REF 
-        git diff --name-only origin/$GITHUB_BASE_REF > changed_files.txt
+        git diff --diff-filter=AMR --name-only origin/$GITHUB_BASE_REF > changed_files.txt
```
pcanal, line 54: Why 'R' (Renamed) and not 'C' (Copied)?  (Related question is a file that is renamed and modifed marked as M, MR or R).  Assuming the answer to this last question is 'MR' or 'M', it sounds like we could just have `-diff-filter=AM`.

hageboeck: I tested, and the status is exclusive. Something that's designated C or R is not M, even if it has been modified. That should mean that we should also run on R.

As for C, copy seems to be almost unreachable, because it only triggers if:
- You pass `diff -C` (detect copies) and
  - The original is modified in the same changeset
  - Or you pass `--find-copies-harder`, which scans the entire repo

pcanal: Fair enough.  Thanks.

PR #19198, opened 2025-06-26
[ntuple] Merger: fix bugs in GenerateZeroPagesForColumns

silverweed:
- nElements was used to grow each zero page instead of nElementsPerPage
- only the last zero page was added to the page groups

This bug manifested as a crash when Union-merging big-enough RNTuples, so that GenerateZeroPagesForColumns actually generates more than 1 page.

This fix will be backported to 6.36.

cc @amete

## Checklist:

- [x] tested changes locally


hahnjo (approved):
Thanks!

PR #19201, opened 2025-06-26
Deprecate template Snapshot overload

vepadulano:
Based on https://github.com/root-project/root/pull/19193 https://github.com/root-project/root/pull/19182 https://github.com/root-project/root/pull/19192


pcanal:
Is this ready for review? The commit starts with `WIP`.

vepadulano:
Thank you @pcanal , the changes are indeed ready for review, the commit message has been updated accordingly.

tree/dataframe/test/dataframe_utils.cxx
```diff
@@ -24,10 +25,9 @@ TEST(RDataFrameUtils, DeduceAllPODsFromDefines)
                .Define("Long64_t_tmp", []() { return Long64_t(0ll); })
                .Define("ULong64_t_tmp", []() { return ULong64_t(0ull); })
                .Define("bool_tmp", []() { return bool(false); });
-   auto c = d.Snapshot<char, unsigned char, int, unsigned int, short, unsigned short, double, float, Long64_t,
-                       ULong64_t, bool>("t", "dataframe_interfaceAndUtils_1.root",
-                                        {"char_tmp", "uchar_tmp", "int_tmp", "uint_tmp", "short_tmp", "ushort_tmp",
-                                         "double_tmp", "float_tmp", "Long64_t_tmp", "ULong64_t_tmp", "bool_tmp"});
+   auto c = d.Snapshot("t", "dataframe_interfaceAndUtils_1.root",
```
pcanal, line 29: ❤️

pcanal (approved):
LGTM.   Ideally the move of the function from the headers to the source would have been done in a separate commit (as is, it is hard to tell if anything changed in those functions).

tree/dataframe/test/dataframe_utils.cxx
hageboeck, line 29: Not related to this PR, but does it make sense to remove the file, or should we just not do it in general and trust that test files are removed periodically?

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -1301,11 +1298,13 @@ public:
    /// df.Snapshot("outputTree", "outputFile.root", {"x"}, opts);
    /// ~~~
    template <typename... ColumnTypes>
-   RResultPtr<RInterface<RLoopManager>>
-   Snapshot(std::string_view treename, std::string_view filename, const ColumnNames_t &columnList,
-            const RSnapshotOptions &options = RSnapshotOptions())
+   R__DEPRECATED(
```
hageboeck, line 1303: Can you confirm that the following problem can be solved for any user?

One is on 6.36, the other on 6.40, and they share an RDF script. They can both run (6.36 is jitting, 6.40 is calling the new interface), and neither runs into issues and warnings, other than time to JIT.

hageboeck: One could add to doxygen:
```c++
/// \deprecated Use the overloads without templates.
```

tree/dataframe/inc/ROOT/RDF/SnapshotHelpers.hxx
hageboeck, line 1: It is impressive how much code leaves the headers of RDF! 🚀

tree/dataframe/inc/ROOT/RDF/SnapshotHelpers.hxx
```diff
@@ -942,8 +92,8 @@ public:
 
    UntypedSnapshotRNTupleHelper(const UntypedSnapshotRNTupleHelper &) = delete;
    UntypedSnapshotRNTupleHelper &operator=(const UntypedSnapshotRNTupleHelper &) = delete;
-   UntypedSnapshotRNTupleHelper(UntypedSnapshotRNTupleHelper &&) = default;
-   UntypedSnapshotRNTupleHelper &operator=(UntypedSnapshotRNTupleHelper &&) = default;
+   UntypedSnapshotRNTupleHelper(UntypedSnapshotRNTupleHelper &&);
+   UntypedSnapshotRNTupleHelper &operator=(UntypedSnapshotRNTupleHelper &&);
```
hageboeck: This breaks nothrow_move_constructible. You can either `=default` in the header, or you have to declare them `noexcept` in header and default them in the cxx.
Otherwise, when helpers would be pushed in e.g. a vector, the compiler is forced to copy them when the vector grows.

Godbolt for playing:
https://godbolt.org/z/jPcorfGov

For performance, defaulting them in the headers would be best, but I guess the outlining happens because we have a forward-declared member that we don't want to include here, correct?

tree/dataframe/inc/ROOT/RDF/SnapshotHelpers.hxx
```diff
@@ -993,8 +143,8 @@ public:
 
    UntypedSnapshotTTreeHelper(const UntypedSnapshotTTreeHelper &) = delete;
    UntypedSnapshotTTreeHelper &operator=(const UntypedSnapshotTTreeHelper &) = delete;
-   UntypedSnapshotTTreeHelper(UntypedSnapshotTTreeHelper &&) = default;
-   UntypedSnapshotTTreeHelper &operator=(UntypedSnapshotTTreeHelper &&) = default;
+   UntypedSnapshotTTreeHelper(UntypedSnapshotTTreeHelper &&);
+   UntypedSnapshotTTreeHelper &operator=(UntypedSnapshotTTreeHelper &&);
```
hageboeck: Same. Either default here, or declare noexcept.

tree/dataframe/inc/ROOT/RDF/SnapshotHelpers.hxx
```diff
@@ -1064,8 +214,8 @@ public:
 
    UntypedSnapshotTTreeHelperMT(const UntypedSnapshotTTreeHelperMT &) = delete;
    UntypedSnapshotTTreeHelperMT &operator=(const UntypedSnapshotTTreeHelperMT &) = delete;
-   UntypedSnapshotTTreeHelperMT(UntypedSnapshotTTreeHelperMT &&) = default;
-   UntypedSnapshotTTreeHelperMT &operator=(UntypedSnapshotTTreeHelperMT &&) = default;
+   UntypedSnapshotTTreeHelperMT(UntypedSnapshotTTreeHelperMT &&);
+   UntypedSnapshotTTreeHelperMT &operator=(UntypedSnapshotTTreeHelperMT &&);
```
hageboeck: Noexcept.

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -17,12 +17,22 @@
...
+#include <TFile.h>
+#include <TLeaf.h>
+#include <TTreeReader.h>
+
+#include <ROOT/REntry.hxx>
+#include <ROOT/RNTuple.hxx>
+#include <ROOT/RNTupleDS.hxx>
+#include <ROOT/RNTupleWriter.hxx>
+#include <ROOT/RTTreeDS.hxx>
+#include <ROOT/TBufferMerger.hxx>
```
hageboeck: It could be slightly better to put the "more modern" headers first, the old ones last. In this way, you would see if the "modern" headers forgot to include an old class, but that's really nitpicking ...

tree/dataframe/inc/ROOT/RDF/SnapshotHelpers.hxx
vepadulano: > For performance, defaulting them in the headers would be best, but I guess the outlining happens because we have a forward-declared member that we don't want to include here, correct?

Correct! And I believe we want to keep that behaviour. I will try your suggestion of adding `noexcept`, thanks!

tree/dataframe/test/dataframe_utils.cxx
vepadulano: It makes a lot of sense. Many of the older tests do not ensure proper file cleanup and that should be improved. For other PRs

vepadulano:
Thank you all for the reviews! I have split the changes in multiple commits to make the history clearer

hageboeck:
@vepadulano I forgot one thing: This very much deserves to be in the release notes for 6.38!

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -303,8 +303,6 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree, ROOT::Internal::RDF:
 }
 } // namespace
 
-namespace ROOT::Internal::RDF {
```
hageboeck: Before this commit, the ROOT build is likely broken, so we would not be able to bisect over this. The function is both qualified and in the namespace, so this commit should be squashed into the commit where the function is moved in its current location, shouldn't it?

vepadulano: You can build the repo at any point of the commit history of this branch without breaking :+1:

PR #19202, opened 2025-06-26
[cppyy] Remove unused template instance pattern rule

### LINKED ISSUE #19189: Unused class rule in dictionary generation
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

I always get the following warning on my machine:
```
Warning: Unused class rule: TypeReduction::*
```
Coming from https://github.com/root-project/root/blob/acde396795dd66f7e1cfeadccf10956cfdf33343/bindings/pyroot/cppyy/cppyy/test/templates.h#L511 paired with https://github.com/root-project/root/blob/acde396795dd66f7e1cfeadccf10956cfdf33343/bindings/pyroot/cppyy/cppyy/test/templates.xml#L38-L40

Removing the line https://github.com/root-project/root/blob/acde396795dd66f7e1cfeadccf10956cfdf33343/bindings/pyroot/cppyy/cppyy/test/templates.xml#L39 also removes the warning from my build. But I don't understand why. Also note that this is not related in anyway with Python/cppyy. In fact, I have created a standalone reproducer, see the following.

It would be nice to understand why this warning appears and whether it is actually safe to remove that line from the xml.

FYI @aaronj0 

### Reproducer

https://github.com/vepadulano/unused-class-rule-reproducer

### ROOT version

master

### Installation method

Any

### Operating system

Linux

### Additional context

_No response_

dpiparo:
as there are no instances to match.

This PR fixes #19189

### DIFF
--- a/bindings/pyroot/cppyy/cppyy/test/templates.xml
+++ b/bindings/pyroot/cppyy/cppyy/test/templates.xml
@@ -36,7 +36,6 @@
   <class pattern="T_WithGreedyOverloads::*" />
 
   <namespace name="TypeReduction" />
-  <class pattern="TypeReduction::*" />
   <function pattern="TypeReduction::*" />
 
   <namespace name="FailedTypeDeducer" />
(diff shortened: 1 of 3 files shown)


hahnjo:
Can this be merged to remove the warning during the build?

hahnjo:
ping @dpiparo. Do we have to wait for upstream or can we get rid of the (annoying, but harmless) warning?

guitargeek (approved):
Since the discussion in the PR to Wims repo didn't conclude yet, I'd suggest to merge this PR to ROOT for now to fix the warning, but with an added patch file to keep track of the differences. I have added this to the PR here. Thanks again @dpiparo for the initiative!

PR #19203, opened 2025-06-26
[Hist] Generalize TF1 constructor using C++17

guitargeek:
Before, the requirement was that the wrapped function had a call operator that *exactly* took C-style arrays as arguments.

To be more general, we check if the function can be invoked with C-style arrays as arguments, even if implicit conversions are involved. This is done using `std::is_invocable_r_v` from C++17.

Care was taken that the detection of vectorized functions still works, which can be tested like this (requires `veccore` and `vc`):

```c++
// Modified reproducer from ROOT-10251
auto originalFF = [](ROOT::Double_v *x , double *par){ (void)par; return *x; };
auto boundFF = std::bind (originalFF,std::placeholders::_1, std::placeholders::_2);
TF1 myFitFunct("myFitFunct", boundFF,0, 1 , 0 );

myFitFunct.Print("v");

std::cout << myFitFunct.IsVectorized() << std::endl;
```

Closes https://its.cern.ch/jira/browse/ROOT-10251.

### DIFF
--- a/hist/hist/inc/TF1.h
+++ b/hist/hist/inc/TF1.h
@@ -175,58 +175,6 @@ namespace ROOT {
       struct TF1Builder<Func *> {
          static void Build(TF1 *f, Func *func);
       };
-
-      /// %Internal class used by TF1 for obtaining the type from a functor
-      /// out of the set of valid operator() signatures.
-      template<typename T>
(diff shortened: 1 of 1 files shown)


atolosadelgado:
Thank you @guitargeek !!

PR #15160, opened 2024-04-05
[docu] Add new ROOT build option: documentation-building

ferdymercury:
# This Pull request:

## Changes or fixes:

This allows to build the documentation as part of the normal ROOT build and sets the basis for integrating it into the GitHub CI. And to test new PullRequests so that they do not break the existing documentation nor do add new warnings to it.

You can still build the documentation as an external project, like is done now.

## Status

Draft, still work to do on the notebook / tutorial part.

## History

This supersedes the old PR https://github.com/root-project/root/pull/9966 that I just closed.

The goals were:

- [x] Migrate old Makefile to modern CMake. See https://github.com/root-project/root/issues/9090
- [x] Allow building the documentation from a git source repository with read-only permissions. I.e. try to not pollute the sources within the process. This follows the Make to CMake migration philosophy. See https://github.com/root-project/root/issues/8947
- [x] As done by [ALICE](https://github.com/AliceO2Group/Monitoring/blob/dev/doc/CMakeLists.txt#L22), use a 'dynamic' Doxyfile declaration, that only specifies what needs to be changed with respect to the default one. This makes easier the maintenance, as you do not need to constantly update the Doxyfile when a new doxygen version arises, and prevents warnings when running in older doxygen versions. This will hopefully contribute to next point:
- [ ]  https://github.com/root-project/root/issues/9953 in combination with https://github.com/ammaraskar/gcc-problem-matcher
- [ ] Allow in the future for automatic meta-documentation of the CMake flag system. See https://github.com/root-project/root/issues/8999
- [x] Potentially add a flag in the main ROOT CMakeLists.txt, to enable the building of the documentation via a normal "add_subdirectory()", so that one does not need to build as a separate process.
- [x] Make the doxygen documentation thread-safe
- [x] Don't let doxygen search for "input" files in the output directory (e.g. it looks now for Mathjax output files that one just "generated")

## Checklist:

- [x] tested changes locally
- [ ] We would need to update https://github.com/root-project/web/blob/main/for_developers/doxygen/index.md
- [ ] Update rootspi scripts to build docu in CI accordingly
- [ ] tutorial generation does not work yet as "out of source". To-do...
- [x] require min doxygen version 1.9.4
- [ ] pre-check that output directory is empty ?
- [ ] if GVIZ not found, disable dot
- [ ] Stamp file generation to only recompile when changes
- [x] Remove unneeded unnecessary files Images.dat etc ? (or prevent that they appear in root source)
[...]

PR #19211, opened 2025-06-27
[CMake] Make it possible to not create `thisroot` setup scripts

guitargeek:
The setup scripts like `thisroot.sh` and friends are disliked by packagers because they mutate the environemnt variables, which are carefully controlled by many package managers.

Here some examples of package definitions that explicitly make an effort to remove these scripts:

  * Arch Linux: https://gitlab.archlinux.org/archlinux/packaging/packages/root/-/blob/main/PKGBUILD?ref_type=heads#L281

  * Conda: https://github.com/conda-forge/root-feedstock/blob/3dfe4c8caf3b9fb8715ed498bc2072ef492ab2d1/recipe/build_root.sh#L490

  * Nix will follow soon now that this issue is fixed: https://github.com/NixOS/nixpkgs/issues/241454

If we make `pip install root` work, we probably also don't need them.

And when using ROOT from LCG or `cmsenv`, one doesn't need these scripts either.

So very often, the setup scripts are not needed or should even not be used, and having them present ist just confusing to the user. This commit therefore suggests a new `thisroot_scripts` built option that can be set to `OFF` if you don't need these scripts.

By the way the scripts are also a bit fragile to the build configuration. For example as soon as one manually specifies `CMAKE_INSTALL_PYTHONDIR` to be different from the default, they don't work anymore because they assume the Python install directory will always be the library install directory.


ferdymercury:
Nice!
Unrelated comment: I also managed to build the doxygen documentation without calling thisroot.sh and without polluting the source folder, as part of the ROOT CMake build or as standalone, see https://github.com/root-project/root/pull/15160

bellenot (approved):

PR #19213, opened 2025-06-27
[skip-ci] Create `histv7`, document terminology and design

hahnjo:


hist/histv7/doc/Terminology.md
```diff
@@ -0,0 +1,53 @@
...
+A *bin content* is the value of a single bin.
+The *bin content type* can be an integer type, a floating-point type, the special `DoubleBinWithError`, or a user-defined type.
+
+A *bin error* is the Poisson error of a bin content.
+With the special `DoubleBinWithError`, it is the square root of the sum of weights squared: $\sqrt{\sum w_i^2}$
+Otherwise it is the square root of the bin content, which is only correct with unweighted filling.
+
+A *bin index* (plural *indices*) refers to a single bin of a dimension.
+A *normal bin* is inside an axis and its index starts from 0.
+*Underflow* and *overflow* bins, also called *flow bins*, are outside the axis and their index has a special value.
```
hageboeck, line 24: Could "flow bins" be an unfortunate abbreviation? I never heard it outside UHI discussions and/or the ROOT team. I find it imprecise, because integers and floating-point numbers under- or overflow, but they don't "flow" (other than through neural networks and the likes). Along the same lines, a value can under- and overflow an axis range, but it doesn't flow around the axis.

hist/histv7/doc/DesignImplementation.md
```diff
@@ -0,0 +1,73 @@
...
+```cpp
+template <typename... A> void Fill(const A &...args) {
+   Fill(std::forward_as_tuple(args...));
+}
+```
+This will forward the arguments as references, so no copy-constructors are called (that could potentially be expensive).
+
+### Arguments with Same Type
+
+In this case, the function has a `std::size N` template argument and accepts a `std::array`.
```
hageboeck:
```suggestion
In this case, the function has a `std::size_t N` template argument and accepts a `std::array`.
```

hageboeck (commented):
Hello, LGTM!

Of the comments I added, the "flow bins" are my biggest concern.

hist/histv7/doc/DesignImplementation.md
```diff
@@ -0,0 +1,73 @@
+# Design and Implementation
```
jblomer, line 1: Maybe "Interface Design"? Unless that will get extended by more aspects.

hist/histv7/doc/Terminology.md
```diff
@@ -0,0 +1,53 @@
...
+A *regular axis* has equidistant bins in the interval $[a, b)$.
+A *variable bin axis* is configured with explicit bin edges $[e_{n}, e_{n+1})$.
+A *categorical axis* has a unique label per bin.
+*Axes* is the plural of axis and usually means the bin configurations for all dimensions of a histogram.
+
+A *bin content* is the value of a single bin.
+The *bin content type* can be an integer type, a floating-point type, the special `DoubleBinWithError`, or a user-defined type.
+
+A *bin error* is the Poisson error of a bin content.
+With the special `DoubleBinWithError`, it is the square root of the sum of weights squared: $\sqrt{\sum w_i^2}$
```
jblomer: For later: can we prevent weighted filling with the wrong bin content type (double instead of `DoubleBinWithError`?

hist/histv7/doc/Terminology.md
```diff
@@ -0,0 +1,53 @@
...
+It is supposed to be exhaustive; any missing term should be added when needed.
+
+An *axis* is a bin configuration in one dimension.
+A *regular axis* has equidistant bins in the interval $[a, b)$.
+A *variable bin axis* is configured with explicit bin edges $[e_{n}, e_{n+1})$.
+A *categorical axis* has a unique label per bin.
+*Axes* is the plural of axis and usually means the bin configurations for all dimensions of a histogram.
+
+A *bin content* is the value of a single bin.
+The *bin content type* can be an integer type, a floating-point type, the special `DoubleBinWithError`, or a user-defined type.
```
jblomer: Maybe `RDoubleBinWithError`?

hist/histv7/doc/Terminology.md
```diff
@@ -0,0 +1,53 @@
...
+
+A *bin error* is the Poisson error of a bin content.
+With the special `DoubleBinWithError`, it is the square root of the sum of weights squared: $\sqrt{\sum w_i^2}$
+Otherwise it is the square root of the bin content, which is only correct with unweighted filling.
+
+A *bin index* (plural *indices*) refers to a single bin of a dimension.
```
jblomer: What does this mean for multi-dimensional histograms?

hist/histv7/doc/Terminology.md
```diff
@@ -0,0 +1,53 @@
...
+*Concurrent filling* allows to modify the same histogram without (external) synchronization.
+
+A *histogram* is the combination of an axes configuration and storage of bin contents.
+For most use cases, it also includes (global) *histogram statistics*.
+On the one hand, these are the number of entries, the sum of weights, and the sum of weights squared.
+The number of *effective entries* can be computed as the ratio $$\frac{(\sum w_i)^2}{\sum w_i^2}$$.
+Furthermore, for each dimension the histogram statistics include the sum of weights times value and the sum of weights times value squared.
+This allows to compute the arithmetic mean and the standard deviation of the values before binning.
+
+A *linear index* starts from 0 up to the total number of bins, potentially including flow bins.
```
jblomer: Possibly: physical index or linearized index

hist/histv7/doc/Terminology.md
```diff
@@ -0,0 +1,53 @@
...
+For a single axis, it places the flow bins after the normal bins.
+The *global index* is a combination of the linear indices from all axes.
```
jblomer: Perhaps give more details here on the flattening.

jblomer (approved):
Great! Very good starting point.

hist/histv7/doc/Terminology.md
hahnjo: Yes, in the prototype I have `static_assert`s failing when trying to use weighted filling on integer bin content types. So far I've opted to allow it for floating-point types because that may be interesting when only looking at the bin content, but not the error. It's possible to reconsider at a later stage.

hist/histv7/doc/Terminology.md
siliataider: Outside of UHI, it is the terminology employed by boost-histogram as well

siliataider (approved):
I think this is a great start to unify the terminology, thanks!

hist/histv7/doc/Terminology.md
hageboeck: OK, let me demonstrate my point a bit better 😄
https://www.google.com/search?q=flow-bin
image

https://www.google.com/search?q=underflow-bin
image

siliataider: I see your point: 'flow bins' isn't widely used in the histogramming world _yet_.
But it appears extensively in boost-histogram documentation (e.g., [views](https://boost-histogram.readthedocs.io/en/latest/user-guide/histogram.html#views)) and in UHI documentation (e.g., [slicing](https://uhi.readthedocs.io/en/latest/indexing.html#slicing)). I think it's gaining traction, and ROOT histograms could benefit from adopting it. 
I personally find '**flow bins**' short and clear, but '**under/overflow bins**' or '**out-of-range bins**' work as well I guess 🤷🏻‍♀️

hahnjo: I agree on your point, Stephan, I had also not really heard the term before Silia mentioned it. I do think though that we can still establish it now, because there is no clear and concise term yet. Thinking a bit during the weekend, a potential alternative would be "excess bins" but that has the disadvantage that it's completely new...

hist/histv7/doc/DesignImplementation.md
hahnjo: It already has some aspects that go beyond interface in the "Miscellaneous" section. The way I view this document are the key choices that lead to the histogram package. For example, one aspect still missing is the goal of being able to support concurrent filling without degrading sequential filling, which we get by storing the bin contents as-is and only using atomic instructions on top. I plan to have the "Code Architecture" (ie how it's organized into classes) in a separate document.

What about "Package Design" maybe, in the sense of "software architecture design"?

hist/histv7/doc/Terminology.md
hahnjo: I like linearized :+1:

hist/histv7/doc/Terminology.md
hahnjo: I'm not sure, in the end I would consider this an implementation detail. I plan to have API functions to compute the global index, but in principle I don't want users to do this themselves...

hist/histv7/doc/Terminology.md
hahnjo: For a multi-dimensional histogram, you'll need an array of bin indices to refer to a single bin content. It's indeed ambiguous how we define a "bin": Is it per dimension or is it only once you give coordinates for all dimensions, ie what really has the bin content in the end?

hageboeck: The ambiguity theoretically exists also for the TH1-style histograms, but in practice, it is always clear from the context if you are talking about a bin on an axis (1-dim), or a bin in the histogram (possibly an index tuple). I guess one could add a sentence about bins in histogram and cover all cases.

hahnjo: Yes, that's why there is no definition of a "bin" itself :innocent: for "bin content" it is clear that it's the bin of a histogram, and for a single "bin index" it makes life easier if that's on a single axis.

Could you concretely propose where you would like to see added what?

hageboeck: I didn't think it was necessary to expand on it initially, but I reacted to Jakob's question. If one wants to resolve the ambiguity (which I think must people naturally understand already), one could say:
```suggestion
A *bin index* (plural *indices*) refers to a single bin of a dimension, a tuple of indices refers to a bin in a histogram.
```

hist/histv7/doc/DesignImplementation.md
jblomer: I think that's fine for now, we'll see later how the documentation bits develop.

hist/histv7/doc/Terminology.md
hahnjo: Ok thanks; I've included this (replacing tuple by array because that will be the type accepted by `GetBinContent`). We may need to revise at a later point to stress that the array must match the dimensionality of the histogram.

hist/histv7/doc/Terminology.md
hahnjo: @hageboeck does this make sense for the moment?

hageboeck: Personally, I would invest the few extra characters to write under/overflow to stick with a term that's already widely adopted.

hahnjo: Based on offline discussion, we will leave the term in for now but I'll try to use it only in contexts where it's unambiguous. If that's still confusing to users, we can re-evaluate and potentially throw it out again.

PR #19218, opened 2025-06-27
[CMake] Set `CMAKE_OSX_SYSROOT` only if CMake generator is Xcode

guitargeek:
Set `CMAKE_OSX_SYSROOT` only if CMake generator is Xcode, because otherwise we don't need it. Also, if not using Xcode ,the `xcrun` executable that is used to figure out this variable is also not available, leading to configuration failures when building ROOT on macOS without Xcode installed.

Follows up on 6bd0dba and ebc38bd.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ if(WIN32)
   cmake_policy(SET CMP0091 OLD)
   set(CMAKE_SKIP_TEST_ALL_DEPENDENCY TRUE)
 endif()
-if(APPLE AND (NOT CMAKE_OSX_SYSROOT OR CMAKE_OSX_SYSROOT STREQUAL ""))
+if((CMAKE_GENERATOR MATCHES Xcode) AND (NOT CMAKE_OSX_SYSROOT OR CMAKE_OSX_SYSROOT STREQUAL ""))
   # The SYSROOT *must* be set before the project() call
   execute_process(COMMAND xcrun --sdk macosx --show-sdk-path
     OUTPUT_VARIABLE SDK_PATH


CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ if(WIN32)
   cmake_policy(SET CMP0091 OLD)
   set(CMAKE_SKIP_TEST_ALL_DEPENDENCY TRUE)
 endif()
-if(APPLE AND (NOT CMAKE_OSX_SYSROOT OR CMAKE_OSX_SYSROOT STREQUAL ""))
+if((CMAKE_GENERATOR MATCHES Xcode) AND (NOT CMAKE_OSX_SYSROOT OR CMAKE_OSX_SYSROOT STREQUAL ""))
```
pcanal: Is it really dependent of the Generator or is it the compiler (XCode's clang)?  (eg I use Ninja+XCode's clang).

guitargeek: I don't know, but you have the perfect setup to figure this out then! Does this PR build for you? If not, it's the compiler and I need to put a check for Apple Clang there. Can you give a hand by doing this check?

pcanal: will do ...

pcanal: So form a fresh directory, with your commit, it seems to work ... but then again it is still 'set':
```
mac-142865:osxroot pcanal$ rg -C3 CMAKE_OSX_SYSROOT CMakeCache.txt 
475-
476-//The product will be built against the headers and libraries located
477-// inside the indicated SDK.
478:CMAKE_OSX_SYSROOT:PATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.5.sdk
479-
```

guitargeek: Hi @pcanal, are you using CMake < 4.0? In that case, according to the docs, this variable is set by CMake:

https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html

But I think you are right: the generator should not matter here, and it's more about the AppleClang compiler. I'll update the PR to check for AppleClang instead of `(CMAKE_GENERATOR MATCHES Xcode)`.

guitargeek: Oh, actually that doesn't work, because `CMAKE_GENERATOR` needs to be set *before* enabling any language according to the same docs. So I can't change to that.

Maybe you upgraded to CMake 4.0 by now? I would need help here by someone checking if with this PR, CMake 4.0, Ninja and AppleClang, ROOT still compiles and if `CMAKE_OSX_SYSROOT` ends up in the CMakeCache.txt.

pcanal: I indeed only have cmake version 3.30.1

guitargeek: Ok I think I have figured out the right thing to do now by looking at the CMake source code itself (see the updated PR description).

Are you fine with this? IMHO there is no need to test more now if the CI is green, because given that this is what CMake did itself I'm pretty confident that it's correct. And it also works for me when building ROOT for nix Darwin.

pcanal:
> This mimics what was done in CMake before version 4.0:

if/since it is no longer done in 4.0, what was is replaced by or why do they not need it anymore (and we do)?

CMakeLists.txt
pcanal: Yes, I am fine with this ... however I can't help to ponder if CMake found a 'better' solution since they no longer use it (seemingly)

pcanal (approved):
This looks good to me (but see inline discussion).

guitargeek:
> if/since it is no longer done in 4.0, what was is replaced by or why do they not need it anymore (and we do)?

I'm trying to understand the PR https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9977

But the more I look at at it, the less I understand why 6bd0dbad38b was needed to begin with, because even in new CMake there is the `xcrun` way of determining `CMAKE_OSX_SYSROOT`` if it is not set:
https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Darwin-Initialize.cmake?ref_type=heads#L283

Well, I don't intend to understand everything :slightly_smiling_face: I was only looking for the minimal changes to make ROOT compile again in `nixpkgs` without patches for Darwin.

CMakeLists.txt
guitargeek: That's true! If they didn't think they need it, why do we? In the future we should try again to remove our `CMAKE_OSX_SYSROOT` detection and see what actually goes wrong then, and maybe find the proper fix. But someone else would need to try. Even though I have some mac minis, I refuse to create an Apple account so I can't use the developer tools :laughing: Why do you have to register somewhere to use your computer....

pcanal: Apparently it is still needed by 'just' for the case when using XCode's compiler ... oh well the change here might be good enough (the risk is that the SDK found conflicts with the one from the compiler when it comes from homebrew for example ... )

PR #19219, opened 2025-06-27
[roottest] use cmake for `io/perf` tests

linev:
1. In `root/io/perf` directory run most macros - not only `julius.C`. 
2. In ReadBuffer.C/WriteBuffer.C exclude tests with direct use of TStreamerInfo methods - which are now private
3. `ShortTest`/`HardTest` takes too long and creating huge files - therefore put under `if(very_long_tests)` 
4. Replicate Makefile logic in `root/io/perf/slowreading` - first create marker file and then run all combinations checking that time is not much differ
5. Enable `root/io/perf/stl` tests with TLorenzVector - were not running with Makefile
6. Enable `root/io/perf/simple` tests - were not running with Makefile
7. Convert `root/io/perf/userdatasets`, all tests are failing and taking extremely long time, therefore use `if(very_long_tests)`


roottest/root/io/perf/ReadBuffer.C
```diff
@@ -32,135 +32,52 @@ public:
...
-#else
-   char *pointer = (char*)&obj;
-#endif
-   TClass *cl = gROOT->GetClass(typeid(obj));
-   TStreamerInfo *info = cl->GetStreamerInfo();
-   
-#if ROOT_VERSION_CODE<= 199169
-      info->WriteBuffer(b, (char*)(&obj),-1);
-#else
-      info->WriteBufferAux(b, &pointer,-1, 1, 0, 0);
```
pcanal, line 118: This function is still public ...

roottest/root/io/perf/slowreading/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+                      MACRO Read.C
+                      MACROARG "\"NuEvent_60_cxx\", \"DST_60.root\", 1"
+                      FIXTURES_REQUIRED root-io-perf-slowreading-files-fixture
+                                        root-io-perf-slowreading-60-fixture
+                      FIXTURES_SETUP root-io-perf-slowreading-mark-fixture)
+  endif()
+
+  foreach(file 60 61 62)
+    ROOTTEST_ADD_TEST(check-macro${macro}-file${file}
+                      MACRO Read.C
```
pcanal, line 27: In the Makefile version (via the trytwice), it was allowing for one of two run to be too slow before failing ...

roottest/root/io/perf/ReadBuffer.C
linev: Yes, but one of the arguments `TCompInfo` is not accessible. 
And I am not sure if I can put nullptr instead.
So I did not find reasonable set of arguments to let it run.
Also there is no public `ReadBufferAux` to make sense test of writing.

roottest/root/io/perf/slowreading/CMakeLists.txt
linev: But why twice is better than once?
Files must be in cache.
If test will fail too often - one can increase margins.

pcanal: Yes this is the other option.  However the net effect is different.  If one increase the margin, it allows for more performance degradation before noticing.   So twice was thought at the time to be better that once because it allowed to ignore more fluctuation (including unexpected disk and/or cache usage due to other usage at the time of the run) while still keeping a lower margin.

For now, let's at least add a comment mentioning that a previous incarnation of the test(s) was running the test a second time if the first try was slower than expected.

roottest/root/io/perf/ReadBuffer.C
pcanal: > Also there is no public ReadBufferAux to make sense test of writing.

I don't remember why but the converse of `WriteBufferAux` is simply `ReadBuffer`.

> Yes, but one of the arguments TCompInfo is not accessible.

Indeed :( ... It looks like 4e9c5a926b9067487f088b5cd80c4e2c3d2ba244 should have made `WriteBufferAux` private.  That commit is part of a large effort to migrate toward using StreamerInfoActions ...  Those performance test were part of making the decision to select the StreamerInfoAction solution ... so indeed they might be obsolete.

roottest/root/io/perf/slowreading/CMakeLists.txt
linev: It is not clear to me how to repeat test twice without failing when first time was too slow.
In principle, one can do it inside macro - if we really want it.

I will add comments

pcanal: > It is not clear to me how to repeat test twice without failing when first time was too slow.

[Just a comment, no action needed] a simple way would have been to keep using the `trytwice.sh` script :)

linev: Any usage of shell scripts excludes Windows - that we do not want

pcanal: Fair enough if we ever need it we would have to come up with a solution/choice for Windows.

PR #19221, opened 2025-06-27
Align behaviour of SetBranchStatus of TChain and TTree

vepadulano:
When a user calls SetBranchStatus("*", false) it means that all branches of the dataset should be deactivated, no matter what other method is called, except if then the user calls SetBranchStatus(someBranch, true) to activate a branch.

So, if a user calls SetBranchStatus to deactivate all branches and then calls SetBranchAddress, this should have no effect. No data should be read because the branch related to SetBranchAddress should not have been activated.

Before this commit, TTree was respecting this rule, but TChain was not. This commit ensures that TChain only sets the status of the branch of the current TTree if the corresponding TChainElement was explicitly provided with a status by the user.


tree/tree/test/TChainRegressions.cxx
```diff
@@ -128,31 +128,109 @@ TEST(TChain, UncommonFileExtension)
...
+   TChain chain(treename);
+   for (auto i = 0; i < nFiles; ++i) {
+      chain.Add(filename);
+   }
+   auto nEntries = callGetEntries ? chain.GetEntries() : chain.GetEntriesFast();
+   if (deactivateAllBranches)
+      chain.SetBranchStatus("*", 0);
+   // read a single branch
+   float random = 0.333333;
+   TBranch *b_random;
```
pcanal:
```suggestion
   TBranch *b_random = nullptr;
```

tree/tree/src/TChain.cxx
```diff
@@ -1446,7 +1446,10 @@ Long64_t TChain::LoadTree(Long64_t entry)
             TIter fnext(fStatus);
             while ((frelement = (TChainElement*) fnext())) {
                Int_t status = frelement->GetStatus();
-               fTree->SetBranchStatus(frelement->GetName(), status);
+               // Only set the branch status if it has a value provided
+               // by the user
+               if (status != -1)
+                  fTree->SetBranchStatus(frelement->GetName(), status);
```
pcanal, line 1452: This is a notable change, let's explain it explicitly the release notes.

PR #19222, opened 2025-06-27
[Python] Avoid hack in setting `__reduce__` attribute of CPPInstance

guitargeek:
The CPPIntance type is immutable, and new attributes can't be set:
```txt
import cppyy
>>> setattr(cppyy._backend.CPPInstance, "test", 10)
  File "<python-input-3>", line 1, in <module>
    setattr(cppyy._backend.CPPInstance, "test", 10)
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: cannot set 'test' attribute of immutable type 'cppyy.CPPInstance'
>>>
```

In our pythonizations, we still want to monkey-patch a `__reduce__` method anyway, and to make Python not complain we to a hack by setting the attribute with the C Python API using `PyObject_GenericSetAttr`. This function doesn't normally do any safety checks, and our monkey patching works.

However, the Python debug build is not having any of that, thanks to a new assert that was added a year ago:

https://github.com/python/cpython/blame/c419af9e277bea7dd78f4defefc752fe93b0b8ec/Objects/object.c#L1921

To make the ROOT Python interface work with debug builds of the Python interpreter, we therefore have to implement adding the reduce method properly.

This commit suggests to achieve this by defining the `__reduce__` method in the immutable CPPInstance type within CPyCppyy, but in such a way that its implementation can be routed to ROOT via a private API.

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/CPPInstance.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/CPPInstance.cxx
@@ -400,6 +400,21 @@ static PySequenceMethods op_as_sequence = {
     0,                             // sq_inplace_repeat
 };
 
+std::function<PyObject *(PyObject *)> &CPPInstance::ReduceMethod() {
+   static std::function<PyObject *(PyObject *)> reducer;
+   return reducer;
+}
+
+PyObject *op_reduce(PyObject *self, PyObject * /*args*/)
+{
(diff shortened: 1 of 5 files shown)


aaronj0 (approved):
LGTM! Thanks for this nice improvement

Issue #9252, opened 2021-11-05
`TTree::CloneTree` is incorrect in case of schema evolution and slow cloning


pcanal:
The problem is that `CloneTree` also copies the class version number of the content of the `TBranch` onto the new branch.

In the case of fast cloning (content of the file is forwarded to the new file as is), this is the correct and necessary behavior.

In the case of slow cloning (copy entry by entry, i.e. your case), this is the wrong thing (it uses the old StreamerInfo to write instead of the new one).

There is some ambiguity on whether the result of CloneTree will be use for fast cloning or for slow cloning and couple of option on how to implement the proper behavior.  In other words, it will take some thoughts on how to implement this.

In the mean-time there is an (annoying but workable) work around.  For each branch whose type underwent schema evolution you can run:

```
   const char *branchname = "MCEventHeader.";
   TBranchElement *be = dynamic_cast<TBranchElement*>(ctree->GetBranch(branchname));
   if (be) {
     auto ind = ctree->GetListOfBranches()->IndexOf(be);
     TBranch *newbr = ctree->Branch(be->GetName(), be->GetClassName(), nullptr);
     ctree->GetListOfBranches()->Remove(newbr);
     ctree->GetListOfBranches()->AddAt(newbr, ind);
     delete be;
     ctree->GetListOfLeaves()->Compress(); // Work around missing protection in CopyAddresses

     tree->CopyAddresses(ctree);
   }
```

If you have more than one branch to adjust, the Compress and CopyAddress can probably be run only once.

(A similar issues could easily be present in`RNtuple`)

pcanal:
[reproducer_9252.tar.gz](https://github.com/root-project/root/files/7488905/reproducer_9252.tar.gz)

Adding a simple reproducer of the failing case. See README in the tar file for instruction.

ferdymercury:
Maybe dupe of https://its.cern.ch/jira/browse/ROOT-5176

tomeichlersmith:
I found this issue after finding a hack solution for a similar case that I described in my ROOT forum post: https://root-forum.cern.ch/t/quiet-writing-failure-when-applying-schema-evolution-while-slow-cloning/64350

Instead of removing the branches that undergo schema evolution, I take the opposite approach: avoid using `CloneTree`/`CopyAddresses` and write my own that runs `Branch` on the output tree for all the input tree branches. This works well _if_ the dictionary is loaded.

Perhaps there are three possible avenues for cloning:

[...]

PR #19227, opened 2025-06-28
[Tree] Delete virtual copy and copy assignment methods in TTree

guitargeek:
The `TTree` class already deletes the copy constructor and the copy assignment, but users might still be tempted to use the virtual copy constructor or copy assignment provided by `TObject::Clone()` and `TObject::Copy()`. But these implementations also make no sense to the TTree.

To avoid user confusion, this commit suggests to override these methods in TTree such that they thow an exception and tell you what to use instead.

Prevents reports like this one in the future:

  * https://its.cern.ch/jira/browse/ROOT-9111

### DIFF
--- a/bindings/pyroot/pythonizations/test/memory.py
+++ b/bindings/pyroot/pythonizations/test/memory.py
@@ -113,8 +113,6 @@ class MemoryStlString(unittest.TestCase):
             "TGraph2D": (100,),
             "TEntryList": ("name", "title"),
             "TEventList": ("name", "title"),
-            "TTree": ("name", "title"),
-            "TNtuple": ("name", "title", "x:y:z"),
         }
         for klass, args in objs.items():
             with self.subTest(klass=klass):
--- a/roofit/roofitcore/src/RooTreeDataStore.cxx
+++ b/roofit/roofitcore/src/RooTreeDataStore.cxx
@@ -369,7 +369,7 @@ void RooTreeDataStore::loadValues(const TTree *t, const RooFormulaVar* select, c
   // We need a custom deleter, because if we don't deregister the Tree from the directory
   // of the original, it tears it down at destruction time!
   auto deleter = [](TTree* tree){tree->SetDirectory(nullptr); delete tree;};
-  std::unique_ptr<TTree, decltype(deleter)> tClone(static_cast<TTree*>(t->Clone()), deleter);
+  std::unique_ptr<TTree, decltype(deleter)> tClone{static_cast<TTree*>(gDirectory->CloneObject(t)), deleter};
   tClone->SetDirectory(t->GetDirectory());
 
   // Clone list of variables
--- a/tree/tree/src/TTree.cxx
+++ b/tree/tree/src/TTree.cxx
@@ -1059,6 +1059,17 @@ TTree::~TTree()
    }
 }
 
+TObject *TTree::Clone(const char *) const
+{
+   Error("Clone", "Not implemented, use TTree::CloneTree instead.");
+   return nullptr;
+}
+
+void TTree::Copy(TObject &) const
+{
+   Error("Copy", "Not implemented, use TTree::CopyTree or TTree::CopyEntries instead.");
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 /// Returns the transient buffer currently used by this TTree for reading/writing baskets.
 
@@ -3204,7 +3215,7 @@ TTree* TTree::CloneTree(Long64_t nentries /* = -1 */, Option_t* option /* = "" *
 
    // Note: For a chain, the returned clone will be
    //       a clone of the chain's first tree.
-   TTree* newtree = (TTree*) thistree->Clone();
+   TTree* newtree = (TTree*) thistree->TNamed::Clone();
    if (!newtree) {
       return nullptr;
    }
--- a/io/io/src/TFileMerger.cxx
+++ b/io/io/src/TFileMerger.cxx
@@ -536,7 +536,7 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type, 
                  && !cl->InheritsFrom( TDirectory::Class() )) {
          R__ASSERT(cl->IsTObject());
          TDirectory::TContext ctxt(current_sourcedir);
-         obj = obj->Clone();
+         obj = gDirectory->CloneObject(obj);
          ownobj = kTRUE;
       }
    } else if (key) {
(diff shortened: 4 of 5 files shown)


ferdymercury:
Tangentially related: https://github.com/root-project/root/issues/9252

pcanal:
Note that `TTree::Clone` is used (intentionally) internally to implement `CloneTTree`.
I.e. the CI failures (at least most of them) are related and require further adjustment to the code.

roofit/roofitcore/src/RooTreeDataStore.cxx
```diff
@@ -366,11 +366,8 @@ void RooTreeDataStore::createTree(RooStringView name, RooStringView title)
 void RooTreeDataStore::loadValues(const TTree *t, const RooFormulaVar* select, const char* /*rangeName*/, Int_t /*nStart*/, Int_t /*nStop*/)
 {
   // Make our local copy of the tree, so we can safely loop through it.
-  // We need a custom deleter, because if we don't deregister the Tree from the directory
-  // of the original, it tears it down at destruction time!
-  auto deleter = [](TTree* tree){tree->SetDirectory(nullptr); delete tree;};
-  std::unique_ptr<TTree, decltype(deleter)> tClone(static_cast<TTree*>(t->Clone()), deleter);
-  tClone->SetDirectory(t->GetDirectory());
+  std::unique_ptr<TTree> tClone{static_cast<TTree*>(const_cast<TTree*>(t)->CloneTree())};
+  tClone->SetDirectory(nullptr);
```
pcanal: Consider setting `gDirectory` to `nullptr` (via the TContext) to ensure that the `TTree` is really in memory.
```suggestion
  {
      TDirectory::TContext ctx{nullptr};
      std::unique_ptr<TTree> tClone{static_cast<TTree*>(const_cast<TTree*>(t)->CloneTree())};
  }
```
but humm ... the previous code and the new code are doing different things (not clear what the 'real' intent is or should have been).

old: Copy just the `TTree` data members themselves and attach the copy to the same file:
net effect: There is 2 concurrent ways to look at the same data in the same file.  Each copy of the TTree can have their own branch address and their own TTreeCache.  They don't affect each other (no change in address and no changing where the cursor is).

new: Copy the tree and all its data .. but in the same file (if it is writeable) well actually in whatever file is current This is very slow if the TTree is big.  The 2 trees do not affect each others but they also have duplicated (potential large) data.

proposed:  Load all the data in memory after possibly a slow copy.  This may or may not fit in memory.  (To speeds the copy up, one could consider attaching to a TMemFile).

So really the essential question before deciding which way to go is to understand what was the intent/point of the cloning of the meta data and what are the performance and memory usage limitation that this routine needs to fit in.

pcanal:
```suggestion
  std::unique_ptr<TTree> tClone{static_cast<TTree*>(t->GetDirectory()->CloneObject(t));
```
is another option ...

tree/tree/src/TTree.cxx
```diff
@@ -1060,6 +1061,17 @@ TTree::~TTree()
    }
 }
 
+TObject *TTree::Clone(const char *) const
+{
+   throw std::runtime_error("TObject::Clone() is not supported by TTree! Use TTree::CloneTree().");
```
pcanal: Minor: I think this is the first exception thrown in this file :)

roofit/roofitcore/src/RooDerivative.cxx
```diff
@@ -90,22 +90,60 @@ RooDerivative::~RooDerivative() = default;
...
-    _rd = std::make_unique<ROOT::Math::RichardsonDerivator>(wf,_eps,true);
-  }
-
-  switch (_order) {
-  case 1: return _rd->Derivative1(_x);
-  case 2: return _rd->Derivative2(_x);
-  case 3: return _rd->Derivative3(_x);
-  }
-  return 0 ;
+   if (!_ftor) {
```
pcanal, line 90: Is this seemingly unrelated commit intentionally in this PR?

guitargeek: Thanks for the review! I'll follow up later. Sorry, indeed I accidentally rebased this PR branch on another development branch. That commit will go in the next push

roofit/roofitcore/src/RooTreeDataStore.cxx
```diff
@@ -369,7 +369,7 @@ void RooTreeDataStore::loadValues(const TTree *t, const RooFormulaVar* select, c
   // We need a custom deleter, because if we don't deregister the Tree from the directory
   // of the original, it tears it down at destruction time!
   auto deleter = [](TTree* tree){tree->SetDirectory(nullptr); delete tree;};
-  std::unique_ptr<TTree, decltype(deleter)> tClone(static_cast<TTree*>(t->Clone()), deleter);
+  std::unique_ptr<TTree, decltype(deleter)> tClone{static_cast<TTree*>(t->GetDirectory()->CloneObject(t)), deleter};
```
vepadulano: What is the fundamental difference between `tree->Clone` and `tree->GetDirectory()->CloneObject(tree)` that makes the latter ok but the first not?

tree/tree/src/TTree.cxx
vepadulano: I agree with Philippe and I think it shouldn't be implemented this way, rather via

```
Error("Clone", "Not implemented, use XXXXXXX instead.");
```

tree/tree/src/TTree.cxx
```diff
@@ -1060,6 +1061,17 @@ TTree::~TTree()
...
+}
+
+void TTree::Copy(TObject &) const
+{
+   throw std::runtime_error(
```
vepadulano: same comment about the `runtime_error`

bindings/pyroot/pythonizations/test/memory.py
```diff
@@ -112,8 +111,6 @@ def test_objects_ownership_with_subdir(self):
             "TGraph2D": (100,),
             "TEntryList": ("name", "title"),
             "TEventList": ("name", "title"),
-            "TTree": ("name", "title"),
-            "TNtuple": ("name", "title", "x:y:z"),
```
vepadulano: Why are these three entries removed from the test cases?

vepadulano (changes requested):
Thank you for this work! I do not disagree with the goal of this PR, but before approval I wanted to clarify a point

From what I can see, this PR disallows calling `TTree::Clone` and `TTree::Copy`, but in practice we are still cloning TTree ourselves in our own code in two ways:

1. via `TDirectory::CloneObject`
2. by calling the base class method `TNamed::Clone`

I just want to make sure that before we go ahead we are aware of what we are doing and think wether there could be better alternatives. Do we need to clone the TTree in those occasions? If so, why don't we use `CloneTree`? At least I believe we should add some comments inplace.

PR #19234, opened 2025-06-30
[skip-ci] Add release notes for 6.36.02

silverweed:


hahnjo:
I think these are automatically created, based on projects added to issues. That's why we should have issues for every bug we fix with backports...

PR #19235, opened 2025-06-30
[metacling] Use InterOps JitCall for destructor calls

aaronj0:
This PR further offloads codegen to CppInterOps `Cpp::Destruct` facility. In line with previous developments this requires a tag bump on InterOp. This also adds back a lock that was missed in `TCling::ExecDefaultConstructor`


aaronj0:
The failures (all `roottest-python-JupyROOT`) on mac14 are unrelated to this PR and concern IPython and Jupyter

PR #19174, opened 2025-06-25
[ci] Enable Fortran on macOS

guitargeek:
Hopefully it works now with the link directories.

Thanks to the work already done by @ferdymercury and @couet in #18234!

PR #19236, opened 2025-06-30
[fortran] fix bug when copying/memmoving string

ferdymercury:
# This Pull request:

## Changes or fixes:

happens maybe only in some gfortran versions
Maybe related: https://www.tek-tips.com/threads/string-copy.1609892/post-6335242

in mac13 and mac14 CI, the last bytes of chopt were being set to garbage memory, which later led to failures due to wrong options being set. chopt was being set to "px exam"
where choptt was "px " and "exam" was the first chars of the parameter PASSCHAR("example") on the C side in h2root.cxx

mac14 still fails (with another error), but mac13 h2root works now.

mac13: there is still the error in rpath for two tests.
mac15: failures due to runners having different gfortran versions?

### DIFF
--- a/.github/workflows/root-ci-config/buildconfig/mac14.txt
+++ b/.github/workflows/root-ci-config/buildconfig/mac14.txt
@@ -1,4 +1,3 @@
-CMAKE_BUILD_RPATH=/usr/local/lib
 asimage_tiff=OFF
 builtin_cfitsio=ON
 builtin_cppzmq=ON
@@ -28,6 +27,7 @@ builtin_zeromq=ON
 builtin_zstd=ON
 ccache=ON
 cocoa=ON
+fortran=OFF
 minuit2_omp=OFF
 pythia8=OFF
 test_distrdf_dask=OFF


hahnjo:
I think instead of guessing around and applying more and more changes, we should carefully understand the problem in depth and then fix it for good

guitargeek:
> I think instead of guessing around and applying more and more changes, we should carefully understand the problem in depth and then fix it for good

Maybe before even doing that, we should be clear on whether we actually want to support `fortran=ON` on mac at this point. It was broken and untested for years, and we only made it work for the newer `macOS` with this PR last week:

 * https://github.com/root-project/root/pull/19174

But we see now it's still fragile, and the binaries we provide only work when the user installed `gfortran` in the same way as it's installed on our CI nodes anyway.

IMHO we need to have a clear strategy on which features of ROOT are supported on which platforms, because we don't necessarily need to support everything everywhere. Attempting that is taking a lot of time and effort, and the features that our users use vary greatly depending on the platform they're on. Once we have decided on these things, we can also make the support status more explicit by making certain configurations error out on certain platforms.

For example, on Windows, `fortran=ON` would still compile, if the user installs a compiler and makes it somehow findable by `enable_language(Fortran)`. But who can guarantee that `h2root` still works on Windows at this point and doesn't give you garbage result and corrupt your data?

So I agree with @hahnjo that we should not necessarily try to guess our way out of the problems, but I think we should go one step further and have a more general discussion on what we support on which platform. Possibly in the next ROOT team meeting.

hageboeck:
#21404 probably solved the problem with passing `CHOPT`. I don't think that all is fixed, but at least the string passing should be stable now.

PR #19237, opened 2025-06-30
[cppyy] Correct management of GIL in concurrency tests

guitargeek:
The GIL should always be acquired via the thread state API when doing Python calls from multithreaded C++.

Fixes the test failure with Python debug builds.

### DIFF
--- a/bindings/pyroot/cppyy/cppyy/test/test_concurrent.py
+++ b/bindings/pyroot/cppyy/cppyy/test/test_concurrent.py
@@ -11,6 +11,11 @@ class TestCONCURRENT:
         cls.data = [3.1415, 2.7183, 1.4142, 1.3807, -9.2848]
 
         cppyy.cppdef("""\
+        // as recommended by:
+        // https://docs.python.org/3/c-api/intro.html#include-files
+        #define PY_SSIZE_T_CLEAN
+        #include "Python.h"
+
         namespace Workers {
             double calc(double d) { return d*42.; }
         }""")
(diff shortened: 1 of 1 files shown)


bindings/pyroot/cppyy/cppyy/test/test_concurrent.py
```diff
@@ -105,17 +110,28 @@ def test04_cpp_threading_with_exceptions(self):
...
                 t = std::thread([this] {
                     int counter = 0;
+
+                    // Each thread needs a Python state object
+                    PyThreadState *pystate = PyThreadState_New(this->interpreterState);
+
                     while (counter++ < 10)
                         try {
+                            PyEval_RestoreThread(pystate);
                             cons->process(counter);
```
Vipul-Cariappa, line 135:
```suggestion
                            PyGILState_STATE gstate = PyGILState_Ensure();
                            cons->process(counter);
                            PyGILState_Release(gstate);
```

Accounted to the exception handling.

guitargeek: Thanks a lot for taking a look!

These functions that you mention are more higher-level wrappers for what I already do. Quoting from the docs you linked:

> The [PyGILState_Ensure()](https://docs.python.org/3/c-api/init.html#c.PyGILState_Ensure) and [PyGILState_Release()](https://docs.python.org/3/c-api/init.html#c.PyGILState_Release) functions do all of the above automatically.

..where all of the above is exactly registering the threads, acquiring the GIL and storing the thread state like I already do.

The reason why I went for this lower-level API is to avoid the overhead or registering and un-registering the thread *inside* the loop. I assume this is more representative of optimized user code. I updated my commit with a clearer explanation why I did this.

But you're right, in the other unit test where there is only a single callback to Python I can just use `Ensure` and `Release` :+1:

In any case, I'll open the same PR for Wims fork to also get feedback from him.

PR #19246, opened 2025-07-02
[cling] Check if `LLVM_SUBMIT_VERSION` is defined, not its value

aaronj0:
Reported here: https://github.com/root-project/cling/issues/551
We check if a variable exists named with the value of LLVM_SUBMIT_VERSION instead of checking if it is defined.

### DIFF
--- a/interpreter/cling/tools/Jupyter/CMakeLists.txt
+++ b/interpreter/cling/tools/Jupyter/CMakeLists.txt
@@ -84,7 +84,7 @@ set_target_properties(libclingJupyter
 if(ENABLE_SHARED)
     if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
         set(LIBCLINGJUPYTER_LINK_FLAGS " -Wl,-compatibility_version -Wl,1")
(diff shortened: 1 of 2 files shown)


aaronj0:
> The latest version doesn't check anymore that the variable is `DEFINED`.
> Also `interpreter/cling/tools/Jupyter/CMakeLists.txt` has the same pattern and should be fixed at the same time.

Thanks, that was an oversight there, that should not be a not false check and retain `DEFINED`. I have addressed that and replaced the pattern in Jupyter. 

> what is the use case for the variable

Based on my limited knowledge when I looked into this issue, this is used for a shared build with LLVM built the[ "apple way"](https://github.com/apple-oss-distributions/llvmCore/blob/main/build_llvm) way (perhaps no longer used?) which uses some autoconf build recipe as seen in `llvmCore`. As a part of that we pass the llvmCore submission number as a link flag.

hahnjo (approved):

aaronj0:
Failures on debian are unrelated and can be seen on master: https://github.com/root-project/root/actions/runs/16028087965/job/45221169747

PR #19251, opened 2025-07-02
Fix TFile gap coalescing

### LINKED ISSUE #19245: TFile crashes when trying to create a gap >2GB
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

When deleting several large, consecutive objects (or many small ones), TFile crashes or corrupts the file. The problem seems to be oversized gaps due to coalescing. If the gap becomes larger than 2GB, ROOT crashes. A gap between 1GB and 2GB is successfully created but an attempt to store a small object in this gap corrupts the file.

### Reproducer

```
#include <vector>

class Large {
public:
   std::vector<char> fVec;
};

auto f = TFile::Open("large.root", "RECREATE");
f->SetCompressionSettings(0);

auto obj = new Large();
obj->fVec.resize(1024*1024*1024 - 100, 'x'); // almost 1 GB

f->WriteObject(obj, "obj1");
f->WriteObject(obj, "obj2");

obj->fVec.resize(1024*1024, 'x');
f->WriteObject(obj, "obj3");

f->Write();
f->Close();
delete f;

f = TFile::Open("large.root", "UPDATE");
f->Delete("obj1;*");

// If the next two lines are uncommented, ROOT crashes
//f->Delete("obj2;*");
//f->Delete("obj3;*");

// Creates a combined gap close to 2GB
f->Delete("obj2;*");
f->Write();
f->Close();
delete f;

f = TFile::Open("large.root", "UPDATE");
f->WriteObject(obj, "obj4"); // corrupts the file
f->Write();
f->Close();
delete f;

f = TFile::Open("large.root");
f->Map("extended");
```

### ROOT version

master

### Installation method

From source

### Operating system

Linux

### Additional context

_No response_

jblomer:
Fixes #19245

### DIFF
--- a/io/io/inc/TFree.h
+++ b/io/io/inc/TFree.h
@@ -23,14 +23,16 @@
 
 #include "TObject.h"
 
-
 class TFree : public TObject {
-
 protected:
    Long64_t        fFirst;            ///<First free word of segment
    Long64_t        fLast;             ///<Last free word of segment
 
 public:
+   /// Prevent gap coalescing from exceeding 2 GB, since the gap size is stored as a signed 4 bytes integer.
+   /// The true limit is 2 GiB but for historical reasons (previous implementation of TFree) we stick to the 2GB limit.
+   static constexpr Int_t kMaxGapSize = 2000 * 1000 * 1000;
+
    TFree();
    TFree(TList *lfree, Long64_t first, Long64_t last);
    ~TFree() override;
--- a/io/io/src/TFile.cxx
+++ b/io/io/src/TFile.cxx
@@ -2149,7 +2149,11 @@ Int_t TFile::Recover()
 
    fEND = Long64_t(size);
 
-   if (fWritable && !fFree) fFree  = new TList;
+   if (fWritable) {
+      if (!fFree)
+         fFree = new TList();
+      fFree->Delete();
+   }
 
    Int_t nrecov = 0;
    nwheader = 1024;
@@ -2222,12 +2226,7 @@ Int_t TFile::Recover()
    if (fWritable) {
       Long64_t max_file_size = Long64_t(kStartBigFile);
       if (max_file_size < fEND) max_file_size = fEND+1000000000;
-      TFree *last = (TFree*)fFree->Last();
-      if (last) {
-         last->AddFree(fFree,fEND,max_file_size);
-      } else {
-         new TFree(fFree,fEND,max_file_size);
-      }
+      new TFree(fFree, fEND, max_file_size);
       if (nrecov) Write();
    }
    return nrecov;
--- a/io/io/src/TFile.cxx
+++ b/io/io/src/TFile.cxx
@@ -1496,27 +1496,59 @@ Bool_t TFile::IsOpen() const
 
 void TFile::MakeFree(Long64_t first, Long64_t last)
 {
-   TFree *f1      = (TFree*)fFree->First();
-   if (!f1) return;
-   TFree *newfree = f1->AddFree(fFree,first,last);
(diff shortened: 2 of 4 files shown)


io/io/src/TFile.cxx
```diff
@@ -1474,27 +1474,55 @@ Bool_t TFile::IsOpen() const
...
+      delete prev;
+   }
+   if (tail->GetFirst() <= nfirst) {
+      nfirst = tail->GetFirst();
+      nlast = tail->GetLast();
+   }
+
+   Long64_t nbytesl = nlast - nfirst + 1;
+   if (nbytesl > TFree::kMaxGapSize)
+      nbytesl = TFree::kMaxGapSize;
```
silverweed: If I understand correctly, here we are "sacrificing" `nbytesl - kMaxGapSize` bytes and leaving them unindexed; shouldn't we also add a new free segment right after `tail` to encompass them if this is not at the end of the file?

io/io/src/TFree.cxx
```diff
@@ -70,17 +70,23 @@ TFree *TFree::AddFree(TList *lfree, Long64_t first, Long64_t last)
    while (idcur) {
```
silverweed: This should be a `do / while` but I guess it's fine either way...

io/io/inc/TFree.h
```diff
@@ -23,14 +23,16 @@
...
-
 class TFree : public TObject {
-
 protected:
    Long64_t        fFirst;            ///<First free word of segment
    Long64_t        fLast;             ///<Last free word of segment
 
 public:
+   /// Prevent gap coalescing from exceeding 2 GB, since the gap size is stored as a signed 4 bytes integer.
+   /// The true limit is 2 GiB but for historical reasons we stick to the 2GB limit.
```
pcanal: Should we add a hint of where the historical reasons can be found? :)

pcanal:
```suggestion
   /// The true limit is 2 GiB but for historical reasons (previous implement of TFree) we stick to the 2GB limit.
```

io/io/src/TFree.cxx
```diff
@@ -70,17 +70,23 @@ TFree *TFree::AddFree(TList *lfree, Long64_t first, Long64_t last)
...
-         TFree *idnext = (TFree*)lfree->After(idcur);
-         if (idnext == 0) return idcur;
-         if (idnext->GetFirst() > last+1) return idcur;
-         idcur->SetLast( idnext->GetLast() );
+         // Merged new segment with previous one; is there a next segment?
+         TFree *idnext = static_cast<TFree *>(lfree->After(idcur));
+         if (idnext == 0)
+            return idcur;
+
+         // Is the next segment adjacent to the newly merged one (and not too big)
```
pcanal: Either
```suggestion
         //  Continue only if the next segment adjacent to the newly merged one (and not too big)
```
or 
Either
```suggestion
         //  Return unless the next segment adjacent to the newly merged one (and not too big)
```
As is the if statement and the comment and logically opposite

io/io/src/TFile.cxx
pcanal: (side note, this lost was already present is the current code).

jblomer: Actually, this condition can only be true for the last virtual free segment (the one from `[fEND + 1 .. Max File Size])`. I made the code more explicit.

jblomer:
... still some non-understood issues on recovery, working on that.

jblomer:
Replaced by https://github.com/root-project/root/pull/22481

PR #19255, opened 2025-07-02
MacVM-14 ARM64 production test

Ni-Langguth:
# This Pull request:

## Changes:
- Adds a runner flair for macos-14 on ARM64 machine
- keeps symlink of ROOT-CI to VM build_drive persistent


pcanal:
Somehow this PR was squash merged (the squash was a good thing) but with the commit log:
```
[ci] Adapt build script to new VM macOS runners
```
without a mention of the PR (the github gui at https://github.com/root-project/root/commit/cae4fa04ad1509d642cbc1530af0294594a50313 does show the PR).

With only the commit log as-is (eg from the command line), it is impossible to know the reasons for the change (at least with the link and/or number of the PR, one can go the jump to a browser and find the information).

In the future when (squash) merging, please ensure that there is at least a mention of the PR number or better yet an explanation of the reasons behind the final form of the change (what does it fix and why that is the right solution).

Thanks for you contribution and your attention.

Issue #11992, opened 2023-01-09
[RF] RooFormulaVar is too eager in variable name substitution


elusian:
- [x] Checked for duplicates

### Describe the bug

When passing a argument to a formula based RooFit function (e.g. `RooFormulaVar` or `RooGenericPdf`), RooFit will search for all occurrences of the object name in the formula and replace them with an ordinal parameter (needed for the underling `TFormula`). However, when the argument is a `RooConst` the name is just a string containing the value and this can break if there are numeric literals inside the formula.
For example
```
RooFormulaVar f("f", "2.7*@0", RooFit::RooConst(2.))
```
becomes inside the `TFormula`
```
x[0].7*x[0]
```
because the name 2 of the `RooConst` matched the 2 of the numeric literal. This of course breaks during formula compilation.

### Expected behavior

The variables do not replace numeric literals no matter the name. Possibly `RooConst`s should be matched only by index?

### To Reproduce

```
RooFormulaVar f("f", "2.7*@0", RooFit::RooConst(2.))
```

### Setup

HEAD of ROOT 6.26 from LCG dev4

### Additional context

`RooFormulaVar f("f", "2.7*@0", RooFit::RooConst(2.7))` creates `x[0]*x[0]`, which is mostly fine but may break if servers are redirected

guitargeek:
> Possibly `RooConst`s should be matched only by index?

Yes, I think this makes sense. Or maybe better: variables where the name is a number should only be matched by index.

elusian:
Haha, yes, it is a bigger framework, encouraged by the ease of creation of `RooConst` inside a `RooArgList` with the new pythonizations.

The solutions looks ok to me.

Ban42:
Hi @guitargeek ,
Just notifying that I started working on this issue as suggested in #19258

PR #19258, opened 2025-07-02
[RF] Implement checking param range

### LINKED ISSUE #7210: [RF] Implement checking of parameter ranges
### Explain what you would like to see improved
Since a while, RooFit can check if parameters are defined with a mathematically allowed range.
https://github.com/root-project/root/blob/c8e050d31d3ed4fefc5f4d28c2daf36953f15a30/roofit/roofit/src/RooGaussian.cxx#L42-L44

This should be implemented for all PDFs where we know that the definition range of the parameters is less than [-inf, inf].

Ban42:
# This Pull request:

## Changes or fixes:
As requested in #7210, adds parameters checks to:
- Breit-Wigner distribution
- Chi-Square distribution
- Decay distribution
- Exponential distribution

## Checklist:

- [ ] tested changes locally: I haven't, I am running into some problems when running roottest, will try to fix those problems and run the test locally in future PRs. This PR looked easy enough to skip them, at least locally. 
- [ ] updated the docs (if necessary): Tell me if I should.

## Questions from me
- I know there is more distributions to check the param of, I just wanted to be sure I was doing what requested before submitting 10+ commits.
- My statistics knowledge is not the best, so I hope I did not make mistakes on parameters ranges :see_no_evil: 
- I used [skip-CI] tag just because I wanted to trigger Jenkins just once after I have created parameter checks on all (or more) distributions. I guess that when removing the tag it will run the tests on the full PR and not only on the last commit, correct?
- I have also inserted the [skip-CI] tag in the commit messages, is this correct or unnecessary?

This PR (partially) fixes #7210 

Thanks!

### DIFF
--- a/roofit/roofit/src/RooChiSquarePdf.cxx
+++ b/roofit/roofit/src/RooChiSquarePdf.cxx
@@ -18,6 +18,7 @@ Here we also implement the analytic integral.
 #include "RooChiSquarePdf.h"
 #include "RooRealVar.h"
 #include "RooBatchCompute.h"
+#include "RooHelpers.h"
 
 #include "TMath.h"
 
@@ -39,6 +40,7 @@ RooChiSquarePdf::RooChiSquarePdf(const char* name, const char* title,
   _x("x", "Dependent", this, x),
   _ndof("ndof","ndof", this, ndof)
 {
+   RooHelpers::checkRangeOfParameters(this, {&x, &ndof}, 0.);
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/roofit/roofit/src/RooBifurGauss.cxx
+++ b/roofit/roofit/src/RooBifurGauss.cxx
@@ -24,6 +24,7 @@ side of maximum value.
 #include <RooBifurGauss.h>
 
 #include "RooBatchCompute.h"
+#include "RooHelpers.h"
 
 #include <RooFit/Detail/MathFuncs.h>
 
@@ -39,6 +40,7 @@ RooBifurGauss::RooBifurGauss(const char *name, const char *title, RooAbsReal &_x
      sigmaR("sigmaR", "Right Sigma", this, _sigmaR)
 
 {
+   RooHelpers::checkRangeOfParameters(this, {&_sigmaL, &_sigmaR}, 0.);
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/roofit/roofit/src/RooBreitWigner.cxx
+++ b/roofit/roofit/src/RooBreitWigner.cxx
@@ -27,6 +27,7 @@ that models a non-relativistic Breit-Wigner shape
 
 #include "RooBreitWigner.h"
 #include "RooRealVar.h"
+#include "RooHelpers.h"
 #include "RooBatchCompute.h"
 
 
@@ -40,6 +41,7 @@ RooBreitWigner::RooBreitWigner(const char *name, const char *title,
   mean("mean","Mean",this,_mean),
   width("width","Width",this,_width)
 {
+   RooHelpers::checkRangeOfParameters(this, {&_width}, 0.);
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/roofit/roofit/src/RooPoisson.cxx
+++ b/roofit/roofit/src/RooPoisson.cxx
@@ -17,24 +17,19 @@ Poisson pdf
 #include "RooMath.h"
 #include "RooNaNPacker.h"
 #include "RooBatchCompute.h"
+#include "RooHelpers.h"
 
 #include <RooFit/Detail/MathFuncs.h>
 
 #include <array>
 
+ ////////////////////////////////////////////////////////////////////////////////
+ /// Constructor
 
-////////////////////////////////////////////////////////////////////////////////
-/// Constructor
-
-RooPoisson::RooPoisson(const char *name, const char *title,
-             RooAbsReal::Ref _x,
-             RooAbsReal::Ref _mean,
-             bool noRounding) :
-  RooAbsPdf(name,title),
-  x("x","x",this,_x),
-  mean("mean","mean",this,_mean),
-  _noRounding(noRounding)
-{
+ RooPoisson::RooPoisson(const char *name, const char *title, RooAbsReal::Ref _x, RooAbsReal::Ref _mean, bool noRounding)
+    : RooAbsPdf(name, title), x("x", "x", this, _x), mean("mean", "mean", this, _mean), _noRounding(noRounding)
+ {
+    RooHelpers::checkRangeOfParameters(this, {&static_cast<RooAbsReal &>(_x), &static_cast<RooAbsReal &>(_mean)}, 0.);
 }
 
 ////////////////////////////////////////////////////////////////////////////////


pcanal:
Why use the 'skip-CI' tag?

pcanal:
> I have also inserted the [skip-CI] tag in the commit messages, is this correct or unnecessary?

It would be misleading.  '[skip-CI]' indicates changes that have no functional effect, which is not the case for this commit/PR.

pcanal:
> I used [skip-CI] tag just because I wanted to trigger Jenkins just once after I have created parameter checks on all (or more) distributions. I guess that when removing the tag it will run the tests on the full PR and not only on the last commit, correct?

I don't think we have a trigger on removing the tag.

Since this PR is still awaiting more changes (including tests?), I move the PR is draft mode.

For this PR, the CI will run only with one the maintainer launching it.  Please let us know when it is ready to run the CI, for example by marking it as ready for review.

guitargeek:
Thank you very much for making the first step towards closing that PR!

Personally, I have not addressed this issue because of two reasons:
  1. Not every PDF class in RooFit gets used a lot, so going through *all* classes is not necessarily effort well spent
  2. The mechanism to check for invalid ranges is fragile, because ranges of parameters can be changed at any time.

But you are right, it's of course better to do *something* rather than to do nothing :)

So let's maybe implement the checks for 2  or 3 more commonly used classes so we can claim success. What about `RooPoisson` and `RooBifurGauss` for example? Could you do this in this PR too?

roofit/roofit/src/RooChiSquarePdf.cxx
```diff
@@ -39,6 +40,7 @@ RooChiSquarePdf::RooChiSquarePdf(const char* name, const char* title,
   _x("x", "Dependent", this, x),
   _ndof("ndof","ndof", this, ndof)
 {
+  RooHelpers::checkRangeOfParameters(this, {&_ndof}, 0.);
```
guitargeek:
```suggestion
  RooHelpers::checkRangeOfParameters(this, {&_x, &_ndof}, 0.);
```
Also the x variable for the chi2 pdf should be positive. If not, users will get NaN pdf/likelihood values, so it's good to warn about this too!

roofit/roofit/src/RooExponential.cxx
```diff
@@ -45,6 +46,7 @@ RooExponential::RooExponential(const char *name, const char *title, RooAbsReal &
      c{"c", "Exponent", this, coefficient},
      _negateCoefficient{negateCoefficient}
 {
+   RooHelpers::checkRangeOfParameters(this, {&coefficient}, 0.);
```
guitargeek:
```suggestion
```
This should not be done, as the class can be used to model both exponential decreases and increases

roofit/roofit/src/RooDecay.cxx
```diff
@@ -64,6 +65,7 @@ RooDecay::RooDecay(const char *name, const char *title,
     _basisExp = declareBasis("exp(-abs(@0)/@1)",tau) ;
     break ;
   }
+  RooHelpers::checkRangeOfParameters(this, {&_tau}, 0.);
```
guitargeek:
```suggestion
```
I would not do this check either (yes, the name `RooDecay` suggests that tau should be positive, but I have no reason to assume that this class technically doesn't work for negative tau as well).

roofit/roofit/src/RooExponential.cxx
Ban42: I assumed that to model exponential decrease the user should keep `coefficient > 0` and have `negateCoefficient = true`?

If you insist will remove it ofc, but it does not seem right to me..

roofit/roofit/src/RooDecay.cxx
Ban42: Ok, I have reverted the commit instead of accepting your commit suggestion because there was also an include statement.

Ban42:
Hi @guitargeek , thanks for the nice words :)

Please have a look at my comments and latest commit and let me know.

> 2. The mechanism to check for invalid ranges is fragile, because ranges of parameters can be changed at any time.

If these parameters are changed using setters or similar, we can add the check also in those functions, at least for the most used distributions. Maybe in another PR? I would be happy to work on it.

roofit/roofit/src/RooExponential.cxx
guitargeek: No, a negative coefficient is always okay. The flag `negateCoefficient` just flips the sign of the coefficient, so you can have a consistent definition of the RooExponential with the definition of an exponential distribution in other frameworks.

Sorry for the confusion, it comes from the name of that flag that was maybe not chosen well! Maybe read it as `flipCoefSign`...

guitargeek:
> If these parameters are changed using setters or similar, we can add the check also in those functions, at least for the most used distributions. Maybe in another PR? I would be happy to work on it.

It's not with simple setters: all objects in the RooFit model (e.g. all RooAbsArg-related classes) are owned by the user, and they can change the range of any variable any time with `RooRealVar::setRange()`. And there is no callback to the pdfs that use the variables when that happens. Maybe one could do the checks again in the `RooMinimizer` whenever a likelihood is numerically minimized, because that's when the range of the parameters matters. But I guess that people reset the parameter range in invalid way like that is quite the edge case, so I wouldn't over-engineer this.

So IMO it would be best to wrap this PR up quickly (it just needs to be fixed to compile again), and then if you're still available to help out maybe address one of the other relatively easy self contained issues :slightly_smiling_face:

  * https://github.com/root-project/root/issues/11992

Of course I'm not expecting anything, but just want to guide you to where I think contributors can have the biggest impact.

roofit/roofit/src/RooExponential.cxx
Ban42: Thanks for explaining, I have reverted the change.

dpiparo:
starting the ci jobs. Thanks @Ban42

dpiparo:
The code is not building correctly, we have at least an issue:
```
roofit/roofit/src/RooChiSquarePdf.cxx: In constructor ‘RooChiSquarePdf::RooChiSquarePdf(const char*, const char*, RooAbsReal&, RooAbsReal&)’:
 ##[error] roofit/roofit/src/RooChiSquarePdf.cxx:43:38: error: could not convert ‘{(&((RooChiSquarePdf*)this)->RooChiSquarePdf::_x), (&((RooChiSquarePdf*)this)->RooChiSquarePdf::_ndof)}’ from ‘<brace-enclosed initializer list>’ to ‘std::initializer_list<const RooAbsReal*>’
    43 |    RooHelpers::checkRangeOfParameters(this, {&_x, &_ndof}, 0.);
       |    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
       |                                      |
       |                                      <brace-enclosed initializer list>
```

ferdymercury:
Thanks for this initiative! A user requested a check also on the class RooJohnson
(see https://its.cern.ch/jira/browse/ROOT-10945)

Ban42:
Hi everyone,

Thanks @guitargeek for the explanations. After I wrap up this issue I will be happy to have a look at #11992 
Thanks @dpiparo for the suggestion, I will check and fix asap.
Hi @ferdymercury I'll add the check to RooJohnson before merging this.

I am taking some time to try to run the roottests on my local machine (some of them are failing and I don't understand why), but I'll come back to this issue very soon.

Ban42:
Hi guys,
I was able to run the tests, it should build now (hopefully :sweat_smile: )

@ferdymercury I had a look at RooJohnson and it looks like the parameters are already checked in the constructor, so you can close that issue on JIRA :)

guitargeek (approved):
Thank you very much for this PR and addressing the review comments!

PR #19262, opened 2025-07-03
[CI] Remove the upload event payload step and allow for using labels to skip CI runs

hageboeck:
- It looks like the upload event payload step is just polluting the artifact storage. Similar information is available directly in the workflow logs.
- Along the lines of "clean build", add the label "skip ci" for an easier way to skip runs

### DIFF
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -83,7 +83,11 @@ jobs:
     # can be skipped if the tag [skip-ci] or [skip ci] is written in the title.
     if: |
         (github.repository_owner == 'root-project' && github.event_name != 'pull_request') ||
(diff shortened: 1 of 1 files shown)


ferdymercury:
> * Along the lines of "clean build", add the label "skip ci" for an easier way to skip runs

Thanks! Looks great.

Maybe an improvement if someone adds or removes the skip-ci label in the course of a review:

using:

```
on:
  pull_request:
    types: [labeled] // or unlabeled see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
```

we could cancel a running build if the label is added, or initiate a new build if the label is removed.

hageboeck:
> we could cancel a running build if the label is added, or initiate a new build if the label is removed.

Hello, it sounds like a good idea, but is probably more complicated than it sounds. Github already cancels jobs when you push, so I would like to leave it at that for the moment.

dpiparo (approved):

PR #19263, opened 2025-07-03
[eve7] fix GlViewerJSRoot and GlViewerThree

linev:
When using TGeoPainter in GlViewerJSRoot,
process geometry description data as is - 3D objects building remains inside painter
This let use navigation and interactivity from TGeoPainter again

Set near/far parameters in PerspectiveCamera of GlViewerThree
If geometry size significantly bigger than default 1000 - camera will not be able to "see" such object

Update JSROOT with several geopainter fixes and three.js r177


js/modules/rntuple.mjs
bellenot, line 1: Are these changes in `rntuple.mjs` intended to be part of this geometry PR?

bellenot (approved):
LGTM, but see my question

js/modules/rntuple.mjs
linev: No, it is current progress of our GSOC student.
Code is not used for production, but my automatic script just copies everything from JSROOT to ROOT repository.

bellenot: Then can you remove it?

linev: Done

PR #19266, opened 2025-07-03
[roottest] cmake for io/newClassDef tests

linev:
* fully remove unused code from `io/newClassDef/event` directory
* simplify several tests which were testing old CINT problems which are resolved like `NESTING_PROBLEM_SOLVED` or `R__GLOBALSTL`
* some tests fail on windows like trying to redefine `protected` and `private` keywords, therefore such tests protected with `win_broken_tests`
* remove `io/newClassDef/showmember` directory while it contains lot of existing tutorials macros which were not used in testing


pcanal (approved):
LGTM (pending CI result).  Thanks.

roottest/root/io/newClassDef/current/CMakeLists.txt
```diff
@@ -14,10 +14,16 @@ ROOTTEST_GENERATE_DICTIONARY(libIoNewClassnamespace
                              NO_CXXMODULE
                              FIXTURES_SETUP root-io-newClassDef-current-namespace-fixture)
 
+# there is no ClassDefT macro with override syntax, therefore suppress warning
```
pcanal: `ClassDefT` is obsolete.  You can used `ClassDefOverride` in its stead.

linev: But that about ClassDefT2?

linev: And may be we can keep old code - to test compatibility?

pcanal: > But that about ClassDefT2?

Currently `ClassDefT` is literally an alias for `ClassDef` and `ClassDefT2` is an empty macro (both since at least v3-03-03 ... 23 years ago :) ).

linev: Ok, lets try with `ClassDefOverride`

linev: It works! 
Was never aware about existing of such defines.

PR #5568, opened 2020-05-07
[cling] DlClose what DlLibMan has DlOpened (ROOT-10659).

Axel-Naumann:
https://its.cern.ch/jira/browse/ROOT-10659

Issue #16270, opened 2024-08-20
[cling] Problems with many DynamicLibraryManager tests on macOS


hahnjo:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

https://github.com/root-project/root/pull/16258 proposes to disable many `DynamicLibraryManager` tests on macOS that never worked on that platform since their introduction (they pass fine on Linux). The tests should be debugged, fixed, and eventually re-enabled.

### Reproducer

.

### ROOT version

`master`

### Installation method

from source

### Operating system

macOS

PR #19268, opened 2025-07-03
[cling] DlClose what DlLibMan has DlOpened in reverse order

ferdymercury:
All credit goes to @Axel-Naumann, this is just a rebase of https://github.com/root-project/root/pull/5568/ to see if it still works / makes sense or if it should just be closed.

Not sure if this is exhaustive, there might be other places to take care:

```
core/metacling/src/TClingCallbacks.cxx
core/base/src/TROOT.cxx
core/metacling/src/TCling.cxx
interpreter/cling/lib/Utils/PlatformPosix.cpp
interpreter/cling/lib/Interpreter/IncrementalParser.cpp
interpreter/CppInterOp/lib/CppInterOp/CppInterOpInterpreter.h
interpreter/cling/lib/Utils/PlatformPosix.cpp
bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
bindings/pyroot/pythonizations/python/ROOT/_jupyroot/helpers/cppcompleter.py
```

Related:
https://its.cern.ch/jira/browse/ROOT-10659
https://its.cern.ch/jira/browse/ROOT-6022

### DIFF
--- a/interpreter/cling/include/cling/Interpreter/DynamicLibraryManager.h
+++ b/interpreter/cling/include/cling/Interpreter/DynamicLibraryManager.h
@@ -56,6 +56,7 @@ namespace cling {
     ///\brief DynamicLibraries loaded by this Interpreter.
     ///
     DyLibs m_DyLibs;
+    std::vector<DyLibHandle> m_OpenDyLibs;
     llvm::StringSet<> m_LoadedLibraries;
 
     ///\brief System's include path, get initialized at construction time.
--- a/interpreter/cling/lib/Interpreter/DynamicLibraryManager.cpp
+++ b/interpreter/cling/lib/Interpreter/DynamicLibraryManager.cpp
@@ -393,6 +393,7 @@ namespace cling {
     if (!insRes.second)
       return kLoadLibAlreadyLoaded;
     m_LoadedLibraries.insert(canonicalLoadedLib);
+    m_OpenDyLibs.push_back(dyLibHandle);
     return kLoadLibSuccess;
   }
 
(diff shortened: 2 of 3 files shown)


ferdymercury:
@hageboeck I see some failures, maybe related with CMake targets. Do I need to add here the clean-build label?

hageboeck:
> @hageboeck I see some failures, maybe related with CMake targets. Do I need to add here the clean-build label?

Most likely yes.

pcanal:
There are still failures, so they maybe actually related.  Let's try one more rerun ...

ferdymercury:
> There are still failures, so they maybe actually related. Let's try one more rerun ...

Yes, the failures seem related.

ferdymercury:
One of the failing tests is https://github.com/root-project/root/blob/master/roottest/root/tree/cloning/runtreeCloneTest2.C

which does not cleanup memory before the end of the script.

Question: should we investigate why the crash happens? Or should we just do a proper cleanup at the end of the script, meaning file->Close() delete file;?

pcanal:
Unfortunately we need to investigate the failure.  However awkward that setup is, it is supported (on a best effort basis only though).

ferdymercury:
Since there are less tests failing in debug mode, I'd bet it's a race condition.
I'm not sure how to debug this though since I don't have mac. Is there a way to activate valgrind in the CI in addition to coredump?

Or do you think rather that the problem is from other dlopen or dlclose places that are not being taken care of?

dpiparo:
restarting the build with the new code in master...

ferdymercury:
Still failing :s

Maybe related? https://github.com/root-project/root/issues/16270

Issue #19241, opened 2025-07-01
THnBase::ProjectionAny does not calculate errors correctly


jgrosseo:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

It seems to me that THnBase::ProjectionAny does not respect the user's choice of calculating the error correctly even when "E" is given as an option. Please find a minimal reproducer attached which compares the behaviour of TH2 -> TH1 (correct) to THn -> THn (not correct).  

Please let me know if I am overlooking something.

### Reproducer

```
void thn()
{
    auto hist = new TH2F("hist", "", 10, 0.5, 10.5, 10, 0.5, 10.5);
    hist->Sumw2();

    for (int i = 0; i < 9; i++)
        hist->Fill(1, 1, 0.1);

    float relError = hist->GetBinError(1, 1) / hist->GetBinContent(1, 1);
    printf("TH2:  %f +- %f --> %f\n", hist->GetBinContent(1, 1), hist->GetBinError(1, 1), relError);

    auto proj1 = hist->ProjectionX();
    relError = proj1->GetBinError(1, 1) / proj1->GetBinContent(1, 1);
    printf("Proj: %f +- %f --> %f\n", proj1->GetBinContent(1), proj1->GetBinError(1), relError);

    int bins[] = {10, 10, 10, 10};
    double xmin[] = {0, 0, 0, 0};
    double xmax[] = {10, 10, 10, 10};
    auto thn = new THnF("thn", "", 4, bins, xmin, xmax);
    thn->Sumw2();

    for (int i = 0; i < 9; i++)
        thn->Fill(0.5, 0.5, 0.5, 0.5, 0.1);

    double values[]{0.5, 0.5, 0.5, 0.5};
    auto binNo = thn->GetBin(values);
    relError = thn->GetBinError(binNo) / thn->GetBinContent(binNo);
    printf("\nTHn:  %f +- %f --> %f\n", thn->GetBinContent(binNo), thn->GetBinError(binNo), relError);

    Int_t dimensions[] = {0, 1};
    // Despite the option "E", the errors are resetted to sqrt(N) instead of keeping the original ones
    THnBase *proj = thn->ProjectionND(2, dimensions, "E");

    binNo = proj->GetBin(values);
    relError = proj->GetBinError(binNo) / proj->GetBinContent(binNo);
    printf("Proj: %f +- %f --> %f\n", proj->GetBinContent(binNo), proj->GetBinError(binNo), relError);
}
```

The output is
```
[...]

PR #19271, opened 2025-07-03
[hist] Fix error calculation in THnBase::ProjectionAny.

hageboeck:
Thanks to jgrosseo!

Fix https://github.com/root-project/root/issues/19241

### DIFF
--- a/hist/hist/inc/THnBase.h
+++ b/hist/hist/inc/THnBase.h
@@ -258,6 +258,14 @@ protected:
       return (TH3D*) ProjectionAny(3, dim, false, option);
    }
 
+   /// @brief Project histogram onto the axes specified in dim.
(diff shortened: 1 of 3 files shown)


dpiparo (approved):
Thanks a lot.
This fix should be backported to 6.36 too.

jgrosseo:
Thanks a lot!

Will you put something in the release notes? A lot of uncertainties were (in most cases) too large but could also be too small because of this...

hageboeck:
> Thanks a lot!
> 
> Will you put something in the release notes? A lot of uncertainties were (in most cases) too large but could also be too small because of this...

The release notes will contain a link to #19241 and the title of the issue. I think that you chose quite a good title already, but we could iterate on it if you think we should make it more visible.

PR #19273, opened 2025-07-03
[tree] raise error if entry not properly loaded in EvalInstance

ferdymercury:
# This Pull request:

## Changes or fixes:

See https://root-forum.cern.ch/t/ntuples-unable-to-read/63847/7

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/tree/treeplayer/src/TTreeFormula.cxx
+++ b/tree/treeplayer/src/TTreeFormula.cxx
@@ -96,12 +96,17 @@ The following method are available from the TFormLeafInfo interface:
 ClassImp(TTreeFormula);
 
 ////////////////////////////////////////////////////////////////////////////////
-
-inline static void R__LoadBranch(TBranch* br, Long64_t entry, bool quickLoad)
+/// The function returns the number of bytes read from the input buffer.
+/// If entry does not exist or (entry!=readEntry && quickload), the function returns 0.
+/// If an I/O error occurs, the function returns -1.
+///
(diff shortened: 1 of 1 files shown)


tree/treeplayer/src/TTreeFormula.cxx
```diff
@@ -3840,8 +3908,8 @@ const char* TTreeFormula::EvalStringInstance(Int_t instance)
       if (fDidBooleanOptimization) {                                                            \
          TBranch *br = leaf->GetBranch();                                                       \
          if (br->GetTree()) {                                                                   \
-            Long64_t treeEntry = br->GetTree()->GetReadEntry();                                 \
-            if (br->GetReadEntry() != treeEntry) br->GetEntry( treeEntry );                     \
+            Long64_t tEntry = br->GetTree()->GetReadEntry();                                    \
```
pcanal, line 3911: Why the variable name shortening?

ferdymercury: It was tEntry in some places and treeEntry in some places, so I chose one.

pcanal: fair enough.  I still prefer the longer name at least here (and maybe in those other places)

hahnjo:
It would have been good to squash the commits, now I think we have intermediate states that fail to build and / or fail the tests...

PR #19274, opened 2025-07-03
[test] only classes deriving from TObject should be marked ClassDef-override

ferdymercury:
# This Pull request:

## Changes or fixes:

compilation error when calling "make" inside test/


guitargeek (approved):
Thanks!

Issue #19269, opened 2025-07-03
Use of -r argument crashes rootcling


linev:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Since a while we are using -r argument in rootcling to be able re-create dictionary and/or rootmap when dictionary file already exists. Starting from ROOT 6.36 use of -r argument crashes ROOT.

Problem not exists with ROOT 6.34 and earlier.

### Reproducer

In the `test/` subdirectory of ROOT just invoke:
```
rootcling -r G__XXX.cxx Event.h EventLinkDef.h 
```
Stack trace is:
```
0  libCling.so    0x00007f7f96e613a8
1  libCling.so    0x00007f7f96e5e947
2  libc.so.6      0x00007f7f90842260
3  libstdc++.so.6 0x00007f7f90d52864 std::ostream::sentry::sentry(std::ostream&) + 4
4  libstdc++.so.6 0x00007f7f90d53469 std::basic_ostream<char, std::char_traits<char>>& std::__ostream_insert<char, std::char_traits<char>>(std::basic_ostream<char, std::char_traits<char>>&, char const*, long) + 41
5  libCling.so    0x00007f7f91e14238 ROOT::TMetaUtils::WriteRulesRegistration(std::ostream&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>>> const&) + 168
6  libCling.so    0x00007f7f91e8521f GenerateFullDict(std::ostream&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>, cling::Interpreter&, RScanner&, std::__cxx11::list<ROOT::TMetaUtils::RConstructorType, std::allocator<ROOT::TMetaUtils::RConstructorType>> const&, bool, bool, bool, bool) + 1247
7  libCling.so    0x00007f7f91e8e036 RootClingMain(int, char**, bool) + 33430
8  libCling.so    0x00007f7f91e92f3d ROOT_rootcling_Driver + 253
9  rootcling      0x0000000000401156 main + 118
10 libc.so.6      0x00007f7f9082b37b
11 libc.so.6      0x00007f7f9082b44b __libc_start_main + 141
12 rootcling      0x0000000000401195 _start + 37
```

### ROOT version

master, 6.36

### Installation method

[...]

PR #19275, opened 2025-07-03
[core] prevent nullptr access in rootcling -r

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/19269

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/dictgen/src/rootcling_impl.cxx
+++ b/core/dictgen/src/rootcling_impl.cxx
@@ -4969,6 +4969,14 @@ int RootClingMain(int argc,
          rootclingRetCode +=  FinalizeStreamerInfoWriting(interp);
       }
    } else {
+      if (!splitDictStream) {
+         if (gOptSplit) {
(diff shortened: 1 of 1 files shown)


core/dictgen/src/rootcling_impl.cxx
```diff
@@ -4969,6 +4969,14 @@ int RootClingMain(int argc,
          rootclingRetCode +=  FinalizeStreamerInfoWriting(interp);
       }
    } else {
```
pcanal, line 4971: This pattern is used elsewhere.
```suggestion
   } else if (!gOptIgnoreExistingDict) {
```
but works only if the rootmap files are being generated not in `GenerateFullDict`.

However, this code has not changed for a while and thus we need to investigate further what 'changed' to provoke the issue (this may or may not just 'hide' the real underlying issue).

ferdymercury: Closing in favor of https://github.com/root-project/root/pull/19400

PR #19277, opened 2025-07-03
[Math] Change TDataPointN::kDimension from static member to template arg

guitargeek:
```
math/mathcore/inc/Math/TDataPointN.h: static UInt_t kDimension;
```

That should likely be a template parameter; multiple KDTrees might
disagree on the dimensionality, so it shouldn't be static!

Changing the template arguments at this time should still be fine, as
the `TDataPointN` class was only introduced in 2011 (https://github.com/root-project/root/commit/52e778230f09cf8df5418257b3372f24323e1eb1) and is
also undocumented, untested, and not covered by release notes or
tutorials.

https://its.cern.ch/jira/browse/ROOT-9323


math/mathcore/inc/Math/TDataPoint.h
```diff
@@ -13,58 +13,128 @@
...
+//
+// This class can be used for describing data points in a high-dimensional space.
+// The (positive) dimension is specified by the first template parameter. The second
+// template parameter can be used to tweak the precision of the stored coordinates. By
+// default all coordinates are stored with 4 byte float precision. In addition to the
+// coordinates a weight can be assigned to each data point allowing the representation
+// of fields in high dimensions.
+// Basic functionality for accessing/modifying the coordinates/weight are provided
+// as well as a comparison method and the basic euclidean metric.
+// End_Html
```
ferdymercury:
```suggestion
/// \brief class representing a data point
///
/// This class can be used for describing data points in a high-dimensional space.
/// The (positive) dimension is specified by the first template parameter. The second
/// template parameter can be used to tweak the precision of the stored coordinates. By
/// default all coordinates are stored with 4 byte float precision. In addition to the
/// coordinates a weight can be assigned to each data point allowing the representation
/// of fields in high dimensions.
/// Basic functionality for accessing/modifying the coordinates/weight are provided
/// as well as a comparison method and the basic euclidean metric.
```

math/mathcore/inc/Math/TDataPoint.h
```diff
@@ -13,58 +13,128 @@
...
-   template<typename _coord_typ>
-   TDataPoint(const _coord_typ* pData,_val_type fWeight = 1);
-#endif
-   //virtual ~TDataPoint() {}
+   static UInt_t Dimension() { return kDimension; }
+   TDataPoint()
+   {
+      // standard constructor
+      //
+      // sets the weight to 1 and initialises all coordinates with 0
```
ferdymercury:
```suggestion
   /// standard constructor
   /// sets the weight to 1 and initialises all coordinates with 0
   TDataPoint()
   {
```

math/mathcore/inc/Math/TDataPoint.h
```diff
@@ -13,58 +13,128 @@
...
+      for (UInt_t k = 0; k < K; ++k)
+         m_vCoordinates[k] = 0;
+   }
 #ifndef __MAKECINT__
-   template<typename _val>
-   value_type   Distance(const TDataPoint<K,_val>& rPoint) const;
+   template <typename _coord_typ>
+   TDataPoint(const _coord_typ *pData, _val_type fWeight = 1)
+   {
+      // constructor initialising the data point from an array
```
ferdymercury: same comments here, convert to doxygen style.
Thanks!!

PR #19279, opened 2025-07-04
[CMake] Standardize Smatrix and GenVector libraries

guitargeek:
It's a bit strange that these two packages separate their dictionaries from the libraries, being the only math libraries that use `ROOT_GENERATE_DICTIONARY` in their CMake code. Especially for modules builds, this is inconsistent, because with that macro you also get `rootmap` files, while we just want `pcm` files. Having both can be problematic in more fragile environments.

It's probably better to use `ROOT_STANDARD_LIBRARY_PACKAGE` all the way.

With this commit, no `rootmap` files from standard ROOT packages remain in the install trees `lib` for modules enabled. Just these files here, where I'm not sure if they are needed:
```txt
libcomplexDict.rootmap
libdequeDict.rootmap
libforward_listDict.rootmap
liblistDict.rootmap
libmap2Dict.rootmap
libmapDict.rootmap
libmultimap2Dict.rootmap
libmultimapDict.rootmap
libmultisetDict.rootmap
libsetDict.rootmap
libunordered_mapDict.rootmap
libunordered_multimapDict.rootmap
libunordered_multisetDict.rootmap
libunordered_setDict.rootmap
libvalarrayDict.rootmap
libvectorDict.rootmap
```

### DIFF
--- a/math/genvector/CMakeLists.txt
+++ b/math/genvector/CMakeLists.txt
@@ -8,30 +8,8 @@
 # CMakeLists.txt file for building ROOT math/genvector package
 ############################################################################
 
-ROOT_LINKER_LIBRARY(GenVector
-    src/3DConversions.cxx
-    src/3DDistances.cxx
-    src/AxisAngle.cxx
-    src/AxisAngleXother.cxx
-    src/BitReproducible.cxx
-    src/Boost.cxx
-    src/BoostX.cxx
-    src/BoostY.cxx
-    src/BoostZ.cxx
-    src/EulerAngles.cxx
-    src/LorentzRotation.cxx
-    src/Quaternion.cxx
-    src/QuaternionXaxial.cxx
-    src/Rotation3D.cxx
-    src/Rotation3DxAxial.cxx
-    src/RotationZYX.cxx
-    src/VectorUtil.cxx
-  DEPENDENCIES
-    Core
-    MathCore
-)
-
-ROOT_GENERATE_DICTIONARY(G__GenVector
+ROOT_STANDARD_LIBRARY_PACKAGE(GenVector
+  HEADERS
     Math/AxisAngle.h
     Math/Boost.h
     Math/BoostX.h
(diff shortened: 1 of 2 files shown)


guitargeek:
Not needed anymore for me.

math/genvector/CMakeLists.txt
```diff
@@ -142,35 +120,31 @@ ROOT_GENERATE_DICTIONARY(G__GenVector
...
+    src/Quaternion.cxx
+    src/QuaternionXaxial.cxx
+    src/Rotation3D.cxx
+    src/Rotation3DxAxial.cxx
+    src/RotationZYX.cxx
+    src/VectorUtil.cxx
   LINKDEF
-    Math/LinkDef_GenVector32.h
-  OPTIONS
+    Math/LinkDef_GenVectorAll.h
```
pcanal, line 142: The 2 sets of dictionaries are completely necessary.  Part of the inner working of the dictionary is to call some internal function based on the signature.  So there is a inner call to (for example)`internal_function(ROOT::Math::Cartesian2D<Double32_t>*)`
However we request the dictionary for both: `ROOT::Math::Cartesian2D<Double32_t>` and `ROOT::Math::Cartesian2D<double>`.  Since `Double32_t` is a typedef to `double`, the C++ type is actually the same and the 2 entities are distinguished only by their character string representation.
In particular this means that if the request for a dictionary for these 2 types is in 2 distinct dictionaries, everything work.  If they are in the same dictionary, then the scaffolding fails because the compiler will not, can not make a difference between `internal_function(ROOT::Math::Cartesian2D<Double32_t>*)` and `internal_function(ROOT::Math::Cartesian2D<double>*)` leading to the incorrect behavior.

guitargeek: Thanks a lot for clarifying @pcanal, that's very much appreciated! I was suspecting something like this is going on, but it's good to understand it better not thanks to you comment!

Issue #13659, opened 2023-09-18
rootprint/rootls missing recursive traversal


ferdymercury:
### Explain what you would like to see improved and how.

`rootprint` does not output canvases that are stored inside a subdirectory.

The CLI should accept a 'recursive' parameter that allows to deal with these cases

### Additional context

https://root-forum.cern.ch/t/saving-all-plots-from-a-root-file-into-separate-png-files/56379/3?u=ferhue

couet:
The code is in `$ROOTSYS/main/python/cmdLineUtils.py` at line 1394. The function retrieving the list of object to be printed is:

```
def _keyListExtended(rootFile,pathSplitList):
    keyList,dirList = keyClassSpliter(rootFile,pathSplitList)
    for pathSplit in dirList: keyList.extend(getKeyList(rootFile,pathSplit))
    keyList = [key for key in keyList if not isDirectoryKey(key)]
    keyListSort(keyList)
    return keyList
```
Somehow it should be extended in order to scan sub-directories.

ferdymercury:
Reopening as the merged PR only implemented rootls but not yet rootprint

PR #19282, opened 2025-07-04
[main] Rewrite rootls in c++

silverweed:
# This Pull request:
Adds a native version of the `rootls` executable. This allows running rootls without Python, meaning one can use it from a ROOT build with pyroot=off.
As a bonus, this version is ~3 to 4x faster than the python version on all the tests I did locally (bringing the "minimal" latency from ~0.5s to ~0.15s).

The output format should be exactly the same, except for some cases I found which rootls.py doesn't handle correctly.

The executable is currently called `rootlscxx` and the original `rootls` is preserved, but ideally it should be phased out after verifying that everything is covered.

I copied the tests for rootls.py and made sure they all pass with rootlscxx as well.

## NOTE
To handle globbing I imported a small header-only library called [wildcards.hpp](https://github.com/zemasoft/wildcards). The license should be compatible with ROOT (boost license).

## Checklist:

- [x] tested changes locally

### DIFF
--- a/main/src/rootls.cxx
+++ b/main/src/rootls.cxx
@@ -1,0 +1,154 @@
+#include <string>
+#include <vector>
+#include <iostream>
+#include <algorithm>
+#include <regex>
+
+#include "wildcards.hpp"
+
+#include <TFile.h>
+#include <TKey.h>
+
+#include <ROOT/StringUtils.hxx>
+
+static const char *const ONE_HELP = "Print content in one column";
+static const char *const LONG_PRINT_HELP = "Use a long listing format.";
+static const char *const TREE_PRINT_HELP = "Print tree recursively and use a long listing format.";
+static const char *const RECURSIVE_PRINT_HELP = "Traverse file recursively entering any TDirectory.";
+static const char *const EPILOG = R"(Examples:
+- rootls example.root
+  Display contents of the ROOT file 'example.root'.
+
+- rootls example.root:dir
+  Display contents of the directory 'dir' from the ROOT file 'example.root'.
+
+- rootls example.root:*
+  Display contents of the ROOT file 'example.root' and his subdirectories.
+
+- rootls file1.root file2.root
+  Display contents of ROOT files 'file1.root' and 'file2.root'.
+
+- rootls *.root
+  Display contents of ROOT files whose name ends with '.root'.
+
+- rootls -1 example.root
+  Display contents of the ROOT file 'example.root' in one column.
+
+- rootls -l example.root
+  Display contents of the ROOT file 'example.root' and use a long listing format.
+
+- rootls -t example.root
+  Display contents of the ROOT file 'example.root', use a long listing format and print trees recursively.
+
+- rootls -r example.root
+  Display contents of the ROOT file 'example.root', traversing recursively any TDirectory.
+)";
+
+
+struct SplitPath {
+  std::vector<std::string> fPathFragments;
+};
+
+struct RootLsSource {
+  std::string fFileName;
+  // List of object paths that match the query
+  std::vector<SplitPath> fObjectPaths;
+};
+
+static void RootLs(const std::vector<RootLsSource> &sourceFiles)
+{
+  int indent = 2 * (sourceFiles.size() > 1);
+  for (const auto &source : sourceFiles) {
+    std::cout << source.fFileName << "\n";
+    for (const auto &path : source.fObjectPaths) {
+      int ind = indent + 2;
+      for (const auto &frag : path.fPathFragments) {
+        for (int i = 0; i < ind; ++i)
+          std::cout << ' ';
+        std::cout << frag << "\n";
+        ind += 2;
+      }
+    }
+  }
+}
+
+static bool MatchesGlob(std::string_view haystack, std::string_view pattern)
+{
+  return wildcards::match(haystack, pattern);
+}
+
+static std::vector<SplitPath> GetMatchingPathsInFile(std::string_view fileName, std::string_view pattern)
+{
+  std::vector<SplitPath> result;
+  
+  auto file = std::unique_ptr<TFile>(TFile::Open(std::string(fileName).c_str(), "READ_WITHOUT_GLOBALREGISTRATION"));
+  if (!file)
+    return result;
+
+  // @Speed: avoid allocating
+  const auto patternSplits = ROOT::Split(pattern, "/");
+
+  // Match all objects at all nesting levels
+  struct DirLevel {
+    TDirectory *fDir;
+    // @Speed this is super inefficient!
+    std::vector<std::string> fAncestorPaths;
+  };
+  std::vector<DirLevel> dirsToVisit { {file.get(), {}} };
+  for (const auto &pat : patternSplits) {
+    std::vector<DirLevel> newDirsToVisit;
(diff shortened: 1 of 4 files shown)


ferdymercury:
Thanks a lot for this! I've always complained silently about that latency.
Side question: do you also plan to port "rootprint" to C++? (If yes, there is an issue with it which I was not able to figure out how to properly solve, see https://github.com/root-project/root/issues/13659)

silverweed:
@ferdymercury Ideally I would like to port all root executables to C++, but I don't have a precise timeline (this is not part of my main tasks so I can only work on this sparingly). I'll keep that in mind when/if I get to work on rootprint, thanks!

main/src/rootls.cxx
```diff
@@ -0,0 +1,743 @@
...
+   }
+   return {columns, rows};
+#endif
+   // Fallback
+   return {80, 25};
+}
+
+using Indent = int;
+
+static void PrintIndent(std::ostream &stream, Indent indent)
```
vepadulano, line 207: Subjective: maybe prefer using an unnamed namespace for all these utils?

main/src/rootls.cxx
```diff
@@ -0,0 +1,743 @@
...
+#if defined(R__UNIX)
+#include <sys/ioctl.h>
+#include <unistd.h>
+#elif defined(R__WIN32)
+#define WIN32_LEAN_AND_MEAN
+#define VC_EXTRALEAN
+#include <windows.h>
+#endif
+
+static const char *const kAnsiNone = "\x1B[0m";
```
vepadulano, line 46: subjective: I would prefer `constexpr auto`.

main/src/rootls.cxx
```diff
@@ -0,0 +1,743 @@
...
+   std::ios defaultFmt(nullptr);
+   stream << month << ' ';
+   stream << std::right << std::setfill('0') << std::setw(2) << datime.GetDay() << ' ';
+   stream << std::setw(2) << datime.GetHour() << ':' << datime.GetMinute() << ' ' << datime.GetYear() << ' ';
+   stream.copyfmt(defaultFmt);
+}
+
+// NOTE: T may be a TTree or a TBranch
+template <typename T>
+static void PrintTTree(std::ostream &stream, T &tree, Indent indent)
```
vepadulano, line 229: Why not just
```suggestion
static void PrintTTree(std::ostream &stream, TTree &tree, Indent indent)
```

main/src/rootls.cxx
```diff
@@ -0,0 +1,743 @@
...
+         if (ClassInheritsFrom(child.fClassName.c_str(), "TTree")) {
+            TTree *tree = child.fKey->ReadObject<TTree>();
+            if (tree) {
+               PrintTTree(stream, *tree, Indent(indent + 2));
+               PrintClusters(stream, *tree, Indent(indent + 2));
+            }
+         }
+      }
+      if ((flags & RootLsArgs::kRecursiveListing) && ClassInheritsFrom(child.fClassName.c_str(), "TDirectory")) {
+         // @Recursion
```
vepadulano: Leftover comment?

main/src/rootls.cxx
```diff
@@ -0,0 +1,743 @@
...
+   int rows = 0, columns = 0;
+   CONSOLE_SCREEN_BUFFER_INFO csbi;
+   if (::GetConsoleScreenBufferInfo(::GetStdHandle(STD_OUTPUT_HANDLE), &csbi)) {
+      columns = csbi.srWindow.Right - csbi.srWindow.Left + 1;
+      rows = csbi.srWindow.Bottom - csbi.srWindow.Top + 1;
+   }
+   return {columns, rows};
+#endif
+   // Fallback
+   return {80, 25};
```
vepadulano: Doesn't this produce two `return` statements at the same scope on Windows?

vepadulano (commented):
I think this is a great initiative, thanks!

main/src/rootls.cxx
silverweed: Because it needs to work also with TBranch which is not a relative of TTree as far as I understand

main/src/rootls.cxx
silverweed: It's actually intentional - I could remove it if you prefer

main/src/rootls.cxx
silverweed: Indeed, thanks for spotting it

main/src/rootls.cxx
vepadulano: My bad, for some reason I read `TTree or TChain` 😓

main/src/rootls.cxx
silverweed: I prefer `static` because it is more "locally readable" (can immediately see which functions are internal without looking for the start or end of the namespace). I could add it for the types but since this is a small standalone executable it's probably unnecessary.

main/src/rootls.cxx
vepadulano: I'm not against it, it just looks a bit incomplete or just placed randomly. If you're using it as a reminder of the places that introduce recursion, I guess it would be nice to give it more context. For example, by writing some doc somewhere in this file and reference it via some kind of tag in every place.

guitargeek:
> The executable is currently called rootlscxx and the original rootls is preserved, but ideally it should be phased out after verifying that everything is covered.

You have a plan to do that phase-out?

vepadulano:
I don't think we should bloat the command line utilities set with another command. I think these changes should be transparent, what is now called `rootlscxx` should just substitute the current `rootls` implementation with no extra command line utility being generated by the ROOT build

main/src/rootls.cxx
```diff
@@ -0,0 +1,762 @@
...
+   PrintUsage fPrintUsageAndExit = PrintUsage::kNo;
+};
+
+struct V2i {
+   int x, y;
+};
+
+static V2i GetTerminalSize()
+{
+   int rows = 80, columns = 25;
```
vepadulano: I think these values may have been inadvertently exchanged

main/src/rootls.cxx
silverweed: I'm never quite sure what `auto` becomes with string literals (I assume `const char *`) so I prefer being explicit in these cases

silverweed:
> I don't think we should bloat the command line utilities set with another command. I think these changes should be transparent, what is now called rootlscxx should just substitute the current rootls implementation with no extra command line utility being generated by the ROOT build

I agree on this, but should we have some kind of transition period (maybe renaming the old rootls to something else) or is it fine to just replace it in one go? In this second case, should it be a separate PR?

vepadulano:
> I agree on this, but should we have some kind of transition period (maybe renaming the old rootls to something else) or is it fine to just replace it in one go? In this second case, should it be a separate PR?

My personal opinion is that these changes are all about the implementation. From the point of view of the interface and the final results nothing changes. So, when I think about a transition period between the old implementation of rootls and the new one, it just sounds to me like we're saying we don't trust the new implementation. In practice, once we have the same coverage of the old implementation with the new one, I don't see why we should have a transition period anyway since it doesn't require any user-facing changes anyway.

main/src/rootls.cxx
pcanal: Maybe it was due to the variable name 'tree' which does not quite advertise the fact that the input could be a branch! :)

Issue #12208, opened 2023-02-02
Deprecating RtypesCore.h


klenze:
Dear ROOT devs,
as you may be aware, recent developments in the standardization of C++ brought the C99 **stdint** to C++11.

When the original Rtypes.h was written in the nineties, there may have been some need for making sure that sizeof(int) was at least 4. Today, some 37 years after the Intel 80386, I suspect that the only systems where ``sizeof(int)==2`` are 16 bit microcontrollers which lack _any_ 32 bit data types and are probably a bit too constrained with regard to addressable memory to run ROOT. 

[RtypesCore.h](https://github.com/root-project/root/blob/master/core/foundation/inc/RtypesCore.h) seems to be a bit outdated. 

From my understanding, Float16_t and Double32_t are meant to warn the user that parts of the mantissa are truncated. Given the prevalence of stdint (where the number denotes the stored size in bits), I am unsure if that meaning is successfully transported. Float7BitMantissa_t might be more verbose. 

Regarding the limits, I notice that for most types, the listed values are the minimum/maximum value that type can take, for example:
```const UInt_t    kMaxUInt     = UInt_t(~0);```
(NB: on a system where ``sizeof(int)==2``, the compiler is (afaik) free to interpret 0 as int, whose complement will then be disappointingly small.)
Under this interpretation, I want to cast doubt on 
```
const Int_t     kMaxUChar    = 256;
const Int_t     kMaxUShort   = 65534;
```
While there were some promising publications of computer scientists claiming to have succeeded in storing 256 in an unsigned char, later attempts to reproduce that research found that the bit sequence was indistinguishable from zero. The general consensus among computer scientists today is that the highest value which can be stored in a uint8_t is in fact 255. On the other hand, a few years later initially doubtful claims that 65535 could be stored in an unsigned short *were* replicated. Today, big tech (e.g. GAFAM) routinely store that value in their 16 bit numbers and even smaller startups are picking up on the trend. ``</sarcasm>``

(Also, while the kMaxUShort error is rendered harmless by the bitshift in kMaxShort, the kMaxUChar error means that kMinChar and kMaxChar are wrong. In fact, Char_t(kMinChar)>Char_t(kMaxChar) is true. However, nine out of ten math professors agree that the maximum of a set of values should be at least as large as the minimum of that set. ``[[citation needed]]``)

[...]

PR #19283, opened 2025-07-04
[skip-ci][core] Discourage in the docu the use of old-style data types from RtypesCore and fix wrong docu of truncated types

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/12208

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/core/foundation/inc/RtypesCore.h
+++ b/core/foundation/inc/RtypesCore.h
@@ -31,6 +31,7 @@
  * Likewise with `ULong_t`.
  * \warning In some architectures, `std::int64_t` may have a different underlying data type (long vs int) than in others
  * and may lead to a different StreamerInfo than in others, thus it might be convenient to use (U)Long64_t instead.
+ * Hence, full fledge embrace of the std::types is actually waiting on ROOT I/O to be extended to support them explicitly.
  */
 
 #include <ROOT/RConfig.hxx>


core/foundation/inc/RtypesCore.h
```diff
@@ -19,6 +19,17 @@
 //                                                                      //
 //////////////////////////////////////////////////////////////////////////
 
+/**
+ * \file
+ * \brief Basic types used by ROOT and required by TInterpreter.
+ * It ensures a portable fixed data type size across systems, since in
+ * the early days, sizeof(int) could be 2 or 4 depending on the architecture.
+ * \deprecated For future designs, unless for very specific needs, consider using instead standard
+ * classes from <cstdint> such as `std::int16_t`, `std::int32_t`, etc. It might be ok
```
silverweed: What's the "It might be ok" there for? Is is a truncated sentence?

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
-typedef const char     Option_t;    //Option string (const char)
-typedef int            Ssiz_t;      //String size (int)
-typedef float          Real_t;      //TVector and TMatrix element type (float)
-typedef long long          Long64_t; //Portable signed long integer 8 bytes
-typedef unsigned long long ULong64_t;//Portable unsigned long integer 8 bytes
+typedef float          Float_t;     ///< Float 4 bytes (float) \deprecated Consider replacing with `float`.
+typedef float          Float16_t;   ///< Float 4 bytes in memory, written to disk as a 3 bytes (24-bits, not 16-bits) float with a truncated (up-to) 16-bit mantissa (12-bit by default in memory), and (7+1)-bits of exponent \deprecated Consider (non-equivalent) alternatives in `<cstdfloat>`.
+typedef double         Double_t;    ///< Double 8 bytes \deprecated Consider replacing with `double`.
+typedef double         Double32_t;  ///< Double 8 bytes in memory, written to disk as a 3 bytes (24-bits, not 32-bits) float with a truncated (up-to) 16-bit mantissa (24-bit by default in memory), and (7+1)-bits of exponent \deprecated Consider (non-equivalent) alternatives in `<cstdfloat>`.
```
ferdymercury: This needs to be confirmed

ferdymercury: See also https://root-forum.cern.ch/t/float16-t-really-encoded-on-16-bits/37518

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
-typedef float          Real_t;      //TVector and TMatrix element type (float)
-typedef long long          Long64_t; //Portable signed long integer 8 bytes
-typedef unsigned long long ULong64_t;//Portable unsigned long integer 8 bytes
+typedef float          Float_t;     ///< Float 4 bytes (float) \deprecated Consider replacing with `float`.
+typedef float          Float16_t;   ///< Float 4 bytes in memory, written to disk as a 3 bytes (24-bits, not 16-bits) float with a truncated (up-to) 16-bit mantissa (12-bit by default in memory), and (7+1)-bits of exponent \deprecated Consider (non-equivalent) alternatives in `<cstdfloat>`. \warning Do not mix this with a half-float such as std::float16_t
+typedef double         Double_t;    ///< Double 8 bytes \deprecated Consider replacing with `double`.
+typedef double         Double32_t;  ///< Double 8 bytes in memory, written to disk as a 3 bytes (24-bits, not 32-bits) float with a truncated (up-to) 16-bit mantissa (24-bit by default in memory), and (7+1)-bits of exponent \deprecated Consider (non-equivalent) alternatives in `<cstdfloat>`. \warning Do not mix this with a single-precision float std::float32_t
```
pcanal:
```suggestion
typedef double         Double32_t;  ///< Double 8 bytes in memory, written to disk as a 4 bytes by default, can be customized to store with less bytes.
```

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
-typedef bool           Bool_t;      //Boolean (0=false, 1=true) (bool)
-typedef unsigned char  Byte_t;      //Byte (8 bits) (unsigned char)
-typedef short          Version_t;   //Class version identifier (short)
-typedef const char     Option_t;    //Option string (const char)
-typedef int            Ssiz_t;      //String size (int)
-typedef float          Real_t;      //TVector and TMatrix element type (float)
-typedef long long          Long64_t; //Portable signed long integer 8 bytes
-typedef unsigned long long ULong64_t;//Portable unsigned long integer 8 bytes
+typedef float          Float_t;     ///< Float 4 bytes (float) \deprecated Consider replacing with `float`.
+typedef float          Float16_t;   ///< Float 4 bytes in memory, written to disk as a 3 bytes (24-bits, not 16-bits) float with a truncated (up-to) 16-bit mantissa (12-bit by default in memory), and (7+1)-bits of exponent \deprecated Consider (non-equivalent) alternatives in `<cstdfloat>`. \warning Do not mix this with a half-float such as std::float16_t
```
pcanal: This is either too much or not enough details.  The actual size on disk depends on the customization.

core/foundation/inc/RtypesCore.h
ferdymercury: Why 4 bytes? It is streamed by default as a UShort and a UChar_t, isn't that 3 bytes?

core/foundation/inc/RtypesCore.h
ferdymercury: But isn't the default 3 bytes? Since you are streaming a UChar_t and a UShort_t ?

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
-typedef short          Font_t;      //Font number (short)
-typedef short          Style_t;     //Style number (short)
-typedef short          Marker_t;    //Marker number (short)
-typedef short          Width_t;     //Line width (short)
-typedef short          Color_t;     //Color number (short)
-typedef short          SCoord_t;    //Screen coordinates (short)
-typedef double         Coord_t;     //Pad world coordinates (double)
-typedef float          Angle_t;     //Graphics angle (float)
-typedef float          Size_t;      //Attribute size (float)
+typedef double         Axis_t;      ///< Axis values type (double) \deprecated Consider replacing with `double`
```
pcanal: Replacing this series of types with their typedefs makes the code less readable by removing the semantic part of the type .... I do not think we should deprecated them.

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
-typedef unsigned short UShort_t;    //Unsigned Short integer 2 bytes (unsigned short)
+typedef char           Char_t;      ///< Signed Character 1 byte (char) \deprecated Consider replacing with `char` or `std::int8_t`
+typedef unsigned char  UChar_t;     ///< Unsigned Character 1 byte (unsigned char) \deprecated Consider replacing with `unsigned char` or `std::uint8_t`
+typedef short          Short_t;     ///< Signed Short integer 2 bytes (short) \deprecated Consider replacing with `short` or `std::int16_t`
+typedef unsigned short UShort_t;    ///< Unsigned Short integer 2 bytes (unsigned short) \deprecated Consider replacing with `unsigned short` or `std::uint16_t`
 #ifdef R__INT16
-typedef long           Int_t;       //Signed integer 4 bytes
-typedef unsigned long  UInt_t;      //Unsigned integer 4 bytes
+typedef long           Int_t;       ///< Signed integer 4 bytes \deprecated Consider replacing with `std::int16_t`
+typedef unsigned long  UInt_t;      ///< Unsigned integer 4 bytes \deprecated Consider replacing with `std::int16_t`
```
pcanal:
```suggestion
typedef long           Int_t;       ///< Signed integer 4 bytes \deprecated Consider replacing with `std::int32_t`
typedef unsigned long  UInt_t;      ///< Unsigned integer 4 bytes \deprecated Consider replacing with `std::int32_t`
```

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
+typedef const char     Option_t;    ///< Option string (const char) \deprecated Consider replacing with `const char`.
+typedef int            Ssiz_t;      ///< String size (int) \deprecated Consider replacing with `int`.
+typedef float          Real_t;      ///< TVector and TMatrix element type (float) \deprecated Consider replacing with `float`.
+typedef long long          Long64_t; ///< Portable signed long integer 8 bytes \deprecated Consider replacing with `long long` or `std::int64_t`.
+typedef unsigned long long ULong64_t;///< Portable unsigned long integer 8 bytes \deprecated Consider replacing with `unsigned long long` or `std::uint64_t`.
 #ifdef _WIN64
-typedef long long      Longptr_t;   //Integer large enough to hold a pointer
-typedef unsigned long long ULongptr_t;  //Unsigned integer large enough to hold a pointer
+typedef long long      Longptr_t;     ///< Integer large enough to hold a pointer \deprecated Consider replacing with `long long` on Windows.
+typedef unsigned long long ULongptr_t;///< Unsigned integer large enough to hold a pointer \deprecated Consider replacing with `unsigned long long` on Windows.
```
pcanal:
```suggestion
typedef unsigned long long ULongptr_t;///< Unsigned integer large enough to hold a pointer \deprecated Consider replacing with `size_t`.
```

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
+typedef short          Version_t;   ///< Class version identifier (short) \deprecated Consider replacing with `short` or `std::int16_t`.
+typedef const char     Option_t;    ///< Option string (const char) \deprecated Consider replacing with `const char`.
+typedef int            Ssiz_t;      ///< String size (int) \deprecated Consider replacing with `int`.
+typedef float          Real_t;      ///< TVector and TMatrix element type (float) \deprecated Consider replacing with `float`.
+typedef long long          Long64_t; ///< Portable signed long integer 8 bytes \deprecated Consider replacing with `long long` or `std::int64_t`.
+typedef unsigned long long ULong64_t;///< Portable unsigned long integer 8 bytes \deprecated Consider replacing with `unsigned long long` or `std::uint64_t`.
 #ifdef _WIN64
-typedef long long      Longptr_t;   //Integer large enough to hold a pointer
-typedef unsigned long long ULongptr_t;  //Unsigned integer large enough to hold a pointer
+typedef long long      Longptr_t;     ///< Integer large enough to hold a pointer \deprecated Consider replacing with `long long` on Windows.
```
pcanal: > ///< .... on Windows.
> ///< .... on UNIX.

This is confusing.  The purpose of the typedef is to avoid having the user use a different type on different platform (where user includes our own code).

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
-constexpr Long64_t kMaxLong64 = Long64_t(kMaxULong64 >> 1);
-constexpr Long64_t kMinLong64 = -kMaxLong64 - 1;
+constexpr ULong64_t kMaxULong64 = ULong64_t(~0LL);         ///< \deprecated Consider replacing with `std::numeric_limits<unsigned long long>::max()` (or `std::uint64_t`)
+constexpr Long64_t kMaxLong64 = Long64_t(kMaxULong64 >> 1);///< \deprecated Consider replacing with `std::numeric_limits<long long>::max()` (or `std::int64_t`)
+constexpr Long64_t kMinLong64 = -kMaxLong64 - 1;           ///< \deprecated Consider replacing with `std::numeric_limits<long long>::lowest()` (or `std::int64_t`)
 
-constexpr ULong_t kBitsPerByte = 8;
-constexpr Ssiz_t kNPOS = ~(Ssiz_t)0;
+constexpr ULong_t kBitsPerByte = 8; ///< \deprecated Consider replacing with `CHAR_BIT` from `<climits>`
+constexpr Ssiz_t kNPOS = ~(Ssiz_t)0;///< \deprecated Consider replacing with `int(-1)` or `std::string::npos`
```
pcanal: Why deprecate this one?  It is the `TString` marker and it is confused to ask the user to hard code or to use a different marker (that may or may not stay the same number).

core/foundation/inc/RtypesCore.h
ferdymercury: But in that case, the alternative data type is not portable. Conclusion, we should not deprecate these ones?

tree/tree/src/TTree.cxx
```diff
@@ -100,9 +100,9 @@ It is strongly recommended to persistify those as objects rather than lists of l
    - `I` : a 32 bit signed integer (`Int_t`)
    - `i` : a 32 bit unsigned integer (`UInt_t`)
    - `F` : a 32 bit floating point (`Float_t`)
-   - `f` : a 21 bit floating point with truncated mantissa (`Float16_t`): 1 for the sign, 8 for the exponent and 12 for the mantissa. Can be customized with suffix `[min,max(,nbits)]`.
+   - `f` : a 24 bit floating point with truncated mantissa (`Float16_t`, stored as 3 bytes, occupies 4 bytes in memory): By default, in disk, only 21 bits are used: 1 for the sign, 8 for the exponent and 12 for the mantissa. Can be customized with suffix `[min,max(,nbits)] `where `nbits` is for the mantissa.
    - `D` : a 64 bit floating point (`Double_t`)
-   - `d` : a 32 bit truncated floating point (`Double32_t`): 1 for the sign, 8 for the exponent and 23 for the mantissa. Can be customized with suffix `[min,max(,nbits)]`.
+   - `d` : a 24 bit floating point with truncated mantissa (`Double32_t`, stored as 3 bytes, occupies 8 bytes in memory): By default, in disk, 1 bit is used for the sign, 8 for the exponent and 15 for the mantissa. Can be customized with suffix `[min,max(,nbits)]` where `nbits` is for the mantissa.
```
pcanal: This does not seem to be accurate/clear.  The default is literally to store the value the same way we store C++ float (4 bytes resulting in the truncation of the both the double's mantissa and exponent).

core/foundation/inc/RtypesCore.h
pcanal: If the alternative is not valid on all platform for 'all' circumstances, I don't think we should deprecate.

core/foundation/inc/RtypesCore.h
pcanal: Yes but I felt that the docstring was too long ... and indeed I missing the part:
```
\deprecated Consider (non-equivalent) alternatives in `<cstdfloat>`.
```
Which I don't understand.  Neither the deprecation (there is no replacement except for some part of RNTuple), nor proposing alternative that are not functionality equivalent.

On another note, I think rather than
```
\warning Do not mix this with a half-float such as std::float16_t
```
which would imply warning against `std::float16_t a; Float16_t b; auto c = a + b;`, we may be meaning:
```
\warning Do not confuse Float16_t on file representation with a half-float such as std::float16_t.
```
And then I tried to add a reminder about `Float16_t` being a real `float` in memory but the sentence get (even more) awkward.

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
-constexpr ULong64_t kMaxULong64 = ULong64_t(~0LL);
-constexpr Long64_t kMaxLong64 = Long64_t(kMaxULong64 >> 1);
-constexpr Long64_t kMinLong64 = -kMaxLong64 - 1;
+constexpr ULong64_t kMaxULong64 = ULong64_t(~0LL);         ///< \deprecated Consider replacing with `std::numeric_limits<unsigned long long>::max()` (or `std::uint64_t`)
+constexpr Long64_t kMaxLong64 = Long64_t(kMaxULong64 >> 1);///< \deprecated Consider replacing with `std::numeric_limits<long long>::max()` (or `std::int64_t`)
+constexpr Long64_t kMinLong64 = -kMaxLong64 - 1;           ///< \deprecated Consider replacing with `std::numeric_limits<long long>::lowest()` (or `std::int64_t`)
 
-constexpr ULong_t kBitsPerByte = 8;
-constexpr Ssiz_t kNPOS = ~(Ssiz_t)0;
+constexpr ULong_t kBitsPerByte = 8; ///< \deprecated Consider replacing with `CHAR_BIT` from `<climits>`
```
pcanal: Isn't there a`C++` version of `CHAR_BIT`?

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
-constexpr ULong64_t kMaxULong64 = ULong64_t(~0LL);
-constexpr Long64_t kMaxLong64 = Long64_t(kMaxULong64 >> 1);
-constexpr Long64_t kMinLong64 = -kMaxLong64 - 1;
+constexpr ULong64_t kMaxULong64 = ULong64_t(~0LL);         ///< \deprecated Consider replacing with `std::numeric_limits<unsigned long long>::max()` (or `std::uint64_t`)
+constexpr Long64_t kMaxLong64 = Long64_t(kMaxULong64 >> 1);///< \deprecated Consider replacing with `std::numeric_limits<long long>::max()` (or `std::int64_t`)
+constexpr Long64_t kMinLong64 = -kMaxLong64 - 1;           ///< \deprecated Consider replacing with `std::numeric_limits<long long>::lowest()` (or `std::int64_t`)
 
-constexpr ULong_t kBitsPerByte = 8;
-constexpr Ssiz_t kNPOS = ~(Ssiz_t)0;
+constexpr ULong_t kBitsPerByte = 8; ///< \deprecated Consider replacing with `std::numeric_limits<unsigned char>::digits`, or with `CHAR_BIT` or `CHAR_WIDTH` from `<climits>`.
```
pcanal:
```suggestion
constexpr ULong_t kBitsPerByte = 8; ///< \deprecated Consider replacing with `std::numeric_limits<unsigned char>::digits`.
```

tree/tree/src/TTree.cxx
```diff
@@ -100,9 +100,9 @@ It is strongly recommended to persistify those as objects rather than lists of l
...
-   - `f` : a 21 bit floating point with truncated mantissa (`Float16_t`): 1 for the sign, 8 for the exponent and 12 for the mantissa. Can be customized with suffix `[min,max(,nbits)]`.
+   - `f` : a 24 bit (or 32) floating point with truncated mantissa (`Float16_t`, stored as 3 bytes by default or as fixed-point arithmetic 4 bytes Int_t if range is customized; occupies 4 bytes in memory): By default, in disk, only 21 bits are used: 1 for the sign, 8 for the exponent and 12 for the mantissa. Can be customized with suffix `[min,max(,nbits)] `where `nbits` is for the mantissa.
    - `D` : a 64 bit floating point (`Double_t`)
-   - `d` : a 32 bit truncated floating point (`Double32_t`): 1 for the sign, 8 for the exponent and 23 for the mantissa. Can be customized with suffix `[min,max(,nbits)]`.
+   - `d` : a 32 (or 24) bit floating point with truncated mantissa (`Double32_t`, stored as a 4 bytes Float_t by default or as 3 bytes if range is customized; occupies 8 bytes in memory): By default, in disk, 1 bit is used for the sign, 7 for the exponent and 24 for the mantissa. Can be customized to 3 bytes (24 bits) with suffix `[min,max(,nbits)]` where `nbits` is for the mantissa.
```
pcanal:
```suggestion
   - `d` : a 32 (or 24) bit floating point with truncated mantissa (`Double32_t`, stored as a 4 bytes Float_t by default or as 3 bytes if range is customized; occupies 8 bytes in memory): By default, in disk, 1 bit is used for the sign, 8 for the exponent and 23 for the mantissa. Can be customized to 3 bytes (24 bits) with suffix `[min,max(,nbits)]` where `nbits` is for the mantissa.
```
As far as I remember the [IEEE 32bits floating point ](https://en.wikipedia.org/wiki/Single-precision_floating-point_format)(and [C++ ](https://en.cppreference.com/w/cpp/language/types.html#Standard_floating-point_types)) mandates the float representation to have 8 bits for the exponent.

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
+typedef double         Double_t;    ///< Double 8 bytes \deprecated Consider replacing with `double`.
+typedef double         Double32_t;  ///< Double 8 bytes in memory, written to disk as a 4 bytes Float_t (32-bits) by default, or as 3 bytes (24-bits) float if range is customized, with a truncated mantissa (24-bit by default in memory, less if range is customized), and (7+1)-bits of exponent \warning Do not confuse Double32_t on file representation with a single-precision float such as std::float32_t
+typedef long double    LongDouble_t;///< Long Double (not portable) \deprecated Consider replacing with `long double`.
+typedef char           Text_t;      ///< General string (char) \deprecated Consider replacing with `char`.
+typedef bool           Bool_t;      ///< Boolean (0=false, 1=true) (bool) \deprecated Consider replacing with `bool`.
+typedef unsigned char  Byte_t;      ///< Byte (8 bits) (unsigned char) \deprecated Consider replacing with `unsigned char` or `std::byte`.
+typedef short          Version_t;   ///< Class version identifier (short) \deprecated Consider replacing with `short` or `std::int16_t`.
```
pcanal:
```suggestion
typedef short          Version_t;   ///< Class version identifier (short)
```
This is a semantic typedef.  (It matters that it is a Version Number, it does not matter (to the user) that it is a short).

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
+typedef const char     Option_t;    ///< Option string (const char) \deprecated Consider replacing with `const char`.
+typedef int            Ssiz_t;      ///< String size (int) \deprecated Consider replacing with `int`.
```
pcanal:
```suggestion
typedef int            Ssiz_t;      ///< String size (int) \deprecated Consider replacing with `std::size_t`.
```
except no it is not the same ... so 
```suggestion
typedef int            Ssiz_t;      ///< String size (currently int)
```
because it is semantic (similar to but different than `size_t`)

core/foundation/inc/RtypesCore.h
```diff
@@ -19,6 +19,17 @@
...
 //////////////////////////////////////////////////////////////////////////
 
+/**
+ * \file
+ * \brief Basic types used by ROOT and required by TInterpreter.
+ * It ensures a portable fixed data type size across systems, since in
+ * the early days, sizeof(int) could be 2 or 4 depending on the architecture.
+ * \deprecated For future designs, unless for very specific needs, consider using instead standard
+ * fixed-width classes from <cstdint> such as `std::int16_t`, `std::int32_t` or from <cstdfloat> for floating types.
+ * \warning `Long_t` has not the same width across platforms, so it should be avoided if portability is envisioned.
```
pcanal, line 29:
```suggestion
 * \warning `Long_t` has not the same width across platforms, so it should be avoided if portability is envisioned.
```
Technically one need to also avoid `std::int64_t` which on some platforms resolve to int and some platform resolve to long, so long different so ROOT I/O (and thus lead to a different StreamerInfo for the same class on the 2 platforms).  If I remember correctly `std::in32_t` always resolve to `int`, so might be fine.

So full fledge embrace of the std::type is actually waiting on ROOT I/O to be extended to support them explicitly.

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
-typedef int            Int_t;       //Signed integer 4 bytes (int)
-typedef unsigned int   UInt_t;      //Unsigned integer 4 bytes (unsigned int)
+typedef int            Int_t;       ///< Signed integer 4 bytes (int) \deprecated Consider replacing with `std::int32_t`
+typedef unsigned int   UInt_t;      ///< Unsigned integer 4 bytes (unsigned int) \deprecated Consider replacing with `std::uint32_t`
 #endif
 #ifdef R__B64    // Note: Long_t and ULong_t are currently not portable types
-typedef int            Seek_t;      //File pointer (int)
-typedef long           Long_t;      //Signed long integer 8 bytes (long)
-typedef unsigned long  ULong_t;     //Unsigned long integer 8 bytes (unsigned long)
+typedef int            Seek_t;      ///< File pointer (int). Size depends on architecture \deprecated Consider replacing with `int`.
```
pcanal:
```suggestion
typedef int            Seek_t;      ///< File pointer (int).
```
This is a semantic typedef.

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
+typedef long           Long_t;      ///< Signed long integer 8 bytes (long). Size depends on architecture \deprecated Consider replacing with `long`
+typedef unsigned long  ULong_t;     ///< Unsigned long integer 8 bytes (unsigned long). Size depends on architecture \deprecated Consider replacing with `unsigned long`
 #else
-typedef int            Seek_t;      //File pointer (int)
-typedef long           Long_t;      //Signed long integer 4 bytes (long)
-typedef unsigned long  ULong_t;     //Unsigned long integer 4 bytes (unsigned long)
+typedef int            Seek_t;      ///< File pointer (int). Size depends on architecture \deprecated Consider replacing with `int`.
```
pcanal:
```suggestion
typedef int            Seek_t;      ///< File pointer (int). 
```

core/foundation/inc/RtypesCore.h
```diff
@@ -34,91 +45,91 @@ class TRootIOCtor;
...
+typedef double         Double32_t;  ///< Double 8 bytes in memory, written to disk as a 4 bytes Float_t (32-bits) by default, or as 3 bytes (24-bits) float if range is customized, with a truncated mantissa (24-bit by default in memory, less if range is customized), and (7+1)-bits of exponent \warning Do not confuse Double32_t on file representation with a single-precision float such as std::float32_t
+typedef long double    LongDouble_t;///< Long Double (not portable) \deprecated Consider replacing with `long double`.
+typedef char           Text_t;      ///< General string (char) \deprecated Consider replacing with `char`.
+typedef bool           Bool_t;      ///< Boolean (0=false, 1=true) (bool) \deprecated Consider replacing with `bool`.
+typedef unsigned char  Byte_t;      ///< Byte (8 bits) (unsigned char) \deprecated Consider replacing with `unsigned char` or `std::byte`.
+typedef short          Version_t;   ///< Class version identifier (short) \deprecated Consider replacing with `short` or `std::int16_t`.
+typedef const char     Option_t;    ///< Option string (const char) \deprecated Consider replacing with `const char`.
```
pcanal: Used mostly semantically ...

pcanal (approved):
Thanks.

PR #19284, opened 2025-07-04
Revert "[roottest] Remove sequential compilation with Ninja."

hahnjo:
This reverts commit e8519e4a8011f199fd7031400962b7153ba8368c.

Since a few weeks, I experienced serveral corrupted incremental builds and tests. I strongly suspect this commit to be the reason. Indeed launching Ninja concurrently that need to build a dependency will execute the same command multiple times, leading to potentially corrupted files. In principle, the same problem applies to Makefiles but there we use the /fast targets and ignore outdated dependencies.


hageboeck:
> Since a few weeks, I experienced serveral corrupted incremental builds and tests. I strongly suspect this commit to be the reason.

We can try reverting this commit, but note that it was merged last week, so it can't be causing something that you are seeing for several weeks. This commit will only affect builds that are part of a `ctest -j`.

hageboeck (approved):
Let's try if this fixes the problem.

hahnjo:
> > Since a few weeks, I experienced serveral corrupted incremental builds and tests. I strongly suspect this commit to be the reason.
> 
> We can try reverting this commit, but note that it was merged last week, so it can't be causing something that you are seeing for several weeks. This commit will only affect builds that are part of a `ctest -j`.

As discussed offline, possible that it's only since 1.5 weeks :sweat_smile: 

I did the test we discussed: a full build, then `touch CMakeLists.txt` and then directly a `ctest -j12`. This indeed corrupted the build as much as manually needing to call `cmake .` and then wanting to rebuild at least all of `core/`. In the past, I also had it rebuilding LLVM but this may depend on exact timing of the race.

Issue #19022, opened 2025-06-12
Can not access REntry fields with DataVector in python


Nowakus:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

An error is reported when trying to access REntry fields of type related to DataVector (a class with a default template argument) in python:

```
entry["HLTNav_RepackedFeatures_MET"]      # field of type: DataVector<xAOD::TrigMissingET_v1>

 File "/cvmfs/sft-nightlies.cern.ch/lcg/nightlies/dev3/Tue/ROOT/HEAD/x86_64-el9-gcc13-opt/lib/ROOT/_pythonization/_rntuple.py", line 28, in _REntry_getitem
    ptr_proxy = self._CallGetPtr(key)
 File "/cvmfs/sft-nightlies.cern.ch/lcg/nightlies/dev3/Tue/ROOT/HEAD/x86_64-el9-gcc13-opt/lib/ROOT/_pythonization/_rntuple.py", line 24, in _REntry_CallGetPtr
    return self._GetPtr[fieldType](key)
TypeError: Could not find "GetPtr<DataVector<xAOD::TrigMissingET_v1>>" (set cppyy.set_debug() for C++ errors):
  Failed to instantiate "GetPtr<DataVector<xAOD::TrigMissingET_v1>>(::ROOT::RFieldToken&)"

```
### Reproducer

```
# standard ATLAS Athena setup:
export ATLAS_LOCAL_ROOT_BASE=/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase
source $ATLAS_LOCAL_ROOT_BASE/user/atlasLocalSetup.sh
asetup main--dev3LCG,latest,Athena,gcc14
python

import ROOT
rnt = ROOT.RNTupleReader.Open("EventData", "/afs/cern.ch/user/m/mnowak/public/DAOD_PHYS.rntuple.pool.root")
entry = rnt.CreateEntry()
rnt.LoadEntry(1, entry)
entry["HLTNav_RepackedFeatures_MET"]

```

### ROOT version

master

### Installation method

LCG dev3

### Operating system

Linux

maszyman:
Just to expand on what @Nowakus wrote, the use case is to read all the fields, something along the lines:

```py
reader = ROOT.RNTupleReader.Open(
    ROOT.TFile.Open(fileName).Get("MetaData")
)
entry = reader.CreateEntry()
reader.LoadEntry(0, entry)

[...]

PR #19286, opened 2025-07-04
An option for the second part of #19022

vepadulano:
This PR provides a way to fix the Python-only side of #19022, see the second commit for a description.


vepadulano:
Confirming that these changes fix the issue also on lxplus with an ATHENA environment and a custom build of ROOT
```
[vpadulan@lxplus997 ~/rntuple-atlas-datavector]$cat repro_mnowak.py
import ROOT
rnt = ROOT.RNTupleReader.Open("EventData", "/afs/cern.ch/user/m/mnowak/public/DAOD_PHYS.rntuple.pool.root")
entry = rnt.CreateEntry()
rnt.LoadEntry(1, entry)
val = entry["HLTNav_RepackedFeatures_MET"]
print(f"Got value {val} of type {type(val)}")
[vpadulan@lxplus997 ~/rntuple-atlas-datavector]$python repro_mnowak.py
Got value {} of type <class cppyy.gbl.DataVector<xAOD::TrigMissingET_v1> at 0xf4195d0>
```

dpiparo (approved):
Thanks for these changes. I would propose to integrate them and backport them to 6.36 at your earliest convenience. This will be the trigger for the release of 6.36.02.

vepadulano:
I think before merging I need to understand why the Windows builds are still red, to avoid having a broken CI

bellenot:
> I think before merging I need to understand why the Windows builds are still red, to avoid having a broken CI

I just had a look and restarted the builds. We'll see...

vepadulano:
I've tried something, let's see if it fixes the Windows CI.

I wanted to specify a general note related to these changes. Even if they work for the reproducer at #19022, there are advantages and disadvantages of choosing this route.

One advantage of this approach is that it does not require any change in the RNTuple library and in cppyy. Also, it does not involve any change in behaviour in core, the new API is completely detached from any pre-existing code paths. One disadvantage is that in order to actually trigger the right function template instantiation there needs to be some manual work, currently done in the Pythonization of `REntry::GetPtr`. In general, if a user were to write a C++ template function that needs to take in a template argument type of the same kind of AtlasLikeDataVector, they would need to try the first template instantiation, see it failing, then retry again after querying the list of alternative class names. This workflow is  exemplified with a test (see `template_instantiation.py`).

Making this approach fully generic would require changes in cppyy which on the surface look quite invasive. For example if a function has multiple template arguments and all of them have multiple alternative class names, one would probably need to trigger all the combinations of template instantiations to look for the right one.

This has been added to the commit message.

dpiparo:
I approved this changes since they have been requested by ATLAS, not only through an issue to GH, but also in the prioritised list given in input to the second Quarterly Report.

vepadulano:
The latest commit fixes the Windows build, I am now rebasing it on top of the existing one to clean the history.

core/cont/src/TClassTable.cxx
```diff
@@ -1030,3 +1048,22 @@ TNamed *ROOT::RegisterClassTemplate(const char *name, const char *file,
    }
    return reg;
 }
+
+std::vector<std::string> TClassTable::GetClassAlternativeNames(const char *cname)
+{
+   UInt_t slot = ROOT::ClassTableHash(cname, fgSize);
```
pcanal, line 1056: I think this might be missing the local lock acquisition (see usage of `lock(GetClassTableMutex());`)

core/cont/src/TClassTable.cxx
```diff
@@ -1030,3 +1048,22 @@ TNamed *ROOT::RegisterClassTemplate(const char *name, const char *file,
    }
    return reg;
 }
+
+std::vector<std::string> TClassTable::GetClassAlternativeNames(const char *cname)
```
pcanal, line 1052: Minor/optimization for later:  This works but requires creating a vector and creating/copying string at each invocation; there might or might not be a way to avoid them (but then again there might be challenge involving other thread modifying the TClassTable (set of collections).

pcanal (approved):
LGTM but there appears to be a missing lock.

core/cont/src/TClassTable.cxx
vepadulano: I thought about giving pointers to the stored class names, but I don't find any indication in the code that they are guaranteed to be owned by the `fgTable` or `fgAlternate` elements.

pcanal: Actually they are meant (but I guess we can not guarantee that) to be 'static' string.  The code is (meant to) be generated by rootcling and looks like:
```
instance.AdoptAlternate(::ROOT::AddClassAlternate("vector<Cyclic>","std::__1::vector<Cyclic, std::__1::allocator<Cyclic>>"));
```

PR #19287, opened 2025-07-04
[roottest] use cmake for `root/io` tests [part5]

linev:
Major change is `io/tree`.
There many tests includes creation and dumping of tree content.
In Makefile it was sequential and dump into single file.
Now creation and reading of each tree run as  separated test via `run.C` macro
Provide individual ref files for each reading test
Use small cmake function to configure dependency via all these tests

Use fixtures for better dependencies in several other tests


roottest/root/io/vararyobj/CMakeLists.txt
```diff
@@ -1,7 +1,52 @@
...
+                  MACRO runvaryingArrayRead_51508.C+
+                  COPY_TO_BUILDDIR varyingArray_51508.root
+                  OUTREF varyingArrayRead_51508.good)
+
+# on the Windows there is no "tail" command, therefore one need
+# to modify test logic to be able compare output log files
+if(NOT MSVC OR win_broken_tests)
+    ROOTTEST_ADD_TEST(varyingArrayWrite
+                    MACRO runvaryingArrayWrite.C+
+                    OUTCNVCMD tail -120
```
pcanal, line 11: It appears that the only difference on the log file might file the script name.  We might be able to handle this with sed instead of tail.

linev: `tail -120` used to extract last 120 lines from test log file - which contains object printing.
And then log after writing compared with log after reading - using plain `diff` command.
Inside the tests many system `rand()` invocations are used - therefore one cannot create ref file directly.
Possible solution will be use of `gRandom` instead - then one can store ref files.

Up to now I just reproduce logic of old tests

pcanal: > Up to now I just reproduce logic of old tests

Where I got confused is that the output file is 128 lines and the 8 lines skipped did not seem remarkable.  Related it seemed that the original test was only comparing the last 3 lines.  I was just wondering if the `tail` was needed at all (while the comp between read and write rather than a static reference is still of course applicable).  But to compare the 2 file we would still need to transform the script name (`sed -e s:runvaryingArrayWrite:runvaryingArrayRead:`).

Either way the `tail -120` is similar to `tail -3` and a more significant change could indeed be handle in a different PR but I did not want to make sure we were not missing something (i.e. that `120` was indeed an arbitrary number between 3 and 126 :) ).

linev: There is some extra output in the beginning, which may vary. 
I guess that's why only last several lines was compared - and I just increase number of these lines

General comment - many tests where such extra output manipulation applied cannot be run on Windows.
Therefore in some bright future one need to analyze all such tests and provide solution for them.
So appearance of  `OUTCNVCMD` is indicator that one need to return to this test later

PR #999, opened 2017-09-15
Cling get fully qualified name use clang tooling

Axel-Naumann:

PR #19292, opened 2025-07-05
[cling] Update policy for `GetFullyQualifiedName` in cling utils

guitargeek:
Just like in CppInterOp, set the `SuppressElaboration` and `FullyQualifiedName` policies when getting the fully qualified name of a QualType.

https://its.cern.ch/jira/browse/ROOT-10150


tree/ntuple/test/ntuple_type_name.cxx
```diff
@@ -216,13 +216,13 @@ TEST(RNTuple, TypeNameTemplatesNestedAlias)
    ASSERT_EQ(2, hashSubfields.size());
    EXPECT_EQ("fHash", hashSubfields[0]->GetFieldName());
    EXPECT_EQ("std::string", hashSubfields[0]->GetTypeName());
-   EXPECT_EQ("EdmHash<1>::value_type", hashSubfields[0]->GetTypeAlias());
+   EXPECT_EQ("EdmHash::value_type", hashSubfields[0]->GetTypeAlias());
```
hahnjo, line 219: This change is incorrect: `EdmHash` is a template class and requires template arguments. Clang stores the inner class outside the template instantiation as an optimization, but the type must have arguments.

tree/ntuple/test/ntuple_type_name.cxx
```diff
@@ -216,13 +216,13 @@ TEST(RNTuple, TypeNameTemplatesNestedAlias)
...
 
    EXPECT_EQ("fHash2", hashSubfields[1]->GetFieldName());
    EXPECT_EQ("std::string", hashSubfields[1]->GetTypeName());
    // FIXME: This should really be EdmHash<1>::value_typeT<EdmHash<1>::value_type>, but this is the value we get from
    // TDataMember::GetFullTypeName right now...
-   EXPECT_EQ("value_typeT<EdmHash<1>::value_type>", hashSubfields[1]->GetTypeAlias());
+   EXPECT_EQ("EdmHash::value_typeT<EdmHash::value_type>", hashSubfields[1]->GetTypeAlias());
```
hahnjo, line 225: This is equally wrong, the expected outcome is even documented in the comment.

tree/ntuple/test/ntuple_type_name.cxx
```diff
@@ -257,7 +257,7 @@ TEST(RNTuple, ContextDependentTypeNames)
       {
          const auto &fdesc = desc.GetFieldDescriptor(desc.FindFieldId("m", fooId));
          EXPECT_EQ(fdesc.GetTypeName(), "std::vector<std::int32_t>");
-         EXPECT_EQ(fdesc.GetTypeAlias(), "MyVec<std::int32_t>");
+         EXPECT_EQ(fdesc.GetTypeAlias(), "CustomStruct::MyVec<std::int32_t>");
```
hahnjo, line 260: The old version may be correct, depending on how you argue: it's the type alias as the user wrote it, inside of `CustomStruct` which happens to have a `typedef`...

hahnjo (changes requested):

tree/ntuple/test/ntuple_type_name.cxx
guitargeek: Yep it's incorrect :( I realized the printing policies are quite brittle in general, and I can't make all cases work. I'll report on the problems I found soon, in a separate GitHub issue I guess.

ferdymercury:
Related: https://github.com/root-project/root/pull/999

guitargeek:
Superseeded by the following PR by @devajithvs, coincidentally opened later the same day, but independently:
  * https://github.com/root-project/root/pull/19293

Dev should be the better expert to bring this to conclusion :slightly_smiling_face:

PR #19293, opened 2025-07-05
[cling] Deduplicate cling::utils::TypeName::GetFullyQualifiedName

### LINKED ISSUE #18622: [cling] Deduplicate `cling::utils::TypeName::GetFullyQualifiedName`

It appears that (part of?) the code was included upstream in 2016 by commit https://github.com/llvm/llvm-project/commit/0dd191a5c4bf27cc8a2b6033436b00f0cbdc4ce7 (including typos in the comments...). By now, the code has probably diverged but it would be good to use the upstream version.

### ROOT version

`master`

### Installation method

from source

### Operating system

any

### Additional context

_No response_

devajithvs:
# This Pull request:

Work in Progress: Uses upstream `GetFullyQualifiedName` but has diverged quite a bit and there are failing tests locally

## Checklist:

- [ ] tested changes locally

This PR fixes https://github.com/root-project/root/issues/18622

### DIFF
--- a/interpreter/cling/lib/Utils/AST.cpp
+++ b/interpreter/cling/lib/Utils/AST.cpp
@@ -10,11 +10,12 @@
 #include "cling/Utils/AST.h"
 
 #include "clang/AST/ASTContext.h"
+#include "clang/AST/DeclTemplate.h"
 #include "clang/AST/DeclarationName.h"
 #include "clang/AST/GlobalDecl.h"
-#include "clang/Sema/Sema.h"
+#include "clang/AST/QualTypeNames.h"
 #include "clang/Sema/Lookup.h"
-#include "clang/AST/DeclTemplate.h"
+#include "clang/Sema/Sema.h"
 
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/StringRef.h"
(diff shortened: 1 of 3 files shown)


guitargeek:
Ha, that's funny! I was just touching the same code because there was an easy Jira issue to fix:
  * https://github.com/root-project/root/pull/19292

Maybe you can take the unit test from that PR and put it here too? The reproducer should also be fixed by what you're doing here. Or we merge my PR before? As you prefer.

devajithvs:
> Ha, that's funny! I was just touching the same code because there was an easy Jira issue to fix:
> 
> * [[cling] Update policy for `GetFullyQualifiedName` in cling utils #19292](https://github.com/root-project/root/pull/19292)
> 
> Maybe you can take the unit test from that PR and put it here too? The reproducer should also be fixed by what you're doing here. Or we merge my PR before? As you prefer.

Yes, I saw your PR. I think it’s better that your PR is merged first. I need to look a bit more into the failing cases.

guitargeek:
Well, my PR has the same problems.

Could it be a clang upstream problem? I noticed that even with LLVM 20, printing out the qual type of the `std::vector<T>::at` method is missing the template arguments of the function argument types:
```txt
std::vector<int>::reference (std::vector<int>::*)(std::vector::size_type)
```
I used the following policy when printing:
```c++
PrintingPolicy Policy(Context->getPrintingPolicy());

Policy.FullyQualifiedName = true;
Policy.SuppressElaboration = true;
```

Maybe this should be fixed upstream?

ferdymercury:
Related: https://github.com/root-project/root/pull/999

PR #19294, opened 2025-07-05
fix: fix progress bar for RDataFrame with Range limits

### LINKED ISSUE #15323: RDF Progress bar seems to ignore Range() calls
### Check duplicate issues.

- [X] Checked for duplicates

### Description

When Range() is called on a RDF object with a progress bar, the progress bar seems to show all the events instead of reducing the events to the values provided in the Range() call.

### Reproducer

```c++
void test() {

  ROOT::RDataFrame df(100);
  ROOT::RDF::Experimental::AddProgressBar(df);
  auto def = df.Define("x", [](){ return gRandom->Rndm(); });
  auto max = def.Range(0, 10);
  auto hist = max.Histo1D("x");
  std::cout << hist->Integral() << "\n";
}
```

The following code shows:
```
root [0] 
Processing test.cxx...
[Total elapsed time: 0:00m  processed files: 0 / 0  processed evts: 100 / 100]                                                                                                                      
10
```

While the progress bar should ideally show 10/10

### ROOT version

Recent master branch

### Installation method

git clone + local compilation

### Operating system

Ubuntu 23.10

### Additional context

_No response_

MohamedElashri:
# This Pull request:

## Changes or fixes:

This PR fixes the progress bar display for RDataFrame when using `Range()` transformations. Previously, the progress bar would always show the total entries from the original dataset, even when using `Range()` to limit the number of processed entries. Now the progress bar correctly shows the actual number of entries being processed.

 Also this PR adds new tutorial `df108_ProgressRange.C` demonstrating the correct usage with `Range()` 

I need to write a unit test but still trying to figure this out so this is a work in progress PR. I 

## Checklist:

- [x] tested changes locally - Everything works!
- [x] updated the docs (if necessary) - If adding new tutorial counts but can work on updating to ROOT docs itself. 

This PR fixes #15323

### DIFF
--- a/tree/dataframe/inc/ROOT/RDFHelpers.hxx
+++ b/tree/dataframe/inc/ROOT/RDFHelpers.hxx
@@ -17,6 +17,7 @@
 #include <ROOT/RDF/RActionBase.hxx>
 #include <ROOT/RDF/RResultMap.hxx>
 #include <ROOT/RResultHandle.hxx> // users of RunGraphs might rely on this transitive include
+#include <ROOT/RResultPtr.hxx>
 #include <ROOT/TypeTraits.hxx>
 
 #include <array>
@@ -276,8 +277,9 @@ SnapshotPtr_t VariationsFor(SnapshotPtr_t resPtr);
 /// \brief Add ProgressBar to a ROOT::RDF::RNode
 /// \param[in] df RDataFrame node at which ProgressBar is called.
 ///
-/// The ProgressBar can be added not only at the RDataFrame head node, but also at any any computational node,
-/// such as Filter or Define.
+/// The ProgressBar can be added not only at the RDataFrame head node, but also at any computational node,
+/// such as Filter or Define. To correctly account for the entries processed, place the progress bar
+/// after transformations that reduce the number of events (e.g. `Range`).
 /// ###Example usage:
 /// ~~~{.cpp}
 /// ROOT::RDataFrame df("tree", "file.root");
@@ -360,6 +362,8 @@ private:
    bool fUseShellColours;
 
    std::shared_ptr<TTree> fTree{nullptr};
+   ROOT::RDF::RResultPtr<ULong64_t> fTotalEntries{};
+
 
 public:
    /// Create a progress helper.
@@ -370,7 +374,8 @@ public:
    /// \param useColors Use shell colour codes to colour the output. Automatically disabled when
    /// we are not writing to a tty.
    ProgressHelper(std::size_t increment, unsigned int totalFiles = 1, unsigned int progressBarWidth = 40,
-                  unsigned int printInterval = 1, bool useColors = true);
+                  unsigned int printInterval = 1, bool useColors = true,
+                  ROOT::RDF::RResultPtr<ULong64_t> totalEntries = {});
 
    ~ProgressHelper() = default;
 
--- a/tree/dataframe/src/RDFHelpers.cxx
+++ b/tree/dataframe/src/RDFHelpers.cxx
@@ -15,6 +15,7 @@
 #include "ROOT/RDF/RLoopManager.hxx" // for RLoopManager
 #include "ROOT/RDF/Utils.hxx"
 #include "ROOT/RResultHandle.hxx" // for RResultHandle, RunGraphs
+#include "ROOT/RResultPtr.hxx"
 
 #include "TROOT.h"      // IsImplicitMTEnabled
 #include "TError.h"     // Warning
@@ -157,7 +158,8 @@ void ThreadsPerTH3(unsigned int N)
 }
 
 ProgressHelper::ProgressHelper(std::size_t increment, unsigned int totalFiles, unsigned int progressBarWidth,
-                               unsigned int printInterval, bool useColors)
+                              unsigned int printInterval, bool useColors,
+                              ROOT::RDF::RResultPtr<ULong64_t> totalEntries)
    : fPrintInterval(printInterval),
      fIncrement{increment},
      fBarWidth{progressBarWidth = int(get_tty_size() / 4)},
@@ -169,6 +171,7 @@ ProgressHelper::ProgressHelper(std::size_t increment, unsigned int totalFiles, u
      fIsTTY{isatty(fileno(stdout)) == 1},
      fUseShellColours{useColors && fIsTTY} // Control characters only with terminals.
 #endif
+      , fTotalEntries(std::move(totalEntries))
 {
 }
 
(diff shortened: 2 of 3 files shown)


vepadulano:
Hi @MohamedElashri ,

Thank you for attempting to fix this issue, really appreciated! I tried your tutorial and I noticed that with the current changes the implementation of the progress bar is only displayed at the end of the event loop, which unfortunately defeats its purpose. You can see with the following example that I created by slightly modifying your tutorial
```cpp
#include <iostream>
#include <ROOT/RDFHelpers.hxx>
#include <ROOT/RLogger.hxx>

auto verbosity = ROOT::RLogScopedVerbosity(ROOT::Detail::RDF::RDFLogChannel(), ROOT::ELogLevel::kInfo);

void df108_ProgressRange()
{
   ROOT::RDataFrame df(1000000);
   auto ranged = df.Range(500);
   ROOT::RDF::Experimental::AddProgressBar(ranged);
   auto h = ranged.Define("x", []() { return double{42}; }).Histo1D<double>("x");
   std::cout << h->GetEntries() << std::endl;
}

int main()
{
df108_ProgressRange();
}
```

The output is
```
Info in <[ROOT.RDF] Info /home/vpadulan/Programs/rootproject/root/tree/dataframe/src/RLoopManager.cxx:900 in void ROOT::Detail::RDF::RLoopManager::Run(bool)>: Starting event loop number 0.
Info in <[ROOT.RDF] Info /home/vpadulan/Programs/rootproject/root/tree/dataframe/src/RLoopManager.cxx:859 in void ROOT::Detail::RDF::RLoopManager::Jit()>: Nothing to jit and execute.
Info in <[ROOT.RDF] Info /home/vpadulan/Programs/rootproject/root/tree/dataframe/src/RLoopManager.cxx:938 in void ROOT::Detail::RDF::RLoopManager::Run(bool)>: Finished event loop number 0 (0s CPU, 0.000498056s elapsed).
[Total elapsed time: 0:00m  processed files: 0 / 0  processed evts: 500 / 500] 
```

As you can see the progressbar only appears at the end of the event loop. I believe there is some hidden trigger of the computation graph in your changes which makes them unfortunately not fit for review at the moment.

MohamedElashri:
Hi @vepadulano 

Thanks for the feedback. This is my first contribution so I wanted just to make it small and limited to the particular problem of wrong progress-bar with `Range()`. But what you are describing is actually the current behavior of `RDataFrame` progress-bar so it is not something new. 

However, I agree that the current progress-bar is useless. I'll work on making a useful version but it would need some time to think about handling large and small number of events/iterations.

MohamedElashri:
@vepadulano 

Hi, I have a version that I think is more suitable. Can you try to test it. The problem with your example is that is has very short lifetime. The `return double{42}` completes in milliseconds so any implementation choice would make it hard to see any progress in the `progessbar`. I made some choices about update frequency that is adaptive. I updated the tutorial for couple of examples about usage (third one is about how not to use it).

Can you please have a look when you get sometime. We can also discuss the criteria for the update frequency. 


PS: To make your example show some intermediate progress we can add some complex operation

```cpp
#include <iostream>
#include <ROOT/RDFHelpers.hxx>
#include <ROOT/RLogger.hxx>
#include <TMath.h>

void df108_ProgressRange()
{
   // Enable verbose logging to see what's happening
   auto verbosity = ROOT::RLogScopedVerbosity(ROOT::Detail::RDF::RDFLogChannel(), ROOT::ELogLevel::kInfo);

   
   ROOT::RDataFrame df(1000000);
   auto ranged = df.Range(500000);
   auto node = ROOT::RDF::AsRNode(ranged);
   ROOT::RDF::Experimental::AddProgressBar(node);
   
   // Add some computation to make it run longer and show intermediate progress
   auto h = ranged.Define("x", []() { 
                     // Add computational work to slow it down
                     double result = 0;
                     for(int i = 0; i < 1000; ++i) {
                        result += TMath::Sin(i * 0.01) * TMath::Cos(i * 0.02);
                     }
                     return result; 
                  }).Histo1D<double>("x");
   
   std::cout << "Histogram entries: " << h->GetEntries() << std::endl;
}
```

kutsibalci:
Not asking to take this over — it is @MohamedElashri's PR and he asked for another look twelve months
ago. But I spent some time measuring this progress bar for #15323 and I think I can answer the
objection that stalled the review, so I would rather put it here than let it sit.

## Why the bar only appeared at the end of @vepadulano's reproducer

That reproducer (`return double{42}`, finishing in milliseconds) is blocked by **two independent
gates**, both of which exist on `master` today (`0eeb28461`) and neither of which this PR introduces:

**1. The event-count gate.** `tree/dataframe/src/RDFHelpers.cxx:465`

```cpp
constexpr std::size_t callbackEveryNEvents = 1000;
...
r.OnPartialResultSlot(callbackEveryNEvents, [progress](unsigned int slot, auto &&arg) { ... });   // :470
```

`ProgressHelper::Update()` is reached only once per **1000 events per slot**. An event loop of 100 —
or of 999 — never reaches it once.

**2. The time gate.** `RDFHelpers.cxx:191` and `:408`

```cpp
fPrintInterval{printInterval == 0 ? (fIsTTY ? 1 : 10) : printInterval}     // :191
...
if (duration_cast<seconds>(system_clock::now() - fLastPrintTime) < fPrintInterval) {  // :408
   return;
}
```

Even when the callback does fire, nothing is printed until a **full second** has elapsed since the
last print (ten seconds when stdout is not a TTY).

So a fast, small loop prints exactly one line — the final one — by construction. **That confirms
@MohamedElashri's reply of 14 July:** it is the existing behaviour of the bar rather than a
regression introduced here. To see an in-run line at all, a reproducer needs both ≥1000 events per
slot and ≥1 second of wall time.

## Corroborating measurement

Independently of this PR, I ran four configurations differing only in what reduces the graph —
`Range(0,10)` with the bar booked before it, booked after it, no `Range` at all, and a `Filter`
keeping 10 of 100. On 6.36.02 the final line was `processed evts: 100 / 100` in **all four**; on
`master` (6.41.01) it was `processed events: 100` in all four, while `Count()` returned 10, 10, 100,
10. Details in [#15323](https://github.com/root-project/root/issues/15323#issuecomment-5239511726).

The useful conclusion for this PR: `Range` is not the variable. The bar ignores *every* graph-level
reduction, because the denominator comes from `RSampleInfo::NEntriesTotal()` via
`ProgressHelper::RegisterNewSample`, which is a property of the input that the graph cannot touch,
while the numerator follows the graph through `OnPartialResultSlot`. Two structurally independent
paths — which is why the fix has to reconcile them rather than adjust one.

(The separate rate defect on that same final line — dividing by a duration already truncated to whole
seconds, so a sub-second loop printed `inf evt/s` — was fixed in #23065 and is merged, so it should
no longer be in the way here.)

@vepadulano @martamaja10 — the substantive question left is the intent one: should the bar report
what was *processed* or the *dataset size*? That is a maintainer's call, and it is the thing this PR
has been waiting on since September 2025. Whichever way it goes, I am happy to help with tests.

*AI disclosure: AI-assisted (Claude Code). The line numbers and constants above were read from
`master` at `0eeb28461` before posting, and the four-configuration measurement was run by me; I take
responsibility for the claims.*

PR #19295, opened 2025-07-07
[skip-ci][hist] Document architecture for initial set of classes

hahnjo:
To enable incremental review, the start of this architecture document is provided upfront. Further classes will be added when implemented.

### DIFF
--- a/hist/histv7/doc/CodeArchitecture.md
+++ b/hist/histv7/doc/CodeArchitecture.md
@@ -1,0 +1,77 @@
+# Code Architecture
+
+This document lists all classes, describes their responsibilities, and details how they fit together.
+
+## Core Classes
+
+These classes are involved when the user calls `Fill`.
+The list is ordered "bottom-up" in terms of functionality.
+
+### `RRegularAxis`, `RVariableBinAxis`, `RCategoricalAxis`
+
+These classes implement a concrete axis type.
+Their main task is to compute the linearized index for a single `Fill` argument:
+```c++
+RLinearizedIndex ComputeLinearizedIndex(double x);
+```
+The `bool` is used to indicate if the return value is valid.
+For example, the argument may be outside the axis with the underflow and overflow bins disabled.
+`RLinearizedIndex` is a simple struct with a `std::size_t index` and `bool valid`.
+It is chosen over `std::optional` because it unifies the return value construction:
+If outside the axis, the validity is just determined by the member property `fEnableFlowBins`.
+
+### `Internal::RAxes`
+
+This class is responsible for managing a histogram's axis configuration.
+It stores the axis objects as a `std::vector` of `std::variant`s.
+Objects of this class are used internally and not exposed to the user.
+Relevant functionality is available through user-facing classes such as `RHistEngine`.
+
+The main function is
+```c++
+template <typename... A>
+RLinearizedIndex ComputeGlobalIndex(const std::tuple<A...> &args) const;
+```
+that dispatches arguments to the individual `ComputeLinearizedIndex` and combines the results.
+If any of the linearized indices is invalid, so will be the combination.
+
+### `RHistEngine`
+
+This class combines an `RAxes` object and storage of bin contents in a `std::vector`.
+During `Fill`, it calls `RAxes::ComputeLinearizedIndex` and then updates the bin content.
+In contrast to `RHist`, this class supports direct concurrent filling via `FillAtomic`.
+
+### `RHistStats`
+
+Manages the (global) histogram statistics, such as the number of entries.
+It also keeps statistics of the unbinned values for each dimension.
+
+### `RHist`
+
+This class combines `RHistEngine`, with its axes and bin contents, and `RHistStats`.
+During `Fill`, it delegates to `RHistEngine::Fill` but also updates the histogram statistics.
+
+## Classes for Weighted Filling
+
+### `RDoubleBinWithError`
+
+A special bin content type that also accumulates the sum of weights squared.
+It can be used as a template argument to `RHistEngine` and `RHist`.
+
+### `RWeight`
+
+A wrapper `struct` for a single `double` value, used for weighted filling to distinguish its type.
+Objects of this type are passed by value.
+
+## Auxiliary Classes
+
+### `RBinIndex`
+
+A single bin index, which is just an integer for normal bins.
+`Underflow()` and `Overflow()` are special values and not ordered with respect to others.
+Objects of this type are passed by value; most notably to `GetBinContent` and `SetBinContent`.
+
+### `RBinIndexRange`
+
+A range of `RBinIndex` from `begin` (inclusive) to `end` (exclusive).
+The class exposes an iterator interface that can be used in range-based loops.


hist/histv7/doc/CodeArchitecture.md
```diff
@@ -0,0 +1,75 @@
...
+The `bool` is used to indicate if the return value is valid.
+For example, the argument may be outside the axis with the underflow and overflow bins disabled.
+`std::pair` is chosen over `std::optional` because it allows to return the same linearized index, even if invalid.
+
+### `Internal::RAxes`
+
+This class is responsible for managing a histogram's axis configuration.
+It stores the axis objects as a `std::vector` of `std::variant`s.
+Objects of this class are used internally and not exposed to the user.
+Relevant functionality is available through user-facing classes such as `RHistEngine`.
```
hahnjo, line 28: Here I'm still unsure if we want this class in `Internal`, or in `Detail` and conditionally exposed to users. I think we don't have to decide immediately, but should keep in mind that (one day) the class name will be part of on-disk data...

hist/histv7/doc/CodeArchitecture.md
```diff
@@ -0,0 +1,76 @@
...
+
+These classes are involved when the user calls `Fill`.
+The list is ordered "bottom-up" in terms of functionality.
+
+### `RRegularAxis`, `RVariableBinAxis`, `RCategoricalAxis`
+
+These classes implement a concrete axis type.
+Their main task is to compute the linearized index for a single `Fill` argument:
+```c++
+std::pair<std::size_t, bool> ComputeLinearizedIndex(double x);
```
jblomer: How about a struct as a return type so that the two properties can have names?

jblomer (approved):

hist/histv7/doc/CodeArchitecture.md
hahnjo: You mean something simple as
```c++
struct RLinearizedIndex final {
   std::size_t index;
   bool valid;
};
```
Let me see how this integrates into the first pieces of the "final" `RRegularAxis` that I put together yesterday, but I think that's a good idea and makes some code easier to understand with less documentation. It might also allow proper overload resolution for methods such as `GetBinContent` that could in principle accept both (many) `RBinIndex` or a single global `RLinearizedIndex`, we'll have to see...

hageboeck: In case the struct gets chosen, should one declare an `bool operator bool() const { return valid; }` to allow for:
```c++
if (const auto index = Compute...(); index) {
  ...
}
```
?

This would, however, very much go in the direction of `std::optional`, so I'm not sure if we're not just reimplementing a part of std::optional in this case.

hist/histv7/doc/CodeArchitecture.md
```diff
@@ -0,0 +1,76 @@
...
+```
+The `bool` is used to indicate if the return value is valid.
+For example, the argument may be outside the axis with the underflow and overflow bins disabled.
+`std::pair` is chosen over `std::optional` because it unifies the return value construction:
+If outside the axis, the validity is just determined by the member property `fEnableFlowBins`.
```
hageboeck, line 21: I don't yet see the difference to the pair or the struct discussed above. They should all be value+bool. If you have under-/overflow enabled, you will just unconditionally access the index. Otherwise, you test for the validity. In that sense, optional is safer, because it prevents you from accidentally accessing an invalid index value.
Would the above be enough to reconsider the choice?

hist/histv7/doc/CodeArchitecture.md
```diff
@@ -0,0 +1,76 @@
...
+It also keeps statistics of the unbinned values for each dimension.
+
+### `RHist`
+
+This class combines `RHistEngine`, with its axes and bin contents, and `RHistStats`.
+During `Fill`, it delegates to `RHistEngine::Fill` but also updates the histogram statistics.
+
+## Classes for Weighted Filling
+
+### `RDoubleBinWithError`
```
hageboeck, line 56: There is a slight ambiguity in this name:
- It could be a bin using `double` as data type, with an error
- Or it could be two bins, with an error

Would you have an idea how to resolve this? Would `RBinWithError` be enough?

hist/histv7/doc/CodeArchitecture.md
hahnjo: Here's what I mean with "it unifies the return value construction:"
```c++
    if (underflow) {
      return {fNumBins, fEnableFlowBins};
    } else if (overflow) {
      return {fNumBins + 1, fEnableFlowBins};
    }
```
With `std::optional`, I have to write something like
```c++
    if (underflow) {
      if (fEnableFlowBins) {
        return fNumBins;
      } else {
        return std::nullopt;
      }
    } else if (overflow) {
      if (fEnableFlowBins) {
        return fNumBins + 1;
      } else {
        return std::nullopt;
      }
    }
```
Sure, I can make this more terse with ternaries and maybe the compiler will even optimize it to the same code as above, but a simple job should be implemented with simple code.

I could also handle the two cases together
```c++
    if ((underflow || overflow) && !fEnableFlowBins) {
      return std::nullopt;
    } else if (underflow) {
      return fNumBins;
    } else if (overflow) {
      return fNumBins + 1;
    }
```
but that's again quite some logic to reason about. Just setting validity to `fEnableFlowBins` seems pretty clear to me.

> If you have under-/overflow enabled, you will just unconditionally access the index. Otherwise, you test for the validity.

No, the caller side (`RAxes`) doesn't (and shouldn't) know if the particular axis has underflow and overflow bins enabled. It always has to check for validity.

> In that sense, optional is safer, because it prevents you from accidentally accessing an invalid index value.

That's not true, `std::optional<T>::operator *` is undefined behavior if there is no value. In that sense, I could argue that accessing a wrong index is better because it's "only" a logic error (in practice probably both are equally bad).

I totally agree that `std::optional` is great for difficult-to-construct types, but I don't see this for `std::size_t`. And the semantics aren't 100% right either from my point of view...

hist/histv7/doc/CodeArchitecture.md
hahnjo: > Let me see how this integrates

Pretty nicely actually; even the test cases get easier to read with `EXPECT_TRUE(linIndex.valid);` :smiley: 

> should one declare an `bool operator bool() const { return valid; }`

Maybe, but I don't see the user interacting with `RLinearizedIndex` a lot. I think it will mostly be our own code in `RAxes` and there I would prefer explicit `.valid` over conversion to `bool`.

hist/histv7/doc/CodeArchitecture.md
hahnjo: To be honest, I'm also not fully happy with the name, but couldn't find a better and short one. I think I would prefer to make it clear from the name that the precision is `double`.

We can consider `RBinWithError<T>` and only allow floating-point types for `T`. Though that also gets us `RBinWithError<float>` which I'm not sure we want due to limited precision...

hist/histv7/doc/CodeArchitecture.md
hahnjo: I'm resolving this conversation since I implemented Jakob's suggestion and the proposed `operator bool()` can be added later on, if necessary. In any case, it's not part of this "Code Architecture" document.

hist/histv7/doc/CodeArchitecture.md
hahnjo: Since the naming is not yet of immediate concern (if I count correctly, weighted filling is only for PR number 10 at this point), I propose to keep this in mind and carry on for now.

hist/histv7/doc/CodeArchitecture.md
hahnjo: @hageboeck does this make sense? We can of course always revisit these decisions later on, but for the moment I quite like `struct RLinearizedIndex` as implemented in https://github.com/root-project/root/pull/19334

hageboeck: > Here's what I mean with "it unifies the return value construction:"

Thanks, that makes it indeed clearer!

hahnjo: Ok, resolving this discussion.

hist/histv7/doc/CodeArchitecture.md
```diff
@@ -0,0 +1,77 @@
...
+The list is ordered "bottom-up" in terms of functionality.
+
+### `RRegularAxis`, `RVariableBinAxis`, `RCategoricalAxis`
+
+These classes implement a concrete axis type.
+Their main task is to compute the linearized index for a single `Fill` argument:
+```c++
+RLinearizedIndex ComputeLinearizedIndex(double x);
+```
+The `bool` is used to indicate if the return value is valid.
```
pcanal, line 17: Since the switch from `pair<size_t,bool>` to `RLinearizedIndex` and with `RLinearizedIndex` not being described (at that point of reading).  This sentence ("The bool ..") is lacking a bit of context.

hist/histv7/doc/CodeArchitecture.md
```diff
@@ -0,0 +1,77 @@
...
+For example, the argument may be outside the axis with the underflow and overflow bins disabled.
+`RLinearizedIndex` is a simple struct with a `std::size_t index` and `bool valid`.
+It is chosen over `std::optional` because it unifies the return value construction:
+If outside the axis, the validity is just determined by the member property `fEnableFlowBins`.
```
pcanal, line 21:
```
If outside the axis, the validity is still valid if flow bins have been enabled (as recorded by the member property `fEnableFlowBins`).
```

hist/histv7/doc/CodeArchitecture.md
```diff
@@ -0,0 +1,77 @@
...
+Objects of this class are used internally and not exposed to the user.
+Relevant functionality is available through user-facing classes such as `RHistEngine`.
+
+The main function is
+```c++
+template <typename... A>
+RLinearizedIndex ComputeGlobalIndex(const std::tuple<A...> &args) const;
+```
+that dispatches arguments to the individual `ComputeLinearizedIndex` and combines the results.
+If any of the linearized indices is invalid, so will be the combination.
```
pcanal, line 36:
```
If any of the linearized indices is invalid, so is the combination.
```
maybe ... to match tenses.

hist/histv7/doc/CodeArchitecture.md
```diff
@@ -0,0 +1,77 @@
...
+
+### `RHistEngine`
+
+This class combines an `RAxes` object and storage of bin contents in a `std::vector`.
+During `Fill`, it calls `RAxes::ComputeLinearizedIndex` and then updates the bin content.
+In contrast to `RHist`, this class supports direct concurrent filling via `FillAtomic`.
```
pcanal, line 42: This could use some more context and justification. (i.e. why this contrast/difference?).

hist/histv7/doc/CodeArchitecture.md
```diff
@@ -0,0 +1,77 @@
...
+### `RWeight`
+
+A wrapper `struct` for a single `double` value, used for weighted filling to distinguish its type.
+Objects of this type are passed by value.
+
+## Auxiliary Classes
+
+### `RBinIndex`
+
+A single bin index, which is just an integer for normal bins.
```
pcanal, line 70: What is the relation ship with `RLinearizedIndex`?

hist/histv7/doc/CodeArchitecture.md
```diff
@@ -0,0 +1,77 @@
...
+`Underflow()` and `Overflow()` are special values and not ordered with respect to others.
```
pcanal, line 71: > not ordered with respect to others.

Why not? and/or why is it important?

hist/histv7/doc/CodeArchitecture.md
hahnjo: For future reference: I called it `RBinWithError` in https://github.com/root-project/root/pull/19927

PR #19299, opened 2025-07-07
[python] Ensure ctest sees results of pytest

vepadulano:
Without the Python script returning the pytest.main result as an exit code, it might happen that ctest thinks that the pytest suite was successfull. Instead, `raise SystemExit` ensures whatever result of pytest is also seen by ctest


siliataider (approved):
Subtle, thank you :)

vepadulano:
Superseded by https://github.com/root-project/root/pull/19301

PR #19300, opened 2025-07-07
[ntuple] RNTupleMerger: fix incorrect assumption about parent field

silverweed:
In GenerateZeroPagesForColumns() it was assumed that the zeroed columns' parent field always comes from the source descriptor. This is not the case, so we now save the correct RNTupleDescriptor that the field desc belongs to in order to safely navigate the field hierarchy.

Fixes a crash observed by @amete in https://gitlab.cern.ch/amete/rootparallelmerger/-/tree/master-rntuple-prototype

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/test/ntuple_merger.cxx
+++ b/tree/ntuple/test/ntuple_merger.cxx
@@ -3419,3 +3419,90 @@ TEST(RNTupleMerger, MergeUntypedSymmetric)
       }
    }
 }
+
+TEST(RNTupleMerger, GenerateZeroPagesIncremental)
+{
+   // Incrementally Union-merge RNTuples with alternating fields, so to trigger generation of zero pages.
+   // This is a minimal reproducer of a bug uncovered by an ATLAS-like merging flow:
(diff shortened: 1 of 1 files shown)


tree/ntuple/src/RNTupleMerger.cxx
```diff
@@ -317,12 +317,15 @@ struct RColumnMergeInfo {
...
    std::string fColumnName;
+   // The column id in the source RNTuple
    ROOT::DescriptorId_t fInputId;
+   // The corresponding column id in the destination RNTuple (the mapping happens in AddColumnsFromField())
    ROOT::DescriptorId_t fOutputId;
    ENTupleColumnType fColumnType;
    // If nullopt, use the default in-memory type
    std::optional<std::type_index> fInMemoryType;
    const ROOT::RFieldDescriptor *fParentField;
+   const ROOT::RNTupleDescriptor *fParentFieldDescriptor;
```
jblomer:
```suggestion
   const ROOT::RFieldDescriptor *fParentFieldDescriptor;
   const ROOT::RNTupleDescriptor *fParentNTupleDescriptor;
```

PR #19305, opened 2025-07-08
[main] replace rootls.py with new C++ version

silverweed:
# This Pull request:
removes `rootls.py` in favor of the new C++ version. Also moves the main utils tests from `roottest/python/cmdLineUtils` to `roottest/main`.


dpiparo (approved):
Thanks for this new implementation! Great!

PR #19306, opened 2025-07-08
[cmake] Disable builtin_zeromq if no internet

bellenot:
Disable `builtin_zeromq`, `builtin_cppzmq`, and `roofit_multiprocess` if there is no internet connection


cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -1872,8 +1879,14 @@ if (roofit_multiprocess)
   endif()
 
   if(builtin_cppzmq)
-    list(APPEND ROOT_BUILTINS cppzmq)
-    add_subdirectory(builtins/zeromq/cppzmq)
+    ROOT_CHECK_CONNECTION("builtin_cppzmq=OFF")
+    if(NO_CONNECTION)
+      message(STATUS "No internet connection, disabling the `builtin_cppzmq` option")
+      set(builtin_cppzmq OFF CACHE BOOL "Disabled because there is no internet connection" FORCE)
```
guitargeek:
```suggestion
      message(STATUS "No internet connection, disabling the `builtin_cppzmq` and `roofit_multiprocess` options")
      set(builtin_cppzmq OFF CACHE BOOL "Disabled because there is no internet connection" FORCE)
      set(roofit_multiprocess OFF CACHE BOOL "Disabled because there is no internet connection" FORCE)
```

I guess in that case we need to switch off RooFit multiprocessing too? It needs both zeromq and cppmq

bellenot: Right, see [lines 1853-1855](https://github.com/root-project/root/pull/19306/files#diff-d2db5be43e703287f436f304b0f6eb75f76c66e20e1c64b3f26223ba436e8e54R1853-R1855)

guitargeek: Sure, but what if the user configured with `builtin_zeromq=OFF` and `builtin_cppzmq=ON`? That's not unreasonable: some environments ship zeromq but not the C++ bindings.

Right now with your PR, if there would be this configuration, `roofit_multiprocess` would not be disabled without internet connection and the build will fail.

bellenot: OK, fine

guitargeek (approved):
Thank you very much!

PR #19307, opened 2025-07-08
[skip-ci] Make sure obsolete doc pages are not kept

couet:
On S3: Add option `--delete` to the `sync` command to remove files not present in the source

on EOS: before copying the new doc remove the old version of the doc to make sure the obsolete pages are removed. This is done using some tempory directory to avoid doc unavailibility.

### DIFF
--- a/.github/workflows/root-docs-ci.yml
+++ b/.github/workflows/root-docs-ci.yml
@@ -185,8 +185,8 @@ jobs:
       run: |
         pwd
         ls -l
-        aws s3 sync ${DOC_DIR}/html/ s3://root/doc/${WEB_DIR_NAME}/
-        aws s3 sync ${DOC_DIR}/ s3://root/doc/${WEB_DIR_NAME}/ --exclude "${DOC_DIR}/html/*"
+        aws s3 sync --delete ${DOC_DIR}/html/ s3://root/doc/${WEB_DIR_NAME}/
+        aws s3 sync --delete ${DOC_DIR}/ s3://root/doc/${WEB_DIR_NAME}/ --exclude "${DOC_DIR}/html/*"
         aws s3 cp ${TAR_NAME}.gz s3://root/download/
 
     - name: Install Kerberos utilities
--- a/.github/workflows/root-docs-ci.yml
+++ b/.github/workflows/root-docs-ci.yml
@@ -212,6 +212,8 @@ jobs:
       run: |
         echo ${RWEBEOS_KT} | base64 -d > ${KT_FILE_NAME}
         kinit -p ${{ secrets.KRB5USER }}@${{ secrets.KRB5REALM }} -kt ${KT_FILE_NAME}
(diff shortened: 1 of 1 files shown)


.github/workflows/root-docs-ci.yml
```diff
@@ -212,6 +212,8 @@ jobs:
       run: |
         echo ${RWEBEOS_KT} | base64 -d > ${KT_FILE_NAME}
         kinit -p ${{ secrets.KRB5USER }}@${{ secrets.KRB5REALM }} -kt ${KT_FILE_NAME}
+        EOS_HOST=${EOS_ENDPOINT#root://}
+        xrdfs "${EOS_HOST}" rm -r "${EOS_BASE_PATH}/doc/${WEB_DIR_NAME}" || true
```
hahnjo: A standard "trick" for these kind of cases is first copying into a second directory, then swap the two folders (which should be pretty fast), and finally delete the old directory (which might again take some time).

dpiparo:
Thanks for this PR. 
I think two things need to be thoroughly checked:
- What is the "--delete" command doing to the rest of the content of the bucket, e.g. releases and files? Why do we care about S3, which is used as backup and is not served in production?
- The current way of treating the docs on EOS will lead to long periods, tens of minutes, during which no documentation is available: deleting-copying anew does not seem the right thing here.

hahnjo:
> The current way of treating the docs on EOS will lead to long periods, tens of minutes, during which no documentation is available: deleting-copying anew does not seem the right thing here.

That's why I proposed copying into a temporary directory and then swapping, as implemented since this morning by @couet.

couet:
_What is the "--delete" command doing to the rest of the content of the bucket, e.g. releases and files?_

The sync command does not delete anything from the destination by default. The `--delete` flag enables this behavior. It will remove any destination files that no longer exist in the source location.

couet:
_Why do we care about S3, which is used as backup and is not served in production?_

I guess it is better to have a clean backup. It does not make sense to backup files which do no exist anymore.

dpiparo:
for xrootd, the way to move directories is explained here https://linux.die.net/man/1/xrdfs : `xrdfs mv`
What we may then need is:
1. copy the doc into a temporary directory
2. move the existing target directory, if existing, to a temporary name
3. move the temporary directory of step 1 into the name of the target directory of 2
4. remove the temporary name of 2
Step `n` can be executed if and only if step `n-1` is successful.
Can this work?

hahnjo:
But I think that's already exactly what is implemented in this PR, no?

dpiparo:
indeed. The PR is ready to be tried, as far as I am concerned @couet

dpiparo (approved):

hahnjo:
@couet the upload failed: https://github.com/root-project/root/actions/runs/17086035449/job/48450211789
```
Server responded with an error: [3010] Removing relative path '-r' is disallowed.`
```

PR #19309, opened 2025-07-08
[tree] fix copy size since Long_t has non-portable width

ferdymercury:


dpiparo:
Thanks for these changes. Could you perhaps share how you arrived to this PR, e.g. the problem you encountered and wanted to fix?

ferdymercury:
Hi, I found this problem in TLeafG about some copy-paste errors in the documentation: https://github.com/root-project/root/pull/19308

So I took a look also at the code to see if there were other occurrences of 'forgot to change' from Long64_t to Long_t, so I found this 'candidate one'. But it's not related to any real problem I've seen, but rather a 'preventive change'.

PR #10470, opened 2022-04-27
[RF] Error out when setting out-of-range variable value instead of clipping

guitargeek:
In previous ROOT versions, if you set the value of a variable with
`RooRealVar::setVal()`, the value was silently clippend when it was
outside the variable range. This silent mutation of data can be
dangerous. With ROOT 6.36, an exception will be thrown instead. If you
know what you are doing and want to restore the old clipping behavior,
you can do so with `RooRealVar::enableSilentClipping()`, but this is not
recommended.

This change caught a logic error in `stressRooFit` where a background
fraction was randomized to be outside the `[0,1]` interval. That part of
the test was therefore removed (it was already removed from the
corresponding tutorial).

Closes https://github.com/root-project/root/issues/6937.

PR #19316, opened 2025-07-08
[RF] Avoid using out-of-range values in RooDerivative

### LINKED ISSUE #18615: RooAbsReal::derivative needs to be fixed

After commit 7641d03, (where ignore policy has been replaced by exception) 
a fix/protection in `RooAbsReal::derivative` is required. 
The case of calculation of numerical derivative close to the edge of the variable needs proper solution.  

### ROOT version

ROOT master 

### Installation method

LCG nightly dev3 cvsfm build 

### Operating system

lxplus 

### Additional context

_No response_

guitargeek:
Avoid using out-of-range values for the numerical differentiation in RooDerivative.

This adds three new special cases:

  * derivative is first order and x-value is close to limit: just use one-directional derivative already implemented in the RichardsonDerivator

  * higher-order derivative but function is constant within floating point precision: just return zero

  * higher-order derivative for non-constant function: throw error, because we don't want to risk wrong derivatives by evaluating outside the range, which will result in x-value-clipping and therefore invalid y value differences

The tutorials fortunately hit the second case with the locally constant function, so we can remove the workaround with the sub-range introduced in #10470.

Closes #18615.

FYI, @VanyaBelyaev

### DIFF
--- a/roofit/roofitcore/src/RooDerivative.cxx
+++ b/roofit/roofitcore/src/RooDerivative.cxx
@@ -90,21 +90,59 @@ RooDerivative::~RooDerivative() = default;
 
 double RooDerivative::evaluate() const
 {
-  if (!_ftor) {
-    _ftor = std::unique_ptr<RooFunctor>{_func.arg().functor(_x.arg(),RooArgSet(),_nset)};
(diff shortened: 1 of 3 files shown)


dpiparo (approved):
LGTM. Thanks.

PR #19319, opened 2025-07-09
[ntuple] Fields tree as dot

magnustymoteus:
# This Pull request:

## Changes or fixes:
This PR introduces the method `RNTupleInspector::GetFieldsTreeAsDot` which gives a string that represents the tree structure of RNTuple fields in .dot language (that can then be displayed in graphviz for example)

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
+++ b/tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
@@ -40,8 +40,16 @@ class RPageSource;
 
 namespace Experimental {
 
-enum class ENTupleInspectorPrintFormat { kTable, kCSV };
-enum class ENTupleInspectorHist { kCount, kNElems, kCompressedSize, kUncompressedSize };
+enum class ENTupleInspectorPrintFormat {
+   kTable,
+   kCSV
+};
+enum class ENTupleInspectorHist {
+   kCount,
+   kNElems,
+   kCompressedSize,
+   kUncompressedSize
+};
 
 // clang-format off
 /**
@@ -93,7 +101,7 @@ public:
          : fColumnDescriptor(colDesc),
            fCompressedPageSizes(compressedPageSizes),
            fElementSize(elemSize),
-           fNElements(nElems) {};
+           fNElements(nElems){};
       ~RColumnInspector() = default;
 
       const ROOT::RColumnDescriptor &GetDescriptor() const { return fColumnDescriptor; }
@@ -101,7 +109,8 @@ public:
       std::uint64_t GetNPages() const { return fCompressedPageSizes.size(); }
       std::uint64_t GetCompressedSize() const
       {
-         return std::accumulate(fCompressedPageSizes.begin(), fCompressedPageSizes.end(), static_cast<std::uint64_t>(0));
+         return std::accumulate(fCompressedPageSizes.begin(), fCompressedPageSizes.end(),
+                                static_cast<std::uint64_t>(0));
       }
       std::uint64_t GetUncompressedSize() const { return fElementSize * fNElements; }
       std::uint64_t GetElementSize() const { return fElementSize; }
@@ -123,7 +132,7 @@ public:
 
    public:
       RFieldTreeInspector(const ROOT::RFieldDescriptor &fieldDesc, std::uint64_t onDiskSize, std::uint64_t inMemSize)
-         : fRootFieldDescriptor(fieldDesc), fCompressedSize(onDiskSize), fUncompressedSize(inMemSize) {};
+         : fRootFieldDescriptor(fieldDesc), fCompressedSize(onDiskSize), fUncompressedSize(inMemSize){};
       ~RFieldTreeInspector() = default;
 
       const ROOT::RFieldDescriptor &GetDescriptor() const { return fRootFieldDescriptor; }
@@ -468,6 +477,23 @@ public:
    GetFieldsByName(std::string_view fieldNamePattern, bool searchInSubfields = true)
    {
       return GetFieldsByName(std::regex{std::string(fieldNamePattern)}, searchInSubfields);
+   }
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Print a .dot string that represents the tree of the (sub)fields of an RNTuple
+   ///
+   /// \param[in] fieldDescriptor The descriptor of the root field (this method works recursively)
+   ///
+
+   void PrintFieldTreeAsDot(const ROOT::RFieldDescriptor &fieldDescriptor, std::ostream &output = std::cout) const;
+
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Print the tree of all the (sub)fields of an RNTuple
+   /// \param[in] output
+   ///
+   /// \see PrintFieldTreeAsDot(const ROOT::RFieldDescriptor &fieldDescriptor, std::ostream &output=std::cout) const
+   void PrintFieldTreeAsDot(std::ostream &output = std::cout) const
+   {
+      PrintFieldTreeAsDot(GetDescriptor().GetFieldZero(), output);
    }
 };
 } // namespace Experimental
--- a/tree/ntupleutil/v7/src/RNTupleInspector.cxx
+++ b/tree/ntupleutil/v7/src/RNTupleInspector.cxx
@@ -516,3 +516,38 @@ ROOT::Experimental::RNTupleInspector::GetFieldsByName(const std::regex &fieldNam
 
    return fieldIds;
 }
+
+void ROOT::Experimental::RNTupleInspector::PrintFieldTreeAsDot(const ROOT::RFieldDescriptor &fieldDescriptor,
+                                                               std::ostream &output) const
+{
+   const auto &tupleDescriptor = GetDescriptor();
+   const bool isZeroField = fieldDescriptor.GetParentId() == ROOT::kInvalidDescriptorId;
+   if (isZeroField) {
+      output << "digraph D {\n";
+      output << "node[shape=box]\n";
+   }
+   const std::string &nodeId = (isZeroField) ? "0" : std::to_string(fieldDescriptor.GetId() + 1);
+   const std::string &fieldName = (isZeroField) ? "RFieldZero" : fieldDescriptor.GetFieldName();
+   const std::string &description = fieldDescriptor.GetFieldDescription();
+   const std::uint32_t &version = fieldDescriptor.GetFieldVersion();
+
+   output << nodeId << "[label=<";
+   if (!isZeroField) {
+      output << "<b>Name: </b>" << fieldName << "<br></br>";
+      output << "<b>Type: </b>" << fieldDescriptor.GetTypeName() << "<br></br>";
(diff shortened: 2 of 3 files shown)


tree/ntupleutil/v7/src/RNTupleInspector.cxx
```diff
@@ -516,3 +516,17 @@ ROOT::Experimental::RNTupleInspector::GetFieldsByName(const std::regex &fieldNam
 
    return fieldIds;
 }
+
+std::string ROOT::Experimental::RNTupleInspector::GetFieldsTreeAsDot(const ROOT::RFieldDescriptor &fieldDescriptor) const {
```
silverweed: I'm wondering whether it would be better to pass a `std::ostream &` as a second param (perhaps defaulted to `std::cout`) rather than returning a string.
Upside: it allows to write the output directly to a file (or any kind of stream) without passing through a string; downside: if one wants the output as a string it's a bit more verbose (you need to pass a stringstream and then convert it to string).
Of course we could simply have both overloads and have one call the other internally, but I suspect that wanting the output as a string is less common than wanting to dump it to file.

tree/ntupleutil/v7/src/RNTupleInspector.cxx
```diff
@@ -516,3 +516,17 @@ ROOT::Experimental::RNTupleInspector::GetFieldsByName(const std::regex &fieldNam
...
+   std::string result;
+   const auto &tupleDescriptor = GetDescriptor();
+   const bool isZero = fieldDescriptor.GetParentId() == ROOT::kInvalidDescriptorId;
+   const std::string &fieldName = (isZero) ? "\"Field Zero\"" : "\""+fieldDescriptor.GetFieldName()+" ("+fieldDescriptor.GetTypeName()+")\"";
+   for(const auto &childFieldId : fieldDescriptor.GetLinkIds()) {
+      const auto &childFieldDescriptor = tupleDescriptor.GetFieldDescriptor(childFieldId);
+      result += fieldName + std::string("->") + "\""+childFieldDescriptor.GetFieldName()+" ("+childFieldDescriptor.GetTypeName()+")\"" + ";\n";
+      result += GetFieldsTreeAsDot(childFieldDescriptor);
+   }
+   if(isZero) result = "digraph D {\n"+result+"}";
```
silverweed: Please run `clang-format` on the code so we keep a consistent style

tree/ntupleutil/v7/src/RNTupleInspector.cxx
magnustymoteus: I agree, in this case I will change the method accordingly, and its name to `RNTupleInspector::PrintFieldsTreeAsDot` (which is I assume more fitting for this requested change)

magnustymoteus:
This is an example of graphviz's rendered output (where Placeholder is a user-defined class containing an integer and a float)
graphviz output

This is obviously very rudimentary right now, but I can work on expanding this today.

tree/ntupleutil/v7/src/RNTupleInspector.cxx
```diff
@@ -516,3 +516,30 @@ ROOT::Experimental::RNTupleInspector::GetFieldsByName(const std::regex &fieldNam
...
+      output << "digraph D {\n";
+   const std::string &fieldId = (isZero) ? "0" : std::to_string(fieldDescriptor.GetId() + 1);
+   const std::string &fieldName = (isZero) ? "RFieldZero" : fieldDescriptor.GetFieldName();
+
+   output << fieldId << "[label=";
+   output << "<<table border=\"0\" cellborder=\"1\" cellspacing=\"0\" cellpadding=\"4\">";
+   output << "<tr> <td> <b>" << fieldName << "</b> </td> </tr>";
+   output << "<tr> <td><table border=\"0\" cellborder=\"0\" cellspacing=\"0\" >";
+   output << "<tr> <td align=\"left\" >" << fieldDescriptor.GetTypeName() << "</td> </tr></table></td> </tr>";
+   output << "<tr> <td align=\"left\">more here<br/>...<br align=\"left\"/></td> </tr></table>>" << ",";
```
silverweed: Is this "more here" a placeholder?

magnustymoteus: It is. Maybe I shouldn't have included that.

On another note, I am thinking about using ROOT's graphviz interface to output the graph instead because the current method is already starting to become a little ugly. The only downside to me is that it'd become a new dependency (with the upside of cleaner code).

silverweed: For now I would avoid involving the Graphviz interface; let's start with the text-only output.
I would simply add the field type and it can already be good as an initial version.
We don't want to spend too much time on this dot output as this is just a "warm up" exercise basically.
I also think that the code is at an acceptable level of cleanliness, given what it must do and that it's all contained in one small function :)

magnustymoteus:
This is the new rendered output with the same example:

graphviz2

Additionally, I removed the (tabular) styling to keep it simple.

tree/ntupleutil/v7/src/RNTupleInspector.cxx
```diff
@@ -516,3 +516,34 @@ ROOT::Experimental::RNTupleInspector::GetFieldsByName(const std::regex &fieldNam
...
+   const std::string &description = fieldDescriptor.GetFieldDescription();
+
+   output << nodeId << "[label=<";
+   if (!isZeroField) {
+      output << "<b>Name: </b>" << fieldName << "<br></br>";
+      output << "<b>Type: </b>" << fieldDescriptor.GetTypeName() << "<br></br>";
+      output << "<b>ID: </b>" << std::to_string(fieldDescriptor.GetId()) << "<br></br>";
+      if (description != "")
+         output << "<b>Description: </b>" << fieldDescriptor.GetFieldDescription() << "<br></br>";
+      output << "<b>Version: </b>" << fieldDescriptor.GetFieldVersion() << "<br></br>";
```
silverweed: Since it's quite uncommon to see versions != 0 I would add that information only when the version is not 0

tree/ntupleutil/v7/src/RNTupleInspector.cxx
```diff
@@ -516,3 +516,34 @@ ROOT::Experimental::RNTupleInspector::GetFieldsByName(const std::regex &fieldNam
...
+   } else
+      output << "<b>" << fieldName << "</b>";
+   output << ">, shape=box]\n";
```
silverweed: I'm pretty sure this `shape=box` can be made common rather than inline by adding one `node [shape=box]` at the beginning of the graph

silverweed (commented):
Looks good! Just a couple more things before we merge

tree/ntupleutil/v7/src/RNTupleInspector.cxx
magnustymoteus: Should be resolved now

tree/ntupleutil/v7/src/RNTupleInspector.cxx
magnustymoteus: Should also be resolved now

vepadulano (commented):
Very nice work! Can I suggest to clean the commit history before merging? That would include squashing formatting/small fixes commits together in one (or a few) main commits and using our convention for the commit message e.g.

```
[ntuple] VERB_IMPERATIVE_FORM description_of_commit_objective
```

tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
```diff
@@ -469,6 +478,23 @@ public:
    {
       return GetFieldsByName(std::regex{std::string(fieldNamePattern)}, searchInSubfields);
    }
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Print a .dot string that represents the tree of the (sub)fields of an RNTuple
+   ///
+   /// \param[in] fieldDescriptor The descriptor of the root field (this method works recursively)
+   ///
+
+   void PrintFieldsTreeAsDot(const ROOT::RFieldDescriptor &fieldDescriptor, std::ostream &output = std::cout) const;
```
enirolf:
```suggestion
   void PrintFieldTreeAsDot(const ROOT::RFieldDescriptor &fieldDescriptor, std::ostream &output = std::cout) const;
```

tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
```diff
@@ -469,6 +478,23 @@ public:
...
+
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Print the tree of all the (sub)fields of an RNTuple
+   /// \param[in] output
+   ///
+   /// \see PrintFieldsTreeAsDot(const ROOT::RFieldDescriptor &fieldDescriptor, std::ostream &output=std::cout) const
+   void PrintFieldsTreeAsDot(std::ostream &output = std::cout) const
```
enirolf:
```suggestion
   void PrintFieldTreeAsDot(std::ostream &output = std::cout) const
```

tree/ntupleutil/v7/src/RNTupleInspector.cxx
```diff
@@ -516,3 +516,38 @@ ROOT::Experimental::RNTupleInspector::GetFieldsByName(const std::regex &fieldNam
...
+      output << "<b>" << fieldName << "</b>";
+   output << ">]\n";
+   for (const auto &childFieldId : fieldDescriptor.GetLinkIds()) {
+      const auto &childFieldDescriptor = tupleDescriptor.GetFieldDescriptor(childFieldId);
+      output << nodeId + "->" + std::to_string(childFieldDescriptor.GetId() + 1) + "\n";
+      PrintFieldsTreeAsDot(childFieldDescriptor, output);
+   }
+   if (isZeroField)
+      output << "}";
+}
```
enirolf: This file misses a final linebreak at the end

tree/ntupleutil/v7/test/ntuple_inspector.cxx
```diff
@@ -813,3 +813,23 @@ TEST(RNTupleInspector, MultiColumnRepresentations)
    EXPECT_EQ(ENTupleColumnType::kReal16, px1Inspector.GetType());
    EXPECT_EQ(1u, px1Inspector.GetNElements());
 }
+
+TEST(RNTupleInspector, FieldsTreeAsDot)
```
enirolf: I believe the output from `PrintFieldsTreeAsDot` for this test is not all too large, right? In that case, perhaps it would be nice to (also) have a 1-to-1 string comparison instead of only checking the substrings below.

enirolf (changes requested):
Very nice work! I have some small nitpicks to address before merging, but after that it's good to go from my side :).

tree/ntupleutil/v7/src/RNTupleInspector.cxx
enirolf: Ah, sorry, I meant for the file itself mostly (i.e. after the closing bracket on l.553)!

magnustymoteus:
The requested changes should be implemented, I'll squash the commits now.

enirolf (approved):
Thank you for the changes, LGTM!

PR #19321, opened 2025-07-09
[WIP][Python] Unify maps used for numpy to cpp conversion

### LINKED ISSUE #18390: Better establish type conversion between numpy and C++ types

There are a few places where we need to convert numpy dtypes to C++ types, for example when adopting buffers of numpy arrays into RVecs. There are at least the following two conversion dictionaries:

1. https://github.com/root-project/root/blob/65a0e12f401a5cb2d6a4af729aa78f251a620409/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rvec.py#L64
2. https://github.com/root-project/root/blob/65a0e12f401a5cb2d6a4af729aa78f251a620409/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rvec.py#L78

Which do not contain currently the same set of semantic key-value pairs. Also, sometimes a builtin C++ type is used, sometimes a ROOT typedef is used, notably in the treatment of integer types. It would be beneficial to establish a set of all the types allowed for conversion between numpy and C++ types and resolve the inconsistencies. My preference would also be to avoid using ROOT typedefs and switch to C++ standard integer types. I expect this may lead to issues that will need to be investigated, see https://github.com/root-project/root/pull/18087

### ROOT version

Any

### Installation method

Any

### Operating system

Any

### Additional context

_No response_

siliataider:
# This Pull request:

> [!WARNING]
> This PR is still under development. I am opening it as a draft to see if this small refactoring breaks a lot of tests, before moving forward with broader changes.

Currently, multiple dictionaries for mapping numpy dtypes to C++ types exist, such as:
- [`_array_interface_dtype_map`](https://github.com/root-project/root/blob/65a0e12f401a5cb2d6a4af729aa78f251a620409/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rvec.py#L64).
- [`cpptypes`](https://github.com/root-project/root/blob/65a0e12f401a5cb2d6a4af729aa78f251a620409/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rvec.py#L78).
- And tests like [`array_conversions`](https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/test/array_conversions.py#L13), [`array_interface`](https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/test/array_interface.py#L13) and [`rtensor`](https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/test/rtensor.py#L21) define their own conversion helpers

The goal is to address some of the inconsistencies by refactoring the existing maps into one clear set of types allowed for conversion between NumPy and c++.

This PR fixes #18390

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_rtensor.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_rtensor.py
@@ -85,13 +85,15 @@ def get_array_interface(self):
     idx1 = cppname.find("RTensor<")
     idx2 = cppname.find(",", idx1)
     dtype = cppname[idx1 + 8 : idx2]
-    # dtype_numpy = _array_interface_dtype_map[dtype]
     for numpy_dtype, type_info in _numpy_dtype_typeinfo_map.items():
         if dtype == type_info["cpp"] or dtype == type_info["typedef"]:
             dtype_numpy = numpy_dtype
             dtype_size = cppyy.sizeof(dtype)
             break
-    dtype_size = cppyy.sizeof(dtype)
+
+    if dtype_numpy:
+        raise RuntimeError(f"Unsupported dtype '{dtype}' found in RTensor.")
+
     endianness = "<" if sys.byteorder == "little" else ">"
     shape = self.GetShape()
     strides = self.GetStrides()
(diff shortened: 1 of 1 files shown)


dpiparo:
Is the intention to deliver this PR for 6.40?

siliataider:
Not planned for 6.40 (unless I get around to it), I will close it for now

PR #19322, opened 2025-07-09
Improve logic to refresh branch addresses of TChain friends 

### LINKED ISSUE #16805: A TChain whose trees have friend TChains stops updating

vepadulano:
Fixes #16805 

Part of the logic of TChain::LoadTree deals with updating the status of friend
trees. First, the current entry is loaded on all friends. In case the friend
tree is a chain that needs to switch to a new tree, the branch addresses of the
friend are also updated.

This logic only considered the case of top-level friends of a TChain. Notably,
it did not consider the case of a main TChain where one (or more) of the trees
in the chain have themselves some friends, but not the main chain itself:

TChain c1
--> TTree t1 <friendswith> TChain c2

This commit condenses the logic that updates friends in a separate method of
TChain. The new method is called in TChain::LoadTree, either in the pre-existing
case of top-level friends or in the new case of friends of the current tree.

The new method, called RefreshFriendAddresses, needs to act:
- On the main chain and on the current tree, in the case of top-level friends
- On the current tree alone, in the case of friends of the current tree

In the two scenarios, either we load the global entry of the chain on the
friends, or the local entry of the current tree. This is done to conserve proper
alignment.

### DIFF
--- a/tree/tree/src/TChain.cxx
+++ b/tree/tree/src/TChain.cxx
@@ -1232,6 +1232,88 @@ Int_t TChain::LoadBaskets(Long64_t /*maxmemory*/)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+/// Refresh branch/leaf addresses of friend trees
+///
+/// The method acts only on the current tree in the chain (fTree), but it may
+/// be called in two different scenarios: when there are friends of the chain
+/// or when there are friends of fTree itself.
+Long64_t TChain::RefreshFriendAddresses()
+{
+   assert(fTree != nullptr);
+
+   bool needUpdate = false;
+   if (auto *innerFriendList = fTree->GetListOfFriends()) {
+      // If the current tree has friends, check if they were mark for update
+      // when switching to the following tree, detect it so that we later we
+      // actually refresh the addresses of the friends.
+      for (auto *frEl : ROOT::Detail::TRangeStaticCast<TFriendElement>(*innerFriendList)) {
+         if (frEl->IsUpdated()) {
+            needUpdate = true;
+            frEl->ResetUpdated();
+         }
+         if (frEl->IsUpdatedForChain()) {
+            needUpdate = true;
+            frEl->ResetUpdatedForChain();
+         }
+      }
+   }
+
+   if (!needUpdate)
+      return 0;
+
+   // Update the branch/leaf addresses and the list of leaves in all
+   // TTreeFormula of the TTreePlayer (if any).
+   for (auto *chainEl : ROOT::Detail::TRangeStaticCast<TChainElement>(*fStatus)) {
+      // Set the branch status of all the chain elements, which may include also
+      // branches that are available in friends. Only set the branch status
+      // if it has a value provided by the user
+      Int_t status = chainEl->GetStatus();
+      if (status != -1)
+         fTree->SetBranchStatus(chainEl->GetName(), status);
+
+      // Set the branch addresses for the newly opened file.
+      void *addr = chainEl->GetBaddress();
+      if (!addr)
+         continue;
+
+      TBranch *br = fTree->GetBranch(chainEl->GetName());
+      TBranch **pp = chainEl->GetBranchPtr();
+      if (pp) {
+         // FIXME: What if br is zero here?
+         *pp = br;
+      }
+      if (!br)
+         continue;
+
+      if (!chainEl->GetCheckedType()) {
+         Int_t res = CheckBranchAddressType(br, TClass::GetClass(chainEl->GetBaddressClassName()),
+                                            (EDataType)chainEl->GetBaddressType(), chainEl->GetBaddressIsPtr());
+         if ((res & kNeedEnableDecomposedObj) && !br->GetMakeClass()) {
+            br->SetMakeClass(true);
+         }
+         chainEl->SetDecomposedObj(br->GetMakeClass());
+         chainEl->SetCheckedType(true);
+      }
+      // FIXME: We may have to tell the branch it should
+      //        not be an owner of the object pointed at.
+      br->SetAddress(addr);
+      if (TestBit(kAutoDelete)) {
+         br->SetAutoDelete(true);
+      }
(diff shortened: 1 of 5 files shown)


tree/tree/src/TChain.cxx
```diff
@@ -1414,83 +1499,21 @@ Long64_t TChain::LoadTree(Long64_t entry)
...
-               fPlayer->UpdateFormulaLeaves();
-            }
-            // Notify user if requested.
-            if (fNotify) {
-               if(!fNotify->Notify()) return -6;
-            }
-         }
+      Long64_t refreshFriendAddressesRet{};
+      if (fFriends) {
+         refreshFriendAddressesRet = RefreshFriendAddresses(*this, *fTree, entry);
```
pcanal: Note that both here and 2 lines below the `fTree` can be reached by calling `GetTree()` on the first argument.

vepadulano: Indeed, but we're already inside an `if (fTree)` branch so I think it's ok to keep it as is

pcanal: My point is that the 2nd argument is technically redundant and moving the call to `GetTree()` inside the function may or may not make the relationship clearer.  (The signature of `RefreshFriendAddresses` allows the 2 input to be unrelated (if called incorrectly) but they are required to match `firstarg == firstarg->GetTree()`)

vepadulano: I see your point, although the correct relationship is `secondarg == firstarg->GetTree()`. I will correct it.

tree/tree/src/TChain.cxx
```diff
@@ -1292,97 +1393,36 @@ Long64_t TChain::LoadTree(Long64_t entry)
...
-            }
-            // Notify user if requested.
-            if (fNotify) {
-               if(!fNotify->Notify()) return -6;
-            }
-         }
+         // If this chain has friends, we need to pass the global entry to keep
+         // the friends aligned to the main chain
+         refreshFriendAddressesRet = RefreshFriendAddresses(*this, entry);
+      } else if (fTree->GetListOfFriends()) {
```
lmoureaux: Why `else`? Both the tree and the chain could have friends. (I don't think we use that though.)

vepadulano: Thanks for the catch! I'll try to create a reproducer for that case, if I manage then I will also change the code accordingly.

tree/tree/src/TChain.cxx
```diff
@@ -1231,6 +1231,87 @@ Int_t TChain::LoadBaskets(Long64_t /*maxmemory*/)
...
+   bool needUpdate = false;
+   if (auto *innerFriendList = fTree->GetListOfFriends()) {
+      // If the current tree has friends, check if they were mark for update
+      // when switching to the following tree, detect it so that we later we
+      // actually refresh the addresses of the friends.
+      for (auto *frEl : ROOT::Detail::TRangeStaticCast<TFriendElement>(*innerFriendList)) {
+         if (frEl->IsUpdated()) {
+            needUpdate = true;
+            frEl->ResetUpdated();
+         } else if (frEl->IsUpdatedForChain()) {
```
pcanal: It is probaly not an 'else' is but both.
```suggestion
         }
         if (frEl->IsUpdatedForChain()) {
```

PR #19323, opened 2025-07-09
[ntuple] improve type name renormalization

### LINKED ISSUE #19116: Re-evaluate the implementation of RField<T>::TypeName

During the investigation for https://github.com/root-project/root/pull/19087, we discovered that the implementation of `RField<T>::TypeName` when `T` is a custom user class (see [here](https://github.com/root-project/root/blob/957ac3356492765c14098525840bcd56f120da67/tree/ntuple/inc/ROOT/RField.hxx#L287)) calls internal functions which lead up to calls to the interpreter. In particular, see [TClassEdit::GetNormalizedName](https://github.com/root-project/root/blob/957ac3356492765c14098525840bcd56f120da67/core/foundation/src/TClassEdit.cxx#L891). Note that this function calls into the interpreter and ROOT meta.

As found out while working on https://github.com/root-project/root/pull/19087, `TClassEdit::GetNormalizedName` does not also run the autoload machinery. In that scenario, this led to the function returning the wrong class name, without having applied the `ROOT::Meta::Selection` rule to strip the second template argument because the class dictionary was not loaded.  

So, even after the changes in https://github.com/root-project/root/pull/19087 are applied, a program that only calls `RField<T>::TypeName`, without ever loading the dictionary information in any other way (e.g. by calling any combination of `TClass::GetClass`, `RNTupleReader::GetModel` or similar), will return the wrong class name.

We should make a decision about the semantics of the `RField<T>::TypeName` method, including the following considerations:
* If the method should return the correct type name as expected by ROOT meta, then further work should be done by the method itself, or somewhere in `TClassEdit` perhaps?
* Should the method just not call the interpreter at all? Then it should be reimplemented it.
* Should the runtime type checks in RNTuple be extended so that initially they try with the fast path (that doesn't call the interpreter) and only then make an attempt at finding the class name via the interpreter?

### ROOT version

master

### Installation method

Any

### Operating system

Any

### Additional context

_No response_

jblomer:
Add different code path for renormalizing a demangled typeid name. Such
type names come from RField<T>::TypeName(). In this case, the compiler
has done already most of the normalization work and we don't need to go
through ROOT meta. There are some parts ROOT meta does that we need to
repeat for this case, e.g. dropping optional template arguments of
stdlib containers.

This PR fixes #19116


tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -805,9 +805,10 @@ e.g. `std::vector<MyEvent>` or `std::vector<std::vector<float>>`.
 ### Type Name Normalization
 
 Type names are stored according to the following normalization rules
-  - The type name of a field has typedefs and usings fully resolved (except for the following rule).
+  - The type name of a field has typedefs and usings fully resolved (except for the following two rules).
   - The integer types `signed char`, `unsigned char`, and `[signed|unsigned](short|int|long[ long])`
     are replaced by the corresponding (at the time of writing) `std::[u]int(8|16|32|64)_t` standard integer typedef.
+  - Supported stdlib types are not further resolved (e.g., `std::string` is _not_ stored as `std::basic_string<char>`).
```
silverweed, line 811: Should we add +1 to the version patch?

silverweed (commented):
Small comment + a warning from the compiler

tree/ntuple/src/RFieldUtils.cxx
```diff
@@ -147,9 +136,185 @@ std::string ROOT::Internal::GetCanonicalTypePrefix(const std::string &typeName)
...
+   while (itr != typeName.end()) {
+      auto c = *itr;
+      itr++;
+
+      if (c != ',')
+         continue;
+
+      R__ASSERT(itr != typeName.end());
+      if (*itr == ' ') {
+         itr = typeName.erase(itr);
```
hahnjo: This potentially needs to copy all characters after `itr`. I think the widely used pattern is the one in `RemoveSpaceBeforeClosingAngleBracket` where you iterate until the end and then call `erase` once.

tree/ntuple/test/ntuple_type_name.cxx
```diff
@@ -58,6 +58,52 @@ TEST(RNTuple, TypeNameNormalization)
...
+   EXPECT_EQ("std::map<std::int32_t,std::int32_t>",
+             ROOT::Internal::GetRenormalizedTypeName(typeid(std::map<int, int>)));
+   EXPECT_EQ("std::unordered_map<std::int32_t,std::int32_t>",
+             ROOT::Internal::GetRenormalizedTypeName(typeid(std::unordered_map<int, int>)));
+   EXPECT_EQ("std::multimap<std::int32_t,std::int32_t>",
+             ROOT::Internal::GetRenormalizedTypeName(typeid(std::multimap<int, int>)));
+   EXPECT_EQ("std::unordered_multimap<std::int32_t,std::int32_t>",
+             ROOT::Internal::GetRenormalizedTypeName(typeid(std::unordered_multimap<int, int>)));
+
+   EXPECT_EQ("CustomStruct", ROOT::Internal::GetRenormalizedTypeName(typeid(class CustomStruct)));
```
hahnjo:
```suggestion
   EXPECT_EQ("CustomStruct", ROOT::Internal::GetRenormalizedTypeName(typeid(CustomStruct)));
```

tree/ntuple/test/ntuple_type_name.cxx
```diff
@@ -58,6 +58,52 @@ TEST(RNTuple, TypeNameNormalization)
...
 }
 
+TEST(RNTuple, TypeNameNormalizationById)
+{
+   EXPECT_EQ("std::int32_t", ROOT::Internal::GetRenormalizedTypeName(typeid(signed)));
+   EXPECT_EQ("std::int32_t", ROOT::Internal::GetRenormalizedTypeName(typeid(std::int32_t)));
+   EXPECT_EQ("double", ROOT::Internal::GetRenormalizedTypeName(typeid(Double32_t)));
+   EXPECT_EQ("float", ROOT::Internal::GetRenormalizedTypeName(typeid(const float)));
+
+   EXPECT_EQ("std::string", ROOT::Internal::GetRenormalizedTypeName(typeid(std::string)));
```
hahnjo: Should we test that the normalization works correctly inside of template arguments, for example `std::map<std::string, std::string>`?

ferdymercury:
Maybe related: https://its.cern.ch/jira/browse/ROOT-5862

roottest/root/ntuple/atlas-datavector/read.cxx
```diff
@@ -21,8 +21,8 @@ TEST(RNTupleAtlasDataVector, Read)
    ASSERT_NO_THROW(entry.GetPtr<AtlasLikeDataVector<CustomStruct>>("my_field"));
 
    const auto typeNameAfter = ROOT::RField<AtlasLikeDataVector<CustomStruct>>::TypeName();
-   const std::string expectedAfter{"AtlasLikeDataVector<CustomStruct>"};
-   // Make sure autoloading happened and the rule to suppress the second template argument kicked in
```
pcanal, line 25: Isn't that removing the test of the actual ATLAS issue?

pcanal: I.e. don't we need both the original testing and a new test?

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -805,10 +805,12 @@ e.g. `std::vector<MyEvent>` or `std::vector<std::vector<float>>`.
...
 
 Type names are stored according to the following normalization rules
-  - The type name of a field has typedefs and usings fully resolved (except for the following rule).
+  - The type name of a field has typedefs and usings fully resolved (except for the following two rules).
   - The integer types `signed char`, `unsigned char`, and `[signed|unsigned](short|int|long[ long])`
     are replaced by the corresponding (at the time of writing) `std::[u]int(8|16|32|64)_t` standard integer typedef.
-  - Qualifiers `volatile` and `const` that do not appear in template arguments are removed.
+  - Supported stdlib types are not further resolved (e.g., `std::string` is _not_ stored as `std::basic_string<char>`).
+  - C style array types (`T[N][M]`) are mapped to stdlib arrays (`std::array<std::array<T,M>,N>`)
+  - Qualifiers `volatile` and `const` that do not appear in template arguments of user-defined types are removed.
```
pcanal, line 813: Does this means that `std::vector<const std::pair<int, double>>` (no user-defined types involved?) is recorded as `std::vector<std::pair<int,double>>` but `MyStruct<const std::pair<int,double>>` is recorded as is?
(and then what about `MyStruct<std::vector<const std::pair<int,double>>>`? )

tree/ntuple/src/RFieldUtils.cxx
```diff
@@ -118,37 +138,288 @@ std::vector<AnglePos> FindTemplateAngleBrackets(const std::string &typeName)
...
+   } else if (typeName == "long long" || typeName == "long long int" || typeName == "signed long long" ||
+              typeName == "signed long long int") {
+      typeName = ROOT::RField<long long int>::TypeName();
+   } else if (typeName == "unsigned long long" || typeName == "unsigned long long int") {
+      typeName = ROOT::RField<unsigned long long int>::TypeName();
+   }
+}
+
+// Note that ROOT Meta already defines GetDemangledTypeName()
+std::string GetRawDemangledTypeName(const std::type_info &ti)
```
pcanal: Do `GetRawDemangledTypeName` and `GetDemangledTypeName` have different functionality (and if yes should they be summarized here)?

tree/ntuple/src/RFieldUtils.cxx
```diff
@@ -118,37 +138,288 @@ std::vector<AnglePos> FindTemplateAngleBrackets(const std::string &typeName)
...
+   char *str = TClassEdit::DemangleName(ti.name(), e);
+   R__ASSERT(str && e == 0);
+   std::string result{str};
+   free(str);
+
+   return result;
+}
+
+// Reverse std::string --> std::basic_string<char>
+void MapStdString(std::string &normalizedTypeName)
```
pcanal: `Map` is not talking to me ... What about `NormalizeStdString`?

tree/ntuple/src/RFieldUtils.cxx
```diff
@@ -118,37 +138,288 @@ std::vector<AnglePos> FindTemplateAngleBrackets(const std::string &typeName)
...
+   }
+
+   // Append the rest of the type from the last closing angle bracket.
+   const auto lastClosePos = angleBrackets.back().second;
+   normName += templatedTypeName.substr(lastClosePos + 1);
+
+   templatedTypeName = normName;
+}
+
+std::string GetRenormalizedMetaTypeName(const std::string &metaNormalizedName)
```
pcanal, line 381: Should we add a short description of the intent of the function?

tree/ntuple/src/RFieldUtils.cxx
```diff
@@ -118,37 +138,288 @@ std::vector<AnglePos> FindTemplateAngleBrackets(const std::string &typeName)
...
+      return ROOT::Internal::GetStandardArrayType(canonicalTypePrefix, dimensions);
+   }
+
+   std::string normName{canonicalTypePrefix};
+   NormalizeTemplateArguments(normName, 0 /* maxTemplateArgs */, GetRenormalizedMetaTypeName);
+
+   return ROOT::Internal::GetStandardArrayType(normName, dimensions);
+}
+
+std::string GetRenormalizedDemangledTypeName(const std::string &demangledName, bool mapStdString)
```
pcanal: Should we add a short description of the intent of the function?

tree/ntuple/doc/BinaryFormatSpecification.md
jblomer: Yes. The last one becomes `MyStruct<std::vector<std::pair<int,double>>>`.

roottest/root/ntuple/atlas-datavector/read.cxx
jblomer: The problem observed by ATLAS is tested above. This particular part tests the `TypeName()` function only, which in the DataVector case is different from `RField<T>::GetTypeName()`. What we test now (after the patch) is that `TypeName()` always gives the same result independent of the currently loaded libraries. 

It's arguably weak for that purpose and we may drop the `TypeName()`  part here altogether.

pcanal: I am missing something as I dont see anymore the test for the short name being there ... I am especially thrown off by the comment that is being removed:
```
Make sure autoloading happened and the rule to suppress the second template argument kicked in
```
which seems explicitly say that the above call and follow test are exactly for the ATLAS scenario ... and/or where should that comment move to?

tree/ntuple/src/RFieldUtils.cxx
jblomer: Yes, `ROOT::Internal::GetDemangledTypeName()` demangles and normalizes the name. I'll extend the comment.

roottest/root/ntuple/atlas-datavector/read.cxx
jblomer: Line 21 is the actual test:
``` 
ASSERT_NO_THROW(entry.GetPtr<AtlasLikeDataVector<CustomStruct>>("my_field"));
```
We ask for a shared pointer without the extra template argument, and it works even tough on disk the extra argument is present.

pcanal: I am quite confused.  If/since new line 24 set the expectation to recording with the long name, then line 21 will work with or without the dictionary (demangled name match the long name).

So unless we also make sure that in the test the 'current' long name does not match the long name when storing the file (I can't recall if we are doing that), then we wouldn't really test the feature ('please ignore the 2nd argument') ... (that is, unless I am still missing something)

tree/ntuple/src/RFieldUtils.cxx
jblomer: Yep, I went for `RenormalizeStdString` because the input is already normalized in some way.

roottest/root/ntuple/atlas-datavector/read.cxx
jblomer: Sorry, you're right! The test didn't make much sense. I added a modified version which I hope makes more sense.

pcanal: At which point in the new test is the autoloading of the library triggered?

jblomer: Actually I'm not sure. I guess by the constructor of `ROOT::RField<AtlasLikeDataVector<CustomStruct>>`? But does it matter?

pcanal: It does in the sense that the previous code/test was checking both before and after the autoloading, ensuring that the autoloading did happen.  Without these tests, the test can later be unwittingly updated to no longer test the autoloading at all (for example by explicitly linking the dictionary library) and thus removing the test of 'RNTuple use of a specific time customized by the users will ensure that the user's library is auto-loaded' (and thus increasing the possibility that we unwittingly stop supporting it)

jblomer: Fair enough. Test is updated.

tree/ntuple/src/RFieldUtils.cxx
```diff
@@ -118,37 +138,280 @@ std::vector<AnglePos> FindTemplateAngleBrackets(const std::string &typeName)
...
+{
+   static const std::string gStringName =
+      GetRenormalizedDemangledTypeName(GetRawDemangledTypeName(typeid(std::string)), false /* renormalizeStdString */);
+
+   // For real nested types of std::string (not typedefs like std::string::size_type), we would need to also check
+   // something like (normalizedTypeName + "::" == gStringName + "::") and replace only the prefix. However, since
+   // such a nested type is not standardized, it currently does not seem necessary to add the logic.
+   if (normalizedTypeName == gStringName) {
+      normalizedTypeName = "std::string";
+      return;
```
hahnjo: `return` not needed, I think

tree/ntuple/src/RFieldUtils.cxx
```diff
@@ -118,37 +138,280 @@ std::vector<AnglePos> FindTemplateAngleBrackets(const std::string &typeName)
...
+          {typeid(std::optional<char>),                  "std::optional<"},
+          {typeid(std::set<char>),                       "std::set<"},
+          {typeid(std::unordered_set<char>),             "std::unordered_set<"},
+          {typeid(std::multiset<char>),                  "std::multiset<"},
+          {typeid(std::unordered_multiset<char>),        "std::unordered_multiset<"},
+          {typeid(std::map<char, char>),                 "std::map<"},
+          {typeid(std::unordered_map<char, char>),       "std::unordered_map<"},
+          {typeid(std::multimap<char, char>),            "std::multimap<"},
+          {typeid(std::unordered_multimap<char, char>),  "std::unordered_multimap<"},
+          {typeid(std::atomic<char>),                    "std::atomic<"}};
```
hahnjo, line 326: Maybe order alphabetically?

silverweed: Or, although it's fuzzy, by how commonly a type is used? (at least for "obvious" types like vector/unique_ptr/maps)

tree/ntuple/src/RFieldUtils.cxx
```diff
@@ -118,37 +138,280 @@ std::vector<AnglePos> FindTemplateAngleBrackets(const std::string &typeName)
...
+   RemoveLeadingKeyword(canonicalTypePrefix);
+   MapIntegerType(canonicalTypePrefix);
+
+   if (canonicalTypePrefix.find('<') == std::string::npos) {
+      // If there are no templates, the function is done.
+      return canonicalTypePrefix;
+   }
+   RemoveSpaceBefore(canonicalTypePrefix, '>');
+   RemoveSpaceAfter(canonicalTypePrefix, ',');
+   RemoveSpaceBefore(canonicalTypePrefix, ','); // MSVC fancies spaces before commas in the demangled name
```
silverweed, line 417: Should we try collapsing these three operations in one to avoid scanning the string three times?

tree/ntuple/src/RFieldUtils.cxx
jblomer: I ordered them according to their appearance in the binary specification. I can add this as a comment.

tree/ntuple/src/RFieldUtils.cxx
jblomer: We could but it seems to me that this is a micro optimization that doesn't gain much, given the other string operations we perform and the fact that this method is not on a critical code path.

roottest/root/ntuple/atlas-datavector/read.cxx
jblomer: @pcanal can this be resolved?

vepadulano (approved):

hahnjo:
@pcanal any progress on reviewing this? If not, I propose we merge this because it is blocking my work on #19342...

roottest/root/ntuple/atlas-datavector/read.cxx
pcanal: Yes.  Thanks.

On a side note (not to be addressed in this PR), we may want to consider reviewing the function names related to typename (for example in this test `TypeName`, `GetTypeName` and `GetDemangledTypeName`) as they do not 'clearly' carry their distinction in meaning (eg. follows RNTuple name convention or ROOT I/O normalized name convention)

PR #19324, opened 2025-07-09
roottest: avoid generated unneeded pcm files

pcanal:
The tests are checking if the dictionary source files can compile.

Having the pcm files around can affect the other tests in the same directory, for example in case those
other test runs before the pcm has been rebuild (and it depend on something that has already been rebuild).

### DIFF
--- a/roottest/cling/stl/map/CMakeLists.txt
+++ b/roottest/cling/stl/map/CMakeLists.txt
@@ -1,4 +1,4 @@
-ROOTTEST_GENERATE_DICTIONARY(dict.C testing.C LINKDEF LinkDef.h)
+# ROOTTEST_GENERATE_DICTIONARY(dict.C testing.C NO_CXXMODULE LINKDEF LinkDef.h)
 
 ROOTTEST_COMPILE_MACRO(MyOpClass.C
                        FIXTURES_SETUP cling-stl-map-MyOpClass-fixtures)
--- a/roottest/cling/parsing/CMakeLists.txt
+++ b/roottest/cling/parsing/CMakeLists.txt
@@ -9,8 +9,8 @@ ROOTTEST_ADD_TEST(runoperatorLessThanTemplate
                   MACRO runoperatorLessThanTemplate.C
                   LABELS roottest regression cling)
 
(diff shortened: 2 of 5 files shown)


pcanal:
@dpiparo @vepadulano This is now ready for review.  It works well but requires a clean build (to get rid of the the stale-no-longer produced pcm files).

.github/workflows/root-ci.yml
```diff
@@ -583,3 +583,19 @@ jobs:
       - name: ccache info (post)
         run: |
           ccache -s || true
+
+  event_file:
```
vepadulano: Looks like spurious changes?

roottest/cling/stl/map/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-ROOTTEST_GENERATE_DICTIONARY(dict.C testing.C LINKDEF LinkDef.h)
+# ROOTTEST_GENERATE_DICTIONARY(dict.C testing.C NO_CXXMODULE LINKDEF LinkDef.h)
```
vepadulano: Remove line if it's not needed

.github/workflows/root-ci.yml
vepadulano: Ok I now saw the commit message mentioning
```
This seems to prevent the display of error logs in the web interface.
```

I believe this should be tackled in a separate PR

pcanal:
> Mostly good, but I believe we should address the CI workflow issue separately

Yes, it actually was.

vepadulano:
> Yes, it actually was.

Ok, does it mean that the third commit of this PR can be removed?

roottest/cling/stl/map/CMakeLists.txt
pcanal: Thanks.  Actually it was just temporarily removed to see the interaction and need to be re-added and 'completed' with a proper test.

.github/workflows/root-ci.yml
pcanal: Yes, it was merged in its own PR.

PR #19328, opened 2025-07-10
[CMake] Add `check_connection` flag to disable connection check

### LINKED ISSUE #19325: ROOT_CHECK_CONNECTION is too aggressive for sites with limited network access
### Check duplicate issues.

- [x] Checked for duplicates

### Description

The CMAKE [ROOT_CHECK_CONNECTION](https://github.com/root-project/root/blob/10d7bd3b942a6bd61034116b91e26f88b70f3418/cmake/modules/SearchInstalledSoftware.cmake#L24C7-L24C29) macro is too aggressive for my use case:
I want to compile ROOT on a cluster with limited network access. I can access Github and any other major code hosting site but everything else is blocked. Therefore, `ROOT_CHECK_CONNECTION` will fail even though access to Github and most Gitlabs is possible.

### Reproducer

Compile ROOT with any package that is builtin but needs network connection

### ROOT version

I am trying to compile ROOT 6.36.02

### Installation method

Build from source

### Operating system

RHEL 8

### Additional context

_No response_

guitargeek:
This can be useful in cases where the connection check will fail, but fetching content from code hosting sites like GitHub works (right now, we imperatively rely on GitHub connection by default, because Clad is downloaded from GitHub).

Closes #19325.


Superharz:
Would it make sense to add a line to the "no connection exists" message in the build setup printout, that this behavior can be overriden with this new flag?

guitargeek:
You are right.

cmake/modules/RootBuildOptions.cmake
```diff
@@ -116,6 +116,7 @@ ROOT_BUILD_OPTION(builtin_zlib OFF "Build bundled copy of zlib")
 ROOT_BUILD_OPTION(builtin_zstd OFF "Build included libzstd, or use system libzstd")
 ROOT_BUILD_OPTION(ccache OFF "Enable ccache usage for speeding up builds")
 ROOT_BUILD_OPTION(cefweb OFF "Enable support for CEF (Chromium Embedded Framework) web-based display")
+ROOT_BUILD_OPTION(check_connection ON "Fail the configuration step if there is not internet connection, but it's required for the build")
```
bellenot:
```suggestion
ROOT_BUILD_OPTION(check_connection ON "Fail the configuration step if there is no internet connection, but it's required for the build")
```

bellenot (approved):
I think that might somehow confuse people and make things break at compile time, but if that can help some corner cases, then fine with me...

Issue #19330, opened 2025-07-10
A partial merge in TFileMerger creates directories that users didn't ask for


hageboeck:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

See details here:
https://root-forum.cern.ch/t/merge-only-selected-subdirectory-of-some-root-files/63867

Basically, this code should not create the directory B:
```py
import ROOT
with ROOT.TFile("test_input.root", "recreate") as outfile:
    outfile.mkdir("A")
    outfile.mkdir("B")

file_merger = ROOT.TFileMerger(False)
file_merger.AddFile("test_input.root")
file_merger.AddObjectNames("A")
file_merger.OutputFile("test_merged.root")
merge_mode = ROOT.TFileMerger.kOnlyListed | ROOT.TFileMerger.kAll | ROOT.TFileMerger.kIncremental
file_merger.PartialMerge(merge_mode)
```

### Reproducer

 See above

### ROOT version

All < 6.38

### Installation method

Not relevant

### Operating system

Not relevant

PR #19331, opened 2025-07-10
Remove empty directories from TFileMerger's output

hageboeck:
- Partial merging creates a full directory hierarchy in an output file,
even if the user didn't ask for the directories to be merged.
- Fix the doxygen of `PartialMerge`

Update:
- ~~[ ] It looks like the tutorial https://root.cern/doc/master/mergeSelective_8C.html is broken, too. I'll add another test and see if it can be fixed.~~

#19330

### DIFF
--- a/io/io/src/TFileMerger.cxx
+++ b/io/io/src/TFileMerger.cxx
@@ -591,8 +591,15 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type, 
       // GetPath(), so we can still figure out where we are in the recursion
 
       // If this folder is a onlyListed object, merge everything inside.
-      if (onlyListed) type &= ~kOnlyListed;
-      status = MergeRecursive(newdir, sourcelist, type);
+      const auto mergeType = onlyListed ? type & ~kOnlyListed : type;
+      status = MergeRecursive(newdir, sourcelist, mergeType);
+
+      if ((type & kOnlyListed) && !(type & kIncremental) && !onlyListed && newdir->GetNkeys() == 0) {
+         // None of the children were merged, and the directory is not listed
+         delete newdir;
+         newdir = nullptr;
+         target->rmdir(obj->GetName());
+      }
       // Delete newdir directory after having written it (merged)
       if (!(type&kIncremental)) delete newdir;
       if (onlyListed) type |= kOnlyListed;
--- a/io/io/test/TFileMergerTests.cxx
+++ b/io/io/test/TFileMergerTests.cxx
@@ -221,3 +221,61 @@ TEST(TFileMerger, ChangeFile)
    gSystem->Unlink("file6640mergerinput_2.root");
    gSystem->Unlink("file6640mergeroutput.root");
 }
+
+TEST(TFileMerger, SelectiveMergeWithDirectories)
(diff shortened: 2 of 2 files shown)


io/io/src/TFileMerger.cxx
```diff
@@ -591,8 +591,15 @@ Bool_t TFileMerger::MergeOne(TDirectory *target, TList *sourcelist, Int_t type,
       // GetPath(), so we can still figure out where we are in the recursion
 
       // If this folder is a onlyListed object, merge everything inside.
-      if (onlyListed) type &= ~kOnlyListed;
-      status = MergeRecursive(newdir, sourcelist, type);
+      const auto mergeType = onlyListed ? type &= ~kOnlyListed : type;
```
silverweed: This assignment inside a ternary operator is weird, please make them two separate lines

silverweed (changes requested):

io/io/src/TFileMerger.cxx
hageboeck: This assignment is actually unintended! Thanks for spotting this!

hageboeck: I removed the assignment now (which could of caused a bug for subsequent directories), and extended the test to ensure that subsequent directories get the same treatment.

pcanal:
> maybe you want an approval also from @pcanal

The current state is draft, is it ready for review?

io/io/test/TFileMergerTests.cxx
```diff
@@ -221,3 +221,47 @@ TEST(TFileMerger, ChangeFile)
...
+      dir->WriteObject(&histo, "Histo_B");
+      dir = infile.mkdir("C")->mkdir("D");
+      dir->WriteObject(&histo, "Histo_D");
+      dir = infile.mkdir("E")->mkdir("F");
+      dir->WriteObject(&histo, "Histo_F");
+   }
+
+   {
+      TFileMerger fileMerger(false);
+      fileMerger.AddFile(input);
```
pcanal: For completeness sake, consider using 2 files, selecting 3 histos in the subdirectory(ies) with one of the histos in both file, one of the histo in only the first file, one of the  histo only in the second file (in addition to all the histo that are not selected and can be in both files).

hageboeck:
> The current state is draft, is it ready for review?

The things that I had pushed are ready for review, but I noticed that the partial merge tutorial is broken. I therefore went back to draft to prevent somebody merging by accident.

io/io/test/TFileMergerTests.cxx
hageboeck: This is a good extension, so I will add this soon-ish.

pcanal:
Note: this PR is likely somewhat related to https://github.com/root-project/root/pull/18679

hageboeck:
> Note: this PR is likely somewhat related to #18679

~~Thanks! I will hold back this one, and re-test once the other one is merged!~~

Update:
I submitted the second test to be included in #18679, and rescoped this PR to only deal with the empty directories.

io/io/test/TFileMergerTests.cxx
hageboeck: Done!

PR #19334, opened 2025-07-10
[hist] Implement initial `RRegularAxis` and `RVariableBinAxis`

hahnjo:
The first versions of these classes expose only `ComputeLinearizedIndex` with simple implementations. Further improvements and optimizations are possible in the future.

This PR also includes basic build system additions and setup for unit testing.

### DIFF
--- a/hist/histv7/inc/ROOT/RLinearizedIndex.hxx
+++ b/hist/histv7/inc/ROOT/RLinearizedIndex.hxx
@@ -1,0 +1,28 @@
+/// \warning This is part of the ROOT 7 prototype! It will change without notice. It might trigger earthquakes. Feedback
+/// is welcome!
+
+#ifndef ROOT_RLinearizedIndex
+#define ROOT_RLinearizedIndex
+
+#include <cstddef>
+
+namespace ROOT {
+namespace Experimental {
+
+/**
+A linearized index that can be invalid.
+
+For example, when an argument is outside the axis and underflow / overflow bins are disabled.
+
+\warning This is part of the ROOT 7 prototype! It will change without notice. It might trigger earthquakes. Feedback is
+welcome!
+*/
+struct RLinearizedIndex final {
+   std::size_t fIndex;
+   bool fValid;
+};
+
+} // namespace Experimental
+} // namespace ROOT
+
+#endif
--- a/hist/histv7/inc/ROOT/RRegularAxis.hxx
+++ b/hist/histv7/inc/ROOT/RRegularAxis.hxx
@@ -1,0 +1,93 @@
+/// \warning This is part of the ROOT 7 prototype! It will change without notice. It might trigger earthquakes. Feedback
+/// is welcome!
+
+#ifndef ROOT_RRegularAxis
+#define ROOT_RRegularAxis
+
+#include "RLinearizedIndex.hxx"
+
+#include <cstddef>
+
+namespace ROOT {
+namespace Experimental {
+
+/**
+A regular axis with equidistant bins in the interval \f$[fLow, fHigh)\f$.
+
+For example, the following creates a regular axis with 10 normal bins between 5 and 15:
+~~~ {.cxx}
+ROOT::Experimental::RRegularAxis axis(10, 5, 15);
+~~~
+
+It is possible to disable underflow and overflow bins by passing `enableFlowBins = false`. In that case, arguments
+outside the axis will be silently discarded.
+
+\warning This is part of the ROOT 7 prototype! It will change without notice. It might trigger earthquakes. Feedback is
+welcome!
+*/
+class RRegularAxis final {
+   /// The number of normal bins
+   std::size_t fNumNormalBins;
+   /// The lower end of the axis interval
+   double fLow;
+   /// The upper end of the axis interval
+   double fHigh;
+   /// The cached inverse of the bin width to speed up ComputeLinearizedIndex
+   double fInvBinWidth; //!
+   /// Whether underflow and overflow bins are enabled
+   bool fEnableFlowBins;
+
+public:
+   /// Construct a regular axis object.
+   ///
+   /// \param[in] numNormalBins the number of normal bins
+   /// \param[in] low the lower end of the axis interval (inclusive)
+   /// \param[in] high the upper end of the axis interval (exclusive)
+   /// \param[in] enableFlowBins whether to enable underflow and overflow bins
+   RRegularAxis(std::size_t numNormalBins, double low, double high, bool enableFlowBins = true)
+      : fNumNormalBins(numNormalBins), fLow(low), fHigh(high), fEnableFlowBins(enableFlowBins)
+   {
+      // FIXME: should validate numNormalBins > 0 and low < high
+      fInvBinWidth = numNormalBins / (high - low);
+   }
+
+   std::size_t GetNumNormalBins() const { return fNumNormalBins; }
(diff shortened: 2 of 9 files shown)


hist/histv7/inc/ROOT/RVariableBinAxis.hxx
```diff
@@ -0,0 +1,84 @@
...
+For example, the following creates an axis with 3 log-spaced bins:
+~~~ {.cxx}
+std::vector<double> binEdges = {1, 10, 100, 1000};
+ROOT::Experimental::RVariableBinAxis axis(binEdges);
+~~~
+
+It is possible to disable underflow and overflow bins by passing `enableFlowBins = false`. In that case, arguments
+outside the axis will be silently discarded.
+*/
+class RVariableBinAxis final {
```
ferdymercury, line 31: Thanks!
Did you consider RVariableAxis as alternative name? (Or RIrregularAxis)

Or alternatively RRegularBinAxis vs RVariableBinAxis

hist/histv7/inc/ROOT/RRegularAxis.hxx
```diff
@@ -0,0 +1,87 @@
...
+      bool underflow = x < fLow;
+      // Put NaNs into overflow bin.
+      bool overflow = !(x < fHigh);
+      if (underflow) {
+         return {fNumNormalBins, fEnableFlowBins};
+      } else if (overflow) {
+         return {fNumNormalBins + 1, fEnableFlowBins};
+      }
+
+      std::size_t bin = (x - fLow) * fInvBinWidth;
```
ferdymercury, line 85: Should this also include the floating point precision loss correction from https://github.com/root-project/root/pull/19033#issuecomment-2996775496

According to @hageboeck, that correction didn't affect performance.

hahnjo: Please see the commit message (that's why I write them...)

hist/histv7/inc/ROOT/RVariableBinAxis.hxx
hahnjo: No:
`RVariableAxis`: it's not the axis that is variable
`RIrregularAxis`: we will have an `RCategoricalAxis` (later)
`RRegularBinAxis`: here it's not the bins that are regular (they are equidistant), but the axis (in analogy to "regular grid")

hist/histv7/inc/ROOT/RLinearizedIndex.hxx
```diff
@@ -0,0 +1,22 @@
+#ifndef ROOT_RLinearizedIndex
```
jblomer, line 4: Here and other files: add ROOT 7 standard comment preamble

hist/histv7/inc/ROOT/RLinearizedIndex.hxx
```diff
@@ -0,0 +1,22 @@
...
+namespace Experimental {
+
+/**
+A linearized index that can be invalid.
+
+For example, when an argument is outside the axis and underflow / overflow bins are disabled.
+*/
+struct RLinearizedIndex final {
+   std::size_t index;
+   bool valid;
```
jblomer: Style: `fIndex`, `fValid`

hist/histv7/inc/ROOT/RRegularAxis.hxx
```diff
@@ -0,0 +1,87 @@
...
+   {
+      // FIXME: should validate numNormalBins > 0 and low < high
+      fInvBinWidth = numNormalBins / (high - low);
+   }
+
+   std::size_t GetNumNormalBins() const { return fNumNormalBins; }
+   std::size_t GetTotalNumBins() const { return fEnableFlowBins ? fNumNormalBins + 2 : fNumNormalBins; }
+   double GetLow() const { return fLow; }
+   double GetHigh() const { return fHigh; }
+   bool AreFlowBinsEnabled() const { return fEnableFlowBins; }
```
jblomer:
```suggestion
   bool HasFlowBins() const { return fEnableFlowBins; }
```

jblomer (commented):
Hello, histv7!

hist/histv7/inc/ROOT/RLinearizedIndex.hxx
hahnjo: Ok, I wasn't sure if we want to follow the style here or keep it closer to STL types (`std::pair` with `first` and `second), but I guess uniformity in ROOT wins.

hist/histv7/inc/ROOT/RLinearizedIndex.hxx
hahnjo: I added the `\warning` paragraphs both to the the file header and the classes, so the user cannot miss them. I'm not a fan of `\author` and `\date` because it's already part of the git file history. And the `Copyright` header is objectively wrong, so I would prefer not to add it to new files I create...

hist/histv7/inc/ROOT/RRegularAxis.hxx
```diff
@@ -0,0 +1,93 @@
...
+   bool fEnableFlowBins;
+
+public:
+   /// Construct a regular axis object.
+   ///
+   /// \param[in] numNormalBins the number of normal bins
+   /// \param[in] low the lower end of the axis interval (inclusive)
+   /// \param[in] high the upper end of the axis interval (exclusive)
+   /// \param[in] enableFlowBins whether to enable underflow and overflow bins
+   RRegularAxis(std::size_t numNormalBins, double low, double high, bool enableFlowBins = true)
```
hageboeck, line 47: One could possibly make this `constexpr`, in case some users want to predefine an axis and reuse it several times.

Don't know if this helps in any way, though.

hageboeck (approved):
Only one small comment

hist/histv7/inc/ROOT/RRegularAxis.hxx
hahnjo: Yes, I think we could for `RRegularAxis` but it's not clear to me how much this would help; users can already cheaply copy the class. It doesn't work for `RVariableBinAxis` until C++20 makes `std::vector` constructors `constexpr`. So I think for symmetry and lack of strong motivation, I would keep it without the annotations for the moment.

hist/histv7/inc/ROOT/RLinearizedIndex.hxx
pcanal: > And the Copyright header is objectively wrong,

How so?

hahnjo: I'll share my reasoning here, but this is not the place to discuss: a) Rene and Fons don't have copyright on files that were created after their active time in the project (and they likely don't even have it for prior contributions, which are owned by the respective contributor and / or usually their employer at the time). b) The year 1995 is wrong in most files because they were created after. c) The usage of dates is generally problematic because it would need updating every year. This is why many projects remove them, see for example https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/.

As I said, this is not the place to discuss. If this PR cannot go in without, I will pause development on `histv7` until it is sorted out.

pcanal: b and c are 'technical' details that could be fixed internally.  a is indeed a much larger (legal) discussion whose resolution in ongoing and involves all the stakeholders that needs to be done offline.

siliataider (approved):
Let's go histv7 🚀

PR #19337, opened 2025-07-11
[ntuple] Remove `/bigobj` flag for `RField.cxx`

hahnjo:
The file was split some time ago, it should not exceed limits anymore.


bellenot (commented):
I'll have to check

jblomer (approved):

hahnjo:
@bellenot is this ok to merge? The CI seems happy, which matches my expectation. If you need more time to check, that's also fine - the change is not on any critical path for me, just something I noticed while passing by

PR #19339, opened 2025-07-11
[ntuple] Correct `RValue::GetRef<T>` usage or use `RValue::GetPtr<void>`

hahnjo:
See the individual commits for the case-by-case changes and explanations.


tree/ntuple/inc/ROOT/RNTupleView.hxx
```diff
@@ -345,6 +345,13 @@ private:
       return fieldId;
    }
 
+   std::uint64_t GetCardinalityValue()
```
silverweed: Can this be `const`?

PR #19341, opened 2025-07-11
[df] Introduce RDatasetSpec for RNTuple

martamaja10:
This PR introduces functionality of RDatasetSpec for RNTuple. Main changes are inside RLoopManager and RNTupleDS - mainly the introduction of WithGlobalRanges. 

Marked as WIP to make sure all tests pass but the main development should be ready for the review.


tree/dataframe/inc/ROOT/RNTupleDS.hxx
```diff
@@ -104,12 +104,18 @@ class RNTupleDS final : public ROOT::RDF::RDataSource {
    std::vector<std::vector<ROOT::Internal::RDF::RNTupleColumnReader *>> fActiveColumnReaders;
 
    ULong64_t fSeenEntries = 0;                ///< The number of entries so far returned by GetEntryRanges()
+   ULong64_t fSeenEntriesRange = 0;
+   ULong64_t fSeenEntriesNonRange = 0;
```
hageboeck: I find the name hard to understand. A small comment could help or if it's not too much work, one could rename it to something like `fSeenEntriesOutOfRange` (if that's what it does).

tree/dataframe/inc/ROOT/RNTupleDS.hxx
```diff
@@ -104,12 +104,18 @@ class RNTupleDS final : public ROOT::RDF::RDataSource {
...
+   ULong64_t counterFileEmpty = 0;
```
hageboeck: Maybe also start it with `fCounter...`?

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -450,41 +495,96 @@ std::optional<std::string> GetRedirectedSampleId(std::string_view path, std::str
...
+
+         for (unsigned int slot{}; slot < fNSlots; slot++) {
+            for (auto &v : fDatasetColumnReaders[slot])
+               v.second.reset();
+         }
+
+      } else {
+         throw std::runtime_error(
+            "More than one RNTuple name was found, please make sure to use RNTuples with the same RnTuple name.");
+      }
```
hageboeck: Is this a temporary limitation?

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -534,13 +537,17 @@ void ROOT::RDF::RNTupleDS::StageNextSources()
...
 {
    assert(fNextRanges.empty());
+
    auto nFiles = fFileNames.empty() ? 1 : fFileNames.size();
    auto nRemainingFiles = nFiles - fNextFileIndex;
+
    if (nRemainingFiles == 0)
       return;
 
+   unsigned int nEntries = 0;
```
hageboeck: To be safe, I would make this a size_t

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -554,12 +561,11 @@ void ROOT::RDF::RNTupleDS::PrepareNextRanges()
          range.fFileName = fFileNames[fNextFileIndex];
          range.fSource->Attach();
          fNextFileIndex++;
-
-         auto nEntries = range.fSource->GetNEntries();
+         nEntries = range.fSource->GetNEntries();
```
hageboeck: `auto nEntries =`? It looks like if could be declared only in the loop.

tree/dataframe/src/RNTupleDS.cxx
hageboeck: Actually, with the below comment, you might be able to remove it.

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -681,29 +689,104 @@ std::vector<std::pair<ULong64_t, ULong64_t>> ROOT::RDF::RNTupleDS::GetEntryRange
...
       if (fCurrentRanges[i].fFirstEntry == 0) {
          // New source
-         fSeenEntries += nEntriesPerSource;
+         fSeenEntriesNonRange += nEntriesPerSource;
          nEntriesPerSource = 0;
       }
-      auto start = fCurrentRanges[i].fFirstEntry + fSeenEntries;
-      auto end = fCurrentRanges[i].fLastEntry + fSeenEntries;
+
+      start = fCurrentRanges[i].fFirstEntry + fSeenEntriesNonRange;
```
hageboeck: Does it need to be out of the loop?

tree/dataframe/test/dataframe_datasetspec.cxx
```diff
@@ -832,6 +836,228 @@ TEST_P(RDatasetSpecTest, TreeInSubdir)
...
+         RDatasetSpec spec;
+         spec.AddSample(samp);
+         spec.AddSample(samp1);
+         auto df_error = ROOT::RDataFrame(spec);
+      }
+
+      catch (const std::runtime_error &err) {
+         EXPECT_EQ(
+            std::string(err.what()),
+            "More than one RNTuple name was found, please make sure to use RNTuples with the same RnTuple name.");
```
hageboeck: Not specifically for this PR, but general for RDF devs:
I always wonder if it makes sense to really test the entire error message. In this way, we always have to touch two places if we wanted to correct small things, such as in this case:
`RnTuple --> RNTuple`

I would either just test that there is any exception (possibly of a specific type), or just test a substring such as `exc.startsWith("More than one RNTuple name");`. In this way, small adjustments or making the error message more helpful doesn't immediately fail the test.

hageboeck (commented):
I see a few shadow warnings in the CI, but you probably saw them, too.

tree/dataframe/inc/ROOT/RNTupleDS.hxx
martamaja10: no, it's not what it does, it counts entries if no global range is requested, I'll add a comment, I could also rename it but I'm not sure what to rename it to to be more evident. It used to be just fSeenEntries which was simply for counting processed entries but now I need to have two separate counters depending on the case. I still need the fSeenEntries to count correctly at the end of processing but depending on whether I have the global ranges or not, I need to count differently first.

martamaja10: maybe fSeenEntriesNonRange --> fSeenEntriesNoGlobalRange and fSeenEntriesRange --> fSeenEntriesWithGlobalRange

tree/dataframe/src/RNTupleDS.cxx
martamaja10: yes, I use it also after the loop in this if statement block: https://github.com/root-project/root/blob/058f1bb150a914ba278c543c1b6aa26d359f0375/tree/dataframe/src/RNTupleDS.cxx#L776

tree/dataframe/src/RLoopManager.cxx
martamaja10: yes, but this is what we have in any RNTuple chaining at this point.

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -450,41 +495,96 @@ std::optional<std::string> GetRedirectedSampleId(std::string_view path, std::str
...
+   } else if (inFile->Get<ROOT::RNTuple>(datasetName[0].data())) {
+
+      // Change the range of entries to be processed
+      fBeginEntry = spec.GetEntryRangeBegin();
+      fEndEntry = spec.GetEntryRangeEnd();
+
+      // Store the samples
+      fSamples = spec.MoveOutSamples();
+      fSampleMap.clear();
```
vepadulano: These lines are shared between the two branches of the `if`, should be extracted

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -450,41 +495,96 @@ std::optional<std::string> GetRedirectedSampleId(std::string_view path, std::str
...
+         fDataSource->SetNSlots(fNSlots);
+
+         for (unsigned int slot{}; slot < fNSlots; slot++) {
+            for (auto &v : fDatasetColumnReaders[slot])
+               v.second.reset();
+         }
+
+      } else {
+         throw std::runtime_error(
+            "More than one RNTuple name was found, please make sure to use RNTuples with the same RnTuple name.");
```
vepadulano:
```suggestion
            "More than one RNTuple name was found, please make sure to use RNTuples with the same name.");
```

tree/dataframe/src/RLoopManager.cxx
```diff
@@ -450,41 +495,96 @@ std::optional<std::string> GetRedirectedSampleId(std::string_view path, std::str
...
-      for (auto &v : fDatasetColumnReaders[slot])
-         v.second.reset();
+
+      fDataSource = std::make_unique<ROOT::Internal::RDF::RTTreeDS>(std::move(chain), spec.GetFriendInfo());
+
+      fDataSource->SetNSlots(fNSlots);
+      for (unsigned int slot{}; slot < fNSlots; slot++) {
+         for (auto &v : fDatasetColumnReaders[slot])
+            v.second.reset();
+      }
```
vepadulano: This part is also shared and can be extracted to avoid repetition

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -534,13 +537,18 @@ void ROOT::RDF::RNTupleDS::StageNextSources()
...
    auto nRemainingFiles = nFiles - fNextFileIndex;
+
    if (nRemainingFiles == 0)
       return;
 
    // Easy work scheduling: one file per slot. We skip empty files (files without entries).
-   if (nRemainingFiles >= fNSlots) {
+
+   if ((nRemainingFiles >= fNSlots) ||
+       (fGlobalEntryRange.has_value() && fGlobalEntryRange->first <= std::numeric_limits<Long64_t>::max() &&
```
vepadulano:
```suggestion
       (fGlobalEntryRange.has_value() && fGlobalEntryRange->first > 0 &&
```

vepadulano: I think there should be no scenario where `fGlobalEntryRange->first <= std::numeric_limits<Long64_t>::max()` is false, unless I'm missing something

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -681,29 +690,105 @@ std::vector<std::pair<ULong64_t, ULong64_t>> ROOT::RDF::RNTupleDS::GetEntryRange
...
-      auto end = fCurrentRanges[i].fLastEntry + fSeenEntries;
+
+      start = fCurrentRanges[i].fFirstEntry + fSeenEntriesNoGlobalRange;
+      auto end = fCurrentRanges[i].fLastEntry + fSeenEntriesNoGlobalRange;
+
       nEntriesPerSource += end - start;
 
-      fFirstEntry2RangeIdx[start] = i;
-      ranges.emplace_back(start, end);
+      if (fGlobalEntryRange.has_value() && fGlobalEntryRange->first <= std::numeric_limits<Long64_t>::max() &&
```
vepadulano: Similar comment here about `fGlobalEntryRange->first <= std::numeric_limits<Long64_t>::max()`, when is this false?

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -534,13 +537,18 @@ void ROOT::RDF::RNTupleDS::StageNextSources()
...
+
    if (nRemainingFiles == 0)
       return;
 
    // Easy work scheduling: one file per slot. We skip empty files (files without entries).
-   if (nRemainingFiles >= fNSlots) {
+
+   if ((nRemainingFiles >= fNSlots) ||
+       (fGlobalEntryRange.has_value() && fGlobalEntryRange->first <= std::numeric_limits<Long64_t>::max() &&
+        fGlobalEntryRange->second <= std::numeric_limits<Long64_t>::max())) {
```
vepadulano: To be honest also this condition looks sketchy. `fGlobalEntryRange->second` is either equal to `std::numeric_limits<Long64_t>::max()` (the default, when the user has not provided a range) or it follows that it will be less than that. So my intuition tells me that also this condition will be always true, when is it not?

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -681,29 +690,105 @@ std::vector<std::pair<ULong64_t, ULong64_t>> ROOT::RDF::RNTupleDS::GetEntryRange
...
-      fFirstEntry2RangeIdx[start] = i;
-      ranges.emplace_back(start, end);
+      if (fGlobalEntryRange.has_value() && fGlobalEntryRange->first <= std::numeric_limits<Long64_t>::max() &&
+          fGlobalEntryRange->second <= std::numeric_limits<Long64_t>::max()) {
+
+         // We need to consider different scenarios for when we have GlobalRanges set by the user.
+         // Consider a simple case of 3 files, with original ranges set as (consecutive entries of 3 files):
+         // [0, 20], [20, 45], [45, 65]
+         // we will now see what happens in each of the scenarios below when GlobalRanges can be set to different
+         // values: a) [2, 5] - we stay in file 1
```
vepadulano:
```suggestion
         // values:
         // a) [2, 5] - we stay in file 1
```

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -721,11 +806,22 @@ void ROOT::RDF::RNTupleDS::InitSlot(unsigned int slot, ULong64_t firstEntry)
...
+          fGlobalEntryRange->second <= std::numeric_limits<Long64_t>::max()) {
+         r->Connect(*fCurrentRanges[idxRange].fSource,
+                    fOriginalRanges[idxRange].first - fCurrentRanges[idxRange].fFirstEntry);
+      } else {
+         r->Connect(*fCurrentRanges[idxRange].fSource, firstEntry - fCurrentRanges[idxRange].fFirstEntry);
+      }
+      // during testing turned out that this is needed in a case like:
+      // std::vector<std::string> fileNames(1000, fname);
+      // auto df1000 = ROOT::RDataFrame(name, fileNames);
+      // df1000.Sum<float>("pt").GetValue()
```
vepadulano: I don't understand what this comment is referring to

tree/dataframe/test/dataframe_datasetspec.cxx
```diff
@@ -832,6 +836,228 @@ TEST_P(RDatasetSpecTest, TreeInSubdir)
...
+   auto samp = ROOT::RDF::Experimental::RSample("mysample", "ntuple", prefix + "0.root");
+   auto samp1 = ROOT::RDF::Experimental::RSample("mysample1", "mytuple", "rntuple_wrong.root");
+
+   RDatasetSpec spec;
+   spec.AddSample(samp);
+   spec.AddSample(samp1);
+
+   EXPECT_THROW(
+      try {
+         auto samp = ROOT::RDF::Experimental::RSample("mysample", "ntuple", prefix + "0.root");
```
vepadulano, line 1011: Careful to warnings about shadowed local variable here

tree/dataframe/test/dataframe_datasetspec.cxx
```diff
@@ -832,6 +836,228 @@ TEST_P(RDatasetSpecTest, TreeInSubdir)
...
+         auto samp1 = ROOT::RDF::Experimental::RSample("mysample1", "mytuple", "rntuple_wrong.root");
```
vepadulano, line 1012: Same thing here

tree/dataframe/test/dataframe_datasetspec.cxx
```diff
@@ -832,6 +836,228 @@ TEST_P(RDatasetSpecTest, TreeInSubdir)
...
+
+         RDatasetSpec spec;
```
vepadulano, line 1014: Same thing here

tree/dataframe/test/dataframe_datasetspec.cxx
```diff
@@ -832,6 +836,228 @@ TEST_P(RDatasetSpecTest, TreeInSubdir)
...
+   }
+
+   auto samp = ROOT::RDF::Experimental::RSample("mysample", "ntuple", "rdatasetspec_rntuple0.root");
+   RDatasetSpec spec;
+   spec.AddSample(samp);
+   auto df1 = ROOT::RDataFrame(spec);
+
+   auto mean = df1.Mean("x");
+
+   EXPECT_EQ(h.GetMean(), mean.GetValue());
```
vepadulano: Probably we want
```suggestion
   EXPECT_FLOAT_EQ(h.GetMean(), mean.GetValue());
```

tree/dataframe/test/spec_rntuple.json
```diff
@@ -0,0 +1,37 @@
...
+         ],
+         "files": [
+            "rdatasetspec_rntuple2.root"
+         ],
+         "metadata": {
+            "lum": 30.0
+         }
+      }
+   }
+}
```
vepadulano: Missing newline at end of file

tree/dataframe/src/RNTupleDS.cxx
martamaja10: I think you are right indeed, I took it from RTTreeDS.cxx but here indeed we may simplify this.

tree/dataframe/src/RNTupleDS.cxx
martamaja10: I again followed what was done for the TTree, but this is indeed not needed here.

tree/dataframe/src/RNTupleDS.cxx
vepadulano: I'm not sure I follow, `RTTreeDS.cxx` checks for the presence of the range via `fGlobalEntryRange.has_value()`

vepadulano: Ok I understand the source of confusion:

`fGlobalEntryRange` is a data member of `RDataSource` base class with type `std::optional<std::pair<ULong64_t, ULong64_t>>`.

In RTTreeDS, we are processing TTree data, where the type used to represent and entry number is `Long64_t`.

In RTTreeDS.cxx at  ROOT::Internal::RDF::RTTreeDS::Initialize() at https://github.com/root-project/root/blob/00565c678832d07c6c73cd2d732236d11c87b710/tree/dataframe/src/RTTreeDS.cxx#L435 the following checks
```
      if (fGlobalEntryRange.has_value() && fGlobalEntryRange->first <= std::numeric_limits<Long64_t>::max() &&
          fGlobalEntryRange->second <= std::numeric_limits<Long64_t>::max()
```

Check three different things, in order:
* fGlobalEntryRange.has_value() checks for the presence of the range, i.e. if the user provided it or not
* fGlobalEntryRange->first <= std::numeric_limits<Long64_t>::max() checks that the `ULong64_t` value contained in `fGlobalEntryRange->first` can be actually represented by a Long64_t
* fGlobalEntryRange->second <= std::numeric_limits<Long64_t>::max() checks that the `ULong64_t` value contained in `fGlobalEntryRange->second` can be actually represented by a Long64_t

So the checks are not controlling the actual values in terms of what the user provided, but it's just a range check for the different representation of the entry number used by TTree w.r.t. the other data sources.

Admittedly, the check is a bit obscure and could deserve a comment, thanks for making me notice this!

martamaja10: indeed, thanks for a clear explanation now! tbh I should have spent a bit more time asking myself if this is really needed instead of blindly following. But we have the resolution now!

tree/dataframe/src/RNTupleDS.cxx
martamaja10: From most of my initial tests, I figured that we could do without the if else statement and I treated everything as if we had the global ranges (`r->Connect(*fCurrentRanges[idxRange].fSource, fOriginalRanges[idxRange].first - fCurrentRanges[idxRange].fFirstEntry);`). 

However, one test in datasource_ntuple.cxx, RNTupleDSTest.ChainMT, in which an rdf is constructed as shown in the comment:
```
std::vector<std::string> fileNames(1000, fname);
auto df1000 = ROOT::RDataFrame(name, fileNames);
df1000.Sum<float>("pt").GetValue();
```
fails.  
 
For the time being, I left original processing for the cases without the GlobalRange so that the test doesn't fail. If you think this should be available to the cases with GlobalRange, we can discuss offline maybe.

vepadulano (approved):
Very nice thank you!! 🚀

vepadulano:
Before merging please squash the commits :+1:

PR #19342, opened 2025-07-11
[ntuple] Check type in `RValue::GetRef<T>` and `GetPtr<T>`

### LINKED ISSUE #18316: [ntuple] RValue::GetRef<T> should type-check

hahnjo:
This requires splitting `RValue` and `RBulkValues` into a separate header. After the changes, the organization is as follows:
 * `RFieldBase.hxx`: defines `RFieldBase` and forward-declares `RFieldBase::RValue` and `RFieldBase::RBulkValues`
 * `RField.hxx`: includes `RFieldBase.hxx` and defines `RField<T>` with all specializations; they provide `RField<T>::TypeName()`
 * `RFieldBaseRValue.hxx`: includes both `RFieldBase.hxx` and `RField.hxx`, defines `RValue` and `RBulkValues`; the typed API uses `RField<T>::TypeName()`

Closes #18316

---

This PR builds on https://github.com/root-project/root/pull/19339, the first five commits should be reviewed there. I put this PR as Draft until it is merged and we are sure we want to organize the headers like this.


tree/ntuple/inc/ROOT/RFieldBaseRValue.hxx
```diff
@@ -0,0 +1,189 @@
...
+   std::shared_ptr<void> fObjPtr;
+   RValue(RFieldBase *field, std::shared_ptr<void> objPtr) : fField(field), fObjPtr(objPtr) {}
+
+   template <typename T>
+   void EnsureMatchingType() const
+   {
+      if constexpr (!std::is_void_v<T>) {
+         if (fField->GetTypeName() != ROOT::RField<T>::TypeName()) {
+            throw RException(
+               R__FAIL("type mismatch: " + fField->GetTypeName() + " vs. " + ROOT::RField<T>::TypeName()));
```
enirolf, line 47: I think the error message could be a bit more descriptive, maybe something like:

```suggestion
               R__FAIL("type mismatch for field \"" + fField->GetFieldName + "\": expected " + fField->GetTypeName() + ", got " + ROOT::RField<T>::TypeName()));
```

tree/ntuple/inc/ROOT/RFieldBaseRValue.hxx
```diff
@@ -34,6 +38,17 @@ private:
    std::shared_ptr<void> fObjPtr;
    RValue(RFieldBase *field, std::shared_ptr<void> objPtr) : fField(field), fObjPtr(objPtr) {}
 
+   template <typename T>
+   void EnsureMatchingType() const
+   {
+      if constexpr (!std::is_void_v<T>) {
+         if (fField->GetTypeName() != ROOT::RField<T>::TypeName()) {
```
jblomer, line 45: When #19323 is merged, I think we want to use `ROOT::Internal::IsMatchingFieldType`. So maybe schedule the two PRs one after the other?

tree/ntuple/inc/ROOT/RFieldBaseRValue.hxx
jblomer, line 1: Since this anyway covers both `RValue` and `RBulkValues`, how about calling the header `RFieldBaseValue.hxx`? I think it would look slightly nicer to me.

jblomer (commented):
Nice! Since this can break user code, I think we add an entry to the release notes.

tree/ntuple/inc/ROOT/RFieldBaseRValue.hxx
hahnjo: Yes right, and we probably also want to extend `roottest/root/ntuple/atlas-datavector/` to check that the procedure is working. Let's pause this PR for the moment and focus on https://github.com/root-project/root/pull/19323 which seems close to done.

jblomer:
Thinking again about it: there is a possible alternative solution using a new, internal helper class, e.g. `RFieldTypeName<T>`. This class would only take care of the type name and would be used both by `RField<T>` and `RValue`. The drawback is that we need to duplicate the template specializations for the new helper class but the advantage is that we don't change the user-visible includes.

hahnjo:
> Thinking again about it: there is a possible alternative solution using a new, internal helper class, e.g. `RFieldTypeName<T>`. This class would only take care of the type name and would be used both by `RField<T>` and `RValue`. The drawback is that we need to duplicate the template specializations for the new helper class but the advantage is that we don't change the user-visible includes.

Yes, it's a possibility. However, it probably requires duplicating (some of) `RIntegralField` and `RIntegralTypeMap`. We can give it a try and see if we like it better.

tree/ntuple/inc/ROOT/RFieldBaseRValue.hxx
hahnjo: Addressed in https://github.com/root-project/root/pull/19544

hahnjo:
After more discussion, we settled on a different approach now implemented in https://github.com/root-project/root/pull/19544.

PR #19344, opened 2025-07-11
[build] Simplify CMake in ATLAS DataVector test directory

vepadulano:
Using the `NO_CXXMODULE` option of `ROOTTEST_GENERATE_DICTIONARY` produces dictionary artifacts that correctly reproduce the ATLAS environment (i.e. there are no C++ module neither modulemap produced in the test build directory).

### DIFF
--- a/roottest/root/ntuple/atlas-datavector/CMakeLists.txt
+++ b/roottest/root/ntuple/atlas-datavector/CMakeLists.txt
@@ -15,28 +15,13 @@ endif()
 # the ATLAS Athena build. Contributors
 # - It always builds the C++ module from the dictionary (in a module build of ROOT)
 # - It changes names of artifacts (e.g. removes the "lib" prefix from the shared library)
-set(CMAKE_ROOTTEST_DICT OFF)
-set(ATLASLIKEDATAVECTOR_DICTNAME AtlasLikeDataVectorDict)
-ROOT_GENERATE_DICTIONARY(
-	${ATLASLIKEDATAVECTOR_DICTNAME}
-	${CMAKE_CURRENT_SOURCE_DIR}/AtlasLikeDataVector.hxx
-	LINKDEF ${CMAKE_CURRENT_SOURCE_DIR}/selection.xml
+ROOTTEST_GENERATE_DICTIONARY(
+    AtlasLikeDataVectorDict
+    NO_CXXMODULE
+    ${CMAKE_CURRENT_SOURCE_DIR}/AtlasLikeDataVector.hxx
+    LINKDEF ${CMAKE_CURRENT_SOURCE_DIR}/selection.xml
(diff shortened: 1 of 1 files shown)


dpiparo (dismissed):
thanks for these changes. I am approving the PR, with the caveat that the debian failure perhaps deserves a closer look before merging.

vepadulano:
Indeed the failure https://github.com/root-project/root/actions/runs/16219472129/job/45796305238?pr=19344#step:11:12437 is worrying
```
   ============================= test session starts ==============================
  platform linux -- Python 3.9.21, pytest-8.4.1, pluggy-1.6.0
  rootdir: /github/home/ROOT-CI
  plugins: anyio-4.9.0
  collected 1 item
  
  ../../../../../src/roottest/root/ntuple/atlas-datavector/template_instantiation.py F [100%]
  
  =================================== FAILURES ===================================
  _________ TestTemplateInstantiation.test_instantiate_function_template _________
  
  self = <template_instantiation.TestTemplateInstantiation object at 0x7f005ea90d30>
  
      def test_instantiate_function_template(self):
          import ROOT
      
          ROOT.gInterpreter.Declare(r"template <typename T> void foo() {}")
      
          # The call raises an exception that the function cannot be found.
          # What is really happening is that a template instantiation is tried
          # with the "wrong" signature, as the class type passed by the user is
          # not what the compiler sees.
          field_type_name = "AtlasLikeDataVector<CustomStruct>"
          with pytest.raises(TypeError):
              ROOT.foo[field_type_name]()
      
          # The first attempt at instantiating the template has had the side
          # effect of loading the dictionary information for AtlasLikeDataVector,
          # including the alternative class type names
          alt_field_type_names = ROOT.TClassTable.GetClassAlternativeNames(field_type_name)
      
          fully_qualified_type_name = "AtlasLikeDataVector<CustomStruct, DataModel_detail::NoBase>"
      
  >       assert len(alt_field_type_names) == 1
  E       assert 0 == 1
  E        +  where 0 = len(vector<string>{})
  
  ../../../../../src/roottest/root/ntuple/atlas-datavector/template_instantiation.py:25: AssertionError
  =========================== short test summary info ============================
  FAILED ../../../../../src/roottest/root/ntuple/atlas-datavector/template_instantiation.py::TestTemplateInstantiation::test_instantiate_function_template - assert 0 == 1
   +  where 0 = len(vector<string>{})
  ============================== 1 failed in 1.71s ===============================
  CMake Error at /github/home/ROOT-CI/build/RootTestDriver.cmake:232 (message):
    error code: 1
```

It's unclear to me whether it happens because of the changes in this PR.

vepadulano:
The problem was understood, there was a missing `FIXTURES_REQUIRED`. The second Python unittest does not need the RNTuple file, but it still needs the class dictionary. The second commit fixes the issue.

pcanal:
Thanks

PR #19345, opened 2025-07-11
[clad] Bump to v2.0

vgvassilev:
The new release introduces improvements across both forward and reverse automatic differentiation modes, including the adoption of static scheduling by default and expanded support for custom derivatives, such as those involving STL containers and iterators. The release enhances compatibility with Clang versions 10 through 20 and extends support for CUDA through updated tape handling. Additional updates include improved error diagnostics, support for more complex object types, and infrastructure for benchmarking.


guitargeek (approved):
Thank you very much!

PR #19351, opened 2025-07-13
[tree] Avoid duplicated tests in treeplayer

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/19346

Also re-enable Windows in some of them by being more fine-grained

## Checklist:

- [ ] tested changes locally
- [ ] updated the docs (if necessary)

PR #19353, opened 2025-07-14
[cmake] Export root and root.exe

bellenot:
Export root and root.exe, following [this discussion](https://root-forum.cern.ch/t/removal-of-some-cmake-targets-in-root-6-36/63885)

### DIFF
--- a/main/CMakeLists.txt
+++ b/main/CMakeLists.txt
@@ -16,7 +16,7 @@ if(NOT WIN32)
   endif()
   ROOT_EXECUTABLE(roots.exe roots.cxx LIBRARIES Core MathCore CMAKENOEXPORT)
 endif()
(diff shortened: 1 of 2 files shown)


ferdymercury:
alma10 failure would be fixed by https://github.com/root-project/root/pull/19351

hageboeck (approved):
:+1:

aphecetche:
@bellenot is it expected that this fix is not (yet ?) in 6.36 ?

aphecetche:
@bellenot is it expected that this fix is not (as far as I can tell) in 6.36.04 ?

hageboeck:
> @bellenot is it expected that this fix is not (as far as I can tell) in 6.36.04 ?

Hello @aphecetche, indeed, for the moment, the export of these executables will only be in v6.38.00 onwards (expected in two to three weeks from now). Will that be sufficient?

aphecetche:
@hageboeck well, yes, I guess that's ok. Thanks.

PR #19355, opened 2025-07-14
[main] rootls: add flag to show RNTuple field hierarchy

silverweed:
# This Pull request:
adds a `-R` flag to `rootls` to print RNTuple field trees recursively (similarly to the -t flag for TTree). Fields are sorted by name.

Sample output (cut for brevity):
```
rootls -R ~/src/iotools/data/gg_data\~zstd.ntuple
ROOT::RNTuple  Dec 18 15:55 2024 mini;1 ""
  SumWeights                 float
  XSection                   float
  channelNumber              std::int32_t
  ditau_m                    float
  eventNumber                std::int32_t
  jet_E                      std::vector<float>
    _0.......................float
  jet_MV2c10                 std::vector<float>
    _0.......................float
  jet_eta                    std::vector<float>
    _0.......................float
  jet_jvt                    std::vector<float>
    _0.......................float
  jet_n                      std::uint32_t
  jet_phi                    std::vector<float>
    _0.......................float
  jet_pt                     std::vector<float>
    _0.......................float
  jet_pt_syst                std::vector<float>
    _0.......................float
  jet_trueflav               std::vector<std::int32_t>
    _0.......................std::int32_t
  jet_truthMatched           std::vector<bool>
    _0.......................bool
  largeRjet_D2               std::vector<float>
    _0.......................float
```

## Checklist:

- [x] tested changes locally


main/src/rootls.cxx
```diff
@@ -73,6 +75,7 @@ positional arguments:
   -t, --treeListing     Print tree recursively and use a long listing format.
   -r, --recursiveListing
                         Traverse file recursively entering any TDirectory.
+  -R, --rntupleListing  Print RNTuples recursively and use a long listing format.
```
enirolf: Perhaps swap the options, so RNTuple will be directly below TTree?

silverweed:
>  use ntuple instead of rntuple (and with it maybe use -n instead of -R)?

A problem I see with that is that technically we have other "ntuple" classes other than RNTuple (like TNTuple), so maybe we want to be extra specific (it's also the name that people who know it know it by so perhaps more familiar and immediately recognizable)

enirolf:
> > use ntuple instead of rntuple (and with it maybe use -n instead of -R)?
> 
> A problem I see with that is that technically we have other "ntuple" classes other than RNTuple (like TNTuple), so maybe we want to be extra specific (it's also the name that people who know it know it by so perhaps more familiar and immediately recognizable)

Yes, that actually also came up in my mind after posting this comment. It's a good reason to keep it like this, then!

enirolf (approved):
LGTM!

main/src/rootls.cxx
```diff
@@ -156,11 +159,12 @@ struct RootLsSource {
...
-      kOneColumn = 1,
-      kLongListing = 2,
-      kTreeListing = 4,
-      kRecursiveListing = 8
+      kNone = 0x0,
+      kOneColumn = 0x1,
+      kLongListing = 0x2,
+      kTreeListing = 0x4,
+      kRecursiveListing = 0x8,
+      kRNTupleListing = 0x10,
```
jblomer: Perhaps swap the two constants.

main/src/rootls.cxx
```diff
@@ -269,6 +273,41 @@ static void PrintClusters(std::ostream &stream, TTree &tree, Indent indent)
...
 
+// Prints an RNTuple field tree recursively.
+static void PrintRNTuple(std::ostream &stream, const ROOT::RNTupleDescriptor &desc, Indent indent,
+                         const ROOT::RFieldDescriptor &rootField, std::size_t minNameLen = 0,
+                         std::size_t minTypeLen = 0)
+{
+   std::size_t maxNameLen = 0, maxTypeLen = 0;
+   std::vector<const ROOT::RFieldDescriptor *> fields;
+   fields.reserve(rootField.GetLinkIds().size());
+   for (const auto &childId : rootField.GetLinkIds()) {
```
jblomer: We already have `GetFieldIterable` with a comparator parameter. Here would be a use case for this.

main/src/rootls.cxx
```diff
@@ -332,13 +371,23 @@ static void PrintNodesDetailed(std::ostream &stream, const RootLsTree &tree,
...
+         }
+      }
+      if (flags & RootLsArgs::kRNTupleListing) {
+         if (ClassInheritsFrom(child.fClassName.c_str(), "ROOT::RNTuple")) {
+            auto *rntuple = child.fKey->ReadObject<ROOT::RNTuple>();
+            if (rntuple) {
+               auto reader = ROOT::RNTupleReader::Open(*rntuple);
+               const auto &desc = reader->GetDescriptor();
+               PrintRNTuple(stream, desc, indent + 2, desc.GetFieldZero());
             }
```
jblomer: Perhaps warn or throw if `!rntuple`

jblomer (commented):
Nice! I think we should add a `roottest` for this.

main/src/rootls.cxx
silverweed: I can use `GetFieldIterable()` to save a line, yes, but why would I need the comparator?

pcanal:
> (and with it maybe use -n instead of -R)?

Why isn't both `-t` and `-R` spelt `-r[ecursive]`?  Why make the distinction between the 2 containers?

silverweed:
@pcanal `-r` already has the meaning of "show every object in the hierarchy including subdirectories" (but without expanding them). That's a useful option to have without necessarily expanding all internal structures of trees and rntuples.

pcanal (approved):

PR #19357, opened 2025-07-14
[ci] Use alma10 for the clang special build

guitargeek:
Use alma10 for the clang special build so that we use a newer clang compiler that will hopefully give us new relevant warnings, and also try to use the stock `alma10` image so that we don't have the duplication like with `alma9` and `alma9-clang`. Installing the `clang` package in the regular image and then setting the correct CMake flags should do the trick.


dpiparo (approved):
thanks for these changes!

PR #19360, opened 2025-07-14
[RF] Restrict VDT usage to architecture-specific BatchCompute libraries.

hageboeck:
- Remove VDT from "main" RooBatchCompute library. Although the headers were included, they were not used.
- Instead, include VDT in the architecture-specific libraries.

In the past, the architecture-specific libraries got their VDT from a PUBLIC target_link_libraries of RooBatchCompute. This is technically not correct, because this one doesn't use any VDT. Now, the VDT usage is only in the implementation of the architecture-specific libraries. This fixes a build failure when VDT headers are not findable in standard system locations or builtin in ROOT.

https://its.cern.ch/jira/browse/SPI-2851


roofit/batchcompute/CMakeLists.txt
```diff
@@ -49,3 +44,11 @@ if (cuda)
   ROOT_LINKER_LIBRARY(RooBatchCompute_CUDA  ${shared_object_sources_cu} TYPE SHARED DEPENDENCIES RooBatchCompute)
```
andresailer, line 44: Does this not need VDT::VDT?

andresailer: Probably not.

roofit/batchcompute/CMakeLists.txt
```diff
@@ -49,3 +44,11 @@ if (cuda)
...
   target_compile_options(RooBatchCompute_CUDA  PRIVATE -lineinfo --expt-relaxed-constexpr)
 endif()
+
+if(vdt OR builtin_vdt)
+  foreach(library in RooBatchCompute_GENERIC RooBatchCompute_SSE4.1 RooBatchCompute_AVX RooBatchCompute_AVX2 RooBatchCompute_AVX512 )
```
andresailer:
```suggestion
  foreach(library RooBatchCompute_GENERIC RooBatchCompute_SSE4.1 RooBatchCompute_AVX RooBatchCompute_AVX2 RooBatchCompute_AVX512 )
```

https://cmake.org/cmake/help/latest/command/foreach.html#foreach
```cmake
foreach(<loop_var> <items>)
  <commands>
endforeach()
```

guitargeek (approved):
Thank you very much!

hageboeck:
The notebook-related failures on fedora are unrelated, since they happen also in other builds.

PR #19364, opened 2025-07-14
[DF] Allow snapshotting from TTree to RNTuple

enirolf:
Snapshotting from TTree to RNTuple was initially disabled, because it was assumed that this would require adopting (parts of) the machinery of the RNTupleImporter to make it work. However, this is all already taken care of by the TTree's datasource.

One caveat is the support of leaflist branches. It is not possible to snapshot these fully, and users always need to explicitly pass their leaf names to `Snapshot`. This, however, is something fundamentally unsupported by the RDataFrame. The reason is that for snapshotting to TTree, it is impossible/overly complex to correctly set the branch addresses in the output TTree for these branches at runtime, without JITting. For RNTuple, however, this is likely more doable by using untyped record fields (similar to how it is done in the importer). This requires some more investigation, however, and if indeed possible will be addressed in a follow-up PR.


tree/dataframe/test/dataframe_snapshot_ntuple.cxx
```diff
@@ -35,6 +35,15 @@ class FileRAII {
    std::string GetPath() const { return fPath; }
 };
 
+template <typename T>
+void expect_vec_eq(const ROOT::RVec<T> &v1, const ROOT::RVec<T> &v2)
+{
+   ASSERT_EQ(v1.size(), v2.size()) << "Vectors 'v1' and 'v2' are of unequal length";
+   for (std::size_t i = 0ull; i < v1.size(); ++i) {
+      EXPECT_EQ(v1[i], v2[i]) << "Vectors 'v1' and 'v2' differ at index " << i;
```
hahnjo: Just a heads up that we may need `EXPECT_FLOAT_EQ` here, depending on what `T` is. I think for the moment it's fine because `{ 1.f, 2.f, 3.f }` are exactly representable.

hahnjo (commented):
Great! LG from an RNTuple perspective, I let RDataFrame experts comment if there's anything else to test in particular (but I think you already discussed offline)

vepadulano (approved):
Thank you!

tree/dataframe/test/dataframe_snapshot_ntuple.cxx
```diff
@@ -511,19 +456,39 @@ void WriteTestTree(const std::string &tname, const std::string &fname)
...
-
    t.Write();
 }
 
-TEST(RDFSnapshotRNTuple, DisallowFromTTreeTemplated)
+TEST(RDFSnapshotRNTuple, FromTTree)
 {
    const auto treename = "tree";
-   FileRAII fileGuard{"RDFSnapshotRNTuple_disallow_from_ttree_templated.root"};
+   FileRAII fileGuard{"RDFSnapshotRNTuple_disallow_from_ttree.root"};
```
pcanal: I am minorly confused by the filename still containing "disallow" ....

enirolf: Oversight, fixed now :)

tree/dataframe/test/dataframe_snapshot_ntuple.cxx
```diff
@@ -544,6 +545,34 @@ TEST(RDFSnapshotRNTuple, FromTTree)
 
    auto reader = RNTupleReader::Open("ntuple", fileGuard.GetPath());
 
+   auto &descriptor = reader->GetDescriptor();
+
+   int nTopLevelFields = 0;
+   for (const auto &_ : descriptor.GetTopLevelFields()) {
```
vepadulano: I see a bunch of unused variable warnings that should be fixed

vepadulano (changes requested):
Changes are great! just a small warning to fix

tree/dataframe/test/dataframe_snapshot_ntuple.cxx
```diff
@@ -511,59 +456,159 @@ TEST(RDFSnapshotRNTuple, TDirectory)
...
+      t.Fill();
+      t.Write();
+   }
 
-   t.Write();
-}
+   void TearDown() override
+   {
+      gSystem->Unlink(fFileName.c_str());
+      std::remove(fFileName.c_str());
```
vepadulano: In principle one of these calls is redundant

PR #19369, opened 2025-07-14
Add support for Windows in the Jupyter kernel for Cling

yjoer:
This PR adds support for Windows in the Jupyter kernel. On Windows (MSVCRT), the stdout and stderr file descriptors are not exported as global symbols. To work around this, we now retrieve them using [_fdopen](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fdopen-wfdopen?view=msvc-170). Beginning with Python 3.12, we can utilize the new cross-platform API [os.set_blocking](https://docs.python.org/3.12/library/os.html#os.set_blocking) to configure pipes as non-blocking.

Since Windows does not offer a direct equivalent to the "select" system call for pipes, I replaced it with [PeekNamedPipe](https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-peeknamedpipe) to check for available data in the pipe. Unlike select, PeekNamedPipe does not allow efficient blocking. As a result, we currently poll every 0.25 seconds, aligning with the flush interval to check for new data and act accordingly.

The outcome looks appealing, with the interpreter's output and errors being correctly streamed.
image

## Checklist:

- [x] tested changes locally

### DIFF
--- a/interpreter/cling/tools/Jupyter/CMakeLists.txt
+++ b/interpreter/cling/tools/Jupyter/CMakeLists.txt
@@ -81,6 +81,10 @@ add_cling_library(libclingJupyter ${ENABLE_SHARED} ${ENABLE_STATIC}
 set_target_properties(libclingJupyter
         PROPERTIES ENABLE_EXPORTS 1)
 
+if(MSVC)
+    set_target_properties(libclingJupyter PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS 1)
+endif()
+
 if(ENABLE_SHARED)
     if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
         set(LIBCLINGJUPYTER_LINK_FLAGS " -Wl,-compatibility_version -Wl,1")
--- a/interpreter/cling/tools/Jupyter/kernel/clingkernel.py
+++ b/interpreter/cling/tools/Jupyter/kernel/clingkernel.py
@@ -24,6 +24,7 @@ import select
 import struct
 import sys
 import threading
+import time
 
 from traitlets import Unicode, Float, Dict, List, CaselessStrEnum
 from ipykernel.kernelbase import Kernel
@@ -38,6 +39,7 @@ class my_void_p(ctypes.c_void_p):
   pass
 
 if sys.platform == 'win32':
+    import msvcrt
     libc = ctypes.cdll.msvcrt
 
     class FILE(ctypes.Structure):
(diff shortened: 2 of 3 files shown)


guitargeek:
This is amazing! Can we maybe get test coverage for this by enabling some tests from here?

https://github.com/root-project/root/blob/master/roottest/python/JupyROOT/CMakeLists.txt#L1

With your development, at least the `ROOT_kernel.ipynb` should work, no?

yjoer:
I'm getting a strange error when using std::cout with the dynamic library, but it doesn't happen when I run the same code in the Cling CLI. It's got me scratching my head. Do you know what's missing?

image

image

bellenot:
@yjoer FYI, some symbols have to be explicitly exported. See for example [core/metacling/src/CMakeLists.txt#L129-L215](https://github.com/root-project/root/blob/master/core/metacling/src/CMakeLists.txt#L129-L215) and [interpreter/cling/tools/driver/CMakeLists.txt#L44-L105](https://github.com/root-project/root/blob/master/interpreter/cling/tools/driver/CMakeLists.txt#L44-L105)

yjoer:
This is incredible, thank you! I aligned the symbols with the ones in [tools/driver/CMakeLists.txt#L44-L105 
](https://github.com/root-project/root/blob/master/interpreter/cling/tools/driver/CMakeLists.txt#L44-L105) and got it working.

image

yjoer:
> With your development, at least the ROOT_kernel.ipynb should work, no?

I believe that's meant for a different kernel. This one doesn't support %cpp or %python magic commands.

interpreter/cling/tools/Jupyter/CMakeLists.txt
```diff
@@ -87,7 +87,66 @@ set_target_properties(libclingJupyter
 
 if(MSVC)
     set_target_properties(libclingJupyter PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS 1)
-endif()
+
+    # RTTI/C++ symbols
+    set(cling_exports ${cling_exports} ??_7type_info@@6B@
```
bellenot: Can you use another variable name, and not reuse `cling_exports`? Something like `jupyter_exports` maybe?

yjoer: Done

bellenot: Thanks!

yjoer:
I managed to run most parts of `ROOT_kernel.ipynb` that do not interact with the ROOT object.

https://github.com/user-attachments/assets/38301bee-df20-4f92-a315-abf53ca590cf

guitargeek:
Super cool! Why does the rest of the notebook not work? You think that can still be fixed?

What would be important for this PR is to have test coverage of what you developed, and the first half of that `ROOT_kernel.ipynb` notebook seems to do the job. How do you think we should proceed? Maybe split up the `ROOT_kernel.ipynb` in two parts, one part that works also on Windows now and then the second half that uses the `ROOT` module? Then we can enable the first part to run on Windows in the CI. Of course, supporting one whole notebook would be even better, but I don't want to ask too much of you.

yjoer:
The kernel updated here is located at https://github.com/root-project/root/tree/master/interpreter/cling/tools/Jupyter, which differs from the one that supports the root framework. I am not aware of any existing test setup for this specific kernel, either.

Issue #18554, opened 2025-04-29
Spurious TMVA related rebuild


pcanal:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Running 2 consecutive builds, the second build would ideally do nothing.  With alma9 CI configuration, a second build rebuilds this:
```
[ 96%] Built target emitFromONNX
TMVA SOFIE Operator Conv:  asymmetric padding not supported. Assume an average padding
[ 96%] Built target SofieCompileModels_ONNX
[ 96%] Building CXX object tmva/sofie/test/CMakeFiles/TestCustomModelsFromONNX.dir/TestCustomModelsFromONNX.cxx.o
[ 96%] Linking CXX executable TestCustomModelsFromONNX
[ 96%] Built target TestCustomModelsFromONNX
[ 96%] Built target emitFromROOT
TMVA SOFIE Operator Conv:  asymmetric padding not supported. Assume an average padding
[ 96%] Built target SofieCompileModels_ROOT
[ 96%] Building CXX object tmva/sofie/test/CMakeFiles/TestCustomModelsFromROOT.dir/TestCustomModelsFromROOT.cxx.o
[ 96%] Linking CXX executable TestCustomModelsFromROOT
```
This is due to the `emitFromONNX` being unconditionally run and producing header files that are input to the next steps.

### Reproducer

```
make
make
```

### ROOT version

master

### Installation method

local build

### Operating system

Linux

ferdymercury:
Maybe related: https://github.com/root-project/root/issues/18432

guitargeek:
Optimally, the header files should not be produced in the the build step but when running the tests. The generated header files are ONNX models converted to C++ by TMVA-SOFIE.
As it is now, TMVA-SOFIE development is quite annoying because if you are in an intermediate development state where not tested SOFIE usecases work, the compilation will fail at the first failing model and you don't have the opportunity to run all the tests after the build to see where you stand. This is something I discussed also yesterday with @lmoneta.

@sanjibansg, can something be done about this?

guitargeek:
[...]

PR #19370, opened 2025-07-14
[tmva][sofie] Don't compile emitted code for testing at build time

guitargeek:
Following a discussion with @lmoneta and @sanjibansg.

Don't compile emitted code for testing at build time, but use the convenient ROOTTEST macros to define the right sequence of emitting code, building, and testing.

Closes https://github.com/root-project/root/issues/18554.

To make this easier, move the `ROOTTEST` macros to the global `RootMacros.cmake` file, which should be fine now that the repositories are merged anyway. What do you think, @pcanal and @linev?


tmva/sofie/test/CMakeLists.txt
```diff
@@ -34,99 +32,71 @@ string(REPLACE ";" ";\n" EMIT_CAPTURES "${ALL_CAPTURES}")
...
-add_executable(emitFromONNX
-   EmitFromONNX_all.cxx
-)
-target_include_directories(emitFromONNX PRIVATE
-   ${CMAKE_SOURCE_DIR}/tmva/sofie/inc
-   ${SOFIE_PARSERS_DIR}/inc
-   ${CMAKE_SOURCE_DIR}/tmva/inc
-   ${CMAKE_CURRENT_BINARY_DIR}   # this is for the protobuf headerfile
-)
+ROOTTEST_GENERATE_EXECUTABLE(emitFromONNX EmitFromONNX_all.cxx LIBRARIES protobuf::libprotobuf ROOTTMVASofie ROOTTMVASofieParser)
```
vepadulano: IIUC this is missing a `FIXTURES_SETUP` which should then be picked in `SofieCompileModels_ONNX` via a corresponding `FIXTURES_REQUIRED` with the same name

tmva/sofie/test/CMakeLists.txt
```diff
@@ -34,99 +32,71 @@ string(REPLACE ";" ";\n" EMIT_CAPTURES "${ALL_CAPTURES}")
...
+      BLAS::BLAS
+      GTest::gtest
+      GTest::gtest_main
+    FIXTURES_REQUIRED
+      sofie-compile-models-onnx
+  )
+  target_include_directories(TestCustomModelsFromONNX PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
+  ROOTTEST_ADD_TEST(TestCustomModelsFromONNX
+                    EXEC ./TestCustomModelsFromONNX
+                    DEPENDS ${GENERATE_EXECUTABLE_TEST})
```
vepadulano: I would strongly suggest to avoid mixing `FIXTURES_*` with `DEPENDS`. In my experience it's better (and clearer) to just use `FIXTURES_*` all the time

tmva/sofie/test/CMakeLists.txt
```diff
@@ -34,99 +32,71 @@ string(REPLACE ";" ";\n" EMIT_CAPTURES "${ALL_CAPTURES}")
...
-add_executable(emitFromROOT
-   EmitFromRoot_all.cxx
-)
-target_include_directories(emitFromROOT PRIVATE
-   ${CMAKE_SOURCE_DIR}/tmva/sofie/inc
-   ${SOFIE_PARSERS_DIR}/inc
-   ${CMAKE_SOURCE_DIR}/tmva/inc
-   ${CMAKE_CURRENT_BINARY_DIR}
+ROOTTEST_GENERATE_EXECUTABLE(emitFromROOT EmitFromRoot_all.cxx
+    LIBRARIES protobuf::libprotobuf RIO ROOTTMVASofie ROOTTMVASofieParser
```
vepadulano, line 71: Similarly this needs a FIXTURES_SETUP with the corresponding FIXTURES_REQUIRED in `SofieCompileModels_ROOT`

tmva/sofie/test/CMakeLists.txt
```diff
@@ -148,34 +118,8 @@ if (ROOT_TORCH_FOUND)
...
-   ${CMAKE_SOURCE_DIR}/tmva/sofie/inc
-   ${CMAKE_SOURCE_DIR}/tmva/inc
-   ${CMAKE_SOURCE_DIR}/core/foundation/inc
-   ${CMAKE_BINARY_DIR}/ginclude   # this is for RConfigure.h
-)
-target_link_libraries(emitGraphIndependent ROOTTMVASofie)
-set_target_properties(emitGraphIndependent PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
-## silence protobuf warnings seen in version 3.0 and 3.6. Not needed from protobuf version 3.17
-target_compile_options(emitGraphIndependent PRIVATE -Wno-unused-parameter -Wno-array-bounds)
+ROOTTEST_GENERATE_EXECUTABLE(emitGNN GNN/EmitGNN.cxx LIBRARIES ROOTTMVASofie)
```
vepadulano: same comment about fixtures here

tmva/sofie/test/CMakeLists.txt
```diff
@@ -148,34 +118,8 @@ if (ROOT_TORCH_FOUND)
...
+ROOTTEST_ADD_TEST(emitGNN EXEC ./emitGNN DEPENDS ${GENERATE_EXECUTABLE_TEST})
+
+ROOTTEST_GENERATE_EXECUTABLE(emitGraphIndependent GNN/EmitGraphIndependent.cxx LIBRARIES ROOTTMVASofie)
```
vepadulano: same comment about fixtures here

tmva/sofie/test/CMakeLists.txt
```diff
@@ -34,99 +32,71 @@ string(REPLACE ";" ";\n" EMIT_CAPTURES "${ALL_CAPTURES}")
...
+      GTest::gtest
+      GTest::gtest_main
+    FIXTURES_REQUIRED
+      sofie-compile-models-root
   )
-  add_dependencies(TestCustomModelsFromROOT SofieCompileModels_ROOT)
+  target_include_directories(TestCustomModelsFromROOT PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
+  ROOTTEST_ADD_TEST(TestCustomModelsFromROOT
+                    EXEC ./TestCustomModelsFromROOT
+                    DEPENDS ${GENERATE_EXECUTABLE_TEST})
```
vepadulano: again here comment about using FIXTURES macros

vepadulano (changes requested):
Thanks, I believe this is the good approach! The setup with the FIXTURES macros is not polished yet, I left some suggestions

guitargeek:
Thanks for the review! You are right, fixtures are clearer than the magic `DEPENDS ${GENERATE_EXECUTABLE_TEST}`

tmva/sofie/test/CMakeLists.txt
```diff
@@ -148,34 +127,8 @@ if (ROOT_TORCH_FOUND)
...
-)
-target_link_libraries(emitGraphIndependent ROOTTMVASofie)
-set_target_properties(emitGraphIndependent PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
-## silence protobuf warnings seen in version 3.0 and 3.6. Not needed from protobuf version 3.17
-target_compile_options(emitGraphIndependent PRIVATE -Wno-unused-parameter -Wno-array-bounds)
+ROOT_EXECUTABLE(emitGNN GNN/EmitGNN.cxx LIBRARIES ROOTTMVASofie)
+ROOT_ADD_TEST(tmva-sofie-EmitGNN COMMAND emitGNN)
+
+ROOT_EXECUTABLE(EmitGraphIndependent GNN/EmitGraphIndependent.cxx LIBRARIES ROOTTMVASofie)
+ROOT_ADD_TEST(tmva-sofie-EmitGraphIndependent COMMAND EmitGraphIndependent)
```
vepadulano, line 134: Is it safe to not use the ROOTTEST_GENERATE_EXECUTABLE macro in this case and thus not use the FIXTURES_* ?

vepadulano (approved):
LGTM, one small remark

vepadulano (commented):
One general comment regarding the change in the build system. Note that even though roottest got merged into root we still have the build option `roottest` which could be `OFF`. In that case, the ROOTTEST macros being defined inside `RootMacros.cmake` might result in missing variables or anything wrong with the cmake. Probably this should be evaluated

tmva/sofie/test/CMakeLists.txt
guitargeek: Yes, it that case it's safe, because the executable source code doesn't depend on code generated by SOFIE.

The SOFIE-emitted code in only some of the test executables was the main motivation of moving the build of the executables to the test time with ROOTTEST_GENERATE_EXECUTABLE. When the SOFIE-generated code was compiled at ROOT build time, any invalid change in the structure of the emitted code caused ROOT to fail building. That happens a lot in intermediate states during development.

Now that the generated code is compiled at test time, one can build ROOT to completion and then check if the emitted code is valid by running the tests, which is way more convenient.

guitargeek:
Good question! I checked to build and run the tests with `roottest=OFF`, and it all works :+1:

Also I added one more commit to fix an unused variable warning.

cmake/modules/RootMacros.cmake
```diff
@@ -2072,3 +2072,1331 @@ function(generateManual name input output)
 
   install(FILES ${output} DESTINATION ${CMAKE_INSTALL_MANDIR}/man1)
 endfunction()
+
+#-------------------------------------------------------------------------------
+#
+#  Former RoottestMacros.cmake starts here
```
pcanal, line 2078: I would have found it much clearer to just move the file as is .... even if it was then 'included' here ....

guitargeek: For the commit history maybe, but my hope was that by bringing these files together, more code duplication will be apparent and inspire code improvements along the way in the future.

pcanal: I think that having the file side by side would have had the same effect ... and preserve commit history (which is too often the only source of documentation of those routines) ...

cmake/modules/RootMacros.cmake
```diff
@@ -2072,3 +2072,1331 @@ function(generateManual name input output)
...
+function(ROOTTEST_TARGETNAME_FROM_FILE resultvar filename)
+
+  get_filename_component(realfp ${filename} ABSOLUTE)
+  get_filename_component(filename_we ${filename} NAME_WE)
+
+  if(DEFINED ROOTTEST_DIR)
+    string(REPLACE "${ROOTTEST_DIR}" "" relativepath ${realfp})
+  else()
+    string(REPLACE "${CMAKE_SOURCE_DIR}" "" relativepath ${realfp})
+  endif()
```
pcanal, line 2170: This is odd.  Why is `ROOTTEST_DIR` not defined?   Why are the two lines semantically different?  Shouldn't the second one be:
```
string(REPLACE "${CMAKE_SOURCE_DIR}/roottest" "" relativepath ${realfp})
```
?

guitargeek: `ROOTTEST_DIR` is only defined inside `roottest/`:
https://github.com/root-project/root/blob/master/roottest/CMakeLists.txt#L270

So if you use the macro in other subdirectories, it's not defined.

That means that the absolute path in `realfp` is not reduced to the relativepath by removing the ROOTTEST_DIR. Then, the `relativepath` will be used to define the target and test names, and if you have absolute paths in your test names it's quite weird.

That's why I introduced the fallback to always strip the `$CMAKE_SOURCE_DIR` from the path. `"${CMAKE_SOURCE_DIR}/roottest"` would not work if your test is not in the `roottest` subdirectory.

pcanal: Fair enough.  It still feels odd that the 2 are treated differently and if
```
string(REPLACE "${CMAKE_SOURCE_DIR}" "" relativepath ${realfp})
```
for all cases does not work then there is likely more corners that might need polishing to make the `ROOTTEST_` macro generically usable ... this of course for another time but in the meantime, I would recommend adding a bit of history/documentation next the code.  Thanks.

guitargeek: You are right, I can improve that with some later updates. I think it's not necessary to have `ROOTTEST_DIR` branch at all, but then the tests will be named like `roottest-roottest-*`, because every ROOTTEST also prefixed the name with `roottest-`. That's weird then. But now that there will always be the roottest subdirectory, we can just use `CMAKE_SOURCE_DIR` all the was and don't add the prefix.

pcanal: > But now that there will always be the roottest subdirectory, we can just use CMAKE_SOURCE_DIR all the was and don't add the prefix.

That sounds about right. (i.e. in the end the testname should not change :) ).

guitargeek: Followed up in:
  * https://github.com/root-project/root/pull/19481

Issue #19267, opened 2025-07-03
[cppyy] New crash when iterating over polymorphic std::vector


Nowakus:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Starting around 24/06/2025 a new crash appeared in 10 years old code where we iterate in python over an std::vector of polymorphic objects created on the C++ side. The inheritance structure is 4 levels deep and contains virtual inheritance that is apparently the reason cppyy trips. The iteration is the faulty part - single object uses work fine.

### Reproducer

For now just quick copy/paste to give the overall feel of the situation:
Here's a brief summary of the inheritance hierarchy:

```
class Analysis::JetFitterTagInfo : public virtual Analysis::IJetFitterTagInfo
class Analysis::IJetFitterTagInfo : public virtual Analysis::BaseTagInfo
class Analysis::BaseTagInfo : public JetTagInfoBase
class JetTagInfoBase : public Analysis::ITagInfo

 class TagInfoTest {   public:
      typedef std::vector<const JetTagInfoBase*>       tagstore_t;
      const tagstore_t makeInfoVector(const char* tag) const {
         tagstore_t store;
         store.push_back( new Analysis::BaseTagInfo(tag) );
         store.push_back( new Analysis::JetFitterTagInfo(tag) );
         return store;
      }
   }; 

>>> import ROOT
>>> for i in ROOT.TagInfoTest().makeInfoVector('bla'):
...   print( i.infoType() )
... 
bla
free(): invalid pointer
 *** Break *** abort
```

### ROOT version

master

### Installation method

ATLAS

### Operating system

Linux

hahnjo:
Hi @guitargeek @Vipul-Cariappa, I was just chatting with Marcin. Based on the date (June 24, +- 2 days), the most likely candidate is https://github.com/root-project/root/commit/e93464a9e95cfcafb2d51b34fe6eb640e3270106. For the moment, the problem is worked around on the Athena side, but would be good to identify this regression.

Vipul-Cariappa:
Hi @Nowakus, could you provide a complete reproducer. Or point me to the implementation of those classes.
I suspect that it is not to do with `std::vector` iteration. It might have to do with some member field being misinterpreted in cppyy.
But I see `free(): invalid pointer` in your report. Not sure how cppyy would cause that.

wlav:
[...]

PR #19374, opened 2025-07-15
[cppyy] Fix type conversion for polymorphic classes

Vipul-Cariappa:
## Changes or fixes:

fixes: https://github.com/root-project/root/issues/19267


dpiparo:
Thanks for this PR. Two minor observations:
1) Why 2 commits? It's not clear from the messages, especially of the second commit
2) Can a test be added to avoid regressions in the future?

dpiparo:
It would also be nice to backport to 6.36...

Vipul-Cariappa:
Yes. I will add a test case and squash the commits.

bindings/pyroot/cppyy/cppyy/test/test_datatypes.py
```diff
@@ -2410,6 +2410,44 @@ def test51_polymorphic_types_in_maps(self):
...
+        public:
+            int jf = 3;
+        };
+
+        std::vector<const JetTag*> make() {
+            std::vector<const JetTag*> res;
+            res.push_back(new BaseTag());
+            res.push_back(new JetFit());
+            return res;
+        }
```
vepadulano: I understand this is as close as possible to the original snippet provided at the linked issue, but can we find a way to avoid explicit memory leaks in the test? I'm guessing creating the vector of objects as a data member of a class and then exposing the vector of pointers via `make()` would still be a good reproducer

bindings/pyroot/cppyy/cppyy/test/test_datatypes.py
```diff
@@ -2410,6 +2410,44 @@ def test51_polymorphic_types_in_maps(self):
...
+        """)
+
+        for i in gbl.make():
+            assert i.name == "NAME"
```
vepadulano, line 2447: What is this really testing actually? The only class in the hierarchy with the `name` data member is the base class? Did I not understand correctly that we want to check also polymorphic behaviour? Shoulnd't all classes in the hierarchy have a `name` member method that overrides the behaviour of the base one and the test should check that each different element of the vector returns the name of the corresponding class?

bindings/pyroot/cppyy/cppyy/test/test_datatypes.py
Vipul-Cariappa: fixed.

bindings/pyroot/cppyy/cppyy/test/test_datatypes.py
Vipul-Cariappa: If the derived class modifies `name`, it does not crash anymore. Therefore, `name` cannot be changed.

vepadulano: After discussion offline I understand the reproducer better, thanks!

vepadulano (approved):
Thank you for the fix!

guitargeek:
> It would also be nice to backport to 6.36...

Backport opened: https://github.com/root-project/root/pull/19409

Thanks Vipul for the fix!

bindings/pyroot/cppyy/cppyy/test/test_datatypes.py
```diff
@@ -2410,6 +2410,41 @@ def test51_polymorphic_types_in_maps(self):
...
+            std::vector<const JetTag*> res;
+            res.push_back(&bt);
+            res.push_back(&jft);
+            res.push_back(&jf);
+            return res;
+        }
+        """)
+
+        for i in gbl.make():
+            assert i.name == "NAME"
```
pcanal, line 2447: How does the test fails without the fix?

Vipul-Cariappa: segfault or assertion error. It is not consistent.

PR #19242, opened 2025-07-01
Update ROOT and Cling to LLVM20

devajithvs:
# This Pull request:
This is a work in progress. Opening a PR to get wider coverage and a better idea on failing tests on different platforms.

TODO: 
- [ ] Organize commits and make it more verbose/remove unneeded commits 
- [ ] Fix the remaining failing tests
- [ ] Update git tags for LLVM

## Checklist:

- [ ] tested changes locally
- [ ] updated the docs (if necessary)

This PR fixes #

PR #19375, opened 2025-07-15
[cling] Use clang-repl-style code completion

devajithvs:
# This Pull request:
This patch eliminates the need to create a nested interpreter for code
completion.

In LLVM 19, some Sema internals became private:
See https://github.com/llvm/llvm-project/commit/874f511

Without this, we would require an additional patch on top of Clang to
access `Sema::CodeCompleter` after upgrading to LLVM 20.

This will fix one of the failing test during [upgrade](https://github.com/root-project/root/pull/19242) in CppInterOp (autocomplete)

## Checklist:

- [ ] tested changes locally

This PR fixes #

### DIFF
--- a/interpreter/cling/include/cling/Interpreter/ClingCodeCompleteConsumer.h
+++ b/interpreter/cling/include/cling/Interpreter/ClingCodeCompleteConsumer.h
@@ -14,6 +14,10 @@
 
 using namespace clang;
 
+namespace clang {
+  class CompilerInstance;
+}
+
 namespace cling {
   /// \brief Create a new printing code-completion consumer that prints its
   /// results to the given raw output stream.
@@ -49,6 +53,23 @@ namespace cling {
     void getCompletions(std::vector<std::string>& completions) {
       completions = m_Completions;
     }
+  };
+
+  struct ClingCodeCompleter {
+    ClingCodeCompleter() = default;
+    std::string Prefix;
+
+    /// \param[in] InterpCI The compiler instance that is used to trigger code
+    ///                     completion.
+    /// \param[in] Content The string where code completion is triggered.
+    /// \param[in] Line The line number of the code completion point.
+    /// \param[in] Col The column number of the code completion point.
+    /// \param[in] ParentCI The running interpreter compiler instance that
+    ///                     provides ASTContexts.
+    /// \param[out] CCResults The completion results.
+    void codeComplete(CompilerInstance* InterpCI, llvm::StringRef Content,
+                      unsigned Line, unsigned Col, CompilerInstance* ParentCI,
+                      std::vector<std::string>& CCResults);
   };
 }
 
--- a/interpreter/cling/lib/Interpreter/CIFactory.cpp
+++ b/interpreter/cling/lib/Interpreter/CIFactory.cpp
@@ -42,6 +42,7 @@
 #include "llvm/Config/llvm-config.h"
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/Option/ArgList.h"
+#include "llvm/Support/Error.h"
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/MemoryBuffer.h"
 #include "llvm/Support/Path.h"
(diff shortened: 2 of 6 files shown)


interpreter/cling/lib/Interpreter/CIFactory.cpp
```diff
@@ -1252,6 +1253,55 @@ namespace {
     }
   }
 
+  // Goal is to make this as close to the function in
+  // clang/lib/Interpreter/Interpreter.cpp taking only clang arguments
+  static llvm::Expected<std::unique_ptr<CompilerInstance>>
```
hahnjo, line 1258: I think this function never returns an error. Instead it prints directly to `cling::errs()` and returns a "successful" `nullptr`...

interpreter/cling/lib/Interpreter/CIFactory.cpp
```diff
@@ -1252,13 +1253,61 @@ namespace {
...
+  CreateCI(const std::vector<const char*>& ClangArgv, std::string ExeName,
```
hahnjo:
```suggestion
  CreateCI(const std::vector<const char*>& ClangArgv, const std::string& ExeName,
```

interpreter/cling/lib/Interpreter/CIFactory.cpp
```diff
@@ -1410,53 +1460,27 @@ namespace {
...
+    std::unique_ptr<clang::driver::Compilation> Compilation;
 
-    const llvm::opt::ArgStringList* CC1Args = GetCC1Arguments(Compilation.get());
-    if (!CC1Args) {
-      cling::errs() << "Could not get cc1 arguments.\n";
+    auto CIOrErr = CreateCI(argvCompile, ExeName, Compilation);
+    if (!CIOrErr) {
+      cling::errs() << "Could not create CI: " << CIOrErr.takeError() << "\n";
       return nullptr;
     }
```
hahnjo: see comment above, I think the function currently returns a "successful" `nullptr` and not actually an error...

interpreter/cling/lib/Interpreter/CIFactory.cpp
```diff
@@ -1635,7 +1639,14 @@ namespace {
...
       Buffer = llvm::MemoryBuffer::getMemBuffer("/*CLING DEFAULT MEMBUF*/;\n");
-    MainFileCC.setBuffer(std::move(Buffer));
+    if (AutoComplete) {
+      // Adapted from upstream clang/lib/Interpreter/Interpreter.cpp
+      // FIXME: Merge with CompilerInstance::ExecuteAction.
+      llvm::MemoryBuffer* MB = Buffer.release();
+      CI->getPreprocessorOpts().addRemappedFile(Filename, MB);
+    } else {
+      MainFileCC.setBuffer(std::move(Buffer));
+    }
```
hahnjo: Can the code path using `addRemappedFile` potentially be used for both code paths?

interpreter/cling/include/cling/Interpreter/ClingCodeCompleteConsumer.h
```diff
@@ -50,6 +54,28 @@ namespace cling {
...
+    /// \param Content [in] The string where code completion is triggered.
+
+    /// \param Line [in] The line number of the code completion point.
+
+    /// \param Col [in] The column number of the code completion point.
+
+    /// \param ParentCI [in] The running interpreter compiler instance that
+    /// provides ASTContexts.
+
+    /// \param CCResults [out] The completion results.
```
hahnjo: comment formatting looks off here?

interpreter/cling/lib/Interpreter/Interpreter.cpp
```diff
@@ -1026,52 +1028,25 @@ namespace cling {
...
-    PushTransactionRAII RAII(this);
+    // arguments for constructing CI
+    auto declCollector = std::make_unique<cling::DeclCollector>();
+    const ModuleFileExtensions& moduleExtensions = {};
 
-    // Triger the code completion.
-    childInterpreter.CodeCompleteInternal(line, cursor);
+    auto InterpCI = std::unique_ptr<clang::CompilerInstance>(
+        CIFactory::createCI("\n", getOptions(), llvmDir.c_str(),
+                            std::move(declCollector), moduleExtensions,
```
hahnjo, line 1044: Not sure if it would be possible to make the `ASTConsumer` optional if we don't really need it... but ok for the current patch

hahnjo (commented):
Thanks, looks mostly good to me! Some minor comments on the calling convention / error return semantics of `CreateCI`

hahnjo (approved):
LGTM, thanks!

PR #19381, opened 2025-07-15
[cmake] disable Gui if minimal=on

silverweed:
# This Pull request:
Adds a `gui` build option that allows disabling the Gui library. It is meant to be used in conjunction with `minimal=on` as no effort is made to make sure that the option is coherent with the set of other options, meaning turning the gui off by itself won't automatically disable all its dependent modules. However, when minimal is on, everything works fine.

### Rationale
When someone builds with `minimal=on` they don't care about graphics (otherwise they'd use `gminimal`) and building the graphics libraries is a waste of time. AFAIK, the gui already doesn't work when building with `minimal=on`, so we can spare time by not building it.

Ideally we should be able to also disable graf2d and graf3d, however too many core packages depend on them to be as easily doable (we'd have to disable half of the math/tree/hist modules and rethink `modules.idx` and `hsimple.root`), so that'll remain a wish for the future.

## Checklist:

- [x] tested changes locally


vepadulano:
Just a fly-by comment

> the gui already doesn't work when building with minimal=on

I can confirm this

CMakeLists.txt
```diff
@@ -399,7 +399,9 @@ add_subdirectory (io)
 add_subdirectory (net)
 add_subdirectory (graf2d)
 add_subdirectory (graf3d)
-add_subdirectory (gui)
+if(gui)
```
hageboeck: I would like to suggest moving to
```suggestion
if(NOT minimal)
```
Because you either build minimal, or gminimal, or "standard" ROOT. A fourth option doesn't seem to work.

bellenot (approved):
Thanks Jack!

silverweed:
@bellenot @dpiparo Following Stephan's suggestion, instead of using a new build option I'm now directly checking for `NOT minimal`, which has 2 benefits:
- avoids introducing a new build option
- avoids unsupported cases where one has gui=off but without minimal=on

tree/ntuplebrowse/test/CMakeLists.txt
```diff
@@ -6,4 +6,8 @@
 
 # @author Jakob Blomer CERN
 
+if(NOT minimal) # gui is not built in minimal
```
bellenot:
```suggestion
if(minimal) # gui is not built in minimal
```

bellenot:
> @bellenot @dpiparo Following Stephan's suggestion, instead of using a new build option I'm now directly checking for `NOT minimal`, which has 2 benefits:
> 
> * avoids introducing a new build option
> * avoids unsupported cases where one has gui=off but without minimal=on

I fully agree!

test/CMakeLists.txt
```diff
@@ -68,7 +68,9 @@ ROOT_EXECUTABLE(eventexe MainEvent.cxx LIBRARIES Event RIO Tree TreePlayer Hist
 ROOT_ADD_TEST(test-event COMMAND eventexe)
 
 #---guitest------------------------------------------------------------------------------------
-ROOT_EXECUTABLE(guitest guitest.cxx LIBRARIES RIO Gui Gpad Hist)
+if(NOT minimal) # gui is not built in minimal
```
hageboeck: I would bind this to the target, so no further changes are required if we would ever touch the build options again. Also, no further comments are needed.
```suggestion
if(TARGET Gui)
```

tree/CMakeLists.txt
```diff
@@ -6,7 +6,9 @@
 
 add_subdirectory(tree)
 add_subdirectory(treeplayer)
-add_subdirectory(treeviewer)
+if(NOT minimal) # gui is not built in minimal
```
hageboeck:
```suggestion
if(TARGET Gui)
```

tree/ntuplebrowse/test/CMakeLists.txt
```diff
@@ -6,4 +6,8 @@
 
 # @author Jakob Blomer CERN
 
+if(minimal) # gui is not built in minimal
```
hageboeck:
```suggestion
if(NOT TARGET Gui)
```

dpiparo:
> > @bellenot @dpiparo Following Stephan's suggestion, instead of using a new build option I'm now directly checking for `NOT minimal`, which has 2 benefits:
> > 
> > * avoids introducing a new build option
> > * avoids unsupported cases where one has gui=off but without minimal=on
> 
> I fully agree!

Fine for me too.

silverweed:
I need to understand why the version using `TARGET Gui` as proposed by @hageboeck is not working (I haven't had time so far)

CMakeLists.txt
```diff
@@ -399,7 +399,9 @@ add_subdirectory (io)
 add_subdirectory (net)
 add_subdirectory (graf2d)
 add_subdirectory (graf3d)
-add_subdirectory (gui)
+if(TARGET Gui)
```
ferdymercury:
```suggestion
if(NOT minimal)
```
The target Gui is only defined in gui/CMakeLists.txt
`ROOT_STANDARD_LIBRARY_PACKAGE(Gui`

so the target is never defined because the subdirectory is not added.

(The other checks below: it's fine to use if (TARGET Gui) since they happen later.)

test/CMakeLists.txt
```diff
@@ -160,7 +162,7 @@ ROOT_ADD_TEST(test-stresslinear-interpreted COMMAND ${ROOT_root_CMD} -b -q -l ${
 
 #--stressGraphics------------------------------------------------------------------------------------
 if(ROOT_opengl_FOUND)
```
ferdymercury:
```suggestion
if(ROOT_opengl_FOUND AND TARGET TreeViewer)
```
stressGraphics depends on TreeViewer, but it is not built if Gui is not built

ferdymercury: An alternative would be to disable `parallelcoord()` test within stressGraphics, depending on if has_include TParallelCoord

ferdymercury (changes requested):
Two changes are needed:

test/CMakeLists.txt:164 `AND TARGET TreeViewer`
main/CMakeLists.txt:115 `if TARGET Gui`

ferdymercury:
Ah, one more thing, apart from vetoing the mentioned tutorials, please also change the title and first post of this PR, to adapt to the current behavior. (disable GUI when minimal is ON).
Thanks!

ferdymercury:
> Ah, one more thing, apart from vetoing the mentioned tutorials, please also change the title and first post of this PR, to adapt to the current behavior. (disable GUI when minimal is ON). Thanks!

An (unrelated) additional point is: there are things in line 76 of config/root-config.in that maybe should not be added to the rootlibs variable if minimal=ON.
Related: https://its.cern.ch/jira/browse/ROOT-8061

tree/CMakeLists.txt
```diff
@@ -6,12 +6,14 @@
 
 add_subdirectory(tree)
 add_subdirectory(treeplayer)
-add_subdirectory(treeviewer)
+if(TARGET Gui)
+  add_subdirectory(treeviewer)
+  add_subdirectory(ntuplebrowse)
```
ferdymercury, line 11: Is this line maybe redundant with the check in tree/ntuplebrowse/test/CMakeLists.txt ?

ferdymercury: Also, I think that in the main CMakeLists.txt
`add_subdirectory (tree)`
happens before
`add_subdirectory (gui)`
so we cannot yet use if (TARGET Gui)
We need to use if (NOT minimal) or change the order in the main CMakeLists so that Gui is defined or not correctly before checking it in tree?

tree/ntuplebrowse/test/CMakeLists.txt
```diff
@@ -6,4 +6,8 @@
 
 # @author Jakob Blomer CERN
 
+if(NOT TARGET Gui) # gui is not built in minimal
+  return()
+endif()
+
```
ferdymercury:
```suggestion
```

(Redundant)

hageboeck (approved):
👍 if you squash.

PR #19383, opened 2025-07-15
[ntuple] Add Escape HTML characters in PrintFieldTreeAsDot

magnustymoteus:
# This Pull request:

## Changes or fixes:
This fixes graphviz getting confused and not rendering a node in cases where certain HTML characters like `>` or `<` are not escaped in its label (e.g field type is `vector<int>`). A small lambda function escapes such characters by for example changing a `>` to `&gt;`.

## Checklist:

- [X] tested changes locally


enirolf (approved):
Nice, LGTM!

PR #19385, opened 2025-07-16
Remove two unused dictionary rules

hahnjo:
The third warning that I see locally is addressed in https://github.com/root-project/root/pull/19202.

### DIFF
--- a/bindings/pyroot/cppyy/cppyy/test/stltypes.xml
+++ b/bindings/pyroot/cppyy/cppyy/test/stltypes.xml
@@ -33,5 +33,4 @@
   <function pattern="ErrorNamespace::*" />
 
   <function name="GetMyErrorCount" />
-  <variable name="N" />
 </lcgdict>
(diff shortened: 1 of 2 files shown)


guitargeek (approved):
Thank you very much!

hahnjo:
@guitargeek does this need a patch file as well? plus maybe 1b41183c4dd3937fd0a709eff8cb3faf03bedcb8?

guitargeek:
Let's see. I'll first open a few PRs for Wim with these commits. If he doesn't want to merge then in the next days for one or the other reason, then I'll create the patch files (actually I need to fix up the `bindings/pyroot/cppyy/sync-upstream` anyway today/in the next days - it got very much out of date).

guitargeek:
PR opened:
  * https://github.com/wlav/cppyy/pull/309

It doesn't include your removal of the unused class rule, because this was only introduced in ROOT anyway by b7892373009df5 from #19105 (instead of merging that PR, we should have just defined the `N` variable on the Python side as I suggested later).

Issue #7541, opened 2021-03-16
Cannot catch exceptions on Apple M1


hahnjo:
### Describe the bug
It's not possible to catch exceptions in the interactive `root` prompt on Apple Silicon, nor does `TRint` take care of uncaught exceptions.

### Expected behavior
The user should be able to catch exceptions, or at least the fallback handler should prevent process termination.

### To Reproduce
The most simple examples are
```
root [0] try { throw 1; } catch (...) { }
libc++abi.dylib: terminating with uncaught exception of type int
```
and
```
root [0] throw 1;
libc++abi.dylib: terminating with uncaught exception of type int
```
(which should be handled in `TRint::HandleTermInput()`).

### Setup
ROOT 6.25/01 on `macphsft25`

### Additional context
Plenty:
 * #7100 fixed catching exceptions from constructors on macOS (and has tests, including unwinding through multiple JIT recursions).
 * [ROOT-10703](https://sft.its.cern.ch/jira/browse/ROOT-10703) and [ROOT-10962](https://sft.its.cern.ch/jira/browse/ROOT-10962) for the original reports.

pcanal:
This reminds me of https://sft.its.cern.ch/jira/browse/ROOT-8544 and https://sft.its.cern.ch/jira/browse/ROOT-8523 which is the end where fixed  by a7b0b3e647409c7510b38198b08ff94fd079f857.

Most likely the way the interpreter sets up the stack frames does not match the expectation of the exception handler (usually implemented in (g)libc).

It is very plausible to be a problem similar to the one leading to the "can not reallocate code" errors.  So I see two plausible path forward: (a) fix the reallocate code error and hopefully it also fix this (b) installed a debug version of (g)libc and trace/debug the exception handlers handling of this case ...

hahnjo:
Ugh, this one might become tricky, even LLVM upstream is unable to handle exceptions during JIT: I tested the most basic
```c++
int main() {
  try {
    throw 1;
  } catch (...) { }
}
```
compiled with `./bin/clang++ -S -emit-llvm throw.cc` and interpreted using `./bin/lli throw.ll`, resulting in
```
libc++abi.dylib: terminating with uncaught exception of type int
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace.
Stack dump:
0.      Program arguments: ./bin/lli throw.ll
[...]

PR #19386, opened 2025-07-16
[io] Register TKey after successful Streamer()

hahnjo:
This avoids corrupting the in-memory `TFile` structure with throwing `Streamer()` functions.

### DIFF
--- a/io/io/src/TKey.cxx
+++ b/io/io/src/TKey.cxx
@@ -327,7 +327,6 @@ TKey::TKey(const void *obj, const TClass *cl, const char *name, Int_t bufsize, T
 
    fBufferRef = new TBufferFile(TBuffer::kWrite, bufsize);
    fBufferRef->SetParent(GetFile());
-   fCycle     = fMotherDir->AppendKey(this);
 
    Streamer(*fBufferRef);         //write key itself
    fKeylen    = fBufferRef->Length();
@@ -338,6 +337,10 @@ TKey::TKey(const void *obj, const TClass *cl, const char *name, Int_t bufsize, T
    clActual->Streamer((void*)actualStart, *fBufferRef); //write object
    lbuf       = fBufferRef->Length();
    fObjlen    = lbuf - fKeylen;
+
+   // Append to mother directory only after the call to Streamer() was
+   // successful (and didn't throw).
+   fCycle = fMotherDir->AppendKey(this);
 
    Int_t cxlevel = GetFile() ? GetFile()->GetCompressionLevel() : 0;
    ROOT::RCompressionSetting::EAlgorithm::EValues cxAlgorithm = static_cast<ROOT::RCompressionSetting::EAlgorithm::EValues>(GetFile() ? GetFile()->GetCompressionAlgorithm() : 0);
--- a/roottest/root/io/customStreamer/CMakeLists.txt
+++ b/roottest/root/io/customStreamer/CMakeLists.txt
@@ -2,3 +2,11 @@ ROOTTEST_ADD_TEST(run
                   MACRO run.C
                   OUTREF customStreamerTest.ref)
 
+ROOTTEST_GENERATE_DICTIONARY(StreamerThrowClass StreamerThrowClass.h
+                             LINKDEF StreamerThrowClass_LinkDef.h
+                             FIXTURES_SETUP root-io-customStreamer-StreamerThrow)
+# ACLiC compiled to work around problems with exceptions in Apple Silicon.
+ROOTTEST_ADD_TEST(StreamerThrow
+                  MACRO StreamerThrow.C+
+                  OUTREF StreamerThrow.ref
+                  FIXTURES_REQUIRED root-io-customStreamer-StreamerThrow)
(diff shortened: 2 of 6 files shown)


hahnjo:
Ok, I give up for the moment on the roottest - somebody has to explain me how to properly register the dependency...

roottest/root/io/customStreamer/CMakeLists.txt
```diff
@@ -2,3 +2,8 @@ ROOTTEST_ADD_TEST(run
                   MACRO run.C
                   OUTREF customStreamerTest.ref)
 
+ROOTTEST_GENERATE_DICTIONARY(StreamerThrowClass StreamerThrowClass.h LINKDEF StreamerThrowClass_LinkDef.h)
```
pcanal:
```suggestion
ROOTTEST_GENERATE_DICTIONARY(StreamerThrowClass StreamerThrowClass.h LINKDEF StreamerThrowClass_LinkDef.h FIXTURES_SETUP root-io-customStreamer-fixtures)
```

roottest/root/io/customStreamer/CMakeLists.txt
```diff
@@ -2,3 +2,8 @@ ROOTTEST_ADD_TEST(run
...
+ROOTTEST_ADD_TEST(StreamerThrow
+                  MACRO StreamerThrow.C
+                  OUTREF StreamerThrow.ref
+                  DEPENDS StreamerThrowClass-build)
```
pcanal:
```suggestion
                  FIXTURES_REQUIRED  root-io-customStreamer-fixtures)
```

io/io/src/TKey.cxx
```diff
@@ -339,6 +338,10 @@ TKey::TKey(const void *obj, const TClass *cl, const char *name, Int_t bufsize, T
    lbuf       = fBufferRef->Length();
    fObjlen    = lbuf - fKeylen;
 
+   // Append to mother directory only after the call to Streamer() was
+   // successful (and didn't throw).
+   fCycle = fMotherDir->AppendKey(this);
```
pcanal, line 343: This change looks good.   As a side note, `fCycle` is used in the 'write key itself' step on old line 332 (and thus switch from writing correct value to writing default value), however this first streaming is 'just' to reserve the space and it is written again (which updated values for fCycle and the sizes) right before returning out of this function.   Additional side note: I did not review thoroughly whenever the other functions between here and the user class Streamer are exception safe but a cursory look says that they are likely to be.

pcanal (approved):
LGTM.  Thanks (of course pending resolution of the test issues :) ).

hahnjo:
@pcanal thanks for the suggestions. I'm still slightly confused because I looked at other test directories (`roottest/cling/dict/ROOT-8739/` and `roottest/cling/dict/enum/` were the first to show up) and they didn't use the `FIXTURES` method. So I guess we have race conditions and missing dependencies in other places?

io/io/src/TKey.cxx
hahnjo: Yes, I also didn't check in detail. I wouldn't be surprised if we leak objects in case of exceptions, but at least it doesn't crash anymore and we have a means to propagate the exception to the user.

pcanal:
Yes, using FIXTURES for dependency is 'new'-ish has indeed not been applied systematically to all existing tests.  See Sergey's last few PR/commits for good examples.  (That said, when running all the test the DEPENDS should have been sufficient to avoid any race).

hahnjo:
Unfortunately didn't work. And the reason is obvious, if I had carefully checked the log before: I'm trying to catch an exception in interpreted code, which is known to not work on Apple Silicon (see https://github.com/root-project/root/issues/7541). I will rewrite as a compiled test tomorrow...

pcanal:
> I will rewrite as a compiled test tomorrow...

This might be as 'simple' as adding #include and a trailing '+' (to use ACLiC) :)

roottest/root/io/customStreamer/CMakeLists.txt
```diff
@@ -2,3 +2,10 @@ ROOTTEST_ADD_TEST(run
                   MACRO run.C
                   OUTREF customStreamerTest.ref)
 
+ROOTTEST_GENERATE_DICTIONARY(StreamerThrowClass StreamerThrowClass.h
+                             LINKDEF StreamerThrowClass_LinkDef.h
+                             FIXTURES_SETUP root-io-customStreamer-StreamerThrow)
+ROOTTEST_ADD_TEST(StreamerThrow
+                  MACRO StreamerThrow.C
```
pcanal:
```suggestion
                  MACRO StreamerThrow.C+
```

pcanal: in addition you can use a variation of
```
ROOTEXE_OPTS -e ".L StreamerThrowClass"
```
to explicitly load the dictionary library.
Side note, if I recall correct that syntax has challenges to use `;`, so for 'quiet' C++ statement we used `-e "(void)somefunction()"`

Furthermore, the ACLiC compilation can be done separately by adding something like:
```
ROOTTEST_COMPILE_MACRO(StreamerThrow.C
                        BUILDLIB StreamerThrowClass
                        FIXTURES_REQUIRED root-io-customStreamer-StreamerThrowLib
                        FIXTURES_SETUP  root-io-customStreamer-StreamerThrowACLiC)
```

PR #19387, opened 2025-07-16
[cmake] Keep `thisroot` scripts with -Dminimal=ON

bellenot:


guitargeek (approved):
Thanks! Indeed, since these scripts have nothing to do with the ROOT build but the deployment, it makes sense that the `minimal` build options doesn't affect it.

pcanal (approved):
Thanks.

PR #19391, opened 2025-07-17
Test Python wheel build github workflow

vepadulano:


bindings/pyroot/cppyy/CPyCppyy/CMakeLists.txt
```diff
@@ -54,6 +54,18 @@ set(libname cppyy)
...
+# site-packages/
+# -- cppyy/
+# -- cppyy_backend/
+# -- libcppyy.so
+# -- libcppyy_backend.so
+# -- libROOTPythonizations.so
+# -- ROOT/
+# With most of the ROOT libraries being stored in `ROOT/lib`. The RPATH must then be updated for the CPython extension
+# shared libraries.
+set_target_properties(${libname} PROPERTIES  INSTALL_RPATH "${CMAKE_INSTALL_RPATH}:$ORIGIN/ROOT/lib")
```
guitargeek, line 68: I *think* all of these hardcoded rpaths should now not be necessary if `CMAKE_INSTALL_PYTHONDIR` and `CMAKE_INSTALL_LIBDIR` are set correctly, at least relative to each other.

There are still some quirks with unneeded artifacts when these directories are not identical, which I'm trying to figure out:
https://github.com/root-project/root/pull/19044

But that should not be a problem for the functionality.

bindings/pyroot/pythonizations/CMakeLists.txt
```diff
@@ -242,7 +254,8 @@ if(NOT MSVC)
   else()
     set_target_properties(${libname} PROPERTIES
         BUILD_RPATH "$ORIGIN/${pymoduledir_to_libdir_build}"
-        INSTALL_RPATH "$ORIGIN/${pymoduledir_to_libdir_install}"
+        # ATTENTION: NEXT PART IS CRUCIAL SEE COMMENT ABOVE
+        INSTALL_RPATH "$ORIGIN/${pymoduledir_to_libdir_install}:${CMAKE_INSTALL_RPATH}:$ORIGIN/lib"
```
guitargeek, line 258: Here you actually see the new mechanism that I introduced so make arbitrary `CMAKE_INSTALL_PYTHONDIR` and `CMAKE_INSTALL_LIBDIR` work. Could you try `-DCMAKE_INSTALL_PYTHONDIR=site-packages/` and `-DCMAKE_INSTALL_LIBDIR=site-packages/ROOT/lib`?

`CMAKE_INSTALL_BINDIR` also needs to be set to `site-packages/ROOT/bin` then, so that the relative rpaths between `lib` and `bin` are not broken.

To summarize: I think if you avoid having the move around `site-packages/`, `site-packages/ROOT/bin` and `site-packages/ROOT/lib` *relative* to each other manually after the install but instead use the right CMake flags, it should work out of the box.

setup.py
```diff
@@ -0,0 +1,128 @@
...
+class ROOTBuild(_build):
+    def run(self):
+        _build.run(self)
+
+        # Configure ROOT build
+        base_opts = shlex.split("cmake -GNinja -Dccache=ON")
+        mode_opts = shlex.split(
+            "-Dbuiltin_nlohmannjson=ON -Dbuiltin_tbb=ON -Dbuiltin_xrootd=ON "  # builtins
+            "-Dbuiltin_lz4=ON -Dbuiltin_lzma=ON -Dbuiltin_zstd=ON -Dbuiltin_xxhash=ON"  # builtins
+            "-Druntime_cxxmodules=ON -Drpath=ON -Dfail-on-missing=ON "  # Generic build configuration
```
guitargeek, line 49:
```suggestion
            "-Druntime_cxxmodules=ON -Drpath=ON -Dfail-on-missing=ON -Dthisroot_scripts=OFF"  # Generic build configuration
```
We don't need the `thisroot` scripts (in fact they would be broken, because they assume `CMAKE_INSTALL_PYTHONDIR` equals `CMAKE_INSTALL_LIBDIR`), and they can be excluded from the build now:

  * https://github.com/root-project/root/pull/19211

vepadulano:
`pull_request` events do not trigger the workflow unless the workflow file is already in master. Closing the PR to avoid too much noise while the workflow is being worked upon

PR #19394, opened 2025-07-17
[cling] Unload all redecls of ClassTemplateDecl

devajithvs:
When unloading a ClassTemplateDecl, no redeclarations were handled.

Additional redecls may still be part of the chain leading to duplicates
and broken redecl chains that cause assertion failures like:
"Passed first decl twice, invalid redecl chain!"

Fixes assertion failures post llvm commit:
https://github.com/llvm/llvm-project/commit/17d8ed7
for the test `roottest-root-meta-cmsUnload-make`

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes # 
The failing test `roottest-root-meta-cmsUnload-make`  in https://github.com/root-project/root/pull/19242


devajithvs:
I've been looking into adding a test case, but it's a bit difficult to test this within the ROOT/Cling testing scope. We could add a simplified test similar to `roottest-root-meta-cmsUnload-make`, but will need LLVM commit https://github.com/llvm/llvm-project/commit/17d8ed7 to trigger an assertion.

vgvassilev:
Ok, any chance in remembering to add the test when that lands. I presume the test case is unloading the test case introduced by the commit you mentioned.

devajithvs:
> Ok, any chance in remembering to add the test when that lands. 

The test with full commit for reference: https://github.com/root-project/root/commit/08e304bca431184f773d41fe677081e34f39bd6b

I was using a simplified version of this test for debugging. But the above test should cover the case when unloading after the upgrade.

`case1.script`
```
#include "inc/Nothing.h"
// gDebug=1;
.L lib/libEdm.so
.L lib/libStrip.so
// step2.cxx does a lookup that provokes a auto-parsing following by
// an (intentional) compilation error and thus a set of unloading.
.x scripts/step2.cxx
// This use to lead to a strange error (complain that a class is a being
// defined a second time on the ... same line in the same file) which was
// due to the unloading now removing the DefinitionData cached in the forward
// declaration decls left (intentionally) by the unloading.
gInterpreter->AutoParse("reco::utils::ClusterTotals");
```

hahnjo:
So just to summarize: `cmsUnload` will test this fix automatically after the upgrade. I don't see a need to add another test for this.

vgvassilev:
`cmsUnload` is a very complex test which is at integration level, I wanted something that tests the particular feature and when it regresses we can debug and fix it easily without having to build ROOT.

interpreter/cling/lib/Interpreter/DeclUnloader.cpp
```diff
@@ -1042,7 +1042,13 @@ namespace cling {
       Successful &=
           VisitClassTemplateSpecializationDecl(*I, /*RemoveSpec=*/false);
 
-    Successful &= VisitRedeclarableTemplateDecl(CTD);
+    // Visit all redeclarations of this template to ensure the full
+    // redecl chain is unloaded properly.
+    // Not doing this will lead to a corrupted decl chain and can cause
+    // assertions later down the line.
+    for (auto* Redecl : CTD->redecls())
```
hahnjo: As discussed, this will need to be
```suggestion
    llvm::SmallVector<RedeclarableTemplateDecl *> redecls(CTD->redecls());
    for (auto* Redecl : redecls)
```
to avoid messing with the redecl chain while iterating.

devajithvs:
Added a simplified test!

vgvassilev:
I was imagining a cling test but that also should do.

PR #5855, opened 2020-06-11
Remove gEnv and gSystem dependencies from TError

jblomer:
In preparation of moving TError to foundation, this PR removes the direct use of gEnv and gSystem from TError.  The functionality is implemented by helper functions instead, which are registered and de-registered upon creation and destruction gEnv and gSystem.  The three tasks implemented by helpers are

- Read the ROOT environment in order to get the value of `Root.ErrorIgnoreLevel`
- Get the last system error message
- Abort the process

PR #19395, opened 2025-07-17
[core] Simplify ErrorHandler implementation

silverweed:
The current implementation uses a stack-allocated buffer + an optionally- allocated thread local heap buffer for vsnprintf. This likely was made to avoid heap allocating whenever not necessary (i.e. in the majority of cases).
However, a subsequent commit introduced a string `bp` that always performs the copy anyway. At this point the code can be simplified to avoid the double buffer and simply allocate straight into the string avoiding an extra copy.

### DIFF
--- a/core/foundation/src/TError.cxx
+++ b/core/foundation/src/TError.cxx
@@ -108,34 +108,20 @@ ErrorHandlerFunc_t GetErrorHandler()
 
 void ErrorHandler(Int_t level, const char *location, const char *fmt, std::va_list ap)
 {
-   thread_local Int_t buf_size(256);
-   thread_local char *buf_storage(nullptr);
-
-   char small_buf[256];
-   char *buf = buf_storage ? buf_storage : small_buf;
-
-   std::va_list ap_copy;
-   va_copy(ap_copy, ap);
-
    if (!fmt)
       fmt = "no error message provided";
 
-   Int_t n = vsnprintf(buf, buf_size, fmt, ap_copy);
-   if (n >= buf_size) {
-      va_end(ap_copy);
+   std::va_list ap_copy;
+   va_copy(ap_copy, ap);
+   // Get the required buffer size
+   const int required_bytes = vsnprintf(nullptr, 0, fmt, ap) + 1;
 
-      buf_size = n + 1;
-      if (buf != &(small_buf[0]))
-         delete[] buf;
-      buf_storage = buf = new char[buf_size];
-
-      // Try again with a sufficiently large buffer
-      va_copy(ap_copy, ap);
-      vsnprintf(buf, buf_size, fmt, ap_copy);
-   }
+   std::string bp;
+   bp.resize(required_bytes - 1, 0);
+   // Actually print the string to the buffer
+   vsnprintf(bp.data(), required_bytes, fmt, ap_copy);
    va_end(ap_copy);
 
-   std::string bp = buf;
    if (level >= kSysError && level < kFatal) {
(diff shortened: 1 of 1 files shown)


hahnjo (approved):
The reasoning makes sense to me.

pcanal:
> This likely was made to avoid heap allocating whenever not necessary (i.e. in the majority of cases).

yes, this is the case (In particular this is useful because of the not so rare case where the ErrorHandler ends up being called during signal handling).

> However, a subsequent commit introduced a string bp

humm :(  That is actually bad.  The original PR https://github.com/root-project/root/pull/5855, the reviewers completely missed this unexpected but significant change (the commit is "remove gEnv and gSystem deps from TError " ...).

We should revert the part of the commit that introduce the unwanted memory allocation rather than doubling down on it ... (or alternatively re-implement the function so that it returns in another way to having no memory allocation in most cases).

pcanal (changes requested):
See comment on memory allocation

hahnjo:
> We should revert the part of the commit that introduce the unwanted memory allocation rather than doubling down on it ... (or alternatively re-implement the function so that it returns in another way to having no memory allocation in most cases).

Can you explain why you think this is critical? After all, we are talking about an *error* handler, so by definition this should not be on the critical path.

silverweed:
@hahnjo unfortunately the ErrorHandler is also called on `Info` and the likes, so it might actually be called repeatedly (although one might argue that the real performance killer in that case is I/O and not memory allocations...)

hahnjo:
... which wasn't a problem for five years...

silverweed:
Nobody complained about it for five years, which is a weaker statement. But I am also not convinced it's something we actually need to address...

silverweed:
Ping @pcanal

pcanal:
> Can you explain why you think this is critical? After all, we are talking about an error handler, so by definition this should not be on the critical path.

It is not a performance issue.  For better or worse, the error handling and printing can and will be used during signal handling.  During the handling of a signal, memory allocation is very much discouraged and often fatal as it can cause further failures ... and those failures can cause the original issue to be obscured.  

So the current situation is that some portion of the crash/unexpected failures are more difficult to debug/understand as they should.  I would not expect (almost) any user to be able to detect that they have a secondary failure due to the ErrorHandling routine memory allocation and thus I don't expect any one to explicitly complain about it.

silverweed:
If allocating memory is an issue, can't we just have a 10 KB thread local buffer from the get go and assume (almost) all error messages will fit into it, truncating when that's not the case?

silverweed:
@pcanal here is an alternative PR to this that only allocates if the stack buffer is not enough: https://github.com/root-project/root/pull/19478

(my proposal for a single static thread local large buffer remains valid - it would make for much simpler code)

silverweed:
Superseded by https://github.com/root-project/root/pull/19478

pcanal:
> (my proposal for a single static thread local large buffer remains valid - it would make for much simpler code)

It might be solution albeit some medium size buffer times a larger number of thread might become quite high (eg. if some user code uses and then discard a lot of threads).  In term of code simplification, indeed most of the code goes but we still need to handle the case where the user (inadvertently or not) is sending more characters than can fit.

silverweed:
@pcanal my proposed way of "handling" it was to simply truncate - but I agree that having an unbounded amount of memory dedicated to it is not ideal (technically that's also the case for the current solution but memory only grows out of control if the user sends large messages in many different threads, so it's much less bad)

PR #19396, opened 2025-07-17
[CMake] Set relative rpaths to ROOT libraries also in `cppyy-backend`

guitargeek:
Like this, the `cppyy-backend` can be loaded without relying on
`LD_LIBRARY_PATH` to be set to the ROOT library directory.

Also, avoid unneeded symlinks in the install tree by installing the shared libraries for Python inside `CMAKE_INSTALL_PYTHONDIR`.


bindings/pyroot/pythonizations/python/ROOT/__init__.py
```diff
@@ -21,14 +21,20 @@
 # Prevent cppyy from filtering ROOT libraries
 os.environ["CPPYY_NO_ROOT_FILTER"] = "1"
 
+root_module_path = os.path.dirname(__file__) # expected to be ${CMAKE_INSTALL_PYTHONDIR}/ROOT
```
pcanal, line 24: Is the expectation the same on Linux and Windows?

bindings/pyroot/cppyy/CPyCppyy/CMakeLists.txt
```diff
@@ -106,15 +106,9 @@ set_property(GLOBAL APPEND PROPERTY ROOT_EXPORTED_TARGETS ${libname})
 
 # Install library
 install(TARGETS ${libname} EXPORT ${CMAKE_PROJECT_NAME}Exports
-                            RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT libraries
-                            LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT libraries
-                            ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT libraries)
-if (NOT MSVC AND NOT CMAKE_INSTALL_LIBDIR STREQUAL CMAKE_INSTALL_PYTHONDIR)
-  # add a symlink to ${libname} in CMAKE_INSTALL_PYTHONDIR
-  set(LIB_FILE_NAME ${CMAKE_SHARED_LIBRARY_PREFIX}${libname}.so)
-  install(CODE "file(CREATE_LINK ${PYTHONDIR_TO_LIBDIR}/${LIB_FILE_NAME}
```
pcanal: Side note for the record.  In the log 8a7f102d600ba0e984afec0f910cb126dfa3ee1d, it is noted:
```
    libcppyy.so and libcppyy_backend.so are special, as they are needed
    by both python modules (namely, by cppyy) and C++ libraries (e.g. by
    TPython), so if CMAKE_INSTALL_PYTHONDIR is different from
    CMAKE_INSTALL_LIBDIR we install these files in CMAKE_INSTALL_LIBDIR
    but also add symlinks to them in CMAKE_INSTALL_PYTHONDIR.
```
and in the commit in this PR, it is answered:
```
There is no need for these libraries to be in the ROOT install directory
because:

  * `libcppyy.so` is a CPython extension, so it should definitely be in
    the Python directory

  * `libcppyy-backend.so` is always loaded dynamically via `ctypes` from
    the Python modules, so if it's path is fixed relative to the ROOT
    and cppyy Python modules by installing it in the same directory,
    that makes its location more predictable.

Hence, this commit suggests to always install these shared libraries to
the Python install directory, and then we also don't need system links
to the ROOT library locations.

```

guitargeek: Oh! That's a very good catch, I was not having the TPython usecase in mind. That is indeed a good reason to have these system links, although now perfect as it doesn't work on Windows. I don't have an alternative solution ready though, so I'll drop that commit.

guitargeek: In the end, using TPython will also import the `ROOT` Python module actually:

https://github.com/root-project/root/blob/master/bindings/tpython/src/TPython.cxx#L187

So there *is* a way around having the CPython extension in the library search paths. But the problem is that now, TPython is linked against it. The long-term perspective is to load the extension dynamically when doing `import ROOT` in TPython, but there is still work need to be done to make that work.

bindings/pyroot/pythonizations/python/ROOT/__init__.py
guitargeek: Yes, that is the same on Windows.

But to make sure the mechanism really works, I also removed the hardcoded cppyy backend paths in the ROOT tests. That should not be needed anymore.

Issue #19402, opened 2025-07-17
TTree/TChain ResetBranchAddress does not propagate the reset to the clones


pcanal:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

As discovered in https://root-forum.cern.ch/t/possible-issue-when-cloning-trees/63873/15, neither `TChain::ResetBranchAddresses`  nor `TTree::ResetBranchAddresses` seems to reset the address of their current list of clones.

### Reproducer

In pseudo code.
```
auto newtree = chain->CloneTree(0);
chain->SetBranchAddress(branchname, &ptr);
chain->ResetBranchAdress();
assert( chain->GetBranch(branchname)->GetAddress() == newtree->GetBranch(branchname)->GetAddress() );  
```
The current code will fail the assert and 'worse' we have `newtree->GetBranch(branchname)->GetAddress() == &ptr`

### ROOT version

master, v6.36, v6.32

### Installation method

any

### Operating system

any

ferdymercury:
Tangentially related: https://its.cern.ch/jira/browse/ROOT-10328

PR #19403, opened 2025-07-18
[tree] also reset branch address of trees in list of clones

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/19402

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/tree/src/TTree.cxx
+++ b/tree/tree/src/TTree.cxx
@@ -8212,7 +8212,8 @@ void TTree::ResetAfterMerge(TFileMergeInfo *info)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-/// Tell a branch to set its address to zero.
+/// Tell a branch to set its address to zero. This also done for all
+/// trees in the list of clones.
 ///
 /// @note If the branch owns any objects, they are deleted.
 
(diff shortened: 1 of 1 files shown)


dpiparo:
Thanks for this proposed solution, that arrived so quickly after the report of the problem! The changes look good to me, but I'd wait to hear @pcanal opinion about this.

tree/tree/src/TTree.cxx
```diff
@@ -8212,19 +8212,29 @@ void TTree::ResetAfterMerge(TFileMergeInfo *info)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-/// Tell a branch to set its address to zero.
+/// Tell a branch to set its address to zero. This also done for all
```
ferdymercury:
```suggestion
/// Tell a branch to set its address to zero. This is also done for all
```

tree/tree/src/TTree.cxx
```diff
@@ -8248,6 +8258,13 @@ void TTree::ResetBranchAddresses()
          }
       }
    }
+   if (GetListOfClones()) {
+      for (TObjLink *lnk = GetListOfClones()->FirstLink(); lnk; lnk = lnk->Next()) {
+         TTree *clone = (TTree *)lnk->GetObject();
+         if (clone)
+            clone->ResetBranchAddresses();
```
pcanal: If we just reset the addresses of the clone, the address are now valid but the current tree and the clone are still separated.  [As a side note calling `ResetBranchAddresses` actually does too much because it also reset branch address that are unrelated to the original, for example branches that have been added to the clone).
```suggestion
            CopyAddresses(clone);
```

tree/tree/src/TTree.cxx
```diff
@@ -8212,19 +8212,29 @@ void TTree::ResetAfterMerge(TFileMergeInfo *info)
...
 void TTree::ResetBranchAddress(TBranch *br)
 {
    if (br && br->GetTree()) {
       br->ResetAddress();
+      if (GetListOfClones()) {
+         for (TObjLink *lnk = GetListOfClones()->FirstLink(); lnk; lnk = lnk->Next()) {
+            TTree *clone = (TTree *)lnk->GetObject();
+            auto clbr = clone ? clone->FindBranch(br->GetName()) : nullptr;
+            if (clbr)
+               clone->ResetBranchAddress(clbr);
```
pcanal, line 8229: We need to call here the subset of `CopyAddresses` related to the branch and it leafs.  This will require a bit of refactoring of `CopyAddresses` :(

ferdymercury: I think this part is too complex for me to refactor, so I would prefer someone more expert taking over this PR

PR #19404, opened 2025-07-18
[win] Fix build and tests with winrtdebug=ON

bellenot:
- Add the `-RTC1` (runtime check) compiler flag with `winrtdebug=ON` and `dev=ON`
- Add the `-D_ITERATOR_DEBUG_LEVEL=0` flag where needed, to work-around the issue with STL iterators and I/O. This produce incompatible debug binaries, but allows to debug ROOT
- Add the `LLVM_ENABLE_DUMP=TRUE` option when `LLVM_BUILD_TYPE` is `Debug` or `RelWithDebInfo`

### DIFF
--- a/core/base/src/TSystem.cxx
+++ b/core/base/src/TSystem.cxx
@@ -3735,7 +3735,10 @@ int TSystem::CompileMacro(const char *filename, Option_t *opt,
    TString optdebFlags;
    if (mode & kDebug)
       optdebFlags = fFlagsDebug + " ";
-   else if (mode & kOpt)
+#ifdef WIN32
+   else
+#endif
+   if (mode & kOpt)
       optdebFlags += fFlagsOpt;
    cmd.ReplaceAll("$Opt", optdebFlags);
 #ifdef WIN32


interpreter/CMakeLists.txt
```diff
@@ -176,8 +176,10 @@ if(builtin_llvm)
     if(CMAKE_BUILD_TYPE MATCHES "Debug"
      OR LLVM_BUILD_TYPE MATCHES "(Debug|RelWithDebInfo)")
       set(LLVM_ENABLE_ASSERTIONS TRUE)
+      set(LLVM_ENABLE_DUMP TRUE)
```
vgvassilev:
```suggestion
     if(NOT DEFINED LLVM_ENABLE_DUMP)
       set(LLVM_ENABLE_DUMP TRUE)
```

interpreter/CMakeLists.txt
```diff
@@ -176,8 +176,10 @@ if(builtin_llvm)
...
     else()
       set(LLVM_ENABLE_ASSERTIONS FALSE)
+      set(LLVM_ENABLE_DUMP FALSE)
```
vgvassilev, line 184: We probably don't need this line.

vgvassilev (approved):
I used this patch and it worked for me. Looks good, given the minor comment.

interpreter/CMakeLists.txt
bellenot: And what happens if one change the value of `LLVM_BUILD_TYPE` and re-configure?

interpreter/CMakeLists.txt
bellenot: I don't understand why this is needed

vgvassilev: If we pass `-DLLVM_ENABLE_DUMP` from the config line we want to respect that, right?

interpreter/CMakeLists.txt
vgvassilev: We respect what the user has specified?

bellenot: OK, fine

interpreter/CMakeLists.txt
```diff
@@ -176,8 +176,11 @@ if(builtin_llvm)
     if(CMAKE_BUILD_TYPE MATCHES "Debug"
      OR LLVM_BUILD_TYPE MATCHES "(Debug|RelWithDebInfo)")
       set(LLVM_ENABLE_ASSERTIONS TRUE)
+     if(NOT DEFINED LLVM_ENABLE_DUMP)
+       set(LLVM_ENABLE_DUMP TRUE)
```
bellenot:
```suggestion
       set(LLVM_ENABLE_DUMP TRUE)
     endif()
```

interpreter/CMakeLists.txt
```diff
@@ -176,8 +176,12 @@ if(builtin_llvm)
...
+     endif()
```
bellenot:
```suggestion
      if(NOT DEFINED LLVM_ENABLE_DUMP)
        set(LLVM_ENABLE_DUMP TRUE)
      endif()
```

tree/ntuple/src/RColumnElement.hxx
```diff
@@ -993,7 +993,7 @@ using Quantized_t = std::uint32_t;
 #ifdef _MSC_VER
    unsigned long idx = 0;
    if (_BitScanReverse(&idx, x))
-      return static_cast<std::size_t>(31 - idx);
+      return static_cast<std::size_t>(32 - idx);
```
bellenot:
```suggestion
      return static_cast<std::size_t>(31 - idx);
```

core/base/src/TSystem.cxx
```diff
@@ -3729,12 +3735,15 @@ int TSystem::CompileMacro(const char *filename, Option_t *opt,
    TString optdebFlags;
    if (mode & kDebug)
       optdebFlags = fFlagsDebug + " ";
-   if (mode & kOpt)
+   else if (mode & kOpt)
```
pcanal: Except for Windows, debug and optimized mode are not exclusive.

bellenot: OK, I'll make a special case for Windows...

PR #19407, opened 2025-07-18
[skip-ci] Delete .travis.yml

ferdymercury:
I guess that Travis is no longer used / the scripts maintained?


dpiparo:
Thanks, I think so: @vgvassilev ?

PR #19411, opened 2025-07-18
[ntuple] Fix wrong assert in Real32Quant

silverweed:
The assert in QuantizeReals() should not assume that all values are in range, as we want to throw an exception at the end of the function if that is the case.

Another mistake conspired to hide this bug from all non-windows builds:  except on Windows, LeadingZeroes() and TrailingZeroes() were incorrectly using the 64-bit version of the intrinsics, where they should have used the 32-bit one. This means they effectively never fired as they ranged from 0 to 63 rather than the intended 0 to 31.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/src/RColumnElement.hxx
+++ b/tree/ntuple/src/RColumnElement.hxx
@@ -996,7 +996,7 @@ using Quantized_t = std::uint32_t;
       return static_cast<std::size_t>(31 - idx);
    return 32;
 #else
-   return static_cast<std::size_t>(__builtin_clzl(x));
+   return static_cast<std::size_t>(__builtin_clz(x));
 #endif
 }
 
@@ -1011,7 +1011,7 @@ using Quantized_t = std::uint32_t;
       return static_cast<std::size_t>(idx);
    return 32;
 #else
-   return static_cast<std::size_t>(__builtin_ctzl(x));
+   return static_cast<std::size_t>(__builtin_ctz(x));
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


bellenot (approved):
I confirm that this fixes `gtest-tree-ntuple-ntuple-packing` on Windows in Debug mode. So I approve it, at least for the Windows part

jblomer (approved):

tree/ntuple/src/RColumnElement.hxx
```diff
@@ -1042,14 +1042,15 @@ int QuantizeReals(Quantized_t *dst, const T *src, std::size_t count, double min,
    for (std::size_t i = 0; i < count; ++i) {
       const T elem = src[i];
 
-      nOutOfRange += !(min <= elem && elem <= max);
+      bool outOfRange = !(min <= elem & elem <= max);
```
pcanal:
```suggestion
      bool outOfRange = !(min <= elem && elem <= max);
```

silverweed: I changed it on purpose as godbolt shows that it can make the loop branchless

hahnjo: Unless actually measured, branchless code isn't necessarily better than what the compiler generates by default. I would suggest following the default pattern of combining two `bool` values using `&&`, as suggested by the warning.

pcanal: > I changed it on purpose as godbolt shows that it can make the loop branchless

Fair enough.  If it had really been the right choice (see Jonas comment) then I would have recommend to add both parenthesis and a comment to clarify that it was intentional and why.

PR #19414, opened 2025-07-19
[RF] Handle variables having numeric value as name in RooFormula

### LINKED ISSUE #11992: [RF] RooFormulaVar is too eager in variable name substitution

Ban42:
# This Pull request:

As requested in #11992 changes the way RooFormulaVar performs variable names substitutions.

## Changes or fixes:
Before performing a substitution in the formula based on variable name, the code now checks if the variable to be substituted has a number as name. 
Variables where the name is a number are now only be matched by index. If there is a variable with a name that is a number that is not a RooConstVar where the value is equal to the name, an exception will be thrown. 

## Checklist:

- [x] tested changes locally: Yes, 227 - gtest-roofit-roofitcore-testRooFormula fails, throwing one of the two exceptions I have created. I either have to update it, or double-check for mistakes on my side.
- [ ] updated the docs : Let me know if I should.
- [x] Add tests: I probably need to add a test for this

This PR fixes #11992

### DIFF
--- a/roofit/roofitcore/test/testRooFormula.cxx
+++ b/roofit/roofitcore/test/testRooFormula.cxx
@@ -1,10 +1,12 @@
 // Tests for the RooFormula
 // Authors: Stephan Hageboeck, CERN  2020
 //          Jonas Rembser, CERN 2023
+//          Andrea Germinario, CERN 2025
 
 #include "../src/RooFormula.h"
 #include <RooFormulaVar.h>
 #include <RooRealVar.h>
+#include <RooConstVar.h>
 
 #include <ROOT/TestSupport.hxx>
 
@@ -56,7 +58,7 @@ TEST(RooFormula, TestDangerousVariableNames)
 {
    RooRealVar dt("dt", "dt", -10, 10);
    RooRealVar x("x", "x", 1.547);
-   RooRealVar zero("0", "0", 0);
+   RooConstVar zero("0", "0", 0);
 
    // Create the formula, triggers an error if the formula doesn't compile
    // correctly because the dangerous variable names haven't been treated right.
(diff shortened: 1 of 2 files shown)


Ban42:
Hi @guitargeek ,
This is the solution I have found which should match the specification you described in #11992 

The test "227 - gtest-roofit-roofitcore-testRooFormula" fails, which means I have made a mistake or I have to update it. Before doing so, I wanted you to have a quick look and confirm this approach is the desired one.

Thank you :)

dpiparo:
let's start with the CI, to see where we stand.

roofit/roofitcore/test/testRooFormula.cxx
```diff
@@ -5,6 +5,7 @@
 #include "../src/RooFormula.h"
```
Ban42, line 6: Hi guys,
I know is a bit silly, but since I have added a new test here can I add my name in the list of authors at the beginning of the file? :grimacing:

dpiparo: yes, of course.

dpiparo:
Let's fix the formatting in the test (which can be coupled with adding the author name to the file). After that, if @guitargeek agrees, for me this is ready to be merged.

dpiparo:
rebased the code and restarted the tests... @guitargeek fyi

PR #19420, opened 2025-07-21
[ntuple] Fix RBitsetField memory corruption on Windows

### LINKED ISSUE #19410: [ntuple] Heap corruption on Windows in ntuple_show
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

In Windows debug builds we observe a heap corruption in ntuple_show:

Image

We should try and reproduce it with asan on Linux

### Reproducer

Launch ntuple_show on Windows Debug

### ROOT version

master

### Installation method

from source

### Operating system

Windows

### Additional context

_No response_

silverweed:
std::bitset on windows behaves differently from clang/gcc: it allocates an array of type T where T may be a u32 or u64 depending on the number of bits of the bitset itself. As we are currently assuming std::bitset is always backed up by an array of `unsigned long` (which is 32 bits on windows), we are miscalculating the required size for a bitset of > 64 bits (e.g. we only allocate 3 words of 32 bits for a bitset<65>, but it really is backed up by 2 64-bit words), resulting in memory corruption.

This PR should fix #19410.

Opening as a draft to let the CI run as I can't seem to run all tests locally.

### DIFF
--- a/tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
+++ b/tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
@@ -82,14 +82,12 @@ public:
 /// Template specializations for C++ std::bitset
 ////////////////////////////////////////////////////////////////////////////////
 
-/// The generic field an `std::bitset<N>`. All compilers we care about store the bits in an array of unsigned long.
+/// The generic field for a `std::bitset<N>`.
+/// GCC and Clang store the bits in an array of unsigned long, whereas MSVC uses either a single uint32_t for `N <= 32`
+/// or an array of uint64_t for `N > 32` (reference: https://github.com/microsoft/STL/blob/main/stl/inc/bitset).
 /// TODO(jblomer): reading and writing efficiency should be improved; currently it is one bit at a time
 /// with an array of bools on the page level.
 class RBitsetField : public RFieldBase {
-   using Word_t = unsigned long;
-   static constexpr std::size_t kWordSize = sizeof(Word_t);
-   static constexpr std::size_t kBitsPerWord = kWordSize * 8;
-
 protected:
    std::size_t fN;
 
@@ -106,14 +104,40 @@ protected:
    void ReadGlobalImpl(ROOT::NTupleSize_t globalIndex, void *to) final;
    void ReadInClusterImpl(RNTupleLocalIndex localIndex, void *to) final;
 
+   size_t WordSize() const
+   {
+#if defined(_MSC_VER)
+      const size_t wordSize = (fN <= sizeof(unsigned long) * 8) ? sizeof(unsigned long) : sizeof(unsigned long long);
+#else
+      const size_t wordSize = sizeof(unsigned long);
+#endif
+      return wordSize;
+   }
+
+   template <typename FUlong, typename FUlonglong, typename... Args>
+   void SelectWordSize(FUlong &&fUlong, FUlonglong &&fUlonglong, Args &&...args);
+
 public:
    RBitsetField(std::string_view fieldName, std::size_t N);
    RBitsetField(RBitsetField &&other) = default;
    RBitsetField &operator=(RBitsetField &&other) = default;
    ~RBitsetField() override = default;
 
-   size_t GetValueSize() const final { return kWordSize * ((fN + kBitsPerWord - 1) / kBitsPerWord); }
-   size_t GetAlignment() const final { return alignof(Word_t); }
+   size_t GetValueSize() const final
+   {
+      const size_t bitsPerWord = WordSize() * 8;
+      return WordSize() * ((fN + bitsPerWord - 1) / bitsPerWord);
+   }
+
+   size_t GetAlignment() const final
+   {
+#if defined(_MSC_VER)
+      return WordSize() == sizeof(unsigned long) ? alignof(unsigned long) : alignof(unsigned long long);
+#else
+      return alignof(unsigned long);
+#endif
+   }
+
    void AcceptVisitor(ROOT::Detail::RFieldVisitor &visitor) const final;
 
    /// Get the number of bits in the bitset, i.e. the `N` in `std::bitset<N>`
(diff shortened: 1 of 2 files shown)


tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
```diff
@@ -82,14 +82,12 @@ public:
 /// Template specializations for C++ std::bitset
 ////////////////////////////////////////////////////////////////////////////////
 
-/// The generic field an `std::bitset<N>`. All compilers we care about store the bits in an array of unsigned long.
+/// The generic field an `std::bitset<N>`.
```
jblomer:
```suggestion
/// The generic field for a `std::bitset<N>`.
```

tree/ntuple/src/RField.cxx
```diff
@@ -647,45 +647,84 @@ void ROOT::RBitsetField::GenerateColumns(const ROOT::RNTupleDescriptor &desc)
...
+      R__ASSERT(false);
+   }
+}
+
+template <typename Word_t>
+static void BitsetAppendImpl(const void *from, size_t nBits, ROOT::Internal::RColumn &column)
 {
+   constexpr auto kBitsPerWord = sizeof(Word_t) * 8;
+
    const auto *asULongArray = static_cast<const Word_t *>(from);
```
jblomer:
```suggestion
   const auto *asWordArray = static_cast<const Word_t *>(from);
```

tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
```diff
@@ -106,14 +104,40 @@ protected:
...
+   size_t GetValueSize() const final
+   {
+      const size_t bitsPerWord = WordSize() * 8;
+      return WordSize() * ((fN + bitsPerWord - 1) / bitsPerWord);
+   }
+
+   size_t GetAlignment() const final
+   {
+#if defined(_MSC_VER)
+      return fN <= sizeof(unsigned long) ? alignof(unsigned long) : alignof(unsigned long long);
```
hahnjo: This looks wrong: `fN` is in bits while `sizeof(unsigned long)` is in bytes...

tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
```diff
@@ -199,7 +223,9 @@ class ROptionalField : public RNullableField {
 
    public:
       ROptionalDeleter(std::unique_ptr<RDeleter> itemDeleter, std::size_t engagementPtrOffset)
-         : fItemDeleter(std::move(itemDeleter)), fEngagementPtrOffset(engagementPtrOffset) {}
+         : fItemDeleter(std::move(itemDeleter)), fEngagementPtrOffset(engagementPtrOffset)
+      {
+      }
```
hahnjo: unrelated change?

hahnjo:
> std::bitset on windows behaves differently from clang/gcc: it allocates an array of type T where T may be a u32 or u64 depending on the number of bits of the bitset itself. As we are currently assuming std::bitset is always backed up by an array of u64, we are miscalculating the required size for a bitset of > 64 bits (e.g. we only allocate 3 words of 32 bits for a bitset<65>, but it really is backed up by 2 64-bit words), resulting in memory corruption.

Can you check this description? As far as I understand (our) code (didn't check the MSVC STL), it uses `u32` only for `fN <= 32`. So for the example of `bitset<65>` it should indeed be backed by two `u64`, no?

tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
silverweed: You're right, I'm missing a `* 8`; thanks!

tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
silverweed: I probably clang-formatted the entire file accidentally

tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
hahnjo: Maybe you want to use
```suggestion
      return WordSize() == sizeof(unsigned long) ? alignof(unsigned long) : alignof(unsigned long long);
```
as you do in `SelectWordSize`?

silverweed:
> Can you check this description? As far as I understand (our) code (didn't check the MSVC STL), it uses u32 only for fN <= 32. So for the example of bitset<65> it should indeed be backed by two u64, no?

In our code prior to this change, we always use an array of `(fN + bitsPerWord - 1) / bitsPerWord` elements to represent the bitset; on windows `bitsPerWord = wordSize * 8 = 4 * 8  = 32` (since on Windows `sizeof(unsigned long) = 4`) so in case of a 65-bit bitset we get:
```
valueSize = wordSize * (fN + bitsPerWord - 1) / bitsPerWord = 4 * (65 + 32 - 1) / 32 = 12
```
which is equivalent to 3 `uint32_t`; but windows actually uses
```
valueSize = longWordSize * (fN + bitsPerLongWord - 1) / bitsPerLongWord = 8 * (65 + 64 - 1) / 64 = 16
```
equivalent to 2 `uint64_t`. Therefore we end up allocating too little memory for it.

tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
silverweed: That works too, yes

hahnjo:
But then

> As we are currently assuming std::bitset is always backed up by an array of u64

should rather say

> As we are currently assuming std::bitset is always backed up by an array of `unsigned long` (which has 32 bits on Windows)

?

silverweed:
Yes, that's right. I'll fix it.

hahnjo:
> Yes, that's right. I'll fix it.

Unfortunately didn't make it into the commit message :neutral_face:

silverweed:
@hahnjo crap, I changed it in the PR itself but forgot to push the local amended commit...

PR #19425, opened 2025-07-22
[hist] Add standalone CMake setup & workflow

hahnjo:


hahnjo:
> Is this the first package we can build stand-alone?

I believe you can also build Minuit2 outside of ROOT, but there's some trickery because it includes some core headers. That's what I want to avoid here, also by having an automatic check that we're not breaking it.

Disclaimer: I'm not actually sure a standalone build will be super useful. If it's annoying to maintain, we can decide to drop it at any time. But for the moment, it's maybe a good idea to ensure that a (new) leaf package actually has that property.

PR #19426, opened 2025-07-22
[hist] Throw on streaming `Experimental` classes

hahnjo:


hahnjo:
This is PR is now rebased and seems to work properly.

jblomer (approved):

PR #19427, opened 2025-07-22
[skip-ci][imt] Remove documentation on ROOT namespace

hahnjo:
This is confusing in the Doxygen documentation, and the comment is actually the explanation why we need `RTaskArenaWrapper`.


vepadulano (approved):
Thanks!

PR #19428, opened 2025-07-22
[ci] Check if Linux platforms work with `builtin_glew=ON`

guitargeek:
Some investigation in the context of this JIRA issue:
https://its.cern.ch/jira/browse/ROOT-10756


guitargeek:
Good to know that this still works!

Issue #11616, opened 2022-10-20
[ntuple] Check the usage of `Long_t` in RField (size of the type might differ between 32- and 64-bit platforms)


jalopezg-r00t:
- [X] Checked for duplicates

### Describe the bug
As discussed in PR #11525 ([here](https://github.com/root-project/root/pull/11525#discussion_r997329752)), the type `Long_t` might have different sizes on 32- or 64- bit platforms.

Both, `RFieldBase::Create()` (see [here](https://github.com/root-project/root/blob/master/tree/ntuple/v7/src/RField.cxx#L77)), and the aforementioned PR incorrectly assume this to be equivalent to `std::int64_t` - which, again, might or not be the case.

jblomer:
Fixed by #16039

hahnjo:
Yes, https://github.com/root-project/root/pull/16039 takes care of `Long_t`, but we still need to update `RProxiedCollectionField` to remove the hard-coded fixed width integer types.

hahnjo:
Yes, I believe with

> we still need to update `RProxiedCollectionField` to remove the hard-coded fixed width integer types.

I mean this part of the constructor:
https://github.com/root-project/root/blob/05e4403401f8be27cf58fce0fc0410e781d5a1ff/tree/ntuple/v7/src/RFieldMeta.cxx#L643-L658

The solution may be as easy as constructing `RField<long>` and so on, and let `RIntegralTypeMap` figure out the rest. That needs a closer look though...

PR #19429, opened 2025-07-22
[ntuple] Use dynamic platform-dependent width rather than hard-coded width of long

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/11616

## Checklist:
- [ ] tested changes locally

### DIFF
--- a/tree/ntuple/src/RFieldMeta.cxx
+++ b/tree/ntuple/src/RFieldMeta.cxx
@@ -532,9 +532,9 @@ ROOT::REnumField::REnumField(std::string_view fieldName, TEnum *enump)
    case kUShort_t: Attach(std::make_unique<RField<uint16_t>>("_0")); break;
    case kInt_t: Attach(std::make_unique<RField<int32_t>>("_0")); break;
    case kUInt_t: Attach(std::make_unique<RField<uint32_t>>("_0")); break;
-   case kLong_t:
+   case kLong_t: Attach(std::make_unique<RField<long>>("_0")); break;
    case kLong64_t: Attach(std::make_unique<RField<int64_t>>("_0")); break;
-   case kULong_t:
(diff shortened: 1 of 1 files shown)


hahnjo (commented):
I think we should generally replace all instances of this pattern (ie also for `kInt_t`, `kUInt_t` etc.) and fully rely on `RField` to do the mapping.

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -532,9 +532,9 @@ ROOT::REnumField::REnumField(std::string_view fieldName, TEnum *enump)
    case kUShort_t: Attach(std::make_unique<RField<uint16_t>>("_0")); break;
    case kInt_t: Attach(std::make_unique<RField<int32_t>>("_0")); break;
    case kUInt_t: Attach(std::make_unique<RField<uint32_t>>("_0")); break;
-   case kLong_t:
+   case kLong_t: Attach(std::make_unique<RField<long>>("_0")); break;
```
pcanal: Completely related question.  Even-though the in-memory size varies (32  or 64 bits), do we want the on-file size to vary (likely the consequence of this change)?

jblomer: It's consistent with the rest of RNTuple writing: integer types are recorded with their bit width at writing time.

ferdymercury:
> I think we should generally replace all instances of this pattern (ie also for `kInt_t`, `kUInt_t` etc.) and fully rely on `RField` to do the mapping.

Do you mean eg changing

`case kUShort_t: Attach(std::make_unique<RField<uint16_t>>("_0")); break;`
 
With

`case kUShort_t: Attach(std::make_unique<RField<UShort_t>>("_0")); break;`

?

PR #19430, opened 2025-07-22
[ci] Check if CI platforms work with `builtin_cfitsio=ON`

guitargeek:
To try in the CI, because it doesn't work for me locally.


guitargeek:
Doesn't work on Windows, but the fix is already on the way for the next cfitsio release (thanks @bellenot for spotting this):
https://github.com/HEASARC/cfitsio/commit/ef0a8967ab2a3ea60bc65157d1e752f59d61b830

Let's update the builtin later then, and for now I mark the PR as draft.

dpiparo:
Could also the RN be updated in this PR?

guitargeek:
> Could also the RN be updated in this PR?

Will do. I have to update the PR once more anyway, to remove the temporary global `builtin_cfitsio=ON` to test compilation on all platforms

guitargeek:
It worked! I'll remove the temporary changes to the CI from this PR now, and then this is ready to review and merge

guitargeek:
@dpiparo, the PR is ready

PR #19431, opened 2025-07-22
[v632][ci] Check if CI platforms work with `builtin_cfitsio=ON`

guitargeek:
To test whether it worked in older releases.


guitargeek:
Ok, so the problem is only on the new Fedoras

PR #19434, opened 2025-07-22
[ntuple] Change RMiniFile::ReadNTupleProperAtOffset

silverweed:
Now it accepts the offset of the payload, its nbytes and uncompressed size rather than the key offset. This will be needed by the RNTuple Attributes.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/src/RMiniFile.cxx
+++ b/tree/ntuple/src/RMiniFile.cxx
@@ -754,14 +754,6 @@ ROOT::RResult<ROOT::RNTuple> ROOT::Internal::RMiniFileReader::GetNTupleProper(st
       return R__FAIL("no RNTuple named '" + std::string(ntupleName) + "' in file '" + fRawFile->GetUrl() + "'");
    }
 
-   auto res = GetNTupleProperAtOffset(offset);
-   return res;
-}
-
-ROOT::RResult<ROOT::RNTuple> ROOT::Internal::RMiniFileReader::GetNTupleProperAtOffset(std::uint64_t keyOffset)
-{
-   auto offset = keyOffset;
-   RTFKey key;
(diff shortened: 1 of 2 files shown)


tree/ntuple/src/RMiniFile.cxx
```diff
@@ -772,29 +764,37 @@ ROOT::RResult<ROOT::RNTuple> ROOT::Internal::RMiniFileReader::GetNTupleProperAtO
...
+{
    // The object length can be smaller than the size of RTFNTuple if it comes from a past RNTuple class version,
    // or larger than it if it comes from a future RNTuple class version.
-   auto bufAnchor = MakeUninitArray<unsigned char>(std::max<size_t>(key.fObjLen, sizeof(RTFNTuple)));
+   auto bufAnchor = MakeUninitArray<unsigned char>(std::max<size_t>(uncompLen, sizeof(RTFNTuple)));
    RTFNTuple *ntuple = new (bufAnchor.get()) RTFNTuple;
 
-   const auto objNbytes = key.GetSize() - key.fKeyLen;
-   if (objNbytes != key.fObjLen) {
+   if (compSize != uncompLen) {
```
pcanal, line 781: Minor question.  Both values are size, both values are length.  Is there value is using a different 'synonym' for each?

pcanal (approved):
LGTM.

tree/ntuple/src/RMiniFile.cxx
silverweed: It's just a convention we use throughout the RNTuple codebase, where size on disk is "size" or "nBytes" and size in memory is "length". I think it's purely arbitrary but we try to be consistent with it.

Issue #19358, opened 2025-07-14
cppyy.gbl.std.function(...) fails for template types like std::vector<int>, ROOT::RVec<int>


siliataider:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

In using cppyy through ROOT, it seems like basic function signatures work with `std::function`, any signature involving template types however (e.g., STL containers or RVec) fails with a `TypeError`, even if the types are valid.

### Reproducer

```
import cppyy

print(cppyy.gbl.std.function("bool(int)"))  # works
print(cppyy.gbl.std.function("bool()"))     # works

print(cppyy.gbl.std.function("bool(std::vector<int>)")) # fails with TypeError
```

Error:
```
TypeError: 'std::function<bool(std::vector<int>)>' is not a known C++ class
```

Even though this is valid c++:
```
std::function<bool(std::vector<int>)> fn = [](std::vector<int> x) {
    return true;
};
```

These also similarily fail:
```
cppyy.gbl.std.function("bool(ROOT::RVec<int>)")
cppyy.gbl.std.function("bool(MyStruct_t<int>)")
```

### ROOT version

6.36

### Installation method

build form source

### Operating system

Linux, Fedora

siliataider:
P.S @Vipul-Cariappa tried the same line with upstream cppyy and it worked

ferdymercury:
Maybe related:
https://github.com/root-project/root/issues/16613
https://github.com/root-project/root/issues/10680
https://github.com/root-project/root/issues/16064

Vipul-Cariappa:
I took a look at this. Not sure if it is worth the effort to fix it now. This will work once we migrate to CppInterOp's cppyy.
#10680 and #16064 also work with CppInterOp's cppyy.

siliataider:
A first debug round shows me there is a parsing issue: [this line](https://github.com/root-project/root/blob/d6c2719c0fc164337d7c64f0d01a1552e705507e/bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx#L499) returns `std::function<bool(vector<int> >` while processing `std::function<bool(vector<int>)>` (missing parenthesis), but I think there is more to it than just the parsing

Vipul-Cariappa:
Yes. There is that thing too. We should consider fixing that I guess. Might enable some of the use cases.

[...]

PR #9138, opened 2021-10-18
[PyROOT] Get names and types of all overloads' signature

vepadulano:
This PR introduces a machinery to get the names and types (separately) of the signature of each overload of a particular `CPPOverload` object. This allows for example for something like.

**EDIT**

Here is the latest status of this feature after some discussion:

```python
>>> import ROOT
>>> ROOT.gInterpreter.Declare("""
 int foo(int a, float b);
 int foo(int a);
 float foo(float b);
 double foo(int a, float b, double c);
... """)
True
>>> ROOT.foo
<cppyy.CPPOverload object at 0x7fc470789fd0>
>>> from pprint import pprint
>>> pprint(ROOT.foo.func_overloads_names)
{'double ::foo(int a, float b, double c)': ('a', 'b', 'c'),
 'float ::foo(float b)': ('b',),
 'int ::foo(int a)': ('a',),
 'int ::foo(int a, float b)': ('a', 'b')}
>>> pprint(ROOT.foo.func_overloads_types)
{'double ::foo(int a, float b, double c)': {'input_types': ('int',
                                                            'float',
                                                            'double'),
                                            'return_type': 'double'},
 'float ::foo(float b)': {'input_types': ('float',), 'return_type': 'float'},
 'int ::foo(int a)': {'input_types': ('int',), 'return_type': 'int'},
 'int ::foo(int a, float b)': {'input_types': ('int', 'float'),
                               'return_type': 'int'}}
```

PR #19437, opened 2025-07-22
[Python][RDF] Enable Support for C++ Free Functions in RDF Define and Filter

siliataider:
# This Pull request:

This PR extends the `RDataFrame` pythonizations by adding support for passing C++ free functions to `.Define()` and `Filter` operations, with support for:
- Templated functions
- Overloaded functions
- Functions taking any input/output combination  of
    - Fundamental types
    - ROOT types (e.g. TObject-derived classes)
    - User-defined C++ classes

The implementation works by converting the free function to `std::function` using the function's signature.

> [!warning]
> This currently doesn't work if the function takes templated types as input arguments like STL containers (e.g. `std::vector<T>`) because of #19358
> Concretely, this doesn't work:
> ```cpp
> bool has_even_sum(const RVec<int>& v) {
>    return Sum(v) % 2 == 0;
> } 
> ```

> [!important]
> Prerequisites / TODO:
> - [x] Bring #9138 up to date and merge it: this implementation relies on the API provided by this PR for selecting the proper overload.
> - [ ] #19358: this would unlock a wide range of use-cases, like the example mentioned above.
>        - This should not block this PR though
> - [x] Update documentation / examples

**Quick demo:**
```python
import ROOT

ROOT.gInterpreter.Declare(""" 
struct MyStruct {
    int a;
    MyStruct() : a(0) {}
    MyStruct(int a_) : a(a_) {}
};
                          
float struct_a(const MyStruct& s) {
    return s.a;
}                          

bool is_even(int x) {
    return x % 2 == 0;
}
                          
bool is_even(int x, float y) {
    return x % 2 == 0 && x == y;
}                                                 

template <typename T>
bool is_even_t(const T x, const float y) {
    return x % 2 == 0 && x == y;
}
""")

rdf = ROOT.RDataFrame(3) \
         .Define("x", "(int)rdfentry_") \
         .Define("y", "(float)rdfentry_") \
         .Define("is_even_x", ROOT.is_even, ["x"]) \
         .Define("is_even_x_y", ROOT.is_even, ["x", "y"]) \
         .Define("is_even_t", ROOT.is_even_t[int], ["x", "y"]) \
         .Define("my_struct", "MyStruct((int)rdfentry_)") \
         .Define("struct_a", ROOT.struct_a, ["my_struct"]) 
         
rdf.Display().Print()
```

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/TemplateProxy.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/TemplateProxy.cxx
@@ -729,6 +729,21 @@ static int tpp_setuseffi(CPPOverload*, PyObject*, void*)
     return 0;                 // dummy (__useffi__ unused)
 }
 
+//-----------------------------------------------------------------------------
+static PyObject* tpp_gettemplateargs(TemplateProxy* self, void*) {
+    if (!self->fTemplateArgs) {
+        Py_RETURN_NONE;
+    }
+
+    Py_INCREF(self->fTemplateArgs);
+    return self->fTemplateArgs;
+}
+
+//-----------------------------------------------------------------------------
+static int tpp_settemplateargs(TemplateProxy*, PyObject*, void*) {
+    PyErr_SetString(PyExc_AttributeError, "__template_args__ is read-only");
+    return -1;
+}
 
 //----------------------------------------------------------------------------
 static PyMappingMethods tpp_as_mapping = {
@@ -739,7 +754,9 @@ static PyGetSetDef tpp_getset[] = {
     {(char*)"__doc__", (getter)tpp_doc, (setter)tpp_doc_set, nullptr, nullptr},
     {(char*)"__useffi__", (getter)tpp_getuseffi, (setter)tpp_setuseffi,
       (char*)"unused", nullptr},
-    {(char*)nullptr,   nullptr,         nullptr, nullptr, nullptr}
+    {(char*)"__template_args__", (getter)tpp_gettemplateargs, (setter)tpp_settemplateargs,
+      (char*)"the template arguments for this method", nullptr},
+    {(char*)nullptr,   nullptr,         nullptr, nullptr, nullptr},
 };
 
 
@@ -770,6 +787,7 @@ static PyObject* tpp_overload(TemplateProxy* pytmpl, PyObject* args)
 {
 // Select and call a specific C++ overload, based on its signature.
     const char* sigarg = nullptr;
+    const char* tmplarg = nullptr;
     PyObject* sigarg_tuple = nullptr;
     int want_const = -1;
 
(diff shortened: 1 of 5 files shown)


bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdf_pyz.py
```diff
@@ -250,14 +250,102 @@ def _convert_to_vector(args):
...
+    """
+    Gets the C++ overload from a cppyy TemplateProxy using the input
+    and template arguments types.
+    """
+    signature = ", ".join(types) if types else ""
+    template_args = func.__template_args__[1:-1] if func.__template_args__ else ""
+    return func.__overload__(signature, template_args)
+
+def get_column_types(rdf, cols):
+    if rdf is None:
```
vepadulano: In what scenario is this branch hit?

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdf_pyz.py
```diff
@@ -250,14 +250,102 @@ def _convert_to_vector(args):
...
+        return next(iter(overload_types))
+
+    for full_sig, info in overload_types.items():
+        input_types = info.get('input_types', ())
+
+        if len(input_types) != len(types):
+            continue
+
+        match = True
+        for expected, actual in zip(input_types, types):
```
vepadulano, line 308: So here `actual` comes from `df.GetColumnType`? Why is a loose match enough in this case?

bindings/pyroot/pythonizations/test/rdf_define_pyz.py
```diff
@@ -113,6 +113,107 @@ def test_std_function(self):
         for x,y in zip(rdf2.Take['ULong64_t']("rdfentry_"), rdf2.Take['ULong64_t']("x")):
```
vepadulano: Thanks for all the new tests! Can you also add some test with more than one input argument where at least one of the types is a user-defined class and/or a `std::` class? I guess for example `std::string` should be fine to circumvent https://github.com/root-project/root/issues/19358

vepadulano (changes requested):
Thank you, this is great! I left a couple of minor remarks

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdf_pyz.py
siliataider: in theory none, I removed it

bindings/pyroot/pythonizations/test/rdf_define_pyz.py
siliataider: yep done

bindings/pyroot/cppyy/CPyCppyy/src/TemplateProxy.cxx
```diff
@@ -813,7 +813,7 @@ static PyObject* tpp_overload(TemplateProxy* pytmpl, PyObject* args)
         scope = ((CPPClass*)pytmpl->fTI->fPyClass)->fCppType;
         cppmeth = Cppyy::GetMethodTemplate(
             scope, pytmpl->fTI->fCppName, proto.substr(1, proto.size()-2));
-    } else if (PyArg_ParseTuple(args, const_cast<char*>("ss|i:__overload__"), &sigarg, &tmplarg, &want_const)) {
+    } else if (PyArg_ParseTuple(args, const_cast<char*>("ss:__overload__"), &sigarg, &tmplarg)) {
```
vepadulano, line 816: Just noticing now that there is this extra line changed in `TemplateProxy.cxx` which is changed in the commit `d5bb6f0` separately from commit `e842df`. I just wanted to ensure this is deliberate.

siliataider: nice catch, fixed, thanks :)

PR #19439, opened 2025-07-23
[skip-ci][eve7] Remove documentation on ROOT namespace

hahnjo:
This is confusing in the Doxygen documentation.

(see also https://github.com/root-project/root/pull/19427)


graf3d/eve7/inc/ROOT/REveVSDStructs.hxx
```diff
@@ -17,7 +17,7 @@
 #include <ROOT/REveVector.hxx>
 
 ////////////////////////////////////////////////////////////////////////////////
-/// VSD Structures
+// VSD Structures
```
ferdymercury, line 20:
```suggestion
/// \file
/// \brief Basic structures for REve VSD concept
```
Could be an alternative if we do not want to lose the docu-overview of this file.

hahnjo: I would prefer not to change anything more than necessary. Also the documentation overview was never associated to the file AFAICT...

linev (approved):

Issue #7786, opened 2021-04-06
Building built-in libAfterImage fails if build path contains `@`


eguiraud:
- [X] Checked for duplicates

### Describe the bug

The symptom is the following:

```
/usr/bin/mkdir -p /data/ssdext4/rdf_scaling/relwithdebinfo@master/include/libAfterImage
installing libAfterImage headers using command -
    "-m 644 <header> /data/ssdext4/rdf_scaling/relwithdebinfo@master/include/libAfterImage" :
afterimage.h afterbase.h ascmap.h asfont.h asim_afterbase.h asimage.h asimagexml.h asstorage.h asvisual.h blender.h bmp.h char2uni.h draw.h export.h imencdec.h import.h pixmap.h scanline.h transform.h ungif.h xcf.h ximage.h xpm.h xwrap.h

/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
/bin/sh: line 30: -m: command not found
gmake[3]: m: Command not found
/bin/sh: line 2: -m: command not found
gmake[3]: *** [Makefile:145: install.static] Error 127
```

The cause seems to be the `sed` substitutions at https://github.com/root-project/root/blob/aee48761f0c4a4d91d0a16c367ca3b66a597a484/graf2d/asimage/src/libAfterImage/configure#L2834 (thanks @chrisburr )

[...]

PR #19443, opened 2025-07-23
Fix missing GIF support in TASimage

hageboeck:
The builtin afterimage by default doesn't enable support for gif.
Therefore, in addition to passing include directories for gif support,
one also has to enable it.

Fix https://github.com/root-project/root/issues/19412
Fixup of https://github.com/root-project/root/pull/18074 https://github.com/root-project/root/commit/ed789d067eb739f01ed5dd4bdddc0744dddba548

Fix https://github.com/root-project/root/issues/7786


roottest/root/graf/testImageFormats.cxx
```diff
@@ -0,0 +1,9 @@
+#include <TGClient.h>
+
+#include <gtest/gtest.h>
+
+TEST(ImageFormats, GIF) {
+  // Import the GoBack icon from ROOT's icon directory
+  auto icon = gClient->GetPicture("GoBack.gif");
+  ASSERT_NE(icon, nullptr);
```
ferdymercury: Thanks for this!
As seen in https://github.com/root-project/root/issues/19412#issuecomment-3097703307
the return value is not a nullptr even if there is an error when loading the file, so this test will always succeed.
Maybe adding
ASSERT_NE(icon->GetMask(), nullptr);
or capturing diagnostic helps?

roottest/root/graf/CMakeLists.txt
```diff
@@ -1,3 +1,3 @@
 ROOTTEST_ADD_TEST(textcopy MACRO  textcopy.C
                            OUTREF textcopy.ref)
-
+ROOT_ADD_GTEST(testImageFormats testImageFormats.cxx LIBRARIES ASImage)
```
ferdymercury:
```suggestion
ROOT_ADD_GTEST(testImageFormats testImageFormats.cxx LIBRARIES ASImage Gui)
```

roottest/root/graf/testImageFormats.cxx
hageboeck: OK, good point. In my local tests, the image was actually `nullptr`, but going one step further by asking for the mask will make it better!
Diagnostics are always captured in all gtests, and will fail the test, so that's a done deal!

hageboeck: > ASSERT_NE(icon->GetMask(), nullptr);

This happens to be nullptr, but the width is not, so I'm using this now.

roottest/root/graf/testImageFormats.cxx
```diff
@@ -0,0 +1,22 @@
+#include <TGClient.h>
+#include <TGPicture.h>
+
+#include <gtest/gtest.h>
+
+// Import a few icons from ROOT's icon directory
+
+TEST(ImageFormats, GIF)
+{
+   auto icon = gClient->GetPicture("GoBack.gif");
```
ferdymercury, line 10:
```suggestion
   if (!gClient)
      gClient = TGClient::Instance();
   auto icon = gClient->GetPicture("GoBack.gif");
```

ferdymercury, line 12:
```suggestion
   ASSERT_NE(gClient, nullptr);
   auto icon = gClient->GetPicture("GoBack.gif");
```

roottest/root/graf/testImageFormats.cxx
```diff
@@ -0,0 +1,22 @@
...
+   ASSERT_NE(icon, nullptr);
+   ASSERT_NE(icon->GetWidth(), 0u);
+   delete icon;
+}
+
+TEST(ImageFormats, PNG)
+{
+   auto icon = gClient->GetPicture("pause.png");
```
ferdymercury, line 22:
```suggestion
   ASSERT_NE(gClient, nullptr);
   auto icon = gClient->GetPicture("pause.png");
```

roottest/root/graf/testImageFormats.cxx
hageboeck: `ASSERT` would still fail the test if the graphics system is not initialised. I think I have an idea, pushing now.

ferdymercury: (Yep, it would just give a more informative error rather than a sigsegv)

ferdymercury (approved):

vepadulano (approved):
LGTM thanks!

PR #19444, opened 2025-07-23
[ntuple] Renormalize type name before comparing

### LINKED ISSUE #19442: [ntuple] Unable to read back RNTuples without normalized type names from v6.34
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

In particular, this affects ATLAS OpenData produced with ROOT v6.34.04.

### Reproducer

Trying to open and read one of the converted PHYSLITE files gives:
```
incompatible type name for field _0: ElementLink<DataVector<xAOD::Jet_v1>> vs. ElementLink<DataVector<xAOD::Jet_v1> >
```
Note the additional space in the on-disk field type name.

### ROOT version

`master` and `v6-36-00-patches`

### Installation method

from source

### Operating system

Linux

### Additional context

FYI @amete @Nowakus

hahnjo:
Fixes #19442

### DIFF
--- a/tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
@@ -823,6 +823,9 @@ public:
    /// In case of invalid field ID, an empty string is returned.
    std::string GetQualifiedFieldName(ROOT::DescriptorId_t fieldId) const;
 
+   /// Adjust the type name of the passed RFieldDescriptor for comparison with another renormalized type name.
+   std::string GetTypeNameForComparison(const RFieldDescriptor &fieldDesc) const;
+
    bool HasFeature(unsigned int flag) const { return fFeatureFlags.count(flag) > 0; }
    std::vector<std::uint64_t> GetFeatureFlags() const;
 
--- a/tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
@@ -657,6 +657,11 @@ private:
    //// All fields above are part of the schema and are cloned when creating a new descriptor from a given one
    //// (see CloneSchema())
 
+   std::uint16_t fVersionEpoch = 0; ///< Set by the descriptor builder when deserialized
+   std::uint16_t fVersionMajor = 0; ///< Set by the descriptor builder when deserialized
+   std::uint16_t fVersionMinor = 0; ///< Set by the descriptor builder when deserialized
+   std::uint16_t fVersionPatch = 0; ///< Set by the descriptor builder when deserialized
+
    std::uint64_t fOnDiskHeaderSize = 0;    ///< Set by the descriptor builder when deserialized
    std::uint64_t fOnDiskHeaderXxHash3 = 0; ///< Set by the descriptor builder when deserialized
    std::uint64_t fOnDiskFooterSize = 0; ///< Like fOnDiskHeaderSize, contains both cluster summaries and page locations
@@ -1528,6 +1533,7 @@ private:
 
 public:
    /// Checks whether invariants hold:
+   /// * RNTuple epoch is valid
    /// * RNTuple name is valid
    /// * Fields have valid parents
    /// * Number of columns is constant across column representations
@@ -1538,6 +1544,10 @@ public:
    /// Copies the "schema" part of `descriptor` into the builder's descriptor.
    /// This resets the builder's descriptor.
    void SetSchemaFromExisting(const RNTupleDescriptor &descriptor);
+
+   void SetVersion(std::uint16_t versionEpoch, std::uint16_t versionMajor, std::uint16_t versionMinor,
+                   std::uint16_t versionPatch);
+   void SetVersionForWriting();
 
    void SetNTuple(const std::string_view name, const std::string_view description);
    void SetFeature(unsigned int flag);
--- a/tree/ntuple/src/RFieldMeta.cxx
+++ b/tree/ntuple/src/RFieldMeta.cxx
@@ -429,11 +429,17 @@ void ROOT::RClassField::BeforeConnectPageSource(ROOT::Internal::RPageSource &pag
 
       rules = FindRules(&fieldDesc);
 
-      // If the field's type name is not the on-disk name but we found a rule, we know it is valid to read
-      // on-disk data because we found the rule according to the on-disk (source) type name and version/checksum.
-      if ((GetTypeName() != fieldDesc.GetTypeName()) && rules.empty()) {
-         throw RException(R__FAIL("incompatible type name for field " + GetFieldName() + ": " + GetTypeName() +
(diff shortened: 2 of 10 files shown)


hahnjo:
To be discussed if renormalization allows cases where we want to error; it seems fine so far, while asking `TClass` would fail the added test regarding default template arguments:
```diff
diff --git a/tree/ntuple/src/RFieldMeta.cxx b/tree/ntuple/src/RFieldMeta.cxx
index 361c798c034..178c000e79d 100644
--- a/tree/ntuple/src/RFieldMeta.cxx
+++ b/tree/ntuple/src/RFieldMeta.cxx
@@ -434,6 +434,9 @@ void ROOT::RClassField::BeforeConnectPageSource(ROOT::Internal::RPageSource &pag
          // Otherwise we require compatible type names, after renormalization (GetTypeName() is already renormalized).
          // However, we must not ask TClass because it might add (new) default template arguments.
          std::string descTypeName = GetRenormalizedTypeName(fieldDesc.GetTypeName());
+         if (auto *c = TClass::GetClass(descTypeName.c_str())) {
+            descTypeName = GetRenormalizedTypeName(c->GetName());
+         }
          if (GetTypeName() != descTypeName) {
             throw RException(R__FAIL("incompatible type name for field " + GetFieldName() + ": " + GetTypeName() +
                                      " vs. " + descTypeName));
```

pcanal:
```
< Reormalize on-disk type in RClassField
> Renormalize on-disk type in RClassField
```

hahnjo:
Here's an updated version of the changes that only renormalizes for spec versions before 1.0.0.1 (i.e. written with ROOT v6.34). Unfortunately, this will not work in all cases according to my quick tests: While ROOT v6.34 did some Meta normalization
```
root [0] auto field = ROOT::Experimental::RFieldBase::Create("v", "ROOT::Math::LorentzVector<ROOT::Math::PtEtaPhiM4D < double >  >").Unwrap()
(std::unique_ptr<ROOT::Experimental::RFieldBase, std::default_delete<ROOT::Experimental::RFieldBase> > &) std::unique_ptr -> 0x3a65900
root [1] field->GetTypeName()
(const std::string &) "ROOT::Math::LorentzVector<ROOT::Math::PtEtaPhiM4D<double> >"
```
it did not resolve typedefs:
```
root [0] auto field = ROOT::Experimental::RFieldBase::Create("v", "ROOT::Math::PtEtaPhiMVector").Unwrap()
(std::unique_ptr<ROOT::Experimental::RFieldBase, std::default_delete<ROOT::Experimental::RFieldBase> > &) std::unique_ptr -> 0x35ce980
root [1] field->GetTypeName()
(const std::string &) "ROOT::Math::PtEtaPhiMVector"
```
I think we cannot automatically fix the latter without being too lenient, for example if the type alias changed in the meantime... The only good news is that we know `TClass` handled the name during writing, so the string is somewhat valid.

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -657,6 +657,11 @@ private:
    //// All fields above are part of the schema and are cloned when creating a new descriptor from a given one
    //// (see CloneSchema())
 
+   std::uint16_t fVersionEpoch = 0; ///< Set by the descriptor builder when deserialized
+   std::uint16_t fVersionMajor = 0; ///< Set by the descriptor builder when deserialized
+   std::uint16_t fVersionMinor = 0; ///< Set by the descriptor builder when deserialized
+   std::uint16_t fVersionPatch = 0; ///< Set by the descriptor builder when deserialized
```
silverweed, line 663: We could also store these as a single `uint32_t` (dedicating 1 byte per field) to have only a single value to compare and juggle around. (Not a big deal either way)

hahnjo:
A corner case that we will actually handle incorrectly is integer type template arguments:
```
root [0] auto field = ROOT::Experimental::RFieldBase::Create("v", "ROOT::Math::LorentzVector<ROOT::Math::PtEtaPhiM4D<std::int64_t> >").Unwrap()
(std::unique_ptr<ROOT::Experimental::RFieldBase, std::default_delete<ROOT::Experimental::RFieldBase> > &) std::unique_ptr -> 0x37f5c60
root [1] field->GetTypeName()
(const std::string &) "ROOT::Math::LorentzVector<ROOT::Math::PtEtaPhiM4D<long> >"
```
During model reconstruction, current ROOT will (correctly) create a field of type `ROOT::Math::LorentzVector<ROOT::Math::PtEtaPhiM4D<std::int64_t>>` on Linux, but it will probably be `ROOT::Math::LorentzVector<ROOT::Math::PtEtaPhiM4D<std::int32_t>>` on Windows...

It's a bit different for `ROOT::Math::PtEtaPhiMVector`: there the typedef resolution will also happen during field creation, but then it will refuse to connect the page source because the renormalization will not resolve it.

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
hahnjo: I think I see the advantage of having `fVersionEpoch` in its own member, to check that it's actually been set, and then I don't think it's much more complicated to keep the others separate as well.

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -822,6 +822,9 @@ public:
    /// In case of invalid field ID, an empty string is returned.
    std::string GetQualifiedFieldName(ROOT::DescriptorId_t fieldId) const;
 
+   /// Adjust the passed typeName for comparison with another renormalized type name.
+   std::string GetTypeNameForComparison(const std::string &typeName) const;
```
jblomer: Perhaps pass a field descriptor here, to make sure the input type name comes from the right place.

jblomer:
> It's a bit different for `ROOT::Math::PtEtaPhiMVector`: there the typedef resolution will also happen during field creation, but then it will refuse to connect the page source because the renormalization will not resolve it.

Should we call `TClassEdit::GetNormalizedName()` first to resolve the typedef?

hahnjo:
> > It's a bit different for `ROOT::Math::PtEtaPhiMVector`: there the typedef resolution will also happen during field creation, but then it will refuse to connect the page source because the renormalization will not resolve it.
> 
> Should we call `TClassEdit::GetNormalizedName()` first to resolve the typedef?

I'm not sure if we want to, the typedef might have changed since writing the data... Yes, it means we will not be able to read such data without further user intervention, but that may be better than silently mistreating it.

hahnjo:
From an offline discussion: For the moment, we will not resolve typedefs and see how far we get. We've effectively lost the information what the actual type was during writing. If needed, we may allow reading back data with an explicit option.

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
hahnjo: Done

Issue #19419, opened 2025-07-21
[RF] FrequentistCalculator cannot handle non-parametric Pdfs


cburgard:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

When trying to use the FrequentistCalculator to conduct a hypothesis test on non-parametric Pdfs, a segfault is encountered.

MWE is attached.

### Reproducer

```
import ROOT

# Observable
x = ROOT.RooRealVar("x", "x", 0, 20)
x.setBins(40)

# Create two underlying shapes for H0 and H1
gauss_H0 = ROOT.RooGaussian("gauss_H0", "gauss_H0", x, ROOT.RooFit.RooConst(10), ROOT.RooFit.RooConst(2))
gauss_H1 = ROOT.RooGaussian("gauss_H1", "gauss_H1", x, ROOT.RooFit.RooConst(12), ROOT.RooFit.RooConst(2))

# Generate toy templates
data_H0 = gauss_H0.generate(ROOT.RooArgSet(x), 10000)
data_H1 = gauss_H1.generate(ROOT.RooArgSet(x), 10000)

# Binned templates (same binning!)
hist_H0 = data_H0.binnedClone()
hist_H1 = data_H1.binnedClone()

# Create RooHistPdfs
pdf_H0 = ROOT.RooHistPdf("pdf_H0", "pdf_H0", ROOT.RooArgSet(x), hist_H0)
pdf_H1 = ROOT.RooHistPdf("pdf_H1", "pdf_H1", ROOT.RooArgSet(x), hist_H1)

# Create pseudo-data from H1 (to simulate a test)
pseudo_data = pdf_H1.generateBinned(ROOT.RooArgSet(x), 1000)

# Dummy parameter (needed by ModelConfig)
dummy = ROOT.RooRealVar("dummy", "dummy", 0.0, -1.0, 1.0)
poi = ROOT.RooArgSet(dummy)

# ModelConfig for H0
mc_H0 = ROOT.RooStats.ModelConfig("H0")
mc_H0.SetPdf(pdf_H0)
mc_H0.SetParametersOfInterest(poi)

# ModelConfig for H1
mc_H1 = ROOT.RooStats.ModelConfig("H1")
mc_H1.SetPdf(pdf_H1)
mc_H1.SetParametersOfInterest(poi)

# Run FrequentistCalculator
fc = ROOT.RooStats.FrequentistCalculator(pseudo_data, mc_H0, mc_H1)
fc.SetToys(500, 500)

[...]

Issue #9490, opened 2022-01-01
[RF] RooMCStudy does not save generated parameters


goi42:
- [x] Checked for duplicates

### Describe the bug

First reported in the forum [here](https://root-forum.cern.ch/t/cannot-access-roomcstudy-genpardataset/46724), `keepGenData` is not sufficient to save generated parameters, despite the claims of the [documentation](https://root.cern.ch/doc/v624/classRooMCStudy.html#aa8654b8fcb5983d834c02489a35c9981). One must *also* specify an *internal* constraint.

### Expected behavior

`genParDataSet` should contain the values of the parameters used in dataset generation if `keepGenData=True` in `RooMCStudy::generate` (or `::run` or `::generateAndFit`).

### To Reproduce

`mcstudy.genParDataSet()` is empty despite setting `keepGenData=True`:
```python
import ROOT as r

ws = r.RooWorkspace("workspace")
x = ws.factory("x[-10, 10]")
pdf1 = ws.factory("Gaussian::pdf1(x, m[-1, 1], s[5, 10])")
pdf2 = ws.factory("Gaussian::pdf2(x, m, s)")
pdf = ws.factory("SUM::pdf(N1[0, 100] * pdf1, N2[0, 100] * pdf2)")

mcstudy = r.RooMCStudy(pdf, r.RooArgSet(x))
mcstudy.generate(1, 1000, True)

gen_ds = mcstudy.genParDataSet()
print("yes" if gen_ds else "no")
```

This is still the case if an external constraint is used:
```python
import ROOT as r

ws = r.RooWorkspace("workspace")
x = ws.factory("x[-10, 10]")
pdf1 = ws.factory("Gaussian::pdf1(x, m[-1, 1], s[5, 10])")
pdf2 = ws.factory("Gaussian::pdf2(x, m, s)")
pdf = ws.factory("SUM::pdf(N1[0, 100] * pdf1, N2[0, 100] * pdf2)")

constraint = ws.factory("Gaussian::constraint(s, cm[7], cs[1])")

mcstudy = r.RooMCStudy(pdf, r.RooArgSet(x), r.RooFit.ExternalConstraints(constraint))
mcstudy.generate(1, 1000, True)

gen_ds = mcstudy.genParDataSet()
print("yes" if gen_ds else "no")
```

This is not the case if an internal constraint is used:
```python
import ROOT as r

ws = r.RooWorkspace("workspace")
x = ws.factory("x[-10, 10]")
[...]

Issue #6949, opened 2020-12-14
RooMCStudyModule finalizeRun() bug


ggraziano98:
I initially reported this on the [root forum](https://root-forum.cern.ch/t/roodllsignificancemcsmodule-merge-error/42702)

### Describe the bug
RooAbsMCStudyModule::finalizeRun() returning a RooDataSet sometimes causes a merge error which is difficult to avoid.
Caused by RooMCStudy::_fitParData and RooAbsMCStudyModule data being different sizes

### Expected behavior
RooMCStudy and RooAbsMCStudyModule data merging correctly,
OR
A way to consistently make the datasets the same size

### To Reproduce
Run included macro

`

#include "RooRealVar.h"
#include "RooGaussian.h"
#include "RooConstVar.h"
#include "RooAddPdf.h"
#include "RooMCStudy.h"
#include "RooFFTConvPdf.h"
#include "RooArgusBG.h"
#include "RooDLLSignificanceMCSModule.h"

using namespace RooFit;
const double xmin = 4.3, xmax = 4.55, bratio = 0.3, m0 = 4.482;
const int nevents = 100;

void min_mcsign_error() {
	RooRealVar x("mresp", "x", 4.3, 4.55);

	// signal
	RooRealVar mean_res("#mean_{gauss}", "mean_res", 4.47);
	RooRealVar sigma_res("#sigma_{gauss}", "sigma_res", 0.05);
	RooGaussian sig("gs_res", "gs_res", x, mean_res, sigma_res);

	// background
	RooRealVar argpar("shape", "argus shape parameter", -7);
	RooRealVar power("power", "argus power parameter", 0.75);
	RooArgusBG bkg("bkg", Form("Background ARGUS, m_{0} = %.3f", m0), x, RooConst(m0), argpar, power);

	// generating distribution
	RooRealVar nsig("nsig", "signal number", bratio);
	RooRealVar nbg("nbg", "bg number", 1.-bratio); 
	RooAddPdf mcdist("distribution", "BW+a", RooArgList(sig, bkg), RooArgList(nsig, nbg));

	// signal used in model
	RooRealVar fitmean_res("#mean_{gauss}", "mean_res", 4.2, 4., 5.);
	RooRealVar fitsigma_res("#sigma_{gauss}", "sigma_res", 0.1, 0., 1.);
	RooGaussian fitsig("gs_res", "gs_res", x, fitmean_res, fitsigma_res);
	
	// background used in model
[...]

PR #19448, opened 2025-07-23
[RF][RS] Sync events per toy with observed data in HypoTest if needed

guitargeek:
For non-extended pdfs, the number of generated events in toy studies
should be synced with the number of events in the observed dataset.
Otherwise, the generation of the toy data will fail.

Closes https://github.com/root-project/root/issues/19419.


ferdymercury:
Thanks!! Does this also have some relation to https://github.com/root-project/root/issues/6949 and https://github.com/root-project/root/issues/9490?

lmoneta (approved):
LGTM!

PR #19450, opened 2025-07-23
[cppyy] fix: instantiation of templated functions

Vipul-Cariappa:
When instantiating with another templated function

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/TemplateProxy.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/TemplateProxy.cxx
@@ -142,6 +142,12 @@ PyObject* TemplateProxy::Instantiate(const std::string& fname,
             } else
                 PyErr_Clear();
 
+            if (!bArgSet && (Py_TYPE(itemi) == &TemplateProxy_Type)) {
+                TemplateProxy *tp = (TemplateProxy*)itemi;
+                PyObject *tmpl_name = CPyCppyy_PyText_FromFormat("decltype(%s%s)", tp->fTI->fCppName.c_str(), tp->fTemplateArgs ? CPyCppyy_PyText_AsString(tp->fTemplateArgs) : "");
+                PyTuple_SET_ITEM(tpArgs, i, tmpl_name);
+                bArgSet = true;
+            }
             if (!bArgSet) {
             // normal case (may well fail)
                 PyErr_Clear();
--- a/bindings/pyroot/cppyy/cppyy/test/test_templates.py
+++ b/bindings/pyroot/cppyy/cppyy/test/test_templates.py
@@ -1341,5 +1341,41 @@ class TestTEMPLATE_TYPE_REDUCTION:
         assert type(e1+e2) == cppyy.gbl.TypeReduction.Expr[int]
 
 
+class TestTEMPLATED_CALLBACK:
+    def setup_class(cls):
+        cls.test_dct = test_dct
+        import cppyy
+        cls.templates = cppyy.load_reflection_info(cls.test_dct)
+    
+    def test01_templated_callbacks(self):
+        import cppyy
+        from cppyy.gbl import std
+
+        cppyy.cppdef(r"""
+        bool foo() { return true; }
+
+        int bar(int a, int b) { return a + b; }
+
+        template <typename T, typename U>
+        T baz(T a, U b, std::string c) {
+            return (T)(a + b) + std::stoi(c);
+        }
+
+        template <typename F, typename R, typename... Args>
+        R returned_callback(F callable, R r, Args... args) {
+            return callable(r, std::forward<Args>(args)...);
+        }
+
+        template <typename F, typename... Args>
+        bool callback(F callable, Args&&... args) {
+            return callable(std::forward<Args>(args)...);
+        }
+        """)
+
+        assert cppyy.gbl.callback(cppyy.gbl.foo)
+        assert cppyy.gbl.returned_callback(cppyy.gbl.bar, 1, 1) == 2
+        assert cppyy.gbl.returned_callback(cppyy.gbl.baz[int, int], 1, 1, std.string("1")) == 3
+
+
 if __name__ == "__main__":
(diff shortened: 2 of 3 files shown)


Vipul-Cariappa:
@siliataider, @vepadulano, You might be interested in this.
This enables what we were discussing yesterday.

interpreter/cling/lib/Interpreter/LookupHelper.cpp
```diff
@@ -1691,7 +1691,7 @@ namespace cling {
         clang::QualType QT = clang::Sema::GetTypeFromParser(Res.get(), &TSI);
         QT = QT.getCanonicalType();
         {
-          ExprValueKind VK = VK_PRValue;
+          ExprValueKind VK = VK_XValue;
```
vgvassilev, line 1696: What's the reason for this change? Maybe add a comment clarifying how changing that changes the result of the routine.

dpiparo:
Very nice. @Vipul-Cariappa would you mind perhaps adding a test?

interpreter/cling/lib/Interpreter/LookupHelper.cpp
Vipul-Cariappa: We require the same change at CppInterOp [line](https://github.com/compiler-research/CppInterOp/blob/82f08c6570b7370ef8eaaab4a0240b2096df501b/lib/CppInterOp/CppInterOp.cpp#L1161).

Vipul-Cariappa:
Done.

bindings/pyroot/cppyy/cppyy/test/test_templates.py
```diff
@@ -1341,5 +1341,37 @@ def test01_reduce_binary(self):
...
+
+        void foo() { std::cout << "foo!\n";}
+
+        void bar(int a, float b) { std::cout << "bar(" << a << "," << b << ")\n"; }
+
+        template <typename T, typename U>
+        void baz(T a, U b, std::string c) { std::cout << "baz(" << a << "," << b << ",\"" << c << "\")\n"; }
+
+        template<typename F, typename... Args>
+        bool dataframe_define_mock(F callable, Args&&... args) {
```
vepadulano: I would give this function a more generic name, not to confuse it with a specific feature

bindings/pyroot/cppyy/cppyy/test/test_templates.py
```diff
@@ -1341,5 +1341,37 @@ def test01_reduce_binary(self):
...
+        #include <utility>
+        #include <iostream>
+        #include <string>
+
+        void foo() { std::cout << "foo!\n";}
+
+        void bar(int a, float b) { std::cout << "bar(" << a << "," << b << ")\n"; }
+
+        template <typename T, typename U>
+        void baz(T a, U b, std::string c) { std::cout << "baz(" << a << "," << b << ",\"" << c << "\")\n"; }
```
vepadulano: Since we've seen that in certain cases the function was called in a way that seemed correct but wasn't, I would actually return values from each of these functions, not using `std::cout`. The values can just be `std::tuple` of the input arguments returned as-is. The values should be checked on the Python side.

bindings/pyroot/cppyy/CPyCppyy/src/TemplateProxy.cxx
```diff
@@ -142,6 +142,12 @@ PyObject* TemplateProxy::Instantiate(const std::string& fname,
             } else
                 PyErr_Clear();
 
+            if (!bArgSet && Py_IS_TYPE(itemi, &TemplateProxy_Type)) {
```
vepadulano: I see the error
```
Py_IS_TYPE was not declared in this scope
```
On Alma8, maybe it's a missing macro from the Python version of that node?

vepadulano (changes requested):

interpreter/cling/lib/Interpreter/LookupHelper.cpp
vgvassilev: Good explanation and comment.

bindings/pyroot/cppyy/cppyy/test/test_templates.py
vepadulano: As discussed offline, since a std::tuple might incur in other issues, it's enough to return something like `T` with `T` being the integer passed as function argument.

vepadulano (approved):
LGTM, thanks!

bindings/pyroot/cppyy/CPyCppyy/src/TemplateProxy.cxx
```diff
@@ -142,6 +142,12 @@ PyObject* TemplateProxy::Instantiate(const std::string& fname,
             } else
                 PyErr_Clear();
 
+            if (!bArgSet && (Py_TYPE(itemi) == &TemplateProxy_Type)) {
+                TemplateProxy *tp = (TemplateProxy*)itemi;
+                PyObject *tmpl_name = CPyCppyy_PyText_FromFormat("decltype(%s%s)", tp->fTI->fCppName.c_str(), tp->fTemplateArgs ? CPyCppyy_PyText_AsString(tp->fTemplateArgs) : "");
```
vgvassilev, line 147: I realize I do not understand that. Are we parsing some strings to pipe to `decltype`?

Vipul-Cariappa: Yes. Because of the way type lookups work in cling, we need that `decltype`. In InterOp the `decltype` is not required. That is; here the lookup happens with a parse step, and that requires `decltype`, whereas in InterOp we can lookup the name/type directly and do not require the `decltype`.

vgvassilev: Can we avoid somehow the that parsing. Don't we have the Decl around?

vgvassilev: Ok, sorry, I am reviewing the wrong fork. Ignore my comment ;) We should fix that on the compiler-research end though...

Issue #19452, opened 2025-07-24
error: no member named 'hessian' in namespace 'clad'


Zehvogel:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Running the below reproducer results in
```
input_line_81:4:7: error: no member named 'hessian' in namespace 'clad'
clad::hessian(TFormula____id2776148674480555124, "p[0]" );
~~~~~~^
```
when using a build that is apparently without clad. Which is a bit annoying as I do not explicitly request the usage of clad anywhere...

### Reproducer

```python
import ROOT
from array import array

f = ROOT.TF1("f", "x[0] * [par_0]")
fit_fun = ROOT.Math.WrappedMultiTF1(f, f.GetNdim())

fit_data = ROOT.Fit.BinData(2, 1)
fit_data.Add(array("d", [0.]), 0.)
fit_data.Add(array("d", [1.]), 2.)

fitter = ROOT.Fit.Fitter()
fitter.SetFunction(fit_fun)
# fitter.SetFunction(fit_fun, False) # silences the error
fitter.Fit(fit_data)
```

### ROOT version

```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.36.000                       https://root.cern |
  | (c) 1995-2025, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Jun 07 2025, 14:20:17                 |
  | From tags/6-36-000@6-36-000                                      |
  | With g++ (Spack GCC) 14.2.0                                      |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------
```

### Installation method

source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh -r 2025-07-24

### Operating system

Alma9

### Additional context

Related: https://github.com/root-project/root/issues/15091

Message comes from here?
https://github.com/root-project/root/blob/aa184349a778d97709778349cc6a28d28fdf8a89/hist/hist/src/TFormula.cxx#L3356-L3358

PR #19454, opened 2025-07-24
[math] return early if clad not available

ferdymercury:
Fixes https://github.com/root-project/root/issues/19452

@Zehvogel would this help? Or you would get the error somewhere else?


Zehvogel:
> @Zehvogel would this help? Or you would get the error somewhere else?

I am not sure, I don't have the time to build a ROOT with this patch at the moment :/

PR #19455, opened 2025-07-24
[RF] Implement mechanism in RooMinimizer to freeze disconnected params

guitargeek:
This mimics what is already implemented in the CascadeMinimizer in CMS Combine.


lmoneta (approved):
LGTM!
Thank you Jonas for this useful improvement!

PR #19456, opened 2025-07-24
Add workflow to build Python wheels

vepadulano:
Now that the workflow is fully functional, see e.g. https://github.com/root-project/root/actions/runs/16493485486, I am reopening this PR.

The wheels can be installed on any Linux x86_64 node via:
```
$: pip install ROOT --index-url=https://root-experimental-python-wheels.web.cern.ch
```

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_rootcli.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_rootcli.py
@@ -1,0 +1,25 @@
+import os
+import subprocess
+import sys
+
+ROOT_HOME = os.path.dirname(__file__)
+
+
+def main() -> None:
+    # Find the ROOT executable from the current installation directory
+    bindir = os.path.join(ROOT_HOME, "bin")
+    rootexe = os.path.join(bindir, "root.exe")
+    if not os.path.exists(rootexe):
+        raise FileNotFoundError(
+            f"Could not find 'root.exe' executable in directory '{bindir}'. "
+            "Something is wrong in the ROOT installation."
+        )
+    # Make sure command line arguments are preserved
+    args = [rootexe] + sys.argv[1:]
+    # Run the actual ROOT executable and return the exit code to the main Python process
+    out = subprocess.run(args)
+    return out.returncode
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/cmake/modules/SearchInstalledSoftware.cmake
+++ b/cmake/modules/SearchInstalledSoftware.cmake
@@ -676,7 +676,11 @@ set(Python3_FIND_FRAMEWORK LAST)
 
 # Even if we don't build PyROOT, one still need python executable to run some scripts
 list(APPEND python_components Interpreter)
-if(pyroot OR tmva-pymva)
+if(pyroot AND NOT (tpython OR tmva-pymva))
+  # We have to only look for the Python development module in order to be able to build ROOT with a pip backend
+  # In particular, it is forbidden to link against libPython.so, see https://peps.python.org/pep-0513/#libpythonx-y-so-1
+  list(APPEND python_components Development.Module)
+elseif(tpython OR tmva-pymva)
   list(APPEND python_components Development)
 endif()
 if(tmva-pymva)
@@ -1759,7 +1763,7 @@ endif(tmva)
 #---Check for PyROOT---------------------------------------------------------------------
 if(pyroot)
 
-  if(Python3_Development_FOUND)
+  if(Python3_Development.Module_FOUND)
     message(STATUS "PyROOT: development package found. Building for version ${Python3_VERSION}")
   else()
     if(fail-on-missing)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,30 @@
+[build-system]
+requires = ["cmake", "setuptools<72", "wheel", "numpy"]
+
+[project]
+name = "ROOT"
+version = "0.1a8"
+requires-python = ">=3.8"
+maintainers = [
+  {name = "Vincenzo Eduardo Padulano", email = "vincenzo.eduardo.padulano@cern.ch"}
+]
+description = "An open-source data analysis framework used by high energy physics and others."
+readme = "README.md"
+license = {file = "LICENSE"}
+dependencies = [
+  "numpy",
+]
+
+# Demonstrate how to expose ROOT CLI tools the Python way. For now only include
+# `root`, other tools can be added later using the same approach.
+[project.scripts]
+root = "ROOT._rootcli:main"
+
+[tool.cibuildwheel]
+# Increase pip debugging output
+build-verbosity = 1
+manylinux-x86_64-image = "manylinux_2_28"
+
+# Install system libraries
+[tool.cibuildwheel.linux]
+before-all = "dnf install -y epel-release && /usr/bin/crb enable && dnf install -y openssl-devel libX11-devel libXpm-devel libXft-devel libXext-devel libuuid-devel libjpeg-devel giflib-devel"
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,138 @@
+"""
+setuptools-based build of ROOT.
+
+This script uses setuptools API to steer a custom CMake build of ROOT. All build
+configuration options are specified in the class responsible for building. A
+custom extension module is injected in the setuptools setup to properly generate
+the wheel with CPython extension metadata. Note that ROOT is first installed via
+CMake into a temporary directory, then the ROOT installation artifacts are moved
+to the final Python environment installation path, which often starts at
+${ENV_PREFIX}/lib/pythonXX.YY/site-packages, before being packaged as a wheel.
+"""
+
+import os
+import pathlib
+import shlex
+import subprocess
+import tempfile
+
+from setuptools import Extension, find_packages, setup
+from setuptools.command.build import build as _build
+from setuptools.command.install import install as _install
+
+# Get the long description from the README file
+SOURCE_DIR = pathlib.Path(__file__).parent.resolve()
+LONG_DESCRIPTION = (SOURCE_DIR / "README.md").read_text(encoding="utf-8")
+
+BUILD_DIR = tempfile.mkdtemp()
+INSTALL_DIR = tempfile.mkdtemp()
+
+# Name given to an internal directory within the build directory
+# used to mimick the structure of the target installation directory
+# in the user Python environment, usually named "site-packages"
+ROOT_BUILD_INTERNAL_DIRNAME = "mock_site_packages"
+
+
+class ROOTBuild(_build):
+    def run(self):
+        _build.run(self)
+
+        # Configure ROOT build
+        configure_command = shlex.split(
+            "cmake "
+            # gminimal=ON enables only a minimal set of components (cling+core+I/O+graphics)
+            "-Dgminimal=ON -Dasimage=ON -Dopengl=OFF "
+            "-Druntime_cxxmodules=ON -Drpath=ON -Dfail-on-missing=ON "  # Generic build configuration
+            # Explicitly turned off components, even though they are already off because of gminimal, we want to keep
+            # them listed here for documentation purposes:
+            # - tmva-pymva, tpython: these components link against libPython, forbidden for manylinux compatibility,
+            #   see https://peps.python.org/pep-0513/#libpythonx-y-so-1
+            # - thisroot_scripts: the thisroot.* scripts are broken if CMAKE_INSTALL_PYTHONDIR!=CMAKE_INSTALL_LIBDIR
+            "-Dtmva-pymva=OFF -Dtpython=OFF -Dthisroot_scripts=OFF "
+            "-Dbuiltin_nlohmannjson=ON -Dbuiltin_tbb=ON -Dbuiltin_xrootd=ON "  # builtins
+            "-Dbuiltin_lz4=ON -Dbuiltin_lzma=ON -Dbuiltin_zstd=ON -Dbuiltin_xxhash=ON "  # builtins
+            "-Dpyroot=ON -Ddataframe=ON -Dxrootd=ON -Dssl=ON -Dimt=ON "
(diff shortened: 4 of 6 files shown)


bindings/pyroot/pythonizations/python/ROOT/_rootcli.py
```diff
@@ -0,0 +1,26 @@
+import os
+import subprocess
+import sys
+
+ROOT_HOME = os.path.dirname(__file__)
+
+
+def main() -> None:
+    # Ensure all ROOT libraries are found at runtime by the process
+    os.environ['LD_LIBRARY_PATH'] = os.path.join(ROOT_HOME, 'lib')
```
guitargeek:
```suggestion
```
Is this really needed? You build with `rpath=ON`, doesn't the binary contain the relative rpath to the library directory then?

bindings/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ if(pyroot)
   endif()
 endif()
 
-if(tpython)
+if(tpython AND NOT (DEFINED ROOT_WHEEL_BUILD AND ROOT_WHEEL_BUILD))
```
guitargeek:
```suggestion
if(tpython)
```
Why not just do `tpython=OFF` in the setup.py?

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -1774,7 +1780,7 @@ if(pyroot)
 endif()
 
 #---Check for TPython---------------------------------------------------------------------
-if(tpython)
+if(tpython AND NOT (DEFINED ROOT_WHEEL_BUILD AND ROOT_WHEEL_BUILD))
```
guitargeek:
```suggestion
if(tpython)
```
Same here.

bindings/pyroot/pythonizations/python/ROOT/__init__.py
```diff
@@ -8,9 +8,29 @@
...
+
+import platform
+if platform.system() != "Windows":
+    try:
+        # Mimic what cling does to find standard headers include path
+        cmd = shlex.split('c++ -xc++ -E -v /dev/null')
+        env = os.environ
+        env.update({'LC_ALL': 'C'})
+        subprocess.run(cmd, env=env, check=True,
+                    stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
guitargeek: Is there no way to avoid unconditionally doing this potentially expensive `subprocess.run` on `import ROOT`?

bindings/pyroot/pythonizations/python/ROOT/_facade.py
```diff
@@ -11,7 +11,6 @@
 from ._numbadeclare import _NumbaDeclareDecorator
 from ._pythonization import pythonization
 
-
```
guitargeek: I wouldn't do these formatting changes if the file is not touched already.

cmake/modules/SearchInstalledSoftware.cmake
```diff
@@ -677,7 +677,13 @@ set(Python3_FIND_FRAMEWORK LAST)
 # Even if we don't build PyROOT, one still need python executable to run some scripts
 list(APPEND python_components Interpreter)
 if(pyroot OR tmva-pymva)
-  list(APPEND python_components Development)
+  if(DEFINED ROOT_WHEEL_BUILD AND ROOT_WHEEL_BUILD)
```
guitargeek: If I understand correctly, we only need the full `Development` component when building `tpython` or `tmva-pymva`.

Maybe this logic works then in general?

```python
if(tpython OR tmva-pymva)
  list(APPEND python_components Development)
elseif(pyroot)
  list(APPEND python_components Development.Module)
endif()
```

Then there is no need for a code branch specific to `ROOT_WHEEL_BUILD`.

setup.py
```diff
@@ -0,0 +1,143 @@
...
+
+class ROOTBuild(_build):
+    def run(self):
+        _build.run(self)
+
+        # Configure ROOT build
+        configure_command = shlex.split(
+            "cmake -GNinja -Dccache=ON "
+            "-Dgminimal=ON -Dasimage=ON -Dopengl=OFF "  # Graphics
+            "-Druntime_cxxmodules=ON -Drpath=ON -Dfail-on-missing=ON -DROOT_WHEEL_BUILD=ON "  # Generic build configuration
```
guitargeek:
```suggestion
            "-Druntime_cxxmodules=ON -Drpath=ON -Dfail-on-missing=ON "  # Generic build configuration
```
Maybe with my suggestions, this is not needed anymore?

setup.py
```diff
@@ -0,0 +1,143 @@
...
+        # leftover directories that still need to be copied
+        self.copy_tree(os.path.join(INSTALL_DIR, "cmake"), os.path.join(root_package_dir, "cmake"))
+        self.copy_tree(os.path.join(INSTALL_DIR, "config"), os.path.join(root_package_dir, "config"))
+        self.copy_tree(os.path.join(INSTALL_DIR, "etc"), os.path.join(root_package_dir, "etc"))
+        self.copy_tree(os.path.join(INSTALL_DIR, "fonts"), os.path.join(root_package_dir, "fonts"))
+        self.copy_tree(os.path.join(INSTALL_DIR, "icons"), os.path.join(root_package_dir, "icons"))
+        self.copy_tree(os.path.join(INSTALL_DIR, "macros"), os.path.join(root_package_dir, "macros"))
+        self.copy_tree(os.path.join(INSTALL_DIR, "man"), os.path.join(root_package_dir, "man"))
+        self.copy_tree(os.path.join(INSTALL_DIR, "README"), os.path.join(root_package_dir, "README"))
+        self.copy_tree(os.path.join(INSTALL_DIR, "tutorials"), os.path.join(root_package_dir, "tutorials"))
```
guitargeek, line 102: The tutorials directory is quite large (21M according to `du -h`). Do we need to include it in the wheel? Can we maybe just delete it?

setup.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # Copy ROOT installation tree to the ROOT package directory in the pip installation path
+        self.copy_tree(os.path.join(INSTALL_DIR, ROOT_BUILD_INTERNAL_DIRNAME), install_path)
+
+        root_package_dir = os.path.join(install_path, "ROOT")
+
+        # After the copy of the "mock" package structure from the ROOT installations, these are the
+        # leftover directories that still need to be copied
+        self.copy_tree(os.path.join(INSTALL_DIR, "cmake"), os.path.join(root_package_dir, "cmake"))
+        self.copy_tree(os.path.join(INSTALL_DIR, "config"), os.path.join(root_package_dir, "config"))
```
guitargeek: The `config` directory contains files that are only relevant when building a project with Makefile against ROOT. Probably that's not needed for a Python wheel (actually I planned to remove this directory from the install tree completely. For `gnuinstall=ON` it's already gone).

setup.py
vepadulano: I think this is interesting. There are two axis of discussion to follow.

First: On the one hand, it's fair to say that people will not use most of those tutorials. On the other hand, I believe it's also nice to provide people with at least some code to check that the ROOT installation is working. So maybe we can just package one or two tutorials? I don't know how to choose though.

Second: I was thinking to use the tutorials as a way to test the wheel as it is installed, so they will need to be available in the wheel. Without this, I don't know what to use to setup a testing environment for the wheel.

setup.py
vepadulano: Ok fine I can remove the `config` directory

setup.py
guitargeek: Right, ok I see, so at least for now we should keep them for the tests :+1:

cmake/modules/SearchInstalledSoftware.cmake
vepadulano: I changed the logic slightly but I agree

bindings/pyroot/pythonizations/python/ROOT/__init__.py
vepadulano: Good question, I don't know of a better way to check this, we can decide to remove it for now, but it's dampening a real problem that some people might incur, i.e. a missing compiler/C++ standard library on their system

guitargeek: Ok if there is nothing else that comes to mind right now we can also keep it. After all, the overhead in `import ROOT` is already huge compared to one `subprocess.run` of the C++ compiler with no inputs

vepadulano:
Thank you @guitargeek ! I addressed all the review comments, in particular:

* I removed the extra `subprocess.run` call in `__init__.py` for the moment. It was meant to demonstrate a way to cope with missing dependencies, but probably we need to have a larger discussion about it
* I kept for now the `tutorials` folder as I believe we can use it as a launcher for tests of the wheel, and the extra 25MB at the moment are not a real issue.

guitargeek (approved):
Thanks for addressing the comments so quickly, it looks very good now!

Can we just do one CI run with the "clean build" flag please, since you changed "SearchInstalledSoftware.cmake"? And I guess you will also squash the commits a bit :smile:

setup.py
```diff
@@ -0,0 +1,141 @@
...
+installation step produces the following output target install directory:
+
+```
+site-packages/
+-- cppyy/
+-- cppyy_backend/
+-- libcppyy.so
+-- libcppyy_backend.so
+-- ROOT/
+-- -- libROOTPythonizations.so
```
guitargeek:
```suggestion
site-packages/
-- cppyy/
-- libcppyy.so
-- ROOT/
-- -- libROOTPythonizations.so
-- -- __init__.py (and other Python sources)
-- -- _pythonization/
```
Just a small nitpick: the `ROOT/` directory contains more than `libROOTPythonizations.so`.

Also, if we would decide to go ahead with #19470, the backend is gone.

vepadulano:
Squashed all changes into a single commit. First commit currently is taken as-is from https://github.com/root-project/root/pull/19465

setup.py
```diff
@@ -0,0 +1,132 @@
...
+ROOT_BUILD_INTERNAL_DIRNAME = "mock_site_packages"
+
+
+class ROOTBuild(_build):
+    def run(self):
+        _build.run(self)
+
+        # Configure ROOT build
+        configure_command = shlex.split(
+            "cmake -GNinja -Dccache=ON "
```
guitargeek:
```suggestion
            "cmake "
```
Now that we have a `setup.py`, people will come and call `pip install root_src/`. Is it really necessary to use `Ninja` and `ccache` in general? These packages are probably not installed anywhere.

I'm also fine keeping this, but it's just something to consider.

guitargeek (approved):
Thanks a lot for the PR! Works great, also locally for me with `pip install root_src/`!

setup.py
```diff
@@ -0,0 +1,132 @@
...
+            "-Dgminimal=ON -Dasimage=ON -Dopengl=OFF "  # Graphics
+            "-Druntime_cxxmodules=ON -Drpath=ON -Dfail-on-missing=ON "  # Generic build configuration
+            "-Dtmva-pymva=OFF -Dtpython=OFF "  # Turn off explicitly components that link against libPython
+            "-Dbuiltin_nlohmannjson=ON -Dbuiltin_tbb=ON -Dbuiltin_xrootd=ON "  # builtins
+            "-Dbuiltin_lz4=ON -Dbuiltin_lzma=ON -Dbuiltin_zstd=ON -Dbuiltin_xxhash=ON "  # builtins
+            "-Dpyroot=ON -Ddataframe=ON -Dxrootd=ON -Dssl=ON -Dimt=ON "
+            "-Droofit=ON "
```
guitargeek, line 55:
```suggestion
            "-Droofit=ON -Dthisroot_scripts=OFF "
```
Something I remembered now: the `thisroot_scripts` are broken if the PYTHONDIR is not the LIBDIR, and I don't think they are necessary for the wheels.

pyproject.toml
```diff
@@ -0,0 +1,30 @@
+[build-system]
+requires = ["cmake", "setuptools<72", "wheel", "ninja", "numpy"]
```
guitargeek: Oh, I see `ninja` will always be there. But what about `ccache`?

pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+root = "ROOT._rootcli:main"
+
+[tool.cibuildwheel]
+# Increase pip debugging output
+build-verbosity = 1
+manylinux-x86_64-image = "manylinux_2_28"
+
+# Install system libraries
+[tool.cibuildwheel.linux]
+before-all = "dnf install -y epel-release && /usr/bin/crb enable && dnf install -y openssl-devel libX11-devel libXpm-devel libXft-devel libXext-devel libuuid-devel libjpeg-devel giflib-devel"
```
guitargeek, line 30: Also here I don't see `ccache`. Is the `ccache` actually available and active?

pyproject.toml
```diff
@@ -0,0 +1,30 @@
+[build-system]
+requires = ["cmake", "setuptools<72", "wheel", "ninja", "numpy"]
+
+[project]
+name = "ROOT"
+version = "0.1a7"
```
guitargeek: Is that still a placeholder?

setup.py
vepadulano: They are fortunately already off by default thanks to `-Dgminimal=ON`. But I can mark them explicitly off with a comment for our own documentation

setup.py
vepadulano: You are right actually, thanks for making me notice. These were leftover from developing on my machine, but they should be removed.

pyproject.toml
vepadulano: Not really a placeholder. Until the Python wheels reach a stable status, it's best to release them tagging a provisional alpha version of ROOT, even though it's clearly connected to a real underlying version of the software itself. I believe by the time the wheels will be built for an official release of ROOT, we will update also the version of the wheel package

guitargeek (approved):
Thanks for addressing the comments!

PR #19458, opened 2025-07-24
[hist] Add first microbenchmark for `RRegularAxis`

hahnjo:
In ROOT, we traditionally had benchmarks separated in the `rootbench` repository. However, for the new `histv7` package, they should evolve together with the code. Therefore this commit introduces a `benchmark` subdirectory, which is only built with the "hidden" configuration flag `histv7_benchmark`. The standalone build exposes an option under the same name.


bellenot (approved):

silverweed:
Should the option be off by default? After all one only wants to enable it when specifically running the benchmarks, which is a very rare occasion in general

hahnjo:
You mean for the standalone build? It's already "weak on" and will skip building the benchmarks when the dependency library is not found. I could turn it fully off by default, that would maybe lose a bit of build coverage...

I was also considering renaming the option to `histv7_benchmark` for consistency with the ROOT build. What do you think?

jblomer (approved):

PR #19459, opened 2025-07-24
[RF] Compress the embedded RooDataHists in workspaces when serializing

guitargeek:
Significantly reduce the size of RooWorkspaces with template histograms on disk.

Should circumvent the 1 GB buffer limit that ATLAS is hitting in the Higgs combination.

### DIFF
--- a/roofit/roofitcore/inc/LinkDef.h
+++ b/roofit/roofitcore/inc/LinkDef.h
@@ -259,6 +259,7 @@
 #pragma link C++ class RooWorkspace- ;
 #pragma link C++ class RooWorkspace::CodeRepo- ;
 #pragma link C++ class RooWorkspace::WSDir+ ;
+#pragma link C++ class RooWorkspace::EmbeddedHisto+ ;
 #pragma link C++ class RooWorkspaceHandle+;
 #pragma link C++ class std::list<TObject*>+ ;
 #pragma link C++ class std::list<RooAbsData*>+ ;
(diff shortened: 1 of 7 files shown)


chenzhl2018:
Hi @guitargeek, thanks for your nice development. Recently, i try to test this development with 12 workspaces of ATLAS Higgs analysis. There are some improvements for some workspaces in disk size, while some failures for some workspaces.  I would like to ask what causes the failure and how to fix it. Thanks a lot.
图片
图片

guitargeek:
Great, thanks for trying it out! Do you also have the comparison for the uncompressed sizes, like reported in the meeting 3 weeks ago?

The errors are because I have not implemented support for template histograms with asymmetric uncertainties, because I thought nobody would do this in practice. Interesting that you have this, because it's not clear at all how to use this in a statistical model :) Maybe it's just an artifact of the way you build the histograms? You can then also reduce the workspace size by removing these asymmetric errors.

In any case, I'll implement support for that in this PR tomorrow, and then you can try again.

chenzhl2018:
Hi @guitargeek, thanks for your quick reply. Yeah, i check the Htautau workspace quickly, and we could find the buffer size decreases largely with the new development. And then i will fit the workspaces and check the workspaces could really work or not. It's good to see the great development.

And i am also curious about the reason  this development could reduce the buffer size. I will appreciate it if you could explain this for me. Thanks a lot.

图片

chenzhl2018:
Hi @guitargeek, i would like to the issue of not implementation for template histograms with asymmetric uncertainties. I think the asymmetric uncertainties of template histograms are the statistical uncertainties of the histos, right? If so, i would like to ask if it's possible to provide one option to remove it or skip this error?

chenzhl2018:
For the channels with successful implementation, there are about 15% buffer size reduction. 
图片

guitargeek:
I added some commits to address comments that I got by @chenzhl2018 via mail.

PR #19460, opened 2025-07-25
[ntuple] Account for aliased pages in on-disk size calculation

enirolf:
For RNTuples that have been written with same-page merging enabled, the pages that were aliased (i.e., repeated from a the previous identical page) should not be considered in the calculation of the total RNTuple size.


tree/ntuple/src/RNTupleDescriptorFmt.cxx
```diff
@@ -124,7 +124,14 @@ void ROOT::RNTupleDescriptor::PrintInfo(std::ostream &output) const
          }
          const auto &pageRange = cluster.second.GetPageRange(column.second.GetPhysicalId());
          auto idx = cluster2Idx[cluster.first];
+         std::uint64_t prevLocatorOffset = 0;
+         std::uint64_t locatorOffset;
          for (const auto &page : pageRange.GetPageInfos()) {
+            locatorOffset = page.GetLocator().GetType() == ROOT::RNTupleLocator::ELocatorType::kTypeDAOS
+                               ? page.GetLocator().GetPosition<RNTupleLocatorObject64>().GetLocation()
+                               : page.GetLocator().GetPosition<std::uint64_t>();
+            if (locatorOffset == prevLocatorOffset)
```
silverweed: I'm pretty sure that `pageRange.GetPageInfos()` is not guaranteed to return the pages in sorted order, so you might miss some duplicates here if they are not adjacent. Usually they will be but it's not a guarantee (and I observed this in the rntviewer)

silverweed: Or rather, the entire set of pages returned by all the calls to `GetPageInfos()` is not necessarily sorted. That said, I'm not sure this has implication for detecting duplicate pages...

hahnjo: I think it has, we are also deduplicating pages between page ranges of different columns. This might even be the primary case, for example for two index columns of two collections that always have the same number of elements.

tree/ntuple/src/RNTupleDescriptorFmt.cxx
```diff
@@ -124,7 +124,14 @@ void ROOT::RNTupleDescriptor::PrintInfo(std::ostream &output) const
...
+               continue;
             nBytesOnStorage += page.GetLocator().GetNBytesOnStorage();
             nBytesInMemory += page.GetNElements() * elementSize;
```
hahnjo: Aliased pages should probably be counted towards `nBytesInMemory` because (at least for the moment) we actually load and decompress them multiple times

enirolf: Here `nBytesInMemory` is only used to calculate and print the compression factor, so I think in this specific place it should not be counted actually. However, this is a good point to take into account for the size counting taking place in the inspector, I will change it there.

tree/ntupleutil/v7/src/RNTupleInspector.cxx
```diff
@@ -60,6 +60,7 @@ void ROOT::Experimental::RNTupleInspector::CollectColumnInfo()
       // to report the size _in memory_ of column elements.
       std::uint32_t elemSize = RColumnElementBase::Generate(colDesc.GetType())->GetSize();
       std::uint64_t nElems = 0;
+      std::unordered_set<std::uint64_t> seenPages{};
```
jblomer, line 63: I think you can move the map down into the first level of the for loop because we don't deduplicate pages across clusters. This will also limit the memory used for the map.

tree/ntuple/src/RNTupleDescriptorFmt.cxx
```diff
@@ -124,15 +125,22 @@ void ROOT::RNTupleDescriptor::PrintInfo(std::ostream &output) const
...
-            info.fNBytesOnStorage += page.GetLocator().GetNBytesOnStorage();
-            ++info.fNPages;
-            ++nPages;
+            locatorOffset = page.GetLocator().GetType() == ROOT::RNTupleLocator::ELocatorType::kTypeDAOS
+                               ? page.GetLocator().GetPosition<RNTupleLocatorObject64>().GetLocation()
+                               : page.GetLocator().GetPosition<std::uint64_t>();
+            auto [_, pageAdded] = seenPages.emplace(locatorOffset);
+            if (pageAdded) {
+               nBytesOnStorage += page.GetLocator().GetNBytesOnStorage();
+               nBytesInMemory += page.GetNElements() * elementSize;
```
jblomer, line 139: I think for the in-memory summary, we should not take into account the page deduplication (because we don't deduplicate in memory).

tree/ntuple/src/RNTupleDescriptorFmt.cxx
```diff
@@ -124,15 +125,22 @@ void ROOT::RNTupleDescriptor::PrintInfo(std::ostream &output) const
...
+               clusters[idx].fNBytesOnStorage += page.GetLocator().GetNBytesOnStorage();
+               clusters[idx].fNBytesInMemory += page.GetNElements() * elementSize;
```
jblomer, line 141: Same here, I think I'd always sum up the in-memory size

tree/ntuple/src/RNTupleDescriptorFmt.cxx
```diff
@@ -124,15 +125,22 @@ void ROOT::RNTupleDescriptor::PrintInfo(std::ostream &output) const
...
+               ++clusters[idx].fNPages;
+               info.fNBytesOnStorage += page.GetLocator().GetNBytesOnStorage();
```
jblomer, line 143: Here it gets tricky. As is, deduplicated pages count for none of the columns. I'd rather error the other way and account for all of the pages (aliased or not) for the _per column_ statistics. If we want to do it correctly, I think we need to keep a `seenPages` map per column and then count deduplicated pages only once. The meaning of this would be "if I only stored this column and nothing else on disk, this is what it takes".

PR #19462, opened 2025-07-25
[DF] Add `signed char` to `TypeID2TypeName`

enirolf:
Discovered when trying to read an RNTuple `std::int8_t`-type field, which is typedeffed as a `signed char`.


vepadulano (approved):
Thanks!

PR #19463, opened 2025-07-25
Fix finding R in CMake on Windows if old (side-by-side) version is present

Blake-Madden:
On Windows, if an old installation of R is present (which no longer has the executable tools in it) which is next to a newer one, then skip it and look in the next R subdirectory. Before, the loop would quit after looking in the first folder.


cmake/modules/FindR.cmake
```diff
@@ -54,7 +54,9 @@ if(WIN32 AND R_EXECUTABLE MATCHES "R_EXECUTABLE-NOTFOUND")
         if(EXISTS "${RX64PATH}\\${R_VERSIONED_SUBDIR}\\bin\\x64\\Rscript.exe")
           set(R_SCRIPT "${RX64PATH}\\${R_VERSIONED_SUBDIR}\\bin\\x64\\Rscript.exe")
         endif()
-        break()
+        if (R_EXECUTABLE AND R_SCRIPT)
+          break()
+        endif()
```
pcanal, line 59:
```suggestion
        if (R_EXECUTABLE AND R_SCRIPT)
          break()
        endif()
        unset(R_EXECUTABLE)
        unset(R_SCRIPT)
```
otherwise, we can end up with the executable coming from one installation and the script from another.

cmake/modules/FindR.cmake
```diff
@@ -69,7 +71,9 @@ if(WIN32 AND R_EXECUTABLE MATCHES "R_EXECUTABLE-NOTFOUND")
         if(EXISTS "${RX86PATH}\\${R_VERSIONED_SUBDIR}\\bin\\x86\\Rscript.exe")
           set(R_SCRIPT "${RX86PATH}\\${R_VERSIONED_SUBDIR}\\bin\\x86\\Rscript.exe")
         endif()
-        break()
+        if (R_EXECUTABLE AND R_SCRIPT)
+          break()
+        endif()
```
pcanal, line 78:
```suggestion
        if (R_EXECUTABLE AND R_SCRIPT)
          break()
        endif()
        unset(R_EXECUTABLE)
        unset(R_SCRIPT)
```
otherwise, we can end up with the executable coming from one installation and the script from another.

PR #19464, opened 2025-07-25
[df] Remove RDatasetSpec header dependency on TTree

vepadulano:
RDatasetSpec depended on RFriendInfo.hxx which transitively includes TTree.h. This commit proposes to remove this dependency by changing the type of the data member that stores information about friend trees as a std::any and cast it when appropriate. In the future, when introducing support for RNTuple joins, this can be extended, changed to std::variant without the need for the extra dependency on TTree.

### DIFF
--- a/tree/dataframe/inc/ROOT/RDF/RDatasetSpec.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/RDatasetSpec.hxx
@@ -11,14 +11,18 @@
 #ifndef ROOT_RDF_RDATASETSPEC
 #define ROOT_RDF_RDATASETSPEC
 
+#include <any>
 #include <limits>
 #include <string>
 #include <utility> // std::pair
 #include <vector>
 
 #include <ROOT/RDF/RSample.hxx>
-#include <ROOT/RFriendInfo.hxx>
 #include <RtypesCore.h> // Long64_t
+
+namespace ROOT::TreeUtils {
+struct RFriendInfo;
+}
 
 namespace ROOT {
 namespace Detail {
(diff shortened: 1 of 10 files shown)


tree/dataframe/inc/ROOT/RDF/RDatasetSpec.hxx
```diff
@@ -75,12 +79,12 @@ public:
 
 private:
    std::vector<RSample> fSamples;             ///< List of samples
-   ROOT::TreeUtils::RFriendInfo fFriendInfo;  ///< List of friends
+   std::any fFriendInfo;  ///< List of friends
```
pcanal, line 82: if we already know it will become an `std::variant`, couldn't we/shouldn't we just use it already (with a single argument)?

vepadulano: To be honest, I don't know if it will become a std::variant. If you like that type better, I can change it. In any case we'll never be able to use the compile-time information since by definition the dataset type can only be known at runtime

vepadulano: Actually thinking about it again, I believe `std::variant` is just a worse alternative in this case since even if we wanted to introduce it and have two variants for TTree and RNTuple we would still need to only forward-declare them and then play the trick of defining the special member methods only in the implementation file. So overall as things currently stand I still like the `std::any` option better, unless you have a clearer idea on why the variant would be better.

martamaja10 (approved):
Thanks Vincenzo!

PR #19465, opened 2025-07-25
[Python] Restructure Python libraries to not require system links

guitargeek:
Restructure Python libraries to not require system links.


hahnjo:
btw "symlink" stands for "symbolic link", not "system" :wink:

guitargeek:
Sorry! Fortunately I can't repeat this mistake because there are no `SYMBOLIC` links left in our CMake code base

PR #19466, opened 2025-07-28
[ntuple] Proper classic browser support

jblomer:
Follow-up of #18889 

Add support for plotting histograms of field data in the classic browser.

For both classic browser and web browser remove intermediate, internal fields from the tree view. E.g., for a vector of float, it will directly histogram the data and not require the user to open the `_0` collection content field.

The result looks like this:

image


linev (approved):
For `gui/browsable` code looks fine.
For classical TBrowser I would wait ok from @bellenot

bellenot (approved):
LGTM, thanks!

PR #19467, opened 2025-07-28
[ntuple] Change GetColumn methods in Inspector

magnustymoteus:
# This Pull request:

## Changes or fixes:
- Renamed `GetColumnsByFieldId` to `GetAllColumnsOfField`
- Added unit tests for `GetAllColumnsOfField` because it also recently became a public method

## Checklist:

- [X] tested changes locally

### DIFF
--- a/tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
+++ b/tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
@@ -277,7 +277,7 @@ public:
    /// \param [in] fieldId The ID of the field for which to collect the columns.
    ///
    /// \return A vector containing the IDs of all columns for the provided field ID.
-   std::vector<ROOT::DescriptorId_t> GetColumnsByFieldId(ROOT::DescriptorId_t fieldId) const;
+   std::vector<ROOT::DescriptorId_t> GetAllColumnsOfField(ROOT::DescriptorId_t fieldId) const;
 
    /////////////////////////////////////////////////////////////////////////////
    /// \brief Get all column types present in the RNTuple being inspected.
--- a/tree/ntupleutil/v7/src/RNTupleInspector.cxx
+++ b/tree/ntupleutil/v7/src/RNTupleInspector.cxx
@@ -127,7 +127,7 @@ ROOT::Experimental::RNTupleInspector::CollectFieldTreeInfo(ROOT::DescriptorId_t 
 }
 
 std::vector<ROOT::DescriptorId_t>
-ROOT::Experimental::RNTupleInspector::GetColumnsByFieldId(ROOT::DescriptorId_t fieldId) const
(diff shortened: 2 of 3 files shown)


silverweed:
Not sure if we actually need the non-recursive `GetColumnIds`, considering that it's basically just a trivial filter on `GetColumnIterable()`...do you have a concrete use case for it?

magnustymoteus:
I am currently using it in my treemap visualization, but yes one can easily replace it with a for loop over the `GetColumnIterable`. I'll remove the method then?

magnustymoteus:
Alright, I've removed `GetColumnIds(fieldId)`, but not `GetColumnIds(colType)` because it's used in `GetPageSizeDistribution`

tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
```diff
@@ -269,15 +269,15 @@ public:
    /// \param[in] colType The column type to collect, as defined by ROOT::ENTupleColumnType.
    ///
    /// \return A vector containing the physical IDs of columns of the provided type.
-   std::vector<ROOT::DescriptorId_t> GetColumnsByType(ROOT::ENTupleColumnType colType);
+   std::vector<ROOT::DescriptorId_t> GetColumnIds(ROOT::ENTupleColumnType colType);
```
silverweed: I think I actually prefer the current name for this method, I find it more explanatory

tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
```diff
@@ -269,15 +269,15 @@ public:
...
 
    /////////////////////////////////////////////////////////////////////////////
    /// \brief Get the columns that make up the given field, including its subfields.
    ///
    /// \param [in] fieldId The ID of the field for which to collect the columns.
    ///
    /// \return A vector containing the IDs of all columns for the provided field ID.
-   std::vector<ROOT::DescriptorId_t> GetColumnsByFieldId(ROOT::DescriptorId_t fieldId) const;
+   std::vector<ROOT::DescriptorId_t> GetColumnIdsRecursive(ROOT::DescriptorId_t fieldId) const;
```
silverweed: For this method, I know that I did say that it should specify it is recursive, but thinking again about it I might also prefer the current name, because it makes it clearer that the ID you're passing is the field id and also the method's documentation says that it includes the subfields. 
Maybe it could be clearer to call it something like `GetAllColumnsOfField` or `GetColumnsOfFieldRecursive`

tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
magnustymoteus: changed it back

tree/ntupleutil/v7/inc/ROOT/RNTupleInspector.hxx
magnustymoteus: Renamed it to `GetAllColumnsOfField`

PR #19468, opened 2025-07-28
[ntuple] Add support for emulating user collections with unknown collection proxies

silverweed:
Previously to this PR, if the user tried to load an emulated user-defined class using a custom CollectionProxy, they would get an untyped Record field which doesn't represent the collection properly.
With this change we instead detect if the field with unknown type is a Collection and, if so, we create an untyped Vector field rather than a Record field, allowing for proper inspection of the underlying values.

Another minor related change: in case of non-emulated fields (so if we have a proper TClass backing the user type) we don't require anymore that the class is at least Interpreted in case of collection proxies, as that is not necessary.

TODO: 
- add a test for the case where we have the dictionary for the user defined class but it's not associated to its original collection proxy (reading in this case should fail unless the user enables emulated fields).
- ~~verify that RNTupleReader::Show() works correctly~~ -> with the latest commit, it does

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/src/RFieldBase.cxx
+++ b/tree/ntuple/src/RFieldBase.cxx
@@ -555,7 +555,7 @@ ROOT::RFieldBase::Create(const std::string &fieldName, const std::string &typeNa
       if (!result) {
          auto cl = TClass::GetClass(typeName.c_str());
 
-         if (cl != nullptr) {
+         if (cl && cl->GetState() > TClass::kForwardDeclared) {
             createContextGuard.AddClassToStack(resolvedType);
             if (cl->GetCollectionProxy()) {
                result = std::make_unique<RProxiedCollectionField>(fieldName, typeName);
--- a/tree/ntuple/src/RFieldBase.cxx
+++ b/tree/ntuple/src/RFieldBase.cxx
@@ -554,15 +554,17 @@ ROOT::RFieldBase::Create(const std::string &fieldName, const std::string &typeNa
 
       if (!result) {
          auto cl = TClass::GetClass(typeName.c_str());
-         // NOTE: if the class is not at least "Interpreted" we currently don't try to construct
-         // the RClassField, as in that case we'd need to fetch the information from the StreamerInfo
-         // rather than from TClass. This might be desirable in the future, but for now in this
-         // situation we rely on field emulation instead.
-         if (cl != nullptr && cl->GetState() >= TClass::kInterpreted) {
+
+         if (cl != nullptr) {
             createContextGuard.AddClassToStack(resolvedType);
             if (cl->GetCollectionProxy()) {
                result = std::make_unique<RProxiedCollectionField>(fieldName, typeName);
-            } else {
+            }
+            // NOTE: if the class is not at least "Interpreted" we currently don't try to construct
+            // the RClassField, as in that case we'd need to fetch the information from the StreamerInfo
+            // rather than from TClass. This might be desirable in the future, but for now in this
+            // situation we rely on field emulation instead.
+            else if (cl->GetState() >= TClass::kInterpreted) {
                if (ROOT::Internal::GetRNTupleSerializationMode(cl) ==
                    ROOT::Internal::ERNTupleSerializationMode::kForceStreamerMode) {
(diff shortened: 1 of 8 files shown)


tree/ntuple/src/RFieldSequenceContainer.cxx
```diff
@@ -464,33 +464,38 @@ void ROOT::RRVecField::AcceptVisitor(ROOT::Detail::RFieldVisitor &visitor) const
...
+   return std::unique_ptr<ROOT::RVectorField>(new RVectorField(fieldName, std::move(itemField), ""));
 }
 
 std::unique_ptr<ROOT::RFieldBase> ROOT::RVectorField::CloneImpl(std::string_view newName) const
 {
    auto newItemField = fSubfields[0]->Clone(fSubfields[0]->GetFieldName());
+   auto isUntyped = GetTypeName().empty();
    return std::unique_ptr<ROOT::RVectorField>(
-      new RVectorField(newName, std::move(newItemField), GetTypeName().empty()));
+      new RVectorField(newName, std::move(newItemField), isUntyped ? std::make_optional("") : std::nullopt));
```
jblomer: Wouldn't that prevent us from cloning an emulated vector field?

tree/ntuple/src/RNTupleReader.cxx
```diff
@@ -218,8 +218,15 @@ void ROOT::RNTupleReader::PrintInfo(const ENTupleInfo what, std::ostream &output
 
 ROOT::RNTupleReader *ROOT::RNTupleReader::GetDisplayReader()
 {
-   if (!fDisplayReader)
-      fDisplayReader = Clone();
+   if (!fDisplayReader) {
+      std::unique_ptr<ROOT::RNTupleModel> fullModel;
+      auto descriptorGuard = fSource->GetSharedDescriptorGuard();
```
jblomer: Minor: this could be combined with the line `fullModel = ` into a block to keep the locking time for the descriptor guard minimal.

jblomer (approved):

tree/ntuple/src/RFieldBase.cxx
```diff
@@ -554,49 +554,69 @@ ROOT::RFieldBase::Create(const std::string &fieldName, const std::string &typeNa
 
       if (!result) {
          auto cl = TClass::GetClass(typeName.c_str());
-         // NOTE: if the class is not at least "Interpreted" we currently don't try to construct
-         // the RClassField, as in that case we'd need to fetch the information from the StreamerInfo
-         // rather than from TClass. This might be desirable in the future, but for now in this
-         // situation we rely on field emulation instead.
-         if (cl != nullptr && cl->GetState() >= TClass::kInterpreted) {
+
+         if (cl != nullptr) {
```
pcanal:
```suggestion
         if (cl != nullptr && cl->GetState() > TClass::kForwardDeclared) {
```
In the forward declared state, the TClass would only know there is a type and no information about it.

tree/ntuple/src/RFieldSequenceContainer.cxx
```diff
@@ -464,33 +464,38 @@ void ROOT::RRVecField::AcceptVisitor(ROOT::Detail::RFieldVisitor &visitor) const
...
-   return std::unique_ptr<ROOT::RVectorField>(new RVectorField(fieldName, std::move(itemField), true));
+   return std::unique_ptr<ROOT::RVectorField>(new RVectorField(fieldName, std::move(itemField), ""));
 }
 
 std::unique_ptr<ROOT::RFieldBase> ROOT::RVectorField::CloneImpl(std::string_view newName) const
 {
    auto newItemField = fSubfields[0]->Clone(fSubfields[0]->GetFieldName());
-   return std::unique_ptr<ROOT::RVectorField>(
-      new RVectorField(newName, std::move(newItemField), GetTypeName().empty()));
+   auto isUntyped = GetTypeName().empty() || (fTraits & kTraitEmulatedField) != 0;
```
pcanal: minor
```suggestion
   auto isUntyped = GetTypeName().empty() || ((fTraits & kTraitEmulatedField) != 0);
```

PR #19469, opened 2025-07-29
Patch builtin Clad version to add missing include

guitargeek:
There was a `#include <exception>` missing in case `NDEBUG` is not defined.


bellenot (approved):
Fine with me if it works 😉

vgvassilev:
Can we move this PR to clad, too?

PR #19470, opened 2025-07-29
[Python] Statically link cppyy_backend inside libcppyy

guitargeek:
Separating the libraries made sense in the past when the ROOT Python interfaces still supported PyPy, where only the backend needed to be loaded and there was no CPython extension. But nowadays, the clingwrapper in the `cppyy_backend` library is only used for the CPython extension `libcppyy`, so we might as well link statically.

This avoids some overhead and boilerplate code:

  * we don't need to set the `CPPYY_BACKEND_LIBRARY` path

  * we don't need system links for `libcppyy_backend.so` if the Python install directory differs from the library install directory

  * we don't need Python code to load the backend library

The change implies that the `rpath` code also needs to be moved from the backend to the CPyCppyy CMake code.


bindings/pyroot/cppyy/CPyCppyy/src/SignalTryCatch.h
```diff
@@ -71,6 +71,11 @@ using CppyyExceptionContext_t = ExceptionContext_t;
         gException = R__old; \
     }
 
-CPYCPPYY_IMPORT CppyyExceptionContext_t *gException;
+// extern, defined in ROOT Core
```
vepadulano, line 74: Where exactly is this defined in ROOT?

vepadulano (approved):
Thank you, I think the changes LGTM for what concerns ROOT! In the future we might want to discuss how to make this configurable in a generic way for other users of cppyy

bindings/pyroot/cppyy/CPyCppyy/src/SignalTryCatch.h
guitargeek: Here:

  * https://github.com/root-project/root/blob/master/core/base/inc/TException.h#L69

vepadulano: Ah! I thought you meant the `CppyyExceptionContext_t` type was defined in ROOT core, instead you meant the symbol, all good thanks!

vepadulano (approved):

guitargeek:
Fix to the unrelated RooFit failures on WIndows: https://github.com/root-project/root/pull/19475

jmcarcell:
I'm finding some issues that lead to the first commit here after bisecting. It seems I can only reproduce on Arch Linux (only tried GCC 15). So if I build ROOT (master or this commit)
``` bash
cmake .. -DCMAKE_CXX_STANDARD=20 -DCMAKE_BUILD_TYPE=Release -Dasimage=ON -Ddavix=OFF -Dfftw3=OFF -Dfitsio=OFF -Dgdml=ON -Dgfal=OFF -Dhttp=ON -Dimt=OFF -Dmathmore=ON -Dmlp=OFF -Dmysql=OFF -Dopengl=ON -Dpgsql=OFF -Dproof=OFF -Dpythia6=OFF -Dpythia8=OFF -Droofit=ON -Droot7=ON -Dspectrum=OFF -Dsqlite=OFF -Dssl=OFF -Dtmva-cpu=OFF -Dtmva=OFF -Dtmva-pymva=OFF -Dwebgui=ON -Dx11=ON -Dxml=OFF -Dxrootd=OFF -Dcuda=OFF -G Ninja
```
then I try to import ROOT from python:
``` bash
$ python -c "import ROOT"
Fatal in <TROOT::InitInterpreter>: cannot load library /usr/local/lib/libCling.so: undefined symbol: _ZN4ROOT10gReadRulesB5cxx11E
```
Any ideas? dev3 seems to work fine (on Alma9), but I have been able to reproduce this on two different machines, both with Arch.

guitargeek:
In the error message, I see it tried to load `libCling` from the system: `/usr/local/lib/libCling.so`

However, do you really install your ROOT build on the system, or is that some other ROOT build that might get in the way, maybe installed with the package manager?

ROOT is mot meant to work with different versions installed in the same environment, and maybe this PR resulted in a new case where installing different ROOT versions in the same environment can become problematic.

Probably it's because we use the standard library resolution mechanisms now, instead of a [custom loader script](https://github.com/wlav/cppyy-backend/blob/master/cling/python/cppyy_backend/loader.py) for cppyy that uses `ctypes.CDLL`.

If that's really the problem, then the fix is to make sure you don't have multiple ROOT versions installed.

jmcarcell:
This is a single install of ROOT in `/usr/local`, I don't have any other installations. I install with `sudo ninja install`, and also I tried a clean build and clean install but I still have that issue. I also tried installing as my user and using that and I get the same result.

guitargeek:
Ok. Then this is a problem. Can you open a GitHub issue about this, based on your original comment here? Then I won't forget to investigate it next week.

PR #19471, opened 2025-07-29
[InterOp] fix cmake to not put anything in install step

Vipul-Cariappa:


guitargeek (approved):
Thank you!

.github/workflows/cppinterop-diff.yml
```diff
@@ -29,4 +29,4 @@ jobs:
         run: |
           rm -rf .git .github discord.svg
       - name: Compare
-        run: diff -ur CppInterOp/ root/interpreter/CppInterOp/
+        run: diff --exclude=CMakeLists.txt -ur CppInterOp/ root/interpreter/CppInterOp/
```
hahnjo, line 32: I just noticed this, IMHO this should be reverted asap or otherwise it will be forgotten and stay around... @guitargeek @aaronj0

PR #19473, opened 2025-07-29
[CMake] Avoid installing unneeded files with `gnuinstall=OFF`

guitargeek:
More detail in the commit descriptions.

### DIFF
--- a/geom/CMakeLists.txt
+++ b/geom/CMakeLists.txt
@@ -15,12 +15,6 @@ add_subdirectory(geompainter)
 
 if(gdml)
   add_subdirectory(gdml)
-  if(NOT gnuinstall)
-    install(DIRECTORY gdml/ DESTINATION geom/gdml
-            FILES_MATCHING PATTERN "*.py"
-            PATTERN "inc" EXCLUDE
-            PATTERN "src" EXCLUDE)
-  endif()
 endif()
 
 if(vecgeom)
--- a/cmake/modules/RootConfiguration.cmake
+++ b/cmake/modules/RootConfiguration.cmake
@@ -809,12 +809,5 @@ install(FILES ${CMAKE_BINARY_DIR}/etc/root.mimes
               ${CMAKE_BINARY_DIR}/etc/system.rootdaemonrc
               DESTINATION ${CMAKE_INSTALL_SYSCONFDIR})
 
-if(NOT gnuinstall)
-  install(FILES ${CMAKE_BINARY_DIR}/config/Makefile.comp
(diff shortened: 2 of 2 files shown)


geom/CMakeLists.txt
```diff
@@ -15,12 +15,6 @@ add_subdirectory(geompainter)
 
 if(gdml)
   add_subdirectory(gdml)
-  if(NOT gnuinstall)
-    install(DIRECTORY gdml/ DESTINATION geom/gdml
-            FILES_MATCHING PATTERN "*.py"
-            PATTERN "inc" EXCLUDE
-            PATTERN "src" EXCLUDE)
-  endif()
```
vepadulano, line 23: It looks good in principle, but do we know where these scripts are used, if at all? What purpose do they serve being only available in the build directory? I guess it would only be fine if they were used in some tests alone and they're not supposed to be exposed to users.

guitargeek: I don't know. Maybe @agheata has an idea?

pcanal: >  What purpose do they serve being only available in the build directory? 

This is not how I read the `install` invocation.  I read it as installing this file 'only' in the install directory (and not in the build directory) but .... only if not using the `gnuinstall` layout.

> but do we know where these scripts are used, if at all? 

We don't know if they are used.  However according to the doc `geom/gdml/doc/index.md` those are used to convert `TGeo` objects to `gdml` from python.

So if anything it is surprising that they are not installed in all cases.   In a related observation they are not tested by ctest at the moment ... Maybe Witek has some left over tests?

guitargeek: Good that you point out the docs, I was not seeing that!

Then we have one more reason to not install these files. They are already part reference guide, where the code to `ROOTwriter.py` is included if you follow the links: https://root.cern.ch/doc/master/group__Geometry__gdml.html

I'm not sure how any user would use these files from this secret install location `geom/gdml`. If they are part of the documentation, is this not enough? The user can copy-paste the file to his project.

Testing is a separate issue, but this PR is not about that. For now, I want to harmonize the `gnuinstall=ON` and `gnuinstall=OFF` installations, because it there are these unexplained differences, it's harder for users and packagers to understand what flags they need to set.

vepadulano: I agree with Jonas, the simplification of the build system has more advantages than installing Python scripts that have never been tested and have never been reachable in any ROOT python installation before.

vepadulano (approved):

Issue #19390, opened 2025-07-16
c.Print when Title contains tex (old ROOT-7087 bug)


kjazatlas:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

hi,
using /cvmfs/sft.cern.ch/lcg/releases/LCG_107a_ATLAS_2/ROOT/6.34.04/x86_64-el9-gcc13-opt/bin/root
on
NAME="AlmaLinux"
VERSION="9.6 (Sage Margay)"

in the pyroot sample code below, if var_name contains tex (eg MSDisplacedVertex) the following error occurs
if var_name does not contain tex  (eg MSDisplacedVerte), all is well.

python test-bug.py
Info in <TCanvas::Print>: pdf file foo has been created
TPDF::Open:0: RuntimeWarning: PDF file already open
Info in <TCanvas::Print>: TeX file foo has been created
 *** Break *** segmentation violation

fixed? in ROOT-7087

```
import ROOT
output_filename = 'foo'

# Create a canvas for drawing the histograms
c = ROOT.TCanvas("c", "Histograms", 800, 600)

# Open the PDF file for multipage output
c.Print(output_filename + "[", "pdf")
var_name = 'MSDisplacedVertex'
c.Print(output_filename, "Title: " + var_name)
c.Print(output_filename + "]")


### Reproducer

setupATLAS
lsetup "root 6.34.04-x86_64-el9-gcc13-opt"
python code.py
import ROOT
output_filename = 'foo'
# Create a canvas for drawing the histograms
c = ROOT.TCanvas("c", "Histograms", 800, 600)

# Open the PDF file for multipage output
c.Print(output_filename + "[", "pdf")
var_name = 'MSDisplacedVertex'
c.Print(output_filename, "Title: " + var_name)
c.Print(output_filename + "]")
```

### ROOT version

/cvmfs/sft.cern.ch/lcg/releases/LCG_107a_ATLAS_2/ROOT/6.34.04/x86_64-el9-gcc13-opt/bin/root

### Installation method

NAME="AlmaLinux" VERSION="9.6 (Sage Margay)"

### Operating system

NAME="AlmaLinux" VERSION="9.6 (Sage Margay)"

ferdymercury:
Indeed, the original ticket 7087

```
   TCanvas *c = new TCanvas();
   c->Print("output.pdf[");
   c->Print("output.pdf", "Title:Vertex");
   c->Print("output.pdf]");
```
also fails on master.

[...]

PR #19474, opened 2025-07-29
Option Title: with title containing tex

couet:
With pdf output it is possible to define a table of content of a multi-pages file. Each page has its name specified using the option "Titlle:" followed by the title of the page. If the title contains a string matching a possible file extension like, for instance, "tex"o r "png", then the pdf generation went wrong and might crash. This PR fixes the problem.

Fixes https://github.com/root-project/root/issues/19390

### DIFF
--- a/graf2d/gpad/src/TPad.cxx
+++ b/graf2d/gpad/src/TPad.cxx
@@ -5123,6 +5123,9 @@ void TPad::Print(const char *filename, Option_t *option)
    TString opt = !option ? opt_default : option;
    Bool_t image = kFALSE;
 
+   Bool_t title = kFALSE;
+   if (strstr(opt,"Title:")) title = kTRUE;
+
    if (!fs1.Length())  {
       psname = GetName();
       psname += opt;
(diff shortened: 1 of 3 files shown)


dpiparo:
could this be backported and merged to 6.36 when merged?

ferdymercury:
Thanks for this nice fix Olivier!
I would suggest adding the following Unit Test:

```
   TCanvas c;
   c.Print("outputJIRA7087.pdf[");
   c.Print("outputJIRA7087.pdf", "Title:Vertex");
   c.Print("outputJIRA7087.pdf]");
   gSystem->Unlink("outputJIRA7087.pdf");
```

couet:
Note the proper syntax is:
```
   c.Print("outputJIRA7087.pdf(");
...
   c.Print("outputJIRA7087.pdf)");
```
With parenthesis... not square brakets.

ferdymercury:
> With parenthesis... not square brackets.

Both options are supported, see https://github.com/root-project/root/commit/e06e246d2666d16b96778b6d0ff68dbb1f1632e1

couet:
This is not equivalent. Try your test with [ and see the doc of TPad::Print

ferdymercury:
Hmmm that documentation you mention says:

```
The "[" and "]" can be used instead of "(" and ")".
```

So this sentence is a bit ambiguous.

My proposal would be to change that line as follows:

`The "[" and "]" can be used instead of "(" and ")" to open / close without actual printing.`

ferdymercury (approved):
Thanks a lot!

PR #19478, opened 2025-07-30
[core] Rework ErrorHandler to avoid allocating memory if possible

silverweed:
Alternative to https://github.com/root-project/root/pull/19395 that doesn't allocate memory unless necessary

### DIFF
--- a/core/foundation/src/TError.cxx
+++ b/core/foundation/src/TError.cxx
@@ -123,41 +123,48 @@ void ErrorHandler(Int_t level, const char *location, const char *fmt, std::va_li
    std::va_list apCopy;
    va_copy(apCopy, ap);
 
-   // Figure out exactly how many bytes we need
-   int nRequired = vsnprintf(nullptr, 0, fmt, ap) + 1;
+   // Write the user string to the current buffer and, at the same time, figure out exactly how many bytes we need.
+   int nWritten = vsnprintf(buf, bufSize, fmt, ap);
+   int nAdditional = 0;
    if (level >= kSysError && level < kFatal) {
       auto sysHandler = GetErrorSystemMsgHandlerRef();
       if (sysHandler)
-         nRequired += strlen(sysHandler()) + 1; // +1 for the whitespace
+         nAdditional = strlen(sysHandler()) + 1; // +1 for the whitespace
       else {
-         nRequired += snprintf(nullptr, 0, " (errno: %d)", errno) + 1;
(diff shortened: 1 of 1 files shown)


core/foundation/src/TError.cxx
```diff
@@ -108,46 +110,59 @@ ErrorHandlerFunc_t GetErrorHandler()
...
+   int nWrittenPost = 0;
    if (level >= kSysError && level < kFatal) {
-      bp.push_back(' ');
-      if (GetErrorSystemMsgHandlerRef())
-         bp += GetErrorSystemMsgHandlerRef()();
-      else
-         bp += std::string("(errno: ") + std::to_string(errno) + ")";
+      auto sysHandler = GetErrorSystemMsgHandlerRef();
+      if (sysHandler) {
+         // NOTE: overwriting the null byte written by the previous vsnprintf
```
pcanal: minor nit pick.  Isn't it 'potential overwrite' if previous message was shorter?

core/foundation/src/TError.cxx
```diff
@@ -108,46 +110,59 @@ ErrorHandlerFunc_t GetErrorHandler()
...
+   char *buf = bufDynStorage ? bufDynStorage : smallBuf;
 
-   Int_t n = vsnprintf(buf, buf_size, fmt, ap_copy);
-   if (n >= buf_size) {
-      va_end(ap_copy);
+   std::va_list apCopy;
+   va_copy(apCopy, ap);
+
+   // Figure out exactly how many bytes we need
+   int nRequired = vsnprintf(nullptr, 0, fmt, ap) + 1;
```
pcanal: Isn't it 'better' (albeit adding some code complexity) to make the first call to `vsnprintf` directly into the buffer so that in the usual case (short messages) there is only one call (as opposed to the current PR code that always requires 2 call to `vsnprintf`).
```suggestion
   int nRequired = vsnprintf(nullptr, 0, fmt, ap) + 1;
```

pcanal (approved):
LGTM (beside performance discussion on the number of calls to `vsnprintf`).  Thanks.

core/foundation/src/TError.cxx
silverweed: The first vsnprintf will print the user's message (and null terminate it), then we extend it by appending the sys error, which is always(?) non-empty, and its first byte will always overwrite the null byte marking the end of the user message

PR #19481, opened 2025-07-31
[CMake] Remove `ROOTTEST_DIR` variable

guitargeek:
It's not needed anymore now that `roottest` was merged into the main repository.

Follows up on a comment by @pcanal:

  * https://github.com/root-project/root/pull/19370#discussion_r2242930620

### DIFF
--- a/cmake/modules/RootMacros.cmake
+++ b/cmake/modules/RootMacros.cmake
@@ -2155,7 +2155,7 @@ endfunction(ROOTTEST_SET_TESTOWNER)
 # Construct a target name for a given file <filename> and store its name into
 # <resultvar>. The target name is of the form:
 #
-#   roottest-<directorypath>-<filename_WE>
+#   <directorypath>-<filename_WE>
 #
 #-------------------------------------------------------------------------------
 function(ROOTTEST_TARGETNAME_FROM_FILE resultvar filename)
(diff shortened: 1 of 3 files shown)


pcanal (approved):
LGTM.  Thanks.

cmake/modules/RootMacros.cmake
```diff
@@ -2155,23 +2155,19 @@ endfunction(ROOTTEST_SET_TESTOWNER)
...
 
   get_filename_component(realfp ${filename} ABSOLUTE)
   get_filename_component(filename_we ${filename} NAME_WE)
 
-  if(DEFINED ROOTTEST_DIR)
-    string(REPLACE "${ROOTTEST_DIR}" "" relativepath ${realfp})
-  else()
-    string(REPLACE "${CMAKE_SOURCE_DIR}" "" relativepath ${realfp})
-  endif()
+  string(REPLACE "${CMAKE_SOURCE_DIR}/" "" relativepath ${realfp})
```
pcanal, line 2166: it is minorly surprising that the trailing '/' is now needed/used while it "seemed" to not be needed before (albeit maybe the else branch was not actually used anymore).

guitargeek: Well, it's just a matter of definition. Before it was needed, because the test name was prefixed with `"roottest${targetname}"` without the dash. So since the trailing `/` after `CMAKE_SOURCE_DIR` was not replaced with nothing, it stayed there and became the dash in `roottest-`. Now, we don't want the test name to start with a `-`, so we also need to get rid of the initial `/` of the relative path.

PR #19483, opened 2025-07-31
[hist] Check arguments in axis constructors

hahnjo:
Also test `ComputeLinearizedIndex` exactly on the edges (this already worked since https://github.com/root-project/root/pull/19334).

### DIFF
--- a/hist/histv7/inc/ROOT/RRegularAxis.hxx
+++ b/hist/histv7/inc/ROOT/RRegularAxis.hxx
@@ -9,6 +9,7 @@
 
 #include <cstddef>
 #include <stdexcept>
+#include <string>
 
 class TBuffer;
 
@@ -44,14 +45,20 @@ class RRegularAxis final {
 public:
    /// Construct a regular axis object.
    ///
-   /// \param[in] numNormalBins the number of normal bins
+   /// \param[in] numNormalBins the number of normal bins, must be > 0
    /// \param[in] low the lower end of the axis interval (inclusive)
-   /// \param[in] high the upper end of the axis interval (exclusive)
+   /// \param[in] high the upper end of the axis interval (exclusive), must be > low
    /// \param[in] enableFlowBins whether to enable underflow and overflow bins
    RRegularAxis(std::size_t numNormalBins, double low, double high, bool enableFlowBins = true)
       : fNumNormalBins(numNormalBins), fLow(low), fHigh(high), fEnableFlowBins(enableFlowBins)
    {
-      // FIXME: should validate numNormalBins > 0 and low < high
+      if (numNormalBins == 0) {
+         throw std::invalid_argument("numNormalBins must be > 0");
+      }
+      if (low >= high) {
+         std::string msg = "high must be > low, but " + std::to_string(low) + " >= " + std::to_string(high);
+         throw std::invalid_argument(msg);
+      }
       fInvBinWidth = numNormalBins / (high - low);
    }
 
--- a/hist/histv7/inc/ROOT/RVariableBinAxis.hxx
+++ b/hist/histv7/inc/ROOT/RVariableBinAxis.hxx
@@ -9,6 +9,7 @@
 
 #include <cstddef>
 #include <stdexcept>
+#include <string>
 #include <utility>
 #include <vector>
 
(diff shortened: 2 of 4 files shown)


hist/histv7/inc/ROOT/RRegularAxis.hxx
```diff
@@ -44,14 +44,19 @@ class RRegularAxis final {
...
    /// \param[in] enableFlowBins whether to enable underflow and overflow bins
    RRegularAxis(std::size_t numNormalBins, double low, double high, bool enableFlowBins = true)
       : fNumNormalBins(numNormalBins), fLow(low), fHigh(high), fEnableFlowBins(enableFlowBins)
    {
-      // FIXME: should validate numNormalBins > 0 and low < high
+      if (numNormalBins == 0) {
+         throw std::invalid_argument("numNormalBins must be > 0");
+      }
+      if (low >= high) {
+         throw std::invalid_argument("high must be > low");
```
ferdymercury:
```suggestion
         throw std::invalid_argument("high (" + std::to_string(high) + ") must be > low (" + std::to_string(low) + ")");
```
Potential improvement, it might facilitate finding which of the eg 5 histograms in the script the user swapped high and low

hahnjo: I considered this and I'm still open to adding this information. However, for "finding the offender" isn't it easier to use the stacktrace? I mean that's why we throw exceptions in the first place...

ferdymercury: Agreed, but then it would be better to fix first: https://its.cern.ch/jira/browse/ROOT-9296

hist/histv7/inc/ROOT/RVariableBinAxis.hxx
```diff
@@ -46,7 +47,16 @@ public:
    RVariableBinAxis(std::vector<double> binEdges, bool enableFlowBins = true)
```
siliataider, line 47: The parameter docstring can maybe be slightly more precise to match the validation logic? something like:
```cpp
/// \param[in] binEdges the (ordered) edges of the normal bins, must define at least one bin (i.e. size >= 2)
```

hist/histv7/inc/ROOT/RVariableBinAxis.hxx
```diff
@@ -41,12 +42,21 @@ class RVariableBinAxis final {
...
    ///
-   /// \param[in] binEdges the (ordered) edges of the normal bins
+   /// \param[in] binEdges the (ordered) edges of the normal bins, must define at least one bin (i.e. size >= 2)
    /// \param[in] enableFlowBins whether to enable underflow and overflow bins
    RVariableBinAxis(std::vector<double> binEdges, bool enableFlowBins = true)
       : fBinEdges(std::move(binEdges)), fEnableFlowBins(enableFlowBins)
    {
-      // FIXME: should validate that fBinEdges is sorted
+      if (fBinEdges.size() < 2) {
+         throw std::invalid_argument("must have >= 2 edges");
```
ferdymercury:
```suggestion
         throw std::invalid_argument("must have >= 2 edges, but has " + std::to_string(fBinEdges.size()));
```

hist/histv7/inc/ROOT/RRegularAxis.hxx
```diff
@@ -44,14 +45,20 @@ class RRegularAxis final {
...
    /// \param[in] low the lower end of the axis interval (inclusive)
-   /// \param[in] high the upper end of the axis interval (exclusive)
+   /// \param[in] high the upper end of the axis interval (exclusive), must be > low
    /// \param[in] enableFlowBins whether to enable underflow and overflow bins
    RRegularAxis(std::size_t numNormalBins, double low, double high, bool enableFlowBins = true)
       : fNumNormalBins(numNormalBins), fLow(low), fHigh(high), fEnableFlowBins(enableFlowBins)
    {
-      // FIXME: should validate numNormalBins > 0 and low < high
+      if (numNormalBins == 0) {
+         throw std::invalid_argument("numNormalBins must be > 0");
```
ferdymercury, line 56:
```suggestion
         throw std::invalid_argument("numNormalBins=" + std::to_string(numNormalBins) + " must be > 0");
```

hahnjo: I don't think we need this: `numNormalBins` is unsigned, so the only value that triggers this exception is 0.

hist/histv7/inc/ROOT/RVariableBinAxis.hxx
hahnjo: For the same reason, I would argue that the user should be able to find the constructor with zero or only one edge.

ferdymercury: Agreed, although, assuming that the stacktrace is not printed, then the error might be cryptic. Imagine that you have 2000-lines long script and you get:

```
root [0] 
Processing crash.cxx...
terminate called after throwing an instance of 'std::invalid_argument'
  what():  must have 2 edges
```

Where is this message coming from, which class? Maybe saying "bin edges" instead of "edges" is more helpful. At least one then know it's the hist class and not the TRectangle class that complained.

Maybe it's because I'm used to the more verbose message such as
Error in TTree::ParseLeaf leaf XX bla bla...
where one gets detailed info.

PR #19484, opened 2025-07-31
[ntuple] Add Merger tests for Streamer fields

silverweed:
Also change the failure in GenerateZeroPages from a R__LOG_FATAL to a ROOT::Result failure (we don't really need to abort the whole process if merging fails)

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/foundation/inc/ROOT/RError.hxx
+++ b/core/foundation/inc/ROOT/RError.hxx
@@ -125,6 +125,7 @@ public:
    void Throw();
 
    /// Used by R__FORWARD_ERROR in order to keep track of the stack trace.
+   [[nodiscard]]
    static RError ForwardError(RResultBase &&result, RError::RLocation &&sourceLocation)
    {
       if (!result.fError) {
--- a/tree/ntuple/src/RNTupleMerger.cxx
+++ b/tree/ntuple/src/RNTupleMerger.cxx
@@ -586,8 +586,10 @@ CompareDescriptorStructure(const ROOT::RNTupleDescriptor &dst, const ROOT::RNTup
 }
 
 // Applies late model extension to `destination`, adding all `newFields` to it.
-static void ExtendDestinationModel(std::span<const ROOT::RFieldDescriptor *> newFields, ROOT::RNTupleModel &dstModel,
-                                   RNTupleMergeData &mergeData, std::vector<RCommonField> &commonFields)
+[[nodiscard]]
+static ROOT::RResult<void>
+ExtendDestinationModel(std::span<const ROOT::RFieldDescriptor *> newFields, ROOT::RNTupleModel &dstModel,
+                       RNTupleMergeData &mergeData, std::vector<RCommonField> &commonFields)
 {
    assert(newFields.size() > 0); // no point in calling this with 0 new cols
 
@@ -636,7 +638,11 @@ static void ExtendDestinationModel(std::span<const ROOT::RFieldDescriptor *> new
       ROOT::Internal::GetProjectedFieldsOfModel(dstModel).Add(std::move(field), fieldMap);
    }
    dstModel.Freeze();
-   mergeData.fDestination.UpdateSchema(changeset, mergeData.fNumDstEntries);
+   try {
+      mergeData.fDestination.UpdateSchema(changeset, mergeData.fNumDstEntries);
+   } catch (const ROOT::RException &ex) {
+      return R__FAIL(ex.GetError().GetReport());
+   }
 
    commonFields.reserve(commonFields.size() + newFields.size());
    for (const auto *field : newFields) {
@@ -644,6 +650,8 @@ static void ExtendDestinationModel(std::span<const ROOT::RFieldDescriptor *> new
       const auto &newFieldInDst = mergeData.fDstDescriptor.GetFieldDescriptor(newFieldInDstId);
       commonFields.emplace_back(*field, newFieldInDst);
    }
+
+   return ROOT::RResult<void>::Success();
 }
 
 // Generates default (zero) values for the given columns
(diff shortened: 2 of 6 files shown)


tree/ntuple/test/ntuple_merger.cxx
```diff
@@ -3506,3 +3506,208 @@ TEST(RNTupleMerger, GenerateZeroPagesIncremental)
...
+            RNTupleMergeOptions opts;
+            opts.fMergingMode = mmode;
+            auto res = merger.Merge(sourcePtrs, opts);
+            ASSERT_TRUE(bool(res));
+         }
+
+         auto reader = RNTupleReader::Open("ntuple", fileGuardOut.GetPath());
+         EXPECT_EQ(reader->GetNEntries(), 20);
+         auto pFoo = reader->GetModel().GetDefaultEntry().GetPtr<CustomStruct>("foo");
+         auto pBar = reader->GetModel().GetDefaultEntry().GetPtr<CustomStruct>("bar");
```
jblomer, line 3578: I think we should also explicitly check that these fields were reconstructed as streamer fields (even though if they weren't, the rest of the test currently happens to fail).

tree/ntuple/test/ntuple_merger.cxx
```diff
@@ -3506,3 +3506,208 @@ TEST(RNTupleMerger, GenerateZeroPagesIncremental)
...
+            EXPECT_EQ(pFoo->s, std::to_string(idx % 10));
+            EXPECT_FLOAT_EQ(pBar->a, idx < 10 ? idx : 2.f * (idx - 10));
+         }
+      }
+   }
+}
+
+TEST(RNTupleMerger, MergeStreamerFieldsFirstMissing)
+{
+   // Merge two files where the second has an additional streamer field - should succeed.
```
jblomer: Why should this succeed? Isn't that late model extension on streamer fields?

silverweed: Currently we only fail if we try generating zero pages, which is not required in this case because we can instead set the first entry index on the second field. That said we might still want to prevent this...

tree/ntuple/src/RNTupleMerger.cxx
```diff
@@ -586,8 +586,10 @@ CompareDescriptorStructure(const ROOT::RNTupleDescriptor &dst, const ROOT::RNTup
 }
 
 // Applies late model extension to `destination`, adding all `newFields` to it.
-static void ExtendDestinationModel(std::span<const ROOT::RFieldDescriptor *> newFields, ROOT::RNTupleModel &dstModel,
-                                   RNTupleMergeData &mergeData, std::vector<RCommonField> &commonFields)
+[[nodiscard]]
```
pcanal, line 589: What is the intent of this addition?

core/foundation/inc/ROOT/RError.hxx
```diff
@@ -125,6 +125,7 @@ public:
    void Throw();
 
    /// Used by R__FORWARD_ERROR in order to keep track of the stack trace.
+   [[nodiscard]]
```
pcanal, line 128: This is a (small) change in behavior, should we mention it in the commit log?

tree/ntuple/test/ntuple_merger.cxx
```diff
@@ -3506,3 +3506,212 @@ TEST(RNTupleMerger, GenerateZeroPagesIncremental)
...
+   
+   {
+      // Gather the input sources
+      std::vector<std::unique_ptr<RPageSource>> sources;
+      sources.push_back(RPageSource::Create("ntuple", fileGuard1.GetPath(), RNTupleReadOptions()));
+      sources.push_back(RPageSource::Create("ntuple", fileGuard2.GetPath(), RNTupleReadOptions()));
+      std::vector<RPageSource *> sourcePtrs;
+      for (const auto &s : sources) {
+         sourcePtrs.push_back(s.get());
+      }
```
pcanal, line 3560: Side note unrelated to the PR.  It feel cumbersome to have to go through this hoop to call `Merge`.  Is it conceivable to have an overload that take a `std::vector<std::unique_ptr<RPageSource>> &` (maybe even `const`) ?

pcanal (approved):
Thanks.

tree/ntuple/test/ntuple_merger.cxx
silverweed: I'm also annoyed by having to do that conversion every time, but I'm not sure if it's worth adding a method for that (in non-test code how often does one end up doing this?)
Ideally we would like to pass any iterator that yields a `RPageSource*`, but afaik that's not easy to do in C++...

tree/ntuple/src/RPageStorage.cxx
```diff
@@ -830,6 +830,11 @@ ROOT::Internal::RPagePersistentSink::AddColumn(ROOT::DescriptorId_t fieldId, RCo
 void ROOT::Internal::RPagePersistentSink::UpdateSchema(const ROOT::Internal::RNTupleModelChangeset &changeset,
                                                        ROOT::NTupleSize_t firstEntry)
 {
+   if (fIsInitialized)
+      for (const auto &field : changeset.fAddedFields)
+         if (field->GetStructure() == ENTupleStructure::kStreamer)
+            throw ROOT::RException(R__FAIL("a Model cannot be extended with Streamer fields"));
+
```
hahnjo, line 837: This change should probably also be tested outside the merger, in `tree/ntuple/test/ntuple_modelext.cxx`, as it's a general restriction...

silverweed: https://github.com/root-project/root/pull/19581

PR #19486, opened 2025-07-31
[ntuple] Prevent creating a RNTupleView of a Streamer field

silverweed:
An RNTupleView created from a Streamer field currently silently reads zeroed values if the dictionary is available, as the View attempts to create a RClassField but fails to find any columns for its subfields. This is not considered an error because of schema evolution, so all fields will be marked as artificial and zero initialized.

To prevent this from happening, explicitly check that the field is not stored as a Streamer field on disk in BeforeConnectPageSource and error out if that's the case.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/inc/ROOT/RNTupleReader.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleReader.hxx
@@ -266,6 +266,8 @@ public:
    /// **Note**: if `T = void`, type checks are disabled. This is not really useful for this overload because
    /// RNTupleView<void> does not give access to the pointer. If required, it is possible to provide an `objPtr` of a
    /// dynamic type, for example via GetView(std::string_view, void *, std::string_view).
+   ///
+   /// Note that creating a View for a Streamer field is not supported.
    template <typename T>
    ROOT::RNTupleView<T> GetView(std::string_view fieldName)
    {
@@ -295,6 +297,8 @@ public:
    /// object types. It is strongly recommended to use an overload that allows passing the `typeName`, such as
    /// GetView(std::string_view, void *, std::string_view). This allows type checks with the on-disk metadata and
    /// enables automatic schema evolution and conversion rules.
+   ///
+   /// Note that creating a View for a Streamer field is not supported.
    template <typename T>
    ROOT::RNTupleView<T> GetView(std::string_view fieldName, std::shared_ptr<T> objPtr)
    {
--- a/tree/ntuple/src/RFieldMeta.cxx
+++ b/tree/ntuple/src/RFieldMeta.cxx
@@ -421,6 +421,9 @@ void ROOT::RClassField::BeforeConnectPageSource(ROOT::Internal::RPageSource &pag
       const ROOT::RNTupleDescriptor &desc = descriptorGuard.GetRef();
       const auto &fieldDesc = desc.GetFieldDescriptor(GetOnDiskId());
 
+      if (fieldDesc.GetStructure() == ENTupleStructure::kStreamer)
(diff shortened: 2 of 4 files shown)


tree/ntuple/inc/ROOT/RNTupleReader.hxx
```diff
@@ -266,6 +266,8 @@ public:
    /// **Note**: if `T = void`, type checks are disabled. This is not really useful for this overload because
    /// RNTupleView<void> does not give access to the pointer. If required, it is possible to provide an `objPtr` of a
    /// dynamic type, for example via GetView(std::string_view, void *, std::string_view).
+   ///
+   /// Note that creating a View for a Streamer field is not supported.
```
pcanal, line 270: Is that a temporary limitation (need further development to support) or fundamental limitation (if it is fundamental, why?)

silverweed: As far as I understand it's something we could do, but at the cost of making the code more complex and I don't think there is a good incentive to do that (at least right now).

jblomer: Correct. We would need to "mark" these types as we do for collection proxies. In the case of streamer fields, the dictionary code could take care of the marking.

hahnjo: Not for the compile time type though...

In general, I think the comment is not 100% accurate: I think you can use a `RNTupleView<void>` with streamer fields (to be tested).

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -421,6 +421,9 @@ void ROOT::RClassField::BeforeConnectPageSource(ROOT::Internal::RPageSource &pag
       const ROOT::RNTupleDescriptor &desc = descriptorGuard.GetRef();
       const auto &fieldDesc = desc.GetFieldDescriptor(GetOnDiskId());
 
+      if (fieldDesc.GetStructure() == ENTupleStructure::kStreamer)
```
jblomer, line 424: Should we (in addition) check that the structural role is `record`?

hahnjo:
I think the core fixes in the PR are unrelated to views: We must error if a user tries to connect a `RClassField` to an on-disk streamer field, and vice versa. Views expose this problem, but I think it goes deeper.

tree/ntuple/inc/ROOT/RNTupleReader.hxx
jblomer: Oh, I guess that's right. But then, isn't it even easier and in `RNTupleViewBase<T>` we just should always use `RFieldBase::Create()`, either with the given type name or with the demangled and renormalized typename of `typeid(T)` (or using `RFieldDescriptor::CreateField()`)?

hahnjo: Maybe, we would lose the nicely typed `RField<T>` though. In any case, I think we must first fix `RClassField` and `RStreamerField` to prevent cross-connects, and as argued that's independent of views.

hahnjo:
What is the status of this after https://github.com/root-project/root/pull/20008? I believe this should "just work", at least in the direction "streamer field on disk, class field in memory"...

silverweed:
> What is the status of this after https://github.com/root-project/root/pull/20008? I believe this should "just work", at least in the direction "streamer field on disk, class field in memory"...

After checking, I confirm creating views for streamer fields now works, so this PR can be closed.

PR #19487, opened 2025-07-31
[Python][RDF] Add support for std::vector and std::array in numba jitted functions

siliataider:
# This Pull request:

## Changes or fixes:
This PR adds support for using `std::vector<T>` and `std::array<T>` as input and return types in functions declared with `ROOT.Numba.Declare` used in RDataFrame `.Define` operations.

This makes it possible to define and use functions like:
```python
@ROOT.Numba.Declare(["std::array<int, 3>"], "std::vector<int>")
def square_vec(x):
    return x * x

rdf = ROOT.RDataFrame(3) \
        .Define("xvec", "std::array{(int)rdfentry_, (int)rdfentry_ + 1, (int)rdfentry_ + 2}") \
        .Define("xvec_square", "Numba::square_vec(xvec)")
```

The implementation reuses the logic for handling RVecs.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
@@ -33,9 +33,37 @@ def _NumbaDeclareDecorator(input_types, return_type = None, name=None):
     except:
         raise Exception('Failed to import cffi')
     import re, sys
+    from typing import Union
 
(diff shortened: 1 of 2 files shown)


vepadulano (approved):
Thank you!

vepadulano:
Please squash commit history before merging :+1:

PR #19488, opened 2025-07-31
[ntuple] Allow merging fields with unknown types

silverweed:
Based on https://github.com/root-project/root/pull/19468 

See commit descriptions for details

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/src/RNTupleMerger.cxx
+++ b/tree/ntuple/src/RNTupleMerger.cxx
@@ -1159,11 +1159,9 @@ ROOT::RResult<void> RNTupleMerger::Merge(std::span<RPageSource *> sources, const
 
    // Merge main loop
    for (RPageSource *source : sources) {
-      if (auto *sourceFile = dynamic_cast<RPageSourceFile *>(source)) {
-         // We need to make sure the streamer info from the source files is loaded otherwise we may not be able
-         // to build the streamer info of user-defined types unless we have their dictionaries available.
-         sourceFile->LoadStreamerInfo();
-      }
+      // We need to make sure the streamer info from the source files is loaded otherwise we may not be able
+      // to build the streamer info of user-defined types unless we have their dictionaries available.
+      source->LoadStreamerInfo();
 
       source->Attach(RNTupleSerializer::EDescriptorDeserializeMode::kForWriting);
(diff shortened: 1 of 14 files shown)


tree/ntuple/src/RNTupleMerger.cxx
```diff
@@ -1136,6 +1136,12 @@ ROOT::RResult<void> RNTupleMerger::Merge(std::span<RPageSource *> sources, const
 
    // Merge main loop
    for (RPageSource *source : sources) {
+      if (auto *sourceFile = dynamic_cast<RPageSourceFile *>(source)) {
+         // We need to make sure the streamer info from the source files is loaded otherwise we may not be able
+         // to build the streamer info of user-defined types unless we have their dictionaries available.
+         sourceFile->LoadStreamerInfo();
+      }
```
jblomer: I'd suggest to make `LoadStreamerInfo` a method of the `RPageSource` base class. In the DAOS page source, we just print a warning (which we can ignore in the tests). This way, we don't forget about this point when we go back to the object store page sources.

silverweed: But what's the advantage of printing this warning to the user if they try to merge DAOS-backed sources (when that will be possible to do)?
In principle there is no action to be taken on their side and the warning is out of their control, so wouldn't it be better if we avoided it on our end?

jblomer: The user action would be submitting a bug report.

silverweed:
Note: this is currently blocked by a consistent crash on Windows x86, which @bellenot is investigating (thanks 🙏🏻 ). The crash definitely seems related to the change and it happens in RNTupleMerger:1171, only if LoadStreamerInfo() is called.

silverweed:
> Note: this is currently blocked by a consistent crash on Windows x86, which @bellenot is investigating (thanks 🙏🏻 ). The crash definitely seems related to the change and it happens in RNTupleMerger:1171, only if LoadStreamerInfo() is called.

Apparently we were invoking UB by static casting every member of the StreamerInfoList to `StreamerInfo*`, but in some cases the last element of the list is actually a `TList`. 
The fix is simply checking if the class is correct before casting. 
Thanks @pcanal and @bellenot  for the help finding the problem

PR #19490, opened 2025-07-31
[RF] Make `RooAbsReal` and `RooAbsPdf` instantiable

guitargeek:
In some cases, it might happen that a CallFunc wrapper for the `RooAbsReal` or `RooAbsPdf` classes is created, which gives compile errors because these classes are virtual.

It is however not a big deal to make these classes non-virtual and avoid these errors. The `RooAbsArg::clone()` method should have been overridden anyway since `RooAbsReal` and `RooAbsPdf` already have copy constructors. The only other method that needs to be implemented is `RooAbsReal::evaluate()`, which is done as well. The implementation throws an exception, because `evaluate()` should always be called from derived classes.

This is a bit of a hotfix and one can argue that it would be nicer to first understand why these CallFunc wrappers are even created, but this issue needs to be resolved quickly and with minimal changes: the fix also needs to be backported to older ROOT versions and is crucial for the ATLAS Run 2 Higgs combination analysis.

Issue reported on the ROOT forum:
https://root-forum.cern.ch/t/cannot-open-a-rooworkspace-once-it-is-saved-allocating-an-object-of-class-type-error/63954

### DIFF
--- a/roofit/roofitcore/inc/RooAbsPdf.h
+++ b/roofit/roofitcore/inc/RooAbsPdf.h
@@ -36,8 +36,11 @@ public:
   RooAbsPdf() ;
   RooAbsPdf(const char *name, const char *title=nullptr) ;
   RooAbsPdf(const char *name, const char *title, double minVal, double maxVal) ;
-  // RooAbsPdf(const RooAbsPdf& other, const char* name=nullptr);
   ~RooAbsPdf() override;
+
+  TObject* clone(const char* newname=nullptr) const override {
+    return new RooAbsPdf(*this,newname);
+  }
 
   // Toy MC generation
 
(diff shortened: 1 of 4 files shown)


pcanal:
Was a ticket opened for the underlying issue?

guitargeek:
Not yet. I'm working on creating a minimal reproducer together with Matthew from ATLAS who posted in the forum. But I'll open the issue before merging this PR

guitargeek:
This PR is fortunately not needed anymore: the underlying problem was identified.

The user was accidentally putting `RooAbsReal/Pdf`-derived classes in their workspace that have no class version themselves (more particular, the likelihoods created by `RooAbsPdf::createNLL`). This is not allowed.

PR #19492, opened 2025-07-31
[roottest] use cmake for root/tree tests

linev:


linev:
Closing in favor of https://github.com/root-project/root/pull/19516, which includes all necessary changes in `root/tree` tests

PR #19494, opened 2025-07-31
[jupyter] Change ROOT to set JUPYTER_CONFIG_PATH not JUPYTER_CONFIG_DIR

### LINKED ISSUE #19320: Jupyter Lab does not start when a read-only ROOT installation is present (loaded via thisroot.sh)
### Check duplicate issues.

- [x] Checked for duplicates

### Description

When ROOT is installed in a read-only location (for normal users), and the ROOT environment has been activated with `source [..]/thisroot.sh`, Jupyter Lab does not start / work anymore!

The steps to reproduce this (and also fix this) are show below.

My investigations have shown, that this is because ROOT (via `thisroot.sh`) sets two environment variables (paths for my installation in `/opt/` - again not writeable by normal users):
- `JUPYTER_CONFIG_DIR=/opt/root/v6_36_00/etc/notebook`
- `JUPYTER_PATH=/opt/root/v6_36_00/etc/notebook`

The problem is the setting of the `JUPYTER_CONFIG_DIR`, which by Jupyter is used to determine the location of the user settings!
- On a fresh install of Jupyter (without any environment variables set), the config directory by default points to `/home/[USER]/.jupyter` (can be shown via `jupyter --config-dir`).
- This directory is of course writeable by the user (as it is in the user home directory).

When ROOT now sets this directory to a non-readable location, Jupyter accordingly is not able to write user settings and crashes accordingly!

From my point of view, there is no reason for ROOT to set this environment variable!
- Setting `JUPYTER_PATH` is sufficient for the `ROOT C++` kernel to be picked up by Jupyter.

**In summary:**
- By setting `JUPYTER_CONFIG_DIR` ROOT makes it impossible for Jupyter to coexist with a read-only ROOT installation.
- **Proposed Fix:** The corresponding lines should be removed from `thisroot.sh` (see https://github.com/root-project/root/blob/10d7bd3b942a6bd61034116b91e26f88b70f3418/config/thisroot.sh#L160)
- The `ROOT C++` kernel seems to work fine, by just setting `JUPYTER_PATH` (based on my quick testing).

**Some relevant References:**
- Jupyter retrieves the `JUPYTER_CONFIG_DIR` env variable here: https://github.com/jupyter/jupyter_core/blob/7cb0a282fad0273a05c92d82d32a5de84eb8ad91/jupyter_core/paths.py#L40
- Its value is then used for creating the user settings location: https://github.com/jupyterlab/jupyterlab/blob/e2fd4c8841a7393f9131fb6b6e8252864a8bd351/jupyterlab/commands.py#L116

### Reproducer

**Basic Setup:**
 
- Install ROOT in a location on the system only writeable by the system root user (for example under `/opt/root/v6_36_00/`)
- Create a new python environment `python3 -m venv env`
- Activate the python environment `source env/bin/activate`
- Install Jupyter-Lab `pip install jupyterlab`

**Normal Jupyter Operation:**

- Start Jupyter Lab `jupyter lab`
  - Jupyter will start normally!
- Stop Jupyter Lab `<CTRL+C>`

**Trigger the Bug**

- Source ROOT: `source /opt/root/v6_36_00/bin/thisroot.sh`
- Start Jupyter Lab again  `jupyter lab`
  - **Jupyter Lab crashes with the following error**: `PermissionError: [Errno 13] Permission denied: '/opt/root/v6_36_00/etc/notebook/migrated'`

**Fix the Bug**

- Unset the JUPYTER_CONFIG_DIR variable `unset JUPYTER_CONFIG_DIR`
- Start Jupyter again `jupyter lab`
  - **Jupyter starts again normally**
  - ROOT notebooks are still available in Jupyter!

### ROOT version

`   ------------------------------------------------------------------
  | Welcome to ROOT 6.36.000                       https://root.cern |
  | Built for linuxx8664gcc on May 25 2025, 16:47:38                 |
  | From tags/v6-36-00@v6-36-00                                      |
  | With c++ (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0                   |
`

### Installation method

Pre-built Binary

### Operating system

Ubuntu 24.04.2 LTS

### Additional context

_No response_

deragent:
# This Pull request:

With the previous version, ROOT explicitely sets the `JUPYTER_CONFIG_DIR` to a folder part of the ROOT installation. In installation, where this folder is not writable by the user, this lead to Jupyter not starting as Jupyter wants to write user settings to files in this folder.

The revised version uses `JUPYTER_CONFIG_PATH` instead of `JUPYTER_CONFIG_DIR`. This has the effect, of adding the ROOT Jupyter config folder as an additional config, but not as the main config folder.

Thus, Jupyter can stil write to the default config folder, but the ROOT provided Jupyter config is still loaded (this is relevant for providing JsROOT locally, instead of loading it from root.cern).

## Changes or fixes:

Closes https://github.com/root-project/root/issues/19320

## Checklist:

- [X ] tested changes locally
- [ X] updated the docs (if necessary)

This PR fixes #19320

### DIFF
--- a/config/thisroot.csh
+++ b/config/thisroot.csh
@@ -154,8 +154,8 @@ if ($?old_rootsys) then
                                  -e "s;^$old_rootsys/etc/notebook:;;g"   \
                                  -e "s;^$old_rootsys/etc/notebook${DOLLAR};;g"`
    endif
-   if ($?JUPYTER_CONFIG_DIR) then
-      setenv JUPYTER_CONFIG_DIR `set DOLLAR='$'; echo $JUPYTER_CONFIG_DIR | \
+   if ($?JUPYTER_CONFIG_PATH) then
+      setenv JUPYTER_CONFIG_PATH `set DOLLAR='$'; echo $JUPYTER_CONFIG_PATH | \
                              sed -e "s;:$old_rootsys/etc/notebook:;:;g" \
                                  -e "s;:$old_rootsys/etc/notebook${DOLLAR};;g"   \
                                  -e "s;^$old_rootsys/etc/notebook:;;g"   \
@@ -226,14 +226,13 @@ else
    setenv JUPYTER_PATH ${ROOTSYS}/etc/notebook
 endif
 
-if ($?JUPYTER_CONFIG_DIR) then
-   setenv JUPYTER_CONFIG_DIR ${ROOTSYS}/etc/notebook:$JUPYTER_CONFIG_DIR
+if ($?JUPYTER_CONFIG_PATH) then
+   setenv JUPYTER_CONFIG_PATH ${ROOTSYS}/etc/notebook:$JUPYTER_CONFIG_PATH
 else
-   setenv JUPYTER_CONFIG_DIR ${ROOTSYS}/etc/notebook
+   setenv JUPYTER_CONFIG_PATH ${ROOTSYS}/etc/notebook
 endif
 
 endif # if ("$thisroot" != "")
 
 set thisroot=
 set old_rootsys=
-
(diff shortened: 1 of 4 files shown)


guitargeek (approved):
Thank you very much!

PR #19499, opened 2025-08-01
[build] Enable running wheel workflow on label

vepadulano:
Now that the workflow file is in the master branch, we should be able to run more triggers from PRs. Let's see

### DIFF
--- a/.github/workflows/python_wheel_build.yml
+++ b/.github/workflows/python_wheel_build.yml
@@ -8,13 +8,14 @@ on:
         type: string
         required: true
         default: "master"
-  push:
-    branches: experimental-pip-install-root
   schedule:
     - cron: '01 1 * * *'
+  pull_request:
+    types: [labeled]
 
 jobs:
   build-wheels:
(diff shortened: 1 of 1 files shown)


guitargeek (approved):
Thank you!

vepadulano:
There's an issue I don't understand with the part of the workflow that uploads the artifacts to EOS
https://github.com/root-project/root/actions/runs/16677771403/job/47214671568?pr=19499#step:7:25
```
kinit: Cannot find KDC for invalid realm name "" while getting initial credentials
```

Seems like the kerberos realm name is not evaluated properly and it is empty, even though the secret in general is clearly working. You can see the action properly running until completion in the last three nights https://github.com/root-project/root/actions/workflows/python_wheel_build.yml

vepadulano:
Unless it's another quirk of github actions that does not expose secrets to workflow modifications happening only in PRs, which could very well be...

vepadulano:
Yes, indeed https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

Issue #19327, opened 2025-07-10
Build Fails on MacOS (ARM, 15)


cjones051073:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

I am attempting to update the MacPorts build of root6 to the latest 6.36.02 release.

ROOT versions based on 6.34.X all build and run just fine.

Build fails with

```
Referenced from: <D677D28C-58A1-3164-A2D0-C88FCD9877C8> /opt/local/var/macports/build/root4Ivhdw/work/build/bin/rootcling
  Reason: tried: '/opt/local/lib/libgcc/libRIO.6.36.so' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/local/lib/libgcc/libRIO.6.36.so' (no such file), '/opt/local/lib/libRIO.6.36.so' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/local/lib/libRIO.6.36.so' (no such file), '/opt/local/var/macports/build/root4Ivhdw/work/build/bin/libRIO.6.36.so' (no such file), '/opt/local/var/macports/build/root4Ivhdw/work/build/bin/../lib/root/libRIO.6.36.so' (no such file), '/opt/local/lib/libgcc/libRIO.6.36.so' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/local/lib/libgcc/libRIO.6.36.so' (no such file), '/opt/local/lib/libRIO.6.36.so' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/local/lib/libRIO.6.36.so' (no such file), '/opt/local/var/macports/build/root4Ivhdw/work/build/bin/libRIO.6.36.so' (no such file), '/opt/local/var/macports/build/root4Ivhdw/work/build/bin/../lib/root/libRIO.6.36.so' (no such file)
```

Configuration used is

```
Executing:  cd "/opt/local/var/macports/build/root4Ivhdw/work/build" && /opt/local/bin/cmake -G "CodeBlocks - Unix Makef
iles" -DCMAKE_BUILD_TYPE=MacPorts -DCMAKE_INSTALL_PREFIX="/opt/local" -DCMAKE_INSTALL_NAME_DIR="/opt/local/lib" -DCMAKE_
[...]

PR #19500, opened 2025-08-01
[CMake] Append to CMAKE_INSTALL_RPATH without force-overriding cache.

hageboeck:
@guitargeek, I would like to suggest this small change for https://github.com/root-project/root/pull/19135.
It's preferable not to touch any cache variables, as they are user-settable (and can be altered after configuring). Given that RootBuildOptions is included at the outermost scope, it's sufficient to declare a new variable with the same name as the cache.

On each CMake run, this variable will be recomputed from the cache variable + whatever we append, leaving the CMakeCache.txt invariant.

I also took the opportunity to clarify what the `rpath` option does, since this was not really explained (see e.g. https://github.com/root-project/root/issues/19327).


guitargeek:
I also have an alternative solution that avoids messing around with the `CMAKE_INSTALL_RPATH` completely:
  * https://github.com/root-project/root/pull/19501

hageboeck:
Closing in favour of #19501

Issue #19479, opened 2025-07-31
Info() logs should not go to `stderr`


ntadej:
### Explain what you would like to see improved and how.

`Info()` logs should not go to `stderr` but `stdout` instead.

For example `hadd` now by default prints messages starting with `Info in <hadd>:` but they go into `stderr` which is unintuitive and confusing.

### ROOT version

ROOT 6.36.02 from `LCG_108` using GCC 15.1.0.

### Installation method

LCG_108

### Operating system

AlmaLinux 9

silverweed:
>  they go into stderr which is unintuitive and confusing

Could you elaborate why? Even though Info is not semantically an "error" I think it's still convenient to have all the logging end up in the same stream. 
It would be nice if this could be conveniently configured though (something like `SetLogStream(kInfo, stdout)`)

ntadej:
Alternatively you could make the verbosity level a bit more flexible. The problem are the new `Info` messages that appeared in 3.36 but were not there before. We'd like to keep those but not the new ones.

The deeper reason is that we wrap `hadd` in a python script where we pipe the whole `stderr` to `logger.error`. Of course we can handle this, but in general info messages are less critical and are often ignored by maybe piping to `/dev/null`. Now it's not possible.

silverweed:
> The problem are the new Info messages that appeared in 3.36 but were not there before.

Which 3.36 are you referring to? Also, could you please write here which messages are bothering you so I can check if they are needed and if we can gate them behind some verbosity settings?

Edit: I guess you refer to 6.36

ferdymercury:
A workaround might be to define your custom error handler. See https://root-forum.cern.ch/t/turning-warnings-into-errors/37872/4?u=ferhue

ntadej:
Sure, but our goal was to use `hadd` as CLI unchanged to avoid reimplementing it.

pcanal:
> The problem are the new Info messages that appeared in 3.36 but were not there before. We'd like to keep those but not the new ones.

Can you be more specific?  Info message are usually not printed.  Can you write here the full output and the part that you want to filter out?

ferdymercury:
Oh, I see, there is a custom logger in hadd that is not related to the default ROOT error handler:
https://github.com/root-project/root/blob/master/main/src/hadd.cxx#L177

ferdymercury:
> `Info` messages that appeared in 6.36 but were not there before. We'd like to keep those but not the new ones

[...]

PR #19503, opened 2025-08-01
[hadd] do not print info to stderr

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/19479

### DIFF
--- a/main/src/hadd.cxx
+++ b/main/src/hadd.cxx
@@ -176,8 +176,8 @@ inline std::ostream &Warn()
 
 inline std::ostream &Info()
 {
-   std::cerr << "Info in <hadd>: ";
-   return std::cerr;
+   std::cout << "Info in <hadd>: ";
+   return std::cout;
 }
 
 using IntFlag_t = uint32_t;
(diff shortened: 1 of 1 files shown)


silverweed:
I don't think we should print the Info() messages on stdout; I'd rather have an option to silence the output altogether if that's the desired outcome (we already have a verbosity option: we can simply not emit Info() if verbosity is less than 1 or something)

ferdymercury:
There was a change of behavior, in ROOT 6.32 hadd printed info messages to std::cout
In ROOT 6.36 they are printed to std::cerr, so I guess the user is surprised or confused by this change.
Maybe the easiest is to document this change in the release notes and close isssue as won't fix.

pcanal:
> close isssue as won't fix.

Before doing that, I would like to hear from the user what explicitly bother them and how, so we can clarify the situation from their point of view.

ferdymercury:
Closing in favor of https://github.com/root-project/root/pull/19707

PR #19504, opened 2025-08-01
[io] seek parent should point to mother dir, not to top level dir

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://its.cern.ch/jira/browse/ROOT-10352

## Checklist:

- [x] tested changes locally

### DIFF
--- a/io/io/test/TFileTests.cxx
+++ b/io/io/test/TFileTests.cxx
@@ -7,6 +7,7 @@
 #include "TFile.h"
 #include "TMemFile.h"
 #include "TDirectory.h"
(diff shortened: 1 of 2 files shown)


io/io/test/TFileTests.cxx
```diff
@@ -7,6 +7,7 @@
 #include "TFile.h"
 #include "TMemFile.h"
 #include "TDirectory.h"
+#include "TDirectoryFile.h"
```
ferdymercury:
```suggestion
```

PR #19509, opened 2025-08-02
[Python] Sync with `cppyy` and `CPyCppyy` master

guitargeek:
Bring in a new fix and unit test.

Also update the `sync-upstream` script.


vepadulano (approved):
Thanks!

PR #19515, opened 2025-08-04
[Python] Remove TString support patch to CPyCppyy

guitargeek:
Actually, `cppyy` can deal with TString arguments without explicit converters, via implicit conversions.

This is covered by numerous tests in `roottest`:

https://github.com/root-project/root/blob/master/roottest/python/basic/PyROOT_basictests.py#L155

The reason why we still kept the custom converters was that the implicit conversion was broken in some cases, which should be fixed now.

### DIFF
--- a/tmva/tmva/inc/TMVA/Factory.h
+++ b/tmva/tmva/inc/TMVA/Factory.h
@@ -96,9 +96,23 @@ namespace TMVA {
       // use TName::GetName and define correct name in constructor
       //virtual const char*  GetName() const { return "Factory"; }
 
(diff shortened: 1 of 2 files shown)


vepadulano (approved):
LGTM feel free to merge when the CI is green

guitargeek:
There are still other implicit conversion problems related to enums, as seen in the TMVA tutorials :( I'll try to get a minimal reproducer for that too and fix it in CPyCppyy, just like the previous case with the explicit constructors.

guitargeek:
Depends on:
  * https://github.com/root-project/root/issues/19517

PR #19516, opened 2025-08-04
[roottest] use cmake for all root/tree tests

linev:


pcanal:
Is there a way to split the PR by group of 2 or 3 directories?

linev:
> Is there a way to split the PR by group of 2 or 3 directories?

Yes, it is possible.
Each commit normally changes one sub-directory.
Plus I add extra commits when need to fix/prevent problem on Windows.

There are 33 commits. How much PRs you are want?

pcanal:
You leave it as is or you can split.  The difference is that it takes a bit to review each directories and the more there is the harder it is to keep track of which one I have reviewed.

linev:
@pcanal 
I create 6 separate PRs.
Let see if they will go through CI

linev:
This big PR splitted on #19522, #19523, #19524, #19525, #19526, #19527

PR #19522, opened 2025-08-04
[roottest] root/tree, part 1

linev:


roottest/root/tree/split/CMakeLists.txt
```diff
@@ -1,7 +1,38 @@
...
-#
-#-------------------------------------------------------------------------------
-ROOTTEST_ADD_OLDTEST()
+ROOTTEST_COMPILE_MACRO(abstract.cxx
+                       FIXTURES_SETUP root-tree-split-abstract-fixture)
+
+ROOTTEST_ADD_TEST(abstract-write
+                  MACRO runabstract.cxx
+                  MACROARG 0
+                  COPY_TO_BUILDDIR abstract.cxx
```
pcanal: is `COPY_TO_BUILDDIR` updating the timestamp.  If it does, I think this will lead to the `ROOTTEST_COMPILE_MACRO(abstract.cxx ..)` to be ignored/re-done at `runabstract.cxx` execution (as it load the macro (as intended) via `ACLiC` which will check the timestamps.

linev: `ROOTTEST_COMPILE_MACRO` always uses macro from source directory.
Extra `COPY_TO_BUILDDIR abstract.cxx` required by `runabstract.cxx` which uses `#include "abstract.cxx+"`
Instead one can try to add `${CMAKE_CURRENT_SOURCE_DIR}` as include directory for running `runabstract.cxx`.

pcanal: > ROOTTEST_COMPILE_MACRO always uses macro from source directory.
> Extra COPY_TO_BUILDDIR abstract.cxx required by runabstract.cxx which uses #include "abstract.cxx+"

Yes this is exactly what I saw.

> Instead one can try to add ${CMAKE_CURRENT_SOURCE_DIR} as include directory for running runabstract.cxx.

Another alternative (I don't know if we can) is to copy the abstract.cxx as part or before the COMPILE_MACRO.

linev: But I see no real need of this - test need to run once

pcanal: Fair but then let's just remove the (seemingly) unneeded `COMPILE_MACRO`.

linev: In a `Makefile` running of the test requires compilation of `abstract.cxx` macro first

pcanal: I am confused.  As far as I can tell (by running with the makefile) In the `Makefile`, the library was built during the write phase (`runabstract.cxx(0)`) and this was not a problem because the output of that step was not compared to a reference file.

In this CMakeFile version, the write phase is also not compared to a reference file, so I am not sure how it fails without `COMPILE_MACRO`.  

My worries was whether `COPY_TO_BUILDDIR` preserved the timestamp (i.e. doing is `cp -p ... `) or not.  If not then we would be build the library twice (at least in some cases - and in those case if the library was really needing to be pre-built, it would have been a problem).

However, I checked and at least on my machine, the `COPY_TO_BUILDDIR` will update the timestamp.  And by putting `ctest` to sleep for a little bit, I saw a rebuild during the execution of the write.
```
$ cat roottest/root/tree/split/abstract-write.log 

Processing /uscms_data/d2/pcanal/eaf_root_working/code/quick-devel/roottest/root/tree/split/runabstract.cxx(0)...
Info in <TUnixSystem::ACLiC>: creating shared library /home/pcanal/root_working/build/quick-devel/roottest/root/tree/split/abstract_cxx.so
(int) 0
```

But since this log is not checked, this is has not effect (except for the very slight inefficiency of the 2 builds).  Feel free to resolve as is if you prefer.

pcanal (approved):
LGTM

roottest/root/tree/split/CMakeLists.txt
linev: I add extra step which copies macro and only then compile it.

PR #19528, opened 2025-08-04
[ntuple] Fixes / Changes for or related to `RField*` and `RValue`

hahnjo:
see individual commits

### DIFF
--- a/tree/ntuple/inc/ROOT/RField.hxx
+++ b/tree/ntuple/inc/ROOT/RField.hxx
@@ -498,8 +498,7 @@ public:
    void AcceptVisitor(ROOT::Detail::RFieldVisitor &visitor) const final;
 };
 
-// Has to be implemented after the definition of all RField<T> types
-// The void type is specialized in RField.cxx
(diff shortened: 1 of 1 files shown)


tree/ntuple/inc/ROOT/RFieldBase.hxx
```diff
@@ -713,9 +715,13 @@ public:
    RValue &operator=(RValue &&other) = default;
    ~RValue() = default;
 
+private:
    std::size_t Append() { return fField->Append(fObjPtr.get()); }
+
+public:
    void Read(ROOT::NTupleSize_t globalIndex) { fField->Read(globalIndex, fObjPtr.get()); }
    void Read(RNTupleLocalIndex localIndex) { fField->Read(localIndex, fObjPtr.get()); }
```
hahnjo, line 723: @jblomer we discussed making the `Read` methods `private` as well. This would be asymmetric to `RBulkValues`, so in the end I decided against it (for the moment)

jblomer (approved):

PR #19529, opened 2025-08-04
[RF] Avoid expensive double loop in `setData()` for likelihooods

guitargeek:
About 2 years ago in 85c5cb4, I made the mistake of introducing a potentially expensive double loop when setting the data for a likelihood: the `RooEvaluatorWapper` iterates over all data variables to set, and for each of them it calls `RooFit::Evaluator::setInput()`, which has a loop over all nodes in the computation graph to identify for which one the data is set.

This caused a big performance hit when instantiating likelihoods with the new vectorizing CPU backend, manifesting in 10 minutes for `createNLL` for the big ATLAS Higgs combination benchmark instead of 1 min.

This commit fixes the problem by caching all nodes in the compute graph in a hash map for fast lookup.


lmoneta (approved):
LGTM!
Thank you Jonas for fixing this!

PR #19532, opened 2025-08-05
[ntuple] fix unit test linking

jblomer:
Fix linking of ntuple_descriptor and ntuple_types unit tests on some platforms (e.g., Arch). Also, remove custom struct library from several unit tests where it is not needed.


silverweed:
> LGTM, but why wasn't this caught before?

The problem appeared on some local machines some time ago (including mine and I believe Jakob's), but not in the CI as we are not testing on these platforms.

PR #19535, opened 2025-08-05
[tree] FindBranch do not remove brackets if candidate name has them

ferdymercury:
Fixes https://its.cern.ch/jira/browse/ROOT-5567


dpiparo:
Thanks a lot for this fix!

PR #19536, opened 2025-08-05
[ntuple] Properly handle unknown columns in CreateModel

### LINKED ISSUE #19533: [ntuple] Unknown column types should throw
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

... as tested by `RNTupleCompat.FutureColumnType`. However, the test has an issue because the custom `RField` specialization reports `FutureColumn` as type name, which is not a known C++ type. Consequently, model recreation fails with "unknown type: FutureColumn" from `RFieldBase::Create`. Fixing the field type (see below) does not throw an exception and fails the test(s).

### Reproducer

```diff
diff --git a/tree/ntuple/test/ntuple_compat.cxx b/tree/ntuple/test/ntuple_compat.cxx
index a6c9e971af8..5c4edeb6231 100644
--- a/tree/ntuple/test/ntuple_compat.cxx
+++ b/tree/ntuple/test/ntuple_compat.cxx
@@ -166,7 +166,7 @@ protected:
    }
 
 public:
-   static std::string TypeName() { return "FutureColumn"; }
+   static std::string TypeName() { return "ROOT::Internal::RTestFutureColumn"; }
    explicit RField(std::string_view name) : RSimpleField(name, TypeName()) {}
    RField(RField &&other) = default;
    RField &operator=(RField &&other) = default;
```

### ROOT version

`master`, but probably also 6.34 and 6.36

### Installation method

from source

### Operating system

Linux

### Additional context

_No response_

silverweed:
# This Pull request:
makes RNTupleDescriptor::CreateModel properly handle unknown column types.

## Changes or fixes:
- check all columns at the start of CreateModel to check for unknown column types. If any are found, we normally throw, but if ForwardCompatible is true we use the information to skip the invalid fields.
- improve the checks done in ntuple_compat to see that we're throwing for the right reason.

## Checklist:

- [x] tested changes locally

This PR fixes #19533

### DIFF
--- a/tree/ntuple/src/RNTupleDescriptor.cxx
+++ b/tree/ntuple/src/RNTupleDescriptor.cxx
@@ -88,6 +88,7 @@ ROOT::RFieldDescriptor::CreateField(const RNTupleDescriptor &ntplDesc, const ROO
       }
    }
 
+   // Untyped records and collections
    if (GetTypeName().empty()) {
       switch (GetStructure()) {
       case ROOT::ENTupleStructure::kRecord: {
@@ -139,7 +140,7 @@ ROOT::RFieldDescriptor::CreateField(const RNTupleDescriptor &ntplDesc, const ROO
       }
 
       return field;
-   } catch (RException &ex) {
+   } catch (const RException &ex) {
       if (options.GetReturnInvalidOnError())
          return std::make_unique<ROOT::RInvalidField>(GetFieldName(), GetTypeName(), ex.GetError().GetReport(),
                                                       ROOT::RInvalidField::RCategory::kGeneric);
@@ -640,6 +641,31 @@ ROOT::RResult<void> ROOT::RNTupleDescriptor::DropClusterGroupDetails(ROOT::Descr
 
 std::unique_ptr<ROOT::RNTupleModel> ROOT::RNTupleDescriptor::CreateModel(const RCreateModelOptions &options) const
 {
+   // Collect all top-level fields that have invalid columns (recursively): by default if we find any we throw an
+   // exception; if we are in ForwardCompatible mode, we proceed but skip of all those top-level fields.
+   std::unordered_set<ROOT::DescriptorId_t> invalidFields;
+   for (const auto &colDesc : GetColumnIterable()) {
+      if (colDesc.GetType() == ROOT::ENTupleColumnType::kUnknown) {
+         auto fieldId = colDesc.GetFieldId();
+         while (1) {
+            const auto &field = GetFieldDescriptor(fieldId);
+            if (field.GetParentId() == GetFieldZeroId())
+               break;
+            fieldId = field.GetParentId();
+         }
+         invalidFields.insert(fieldId);
+
+         // No need to look for all invalid fields if we're gonna error out anyway
+         if (!options.GetForwardCompatible())
+            break;
+      }
+   }
+
+   if (!options.GetForwardCompatible() && !invalidFields.empty())
+      throw ROOT::RException(R__FAIL(
+         "cannot create Model: descriptor contains unknown column types. Use 'SetForwardCompatible(true)' on the "
+         "RCreateModelOptions to create a partial model containing only the fields made up by known columns."));
+
    auto fieldZero = std::make_unique<ROOT::RFieldZero>();
    fieldZero->SetOnDiskId(GetFieldZeroId());
    auto model = options.GetCreateBare() ? RNTupleModel::CreateBare(std::move(fieldZero))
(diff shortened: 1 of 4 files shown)


tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -648,9 +674,27 @@ std::unique_ptr<ROOT::RNTupleModel> ROOT::RNTupleDescriptor::CreateModel(const R
...
-      if (field->GetTraits() & ROOT::RFieldBase::kTraitInvalidField)
+
+      // If we got an InvalidField here, figure out if it's a hard error or if the field must simply be skipped.
+      // The only case where it's not a hard error is if the field has an unknown structure, as that case is
+      // covered by the ForwardCompatible flag (note that if the flag is off we would not get here
+      // in the first place, so we don't need to check for that flag again).
+      if (field->GetTraits() & ROOT::RFieldBase::kTraitInvalidField) {
+         const auto &invalid = static_cast<const RInvalidField &>(*field);
+         const auto cat = invalid.GetCategory();
+         bool mustThrow = cat != RInvalidField::RCategory::kUnknownStructure;
```
jblomer, line 691: Can't we skip fields with unknown/invalid types, too?

silverweed: I don't think we should, because those do not depend on the binary format so they would not fall under the "forward compatibility" umbrella...

jblomer: Let's assume we add support for `std::chrono::timepoint`, e.g. as a new leaf field with a 64bit column. How can old readers be configured to read everything but fields with that new type?

jblomer: Actually, perhaps this case and others like it are indeed better handled under the type emulation umbrella. I think we just need to generalize it a bit. Feel free to resolve this thread.

jblomer (approved):

tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -640,6 +641,31 @@ ROOT::RResult<void> ROOT::RNTupleDescriptor::DropClusterGroupDetails(ROOT::Descr
 
 std::unique_ptr<ROOT::RNTupleModel> ROOT::RNTupleDescriptor::CreateModel(const RCreateModelOptions &options) const
 {
+   // Collect all top-level fields that have invalid columns (recursively): by default if we find any we throw an
+   // exception; if we are in ForwardCompatible mode, we proceed but skip of all those top-level fields.
+   std::vector<ROOT::DescriptorId_t> invalidFields;
```
hahnjo: Can't a top-level field have multiple columns with unknown column type below it? Then we should probably avoid duplicates in `invalidFields`...

tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -640,6 +641,31 @@ ROOT::RResult<void> ROOT::RNTupleDescriptor::DropClusterGroupDetails(ROOT::Descr
...
+   for (const auto &colDesc : GetColumnIterable()) {
+      if (colDesc.GetType() == ROOT::ENTupleColumnType::kUnknown) {
+         auto fieldId = colDesc.GetFieldId();
+         while (1) {
+            const auto &field = GetFieldDescriptor(fieldId);
+            if (field.GetParentId() == GetFieldZeroId() || R__unlikely(field.GetParentId() == kInvalidDescriptorId))
```
hahnjo: When is the parent Id `kInvalidDescriptorId`? Should this be a `R__ASSERT`?

tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -648,9 +674,27 @@ std::unique_ptr<ROOT::RNTupleModel> ROOT::RNTupleDescriptor::CreateModel(const R
...
+      if (std::find(invalidFields.begin(), invalidFields.end(), topDesc.GetId()) != invalidFields.end()) {
+         // Field contains invalid columns: skip it
+         continue;
+      }
+
       auto field = topDesc.CreateField(*this, createFieldOpts);
-      if (field->GetTraits() & ROOT::RFieldBase::kTraitInvalidField)
+
+      // If we got an InvalidField here, figure out if it's a hard error or if the field must simply be skipped.
+      // The only case where it's not a hard error is if the field has an unknown structure, as that case is
```
hahnjo, line 685: side remark: do we check the structural role of non-top-level fields? During model reconstruction, if I remember correctly, we will call `RFieldDescriptor::CreateField` for the top-level field, but the subfields are created according to the current data model. They would only be checked when connecting, with https://github.com/root-project/root/pull/19539...

tree/ntuple/src/RNTupleDescriptor.cxx
silverweed: I'll just remove the check as we will throw in GetFieldDescriptor if we find an invalid id anyway

tree/ntuple/src/RNTupleDescriptor.cxx
silverweed: I'm using a vector as I expect it to usually contain very few elements, but I can replace it with a set for pathological cases

tree/ntuple/src/RNTupleDescriptor.cxx
silverweed: I'm not sure; maybe we should do the check directly in `RFieldBase::Create` (when we have the descriptor)?

hahnjo:
Do we want this fix potentially backported to v6.36, given that it affects forward-compatibility?

PR #19537, opened 2025-08-05
[RF] Make RooFit::MsgTopic enum name for numeric integration consistent

### LINKED ISSUE #19422: Make `RooFit::MsgTopic` enum name for numeric integration consistent

For numeric integration, the name of the enum is different than the name returned by the message service

https://github.com/root-project/root/blob/b41273e1be7164c39ab2abb4c9c4a8bdb8fc1384/roofit/roofitcore/src/RooMsgService.cxx#L103-L105

This leads to the confusing scenario where one can not use the name returned by `RooMsgService.instance().Print()` to add the numeric integration topic to a debug stream

example:
```python
import ROOT

ROOT.RooMsgService.instance().Print()
# Active Message streams
# [0] MinLevel = PROGRESS Topic = Generation Minimization Plotting Fitting Integration LinkStateMgmt Eval Caching  Optimization ObjectHandling InputArguments Tracing Contents DataHandling NumericIntegration FastEvaluations 
# [1] MinLevel = INFO Topic = Minimization Plotting Fitting Eval Caching ObjectHandling InputArguments DataHandling NumericIntegration 
# [2] MinLevel = INFO Topic = HistFactory 
ROOT.RooMsgService.instance().addStream(ROOT.RooFit.DEBUG, Topic=ROOT.RooFit.NumericIntegration)
# AttributeError: <namespace cppyy.gbl.RooFit at 0x7ea2790> has no attribute 'NumericIntegration'. Full details:
#   type object 'RooFit' has no attribute 'NumericIntegration'
#   'RooFit::NumericIntegration' is not a known C++ class
#   'NumericIntegration' is not a known C++ template
#   'NumericIntegration' is not a known C++ enum
ROOT.RooMsgService.instance().addStream(ROOT.RooFit.DEBUG, Topic=ROOT.RooFit.NumIntegration)
# works
```

An easy way out would be to change the name to also just be `NumIntegration` but I think the more verbose one is more user friendly. Changing the enum is probably a bit though as it has been like this forever now, but as it is just a number in the end it would be possible to just add another member `RooFit.NumericIntegration` with the same value to the namespace and maybe even to add a deprecation notice for the old one?

### ROOT version

any

### Installation method

any

### Operating system

any

### Additional context

_No response_

guitargeek:
Use the new enum name in all the RooFit code and tutorials, and preserve an alias with the old name for full backwards compatibility.

Closes #19422.

### DIFF
--- a/roofit/histfactory/test/testHistFactory.cxx
+++ b/roofit/histfactory/test/testHistFactory.cxx
@@ -86,7 +86,7 @@ TEST(Sample, CopyAssignment)
 
 TEST(HistFactory, Read_ROOT6_16_Model)
 {
-   RooHelpers::LocalChangeMsgLevel chmsglvl{RooFit::WARNING, 0u, RooFit::NumIntegration, true};
+   RooHelpers::LocalChangeMsgLevel chmsglvl{RooFit::WARNING, 0u, RooFit::NumericIntegration, true};
 
    std::string filename = "./ref_6.16_example_UsingC_channel1_meas_model.root";
    std::unique_ptr<TFile> file(TFile::Open(filename.c_str()));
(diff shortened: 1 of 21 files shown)


lmoneta (approved):
LGTM!
Thanks for fixing this issue.

PR #19539, opened 2025-08-05
[ntuple] Enforce field compatibility rules

### LINKED ISSUE #19256: [ntuple] Schema evolution of repetitive fields
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

It appears repetitive fields currently don't check against the on-disk metadata.

### Reproducer

```c++

#include <array>
#include <bitset>
#include <cstdint>

void ntuple_array_bitset() {
  {
    auto arr = model->MakeField<std::array<std::int32_t, 3>>("arr");
    auto bs = model->MakeField<std::bitset<3>>("bs");
    auto writer = ROOT::RNTupleWriter::Recreate(std::move(model), "ntpl", "ntuple_array_bitset.root");
    *arr = {1, 2, 3};
    *bs = 5;
    writer->Fill();
    writer->Fill();
  }

  {
    auto arr = model->MakeField<std::array<std::int32_t, 5>>("arr");
    auto bs = model->MakeField<std::bitset<5>>("bs");
    auto reader = ROOT::RNTupleReader::Open(std::move(model), "ntpl", "ntuple_array_bitset.root");
    reader->LoadEntry(0);
    std::cout << "0: arr = [" << arr->at(0) << ", " << arr->at(1) << ", " << arr->at(2) << ", " << arr->at(3) << ", " << arr->at(4) << "], bs = " << bs->to_ulong() << "\n";
    reader->LoadEntry(1);
  }
}
```

Output:
```
Processing ntuple_array_bitset.C...
0: arr = [1, 2, 3, 1, 2], bs = 13
  what():  entry with index 6 out of bounds
At:
  virtual ROOT::Internal::RPageRef ROOT::Internal::RPageSource::LoadPage(ColumnHandle_t, ROOT::NTupleSize_t)
```
(showing that the elements were mixed across entries)

### ROOT version

`master`

### Installation method

from source

### Operating system

Linux

### Additional context

_No response_

jblomer:
Systematically check compatibility of in-memory and on-disk field when connecting a page source.

The current state of the PR adds the basic infrastructure. A number of checks in derived classes are still too relaxed. The exact details of automatic transformations will be carved out in follow-up PRs.

Fixes  #19256.


tree/ntuple/inc/ROOT/RFieldBase.hxx
```diff
@@ -510,8 +510,13 @@ protected:
...
    /// and not backed by on-disk information.
-   /// By default the implementation will fail if the on-disk ID is set and CompareOnDisk() reports any differences.
-   virtual void BeforeConnectPageSource(ROOT::Internal::RPageSource &) {}
+   virtual void BeforeConnectPageSource(ROOT::Internal::RPageSource &source)
+   {
+      // The default implementation throws an exception if the on-disk ID is set and there are any differences
+      // to the on-disk field. Derived classes may overwrite this and losen the checks to support automatic schema
+      // evolution.
+      EnsureCompatibleOnDiskField(source);
+   }
```
hahnjo: Maybe outline to `RFieldBase.cxx`?

tree/ntuple/src/RField.cxx
```diff
@@ -783,6 +801,17 @@ std::size_t ROOT::RNullableField::AppendValue(const void *from)
...
 
+void ROOT::RNullableField::BeforeConnectPageSource(Internal::RPageSource &source)
+{
+   if (IsArtificial())
+      return;
+
+   const auto descriptorGuard = source.GetSharedDescriptorGuard();
+   const auto &fieldDesc = descriptorGuard->GetFieldDescriptor(GetOnDiskId());
+   EnsureCompatibleOnDiskField(fieldDesc, kDiffTypeName);
+   EnsureCompatibleTypePrefix(fieldDesc, {"std::optional<", "std::unique_ptr<"});
```
hahnjo: Maybe make the prefixes a `static std::vector<std::string>` so they don't need to be constructed on every invocation?

tree/ntuple/src/RField.cxx
```diff
@@ -1003,6 +1032,17 @@ std::vector<ROOT::RFieldBase::RValue> ROOT::RAtomicField::SplitValue(const RValu
...
 
+void ROOT::RAtomicField::BeforeConnectPageSource(Internal::RPageSource &source)
+{
+   if (IsArtificial())
+      return;
+
+   const auto descriptorGuard = source.GetSharedDescriptorGuard();
+   const auto &fieldDesc = descriptorGuard->GetFieldDescriptor(GetOnDiskId());
+   EnsureCompatibleOnDiskField(fieldDesc, kDiffTypeName);
+   EnsureCompatibleTypePrefix(fieldDesc, {"std::atomic<"});
```
hahnjo: same

tree/ntuple/src/RFieldBase.cxx
```diff
@@ -1031,6 +1032,61 @@ void ROOT::RFieldBase::ConnectPageSource(ROOT::Internal::RPageSource &pageSource
...
+{
+   std::uint32_t diffBits = 0;
+   if (GetFieldVersion() != fieldDesc.GetFieldVersion())
+      diffBits |= kDiffFieldVersion;
+   if (GetTypeVersion() != fieldDesc.GetTypeVersion())
+      diffBits |= kDiffTypeVersion;
+   if (GetStructure() != fieldDesc.GetStructure())
+      diffBits |= kDiffStructure;
+   if (GetTypeName() != fieldDesc.GetTypeName())
+      diffBits |= kDiffTypeName;
```
hahnjo: A (minor) inefficiency is that we first compare the type names in all cases and then discard the result depending on `ignoreBits` above. Not sure if actually relevant in practice...

tree/ntuple/inc/ROOT/RField/RFieldProxiedCollection.hxx
```diff
@@ -168,6 +168,11 @@ protected:
    std::size_t AppendImpl(const void *from) final;
    void ReadGlobalImpl(ROOT::NTupleSize_t globalIndex, void *to) final;
 
+   void BeforeConnectPageSource(Internal::RPageSource &source) final
+   {
+      EnsureCompatibleOnDiskField(source, kDiffTypeName);
```
hahnjo: Hm, I wanted to comment that "we may need to ignore `kDiffTypeVersion` as well", but it appears we are not storing the type version of a proxied collection field. Is that correct?

tree/ntuple/inc/ROOT/RField/RFieldRecord.hxx
```diff
@@ -101,6 +101,11 @@ protected:
       fSize += GetItemPadding(fSize, fMaxAlignment);
    }
 
+   void BeforeConnectPageSource(ROOT::Internal::RPageSource &source) override
+   {
+      EnsureCompatibleOnDiskField(source, kDiffTypeName | kDiffTypeVersion);
```
hahnjo: What is the type version of a record field?

tree/ntuple/src/RFieldSequenceContainer.cxx
```diff
@@ -85,6 +85,17 @@ void ROOT::RArrayField::ConstructValue(void *where) const
...
 
+void ROOT::RArrayField::BeforeConnectPageSource(Internal::RPageSource &source)
+{
+   if (IsArtificial())
+      return;
+
+   const auto descriptorGuard = source.GetSharedDescriptorGuard();
+   const auto &fieldDesc = descriptorGuard->GetFieldDescriptor(GetOnDiskId());
+   EnsureCompatibleOnDiskField(fieldDesc, kDiffTypeName);
+   EnsureCompatibleTypePrefix(fieldDesc, {"std::array<"});
```
hahnjo: `static std::vector<std::string>`?

tree/ntuple/src/RFieldSequenceContainer.cxx
```diff
@@ -695,6 +706,17 @@ std::vector<ROOT::RFieldBase::RValue> ROOT::RField<std::vector<bool>>::SplitValu
...
 
+void ROOT::RField<std::vector<bool>>::BeforeConnectPageSource(Internal::RPageSource &source)
+{
+   if (IsArtificial())
+      return;
+
+   const auto descriptorGuard = source.GetSharedDescriptorGuard();
+   const auto &fieldDesc = descriptorGuard->GetFieldDescriptor(GetOnDiskId());
+   EnsureCompatibleOnDiskField(fieldDesc, kDiffTypeName);
+   EnsureCompatibleTypePrefix(fieldDesc, {"std::vector<"});
```
hahnjo: `static std::vector<std::string>`?

pcanal:
> On the other hand, all field types that actually need to do something elaborate in BeforeConnectPageSource probably also have custom compatibility constraints, so probably it's fine...

Except maybe that the developer overloading `BeforeConnectPageSource` may or may not remember to also ensure type compatibility ...

tree/ntuple/inc/ROOT/RField/RFieldProxiedCollection.hxx
pcanal: > but it appears we are not storing the type version of a proxied collection field. Is that correct?

I think that would be a 'bug' as not knowing the 'version' or 'checksum' of the content of a collection (that has a proxy) would lead to not be able to schema evolve it.

tree/ntuple/src/RFieldBase.cxx
```diff
@@ -1028,11 +1029,74 @@ void ROOT::RFieldBase::ConnectPageSource(ROOT::Internal::RPageSource &pageSource
...
+{
+   // For computed fields added by schema evolution (e.g., new class member, new base class,
+   // field targeted by I/O customization rule), ignore on-disk information.
+   if (!fIsArtificial) {
+      const auto descriptorGuard = source.GetSharedDescriptorGuard();
+      EnsureCompatibleOnDiskField(descriptorGuard->GetFieldDescriptor(GetOnDiskId()), ignoreBits);
+   }
+}
+
+void ROOT::RFieldBase::EnsureCompatibleOnDiskField(const RFieldDescriptor &fieldDesc, std::uint32_t ignoreBits) const
```
pcanal: Unless I am mistaken, this function currently/for-now does not quite live up to his name (rather than 'compatible with' we have (for now) 'strict' equality, should we document the current state (and/or update the name if this is the long term intent)?

tree/ntuple/inc/ROOT/RField/RFieldProxiedCollection.hxx
hahnjo: Note that `RProxiedCollectionField` is the collection parent field. The item field with the actual content is (most of the times) a regular `RClassField` and schema evolution will work fine. The question is whether there is need to somehow "schema evolve" the collection field itself. We don't have it for `std::vector`, `RVec` etc, but I'm not sufficiently familiar with proxied collections to know what is correct. In any case, not really related to this PR, so please feel free to resolve @jblomer

pcanal: In streamed I/O, we do record a version number of the collection itself (the same number for all collections) which is recording the way the collection is streamed, in particular whether its content is streamed memberwise or objectwise.  If I recall correctly we have used this information (beside memberwise/objectwise) to evolve the file format.  So I would venture that similarly for `RNTuple` a change in the way collections are stored would be a file format change/evolution.

jblomer: For this purpose, we have the _field version_ in RNTuple, which is distinct from the _type version_.

tree/ntuple/inc/ROOT/RField/RFieldRecord.hxx
jblomer: Zero on write, the corresponding class version when used for emulated classes.

tree/ntuple/src/RFieldBase.cxx
jblomer: It's not a strict equality. We don't require, e.g., the field description nor the type alias to be identical.

jblomer:
The `BeforeConnectPageSource()` method has gotten a spin-off, `ReconcileOnDiskField()`, to give the field compatibility aspect a more targeted place.

tree/ntuple/src/RFieldBase.cxx
pcanal: Humm ... we usually consider the field description/type comment to be not relevant to the I/O, so fine but all the rest is strict equality (as far as I can read `CompareOnDiskField`) even the type alias is still semantically strict equality of types (it is 'just' a different spelling of the same normalized name pointing to the same semantic type [The non-equality or compatibility of type[name] in the existing ROOT I/O code correspond semantically to a 'renaming I/O customization rule' or a an implicit conversion in case of STL collection: `std::vector<short>` -> `std::vector<long>` ]

jblomer: Renamed `Compatible` to `Matching`

tree/ntuple/src/RFieldSequenceContainer.cxx
```diff
@@ -74,6 +74,15 @@ void ROOT::RArrayField::ReadInClusterImpl(RNTupleLocalIndex localIndex, void *to
    }
 }
 
+void ROOT::RArrayField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
+{
+   static const std::vector<std::string> prefixes = {"std::array<"};
+
+   const auto &fieldDesc = desc.GetFieldDescriptor(GetOnDiskId());
+   EnsureMatchingOnDiskField(fieldDesc, kDiffTypeName);
+   EnsureMatchingTypePrefix(fieldDesc, {"std::array<"});
```
hahnjo:
```suggestion
   EnsureMatchingTypePrefix(fieldDesc, prefixes);
```

tree/ntuple/src/RField.cxx
```diff
@@ -59,6 +59,20 @@ void ROOT::RCardinalityField::GenerateColumns(const ROOT::RNTupleDescriptor &des
    GenerateColumnsImpl<ROOT::Internal::RColumnIndex>(desc);
 }
 
+void ROOT::RCardinalityField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
+{
+   const auto &fieldDesc = desc.GetFieldDescriptor(GetOnDiskId());
+   EnsureMatchingOnDiskField(fieldDesc, kDiffTypeVersion | kDiffStructure | kDiffTypeName);
+   if (fieldDesc.GetStructure() == ENTupleStructure::kLeaf) {
```
hahnjo:
```suggestion
   if (fieldDesc.GetStructure() == ENTupleStructure::kPlain) {
```
(after rebasing)

tree/ntuple/test/ntuple_evolution_type.cxx
```diff
@@ -0,0 +1,124 @@
...
+   std::unique_ptr<ROOT::RFieldBase> CloneImpl(std::string_view newName) const final
+   {
+      return std::make_unique<RNewStringField>(newName, GetTypeName(), fFieldVersion, fTypeVersion);
+   }
+   void ConstructValue(void *where) const final { new (where) std::string(); }
+   std::unique_ptr<RDeleter> GetDeleter() const final { return std::make_unique<RTypedDeleter<std::string>>(); }
+
+public:
+   RNewStringField(std::string_view name, std::string_view type, std::uint32_t fieldVersion, std::uint32_t typeVersion)
+      : ROOT::RFieldBase(name, type, ROOT::ENTupleStructure::kLeaf, false /* isSimple */),
```
hahnjo:
```suggestion
      : ROOT::RFieldBase(name, type, ROOT::ENTupleStructure::kPlain, false /* isSimple */),
```
(after rebasing)

hahnjo (approved):
LGTM, one minor comment and two notes to change `kLeaf` to `kPlain`.

tree/ntuple/src/RFieldBase.cxx
jblomer: @pcanal can this be resolved?

pcanal: yes

pcanal (approved):
Thanks.

PR #19541, opened 2025-08-05
[roottest] use cmake for root/treeformula/array tests

linev:
Several tests need to be adjust to let compile macros before starting tests


roottest/root/treeformula/array/CMakeLists.txt
```diff
@@ -1,7 +1,87 @@
...
-# Placeholder file to translate the tests to the new CTest system. Meanwhile we
-# define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
-#
-#-------------------------------------------------------------------------------
-ROOTTEST_ADD_OLDTEST(LABELS longtest)
+ROOTTEST_COMPILE_MACRO(Data.cxx
+                       FIXTURES_SETUP root-treeformula-array-Data_cxx-fixture)
+
+ROOTTEST_ADD_TEST(filltree
+                  MACRO fillTree.cxx
```
pcanal, line 5: Side note: in previous version fillTree.cxx was used compiled, however it does not appear to be important to the test.

pcanal, line 5: For the record:
> do not compile fillTree.cxx macro
> On Windows this requires generation of dictionary for classes from Data.h. Which is duplication of the code.

Actually, all that is needed is that `fillTree.cxx` is compiled with `Data_cxx.so/dll` loaded (as was previously done by `runFillTree.C`).  With `ROOTTEST_COMPILE_MACRO` this can be done with `BUILDLIB Data_cxx`, however, as mentioned above, `fillTree.cxx` does not actually need to be compiled.

pcanal:
Side note, the test `nonsplit.C` which was already not run with the Makefile can not be enable because of https://github.com/root-project/root/issues/19687

PR #19543, opened 2025-08-05
[CI] Test installed headers for include errors

hageboeck:
Add a new step to the CI. In this step, every installed header is checked for syntax errors.
Some headers which are installed are exempt, for example from TMVA, vdt, bvh, and single instances such as headers that aren't supposed to be used standalone. Minimising the suppressions and fixing the corresponding headers will be moved into a separate issue.

The test works as follows (as done in https://github.com/xrootd/xrootd by @amadio): the compiler is invoked to only run the syntax checks. If the header doesn't include all headers it depends on, this will provoke errors. This means that this header works by accident, because it's included in another TU that was already including the relevant dependencies.
This is less aggressive than IncludeWhatYouUse, but still validates that all headers can be included on their own.

Note, some commits were split off to other PRs:
- #23242
- *Some more in previous PRs that I didn't find right now*


geom/geom/inc/TGeoTypedefs.h
```diff
@@ -15,6 +15,14 @@
...
 #include <vector>
 
+namespace ROOT::Geom {
+struct Vertex_t;
+}
+
+namespace ROOT::Geom {
+struct Vertex_t;
+}
+
```
guitargeek:
```suggestion
```
Copy-paste artefact?

cmake/modules/RootMacros.cmake
```diff
@@ -1223,60 +1223,69 @@ function(ROOT_FIND_DIRS_WITH_HEADERS result_dirs)
...
+
+  if(ARG_HEADERS)
+    # Headers have been listed explicitly, find them one by one
+    foreach(regex ${ARG_FILTER} "LinkDef")
+      list(FILTER ARG_HEADERS EXCLUDE REGEX "${regex}")
+    endforeach()
+    foreach(header ${ARG_HEADERS})
+      file(GLOB globbed_header ${header} */${header})
+      if(globbed_header STREQUAL "")
+	message(SEND_ERROR "No header corresponding to ${header} found in ${CMAKE_CURRENT_SOURCE_DIR}")
```
```suggestion
	     message(SEND_ERROR "No header corresponding to ${header} found in ${CMAKE_CURRENT_SOURCE_DIR}")
```
Indentation level seems off

guitargeek (approved):
LGTM! Just two minor poiishes that I'd suggest.

Also, maybe as a followup, what about running this in our nice post-install ctest battery (which you also introduced)? Then it's easier to do these checks also outside the context of the CI, which would be nice for packages and local builds.

test/PostInstall/check-headers.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
guitargeek:
```suggestion
#!/usr/bin/env bash
```
The usual comment from the NixOS user, where `/bin/bash` doesn't exist 🙂

geom/geom/inc/TGeoTypedefs.h
hageboeck: It's a fix. The header doesn't compile otherwise because the vertex definition is used later.

hageboeck: Oh, the doubling! That's probably a rebase error. 🙂

PR #19544, opened 2025-08-05
[ntuple] Check type in `RValue::GetRef<T>` and `GetPtr<T>` (version 2)

hahnjo:
This is a bit tricky because `RFieldBase.hxx` cannot depend on `RField` specializations in `RField.hxx`. We considered factoring `TypeName()` into a separate `RFieldTypeName` trait, but that would have still required including a bunch of STL headers to specialize for all supported containers. In the end, we settled on taking the `typeid` and passing the `std::type_info` to `Internal::GetRenormalizedTypeName`.

However, this comes with a run time cost, especially for fundamental types. In a microbenchmark, I measure 24ns per (successful) call to `REntry::GetPtr<std::int32_t>` while a naive implementation of always calling `GetRenormalizedTypeName` takes ~190ns per `RValue::GetPtr`. For more complex types, for example `ROOT::Math::PtEtaPhiMVector` with a number of templates, the difference is less pronounced if at all measurable with around 2000ns both.

To avoid this cost on repeated calls, we cache the `std::type_info` of matching types and try if it's equal to the `typeid`, which is basically a pointer comparison. This brings the (amortized) cost down to 8ns for any type! In the future, we will want to reuse the caching approach of `RValue::EnsureMatchingType` also for `REntry`.

Closes https://github.com/root-project/root/issues/18316


hahnjo:
For future reference, here are the microbenchmarks cited in the commit message:
```c++
#include <Math/Vector4D.h>

#include <ROOT/REntry.hxx>
#include <ROOT/RFieldBase.hxx>

#include <benchmark/benchmark.h>

#include <cstdint>
#include <utility>

static void REntry_GetPtr_int32(benchmark::State& state) {
  auto model = ROOT::RNTupleModel::CreateBare();
  auto field = ROOT::RFieldBase::Create("i32", "std::int32_t").Unwrap();
  model->AddField(std::move(field));
  model->Freeze();
  auto token = model->GetToken("i32");
  auto entry = model->CreateEntry();
  for (auto _ : state) {
    benchmark::DoNotOptimize(entry->GetPtr<std::int32_t>(token));
  }
}
BENCHMARK(REntry_GetPtr_int32);

static void REntry_GetPtr_LorentzVector(benchmark::State& state) {
  auto model = ROOT::RNTupleModel::CreateBare();
  auto field = ROOT::RFieldBase::Create("v", "ROOT::Math::PtEtaPhiMVector").Unwrap();
  model->AddField(std::move(field));
  model->Freeze();
  auto token = model->GetToken("v");
  auto entry = model->CreateEntry();
  for (auto _ : state) {
    benchmark::DoNotOptimize(entry->GetPtr<ROOT::Math::PtEtaPhiMVector>(token));
  }
}
BENCHMARK(REntry_GetPtr_LorentzVector);

static void RValue_GetPtr_int32(benchmark::State& state) {
  auto field = ROOT::RFieldBase::Create("i32", "std::int32_t").Unwrap();
  auto value = field->CreateValue();
  for (auto _ : state) {
    benchmark::DoNotOptimize(value.GetPtr<std::int32_t>());
  }
}
BENCHMARK(RValue_GetPtr_int32);

static void RValue_GetPtr_LorentzVector(benchmark::State& state) {
  auto field = ROOT::RFieldBase::Create("v", "ROOT::Math::PtEtaPhiMVector").Unwrap();
  auto value = field->CreateValue();
  for (auto _ : state) {
    benchmark::DoNotOptimize(value.GetPtr<ROOT::Math::PtEtaPhiMVector>());
  }
}
BENCHMARK(RValue_GetPtr_LorentzVector);

BENCHMARK_MAIN();
```

tree/ntuple/inc/ROOT/RFieldBase.hxx
```diff
@@ -706,6 +710,8 @@ private:
    RFieldBase *fField = nullptr;  ///< The field that created the RValue
    /// Set by Bind() or by RFieldBase::CreateValue(), RFieldBase::SplitValue() or RFieldBase::BindValue()
    std::shared_ptr<void> fObjPtr;
+   mutable const std::type_info *fTypeInfo = nullptr;
```
jblomer: I think this should be an atomic to keep thread-safety on the const methods.

hahnjo: Implemented - do we promise thread-safety of `const` methods somewhere? It unfortunately requires custom copy/move constructors and operators for `RValue`...

hahnjo:
The latest push also avoids extra type checks in `RNTupleView<T>`, which works around a renormalization bug that I'm fixing separately in https://github.com/root-project/root/pull/19550 (thanks @bellenot for confirming my suspicion).

vepadulano (approved):
Cool thanks!

tree/ntuple/inc/ROOT/RFieldBase.hxx
jblomer: I think we do here: https://github.com/root-project/root/blob/25c19af92ec29c2146417c34d94439dfedbe5f48/tree/ntuple/doc/Architecture.md?plain=1#L28

in combination with the explanation of "conditionally thread-safe" here: https://root.cern/manual/multi_threading/

We'll need custom move and copy constructors anyway at some point.

jblomer (approved):

PR #15152, opened 2024-04-05
Cluster-based scheduling of RNTuple processing in distributed mode

vepadulano:
The logic in `RNTupleDS::PrepareNextRanges` and `RNTupleDS::GetRanges` had to be changed a bit to accommodate the new case, I think it could be streamlined but I didn't want to change too many things in the same PR.

PR #19546, opened 2025-08-05
[df] Cluster-based scheduling of RNTuple processing in distributed mode

martamaja10:
Revival of https://github.com/root-project/root/pull/15152.


tree/dataframe/inc/ROOT/RNTupleDS.hxx
```diff
@@ -17,6 +17,7 @@
 #ifndef ROOT_RNTupleDS
 #define ROOT_RNTupleDS
 
+#include <ROOT/RDataFrame.hxx>
```
vepadulano: Seeing this again, would be better to avoid the extra header inclusion. As I see from the rest of the code, it should not be necessary actually

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -798,7 +798,11 @@ public:
    /// We know the number of entries from adding the cluster summaries
    ROOT::NTupleSize_t GetNEntries() const { return fNEntries; }
    ROOT::NTupleSize_t GetNElements(ROOT::DescriptorId_t physicalColumnId) const;
-
+   struct RClusterBoundaries {
+      ROOT::NTupleSize_t fFirstEntry;
+      ROOT::NTupleSize_t fLastEntry;
+   };
+   std::vector<RClusterBoundaries> GetClusterBoundaries() const;
```
vepadulano: Probably it would be better to start with this new API of RNTupleDescriptor in the `private` section and only access it internally, @jblomer thoughts?

tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -337,6 +337,24 @@ ROOT::NTupleSize_t ROOT::RNTupleDescriptor::GetNElements(ROOT::DescriptorId_t ph
...
+{
+   ROOT::RNTupleDescriptor::RClusterBoundaries clusterBoundaries;
+   std::vector<ROOT::RNTupleDescriptor::RClusterBoundaries> clusterBoundariesVec;
+   clusterBoundariesVec.reserve(GetNClusters());
+   auto clusterId = FindClusterId(0, 0);
+   while (clusterId != ROOT::kInvalidDescriptorId) {
+      const auto &clusterDesc = GetClusterDescriptor(clusterId);
+      clusterBoundaries.fFirstEntry = clusterDesc.GetFirstEntryIndex();
+      clusterBoundaries.fLastEntry = clusterDesc.GetFirstEntryIndex() + clusterDesc.GetNEntries();
+      clusterBoundariesVec.emplace_back(clusterBoundaries);
```
vepadulano: Probably this could also work
```suggestion
   std::vector<ROOT::RNTupleDescriptor::RClusterBoundaries> clusterBoundariesVec;
   clusterBoundariesVec.reserve(GetNClusters());
   auto clusterId = FindClusterId(0, 0);
   while (clusterId != ROOT::kInvalidDescriptorId) {
      const auto &clusterDesc = GetClusterDescriptor(clusterId);
      clusterBoundariesVec.emplace_back({clusterDesc.GetFirstEntryIndex(), clusterDesc.GetFirstEntryIndex() + clusterDesc.GetNEntries()});
```

vepadulano (changes requested):
Thank you so much! From my point of view we're very close to the finish line, minor comments.

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
martamaja10: now it is done with RClusterBoundaries in the ROOT::Internal namespace while the GetRNTupleClusterBoundaries method is private.

pcanal: > Probably it would be better to start with this new API of RNTupleDescriptor in the private section and only access it internally...

Why?  It seems to be a generic need for any tool that wants to iterate over clusters.  For example see: https://github.com/SMShovan/RNTuple-I-O-Config/blob/main/src/Utils.cpp which essentially implement the same thing... (albeit it is only 20 lines ... but still)

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -689,6 +701,8 @@ private:
    /// when merging two RNTuples.
    RNTupleDescriptor CloneSchema() const;
 
+   std::vector<ROOT::Internal::RNTupleClusterBoundaries> GetRNTupleClusterBoundaries() const;
```
jblomer: I think if we anyway use a free-standing function to do the work, we don't need this method.

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -59,6 +59,16 @@ class RFieldDescriptorBuilder;
...
 RNTupleDescriptor CloneDescriptorSchema(const RNTupleDescriptor &desc);
+struct RNTupleClusterBoundaries {
+   ROOT::NTupleSize_t fFirstEntry;
+   ROOT::NTupleSize_t fLastEntry;
+};
+
+const std::vector<ROOT::Internal::RNTupleClusterBoundaries>
+GetVecRNTupleClusterBoundaries(const ROOT::RNTupleDescriptor &RNTupDesc);
+
+class RSharedDescriptorGuard;
```
jblomer: It seems that we don't need that.

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -59,6 +59,16 @@ class RFieldDescriptorBuilder;
 class RNTupleDescriptorBuilder;
 
 RNTupleDescriptor CloneDescriptorSchema(const RNTupleDescriptor &desc);
+struct RNTupleClusterBoundaries {
+   ROOT::NTupleSize_t fFirstEntry;
+   ROOT::NTupleSize_t fLastEntry;
```
jblomer:
```suggestion
   ROOT::NTupleSize_t fFirstEntry = kInvalidNTupleIndex;
   ROOT::NTupleSize_t fLastEntry = kInvalidNTupleIndex;
```

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -59,6 +59,16 @@ class RFieldDescriptorBuilder;
...
+};
+
+const std::vector<ROOT::Internal::RNTupleClusterBoundaries>
```
jblomer: Let the use decide what to do with it...
```suggestion
std::vector<ROOT::Internal::RNTupleClusterBoundaries>
```

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -59,6 +59,16 @@ class RFieldDescriptorBuilder;
...
+GetVecRNTupleClusterBoundaries(const ROOT::RNTupleDescriptor &RNTupDesc);
```
jblomer:
```suggestion
GetRNTupleClusterBoundaries(const ROOT::RNTupleDescriptor &desc);
```

or maybe even (since it is internal)
```suggestion
GetClusterBoundaries(const ROOT::RNTupleDescriptor &desc);
```

tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -337,6 +337,22 @@ ROOT::NTupleSize_t ROOT::RNTupleDescriptor::GetNElements(ROOT::DescriptorId_t ph
    return result;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// Return the cluster boundaries for each cluster in this RNTuple.
+std::vector<ROOT::Internal::RNTupleClusterBoundaries> ROOT::RNTupleDescriptor::GetRNTupleClusterBoundaries() const
+{
+   std::vector<ROOT::Internal::RNTupleClusterBoundaries> clusterBoundariesVec;
```
jblomer:
```suggestion
   std::vector<Internal::RNTupleClusterBoundaries> boundaries;
```

tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -337,6 +337,22 @@ ROOT::NTupleSize_t ROOT::RNTupleDescriptor::GetNElements(ROOT::DescriptorId_t ph
...
+   clusterBoundariesVec.reserve(GetNClusters());
+   auto clusterId = FindClusterId(0, 0);
+   while (clusterId != ROOT::kInvalidDescriptorId) {
+      const auto &clusterDesc = GetClusterDescriptor(clusterId);
+      clusterBoundariesVec.emplace_back(ROOT::Internal::RNTupleClusterBoundaries{
+         clusterDesc.GetFirstEntryIndex(), clusterDesc.GetFirstEntryIndex() + clusterDesc.GetNEntries()});
```
jblomer, line 355: Shouldn't this be
```suggestion
         R__ASSERT(clusterDesc.GetNEntries() > 0);
         clusterDesc.GetFirstEntryIndex(), clusterDesc.GetFirstEntryIndex() + clusterDesc.GetNEntries() - 1});
```

martamaja10: Thank you for the suggestion. I am not sure if this is needed though. When I was adding the RDatasetSpec support in RNTuple I added a test that compares the output of RNTupleReader with the result of using RDatasetSpec (https://github.com/root-project/root/blob/3ee384d23bd968306bbce05e0f3c9724b777e67e/tree/dataframe/test/dataframe_datasetspec.cxx#L1029). When I do the subtraction, I get wrong result: 
```
using RNTupleReader: 
  h.GetMean()
    Which is: 2

using datasetspec: 
  mean.GetValue()
    Which is: 1.5 
    ```
Processing of distributed RDF fails then as well. But we can also discuss this offline.

Comment (jblomer) [reply tree/ntuple/src/RNTupleDescriptor.cxx:355]: Yep, let's discuss offline / in front of a screen.

File: tree/dataframe/inc/ROOT/RNTupleDS.hxx (hunk 1 above)
Comment (vepadulano) [inline tree/dataframe/inc/ROOT/RNTupleDS.hxx]: I see this include is still present, can we remove it?

File: tree/dataframe/src/RNTupleDS.cxx [hunk 11]
```diff
@@ -35,17 +35,17 @@
...
+ * \class ROOT::RDF::RNTupleDS
+ * \ingroup dataframe
+ * \brief The RDataSource implementation for RNTuple. It lets RDataFrame read RNTuple data.
+ *
+ * An RDataFrame that reads RNTuple data can be constructed using FromRNTuple().
+ *
+ * For each column containing an array or a collection, a corresponding column `#colname` is available to access
+ * `colname.size()` without reading and deserializing the collection values.
+ *
+ **/
```
vepadulano: Is the extra space indentation needed?

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -59,6 +59,15 @@ class RFieldDescriptorBuilder;
 class RNTupleDescriptorBuilder;
 
 RNTupleDescriptor CloneDescriptorSchema(const RNTupleDescriptor &desc);
+struct RNTupleClusterBoundaries {
+   ROOT::NTupleSize_t fFirstEntry = kInvalidNTupleIndex;
+   ;
```
vepadulano: spurious extra ;

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -59,6 +59,15 @@ class RFieldDescriptorBuilder;
...
+   ROOT::NTupleSize_t fLastEntryPlusOne = kInvalidNTupleIndex;
+   ;
```
vepadulano: spurious extra ;

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -59,6 +59,15 @@ class RFieldDescriptorBuilder;
...
+};
+
+std::vector<ROOT::Internal::RNTupleClusterBoundaries> GetClusterBoundaries(const RNTupleDescriptor &desc);
```
vepadulano, line 67:
```suggestion
std::vector<RNTupleClusterBoundaries> GetClusterBoundaries(const ROOT::RNTupleDescriptor &desc);
```

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -694,6 +704,8 @@ private:
    /// when merging two RNTuples.
    RNTupleDescriptor CloneSchema() const;
 
+   std::vector<ROOT::Internal::RNTupleClusterBoundaries> GetClusterBoundaries() const;
```
vepadulano:
```suggestion
   std::vector<Internal::RNTupleClusterBoundaries> GetClusterBoundaries() const;
```

vepadulano: Also, this method should be moved to the `private` section IIUC from the previous conversations

martamaja10: Maybe we could have a quick chat offline? I'm not sure anymore based on some earlier suggestions in the conversation above.

PR #19551, opened 2025-08-06
Remove occurences of legacy-style naming

vepadulano:
* Remove two old headers which were not included in the library anyway
* Change one occurrence of the old spelling in a tutorial
* Remaining old TDS namespace and keyword occurrences


martamaja10 (approved):
Nice! Thanks :)

PR #19552, opened 2025-08-06
[roottest] Fix some warnings on mac

silverweed:
Seen in [this PR](https://github.com/root-project/root/pull/19536)


roottest/cling/offset/runvbase.C
```diff
@@ -35,7 +35,7 @@ struct Bottom : public Mid1, Mid2
 #pragma warning(push)
 #pragma warning(disable : 4584)
 #endif
-#ifdef __CLING__
+#if defined(__CLING__) || defined(__clang__)
```
hahnjo: I think
```suggestion
#if defined(__clang__)
```
should be sufficient

hahnjo (approved):
Otherwise LGTM

Issue #14581, opened 2024-02-04
[ROOT-9733] TRandom3 does not implement perfectly a Mersenne Twister PRNG


vepadulano:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

From https://its.cern.ch/jira/browse/ROOT-9733

The {{TRandom3}} is a random number engine based on Mersenne Twister algorithm, defined in {{TRandom3.cxx}}. Since 2002 (see [commit 026ec9c0f458be61825e7335ec51e115ec6cbf4c](https://github.com/root-project/root/commit/026ec9c0f458be61825e7335ec51e115ec6cbf4c)), an eventual zero return value is discarded by {{TRandom3::Rndm()}}. This does not allow the algorithm to return the same sequence of the C++11 standard implementation {{std::mt19937}} (when seeded with the same value - except for seed zero, since handled in a special way by {{TRandom3::SetSeed()}} - and multiplied by 2^-32).

A related mistake is present also in the description of the class, where it states the algorithm period to be 2^19937 - 1. If you discard the zeros, of course the period will be (very slightly) less than it.

In conclusion, {{TRandom3}} is an imperfect implementation of the Mersenne Twister 32-bit algoritm. Furthermore, the documentation does not explain this small difference.

 

### Reproducer

As per the description

### ROOT version

Any

### Installation method

Any

### Operating system

Any

ferdymercury:
Another issue is that the docu says it includes 1, but it doesn't, since:

`UINT_MAX* 2.3283064365386963e-10`
is
`0.999999999767169356`

as noted in https://github.com/root-project/root/commit/07c0ad7ef9098fb58e800903d203c472c5ec5b1f

PR #14702, opened 2024-02-13
Removed skipping of 0 (zero) in TRandom3.Rndm()

kjvbrt:
# This Pull request: Removes skipping of 0 (zero) in TRandom3.Rndm()

## Changes or fixes:
* Removes skipping of 0 (zero) in TRandom3.Rndm()
* Updates documentation, adds warnings

## Checklist:

- [x] tested changes locally
- [x] updated the docs (if necessary)

This PR fixes #14581

PR #19554, opened 2025-08-06
[nfc][math] Document shortcoming of TRandom3

ferdymercury:
# This Pull request:

## Changes or fixes:

And clarify / correct some descriptions

Fixes https://github.com/root-project/root/issues/14581

## Checklist:

- [ ] tested changes locally
- [x] updated the docs (if necessary)

PR #19557, opened 2025-08-06
[math] Implement Mersenne Twister generator that does not skip elements of the sequence

ferdymercury:
Fixes https://github.com/root-project/root/issues/14581

Supersedes https://github.com/root-project/root/pull/14702

### DIFF
--- a/math/mathcore/CMakeLists.txt
+++ b/math/mathcore/CMakeLists.txt
@@ -93,6 +93,7 @@ set(HEADERS
   TRandom1.h
   TRandom2.h
   TRandom3.h
+  TRandom4.h
   TRandomGen.h
   TStatistic.h
   VectorizedTMath.h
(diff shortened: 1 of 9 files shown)


lmoneta (commented):
I am not sure if it is a good idea to introduce another class implementing another random generator, which is anyway not optimal, to fix a small issue. Either we decide to leave the current TRandom3 as it is, or we remove the skipping of the zero with the result that new versions of ROOT will generate different sequences. 
Since TRandom3 is not the best pseudo-random number generator, we should recommend using TRandomRanluxpp, I am in favour of leaving it as it is.

ferdymercury:
> Since TRandom3 is not the best pseudo-random number generator, we should recommend using TRandomRanluxpp, I am in favour of leaving it as it is

In that case, I would suggest merging https://github.com/root-project/root/pull/19554/files and closing issue as won't fix.
(And closing this PR too)

PR #19562, opened 2025-08-07
[RDF] Extend three RDF tests

hageboeck:
During work on snapshot with variations, these three snapshot tests were refined slightly to make the failures more readable or to slightly extend coverage.


tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -1350,6 +1359,36 @@ TEST_F(RDFSnapshotArrays, MultiThreadJitted)
    ROOT::DisableImplicitMT();
 }
 
+TEST_F(RDFSnapshotArrays, RedefineArrayMT)
+{
+   static constexpr unsigned int newArraySize = 6u;
+   static_assert(kFixedSize != newArraySize);
+   ROOT::EnableImplicitMT(3);
```
vepadulano: RAII struct that resets the IMT state and deletes the file?

tree/dataframe/test/dataframe_snapshot.cxx
```diff
@@ -750,12 +750,10 @@ void ReadWriteCarray(const char *outFileNameBase)
...
       EXPECT_EQ(rv.GetSize(), 100000u);
       EXPECT_EQ(rvb.GetSize(), 100000u);
-      for (auto e : rv)
-         EXPECT_EQ(e, 84);
-      for (auto e : rvb)
-         EXPECT_TRUE(e);
-      for (auto e : rvl)
-         EXPECT_EQ(e, 42);
+      EXPECT_EQ(rvl.GetSize(), 100000u);
+      EXPECT_TRUE(std::all_of(rv.begin(), rv.end(), [](auto elm){ return elm == 84; })) << rv[0] << " " << rv[1] << " " << rv[2];
```
vepadulano: Sorry I don't remember the details of this test, why are you printing rv[0], rv[1] and rv[2] specifically in case of failure?

vepadulano (changes requested):
Thanks! Minor suggestions

tree/dataframe/test/dataframe_snapshot.cxx
hageboeck: Previously, it was printing some 10k failure messages. Now it's only one message, and you see the first three examples. I admit that three is arbitrary, but it was a better number than 10k.

tree/dataframe/test/dataframe_snapshot.cxx
hageboeck: It's a copy of the ST test, made to look similar to the other MT tests in this file (see e.g. the context above). If we want to do it with RAII, it should probably be done for all the MT tests in this file.

vepadulano: Let's leave it as is then

hageboeck: Too late, there is a new commit now. 😅

vepadulano (approved):
LGTM thanks! Let's wait for the CI to be green before merging.

PR #19563, opened 2025-08-07
[TestSupport] Remove a special case for RNTuple.

hageboeck:
Now that RNT is not experimental, TestSupport doesn't need to mask its warning messages when running gtest.
The remaining warning message about merging can be masked explicitly (which is already happening in ntuple_merger.cxx).

### DIFF
--- a/core/testsupport/src/TestSupport.cxx
+++ b/core/testsupport/src/TestSupport.cxx
@@ -58,12 +58,6 @@ static struct ForbidDiagnostics {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


silverweed:
Actually merging RNTuples is still experimental so I believe this should remain until the RNTupleMerger is out of Experimental

hageboeck:
> Actually merging RNTuples is still experimental so I believe this should remain until the RNTupleMerger is out of Experimental

Initially, I thought to replace it by "active" checking, but seeing the number of warnings in the files, I tend to agree. 🙂

PR #19565, opened 2025-08-07
[roottest] use cmake for root/treeproxy tests

linev:
In many cases one need to copy scripts into build directory and
configure macro path with:
```
                  ROOTEXE_OPTS -e "(void)gROOT->SetMacroPath(\"${CMAKE_CURRENT_BINARY_DIR}\")"
```
Otherwise TTreeProxyGenerator not able to find scripts.

Also crucial to have some compiled macros in current directory to get correct declaration file set for created TClass instance - like in case of `TTrack` class. When not set - created selected will miss `Track.h` include and will not work.


linev:
@pcanal All failures unrelated to PR, they appears in all other PRs

roottest/root/treeproxy/stlvec/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+ROOTTEST_ADD_TEST(selector
+                  MACRO createtlorentzvecproxy.C
+                  COPY_TO_BUILDDIR tlorentzvec.root analyze.h analyze.C cutting.h cutting.C
+                  FIXTURES_SETUP root-treeproxy-stlvec-selector-fixture)
+
+ROOTTEST_ADD_TEST(tlorentzvec
+                  MACRO runtlorentzvec.C
```
pcanal, line 7: Side note: In the Makefile version ` tlorentzvecProxy.h` was compiled as a separate step, it is now compiled when running `runtlorentzvec.C`.  This is fine as `ROOTTEST_ADD_TEST` filters out the `ACLiC` info message about compilation.

roottest/root/treeproxy/CMakeLists.txt
```diff
@@ -1,8 +1,124 @@
...
+                  ROOTEXE_OPTS -e "(void)gROOT->SetMacroPath(\"${CMAKE_CURRENT_BINARY_DIR}\")"
+                  FIXTURES_REQUIRED root-treeproxy-createvaldim3-fixture
+                  FIXTURES_SETUP root-treeproxy-val3dimSel-fixture)
+
+ROOTTEST_COMPILE_MACRO(runvaldim3.C
+                       FIXTURES_REQUIRED root-treeproxy-Track-fixture
+                       FIXTURES_SETUP root-treeproxy-runvaldim3-fixture)
+
+ROOTTEST_ADD_TEST(valdim3
+                  MACRO runvaldim3.C+
```
pcanal, line 66: Consider building `val3simsel.h` (from `val3dimSel.h` generated by `val3dimSel/createvaldim3`) as a separate step.

pcanal (approved):
LGTM.  Thanks.

roottest/root/treeproxy/CMakeLists.txt
linev: I will submit it as separate PR while compilation of generated selector can be not trivial

roottest/root/treeproxy/stlvec/CMakeLists.txt
linev: Same here - compilation of generated selector is not that obvious. I will try to make it in new PR.

PR #19566, opened 2025-08-07
Add extra code to specify workflow conditions

vepadulano:
I noticed the workflow is being skipped (e.g. https://github.com/root-project/root/actions/runs/16792883609). The gh UI doesn't say why. I have a suspicion that it is yet another undocumented behaviour. This commit suggests to add extra code to specify each and every  scenario where the workflow can run.

Even though it is redundant with the `on` section, it looks like this is the way to have the workflow properly run.

### DIFF
--- a/.github/workflows/python_wheel_build.yml
+++ b/.github/workflows/python_wheel_build.yml
@@ -15,7 +15,10 @@ on:
 
 jobs:
   build-wheels:
-    if: contains(github.event.pull_request.labels.*.name, 'build-python-wheels')
+    if: |
+      github.event_name == 'schedule' ||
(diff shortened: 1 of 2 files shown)


vepadulano:
Something broke the workflow in the meantime :) investigating...

hahnjo:
You probably want to skip the upload step for PR builds. It currently doesn't have access to the secret, and even if it had I'm not sure it should actually upload something to the "production" repository...

vepadulano:
> You probably want to skip the upload step for PR builds. It currently doesn't have access to the secret, and even if it had I'm not sure it should actually upload something to the "production" repository...

Thank you! I was thinking the same, will do it in a followup PR

Issue #19568, opened 2025-08-07
Ineffective `TMatrixTSym` symmetry check during construction


Zehvogel:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

This check in the constructor doesn't work

https://github.com/root-project/root/blob/a535d37f04dbfb389c5a375065fde0e6a1542423/math/matrix/src/TMatrixTSym.cxx#L69-L71

because of

https://github.com/root-project/root/blob/a535d37f04dbfb389c5a375065fde0e6a1542423/math/matrix/inc/TMatrixTSym.h#L99

making it possible to construct non-symmetric matrices that tell you that they are symmetric.

### Reproducer

```python
import ROOT

v = ROOT.RVecD([1., 2., 3., 4.])
m = ROOT.TMatrixDSym(2, v.data())
print(m(0,1))
# 2.0
print(m(1,0))
# 3.0
print(m.IsSymmetric())
# True
```

### ROOT version

```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.36.000                       https://root.cern |
  | (c) 1995-2025, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Jun 07 2025, 14:20:17                 |
  | From tags/6-36-000@6-36-000                                      |
  | With g++ (Spack GCC) 14.2.0                                      |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------
```

### Installation method

source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh

### Operating system

Alma9

### Additional context

there are similar ineffective checks in some of the other constructors

PR #19570, opened 2025-08-07
[math] properly check for symmetry when constructing TSym

ferdymercury:
Fixes https://github.com/root-project/root/issues/19568


hageboeck:
The fedora failures are unrelated, and I don't see a good reason to clang-format the surrounding code, so I'll merge as is.

PR #19571, opened 2025-08-07
[roottest] improve cmake for `root/collection`

linev:
Use explicit list of tests - one test was not present in automatic list 
Copy ROOT file only for specific test
Remove unnecessary `.rootrc`


linev:
I do not change tests - just adding fixtures to have clear order of tests invocation.
`DEPENDS` argument of `ROOTTEST_ADD_TEST` does not invoke required test

PR #19575, opened 2025-08-08
[roottest] use fixtures for dataframe tests

linev:
There are many executables and each receives fixture to correctly generate it for correspondent test
Using DEPENDS do not allow to correctly run individual tests

### DIFF
--- a/roottest/root/dataframe/CMakeLists.txt
+++ b/roottest/root/dataframe/CMakeLists.txt
@@ -1,4 +1,4 @@
-if(NOT ROOT_dataframe_FOUND)
+if(NOT dataframe)
   return()
 endif()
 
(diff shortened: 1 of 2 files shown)


roottest/root/dataframe/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ endif()
 
 set(DFLIBRARIES Core RIO Hist Tree MathCore TreePlayer ROOTDataFrame ROOTVecOps)
 
-if(ROOT_root7_FOUND)
+if(root7)
   list(APPEND DFLIBRARIES ROOTNTuple)
 endif()
```
vepadulano: Nowadays this `if` is not needed anymore, `ROOTNTuple` is built by default, can you remove the if?

vepadulano (approved):
Thank you so much!

PR #19576, opened 2025-08-08
[roottest] improve `root/math`, `root/fit`, `root/hist` tests

linev:
Correctly configure reference files for `math/smatrix` tests
Remove no longer used `math/genvector` and `math/mathcore` - tests were moved to source directory
Remove not necessary `.rootrc` file - it automatically created for each dir


pcanal (approved):
LGTM.  Thanks.

PR #19579, opened 2025-08-08
[CMake] Also check for sklearn if the test requires it via xgboost

guitargeek:
Fixes the failures in the nighlies on Fedora 43, which uses Python 3.14 where xgboost is available but sklearn is not.


vepadulano (approved):
LGTM, but check for typos `sklean` --> `sklearn`

guitargeek:
This PR caused failues on Windows, which I should have caught in the CI but I didn't wait for the Windows tests to complete. I thought since xgboost is not available on the Windows anyway, this would have no effect :( Sorry for the red nightlies over the weekend :pray:

PR #19581, opened 2025-08-08
[ntuple] add test for model extension with streamer fields

silverweed:
- [x] tested changes locally


hahnjo (approved):
Thanks!

PR #19584, opened 2025-08-08
Adjust stressGraphics_web.ref

linev:
On newest fedora machines PNG variation for two
basic tests larger than before


vepadulano:
Isn't the test still failing on both platforms even with these changes?

linev:
Yes, there are other tests failing. 
I have to provide more changes

test/stressGraphics_web.ref
```diff
@@ -1,33 +1,33 @@
 Test#   SVG1Ref#  SVG1Err#  PDFRef#   PDFErr#   JPGRef#   JPGErr#   PNGRef#   PNGErr#   SVG2Ref#  SVG2Err#
     1      1356        58      4268        50     75219     29324     15193     10000      1356        58
-    2     13739        50     14950       161     33621      3693     17791      3815     13739        50
+    2     13739        50     14950       161     33621      3693     17791      7000     13739        50
     3       611        50      3802        50     27412     10485     24438     14774       611        50
-    4     49021       306     93647       842    304049     30830    139824     34984     49931       306
+    4     49021       306     93647       842    304049     30830    139824     60000     49931       306
```
pcanal, line 5: This seems like a large difference.  Is it really just 'noise' and not visible/noticeable differences?

linev: Yes, in some cases variations very large.
But I pretty sure that this is real variations in PNG files.

pcanal: > But I pretty sure that this is real variations in PNG files.

I am confused by this wording.  Is the output 'looking different' (will affect user experience) or is the same output 'simply encoded and compressed poorly by the new version'  (will NOT affect user experience except for bigger files).

linev: This PNG created by chrome browser. I cannot reproduce it myself - while I do not have exactly this version of chrome. 

But all other files - especially SVG and PDF - remains stable. Therefore I wrote that PNG generated from same SVG file should be ok.

PR #19586, opened 2025-08-08
[core] Move `TClass` tests to `core/meta`

hahnjo:
The test in `core/clingutils` is meant for unit testing of `TClingUtils` and statically links `ClingUtils` and Cling/Clang/LLVM libraries. Calling Cling, `TInterpreter`, or `TClass` results in two copies of the functions being around (in `libCling.so` and the test binary) with not much guarantees which ones are called at which moment.

(causes a test failure with the upgrade to LLVM 20)

### DIFF
--- a/core/clingutils/test/TClingUtilsTests.cxx
+++ b/core/clingutils/test/TClingUtilsTests.cxx
@@ -21,6 +21,7 @@
 // test was to do any of the above, there would be two copies of functions around (in libCling.so and the test binary)
 // with not much guarantees which ones are called at which moment.
 
+#include <DllImport.h>
 #include <ROOT/FoundationUtils.hxx>
 
 #include "gtest/gtest.h"
@@ -90,3 +91,13 @@ TEST(TClingUtilsTests, GetRealPath)
    std::remove("./realfile2");
 #endif // not R__WIN32
 }
+
+// Forward-declare gCling to not include TInterpreter.h just for checking that the interpreter has not been initialized.
+class TInterpreter;
+R__EXTERN TInterpreter *gCling;
+
+class InterpreterCheck : public testing::Environment {
+   void TearDown() override { ASSERT_EQ(gCling, nullptr); }
+};
+
+testing::Environment *gInterpreterCheck = testing::AddGlobalTestEnvironment(new InterpreterCheck);


devajithvs (approved):
LGTM! Thanks.

core/clingutils/test/TClingUtilsTests.cxx
```diff
@@ -15,15 +15,17 @@
  *************************************************************************/
 
 #include <TClingUtils.h>
-#include <TClass.h>
-#include <TInterpreter.h>
+
+// This test must under no circumstances include other ROOT headers. It must not call into Cling, TInterpreter, TClass.
+// It is meant for unit testing of TClingUtils and statically links ClingUtils and Cling/Clang/LLVM libraries. If the
+// test was to do any of the above, there would be two copies of functions around (in libCling.so and the test binary)
+// with not much guarantees which ones are called at which moment.
```
vgvassilev, line 22: Can we somehow think of a way to enforce this comment programmatically? Maybe dlsym a symbol from libCling.so?

hahnjo: I'm not sure how this would work since we will have two copies of the functions around, the point is just that we must not use the ones in `libCling.so`. I don't think this can be enforced, except not linking `libCore` (which I tried, but doesn't work because `core/clingutils` depends on `core/foundation`...).

vgvassilev: I was thinking something like `dlsym(gInterpreterLib, some_libcling_symbol)`.

pcanal:
> (causes a test failure with the upgrade to LLVM 20)

Having twice the same symbols is indeed bad.  What was the symptoms in the failures?

pcanal (approved):
Thanks. (and indeed a test using `TClass` belongs to core/meta).

dpiparo (approved):

core/clingutils/test/TClingUtilsTests.cxx
hahnjo: Sure, but what do I do with the return value? In any case, does this discussion block merging this PR? As much as I would like to enforce the constraint, I don't currently see how to do that as I said.

hahnjo:
> > (causes a test failure with the upgrade to LLVM 20)
> 
> Having twice the same symbols is indeed bad. What was the symptoms in the failures?

There were two symptoms with LLVM 20, depending on your "luck":
1. Without JITLink, target machines & triples were registered in the static copy of the symbols and then `libCling.so` could not find them. This was "caused" by a change in symbol visibility in LLVM 20 and left us confused for quite some time.
2. When hacking around this (for example by reverting that symbol visibility change), `llvm::Module` destruction would fail at the end of the test execution because hash values were different. That's because of a second change that now uses a process-specific seed for hashing. That is implemented by taking the address of a known function (`install_fatal_error_handler`) and relying on address space layout randomization (ASLR) to load code at different virtual addresses every time. The problem here is that we have two `install_fatal_error_handler`, one in the static binary and one in `libCling.so`.

We focused on the second point and eventually I guessed what is going on when seeing the size of `coreclingutilstestUnit`.

core/clingutils/test/TClingUtilsTests.cxx
vgvassilev: If we return result is non-zero, then you assert. That's all it seems its needed. Does not block it but we can probably experiment quickly if the approach helps.

vgvassilev: something like `assert ((!gInterpreterLib || !dlsym(gInterpreterLib, some_libcling_sym) && "libCling somehow loaded!);`

hahnjo: This doesn't work conceptually: `gInterpreterLib` is a `static` variable in `core/base/src/TROOT.cxx`, I cannot access it from the test. We can check `gCling == nullptr` though, in contrast to `gInterpreter` it doesn't auto-initialize.

vgvassilev: > This doesn't work conceptually: gInterpreterLib is a static variable in core/base/src/TROOT.cxx

Can't we forward declare it with `extern void *gInterpreterLib;` and use it. The linker will know what to do there...

EDIT: Perhaps we will need to remove the `static` part of `gInterpreterLib` but should not be a huge deal.

vgvassilev: > We can check gCling == nullptr though, in contrast to gInterpreter it doesn't auto-initialize.

That should probably do the trick, too...

PR #19587, opened 2025-08-08
[build] Add post-install test for Python bindings

vepadulano:


test/PostInstall/CMakeLists.txt
```diff
@@ -18,4 +18,22 @@ if(BUILD_TESTING)
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
   set_tests_properties(run-hsimple   PROPERTIES FIXTURES_SETUP    HSIMPLE)
   set_tests_properties(read-hsimple  PROPERTIES FIXTURES_REQUIRED HSIMPLE)
+
+  find_package(Python3 3.8 REQUIRED)
```
hageboeck, line 22: If I'm not mistaken, this specifically asks for Python 3.8. Isn't this overly restrictive?

vepadulano: No, this is the minimum compatible version. In order for it to be exact, the `EXACT` keyword should be used in the call to `find_package`. This is copied as-is from https://github.com/root-project/root/blob/1b970322d0b65d5b9152ca98eeee9528752114d9/cmake/modules/SearchInstalledSoftware.cmake#L710

PR #19591, opened 2025-08-09
Don't include TMathBase.h in TString.h

guitargeek:
The TString header only uses `TMath` functions that are readily available in the standard library.

The `TMathBase.h` header is quite big, and avoiding to include it reduces the overall cold-ccache compile time of the full ROOT project (excluding LLVM/Clang) with all default features by about 2 % on my machine, namely from 870 s to 855 s on average (I repeated the measurement several times, they are reproducible to a second).

In the places there the `TMath` functions were used but only indirectly included via `TString.h` inclusion, the commit suggests to use the `<cmath>` functions instead.

### DIFF
--- a/README/ReleaseNotes/v638/index.md
+++ b/README/ReleaseNotes/v638/index.md
@@ -95,6 +95,13 @@ the deprecation period was extended.
 ## JavaScript ROOT
 - A new configuration option `Jupyter.JSRoot` was added in .rootrc to set the default mode for JSROOT in Jupyter notebooks (on or off).
 
+### Optimization of ROOT header files
+
+More unused includes were removed from ROOT header files.
+For instance, `#include "TMathBase.h"` was removed from `TString.h`.
+This change may cause errors during compilation of ROOT-based code. To fix it, provide missing the includes
+where they are really required.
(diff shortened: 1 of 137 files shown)


dpiparo (approved):
Thanks for this refactoring!
Minor observation for the commit message: "In the places there the `TMath` functions" --> "In the places where the `TMath` functions"?

dpiparo:
Not specific to this PR, but more in general: can an automated way be envisaged to spot similar opportunities for improving build time? Also because in ROOT buildtime and runtime, thanks to C++ interpretation, have a very much blurred borders...

guitargeek:
> Not specific to this PR, but more in general: can an automated way be envisaged to spot similar opportunities for improving build time? Also because in ROOT buildtime and runtime, thanks to C++ interpretation, have a very much blurred borders...

That's a good question. I don't know how to automatize this, but I have manually identified a few spots that I might address later.

By the way, the motivation for this PR were RooFit compile times in particular. RooFit is a pretty huge codebase inside ROOT, and compile times were creeping up gradually. Also because it has so many test binaries...

I'll try to get this under control gradually, but changes should also not be too aggressive. Users might also rely on these indirect inclusions and get annoyed when the code doesn't compile anymore. The removal of `TMathBase.h` from `TString.h` is already quite the big change in that sense. In one release cycle we should probably not do much more than that.

dpiparo:
Given the magnitude of the change, does it make sense to already add some wording to the RNs?

PR #19594, opened 2025-08-10
[tutorials] Disable `tmva100_DataPreparation` tutorial on Windows

guitargeek:
In commit https://github.com/root-project/root/commit/042f6985a13464f410fd3eb1e1e48681280b592f, I removed the usage of the `tmva10*.py` wild card for vetoing if xgboost is not found, because it caused false positives: the `tmva100_DataPreparation` tutorial is actually not using xgboost at all.

But now we see in the nightlies that this tutorial crashes on Windows, so it needs to be vetoed in that case.

Also, in a separate commit, move `import xgboost` inside the main function of the tmva101 tutorial, such that other tutorials that use the tmva101 tutorial as a library don't inherit this dependency

### DIFF
--- a/tutorials/CMakeLists.txt
+++ b/tutorials/CMakeLists.txt
@@ -113,6 +113,10 @@ if (NOT dataframe)
     # RDataFrame tutorial in graphs
     list(APPEND dataframe_veto visualisation/graphs/gr109_timeSeriesFromCSV_RDF.C)
     # TMVA tutorials dependent on RDataFrame
+    list(APPEND dataframe_veto machine_learning/tmva100_DataPreparation.py
+        machine_learning/tmva101_Training.py # depends on tmva100_DataPreparation.py
+        machine_learning/tmva102_Testing.py # depends on tmva101_Training.py
+    )
     list(APPEND dataframe_veto machine_learning/tmva*.C)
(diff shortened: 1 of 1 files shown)


hahnjo:
https://github.com/root-project/root/pull/19579 found this during testing, but was unfortunately merged before the CI could complete...

tutorials/CMakeLists.txt
```diff
@@ -105,6 +113,7 @@ if (NOT dataframe)
     # RDataFrame tutorial in graphs
     list(APPEND dataframe_veto visualisation/graphs/gr109_timeSeriesFromCSV_RDF.C)
     # TMVA tutorials dependent on RDataFrame
+    list(APPEND dataframe_veto machine_learning/tmva100_DataPreparation.py)
```
hahnjo: I guess this also needs to veto `tmva101` and `tmva102`? Might be worth to point out in the commit message / move to a separate commit

hahnjo (approved):
LGTM with one comment about vetoing without `dataframe`

tutorials/CMakeLists.txt
guitargeek: You are right, thanks! I have moved this change to a separate commit and also added `tmva101` and `tmva102`.

PR #19596, opened 2025-08-11
[hist] Implement initial `RAxes`

hahnjo:
This first version supports `RRegularAxis` and `RVariableBinAxis` and exposes only `ComputeTotalNumBins` and `ComputeGlobalIndex`.

### DIFF
--- a/hist/histv7/inc/ROOT/RAxes.hxx
+++ b/hist/histv7/inc/ROOT/RAxes.hxx
@@ -58,6 +58,8 @@ public:
             totalNumBins *= regular->GetTotalNumBins();
          } else if (auto *variable = std::get_if<RVariableBinAxis>(&axis)) {
             totalNumBins *= variable->GetTotalNumBins();
+         } else {
+            throw std::logic_error("unimplemented axis type");
          }
       }
       return totalNumBins;
@@ -75,6 +77,8 @@ private:
       } else if (auto *variable = std::get_if<RVariableBinAxis>(&axis)) {
          index *= variable->GetTotalNumBins();
          linIndex = variable->ComputeLinearizedIndex(std::get<I>(args));
+      } else {
+         throw std::logic_error("unimplemented axis type");
       }
       if (!linIndex.fValid) {
          return {0, false};


hist/histv7/inc/ROOT/RAxes.hxx
```diff
@@ -0,0 +1,114 @@
...
+   /// \param[in] args the arguments
+   /// \return the global index that may be invalid
+   template <typename... A>
+   RLinearizedIndex ComputeGlobalIndex(const std::tuple<A...> &args) const
+   {
+      if (sizeof...(A) != fAxes.size()) {
+         throw std::invalid_argument("invalid number of arguments to ComputeGlobalIndex");
+      }
+      return ComputeGlobalIndex<0, A...>(0, args);
+   }
```
hageboeck, line 105: Are users expected to call this function or will they mostly interact with `RHist`? If yes, what are your thoughts on providing a convenience overload like this:
```diff
- ax.ComputeGlobalIndex(std::make_tuple(1));
+ ax.ComputeGlobalIndex(1);
```

hahnjo: From `doc/CodeArchitecture.md`:
> Objects of this class (`Internal::RAxes`) are used internally and not exposed to the user.
> Relevant functionality is available through user-facing classes such as `RHistEngine`.

See also the section about "Functions with Variable Number of Arguments" in `doc/DesignImplementation.md`.

So no, users are not expected to call `ComputeGlobalIndex` directly and I would therefore prefer to have as few overloads as possible.

hist/histv7/inc/ROOT/RAxes.hxx
```diff
@@ -0,0 +1,114 @@
...
+
+namespace ROOT {
+namespace Experimental {
+namespace Internal {
+
+/**
+Bin configurations for all dimensions of a histogram.
+
+\warning This is part of the ROOT 7 prototype! It will change without notice. It might trigger earthquakes. Feedback is
+welcome!
```
jblomer: I think we don't necessarily need this banner for an experimental class.

hist/histv7/inc/ROOT/RAxes.hxx
```diff
@@ -0,0 +1,114 @@
...
+
+private:
+   std::vector<AxisVariant> fAxes;
+
+public:
+   /// \param[in] axes the axis objects, must have size > 0
+   explicit RAxes(std::vector<AxisVariant> axes) : fAxes(std::move(axes))
+   {
+      if (fAxes.empty()) {
+         throw std::invalid_argument("must have at least 1 axis object");
```
jblomer, line 39: We should consider letting ROOT always throw an `RException` so that users can consistently catch a "ROOT error".

hist/histv7/inc/ROOT/RAxes.hxx
```diff
@@ -0,0 +1,114 @@
...
+      }
+   }
+
+   std::size_t GetNumDimensions() const { return fAxes.size(); }
```
jblomer, line 43:
```suggestion
   std::size_t GetNDimensions() const { return fAxes.size(); }
```

hist/histv7/inc/ROOT/RAxes.hxx
```diff
@@ -0,0 +1,114 @@
...
+   const std::vector<AxisVariant> &GetVector() const { return fAxes; }
```
jblomer:
```suggestion
   const std::vector<AxisVariant> &Get() const { return fAxes; }
```

hist/histv7/inc/ROOT/RAxes.hxx
```diff
@@ -0,0 +1,114 @@
...
+
+   friend bool operator==(const RAxes &lhs, const RAxes &rhs) { return lhs.fAxes == rhs.fAxes; }
+
+   /// Compute the total number of bins for all axes.
+   ///
+   /// It is the product of each dimension's total number of bins.
+   ///
+   /// \return the total number of bins
+   std::size_t ComputeTotalNumBins() const
```
jblomer, line 53:
```suggestion
   std::size_t ComputeTotalNBins() const
```

or maybe just
```suggestion
   std::size_t ComputeNBins() const
```

hist/histv7/inc/ROOT/RAxes.hxx
```diff
@@ -0,0 +1,114 @@
...
+   {
+      std::size_t totalNumBins = 1;
+      for (auto &&axis : fAxes) {
+         if (auto *regular = std::get_if<RRegularAxis>(&axis)) {
+            totalNumBins *= regular->GetTotalNumBins();
+         } else if (auto *variable = std::get_if<RVariableBinAxis>(&axis)) {
+            totalNumBins *= variable->GetTotalNumBins();
+         }
```
jblomer, line 63: Add an else with an `R__ASSERT` or `assert` so that we don't forget this spot if we new axis types.

hist/histv7/inc/ROOT/RAxes.hxx
```diff
@@ -0,0 +1,114 @@
...
+   {
+      const auto &axis = fAxes[I];
+      RLinearizedIndex linIndex;
+      if (auto *regular = std::get_if<RRegularAxis>(&axis)) {
+         index *= regular->GetTotalNumBins();
+         linIndex = regular->ComputeLinearizedIndex(std::get<I>(args));
+      } else if (auto *variable = std::get_if<RVariableBinAxis>(&axis)) {
+         index *= variable->GetTotalNumBins();
+         linIndex = variable->ComputeLinearizedIndex(std::get<I>(args));
+      }
```
jblomer, line 82: Add a failing default branch.

hist/histv7/inc/ROOT/RAxes.hxx
hahnjo: (I guess you mean "internal" class?)

hist/histv7/inc/ROOT/RAxes.hxx
hahnjo: I briefly discussed this topic with @hageboeck and we're leaning towards keeping `std::invalid_argument` for two reasons: First, `std::invalid_argument` has established semantics (also used by RooFit and RDataFrame) and we're throwing it during object construction. This is usually a problem during setup before the event loop, and not like an IO exception because of corrupt data. Second, it would introduce a dependency on Core which I would like to avoid for the standalone build...

hist/histv7/inc/ROOT/RAxes.hxx
hahnjo: Ah yes, this is something I "blindly" copied from my prototype and didn't think about the "ROOT7 conventions" . I'm OK with changing this, but then I'd like to do it consistently for `RRegularAxis` and `RVariableBinAxis` as well. Even though `GetNNormalBins` looks slightly awkward there, but any convention will break down eventually...

If that's ok, I will address this consistently in a follow-up PR.

hist/histv7/inc/ROOT/RAxes.hxx
hahnjo: see above

hist/histv7/inc/ROOT/RAxes.hxx
hahnjo: I need to test here that it doesn't add overhead. Otherwise @hageboeck and me discussed adding an "unreachable" implementation that throws in `#ifndef NDEBUG` builds, but otherwise just compiles to an illegal instruction. The problem I see here is that we will only find it during testing when we actually think about it, and then we will detect it anyway based on the incorrect behavior...

hahnjo: There is some overhead to putting a failing default branch (up to 10% in the microbenchmark, depending on the compiler). However, it might well be within noise together with future changes to check the argument types for a given axis type. So for now, I added a `throw std::logic_error` (in a separate commit) that will also be clear to users in case we miss it during testing.

hist/histv7/inc/ROOT/RAxes.hxx
jblomer: Ok, not something for this PR. This may well be a good approach but I'd still like to clarify how we want to use exceptions consistently in ROOT.

PR #19598, opened 2025-08-11
Minor changes to make build output less verbose

guitargeek:
See commit descriptions.


cmake/modules/SetROOTVersion.cmake
```diff
@@ -12,9 +12,7 @@
 # - GIT_DESCRIBE_ALL: output of `git describe --all` if source directory is git repo
 
 
-cmake_minimum_required(VERSION 3.16 FATAL_ERROR)
-
-find_package(Git)
+cmake_minimum_required(VERSION 3.20 FATAL_ERROR)
```
pcanal, line 15: Is CMake 3.20 the minimum default cmake install on all our platforms?

core/rint/src/TRint.cxx
```diff
@@ -498,7 +498,6 @@ void TRint::Run(Bool_t retrn)
    }
 
    if (QuitOpt()) {
-      printf("\n");
```
pcanal, line 501: For better or worse, this 'can' be noticed by users (eg. that have some automation of sorts that includes diff or what not).  So we do (yes, sigh ... :) ) need to explicitly announce it in the release notes (I vaguely remember at least one issue related to this kind of empty line).

cmake/modules/SetROOTVersion.cmake
pcanal: Nevermind, this is already the required version.

core/rint/src/TRint.cxx
guitargeek: Ok why not, but who would do diffs of running `root` with not input files and no interactivity :smiley: We do it to implicitly create modules, but that's probably it.

Maybe some people use `root -b -q` to get the banner to validate the ROOT version...

guitargeek:
Thanks for the review @pcanal!

core/rint/src/TRint.cxx
pcanal: Thanks for adding the release notes :)

CMakeLists.txt
```diff
@@ -381,9 +381,14 @@ add_custom_target(move_artifacts DEPENDS ${stamp_file} ${all_artifacts})
 add_subdirectory (interpreter)
 
 # Update etc/gitinfo.txt for every build.
+find_package(Git)
 add_custom_target(gitinfotxt
```
pcanal, line 385: Should there be a:
```
if(Git_FOUND)
   add_custom_target(gitinfotxt
      ....
   )
else()
   message(WARNING "git executable not found, commit info will be inaccurate")
endif()
```
and/or how do we fail if git is missing?

pcanal: Nevermind, `SetROOTVersion.cmake` which is called from `cmake/modules/UpdateGitInfo.cmake` ends up warning if `Git_FOUND` is false.

core/rint/src/TRint.cxx
pcanal: I still can find the exact report.  I am slightly remembering that the user/experiment was needing to make a distinction between `root -b -l -q` returning absolutely nothing and `root.exe -b -l -q` having that extra line ... can't yet remember the exact context.

pcanal: and I might be confused with other (similar but unrelated discussion like https://github.com/root-project/cling/issues/61#issuecomment-256393398)

CMakeLists.txt
guitargeek: Yes, that's the idea

PR #19599, opened 2025-08-11
[df] Support MT snapshotting to RNTuple

hahnjo:
Switch the existing code to use the `RNTupleParallelWriter` with one `RNTupleFillContext` per slot. This should be (almost) as efficient as the `RNTupleWriter` (one additional cloned `RNTupleModel` for the only fill context), but save quite a bit of code duplication and in testing effort.

(Draft for early comments while testing functionality and scalability; I can split out the preparatory commits into a separate PR if wanted)


pcanal:
> (one additional cloned RNTupleModel for the only fill context)

I am confused by the wording of this part of the commit log.  `only` suggest there is a single fill context for the whole process ... which is contradicted by `one fill context per slot`.  What is means by the above sentence?

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -874,11 +868,41 @@ void ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::Initialize()
...
+
+void ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::Exec(unsigned int slot, const std::vector<void *> &values)
+{
+   auto &fillContext = fFillContexts[slot];
+   auto &outputEntry = fEntries[slot];
+   assert(values.size() == fFieldTokens.size());
+   for (decltype(values.size()) i = 0; i < values.size(); i++) {
+      outputEntry->BindRawPtr(fFieldTokens[i], values[i]);
+   }
+   fillContext->Fill(*outputEntry);
```
pcanal, line 892: Isn't it meant to be `FillNoFlush`?

hahnjo:
> > (one additional cloned RNTupleModel for the only fill context)
> 
> I am confused by the wording of this part of the commit log. `only` suggest there is a single fill context for the whole process ... which is contradicted by `one fill context per slot`. What is means by the above sentence?

Agreed, it's not well formulated. What I'm trying to say is that for sequential snapshotting (that is already supported before the PR), changing from the `RNTupleWriter` to the `RNTupleParallelWriter` has only a slight overhead. I will revise the commit message.

tree/dataframe/src/RDFSnapshotHelpers.cxx
hahnjo: It's not needed here because `UntypedSnapshotRNTupleHelper` exclusively owns the `TFile` and there is only one `RNTupleParallelWriter` appending to it. Without the need to lock on the user / "framework" side, there is no benefit to using `FillNoFlush` (except it's longer and more code to write).

tree/dataframe/test/dataframe_snapshot_ntuple.cxx
```diff
@@ -614,23 +626,22 @@ struct TIMTEnabler {
...
-      *sdf;
-      FAIL() << "MT snapshotting to RNTuple is not supported yet";
-   } catch (const std::runtime_error &err) {
-      EXPECT_STREQ(err.what(), "Snapshot: Snapshotting to RNTuple with IMT enabled is not supported yet.");
-   }
+   auto sdf = df.Snapshot("ntuple", fileGuard.GetPath(), {"x"}, opts);
+   *sdf;
+
+   auto reader = RNTupleReader::Open("ntuple", fileGuard.GetPath());
+   EXPECT_EQ(25, reader->GetNEntries());
```
enirolf: Perhaps we can extend the test a bit and also check for contents, like:
```cpp
   auto df = ROOT::RDataFrame(25ull).Define("x", [](std::uint64_t i){ return i; }, {"rdfentry_"});

   RSnapshotOptions opts;
   opts.fOutputFormat = ROOT::RDF::ESnapshotOutputFormat::kRNTuple;

   auto sdf = df.Snapshot("ntuple", fileGuard.GetPath(), {"x"}, opts);
   *sdf;
   
   auto sum = sdf->Sum<std::uint64_t>("x");
   EXPECT_EQ(276, sum.GetValue());

   auto reader = RNTupleReader::Open("ntuple", fileGuard.GetPath());
   EXPECT_EQ(25, reader->GetNEntries());
```

enirolf (commented):
Really really nice! I'm wondering if for completeness we should also add some of the existing tests as MT tests :thinking:

tree/dataframe/test/dataframe_snapshot_ntuple.cxx
hahnjo: Done. However, I computed a different expected value and the test confirms the sum is indeed 300 :sweat_smile:

hahnjo:
> I'm wondering if for completeness we should also add some of the existing tests as MT tests 🤔

Yes, I wasn't sure either. We're of course massively benefiting that we use the exact same model creation and filling code that you already wrote all the tests for. Additionally, we have the challenge of MT scheduling, so even the best test case will have to deal with non-determinism and potentially still not test the relevant thing because all events end up on a single thread...

enirolf:
> > I'm wondering if for completeness we should also add some of the existing tests as MT tests 🤔
> 
> Yes, I wasn't sure either. We're of course massively benefiting that we use the exact same model creation and filling code that you already wrote all the tests for. Additionally, we have the challenge of MT scheduling, so even the best test case will have to deal with non-determinism and potentially still not test the relevant thing because all events end up on a single thread...

That's fair, not to forget about the fact that the behavior of the parallel writer is already tested in isolation. I'm okay with leaving this like this then!

enirolf (approved):
LGTM! Would probably be good to get a second approval from someone else as well :)

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -864,20 +868,41 @@ void ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::Initialize()
...
    }
-   fWriter->Fill(*fOutputEntry);
+   fillContext->Fill(*outputEntry);
+}
+
+void ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::FinalizeTask(unsigned int slot)
+{
+   // In principle we would not need to flush a cluster here, but we want to benefit from parallelism for compression.
+   // NB: RNTupleFillContext::FlushCluster() is a nop if there is no new entry since the last flush.
+   fFillContexts[slot]->FlushCluster();
```
vepadulano, line 899: With Snapshot to TTree, users can specify a value (expressed in number of entries) for the size of the output TTree clusters via [fAutoFlush](https://root.cern/doc/master/structROOT_1_1RDF_1_1RSnapshotOptions.html#a3a766c08900892477189398d61fdd214). I wonder how this line impacts this feature, which at the moment is not supported for the Snapshot to RNTuple, but I imagine it would be requested at some point.

vepadulano (approved):
LGTM, thanks a lot! I have one question left, does not need to be addressed in this PR specifically.

tree/dataframe/src/RDFSnapshotHelpers.cxx
pcanal:
```suggestion
   // Any synchronization needed are handled by the underlying `RNTupleParallelWriter`
   // which has exclusive access to its `TFile`.
   fillContext->Fill(*outputEntry);
```

hahnjo: I added a corresponding comment in `UntypedSnapshotRNTupleHelper::Initialize` instead of separately in `Exec` and `FinalizeTask`.

tree/dataframe/src/RDFSnapshotHelpers.cxx
hahnjo: Yes, that's a good point. In RNTuple, we have options to express the target cluster size, but no (automatic) setting for the number of entries per cluster. Note that `UntypedSnapshotTTreeHelperMT` is also questionable in that regard because it writes / flushes output trees in `FinalizeTask`. So for RDataFrame snapshotting to TTree in MT mode, it's already more of an upper guarantee, but I guess many cluster will actually be smaller than the user wants them...

This probably requires further discussion - I opened #19784 and will resolve this thread for the moment.

PR #19600, opened 2025-08-11
[WIP] Integrate testing to the Python wheels workflow

siliataider:


test_tutorials/test_tutorials.py
```diff
@@ -0,0 +1,70 @@
...
+            check=True,
+            timeout=60,
+            capture_output=True,
+            text=True
+        )
+    except subprocess.TimeoutExpired:
+        pytest.skip(f"Tutorial {tutorial} timed out")
+    except subprocess.CalledProcessError as e:
+        if e.returncode == -signal.SIGILL or e.returncode == 132:
+            pytest.skip(f"Skipping {tutorial.name} (illegal instruction on this platform)")
```
pcanal: Why are we skipping rather failing those test/tutorials?

siliataider: Just for my debugging, they will absolutely be marked as failed indeed

hahnjo:
hm, you may want to implement the equivalent of https://github.com/root-project/root/commit/771bf406a149. There were three jobs running for old versions of this PR...

siliataider:
> hm, you may want to implement the equivalent of [771bf406a149](https://github.com/root-project/root/commit/771bf406a149). There were three jobs running for old versions of this PR...

Thanks! #19818

.github/wheel-tests/requirements.txt
vepadulano, line 1: Do we strictly need this file? Can't we just use the top-level `requirements.txt`? I acknowledge that one has more packages than this test suite currently requires, but having only one `requirements.txt` makes maintenance easier, unless there is a strong reason to keep this one.

.github/wheel-tests/test_tutorials.py
vepadulano, line 1: I would place the `wheel-tests` folder inside `ROOT_DIR/test` instead of `ROOT_DIR/.github`

.github/wheel-tests/test_tutorials.py
```diff
@@ -0,0 +1,102 @@
...
+subdirs = ["analysis/dataframe", "analysis/tree", "hist", "io/ntuple", "roofit/roofit"]
+
+SKIP_TUTORIALS = {
+    "rf503_wspaceread.py",
+    "rf615_simulation_based_inference.py",
+    "rf618_mixture_models.py",
+    "ntpl004_dimuon.C",
+    "ntpl008_import.C",
+    "distrdf004_dask_lxbatch.py",
+}
```
vepadulano, line 26: The tutorials we skip should also have some documentation about the reason they are skipped. Ideally we wouldn't skip anyone, but that is not always possible. By quickly looking at this list, I can tell that `distrdf004_dask_lxbatch.py` needs to be skipped always because it only works on lxplus. For the two RNTuple tutorials, I guess they are skipped because they require reading remote data via HTTP and our wheels don't enable that yet. For the others I'm not sure what's the reason for skipping yet.

vepadulano (changes requested):
Great work! Having continuous testing enabled on the Python wheels will help towards a smooth deployment of the ROOT package on PyPI! I have left a couple of minor comments to be addressed before merging.

test/wheels/test_tutorials.py
```diff
@@ -0,0 +1,99 @@
...
+
+def test_cpp_tutorials_are_detected():
+    assert len(cpp_tutorials) > 0
+
+
+@pytest.mark.parametrize("tutorial", cpp_tutorials, ids=lambda p: p.name)
+def test_cpp_tutorial(tutorial):
+    try:
+        result = subprocess.run(
+            [sys.executable, "-c", f'import ROOT; ROOT.gROOT.ProcessLine(".x {tutorial}")'],
```
vepadulano: Can't we use directly the `root` executable provided with the wheel? This would also provide coverage for the executable itself on top of running the tutorials.

.github/wheel-tests/requirements.txt
siliataider: I think a reason manifested: the runners are running out of space and the workflow is always failing, for now I put it back #20487

PR #19602, opened 2025-08-11
Fix Crystal Ball integral near n = 1 for RooCBShape and RooCrystalBall

MartinDuyTat:
# This Pull request:
Fix bug in calculation of Crystal Ball integrals used in RooFit
## Changes
- Analytical integral of RooCBShape
- Analytical integral of RooCrystalBall
## Why?
In the integral of the Crystal Ball function, the analytical expression results in 0/0 when ```n = 1```. To circumvent this, when ```abs(n - 1) < 1e-5``` an alternative calculation is used which can be derived by a Taylor expansion.

In particular, the expression
```
r / (1.0 - n) * (std::pow(r / (b - tmin), n - 1.0) - std::pow(r / (b - tmax), n - 1.0));
```
is approximated to 
```
std::pow(r, n) * (log(b - tmin) - log(b - tmax));
```
This is done by expanding the quantity inside the parentheses to first order in ```n - 1```.

However, since there is a factor ```1.0 - n``` in front of the parentheses, this expression effectively becomes a zeroth order expansion in ```n - 1```. This is problematic because the pre-factor ```std::pow(r, n)``` is in fact a function of ```n```, with a non-zero first order term in ```n - 1```.

Therefore, the quantity inside the parentheses really should be expanded to second order in ```n - 1```, otherwise the whole expression will have an incorrect value at first order in ```n - 1```.

Here are plots that show the issue, and also the ```C++``` scripts that produced the plots. I have scanned the relevant functions across ```n = 1``` and one can clearly see that the current implementation is wrong, and this PR ensures that the gradient is correct near ```n = 1```.

[RooCrystalBall_fix.pdf](https://github.com/user-attachments/files/21713449/RooCrystalBall_fix.pdf)
[RooCBShape_fix.pdf](https://github.com/user-attachments/files/21713450/RooCBShape_fix.pdf)
[RooCBShape_fix.txt](https://github.com/user-attachments/files/21713468/RooCBShape_fix.txt)
[RooCrystalBall_fix.txt](https://github.com/user-attachments/files/21713469/RooCrystalBall_fix.txt)

## Checklist:
- [x] tested changes locally

### DIFF
--- a/math/mathcore/src/ProbFuncMathCore.cxx
+++ b/math/mathcore/src/ProbFuncMathCore.cxx
@@ -107,7 +107,7 @@ namespace Math {
          MATH_ERROR_MSG("crystalball_integral", "CrystalBall function not defined at alpha=0");
          return 0.;
       }
-      bool useLog = (n == 1.0);
+      bool useLog = (std::abs(n - 1.) < 1.0e-5);
       if (n <= 0)
          MATH_WARN_MSG("crystalball_integral", "No physical meaning when n<=0");
 
@@ -132,11 +132,13 @@ namespace Math {
          double B = r - abs_alpha;
 
          if (!useLog) {
-            intpow = A * (1 - std::pow(r / (B - z), n - 1)) / (n - 1);
+            intpow = A * (1. - std::pow(r / (B - z), n - 1.)) / (n - 1.);
          }
          else {
             // for n=1 the primitive of 1/x is log(x)
-            intpow = A * std::pow(r, n - 1) * (std::log(B - z) - std::log(r));
+            double log_B_z = std::log(B - z);
+            double log_r = std::log(r);
+            intpow = A * std::pow(r, n - 1.) * (log_B_z - log_r + 0.5 * (1. - n) * (log_B_z * log_B_z - log_r * log_r));
          }
          intgaus = sqrtpiover2 * (1. + ROOT::Math::erf(abs_alpha * oneoversqrt2));
       }


ferdymercury:
Thanks!
- Where is the check `abs(n-1)<1e-5` ?
- Can you also update math/mathcore/src/ProbFuncMathCore.cxx

MartinDuyTat:
>     * Where is the check `abs(n-1)<1e-5` ?
The check is done on lines 270 and 274 in ```roofit/roofit/src/RooCrystalBall.cxx``` and line 691 in ```roofit/roofitcore/inc/RooFit/Detail/MathFuncs.h```.

>     * Can you also update math/mathcore/src/ProbFuncMathCore.cxx
I don't think this one is necessary?  On line 113 I only see ```bool useLog = (n == 1.0)```, and at this single point ```n = 1``` I believe the logarithmic approximation is only needed at zeroth order.

ferdymercury:
> On line 113 I only see `bool useLog = (n == 1.0)`, and at this single point `n = 1` I believe the logarithmic approximation is only needed at zeroth order.

Thanks a lot for clarifying!
If you generate with your script the same canvas you attached, but using the function from ProbFundMathCore, does the function behave correctly in the range n = 1 - 1e-6 or does it show some artefacts ?
If it shows artefacts, we should try to make everything consistent by also changing bool useLog =  abs(n-1) < 1e-5 and adding 2nd order, as in the other two files.

MartinDuyTat:
> > On line 113 I only see `bool useLog = (n == 1.0)`, and at this single point `n = 1` I believe the logarithmic approximation is only needed at zeroth order.
> 
> Thanks a lot for clarifying! If you generate with your script the same canvas you attached, but using the function from ProbFundMathCore, does the function behave correctly in the range n = 1 - 1e-6 or does it show some artefacts ? If it shows artefacts, we should try to make everything consistent by also changing bool useLog = abs(n-1) < 1e-5 and adding 2nd order, as in the other two files.

Hi, for the function ```ROOT::Math::crystalball_integral``` I just scanned ```n - 1``` across the range ```[-1e-6, 1e-6]```, in ```1e-8``` increments, using these parameters:
```
x = -2.0
alpha = 0.5
sigma = 1.0
mean = 0.0
```
In the first attachment you can see that there's no weird behaviour, other than what looks like float-point errors. I assume these small float-point errors is the reason why the RooFit implementation uses the logarithmic approximation in some finite range around ```n = 1```. It looks like very small noise, as you can see in the second attachment where I've scanned around a larger range ```[-1e-5, 1e-5]```. But I haven't tested the function for different shape parameters and I don't know if there are any cases where these float-point errors become large.

[MathCore_crystalball_fix.pdf](https://github.com/user-attachments/files/21718034/MathCore_crystalball_fix.pdf)
[MathCore_crystalball_fix_wide_range.pdf](https://github.com/user-attachments/files/21718115/MathCore_crystalball_fix_wide_range.pdf)

guitargeek (changes requested):
Thank you @MartinDuyTat for this great PR!

For consistency, I would indeed suggest to use the same tolerance for the `TMath` function before merging this PR:
  * https://github.com/root-project/root/blob/master/math/mathcore/src/ProbFuncMathCore.cxx#L110

math/mathcore/src/ProbFuncMathCore.cxx
```diff
@@ -132,11 +132,14 @@ namespace Math {
...
 
          if (!useLog) {
-            intpow = A * (1 - std::pow(r / (B - z), n - 1)) / (n - 1);
+            intpow = A * (1. - std::pow(r / (B - z), n - 1.)) / (n - 1.);
          }
          else {
             // for n=1 the primitive of 1/x is log(x)
-            intpow = A * std::pow(r, n - 1) * (std::log(B - z) - std::log(r));
+	    double log_B_z = std::log(B - z);
+	    double log_r = std::log(r);
```
ferdymercury:
```suggestion
            double log_B_z = std::log(B - z);
            double log_r = std::log(r);
```

MartinDuyTat:
> Thank you @MartinDuyTat for this great PR!
> 
> For consistency, I would indeed suggest to use the same tolerance for the `TMath` function before merging this PR:
> 
>     * https://github.com/root-project/root/blob/master/math/mathcore/src/ProbFuncMathCore.cxx#L110

Hi @guitargeek I've just updated the ```TMath``` function to use the same tolerance and I've put in the same fix there as well. I'm also attaching a comparison between the original and new calculation.

[MathCore_crystalball_fix.pdf](https://github.com/user-attachments/files/21751025/MathCore_crystalball_fix.pdf)

math/mathcore/src/ProbFuncMathCore.cxx
MartinDuyTat: Hi @ferdymercury I'm not sure if it's my browser/editor or if I've missed something, but it doesn't seem to me like there are any changes here?

guitargeek: It seems you have used tabs for indentation, but we exclusively use spaces in the ROOT codebase, so I think @ferdymercury is suggesting to change that.

That's a nitpick, admittedly, but he's right :slightly_smiling_face:

MartinDuyTat: Ah, many thanks for pointing that out!

MartinDuyTat:
Hi @guitargeek @ferdymercury I see several checks are failing, but I don't think it's related to my changes? Do you need any further input from my side to merge this?

guitargeek (approved):
Thank you very much, also for addressing the comments!

PR #19605, opened 2025-08-11
[hist] Improve limits of THLimitsFinder.

hageboeck:
THLimitsFinder sometimes removes the first/last bin of an axis range. Here, it is ensured that the min and max of a buffered range is part of the axis range, and not removed by accident.
This fixes issues with TTree::Draw such as the one described in https://root-forum.cern.ch/t/bug-or-feature-in-ttree-draw/62862/

This supersedes #17689


ferdymercury (approved):
LGTM thanks!

I would propose this test:

```
TTree t;
Float_t x;
t.Branch("x", &x);
x = -999;
t.Fill();
x = 0;
t.Fill();
t.Draw("x");
auto h = (TH1*)gROOT->FindObject("htemp");
ASSERT_EQ(h->GetEntries(), h->GetEffectiveEntries());
```

hageboeck:
> LGTM thanks!
> 
> I would propose this test:
> 
> ```
> TTree t;
> Float_t x;
> t.Branch("x", &x);
> x = -999;
> t.Fill();
> x = 0;
> t.Fill();
> t.Draw("x");
> auto h = (TH1*)gROOT->FindObject("htemp");
> ASSERT_EQ(h->GetEntries(), h->GetEffectiveEntries());
> ```

@ferdymercury good test! It's added now.

ferdymercury (approved):
Just a nitpick at the end of DrawAutoBinning:

```
delete h;
delete gROOT->FindObject("c1");
```

hageboeck:
> Thanks! Just a nitpick at the end of DrawAutoBinning:
> 
> ```
> delete h;
> delete gROOT->FindObject("c1");
> ```

👍, updated for h. The canvas isn't created in batch mode, though, so that should be be unnecessary in this test.

ferdymercury:
> The canvas isn't created in batch mode, though, so that should be be unnecessary in this test.

Are you sure?

```
root -l -b
root [0] TTree t;
root [1] Float_t x;
root [2] t.Branch("x", &x);
root [3] x = -999;
root [4] t.Fill();
root [5] x = 0;
root [6] t.Fill();
root [7] t.Draw("x");
Info in <TCanvas::MakeDefCanvas>:  created default TCanvas with name c1
root [8] gROOT->FindObject("c1")
(TObject *) 0x5c740608b9c0
root [9] delete gROOT->FindObject("c1")
root [10] gROOT->FindObject("c1")
(TObject *) nullptr
```

hageboeck:
The test failures for the graphics are real. This will require more investigation.

hist/hist/src/THLimitsFinder.cxx
```diff
@@ -337,10 +337,16 @@ void THLimitsFinder::Optimize(Double_t A1,  Double_t A2,  Int_t nold ,
...
+   if (al - BinLow >= atest && al < BinLow + BinWidth) {
+      // Suppress the first bin, but only if al doesn't fall into it
+      BinLow += BinWidth;
+      nbins--;
+   }
+   if (BinHigh - ah >= atest && BinHigh - BinWidth > ah) {
+      // Suppress the last bin, but only if ah doesn't fall into it
+      BinHigh -= BinWidth;
+      nbins--;
+   }
```
ferdymercury: Shouldn't it be `al > BinLow + BinWidth` ? instead of `<` ?
Or maybe even easier:
```suggestion
   if (al >= BinLow + BinWidth + atest) {
      // Suppress the first bin, but only if al doesn't fall into it
      BinLow += BinWidth;
      nbins--;
   }
   if (ah <= BinHigh - BinWidth - atest) {
      // Suppress the last bin, but only if ah doesn't fall into it
      BinHigh -= BinWidth;
      nbins--;
   }
```

tree/tree/test/TTreeRegressions.cxx
```diff
@@ -279,3 +279,22 @@ TEST(TTreeRegressions, TTreeFormulaMemberIndex)
...
+   t.Branch("x", &x);
+   x = -999;
+   t.Fill();
+   x = 0;
+   t.Fill();
+   t.Draw("x");
+   auto h = (TH1 *)gROOT->FindObject("htemp");
+   ASSERT_NE(h, nullptr);
+   EXPECT_EQ(h->GetEntries(), h->GetEffectiveEntries());
+   delete h;
```
ferdymercury, line 309:
```suggestion
   delete h;
   delete gROOT->FindObject("c1");
```

hist/hist/src/THLimitsFinder.cxx
hageboeck: Yes, indeed.
But looking at this logic again, it's clear that it may truncate the axis range, so one shouldn't use it. At the same time, I cannot remove it (I tested), because this breaks the "looks" of other graphics components.
I think I now found a solution that works. It's already pushed.

hist/hist/src/THLimitsFinder.cxx
```diff
@@ -381,8 +383,8 @@ void THLimitsFinder::OptimizeLimits(Int_t nbins, Int_t &newbins, Double_t &xmin,
       xmin = -1;
       xmax = 1;
    } else {
-      xmin    = binlow;
-      xmax    = binhigh;
+      xmin = std::min(binlow, xmin);
+      xmax = std::max(binhigh, xmax + 0.01 * (xmax - xmin));
```
ferdymercury, line 386:
```suggestion
      xmax = std::max(binhigh, xmax + 1e-12 * (xmax - xmin));
```
Does it only work with 1% ?

hageboeck: Not exclusively, but I think it's actually good to slightly but noticeably extend the axis beyond the max value. 1% is of course an arbitrary number, but I really intended to add a visible margin. After all, this auto-binning would mostly be used for a first pass to have a look at the histogram, and afterwards, users will likely choose their own limits. 5% or 10% would probably also be numbers that work OK.

1E-12 on the other hand is a *very* tight margin, probably hard to discern visually. It might also be at risk of being eaten up by floating-point-precision issues if the range is already encompassing 3 orders of magnitude. So I would rather go larger than smaller with that margin.
What do you think?

ferdymercury: agreed, but it would be weird that there is a right margin but not a left margin? (That's why I was thinking of 1e-12 or 1e-6)
Anyway, both options work fine.

hist/hist/src/THLimitsFinder.cxx
```diff
@@ -381,8 +383,8 @@ void THLimitsFinder::OptimizeLimits(Int_t nbins, Int_t &newbins, Double_t &xmin,
       xmin = -1;
       xmax = 1;
    } else {
-      xmin    = binlow;
-      xmax    = binhigh;
+      xmin = std::min(binlow, xmin);
```
ferdymercury:
```suggestion
      xmin = std::min(binlow, xmin - 0.01 * (xmax - xmin));
```
Or leaving as is, but then changing below 0.01 to 0.001 so that the asymmetry is no longer visible by eye

hist/hist/test/test_TH1.cxx
```diff
@@ -54,6 +54,19 @@ TEST(THLimitsFinder, Degenerate)
    EXPECT_GE(xmax, centralValue + 5.);
 }
 
+// see https://root-forum.cern.ch/t/bug-or-feature-in-ttree-draw/62862
+// Due to a poor binning choice in THLimitsFinder, the histograms in
+// TTree::Draw might not contain contain all values.
```
guitargeek:
```suggestion
// TTree::Draw might not contain all values.
```
Just a typo

guitargeek (approved):
Thank you very much!

Issue #14770, opened 2024-02-20
TBuffer* classes should abort in case the 1GB limit is being hit


vepadulano:
### Explain what you would like to see improved and how.

See https://github.com/root-project/root/pull/14627#issuecomment-1953632146

### ROOT version

Any

### Installation method

Any

### Operating system

Any

ferdymercury:
There are also some issues pending in TString and TBufferFile constructor classes, that have things like:

```
TString::TString(const char *cs, Ssiz_t n)
TBuffer::Expand(Int_t newsize
TBuffer::TBuffer(EMode mode, Int_t bufsiz)
etc.
```

which could be made Long64_t and adding Fatal

vepadulano:
> which could be made Long64_t and adding Fatal

Makes sense! Re-opening this issue until those are fixed

Issue #6734, opened 2020-10-30
Overcome 1GB size limit for IO buffers


sawenzel:
### Explain what you would like to see improved

ROOT currently has a size limitation (~1GB) for the number of bytes
that can go into a single entry of a TBranch. This limitation is due to using
code based on 32bit integers in the IO mechanism.

For Run3, the ALICE experiment is turning from an event (singe collision) - based data processing
to a processing based on event-collections (timeframes). A timeframe becomes the basic data index. In consequence
this puts us into a situation in which we often need to store/serialize very large amounts of data per TBranch entries.

So far, we have worked around this by either giving up on ROOT serialization features, or writing specialized IO containers.
But it is foreseeable that the problem aggravates and a fix in the ROOT ecosystem would be a much better solution.

### Optional: share how it could be improved

Please improve on the ~1GB size limitation, which is does not correspond to modern memory capabilities and 64bit processing.
As a short term solution, gaining a factor of 2 would already help us. This might be possible by reviewing the use of int32 vs uint32 etc.

davidrohr:
To add my opinion: I think this is quite a central ROOT feature, and supporting larger buffers will become imperative at some point anyway, so I don't see a strong reason to delay such changes.
If backward compatibility is a concern, I think it would even be possible (or acceptable for ALICE) to create the file with a special flag or so, such that it cannot be opened with old ROOT versions.
At least the current behavior seems very outdated to me. The first time I ran into this 1GB limit, it was actually the last thing I thought of to suspect the buffer size, in particular since the limit is 1 GB not 4 GB.

cxwx:
1 GB limit the app for large skymap and earth map.

PR #19606, opened 2025-08-11
[io] Properly abort when buffer size overflows max integer

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/14770

And is a first step towards https://github.com/root-project/root/issues/6734

## Checklist:

- [x] tested changes locally


pcanal:
Can you make a separate PR with only the change in variable name (`bufsize`) to reduce the noise of this more challenging PR.

io/io/src/TDirectoryFile.cxx
```diff
@@ -24,6 +24,7 @@ End_Macro
 */
 
 #include <iostream>
+#include "RtypesCore.h"
```
ferdymercury:
```suggestion
```

io/io/src/TFile.cxx
```diff
@@ -81,6 +81,7 @@ The structure of a directory is shown in TDirectoryFile::TDirectoryFile
 </details>
 */
 
+#include "RtypesCore.h"
```
ferdymercury:
```suggestion
```

core/base/inc/TBuffer.h
```diff
@@ -78,16 +78,16 @@ class TBuffer : public TObject {
...
    Int_t    GetBufferVersion() const { return fVersion; }
    Bool_t   IsReading() const { return (fMode & kWrite) == 0; }
    Bool_t   IsWriting() const { return (fMode & kWrite) != 0; }
    void     SetReadMode();
    void     SetWriteMode();
-   void     SetBuffer(void *buf, UInt_t bufsize = 0, Bool_t adopt = kTRUE, ReAllocCharFun_t reallocfunc = nullptr);
+   void     SetBuffer(void *buf, Long64_t bufsize = 0, Bool_t adopt = kTRUE, ReAllocCharFun_t reallocfunc = nullptr);
    ReAllocCharFun_t GetReAllocFunc() const;
    void     SetReAllocFunc(ReAllocCharFun_t reallocfunc = nullptr);
    void     SetBufferOffset(Int_t offset = 0) { fBufCur = fBuffer+offset; }
```
jblomer: I think we should update this method in the same PR.

io/io/inc/TFile.h
```diff
@@ -404,8 +404,8 @@ class TFile : public TDirectoryFile {
 
    static void         SetFileBytesRead(Long64_t bytes = 0);
    static void         SetFileBytesWritten(Long64_t bytes = 0);
-   static void         SetFileReadCalls(Int_t readcalls = 0);
-   static void         SetReadaheadSize(Int_t bufsize = 256000);
+   static void         SetFileReadCalls(Long64_t readcalls = 0);
+   static void         SetReadaheadSize(Long64_t bytes = 256000);
```
jblomer, line 408: Maybe put those in another commit/PR.

jblomer (commented):
Cool, many thanks!

I think generally we want to use `std::size_t` for buffer sizes instead of `Long64_t`.

We should probably also update `TBuffer::[Read|Write]Buf`, `TBuffer::ReadString`, `TBuffer::MapObject`, `TBuffer::[Check|Set]ByteCount`, `TBuffer::SetBufferDisplacement`.

Maybe also `TBuffer::[Read|Write]Clones`.

Sometimes we are checking for `kMaxBufferSize`, sometimes for `kMaxInt`. Shouldn't we always check for `kMaxBufferSize`?

Regarding commit messages, I'd suggest

    [NFC] remove unused headers

and

    [io] accept and check long buffer size params

with an explanation why we (at this point) allow for long buffer sizes but then abort when they are actually used.

@pcanal: do we have an indication that the optimization of initializing the buffer size to the average buffer size seen so far in the file is actually useful? There are certainly write patterns where it hurts rather than helps. Removing this optimization would get us a fair amount of simplification in a number of read/write APIs.

ferdymercury:
> I think generally we want to use `std::size_t` for buffer sizes instead of `Long64_t`.

Even if that changes the data type from signed to unsigned?

Also, wouldn't it be better to have ULong64_t instead of std::size_t?

Since size_t is unsigned int for 32-bit targets and unsigned long long or unsigned long for 64-bit targets, depending on the C++ implementation and the compilation target. 
Where as ULong64_t would always be unsigned long long.

jblomer:
> Even if that changes the data type from signed to unsigned?
> 
> Also, wouldn't it be better to have ULong64_t instead of std::size_t?
> 
> Since size_t is unsigned int for 32-bit targets and unsigned long long or unsigned long for 64-bit targets, depending on the C++ implementation and the compilation target. Where as ULong64_t would always be unsigned long long.

In my opinion, that's the point. The `std::size_t` type is the integer type describing a length of something in memory. This is platform-dependent. On a 32bit platform, e.g., there will never be a buffer > 4GB.

ferdymercury:
> In my opinion, that's the point. The `std::size_t` type is the integer type describing a length of something in memory. This is platform-dependent. On a 32bit platform, e.g., there will never be a buffer > 4GB.

Sounds reasonable. What if a TTree in the future contains a big entry over 4GB ? Does it mean that it won't be read in 32-bits? How do we error out then, or it's silently cropped? Or is it going to be emulated as several buffers one after the other?
Using a data type larger than the actual type (ulong64 vs size_t) leaves room for detecting those kind of situations.
But maybe it's all very corner cases and is not worth?

jblomer:
> > In my opinion, that's the point. The `std::size_t` type is the integer type describing a length of something in memory. This is platform-dependent. On a 32bit platform, e.g., there will never be a buffer > 4GB.
> 
> Sounds reasonable. What if a TTree in the future contains a big entry over 4GB ? Does it mean that it won't be read in 32-bits? How do we error out then, or it's silently cropped? Or is it going to be emulated as several buffers one after the other? Using a data type larger than the actual type (ulong64 vs size_t) leaves room for detecting those kind of situations. But maybe it's all very corner cases and is not worth?

I think generally we have to distinguish between the in-memory buffer and what's serialized to disk. If a big atomic object (e.g., a histogram) is serialized to disk, it can't be read back on 32bit platforms. I think that's fine and unavoidable. The 32bit machine is simply not capable enough. On disk, of course, we will need to represent the size of objects in a platform-independent way. I think that the deserialization of the object length will be the proper point to throw errors.

Regarding the concrete on-disk representation, the plan is to chunk large objects in multiple keys to keep the changes to the TFile on-disk format minimal.

pcanal:
> do we have an indication that the optimization of initializing the buffer size to the average buffer size seen so far in the file is actually useful? 

This is hard to really measure for sure as it is of course very dependent of the actual workload.  When this was introduced, this was in direct reaction to issues related to not only memory fragmentation (increase of process virtual size due to the inability to re-use some memory that just a tad too small) but also thread scaling (by reducing the amount of memory allocation which requires (in most cases) the system to take a global lock).

pcanal:
> On a 32bit platform, e.g., there will never be a buffer > 4GB.

On the other hand we need to also make sure we probably error-out when there is a request for it ...

pcanal:
> What if a TTree in the future contains a big entry over 4GB ? ... Or is it going to be emulated as several buffers one after the other? 

That is the current plan.

io/io/src/TDirectoryFile.cxx
```diff
@@ -24,7 +24,6 @@ End_Macro
 */
 
 #include <iostream>
-#include "RtypesCore.h"
```
pcanal: Please, open a separate PR for this change.

ferdymercury: Sorry, it's not a change, I had somehow added it by accident and then removed it back. (Or viceversa). Probably as a consequence of a conflict during the rebase.

core/base/inc/TObject.h
```diff
@@ -170,8 +170,8 @@ class TObject {
    virtual void        SetDrawOption(Option_t *option="");  // *MENU*
    virtual void        SetUniqueID(UInt_t uid);
    virtual void        UseCurrentStyle();
-   virtual Int_t       Write(const char *name = nullptr, Int_t option = 0, Int_t bufsize = 0);
-   virtual Int_t       Write(const char *name = nullptr, Int_t option = 0, Int_t bufsize = 0) const;
+   virtual Int_t       Write(const char *name = nullptr, Int_t option = 0, Long64_t bufsize = 0);
```
pcanal, line 173: That might be necessary but is a serious problem.  This function is overload a lot in both our code but also very possibly in user code.  Unless those user have upgrade their code to use the `override` keyword (which is unlikely in my opinion), their code will compile correctly but do the wrong thing (revert to use the default behavior rather than their customization .... )

ferdymercury: What if we implement a dummy

`Int_t       Write(const char *name, Int_t option, Int_t bufsize) const final = delete;`

to trigger a compilation error, or at least a warning and avoid that silenced wrong behavior?

NOPE, this does not work because there will be ambiguous overload of default parameters with function above.

pcanal: That would indeed provoke a compilation error ....

ferdymercury: > their code will compile correctly but do the wrong thing (revert to use the default behavior rather than their customization .... )

I guess their code will probably not compile with an error due to ambiguous overload of default parameters

pcanal: The variation:
```
[virtual] Int_t       Write(const char *name, Int_t option, Int_t bufsize) const final { return Write(name, option, (Long64_t)bufsize); }
```
works well for function but is not quite enough for constructor.  There is no way to prevent the derived class from having its own constructor that still take an Int_t.   More annoyingly introducing/keeping the constructor with `Int_t` prevent the promotion-to and usage of the constructor with `Long64_t`.  So ideas like `= delete`-ing the `Int_t` constructor or having the `Int_t` constructor issue a deprecation message don't quite work out nicely.

pcanal: The constructor issue is probably not that important as it is only to set the initial size, which will still be limited to less than 1GB.  @ferdymercury could you add a new commit with the `final overload that delegates to the 64bits version` trick and to help the user, add a comment at the end of the lines, for example:

```
Int_t Write(const char *name, Int_t option, Int_t bufsize) const final // Please override the `Long64_t` overload instead of the `Int_t` version.
{ return Write(name, option, (Long64_t)bufsize); }
```
which (hopefully) we lead to error messages with the right hint:
```
<source>:13:18: error: declaration of 'write' overrides a 'final' function
   13 |     virtual void write(Int_t) { };
      |                  ^
<source>:7:18: note: overridden virtual function is here
    7 |     virtual void write(Int_t i) final // Please override the `Long64_t` overload instead of the `Int_t` version.
      |    
```

ferdymercury: That doesn't work either...
I had tried that 3 weeks ago, but then one gets flooded with warning of the type:

```
In file included from /opt/root_src/core/base/inc/TNamed.h:25,
                 from /opt/root_src/core/base/inc/TDirectory.h:24,
                 from /opt/root_src/core/base/inc/TROOT.h:28,
                 from /opt/root_src/core/metacling/src/rootclingTCling.cxx:28:
/opt/root_src/core/base/inc/TObject.h:176:18: warning: ‘virtual Int_t TObject::Write(const char*, Int_t, Int_t)’ was hidden [-Woverloaded-virtual]
  176 |    virtual Int_t Write(const char *name, Int_t option, Int_t bufsize) final
      |                  ^~~~~
In file included from /opt/root_src/core/cont/inc/TSeqCollection.h:25,
                 from /opt/root_src/core/cont/inc/TObjArray.h:25,
                 from /opt/root_src/core/meta/inc/TClass.h:30,
                 from /opt/root_src/core/base/inc/TDirectory.h:25,
                 from /opt/root_src/core/base/inc/TROOT.h:28,
                 from /opt/root_src/core/metacling/src/rootclingTCling.cxx:28:
/opt/root_src/core/cont/inc/TCollection.h:211:23: note:   by ‘TCollection::Write’
  211 |    Int_t              Write(const char *name = nullptr, Int_t option = 0, Long64_t bufsize = 0) const override;
      |                       ^~~~~
```

because of the default params.

What could work is to define the function without the "final" keyword. Then users would get a compilation error, if they used override on the Int_t version:
`marked ‘override’, but does not override`
However, if they didn't use the keyword override, then there would be no warning/error at all.

PR #19607, opened 2025-08-11
[core] reduce maxsize of string to prevent int overflow when calling Align

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes a alloc-size-larger compiler warning often seen in the CI of the PRs.

## Checklist:

- [ ] tested changes locally

This PR fixes #

### DIFF
--- a/core/base/inc/TString.h
+++ b/core/base/inc/TString.h
@@ -255,9 +255,9 @@ private:
    char          *GetPointer() { return IsLong() ? GetLongPointer() : GetShortPointer(); }
    const char    *GetPointer() const { return IsLong() ? GetLongPointer() : GetShortPointer(); }
 #ifdef R__BYTESWAP
-   static Ssiz_t  MaxSize() { return kMaxInt - 1; }
+   static Ssiz_t  MaxSize() { return kMaxInt - 1 - (kAlignment - 1); }
 #else
-   static Ssiz_t  MaxSize() { return (kMaxInt >> 1) - 1; }
+   static Ssiz_t  MaxSize() { return (kMaxInt >> 1) - 1 - (kAlignment - 1); }
 #endif
    void           UnLink() const { if (IsLong()) delete [] fRep.fLong.fData; }
    void           Zero() {


core/base/inc/TString.h
```diff
@@ -255,9 +255,9 @@ friend std::strong_ordering operator<=>(const TString &s1, const TString &s2) {
    char          *GetPointer() { return IsLong() ? GetLongPointer() : GetShortPointer(); }
    const char    *GetPointer() const { return IsLong() ? GetLongPointer() : GetShortPointer(); }
 #ifdef R__BYTESWAP
-   static Ssiz_t  MaxSize() { return kMaxInt - 1; }
+   static Ssiz_t  MaxSize() { return kMaxInt - 1 - (kAlignment - 1); }
```
pcanal: This seems the wrong direction.  It seems that `Align` is used in a single place:
```
static Ssiz_t  Recommend(Ssiz_t s) { return (s < kMinCap ? kMinCap : Align(s+1)) - 1; }
```
which I guess could be 'fixed' by doing:
```
static Ssiz_t  Recommend(Ssiz_t s) {
    auto aligned = Align(s+1)) - 1;
    if (s < kMinCap)
       return kMinCap;
    else if (aligned > kMaxInt)
       return MaxSize();
    else
       return aligned;
```

ferdymercury: Hmmm I kind of disagree.
As a user, if I call a public method called MaxSize() saying to me that I can have up to N characters in my string, if I then get an error when building a string of N - 8 characters, I would be surprised.

It can be fixed as you say, but then we would have to add docu saying that MaxSize is not actually MaxSize.

pcanal: Humm ... What about:
```
static Ssiz_t  Recommend(Ssiz_t s) {
    auto aligned = Align(s+1)) - 1;
    if (s < kMinCap)
       return kMinCap;
    else if (aligned > kMaxInt)
       return s;
    else
       return aligned;
}
```

>  if I then get an error when building a string of N - 8 characters, I would be surprised.

Both the version of `Recommend` I just wrote would let you do that (delta possible typos and off by hand errors)
In the first one if you pass `N-4` you would get `MaxSize()` which is 'greater than requested' so match the documentation.
In the second one, if you also pass `N-4`, you would get `N-4`.

ferdymercury: Ok, do you mean that having a misaligned TString will not pose problems?

ferdymercury: in any case, the problem is also that the signature of Align would have to be changed to return a Long64_t which is maybe not so ideal (performance?).

pcanal:
```
// `s` is expected to be under MaxSize() (kMaxInt-1)
static Ssiz_t  Recommend(Ssiz_t s) {
    if (s < kMinCap)
       return kMinCap;
    else if ((s+1) > (MaxSize() - kAlignment))
       return s;
    else 
       return Align(s+1)) - 1;
}
```
(also making `MaxSize` a constexpr function)

ferdymercury:
```suggestion
   static Ssiz_t  MaxSize() { return kMaxInt - 1; }
```

core/base/inc/TString.h
```diff
@@ -255,9 +255,9 @@ friend std::strong_ordering operator<=>(const TString &s1, const TString &s2) {
...
 #else
-   static Ssiz_t  MaxSize() { return (kMaxInt >> 1) - 1; }
+   static Ssiz_t  MaxSize() { return (kMaxInt >> 1) - 1 - (kAlignment - 1); }
```
ferdymercury:
```suggestion
   static Ssiz_t  MaxSize() { return (kMaxInt >> 1) - 1; }
```

core/base/src/TString.cxx
```diff
@@ -23,6 +23,14 @@ strings (<15 on 64-bit and <11 on 32-bit) are contained in the
 TString internal data structure without the need for mallocing the
 required space.
 
+\note TString can store a maximum of MaxSize()=2147483631 characters.
```
ferdymercury:
```suggestion
\note TString can store a maximum of MaxSize()=2147483646 characters.
```

core/base/src/TString.cxx
```diff
@@ -1220,11 +1228,11 @@ void TString::AssertElement(Ssiz_t i) const
 Ssiz_t TString::AdjustCapacity(Ssiz_t oldCap, Ssiz_t newCap)
 {
    Ssiz_t ms = MaxSize();
-   if (newCap > ms - 1) {
+   if (newCap > ms) {
       Fatal("TString::AdjustCapacity", "capacity too large (%d, max = %d)",
             newCap, ms);
    }
-   Ssiz_t cap = oldCap < ms / 2 - kAlignment ?
+   Ssiz_t cap = oldCap < ms / 2 ?
```
ferdymercury:
```suggestion
   Ssiz_t cap = oldCap < ms / 2 - kAlignment ?
```

core/base/inc/TString.h
ferdymercury: Done as suggested!

core/base/inc/TString.h
```diff
@@ -230,7 +230,15 @@ friend std::strong_ordering operator<=>(const TString &s1, const TString &s2) {
...
-   static Ssiz_t  Recommend(Ssiz_t s) { return (s < kMinCap ? kMinCap : Align(s+1)) - 1; }
+   // `s` is expected to be <= MaxSize() = (kMaxInt-1) ; `s + 1` includes the terminating nullchar
+   static Ssiz_t  Recommend(Ssiz_t s) {
+      if (s < kMinCap)
+         return kMinCap;
+      else if (s > MaxSize() - (kAlignment - 1))
+         return s;
+      else 
+         return Align(s + 1) - 1;
+   }
```
pcanal:
```suggestion
   // `s` is expected to be <= MaxSize() = (kMaxInt-1) ; `s + 1` includes the terminating nullchar
   static Ssiz_t  Recommend(Ssiz_t s)
   {
      if (s < kMinCap)
         return kMinCap;
      else if (s > MaxSize() - (kAlignment - 1))
         return s;
      else
         return Align(s + 1) - 1;
   }
```

core/base/inc/TString.h
```diff
@@ -255,9 +264,9 @@ friend std::strong_ordering operator<=>(const TString &s1, const TString &s2) {
    char          *GetPointer() { return IsLong() ? GetLongPointer() : GetShortPointer(); }
    const char    *GetPointer() const { return IsLong() ? GetLongPointer() : GetShortPointer(); }
 #ifdef R__BYTESWAP
-   static Ssiz_t  MaxSize() { return kMaxInt - 1; }
+   static constexpr Ssiz_t  MaxSize() { return kMaxInt - 1; }
```
pcanal:
```suggestion
   static constexpr Ssiz_t MaxSize() { return kMaxInt - 1; }
```

core/base/inc/TString.h
```diff
@@ -255,9 +264,9 @@ friend std::strong_ordering operator<=>(const TString &s1, const TString &s2) {
    char          *GetPointer() { return IsLong() ? GetLongPointer() : GetShortPointer(); }
    const char    *GetPointer() const { return IsLong() ? GetLongPointer() : GetShortPointer(); }
 #ifdef R__BYTESWAP
-   static Ssiz_t  MaxSize() { return kMaxInt - 1; }
+   static constexpr Ssiz_t MaxSize() { return kMaxInt - 1; }
 #else
-   static Ssiz_t  MaxSize() { return (kMaxInt >> 1) - 1; }
+   static constexpr Ssiz_t  MaxSize() { return (kMaxInt >> 1) - 1; }
```
pcanal:
```suggestion
   static constexpr Ssiz_t MaxSize() { return (kMaxInt >> 1) - 1; }
```

core/base/src/TString.cxx
```diff
@@ -1220,12 +1228,12 @@ void TString::AssertElement(Ssiz_t i) const
...
    Ssiz_t ms = MaxSize();
-   if (newCap > ms - 1) {
+   if (newCap > ms) {
       Fatal("TString::AdjustCapacity", "capacity too large (%d, max = %d)",
             newCap, ms);
    }
-   Ssiz_t cap = oldCap < ms / 2 - kAlignment ?
-                Recommend(std::max(newCap, 2 * oldCap)) : ms - 1;
+   Ssiz_t cap = oldCap <= ms / 2 ?
+                Recommend(std::max(newCap, 2 * oldCap)) : ms;
```
pcanal:
```suggestion
   Ssiz_t cap = oldCap <= ms / 2 ? Recommend(std::max(newCap, 2 * oldCap)) : ms;
```

core/base/src/TString.cxx
```diff
@@ -23,6 +23,14 @@ strings (<15 on 64-bit and <11 on 32-bit) are contained in the
 TString internal data structure without the need for mallocing the
 required space.
 
+\note TString can store a maximum of MaxSize()=2147483646 characters + 1 terminating null.
```
ferdymercury:
```suggestion
\note TString can store a maximum of MaxSize()=2147483646 characters; ie 2147483647 bytes if you include the terminating null.
```

PR #19611, opened 2025-08-11
[Python] Fix TMVA pythonizations when built with `dataframe=OFF`

guitargeek:
The registration of the TMVA RBatchGenerator pythonizations already works fine with `dataframe=OFF` in principle because it's inside a try-except block in `_facade.py`.

However, the `_rbatchgenerator` module that is conditional on `dataframe=ON` is imported globally in the `__init__.py` of the TMVA pythonizations, and this will still result in an uncaught exception.

Importing `_rbatchgenerator` locally fixes the problem.

Personally I build usually with `dataframe=OFF` but `roofit=ON` and `tmva=ON`, because I develop RooFit and occasionally do some TMVA stuff :slightly_smiling_face:


dpiparo:
Should this be ported to 636?

guitargeek:
Yes, nothing speaks against that!

PR #19593, opened 2025-08-09
Fix handling for CONST properties in CreateOpticalSurfaceN method

KoljaFrahm:
# This Pull request:

## Checklist:

- [x] tested changes locally
- [ ] updated the docs (if necessary)

This PR fixes https://github.com/root-project/root/issues/19592

PR #19612, opened 2025-08-12
[636] Fix handling for CONST properties in CreateOpticalSurfaceN method

dpiparo:
BP of https://github.com/root-project/root/pull/19593


agheata (approved):
Thanks!

PR #19613, opened 2025-08-12
[ci] Remove `rpath=ON` from global configuration

hahnjo:
Also remove non-working link for `rpath` deprecation, which is currently very confusing for users building from source.

### DIFF
--- a/.github/workflows/root-ci-config/buildconfig/global.txt
+++ b/.github/workflows/root-ci-config/buildconfig/global.txt
@@ -75,7 +75,6 @@ roofit_multiprocess=ON
 root7=ON
 rootbench=OFF
 roottest=ON
-roottest_force_checkout=OFF
 runtime_cxxmodules=ON
 shadowpw=OFF
 shared=ON


.github/workflows/root-ci-config/buildconfig/global.txt
```diff
@@ -76,7 +76,6 @@ root7=ON
 rootbench=OFF
 roottest=ON
 roottest_force_checkout=OFF
-rpath=ON
```
guitargeek, line 79: What about the `roottest_force_checkout` just one line above? Seems irrelevant too now, maybe we can also remove it in the PR?

cmake/modules/RootBuildOptions.cmake
```diff
@@ -408,7 +408,6 @@ endforeach()
 
 if(DEFINED rpath)
   message(DEPRECATION ">>> Option 'rpath' is deprecated and without effect."
-      " See https://root.cern/doc/v638/release-notes.html"
```
guitargeek, line 411: This link will be valid once 6.38 is out.

In the past, the release notes for the currently-developed release were available under this link already, which was quite convenient because then one can refer to the release notes without having intermediate dead links in ROOT `master`. @dpiparo, maybe that's worth to do again?

guitargeek (approved):
Thanks! I would prefer if the release note link would already work, like it was the case for past release cycles, but maybe that's also questionable because there is actually no release with the upcoming version number yet.

Removing the link is definitely fine too!

cmake/modules/RootBuildOptions.cmake
hahnjo: I'm removing it for now, we can always revisit later on.

ferdymercury:
Related: https://github.com/root-project/root/pull/19626
Feel free to take it over into your PR and close that one.

cmake/modules/RootBuildOptions.cmake
guitargeek: Sure!

Issue #11883, opened 2022-12-13
TTree indexes not correctly created when using branches made of vectors


midnightpride:
- [x] Checked for duplicates
### Describe the bug
If I try to build a TTreeIndex using a specific element of a (pointer to) vector stored in the tree (like `BuildIndex("values[5]")`) the expression is always evalued as 0. This happens both with `BuildIndex` and with the `TTreeIndex` constructor.
### Expected behavior
The expression used in the creation of the index should access correctly the given element of the vector (as it does when using the same expression in the `TTree::Scan` function)
### To Reproduce
I have prepared a small macro that creates a tree with a minimal structure and show how the indexing fails: https://cernbox.cern.ch/s/JXlFpygmNVVrbjP
### Setup
Tested with both root 6.26/08 and 5.34/25

ferdymercury:
Cross-posting reproducer:
```
#include "TFile.h"
#include "TTree.h"
#include "TTreeIndex.h"
#include <vector>

using namespace std;

int testIndex()
{
    const int nEvents = 1000;
    const size_t nElements = 30;
    TFile *f = TFile::Open("test.root", "recreate");
    vector<int> *otherNumbers = new vector<int>(nElements);
    int eventNumber;
    TTree *t = new TTree("tree", "tree");
    t->Branch("eventNumber", &eventNumber);
    t->Branch("otherNumbers", &otherNumbers);
    for (eventNumber = 0; eventNumber < nEvents; eventNumber++)
    {
        for (int j = 0; j < nElements; ++j)
        {
            (*otherNumbers)[j] = 2. * eventNumber + 3500 * j;
        }
        t->Fill();
    }
    f->Write();
    delete t;
    f->Close();
    delete f;
    delete otherNumbers;

    // Open in read mode
    f = TFile::Open("test.root");
    t = (TTree *)f->Get("tree");
    TTreeIndex *firstIndex = new TTreeIndex(t, "eventNumber", "otherNumbers[10]");
    firstIndex->Print("10"); // doesn't work: minor is always 0
    // use Scan to show what otherNumbers[10] contains
    t->Scan("eventNumber:otherNumbers[10]", "", "", 10);
    TTreeIndex *secondIndex = new TTreeIndex(t, "eventNumber", "otherNumbers");
    secondIndex->Print("10"); // minor is always otherNumbers[0]

    delete firstIndex;
    delete secondIndex;
    delete t;
    f->Close();
    delete f;
    return 0;
}
int main()
{
[...]

PR #19619, opened 2025-08-12
[tree] fix TTree::BuildIndex with a formula containing variable array index

ferdymercury:
# This Pull request:

## Changes or fixes:

fyi @amete @midnightpride

Fixes https://github.com/root-project/root/issues/11883

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/treeplayer/src/TTreeIndex.cxx
+++ b/tree/treeplayer/src/TTreeIndex.cxx
@@ -11,6 +11,8 @@
 
 /** \class TTreeIndex
 A Tree Index with majorname and minorname.
+
+If minorname is "0" (default arg in TTree::BuildIndex), just majorname will be used.
 */
 
 #include "TTreeIndex.h"


roottest/root/tree/index/runvarsizearr.C
```diff
@@ -0,0 +1,32 @@
+#include "TTree.h"
```
pcanal, line 1: Can you `clang-format` this file?

roottest/root/tree/index/runvarsizearr.C
```diff
@@ -0,0 +1,32 @@
...
+            otherNumbers[j] = - 2. * eventNumber + 3500 * j;
+        }
+        t.Fill();
+    }
+    std::cout << t.GetEntries() << std::endl;
+    // use Scan to show what otherNumbers[10] contains
+    t.Scan("eventNumber:otherNumbers", "", "", nEvents);
+
+    TTreeIndex firstIndex(&t, "otherNumbers[2]", "eventNumber");
+    firstIndex.Print("2"); // doesn't work: minor is always 0
```
pcanal: Is it supposed to not work or did it used to not work and now is fixed?

ferdymercury:
```suggestion
    firstIndex.Print("2"); // wasn't working before: major was always a garbage value
```

roottest/root/tree/index/runvarsizearr.C
```diff
@@ -0,0 +1,32 @@
...
+    TTreeIndex secondIndex(&t, "otherNumbers", "eventNumber");
+    secondIndex.Print("2"); // minor is always otherNumbers[0]
```
ferdymercury:
```suggestion
    secondIndex.Print("2"); // major is always otherNumbers[0] since index not specified (before and after fix)
```

roottest/root/tree/index/runvarsizearr.C
```diff
@@ -0,0 +1,32 @@
...
+    {
+        eventNumber = i;
+        for (size_t j = 0; j < nElements; ++j)
+        {
+            otherNumbers[j] = - 2. * eventNumber + 3500 * j;
+        }
+        t.Fill();
+    }
+    std::cout << t.GetEntries() << std::endl;
+    // use Scan to show what otherNumbers[10] contains
```
ferdymercury:
```suggestion
    // use Scan to show what otherNumbers[2] contains
```

roottest/root/tree/index/runvarsizearr.C
ferdymercury: Done

tree/treeplayer/src/TTreeIndex.cxx
```diff
@@ -11,6 +11,8 @@
 
 /** \class TTreeIndex
 A Tree Index with majorname and minorname.
+
+If minorname is "0", just majorname will be used.
```
vepadulano: Is this hard-coded somewhere, and this extra line is just for documentation?

tree/treeplayer/src/TTreeIndex.cxx
```diff
@@ -180,6 +182,14 @@ TTreeIndex::TTreeIndex(const TTree *T, const char *majorname, const char *minorn
...
          fMinorFormula->UpdateFormulaLeaves();
       }
+      if ((fMajorFormula->GetNdata() + fMinorFormula->GetNdata()) <= 0) {
+         // Calling GetNdata is essential before calling EvalInstance, otherwise a wrong
+         // result is silently returned by EvalInstance below if formula is value from a variable-sized array
+         // We raise an error to prevent the if clause being optimized-out if we do not use the return
+         Error("TTreeIndex",
+                 "In tree entry %lld, Ndata in formula is zero for both '%s' and '%s'", i,
+                 fMajorName.Data(), fMinorName.Data());
+      }
```
vepadulano, line 191: Here the `GetNdata()` calls are to make sure that for the current event the number of elements of the collection in either fMajorFormula o fMinorFormula is greater than zero?

roottest/root/tree/index/runvarsizearr.C
```diff
@@ -0,0 +1,30 @@
...
+      t.Fill();
+   }
+   std::cout << t.GetEntries() << std::endl;
+   // use Scan to show what otherNumbers[2] contains
+   t.Scan("eventNumber:otherNumbers", "", "", nEvents);
+
+   TTreeIndex firstIndex(&t, "otherNumbers[2]", "eventNumber");
+   firstIndex.Print("2"); // wasn't working before: major was always a garbage value
+   TTreeIndex secondIndex(&t, "otherNumbers", "eventNumber");
+   secondIndex.Print("2"); // major is always otherNumbers[0] since index not specified (before and after fix)
```
vepadulano: It's not clear to me what the "expected" behaviour is here. The collection `otherNumbers` always has 3 elements, so `otherNumbers[2]` should always be a valid value. So what was the previously faulty behaviour that this test is checking now?

tree/treeplayer/src/TTreeIndex.cxx
ferdymercury: Thanks for the review !

This information serves as documentation, and was before in the User-guide, which is now deprecated.
_(To build an index with only `majorname`, specify `minorname="0"` (default).)_

Thus I thought of rescuing that information here as Doxygen format.

Nothing changed in the underlying code: This all happens because of the default argument we have in the TTree header:

`virtual Int_t           BuildIndex(const char *majorname, const char *minorname = "0", bool long64major = false, bool long64minor = false);`

So minorname formula returns always 0, so std::sort does nothing, ie effectively we are just taking into account majorname, which is a mandatory parameter, unlike minorname which is by default ignored ("0").

tree/treeplayer/src/TTreeIndex.cxx
ferdymercury: Yep, you are right.
Note also that GetNdata must always be called for multi-dim formulas before calling EvalInstance, even if one does not care about the exact return value.

roottest/root/tree/index/runvarsizearr.C
ferdymercury: Left: before, had a wrong value of otherNumbers[2], Right: after, correct value and properly sorted by major key (note event 1 and 0 now in inverse order)

Image

The secondIndex.Print statement was already correct before, it does not change, but I left it to show that the newly introduced code does not break that.

tree/treeplayer/src/TTreeIndex.cxx
vepadulano: Thank you, I agree it's a good idea to resurrect this piece of documentation. Since it's in principle not connected to fixing the linked issue, could you just move this into a separate commit of this PR?

ferdymercury:
Green now :)

PR #19620, opened 2025-08-12
[CPyCppyy] Add pythonizations for `string_view`

### LINKED ISSUE #19476: `std::string_view` adds characters to strings in Python
### Check duplicate issues.

- [x] Checked for duplicates

### Description

With cppyy 3.5.0, having a `std::string_view` will append `[N]`, with `N` being the number of characters of the string when printing a `std::string_view`.

There is `s.data()` that will give the expected string, but that means if getting the value from a function, I need to know if it's a string or a string view, while this is not needed in C++. I would say this is unexpected, I would expect comparisons and usage of `std::string_view` to be similar to `std::string`.

This seems not be an issue from `cppyy` standalone: https://github.com/wlav/cppyy/issues/313#issuecomment-3133909956

### Reproducer

``` python
import cppyy
cppyy.cppdef(
"""
std::string_view s = "Hello, World!";
""")

from cppyy.gbl import s
print(s)
```
The answer is `"Hello, World!"[13]`. But then, this means I can not compare a `std::string_view` to a string in python, because this will not work without the `[N]`:
``` python
print(s == "Hello, World!")  # Should print True, prints False
print(s == "Hello, World![13]")  # Also False
print(s == '"Hello, World!"[13]')  # Also False
print(str(s) == '"Hello, World!"[13]')  # This is True
cppyy.cppdef(
"""
bool is_equal_cpp = s == "Hello, World!";
"""
)
from cppyy.gbl import is_equal_cpp
print(is_equal_cpp)  # Should print True, prints True
```

### ROOT version

```
  | Welcome to ROOT 6.36.02                        https://root.cern |
  | Built for linuxx8664gcc on Jul 26 2025, 08:01:37                 |
  | From tags/v6-36-02@v6-36-02                                      |
  | With g++ (GCC) 15.1.1 20250425                                   |
```

### Installation method

Built from source, Spack for the Key4hep stack, LCG stacks

### Operating system

Linux, Alma 9, Arch Linux

### Additional context

_No response_

guitargeek:
This also makes sure that the `string_view` converts to Python strings correctly.

Closes #19476.

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/Pythonize.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/Pythonize.cxx
@@ -1136,15 +1136,15 @@ static int PyObject_Compare(PyObject* one, PyObject* other) {
 }
 #endif
 static inline
-PyObject* CPyCppyy_PyString_FromCppString(std::string* s, bool native=true) {
+PyObject* CPyCppyy_PyString_FromCppString(std::string_view s, bool native=true) {
     if (native)
-        return PyBytes_FromStringAndSize(s->data(), s->size());
-    return CPyCppyy_PyText_FromStringAndSize(s->data(), s->size());
+        return PyBytes_FromStringAndSize(s.data(), s.size());
+    return CPyCppyy_PyText_FromStringAndSize(s.data(), s.size());
 }
 
 static inline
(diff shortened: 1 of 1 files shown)


dpiparo:
a build seemed affected by a network glitch. I restarted it to double-check ...

dpiparo (approved):

PR #19621, opened 2025-08-12
[roottest] Make `ALICE-18972` test conditional on `geom`

guitargeek:
The test includes the TGeomManager, so it should be only added if ROOT was built with `geom=ON`.


roottest/cling/parsing/ALICE-18972/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 ROOTTEST_GENERATE_REFLEX_DICTIONARY(ALICE_18972 mylib.h SELECTION mylib_sel.xml)
 
-ROOTTEST_ADD_TEST(ALICE_18972
-                  MACRO run_ALICE_18972.C
-                  DEPENDS ${GENERATE_REFLEX_TEST})
-
+if(ROOT_geom_FOUND)
```
pcanal, line 3:
```suggestion
if(geom)
```
is what we seem to use in most other places.

pcanal (approved):
LGTM.

roottest/cling/parsing/ALICE-18972/CMakeLists.txt
guitargeek: In `roottest`, we're using the flags that are defined when you're including `ROOT` as a package in CMake, because `roottest` used to be an independent project

pcanal: I think we should progressively (or faster if need be) converge on using a single notation since now roottest is no longer used as independent project.

PR #19472, opened 2025-07-29
Add UHI integration for 8 histogram tutorials and update index.md documentation

senabibi:
# This Pull request:
## Changes or fixes:
1. **Core Implementation**:
   - Integrated UHI into 8 histogram tutorials

2. **Documentation**:
   - Added Doxygen documentation for all UHI-integrated tutorials

## Checklist:
- [x] Tested changes locally (all 8 tutorials)
- [x] Updated documentation (Doxygen + tutorial index)

PR #19623, opened 2025-08-12
[tutorials] Fix vetos for UHI tutorials

guitargeek:
Following up on #19472. One can't do `file(GLOB ..)` with two separate output variables. Also, don't claim that the tutorial `hist002_TH1_fillrandom_userfunc_uhi.py` requires mplhep and matplotlib, because it only requires NumPy.

### DIFF
--- a/tutorials/CMakeLists.txt
+++ b/tutorials/CMakeLists.txt
@@ -934,8 +934,14 @@ if(ROOT_pyroot_FOUND)
       machine_learning/RBatchGenerator_TensorFlow.py
   )
 
-  file(GLOB requires_matplotlib requires_mplhep RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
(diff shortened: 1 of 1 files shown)


dpiparo:
Thanks for the PR. Its description is also about properly expressing dependencies on python modules, however the commit does not seem to reflect that: am I misunderstanding something?

tutorials/CMakeLists.txt
```diff
@@ -890,6 +890,7 @@ if(ROOT_pyroot_FOUND)
       analysis/dataframe/df017_vecOpsHEP.py
       analysis/dataframe/df026_AsNumpyArrays.py
       analysis/dataframe/df032_RDFFromNumpy.py
+      hist/hist002_TH1_fillrandom_userfunc_uhi.py
```
siliataider:
```suggestion
```
`hist/hist002_TH1_fillrandom_userfunc_uhi.py` is already present on line 917 (below)

Issue #13122, opened 2023-06-29
`TF1::DrawCopy()` does not copy histogram properties


linev:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

If TF1::DrawCopy() is used to draw the function, all axis settings (which are stored in histogram) are lost.

### Reproducer

This happens for instance in `tutorials/math/Bessel.C` macro where configured axis titles are not displayed.

### ROOT version

master and all recent releases

### Installation method

build from source

### Operating system

OpenSUSE Linux

couet:
I'll check. Thanks for the report

couet:
I am not sure how to reproduce this issue. I did the following:
1) run the macro Bessel.C
2) Save the canvas to a root file
3) reopen root with that saved file and display the canvas
=> The plots are exactly the same as when I execute the macro directly

linev:
Try to store canvas from `tutorials/graphs/annotation3d.C` macro.

It modifies axes properties in the macro itself - like title or some special labels.

You can do same with `Bessel.C` - but interactively before storing canvas to ROOT file

couet:
I see that with `annotation3d.C` the changed labels are not stored in the canvas when it is displayed back. But it has nothing to do with `DrawCopy` because this macro uses only `Draw`.

couet:
It fails already now. I am preparing a PR for the  `.C` case first ...

couet:
This one fixes the .C part: https://github.com/root-project/root/pull/13228

couet:
Now we have:
```
auto f = new TF1("x","x",-1,1);
f->Draw();
f->GetXaxis()->ChangeLabel(5, -1, -1, -1, -1, -1,"CHANGED");
c1->SaveAs("c1.C");
c1->SaveAs("c1.root");
```
The changed label is saved in the .C file but not in the root file.

couet:
The Streamer saves only the TF1 data. That's why the changed labels are not saved.
see: https://github.com/root-project/root/blob/333eaf379647157944e54f6f3317b3b586592406/hist/hist/src/TF1.cxx#L3589

couet:
Indeed `TF1` the data member `fHistogram` is not persistent (see `TF1.h`) this member holds the axis. Therefore the axis are not persistent too and not saved in the root file. That's a design choice to not make `fHistogram` persistent. I do not think that ca be changed.

couet:
[...]

PR #19627, opened 2025-08-13
Fix TF1::DrawCopy issue

couet:
TF1::Copy didn't copy fHistogram if it exited.
It fixes this issue: https://github.com/root-project/root/issues/13122

### DIFF
--- a/hist/hist/src/TF1.cxx
+++ b/hist/hist/src/TF1.cxx
@@ -1029,7 +1029,13 @@ void TF1::Copy(TObject &obj) const
    ((TF1 &)obj).fParMax    = fParMax;
    ((TF1 &)obj).fParent    = fParent;
    ((TF1 &)obj).fSave      = fSave;
-   ((TF1 &)obj).fHistogram = nullptr;
+   if (fHistogram) {
+      auto *h1 = (TH1*)fHistogram->Clone();
+      h1->SetDirectory(nullptr);
+      ((TF1 &)obj).fHistogram = h1;
+   } else {
+      ((TF1 &)obj).fHistogram = nullptr;
+   }
    ((TF1 &)obj).fMethodCall = nullptr;
    ((TF1 &)obj).fNormalized = fNormalized;
    ((TF1 &)obj).fNormIntegral = fNormIntegral;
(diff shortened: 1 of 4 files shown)


dpiparo:
BP to 636 when all tests are passing?

hist/hist/src/TF1.cxx
```diff
@@ -1368,6 +1368,7 @@ TF1 *TF1::DrawCopy(Option_t *option) const
    Copy(*newf1);
    newf1->AppendPad(option);
    newf1->SetBit(kCanDelete);
+   if (fHistogram) newf1->SetHistogram(fHistogram);
```
ferdymercury: Thanks!

Isn't this solution without a call to Clone() dangerous? It's copying the pointer address, no the histogram parameters.
If one deletes the new TF1, the old TF1 will be broken since fHistogram will be deleted?

I would also suggest moving this statement to TF1::Copy()

as well as creating a unit test as the one you wrote in the Issue.

couet: May be instead of Setting the original histogram we can set a copy of it. 
I agree with you 2 other points.

couet:
It needs more checks. It does not work yet.

ferdymercury:
> It needs more checks. It does not work yet.

Interesting. I thought that copy-pasting

```
   if (fHistogram) {
      obj->fHistogram = (TH1*)fHistogram->Clone();
      obj->fHistogram->SetDirectory(nullptr);
   }
```
   
   as done in `TF1::Clone` should do the job.

ferdymercury (changes requested):
Thanks a lot!

Could you add this gtest?

```
TEST(TF1, DrawCopy)
{
   auto f = new TF1("fa1","x*x",-3,3);
   f->GetXaxis()->SetTitle("xtitle");
   auto fcopy = f->DrawCopy();
   EXPECT_EQ(fcopy->GetXaxis()->GetTitle(), "xtitle");
}
```

couet:
indeed I am on it

ferdymercury (approved):
Thanks a lot! LGTM

ferdymercury (approved):

couet:
It looks like the failures are not related to this PR

PR #19618, opened 2025-08-12
[math][genvector] Implement pair acoplanarity and asymmetry functions

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/19338

Fyi @Rik002

## Checklist:

- [ ] tested changes locally
- [x] updated the docs (if necessary)

PR #19629, opened 2025-08-13
[math] Remove the GenVector namespace to improve doxygen.

hageboeck:
With free functions being added to for the GenVector package in #19618, it will be a problem to find these functions in the ROOT documentation. This is because of a name clash between the GenVector namespace vs the GenVector doxygen group.
Doxygen links to the former, whereas the latter has all the documentation. Since the namespace was only wrapping the Throw function, GenVector::Throw can be replaced by GenVector_Throw, and the namespace can be removed. This should make the doxygen documentation more useful.


guitargeek (approved):
Thanks, that's a very good idea!

PR #19630, opened 2025-08-13
[CMake]: Fix the -Dcheck_connection=OFF setting

andresailer:
If we set check_connection=OFF we should not then check the connection usually we should use return(), but we are in a macro so we cannot use that here, so we need to wrap things in else()

# This Pull request:

## Changes or fixes:

Fixes things when `-Dcheck_connection=OFF` is set

## Checklist:

- [x] tested changes locally

This PR fixes SPI builds with bad CERN network for root.cern or something.


andresailer:
* This is so great to test, because getting the file just randomly works...
* The diff is more obvious if you disable white-space changes: https://github.com/root-project/root/pull/19630/files?diff=unified&w=1

PR #19632, opened 2025-08-13
[ci] Set `check_connection=OFF` globally

guitargeek:
This way, we are not affected by problems downloading the connection check file from `root.cern`.

### DIFF
--- a/.github/workflows/root-ci-config/buildconfig/global.txt
+++ b/.github/workflows/root-ci-config/buildconfig/global.txt
@@ -34,6 +34,7 @@ builtin_zlib=OFF
 builtin_zstd=OFF
 ccache=ON
 cefweb=OFF
+check_connection=OFF
 clad=ON
 clingtest=OFF
 cocoa=OFF


andresailer:
Some tests are going to fail downloading files from root.cern if that server has issues.

guitargeek:
That's true, but at least it will actually build and run the tests :slightly_smiling_face:

Thanks for having tried and fixed this flag, by the way!

dpiparo:
thanks for this upgrade. Could this be documented in the 6.38 flags here https://root.cern/install/build_from_source/#all-build-options ?

dpiparo (approved):

guitargeek:
Sure! https://github.com/root-project/web/pull/1143

pcanal:
> Set check_connection=OFF globally

Do we still test `check_connection=ON` somewhere?

guitargeek:
No. But you're right, we should do it at least on one platform. I'll change that

PR #19637, opened 2025-08-13
[cling][AST] Desugar TemplateName before fully qualifying it

devajithvs:
TemplateArgument can return a QualType that may contain unqualified parts.

Before (incorrect):
  __gnu_cxx::__common_pool_policy<__pool, true>

After (correct):
  __gnu_cxx::__common_pool_policy<__gnu_cxx::__pool, true>

This appeared during LLVM20 upgrade in normalization of types like:

`LHCb::FastAllocVector<int,__gnu_cxx::__mt_alloc<int,__gnu_cxx::__common_pool_policy<__gnu_cxx::__pool,true>`

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes #

### DIFF
--- a/interpreter/cling/lib/Utils/AST.cpp
+++ b/interpreter/cling/lib/Utils/AST.cpp
@@ -849,7 +849,14 @@ namespace utils {
 
     if (arg.getKind() == TemplateArgument::Template) {
       TemplateName tname = arg.getAsTemplate();
-      // Note: should we not also desugar?
+      // Desugar before fully qualifying.
+      if (std::optional<TemplateName> UnderlyingOrNone =
+              tname.desugar(/*IgnoreDeduced=*/false)) {
+        if (*UnderlyingOrNone != tname) {
+          tname = *UnderlyingOrNone;
+          changed = true;
+        }
+      }
       changed = GetFullyQualifiedTemplateName(Ctx, tname);
       if (changed) {
         arg = TemplateArgument(tname);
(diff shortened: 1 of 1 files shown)


devajithvs:
I'm trying to find a way to add a test case. But there doesn't seem to be an easy way to create an unqualified type in a test.

devajithvs:
Closing this and moving this to the main LLVM20 PR.

pcanal:
> But there doesn't seem to be an easy way to create an unqualified type in a test.

I wonder if this does it:
```
namespace test__gnu_cxx {
   struct __pool {};
   template <typename P = __pool, bool C = true> struct test__common_pool_policy {};
   template <typename T, typename P = test__common_pool_policy<> > struct test__mt_alloc {};
}
namespace test_LHCb
{
    template <typename T, typename A = test__gnu_cxx::test__mt_alloc<int>>
    struct FastAllocVector {};
}
```

interpreter/cling/lib/Utils/AST.cpp
```diff
@@ -1286,8 +1293,20 @@ namespace utils {
...
-          // Note: should we not also desugar?
-          bool changed = GetFullyQualifiedTemplateName(Ctx, tname);
+          bool changed = false;
+
+          // In some case tname can be unqualified, desugar first.
+          // The following example fixed:
+          // __common_pool_policy<__pool,true> ->
+          // __common_pool_policy<__gnu_cxx::__pool, true>
+          if (std::optional<TemplateName> UnderlyingOrNone =
+                  tname.desugar(/*IgnoreDeduced=*/false)) {
```
pcanal, line 1303: @devajithvs Rather than 'desugar' directly, we may or may not need 'GetPartiallyDesugaredTypeImpl`. i.e When testing, also test that we end up with the intended result with `Double32_t`

PR #19639, opened 2025-08-14
[test] reset branch address before pointer goes out of scope

ferdymercury:
and do not call scan if tree is nullptr

Found out by asan in https://github.com/root-project/root/actions/runs/16941754812/job/48012438974?pr=19636


dpiparo:
Thanks.

PR #19640, opened 2025-08-14
[ntuple] Renormalize type name before comparing [v6.36]

hahnjo:
Backport of #19444, pending confirmation that this fixes the issue in a nightly build of Athena


jblomer (approved):

hahnjo:
The fix works with the latest Athena built using ROOT `master`, so I think we can merge this backport and release with 6.36.04.

dpiparo:
Thanks for these changes.

PR #19642, opened 2025-08-14
[df] Fix ranges assignment logic for WithGlobalRanges cases

martamaja10:
This was motivated by the sporadically failing MT version of the datasetspec test. A few slots were working on the same range of events (this became visible while we had to skip multiple files before accessing the file with an entry from the GlobalRange). A few other minor glitches are fixed as well.


tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -727,29 +727,30 @@ std::vector<std::pair<ULong64_t, ULong64_t>> ROOT::RDF::RNTupleDS::GetEntryRange
          // The first case
          if (fGlobalEntryRange->first >= start && fGlobalEntryRange->second <= end) {
             fOriginalRanges.emplace_back(start, end);
-            nEntriesPerRange += fGlobalEntryRange->second - fGlobalEntryRange->first;
-            ranges.emplace_back(fGlobalEntryRange->first, fGlobalEntryRange->second);
             fFirstEntry2RangeIdx[fGlobalEntryRange->first] = i;
+            ranges.emplace_back(fGlobalEntryRange->first, fGlobalEntryRange->second);
+            nEntriesPerRange += fGlobalEntryRange->second - fGlobalEntryRange->first;
```
hahnjo, line 731: (note to future me: this is only moving the statements)

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -727,29 +727,30 @@ std::vector<std::pair<ULong64_t, ULong64_t>> ROOT::RDF::RNTupleDS::GetEntryRange
          // The first case
          if (fGlobalEntryRange->first >= start && fGlobalEntryRange->second <= end) {
```
hahnjo, line 727: (note to future me: this case is for `fGlobalEntryRange` is entirely within the current file)

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -727,29 +727,30 @@ std::vector<std::pair<ULong64_t, ULong64_t>> ROOT::RDF::RNTupleDS::GetEntryRange
...
-            ranges.emplace_back(fGlobalEntryRange->first, fGlobalEntryRange->second);
             fFirstEntry2RangeIdx[fGlobalEntryRange->first] = i;
+            ranges.emplace_back(fGlobalEntryRange->first, fGlobalEntryRange->second);
+            nEntriesPerRange += fGlobalEntryRange->second - fGlobalEntryRange->first;
          }
 
          // The second case:
-         else if (fGlobalEntryRange->first <= end && fGlobalEntryRange->second > end) {
+         // The `fGlobalEntryRange->first <= end` condition is to distinguish this case from the 4th case.
+         else if (fGlobalEntryRange->second > end && fGlobalEntryRange->first <= end) {
```
hahnjo: (note for future me: this is only changing the order of two conditions. This case is taken if `fGlobalEntryRange` extends past the current file (`fGlobalEntryRange->second > end`). And because we process files in order and adjust `fGlobalEntryRange` below, I think we are guaranteed `assert(start <= fGlobalEntryRange->first)`).

I think we may need to be more strict here and change to `fGlobalEntryRange->first < end` because otherwise we have an empty range in the current file if they are equal...

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -727,29 +727,30 @@ std::vector<std::pair<ULong64_t, ULong64_t>> ROOT::RDF::RNTupleDS::GetEntryRange
...
             fOriginalRanges.emplace_back(start, end);
             fFirstEntry2RangeIdx[fGlobalEntryRange->first] = i;
             ranges.emplace_back(fGlobalEntryRange->first, end);
             std::optional<std::pair<ULong64_t, ULong64_t>> newvalues({end, fGlobalEntryRange->second});
             fGlobalEntryRange.swap(newvalues);
          }
          // The third case, needed to correctly quit processing if we only stay in the first file
-         else if (fGlobalEntryRange->first < start) {
+         else if (fGlobalEntryRange->second < start) {
```
hahnjo, line 744: (note to future me: this case is (now corrected) for quitting when `fGlobalEntryRange` ends before the `start` of the current file)

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -727,29 +727,30 @@ std::vector<std::pair<ULong64_t, ULong64_t>> ROOT::RDF::RNTupleDS::GetEntryRange
...
             return ranges;
          }
 
          // The fourth case:
          else if (fGlobalEntryRange->first > end) {
```
hahnjo: (note to future me: this case is for skipping a file if `fGlobalRange` starts after the `end` of the current file)

Shouldn't the condition be
```suggestion
         // The fourth case:
         else if (fGlobalEntryRange->first >= end) {
```
?

tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -727,29 +727,30 @@ std::vector<std::pair<ULong64_t, ULong64_t>> ROOT::RDF::RNTupleDS::GetEntryRange
...
-            fFirstEntry2RangeIdx[0] = i;
             fOriginalRanges.emplace_back(start, end);
-            ranges.emplace_back(0, 0);
+            fFirstEntry2RangeIdx[start] = i;
+            ranges.emplace_back(start, end);
```
hahnjo: Why do we need to create a range in this case? And why is it non-empty, won't this result in processing the entire file?

martamaja10: yes, you're right, only the map needs to be changed and the ranges should remain (0,0), I will add checking number of processed entries to the test actually

tree/dataframe/src/RNTupleDS.cxx
martamaja10: correct! Will fix

tree/dataframe/src/RNTupleDS.cxx
martamaja10: yes, you're right, thanks for catching this.

tree/dataframe/src/RNTupleDS.cxx
hahnjo: But aren't we back to square one now with `ranges.emplace_back(0, 0)`? That will cause `firstEntry = 0` in `InitSlot` and then operate on the same page source...

PR #19643, opened 2025-08-14
[hist] Implement `RBinIndex`

hahnjo:
This is a typed version of a bin number, with special values for underflow and overflow bins.

### DIFF
--- a/hist/histv7/inc/ROOT/RBinIndex.hxx
+++ b/hist/histv7/inc/ROOT/RBinIndex.hxx
@@ -50,10 +50,17 @@ public:
 
    RBinIndex &operator+=(std::size_t a)
    {
-      assert(IsNormal());
-      fIndex += a;
-      // TODO: is it possible to catch overflows?
-      assert(IsNormal());
+      if (!IsNormal()) {
+         // Arithmetic operations on special values go to InvalidIndex.
+         fIndex = InvalidIndex;
+      } else {
+         std::size_t old = fIndex;
+         fIndex += a;
+         if (fIndex < old || !IsNormal()) {
+            // The addition wrapped around, go to InvalidIndex.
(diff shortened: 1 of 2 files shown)


hist/histv7/inc/ROOT/RBinIndex.hxx
```diff
@@ -0,0 +1,144 @@
...
+   /// Construct a bin index for a normal bin.
+   RBinIndex(std::size_t index) : fIndex(index) { assert(IsNormal()); }
+
+   /// Return the index for a normal bin.
+   std::size_t GetIndex() const
+   {
+      assert(IsNormal());
+      return fIndex;
+   }
+   bool IsNormal() const { return fIndex < UnderflowIndex; }
```
hageboeck, line 46: Would it make sense to repeat here in a doxygen string what the definition of "Normal" is? I'm thinking particularly that a bin 4M is unlikely to be "valid" for a given axis, but still considered `Normal()`.

hist/histv7/inc/ROOT/RBinIndex.hxx
```diff
@@ -0,0 +1,144 @@
...
+      return old;
+   }
+
+   RBinIndex &operator-=(std::size_t a)
+   {
+      assert(IsNormal());
+      assert(fIndex >= a);
+      fIndex -= a;
+      assert(IsNormal());
+      return *this;
```
hageboeck: With assertions off, one can easily make it into under-/overflow or invalid values. I'm wondering whether always setting it to invalid would help in case of `fIndex < a`.

hahnjo: As discussed offline, arithmetic operations are now always defined and go to `InvalidIndex` if wrapping around.

PR #19644, opened 2025-08-14
[ntuple] fix for I/O rule handling

jblomer:
Fixes the case where the source/target class versions are identical but their names are not.

Seen in https://root-forum.cern.ch/t/manual-schema-evolution-with-i-o-rules-and-branches-containing-vector-t/64026

### DIFF
--- a/tree/ntuple/test/CustomStructLinkDef.h
+++ b/tree/ntuple/test/CustomStructLinkDef.h
@@ -154,4 +154,10 @@
 
 #pragma link C++ class RelativelyLargeStruct + ;
 
+#pragma link C++ class v1::Vector3D+;
+#pragma link C++ class v1::ExampleMC+;
+#pragma link C++ class v2::ExampleMC+;
+#pragma read sourceClass = "v1::ExampleMC" source = "v1::Vector3D fSpin" version="[1-]" targetClass = \
+   "v2::ExampleMC" target = "fHelicity" code = "{ fHelicity = onfile.fSpin.fZ; }"
+
 #endif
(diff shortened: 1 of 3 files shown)


tree/ntuple/test/CustomStructLinkDef.h
```diff
@@ -154,4 +154,10 @@
 
 #pragma link C++ class RelativelyLargeStruct + ;
 
+#pragma link C++ class v1::Vector3D+;
+#pragma link C++ class v1::ExampleMC+;
+#pragma link C++ class v2::ExampleMC+;
+#pragma read sourceClass = "v1::ExampleMC" source = "v1::Vector3D fSpin" targetClass = \
```
pcanal:
```suggestion
#pragma read sourceClass = "v1::ExampleMC" source = "v1::Vector3D fSpin" version="[1-]" targetClass = \
```
If I remember correctly rootcling complains without the 'version' tag.

pcanal (approved):
LTGM.  Thanks.

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -330,7 +330,7 @@ ROOT::DescriptorId_t ROOT::RClassField::LookupMember(const ROOT::RNTupleDescript
 void ROOT::RClassField::SetStagingClass(const std::string &className, unsigned int classVersion)
 {
    TClass::GetClass(className.c_str())->GetStreamerInfo(classVersion);
-   if (classVersion != GetTypeVersion()) {
+   if (classVersion != GetTypeVersion() || className != GetTypeName()) {
```
pcanal, line 333: As a side note, comparing the TClass pointers is a tad bit more stable than comparing the classname as `TClass::Get Class` will deal with typographical and typedefs differences.

jblomer: Noted. In this case, we only have one `TClass` pointer (the other one is being figured out by the method) and we compare normalized type names.

PR #19651, opened 2025-08-15
[tutorials] Don't import TensorFlow before ROOT

guitargeek:
TensorFlow has to be imported after ROOT to avoid LLVM symbol clashes if
ROOT was built with LLVM in Debug mode and TensorFlow>=2.20.0.


guitargeek:
PR was already green, but I needed to update to fix a typo...

dpiparo (approved):
thanks.

PR #19656, opened 2025-08-15
RooFit: Fix plotOn normalization for extended PDFs

### LINKED ISSUE #18929: [RF] Plotting extended pdf together with data unexpectedly scales pdf to number of events in data
When plotting an extended pdf (i.e. a pdf that also makes a prediction for the number of events) together with a RooFit dataset, users expect that the plotted pdf is normalized to the number of predicted events, in order to see what the actual prediction is and meaningfully compare to the data.

However, the pdf is automatically scaled to the *observed* number of events in the dataset, so there is no way to get visual feedback on whether the predicted number of events makes sense. This is useful for models that don't make any prediction on the total number of events, but misleading otherwise. See also the following forum post where this came up:
https://root-forum.cern.ch/t/roofit-background-component-not-staying-constant-despite-rooconstvar-setconstant-true/63650/2

Usually, this problem stays under the radar because post-fit event number predictions usually match the observed number of events. But in the case where the shape is correlated with the normalization, this is not the case, and the resulting small difference between post-fit and observed number of events can make you hesitant and open a forum post at best, or in the worst case the scaling to the observed number of events makes you believe the fit was good while it was actually not.

One can work around this by introducing additional scale factors (see the `Normalization` command argument for [RooAbsReal::plotOn()](https://root.cern.ch/doc/master/classRooAbsReal.html#ae26e30ce9fcf11b820a7a8e26f4aa424)), but this should not be necessary. It would be better if extended pdfs are by default normalized to the predicted number of events when plotting, not to the observed number of events.

This is a good first issue to help with RooFit development, since the required changes should be little. The challenge is mostly to figure out where the scaling to the number of events in the datasets happens, and then add a code path for the case where the RooAbsPdf makes a prediction on the expected events (i.e. the return value of [RooAbsPdf::expectedEvents()](https://root.cern.ch/doc/master/classRooAbsPdf.html#a2debaf3731b6d5f6027ee1ede917af01) is non-zero.

Reproducer:
```c++
RooRealVar x{"x", "x", 0, 1};
RooRealVar n{"n", "n", 5000, 0, 20000};

x.setBins(1);

RooUniform pdf{"pdf", "pdf", x};
std::unique_ptr<RooAbsData> data{pdf.generateBinned(x, 10000)};

RooExtendPdf extPdf{"ext_pdf", "ext_pdf", pdf, n};

auto c1 = new TCanvas{"c1", "c1"};

auto frame = x.frame();
data->plotOn(frame);
extPdf.plotOn(frame);

// Should be the value of n
std::cout << frame->getCurve()->GetY()[1] << std::endl;

frame->Draw();

c1->SaveAs("plot.png");
```
The output is:
```
10000
```
But we expect the curve to represent the actual number of expected events, which is 5000 in this reproducer.
Image

JasMehta08:
# This Pull Request:

## Changes or fixes:

This PR fixes an issue where the `plotOn` function incorrectly normalizes an extended PDF to the number of events in a dataset, rather than to its own predicted number of events. This behavior is misleading when the PDF's normalization is a key part of the model's prediction.

The default logic in `RooAbsPdf::plotOn` is modified to check if a PDF is extended by calling `expectedEvents()`. If it is, the function now uses the PDF's prediction for normalization; otherwise, it retains the old behavior of scaling to the data for non-extended PDFs.

## Checklist:

- [x] tested changes locally

## Testing 
I have verified this fix locally with several tests:

1.  **Reproducer Script:** Confirmed that for an extended PDF, the plotted curve is now normalized to the PDF's `expectedEvents()`.

    ```cpp
    #include "RooRealVar.h"
    #include "RooUniform.h"
    #include "RooDataSet.h"
    #include "RooExtendPdf.h"
    #include "RooPlot.h"
    #include "TCanvas.h"
    #include "RooCurve.h"
    #include <iostream>
    #include <memory>

    void reproducer() {
        RooRealVar x{"x", "x", 0, 1};
        RooRealVar n{"n", "n", 5000, 0, 20000};
        x.setBins(1);
        RooUniform pdf{"pdf", "pdf", x};
        std::unique_ptr<RooAbsData> data{pdf.generateBinned(x, 10000)};
        RooExtendPdf extPdf{"ext_pdf", "ext_pdf", pdf, n};
        auto c1 = new TCanvas{"c1", "c1"};
        auto frame = x.frame();
        data->plotOn(frame);
        extPdf.plotOn(frame);
        std::cout << frame->getCurve()->GetY()[0] << std::endl;
        frame->Draw();
        c1->SaveAs("plot.png");
    }
    ```
image

2.  **Regression Test:** Confirmed that for a standard, non-extended `RooGaussian`, the plotted curve is still correctly scaled to the data.

```cpp
#include "RooRealVar.h"
#include "RooGaussian.h"
#include "RooDataSet.h"
#include "RooPlot.h"
#include "TCanvas.h"
#include "RooConstVar.h" 
#include <memory>

void regression_test() {
    // This script checks that non-extended PDFs still
    // correctly scale to the data.

    RooRealVar x("x", "x", -10, 90);
    x.setBins(10);

    // Create the mean and sigma as named RooConstVar objects first
    RooConstVar mean("mean", "mean", 6.0);
    RooConstVar sigma("sigma", "sigma", 3.0);

    // Now pass these named objects to the constructor
    RooGaussian pdf("pdf", "pdf", x, mean, sigma);

    // A dataset with 500 events.
    std::unique_ptr<RooDataSet> data{pdf.generate(x, 500)};

    // Plot both.
    TCanvas* c1 = new TCanvas("c1", "c1");
    RooPlot* frame = x.frame();
    data->plotOn(frame);
    pdf.plotOn(frame); // The curve should scale to the data.

    frame->Draw();
    c1->SaveAs("regression_test.png");
}
```

image

3.  **Comprehensive Test:** A single script was run to verify the default behavior and the `RelativeExpected` and `NumEvent` normalization overrides. The results are shown in the plot below, and all components behave as expected.
```cpp
#include "RooRealVar.h"
#include "RooGaussian.h"
#include "RooExtendPdf.h"
#include "RooDataSet.h"
#include "RooPlot.h"
#include "TCanvas.h"
#include "RooConstVar.h"
#include <memory>

void reproducer() {
    TCanvas* c1 = new TCanvas("c1", "Comprehensive plotOn Test", 1200, 400);
    c1->Divide(3, 1); // Create a 3x1 grid of plots

    // --- Common variables for all tests ---
    RooRealVar x("x", "x", 0, 10);
    x.setBins(20);

    // --- Test 1: Default behavior for NON-EXTENDED PDF ---
    c1->cd(1); // Move to the first pad
    RooPlot* frame1 = x.frame(RooFit::Title("Test 1: Default (Non-Extended)"));
    
    RooConstVar mean1("mean1", "mean1", 5.0);
    RooConstVar sigma1("sigma1", "sigma1", 1.0);
    RooGaussian pdf1("pdf1", "pdf1", x, mean1, sigma1);
    
    std::unique_ptr<RooDataSet> data1{pdf1.generate(x, 1000)};
    
    data1->plotOn(frame1);
    pdf1.plotOn(frame1); // Should scale to 1000 events in data
    frame1->Draw();

    // --- Test 2: RelativeExpected behavior for EXTENDED PDF ---
    c1->cd(2); // Move to the second pad
    RooPlot* frame2 = x.frame(RooFit::Title("Test 2: RelativeExpected"));

    RooRealVar n2("n2", "n2", 2000); // PDF predicts 2000 events
    RooExtendPdf pdf2("pdf2", "pdf2", pdf1, n2);

    // Create dummy data with a DIFFERENT number of events
    std::unique_ptr<RooDataSet> data2{pdf1.generate(x, 5000)};

    data2->plotOn(frame2);
    // Explicitly ask to scale to the PDF's own prediction
    pdf2.plotOn(frame2, RooFit::Normalization(1.0, RooAbsPdf::RelativeExpected));
    frame2->Draw();

    // --- Test 3: NumEvent behavior ---
    c1->cd(3); // Move to the third pad
    RooPlot* frame3 = x.frame(RooFit::Title("Test 3: NumEvent"));

    // Use the same non-extended pdf1 and data1
    data1->plotOn(frame3);
    // Explicitly scale the PDF to a specific number of events (e.g., 2500)
    pdf1.plotOn(frame3, RooFit::Normalization(2500, RooAbsPdf::NumEvent));
    frame3->Draw();

    c1->SaveAs("full_test.png");
}
```
image

Fixes #18929

### DIFF
--- a/roofit/roofitcore/src/RooAbsPdf.cxx
+++ b/roofit/roofitcore/src/RooAbsPdf.cxx
@@ -2146,7 +2146,21 @@ RooPlot* RooAbsPdf::plotOn(RooPlot* frame, RooLinkedList& cmdList) const
         scaleFactor *= rangeNevt/nExpected ;
 
       } else {
-        scaleFactor *= frame->getFitRangeNEvt()/nExpected ;
+        // First, check if the PDF *can* be extended.
+        if (this->canBeExtended()) {
+            // If it can, get the expected events.
+            const double nExp = expectedEvents(frame->getNormVars());
+            if (nExp > 0) {
+                // If the prediction is valid, use it for normalization.
+                scaleFactor *= nExp / nExpected;
+            } else {
+                // If prediction is not valid (e.g. 0), fall back to data.
+                scaleFactor *= frame->getFitRangeNEvt() / nExpected;
+            }
+        } else {
+            // If the PDF can't be extended, just use the data.
+            scaleFactor *= frame->getFitRangeNEvt() / nExpected;
+        }
       }
     } else if (stype==RelativeExpected) {
       scaleFactor *= nExpected ;


JasMehta08:
I have figured out whats wrong with my pull request 

Not all pdfs are required to have extended pdfs so it would cause trouble there and also that some of the tests were having the wrong output so i will have to fix that too. I will do this in sometime.

JasMehta08:
i changed the code a to make sure to ask if the pdfs are extendable and then i ran most of the tests that i could that were failing in the tutorial tests and they passed so i think this should fix it.

(also this is my first time contributing could help me understand how can i test all the CI tests locally if even possible to do so)

JasMehta08:
@guitargeek i went through the failing tests log and have figured out that the problem is being solved which is causing the stress tests to fail when it is comparing to the expected answers. Can you take a look at my code and tell me if you feel that it is right according to you, so, that i can go ahead and change the tests which are present in the stress tests.

JasMehta08:
Hi @guitargeek ,

I've pushed a new version of the code that I believe is a more robust fix for the CI failures. Could a maintainer please approve the workflows to run so I can verify the solution?

Thank you!

JasMehta08:
the CI tests are passing, the ones which did not pass are having error relating to, 

Failed to open the S3 connection: You must provide an auth secret.

which i think was a error with the CI itself not with the bug fix could you please look into it @guitargeek 

Thank you!

PR #19657, opened 2025-08-15
[v634][tutorials] Don't import TensorFlow before ROOT

guitargeek:
Backport.


guitargeek:
Forgot that this branch is closed

PR #19658, opened 2025-08-15
[interpreter] Decouple Cling assertions from `LLVM_BUILD_TYPE`

hahnjo:
Reset `CMAKE_BUILD_TYPE` after LLVM to make Cling (and CppInterOp) follow the general ROOT build type and the `asserts` option. This allows having Cling assertions even with `RelWithDebInfo` builds, the default `LLVM_BUILD_TYPE=Release`, and `asserts=ON` (as we have in the CI).

### DIFF
--- a/interpreter/CMakeLists.txt
+++ b/interpreter/CMakeLists.txt
@@ -190,6 +190,8 @@ if(builtin_llvm)
   # in production builds.
   set(LLVM_UNREACHABLE_OPTIMIZE FALSE)
 
+  set(_build_type ${CMAKE_BUILD_TYPE})
+
   # Multi-configuration generators ignore CMAKE_BUILD_TYPE, so
   # in that case we set the flags for all configurations to the
   # flags of the build type assigned to LLVM_BUILD_TYPE.
@@ -241,6 +243,7 @@ if(builtin_llvm)
   endif()
 
   set(CMAKE_CXX_STANDARD ${_cxx_standard})
+  set(CMAKE_BUILD_TYPE ${_build_type})
 
   set(LLVM_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/llvm-project/llvm/include
     ${CMAKE_CURRENT_BINARY_DIR}/llvm-project/llvm/include


pcanal:
With debug `LLVM`, `rootcling` is significantly slower (and thus regular ROOT build are slower).   So a related question is whether with just `Cling` in debug mode, the build are significantly or just (hopefully) slightly slower.

dpiparo:
How slower are builds @pcanal @hahnjo ?

hahnjo:
According to my tests (with `-DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -Dasserts=ON -DLLVM_ENABLE_ASSERTIONS=ON`, meaning `LLVM_BUILD_TYPE=Release`), I don't see an influence on the overall build time - simply because it's dominated by compilation time, especially of LLVM itself. An incremental `ninja` after deleting the modules (`rm lib/*.pcm`) takes a minor hit, from 4m51s to 5m23s. Note that `TCling` and friends were already in `Debug` mode before.

In my opinion, this is acceptable - after all the developer asked for `Debug`. I plan to merge later today unless I hear objection until then.

hahnjo:
> According to my tests (with `-DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -Dasserts=ON -DLLVM_ENABLE_ASSERTIONS=ON`, meaning `LLVM_BUILD_TYPE=Release`), I don't see an influence on the overall build time - simply because it's dominated by compilation time, especially of LLVM itself. An incremental `ninja` after deleting the modules (`rm lib/*.pcm`) takes a minor hit, from 4m51s to 5m23s. Note that `TCling` and friends were already in `Debug` mode before.

For comparison, a build with `LLVM_BUILD_TYPE=Debug` takes 15m39s for building the dictionaries + modules. That's the slowness @pcanal mentioned.

vgvassilev:
> According to my tests (with `-DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -Dasserts=ON -DLLVM_ENABLE_ASSERTIONS=ON`, meaning `LLVM_BUILD_TYPE=Release`), I don't see an influence on the overall build time - simply because it's dominated by compilation time, especially of LLVM itself. An incremental `ninja` after deleting the modules (`rm lib/*.pcm`) takes a minor hit, from 4m51s to 5m23s. Note that `TCling` and friends were already in `Debug` mode before.
> 
> In my opinion, this is acceptable - after all the developer asked for `Debug`. I plan to merge later today unless I hear objection until then.

Maybe one quick test to do is the pch build times for `-Druntime_cxxmodules=Off`. That normally takes more than 6-7 minutes with rootcling built with llvm in debug.

hahnjo:
> Maybe one quick test to do is the pch build times for `-Druntime_cxxmodules=Off`. That normally takes more than 6-7 minutes with rootcling built with llvm in debug.

I checked with `runtime_cxxmodules=OFF` and the time for generating `etc/allDict.cxx.pch` indeed takes a large hit, from 1m14s to 3m48s (factor 3x). I did some quick profiling and this is all because of this linear duplicates check here:
https://github.com/root-project/root/blob/1a06266d75fca88b6045c8cfafd7e4f459337998/interpreter/cling/lib/Interpreter/Transaction.cpp#L150-L177
Temporarily commenting out these lines brings down the time to 1m25s. Since this is not the default configuration, I'm going to merge. If somebody is bothered by this slowdown (for `Debug` builds), the check can probably be improved...

vgvassilev:
> > Maybe one quick test to do is the pch build times for `-Druntime_cxxmodules=Off`. That normally takes more than 6-7 minutes with rootcling built with llvm in debug.
> 
> I checked with `runtime_cxxmodules=OFF` and the time for generating `etc/allDict.cxx.pch` indeed takes a large hit, from 1m14s to 3m48s (factor 3x). I did some quick profiling and this is all because of this linear duplicates check here:
> 
> https://github.com/root-project/root/blob/1a06266d75fca88b6045c8cfafd7e4f459337998/interpreter/cling/lib/Interpreter/Transaction.cpp#L150-L177
> 
> Temporarily commenting out these lines brings down the time to 1m25s. Since this is not the default configuration, I'm going to merge. If somebody is bothered by this slowdown (for `Debug` builds), the check can probably be improved...

Maybe we can wrap it in if llvm is in debug -- usually if you care about these you are an llvm dev that will have llvm in debug mode.

PR #19659, opened 2025-08-15
[RF] Adding RooMultiReal to per request from ATLAS

GalinBistrev2:
This change wants to implement a new class which work like RooMultiPdf , but instead of pdfs it deals with RooRealVars.
testRooMultiPdf.cxx was changed to testRooMultiReal.cxx  in order everything to be more compact. All the tests are passing


guitargeek (approved):
Thank you very much Galin!

PR #19660, opened 2025-08-15
[CPyCppyy] Add smart pointer conversion policy flag and remove local memory heuristics flag

guitargeek:
This is to reduce the patches between Wim's cppyy and ROOT's (if @wlav is fine with it :slightly_smiling_face: ).


guitargeek:
The upstream review has not happened yet, but I think it's important to merge this early to get some exposure of this change to the users of ROOT `master` before the 6.38 release

PR #19663, opened 2025-08-16
[core] Generalize TROOT::GetLibDir()

guitargeek:
The problem with `TROOT::GetLibDir()` was that it either hardcodes the install directory inside it if `ROOTLIBDIR` was set (which is the case for `gnuinstall=ON`), or it assumes that it's always called `lib` and is inside the `$ROOTSYS`, which might or might not be set. This is broken if the `CMAKE_INSTALL_LIBDIR` is different from the default `lib`.

This commit suggests to automatically figure out the ROOT library directory instead.

https://its.cern.ch/jira/browse/ROOT-9569

### DIFF
--- a/core/base/src/TROOT.cxx
+++ b/core/base/src/TROOT.cxx
@@ -71,6 +71,7 @@ of a main program creating an interactive version is shown below:
 #include <ROOT/RVersion.hxx>
 #include "RConfigure.h"
 #include "RConfigOptions.h"
+#include <filesystem>
 #include <string>
 #include <map>
 #include <set>
@@ -109,8 +110,12 @@ FARPROC dlsym(void *library, const char *function_name)
       return ::GetProcAddress((HMODULE)library, function_name);
    }
 }
+#elif defined(__APPLE__)
+#include <dlfcn.h>
+#include <mach-o/dyld.h>
 #else
 #include <dlfcn.h>
+#include <link.h>
 #endif
 
 #include <iostream>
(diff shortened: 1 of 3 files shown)


dpiparo (approved):

core/base/src/TROOT.cxx
```diff
@@ -3007,23 +3012,71 @@ const TString& TROOT::GetBinDir() {
...
-   } else {
-      const static TString rootlibdir = ROOTLIBDIR;
+const TString &TROOT::GetLibDir()
+{
+   static bool haveLooked = false;
+   static TString rootlibdir;
+   if (haveLooked)
       return rootlibdir;
+
+   haveLooked = true;
```
hahnjo, line 3033: This caching logic is not thread-safe, it needs at least atomic's or (more probably) locking

pcanal: @guitargeek A good solution is:
```
static TString rootlibdir  = somefunction_or_even_a_lambda();
```
which will be guaranteed to run only once and be made thread safe by the compiler (requirement from the C++ standard). (It is clear that the existing was not thread safe but let's take this occasion to improve it).

guitargeek: I didn't worry about thread safety, because this function is called the first time in the `TROOT` constructor. I assumed the `TROOT` constructor is not called concurrently anyway. Or can you start multiple ROOT from several threads concurrently? Why does thread safety matter here?

pcanal: Fair enough. I had not noticed/checked where this function was currently called from ... It is probably just fine as is.

As a s side note, when updating code in core, it is always good to worry about thread-safely even in case like this as future refactor (however unlikely they are) may move the function first call somewhere else ... and, in case like this one, switching to a thread-safe pattern is not onerous.

guitargeek: Ok! In a future refactor, I would then make the result of the library dir a data member of `TROOT` that is set in the constructor, instead of a static variable :)

PR #19665, opened 2025-08-17
[CMake] Don't cache `CPPINTEROP_INCLUDE_DIRS` and `CLING_INCLUDE_DIRS`

guitargeek:
Cached variables end up in `root-config --config`, and we don't want to have absolute paths to the bulid directory inside its output.

After this change, no trace of the full ROOT source or build directory path is left in the output of `root-config --config`.

Closes https://its.cern.ch/jira/browse/ROOT-9481

### DIFF
--- a/core/metacling/src/CMakeLists.txt
+++ b/core/metacling/src/CMakeLists.txt
@@ -50,7 +50,7 @@ target_include_directories(MetaCling SYSTEM PRIVATE
   ${CLANG_INCLUDE_DIRS}
   ${LLVM_INCLUDE_DIRS}
   ${CLAD_INCLUDE_DIRS}
-  ${CPPINTEROP_INCLUDE_DIRS}
+  ${CMAKE_SOURCE_DIR}/interpreter/CppInterOp/include
 )
 
 target_include_directories(MetaCling PRIVATE
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -376,6 +376,9 @@ unset(artifact_files)
 
 add_custom_target(move_artifacts DEPENDS ${stamp_file} ${all_artifacts})
 
+if(builtin_cling)
+    set(CLING_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/interpreter/cling/include)
+endif()
 
 #---Add the main sources ---------------------------------
 add_subdirectory (interpreter)
--- a/interpreter/CMakeLists.txt
+++ b/interpreter/CMakeLists.txt
@@ -467,8 +467,6 @@ if (builtin_cling)
   add_subdirectory(cling EXCLUDE_FROM_ALL)
   add_dependencies(CLING ${CLING_LIBRARIES})
 
-  set(CLING_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/cling/include CACHE STRING "")
-
 
 
   #---These are the libraries that we link ROOT with CLING---------------------------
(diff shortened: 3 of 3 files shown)


pcanal:
> After this change, no trace of the full ROOT source or build directory path is left in the output of root-config --config.

Is that true even for 'maximal' use of builtins? (mine is probably 'old' but seems to list LZ4, ZSTD, Glew, xxHash, FTGL)

core/metacling/src/CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ target_include_directories(MetaCling SYSTEM PRIVATE
   ${CLANG_INCLUDE_DIRS}
   ${LLVM_INCLUDE_DIRS}
```
pcanal, line 51: Cache variable stored in `root-config` are filtered around `CMakeLists.txt:611+` and it turns out that `LLVM_INCLUDE_DIRS` (which is cached) is not selected due to an error/mistake.  We have there:
```
     (NOT ${${var}} STREQUAL "") AND
```
which is 'false' for `LLVM_INCLUDE_DIRS` (at least in my case) only because it has 2 components (and this `;` which seems to mess things up).  The correct syntax:
```
     (NOT ${var} STREQUAL "") AND
```
leads to `LLVM_INCLUDE_DIRS` being added to `root-config` .. but somewhere in a broken way (the `;` is there unescaped/unquoted)

So we should fix the above mentioned line **and** we should extend the list of excluded variable (probably all the builtins in addition to LLVM).

guitargeek: I don't see what's the advantage of filtering out these cached variables over just avoiding to cache them, as suggested in this PR. Why would you do this? It makes the code more complicated, because the veto list needs to be kept in sync with the cached variables...

guitargeek:
> > After this change, no trace of the full ROOT source or build directory path is left in the output of root-config --config.
> 
> Is that true even for 'maximal' use of builtins? (mine is probably 'old' but seems to list LZ4, ZSTD, Glew, xxHash, FTGL)

Probably not, but I think it doesn't matter. The people who care about their `root-config --config` being clean are probably the same people who disable builtins where they can :slightly_smiling_face:

Quoting from the original JIRA ticket:

> Although these references seem harmless for the time being (i.e. not being actually used anywhere), we should remove them to ensure they won't become a problem in the future.

So since we don't really know why these references are a problem yet, I don't think it's worth to do a more thorough investigation, and I'd advocate for addressing the JIRA ticket for the default build, as suggested in this PR.

core/metacling/src/CMakeLists.txt
pcanal: My slight point was that this current PR is slightly inconsistent as it does not 'also' uncache (or filter out) the other `INCLUDE_DIRS`  in particular the very similar `LLVM_INCLUDE_DIRS` and `CLANG_INCLUDE_DIRS`.  i.e. I don't see why we would care about `CLING_INCLUDE_DIRS` not being in the `root-config` but not care about the other 2 ...  and if I correct that all 3 are similar, we are currently a 'bug fix' (the one mentioned in https://github.com/root-project/root/pull/19665#discussion_r2283644342) away from inadvertently reintroducing a slight various of the issue being addressed here.

ferdymercury: Tangentially related: https://its.cern.ch/jira/browse/ROOT-8061
And https://root-forum.cern.ch/t/creating-an-application-root-config-treeviewer-missing/63356/3

ferdymercury:
> Closes https://its.cern.ch/jira/browse/ROOT-9481

Gentle reminder to close the JIRA Issue.

Issue #15792, opened 2024-06-07
root_generate_dictionary should check target OUTPUT_NAME


lmoureaux:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

When a dictionary is generated for a library **target** called `foo`, whose `OUTPUT_NAME` is set to **bar**, `root_generate_dictionary` produces dictionary files called `libfoo_rdict.pcm` and `libfoo.rootmap`.

### Reproducer

`ROOTBug.cc` and `LinkDef.h` are both empty files. Contents of `CMakeLists.txt`:
```cmake
cmake_minimum_required(VERSION 3.16 FATAL_ERROR)
project(ROOTBug)
find_package(ROOT 6.28 REQUIRED)

# https://github.com/root-project/root/issues/8308#issuecomment-1143791946
set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")

# another bug... INSTALL FILES given no destination!
include(GNUInstallDirs)

add_library(ROOTBug SHARED ROOTBug.cc)
target_link_libraries(ROOTBug PRIVATE ROOT::Core)
set_target_properties(ROOTBug PROPERTIES OUTPUT_NAME ROOTWorks)  # Important!
root_generate_dictionary(ROOTWorksDict MODULE ROOTBug LINKDEF LinkDef.h)

install(TARGETS ROOTBug)
```
Build commands:
```
$ cmake -B build -DCMAKE_INSTALL_PREFIX=install
$ cmake --build build --target install
[ 25%] Generating ROOTBugDict.cxx, libROOTBug_rdict.pcm, libROOTBug.rootmap
[ 50%] Building CXX object CMakeFiles/ROOTBugDict.dir/ROOTBugDict.cxx.o
[ 50%] Built target ROOTBugDict
[ 75%] Linking CXX shared library libROOTWorks.so
[100%] Built target ROOTBug
Install the project...
-- Install configuration: ""
-- Installing: /nfs/dust/cms/user/mourelou/darwin/tests/install/lib64/libROOTBug_rdict.pcm
-- Installing: /nfs/dust/cms/user/mourelou/darwin/tests/install/lib64/libROOTBug.rootmap
-- Installing: /nfs/dust/cms/user/mourelou/darwin/tests/install/lib64/libROOTWorks.so
-- Set runtime path of "/nfs/dust/cms/user/mourelou/darwin/tests/install/lib64/libROOTWorks.so" to ""
[...]

PR #19666, opened 2025-08-17
[CMake] Use output name of `MODULE` target also for dictionary name

guitargeek:
Use output name of `MODULE` target also for the dictionary name, if the target is defined. This makes the dictionary and library file names more consistent.

Closes https://github.com/root-project/root/issues/15792.


cmake/modules/RootMacros.cmake
```diff
@@ -958,7 +954,7 @@ function(ROOT_LINKER_LIBRARY library)
     endif()
   endif()
 
-  set_target_properties(${library} PROPERTIES OUTPUT_NAME ${library_name})
+  set_target_properties(mylib PROPERTIES PREFIX ${prefix})
```
pcanal: is `mylib` intended here?

guitargeek: No, sorry

guitargeek:
Hi @pcanal, thanks for taking a look! There were too many moving parts. Now I intend to first get some sane suffix and prefix settings also on Windows and continue from there.

cmake/modules/RootMacros.cmake
```diff
@@ -958,7 +954,7 @@ function(ROOT_LINKER_LIBRARY library)
     endif()
   endif()
 
-  set_target_properties(${library} PROPERTIES OUTPUT_NAME ${library_name})
+  set_target_properties(${library} PROPERTIES PREFIX ${libprefix})
```
pcanal: Does it have an effect on the linux library location/names?

guitargeek: No, because for platforms other than Windows, we already do `set(libprefix ${CMAKE_SHARED_LIBRARY_PREFIX})` in the beginning of the file. So `${libprefix}` is already the default, and re-setting this at the target level is a no-op

pcanal: Fair enough.

pcanal: On the other hand this does not seem enough to make it consistent.  The log says:
```
Creating library C:/ROOT-CI/build/lib/Core.lib and object C:/ROOT-CI/build/lib/Core.exp
```
and then
```
LINK : fatal error LNK1181: cannot open input file 'libCore.lib' [C:\ROOT-CI\build\roottest\root\tree\readrules\gh-14462\gh14462regression.vcxproj]
```

guitargeek: Yes, there is a real problem :( I'm still thinking about it and also asked @bellenot for help

guitargeek:
I think I got it now:

  * https://cmake.org/cmake/help/latest/prop_tgt/IMPORT_PREFIX.html

bellenot:
> I think I got it now:
> 
> * https://cmake.org/cmake/help/latest/prop_tgt/IMPORT_PREFIX.html

Good catch!

bellenot (approved):
LGTM, thanks Jonas!

PR #19673, opened 2025-08-18
[roottest] use cmake for root/meta tests, part 2

linev:


pcanal (approved):
LVGTM.  Thank you very much for your patience.

PR #19672, opened 2025-08-18
[roottest] use cmake for root/meta tests, part 1

linev:

PR #19676, opened 2025-08-18
[roottest] convert to cmake meta/ROOT-7181 tests

linev:
Like in cmsUnload, provide special runscript.C which loads commands and invoke them one by one.

Tests failing on most platforms with assert message from clang, to be investigated

### DIFF
--- a/roottest/root/meta/ROOT-7181/inc/jet_selection.xml
+++ b/roottest/root/meta/ROOT-7181/inc/jet_selection.xml
@@ -27,6 +27,11 @@
    <class name="ElementLinkVector<xAOD::JetContainer_v1>" />
    <class name="std::vector<ElementLinkVector<xAOD::JetContainer_v1> >" />
 
+   <class name="DataVector<xAOD::Jet_v1>" />
+   <class name="ElementLink<DataVector<xAOD::Jet_v1>>" />
+   <class name="std::vector<ElementLink<DataVector<xAOD::Jet_v1>>>" />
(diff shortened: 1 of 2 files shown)


roottest/root/meta/ROOT-7181/scripts/case4.script
```diff
@@ -1,5 +1,5 @@
-.L lib/libjet.so
-.L lib/libbtag.so 
+gSystem->Load("libjet_dictrflx");
```
pcanal: I am curious, why those changes?

pcanal:
> Like in cmsUnload, provide special runscript.C which loads commands and invoke them one by one.

If that PR is already open and/or merged, can we link it here?

roottest/root/meta/ROOT-7181/scripts/case4.script
linev: First of all, `.so` is not working on Windows.
And second - use of `.L libjet_dictrflx` requires changes of default macro path.
So I trying to load library with `gSystem->Load`. 
Test perfectly works on my node, but fails on all CI nodes.
So I need to investigate further

linev:
> > Like in cmsUnload, provide special runscript.C which loads commands and invoke them one by one.
> 
> If that PR is already open and/or merged, can we link it here?

Change of cmsUnload is here: https://github.com/root-project/root/pull/19672
But they do not depend from each other. 
In both tests instead of redirecting of standard input one just read text file line by line and executing `ProcessLine`

linev:
Closing in favor of https://github.com/root-project/root/pull/20424

PR #19678, opened 2025-08-18
Update RooSimultaneous.cxx - protect evaluate against missing pdf for current categorical state

will-cern:
A user encountered a nasty crash when evaluating a pdf that didn't have the current state pdf defined for it. Should just return 0, not crash out!


guitargeek:
Why zero and not NaN?

will-cern:
Do we have a policy of returning nan for evaluations of PDFs outside of their domain? If so, happy to make it a nan. Is that the behaviour of your new RooMultiPdfs too?

PR #19681, opened 2025-08-19
[df] Some cleanups to `RNTupleDS`

hahnjo:


tree/dataframe/src/RNTupleDS.cxx
```diff
@@ -687,6 +687,7 @@ std::vector<std::pair<ULong64_t, ULong64_t>> ROOT::RDF::RNTupleDS::GetEntryRange
    // We remember the connection from first absolute entry index of a range to its REntryRangeDS record
    // so that we can properly rewire the column reader in InitSlot
    fFirstEntry2RangeIdx.clear();
+   fOriginalRanges.clear();
```
hahnjo, line 689: @martamaja10 I think I found the case where this makes a difference: there have to be more files than slots so the ranges are split over multiple invocations of `GetEntryRange`. I can trigger this manually by lowering the number of slots passed to `ROOT::EnableImplicitMT`:
```diff
diff --git a/tree/dataframe/test/dataframe_datasetspec.cxx b/tree/dataframe/test/dataframe_datasetspec.cxx
index 18aeb3bda7a..50215cce9cb 100644
--- a/tree/dataframe/test/dataframe_datasetspec.cxx
+++ b/tree/dataframe/test/dataframe_datasetspec.cxx
@@ -88,7 +88,7 @@ std::vector<RTestSample> data{
 // The helper class is also responsible for the creation and the deletion of the root specTestFiles
 class RDatasetSpecTest : public ::testing::TestWithParam<bool> {
 protected:
-   RDatasetSpecTest() : NSLOTS(GetParam() ? std::min(4u, std::thread::hardware_concurrency()) : 1u)
+   RDatasetSpecTest() : NSLOTS(GetParam() ? std::min(2u, std::thread::hardware_concurrency()) : 1u)
    {
       if (GetParam())
          ROOT::EnableImplicitMT(NSLOTS);
```
with a failure of `MT/RDatasetSpecTest.RNTupleWithGlobalRanges/0`:
```
520: [ RUN      ] MT/RDatasetSpecTest.RNTupleWithGlobalRanges/0                                                                                                                                         
520: RDataFrame::Run: event loop was interrupted                                                                                                                                                        
520: unknown file: Failure                                                                                                                                                                              
520: C++ exception with description "entry with index 14 out of bounds                                                                                                                                  
520: At:                                                                                                                                                                                                
520:   virtual ROOT::Internal::RPageRef ROOT::Internal::RPageSource::LoadPage(ColumnHandle_t, ROOT::NTupleSize_t) [/home/jhahnfel/ROOT/src/tree/ntuple/src/RPageStorage.cxx:401]                        
520: " thrown in the test body.                                                                                                                                                                         
520: [  FAILED  ] MT/RDatasetSpecTest.RNTupleWithGlobalRanges/0, where GetParam() = true (298 ms)
```
I didn't add this as a test yet, maybe you can repeat your checks with fewer slots than files / more files than slots?

martamaja10: okay, I'll do that, thanks!

martamaja10: it's correct, the issue was there and after this change it is resolved

martamaja10 (approved):
Thank you Jonas for the fix with clearing the fOriginalRanges and for the simplifications! I also tested these changes locally with changes introduced in the PR for the cluster-based scheduling of RNTuples in distributed mode (https://github.com/root-project/root/pull/19546/files) and things work there as well.

PR #19685, opened 2025-08-19
[RF] Adding Discrete profiling to RooMinimizer

GalinBistrev2:
This pull request adds Discrete profiling to RooMinimizer


roofit/roofitcore/src/RooMultiPdf.cxx
```diff
@@ -86,4 +86,5 @@ void RooMultiPdf::getParametersHook(const RooArgSet *nset, RooArgSet *list, bool
 
    list->removeAll();
    getCurrentPdf()->getParameters(nset, *list, stripDisconnected);
+   list->add(*x);
```
guitargeek, line 89: Thanks for the PR and addressing the comments I have done offline! Could you also make the same change to the `RooMultiReal` to keep things consistent? Thanks!

PR #19693, opened 2025-08-20
[ntuple] Use a view-based interface in `RNTupleProcessor`

enirolf:
Demonstrator for an `RNTupleProcessor` interface that implements `GetView<T>`, similar to the `RNTupleReader`. Rather than an `RNTupleView<T>`, it returns a `RNTupleProcessorView<T>` for compatibility reasons, but the public interface is identical.

This interface change requires a significant overhaul of the internal `RNTupleProcessor` machinery, most notably removing the use of `RNTupleModel` altogether, the introduction of the `RNTupleProcessorEntry` and the internal state management of the processors. Each processor now implements the following methods:

#### `Connect(std::shared_ptr<Internal::RNTupleProcessorEntry> entry = nullptr)`
Called when the processor is first needed. `Connect` sets the entry, which is either created and filled, or passed from a parent processor.
* In the case of the single processor, connect the underlying RNTuple's page source.

#### `Update(ROOT::NTupleSize_t globalIndex)`
Update the internal state of the processor to be able to read values at `globalIndex`. 
* In the case of a chain, if necessary connect a new processor in the chain, and update the internal state of the current nested processor. 
* In the case of a join, update the internal states of the primary and auxiliary processors. If the join is unaliged, use the join table to find the auxiliary entry number corresponding to `globalIndex` and use this to update the auxiliary processor.

#### `HasNextEntry()`
Used by the iterator. Check if the processor has an entry following the current one (`fCurrentEntryNumber`).

#### `HasField(std::string_view)`
Recursively check whether a field can exists in the processor's schema, down to the single processor(s).

#### `CreateField(std::string_view fieldName, std::string_view typeName = "")`
Create a new field for processing. Recurses down to the single processor(s), where the field is connected to the page source.

#### ```GetLocalEntryIndex(std::string_view fieldName)```
Get the entry index corresponding to the on-disk entry index of the single processor the field belongs to. Used by `RNTupleProcessorView` to read the values.

## To discuss
The view-based interface now also makes random entry access through the processor possible, which initially we intended not to allow/provide. I don't think this should be a big blocker, as long as we clearly document that the recommended way to interact with the processor through its iterator, but I'm interested in opinions on this point.

### DIFF
--- a/tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
@@ -120,7 +120,6 @@ protected:
    ROOT::NTupleSize_t fNEntries = kInvalidNTupleIndex;
 
    ROOT::NTupleSize_t fCurrentEntryIndex = kInvalidNTupleIndex; //< Current processor entry index
-   std::size_t fCurrentProcessorNumber = 0;                     //< Number of the currently open inner processor
 
    static ENTupleProcessorKind GetKind() { return fKind; }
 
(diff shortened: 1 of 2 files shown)


enirolf:
Closed in favor of #19111.

PR #19694, opened 2025-08-20
[hist] Implement `RBinIndexRange`

hahnjo:
The interface allows convenient iteration over `RBinIndex`.

Builds on https://github.com/root-project/root/pull/19643, first two commits with the implementation of `RBinIndex` should be reviewed there.

---

A bit of context on constructing `RBinIndexRange`: Initially I considered a plain constructor and / or static factory methods that would allow creating bin index ranges without an axis object. That was kind of awkward because of conventions (`end` is exclusive, so to have a range including `RBinIndex::Overflow()` we use the invalid `RBinIndex`), the difference in axis configurations (are flow bins enabled?), and in axis types (a categorical axis will only have an overflow, but not an underflow bin). When discussing with @hageboeck, we came to the conclusion that (for the moment) we prefer creating `RBinIndexRange`s from axis objects. This can also enforce a number of constraints, for example that the normal bins are actually valid for the given axis configuration. To simplify testing, I opted for a `Internal::CreateBinIndexRange` that is also used by `Get{Normal,Full}Range`.


hist/histv7/test/hist_variable.cxx
```diff
@@ -194,3 +196,81 @@ TEST(RVariableBinAxis, GetLinearizedIndex)
...
+      EXPECT_EQ(normal.GetEnd(), index5);
+      EXPECT_EQ(std::distance(normal.begin(), normal.end()), 4);
+   }
+
+   {
+      const auto index1 = RBinIndex(1);
+      const auto empty = axis.GetNormalRange(index1, index1);
+      EXPECT_EQ(empty.GetBegin(), index1);
+      EXPECT_EQ(empty.GetEnd(), index1);
+      EXPECT_EQ(empty.begin(), empty.end());
```
hageboeck, line 241: By construction, `std::distance(begin, end)` should be 0, but it might help in case of future changes.

hist/histv7/inc/ROOT/RBinIndexRange.hxx
```diff
@@ -0,0 +1,154 @@
...
+         return *this;
+      }
+      Iterator operator++(int)
+      {
+         Iterator old = *this;
+         operator++();
+         return old;
+      }
+
+      RBinIndex operator*() const { return fIndex; }
```
hageboeck, line 108: I assume it's not a reference expecting it to be used mostly in C++20 and later?

hageboeck (approved):
Almost nothing to add. :+1:

hist/histv7/inc/ROOT/RBinIndexRange.hxx
hahnjo: After quite some research, I changed the `iterator_category` to `std::input_iterator_tag` because it turns out that forward iterators have a requirement that we cannot fulfill:
https://eel.is/c++draft/forward.iterators#6 24.3.5.5.6
> If a and b are both dereferenceable, then a == b if and only if *a and *b are bound to the same object.

We cannot make this happen for member objects of the iterator.

PR #19696, opened 2025-08-20
[ntuple] Document `RNTupleProcessor::PrintStructure`

enirolf:


tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -142,6 +142,10 @@ protected:
    /// with respect to the processor's position in the chain.
    virtual void AddEntriesToJoinTable(Internal::RNTupleJoinTable &joinTable, ROOT::NTupleSize_t entryOffset = 0) = 0;
 
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Processor-specific implementation for printing its structure, called by PrintStructure().
+   ///
+   /// \param[in] output Output stream to print to.
```
silverweed: Technically `output` is a `param[in,out]`

silverweed (approved):

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -198,6 +202,24 @@ public:
    /// \return A reference to the entry used by the processor.
    const ROOT::REntry &GetEntry() const { return *fEntry; }
 
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Print a graphical representation of the processor composition.
+   ///
+   /// \param[in] output Stream to print to (default is stdout).
```
pcanal: See https://github.com/root-project/root/pull/19696/files#r2287917929

tree/ntuple/inc/ROOT/RNTupleProcessor.hxx
```diff
@@ -198,6 +202,24 @@ public:
...
+   /// RNTuples will be printed as follows:
+   /// ~~~
+   /// +-----------------------------+ +-----------------------------+
+   /// | ntuple                      | | ntuple_aux                  |
+   /// | ntuple.root                 | | ntuple_aux1.root            |
+   /// +-----------------------------+ +-----------------------------+
+   ///                                 +-----------------------------+
+   ///                                 | ntuple_aux                  |
+   ///                                 | ntuple_aux2.root            |
+   ///                                 +-----------------------------+
```
pcanal, line 221: I assume that `PrintStructure` does not (cannot) know the number of rows/entries in each.  Due to that missing information we get a picture that is correct but a bit skewed, i.e. if/when the chain and the primary RNTuple have the same total number of rows, the picture:
```
   /// +-----------------------------+ +-----------------------------+
   /// | ntuple                      | | ntuple_aux                  |
   /// | ntuple.root                 | | ntuple_aux1.root            |
   /// |                             | +-----------------------------+
   /// |                             | +-----------------------------+
   /// |                             | | ntuple_aux                  |
   /// |                             | | ntuple_aux2.root            |
   /// +-----------------------------+ +-----------------------------+
```
might be clearer.  The first picture give an impression of unbalance and thus that it needs/requires a lookup index to match the two sets of rows.

enirolf: I can look into how much work it is to make the things "aligned". This PR only adds documentation so I would address it in a separate one. All things considered, I currently can't put it super high on my priority list, however.

jblomer (approved):

Issue #19491, opened 2025-07-31
Bug in Numba + ROOT


acampove:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Hi,

ROOT seems to be raising exceptions when using numba in certain cases

### Reproducer

```python
import ROOT
import pandas as pnd

from ROOT import RDataFrame, EnableImplicitMT, Math

EnableImplicitMT(4)

@ROOT.Numba.Declare(['float'], 'float')
def my_func(x):
    vec = Math.PtEtaPhiMVector(1, 1, 1, x)
    return vec.M()

rdf = RDataFrame(10)
rdf = rdf.Define('a', 'gRandom->Uniform()')
rdf = rdf.Define('b', 'Numba::my_func(a)')
rdf = rdf.Define('c', 'ROOT::Math::PtEtaPhiMVector vec(1, 1, 1, a); return vec.M();')

data = rdf.AsNumpy()
df   = pnd.DataFrame(data)
print(df)
```

### ROOT version

6.34.04 in AlmaLinux9

### Installation method

micromamba, binary

### Operating system

AlmaLinux9

siliataider:
Dear @acampove,

Thank you for your report!

The behavior you described is not a 'bug' in ROOT itself, but rather a current limitation of the [`cppyy.numba_ext`](https://cppyy.readthedocs.io/en/latest/numba.html) integration.

Specifically, the issue arises because constructing C++ objects inside `numba.njit` functions is not yet supported. This is documented in the [cppyy documentation on Numba support](https://cppyy.readthedocs.io/en/latest/numba.html#examples):

> Instances of C++ classes can be passed into Numba traces. They can be returned from functions called within the trace, but cannot yet be returned from the trace. Their public data is accessible (read-only) if of built-in type and their public methods can be called, for which overload selection works.

In your use case, you're attempting to construct a `ROOT::Math::PtEtaPhiMVector` object inside the `@ROOT.Numba.Declare` decorated function, and this is not supported.
This is not specific to RDataFrame or ROOT’s `@ROOT.Numba.Declare` mechanism. You can reproduce the exact same limitation in pure Python with Numba and cppyy:

- You can do the following:
```py
@numba.njit
def f(x):
    return x.M()

v = ROOT.Math.PtEtaPhiMVector(1, 2, 3, 4)
print(f(v)) # prints 4.0
```

- But not:
```py
@numba.njit
def f(x):
[...]

PR #19697, opened 2025-08-20
[Python][RDF] Add support for (some) root classes in numba jitted functions

siliataider:
# This Pull request:

## Changes or fixes:
This PR extends `Numba` support in the `@ROOT.Numba.Declare` decorator to ROOT C++ classes recognized by the `cppyy` `Numba` extension (see [cppyy docs](https://cppyy.readthedocs.io/en/latest/numba.html)).
- `Numba` can infer supported types automatically when using `@numba.jit(nopython=True)` without passing an explicit signature. Any class understood by cppyy’s Numba extension is now supported.
- For `numba.cfunc`, generating a native C callback for these classes fails. As a workaround, we manually create the C function using `ctypes.CFUNCTYPE` and pass its address to the C++ wrapper.

>[!WARNING]
> This is an experimental feature. It is still incomplete and has only been tested on Linux on x86_64.
> See https://cppyy.readthedocs.io/en/latest/numba.html#numba-support

### Demo
```python
@ROOT.Numba.Declare(["ROOT::Math::PtEtaPhiMVector"], "float")
def get_m(x: ROOT.Math.PtEtaPhiMVector) -> float:
    return x.M()

rdf = ROOT.RDataFrame(3) \
    .Define("x", "(int)rdfentry_") \
    .Define("v", "ROOT::Math::PtEtaPhiMVector(1, 2, 3, 2*x)") \
    .Define("v_m", "Numba::get_m(v)")
```

### Limitations

These classes cannot be constructed within the function or returned, only their public methods can be called, as per cppyy’s documentation:
> Instances of C++ classes can be passed into Numba traces. They can be returned from functions called within the trace, but cannot yet be returned from the trace. Their public data is accessible (read-only) if of built-in type and their public methods can be called, for which overload selection works.

Thus, the use case described in [#19491](https://github.com/root-project/root/issues/19491) is still not supported.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
@@ -10,10 +10,12 @@ def _NumbaDeclareDecorator(input_types, return_type=None, name=None):
     wrapper functions with the given C++ types for input and return types. Eventually,
     the Python callable is accessible in the Numba namespace in C++.
 
-    The implementation first jits with numba the Python callable. We support fundamental types and
-    ROOT::VecOps::RVecs thereof. Note that you can get the jitted Python callable by the attribute
-    numba_func. The C++ types are converted to the respective numba types and RVecs are accessible
-    in Python by numpy arrays. After jitting the actual Python callable, we jit another Python wrapper,
+    The implementation first jits with numba the Python callable. We support fundamental types as well as
+    ROOT::VecOps::RVec, std::vector, and std::array thereof, along with ROOT C++ classes recognized
+    by the cppyy Numba extension (see https://cppyy.readthedocs.io/en/latest/numba.html).
+    Note that you can get the jitted Python callable by the attribute numba_func.
+    The C++ types are converted to the respective numba types and RVecs are accessible in Python
+    by numpy arrays. After jitting the actual Python callable, we jit another Python wrapper,
     which converts the Python signature to a C-friendly signature. The wrapper code is accessible by
     the attribute __py_wrapper__. Next, the Python wrapper is given to cling to jit a C++ wrapper function,
     making the original Python callable accessible in C++. The wrapper code in C++ is accessible by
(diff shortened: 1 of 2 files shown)


bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
```diff
@@ -65,15 +68,17 @@ def get_container_cpp_name(cpp_type: str) -> Union[str, None]:
...
         """
-        Remove ROOT:: and VecOps:: namespaces
+        Make sure all namespaces are included
+        e.g. "RVecI" -> "ROOT::RVecI"
         """
-        return t.replace("ROOT::", "").replace("VecOps::", "")
-
-    input_types = [normalize_typename(t) for t in input_types]
-    return_type = normalize_typename(return_type) if return_type is not None else None
+        container_type = get_container_type(cpp_type)
```
guitargeek:
```suggestion
        container_type = cpp_type if cpp_type in CONTAINER_TYPES else None
```
I think this `get_container_type()` function really obfuscates the code. I would be more explicit, then one has to jump around less to understand the code.

In general, I would avoid having too small functions that can be better inlined for the sake of readability. Also `is_container_type(cpp_type)` could be `cpp_type in CONTAINER_TYPES`. Then when reading the code I understand directly "okay it's checking if it's in the dictionary", and with `is_container_type(cpp_type)` I first assume there is more complicated logic that warrants a separate function.

That's also personal preference of course!

bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
```diff
@@ -65,15 +68,17 @@ def get_container_cpp_name(cpp_type: str) -> Union[str, None]:
...
+        cpp_name = get_container_cpp_name(cpp_type)
+        if cpp_name is None or any(cpp_type.startswith(name) for name in cpp_name):
+            return cpp_type
+        ret = cpp_type.replace(container_type, cpp_name[0])
```
guitargeek:
```suggestion
        ret = cpp_type.replace(cpp_type, cpp_name[0])
```
If I understand the logic correctly, `container_type` is just `cpp_type`, no? Then we don't even need the `container_type` variable (fewer local variables means less confusion).

bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
```diff
@@ -276,6 +287,21 @@ def inner(func, input_types=input_types, return_type=return_type, name=name):
         else:
             pywrapper_return = "return r"
 
+        # bind the arguments that are passed as void* pointers to the correct type
+        pywrapper_bind_lines = "\n    ".join(
+            [
+            "import cppyy" if any(
+                not is_container_type(t) and map_cpp_type(t, as_ctypes=True) == ctypes.c_void_p
+                for t in input_types
+            ) else "",
```
guitargeek:
```suggestion
            "import cppyy",
```
Python caches imported modules, so if `cppyy` is already imported then `"import cppyy"` has no relevant overhead. And when you use ROOT, it is already imported.

I would consider just always doing `"import cppyy"` to simplify the logic

bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
guitargeek: Nevermind, I understand it now! The one line functions need to check if the key is a substring of the `cpp_type`

guitargeek:
Hi @siliataider, there are test failures on the ARM platforms. Can you fix them before I wrap up the review? Thanks!

bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
siliataider: Not really, `cpp_type` is for ex `RVec<int>&`, `container_type` is `RVec` and `cpp_name[0]` is `ROOT::RVec`.
I will rename the variables and add a comment I see how it's a bit confusing

bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
siliataider: I'll do that thanks

siliataider: This is not really about the overhead, doing `import cppyy` inside `@numba.cfunc` does not work, so I only add it if I know that I will be using `ctypes.CFUNCTYPE` to create the c callback.
I wrote it differently and added a comment.

bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
```diff
@@ -208,17 +235,31 @@ def inner(func, input_types=input_types, return_type=return_type, name=name):
...
         except:  # noqa E722
-            raise Exception("Failed to jit Python callable {} with numba.jit".format(func))
+            try:
+                # Fallback to letting numba infer the signature
+                import cppyy.numba_ext  # noqa F401
+                import platform
+                import warnings
+
+                if platform.machine() != "x86_64":
+                    warnings.warn(
```
vepadulano: I'd signal this with a hard error, since we're not sure that by the next ROOT release this part of the cppyy-numba integration will be improved, and an error is better than silently returning wrong data.

bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
```diff
@@ -208,17 +235,31 @@ def inner(func, input_types=input_types, return_type=return_type, name=name):
...
                 nb_return_type = nbjit.nopython_signatures[-1].return_type
         except:  # noqa E722
-            raise Exception("Failed to jit Python callable {} with numba.jit".format(func))
+            try:
+                # Fallback to letting numba infer the signature
+                import cppyy.numba_ext  # noqa F401
+                import platform
+                import warnings
+
+                if platform.machine() != "x86_64":
```
vepadulano:
```suggestion
                if not (platform.system() == "Linux" and platform.machine() == "x86_64"):
```

vepadulano (approved):
Thank you, very nice work! I left a small comment.

bindings/pyroot/pythonizations/python/ROOT/_numbadeclare.py
siliataider: Makes sense, I changed it this way

PR #19699, opened 2025-08-20
[rootest] refine root/io tests, part 2

linev:
1. Use cmake for io/dirdelete test
2. Link executable with dictionary lib in `io/evolution/enums` test
3. Use more distinct fixtures names in several io tests

### DIFF
--- a/roottest/root/io/evolution/enums/CMakeLists.txt
+++ b/roottest/root/io/evolution/enums/CMakeLists.txt
@@ -1,25 +1,29 @@
-# We won't be creating a shared library so we don't want to generate a rootmap file.
-set(CMAKE_ROOTTEST_NOROOTMAP ON)
-ROOT_GENERATE_DICTIONARY(testVectorOfEnumsDict ${CMAKE_CURRENT_SOURCE_DIR}/classes.hxx
-                         LINKDEF classesLinkdef.h)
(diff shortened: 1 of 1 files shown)


roottest/root/io/dirdelete/CMakeLists.txt
```diff
@@ -1,7 +1,3 @@
-#-------------------------------------------------------------------------------
-#
-# Placeholder file to translate the tests to the new CTest system. Meanwhile we
-# define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
-#
-#-------------------------------------------------------------------------------
-ROOTTEST_ADD_OLDTEST(LABELS longtest)
+ROOTTEST_ADD_TEST(dirdelete
+                  MACRO rundirdelete.C
+                  OUTREF dirdelete.fail.ref)
```
pcanal, line 3: Here we lose the information that the test is actually failing.  We could either leave it as is or we could use:

```suggestion
                  OUTREF dirdelete.ref
                  WILLFAIL)
```
either way the issue is now reported at https://github.com/root-project/root/issues/20314

pcanal (approved):
Thanks for you patience.

roottest/root/io/dirdelete/CMakeLists.txt
linev: Moving discussion to the issue

PR #19700, opened 2025-08-20
[roottest] refine root/io tests, part 3

linev:


linev:
This PR may also fix sporadic failures which one tries to address in https://github.com/root-project/root/pull/19840

PR #19701, opened 2025-08-20
[tmva][sofie] fix warning due to incorrect else condition in ROperator_BasicBinary

sanjibansg:
This PR fixes the warnings produced due to the incorrect positioning of the else condition in the Basic Binary Operator of SOFIE.


guitargeek:
For the reviewer: The diff is more obvious if you disable white-space changes: https://github.com/root-project/root/pull/19701/files?diff=unified&w=1

PR #19702, opened 2025-08-20
Speed up CTest, particularly for RDataFrame

hageboeck:
### Allow Ninja compilation in CTest to run in parallel with other tests

roottest Ninja targets are rebuilt as part of a ctest run. If two tests invoke Ninja, race conditions will occur.
To prevent that, ROOT used to run all tests with Ninja compilation serially, reducing the CTest throughput.

Here, by using a `RESOURCE_LOCK` in ctest, Ninja compilations are not run from multiple tests at the same time. They may, however, run in parallel with pre-compiled tests.

Note that this might break if the main build tree is outdated. That is:

| Bad | Good |
|---|---|
|`touch core/...` | `touch core/...` |
| ctest -jN | ninja && ctest -jN |

### Allow multi-threaded RDF tutorials to run in parallel with others
ROOT used to forbid running an MT RDF tutorial in parallel with another MT RDF tutorial. When running `ctest -R dataframe -jN`, this can lead to a long tail of tutorials running completely sequentially. Nevertheless, since ROOT honours `ROOT_MAX_THREADS`, each was only running with 4 cores.
Here, the strict `RESOURCE_LOCK` is removed, and CMake is just informed about the required number of CPU cores. This allows for running a few of the MT tutorials in parallel.


cmake/modules/RootMacros.cmake
```diff
@@ -2369,7 +2369,7 @@ macro(ROOTTEST_COMPILE_MACRO filename)
   endif()
   set_property(TEST ${COMPILE_MACRO_TEST} PROPERTY ENVIRONMENT ${ROOTTEST_ENVIRONMENT})
   if(CMAKE_GENERATOR MATCHES Ninja AND NOT MSVC)
-    set_property(TEST ${COMPILE_MACRO_TEST} PROPERTY RUN_SERIAL true)
+    set_property(TEST ${COMPILE_MACRO_TEST} PROPERTY NINJA_COMPILATION true)
```
pcanal: I could not find `NINJA_COMPILATION` in the CMake documentation, nor the rest of the code.  How does it work?

tutorials/CMakeLists.txt
```diff
@@ -1030,10 +1030,8 @@ if(ROOT_pyroot_FOUND)
                 ${py_will_fail})
 
     if(${t} IN_LIST multithreaded)
-      # Makes sure that this doesn't run in parallel with other multithreaded tutorials, and that cmake doesn't start too
-      # many other tests. That we use 4 processors is actually a lie, because IMT takes whatever it finds.
-      # However, even this poor indication of MT behaviour is a good hint for cmake to reduce congestion.
-      set_tests_properties(${tutorial_name} PROPERTIES RESOURCE_LOCK multithreaded PROCESSORS ${NProcessors})
+      # Makes sure that this doesn't run in parallel with too many other multithreaded tutorials.
```
pcanal, line 1048:
```suggestion
      # Makes sure that this doesn't run in parallel with too many other multithreaded tutorials.
      # The number of threads used by the tutorials is limited by setting ROOT_MAX_THREADS (see above).
```

cmake/modules/RootMacros.cmake
hageboeck: That's actually a mistake originating from search+replace. 🙂
It should read `RESOURCE_LOCK NINJA_COMPILATION`, I'll change that.

pcanal (approved):
Thanks.

cmake/modules/RootMacros.cmake
```diff
@@ -2714,7 +2714,7 @@ macro(ROOTTEST_GENERATE_EXECUTABLE executable)
   endif()
 
   if(CMAKE_GENERATOR MATCHES Ninja AND NOT MSVC)
-    set_property(TEST ${GENERATE_EXECUTABLE_TEST} PROPERTY RUN_SERIAL true)
+    set_property(TEST ${GENERATE_EXECUTABLE_TEST} PROPERTY RESOURCE_LOCK NINJA_COMPILATION)
```
pcanal: Important note:  This changes works ***IF AND ONLY IF*** the ninja compilation does not induce spurious rebuild of the core libraries .. which unfortunately does happens from time to time (for example on Windows or any other machine with 'low' timestamp granularity where ninja gets confused and rebuild things ... things may or may not be better after we recently removed a large swat of spurious updates/rebuild when doing 2 consecutive builds)

guitargeek:
For the `ROOTTEST_COMPILE_MACRO`, I think the simpler solution would be to not wrap the `root ... macro.C` command in a target, as I propose in this PR:

  * https://github.com/root-project/root/pull/19780

cmake/modules/RootMacros.cmake
hageboeck: As can be seen from the line above, Windows was never part of this adventure, so I guess that we don't need to worry about this in this context.

Regarding rebuilds, I'm testing at the moment a fix for the outdated build tree problem mentioned in the PR description. Will ping once done.

hageboeck: OK, this has hopefully been addressed now:
The PR description was updated, and a test fixture was added that refreshes the entire build tree before invoking Ninja in any of the tests.

I see two possible failure modes:
- A truncated Ninja file due to a previous build with race conditions (Ninja warns about that): Fixed by rerunning cmake.
- Low timestamp granularity --> spurious rebuild: All changes in this PR have always been in an "`if`" that excludes Windows, so unless there is another platform with low granularity, no spurious rebuilds should occur. If they did, let's not use Ninja on those platforms. 🙂

hageboeck:

PR #19707, opened 2025-08-21
[hadd] Improve verbosity settings

### LINKED ISSUE #19479: Info() logs should not go to `stderr`

silverweed:
New verbosity works like this:

  v0: only output errors
  v1: only output errors and warnings
  v2: output errors, warnings and TFileMerger messages
  v3: output all messages

This is largely backwards-compatible as the default verbosity was and still is 99. 

One message (and one error!) were for some reason only emitted when verbosity was exactly 1, which makes no sense. This PR changes it to a regular Info() and Err() respectively. This means that by default hadd will now emit an additional message at the end of the merging process.

## Checklist:

- [x] tested changes locally

This PR fixes #19479

### DIFF
--- a/main/src/hadd.cxx
+++ b/main/src/hadd.cxx
@@ -63,16 +63,21 @@
                        whitespaces or '/'. You can also pass TDirectory names, which apply to the entire directory
                        content. Lines beginning with '#' are ignored. If this flag is passed, "-Ltype" MUST be
                        passed as well.
-  \param -Ltype <SkipListed|OnlyListed> Sets the type of operation performed on the objects listed in FILE given with the
+  \param -Ltype <SkipListed|OnlyListed> Sets the type of operation performed on the objects listed in FILE given with
+  the
                        "-L" flag. "SkipListed" will skip all the listed objects; "OnlyListed" will only merge those
                        objects. If this flag is passed, "-L" must be passed as well.
   \param -n <N_FILES>  Open at most `N` files at once (use 0 to request to use the system maximum - which is also
                        the default). This number includes both the input reading files as well as the output file.
-                       Thus, if set to 1, it will be automatically replaced to a minimum of 2. If set to a too large value,
-                       it will be clipped to the system maximum.
+                       Thus, if set to 1, it will be automatically replaced to a minimum of 2. If set to a too large
+  value, it will be clipped to the system maximum.
   \param -O            Re-optimize basket size when merging TTree
   \param -T            Do not merge Trees
-  \param -v [LEVEL]    Explicitly set the verbosity level: 0 request no output, 99 is the default
+  \param -v [LEVEL]    Explicitly set the verbosity level:
+                            <= 0 = only output errors;
+                            1 = only output errors and warnings;
+                            2 = output minimal informative messages, errors and warnings;
+                            >= 3 = output all messages (default).
   \return hadd returns a status code: 0 if OK, 1 otherwise
 
   For example assume 3 files f1, f2, f3 containing histograms hn and Trees Tn
@@ -155,6 +160,7 @@
 #include <optional>
 #include <sstream>
 #include <string>
+#include <streambuf>
 
 #ifndef R__WIN32
 #include "ROOT/TProcessExecutor.hxx"
@@ -162,22 +168,54 @@
 
 ////////////////////////////////////////////////////////////////////////////////
 
-inline std::ostream &Err()
+// NOTE: TFileMerger will use PrintLevel = gVerbosity - 1. If PrintLevel is < 1, it will print nothing, otherwise
+// it will print everything. To give some granularity to hadd, we do the following:
+// gVerbosity = 0: only print hadd errors
+// gVerbosity = 1: only print hadd errors + warnings
+// gVerbosity = 2: print hadd errors + warnings and TFileMerger messages
+// gVerbosity > 2: print all hadd and TFileMerger messages.
+static int gVerbosity = 99;
+
+namespace {
+
+class NullBuf : public std::streambuf {
+public:
+   int overflow(int c) final { return c; }
+};
+
+class NullStream : public std::ostream {
+   NullBuf fBuf;
+
+public:
+   NullStream() : std::ostream(&fBuf) {}
+};
+
+} // namespace
+
+static NullStream &GetNullStream()
+{
+   static NullStream nullStream;
+   return nullStream;
+}
+
+static inline std::ostream &Err()
 {
    std::cerr << "Error in <hadd>: ";
    return std::cerr;
 }
 
-inline std::ostream &Warn()
+static inline std::ostream &Warn()
 {
-   std::cerr << "Warning in <hadd>: ";
-   return std::cerr;
+   std::ostream &s = gVerbosity < 1 ? GetNullStream() : std::cerr;
+   s << "Warning in <hadd>: ";
+   return s;
 }
 
-inline std::ostream &Info()
+static inline std::ostream &Info()
 {
-   std::cerr << "Info in <hadd>: ";
-   return std::cerr;
+   std::ostream &s = gVerbosity < 3 ? GetNullStream() : std::cerr;
+   s << "Info in <hadd>: ";
+   return s;
 }
 
 using IntFlag_t = uint32_t;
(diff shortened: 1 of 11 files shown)


dpiparo (approved):
Thanks for putting some order in this!

main/src/hadd.cxx
```diff
@@ -155,29 +160,62 @@
...
 ////////////////////////////////////////////////////////////////////////////////
 
-inline std::ostream &Err()
+// NOTE: TFileMerger will use PrintLevel = gVerbosity - 1. If PrintLevel is < 1, it will print nothing, otherwise
+// it will print everything. To give some granularity to hadd, we do the following:
+// gVerbosity = 0: only print hadd errors
+// gVerbosity = 1: only print hadd errors + warnings
+// gVerbosity = 2: print hadd errors + warnings and TFileMerger messages
+// gVerbosity > 2: print all hadd and TFileMerger messages.
+static int gVerbosity = 99;
```
pcanal: Minor
```suggestion
static int gHaddVerbosity = 99;
```
to clarify to the reader (that is not also checking the declaration) of the point of use that this is limited to hadd.

roottest/root/io/hadd/hadd_args_cachesize.ref
```diff
@@ -3,3 +3,4 @@ Info in <hadd>: Using cachesize=100M
 Info in <hadd>: compression setting for all output: 101
 hadd Source file 1: hadd_args_in.root
 hadd Target path: hadd_args_cachesize_out.root:/
+Info in <hadd>: merged 1 (1) input (partial) files into hadd_args_cachesize_out.root
```
pcanal: Those message were not meant to be seen by default.

main/src/hadd.cxx
```diff
@@ -155,29 +160,62 @@
...
 }
 
-inline std::ostream &Info()
+static inline std::ostream &Info()
 {
-   std::cerr << "Info in <hadd>: ";
-   return std::cerr;
+   std::ostream &s = gVerbosity < 3 ? GetNullStream() : std::cerr;
+   s << "Info in <hadd>: ";
+   return s;
```
pcanal:
```suggestion
static inline std::ostream &Info(int level)
{
   std::ostream &s = gVerbosity < level ? GetNullStream() : std::cerr;
   s << "Info in <hadd>: ";
   return s;
```

main/src/hadd.cxx
```diff
@@ -63,16 +63,21 @@
...
+                       Thus, if set to 1, it will be automatically replaced to a minimum of 2. If set to a too large
+  value, it will be clipped to the system maximum.
   \param -O            Re-optimize basket size when merging TTree
   \param -T            Do not merge Trees
-  \param -v [LEVEL]    Explicitly set the verbosity level: 0 request no output, 99 is the default
+  \param -v [LEVEL]    Explicitly set the verbosity level:
+                            <= 0 = only output errors;
+                            1 = only output errors and warnings;
+                            2 = output minimal informative messages, errors and warnings;
+                            >= 3 = output all messages (default).
```
pcanal:
```suggestion
                            <= 0 = only output errors;
                            1 = only output errors and warnings;
                            2 = output minimal informative messages, errors and warnings (default);
                            >= 3 = output all messages.
```

main/src/hadd.cxx
```diff
@@ -674,8 +712,7 @@ int main(int argc, char **argv)
    }
    targetname = argv[args.fOutputArgIdx];
 
-   if (verbosity > 1)
-      Info() << "target file: " << targetname << "\n";
+   Info() << "target file: " << targetname << "\n";
```
pcanal:
```suggestion
   Info(2) << "target file: " << targetname << "\n";
```

main/src/hadd.cxx
```diff
@@ -674,8 +712,7 @@ int main(int argc, char **argv)
...
 
    if (args.fCacheSize)
       Info() << "Using " << cacheSize << "\n";
```
pcanal:
```suggestion
      Info(2) << "Using " << cacheSize << "\n";
```

main/src/hadd.cxx
```diff
@@ -758,21 +795,22 @@ int main(int argc, char **argv)
...
-   if (verbosity > 1) {
-      if (args.fKeepCompressionAsIs && !args.fReoptimize)
-         Info() << "compression setting for meta data: " << newcomp << '\n';
-      else
-         Info() << "compression setting for all output: " << newcomp << '\n';
-   }
+   if (args.fKeepCompressionAsIs && !args.fReoptimize)
+      Info() << "compression setting for meta data: " << newcomp << '\n';
+   else
+      Info() << "compression setting for all output: " << newcomp << '\n';
```
pcanal:
```suggestion
   if (args.fKeepCompressionAsIs && !args.fReoptimize)
      Info(2) << "compression setting for meta data: " << newcomp << '\n';
   else
      Info(2) << "compression setting for all output: " << newcomp << '\n';
```

main/src/hadd.cxx
```diff
@@ -893,16 +931,12 @@ int main(int argc, char **argv)
 #endif
 
    if (status) {
-      if (verbosity == 1) {
-         Info() << "merged " << allSubfiles.size() << " (" << fileMerger.GetMergeList()->GetEntries()
-                << ") input (partial) files into " << targetname << ".\n";
-      }
+      Info() << "merged " << allSubfiles.size() << " (" << fileMerger.GetMergeList()->GetEntries()
+             << ") input (partial) files into " << targetname << "\n";
```
pcanal:
```suggestion
      Info(3) << "merged " << allSubfiles.size() << " (" << fileMerger.GetMergeList()->GetEntries()
             << ") input (partial) files into " << targetname << "\n";
```

pcanal (changes requested):
See previous comment.

main/src/hadd.cxx
```diff
@@ -893,16 +931,12 @@ int main(int argc, char **argv)
...
       return 0;
    } else {
-      if (verbosity == 1) {
-         Err() << "failure during the merge of " << allSubfiles.size() << " ("
-               << fileMerger.GetMergeList()->GetEntries() << ") input (partial) files into " << targetname << ".\n";
-      }
+      Err() << "failure during the merge of " << allSubfiles.size() << " (" << fileMerger.GetMergeList()->GetEntries()
```
pcanal, line 939: I assume the message was previously only optionally printed because it was assumed that the main cause of failure would already have been printed message and this was 'only' extra details about the setups.

Unrelated to this PR, the message could be improved to clarify what the 2 numbers means.

roottest/root/io/hadd/hadd_args_cachesize.ref
silverweed: Then how should we treat them? It's odd that the user has to decrease the verbosity to see them.
Should we set the default verbosity to 3 and gate those messages to level 4+?

silverweed:
I agree on adding the multiple levels to Info(), but what about all the messages you didn't correct? Are those all `2`? Or `3`?
If, as the description says, 2 is only for "minimal informative messages", I would expect most messages to be at level 3, but that would change what is shown by default if we set the default to 2...

pcanal:
I think that `2 is for "minimal informative messages",` is the 'right' default.  

I already made some proposition inline, here is another stab:
```
601:         Info(2) << "added " << nObjects << " object from filter file '" << *filterFileName << "'\n"; 
647:      Info(2) << "parallelizing  with " << nProcesses << " processes.\n";
678:      Info(2) << "target file: " << targetname << "\n";
681:      Info(3) << "Using " << cacheSize << "\n";
763:         Info(2) << "compression setting for meta data: " << newcomp << '\n';
765:         Info(2) << "compression setting for all output: " << newcomp << '\n';
784:      Info(2) << "each process should handle at least 3 files for efficiency."
897:         Info(3) << "merged " << allSubfiles.size() << " (" << fileMerger.GetMergeList()->GetEntries()
```

pcanal (approved):
LGTM.  Thanks.

PR #19709, opened 2025-08-21
[roottest] use cmake for io/newstl tests, part 4

linev:
Compile macros in advance and use them from srun.C script 
Avoid usage of rootlogon.C - which automatically involved for all scripts 
Adjust stlIoTest.ref with actual warnings
Remove execute flag for many files
Mark as failing readNoLib.C - was fully disabled in Makefile

### DIFF
--- a/roottest/root/io/newstl/CMakeLists.txt
+++ b/roottest/root/io/newstl/CMakeLists.txt
@@ -52,7 +52,7 @@ endif()
 ROOTTEST_ADD_TEST(ComplexTest
                   MACRO runComplexTest.C
                   ROOTEXE_OPTS ${_complextest_load}
-                  OUTREF ComplexTest.ref
+                  OUTREF ComplexTest${ref_suffix}
                   FIXTURES_REQUIRED root-io-newstl-ComplexTest_h-fixture)
 
 ROOTTEST_ADD_TEST(SampleFile
--- a/roottest/root/io/newstl/ComplexTest_win64.ref
+++ b/roottest/root/io/newstl/ComplexTest_win64.ref
@@ -1,0 +1,8 @@
+
+Processing runComplexTest.C...
+Warning in <TStreamerInfo::Build>: _Complex_base<double,_C_double_complex>: base class _C_double_complex has no streamer or dictionary it will not be saved
+Warning in <TStreamerInfo::Build>: complex<double>: base class _Complex_base<double,_C_double_complex> has no streamer or dictionary it will not be saved
+The complex object was set properly:
+3 6
+The complex object was read properly:
+3 6
--- a/roottest/root/io/newstl/ComplexTest_win32.ref
+++ b/roottest/root/io/newstl/ComplexTest_win32.ref
@@ -1,0 +1,8 @@
+
+Processing runComplexTest.C...
+Warning in <TStreamerInfo::Build>: _Complex_base<double,_C_double_complex>: base class _C_double_complex has no streamer or dictionary it will not be saved
+Warning in <TStreamerInfo::Build>: complex<double>: base class _Complex_base<double,_C_double_complex> has no streamer or dictionary it will not be saved
+The complex object was set properly:
+3 6
+The complex object was read properly:
+3 6


roottest/root/io/newstl/CMakeLists.txt
```diff
@@ -1,7 +1,107 @@
...
+  list(GET _holders ${index} holder)
+  list(GET _tests ${index} test)
+  ROOTTEST_COMPILE_MACRO(${holder}.C
+                         BUILDLIB TestHelpers_C
+                         FIXTURES_REQUIRED root-io-newstl-TestHelpers-fixture
+                         FIXTURES_SETUP root-io-newstl-${holder}-fixture)
+  ROOTTEST_COMPILE_MACRO(${test}.C
+                         BUILDLIB ${holder}_C
+                         FIXTURES_REQUIRED root-io-newstl-${holder}-fixture
+                         FIXTURES_SETUP root-io-newstl-test-fixture)
```
pcanal, line 17:
```suggestion
                         FIXTURES_SETUP root-io-newstl-${test}-fixture)
```

linev: It is done intentionally.
All macros have to be compiled to fulfill `root-io-newstl-test-fixture` fixture.
And only then `srun.C` macro will be executed.

With your approach one will define 8 independent fixtures and have to specify them all for running `srun.C`.

roottest/root/io/newstl/ComplexTest_win64.ref
```diff
@@ -0,0 +1,8 @@
+
+Processing runComplexTest.C...
+Warning in <TStreamerInfo::Build>: _Complex_base<double,_C_double_complex>: base class _C_double_complex has no streamer or dictionary it will not be saved
+Warning in <TStreamerInfo::Build>: complex<double>: base class _Complex_base<double,_C_double_complex> has no streamer or dictionary it will not be saved
```
pcanal, line 4: Those are an issue per se.  I realize that the test was not run on Windows before and so we did not noticed.  We should open an issue to track the needed removal of those (i.e. by updating the I/O code to properly handle std::complex that have the same implementation as Windows).

roottest/root/io/newstl/ComplexTest.h
```diff
@@ -1,25 +1,32 @@
...
-   
-  ClassDef(Test, 1);
+
+  ClassDefOverride(Test, 1);
 };
+
+#ifdef __ROOTCLING__
+#pragma link C++ class std::complex<double>+;
+#pragma link C++ class Test+;
+#endif
```
pcanal, line 30: Why are those needed?  The one from `Test` is redundant (but harmless) since the header is handled by `ACLiC`.  The dictionary for `std::complex<double>` is meant to be pre-builts:
```
$ root.exe -b -l 
root [0] TClass::GetClass("std::complex<double>")
(TClass *) 0x10e708080
root [1] TClass::GetClass("std::complex<double>")->IsLoaded()
(bool) true
```
So should be redundant (and wasteful - dictionary generated twice).

pcanal: If I remember correctly, the `complex` dictionary are provided by `core/clingutils/src/complexLinkdef.h` (which may or may not be used on Windows!?)

linev: After two months I am not sure why I add these lines.
Probably trying to solve issue on Windows.

roottest/root/io/newstl/srun.C
```diff
@@ -1,104 +1,120 @@
...
+      result = gSystem->CompileMacro(srcpath + "ntest.C","k");
       if (result != 0 && result != 1) return;
 
-      if (exec) gROOT->ProcessLine("ntest(exec-1)");
+      if (exec) gROOT->ProcessLine("ntest" + arg);
 
    }
 
-   pass++;
+   // pass++;
```
pcanal, line 119: Why remove this? (still used in `run.C`)

roottest/root/io/newstl/sample_writer.C
```diff
@@ -1,13 +1,9 @@
...
+
+#ifndef ClingWorkAroundMissingDynamicScope
   gROOT->ProcessLine(".L sample_bx_classes.C+");
 #endif
 
-#if defined(ClingWorkAroundMissingDynamicScope) && !defined(SECOND_RUN)
-#define SECOND_RUN
-  gROOT->ProcessLine(".x sample_writer.C");
-
-#else
```
pcanal, line 10: For the record, this trades off a (noticeable) simplification of the code with an increase in complexity in calling the script by hand (need to know that a library needs to be loaded explicitly).  This is mitigated by the fact that the invocation is provided by `ctest -V` (or log) when needed.

roottest/root/io/newstl/srun.C
linev: This variable was defined in `rootlogon.C` script - which is now completely removed.
Not sure if run.C used.

linev: As I can see - run.C is not used and can be removed as well

roottest/root/io/newstl/test.C
```diff
@@ -28,8 +28,9 @@ template <class HolderClass> void write(const char *testname, int nEntry = 3) {
    dirname.ReplaceAll("/","-");
 
    gSystem->mkdir(dirname);
```
pcanal, line 30: Note that in the `Makefile` we were also calling `mkdir` and that directory in order to avoid the race condition of several test call this `mkdir` on the same directory at the same time.  I am not sure whether this is still relevant/needed (Probably not)

linev: Tests are not running in parallel - they compiled in parallel, but then executed one-by-one from srun.C macro.
So there should be no race here

roottest/root/io/newstl/readNoLib.C
```diff
@@ -1,6 +1,20 @@
-void readNoLib(const char *filename = 0) {
-   TFile *_file0 = TFile::Open(filename ? filename : "4-01-03/vector.root");
-   
+int readNoLib(const char *fname = "vector.root")
+{
+   TString dirname = gROOT->GetVersion();
```
pcanal, line 3: If I understand correctly, the current directory in which the test is moving from `root_src/roottest/root/io/newstl` (within the source) to `build_dir/roottest/root/io/newstl`.

It is no clear whether we need to move the directory in which those output file are being generated.  Those files are intended to be 'precious' and be available to be gathered for testing of forward compatibility.

In the old days, the roottest source dir would be kept around long term (on the developers development's machine) and naturally test over time that forward compatibility by keeping around those files whenever the local version the ROOT source would switch over to a new version.

With the change to CMake, they are now being produced within a non-precious area (the build directory is easily discarded since it is 'fully' reproducible ... except for those files .... ).

Side note: Currently (even with the Makefile) the CI does not leverage this features, since it always starts mostly from scratch and does not download/install any older version.

@dpiparo FYI

linev: This I did not want to change in the PR.
Later one can do it if we want.

roottest/root/io/newstl/ComplexTest_win64.ref
pcanal: Moving this discussion to https://github.com/root-project/root/issues/20312

roottest/root/io/newstl/ComplexTest.h
pcanal: Moving this discussion to https://github.com/root-project/root/issues/20312

roottest/root/io/newstl/srun.C
pcanal: > run.C is not used and can be removed as well

Fair enough.  Indeed it looks obsolete.

roottest/root/io/newstl/test.C
pcanal: Fair enough.  Thanks.

roottest/root/io/newstl/readNoLib.C
pcanal: It is unfortunate I did not catch this sooner.  As it is, I agree that this should be handle in a new PR, so I opened https://github.com/root-project/root/issues/20313

roottest/root/io/newstl/srun.C
linev: Will be removed in following PRs

PR #19710, opened 2025-08-21
[Python] Implement NumPy array lifetime management for RDataFrame in C++

### LINKED ISSUE #19706: Crash when exiting Python interpreter after using `ROOT.RDF.FromNumpy`
### Check duplicate issues.

- [x] Checked for duplicates

### Description

I would expect no crash

### Reproducer

```python
# repro.py
import ROOT
import numpy as np

x = np.array([1, 2, 3, 4, 5], dtype=np.float64)
y = np.array([2, 4, 6, 8, 10], dtype=np.float64)

df = ROOT.RDF.FromNumpy({'x': x, 'y': y})
```

```
python -i repro.py
>>> <Ctrl+D>

terminate called after throwing an instance of 'CPyCppyy::PyException'
  what():  TypeError: callable was deleted
 *** Break *** abort

<stack trace>
```

### ROOT version

```
  | Welcome to ROOT 6.36.000                       https://root.cern |
  | Built for linuxx8664gcc on Jun 07 2025, 14:20:17                 |
  | From tags/6-36-000@6-36-000                                      |
  | With g++ (Spack GCC) 14.2.0                                      |
```

### Installation method

source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh

### Operating system

Alma9

### Additional context

_No response_

guitargeek:
Implement NumPy array lifetime management for RDataFrame purely in C++, so we don't need to call back into Python. This is more stable, as the Python callback might be destroyed before the RDataSource when the Python interpreter is shut down.

Follows up on #15031.

We can further follow up on this by moving the Python object deleter callback part of the Cppyy API, so we don't have to use `gInterpreter.Declare`.

Closes #19706.

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdataframe.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdataframe.py
@@ -540,11 +540,6 @@ def _make_name_rvec_pair(key, value):
     return ROOT.std.pair["std::string", type(pyvec)](key, ROOT.std.move(pyvec))
 
 
-# For references to keep alive the NumPy arrays that are read by
-# MakeNumpyDataFrame.
-_numpy_data = {}
-
-
 def _MakeNumpyDataFrame(np_dict):
     r"""
     Make an RDataFrame from a dictionary of numpy arrays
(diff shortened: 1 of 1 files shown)


siliataider (approved):
More stable indeed, thank you!

wlav:
I have no clue what's going on here, but as a quick note, although `Py_INCREF` and `Py_DECREF` are macros, `Py_IncRef` and `Py_DecRef` are functions that you can just call, so that `cppdef` seems to be unnecessary:
```
>>> import cppyy
>>> cppyy.include("Python.h")
True
>>> cppyy.gbl.Py_IncRef
<cppyy.CPPOverload object at 0x104c34540>
>>> cppyy.gbl.Py_DecRef
<cppyy.CPPOverload object at 0x104c2ef00>
>>>
```
(Note that you don't even need to include "Python.h" b/c these functions are part of the stable API, so it's safe to redeclare them and the linker symbols exists if you're in the Python interpreter.)

guitargeek:
Thanks for the comment! Good point, I can also call the functions in Python. But the point of this PR was to avoid callbacks to Python.

What's going on here is that we need to pass a callback to `Py_DecRef` to the RDataFrame C++ so it can decrement the reference count for some NumPy arrays when it doesn't need them anymore. But the callback needs to by implemented in C++, because otherwise the callback might be destroyed before the RDataFrame. And we can't put a lifeline on the Python side, because the NumPy datasource is a member of the RDataFrame that is moved around in C++ when the RDataFrame is transformed by adding operations (kind of like a shared pointer between different RDataFrame instances).

But admittedly, this Python-free callback to do decref is probably only needed in this corner case. But if there are more cases like this, I will come back to it.

The `gInterpreter` declare call doesn't harm here because it only happens once, so we can keep it like this!

wlav:
> But the point of this PR was to avoid callbacks to Python.

I don't understand that: whether you call the function or the macro, an undetermined amount of Python code may be triggered from this callback if the subjected `PyObject`'s ref-count hits zero. You have to hold the GIL for example, when executing this callback.

guitargeek:
Fair enough, calling `Py_DECREF`/`Py_DecRef` can imply more Python code that is called, that is true. That might also be the problem, especially when decref is called without holding the GIL, but it's not the immediate problem that this PR fixed.

Let me say it differently: before this PR, a Python lambda function was used as a callback to indirectly wrap the desired reference count decrements. These lambda functions were wrapped in `std::function<void()>` for the C++ side, something that CPyCppyy takes care of. But to be really correct and avoid crashes when shutting down the interpreter, the `std::function` would need to hold a hard reference to the Python lamda function, and release it when the `std::function` gets destroyed. That would be quite tricky to implement. Right now, Python callables passed to C++ are borrowed. To work around this limitation, I refactored the code to not use a Python callable as a callback, but instead use a C++ callable that does what we want (in this case, decrementing the reference count of the NumPy arrays).

In the end, the problem is that I want to pass a hard reference to a Python object (in this case the callback) to C++ code that doesn't know anything about Python.

wlav:
One more nitpick: the references aren't borrowed, but weak references. Shouldn't crash, but raise a Python exception when the lambda has gone away (which, yes, would be problematic during shutdown). An alternative then would have been to use a Python function rather than a Python lambda.

guitargeek:
> An alternative then would have been to use a Python function rather than a Python lambda.

Right, I also tried that, but that was not possible because the function needs to capture a variable (the PyObject pointer). I would have avoided that by also storing the pointer on the C++ side, that's true, but then I would have had to change the C++ code in RVecDS too. And I didn't want the details of MakeNumpyDataFrame to leak into this general data source too much

PR #19718, opened 2025-08-21
[CMake] Remove overriding `CMAKE_OSX_SYSROOT`

guitargeek:
The internet says that newer CMake versions got more reliable in figuring out the SDK path. So setting `CMAKE_OSX_SYSROOT` ourselves is maybe not needed anymore.

Follows up on #19218.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,19 +12,6 @@ if(WIN32)
   cmake_policy(SET CMP0091 OLD)
   set(CMAKE_SKIP_TEST_ALL_DEPENDENCY TRUE)
 endif()
-if(CMAKE_HOST_APPLE AND (NOT CMAKE_OSX_SYSROOT OR CMAKE_OSX_SYSROOT STREQUAL ""))
-  # The SYSROOT *must* be set before the project() call
-  execute_process(COMMAND xcrun --sdk macosx --show-sdk-path
-    OUTPUT_VARIABLE SDK_PATH
-    ERROR_QUIET
-    OUTPUT_STRIP_TRAILING_WHITESPACE)
-
-  if(NOT EXISTS "${SDK_PATH}")
-    message(FATAL_ERROR "Could not detect macOS SDK path")
-  endif()
-
-  set(CMAKE_OSX_SYSROOT "${SDK_PATH}" CACHE PATH "SDK path" FORCE)
-endif()
 
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR AND NOT ALLOW_IN_SOURCE)
   message(FATAL_ERROR


guitargeek:
I don't understand what is going on at this point. PR #18318, which introduced this code, said it was to fix build errors with CMake 4. However, the two macOS platforms that use CMake 4.0.3 (`mac15` and `mac-beta`) already succeeded building. @devajithvs, can you maybe elaborate on why #18318 was needed in the first place? As I said already in https://github.com/root-project/root/pull/19218#issuecomment-3211850858, after studying the CMake source code, I don't see the reason why things should have failed with CMake 4 to begin with.

devajithvs:
This was needed because, when CMake 4 on macos-15-4 first became available, it failed to set the `CMAKE_OSX_SYSROOT` variable. As a result, ROOT would fail to compile since the correct modulemap could not be found, as discussed here: https://root-forum.cern.ch/t/issue-with-macos-15-4-xcode-16-3/63327/2 - this was an emergency fix back then.

guitargeek:
Thanks for giving the context! Do we still need it now? CMake 4.0 probably went to not setting `CMAKE_OSX_SYSROOT` for a good reason, so maybe we should not reproduce the previous 3.x behavior

devajithvs:
Yes, I agree. If the build doesn't fail anymore. It's nice to revert to the previous behavior.

devajithvs (approved):

devajithvs:
@couet reported some failing builds post merging this on mac-beta (same error as in: https://root-forum.cern.ch/t/issue-with-macos-15-4-xcode-16-3/63327). Configuring with `-DCMAKE_OSX_SYSROOT=$(xcrun --sdk macosx --show-sdk-path)` fixed it. 

cc: @guitargeek

guitargeek:
That's strange, do you have any hint why we don't see it in the CI?

devajithvs:
> That's strange, do you have any hint why we don't see it in the CI?

Not clear. It is indeed strange. I (and @couet) tried a build on the same machine with the exact same configuration as the nightlies and was not able to get it to build.

vepadulano:
Hi @devajithvs @guitargeek . I have been debugging build failures on my Mac laptop after a recent rebase and I could trace it to this PR. With this change ROOT does not build on my laptop. Here is some info from my system
```
vpadulan@vpadulan-macbook root % cmake --version
cmake version 4.1.1

CMake suite maintained and supported by Kitware (kitware.com/cmake).
vpadulan@vpadulan-macbook root % clang --version
Apple clang version 17.0.0 (clang-1700.0.13.5)
Target: arm64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
vpadulan@vpadulan-macbook root % sw_vers 
ProductName:		macOS
ProductVersion:		15.6.1
BuildVersion:		24G90
vpadulan@vpadulan-macbook root % xcrun --show-sdk-version
15.5
vpadulan@vpadulan-macbook root % pkgutil --pkg-info=com.apple.pkg.CLTools_Executables | grep version
version: 16.4.0.0.1.1747106510
```

Please get in touch with me and let's try to understand the situation better. We have to avoid that this happens again to users in the wild as the error during the build is absolutely obscure.

PR #19719, opened 2025-08-21
[ci] Remove redundant variables in global CI config

guitargeek:
Also, do the internet connection check in one of our CI configurations, as requested by @pcanal (https://github.com/root-project/root/pull/19632#issuecomment-3200517980).

More (maybe too much) detail in the commit descriptions.


.github/workflows/root-ci-config/buildconfig/mac15.txt
```diff
@@ -23,9 +23,8 @@ builtin_xrootd=ON
 builtin_xxhash=ON
 builtin_zeromq=ON
 builtin_zstd=ON
+check_connection=ON
```
pcanal: The main challenge with picking this as the sole `check_connection` build is that overtime (several years as you mentioned) we will forget to migrate it to a new mac when mac15 is retired.  For this reason, I would actually set it to mac-beta instead ...

guitargeek: Yes okay, I changed that. I had no strong preference after all

hahnjo:
This breaks the option hashing in our CI script. I'm revert in https://github.com/root-project/root/pull/19721.

To elaborate a bit more, I disagree with the reasons given in the commit message: Our CI's purpose is *not* to test the default configuration, but to make sure that a stable set of (more or less) carefully chosen platforms and options continues to work for every change we commit. In order to incorporate changes in the configuration, our CI script hashes all options to determine if an incremental build makes sense. By relying on default values that can potentially change, this mechanism is rendered useless.

guitargeek:
We use the same configuration for testing and for building the binaries that we ship to the users, if I understand correctly. If the `global.txt` would only be used for our CI tests and then we have a separate configuration for the releases that only deviates from the defaults in a controlled and justified way, then I would agree with you.

But if `globals.txt` is used for the binaries, option divergence ultimately leads to user confusion, like "Why does this script from my colleague not work on my PC? He built ROOT from source and I downloaded the binaries". I have had these questions before.

Let me also add that I think it's good that we use the same configuration for CI and releases. because like this we actually test what we intend to release.

PR #19720, opened 2025-08-21
[ci] Use `alma8` image with Python 3.12

guitargeek:
Let's see if this works. We need to update the Python version in `alma8`, because right now we're using a random Python version in Alma 8 that is not supported by RHEL anymore.

See https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle

The only Python versions in RHEL 8 that are supported until its end-of-life are Python 3.6 and Python 3.12. Since ROOT requires Python 3.12, we need to use Alma 8 with Python 3.12.

Right now, our CI images install Python 3.8 on RHEL 3.8, but this intermediate version is not supported anymore by RHEL.


hahnjo:
I think this is a bad idea: As far as I can tell, `alma8` is currently the only platform testing our minimum of Python 3.8. In my opinion, if we claim to support it, we must make sure it works. Is there a concrete problem of keeping Python 3.8 for `alma8`? I don't think we care about updates and security fixes for our CI...

guitargeek:
> I think this is a bad idea: As far as I can tell, `alma8` is currently the only platform testing our minimum of Python 3.8. In my opinion, if we claim to support it, we must make sure it works.

Fortunately this is not a problem anymore, because we have nightly Python builds for all Python version we support.

> Is there a concrete problem of keeping Python 3.8 for `alma8`? I don't think we care about updates and security fixes for our CI...

We use the same config for the release, and there we should care about it. Right now, we are shipping binaries built against a Python version that RHEL retired over 2 years ago:

Screenshot_2025-08-19_15-57-58

hahnjo:
> Fortunately this is not a problem anymore, because we have nightly Python builds for all Python version we support.

Yes, but not for every PR build that may break it by using syntax that doesn't work in Python 3.8...

guitargeek:
That's true, so maybe we can make this change to the CI only once we change the minimum supported Python version.

In any case, the tests don't even pass :laughing: Let's maybe come back to this later when we're actually getting closer to 6.38 branching/release.

PR #19721, opened 2025-08-22
Revert "[ci] Remove redundant variables in global CI config"

hahnjo:
This reverts commit 2f2a29ba6384fe2f26716c385769cbcdd5b60df8. It breaks the option hashing in our CI script.


hageboeck:
> This reverts commit [2f2a29b](https://github.com/root-project/root/commit/2f2a29ba6384fe2f26716c385769cbcdd5b60df8). It breaks the option hashing in our CI script.

While I agree that having removed these options changes the hash, does it also lead to clashes? `sha("<fewer defaults>" + <a few special options>)` should still lead to pretty unique hashes in comparison to `("<constant set of defaults>" + "<a few special options>")`, shouldn't it?

Sorry @dpiparo, but I have to say it again 😄:
> Maybe we should invest a few extra minutes of compile time to build from scratch with a big ccache. This would make all *previous-build-folder-related* problems go away and simplify the CI scripts.

hahnjo:
> > This reverts commit [2f2a29b](https://github.com/root-project/root/commit/2f2a29ba6384fe2f26716c385769cbcdd5b60df8). It breaks the option hashing in our CI script.
> 
> While I agree that having removed these options changes the hash, does it also lead to clashes? `sha("<fewer defaults>" + <a few special options>)` should still lead to pretty unique hashes in comparison to `("<constant set of defaults>" + "<a few special options>")`, shouldn't it?

I'm not worried about clashes between platforms, but the problem is that changing a default in `cmake/modules/RootBuildOptions.cmake` will not change the hash and our CI will attempt an incremental build. That's why (as far as I understand), explicit spelling of all options was added in the first place.

hageboeck:
> I'm not worried about clashes between platforms, but the problem is that changing a default in `cmake/modules/RootBuildOptions.cmake` will not change the hash and our CI will attempt an incremental build. That's why (as far as I understand), explicit spelling of all options was added in the first place.

This makes total sense, but you also named a possible solution: 🙂
Hash the options string + `cmake/modules/RootBuildOptions.cmake`

(Or we don't do incremental builds. 😉)

guitargeek:
> I'm not worried about clashes between platforms, but the problem is that changing a default in cmake/modules/RootBuildOptions.cmake will not change the hash and our CI will attempt an incremental build. That's why (as far as I understand), explicit spelling of all options was added in the first place.

This problem is that when someone changes `RootBuildOptions.cmake`, the risk that they forget to update `globals.txt` is high and as a result, the CI config uncontrollably diverges from the default config over time.

Stephans proposed solutions both make sense to me! The drawback of hashing `RootBuildOptions.cmake` is that we get false positives when someone just changes a typo in that file or something, but that's rare and most changes in that file should better result in a clean rebuild anyway.

guitargeek:
Coincidentally, the failues in `ubuntu2510` show us another potential problem with the incremental builds, because of a glibc update:
```txt
2025-08-22T06:22:41.0697191Z   Possible C++ standard library mismatch, compiled with __GLIBCXX__ '20250804'
2025-08-22T06:22:41.0697518Z   Extraction of runtime standard library version was: '20250812'
```

ferdymercury:
> Hash the options string + `cmake/modules/RootBuildOptions.cmake`

cmake has an option to print all current options, so no need to concatenate maybe?
https://stackoverflow.com/questions/16851084/how-to-list-all-cmake-build-options-and-their-default-values

That way it will not be sensitive to eg just adding a newline to that file

hahnjo:
> > Hash the options string + `cmake/modules/RootBuildOptions.cmake`
> 
> cmake has an option to print all current options, so no need to concatenate maybe?

We need the hash before actually running `cmake`, to download the "right" binary artifacts in PR builds.

ferdymercury:
> We need the hash before actually running `cmake`, to download the "right" binary artifacts in PR builds.

Couldn't you do:

```
mkdir -p /tmp/aux && cd /tmp/aux
cmake /path/to/src -Dwhatever-os-options
cmake -L > /tmp/currentoptions.txt
rm -r /tmp/aux
Download hash matching currentoptions.txt
```

Or is that adding too much time delay due to the extra temporary configure step?

hahnjo:
As far as I can tell, this has still not been handled, which leaves the CI in a state where you need some expert knowledge in order to avoid bad surprises. Unless there is high-priority effort going into this, I would urge to revert the change and at least get back to a "known" state. @dpiparo

guitargeek:
So do I understand correctly that you are fine with the solution proposed by @hageboeck?

> Hash the options string + cmake/modules/RootBuildOptions.cmake


I was waiting if you had any comments on the suggested action before proceeding to open a PR with it.

hahnjo:
I haven't thought it through. I'd also need to see it in code and we must be sure that there are no other places that have funny side consequences on the options. Generally, I would still prefer to test a fixed set of options, I think that is a useful way for the CI. In the absence of code ready *now*, I still think we should revert.

PR #19722, opened 2025-08-22
[main] replace rootbrowse.py with C++ version

silverweed:
Similarly to https://github.com/root-project/root/pull/19305 with rootls, this PR replaces `rootbrowse.py` with a native version. 
Since rootbrowse is just a thin wrapper over TBrowser, this change is much smaller than for rootls and should be quite straightforward.

**NOTE**: the new rootbrowse is currently untested on Windows and Mac, so before merging this I would much appreciate if someone could help me with that.

## Checklist:

- [x] tested changes locally


main/src/rootbrowse.cxx
```diff
@@ -0,0 +1,130 @@
...
+static const char *const kLongHelp = R"(
+Open a ROOT file in a TBrowser
+
+positional arguments:
+  FILE           Input file
+
+options:
+  -h, --help     show this help message and exit
+  -w, --web WEB  Configure webdisplay. For all possible values, see:
+                 https://root.cern/doc/v636/classTROOT.html#a1749472696545b76a6b8e79769e7e773
```
silverweed: Not sure if there is a way to point to the latest doc consistently...

main/src/rootbrowse.cxx
```diff
@@ -0,0 +1,130 @@
...
+///
+/// Command line tool to open a ROOT file on a TBrowser
+///
+/// \author Giacomo Parolini <giacomo.parolini@cern.ch>
+/// \date 2025-08-21
+#include <cstring>
+#include <iostream>
+#include <thread>
+#include <chrono>
+#include <memory>
```
jblomer: Standard includes come last; also order alphabetically

main/src/rootbrowse.cxx
```diff
@@ -0,0 +1,130 @@
...
+
+#include <ROOT/RLogger.hxx>
+
+#include <TError.h>
+#include <TFile.h>
+#include <TROOT.h>
+#include <TSystem.h>
+#include <TBrowser.h>
+#include <TApplication.h>
```
jblomer: Order alphabetically

main/src/rootbrowse.cxx
jblomer: That would be great! Or this string needs to be updated as part of the release procedure

main/src/rootbrowse.cxx
```diff
@@ -0,0 +1,130 @@
...
+Open a ROOT file in a TBrowser
+
+positional arguments:
+  FILE           Input file
+
+options:
+  -h, --help     show this help message and exit
+  -w, --web WEB  Configure webdisplay. For all possible values, see:
+                 https://root.cern/doc/v636/classTROOT.html#a1749472696545b76a6b8e79769e7e773
+  -wf, --webOff  Invoke the normal TBrowser (not the web version)
```
jblomer:
```suggestion
  -wf, --webOff  Invoke the classic TBrowser (not the web version)
```

main/src/rootbrowse.cxx
```diff
@@ -0,0 +1,130 @@
...
+   }
+
+   return outArgs;
+}
+
+int main(int argc, char **argv)
+{
+   auto args = ParseArgs(const_cast<const char **>(argv) + 1, argc - 1);
+   if (args.fPrintHelp != RootBrowseArgs::EPrintUsage::kNo) {
+      std::cerr << "usage: rootbrowse [-w WEB|-wf] <file.root>\n";
```
jblomer: Maybe have another static string next to `kLongHelp`, to see both messages together.

main/src/rootbrowse.cxx
```diff
@@ -0,0 +1,130 @@
...
+   }
+
+   // NOTE: we need to instantiate TApplication ourselves, otherwise TBrowser
+   // will create a batch application that cannot show graphics.
+   TApplication app("rootbrowse", nullptr, nullptr);
+
+   auto browser = std::make_unique<TBrowser>();
+   std::cout << "Press ctrl+c to exit.\n";
+   while (!gROOT->IsInterrupted() && !gSystem->ProcessEvents()) {
+      std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
jblomer, line 166: So there is no other way but busy waiting to run the event loop?

silverweed: I didn't investigate much, but this is basically what rootbrowse.py was doing. Maybe @bellenot @couet or @linev know a better way, but I'd personally consider it only if it's not significantly more complicated than this, since this is not exactly hogging the CPU anyway

main/src/rootbrowse.cxx
```diff
@@ -0,0 +1,131 @@
...
+static const char *const kLongHelp = R"(
+Open a ROOT file in a TBrowser
+
+positional arguments:
+  FILE           Input file
+
+options:
+  -h, --help     show this help message and exit
+  -w, --web WEB  Configure webdisplay. For all possible values, see:
+                 https://root.cern/doc/v636/classTROOT.html#a1749472696545b76a6b8e79769e7e773
```
ferdymercury:
```suggestion
                 https://root.cern/doc/master/classTROOT.html#a1749472696545b76a6b8e79769e7e773
```

silverweed: This will indeed update automatically but it is also ahead of the latest release, so not really what I want

ferdymercury: A workaround could be to define 

 https://root.cern/doc/latest-stable/
 
 in the webserver as a redirect to the current release.
 
 Then we would not need to modify the cpp code every time, just the redirect.

silverweed: @dpiparo do you think we could do this? (assuming we don't have such a mechanism already)

main/src/rootbrowse.cxx
pcanal:
```
https://root.cern/doc/master/classTROOT.html#a1749472696545b76a6b8e79769e7e773
```
works (maybe, I am not sure how stable the hash is).
@dpiparo It would also be nice if
```
https://root.cern/doc/latest-stable/classTROOT.html#a1749472696545b76a6b8e79769e7e773
```
worked ...

PS. Apologies to @ferdymercury who already proposed this.

main/CMakeLists.txt
```diff
@@ -111,6 +111,9 @@ if(WIN32)
   set_property(TARGET rootcling APPEND_STRING PROPERTY LINK_FLAGS " -STACK:4000000")
 endif()
 
+if(gui)
```
vepadulano: There is no target `gui` in our build system

vepadulano (changes requested):
Reporting from my MacOS 15 machine:

* calling `rootbrowse` by default starts the webgui, which just hangs and does not produce anything on screen:

```
08:56:03 (rootdev) vpadulan@vpadulan-macbook rootproject → rootbrowse

ROOT comes with a web-based browser, which is now being started.
Revert to TBrowser by setting "Browser.Name: TRootBrowser" in rootrc file or
by starting "root --web=off"
Find more info on https://root.cern/for_developers/root7/#rbrowser
Info in <THttpEngine::Create>: Starting HTTP server on port 127.0.0.1:8846
Press ctrl+c to exit.
```

* setting "Browser.Name: TRootBrowser" in rootrc does not work, `rootbrowse` does not pick up the option and still tries to start the web graphics


@linev the fact that the web graphics hang on MacOS is troublesome

linev:
> @linev the fact that the web graphics hang on MacOS is troublesome

@vepadulano I see no problem on my Mac with master branch.

```
root --web=safari -e 'new TBrowser'  
```

or  
  
```
root --web=safari -e 'new TBrowser'  
```

It is troublesome that simple commands is not know to everybody.

main/src/rootbrowse.cxx
silverweed: With Danilo we decided we will do this, however we need a good way to make it obvious which version a user is looking at, and not just "latest-stable". This will probably need a little work on the website js - I will do it as soon as I manage.

vepadulano:
What command do you mean specifically? I'm just testing `rootbrowse` as it is the main goal of this PR

main/src/rootbrowse.cxx
ferdymercury: a more pinpointed workaround would be to define the message as:

"https://root.cern/doc/" + R__VERSION_MAJOR % 2 == 0? "v"#R__VERSION_MAJOR#R_VERSION_MINOR : "master" + "/classTROOT.html#a1749472696545b76a6b8e79769e7e773"

which would then point at the "right" version of the rootbrowse you are reunning. Since, if you are running 6.36 and stable is 6.38, you don't really want to see the docu of 6.38, assuming that the new flags have been added that are not in 6.36 which may lead to confusion.

main/src/rootbrowse.cxx
ferdymercury: Why not doing it on the c++ side or even CMake side? since R__VERSION_MAJOR is known, we can decide whether the help prints:
v636 or v634 or master, etc.
For non-even dev versions, either "master" or round-up to even number.
This could be a generic CMake or C++ function that would be useful for everybody, not just rootbrowse.

main/CMakeLists.txt
ferdymercury:
```suggestion
if(TARGET Gui)
```

main/src/rootbrowse.cxx
```diff
@@ -0,0 +1,131 @@
...
+      if (!file || file->IsZombie()) {
+         R__LOG_WARNING(RootBrowseLog()) << "File " << args.fFileName << " does not exist or is unreadable.";
+      }
+      gErrorIgnoreLevel = kUnset;
+   }
+
+   // NOTE: we need to instantiate TApplication ourselves, otherwise TBrowser
+   // will create a batch application that cannot show graphics.
+   TApplication app("rootbrowse", nullptr, nullptr);
+
```
ferdymercury: General comment on why I never use rootbrowse:
You sometimes come to a point where you would like to type something in the terminal and you cannot.

My question: wouldn't it be easier to just call, as an alias:
root -l %arg1 --web=%arg2 -e 'auto br = new TBrowser;'

rather than creating a TApplication ?

Or, alternatively, create a TRint instead of a TApplication, as in rmain.cxx

linev:
@vepadulano 

Old `rootbrowse` based on python. And ROOT graphics - classical one and web-based - does not work reliably on mac with python.

There is PR which provides reasonable solution for python on all platforms - including Windows and Mac.
PR is https://github.com/root-project/root/pull/17587

It is pending since January.

silverweed:
@linev Vincenzo was referring to this version of rootbrowse, so Python is not involved. We later found that the web graphics was not actually "hanging", but rather opening a window in another desktop without moving automatically to it - this is obviously better than hanging, but still not a great experience for a user. Is it expected to do that or should the web window also get focused automatically (like the classic graphics do)?

main/src/rootbrowse.cxx
silverweed: Maybe this would indeed be a better option. Begs the question why it wasn't done like that so far though - if there are reasons for it I am not aware of them...

main/CMakeLists.txt
silverweed: Actually we don't even need to check for it - it's always built

main/src/rootbrowse.cxx
silverweed: This could be also a good solution indeed. Now I'm wondering if the hash after `classTROOT.html` is stable throughout versions though.

main/CMakeLists.txt
ferdymercury: even with -Dminimal=ON?

main/src/rootbrowse.cxx
silverweed: I talked with @bellenot and there is a simple solution for the classical graphics, but as far as I understand it doesn't work out of the box for the webgraphics, so for now I'm keeping it as-is.

main/src/rootbrowse.cxx
ferdymercury: > classTROOT.html#a1749472696545b76a6b8e79769e7e773

@albert-github do you know how stable are the doxygen-generated hashes? Do they change between doxygen versions? Or only if one changes the signature of the corresponding C++ function? Thks in advance.

main/CMakeLists.txt
ferdymercury: Oh, I see, it's https://github.com/root-project/root/pull/19381

main/CMakeLists.txt
```diff
@@ -111,6 +111,7 @@ if(WIN32)
   set_property(TARGET rootcling APPEND_STRING PROPERTY LINK_FLAGS " -STACK:4000000")
 endif()
 
+ROOT_EXECUTABLE(rootbrowse src/rootbrowse.cxx LIBRARIES RIO Core Rint Gui)
```
ferdymercury, line 114:
```suggestion
if(NOT minimal)
   ROOT_EXECUTABLE(rootbrowse src/rootbrowse.cxx LIBRARIES RIO Core Rint Gui)
endif()
```
Since GUI is anyway broken with minimal=ON.
See https://github.com/root-project/root/pull/19381#issuecomment-3074078350

linev:
> Is it expected to do that or should the web window also get focused automatically (like the classic graphics do)?

I do not have control over desktop applications when starting external browser via the shell functionality.
Each desktop manager like KDE or Gnome or Windows or Mac have its own methods to control this.
The only exception is use of CEF or Qt6 WebEngine - but they sometime have other problems and normally not present on the system

main/CMakeLists.txt
silverweed: I'd rather not add complexity to the build system if not necessary - in minimal `rootbrowse` will be compiled and just fail gracefully to open.

main/src/rootbrowse.cxx
albert-github: They should be stable between doxygen versions and should only change in case of a change of signature of the corresponding C++ function.

main/CMakeLists.txt
ferdymercury: Sounds reasonable. But if you merge later on https://github.com/root-project/root/pull/19381 users will get a build error when setting -Dminimal=ON.

main/src/rootbrowse.cxx
silverweed: I still don't hate the idea of having a `latest-stable` link, regardless of what we decide to do with this PR

main/CMakeLists.txt
silverweed: Sure, if that gets merged I'll take care of it - should be hard to miss, since I always use minimal in one of my build trees :)

main/src/rootbrowse.cxx
silverweed: https://github.com/root-project/web/pull/1146

vepadulano (approved):
Tested on my MacOS ARM laptop, works with both traditional graphics and web graphics, thanks!

PR #19724, opened 2025-08-22
[RDF] Prepare RActionSnapshot for variations.

hageboeck:
Convert the fPreviousNode pointer into a vector of pointers. For now, it will only have a single entry, but this will allow for extending RActionSnapshot to systematic variations later.

This is part 3 of the work in #19725. Feel free to merge it if it gets approved and the CI is green.


vepadulano (approved):
Thank you! I will wait for the CI to rerun before merging

PR #19725, opened 2025-08-22
[RDF] Add snapshot with variations

hageboeck:
**Note**:
This PR is split into several smaller parts to simplify the review. Many of the commits will disappear once it gets rebased.
When this happens, the description will be extended and it will come out of draft mode.


tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -0,0 +1,435 @@
...
+   options.fIncludeVariations = true;
+
+   auto cuts = [](int x, int y) { return x % 2 == 0 && y % 3 == 0; };
+
+   ROOT::RDataFrame(N)
+      .Define("x", [](ULong64_t e) -> int { return e; }, {"rdfentry_"})
+      .Define("entry", [](ULong64_t e) { return e; }, {"rdfentry_"})
+      .Vary(
+         "x",
+         [NSystematics](int x) {
```
ferdymercury:
```suggestion
         [](int x) {
```
to avoid mac26 warning?

hageboeck: Would be great, but then it wouldn't compile on Windows 😅 

The only solution is to use `[=]` it seems, so I'll push that later.

tree/dataframe/inc/ROOT/RDF/RActionSnapshot.hxx
```diff
@@ -65,6 +67,27 @@ public:
       fIsDefine.reserve(nColumns);
       for (auto i = 0u; i < nColumns; ++i)
          fIsDefine.push_back(colRegister.IsDefineOrAlias(columns[i]));
+
+      if (const auto &variations = GetVariations(); !variations.empty()) {
```
vepadulano: This is running irrespective of the option set by the user?

tree/dataframe/inc/ROOT/RDF/RActionSnapshot.hxx
```diff
@@ -89,6 +112,26 @@ public:
...
+         // that get passed to the helpers
+         auto const &variations = GetVariations();
+         for (unsigned int variationIndex = 0; variationIndex < variations.size(); ++variationIndex) {
+            auto const &readers =
+               GetUntypedColumnReaders(slot, r, RActionBase::GetColRegister(), *fLoopManager,
+                                       RActionBase::GetColumnNames(), fColTypeIDs, variations[variationIndex]);
+            for (unsigned int i = 0; i < readers.size(); ++i) {
+               if (fValues[slot][i] != readers[i]) {
+                  fValues[slot].push_back(readers[i]);
+                  fHelper.RegisterVariedColumn(slot, i, i, 0, "nominal");
```
vepadulano: Here it looks like nominal is being registered many times, can you elaborate on the logic?

tree/dataframe/inc/ROOT/RDF/RTreeColumnReader.hxx
```diff
@@ -108,6 +110,23 @@ private:
...
+   std::unique_ptr<ROOT::Detail::RDF::RColumnReaderBase> fValueReader;
+   std::unique_ptr<TTreeReaderValue<uint64_t>> fTreeValueMask;
+   unsigned int fMaskIndex = 0;
+
+   void *GetImpl(Long64_t) override;
+
+public:
+   RMaskedColumnReader(TTreeReader &r, std::unique_ptr<ROOT::Detail::RDF::RColumnReaderBase> valueReader,
+                       std::string_view maskName, unsigned int maskIndex);
+   virtual ~RMaskedColumnReader();
```
vepadulano: Is this needed?

tree/dataframe/inc/ROOT/RDF/SnapshotHelpers.hxx
```diff
@@ -236,6 +265,46 @@ public:
...
+public:
+   SnapshotHelperWithVariations(std::string_view filename, std::string_view dirname, std::string_view treename,
+                                const ColumnNames_t & /*vbnames*/, const ColumnNames_t &bnames,
+                                const RSnapshotOptions &options, std::vector<bool> && /*isDefine*/,
+                                ROOT::Detail::RDF::RLoopManager *outputLoopMgr,
+                                ROOT::Detail::RDF::RLoopManager *inputLoopMgr,
+                                const std::vector<const std::type_info *> &colTypeIDs);
+
+   SnapshotHelperWithVariations(SnapshotHelperWithVariations &&) = default;
+   ~SnapshotHelperWithVariations();
```
vepadulano: I suggest fully implementing rule of five

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -930,3 +995,244 @@ ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::MakeNew(void *newName)
...
+   };
+   std::vector<Bitmask> fBitMasks;
+
+   std::unordered_map<std::string, unsigned int> fBranchToVariationMapping;
+   // The corresponding ROOT dictionary is declared in core/clingutils/src
+   std::unordered_map<std::string, std::pair<std::string, unsigned int>> fBranchToBitmaskMapping;
+   unsigned int fNBits = 0;
+
+   SnapshotOutputWriter(TFile *file, TTree *tree) : fFile{file}, fTree{tree} {}
+   ~SnapshotOutputWriter()
```
vepadulano, line 1028: Here another opportunity to implement rule of five

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -930,3 +995,244 @@ ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::MakeNew(void *newName)
...
+   {
+      if (!fBranchToBitmaskMapping.empty()) {
+         fFile->WriteObject(&fBranchToBitmaskMapping,
+                            (std::string{"R_rdf_branchToBitmaskMapping_"} + fTree->GetName()).c_str());
+      }
```
vepadulano, line 1033: Sorry I'm getting lost here, isn't the mapping another branch of the output tree? Here instead it is an object in the output file

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -930,3 +995,244 @@ ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::MakeNew(void *newName)
...
+         fTree.reset();
+         fFile.reset();
+
+         if (fOutputLoopManager)
+            fOutputLoopManager->SetDataSource(std::make_unique<ROOT::Internal::RDF::RTTreeDS>(tree, file));
+      }
+   }
+
+   /// Register a branch and corresponding systematic uncertainty. The index returned is the global index of this
+   /// systematic.
```
vepadulano: What is the "index returned" here? Maybe the comment does not apply anymore?

vepadulano (commented):
Thank you for such a great effort! I have a first round of comments for discussion.

tree/dataframe/src/RDFSnapshotHelpers.cxx
hageboeck: It's indeed both. For every 64 variations, we add one branch to the output tree. But to know which variation is which bit in which branch, we write a map that maps branchName --> (branchWithBitmask, index) pairs.
The masked column readers will check for this map, and start the masking logic if their column shows up in this map.

tree/dataframe/src/RDFSnapshotHelpers.cxx
hageboeck: Indeed, I will update this comment.

tree/dataframe/inc/ROOT/RDF/RActionSnapshot.hxx
hageboeck: It's correct, but maybe not obvious:
There are three column types:
1. Nominal and always visible, i.e. no variations 
2. Nominal and maybe masked, i.e., this column has variations
3. Varied

The if in line 125 determines if the column is type 1. In that case, nothing needs to be done.
If the column reader of the variation differs from the original, the column undergoes variations. We therefore have to register both type 2 and type 3 to the output helper, so both can be masked out if needed.

I'll add a few comments, too.

tree/dataframe/inc/ROOT/RDF/RTreeColumnReader.hxx
hageboeck: It's here to outline the destructor, so the header can get away with forward decls.

hageboeck: In the .cxx, it's `= default`.

tree/dataframe/inc/ROOT/RDF/SnapshotHelpers.hxx
hageboeck: The destructor is defaulted and outlined, so the compiler-generated versions will be OK. I'll check if the destructor can be defaulted in the header, though, to make it visible.

**Update**: Now destructor and move are defaulted in the header; copy disabled until proven necessary. Must have been a remnant from an earlier version of the class.

tree/dataframe/src/RDFSnapshotHelpers.cxx
hageboeck: That's a good point. The default move constructor wouldn't have done any cleanup of the moved-from objects, so I disabled copy/move until proven to be necessary.

tree/dataframe/inc/ROOT/RDF/RActionSnapshot.hxx
hageboeck: It is. It doesn't affect correctness, but it's unnecessary.

hageboeck:
@vepadulano Changes based on the discussion above now pushed as `!fixup` commits if you want to have a look.

tree/dataframe/inc/ROOT/RDF/RTreeColumnReader.hxx
vepadulano: I see now what you mean, but then it should be
```suggestion
   ~RMaskedColumnReader() override;
```
And ideally also rule of five should be implemented, but not necessarily

tree/dataframe/inc/ROOT/RDFHelpers.hxx
```diff
@@ -222,6 +222,10 @@ namespace Experimental {
 template <typename T>
 RResultMap<T> VariationsFor(RResultPtr<T> resPtr)
 {
+   using SnapshotResult_t = ROOT::RDF::RInterface<ROOT::Detail::RDF::RLoopManager, void>;
+   static_assert(!std::is_same_v<T, SnapshotResult_t>,
+                 "Snapshot with variations only can be enabled via RSnapshotOptions.");
```
vepadulano:
```suggestion
                 "Snapshot with variations can only be enabled via RSnapshotOptions.");
```

tree/dataframe/src/RTreeColumnReader.cxx
```diff
@@ -105,3 +107,23 @@ ROOT::Internal::RDF::RTreeUntypedArrayColumnReader::RTreeUntypedArrayColumnReade
...
+ROOT::Internal::RDF::RMaskedColumnReader::~RMaskedColumnReader() = default;
+
+void *ROOT::Internal::RDF::RMaskedColumnReader::GetImpl(Long64_t event)
+{
+   const std::bitset<64> mask{*fTreeValueMask->Get()};
+   if (mask.test(fMaskIndex) == false)
+      return nullptr;
+
+   return fValueReader->TryGet<void>(event);
+}
```
vepadulano: missing line at end of file

tree/dataframe/test/CMakeLists.txt
```diff
@@ -57,6 +57,8 @@ ROOT_GENERATE_DICTIONARY(DummyDict ${CMAKE_CURRENT_SOURCE_DIR}/DummyHeader.hxx
                          MODULE dataframe_snapshot_emptyoutput LINKDEF DummyHeaderLinkDef.hxx OPTIONS -inlineInputHeader
                          DEPENDENCIES ROOTVecOps GenVector)
 ROOT_ADD_GTEST(dataframe_snapshot_ntuple dataframe_snapshot_ntuple.cxx LIBRARIES ROOTDataFrame ROOTNTuple NTupleStruct)
+ROOT_ADD_GTEST(dataframe_snapshotWithVariations dataframe_snapshotWithVariations.cxx TwoFloatsDict.cxx LIBRARIES ROOTDataFrame)
+# add_dependencies(dataframe_snapshotWithVariations TwoFloatsDict)
```
vepadulano: Spurious leftover comment?

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -0,0 +1,435 @@
...
+            EXPECT_EQ(tf_Up->b, i + 1.f);
+            EXPECT_EQ(tf_Down->a, i);
+            EXPECT_EQ(tf_Down->b, i - 1.f);
+         }
+      }
+   }
+
+   if (!HasFailure())
+      gSystem->Unlink(filename);
+}
```
vepadulano: missing line at end of file

tree/dataframe/inc/ROOT/RDF/RActionSnapshot.hxx
```diff
@@ -119,9 +167,27 @@ public:
...
-      if (fPrevNodes.front()->CheckFilters(slot, entry))
-         CallExec(slot, entry);
+      if constexpr (std::is_same_v<Helper, SnapshotHelperWithVariations>) {
+         // check if entry passes all filters
+         std::vector<bool> filterPassed(fPrevNodes.size(), false);
+         for (unsigned int variation = 0; variation < fPrevNodes.size(); ++variation) {
+            filterPassed[variation] = fPrevNodes[variation]->CheckFilters(slot, entry);
+         }
+
+         if (std::any_of(filterPassed.begin(), filterPassed.end(), [](bool val) { return val; })) {
```
vepadulano, line 179: Maybe add a note here: in principle, this is where we could inject different behaviours for this feature IIUC:
* write all events that pass at least one filter among the nominal+varied columns (i.e. what happens right now)
* write only events where all filters pass for all columns
* write only events where no filters pass (needed at all?)

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -364,6 +366,69 @@ void SetBranchesHelper(TTree *inputTree, TTree &outputTree,
...
+void ROOT::Internal::RDF::RBranchData::ClearBranchContents()
+{
+   if (!fOutputBranch)
+      return;
+
+   if (auto fundamental = std::get_if<FundamentalType>(&fTypeData); fundamental) {
+      fundamental->fBytes.fill(std::byte{0});
+   } else {
+      // TTree expects pointer to pointer, to figure out who allocates the object:
+      fOutputBranch->SetAddress(EmptyInstance(/*pointerToPointer=*/true));
```
vepadulano, line 428: Don't we need here to bind `fTypeData` to `EmptyInstance` before calling `SetAddress`?

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -930,3 +995,260 @@ ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::MakeNew(void *newName)
...
+      std::unique_ptr<uint64_t> branchBuffer{new uint64_t{}};
+   };
+   std::vector<Bitmask> fBitMasks;
+
+   std::unordered_map<std::string, unsigned int> fBranchToVariationMapping;
+   // The corresponding ROOT dictionary is declared in core/clingutils/src
+   std::unordered_map<std::string, std::pair<std::string, unsigned int>> fBranchToBitmaskMapping;
+   unsigned int fNBits = 0;
+
+   SnapshotOutputWriter(TFile *file, TTree *tree) : fFile{file}, fTree{tree} {}
```
vepadulano: nitpick
```suggestion
   SnapshotOutputWriter(TFile *file, TTree *tree) : fFile{file}, fTree{tree} {assert(file != nullptr); assert(tree != nullptr);}
```

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -930,3 +995,260 @@ ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::MakeNew(void *newName)
...
+
+   /// Register a branch and corresponding systematic uncertainty.
+   /// This will create an entry in the mapping from branch names to bitmasks, so the corresponding
+   /// column can be masked if it doesn't contain valid entries. This mapping is written next to the
+   /// tree into the output file.
+   void RegisterBranch(std::string const &branchName, unsigned int variationIndex)
+   {
+      if (auto it = fBranchToVariationMapping.find(branchName); it != fBranchToVariationMapping.end()) {
+         if (variationIndex != it->second) {
+            throw std::logic_error("Branch " + branchName + " is being registered with different variation index.");
```
vepadulano: Maybe add the following (or maybe consider also writing the full name of the variation in the error message?)
```suggestion
            throw std::logic_error("Branch " + branchName + " is being registered with different variation index than the expected one: " + std::to_string(variationIndex));
```

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -930,3 +995,260 @@ ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::MakeNew(void *newName)
...
+ROOT::Internal::RDF::SnapshotHelperWithVariations::SnapshotHelperWithVariations(
+   std::string_view filename, std::string_view dirname, std::string_view treename, const ColumnNames_t &vbnames,
+   const ColumnNames_t &bnames, const RSnapshotOptions &options, std::vector<bool> &&isDefine,
+   ROOT::Detail::RDF::RLoopManager *outputLoopMgr, ROOT::Detail::RDF::RLoopManager *inputLoopMgr,
+   const std::vector<const std::type_info *> &colTypeIDs)
+   : fOptions(options), fInputLoopManager{inputLoopMgr}, fOutputLoopManager{outputLoopMgr}
+{
+   EnsureValidSnapshotTTreeOutput(fOptions, std::string(treename), std::string(filename));
+
+   TFile::TContext fileCtxt;
```
vepadulano: I usually see `TDirectory::TContext` being used in such cases, any reason to use `TFile::TContext` here?

tree/dataframe/src/RDFSnapshotHelpers.cxx
vepadulano: I now see that you have a `if (!fOutputHandle->fFile)` condition a few lines below to check at least for the validity of the input TFile. I would consider moving these checks in the constructor of `SnapshotOutputWriter` just to have connected logic closer together

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -930,3 +995,260 @@ ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::MakeNew(void *newName)
...
+   fOutputHandle = std::make_shared<SnapshotOutputWriter>(
+      TFile::Open(filename.data(), fOptions.fMode.c_str(), /*ftitle=*/"",
+                  ROOT::CompressionSettings(fOptions.fCompressionAlgorithm, fOptions.fCompressionLevel)),
+      nullptr);
+   if (!fOutputHandle->fFile)
+      throw std::runtime_error(std::string{"Snapshot: could not create output file "} + std::string{filename});
```
vepadulano, line 1140: See my other comment in the constructor of `SnapshotOutputWriter`

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -930,3 +995,260 @@ ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::MakeNew(void *newName)
...
+   auto outputBranchNames = ReplaceDotWithUnderscore(bnames);
+
+   fBranchData.reserve(vbnames.size());
+   for (unsigned int i = 0; i < vbnames.size(); ++i) {
+      fOutputHandle->RegisterBranch(outputBranchNames[i], 0);
+      fBranchData.emplace_back(vbnames[i], outputBranchNames[i], isDefine[i], colTypeIDs[i]);
+   }
+}
+
+// SnapshotHelperWithVariations::~SnapshotHelperWithVariations() = default;
```
vepadulano: I guess this line can be removed since it's commented now

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -930,3 +995,260 @@ ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::MakeNew(void *newName)
...
+/// If a nominal column is registered here, it is written without changes, but it means that it will be masked
+/// in case its selection cuts don't pass.
+/// \param slot Task ID for MT runs.
+/// \param columnIndex Index where the data of this column will be passed into the helper.
+/// \param originalColumnIndex If the column being registered is a variation of a "nominal" column, this designates the
+/// original.
+///         Properties such as name and output type are cloned from the original.
+/// \param variationName The variation that this column belongs to. If "nominal" is used, this column is considered as
+/// the original.
+void SnapshotHelperWithVariations::RegisterVariedColumn(unsigned int /*slot*/, unsigned int columnIndex,
```
vepadulano: This and other `SnapshotHelperWithVariations` methods are defined in this implementation file within a `namespace ROOT::Internal::RDF {` scope. I believe it's better to define them outside, so that the explicit spelling `ROOT::Internal::RDF::SnapshotHelperWithVariations` must be used everywhere (like it's done in the constructor definition). Clearly `SnapshotOutputWriter` is exempt because it's defined directly within this file.

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -0,0 +1,435 @@
+#include <ROOT/RDataFrame.hxx>
```
vepadulano, line 1: Offline we discussed about adding one more test: Snapshot with variations of two different TTrees in the same file, to check that the bitmask mappings are not interefering with each other.

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -0,0 +1,435 @@
...
+#include <TLeaf.h>
+#include <TTree.h>
+#include <TSystem.h>
+
+#include <gtest/gtest.h>
+
+#include <bitset>
+#include <numeric>
+
+constexpr bool verbose = false;
```
vepadulano, line 18: How can this be controlled? Is it something useful only when changing the file locally?

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -0,0 +1,435 @@
...
+            EXPECT_EQ(tf_Up->a, i);
+            EXPECT_EQ(tf_Up->b, i + 1.f);
+            EXPECT_EQ(tf_Down->a, i);
+            EXPECT_EQ(tf_Down->b, i - 1.f);
+         }
+      }
+   }
+
+   if (!HasFailure())
+      gSystem->Unlink(filename);
```
vepadulano: Here and in all similar `Unlink`/`std::remove` calls, I would propose to have a RAII structure to manage file cleanup to be more exception-safe

tree/dataframe/inc/ROOT/RDF/RTreeColumnReader.hxx
hageboeck: We can add the override, indeed (*done*). As long as the destructor is defaulted, we are good with the rule of five, because it states "define or =delete".

tree/dataframe/src/RDFSnapshotHelpers.cxx
hageboeck: Inside `EmtpyInstance()`, the type variant gets checked. An empty instance is created on the fly if needed, so `SetBranchAddress` gets bound to the same object no matter how often you call this.

For fundamental types, however, (re-)binding the branch address again and again is just a waste of CPU cycles. It gets done once in every `InitTask()`, and isn't touched again like for branches that don't participate in variations.

tree/dataframe/src/RDFSnapshotHelpers.cxx
hageboeck: Ups, I misremembered the name. 😅
They happen to be the same class due to the inheritance, but there is no reason to refer to the class of TDirectory through TFile ...

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
hageboeck: It's activating code that will print the trees, and this is I believe useful if anybody would like to debug the tests. I would like to leave it there (but disabled) so that future maintainers don't need to rewrite the code.
The constexpr is the best way to do it, because we get syntax checking, but no actual code is emitted. It's similar to what the kernel does with `if (false) {`.

tree/dataframe/src/RDFSnapshotHelpers.cxx
hageboeck: I added the suggested assertion, but for release builds, we will need to keep this exception. Furthermore, the OutputWriter doesn't know the filename if file creation fails, so the exception needs to be where we open the file.

tree/dataframe/inc/ROOT/RDF/RTreeColumnReader.hxx
vepadulano: Rule of five states that a class which declares (which includes `=default` and `=delete`) anyone of destructor, copy constructor, move constructor, copy assignment, move assignment should also declare all others. Again, that's a guideline, so I won't block this PR because the rule is not implemented.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
hageboeck: :+1: See new commit.

tree/dataframe/inc/ROOT/RDF/RTreeColumnReader.hxx
```diff
@@ -108,6 +110,23 @@ private:
...
+   std::unique_ptr<ROOT::Detail::RDF::RColumnReaderBase> fValueReader;
+   std::unique_ptr<TTreeReaderValue<uint64_t>> fTreeValueMask;
+   unsigned int fMaskIndex = 0;
+
+   void *GetImpl(Long64_t) override;
+
+public:
+   RMaskedColumnReader(TTreeReader &r, std::unique_ptr<ROOT::Detail::RDF::RColumnReaderBase> valueReader,
+                       std::string_view maskName, unsigned int maskIndex);
+   virtual ~RMaskedColumnReader() override;
```
vepadulano: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final


```suggestion
   ~RMaskedColumnReader() override;
```

vepadulano (approved):
Fantastic, thank you so much!

rbarrue:
Hi @hageboeck thank you very much for this work. Any idea on the timeline for this PR to get merged ?

hageboeck:
> Hi @hageboeck thank you very much for this work. Any idea on the timeline for this PR to get merged ?

Yes, today! We were battling with test failures on Windows, but the cause is now understood.

rbarrue:
Hi @hageboeck a question: from what I got, this only saves the quantities that are directly varied. It doesn't vary downstream quantities that depend on the directly varied quantities right ?

This would require a Snapshot version of VariationsFor(), that is not yet implemented, right ?

hageboeck:
> Hi @hageboeck a question: from what I got, this only saves the quantities that are directly varied. It doesn't vary downstream quantities that depend on the directly varied quantities right ?
> 
> This would require a Snapshot version of VariationsFor(), that is not yet implemented, right ?

It varies also downstream defines, see here:
https://root.cern.ch/doc/master/classROOT_1_1RDataFrame.html#systematics

PR #19726, opened 2025-08-22
[RDF] Prepare for snapshot with variations 2

hageboeck:
Next to cleaning up a few white spaces and moving a docstring to the right place, two changes are necessary to implement snapshot with variations. These affect column readers for systematic variations:
Previously, each variation got its own column reader, even if the underlying `Define` was the same (because a column isn't affected by variations). In this case, SnapshotWithVariations would write the column again and again, since it doesn't have a way to check if the column is affected by variations.

Here, column readers are reused if a column is identical to nominal, and this will be detected by snapshot to write the column only once. This goes both for Jitted as well as compiled defines.

Lastly, an exception for variations of snapshot could be removed and replaced with a `static_assert`, since snapshot will be activated in RSnapshotOptions.

This is part 2 of the work in #19725.

### DIFF
--- a/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
@@ -1251,6 +1251,27 @@ public:
    /// \param[in] options RSnapshotOptions struct with extra options to pass to the output TFile and TTree/RNTuple.
    /// \return a `RDataFrame` that wraps the snapshotted dataset.
    ///
+   template <typename... ColumnTypes>
+   R__DEPRECATED(
+      6, 40, "Snapshot does not need template arguments anymore, you can safely remove them from this function call.")
+   RResultPtr<RInterface<RLoopManager>> Snapshot(std::string_view treename, std::string_view filename,
+                                                 const ColumnNames_t &columnList,
+                                                 const RSnapshotOptions &options = RSnapshotOptions())
+   {
+      return Snapshot(treename, filename, columnList, options);
+   }
+
+   ////////////////////////////////////////////////////////////////////////////
+   /// \brief Save selected columns to disk, in a new TTree or RNTuple `treename` in file `filename`.
+   /// \param[in] treename The name of the output TTree or RNTuple.
+   /// \param[in] filename The name of the output TFile.
+   /// \param[in] columnList The list of names of the columns/branches/fields to be written.
+   /// \param[in] options RSnapshotOptions struct with extra options to pass to TFile and TTree/RNTuple.
+   /// \return a `RDataFrame` that wraps the snapshotted dataset.
+   ///
+   /// This function returns a `RDataFrame` built with the output TTree or RNTuple as a source.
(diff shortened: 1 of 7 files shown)


tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -2362,13 +2361,13 @@ public:
    /// auto myGAE2 = myDf.GraphAsymmErrors<f, f, f, f, f, f>("xValues", "yValues", "exl", "exh", "eyl", "eyh");
    /// ~~~
    ///
-   /// `GraphAssymErrors` should also be used for the cases in which values associated only with 
-   /// one of the axes have associated errors. For example, only `ey` exist and `ex` are equal to zero. 
-   /// In such cases, user should do the following: 
+   /// `GraphAssymErrors` should also be used for the cases in which values associated only with
```
vepadulano:
```suggestion
   /// `GraphAsymmErrors` should also be used for the cases in which values associated only with
```

tree/dataframe/inc/ROOT/RDFHelpers.hxx
```diff
@@ -222,6 +222,10 @@ namespace Experimental {
 template <typename T>
 RResultMap<T> VariationsFor(RResultPtr<T> resPtr)
 {
+   using SnapshotResult_t = ROOT::RDF::RInterface<ROOT::Detail::RDF::RLoopManager, void>;
+   static_assert(!std::is_same_v<T, SnapshotResult_t>,
+                 "Snapshot with variations only can be enabled via RSnapshotOptions.");
```
vepadulano: I'm not sure we want to include this error in this PR, since the option is not added to RSnapshotOptions yet.

tree/dataframe/src/RDFHelpers.cxx
```diff
@@ -141,15 +141,7 @@ unsigned int ROOT::RDF::RunGraphs(std::vector<RResultHandle> handles)
    return uniqueLoops.size();
 }
 
-ROOT::RDF::Experimental::SnapshotPtr_t ROOT::RDF::Experimental::VariationsFor(ROOT::RDF::Experimental::SnapshotPtr_t)
-{
-   throw std::logic_error("Varying a Snapshot result is not implemented yet.");
-}
```
vepadulano: Also hesitant about removing this in this PR since the actual implementation is not available yet.

vepadulano (changes requested):
LGTM overall but I'm hesitant in mentioning that there will be an implementation that can be activated via RSnapshotOptions in this PR since the implementation is not given at the same time

tree/dataframe/inc/ROOT/RDFHelpers.hxx
hageboeck: It was easy to take out this commit. It will anyway come back when the later PRs are rebased.

vepadulano (approved):
Thanks!

PR #19727, opened 2025-08-22
[RDF] Prepare for snapshot with variations (part 4)

hageboeck:
This is part 4 of the work in #19725.
- [x] Please review part 2 #19726 first, since the commits will disappear from here once it gets rebased.

This PR is a refactor of the members of all TTree snapshot helpers, which will allow keeping all necessary data for a column in a single struct. This will simplify snapshot with variations, since varied columns can be created by copying the struct of the original column, and overriding a few members.
If the data hadn't been colocated like it's done here, it would become much harder to keep everything in a consistent state for both nominal and variations.


tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -49,35 +86,51 @@ TBranch *SearchForBranch(TTree *inputTree, const std::string &branchName)
...
-         // The null branch address is a placeholder. It will be set when SetBranchesHelper is called for `sizeLeafName`
-         auto *outputBranch = outputTree.Branch(sizeLeafName.c_str(), static_cast<void *>(nullptr),
-                                                (sizeLeafName + '/' + sizeTypeStr).c_str(), bufSize);
-         outputBranches.Insert(sizeLeafName, outputBranch);
+      if (sizeLeaf) {
+         // The array branch `bname` has dynamic size stored in leaf `sizeLeafName`, so we need to ensure that it's
+         // in the output tree.
+         auto sizeLeafIt =
+            std::find_if(outputBranches.begin(), outputBranches.end(),
+                         [&sizeLeafName](RBranchData const &bd) { return bd.fOutputBranchName == sizeLeafName; });
```
vepadulano: Nitpick: in RDF there is the practice of using const-qualifiers on the left of the type

```suggestion
                         [&sizeLeafName](const RBranchData &bd) { return bd.fOutputBranchName == sizeLeafName; });
```

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -49,35 +86,51 @@ TBranch *SearchForBranch(TTree *inputTree, const std::string &branchName)
...
+            const auto indexBeforeEmplace = std::distance(outputBranches.begin(), thisBranch);
+            outputBranches.emplace_back("", sizeLeafName, /*isDefine=*/false, /*typeID=*/nullptr,
+                                        /*outputBranch=*/nullptr);
+            thisBranch = outputBranches.begin() + indexBeforeEmplace;
+            sizeLeafIt = outputBranches.end() - 1;
+         }
+         if (!sizeLeafIt->fOutputBranch) {
+            // The size leaf was emplaced, but not initialised yet
+            const auto sizeTypeStr = ROOT::Internal::RDF::TypeName2ROOTTypeName(sizeLeaf->GetTypeName());
+            // Use Original basket size for Existing Branches otherwise use Custom basket Size.
```
vepadulano:
```suggestion
            // Use original basket size for existing branches otherwise use custom basket size.
```

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -449,17 +499,16 @@ void ROOT::Internal::RDF::UntypedSnapshotTTreeHelper::UpdateCArraysPtrs(const st
    // associated to those is re-allocated. As a result the value of the pointer can change therewith
    // leaving associated to the branch of the output tree an invalid pointer.
    // With this code, we set the value of the pointer in the output branch anew when needed.
-   assert(values.size() == fBranches.size());
+   assert(values.size() <= fBranchData.size());
```
vepadulano: Can you remind me of the reason for this change? IIRC it's because the Snapshot helper will be able to deal both with the nominal-only case and with the case including systematic variations so there may be less values than the number of branches. But just by typing this out something doesn't click in my mind :/

hageboeck: This looks to be an outdated change, which results from porting the previous implementation over to the type-less snapshot. Good spot!

Since the variations are handled in a different helper now, the old assertion might still hold. I'll try reverting this now.

hageboeck: The revert was successful. The assertions work as before now.

PR #19729, opened 2025-08-22
[tmva][sofie] Fix initialisation of BasicBinary operator

lmoneta:
In the initialization of BasicBinary operator  do not change the inputshapes of the tensors, just compute the output shape. Do instead broadcasting of tensors and implement the binary operation at initialization phase  if input tensors are constant.

### DIFF
--- a/tmva/sofie/inc/TMVA/ROperator_BasicBinary.hxx
+++ b/tmva/sofie/inc/TMVA/ROperator_BasicBinary.hxx
@@ -129,33 +129,37 @@ public:
       // check if need to broadcast at initialization time if shapes are known and different
       // (we could broadcast the tensor tensor to maximum values of dynamic shapes - to be done)
       // case of known shapes
+      // if shapes are known find the output shape from broadcasting
       if (dynamicInputs == 0) {
          auto ret = UTILITY::MultidirectionalBroadcastShape(fShapeA, fShapeB);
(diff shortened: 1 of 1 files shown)


tmva/sofie/inc/TMVA/ROperator_BasicBinary.hxx
```diff
@@ -129,33 +129,40 @@ public:
...
-         if (broadcast) {
-            // Y is the common shape of A and B
-            bool broadcastA = ret.first & 2;
-            bool broadcastB = ret.first & 1;
-            // Broadcast A to Y
-            if (broadcastA) {
-               fNBroadcastedA = "Broadcasted" + fNA + "to" + fNY;
-               if (model.IsConstantTensor(fNA)) {
+         if (model.IsConstantTensor(fNA) && model.IsConstantTensor(fNB)) {
+            auto ret = UTILITY::MultidirectionalBroadcastShape(fShapeA, fShapeB);
```
sanjibansg: `ret` is redefined here

lmoneta: Thank you for spotting this. It should be fixed now

sanjibansg (commented):
One last thing, maybe we need to fix the clang-format

PR #19732, opened 2025-08-23
[hist] Implement initial `RHistEngine`

hahnjo:
It combines an `RAxes` object and storage of bin contents.

### DIFF
--- a/hist/histv7/inc/ROOT/RHistEngine.hxx
+++ b/hist/histv7/inc/ROOT/RHistEngine.hxx
@@ -1,0 +1,212 @@
+/// \file
+/// \warning This is part of the %ROOT 7 prototype! It will change without notice. It might trigger earthquakes.
+/// Feedback is welcome!
+
+#ifndef ROOT_RHistEngine
+#define ROOT_RHistEngine
+
+#include "RAxes.hxx"
+#include "RBinIndex.hxx"
+#include "RLinearizedIndex.hxx"
+#include "RRegularAxis.hxx"
+
+#include <array>
+#include <cassert>
+#include <stdexcept>
+#include <tuple>
+#include <utility>
+#include <vector>
+
+class TBuffer;
+
+namespace ROOT {
+namespace Experimental {
+
+/**
+A histogram data structure to bin data along multiple dimensions.
+
+Every call to \ref Fill(const A &... args) "Fill" bins the data according to the axis configuration and increments the
+bin content:
+\code
+ROOT::Experimental::RHistEngine<int> hist(10, 5, 15);
+hist.Fill(8.5);
+// hist.GetBinContent(ROOT::Experimental::RBinIndex(3)) will return 1
+\endcode
+
+The class is templated on the bin content type. For counting, as in the example above, it may be an integer type such as
+`int` or `long`. Narrower types such as `unsigned char` or `short` are supported, but may overflow due to their limited
+range and must be used with care. For weighted filling, the bin content type must be a floating-point type such as
+`float` or `double`. Note that `float` has a limited significant precision of 24 bits.
+
+An object can have arbitrary dimensionality determined at run-time. The axis configuration is passed as a vector of
+RAxisVariant:
+\code
+std::vector<ROOT::Experimental::RAxisVariant> axes;
+axes.push_back(ROOT::Experimental::RRegularAxis(10, 5, 15));
+axes.push_back(ROOT::Experimental::RVariableBinAxis({1, 10, 100, 1000}));
+ROOT::Experimental::RHistEngine<int> hist(axes);
+// hist.GetNDimensions() will return 2
+\endcode
+
+\warning This is part of the %ROOT 7 prototype! It will change without notice. It might trigger earthquakes.
+Feedback is welcome!
+*/
+template <typename BinContentType>
+class RHistEngine final {
+   /// The axis configuration for this histogram. Relevant methods are forwarded from the public interface.
+   Internal::RAxes fAxes;
+   /// The bin contents for this histogram
+   std::vector<BinContentType> fBinContents;
+
+public:
+   /// Construct a histogram engine.
+   ///
+   /// \param[in] axes the axis objects, must have size > 0
+   explicit RHistEngine(std::vector<RAxisVariant> axes) : fAxes(std::move(axes))
+   {
+      fBinContents.resize(fAxes.ComputeTotalNBins());
+   }
+
+   /// Construct a one-dimensional histogram engine with a regular axis.
+   ///
+   /// \param[in] nNormalBins the number of normal bins, must be > 0
+   /// \param[in] low the lower end of the axis interval (inclusive)
+   /// \param[in] high the upper end of the axis interval (exclusive), must be > low
+   /// \par See also
+   /// the \ref RRegularAxis::RRegularAxis(std::size_t nNormalBins, double low, double high, bool enableFlowBins)
+   /// "constructor of RRegularAxis"
+   RHistEngine(std::size_t nNormalBins, double low, double high) : RHistEngine({RRegularAxis(nNormalBins, low, high)})
+   {
+   }
+
+   // Copy constructor and assignment operator are deleted to avoid surprises.
+   RHistEngine(const RHistEngine<BinContentType> &) = delete;
+   RHistEngine(RHistEngine<BinContentType> &&) = default;
+   RHistEngine<BinContentType> &operator=(const RHistEngine<BinContentType> &) = delete;
+   RHistEngine<BinContentType> &operator=(RHistEngine<BinContentType> &&) = default;
+   ~RHistEngine() = default;
+
+   const std::vector<RAxisVariant> &GetAxes() const { return fAxes.Get(); }
+   std::size_t GetNDimensions() const { return fAxes.GetNDimensions(); }
+   std::size_t GetTotalNBins() const { return fBinContents.size(); }
+
+   /// Get the content of a single bin.
+   ///
+   /// \code
+   /// ROOT::Experimental::RHistEngine<int> hist({/* two dimensions */});
+   /// std::array<ROOT::Experimental::RBinIndex, 2> indices = {3, 5};
+   /// int content = hist.GetBinContent(indices);
+   /// \endcode
+   ///
+   /// \note Compared to TH1 conventions, the first normal bin has index 0 and underflow and overflow bins are special
+   /// values. See also the class documentation of RBinIndex.
+   ///
+   /// Throws an exception if the number of indices does not match the axis configuration or the bin is not found.
+   ///
+   /// \param[in] indices the array of indices for each axis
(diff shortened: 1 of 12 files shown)


hist/histv7/inc/ROOT/RHistEngine.hxx
```diff
@@ -0,0 +1,198 @@
...
+// hist.GetNDimensions() will return 2
+\endcode
+
+\warning This is part of the %ROOT 7 prototype! It will change without notice. It might trigger earthquakes.
+Feedback is welcome!
+*/
+template <typename T>
+class RHistEngine final {
+public:
+   using BinContentType = T;
```
hageboeck: Although functionally equivalent, wouldn't this remove a mental indirection, and possibly even help with syntax highlighting depending on the editor config?
```diff
- template <typename T>
+ template<typename BinContentType>
  class RHistEngine final {
  public:
-    using BinContentType = T;
```

hist/histv7/inc/ROOT/RHistEngine.hxx
```diff
@@ -0,0 +1,198 @@
...
+      fBinContents.resize(fAxes.ComputeTotalNBins());
+   }
+
+   /// Construct a one-dimensional histogram engine with a regular axis.
+   ///
+   /// \param[in] nNormalBins the number of normal bins, must be > 0
+   /// \param[in] low the lower end of the axis interval (inclusive)
+   /// \param[in] high the upper end of the axis interval (exclusive), must be > low
+   /// \sa the \ref RRegularAxis::RRegularAxis(std::size_t nNormalBins, double low, double high, bool enableFlowBins)
+   /// "constructor of RRegularAxis"
```
hageboeck: If I'm not mistaken `\sa` tries to generate links to every word that follows, so also to "the".
I didn't test whether `\ref name "link text"` works correctly, but most probably, the "the" should be removed.

A not-so-great example can be found here:
https://root.cern/doc/master/classTTree.html#aced3b1ddb415f809ad7bb33cd7356f14
originating from
https://github.com/root-project/root/blob/1b908574e9ff615bcb47810459c208bd42376289/tree/tree/inc/TTree.h#L722-L728


Another solution is to replace `\sa` by a plain "See", and add a blank line before, which should generate a normal doxygen paragraph.

hist/histv7/inc/ROOT/RHistEngine.hxx
```diff
@@ -0,0 +1,198 @@
...
+   /// std::array<ROOT::Experimental::RBinIndex, 2> indices = {3, 5};
+   /// int content = hist.GetBinContent(indices);
+   /// \endcode
+   ///
+   /// Throws an exception if the number of indices does not match the axis configuration or the bin is not found.
+   ///
+   /// \param[in] indices the array of indices for each axis
+   /// \return the bin content
+   /// \sa the \ref GetBinContent(const A &... args) const "variadic function template overload" accepting arguments
+   /// directly
```
hageboeck: Same as above, maybe "the" and "accepting arguments directly" should be removed. Or convert to "`See \ref xxx "yyy"`"

(The same goes for the following functions.)

hist/histv7/inc/ROOT/RHistEngine.hxx
```diff
@@ -0,0 +1,198 @@
...
+      }
+      assert(index.fIndex < fBinContents.size());
+      return fBinContents[index.fIndex];
+   }
+
+   /// Get the content of a single bin.
+   ///
+   /// \code
+   /// ROOT::Experimental::RHistEngine<int> hist({/* two dimensions */});
+   /// int content = hist.GetBinContent(ROOT::Experimental::RBinIndex(3), ROOT::Experimental::RBinIndex(5));
```
hageboeck, line 131: Looking at the constructor of RBinIndex, this should work, too, wouldn't it?
```
hist.GetBinContent(3, 5);
```
Is it worth pointing this out in the example? If users don't want to ask for under/overflow, they might be happy to see that simple indexing is possible.

hist/histv7/inc/ROOT/RHistEngine.hxx
```diff
@@ -0,0 +1,198 @@
...
+   ///
+   /// \param[in] args the arguments for each axis
+   /// \sa the \ref Fill(const A &... args) "variadic function template overload" accepting arguments directly
+   template <typename... A>
+   void Fill(const std::tuple<A...> &args)
+   {
+      // We could rely on RAxes::ComputeGlobalIndex to check the number of arguments, but its exception message might
+      // be confusing for users.
+      if (sizeof...(A) != GetNDimensions()) {
+         throw std::invalid_argument("invalid number of arguments to Fill");
```
hageboeck, line 175: Could one maybe format `"<sizeof...(A)> vs <GetNDimensions>"` into the string to make `exception::what()` more insightful, and also add the function name?

hist/histv7/inc/ROOT/RHistEngine.hxx
```diff
@@ -0,0 +1,198 @@
...
+   /// \return the bin content
+   /// \sa the \ref GetBinContent(const A &... args) const "variadic function template overload" accepting arguments
+   /// directly
+   template <std::size_t N>
+   const BinContentType &GetBinContent(const std::array<RBinIndex, N> &indices) const
+   {
+      // RAxes::ComputeGlobalIndex will check the number of indices.
+      RLinearizedIndex index = fAxes.ComputeGlobalIndex(indices);
+      if (!index.fValid) {
+         throw std::invalid_argument("bin not found");
```
hageboeck: Maybe one could make this a bit easier to trace by adding "RHistEngine::GetBinContent(): bin not found"?

hist/histv7/inc/ROOT/RHistEngine.hxx
hahnjo: The idea was to expose the bin content type with a `using` declaration, but that is not allowed to shadow the template parameter. But in the end, maybe we don't actually need `RHistEngine<T>::BinContentType` - I've removed it for now as you suggested.

hist/histv7/inc/ROOT/RHistEngine.hxx
hahnjo: In principle, every word in `\sa` should be checked, but only auto-linked when available. It's not clear to me why that's the case for generic words such as "the"... I'll investigate what to do.

hist/histv7/inc/ROOT/RHistEngine.hxx
hahnjo: So far the other exception messages don't prefix the function name. I would argue that the stack trace should be used to find where the exception was thrown...

hist/histv7/inc/ROOT/RHistEngine.hxx
hahnjo: In principle, I want to encourage users to explicitly work with `RBinIndex`. As discussed, I extended the snippet code and added a note about the change of conventions compared to TH1.

hist/histv7/inc/ROOT/RHistEngine.hxx
hahnjo: I was worried that dynamic exception messages with `std::string` could have a performance hit, and it seems to be the case when built with Clang:
```
RHistEngine_int_Regular1/Fill/4096       6645 ns         6637 ns       106185
RHistEngine_int_Regular2/Fill/4096      25946 ns        25915 ns        26760
```
goes to
```
RHistEngine_int_Regular1/Fill/4096      34293 ns        34253 ns        20438
RHistEngine_int_Regular2/Fill/4096      35925 ns        35883 ns        19754
```
GCC seems to be unaffected, maybe because of different code layout? I would like to investigate; if possible without performance hit, we can later on implement it consistently also for the exceptions in `RAxes`.

hist/histv7/inc/ROOT/RHistEngine.hxx
hahnjo: As discussed offline, there are too many functions and other things in the global namespace for the full ROOT documentation. I've converted all `\sa` to the user defined heading `\par See also`.

hist/histv7/inc/ROOT/RHistEngine.hxx
hahnjo: Right, so with dynamic `std::string` concatenation neither `Fill()` nore `ComputeGlobalIndex()` are inlined anymore, which explains the performance penalty. It's possible to "fix" by making sure to outline the construction of the `std::string` into a separate function:
```c++
std::string InvalidNumberOfArguments(std::size_t N) const
{   
   std::string msg = "invalid number of arguments to Fill: ";
   msg += std::to_string(N) + " for " + std::to_string(GetNDimensions()) + " dimensions";
   return msg;
}

throw std::invalid_argument(InvalidNumberOfArguments(sizeof...(A)));
```

This gives:
```
RHistEngine_int_Regular1/Fill/4096       6640 ns         6632 ns       106346
RHistEngine_int_Regular2/Fill/4096      25708 ns        25678 ns        27067
```

However I fear this depends on the compiler (version) and might change at any point. I think I would prefer to keep constant exception message strings for performance critical functions.

Fun fact, I first tried to outline `ThrowInvalidNumberOfArguments`:
```
RHistEngine_int_Regular1/Fill/4096      18731 ns        18709 ns        37460
RHistEngine_int_Regular2/Fill/4096      30178 ns        30143 ns        23226
```
That's worse because the compiler cannot see `throw` in that case; adding `[[unlikely]]` brings it further down to:
```
RHistEngine_int_Regular1/Fill/4096      11911 ns        11897 ns        61084
RHistEngine_int_Regular2/Fill/4096      21541 ns        21516 ns        32604
```

hageboeck: OK, let's leave it like that.

hist/histv7/inc/ROOT/RHistEngine.hxx
hahnjo: Actually my statement is not 100% correct, most exception messages have "in \<method\>". Let's do the same here, and come back to this later when we discuss exception types and messages in the team.

hist/histv7/inc/ROOT/RHistEngine.hxx
```diff
@@ -0,0 +1,212 @@
...
+   /// \param[in] args the arguments for each axis
+   /// \par See also
+   /// the \ref Fill(const A &... args) "variadic function template overload" accepting arguments directly
+   template <typename... A>
+   void Fill(const std::tuple<A...> &args)
+   {
+      // We could rely on RAxes::ComputeGlobalIndex to check the number of arguments, but its exception message might
+      // be confusing for users.
+      if (sizeof...(A) != GetNDimensions()) {
+         throw std::invalid_argument("invalid number of arguments to Fill");
```
pcanal, line 175: Consider adding a comment (here and probably elsewhere) warning future maintainer to not 'improve' the error message at the expense of performance.

PR #19735, opened 2025-08-24
[RF] Don't format infinities to large numbers anymore in `codegen`

guitargeek:
By now, Clad should have learned how to deal with the `inf` infinities.

The unit tests are tweaked the check this.


vgvassilev (approved):
Quite nice simplification. LGTM!

Issue #19077, opened 2025-06-18
Crash when exiting python if a writable TFile is still open (appears to be modification of freed object?)


will-cern:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

I have been trying to isolate a nasty crash that occurs when python exits. It appears to be triggered when there are writable TFiles still open at the end of the script. The reproducer looks a bit strange but is reflecting the sort of file my code was building. I do not know if the problem is related to the specific writing of those objects (a FitConfig and a RooLinkedList) but the reproducer does at least reproduce the crash 9 times out of 10. 

If the files are closed before the end of the script, there is no crash.

The crash seems to occur because of the call to the WriteStreamerInfo method of the writable files when they are being closed. It appears as if something that the StreamerInfo needs has perhaps already been destroyed?

When the crash occurs on my mac, it reports that there is a modification of an already freed piece of memory:

```
python(80807,0x7ff8572ddfc0) malloc: Heap corruption detected, free list is damaged at 0x600006d00240
```
Here's the stack trace I get from attaching a debugger:

```
__pthread_kill 0x00007ff813d6414a
pthread_kill 0x00007ff813d9cebd
abort 0x00007ff813cc2a79
malloc_vreport 0x00007ff813bc73a9
malloc_zone_error 0x00007ff813beac5b
nanov2_guard_corruption_detected 0x00007ff813bdc389
nanov2_allocate_outlined 0x00007ff813bdb692
nanov2_malloc 0x00007ff813bba3c7
operator new(unsigned long) 0x00007ff813d5734a
std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>* std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>>::__push_back_slow_path<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) 0x0000000105af6c4f
[...]

PR #19737, opened 2025-08-25
update xroofit

will-cern:
This PR updates xRooFit to match latest development version. Once merged I will create a tag for it in the main xroofit repo.

### DIFF
--- a/io/io/inc/TMemFile.h
+++ b/io/io/inc/TMemFile.h
@@ -106,6 +106,8 @@ public:
 
            void        Print(Option_t *option="") const override;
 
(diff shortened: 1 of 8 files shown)


dpiparo:
Thanks @will-cern . Running the changes through the CI.

will-cern:
@pcanal the `SetWritable` method is introduced to `TMemFile` to workaround a bug affecting pyROOT when it has open files at the time it exits (see https://github.com/root-project/root/issues/19077). By allowing me to mark the TMemFile as read-only even though I created it in write-access mode allows me to bypass the crash.

guitargeek (approved):
Thank you very much for the update!

PR #19738, opened 2025-08-25
[test] fix name conflict when multithreading

ferdymercury:
# This Pull request:

## Changes or fixes:

see https://github.com/root-project/root/actions/runs/17157474273/job/48677881584?pr=19688

since TestObj.cpp creates same-named file in same directory.

### DIFF
--- a/roottest/root/tree/array/execStdArray.C
+++ b/roottest/root/tree/array/execStdArray.C
@@ -1,5 +1,5 @@
 void CreateTreeWithStdArray(){
-   TFile f("test.root", "RECREATE");
+   TFile f("testStdArray.root", "RECREATE");
    TTree t("tree", "tree");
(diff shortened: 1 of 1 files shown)


linev (commented):
Do you have example of such conflict?

File "test.root" created in dedicated directory for `root/tree/array` tests
and therefore should not conflict with any other tests

ferdymercury:
> Do you have example of such conflict?

https://github.com/root-project/root/actions/runs/17157474273/job/48677881584?pr=19688

> File "test.root" created in dedicated directory for `root/tree/array` tests

TestObj.cpp creates same-named file in the same `root/tree/array` directory.

linev (approved):
Now I see. Thanks to fixing it.

PR #19740, opened 2025-08-25
[tmva][sofie] Document why Tanh must stay std::tanh, use size_t loop idx

olia110:
Keep std::tanh in the generated code for the Tanh operator and add a comment deterring exp-based rewrites like (e^{2x}-1)/(e^{2x}+1): they produce NaN for x > ~44 in single precision (exp overflows to inf, giving inf/inf) and lose all relative accuracy near x = 0 through cancellation, while measurements on Linux (glibc 2.42) show only a 20-25% speedup. Vectorizing the generated loop (SIMD tanhf via libmvec with -ffast-math) is ~4x and does not compromise numerical quality.

Also switch the loop index from int to size_t.

### DIFF
--- a/tmva/sofie/inc/TMVA/ROperator_Tanh.hxx
+++ b/tmva/sofie/inc/TMVA/ROperator_Tanh.hxx
@@ -57,7 +57,14 @@ public:
       std::stringstream out;
       size_t length = ConvertShapeToLength(fShape);
       out << "\n//------ TANH\n";
-      out << SP << "for (int id = 0; id < " << length << " ; id++){\n";
+      // Deliberately std::tanh: rewriting this as (e^{2x}-1)/(e^{2x}+1) (or
+      // similar exp-based formulas) returns NaN for x > ~44 in single
+      // precision (exp overflows to inf, giving inf/inf) and loses all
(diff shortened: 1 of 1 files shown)


tmva/sofie/inc/TMVA/ROperator_Tanh.hxx
```diff
@@ -6,6 +6,8 @@
 #include "TMVA/RModel.hxx"
 
 #include <sstream>
+#include <vector>
+#include <string>
```
sanjibansg: why do we need to include the `vector` and `<string>` libraries here?

sanjibansg (approved):
LGTM! Just one small comment

lmoneta (commented):
These changes should be tested also on Linux and I think there is a problem with the function returning NaN for large value of x (inf/inf). I would test further if other improvements are possible without affecting the numerical quality

guitargeek (approved):
@lmoneta is right, the proposed optimization was not numerically safe.

Let's stay with `std::tanh`, and instead commit a little inline comment to deter the next person to look at the code from making the same "optimization".

Thanks @olia110 for kickstarting the conversation about this!

PR #19561, opened 2025-08-07
[tree] allow skipping longdoubles when building tree index

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/19560

## Checklist:

- [ ] tested changes locally
- [x] updated the docs (if necessary)

PR #19741, opened 2025-08-25
[tree] Fix long64 tree index losing precision on 64-bit long double platforms and enable forgotten test

### LINKED ISSUE #19560: TTreeIndex fails under Valgrind due to long double emulation
### Check duplicate issues.

- [x] Checked for duplicates

### Description

In ATLAS, we make use of custom `TTree` indices via `TTree::BuildIndex()`. Recently, we encountered an issue when running jobs under Valgrind: entries that are accessible under normal execution become unreadable when running the same code under Valgrind.

Upon investigation, it appears that `TTreeIndex` uses `long double` for internal calculations. However, Valgrind does not fully emulate 80-bit `long double` precision on `x86_64` systems. It simulates `long double` as a 64-bit `double`, losing precision beyond a certain limit.

As a result, when index values fall in the range where `double` loses precision but `long double` remains exact, the index lookup fails under Valgrind. This affects our ability to read ROOT files in Valgrind-based memory checks when using such index values.

We understand that this is fundamentally a limitation of Valgrind’s `long double` emulation, not ROOT itself, but we’re wondering if anything can be done on the ROOT side to work around the issue. Many thanks.

### Reproducer

```
#include <TFile.h>
#include <TTree.h>
#include <TTreeIndex.h>

#include <format>
#include <iostream>

int main() {

  // A large enough index that can be represented by long double but not double
  const Long64_t value = 9007199254743041LL;

  // Write
  {
    TFile file("file.root", "RECREATE");
    TTree tree("tree", "tree");
    Long64_t index{value};
    tree.Branch("index", &index);
    tree.Fill();
    tree.Write();
    file.Close();
  }

  //Read
  {
    TFile file("file.root", "READ");
    TTree* tree = file.Get<TTree>("tree");
    tree->BuildIndex("index");
    auto entry = tree->GetEntryNumberWithIndex(value); 
    std::cout << std::format("Entry: {} for Value: {}", entry, value) << std::endl;
    file.Close();
  }

  return 0;
}
```

Running this against the head of main (e.g., `dev3LCG` nightly), one gets:

```
Entry: 0 for Value: 9007199254743041
```

under normal execution while it fails under Valgrind with:

```
Warning in <TTreeIndex::TTreeIndex>: In tree entry 0, major value index=9007199254743040.000000 possibly out of range for internal `long double`
Entry: -1 for Value: 9007199254743041
```

### ROOT version

ROOT Version: 6.37.01
Built for linuxx8664gcc on Aug 05 2025, 22:50:58
From heads/master@v6-37-01-7673-g25c19af92e

### Installation method

LCG

### Operating system

EL9

### Additional context

_No response_

ferdymercury:
This is a fixup of https://github.com/root-project/root/pull/19561 that was supposed to fix https://github.com/root-project/root/issues/19560

A test was created there, but it was forgotten to add it to the CMakeLists. As a consequence, the CI looked green but the test was not running at all, and it was actually broken.

This PR reenables the test as well as fixes the TTree code to unbreak the test.

### DIFF
--- a/tree/treeplayer/src/TTreeIndex.cxx
+++ b/tree/treeplayer/src/TTreeIndex.cxx
@@ -189,7 +189,7 @@ TTreeIndex::TTreeIndex(const TTree *T, const char *majorname, const char *minorn
          Error("TTreeIndex", "In tree entry %lld, Ndata in formula is zero for both '%s' and '%s'", i,
                fMajorName.Data(), fMinorName.Data());
       }
-      auto GetAndRangeCheck = [this](bool isMajor, Long64_t entry) {
+      auto GetAndRangeCheck = [this](bool isMajor, Long64_t entry) -> Long64_t {
          LongDouble_t ret = (isMajor ? fMajorFormula : fMinorFormula)->EvalInstance<LongDouble_t>();
          // Check whether the value (vs significant bits) of ldRet can represent
          // the full precision of the returned value. If we return 10^60, the
@@ -206,11 +206,16 @@ TTreeIndex::TTreeIndex(const TTree *T, const char *majorname, const char *minorn
                     "In tree entry %lld, %s value %s=%Lf possibly out of range for internal `long double`", entry,
                     isMajor ? "major" : "minor", isMajor ? fMajorName.Data() : fMinorName.Data(), ret);
          }
-         return ret;
+         return static_cast<Long64_t>(ret); // static_cast necessary for use in ternary expression below
       };
       auto GetLong64 = [this](bool isMajor) {
          return (isMajor ? fMajorFormula : fMinorFormula)->EvalInstance<Long64_t>();
       };
+      // Note: in ternary expression ensure both branches have same return data type Long64_t.
+      // otherwise, with LongDouble_t, the expression is implicitly cast to their
+      // common type `long double` regardless of the branch taken. On platforms where
+      // `sizeof(long double) == sizeof(double)` (macOS arm64, Windows) this silently
+      // rounds the exact Long64_t value, defeating the purpose of the long64 flag.
       tmp_major[i] = long64major ? GetLong64(true) : GetAndRangeCheck(true, i);
       tmp_minor[i] = long64minor ? GetLong64(false) : GetAndRangeCheck(false, i);
    }
--- a/roottest/root/tree/index/CMakeLists.txt
+++ b/roottest/root/tree/index/CMakeLists.txt
@@ -15,6 +15,10 @@ ROOTTEST_ADD_TEST(index64
                   MACRO runindex64.C
                   OUTREF index64.ref)
 
+ROOTTEST_ADD_TEST(indexl64
+                  MACRO runindexl64.C
+                  OUTREF indexl64.ref)
+
 ROOTTEST_ADD_TEST(varsizearr
                   MACRO runvarsizearr.C
                   OUTREF varsizearr.ref)
--- a/roottest/root/tree/index/runindexl64.C
+++ b/roottest/root/tree/index/runindexl64.C
@@ -5,9 +5,8 @@
 bool test(TTree *);
 
 const char *fname = "indexl64.root";
-https : // github.com/root-project/root/pull/19561
-        const Long64_t bigval =
-           0x0FFFFFFFFFFFFFFF; // here we skip long double, so we can go higher than with runindex64.C
+// https://github.com/root-project/root/pull/19561
+const Long64_t bigval = 0x0FFFFFFFFFFFFFFF; // here we skip long double, so we can go higher than with runindex64.C
 
 int runindexl64()
 {


ferdymercury:
It seems Windows/MacOS still do not like big long64 values... :s

ferdymercury:
@guitargeek since your last PRs touch some tree code, could you maybe check if Claude can also help with reenabling this broken test?

guitargeek:
> @guitargeek since your last PRs touch some tree code, could you maybe check if Claude can also help with reenabling this broken test?

The agentic Claude Code workflow that I set up runs in a sandboxed Linux VM (I want to use it in auto mode without risking the files on my workstation), but I asked it to attempt to fix it anyway by providing the CI logs. The change it suggests looks harmless, let's see if it works!

ferdymercury:
Thanks a lot !!

tree/treeplayer/src/TTreeIndex.cxx
```diff
@@ -211,8 +211,20 @@ TTreeIndex::TTreeIndex(const TTree *T, const char *majorname, const char *minorn
...
+      // `sizeof(long double) == sizeof(double)` (macOS arm64, Windows) this silently
+      // rounds the exact Long64_t value, defeating the purpose of the long64 flag.
+      if (long64major)
+         tmp_major[i] = GetLong64(true);
+      else
+         tmp_major[i] = GetAndRangeCheck(true, i);
+      if (long64minor)
+         tmp_minor[i] = GetLong64(false);
+      else
+         tmp_minor[i] = GetAndRangeCheck(false, i);
```
ferdymercury: Seems to work, thanks !

I'd suggest this as a more compact alternative fix, but either way seems fine to me, so feel free to apply or not:

(It should be in place of 192 to 227, not sure why Github is not diffing correctly)

```suggestion
      auto GetAndRangeCheck = [this](bool isMajor, Long64_t entry) {
         LongDouble_t ret = (isMajor ? fMajorFormula : fMinorFormula)->EvalInstance<LongDouble_t>();
         // Check whether the value (vs significant bits) of ldRet can represent
         // the full precision of the returned value. If we return 10^60, the
         // value fits into a long double, but if sizeof(long double) ==
         // sizeof(double) it cannot store the ones: the value returned by
         // EvalInstance() only stores the higher bits.
         LongDouble_t retCloserToZero = ret;
         if (ret > 0)
            retCloserToZero -= 1;
         else
            retCloserToZero += 1;
         if (retCloserToZero == ret) {
            Warning("TTreeIndex",
                    "In tree entry %lld, %s value %s=%Lf possibly out of range for internal `long double`", entry,
                    isMajor ? "major" : "minor", isMajor ? fMajorName.Data() : fMinorName.Data(), ret);
         }
         return static_cast<Long64_t>(ret); // static_cast necessary for use in ternary expression below
      };
      auto GetLong64 = [this](bool isMajor) {
         return (isMajor ? fMajorFormula : fMinorFormula)->EvalInstance<Long64_t>();
      };
      // Note: in ternary expression ensure both branches have same return data type Long64_t.
      // otherwise, with LongDouble_t, the expression is implicitly cast to their
      // common type `long double` regardless of the branch taken. On platforms where
      // `sizeof(long double) == sizeof(double)` (macOS arm64, Windows) this silently
      tmp_major[i] = long64major ? GetLong64(true) : GetAndRangeCheck(true, i);
      tmp_minor[i] = long64minor ? GetLong64(false) : GetAndRangeCheck(false, i);
```

tree/treeplayer/src/TTreeIndex.cxx
```diff
@@ -206,11 +207,15 @@ TTreeIndex::TTreeIndex(const TTree *T, const char *majorname, const char *minorn
...
-         return ret;
+         return static_cast<Long64_t>(ret); // static_cast necessary for use in ternary expression below
       };
       auto GetLong64 = [this](bool isMajor) {
          return (isMajor ? fMajorFormula : fMinorFormula)->EvalInstance<Long64_t>();
       };
+      // Note: in ternary expression ensure both branches have same return data type Long64_t.
+      // otherwise, with LongDouble_t, the expression is implicitly cast to their
+      // common type `long double` regardless of the branch taken. On platforms where
+      // `sizeof(long double) == sizeof(double)` (macOS arm64, Windows) this silently
```
ferdymercury, line 217: sentence unfinished?

guitargeek: Thanks for spotting, I missed this when copying your suggestion from the diff in the GitHub interface (it was indeed a bit weird that it couldn't display it correctly).

guitargeek (approved):
Thanks a lot for re-adding the test! The fix that I needed to apply was just to add the `static_cast<Long64_t>(ret)` to ensure the ternary expression has the right type.

ferdymercury:
Thaanks a lot @guitargeek , could you backport to 6.40?

Fyi @amete

PR #19742, opened 2025-08-25
[tmva][sofie] Fix binary operators for dynamic tensors

lmoneta:
After optimising the binary operators to avoid broadcasting, the fix needed to be done also for the case of dynamic tensors

Fix also an issue in simplifying the resulting output shape of Reshape when having dynamic tensor shapes


sanjibansg (approved):
LGTM! Thanks for fixing this

PR #19743, opened 2025-08-25
[tmva][sofie] Softmax contiguous memory fast path

olia110:
# This Pull request:
An optimized code path has been added for the common case where the Softmax axis is the last dimension (axis == size - 1).

### DIFF
--- a/tmva/sofie/inc/TMVA/ROperator_Softmax.hxx
+++ b/tmva/sofie/inc/TMVA/ROperator_Softmax.hxx
@@ -60,10 +60,10 @@ public:
       std::stringstream out;
       size_t size = fShape.size();
       auto length_str = ConvertDimShapeToLength(fShape);
-      int axis = fAttrAxis < 0 ? size + fAttrAxis : fAttrAxis;
+      size_t axis = fAttrAxis < 0 ? size + fAttrAxis : fAttrAxis;
 
       // Check if this is the special case where memory is contiguous.
-      if (axis == static_cast<int>(size - 1)) {
+      if (axis == size - 1) {
          std::string axis_size = fShape[axis].GetVal();
          std::string num_rows;
          if (IsInteger(length_str) && IsInteger(axis_size)) {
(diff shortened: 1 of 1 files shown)


tmva/sofie/inc/TMVA/ROperator_Softmax.hxx
```diff
@@ -58,199 +58,108 @@ public:
          throw std::runtime_error("TMVA SOFIE Operator Softmax called to Generate without being initialized first");
       }
       std::stringstream out;
-      int size = fShape.size();
-      auto length = ConvertDimShapeToLength(fShape);
-      auto stride = UTILITY::ComputeStrideFromShape(fShape);
+      size_t size = fShape.size();
+      auto length_str = ConvertDimShapeToLength(fShape);
+      size_t length = std::stoul(length_str);
```
lmoneta: you don't need to convert to an integer, since in some case it might fail : e.g length  = " N * 32"

tmva/sofie/inc/TMVA/ROperator_Softmax.hxx
```diff
@@ -58,199 +58,108 @@ public:
...
-         out << SP << SP << "tensor_" << fNY << "[i] = std::exp(tensor_" << fNX << "[i] - vmax);\n";
-         out << SP << SP << "sum += tensor_" << fNY << "[i];\n";
-         out << SP << "}\n";
-         out << SP << "for (size_t i = 0; i < " << length << " ; i++){\n";
-         out << SP << SP << "tensor_" << fNY << "[i] /= sum;\n";
+
+      // Check if this is the special case where memory is contiguous.
+      if (axis == static_cast<int>(size - 1)) {
+         size_t axis_size = std::stoul(fShape[axis].GetVal());
+         size_t num_rows = length / axis_size;
```
lmoneta: In case of length is not an integer , this could be written as following: 
```
std::string axis_size = fShape[axis].GetVal());
std::string num_rows; 
if (IsInteger(length) && IsInteger(axis_size)) { 
  num_rows = std::to_string( std::stoul(length_str)/ std::stoul(fShape[axis].GetVal()));
} else  {
  num_rows = "(" + length_str + ") /" + fShape[axis].GetVal():
}

**Review by lmoneta — CHANGES_REQUESTED** [review]
Thank you for the PR, 
Just some corrections for the case the shapes are not fully specified as integer values, but parametrised (dynamic)

**Review by sanjibansg — APPROVED** [review]
LGTM! Thanks for this!

File: tmva/sofie/inc/TMVA/ROperator_Softmax.hxx [hunk 3]
```diff
@@ -58,199 +58,113 @@ public:
          throw std::runtime_error("TMVA SOFIE Operator Softmax called to Generate without being initialized first");
       }
       std::stringstream out;
-      int size = fShape.size();
-      auto length = ConvertDimShapeToLength(fShape);
-      auto stride = UTILITY::ComputeStrideFromShape(fShape);
+      size_t size = fShape.size();
+      auto length_str = ConvertDimShapeToLength(fShape);
       int axis = fAttrAxis < 0 ? size + fAttrAxis : fAttrAxis;
```
lmoneta: Here we could have "size_t" for axis instead of int

tmva/sofie/inc/TMVA/ROperator_Softmax.hxx
```diff
@@ -58,199 +58,113 @@ public:
...
-         out << SP << fType << " sum = 0.0;\n";
-         out << SP << "for (size_t i = 0; i < " << length << " ; i++){\n";
-         out << SP << SP << "tensor_" << fNY << "[i] = std::exp(tensor_" << fNX << "[i] - vmax);\n";
-         out << SP << SP << "sum += tensor_" << fNY << "[i];\n";
-         out << SP << "}\n";
-         out << SP << "for (size_t i = 0; i < " << length << " ; i++){\n";
-         out << SP << SP << "tensor_" << fNY << "[i] /= sum;\n";
+
+      // Check if this is the special case where memory is contiguous.
+      if (axis == static_cast<int>(size - 1)) {
```
lmoneta: if we define axis as size_t we don;t need the casting here

tmva/sofie/inc/TMVA/ROperator_Softmax.hxx
```diff
@@ -58,199 +58,113 @@ public:
...
+         std::string axis_size = fShape[axis].GetVal();
+         std::string num_rows;
+         if (IsInteger(length_str) && IsInteger(axis_size)) {
+            num_rows = std::to_string(std::stoul(length_str) / std::stoul(axis_size));
+         } else {
+            num_rows = "(" + length_str + ") / (" + axis_size + ")";
+         }
+         
+         out << "\n" << SP << "//------ SOFTMAX - " << size << "  " << length_str << "  " << axis << "\n";
+         out << SP << "for (size_t i = 0; i < " << num_rows << "; ++i) {\n";
```
lmoneta: we could use here `int` in the generated code to save memory, as it is done in the general case below

lmoneta: Thanks Olha  for addressing the comments!

Issue #18924, opened 2025-05-31
PyROOT: assigning a numpy array to a std vector leaks memory


lwpiotr:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Assigning a numpy array to a std vector leaks memory. The attached script starts with 600 MB memory use and quickly goes into GBs. The commented-out vector::clear() does not help.

### Reproducer

```Python
#!/usr/bin/python

import numpy as np
import ROOT

v = ROOT.vector["vector<vector<int>>"]()
a = np.arange(100**3, dtype=np.int32).reshape(100,100,100)

for i in range(10000):
    #v.clear()
    v += a
```

### ROOT version

heads/master@v6-37-01-6541-g9a9e7697e4

### Installation method

Source

### Operating system

Linux

PR #19745, opened 2025-08-26
[Python] Don't add item converters to InitializerListConverter forever

guitargeek:
When using a `std::initializer_list` multiple times for a specific call, element converters should only be added if not already present, and not forever.

Reproducer:
```Python
import numpy as np
import ROOT
import psutil
import os

def memory_usage_mb():
    process = psutil.Process(os.getpid())
    return process.memory_info().rss / 1024**2

n = 10000
a = np.arange(n, dtype=np.int32).reshape(n,1)

ROOT.gInterpreter.Declare("""

void my_insert(std::initializer_list<std::vector<int>>) {}

""")

for i in range(100):
    ROOT.my_insert(a)

    if i % 10 == 0:
        print(f"Iteration {i}: Memory usage = {memory_usage_mb():.2f} MB")
```

Closes #https://github.com/root-project/root/issues/18924.


ferdymercury:
> Closes ##18924.

Thanks! There is a stray hash here preventing the automatic closing.

PR #19746, opened 2025-08-26
[ACLiC] Allow unresolved symbols for macOS

devajithvs:
This is a follow up to: https://github.com/root-project/root/pull/19716#issuecomment-3214377372

MacOS does not allow unresolved symbols in dynamic libraries unless passed `-Wl,-undefined,dynamic_lookup`.

On linux, ACLiC relies on leaving symbols unresolved and resolving them later at runtime.

We now add the flag when relinking on macOS. Like how it was done in https://github.com/root-project/root/pull/11432

To reproduce this behaviour:
    
```bash
$BUILD/bin/root.exe -e "gSystem->SetBuildDir(\"$BUILD/roottest/root/tree/addresses\", true)" \
-q -l -b "$ROOT/roottest/scripts/build.C(\"$ROOT/roottest/root/tree/addresses/ConfigRecord.cxx\",\"\",\"\")"
```

Which will result in:
```
Undefined symbols for architecture arm64:
  "RecRecordImp<RecHeader>::Class()", referenced from:
      RecRecordImp<RecHeader>::IsA() const in ConfigRecord_cxx_ACLiC_dict.o
      RecRecordImp<RecHeader>::ShowMembers(TMemberInspector&) const in ConfigRecord_cxx_ACLiC_dict.o
  "RecRecordImp<RecHeader>::Streamer(TBuffer&)", referenced from:
      vtable for RecRecordImp<RecHeader> in ConfigRecord_cxx_ACLiC_dict.o
  "RecHeader::Print(char const*) const", referenced from:
      RecRecordImp<RecHeader>::Print(char const*) const in ConfigRecord_cxx_ACLiC_dict.o
  "vtable for Context", referenced from:
      RecRecordImp<RecHeader>::RecRecordImp() in ConfigRecord_cxx_ACLiC_dict.o
   NOTE: a missing vtable usually means the first non-inline virtual member function has no definition.
  "vtable for RecHeader", referenced from:
      ConfigRecord::~ConfigRecord() in ConfigRecord_cxx_ACLiC_dict.o
      ConfigRecord::~ConfigRecord() in ConfigRecord_cxx_ACLiC_dict.o
      ROOT::deleteArray_ConfigRecord(void*) in ConfigRecord_cxx_ACLiC_dict.o
      RecRecordImp<RecHeader>::RecRecordImp() in ConfigRecord_cxx_ACLiC_dict.o
      RecRecordImp<RecHeader>::~RecRecordImp() in ConfigRecord_cxx_ACLiC_dict.o
      RecRecordImp<RecHeader>::~RecRecordImp() in ConfigRecord_cxx_ACLiC_dict.o
   NOTE: a missing vtable usually means the first non-inline virtual member function has no definition.
ld: symbol(s) not found for architecture arm64
```

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes #

### DIFF
--- a/core/base/src/TSystem.cxx
+++ b/core/base/src/TSystem.cxx
@@ -3828,6 +3828,10 @@ int TSystem::CompileMacro(const char *filename, Option_t *opt,
          ForeachSharedLibDep(library, CollectF);
 
          TString relink_cmd = cmd.Strip(TString::kTrailing, ';');
+#ifdef R__MACOSX
+         // Allow linking to succeed despite the missing symbols.
+         relink_cmd.ReplaceAll("-dynamiclib", "-dynamiclib -Wl,-w -Wl,-undefined,dynamic_lookup");
+#endif
          relink_cmd += depLibsFullPaths;
          if (verboseLevel > 3 && withInfo) {
             ::Info("ACLiC", "relinking against all dependencies");
--- a/roottest/root/tree/addresses/CMakeLists.txt
+++ b/roottest/root/tree/addresses/CMakeLists.txt
@@ -147,8 +147,7 @@ ROOTTEST_ADD_TEST(ursula
 #for some classes
 if(NOT MSVC OR win_broken_tests)
   ROOTTEST_COMPILE_MACRO(ConfigRecord.cxx
-                         FIXTURES_SETUP root-tree-addresses-ConfigRecord-fixture
-                         FIXTURES_REQUIRED root-tree-addresses-sueloader-fixture)
+                         FIXTURES_SETUP root-tree-addresses-ConfigRecord-fixture)
 
   ROOTTEST_COMPILE_MACRO(sueloader.C
                          FIXTURES_SETUP root-tree-addresses-sueloader-fixture)


linev:
Exactly same problem exists on Windows.
So fixing it on Mac does not solve it in general.

In the ROOTTEST_COMPILE_MACRO it can be solved by using `BUILDLIB` argument with macro where such symbols defined. See examples in `roottest/root/io/references` directory.

So if there is problem with compilation of ConfigRecord.cxx - one can add missing symbols using BUILDLIB argument

devajithvs:
> So if there is problem with compilation of ConfigRecord.cxx - one can add missing symbols using BUILDLIB argument

As @pcanal mentioned here: https://github.com/root-project/root/pull/19716#discussion_r2291881524, I'm not sure if the generation of `ConfigRecord_cxx` should depend on `sueloader_C` for the symbols.

linev:
https://github.com/root-project/root/pull/19747 

This is alternative solution for problem in roottest
I would not change ACLiC behavior only because of it

linev:
> As @pcanal mentioned here: [#19716 (comment)](https://github.com/root-project/root/pull/19716#discussion_r2291881524), I'm not sure if the generation of `ConfigRecord_cxx` should depend on `sueloader_C` for the symbols.

From your error message - compilation of ConfixRecord dictionary requires some symbols from sueloader.C

linev:
https://github.com/root-project/root/pull/19747. solves problem on Windows.
So it also should work on Mac without need to change ACLiC linking flags

core/base/src/TSystem.cxx
```diff
@@ -3828,6 +3828,10 @@ int TSystem::CompileMacro(const char *filename, Option_t *opt,
          ForeachSharedLibDep(library, CollectF);
 
          TString relink_cmd = cmd.Strip(TString::kTrailing, ';');
+#ifdef R__MACOSX
+         // Allow linking to succeed despite the missing symbols.
+         relink_cmd.ReplaceAll("-dynamiclib", "-dynamiclib -Wl,-w -Wl,-undefined,dynamic_lookup");
```
pcanal, line 3833: This should not be necessary and should be a user selected choice.

devajithvs:
> #19747. solves problem on Windows. So it also should work on Mac without need to change ACLiC linking flags

Thanks! I’m closing this in favour of that!

PR #19748, opened 2025-08-26
[ntuple] add streaming vector tutorial (v2)

jblomer:
Replaces #17139 with comments in that PR incorporated.

### DIFF
--- a/tutorials/io/ntuple/ntpl016_streaming_vector.C
+++ b/tutorials/io/ntuple/ntpl016_streaming_vector.C
@@ -1,0 +1,199 @@
+/// \file
+/// \ingroup tutorial_ntuple
+///
+/// Example of a streaming vector: a special purpose container that reads large vectors piece-wise.
+///
+/// \macro_code
+///
+/// \date November 2024
+/// \author Peter van Gemmeren, the ROOT Team
+
+#include <ROOT/RNTupleModel.hxx>
+#include <ROOT/RNTupleReader.hxx>
+#include <ROOT/RNTupleReadOptions.hxx>
+#include <ROOT/RNTupleRange.hxx>
+#include <ROOT/RNTupleTypes.hxx>
+#include <ROOT/RNTupleView.hxx>
+#include <ROOT/RNTupleWriter.hxx>
+
+#include <TRandom3.h>
+
+#include <cstdint>
+#include <iostream>
+#include <vector>
+#include <utility>
+
+constexpr char const *kFileName = "ntpl016_streaming_vector.root";
+constexpr char const *kNTupleName = "ntpl";
+constexpr char const *kFieldName = "LargeVector";
+constexpr unsigned int kNEvents = 10;
+constexpr unsigned int kNElementsPerVector = 1000000;
+
+// Create an RNTuple with a single vector field. Every entry contains a large vector of random integers.
+// The vector should be seen as too large to be held entirely in memory during reading.
+void CreateRNTuple()
+{
+   auto model = ROOT::RNTupleModel::Create();
+   auto ptrLargeVector = model->MakeField<std::vector<std::uint32_t>>(kFieldName);
+   auto writer = ROOT::RNTupleWriter::Recreate(std::move(model), kNTupleName, kFileName);
+
+   auto prng = std::make_unique<TRandom3>();
+   prng->SetSeed();
+
+   for (ROOT::NTupleSize_t i = 0; i < kNEvents; i++) {
+      ptrLargeVector->clear();
+      for (std::size_t j = 0; j < kNElementsPerVector; j++)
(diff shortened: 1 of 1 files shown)


tutorials/io/ntuple/ntpl016_streaming_vector.C
```diff
@@ -0,0 +1,191 @@
...
+ */
+
+// The StreamingVectorView class allows iteration over an RNTuple on-disk vector of element type T.
+// Unlike an std::vector, this class does not provide random-access but only allows to iterate the data elements
+// from beginning to end.
+// Internally, it uses an RNTupleCollection view and an item view to load chunks of the vector elements into memory,
+// so that never the entire vector needs to stay in memory.
+// Note that we don't need to implement loading chunks of data explicitly. Simply by asking for a single vector element
+// at every iteration step, the RNTuple views will take care of keeping only the currently required data pages
+// in memory.
```
silverweed: It would be helpful here to say how one can tune the maximum amount of memory used by the StreamingVector

tutorials/io/ntuple/ntpl016_streaming_vector.C
```diff
@@ -0,0 +1,191 @@
...
+#include <cstdint>
+#include <iostream>
+#include <vector>
+#include <utility>
+
+constexpr char const *kFileName = "ntpl015_streaming_vector.root";
+constexpr char const *kNTupleName = "ntpl";
+constexpr char const *kFieldName = "LargeVector";
+constexpr unsigned int kNEvents = 10;
+constexpr unsigned int kVectorSize = 1000000;
```
hahnjo: `alma9 modules_off runtime_cxxmodules=Off` is slightly unhappy about this line because `TGeometry.h` defines `const Int_t kVectorSize = 3;` I guess we have to rename the constant here...

tutorials/io/ntuple/ntpl016_streaming_vector.C
```diff
@@ -0,0 +1,191 @@
...
+#include <ROOT/RNTupleWriter.hxx>
+
+#include <TRandom3.h>
+
+#include <cstdint>
+#include <iostream>
+#include <vector>
+#include <utility>
+
+constexpr char const *kFileName = "ntpl015_streaming_vector.root";
```
hahnjo:
```suggestion
constexpr char const *kFileName = "ntpl016_streaming_vector.root";
```

tutorials/io/ntuple/ntpl016_streaming_vector.C
```diff
@@ -0,0 +1,191 @@
...
+   // The index of the entry from which the vector should be read
+   ROOT::NTupleSize_t fEntry{0};
+   // The size of the collection in fEntry
+   ROOT::NTupleSize_t fSize{0};
+
+public:
+   // A lightweight iterator used in StreamingVectorView::begin() and StreamingVectorView::end().
+   // Used to iterate over the elements of an RNTuple on-disk vector for a certain entry.
+   // Dereferencing the iterator returns the corresponding value of the item view.
+   class iterator {
```
hahnjo: Should probably have the usual iterator `using` definitions, for `std::iterator_traits`

tutorials/io/ntuple/ntpl016_streaming_vector.C
hahnjo: I think the important point is to turn off the cluster cache, otherwise we will load entire clusters anyway. Beyond that, the memory consumption should be that of a page, which is determined by the input file. The reading code doesn't have much control over it...

hahnjo (approved):
Thanks for the other fixes to `RNTupleLocalRange` and `RNTupleCollectionView` - we may consider backporting these to the next release of v6.36. Some comments inline for the `ntpl016_streaming_vector.C` tutorial.

tutorials/io/ntuple/ntpl016_streaming_vector.C
pcanal: Don't we have (yet) the feature of 'only' loading the active columns in the cache?  (Reading the column ahead of time through the cache would allow to reduce the number of transactions with a remote server).

hahnjo: We only load active columns ("entire clusters" was maybe not 100% precise), but the user still has no control over how much data that actually is.

pcanal: And indeed if the cluster/one entry does not really fit in memory, we can't load the whole cluster even for just one column. To avoid having reads of only one (small) page, we would need to either configure to have 'large' page on the write side or upgrade the cache to be able to read partial clusters (if we don't have this feature yet - if I remember correct TTreeCache already does).

PR #19749, opened 2025-08-26
[build] Fix script for conda MacOS

vepadulano:
Note: this needs to be backported to 6.32 as well

### DIFF
--- a/cmake/scripts/std_modulemap_darwin_fix.py
+++ b/cmake/scripts/std_modulemap_darwin_fix.py
@@ -25,6 +25,13 @@ def main():
     output_path = sys.argv[1]
     sdk = get_sdk_path()
     cpp_modulemap = os.path.join(sdk, "usr/include/c++/v1/module.modulemap")
+    if not os.path.exists(cpp_modulemap):
+        # Try again if we are running a conda build. conda-forge ships the MacOS SDK stripped of libc++.
+        # Instead, the standard libraries are shipped as a separate package which must be declared as a build
+        # dependency and thus ends up in the $BUILD_PREFIX path. For explanation, see
+        # https://conda-forge.zulipchat.com/#narrow/channel/516932-compilers/topic/Organisation.20of.20libc.2B.2B.20and.20SDKs.20on.20MacOS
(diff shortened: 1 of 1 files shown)


cmake/scripts/std_modulemap_darwin_fix.py
```diff
@@ -25,6 +25,15 @@ def main():
...
     cpp_modulemap = os.path.join(sdk, "usr/include/c++/v1/module.modulemap")
+    if not os.path.exists(cpp_modulemap):
+        # Try again if we are running a conda build. conda-forge ships the MacOS SDK stripped of libc++
+        # Instead, the standard libraries are shipped as a separate package which must be declared as a build
+        # dependency and thus ends up in the $BUILD_PREFIX path. For explanation, see
+        # https://conda-forge.zulipchat.com/#narrow/channel/516932-compilers/topic/Organisation.20of.20libc.2B.2B.20and.20SDKs.20on.20MacOS
+        if (os.environ.get("CONDA_BUILD", "0") == "1") and (os.environ.get("BUILD_PREFIX", None) is not None):
+            cpp_modulemap = os.path.join(os.environ["BUILD_PREFIX"], "include/c++/v1/module.modulemap")
+        else:
+            raise FileNotFoundError(f"Cannot find libc++ modulemap at {cpp_modulemap}")
```
devajithvs: This might be redundant as we already have the check below. 

I was also wondering if there is a way to detect that we're in a conda build and maybe modify `get_sdk_path` to get the right sdk. But if not, LGTM

devajithvs (approved):

cmake/scripts/std_modulemap_darwin_fix.py
vepadulano: Thanks for the catch, indeed the `raise` is redundant I will remove it!

>  maybe modify get_sdk_path to get the right sdk.

`get_sdk_path` is working perfectly well, also in the conda build. The real issue is that the SDK shipped by conda is stripped explicitly of any and all C and C++ headers, libraries *and* modulemaps. Those are shipped separately (the C++ headers,libraries and modulemaps via the `libcxx`) conda package and thus even if `get_sdk_path` returns the correct SDK path, it's not the path that contains the modulemap :)

Issue #16488, opened 2024-09-20
WebGUI doesn't work with snap sandboxing


Axel-Naumann:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Hi,

starting the webgui (`new TBrowser`) opens a chromium window which complains:
```
Your file couldn’t be accessed
It may have been moved, edited, or deleted.
ERR_FILE_NOT_FOUND
```
I guess this corresponds to the following process:
```
axel       17508    3604  4 15:51 pts/0    00:00:01     /snap/chromium/2950/usr/lib/chromium-browser/chrome --password-store --disable-features=TFLiteLanguageDetectionEnabled --window-size=1200,700 --new-window --app=file:///tmp/root_start_t9pNYd.html
```

Note that this is a snap package; the browser might not be allowed to access files in `/tmp` as per https://forum.snapcraft.io/t/can-browsers-open-local-files/3016

(*I* can work around this, but I thought it'd still be worthwhile to report this.)

Cheers, Axel

### Reproducer

`new TBrowser` with web gui.

### ROOT version

master

### Installation method

source

### Operating system

Ubuntu 24.04

linev:
Hi Axel,

Can you try to add `--allow-file-access-from-files` to the command line?
It can be done with the following entry in .rootrc file:

```
WebGui.ChromeInteractive: $prog $geometry --new-window  --allow-file-access-from-files --app='$url'  &
```

ferdymercury:
> WebGui.ChromeInteractive: $prog $geometry --new-window  --allow-file-access-from-files --app='$url'  &

@linev I confirm that this solves the issue.

Question: should this be documented?
Or should we just close this issue as Won't fix, considering that it's not a safe flag?
See https://stackoverflow.com/a/33801287/7471760

I think a better solution would be to allow to define a Variable in the .rootrc file,
something like WebGui.TempFolder, which is /tmp by default, but can be modified to /home/user/mytmp so that it works fine with snap sandboxing.

linev:
That is value for you in `WebGui.Chrome` in `etc/system.rootrc` file?
I already implement workaround for snap chromium - if its name "/snap/bin/chromium".

See https://github.com/root-project/root/pull/18311

Probably another name variants can be tested.

And you always can redefine shell variable `TMPDIR` before running ROOT.

ferdymercury:
> hat is value for you in `WebGui.Chrome` in `etc/system.rootrc` file?

It's empty.

[...]

PR #19757, opened 2025-08-27
[skip-ci] document workaround for snap sandboxing and webgui

ferdymercury:
Fixes https://github.com/root-project/root/issues/16488

### DIFF
--- a/core/base/src/TROOT.cxx
+++ b/core/base/src/TROOT.cxx
@@ -2817,7 +2817,7 @@ void TROOT::SetBatch(Bool_t batch)
 /// `webdisplay` parameter may contain:
 ///
 ///  - "firefox": select Mozilla Firefox browser for interactive web display
-///  - "chrome": select Google Chrome browser for interactive web display. Can also be set to "chromium-browser"
+///  - "chrome": select Google Chrome browser for interactive web display. Can also be set to "chromium"
 ///  - "edge": select Microsoft Edge browser for interactive web display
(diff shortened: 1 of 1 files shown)


linev:
It is not proper place for such documentation.

Either one  add such remark in TROOT::SetWebDisplay documentation.

Or in RWebWindowsManager::ShowWindow method

In ideal case one need to create documentation for  `RWebDisplayHandle::Display` method
which implements full logic of starting web browsers.

ferdymercury:
> It is not proper place for such documentation.

Moved as suggested, thanks.

core/base/src/TROOT.cxx
```diff
@@ -2817,7 +2817,7 @@ void TROOT::SetBatch(Bool_t batch)
 /// `webdisplay` parameter may contain:
 ///
 ///  - "firefox": select Mozilla Firefox browser for interactive web display
-///  - "chrome": select Google Chrome browser for interactive web display
+///  - "chrome": select Google Chrome browser for interactive web display. Can also be set to "chromium-browser"
```
linev: Actually "chrome" and "chromium" are equivalent.
"chromium-browser" considered as "unknown" type of browser and handled in very generic way

linev:
```suggestion
///  - "chrome": select Google Chrome browser for interactive web display. Can also be set to "chromium"
```

ferdymercury:
Thanks! (Please squash on merge)

Issue #19076, opened 2025-06-18
[roottest] test `roottest-root-io-evolution-fixarr2` fails on some Mac


linev:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

After https://github.com/root-project/root/pull/19012 will be merged,
large `roottest/root/io/evolution` test will be split on many small tests.
One of such tests `roottest-root-io-evolution-fixarr2` crashes on some Mac platforms and therefore was disabled there.
Stack trace looks like:
```
 [/usr/lib/system/libsystem_platform.dylib] _sigtramp (no debug info)
  [/Users/sftnight/ROOT-CI/build/lib/libRIO.so] TFile::Close(char const*) /Users/sftnight/ROOT-CI/src/io/io/src/TFile.cxx:1036
  [/Users/sftnight/ROOT-CI/build/lib/libRIO.so] TFile::~TFile() /Users/sftnight/ROOT-CI/src/io/io/src/TFile.cxx:580
  [/Users/sftnight/ROOT-CI/build/lib/libRIO.so] TFile::~TFile() /Users/sftnight/ROOT-CI/src/io/io/src/TFile.cxx:569
  [/Users/sftnight/ROOT-CI/build/roottest/root/io/evolution/fixarr2_C.so] read(char const*) (no debug info)
```

### Reproducer

One need to remove `if(NOT APPLE)` from cmake file and then:

`ctest -R roottest-root-io-evolution-fixarr2`

### ROOT version

master

### Installation method

build from source

### Operating system

Mac

dpiparo:
@linev I cannot reproduce: what mac platform exactly?

linev:
I saw failure on Mac15 platform. See here:
https://github.com/root-project/root/actions/runs/15714091698/job/44279370611?pr=19012#step:6:8703

But first PR must be merged, only then fixarr2 will appear - but not on Mac.

This test was never executed with Makefiles.

linev:
And on my mac it crashes in 10% of runs

dpiparo:
It's great we have this increased coverage. Thanks for that. We need the fix to be one commit of the PR that enables the test, right?

linev:
For the moment test is disabled like:

```
if(NOT APPLE)
  ROOTTEST_ADD_TEST(fixarr2
                    MACRO fixarr2.C+
                    OUTREF fixarr2.ref
                    LABELS longtest io
[...]

PR #19760, opened 2025-08-27
[test] make fixarr3 and fixarr2 non-concurrent

ferdymercury:
Fixes https://github.com/root-project/root/issues/19076

This way they are executed one after the other, as they were in the original Makefile.

One hypothesis is that, even in "READ" mode, same ROOT file accessed concurrently might interfere. see eg https://github.com/root-project/root/pull/18954
due to TFile::SysClose or so.

Not sure if an alternative approach would be to read without global registration or so?

### DIFF
--- a/roottest/root/io/evolution/CMakeLists.txt
+++ b/roottest/root/io/evolution/CMakeLists.txt
@@ -139,20 +139,18 @@ ROOTTEST_ADD_TEST(fixarr1
                   FIXTURES_REQUIRED root-io-evolution-fixarr1-fixture
                   FIXTURES_SETUP root-io-evolution-fixarrfile-fixture)
 
-# tests fails on some Mac platforms generating stacktrace
-if(NOT APPLE)
-  ROOTTEST_ADD_TEST(fixarr2
-                    MACRO fixarr2.C+
-                    OUTREF fixarr2.ref
-                    LABELS longtest io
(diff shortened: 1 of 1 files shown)


linev:
It would be really strange if concurrent reading makes problems

ferdymercury:
You mentioned that if failed 10% of the time on your Mac. Does the percentage change after adding this fix ?
Right now mac13, mac15 and mac15 went well.

> It would be really strange if concurrent reading makes problems

Yeah, weird. Maybe the problem is the "delete" part after GetObject?

linev:
Last days I see no any failures for these tests.

And any GetObject problems cannot affect other process which only reads same file.

ferdymercury:
> Last days I see no any failures for these tests.

even after reenabling fixarr2 ?

I am exploring that in the CI, see https://github.com/root-project/root/pull/19762

ferdymercury:
In the meantime, while the underlying issue is understood: isn't it better to just reenable this test in serial mode as I propose here, rather than not running it at all?

linev:
I checked again on Mac platform where `fixarr2` test fails.
It also fails when `fixarr3` completely disabled. 
So there is no sense to make them sequential - `fixarr2` simply has problems.

ferdymercury:
Thanks for checking!

PR #19767, opened 2025-08-27
[Python][RDF] Extend FunctionJitter with newest supported numba jitting types

siliataider:
# This Pull request:

## Changes or fixes:
This is a follow-up to:
-  [Python][RDF] Add support for (some) root classes in numba jitted functions #19697 
-  [Python][RDF] Add support for std::vector and std::array in numba jitted functions #19487 

This PR extends the `FunctionJitter` class to call the `Numba.Declare` decorator with the right list of arguments, including the newest supported types in case the user doesn't explicitly specify the input and output types of the function to jit.
This is no longer limited to fundamental types and RVecs of fundamental types.

## Checklist:

- [x] tested changes locally


vepadulano (approved):
Great, thanks!

PR #19769, opened 2025-08-28
[CMake] Build `rootcint` and `genreflex` as separate targets

guitargeek:
This makes the CMake code more robust.

Right now, we use some `install(CODE "execute_process(COMMAND ln -f ...`
solution on unix to install `rootcint` and `genreflex`. This does not
work in all cases, either because of the usage of `\$ENV{DESTDIR}` when
`DESTDIR` is not set, or because hard links are not allowed.

Always copying `rootcling` - already in the build tree - would avoid
that problem, but by copying we risk sidestepping the CMake mechanisms
to set the RPath correctly when installing the copies, which are not
actual targets.

To make makes things simpler and more robust, this commit suggests to
build the `rootcing` and `genreflex` executables as separate targets
from the same source. The cost is very little cost in memory
(`rootcling` is only 31K, so copying two times only increases the size
of ROOTs `bin` directory by 1.5 %) and little in compile time (the extra
compile time is less than a second, not noticable in parallel builds).

### DIFF
--- a/main/CMakeLists.txt
+++ b/main/CMakeLists.txt
@@ -94,55 +94,33 @@ set_source_files_properties(src/rootcling.cxx PROPERTIES
   VISIBILITY_INLINES_HIDDEN "ON"
 )
 
-ROOT_EXECUTABLE(rootcling src/rootcling.cxx LIBRARIES RIO Cling Core Rint)
+set(rootcling_exe_names rootcling genreflex rootcint)
 
-# rootcling includes the ROOT complex header which would build the complex
-# dictionary with modules. To make sure that rootcling_stage1 builds this
-# dict before we use it, we add a dependency here.
-add_dependencies(rootcling complexDict)
+foreach(exe_name IN LISTS rootcling_exe_names)
+  ROOT_EXECUTABLE(${exe_name} src/rootcling.cxx LIBRARIES RIO Cling Core Rint)
 
-target_include_directories(rootcling PRIVATE
-        ${CMAKE_SOURCE_DIR}/core/metacling/res
-        ${CMAKE_SOURCE_DIR}/core/dictgen/res
-        ${CMAKE_SOURCE_DIR}/io/rootpcm/res)
-set_property(TARGET rootcling PROPERTY ENABLE_EXPORTS 1)
-if(WIN32)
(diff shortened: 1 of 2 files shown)


guitargeek:
To hide whitespace changes for easier review:
  * https://github.com/root-project/root/pull/19769/files?diff=unified&w=1

main/CMakeLists.txt
```diff
@@ -94,52 +94,30 @@ set_source_files_properties(src/rootcling.cxx PROPERTIES
...
+  if(WIN32)
+    set_target_properties(${exe_name} PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS 1)
+    set_property(TARGET ${exe_name} APPEND_STRING PROPERTY LINK_FLAGS " -STACK:4000000")
+  endif()
+endforeach()
 
-ROOT_EXECUTABLE(rootls src/rootls.cxx LIBRARIES RIO Tree Core Rint ROOTNTuple)
+# To inherit the dependencies from rootcling
+add_dependencies(genreflex rootcling)
+add_dependencies(rootcint rootcint)
```
vepadulano, line 120: Is this maybe?

```suggestion
add_dependencies(rootcint rootcling)
```

guitargeek: Oh... Yes thanks!

pcanal:
> The cost is ... little cost ...

Side note:  We do not support static library builds but we have had request for it, this would (silently) make them [avoidably] (even) bigger.

pcanal:
> when DESTDIR is not set

I assume this is solvable by ... setting it :) (and/or handling the case more explicitly).

So the bigger hurdle is:

> or because hard links are not allowed.

Where/when are hard links not allowed?

guitargeek:
Also Python wheels (I can look up the source again when on the computer)

guitargeek:
> > The cost is ... little cost ...
> 
> Side note: We do not support static library builds but we have had request for it, this would (silently) make them [avoidably] (even) bigger.

We are far from this anyway. Progress on static builds is tracked in this issue now:
https://github.com/root-project/root/issues/19944

If we have to provide static builds, the situation with the ROOT executables can be revisited.

Issue #19770, opened 2025-08-28
`RooWorkspace` return values unintuitive?


Zehvogel:
In some of the methods of `RooWorkspace,` e.g. `import`, it is documented that they return `true` when there was an error and `false` otherwise. A practice that I would guess was inherited from the usual meaning of a non-zero return value indicating an error and `0` being equal to `false`. There are also some methods (like `writeToFile`) that have no internal error handling and always return `false`.

https://github.com/root-project/root/blob/df80b8e2c42cec907af63082852858b0709549b2/roofit/roofitcore/src/RooWorkspace.cxx#L2042-L2047

In most circumstances, this does not matter as most people probably never check these return values. However, in today's world of doing analysis in Jupyter notebooks, the return value of a statement on the last line of a cell gets displayed automatically. Writing a workspace to a file and then receiving the message `false` like below makes at least me believe that something went wrong, and I start looking for the non-existent mistake.

<img width="505" height="158" alt="Image" src="https://github.com/user-attachments/assets/799ab709-286f-4c4a-accb-6a352c92d2ac" />

Maybe this behaviour could be improved?

ferdymercury:
Thanks for the report!

A fix for writeToFile: https://github.com/root-project/root/pull/19771

For Import, since it's documented, I'd say it's too late to change it (people might rely on it?). So the obvious workaround would be that you write in your Jupyter Notebook:

`w.Import(model) == 0`

PR #19771, opened 2025-08-28
[RF] proper error handling and return value when file-write

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/19770

### DIFF
--- a/roofit/roofitcore/src/RooWorkspace.cxx
+++ b/roofit/roofitcore/src/RooWorkspace.cxx
@@ -263,6 +263,7 @@ RooWorkspace::~RooWorkspace()
 /// Import a RooAbsArg or RooAbsData set from a workspace in a file. Filespec should be constructed as "filename:wspacename:objectname"
 /// The arguments will be passed to the relevant import() or import(RooAbsData&, ...) import calls
 /// \note From python, use `Import()`, since `import` is a reserved keyword.
+/// \return due to historical reasons: false (0) on success and true (1) on failure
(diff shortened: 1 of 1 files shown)


roofit/roofitcore/src/RooWorkspace.cxx
```diff
@@ -2038,12 +2045,15 @@ bool RooWorkspace::cd(const char* path)
...
 bool RooWorkspace::writeToFile(const char* fileName, bool recreate)
 {
-  TFile f(fileName,recreate?"RECREATE":"UPDATE") ;
-  Write() ;
-  return false ;
+  TFile f(fileName, recreate ? "RECREATE" : "UPDATE");
+  if (f.IsZombie())
+    return false;
+  auto bytes = Write();
+  return bytes > 0;
```
pcanal: Should we take this opportunity to 'easily' add access to writing remote files?
```suggestion
  std::unique_ptr<TFile> f{ TFile::Open(fileName, recreate ? "RECREATE" : "UPDATE") };
  if (!f || f->IsZombie())
    return false;
  auto bytes = Write();
  return bytes > 0;
```

Zehvogel:
Thx for also adding all the additional documentation! I did not realise before this that there are even some methods in there that return `true` and `false` the other way around :D

guitargeek (approved):
Thank you very much! These fixes are always appreciated :+1:

Issue #16981, opened 2024-11-19
Many roottest tests fail if you build ROOT with `-DNDEBUG` in `CXXFLAGS`


amadio:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

Example test failure output from roottest:
```
-- BEGIN OUTDIFF OUTPUT --
--- /srv/root/src/roottest/root/aclic/load/execautoload.ref	Tue Mar 17 09:11:36 2020
+++ /srv/root/roottest_build/root/aclic/load/autoload.log	Tue Nov 19 16:37:44 2024
@@ -1 +1,7 @@
 
+ROOT_cli_0:1:9: warning: 'NDEBUG' macro redefined [-Wmacro-redefined]
+#define NDEBUG
+        ^
+<command line>:5:9: note: previous definition is here
+#define NDEBUG 1
+        ^

-- END OUTDIFF OUTPUT --
CMake Error at /usr/lib64/cmake/ROOT/RootTestDriver.cmake:259 (message):
  compare 'stdout' error: 1
```

### Reproducer

Build ROOT with `-DNDEBUG` added to `CXXFLAGS`, build roottest against it and run the tests.

### ROOT version

```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.32.08                        https://root.cern |
  | (c) 1995-2024, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Nov 19 2024, 13:10:09                 |
  | From tags/6-32-08@6-32-08                                        |
  | With                                                             |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------
```

### Installation method

emerge

### Operating system

Gentoo Linux

### Additional context

I installed ROOT in the system, built with `CXXFLAGS=-O2 -g -DNDEBUG -fno-omit-frame-pointer -pipe`, then built and ran roottest tests against it, result is:
```
51% tests passed, 547 tests failed out of 1117
```
Almost all of them due to the macro redefinition.

Maybe replacing `#define NDEBUG` with
```cpp
#ifndef NDEBUG
#define NDEBUG
#endif
```
everywhere might fix it. However, LLVM does add `-UNDEBUG` in their compilation, so it might not be so simple.

dpiparo:
[...]

PR #19775, opened 2025-08-29
[test] fix potential clash with predefined NDEBUG flag

ferdymercury:
Fixes https://github.com/root-project/root/issues/16981
(I did not test though this, I just grepped where it was defined, so not sure if it fixes all cases).

@amadio do you find other tests with current master that fail apart from the fixed one?


guitargeek:
What happens when you remove the workaround with defining `NDEBUG 1` before including the assert header completely? There is at least a chance that it's not needed anymore... Can you try this in a CI run, please? The workaround was introduced 12 years ago:

https://github.com/root-project/roottest/blame/master/root/meta/genreflex/ROOT-5768/PyCool.h

ferdymercury:
> Can you try this in a CI run, please?

https://github.com/root-project/root/pull/19792

guitargeek:
Superseded by https://github.com/root-project/root/pull/19792

PR #19779, opened 2025-08-30
Revert "[CMake] Never force `RUN_SERIAL` in roottest when using Ninja"

hahnjo:
This reverts commit c8d6ef5f16553a81c6596815115c83e3053f3f16, again: The same change was already attempted in commit e8519e4a80 ("[roottest] Remove sequential compilation with Ninja.") and reverted in bab6bef828 which also holds some explanations and pointers why Ninja cannot be launched concurrently.


guitargeek (approved):
Thanks! I was only checking the git history of the `roottest` repository, so I missed the reverts that already happened in this repository. Sorry!

If you want to revert this again why not. But still, I don't see the advantage of this revert and wouldn't do it.

  * Having this `RUN_SERIAL` slows down things so much that ninja becomes unusable for development purposes when you like to validate your changes by running `roottest`. And you can't work around it without changing the source code. Or you see a way?

* The problem that you describe with doing an incremental build when running ctest makes sense, but it can easily be avoided by always building before running ctest, like we do in the CI, so ctest in run on a "clean" state

If you really want that we support this pattern of doing `ctest` straight from a dirty source directory, I'm sure we can find another solution without compromising parallel testing when using Ninja.

cmake/modules/RootMacros.cmake
```diff
@@ -2367,6 +2367,9 @@ macro(ROOTTEST_COMPILE_MACRO filename)
     set_property(TEST ${COMPILE_MACRO_TEST} PROPERTY FAIL_REGULAR_EXPRESSION "Warning in")
   endif()
   set_property(TEST ${COMPILE_MACRO_TEST} PROPERTY ENVIRONMENT ${ROOTTEST_ENVIRONMENT})
+  if(CMAKE_GENERATOR MATCHES Ninja AND NOT MSVC)
+    set_property(TEST ${COMPILE_MACRO_TEST} PROPERTY RUN_SERIAL true)
+  endif()
```
guitargeek, line 2372:
```suggestion
```

I think this is already not needed: the tests build a target that just calls a custom command, so they should have no dependencies anyway. But maybe it still triggers a re-configuration if the `CMakeLists` are changed...

guitargeek: Let's see how it goes:
  * https://github.com/root-project/root/pull/19780

guitargeek (changes requested):
Let's keep not bring this back for the `ROOTTEST_COMPILE_MACRO` at least, because there it's easy to avoid the problem. For the other cases, we can see later how building CMake targets can be avoided during test time.

hahnjo:
> Having this `RUN_SERIAL` slows down things so much that ninja becomes unusable for development purposes when you like to validate your changes by running `roottest`. And you can't work around it without changing the source code. Or you see a way?

As I already wrote in https://github.com/root-project/root/pull/19766#issuecomment-3239178119, Stephan has been working on https://github.com/root-project/root/pull/19702 which would "[prevent] only concurrent Ninja invocations, but otherwise [increase] parallelism".

> The problem that you describe with doing an incremental build when running ctest makes sense, but it can easily be avoided by always building before running ctest, like we do in the CI, so ctest in run on a "clean" state

Most of the times yes, and that's my standard procedure (`ninja && ctest -j12`). But it happens to me every now and then that something decides it needs to run `cmake` again, even though `ninja` went through.

> If you really want that we support this pattern of doing `ctest` straight from a dirty source directory, I'm sure we can find another solution without compromising parallel testing when using Ninja.

I don't want to actively support it, but as written above it sometimes happens spuriously. In addition, it can happen by mistake and it is really frustrating to end up with a fully corrupt build directory. I have this right now in my main build directory, crashing with LLVM assertions due to corrupt files, and even I couldn't figure out which file I have to delete to "save" my directory. This is just a huge waste of time and hinders development.

cmake/modules/RootMacros.cmake
hahnjo: Maybe, maybe not. A passing CI is not a guarantee that a change in the build system is correct. In any case, this PR is a pure revert of commit c8d6ef5f16553a81c6596815115c83e3053f3f16 and not the place to discuss other changes.

PR #19780, opened 2025-08-30
[CMake] Don't wrap command in `ROOTTEST_COMPILE_MACRO` in CMake target

guitargeek:
Instead of running the `root ... <macro>.C+` build step as the command
associated with a custom CMake target, just run the command directly in
the test added with `add_test`.

This simplification is possible because for the custom CMake target, no
dependencies are ever specified. This makes building it not different
from just running the underlying command directly.

Do some further CMake code simplifications in a second commit (more detail in the commit description).

### DIFF
--- a/cmake/modules/RootMacros.cmake
+++ b/cmake/modules/RootMacros.cmake
@@ -2314,17 +2314,14 @@ macro(ROOTTEST_COMPILE_MACRO filename)
     list(APPEND RootMacroDirDefines "-e;#define ${d}")
   endforeach()
 
-  set(RootMacroBuildDefines
-        -e "#define CMakeEnvironment"
-        -e "#define CMakeBuildDir \"${CMAKE_CURRENT_BINARY_DIR}\""
-        -e "gSystem->AddDynamicPath(\"${CMAKE_CURRENT_BINARY_DIR}\")"
-        -e "gROOT->SetMacroPath(\"${CMAKE_CURRENT_SOURCE_DIR}\")"
-        -e "gInterpreter->AddIncludePath(\"-I${CMAKE_CURRENT_BINARY_DIR}\")"
-        -e "gSystem->AddIncludePath(\"-I${CMAKE_CURRENT_BINARY_DIR}\")"
-        -e "gSystem->SetBuildDir(\"${CMAKE_CURRENT_BINARY_DIR}\", true)"
-        ${RootMacroDirDefines})
-
-  set(root_compile_macro ${ROOT_root_CMD} ${RootMacroBuildDefines} -q -l -b)
+  set(root_compile_macro ${CMAKE_COMMAND} -E env
+      ROOT_LIBRARY_PATH="${CMAKE_CURRENT_BINARY_DIR}"
+      ROOT_INCLUDE_PATH="${CMAKE_CURRENT_BINARY_DIR}"
+      ${ROOT_root_CMD}
+      -e "gSystem->SetBuildDir(\"${CMAKE_CURRENT_BINARY_DIR}\", true)"
+      ${RootMacroDirDefines}
+      -q -l -b
+  )
 
   get_filename_component(realfp ${filename} ABSOLUTE)
   if(MSVC)
--- a/cmake/modules/RootMacros.cmake
+++ b/cmake/modules/RootMacros.cmake
@@ -2291,8 +2291,7 @@ endfunction(ROOTTEST_ADD_AUTOMACROS)
 #-------------------------------------------------------------------------------
 #
 # macro ROOTTEST_COMPILE_MACRO(<filename> [BUILDOBJ object] [BUILDLIB lib]
-#                                         [FIXTURES_SETUP ...] [FIXTURES_CLEANUP ...] [FIXTURES_REQUIRED ...]
-#                                         [DEPENDS dependencies...])
+#                                         [FIXTURES_SETUP ...] [FIXTURES_CLEANUP ...] [FIXTURES_REQUIRED ...])
 #
 # This macro creates and loads a shared library containing the code from
 # the file <filename>. A test that performs the compilation is created.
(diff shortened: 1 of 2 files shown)


cmake/modules/RootMacros.cmake
```diff
@@ -2436,7 +2436,7 @@ macro(ROOTTEST_GENERATE_DICTIONARY dictname)
   set(targetname_libgen ${dictname}libgen)
 
   add_library(${targetname_libgen} EXCLUDE_FROM_ALL SHARED ${dictname}.cxx)
-  set_property(TARGET ${executable} PROPERTY BUILD_WITH_INSTALL_RPATH OFF) # will never be installed anyway
+  set_property(TARGET ${targetname_libgen} PROPERTY BUILD_WITH_INSTALL_RPATH OFF) # will never be installed anyway
```
hahnjo: This change is unrelated to `ROOTTEST_COMPILE_MACRO` and I argue it should go into a separate PR.

hahnjo (commented):
> Wrapping to command inside a Cmake target doesn't seem to be necessary, because in the test, the target is forced to be built without dependencies anyway (e.g. `cmake --build . --target my_macro_test/fast --always-make` when using makefiles).

This is only partially true, for `Unix Makefiles`. With `Ninja`, we don't use `/fast` nor `--always-make` AFAICT

guitargeek:
The intention of the original macro is apparently to only build this one target that runs the command. The fact that this was not done correctly for Ninja doesn't mean that rebuilding all dependencies was the intended behavior. I suppose it was just not correctly implemented.

cmake/modules/RootMacros.cmake
guitargeek: Sure!

cmake/modules/RootMacros.cmake
```diff
@@ -2340,36 +2340,18 @@ macro(ROOTTEST_COMPILE_MACRO filename)
                             WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 
   if(ARG_DEPENDS)
-    set(deps ${ARG_DEPENDS})
+    message(WARNING "Specifying DEPENDS in ROOTTEST_COMPILE_MACRO has no effect.")
```
guitargeek: Note that this warning is never hit. Nowhere in the codebase to we specify `DEPENDS` in a `ROOTTEST_COMPILE_MACRO`. Together with the dependencies being ignored via the `\fast` target when using Makefiles, that makes it clear to me that it was never the intention to have dependencies for these macro tests.

The problem is that even with no dependencies, a re-configuration will still be triggered if the CMake code was changed, which is the reason thy the `RUN_SERIAL true` workaround was needed.

hahnjo:
In principle makes sense to me; maybe rephrase the first commit message because it still sounds wrong that dependencies would be ignored "anyway".

Also adding @hageboeck because it's related to https://github.com/root-project/root/pull/19702

cmake/modules/RootMacros.cmake
hageboeck: I would like to suggest removing `DEPENDS` from the allowed arguments altogether.

cmake/modules/RootMacros.cmake
```diff
@@ -2315,17 +2315,14 @@ macro(ROOTTEST_COMPILE_MACRO filename)
...
-
-  set(root_compile_macro ${ROOT_root_CMD} ${RootMacroBuildDefines} -q -l -b)
+  set(root_compile_macro ${CMAKE_COMMAND} -E env
+      ROOT_LIBRARY_PATH="${CMAKE_CURRENT_BINARY_DIR}"
+      ROOT_INCLUDE_PATH="${CMAKE_CURRENT_BINARY_DIR}"
+      ${ROOT_root_CMD}
+      -e "gSystem->SetBuildDir(\"${CMAKE_CURRENT_BINARY_DIR}\", true)"
+      ${RootMacroDirDefines}
+      -q -l -b
+  )
```
hageboeck, line 2324: I'm all for simplifying, but maybe a more detailed commit message would be nice.
So far, I managed to find out using grep that the two `#define`s aren't needed, but why is the MacroPath not needed?

hageboeck (commented):
I find the explanation confusing. I'll come discuss this in person.

cmake/modules/RootMacros.cmake
guitargeek: I have added this info to the message of the second commit, and also rewritten the message of the first commit to be more concise.

hageboeck (approved):
Much clearer, thank you!

roottest/scripts/build.C
```diff
@@ -75,9 +75,6 @@ void build(const char *filename, const char *lib = 0, const char *obj = 0, const
    }
    // fprintf(stderr,"from %s to %s\n",filename,fname.Data());
    int result = gSystem->CompileMacro(fname,"kc");
-#elif defined(CMakeEnvironment) && defined(CMakeBuildDir)
-   // fprintf(stderr,"CmakeBuildDir: %s filename: %s", CMakeBuildDir, filename);
-   int result = gSystem->CompileMacro(filename,"kc-", libname, CMakeBuildDir);
```
pcanal, line 80: This seems to be an unintentional behavior change.  If I read correctly, before this PR, `ROOTTEST_COMPILE_MACRO` was always setting `CMakeEnvironment` and `CMakeBuildDir` and thus always passing ` "kc-" `, while it now passes "only" ` "kc" ` thus removing the `-` and changing the output directory layout.

pcanal:
> Simplify ROOTTEST_COMPILE_MACRO to shorten test command
>    Add include path and library path via environment variables

How does switching from an argument to an environment variable concretely help?  If one ever tries to reproduce the failing compilation 'exactly', one now needs to copy paste (the extended version  of):
```
      ROOT_LIBRARY_PATH="${CMAKE_CURRENT_BINARY_DIR}"
      ROOT_INCLUDE_PATH="${CMAKE_CURRENT_BINARY_DIR}"
      ${ROOT_root_CMD}
      -e "gSystem->SetBuildDir(\"${CMAKE_CURRENT_BINARY_DIR}\", true)"
      ${RootMacroDirDefines}
      -q -l -b ....
```
instead of the extended version of:
```
      ${ROOT_root_CMD}
      -e "gSystem->AddDynamicPath(\"${CMAKE_CURRENT_BINARY_DIR}\")"
      -e "gInterpreter->AddIncludePath(\"-I${CMAKE_CURRENT_BINARY_DIR}\")"
      -e "gSystem->SetBuildDir(\"${CMAKE_CURRENT_BINARY_DIR}\", true)"
      ${RootMacroDirDefines}
      -q -l -b ... 
```

roottest/scripts/build.C
guitargeek: That was done on purpose to harmonize with the code path for Windows above, because specifying the output directory to `CMAKE_CURRENT_BINARY_DIR` was redundant with setting the working directory in the CMake command (`WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}`).

guitargeek:
> How does switching from an argument to an environment variable concretely help? If one ever tries to reproduce the failing compilation 'exactly', one now needs to copy paste (the extended version of):

That's true, the difference is marginal. But steering include and library paths via environment variables compared to C++ statements feels a bit cleaner to me, and also I think this was a good opportunity to test that the `ROOT_INCLUDE_PATH` and `ROOT_LIBRARY_PATH` environment variables work as intended. These are not covered anywhere else in the tests.

PR #19782, opened 2025-09-01
[tmva][sofie] Fix a bug in generating the code for LayerNormalization

lmoneta:
This is a first fix for LayerNorm. The code should be improved to be more efficient by avoiding broadcasting and combining the loops as it is done in the BatchNormalization layer


lmoneta:
Already implemented in a different PR

PR #19785, opened 2025-09-01
[RF][Docs] Remove a \see in the RooDataHist/Set documentation.

hageboeck:
Currently, the documentation seems to show an empty section, since a section title is followed by a \see command, which generates its own section.
It's better to replace it by the plain word "See xxx".

See the second section [here](https://root.cern/doc/master/classRooDataHist.html):
image


guitargeek (approved):
Thank you!

Issue #7311, opened 2021-03-01
Cannot build ROOT if some paths contain spaces


eguiraud:
- [X] Checked for duplicates

### Describe the bug

The problem  and steps to reproduce it are described [in this forum post](https://root-forum.cern.ch/t/trouble-installing-root-w-python3/43694/11).

eguiraud:
This is still an issue, there are some problems already at configuration time:

```
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 105: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 106: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 107: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 108: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 109: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 110: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 111: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 113: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 118: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 122: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 123: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 124: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 125: ${COMPILEDATA}.tmp: ambiguous redirect
/home/blue/ROOT/root sources/build/unix/compiledata.sh: line 126: ${COMPILEDATA}.tmp: ambiguous redirect
[...]

PR #19786, opened 2025-09-01
[cmake] avoid OS-dependent cat

ferdymercury:
# This Pull request:

## Changes or fixes:

since CMake 3.18 there is an inbuilt tool.

Helps towards https://github.com/root-project/root/issues/7311

Probably, the logic in the generation of the modulemap can be further optimized by avoiding that "intermediate copy" of unix/modulemap and doing all via "cat", right?

## Checklist:

- [x] tested changes locally

This PR fixes #


vgvassilev (approved):
Lgtm!

hageboeck:
@ferdymercury thanks, I guess this is a good simplification!

However, looking at the code, I think it can be made even simpler (not necessarily your job, though). Let me check if I can find a way to generate the file directly using `configure_file`.

ferdymercury:
Thanks, closing in favor of your nice compact solution!

PR #19790, opened 2025-09-01
[roottest] don't disable all hadd tests if pyroot is off

silverweed:


guitargeek (approved):
Very good, thanks!

vepadulano (approved):
LGTM Thanks!

PR #12292, opened 2023-02-12
Fix cpt build rules

vgvassilev:

PR #19796, opened 2025-09-01
[cling] Remove cpt.py cling packaging tool

ferdymercury:
It was very outdated and broken, with over 20 bug reports.
See https://github.com/root-project/root/pull/12292#issuecomment-3242171610

appveyor yml is also removed.

Now there is the llvm monorepo and updated build instructions, as well as a work-in-progress GitHub CI script


vgvassilev:
Can we make sure that no cpt is mentioned anywhere in the codebase?

ferdymercury:
> Can we make sure that no cpt is mentioned anywhere in the codebase?

Yes, I have done a grep and it's nowhere else to be found.

Issue #19798, opened 2025-09-01
Removed option `tbb` is still considered to enable secret optimization of `ROOT::gCoreMutex`


guitargeek:
### Description

A long time ago, ROOT had a `tbb` build option, but it was removed in the ROOT 6.08 development cycle in dd899a5cacbb from 2015, by @peremato.

However, some build configurations still use it, like CMSSW and Conda:
  * https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_15_1_X/master/root.spec#L78
  * https://github.com/conda-forge/root-feedstock/blob/main/recipe/build_root.sh#L342

That would not be a problem per se, but then, in 2021, @bendavid introduced a new optimization that is enabled only with the long-time removed `tbb` flag (or `builtin_tbb`), in 6c8b77d959:

  * https://github.com/root-project/root/pull/7260

So we ended up with a secret `tbb=ON` configuration that our CI doesn't test, and a CMake configuration option that has a pretty fundamental effect but is not documented.

Now, this comes back to bite us, because loading `libtbb` and especially `libtbbmalloc` when loading `libRIO` -> `libThread` -> `libtbb` can have unintended side effects. The `libRIO` is always loaded when the interpreter is initialized, and tbb then loads its custom allocator. I think this can result in mismatching allocation/deallocation, when ROOT is imported in Python after importing other Python modules that already used the same shared libraries as ROOT but with other allocators.

The example is this Conda issue here:
  * https://github.com/conda-forge/root-feedstock/issues/253

Both TensorFlow and ROOT use the `libzstd` support in LLVM to decompress C++ modules (I think), and switching allocators mid-way might cause the reported crash.

I don't know yet what's the best action here, but for sure we should not have an undocumented build option that changes the behavior of `ROOT::gCoreMutex`.

bendavid:
Well I guess the underlying problem here is bad symbol visibility behaviour from Tensorflow (we had tons of problems with this in our analysis code related to symbol conflicts with Eigen that we ended up patching out, but I haven't gotten around to upstreaming this yet: https://gitlab.cern.ch/bendavid/cmswmassdocker/-/blob/f987d251941408a4452f8e95272808d26024cfd1/build/tensorflow/symbol-visibility.patch )

[...]

PR #6919, opened 2020-12-06
Further improvements to reduce lock contention

bendavid:
Further reduction in lock contention when using RDataFrame with a large number of threads and/or files, by improving the global read write lock and migrating one hot spot in TBufferFile to use it.

PR #19799, opened 2025-09-01
[ci] Check what happens when building with `tbb=ON`

guitargeek:
As described in #19798, we have an old build option that was supposedly removed, but it is still considered in the implementation of `ROOT::gCoreMutex`.

It would be good to know what happens when testing ROOT with it (no differences are expected from the performance optimization with TBB, and actually it was always enabled when building with `builtin_tbb=ON` as well, as we do for macOS and Windows).

See also:
  * https://github.com/root-project/root/issues/19798


bendavid:
I'm confused, doesn't RecurseCountsThreadLocal still depend on tbb::enumerable_thread_specific in this case?

guitargeek:
Oh, pardon I forgot to remove the `tbb::enumerable_thread_specific<LocalCounts> fLocalCounts` data member! It's updated now.

bendavid:
So there's a maybe longstanding issue here related to https://github.com/root-project/root/commit/bd1894b2bc8306c789557ddf12266c9af6259046 and discussed in https://github.com/root-project/root/pull/6919#discussion_r537705393

Given that this is similar to the old implementation in https://github.com/root-project/root/blob/5028941cbe68d777d16691663a58165da14f01de/core/thread/inc/ROOT/TReentrantRWLock.hxx#L27-L86

guitargeek:
Oh! I see. Interesting

guitargeek:
I wonder how TBB avoids this problem that Philippe described...

pcanal:
> I wonder how TBB avoids this problem that Philippe described...

The difference between `UniqueLockRecurseCount` and `RecurseCounts` (and likely the TBB implementation) was switching from using `thread_local` (which, if I recall correctly, requires/required via a call to `tls_get_addr_tail` to indirectly take the same lock as ...something else maybe related to library loading or more likely static initialization) to only using `std::this_thread::get_id()`.

The real challenge with this area is that seeing it work in most (all?) cases does not tell us we don't have the problem ... as I was unable at the time to deterministically reproduce the issue (see log https://github.com/root-project/root/commit/bd1894b2bc8306c789557ddf12266c9af6259046).

At the very least, one should revert bd1894b (i.e. using the old RecurseCounts) to check if we see the problem.  If we see the problem then we have some degree of confidence that having the alternative code not failing means that they might not have the same problem.  If we do not see the problem then ... we can not tell :(

pcanal:
In [tlslock.tar.gz](https://github.com/user-attachments/files/22103891/tlslock.tar.gz) I got to the spirit of what I remember
```
. go
Starting execution
Start thread 1 workload
Start thread 2 workload
Thread 2 sleeps 2 seconds while holding the lock
Inner most counter for thread 0x16f7eb000 is 1
Thread 1 will request the lock
Thread 2 done sleeping and will initialize a thread_local
# Process is stuck forever.
```
on macos and linux.  The process is actually stuck in the ROOT lock ....

Updating the example by commenting out `ROOT::EnableThreadSafety();` in `main.cxx` shows a slightly different behavior but then again it is not using `gSystem` is a thread safe way

thread 2 is waiting for the lock:
```
#0  0x00007ffff3e873f0 in __lll_lock_wait () from /lib64/libc.so.6
#1  0x00007ffff3e8d582 in pthread_mutex_lock@@GLIBC_2.2.5 () from /lib64/libc.so.6
#2  0x0000000000404505 in __gthread_mutex_lock (__mutex=0x4081c0 <gMainMutex>) at /cvmfs/cms.cern.ch/el9_amd64_gcc11/external/gcc/11.4.1-30ebdc301ebd200f2ae0e3d880258e65/include/c++/11.4.1/x86_64-redhat-linux-gnu/bits/gthr-default.h:749
#3  0x0000000000404634 in std::mutex::lock (this=0x4081c0 <gMainMutex>) at /cvmfs/cms.cern.ch/el9_amd64_gcc11/external/gcc/11.4.1-30ebdc301ebd200f2ae0e3d880258e65/include/c++/11.4.1/bits/std_mutex.h:100
#4  0x00007ffff45374ad in innermost () at inner.cxx:21
```
while the other thread is waiting on the system lock
```
(gdb) thread 3
[Switching to thread 3 (LWP 4021866)]
#0  0x00007ffff3e873f0 in __lll_lock_wait () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff3e873f0 in __lll_lock_wait () from /lib64/libc.so.6
#1  0x00007ffff3e8d5dd in pthread_mutex_lock@@GLIBC_2.2.5 () from /lib64/libc.so.6
#2  0x00007ffff7fd3393 in _dl_open (file=0x7fffcc000c40 "/var/tmp/tlslock/libWithThreadLocal.so", mode=-2147483391,
    caller_dlopen=0x7fffe4cbf2be <cling::utils::platform::DLOpen(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >*)+30>,
    nsid=-2, argc=1, argv=0x7fffffffbc18, env=0x7fffec0226f0) at dl-open.c:819
```

it is not clear whether the example is too convoluted.

pcanal:
This second example does also show the essence of the problem without ROOT.  
[tlslock2.tar.gz](https://github.com/user-attachments/files/22105964/tlslock2.tar.gz)

Side note: I am guessing the example can be simply further (by removing one of the layer of libraries)

bendavid:
One thing I'm not sure I understood here.  Would it be sufficient to ensure that the thread_local variables are initialized during the loading of libThread (and/or in the subsequent creation of new therads)?  (as opposed to potentially later at the first call to GetLocal()?

pcanal:
> Would it be sufficient to ensure that the thread_local variables are initialized during the loading of libThread

it can not be done, can it?  The official `thread_local` variables need to be initialized for each (new) threads.

bendavid:
Yes, but in this case it could/should be initialized for all existing threads at library load time and for any additional threads immediately when the thread is created.

(As opposed to when execution reaches some particular local scope, I think as in 

https://en.cppreference.com/w/cpp/language/storage_duration.html#Static_block_variables

)

pcanal:
> Yes, but in this case it could/should be initialized for all existing threads

I can see how that would work for a static variable .. but I am confused on how it could work for a thread_local since there is an infinite (indeterminated) number of (upcoming) threads.

bendavid:
Right, for threads created after the library is loaded I would assume there are some cases where the thread_local variables would be initialized immediately upon the creation of the thread.  (but not necessarily for the thread_local block variable case I linked above)

pcanal:
> Right, for threads created after the library is loaded I would assume there are some cases where the thread_local variables would be initialized immediately upon the creation of the thread. (but not necessarily for the thread_local block variable case I linked above)

Humm ... intriguing ... We might indeed be able to use a 'thread_local` declared in the global scope rather than inside a function in order to avoid the issue.

bendavid:
Yes, or possibly as a static thread_local class data member

bendavid:
I looked into this a bit, and it seems  that the thread_local variables are ~always initialized lazily.

To avoid the original deadlock, it might be sufficient to just make sure that all calls to fRecurseCounts.GetLocal() in TReentrantRWLock are made before taking the mutex lock.

e.g. switch the order of these two lines, and similar elsewhere

https://github.com/root-project/root/blob/816001bde4e7b057d12aeb3cc9df20a089e90373/core/thread/src/TReentrantRWLock.cxx#L173-L175

guitargeek:
In the end, we just went with the tbb dependency:
  * https://github.com/root-project/root/pull/20564

PR #19800, opened 2025-09-02
[main] Fix a bit help message of root.

silverweed:
The description was trying to refer to a website but no URL was there. The message was also compactified a bit to save vertical space.

### DIFF
--- a/core/base/src/root-argparse.py
+++ b/core/base/src/root-argparse.py
@@ -2,9 +2,10 @@ import argparse
 
 def get_argparse():
 	parser = argparse.ArgumentParser(add_help=False, prog='root',
-	description = """ROOTs Object-Oriented Technologies.\n
-root is an interactive interpreter of C++ code. It uses the ROOT  framework.  For  more information on ROOT, please refer to\n
-An extensive Users Guide is available from that site (see below).
+	description = """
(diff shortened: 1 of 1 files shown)


ferdymercury (approved):
Thanks!

Not related to this PR:
I find root-argparse, related Python cmdlineOption helper scripts and generateHeader RootMacro (used for hadd, dictgen, root, ...) an unnecessary hetero-complexity, when there are ready-to-go header-only C++ libraries such as https://github.com/CLIUtils/CLI11 that could be linked via CMake without problems.

Also, there was some discussion about removing root wrapper and just redirecting to root.exe, in which case this help would have to be moved there, too.
see https://its.cern.ch/jira/browse/ROOT-10697

silverweed:
> I find root-argparse, related Python cmdlineOption helper scripts and generateHeader RootMacro (used for hadd, dictgen, root, ...) an unnecessary hetero-complexity

100% agree, I will see them gone at some point in the future.

Using a common C++ library to parse options already emerged some time ago, but there is the issue of executables using heterogeneous and non-standard styles for options (e.g. some executables accept single-dash long options, and some use weird conventions where the same flag means different things depending on the argument).
These can probably all be resolved, but needs to be thought through a bit.

I would also be in favor of removing the root wrapper, but again the implications needs to be considered first.

PR #19804, opened 2025-09-02
[CMake] Simplify writing of modulemap.

hageboeck:
Instead of a two-step process that writes the dynamic contents at CMake configure time, and a build-time step that concatenates the static and dynamic part of the modulemap, CMake's configure_file is used to directly write the full contents at configure time.

This is carrying the simplifications in #19786 even further.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -420,36 +420,14 @@ get_property(__allBuiltins GLOBAL PROPERTY ROOT_BUILTIN_TARGETS)
 add_custom_target(move_headers ALL DEPENDS ${__allHeaders} ${__allBuiltins} gitinfotxt)
 
 #---CXX MODULES-----------------------------------------------------------------------------------
-if(MSVC)
-  set(_os_cat "type")
-else()
-  set(_os_cat "cat")
-endif()
-cmake_path(CONVERT "${CMAKE_BINARY_DIR}/include/module.modulemap.extra" TO_NATIVE_PATH_LIST _from_native)
-cmake_path(CONVERT "${CMAKE_BINARY_DIR}/include/ROOT.modulemap" TO_NATIVE_PATH_LIST _to_native)
-
-add_custom_target(copymodulemap DEPENDS "${CMAKE_BINARY_DIR}/include/ROOT.modulemap")
-add_custom_command(
-		  OUTPUT "${CMAKE_BINARY_DIR}/include/ROOT.modulemap"
-                  DEPENDS cmake/unix/module.modulemap "${CMAKE_BINARY_DIR}/include/module.modulemap.extra"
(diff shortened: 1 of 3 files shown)


ferdymercury (approved):

CMakeLists.txt
```diff
@@ -421,36 +421,14 @@ get_property(__allBuiltins GLOBAL PROPERTY ROOT_BUILTIN_TARGETS)
...
-string(REPLACE ";" "" __modulemap_extra_content "${__modulemap_extra_content}")
-# Write module.modulemap.extra to a temporary file first, to not touch module.modulemap.extra
-# if it's unchanged.
-file(WRITE "${CMAKE_BINARY_DIR}/include/module.modulemap.extra.tmp" "${__modulemap_extra_content}")
-configure_file("${CMAKE_BINARY_DIR}/include/module.modulemap.extra.tmp"
-    "${CMAKE_BINARY_DIR}/include/module.modulemap.extra"
-    COPYONLY)
+string(REPLACE ";" "" ROOT_GENERATED_MODULEMAP_CONTENT "${__modulemap_extra_content}")
+# Now append the generated content to the original modulemap
+configure_file(${CMAKE_SOURCE_DIR}/cmake/unix/module.modulemap ${CMAKE_BINARY_DIR}/include/ROOT.modulemap)
```
vgvassilev, line 429: Would that work if somebody reconfigure with different options? Eg. `cmake -DTMVA=On` then `cmake -DTMVA=Off`.

hageboeck: Yes, it does! I tested it, and the CMAKE variable is filled into the file again

hageboeck: (and of course, when you reconfigure with different options, the time stamp is updated)

vgvassilev: Hm... okay. I tried that long time ago and failed to make it work but perhaps now we have all the dependencies tracked right...

hageboeck: What concerns the build system, the only dependencies that remain are `*.pcm -> ${BINDIR}/include/ROOT.modulemap`
The contents of the modulemap are (re-)generated during every CMake run, and if it comes out to be the same contents, the file remains untouched.

ferdymercury: Btw there is this comment here https://github.com/root-project/root/blob/e928d4210da6525f1447b3030000b33b6e9c79d0/core/base/src/TSystem.cxx#L3572 that might be relevant.

hageboeck: @vgvassilev is the comment above still relevant?

vgvassilev: Somewhat relevant but requires a lot of work. We could move all of the logic of modulemap creation to rootcling to merge the paths with Aclic. This however has other underwater rocks.

dpiparo (approved):

CMakeLists.txt
hageboeck: I'll resolve the discussion at this point, since it's not in the scope of this PR.

PR #19808, opened 2025-09-03
Revert "[NFC][cling] Add `CreateCI` to match upstream structure"

hahnjo:
This reverts commit 1a2ed14a72bc1f187e2ee40cdf544bae6b159077. It introduced a `stack-use-after-scope` because the `Compilation` has a reference to the `Driver` (on the stack). I tried moving back only these two variables and pass `CC1Args` to `CreateCI`, but that does not work either because it needs a `DiagnosticsEngine`. For now I don't see a good way out, so revert this refactoring to solve the concrete issue.


devajithvs (approved):
LGTM, thanks. Even if we could fix it moving variables to CreateCI, it would diverge from upstream structure and it's better to revert this and maybe think of something later to try and simplify CIFactory.

PR #19809, opened 2025-09-03
[CMake] Fix `asan` builds with Clang

hahnjo:
* `SearchInstalledSoftware` before adding linker flags, to avoid interference with `-z undefs`
 * Look for new `libclang_rt.asan.so` in architecture dependent directory

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,6 +113,14 @@ include(RootMacros)
 include(CheckAssembler)
 include(CheckIntrinsics)
 
+#---Enable testing------------------------------------------------------------------------------
+if(testing)
+  enable_testing()
+endif()
+
+#---Here we look for installed software and switch on and off the different build options-------
+include(SearchInstalledSoftware)
+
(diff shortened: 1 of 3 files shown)


hahnjo:
There is one more issue with `CppInterOp` shared libraries for testing, this needs https://github.com/compiler-research/CppInterOp/pull/706 upstream.

CMakeLists.txt
```diff
@@ -113,6 +113,15 @@ include(RootMacros)
...
 include(CheckIntrinsics)
 
+#---Enable CTest package -----------------------------------------------------------------------
+#include(CTest)
+if(testing)
+  enable_testing()
+endif()
+
+#---Here we look for installed software and switch on and of the different build options--------
+include(SearchInstalledSoftware)
```
hageboeck: Not related to the PR, but a good opportunity to clean up and fix a typo.
```suggestion
if(testing)
  enable_testing()
endif()

#---Here we look for installed software and switch on and off the different build options--------
include(SearchInstalledSoftware)
```

PR #19811, opened 2025-09-03
[Python] Fix RPATH info for ROOTPythonizations library

fuhlig1:
# This Pull request:

## Changes or fixes:

The RPATH was not properly added to the library info. The PR adds the needed information using the macro ROOT_APPEND_LIBDIR_TO_INSTALL_RPATH.
The  ROOTPythonizations library was the only shared library which missed the information.

## Checklist:

- [x ] tested changes locally

This PR fixes #

### DIFF
--- a/bindings/pyroot/pythonizations/CMakeLists.txt
+++ b/bindings/pyroot/pythonizations/CMakeLists.txt
@@ -182,6 +182,8 @@ else()
   target_link_libraries(${libname} PUBLIC -Wl,--unresolved-symbols=ignore-all)
 endif()
 
+ROOT_APPEND_LIBDIR_TO_INSTALL_RPATH(${libname} ${CMAKE_INSTALL_LIBDIR})
+
 target_include_directories(${libname}
     PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/inc>)
 


guitargeek:
Hi @fuhlig1, thanks for opening this PR! It's great you're trying things our with ROOT master and fix problem!

However, I'm not sure if this fix is correct. The correct rpath is already added to the pythonizations library here:
https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/CMakeLists.txt#L228
```cmake
if(NOT MSVC)
  # Make sure that relative RUNPATH to main ROOT libraries is always correct.
  ROOT_APPEND_LIBDIR_TO_INSTALL_RPATH(${libname} ${pymoduledir_install})
endif()

# Install library
install(TARGETS ${libname} EXPORT ${CMAKE_PROJECT_NAME}Exports
                            RUNTIME DESTINATION ${pymoduledir_install} COMPONENT libraries
                            LIBRARY DESTINATION ${pymoduledir_install}  COMPONENT libraries
                            ARCHIVE DESTINATION ${pymoduledir_install} COMPONENT libraries)
```

So the library gets installed into `${pymoduledir_install}`, and then we add the relative RPATH from `${pymoduledir_install}` to the `${CMAKE_INSTALL_LIBDIR}` in the `ROOT_APPEND_LIBDIR_TO_INSTALL_RPATH` macro. That this works is validated by our nightly Python wheels.

So why is it necessary for you to do `ROOT_APPEND_LIBDIR_TO_INSTALL_RPATH(${libname} ${CMAKE_INSTALL_LIBDIR})`, even thought the pythonizations library is installed to `${pymoduledir_install}` and not `${CMAKE_INSTALL_LIBDIR}`? What is your CMake config? Can you check the RPATH of the pythonization library in the install tree and let me know? Right now I don't understand why this PR is needed, and that needs to be understood first.

fuhlig1:
Hi,

thanks for the fast response. I stumbled over the inconsistency meanwhile myself. I was creating the patch for v6.36.04 and then added it to master without further checking. So probably the issue was already resolved in the upstream master.
I will check what was done in master.

I will close the issue.

guitargeek:
Thanks! Let me know if I can help by doing backports to 6.36

PR #19813, opened 2025-09-03
[cmake][win] Fix conda builds when using externals instead of builtins

bellenot:


tree/ntuple/test/CMakeLists.txt
```diff
@@ -78,7 +78,9 @@ if(MSVC)
 endif()
 ROOT_ADD_GTEST(rfield_vector rfield_vector.cxx LIBRARIES ROOTNTuple CustomStruct)
 
-ROOT_ADD_GTEST(ntuple_minifile ntuple_minifile.cxx LIBRARIES ROOTNTuple Physics Tree)
+if(NOT MSVC OR builtin_gtest)
```
pcanal, line 81: Why does this test fail on MSVC with an external gtest?

io/xmlparser/CMakeLists.txt
```diff
@@ -28,10 +28,16 @@ ROOT_STANDARD_LIBRARY_PACKAGE(XMLParser
     Core
 )
 
+if(MSVC AND NOT builtin_clang)
+  # fixes an issue when building with conda not finding iconv.h
```
pcanal: Consider adding a more explicit description of the issue being solved/worked-around.

tree/ntuple/test/CMakeLists.txt
bellenot: There is an undefined symbol, and if I use the proper flag to fix it, many (all?) the other tests fail with symbols already defined... But this is not a big issue, AFAIK the tests are not run in Conda (but I do...)

io/xmlparser/CMakeLists.txt
bellenot: Will try to come with something simple (it's Conda, so it's complicated/out of control)

bellenot: I fixed the issue outside

tree/ntuple/test/CMakeLists.txt
pcanal: See concrete proposal
```suggestion
if(NOT MSVC OR builtin_gtest)
  # As of 2025-09, the external gtest provided by Conda is not link compatible with ROOT.
  # As is this test fails with (at least) this symbols XYZ missing
  # Addressing locally this missing symbol leads to other symbols no longer be found in the other tests.
```
and replace XYZ by the symbol that you first saw missing.

bellenot: Done

Issue #19814, opened 2025-09-03
Regression in 6.34: segfault when using TTreeReader on partially initialized TChain


wdconinc:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Using `TTreeReader tree_reader(tree)` to define a TTreeReader on a TChain, and subsequently setting the range with `tree_reader.SetEntriesRange(0, tree->GetEntries())` fails with a segfault as of ROOT v6.34.
```
===========================================================
#8  0x00007ae157f5aae3 in TTree::GetReadCache(TFile*) const () from /opt/software/linux-x86_64_v2/root-6.34.08-igalfw34wnajhjntnzlwq6u546prgstg/lib/root/libTree.so.6.34.08
#9  0x00007ae157f64548 in TTree::GetReadCache(TFile*, bool) () from /opt/software/linux-x86_64_v2/root-6.34.08-igalfw34wnajhjntnzlwq6u546prgstg/lib/root/libTree.so.6.34.08
#10 0x00007ae157d777d1 in TTreeReader::Restart() () from /opt/software/linux-x86_64_v2/root-6.34.08-igalfw34wnajhjntnzlwq6u546prgstg/lib/root/libTreePlayer.so.6.34.08
#11 0x00007ae157d7be87 in TTreeReader::SetEntriesRange(long long, long long) () from /opt/software/linux-x86_64_v2/root-6.34.08-igalfw34wnajhjntnzlwq6u546prgstg/lib/root/libTreePlayer.so.6.34.08
#12 0x00007ae158a69144 in ?? ()
#13 0x00007ffe689d2890 in ?? ()
#14 0x00007ae157e1f640 in vtable for TTreeReader () from /opt/software/linux-x86_64_v2/root-6.34.08-igalfw34wnajhjntnzlwq6u546prgstg/lib/root/libTreePlayer.so.6.34.08
#15 0x0202400000000000 in ?? ()
#16 0x00005ab53c512700 in ?? ()
#17 0x0000000000000000 in ?? ()
===========================================================
```

### Reproducer

With `test.C`:
```cpp
#include <TChain.h>
#include <TTreeReader.h>

void test(std::string file) {

    auto tree = new TChain("events"); 
    tree->Add(file.c_str()); 
    TTreeReader tree_reader(tree); // !the tree reader

    //...

    tree_reader.SetEntriesRange(0, tree->GetEntries());

}
```
[...]

PR #19817, opened 2025-09-04
[treeplayer] prevent nullptr access in Restart

ferdymercury:
Fixes https://github.com/root-project/root/issues/19814

@wdconinc could you test on your side with this patch?

### DIFF
--- a/tree/treeplayer/test/regressions.cxx
+++ b/tree/treeplayer/test/regressions.cxx
@@ -285,7 +285,7 @@ TEST(TTreeReaderRegressions, UninitializedChain)
       TTreeReader reader(&ch);
       reader.SetEntriesRange(0, ch.GetEntries());
       EXPECT_EQ(reader.GetEntries(), 1);
-      TTreeReaderValue x(reader, brname);
+      TTreeReaderValue<int> x(reader, brname);
       EXPECT_TRUE(reader.Next());
       EXPECT_EQ(*x, refval);
    }


vepadulano (approved):
LGTM and thank you **so** much! Before merging, I will wait for @pcanal's review and for confirmation that this indeed fixes the user issue (although it already looks very promising with the unittest).

tree/treeplayer/src/TTreeReader.cxx
```diff
@@ -567,9 +567,14 @@ void TTreeReader::Restart()
    fProxiesSet = false; // we might get more value readers, meaning new proxies.
    fEntry = -1;
    if (const auto curFile = fTree->GetCurrentFile()) {
-      if (auto tc = fTree->GetTree()->GetReadCache(curFile, true)) {
-         tc->DropBranch("*", true);
-         tc->ResetCache();
+      if (!fTree->GetTree()) {
+         fTree->LoadTree(0);
```
pcanal, line 571: Shouldn't it be:
```suggestion
         fTree->LoadTree(lowest_entry_in_the_entry_range);
```
?

ferdymercury: I don't think so, because Restart is only called if reading from the first entry of the tree:

```
if (beginEntry - 1 < 0)
      // Reset the cache if reading from the first entry of the tree
      Restart();
```
See TTreeReader.cxx:548

ferdymercury: See also the documentation of Restart:

`/// Restart a Next() loop from entry 0 (of TEntryList index 0 of fEntryList is set).`

ferdymercury: Btw, plz squash-on-merge once approved.

pcanal: Fair enough.  Thanks.

wdconinc:
> @wdconinc could you test on your side with this patch?

Sorry for the delay. I haven't been able to test this yet (curiously I may have run into a gitlab pipeline bug, and then my internet was down all evening). Hope to confirm today.

Issue #6671, opened 2020-10-20
TH1F bin content saturation


stwunsch:
See for comparison the screenshot below.

The upper plot was done with `TTree.Draw`:

```python
>>> import ROOT
>>> f = ROOT.TFile('DYJetsToLL.root')
>>> t = f.Get('Events')
>>> t.Draw('GenPart_pdgId')
```

The lower plot was done with `RDataFrame.Histo1D`:

```python
>>> import ROOT
>>> c = ROOT.TCanvas()
>>> h = ROOT.RDataFrame('Events', 'DYJetsToLL.root').Histo1D('GenPart_pdgId')
>>> h.Draw()
```

![Screenshot from 2020-10-20 10-38-16](https://user-images.githubusercontent.com/6951222/96562487-1974b880-12c1-11eb-9bc5-1006c78031e5.png)

I've used ROOT 6.22/02 and you can download the file here:

http://opendata.web.cern.ch/record/12353

bellenot:
@stwunsch which one is correct?

stwunsch:
Good question! I guess the lower one, also checked by counting only the respective entries in the `GenPart_pdgId` vector.

stwunsch:
Well, it's a DY sample, so physically spoken we would expect the upper one (?) Let's say, it's not so clear what's going on :sweat_smile:

eguiraud:
Can you enforce the same exact binning in both and also try with raw TTree and raw TTreeReader? If RDF or TTree is producing wrong histograms it's a critical issue if not a release blocker.

stwunsch:
@eguiraud Do we agree that the code above should produce (in principle) the same output?

stwunsch:
@eguiraud Here you can find the plots produced with the same binning.

```python
>>> import ROOT
>>>  f = ROOT.TFile('DYJetsToLL.root')
>>> t = f.Get('Events')
>>> h = ROOT.TH1F('h', 'h', 100, -20, 20)
>>> c = ROOT.TCanvas()
>>> t.Draw('GenPart_pdgId>>h')
```

```python
>>> import ROOT
>>> c = ROOT.TCanvas()
>>> h = ROOT.RDataFrame('Events', 'DYJetsToLL.root').Histo1D(('h', 'h', 100, -20, 20), 'GenPart_pdgId')
>>> h.Draw()
```

[...]

PR #19819, opened 2025-09-04
[hist] Implement weighted filling

hahnjo:
It is only supported for floating-point bin content types.

### DIFF
--- a/hist/histv7/benchmark/hist_benchmark_engine.cxx
+++ b/hist/histv7/benchmark/hist_benchmark_engine.cxx
@@ -1,5 +1,6 @@
 #include <ROOT/RHistEngine.hxx>
 #include <ROOT/RRegularAxis.hxx>
+#include <ROOT/RWeight.hxx>
 
 #include <benchmark/benchmark.h>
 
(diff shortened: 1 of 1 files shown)


ferdymercury:
Thanks!

In the past, some users complained about the fact that Weighted Filling with floating point content does not warn about overflows.

Maybe for the new interfaces it would be worth to define an additional function SafeFill(x, const double tol = xx) function that does this with a penalty overhead.

See related discussion: https://github.com/root-project/root/issues/6671

SafeFill for integer content types is easier since there is no tolerance there, just the limit.

hist/histv7/inc/ROOT/RWeight.hxx
```diff
@@ -0,0 +1,28 @@
+/// \file
+/// \warning This is part of the %ROOT 7 prototype! It will change without notice. It might trigger earthquakes.
+/// Feedback is welcome!
+
+#ifndef ROOT_RWeight
+#define ROOT_RWeight
+
+#include <cstddef>
```
jblomer: May not be needed

hist/histv7/inc/ROOT/RWeight.hxx
```diff
@@ -0,0 +1,28 @@
...
+namespace ROOT {
+namespace Experimental {
+
+/**
+A weight for filling histograms.
+
+\warning This is part of the %ROOT 7 prototype! It will change without notice. It might trigger earthquakes.
+Feedback is welcome!
+*/
+struct RWeight final {
```
jblomer, line 17: General remark (not for this PR): at some point we should discuss which class names the v7 histograms will make available in the ROOT namespace and if we want to prefix them somehow. For instance, `ROOT::RWeight` is rather generic and may be applicable to other modules besides histograms.

hahnjo: Yes, we'll need to have this discussion somewhat soon, we need to decide on compromises between clarity and user convenience with short but generic names...

hahnjo:
> In the past, some users complained about the fact that Weighted Filling with floating point content does not warn about overflows.
> 
> Maybe for the new interfaces it would be worth to define an additional function SafeFill(x, const double tol = xx) function that does this with a penalty overhead.
> 
> See related discussion: #6671
> 
> SafeFill for integer content types is easier since there is no tolerance there, just the limit.

That's an interesting proposal, I took note. It may be difficult to find generic incarnations that work well with weighted filling. It's also the question whether an additional interface is actually helpful: If you use it, chances are you already thought about the possibility of overflows. Otherwise the default `Fill` still won't protect you...

hist/histv7/test/hist_engine.cxx
```diff
@@ -158,3 +158,78 @@ TEST(RHistEngine, FillInvalidNumberOfArguments)
...
+{
+   static constexpr std::size_t Bins = 20;
+   const RRegularAxis axis(Bins, 0, Bins);
+   RHistEngine<float> engine({axis});
+
+   engine.Fill(-100, RWeight(0.25));
+   for (std::size_t i = 0; i < Bins; i++) {
+      engine.Fill(i, RWeight(0.1 + i * 0.03));
+   }
+   engine.Fill(100, RWeight(0.75));
```
hageboeck, line 172: Would it make sense to use at least two dimensions in the test?
Errors stemming from incorrectly computing the global index won't show in this test (but might well be covered by other tests, hence my question).

hist/histv7/test/hist_engine.cxx
```diff
@@ -158,3 +158,78 @@ TEST(RHistEngine, FillInvalidNumberOfArguments)
...
+   EXPECT_EQ(engine.GetBinContent(RBinIndex::Overflow()), 0.75);
+}
+
+TEST(RHistEngine, FillTupleWeight)
+{
+   static constexpr std::size_t Bins = 20;
+   const RRegularAxis axis(Bins, 0, Bins);
+   RHistEngine<float> engine({axis});
+
+   engine.Fill(std::make_tuple(-100), RWeight(0.25));
```
hageboeck, line 187: Again, wondering if > 1D would be more comprehensive.

hist/histv7/test/hist_engine.cxx
hageboeck: OK, found it. The throw tests below are testing for the correct number of arguments.

PR #19825, opened 2025-09-04
Rename CINT to CLING and MAKECINT to ROOTCLING

ferdymercury:
# This Pull request:

## Changes or fixes:

Only functional changes here, pulled out according to @pcanal comment https://github.com/root-project/root/pull/19807#discussion_r2322541134


core/base/inc/RtypesImp.h
```diff
@@ -44,7 +44,7 @@ inline void operator delete(void*, ROOT::Internal::TOperatorNewHelper*) { }
 
 // The STL GenerateInitInstance are not unique and hence are declared static
 // (not accessible outside the dictionary and not linker error for duplicate)
-#if defined(__CINT__)
+#if defined(__CLING__)
```
pcanal: This one is interesting.  The #ifdef is literally `if we are in CINT workaround problem in CINT`, so the 'right change is actually to drop completely the `#ifdef __CINT__` part and keep only the `#else` part

interpreter/cling/include/cling/cint/multimap
```diff
@@ -1,4 +1,4 @@
-#ifdef __CINT__
```
pcanal, line 1: We likely (hopefully?) no longer need those files.

gui/gui/inc/TGTextViewStream.h
```diff
@@ -17,7 +17,7 @@
 #include <streambuf>
 #include <iostream>
 
-#if defined (R__WIN32) && defined (__MAKECINT__)
+#if defined (R__WIN32) && defined (__ROOTCLING__)
 typedef basic_streambuf<char, char_traits<char> > streambuf;
 #endif
```
pcanal: Let's try:
```suggestion

```

roottest/cling/parsing/semicolon.C
```diff
@@ -3,7 +3,7 @@ void semicolon(const char *string = "/star/simu/simu/jwebb/11-12-2010-pp500-pile
   cout << string << endl;
 }
 
-#if !defined(__CINT__) && !defined(__CLING__)
+#if !defined(__CLING__) && !defined(__CLING__)
```
pcanal: odd looking :)

roottest/cling/templateMembers/constructor.C
```diff
@@ -19,7 +19,7 @@ struct A {
   template <class U>
    A(const U&) { 
       cout << "c'tor taking " << 
-#if defined(__CINT__) || defined(_MSC_VER)
+#if defined(__CLING__) || defined(_MSC_VER)
```
pcanal: This category of changes, I would keep the __CINT__ for now or actually review whether it is needed at all:
```suggestion
#if defined(_MSC_VER)
```

ferdymercury:
I am not sure on how to deal with the remaining failing tests

pcanal:
> I am not sure on how to deal with the remaining failing tests

We will have to understand them on a case by case basis ... a priori any changes to 'remove' CINT renmant should not affect any behaviors.

ferdymercury:
> We will have to understand them on a case by case basis ... a priori any changes to 'remove' CINT renmant should not affect any behaviors.

```
roottest/cling/dict/runoperators.C
roottest/root/tree/stl/runmixing.C
```
fail both due to the same reason:
```
 fatal error: 'header.h+' file not found
  #include "header.h+"
           ^~~~~~~~~~~~~~~~~~~
```

They are being included with a "+" at the end. Which I do not know if this is supported or documented somewhere in doxygen???

Then we have:
```
runoperatorEqual.C...
  Class StThreeVector not found
```

Later we have:
`roottest-cling-function-rundefaultParams_WILL_FAIL`
which now passes instead of failing...

and
`test/stressMathCore.cxx`
who does not like ROOT::Math not being in the namespace
and also:
```
Warning: /github/home/ROOT-CI/src/test/stressMathCore.cxx:386:7: warning: parentheses were disambiguated as a function declaration [-Wvexing-parse]
        CREATE_DIST(beta);
        ^~~~~~~~~~~~~~~~~
```

roottest/root/tree/stl/runmixing.C
```diff
@@ -3,16 +3,16 @@
 
 namespace std {}; using namespace std;
 
-#if defined(__CINT__) && !defined(__MAKECINT__)
+#if defined(__CLING__) && !defined(__ROOTCLING__)
```
pcanal:
```
 #include "MyClass.h+"
```
must only be seen while running the interpreter (i.e. neither during compilation nor during the execution of rootcling).  This syntax tells Cling to first compile the header with ACLiC.  The error:
```
 fatal error: 'header.h+' file not found
  #include "header.h+"
           ^~~~~~~~~~~~~~~~~~~
```
indicates it is seen in a different context.  We may or may not need to 'fix' something in the definition.   I would separate these (type of) changes for a distinct PR.  (i.e. this may need more than just a respelling).

pcanal:
@ferdymercury We may also want to split this PR in much smaller chunks in general. With 325 files changed, it is challenging to properly review them.  We can either split by directories or maybe by 'theme' (for example a PR for only all the linkdef/#pragma changes).

ferdymercury:
> @ferdymercury We may also want to split this PR in much smaller chunks in general. With 325 files changed, it is challenging to properly review them. We can either split by directories or maybe by 'theme' (for example a PR for only all the linkdef/#pragma changes).

Hmm that should be doable, however, in principle this can be already done in the "Files changed" tab, by using "Hold shift + click to select a range". I had already split the commits by folders with that in mind.

Btw, we have now only two tests failing:
```
roottest-cling-function-rundefaultParams_WILL_FAIL
roottest-cling-templateMembers-runoperatorEqual
```

pcanal:
> however, in principle this can be already done in the "Files changed" tab,

On the other hand, I have a hard time remembering how far I reviewed if/when I have to split the review in several time sessions.

> Btw, we have now only two tests failing:

Good :)

core/dictgen/src/rootcling_impl.cxx
```diff
@@ -4486,8 +4486,8 @@ int RootClingMain(int argc,
    TClassEdit::Init(&helper);
 
    // flags used only for the pragma parser:
-   clingArgs.push_back("-D__CINT__");
-   clingArgs.push_back("-D__MAKECINT__");
+   clingArgs.push_back("-D__CINT__"); // backward compatibility. Now __CLING__ is used
+   clingArgs.push_back("-D__MAKECINT__"); // backward compatibility. Now __ROOTCLING__ is used
```
pcanal:
```suggestion
   clingArgs.push_back("-D__CINT__"); // backward compatibility. Now __CLING__ should be used instead
   clingArgs.push_back("-D__MAKECINT__"); // backward compatibility. Now __ROOTCLING__ should used instead
```

ferdymercury:
Now we only have some OS-dependent fails of one test:
`rundefaultParams_WILL_FAIL`

Could it be related to the CXX standard?

What would you suggest for solving it?

pcanal:
> Now we only have some OS-dependent fails of one test:
rundefaultParams_WILL_FAIL

What is the error?

ferdymercury:
> What is the error?

No error, the script rundefaultParams just runs correctly. I guess it was expected to always throw an error but now it doesn't, on some platforms.

pcanal:
> No error, the script rundefaultParams just runs correctly. I guess it was expected to always throw an error but now it doesn't, on some platforms.

I guess we'll have to debug locally the before and after (and investigate the intent of the test) to figure out what changes and whether it is for the better or worse.

ferdymercury:
> I guess we'll have to debug locally the before and after (and investigate the intent of the test) to figure out what changes and whether it is for the better or worse.

I guess this can help:
https://github.com/root-project/root/commit/989bd23a250d6a25d9f17d96489262dbb6f0e872
https://github.com/root-project/root/commit/70258f57e580c0aed9c0f44977206150b52cfffd

ferdymercury:
CI is green now. Failures are unrelated.

PR #19828, opened 2025-09-05
[df] Exclude static type maps from coverage

hahnjo:


hageboeck:
As discussed offline, I would say what codecov outputs currently is not sufficient to embark on an endeavour to annotate the code with comments.

pcanal:
> I would say what codecov outputs currently is not sufficient 

What do you mean and/or what additional output do you have in mind?

hageboeck:
> > I would say what codecov outputs currently is not sufficient
> 
> What do you mean and/or what additional output do you have in mind?

Let's take the file touched in this PR as an example:
https://app.codecov.io/gh/root-project/root/blob/master/tree%2Fdataframe%2Fsrc%2FRDFUtils.cxx#L157

I'm sure this function is called during RDF execution, so we should expect it to be covered.
- When the static map is initialised (i.e. on the first call of this function), I would expect line 157 to be covered. It's not. For a reason we don't understand, it has partial coverage. Does this mean that sometimes the map is initialised and sometimes not? I don't believe so.
- Going on, line 159 is a find in a map. It has partial coverage. Why partial though? Obviously, since the lines below have coverage, we hit both the case that an item is found (and the return is executed) *and* the case that the type name is not found (and we have to continue in line 162). So both branches are taken. We don't understand why this should result in partial coverage.
- The same can be said for the throws. They seem to get hit, so why is the coverage partial?
- Going on to line 206, the topic of this PR. The function name has coverage, so I take this as "it has been called". In that case, *by construction*, the static map should have been initialised. The initialiser list of that map has no coverage, though. Looks like a bug to me.

And these patterns seem to continue through other files we spot checked. I would therefore argue that what codecov outputs is not correct. That reminded me of a big campaign Wouter undertook in RooFit to make coverity happy. We are not using it, any more, though, so all these comments that had been added throughout the codebase are now just dead text.
That lead me to conclude with Jonas that adding codecov comments to RDF is worth the effort only if:
- The output of the tool is shown to be correct (I'm unsure about that at this point)
- And if we expect to be using it for an extended period of time. I'm not sure either if we will go more into codecov or away from it.

pcanal:
> I would expect line 157 to be covered. It's not. For a reason we don't understand, it has partial coverage.

It may indeed be a bug in gcov and/or codecov. But it is at least a lack of information.  Some post/documentation hints at the reason there being 'indirect' and something in the call (the implied constructor) not being fully covered.

> Going on to line 206, the topic of this PR. The function name has coverage, so I take this as "it has been called". In that case, by construction, the static map should have been initialised. The initialiser list of that map has no coverage, though. Looks like a bug to me.

It does to me too.  

> And if we expect to be using it for an extended period of time. I'm not sure either if we will go more into codecov or away from it.

I think that even with the bug mentioned, codecov is useful in particular if we worry more about the delta than the absolute value.  So I don't we should migrate away from it unless/until we migrate to a new/different code coverage tool.  I would consider annotating the lines that we do know that are not executed in test (eg. error condition that we don't know how or don't want to test) ...

hageboeck:
> > And if we expect to be using it for an extended period of time. I'm not sure either if we will go more into codecov or away from it.
> 
> I think that even with the bug mentioned, codecov is useful in particular if we worry more about the delta than the absolute value. So I don't we should migrate away from it unless/until we migrate to a new/different code coverage tool. I would consider annotating the lines that we do know that are not executed in test (eg. error condition that we don't know how or don't want to test) ...

I'm with you that's its indeed useful as a delta tool! I would also say that it should be maintained until something significantly better comes up. Using deltas, however, implies that adding annotations like in this PR to deal with bugs or not-well-understood details is not necessarily useful. These won't impact the deltas in work that is to follow.

hahnjo:
Closing this PR for the moment

PR #19558, opened 2025-08-06
[tmva][sofie] Profiler code generation for SOFIE

olia110:
# This Pull request:

This PR introduces a time-based profiler for the SOFIE inference engine. It provides developers with the tools to analyze the performance of generated C++ models by measuring the execution time of each individual operator, as well as the total inference time.

## Changes or fixes:

-   A new option, `SOFIE::Options::kProfile`, is added to the `RModel::Generate()` method to enable the feature.
-   When enabled, a new `RModelProfiler` helper class instruments the generated `doInfer()` method with `std::chrono` timers.
-   Utility functions like `PrintProfilingResults()` and `GetOpAvgTime()` are added to the generated `Session` struct to access the collected timing data.
## Checklist:

- [x] Tested changes locally by running inference in a loop and verifying the timing results.
- [ ] updated the docs (if necessary)

PR #19829, opened 2025-09-05
[tmva][sofie] Profiler code generation for SOFIE

lmoneta:
This PR introduces a time-based profiler for the SOFIE inference engine. It provides developers with the tools to analyze the performance of generated C++ models by measuring the execution time of each individual operator, as well as the total inference time.

### Changes or fixes:

A new option, SOFIE::Options::kProfile, is added to the RModel::Generate() method to enable the feature.
When enabled, a new RModelProfiler helper class instruments the generated doInfer() method with std::chrono timers.
Utility functions like PrintProfilingResults() and GetOpAvgTime() are added to the generated Session struct to access the collected timing data.
Checklist:

 Tested changes locally by running inference in a loop and verifying the timing results.
 updated the docs (if necessary)

This PR replaces #19558 by fixing the conflicts with rebasing to master

### DIFF
--- a/tmva/sofie/inc/TMVA/RModel.hxx
+++ b/tmva/sofie/inc/TMVA/RModel.hxx
@@ -45,6 +45,8 @@ private:
    MemoryPoolInfo fIntermediateMemoryInfo;    ///<!  intermediate memory info (transient)
    std::unordered_map<std::string_view, size_t> fIntermediateTensorFrequencyLookup;    ///<!  lookup table for intermediate tensor frequency (transient)
 
+   std::string fExtraCodeForDimShapes; // extra code needed for initialization of dynamic parameters (e.g. number of non zero elements in NonZero operator)
+
 public:
    /**
        Default constructor. Needed to allow serialization of ROOT objects. See
@@ -108,6 +110,7 @@ public:
 
    void AddShapeTensor(const std::string & name, const std::vector<Dim> & shapeValues, bool scalar = false);
 
+   void AddExtraCodeForDimShapes(const std::string & code) { fExtraCodeForDimShapes += code; }
 
    // add and initialize subgraph to the model
    void InitializeSubGraph(std::shared_ptr<RModel>  graph);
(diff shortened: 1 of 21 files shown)


dpiparo:
Is there perhaps a way in which we could test this new nice feature?

lmoneta:
Yes, I think we should add a tutorial for this, generting code instrumented with timers and then producing the results. 
@olia110 can you add this new tutorial?

olia110:
Yes, thank you. I will look how to add some testing and will add a tutorial.

tmva/sofie/inc/TMVA/RModel.hxx
```diff
@@ -11,16 +11,24 @@ namespace SOFIE {
...
+
    int fVerbose = 0;
    int fBatchSize = -1;
    long fReadPos = 0;  // reading file position
+
    size_t fConstantTensorSize = 0; // size  (in Bytes) of the allocated constant tensors
    size_t fWeightsTensorSize = 0;  // size  (in Bytes) of the allocated weight tensors
    size_t fOtherTensorSize = 0;    // size  (in Bytes) of intermediate tensors which are not managed by the memory pool
 
+   std::string fProfilerGC = "";
```
guitargeek: Adding a data member meas you have to increase the class version, otherwise you break backwards compatible IO.

But I'm happy that I have the opportunity to comment on this here! Because I think in general, supporting IO for RModel is not optimal. It bars us from changing the schema as we like, and I'm not sure what's the use case?

Furthermore, as SOFIE is still in the `Experimental` namespace, I think supporting persistent IO for these classes is in contradiction with the experimental namespace.

guitargeek (changes requested):
I didn't look at the full diff yet, but this definitely requires a change in the RModel class number. Or - which I would prefer - remove IO capability from RModel altogether. What's the use case for that IO? If ROOT with SOFIE is installed to read a possible RModel from a file, you can also just recreate the model from ONNX on the fly. I think sticking to ONNX as the persistent model format is cleaner.

PR #19830, opened 2025-09-05
[hist] Implement additional `RHistEngine` operations

hahnjo:
* `RHistEngine::Add`
 * `RHistEngine::Clear`
 * `RHistEngine::Clone`

### DIFF
--- a/hist/histv7/inc/ROOT/RHistEngine.hxx
+++ b/hist/histv7/inc/ROOT/RHistEngine.hxx
@@ -82,11 +82,26 @@ public:
    {
    }
 
-   // Copy constructor and assignment operator are deleted to avoid surprises.
+   /// The copy constructor is deleted.
+   ///
+   /// Copying all bin contents can be an expensive operation, depending on the number of bins. If required, users can
+   /// explicitly call Clone().
    RHistEngine(const RHistEngine<BinContentType> &) = delete;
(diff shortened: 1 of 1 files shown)


hist/histv7/inc/ROOT/RHistEngine.hxx
```diff
@@ -173,6 +173,18 @@ public:
       }
    }
 
+   /// Clone this histogram engine.
+   ///
+   /// \return the clone object
+   RHistEngine<BinContentType> Clone() const
```
hageboeck, line 199: Why explicitly Clone and not the copy constructor?

hahnjo: We had discussed offline that providing copy constructor and assignment operator in all cases is less confusing to users, even if it means they may copy too often. I still came up with one use case for explicit `Clone()`, when used from Python. I can re-implement it in terms of the copy constructor.

hahnjo: More offline discussion including @jblomer with the tendency to take a "purist" approach for now: Since we want `Clone()`, we will document the choice not to implement copy constructor and assignment. If this causes major problems, we can still implement them later on but the other way around of removing them is more difficult / impossible.

hageboeck (approved):

PR #19832, opened 2025-09-05
[ntuple] Default `RBulkValues` move constructor and assignment

hahnjo:
All members are moveable and there seems to be no custom code.

### DIFF
--- a/tree/ntuple/src/RFieldBase.cxx
+++ b/tree/ntuple/src/RFieldBase.cxx
@@ -122,35 +122,6 @@ void ROOT::RFieldBase::RValue::BindRawPtr(void *rawPtr)
 
 //------------------------------------------------------------------------------
 
-ROOT::RFieldBase::RBulkValues::RBulkValues(RBulkValues &&other)
-   : fField(other.fField),
-     fValueSize(other.fValueSize),
-     fCapacity(other.fCapacity),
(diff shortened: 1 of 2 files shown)


tree/ntuple/src/RFieldBase.cxx
```diff
@@ -122,35 +122,6 @@ void ROOT::RFieldBase::RValue::BindRawPtr(void *rawPtr)
...
-   : fField(other.fField),
-     fValueSize(other.fValueSize),
-     fCapacity(other.fCapacity),
-     fSize(other.fSize),
-     fIsAdopted(other.fIsAdopted),
-     fNValidValues(other.fNValidValues),
-     fFirstIndex(other.fFirstIndex)
-{
-   std::swap(fDeleter, other.fDeleter);
-   std::swap(fValues, other.fValues);
```
silverweed, line 135: I think this swap is necessary, otherwise the pointer may just get copied and we'd end up freeing the values twice in the destructor, no?

hahnjo: Hm, that's a good point - I found this code because our coverage report says we're not testing it at all, so maybe we should do this and see.

hahnjo:
Replaced by https://github.com/root-project/root/pull/19979

PR #19833, opened 2025-09-05
[6.36] [treeplayer] prevent nullptr access in Restart (#19817)

vepadulano:
Backport of #19817


hahnjo:
Note that the test fails on `mac15` after merging to `master`: https://github.com/root-project/root/actions/runs/17494695421/job/49692399625

PR #19840, opened 2025-09-07
Add missing ctest dependency for readPtrColls tests

LukasBreitwieser:


dpiparo:
Nice catch!

roottest/root/io/uniquePointer/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ ROOTTEST_COMPILE_MACRO(readPtrColls.C
 
 ROOTTEST_ADD_TEST(readPtrColls
                   MACRO readPtrColls.C+
+                  DEPENDS writePtrColls
```
hahnjo, line 24: This should not be needed, we use fixtures for this. When I looked at the logs, writePtrColls was nicely executed at the right place...

linev: Yes, in most of the places one should replace `DEPENDS` by fixtures.

LukasBreitwieser: Thanks for your comments. I also saw in the logs that the `writePtrColls` was executed and finished before the read tests, just wanted to make sure this is guaranteed. However, it seems that I misinterpreted the [FIXTURE_DEPENDS](https://cmake.org/cmake/help/latest/prop_test/FIXTURES_REQUIRED.html) documentation. Will close this PR.

linev:
Some side effect with Windows can be generation of rootmap file.
So with default settings in the directory appears several libraries implementing same class - and ROOT tries load all of them when running macros.
Windows shows warnings like:
```
Warning in <TInterpreter::ReadRootmapFile>: While processing ./readUniqPtrColls_C.rootmap, class MyObj was found to be associated to readUniqPtrColls_C although it is already associated to readPtrColls_C
  Warning in <TInterpreter::ReadRootmapFile>: While processing ./readUniqPtrColls_C.rootmap, class MyTest was found to be associated to readUniqPtrColls_C although it is already associated to readPtrColls_C
```

But then fails.

PR #19700 removes `.rootrc` from the directory - suppressing default behavior on Windows.
Now rootmap should not exists and failure should not appear again

PR #19842, opened 2025-09-08
[net] Add libcurl based support for files on HTTP remotes

jblomer:
Add new `TFile` and `RRawFile` plugins that use libcurl for the HTTP transfers. This can eventually replace the Davix dependency and also the older `TWebFile` and `TS3WebFile`. The curl library gives access to a modern HTTP implementation (e.g., h2), and it works on Linux, macOS, and Windows with the respective native SSL stacks.

The current implementation does HEAD and (multi-)range GET queries, which provides the basics for efficient use of HTTP in ROOT. 

Authentication can relatively easily be added, since S3 (AWSv4 signatures), bearer token authentication and X.509 client certificates are already supported by libcurl.

Meta-link support can in principle be added if still needed. It requires XML parsing of the metalink snippet.

Write support (POST, PUT) requests can in principle be added if still needed, but require a closer look.


bellenot:
@jblomer I'll check

net/curl/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+    src/RCurlConnection.cxx
+    src/RRawFileCurl.cxx
+    src/TCurlFile.cxx
+  LIBRARIES
+    CURL::libcurl
+  DEPENDENCIES
+    RIO
+)
+
+target_compile_options(RCurlHttp PRIVATE -Wno-deprecated-declarations)
```
bellenot:
```suggestion
if(NOT MSVC)
  target_compile_options(RCurlHttp PRIVATE -Wno-deprecated-declarations)
endif()
```

bellenot:
Note that after successful build on Windows, some of the tests fail like:
```
The following tests FAILED:
        157 - gtest-io-io-RRawFile (Failed)
        262 - gtest-net-curl-rawfile-curl (Failed)
        263 - gtest-net-curl-tfile-curl (Failed)
        488 - gtest-tree-dataframe-datasource-csv (Failed)
        493 - gtest-tree-dataframe-datasource-sqlite (Failed)
Errors while running CTest
```

bellenot:
Thanks @jblomer . FYI, I'll have to install `libcurl` and modify the environment on the build nodes (one after the other, by hand). Will try to do that tonight or over the weekend, and then I'll restart the builds on Windows.

io/io/src/RRawFile.cxx
```diff
@@ -75,8 +75,7 @@ ROOT::Internal::RRawFile::Create(std::string_view url, ROptions options)
        transport == "root" || transport == "roots" ) {
       std::string plgclass = transport.compare( 0, 4, "http" ) == 0 ?
                              "RRawFileDavix" : "RRawFileNetXNG";
-      if (TPluginHandler *h = gROOT->GetPluginManager()->
-          FindHandler("ROOT::Internal::RRawFile", std::string(url).c_str())) {
+      if (TPluginHandler *h = gROOT->GetPluginManager()->FindHandler("RRawFile", std::string(url).c_str())) {
```
bellenot:
```suggestion
      if (TPluginHandler *h = gROOT->GetPluginManager()->
          FindHandler("::ROOT::Internal::RRawFile", std::string(url).c_str())) {
```

bellenot (changes requested):

io/io/src/RRawFile.cxx
jblomer: Thank you! Fixed, although it still fails on my devel machine (arch)...

net/curl/inc/LinkDef.h
```diff
@@ -0,0 +1,5 @@
+#ifdef __CLING__
+
+#pragma link C++ class TCurlFile+;
```
bellenot, line 4:
```suggestion
#pragma link C++ class TCurlFile+;
#pragma link C++ class ROOT::Internal::RRawFileCurl+;
```
That makes it working on Windows

bellenot (changes requested):

PR #19844, opened 2025-09-08
[ntuple] fix automatic schema evolution of variant

### LINKED ISSUE #19253: [ntuple] Schema evolution of `std::variant`
### Description

We currently have no `RVariantField::BeforeConnectPageSource` that either checks or correctly schema evolves a `std::variant` field.

### Reproducer

```c++

#include <cstdint>
#include <string>
#include <variant>

void ntuple_variant() {
  {
    using VarType = std::variant<std::int32_t, std::string>;
    auto var = model->MakeField<VarType>("var");
    auto writer = ROOT::RNTupleWriter::Recreate(std::move(model), "ntpl", "ntuple_variant.root");

    *var = 42;
    writer->Fill();
    *var = "abc";
    writer->Fill();
  }

  { // 1
    using VarType = std::variant<std::int32_t, std::string, float>;
    auto var = model->MakeField<VarType>("var");
    try {
      auto reader = ROOT::RNTupleReader::Open(std::move(model), "ntpl", "ntuple_variant.root");
    } catch (const ROOT::RException &e) {
      std::cout << e.GetError().GetReport() << "\n";
    }
  }

  { // 2
    using VarType = std::variant<std::int32_t>;
    auto var = model->MakeField<VarType>("var");
    auto reader = ROOT::RNTupleReader::Open(std::move(model), "ntpl", "ntuple_variant.root");
    reader->LoadEntry(0);
    std::cout << "0: index = " << var->index() << "\n";
    reader->LoadEntry(1);
    std::cout << "1: index = " << var->index() << "\n";
  }
}
```

This macro writes a `std::variant<std::int32_t, std::string>` and then demonstrates two problems:
1. When trying to read back as a `std::variant<std::int32_t, std::string, float>`, the additional `float` leaf field complains. I think it would be good to automatically support this case.
```
No on-disk field information for `var._2`
At:
  const ROOT::RFieldBase::ColumnRepresentation_t &ROOT::RFieldBase::EnsureCompatibleColumnTypes(const ROOT::RNTupleDescriptor &, std::uint16_t) const
```
2. The second case is more severe: Here the user tries to read back into a `std::variant<std::int32_t>` and this crashes once the on-disk switch tag indicates a value in the second item field:
https://github.com/root-project/root/blob/70c900f6b9dfbaa8e501d540b7ec43ddc0133109/tree/ntuple/src/RFieldMeta.cxx#L1258-L1262

### Additional context

We discussed a conservative approach to check exact identity of the type name.

jblomer:
Fixes #19253

### DIFF
--- a/tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
+++ b/tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
@@ -413,6 +413,8 @@ protected:
 
    void CommitClusterImpl() final;
 
+   void ReconcileOnDiskField(const RNTupleDescriptor &desc) final;
+
 public:
    RVariantField(std::string_view fieldName, std::vector<std::unique_ptr<RFieldBase>> itemFields);
    RVariantField(RVariantField &&other) = default;
--- a/tree/ntuple/src/RFieldMeta.cxx
+++ b/tree/ntuple/src/RFieldMeta.cxx
@@ -1297,6 +1297,19 @@ void ROOT::RVariantField::GenerateColumns(const ROOT::RNTupleDescriptor &desc)
    GenerateColumnsImpl<ROOT::Internal::RColumnSwitch>(desc);
 }
 
+void ROOT::RVariantField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
+{
+   static const std::vector<std::string> prefixes = {"std::variant<"};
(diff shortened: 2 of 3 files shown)


pcanal:
Could we enhance the git log to described what 'fixed' means in this context?

hahnjo (approved):
LGTM

PR #19846, opened 2025-09-09
[ntuple] Fix automatic schema evolution of record fields

### LINKED ISSUE #19254: [ntuple] Schema evolution of record fields
### Check duplicate issues.

- [x] Checked for duplicates

### Description

We currently have no `RRecordField::BeforeConnectPageSource` that either checks or correctly schema evolves record fields.

### Reproducer

```c++

#include <cstdint>
#include <string>
#include <tuple>

void ntuple_tuple() {
  {
    using TupType = std::tuple<std::int32_t, std::string>;
    auto tup = model->MakeField<TupType>("tup");
    auto writer = ROOT::RNTupleWriter::Recreate(std::move(model), "ntpl", "ntuple_tuple.root");

    *tup = {42, "abc"};
    writer->Fill();
  }

  { // 1
    using TupType = std::tuple<std::int32_t, std::string, float>;
    auto tup = model->MakeField<TupType>("tup");
    try {
      auto reader = ROOT::RNTupleReader::Open(std::move(model), "ntpl", "ntuple_tuple.root");
    } catch (const ROOT::RException &e) {
      std::cout << e.GetError().GetReport() << "\n";
    }
  }

  { // 2
    using TupType = std::tuple<std::string>;
    auto tup = model->MakeField<TupType>("tup");
    try {
    auto reader = ROOT::RNTupleReader::Open(std::move(model), "ntpl", "ntuple_tuple.root");
      reader->LoadEntry(0);
      std::cout << "0: string = " << std::get<0>(*tup) << "\n";
    } catch (const ROOT::RException &e) {
      std::cout << e.GetError().GetReport() << "\n";
    }
  }
}
```

This macro writes a `std::tuple<std::int32_t, std::string>` and then demonstrates two cases:
1. When trying to read back as a `std::tuple<std::int32_t, std::string, float>`, the additional `float` leaf field complains:
```
No on-disk field information for `tup._2`
At:
  const ROOT::RFieldBase::ColumnRepresentation_t &ROOT::RFieldBase::EnsureCompatibleColumnTypes(const ROOT::RNTupleDescriptor &, std::uint16_t) const
```
2. With an imposed field type of `std::variant<std::string>`, the user could expect that the string is read back. Instead it (correctly) throws:
```
On-disk column types {`SplitInt32`} for field `tup._0` cannot be matched to its in-memory type `std::string` (representation index: 0)                                                                 
At:
  const ROOT::RFieldBase::ColumnRepresentation_t &ROOT::RFieldBase::EnsureCompatibleColumnTypes(const ROOT::RNTupleDescriptor &, std::uint16_t) const
```

### ROOT version

`master`

### Installation method

from source

### Operating system

Linux

### Additional context

We discussed a conservative approach to check exact identity of the type name.

jblomer:
Includes automatic schema evolution for `std::pair` and `std::tuple`, which are implemented as descendants of the `RRecordField`.

Fixes #19254

### DIFF
--- a/tree/ntuple/src/RField.cxx
+++ b/tree/ntuple/src/RField.cxx
@@ -607,6 +607,13 @@ void ROOT::RRecordField::ReadInClusterImpl(RNTupleLocalIndex localIndex, void *t
 
 void ROOT::RRecordField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
 {
+   if (fTraits & kTraitEmulatedField) {
+      // The field has been explicitly constructed following the on-disk information. No further reconcilation needed.
+      return;
+   }
+   // Note that the RPairField and RTupleField descendants have their own reconcilation logic
+   R__ASSERT(GetTypeName().empty());
+
    const auto &fieldDesc = desc.GetFieldDescriptor(GetOnDiskId());
    EnsureMatchingOnDiskField(fieldDesc, kDiffTypeName | kDiffTypeVersion);
 
--- a/tree/ntuple/inc/ROOT/RField/RFieldRecord.hxx
+++ b/tree/ntuple/inc/ROOT/RField/RFieldRecord.hxx
@@ -63,6 +63,7 @@ class RRecordField : public RFieldBase {
 
    /// If `emulatedFromType` is non-empty, this field was created as a replacement for a ClassField that we lack a
    /// dictionary for and reconstructed from the on-disk information.
+   /// Used by the public constructor and by Internal::CreateEmulatedRecordField().
    RRecordField(std::string_view fieldName, std::vector<std::unique_ptr<RFieldBase>> itemFields,
                 std::string_view emulatedFromType);
 
@@ -82,6 +83,8 @@ protected:
    void ReadGlobalImpl(ROOT::NTupleSize_t globalIndex, void *to) final;
    void ReadInClusterImpl(RNTupleLocalIndex localIndex, void *to) final;
 
+   /// Used by RPairField and RTupleField descendants. These descendants have their own logic to attach the subfields
+   /// that ensure that the resulting memory layout matches std::pair or std::tuple, resp.
    RRecordField(std::string_view fieldName, std::string_view typeName);
 
    void AttachItemFields(std::vector<std::unique_ptr<RFieldBase>> itemFields);
@@ -106,6 +109,8 @@ protected:
 public:
    /// Construct a RRecordField based on a vector of child fields. The ownership of the child fields is transferred
    /// to the RRecordField instance.
+   /// The resulting field uses a memory layout for its values as if there was a struct consisting of the passed
+   /// item fields.
    RRecordField(std::string_view fieldName, std::vector<std::unique_ptr<RFieldBase>> itemFields);
    RRecordField(RRecordField &&other) = default;
    RRecordField &operator=(RRecordField &&other) = default;


tree/ntuple/inc/ROOT/RFieldBase.hxx
```diff
@@ -421,6 +420,11 @@ protected:
    static void CallConstructValueOn(const RFieldBase &other, void *where) { other.ConstructValue(where); }
    static std::unique_ptr<RDeleter> GetDeleterOf(const RFieldBase &other) { return other.GetDeleter(); }
 
+   /// Allow parents to mark their childs as artificial fields (used in class and record fields)
+   void CallSetArtificialOn(RFieldBase &other) { other.SetArtificial(); }
+   /// Allow class fields to adjust the type alias of their members
+   void SetTypeAliasOf(RFieldBase &other, const std::string &alias) { other.fTypeAlias = alias; }
```
hahnjo: I believe these functions can be `static`?

```suggestion
   static void CallSetArtificialOn(RFieldBase &other) { other.SetArtificial(); }
   /// Allow class fields to adjust the type alias of their members
   static void SetTypeAliasOf(RFieldBase &other, const std::string &alias) { other.fTypeAlias = alias; }
```

tree/ntuple/src/RField.cxx
```diff
@@ -605,7 +606,20 @@ void ROOT::RRecordField::ReadInClusterImpl(RNTupleLocalIndex localIndex, void *t
 
 void ROOT::RRecordField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
```
hahnjo, line 608: If I understand correctly, this function is now only for untyped record fields because `RPairField` and `RTupleField` provide their own override. Maybe we can add a comment and
```c++
R__ASSERT(fName.empty());
```
(or whatever other method we have to identify untyped record fields) to ensure this invariant is not violated in the future?

tree/ntuple/src/RField.cxx
```diff
@@ -605,7 +606,20 @@ void ROOT::RRecordField::ReadInClusterImpl(RNTupleLocalIndex localIndex, void *t
...
 {
-   EnsureMatchingOnDiskField(desc.GetFieldDescriptor(GetOnDiskId()), kDiffTypeName | kDiffTypeVersion);
+   const auto &fieldDesc = desc.GetFieldDescriptor(GetOnDiskId());
+   EnsureMatchingOnDiskField(fieldDesc, kDiffTypeName | kDiffTypeVersion);
+
+   // The on-disk ID of subfields is matched by field name. So we inherently support reordering of fields
+   // and we will ignore extra on-disk fields.
+   // It remains to mark the extra in-memory fields as artificial.
+   std::unordered_set<std::string> onDiskSubfields;
```
silverweed: These can be `string_view`s to avoid an unneeded copy

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -1151,6 +1166,22 @@ ROOT::RTupleField::RTupleField(std::string_view fieldName, std::vector<std::uniq
    }
 }
 
+void ROOT::RTupleField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
+{
+   static const std::vector<std::string> prefixes = {"std::pair<", "std::tuple<"};
```
silverweed, line 1189: We should probably change `EnsureMatchingTypePrefix` to accept a `span` so we can avoid creating `vector`s for things that should really be static arrays

jblomer: Ok, I'd like to do that in a separate PR.

tree/ntuple/src/RField.cxx
jblomer: Good point! It actually revealed another bug.

hahnjo (approved):
LGTM, thanks!

PR #19852, opened 2025-09-09
Modernize Python hsimple tutorial

vepadulano:
See the first commit for the actual changes.

### DIFF
--- a/tutorials/hsimple.py
+++ b/tutorials/hsimple.py
@@ -15,69 +15,67 @@
 ## \author Wim Lavrijsen, Enric Tejedor
 ## \author Vincenzo Eduardo Padulano (CERN), 09.2025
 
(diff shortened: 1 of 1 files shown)


vepadulano:
@guitargeek I thought about that, but then I read the docs and discovered that `SetSeed(0)` actually **does not** set the seed to zero, instead initialises the seed to a random value every time

https://root.cern.ch/doc/master/classTRandom.html#a2007ae15ce828266aecb783ed3410e8b

I guess it means that the other tutorials using numpy RNG are actually doing something different than what was done by gRandom in the past

guitargeek:
Oh, interesting I didn't know! Not a very type-safe interface that NumPy has there, and `SetSeed(None)` would have been more pythonic. Sorry I got fooled by my intuition. Ok it's good that you also thought about that, and whatever you decide is fine for this tutorial :+1:

PR #19854, opened 2025-09-10
[DO NOT MERGE] Debug MacOS build

vepadulano:
This is to debug the failures on my MacOS machine introduced by https://github.com/root-project/root/pull/19718

As commented https://github.com/root-project/root/pull/19718#issuecomment-3241681624, there seems to be some configuration where also on the CI nodes the build fails, but it was not understood why.

### DIFF
--- a/interpreter/CMakeLists.txt
+++ b/interpreter/CMakeLists.txt
@@ -1,6 +1,10 @@
 # Do not depend on CMake scripts from the rest of the ROOT build
 set(CMAKE_MODULE_PATH "")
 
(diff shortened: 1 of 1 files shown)


vepadulano:
Mystery solved: our Mac CI runs have the `SDKROOT` variable set https://github.com/root-project/root/actions/runs/17606774966/job/50019156945?pr=19854#step:6:500

I have no idea what is setting that variable. Logging to the same machine via `ssh` shows the variable is not set by default in the system. So something must be happening only during the setup of our CI.

PR #19855, opened 2025-09-10
Revert "[CMake] Remove overriding `CMAKE_OSX_SYSROOT`"

vepadulano:
This reverts commit e3a10abc23eaf82aeafff9b14f879ca72e6c03f9.

The MacOS builds on the CI were working by chance as shown at https://github.com/root-project/root/pull/19854.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,19 @@ if(WIN32)
   cmake_policy(SET CMP0091 OLD)
   set(CMAKE_SKIP_TEST_ALL_DEPENDENCY TRUE)
 endif()
+if(CMAKE_HOST_APPLE AND (NOT CMAKE_OSX_SYSROOT OR CMAKE_OSX_SYSROOT STREQUAL ""))
+  # The SYSROOT *must* be set before the project() call
+  execute_process(COMMAND xcrun --sdk macosx --show-sdk-path
+    OUTPUT_VARIABLE SDK_PATH
+    ERROR_QUIET
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+  if(NOT EXISTS "${SDK_PATH}")
+    message(FATAL_ERROR "Could not detect macOS SDK path")
+  endif()
+
+  set(CMAKE_OSX_SYSROOT "${SDK_PATH}" CACHE PATH "SDK path" FORCE)
+endif()
 
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR AND NOT ALLOW_IN_SOURCE)
   message(FATAL_ERROR


devajithvs (approved):

devajithvs:
We still need to figure out why this is happening and what was happening before `CMake 4`. But let's get this in to not break the builds.

guitargeek:
Also, if the reason that is was not necessary in our CI to set `CMAKE_OSX_SYSROOT` was that `SDKROOT` was already set in our CI, can we make sure that our CI doesn't do that before merging this? Our CI configuration should not set any secret CMake variables that are necessary to compile. IMO, we should have a configuration similar to the standard user configuration, so that we are sensitive to possible problems with it.

vepadulano:
> Also, if the reason that is was not necessary in our CI to set CMAKE_OSX_SYSROOT was that SDKROOT was already set in our CI, can we make sure that our CI doesn't do that before merging this? Our CI configuration should not set any secret CMake variables that are necessary to compile. IMO, we should have a configuration similar to the standard user configuration, so that we are sensitive to possible problems with it.

As I said in my comment above, I have no idea who/what is messing with our CI runners so that the `SDKROOT` variable is set. I unfortunately don't have time to investigate that. I refuse to block this PR because of that, because that means forcing any dev compiling ROOT master on MacOS to also include this commit reverting the change and that is not sustainable

guitargeek:
Which comment above? Maybe you mean the private discussion on mattermost :smiley: Or did I miss something?

Yes sure, no need to block this PR for this, but can we maybe write at least some notes? So if this code is touched again in the future it's clear what the constraints are and that there is something in the CI that can be improved about not setting the `SDKROOT`. A `TODO` in the touched CMake code is the best IMHO because the most visible.  Commit message or PR description is also fine, but I think in-code comments are less likely to be missed.

guitargeek:
Ah I see now, the information is indirectly available via the link to https://github.com/root-project/root/pull/19854 ! Okay that's maybe alright, but I think directly having it in the code comments or commit messages will be better.

Sorry for being pedantic, but we need to minimize the risk that people like me break things in the future again because they are missing the context :slightly_smiling_face:

guitargeek (approved):
Thanks for considering my request about checking if `xcrun` is available :pray:

Good now for me, maybe only consider an inline note on why this CMake code block is seemingly not needed in our CI. And that it would be better to avoid setting implicitly required CMake variables in the CI.

vepadulano:
> Good now for me, maybe only consider an inline note on why this CMake code block is seemingly not needed in our CI. And that it would be better to avoid setting implicitly required CMake variables in the CI.

Yes, done in the latest commit 👍

PR #19569, opened 2025-08-07
[roottest] enable several `root/aclic` tests

linev:
1. Implement new version of `aclic/nolinkdep` tests. Seems to be they were testsed before with CINT and were not working with cling. Now at least basic tests are works. ~~Plus linking between `script2.C` and `single.C`~~. Also renaming of script2.C into script.C is testing
2. Enable again `aclic/offset` test, was completely skipped
3. Remove many `.rootrc` file which are now not necessary

PR #19857, opened 2025-09-10
[roottest] use cmake in root/aclic tests

linev:
Use fixtures
Replace DEPENDS by explicit compilation of macro

It replacing https://github.com/root-project/root/pull/19569


pcanal (approved):
Thanks.

Issue #19850, opened 2025-09-09
Error: "no member named 'getenv' in the global namespace" during the compilation with libc++


YanzhaoW:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Hi,

I'm trying to install ROOT via clang with libc++. The installation failed due to the following error:

```text
[2130/5199] Building CXX object core/foundation/CMakeFiles/Foundation_Stage1.dir/src/FoundationUtils.cxx.o
FAILED: core/foundation/CMakeFiles/Foundation_Stage1.dir/src/FoundationUtils.cxx.o
/usr/bin/clang++-20  -I/home/yanwang/software/src/root/core/foundation/inc -I/home/yanwang/software/src/root/core/foundation/res -I/home/yanwang/software/src/root/core/cont/inc -I/home/yanwang/software/src/root/core/clingutils/res -I/home/yanwang/software/src/root/core/clingutils/inc -I/home/yanwang/software/src/root/core/clib/inc -I/home/yanwang/software/src/root/core/base/inc -I/home/yanwang/software/src/root-build/ginclude -fcolor-diagnostics -Wc++11-narrowing -Wsign-compare -Wsometimes-uninitialized -Wconditional-uninitialized -Wheader-guard -Warray-bounds -Wcomment -Wtautological-compare -Wstrncat-size -Wloop-analysis -Wbool-conversion -pipe  -Wall -W -Woverloaded-virtual -fsigned-char -pthread -stdlib=libc++ -O3 -DNDEBUG -std=c++23 -fPIC -fvisibility-inlines-hidden    -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fno-strict-aliasing -Wwrite-strings -Wno-shadow -Wno-unused-parameter -MD -MT core/foundation/CMakeFiles/Foundation_Stage1.dir/src/FoundationUtils.cxx.o -MF core/foundation/CMakeFiles/Foundation_Stage1.dir/src/FoundationUtils.cxx.o.d -o core/foundation/CMakeFiles/Foundation_Stage1.dir/src/FoundationUtils.cxx.o -c /home/yanwang/software/src/root/core/foundation/src/FoundationUtils.cxx
/home/yanwang/software/src/root/core/foundation/src/FoundationUtils.cxx:167:36: error: no member named 'getenv' in the global namespace
[...]

PR #19864, opened 2025-09-10
[core] fix getenv namespace error with libc++

ferdymercury:
Fixes https://github.com/root-project/root/issues/19850

### DIFF
--- a/bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx
+++ b/bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx
@@ -39,7 +39,7 @@
 #include "TThread.h"
 
 // Standard
-#include <assert.h>
+#include <cassert>
 #include <algorithm>     // for std::count, std::remove
 #include <climits>
 #include <stdexcept>
@@ -47,9 +47,9 @@
 #include <new>
 #include <set>
 #include <sstream>
-#include <signal.h>
-#include <stdlib.h>      // for getenv
-#include <string.h>
+#include <csignal>
+#include <cstdlib>      // for getenv
+#include <cstring>
 #include <typeinfo>
 
 // temp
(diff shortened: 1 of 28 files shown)


guitargeek (approved):
Thanks a lot! Will you take care of the backport to 6.36, or should I do it?

ferdymercury:
> Thanks a lot! Will you take care of the backport to 6.36, or should I do it?

I could do it. The question to me before is:

- There are more ::getenv throughout the codebase without std:: prepended. Not sure if they should also be changed, for consistency or if that will break things on Windows or who knows what.
- The user also reports a failure with cxxmodules after fixing this, (see same issue)

guitargeek:
Good points!

  * we should use `std::getenv` everywhere. The idea of moden C++ is to not use the C headers like `stdlib.h`, but instead use the standard library headers like `cstdlib`. See also this clang tidy check:
    https://clang.llvm.org/extra/clang-tidy/checks/modernize/deprecated-headers.html
    So it's probably a good opportunity to use `std::getenv` all the way :+1:. Can we do it in this PR?
    This should also work on Windows, as the functions do the same thing.
 * I have seen that, that's a shame... Let's continue the discussion in that PR and ping our macOS expert @couet  when we're stuck


See also:
  * https://en.cppreference.com/w/cpp/header/cstdlib.html
  * https://en.cppreference.com/w/c/header/stdlib.html

ferdymercury:
> Can we do it in this PR?

Done as suggested

interpreter/CppInterOp/lib/CppInterOp/Compatibility.h
```diff
@@ -23,7 +23,7 @@ static inline char* GetEnv(const char* Var_Name) {
   getenv_s(&sz, Env, sz, Var_Name);
   return Env;
 #else
-  return getenv(Var_Name);
+  return std::getenv(Var_Name);
```
pcanal: We probably should address this change upstream rather than here.
```suggestion
  return getenv(Var_Name);
```

ferdymercury: Is now https://github.com/compiler-research/CppInterOp/pull/709

PR #19865, opened 2025-09-10
[cmake][win] Fix deadly dependency on rootcling

bellenot:
Fix a rootcling dependency issue on Windows, leading to this kind of error:
```
          Start 1866: roottest-root-io-datamodelevolution-misc-MyClass-build
        Test #1791: roottest-root-io-autoptr-test_auto_ptr_v2_dict-build ..............................................***Failed   55.59 sec
MSBuild version 17.9.5+33de0b227 for .NET Framework

  Updating etc/gitinfo.txt.
  ZLIB.vcxproj -> C:\ROOT-CI\build\lib\ZLIB.lib
  lz4.vcxproj -> C:\ROOT-CI\build\lib\lz4.lib
  xxhash.vcxproj -> C:\ROOT-CI\build\lib\xxhash.lib
  zstd.vcxproj -> C:\ROOT-CI\build\lib\zstd.lib
  obj.clangSupport.vcxproj -> C:\ROOT-CI\build\interpreter\llvm-project\llvm\tools\clang\lib\Support\obj.clangSupport.dir\Release\obj.clangSupport.lib
  LLVMDemangle.vcxproj -> C:\ROOT-CI\build\interpreter\llvm-project\llvm\Release\lib\LLVMDemangle.lib

  clingMetaProcessor.vcxproj -> C:\ROOT-CI\build\lib\clingMetaProcessor.lib
C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.CppBuild.targets(2563,5): warning : Remote deployment might be slow/inefficient. The process cannot access the file 'C:\ROOT-CI\build\interpreter\llvm-project\llvm\tools\clang\include\clang\Basic\Win32\Release\ClangRISCVSiFiveVectorBuiltinSema\ClangRISCVSiFiveVectorBuiltinSema.recipe' because it is being used by another process. [C:\ROOT-CI\build\interpreter\cling\lib\MetaProcessor\clingMetaProcessor.vcxproj]
  ClingUtils.vcxproj -> C:\ROOT-CI\build\core\clingutils\ClingUtils.dir\Release\ClingUtils.lib
  Dictgen.vcxproj -> C:\ROOT-CI\build\core\dictgen\Dictgen.dir\Release\Dictgen.lib
  clangCppInterOp.vcxproj -> C:\ROOT-CI\build\lib\clangCppInterOp.lib
  MetaCling.vcxproj -> C:\ROOT-CI\build\core\metacling\src\MetaCling.dir\Release\MetaCling.lib
  Clib.vcxproj -> C:\ROOT-CI\build\core\clib\Clib.dir\Release\Clib.lib
  Foundation_Stage1.vcxproj -> C:\ROOT-CI\build\core\foundation\Foundation_Stage1.dir\Release\Foundation_Stage1.lib
LINK : fatal error LNK1104: cannot open file 'C:\ROOT-CI\build\bin\rootcling_stage1.exe' [C:\ROOT-CI\build\core\rootcling_stage1\rootcling_stage1.vcxproj]
```
when running (building) many tests, and making the builds much slower as well


cmake/modules/RootMacros.cmake
```diff
@@ -619,9 +619,13 @@ function(ROOT_GENERATE_DICTIONARY dictionary)
...
-      set(command ${CMAKE_COMMAND} -E env "ROOTIGNOREPREFIX=1" $<TARGET_FILE:rootcling> -rootbuild)
-      # Modules need RConfigure.h copied into include/.
-      set(ROOTCINTDEP rootcling rconfigure)
+      if(MSVC)
+        set(command ${CMAKE_COMMAND} -E env "ROOTIGNOREPREFIX=1" ${CMAKE_BINARY_DIR}/bin/rootcling.exe -rootbuild)
+      else()
+        set(command ${CMAKE_COMMAND} -E env "ROOTIGNOREPREFIX=1" $<TARGET_FILE:rootcling> -rootbuild)
+        # Modules need RConfigure.h copied into include/.
+        set(ROOTCINTDEP rootcling rconfigure)
+      endif()
```
```suggestion
        set(command ${CMAKE_COMMAND} -E env "ROOTIGNOREPREFIX=1" ${CMAKE_BINARY_DIR}/bin/rootcling.exe -rootbuild)
      set(command ${CMAKE_COMMAND} -E env "ROOTIGNOREPREFIX=1" $<TARGET_FILE:rootcling> -rootbuild)
      if(NOT MSVC)
        # Modules need RConfigure.h copied into include/.
        set(ROOTCINTDEP rootcling rconfigure)
      endif()
```
AFAIK, using generator expressions like `$<TARGET_FILE:rootcling>` don't imply dependencies, which is probably exactly why the `set(ROOTCINTDEP rootcling rconfigure)` below is needed.

So if you want to avoid this deadly dependency, it's probably cleaner to only change the dependencies.

Also, if the `rootcling` dependency is apparently not necessary for Windows at all? Why is it needed for the other platforms? Can we maybe just go with:

```CMake
      set(command ${CMAKE_COMMAND} -E env "ROOTIGNOREPREFIX=1" $<TARGET_FILE:rootcling> -rootbuild)
```
If that works, that means the appropriate dependencies are probably already inherited via some other dependencies, and we can remove this line. Doing manual dependency management is always sub-optimal.

bellenot: No, I tried and it fails. There is maybe a bug or it's a MS issue... The PR is only working fine the way it is. And I'll be on vacation starting basically now. So it's basically an emergency PR, but we can review this once I'm back if you want.
N.B. Without this patch, builds on Windows now take ages and many tests can (and will) fail

bellenot:
> Thanks for the PR!
> 
> See the inline comment. If it's about dependencies, I would only remove the line relevant to dependencies, maybe even on all platforms.

You're welcome! And see my inline reply (I have no more time for investigations...)

cmake/modules/RootMacros.cmake
guitargeek: Oh, that's a bummer :( Okay thanks for the explanation! Probably it's specific to MSVC being a multi-configuration generator with some more implicit targets.

Sure, let's see when you're back if these `rootcling` invocations can be harmonized!

PR #19868, opened 2025-09-10
[tree] add missing header

ferdymercury:
Compilation error otherwise with libcxx

See 
https://github.com/root-project/root/issues/19850#issuecomment-3276474407


vepadulano (approved):
Thanks!

PR #19869, opened 2025-09-11
[RF] Add base-case overload for `RooAbsReal::matchArgs` implementation

guitargeek:
In user frameworks based on RooFit, there were linker errors with ROOT `master`:

```txt
undefined reference to RooAbsReal::matchArgs(RooArgSet const&, RooArgSet&, RooArgProxy const&) const
```

This can be fixed by treating the case of no extra arguments to the variadic `matchArgs` function as a separate overload, so it get consistently not inlined.

Thanks @cburgard for noticing and reporting!

Follows up on 85186f3d8e8fea3, which implemented `RooAbsReal::matchArgs` with variadic templates.

### DIFF
--- a/roofit/roofitcore/inc/RooAbsReal.h
+++ b/roofit/roofitcore/inc/RooAbsReal.h
@@ -437,6 +437,8 @@ protected:
     return result;
   }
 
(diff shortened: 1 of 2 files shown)


guitargeek:
Not clear why this was needed. The base case was already covered by the existing overloads.

PR #16972, opened 2024-11-18
[ci] Enable libc++ for the clang build on Linux

dpiparo:
Revive PR https://github.com/root-project/root/pull/15918

PR #13975, opened 2023-10-29
[cmake] Add dev to macOS, dev uses checked stdlib.

Axel-Naumann:
# This Pull request:

## Checklist:

- [ ] tested changes locally
- [ ] updated the docs (if necessary)

This PR fixes #

PR #19870, opened 2025-09-11
[ci] Enable libc++ for the clang build on Linux

ferdymercury:
# This Pull request:

## Changes or fixes:

explore issues after fixes from https://github.com/root-project/root/issues/19850
revive https://github.com/root-project/root/pull/16972
Tests https://github.com/root-project/root/issues/19940
Related: https://github.com/root-project/root/pull/13975/files

### DIFF
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -427,7 +427,7 @@ jobs:
             platform_config: alma10-clang_ninja
             is_special: true
             property: "clang Ninja builtins auto-registration off"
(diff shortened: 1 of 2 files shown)


ferdymercury:
@guitargeek It seems Alma Linux 10 does not have the libcxx-devel available, which means `<filesystem>` etc can not be found, it's only available for g++14 and g++15.

So clang is supposed to work with gnu library headers, not with libc++.

Maybe Fedora is a better alternative for these tests.

PR #19871, opened 2025-09-11
[cmake] substitute deprecated commands

ferdymercury:
leading otherwise to warnings in newer CMake versions

(Please squash on merge.)

### DIFF
--- a/cmake/modules/CheckAssembler.cmake
+++ b/cmake/modules/CheckAssembler.cmake
@@ -6,12 +6,12 @@
 
 # Helper macro for checking binutils available in OS
 macro(root_check_assembler)
-   execute_process(COMMAND ${CMAKE_CXX_COMPILER} ARGS -print-prog-name=as OUTPUT_VARIABLE _as)
+   execute_process(COMMAND ${CMAKE_CXX_COMPILER} -print-prog-name=as OUTPUT_VARIABLE _as ERROR_VARIABLE _as OUTPUT_STRIP_TRAILING_WHITESPACE)
    mark_as_advanced(_as)
    if(NOT _as)
       message(WARNING "Could not find 'as', the assembler used by GCC.")
    else()
-      execute_process(COMMAND ${_as} ARGS --version OUTPUT_VARIABLE _as_version)
+      execute_process(COMMAND ${_as} --version OUTPUT_VARIABLE _as_version ERROR_VARIABLE _as_version OUTPUT_STRIP_TRAILING_WHITESPACE)
       string(REGEX REPLACE "\\([^\\)]*\\)" "" _as_version "${_as_version}")
       string(REGEX MATCH "[1-9]\\.[0-9]+(\\.[0-9]+)?" _as_version "${_as_version}")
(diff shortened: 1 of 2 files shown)


cmake/modules/CheckAssembler.cmake
```diff
@@ -6,12 +6,12 @@
 
 # Helper macro for checking binutils available in OS
 macro(root_check_assembler)
-   exec_program(${CMAKE_CXX_COMPILER} ARGS -print-prog-name=as OUTPUT_VARIABLE _as)
+   execute_process(COMMAND ${CMAKE_CXX_COMPILER} ARGS -print-prog-name=as OUTPUT_VARIABLE _as)
```
pcanal: Are you sure this works as expected? (See https://cmake.org/cmake/help/latest/command/exec_program.html).  There seems to need 'more' change to adapt to the new function API.

ferdymercury:
```suggestion
   execute_process(COMMAND ${CMAKE_CXX_COMPILER} -print-prog-name=as OUTPUT_VARIABLE _as ERROR_VARIABLE _as OUTPUT_STRIP_TRAILING_WHITESPACE)
```

cmake/modules/CheckAssembler.cmake
```diff
@@ -6,12 +6,12 @@
...
    mark_as_advanced(_as)
    if(NOT _as)
       message(WARNING "Could not find 'as', the assembler used by GCC.")
    else()
-      exec_program(${_as} ARGS --version OUTPUT_VARIABLE _as_version)
+      execute_process(COMMAND ${_as} ARGS --version OUTPUT_VARIABLE _as_version)
```
ferdymercury:
```suggestion
      execute_process(COMMAND ${_as} --version OUTPUT_VARIABLE _as_version ERROR_VARIABLE _as_version OUTPUT_STRIP_TRAILING_WHITESPACE)
```

cmake/modules/CheckCompiler.cmake
```diff
@@ -65,7 +65,7 @@ endif()
 
 #----Test if clang setup works----------------------------------------------------------------------
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-  exec_program(${CMAKE_CXX_COMPILER} ARGS "--version 2>&1 | grep version" OUTPUT_VARIABLE _clang_version_info)
+  execute_process(COMMAND ${CMAKE_CXX_COMPILER} ARGS "--version 2>&1 | grep version" OUTPUT_VARIABLE _clang_version_info)
```
ferdymercury:
```suggestion
  execute_process(COMMAND ${CMAKE_CXX_COMPILER} "--version 2>&1 | grep version" OUTPUT_VARIABLE _clang_version_info ERROR_VARIABLE _clang_version_info OUTPUT_STRIP_TRAILING_WHITESPACE)
```

cmake/modules/CheckAssembler.cmake
ferdymercury: Good catch. Amended.

PR #19872, opened 2025-09-12
[clang-tidy] modernize deprecated headers

ferdymercury:
https://rocm.docs.amd.com/projects/llvm-project/en/latest/LLVM/clang-tools/html/clang-tidy/checks/modernize/deprecated-headers.html

### DIFF
--- a/math/mathcore/inc/Math/Math.h
+++ b/math/mathcore/inc/Math/Math.h
@@ -21,7 +21,8 @@
 
 #if defined(__sun) || defined(_MSC_VER)
 //Microsoft and solaris definition of cmath does not include math.h which has the definitions of numerical constants
-#include <cmath>
+#include <math.h> // for M_PI
+// TODO replace with std::numbers::pi once minimum version is C++20, and remove this code block
 #endif
 
 


math/vecops/inc/ROOT/RVec.hxx
```diff
@@ -24,7 +24,7 @@
       #ifndef _USE_MATH_DEFINES
          #define _USE_MATH_DEFINES
       #endif
-      #include <math.h>
+      #include <cmath>
```
guitargeek: I think we should keep `math.h` in this particular case for now, because `cmath` doesn't define `M_PI`.

We can migrate to the modern C++ version of pi (`std::numbers::pi`) once we use C++20 as the minimum standard version:
https://cppreference.net/cpp/numeric/constants.html

Then, also this `math.h` include will go away.

smuzaffar:
@ferdymercury @guitargeek , while testing latest ROOT master changes we notice massive failure like [a] (`fatal error: cannot open file .../stdlib.h`) . Could these be due to this change in ROOT ? Looks like rootcling is trying to find the header `stdlib.h` in the build path which is not available during runtime.

[a]  https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-ff4129/48126/addOnTests/hlt_data_PRef/step2.log
```
fatal error: cannot open file '/data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/el8_amd64_gcc12/external/gcc/12.3.1-40d504be6370b5a30e3947a6e575ca28/bin/../lib/gcc/x86_64-redhat-linux-gnu/12.3.1/../../../../include/c++/12.3.1/stdlib.h': No such file or directory
15-Sep-2025 16:47:10 CEST  Closed file file:RelVal_Raw_PRef_DATA.root
----- Begin Fatal Exception 15-Sep-2025 16:47:10 CEST-----------------------
An exception of category 'FatalRootError' occurred while
   [0] Constructing the EventProcessor
   [1] Constructing module: class=L2MuonProducer label='hltL2Muons'
   [2] While initializing meta data for branch: L2MuonTrajectorySeedsToManyL2MuonTrajectorySeedsAssociation_hltL2Muons__HLTPRef.
   [3] Calling edm::TypeWithDict::byName()
   Additional Info:
      [a] Getting TClass for edm::AssociationMap<edm::OneToMany<std::vector<L2MuonTrajectorySeed>,std::vector<L2MuonTrajectorySeed>,unsigned int> >
      [b] Fatal Root Error: @SUB=TInterpreter::AutoParse
Error parsing payload code for class L2MuonTrajectorySeed with content:

#line 1 "DataFormatsMuonSeed_xr dictionary payload"
```

ferdymercury:
> Could these be due to this change in ROOT ? Looks like rootcling is trying to find the header `stdlib.h` in the build path which is not available during runtime.

Maybe. Does it also happen with a clean-build from scratch?

guitargeek:
Thanks for the report @smuzaffar! Can you try with
  * https://github.com/root-project/root/pull/19893

?

PR #19866, opened 2025-09-10
Cleanup also current file in TChain::GetEntries 

vepadulano:
Followup to https://github.com/root-project/root/pull/19817

In the method we need to check if the current TTree has a TTreeCache associated with the current file. The previous logic was getting the pointer to the current file but calling LoadTree afterwards, which deleted the current file and switched it for a new instance. In turn, this would cause bad reads as highlighted by a run of the associated unit test with the address sanitizer:

```
==18501==ERROR: AddressSanitizer: heap-use-after-free on address 0x61800004be60 at pc 0x000108735e2c bp 0x00016b501950 sp 0x00016b501948
READ of size 8 at 0x61800004be60 thread T0
    #0 0x000108735e28 in TFile::GetCacheRead(TObject const*) const TFile.cxx:1262
    https://github.com/root-project/root/issues/1 0x00010a6e0758 in TTree::GetReadCache(TFile*) const TTree.cxx:6403
    https://github.com/root-project/root/pull/2 0x00010a6e79c0 in TTree::GetReadCache(TFile*, bool) TTree.cxx:6416
    https://github.com/root-project/root/pull/3 0x00010acd2f58 in TTreeReader::Restart() TTreeReader.cxx:574
    https://github.com/root-project/root/pull/4 0x00010acd2634 in TTreeReader::SetEntriesRange(long long, long long) TTreeReader.cxx:550
    https://github.com/root-project/root/pull/5 0x0001048fd23c in main test_ttreereader_uninitialized_chain.cpp:28
    https://github.com/root-project/root/pull/6 0x000190a96b94 in start+0x17b8 (dyld:arm64e+0xfffffffffff3ab94)

0x61800004be60 is located 480 bytes inside of 824-byte region [0x61800004bc80,0x61800004bfb8)
freed by thread T0 here:
    #0 0x00010c757b0c in _ZdlPv+0x74 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x4bb0c)
    https://github.com/root-project/root/issues/1 0x000107556f6c in TStorage::ObjectDealloc(void*) TStorage.cxx:325
    https://github.com/root-project/root/pull/2 0x0001074a8a68 in TObject::operator delete(void*) TObject.cxx:1189
    https://github.com/root-project/root/pull/3 0x000108732650 in TFile::~TFile() TFile.cxx:569
    https://github.com/root-project/root/pull/4 0x00010a5c6e04 in TChain::LoadTree(long long) TChain.cxx:1449
[...]

PR #19873, opened 2025-09-12
Rework TChain::GetEntries to avoid side-effects 

### LINKED ISSUE #19814: Regression in 6.34: segfault when using TTreeReader on partially initialized TChain

vepadulano:
Fixes #19814 and also the recent CI failures such as https://github.com/root-project/root/actions/runs/17610706130/job/50031825233?pr=19860#step:6:4762

The second commit contains the main change: reworking the `TChain::GetEntries` method. The issue with it was that it was causing one side-effect too many: the last opened file was not being properly closed. This in turn triggered the if condition in `TTreeReader::Restart` and the segfaults due to bad memory accesses. This is an alternative solution to https://github.com/root-project/root/pull/19866 . The advantage IMHO is that it requires changing fewer places and better conveys the intent of the method: a for loop opening the files as opposed to calling LoadTree until all files have been opened. On top of course of having far less side-effects (AFAICT none).

### DIFF
--- a/tree/tree/src/TChain.cxx
+++ b/tree/tree/src/TChain.cxx
@@ -905,14 +905,24 @@ Long64_t TChain::GetEntries() const
       // and `LoadTree` will be no-op.
       if (kLoadTree & fFriendLockStatus)
          return fEntries;
-      const auto readEntry = fReadEntry;
-      auto *thisChain = const_cast<TChain *>(this);
-      thisChain->LoadTree(TTree::kMaxEntries - 1);
-      thisChain->InvalidateCurrentTree();
-      if (readEntry >= 0)
-         thisChain->LoadTree(readEntry);
-      else
-         thisChain->fReadEntry = readEntry;
+      Long64_t totalEntries{};
+      for (auto chainEl : ROOT::Detail::TRangeStaticCast<TChainElement>(fFiles)) {
+         if (chainEl->GetEntries() != TTree::kMaxEntries) {
+            totalEntries += chainEl->GetEntries();
+            continue;
+         }
+         TDirectory::TContext ctxt;
+         std::unique_ptr<TFile> curFile{TFile::Open(chainEl->GetTitle(), "READ_WITHOUT_GLOBALREGISTRATION")};
+         if (!curFile || curFile->IsZombie()) {
+            continue;
+         }
+         std::unique_ptr<TTree> curTree{curFile->Get<TTree>(chainEl->GetName())};
+         if (!curTree) {
+            continue;
+         }
+         totalEntries += curTree->GetEntries();
+      }
+      const_cast<TChain *>(this)->fEntries = totalEntries;
    }
    return fEntries;
 }
--- a/tree/tree/src/TChainElement.cxx
+++ b/tree/tree/src/TChainElement.cxx
@@ -30,7 +30,7 @@ TChainElement::TChainElement() : TNamed(),fBaddress(nullptr),fBaddressType(0),
 {
    fNPackets   = 0;
    fPackets    = nullptr;
-   fEntries    = 0;
+   fEntries    = TTree::kMaxEntries;
    fPacketSize = 100;
    fStatus     = -1;
    ResetBit(kHasBeenLookedUp);
@@ -45,7 +45,7 @@ TChainElement::TChainElement(const char *name, const char *title)
 {
    fNPackets   = 0;
    fPackets    = nullptr;
-   fEntries    = 0;
+   fEntries    = TTree::kMaxEntries;
    fPacketSize = 100;
    fStatus     = -1;
    ResetBit(kHasBeenLookedUp);
--- a/tree/treeplayer/src/TTreeReader.cxx
+++ b/tree/treeplayer/src/TTreeReader.cxx
@@ -567,14 +567,9 @@ void TTreeReader::Restart()
    fProxiesSet = false; // we might get more value readers, meaning new proxies.
    fEntry = -1;
    if (const auto curFile = fTree->GetCurrentFile()) {
-      if (!fTree->GetTree()) {
-         fTree->LoadTree(0);
-      }
-      if (fTree->GetTree()) {
-         if (auto tc = fTree->GetTree()->GetReadCache(curFile, true)) {
-            tc->DropBranch("*", true);
-            tc->ResetCache();
-         }
+      if (auto tc = fTree->GetTree()->GetReadCache(curFile, true)) {
+         tc->DropBranch("*", true);
+         tc->ResetCache();
       }
    }
 }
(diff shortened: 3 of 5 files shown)


vepadulano:
Turns out there are even more side-effects caused by `TChain::GetEntries` than I could see/foresee :)

tree/tree/src/TChain.cxx
```diff
@@ -905,14 +905,20 @@ Long64_t TChain::GetEntries() const
...
-      const auto readEntry = fReadEntry;
-      auto *thisChain = const_cast<TChain *>(this);
-      thisChain->LoadTree(TTree::kMaxEntries - 1);
-      thisChain->InvalidateCurrentTree();
-      if (readEntry >= 0)
-         thisChain->LoadTree(readEntry);
-      else
-         thisChain->fReadEntry = readEntry;
+      Long64_t totalEntries{};
+      for (auto chainEl : ROOT::Detail::TRangeStaticCast<TChainElement>(fFiles)) {
```
pcanal, line 909: One notable 'efficiency' degradation.  This opens all files (except if the last file has already been reached).  The previous implementation was 'only' opening the files that had not already been opened.

There seems to be some 'bug' in TChainElement (it seems confused on whether the default value of fEntries is zero or `TTree::kMaxEntries`
```suggestion
      for (auto chainEl : ROOT::Detail::TRangeStaticCast<TChainElement>(fFiles)) {
        if (chainEl->fEntries != 0 && chainEl->fEntries != TTree::kMaxEntries) {
           totalEntries += chainEl->GetEntries();
           continue;
        }
```

vepadulano:
@pcanal  I implemented the changes we discussed in the last commit. Let me know what you think and let's see if the CI will agree with us.

tree/tree/src/TChainElement.cxx
```diff
@@ -30,7 +30,7 @@ TChainElement::TChainElement() : TNamed(),fBaddress(nullptr),fBaddressType(0),
 {
    fNPackets   = 0;
    fPackets    = nullptr;
-   fEntries    = 0;
+   fEntries = TTree::kMaxEntries;
```
ferdymercury: Should this maybe be kMaxEntries+1 ?
Since otherwise kMaxEntries can not be safely reached?

vepadulano: I don't understand where the question is coming from, is there any actual use case for this? Where should `kMaxEntries` be safely reached? `kMaxEntries = std::numeric_limits<Long64_t>::max();` so `kMaxEntries+1` seems like a really bad idea anyway. I cannot see it being done anywhere in the codebase (hoping I'm not missing something).

ferdymercury: Fair enough, thanks for the explanation. Could you instead modify the following in TTree.h:

```
- // Used as the max value for any TTree range operation.
+ /// Used as the max value for any TTree range operation. The Nentries in a TTree should be strictly smaller than this value.
```

vepadulano: That can be done, but I'll keep it for another PR

vepadulano:
> Also, does this change with the cache maybe impact https://its.cern.ch/jira/browse/ROOT-7973 ?

I don't think so. I think the two issues are not related.

tree/treeplayer/src/TTreeReader.cxx
```diff
@@ -567,14 +567,9 @@ void TTreeReader::Restart()
    fProxiesSet = false; // we might get more value readers, meaning new proxies.
    fEntry = -1;
    if (const auto curFile = fTree->GetCurrentFile()) {
-      if (!fTree->GetTree()) {
```
pcanal, line 570: Why is removing the check for `nullptr` an improvement (and/or why is it guaranteed to not be `nullptr` (in which case we should add a comment explaining why it is)).

vepadulano: The `nullptr` check is fine but the call to `LoadTree` was not necessary. In the latest commit I have added nullptr checks for all variables in this part of the function.

PR #19874, opened 2025-09-12
[roottest] adjust root/hist tests

linev:
Remove unused rootlogon.C script
Better format cmake files
Adjust C++ formating in some tests

### DIFF
--- a/roottest/root/hist/CMakeLists.txt
+++ b/roottest/root/hist/CMakeLists.txt
@@ -5,6 +5,7 @@ ROOTTEST_ADD_TEST(mergeTH2Poly
 ROOTTEST_ADD_TEST(assertSparseToTHn
                   MACRO assertSparseToTHn.C
                   COPY_TO_BUILDDIR effectiveCorrection.root)
+
 if(fortran AND CMAKE_Fortran_COMPILER AND NOT CMAKE_Fortran_COMPILER_ID STREQUAL Intel)
   ROOTTEST_ADD_TEST(h2root
                     COMMAND h2root mb4i1.hbook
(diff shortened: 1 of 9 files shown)


roottest/root/hist/CMakeLists.txt
```diff
@@ -1,10 +1,10 @@
 ROOTTEST_ADD_TEST(mergeTH2Poly
-                  MACRO th2PolyMerge.C
-                  ${WILLFAIL_ON_WIN32})
```
pcanal: This test appears to still be 'failing' on Windows.

linev: I just want to see why it fails. It creates different number of entries.

pcanal: Indeed on Windows the output is: 
```
Processing C:/ROOT-CI/src/roottest/root/hist/th2PolyMerge.C...
We expect two warnings to be printed:
 1) About different number of bins
 2) About different contours
totEntries 4599
(int) 1
```
which return early before it can print the expected warnings:
```
Processing th2PolyMerge.C...
We expect two warnings to be printed:
 1) About different number of bins
 2) About different contours
totEntries 4548
Error in <TH2Poly::Add>: Attempt to add histograms with different number of bins
Warning in <TH2Poly::Merge>: An issue was encountered during the merge operation.
totEntries additional bin 0
Error in <TH2Poly::Add>: Attempt to add histograms with different bin limits
Warning in <TH2Poly::Merge>: An issue was encountered during the merge operation.
totEntries diff contour 0
(int) 0
```

PR #19875, opened 2025-09-12
[ntuple] Remove type name from public field constructors

jblomer:
Several RFieldBase descendants still allow for passing the type name in a public constructor. This is dangerous as it potentially allows to write arbitrary type names to disk. Remove / rework these constructors.

### DIFF
--- a/tree/ntuple/inc/ROOT/RField/RFieldProxiedCollection.hxx
+++ b/tree/ntuple/inc/ROOT/RField/RFieldProxiedCollection.hxx
@@ -281,7 +281,13 @@ public:
 /// The generic field for a `std::map<KeyType, ValueType>` and `std::unordered_map<KeyType, ValueType>`
 class RMapField : public RProxiedCollectionField {
 public:
-   RMapField(std::string_view fieldName, std::string_view typeName, std::unique_ptr<RFieldBase> itemField);
+   enum class EMapType {
+      kMap,
+      kUnorderedMap,
+      kMultiMap,
+      kUnorderedMultiMap
+   };
+   RMapField(std::string_view fieldName, EMapType mapType, std::unique_ptr<RFieldBase> itemField);
    RMapField(RMapField &&other) = default;
    RMapField &operator=(RMapField &&other) = default;
    ~RMapField() override = default;
@@ -296,7 +302,7 @@ public:
    }
 
    explicit RField(std::string_view name)
-      : RMapField(name, TypeName(), std::make_unique<RField<std::pair<KeyT, ValueT>>>("_0"))
+      : RMapField(name, EMapType::kMap, std::make_unique<RField<std::pair<KeyT, ValueT>>>("_0"))
    {
    }
    RField(RField &&other) = default;
@@ -313,7 +319,7 @@ public:
    }
 
    explicit RField(std::string_view name)
-      : RMapField(name, TypeName(), std::make_unique<RField<std::pair<KeyT, ValueT>>>("_0"))
+      : RMapField(name, EMapType::kUnorderedMap, std::make_unique<RField<std::pair<KeyT, ValueT>>>("_0"))
    {
    }
    RField(RField &&other) = default;
@@ -330,7 +336,7 @@ public:
    }
 
    explicit RField(std::string_view name)
-      : RMapField(name, TypeName(), std::make_unique<RField<std::pair<KeyT, ValueT>>>("_0"))
+      : RMapField(name, EMapType::kMultiMap, std::make_unique<RField<std::pair<KeyT, ValueT>>>("_0"))
    {
    }
    RField(RField &&other) = default;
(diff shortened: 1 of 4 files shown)


tree/ntuple/src/RField.cxx
```diff
@@ -761,9 +761,10 @@ void ROOT::RBitsetField::AcceptVisitor(ROOT::Detail::RFieldVisitor &visitor) con
 
 //------------------------------------------------------------------------------
 
-ROOT::RNullableField::RNullableField(std::string_view fieldName, std::string_view typeName,
+ROOT::RNullableField::RNullableField(std::string_view fieldName, const std::string &typePrefix,
                                      std::unique_ptr<RFieldBase> itemField)
-   : ROOT::RFieldBase(fieldName, typeName, ROOT::ENTupleStructure::kCollection, false /* isSimple */)
+   : ROOT::RFieldBase(fieldName, typePrefix + "<" + itemField->GetTypeName() + ">", ROOT::ENTupleStructure::kCollection,
```
pcanal, line 766: As implied by:
> This is dangerous as it potentially allows to write arbitrary type names to disk.

doesn't the 'prefix' then also need to be normalized?  (At least here) wasn't the 'real' issue that the 2 types (`typeName` and the one derived from `itemField->GetTypeName()` being potentially different?

tree/ntuple/inc/ROOT/RField/RFieldProxiedCollection.hxx
```diff
@@ -365,7 +368,13 @@ public:
 /// The generic field for a `std::set<Type>` and `std::unordered_set<Type>`
 class RSetField : public RProxiedCollectionField {
 public:
-   RSetField(std::string_view fieldName, std::string_view typeName, std::unique_ptr<RFieldBase> itemField);
+   enum class ESetType {
+      kSet,
+      kUnorderedSet,
+      kMultiSet,
+      kUnorderedMultiSet
```
pcanal, line 375: Is this enum set limited to the actual STL collection?  How is a proxied map-like collection handled?

tree/ntuple/src/RField.cxx
jblomer: No, because the constructor that accepts a prefix is protected. I could have used the full type name instead of the type prefix approach. The reason I didn't is that the code calling this constructor would then both get the type name from the item field _and_ move the item field to the base class. And doing both these things in a single call is illegal.

tree/ntuple/inc/ROOT/RField/RFieldProxiedCollection.hxx
jblomer: A custom collection proxy is not created via `R[Set|Map]Field` but through their base class `RProxiedCollectionField`. This base class accepts a type name but it validates it though `TClass`. So arguably for `RSetField` / `RMapField` the situation was not quite that bad...

tree/ntuple/src/RField.cxx
pcanal: Wasn't the previous constructor also protected?

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -703,8 +708,11 @@ ROOT::RProxiedCollectionField::RProxiedCollectionField(std::string_view fieldNam
 std::unique_ptr<ROOT::RFieldBase> ROOT::RProxiedCollectionField::CloneImpl(std::string_view newName) const
 {
    auto newItemField = fSubfields[0]->Clone(fSubfields[0]->GetFieldName());
-   return std::unique_ptr<RProxiedCollectionField>(
-      new RProxiedCollectionField(newName, GetTypeName(), std::move(newItemField)));
+   auto clone =
+      std::unique_ptr<RProxiedCollectionField>(new RProxiedCollectionField(newName, fProxy->GetCollectionClass()));
```
hahnjo, line 730: While you are modifying the line, maybe use `std::make_unique`?

hahnjo (commented):
In principle LGTM. It smells a bit that we have to repeat the field type names in a number of places and sometimes passed through `TClass`, but I don't have a better proposal for the moment. See also question by Philippe regarding the constructor of `RNullableField`.

tree/ntuple/src/RFieldMeta.cxx
jblomer: This constructor of `RProxiedCollectionField` is protected.

tree/ntuple/src/RField.cxx
jblomer: Yes, the problem was not with `RNullableField` but with its descendants `ROptionalField` and `RUniquePtrField` that do have a public constructor.

hahnjo (approved):

tree/ntuple/src/RField.cxx
jblomer: Can this be resolved?

jblomer: @pcanal ping

jblomer: Resolving, please feel free to open a new issue if there are still questions.

PR #19877, opened 2025-09-12
[ntuple] cache number of entries in reader

jblomer:
Follow up of #19763

### DIFF
--- a/tree/ntuple/inc/ROOT/RNTupleReader.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleReader.hxx
@@ -82,6 +82,8 @@ private:
    /// Retrieving descriptor data from an RNTupleReader is supposed to be for testing and information purposes,
    /// not on a hot code path.
    std::optional<ROOT::RNTupleDescriptor> fCachedDescriptor;
+   /// We know that the RNTupleReader is always reading a single RNTuple, so the number of entries is fixed.
+   ROOT::NTupleSize_t fNEntries = 0;
    Experimental::Detail::RNTupleMetrics fMetrics;
    /// If not nullopt, these will be used when creating the model
    std::optional<ROOT::RNTupleDescriptor::RCreateModelOptions> fCreateModelOptions;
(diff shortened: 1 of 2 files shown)


tree/ntuple/src/RNTupleReader.cxx
```diff
@@ -258,6 +258,13 @@ const ROOT::RNTupleDescriptor &ROOT::RNTupleReader::GetDescriptor()
    return *fCachedDescriptor;
 }
 
+ROOT::NTupleSize_t ROOT::RNTupleReader::GetNEntries() const
+{
+   if (fCachedNEntries == ROOT::kInvalidNTupleIndex)
+      fCachedNEntries = fSource->GetNEntries();
```
pcanal: Is there any way `fSource` could be a `nullptr`?  Either way, could the `fCachedNEntries` be set as soon as `fSource` is set?

tree/ntuple/inc/ROOT/RNTupleReader.hxx
```diff
@@ -82,6 +83,8 @@ private:
    /// Retrieving descriptor data from an RNTupleReader is supposed to be for testing and information purposes,
    /// not on a hot code path.
    std::optional<ROOT::RNTupleDescriptor> fCachedDescriptor;
+   /// We know that the RNTupleReader is always reading a single RNTuple, so the number of entries is fixed.
+   mutable std::atomic<ROOT::NTupleSize_t> fCachedNEntries = kInvalidNTupleIndex;
```
silverweed: Why does this need to be atomic? It only gets written under a lock and it's then only read afterwards, no?

tree/ntuple/src/RNTupleReader.cxx
jblomer: Good point. Changed.

tree/ntuple/inc/ROOT/RNTupleReader.hxx
jblomer: See above, changed.

pcanal (approved):

tree/ntuple/inc/ROOT/RNTupleReader.hxx
```diff
@@ -82,6 +82,8 @@ private:
    /// Retrieving descriptor data from an RNTupleReader is supposed to be for testing and information purposes,
    /// not on a hot code path.
    std::optional<ROOT::RNTupleDescriptor> fCachedDescriptor;
+   /// We know that the RNTupleReader is always reading a single RNTuple, so the number of entries is fixed.
+   ROOT::NTupleSize_t fCachedNEntries = kInvalidNTupleIndex;
```
silverweed: Does this default make sense? Wouldn't 0 be a more apt choice?

silverweed: Also, maybe it's fine to just call it `fNEntries` at this point?

PR #19881, opened 2025-09-15
[roottest] use cmake for cling tests [part 1]

linev:
1. Convert cling/template and cling/reflex to cmake
2. Remove tests which not run with Makefile and cannot be run - while Cintex was removed
3. Remove unused Makefile and .rootrc files
4. Do not copy macros to build directory - they can run from source dir as well


roottest/cling/reflex/CMakeLists.txt
```diff
@@ -1,7 +1,41 @@
...
-# define a CTest test that calls 'make' in ${CMAKE_CURRENT_SOURCE_DIR}
-#
-#-------------------------------------------------------------------------------
-ROOTTEST_ADD_OLDTEST(LABELS longtest)
+ROOTTEST_GENERATE_REFLEX_DICTIONARY(classVersion
+                                    classVersion.h
+                                    SELECTION classVersion_selection.xml
+                                    FIXTURES_SETUP cling-reflex-classVersion-fixture)
+
+# test fail on Windows - with warning:  no dictionary for class WithClassVersion is available
```
pcanal, line 6: This is strange.  I don't (easily) see any reason for the failure.  The dictionary for `WithClassVersion` should be generated and it is explicitly loaded .... At the very least we should open an issue to have this investigated.

linev: I create issue https://github.com/root-project/root/issues/20412

PR #19882, opened 2025-09-15
[CMake] Always compile `gui/treemap`

hahnjo:
It is unconditionally used by `ROOTNTupleBrowse`.

### DIFF
--- a/gui/CMakeLists.txt
+++ b/gui/CMakeLists.txt
@@ -10,6 +10,7 @@ add_subdirectory(fitpanel)
 add_subdirectory(guibuilder)
 add_subdirectory(guihtml)
 add_subdirectory(recorder)
+add_subdirectory(treemap)
 
 if(webgui)
    if(cefweb)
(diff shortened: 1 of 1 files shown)


silverweed (approved):
Thanks!
cc @magnustymoteus

guitargeek:
This PR doesn't completely solve the compilation failures with ROOT master for me :slightly_frowning_face: I build with `webgui=OFF` and `root7=OFF`. This doesn't work now because `gui/treemap` is always built with this PR, but it depends on `ROOTNTupleUtil`, which is barred behind `root7`:
https://github.com/root-project/root/blob/master/tree/ntupleutil/CMakeLists.txt#L12

Can this also be fixed in this PR? I think it's maybe better to address all breakages caused by 5ba79a9a81e12f in one single PR to keep a better overview, IMHO.

hahnjo:
It's a separate PR because it needs more changes: https://github.com/root-project/root/pull/19883

PR #19883, opened 2025-09-15
[ntuple] Move `ntupleutil` out of ROOT 7

hahnjo:
All RNTuple code should always compile, while some interfaces remain in `ROOT::Experimental`.


jblomer (approved):
At some point we need to remove the ROOT7 banner from the files, too. But we can do that if code moves out of experimental.

PR #19887, opened 2025-09-15
[builtins] Update nlohmann::json from 3.10.5 to 3.12.0

silverweed:
This gets rid of the deprecation warning about `operator ""`

### DIFF
--- a/builtins/nlohmann/json.hpp
+++ b/builtins/nlohmann/json.hpp
@@ -1,31 +1,10 @@
-/*
-    __ _____ _____ _____
- __|  |   __|     |   | |  JSON for Modern C++
-|  |  |__   |  |  | | | |  version 3.10.5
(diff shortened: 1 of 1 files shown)


pcanal:
I assume its all fine :), github web says: "8,906 additions, 5,320 deletions not shown because the diff is too large. Please use a local Git client to view these changes. " ! :)

silverweed:
> I assume its all fine :), github web says: "8,906 additions, 5,320 deletions not shown because the diff is too large. Please use a local Git client to view these changes. " ! :)

According to the [release notes](https://github.com/nlohmann/json/releases), both minor versions should be backward compatible.
An alternative to a full upgrade would be to selectively patch those operators and preserve the vendored version as-is, but we most likely will at some point need to upgrade (e.g. if we switch to c++20)

Issue #19867, opened 2025-09-10
ROOT dictionary crash related to default template arguments


scott-snyder:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

hi -

Since the Sep. 4 nightly, we've been seeing a failure in the dev3LCG ATLAS
build.  This can be seen within that build by simply doing:

```
$ . /cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/user/atlasLocalSetup.sh
$ asetup main--dev3LCG,Athena,gcc15,r2025-09-08*
$ python
>>> import ROOT
>>> ROOT.TClass.GetClass('DataVector<INavigable4Momentum,DataModel_detail::NoBase>')
 *** Break *** segmentation violation
```

If i switch to the debug build of the ROOT libraries, i instead get this
assertion failure:

```
python: /build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-HEAD/src/ROOT/HEAD/interpreter/cling/lib/Interpreter/Transaction.cpp:98: void cling::Transaction::addNestedTransaction(cling::Transaction*): Assertion `!m_Unloading && "Must not nest within unloading transaction"' failed.
```

with this stack trace:

```
#4  0x00007ffff7437886 in __assert_fail () from /lib64/libc.so.6
#5  0x00007fffecb04356 in cling::Transaction::addNestedTransaction (
    this=0x86240b0, nested=0x8555310)
    at /build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-HEAD/src/ROOT/HEAD/interpreter/cling/lib/Interpreter/Transaction.cpp:98
#6  0x00007fffecc053a7 in cling::IncrementalParser::beginTransaction (
    this=0x7a7d80, Opts=...)
    at /build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-HEAD/src/ROOT/HEAD/interpreter/cling/lib/Interpreter/IncrementalParser.cpp:497
#7  0x00007fffecad6190 in cling::Interpreter::PushTransactionRAII::PushTransactionRAII (this=0x7ffffffeec40, i=0x45e540)
    at /build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-HEAD/src/ROOT/HEAD/interpreter/cling/lib/Interpreter/Interpreter.cpp:107
#8  0x00007fffec7feb24 in TCling::LoadEnums (this=0x53e970, enumList=...)
[...]

PR #19892, opened 2025-09-15
[core] Extend `PushTransactionRAII` in `TClingTypeInfo::Property()`

hahnjo:
This fixes a failing assertion `getState() == kCollecting` during `pyunittests-roottest-root-ntuple-atlas-datavector-rentry-getptr` on `mac13` since effectively enabling assertions in Cling with commit 842dad25d9.

It is an alternative to commit a54f6eeff0 ("[core] Handle deserialization during TCling::AutoLoad") that causes issues with autoparsing and (suppressed) default template arguments as reported in https://github.com/root-project/root/issues/19867.

FYI @scott-snyder


hahnjo:
> LGTM. Reducing the scope/frequency of the transactions is indeed usually a gain.

Note that this is not fully clear here: In principle, there can be many calls to `TClingTypeInfo::Property()` per autoload. That's why I put it on the outermost level at first. However, having a single transaction in the case of classes is probably beneficial, at the expense for enums.

PR #19893, opened 2025-09-15
[core] Include `stdlib.h` when running rootcling

guitargeek:
Follows up on 0f338acb8f53, including the `stdlib.h` header to prevent later lookup that might fail in the CMSSW environment.

### DIFF
--- a/core/dictgen/src/rootcling_impl.cxx
+++ b/core/dictgen/src/rootcling_impl.cxx
@@ -2805,6 +2805,7 @@ void CreateDictHeader(std::ostream &dictStream, const std::string &main_dictname
                // write one and include in there a few things for backward
                // compatibility.
                << "\n/*******************************************************************/\n"
+               << "#include <stdlib.h>\n" // Has been included historically and CMSSW relies on that
                << "#include <cstddef>\n"
                << "#include <cstdio>\n"
                << "#include <cstdlib>\n"
@@ -4473,7 +4474,8 @@ int RootClingMain(int argc,
       }
    }
 
-   if (interp.declare("#include <string>\n" // For the list of 'opaque' typedef to also include string.
+   if (interp.declare("#include <stdlib.h>\n" // Has been implicitly included historically and CMSSW relies on that
+                      "#include <string>\n" // For the list of 'opaque' typedef to also include string.
                       "#include <RtypesCore.h>\n" // For initializing TNormalizedCtxt.
                       "namespace std {} using namespace std;") != cling::Interpreter::kSuccess) {
       ROOT::TMetaUtils::Error(nullptr, "Error loading the default header files.\n");


guitargeek:
FYI @smuzaffar

smuzaffar:
thanks @guitargeek , I have started cmssw tests via https://github.com/cms-sw/root/pull/226

smuzaffar:
cmssw tests still fail. I noticed that if we do not build cuda.pcm during root build then we do not get this missing header error. It might be again the issue we have seen in https://github.com/cms-sw/cmssw/issues/47774#issuecomment-2893305330 . I am now testing https://github.com/cms-externals/HLS_arbitrary_Precision_Types/pulls ( replace `stdlib.h` with` cstdlib`) to see if cmssw tests pass with newer root master

guitargeek:
Thank you! So I'll wait for your next report before continuing the investigation on the ROOT side

smuzaffar:
@guitargeek , cleaning up our hls external to not have `stdlib.h` helped (now number of failures are a lot less ) but we still have same error ( see https://github.com/cms-sw/cmsdist/pull/10077#issuecomment-3299220402 ). While I am still testing this change but I was worndering if we also need to add `stdlib.h` in https://github.com/root-project/root/pull/19864/files#diff-ad804f06151652d3de9c1ff33a9cf01433c632a09aba63972d51a9c6bca572ad

guitargeek:
Oh, I missed that! Good catch. Is it okay if I update this PR right now so you can try it out?

In any case, I think the update to C++ headers to `HLS_arbitrary_Precision_Types` is a good change to do. You intend to make that change even if the problem can be fixed on the ROOT side? Since you also have it in your `clang-tidy`:
https://github.com/cms-sw/cmssw/blob/master/.clang-tidy#L7
And using the C++ headers consistently if already enforced in `cmssw` would make sense to me.

smuzaffar:
> Oh, I missed that! Good catch. Is it okay if I update this PR right now so you can try it out?

yes please feel free to update this PR

smuzaffar:
>  You intend to make that change even if the problem can be fixed on the ROOT side? 

yes, https://github.com/cms-externals/HLS_arbitrary_Precision_Types/pull/1 is independent PR and will go in regardless of root changes

guitargeek:
> yes please feel free to update this PR

Ok! It's updated

core/dictgen/src/rootcling_impl.cxx
```diff
@@ -2805,6 +2805,7 @@ void CreateDictHeader(std::ostream &dictStream, const std::string &main_dictname
                // write one and include in there a few things for backward
                // compatibility.
                << "\n/*******************************************************************/\n"
+               << "#include <stddef.h>\n" // Has been included historically and CMSSW relies on that
```
smuzaffar: @guitargeek , shouldn't this be `stdlib.h` ?

pcanal: Humm ... why is either of them needed here?  Not having the right header here would lead to a compilation error of the dictionary file and ... would (likely) indicates that there is a missing include in the user header file.

core/dictgen/src/rootcling_impl.cxx
```diff
@@ -4473,7 +4474,8 @@ int RootClingMain(int argc,
       }
    }
 
-   if (interp.declare("#include <string>\n" // For the list of 'opaque' typedef to also include string.
+   if (interp.declare("#include <stdlib.h>\n" // Has been implicitly included historically and CMSSW relies on that
```
pcanal, line 4477: This does not seem to be a historical need but rather something more complicated (if it was just historical, it could be equally fixed by changes (adding missing header) in CMSSW ... this does not appear to be the case).

core/dictgen/src/rootcling_impl.cxx
guitargeek: I agree with both of you :slightly_smiling_face: PR is updated.

And good that you chime in, @pcanal! Indeed it's likely a missing include in the CMSSW code.

So I guess the question is really if we are allowed to beak dictionary compilation by removing includes that we don't want anymore but the user relied on?

My approach was to make sure that users like CMSSW don't have to change their code, but I would be grateful to read your perspective in more detail.

core/dictgen/src/rootcling_impl.cxx
guitargeek: I didn't put much thought into these comments yet and first would like to see if this PR is actually necessary to fix the CMSSW dictionary compilation

core/dictgen/src/rootcling_impl.cxx
pcanal: (a) Was there really a **compilation** failures in CMSSW?
(b) Yes, I think at this point we can afford a small breakage in the dictionary compilation since those would be 'clear' that somehow the user header passed to `rootcling` are not standalone but actually interdependent on how they are included (aka the headers are in my opinion broken :) ).

guitargeek: Sorry, not compilation errors, but errors at runtime in `TInterpreter::AutoParse` (see https://github.com/root-project/root/pull/19872#issuecomment-3292625349).

Are you familiar with these errors?

pcanal: > Sorry, not compilation errors, 

Then this part of the change is not needed.

pcanal:
> but errors at runtime in TInterpreter::AutoParse (see https://github.com/root-project/root/pull/19872#issuecomment-3292625349).
> Are you familiar with these errors?

What I see first is:
```
fatal error: cannot open file '/data/cmsbld/jenkins/workspace/ib-run-pr-tests/testBuildDir/el8_amd64_gcc12/external/gcc/12.3.1-40d504be6370b5a30e3947a6e575ca28/bin/../lib/gcc/x86_64-redhat-linux-gnu/12.3.1/../../../../include/c++/12.3.1/stdlib.h': No such file or directory
```
which seems to be some sort of relocation error combines with `stdlib.h` not being explicitly loaded in the interpreter.  So indeed the change around https://github.com/root-project/root/pull/19893/files#r2353004215 might solve the problem.  However the underlying issue might be the relocation issue (can't find `stdlib.h` if it 'moved' since the module were generated is the "real" issue.  It might also be simply that `stdlib.h` is (no longer) part of a module (which I guess might be yet another workaround).

smuzaffar:
@guitargeek , CMSSW tests passed after fixing hls and clhep headers to not include `stdlib.h`. Latest root master changes are already integrated in our special CMSSW ROOT6 IBs.  I think we do not need this change. Thanks for all the help

guitargeek:
Thanks a lot for working around this on the CMSSW side, @smuzaffar!

PR #19153, opened 2025-06-24
[do not merge] RNTuple Attributes prototype

silverweed:
# This Pull request:
Draft PR for the RNTuple Attributes prototype.
This prototype allows writing, reading and merging TFile-based RNTuples containing Attributes. Note that the implementation is not final and cuts several corners in order to provide a proof of concept.

Note that this new optional feature requires the Binary Format to be bumped by one minor version.

PR #19894, opened 2025-09-16
[ntuple] Attributes: update specs and add de/serialization

silverweed:
First PR of a series to merge the RNTuple Attributes into master. The final result will be [this](https://github.com/root-project/root/pull/19153), although the commits will be reorganized to be more coherent and reviewable.

This first PR updates the binary format specification (introducing a new minor version) and updates the Serializer and Descriptor code to match. This is backward-compatible and no Attribute can be written yet since the writer API will be introduced later.

## Checklist:

- [x] tested changes locally


tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -684,6 +685,46 @@ The entry range allows for finding the right page list for random access request
...
+- a 16-bit unsigned integer containing the Attribute Schema Version.
+- a 48-bit unsigned integer containing the uncompressed size of the Anchor;
+- a Locator pointing to the Attribute RNTuple's Anchor;
+- a String containing the Attribute Set's name;
+
+```
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|   Attribute Schema Version    |                               |
```
jblomer: I think we need to describe the schema version and what it means for readers/writers.
I would also prefer, I think, power-of-two wide integers. I think it's fine to have a 32bit anchor length. We could then have a 16bit schema version and a 16bit reserved or flags field.

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -684,6 +685,46 @@ The entry range allows for finding the right page list for random access request
...
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                      Attribute Set Name                       |
+.                             ...                               .
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+```
+
+All linked Attribute Sets must have a distinct name.
+
```
jblomer, line 702: We should have a description of the attribute ntuple schema here, too. Or  the complete structure of this section with TODOs, if this is more appropriate for a different PR.

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -67,6 +67,28 @@ struct RNTupleClusterBoundaries {
 std::vector<ROOT::Internal::RNTupleClusterBoundaries> GetClusterBoundaries(const RNTupleDescriptor &desc);
 } // namespace Internal
 
+namespace Experimental::Internal {
+
+struct RNTupleAttrSetDescriptor {
+   std::uint64_t fAnchorUncompLen = 0;
```
jblomer:
```suggestion
   std::uint64_t fAnchorLength = 0; ///< uncompressed size of the linked anchor
```

tree/ntuple/doc/BinaryFormatSpecification.md
jblomer: As for the schema version, I'd suggest a semantic versioning scheme where the minor version indicates new, optional internal fields, and the major version indicates a breaking change, e.g. changing from a single range to a vector of ranges. Major and minor could possibly go into 8-bit integers each (if this is not considered too tight). Or we make the schema version 32 bit wide and major and minor go into 16 bit ints each.

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -67,6 +67,28 @@ struct RNTupleClusterBoundaries {
...
+   std::uint16_t fSchemaVersion = 0;
+   // The locator of the AttributeSet anchor.
+   // In case of kTypeFile, it points to the beginning of the Anchor's payload.
+   // NOTE: Only kTypeFile is supported at the moment.
+   RNTupleLocator fLocator;
```
jblomer:
```suggestion
   RNTupleLocator fAnchorLocator;
```

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -645,6 +667,8 @@ and backward compatibility when the metadata evolves.
 class RNTupleDescriptor final {
    friend class Internal::RNTupleDescriptorBuilder;
    friend RNTupleDescriptor Internal::CloneDescriptorSchema(const RNTupleDescriptor &desc);
+   friend const std::vector<Experimental::Internal::RNTupleAttrSetDescriptor> &
+   ROOT::Experimental::Internal::GetAttributeSets(const RNTupleDescriptor &desc);
```
jblomer: I assume this is temporary until the attribute sets move out of experimental?

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -841,6 +867,9 @@ public:
    bool HasFeature(unsigned int flag) const { return fFeatureFlags.count(flag) > 0; }
    std::vector<std::uint64_t> GetFeatureFlags() const;
 
+   /// Returns the names of all Attribute Sets associated to this RNTuple.
+   std::vector<std::string> GetAttributeSetNames() const;
```
jblomer: I wonder if this will stay in the final API. Wouldn't we have internally a map string -> attribute set descriptor, which we query by a `GetAttributeSet(name)`? And for the size a `GetNAttrSets()` and for iteration an `AttrSetIterable`, like we have for the other descriptor types?

tree/ntuple/src/RNTupleSerialize.cxx
```diff
@@ -1918,36 +1965,95 @@ ROOT::RResult<void> ROOT::Internal::RNTupleSerializer::DeserializeFooter(const v
...
+   auto fnBufSizeLeft = [&]() { return bufSize - (bytes - base); };
+
+   std::uint64_t frameSize;
+   if (auto res = DeserializeFrameHeader(bytes, fnBufSizeLeft(), frameSize)) {
+      bytes += res.Unwrap();
+   } else {
+      return R__FORWARD_ERROR(res);
+   }
+   std::uint64_t uncompLenAndSchemaVersion;
+   bytes += DeserializeUInt64(bytes, uncompLenAndSchemaVersion);
```
jblomer: Requires checking if the remaining buffer size is big enough for the 64bit integer.

tree/ntuple/test/ntuple_serialize.cxx
```diff
@@ -2059,3 +2059,25 @@ TEST(RNTuple, SerializeMultiColumnRepresentationDeferredInMainHeader)
    EXPECT_EQ(expect2_0, columnRange2_0);
    EXPECT_EQ(expect2_1, columnRange2_1);
 }
+
+TEST(RNTuple, SerializeAttrSets)
```
jblomer, line 2063: Perhaps add a test that serializes/deserializes an entire ntuple descriptor with attribute sets.

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
silverweed: Depends on whether or not we want `RNTupleAttrSetDescriptor` exposed (not in `Internal`) or not; for now I didn't see a strong reason (other than "symmetry") to have it non-Internal, since, other than the name, the descriptor contains low-level information that are only useful if you want to locate it in the underlying storage (which you should not do directly but rather through the upcoming `AttrSetReader`.
That said, we might also expose the Descriptor publicly and gradually decide which fields we want to expose from it.

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
silverweed: If we go for the same API as the other descriptors (see point above) then yes, this would become redundant

silverweed:
I updated the PR and uniformed the `RNTupleAttrSetDescriptor` with the other descriptor classes.

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -684,6 +685,63 @@ The entry range allows for finding the right page list for random access request
...
+The Attribute Schema Version describes the internal schema of the linked Attribute Set RNTuple.
+A change in Major version number indicates a breaking, non-forward-compatible change in the schema: readers should
+refuse reading an Attribute Set whose Major Schema Version is unknown.
+A change in Minor version number indicates the presence of optional additional fields in the schema: readers should
+still be able to read the Attribute Set as before, ignoring any new field.
+
+The current Attribute Schema Version is **1.0**. It has the following fields (in the following order):
+  1. `_rangeStart` (type `NTupleSize_t`): the start of the range that each Attribute Entry refers to;
+  2. `_rangeLen` (type `NTupleSize_t`): the length of the range that each Attribute Entry refers to;
```
jblomer: I think we may want to elaborate what length zero means

tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -1474,3 +1502,25 @@ ROOT::RNTupleDescriptor::RExtraTypeInfoDescriptorIterable ROOT::RNTupleDescripto
...
+}
+
+bool ROOT::Experimental::RNTupleAttrSetDescriptor::operator==(const RNTupleAttrSetDescriptor &other) const
+{
+   return fAnchorLength == other.fAnchorLength && fSchemaVersionMajor == other.fSchemaVersionMajor &&
+          fSchemaVersionMinor == other.fSchemaVersionMinor && fAnchorLocator == other.fAnchorLocator &&
+          fName == other.fName;
+};
+
+ROOT::Experimental::RNTupleAttrSetDescriptor ROOT::Experimental::RNTupleAttrSetDescriptor::Clone() const
```
jblomer, line 1536: Clone of fAnchorLocator seems missing

tree/ntuple/src/RNTupleSerialize.cxx
```diff
@@ -1799,6 +1799,23 @@ ROOT::RResult<std::uint32_t> ROOT::Internal::RNTupleSerializer::SerializeFooter(
       return R__FORWARD_ERROR(res);
    }
 
+   // Attributes
+   frame = pos;
+   const auto nAttributeSets = desc.GetNAttributeSets();
```
jblomer, line 1804: Let's add a warning if `nAttributeSets > 0` that attributes are still experimental

tree/ntuple/src/RNTupleSerialize.cxx
```diff
@@ -1918,36 +1962,98 @@ ROOT::RResult<void> ROOT::Internal::RNTupleSerializer::DeserializeFooter(const v
...
+   // Testing for > 8 because bufSize includes the checksum.
+   if (fnBufSizeLeft() > 8) {
+      std::uint32_t nAttributeSets;
+      frame = bytes;
+      if (auto res = DeserializeFrameHeader(bytes, fnBufSizeLeft(), frameSize, nAttributeSets)) {
+         bytes += res.Unwrap();
+      } else {
+         return R__FORWARD_ERROR(res);
+      }
+      for (std::uint32_t attrSetId = 0; attrSetId < nAttributeSets; ++attrSetId) {
```
jblomer, line 2012: Same here, if `nAttributeSets > 0` let's print a warning

tree/ntuple/doc/BinaryFormatSpecification.md
silverweed: How prescriptive do we want to be with them? I think we probably don't want to mandate how a reader treats them, but we want to make sure they are recognized as 'valid'. That said, we never say anything about merging in the Specification...

tree/ntuple/src/RNTupleSerialize.cxx
silverweed: Does this have implication on the readability of the data? @hahnjo pointed out that we may want to have a way to write RNTuples with no format changes until we have finalized the decision about the format (which also implies we should not merge any change to the specs until then).
I think we need an internal discussion on this before this PR can proceed (or at least before the specs are merged).

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -627,6 +627,7 @@ The footer envelope has the following structure:
 - Header checksum (XxHash-3 64bit)
 - Schema extension record frame
 - List frame of cluster group record frames
+- List frame of linked Attribute Sets
```
hahnjo: To stay consistent with the rest of the section, should this rather be
```suggestion
- List frame of attribute set record frames
```
?

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -684,6 +685,64 @@ The entry range allows for finding the right page list for random access request
...
+An RNTuple may have zero or more linked Attribute Sets, containing metadata.
+Each Attribute Set is stored on disk as an RNTuple and the Anchor of each RNTuple is linked to by the main
+RNTuple's footer.
+
+An Attribute Set RNTuple has a number of restrictions compared to a regular RNTuple:
+
+1. it cannot have linked Attribute RNTuples itself;
+2. the Alias columns sections, both in its header and footer, must be empty (i.e. none of the Attribute Set RNTuple's
+   Fields can be Projected Fields);
+3. none of its fields may have a structural role of 0x04 (i.e. it must not contain a ROOT streamer object);
```
hahnjo: For consideration, maybe move into a separate top-level section (between Envelopes and Mappings) explaining Attribute Sets rather than nested inside the Footer Envelope?

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -684,6 +685,64 @@ The entry range allows for finding the right page list for random access request
...
+
+The Attribute Schema Version describes the internal schema of the linked Attribute Set RNTuple.
+A change in Major version number indicates a breaking, non-forward-compatible change in the schema: readers should
+refuse reading an Attribute Set whose Major Schema Version is unknown.
+A change in Minor version number indicates the presence of optional additional fields in the schema: readers should
+still be able to read the Attribute Set as before, ignoring any new field.
+
+The current Attribute Schema Version is **1.0**. It has the following fields (in the following order):
+  1. `_rangeStart` (type `NTupleSize_t`): the start of the range that each Attribute Entry refers to;
+  2. `_rangeLen` (type `NTupleSize_t`): the length of the range that each Attribute Entry refers to.
```
hahnjo: `NTupleSize_t` is a `typedef` that I believe we are not treating specially during type normalization. In any case, it's not defined so far in the spec, I think it would be better to resolve this to `std::uint64_t`.

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -684,6 +685,64 @@ The entry range allows for finding the right page list for random access request
...
+##### Attribute Schema Version
+Each Attribute Set is created with a user-defined Model. This Model is not used directly by the underlying Attribute
+Set RNTuple, but it is augmented with internal fields used to store additional data that serve to associate each
+entry in the Attribute Set with those in the main RNTuple.
+
+The Attribute Schema Version describes the internal schema of the linked Attribute Set RNTuple.
+A change in Major version number indicates a breaking, non-forward-compatible change in the schema: readers should
+refuse reading an Attribute Set whose Major Schema Version is unknown.
+A change in Minor version number indicates the presence of optional additional fields in the schema: readers should
+still be able to read the Attribute Set as before, ignoring any new field.
```
hahnjo: This sub-sub-sub-section could then also move to the new top-level section.

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -684,6 +685,64 @@ The entry range allows for finding the right page list for random access request
...
+|             Attribute Anchor Uncompressed Size                |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                   Attribute Anchor Locator                    |
+.                             ...                               .
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                      Attribute Set Name                       |
+.                             ...                               .
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
hahnjo: For variable record frames, we otherwise use "followed by [...]" (the attribute anchor locator and the attribute set name)

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -1141,6 +1194,59 @@ public:
...
+class RNTupleAttrSetDescriptorIterable final {
+private:
+   /// The associated RNTuple for this range.
+   const RNTupleDescriptor &fNTuple;
+
+public:
+   class RIterator final {
+   private:
+      using Iter_t = std::vector<RNTupleAttrSetDescriptor>::const_iterator;
+      /// The wrapped map iterator
```
hahnjo: It's not a map...

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -1141,6 +1194,59 @@ public:
...
+      Iter_t fIter;
+
+   public:
+      using iterator_category = std::bidirectional_iterator_tag;
```
hahnjo: The tag is not correct, the iterator doesn't implement enough methods. I think we recently concluded that even for iterators wrapping `std::vector`, we only want to expose `std::forward_iterator_tag` (see `RExtraTypeInfoDescriptorIterable`)

hahnjo (commented):
some comments on the spec additions and the iterator implementations

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +800,58 @@ In every cluster, every field has exactly one primary column representation.
...
+An RNTuple may have zero or more linked Attribute Sets, containing metadata.
+Each Attribute Set is stored on disk as an RNTuple and the Anchor of each RNTuple is linked to by the main
+RNTuple's footer.
+
+An Attribute Set RNTuple has a number of restrictions compared to a regular RNTuple:
+
+1. it cannot have linked Attribute RNTuples itself;
+2. the Alias columns sections, both in its header and footer, must be empty (i.e. none of the Attribute Set RNTuple's
+   Fields can be Projected Fields);
+3. none of its fields may have a structural role of 0x04 (i.e. it must not contain a ROOT streamer object);
```
pcanal, line 829: Is it a choice or a technical limitation?

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +800,58 @@ In every cluster, every field has exactly one primary column representation.
...
+## Linked Attribute Sets
+
+An RNTuple may have zero or more linked Attribute Sets, containing metadata.
+Each Attribute Set is stored on disk as an RNTuple and the Anchor of each RNTuple is linked to by the main
+RNTuple's footer.
+
+An Attribute Set RNTuple has a number of restrictions compared to a regular RNTuple:
+
+1. it cannot have linked Attribute RNTuples itself;
+2. the Alias columns sections, both in its header and footer, must be empty (i.e. none of the Attribute Set RNTuple's
```
pcanal: Is there (somewhere else) a more detailed explanation/reason for this limitation?

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +800,58 @@ In every cluster, every field has exactly one primary column representation.
...
+- a 32-bit unsigned integer follows, containing the uncompressed size of the Anchor.
+
+These fields are followed by: 
+
+- a locator pointing to the Attribute RNTuple's anchor;
+- a string containing the Attribute Set's name. All linked Attribute Sets must have a non-empty, distinct name.
+
+### Attribute Schema Version
+Each Attribute Set is created with a user-defined Model. This Model is not used directly by the underlying Attribute
+Set RNTuple, but it is augmented with internal fields used to store additional data that serve to associate each
```
pcanal: When inspecting the `Attribute Set RNTuple` will the additional fields be exposed or hidden? (And If they are exposed would a user be able to distinguish the implicit vs the explicit part of the model?)

tree/ntuple/src/RNTupleSerialize.cxx
```diff
@@ -1918,36 +1966,102 @@ ROOT::RResult<void> ROOT::Internal::RNTupleSerializer::DeserializeFooter(const v
...
+   std::uint64_t frameSize;
+   if (auto res = DeserializeFrameHeader(bytes, fnBufSizeLeft(), frameSize)) {
+      bytes += res.Unwrap();
+   } else {
+      return R__FORWARD_ERROR(res);
+   }
+   if (fnBufSizeLeft() < static_cast<int>(sizeof(std::uint64_t)))
+      return R__FAIL("record frame too short");
+   std::uint16_t vMajor, vMinor;
+   bytes += DeserializeUInt16(bytes, vMajor);
```
pcanal, line 2043: Should there be a:
```
if (vMajor > 1)
	throw;
```
and/or did I misunderstand:
```
A change in Major version number indicates a breaking, non-forward-compatible change in the schema: readers should
refuse reading an Attribute Set whose Major Schema Version is unknown.
```
and/or should that statement be weakened?

tree/ntuple/doc/BinaryFormatSpecification.md
silverweed: It's a choice (see other answer)

tree/ntuple/doc/BinaryFormatSpecification.md
silverweed: We decided that we want to keep the Model of the attribute sets "as simple as possible" by removing certain advanced functionalities that we don't see as useful in the context of metadata. We may relax these restrictions in the future should the need arise.

tree/ntuple/doc/BinaryFormatSpecification.md
silverweed: They will be hidden: the user doesn't have access to the inner model but only to the fields they defined on the user model.

tree/ntuple/src/RNTupleSerialize.cxx
silverweed: Yes, we should add that check, thanks.

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +800,58 @@ In every cluster, every field has exactly one primary column representation.
 All other representations must be suppressed.
 Note that the primary column representation can change from cluster to cluster.
 
+## Linked Attribute Sets
+
+An RNTuple may have zero or more linked Attribute Sets, containing metadata.
```
enirolf:
```suggestion
An RNTuple may have zero or more linked Attribute Sets, containing user-defined metadata.
```

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +800,58 @@ In every cluster, every field has exactly one primary column representation.
...
+Each Attribute Set is stored on disk as an RNTuple and the Anchor of each RNTuple is linked to by the main
```
enirolf: "anchor" is not capitalized anywhere else:
```suggestion
Each Attribute Set is stored on disk as an RNTuple and the anchor of each RNTuple is linked to by the main
```

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +800,58 @@ In every cluster, every field has exactly one primary column representation.
...
+RNTuple's footer.
+
+An Attribute Set RNTuple has a number of restrictions compared to a regular RNTuple:
+
+1. it cannot have linked Attribute RNTuples itself;
+2. the Alias columns sections, both in its header and footer, must be empty (i.e. none of the Attribute Set RNTuple's
+   Fields can be Projected Fields);
```
enirolf:
```suggestion
2. the alias columns sections, both in its header and footer, must be empty (i.e. none of the Attribute Set RNTuple's
   fields can be projected fields);
```

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +800,58 @@ In every cluster, every field has exactly one primary column representation.
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|     Schema Version Major     |     Schema Version Minor       |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|             Attribute Anchor Uncompressed Size                |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+```
+
+- The first 32 bits contain the _Attribute Schema Version_. This is split into a _Major_ (16 LSB) and a
+  _Minor_ (16 MSB) version. The Schema Version is described below;
+- a 32-bit unsigned integer follows, containing the uncompressed size of the Anchor.
```
enirolf:
```suggestion
- a 32-bit unsigned integer follows, containing the uncompressed size of the Attribute Anchor.
```

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +800,58 @@ In every cluster, every field has exactly one primary column representation.
...
+
+These fields are followed by: 
+
+- a locator pointing to the Attribute RNTuple's anchor;
+- a string containing the Attribute Set's name. All linked Attribute Sets must have a non-empty, distinct name.
+
+### Attribute Schema Version
+Each Attribute Set is created with a user-defined Model. This Model is not used directly by the underlying Attribute
```
enirolf:
```suggestion
Each Attribute Set is created with a user-defined model. This model is not used directly by the underlying Attribute
Set RNTuple, but it is augmented with internal fields. These fields store additional data that associate each
entry in the Attribute Set with those in the main RNTuple.
```

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +800,58 @@ In every cluster, every field has exactly one primary column representation.
...
+A change in Major version number indicates a breaking, non-forward-compatible change in the schema: readers should
+refuse reading an Attribute Set whose Major Schema Version is unknown.
+A change in Minor version number indicates the presence of optional additional fields in the schema: readers should
+still be able to read the Attribute Set as before, ignoring any new field.
+
+The current Attribute Schema Version is **1.0**. It has the following fields (in the following order):
+  1. `_rangeStart` (type `std::uint64_t`): the start of the range that each Attribute Entry refers to;
+  2. `_rangeLen` (type `std::uint64_t`): the length of the range that each Attribute Entry refers to.
+     Note that `_rangeLen == 0` is valid and refers to an empty range;
+  3. `_userModel` (Untyped Record): a Record field that serves as the root field to the user-provided RNTupleModel
```
enirolf:
```suggestion
  3. `_userModel` (untyped record): a record-type field that serves as the root field to the user-provided RNTupleModel
     used by the Attribute Set RNTuple. Each user-defined field that was attached to the Field Zero in the user-provided
```

tree/ntuple/inc/ROOT/RNTupleSerialize.hxx
```diff
@@ -36,6 +36,13 @@ class RNTupleDescriptor;
 class RClusterDescriptor;
 enum class EExtraTypeInfoIds;
 
+namespace Experimental {
+class RNTupleAttrSetDescriptor;
+namespace Internal {
+class RNTupleAttrSetDescriptorBuilder;
+}
```
enirolf:
```suggestion
} // namespace Internal
```

tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -1359,6 +1374,19 @@ void ROOT::Internal::RNTupleDescriptorBuilder::ReplaceExtraTypeInfo(RExtraTypeIn
...
 }
 
+ROOT::RResult<void>
+ROOT::Internal::RNTupleDescriptorBuilder::AddAttributeSet(Experimental::RNTupleAttrSetDescriptor &&attrSetDesc)
+{
+   auto &attrSets = fDescriptor.fAttributeSets;
+   if (std::find_if(attrSets.begin(), attrSets.end(), [&name = attrSetDesc.GetName()](const auto &desc) {
+          return desc.GetName() == name;
+       }) != attrSets.end()) {
+      return R__FAIL("attribute sets with duplicate names");
```
enirolf, line 1402:
```suggestion
      return R__FAIL("AttributeSets with duplicate names");
```
(alternatively, change the CamelCase in line 1114)

tree/ntuple/doc/BinaryFormatSpecification.md
silverweed: I don't think this is necessarily true, as some metadata may be automatically defined (e.g. ROOT's internal attributes). It is anyway not relevant for the purposes of specification

tree/ntuple/src/RNTupleSerialize.cxx
silverweed: Actually I think I'll add it in a later PR when I define a constant with the current major version

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +800,58 @@ In every cluster, every field has exactly one primary column representation.
...
+```
+
+- The first 32 bits contain the _Attribute Schema Version_. This is split into a _Major_ (16 LSB) and a
+  _Minor_ (16 MSB) version. The Schema Version is described below;
+- a 32-bit unsigned integer follows, containing the uncompressed size of the Attribute Anchor.
+
+These fields are followed by: 
+
+- a locator pointing to the Attribute RNTuple's anchor;
+- a string containing the Attribute Set's name. All linked Attribute Sets must have a non-empty, distinct name.
```
hahnjo: Sorry for not being clear: I think this part must stay in the "Footer Envelope" section. The remaining restrictions and explanations should stay in this new section, thanks.

tree/ntuple/doc/BinaryFormatSpecification.md
enirolf: I think it is relevant to distinguish between the type of metadata the attributes store and the other kind of metadata (i.e. the ones in the header and footer) that is mentioned throughout this spec. It's true that "user-defined" may not be complete enough, how about something like "entry metadata", "metadata over (ranges of) entries" or "metadata related to the stored contents" instead?

silverweed: I see; I'll think of an appropriate phrasing.

silverweed: In the end I think "user-defined" is fine. @jblomer should the specs explicitly reserve the "ROOT" attribute namespace?

jblomer: Yes, I think that would be good.

silverweed: After an offline discussion we decided to reserve all names starting with `__` for implementation-defined metadata. Our implementation will use the `__ROOT` "namespace" but that is not part of the specification.

tree/ntuple/doc/BinaryFormatSpecification.md
pcanal: Do we say this somewhere in the text?  If a user "need" these advanced functionalities for their meta-data they need some input on whether they have no choice but to use something else or if they can talk to us first :)

silverweed: I don't think we need to spell it out in the specification; if a user needs these functionalities they will probably just ask us regardless (but I don't see a situation where one would _actually_ need these kind of functionalities for metadata: if they do they're most likely doing something wrong).

tree/ntuple/doc/BinaryFormatSpecification.md
hahnjo: Sorry for coming back to this, I still find the added document structure confusing. IMHO I was expecting something like the following:
```
## Envelopes
### Footer Envelope
[...]
#### Cluster Group Record Frame
#### Linked Attribute Set Record Frame
[...]
## Linked Attribute Sets
### Reserved Attribute Set Names
### Attribute Schema Version
## Mapping [...]
```

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +838,25 @@ In every cluster, every field has exactly one primary column representation.
...
+A change in Major version number indicates a breaking, non-forward-compatible change in the schema: readers should
+refuse reading an Attribute Set whose Major Schema Version is unknown.
+A change in Minor version number indicates the presence of optional additional fields in the schema: readers should
+still be able to read the Attribute Set as before, ignoring any new field.
+
+The current Attribute Schema Version is **1.0**. It has the following fields (in the following order):
+  1. `_rangeStart` (type `std::uint64_t`): the start of the range that each Attribute Entry refers to;
+  2. `_rangeLen` (type `std::uint64_t`): the length of the range that each Attribute Entry refers to.
+     Note that `_rangeLen == 0` is valid and refers to an empty range;
+  3. `_userModel` (untyped record): a record-type field that serves as the root field to the user-provided RNTupleModel
```
hahnjo: So far, the specification doesn't define the concept of a model, and `RNTupleModel` is an implementation detail in ROOT. I think the specification calls it "schema description" (section on "Header Envelope")

tree/ntuple/doc/BinaryFormatSpecification.md
silverweed: Hopefully I got it right this time!

tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx
```diff
@@ -67,6 +67,53 @@ struct RNTupleClusterBoundaries {
 std::vector<ROOT::Internal::RNTupleClusterBoundaries> GetClusterBoundaries(const RNTupleDescriptor &desc);
 } // namespace Internal
 
+namespace Experimental {
+namespace Internal {
+class RNTupleAttrSetDescriptorBuilder;
+}
```
jblomer: I dont think we need the forward declaration.

tree/ntuple/doc/BinaryFormatSpecification.md
```diff
@@ -799,6 +815,47 @@ In every cluster, every field has exactly one primary column representation.
...
+1. it cannot have linked attribute RNTuples itself;
+2. the alias columns sections, both in its header and footer, must be empty (i.e. none of the attribute set RNTuple's
+   fields can be projected fields);
+3. none of its fields may have a structural role of 0x04 (i.e. it must not contain a ROOT streamer object);
+
+All linked attribute sets must have a non-empty, distinct name.
+
+### Reserved Attribute Set Names
+
+in the "Linked Attribute Sets" paragraphAttribute set names starting with `__` (two underscores) are reserved for internal use by implementations: compliant
```
hahnjo: It looks like there is some leftover here?
```suggestion
Attribute set names starting with `__` (two underscores) are reserved for internal use by implementations: compliant
```

jblomer (commented):
I think we need one more change with this PR: the RNTuple anchor needs to update the schema version to match the new specification.

tree/ntuple/doc/BinaryFormatSpecification.md
silverweed: Oh, I must have accidentally pasted it, thanks for spotting it!

silverweed:
>  It's still mentioning "model" in some places (notably in the _userModel field name), not sure if we want to fully remove it...

I'll remove it from the leftover places; I guess at this point I should also rename the field to `_userSchema`?

hahnjo:
> > It's still mentioning "model" in some places (notably in the _userModel field name), not sure if we want to fully remove it...
> 
> I'll remove it from the leftover places; I guess at this point I should also rename the field to `_userSchema`?

Or actually `_userData` since that's what the record field contains?

tree/ntuple/test/ntuple_minifile.cxx
```diff
@@ -76,10 +76,10 @@ TEST(MiniFile, Stream)
...
    RMiniFileReader reader(rawFile.get());
    auto ntuple = reader.GetNTuple("MyNTuple").Inspect();
-   EXPECT_EQ(1u, ntuple.GetVersionEpoch());
-   EXPECT_EQ(0u, ntuple.GetVersionMajor());
-   EXPECT_EQ(0u, ntuple.GetVersionMinor());
-   EXPECT_EQ(2u, ntuple.GetVersionPatch());
+   EXPECT_EQ(ROOT::RNTuple::kVersionEpoch, ntuple.GetVersionEpoch());
+   EXPECT_EQ(ROOT::RNTuple::kVersionMajor, ntuple.GetVersionMajor());
+   EXPECT_EQ(ROOT::RNTuple::kVersionMinor, ntuple.GetVersionMinor());
+   EXPECT_EQ(ROOT::RNTuple::kVersionPatch, ntuple.GetVersionPatch());
```
hahnjo: IIRC the point of these tests was precisely that we were forced to update them and think about changing the version number...

PR #19899, opened 2025-09-16
[hist] Implement `RHistStats`

hahnjo:
Histogram statistics of unbinned values to compute the number of effective entries as well as the arithmetic mean and other statistical quantities per dimension.

### DIFF
--- a/hist/histv7/inc/ROOT/RHistStats.hxx
+++ b/hist/histv7/inc/ROOT/RHistStats.hxx
@@ -192,6 +192,9 @@ public:
       }
       double mean = ComputeMean(dim);
       double var = ComputeVariance(dim);
+      if (var == 0) {
+         return 0;
+      }
       double EWX3 = stats.fSumWX3 / fSumW;
       double EWX2 = stats.fSumWX2 / fSumW;
       return (EWX3 - 3 * EWX2 * mean + 2 * mean * mean * mean) / std::pow(var, 1.5);
@@ -226,6 +229,9 @@ public:
       }
       double mean = ComputeMean(dim);
       double var = ComputeVariance(dim);
+      if (var == 0) {
+         return 0;
+      }
       double EWX4 = stats.fSumWX4 / fSumW;
       double EWX3 = stats.fSumWX3 / fSumW;
       double EWX2 = stats.fSumWX2 / fSumW;
--- a/hist/histv7/inc/ROOT/RHistStats.hxx
+++ b/hist/histv7/inc/ROOT/RHistStats.hxx
@@ -158,6 +158,70 @@ public:
    /// \return the standard deviation of unbinned values
    double ComputeStdDev(std::size_t dim = 0) const { return std::sqrt(ComputeVariance(dim)); }
 
+   // clang-format off
+   /// Compute the skewness of unbinned values.
+   ///
+   /// The skewness is the third standardized moment:
+   /// \f[
+   /// E\left[\left(\frac{X - \mu}{\sigma}\right)^3\right]
+   /// \f]
+   /// With support for weighted filling and after some rewriting, it is computed as:
+   /// \f[
+   /// \frac{\frac{\sum w_i \cdot x_i^3}{\sum w_i} - 3 \cdot \frac{\sum w_i \cdot x_i^2}{\sum w_i} \cdot \mu + 2 \cdot \mu^3}{\sigma^3}
+   /// \f]
+   ///
+   /// \param[in] dim the dimension index, starting at 0
+   /// \return the skewness of unbinned values
+   // clang-format on
+   double ComputeSkewness(std::size_t dim = 0) const
+   {
+      // First get the statistics, which includes checking the argument.
+      auto &stats = fDimensionStats.at(dim);
+      if (fSumW == 0) {
+         return 0;
+      }
+      double mean = ComputeMean(dim);
+      double var = ComputeVariance(dim);
+      double EWX3 = stats.fSumWX3 / fSumW;
+      double EWX2 = stats.fSumWX2 / fSumW;
+      return (EWX3 - 3 * EWX2 * mean + 2 * mean * mean * mean) / std::pow(var, 1.5);
+   }
+
+   // clang-format off
+   /// Compute the (excess) kurtosis of unbinned values.
+   ///
+   /// The kurtosis is based on the fourth standardized moment:
+   /// \f[
+   /// E\left[\left(\frac{X - \mu}{\sigma}\right)^4\right]
+   /// \f]
+   /// The excess kurtosis subtracts 3 from the standardized moment to have a value of 0 for a normal distribution:
+   /// \f[
+   /// E\left[\left(\frac{X - \mu}{\sigma}\right)^4\right] - 3
+   /// \f]
+   ///
+   /// With support for weighted filling and after some rewriting, the (excess kurtosis) is computed as:
+   /// \f[
+   /// \frac{\frac{\sum w_i \cdot x_i^4}{\sum w_i} - 4 \cdot \frac{\sum w_i \cdot x_i^3}{\sum w_i} \cdot \mu + 6 \cdot \frac{\sum w_i \cdot x_i^2}{\sum w_i} \cdot \mu^2 - 3 \cdot \mu^4}{\sigma^4} - 3
+   /// \f]
+   ///
+   /// \param[in] dim the dimension index, starting at 0
+   /// \return the (excess) kurtosis of unbinned values
+   // clang-format on
+   double ComputeKurtosis(std::size_t dim = 0) const
+   {
+      // First get the statistics, which includes checking the argument.
+      auto &stats = fDimensionStats.at(dim);
+      if (fSumW == 0) {
+         return 0;
+      }
+      double mean = ComputeMean(dim);
+      double var = ComputeVariance(dim);
+      double EWX4 = stats.fSumWX4 / fSumW;
+      double EWX3 = stats.fSumWX3 / fSumW;
+      double EWX2 = stats.fSumWX2 / fSumW;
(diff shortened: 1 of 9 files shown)


hahnjo:
Ping, this is the last PR needed before I can put together the "final" `RHist = RHistEngine + RHistStats`

hist/histv7/inc/ROOT/RHistStats.hxx
```diff
@@ -0,0 +1,351 @@
...
+// stats.ComputeMean() will return 5.0
+\endcode
+
+\warning This is part of the %ROOT 7 prototype! It will change without notice. It might trigger earthquakes.
+Feedback is welcome!
+*/
+class RHistStats final {
+public:
+   /// Statistics for one dimension.
+   struct DimensionStats final {
```
jblomer:
```suggestion
   struct RDimensionStats final {
```

hist/histv7/inc/ROOT/RHistStats.hxx
```diff
@@ -0,0 +1,351 @@
...
+
+      void Add(double x)
+      {
+         fSumWX += x;
+         fSumWX2 += x * x;
+         fSumWX3 += x * x * x;
+         fSumWX4 += x * x * x * x;
+      }
+
+      void Add(double x, double w)
```
jblomer, line 55: Should we use RWeight here, too?

hist/histv7/inc/ROOT/RHistStats.hxx
```diff
@@ -0,0 +1,351 @@
...
+   /// \f[
+   /// \sigma^2 = \frac{1}{\sum w_i} \sum(w_i \cdot x_i - \mu)^2
+   /// \f]
+   /// With some rewriting, this is equivalent to:
+   /// \f[
+   /// \sigma^2 = \frac{\sum w_i \cdot x_i^2}{\sum w_i} - \frac{(\sum w_i \cdot x_i)^2}{(\sum w_i)^2}
+   /// \f]
+   ///
+   /// This function does not include Bessel's correction needed for an unbiased estimator of population variance.
+   /// In other words, the return value is a biased estimation lower than the actual population variance.
```
jblomer, line 139: Do we want this?

hist/histv7/inc/ROOT/RHistStats.hxx
```diff
@@ -0,0 +1,351 @@
...
+   ///
+   /// \f[
+   /// \mu = \frac{\sum w_i \cdot x_i}{\sum w_i}
+   /// \f]
+   ///
+   /// \param[in] dim the dimension index, starting at 0
+   /// \return the arithmetic mean of unbinned values
+   double ComputeMean(std::size_t dim = 0) const
+   {
+      // First get the statistics, which includes checking the argument.
```
jblomer, line 119: Here and further down: not sure if we need the argument.

hist/histv7/inc/ROOT/RHistStats.hxx
```diff
@@ -0,0 +1,351 @@
...
+      auto &stats = fDimensionStats.at(dim);
+      if (fSumW == 0) {
+         return 0;
+      }
+      double mean = ComputeMean(dim);
+      double var = ComputeVariance(dim);
+      double EWX3 = stats.fSumWX3 / fSumW;
+      double EWX2 = stats.fSumWX2 / fSumW;
+      return (EWX3 - 3 * EWX2 * mean + 2 * mean * mean * mean) / std::pow(var, 1.5);
```
jblomer, line 201: Can't `var` be zero? E.g. after filling 1?

hist/histv7/inc/ROOT/RHistStats.hxx
```diff
@@ -0,0 +1,351 @@
...
+   /// Compute the arithmetic mean of unbinned values.
+   ///
+   /// \f[
+   /// \mu = \frac{\sum w_i \cdot x_i}{\sum w_i}
+   /// \f]
+   ///
+   /// \param[in] dim the dimension index, starting at 0
+   /// \return the arithmetic mean of unbinned values
+   double ComputeMean(std::size_t dim = 0) const
+   {
```
ferdymercury, line 118: Is it planned to have Kahan sum correction, as with RDF.Mean?

hist/histv7/inc/ROOT/RHistStats.hxx
hahnjo: Not sure, so far we use `RWeight` only in the user-facing interface functions and then "unpack" it before adding to the bin content or statistics.

hist/histv7/inc/ROOT/RHistStats.hxx
hahnjo: The problem is I'm not sure there is a general way to apply the correction with weighted filling. Usually you use `N - 1` instead of `N`, but just subtracting 1 from `fSumW` doesn't make sense. So instead for the moment, I'm more clearly documenting what has been implemented for the past 30+ years...

hist/histv7/inc/ROOT/RHistStats.hxx
hahnjo: I'm not sure what you mean: `dim` is needed to get the statistics for a particular dimension.

hist/histv7/inc/ROOT/RHistStats.hxx
hahnjo: Currently not planned, unless requested for a clear use case.

hist/histv7/inc/ROOT/RHistStats.hxx
hahnjo: Yes, you are absolutely right! I thought about `fSumW == 0` in some places, but missed `var == 0`. The protection is still not perfect because it compares exact equality, but at least users won't get NaN. Fun fact: `TH1` doesn't have the protection either, it returns NaN...

hist/histv7/inc/ROOT/RHistStats.hxx
```diff
@@ -0,0 +1,222 @@
...
+      fDimensionStats.resize(nDimensions);
+   }
+
+   std::size_t GetNDimensions() const { return fDimensionStats.size(); }
+
+   std::uint64_t GetNEntries() const { return fNEntries; }
+   double GetSumW() const { return fSumW; }
+   double GetSumW2() const { return fSumW2; }
+
+   const RDimensionStats &GetDimensionStats(std::size_t dim = 0) const { return fDimensionStats.at(dim); }
```
hageboeck, line 92: Is the default of 0 such a good idea? Wouldn't it be better to force the caller to think about which dimension they need?

hist/histv7/inc/ROOT/RHistStats.hxx
```diff
@@ -0,0 +1,222 @@
...
+ROOT::Experimental::RHistStats stats(1);
+stats.Fill(8.5);
+stats.Fill(1.5);
+// stats.ComputeMean() will return 5.0
+\endcode
+
+\warning This is part of the %ROOT 7 prototype! It will change without notice. It might trigger earthquakes.
+Feedback is welcome!
+*/
+class RHistStats final {
```
hageboeck, line 38: Is something like a `clear()` planned? Or is this part of RHist, which could instantiate an empty RHistStats?

hist/histv7/inc/ROOT/RHistStats.hxx
```diff
@@ -0,0 +1,222 @@
...
+public:
+   /// Statistics for one dimension.
+   struct RDimensionStats final {
+      double fSumWX = 0.0;
+      double fSumWX2 = 0.0;
+      double fSumWX3 = 0.0;
+      double fSumWX4 = 0.0;
```
hageboeck, line 45: `X` might be a bit misleading, because this could be used for any dimension. Doesn't affect the results, though ...

hist/histv7/test/hist_stats.cxx
```diff
@@ -0,0 +1,136 @@
...
+
+   static constexpr std::size_t Entries = 20;
+   for (std::size_t i = 0; i < Entries; i++) {
+      stats.Fill(1);
+   }
+
+   ASSERT_EQ(stats.GetNEntries(), Entries);
+   EXPECT_DOUBLE_EQ(stats.GetSumW(), Entries);
+   EXPECT_DOUBLE_EQ(stats.GetSumW2(), Entries);
+   EXPECT_DOUBLE_EQ(stats.ComputeNEffectiveEntries(), Entries);
```
hageboeck, line 99: ~~Although this achieves coverage, I'm not fully convinced that testing with a weight of 1 is the most interesting test for the effective entries.~~

Got it. I went commit-by-commit, so the weighted filling wasn't visible yet.

hist/histv7/inc/ROOT/RHistStats.hxx
hahnjo: The idea was that for one-dimensional histograms, I didn't want to bother the user to write out that he meant the only dimension. I could try to restrict the default to that case, but I'm not sure it's worth the effort...

hist/histv7/inc/ROOT/RHistStats.hxx
hahnjo: Yes, implemented in https://github.com/root-project/root/pull/19959/commits/dd275dfda2fe3166421ed63730f8d4ef9b724879 exactly for usage in `RHist`

hist/histv7/inc/ROOT/RHistStats.hxx
hahnjo: It can be implemented, but it requires additional overloads that then only work for some object configurations. I think I would let it this way for the moment, and we can reconsider if the default argument is confusing to users.

hist/histv7/inc/ROOT/RHistStats.hxx
hahnjo: Discussed offline, it's about whether we would need the comment. We leave it as-is for the moment, it also doesn't hurt.

PR #19904, opened 2025-09-16
[ntuple] Add some internal utilities to the storage layer

silverweed:
**NOTE**: this PR has been split into multiple ones:
- #21007
- #21010
- #21063

Add a bunch of functions to poke through the abstraction of the storage classes (specifically, to get the underlying TDirectory and RMiniFileReader).  These functionalities (all Internal) will be needed by the RNTuple Attributes (see [the full PR](https://github.com/root-project/root/pull/19153) for details)

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/inc/ROOT/RMiniFile.hxx
+++ b/tree/ntuple/inc/ROOT/RMiniFile.hxx
@@ -32,13 +32,14 @@ class TVirtualStreamerInfo;
 
 namespace ROOT {
 
-namespace Internal {
-class RRawFile;
-}
-
 class RNTupleWriteOptions;
 
 namespace Internal {
+
+class RRawFile;
+
+TDirectory *GetUnderlyingDirectory(ROOT::Internal::RNTupleFileWriter &writer);
+
 /// Holds status information of an open ROOT file during writing
 struct RTFileControlBlock;
 
@@ -68,9 +69,6 @@ private:
    /// Used when the file turns out to be a TFile container. The ntuplePath variable is either the ntuple name
    /// or an ntuple name preceded by a directory (`myNtuple` or `foo/bar/myNtuple` or `/foo/bar/myNtuple`)
    RResult<RNTuple> GetNTupleProper(std::string_view ntuplePath);
-   /// Loads an RNTuple anchor from a TFile at the given file offset (unzipping it if necessary).
-   RResult<RNTuple>
-   GetNTupleProperAtOffset(std::uint64_t payloadOffset, std::uint64_t compSize, std::uint64_t uncompLen);
 
    /// Searches for a key with the given name and type in the key index of the directory starting at offsetDir.
    /// The offset points to the start of the TDirectory DATA section, without the key and without the name and title
@@ -84,6 +82,9 @@ public:
    explicit RMiniFileReader(ROOT::Internal::RRawFile *rawFile);
    /// Extracts header and footer location for the RNTuple identified by ntupleName
    RResult<RNTuple> GetNTuple(std::string_view ntupleName);
+   /// Loads an RNTuple anchor from a TFile at the given file offset (unzipping it if necessary).
+   RResult<RNTuple>
+   GetNTupleProperAtOffset(std::uint64_t payloadOffset, std::uint64_t compSize, std::uint64_t uncompLen);
    /// Reads a given byte range from the file into the provided memory buffer.
    /// If `nbytes > fMaxKeySize` it will perform chunked read from multiple blobs,
    /// whose addresses are listed at the end of the first chunk.
@@ -109,6 +110,8 @@ A stand-alone version of RNTuple can remove the TFile based writer.
 */
 // clang-format on
 class RNTupleFileWriter {
+   friend TDirectory *ROOT::Internal::GetUnderlyingDirectory(ROOT::Internal::RNTupleFileWriter &writer);
+
 public:
    /// The key length of a blob. It is always a big key (version > 1000) with class name RBlob.
    static constexpr std::size_t kBlobKeyLen = 42;
@@ -254,7 +257,7 @@ public:
    void WriteIntoReservedBlob(const void *buffer, size_t nbytes, std::int64_t offset);
    /// Ensures that the streamer info records passed as argument are written to the file
    void UpdateStreamerInfos(const ROOT::Internal::RNTupleSerializer::StreamerInfoMap_t &streamerInfos);
-   /// Writes the RNTuple key to the file so that the header and footer keys can be found
+   /// Writes the RNTuple key to the file so that the header and footer keys can be found.
    void Commit(int compression = RCompressionSetting::EDefaults::kUseGeneralPurpose);
 };
 
--- a/tree/ntuple/inc/ROOT/RPageSinkBuf.hxx
+++ b/tree/ntuple/inc/ROOT/RPageSinkBuf.hxx
@@ -152,6 +152,11 @@ public:
    void CommitDatasetImpl() final;
 
    RPage ReservePage(ColumnHandle_t columnHandle, std::size_t nElements) final;
+
+   std::unique_ptr<RPageSink> DeriveFor(std::string_view newName, const ROOT::RNTupleWriteOptions &opts) const final
+   {
+      return std::make_unique<RPageSinkBuf>(fInnerSink->DeriveFor(newName, opts));
+   }
 }; // RPageSinkBuf
 
 } // namespace Internal
--- a/tree/ntuple/inc/ROOT/RPageStorage.hxx
+++ b/tree/ntuple/inc/ROOT/RPageStorage.hxx
@@ -47,9 +47,9 @@ namespace ROOT {
 class RNTupleModel;
 
 namespace Internal {
-
-class RPageAllocator;
 class RColumn;
+class RMiniFileReader;
+class RPageAllocator;
 struct RNTupleModelChangeset;
 
 enum class EPageStorageType {
@@ -312,6 +312,10 @@ public:
    virtual const ROOT::RNTupleDescriptor &GetDescriptor() const = 0;
 
    virtual ROOT::NTupleSize_t GetNEntries() const = 0;
+
+   /// Creates a new RPageSink linked to the same underlying storage as this, writing to a new RNTuple called `newName`.
+   /// The existing sink will stay valid. The existing sink and the new one must not write concurrently.
+   virtual std::unique_ptr<RPageSink> DeriveFor(std::string_view newName, const RNTupleWriteOptions &opts) const = 0;
 
    /// Physically creates the storage container to hold the ntuple (e.g., a keys a TFile or an S3 bucket)
    /// Init() associates column handles to the columns referenced by the model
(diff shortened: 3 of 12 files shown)


silverweed:
> Do we really need `GetUnderlyingDirectory()`? Since the attribute anchor is a hidden key, we can just store it in the ROOT file's root directory.

Not sure...the reason I need it in the full PR is to [create the AttrSetWriter](https://github.com/root-project/root/pull/19153/files#diff-e37e5de3bf6e5ff193af058ca15a5ccab9215ce08e9f4a4f1a7b82e9351806e6R115) which then uses it [create a PageSinkFile](https://github.com/root-project/root/pull/19153/files#diff-9724e97c8fbe3b9752ab174a04652de144b2d2c9469caa7f0b9f791d3aaee06eR101) using it. Even if we wanted to store the key in the root directory we'd still need the underlying TFile for it...

tree/ntuple/inc/ROOT/RNTupleFillContext.hxx
```diff
@@ -103,7 +103,7 @@ private:
    std::size_t FillImpl(Entry &entry)
    {
       ROOT::RNTupleFillStatus status;
-      FillNoFlush(entry, status);
+      FillNoFlushImpl(entry, status);
```
hahnjo: unrelated to this PR...

tree/ntuple/inc/ROOT/RPageSinkBuf.hxx
```diff
@@ -148,6 +148,11 @@ public:
    void CommitDatasetImpl() final;
 
    RPage ReservePage(ColumnHandle_t columnHandle, std::size_t nElements) final;
+
+   std::unique_ptr<RPageSink> CloneWithNewRNTuple(std::string_view newName) const final
+   {
+      return fInnerSink->CloneWithNewRNTuple(newName);
+   }
```
hahnjo: I feel the semantics are not well defined: As it stands, it will return a page sink that is not buffered anymore.

tree/ntuple/inc/ROOT/RPageStorageDaos.hxx
```diff
@@ -77,7 +77,8 @@ struct RDaosNTupleAnchor {
    /// The object class for user data OIDs, e.g. `SX`
    std::string fObjClass{};
 
-   bool operator ==(const RDaosNTupleAnchor &other) const {
+   bool operator==(const RDaosNTupleAnchor &other) const
+   {
```
hahnjo, line 81: unrelated to this PR

tree/ntuple/inc/ROOT/RNTupleFillContext.hxx
```diff
@@ -114,6 +114,9 @@ private:
    RNTupleFillContext &operator=(const RNTupleFillContext &) = delete;
 
 public:
+   RNTupleFillContext(RNTupleFillContext &&) = default;
+   RNTupleFillContext &operator=(RNTupleFillContext &&) = default;
+
```
hahnjo: unrelated to this PR (I believe, unless it's needed somewhere?)

tree/ntuple/src/RPageStorage.cxx
```diff
@@ -1198,8 +1199,8 @@ void ROOT::Internal::RPagePersistentSink::CommitStagedClusters(std::span<RStaged
          if (!columnInfo.fIsSuppressed)
             continue;
          const auto colId = columnInfo.fPageRange.GetPhysicalColumnId();
-         // For suppressed columns, we need to reset the first element index to the first element of the next (upcoming)
-         // cluster. This information has been determined for the committed cluster descriptor through
+         // For suppressed columns, we need to reset the first element index to the first element of the next
+         // (upcoming) cluster. This information has been determined for the committed cluster descriptor through
```
hahnjo: also unrelated to this PR

tree/ntuple/src/RPageStorageFile.cxx
```diff
@@ -305,6 +317,15 @@ ROOT::Internal::RPageSourceFile::CreateFromAnchor(const RNTuple &anchor, const R
    return pageSource;
 }
 
+std::unique_ptr<ROOT::Internal::RPageSourceFile>
+ROOT::Internal::RPageSourceFile::OpenWithDifferentAnchor(const RNTuple &anchor, const ROOT::RNTupleReadOptions &options)
+{
+   auto pageSource = std::make_unique<RPageSourceFile>("", fFile->Clone(), options);
+   pageSource->fAnchor = anchor;
+   pageSource->fNTupleName = pageSource->fDescriptorBuilder.GetDescriptor().GetName();
```
hahnjo: How is this initialized? Above you are passing `""` for `ntupleName`, is the header already loaded somewhere?

tree/ntuple/inc/ROOT/RMiniFile.hxx
```diff
@@ -84,6 +85,9 @@ public:
    explicit RMiniFileReader(ROOT::Internal::RRawFile *rawFile);
    /// Extracts header and footer location for the RNTuple identified by ntupleName
    RResult<RNTuple> GetNTuple(std::string_view ntupleName);
+   /// Loads an RNTuple anchor from a TFile at the given file offset (unzipping it if necessary).
+   RResult<RNTuple>
+   GetNTupleProperAtOffset(std::uint64_t payloadOffset, std::uint64_t compSize, std::uint64_t uncompLen);
```
hahnjo, line 87: Why do we need to be`friend` `RPageSource` *and* make this method `public`?

hahnjo (commented):
Some comments inline. I'm not actually sure it helps to review internal methods and additions to the `RPageStorage` abstraction layer without seeing the immediate need. Maybe one has to look at the final PR all the time, but that is quite some mental overhead. It feels like we are now vertically building the implementation, instead of in logical chunks...

tree/ntuple/inc/ROOT/RPageSinkBuf.hxx
silverweed: I'm not entirely sure what should happen here: it's not necessarily true that if the parent sink is buffered then we also want to buffer the clone. Arguably, returning a non-buffered sink gives the most flexibility as the caller can decide whether to wrap it in a buffered sink or not.

tree/ntuple/src/RPageStorageFile.cxx
silverweed: By "initialized" you mean attached? At the moment it needs to be done manually, like when you create a new PageSource. This is done in 2 places in the [full PR](https://github.com/root-project/root/pull/19153): in `ReadAttributeSet`, by the created Reader, and in `RNTupleMerger::OpenAttributeSource`

tree/ntuple/inc/ROOT/RPageSinkBuf.hxx
hahnjo: I would argue it's not a clone in this case... Maybe it's a matter of finding a more suitable name?

tree/ntuple/src/RPageStorageFile.cxx
hahnjo: Somewhat: Unless I'm missing something, I think the effect of this line is always that `pageSource->fNTupleName = "";` because the descriptor builder doesn't have more information (yet).

silverweed: I see. It's indeed possible that that line is useless (I still need to verify), but isn't this the case also for `CreateFromAnchor` then?

tree/ntuple/inc/ROOT/RPageSinkBuf.hxx
silverweed: I can't come up with anything better than `CreateNewWithNewRNTuple`...being a private internal method it might be ok even though it's ugly.

tree/ntuple/src/RPageStorageFile.cxx
hahnjo: That is very much possible...

silverweed: https://github.com/root-project/root/pull/20017 (if that solution is fine I will also change this PR accordingly)

silverweed:
> I'm not actually sure it helps to review internal methods and additions to the RPageStorage abstraction layer without seeing the immediate need. Maybe one has to look at the final PR all the time, but that is quite some mental overhead

I split it like this because this way it could be reviewed in parallel with the other PR rather than being blocked by it, but if you prefer I can also close this, wait for the other to be merged and then put this change together with the parts that use it (probably doing one PR for attribute reading, one for attribute writing and one for merging)

tree/ntuple/inc/ROOT/RPageStorage.hxx
```diff
@@ -313,6 +313,10 @@ public:
 
    virtual ROOT::NTupleSize_t GetNEntries() const = 0;
 
+   /// Creates a new RPageSink linked to the same underlying storage as this, writing to a new RNTuple called `newName`.
+   /// The existing sink will stay valid. The existing sink and the new one must not write concurrently.
+   virtual std::unique_ptr<RPageSink> CreateNewWithNewRNTuple(std::string_view newName) const = 0;
```
jblomer: How about `CreateFor` or `DeriveFor` or `[Create|Derive]With(Name)`

tree/ntuple/src/RPageStorageFile.cxx
```diff
@@ -252,6 +252,18 @@ void ROOT::Internal::RPageSinkFile::CommitDatasetImpl(unsigned char *serializedF
    fWriter->Commit(GetWriteOptions().GetCompression());
 }
 
+std::unique_ptr<ROOT::Internal::RPageSink>
+ROOT::Internal::RPageSinkFile::CreateNewWithNewRNTuple(std::string_view newName) const
+{
+   if (auto *dir = Internal::GetUnderlyingDirectory(*fWriter)) {
+      auto opts = ROOT::RNTupleWriteOptions();
```
jblomer:
```suggestion
      ROOT::RNTupleWriteOptions opts;
      opts.SetCompression(GetWriteOptions().GetCompression());
```

jblomer: I guess we also want to use the same setting for implicit multi-threading and page checksums. Possibly other options, too. Does it make sense to copy all the options and reset selected ones to their default?

tree/ntuple/inc/ROOT/RPageStorage.hxx
```diff
@@ -808,6 +812,8 @@ public:
    virtual std::vector<std::unique_ptr<ROOT::Internal::RCluster>>
    LoadClusters(std::span<ROOT::Internal::RCluster::RKey> clusterKeys) = 0;
 
+   virtual RMiniFileReader *GetUnderlyingReader() { return nullptr; }
```
jblomer, line 815: Do we need that? Aren't we using the reader only within `ReadAttrSet`?

tree/ntuple/src/RPageStorageFile.cxx
silverweed: I think we should go through each option and see if it makes sense to copy or not for the attributes RNTuple...the particularly delicate ones are those that have memory usage implications (we don't want the user to inadvertently double their memory usage if they create an attribute set). 
Those you mentioned surely make sense to me. Same page merging as well. Not sure about the direct IO...

silverweed: That said, since this specific place in the code has no idea of the context, it probably makes more sense that this function accepts WriteOptions from outside and the caller decides what to pass to it.

tree/ntuple/inc/ROOT/RPageStorage.hxx
silverweed: I think I like `DeriveFor` the best

tree/ntuple/inc/ROOT/RPageStorage.hxx
silverweed: It is also used by the RNTupleMerger, since we need to fetch the Anchor when opening the attribute sets during merging (see https://github.com/root-project/root/pull/19153/files#diff-3fe10805b28dcc612504e58728446c32a3f70f4e42b5a6d62e46d07facd46766R127)

PR #19907, opened 2025-09-16
[ntuple] Automatic schema evolution from enum to int

jblomer:


tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -169,6 +169,23 @@ bool ROOT::RFieldDescriptor::IsCustomClass() const
...
+   if (fStructure != ROOT::ENTupleStructure::kPlain)
+      return false;
+   if (fTypeName.rfind("std::", 0) == 0)
+      return false;
+
+   auto subFieldId = desc.FindFieldId("_0", fFieldId);
+   if (subFieldId == kInvalidDescriptorId)
+      return false;
+
+   static const std::unordered_set<std::string> gIntTypeNames = {"char",          "std::int8_t",   "std::uint8_t",
```
silverweed: Can't `bool` also be a valid option here?

jblomer: I don't think so, bool can't be a backing type for an enum.

silverweed: Are you sure? It seems supported by both [clang](https://godbolt.org/z/vhe3WherG) and [GCC](https://godbolt.org/z/evaaPEneq)...

ferdymercury: Yep, enum works with integer types, and bool is an (extended) integer type (0 and 1), much like char.
Related: https://stackoverflow.com/a/68321667/7471760

jblomer: Interesting! Always something new to learn.

Ok, that also means that we need to enable writing of such enums. Currently, the `TEnumField` constructor doesn't allow it.

jblomer: See #19924

jblomer: and fixed here, too.

tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -169,6 +169,23 @@ bool ROOT::RFieldDescriptor::IsCustomClass() const
...
+      return false;
+
+   auto subFieldId = desc.FindFieldId("_0", fFieldId);
+   if (subFieldId == kInvalidDescriptorId)
+      return false;
+
+   static const std::unordered_set<std::string> gIntTypeNames = {
+      "bool",          "char",         "std::int8_t",   "std::uint8_t", "std::int16_t",
+      "std::uint16_t", "std::int32_t", "std::uint32_t", "std::int64_t", "std::uint64_t"};
+   return gIntTypeNames.count(desc.GetFieldDescriptor(subFieldId).GetTypeName()) > 0;
```
silverweed: Given the very low number of elements I'm pretty sure a plain array search would be faster than a set lookup...not sure if important here though.

tree/ntuple/src/RField.cxx
```diff
@@ -92,6 +92,40 @@ const ROOT::RField<ROOT::RNTupleCardinality<std::uint64_t>> *ROOT::RCardinalityF
...
+
+   if (fieldDesc.IsCustomEnum(desc)) {
+      SetOnDiskId(desc.FindFieldId("_0", GetOnDiskId()));
+      return;
+   }
+
+   static const std::unordered_set<std::string> gIntegralTypeNames = {
+      "bool",          "char",         "std::int8_t",   "std::uint8_t", "std::int16_t",
+      "std::uint16_t", "std::int32_t", "std::uint32_t", "std::int64_t", "std::uint64_t"};
+   if (gIntegralTypeNames.count(fieldDesc.GetTypeName()) == 0) {
```
silverweed: See other comment about the lookup

tree/ntuple/src/RNTupleDescriptor.cxx
jblomer: True, it probably makes both compile time and runtime a bit faster.

hahnjo (approved):
LGTM

PR #19916, opened 2025-09-18
[nfc] fix typo in tree draw iteration docu

ferdymercury:


silverweed:
I'm not sure if this makes it clearer - neither description says whether the range is inclusive or exclusive so they both may be both correct and incorrect depending who you ask.

Issue #9530, opened 2022-01-11
TH1D::GetRandom and histograms with non-uniform binning


andrea-celentano:
Assume to have a TH1D object, with non-uniform binning, representing some kind of differential quantity (for example, a particle flux as a function of the particle energy, dN/dE). The `TH1D::GetRandom` method will return random numbers distributed according to the histogram bins height, rather than the histogram bins area.

The code attached here (a ROOT macro) reproduces this behavior (I attach it as a .txt file, since .C files are not supported)
[test.txt](https://github.com/root-project/root/files/7845002/test.txt)

couet:
It seems think that's not graphics and that more for @lmoneta

andrea-celentano:
Hi,
I am trying to follow the instructions reported [here](https://github.com/root-project/root/blob/c2f028feb6da152ff07020b3b4b2a961c1af315e/CONTRIBUTING.mdl) to fork the ROOT repository, create a new branch, suggest a fix, and then do a pull request. I agree that this is not a graphic issue, rather a numerical one.

andrea-celentano:
I made a pull request for this (9538), but I do not know how to link that to this issue (there is a "Linked pull requests" list on the web page, but I cannot modify it).

lmoneta:
Hi, 

 I think this is not a bug, but maybe a different way of representing the non-uniform binning histogram. 
 You first scale the original histogram by the bin width to represent a density, this is fine and then you generate random number from this scaled histogram. If you doing this you the histogram represent now a density and of course you need to account of the bin width when calling`TH1::GetRandom`. If you would use the original un-scaled histogram, it would be fine. 
 I am afraid adding the fix that you suggest it will add some inconsistency. Maybe the way to proceed is adding in GetRandom an option to consider the bin width in case of density histograms ?

Cheers

Lorenzo

andrea-celentano:
Dear Lorenzo,
I agree that this is a matter related to what a histogram is, and what the operation of extracting random numbers from it should be. I am totally in favor of adding an option in `TH1::GetRandom` to allow both methods.

ferdymercury:
attempt to support both options:

https://github.com/root-project/root/pull/19918

PR #19918, opened 2025-09-18
[hist] Allow GetRandom to work with density-histograms

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/9530

[Plz squash]

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/hist/hist/src/TH1.cxx
+++ b/hist/hist/src/TH1.cxx
@@ -2531,7 +2531,7 @@ Double_t TH1::ComputeIntegral(Bool_t onlyPositive, Option_t *option)
    for (Int_t binz=1; binz <= nbinsz; ++binz) {
       Double_t zWidth = (fDimension > 2) ? fZaxis.GetBinWidth(binz) : 1;
       for (Int_t biny=1; biny <= nbinsy; ++biny) {
-         Double_t yWidth= (fDimension > 1) ? fYaxis.GetBinWidth(biny) : 1;
+         Double_t yWidth = (fDimension > 1) ? fYaxis.GetBinWidth(biny) : 1;
          for (Int_t binx=1; binx <= nbinsx; ++binx) {
             Double_t xWidth = fXaxis.GetBinWidth(binx);
             ++ibin;
@@ -5023,7 +5023,7 @@ void TH1::GetBinXYZ(Int_t binglobal, Int_t &binx, Int_t &biny, Int_t &binz) cons
 /// @note Only valid for 1-d histograms. Use GetRandom2 or GetRandom3 otherwise.
 /// If the histogram has a bin with negative content, a NaN is returned.
 
-Double_t TH1::GetRandom(TRandom * rng, Option_t *option) const
+Double_t TH1::GetRandom(TRandom *rng, Option_t *option) const
 {
    if (fDimension > 1) {
       Error("GetRandom","Function only valid for 1-d histograms");
@@ -5033,10 +5033,11 @@ Double_t TH1::GetRandom(TRandom * rng, Option_t *option) const
    Double_t integral = 0;
    // compute integral checking that all bins have positive content (see ROOT-5894)
    if (fIntegral) {
-      if (fIntegral[nbinsx+1] != fEntries) integral = const_cast<TH1*>(this)->ComputeIntegral(true, option);
+      if (fIntegral[nbinsx + 1] != fEntries)
+         integral = const_cast<TH1 *>(this)->ComputeIntegral(true, option);
       else  integral = fIntegral[nbinsx];
    } else {
-      integral = const_cast<TH1*>(this)->ComputeIntegral(true, option);
+      integral = const_cast<TH1 *>(this)->ComputeIntegral(true, option);
    }
    if (integral == 0) return 0;
    // return a NaN in case some bins have negative content
--- a/hist/hist/inc/TH1.h
+++ b/hist/hist/inc/TH1.h
@@ -550,7 +550,7 @@ public:
    TVirtualHistPainter *GetPainter(Option_t *option="");
 
    virtual Int_t    GetQuantiles(Int_t n, Double_t *xp, const Double_t *p = nullptr);
-   virtual Double_t GetRandom(TRandom * rng = nullptr, Option_t *option = "") const;
+   virtual Double_t GetRandom(TRandom *rng = nullptr, Option_t *option = "") const;
    virtual void     GetStats(Double_t *stats) const;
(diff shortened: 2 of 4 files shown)


hist/hist/src/TH1.cxx
```diff
@@ -5021,10 +5031,10 @@ Double_t TH1::GetRandom(TRandom * rng) const
    Double_t integral = 0;
    // compute integral checking that all bins have positive content (see ROOT-5894)
    if (fIntegral) {
-      if (fIntegral[nbinsx+1] != fEntries) integral = ((TH1*)this)->ComputeIntegral(true);
+      if (fIntegral[nbinsx+1] != fEntries) integral = ((TH1*)this)->ComputeIntegral(true, option);
```
hageboeck: Shouldn't this work?
```suggestion
      if (fIntegral[nbinsx+1] != fEntries) integral = ComputeIntegral(true, option);
```

hist/hist/src/TH1.cxx
```diff
@@ -5021,10 +5031,10 @@ Double_t TH1::GetRandom(TRandom * rng) const
...
       else  integral = fIntegral[nbinsx];
    } else {
-      integral = ((TH1*)this)->ComputeIntegral(true);
+      integral = ((TH1*)this)->ComputeIntegral(true, option);
```
hageboeck:
```suggestion
      integral = ComputeIntegral(true, option);
```

hist/hist/src/TH1.cxx
```diff
@@ -2505,12 +2505,15 @@ void TH1::ClearUnderflowAndOverflow()
 ///  The resulting integral is normalized to 1.
 ///  If the routine is called with the onlyPositive flag set an error will
 ///  be produced in case of negative bin content and a NaN value returned
+///  \param option (optional) Set it to "width" if your non-uniform bin contents
+///  represent a density rather than counts. In that case, we internally multiply
+///  by the bin width/area so that fIntegral is in the "counts" and not on the original "density" domain.
```
hageboeck: We should maybe first explain what the default does, and then go to the options.
This is what I understand the function to do (please correct if wrong):
```suggestion
///  \param onlyPositive If set to true, an error will be produced and NaN will be returned
///  when a bin with negative number of entries is encountered.
///  \param option
///  - `""` (default) Compute the cumulative density function.
///  - `"width"` When using a histogram that has been scaled to represent densities instead of event counts, 
///     this options multiplies the densities with the bin width.
///     The cumulative function in that case represents counts instead of probabilities.
```

hist/hist/src/TH1.cxx
```diff
@@ -5005,13 +5014,14 @@ void TH1::GetBinXYZ(Int_t binglobal, Int_t &binx, Int_t &biny, Int_t &binz) cons
 /// is evaluated, normalized to one.
 ///
 /// @param rng (optional) Random number generator pointer used (default is gRandom)
+/// @param option (optional) Set it to "width" if your non-uniform bin contents represent a density rather than counts
 ///
 /// The integral is automatically recomputed if the number of entries
 /// is not the same then when the integral was computed.
-/// NB Only valid for 1-d histograms. Use GetRandom2 or 3 otherwise.
+/// @note Only valid for 1-d histograms. Use GetRandom2 or GetRandom3 otherwise.
 /// If the histogram has a bin with negative content a NaN is returned
```
hageboeck:
```suggestion
/// @note Only valid for 1-d histograms. Use GetRandom2 or GetRandom3 otherwise.
/// If the histogram has a bin with negative content, a NaN is returned.
```

hist/hist/src/TH1.cxx
ferdymercury:
```suggestion
///  \param onlyPositive If set to true, an error will be produced and NaN will be returned
///  when a bin with negative number of entries is encountered.
///  \param option
///  - `""` (default) Compute the cumulative density function assuming current bin contents represent counts.
///  - `"width"` Computes the cumulative density function assuming current bin contents represent densities.
```

hist/hist/src/TH1.cxx
```diff
@@ -2505,12 +2505,17 @@ void TH1::ClearUnderflowAndOverflow()
 ///  The resulting integral is normalized to 1.
 ///  If the routine is called with the onlyPositive flag set an error will
 ///  be produced in case of negative bin content and a NaN value returned
+///  \param onlyPositive If set to true, an error will be produced and NaN will be returned
+///  when a bin with negative number of entries is encountered.
+///  \param option
+///  - `""` (default) Compute the cumulative density function assuming current bin contents represent counts.
+///  - `"width"` Computes the cumulative density function assuming current bin contents represent densities.
```
ferdymercury, line 2512: In both cases, the result is a normalized CDF function. It's just that, to properly calculate it, the input of this function has always to be "counts". So, if the input is density, we pre-multiply it by "binwidth" so that it's counts now (in the auxiliary variable then passed to the function, original input is not changed).

hist/hist/src/TH1.cxx
ferdymercury: I get now: error: 'this' argument to member function 'ComputeIntegral' has type 'const TH1', but function is not marked const

hist/hist/src/TH1.cxx
```diff
@@ -5021,10 +5033,10 @@ Double_t TH1::GetRandom(TRandom * rng) const
    Double_t integral = 0;
    // compute integral checking that all bins have positive content (see ROOT-5894)
    if (fIntegral) {
-      if (fIntegral[nbinsx+1] != fEntries) integral = ((TH1*)this)->ComputeIntegral(true);
+      if (fIntegral[nbinsx+1] != fEntries) integral = ComputeIntegral(true, option);
```
ferdymercury:
```suggestion
      if (fIntegral[nbinsx+1] != fEntries) integral = const_cast<TH1*>(this)->ComputeIntegral(true, option);
```

hist/hist/src/TH1.cxx
```diff
@@ -5021,10 +5033,10 @@ Double_t TH1::GetRandom(TRandom * rng) const
...
       else  integral = fIntegral[nbinsx];
    } else {
-      integral = ((TH1*)this)->ComputeIntegral(true);
+      integral = ComputeIntegral(true, option);
```
ferdymercury:
```suggestion
      integral = const_cast<TH1*>(this)->ComputeIntegral(true, option);
```

hageboeck (approved):

Issue #19889, opened 2025-09-15
Mixing Clang 20 and ROOT causes very sporadic crashes


ktf:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Since we bumped to Clang 20 (while keeping everything else unchanged), we see the following stacktrace:

```
#0  0x000014a519934570 in cling::Interpreter::runAndRemoveStaticDestructors() () from /opt/alisw/el9/ROOT/v6-32-06-alice9-8/lib/libCling.so
#1  0x000014a519843d4c in TCling::ResetGlobals (this=0x1d68cd0) at /local/workspace/DailyBuilds/WeeklyO2Release/daily-tags.eyARqWDKEt/SOURCES/ROOT/v6-32-06-alice9/v6-32-06-alice9/core/metacling/src/TCling.cxx:3756                                                                                                                                                                                                                    
#2  0x000014a52aa5ba8b in TROOT::EndOfProcessCleanups (this=0x14a52acbb500 <ROOT::Internal::GetROOT1()::alloc>) at /local/workspace/DailyBuilds/WeeklyO2Release/daily-tags.eyARqWDKEt/SOURCES/ROOT/v6-32-06-alice9/v6-32-06-alice9/core/base/src/TROOT.cxx:1257                                                                                                                                                                          
#3  0x000014a52ab59fcf in TUnixSystem::Exit (this=<optimized out>, code=139, mode=true) at /local/workspace/DailyBuilds/WeeklyO2Release/daily-tags.eyARqWDKEt/SOURCES/ROOT/v6-32-06-alice9/v6-32-06-alice9/core/unix/src/TUnixSystem.cxx:2191                                                                                                                                                                                            
#4  0x000014a52ab6229b in TUnixSystem::DispatchSignals (this=0x19cf920, sig=<optimized out>) at /local/workspace/DailyBuilds/WeeklyO2Release/daily-tags.eyARqWDKEt/SOURCES/ROOT/v6-32-06-alice9/v6-32-06-alice9/core/unix/src/TUnixSystem.cxx:3675                                                                                                                                                                                       
#5  <signal handler called>
#6  std::local_Rb_tree_rotate_left (__x=0x2d977c0, __root=@0x2e16a00: 0x26f75a0) at ../../../../../gcc/libstdc++-v3/src/c++98/tree.cc:138
[...]

PR #19923, opened 2025-09-19
[core] Use version script to hide symbols in libCling

hahnjo:
We already use `-fvisibility=hidden` and `-fvisibility-inlines-hidden` to generate symbols that are not exported from the shared object. However, these two options don't apply for non-inline template instantiations, which can still cause problems when a STL container is instantiated with an LLVM type that may change size between LLVM versions.

Related to https://github.com/root-project/root/issues/19889

### DIFF
--- a/core/metacling/src/libCling.script
+++ b/core/metacling/src/libCling.script
@@ -1,0 +1,13 @@
+{
+  global:
+    CreateInterpreter;
+    DestroyInterpreter;
+    TCling__*;
+    _Z26TCling__TEST_isInvalidDeclP11ClassInfo_t;
+    TROOT__*;
+    ROOT_rootcling_Driver;
+    _ZN5cling*;
+    cling_runtime_internal_throwIfInvalidPointer;
+  local:
+    *;
+};
--- a/core/metacling/src/CMakeLists.txt
+++ b/core/metacling/src/CMakeLists.txt
@@ -229,6 +229,8 @@ if(NOT APPLE AND NOT MSVC)
   # and by ALICE in https://github.com/root-project/root/issues/19889
   # Only needed for Linux: Mac uses linker namespaces and  Windows explicit export/import
   target_link_libraries(Cling PRIVATE -Wl,-Bsymbolic)
+  target_link_libraries(Cling PRIVATE -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/libCling.script)
+  set_target_properties(Cling PROPERTIES LINK_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/libCling.script)
 endif()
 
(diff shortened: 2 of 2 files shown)


core/metacling/src/libCling.script
```diff
@@ -0,0 +1,13 @@
+{
```
vgvassilev, line 1: The linker script approach used to make the library artifact much larger than without and using hidden visibility.

hahnjo: Here I'm proposing to use the linker script in addition to all compilation flags that we already apply. In a `Release` build with GCC, this actually reduces the size of `libCling.so` by 822 kB (of 120 MB).

vgvassilev: Ah, okay, I see. Some linkers (wasm for sure) do not support linker scripts. Can we check if the linker supports a linker script and then use it?

vgvassilev: I am a little surprised to hear that the overall size is 120MB. It used to be between 30 and 50MB several years ago. Do we know what has changed?

hahnjo: The change is about building `libCling` in ROOT. I don't think we support wasm here, this is about linking on Linux where it works.

hahnjo: Regarding the size: ROOT v6.26 (released in early 2022) had LLVM 9 and `libCling.so` was already 73 MB. For ROOT v6.28 (released in early 2023), we upgraded to LLVM 13 and `libCling.so` grew to 100 MB. So the answer is that it's LLVM growing in size.

vgvassilev: Okay, that’s a bit unfortunate as we use the same surface naively.

hahnjo:
> If we forward declare llvm::DebugFlag would it still work after this change on the root prompt?

In principle nothing changes for symbol resolution from interpreted code, since we have a special handle to `libCling` anyway. However, I can't get it to work, even before the change. I tried:
```
root [0] namespace llvm { extern bool DebugFlag; }
root [1] ::llvm::DebugFlag = true;
root [2] std::cout << "test\n";
test
```
What am I missing?

vgvassilev:
> > If we forward declare llvm::DebugFlag would it still work after this change on the root prompt?
> 
> In principle nothing changes for symbol resolution from interpreted code, since we have a special handle to `libCling` anyway. However, I can't get it to work, even before the change. I tried:
> 
> ```
> root [0] namespace llvm { extern bool DebugFlag; }
> root [1] ::llvm::DebugFlag = true;
> root [2] std::cout << "test\n";
> test
> ```
> 
> What am I missing?

Okay, I thought that would help with symbol reachability.  In principle this should not be reachable. The advantage of the linker script is that we could export the cling runtime and hide everything else. This way we won’t need libCling in the jit.

PR #19925, opened 2025-09-19
[browsable] Change v7 draw as default for treemap visualization

magnustymoteus:
# This Pull request:

## Changes or fixes:
Make `kActDraw7` default for drawing `RTreeMapElement` on `RBrowser`

## Checklist:

- [X] tested changes locally

### DIFF
--- a/gui/browsable/src/RNTupleBrowseProvider.cxx
+++ b/gui/browsable/src/RNTupleBrowseProvider.cxx
@@ -68,13 +68,10 @@ public:
    }
 
    /** Default action is to draw the treemap */
-   EActionKind GetDefaultAction() const override { return kActDraw6; /*temporary, replace with kActDraw7; */ }
+   EActionKind GetDefaultAction() const override { return kActDraw7; }
 
    /** Check if visualization is capable of being drawn */
-   bool IsCapable(EActionKind kind) const override
-   {
-      return kind == kActDraw6;
-   }
+   bool IsCapable(EActionKind kind) const override { return kind == kActDraw6 || kind == kActDraw7; }
 
    /** Create item with TreeMap icon */
    std::unique_ptr<RItem> CreateItem() const override
@@ -476,4 +473,3 @@ public:
    ~RNTupleBrowseProvider() override { RegisterNTupleFunc(nullptr); }
 
 } newRNTupleBrowseProvider;
-


gui/browsable/src/RNTupleBrowseProvider.cxx
```diff
@@ -68,12 +68,15 @@ class RTreeMapElement : public RElement {
    }
 
    /** Default action is to draw the treemap */
-   EActionKind GetDefaultAction() const override { return kActDraw6; /*temporary, replace with kActDraw7; */ }
+   EActionKind GetDefaultAction() const override { return kActDraw7; }
```
linev, line 71: Why you want to change to RCanvas by default? 
In JSROOT code we used `ensureTCanvas` - so by default TCanvas is used for drawing inside jsroot.

gui/browsable/src/RNTupleBrowseProvider.cxx
```diff
@@ -68,12 +68,15 @@ class RTreeMapElement : public RElement {
...
 
    /** Check if visualization is capable of being drawn */
    bool IsCapable(EActionKind kind) const override
    {
-      return kind == kActDraw6;
+      if ((kind == kActDraw6) || (kind == kActDraw7))
+         return GetDefaultAction() == kActDraw7;
```
linev: In principle, both TCanvas and RCanvas drawing can be supported here.
So one can do:
```suggestion
         return true;
```

gui/browsable/src/RNTupleBrowseProvider.cxx
magnustymoteus: Because the TCanvas drawing for RTreeMap doesn't have interactive features since it uses the C++ `Paint` method of `RTreeMapPainter`. Opening histograms on RBrowser also gives the RCanvas drawing by default, so I don't see why not do the same for RTreeMap?

linev: For now `RTreeMapPainter` only works in the web mode, so `Paint` method is not important.

Actually, I do not understand why interactivity on TCanvas should not work. 

For now I propose following.
You make `kActDraw7` as default action, but:
```
   bool IsCapable(EActionKind kind) const override
   {
      return kind == kActDraw6 || kind == kActDraw7;
   }
```
So one can display object on TCanvas when it there and one can debug why interactivity does not work

linev (approved):
Let do like this

PR #19714, opened 2025-08-21
[ntuple] rename RInvalidField::RCategory to ECategory

silverweed:
# This Pull request:
renames `RInvalidField::RCategory` to `RInvalidField::ECategory`, more appropriate for an enum.
Marks `RCategory` as deprecated, to be removed in 6.40.

## Checklist:

- [x] tested changes locally
- [ ] updated the docs (if necessary)

PR #19929, opened 2025-09-19
[ntuple] remove GetCollectionInfo() from collection fields

jblomer:
Several collection fields still had a public GetCollectionInfo() method. It was an oversight not to remove them earlier. These methods are unneeded and potentially dangerous to expose.


silverweed:
Aren't we supposed to deprecate them and remove them in 6.40 (or even 42 given what we said [here](https://github.com/root-project/root/pull/19714) for `RInvalidField::RCategory`)?

jblomer:
I think this case is different, because it is really a bug to have them in in the first place. So, *if* they are used (unlikely), I'd actually prefer to see breakage.

silverweed (approved):

PR #19930, opened 2025-09-19
[hist] Enable chi2 tests for TProfile

hageboeck:
In reaction to this question
https://root-forum.cern.ch/t/chi2test-using-tprofile/64156

- Override the Chi2Test() function for TProfiles, so the "WW" option is used to obtain correct errors.
- Test that the profiles have the correct error option set (which is the error of the mean)
- Add unit tests.

**Note**: I decided to ignore clang-format for the headers, because the indentation would look out of place.

### DIFF
--- a/hist/hist/test/CMakeLists.txt
+++ b/hist/hist/test/CMakeLists.txt
@@ -29,6 +29,7 @@ ROOT_ADD_GTEST(testTGraphSorting test_TGraph_sorting.cxx LIBRARIES Hist)
 ROOT_ADD_GTEST(testSpline test_spline.cxx LIBRARIES Hist)
 ROOT_ADD_GTEST(testTF1Simple test_tf1_simple.cxx LIBRARIES Hist RIO)
 ROOT_ADD_GTEST(testTF1DrawCopy test_tf1_drawcopy.cxx LIBRARIES Hist)
+ROOT_ADD_GTEST(test_TProfile test_TProfile.cxx LIBRARIES Hist)
 
 if(fftw3)
   ROOT_ADD_GTEST(testTF1 test_tf1.cxx LIBRARIES Hist)
--- a/hist/hist/test/test_TProfile.cxx
+++ b/hist/hist/test/test_TProfile.cxx
@@ -1,0 +1,93 @@
+#include "gtest/gtest.h"
+
+#include "TProfile.h"
+#include "TProfile2D.h"
+#include "TProfile3D.h"
+#include "TRandom.h"
+
+#include "ROOT/TestSupport.hxx"
+
+template <typename T>
+void runTest(T const &reference, T const &sameDistr, T const &differentDistr)
+{
+   EXPECT_EQ(reference.Chi2Test(&sameDistr), reference.Chi2Test(&sameDistr, "WW"));
+   EXPECT_EQ(reference.Chi2Test(&sameDistr),
+             reference.Chi2Test(&sameDistr, "P WW UW")); // Need more than just the default option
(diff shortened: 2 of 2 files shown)


ferdymercury (approved):
Thanks! LGTM

hist/hist/src/TProfile.cxx
```diff
@@ -416,6 +416,25 @@ Int_t TProfile::BufferFill(Double_t x, Double_t y, Double_t w)
...
 
+////////////////////////////////////////////////////////////////////////////////
+/// Run a Chi2Test between two TProfiles.
+/// This calls TH1::Chi2Test() with the option "WW".
+
+Double_t TProfile::Chi2Test(const TH1 *h2, Option_t *option, Double_t *res) const
+{
+   TString opt = option;
+   opt.ToUpper();
+   opt += "WW";
```
wiso: should this happen only if h2 is a TProfile?
Actually, does it make sense to compute a chi2 between a TProfile and a non-TProfile?

hageboeck: Generally, you *could* compare something that's not a TProfile with a TProfile, but I don't see how this makes sense. I was thinking that users either compare "normal" histograms with normal histograms, or profiles with profiles. For mixed cases, it probably depends on the case whether this makes sense.

On second thought, I agree that this modification of the options is safe only when `h2` is also a profile. I'll change the code accordingly, and leave the options for the mixed cases as the user inputs them.

wiso: ok, maybe you can consider overloading only the case where the other histogram is actually a `TProfile` and not a generic `TH1`.

hageboeck: I considered this, but an overload `TProfile::Chi2Test(TProfile*, ...)` could not be called via a `TH1` pointer. So this one
```c++
TH1* profile = new TProfile(...);
profile->Chi2Test( new TProfile(...) );
```
would call `TH1::Chi2Test(TH1*)`. This might be OK if users use their TProfiles always as TProfiles, but maybe polymorphism isn't such a bad thing here, when you e.g. want to collect histograms and profiles in the same container.
For this reason, I tend towards overriding instead of overloading. That would get us back to casting `h2`, and if the cast succeeds, the options get set to `WW`. If it fails, the users can choose any option they desire, like you did with the original version of the function.

hist/hist/src/TProfile.cxx
```diff
@@ -416,6 +416,30 @@ Int_t TProfile::BufferFill(Double_t x, Double_t y, Double_t w)
...
+
+   if (auto other = dynamic_cast<const TProfile *>(h2); other) {
+      if (fErrorMode != kERRORMEAN || other->fErrorMode != kERRORMEAN)
+         Error("Chi2Test", "Chi2 tests need TProfiles in 'error of mean' mode.");
+
+      opt += "WW";
+      opt.ReplaceAll("UU", "WW");
+      opt.ReplaceAll("UW", "WW");
+   } else if (!opt.Contains("WW")) {
+      Error("Chi2Test", "TProfiles need to be tested with the 'W' option. Either use option 'WW' or use histogram.Chi2Test(<profile>, 'UW')");
```
guitargeek:
```suggestion
      Error("Chi2Test", "TProfiles need to be tested with the 'W' option. Either use option 'WW' or use histogram.Chi2Test(<profile>, 'UW')");
      return 0;
```

Maybe we should also return zero in the error case, to make it more clear to the users that something is wrong even if they don't read the logs? Also in the other three cases.

hageboeck:
> I would just prefer to return zero in the error case

Good idea! I'll change it to return 0.

guitargeek:
Thanks!

PR #19932, opened 2025-09-21
[clad] Check clad master before v2.1

vgvassilev:


vgvassilev:
@guitargeek, can you take a look at the failures and interpret them for us? IIUC, there is nothing clad related...

hahnjo:
You probably want a clean build because there may be files left over when just changing the `GIT_TAG` version

guitargeek:
> @guitargeek, can you take a look at the failures and interpret them for us? IIUC, there is nothing clad related...

There are comparison failures with numeric gradients now that fail all over the place, so the current failures seem Clad related.

I'll have time to investigate this later in the afternoon after the ROOT meeting. I can create a reproducer and open an issue in Clad a usual.

vgvassilev:
Thank you, @guitargeek!

cc: @petrozarytskyi

guitargeek:
I'm checking now Clad `master` with Petros possible fix on top:
  * https://github.com/root-project/root/pull/19961

guitargeek:
Note that while the ROOT CI will pass now, there is still a regression in Clad `master` that broke the CMS Combine likelihoods:
  * https://github.com/vgvassilev/clad/issues/1573

vgvassilev:
> Note that while the ROOT CI will pass now, there is still a regression in Clad `master` that broke the CMS Combine likelihoods:
> 
>     * [Regression in Clad `master` when aliasing array parameters in inner function vgvassilev/clad#1573](https://github.com/vgvassilev/clad/issues/1573)

Ah, okay, thanks for clarifying.

PR #19935, opened 2025-09-22
[ntuple] automatic evolution from fixed-size array to RVec

jblomer:
In contrast to other automatic evolution mechanisms, this one cannot use `ReconcileOnDiskField()` because for proper support the type of the field itself must change. In principle, we could stuff all the logic into RRecField, but it would be quite messy. We also discussed adding a "virtual column" for the fixed-size indexes; this doesn't work, however, because the index column stores offsets and not vector lengths, and it is also used to translate between global and local indexing.

Thus, this PR adds the infrastructure to replace fields of just created field hierarchies (models). This is legit as long as the fields weren't exposed to the user yet, where the user, e.g., could have already created entries. As a result, the automatic evolution from fixed-size to RVec (and later std::vector) won't work with imposed models. But it does work for views and models reconstructed from disk, which are the most common use cases.

The infrastructure will also be used to switch between streamer field and class field.

### DIFF
--- a/tree/ntuple/inc/ROOT/RField.hxx
+++ b/tree/ntuple/inc/ROOT/RField.hxx
@@ -208,7 +208,7 @@ protected:
    void ReadGlobalImpl(ROOT::NTupleSize_t globalIndex, void *to) final;
    void ReadInClusterImpl(RNTupleLocalIndex localIndex, void *to) final;
 
-   void BeforeConnectPageSource(ROOT::Internal::RPageSource &pageSource) final;
+   std::unique_ptr<RFieldBase> BeforeConnectPageSource(ROOT::Internal::RPageSource &pageSource) final;
    void ReconcileOnDiskField(const RNTupleDescriptor &desc) final;
 
 public:
@@ -263,7 +263,7 @@ protected:
    // Returns the list of seen streamer infos
    ROOT::RExtraTypeInfoDescriptor GetExtraTypeInfo() const final;
 
-   void BeforeConnectPageSource(ROOT::Internal::RPageSource &source) final;
+   std::unique_ptr<RFieldBase> BeforeConnectPageSource(ROOT::Internal::RPageSource &source) final;
    void ReconcileOnDiskField(const RNTupleDescriptor &desc) final;
 
 public:
--- a/tree/ntuple/inc/ROOT/RField.hxx
+++ b/tree/ntuple/inc/ROOT/RField.hxx
@@ -49,9 +49,26 @@ namespace Detail {
 class RFieldVisitor;
 } // namespace Detail
 
+class RFieldZero;
+namespace Internal {
+void SetAllowFieldSubstitutions(RFieldZero &fieldZero, bool val);
+}
+
 /// The container field for an ntuple model, which itself has no physical representation.
 /// Therefore, the zero field must not be connected to a page source or sink.
 class RFieldZero final : public RFieldBase {
+   friend void ROOT::Internal::SetAllowFieldSubstitutions(RFieldZero &, bool);
+
+   /// If field substitutions are allowed, upon connecting to a page source the field hierarchy will replace created
+   /// fields by fields that match the on-disk schema. This happens for
+   ///     - Vector fields (RVectorField, RRVecField) that connect to an on-disk fixed-size array
+   ///     - Streamer fields that connect to an on-disk class field
+   /// Field substitutions must not be enabled when the field hierarchy already handed out RValue objects because
+   /// they would leave dangling field pointers to the replaced fields. This is used in cases when the field/model
+   /// is created by RNTuple (not imposed), before it is made available to the user.
+   /// This flag is reset on Clone().
+   bool fAllowFieldSubstitutions = false;
+
 protected:
    std::unique_ptr<RFieldBase> CloneImpl(std::string_view newName) const final;
    void ConstructValue(void *) const final {}
@@ -64,6 +81,8 @@ public:
    size_t GetAlignment() const final { return 0; }
 
    void AcceptVisitor(ROOT::Detail::RFieldVisitor &visitor) const final;
+
+   bool GetAllowFieldSubstitutions() const { return fAllowFieldSubstitutions; }
 };
 
 /// Used in RFieldBase::Check() to record field creation failures.
--- a/tree/ntuple/inc/ROOT/RField.hxx
+++ b/tree/ntuple/inc/ROOT/RField.hxx
@@ -83,6 +83,8 @@ public:
    void AcceptVisitor(ROOT::Detail::RFieldVisitor &visitor) const final;
 
    bool GetAllowFieldSubstitutions() const { return fAllowFieldSubstitutions; }
+   /// Moves all subfields into the returned vector.
+   std::vector<std::unique_ptr<RFieldBase>> ReleaseSubfields();
 };
 
 /// Used in RFieldBase::Check() to record field creation failures.
--- a/tree/ntuple/src/RField.cxx
+++ b/tree/ntuple/src/RField.cxx
@@ -31,6 +31,11 @@
 #include <type_traits>
 #include <unordered_set>
 
+void ROOT::Internal::SetAllowFieldSubstitutions(RFieldZero &fieldZero, bool val)
+{
+   fieldZero.fAllowFieldSubstitutions = val;
+}
+
 std::unique_ptr<ROOT::RFieldBase> ROOT::RFieldZero::CloneImpl(std::string_view /*newName*/) const
 {
    auto result = std::make_unique<RFieldZero>();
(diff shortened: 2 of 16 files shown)


tree/ntuple/inc/ROOT/RField.hxx
```diff
@@ -49,9 +49,25 @@ namespace Detail {
...
+namespace Internal {
+void SetAllowFieldSubstitutions(RFieldZero &fieldZero, bool val);
+}
+
 /// The container field for an ntuple model, which itself has no physical representation.
 /// Therefore, the zero field must not be connected to a page source or sink.
 class RFieldZero final : public RFieldBase {
+   friend void ROOT::Internal::SetAllowFieldSubstitutions(RFieldZero &, bool);
+
+   /// If field substitions are allowed, upon connecting to a page source the field hierarchy will replace created
```
silverweed:
```suggestion
   /// If field substitutions are allowed, upon connecting to a page source the field hierarchy will replace created
```

tree/ntuple/src/RField.cxx
```diff
@@ -31,14 +31,28 @@
...
    for (auto &f : fSubfields)
       result->Attach(f->Clone(f->GetFieldName()));
    return result;
 }
 
+std::vector<std::unique_ptr<ROOT::RFieldBase>> ROOT::RFieldZero::ReleaseSubfields()
+{
+   std::vector<std::unique_ptr<ROOT::RFieldBase>> result;
+   result.reserve(fSubfields.size());
+   std::move(fSubfields.begin(), fSubfields.end(), std::back_inserter(result));
```
silverweed: I think you can also do
```suggestion
   std::vector<std::unique_ptr<ROOT::RFieldBase>> result;
   std::swap(fSubfields, result);
```

tree/ntuple/src/RFieldBase.cxx
```diff
@@ -1004,7 +1004,26 @@ void ROOT::RFieldBase::ConnectPageSource(ROOT::Internal::RPageSource &pageSource
...
       throw RException(R__FAIL("setting description only valid when connecting to a page sink"));
 
-   BeforeConnectPageSource(pageSource);
+   auto substitute = BeforeConnectPageSource(pageSource);
+   if (substitute) {
+      const RFieldBase *itr = this;
+      while (itr->GetParent()) {
+         itr = itr->GetParent();
+      }
+      if (auto fieldZero = dynamic_cast<const RFieldZero *>(itr); fieldZero->GetAllowFieldSubstitutions()) {
```
silverweed: This should probably be an `assert` + `static_cast` (since we're not checking the returned pointer)

tree/ntuple/src/RFieldSequenceContainer.cxx
```diff
@@ -253,30 +254,28 @@ std::size_t ROOT::RRVecField::AppendImpl(const void *from)
...
+      throw RException(R__FAIL("RVec too large: " + std::to_string(nItems)));
+   }
 
-   // Read collection info for this entry
-   ROOT::NTupleSize_t nItems;
-   RNTupleLocalIndex collectionStart;
-   fPrincipalColumn->GetCollectionInfo(globalIndex, &collectionStart, &nItems);
-   char *begin = reinterpret_cast<char *>(*beginPtr); // for pointer arithmetics
+   auto [beginPtr, sizePtr, capacityPtr] = GetRVecDataMembers(rvec);
+   auto begin = *reinterpret_cast<unsigned char **>(beginPtr); // for pointer arithmetics
```
silverweed: I would rather write this as the slightly more convincingly type-safe
```cpp
auto begin = static_cast<unsigned char *>(*beginPtr);
```
(same for the other cases)

tree/ntuple/src/RFieldSequenceContainer.cxx
```diff
@@ -748,52 +778,10 @@ std::unique_ptr<ROOT::RFieldBase> ROOT::RArrayAsRVecField::CloneImpl(std::string
 void ROOT::RArrayAsRVecField::ConstructValue(void *where) const
 {
    // initialize data members fBegin, fSize, fCapacity
+   // currently the inline buffer is left uninitialized
    void **beginPtr = new (where)(void *)(nullptr);
    std::int32_t *sizePtr = new (reinterpret_cast<void *>(beginPtr + 1)) std::int32_t(0);
```
silverweed:
```suggestion
   std::int32_t *sizePtr = new (static_cast<void *>(beginPtr + 1)) std::int32_t(0);
```

tree/ntuple/test/ntuple_evolution_type.cxx
```diff
@@ -207,3 +207,24 @@ TEST(RNTupleEvolution, CheckPairTuple)
...
+      auto writer = ROOT::RNTupleWriter::Recreate(std::move(model), "ntpl", fileGuard.GetPath());
+
+      *a = {1, 2};
+      writer->Fill();
+   }
+
+   auto reader = RNTupleReader::Open("ntpl", fileGuard.GetPath());
+
+   auto a = reader->GetView<ROOT::RVec<short>>("a");
+   EXPECT_TRUE(typeid(a.GetField()) == typeid(ROOT::RArrayAsRVecField));
```
silverweed: Can't this be an `EXPECT_EQ` as well?

tree/ntuple/src/RFieldBase.cxx
jblomer: Indeed, this condition is flawed; rewritten as
```
if (typeid(*itr) == typeid(RFieldZero) && static_cast<const RFieldZero *>(itr)->GetAllowFieldSubstitutions()) {
```

tree/ntuple/src/RFieldSequenceContainer.cxx
jblomer: Fixed by a change to `GetRVecDataMembers()`

tree/ntuple/src/RFieldBase.cxx
silverweed: Is there a legitimate case where `typeid(*itr) != typeid(RFieldZero)`? If not I think the assertion would be more appropriate

jblomer: Yes, fields may be free-standing and not rooted in an `RFieldZero`

silverweed (approved):
LGTM but I would wait for a second approval.
Also see the warnings about typeid in the test

tree/ntuple/src/RField.cxx
```diff
@@ -31,9 +31,15 @@
...
 
+void ROOT::Internal::SetAllowFieldSubstitutions(RFieldZero &fieldZero, bool val)
+{
+   fieldZero.fAllowFieldSubstitutions = val;
+}
+
 std::unique_ptr<ROOT::RFieldBase> ROOT::RFieldZero::CloneImpl(std::string_view /*newName*/) const
 {
    auto result = std::make_unique<RFieldZero>();
+   result->fAllowFieldSubstitutions = fAllowFieldSubstitutions;
```
hahnjo: Do we want to preserve this on cloning or should it be reset?

tree/ntuple/test/ntuple_evolution_shape.cxx
```diff
@@ -956,6 +956,51 @@ struct RenamedIntermediateDerived : public RenamedIntermediate2 {
    }
 }
 
+TEST(RNTupleEvolution, ArrayAsVector)
+{
+   // RNTuple currently does not support automatic schema evolution when a class is renamed.
```
hahnjo: This comment should probably be removed

tree/ntuple/test/ntuple_largevector.cxx
```diff
@@ -0,0 +1,45 @@
...
+   FileRaii fileGuard("test_ntuple_large_vector.root");
+
+   // write out a vector too large for RVec
+   {
+      auto m = RNTupleModel::Create();
+      auto vec = m->MakeField<std::vector<int8_t>>("v");
+      auto writer = RNTupleWriter::Recreate(std::move(m), "r", fileGuard.GetPath());
+      vec->push_back(1);
+      writer->Fill();
+      vec->resize(std::numeric_limits<std::int32_t>::max());
```
hahnjo, line 16: I think this requires 8 GiB of memory, I'm not sure if we want to run this in the CI by default...

tree/ntuple/src/RField.cxx
jblomer: Good point! I think we want it reset.

tree/ntuple/test/ntuple_largevector.cxx
jblomer: It's 4G according to /usr/bin/time, but still... 
Disabled the test.

hahnjo (approved):

PR #19937, opened 2025-09-22
[ntuple] automatic evolution in and out std::atomic

jblomer:
Allow for std::atomic<T> --> T (or compatible) automatic conversion and vice versa.


tree/ntuple/test/ntuple_evolution_type.cxx
```diff
@@ -207,3 +207,35 @@ TEST(RNTupleEvolution, CheckPairTuple)
...
    EXPECT_DOUBLE_EQ(2.0, std::get<1>(p(0)));
 }
+
+TEST(RNTupleEvolution, CheckAtomic)
+{
+   FileRaii fileGuard("test_ntuple_evolution_check_atomic.root");
+   {
+      auto model = ROOT::RNTupleModel::Create();
+      auto a = model->MakeField<std::atomic<std::int32_t>>("a");
+      auto i = model->MakeField<std::int32_t>("i");
```
silverweed: Maybe we should also test with a class type (and one where `is_lock_free` is false)

tree/ntuple/src/RField.cxx
```diff
@@ -1050,8 +1050,11 @@ void ROOT::RAtomicField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
    static const std::vector<std::string> prefixes = {"std::atomic<"};
```
hahnjo: not used anymore

tree/ntuple/test/CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ ROOT_GENERATE_DICTIONARY(RNTupleDescriptorDict ${CMAKE_CURRENT_SOURCE_DIR}/RNTup
                          DEPENDENCIES RIO)
 
 ROOT_ADD_GTEST(ntuple_endian ntuple_endian.cxx LIBRARIES ROOTNTuple)
-ROOT_ADD_GTEST(ntuple_evolution_type ntuple_evolution_type.cxx LIBRARIES ROOTNTuple)
+ROOT_ADD_GTEST(ntuple_evolution_type ntuple_evolution_type.cxx LIBRARIES ROOTNTuple atomic)
```
hahnjo: Should this use `ROOT_ATOMIC_LIBS`? Not sure...

jblomer: I think it shouldn't. `ROOT_ATOMIC_LIBS` contains `atomic` when it is needed even for the normally lock-free cases, AFAICS.

pcanal: That is not my reading.  It seems to be literally that `ROOT_ATOMIC_LIBS` is the result of a test that checks whether `std::atomic` linking require an 'atomic' library or not.

tree/ntuple/src/RField.cxx
```diff
@@ -1096,11 +1096,12 @@ std::unique_ptr<ROOT::RFieldBase> ROOT::RAtomicField::CloneImpl(std::string_view
...
 {
-   static const std::vector<std::string> prefixes = {"std::atomic<"};
-
    const auto &fieldDesc = desc.GetFieldDescriptor(GetOnDiskId());
-   EnsureMatchingOnDiskField(fieldDesc, kDiffTypeName).ThrowOnError();
-   EnsureMatchingTypePrefix(fieldDesc, prefixes).ThrowOnError();
+   if (fieldDesc.GetTypeName().rfind("std::atomic<", 0) == 0) {
+      EnsureMatchingOnDiskField(fieldDesc, kDiffTypeName).ThrowOnError();
+   } else {
+      fSubfields[0]->SetOnDiskId(GetOnDiskId());
```
pcanal, line 1172: I am missing (not seeing) the part of the code that make sure that the non-atomic value is compatible with the in memory type.  For example:
```
auto v10 = reader->GetView<CustomAtomicNotLockFree>("atomicInt");
```
in the new test should still throw.

tree/ntuple/test/CMakeLists.txt
jblomer: On my machine (arch), `ROOT_ATOMIC_LIBS` is empty. If I use `is_lock_free()`, I need `-latomic`. For `std::atomic<int>` etc. I don't.

tree/ntuple/src/RField.cxx
jblomer: The new test already tests that `reader->GetView<std::atomic<std::byte>>("atomicInt")` throws.

The `RFieldBase::ConnectPageSource()` method is recursive. Checking of the inner type is done when it comes to calling the method on the subfield.

tree/ntuple/test/CMakeLists.txt
jblomer: But there is obviously an issue on EL platforms. I'm confused. Perhaps the check for `ROOT_ATOMIC_LIBS` is incomplete/buggy for arch.

tree/ntuple/src/RField.cxx
pcanal: > The new test already tests that reader->GetView<std::atomic<std::byte>>("atomicInt") throws.

Is that exercising the branch on line 1101 or 1103?

jblomer: You're right, the existing line tests something different. I added `reader->GetView<CustomAtomicNotLockFree>("atomicInt")` (which throws as expected).

tree/ntuple/test/CMakeLists.txt
jblomer: For now, my suggestion is to merge the PR without testing `std::atomic<MyClass>` and to add that later once we figured out the linking issue.

pcanal: That is fine as long as we keep track either by opening the next/new PR 'right away' or opening an issue.

tree/ntuple/src/RField.cxx
jblomer: @pcanal Can this be resolved?

tree/ntuple/test/CMakeLists.txt
jblomer: See #20078

tree/ntuple/src/RField.cxx
pcanal: Yes.

pcanal: Thanks.

Issue #19879, opened 2025-09-14
add user-defined handling of NaN and inf to RDataFrame statistics


ynikitenko:
### Feature description

I analysed a tree in PyROOT with RDataFrame. Statistics for most variables could be calculated correctly. However, for one column I got nan-s. After further inspection it turned out that one event from 8 millions contained a NaN in that column. I tried *Stats* and *Mean*, and they both returned NaN. NaN is the correct answer for that data, but it is not practical.

The possible solutions could be:

1. RDataFrame could optionally filter irregular numbers (like NaN or inf).
2. Statistics functions of RDataFrame optionally support that. Tree Drawer simply ignored that value and calculated the mean (already noted [here](https://root-forum.cern.ch/t/new-defined-column-in-rdataframe-return-nan-value/42403/6)). One could provide an option to Stats() or Mean() to ignore NaN-s.
3. Documentation is added for user to handle that. There is a similar section [Working with missing values in the dataset](https://root.cern/doc/master/classROOT_1_1RDataFrame.html#missing-values) in the RDataFrame reference. NaNs are common in statistical analysis, so they would be worth explanation too.

The approach with RDataFrame looks cleaner; at the same time, another method like "FilterNaNs" for RDataFrame looks like feature creep. In Python that could be solved with a keyword argument (option); for C++ it is less clear to me (5 overloaded functions neither look good). An improved guideline is always possible.

### Alternatives considered

I could solve it in PyROOT with

`ROOT.gInterpreter.Declare("auto myisnan = static_cast<bool (*)(float)>(&isnan);")
dr_cmx_no_cut = df.Filter("!myisnan(displacement.dr_proj_cmx_cm)")\
                                   .Stats("displacement.dr_proj_cmx_cm")
`

However, that is a) non-evident, especially for Python users, b) filtering the same column with a dedicated function looks redundant (maybe there is a better syntax for that).

### Additional context

Not sure whether it is connected, but there was a similar [topic](https://root-forum.cern.ch/t/rdataframe-with-branch-contain-nan/41653) on the ROOT Forum.

ferdymercury:
In Scipy they solve this with an extra arg: https://docs.scipy.org/doc/scipy/dev/api-dev/nan_policy.html

dpiparo:
Thanks for taking the time to share your user story: this approach helps us greatly!

[...]

PR #19939, opened 2025-09-22
[nfc][RDF] mention shortcut for omitting NaNs when calculating mean

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/19879

### DIFF
--- a/tree/dataframe/src/RDataFrame.cxx
+++ b/tree/dataframe/src/RDataFrame.cxx
@@ -86,6 +86,7 @@ You can directly see RDataFrame in action in our [tutorials](https://root.cern/d
    - [Creating an RDataFrame from a dataset specification file](\ref rdf-from-spec)
    - [Adding a progress bar](\ref progressbar)
    - [Working with missing values in the dataset](\ref missing-values)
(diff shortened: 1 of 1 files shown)


tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1856,6 +1857,19 @@ Note that working with missing values is currently supported with a TTree-based
...
+
+RDataFrame does not treat NaNs or Inf values in any special way, ie they will
+propagate to the final result. There is no configurable flag such as Scipy's
+`nan_policy=propagate,omit,raise`.
+The `omit` behavior can be easily mimicked by a user call to a filter, e.g.:
+
+\code{.py}
+df.Filter("!std::isnan(x)").Mean("x")
+df.Filter("!std::isinf(x)").Mean("x")
+\endcode
```
```suggestion
RDataFrame does not treat NaNs or infinities beyond what the floating-point standards require, i.e. they will
propagate to the final result.
Non-finite numbers can be suppressed using Filter(), e.g.:

\code{.py}
df.Filter("std::isfinite(x)").Mean("x")
\endcode
```

PR #19941, opened 2025-09-22
[CMake] Bump the minimum required version to 3.23

ferdymercury:
This version of CMake was released in March 2022
https://github.com/Kitware/CMake/releases/tag/v3.23.0

This version is required to use FILE_SETs, which could help towards https://github.com/root-project/root/issues/16327 and https://github.com/root-project/root/pull/18419.

### DIFF
--- a/hist/histv7/standalone.cmake
+++ b/hist/histv7/standalone.cmake
@@ -13,8 +13,8 @@ target_include_directories(ROOTHist INTERFACE
 target_compile_features(ROOTHist INTERFACE cxx_std_17)
 
 # Install header files manually: PUBLIC_HEADER has the disadvantage that CMake
-# flattens the directory structure on install, and FILE_SETs are only available
-# with CMake v3.23.
+# flattens the directory structure on install
+# TO-DO: use FILE_SETs instead
 install(FILES ${histv7_headers} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/ROOT)
 
 include(CTest)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 # For the licensing terms see $ROOTSYS/LICENSE.
 # For the list of contributors see $ROOTSYS/README/CREDITS.
 
-cmake_minimum_required(VERSION 3.20 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.23 FATAL_ERROR)
 
 if(WIN32)
   # Set CMP0091 (MSVC runtime library flags are selected by an abstraction) to OLD
@@ -42,7 +42,7 @@ foreach(policy ${policy_new})
   endif()
 endforeach()
 
-set(policy_old CMP0116)
+set(policy_old CMP0116 CMP0126)
 foreach(policy ${policy_old})
   if(POLICY ${policy})
     cmake_policy(SET ${policy} OLD)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 # For the licensing terms see $ROOTSYS/LICENSE.
 # For the list of contributors see $ROOTSYS/README/CREDITS.
 
-cmake_minimum_required(VERSION 3.20 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.23 FATAL_ERROR)
 
 if(WIN32)
   # Set CMP0091 (MSVC runtime library flags are selected by an abstraction) to OLD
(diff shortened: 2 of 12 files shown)


guitargeek:
Unfortunately, we still have older CMake versions on the platforms that we still support:

* ubuntu2204 - cmake version 3.22.1 (end of life April 2027)
* debian125 - cmake version 3.25.1 (end of life [30 Jun 2028](https://endoflife.date/debian))
* alma8 - cmake version 3.26.5
* ubuntu2404 - cmake version 3.28.3
* macOS 14 - cmake version 3.29.5

On Windows, I don't know.

So I think in practice, that means we can only require CMake 3.23+ when Ubuntu 22.04 is end-of-life in two years :slightly_frowning_face:

By the way, so far our strategy has been to just use the [same minimum CMake version as LLVM](https://github.com/llvm/llvm-project/blob/main/llvm/CMakeLists.txt#L2).

ferdymercury:
@dannf @roehling
Is there any remote plan / chance to bump Ubuntu's 22.04 cmake version from 3.22 to 3.23 ?

ferdymercury:
> So I think in practice, that means we can only require CMake 3.23+ when Ubuntu 22.04 is end-of-life in two years

Question: wouldn't it be an option to ask those users to install cmake from snap or from the official cmake PPA in that corner case?

ferdymercury:
Thanks, but it seems cmake is not found in the Ubu22 image

bellenot:
AFAIK, we never force users to take another version of CMake than the one installed by default on the distros. That's what defines the minimal version of CMake. At least it was like this so far. @dpiparo can correct me if I'm wrong

guitargeek:
> Thanks, but it seems cmake is not found in the Ubu22 image

That's just an effect of the incremental build I guess

> AFAIK, we never force users to take another version of CMake than the one installed by default on the distros. That's what defines the minimal version of CMake. At least it was like this so far. @dpiparo can correct me if I'm wrong

Yes, that's what we have done so far, but maybe at some point we should discuss if we should keep it like that when Ubuntu doesn't update it's CMake version in 5 years at all. Even RedHat does that.

The only thing that I changed now was to install a newer CMake version on the Ubuntu 22.04 CI image, which is independent of what we require from our users.

guitargeek:
@ferdymercury, you have any idea what is going on with all these failures? Maybe it would help to rebase the commit on master and push again to this PR branch?

ferdymercury:
Force-pushing didn't help.

It might be the underlying call to cmake_policy(3.23) causes issues.?

I checked in https://man.archlinux.org/man/cmake-policies.7.en#Policies_Introduced_by_CMake_3.23
but didn't see anything obvious.

Not sure if we need to change anything else in our CMakeLists around these lines:
```
set(policy_new CMP0072 CMP0076 CMP0077 CMP0079
    CMP0156 CMP0179 #deduplicate static libraries when linker supports it
)
set(policy_old CMP0116 CMP0135 CMP0144)
```

ferdymercury:
I can reproduce the failure locally on Ubuntu22 (only after clearing full build dir)

The problem is from 3.20 to 3.21:

This works:
```
cmake_minimum_required(VERSION 3.23 FATAL_ERROR)
cmake_policy(VERSION 3.20)
```

This or higher fails:
```
cmake_minimum_required(VERSION 3.23 FATAL_ERROR)
cmake_policy(VERSION 3.21)
```

So maybe related with https://man.archlinux.org/man/cmake-policies.7.en#CMP0126

ferdymercury:
Fixed now.

hageboeck:
👉 https://github.com/root-project/root/pull/19941#issuecomment-3320164536

ferdymercury:
Side note: the merged builtin-libgif patch (https://github.com/root-project/root/pull/21578) requires now 3.23 already in master

guitargeek:
True, but that's only a requirement if you build with `builtin_gif=ON`, which usually doesn't happen on Linux as giflib is available. So always requiring CMake 3.23 when building ROOT would still be more drastic (I'm not against it, just making a note here).

ferdymercury:
> By the way, so far our strategy has been to just use the [same minimum CMake version as LLVM](https://github.com/llvm/llvm-project/blob/main/llvm/CMakeLists.txt#L2).

Related discussion: https://discourse.llvm.org/t/rfc-2026-upgrading-llvm-s-minimum-required-cmake-version/89829

ferdymercury:
New discussion: https://discourse.llvm.org/t/rfc-raising-minimum-required-cmake-version-to-3-31

And first PRs are already within Clang (https://github.com/llvm/llvm-project/commit/e192ec42b426f6ffbc25419238a90f3c59088587)

hist/histv7/standalone.cmake
```diff
@@ -13,8 +13,8 @@ target_include_directories(ROOTHist INTERFACE
 target_compile_features(ROOTHist INTERFACE cxx_std_17)
 
 # Install header files manually: PUBLIC_HEADER has the disadvantage that CMake
-# flattens the directory structure on install, and FILE_SETs are only available
-# with CMake v3.23.
+# flattens the directory structure on install
+# TO-DO: use FILE_SETs instead
```
hahnjo: There is a separate `cmake_minimum_required` in `hist/histv7/CMakeLists.txt`. However, I don't want to bump this yet (unless there are really good reasons), so please revert this change.

ferdymercury: Done!

PR #19943, opened 2025-09-22
Never append to library path in `root` executable

guitargeek:
This follows up on 26d24de5db607, where it was made sure that ROOT executables always set the required RPATHs to the ROOT libraries, unless the user disables it with `CMAKE_SKIP_INSTALL_RPATH` at build time.

Therefore, the code path to set the `LD_LIBRARY_PATH` in the `root` executable was still kept for the case where `CMAKE_SKIP_INSTALL_RPATH` was enabled.

However, when thinking about it further, this is not required. The reason to set `CMAKE_SKIP_INSTALL_RPATH` is usually because ROOT is intended to be installed into directories that are already in the paths, like in the MacPorts configuration. So if the user doesn't need rpaths, they also don't need ROOT to figure out its library paths.

Therefore, this commit suggests so simply never set the library path in the `root` executable.

Closes Closed by: https://its.cern.ch/jira/browse/ROOT-10697

### DIFF
--- a/rootx/CMakeLists.txt
+++ b/rootx/CMakeLists.txt
@@ -27,7 +27,3 @@ target_include_directories(root PRIVATE
   ${CMAKE_SOURCE_DIR}/core/meta/inc    # for TGenericClassInfo.h
   ${CMAKE_BINARY_DIR}/ginclude         # for RConfigure.h and rootCommandLineOptionsHelp.h
 )
-
-if(NOT MSVC AND NOT CMAKE_SKIP_INSTALL_RPATH)
-  target_compile_definitions(root PRIVATE IS_RPATH_BUILD)
-endif()
--- a/rootx/src/rootx.cxx
+++ b/rootx/src/rootx.cxx
@@ -8,10 +8,6 @@
 // Rootx is a small front-end program that starts the main ROOT module. //
 // This program is called "root" in the $ROOTSYS/bin directory and the  //
 // real ROOT executable is now called "root.exe" (formerly "root").     //
-// Rootx puts up a splash screen giving some info about the current     //
-// version of ROOT and, more importantly, it sets up the required       //
-// LD_LIBRARY_PATH, SHLIB_PATH and LIBPATH environment variables        //
-// (depending on the platform).                                         //
 //                                                                      //
 //////////////////////////////////////////////////////////////////////////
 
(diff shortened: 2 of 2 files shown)


ferdymercury:
Thanks! Is the Dlopen trick in TUnixSystem still needed?
https://github.com/root-project/root/blob/f28225c19f29deffe00728c8350d92b8cfa0da7f/core/unix/src/TUnixSystem.cxx#L4625-L4650

guitargeek:
I think that's unrelated and maybe still needed to fix [https://its.cern.ch/jira/browse/ROOT-5858](https://its.cern.ch/jira/browse/ROOT-5858). See also the commit that reproduces it:

  * https://github.com/root-project/root/commit/0a0b2dd9a4a56e978f444f3d024b67fc4b7905b4

Why you're asking? Is there another GitHub issue or Jira ticket related to it?

ferdymercury:
> Why you're asking?

Just because it was mentioned in the description of https://its.cern.ch/jira/browse/ROOT-9758

pcanal:
>  The reason to set CMAKE_SKIP_INSTALL_RPATH is usually because ROOT is intended to be installed into directories that are already in the paths, 

Can you produce a relocatable build without `CMAKE_SKIP_INSTALL_RPATH` ?

> because ROOT is intended to be installed into directories that are already in the paths

Yes maybe ... but one of the 'point' of the `root` executable is to be (as opposed to `root.exe`) self sufficient (i.e. no need to set `LD_LIBRARY_PATH` nor `ROOTSYS`) even if the build was relocated.

guitargeek:
> > The reason to set CMAKE_SKIP_INSTALL_RPATH is usually because ROOT is intended to be installed into directories that are already in the paths,

Yes, that works with ROOT master right now and will continue to work with this PR, which removes code path that is already disabled via preprocessor macro in the case of `CMAKE_SKIP_INSTALL_RPATH=OFF`.

> Can you produce a relocatable build without `CMAKE_SKIP_INSTALL_RPATH` ?
> 
> > because ROOT is intended to be installed into directories that are already in the paths
> 
> Yes maybe ... but one of the 'point' of the `root` executable is to be (as opposed to `root.exe`) self sufficient (i.e. no need to set `LD_LIBRARY_PATH` nor `ROOTSYS`) even if the build was relocated.

Sorry maybe the PR description was not clear enough. I meant *if* the user has set `CMAKE_SKIP_INSTALL_RPATH=ON` in the build (which is a rare case and in my experience only done by packagers), they probably also intend to install ROOT into directories that are already in the paths, as they explicitly chose to neutralize the RPATH mechanism.

PR #19946, opened 2025-09-22
Make sure we consistently use system openblas in SOFIE GNN tests

guitargeek:
This avoids crashes because of mixing openblas versions on some platforms.

The second commit includes the actual fix, the first commit is just code formatting.


guitargeek:
The remaining test failures are unrelated to this PR.

Issue #19942, opened 2025-09-22
allow setting name of TStatistic or RDataFrame.Stats


ynikitenko:
### Explain what you would like to see improved and how.

I use RDataFrame.Stats. That does not allow setting the name of the resulting statistics. This is a pity, since RDataFrame by design allows producing many results at once.
At the same time, TStatistics does not allow changing the name after its creation.

As a result, printing statistics results in many columns with "OBJ: TStatistic". 

It would be good to allow setting name to TStatistic. Alternatively, one may allow setting name in the RDataFrame.Stats method.

### ROOT version

Any.

### Installation method

any

### Operating system

any

ferdymercury:
would this work for you, for an object st of class TStatistic?

`st.SetName("myname")` ?

ynikitenko:
> would this work for you, for an object st of class TStatistic?
> 
> `st.SetName("myname")` ?

Yes, this looks good. Honestly, I would also check other ROOT objects for that property (like a check list for good design). I've also noted that there is no copy constructor for TStatistic (which could set a new name and mitigate that problem).
But this solution looks good and necessary. Thank you.

PR #19948, opened 2025-09-22
[math] allow setting TStatistic name

ferdymercury:
Closes https://github.com/root-project/root/issues/19942


guitargeek:
Damn I'm sorry I didn't squash the commits! My bad, it didn't enter my mind that this simple addition could be more than one commit, so I didn't even check and only noticed when scrolling the page back up after merging :slightly_frowning_face:

PR #19958, opened 2025-09-23
[rfile] Introduce RFile first prototype

silverweed:
This PR adds a first version of the new RFile prototype, implementing basic Open/Append/Recreate/Get/Put methods. 

For a sneak peek of the final shape of the prototype, see [here](https://github.com/root-project/root/pull/19151/)

## Checklist:

- [x] tested changes locally

### DIFF
--- a/io/io/inc/ROOT/RFile.hxx
+++ b/io/io/inc/ROOT/RFile.hxx
@@ -131,13 +131,18 @@ public:
 
    ///// Factory methods /////
 
-   /// Opens the file for reading
+   /// Opens the file for reading. `path` may be a regular file path or a remote URL.
+   /// \throw ROOT::RException if the file at `path` could not be opened.
    static std::unique_ptr<RFile> Open(std::string_view path);
 
-   /// Opens the file for reading/writing, overwriting it if it already exists
+   /// Opens the file for reading/writing, overwriting it if it already exists.
+   /// \throw ROOT::RException if a file could not be created at `path` (e.g. if the specified
+   /// directory tree does not exist).
    static std::unique_ptr<RFile> Recreate(std::string_view path);
 
-   /// Opens the file for updating
+   /// Opens the file for updating, creating a new one if it doesn't exist.
+   /// \throw ROOT::RException if the file at `path` could neither be read nor created
+   /// (e.g. if the specified directory tree does not exist).
    static std::unique_ptr<RFile> Update(std::string_view path);
 
    ///// Instance methods /////
@@ -185,7 +190,7 @@ public:
       PutInternal(path, obj, flags);
    }
 
-   /// Writes all objects to disk with the file structure.
+   /// Writes all objects and the file structure to disk.
    /// Returns the number of bytes written.
    size_t Flush();
 
--- a/io/io/inc/ROOT/RFile.hxx
+++ b/io/io/inc/ROOT/RFile.hxx
@@ -34,7 +34,66 @@ ROOT::RLogChannel &RFileLog();
 \ingroup RFile
 \brief An interface to read from, or write to, a ROOT file, as well as performing other common operations.
 
-TODO: more in-depth explanation
+## When and why should you use RFile
+
+RFile is a modern and minimalistic interface to ROOT files, both local and remote, that can be used instead of TFile
+when the following conditions are met:
+- you want a simple interface that makes it easy to do things right and hard to do things wrong;
+- you only need basic Put/Get operations and don't need the more advanced TFile/TDirectory functionalities;
+- you want more robustness and better error reporting for those operations;
+- you want clearer ownership semantics expressed through the type system rather than having objects "automagically"
+  handled for you via implicit ownership of raw pointers.
+
+RFile doesn't try to cover the entirety of use cases covered by TFile/TDirectory/TDirectoryFile and is not
+a 1:1 replacement for them. It is meant to simplify the most common use cases and make them easier to handle by
+minimizing the amount of ROOT-specific quirks and conforming to more standard C++ practices.
+
+## Ownership model
+
+RFile handles ownership via smart pointers, typically std::unique_ptr.
+
+When getting an object from the file (via RFile::Get) you get back a unique copy of the object. Calling `Get` on the
+same object twice produces two independent clones of the object. The ownership over that object is solely on the caller
+and not shared with the RFile. Therefore, the object will remain valid after closing or destroying the RFile that
+generated it. This also means that any modification done to the object are **not** reflected to the file automatically:
+to update the object in the file you need to write it again (via RFile::Overwrite).
+
+RFile::Put and RFile::Overwrite are the way to write objects to the file. Both methods take a const reference to the
+object to write and don't change the ownership of the object in any way. Calling Put or Overwrite doesn't guarantee that
+the object is immediately written to the underlying storage: to ensure that, you need to call RFile::Flush (or close the
+file).
+
+## Directories
+
+Differently from TFile, the RFile class itself is not also a "directory". In fact, there is no RDirectory class at all.
+
+Directories are still an existing concept in RFile (since they are a concept in the ROOT binary format),
+but they are usually interacted with indirectly, via the use of filesystem-like string-based paths. If you Put an object
+in an RFile under the path "path/to/object", "object" will be stored under directory "to" which is in turn stored under
+directory "path". This hierarchy is encoded in the ROOT file itself and it can provide some optimization and/or
+conveniencies when querying objects.
+
+For the most part, it is convenient to think about RFile in terms of a key-value storage where string-based paths are
+used to refer to arbitrary objects. However, given the hierarchical nature of ROOT files, certain filesystem-like
+properties are applied to paths, for ease of use: the '/' character is treated specially as the directory separator;
+multiple '/' in a row are collapsed into one (since RFile doesn't allow directories with empty names).
+
+At the moment, RFile doesn't allow getting directories via Get, nor writing ones via Put (this may change in the
+future).
+
+## Sample usage
+Opening an RFile (for writing) and writing an object to it:
+~~~{.cpp}
+auto rfile = ROOT::RFile::Recreate("my_file.root");
+auto myObj = TH1D("h", "h", 10, 0, 1);
+rfile->Put(myObj.GetName(), myObj);
+~~~
+
+Opening an RFile (for reading) and reading an object from it:
+~~~{.cpp}
+auto rfile = ROOT::RFile::Open("my_file.root");
+auto myObj = file->Get<TH1D>("h");
+~~~
 */
 class RFile final {
    enum PutFlags {
--- a/io/io/inc/ROOT/RFile.hxx
+++ b/io/io/inc/ROOT/RFile.hxx
@@ -1,0 +1,140 @@
+/// \file ROOT/RFile.hxx
+/// \ingroup Base ROOT7
+/// \author Giacomo Parolini <giacomo.parolini@cern.ch>
+/// \date 2025-03-19
+/// \warning This is part of the ROOT 7 prototype! It will change without notice. It might trigger earthquakes. Feedback
+/// is welcome!
+
+#ifndef ROOT7_RFile
+#define ROOT7_RFile
+
+#include <ROOT/RError.hxx>
+
+#include <memory>
+#include <string_view>
+#include <typeinfo>
+
+class TFile;
+class TKey;
+
+namespace ROOT {
+namespace Experimental {
+
+class RFile;
+struct RFileKeyInfo;
+
+namespace Internal {
+
+ROOT::RLogChannel &RFileLog();
+
+} // namespace Internal
+
+/**
+\class ROOT::Experimental::RFile
+\ingroup RFile
+\brief An interface to read from, or write to, a ROOT file, as well as performing other common operations.
+
+TODO: more in-depth explanation
+*/
+class RFile final {
+   enum PutFlags {
+      kPutAllowOverwrite = 0x1,
+      kPutOverwriteKeepCycle = 0x2,
+   };
+
+   std::unique_ptr<TFile> fFile;
+
+   // Outlined to avoid including TFile.h
+   explicit RFile(std::unique_ptr<TFile> file);
+
+   /// Gets object `path` from the file and returns an **owning** pointer to it.
+   /// The caller should immediately wrap it into a unique_ptr of the type described by `type`.
+   [[nodiscard]] void *GetUntyped(std::string_view path, const std::type_info &type) const;
+
+   /// Writes `obj` to file, without taking its ownership.
+   void PutUntyped(std::string_view path, const std::type_info &type, const void *obj, std::uint32_t flags);
+
+   /// \see Put
+   template <typename T>
+   void PutInternal(std::string_view path, const T &obj, std::uint32_t flags)
+   {
+      PutUntyped(path, typeid(T), &obj, flags);
+   }
+
+   /// Given `path`, returns the TKey corresponding to the object at that path (assuming the path is fully split, i.e.
+   /// "a/b/c" always means "object 'c' inside directory 'b' inside directory 'a'").
+   /// IMPORTANT: `path` must have been validated/normalized via ValidateAndNormalizePath() (see RFile.cxx).
+   TKey *GetTKey(std::string_view path) const;
+
+public:
+   // This is arbitrary, but it's useful to avoid pathological cases
+   static constexpr int kMaxPathNesting = 1000;
+
+   ///// Factory methods /////
+
+   /// Opens the file for reading
+   static std::unique_ptr<RFile> Open(std::string_view path);
+
+   /// Opens the file for reading/writing, overwriting it if it already exists
+   static std::unique_ptr<RFile> Recreate(std::string_view path);
+
+   /// Opens the file for updating
+   static std::unique_ptr<RFile> Update(std::string_view path);
+
+   ///// Instance methods /////
+
+   // Outlined to avoid including TFile.h
+   ~RFile();
+
+   /// Retrieves an object from the file.
+   /// `path` should be a string such that `IsValidPath(path) == true`, otherwise an exception will be thrown.
+   /// See \ref ValidateAndNormalizePath() for info about valid path names.
+   /// If the object is not there returns a null pointer.
+   template <typename T>
+   std::unique_ptr<T> Get(std::string_view path) const
+   {
+      void *obj = GetUntyped(path, typeid(T));
+      return std::unique_ptr<T>(static_cast<T *>(obj));
+   }
+
+   /// Puts an object into the file.
+   /// The application retains ownership of the object.
+   /// `path` should be a string such that `IsValidPath(path) == true`, otherwise an exception will be thrown.
+   /// See \ref ValidateAndNormalizePath() for info about valid path names.
+   ///
+   /// Throws a RException if `path` already identifies a valid object or directory.
+   /// Throws a RException if the file was opened in read-only mode.
+   template <typename T>
+   void Put(std::string_view path, const T &obj)
+   {
+      PutInternal(path, obj, /* flags = */ 0);
+   }
+
+   /// Puts an object into the file, overwriting any previously-existing object at that path.
+   /// The application retains ownership of the object.
+   ///
+   /// If an object already exists at that path, it is kept as a backup cycle unless `backupPrevious` is false.
+   /// Note that even if `backupPrevious` is false, any existing cycle except the latest will be preserved.
+   ///
+   /// Throws a RException if `path` is already the path of a directory.
+   /// Throws a RException if the file was opened in read-only mode.
+   template <typename T>
+   void Overwrite(std::string_view path, const T &obj, bool backupPrevious = true)
+   {
+      std::uint32_t flags = kPutAllowOverwrite;
+      flags |= backupPrevious * kPutOverwriteKeepCycle;
+      PutInternal(path, obj, flags);
+   }
+
+   /// Writes all objects to disk with the file structure.
+   /// Returns the number of bytes written.
+   size_t Flush();
+
+   /// Flushes the RFile if needed and closes it, disallowing any further reading or writing.
+   void Close();
+};
+
+} // namespace Experimental
+} // namespace ROOT
+
+#endif
--- a/io/io/src/RFile.cxx
+++ b/io/io/src/RFile.cxx
@@ -190,6 +190,9 @@ std::unique_ptr<RFile> RFile::Open(std::string_view path)
 
    TDirectory::TContext ctx(nullptr); // XXX: probably not thread safe?
    auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "READ_WITHOUT_GLOBALREGISTRATION"));
+   if (!tfile || tfile->IsZombie())
+      throw ROOT::RException(R__FAIL("failed to open file " + std::string(path) + " for reading"));
+
    auto rfile = std::unique_ptr<RFile>(new RFile(std::move(tfile)));
    return rfile;
 }
@@ -200,6 +203,9 @@ std::unique_ptr<RFile> RFile::Update(std::string_view path)
 
    TDirectory::TContext ctx(nullptr); // XXX: probably not thread safe?
    auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "UPDATE_WITHOUT_GLOBALREGISTRATION"));
+   if (!tfile || tfile->IsZombie())
+      throw ROOT::RException(R__FAIL("failed to open file " + std::string(path) + " for updating"));
+
    auto rfile = std::unique_ptr<RFile>(new RFile(std::move(tfile)));
    return rfile;
 }
@@ -210,6 +216,9 @@ std::unique_ptr<RFile> RFile::Recreate(std::string_view path)
 
    TDirectory::TContext ctx(nullptr); // XXX: probably not thread safe?
    auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "RECREATE_WITHOUT_GLOBALREGISTRATION"));
+   if (!tfile || tfile->IsZombie())
+      throw ROOT::RException(R__FAIL("failed to open file " + std::string(path) + " for writing"));
+
    auto rfile = std::unique_ptr<RFile>(new RFile(std::move(tfile)));
    return rfile;
 }
@@ -281,7 +290,7 @@ void *RFile::GetUntyped(std::string_view pathSV, const std::type_info &type) con
       if (auto autoAddFunc = cls->GetDirectoryAutoAdd(); autoAddFunc) {
          autoAddFunc(obj, nullptr);
       }
-   } else if (key && !GetROOT()->IsBatch()) { // XXX: do we really want this warning?
+   } else if (key && !GetROOT()->IsBatch()) {
       R__LOG_WARNING(RFileLog()) << "Tried to get object '" << path << "' of type " << cls->GetName()
                                  << " but that path contains an object of type " << key->GetClassName();
    }
--- a/io/io/src/RFile.cxx
+++ b/io/io/src/RFile.cxx
@@ -1,0 +1,372 @@
+/// \file v7/src/RFile.cxx
+/// \ingroup Base ROOT7
+/// \author Giacomo Parolini <giacomo.parolini@cern.ch>
+/// \date 2025-03-19
+/// \warning This is part of the ROOT 7 prototype! It will change without notice. It might trigger earthquakes. Feedback
+/// is welcome!
+
+#include "ROOT/RFile.hxx"
+
+#include <ROOT/StringUtils.hxx>
+#include <ROOT/RError.hxx>
+
+#include <Byteswap.h>
+#include <TError.h>
+#include <TFile.h>
+#include <TKey.h>
+#include <TROOT.h>
+
+#include <algorithm>
+#include <cstring>
+
+ROOT::RLogChannel &ROOT::Experimental::Internal::RFileLog()
+{
+   static ROOT::RLogChannel sLog("ROOT.File");
+   return sLog;
+}
+
+using ROOT::Experimental::RFile;
+using ROOT::Experimental::Internal::RFileLog;
+
+static void CheckExtension(std::string_view path)
+{
+   if (ROOT::EndsWith(path, ".xml")) {
+      throw ROOT::RException(R__FAIL("ROOT::RFile doesn't support XML files."));
+   }
+
+   if (!ROOT::EndsWith(path, ".root")) {
+      R__LOG_WARNING(RFileLog()) << "ROOT::RFile only supports ROOT files. The preferred file extension is \".root\"";
+   }
+}
+
+namespace {
+enum class ENameCycleError {
+   kNoError,
+   kAnyCycle,
+   kInvalidSyntax,
+   kCycleTooLarge,
+   kNameEmpty,
+   kCOUNT
+};
+
+struct RNameCycleResult {
+   std::string fName;
+   std::optional<std::int16_t> fCycle;
+   ENameCycleError fError;
+};
+} // namespace
+
+static const char *ToString(ENameCycleError err)
+{
+   static const char *const kErrorStr[] = {"", "", "invalid syntax", "cycle is too large", "name is empty"};
+   static_assert(std::size(kErrorStr) == static_cast<std::size_t>(ENameCycleError::kCOUNT));
+   return kErrorStr[static_cast<std::size_t>(err)];
+}
+
+static ENameCycleError DecodeNumericCycle(const char *str, std::optional<std::int16_t> &out)
+{
+   uint32_t res = 0;
+   do {
+      if (!isdigit(*str))
+         return ENameCycleError::kInvalidSyntax;
+      if (res * 10 > std::numeric_limits<std::int16_t>::max())
+         return ENameCycleError::kCycleTooLarge;
+      res *= 10;
+      res += *str - '0';
+   } while (*++str);
+
+   assert(res < std::numeric_limits<std::int16_t>::max());
+   out = static_cast<std::int16_t>(res);
+
+   return ENameCycleError::kNoError;
+}
+
+static RNameCycleResult DecodeNameCycle(std::string_view nameCycleRaw)
+{
+   RNameCycleResult result{};
+
+   if (nameCycleRaw.empty())
+      return result;
+
+   // Scan the string to find the name length and the semicolon
+   std::size_t semicolonIdx = nameCycleRaw.find_first_of(';');
+
+   if (semicolonIdx == 0) {
+      result.fError = ENameCycleError::kNameEmpty;
+      return result;
+   }
+
+   // Verify that we have at most one ';'
+   if (nameCycleRaw.substr(semicolonIdx + 1).find_first_of(';') != std::string_view::npos) {
+      result.fError = ENameCycleError::kInvalidSyntax;
+      return result;
+   }
+
+   result.fName = nameCycleRaw.substr(0, semicolonIdx);
+   if (semicolonIdx < std::string_view::npos) {
+      if (semicolonIdx == nameCycleRaw.length() - 1 && nameCycleRaw[semicolonIdx] == '*')
+         result.fError = ENameCycleError::kAnyCycle;
+      else
+         result.fError = DecodeNumericCycle(nameCycleRaw.substr(semicolonIdx + 1).data(), result.fCycle);
+   }
+
+   return result;
+}
+
+/// This function first validates, then normalizes the given path in place.
+///
+/// Returns an empty string if `path` is a suitable path to store an object into a RFile,
+/// otherwise returns a description of why that is not the case.
+///
+/// A valid object path must:
+///   - not be empty
+///   - not contain the character '.'
+///   - not contain ASCII control characters or whitespace characters (including tab or newline).
+///   - not contain more than RFile::kMaxPathNesting path fragments (i.e. more than RFile::kMaxPathNesting - 1 '/')
+///   - not end with a '/'
+///
+/// In addition, when *writing* an object to RFile, the character ';' is also banned.
+///
+/// Passing an invalid path to Put will cause it to throw an exception, and
+/// passing an invalid path to Get will always return nullptr.
+///
+/// If required, `path` is modified to make its hierarchy-related meaning consistent. This entails:
+/// - combining any consecutive '/' into a single one;
+/// - stripping any leading '/'.
+///
+static std::string ValidateAndNormalizePath(std::string &path)
+{
+   ////// First, validate path.
+
+   if (path.empty())
+      return "path cannot be empty";
+
+   if (path.back() == '/')
+      return "path cannot end with a '/'";
+
+   bool valid = true;
+   for (char ch : path) {
+      // Disallow control characters, tabs, newlines, whitespace and dot.
+      // NOTE: not short-circuiting or early returning to enable loop vectorization.
+      valid &= !(ch < 33 || ch == '.');
+   }
+   if (!valid)
+      return "path cannot contain control characters, whitespaces or dots";
+
+   //// Path is valid so far, normalize it.
+
+   // Strip all leading '/'
+   {
+      auto nToStrip = 0u;
+      const auto len = path.length();
+      while (nToStrip < len && path[nToStrip] == '/')
+         ++nToStrip;
+
+      if (nToStrip > 0)
+         path.erase(0, nToStrip);
+   }
+
+   // Remove duplicate consecutive '/'
+   const auto it = std::unique(path.begin(), path.end(), [](char a, char b) { return (a == '/' && b == '/'); });
+   path.erase(it, path.end());
+
+   //// After the path has been normalized, check the nesting level by counting how many slashes it contains.
+   const auto nesting = std::count(path.begin(), path.end(), '/');
+   if (nesting > RFile::kMaxPathNesting)
+      return "pathView contains too many levels of nesting";
+
+   return "";
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////////////
+
+RFile::RFile(std::unique_ptr<TFile> file) : fFile(std::move(file)) {}
+
+RFile::~RFile() = default;
+
+std::unique_ptr<RFile> RFile::Open(std::string_view path)
+{
+   CheckExtension(path);
+
+   TDirectory::TContext ctx(nullptr); // XXX: probably not thread safe?
+   auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "READ_WITHOUT_GLOBALREGISTRATION"));
+   auto rfile = std::unique_ptr<RFile>(new RFile(std::move(tfile)));
+   return rfile;
+}
+
+std::unique_ptr<RFile> RFile::Update(std::string_view path)
+{
+   CheckExtension(path);
+
+   TDirectory::TContext ctx(nullptr); // XXX: probably not thread safe?
+   auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "UPDATE_WITHOUT_GLOBALREGISTRATION"));
+   auto rfile = std::unique_ptr<RFile>(new RFile(std::move(tfile)));
+   return rfile;
+}
+
+std::unique_ptr<RFile> RFile::Recreate(std::string_view path)
+{
+   CheckExtension(path);
+
+   TDirectory::TContext ctx(nullptr); // XXX: probably not thread safe?
+   auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "RECREATE_WITHOUT_GLOBALREGISTRATION"));
+   auto rfile = std::unique_ptr<RFile>(new RFile(std::move(tfile)));
+   return rfile;
+}
+
+TKey *RFile::GetTKey(std::string_view path) const
+{
+   // In RFile, differently from TFile, when dealing with a path like "a/b/c", we always consider it to mean
+   // "object 'c' in subdirectory 'b' of directory 'a'". We don't try to get any other of the possible combinations,
+   // including the object called "a/b/c".
+   std::string fullPath = std::string(path);
+   char *dirName = fullPath.data();
+   char *restOfPath = strchr(dirName, '/');
+   TDirectory *dir = fFile.get();
+   while (restOfPath) {
+      // Truncate `dirName` to the position of this '/'.
+      *restOfPath = 0;
+      ++restOfPath;
+      // `restOfPath` should always be a non-empty string unless `path` ends with '/' (which it shouldn't, as we are
+      // supposed to have normalized it before calling this function).
+      assert(*restOfPath);
+
+      dir = dir->GetDirectory(dirName);
+      if (!dir)
+         return nullptr;
+
+      dirName = restOfPath;
+      restOfPath = strchr(restOfPath, '/');
+   }
+   // NOTE: after this loop `dirName` contains the base name of the object.
+
+   // Get the leaf object from the innermost directory.
+   TKey *key = dir->FindKey(dirName);
+   if (key) {
+      // For some reason, FindKey will not return nullptr if we asked for a specific cycle and that cycle
+      // doesn't exist. It will instead return any key whose cycle is *at most* the requested one.
+      // This is very confusing, so in RFile we actually return null if the requested cycle is not there.
+      RNameCycleResult res = DecodeNameCycle(dirName);
+      if (res.fError != ENameCycleError::kAnyCycle) {
+         if (res.fError != ENameCycleError::kNoError) {
+            R__LOG_ERROR(RFileLog()) << "error decoding namecycle '" << dirName << "': " << ToString(res.fError);
+            key = nullptr;
+         } else if (res.fCycle && *res.fCycle != key->GetCycle()) {
+            key = nullptr;
+         }
+      }
+   }
+   return key;
+}
+
+void *RFile::GetUntyped(std::string_view pathSV, const std::type_info &type) const
+{
+   if (!fFile)
+      throw ROOT::RException(R__FAIL("File has been closed"));
+
+   std::string path{pathSV};
+
+   const TClass *cls = TClass::GetClass(type);
+   if (!cls)
+      throw ROOT::RException(R__FAIL(std::string("Could not determine type of object ") + path));
+
+   if (auto err = ValidateAndNormalizePath(path); !err.empty())
+      throw RException(R__FAIL("Invalid object path '" + path + "': " + err));
+
+   TKey *key = GetTKey(path);
+   void *obj = key ? key->ReadObjectAny(cls) : nullptr;
+
+   if (obj) {
+      // Disavow any ownership on `obj`
+      if (auto autoAddFunc = cls->GetDirectoryAutoAdd(); autoAddFunc) {
+         autoAddFunc(obj, nullptr);
+      }
+   } else if (key && !GetROOT()->IsBatch()) { // XXX: do we really want this warning?
+      R__LOG_WARNING(RFileLog()) << "Tried to get object '" << path << "' of type " << cls->GetName()
+                                 << " but that path contains an object of type " << key->GetClassName();
+   }
+
+   return obj;
+}
+
+void RFile::PutUntyped(std::string_view pathSV, const std::type_info &type, const void *obj, std::uint32_t flags)
+{
+   const TClass *cls = TClass::GetClass(type);
+   if (!cls)
+      throw ROOT::RException(R__FAIL(std::string("Could not determine type of object ") + std::string(pathSV)));
+
+   std::string path{pathSV};
+   if (auto err = ValidateAndNormalizePath(path); !err.empty())
+      throw RException(R__FAIL("Invalid object path '" + path + "': " + err));
+
+   if (path.find_first_of(';') != std::string_view::npos) {
+      throw RException(
+         R__FAIL("Invalid object path '" + path +
+                 "': character ';' is used to specify an object cycle, which only makes sense when reading."));
+   }
+
+   if (!fFile)
+      throw ROOT::RException(R__FAIL("File has been closed"));
+
+   if (!fFile->IsWritable())
+      throw ROOT::RException(R__FAIL("File is not writable"));
+
+   // If `path` refers to a subdirectory, make sure we always write in an actual TDirectory,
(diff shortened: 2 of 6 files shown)


.github/workflows/root-ci.yml
```diff
@@ -107,8 +107,10 @@ jobs:
             arch: ARM64
             overrides: ["CMAKE_CXX_STANDARD=23"]
           - platform: mac26
+            overrides: ["rfile=ON"]
```
hageboeck: I would suggest to put these in
https://github.com/root-project/root/blob/master/.github/workflows/root-ci-config/buildconfig/mac26.txt

and similar as "the default". I wouldn't consider the addition of rfile as an "override".

io/io/inc/ROOT/RFile.hxx
hageboeck, line 1: After yesterday's discussion, is it wise to put this in `io/v7`? Version numbers in file names or configs will just create headaches. How about putting it in `io/rfile`, where it can remain also in the future?

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,131 @@
...
+class RFile;
+struct RFileKeyInfo;
+
+namespace Internal {
+
+ROOT::RLogChannel &RFileLog();
+
+} // namespace Internal
+
+class RFile final {
```
hageboeck, line 98: `final` will prevent users from inheriting and possibly overriding. Is this the intention?

hageboeck: A larger docstring for the class documentation would be desirable, but it
- can be added later
- and could be placed in the `.cxx`, so you don't have to recompile everything when updating the docs.

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,131 @@
...
+
+   std::unique_ptr<TFile> fFile;
+
+   explicit RFile(std::unique_ptr<TFile> file) : fFile(std::move(file)) {}
+
+   /// Gets object `path` from the file and returns an **owning** pointer to it.
+   /// The caller should immediately wrap it into a unique_ptr of the type described by `type`.
+   [[nodiscard]] void *GetUntyped(std::string_view path, const TClass &type) const;
+
+   /// Writes `obj` to file, without taking its ownership.
```
hageboeck, line 113: Maybe this makes it clearer?
```suggestion
   /// Writes a copy of `obj` to file, without taking its ownership.
```

hageboeck: An explanation of `type` and `flags` would be great, too.

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,131 @@
...
+   {
+      const TClass *cls = TClass::GetClass(typeid(T));
+      if (!cls) {
+         throw ROOT::RException(R__FAIL(std::string("Could not determine type of object ") + std::string(path)));
+      }
+      void *obj = GetUntyped(path, *cls);
+      return std::unique_ptr<T>(static_cast<T *>(obj));
+   }
+
+   /// Puts an object into the file.
```
hageboeck, line 164:
```suggestion
   /// Puts a copy of an object into the file.
```

io/v7/rfile/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,131 @@
...
+      }
+      void *obj = GetUntyped(path, *cls);
+      return std::unique_ptr<T>(static_cast<T *>(obj));
+   }
+
+   /// Puts an object into the file.
+   /// The application retains ownership of the object.
+   /// `path` should be a string such that `IsValidPath(path) == true`.
+   ///
+   /// Throws a RException if a directory already exists at `path`.
```
hageboeck: I understand what this means, but it sounds like this is forbidden:
```
f.Put("a/b/obj1", ...);
f.Put("a/b/obj2", ...);
```

io/io/inc/ROOT/RFile.hxx
hageboeck: Sorry, I see now that this is private, so the request has low priority.

io/v7/rfile/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,131 @@
...
+/// \date 2025-03-19
+/// \warning This is part of the ROOT 7 prototype! It will change without notice. It might trigger earthquakes. Feedback
+/// is welcome!
+
+#ifndef ROOT7_RFile
+#define ROOT7_RFile
+
+#include <ROOT/RError.hxx>
+
+#include <TFile.h>
```
hageboeck: It looks to me that if you outline the destructor of RFile, you could get away with a forward declaration of TFile, only requiring TClass?

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,131 @@
...
+   /// Throws a RException if a directory already exists at `path`.
+   /// Throws a RException if an object already exists at `path`.
+   /// Throws a RException if the file was opened in read-only mode.
+   template <typename T>
+   void Put(std::string_view path, const T &obj)
+   {
+      PutInternal(path, obj, /* flags = */ 0);
+   }
+
+   /// Puts an object into the file, overwriting any previously-existing object at that path.
```
hageboeck, line 177:
```suggestion
   /// Puts a copy of an object into the file, overwriting any previously-existing object at that path.
```

io/v7/rfile/inc/ROOT/RFile.hxx
hageboeck: Doxygen should add special highlighting if you used:
```c++
/// \throw RException <description>
/// \throw OtherType <description>

```

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,131 @@
...
+   void Overwrite(std::string_view path, const T &obj, bool backupPrevious = true)
+   {
+      std::uint32_t flags = kPutAllowOverwrite;
+      flags |= backupPrevious * kPutOverwriteKeepCycle;
+      PutInternal(path, obj, flags);
+   }
+
+   /// Writes all objects to disk with the file structure.
+   /// Returns the number of bytes written.
+   size_t Write();
```
hageboeck: Does the number of bytes written make sense here if the other write functions return `void`?

io/v7/rfile/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,131 @@
...
+      std::uint32_t flags = kPutAllowOverwrite;
+      flags |= backupPrevious * kPutOverwriteKeepCycle;
+      PutInternal(path, obj, flags);
+   }
+
+   /// Writes all objects to disk with the file structure.
+   /// Returns the number of bytes written.
+   size_t Write();
+
+   /// Closes the RFile, disallowing any further reading or writing.
```
hageboeck: I assume that this will trigger `Write()`, though?

io/v7/rfile/test/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+ROOT_ADD_GTEST(rfile rfile.cxx INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/tree/ntuple/v7/inc LIBRARIES Core Hist RIO ROOTFile ROOTNTuple)
```
hageboeck: Hacking the include path in this way is maybe OK for the test, but not that great in the long term. What if it gets moved?

io/v7/rfile/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+HEADERS
+  ROOT/RFile.hxx
+SOURCES
+  src/RFile.cxx
+ LINKDEF
+  LinkDef.h
+DEPENDENCIES
+  RIO
+  Tree
+  Hist
```
hageboeck: Can we break the link to `Hist` (see other comments)?

[TClass should be able to achieve unsetting the directory using function pointers.](https://github.com/root-project/root/blob/92d8e56aaa2adaf20fc2fb1a8616321620e68614/core/meta/inc/TClass.h#L476)

hageboeck: You hopefully only need it for the tests. But if you do, you could place these in roottest, so the `io/rfile` tests can run without linking to `Hist`.

io/v7/rfile/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+#include "ROOT/RFile.hxx"
+
+#include <ROOT/StringUtils.hxx>
+#include <ROOT/RError.hxx>
+
+#include <Byteswap.h>
+#include <TError.h>
+#include <TTree.h>
+#include <TGraph2D.h>
+#include <TH1.h>
```
hageboeck: You might be able to remove these, see below.

io/v7/rfile/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+
+   if (obj) {
+      // Disavow any ownership on `obj`
+      // NOTE: all these objects inherit from TObject as their first parent, so we can use static_cast.
+      if (type.InheritsFrom("TTree"))
+         static_cast<TTree *>(obj)->SetDirectory(nullptr);
+      else if (type.InheritsFrom("TH1"))
+         static_cast<TH1 *>(obj)->SetDirectory(nullptr);
+      else if (type.InheritsFrom("TGraph2D"))
+         static_cast<TGraph2D *>(obj)->SetDirectory(nullptr);
```
hageboeck: I *think* you could use [TClass::GetDirectoryAutoAdd()](https://github.com/root-project/root/blob/92d8e56aaa2adaf20fc2fb1a8616321620e68614/core/meta/inc/TClass.h#L476), and you should be able to set the directories without depending on the corresponding headers or linking to the libs.
More or less:
```c++
if (auto autoAddFunc = type.GetDirectoryAutoAdd(); autoAddFunc) {
   autoAddFunc(obj, nullptr);
}
```

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+#include <TError.h>
+#include <TTree.h>
+#include <TGraph2D.h>
+#include <TH1.h>
+#include <TKey.h>
+#include <TROOT.h>
+
+#include <cstring>
+
+ROOT::RLogChannel &ROOT::Experimental::Internal::RFileLog()
```
hageboeck, line 22: Could this simply be `ROOT::Internal`? We have the freedom to do whatever we want with internal, so I don't see a need to have it in experimental only to move it out later.

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+{
+   static ROOT::RLogChannel sLog("ROOT.RFile");
+   return sLog;
+}
+
+using ROOT::Experimental::RFile;
+using ROOT::Experimental::Internal::RFileLog;
+
+static void ThrowIfExtensionIsNotRoot(std::string_view path)
```
hageboeck: Move to anonymous namespace?

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+};
+
+struct RNameCycleResult {
+   std::string fName;
+   std::optional<std::int16_t> fCycle;
+   ENameCycleError fError;
+};
+} // namespace
+
+static const char *ToString(ENameCycleError err)
```
hageboeck, line 59: Maybe also to anonymous namespace, as the ones below.

io/io/src/RFile.cxx
hageboeck: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+/// Returns an empty string if `path` is a suitable path to store an object into a RFile,
+/// otherwise returns a description of why that is not the case.
+///
+/// A valid object path must:
+///   - not be empty
+///   - not contain the character '.'
+///   - not contain ASCII control characters or whitespace characters (including tab or newline).
+///   - not contain more than RFile::kMaxPathNesting path fragments (i.e. more than RFile::kMaxPathNesting - 1 '/')
+///
+/// In addition, when *writing* an object to RFile, the character ';' is also banned.
```
hageboeck, line 128: This is a very useful description. Could this be copied also to the description of `Get` / `Put` or the class description?

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+
+   //// After the path has been normalized, check the nesting level by counting how many slashes it contains.
+   const auto nesting = std::count(path.begin(), path.end(), '/');
+   if (nesting > RFile::kMaxPathNesting)
+      return "pathView contains too many levels of nesting";
+
+   return "";
+}
+
+/////////////////////////////////////////////////////////////////////////////////////////////////
```
hageboeck, line 181: Consider extending the documentation, (possibly in a different PR if desired)

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+
+std::unique_ptr<RFile> RFile::Open(std::string_view path)
+{
+   ThrowIfExtensionIsNotRoot(path);
+
+   auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "READ_WITHOUT_GLOBALREGISTRATION"));
+   auto rfile = std::unique_ptr<RFile>(new RFile(std::move(tfile)));
+   return rfile;
```
hageboeck, line 197: Lazy way. 🙂 
```suggestion
   return std::make_unique<RFile>(std::move(tfile));
```

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+}
+
+std::unique_ptr<RFile> RFile::OpenForUpdate(std::string_view path)
```
hageboeck: `///`

io/v7/rfile/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+   char *restOfPath = strchr(dirName, '/');
+   TDirectory *dir = fFile.get();
+   while (restOfPath) {
+      // Truncate `dirName` to the position of this '/'.
+      *restOfPath = 0;
+      ++restOfPath;
+
+      // This can only happen if `path` ends with '/' (which it shouldn't).
+      if (R__unlikely(!*restOfPath))
+         return nullptr;
```
hageboeck: If it shouldn't, is this a case for an assertion?

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+      // For some reason, FindKey will not return nullptr if we asked for a specific cycle and that cycle
+      // doesn't exist. It will instead return any key whose cycle is *at most* the requested one.
+      // This is very confusing, so in RFile we actually return null if the requested cycle is not there.
+      RNameCycleResult res = DecodeNameCycle(dirName);
+      if (res.fError != ENameCycleError::kAnyCycle) {
+         if (res.fError != ENameCycleError::kNoError) {
+            R__LOG_ERROR(RFileLog()) << "error decoding namecycle '" << dirName << "': " << ToString(res.fError);
+            key = nullptr;
+         } else if (res.fCycle && *res.fCycle != key->GetCycle()) {
+            key = nullptr;
```
hageboeck, line 264: I'm unsure, so therefore the question: Is `key` leaking now or is it owned by the file?

io/v7/rfile/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+std::unique_ptr<RFile> RFile::Recreate(std::string_view path)
+{
+   ThrowIfExtensionIsNotRoot(path);
+
+   auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "RECREATE_WITHOUT_GLOBALREGISTRATION"));
+   auto rfile = std::unique_ptr<RFile>(new RFile(std::move(tfile)));
+   return rfile;
+}
+
+TKey *RFile::GetTKey(const char *path) const
```
hageboeck: Who owns the returned key?

Why does this one take `const char*`, whereas the rest takes string_view?

io/v7/rfile/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+
+   if (auto err = ValidateAndNormalizePath(path); !err.empty()) {
+      R__LOG_ERROR(Internal::RFileLog()) << "Invalid object path '" << path << "': " << err;
+      return nullptr;
+   }
+
+   if (!fFile)
+      throw ROOT::RException(R__FAIL("File has been closed"));
+
+   TKey *key = GetTKey(path.c_str());
```
hageboeck: By now, I get the impression that the key is owned by the file. Just adding this comment to confirm

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,357 @@
...
+   if (obj) {
+      // Disavow any ownership on `obj`
+      // NOTE: all these objects inherit from TObject as their first parent, so we can use static_cast.
+      if (type.InheritsFrom("TTree"))
+         static_cast<TTree *>(obj)->SetDirectory(nullptr);
+      else if (type.InheritsFrom("TH1"))
+         static_cast<TH1 *>(obj)->SetDirectory(nullptr);
+      else if (type.InheritsFrom("TGraph2D"))
+         static_cast<TGraph2D *>(obj)->SetDirectory(nullptr);
+   } else if (key && !GetROOT()->IsBatch()) { // XXX: do we really want this warning?
```
hageboeck: If this is the only warning/error that's generated, it's indeed helpful.
If `ReadObjectAny` gives more insight into what went wrong, it's maybe not necessary.

io/v7/rfile/test/rfile.cxx
```diff
@@ -0,0 +1,397 @@
+#include "gtest/gtest.h"
+#include "gmock/gmock.h"
```
hageboeck, line 2: Is it using `gmock`? I might have removed `gmock` from the automatically linked CMake targets, so you might have to add it if required.

io/v7/rfile/test/rfile.cxx
```diff
@@ -0,0 +1,397 @@
...
+
+   auto file = RFile::Open(fileGuard.GetPath());
+   auto hist = file->Get<TH1D>("hist");
+   EXPECT_TRUE(hist);
+
+   EXPECT_FALSE(file->Get<TH1D>("inexistent"));
+   EXPECT_FALSE(file->Get<TH1F>("hist"));
+   EXPECT_TRUE(file->Get<TH1>("hist"));
+
+   // We do NOT want to globally register RFiles ever.
```
hageboeck, line 66: * Maybe yes if the browser is supposed to show a list of open files. But if we decide that that's what it should do, one should maybe do this in a different list with weak pointers or so ...

io/v7/rfile/test/rfile.cxx
```diff
@@ -0,0 +1,397 @@
...
+   hist2.FillRandom("gaus", 10);
+   EXPECT_THROW(file->Put("hist", hist2), ROOT::RException);
+
+   // Try with Overwrite, should work (and preserve the old object)
+   file->Overwrite("hist", hist2);
+   {
+      auto hist = file->Get<TH1D>("hist");
+      ASSERT_TRUE(hist);
+      EXPECT_EQ(static_cast<int>(hist->GetEntries()), 10);
+
```
hageboeck, line 247: Does this hold?
```suggestion
EXPECT_EQ(*file->Get<TH1D>("hist;2"), *hist);
```

io/v7/rfile/test/rfile.cxx
```diff
@@ -0,0 +1,397 @@
...
+   }
+}
+
+#ifdef R__HAS_DAVIX
+TEST(RFile, RemoteRead)
+{
+   constexpr const char *kFileName = "http://root.cern/files/RNTuple.root";
+
+   auto file = RFile::Open(kFileName);
+   auto ntuple = file->Get<ROOT::RNTuple>("Contributors");
```
hageboeck: Can this gracefully fail with GTEST_SKIP() if one is offline?

io/v7/rfile/test/rfile.cxx
```diff
@@ -0,0 +1,397 @@
...
+      for (int runIdx = 0; runIdx < kNRuns; ++runIdx) {
+         const auto runDir = dir + "/run" + (runIdx + 1);
+
+         const auto kNHist = 10;
+         for (int i = 0; i < kNHist; ++i) {
+            const auto histName = std::string("h") + (i + 1);
+            const auto histPath = runDir + "/hists/" + histName;
+            const auto histTitle = std::string("Histogram #") + (i + 1);
+            TH1D hist(histName, histTitle, 100, -10 * (i + 1), 10 * (i + 1));
+            file->Put(histPath, hist);
```
hageboeck, line 399: I'm not sure this test increases coverage. Also, if it was covering a new case, wouldn't you at least have to check that the objects can be read / listed?

io/io/src/RFile.cxx
silverweed: I don't like using anonymous namespaces as it obscures whether the function is static or not when it gets larger than about a page. I only use it for types because there is no alternative for them.

io/v7/rfile/src/RFile.cxx
silverweed: The key is owned by `fFile` (or one of its subdirectory) and we're getting an unowning pointer. I felt it's redundant to specify given it's implied by the signature, but I can specify it (I'm just afraid that if I do then I'm somehow implying that a raw pointer doesn't necessarily imply absence of ownership).
For the `const char*`, no good reason - I'll change that.

io/io/src/RFile.cxx
silverweed: `GetKey` returns an unowning pointer, so no leaking is happening

io/io/inc/ROOT/RFile.hxx
silverweed: Which other write functions? You mean `Put`?
I think it makes sense, also considering that `TFile::Write()` returns the information, so it makes sense to use it.

io/io/inc/ROOT/RFile.hxx
silverweed: > `final` will prevent users from inheriting and possibly overriding. Is this the intention?

Considering we're not defining any virtual method (including the destructor) I don't see how the user could effectively inherit this class anyway. If we change our mind later we can just remove the `final`, but I consider it a good default.

io/io/inc/ROOT/RFile.hxx
silverweed: Not sure about the "a copy of" part; the object is not really copied in memory, it's just serialized into the file. At least, that's the mental model that the both us and the user should have, regardless of how TFile actually implements it

io/v7/rfile/inc/ROOT/RFile.hxx
silverweed: I'm not sure the doxygen stuff fits well here as it looks like it's meant to be used with different exception types; here we only have one exception type that may be thrown for a number of different cases (plus it reads slightly better if you're looking at the code rather than the doxygen doc)

io/io/inc/ROOT/RFile.hxx
silverweed: See comment above

io/io/src/RFile.cxx
silverweed: I don't see a strong reason to prefer one or the other, but having all in Experimental makes a bit more sense to me...that's what we did for RNTuple as well.

io/io/src/RFile.cxx
silverweed: Yes, but then we would have to remember to update all the places if we change these rules. It might still be worth it; opinions?

io/io/src/RFile.cxx
silverweed: ?

io/io/inc/ROOT/RFile.hxx
silverweed: In the end I just put it in `io/io` to avoid CMake headaches. Since it's always built anyway it doesn't make much of a difference in my opinion.

io/io/src/RFile.cxx
silverweed: `make_unique` can't be used with private ctors; as for the local variable, it's a personal preference motivated by ease of debugging

io/v7/rfile/test/rfile.cxx
silverweed: I don't think the equality operator is defined for TH1..

io/v7/rfile/test/rfile.cxx
silverweed: It is apparently needed by `testing::HasSubstr`

io/v7/rfile/test/rfile.cxx
silverweed: I don't know...how can we reliably detect whether we're offline? Do we do this in other tests?

io/io/test/CMakeLists.txt
```diff
@@ -18,6 +18,8 @@ if(uring AND NOT DEFINED ENV{ROOTTEST_IGNORE_URING})
   ROOT_ADD_GTEST(RIoUring RIoUring.cxx LIBRARIES RIO)
 endif()
 
+ROOT_ADD_GTEST(rfile rfile.cxx LIBRARIES RIO Hist ROOTNTuple)
```
hahnjo: In principle, `ROOTNTuple` is a higher level in terms of layering than `RIO`, not sure if it's a good idea to depend on it (even for a test)...

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,147 @@
...
+
+   /// \see Put
+   template <typename T>
+   void PutInternal(std::string_view path, const T &obj, std::uint32_t flags)
+   {
+      const TClass *cls = TClass::GetClass(typeid(T));
+      if (!cls) {
+         throw ROOT::RException(R__FAIL(std::string("Could not determine type of object ") + std::string(path)));
+      }
+      PutUntyped(path, *cls, &obj, flags);
```
hahnjo: In principle, we could even avoid a dependency of the other on `TClass` by only forwarding the result of `typeid(T)` and call `TClass::GetClass` only in `RFile.cxx`. Not sure if this buys us anything...

hahnjo (commented):
Two small comments, but in general I'll leave review to others...

io/io/test/CMakeLists.txt
silverweed: Isn't that the case also for `Hist`?

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,138 @@
...
+   ///// Factory methods /////
+
+   /// Opens the file for reading
+   static std::unique_ptr<RFile> Open(std::string_view path);
+
+   /// Opens the file for reading/writing, overwriting it if it already exists
+   static std::unique_ptr<RFile> Recreate(std::string_view path);
+
+   /// Opens the file for updating
+   static std::unique_ptr<RFile> OpenForUpdate(std::string_view path);
```
jblomer:
```suggestion
   static std::unique_ptr<RFile> Update(std::string_view path);
```

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,138 @@
...
+   void Overwrite(std::string_view path, const T &obj, bool backupPrevious = true)
+   {
+      std::uint32_t flags = kPutAllowOverwrite;
+      flags |= backupPrevious * kPutOverwriteKeepCycle;
+      PutInternal(path, obj, flags);
+   }
+
+   /// Writes all objects to disk with the file structure.
+   /// Returns the number of bytes written.
+   size_t Write();
```
jblomer:
```suggestion
   size_t Flush();
```
or
```suggestion
   size_t Commit();
```

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,364 @@
...
+ROOT::RLogChannel &ROOT::Experimental::Internal::RFileLog()
+{
+   static ROOT::RLogChannel sLog("ROOT.RFile");
+   return sLog;
+}
+
+using ROOT::Experimental::RFile;
+using ROOT::Experimental::Internal::RFileLog;
+
+static void ThrowIfExtensionIsNotRoot(std::string_view path)
```
jblomer: Do we really want to enforce the restriction on the file extension? I'd leave it up to the users. If they need to backup their data to "data.root.1", why forbid it...

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,364 @@
...
+} // namespace
+
+static const char *ToString(ENameCycleError err)
+{
+   static const char *const kErrorStr[] = {"", "", "invalid syntax", "cycle is too large", "name is empty"};
+   static_assert(std::size(kErrorStr) == static_cast<std::size_t>(ENameCycleError::kCOUNT));
+   return kErrorStr[static_cast<std::size_t>(err)];
+}
+
+static ENameCycleError DecodeNumericCycle(const char *str, std::optional<std::int16_t> &out)
```
jblomer, line 66:
```suggestion
static ENameCycleError DecodeNumericCycle(const char *str, std::int16_t &out)
```

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,364 @@
...
+
+RFile::~RFile() = default;
+
+std::unique_ptr<RFile> RFile::Open(std::string_view path)
+{
+   ThrowIfExtensionIsNotRoot(path);
+
+   auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "READ_WITHOUT_GLOBALREGISTRATION"));
+   auto rfile = std::unique_ptr<RFile>(new RFile(std::move(tfile)));
+   return rfile;
```
jblomer, line 197:
```suggestion
   return std::unique_ptr<RFile>(new RFile(std::move(tfile)));
```

io/io/test/rfile.cxx
```diff
@@ -0,0 +1,395 @@
...
+   try {
+      file->Put("a/b/c", s);
+      FAIL() << "creating a directory over a leaf path should fail.";
+   } catch (const ROOT::RException &ex) {
+      EXPECT_THAT(ex.what(), testing::HasSubstr("'a/b'"));
+      EXPECT_THAT(ex.what(), testing::HasSubstr("name already taken"));
+   }
+}
+
+#ifdef R__HAS_DAVIX
```
jblomer, line 370: HTTP capabilities may be there even without Davix (e.g., TWebFile).

jblomer (commented):
Very nice take-off!

io/io/src/RFile.cxx
silverweed: The idea was to prevent the user from trying opening files with unsupported extensions like `.xml` (which is something you could do with `TFile::Open`), but maybe it's not a good enough reason to prevent valid use cases like you say...

io/io/test/rfile.cxx
silverweed: Is there a reliable way to tell if this has a chance of succeeding and skipping the test if it's not the case?

io/io/src/RFile.cxx
silverweed: I would rather keep the lines separated for debuggability reasons

io/io/src/RFile.cxx
silverweed: If I change this signature I also have to change the caller site (and a couple of other places - basically make `RNameCycleResult::fCycle` not optional.
Is there a particular reason to want that?

io/v7/rfile/inc/ROOT/RFile.hxx
hageboeck: Granted the doxygen part looks worse in the code. An doxygen, I believe it would look better though, see e.g. here:
https://root.cern.ch/doc/master/classRooAbsCollection.html#aa4a63f734e815ff740d6d65dbe2c081d

To list multiple conditions where the same exception might be thrown, one could probably use
```
/// \throw RExecption 
/// - If A
/// - If B, ...
/// - If C, ...
```
But I have to admit that I didn't try that yet.

io/io/inc/ROOT/RFile.hxx
hageboeck: Yes, I meant `Put` for example. I don't understand the use of the "bytes written" return value, have always been wondering about the reason for it, but I will accept that it doesn't do any harm.

io/io/src/RFile.cxx
hageboeck: Updating / syncing it is indeed cumbersome. A way around this would be to add `see \ref ValidateAndNormalizePath() for info about valid path names` to `Get` / `Put`

io/v7/rfile/test/rfile.cxx
hageboeck: I overestimated the feature set of TH1 ... 😅

io/v7/rfile/test/rfile.cxx
hageboeck: If that's in use, it's indeed required. 👍

io/io/src/RFile.cxx
hageboeck: That was supposed to mean that a docstring might be helpful at some point for users.

io/v7/rfile/test/rfile.cxx
hageboeck: Not to my knowledge. One could prevent one type of spurious failure by pinging:
```
auto file = RFile::Open(kFileName);
if (!file && ! <ping http://root.cern>) GTEST_SKIP();
```
But maybe that's overdoing it.

io/io/src/RFile.cxx
silverweed: Actually, a (small) docstring is already there (in the header file)

io/io/src/RFile.cxx
silverweed: ReadObjectAny silently returns `nullptr`, so we probably want to keep this warning.

io/io/src/RFile.cxx
silverweed: The problem with that is that `Get`/`Put` are only defined in the header, and `ValidateAndNormalizePath()` is only defined in the cxx...not sure how to solve this in a nice way.

io/io/src/RFile.cxx
jblomer: Yeah, I didn't follow the chain properly. Fine then.

io/io/src/RFile.cxx
hageboeck: Where these are defined shouldn't be an issue for doxygen. Adding this line to the `Get / Put` docstrings should solve it:
```
/// See \ref ValidateAndNormalizePath() for info about valid path names.
```

io/v7/rfile/test/rfile.cxx
silverweed: We can talk about this when the problem emerges I think

io/io/test/rfile.cxx
silverweed: Added a TODO for now

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,199 @@
...
+
+   ///// Factory methods /////
+
+   /// Opens the file for reading
+   static std::unique_ptr<RFile> Open(std::string_view path);
+
+   /// Opens the file for reading/writing, overwriting it if it already exists
+   static std::unique_ptr<RFile> Recreate(std::string_view path);
+
+   /// Opens the file for updating
```
jblomer: Documentation should say if the file is created if it doesn't exist yet or if it has to be already present.

jblomer (approved):

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,199 @@
...
+   /// Throws a RException if the file was opened in read-only mode.
+   template <typename T>
+   void Overwrite(std::string_view path, const T &obj, bool backupPrevious = true)
+   {
+      std::uint32_t flags = kPutAllowOverwrite;
+      flags |= backupPrevious * kPutOverwriteKeepCycle;
+      PutInternal(path, obj, flags);
+   }
+
+   /// Writes all objects to disk with the file structure.
```
hageboeck:
```suggestion
   /// Writes all objects and the file structure to disk.
```

io/io/inc/ROOT/RFile.hxx
```diff
@@ -34,7 +34,66 @@ ROOT::RLogChannel &RFileLog();
...
+
+When getting an object from the file (via RFile::Get) you get back a unique copy of the object. Calling `Get` on the
+same object twice produces two independent clones of the object. The ownership over that object is solely on the caller
+and not shared with the RFile. Therefore, the object will remain valid after closing or destroying the RFile that
+generated it. This also means that any modification done to the object are **not** reflected to the file automatically:
+to update the object in the file you need to write it again (via RFile::Overwrite).
+
+RFile::Put and RFile::Overwrite are the way to write objects to the file. Both methods take a const reference to the
+object to write and don't change the ownership of the object in any way. Calling Put or Overwrite doesn't guarantee that
+the object is immediately written to the underlying storage: to ensure that, you need to call RFile::Flush (or close the
```
hageboeck, line 63: I find this surprising. Doesn't `TFile::WriteObjectAny` write objects immediately? I know that it doesn't write the footer, but couldn't one theoretically recover objects that have been written with `WriteObject`?

io/io/inc/ROOT/RFile.hxx
silverweed: Added a commit to clarify the behavior and to throw exceptions when TFile::Open returns an invalid file

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,199 @@
...
+file).
+
+## Directories
+
+Differently from TFile, the RFile class itself is not also a "directory". In fact, there is no RDirectory class at all.
```
hageboeck, line 68: I know that inheritance can be expressed as an "is-a" relationship, but maybe this reads a bit smoother:
```suggestion
Differently from TFile, the RFile class does not act as a directory. In fact, there is no RDirectory class at all.
```

hageboeck (approved):
👍 Just a few small things below.

io/io/inc/ROOT/RFile.hxx
hageboeck: As just discussed, there might be a TFile-level write cache, so the statement is correct.

io/io/inc/ROOT/RFile.hxx
silverweed: I don't know; for me it's a bit clearer to say that it's (taxonomically and semantically) not a directory, rather than it doesn't "act as a directory" (what does that mean exactly? Considering it has the capability of "storing objects under itself" you could argue that it does somewhat act as a directory...)

hageboeck: You are probably right. I was thinking of the less experienced users, but maybe at this stage, we should actually aim at the experienced users, because these are most likely to follow the developments. OK to resolve this.

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,204 @@
...
+   template <typename T>
+   void PutInternal(std::string_view path, const T &obj, std::uint32_t flags)
+   {
+      PutUntyped(path, typeid(T), &obj, flags);
+   }
+
+   /// Given `path`, returns the TKey corresponding to the object at that path (assuming the path is fully split, i.e.
+   /// "a/b/c" always means "object 'c' inside directory 'b' inside directory 'a'").
+   /// IMPORTANT: `path` must have been validated/normalized via ValidateAndNormalizePath() (see RFile.cxx).
+   TKey *GetTKey(std::string_view path) const;
```
pcanal, line 126: Consider:
```suggestion
   TKey *GetKey(std::string_view path) const;
```

io/io/inc/ROOT/RFile.hxx
pcanal: In the general case, it can be used to test that what was written is what (or more exactly the amount that) was expected to be written.

io/io/test/CMakeLists.txt
pcanal: Those test should probably be moved to roottest

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,381 @@
...
+
+   if (obj) {
+      // Disavow any ownership on `obj`
+      if (auto autoAddFunc = cls->GetDirectoryAutoAdd(); autoAddFunc) {
+         autoAddFunc(obj, nullptr);
+      }
+   } else if (key && !GetROOT()->IsBatch()) {
+      R__LOG_WARNING(RFileLog()) << "Tried to get object '" << path << "' of type " << cls->GetName()
+                                 << " but that path contains an object of type " << key->GetClassName();
+   }
```
pcanal, line 296: This is debatable.   What if the user wants to use this in a script used interactively?  For example, the inner part of the script might be interrogating what 'user type' of file this before moving forward (eg. do something for file containing `TTree` with one name and another for files containing a `RNTuple` with a same name.

io/io/inc/ROOT/RFile.hxx
pcanal: > I was thinking of the less experienced users, but maybe at this stage, we should actually aim at the experienced users, because these are most likely to follow the developments. 

Except, aren't we also unlikely to fully revisit the documentation to make it 'for all user' later?

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,204 @@
...
+when the following conditions are met:
+- you want a simple interface that makes it easy to do things right and hard to do things wrong;
+- you only need basic Put/Get operations and don't need the more advanced TFile/TDirectory functionalities;
+- you want more robustness and better error reporting for those operations;
+- you want clearer ownership semantics expressed through the type system rather than having objects "automagically"
+  handled for you via implicit ownership of raw pointers.
+
+RFile doesn't try to cover the entirety of use cases covered by TFile/TDirectory/TDirectoryFile and is not
+a 1:1 replacement for them. It is meant to simplify the most common use cases and make them easier to handle by
+minimizing the amount of ROOT-specific quirks and conforming to more standard C++ practices.
```
pcanal, line 49: I find the tone a bit harsh.  Consider:
```
RFile is a modern and minimalistic interface to ROOT files, both local and remote, that can be used instead of TFile
when you only need basic Put/Get operations and don't need the more advanced TFile/TDirectory functionalities.  It provides:
- a simple interface that makes it easy to do things right and hard to do things wrong,
- more robustness and better error reporting for those operations,
- clearer ownership semantics expressed through the type system.

RFile doesn't cover the entirety of use cases covered by TFile/TDirectory/TDirectoryFile and is not
a 1:1 replacement for them.  It is meant to simplify the most common use cases by following newer standard C++ practices.
```

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,204 @@
...
+to update the object in the file you need to write it again (via RFile::Overwrite).
+
+RFile::Put and RFile::Overwrite are the way to write objects to the file. Both methods take a const reference to the
+object to write and don't change the ownership of the object in any way. Calling Put or Overwrite doesn't guarantee that
+the object is immediately written to the underlying storage: to ensure that, you need to call RFile::Flush (or close the
+file).
+
+## Directories
+
+Differently from TFile, the RFile class itself is not also a "directory". In fact, there is no RDirectory class at all.
```
pcanal, line 68:
```suggestion
Unlike TFile, the RFile class does not function as a directory. Moreover, there is no RDirectory class provided.
```

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,204 @@
...
+
+Directories are still an existing concept in RFile (since they are a concept in the ROOT binary format),
+but they are usually interacted with indirectly, via the use of filesystem-like string-based paths. If you Put an object
```
pcanal, line 71: > but they are usually interacted with indirectly, via the use of filesystem-like string-based paths. If you Put an object

Is the "usually" intentional?  It implies that there is (or will be) other to interact with the directories.  Is that the intent?

io/io/inc/ROOT/RFile.hxx
```diff
@@ -0,0 +1,204 @@
...
+Opening an RFile (for reading) and reading an object from it:
+~~~{.cpp}
+auto rfile = ROOT::RFile::Open("my_file.root");
+auto myObj = file->Get<TH1D>("h");
+~~~
+*/
+class RFile final {
+   enum PutFlags {
+      kPutAllowOverwrite = 0x1,
+      kPutOverwriteKeepCycle = 0x2,
```
pcanal, line 101: Did we document these flags?

io/io/test/rfile.cxx
```diff
@@ -0,0 +1,500 @@
...
+      tfile->WriteObject(&hist, "hist");
+   }
+
+   auto file = RFile::Open(fileGuard.GetPath());
+   auto hist = file->Get<TH1D>("hist");
+   EXPECT_TRUE(hist);
+
+   EXPECT_FALSE(file->Get<TH1D>("inexistent"));
+   EXPECT_FALSE(file->Get<TH1F>("hist"));
+   EXPECT_TRUE(file->Get<TH1>("hist"));
```
pcanal, line 64: ~Should we delete the result?~

io/io/test/rfile.cxx
```diff
@@ -0,0 +1,500 @@
...
+
+   std::string foo = "foo";
+   EXPECT_THROW(file->Put("foo", foo), ROOT::RException);
+}
+
+TEST(RFile, OpenInexistent)
+{
+   FileRaii fileGuard("does_not_exist.root");
+
+   // make sure that the file really does not exist, in case a previous test didn't clean it up.
```
pcanal, line 77: Is there any test that intentionally create the file named 'does_not_exist.root' ?  The comment seems to imply there is.

io/io/test/rfile.cxx
```diff
@@ -0,0 +1,500 @@
...
+   }
+   try {
+      auto f = RFile::Recreate("/a/random/directory/that/definitely/does_not_exist.root");
+      FAIL() << "trying to create a file under an inexistent directory should throw";
+   } catch (const ROOT::RException &e) {
+      EXPECT_THAT(e.what(), testing::HasSubstr("failed to open file"));
+   }
+
+   // This succeeds because Update creates the file if it doesn't exist.
+   EXPECT_NO_THROW(RFile::Update("does_not_exist.root"));
```
pcanal, line 104: This answer: https://github.com/root-project/root/pull/19958/files#r2411952125.

I recommend using a different name:
```suggestion
    FileRaii updateFileGuard("created_by_update.root");
   gSystem->Unlink(updateFileGuard.GetPath().c_str());
   EXPECT_NO_THROW(RFile::Update("created_by_update.root"));
```

io/io/test/rfile.cxx
```diff
@@ -0,0 +1,500 @@
...
+      hist->Fill(4);
+      file->Put("hist", *hist);
+   }
+
+   {
+      auto file = RFile::Open(fileGuard.GetPath());
+      EXPECT_EQ(gDirectory, gROOT);
+      auto hist = file->Get<TH1D>("hist");
+      EXPECT_EQ(hist->GetDirectory(), nullptr);
+      ASSERT_NE(hist, nullptr);
```
pcanal, line 152:
```suggestion
      ASSERT_NE(hist, nullptr);
      EXPECT_EQ(hist->GetDirectory(), nullptr);
```
Otherwise in case of failure of the `ASSERT_NE`, the actual result is a segmentation fault in the `EXPECT_EQ`

io/io/test/rfile.cxx
pcanal: Nevermind, the return value is a `unique_ptr` :)

io/io/test/rfile.cxx
pcanal: See https://github.com/root-project/root/pull/19958/files#r2411956485

io/io/test/rfile.cxx
```diff
@@ -0,0 +1,500 @@
...
+      EXPECT_NE(hist, nullptr);
+   }
+}
+
+TEST(RFile, WrongExtension)
+{
+   {
+      FileRaii fileGuard("test_rfile_wrong.root.1");
+      ROOT::TestSupport::CheckDiagsRAII diagsRaii;
+      diagsRaii.requiredDiag(kWarning, "ROOT.File", "preferred file extension is \".root\"", false);
```
pcanal, line 271: Why is this a warning rather than an Info?  (Reminder: tools can tell whether a file is a ROOT file by reading the first few bytes of the file (MIME format)).

io/io/src/RFile.cxx
```diff
@@ -0,0 +1,381 @@
...
+   static ROOT::RLogChannel sLog("ROOT.File");
+   return sLog;
+}
+
+using ROOT::Experimental::RFile;
+using ROOT::Experimental::Internal::RFileLog;
+
+static void CheckExtension(std::string_view path)
+{
+   if (ROOT::EndsWith(path, ".xml")) {
```
pcanal, line 33: Why is `.xml` so special?

io/io/inc/ROOT/RFile.hxx
silverweed: I use `TKey` in RFile whenever I want to point out that it's specifically a `TKey` - which should only be used internally and never exposed in the public API.

io/io/src/RFile.cxx
silverweed: It is certainly debatable and I'm still not sure about the best compromise. I guess a flag could help, though it's a bit unergonomic. Another option would be to use R__LOG_INFO which iirc is off by default

io/io/inc/ROOT/RFile.hxx
silverweed: Why do you think this is better than the current version?

io/io/inc/ROOT/RFile.hxx
silverweed: Yes, it implies that we "reserve the right" to also provide directory querying functionality at some point (which we know we'll probably need in some form)

io/io/inc/ROOT/RFile.hxx
silverweed: They are internal and pretty self-explanatory, but I can add doc comments to them

io/io/src/RFile.cxx
silverweed: Because `TFile::Open` has a plugin that tries to open the file as XML if that's the extension and we don't want to support XML files in RFile (there might be more unwanted extensions that I missed, but this is the one I'm aware of).

Edit: I guess there's at least also `.zip`, we should ban that as well for now.

io/io/inc/ROOT/RFile.hxx
pcanal: Since it is internal, it does not really matter that much.  I still find it a bit awkward unless we envisioned the possibility that there would eventually be a `RKey`.

io/io/src/RFile.cxx
pcanal: There is also `.zip`:
```
etc/plugins/TArchiveFile/P010_TZIPFile.C:   gPluginMgr->AddHandler("TArchiveFile", ".+[.]zip$", "TZIPFile",
etc/plugins/TFile/P080_TXMLFile.C:   gPluginMgr->AddHandler("TFile", ".+[.]xml$", "TXMLFile",
```

io/io/inc/ROOT/RFile.hxx
pcanal: Actually, it took me to look at the code to understand the 2nd one, I first read it as:
"kPut Overwrite[/change] [the default behavior] Keep Cycle" [aka 'problably' not keep the cycle]
while I guess it is meant to be read as
"[The] kPut Overwrite [operation will] Keep [the existing last] Cycle" 

The misunderstanding is based on my misunderstanding in that context of "Overwrite" as a noun or as a verb.  The fact that in `kPutAllowOverwrite` the 2nd word is treated as a verb lead my brain to infer that the 2nd word of the other enum was also to be treated as verb :).

io/io/inc/ROOT/RFile.hxx
pcanal: I don't think that 'usually' is effectively expressing "reserve the right".  In the contrary, it reads as if there is _already_ other ways.  A closer adverb might 'for now only interacted with indirectly.'

io/io/inc/ROOT/RFile.hxx
pcanal: It reads (better) to me 'english-wise'.  `Unlike` is more (american) English than `Differently from`.  The rest of that sentence is rephrase in a function way rather than indirectly focusing on the C++ semantic of inheritance.  The `In fact ... at all.` was making the 2nd sentences stand (too much) as a reinforcement of the first rather than its owns message.  Instead of saying 'we also do not plan to offer a standalone RDirectory facility', the `In fact ... at all` can be (mis?) read as 'RFile is not a directory **solely** because we are not implementing RDirectory [but if we did implement it, RFile would inherit from it]' (the last part is a slight exaggeration to clarify the drift)'

silverweed: Ok for the "unlike", I agree.
The point of the sentence for me was specifically to focus on the C++ semantic rather than the functionality (see [here](https://github.com/root-project/root/pull/19958#discussion_r2406021248)), as I think it's a relatively big change from both TFile and the older RFile prototype. However, rereading the sentence, maybe the intention was not that clear.
I have a bit of a problem with the "function as a directory" wording as it's not very clear in my opinion.
Perhaps I should just drop the paragraph altogether, I don't think it adds much value to the documentation after all.

io/io/inc/ROOT/RFile.hxx
silverweed: Fair, I added documentation to them in the followup PR.

io/io/src/RFile.cxx
pcanal: Let's discuss this as the I/O meeting

io/io/inc/ROOT/RFile.hxx
pcanal: It can indeed make sense to drop the first 2 sentences and start at `Directories are still ...` maybe rephrased now as (for example) 
```
Directories are semantically supported by the `RFile` interfaces as they are a concrete part of the ROOT binary format.
Directories are interacted with solely via the use of filesystem-like string-based paths. 
```

silverweed: See if [the updated phrasing](https://github.com/root-project/root/pull/20056/files#diff-147e8251f171f7036f5ff981fe7fbca18f3628345f7680e411e3913ec54de38eR156) works for you :)

PR #19959, opened 2025-09-23
[hist] Implement initial `RHist`

hahnjo:
It is a combination of `RHistEngine` and `RHistStats`.

### DIFF
--- a/hist/histv7/inc/LinkDef.h
+++ b/hist/histv7/inc/LinkDef.h
@@ -1,10 +1,24 @@
-// For RHistEngine, we request dictionaries for the most commonly used bin content types. This results in proper error
-// messages when trying to stream. Other instantiations will be caught by the RAxes member.
+// For RHist and RHistEngine, we request dictionaries for the most commonly used bin content types. This results in
+// proper error messages when trying to stream. Other instantiations will be caught by the RAxes member.
+#pragma link C++ class ROOT::Experimental::RHist<int>-;
+#pragma link C++ class ROOT::Experimental::RHist<unsigned>-;
+#pragma link C++ class ROOT::Experimental::RHist<long>-;
+#pragma link C++ class ROOT::Experimental::RHist<unsigned long>-;
+#pragma link C++ class ROOT::Experimental::RHist<long long>-;
+#pragma link C++ class ROOT::Experimental::RHist<unsigned long long>-;
+#pragma link C++ class ROOT::Experimental::RHist<float>-;
+#pragma link C++ class ROOT::Experimental::RHist<double>-;
+#pragma link C++ class ROOT::Experimental::RHist<ROOT::Experimental::RBinWithError>-;
+
 #pragma link C++ class ROOT::Experimental::RHistEngine<int>-;
+#pragma link C++ class ROOT::Experimental::RHistEngine<unsigned>-;
 #pragma link C++ class ROOT::Experimental::RHistEngine<long>-;
+#pragma link C++ class ROOT::Experimental::RHistEngine<unsigned long>-;
 #pragma link C++ class ROOT::Experimental::RHistEngine<long long>-;
+#pragma link C++ class ROOT::Experimental::RHistEngine<unsigned long long>-;
 #pragma link C++ class ROOT::Experimental::RHistEngine<float>-;
 #pragma link C++ class ROOT::Experimental::RHistEngine<double>-;
+#pragma link C++ class ROOT::Experimental::RHistEngine<ROOT::Experimental::RBinWithError>-;
 
 #pragma link C++ class ROOT::Experimental::RHistStats-;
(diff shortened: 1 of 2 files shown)


hist/histv7/inc/ROOT/RHist.hxx
```diff
@@ -0,0 +1,285 @@
...
+\code
+ROOT::Experimental::RHist<int> hist(10, 5, 15);
+hist.Fill(8.5);
+// hist.GetBinContent(ROOT::Experimental::RBinIndex(3)) will return 1
+\endcode
+
+The class is templated on the bin content type. For counting, as in the example above, it may be an integer type such as
+`int` or `long`. Narrower types such as `unsigned char` or `short` are supported, but may overflow due to their limited
+range and must be used with care. For weighted filling, the bin content type must be a floating-point type such as
+`float` or `double`, or the special type RBinWithError. Note that `float` has a limited significant precision of 24
```
hageboeck:
```suggestion
`float` or `double`, or the special type RBinWithError. Note that `float` has a limited significand precision of 24
```

hist/histv7/test/hist_hist.cxx
```diff
@@ -0,0 +1,124 @@
+#include "hist_test.hxx"
+
+#include <tuple>
+#include <variant>
+#include <vector>
+
+// Mostly, RHist = RHistEngine + RHistStats which are tested individually. Here we mostly check that the forwarding
+// works correctly.
+
```
hageboeck, line 10: I expect these to pass, but somehow persistifying these guarantees here at the top level would guarantee efficient resizes of `std::vector<RHist<...>>` in the future:

```suggestion
static_assert(std::is_nothrow_move_constructible_v<RHist<int>>);
static_assert(std::is_nothrow_move_assignable_v<RHist<int>>);
```

hist/histv7/inc/ROOT/RHist.hxx
```diff
@@ -0,0 +1,252 @@
...
+axes.push_back(ROOT::Experimental::RVariableBinAxis({1, 10, 100, 1000}));
+ROOT::Experimental::RHist<int> hist(axes);
+// hist.GetNDimensions() will return 2
+\endcode
+
+\warning This is part of the %ROOT 7 prototype! It will change without notice. It might trigger earthquakes.
+Feedback is welcome!
+*/
+template <typename BinContentType>
+class RHist final {
```
hageboeck, line 55: Similar to my question in the stats, is `Clear()` a thing?
I have seen quite a number of `auto clone hist2 = hist.Clone(); hist2->Clear()`. Less efficient than constructing it from scratch, but it ensures that the two histograms are compatible in all dimensions.

hist/histv7/inc/LinkDef.h
```diff
@@ -1,10 +1,18 @@
-// For RHistEngine, we request dictionaries for the most commonly used bin content types. This results in proper error
-// messages when trying to stream. Other instantiations will be caught by the RAxes member.
+// For RHist and RHistEngine, we request dictionaries for the most commonly used bin content types. This results in
+// proper error messages when trying to stream. Other instantiations will be caught by the RAxes member.
+#pragma link C++ class ROOT::Experimental::RHist<int>-;
```
hageboeck, line 3: Shouldn't the unsigned types be included? When only counting up, these seem natural.

hist/histv7/inc/ROOT/RHist.hxx
hageboeck: Found it.

hist/histv7/inc/LinkDef.h
hahnjo: Yes maybe. I thought about them initially, but can't figure out a reason why I didn't request dictionaries for them in the end. I would like to do this in a followup PR because then we also need them for `RHistEngine`.

hist/histv7/test/hist_hist.cxx
hahnjo: Yes, these pass at the moment. I would propose to handle this in the same followup PR as the dictionaries for `unsigned` integer types since we want the same for `RHistEngine` (even though this is probably transitively required if we test for `RHist`)

hageboeck: Sure, no problem. Same for the unsigned-type dictionaries.

hist/histv7/inc/LinkDef.h
hahnjo: Never mind, I forgot that I'm already fixing up the dictionaries for `RBinWithError`, so might as well include it here.

hist/histv7/test/hist_hist.cxx
hahnjo: also included

PR #19966, opened 2025-09-24
[ntuple] automatic evolution of nullable fields

jblomer:
Conversion between `std::unique_ptr<T>` and `std::optional<T>` as well as `T` --> `std::unique_ptr|std::optional<T>`.

### DIFF
--- a/tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
+++ b/tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
@@ -195,6 +195,9 @@ class RNullableField : public RFieldBase {
    ROOT::Internal::RColumnIndex fNWritten{0};
 
 protected:
+   // For reading, indicates that we read type T as a nullable field of type T, i.e. the value is always present
+   bool fIsEvolvedFromInnerType = false;
+
    const RFieldBase::RColumnRepresentations &GetColumnRepresentations() const final;
    void GenerateColumns() final;
    void GenerateColumns(const ROOT::RNTupleDescriptor &) final;
--- a/tree/ntuple/src/RField.cxx
+++ b/tree/ntuple/src/RField.cxx
@@ -849,18 +849,20 @@ const ROOT::RFieldBase::RColumnRepresentations &ROOT::RNullableField::GetColumnR
                                                   {ENTupleColumnType::kIndex64},
                                                   {ENTupleColumnType::kSplitIndex32},
                                                   {ENTupleColumnType::kIndex32}},
-                                                 {});
+                                                 {{}});
    return representations;
 }
 
 void ROOT::RNullableField::GenerateColumns()
 {
+   R__ASSERT(!fIsEvolvedFromInnerType);
    GenerateColumnsImpl<ROOT::Internal::RColumnIndex>();
 }
 
 void ROOT::RNullableField::GenerateColumns(const ROOT::RNTupleDescriptor &desc)
 {
-   GenerateColumnsImpl<ROOT::Internal::RColumnIndex>(desc);
+   if (!fIsEvolvedFromInnerType)
+      GenerateColumnsImpl<ROOT::Internal::RColumnIndex>(desc);
 }
 
 std::size_t ROOT::RNullableField::AppendNull()
@@ -881,8 +883,16 @@ void ROOT::RNullableField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
 {
    static const std::vector<std::string> prefixes = {"std::optional<", "std::unique_ptr<"};
 
-   EnsureMatchingOnDiskField(desc, kDiffTypeName).ThrowOnError();
-   EnsureMatchingTypePrefix(desc, prefixes).ThrowOnError();
+   auto success = EnsureMatchingOnDiskField(desc, kDiffTypeName);
+   if (!success) {
+      fIsEvolvedFromInnerType = true;
+   } else {
+      success = EnsureMatchingTypePrefix(desc, prefixes);
+      fIsEvolvedFromInnerType = !success;
+   }
+
+   if (fIsEvolvedFromInnerType)
+      fSubfields[0]->SetOnDiskId(GetOnDiskId());
 }
 
 ROOT::RNTupleLocalIndex ROOT::RNullableField::GetItemIndex(ROOT::NTupleSize_t globalIndex)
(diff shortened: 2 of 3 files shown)


tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
```diff
@@ -205,9 +208,12 @@ protected:
 
    void ReconcileOnDiskField(const RNTupleDescriptor &desc) final;
 
-   /// Given the index of the nullable field, returns the corresponding global index of the subfield or,
-   /// if it is null, returns `kInvalidNTupleIndex`
-   RNTupleLocalIndex GetItemIndex(ROOT::NTupleSize_t globalIndex);
+   /// Given the index of the nullable field, returns the corresponding cluster-local index of the subfield or,
```
enirolf:
```suggestion
   /// Given the global index of the nullable field, returns the corresponding cluster-local index of the subfield or,
```

tree/ntuple/inc/ROOT/RField/RFieldSTLMisc.hxx
```diff
@@ -205,9 +208,12 @@ protected:
...
+   /// if it is null, returns a default constructed RNTupleLocalIndex
+   RNTupleLocalIndex GetItemIndex(NTupleSize_t globalIndex);
+   /// Given the index of the nullable field, returns the corresponding local index of the subfield or,
```
enirolf:
```suggestion
   /// Given the cluster-local index of the nullable field, returns the corresponding cluster-local index of the subfield or,
```

tree/ntuple/src/RField.cxx
```diff
@@ -830,8 +831,13 @@ void ROOT::RNullableField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
    static const std::vector<std::string> prefixes = {"std::optional<", "std::unique_ptr<"};
 
    const auto &fieldDesc = desc.GetFieldDescriptor(GetOnDiskId());
-   EnsureMatchingOnDiskField(fieldDesc, kDiffTypeName);
-   EnsureMatchingTypePrefix(fieldDesc, prefixes);
+   try {
+      EnsureMatchingOnDiskField(fieldDesc, kDiffTypeName);
+      EnsureMatchingTypePrefix(fieldDesc, prefixes);
+   } catch (const RException &) {
+      fSubfields[0]->SetOnDiskId(GetOnDiskId());
```
silverweed: This looks a bit sketchy to me: since we're using RException for everything, how can we know that it's not a fatal exception, possibly coming from a deep call inside the Ensure methods?

Unrelated, for the Ensure methods I would adopt the same approach as e.g. `VerifyXxHash3`, where the call returns a `RResult` rather than throwing an exception directly (of course in a separate PR)

tree/ntuple/test/ntuple_evolution_type.cxx
```diff
@@ -207,3 +207,32 @@ TEST(RNTupleEvolution, CheckPairTuple)
...
+      *i = 13;
+      writer->Fill();
+   }
+
+   auto reader = RNTupleReader::Open("ntpl", fileGuard.GetPath());
+
+   auto v1 = reader->GetView<std::unique_ptr<std::int64_t>>("o");
+   auto v2 = reader->GetView<std::optional<std::int64_t>>("u");
+   auto v3 = reader->GetView<std::unique_ptr<std::int64_t>>("i");
+   auto v4 = reader->GetView<std::optional<std::int64_t>>("i");
```
enirolf, line 307: What happens if I try to create a view for "i" with an inner type of the nullable field that cannot represent the original on-disk type (e.g., `std::optional<std::string>`?

enirolf (commented):
Nice! I have two minor documentation suggestions and a question, see below.

tree/ntuple/test/ntuple_evolution_type.cxx
jblomer: An exception will be thrown because the inner std::string field does not know how to read from an on-disk int field. I can add this to the test.

tree/ntuple/src/RField.cxx
hahnjo: Also similar question to https://github.com/root-project/root/pull/19937#pullrequestreview-3263430792: Is unconditionally delegating to the item field maybe confusing for the error message?

jblomer: > Unrelated, for the Ensure methods I would adopt the same approach as e.g. `VerifyXxHash3`, where the call returns a `RResult` rather than throwing an exception directly (of course in a separate PR)

Makes sense. See https://github.com/root-project/root/pull/19971

tree/ntuple/src/RField.cxx
```diff
@@ -860,7 +860,8 @@ void ROOT::RNullableField::GenerateColumns()
 
 void ROOT::RNullableField::GenerateColumns(const ROOT::RNTupleDescriptor &desc)
 {
-   GenerateColumnsImpl<ROOT::Internal::RColumnIndex>(desc);
+   if (!fIsEvolvedFromInnerType)
```
hahnjo, line 864: Maybe add `R__ASSERT(!fIsEvolvedFromInnerType)` in `RNullableField::GenerateColumns` (as we (will) have in `RField<std::vector<bool>>::GenerateColumns()`)?

tree/ntuple/src/RField.cxx
```diff
@@ -1010,22 +1027,38 @@ std::size_t ROOT::ROptionalField::AppendImpl(const void *from)
...
+         fItemDeleter->operator()(to, true /* dtorOnly */);
+      *engagementPtr = false;
    }
 }
 
+void ROOT::ROptionalField::ReadGlobalImpl(ROOT::NTupleSize_t globalIndex, void *to)
+{
+   auto itemIndex = GetItemIndex(globalIndex);
+   const bool hasOnDiskValue = itemIndex.GetIndexInCluster() != ROOT::kInvalidNTupleIndex;
+   PrepareRead(to, hasOnDiskValue);
```
hahnjo, line 1072: A thought for the future (doesn't need to be addressed in this PR): We may also want to return a pointer from `ROptionalField::PrepareRead` because I don't think the C++ standard guarantees that the contained value is the first member of `std::optional` nor that it necessarily is at the same address (for example with(out) empty base class optimization).

hahnjo (approved):
LGTM

Issue #7684, opened 2021-03-25
[PyROOT] Importing ROOT in JupyterLab reduces kernel performance


agoose77:
After importing `ROOT` in a Jupyter Notebook in which Jupyter Widgets are used, the performance of figure interactions like panning and zooming drastically falls off. After identifying ROOT as the culprit, I found that the poor performance is caused by sleeping invoked during [these hooks](https://github.com/root-project/root/blob/aedbc69c9c41a9d3eb6a5e24eb7382547141d85a/bindings/jupyroot/python/JupyROOT/helpers/utils.py#L400-L401) which are run for each kernel execution.

agoose77:
An interactive binder: [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gist/agoose77/7e9a8c146a7dc68c87f47678d1f2e51c/HEAD?filepath=%2Flab)

etejedor:
@agoose77 thank you very much for this extensive research you did!

Do you think you could apply this to `JupyROOT.helpers.utils`?

agoose77:
@etejedor sure, I'll add this to my TODO list :)

PR #19967, opened 2025-09-24
[jupyter] optimize jsroot loading

linev:
1. Use JavaScript `import()` function to load jsroot. Avoiding by-hand creation of script elements
2. Solves problem of multiple jsroot loading by different plots
3. Make code much more compact, produce the only global function per plot
4. Keep requirejs part to support Visual Studio Code
5. Tested both in `jupyter lab` and `jupyter notebook`

### DIFF
--- a/bindings/jupyroot/python/JupyROOT/helpers/utils.py
+++ b/bindings/jupyroot/python/JupyROOT/helpers/utils.py
@@ -49,60 +49,37 @@ _jsNotDrawableClassesPatterns = ["TEve*"]
 
 _jsCanvasWidth = 800
 _jsCanvasHeight = 600
-_jsCode = """
 
+_jsCode = """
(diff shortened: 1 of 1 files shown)


ferdymercury:
Thanks a lot! Do you know if this change affects: https://github.com/root-project/root/issues/7684 ?

guitargeek:
Nice! I was also thinking some time ago that this should be improved :+1:

linev:
> Thanks a lot! Do you know if this change affects: #7684 ?

I don't think so. pre/post execute hooks remain as is. 
I just optimize loading of jsroot in case of multiple jsroot drawing stored in the notebook. 
PR is triggered by [recent post on the forum](https://root-forum.cern.ch/t/how-to-correctly-display-tcanvas-in-jupyter-notebook/64182/16).

I also will need to update JSROOT which should workaround some jupyter HTML layout behavior.

PR #19970, opened 2025-09-24
[jupyter] use zipping for stored json

linev:
Up to now plain json code embed into ipython notebooks.
This is very inefficient and can make notebook huge.
Especially because of fact that notebook stored as json - so any double-quote `"` symbol should be escaped.

Therefore use standard compression function of ROOT and then base64 encoding to produce string suitable for embeding. This done in `TBufferJSON::zipJSON()` method.

On JSROOT side just provide correspondent `unzipJSON()` function.

And modify jupyter `utils.py` code to use such functionality.

### DIFF
--- a/bindings/jupyroot/python/JupyROOT/helpers/utils.py
+++ b/bindings/jupyroot/python/JupyROOT/helpers/utils.py
@@ -57,10 +57,12 @@ _jsCode = """
    function process_{jsDivId}() {{
       function drawPlot(Core) {{
          Core.settings.HandleKeys = false;
-         const obj = Core.parse({jsonContent});
-         Core.draw("{jsDivId}", obj, "{jsDrawOptions}");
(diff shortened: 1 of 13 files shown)


linev:
Here one can compare new produced notebook code:

https://github.com/linev/misc/blob/master/notebook/Simple.ipynb?short_path=26ba30e#L30-L35

with old stored notebooks:

https://github.com/linev/misc/blob/master/notebook/Simple0.ipynb?short_path=58fd63c#L32-L36

guitargeek (approved):
Thank you very much for implementing this so quickly!

Also it's good that you used `zlib` for the compression, as ROOT is always guaranteed to include this compression library (either via `builtin_zlib` or not). So it should be possible to display notebooks produced with any build configuration of ROOT with any other build configuration.

linev:
Any supported compression algorithm can be used here.
For the moment I specify `kZLIB`, but it can be any.

PR #19971, opened 2025-09-24
[ntuple] use RResult in field reconciliation

jblomer:
Let EnsureMatchingOnDiskField() and EnsureMatchingTypePrefix() return an RResult<void> instead of directly throwing.

Follow-up of https://github.com/root-project/root/pull/19966#discussion_r2374731576


silverweed (approved):
Thanks!

PR #19974, opened 2025-09-25
[thisroot] resolve symbolic links

ferdymercury:
Fixes https://its.cern.ch/jira/browse/ROOT-10415

### DIFF
--- a/config/thisroot.csh
+++ b/config/thisroot.csh
@@ -58,6 +58,11 @@ endif
 if ($?thisrootdir) then
 
 setenv ROOTSYS "`(dirname ${thisrootdir})`"
+if ( ! -f "$ROOTSYS/bin/root-config" ) then
+    echo "ERROR: root-config not found under ROOTSYS=\"$ROOTSYS/\"" >&2
+    set ROOTSYS =
+    return 1
+endif
 
 if ($?old_rootsys) then
    setenv PATH `set DOLLAR='$'; echo $PATH | sed -e "s;:$old_rootsys/bin:;:;g" \


config/thisroot.sh
```diff
@@ -238,11 +238,27 @@ if [ -z "${SOURCE}" ]; then
...
+      thisroot="$(dirname "${SOURCE}")"
+      
+      ROOTSYS=$(cd "${thisroot}/.." > /dev/null && pwd); export ROOTSYS
+      if [ -z "$ROOTSYS" ]; then
+         echo "ERROR: \"cd ${thisroot}/..\" or \"pwd\" failed" >&2
+         return 1
+      fi
+   else
+      echo "Missing file ${SOURCE}"
+      ROOTSYS=; export ROOTSYS
```
hageboeck: I think using the canonicalise function of readlink should make this very easy. We just need to take away `bin/` afterwards
```suggestion
   thisrootdir=$(dirname $(readlink -e "${SOURCE}"))                                                                          
   export ROOTSYS=${thisrootdir%/*}
```

hageboeck (commented):
Hello, great initiative! I believe I found a hack to do this in a two-liner. I tested locally with the original thisroot.sh and with a link, and both resulted in what I expected.

config/thisroot.sh
ferdymercury: Thanks, nice! but is this correctly printing out error messages in case link is broken, folder .. does not exist, etc?

ferdymercury: Also two side questions:
does the parameter expansion %/* also work for ksh, zsh, ... ?
would it also work for the .csh and .fish script variants?

hageboeck: Good questions ...
If the link is broken, there is no script to source I believe. So I don't think that this case needs to be considered.

If the folder does not exist, I believe there is no script to source either, so I guess we don't need to do anything.

I didn't look up the syntax for other shells, because this is the `.sh` variant.

ferdymercury:
In fish, it's:

```
set thisroot (path dirname $SOURCE)
set -xg ROOTSYS (set oldpwd $PWD; cd $thisroot/.. > /dev/null;pwd;cd $oldpwd; set -e oldpwd)
```

In csh, it's:

```
set thisroot="`dirname ${thisfile}`"
setenv ROOTSYS "`(cd ${thisroot}/..;pwd)`"
```

But I need help here on how to change it

hageboeck (approved):
LGTM
It's a very good idea to check for root-config, because that will detect most errors!

config/thisroot.sh
```diff
@@ -238,15 +238,16 @@ if [ -z "${SOURCE}" ]; then
       return 1
    fi
 else
-   # get param to "."
-   thisroot="$(dirname "${SOURCE}")"
-   ROOTSYS=$(cd "${thisroot}/.." > /dev/null && pwd); export ROOTSYS
-   if [ -z "$ROOTSYS" ]; then
-      echo "ERROR: \"cd ${thisroot}/..\" or \"pwd\" failed" >&2
+   thisrootdir=$(dirname $(readlink -e "${SOURCE}"))
```
hageboeck: I just tried this on mac and linux, and the `-e` option is not supported on mac. This seems to work instead:
```suggestion
   thisrootdir=$(dirname $(realpath "${SOURCE}"))                                                                          
```

hageboeck:
@ferdymercury we must have made a mistake during merging. I believe that this PR https://github.com/ferdymercury/root/pull/11 fixes it.

ferdymercury:
> @ferdymercury we must have made a mistake during merging. I believe that this PR [ferdymercury#11](https://github.com/ferdymercury/root/pull/11) fixes it.

Thanks a lot! Seems fixed now.

I'll leave it as "Draft" while we think on how to deal with https://github.com/root-project/root/pull/19974#issuecomment-3333742202

PR #19976, opened 2025-09-25
[df] Add `RNTupleWriteOptions` to `RSnapshotOptions` and add format-compatibility checks

### LINKED ISSUE #19784: [df] RNTuple snapshot + TTree-specific options
Follow-up from https://github.com/root-project/root/pull/19599#discussion_r2298425696: RNTuple snapshotting currently does not support `fAutoFlush`, which doesn't really map to RNTuple's target cluster size. We need to discuss how and if we want to support it.

Note that `UntypedSnapshotTTreeHelperMT` is also questionable in that regard because it writes / flushes output trees in `FinalizeTask`. So for RDataFrame snapshotting to TTree in MT mode, it's already more of an upper guarantee, but I guess many cluster will actually be smaller than the user wants them...

enirolf:
For snapshotting to RNTuple, users can now pass an `RNTupleWriteOptions` to `RSnapshotOptions` to configure the output RNTuple. Compression settings that have been set directly through `RSnapshotOptions::fCompressionAlgorithm` and `RSnapshotOptions::fCompressionLevel` are propagated to `RSnapshotOptions::fNTupleWriteOpts`, provided that they haven't been set there already as well.

In addition, a check has been added to warn users when they set options that are specific to one output format, but `RSnapshotOptions.fOutputFormat` has been set to use the other.

Closes #19784.

### DIFF
--- a/tree/dataframe/inc/ROOT/RSnapshotOptions.hxx
+++ b/tree/dataframe/inc/ROOT/RSnapshotOptions.hxx
@@ -11,8 +11,9 @@
 #ifndef ROOT_RSNAPSHOTOPTIONS
 #define ROOT_RSNAPSHOTOPTIONS
 
+#include "ROOT/RNTupleWriteOptions.hxx"
 #include <Compression.h>
-#include <optional>
+
 #include <string_view>
 #include <string>
 
@@ -31,6 +32,7 @@ struct RSnapshotOptions {
    RSnapshotOptions() = default;
    RSnapshotOptions(std::string_view mode, ECAlgo comprAlgo, int comprLevel, int autoFlush, int splitLevel, bool lazy,
                     bool overwriteIfExists = false, bool vector2RVec = true, int basketSize = -1,
+                    ROOT::RNTupleWriteOptions ntupleWriteOpts = ROOT::RNTupleWriteOptions(),
                     ESnapshotOutputFormat outputFormat = ESnapshotOutputFormat::kDefault)
       : fMode(mode),
         fCompressionAlgorithm(comprAlgo),
@@ -41,6 +43,7 @@ struct RSnapshotOptions {
         fOverwriteIfExists(overwriteIfExists),
         fVector2RVec(vector2RVec),
         fBasketSize(basketSize),
+        fNTupleWriteOpts(ntupleWriteOpts),
         fOutputFormat(outputFormat)
    {
    }
@@ -55,7 +58,8 @@ struct RSnapshotOptions {
    bool fVector2RVec = true;         ///< If set to true will convert std::vector columns to RVec when saving to disk
    int fBasketSize = -1;             ///< Set a custom basket size option. For more details, see
                                      ///< https://root.cern/manual/trees/#baskets-clusters-and-the-tree-header
-   ESnapshotOutputFormat fOutputFormat = ESnapshotOutputFormat::kDefault; ///< Which data format to write to
+   ROOT::RNTupleWriteOptions fNTupleWriteOpts = ROOT::RNTupleWriteOptions(); ///< RNTuple-specific write options
+   ESnapshotOutputFormat fOutputFormat = ESnapshotOutputFormat::kDefault;    ///< Which data format to write to
 };
 } // namespace RDF
 } // namespace ROOT
--- a/tree/dataframe/src/RDFInterfaceUtils.cxx
+++ b/tree/dataframe/src/RDFInterfaceUtils.cxx
@@ -939,6 +939,37 @@ void CheckForDuplicateSnapshotColumns(const ColumnNames_t &cols)
    }
 }
 
+void CheckSnapshotOptionsFormatCompatibility(const ROOT::RDF::RSnapshotOptions &opts)
+{
+   const ROOT::RDF::RSnapshotOptions defaultSnapshotOpts;
+   if ((opts.fOutputFormat == ROOT::RDF::ESnapshotOutputFormat::kTTree ||
+        opts.fOutputFormat == ROOT::RDF::ESnapshotOutputFormat::kDefault) &&
+       opts.fNTupleWriteOpts != defaultSnapshotOpts.fNTupleWriteOpts) {
+      Warning(
+         "Snapshot",
+         "The RNTuple-specific fNTupleWriteOptions option in RSnapshotOptions has been set, but the output format is "
+         "set to TTree, so this option won't have any effect. Use the other options available in RSnapshotOptions to "
+         "configure the output TTree. Alternatively, change fOutputFormat to snapshot to RNTuple instead.");
+   } else if (opts.fOutputFormat == ROOT::RDF::ESnapshotOutputFormat::kRNTuple) {
+      if (opts.fAutoFlush != defaultSnapshotOpts.fAutoFlush) {
+         Warning("Snapshot", "The TTree-specific fAutoFlush option in RSnapshotOptions has been set, but the output "
+                             "format is set to RNTuple, so this option won't have any effect. Use the "
+                             "fNTupleWriteOptions option available in RSnapshotOptions to configure the output "
+                             "RNTuple. Alternatively, change fOutputFormat to snapshot to TTree instead.");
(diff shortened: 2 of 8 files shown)


tree/dataframe/src/RDFInterfaceUtils.cxx
```diff
@@ -939,6 +939,37 @@ void CheckForDuplicateSnapshotColumns(const ColumnNames_t &cols)
...
+         "Snapshot",
+         "The RNTuple-specific fNTupleWriteOptions option in RSnapshotOptions has been set, but the output format is "
+         "set to TTree, so this option won't have any effect. Use the other options available in RSnapshotOptions to "
+         "configure the output TTree. Alternatively, change fOutputFormat to snapshot to RNTuple instead.");
+   } else if (opts.fOutputFormat == ROOT::RDF::ESnapshotOutputFormat::kRNTuple) {
+      if (opts.fAutoFlush != defaultSnapshotOpts.fAutoFlush) {
+         Warning("Snapshot", "The TTree-specific fAutoFlush option in RSnapshotOptions has been set, but the output "
+                             "format is set to RNTuple, so this option won't have any effect. Use the "
+                             "fNTupleWriteOptions option available in RSnapshotOptions to configure the output "
+                             "RNTuple. Alternatively, change fOutputFormat to snapshot to TTree instead.");
```
pcanal: Consider:

```suggestion
      const char *issueWith = nullptr;
      if (opts.fAutoFlush != defaultSnapshotOpts.fAutoFlush) {
         issueWith =  "fAutoFlush";
      } else if ....
          ...
      }
      Warning("Snapshot", "The TTree-specific %s option in RSnapshotOptions has been set, but the output "
                          "format is set to RNTuple, so this option won't have any effect. Use the "
                          "fNTupleWriteOptions option available in RSnapshotOptions to configure the output "
                          "RNTuple. Alternatively, change fOutputFormat to snapshot to TTree instead.", issueWith);
```

tree/dataframe/src/RDFSnapshotHelpers.cxx
```diff
@@ -852,8 +852,14 @@ void ROOT::Internal::RDF::UntypedSnapshotRNTupleHelper::Initialize()
    }
    model->Freeze();
 
-   ROOT::RNTupleWriteOptions writeOptions;
-   writeOptions.SetCompression(fOptions.fCompressionAlgorithm, fOptions.fCompressionLevel);
+   // Propagate fOptions.fCompressionAlgorithm and fOptions.fCompressionLevel to fOptions.fNTupleWriteOpts *only* if the
+   // compression settings in fNTupleWriteOpts have not been changed, and the compression algorithm or level in fOptions
+   // have.
+   if (fOptions.fNTupleWriteOpts.GetCompression() == RCompressionSetting::EDefaults::kUseGeneralPurpose &&
+       (fOptions.fCompressionAlgorithm != RCompressionSetting::EAlgorithm::kZLIB ||
```
jblomer: I think that leaves the possibility that a user explicitly sets fCompression... to what happens to be the current default, but then is surprised because RNTuple doesn't pick it up but continue to use its zstd default. I don't have a good solution though, maybe that's ok.

enirolf: Hmm, good point. Perhaps to avoid situations like these it might be better to also warn users for `fCompression...`, and require them to be set through `fNTupleWriteOpts` only.

enirolf: After the merger of #20030, the behavior is now such that `fCompression[...]` will be propagated to `fNTupleWriteOpts` if and only if in there it has not changed from the default. I think that should (reasonably) remove possible ambiguities regarding what compression is used.

enirolf:
After discussing offline with @vepadulano, we concluded that in the end it makes more sense to throw an exception in case either of `RSnapshotOptions::fCompression[...]` *and* `RNTupleWriteOptions::fCompression` has been set to conflicting (non-default) settings. I've updated the first commit accordingly.

tree/dataframe/inc/ROOT/RSnapshotOptions.hxx
```diff
@@ -11,8 +11,9 @@
 #ifndef ROOT_RSNAPSHOTOPTIONS
 #define ROOT_RSNAPSHOTOPTIONS
 
+#include "ROOT/RNTupleWriteOptions.hxx"
```
vepadulano: Apologies for bringing this up only now. By looking at the code more closely, I am reminded that a lot of work has been put into removing all header inclusions in RDataFrame of other parts of ROOT I/O (namely TTree, TChain, TFile and related headers). This now introduces another header inclusion which is logically similar to including something from TTree, and I'm afraid it goes against the rest of the work that was done. I don't have yet a clear idea on how to overcome this.

vepadulano: There are two ways I can see:

* Pass the RNTupleWriteOptions via `std::unique_ptr` and default to `nullptr`.
* Unpack the various options from RNTupleWriteOptions flat into RSnapshotOptions, and then only internally construct the RNTupleWriteOptions in the .cxx

vepadulano: The second option would have the slight advantage of removing the need for the checks of compression settings, we just wouldn't create another compression setting option in the RSnapshotOptions

enirolf: Ah, thanks for bringing this up. I agree that the second option seems more favorable. That would also keep the struct as simple as possible. Perhaps it's not necessary to add *all* write options for the time being, but only the ones we expect are more commonly tweaked, but on the other hand it also doesn't cost much more effort to just add them all at once.

enirolf: After discussing offline, a subset of the write options has been added directly to `RSnapshotOptions`. These were chosen because we foresee them to be most immediately useful to users of RDataFrame and Snapshot. Of course, if the need arises, the remaining options can be added as well.

tree/dataframe/inc/ROOT/RSnapshotOptions.hxx
```diff
@@ -25,38 +24,194 @@ enum class ESnapshotOutputFormat {
...
+<td><code>fSplitLevel</code></td>
+<td><code>int</code></td>
+<td>99</td>
+<td>Split level of the output branches</td>
+</tr>
+<tr>
+<td><code>fBasketSize</code></td>
+<td><code>int</code></td>
+<td>-1</td>
+<td>Output basket size</td>
```
pcanal: If I recall correctly `-1` is not a valid value for `SetBasketSize`
```suggestion
<td>Output basket size (-1 to request the TTree's default)</td>
```

tree/dataframe/test/dataframe_snapshot_ntuple.cxx
```diff
@@ -130,6 +130,41 @@ TEST(RDFSnapshotRNTuple, WriteOpts)
    EXPECT_EQ(2, reader->GetDescriptor().GetClusterDescriptor(0).GetPageRange(0).GetPageInfos().size());
    EXPECT_EQ(2, reader->GetDescriptor().GetClusterDescriptor(1).GetPageRange(0).GetPageInfos().size());
    EXPECT_FALSE(reader->GetDescriptor().GetClusterDescriptor(0).GetPageRange(0).GetPageInfos()[0].HasChecksum());
+
+   // Setting TTree-specific options while the output format is set to RNTuple should result in a warning
+   {
+      RSnapshotOptions opts;
```
hahnjo: The compilers warn that `opts` shadows the variable at the parent scope

enirolf:
> The PR summary says the default for `fInitialUnzippedPageSize` is 258 B 😉

Thanks for the catch, also fixed now in the actual docs :)

vepadulano (approved):
Very nice, thank you! If I see it correctly, there's still the `opts` variable being shadowed by other `opts` variables in `TEST(RDFSnapshotRNTuple, WriteOpts)`. With that fixed this should be good to go.

PR #19977, opened 2025-09-25
[ci] add debian13

silverweed:
Add new debian stable to the CI. Will be backported to 636


.github/workflows/root-ci.yml
```diff
@@ -391,6 +391,8 @@ jobs:
             overrides: ["CMAKE_CXX_STANDARD=23"]
           - image: debian125
             overrides: ["CMAKE_CXX_STANDARD=20", "dev=ON", "CMAKE_CXX_FLAGS=-Wsuggest-override"]
+          - image: debian13
+            overrides: ["CMAKE_CXX_STANDARD=20", "CMAKE_CXX_FLAGS=-Wsuggest-override"]
```
guitargeek: Why not `dev=ON` as well? Or mode the `dev=ON` from `debian125` to `debian13`? It's probably nice to turn the warnings into errors for the newer compiler that we get with `debian13`?

dpiparo: maybe, can we set in the configuration file, and not override, c++23 as a standard given that the distro is new?

guitargeek (approved):
Thank you very much!

.github/workflows/root-ci.yml
hahnjo: We usually don't set `CMAKE_CXX_STANDARD` in the configuration files because it means that release binaries are also built that way. In the past, we wanted to have the compiler default...

PR #19978, opened 2025-09-25
[webgui] handle ClearOnClose for pending connections

linev:
`RWebWindow::ClearOnClose` let destroy handle when all connections are closed by widget - or when widget is destroyed.

But if pending connection is not reconnected and finally removed, `ClearOnClose` was not emptied. 
Therefore add such check when pending connections processed.

Add separate timeout for reconnecting of pending connections.
Default is 15 seconds - while in such case browser process already exists and only JavaScript has to be reinitialized.
Can be changed via ` WebGui.ReconnectTmout` rootrc parameter.

If such timeout zero or negative - connection is not possible to reconnect.

### DIFF
--- a/gui/browserv7/src/RBrowser.cxx
+++ b/gui/browserv7/src/RBrowser.cxx
@@ -303,7 +303,7 @@ RBrowser::RBrowser(bool use_rcanvas)
    fWebWindow = RWebWindow::Create();
    if (!fWebWindow)
       return;
-   
+
    fWebWindow->SetDefaultPage("file:rootui5sys/browser/browser.html");
 
    std::string sortby = gEnv->GetValue("WebGui.Browser.SortBy", "name"),
@@ -591,6 +591,15 @@ void RBrowser::Hide()
 {
    if (fWebWindow)
       fWebWindow->CloseConnections();
+}
+
+///////////////////////////////////////////////////////////////////////////////////////////////////////
+/// Run widget Sync method - processing pending actions
+
+void RBrowser::Sync()
+{
+   if (fWebWindow)
+      fWebWindow->Sync();
 }
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/gui/webdisplay/src/RWebWindow.cxx
+++ b/gui/webdisplay/src/RWebWindow.cxx
@@ -647,6 +647,8 @@ void RWebWindow::CheckPendingConnections()
 
    ConnectionsList_t selected;
 
+   bool do_clear_on_close = false;
+
    {
       std::lock_guard<std::mutex> grd(fConnMutex);
 
@@ -663,7 +665,12 @@ void RWebWindow::CheckPendingConnections()
       };
 
       fPendingConn.erase(std::remove_if(fPendingConn.begin(), fPendingConn.end(), pred), fPendingConn.end());
+
+      do_clear_on_close = (selected.size() > 0) && (fPendingConn.size() == 0) && (fConn.size() == 0);
    }
+
+   if (do_clear_on_close)
+      fClearOnClose.reset();
 }
 
 
--- a/gui/webdisplay/src/RWebWindow.cxx
+++ b/gui/webdisplay/src/RWebWindow.cxx
@@ -643,7 +643,8 @@ void RWebWindow::CheckPendingConnections()
 
    timestamp_t stamp = std::chrono::system_clock::now();
 
-   float tmout = fMgr->GetLaunchTmout();
+   float launch_tmout = fMgr->GetLaunchTmout(),
+         reconnect_tmout = fMgr->GetReconnectTmout();
 
    ConnectionsList_t selected;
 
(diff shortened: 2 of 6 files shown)


gui/webdisplay/src/RWebWindow.cxx
```diff
@@ -643,16 +643,21 @@ void RWebWindow::CheckPendingConnections()
 
    timestamp_t stamp = std::chrono::system_clock::now();
 
-   float tmout = fMgr->GetLaunchTmout();
+   float launch_tmout = fMgr->GetLaunchTmout(),
+         reconnect_tmout = fMgr->GetReconnectTmout();
 
    ConnectionsList_t selected;
 
+   bool do_clear_on_close = false;
```
silverweed:
```suggestion
   bool doClearOnClose = false;
```

gui/webdisplay/src/RWebWindow.cxx
```diff
@@ -643,16 +643,21 @@ void RWebWindow::CheckPendingConnections()
 
    timestamp_t stamp = std::chrono::system_clock::now();
 
-   float tmout = fMgr->GetLaunchTmout();
+   float launch_tmout = fMgr->GetLaunchTmout(),
+         reconnect_tmout = fMgr->GetReconnectTmout();
```
silverweed:
```suggestion
   float launchTmout = fMgr->GetLaunchTmout();
   float reconnectTmout = fMgr->GetReconnectTmout();
```

gui/webdisplay/src/RWebWindow.cxx
```diff
@@ -663,7 +668,12 @@ void RWebWindow::CheckPendingConnections()
       };
 
       fPendingConn.erase(std::remove_if(fPendingConn.begin(), fPendingConn.end(), pred), fPendingConn.end());
+
+      do_clear_on_close = (selected.size() > 0) && (fPendingConn.size() == 0) && (fConn.size() == 0);
    }
+
+   if (do_clear_on_close)
+      fClearOnClose.reset();
```
silverweed, line 676: This is quite an odd way to implement a callback: why is a `shared_ptr` needed? In particular, if the caller of `SetClearOnClose` keeps this alive they won't get the signal at this time, no?

gui/webdisplay/src/RWebWindow.cxx
silverweed: For consistency with ROOT recommended coding style

gui/webdisplay/src/RWebWindow.cxx
linev: Because this functionality already exists for macros like this:
```
auto br = std::make_shared<ROOT::RBrowser>(true);
br->ClearOnClose(br);
```

It is the way to store `shared_ptr` and release it when widget closed.
Same is for RFitPanel, RGeomViewer and most of other web widgets

linev: Without such functionality `shared_ptr` immediately deleted at the end of the macro execution

gui/webdisplay/src/RWebWindow.cxx
```diff
@@ -643,16 +643,21 @@ void RWebWindow::CheckPendingConnections()
 
    timestamp_t stamp = std::chrono::system_clock::now();
 
-   float tmout = fMgr->GetLaunchTmout();
+   float launch_tmout = fMgr->GetLaunchTmout();
```
silverweed: These variables still use the "wrong" casing

gui/browserv7/src/RBrowser.cxx
```diff
@@ -593,6 +593,15 @@ void RBrowser::Hide()
       fWebWindow->CloseConnections();
 }
 
+///////////////////////////////////////////////////////////////////////////////////////////////////////
+/// Run widget Sync method - processing pending actions
+
+void RBrowser::Sync()
```
silverweed, line 599: When is this method called?

linev: I put it here for future use.
If one want to handle timeouts with connection in RBrowser,
one should regularly call this function

silverweed: So in the case of rootbrowse for example, would I have to call this method in the [busy loop at the end](https://github.com/root-project/root/blob/master/main/src/rootbrowse.cxx#L165)?
Or should that be done by `RWebBrowserImp`? (e.g. in `Refresh`  - which I would then need to call probably)

linev: For the moment there was no real need to run this event loop for `RBrowser`. 
For instance, if you will set timeout to -1, closed connection immediately will trigger `ClearOnClose` cleanup.
But be aware that pressing `Ctrl-R` in the web browser first close connection and then client will try reconnect again. So with negative timeout here it will not be possible. And with normal timeout one need to call `Sync` function regularly - if you are interesting in `ClearOnClose` cleanup.

Probably, `RWebBrowserImp` is proper place for implementing such timer which runs function each second.

silverweed: It's a bit unfortunate that we have to do this polling though - this was supposed to be a signal/callback (otherwise we might as well have a `IsConnectionOpen()` method that we can query, rather than `Sync()` which goes through additional hoops).
We could probably avoid that if we had a way to set this ClearOnClose object onto the Browser itself, no?

linev: Yes, we can provide extra method to check connections - and invoke `Sync()` inside this method.

silverweed (approved):

PR #19982, opened 2025-09-25
[draft] detecting close of RBrowser

linev:
Follow up for https://github.com/root-project/root/pull/19978

Show how `ClearOnClose` can be used for RBrowser.
To get signal immediately with browser close, 
one should change `gEnv->SetValue("WebGui.ReconnectTmout", -1);`

### DIFF
--- a/gui/browserv7/src/RWebBrowserImp.cxx
+++ b/gui/browserv7/src/RWebBrowserImp.cxx
@@ -20,6 +20,14 @@
 
 using namespace ROOT;
 
+class CleanupHandle {
+      RWebBrowserImp *fImp = nullptr;
+   public:
+
+      CleanupHandle(RWebBrowserImp *imp) { fImp = imp; }
+      ~CleanupHandle() { printf("CleanupHandle destroyed\n"); }
+};
+
 ////////////////////////////////////////////////////////////////////////////////////////
 /// Default constructor
 
@@ -29,6 +37,8 @@ RWebBrowserImp::RWebBrowserImp(TBrowser *b) : TBrowserImp(b)
 
    fWebBrowser = std::make_shared<RBrowser>();
    fWebBrowser->AddTCanvas();
+
+   fWebBrowser->ClearOnClose(std::make_shared<CleanupHandle>(this));
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


silverweed:
I'm not sure how to translate this for my use case: in rootbrowse.cxx I have this code:
```c++
int main() {
   // ... 
 
   auto browser = std::make_unique<TBrowser>();
  
   // For classic graphics: ensure rootbrowse quits when the window is closed
   if (auto imp = browser->GetBrowserImp()) {
      if (auto mainframe = imp->GetMainFrame()) {        
         mainframe->Connect("CloseWindow()", "TApplication", &app, "Terminate()");
      }
   }
  
  // TODO: do the same as the paragraph above but for the web graphics

   std::cout << "Press ctrl+c to exit.\n";
   while (!gROOT->IsInterrupted() && !gSystem->ProcessEvents()) {
      std::this_thread::sleep_for(std::chrono::milliseconds(10));
   }
   return 0;
}
```

Ideally I don't want to "know" about the internal `RBrowser` from here, I want to talk to `browser` or `browser->GetBrowserImp()`. 
I want to call `TApplication::Terminate()` when the `ClearOnClose` is triggered, but how do I do it from here? (without modifying the `TBrowserImp` interface further?)

linev:
> Ideally I don't want to "know" about the internal `RBrowser` from here, I want to talk to `browser` or `browser->GetBrowserImp()`.

That's why I add last commit with changes of `RWebBrowserImp` class.
It can assign `ClearOnClose` handle which is destroyed when last connection is closed.
And provide its own handler here.

So PR is incomplete and just shows the direction. 
One still need add some handler to base `TBrowserImp` class and invoke it in correspondent implementation properly.

Also one should be careful - `ClearOnClose` handle destroyed also during regular ROOT shutdown when instance of `RWebBrowserImp` can be already deleted.

silverweed:
> To get signal immediately with browser close, one should change `gEnv->SetValue("WebGui.ReconnectTmout", -1);`

Can't we add a `SetReconnectTmout` instead of having to pass through `TEnv` (which does an insane amount of work to Set/Get values compared to what could be in principle done by setting a single class member)? 
Also, as far as I understand it's currently impossible to change this timeout after having created the RWebBrowserImp, which is not ideal

PR #19985, opened 2025-09-25
[gui] Add browsable RNTupleItem

magnustymoteus:
# This Pull request:

## Changes or fixes:
This PR adds `RNTupleItem` class which overrides `Compare` method to ensure correct order in RBrowser (visualization element always appears before fields)

cc @silverweed 

## Checklist:

- [X] tested changes locally

### DIFF
--- a/gui/browsable/inc/ROOT/Browsable/RNTupleItem.hxx
+++ b/gui/browsable/inc/ROOT/Browsable/RNTupleItem.hxx
@@ -1,0 +1,56 @@
+/*************************************************************************
+ * Copyright (C) 1995-2025, Rene Brun and Fons Rademakers.               *
+ * All rights reserved.                                                  *
+ *                                                                       *
+ * For the licensing terms see $ROOTSYS/LICENSE.                         *
+ * For the list of contributors see $ROOTSYS/README/CREDITS.             *
+ *************************************************************************/
+
+#ifndef ROOT7_Browsable_RNTupleItem
+#define ROOT7_Browsable_RNTupleItem
+
+#include <ROOT/Browsable/RItem.hxx>
+
+namespace ROOT {
+namespace Browsable {
+
+/** \class RNTupleItem
+\ingroup rbrowser
+\brief Representation of an RNTuple item in the browser
+\author Patryk Tymoteusz Pilichowski
+*/
+
+class RNTupleItem : public RItem {
+public:
+   enum ECategory {
+      kField,
+      kVisualization
+   };
+
+   RNTupleItem() = default;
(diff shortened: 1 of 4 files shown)


gui/browsable/inc/ROOT/Browsable/RNTupleItem.hxx
```diff
@@ -0,0 +1,49 @@
...
+
+/** \class RNTupleItem
+\ingroup rbrowser
+\brief Representation of an RNTuple item in the browser
+\author Patryk Tymoteusz Pilichowski
+*/
+
+class RNTupleItem : public RItem {
+protected:
+   bool visualization{false}; ///< is visualization item
```
silverweed: For the sake of future additions, I'd make this into an enum (ECategory?)

gui/browsable/inc/ROOT/Browsable/RNTupleItem.hxx
```diff
@@ -0,0 +1,49 @@
...
+   RNTupleItem(const std::string &_name, int _nchilds = 0, const std::string &_icon = "", bool _visualization = false)
+      : RItem(_name, _nchilds, _icon), visualization(_visualization)
+   {
+   }
+   // must be here, one needs virtual table for correct streaming of sub-classes
+   virtual ~RNTupleItem() = default;
+
+   bool IsVisualization() const { return visualization; }
+
+   virtual bool Compare(const RItem *b, const std::string &s) const override
```
silverweed:
```suggestion
   bool Compare(const RItem *b, const std::string &s) const override
```

gui/browsable/inc/ROOT/Browsable/RNTupleItem.hxx
```diff
@@ -0,0 +1,49 @@
...
+   {
+      auto tuple_b = dynamic_cast<const RNTupleItem *>(b);
```
silverweed, line 43: I believe that if `this->IsVisualization()` then we can automatically declare this the "winner" (meaning it should compare smaller than `b`). This is because we only expect to have 1 Visualization item and it should always be displayed before any other category. So probably you can just do:
```c++
return IsVisualization() || RItem::Compare(b, s);
```

magnustymoteus: This doesn't work for all RNTuples. The visualization RNTupleItem would always have to be the first one in an iteration for this to work, which doesn't seem to be the case

silverweed: I guess it should actually be 
```c++
if (IsVisualization()) return true;
if (b->IsVisualization()) return false;
return RItem::Compare(b, s);
```
if we assume that `b` is always a RNTupleItem. But let's not assume that, just in case, and keep your version.

Issue #19965, opened 2025-09-23
thisroot.sh fails on macos when called from script with #!/bin/sh


wright20:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Use case is a shell script with shebang #!/bin/sh that executes thisroot.sh prior to running root commands within the script.

If the mac user has chosen bash as the default shell, then thisroot.sh incorrectly identifies the shell as "sh" instead of "bash" and runs the wrong setup commands, resulting in an error.

On macos /bin/sh is a binary that re-execs itself to another shell. See the manpage on macos

sh is a POSIX-compliant command interpreter (shell).  It is implemented by re-execing as either bash(1), dash(1), or
     zsh(1) as determined by the symbolic link located at /private/var/select/sh.

### Reproducer

create file runroot.sh (does nothing except setup root environment inside the script):

#!/bin/sh
source path-to-root/bin/thisroot.sh

execute the shell script:

runroot.sh

produces error message:

...root_v6.32.16/bin/thisroot.sh-orig: line 253: root-config: command not found

### ROOT version

%60%60%60%0AROOT v6.32.16%0ABuilt for macosxarm64 on Sep 17 2025, 02:40:39%0AFrom tags%2F6-32-16@6-32-16%0AWith Apple clang version 17.0.0 (clang-1700.3.19.1)%0ABinary directory: %2FUsers%2Fwright20%2Fcern%2Froot_v6.32.16%2Fbin%0A%60%60%60

### Installation method

build from source

### Operating system

MacOS 15.7

### Additional context

I think the error is due to getTrueShellExeName() function which returns "sh" as the SHELLNAME which is then used on line 199 to decide how to find the path to thisroot.sh. Instead of the "bash" method (which works) it assumes "dash or ksh" method using $(lsof ...) which fails.

ferdymercury:
Could you retry with Root master? There were changes recently in this script.
You Just need to copy paste the thisroot script no need to reinstall everything

wright20:
Testing with https://github.com/root-project/root/blob/master/config/thisroot.sh
Copied to local laptop as thisroot.sh-master

Same error:
./thisroot.sh-master: line 255: root-config: command not found

ferdymercury:
Thanks for checking.

Can you try:
`. thisroot.sh-master`
instead of
`./thisroot.sh-master`
?

I cannot reproduce the problem on Linux, so might be a mac issue.

[...]

PR #19987, opened 2025-09-25
[thisroot] resolve reexec sh on macos

ferdymercury:
Fixes https://github.com/root-project/root/issues/19965

@wright20 could you check if this solves the issue?

### DIFF
--- a/config/thisroot.sh
+++ b/config/thisroot.sh
@@ -190,16 +190,10 @@ getTrueShellExeName() { # mklement0 https://stackoverflow.com/a/23011530/7471760
    # target.
    while nextTarget=$(readlink "$trueExe"); do trueExe=$nextTarget; done
    
-   if [ "$(basename "$trueExe")" = "sh" ]; then
-      # on MacOS, sh is NOT a symlink but an executable, so trueExe is 'sh' but it will
-      # re-exec according to symlink at /private/var/select/sh, so resolve it
-      if [ -L "/private/var/select/sh" ] && [ -e "/private/var/select/sh" ] ; then
-         trueExe="/private/var/select/sh";
-         while nextTarget=$(readlink "$trueExe"); do trueExe=$nextTarget; done
-      else
-         # otherwise default to bash
-         trueExe="/bin/bash"
-      fi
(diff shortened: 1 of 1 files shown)


config/thisroot.sh
```diff
@@ -189,6 +189,18 @@ getTrueShellExeName() { # mklement0 https://stackoverflow.com/a/23011530/7471760
...
+      # on MacOS, sh is NOT a symlink but an executable, so trueExe is 'sh' but it will
+      # re-exec according to symlink at /private/var/select/sh, so resolve it
+      if [ -L "/private/var/select/sh" ] && [ -e "/private/var/select/sh" ] ; then
+         trueExe="/private/var/select/sh";
+         while nextTarget=$(readlink "$trueExe"); do trueExe=$nextTarget; done
+      else
+         # otherwise default to bash
+         trueExe="/bin/bash"
+      fi
+   fi
```
hageboeck: I tested on mac, and this seems to solve the problem:
```suggestion
   if [ "$(basename $trueExe)" = "sh" -a -L "/private/var/select/sh" ]; then
      # on MacOS, sh re-execs /private/var/select/sh, so resolve it
      trueExe="/private/var/select/sh";
      while nextTarget=$(readlink "$trueExe"); do trueExe=$nextTarget; done
   fi
```

hageboeck: I would not include the fallback to bash, because having a link at `/private/var/select` has to be resolvable. Otherwise, macOS would be very broken.

ferdymercury: Ok. But see sh manual:

If /private/var/select/sh does not exist or
does not point to a valid shell, sh will use one of the supported shells.

hageboeck: I see. In that case, sometimes it might work, and other times it might not. Having or not having the fallback both seem to be not-so-perfect solutions, wouldn't you say?

ferdymercury: Yep, agreed. I'd say like this is fine, since it covers 99.9% of the cases and we reduce the number of lines in the code.

hageboeck:
I tested this change + the one in #19974, and I'm able to source a symlink both in sh and bash now.

hageboeck (approved):

Issue #10423, opened 2022-04-20
Reading XYZVectors with TTreeReader is broken in some cases


eguiraud:
Reproducer:

```cpp
#include <Math/Vector3D.h>
#include <ROOT/RDataFrame.hxx>
#include <vector>
#include <TClass.h>

int main() {
  {
    TFile f("f.root", "recreate");
    TTree t("t", "t");
    ROOT::Math::XYZVector x(1, 2, 3);
    std::vector<ROOT::Math::XYZVector> y{ROOT::Math::XYZVector(4,5,6)};
    t.Branch("x", &x);
    // original line:
    //     t.Branch("y", &y); // commenting this line "fixes" the problem!
    // Actual trigger:
    auto c = TClass::GetClass("std::vector<ROOT::Math::XYZVector>");  // commenting this line "fixes" the problem!
    t.Fill();
    t.Write();
  }

  TFile f("f.root");
  TTreeReader r("t", &f);
  TTreeReaderValue<ROOT::Math::XYZVector> rx(r, "x");
  r.Next();
  *rx;
}
```
yields:

```
Error in <TTreeReaderValueBase::CreateProxy()>: The branch x contains data of type ROOT::Math::DisplacementVector3D<ROOT::Math::Cartesian3D<Double32_t>,ROOT::Math::DefaultCoordinateSystemTag>. It cannot be accessed by a TTreeReaderValue<ROOT::Math::DisplacementVector3D<ROOT::Math::Cartesian3D<double>,ROOT::Math::DefaultCoordinateSystemTag>>
Error in <TTreeReaderValue::Get()>: Value reader not properly initialized, did you call TTreeReader::Set(Next)Entry() or TTreeReader::Next()?
```

Interestingly, if branch `"y"` is not written out, then branch `"x"` can be read correctly.

dudarboh:
@pcanal, @eguiraud, @Axel-Naumann,

Dear all, could I ask if there is any progress on this?
Looking at the code, I found that somewhat similar problem has already existed:
https://sft.its.cern.ch/jira/browse/ROOT-8731

Wouldn't the solution be straightforward? Maybe I can try to contribute?

cheers

Axel-Naumann:
@pcanal do we really need to teach TTreeReader to check for schema evolution between source and target, or is there maybe a better question that TTreeReader can ask, better than "is source and target class the same?"

ferdymercury:
[...]

PR #19988, opened 2025-09-26
[test] Add test for vector of XYZVector in TTreeReaderValue

ferdymercury:
Closes https://github.com/root-project/root/issues/10423
by adding a test to it.

The issue seems solved in the meanwhile.


dpiparo:
Thanks a lot for this test.

PR #19991, opened 2025-09-26
[ntuple] Optimize bulk read for (RVec of) array of PoD

### LINKED ISSUE #16892: [ntuple] Add optimized bulk read of fixed-size arrays

The RArrayField should provide an optimized `ReadBulkImpl`, at least for simple types. That would allow bulk-reading of fixed-size arrays without the mask evaluation. Requested by ALICE.

We should also add bulk read optimization to the cardinality fields (possibly to be moved to another issue).

### ROOT version

master

### Installation method

n/a

### Operating system

n/a

### Additional context

_No response_

jblomer:
Fixes #16892

@ktf FYI

### DIFF
--- a/tree/ntuple/inc/ROOT/RNTupleTypes.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleTypes.hxx
@@ -163,6 +163,11 @@ public:
       return RNTupleLocalIndex(fClusterId, fIndexInCluster - off);
    }
 
+   RNTupleLocalIndex operator*(ROOT::NTupleSize_t repetitionFactor) const
+   {
+      return RNTupleLocalIndex(fClusterId, fIndexInCluster * repetitionFactor);
+   }
+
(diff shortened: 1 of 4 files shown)


tree/ntuple/inc/ROOT/RNTupleTypes.hxx
```diff
@@ -163,6 +163,11 @@ public:
       return RNTupleLocalIndex(fClusterId, fIndexInCluster - off);
    }
 
+   RNTupleLocalIndex operator*(ROOT::NTupleSize_t scaleFactor) const
```
pcanal: Wouldn't
```suggestion
   RNTupleLocalIndex operator*(ROOT::NTupleSize_t repetitionFactor) const
```
be closer to the semantic?

pcanal (approved):
LGTM.

PR #19997, opened 2025-09-28
[ntuple] automatic schema evolution of fixed-size array to std::vector

jblomer:


tree/ntuple/src/RFieldSequenceContainer.cxx
```diff
@@ -814,36 +830,39 @@ void ROOT::RField<std::vector<bool>>::GenerateColumns()
 
 void ROOT::RField<std::vector<bool>>::GenerateColumns(const ROOT::RNTupleDescriptor &desc)
 {
-   GenerateColumnsImpl<ROOT::Internal::RColumnIndex>(desc);
+   if (fOnDiskNRepetitions == 0)
```
hahnjo, line 807: Should this `assert` because writing should only be called with `fOnDiskNRepetitions == 0`?

tree/ntuple/src/RFieldSequenceContainer.cxx
```diff
@@ -649,6 +649,28 @@ void ROOT::RVectorField::ReadGlobalImpl(ROOT::NTupleSize_t globalIndex, void *to
    }
 }
 
+void ROOT::RVectorField::ReadInClusterImpl(ROOT::RNTupleLocalIndex localIndex, void *to)
```
hahnjo: curious: Why do we need this? Since we have an index column, the default implementation of `RFieldBase::ReadInClusterImpl` should be able to do the translation to a global index.

jblomer: Yes, we don't necessarily need this. It is a small performance improvement not to translate between global and cluster-local indexing (i.e., we already know the cluster in which we operate and can use that information in ReadInClusterImpl(). Now that we have factored out much of the heavy logic, it seemed not to cumbersome to add this override. But we can also remove it.

tree/ntuple/src/RFieldSequenceContainer.cxx
jblomer: You mean assert in the other overload of `GenerateColumns()`?

hahnjo: Ah yes. I was looking for the overload that is called during writing to have the equivalent of the check that was in `RArrayAsBoolVectorField`

tree/ntuple/src/RFieldSequenceContainer.cxx
hahnjo: Ok, your call. In the end, I would be surprised if this contributes much because the index calculations are purely in memory IIRC...

jblomer: Perhaps you're right. Let's try with less code.

tree/ntuple/src/RFieldSequenceContainer.cxx
jblomer: I added an assert to ensure `fOnDiskNRepetitions` is only used for reading

hahnjo (approved):
LGTM, thanks!

PR #19999, opened 2025-09-29
[Python][RF] Introduce `Klass::contains` -> `Klass.__contains__` pythonization and use it in RooFit

### LINKED ISSUE #19986: [PyROOT] Proper way to check if snapshot exists in workspace

In C++ ROOT, one can check if a snapshot exists in a workspace by checking if `<workspace>.getSnapshot(<snapshot_name>)` returns a null pointer.

However in pyROOT, there's no standard way to check if an object is a null pointer. Take for example the following:

```python
import ROOT

ws = ROOT.RooWorkspace('ws')
ws.saveSnapshot('empty_snapshot', ROOT.RooArgSet())

empty_snapshot = ws.getSnapshot('empty_snapshot')
non_existent_snapshot = ws.getSnapshot('non_existent_snapshot')

def naive_has_snapshot(snapshot: "ROOT.RooArgSet") -> bool:
    if not snapshot:
        return False
    return True


# returns False
naive_has_snapshot(empty_snapshot)
# returns False
naive_has_snapshot(non_existent_snapshot)
```

There will be no difference between a non-existent snapshot or an empty snapshot.

A workaround will be to check also the repr of the object:
```python
def proper_has_snapshot(snapshot: "ROOT.RooArgSet") -> bool:
    if (not snapshot) and ('0x(nil)' in empty_snapshot.__repr__()):
        return False
    return True
    
# returns True
proper_has_snapshot(empty_snapshot)
# returns False
proper_has_snapshot(non_existent_snapshot)
```

But it is cumbersome and non-trivial for most users.

Would there be a better way to check if a snapshot exists in a workspace? Like adding a method `<workspace>.hasSnapshot(<snapshot_name>)` ?

### ROOT version

ROOT 6.34.04

### Installation method

conda

### Operating system

Linux

### Additional context

_No response_

guitargeek:
This is inspired by the existing `Klass::size` -> `Klass.__len__` pythonization, lifting another STL-typical C++ method to a magic Python method for the sake of Pythonization.

This PR enables for example checking for the existence of a specific snapshot in a RooWorkspace.

```python
import ROOT

name_1 = 'empty_snapshot'
name_2 = 'non_existent_snapshot'


ws = ROOT.RooWorkspace('ws')
ws.saveSnapshot(name_1, ROOT.RooArgSet())

empty_snapshot = ws.getSnapshot(name_1)
non_existent_snapshot = ws.getSnapshot(name_2)

ws.Print()

print(name_1 in ws.getSnapshots())
print(name_2 in ws.getSnapshots())
```

Test coverage is provided by the cppyy unit tests.

Closes #19986.


aaronj0 (approved):
The changes on the cppyy side look good to me! Thanks for this improvement in the `contains()` pythonizations utilizing the C++ implementation

guitargeek:
Thanks for the review! As the change to RooFit was also seen as useful and intuitive by @AlkaidCheng who opened the issue, I'll merge.

PR #20001, opened 2025-09-29
[RF] Expose again interface to write NLL codegen debug macro

guitargeek:
This experimental public interface was so far implemented in the
`RooFuncWrapper` class, but it got lost from the public interface when
that class was merged into the implementation details of the
`RooEvaluatorWrapper` (in commit https://github.com/root-project/root/commit/5b9d51df7f5963b15772a6da9a8f5eaeef53c981).

Writing the codegen debug macros is however quite useful for developing,
so this commit brings back that functionality as a free function in the
`RooFit::Experimental` namespace.


vgvassilev (approved):
Lgtm!

PR #20003, opened 2025-09-29
[CMake] More cleanups to libCling linking

hahnjo:


core/metacling/src/CMakeLists.txt
```diff
@@ -130,6 +130,8 @@ ROOT_LINKER_LIBRARY(Cling
         $<TARGET_OBJECTS:clangCppInterOp>
         LIBRARIES ${CLING_LIBRARIES} ${LINK_LIBS} ${CLING_PLUGIN_LINK_LIBS})
 
+target_link_libraries(Cling PRIVATE Core RIO)
```
pcanal, line 133: I suspect that the reason that we did not do this (eventhough it is probably required as Windows needs it) is the fact that there is a potential risk of (at least semantic) issue if `libCling` is loaded before `libCore` is fully loaded (because `libCore` could then trigger an attempt to load `libCling`) but as one of the commit essentially says `nobody should be doing that anyway` (loading or linking directly against `libCling`)

PR #20006, opened 2025-09-29
[rootx] Turn `root` executable into symbolic link to `root.exe`

guitargeek:
There is no need to have different executables anymore that start each
other anymore, but for backwards compatibility we keep `root.exe` also
for macOS and Linux. Before, `root` was launching `root.exe` after doing
the following things:

  1. In the past, show the splash screen
     (already removed in 6.36 with commit https://github.com/guitargeek/root/commit/4042121fd6b6308bc3f714e24c1856f938677a1a)
  2. In the past, set `(DY)LD_LIBRARY_PATH`
     (already removed in 6.38 with commit https://github.com/guitargeek/root/commit/daf50eeaca9dde84918e13c20a6e0203e3014ada)
  3. Set `ROOTSYS`

The value of `ROOTSYS` can be inferred from the opened shared libraries
(a mechanism introduced in commit ????), which can always correctly be
opened because the `RPATH` is now always set (as of commit 26d24de).

The `root.exe` was launched with the full path. That's also not needed,
because `root` and `root.exe` were in the same directory anyway. So if
`root` was in the path already, `root.exe` will be too.

In conclusion, there is no need anymore for a separate "wrapper"
executable and this commit suggests to make the former `root` wrapper
simply a copy of `root.exe`.

This is similar to what was [done for rootcling](https://github.com/root-project/root/pull/19769). The actual executable is small anyway, being linked against the ROOT shared libraries.


rootx/src/rootx.cxx
```diff
@@ -1,269 +0,0 @@
...
-         kill(gChildpid, SIGCONT);       // if parent wakes up, wake up child
-      }
-   } while (WIFSTOPPED(status));
-
-   exit(0);
-}
-
-static void PrintUsage()
-{
-   fprintf(stderr, kCommandLineOptionsHelp);
```
ferdymercury: Shouldn't this Help Messages be moved now to root.exe?

ferdymercury (changes requested):
Thanks a lot!

Does this close https://its.cern.ch/jira/browse/ROOT-10697 ?

Please also fix these messages:
```
core/base/src/TApplication.cxx:         fprintf(stderr, "ROOT splash screen is not visible with root.exe, use root instead.\n");
roofit/histfactory/config/prepareHistFactory:# we will switch here to use root.exe as it is done for the whole ROOT.
```

And the lost usage-help part.

Potentially add this change to the release-notes 6.38

rootx/src/rootx.cxx
guitargeek: It is already there too:
https://github.com/root-project/root/blob/master/core/base/src/TApplication.cxx#L360

rootx/src/rootx.cxx
```diff
@@ -1,269 +0,0 @@
...
-int main(int argc, char **argv)
-{
-   char **argvv;
-   char  arg0[kMAXPATHLEN];
-
-#ifdef ROOTPREFIX
-   if (std::getenv("ROOTIGNOREPREFIX")) {
-#endif
-   // Try to set ROOTSYS depending on pathname of the executable
-   SetRootSys();
```
pcanal, line 164: There are still a few places in the code that looks up `ROOTSYS` (and thus could potentially lead to a difference in behavior between `root` and `root.exe`).  Some (eg the one in `TUnixSystem.cxx`) have additional search to find the 'right' value when not set) but (seemingly) not all. 

As part of this PR, we need to review those later ones and either remove if the result is not used or annotated them as needed.

guitargeek: Absolutely. I should mark this PR as draft to make clear that more work is needed.

pcanal:
> is a symbolic link to it.

Rather than this technical choice, we may want to just copy/duplicate.  We just replaced the symbolic link genreflex/rootcling by a copy.

hageboeck:
> > is a symbolic link to it.
> 
> Rather than this technical choice, we may want to just copy/duplicate. We just replaced the symbolic link genreflex/rootcling by a copy.

With debug symbols, they are < 100k, so this would be a viable choice.

guitargeek:
About how `ROOTSYS` could still be used, I'm suggesting to not set `ROOTSYS` anymore in our tests:
https://github.com/root-project/root/pull/20408

Like this, our tests have a higher change of spotting problems if setting `ROOTSYS` in the wrapping `root` executable is missing.

ferdymercury:
`roottest/root/rint/CMakeLists.txt:                  COMMAND $<TARGET_FILE:root> -b -q -l -e ".libraries"` might need `root.exe` instead of `root`?

ferdymercury (commented):
thanks!!
maybe update PR title according to new copy behavior

PR #20007, opened 2025-09-29
[ntuple] finalize automatic schema evolution between collections

jblomer:


tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -882,12 +882,34 @@ ROOT::RMapField::RMapField(std::string_view fieldName, EMapType mapType, std::un
...
 }
 
+std::unique_ptr<ROOT::RFieldBase> ROOT::RSetField::CloneImpl(std::string_view newName) const
+{
+   return std::make_unique<RSetField>(newName, fSetType, fSubfields[0]->Clone(fSubfields[0]->GetFieldName()));
+}
+
+void ROOT::RSetField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
+{
+   static const std::vector<std::string> prefixesRegular = {"std::set<", "std::unordered_set<", "std::map<"};
```
hahnjo: I guess we discussed this, but I forgot the reasoning again, sorry: Why can we read into a `kUnorderedSet` from a `std::map`, but not from a `std::unordered_map`?

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -871,14 +871,35 @@ void ROOT::RProxiedCollectionField::AcceptVisitor(ROOT::Detail::RFieldVisitor &v
...
 }
 
+std::unique_ptr<ROOT::RFieldBase> ROOT::RMapField::CloneImpl(std::string_view newName) const
+{
+   return std::make_unique<RMapField>(newName, fMapType, fSubfields[0]->Clone(fSubfields[0]->GetFieldName()));
+}
+
+void ROOT::RMapField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
+{
+   static const std::vector<std::string> prefixesRegular = {"std::map<", "std::unordered_map<", "std::set<"};
```
hahnjo: similarly here: `std::unordered_set`?

tree/ntuple/src/RFieldMeta.cxx
jblomer: I think from the point of view of I/O we could. But to make it happen, we would need an `unordered_set<pair>`, which is missing a default hash function...

pcanal: > from a `std::map`, but not from a `std::unordered_map`
> we would need an `unordered_set<pair>`, which is missing a default hash function...

Why is this an issue for a `std::unordered_map`?

And is it a fundamental issue or just an issue for the test?  I.e. if a user had written a `map` and wanted to read it into a `unordered_set` they would have have provided the necessary hash overload, isn't it?

jblomer: Solved, supported and tested with a custom hash for `std::pair<(short) int, (short) int>`.

PR #20008, opened 2025-09-29
[ntuple] automatic schema evolution from streamer to class field

jblomer:


tree/ntuple/test/rfield_streamer.cxx
```diff
@@ -350,6 +350,36 @@ TEST(RField, StreamerSchemaEvolution)
...
+      auto writer = RNTupleWriter::Recreate(std::move(model), "ntpl", fileGuard.GetPath());
+      auto ptr = writer->GetModel().GetDefaultEntry().GetPtr<CustomStreamer>("f");
+      ptr->a = 3.14;
+      writer->Fill();
+   }
+
+   auto cl = TClass::GetClass("CustomStreamer");
+   ASSERT_TRUE(cl != nullptr);
+   cl->CreateAttributeMap();
+   cl->GetAttributeMap()->AddProperty("rntuple.streamerMode", "false");
```
hahnjo: Can we use a different class for this? Because the attributes stick around in the process, this may cause issues if tests are reordered either in code or with `--gtest_shuffle`.

hahnjo (approved):
LGTM with a comment on the test

PR #20012, opened 2025-09-30
[CMake] Fix an RPath issue

hageboeck:
Usually, CMake automatically sets RUNPATH when target_link_libraries is
used. However, since Core doesn't have other ROOT dependencies, its
RUNPATH might only contain external dependencies or remain empty.
To enable Core to find other ROOT libraries, its current location is
added to its BUILD_RPATH.
If this isn't done, dlopen might find ROOT libraries which are installed in system locations.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -26,6 +26,14 @@ add_custom_target(rconfigure ALL DEPENDS ${CMAKE_BINARY_DIR}/include/RConfigure.
 
 ROOT_LINKER_LIBRARY(Core BUILTINS LZMA)
 
+# This sets RUNPATH for libCore, which enables it to dlopen any ROOT library in its directory.
+# If omitted, Core might try to load libraries from a ROOT that's installed in the system.
+if(APPLE)
+  set_property(TARGET Core APPEND PROPERTY BUILD_RPATH @loader_path)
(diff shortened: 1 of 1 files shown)


guitargeek:
I have solved a similar problem in 47caba5 for TROOT, so `TROOT::GetLibDir()` should give you the right path already. Maybe you can re-use that?

core/CMakeLists.txt
```diff
@@ -26,6 +26,14 @@ add_custom_target(rconfigure ALL DEPENDS ${CMAKE_BINARY_DIR}/include/RConfigure.
 
 ROOT_LINKER_LIBRARY(Core BUILTINS LZMA)
 
+# This sets RUNPATH for libCore, which enables it to dlopen any ROOT library in its directory.
```
guitargeek, line 29: This is a very good addition in any case :+1:

hageboeck:
> I have solved a similar problem in [47caba5](https://github.com/root-project/root/commit/47caba598a1529cbbacc170a0fb6c0d139bd352c) for TROOT, so `TROOT::GetLibDir()` should give you the right path already. Maybe you can re-use that?

@guitargeek I removed the changes to the path searching from this PR. I would need more time to rework this.

hageboeck:

PR #20019, opened 2025-10-01
[core] Reduce size of libCling

hahnjo:
Garbage collect unused sections and strip the library after build, reducing its size from 120 MB to 88 MB in a `Release` configuration with GCC. See the individual commit messages for more details.

FYI @vepadulano maybe interesting for the Python wheels, though we will grow in size again with each LLVM upgrade...

### DIFF
--- a/core/metacling/src/CMakeLists.txt
+++ b/core/metacling/src/CMakeLists.txt
@@ -236,6 +236,10 @@ if(NOT APPLE AND NOT MSVC)
   # After inlining and hiding symbols, some of them are not referenced anymore
   # and can be collected by the linker, reducing the library size.
   target_link_libraries(Cling PRIVATE -Wl,--gc-sections)
+  # Additionally, for Release builds, strip the binary to remove symbol tables.
+  add_custom_command(TARGET Cling POST_BUILD
+    COMMAND $<$<CONFIG:Release>:${CMAKE_STRIP}>
+    ARGS $<TARGET_FILE:Cling>)
 endif()
 
 if (CMAKE_SYSTEM_NAME MATCHES FreeBSD)


vgvassilev (approved):
Lgtm!

We might be able to do better because even if libCling is a shared library it is self contained. That is we don’t expose any of these llvm interfaces outside. I am not sure if we can express that to the linker but practically we can strip a lot of symbols as if we were building a binary.

hahnjo:
> We might be able to do better because even if libCling is a shared library it is self contained. That is we don’t expose any of these llvm interfaces outside. I am not sure if we can express that to the linker but practically we can strip a lot of symbols as if we were building a binary.

Hiding symbols, garbage collecting unused stuff and `strip` already does this effectively.

vgvassilev:
> > We might be able to do better because even if libCling is a shared library it is self contained. That is we don’t expose any of these llvm interfaces outside. I am not sure if we can express that to the linker but practically we can strip a lot of symbols as if we were building a binary.
> 
> Hiding symbols, garbage collecting unused stuff and `strip` already does this effectively.

I believe if I build some binary on top of llvm built with hidden visibility I will get the overall size to be smaller. In addition, it seems some of the linked in JIT components might not be needed. Forgot what those were but I am surprised that if the linker does not find them.

vepadulano (approved):
Nice ideas, thanks!

core/metacling/src/CMakeLists.txt
```diff
@@ -236,6 +236,10 @@ if(NOT APPLE AND NOT MSVC)
   # After inlining and hiding symbols, some of them are not referenced anymore
   # and can be collected by the linker, reducing the library size.
   target_link_libraries(Cling PRIVATE -Wl,--gc-sections)
+  # Additionally, for Release builds, strip the binary to remove symbol tables.
+  add_custom_command(TARGET Cling POST_BUILD
+    COMMAND $<$<CONFIG:Release>:${CMAKE_STRIP}>
+    ARGS $<TARGET_FILE:Cling>)
```
hageboeck, line 242: If I'm not misreading this, a not-release build would see the following command:
```
"" path/to/libCling.so
```
because the generator expression would expand to empty.

hahnjo: Yes, and this seems supported to not have the command run at all:
 * https://github.com/Kitware/CMake/blob/f2afbff2c48a0479cc129c8eb2cc40e9181f0c51/Source/cmLocalNinjaGenerator.cxx#L598
 * https://github.com/Kitware/CMake/blob/f2afbff2c48a0479cc129c8eb2cc40e9181f0c51/Source/cmLocalVisualStudioGenerator.cxx#L227
 * https://github.com/Kitware/CMake/blob/f2afbff2c48a0479cc129c8eb2cc40e9181f0c51/Source/cmLocalVisualStudioGenerator.cxx#L229

I'm struggling to find a better way; we could wrap the argument as `$<$<CONFIG:Release>:$<TARGET_FILE:Cling>>`, but as far as I understand whitespace in generator expressions leads to very surprising results as well...

PR #20020, opened 2025-10-01
[Python][UHI] Integrate the UHI testing suite

siliataider:
# This Pull request:

## Changes or fixes:
- This PR integrates the [UHI indexing test suite](https://uhi.readthedocs.io/en/latest/testing.html) for ROOT histograms by adding `Indexing1D`,  `Indexing2D` and `Indexing3D` test helpers specialized for `ROOT.TH1` histograms.

- For all tests to pass, a temporary equality patch (`__eq__)` for `ROOT.TH1` within the test context is implemented checking histogram values instead of default object identity.

- Since serialization is not yet implemented, the histograms are manually created and set in `make_histogram`.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/bindings/pyroot/pythonizations/test/uhi_test_suite.py
+++ b/bindings/pyroot/pythonizations/test/uhi_test_suite.py
@@ -1,0 +1,90 @@
+from __future__ import annotations
+
+import numpy as np
+import pytest
+import ROOT
+import uhi.testing.indexing
+from ROOT._pythonization._uhi import _temporarily_disable_add_directory
+
+
+def th1_eq(self, other):
+    if self is other:
+        return True
+    return (
+        isinstance(other, type(self))
+        and self.kind == other.kind
+        and np.array_equal(self.values(), other.values())
+        and np.array_equal(self.counts(), other.counts())
+        and np.array_equal(self.variances(), other.variances())
+        and all(a == b for a, b in zip(self.axes, other.axes))
+    )
+
+
+@pytest.fixture(autouse=True)
+def patch_th1_eq(monkeypatch):
+    """
+    Temporarily patch ROOT.TH1.__eq__ for tests
+    Since __eq__ pythonizations are not defined, the comparison
+    would fall back to object identity (`is`) and would fail
+    """
+    monkeypatch.setattr(ROOT.TH1, "__eq__", th1_eq)
+
+
+class TestIndexing1D(uhi.testing.indexing.Indexing1D[ROOT.TH1D]):
+    tag = ROOT.uhi
+
+    @classmethod
+    def make_histogram(cls) -> ROOT.TH1D:
+        """
+        Return a 1D histogram with 10 bins from 0 to 1. Each bin value increases
+        by 2 compared to the previous one, starting at 0. The underflow bin is 3,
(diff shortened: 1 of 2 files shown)


bindings/pyroot/pythonizations/test/uhi_test_suite.py
```diff
@@ -0,0 +1,67 @@
...
+def patch_th1_eq(monkeypatch):
+    """
+    Temporarily patch ROOT.TH1.__eq__ for tests
+    Since __eq__ pythonizations are not defined, the comparison
+    would fall back to object identity (`is`) and would fail
+    """
+    monkeypatch.setattr(ROOT.TH1, "__eq__", th1_eq)
+
+
+class TestIndexing1D(uhi.testing.indexing.Indexing1D[ROOT.TH1]):
```
vepadulano: Minor thing: It looks weird to have `ROOT.TH1` as the input class for all three dimensions tests. I wonder if it would still work if we had `ROOT.TH2` for the Indexing2D and `ROOT.TH3` for the Indexing3D?

bindings/pyroot/pythonizations/test/uhi_test_suite.py
```diff
@@ -0,0 +1,67 @@
...
+    tag = ROOT.uhi
+
+    @classmethod
+    def make_histogram(cls) -> ROOT.TH1D:
+        with _temporarily_disable_add_directory():
+            h = ROOT.TH1D("h", "h", 10, 0, 1)
+            h[...] = np.array([3.0, 0.0, 2.0, 4.0, 6.0, 8.0, 10.0, 12.0, 14.0, 16.0, 18.0, 1.0])
```
vepadulano, line 47: Are the values arbitrary, or should they follow any pattern? If they can be arbitrary, what is the testing suite actually testing?

vepadulano (changes requested):
Super nice thanks! I believe the test should also be added to the CMake bindings/pyroot/pythonizations/test/CMakeLists.txt so that it gets picked up by the CI

bindings/pyroot/pythonizations/test/uhi_test_suite.py
siliataider: The values are in fact not arbitrary: for each histogram you get a description in the base class of the number of bins and expected values ([1D](https://github.com/scikit-hep/uhi/blob/main/src/uhi/testing/indexing.py#L68), [2D](https://github.com/scikit-hep/uhi/blob/main/src/uhi/testing/indexing.py#L391), and [3D](https://github.com/scikit-hep/uhi/blob/main/src/uhi/testing/indexing.py#L591))

vepadulano: Ok thanks! Can you add a permanent link to that docstring for each class?

Looking at the source code you link, I realise that the `T` type, which is the one specified at the class level, should also be the same `T` type returned by `make_histogram`. So either the return type is changed to `ROOT.TH1` or the `T` is changed to `ROOT.TH1D`, for proper type correctness.

vepadulano (approved):
LGTM! Feel free to merge once the CI is green

PR #20022, opened 2025-10-01
[cling][test] Fix failing cling tests in master

devajithvs:
# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes #


interpreter/cling/test/Prompt/ValuePrinter/SourceLocation.C
```diff
@@ -6,24 +6,18 @@
 // LICENSE.TXT for details.
 
 //------------------------------------------------------------------------------
-// RUN: cat %s | %cling | FileCheck %s
+// This test printing of std::source_location, which is a C++20 feature.
+// Running it with an older language standard is meaningless and brittle to test
+// Explicitly enforce -std=c++20 here.
+// RUN: cat %s | %cling -std=c++20 | FileCheck %s
```
hahnjo: We are not guaranteed that the C++ library actually supports C++ 20; for example, on AlmaLinux 8 it does not...

devajithvs: I see, I will revert that. There are some issues in cling when running preprocessor conditionals:
```C++
#include <iostream>
#define SOMETHING 1
#ifndef SOMETHING
std::cout << "Shouldn't run!";
#else
void getsrcloc() { return; }
#endif
```

This fails with: 
```bash
input_line_13:5:18: error: function definition is not allowed here
void getsrcloc() { return; }
                 ^
```

I'm looking into this. The above code doesn't run on clang-repl either.

hahnjo: You may need `.rawInput` I think

devajithvs: Thanks!

interpreter/cling/test/Prompt/ValuePrinter/SourceLocation.C
```diff
@@ -8,22 +8,27 @@
 //------------------------------------------------------------------------------
 // RUN: cat %s | %cling | FileCheck %s
```
hahnjo: Should this still specify `-std=c++20`?

interpreter/cling/test/Prompt/ValuePrinter/SourceLocation.C
```diff
@@ -8,22 +8,27 @@
...
 }
-getsrcloc()
+#else
+// Hack to prevent failure if __cpp_lib_source_location feature does not exist!
+void getsrcloc() {
+  std::cout << "(std::source_location) ";
+  std::cout << "CHECK_SRCLOC:42:std::source_location getsrcloc()\n";
+}
 #endif
+.rawInput 0
```
hahnjo, line 31: Just for curiosity: We need to always wrap it into a function because we need `.rawInput 0` before printing the return value?

hahnjo (approved):

interpreter/cling/test/Prompt/ValuePrinter/SourceLocation.C
devajithvs: Yes, exactly. Value printing only happens when we switch back to `.rawInput 0`

interpreter/cling/test/Prompt/ValuePrinter/SourceLocation.C
devajithvs: Done!

PR #20023, opened 2025-10-02
[ntuple] better error messages in field reconciliation

jblomer:
When matching the in-memory field to the on-disk field fails, print a trace of the type hierarchy in-memory and on-disk that helps users to pinpoint the violating sub field.

Follow-up of https://github.com/root-project/root/pull/19937#pullrequestreview-3263430792

### DIFF
--- a/tree/ntuple/src/RFieldUtils.cxx
+++ b/tree/ntuple/src/RFieldUtils.cxx
@@ -731,7 +731,7 @@ std::string ROOT::Internal::GetTypeTraceReport(const RFieldBase &field, const RN
 
    auto fnGetLine = [](const std::string &fieldName, const std::string &fieldType, DescriptorId_t fieldId,
                        int level) -> std::string {
-      return std::string(2 * level, ' ') + fieldName + " [" + fieldType + "] (" + std::to_string(fieldId) + ")\n";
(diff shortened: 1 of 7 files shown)


tree/ntuple/src/RFieldUtils.cxx
```diff
@@ -722,3 +723,41 @@ bool ROOT::Internal::IsMatchingFieldType(std::string_view actualTypeName, std::s
    // Thus, we check again using first ROOT Meta normalization followed by RNTuple re-normalization.
    return (actualTypeName == ROOT::Internal::GetRenormalizedTypeName(ROOT::Internal::GetDemangledTypeName(ti)));
 }
+
+std::string ROOT::Internal::GetTypeTraceReport(const RFieldBase &field, const RNTupleDescriptor &desc)
```
pcanal, line 727: Could we/Should we add a test that this routine print what we want it to print?

pcanal (approved):
LGTM (delta the missing test and/or sample output in the commit)

silverweed:
Should we add a test for this?

tree/ntuple/src/RFieldUtils.cxx
jblomer: Test added

Issue #19265, opened 2025-07-03
ROOT builds for Debian/Ubuntu


guitargeek:
### Feature description

This subject often comes up in discussions, like just not in a [discussion about rpath and soversion](https://github.com/root-project/root/issues/18861#issuecomment-3031170983).

I therefore suggest to create a GitHub issue to discuss and work towards this goal, superseding the JIRA tickets where this was previously discussed:

  * https://its.cern.ch/jira/browse/ROOT-6428
  * https://its.cern.ch/jira/browse/ROOT-10878

We would definitely appreciate help here, in the best case by an external contributor who is already engaged with Debian packaging. Any PR to ROOT that goes in the direction of making it easier to generate `.deb` packages is highly welcome!

Debian 13 will be released this month I guess, but then we have the next Debian development cycle of about two years to make this happen.

### Alternatives considered

sdf

### Additional context

sdf

ferdymercury:
Copy-pasting from the other thread: Maybe related / convenient for this:

https://its.cern.ch/jira/browse/ROOT-6428
https://its.cern.ch/jira/browse/ROOT-6446
https://its.cern.ch/jira/browse/ROOT-8588
https://its.cern.ch/jira/browse/ROOT-8589
https://its.cern.ch/jira/browse/ROOT-9403
https://its.cern.ch/jira/browse/ROOT-9443
https://its.cern.ch/jira/browse/ROOT-9481
https://its.cern.ch/jira/browse/ROOT-9569
https://its.cern.ch/jira/browse/ROOT-9758
https://its.cern.ch/jira/browse/ROOT-10021
https://its.cern.ch/jira/browse/ROOT-10151
https://its.cern.ch/jira/browse/ROOT-10415
https://its.cern.ch/jira/browse/ROOT-10697
https://its.cern.ch/jira/browse/ROOT-10756
https://its.cern.ch/jira/browse/ROOT-10878

 * https://github.com/root-project/root/issues/6771
 * https://github.com/root-project/root/issues/7311
 * https://github.com/root-project/root/issues/8357
 * https://github.com/root-project/root/issues/8655
 * https://github.com/root-project/root/issues/8923
 * https://github.com/root-project/root/issues/9736
 * https://github.com/root-project/root/issues/10148
 * https://github.com/root-project/root/issues/10704
 * https://github.com/root-project/root/issues/10985
 * https://github.com/root-project/root/issues/11429
 * https://github.com/root-project/root/issues/11612
[...]

PR #20026, opened 2025-10-03
[Minuit2] Make function passing purely `std::function` based

guitargeek:
This makes is possible to use Minuit2 directly from Python, as we can benefit from the `std::function` pythonization.

It also makes the Minuit2 code more consistent, because `std::function` was already used for the G2 (second derivatives) and the Hessian.

Here is an example of how the base Minuit 2 can be used from Python now:
```python
import ROOT

# Define function to minimize
def fcn(params):
    x, y = params[0], params[1]
    return (x - 2.0)**2 + (y + 1.0)**2 + 1.0

# Wrap into Minuit2 FCNBase
func = ROOT.Minuit2.FCNAdapter(fcn)

# Initial values and step sizes as std::vector<double>
start_vals = ROOT.std.vector['double']([0.0, 0.0])
step_sizes = ROOT.std.vector['double']([0.1, 0.1])

# Create the minimizer
minimizer = ROOT.Minuit2.VariableMetricMinimizer()


state = ROOT.Minuit2.MnUserParameterState(start_vals, step_sizes)

# Run minimization directly
result = minimizer.Minimize(func, state)

# Inspect results
print("Minimization valid?", result.IsValid())
print("Minimum f =", result.Fval())
print("Best-fit parameters:")
for i in range(result.UserParameters().Params().size()):
    val = result.UserParameters().Params()[i]
    err = result.UserParameters().Errors()[i]
    print(f"  param[{i}] = {val} ± {err}")
```

This makes the base Minuit 2 code easily usable from Python, as promised in the ROOT plan of work. With "base Minuit 2 code", I mean the same classes that would also be built with Minuit 2 standalone, not using any other ROOT abstraction layers like the `ROOT::Math::Minimizer` interface, the `ROOT::Math::Fitter` or RooFit.

### DIFF
--- a/README/ReleaseNotes/v638/index.md
+++ b/README/ReleaseNotes/v638/index.md
@@ -70,6 +70,11 @@ The following people have contributed to this new version:
 ### Minuit2
 
 * Behavior change: building ROOT using `minuit2_omp=ON` option no longer enables OpenMP parallelization by default. One has to call now additionally GradientCalculator::SetParallelOMP().
+* The functionality of the `FCNGradAdapter` got absorbed into the `FCNAdapter`. This also means that the `FCNGradAdapter.h` header is gone.
+* The `ROOT::Minuit2::FCNAdapter` is no longer templated and now handle only functions with this exact signatures:
+  * `double(double const *)` for the wrapped function
+  * `void(double const *, double *)` for the gradient
+  The advantage of this restriction is that the `FCNAdapter` can now also wrap any Python function that respects this interface.
 
 ## RooFit
 


math/minuit2/inc/Minuit2/FCNAdapter.h
```diff
@@ -23,29 +24,119 @@ namespace Minuit2 {
...
+
+   FCNAdapter(Function f, double up = 1.) : fUp(up), fFunc(std::move(f)) {}
+
+   void SetGradientBufferSize(std::size_t n) { fGrad.resize(n); }
+
+   bool HasGradient() const override { return bool(fGradFunc); }
+   bool HasG2() const override { return bool(fG2Func); }
+   bool HasHessian() const override { return bool(fHessianFunc); }
+
+   double operator()(std::vector<double> const &v) const override { return fFunc(&v[0]); }
```
hageboeck: This looks like less pointer magic, even if it's the same:
```suggestion
   double operator()(std::vector<double> const &v) const override { return fFunc(v.data()); }
```

math/minuit2/inc/Minuit2/FCNAdapter.h
```diff
@@ -23,29 +24,119 @@ namespace Minuit2 {
...
 
 public:
-   FCNAdapter(const Function &f, double up = 1.) : fFunc(f), fUp(up) {}
+   using Function = std::function<double(double const *)>;
+   using GradFunction = std::function<void(double const *, double *)>;
+   using GradWithPrevResultFunction = std::function<void(double const *, double *, double *, double *, double *)>;
+   using G2Function = std::function<std::vector<double>(std::vector<double> const &)>;
+   using HessianFunction = std::function<bool(std::vector<double> const &, double *)>;
+
+   FCNAdapter(Function f, double up = 1.) : fUp(up), fFunc(std::move(f)) {}
```
hageboeck: It might be tricky to read the documentation on doxygen, since the `using` declaration is probably not visible in the context of the constructor. Users might have to click on it to understand what types they can pass.
From a documentation perspective, I would therefore recommend using:

```suggestion
   FCNAdapter(std::function<double(double const *)> f, double up = 1.) : fUp(up), fFunc(std::move(f)) {}
```

math/minuit2/inc/Minuit2/FCNAdapter.h
```diff
@@ -23,29 +24,119 @@ namespace Minuit2 {
...
+   bool HasG2() const override { return bool(fG2Func); }
+   bool HasHessian() const override { return bool(fHessianFunc); }
+
+   double operator()(std::vector<double> const &v) const override { return fFunc(&v[0]); }
 
-   double operator()(std::vector<double> const& v) const override { return fFunc.operator()(&v[0]); }
-   double operator()(const double *v) const { return fFunc.operator()(v); }
    double Up() const override { return fUp; }
 
+   std::vector<double> Gradient(std::vector<double> const &v) const override
```
hageboeck, line 77: Are these documented in the base class?

hageboeck: [I checked](https://root.cern.ch/doc/master/classROOT_1_1Minuit2_1_1FCNBase.html#acdb78b3e9fe24abc5f34dcd75fd62e76). Not so much. Would it be in scope to improve that a bit?

math/minuit2/inc/Minuit2/FCNAdapter.h
```diff
@@ -23,29 +24,119 @@ namespace Minuit2 {
...
+   {
+      fGradFunc(&v[0], &fGrad[0]);
+      return fGrad;
+   }
+   std::vector<double> GradientWithPrevResult(std::vector<double> const &v, double *previous_grad, double *previous_g2,
```
hageboeck: Same here, the arguments could benefit from a bit of explaining how to use them.

math/minuit2/inc/Minuit2/FCNAdapter.h
```diff
@@ -23,29 +24,119 @@ namespace Minuit2 {
...
+
+      return fHessian;
+   }
+
+   void SetReturnsInMinuit2ParameterSpace(bool flag) { fReturnsInMinuit2ParameterSpace = flag; }
+
+   void SetGradientFunction(GradFunction f) { fGradFunc = std::move(f); }
+   void SetG2Function(G2Function f) { fG2Func = std::move(f); }
+   void SetGradWithPrevResultFunction(GradWithPrevResultFunction f) { fGradWithPrevResultFunc = std::move(f); }
+   void SetHessianFunction(HessianFunction f) { fHessianFunc = std::move(f); }
```
hageboeck: Maybe also add small docstring explaining what `G2` or similar means.

math/minuit2/inc/Minuit2/FCNAdapter.h
```diff
@@ -23,29 +24,119 @@ namespace Minuit2 {
...
    double fUp;
+   mutable std::vector<double> fGrad;
+   mutable std::vector<double> fHessian;
+   mutable std::vector<double> fG2Vec;
+
+   Function fFunc;
+   GradFunction fGradFunc;
+   GradWithPrevResultFunction fGradWithPrevResultFunc;
+   G2Function fG2Func;
+   mutable HessianFunction fHessianFunc;
```
hageboeck: Why is this one mutable, but not the others?

hageboeck (approved):
LGTM, but note that an entry in the release notes might be in order:
- `FCNGradAdapter.h` is gone, so code might fail to compile
- `FCNAdapter<xxx>` has to be converted to `FCNAdapter`

math/minuit2/inc/Minuit2/FCNAdapter.h
guitargeek: Good point!

math/minuit2/inc/Minuit2/FCNAdapter.h
guitargeek: That's a good observation. I digged a bit on how this function is used, and I realized it's only relevant for the implementation details of the RooFit multiprocessing. So I would rather not that this is used by any one else.

I'll try to either refactor the code such that this public interface is not needed, or otherwise I make it clear in the docs that this is a RooFit implementation detail leaking out that should not be used.

math/minuit2/inc/Minuit2/FCNAdapter.h
```diff
@@ -23,29 +24,115 @@ namespace Minuit2 {
...
 @ingroup Minuit
 
 */
 
-template <class Function>
 class FCNAdapter : public FCNBase {
 
 public:
-   FCNAdapter(const Function &f, double up = 1.) : fFunc(f), fUp(up) {}
+   FCNAdapter(std::function<double(double const *)> f, double up = 1.) : fUp(up), fFunc(std::move(f)) {}
```
pcanal, line 45: I guess that this could (for better) lead to compilation error if the user function does not have this exact signature (but was 'compatible enough' with the code?  If so, shouldn't we (already) update the Release Notes accordingly?

README/ReleaseNotes/v638/index.md
```diff
@@ -65,6 +65,8 @@ The following people have contributed to this new version:
 ### Minuit2
 
 * Behavior change: building ROOT using `minuit2_omp=ON` option no longer enables OpenMP parallelization by default. One has to call now additionally GradientCalculator::SetParallelOMP().
+* The functionality of the `FCNGradAdapter` got absorbed into the `FCNAdapter`. This also meanst that the `FCNGradAdapter.h` header is gone.
```
pcanal: Is `FCNGradAdapter.h` rarely enough (/ never) used by users that we don't need to go through a deprecation phase?
```suggestion
* The functionality of the `FCNGradAdapter` got absorbed into the `FCNAdapter`. This also means that the `FCNGradAdapter.h` header is gone.
```

README/ReleaseNotes/v638/index.md
```diff
@@ -65,6 +65,8 @@ The following people have contributed to this new version:
...
+* The `FCNAdapter` is not a templated class anymore, but instead uses `std::function` to wrap the functions. Please remove the constructor template argument.
```
pcanal: I am not sure I understand the call for action :( [I am probably just missing the actual context in which this is used]

math/minuit2/inc/Minuit2/FCNAdapter.h
guitargeek: This `FCNAdapter` class was never exposed to the users before in any tutorials. I considered it as an implementation detail of the `Minuit2Minimizer`, that wraps Minuit 2 into the `ROOT::Math::Minimizer` interface.

But I needed an adapter class to make Minuit 2 usable from Python via the `std::function` pythonization, to the adapter class with the modifications suggested in this PR fits the bill nicely! So I want to soon expose it to the users with this new interface.

The end goal are some Minuit 2 tutorials in C++ and Python that showcase using the `FCNAdapter` with all the possible optional functions (Hessians, second derivatives, and gradients). So far users don't even know that external Hessians and second derivatives are possible.

README/ReleaseNotes/v638/index.md
guitargeek: You're right. Since the class is probably not present in user code, the call to action is superfluous

README/ReleaseNotes/v638/index.md
guitargeek: Yes. I can check by browsing more user code on the web, but I think it's not used anywhere except for our own `Minuit2Minimizer` wrapper.

math/minuit2/inc/Minuit2/FCNAdapter.h
pcanal: > This FCNAdapter class was never exposed to the users before in any tutorials. I considered it as an implementation detail of the Minuit2Minimizer, that wraps Minuit 2 into the ROOT::Math::Minimizer interface.
...
> The end goal are some Minuit 2 tutorials in C++ and Python that showcase using the FCNAdapter

So if I understand correctly `FCNAdapter`
- is part of the public interface (it is directly in the `ROOT::Minuit2` namespace),
- is meant to be part of the public interface (will soon have tutorials indicating how to use it)
- Needs to changes
- May or may not be already in use by adventurous users ... because we did not yet have any tutorials.

Are those more or so correct?  If they are, then, in my opinion, because the class was already public and maybe already used but was likely very very rarely used, we can change it usage without deprecation but we need to explain in some (small) way what no longer works.  Some variation on the theme `FCNAdapter is no longer templated and now handle only functions with this exact signature: ... `

guitargeek: Ok, yes your statement is correct. You propose to put this in the release notes, or also the docs?

pcanal: Release note only, this is a differential statement. (Although the documentation may want to be explicit (directly or indirectly via a typedef) about the signature required)

math/minuit2/inc/Minuit2/FCNAdapter.h
guitargeek: Yes. I have now added docs where the `FCNBase` class was missing it.

math/minuit2/inc/Minuit2/FCNAdapter.h
guitargeek: In the end, I managed to get rid of these RooFit implementation details at least from the `Math` function base class, and from the `FCNAdapter`. It's still in the `FCNBase`, but I changed the docs such that it's clear that it should not be used.

math/minuit2/inc/Minuit2/FCNAdapter.h
guitargeek: Because it checks if the evaluation of the Hessian failed when calling it, and if it failed, this function will be nulled to that it won't be called again.

I will probably change this Hessian interface in a future PR, when also adding Hessian support to RooFit.

math/minuit2/inc/Minuit2/FCNAdapter.h
ferdymercury: > I considered it as an implementation detail of the `Minuit2Minimizer`, that wraps Minuit 2 into the `ROOT::Math::Minimizer` interface.

Related comment on should-be-private headers: https://github.com/root-project/root/issues/19265#issuecomment-3319904634

guitargeek: Thanks for linking that, it's an important thing to keep in mind.

The problem with Minuit 2 is that it was written in a time where people didn't really consider making sure that no unneeded header files are exposed to the user. I have the impression that people just put all header files in `inc/`, even if they are only needed at build time. So for Minuit 2, we probably don't even need a subfolder for private headers, but it will be enough to move many headers to `src/`. To be done at a later point.

PR #20027, opened 2025-10-03
[math/RF] Reduce test failures when ROOT is built without clad or gsl

hageboeck:
When working with ROOT where some features were off, I encountered test failures. These commits reduce some of those, namely:
- Add fallbacks to other RooFit backends when clad is off. The tests continue to run also with the numeric backend.
- When selecting an unsupported algorithm in RootFinder, throw an exception instead of crashing.


tutorials/roofit/roofit/rf619_discrete_profiling.C
```diff
@@ -118,8 +118,16 @@ void rf619_discrete_profiling()
...
+   // Create an NLL and minimize it via the discrete profiling method.
+   std::unique_ptr<RooAbsReal> nll1;
+   try {
+      // Try using an NLL with automatic differentiation if available in ROOT
+      nll1.reset(simPdf.createNLL(*data, EvalBackend("codegen")));
+   } catch (std::runtime_error &e) {
+      std::cout << "Automatic differentiation not activated, falling back to numeric mode (" << e.what() << ")\n";
+      // Fallback to numeric differentiation
+      nll1.reset(simPdf.createNLL(*data));
+   }
```
guitargeek:
```suggestion
   // Create an NLL and minimize it via the discrete profiling method.

   nll1.reset(simPdf.createNLL(*data));
```
Sorry, that was an oversight when writing the tutorial. We should use the default backend here (we only wanted to check that it works with "codegen" during the development).

guitargeek (changes requested):
Thanks for fixing this! I only have one request about the tutorial

guitargeek (approved):
Thanks! LGTM

roofit/roofitcore/test/testRooMulti.cxx
```diff
@@ -78,7 +83,7 @@ TEST(RooMultiPdfTest, FitConvergesAndReturnsReasonableResult)
 
    // Fit 1 - RooMultiPdf fit
 
-   std::unique_ptr<RooAbsReal> nll{multipdf.createNLL(*data, EvalBackend("codegen"))};
```
vgvassilev, line 81: Can we do it the opposite way. We have the ifdef in the `EvalBackend("codegen")` where we check if clad is not on, we warn and switch to `EvalBackend::defaultValue()`?

hageboeck: I guess that possible. @guitargeek, OK with you?

hageboeck: OK, I talked to Jonas, and we propose to do the following:
- testRooMulti only runs with the default backend.
- codegen gets tested already extensively in the tests that loop through all backends, so we have good coverage.
- And we add an error if users choose codegen while clad is disabled.

vgvassilev: Makes sense. I hope the ifdef-s will be in a single place.

hageboeck: They are now, and I think I covered the point that all code paths run through. Please have look.

vgvassilev: Nice. Go ahead!

hageboeck:

PR #20032, opened 2025-10-05
[Math] Make building old C++ Minuit implementation optional

guitargeek:
Minuit 2 has been the default for 2 years, and one day we should consider not building ROOT with the old C++ TMinuit implementation anymore.

To make it possible to try this out, introduce a new `minuit1` flag that is on by default.

Building with `minuit1=OFF` will also help to ensure that we are actually using Minuit 2 in all our tests. With this, it was uncovered
that the `stressRooFit` actually didn't use Minuit 2, even thought it's the default minimizer in RooFit as well. Therefore, the default minimizer setting in `stressRooFit`/`stressRooStats` are changed to `Minuit2`, reference files are updated, and new tests with the old Minuit 1 are added to keep previous test coverage.


pcanal:
There seems to be unrelated commits in the PR (roofit value servers)

guitargeek:
They are related. Using Minuit 2 consistently unearthed some problems in RooFit, because Minuit 2 is more verbose with errors.

So I use this PR to also see how I can fix these problems. The goal is to create spin-of PRs for the RooFit fixes (many are already in), and then finally come back to this PR and only include the Minuit changes.

dpiparo:
The `modules_off` build seems to suffer from errors which do not seem due to glitches. However, they could be due to some artifacts, interfering with the creation of the PCH. I am restarting the builds as *clean builds* to then be able to assess whether the changes are fine to be integrated.

dpiparo:
nope. The problem persists @guitargeek

guitargeek:
The problem should be gone now. I have fixed some edge cases with the new CPU evaluation backend in the recent months, so all tests should work now (before, a problematic fit in `stressRooFit` was silently passing, because the old Minuit 1 was less verbose, not printing an "Error" message that caused the test to fail, which was grepping for "Error" in its passing condition)

guitargeek:
All tests pass, in particular the RooFit ones. Hence, this experiment PR showed that the new Minuit 2 can completely replace Minuit 1 *in principle*, which is what I wanted to know.

PR #20034, opened 2025-10-06
[RF] Avoid implicit creation of normalization set in RooNumRunningInt

guitargeek:
Fixes this error when using the RooNumRunningInt class:

```
what():  calling RooAbsReal::getVal() with r-value references to the
normalization set is not allowed, because it breaks RooFits caching
logic and potentially introduces significant overhead. Please
explicitly create the RooArgSet outside the call to getVal().
```


lmoneta (approved):
LGTM!
Thank you for fixing this error!

PR #20040, opened 2025-10-06
[RF] Avoid compute graph desync in RooFixedProdPdf

guitargeek:
The RooFixedProdPdf used an antipattern that should be avoided, and it was actually causing some problems in fits with the new CPU evaluation backend.

The problem was that the class held on to some RooAbsArgs for its evaluation in some containers that were separate from RooFit server-client interface (the container was the
`std::unique_ptr<RooProdPdf::CacheElem>`).

This commit fixes the situation by avoiding this cache element member completely and consistently using the server proxies instead.


lmoneta (approved):
LGTM!
Thank you Jonas for the fix

Issue #13239, opened 2023-07-13
TBranch::GetBulkEntries does not work for variable-sized arrays


eguiraud:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

As per the title, `TBranch::GetBulkEntries` does not work for variable-sized arrays, while it works for static-sized arrays and scalars.

### Reproducer

```cpp
#include <ROOT/RDataFrame.hxx>
#include <ROOT/RVec.hxx>
#include <TBufferFile.h>
#include <TFile.h>
#include <TTree.h>

void FillTree(const char *filename, const char *treeName) {
  TFile f(filename, "recreate");
  TTree t(treeName, treeName);
  int n = 2;
  double *x = new double[2];
  x[0] = 42;
  x[1] = 84;
  t.Branch("n", &n);
  t.Branch("x", x, "x[n]/D");
  t.Fill();
  t.Write();
  f.Close();
}

int main() {
  auto filename = "repro_getbulkread.root";
  auto treename = "t";
  FillTree(filename, treename);

  TFile f(filename);
  auto *t = f.Get<TTree>(treename);

  t->Scan();
  
  auto *b = t->GetBranch("x");
  TBufferFile buf(TBuffer::kWrite, 10000);
  const auto n = b->GetBulkRead().GetBulkEntries(0, buf);
  assert(n == 1);
}
```

### ROOT version

6.29

### Installation method

any

### Operating system

any

ferdymercury:
This gives now an error:

`Error in <TBranch::GetBulkEntries>: Leaf failed to read.`

PR #20055, opened 2025-10-08
[tree] GetBulkEntries fallback when N=1 and Ndata

ferdymercury:
Fixes https://github.com/root-project/root/issues/13239


pcanal:
Can we add/extend a test for this?

PR #20056, opened 2025-10-08
[rfile] Introduce GetKeys() methods

silverweed:
# This Pull request:
is a followup of #19958. Other than addressing @pcanal's comments, this PR introduces the `GetKeys()` and `GetKeysNonRecursive()` methods, which iterate over the keys of some subdirectory.
They do so via an iterator that returns an abridged version of TKey, currently containing only barebones information (more may be added later).

## Checklist:

- [x] tested changes locally


io/io/inc/ROOT/RFile.hxx
```diff
@@ -11,11 +11,13 @@
...
 
 #include <memory>
+#include <iostream>
 #include <string_view>
 #include <typeinfo>
 
-class TFile;
+class TIterator;
 class TKey;
+class TFile;
```
jblomer: Order alphabetically?

io/io/inc/ROOT/RFile.hxx
```diff
@@ -29,6 +31,93 @@ ROOT::RLogChannel &RFileLog();
...
+Every object inside a ROOT file has an associated "Key" which contains metadata on the object, such as its name, type
+etc.
+Querying this information can be done via RFile::GetKeys() or RFile::GetKeysNonRecursive. Reading an object's Key
+doesn't deserialize the full object, so it's a relatively lightweight operation.
+*/
+struct RFileKeyInfo {
+   std::string fName;
+   std::string fTitle;
+   std::string fClassName;
+   std::uint16_t fCycle;
```
jblomer: Default-initialize

io/io/inc/ROOT/RFile.hxx
```diff
@@ -29,6 +31,93 @@ ROOT::RLogChannel &RFileLog();
...
+\class ROOT::Experimental::RFileKeyInfo
+\ingroup RFile
+\brief Information about an RFile object's Key.
+
+Every object inside a ROOT file has an associated "Key" which contains metadata on the object, such as its name, type
+etc.
+Querying this information can be done via RFile::GetKeys() or RFile::GetKeysNonRecursive. Reading an object's Key
+doesn't deserialize the full object, so it's a relatively lightweight operation.
+*/
+struct RFileKeyInfo {
```
jblomer: I'm not sure this is the best name, because the keys are the same whether it is TFile or RFile. I'd probably prefer `RFile::RKeyInfo` or `Detail::RKeyInfo`... (not entirely sure about those two either).

io/io/inc/ROOT/RFile.hxx
```diff
@@ -29,6 +31,93 @@ ROOT::RLogChannel &RFileLog();
...
+   std::string fTitle;
+   std::string fClassName;
+   std::uint16_t fCycle;
+};
+
+/// The iterable returned by RFile::GetKeys() and RFile::GetKeysNonRecursive()
+class RFileKeyIterable final {
+   using Pattern_t = std::string;
+
+   TFile *fFile;
```
jblomer: default-initialize

io/io/inc/ROOT/RFile.hxx
```diff
@@ -196,6 +285,31 @@ public:
...
+      return RFileKeyIterable(fFile.get(), rootPath, RFileKeyIterable::kRecursive);
+   }
+
+   /// Returns an iterable over all paths of objects written into this RFile contained in the directory "rootPath".
+   /// The returned paths are always "absolute" paths: they are not relative to `rootPath`.
+   /// Keys relative to directories are not returned: only those relative to leaf objects are.
+   /// If `rootPath` is the path of a leaf object, only `rootPath` itself will be returned.
+   /// This only returns the immediate children of `rootPath`. If you want to recurse into the subdirectories of
+   /// `rootPath`, use GetKeys().
+   RFileKeyIterable GetKeysNonRecursive(std::string_view rootPath = "") const
```
jblomer: I think I'd prefer a single method and expose the flags (`RFileKeyIterable::kNone`, `RFileKeyIterable::kRecursive`).

io/io/inc/ROOT/RFile.hxx
```diff
@@ -196,6 +285,31 @@ public:
...
    /// Flushes the RFile if needed and closes it, disallowing any further reading or writing.
    void Close();
+
+   /// Returns an iterable over all paths of objects written into this RFile starting at path "rootPath".
+   /// The returned paths are always "absolute" paths: they are not relative to `rootPath`.
+   /// Keys relative to directories are not returned: only those relative to leaf objects are.
+   /// If `rootPath` is the path of a leaf object, only `rootPath` itself will be returned.
+   /// This recurses on all the subdirectories of `rootPath`. If you only want the immediate children of `rootPath`,
+   /// use GetKeysNonRecursive().
+   RFileKeyIterable GetKeys(std::string_view rootPath = "") const
```
jblomer: Perhaps I'd rather call this `GetKeysListing()` or `ListKeys()`. I think the documentation should also mention that a file may contain hidden keys that will not be found by this method (but by `TFile::Map()`.

io/io/src/RFile.cxx
```diff
@@ -30,13 +32,17 @@ using ROOT::Experimental::Internal::RFileLog;
...
-   }
 
-   if (!ROOT::EndsWith(path, ".root")) {
-      R__LOG_WARNING(RFileLog()) << "ROOT::RFile only supports ROOT files. The preferred file extension is \".root\"";
-   }
+   if (ROOT::EndsWith(path, ".zip"))
+      throw ROOT::RException(R__FAIL("ROOT::RFile doesn't support Zip files."));
+
+   if (!ROOT::EndsWith(path, ".root"))
+      R__LOG_INFO(RFileLog()) << "ROOT::RFile only supports ROOT files. The preferred file extension is \".root\"";
```
jblomer: Coming back to this, LHCb, for instance, stores ROOT files with the ending `.dst`. Also, I guess zip files could be supported in archive mode.

Anyway, I'd be in favor not to check the extension. If you want to check at all, then you can in case of failure try to open the file in raw mode and check if the file starts with the magic bytes.

io/io/src/RFile.cxx
```diff
@@ -180,49 +171,52 @@ static std::string ValidateAndNormalizePath(std::string &path)
...
 {
-   CheckExtension(path);
-
    TDirectory::TContext ctx(nullptr); // XXX: probably not thread safe?
    auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "READ_WITHOUT_GLOBALREGISTRATION"));
    if (!tfile || tfile->IsZombie())
       throw ROOT::RException(R__FAIL("failed to open file " + std::string(path) + " for reading"));
 
+   if (tfile->IsRaw() || !tfile->IsBinary() || tfile->IsArchive())
+      throw ROOT::RException(R__FAIL("Opened file " + std::string(path) + " is not a ROOT file"));
```
pcanal:
```suggestion
      throw ROOT::RException(R__FAIL("Opened file " + std::string(path) + " is not a ROOT binary file"));
```
whether we want to hint on whether xml and archive file will be supported at some point is another question.

io/io/src/RFile.cxx
```diff
@@ -180,49 +171,52 @@ static std::string ValidateAndNormalizePath(std::string &path)
...
 {
-   CheckExtension(path);
-
    TDirectory::TContext ctx(nullptr); // XXX: probably not thread safe?
    auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "UPDATE_WITHOUT_GLOBALREGISTRATION"));
    if (!tfile || tfile->IsZombie())
       throw ROOT::RException(R__FAIL("failed to open file " + std::string(path) + " for updating"));
 
+   if (tfile->IsRaw() || !tfile->IsBinary() || tfile->IsArchive())
+      throw ROOT::RException(R__FAIL("Opened file " + std::string(path) + " is not a ROOT file"));
```
pcanal:
```suggestion
      throw ROOT::RException(R__FAIL("Opened file " + std::string(path) + " is not a ROOT binary file"));
```

io/io/src/RFile.cxx
```diff
@@ -180,49 +171,52 @@ static std::string ValidateAndNormalizePath(std::string &path)
...
 {
-   CheckExtension(path);
-
    TDirectory::TContext ctx(nullptr); // XXX: probably not thread safe?
    auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "RECREATE_WITHOUT_GLOBALREGISTRATION"));
    if (!tfile || tfile->IsZombie())
       throw ROOT::RException(R__FAIL("failed to open file " + std::string(path) + " for writing"));
 
+   if (tfile->IsRaw() || !tfile->IsBinary() || tfile->IsArchive())
+      throw ROOT::RException(R__FAIL("Opened file " + std::string(path) + " is not a ROOT file"));
```
pcanal:
```suggestion
      throw ROOT::RException(R__FAIL("Opened file " + std::string(path) + " is not a ROOT binary file"));
```

io/io/inc/ROOT/RFile.hxx
```diff
@@ -29,6 +31,95 @@ ROOT::RLogChannel &RFileLog();
...
+   enum class ECategory : std::uint16_t {
+      kInvalid,
+      kObject,
+      kDirectory
+   };
+
+   std::string fName;
+   std::string fTitle;
+   std::string fClassName;
+   std::uint16_t fCycle = 0;
```
pcanal: Here we choose to duplicate the information rather than forwarding it from the live `TKey`.  What are the pros and cons?

io/io/inc/ROOT/RFile.hxx
```diff
@@ -65,13 +155,11 @@ file).
...
-Directories are still an existing concept in RFile (since they are a concept in the ROOT binary format),
-but they are usually interacted with indirectly, via the use of filesystem-like string-based paths. If you Put an object
-in an RFile under the path "path/to/object", "object" will be stored under directory "to" which is in turn stored under
-directory "path". This hierarchy is encoded in the ROOT file itself and it can provide some optimization and/or
-conveniencies when querying objects.
+Even though there is no equivalent of TDirectory in the RFile API, directories are still an existing concept in RFile
+(since they are a concept in the ROOT binary format). However they are for now only interacted with indirectly, via the
+use of filesystem-like string-based paths. If you Put an object in an RFile under the path "path/to/object", "object"
+will be stored under directory "to" which is in turn stored under directory "path". This hierarchy is encoded in the
+ROOT file itself and it can provide some optimization and/or conveniencies when querying objects.
```
pcanal:
```suggestion
ROOT file itself and it can provide some optimization and/or conveniences when querying objects.
```

io/io/inc/ROOT/RFile.hxx
```diff
@@ -196,6 +293,23 @@ public:
 
    /// Flushes the RFile if needed and closes it, disallowing any further reading or writing.
    void Close();
+
+   /// Returns an iterable over all paths of objects written into this RFile starting at path "rootPath".
```
pcanal: The sentence does not give wiggle room to include the case `(flags & kListRecursive) == 0`.  We should add a conditional somewhere, for example
```suggestion
   /// Returns an iterable over all paths of objects written into this RFile starting at path "rootPath" (defaulting to also include the content of sub-directories)
```
(but the proposed words don't seem to match the first part (path vs directories))

io/io/src/RFile.cxx
```diff
@@ -369,6 +363,138 @@ void RFile::PutUntyped(std::string_view pathSV, const std::type_info &type, cons
...
+         fIterStack.emplace_back(innerIter, dirPath + dirSep + dir->GetName());
+         if (!includeDirs)
+            continue;
+      } else if (!includeObj) {
+         continue;
+      }
+
+      // Reconstruct the full path of the key
+      const auto &fullPath = dirPath + dirSep + key->GetName();
+      const auto nesting = fIterStack.size() - 1;
```
pcanal:
```suggestion
      const std::uint16_t nesting = fIterStack.size() - 1;
```
To avoid the warnings:
```
comparison of integer expressions of different signedness: ‘const long unsigned int’ and ‘int’ [-Wsign-compare]
```
Alternatively `fRootDirNesting` could be made the same as `decltype(fIterStack.size())`

io/io/src/RFile.cxx
```diff
@@ -180,49 +171,52 @@ static std::string ValidateAndNormalizePath(std::string &path)
...
 std::unique_ptr<RFile> RFile::Open(std::string_view path)
 {
-   CheckExtension(path);
-
    TDirectory::TContext ctx(nullptr); // XXX: probably not thread safe?
    auto tfile = std::unique_ptr<TFile>(TFile::Open(std::string(path).c_str(), "READ_WITHOUT_GLOBALREGISTRATION"));
    if (!tfile || tfile->IsZombie())
       throw ROOT::RException(R__FAIL("failed to open file " + std::string(path) + " for reading"));
 
+   if (tfile->IsRaw() || !tfile->IsBinary() || tfile->IsArchive())
```
pcanal: Why is `RFile` sensitive to the backend (binary vs text) used by `TFIle`?

silverweed: Because we don't want to promise to support non-binary root files in RFile (always in case of us changing the implementation to not use TFile in the future)

io/io/inc/ROOT/RFile.hxx
silverweed: We could as well hide these and have getter methods instead, which could query the internal TKey. No reason in particular if not simplicity; for me the only important part is not to expose TKey directly, but it can make sense to use it internally to save a few copies.
I think we do want the getters in the end, so I'll change it.

EDIT: actually there is an obvious con to using the internal TKey: the RKeyInfo will have the same lifetime as the file and will become invalid when the file is closed (unless we somehow keep the TKey alive, but that I'd rather not do).
So, even though using the getters is probably better in case we want to change how it works in the future, I'd still copy the information in order to not have lifetime issues, at least for now.

io/io/inc/ROOT/RFile.hxx
silverweed: I'm not sure about making it a subclass of RFile (as you say the keys are a concept of a ROOT file in general), and putting it in Detail is also a bit weird because it's not really part of the "advanced interface"...

io/io/inc/ROOT/RFile.hxx
```diff
@@ -196,6 +319,39 @@ public:
 
    /// Flushes the RFile if needed and closes it, disallowing any further reading or writing.
    void Close();
+
+   /// Returns an iterable over all keys of objects and/or directories written into this RFile starting at path
+   /// `rootPath` (defaulting to include the content of all subdirectories).
```
enirolf: Maybe this is a stretch,but I'm wondering whether using the word `rootPath` could lead to confusion, given that the ROOT files themselves are `ABC.root` and perhaps could be understood by some users as `rootPath`...

io/io/src/RFile.cxx
```diff
@@ -179,50 +170,63 @@ static std::string ValidateAndNormalizePath(std::string &path)
...
+      throw ROOT::RException(R__FAIL("Opened file " + std::string(path) + " is not a ROOT binary file"));
+
```
enirolf: Since this is duplicated across `Open`, `Update` and `Recreate`, would it make sense to have a singular `EnsureIsBinary` function?

io/io/src/RFile.cxx
```diff
@@ -369,6 +373,139 @@ void RFile::PutUntyped(std::string_view pathSV, const std::type_info &type, cons
...
+      assert(!fIterStack.empty());
+      const std::size_t nesting = fIterStack.size() - 1;
+
+      // Skip key if it's not a child of root dir
+      if (!ROOT::StartsWith(fullPath, fPattern))
+         continue;
+
+      // Check that we are in the same directory as "rootDir".
+      // Note that for directories we list both the root dir and the immediate children (in non-recursive mode).
+      if (!recursive && nesting != fRootDirNesting && (!isDir || nesting != fRootDirNesting + 1))
```
enirolf: Maybe you've already seen it, but this line is causing a `comparison of integer expressions of different signedness` warning on a bunch of platforms

io/io/inc/ROOT/RFile.hxx
silverweed: Yeah, the word is overloaded unfortunately...maybe `basePath` works better?

io/io/src/RFile.cxx
silverweed: I'm not sure why, since `nesting` and `fRootDirNesting` are both supposed to be unsigned...

enirolf: Apparently adding `1` to a `std::uint16_t` auto-casts it to an `int`: https://godbolt.org/z/sYvcsboee...

silverweed: I guess it makes "sense". Thanks, I'll fix it

io/io/inc/ROOT/RFile.hxx
```diff
@@ -37,16 +37,15 @@ ROOT::RLogChannel &RFileLog();
 ## When and why should you use RFile
 
 RFile is a modern and minimalistic interface to ROOT files, both local and remote, that can be used instead of TFile
-when the following conditions are met:
-- you want a simple interface that makes it easy to do things right and hard to do things wrong;
-- you only need basic Put/Get operations and don't need the more advanced TFile/TDirectory functionalities;
-- you want more robustness and better error reporting for those operations;
-- you want clearer ownership semantics expressed through the type system rather than having objects "automagically"
-  handled for you via implicit ownership of raw pointers.
+when you only need basic Put/Get operations and don't need the more advanced TFile/TDirectory functionalities.
```
hageboeck, line 163: Slightly off-topic, but I would prefer seeing this text in `RFile.cxx` or in an associated `.md` in the future. The reasoning is that documentation updates shouldn't recompile the world.

io/io/inc/ROOT/RFile.hxx
```diff
@@ -196,6 +319,39 @@ public:
 
    /// Flushes the RFile if needed and closes it, disallowing any further reading or writing.
    void Close();
+
+   /// Returns an iterable over all keys of objects and/or directories written into this RFile starting at path
+   /// `rootPath` (defaulting to include the content of all subdirectories).
+   /// By default, keys relative to directories are not returned: only those relative to leaf objects are.
```
hageboeck:
```suggestion
   /// By default, keys referring to directories are not returned: only those referring to leaf objects are.
```

io/io/inc/ROOT/RFile.hxx
```diff
@@ -196,6 +319,39 @@ public:
...
+   /// If `(flags & kListObject) != 0`, the listing will include keys of non-directory objects (default);
+   /// If `(flags & kListDirs) != 0`, the listing will include keys of directory objects;
+   /// If `(flags & kListRecursive) != 0`, the listing will recurse on all subdirectories of `rootPath` (default),
+   /// otherwise it will only list immediate children of `rootPath`.
+   ///
+   /// Example usage:
+   /// ~~~{.cpp}
+   /// for (auto keyInfo : file->ListKeys()) {
+   ///     /* iterate over all objects in the RFile */
+   /// }
```
hageboeck, line 344: At least this first example could include what to do with the `value_type`s in the body. So far, it's obscured by the `auto` and the typedefs, and one needs to follow the typedefs to figure out how to get to an object.

io/io/src/RFile.cxx
```diff
@@ -361,6 +373,140 @@ void RFile::PutUntyped(std::string_view pathSV, const std::type_info &type, cons
...
+         fIterStack.pop_back();
+         continue;
+      }
+
+      assert(keyObj->IsA() == TClass::GetClass<TKey>());
+      auto key = static_cast<TKey *>(keyObj);
+
+      const auto dirSep = (dirPath.empty() ? "" : "/");
+      const auto isDir =
+         strcmp(key->GetClassName(), "TDirectory") == 0 || strcmp(key->GetClassName(), "TDirectoryFile") == 0;
```
hageboeck: Could there be TDirectory-derived types that would qualify, too, or is this forbidden by TFile?
i.e., do we also have to check if <type> derives from TDirectory?

io/io/src/RFile.cxx
```diff
@@ -361,6 +373,140 @@ void RFile::PutUntyped(std::string_view pathSV, const std::type_info &type, cons
...
+   keyInfo.fTitle = key->GetTitle();
+   return keyInfo;
+}
+
+void RFile::Print(std::ostream &out) const
+{
+   std::vector<RKeyInfo> keys;
+   auto keysIter = ListKeys();
+   for (const auto &key : keysIter) {
+      keys.emplace_back(key);
```
hageboeck, line 497: Consider `vector.reserve()` if the number of elements can be obtained.

io/io/test/rfile.cxx
```diff
@@ -315,6 +353,42 @@ TEST(RFile, WriteReadInTFileDir)
...
+   FileRaii fileGuard("test_rfile_iteratekeys.root");
+
+   {
+      auto file = RFile::Recreate(fileGuard.GetPath());
+      TH1D a;
+      auto b = std::make_unique<std::string>();
+      std::string c = "0";
+      file->Put("a", a);
+      file->Put("b", *b);
+      file->Put("c", c);
```
hageboeck, line 367: Consider adding:
```suggestion
      file->Put("c", c);
      file->Put("/foo/bar/c", c);
```

io/io/test/rfile.cxx
```diff
@@ -423,35 +593,30 @@ TEST(RFile, Closing)
...
-TEST(RFile, InvalidPaths)
+TEST(RFile, GetAfterOverwriteNoBackup)
 {
-   FileRaii fileGuard("test_rfile_invalidpaths.root");
+   FileRaii fileGuard("test_rfile_getafternobackup.root");
 
    auto file = RFile::Recreate(fileGuard.GetPath());
+   std::string s = "foo";
+   file->Put("s", s);
+   file->Overwrite("s", s, false);
```
hageboeck, line 605: Consider:
```suggestion
   file->Put("s", s);
   s = "bar";
   file->Overwrite("s", s, false);
```

io/io/inc/ROOT/RFile.hxx
silverweed: > The reasoning is that documentation updates shouldn't recompile the world.

In a way that makes sense, but the counter-reasoning is that a user that only has access to headers (e.g. because they installed ROOT via the system package manager) should still be able to read the documentation from them. So I'm not very convinced about it, especially considering that documentation is not expected to change frequently in the long run

hageboeck: I think ROOT wasn't doing this in the past, but it's a good argument to start doing it now ...

io/io/src/RFile.cxx
silverweed: I'm not entirely sure but I think there can never be directories that are not those two specific types...maybe @pcanal can answer better.

io/io/src/RFile.cxx
silverweed: Unfortunately getting the total number of keys is not that easy as it requires visiting the whole directory tree, so we can't really get the keys count efficiently before the iteration. (`TDirectory::GetNkeys()` and `TDirectory::GetListOfKeys()` both return only the top-level keys)

io/io/src/RFile.cxx
pcanal: There is nothing that really prevents it ... In a similar context `TFileMerger` does:
```
TClass *cl = TClass::GetClass(key->GetClassName());
...
if (cl->InheritsFrom(TDirectory::Class())
...
```
Similarly there are several classes inheriting from `TKey` (although there only the GUI seems to check for inheritance and thus it is not clear those derived classes are/can-be used in the directories' list of keys)

io/io/inc/ROOT/RFile.hxx
```diff
@@ -196,6 +319,40 @@ public:
...
+   ///     /* iterate over all directories under directory "foo", recursively */
+   /// }
+   /// ~~~
+   RFileKeyIterable ListKeys(std::string_view basePath = "", std::uint32_t flags = kListObjects | kListRecursive) const
+   {
+      return RFileKeyIterable(fFile.get(), basePath, flags);
+   }
+
+   /// Prints the internal structure of this RFile to the given stream.
+   void Print(std::ostream &out = std::cout) const;
```
pcanal, line 361: Related to `Print`'s API.

Historically we had `Print(const char* opt ="")` and `ls(const char *opt = "")` with the later usually printing a more compact version of the former.

We may want to consider having a consistent 'experience' for printing information about the object across the new APIs.  (Yes, it is no longer imposed by `TObject` but I believe we should still be consistent).

* Examples that could lead to the desire to customize the output for RFile:

How are we planning of supporting the user 'pre-placing' object hierarchy (currently supported via the shared ownership).  At some point I think we consider one of the option being `RFile` having a collection of `shared_pointer`.  Thus one might want to Print `different` things about the `RFile`.

Another reason for wanting customization would be to 'limit' the amount of information printed (for example to the first level).

* Current RNTuple related interfaces (subset that has `Print` in the name)

```
tree/ntuple/inc/ROOT/RNTupleMetrics.hxx:   void Print(std::ostream &output, const std::string &prefix = "") const;
tree/ntuple/inc/ROOT/RNTupleReader.hxx:   void PrintInfo(const ENTupleInfo what = ENTupleInfo::kSummary, std::ostream &output = std::cout) const;
tree/ntuple/inc/ROOT/RNTupleProcessor.hxx:   virtual void PrintStructureImpl(std::ostream &output) const = 0;
tree/ntuple/inc/ROOT/RNTupleProcessor.hxx:   void PrintStructure(std::ostream &output = std::cout) { PrintStructureImpl(output); }
tree/ntuple/inc/ROOT/RFieldVisitor.hxx:   void PrintIndent();
tree/ntuple/inc/ROOT/RFieldVisitor.hxx:   void PrintName(const ROOT::RFieldBase &field);
tree/ntuple/inc/ROOT/RFieldVisitor.hxx:   void PrintCollection(const ROOT::RFieldBase &field);
tree/ntuple/inc/ROOT/RFieldVisitor.hxx:   void PrintRecord(const ROOT::RFieldBase &field);
tree/ntuple/inc/ROOT/RNTupleDescriptor.hxx:   void PrintInfo(std::ostream &output) const;
```

io/io/src/RFile.cxx
silverweed: I replaced the check with a proper `InheritsFrom`

io/io/inc/ROOT/RFile.hxx
silverweed: I agree that we will need a more customizable Print, but I think it would be the job of a future PR to define that...in this PR it's mostly there for quick debugging, not intended as a final interface.
We might also want to discuss about this interface uniformity issue that you bring up, as the convenience of a single `ls()` across "all" possible objects is undeniable from the user's perspective (mostly in the context of interactive use)

io/io/inc/ROOT/RFile.hxx
vepadulano: I like `RFileKeyInfo` because it tells me this is information about a key in a file, the `R` is just there because the class is new, I don't see it linked just to RFile in this case (that would have been spelled `RRFileKeyInfo`).

That being said, I could think of other possible names e.g. `RROOTFileKeyInfo`, `RKeyInFileInfo`.

io/io/inc/ROOT/RFile.hxx
vepadulano: My absolutely completely personal view on this. if we pick random 100 users and ask them about our docs:

* 80 will say they never read it
* 18 will say they read it on https://root.cern/doc/master/
* 2 will say they read it via headers

(I'm keeping out search engine/ROOT forum/LLM queries from this count)

So I don't believe the placement between headers or source files will make a difference for our users. On the other hand, if it makes a difference for us, it might still be worth understanding how to approach this best. That being said, we have never made a clear decision on this topic before.

io/io/inc/ROOT/RFile.hxx
```diff
@@ -29,6 +32,122 @@ ROOT::RLogChannel &RFileLog();
...
 
+/// Given a "path-like" string (like foo/bar/baz), returns a pair `{ dirName, baseName }`.
+/// `baseName` will be empty if the string ends with '/'.
+/// `dirName` will be empty if the string contains no '/'.
+/// `dirName`, if not empty, always ends with a '/'.
+/// NOTE: this function does no semantic checking or path expansion, nor does it interact with the
+/// filesystem in any way (so it won't follow symlink or anything like that).
+/// Moreover it doesn't trim the path in any way, so any leading or trailing whitespaces will be preserved.
+/// This function does not perform any copy: the returned string_views have the same lifetime as `path`.
+std::pair<std::string_view, std::string_view> DecomposePath(std::string_view path);
```
vepadulano, line 45: This one sounds like an internal/utility function. Have you considered moving it to the implementation file directly? If it must stay in the header, I believe it should go in `Internal` namespace

io/io/inc/ROOT/RFile.hxx
```diff
@@ -29,6 +32,122 @@ ROOT::RLogChannel &RFileLog();
...
+
+class RFileKeyIterable;
+
+/**
+\class ROOT::Experimental::RKeyInfo
+\ingroup RFile
+\brief Information about an RFile object's Key.
+
+Every object inside a ROOT file has an associated "Key" which contains metadata on the object, such as its name, type
```
vepadulano, line 56: Why the `"Key"`? I find both the capitalisation and the usage of quotes a bit misleading.

io/io/inc/ROOT/RFile.hxx
```diff
@@ -29,6 +32,122 @@ ROOT::RLogChannel &RFileLog();
...
+      struct RIterStackElem {
+         // This is ugly, but TList returns an (owning) pointer to a polymorphic TIterator...and we need this class
+         // to be copy-constructible.
+         std::shared_ptr<TIterator> fIter;
+         std::string fDirPath;
+
+         // Outlined to avoid including TIterator.h
+         RIterStackElem(TIterator *it, const std::string &path = "");
+         // Outlined to avoid including TIterator.h
+         ~RIterStackElem();
```
vepadulano, line 110: Ideally I would encourage use of rule of five, especially since this class is required to be copy-constructible

io/io/src/RFile.cxx
```diff
@@ -178,51 +169,61 @@ static std::string ValidateAndNormalizePath(std::string &path)
...
-      throw ROOT::RException(R__FAIL("failed to open file " + std::string(path) + " for writing"));
+   EnsureFileOpenAndBinary(tfile.get(), path);
 
    auto rfile = std::unique_ptr<RFile>(new RFile(std::move(tfile)));
    return rfile;
 }
 
+RFile::RFile(std::unique_ptr<TFile> file) : fFile(std::move(file)) {}
+
+RFile::~RFile() = default;
```
vepadulano, line 225: Looking at this RFile destructor here also triggers my personal suggestion for employing rule of five.

io/io/inc/ROOT/RFile.hxx
silverweed: I think this can be a useful function for users to manipulate paths more easily, but arguably it could be considered "advanced" enough to go into `Detail`

io/io/inc/ROOT/RFile.hxx
silverweed: Because this sentence is defining what a "Key" is in the context of a ROOT file

io/io/inc/ROOT/RFile.hxx
silverweed: Is it really useful in this case though? The only reason we're defining the dtor explicitly is to outline it in the C++ file, but it's not doing anything special, so I don't think we need to also specify the ctors

io/io/src/RFile.cxx
silverweed: Same consideration as before: as far as I understand the "rule of five" is only useful if any of the "five" is doing non-default stuff (e.g. dealing with non-RAII resources), which in this case is not true

io/io/inc/ROOT/RFile.hxx
silverweed: I have always found documentation in the implementation a very weird choice and my personal preference is having it on the declaration.
If we never made a decision and we don't have a clear rule, my preference is to keep it as-is, unless there are strong reasons to do otherwise.

io/io/inc/ROOT/RFile.hxx
vepadulano: Ok, then I have two followup comments:

* If the function is generic enough, i.e. it could be useful on its own even if `RFile` did not exist, then it should be placed in `core/foundation/inc/ROOT/StringUtils.hxx` or somewhere close imho.
* If it's a function that can be situationally useful but only for users of RFile, then I would still put it at least in `Detail` and then it could be left in this header.

io/io/src/RFile.cxx
vepadulano: The cpp core guidelines suggests to implement rule of five even in case the special member function is `= default`ed https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all . But it is a guideline after all, so I'm not putting it as a blocker for the PR (in fact I already approved)

io/io/inc/ROOT/RFile.hxx
silverweed: Putting it in StringUtils potentially makes sense, but the thing is this function means "path" in the sense of "object path inside a ROOT file" specifically, not "filesystem path" (in fact it doesn't handle stuff like normalizing multiple `/` or considering `\\` as a path separator on windows), so it's a relatively specific thing. That's why putting it in RFile for me "scopes" it a little better (though maybe it should be even better scoped by having it into a specific RFile namespace).
All in all, I'm not so sure anymore that we should expose it after all.

pcanal: I would vote for `Details` in this file or not publish it (beyond where it is needed) ... (Side note: since it has not semantic, it seems that the effect is similar to a split (on character `/`) of the last component (with the unusual feature of keeping the trailing '/' in the left hand/basename).

io/io/inc/ROOT/RFile.hxx
pcanal: Fair enough.  Thanks.

io/io/src/RFile.cxx
silverweed: In this case, given it's a private, non-virtual struct, I would rather not add the noise of the "other four" declarations. After all the idea is to prevent the compiler from deleting the copy/move ctor/assignments when one of them is explicitly declared, but this doesn't happen when you only declare the dtor (as the core guideline sort-of-implicitly says).
This would really be a "rule of zero" case, were it not for the non-semantically relevant fact that we want to hide the TIterator.h inclusion from the header.

PR #20057, opened 2025-10-08
[Minuit2] Typesafe strategy handling

guitargeek:
It should not be possible to convert the MnStrategy object to an
integer, because that means the strategy can be used in arbitrary ways
inside the Minuit 2 algorithms.

It would be better if all parameters dependent on the strategy can be
retreived with explicit getter frunctions from the strategy object, such
that all strategy dependence is clear just from reading the public
interface.

To be followed up by improving the documentation.

### DIFF
--- a/math/minuit2/inc/Minuit2/MnStrategy.h
+++ b/math/minuit2/inc/Minuit2/MnStrategy.h
@@ -46,21 +46,11 @@ public:
 
    int StorageLevel() const { return fStoreLevel; }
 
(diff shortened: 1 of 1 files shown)


guitargeek:
> Given these improvements in the Strategy class will it be possible to have independent solutions, like do not compute initial Hessian, but always recompute the Hessian at the end (having a -infinity threshold)?

Yes, this is something we can do next if needed! Thanks for the review.

PR #20062, opened 2025-10-09
[ci] Disable `builtin_vc` on `mac15`

hahnjo:
It fails to build with the latest SDK because of missing includes and was already off for `mac26` and `mac-beta` (probably for that reason).


hahnjo:
The failure of `gtest-tree-ntuple-ntuple-evolution-shape` on `alma9` is addressed by https://github.com/root-project/root/pull/20067 - there are two tests that write to the same file.

PR #20065, opened 2025-10-09
[hist] Test user-defined bin content type, support weighted filling

hahnjo:
More discussions will be necessary to figure out if and how we want to support user-defined weights.

### DIFF
--- a/hist/histv7/test/hist_hist.cxx
+++ b/hist/histv7/test/hist_hist.cxx
@@ -120,6 +120,10 @@ TEST(RHist, FillWeight)
    hist.Fill(8.5, RWeight(0.8));
    hist.Fill(std::make_tuple(9.5), RWeight(0.9));
 
+   EXPECT_FLOAT_EQ(hist.GetBinContent(RBinIndex(8)), 0.8);
+   std::array<RBinIndex, 1> indices = {9};
+   EXPECT_FLOAT_EQ(hist.GetBinContent(indices), 0.9);
+
    EXPECT_EQ(hist.GetStats().GetNEntries(), 2);
    // Cross-checked with TH1
    EXPECT_FLOAT_EQ(hist.GetStats().ComputeNEffectiveEntries(), 1.9931034);


hahnjo:
ping

hist/histv7/inc/ROOT/RHistEngine.hxx
```diff
@@ -309,8 +307,7 @@ public:
    {
       auto t = std::forward_as_tuple(args...);
       if constexpr (std::is_same_v<typename Internal::LastType<A...>::type, RWeight>) {
-         static_assert(SupportsWeightedFilling,
-                       "weighted filling is only supported for floating-point bin content types");
+         static_assert(SupportsWeightedFilling, "weighted filling is not supported for integral bin content types");
```
pcanal, line 310: How hard would be it to add a test checking the failing case (i.e. requiring a failing compilation as success)?

pcanal (approved):
LGTM.

hist/histv7/inc/ROOT/RHistEngine.hxx
hahnjo: Quite hard: We would either need a small source file and somehow convince `ctest` that a failing compilation means success (no idea how to do that), or somehow do this via the interpreter (which requires the rest of ROOT). Using SFINAE to test is currently not an option with how it's written, that only works with substitution failures in the declarations but the `static_assert` is in the body.

hageboeck (approved):

hist/histv7/inc/ROOT/RHistEngine.hxx
pcanal: As a side note an example of testing for compilation failures

   https://ibob.bg/blog/2022/10/04/testing-build-failure-with-cmake/
   
Summary: the test runs CMake on either a target excluded from the default builds or using a standalone CMake file
and could be made generic enough ... but indeed this seems overkill for us (or similarly, very low priority).

PR #20068, opened 2025-10-09
[Math] Migrate from Vc to `std::experimental::simd` - show what happens for all platforms with `vecccore=ON`

guitargeek:
Do not merge. This is just for a demo CI run


guitargeek:
Not needed anymore: with the new suggestion to also drop the VecCore dependency, we can anyway enable the `std::experimental::simd` features on all supported platforms.

PR #20069, opened 2025-10-09
Deprecate `vc` and `builtin_vc` build options

guitargeek:
The `vc` flag enabled two features:

 1. Some vectorized Math functions, but they didn't get adopted much by our users because you need to use specific SIMD types.
 2. A vectorized backend for the VecCore-powered TFormula implementation, but this doesn't pass the unit tests.

Two drawbacks of removing it:
   1. If the user go through the effort of using the SIMD types for vectorization, we can expect them to implement the required math functions themselves (ROOT didn't provide many any way, `VectorizedTMath.cxx` had only 250 lines).
  2. In the cases where the VecCore/Vc backend for TFormula actually worked and the user enabled vectorization via the [TFormula constructor argument](https://root.cern.ch/doc/master/classTFormula.html#a32fcf259ba041e3f499a0a22de363165), there might be a performance regression.

The upside of removing it:

  1. We remove code that is untested in the CI because enabled it makes tests fail (see this [CI test run](https://github.com/root-project/root/pull/20068)). Leaving it might harm users with silently wrong results.
  2. Get rid on one (builtin) dependency.
  3. Reduced build option matrix.
  4. We don't need to worry about how/if builtin Vc headers should be available at install time.
  5. Vc is in **maintainance mode and no longer actively developed** (see it's [README](https://github.com/VcDevel/Vc))

Closes also the following issues:

  * https://its.cern.ch/jira/browse/ROOT-10614 - TFormula tests fail when vectorization is enabled
  * https://github.com/root-project/root/issues/18949


guitargeek:
The failure on `alma8` is unrelated: it is a artifact from upgrading Python on that platform and doing incremental builds, so it's still looking for the Python3.8 development files that are gone.

guitargeek:
Let's keep these options for now, and see how we can proceed with the migration to `std::simd` first.

PR #20073, opened 2025-10-09
[main] Introduce cmdline option parser to share some code between executables

silverweed:
We now have enough C++ executables (with more to be ported) to justify sharing some code between them, in particular option parsing.

This PR introduces a simple option parser class that covers most of our cases. Not all cases (e.g. `hadd` keeps its custom parsing because it's a bit weird), but it's already enough to cover `rootls` and `rootbrowse`, and will in the future also cover `rootcp`, `rootmv` etc.

The parser is documented with code examples and properly tested and it's about 300 lines of codes (about an order of magnitude less than e.g. [cxxopts](https://github.com/jarro2783/cxxopts/) which in my opinion is overkill for our use case).

## Checklist:

- [x] tested changes locally

### DIFF
--- a/main/src/optparse.hxx
+++ b/main/src/optparse.hxx
@@ -56,6 +56,7 @@
 #include <charconv>
 #include <cstring>
 #include <iostream>
+#include <limits>
 #include <optional>
 #include <sstream>
 #include <stdexcept>
@@ -208,9 +209,14 @@ public:
       static_assert(std::is_integral_v<T> || std::is_floating_point_v<T>);
 
       if (auto val = GetFlagValue(name); !val.empty()) {
+         // NOTE: on paper std::from_chars is supported since C++17, however some compilers don't properly support it
+         // (e.g. it's not available at all on MacOS < 26 and only the integer overload is available in AlmaLinux 8).
+         // There is also no compiler define that we can use to determine the availability, so we just use it only
+         // from C++20 and hope for the best.
+#if __cplusplus >= 202002L && !defined(__APPLE__)
          T converted;
          auto res = std::from_chars(val.data(), val.data() + val.size(), converted);
-         if (res.ptr == val.end() && res.ec == std::errc{}) {
+         if (res.ptr == val.data() + val.size() && res.ec == std::errc{}) {
             return converted;
          } else {
             std::stringstream err;
@@ -225,6 +231,32 @@ public:
             else
                throw std::invalid_argument(err.str());
          }
+#else
+         std::conditional_t<std::is_integral_v<T>, long long, long double> converted;
+         std::size_t unconvertedPos;
+         if constexpr (std::is_integral_v<T>) {
+            converted = std::stoll(std::string(val), &unconvertedPos);
+         } else {
+            converted = std::stold(std::string(val), &unconvertedPos);
+         }
+
+         const bool isOor = converted > std::numeric_limits<T>::max();
+         if (unconvertedPos != val.size() || isOor) {
+            std::stringstream err;
+            err << "Failed to parse flag `" << name << "` with value `" << val << "`";
+            if constexpr (std::is_integral_v<T>)
+               err << " as an integer.\n";
+            else
+               err << " as a floating point number.\n";
+
+            if (isOor)
+               throw std::out_of_range(err.str());
+            else
+               throw std::invalid_argument(err.str());
+         }
+
+         return converted;
+#endif
       }
       return std::nullopt;
    }
--- a/main/src/rootbrowse.cxx
+++ b/main/src/rootbrowse.cxx
@@ -76,7 +76,7 @@ static RootBrowseArgs ParseArgs(const char **args, int nArgs)
       return outArgs;
    }
 
-   if (opts.GetBooleanFlag("help")) {
+   if (opts.GetSwitch("help")) {
       outArgs.fPrintHelp = RootBrowseArgs::EPrintUsage::kLong;
       return outArgs;
    }
@@ -84,7 +84,7 @@ static RootBrowseArgs ParseArgs(const char **args, int nArgs)
    if (auto web = opts.GetFlagValue("web"); !web.empty())
       outArgs.fWeb = web;
 
-   if (opts.GetBooleanFlag("webOff"))
+   if (opts.GetSwitch("webOff"))
       outArgs.fWeb = "off";
 
    if (opts.GetArgs().empty())


main/src/optparse.hxx
```diff
@@ -0,0 +1,333 @@
...
+/// MyAppOpts ParseArgs(const char **args, int nArgs) {
+///    ROOT::RCmdLineOpts opts;
+///    // will parse '-c VAL', '--compress VAL' or '--compress=VAL'
+///    opts.AddFlag({"-c", "--compress"}, RCmdLineOpts::EFlagType::kWithArg);
+///    // will toggle a boolean flag '--recreate' (no args).
+///    opts.AddFlag({"--recreate"});
+///    opts.AddFlag({"-o"}, RCmdLineOpts::EFlagType::kWithArg);
+///
+///    // NOTE: `args` should not contain the program name! It should usually be `argc + 1`.
+///    opts.Parse(args, nArgs);
```
pcanal, line 18:
```suggestion
///    // NOTE: `args` should not contain the program name! It should usually be `argc + 1`.
///    // For example `main(char **argv, int argc)` might call this function as: 
///    //    `ParseArgs(const_cast<const_cast<const char**>(argv) + 1, argc - 1);`
///    opts.Parse(args, nArgs);
```
(Without this, I was side tracked for a while pondering whether the comment wanted the last line to be `opts.Parse(args, nArgs+1)` )

main/src/rootls.cxx
```diff
@@ -784,9 +746,9 @@ int main(int argc, char **argv)
    gErrorIgnoreLevel = kError;
 
    auto args = ParseArgs(const_cast<const char **>(argv) + 1, argc - 1);
```
pcanal, line 748:
```suggestion
   auto args = ParseArgs(argv + 1, argc - 1);
```
I.e. is the `const_cast` still needed?

pcanal: Nevermind, I mis-read the direction of the `const_cast` (it is adding the `const`)

main/test/optparse_test.cxx
```diff
@@ -0,0 +1,450 @@
...
+   EXPECT_TRUE(opts.GetBooleanFlag("f"));
+   EXPECT_FALSE(opts.GetBooleanFlag("g"));
+}
+
+TEST(OptParse, OptParseString)
+{
+   ROOT::RCmdLineOpts opts;
+   opts.AddFlag({"--foo"}, ROOT::RCmdLineOpts::EFlagType::kWithArg);
+
+   const char *args[] = {"--foo", "bar"};
```
pcanal, line 58: Do we test something like:
```
 const char *args[] = {"somename", "--foo", "bar", "-abc"};
```
and
```
 const char *args[] = {"somename", "--noargopt", "bar"};
```
```
 const char *args[] = {"--noargopt", "bar"};
```
i.e. location of positional arguments.  (In particular the first one might appear by mistake if `argv` is passed instead of `argv+1`)

ferdymercury:
Thanks a lot for the initiative!

Even if it's 300 lines of code, isn't it better to rely on something external that gets auto-maintained, even if it's bigger in number of lines of code?
Such as the header-only https://github.com/CLIUtils/CLI11 by @henryiii

Or to understand better the motivation, what do we gain by having a small parser for which we have to add our own tests?
- Is it because then compile time is smaller?
- Or is initialization time more performant?
- Or is the overkill of CLI11 or cxxopt increasing the size of the produced binaries or sth like that?
- Or to not rely on external sources?

silverweed:
> Thanks a lot for the initiative!
> 
> Even if it's 300 lines of code, isn't it better to rely on something external that gets auto-maintained, even if it's bigger in number of lines of code? Such as the header-only https://github.com/CLIUtils/CLI11
> 
> Or to understand better the motivation, what do we gain by having a small parser for which we have to add our own tests?
> 
>     * Is it because then compile time is smaller?
> 
>     * Or is initialization time more performant?
> 
>     * Or is the overkill of CLI11 or cxxopt increasing the size of the produced binaries or sth like that?
> 
>     * Or to not rely on external sources?

It's mostly the last point (though the first is also somewhat important).
Simply put, if the functionality is not particularly complex (as is this case) it's always better to have your own code that you know what it does and can easily control than an external dependency that you need to maintain and complicates your build system. 
Furthermore, if such dependency is 10x the code size (without providing at least 10x the benefits) then you would just be carrying around much more weight than you really need. 
External dependencies are not free and should only be used where there is a really good argument for them, which in this case I don't see.

main/test/optparse_test.cxx
silverweed: currently positional arguments may appear in any place, so in the first example `somename` would be the first (and only) positional arg, in the second `somename` and `bar` would be the positional args and in the third only `bar`.
I will add the tests for the first 2 cases (as the third I believe is already covered)

henryiii:
On the flipside, a library that is heavily used is more likely to come across all of the weird edge cases that are going to end up making the custom implementation significantly longer eventually.

When I designed CLI11, I did have ROOT in mind, and with that you could actually make it easy to have a user extend an application. That would definitely be more than 10x functionality.

silverweed:
> weird edge cases that are going to end up making the custom implementation significantly longer eventually.

This is an unprovable assumption as of now - if we start finding such edge cases regularly then this might tilt the decision towards an external library. It's very easy to change our mind later after all, this is all "leaf code", which brings me to the second point:

>you could actually make it easy to have a user extend an application

I don't see the point here: the user cannot "extend" a binary like `rootls` or `rootcp` as in "use it as a library", and if they want to write some sort of wrapper around them they are free to bring their own option parsing library. `optparse.hxx` is not meant to be used as a library, which is why it's in `src` and not in `inc/ROOT`.

> That would definitely be more than 10x functionality.

Again, this is conjecture which can only be proven with actually trying both things. My position is: let's start with the simpler option, then if we realize it was a mistake we can easily swap the implementation later.

main/src/optparse.hxx
```diff
@@ -0,0 +1,367 @@
...
+///
+/// Small utility to parse cmdline options.
+///
+/// Usage:
+/// ~~~{.cpp}
+/// MyAppOpts ParseArgs(const char **args, int nArgs) {
+///    ROOT::RCmdLineOpts opts;
+///    // will parse '-c VAL', '--compress VAL' or '--compress=VAL'
+///    opts.AddFlag({"-c", "--compress"}, RCmdLineOpts::EFlagType::kWithArg);
+///    // will toggle a boolean flag '--recreate' (no args).
```
jblomer:
```suggestion
///    // will toggle a switch '--recreate' (no args).
```

main/src/optparse.hxx
```diff
@@ -0,0 +1,367 @@
...
+#include <string>
+#include <string_view>
+#include <vector>
+
+namespace ROOT {
+
+class RCmdLineOpts {
+public:
+   enum class EFlagType {
+      kBoolean,
```
jblomer:
```suggestion
      kSwitch,
```

main/src/optparse.hxx
```diff
@@ -0,0 +1,367 @@
...
+      kWithArg
+   };
+
+   struct RFlag {
+      std::string fName;
+      std::string fValue;
```
jblomer, line 78: Should this be an optional to distinguish a switch from an empty string argument?

main/src/optparse.hxx
```diff
@@ -0,0 +1,367 @@
...
+      for (std::size_t i = 0; i < nArgs && fErrors.empty(); ++i) {
+         const char *arg = args[i];
+
+         if (strcmp(arg, "--") == 0) {
+            forcePositional = true;
+            continue;
+         }
+
+         bool isFlag = !forcePositional && arg[0] == '-';
+         if (isFlag) {
```
jblomer: Maybe swap the if-else blocks so that the short one is handled first (reduce scrolling).

main/src/optparse.hxx
```diff
@@ -0,0 +1,367 @@
...
+      for (const auto &err : fErrors)
+         stream << err << "\n";
+      return !fErrors.empty();
+   }
+
+   void AddFlag(std::initializer_list<std::string_view> aliases, EFlagType type = EFlagType::kBoolean,
+                std::string_view help = "")
+   {
+      int aliasIdx = -1;
+      for (auto f : aliases) {
```
jblomer, line 136: Perhaps: check that the alias is neither `-` nor contains a space (and possibly other funny stuff, newline, NULL, etc). Maybe enough to say in the documentation what is allowed.

jblomer (approved):
Nice! Two general comments:
  - I'd call "boolean flag" a switch
  - Perhaps call out in the initial documentation that repeated flags are not supported (e.g., no `root-ssh -vvv`)

main/src/optparse.hxx
silverweed: Is the distinction relevant? As far as I see it, it would only be relevant if the user is iterating the flags list and wants to know which flag is a switch and which is not; but the user *knows* the semantic value of the flags already (as he defined them), so my impression is that it's fine to collapse the two cases.
But maybe I'm missing some other benefits to this, feel free to let me know if you have any in mind

main/src/optparse.hxx
silverweed: The alias already cannot be `-` (or start with a `-` after the prefix) as that is prevented by the combination of the two `if`s; I could validate the characters, but I think it's a bit overly defensive (the user will almost always define these aliases as literal strings and probably won't shoot themselves in the foot by using weird characters in flags)

PR #20074, opened 2025-10-09
[io] ROOT::Detail::TKeyMapIterable: return ref in operator++()

silverweed:
The [requirements](https://en.cppreference.com/w/cpp/named_req/Iterator.html) for an iterator mandate that `operator++()` returns a reference to the iterator, not a copy.


silverweed:
Update: I also changed the iterator tag from Forward to Input (see commit description)

hahnjo (approved):

PR #20078, opened 2025-10-09
[ntuple] add evolution test for atomic<class>

jblomer:
Follow-up of (and based on) #19937 

Linking of libatomic needs to be sorted out.

### DIFF
--- a/tree/ntuple/src/RField.cxx
+++ b/tree/ntuple/src/RField.cxx
@@ -1165,10 +1165,12 @@ std::unique_ptr<ROOT::RFieldBase> ROOT::RAtomicField::CloneImpl(std::string_view
 
 void ROOT::RAtomicField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
 {
-   static const std::vector<std::string> prefixes = {"std::atomic<"};
-
-   EnsureMatchingOnDiskField(desc, kDiffTypeName).ThrowOnError();
-   EnsureMatchingTypePrefix(desc, prefixes).ThrowOnError();
+   const auto &fieldDesc = desc.GetFieldDescriptor(GetOnDiskId());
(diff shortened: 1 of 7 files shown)


tree/ntuple/src/RField.cxx
```diff
@@ -1165,10 +1165,12 @@ std::unique_ptr<ROOT::RFieldBase> ROOT::RAtomicField::CloneImpl(std::string_view
 
 void ROOT::RAtomicField::ReconcileOnDiskField(const RNTupleDescriptor &desc)
 {
-   static const std::vector<std::string> prefixes = {"std::atomic<"};
-
-   EnsureMatchingOnDiskField(desc, kDiffTypeName).ThrowOnError();
-   EnsureMatchingTypePrefix(desc, prefixes).ThrowOnError();
+   const auto &fieldDesc = desc.GetFieldDescriptor(GetOnDiskId());
+   if (fieldDesc.GetTypeName().rfind("std::atomic<", 0) == 0) {
```
pcanal, line 1169: Is this the 'right' test or would `fieldDesc.IsStdAtomic()` be sufficient/equivalent?

tree/ntuple/test/CustomStruct.hxx
```diff
@@ -36,6 +36,11 @@ enum class CustomEnumUInt32 : unsigned int {};
 enum class CustomEnumInt64 : long int {};
 enum class CustomEnumUInt64 : unsigned long int {};
 
+// Used for std::atomic tests as an example of a class that is not lock-free.
```
pcanal, line 39:
```suggestion
// Used for std::atomic tests as an example of a class whose corresponding std::atomic is not lock-free.
```

pcanal (approved):
LFTM.

PR #20079, opened 2025-10-09
[skip-ci][NFC][ntuple] add schema evolution docs

jblomer:


tree/ntuple/doc/SchemaEvolution.md
```diff
@@ -0,0 +1,209 @@
...
+into in-memory models that are different but compatible to the on-disk schema.
+
+Schema evolution allows for data models to evolve over time
+such that old data can be read into current models ("backward compatibility")
+and old software can read newer data models ("forward compatibility").
+For instance, data model authors may over time add and reorder class members, change data types
+(e.g. `std::vector<float>` --> `ROOT::RVec<double>`), rename classes, etc.
+
+ROOT applies automatic schema evolution rules for common, safe and unambiguous cases.
+Users can complement the automatic rules by manual schema evolution ("I/O read rules")
```
pcanal: We have been waffling on the exact wording for the name of the rule but we have use the following wording quite a bit:
```suggestion
Users can complement the automatic rules by manual schema evolution ("I/O customization rules")
```

tree/ntuple/doc/SchemaEvolution.md
```diff
@@ -0,0 +1,209 @@
...
+where custom code snippets implement the transformation logic.
+In case neither automatic nor manual schema evolution rules suffice
```
pcanal:
```suggestion
In case neither automatic nor provided manual schema evolution rules suffice
```
Or
```suggestion
In case neither automatic nor any of the provided I/O customization rules suffice
```

tree/ntuple/doc/SchemaEvolution.md
```diff
@@ -0,0 +1,209 @@
...
+to transform the on-disk schema into the in-memory model, ROOT will error out and refrain from reading data.
+
+This document describes schema evolution in RNTuple.
```
pcanal:
```suggestion
This document describes schema evolution support implemented in RNTuple.
```
(maybe)

tree/ntuple/doc/SchemaEvolution.md
pcanal, line 35:
```suggestion
Note for classes that have been given a version number, users must increase the class version number whenever the persistent layout changes (members marked as transient do not participate in the persistent layout).
```
In addition there is a quirk that we currently also need to increase the version number of a derived class if the base class changes.  It is a really annoying feature and we should 'fix' it and thus it may or may not be worth mentioned it here.

pcanal: What do you mean? The `TClass` will be confused and issue an error/warning message if there is 2 StreamerInfo with the same version number but different checksum.

In the subset of cases where the StreamerInfo for the class stored in RNTuple is not stored in the file, then the above might be true as then the RNTuple layout is then clearly known to be its own but having the same version number marked in the RNTuple and the in memory layout will make the use of rules 'harder'.

Note that this is distinct from the case where the class is not (user) versioned at all.  In this case each schema layout is distinct and look up is based on the CheckSum value.

pcanal: This may or may not deserve as asterix.  I am guess the bound check is based on the underlying type rather than on the list of 'valid'/'explicit' enums values.

pcanal: Does we (eventually) want a (possibly optional) bound check here?

pcanal, line 87: Should we add a mention of the cases with fixed length collections (both vs other fixed length collections and variable collections).  For example are the following supported?:
| In-memory type                   | Compatible on-disk types             | 
| -------------------------------- | ------------------------------------ | 
| int[8] | int[4] | 
| std::vector\<int\> | int[4] | 
| std::array\<int,4\> | int[4] |

And vice et versa.

pcanal: Note: std::array -> variable-length collection is already mentioned (but not the reverse).

pcanal, line 54: Should this mention (as a preview of the text) below whether this include user types related via a `renaming rule`?

pcanal, line 103: `std::list` and `std::deque` are not explicitly listed, are they supported here?  (If they are not actively disallowed then they will likely just 'work' via the 'User-defined collection' code path (i.e. they have a CollectionProxy)).

pcanal, line 115: This might deserve a note/comment/asterix hinting at the reason why other transformation can not be supported.

pcanal: (minor)

```suggestion
For the exact syntax of customization rules, please refer to the ROOT manual.
```

pcanal, line 188: Should we mention that if the `given type` differs from the `on file type`, the source member will undergo schema evolution before being passed to the rule's function?

pcanal:
```
A source member can be read them into any type compatible to its on-disk type
```
What is the (them) part refering to?

tree/ntuple/doc/SchemaEvolution.md
```diff
@@ -0,0 +1,210 @@
...
+
+User-defined classes can automatically evolve their layout in the following ways.
+Note that users should increase the class version number when the layout changes.
+However, for RNTuple automatic rules that is not mandatory;
+RNTuple will always compare the current on-disk layout with the in-memory model.
+
+| Layout Change                           | Also supported in Untyped Records | Comment              |
+| --------------------------------------- | --------------------------------- | -------------------- |
+| Remove member                           | Yes                               | Match by member name |
+| Add member                              | Yes                               | Match by member name |
```
silverweed: Should we specify in the comment that new members are default initialized?

tree/ntuple/doc/SchemaEvolution.md
```diff
@@ -0,0 +1,210 @@
...
+| `bool`                      | `char`                      |                      |
+|                             | `std::[u]int[8,16,32,64]_t` |                      |
+|                             | enum                        |                      |
+|-----------------------------|-----------------------------|----------------------|
+| `char`                      | `bool`                      |                      |
+|                             | `std::[u]int[8,16,32,64]_t` | with bounds check    |
+|                             | enum                        | with bounds check    |
+|-----------------------------|-----------------------------|----------------------|
+| `std::[u]int[8,16,32,64]_t` | `bool`                      |                      |
+|                             | `char`                      |                      |
```
silverweed: Probably we should specify the behavior when the value doesn't fit in the `char`

tree/ntuple/doc/SchemaEvolution.md
```diff
@@ -0,0 +1,210 @@
...
+|                             | `std::[u]int[8,16,32,64]_t` | with bounds check    |
+|                             | enum                        | with bounds check    |
+|-----------------------------|-----------------------------|----------------------|
+| enum                        | enum of different type      | with bounds check    |
+|-----------------------------|-----------------------------|----------------------|
+| float                       | double                      |                      |
+|-----------------------------|-----------------------------|----------------------|
+| double                      | float                       |                      |
```
silverweed: No bounds check?

tree/ntuple/doc/SchemaEvolution.md
```diff
@@ -0,0 +1,210 @@
...
+If a class property changes from using an RNTuple streamer field to a using regular RNTuple class field,
+existing files with on-disk streamer fields will continue to read as streamer fields.
+This can be seen as "schema evolution out of streamer fields".
+
+## Manual schema evolution (I/O customization rules)
+
+ROOT I/O customization rules allow for custom code handling the transformation
+from the on-disk schema to the in-memory model.
+Customization rules are part of the class dictionary.
+For the exact syntax of customization rules, we refer to the ROOT manual.
```
silverweed: Maybe a link to the manual here?

tree/ntuple/doc/SchemaEvolution.md
```diff
@@ -0,0 +1,210 @@
...
+
+Generally, customization rules consist of
+  - A target class.
+  - Target members of the target class, i.e. those class members whose value is set by the rule.
+    Target members must be direct members, i.e. not part of a base class.
+  - A source class (possibly having a different class name than the target class)
+    together with class versions or class checksums
```
silverweed, line 185: It could be useful to mention that the class checksum can be retrieved from `TClass::GetCheckSum()`

tree/ntuple/doc/SchemaEvolution.md
```diff
@@ -0,0 +1,210 @@
...
+    that describe all the possible on-disk class versions the rule applies to.
+  - Source members of the source class; the given source members will be read as the given type.
+    Source members can also be from a base class.
+    Note that there is no way to specify a base class member that has the same name as a member in the derived class.
+  - The custom code snippet; the code snippet has access to the (whole) target object and to the given source members.
```
silverweed, line 192: I think an example of a customization rule would be very useful to anyone not already familiar with them.

tree/ntuple/doc/SchemaEvolution.md
hahnjo: Yes, we discussed a check based on the floating-point class.

tree/ntuple/doc/SchemaEvolution.md
hahnjo: I think for this line of the table, it's not needed because `double` can store a superset of values that can be represented by `float` on-disk.

tree/ntuple/doc/SchemaEvolution.md
hahnjo: Neither are supported in RNTuple at the moment: https://github.com/root-project/root/blob/master/tree/ntuple/doc/BinaryFormatSpecification.md#stdlib-types-and-collections

hahnjo:
```suggestion
| `ROOT::RVec<T>`                  | `std::vector<T'>`                    | with size check                       |
```

hahnjo: `std::vector` and / or `ROOT::RVec`

hahnjo: `std::vector` and / or `ROOT::RVec`

hahnjo, line 167: This is now listed in the table, do we need / want to discuss it separately?

tree/ntuple/doc/SchemaEvolution.md
silverweed: Ah right, this is the "read" table - then I guess I meant the line above

tree/ntuple/doc/SchemaEvolution.md
pcanal: Unless we actively do 'something' explicit about it, the `std::list` (and likely `std::deque`) should be storable via the `CollectionProxy`.   (I.e. technically they would be handled here in the `User-defined collection`)

hahnjo: Yes, that's indeed a problem since we only have the check in `RClassField`:
https://github.com/root-project/root/blob/fe2834d58f1604e273df5f3accea66fcf749387f/tree/ntuple/src/RFieldMeta.cxx#L134-L136

```
root [0] ROOT::RFieldBase::Create("l", "std::list<int>").Unwrap()
(std::unique_ptr<ROOT::RFieldBase, std::default_delete<ROOT::RFieldBase> >) std::unique_ptr -> 0x3737380
root [1] typeid(*ROOT::RFieldBase::Create("l", "std::list<int>").Unwrap().get()).name()
(const char *) "N4ROOT23RProxiedCollectionFieldE"
```

jblomer: See #20206

jblomer: I think the current sentence leaves room both interpretations, so perhaps let's for the moment not make this question subject of this document.

jblomer: Yes, I didn't consider the bigger picture, removed the two lines.

jblomer: Only the reverse is in the table

jblomer: removed `them`

enirolf, line 169:
```suggestion
If a class property changes from using an RNTuple streamer field to using a regular RNTuple class field,
```

enirolf: Perhaps worth adding a link:


```suggestion
For the exact syntax of customization rules, please refer to the [ROOT manual](https://root.cern/manual/io/#dealing-with-changes-in-class-layouts-schema-evolution).
```

enirolf, line 238: Isn't this the case for RNTuple? It is mentioned in the table, line 134.

jblomer: Done.

jblomer: There is no special automatic evolution for fixed-length collections (other than the recursive evolution of the elements). At the moment, the length of the array must not change and there is no automatic transformation from variable-length to fixed-length collection.

For `int[4]` and `std::array<int, 4>`, they are considered different spellings and both end up on disk as `std::array<int, 4>` due to the type normalization.

jblomer: Perhaps github got confused, isn't the suggestion what's already there?

jblomer: It's only the reverse that works, `T'` on disk into `std::unique_ptr<T>`.

pcanal: So it may be worth mentioning the part "At the moment, the length of the array must not change and there is no automatic transformation from variable-length to fixed-length collection." (unless it is already mentioned :) )

pcanal: Why block it?  A priori it works 'perfectly' through  the CollectionProxy without intervention. (except for the user needing to generate the dictionary).

jblomer: We decided that we want to have control over which stdlib types we enable and only allow them explicitly. We can extend the current list but then this should be a conscious decision and involve updating the specs.

jblomer: Added a note in a separate section

jblomer: Added an introduction paragraph to the section

hahnjo:
```suggestion
| `ROOT::RVec<T>`                  | `std::vector<T'>`                    | with size check                       |
```

hahnjo: Thinking about this again, it may lose entries because the set is unique on the pair of `K, V` while the map enforces uniqueness of `K`. Not sure if this is a problem in practice...

hahnjo (approved):
LGTM, two minor comments

tree/ntuple/doc/SchemaEvolution.md
enirolf: Ah yes, I see now. I confused the table column headers :)

PR #20080, opened 2025-10-10
Fix warnings when compiling with Clang 21

guitargeek:
Fixing warnings like:
```txt
In file included from /home/rembserj/code/root/root_src/core/metacling/src/TClingCallFunc.cxx:32:
/home/rembserj/code/root/root_src/core/metacling/src/TCling.h:372:17: warning: virtual method 'GetFunctionName' is inside a 'final' class and can never be overridden [-Wunnecessary-virtual-specifier]
  372 |    virtual void GetFunctionName(const clang::Decl *decl, std::string &name) const;
      |                 ^
1 warning generated.
```
or
```
/home/rembserj/code/root/root_src/tree/ntuple/test/ntuple_parallel_writer.cxx:97:29: warning: variable 'pt1' is uninitialized when passed as a const pointer argument here [-Wuninitialized-const-pointer]
   97 |       e1->BindRawPtr("pt", &pt1);
      |
```

### DIFF
--- a/core/metacling/src/TCling.h
+++ b/core/metacling/src/TCling.h
@@ -191,7 +191,7 @@ protected:
 
 public: // Public Interface
 
-   virtual ~TCling();
+   ~TCling() final;
    TCling(const char* name, const char* title, const char* const argv[], void *interpLibHandle);
 
    void    AddIncludePath(const char* path) final;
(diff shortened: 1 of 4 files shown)


core/metacling/src/TCling.h
```diff
@@ -219,11 +219,11 @@ class TCling final : public TInterpreter {
    const char* GetClassSharedLibs(const char* cls, bool skipCore = true) final;
    const char* GetSharedLibDeps(const char* lib, bool tryDyld = false) final;
    const char* GetIncludePath() final;
-   virtual const char* GetSTLIncludePath() const final;
+   const char* GetSTLIncludePath() const final override;
```
vepadulano: `final` implies `override`, check also similar occurrences.

core/metacling/src/TCling.h
```diff
@@ -191,7 +191,7 @@ class TCling final : public TInterpreter {
 
 public: // Public Interface
 
-   virtual ~TCling();
+   ~TCling() override;
```
vepadulano: This would be a nitpick, but between `override` and `final` here I would prefer the latter simply for consistency with other methods of this class which specify `final` already

guitargeek: I agree. Thanks!

core/metacling/src/TCling.h
guitargeek: Right,  forgot about that! Thanks for the catch

guitargeek:
Thanks @vepadulano for the review! I have addressed your comments.

PR #20082, opened 2025-10-11
[RF] Consistent minimizer defaults in `RooAbsPdf::fitTo()`

guitargeek:
Take the default offsetting flag value from the default `MinimizerConfig` struct, where the offsetting flag is correctly initialized to `int doOffset = -1`:
https://github.com/root-project/root/blob/master/roofit/roofitcore/src/FitHelpers.cxx#L247

This means to not change any offset setting by default.

The problem was that the default was set to zero before, which tells the minimizer to explicitly disable offsetting, even if the likelihood was created with offsetting.

This fixes an issue reported on the forum:
https://root-forum.cern.ch/t/cannot-use-offset-with-rooabspdf-fitto/64293

Also, make sure that all other default config values are taken from the default options struct for consistency.

To be backported to the previous active ROOT releases.

### DIFF
--- a/roofit/roofitcore/src/FitHelpers.cxx
+++ b/roofit/roofitcore/src/FitHelpers.cxx
@@ -246,7 +246,7 @@ struct MinimizerConfig {
    int maxCalls = -1;
    int doOffset = -1;
    int parallelize = 0;
-   bool enableParallelGradient = true;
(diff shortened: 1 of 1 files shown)


dpiparo:
Fixes #20083

PR #20086, opened 2025-10-12
[Python] Raise minimum required Python version from 3.8 to 3.9

guitargeek:
Python 3.8 has been end-of-life for one year already. The last platform that used it was AlmaLinux 8, but also there it has reached end of life. Instead, RHEL has committed to support Python 3.12 until the end-of-life of RHEL 8 itself:
https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle

Therefore, our CI images for `alma8` also use Python 3.12 now.

And although Python 3.9 has also reached end-of-life now, we still need to support it because it's the system Python version on macOS.

See:
https://devguide.python.org/versions/

### DIFF
--- a/.github/workflows/python_wheel_build.yml
+++ b/.github/workflows/python_wheel_build.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        target: [cp38-manylinux_x86_64, cp39-manylinux_x86_64, cp310-manylinux_x86_64, cp311-manylinux_x86_64, cp312-manylinux_x86_64, cp313-manylinux_x86_64]
+        target: [cp39-manylinux_x86_64, cp310-manylinux_x86_64, cp311-manylinux_x86_64, cp312-manylinux_x86_64, cp313-manylinux_x86_64]
     name: ${{ matrix.target }}
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


guitargeek:
> Thank you! I wasn't sure we could already make the move because of RHEL, but with the extra info you provide I think it's safe. We could even discuss about this for MacOS, in the end on that platform the "system Python version" is not as clearly defined, because versions of the official Python binaries can be downloaded at https://www.python.org/downloads/macos/

Yes, we can think about that, but not sure we should do it now given that the macOS runners seem kind of unstable to me :slightly_smiling_face:

Also note that `alma9` is still using Python 3.9, and different from the `alma8` situation, Red Hat has not committed to supporting any other Python version until the end of the RHEL 9 lifetime. If they follow the same pattern as RHEL 8, we can expect them to ship Python 3.15 in 2027 (see the section "RHEL 8 Full Life Application Streams Release Life Cycle" in the link a posted above). And we have to support Python 3.10 as well until 2027 because of Ubuntu 22.04.

So as I see it, the next minimum Python version requirement raise we can make with our current approach for Linux is Python 3.9 -> 3.12 in April 2027. That means finding a solution for macOS to upgrade the Python version is not urgent.

> Another question, when you say that our alma8 CI runs python 3.12, is it because you modified the image configuration or because the alma8 runner picks up the latest Python by default?

I have explicitly updated it in `root-ci-images`:
https://github.com/root-project/root-ci-images/commit/1b5658fd92a942e54ef4da8322130821976cd0af#diff-9b3e6b0f4447054ea45151f30aab4cfb65ae8ee729909d5721dae43908d757e4

> Lastly, I believe this PR is just missing a line in the release notes

Done!

PR #20093, opened 2025-10-13
[RF] generateAsimov support RooProdPdf of RooSimultaneous

ferdymercury:
Explores solution proposed by user here: https://its.cern.ch/jira/browse/ROOT-4900


guitargeek (approved):
Thank you very much! Makes sense to me :+1:

PR #19795, opened 2025-09-01
[build] Support source or build directory with spaces

ferdymercury:
With these changes, I was able to build ROOT with -Dminimal=ON locally, using a space in both the source dir and the build dir.

Trying now on the CI.

Fixes https://github.com/root-project/root/issues/7311

PR #20098, opened 2025-10-13
Remove quotes around DEFAULT_ROOT_INCLUDE_PATH to address build failure on stacks

LukasBreitwieser:
as reported by Andre Sailer on #19179

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -579,10 +579,10 @@ add_custom_target(hsimple ALL DEPENDS tutorials/hsimple.root)
 add_dependencies(hsimple onepcm)
 if(WIN32)
   set(hsimple_cmd COMMAND ${CMAKE_COMMAND} -E env PATH="${CMAKE_RUNTIME_OUTPUT_DIRECTORY}\\\;%PATH%"
-                          ROOT_INCLUDE_PATH="${DEFAULT_ROOT_INCLUDE_PATH}"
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -579,10 +579,10 @@ add_custom_target(hsimple ALL DEPENDS tutorials/hsimple.root)
 add_dependencies(hsimple onepcm)
 if(WIN32)
   set(hsimple_cmd COMMAND ${CMAKE_COMMAND} -E env PATH="${CMAKE_RUNTIME_OUTPUT_DIRECTORY}\\\;%PATH%"
-                          ROOT_INCLUDE_PATH="${DEFAULT_ROOT_INCLUDE_PATH}"
+                          ROOT_INCLUDE_PATH=${DEFAULT_ROOT_INCLUDE_PATH}
```
pcanal, line 582: Can you describe in more details (and/or extract the relevant bits of the referenced issue) what is being address here?   In first approximation, this looks like it would make support for paths with spaces harder (i.e. this part would stop working).

ferdymercury: Support for spaces was already discontinued, see https://github.com/root-project/root/pull/19795
And https://github.com/root-project/root/pull/20054

andresailer: The problem is that CMake escapes the quotes when the command is executed (at least on Linux) which renders the ROOT_INCLUDE_PATH broken.

Issue #13843, opened 2023-10-11
RConcurrentHashColl.hxx 


pcanal:
### Explain what you would like to see improved and how.

Improve RConcurrentHashColl by moving it declaration/implementation from Thread to IO and by renaming some of the functions to clarify their intent.  See
https://github.com/root-project/root/pull/12847/files#r1232327876
and
https://github.com/root-project/root/pull/12847/files#r1232331243

### ROOT version

master

### Installation method

any

### Operating system

any

### Additional context

Related to the solution for https://github.com/root-project/root/issues/12783

pcanal:
See also https://github.com/root-project/root/pull/12847#discussion_r1232366997

PR #20105, opened 2025-10-14
[io] Move RConcurrentHashColl from core/thread to io/io

ferdymercury:
Fixes https://github.com/root-project/root/issues/13843

It's a utility for StreamerInfo, not a threading utility. It was historically in core/thread since originally it was used only in multi-thread environments.

Rename a internal function to clarify its intent and improve docu.


guitargeek (approved):
Thank you very much!

Issue #8518, opened 2021-06-23
Improve diagnostic information about module rebuilds


chrisburr:
### Is your feature request related to a problem? Please describe.

Currently it's hard to debug issues that happen when the modules are rebuilt as there is no information displayed about why ROOT decided to rebuild them. For an example see https://github.com/conda-forge/root-feedstock/issues/136.

### Describe the solution you'd like

When rebuilding modules information will be displayed such as: `/usr/include/foo.h got updated, must regenerate module bar`

### Additional context

I've tried a few thing to try and force a module rebuild, e.g. editing headers, touching files and installing different versions of dependencies but ROOT never seems to notice. The best I've managed is to delete all `lib/*.pcm` files which obviously triggers them to be rebuilt:

```bash
$ root
input_line_1:1:10: remark: building module 'std' as '/home/cburr/miniconda3/envs/test/lib/std.pcm' [-Rmodule-build]
#include <new>
         ^
input_line_1:1:10: remark: finished building module 'std' [-Rmodule-build]
input_line_2:1:10: remark: building module 'Cling_Runtime' as '/home/cburr/miniconda3/envs/test/lib/Cling_Runtime.pcm' [-Rmodule-build]
#include "cling/Interpreter/RuntimeUniverse.h"
         ^
While building module 'Cling_Runtime' imported from input_line_2:1:
In file included from <module-includes>:3:
/home/cburr/miniconda3/envs/test/etc//cling/Interpreter/DynamicLookupRuntimeUniverse.h:16:10: remark: building module 'Cling_Runtime_Extra' as '/home/cburr/miniconda3/envs/test/lib/Cling_Runtime_Extra.pcm' [-Rmodule-build]
#include "cling/Interpreter/DynamicExprInfo.h"
         ^
```

I don't know if ROOT is supposed to be able to recover from this but I run in to three issues, the first is the known `tbb` dependency issue in 6.24.0 that is already fixed for the next patch release.

The second issue is that `MathCore.pcm` fails to build with:

```bash
<<< cling interactive line includer >>>: remark: building module 'MathCore' as '/home/cburr/miniconda3/envs/test/lib/MathCore.pcm' [-Rmodule-build]
While building module 'MathCore':
In file included from <module-includes>:3:
In file included from /home/cburr/miniconda3/envs/test/include/Fit/Chi2FCN.h:19:
[...]

Issue #7704, opened 2021-03-26
Is the error message about out of data pcm inverted ?


pcanal:
Hi,

After touch a header file in `io/io/inc` and running **only** `ninja RIO` then running `root.exe -b -l -q` I get:
```
<<< cling interactive line includer >>>: fatal error: module file '/home/pcanal/root_working/build/master-debug/lib/RIO.pcm' is out of date and needs to be rebuilt: module file out of date
<<< cling interactive line includer >>>: note: imported by module 'Hist' in '/home/pcanal/root_working/build/master-debug/lib/Hist.pcm'
....
```

This seems "reversed and indeed:

```
ls -lrt lib/RIO.pcm lib/Hist.pcm
-rw-r--r-- 1 pcanal pcanal 5537736 Mar 25 21:11 lib/Hist.pcm
-rw-r--r-- 1 pcanal pcanal 3220264 Mar 25 21:13 lib/RIO.pcm
```

Maybe either:
```
<<< cling interactive line includer >>>: fatal error: module file '/home/pcanal/root_working/build/master-debug/lib/Hist.pcm' is out of date and needs to be rebuilt: module file out of date
<<< cling interactive line includer >>>: note: imported module 'RIO' in '/home/pcanal/root_working/build/master-debug/lib/RIO.pcm' is newer.
....
```

```
<<< cling interactive line includer >>>: fatal error: module file '/home/pcanal/root_working/build/master-debug/lib/RIO.pcm' is newer and all dependent module needs to be rebuilt
<<< cling interactive line includer >>>: note: imported by module 'Hist' in '/home/pcanal/root_working/build/master-debug/lib/Hist.pcm'
....
```

Cheers,
Philippe.

PS Full error message:
```
<<< cling interactive line includer >>>: fatal error: module file '/home/pcanal/root_working/build/master-debug/lib/RIO.pcm' is out of date and needs to be rebuilt: module file out of date
<<< cling interactive line includer >>>: note: imported by module 'Hist' in '/home/pcanal/root_working/build/master-debug/lib/Hist.pcm'
Failed to load module Hist
Failed to load module RDAVIX
Failed to load module RooFitMore
Failed to load module RooFitRDataFrameHelpers
Failed to load module FITSIO
Failed to load module FFTW
Failed to load module MathMore
[...]

PR #20106, opened 2025-10-14
[cmake] Add explicit RIO dependency in MathCore

silverweed:
As an attempt to fix the build issues ("RIO.pcm is out of date") seen e.g. [here](https://github.com/root-project/root/pull/20074) on debian (due to `dev=on`)


ferdymercury:
Thanks. One question: isn't the failure happening in the generate-dictionary of tree/ntuple (G__ROOTNTuple.cxx) rather than in MathCore?

Related solution: https://github.com/root-project/root/pull/18858
and issues:
https://github.com/root-project/root/issues/7704
https://github.com/root-project/root/issues/8518

silverweed:
> Thanks. One question: isn't the failure happening in the generate-dictionary of tree/ntuple (G__ROOTNTuple.cxx) rather than in MathCore?
> 
> Related solution: #18858 and issues: #7704 #8518

Not sure, but considering RNTuple already lists RIO in its dependencies I thought MathCore was more likely to be the culprit (if the two dictionaries are built in parallel it would make sense - then again, no idea if they are)

hageboeck:
I'm not convinced yet that this is the solution. See this:
```
echo "  ROOT/RRawFile.hxx
  ROOT/RRawFileTFile.hxx
  ${rawfile_local_headers}
  ROOT/TBufferMerger.hxx
  TArchiveFile.h
  TBufferFile.h
  TBufferText.h
  TBufferIO.h
  TBufferJSON.h
  TCollectionProxyFactory.h
  TContainerConverters.h
  TEmulatedMapProxy.h
  TEmulatedCollectionProxy.h
  TDirectoryFile.h
  TFileCacheRead.h
  TFileMerger.h
  TFree.h
  TFileCacheWrite.h
  TFilePrefetch.h
  TFile.h
  TFPBlock.h
  TGenCollectionStreamer.h
  TGenCollectionProxy.h
  TKey.h
  TKeyMapFile.h
  TLockFile.h
  TMemFile.h
  TMapFile.h
  TMakeProject.h
  TStreamerInfoActions.h
  TVirtualCollectionIterators.h
  TStreamerInfo.h
  TZIPFile.h" | while read file; do git grep $file math/mathcore/; done
```

This is all headers listed in RIO's CMake, and checking if they are used anywhere in mathcore. The only hit is `math/mathcore/test/stress`, but this one explicitly lists that dependency. We would therefore introduce a spurious dependency.
I'm not sure that this is really causing the pcm problems, so introducing an unneeded dependency would go against all efforts of reducing those dependencies that we tried to do in the past.

silverweed:
Thank you @hageboeck, I mistakenly thought RIO had a dependency on TFile but then you made me realize it's specifically avoided via an eval-style trick.
I guess we need to dig deeper.
Closing this.

PR #20112, opened 2025-10-15
[ntuple] Merger: avoid resealing if not needed

silverweed:
Currently we either "fast merge" the page (i.e. we copy it as-is to the destination) or we reseal it if fast merging is not possible (which happens where the compression or the on-disk encoding don't match).
There is a middle ground where we can avoid resealing if only the page compression differs but not the on-disk encoding.
We were not currently leveraging this and doing the extra work of resealing the pages in all cases where just recompression could be done.

With this PR we introduce this middle ground case ("L2 merging", aka "recompress merging") which skips the resealing (thus the re-encoding) of the pages whenever possible.

([see here](https://codimd.web.cern.ch/s/xI6A5V4_w#) for the L2 terminology)

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/src/RNTupleMerger.cxx
+++ b/tree/ntuple/src/RNTupleMerger.cxx
@@ -301,6 +301,7 @@ struct RResealFunc {
    ROOT::Internal::RPageAllocator &fPageAlloc;
    std::uint8_t *fBuffer;
    std::size_t fBufSize;
+   const ROOT::RNTupleWriteOptions &fWriteOpts;
 
    void operator()() const
    {
@@ -310,7 +311,7 @@ struct RResealFunc {
       sealConf.fPage = &page;
       sealConf.fBuffer = fBuffer;
       sealConf.fCompressionSettings = *fMergeOptions.fCompressionSettings;
-      sealConf.fWriteChecksum = fSealedPage.GetHasChecksum();
+      sealConf.fWriteChecksum = fWriteOpts.GetEnablePageChecksums();
       assert(fBufSize >= fSealedPage.GetDataSize() + fSealedPage.GetHasChecksum() * sizeof(std::uint64_t));
       auto refSealedPage = RPageSink::SealPage(sealConf);
       fSealedPage = refSealedPage;
@@ -916,6 +917,7 @@ RNTupleMerger::MergeCommonColumns(ROOT::Internal::RClusterPool &clusterPool,
                   *fPageAlloc,
                   buffer.get(),
                   bufSize,
+                  mergeData.fDestination.GetWriteOptions()
                });
             } else {
                RTaskVisitor{fTaskGroup}(RChangeCompressionFunc{


tree/ntuple/src/RNTupleMerger.cxx
```diff
@@ -267,6 +267,38 @@ struct RChangeCompressionFunc {
...
+   {
+      assert(fSrcColElement.GetIdentifier() == fDstColElement.GetIdentifier());
+
+      const auto bytesPacked = fSrcColElement.GetPackedSize(fSealedPage.GetNElements());
+      // TODO: this buffer could be kept and reused across pages
+      auto unzipBuf = MakeUninitArray<unsigned char>(bytesPacked);
+      ROOT::Internal::RNTupleDecompressor::Unzip(fSealedPage.GetBuffer(), fSealedPage.GetDataSize(), bytesPacked,
+                                                 unzipBuf.get());
+
+      // TODO: maybe we should check writeOpts.fWriteChecksum rather than fSealedPage.GetHasChecksum()...
```
jblomer: Probably.. or else how do we ensure that the correct checksum bit ends up in the page list?

silverweed: Changed

tree/ntuple/src/RNTupleMerger.cxx
```diff
@@ -267,6 +267,38 @@ struct RChangeCompressionFunc {
...
+
+   void operator()() const
+   {
+      assert(fSrcColElement.GetIdentifier() == fDstColElement.GetIdentifier());
+
+      const auto bytesPacked = fSrcColElement.GetPackedSize(fSealedPage.GetNElements());
+      // TODO: this buffer could be kept and reused across pages
+      auto unzipBuf = MakeUninitArray<unsigned char>(bytesPacked);
+      ROOT::Internal::RNTupleDecompressor::Unzip(fSealedPage.GetBuffer(), fSealedPage.GetDataSize(), bytesPacked,
+                                                 unzipBuf.get());
```
hahnjo, line 283: As discussed, we should validate the checksum before decompressing

tree/ntuple/src/RNTupleMerger.cxx
```diff
@@ -267,6 +267,38 @@ struct RChangeCompressionFunc {
...
+
+      const auto checksumSize = fWriteOpts.GetEnablePageChecksums() * sizeof(std::uint64_t);
```
hahnjo, line 285: `RResealFunc` enables page checksums iff `fSealedPage.GetHasChecksum()`. Is there a reason to deviate from this logic for pure recompression?

tree/ntuple/src/RNTupleMerger.cxx
```diff
@@ -831,18 +885,41 @@ RNTupleMerger::MergeCommonColumns(ROOT::Internal::RClusterPool &clusterPool,
...
+                  sealedPage,
+                  *fPageAlloc,
+                  buffer.get(),
+                  bufSize,
+                  mergeData.fDestination.GetWriteOptions()
+               });
+            }
+
+            assert(task.has_value());
+            std::visit(RTaskVisitor{fTaskGroup}, *task);
```
hahnjo: Can this be expressed easier with `std::optional<std::variant>`, by just passing temporaries to `RTaskVisitor{fTaskGroup}`. Like

```suggestion
            if (needsResealing) {
               RTaskVisitor{fTaskGroup}(RResealFunc{ /* ... */ });
            } else {
               RTaskVisitor{fTaskGroup}(RChangeCompressionFunc{ /* ... */ });
            }
```

tree/ntuple/src/RNTupleMerger.cxx
silverweed: Shouldn't we rather change `RResealFunc` to also use `fWriteOpts.GetEnablePageChecksums()`? I'd argue it's the "most authoritative" option since it's the one that the user can control

hahnjo: Yes maybe. `fWriteOpts` comes from the user-provided page sink, right? Then the only place where we don't respect that is L1 fast-merging...

tree/ntuple/src/RNTupleMerger.cxx
hahnjo: This is not resolved AFAICT

tree/ntuple/src/RNTupleMerger.cxx
```diff
@@ -276,12 +311,26 @@ struct RChangeCompressionFunc {
       sealConf.fPage = &page;
       sealConf.fBuffer = fBuffer;
       sealConf.fCompressionSettings = *fMergeOptions.fCompressionSettings;
-      sealConf.fWriteChecksum = fSealedPage.GetHasChecksum();
+      sealConf.fWriteChecksum = fWriteOpts.GetEnablePageChecksums();
```
hahnjo, line 314: This should be a separate commit if possible because it changes behavior

tree/ntuple/src/RNTupleMerger.cxx
silverweed: I added the verification at line 277: is it not what you meant?

hahnjo: Sorry, missed that line :grimacing:

PR #20113, opened 2025-10-15
[CI] Don't upload build caches to S3 for Mac15 and Mac26.

hageboeck:
For Mac15 and Mac26, the source and build caches aren't used since 4c3335e, so there is no urgent need to generate them at all. Since the Mac VMs sometimes hang for several hours at the upload step, it seems less fragile to not upload or download anything.

### DIFF
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -123,6 +123,25 @@ jobs:
       ${{ (github.event_name != 'schedule' && github.event_name != 'workflow_dispatch' && join( matrix.overrides, ', ' )) || '' }}
 
     steps:
+      - name: Node state
+        shell: bash -leo pipefail {0}
+        run: |
+          echo $PATH
+          which cmake || true
+          find /opt -name cmake -type f || true
+          find /usr -name cmake -type f || true
+          cmake --version || true
+          which c++ || true
+          c++ --version || true
+          uname -a || true
+          sw_vers || true
+          uptime || true
+          df || true
+          echo $SHELL
+          echo $0
+          python3 -m site || true
+          pkgutil --pkg-info=com.apple.pkg.CLTools_Executables || true
+
       - name: Checkout
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


Ni-Langguth (commented):
How about adding these lines to Node state to find out even more about the environment?
```
echo $SHELL
echo $0
python3 -m site
pkgutil --pkg-info=com.apple.pkg.CLTools_Executables
```

.github/workflows/root-ci.yml
```diff
@@ -123,6 +123,25 @@ jobs:
       ${{ (github.event_name != 'schedule' && github.event_name != 'workflow_dispatch' && join( matrix.overrides, ', ' )) || '' }}
 
     steps:
+      - name: Node state
```
pcanal, line 126: Is this step meant to be merged?

.github/workflows/root-ci.yml
```diff
@@ -203,8 +222,8 @@ jobs:
 
       - name: Update build cache after push to release branch
         shell: bash -leo pipefail {0}
-        if:   github.event_name == 'push'
-        run: ".github/workflows/root-ci-config/build_root.py 
+        if:   ${{ github.event_name == 'push' && !matrix.platform == 'mac15' && !matrix.platform == 'mac26' }}
```
pcanal, line 225: Is there a way to link/relate closer this with the disabling of the use?  (I.e. to avoid someone later one trying to renable the use by forgetting to revert this).

.github/workflows/root-ci.yml
Ni-Langguth: Yes, we would like to add some diagnostics for the node to the CI runs.

.github/workflows/root-ci.yml
hageboeck: We might be able to write some info in the environment variables, like so:
```
<download cached artifacts>
echo ROOT_CI_UPLOAD_ARTIFACTS=1 >> $GITHUB_ENV
```
This environment variable could be used to trigger the upload step. What do you think @pcanal ?

hageboeck:
Superseded by 6504e9653d4c0c8fc2aa4d6b5977e200d36f2cbd

PR #20120, opened 2025-10-16
Prevent TSVG::Text crash when wrong font id specified [6.36]

linev:
Font outside 1..15 range not exists in SVG and does not supported
When used - ROOT crashes

Backported from https://github.com/root-project/root/pull/20111


couet (approved):

linev:
`gtest-tmva-sofie-TestSofieModels` failure has nothing to do with PR

PR #20121, opened 2025-10-16
[main] hadd: properly handle help flag and add tests

silverweed:
# This Pull request:
makes hadd handle `-h`, `-help` or `--help` like the other flags, i.e. accepting it in any valid position rather than just as the first flag.
Note that we accept all versions because a user that requires help on the command shouldn't have to guess the proper form of the flag, we just accept all canonical ones.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/main/src/hadd.cxx
+++ b/main/src/hadd.cxx
@@ -664,7 +664,7 @@ int main(int argc, char **argv)
    const HAddArgs &args = *argsOpt;
 
    if (args.fHelp) {
-      fprintf(stderr, kCommandLineOptionsHelp);
+      fputs(kCommandLineOptionsHelp, stderr);
       return 0;
    }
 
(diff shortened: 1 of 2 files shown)


main/src/hadd.cxx
```diff
@@ -653,16 +658,16 @@ static Int_t ParseFilterFile(const std::optional<std::string> &filterFileName,
 
 int main(int argc, char **argv)
 {
-   if (argc < 3 || "-h" == std::string(argv[1]) || "--help" == std::string(argv[1])) {
```
jblomer, line 656: Do I understand correctly that if we provide no arguments, we won't show a help string anymore? Should we?

silverweed: That's right; I now added a print of the short help msg if input or output files are missing

PR #20123, opened 2025-10-16
[gpad] provide methods to use double as pixel coordinates

linev:
Conversion methods to/from pixel in TVirtualPad now provided with double

Use these methods in TLatex for transformation.  Especially due to rotation double precision let keep original coordiantes


linev:
All failed tests do not use graphics

PR #20126, opened 2025-10-16
[CI] Prepare Mac jobs for running both on physical machines and homebrew-based VMs

hageboeck:
- [x] *This includes commits from #20548, so this one should be merged first.*

Here, the CI of `master` is prepared for running both on physical Macs as well as VMs that are using homebrew and Python virtual environments. To make the workflow steps robust for using one or the other, the build&test steps will check for a folder where a Python virtual environment should be located, and if one is found, they source the `activate` script before starting the build and test.

All workflow steps on Mac are now run in bash login shells, so environments from `.bash_profile` or `.profile` will be visible in each step.

### DIFF
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -101,12 +101,13 @@ jobs:
       run:
         shell: bash -leo pipefail {0} # Use login shell, so profile scripts get sourced
 
+    env:
+      VIRTUAL_ENV_DIR: ${{ github.workspace }}/ROOT_CI_VENV
+
     strategy:
       fail-fast: false
       matrix:
-        # Specify platform + arch + (optional) build option overrides
-        #
-        # Common configs: {Release,Debug,RelWithDebInfo)
+        # Specify platform + optional build option overrides
         # Build options: https://root.cern/install/build_from_source/#all-build-options
         include:
           - platform: mac14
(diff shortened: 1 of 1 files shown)


dpiparo (approved):

.github/workflows/root-ci.yml
```diff
@@ -129,9 +129,9 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
-          ref: ${{ inputs.ref_name }}
```
hahnjo, line 134: This effectively reverts 8bb62ae4f0697323460ee0883b56bcad0047e3f3 which I believe will break nightly builds on future release branches. There the call sequence is something like the following:
1. On the `master` branch, we have a `schedule` workflow. This runs with the ref being the latest commit on `master`.
2. That `uses` the `root-ci.yml` from a fixed branch. However, if I remember correctly, this does not change the context.
3. The `root-ci.yml` from the branch runs, but its `checkout` will download `master`. This gets used for the `buildconfig` and the `build_root.py` script.
4. Eventually, `build_root.py` will download the "right" branch for building.

The critical point is step 3 where we want to make sure that we are actually using the CI configuration of the branch, including platforms and options. I think we need to restore the `ref:` options.

hageboeck: OK, thanks for the explanation!
I anyway needed to apply some changes to test the new runner. 🙂

Issue #19768, opened 2025-08-27
JSON output and infinity.


pcanal:
We have:
```
$ root.exe -b -l
root [0] struct Content { double d = std::numeric_limits<double>::infinity(); }
root [1] Content c;
root [2] auto s = TBufferJSON::ToJSON(&c)
(TString &) "{
  "_typename" : "Content",
  "d" : 2e308
}"[44]
root [3] Content *p = nullptr;
root [4] TBufferJSON::FromJSON<Content>(p, s)
libc++abi: terminating due to uncaught exception of type nlohmann::detail::out_of_range: [json.exception.out_of_range.406] number overflow parsing '2e308'
Abort trap: 6
```
At as side note we also have:
```
root [0] struct Content { double d = std::numeric_limits<double>::infinity(); }
root [1] Content *p = nullptr;
root [2] TString s = "{ \"_typename\" : \"Content\", \"d\" : null }"
(TString &) "{ "_typename" : "Content", "d" : null }"[39]
root [3] TBufferJSON::FromJSON<Content>(p, s)
(bool) true
root [4] p->d
(double) nan
```

So `null` represent nan rather than infinity.

pcanal:
It is not clear what can be done.  Ideas include changing the writing to `"+inf"` or `"-inf"` but that leads:
```
root [0] struct Content { double d = std::numeric_limits<double>::infinity(); }; Content *p = nullptr;
root [2] TString s = "{ \"_typename\" : \"Content\", \"d\" : \"+inf\" }"
(TString &) "{ "_typename" : "Content", "d" : "+inf" }"[41]
root [3] TBufferJSON::FromJSON<Content>(p, s)
libc++abi: terminating due to uncaught exception of type nlohmann::detail::type_error: [json.exception.type_error.302] type must be number, but is string
Abort trap: 6
```
In either case, maybe catching the exception and noting the text of the exception:
```
libc++abi: terminating due to uncaught exception of type nlohmann::detail::out_of_range: [json.exception.out_of_range.406] number overflow parsing '2e308'
```
Another option might be to change the value during the writing to write `std::numeric_limits<double>::max()` instead of `2e308`

ferdymercury:
Could it work to stream the double using std::hexfloat and then reconstruct it back using strtod ?

pcanal:
If I remember correctly the actual printing and reading is done via the 'nholmann' JSON library

ferdymercury:
I see. A (cumbersome) workaround could be:
```
[...]

PR #20128, opened 2025-10-16
[io][json] store inf, nan as strings and read them back properly

ferdymercury:
Fixes https://github.com/root-project/root/issues/19768

### DIFF
--- a/io/io/test/TBufferJSONTests.cxx
+++ b/io/io/test/TBufferJSONTests.cxx
@@ -103,17 +103,17 @@ struct ContentF {
 };
 TEST(TBufferJSON, SpecialNumbersFloat)
 {
-   gInterpreter->ProcessLine("struct ContentF { float v; };");
+   gInterpreter->Declare("struct ContentF { float v; };");
    ContentF *p = nullptr;
-   TBufferJSON::FromJSON<ContentF>(p, "{ \"_typename\" : \"ContentF\", \"v\" : inff }");
+   TBufferJSON::FromJSON<ContentF>(p, "{ \"_typename\" : \"ContentF\", \"v\" : \"inff\" }");
    EXPECT_TRUE(p && std::isinf(p->v) && (p->v > 0));
    delete p;
    p = nullptr;
-   TBufferJSON::FromJSON<ContentF>(p, "{ \"_typename\" : \"ContentF\", \"v\" : -inff }");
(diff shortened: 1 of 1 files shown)


ferdymercury:
I need some help with the Test. It works fine if I run it by copy-pasting it into the interpreter. So it seems like a missing dict that I did not manage to generate well in the scripted version.

io/io/src/TBufferJSON.cxx
```diff
@@ -665,6 +665,8 @@ Bool_t TBufferJSON::IsSkipClassInfo(const TClass *cl) const
 
 TString TBufferJSON::ConvertToJSON(const void *obj, const TClass *cl, Int_t compact, const char *member_name)
 {
+   if (!cl)
+      Error("ConvertToJSON", "%s", "Unknown class (probably missing dictionary).");
```
pcanal:
```suggestion
      ::Error("TBufferJOSN::ConvertToJSON", "%s", "Unknown class (probably missing dictionary).");
```

ferdymercury:
```suggestion
      ::Error("TBufferJSON::ConvertToJSON", "Unknown class (probably missing dictionary).");
```

linev:
Proposed approach is not portable.

So any other JSON parser will fail completely parsing such string - and not only this single member.
JSON normally not used for objects I/O inside ROOT but rather for data exchange with other applications.

So provided solution can be used - but not by default.
One can provide special flag in `compact` argument to enable such feature. 
See `TBufferJSON.h` around lines 35 - 48.
Something like `kStoreInfinity = 500`. 
And then adjust `TBufferJSON::SetCompact` method.

io/io/src/TBufferJSON.cxx
```diff
@@ -481,7 +481,7 @@ class TJSONStackObj : public TObject {
 
 TBufferJSON::TBufferJSON(TBuffer::EMode mode)
    : TBufferText(mode), fOutBuffer(), fOutput(nullptr), fValue(), fStack(), fSemicolon(" : "), fArraySepar(", "),
-     fNumericLocale(), fTypeNameTag("_typename")
+     fNumericLocale(), fStoreInfNaN(), fTypeNameTag("_typename")
```
linev: Either provide default value (kFALSE) or simply do not place it.

ferdymercury:
CI is finally green :)

linev (approved):
This looks fine for me

PR #20130, opened 2025-10-17
[tmva][sofie] install onnxscript from within LinearModelGenerator

sanjibansg:
This PR tries to install onnxscript from within the `LinearModelGenertor.py` file if the package does not already exist to avoid failures in the CI.


vepadulano (changes requested):
Sorry but I believe this is really not good practice

sanjibansg:
> Sorry but I believe this is really not good practice

Yes, this might not be very good, we are also investigating this more, since there are some other cases failing for this test.

guitargeek:
Superseded by #20149.

PR #20142, opened 2025-10-20
[cmake] Disable failing test

bellenot:
`TestSofieModels` is failing on alma8, alma10, ubuntu2204, and ubuntu2404

### DIFF
--- a/tmva/sofie/test/CMakeLists.txt
+++ b/tmva/sofie/test/CMakeLists.txt
@@ -122,8 +122,7 @@ if (ROOT_TORCH_FOUND)
       OUTPUT_VARIABLE module_version
       OUTPUT_STRIP_TRAILING_WHITESPACE
       ERROR_QUIET)
-    message(STATUS "checking torch version - status = ${status} - module_version = ${module_version}")
-    if(${status} EQUAL 0 AND ${module_version} VERSION_GREATER_EQUAL 2.9.0)
+    if(${status} EQUAL 0 AND ${module_version} VERSION_LESS 2.9.0)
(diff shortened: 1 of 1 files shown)


guitargeek:
Can we be maybe more specific here, i.e. mention that the tests fail with the new PyTorch 2.9.0? Also, is should be possible to disable the tests conditional on the Python package version (see our FIND_PYTHON_PACKAGE macro).

We should also open an issue to remind us that  PyTorch 2.9.0 support needs to be taken care of.

bellenot:
> Can we be maybe more specific here, i.e. mention that the tests fail with the new PyTorch 2.9.0? Also, is should be possible to disable the tests conditional on the Python package version (see our FIND_PYTHON_PACKAGE macro).
> 
> We should also open an issue to remind us that PyTorch 2.9.0 support needs to be taken care of.

I didn't know it was specific to PyTorch 2.9.0... And I'll check how to disable it conditionally (just trying to do my shifter job here...)

bellenot:
@guitargeek the PR has been updated and a GH issue created (https://github.com/root-project/root/issues/20145)

bellenot:
@guitargeek it doesn't work...

guitargeek:
It looks like the check for `pytorch` 2.9.0 that you implemented doesn't work yet :slightly_frowning_face: 

Yes, I'm sure the problem is PyTorch 2.9.0 because the failures appeared in the CI once this version got released a few days ago:
https://pypi.org/project/torch/#history

It would just be very nice to still run the test on platforms with older pytorch versions, so we don't completely lose test coverage on Linux for this test

bellenot:
@guitargeek my bad, I used the wrong logic (thanks @linev for letting me know). Should be fine now

PR #20146, opened 2025-10-20
[Python][UHI] Fix handling of infinite upper bin edges in sliced histograms

### LINKED ISSUE #20077: Histogram projections and UHI fail on histograms with axes having infinite upper edges.
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Infinite bin edges are useful for example in event categorizations, where for example events above a final threshold in a variable are placed in a specific category that is not accurately described as an overflow bin. However, when creating projections from histograms that include axes with an infinite upper edge, the contents of the highest bins are not filled in to the projected histogram. This behaviour is undocumented. The below reproducer uses a TH2D to demonstrate the problem, but the same applies at least to the Projection3D method of TH3 histograms.

The new ROOT UHI implementation also fails to operate on dimensions with infinite bin edges so it cannot be used as a workaround in pyROOT. The behaviour here is more clearly incorrect though, as all resulting bins end up empty instead of only the bin that extends to infinity, as is the case when using traditional ROOT methods.

I don't imagine a fix for this should take too much effort, but I am unfortunately too busy with other work to contribute a PR for this in the near future. Of course a large value could be used in place of infinity as a placeholder, but magic numbers are not nice.

### Reproducer

```python
# file repro.py
import numpy as np
import ROOT
import ROOT.uhi as uhi

def inspect(proj, full, y):
    expected = full.values()[:, y-1].sum()
    print(f"proj(y={y}) = {proj.GetBinContent(y)}, expected {expected}")

def test_ops(h):
    proj = h.ProjectionY()
    print("Original histogram")
    print(h.values())
    print("ProjectionY")
    print(proj.values())
    for y in range(1, h.GetNbinsY() + 1):
        inspect(proj, h, y)

    uhi_projection = h[uhi.sum,:]
    print("uhi projection")
    print(uhi_projection.values())

    uhi_slice = h[1,:]
    print("uhi slice")
    print(uhi_slice.values())

nx = 2
ny = 3
h = ROOT.TH2D(
    "h", "h;X;Y;Z",
    nx, np.arange(nx + 1, dtype=float),
    ny, np.arange(ny + 1, dtype=float),
)

for ix in range(1, h.GetNbinsX()+1):
    for iy in range(1, h.GetNbinsY()+1):
            val = 10*ix + iy
            h.SetBinContent(ix, iy, val)
print("Testing standard histogram")
test_ops(h)

h_inf = ROOT.TH2D(
    "h_inf", "h;X;Y;Z",
    nx, np.arange(nx + 1, dtype=float),
    ny, np.concatenate((np.arange(ny , dtype=float), np.array([np.inf]))),
)

for ix in range(1, h_inf.GetNbinsX()+1):
    for iy in range(1, h_inf.GetNbinsY()+1):
            val = 10*ix + iy
            h_inf.SetBinContent(ix, iy, val)
print("Testing histogram with infinite high bin")
test_ops(h_inf)

```
Example output from `python repro.py`:
```
Testing standard histogram
Original histogram
[[11. 12. 13.]
 [21. 22. 23.]]
ProjectionY
[32. 34. 36.]
proj(y=1) = 32.0, expected 32.0
proj(y=2) = 34.0, expected 34.0
proj(y=3) = 36.0, expected 36.0
uhi projection
[32. 34. 36.]
uhi slice
[21. 22. 23.]
Testing histogram with infinite high bin
Original histogram
[[11. 12. 13.]
 [21. 22. 23.]]
ProjectionY
[32. 34.  0.]
proj(y=1) = 32.0, expected 32.0
proj(y=2) = 34.0, expected 34.0
proj(y=3) = 0.0, expected 36.0
uhi projection
[0. 0. 0.]
uhi slice
[0. 0. 0.]
```

### ROOT version

```
ROOT v6.36.04
Built for linuxx8664gcc on Sep 24 2025, 16:10:48
From tags/6-36-04@6-36-04
With c++ (GCC) 15.2.1 20250813
Binary directory: /usr/bin
```

### Installation method

pre-built binaries from Arch Linux repositories and pre-built binaries from conda-forge 

### Operating system

Linux

### Additional context

_No response_

siliataider:
# This Pull request:

## Changes or fixes:
When slicing histograms containing an axis with an infinite upper bin edge, the `SliceHistoInPlace` method produced invalid axis limits (`NaN`) in the resulting slice. This was caused by calling the `SetBins` overload using` (lowEdge, upEdge)` limits, which fails for axes that extend to infinity. Since the resulting sliced histogram had wrong axis values, any following operation including projections had the wrong output.

This PR updates `SliceHistoInPlace` to compute and store the full bin edge arrays for each dimension instead.

The reproducer from issue #20077 now yields identical results for both `h.ProjectionY()` and `h[uhi.sum, :]`.

## Checklist:

- [x] tested changes locally

> [!NOTE]  
> This PR fixes #20077 partially by addressing the second part of the issue
> This implementation makes sure the UHI projection result is the same the original ROOT projection methods, it doesn't however change the behavior of `ProjectionY()` described in the first part of the issue issue

### DIFF
--- a/bindings/pyroot/pythonizations/test/uhi_indexing.py
+++ b/bindings/pyroot/pythonizations/test/uhi_indexing.py
@@ -2,6 +2,7 @@
 Tests to verify that TH1 and derived histograms conform to the UHI Indexing interfaces (setting, accessing and slicing).
 """
 
+import numpy as np
 import pytest
 import ROOT
 from ROOT._pythonization._uhi import _get_axis, _get_processed_slices, _overflow, _shape, _underflow
@@ -324,6 +325,36 @@ class TestTH1Indexing:
         expected = np.full(_shape(hist_setup), 3, dtype=np.int64)
         hist_setup[...] = expected
         assert list(hist_setup) == expected.flatten().tolist()
+
+
+class TestInfiniteEdges:
+    def setup_method(self):
+        # create a 2D histogram with an infinite upper edge on the Y axis
+        xedges = np.array([0.0, 1.0, 2.0], dtype="float64")
+        yedges = np.array([0.0, 1.0, 2.0, np.inf], dtype="float64")
+        self.h_inf = ROOT.TH2D("h_inf", "h_inf", len(xedges) - 1, xedges, len(yedges) - 1, yedges)
+
+        for i in range(1, self.h_inf.GetNbinsX() + 1):
+            for j in range(1, self.h_inf.GetNbinsY() + 1):
+                self.h_inf.SetBinContent(i, j, 10 * i + j)
+
+    def test_uhi_projection_preserves_content(self):
+        """check that UHI projection behaves like standard projection"""
+        # projection on X axis
+        proj_x_ref = self.h_inf.ProjectionX()
+        proj_x_uhi = self.h_inf[:, ROOT.uhi.sum]
+        ref_values = proj_x_ref.values()
+        uhi_values = proj_x_uhi.values()
+
+        assert np.allclose(uhi_values, ref_values)
+
+        # projection on Y axis
+        proj_y_ref = self.h_inf.ProjectionY()
+        proj_y_uhi = self.h_inf[ROOT.uhi.sum, :]
+        ref_values = proj_y_ref.values()
+        uhi_values = proj_y_uhi.values()
+
+        assert np.allclose(uhi_values, ref_values)
 
 
 if __name__ == "__main__":


hist/hist/inc/TH1.h
```diff
@@ -214,17 +214,19 @@ class TH1 : public TNamed, public TAttLine, public TAttFill, public TAttMarker {
...
          nBins[d] = end - start;
-         lowEdge[d] = axis.GetBinLowEdge(start);
-         upEdge[d] = axis.GetBinLowEdge(end);
          totalBins[d] = axis.GetNbins() + 2;
          args[2 * d] = start;
          args[2 * d + 1] = end;
+         // Compute new edges
+         for (int b = start; b <= end; ++b)
+            edges[d].push_back(axis.GetBinLowEdge(b));
+         edges[d].push_back(axis.GetBinUpEdge(end));
```
vepadulano, line 229: So practically the change is that now we are supporting variable bin edges, correct? So it's doing slightly more than just adding support for the case of the infinite value in the bin edge. I'm just trying to understand if I got it right

siliataider: Yes, this now doesn’t recompute or redefine the bin edges, it simply slices away the unused part of the histogram and retrieves the original remaining bin edges making the behavior more correct and naturally extending support for variable-width bins

siliataider:
The failing test (`483:gtest-tmva-sofie-TestSofieModels`) on all platforms is unrelated to this PR

hist/hist/inc/TH1.h
vepadulano: Ok nice!

vepadulano (approved):
Thank you, this is definitely an improvement! IIUC, TH1 and derivates still have an issue with handling infinite bin edges. If that is the case, that should be reported in the linked issue, ideally with a C++ reproducer, or a new one should be opened to not lose sight of that.

Issue #20145, opened 2025-10-20
`TestSofieModels` is failing with PyTorch >= 2.9.0


bellenot:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

`gtest-tmva-sofie-TestSofieModels` is failing on alma8, alma10, ubuntu22, and ubuntu2404, as shown below:
```
   864/2687 Test  #439: gtest-tmva-sofie-TestSofieModels ..................................................................***Failed   30.62 sec
  Running main() from ./googletest/src/gtest_main.cc
  [==========] Running 14 tests from 1 test suite.
  [----------] Global test environment set-up.
  [----------] 14 tests from SOFIE
  [ RUN      ] SOFIE.Linear_B1
  using batch-size = 1 input dim = 10 nlayers = 4
  input data torch.Size([1, 10])
  tensor([[1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]])
  W1020 01:57:39.124000 78746 torch/onnx/_internal/exporter/_registration.py:107] torchvision is not installed. Skipping torchvision::nms
  [torch.onnx] Obtain model graph for `Net([...]` with `torch.export.export(..., strict=False)`...
  [torch.onnx] Obtain model graph for `Net([...]` with `torch.export.export(..., strict=False)`... ✅
  [torch.onnx] Run decomposition...
  [torch.onnx] Run decomposition... ✅
  [torch.onnx] Translate the graph into ONNX...
  [torch.onnx] Translate the graph into ONNX... ✅
  output data : shape,  torch.Size([1, 4])
  tensor([[ 0.0202, -0.0144,  0.0183, -0.1695]], grad_fn=<AddmmBackward0>)
  /github/home/ROOT-CI/build/tmva/sofie/test/LinearModelGenerator.py:148: UserWarning: Converting a tensor with requires_grad=True to a scalar may lead to unexpected behavior.
  Consider using tensor.detach() first. (Triggered internally at /pytorch/torch/csrc/autograd/generated/python_variable_methods.cpp:836.)
    f.write(str(float(yvec[i]))+" ")
  executing python3 LinearModelGenerator.py   1  10  4
  parsing file LinearModel_B1.onnx
  generating model.....
  writing model as header .....
  output written in  LinearModel_B1.hxx
  doing inference.....(std::vector<float>) { inff, inff, inff, inff }
   result inf reference 0.0201594
[...]

PR #20149, opened 2025-10-20
[tmva][sofie]  Fix testSofieModels for new torch version

lmoneta:
New torch versions comes with a new export mode to ONNX. One needs to specify first to have a single onnx file with the weights (external_data=False) and then in some case (new export mode, which is with dynamo=True) does not work for batchNorm and recurrent network. Disable it for these cases.

This PR should fix the test TestSofieModels

### DIFF
--- a/tmva/sofie/test/CMakeLists.txt
+++ b/tmva/sofie/test/CMakeLists.txt
@@ -116,24 +116,14 @@ if (ROOT_TORCH_FOUND)
   configure_file(RecurrentModelGenerator.py  RecurrentModelGenerator.py COPYONLY)
 
   if (BLAS_FOUND)
-    execute_process(COMMAND "${Python3_EXECUTABLE}" "-c"
-                            "import torch; print(getattr(torch, '__version__', 'unknown'))"
-      RESULT_VARIABLE status
-      OUTPUT_VARIABLE module_version
-      OUTPUT_STRIP_TRAILING_WHITESPACE
-      ERROR_QUIET)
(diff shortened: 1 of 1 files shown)


tmva/sofie/test/TestSofieModels.cxx
```diff
@@ -11,7 +11,7 @@
 #include "TMVA/RModelParser_ONNX.hxx"
 
 #include  "gtest/gtest.h"
-#define USE_ONNXSIM
+//#define USE_ONNXSIM
```
sanjibansg, line 14: maybe we can remove this line since it is no longer required now
```suggestion

```

guitargeek:
Hi! Could you add the revert of commit 5f3a91fb168f29 here? Thanks!

tmva/sofie/test/TestSofieModels.cxx
lmoneta: I think is better to keep it, to remeber to enable it if needed

ferdymercury:
Fixes https://github.com/root-project/root/issues/20145

vepadulano:
Are we sure this is fixing all platforms? I can see both Mac runners failing in this PR https://github.com/root-project/root/actions/runs/18656813948/job/53187961355 https://github.com/root-project/root/actions/runs/18656813948/job/53187961360

hahnjo:
FWIW the failure (that was already visible in this PR!) is
```
TypeError: export() got an unexpected keyword argument 'dynamo'
```
So probably this PR *requires* the new PyTorch version. I'm not sure that this is what we want...

PR #20153, opened 2025-10-21
[tmva][sofie] Add a check on torch version in TestSofieModels

lmoneta:
Check that the input torch version support the new dynamo torch exporter to ONNX

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,7 @@ dm-sonnet # used for GNNs
 graph_nets
 onnx
 onnxscript
+packaging
 
 # TMVA: PyMVA interfaces
 scikit-learn


sanjibansg (approved):
LGTM! thanks!

tmva/sofie/test/Conv1dModelGenerator.py
```diff
@@ -132,14 +132,20 @@ def main():
...
-         model,
-         xinput,
-         name + ".onnx",
-         export_params=True,
-         dynamo=True,
-         external_data=False
-      )
+      #check torch version
+      v = torch.__version__
+      from packaging.version import Version
```
vepadulano, line 137: `packaging` is not a standard Python module, https://pypi.org/project/packaging/ . If we need to use it in these tests, it must be added to `requirements.txt` before merging this PR. Even better, you could think about not using it altogether

lmoneta: I think it worked because it is probably already install as part of one of the other dependencies. Since it is needed to managed version numbers of the different packages, API in Python changes frequently without backward compatibility, I would add it to `requirements.txt`

PR #20159, opened 2025-10-21
[ntuple] Check dynamic type of polymorphic objects

hahnjo:
For a field of type `std::unique_ptr<T>`, where `T` is a polymorphic class (that declares or inherits at least one virtual method), the user may put a pointer to an object of a derived class. This class would be sliced, contrary to user expectation from classical ROOT IO, so throw an exception in this case.

The implementation uses a strategy similar to `TIsAProxy`: If the item type is a polymorphic class, the `RUniquePtrField` constructor determines the expected dynamic type using `TClass::GetTypeInfo()`. Then for every append, the object pointer is cast to a dummy type that allows to use the `typeid` operator to get the actual dynamic type.

Note: I did not benchmark the run-time cost; in all current implementations, it should be two pointer loads (vtable ptr from object, `std::type_info` ptr from vtable) and the comparison operator of `std::type_info`.

Note 2: In principle, this could break users that misused fields of polymorphic class type before. It may have even worked if the derived class didn't specify additional members, but I'm not sure we can or want to support this special case (for the moment).

### DIFF
--- a/tree/ntuple/src/RField.cxx
+++ b/tree/ntuple/src/RField.cxx
@@ -947,6 +947,8 @@ std::size_t ROOT::RUniquePtrField::AppendImpl(const void *from)
    if (*typedValue) {
       const void *obj = typedValue->get();
       if (fPolymorphicTypeInfo != nullptr) {
+         // This cast allows getting the dynamic type of polymorphic objects. A similar strategy is employed by
+         // TIsAProxy. If one of them needs updating because of changes in C++, also check the other one.
          const std::type_info &t = typeid(*static_cast<const PolymorphicClass *>(obj));
          if (t != *fPolymorphicTypeInfo) {
             std::string msg = "invalid dynamic type of object, expected " + fSubfields[0]->GetTypeName();
--- a/core/meta/src/TIsAProxy.cxx
+++ b/core/meta/src/TIsAProxy.cxx
@@ -103,6 +103,8 @@ TClass* TIsAProxy::operator()(const void *obj)
       return fClass;
    }
 
+   // This cast allows getting the dynamic type of polymorphic objects. A similar strategy is employed in RNTuple's
+   // RUniquePtrField::AppendImpl. If one of them needs updating because of changes in C++, also check the other one.
    DynamicType* ptr = (DynamicType*)obj;
    const std::type_info* typ = &typeid(*ptr);
 


tree/ntuple/src/RField.cxx
```diff
@@ -918,9 +918,19 @@ void ROOT::RNullableField::AcceptVisitor(ROOT::Detail::RFieldVisitor &visitor) c
 
 //------------------------------------------------------------------------------
 
+namespace {
+// Dummy class to determine the dynamic type of any polymorphic user object.
+struct PolymorphicClass {
+   virtual ~PolymorphicClass() {}
```
silverweed: Can this be `= default`?

tree/ntuple/src/RField.cxx
```diff
@@ -933,7 +943,14 @@ std::size_t ROOT::RUniquePtrField::AppendImpl(const void *from)
 {
    auto typedValue = static_cast<const std::unique_ptr<char> *>(from);
    if (*typedValue) {
-      return AppendValue(typedValue->get());
+      const void *obj = typedValue->get();
+      if (fPolymorphicTypeInfo != nullptr) {
+         const std::type_info &t = typeid(static_cast<const PolymorphicClass *>(obj));
+         if (t != *fPolymorphicTypeInfo) {
+            throw RException(R__FAIL("invalid dynamic type of object, expected " + fSubfields[0]->GetTypeName()));
```
silverweed: Maybe we should we explicitly tell in the error that this kind of upcasting is not supported, as the user might expect it to work

tree/ntuple/src/RField.cxx
```diff
@@ -933,7 +943,14 @@ std::size_t ROOT::RUniquePtrField::AppendImpl(const void *from)
 {
    auto typedValue = static_cast<const std::unique_ptr<char> *>(from);
    if (*typedValue) {
-      return AppendValue(typedValue->get());
+      const void *obj = typedValue->get();
+      if (fPolymorphicTypeInfo != nullptr) {
+         const std::type_info &t = typeid(static_cast<const PolymorphicClass *>(obj));
```
pcanal: (minor)  Consider adding a comment here (and possibly in `TIsAproxy`) referring to this technique being used in the other place, in case there is a change in C++ that lead us to need to update it. (unless it is not undefined behavior).

tree/ntuple/src/RField.cxx
pcanal:
```suggestion
            throw RException(R__FAIL("invalid dynamic type of object, expected " + fSubfields[0]->GetTypeName() + " but was passed " + t.name()));
```

tree/ntuple/src/RField.cxx
```diff
@@ -933,7 +943,14 @@ std::size_t ROOT::RUniquePtrField::AppendImpl(const void *from)
 {
    auto typedValue = static_cast<const std::unique_ptr<char> *>(from);
    if (*typedValue) {
-      return AppendValue(typedValue->get());
+      const void *obj = typedValue->get();
```
pcanal, line 948: Especially since/if we expect `*typedValue` to most often be not nullptr. Wouldn't the following be (vaguely) faster:
```
const void *obj = typedValue->get();
if (obj) {
```

pcanal (approved):
LGTM.

> Note 2: In principle, this could break users that misused fields of polymorphic class type before. It may have even worked if the derived class didn't specify additional members, but I'm not sure we can or want to support this special case (for the moment).

How hard it is be to add a way for the user to add a flag/something to tell `RNTuple` that they do indeed indeed to pass derived classes and that they really want to be sliced?  (i.e. suppress the error).

jblomer (approved):

tree/ntuple/src/RField.cxx
hahnjo: Done. Though note that the class is not actually used, we just need one `virtual` method...

tree/ntuple/src/RField.cxx
hahnjo: `if (*typedValue)` calls the `bool` conversion operator, which does exactly this check, so in practice I would expect the compiler to optimize this away. (It's also somewhat unrelated to this PR, so resolving.)

tree/ntuple/src/RField.cxx
hahnjo: Done. I threw in an additional `TClassEdit::DemangleTypeIdName` so that the error message in the test is
```
invalid dynamic type of object, expected PolymorphicBase but was passed PolymorphicDerived
```
instead of
```
invalid dynamic type of object, expected PolymorphicBase but was passed 18PolymorphicDerived
```
(and worse for classes in `namespace`s)

And this actually told me that I was missing a pointer dereference in the `typeid` argument, so this PR would have broken storing any polymorphic object in a `std::unique_ptr` :grimacing: now fixed and more tests added!

tree/ntuple/src/RField.cxx
hahnjo: Done. (I'm pretty sure it's UB)

hahnjo:
> > Note 2: In principle, this could break users that misused fields of polymorphic class type before. It may have even worked if the derived class didn't specify additional members, but I'm not sure we can or want to support this special case (for the moment).
> 
> How hard it is be to add a way for the user to add a flag/something to tell `RNTuple` that they do indeed indeed to pass derived classes and that they really want to be sliced? (i.e. suppress the error).

I think not too difficult: Probably we would want this in the `RNTupleWriteOptions`, which we can get from the page sink. We currently don't have a `virtual` hook similar to `BeforeConnectPageSource`, but that should not be too hard to invent. Then the implementation in `RUniquePtrField` would only need to reset `fPolymorphicTypeInfo = nullptr` which effectively disables the check (and any overhead).

silverweed (approved):

PR #20161, opened 2025-10-22
[gui] Avoid several crashes when the images are not found

silverweed:
Opening the classic browser when some pictures are missing causes crashes due to missing checks. With this PR the browser will simply not display the missing elements (and it still works somewhat fine) in such cases.

### DIFF
--- a/gui/gui/src/TGTextEditor.cxx
+++ b/gui/gui/src/TGTextEditor.cxx
@@ -240,7 +240,7 @@ TGTextEditor::TGTextEditor(const char *filename, const TGWindow *p, UInt_t w,
       fToolBar->RemoveFrame(fComboCmd);
       fLabel->UnmapWindow();
       fToolBar->RemoveFrame(fLabel);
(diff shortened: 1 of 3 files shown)


gui/gui/src/TGTextEditor.cxx
```diff
@@ -866,7 +866,7 @@ Bool_t TGTextEditor::HandleTimer(TTimer *t)
    }
    else {
       fMenuEdit->EnableEntry(kM_EDIT_PASTE);
-      if (fToolBar->GetButton(kM_EDIT_PASTE)->GetState() == kButtonDisabled)
+      if (auto btn = fToolBar->GetButton(kM_EDIT_PASTE); btn && btn->GetState() == kButtonDisabled)
          fToolBar->GetButton(kM_EDIT_PASTE)->SetState(kButtonUp);
```
linev: One can use here `btn->SetState(kButtonUp);`

linev:
For me changes looks ok.
May be @bellenot will approve them?

PR #20162, opened 2025-10-22
[hist] Implement initial `RCategoricalAxis`

hahnjo:
... and integrate into `RAxes`.

### DIFF
--- a/hist/histv7/test/hist_axes.cxx
+++ b/hist/histv7/test/hist_axes.cxx
@@ -17,19 +17,23 @@ TEST(RAxes, Constructor)
    }
    bins.push_back(BinsY);
    const RVariableBinAxis variableBinAxis(bins);
+   const std::vector<std::string> categories = {"a", "b", "c"};
+   const RCategoricalAxis categoricalAxis(categories);
 
-   RAxes axes({regularAxis, variableBinAxis});
-   EXPECT_EQ(axes.GetNDimensions(), 2);
+   RAxes axes({regularAxis, variableBinAxis, categoricalAxis});
+   EXPECT_EQ(axes.GetNDimensions(), 3);
    const auto &v = axes.Get();
-   ASSERT_EQ(v.size(), 2);
+   ASSERT_EQ(v.size(), 3);
    EXPECT_EQ(v[0].index(), 0);
    EXPECT_EQ(v[1].index(), 1);
+   EXPECT_EQ(v[2].index(), 2);
    EXPECT_TRUE(std::get_if<RRegularAxis>(&v[0]) != nullptr);
    EXPECT_TRUE(std::get_if<RVariableBinAxis>(&v[1]) != nullptr);
+   EXPECT_TRUE(std::get_if<RCategoricalAxis>(&v[2]) != nullptr);
 
-   std::vector<RAxisVariant> newAxes{variableBinAxis, regularAxis};
+   std::vector<RAxisVariant> newAxes{categoricalAxis, variableBinAxis, regularAxis};
    axes = RAxes(newAxes);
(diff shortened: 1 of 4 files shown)


hist/histv7/inc/ROOT/RCategoricalAxis.hxx
```diff
@@ -0,0 +1,177 @@
...
+            std::string msg = "duplicate category '" + fCategories[i] + "' for bin " + std::to_string(i);
+            throw std::invalid_argument(msg);
+         }
+      }
+   }
+
+   std::size_t GetNNormalBins() const { return fCategories.size(); }
+   std::size_t GetTotalNBins() const { return fEnableOverflowBin ? fCategories.size() + 1 : fCategories.size(); }
+   const std::vector<std::string> &GetCategories() const { return fCategories; }
+   std::string GetCategory(std::size_t bin) const { return fCategories[bin]; }
```
hageboeck: Is this called by the implementation (i.e. important for performance)? If not, one could consider `fCategories.at(bin)` to not leave user input unchecked.

hist/histv7/inc/ROOT/RCategoricalAxis.hxx
```diff
@@ -0,0 +1,177 @@
...
+      }
+      if (!end.IsNormal()) {
+         throw std::invalid_argument("end must be a normal bin");
+      }
+      if (end.GetIndex() > fCategories.size()) {
+         throw std::invalid_argument("end must be inside or past the axis");
+      }
+      if (!(end >= begin)) {
+         throw std::invalid_argument("end must be >= begin");
+      }
```
hageboeck, line 154: In the STL, `begin == end` would create an empty range. Would it make sense to allow this?

hist/histv7/test/hist_axes.cxx
```diff
@@ -77,10 +89,12 @@ TEST(RAxes, ComputeTotalNBins)
    }
    bins.push_back(BinsY);
    const RVariableBinAxis variableBinAxis(bins);
-   const RAxes axes({regularAxis, variableBinAxis});
+   const std::vector<std::string> categories = {"a", "b", "c"};
+   const RCategoricalAxis categoricalAxis(categories);
+   const RAxes axes({regularAxis, variableBinAxis, categoricalAxis});
 
    // Both axes include underflow and overflow bins.
```
hageboeck: The comment is slightly outdated now, but maybe not worth changing it ...

hageboeck (approved):

hist/histv7/inc/ROOT/RCategoricalAxis.hxx
hahnjo: In fact, we don't need the method at all, the user can for now call `GetCategories()` and then access the `vector` directly. I also removed `RVariableBinAxis::GetBinEdge()` when copying from EPHist...

Edit: ... and even worse, the method was returning a copy of the `std::string` :scream:

hist/histv7/inc/ROOT/RCategoricalAxis.hxx
hahnjo: Yes, this is allowed and used (+ tested) for empty ranges. Maybe the condition is a bit confusing? It's equivalent to `if (end < begin)`...

hist/histv7/inc/ROOT/RCategoricalAxis.hxx
hageboeck: That's a good solution 🙂

hist/histv7/inc/ROOT/RCategoricalAxis.hxx
hageboeck: Yes, indeed! 😅

Issue #20138, opened 2025-10-19
Improper behavior of TCanvas::Divide


rlalik:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Documentation for TCanvas::Divide says:

> Automatic pad generation by division.
> 
> * The current canvas is divided in nx by ny equal divisions (pads).
> * xmargin is the space along x between pads in percent of canvas.
> * ymargin is the space along y between pads in percent of canvas.
> * color is the color of the new pads. If 0, color is the canvas color.

But setting margins to 0 results in:
1. Not setting up colors of subpards
2. Adding top-right padding to the main canvas.

The documentation also says that:

> Note3: in case xmargin <=0 and ymargin <= 0, there is no space between pads. The current pad margins are recomputed to optimize the layout.

Not sure what the second sentence means, what pad margins are recomputed?

### Reproducer

Here a couple of examples of proper and improper behaviors:
1. This works as expected - OK
```c++
{
    TCanvas * can = new TCanvas("can", "can", 800, 800);
    can->Divide(2, 2, 0.01, 0.01, 2);
}
```
<img width="800" height="800" alt="Image" src="https://github.com/user-attachments/assets/61da2300-678b-4e93-98dc-7727a5968902" />

2. This gives white pads whereas they should be red - NOT OK
```c++
{
    TCanvas * can = new TCanvas("can", "can", 800, 800);
    can->Divide(2, 2, 0.0, 0.0, 2);
}
```
<img width="800" height="800" alt="Image" src="https://github.com/user-attachments/assets/49a4e793-18db-423c-aaf0-942152728938" />

3. After putting colours manually, we see odd top-right margin - NOT OK
```c++
{
    TCanvas * can = new TCanvas("can", "can", 800, 800);
    
    can->Divide(2, 2, 0.0, 0.0, 2);
    
    can->cd(1);
    gPad->SetFillColor(30);
    gPad->Modified();
    gPad->Update();
    
    can->cd(2);
    gPad->SetFillColor(38);
    gPad->Modified();
    gPad->Update();
    
    can->cd(3);
    gPad->SetFillColor(41);
    gPad->Modified();
    gPad->Update();
    
    can->cd(4);
    gPad->SetFillColor(42);
    gPad->Modified();
    gPad->Update();
    
    can->cd(0);
    gPad->SetFillColor(2);
    gPad->Modified();
    gPad->Update();
}
```
[...]

PR #20165, opened 2025-10-22
[graf2d] allow zero-margins between pads and respect color for negative margins

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/20138

fyi @rlalik

a

## Checklist:

- [x] tested changes locally

### DIFF
--- a/graf2d/gpad/src/TPad.cxx
+++ b/graf2d/gpad/src/TPad.cxx
@@ -1251,7 +1251,7 @@ Int_t TPad::DistancetoPrimitive(Int_t px, Int_t py)
 ///             points to the current pad. One can use gPad to set attributes
 ///             of the current pad.
 ///
-/// __Note3:__  in case xmargin < 0 and ymargin < 0, there is no space
+/// __Note3:__  in case xmargin < 0 or ymargin < 0, there is no space
 ///             between pads. The current pad margins are recomputed to
 ///             optimize the layout.
 


rlalik:
I didn't test the PR code but this seems to be the solution. However, since this was broken and is being fixed, I would like to propose some change. Current behavior is that, when setting non-zero positive margin, the margin is applied to all pads, also on the outer edges, not only between pads.  But sometimes one wants to increase the margin to separate the pads, for better visibility or so, but actually we don't want to increase the margin on the outer edges, where there is no other pad to keep distance from.

So my proposition would be, when the margin is positive, have all margins set - this will be the default behavior, compatible with 99% existing code. But when the margin value is negative, only the inner margin between pads are set, and the outer margin are kept 0. Setting margin to be 0 removes the margins between pads.

A such implementation would be flexible I think for almost all cases.

So with `canvas->Divide(2, 2, -0.02, -0.02)` I would get something like this:
can_no_outer_margins
and with `canvas->Divide(2, 2, 0.02, -0.02)` I would get something like this:
can_no_outer_margins
Could it be done?

ferdymercury:
@rlalik would you like to propose a code modification for the new feature you are mentioning? It should go just below the if statement I modified

graf2d/gpad/src/TPad.cxx
```diff
@@ -1298,7 +1298,7 @@ void TPad::Divide(Int_t nx, Int_t ny, Float_t xmargin, Float_t ymargin, Int_t co
          }
       }
    } else {
-      // special case when xmargin <= 0 && ymargin <= 0
+      // special case when xmargin < 0 && ymargin < 0
```
ferdymercury:
```suggestion
      // special case when xmargin < 0 or ymargin < 0
```

graf2d/gpad/src/TPad.cxx
```diff
@@ -1251,7 +1251,7 @@ Int_t TPad::DistancetoPrimitive(Int_t px, Int_t py)
 ///             points to the current pad. One can use gPad to set attributes
 ///             of the current pad.
 ///
-/// __Note3:__  in case xmargin <=0 and ymargin <= 0, there is no space
+/// __Note3:__  in case xmargin < 0 and ymargin < 0, there is no space
```
ferdymercury:
```suggestion
/// __Note3:__  in case xmargin < 0 or ymargin < 0, there is no space
```

rlalik:
> @rlalik would you like to propose a code modification for the new feature you are mentioning? It should go just below the if statement I modified

So I propose that this gets merged and then I submit PR with my proposed change to add my requested feature.

linev:
I reopen PR to restart CI

ferdymercury:
CI green now :)

rlalik:
Ok, I made tests with following macro:
```cpp
{
    TCanvas * can_0 = new TCanvas("can_0", "can_0", 200, 200, 200, 200);
    can_0->Divide(2, 2, 0.1, 0.1, 2);
    can_0->Draw();

    TCanvas * can_1 = new TCanvas("can_1", "can_1", 408, 200, 200, 200);
    can_1->Divide(2, 2, 0.0, 0.1, 2);
    can_1->Draw();

    TCanvas * can_2 = new TCanvas("can_2", "can_2", 616, 200, 200, 200);
    can_2->Divide(2, 2, 0.1, 0.0, 2);
    can_2->Draw();

    TCanvas * can_3 = new TCanvas("can_3", "can_3", 824, 200, 200, 200);
    can_3->Divide(2, 2, 0.0, 0.0, 2);
    can_3->Draw();

    TCanvas * can_4 = new TCanvas("can_4", "can_4", 200, 432, 200, 200);
    can_4->Divide(2, 2, -0.1, -0.1, 2);
    can_4->Draw();

    TCanvas * can_5 = new TCanvas("can_5", "can_5", 408, 432, 200, 200);
    can_5->Divide(2, 2, 0.1, -0.1, 2);
    can_5->Draw();

    TCanvas * can_6 = new TCanvas("can_6", "can_6", 616, 432, 200, 200);
    can_6->Divide(2, 2, -0.1, 0.1, 2);
    can_6->Draw();

    TCanvas * can_7 = new TCanvas("can_7", "can_7", 824, 432, 200, 200);
    can_7->Divide(2, 2, 0.1, 0.1, 2);
    can_7->Draw();
}
```
image

Canvases 4...6 still seems to behave odd with negative margin value. I didn't yet implement my feature, this is the version after this PR merge.

I will try to implement my negative margins.

PR #20167, opened 2025-10-22
[rfile] Add pythonizations

silverweed:
# This Pull request:
- introduces `RFile::GetKeyInfo` (necessary for the non-templated `Get` pythonization)
- pythonizes the factory methods, `Get`, `Put`, `GetKeyInfo` and `ListKeys`.

Get/Put: allow them to be called without template arguments, check the type at runtime
GetKeyInfo: return None instead of std::nullopt
ListKeys: use keyword arguments instead of bitmask for flags

## Checklist:

- [x] tested changes locally

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rfile.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rfile.py
@@ -1,0 +1,148 @@
+# Author: Giacomo Parolini CERN 04/2025
+
+r"""
+\pythondoc RFile
+
+TODO: document RFile
+
+\code{.py}
+# TODO code example
+\endcode
+
+\endpythondoc
+"""
+
+from . import pythonization
+
+
+class _RFile_Get:
+    """
+    Allow access to objects through the method Get().
+    This is pythonized to allow Get() to be called both with and without a template argument.
+    """
+
+    def __init__(self, rfile):
+        self._rfile = rfile
+
+    def __call__(self, namecycle):
+        """
+        Non-templated Get()
+        """
+        import ROOT
+        import cppyy
+
+        key = self._rfile.GetKeyInfo(namecycle)
+        if key:
+            obj = ROOT.Experimental.Internal.RFile_GetObjectFromKey(self._rfile, key)
+            return cppyy.bind_object(obj, key.GetClassName())
+        # No key
+        return None
+
+    def __getitem__(self, template_arg):
+        """
+        Templated Get()
+        """
+
+        def getitem_wrapper(namecycle):
+            obj = self._rfile._OriginalGet[template_arg](namecycle)
+            return obj if obj else None
+
+        return getitem_wrapper
+
+
+class _RFile_Put:
+    """
+    Allow writing objects through the method Put().
+    This is pythonized to allow Put() to be called both with and without a template argument.
+    """
+
+    def __init__(self, rfile):
+        self._rfile = rfile
+
+    def __call__(self, name, obj):
+        """
+        Non-templated Put()
+        """
+        objType = type(obj)
+        if objType == str:
+            # special case: automatically convert python str to std::string
+            className = "std::string"
+        elif not hasattr(objType, '__cpp_name__'):
+            raise TypeError(f"type {objType} is not supported by ROOT I/O") 
+        else:
+            className = objType.__cpp_name__
+        self._rfile.Put[className](name, obj)
+
+    def __getitem__(self, template_arg):
+        """
+        Templated Put()
+        """
+        return self._rfile._OriginalPut[template_arg]
+
+
+def _RFileExit(obj, exc_type, exc_val, exc_tb):
+    """
+    Close the RFile object.
+    Signature and return value are imposed by Python, see
+    https://docs.python.org/3/library/stdtypes.html#typecontextmanager.
+    """
+    obj.Close()
+    return False
+
+
+def _RFileOpen(original):
+    """
+    Pythonization for the factory methods (Recreate, Open, Update)
+    """
+
+    def rfile_open_wrapper(klass, *args):
+        rfile = original(*args)
+        rfile._OriginalGet = rfile.Get
+        rfile.Get = _RFile_Get(rfile)
+        rfile._OriginalPut = rfile.Put
+        rfile.Put = _RFile_Put(rfile)
+        return rfile
+
+    return rfile_open_wrapper
+
+
+def _RFileInit(rfile):
+    """
+    Prevent the creation of RFile through constructor (must use a factory method)
+    """
+    raise NotImplementedError("RFile can only be created via Recreate, Open or Update")
+
+
+def _GetKeyInfo(rfile, path):
+    key = rfile._OriginalGetKeyInfo(path)
+    if key.has_value():
+        return key.value()
+    return None
+
+
+def _ListKeys(rfile, basePath="", listObjects = True, listDirs = False, listRecursive = True):
+    from ROOT.Experimental import RFile
+
+    flags = (listObjects * RFile.kListObjects) | (listDirs * RFile.kListDirs) | (listRecursive * RFile.kListRecursive)
+    iter = rfile._OriginalListKeys(basePath, flags)
+    return iter
+
+
+@pythonization("RFile", ns="ROOT::Experimental")
+def pythonize_rfile(klass):
+    # Explicitly prevent to create a RFile via ctor
+    klass.__init__ = _RFileInit
+
+    # Pythonize factory methods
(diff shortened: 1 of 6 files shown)


bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rfile.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+
+    def __init__(self, rfile):
+        self._rfile = rfile
+
+    def __call__(self, name, obj):
+        """
+        Non-templated Put()
+        """
+        className = type(obj).__cpp_name__
```
vepadulano: Maybe add here a touch of sanitization incase the passed object does not have `__cpp_name__`

vepadulano: A case that I guess would deserve its own unittest

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rfile.py
```diff
@@ -0,0 +1,156 @@
...
+    Pythonization for the factory methods (Recreate, Open, Update)
+    """
+
+    def rfile_open_wrapper(klass, *args):
+        rfile = original(*args)
+        rfile._OriginalGet = rfile.Get
+        rfile.Get = _RFile_Get(rfile)
+        rfile._OriginalPut = rfile.Put
+        rfile.Put = _RFile_Put(rfile)
+        return rfile
```
vepadulano, line 104: It's slightly weird to see these lines here as they do not relate to the `Open` method per se. Usually these are placed in the function that is wrapped by the `@pythonization` decorator

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rfile.py
```diff
@@ -0,0 +1,156 @@
...
+
+    return rfile_open_wrapper
+
+
+def _RFileInit(rfile):
+    """
+    Prevent the creation of RFile through constructor (must use a factory method)
+    """
+    raise NotImplementedError("RFile can only be created via Recreate, Open or Update")
```
vepadulano, line 113: Nice!

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rfile.py
```diff
@@ -0,0 +1,156 @@
...
+    if key.has_value():
+        return key.value()
+    return None
+
+
+# kwargs are:
+#   listObjects (default: True)
+#   listDirs (default: False)
+#   listRecursive (default: True)
+def _ListKeys(rfile, basePath="", **kwargs):
```
vepadulano: What is preventing us from defining the signature as
```suggestion
def _ListKeys(rfile, basePath="", listObjects=True, listDirs=False, listRecursive=True):
```

io/io/src/RFile.cxx
```diff
@@ -269,21 +280,27 @@ TKey *RFile::GetTKey(std::string_view path) const
...
 
-   const TClass *cls = TClass::GetClass(type);
    if (!cls)
-      throw ROOT::RException(R__FAIL(std::string("Could not determine type of object ") + path));
+      throw ROOT::RException(R__FAIL(std::string("Could not determine type of object ") + pathStr));
 
-   if (auto err = ValidateAndNormalizePath(path); !err.empty())
-      throw RException(R__FAIL("Invalid object path '" + path + "': " + err));
+   if (auto err = ValidateAndNormalizePath(pathStr); !err.empty())
+      throw RException(R__FAIL("Invalid object pathStr '" + pathStr + "': " + err));
```
vepadulano, line 301: Maybe clearer
```suggestion
      throw RException(R__FAIL("Invalid object path '" + pathStr + "': " + err));
```

io/io/test/rfile.py
```diff
@@ -0,0 +1,115 @@
...
+
+        with RFile.Open(fileName) as rfile:
+            hist = rfile.Get("hist")
+            self.assertNotEqual(hist, None)
+            self.assertEqual(rfile.Get[ROOT.TH1D]("inexistent"), None)
+            self.assertEqual(rfile.Get[ROOT.TH1F]("hist"), None)
+            self.assertNotEqual(rfile.Get[ROOT.TH1]("hist"), None)
+
+            with self.assertRaises(ROOT.RException) as _cm:
+                rfile.Put("foo", hist)
```
vepadulano, line 29: Sorry I must have lost something, why does this raise an exception?

silverweed: Because the file is opened as read-only (I'll add a comment to explain it)

vepadulano: That makes absolute sense, I was just looking at the function call itself, apologies :)

io/io/src/RFile.cxx
```diff
@@ -513,3 +530,24 @@ void RFile::Close()
...
+   RKeyInfo keyInfo;
+   keyInfo.fPath = ReconstructFullKeyPath(*key);
+   keyInfo.fClassName = key->GetClassName();
+   keyInfo.fCycle = key->GetCycle();
+   keyInfo.fTitle = key->GetTitle();
+
+   return keyInfo;
+}
+
+void *ROOT::Experimental::Internal::GetRFileObjectFromKey(RFile &file, const RKeyInfo &key)
```
pcanal:
```suggestion
void *ROOT::Experimental::Internal::RFileGetObjectFromKey(RFile &file, const RKeyInfo &key)
```
or just (since `RFile` is already implied from the signature)
```suggestion
void *ROOT::Experimental::Internal::GetObjectFromKey(RFile &file, const RKeyInfo &key)
```
The original reads (to me) as 'Get' 'the RFile Object' (i.e. an object of type `RFile`) `from` `the Key` (as opposed to (I suppose) be meant to read as 'C-Style wrapper around `RFile::GetObjectFromKey` ... so I guess another option would be `RFile__GetObjectFromKey` )

io/io/inc/ROOT/RFile.hxx
```diff
@@ -231,7 +239,8 @@ class RFile final {
 
    /// Gets object `path` from the file and returns an **owning** pointer to it.
    /// The caller should immediately wrap it into a unique_ptr of the type described by `type`.
-   [[nodiscard]] void *GetUntyped(std::string_view path, const std::type_info &type) const;
+   [[nodiscard]] void *GetUntyped(std::string_view path,
+                                  std::variant<const char *, std::reference_wrapper<const std::type_info>> type) const;
```
pcanal, line 244: Since it is private and only use in the source file, would this be an option:
```suggestion
   template <typename T>
   [[nodiscard]] void *GetUntyped(std::string_view path,
                                  T type) const;
```
(`TClass::GetClass` itself has the 2 overloads).

io/io/inc/ROOT/RFile.hxx
```diff
@@ -23,13 +24,17 @@ class TKey;
...
 class RFile;
-struct RFileKeyInfo;
 
 namespace Internal {
 
 ROOT::RLogChannel &RFileLog();
 
+/// Returns an **owning** pointer to the object referenced by `key`. The caller must delete this pointer.
+/// This method is meant to only be used by the pythonization.
+[[nodiscard]] void *GetRFileObjectFromKey(RFile &file, const RKeyInfo &key);
```
jblomer: I think I would prefer a special handling of `Get<void>(...)` that doesn't check types.

jblomer (commented):
LGTM, I'll leave approval up to the other reviewers

io/io/inc/ROOT/RFile.hxx
silverweed: I think this can't work because it's also called from the header (in `Get`,which cannot be moved to the cxx)

io/io/inc/ROOT/RFile.hxx
silverweed: I'm not sure if you mean:
1. in this function (but since it's not templated I'm not sure what you mean)
2. *instead* of this function
3. in `RFile::Get` in general, unrelated to this function

io/io/inc/ROOT/RFile.hxx
pcanal: Fair enough.  This is then down to a stylistic choice.  You could still use the template but with one level of indirection (2 non templated one line overloads implemented in the source file and 1 templated `Impl` function)

io/io/inc/ROOT/RFile.hxx
jblomer: I mean in `RField::Get` in general, which would then make this function not needed.

PR #20097, opened 2025-10-13
[HS3] Minor Bugfixes, added Streamers and name sanitizing scheme 

Phmonski:
# Correction of PR #20052 
# This Pull request: 
- Added Streamers:
   - RooDecay
   - RooDerivative
   - RooAddModel
   - RooGaussModel
   - RooTruthModel

- New name sanitizing Tool

- Minor Bugfixes

PR #20171, opened 2025-10-22
[RF] ToyMCSampler improvements for binned toys

cburgard:
# This Pull request:

Adds the ability to control the ToyMCSampler to emit binned toys.
Makes use of this new capability in the HypoTestCalculator in cases where the observed data is binned.

## Checklist:

- [x] tested changes locally

This PR fixes #

### DIFF
--- a/roofit/roostats/src/HypoTestCalculatorGeneric.cxx
+++ b/roofit/roostats/src/HypoTestCalculatorGeneric.cxx
@@ -63,8 +63,20 @@ HypoTestCalculatorGeneric::HypoTestCalculatorGeneric(
                                                   *altModel.GetPdf(),
                                                   altModel.GetSnapshot());
 
-      fDefaultSampler = new ToyMCSampler(*fDefaultTestStat, 1000);
-      fTestStatSampler = fDefaultSampler;
+      auto toymcs = new ToyMCSampler(*fDefaultTestStat, 1000);
+      const bool dataIsBinned = dynamic_cast<const RooDataHist*>(fData);
+      if (dataIsBinned) {
+         // Ensure the ToyMCSampler generates toys with the same structure as the observed data.
+         // TODO: understand why this is needed only in the RooDataHist case,
+         // but results in backwards-incompatibility for RooDataSet, which
+         // manifests for example in the stressRooStats tests.
+         // See also: https://github.com/root-project/root/pull/20171
+         toymcs->SetProtoData(&data);
+      }
+      toymcs->SetGenerateBinned(dataIsBinned);  // if observed is RooDataHist -> generate RooDataHist toys
+
+      fDefaultSampler = toymcs;
+      fTestStatSampler = toymcs;
    }
 
 
--- a/roofit/roofitcore/inc/RooGlobalFunc.h
+++ b/roofit/roofitcore/inc/RooGlobalFunc.h
@@ -358,7 +358,7 @@ RooCmdArg Conditional(const RooArgSet& pdfSet, const RooArgSet& depSet, bool dep
  * @{
  */
 // RooAbsPdf::generate arguments
-RooCmdArg ProtoData(const RooDataSet& protoData, bool randomizeOrder=false, bool resample=false) ;
+RooCmdArg ProtoData(const RooAbsData& protoData, bool randomizeOrder=false, bool resample=false) ;
 RooCmdArg NumEvents(Int_t numEvents) ;
 RooCmdArg NumEvents(double numEvents) ;
 RooCmdArg AutoBinned(bool flag=true) ;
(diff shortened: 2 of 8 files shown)


cburgard:
This PR is still WIP. This is because the use case relies on https://github.com/root-project/root/pull/20097 being merged first (even though the functionalities are not closely related).

guitargeek:
I have merged #20097. Could you please rebase this PR?

Also, could you please add a commit where you add both of you to the list of contributors for ROOT 6.38 (insert in alphabetical order)?
https://github.com/root-project/root/blob/master/README/ReleaseNotes/v638/index.md

Thanks a lot!

cburgard:
> I have merged #20097. Could you please rebase this PR?
> 
> Also, could you please add a commit where you add both of you to the list of contributors for ROOT 6.38 (insert in alphabetical order)? https://github.com/root-project/root/blob/master/README/ReleaseNotes/v638/index.md
> 
> Thanks a lot!

Thanks a lot!

Done and done. I've removed the WIP marker now.

guitargeek (approved):
Thank you very much! This change makes sense to me.

roofit/roostats/src/HypoTestCalculatorGeneric.cxx
```diff
@@ -63,8 +63,14 @@ HypoTestCalculatorGeneric::HypoTestCalculatorGeneric(
                                                   *altModel.GetPdf(),
                                                   altModel.GetSnapshot());
 
-      fDefaultSampler = new ToyMCSampler(*fDefaultTestStat, 1000);
-      fTestStatSampler = fDefaultSampler;
+      auto toymcs = new ToyMCSampler(*fDefaultTestStat, 1000);
+      // --- Ensure the ToyMCSampler generates toys with the same structure as the observed data
+      toymcs->SetProtoData(&data);
```
guitargeek: This is a change in behavior that likely causes the RooStats tests to fail. I'll take a look.

guitargeek (changes requested):
This PR actually introduces a behavioral change that causes the RooStats tests like `stressRooStats` to fail. Even if the warning about duplicate args is silenced in [this line](https://github.com/root-project/root/blob/master/roofit/roofitcore/src/RooSimultaneous.cxx#L952) by doing `allVars.add(*prototype->get(), true)`.

@cburgard, did you intend the behavior change?

cburgard:
> This PR actually introduces a behavioral change that causes the RooStats tests like `stressRooStats` to fail. Even if the warning about duplicate args is silenced in [this line](https://github.com/root-project/root/blob/master/roofit/roofitcore/src/RooSimultaneous.cxx#L952) by doing `allVars.add(*prototype->get(), true)`.
> 
> @cburgard, did you intend the behavior change?

This is a bit of a hard question. I'll try to explain my reasoning here.

The core issue that triggered this PR is this:

When running the ToyMCSampler on a binned dataset, before this fix, this would happen:

```
cppyy.gbl.std.runtime_error: RooStats::HypoTestResult* RooStats::HypoTestCalculatorGeneric::GetHypoTest() => runtime_error: Error in RooAbsReal::setData(): only resetting with same-structured data is supported.
```

That is because the ToyMCSampler would always default to an unbinned dataset, and the way the HypoTestCalculator is setup it would not communicate the shape of the dataset to the underlying sampler, triggering this error.

In that sense, the change in behavior is intentional and is the main reason for this fix.

I would, however, have hoped that it would not affect the behavior on datasets that are already unbinned, where the ToyMCSampler (supposedly?) worked correctly already before.

The change of the behavior for binned datasets is intended and the point of this fix.
I do not currently have a test case to check whether there is any significant change in behavior also for unbinned datsets, and would certainly have hoped that any change there is minor or perhaps also to be considered a "fix" (perhaps there are some other properties of the toys that are more reflective of the example dataset now than they were before?)
But if there is an actual change in the quality of the estimate (that is, the resulting p-values of the test change for a scenario that is recognized as a unit test that was working correctly before my fix), that is probably then to be considerend "non-intentional".

So perhaps to fling this question back to you: Is there a unit test that produces notably different results than it did before, or is this just a case where the numerics changed a bit due to the toys now being more reflective of the example dataset than they were before?

guitargeek:
> So perhaps to fling this question back to you: Is there a unit test that produces notably different results than it did before, or is this just a case where the numerics changed a bit due to the toys now being more reflective of the example dataset than they were before?

Unfortunately, all the stressRooStats tests fail all over the place, with vastly different numeric results.

I discussed with Carsten privately, and we agreed that for now the change in behavior (i.e. setting the proto data) should only apply when using **RooDataHist**, because this was not supported before and therefore there is no backwards incompatible change.

roofit/roostats/src/HypoTestCalculatorGeneric.cxx
guitargeek: See https://github.com/root-project/root/pull/20171#issuecomment-3678918489

Issue #20174, opened 2025-10-22
TH1 and derivates projections produce incorrect results for histograms with infinite bin edges


siliataider:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

This issue continues the discussion from #20077, which originally reported that histogram projections and UHI operations fail for histograms with axes having infinite upper edges.

Part of the problem, the UHI slicing and projection behavior differing from the standard ROOT projection functions, was addressed in #20146.

However, the core C++ histogram projection methods ([TH2::ProjectionY](https://root.cern.ch/doc/v636/classTH2.html#a04917d549aa802ce2489692da74348f8) in this example) still produce incorrect results: when a histogram has an axis with an infinite upper bin edge, the contents of the highest bin are not included in the projected histogram.

### Reproducer

```cpp
int main() {
   Double_t xedges[] = {0., 1., 2.};
   Double_t yedges[] = {0., 1., 2., std::numeric_limits<double>::infinity()};

   TH2D h("h_inf", "h_inf;X;Y", 2, xedges, 3, yedges);

   h.SetBinContent(1, 1, 11.);
   h.SetBinContent(2, 1, 21.);
   h.SetBinContent(1, 2, 12.);
   h.SetBinContent(2, 2, 22.);
   h.SetBinContent(1, 3, 13.);
   h.SetBinContent(2, 3, 23.);

   std::cout << "ProjectionY values:" << std::endl;
   auto projY = h.ProjectionY();
   for (int i = 1; i <= projY->GetNbinsX(); ++i) {
      std::cout << "  bin " << i << ": " << projY->GetBinContent(i) << ", expected " << h.Integral(1, 2, i, i) << std::endl;
   }

   return 0;
}
```

Output:
```shell
ProjectionY values:
  bin 1: 32, expected 32
  bin 2: 34, expected 34
  bin 3: 0, expected 36
```

### ROOT version

6.37.01

### Installation method

all

### Operating system

all

ferdymercury:
The same happens with -infinite lower bin edges

@aapo-kossi an obvious workaround in the meantime would be to set the upper bin edge to 1e12 or sth like that.
Note that functions like GetBinCenter do not really work if an edge is infinite. Whereas if you put 1e12 it will be fine.

hageboeck:
[...]

PR #20176, opened 2025-10-23
[hist] do not look at bin center if one edge is infinite

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/20174

## Checklist:

- [x] tested changes locally

This PR fixes #

### DIFF
--- a/hist/hist/src/TAxis.cxx
+++ b/hist/hist/src/TAxis.cxx
@@ -477,15 +477,38 @@ Int_t TAxis::GetLast() const
 
 ////////////////////////////////////////////////////////////////////////////////
 /// Return center of bin
+///
+/// If fXmin is -inf, it will return -inf even if fXmax is finite
+/// If fXmax is +inf, it will return +inf even if fXmin is finite
 
 Double_t TAxis::GetBinCenter(Int_t bin) const
 {
    Double_t binwidth;
    if (!fXbins.fN || bin<1 || bin>fNbins) {
       binwidth = (fXmax - fXmin) / Double_t(fNbins);
+      if (std::isinf(binwidth)) {
+         if (std::isfinite(fXmin))
+            return fXmax;
+         else if (std::isfinite(fXmax))
+            return fXmin;
+         else if (fXmin == -std::numeric_limits<double>::infinity() && fXmax == std::numeric_limits<double>::infinity())
+            return 0.;
+         else
+            return std::numeric_limits<double>::quiet_NaN();
+      }
       return fXmin + (bin - 0.5) * binwidth;
    } else {
       binwidth = fXbins.fArray[bin] - fXbins.fArray[bin-1];
+      if (std::isinf(binwidth)) {
+         if (std::isfinite(fXbins.fArray[bin-1]))
+            return fXbins.fArray[bin];
+         else if (std::isfinite(fXbins.fArray[bin]))
+            return fXbins.fArray[bin-1];
+         else if (fXbins.fArray[bin-1] == -std::numeric_limits<double>::infinity() && fXbins.fArray[bin] == std::numeric_limits<double>::infinity())
+            return 0.;
+         else
+            return std::numeric_limits<double>::quiet_NaN();
+      }
       return fXbins.fArray[bin-1] + 0.5*binwidth;
    }
 }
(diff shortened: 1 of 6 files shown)


hist/hist/src/TAxis.cxx
```diff
@@ -477,13 +477,16 @@ Int_t TAxis::GetLast() const
...
+/// If fXmin is -inf, it will return -nan even if fXmax is finite
+/// If fXmax is +inf, it will return +inf even if fXmin is finite
 
 Double_t TAxis::GetBinCenter(Int_t bin) const
 {
    Double_t binwidth;
    if (!fXbins.fN || bin<1 || bin>fNbins) {
       binwidth = (fXmax - fXmin) / Double_t(fNbins);
-      return fXmin + (bin-1) * binwidth + 0.5*binwidth;
+      return fXmin + (bin - 0.5) * binwidth;
```
vepadulano: I have some doubts about this logic since we're introducing handling of infinite values. To me it's not so clear that they should not be treated separately. For example, I would have assumed we would have just added some logic to the function of the type

```
    double ret = fXmin + (bin - 0.5) * binwidth;
    if (std::isinf(ret) || std::isnan(ret))
        return std::numeric_limits<double>::quiet_NaN();
```

At least to have a homogeneous return value in this class of exceptional cases. I guess this also relates to users' expectations and how they would need to treat this afterwards. Thoughts @lmoneta ?

ferdymercury: Hmmm my view: I'd say that the bin center return value is correctly defined, it's infinite. It's not NaN.
So the problem is that one should not rely on GetBinCenter for finding corresponding bins between two axes.
A more elegant solution might be just to look linearly bin by bin what are the edges of each of the two axes and then create map from one to the other. This way one would not rely at all on GetBinCenter and eps.
But I thought the solution with eps was simple and fast enough.
Of course, another approach would be to say: infinite edges are not allowed in TH1, use instead 1e30 or sth like that, and then we get rid of the complexity of dealing with this.

vepadulano: > I'd say that the bin center return value is correctly defined, it's infinite. It's not NaN.

It's only infinite in case of the positive infinite upper bin edge. In case of the negative infinite lower bin edge, then it's going to be a NaN because of `fXmin + (bin - 0.5) * binwidth;` becoming practically a `-inf + inf` computation. So I don't see that the return value it's always correctly defined.

ferdymercury: You're right!

The drawback of all this is that we should update many function to deal with this.

GetBinLowEdge
GetBinUpEdge
GetBinWidth
GetBinCenterLog

So maybe we should just say -inf/+inf is unsupported? (workaround for the user is to just use 1e30 in that case?

aapo-kossi: As a user I agree with the sentiment that NaN could be returned for bin centers when there are infinite edges involved, or even an error thrown. However, this side-steps the issue that originally caused this PR to be created, which is that projecting a histogram to an axis that includes an infinite bin edge causes that bin to become empty even though intuitively the edges of bins in the remaining axis/axes is not involved in the calculation of the contents, which are just the sum of the corresponding bins along the other axis/axes. In my opinion there would be two clean solutions for this with varying amount of effort in coordination/implementation:
1. Document clearly in either the histogram classes or projection methods that infinite bin edges are not supported either at all or for producing histogram projections. This, as mentioned in the discussion of #20201, is more of a global issue within ROOT and might have to be separately handled in the wider case that inf values are completely disallowed.
2. Use bin indexing based logic in the DoProjection method instead of looking up the output bin by a coordinate value. I am not entirely sure how feasible this would be because I'm not so familiar with all the different projection cases but I don't Immediately see how it couldn't work.

vepadulano: Hi @aapo-kossi, thanks for your inputs, I believe they are extremely valuable! I see a need to first decide at the ROOT project level what is the actual policy for non-finite bin edges in ROOT histograms. Both the issue you opened and this PR show that, for all intents and purposes, the current status is "they are not supported and this fact is not documented". In the very short term, we could document this status, keeping the door open for a more educated decision regarding whether to disallow them by policy or to allow them with certain rules.

PR #20177, opened 2025-10-23
[ntuple] add exporter option to decompress pages

jblomer:
Helps for compression studies: to experiment with different compression algorithms, we want to compare packed but uncompressed pages (i.e., different to just uncompressing the file with hadd and exporting the uncompressed file).

### DIFF
--- a/tree/ntupleutil/src/RNTupleExporter.cxx
+++ b/tree/ntupleutil/src/RNTupleExporter.cxx
@@ -13,13 +13,16 @@
  * For the list of contributors see $ROOTSYS/README/CREDITS.             *
  *************************************************************************/
 
+#include <ROOT/RError.hxx>
 #include <ROOT/RNTupleExporter.hxx>
 #include <ROOT/RPageStorage.hxx>
 #include <ROOT/RNTupleDescriptor.hxx>
+#include <ROOT/RNTupleZip.hxx>
 #include <ROOT/RClusterPool.hxx>
 #include <ROOT/RLogger.hxx>
 #include <fstream>
 #include <sstream>
+#include <vector>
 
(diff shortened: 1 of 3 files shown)


tree/ntupleutil/src/RNTupleExporter.cxx
```diff
@@ -158,22 +165,33 @@ RNTupleExporter::ExportPages(ROOT::Internal::RPageSource &source, const RPagesOp
...
                   R__FAIL(std::string("output path ") + options.fOutputPath + " does not exist or is not writable!"));
+            }
 
-            outFile.write(reinterpret_cast<const char *>(pageBuf), pageBufSize);
+            // dump the page
+            const auto *pageBuf = static_cast<const char *>(onDiskPage->GetAddress());
+            if (options.fFlags & RPagesOptions::kDecompress) {
+               const auto nbytesPacked = colElement->GetPackedSize(pageInfo.GetNElements());
+               const auto nbytesData = pageInfo.GetLocator().GetNBytesOnStorage();
+               auto outBuf = ROOT::Internal::MakeUninitArray<char>(nbytesPacked);
```
silverweed: Not sure how impactful on overall performance, but we could allocate this buffer once at the start and then grow it when necessary, to avoid potentially hundreds of thousands of allocations/deallocations

enirolf (approved):
LGTM, thanks!

tree/ntupleutil/src/RNTupleExporter.cxx
jblomer: Good point!

silverweed (approved):
lgtm!

PR #20180, opened 2025-10-23
[hist] Implement initial `RHistAutoAxisFiller`

hahnjo:
To fill a regular one-dimensional histogram without specifying an axis interval during construction.

### DIFF
--- a/hist/histv7/inc/ROOT/RHistAutoAxisFiller.hxx
+++ b/hist/histv7/inc/ROOT/RHistAutoAxisFiller.hxx
@@ -56,6 +56,8 @@ private:
    std::size_t fNNormalBins;
    /// The maximum buffer size until Flush() is automatically called
    std::size_t fMaxBufferSize;
+   /// The fraction of the axis interval to use as margin
+   double fMarginFraction;
 
    using BufferElement = std::conditional_t<SupportsWeightedFilling, std::pair<double, RWeight>, double>;
 
@@ -71,8 +73,9 @@ public:
    ///
    /// \param[in] nNormalBins the number of normal bins, must be > 0
    /// \param[in] maxBufferSize the maximum buffer size, must be > 0
-   explicit RHistAutoAxisFiller(std::size_t nNormalBins, std::size_t maxBufferSize = 1024)
-      : fNNormalBins(nNormalBins), fMaxBufferSize(maxBufferSize)
+   /// \param[in] marginFraction the fraction of the axis interval to use as margin, must be > 0
+   explicit RHistAutoAxisFiller(std::size_t nNormalBins, std::size_t maxBufferSize = 1024, double marginFraction = 0.05)
+      : fNNormalBins(nNormalBins), fMaxBufferSize(maxBufferSize), fMarginFraction(marginFraction)
    {
       if (nNormalBins == 0) {
          throw std::invalid_argument("nNormalBins must be > 0");
@@ -80,10 +83,14 @@ public:
       if (maxBufferSize == 0) {
          throw std::invalid_argument("maxBufferSize must be > 0");
       }
+      if (marginFraction <= 0) {
+         throw std::invalid_argument("marginFraction must be > 0");
+      }
    }
 
    std::size_t GetNNormalBins() const { return fNNormalBins; }
    std::size_t GetMaxBufferSize() const { return fMaxBufferSize; }
+   double GetMarginFraction() const { return fMarginFraction; }
 
 private:
    void BufferImpl(double x, RWeight weight)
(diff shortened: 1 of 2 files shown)


hist/histv7/inc/ROOT/RHistAutoAxisFiller.hxx
```diff
@@ -0,0 +1,195 @@
...
+         // Silence compiler warning about unused parameter
+         (void)weight;
+         fBuffer.push_back(x);
+      }
+      if (x < fMinimum) {
+         fMinimum = x;
+      }
+      if (x > fMaximum) {
+         fMaximum = x;
+      }
```
jblomer:
```suggestion
      fMinimum = std::min(x, fMinimum);
      fMaximum = std::max(x, fMaximum);
```

hist/histv7/inc/ROOT/RHistAutoAxisFiller.hxx
```diff
@@ -0,0 +1,195 @@
...
+      fBuffer.clear();
+   }
+
+   /// Return the constructed histogram.
+   ///
+   /// \see Flush()
+   RHist<BinContentType> &GetHist()
+   {
+      Flush();
```
jblomer, line 193: That means that we can't get an empty histogram from the auto filler. Is this the desired behavior?

hist/histv7/inc/ROOT/RHistAutoAxisFiller.hxx
```diff
@@ -0,0 +1,195 @@
...
+         return;
+      }
+      BufferImpl(x, weight);
+   }
+
+   /// Flush the buffer of entries and construct the histogram.
+   ///
+   /// Throws an exception if the buffer is empty, the axis interval cannot be determined, or if it would be empty
+   /// because the minimum equals the maximum.
+   void Flush()
```
jblomer, line 153: Do we want/need to expose this publicly?

hist/histv7/test/hist_auto.cxx
```diff
@@ -0,0 +1,162 @@
...
+   RHistAutoAxisFiller<int> filler(Bins);
+
+   filler.Fill(-1);
+   filler.Fill(0);
+
+   auto &hist = filler.GetHist();
+   EXPECT_EQ(hist.GetBinContent(RBinIndex::Underflow()), 0);
+   EXPECT_EQ(hist.GetBinContent(RBinIndex::Overflow()), 0);
+}
+
```
ferdymercury, line 89: Thanks!
For completeness, I would propose copy-pasting the CTest "TTreeDraw_AutoBinning" from https://github.com/root-project/root/pull/19605/files into this class

Ie filling -999 and 0 and checking that auto-axis does not put them in the overflow or underflow.

hahnjo: I briefly looked at the test case, which additional issue does it cover? In other words, what is special about `-999`? Note that the test for `Fill` above also already checks that all values landed in the normal bins. I'm against adding more tests that don't increase coverage just for completeness.

ferdymercury: > In other words, what is special about `-999`?

In a case where the user filled with 0 and -999, the automatic range calculation was sometimes missing the 0. (putting it in the overflow rather than in the last bin), due to some floating precision effects if I recall correctly.

https://root-forum.cern.ch/t/bug-or-feature-in-ttree-draw/62862

(Checking 0 and -1 might not trigger the issue since the relative error is smaller)

hahnjo: From the forum:
> In this example the negatives (-999) are relatively far from 0, but I also tested down to -1 and the same still happens.

That's also what I saw with my first naive implementation, it was just the upper bound of zero that was problematic if used as a factor. This should be avoided by relying on `std::nextafter`, I hope.

hist/histv7/inc/ROOT/RHistAutoAxisFiller.hxx
hahnjo: Yes correct. It's a good question what we want to happen in this case. If we want to allow an empty histogram, what would be the axis interval and what would we want to happen in case of further fills? Would we modify the constructed histogram (potentially) behind the users' back? If there are many fills, how would the user tell is that *now* is the time to flush the buffer, while they are already looking at the histogram?

I think my proposal for the moment would be to disallow and see if it poses problems that we find a good compromise how to address.

hist/histv7/inc/ROOT/RHistAutoAxisFiller.hxx
hahnjo: Another good question. A potential use case would be where the user designates a limited set of "sample" entries that are potentially not aligned with `fMaxBufferSize`, or multiple `RHistAutoAxisFiller` with different fill frequencies that the user wants to "synchronize".

hist/histv7/inc/ROOT/RHistAutoAxisFiller.hxx
```diff
@@ -0,0 +1,147 @@
...
+         throw std::runtime_error("could not determine axis interval");
+      }
+      if (fMinimum == fMaximum) {
+         throw std::runtime_error("axis interval is empty");
+      }
+
+      // Slightly increase the upper limit to make sure the maximum is included in the last bin.
+      double high = std::nextafter(fMaximum, std::numeric_limits<double>::infinity());
+      assert(high > fMaximum);
+      fHist.emplace(fNNormalBins, std::make_pair(fMinimum, high));
```
hageboeck: I would suggest to extend the range with a margin of a few percent to each side.

I observed (and did it myself) that many people do it anyway when they create their "final" plots because it looks a bit more pleasing, but it also ensures that future values close to the min/max will be inside the range. The threshold is obviously arbitrary, one could start with a margin of 5%. Note also that this will probably be used mostly for the first attempts, and fixed limits will be decided based on these first looks. In that case, it's more important that the entire distribution is in range than computing the tightest possible range.
```
const auto margin = 0.05 * (fMaximum - fMinimum);
const auto high = fMaximum + margin;
const auto low = fMinimum - margin;
```

hahnjo: Done. I made the fraction configurable, should the user wish to change it.

PR #20183, opened 2025-10-23
[hist] Implement `RHistEngine::FillAtomic`

hahnjo:


hahnjo:
It seems the latest code works fine to implement atomic operations on Windows :partying_face:

hahnjo:
> Perhaps add a small multi-threaded (stress) test to check that the addition is indeed atomic.

Done (it's of course not a guarantee that everything is right since reliably testing race conditions is hard)

hahnjo:
ping

siliataider (approved):
Really nice!

PR #20191, opened 2025-10-24
[skip-ci][ntuple][doc] Add reference doc for RNTuple merging

silverweed:
Following [this comment](https://github.com/root-project/root/pull/20112#pullrequestreview-3375618729)'s suggestion.
This documentation is meant to be useful for future reference for ourselves (and maybe advanced users who need this information) and we should try to keep it up-to-date as the Merger is updated.

### DIFF
--- a/tree/ntuple/doc/Merging.md
+++ b/tree/ntuple/doc/Merging.md
@@ -18,15 +18,17 @@ Please note that the RNTupleMerger is currently experimental and the content of 
 ## Goal
 The goal of the RNTuple merging process is producing one output RNTuple from *N* input RNTuples that can be used as if it were produced directly in the merged state. This means that:
 
-* R1: All fields in the output RNTuple are accessible and have a type compatible with the original fields of the input RNTuples.
+* R1: All fields in the output RNTuple are accessible and have a type compatible<sup>1</sup> with the original fields of the input RNTuples.
 * R2: The values of those fields are a concatenation of the original fields. If the first input RNTuple had *M* entries, the first *M* entries of the output RNTuple map to those entries; entry *M+1* maps to the first entry of the second input RNTuple, and so on.
+
+<sup>1</sup>: currently "compatible" means "identical". This may be extended in the future to include fields that have convertible types.
 
 At a lower level, we require that:
 
 * R3: the output RNTuple has the **same non-extended schema description** as the **first input RNTuple**;
 * R4: the output RNTuple has an **extended schema description** that is:
     *  **strictly equal** to the first input (`Filter` and `Strict` mode), or
-    *  **equal or a superset** of the first input (`Union` mode)
+    *  **a superset** of the first input (`Union` mode)
 
 Consequences of R3 and R4:
 
@@ -36,40 +38,49 @@ Consequences of R3 and R4:
 The following properties are currently true but they are subject to change:
 
 * P1: all output pages have the **same compression** (which may be different from the input pages' compression);
-* P2: the output clusters are **the same as the input clusters**;
-* P3: the output RNTuple **always has 1 cluster group**
+* P2: all pages in the same output column have the **same encoding** (which may be different from the inputs' encoding);
+* P3: the output clusters are **the same as the input clusters**;
+* P4: the output RNTuple **always has 1 cluster group**
+
+Note that these properties influence and are influenced by the level of merging used.
+E.g. P1 and P2 are currently true because we only support L1 merging of pages with identical compressions. This is a limitation that we intend to lift at some point (both for L1 and L0 if we ever support it).
+P3 and P4 would not necessarily be true with L4 support (which might be desirable in some cases, e.g. to group pages into smaller/larger clusters).
+
+Therefore we *will* want to drop these properties at some point, in order to improve the capabilities of the Merger.
 
 ## High-level description
-The merging process requires at least 1 input, consisting in a `RPageSource`.
+The merging process requires at least 1 input, in the form of an `RPageSource`.
 
-The first input is attached in `EDescriptorDeserializeMode::kForWriting` mode, which doesn't collate the extended header with the non-extended header. Since we use the first input's descriptor as the output schema (barring late model extensions, see later), opening in `kForWriting` mode allows us to write the output to disk preserving the non-extended schema description as per requirement R3. A consequence of this choice is that the merger never produces (new) deferred columns in the output RNTuple's header.
+The first input is attached in `EDescriptorDeserializeMode::kForWriting` mode, which doesn't collate the extended header with the non-extended header. Since we use the first input's descriptor as the output schema (barring late model extensions, see later), opening in `kForWriting` mode allows us to write the output to disk while preserving the non-extended schema description as per requirement R3. A consequence of this choice is that the merger never produces (new) deferred columns in the output RNTuple's header.
 
-In `Union` mode only, we allow any following input RNTuple to define new fields that don't appear in the first input. These fields, after being validated, are late model extended into the output model and will thus appear in the output RNTuple's extended header on disk. This means that all columns that were not part of the first input's schema become deferred columns in the output RNTuple (unless the first source had 0 entries).
+In `Union` mode only, we allow any subsequent input RNTuple to define new fields that don't appear in the first input. These fields, after being validated, are late model extended into the output model and will thus appear in the output RNTuple's extended header on disk. This means that all columns that were not part of the first input's schema become deferred columns in the output RNTuple (unless the first source had 0 entries).
 
 ## Descriptor compatibility and validation
 Whenever a new input is processed, we compare its descriptor with the output descriptor to verify that merging is possible.
 
 The comparison function does 3 main things:
-- collect all "extra destination fields" (i.e. fields that exist in the destination but not in this input RNTuple)
-- collect all "extra source fields"
+- collect all "extra destination fields" (i.e. fields that exist in the output but not in this input RNTuple)
+- collect all "extra source fields" from the input RNTuple
 - collect and validate all common fields.
 
 If the Merging Mode is set to **Filter** we require the "extra destination fields" list to be empty.
-If the Merging Mode is set to **Strict** we require both "extra destination fields" and "extra source fields" to be empty.
+If the Merging Mode is set to **Strict** we require both the "extra destination fields" and "extra source fields" lists to be empty.
 If the Merging Mode is set to **Union**, the "extra source fields" list is used to late model extend the destination model.
 
 As for common fields, they are matched by name and validated as follows:
 - any field that is projected in the destination must be also projected in the source and must be projected to the same field;
 - any field that is not projected in the destination must also not be projected in the source;
-- the field types names must be **identical** (*this could probably be relaxed in the future to allow for different but compatible types*)
+- the field types names must be **identical** (*this could probably be relaxed in the future to allow for different but compatible types - see requirement R1*)
 - the type checksums, if present, must be identical. Note that if a field has a type checksum and the other doesn't, we consider this valid (*is this sound?*);
 - the type versions must be identical;
 - the fields' structural roles must be identical;
-- the column representations must match, as follows:
+- the column representations must match<sup>1</sup>, as follows:
     - the source and destination fields must have the same number of columns;
     - the types of each column must either be identical or one must be the split/unsplit version of the other;
     - the bits on storage of both columns must be identical;
     - the value range of both columns must be identical;
-    - the representation index of the each source column must be 0; (*why?*)
-- if the fields have subfields, the number of subfields must be identical, and each source subfield is recursively validated against  its destination counterpart via all the rules described in this list.
+    - the representation index of the each source column must be 0 (i.e. we currently don't support multiple columns representations while merging);
+- if the fields have subfields, the number of subfields must be identical, and each source subfield is recursively validated against its destination counterpart via all the rules described in this list.
 
+
+<sup>1</sup>: these restrictions will likely not be required for L4 merging.


tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+    * **L3 Merge** (*reseal* merge): merge a page by resealing/repacking it (implies decompressing/recompressing, may change the encoding)
+    * **L4 Merge** (*slow* merge): merge a page by re-reading and re-writing all values, potentially changing the pages' boundaries and clustering (implies resealing and recompressing)
+
+(*NOTE: currently there is no guarantee for the user about which mode will be used to generate the merged RNTuple. At the moment, L0 and L4 are never used; L1 is used when possible, otherwise L3 is used. Note that we currently don't use L2 because in general when recompressing we might need to change encoding. Improvements on this front are possible and in principle we should be able to use L2 when the encoding doesn't need to change.*)
+
+## Goal
+The goal of the RNTuple merging process is producing one output RNTuple from *N* input RNTuples that can be used as if it were produced directly in the merged state. This means that:
+
+* R1: All fields in the output RNTuple are accessible and have a type compatible with the original fields of the input RNTuples.
```
vepadulano: About the "compatible" bit, it would be good to also clarify how compatibility is defined in this context

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+* P2: the output clusters are **the same as the input clusters**;
+* P3: the output RNTuple **always has 1 cluster group**
+
+## High-level description
+The merging process requires at least 1 input, consisting in a `RPageSource`.
+
+The first input is attached in `EDescriptorDeserializeMode::kForWriting` mode, which doesn't collate the extended header with the non-extended header. Since we use the first input's descriptor as the output schema (barring late model extensions, see later), opening in `kForWriting` mode allows us to write the output to disk preserving the non-extended schema description as per requirement R3. A consequence of this choice is that the merger never produces (new) deferred columns in the output RNTuple's header.
+
+In `Union` mode only, we allow any following input RNTuple to define new fields that don't appear in the first input. These fields, after being validated, are late model extended into the output model and will thus appear in the output RNTuple's extended header on disk. This means that all columns that were not part of the first input's schema become deferred columns in the output RNTuple (unless the first source had 0 entries).
```
vepadulano: > unless the first source had 0 entries

In which case the non-extended output schema is equal to the non-extended schema of the second input? Or is there something more?

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+- collect and validate all common fields.
+
+If the Merging Mode is set to **Filter** we require the "extra destination fields" list to be empty.
+If the Merging Mode is set to **Strict** we require both "extra destination fields" and "extra source fields" to be empty.
+If the Merging Mode is set to **Union**, the "extra source fields" list is used to late model extend the destination model.
+
+As for common fields, they are matched by name and validated as follows:
+- any field that is projected in the destination must be also projected in the source and must be projected to the same field;
+- any field that is not projected in the destination must also not be projected in the source;
+- the field types names must be **identical** (*this could probably be relaxed in the future to allow for different but compatible types*)
```
vepadulano: Although not in direct contrast with R1, this sounds related to it, maybe it should be mentioned?

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+- the type checksums, if present, must be identical. Note that if a field has a type checksum and the other doesn't, we consider this valid (*is this sound?*);
```
pcanal, line 74: Side note that mode `(L4)` would allow this to be fully relaxed.
- the input any checksum will do (within the constraint of regular schema evolution support)
- the destination checksum would have the same relationship to the in-memory class layout/checksum as with regular RNTuple Write (i.e. be the same)

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+- the type versions must be identical;
+- the fields' structural roles must be identical;
+- the column representations must match, as follows:
```
pcanal: Those limitations are also likely removable for mode `(L4)`

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+* all projected fields in the input RNTuples remain projected in the output and have the same source field.
+ 
+The following properties are currently true but they are subject to change:
+
+* P1: all output pages have the **same compression** (which may be different from the input pages' compression);
+* P2: the output clusters are **the same as the input clusters**;
+* P3: the output RNTuple **always has 1 cluster group**
+
+## High-level description
+The merging process requires at least 1 input, consisting in a `RPageSource`.
```
jblomer:
```suggestion
The merging process requires at least 1 input in the form of an `RPageSource`.
```

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+If the Merging Mode is set to **Filter** we require the "extra destination fields" list to be empty.
+If the Merging Mode is set to **Strict** we require both "extra destination fields" and "extra source fields" to be empty.
+If the Merging Mode is set to **Union**, the "extra source fields" list is used to late model extend the destination model.
+
+As for common fields, they are matched by name and validated as follows:
+- any field that is projected in the destination must be also projected in the source and must be projected to the same field;
+- any field that is not projected in the destination must also not be projected in the source;
+- the field types names must be **identical** (*this could probably be relaxed in the future to allow for different but compatible types*)
+- the type checksums, if present, must be identical. Note that if a field has a type checksum and the other doesn't, we consider this valid (*is this sound?*);
+- the type versions must be identical;
```
jblomer, line 75: we should also make sure that the field versions are identical (even though we don't use them yet)

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+- the fields' structural roles must be identical;
+- the column representations must match, as follows:
+    - the source and destination fields must have the same number of columns;
+    - the types of each column must either be identical or one must be the split/unsplit version of the other;
+    - the bits on storage of both columns must be identical;
+    - the value range of both columns must be identical;
+    - the representation index of the each source column must be 0; (*why?*)
```
jblomer: Not sure.. perhaps it means that we currently don't support multi-column representations as an input?

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+The following properties are currently true but they are subject to change:
+
+* P1: all output pages have the **same compression** (which may be different from the input pages' compression);
+* P2: the output clusters are **the same as the input clusters**;
+* P3: the output RNTuple **always has 1 cluster group**
+
+## High-level description
+The merging process requires at least 1 input, consisting in a `RPageSource`.
+
+The first input is attached in `EDescriptorDeserializeMode::kForWriting` mode, which doesn't collate the extended header with the non-extended header. Since we use the first input's descriptor as the output schema (barring late model extensions, see later), opening in `kForWriting` mode allows us to write the output to disk preserving the non-extended schema description as per requirement R3. A consequence of this choice is that the merger never produces (new) deferred columns in the output RNTuple's header.
```
enirolf:
```suggestion
The first input is attached in `EDescriptorDeserializeMode::kForWriting` mode, which doesn't collate the extended header with the non-extended header. Since we use the first input's descriptor as the output schema (barring late model extensions, see later), opening in `kForWriting` mode allows us to write the output to disk, preserving the non-extended schema description as per requirement R3. A consequence of this choice is that the merger never produces (new) deferred columns in the output RNTuple's header.
```

tree/ntuple/doc/Merging.md
enirolf:
```suggestion
In `Union` mode only, we allow any subsequent input RNTuple to define new fields that don't appear in the first input. These fields, after being validated, are late model extended into the output model and will thus appear in the output RNTuple's extended header on disk. This means that all columns that were not part of the first input's schema become deferred columns in the output RNTuple (unless the first source had 0 entries).
```

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+
+In `Union` mode only, we allow any following input RNTuple to define new fields that don't appear in the first input. These fields, after being validated, are late model extended into the output model and will thus appear in the output RNTuple's extended header on disk. This means that all columns that were not part of the first input's schema become deferred columns in the output RNTuple (unless the first source had 0 entries).
+
+## Descriptor compatibility and validation
+Whenever a new input is processed, we compare its descriptor with the output descriptor to verify that merging is possible.
+
+The comparison function does 3 main things:
+- collect all "extra destination fields" (i.e. fields that exist in the destination but not in this input RNTuple)
```
enirolf:
```suggestion
- collect all "extra destination fields" (i.e. fields that exist in the output but not in this input RNTuple)
```

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+- collect all "extra source fields"
```
enirolf:
```suggestion
- collect all "extra source fields" from the input RNTuple
```

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+- collect and validate all common fields.
+
+If the Merging Mode is set to **Filter** we require the "extra destination fields" list to be empty.
+If the Merging Mode is set to **Strict** we require both "extra destination fields" and "extra source fields" to be empty.
```
enirolf:
```suggestion
If the Merging Mode is set to **Strict** we require both the "extra destination fields" and "extra source fields" lists to be empty.
```

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+- the type checksums, if present, must be identical. Note that if a field has a type checksum and the other doesn't, we consider this valid (*is this sound?*);
+- the type versions must be identical;
+- the fields' structural roles must be identical;
+- the column representations must match, as follows:
+    - the source and destination fields must have the same number of columns;
+    - the types of each column must either be identical or one must be the split/unsplit version of the other;
+    - the bits on storage of both columns must be identical;
+    - the value range of both columns must be identical;
+    - the representation index of the each source column must be 0; (*why?*)
+- if the fields have subfields, the number of subfields must be identical, and each source subfield is recursively validated against  its destination counterpart via all the rules described in this list.
```
enirolf:
```suggestion
- if the fields have subfields, the number of subfields must be identical, and each source subfield is recursively validated against its destination counterpart via all the rules described in this list.
```

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+* R4: the output RNTuple has an **extended schema description** that is:
+    *  **strictly equal** to the first input (`Filter` and `Strict` mode), or
+    *  **equal or a superset** of the first input (`Union` mode)
+
+Consequences of R3 and R4:
+
+* all columns that were deferred in the first RNTuple remain deferred in the output RNTuple and have the same FirstElementIdx.
+* all projected fields in the input RNTuples remain projected in the output and have the same source field.
+ 
+The following properties are currently true but they are subject to change:
```
enirolf, line 38: Maybe worth mentioning that these properties dictate with level of merging is used?

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+
+* P1: all output pages have the **same compression** (which may be different from the input pages' compression);
+* P2: the output clusters are **the same as the input clusters**;
+* P3: the output RNTuple **always has 1 cluster group**
```
enirolf: P4: all output columns have the same encoding?

tree/ntuple/doc/Merging.md
silverweed: This is currently kept vague on purpose as we don't have a precise definition valid for the merger yet. At the moment "compatible" for the merger means "exactly the same type", but we know we can do better in principle. I can add a note that says this in the document if you'd like.

tree/ntuple/doc/Merging.md
silverweed: No, the schema will still be the union of the first two but the columns of the second won't be deferred because they still start at index 0 (since the first input had 0 entries).

Btw you made me realize we weren't testing this case so I added a test: https://github.com/root-project/root/pull/20241

tree/ntuple/doc/Merging.md
silverweed: Right, we currently don't support extending the model for column representations; so this is a requirement that may be lifted in the future

tree/ntuple/doc/Merging.md
silverweed: Maybe a `while` instead of the comma would communicate the intended message better? (the idea is that we could still write the output to disk, but we wouldn't be preserving the schema description - whereas the comma makes it sound as if the "writing to disk" is the thing that would be allowed by this mode)

tree/ntuple/doc/Merging.md
silverweed: Correct, though it's more like "all pages in the same output column have the same encoding"

tree/ntuple/doc/Merging.md
silverweed: https://github.com/root-project/root/pull/20243

tree/ntuple/doc/Merging.md
```diff
@@ -0,0 +1,75 @@
...
+
+* R1: All fields in the output RNTuple are accessible and have a type compatible with the original fields of the input RNTuples.
+* R2: The values of those fields are a concatenation of the original fields. If the first input RNTuple had *M* entries, the first *M* entries of the output RNTuple map to those entries; entry *M+1* maps to the first entry of the second input RNTuple, and so on.
+
+At a lower level, we require that:
+
+* R3: the output RNTuple has the **same non-extended schema description** as the **first input RNTuple**;
+* R4: the output RNTuple has an **extended schema description** that is:
+    *  **strictly equal** to the first input (`Filter` and `Strict` mode), or
+    *  **equal or a superset** of the first input (`Union` mode)
```
hahnjo: nit: superset already includes equality :wink:

PR #20192, opened 2025-10-24
Allow both 'bool' and 'Bool_t' type conversions to numpy booleans

### LINKED ISSUE #20081: Also consider `Bool_t` in numpy array conversion

As reported at https://root-forum.cern.ch/t/bool-branch-becomes-python-object-with-rdataframe-asnumpy/64305

The logic at https://github.com/root-project/root/blob/477137c68fac415c9b186b3f4ca7a0d9247b2482/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdataframe.py#L302 only considers the `bool` case, but not the `Bool_t` case, which is what a TTree branch holding boolean values will display to RDataFrame.

### ROOT version

Any

### Installation method

Any

### Operating system

Any

### Additional context

_No response_

vepadulano:
Fixes #20081

### DIFF
--- a/bindings/pyroot/pythonizations/test/rdataframe_asnumpy.py
+++ b/bindings/pyroot/pythonizations/test/rdataframe_asnumpy.py
@@ -7,7 +7,7 @@ from pathlib import Path
 import numpy as np
 import ROOT
 from ROOT._pythonization._rdataframe import _clone_asnumpyresult
-
+import os
 
 def make_tree(*dtypes):
     """
@@ -90,10 +90,22 @@ class RDataFrameAsNumpy(unittest.TestCase):
         Test bool data-type as a special case since we cannot adopt
         the std::vector<bool> with numpy arrays
         """
-        df = ROOT.RDataFrame(2).Define("x", "bool(rdfentry_)")
+        treename = "test_branch_bool"
+        filename = "test_branch_bool.root"
+        # Snapshot a TTree so that column 'x' will be of type 'Bool_t'
+        ROOT.RDataFrame(2).Define("x", "bool(rdfentry_)").Snapshot(treename, filename)
+        # The column 'y' will instead have type 'bool'
+        df = ROOT.RDataFrame(treename, filename).Define("y", "bool(rdfentry_)")
+        self.assertEqual(df.GetColumnType("x"), "Bool_t")
+        self.assertEqual(df.GetColumnType("y"), "bool")
         npy = df.AsNumpy()
-        self.assertFalse(bool(npy["x"][0]))
-        self.assertTrue(bool(npy["x"][1]))
+        # Both numpy arrays should have dtype bool
+        self.assertEqual(npy["x"].dtype, bool)
+        self.assertEqual(npy["y"].dtype, bool)
+        self.assertFalse(npy["x"][0])
+        self.assertTrue(npy["x"][1])
+        self.assertFalse(npy["y"][0])
+        self.assertTrue(npy["y"][1])
 
     def test_read_array(self):
         """
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdataframe.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rdataframe.py
@@ -296,10 +296,6 @@ def RDataFrameAsNumpy(
     result_ptrs = {}
     for column in columns:
         column_type = df.GetColumnType(column)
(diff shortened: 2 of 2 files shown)


bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rvec.py
```diff
@@ -62,16 +62,73 @@
...
+    "std::size_t": "i",
+    # Extended standard integer types
+    "std::int8_t": "i",
+    "std::int16_t": "i",
+    "std::int32_t": "i",
+    "std::int64_t": "i",
+    "std::uint8_t": "u",
+    "std::uint16_t": "u",
+    "std::uint32_t": "u",
+    "std::uint64_t": "u",
```
guitargeek, line 103:
```suggestion
    "std::int8_t": "i1",
    "std::int16_t": "i2",
    "std::int32_t": "i4",
    "std::int64_t": "i8",
    "std::uint8_t": "u1",
    "std::uint16_t": "u2",
    "std::uint32_t": "u4",
    "std::uint64_t": "u8",
```
Also for the other non-32 bit integer types added in this dictionary.

Quoting from the NumPy docs you linked:
> A string providing the basic type of the homogeneous array The basic string format consists of 3 parts: a character describing the byteorder of the data (<: little-endian, >: big-endian, |: not-relevant), a character code giving the basic type of the array, and an integer providing the number of bytes the type uses.


The endian part is not important because we can assume always little endian, but I would make sure we use a consistent number of bytes. Users might otherwise we surprised by unexpectedly high memory usage when taking data from an RDataFrame.

bindings/pyroot/pythonizations/test/rdataframe_asnumpy.py
```diff
@@ -90,10 +90,25 @@ def test_branch_bool(self):
...
+            npy = df.AsNumpy()
+            # Both numpy arrays should have dtype bool
+            self.assertEqual(npy["x"].dtype, bool)
+            self.assertEqual(npy["y"].dtype, bool)
+            self.assertFalse(npy["x"][0])
+            self.assertTrue(npy["x"][1])
+            self.assertFalse(npy["y"][0])
+            self.assertTrue(npy["y"][1])
+        finally:
+            os.remove(filename)
```
guitargeek: This results in test failures on Windows:
```txt
2025-10-24T20:22:30.9438074Z ----------------------------------------------------------------------
2025-10-24T20:22:30.9438634Z Traceback (most recent call last):
2025-10-24T20:22:30.9439374Z   File "C:\ROOT-CI\src\bindings\pyroot\pythonizations\test\rdataframe_asnumpy.py", line 111, in test_branch_bool
2025-10-24T20:22:30.9440075Z     os.remove(filename)
2025-10-24T20:22:30.9440834Z PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'test_branch_bool.root'
```

There are many other tests that don't clean up temporary files that they produce either. So IMHO it's fine to fix the test on Windows by just keeping the file around.

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rvec.py
```diff
@@ -62,16 +62,73 @@
...
+    "char16_t": "i",
+    "char32_t": "i",
+    # Floating-point types
+    # C++ standard floating-point types
+    "float": "f",
+    "double": "f",
+    "long double": "f",
+    # ROOT floating-point types
+    "Float_t": "f",
+    "Double_t": "f",
```
guitargeek, line 131:
```suggestion
    "float": "f4",
    "double": "f8",
    "long double": "f16",
    # ROOT floating-point types
    "Float_t": "f4",
    "Double_t": "f8",
```
These types are also understood by NumPy, and without using `f8` and `f16` you would actually loose information here because the 4-byte float32 is the default `f`:

```txt
Python 3.13.8 (main, Oct  7 2025, 12:01:51) [GCC 14.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> a = np.array([0], dtype="f")
>>> a.dtype
dtype('float32')
>>>
```

guitargeek (approved):
Nevermind, my previous change request. I have now remembered that the size is later appended to the dtype string as the result of `sizeof`.

So this PR looks good to me, except for the failure on Windows, which can be easily avoided by not deleting the file!

PR #14246, opened 2023-12-15
[cling] Enable JITLink for AArch64 and x86_64 on Linux

devajithvs:
# This Pull request:

## Checklist:

- [ ] tested changes locally
- [ ] updated the docs (if necessary)

This PR fixes #

PR #20193, opened 2025-10-24
[cling] Enable JITLink debugging/profiling

devajithvs:
# This Pull request:

`CLING_PROFILE=1` enables perf profiling using `jitdump` format.

Profiling now requires a `perf inject` step (with JITLink):
```
perf record -k 1 <cling>
perf inject -j -i perf.data -o perf.jitted.data
perf report -i perf.jitted.data
```

## Changes or fixes:
Should be merged before merging: https://github.com/root-project/root/pull/14246

## Checklist:

- [x] tested changes locally on gdb, lldb and perf version 6.14.0

This PR fixes #

### DIFF
--- a/interpreter/cling/lib/Interpreter/IncrementalJIT.cpp
+++ b/interpreter/cling/lib/Interpreter/IncrementalJIT.cpp
@@ -19,9 +19,13 @@
 #include <clang/Basic/TargetOptions.h>
 #include <clang/Frontend/CompilerInstance.h>
 
+#include <llvm/ExecutionEngine/Orc/Debugging/DebugInfoSupport.h>
+#include <llvm/ExecutionEngine/Orc/Debugging/DebuggerSupport.h>
+#include <llvm/ExecutionEngine/Orc/Debugging/PerfSupportPlugin.h>
 #include <llvm/ExecutionEngine/Orc/JITTargetMachineBuilder.h>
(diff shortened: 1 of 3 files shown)


interpreter/cling/lib/Interpreter/IncrementalJIT.cpp
```diff
@@ -40,7 +44,57 @@ using namespace llvm;
...
+static LLVM_ATTRIBUTE_USED void linkComponents() {
+  errs() << "Linking in runtime functions\n"
+         << (void*)&llvm_orc_registerJITLoaderPerfStart << '\n'
+         << (void*)&llvm_orc_registerJITLoaderPerfEnd << '\n'
+         << (void*)&llvm_orc_registerJITLoaderPerfImpl << '\n';
+}
+
 namespace {
+  // This could potentially be upstreamed, similar to enableDebuggerSupport()
+  Error enablePerfSupport(LLJIT& J) {
```
hahnjo, line 56: The compiler warns that this function is unused on macOS. I think we could just always call it, then the user would get a proper error message when trying to set `CLING_PROFILE`.

hahnjo (approved):
LGTM. It might be good to have some documentation on `CLING_DEBUG` and `CLING_PROFILE` and the new way of using it. I think so far, we only had it in the release notes for 6.28...

vgvassilev (approved):

PR #20194, opened 2025-10-24
[Python][UHI] Start introducing UHI serialization

siliataider:
# This Pull request:

This PR introduces the first steps toward enabling UHI serialization for ROOT histograms (see https://github.com/scikit-hep/uhi/pull/167).

It implements
- `_to_uhi_` for ROOT histograms: provides an Intermediate Representation of the histogram described by the [UHI schema](https://uhi.readthedocs.io/en/latest/serialization.html#full-schema). 
- `_from_uhi_`  classmethod: converts from a UHI histogram to a ROOT histogram.

Minimal reproducer:
```py
import json

import numpy as np
import uhi.io.json

import ROOT

# Create a histogram
h = ROOT.TH1D("h", "h", 10, -5, 5)
h[...] = np.arange(10)

# Serialize to JSON
ir = json.loads(json.dumps(h, default=uhi.io.json.default), object_hook=uhi.io.json.object_hook)

# Deserialize via constructor
h_loaded_ctor = ROOT.TH1D(ir)

# Deserialize via classmethod
h_loaded_clsmethod = ROOT.TH1D._from_uhi_(ir)

assert np.array_equal(h.values(), h_loaded_ctor.values()) and np.array_equal(h.values(), h_loaded_clsmethod.values())
```

### DIFF
--- a/bindings/pyroot/pythonizations/test/uhi_test_suite.py
+++ b/bindings/pyroot/pythonizations/test/uhi_test_suite.py
@@ -39,13 +39,12 @@ class TestIndexing1D(uhi.testing.indexing.Indexing1D[ROOT.TH1D]):
         Return a 1D histogram with 10 bins from 0 to 1. Each bin value increases
         by 2 compared to the previous one, starting at 0. The underflow bin is 3,
         and the overflow bin is 1.
-        Reference base class:
-        https://github.com/scikit-hep/uhi/blob/7d2cc3c169b51b5538e1380e66b892f451c28c1a/src/uhi/testing/indexing.py#L68
+        The histogram is constructed from the UHI schema returned by `get_uhi()`, a utility
+        defined in the reference base class:
+        https://github.com/scikit-hep/uhi/blob/7bf20edac71ba5fa648f6b33ccdd579df7f5e1a0/src/uhi/testing/indexing.py#L77
         """
         with _temporarily_disable_add_directory():
-            h = ROOT.TH1D("h", "h", 10, 0, 1)
-            h[...] = np.array([3.0, 0.0, 2.0, 4.0, 6.0, 8.0, 10.0, 12.0, 14.0, 16.0, 18.0, 1.0])
-            return h
+            return ROOT.TH1D(cls.get_uhi())
 
 
 class TestIndexing2D(uhi.testing.indexing.Indexing2D[ROOT.TH2D]):
@@ -56,15 +55,12 @@ class TestIndexing2D(uhi.testing.indexing.Indexing2D[ROOT.TH2D]):
         """
         Return a 2D histogram with [2,5] bins. The contents are x+2y, where x and y
         are the bin indices.
-        Reference base class:
-        https://github.com/scikit-hep/uhi/blob/7d2cc3c169b51b5538e1380e66b892f451c28c1a/src/uhi/testing/indexing.py#L391
+        The histogram is constructed from the UHI schema returned by `get_uhi()`, a utility
+        defined in the reference base class:
+        https://github.com/scikit-hep/uhi/blob/7bf20edac71ba5fa648f6b33ccdd579df7f5e1a0/src/uhi/testing/indexing.py#L399
         """
         with _temporarily_disable_add_directory():
-            h2 = ROOT.TH2D("h2", "h2", 2, 0, 2, 5, 0, 5)
-            x, y = np.mgrid[0:2, 0:5]
-            data = np.pad(x + 2 * y, 1, mode="constant")
-            h2[...] = data
-            return h2
+            return ROOT.TH2D(cls.get_uhi())
 
 
 class TestIndexing3D(uhi.testing.indexing.Indexing3D[ROOT.TH3D]):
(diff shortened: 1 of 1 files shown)


bindings/pyroot/pythonizations/python/ROOT/_pythonization/_th1.py
```diff
@@ -252,6 +252,11 @@ def _enable_numpy_fill(klass):
         klass._Fill = klass.Fill
         klass.Fill = _FillWithArrayTH1
 
+    # Add serialization features
+    from ROOT._pythonization._uhi import _add_serialization_features
```
vepadulano: This import statement could be moved outside of the loop

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_th1.py
```diff
@@ -265,7 +270,7 @@ def pythonize_th1(klass):
     klass._Original_SetDirectory = klass.SetDirectory
     klass.SetDirectory = _SetDirectory_SetOwnership
 
-    # Add UHI indexing features
+    # Add UHI indexing and serialization features
     _add_indexing_features(klass)
```
vepadulano: The function name does not reflect the addition to the comment

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_th2.py
```diff
@@ -83,6 +83,11 @@ def _FillWithArrayTH2(self, *args):
     # Add UHI plotting features
     pythonization(klass)(_add_plotting_features)
 
+    # Add serialization features
+    from ROOT._pythonization._uhi import _add_serialization_features
```
vepadulano: This import statement could be moved outside of the loop

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_th3.py
```diff
@@ -31,3 +31,8 @@
 
     # Add UHI plotting features
     pythonization(klass)(_add_plotting_features)
+
+    # Add serialization features
+    from ROOT._pythonization._uhi import _add_serialization_features
```
vepadulano: This import statement could be moved outside of the loop

bindings/pyroot/pythonizations/test/uhi_test_suite.py
```diff
@@ -43,9 +43,7 @@ def make_histogram(cls) -> ROOT.TH1D:
         https://github.com/scikit-hep/uhi/blob/7d2cc3c169b51b5538e1380e66b892f451c28c1a/src/uhi/testing/indexing.py#L68
         """
         with _temporarily_disable_add_directory():
-            h = ROOT.TH1D("h", "h", 10, 0, 1)
-            h[...] = np.array([3.0, 0.0, 2.0, 4.0, 6.0, 8.0, 10.0, 12.0, 14.0, 16.0, 18.0, 1.0])
-            return h
+            return ROOT.TH1D(cls.get_uhi())
```
vepadulano, line 47: Sorry, I can't see where the `get_uhi` method is defined

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_uhi/serialization.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    hist_values = storage_dict["values"]
+    hist[...] = hist_values
+
+
+def _to_uhi_(self) -> dict[str, Any]:
+    from .tags import _get_axis
+
+    return {
+        "uhi_schema": 1,
```
vepadulano, line 149: Is this mandated by the UHI, i.e. this says we are conforming to the UHI schema version 1?

siliataider: Yes, https://uhi.readthedocs.io/en/latest/serialization.html

> There is one more required top-level key: "uhi_schema", which must be set to 1 currently. If there is a future revision with a backward incompatible change, this will be bumped to 2

bindings/pyroot/pythonizations/test/uhi_test_suite.py
siliataider: It's a utility [defined in the parent test class](https://github.com/scikit-hep/uhi/blob/7bf20edac71ba5fa648f6b33ccdd579df7f5e1a0/src/uhi/testing/indexing.py#L77), I'll link them thanks

siliataider:
Thanks for the review @vepadulano , and yes docs will come in a later PR

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_uhi/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+Implementation of the serialization component of the UHI
+"""
+
+
+def _TH1_Constructor(self, *args, **kwargs):
+    """
+    If UHI IR is detected, use the UHI deserialization constructor
+    else forward to the original
+    """
+    if len(args) == 1 and isinstance(args[0], dict):
```
vepadulano: Now I understand this bit better: we are checking if the user is trying to construct a histogram from a dictionary, with the assumption that it is compliant with the UHI schema. The feature sounds cool, but I would try to be a bit more conservative in first approximation, e.g. by enforcing that the dictionary at least has the pair `"uhi_schema": 1`. I understand that is not full enforcing, the user could always insert that pair themselves, but then they'd be falling out of the convention.

siliataider: Makes sense, I added some extra checks based on your suggestion [here](https://github.com/siliataider/root/blob/01285c42e9c54f7c87749329e9aded4b3fb75077/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_uhi/serialization.py#L134) and [here](https://github.com/siliataider/root/blob/01285c42e9c54f7c87749329e9aded4b3fb75077/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_uhi/__init__.py#L81) and [tests](https://github.com/siliataider/root/blob/01285c42e9c54f7c87749329e9aded4b3fb75077/bindings/pyroot/pythonizations/test/uhi_serialization_json.py#L82)

bindings/pyroot/pythonizations/test/uhi_serialization_json.py
vepadulano, line 1: I believe this file should also be added as a test in the CMake at `bindings/pyroot/pythonizations/test/CMakeLists.txt`, right?

PR #20204, opened 2025-10-27
[hist] fix bug in margin not being symmetric

ferdymercury:
2nd delta was different since xmin was modified inbetween

follow-up of https://github.com/root-project/root/pull/19605/files

### DIFF
--- a/hist/hist/src/THLimitsFinder.cxx
+++ b/hist/hist/src/THLimitsFinder.cxx
@@ -382,7 +382,7 @@ void THLimitsFinder::OptimizeLimits(Int_t nbins, Int_t &newbins, Double_t &xmin,
       xmin = -1;
       xmax = 1;
    } else {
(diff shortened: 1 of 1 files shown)


hist/hist/src/THLimitsFinder.cxx
```diff
@@ -382,8 +382,9 @@ void THLimitsFinder::OptimizeLimits(Int_t nbins, Int_t &newbins, Double_t &xmin,
       xmin = -1;
       xmax = 1;
    } else {
-      xmin = std::min(binlow, xmin - 0.01 * (xmax - xmin));
-      xmax = std::max(binhigh, xmax + 0.01 * (xmax - xmin));
+      auto delta = 0.01 * (xmax - xmin);
```
hageboeck:
```suggestion
      const auto delta = 0.01 * (xmax - xmin);
```

PR #20209, opened 2025-10-28
[graf2d] allow zero outer margins when dividing canvas

rlalik:
# This Pull request:
Implements feature proposed here: https://github.com/root-project/root/pull/20165#issuecomment-3434969088
## Changes:
* When `nx` or `ny` is negative, the pads is divided according to `abs(nx)` or `abs(ny)`, but the outer margins is set to 0. The inner margins are equal to `xmargin` and `ymargin`, respectively. Demo of the new feature in the bottom row of the image:
image

Created with macro:

```cpp
void can_test(int nx = 3, int ny = 3, int color = 46) {
    /* can_0 to can_6 here, not relevant */
    TCanvas * can_7 = new TCanvas("can_7", "can_7", 200, 664, 200, 200);
    can_7->Divide(-nx, -ny, 0.05, 0.05, color);
    can_7->Draw();

    TCanvas * can_8 = new TCanvas("can_8", "can_8", 408, 664, 200, 200);
    can_8->Divide(-nx, ny, 0.05, 0.05, color);
    can_8->Draw();

    TCanvas * can_9 = new TCanvas("can_9", "can_9", 616, 664, 200, 200);
    can_9->Divide(nx, -ny, 0.05, 0.05, color);
    can_9->Draw();
}
```
## Checklist:

- [x] tested changes locally

Also related to #20138


graf2d/gpad/src/TPad.cxx
```diff
@@ -1269,30 +1272,44 @@ void TPad::Divide(Int_t nx, Int_t ny, Float_t xmargin, Float_t ymargin, Int_t co
...
-   if (ny <= 0) ny = 1;
-   Int_t ix, iy;
-   Double_t x1, y1, x2, y2, dx, dy;
-   TPad *pad;
+   if (nx == 0)
+      nx = 1;
+   if (ny == 0)
+      ny = 1;
+
+   // Outer margins, equal to half of the inner margin. Set to zero in nx or ny are negative.
```
ferdymercury:
```suggestion
   // Outer margins, equal to half of the inner margin. Set to zero if nx or ny are negative.
```

rlalik: Fixed. I also force pushed the commit, made no sense for a such small change.

rlalik:
I force pushed small fix. In the previous version I was generating margins twice smaller than in the original implementation. So in my implementation whenever I use `xmargin` or `ymargin` I must use `2 * x(y)margin` to achieve backward compatibility. Now is fixed. I also updated the image and test macro.

couet (approved):

rlalik:
@couet But I started to have a second thoughts. Unfortunately, my proposed solution, if run on older ROOT version, will fall for the previous behavior:
```cpp
   if (nx <= 0) nx = 1;
   if (ny <= 0) ny = 1;
```
and if someones does `can->cd(3)`, the code will silently crash without explanation.

So perhaps other idea would be one of two:
1. What is the special case for `xmargin` and `ymargin` `< 0`? I asked a couple of time about this but never got answer. Is it really used? It could be used (and it was my initial idea) to use negative values for zero outer margins, if this case isn't really used by anyone - and I don't know why it should be used as it produces odd case.
2. Change signature of `Divide` function to:
```cpp
void TPad::Divide(Int_t nx, Int_t ny, Float_t xmargin, Float_t ymargin, Int_t color, Float_t xmargin_outer = 999.0, Float_t ymargin_outer = 999.0)
```
or make an overload:
```cpp
void TPad::Divide(Int_t nx, Int_t ny, Float_t xmargin, Float_t ymargin, Int_t color, Float_t xmargin_outer, Float_t ymargin_outer)
```
This is more flexible as allows to adjust the outer margin - may not be 0 but may also be smaller than inner margins. If value is 1 or more (for the case with defaults), then inner `x(y)margins` are used as the outer margin. This will also not work with the older ROOT but at least will give compilation/execution error which user can lookup in the documentation. But this is ABI breaking change.

couet:
I do not think a negative marging has some meaning. It is not allowed. In SetMargin one can see this protection:
```
if ([margin < 0 || margin >1) margin = 0.1;
```

rlalik:
> I do not think a negative marging has some meaning. It is not allowed. In SetMargin one can see this protection:
> 
> ```
> if ([margin < 0 || margin >1) margin = 0.1;
> ```

But documentation says something else. Please see this comment: https://github.com/root-project/root/issues/20138#issuecomment-3453402963 and see the image which is just above. The lower row of canavses has odd behavior - non-zero top-right margins. I reported that in my first issue. And in the code the top-right margin is somehow treated in a special way. So perhaps this special case should be removed? Perhaps it is some very obsolete code and not relevant today?

couet:
Ok I need to check....

couet:
The margins <0 means there is now space between the pads. May be there is somthing wrong in that case. Does your PR fixes it ? That's very old code  which was not change since 21 years. May be it can be improved as you suggest.

rlalik:
I tried to not modify this behavior as it was explicitly documented, so assumed that it is important to keep it as it is. So I changed from initial idea of using negative margins to negative divisions numbers. But if you say, I can get back to initial idea of negative margins, and remove the this old obsolete code.

couet:
Yes negative margins means no spaces betwen the pads. This code was introduced by this PR: https://github.com/root-project/root/commit/e2cbbfc8f6a9aac832ba111ef9b9e07075cff01a

it says: 

_Extend the functionality of TPad::Divide when xmargin=0 and ymargin=0.
In this case the pad is divided with no space between the subpads.
The size of each useful frame inside each pad is identical.
eg, try the following script:_
```
void divpad(Int_t nx=3, Int_t ny=2) {
   TCanvas *c1 = new TCanvas("c1");
   c1->Divide(nx,ny,0,0);
   Int_t number = 0;
   TH1F *h = new TH1F("h","h",100,-3.3,3.3);
   h->GetXaxis()->SetLabelFont(53);
   h->GetXaxis()->SetLabelSize(10);
   h->GetYaxis()->SetLabelFont(53);
   h->GetYaxis()->SetLabelSize(10);
   h->SetMaximum(30*nx*ny);
   h->SetFillColor(42);
   for (Int_t i=0;i<nx*ny;i++) {
      number++;
      c1->cd(number);
      h->FillRandom("gaus",1000);
      h->DrawCopy();
   }
}
```
When you run this example it looks fine.

rlalik:
Margins set to 0 are fine. But margins set to negative number give the odd top-right non-negative margin, and left-bottom zero margin. See the middle row of the image on top of the page.

couet:
In the example above (given in the original PR) if i put negative margins like:
```
c1->Divide(nx,ny,.1,-1);
```
I get exactly the same plot as the original with margins = 0.

rlalik:
I took your example and made only this change: `c1->Divide(nx,ny,1,-1, 38);` to add color to subpad and set negative margin:
```cpp
void divpad(Int_t nx=3, Int_t ny=2) {
   TCanvas *c1 = new TCanvas("c1");
   c1->Divide(nx,ny, 1, -1, 38);
   Int_t number = 0;
   TH1F *h = new TH1F("h","h",100,-3.3,3.3);
   h->GetXaxis()->SetLabelFont(53);
   h->GetXaxis()->SetLabelSize(10);
   h->GetYaxis()->SetLabelFont(53);
   h->GetYaxis()->SetLabelSize(10);
   h->SetMaximum(30*nx*ny);
   h->SetFillColor(42);
   for (Int_t i=0;i<nx*ny;i++) {
      number++;
      c1->cd(number);
      h->FillRandom("gaus",1000);
      h->DrawCopy();
   }
}
```
The result is following:
c1
Can you past image generated by this exact figure on your system? You are plotting white backgrounds so you may not be able to notice the odd margins.

ferdymercury:
> the odd margins

I guess it leaves as much margin on the right part as the space on the left for the y axis labels?  Same for top vs bottom x labels. So that the axes lines are centered.

couet:
This is done on purpose. This divide allows to have the exact save drawing area for all the pads. The right and top margins (green in your case) compensate the left and bottom margins in which the  axis are drawn. That's what _"The current pad margins are recomputed to optimize the layout"_ means. This avoid the effect you get with 0 margins in which the drawing area are not all the same:


```
void divpad(Int_t nx=3, Int_t ny=2) {
   TCanvas *c1 = new TCanvas("c1");
   c1->Divide(nx,ny,0.00001,0.00001);
   Int_t number = 0;
   TH1F *h = new TH1F("h","h",100,-3.3,3.3);
   h->GetXaxis()->SetLabelFont(53);
   h->GetXaxis()->SetLabelSize(10);
   h->GetYaxis()->SetLabelFont(53);
   h->GetYaxis()->SetLabelSize(10);
   h->SetMaximum(30*nx*ny);
   h->SetFillColor(42);
   for (Int_t i=0;i<nx*ny;i++) {
      number++;
      auto p = c1->cd(number);
      if (i!= 0 && i!=3) p->SetLeftMargin(0.0);
      p->SetRightMargin(0.0);
      if (i<3) p->SetBottomMargin(0.0);
      p->SetTopMargin(0.0);
      h->FillRandom("gaus",1000);
      h->DrawCopy();
   }
}
```
Screenshot 2025-10-28 at 16 24 15

This divide (with negative margins) allows to get easily common axis for all pads making sure the drawing area has the same size in all pads.

rlalik:
Ok, after this explanation it is much more clearer for me why this mode is introduced, so also it shouldn't be touched. I don't know, maybe it is only me, but perhaps the documentation could be slightly reworded to made it clearer. But as I just say, maybe it's only me who didn't get the idea.

But that means I have no good way to implement my idea of outer margins. The only idea left for me for now is to:
1. Change signature of Divide function to:
```cpp
void TPad::Divide(Int_t nx, Int_t ny, Float_t xmargin, Float_t ymargin, Int_t color, Float_t xmargin_outer = 999.0, Float_t ymargin_outer = 999.0)
```
2. or make an overload:
```cpp
void TPad::Divide(Int_t nx, Int_t ny, Float_t xmargin, Float_t ymargin, Int_t color, Float_t xmargin_outer, Float_t ymargin_outer)
```

couet:
Yes a better explanation will be useful. 
Regarding the outer margins is it not already there via the canvas margins ?
```
void divpad(Int_t nx=3, Int_t ny=2) {
   TCanvas *c1 = new TCanvas("c1");
   c1->SetLeftMargin(0.3);
   c1->SetRightMargin(0.3);
   c1->SetTopMargin(0.3);
   c1->SetBottomMargin(0.3);
   c1->Divide(nx,ny,-1);
   Int_t number = 0;
   TH1F *h = new TH1F("h","h",100,-3.3,3.3);
   h->GetXaxis()->SetLabelFont(53);
   h->GetXaxis()->SetLabelSize(10);
   h->GetYaxis()->SetLabelFont(53);
   h->GetYaxis()->SetLabelSize(10);
   h->SetMaximum(30*nx*ny);
   h->SetFillColor(42);
   for (Int_t i=0;i<nx*ny;i++) {
      number++;
      c1->cd(number);
      h->FillRandom("gaus",1000);
      h->DrawCopy();
   }
}
```
Also [this example ](https://root.cern/doc/master/canvas2_8C.html)is may be useful.

rlalik:
No, this example works only if you use the optimized mode. It will not work if both margins are non-negative. As it was in the old code (before my PR):
```cpp
      dy = 1/Double_t(ny);
      dx = 1/Double_t(nx);
      for (iy=0;iy<ny;iy++) {
         y2 = 1 - iy*dy - ymargin;
         y1 = y2 - dy + 2*ymargin;
```
so this line ignores the top canvas margins. It could be changed to work with top margins and in my opinion it would be most elegant, but will break existing macros. But with proper documentation, could we live with this backward compatibility breakage?

rlalik:
So, that would be my proposal to have TPad::Divide which respects the top-level canvas margins. The pros is better control of the layout; cons, is the broken of existing macros.
 If the old code was:
```cpp
c->Divide(3, 2, 0.02, 0.01);
```
the new code must be:
```cpp
c->SetLeftMargin(0.02); // old xmargin
c->SetBottomMargin(0.01); // old ymargin
c->SetRightMargin(0.02); // old xmargin
c->SetTopMargin(0.01); // old ymargin
c->Divide(3, 2, 0.04, 0.02); // two times old x,y margins
```
to achieve the same result.

This is how the pad would look with some custom:
canvas_divide_custom
and this if default are used:
canvas_divide_default
and this if one used code to reproduce old defaults:
canvas_divide_old_default

rlalik:
I also updated the documentation. In case that this PR would not be accepted, one has to update the doc of `TPad::Divide`, since this is incorrect:
```cpp
///  - xmargin is the space along x between pads in percent of canvas.
///  - ymargin is the space along y between pads in percent of canvas.
```
This is more correct (could be add to #20260):
```cpp
///  - xmargin is the space around each pad along x, distance between pads along x is equal to 2*xmargin.
///  - ymargin is the space around each pad along y, distance between pads along y is equal to 2*ymargin.
```
With this PR, the old documentation is correct as `xmargin` and `ymargin` are now distances between pads, so no need for this change.

rlalik:
Here are a couple of more scripts to test the effect of this PR:
```cpp
/*
 * Make canvas:
 * @param name canvas name/title
 * @param sx, sy windows screen position x, y
 * @param wx, wy window size
 * @param nx, ny canvas division
 * @param xm, ym canvas inner margins
 * @param ml, mb, mr, bt top-level canvas margins (outer margis) left, bottom,
 * right, top
 */
void make_canvas(const char *name, int sx, int sy, int wx, int wy, int nx, int ny, float xm, float ym, float ml,
                 float mb, float mr, float mt)
{
   auto c = new TCanvas(name, name, sx, sy, wx, wy);
   // c->SetFillColor(38);

   c->SetLeftMargin(ml);
   c->SetBottomMargin(mb);
   c->SetRightMargin(mr);
   c->SetTopMargin(mt);

   c->Divide(nx, ny, xm, ym, 46);

   auto h = new TH1F("", "", 100, -3.3, 3.3);
   h->GetXaxis()->SetLabelFont(43);
   h->GetXaxis()->SetLabelSize(12);
   h->GetYaxis()->SetLabelFont(43);
   h->GetYaxis()->SetLabelSize(12);
   h->GetYaxis()->SetNdivisions(505);
   h->SetMaximum(30 * nx * ny);
   h->SetFillColor(41);

   Int_t number = 0;
   for (Int_t i = 0; i < nx * ny; i++) {
      number++;
      c->cd(number);
      h->FillRandom("gaus", 1000);
      h->DrawCopy();
   }
}

void can_test2()
{
   auto wx = 200; // width and heigh
   auto wy = 200;
   auto sx = 200; // initial canvas position
   auto sy = 200;
   auto dx = 208; // canvas next screen x position
   auto dy = 232;

   auto nx = 3;
   auto ny = 2;

   make_canvas("can_0_0", sx + dx * 0, sy + dy * 0, wx, wy, nx, ny, 0.05, 0.05, 0.1, 0.1, 0.1, 0.1);
   make_canvas("can_0_1", sx + dx * 1, sy + dy * 0, wx, wy, nx, ny, 0.01, 0.01, 0.1, 0.1, 0.1, 0.1);
   make_canvas("can_0_2", sx + dx * 2, sy + dy * 0, wx, wy, nx, ny, 0.00, 0.00, 0.1, 0.1, 0.1, 0.1);

   make_canvas("can_1_0", sx + dx * 0, sy + dy * 1, wx, wy, nx, ny, 0.05, 0.05, 0, 0, 0, 0);
   make_canvas("can_1_1", sx + dx * 1, sy + dy * 1, wx, wy, nx, ny, 0.01, 0.01, 0, 0, 0, 0);
   make_canvas("can_1_2", sx + dx * 2, sy + dy * 1, wx, wy, nx, ny, 0.00, 0.00, 0, 0, 0, 0);

   make_canvas("can_2_0", sx + dx * 0, sy + dy * 2, wx, wy, nx, ny, 0.05, 0.05, 0.05, 0.05, 0, 0);
   make_canvas("can_2_1", sx + dx * 1, sy + dy * 2, wx, wy, nx, ny, 0.01, 0.01, 0, 0.05, 0, 0.05);
   make_canvas("can_2_2", sx + dx * 2, sy + dy * 2, wx, wy, nx, ny, 0.00, 0.00, 0.05, 0, 0.05, 0);
}
```
image

couet:
I modified my PR regarding the documention is it ok for you If I merge it ? https://github.com/root-project/root/pull/20260
Regarding your new version of this PR I need to look more closely.

couet:
It looks like you need to rebase/take the mods I did in the help.

rlalik:
> It looks like you need to rebase/take the mods I did in the help.

I did rebase after your PR being merged.

graf2d/gpad/src/TPad.cxx
```diff
@@ -1225,13 +1225,12 @@ Int_t TPad::DistancetoPrimitive(Int_t px, Int_t py)
...
-///  - ymargin defines the vertical spacing around each pad as a percentage of the canvas
-///    height. Therefore, the distance between two adjacent pads along the y-axis is equal
-///    to twice the ymargin value.
+///  - xmargin defines the horizontal spacing between each pad as a percentage of the canvas
+///    width.
+///  - ymargin defines the vertical spacing between each pad as a percentage of the canvas
+///    height.
 ///  - color is the color of the new pads. If 0, color is the canvas color.
+///  - All pads are contained within the inner area defined by the canvas margins.
 ///
```
ferdymercury, line 1234:
```suggestion
/// - Note that, if you don't have a background color of your pad, the spacing between pads might look larger than specified,
/// since in the default case, each pad has internally an empty space on the right equal to the space filled on the left for the y axis labels 
```

Maybe sth along these lines would be helpful, since most people don't put background color on their pads.

tutorials/visualisation/graphics/canvas_divide_example.C
```diff
@@ -0,0 +1,118 @@
+// variants: 0  - custom values
+//           1  - default values
+//           2+ - old default values
```
ferdymercury: Maybe specify that old refers to anything before 6.40

ferdymercury (approved):

tutorials/visualisation/graphics/canvas_divide_example.C
rlalik: No chance of having this in 6.38.0?

ferdymercury: Not sure, that for couet to decide. I think 6.38.0 is in feature-freeze status.

rlalik: I enriched the example description (not committed yet):
```
/// ROOT 6.40.0 changed how the canvas are divided into subpads. Before, the TPad::Divide
/// with typical arguments (positive `xmargin` and `ymargin` values) function was not respecting
/// canvas own margins for the inner area. This limited flexibility of defining canvas layout.
///
/// As it was changed and fixed in 6.40.0, this macro demonstrates how the new TPad::Divide function
/// works, what are the current default values, and how to reproduce the old behaviour.
///
/// This example can be run with optional argument (default is 0):
///  * 0  - will demonstrate current custom layout behaviour,
///  * 1  - will show default values (starting from 6.40.0) where the canvas margins are respected
///         and the subpad canvas are customised (the default values are put explicitly because we
///         also want to modify the pads background colour for better visibility of the layout),
///  * 2 (or anything else than 0, 1) - will restore old default values (for root before 6.40.0).
///      Note that the `xmargin` and `ymargin` are doubled in TPad::Divide call in respect to the
///      old defaults, and the canvas margins also must be modified.
/// ~~~{.cpp}
/// const auto xmargin = 0.01; // old default xmargin
/// const auto ymargin = 0.01; // old default ymargin
/// c->SetMargin(xmargin, xmargin, ymargin, ymargin);
/// c->Divide(nx, ny, 2 * xmargin, 2 * ymargin, 46);
/// ~~~
```
Let me know if you would like to change something in it before I commit.

graf2d/gpad/src/TPad.cxx
```diff
@@ -1225,13 +1225,16 @@ Int_t TPad::DistancetoPrimitive(Int_t px, Int_t py)
 /// Automatic pad generation by division.
 ///
 ///  - The current canvas is divided in nx by ny equal divisions (pads).
-///  - xmargin defines the horizontal spacing around each pad as a percentage of the canvas
-///    width. Therefore, the distance between two adjacent pads along the x-axis is equal
-///    to twice the xmargin value.
-///  - ymargin defines the vertical spacing around each pad as a percentage of the canvas
-///    height. Therefore, the distance between two adjacent pads along the y-axis is equal
-///    to twice the ymargin value.
+///  - xmargin defines the horizontal spacing between each pad as a percentage of the canvas
```
couet, line 1228: This is overwriting the changed I did in the help. It goes back to the old help.

couet (commented):
the PR is going back it overwrites the changes I did

graf2d/gpad/src/TPad.cxx
ferdymercury: I think he is proposing to change behavior in 6.40, and the new behavior matches by chance the old help message

rlalik: Yes, I already wrote that here https://github.com/root-project/root/pull/20260#issuecomment-3481044532 that f this PR is accepted, the old doc will stay valid and no need to change the x/y margin definition. But I think, in the end, the documentation become better, since I reworded your documentation to have it better in this PR.

tutorials/visualisation/graphics/canvas_divide_example.C
```diff
@@ -0,0 +1,118 @@
...
+void canvas_divide_example(int use_variant = 0)
```
couet, line 32: Look at the other examples: they all have a standard header. You should add a header

rlalik: I had it already but not commited yet, but now it is in. You can have a look on that.

linev:
I re-open PR to start CI again.

ferdymercury (commented):
I'd suggest mentioning this also in the ReleaseNotes

linev:
I see very long discussion - and code which is no longer corresponds to original description of the PR.
And there are many failing tests.

In general I like idea to respect margins when divide pad, 
but at the same time I want to preserve default behavior of `TPad::Divide()` method.

My proposed solution is following.

Meaning of `nx` and `ny` parameters is not changed. 

That will be changed - `xmargin` and `ymargin` can be positive or negative. 

If both positive (` >= 0`) - `TPad::Divide()` works as before.

When `xmargin`  negative - pad division inside configured X-margins is performed, as a gap absolute value of `xmargin` will be used. Same for `ymargin`.

So one can have flexibility to configure independently boundary around subpad (via margins) and gaps between subpads. I guess, it was main point of Rafal.

@rlalik Are you willing to redo your PR or should I made it.

rlalik:
@linev from the discussion in thes PR, specifically https://github.com/root-project/root/pull/20209#issuecomment-3456907403 and following (and also before), negative values of x or y margin have special meaning of adjusting right margin to match left margin, it is in the `TPad::Divide` implementation:
```c++
   if (xmargin >= 0 && ymargin >= 0) {
      //general case
      // ... code here
   } else {
      // special case when xmargin < 0 or ymargin < 0
      // ... code here
   }
  ```
So not sure if we should change this. My proposed behavior with negative nx/ny do not modify old behavior but introduces new feature for values before not used (or at least not documented).

linev:
1. I did not found place where negative `nx` or `ny` are handled in your code.
2. When `(xmargin >= 0 && ymargin >= 0)` code should remain exactly the same.
3.  I want to handle X and Y direction independently. When `xmargin < 0`, different approach applied for X axis, but Y remain the same.

PR #20211, opened 2025-10-28
[net] TParallelMergingFile: don't rely on nbytes for UploadAndReset()

silverweed:
The `nbytes` value returned by TDirectoryFile::Write() is only > 0 if we write at least one TObject to the file, which is not the case for non-TObject-based objects that handle their own writing (e.g. RNTuple through RMiniFile).
This means that if we write only non-TObjects in a TParallelMergingFile we will never call UploadAndReset() by only checking nbytes. With this PR, we instead use GetBytesWritten() to check if anything gets written by TDirectoryFile::Write(), which works even in case of non- TObjects.

cc @amete - this should fix one of the issues you observed with your parallel file merger when you only merge RNTuples.

### DIFF
--- a/net/net/test/testParallelMergingFile.cxx
+++ b/net/net/test/testParallelMergingFile.cxx
@@ -67,7 +67,7 @@ static void Server(std::unique_ptr<TServerSocket> ss, const std::string &outFile
          msg->ReadTString(filename);
          msg->ReadLong64(length);
 
-         // XXX: this lock is here to work around https://github.com/root-project/root/issues/20641
+         // NOTE: this lock is here to work around https://github.com/root-project/root/issues/20641
          // Note that we don't care about minimizing its scope since we are not testing multithreaded scaling here
          // and it's fine if the client's and server's operations get serialized by the mutexes.
          std::scoped_lock<std::mutex> lock(gMutex);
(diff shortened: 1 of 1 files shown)


net/net/test/testParallelMergingFile.cxx
```diff
@@ -0,0 +1,181 @@
...
+         // Receive and merge data
+         Int_t clientId;
+         TString filename;
+         Long64_t length;
+         msg->ReadInt(clientId);
+         msg->ReadTString(filename);
+         msg->ReadLong64(length);
+
+         // XXX: this lock is here to work around https://github.com/root-project/root/issues/20641
+         std::scoped_lock<std::mutex> lock(gMutex);
```
pcanal, line 73: We should note in a comment if the scope is already minimal or if it does not matter (eg. we are not testing for maximal scaling).

net/net/test/testParallelMergingFile.cxx
```diff
@@ -0,0 +1,183 @@
...
+      default: std::cout << "ignoring message of kind " << msg->What() << "\n";
+      }
+
+      delete msg;
+   }
+}
+
+TEST(TParallelMergingFile, UploadAndResetNonTObject)
+{
+   constexpr auto sockPath = "/tmp/parallelMergeTest.sock";
```
jblomer: I think I'd prefer a `TempFileName()` constructed path rather than a fully hardcoded one.

net/net/test/testParallelMergingFile.cxx
```diff
@@ -0,0 +1,183 @@
...
+      auto px = entry->GetPtr<float>("px");
+      auto py = entry->GetPtr<float>("py");
+      int nSent = 0;
+      for (int i = 0; i < kNEntries; ++i) {
+         *px = i;
+         *py = 2 * i;
+         writer->Fill(*entry);
+         if (i > 0 && (i % (kNEntries / kNUploads)) == 0) {
+            {
+               // XXX: this lock is here to work around https://github.com/root-project/root/issues/20641
```
jblomer, line 147: Why `XXX`?

net/net/test/testParallelMergingFile.cxx
```diff
@@ -0,0 +1,183 @@
...
+            py = entry->GetPtr<float>("py");
+         }
+      }
+   }
+
+   delete file;
+
+   serverThread.join();
+
+   {
```
jblomer: This scope seems unnecessary.

net/net/test/testParallelMergingFile.cxx
silverweed: Just a way to signal that it's something "out of place" that ideally shouldn't be there. I can remove it if you prefer

net/net/test/testParallelMergingFile.cxx
```diff
@@ -0,0 +1,183 @@
...
+      case kMESS_ANY: {
+         // Receive and merge data
+         Int_t clientId;
+         TString filename;
+         Long64_t length;
+         msg->ReadInt(clientId);
+         msg->ReadTString(filename);
+         msg->ReadLong64(length);
+
+         // XXX: this lock is here to work around https://github.com/root-project/root/issues/20641
```
pcanal:
```suggestion
         // NOTE: this lock is here to work around https://github.com/root-project/root/issues/20641
```

net/net/test/testParallelMergingFile.cxx
pcanal:
```suggestion
               // NOTE: this lock is here to work around https://github.com/root-project/root/issues/20641
```

silverweed: Oh, I fixed one and forgot the other...

PR #20201, opened 2025-10-26
[hist] raise error when axis edges are inf/nan

ferdymercury:
This is related to https://github.com/root-project/root/issues/20174 and the discussion in https://github.com/root-project/root/pull/20176

fyi @aapo-kossi

Should we completely disallow infinite bin edges, since it violates a bit the assumptions of under and overflow bins? (As well as complicates the implementation of GetBinCenter, Low and Up Edge, etc). And the workaround is quite easy, just setting 1e30 / 1-e30 as variable bin edge.

Or should we only allow inf for variable bin edges, and not for fixed binwidth since there it does not make sense?

Closes https://github.com/root-project/root/issues/20174

PR #20217, opened 2025-10-28
[hist] Don't use FindBin/GetBinCenter in TH1 derived projections

### LINKED ISSUE #20174: TH1 and derivates projections produce incorrect results for histograms with infinite bin edges

aapo-kossi:
Since outgoing bin indices can be computed depending only on the option to keep original binning and set axis ranges, there is no need to use FindBin and GetBinCenter to get the output bin where integrated contents are to be placed. This PR removes the usage of GetBinCenter and FindBin from the output bin index calculation, instead using simple index arithmetic. This fixes the corner-case where non-finite bin edges cause projected bins to be empty due to GetBinCenter returning a bin index that is not actually inside the bin. A slight performance improvement could also be expected, but this is unimportant since I don't expect a histogram projection to be a bottleneck in anyone's analysis.

I tried adding more tests to verify that the new logic is sound and equivalent to the existing implementation in cases other than the mentioned infinite bin edges. I think someone more familiar with the internals should make sure I have covered all use cases so that this change doesn't introduce any new bugs.

Fixes https://github.com/root-project/root/issues/20174

## Checklist:

- [x] tested changes locally

This PR fixes #20174

### DIFF
--- a/hist/hist/src/TH2.cxx
+++ b/hist/hist/src/TH2.cxx
@@ -2272,10 +2272,25 @@ TH1D *TH2::DoProjection(bool onX, const char *name, Int_t firstbin, Int_t lastbi
    // if the out axis has labels and is extendable, temporary make it non-extendable to avoid adding extra bins
    Bool_t extendable = outAxis->CanExtend();
    if ( labels && extendable ) h1->GetXaxis()->SetCanExtend(kFALSE);
-   for ( Int_t outbin = 0; outbin <= outAxis->GetNbins() + 1;  ++outbin) {
+
+   // compute loop indices
+   Int_t loop_start = 0;
+   Int_t loop_end = outAxis->GetNbins() + 1;
+   Int_t binOut = 0;
+   if (outAxis->TestBit(TAxis::kAxisRange)) {
+      loop_start = firstOutBin;
+      loop_end = lastOutBin;
+      if (originalRange)
+         // place results at original bin indices
+         binOut = firstOutBin;
+      else
+         // start from the first bin
+         binOut = 1;
+   }
+
+   for (Int_t outbin = loop_start; outbin <= loop_end; ++outbin, ++binOut) {
       err2 = 0;
       cont = 0;
-      if (outAxis->TestBit(TAxis::kAxisRange) && ( outbin < firstOutBin || outbin > lastOutBin )) continue;
 
       for (Int_t inbin = firstbin ; inbin <= lastbin ; ++inbin) {
          Int_t binx, biny;
@@ -2292,9 +2307,8 @@ TH1D *TH2::DoProjection(bool onX, const char *name, Int_t firstbin, Int_t lastbi
             err2  += exy*exy;
          }
       }
-      // find corresponding bin number in h1 for outbin
-      Int_t binOut = h1->GetXaxis()->FindBin( outAxis->GetBinCenter(outbin) );
-      h1->SetBinContent(binOut ,cont);
+
+      h1->SetBinContent(binOut, cont);
       if (computeErrors) h1->SetBinError(binOut,TMath::Sqrt(err2));
       // sum  all content
       totcont += cont;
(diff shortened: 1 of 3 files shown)


aapo-kossi:
This should be viewed as an alternative approach to #20176, also related to #20201. If/while infinite bin edges are fully disallowed then this is mostly irrelevant and probably unnecessary, but I do not know if that is needed if a fix is this simple.

ferdymercury:
Thanks! Side note: the drawback of supporting this is that one would have also have to fix GetBinCenter, GetBinLowEdge independently of whether it's used or not for the projection), and fixing it might affect performance.
Likewise, the definition of the overflow and underflow bin lose meaning when the first bin is already -inf

aapo-kossi:
> Thanks! Side note: the drawback of supporting this is that one would have also have to fix GetBinCenter, GetBinLowEdge independently of whether it's used or not for the projection), and fixing it might affect performance. Likewise, the definition of the overflow and underflow bin lose meaning when the first bin is already -inf

I agree with the sentiment that those methods could ideally return something reasonable, but since the returned values will either be infinite or NaN, they shouldn't be relied upon for any calculation other than possibly checking that they are not finite. The current implementation seemingly can already do this, though with some inconsistency on if the return value is inf or NaN.

To be more concrete, I don't expect a bin center especially to be very meaningful if we are talking about a bin with infinite width. Having a finite width for the bin in question could just be a precondition for using that method in my opinion, for example. And for getting the bin edges, I don't think uniformly binned axes should need to accept infinities.

A flow bin mostly does become meaningless for a histogram extending to infinity, yes (except for filling at +inf itself).

All this to say that there may be other problems with accepting infs that may need to be considered together as a whole, but in isolation this PR doesn't actually do any handling of infinities. It does fix an issue that would prevent them from being supported and in my opinion it simplifies the projection operations by removing unnecessary checking of axis coordinates (given I haven't missed a case that actually makes it necessary).

hageboeck:
Hello @aapo-kossi, thanks for the initiative! ROOT is currently branching off the v6.38 release, so we'll need a bit of time to decide how to go on.

aapo-kossi:
Closed since inf edges now throw and I don't see sense in changing the approach before RHist is adopted.

PR #20222, opened 2025-10-29
Fixes for TTree with complex friend hierarchies

vepadulano:
This PR collects the fixes and correlated testing in the latest campaign to fix the CMS DAS MiniAOD analysis framework, the only MiniAOD analysis system supported by the CMS Common Analysis Tools group (see https://indico.cern.ch/event/1330797/papers/5796853/files/13266-CAT_ACAT24_proceeding.pdf)

### DIFF
--- a/tree/treeplayer/test/regressions.cxx
+++ b/tree/treeplayer/test/regressions.cxx
@@ -488,3 +488,82 @@ TEST(TTreeScan, chainNameWithDifferentTreeName)
          throw std::runtime_error("Could not retrieve TTreePlayer from main tree!");
    }
 }
+
+// https://github.com/root-project/root/issues/20249
+TEST(TTreeScan, TTreeGetBranchOfFriendTChain)
+{
+   struct DatasetRAII {
+      const char *fTreeNameStepZero{"tree_20249_zero"};
+      const char *fFileNameStepZero{"tree_20249_zero.root"};
+      const char *fTreeNameStepOne{"tree_20249_one"};
+      const char *fFileNameStepOne{"tree_20249_one.root"};
+
+      void WriteData(const char *name, const char *treename, int first, int last)
+      {
+         auto file = std::make_unique<TFile>(name, "RECREATE");
+         auto tree = std::make_unique<TTree>(treename, treename);
+
+         int value{};
+         tree->Branch("value", &value);
+
+         for (value = first; value < last; ++value) {
+            tree->Fill();
+         }
+
+         file->Write();
+      }
+
+      DatasetRAII()
+      {
+         WriteData(fFileNameStepZero, fTreeNameStepZero, 3, 7);
+         WriteData(fFileNameStepOne, fTreeNameStepOne, 0, 4);
+      }
+
+      ~DatasetRAII()
+      {
+         std::remove(fFileNameStepZero);
+         std::remove(fFileNameStepOne);
+      }
+   } dataset;
+
+   auto chain0 = std::make_unique<TChain>("stepzerochain");
+   chain0->Add((std::string(dataset.fFileNameStepZero) + "?#" + dataset.fTreeNameStepZero).c_str());
+
+   auto file1 = std::make_unique<TFile>(dataset.fFileNameStepOne);
+   std::unique_ptr<TTree> tree1{file1->Get<TTree>(dataset.fTreeNameStepOne)};
+   tree1->AddFriend(chain0.get());
+
+   ASSERT_NE(tree1->FindBranch("stepzerochain.value"), nullptr);
+
+   std::ostringstream strCout;
+   {
+      if (auto *treePlayer = static_cast<TTreePlayer *>(tree1->GetPlayer())) {
+         struct FileRAII {
+            const char *fPath;
+            FileRAII(const char *name) : fPath(name) {}
+            ~FileRAII() { std::remove(fPath); }
+         } redirectFile{"tree_20249_regression_redirect.txt"};
+         treePlayer->SetScanRedirect(true);
+         treePlayer->SetScanFileName(redirectFile.fPath);
+
+         tree1->Scan("value:stepzerochain.value", "", "colsize=24");
+
+         std::ifstream redirectStream(redirectFile.fPath);
+         std::stringstream redirectOutput;
+         redirectOutput << redirectStream.rdbuf();
+
+         const static std::string expectedScanOut{
+            R"Scan(******************************************************************
+*    Row   *                    value *      stepzerochain.value *
+******************************************************************
+*        0 *                        0 *                        3 *
+*        1 *                        1 *                        4 *
+*        2 *                        2 *                        5 *
+*        3 *                        3 *                        6 *
+******************************************************************
+)Scan"};
+         EXPECT_EQ(redirectOutput.str(), expectedScanOut);
+      } else
+         throw std::runtime_error("Could not retrieve TTreePlayer from main tree!");
+   }
+}


vepadulano:
This PR now
Fixes #20226
Fixes #20228

See the related commits

vepadulano:
> There are more places where fixing the potential integer overflow is advisable, see https://github.com/root-project/root/pull/20227/files, in TTreePlayer, but also in TSpider, TChain

Thanks for pointing these out! I acknowledge there may be more places to fix. For each of those, we need a corresponding reproducer that fails without the fix and works with the fix. For now those places are not interfering immediately with the code changes needed for this PR, so I would in principle tackle them separately. I'm still developing more tests offline which will end up in this PR. If while developing those tests I see that some other integer overflow is the problem, then I will create the corresponding issue+commit in this PR.

> precalculating lastentry is already being used in three other places in the ROOT tree code

Where exactly? I can only see in `TChain::Loop` in a dead code section (preceded by `#if 0`). There is another instance in `tree/treeplayer`, but there the way it's calculated is not trivially applicable to the two int overflow fixes added in this PR.

ferdymercury:
> For each of those, we need a corresponding reproducer that fails without the fix and works with the fix. For now those places are not interfering immediately with the code changes needed for this PR, so I would in principle tackle them separately.

Thanks, got it. An easy failures-reproducer of two of the locations: TTreeplayer lines 237 and 620:

change line 83 of tutorials/io/tree/tree109_friend.C with
`auto TF = T->CopyTree("z<10", 0, TTree::kMaxEntries, 2);`
and you'll get an integer overflow. It affects both locations/functions of TTreePlayer. (One should not need to know how many entries the tree has in total even if he wants to skip a few from the beginning).

(For the other locations I do not have a repro.)

tree/tree/src/TChain.cxx
```diff
@@ -804,12 +805,17 @@ Long64_t TChain::Draw(const char* varexp, const char* selection,
 
 TBranch* TChain::FindBranch(const char* branchname)
 {
+   std::string name{branchname};
+   auto prefix = std::string(GetName()) + '.';
+   if (ROOT::StartsWith(name, prefix)){
```
pcanal: white space
```suggestion
   if (ROOT::StartsWith(name, prefix)) {
```

tree/tree/src/TChain.cxx
```diff
@@ -804,12 +805,17 @@ Long64_t TChain::Draw(const char* varexp, const char* selection,
 
 TBranch* TChain::FindBranch(const char* branchname)
 {
+   std::string name{branchname};
```
pcanal: Can you add a code comment explaining the net/resulting effect of conditionally stripping the chain's name from the branch name?

pcanal: If/since this is the correct behavior for a chain, why is it not the correct behavior for a `TTree` (i.e. did we investigate the consequence of putting that code to be in `TTree::FindBranch` instead of here?

tree/tree/src/TTree.cxx
```diff
@@ -6122,30 +6122,52 @@ const char* TTree::GetFriendAlias(TTree* tree) const
    if (kGetFriendAlias & fFriendLockStatus) {
       return nullptr;
    }
-   if (!fFriends) {
+   TFriendLock lock(const_cast<TTree *>(this), kGetFriendAlias);
```
pcanal, line 6130: Why is the function no longer starting with
```
if (!friend)
	return nullptr;
```
?

pcanal: This is because this could be a chain (and there is no `TChain::GetFriendAlias`, adding it would be another valid implementation choice).  So that early return would now be:
```
if (!friend && GetTree() == this)
	return nullptr;
```

tree/tree/src/TTree.cxx
```diff
@@ -6122,30 +6122,52 @@ const char* TTree::GetFriendAlias(TTree* tree) const
...
-   while ((fe = (TFriendElement*) nextf())) {
-      TTree* t = fe->GetTree();
-      if (t == tree) {
-         return fe->GetName();
-      }
-      // Case of a chain:
-      if (t && t->GetTree() == tree) {
-         return fe->GetName();
+
+   // Then, check if this is a TChain and the current tree has friends
```
pcanal, line 6152: We might  not be able to avoid it but in the usual case, this should be redundant.

If this is a chain and it has friends, it will make its underlying tree have the friends too (so same semantic friends).

The case that is non-redundantly covered here is the case where the underlying TTree of a TChain has its own persistent friends.

If this analysis is correct, we ought to clarify it in the code comment (for example to avoid a future maintainer thinking only of the former case and wanting to remove this code)

tree/tree/src/TChain.cxx
vepadulano: The equivalent in `TTree::FindBranch` is already there https://github.com/root-project/root/blob/4d196eacc4cde31343e5c5705d6e95091d1bef2d/tree/tree/src/TTree.cxx#L4903-L4908 

This is just aligning that logic to TChain as well

tree/tree/src/TTree.cxx
vepadulano: Correct! Yes indeed we could still have that early return

tree/tree/src/TChain.cxx
pcanal:
```suggestion
   // This will allow the branchname to be preceded by the name of this chain.
   // See similar code in TTree::FindBranch
   std::string name{branchname};
```

tree/tree/src/TTree.cxx
vepadulano: I believe this is already non-redundant because of the condition
```
innerListOfFriends && innerListOfFriends != fFriends
```
but I'm also adding a comment to make this explicit

vepadulano:
@pcanal I addressed all your comments. I also added in the last commit an extension to one of the testing suites that I introduced in an earlier commit, now it considers all cases in which a call to TChain::Scan is accessing one of the friend chains with a name of chain that may be different than the name of tree it holds. ATOW, this PR introduces the following amount of tests:

`gh20033.cxx`
```
586: [----------] Global test environment tear-down
586: [==========] 4080 tests from 3 test suites ran. (17421 ms total)
586: [  PASSED  ] 4064 tests.
586: [  SKIPPED ] 16 tests, listed below:
```

`regressions.cxx`
```
589: [----------] 3 tests from TTreeScan (4 ms total)
589: 
589: [----------] 1 test from TTreeDraw (0 ms total)
```

vepadulano:
As discussed privately with @pcanal, I have reorganised the commit history of this PR to have all the code changes at the beginning, and separate every type of new test scenario added in a different commit, appended after the changes. Another PR has been opened just with the test commits to see the failures in isolation --> https://github.com/root-project/root/pull/20263

vepadulano:
Pasting here errors from the tests seen in the other PR.

## Regression tests

```
[ RUN      ] TTreeScan.IntOverflow
File <tree_20226_regression_redirect.txt> created
/github/home/ROOT-CI/src/tree/treeplayer/test/regressions.cxx:396: Failure
  redirectOutput.str()
    Which is: "************************\n*    Row   *       val *\n************************\n*        3 *         3 *\n************************\n"
  expectedScanOut
    Which is: "************************\n*    Row   *       val *\n************************\n*        3 *         3 *\n*        4 *         4 *\n*        5 *         5 *\n*        6 *         6 *\n*        7 *         7 *\n*        8 *         8 *\n*        9 *         9 *\n************************\n"
With diff:
@@ +3,9 @@
 ************************
 *        3 *         3 *
+*        4 *         4 *
+*        5 *         5 *
+*        6 *         6 *
+*        7 *         7 *
+*        8 *         8 *
+*        9 *         9 *
 ************************\n


[  FAILED  ] TTreeScan.IntOverflow (56 ms)
[ RUN      ] TTreeScan.chainNameWithDifferentTreeName
/github/home/ROOT-CI/src/tree/treeplayer/test/regressions.cxx:457: Failure
Expected: (c.FindBranch("differentNameForChain.val")) != (nullptr), actual: NULL vs (nullptr)

[  FAILED  ] TTreeScan.chainNameWithDifferentTreeName (2 ms)
[ RUN      ] TTreeScan.TTreeGetBranchOfFriendTChain
/github/home/ROOT-CI/src/tree/treeplayer/test/regressions.cxx:536: Failure
Expected: (tree1->FindBranch("stepzerochain.value")) != (nullptr), actual: NULL vs (nullptr)

[  FAILED  ] TTreeScan.TTreeGetBranchOfFriendTChain (3 ms)
[----------] 3 tests from TTreeScan (63 ms total)

[----------] 1 test from TTreeDraw
[ RUN      ] TTreeDraw.IntOverflow
/github/home/ROOT-CI/src/tree/treeplayer/test/regressions.cxx:430: Failure
  nEntriesSelected
    Which is: 1
  7
```

## New tests for complex friendship hierarchies scenarios

```
1500/3660 Test  #672: gtest-tree-treeplayer-treeplayer-gh20033 ..........................................................***Failed    0.61 sec
Running main() from /builddir/build/BUILD/googletest-1.14.0/googletest/src/gtest_main.cc
[==========] Running 4080 tests from 3 test suites.
[----------] Global test environment set-up.
[----------] 176 tests from Run/GH20033Regression
[ RUN      ] Run/GH20033Regression.Test/StepOneTTree_StepTwoTTree_StepThreeTTree_StepFourTTree_0
/github/home/ROOT-CI/src/core/testsupport/src/TestSupport.cxx:93: Failure
Failed
Received unexpected diagnostic of severity 4000 at 'TUnixSystem::DispatchSignals' reading 'segmentation violation'.
Suppress those using ROOT/TestSupport.hxx

 Generating stack trace...
 0x00007f5ab1fbab12 in TTreePlayer::Scan(char const*, char const*, char const*, long long, long long) at /github/home/ROOT-CI/src/tree/treeplayer/src/TTreePlayer.cxx:2774 (discriminator 2) from /github/home/ROOT-CI/build/lib/libTreePlayer.so
 0x000000000041c4b6 in GH20033Regression_Test_Test::TestBody() at /github/home/ROOT-CI/src/tree/treeplayer/test/gh20033.cxx:396 from /github/home/ROOT-CI/build/tree/treeplayer/test/treeplayer_gh20033
 0x00007f5ab0bbca02 in <unknown> from /lib64/libgtest.so.1.14.0
 0x00007f5ab0ba7996 in testing::Test::Run() + 0xd6 from /lib64/libgtest.so.1.14.0
 0x00007f5ab0ba7b95 in testing::TestInfo::Run() + 0x1d5 from /lib64/libgtest.so.1.14.0
 0x00007f5ab0ba7d9f in testing::TestSuite::Run() + 0x1bf from /lib64/libgtest.so.1.14.0
 0x00007f5ab0bb41d0 in testing::internal::UnitTestImpl::RunAllTests() + 0x340 from /lib64/libgtest.so.1.14.0
 0x00007f5ab0bb4808 in testing::UnitTest::Run() + 0x138 from /lib64/libgtest.so.1.14.0
 0x00007f5ab1ed5114 in main + 0x44 from /lib64/libgtest_main.so.1.14.0
 0x00007f5ab0122348 in <unknown> from /lib64/libc.so.6
 0x00007f5ab012240b in __libc_start_main at :? from /lib64/libc.so.6
 0x0000000000418205 in _start + 0x25 from /github/home/ROOT-CI/build/tree/treeplayer/test/treeplayer_gh20033
CMake Error at /github/home/ROOT-CI/src/cmake/modules/RootTestDriver.cmake:232 (message):
  error code: 139
```

vepadulano:
Here is a link to the test runs in the other PR https://github.com/root-project/root/actions/runs/18983943721?pr=20263

vepadulano:
The latest force-push is just to introduce a small improvement. In TChain::FindBranch and TChain::FindLeaf we don't need the copy of the std::string_view to ensure the character array is null-terminated. That's because we're only truncating from the beginning of the view (removing the leading `chainName.` part), and not from the end. The input `const char *` into the function is null-terminated already.

PR #20223, opened 2025-10-29
[win] Skip unhandled exceptions on Windows

bellenot:


io/io/test/rfile.py
```diff
@@ -1,4 +1,5 @@
 import os
+import platform
 import unittest
 import ROOT
```
enirolf, line 5: I believe adding an additional newline fixes the `ruff` error:
```suggestion
import os
import platform
import unittest

import ROOT
```

bellenot: Thanks Florine!

Issue #20208, opened 2025-10-27
Flag to disable cling's `NVPTX` target


LukasBreitwieser:
### Explain what you would like to see improved and how.

In order to reduce binary size and compile time, @ktf asked for a way to disable the LLVM `NVPTX` build target. 
Currently, NVPTX is added in all cases, even if  ROOT is built with `-Dcuda=off -Dcudnn=off`:
[see here](https://github.com/root-project/root/blob/master/interpreter/CMakeLists.txt#L96):
```cmake
if(NOT "${ROOT_CLING_TARGET}" STREQUAL "all")
  string(APPEND ROOT_CLING_TARGET ";NVPTX")
endif()
```
and a check [here](https://github.com/root-project/root/blob/master/interpreter/cling/CMakeLists.txt#L121C1-L124C8):
```cmake
if( NOT "NVPTX" IN_LIST LLVM_TARGETS_TO_BUILD)
  message(FATAL_ERROR "NVPTX backend is not activated\n"
    "Please enable it via -DLLVM_TARGETS_TO_BUILD=\"host;NVPTX\"")
endif()
```

Commenting these lines leads to build errors related to the `IncrementalCUDADeviceCompiler`.

### ROOT version

current master (`e76402aad8`)

### Installation method

Build from sourc

### Operating system

Ubuntu 24.04

dpiparo:
This is a great opportunity to review whether we can remove even more targets, therewith reducing binary sizes and compile time.

dpiparo:
fixed by https://github.com/root-project/root/pull/20225

PR #20225, opened 2025-10-29
Enable LLVM NVPTX target only if ROOT's cmake option 'cuda' is enabled

LukasBreitwieser:
In order to reduce binary size and compile time, we add the LLVM's NVPTX target only if ROOT's cmake option 'cuda' was enabled. See: https://github.com/root-project/root/issues/20208

Add null object definition for IncrementalCUDADeviceCompiler to avoid that the client code of this class needs to be wrapped in #ifdefs, in order to compile.

### DIFF
--- a/interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
+++ b/interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
@@ -102,7 +102,7 @@ namespace cling {
     LLVMInitializeNVPTXAsmPrinter();
 
     m_Init = true;
-#endif  // LLVM_HAS_NVPTX_TARGET
+#endif // LLVM_HAS_NVPTX_TARGET
   }
 
   void IncrementalCUDADeviceCompiler::setCuArgs(


interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
```diff
@@ -8,6 +8,7 @@
 //------------------------------------------------------------------------------
 
 #include "cling/Interpreter/IncrementalCUDADeviceCompiler.h"
+#ifdef ROOT_HAS_CUDA
```
vgvassilev: We can add that file in cmake if configured with the nvptx target. We should not need these ifdefs.

LukasBreitwieser: This line should be changed to `#ifdef LLVM_HAS_NVPTX_TARGET`. I used `ROOT_HAS_CUDA` before and missed to replace this occurrence.

About your comment. Do you suggest to conditionally add this source file to the cling library only if NVPTX is enabled? If so, the definition of the null object would need to move from the source to the header.

vgvassilev: I meant that this file should be part of the library only if we have the cuda target. The change should be here:

https://github.com/root-project/root/blob/2e4aeaa3ddcddb4b4399ee032546053597b21735/interpreter/cling/lib/Interpreter/CMakeLists.txt#L98

where we add this file only if we enabled the nvptx target in llvm.

LukasBreitwieser: ok, then I understood correctly. 
But that would mean that the following would need to move to the header.
https://github.com/root-project/root/pull/20225/files#diff-39c31edadb6da5c59aa145b14a24c93e7b5af06777e7cc44f8b5761cfdff7404R446-R476

The llvm build system seems to dislike this option:
```
  LLVM's build system enforces that all source files are added to a build
  target, that exactly one build target exists in each directory, and that
  this target lists all files in that directory.  If you want multiple
  targets in the same directory, add PARTIAL_SOURCES_INTENDED to the target
  specification, though it is discouraged.
 ```

Comment (vgvassilev) [reply interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp]: Yeah, that's why I think this is a mess but okay.

We can add `PARTIAL_SOURCES_INTENDED` switch (check out how it is done in llvm itself). I believe llvm will complain if we try to run code and we don't have the relevant target support. I am not sure if we need anything else there..

**Comment by dpiparo** [comment]
Do I understand correctly that if the tests pass, we will merge this change?

**Comment by dpiparo** [comment]
and, btw, what is the improvement in size and compile time?

File: interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp (hunk 1 above)
Comment (hahnjo) [reply interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp]: If I understand correctly, the issue is that functions to initialize the NVPTX target don't exist, like `LLVMInitializeNVPTXTargetInfo`. So it's not about LLVM complaining, but Cling won't even compile.

File: interpreter/cling/include/cling/Interpreter/IncrementalCUDADeviceCompiler.h [hunk 2]
```diff
@@ -6,10 +6,10 @@
 // of Illinois Open Source License or the GNU Lesser General Public License. See
 // LICENSE.TXT for details.
 //------------------------------------------------------------------------------
-
 #ifndef CLING_INCREMENTAL_CUDA_DEVICE_JIT_H
 #define CLING_INCREMENTAL_CUDA_DEVICE_JIT_H
 
+#ifdef LLVM_HAS_NVPTX_TARGET
```
hahnjo: This likely needs at least an include of the LLVM config to have somebody `#define` that macro

interpreter/cling/include/cling/Interpreter/IncrementalCUDADeviceCompiler.h
```diff
@@ -62,16 +62,15 @@ namespace cling {
                        const uint32_t fatbinFlags, const bool verbose,
                        const bool debug,
                        const std::vector<std::string> additionalPtxOpt)
-          : cppStdVersion(cppStdVersion),
-            hostTriple(hostTriple), smVersion(smVersion),
-            fatbinFlags(fatbinFlags), verbose(verbose), debug(debug),
-            additionalPtxOpt(additionalPtxOpt) {}
+          : cppStdVersion(cppStdVersion), hostTriple(hostTriple),
+            smVersion(smVersion), fatbinFlags(fatbinFlags), verbose(verbose),
+            debug(debug), additionalPtxOpt(additionalPtxOpt) {}
```
hahnjo: This formatting change is unrelated, as are many other modifications in this file. Please revert from this PR, and configure your editor / IDE to only (re)format lines that are modified.

hahnjo (changes requested):

hahnjo:
> and, btw, what is the improvement in size and compile time?

I didn't measure, but I would expect only little: the backends are a very small part of the entire LLVM libraries, the generic infrastructure, optimization passes and JIT functionality is much bigger.

vgvassilev:
> > and, btw, what is the improvement in size and compile time?
> 
> I didn't measure, but I would expect only little: the backends are a very small part of the entire LLVM libraries, the generic infrastructure, optimization passes and JIT functionality is much bigger.

I've mentioned privately that there is a little point in this change even if requested by an experiment. I do not oppose it but I do not see a benefit either except for adding some more conditionals in the codebase..

interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
devajithvs: I was wondering if we could do the same with:
```diff
diff --git a/interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp b/interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
index a492add8a01..6c284f9efea 100644
--- a/interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
+++ b/interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
@@ -16,6 +16,7 @@
 #include "clang/Frontend/CompilerInstance.h"
 #include "clang/Lex/HeaderSearchOptions.h"
 
+#include "llvm/Config/llvm-config.h" // for LLVM_HAS_NVPTX_TARGET
 #include "llvm/Support/Process.h"
 #include "llvm/Support/Program.h"
 #include "llvm/Support/raw_ostream.h"
@@ -93,6 +94,7 @@ namespace cling {
       return;
     }
 
+#if LLVM_HAS_NVPTX_TARGET
     // initialize NVPTX backend
     LLVMInitializeNVPTXTargetInfo();
     LLVMInitializeNVPTXTarget();
@@ -100,6 +102,7 @@ namespace cling {
     LLVMInitializeNVPTXAsmPrinter();
 
     m_Init = true;
+#endif
   }
 
   void IncrementalCUDADeviceCompiler::setCuArgs(
   ```

**Review by devajithvs — CHANGES_REQUESTED** [review]

**Comment by vgvassilev** [comment]
Including `"llvm/Config/llvm-config.h"` will break the builtin_clang=Off I think.

cc: @guitargeek

**Comment by hahnjo** [comment]
> Including `"llvm/Config/llvm-config.h"` will break the builtin_clang=Off I think.
> 
> cc: @guitargeek

Why? It is included by a number of headers and installed as part of LLVM. `builtin_clang` has nothing to do with it.

**Comment by pcanal** [comment]
> > >   and, btw, what is the improvement in size and compile time?
> I didn't measure, but I would expect only little: the backends are a very small part of the entire LLVM libraries, the generic infrastructure, optimization passes and JIT functionality is much bigger.
> I've mentioned privately that there is a little point in this change even if requested by an experiment. I do not oppose it but I do not see a benefit either except for adding some more conditionals in the codebase..

I agree with Vassil.  Before merging this we ought to measure the change in size and time as the stated motivation in https://github.com/root-project/root/issues/20208 is:
```
In order to reduce binary size and compile time ...
```
If this PR does not accomplish this, it sounds like the extra complication is not worth it.
However, if the requester had additional motivation/need behind the request then we ought to update the issue and re-review the PR in view of those.

TLDR: Please measure the gains :)

File: interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp (hunk 1 above)
Comment (LukasBreitwieser) [reply interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp]: Thanks for your comments!

**Comment by LukasBreitwieser** [comment]
This PR reduces the size of ROOT's install directory for a release build by a mere `2020 kB`, which is a `0.3%` improvement. For the build time I don't have a number right now.

Due to your comments, the complication is now quite small. So I would argue that it would be worth merging.

**Comment by vgvassilev** [comment]
My problem with merging this is that we lose the ability to call cling -cuda for such builds, not a big deal if we have diagnostics for the case but that makes the distribution less flexible for not a real impact… 

I will let other decide though…

**Comment by ferdymercury** [comment]
Maybe related: https://github.com/root-project/root/pull/20231

File: interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp [hunk 4]
```diff
@@ -93,13 +94,15 @@ namespace cling {
       return;
     }
 
+#if LLVM_HAS_NVPTX_TARGET == 1
```
hahnjo:
```suggestion
#if LLVM_HAS_NVPTX_TARGET
```

interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
```diff
@@ -93,13 +94,15 @@ namespace cling {
...
     // initialize NVPTX backend
     LLVMInitializeNVPTXTargetInfo();
     LLVMInitializeNVPTXTarget();
     LLVMInitializeNVPTXTargetMC();
     LLVMInitializeNVPTXAsmPrinter();
 
     m_Init = true;
+#endif  // #if LLVM_HAS_NVPTX_TARGET == 1
```
hahnjo:
```suggestion
#endif  // LLVM_HAS_NVPTX_TARGET
```

interpreter/CMakeLists.txt
```diff
@@ -92,7 +92,7 @@ else()
   set(ROOT_CLING_TARGET "all")
 endif()
 
-if(NOT "${ROOT_CLING_TARGET}" STREQUAL "all")
+if(cuda)
```
hahnjo: I think we want to keep the previous condition:
```suggestion
if(cuda AND NOT "${ROOT_CLING_TARGET}" STREQUAL "all")
```

hahnjo (approved):
Given the small changes required and that https://github.com/root-project/root/pull/20231 proposes a less correct approach, I personally think we should merge this even if the gain is small. Maybe we should update the commit message with a different motivation (and squash the commits). For me, that would be the fact that CUDA in Cling is an optional feature that we know doesn't properly work from inside ROOT (with modules). I am less worried about standalone `cling` since there the user has to actively work to turn off backends.

pcanal (approved):
Fair enough.

PR #20230, opened 2025-10-29
Remove unused method TChain::Loop

vepadulano:
Following up on this comment https://github.com/root-project/root/pull/20222#issuecomment-3462646223

### DIFF
--- a/tree/tree/inc/TChain.h
+++ b/tree/tree/inc/TChain.h
@@ -127,8 +127,7 @@ public:
    bool      InPlaceClone(TDirectory *newdirectory, const char *options = "") override;
    Int_t     LoadBaskets(Long64_t maxmemory) override;
    Long64_t  LoadTree(Long64_t entry) override;
-           void      Lookup(bool force = false);
-   virtual void      Loop(Option_t *option="", Long64_t nentries=kMaxEntries, Long64_t firstentry=0); // *MENU*
+   void      Lookup(bool force = false);
    void      ls(Option_t *option="") const override;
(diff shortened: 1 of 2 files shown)


tree/tree/src/TChain.cxx
```diff
@@ -1807,45 +1807,9 @@ void TChain::Lookup(bool force)
 ////////////////////////////////////////////////////////////////////////////////
 /// Loop on nentries of this chain starting at firstentry.  (NOT IMPLEMENTED)
 
-void TChain::Loop(Option_t* option, Long64_t nentries, Long64_t firstentry)
+void TChain::Loop(Option_t *, Long64_t, Long64_t)
```
pcanal: I think we can safely remove this routine altogether.  Its `TTree` sibling was removed 22 years ago: 5621165b06fd8a2bbffbe22009b9e66dd522d81b

tree/tree/inc/TChain.h
```diff
@@ -127,8 +127,7 @@ class TChain : public TTree {
    bool      InPlaceClone(TDirectory *newdirectory, const char *options = "") override;
    Int_t     LoadBaskets(Long64_t maxmemory) override;
    Long64_t  LoadTree(Long64_t entry) override;
-           void      Lookup(bool force = false);
-   virtual void      Loop(Option_t *option="", Long64_t nentries=kMaxEntries, Long64_t firstentry=0); // *MENU*
+   void Lookup(bool force = false);
```
pcanal: The spacing appear to not match (i.e. line 130 `Lookup` should be align with `LoadTree`).

PR #20231, opened 2025-10-29
[cling] Enable cling to be built with Emscripten

mcbarton:
# This Pull request:

## Changes or fixes:

This change enables cling to be built with Emscripten. The Emscripten build of cling doesn't currently work, but these changes allow it to build. If this goes in I will begin debugging what changes are needed to enable it to function in a web browser and node through CppInterOp, which has an Emscripten ci (including tests).

## Checklist:

- [x] tested changes locally

This PR fixes #

### DIFF
--- a/interpreter/cling/CMakeLists.txt
+++ b/interpreter/cling/CMakeLists.txt
@@ -118,7 +118,7 @@ else()
   endif()
 endif()
 
-if( NOT "NVPTX" IN_LIST LLVM_TARGETS_TO_BUILD)
+if( NOT "NVPTX" IN_LIST LLVM_TARGETS_TO_BUILD AND NOT EMSCRIPTEN)
   message(FATAL_ERROR "NVPTX backend is not activated\n"
     "Please enable it via -DLLVM_TARGETS_TO_BUILD=\"host;NVPTX\"")
 endif()
--- a/interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
+++ b/interpreter/cling/lib/Interpreter/IncrementalCUDADeviceCompiler.cpp
@@ -92,13 +92,13 @@ namespace cling {
       llvm::errs() << "Could not create PTX interpreter instance\n";
       return;
     }
-
+#ifndef __EMSCRIPTEN__
     // initialize NVPTX backend
     LLVMInitializeNVPTXTargetInfo();
     LLVMInitializeNVPTXTarget();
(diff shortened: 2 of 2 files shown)


ferdymercury:
Maybe related: https://github.com/root-project/root/pull/20225

devajithvs (commented):
I think the changes in https://github.com/root-project/root/pull/20225 should fix Cling build with Emscripten. Let's wait for that before making this change.

hahnjo (changes requested):
In my opinion, https://github.com/root-project/root/pull/20225 is the more correct approach because it's not about building with Emscripten, but really about the missing NVPTX backend.

vgvassilev:
> In my opinion, #20225 is the more correct approach because it's not about building with Emscripten, but really about the missing NVPTX backend.

What has to do the emscripten work with some other PR we discussed. Is there a reason to block the current PR?

hahnjo:
> > In my opinion, #20225 is the more correct approach because it's not about building with Emscripten, but really about the missing NVPTX backend.
> 
> What has to do the emscripten work with some other PR we discussed. Is there a reason to block the current PR?

I don't understand the first sentence of the comment. And yes, I object to this PR because `#ifndef __EMSCRIPTEN__` is wrong and `#if LLVM_HAS_NVPTX_TARGET` from https://github.com/root-project/root/pull/20225 is correct.

vgvassilev:
> and `#if LLVM_HAS_NVPTX_TARGET` from #20225 is correct.

~~In your opinion.~~

Now I see what you meant. Yes, we should not wrap the `nvptx` initialization around emscripten clauses.

hahnjo:
> > and `#if LLVM_HAS_NVPTX_TARGET` from #20225 is correct.
> 
> In your opinion.

Yes, obviously. You asked for the reason to block this PR, there it is. I'm open to discussion, but that's the technical arguments I see so far.

vgvassilev:
> > > and `#if LLVM_HAS_NVPTX_TARGET` from #20225 is correct.
> > 
> > 
> > In your opinion.
> 
> Yes, obviously. You asked for the reason to block this PR, there it is. I'm open to discussion, but that's the technical arguments I see so far.

Apologies -- I overlooked. Updated my comment...

mcbarton:
Closing as I checked locally and https://github.com/root-project/root/pull/20225 achieves the same result of allowing cling to be compiled with Emscripten.

PR #20236, opened 2025-10-30
Fix many problems in `TGraphPainter::PaintGraphReverse`

linev:
This method used when graph drawn with "rx" or "ry" draw option.

Example is: https://github.com/root-project/root/blob/master/roottest/graphics/macros/graphs/graphreverse.C

Following problems are resolved:

1. Zooming of Y axis. While TH1 used for axis painting. Y min/max stored in `fMininum/fMaximum` members of histogram. So one have to use these values for painting of Y axis.
2.  `Unzoom` of Y axis - fMinimum/fMaximum of histogram set to 0 and need to be properly configured again.
3. Memory leak - new TGaxis object and reversed graph was not deleted
4. Eliminate `gPad->Update()` inside painter which completely screw-up painting of TCanvas making wired PS/SVG output. Found with new stressGraphics tests in https://github.com/root-project/root/pull/20216


couet (approved):

dpiparo:
Thanks!

PR #20238, opened 2025-10-30
[cmake][tmva] Disable TMVA_CNN_Classification.py if tensorflow is not found

bellenot:


dpiparo:
Do we also need to do something about other tutorials that require TF? For example `RBatchGenerator_TensorFlow.py`?

bellenot:
> RBatchGenerator_TensorFlow

Well, I'll have to check on Linux, I only discovered TMVA_CNN_Classification.py because some other ones are disabled on WIndows:
```
if(MSVC AND NOT win_broken_tests)
  # RBatchGenerator tutorials don't work on Windows at the moment.
  list(APPEND dataframe_veto machine_learning/RBatchGenerator_NumPy.py)
  list(APPEND dataframe_veto machine_learning/RBatchGenerator_TensorFlow.py)
  list(APPEND dataframe_veto machine_learning/RBatchGenerator_PyTorch.py)
  [...]
```

bellenot:
@dpiparo ok thanks to your comment, I think I found the proper way of fixing this:
```
  file(GLOB requires_tensorflow RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
      machine_learning/RBatchGenerator_TensorFlow.py
      machine_learning/TMVA_CNN_Classification.py
  )
```

PR #20244, opened 2025-10-30
[graphics] Implement TGraphPolar::SavePrimitive

linev:
Was completely missing before.

Extend functionality of `SavePrimitiveVector` to handle empty arrays or arrays with only zeros
which can be replaced by "nullptr" in many constrctors like in `TGraphErrors`.
Use this functionality in meaningful places.

Fixes issue discovered with https://github.com/root-project/root/pull/20216


linev:
PR was ok before, only stressGraphics can have problem which I want to fix fix next PR

PR #20246, opened 2025-10-30
[graphics] Significant `TGraphPainter` improvment

linev:
Do not create graph histogram when no "A" draw option was specified.

In older ROOT versions `TGraph::GetHistogram()` method was not creating histogram, but since several years it does.
As result now any painting of `TGraph` object leads to creation of auxiliary histogram even when it does not used at all.
For objects with many graphs (like `TH2Poly`) this makes significant overhead in computation, in storage, in memory consumption. So try to avoid creation of such histogram if possible.

After this PR stored `TH2Poly` used significantly less space on disk. And painting performed faster.

Also plain `TGraph` painting will work faster. 

Need to be tested with newer stressGraphics

### DIFF
--- a/hist/histpainter/src/THistPainter.cxx
+++ b/hist/histpainter/src/THistPainter.cxx
@@ -1051,7 +1051,7 @@ is the color change between cells.
 
 The color palette in TStyle can be modified via `gStyle->SetPalette()`.
 
-All the non-empty bins are painted. Empty bins (bins with content and error equal to 0) are 
+All the non-empty bins are painted. Empty bins (bins with content and error equal to 0) are
 not painted unless some bins have a negative content because in that case the null bins
 might be not empty.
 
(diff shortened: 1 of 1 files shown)


ferdymercury:
Thanks @linev for this nice modification. I have a recurrent issue that could be related with this (or maybe that could be fixed at the same time than this change).

```
TGraph gr;
for(auto i = 0; i < 100; ++i) gr.AddPoint(i,i);
gr.AddPoint(10000, 0)
gr.Draw("A*")
// Try zooming x axis from 0 to 50, it does not work
```

This is because the underlying fixed-bin-width histogram is created with 101 bins, so bin 1 includes all first 100 points. See TGraph.cxx:1491

Does this PR change that behavior? If not, is there a way for the user to enable a "high-resolution" graph that has 10000 bins?

My current workaround is to artificially repeat the last point:
`for(auto i = 0; i < 1000; ++i) gr.AddPoint(10000,0);`

then I can easily zoom wherever I want.

So... is there a way to improve this behavior, or at least let the user specify the number of bins for the underlying TAxis, rather than just depending on the number of points?

For example, changing:
```
Int_t npt = 100;
if (fNpoints > npt) npt = fNpoints;
```
 
with:
If parent==TGraph:
`Int_t npt = GetParentGraph()->GetUserNpts();`

so that user can define this high-resolution zooming from script.

Another option could be to have a histogram with variable bin edges I guess. Or not relying on a histogram for objects such as TGraph.

Unless there is a way to decouple TGraph painting / zooming from fixed-bin-width underlying TAxis

If you think this is out of scope, I can open a separate issue.

linev:
Hi @ferdymercury. 
This PR should not change behavior of graph painter (I hope so), 
but will improve memory consumption and performance for many use cases. 

> // Try zooming x axis from 0 to 50, it does not work

For such special use-case you can directly create histogram like:

```
  auto h1 = new TH1F("hist", "hist", 10000, -10., 11000.);
  h1->SetMaximum(120);
  h1->SetStats(0);
  gr.SetHistogram(h1);
```

This should be done before first drawing of the graph.


Or simply start canvas in web mode - there zooming works perfectly.

Issue #20252, opened 2025-10-31
TGraph::SetHistogram() can leak memory


linev:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

When new histogram set to `TGraph` - old is lost and leads to memory leak.
Normally previous histogram should be deleted at this moment.
Histogram is owned by `TGraph` object and deleted in the destructor of `TGraph`

### Reproducer

```
{
  auto gr = new TGraph;
  for(int i = 0; i < 100; ++i) gr->AddPoint(i,i);
  gr->AddPoint(10000, 0);

  for (int n=0;n<1000;++n) {
     auto h1 = new TH1F("hist", "hist", 10000, -10., 11000.);
     h1->SetStats(0);
    h1->SetMaximum(120);
     gr->SetHistogram(h1);
}
  gr->Draw("A*");
}
```

### ROOT version

master

### Installation method

Build from source

### Operating system

Linux

PR #20254, opened 2025-10-31
[hist] avoid memleak when SetHistogram, mention transferred ownership

ferdymercury:
Fixes https://github.com/root-project/root/issues/20252

Please squash on merge.

### DIFF
--- a/hist/hist/src/TGraph.cxx
+++ b/hist/hist/src/TGraph.cxx
@@ -2266,6 +2266,13 @@ void TGraph::SetHighlight(Bool_t set)
    painter->SetHighlight(this);
 }
 
+/// Set the histogram underlying the TGraph. This transfers the ownership of h to the TGraph. The preexisting fHistogram will be deleted.
+void  TGraph::SetHistogram(TH1F *h)
+{
+   delete fHistogram;
+   fHistogram = h;
+}
+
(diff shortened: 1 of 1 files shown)


hist/hist/src/TGraph.cxx
```diff
@@ -2266,6 +2266,13 @@ void TGraph::SetHighlight(Bool_t set)
    painter->SetHighlight(this);
 }
 
+/// Set the histogram underlying the TGraph. This transfers the ownership of h to the TGraph. The preexisting fHistogram will be deleted.
+void  TGraph::SetHistogram(TH1F *h)
+{
+   delete fHistogram;
```
dpiparo, line 2272: Would it be useful to perhaps check if the pointer is not null before deleting?

ferdymercury: It's not needed since `delete nullptr;` is safe to call. (https://stackoverflow.com/questions/6731331/is-it-still-safe-to-delete-nullptr-in-c0x)

linev:
This solution can be dangerous - if users was aware about such leak and tried to cleanup memory himself.
Probably instead of simple `delete fHistogram` one can add extra checks like:
```
if (fHistorgam && !fHistorgam->IsZombie())
   delete fHistorgam;
```
Although it is not perfect solution as well.

PR #20256, opened 2025-10-31
[codecov] Include partials as hits

hahnjo:
After this change, partially covered lines will also count in the Coverage %.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,3 @@
+parsers:
+  cobertura:
+    partials_as_hits: true


pcanal:
Reporting about branch coverage data as a summary metric is indeed probably not useful.  We should, however, probably consider adding a mode to see it, **on demand**, for a specific patch when one can look carefully at the code. e.g. for coverage-guided testing.

Example of ' technically correct' but unhelpful reports:
```
   int *m = new int[3];
   delete m;  // report about the implied `if (m)` not being fully covered
``` 
and in the following, it also points out that some of the implied cases are not covered.
```
  try {
    throw std::runtime_error("");
  } catch (...) {
     // nothing
  }
```

pcanal (approved):
This helps.  We probably should offer an on-demand option to turn it off (maybe).

Issue #16325, opened 2024-08-28
[ntuple] Better customization options for RClusterPool (read-ahead)


jblomer:
### Explain what you would like to see improved and how.

The RClusterPool should be improved
- to work with a memory budget rather than a fixed number of clusters to read-ahead
- allow users to exclude/include specific fields

### ROOT version

master

### Installation method

n/a

### Operating system

n/a

PR #20257, opened 2025-10-31
[ntuple] Add support for efficient multi-stream reading

jblomer:
Enables a shared RNTupleReader to read multiple streams efficiently. On the page source layer, add the possibility to pin and unpin clusters. Data from pinned clusters will not be evicted from the cluster pool or the page pool.

Extend the RNTupleReader API by "active entry tokens". Active entry tokens keep an entry alive in the cache. Internally, the active entries turn into a reference counter for the corresponding cluster, so that the clusters are pinned and unpinned correctly.

Active entry tokens should provide a flexible API not only to support multiple streams but also to keep, e.g., certain (past) reference events alive.

While this functionality is different from the description in #16325, it may be the flexibility that is actually needed.

@Dr15Jones FYI

Graphical output of the tutorial:

image

### DIFF
--- a/tree/ntuple/doc/Architecture.md
+++ b/tree/ntuple/doc/Architecture.md
@@ -251,6 +251,11 @@ Data is populated to an explicit `REntry` or the model's default entry through `
 The reader can create `RNTupleView` objects for the independent reading of individual fields.
 The reader can create `RBulkValues` objects for bulk reading of individual fields.
 
+By default, the caches for compressed and uncompressed data used by the reader will assume sequential access
+(or at least a read pattern that is monotonic in the entry number).
+The reader can, however, create "active entry tokens" that are used keep data of past entries pinned in the cache.
+This can be useful to implement multi-stream reading where multiple threads share a single reader.
+
 Additionally, the reader provides access to a cached copy of the descriptor.
 It can display individual entries (`RNTupleReader::Show()`) and summary information (`RNTupleReader::PrintInfo()`).
 
@@ -424,6 +429,8 @@ That means that RDataFrame uses a separate data source for every thread, each of
 
 ### Concurrent Readers
 Multiple readers can read the same RNTuple concurrently as long as access to every individual reader is sequential.
+The reader support efficient multi-stream access (multiple threads/streams sharing the same reader)
+as long as reader access is locked and every stream uses an active entry token.
 
 ### Parallel REntry Preparation
 Multiple `REntry` object can be concurrently prepared by multiple threads.


tree/ntuple/src/RClusterPool.cxx
```diff
@@ -183,7 +183,21 @@ class RProvides {
...
+      auto descriptorGuard = fPageSource.GetSharedDescriptorGuard();
+
+      for (ROOT::DescriptorId_t i = 1, next = cid; i < 2 * fClusterBunchSize; ++i) {
+         next = descriptorGuard->FindNextClusterId(next);
+         if (next == ROOT::kInvalidNTupleIndex ||
+             !fPageSource.GetEntryRange().IntersectsWith(descriptorGuard->GetClusterDescriptor(next))) {
+            break;
+         }
+
+         keep.insert(next);
```
hahnjo, line 194: This modifies the `keep` container that we are currently iterating over - did you mean to iterate over `fPinnedClusters` instead?

tree/ntuple/inc/ROOT/RNTupleReader.hxx
```diff
@@ -64,7 +66,59 @@ std::cout << "myNTuple has " << reader->GetNEntries() << " entries\n";
...
+      /// SetEntryNumber() should be called before the corresponding entry is used (through LoadEntry() or views).
+      void SetEntryNumber(NTupleSize_t entryNumber);
+      /// Release the entry number, i.e. allow the corresponding data to be evicted from caches.
+      /// Called implicitly on destruction.
+      void Reset();
+   };
+
 private:
+   /// Shared data structure between the reader and all the issued active entry tokens.
+   struct RActiveEntriesControlBlock {
```
hahnjo, line 71: Is it possible to move the definition before `RActiveEntryToken`, ie without the need for the forward declaration?

tree/ntuple/src/RNTupleReader.cxx
```diff
@@ -22,6 +22,102 @@
...
+{
+   fPtrControlBlock = other.fPtrControlBlock;
+   SetEntryNumber(other.fEntryNumber);
+}
+
+ROOT::RNTupleReader::RActiveEntryToken::RActiveEntryToken(RActiveEntryToken &&other)
+{
+   fPtrControlBlock = other.fPtrControlBlock;
+   SetEntryNumber(other.fEntryNumber);
+   other.Reset();
```
hahnjo: For the move-constructor, we should be able to "steal" the state of `other`: set `fEntryNumber = other.fEntryNumber` and `other.fEntryNumber = kInvalidNTupleIndex`. That should avoid refcount changes

tree/ntuple/src/RNTupleReader.cxx
```diff
@@ -22,6 +22,102 @@
...
+ROOT::RNTupleReader::RActiveEntryToken &ROOT::RNTupleReader::RActiveEntryToken::operator=(RActiveEntryToken &&other)
+{
+   if (this == &other)
+      return *this;
+   if (other.fPtrControlBlock.get() != fPtrControlBlock.get()) {
+      Reset();
+      fPtrControlBlock = other.fPtrControlBlock;
+   }
+   SetEntryNumber(other.fEntryNumber);
+   other.Reset();
```
hahnjo: same

tree/ntuple/test/ntuple_multi_stream.cxx
```diff
@@ -0,0 +1,164 @@
...
+      writer->Fill();
+   }
+
+   {
+      auto reader = ROOT::RNTupleReader::Open("ntpl", fileGuard.GetPath());
+      reader->EnableMetrics();
+
+      auto t1 = reader->CreateActiveEntryToken();
+      t1.SetEntryNumber(0);
+      auto t2 = reader->CreateActiveEntryToken(); // t1;
```
hahnjo: May be a good opportunity to the copy constructor

tree/ntuple/doc/Architecture.md
```diff
@@ -251,6 +251,11 @@ Data is populated to an explicit `REntry` or the model's default entry through `
 The reader can create `RNTupleView` objects for the independent reading of individual fields.
 The reader can create `RBulkValues` objects for bulk reading of individual fields.
 
+By default, the caches for compressed and uncompressed data used by the reader will assume sequential access
+(or at least a read pattern that is monotonic in the entry number).
+The reader can, however, create "active entry tokens" that are used keep data of past entries pinned in the cache.
+This can be useful to implement "event mixing" or multi-stream reading where multiple threads share a single reader.
```
hahnjo: I think "event mixing" is something else in HEP (overlaying multiple collision events), maybe we can find a different term? Maybe "out-of-order reading"?

tree/ntuple/src/RNTupleReader.cxx
jblomer: I don't think we can do the same for move assignment: here, `this` and `other` are initialized and may or may not point to the same cluster/entry. I think here it is clearer to go through proper refcounting.

tree/ntuple/src/RClusterPool.cxx
jblomer: Thanks, good catch!

tree/ntuple/src/RNTupleReader.cxx
hahnjo: I'm not sure I see the issue, I think we could just swap the `fEntryNumber` and let `other.Reset()` or even its eventual destruction decrement the refcount. But anyway a minor comment not that relevant

jblomer: Oh, yes. You're right.

PR #20258, opened 2025-10-31
[skip-ci][tutorials] refactor roofit tutorial

siliataider:
[This failure](https://github.com/root-project/root/actions/runs/18940952328/job/54110599747) appeared when testing the python wheels in #19600:

```shell
    legend2.SetFillColor("kWhite")
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
TypeError: none of the 2 overloaded methods succeeded. Full details:
  void TAttFill::SetFillColor(TColorNumber) =>
    TypeError: could not convert argument 1
  short conversion expects an integer object
```

### DIFF
--- a/tutorials/roofit/roofit/rf615_simulation_based_inference.py
+++ b/tutorials/roofit/roofit/rf615_simulation_based_inference.py
@@ -33,8 +33,8 @@
 ## \date July 2024
 ## \author Robin Syring
 
-import ROOT
 import numpy as np
+import ROOT
 from sklearn.neural_network import MLPClassifier
 
 # The samples used for training the classifier in this tutorial / rescale for more accuracy
@@ -45,8 +45,7 @@ ROOT.RooMsgService.instance().setGlobalKillBelow(ROOT.RooFit.WARNING)
 
 
 # Morphing as a baseline
-def morphing(setting):
-
+def morphing(setting, workspace):
     # Define binning for morphing
     grid = ROOT.RooMomentMorphFuncND.Grid(ROOT.RooBinning(4, 0.0, 4.0))
     x_var.setBins(50)
@@ -199,7 +198,7 @@ nllr_learned = pdf_learned.createNLL(obs_data)
 ROOT.SetOwnership(nllr_learned, True)
 
 # Compute the morphed nll
-morphing(ROOT.RooMomentMorphFuncND.Linear)
+morphing(ROOT.RooMomentMorphFuncND.Linear, workspace)
 nll_morph = workspace["morph"].createNLL(obs_data)
 ROOT.SetOwnership(nll_morph, True)
 


vepadulano:
@guitargeek I vaguely remember some Pythonizations in RooFit that converted strings with the same name as ROOT constants to the actual constant value to allow for situations similar to the one of this tutorial. Maybe this particular overload is not Pythonized? I could also be misremembering.

tutorials/roofit/roofit/rf615_simulation_based_inference.py
```diff
@@ -33,8 +33,8 @@
 ## \date July 2024
 ## \author Robin Syring
 
-import ROOT
```
vepadulano, line 36: Any particular reason to move this import statement (I imagine it's ruff format)?

tutorials/roofit/roofit/rf615_simulation_based_inference.py
```diff
@@ -58,25 +57,25 @@ def morphing(setting):
         # Define the sampled gausians
         mu_help = ROOT.RooRealVar(f"mu{i}", f"mu{i}", i)
         help = ROOT.RooGaussian(f"g{i}", f"g{i}", x_var, mu_help, sigma)
-        workspace.Import(help, Silence=True)
+        workspace.Import(help, Silence=True)  # noqa F821
```
vepadulano: Why the noqa?

tutorials/roofit/roofit/rf615_simulation_based_inference.py
siliataider: Yes Ruff moved NumPy up (ordering by group)

tutorials/roofit/roofit/rf615_simulation_based_inference.py
siliataider: ```F821 Undefined name `workspace` ```

vepadulano: Ah I see now! This is thanks to the magic of Python variable scoping 😓 The `workspace` variable is initialised at line 159 of the tutorial and this function implicitly inherits it from the global scope by chance only because it's executed after the variable is initialised in the global scope.

guitargeek: Can we refactor the tutorial such that `workspace` is passed as an argument to the function where this warning happens? Then we don't rely on Pythons special variable scoping

guitargeek (changes requested):
Thanks a lot! I'd just suggest so refactor the tutorial such that we don't rely on Pythons variable scoping.

guitargeek:
> @guitargeek I vaguely remember some Pythonizations in RooFit that converted strings with the same name as ROOT constants to the actual constant value to allow for situations similar to the one of this tutorial. Maybe this particular overload is not Pythonized? I could also be misremembering.

This looks like a bug in the implicit conversion of `cppyy`. It should work, because `TColorNumber` can be constructed from a string:
https://github.com/root-project/root/blob/master/core/base/inc/TColor.h#L145

So we can merge this PR to make things work, but also open a GitHub issue about this and revert the change once the string version works again (probably with the cppyy update by @Vipul-Cariappa and @aaronj0). I will investigate and open such an issue.

guitargeek:
Ok, it has nothing to do with cppyy, I just didn't know that two implicit conversions at once are not allowed in C++ :slightly_smiling_face:

So this can be fixed on the C++ side: https://github.com/root-project/root/pull/20271

Let's see how the CI responds to that PR, but if it works In don't think we need to change from `"kWhite"` back to the more verbose `ROOT.kWhite`. It would still be worth it to do the refactoring to avoid relying on Python-specific variable scoping though!

Thanks a lot for catching this in any case

siliataider:
Sure, then I'll wait for the status of that PR before changing this one

guitargeek (approved):
Thank you very much!

PR #20260, opened 2025-10-31
[skip-ci] improve TPad::Divide help

couet:
improve TPad::Divide help as requested here https://github.com/root-project/root/pull/20209

### DIFF
--- a/graf2d/gpad/src/TPad.cxx
+++ b/graf2d/gpad/src/TPad.cxx
@@ -1264,10 +1264,7 @@ Int_t TPad::DistancetoPrimitive(Int_t px, Int_t py)
 ///    void divpad(Int_t nx=3, Int_t ny=2) {
 ///       gStyle->SetOptStat(0);
 ///       auto C = new TCanvas();
-///       C->SetLeftMargin(0.3);
-///       C->SetRightMargin(0.3);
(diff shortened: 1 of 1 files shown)


rlalik:
This part of the documentation is also incorrect:
```cpp
///  - xmargin is the space along x between pads in percent of canvas.
///  - ymargin is the space along y between pads in percent of canvas.
```
The more correct could look like:
```cpp
///  - xmargin is the space around each pad along x, distance between pads along x is equal to 2*xmargin.
///  - ymargin is the space around each pad along y, distance between pads along y is equal to 2*ymargin.
```

rlalik:
The new doc sounds really good. I was wondering only about perhaps replacing:
```cpp
C->SetLeftMargin(0.3);
C->SetRightMargin(0.3);
C->SetTopMargin(0.3);
C->SetBottomMargin(0.3);
```
with
```cpp
C->SetMargin(0.3, 0.3, 0.3, 0.3);
```
The former is very verbose, takes up vertical space. I like the way of setting all margins at once.

But beside that I think this can be merged. Unless you would like to wait for decision of my PR, because this will again change the definition (the old one will stay valid).

couet:
good I'll merge

Issue #20015, opened 2025-09-30
libcppyy shows up in top-level site-packages


hageboeck:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

In the spirit of #14917, Gentoo with ROOT master warns about:
```
 * Verifying compiled files for python3.13                                                                                                                                                                                             
 *                                                                                                                                                                                                                                              
 * QA Notice: The following unexpected files/directories were found                                                                                                                                                                             
 * top-level in the site-packages directory:                                                                                                                                                                                                    
 *                                                                                                                                                                                                                                              
 *   /usr/lib/python3.13/site-packages/libcppyy.so.6.37.01                                                                                                                                                                                      
 *                                                                                                                                                                                                                                              
 * This is most likely a bug in the build system.  More information                                                                                                                                                                             
 * can be found in the Python Guide:                                                                                                                                                                                                           
 * https://projects.gentoo.org/python/guide/qawarn.html#stray-top-level-files-in-site-packages  
```

### Reproducer

Install root-master in the system

### ROOT version

Master

### Installation method

Source

### Operating system

Linux

### Additional context

_No response_

PR #20262, opened 2025-10-31
[NFC] Update release notes with build system, math and RDF changes.

hageboeck:


README/ReleaseNotes/v638/index.md
```diff
@@ -143,21 +153,23 @@ If you want to keep using `TList*` return values, you can write a small adapter
...
 
 ### Distributed RDataFrame
 - Processing of RNTuples is now done in the same way as for TTrees, as it is a more efficient way of distributing the computations. The API remains unchanged.
 
 ## Python Interface
 
-ROOT dropped support for Python 3.8, meaning ROOT now requires at least Python 3.9.
+- ROOT dropped support for Python 3.8, meaning ROOT now requires at least Python 3.9.
+- `libcppyy.so` was moved from the top-level `site-packages` to `site-packages/cppyy`.
```
vepadulano: This point sounds to me like a bit too low-level detail for RN, it's more like something I would tie to a resolved issue. An alternative could be describing the same things in more high-level terms, e.g. "the cppyy library is now installed in a relative directory to the user Python installation directory (usually `site-packages` in the current Python environment). This provides better isolation and is more in line with standard Python packaging practices. Certain Linux distributions such as Gentoo raise warnings about files installed at the top-level of the `site-packages` folder https://projects.gentoo.org/python/guide/qawarn.html#stray-top-level-files-in-site-packages"

hageboeck: Thanks, that's a good point! This reminded me that there is indeed an issue for it, #20015, so it will show up in the release notes automatically. I'll remove the line.

PR #20271, opened 2025-11-03
[core] Add TColorNumber constructor from a C-style string

guitargeek:
We also require an overload for C-style strings. That's because the TColorNumber is designed to be constructed implicitly, and C++ won't do two user-defined conversions in a single implicit conversion chain.


siliataider:
Thanks! I adapted #20258 accordingly

PR #20272, opened 2025-11-03
[ntpl] Allow XRootD redirection with EOS files

### LINKED ISSUE #20213: Implement RNTuple redirection from FUSE to XRootD

In pull request #11644 a redirection was implemented to allow ROOT to bypass FUSE access to EOS and instead open the file via XRootD, which has better performance. Unfortunately, this redirection does not work for `RNTuple`, as the way it opens `TFile`s bypasses the place where the hook was created to perform this redirection. It would be nice to create a similar redirection that worked also for `RNTuple`. The place this has to be done is likely around here:

https://github.com/root-project/root/blob/a460d3f338106128c0c994c0cf7aba3c409f960d/tree/ntuple/src/RPageStorageFile.cxx#L281-L306

### ROOT version

ROOT 6.36.04

### Installation method

emerge

### Operating system

Gentoo Linux

### Additional context

See also https://root.cern/doc/v628/release-notes.html#faster-reading-from-eos

vepadulano:
Fixes #20213

With these changes I can see the redirection happening when reading an RNTuple on EOS from lxplus (going from `ntpl001_staff.root` to `root://eoshome-v.cern.ch//eos/user/v/vpadulan/ntpl001_staff.root`

```
Breakpoint 1, ROOT::Internal::RRawFile::Create (url=..., options=...) at /tmp/vpadulan/root/io/io/src/RRawFile.cxx:88       
88         std::string transport = GetTransport(url);                                                                       
(gdb) n  
89         if (transport == "file") {
(gdb) p url                                                                                                                 
$1 = {_M_len = 18, _M_str = 0x4030a3 "ntpl001_staff.root"}  
[...]
103                    if (auto baseName = fileNameFromUrl.substr(fileNameFromUrl.find_last_of("/") + 1);
(gdb) 
105                       return Create(xurl, options);
(gdb) s
[...]
Breakpoint 1, ROOT::Internal::RRawFile::Create (url=..., options=...) at /tmp/vpadulan/root/io/io/src/RRawFile.cxx:88       
88         std::string transport = GetTransport(url);                                                                       
(gdb) n                                                                                                                     
89         if (transport == "file") {                                                                                       
(gdb) 
113        if (transport == "http" || transport == "https" ||
(gdb) p url
$8 = {_M_len = 64, _M_str = 0x136a790 "root://eoshome-v.cern.ch//eos/user/v/vpadulan/ntpl001_staff.root"}
(gdb) p transport
$9 = {static npos = 18446744073709551615, 
  _M_dataplus = {<std::allocator<char>> = {<std::__new_allocator<char>> = {<No data fields>}, <No data fields>}, 
    _M_p = 0x7fffffffa230 "root"}, _M_string_length = 4, {
    _M_local_buf = "root\000\177\000\000\220\242\377\377\377\177\000", _M_allocated_capacity = 139639930187634}}
(gdb) 
```

There might be more places where the redirection needs to happen though.

### DIFF
--- a/io/io/CMakeLists.txt
+++ b/io/io/CMakeLists.txt
@@ -21,6 +21,7 @@ if (uring)
 endif ()
 
 ROOT_LINKER_LIBRARY(RIO
+  src/InternalIOUtils.cxx
   src/RConcurrentHashColl.cxx
   src/RFile.cxx
   src/RRawFile.cxx
@@ -66,6 +67,8 @@ ROOT_LINKER_LIBRARY(RIO
 )
 
 target_include_directories(RIO PRIVATE ${CMAKE_SOURCE_DIR}/core/clib/res)
+# PUBLIC so that dependent libraries e.g. ROOTDataFrame may find the internal header too
+target_include_directories(RIO PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/res>)
 target_link_libraries(RIO PUBLIC ${ROOT_ATOMIC_LIBS})
 target_link_libraries(RIO PRIVATE nlohmann_json::nlohmann_json)
 
--- a/io/io/CMakeLists.txt
+++ b/io/io/CMakeLists.txt
@@ -21,6 +21,7 @@ if (uring)
 endif ()
 
 ROOT_LINKER_LIBRARY(RIO
+  src/InternalIOUtils.cxx
   src/RConcurrentHashColl.cxx
   src/RFile.cxx
   src/RRawFile.cxx
@@ -66,6 +67,7 @@ ROOT_LINKER_LIBRARY(RIO
 )
 
 target_include_directories(RIO PRIVATE ${CMAKE_SOURCE_DIR}/core/clib/res)
+target_include_directories(RIO PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/res>)
 target_link_libraries(RIO PUBLIC ${ROOT_ATOMIC_LIBS})
 target_link_libraries(RIO PRIVATE nlohmann_json::nlohmann_json)
 
--- a/io/io/res/ROOT/InternalIOUtils.hxx
+++ b/io/io/res/ROOT/InternalIOUtils.hxx
@@ -1,0 +1,24 @@
+// Author: Vincenzo Eduardo Padulano (CERN), 08/2026
+
+#ifndef ROOT_IO_UTILS
+#define ROOT_IO_UTILS
+
+#include <optional>
+#include <string>
+#include <string_view>
+
+namespace ROOT::Internal {
+
+/// \brief Get extended attribute value from path
+/// \param path Path to the file to check
+/// \param xattr Extended attribute to evaluate
+/// \return The string containing the extended attribute value if found, std::nullopt otherwise
+std::optional<std::string> GetXAttrVal(std::string_view path, std::string_view xattr);
+
+/// \brief Redirects the input path to the equivalent XRootD URL on EOS
+/// \param inputUrl The input path to redirect
+/// \return The redirected URL in case of successful redirection, std::nullopt otherwise
+std::optional<std::string> GetEOSRedirectedXRootURL(std::string_view inputPath);
+} // namespace ROOT::Internal
+
+#endif
--- a/io/io/res/ROOT/InternalIOUtils.hxx
+++ b/io/io/res/ROOT/InternalIOUtils.hxx
@@ -1,0 +1,24 @@
+// Author: Vincenzo Eduardo Padulano (CERN), 11/2025
+
+#ifndef ROOT_IO_UTILS
+#define ROOT_IO_UTILS
+
+#include <optional>
+#include <string>
+#include <string_view>
+
+namespace ROOT::Internal {
+
+/// \brief Get extended attribute value from path
+/// \param path Path to the file to check
+/// \param xattr Extended attribute to evaluate
+/// \return The string containing the extended attribute value if found, std::nullopt otherwise
+std::optional<std::string> GetXAttrVal(std::string_view path, std::string_view xattr);
+
+/// \brief Redirects the input URL to the equivalent XRootD path on EOS
+/// \param inputUrl The input URL to redirect
+/// \return The redirected URL in case of successful redirection, std::nullopt otherwise
+std::optional<std::string> GetEOSRedirectedXRootURL(std::string_view inputURL);
+} // namespace ROOT::Internal
+
+#endif
--- a/io/io/src/InternalIOUtils.cxx
+++ b/io/io/src/InternalIOUtils.cxx
@@ -1,0 +1,66 @@
+#include "ROOT/InternalIOUtils.hxx"
+
+#include "ROOT/RConfig.hxx" // R__UNIX
+
+#ifdef R__UNIX
+// getxattr
+#ifdef R__FBSD
+#include <sys/extattr.h>
+#else
+#include <sys/xattr.h>
+#endif
+
+#ifdef R__MACOSX
+/* On macOS getxattr takes two extra arguments that should be set to 0 */
+#define getxattr(path, name, value, size) getxattr(path, name, value, size, 0u, 0)
+#endif
+
+#ifdef R__FBSD
+#define getxattr(path, name, value, size) extattr_get_file(path, EXTATTR_NAMESPACE_USER, name, value, size)
+#endif
+
+#include "ROOT/StringUtils.hxx" // ROOT::StartsWith
+#include "TEnv.h"               // TEnv::GetValue
+#endif
+
+std::optional<std::string>
+ROOT::Internal::GetXAttrVal([[maybe_unused]] std::string_view path, [[maybe_unused]] std::string_view xattr)
+{
+#ifdef R__UNIX
+   // First call to getxattr evaluates the length of the extended attribute value
+   if (auto len = getxattr(path.data(), xattr.data(), nullptr, 0); len > 0) {
+      std::string xval(len, 0);
+      // Second call extracts the extended attribute value, checking it's of the correct length
+      if (getxattr(path.data(), xattr.data(), xval.data(), len) == len)
+         return xval;
+   }
+#endif
+   return std::nullopt;
+}
+
+std::optional<std::string> ROOT::Internal::GetEOSRedirectedXRootURL([[maybe_unused]] std::string_view inputURL)
+{
+#ifdef R__UNIX
+   if (inputURL.empty() || inputURL.back() == '/')
+      return std::nullopt;
+
+   if (gEnv->GetValue("TFile.CrossProtocolRedirects", 1) != 1)
+      return std::nullopt;
+
+   auto xurl = ROOT::Internal::GetXAttrVal(inputURL, "eos.url.xroot");
+   if (!xurl)
+      return std::nullopt;
+
+   auto baseName = inputURL.substr(inputURL.find_last_of("/") + 1);
+   // Sometimes the `getxattr` call may return an invalid URL due
+   // to the POSIX attribute not being yet completely filled by EOS.
+   if (!std::equal(baseName.crbegin(), baseName.crend(), xurl->crbegin()))
+      return std::nullopt;
+
+   // Ensure the redirected URL actually starts with the XRootD protocol string
+   if (ROOT::StartsWith(*xurl, "root://") || ROOT::StartsWith(*xurl, "xroot://") ||
+       ROOT::StartsWith(*xurl, "roots://") || ROOT::StartsWith(*xurl, "xroots://"))
+      return xurl;
+#endif
+   return std::nullopt;
+}
--- a/io/io/src/InternalIOUtils.cxx
+++ b/io/io/src/InternalIOUtils.cxx
@@ -1,0 +1,82 @@
+#include "ROOT/InternalIOUtils.hxx"
+
+#include "ROOT/RConfig.hxx" // R__UNIX
+
+#ifdef R__UNIX
+
+#include <ROOT/RLogger.hxx>
+
+ROOT::RLogChannel &InternalLogChannel()
+{
+   static ROOT::RLogChannel sLog("ROOT.Internal");
+   return sLog;
+}
+
+// getxattr
+#ifdef R__FBSD
+#include <sys/extattr.h>
+#else
+#include <sys/xattr.h>
+#endif
+
+#ifdef R__MACOSX
+/* On macOS getxattr takes two extra arguments that should be set to 0 */
+#define getxattr(path, name, value, size) getxattr(path, name, value, size, 0u, 0)
+#endif
+
+#ifdef R__FBSD
+#define getxattr(path, name, value, size) extattr_get_file(path, EXTATTR_NAMESPACE_USER, name, value, size)
+#endif
+
+#include "ROOT/StringUtils.hxx" // ROOT::StartsWith
+#include "TEnv.h"               // TEnv::GetValue
+#endif
+
+std::optional<std::string>
+ROOT::Internal::GetXAttrVal([[maybe_unused]] std::string_view path, [[maybe_unused]] std::string_view xattr)
+{
+#ifdef R__UNIX
+   // First call to getxattr evaluates the length of the extended attribute value
+   if (auto len = getxattr(path.data(), xattr.data(), nullptr, 0); len >= 0) {
+      std::string xval(len, 0);
+      // Second call extracts the extended attribute value, checking it's of the correct length
+      if (getxattr(path.data(), xattr.data(), xval.data(), len) == len)
+         return xval;
+   }
+#endif
+   return std::nullopt;
+}
+
+std::optional<std::string> ROOT::Internal::GetEOSRedirectedXRootURL([[maybe_unused]] std::string_view inputPath)
+{
+#ifdef R__UNIX
+   if (inputPath.empty() || inputPath.back() == '/')
+      return std::nullopt;
+
+   if (gEnv->GetValue("TFile.CrossProtocolRedirects", 1) != 1)
+      return std::nullopt;
+
+   auto xurl = ROOT::Internal::GetXAttrVal(inputPath, "eos.url.xroot");
+   if (!xurl)
+      return std::nullopt;
+
+   auto baseName = inputPath.substr(inputPath.find_last_of("/") + 1);
+   // Sometimes the `getxattr` call may return an invalid URL due
+   // to the POSIX attribute not being yet completely filled by EOS.
+   if (!std::equal(baseName.crbegin(), baseName.crend(), xurl->crbegin())) {
+      R__LOG_WARNING(InternalLogChannel())
+         << "Could not find path base name '" << baseName << "' in redirected URL '" << *xurl << "'.";
+      return std::nullopt;
+   }
+
+   // Ensure the redirected URL actually starts with the XRootD protocol string
+   if (ROOT::StartsWith(*xurl, "root://") || ROOT::StartsWith(*xurl, "xroot://") ||
+       ROOT::StartsWith(*xurl, "roots://") || ROOT::StartsWith(*xurl, "xroots://"))
+      return xurl;
+   else
+      R__LOG_WARNING(InternalLogChannel())
+         << "Redirected URL '" << *xurl << "' does not begin with any valid XRootD protocol string.";
+
+#endif
+   return std::nullopt;
+}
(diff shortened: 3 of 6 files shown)


io/io/src/RRawFile.cxx
```diff
@@ -68,6 +90,23 @@ ROOT::Internal::RRawFile::Create(std::string_view url, ROptions options)
...
+         ssize_t len = getxattr(url.data(), "eos.url.xroot", nullptr, 0);
+         if (len > 0) {
+            std::string xurl(len, 0);
+            std::string fileNameFromUrl{url.data()};
+            if (getxattr(fileNameFromUrl.c_str(), "eos.url.xroot", &xurl[0], len) == len) {
+               // Sometimes the `getxattr` call may return an invalid URL due
+               // to the POSIX attribute not being yet completely filled by EOS.
+               if (auto baseName = fileNameFromUrl.substr(fileNameFromUrl.find_last_of("/") + 1);
+                   std::equal(baseName.crbegin(), baseName.crend(), xurl.crbegin())) {
+                  return Create(xurl, options);
```
jblomer: I think we should check that the result actually has the XRootD protocol prefix. Otherwise a malfunctioning file system can send us in an endless loop.

io/io/src/RRawFile.cxx
```diff
@@ -68,6 +90,23 @@ ROOT::Internal::RRawFile::Create(std::string_view url, ROptions options)
 #ifdef _WIN32
       return std::unique_ptr<RRawFile>(new RRawFileWin(url, options));
 #else
+      // If URL is a file on an EOS FUSE mount, attempt redirection to XRootD protocol.
+      if (gEnv->GetValue("TFile.CrossProtocolRedirects", 1) == 1) {
+         // At the moment the url parameter can only be the file name
+         ssize_t len = getxattr(url.data(), "eos.url.xroot", nullptr, 0);
```
jblomer: I think we should combine the logic here and in TFile into a common header (in RIO) that gives us one function to read an extended attribute.

jblomer (commented):
Thanks for addressing this!

io/io/src/IOUtils.cxx
```diff
@@ -0,0 +1,80 @@
...
+   return -1;
+#endif
+}
+
+std::optional<std::string> ROOT::Internal::GetEOSRedirectedXRootUrl(std::string_view inputUrl)
+{
+   if (inputUrl.empty())
+      return std::nullopt;
+
+#ifdef R__UNIX
```
pcanal: (minor)
```suggestion
#ifdef R__UNIX
   if (inputUrl.empty())
      return std::nullopt;

```

io/io/src/IOUtils.cxx
```diff
@@ -0,0 +1,80 @@
...
+
+#ifdef R__FBSD
+#define getxattr(path, name, value, size) extattr_get_file(path, EXTATTR_NAMESPACE_USER, name, value, size)
+#endif
+
+#include "ROOT/StringUtils.hxx" // ROOT::StartsWith
+#include "TEnv.h"               // TEnv::GetValue
+#endif
+
+auto ROOT::Internal::GetXAttrSize(std::string_view path, std::string_view xattr)
```
pcanal:
```suggestion
auto ROOT::Internal::GetXAttrSize(... path, ... xattr)
```
Unfortunately the optimization provided by `std::string_view` is not helping here.  There is no guarantees from `std::string_view` is null terminated, so with the proposed interface, one need to first turn the argument into `std::string` before passing them to `getxattr`.

vepadulano:
To the best of my understanding, this PR is simply moving the code that used to live separately in `TFile::Open` and `RLoopManager::ChangeSpec` extracting it into a common place. I have been testing these changes on EOS. Both `TFile::Open` and `RNTupleReader::Open` work flawlessly. For some reason, with the redirection happening in RDataFrame to store also the redirected URL as a sample ID, I see this error
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Full sample identifier 'root://eoshome-v.cern.ch//eos/user/v/vpadulan/rdf_sample.root/events' cannot be found in the available samples.
```
which is the same that triggered  https://github.com/root-project/root/pull/17544 . What's even more worrying is that I don't see the error on every test execution, but sometimes the test also runs without failing. Before merging this PR I need to understand why that happens. It's possible the problem is still there somehow also without the code changes of this PR.

io/io/inc/ROOT/InternalIOUtils.hxx
```diff
@@ -0,0 +1,31 @@
...
+ * For the list of contributors see $ROOTSYS/README/CREDITS.             *
+ *************************************************************************/
+
+// Author: Vincenzo Eduardo Padulano (CERN), 11/2025
+
+#ifndef ROOT_IO_UTILS
+#define ROOT_IO_UTILS
+
+#include <string>
+#include <optional>
```
jblomer: Reorder alphabetically

io/io/src/InternalIOUtils.cxx
```diff
@@ -0,0 +1,69 @@
...
+            if (auto baseName = inputSV.substr(inputSV.find_last_of("/") + 1);
+                std::equal(baseName.crbegin(), baseName.crend(), xurl->crbegin())) {
+               // Ensure the redirected URL actually starts with the XRootD protocol string
+               if (ROOT::StartsWith(*xurl, "root://")) {
+                  return xurl;
+               }
+            }
+         }
+      }
+   }
```
jblomer: To increase readability, I'd reverse the if statements, like

```
if (gEnv->GetValue("TFile.CrossProtocolRedirects", 1) != 1)
   return std::nullopt;

auto xurl = ROOT::Internal::GetXAttrVal(inputURL, "eos.url.xroot");
if (!xurl)
   return std::nullopt;

...
```

io/io/src/InternalIOUtils.cxx
```diff
@@ -0,0 +1,69 @@
...
+
+   if (gEnv->GetValue("TFile.CrossProtocolRedirects", 1) == 1) {
+      if (auto xurl = ROOT::Internal::GetXAttrVal(inputURL, "eos.url.xroot")) {
+         // Sometimes the `getxattr` call may return an invalid URL due
+         // to the POSIX attribute not being yet completely filled by EOS.
+         if (inputSV.back() != '/') {
+            if (auto baseName = inputSV.substr(inputSV.find_last_of("/") + 1);
+                std::equal(baseName.crbegin(), baseName.crend(), xurl->crbegin())) {
+               // Ensure the redirected URL actually starts with the XRootD protocol string
+               if (ROOT::StartsWith(*xurl, "root://")) {
```
jblomer: The schema can also be `xroot://`

io/io/src/InternalIOUtils.cxx
```diff
@@ -0,0 +1,69 @@
...
+{
+#ifdef R__UNIX
+   std::string_view inputSV{inputURL};
+   if (inputSV.empty())
+      return std::nullopt;
+
+   if (gEnv->GetValue("TFile.CrossProtocolRedirects", 1) == 1) {
+      if (auto xurl = ROOT::Internal::GetXAttrVal(inputURL, "eos.url.xroot")) {
+         // Sometimes the `getxattr` call may return an invalid URL due
+         // to the POSIX attribute not being yet completely filled by EOS.
```
jblomer: Hm, that doesn't sound great. Is there a corresponding bug report for EOS?

jblomer (commented):
In principle looks good to me! We should think about how to test it. Extended attributes are file system dependent, so we may need to add some code just for testing, e.g. something that lets us control what `GetXAttrVal` returns under unit tests.

io/io/src/InternalIOUtils.cxx
vepadulano: Unfortunately not, I had reported this in the past, could be that this was addressed meanwhile, but I'm not sure. I would not tie this PR to that investigation.

pcanal: I thought, this was more or so a race condition :(

jblomer: I don't think we need to wait for the problem to be solved in order to merge this PR, but I would prefer to link the issue to a report on the EOS side. Could you ask or open an issue there?

Otherwise I'm afraid that the workaround will stay there forever.

vepadulano:
Regarding the issues with the `eos.url.xroot` attribute being wrong:

* https://github.com/cern-eos/eos/issues/41 describes the issue I remembered about the base name sometimes not being populated. I remember this still happening last year, can't say for sure at this point, but at some point it was there.
* https://github.com/cern-eos/eos/issues/50 describes a new regression that introduces double forward slash in the XRootD URL, thus breaking all versions of ROOT relying on this redirection

I would wait for the latter issue to also be confirmed fix before moving on with this PR, FYI @amadio.

jblomer:
@vepadulano it looks like this PR can advance

io/io/src/InternalIOUtils.cxx
```diff
@@ -0,0 +1,70 @@
...
+      return std::nullopt;
+
+   auto baseName = inputSV.substr(inputSV.find_last_of("/") + 1);
+   // Sometimes the `getxattr` call may return an invalid URL due
+   // to the POSIX attribute not being yet completely filled by EOS.
+   if (!std::equal(baseName.crbegin(), baseName.crend(), xurl->crbegin()))
+      return std::nullopt;
+
+   // Ensure the redirected URL actually starts with the XRootD protocol string
+   if (ROOT::StartsWith(*xurl, "root://") || ROOT::StartsWith(*xurl, "xroot://"))
```
amadio: You can safely remove this check. Alternatively, add also the TLS versions too : `roots://` and `xroots://`.

io/io/src/InternalIOUtils.cxx
```diff
@@ -0,0 +1,70 @@
...
+#endif
+
+std::optional<std::string> ROOT::Internal::GetXAttrVal(const char *path, const char *xattr)
+{
+#ifdef R__UNIX
+   // First call to getxattr evaluates the length of the extended attribute value
+   if (auto len = getxattr(path, xattr, nullptr, 0); len > 0) {
+      std::string xval(len, 0);
+      // Second call extracts the extended attribute value, checking it's of the correct length
+      if (getxattr(path, xattr, &xval[0], len) == len)
```
silverweed:
```suggestion
      if (getxattr(path, xattr, xval.data(), len) == len)
```

io/io/src/InternalIOUtils.cxx
```diff
@@ -0,0 +1,70 @@
...
+   (void)xattr;
+#endif
+   return std::nullopt;
+}
+
+std::optional<std::string> ROOT::Internal::GetEOSRedirectedXRootURL(const char *inputURL)
+{
+#ifdef R__UNIX
+   std::string_view inputSV{inputURL};
+   if (inputSV.empty() || inputSV.back() == '/')
```
silverweed: Since constructing a `string_view` with an invalid (including null) string is UB, I'd add:
```suggestion
   if (!inputURL) 
       return std::nullopt;
   
   std::string_view inputSV{inputURL};
   if (inputSV.empty() || inputSV.back() == '/')
```

vepadulano:
The two EOS issues I had mentioned previously seem to be fixed, although I am now looking at another failure that is happening. This is the ROOT reproducer
```cpp
#include <ROOT/RSnapshotOptions.hxx>

void write(const char *dataset, const char *path, const char *type) {
  ROOT::RDataFrame df{5};
  ROOT::RDF::RSnapshotOptions opts;
  if (strcmp(type, "ttree") == 0)
    opts.fOutputFormat = ROOT::RDF::ESnapshotOutputFormat::kTTree;
  else
    opts.fOutputFormat = ROOT::RDF::ESnapshotOutputFormat::kRNTuple;
  df.Define("x", []() { return 42; }).Snapshot(dataset, path, "", opts);
}

void test_rntuplereader_open() {
  write("events", "anotherfile.root", "rntuple");
  write("events", "rntuplereader_myfile.root", "rntuple");
  auto r = ROOT::RNTupleReader::Open("events", "rntuplereader_myfile.root");
}

int main() {
  test_rntuplereader_open();
}
```

The logic is fairly simple:

* Write a file named "anotherfile.root" with an RNTuple inside called "events"
* Write a file named "rntuplereader_myfile.root" with an RNTuple inside called "events"
* Open the second file with RNTupleReader

It looks like writing two files one after another is causing initially some corruption in the file being read, then on subsequent runs of the same application shows an error message about not having permission to write to the first file

```
[vpadulan@lxplus9104 /eos/user/v/vpadulan/xrootd_redirection_tests]$./xrootd_redirection_rntuple.out 
  what():  invalid read (expected bytes: 4, read: 0)
At:
  ROOT::Internal::RMiniFileReader::TryReadBuffer(void*, size_t, uint64_t)::<lambda(std::size_t, std::size_t)> [/tmp/vpadulan/rootproject/root/tree/ntuple/src/RMiniFile.cxx:895]

Aborted (core dumped)
[vpadulan@lxplus9104 /eos/user/v/vpadulan/xrootd_redirection_tests]$./xrootd_redirection_rntuple.out 
  what():  The file 'root://eoshome-v.cern.ch//eos/user/v/vpadulan/xrootd_redirection_tests/anotherfile.root' given to RNTupleParallelWriter is not writable. Open it with 'UPDATE' or 'RECREATE' if you want to write into it.
At:
  static std::unique_ptr<ROOT::RNTupleParallelWriter> ROOT::RNTupleParallelWriter::Append(std::unique_ptr<ROOT::RNTupleModel>, std::string_view, TDirectory&, const ROOT::RNTupleWriteOptions&) [/tmp/vpadulan/rootproject/root/tree/ntuple/src/RNTupleParallelWriter.cxx:196]

Aborted (core dumped)
[vpadulan@lxplus9104 /eos/user/v/vpadulan/xrootd_redirection_tests]$./xrootd_redirection_rntuple.out 
  what():  The file 'root://eoshome-v.cern.ch//eos/user/v/vpadulan/xrootd_redirection_tests/anotherfile.root' given to RNTupleParallelWriter is not writable. Open it with 'UPDATE' or 'RECREATE' if you want to write into it.
At:
  static std::unique_ptr<ROOT::RNTupleParallelWriter> ROOT::RNTupleParallelWriter::Append(std::unique_ptr<ROOT::RNTupleModel>, std::string_view, TDirectory&, const ROOT::RNTupleWriteOptions&) [/tmp/vpadulan/rootproject/root/tree/ntuple/src/RNTupleParallelWriter.cxx:196]

Aborted (core dumped)
```


I am at the moment trying to see if I can create an EOS-only reproducer without ROOT.

vepadulano:
The error from the parallel writer about the file not being writable is due to TNetXNGFile not properly sanitising the input mode option "RECREATE_WITHOUT_GLOBALREGISTRATION" which is then parsed to the appropriate XRootD file open mode. Since it's not recognised, the file is opened in read only mode https://github.com/root-project/root/blob/63097c0dd7836be079a07291cb50df85e0ceb0ec/net/netxng/src/TNetXNGFile.cxx#L180 . The patch is straightforward in this case.

The other issue about the invalid read from RMiniFileReader is still unsolved.

vepadulano:
The invalid read is caused by not closing the output TFile in `UntypedSnapshotRNTupleHelper::Finalize` https://github.com/root-project/root/blob/63097c0dd7836be079a07291cb50df85e0ceb0ec/tree/dataframe/src/RDFSnapshotHelpers.cxx#L967 .

vepadulano:
Should be ready for re-review now :+1:

io/io/res/ROOT/InternalIOUtils.hxx
hageboeck, line 1: Could this be `io/io/res`, so during build, the header can be used by dependent source files, but we would never install it?

io/io/CMakeLists.txt
```diff
@@ -79,6 +80,7 @@ set_source_files_properties(src/RConcurrentHashColl.cxx
   PROPERTIES COMPILE_FLAGS -I${CMAKE_SOURCE_DIR}/core/foundation/res)
 
 ROOT_GENERATE_DICTIONARY(G__RIO
+  ROOT/InternalIOUtils.hxx
```
hageboeck: If the header goes to `res/`, this is hopefully unnecessary.

hageboeck (commented):
This looks good to me, but I would try to make this header completely invisible outside of ROOT.

amadio (commented):
Looks good! Thanks a lot for making this work with RNTuple too!

io/io/src/InternalIOUtils.cxx
```diff
@@ -0,0 +1,66 @@
...
+#include "ROOT/StringUtils.hxx" // ROOT::StartsWith
+#include "TEnv.h"               // TEnv::GetValue
+#endif
+
+std::optional<std::string>
+ROOT::Internal::GetXAttrVal([[maybe_unused]] std::string_view path, [[maybe_unused]] std::string_view xattr)
+{
+#ifdef R__UNIX
+   // First call to getxattr evaluates the length of the extended attribute value
+   if (auto len = getxattr(path.data(), xattr.data(), nullptr, 0); len > 0) {
```
jblomer: I think in principle you can have an extended attribute value which is the empty string.
```suggestion
   if (auto len = getxattr(path.data(), xattr.data(), nullptr, 0); len >= 0) {
```

io/io/res/ROOT/InternalIOUtils.hxx
```diff
@@ -0,0 +1,24 @@
...
+/// \brief Get extended attribute value from path
+/// \param path Path to the file to check
+/// \param xattr Extended attribute to evaluate
+/// \return The string containing the extended attribute value if found, std::nullopt otherwise
+std::optional<std::string> GetXAttrVal(std::string_view path, std::string_view xattr);
+
+/// \brief Redirects the input URL to the equivalent XRootD path on EOS
+/// \param inputUrl The input URL to redirect
+/// \return The redirected URL in case of successful redirection, std::nullopt otherwise
+std::optional<std::string> GetEOSRedirectedXRootURL(std::string_view inputURL);
```
jblomer: We should assume that the input is a local path, not a URL.
```suggestion
std::optional<std::string> GetEOSRedirectedXRootURL(std::string_view inputPath);
```

io/io/src/InternalIOUtils.cxx
```diff
@@ -0,0 +1,66 @@
...
+      return std::nullopt;
+
+   auto xurl = ROOT::Internal::GetXAttrVal(inputURL, "eos.url.xroot");
+   if (!xurl)
+      return std::nullopt;
+
+   auto baseName = inputURL.substr(inputURL.find_last_of("/") + 1);
+   // Sometimes the `getxattr` call may return an invalid URL due
+   // to the POSIX attribute not being yet completely filled by EOS.
+   if (!std::equal(baseName.crbegin(), baseName.crend(), xurl->crbegin()))
```
jblomer: We probably want to warn in this case.

io/io/src/InternalIOUtils.cxx
```diff
@@ -0,0 +1,66 @@
...
+      return std::nullopt;
+
+   // Ensure the redirected URL actually starts with the XRootD protocol string
+   if (ROOT::StartsWith(*xurl, "root://") || ROOT::StartsWith(*xurl, "xroot://") ||
```
jblomer, line 73: Also here, the other branch should warn.

io/io/CMakeLists.txt
```diff
@@ -66,6 +67,7 @@ ROOT_LINKER_LIBRARY(RIO
 )
 
 target_include_directories(RIO PRIVATE ${CMAKE_SOURCE_DIR}/core/clib/res)
+target_include_directories(RIO PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/res>)
```
jblomer, line 71: Why not

```suggestion
target_include_directories(RIO PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/res>)
```

jblomer (approved):
Apologies for having let this go abandoned!

io/io/res/ROOT/InternalIOUtils.hxx
vepadulano: RRawFile::Create does not make that assumption, see https://github.com/root-project/root/blob/3d1f57342af5d8eadc70ac235e36be9a30999e9d/io/io/src/RRawFile.cxx#L64

vepadulano: Although maybe you mean that this specific function is only meant to work for the POSIX path to XRootD URL redirection feature on EOS, so here we can/should assume that the input string is a path to a file

io/io/CMakeLists.txt
vepadulano: So that dependent libraries may find the header too

PR #20273, opened 2025-11-03
[io] TFile: allow using cache with READ_WITHOUT_GLOBALREGISTRATION

### LINKED ISSUE #15028: [RDataFrame] Unable to cacheread remote file
### Check duplicate issues.

- [X] Checked for duplicates

### Description

When input files to `RDataFrame` are remote files, force caching of remote files does not work and the remote files will be downloaded every time. 

### Reproducer

```
import os
import ROOT

user = os.environ['USER']
outdir = f"/eos/user/{user[0]}/{user}"
filename = os.path.join(outdir, "test.root")
# create dummy root file
ROOT.RDataFrame(100).Define("x", "1").Snapshot("test", filename)

ROOT.TFile.SetCacheFileDir("/tmp", True, True)
# this does not trigger loading of cached root file
ROOT.RDataFrame("test", f"root://eosuser.cern.ch/{filename}").Sum("x").GetValue()
```

This is because internally RDataFrame will create a TChain using [`ROOT.Internal.TreeUtils.MakeChainForMT(treename)`](https://root.cern/doc/master/RLoopManager_8cxx_source.html#l01169), which creates a `TChain` object with the mode `ROOT.TChain.kWithoutGlobalRegistration`. This in turn forces the TFile open option to be ["READ_WITHOUT_GLOBALREGISTRATION"](https://root.cern/doc/master/TChain_8cxx_source.html#l00505). This causes the TFile to be opened without caching since it only checks the [fgCacheFileForce](https://root.cern/doc/master/classTFile.html#a04d17cde8f18d500653077fc7d7dc3ac) flag when option is ["READ"](https://root.cern/doc/master/TFile_8cxx_source.html#l04182)

### ROOT version

6.30/04 (LCG105a)

### Installation method

LCG (Swan)

### Operating system

Linux

### Additional context

_No response_

silverweed:
`SetCacheFileDir` was ignored if the file is opened without globalregistration (e.g. by RDF). With this PR we cover that case as well.

## Checklist:

- [x] tested changes locally

This PR fixes #15028

### DIFF
--- a/io/io/test/TFileTests.cxx
+++ b/io/io/test/TFileTests.cxx
@@ -141,6 +141,15 @@ TEST(TFile, ReadWithoutGlobalRegistrationNet)
    const auto netFile = "root://eospublic.cern.ch//eos/root-eos/h1/dstarmb.root";
    TestReadWithoutGlobalRegistrationIfPossible(netFile);
 }
+TEST(TFile, ReadWithCacheWithoutGlobalRegistration)
+{
+   const auto webFile = "http://root.cern/files/h1/dstarmb.root";
+   TFile::SetCacheFileDir(".");
+   delete TFile::Open(webFile, "READ_WITHOUT_GLOBALREGISTRATION");
+   EXPECT_TRUE(gSystem->AccessPathName("./files/h1/dstarmb.root"));
+   TFile::SetCacheFileDir("");
+   gSystem->Unlink("./files");
+}
 #endif
 #endif
 
@@ -295,3 +304,4 @@ TEST(TFile, PersistTObjectStdArray)
    }
    gSystem->Unlink(filename);
 }
+


io/io/src/TFile.cxx
```diff
@@ -3883,7 +3883,8 @@ TFile *TFile::Open(const char *url, Option_t *options, const char *ftitle,
 
       // check if we read through a file cache
       if (!strcasecmp(option, "CACHEREAD") ||
-         ((!strcasecmp(option,"READ") || !option[0]) && fgCacheFileForce)) {
+          ((!strcasecmp(option, "READ") || !strcasecmp(option, "READ_WITHOUT_GLOBALREGISTRATION") || !option[0]) &&
```
pcanal, line 3886: Side note: this new code is not tested ... but `CACHEREAD` is also not explicitly tested either.

silverweed:
Added a simple test

PR #20276, opened 2025-11-03
[RF][HS3] Clean Expressions in RooGenPdf and RooFormularVar

Phmonski:
# This Pull request:
- Changes to Expressions to exclude TMath:: 
- Code clean-up
- Include RooResolutionModels to nameSanitization

# Major BugFix:
- Data axes changed due to sorting variable names when reading .json Files

## Changes or fixes:

This PR fixes HS3 Issue: https://github.com/scipp-atlas/pyhs3/issues/69#issuecomment-3081740379

### DIFF
--- a/roofit/hs3/src/RooJSONFactoryWSTool.cxx
+++ b/roofit/hs3/src/RooJSONFactoryWSTool.cxx
@@ -483,18 +483,13 @@ void exportAttributes(const RooAbsArg *arg, JSONNode &rootnode)
  *
  * @param ws The RooWorkspace in which the observables will be created.
  * @param node The JSONNode containing information about the observables to be created.
- * @param out The RooArgSet to which the created observables will be added.
+ * @param out The RooAbsCollection to which the created observables will be added.
  * @return void
  */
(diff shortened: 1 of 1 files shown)


roofit/hs3/src/JSONFactories_RooFitCore.cxx
```diff
@@ -701,6 +684,19 @@ class RooFormulaArgStreamer : public RooFit::JSONIO::Exporter {
       elem["expression"] << expression.Data();
       return true;
    }
+
+private:
+   void cleanExpression(TString &expr) const
+   {
+      expr.ReplaceAll("TMath::Exp", "exp");
```
ferdymercury, line 691: maybe std::exp and so on?

cburgard: No, this is intentional:
https://hep-statistics-serialization-standard.github.io/#sec:generic_expression

If we really want "std::exp" for the JIT, the addition of the "std::" prefix should happen during re-import.

guitargeek (approved):
Thank you very much!

PR #20277, opened 2025-11-03
[hist] prevent Unzoom from Zooming in

ferdymercury:
# This Pull request:

## Changes or fixes:

Default drawing histogram option is to leave a YMARGIN range of 10% in YMAX. So if you call Unzoom, it was calling GetMaximum, which calculated the exact maximum, and then called SetMaximum, thus overwriting the stored value. Use instead GetMaximumStored to prevent recalculation if not set by user.

Also add sanity check before casting

Reproducer:

```
TH1F h("h", "", 100, 0, 10);
h->Fill(2);
h.Draw();
h.GetYaxis()->UnZoom(); // This zooms in rather than out.
```

## Checklist:

- [x] tested changes locally

### DIFF
--- a/hist/hist/src/TAxis.cxx
+++ b/hist/hist/src/TAxis.cxx
@@ -1268,7 +1268,7 @@ void TAxis::UnZoom()
 
    //unzoom object owning this axis
    SetRange(0,0);
-   TH1 *hobj1 = GetParent() && GetParent()->InheritsFrom(TH1::Class()) ? static_cast<TH1 *>(GetParent()) : nullptr;
+   TH1 *hobj1 = dynamic_cast<TH1 *>(GetParent());
    if (!strstr(GetName(),"xaxis")) {
       if (!hobj1) return;
       if (hobj1->GetDimension() == 2) {


couet:
_Default drawing histogram option is to leave a YMARGIN range of 10% in YMAX._

it is not a fixed vaule of 10%: it can be set via TStyle::SetHisTopMargin and it can be 0.

ferdymercury:
> _Default drawing histogram option is to leave a YMARGIN range of 10% in YMAX._
> 
> it is not a fixed vaule of 10%: it can be set via TStyle::SetHisTopMargin and it can be 0.

Thanks for the clarification. This does not change the underlying issue, the correct wording would be:

Default drawing histogram option is to leave a margin of X% in YMAX, where X is defined by TStyle::SetHistTopMargin.

When one clicks on "Unzoom", the x axis is "zoomed in", rather than 'kept as is' or zoomed out, which is quite counterintuitive.

hist/hist/src/TAxis.cxx
```diff
@@ -1268,7 +1268,7 @@ void TAxis::UnZoom()
 
    //unzoom object owning this axis
    SetRange(0,0);
-   TH1 *hobj1 = (TH1*)GetParent();
+   TH1 *hobj1 = GetParent() && GetParent()->InheritsFrom(TH1::Class()) ? static_cast<TH1 *>(GetParent()) : nullptr;
```
silverweed: Why not simply 
```c++
TH1 *hobj1 = dynamic_cast<TH1*>(GetParent());
```
?

guitargeek (approved):
Thanks! The comment by @silverweed was addressed, and @couet already approved it. So I see no reason to have this PR on hold

PR #20278, opened 2025-11-03
Store typecode in `TLeaf`

tomeichlersmith:
This PR is originating from my solution to my ROOT Forum post about slow cloning while performing schema evolution: https://root-forum.cern.ch/t/quiet-writing-failure-when-applying-schema-evolution-while-slow-cloning/64350/11

I noticed while developing that solution that I could not figure out how to access a `TLeaf`'s typecode without (a) using a map between the type's full name and the single-character type code or (b) a long if/else tree checking if my `TLeaf` is one of the many different implementations for the different types.

While either of these solutions could work, I would prefer to simply store the single-character typecode in the `TLeaf` base class so that code like mine can be simplified and more maintainable.

If this type of no-issue change is unacceptable, feel free to drop it. I am not overly committed to it.

### DIFF
--- a/tree/tree/src/TLeaf.cxx
+++ b/tree/tree/src/TLeaf.cxx
@@ -73,7 +73,7 @@ TLeaf::TLeaf()
 
 TLeaf::TLeaf(TBranch *parent, const char* name, const char *type)
    : TNamed(name, name)
-   , fTypeCode(type[0])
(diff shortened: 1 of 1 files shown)


pcanal:
Just to note that we are looking at this PR.  We are investigating whether alternative that do not require an additional data member would be viable.

PR #20284, opened 2025-11-04
[skip-ci][DF] Add translations of TTree::Draw to RDataFrame

martamaja10:
First attempt of adding some "Rosetta stone" table. Open to discussion what else to include already now and what should be removed (note the last two rows of the table). The location of this table can also be modified. 

Screenshots updated after second round of fixes (post Stephan's and Vincenzo's review) 

Screenshot 2025-11-06 at 11 30 12

### DIFF
--- a/tree/dataframe/src/RDataFrame.cxx
+++ b/tree/dataframe/src/RDataFrame.cxx
@@ -264,7 +264,7 @@ df.Filter([] (ULong64_t e) { return e == 1; }, {"event"}).Histo1D<RVec<float>>("
    <td>
 ~~~{cpp}
 // object selection: for each event, fill histogram with array of selected pts
-tree->Draw('Muon_pt', 'Muon_pt > 100')
+tree->Draw('Muon_pt', 'Muon_pt > 100');
 ~~~
    </td>
    <td>
@@ -1915,6 +1915,178 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
 \code{.py}
 df.Filter("std::isfinite(x)").Mean("x")
 \endcode
+
+### Translating TTree::Draw to RDataFrame 
+
+<table>
+<tr>
+   <td>
+      <b>TTree::Draw</b>
+   </td>
+   <td>
+      <b>ROOT::RDataFrame</b>
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{.cpp}
+// Get the tree and Draw a histogram of x for selected y values
+auto *tree = file->Get<TTree>("myTree");
+tree->Draw("x", "y > 2");
+~~~
+   </td>
+   <td>
+~~~{.cpp}
+ROOT::RDataFrame df("myTree", file);
+df.Filter("y > 2").Histo1D("x")->Draw();
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{.cpp}
+// Draw a histogram of "jet_eta" with the desired weight
+tree->Draw("jet_eta", "weight*(event == 1)");
+~~~
+   </td>
+   <td>
+~~~{.cpp}
+df.Filter("event == 1").Histo1D("jet_eta", "weight")->Draw();
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Draw a histogram filled with values resulting from calling a method of the class of the `event` branch in the TTree.
+tree->Draw("event.GetNtrack()");
+
+~~~
+   </td>
+   <td>
+~~~{cpp}
+auto df1 = df.Define("NTrack","event.GetNtrack()");
+df1.Histo1D("NTrack")->Draw();
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Draw only every 10th event
+tree->Draw("fNtrack","fEvtHdr.fEvtNum%10 == 0");
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// Use the Filter operation together with the special RDF column: `rdfentry_`
+df.Filter("rdfentry_ % 10 == 0").Histo1D("fNtrack")->Draw();
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// object selection: for each event, fill histogram with array of selected pts
+tree->Draw('Muon_pt', 'Muon_pt > 100');
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// with RDF, arrays are read as ROOT::VecOps::RVec objects
+df.Define("good_pt", "Muon_pt[Muon_pt > 100]").Histo1D("good_pt")->Draw();
+~~~
+   </td>
+</tr>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Draw the histogram and fill hnew with it
+tree->Draw("sqrt(x)>>hnew","y>0"); 
+
+// Retrieve hnew from the current directory
+TH1F *hnew = (TH1F*)gDirectory->Get("hnew");
+
+// Retrieve hnew from the current Pad
+TH1F *hnew = (TH1F*)gPad->GetPrimitive("hnew");
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// We pass histogram constructor arguments to the Histo1D operation, to easily give the histogram a name
+auto hist = df.Filter("y>0").Histo1D({"hnew","hnew",10, 0, 10},"x");
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Draw a 1D Profile histogram instead of TH2F
+tree->Draw("y:x","","prof");
+
+// Draw a 2D Profile histogram instead of TH3F
+tree->Draw("z:y:x","","prof");
+~~~
+   </td>
+   <td>
+~~~{cpp}
+
+// Draw a 1D Profile histogram
+auto profile1D = df.Profile1D("x", "y");
+
+// Draw a 2D Profile histogram
+auto profile2D = df.Profile2D("x", "y", "z");
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// This command draws 2 entries starting with entry 5
+tree->Draw("x", "","", 2, 5);
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// Range function with arguments begin, end
+auto histo_range = df.Range(5,7).Histo1D<int>("x");
+histo_range->Draw();
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Draw the X() component of the 
+// ROOT::Math::DisplacementVector3D in vec_list
+tree->Draw("vec_list.X()");
+~~~
+   </td>
+   <td>
+~~~{cpp}
+auto histo = df.Define("x", "ROOT::RVecD out; for(const auto &el: vec_list) out.push_back(el.X()); return out;").Histo1D("x");
+histo->Draw();
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Gather all values from a branch holding a collection per event, `pt`, 
+// and fill a histogram so that we can count the total number of values across all events 
+tree->Draw("pt>>histo");
+TH1D *histo = (TH1D *)gDirectory->Get("histo");
+histo->GetEntries();
+~~~
+   </td>
+   <td>
+~~~{cpp}
+df.Histo1D("pt")->GetEntries();
+~~~
+   </td>
+</tr>
+</table>
 
 */
 // clang-format on


tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,264 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+   <td>
+~~~{cpp}
+// Draw only every 10th event
+tree->Draw("fNtrack","fEvtHdr.fEvtNum%10 == 0")
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// Use the Filter operation together with the special RDF column: `rdfentry_`
+auto histo = df.Filter("rdfentry_ % 10 == 0").Histo1D<int>("fNtrack")
```
```suggestion
auto histo = df.Filter("rdfentry_ % 10 == 0").Histo1D<int>("fNtrack");
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,264 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+// note: this will not work in RDataFrame version
+// tree->Draw("GetNtrack()");
+
+~~~
+   </td>
+   <td>
+~~~{cpp}
+auto df1 = df.Define("NTrack","event.GetNtrack()")
+// note: auto df2 = df.Define("ntrack","GetNtrack()") will not work 
+auto df2 = df1.Histo1D<int>("NTrack")
```
```suggestion
auto df1 = df.Define("NTrack","event.GetNtrack()")
             .Histo1D<int>("NTrack")
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,264 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+// Retrieve hnew from the current Pad
+TH1F *hnew = (TH1F*)gPad->GetPrimitive("hnew")
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// Save the histogram to a file
+TFile f("myhistos.root","new")
+auto hist = df.Filter("y>0").Histo1D<float>({"myhisto","myhisto",10, 0, 10},"x")
+hist->Write()
```
hageboeck: Maybe we could already use ROOT-7 syntax:
```suggestion
f.WriteTObject(hist.GetPtr());
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,264 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+
+// Retrieve hnew from the current Pad
+TH1F *hnew = (TH1F*)gPad->GetPrimitive("hnew")
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// Save the histogram to a file
+TFile f("myhistos.root","new")
+auto hist = df.Filter("y>0").Histo1D<float>({"myhisto","myhisto",10, 0, 10},"x")
```
```suggestion
TFile f("myhistos.root","new");
auto hist = df.Filter("y>0").Histo1D<float>({"myhisto","myhisto",10, 0, 10},"x");
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,264 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+tree->Draw("z:y:x","","prof")
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// Draw a 1D Profile histogram
+auto profile1D = df.Profile1D<int, float, double>({"profName", "profTitle", 10, 0, 10}, "x", "y");
+
+// Draw a 2D Profile histogram
+auto profile2D = df.Profile2D<int, float, double, int>({"profName", "profTitle", 10u, 0, 10, 10, 0, 10}, "x", "y", "z");
```
hageboeck: Are the template parameters correct? It may work without, but if we name them, don't you only need to name them for the columns being passed? Since there is no weight, 1D should have two, and 2D three template parameters, unless I'm overlooking something.

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,264 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+// This command draws the maximum value of the X-axis for each histogram.
+tree->Draw("fH.fXaxis.fXmax"); // fH is of type TH1F stored in class Event 
+tree->Draw("fH.fXaxis.GetXmax()"); // fH is of type TH1F stored in class Event 
+tree->Draw("fH.GetXaxis().fXmax"); //fH is of type TH1F stored in class Event 
+tree->Draw("GetHistogram().GetXaxis().GetXmax()"); 
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// RDF ?
```
hageboeck: I would say we don't mention this. Histograms in trees or RNTuples are hopefully niche enough that those users would know how to write their own action.

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,264 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+   <td>
+~~~{cpp}
+// RDF ? 
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Special functions and variables
```
hageboeck: I would say we split this off in another PR, so that this one can go in faster.

hageboeck (approved):
Very nice, thanks!

silverweed:
Would it make sense to have 3 columns on the table: one with TTree::Draw, one with "RDF one-liner" (interpreted, as few intermediate variables as possible) and one with "RDF fully compiled", with proper lambdas/types and not as brief as the second column?

tree/dataframe/src/RDataFrame.cxx
martamaja10: you're right, the 4th here or the 3rd in 1D profile are for the weights (it should also work without the template parameters)

martamaja10:
> Would it make sense to have 3 columns on the table: one with TTree::Draw, one with "RDF one-liner" (interpreted, as few intermediate variables as possible) and one with "RDF fully compiled", with proper lambdas/types and not as brief as the second column?

Thanks for this comment Jack! You're right, maybe having the minimal version here could be better and then showing the fully compiled version separately would also make sense. What I fear with 3 columns here is that maybe it will make it hard to read/comprehend quickly. Maybe this PR should focus on the simplest translation and then we add on top of it later? What do you think @vepadulano @hageboeck ?

hageboeck:
> Maybe this PR should focus on the simplest translation and then we add on top of it later? What do you think vepadulano hageboeck ?

We could surely merge it now with these two columns, wait one night for it to show up in the docs, and subsequently test if the space allows us to have the third column.
Vincenzo had the same question for a Python column, so we would arrive already at 4 columns.

Maybe one could make a table with the most concise one-liners (actually, one statement), in C++ and Python. And then we could think about a second table with the "expert" fully compiled versions?
The reason I make a distinction between one-liner and one statement is that this:
```c++
auto histo = df.Filter(...)
               .Define(....)
               .Histo1D(...);
```
seems easier to present and read than:
```c++
auto histo = df.Filter(...).Define(....).Histo1D(...);
```

silverweed:
Let's not forget that if the table is meant to be viewed on a webpage we can also have it dynamic (keep it with 2 columns but let the user decide which "flavor" of RDF is displayed with a dropdown menu or similar) - similarly to what e.g. Windows does for the [WPF docs](https://learn.microsoft.com/en-us/dotnet/api/system.windows.window?view=windowsdesktop-9.0) (notice the little dropdown menu that lets you select the programming language)

ferdymercury:
Thanks for this!

It's not for TTree::Draw but it would be nice to also document a translation of:

```
tree->SetBranchAddress("time", &time);
auto prevTime = 0;
for(Long64_t i = 0; i<nEntries; ++i) {
    tree->GetEntry(i);
    hist->Fill(time - prevTime);
    prevTime = time;
}
```

(if a histogram of time differences of consecutive entries can be achieved at all with non-multithreaded RDF)

hageboeck:
> Thanks for this!
> 
> It's not for TTree::Draw but it would be nice to also document a translation of:
> 
> ```
> tree->SetBranchAddress("time", &time);
> auto prevTime = 0;
> for(Long64_t i = 0; i<nEntries; ++i) {
>     tree->GetEntry(i);
>     hist->Fill(time - prevTime);
>     prevTime = time;
> }
> ```
> 
> (if a histogram of time differences of consecutive entries can be achieved at all with non-multithreaded RDF)

Yes, you can, but you need a mutable lambda:
```c++
  ROOT::RDataFrame rdf(10);
  unsigned long long old = 0;
  auto histo = rdf.Define("timediff", [old = 0](unsigned long long entry) mutable {
      const auto result = entry - old;
      old = entry;
      return result;
      }, {"rdfentry_"}).Histo1D({"name", "title", 10, 0, 10}, "timediff");
```
image

vepadulano:
> Maybe this PR should focus on the simplest translation and then we add on top of it later? What do you think @vepadulano @hageboeck ?

I believe this PR should show the simplest/shortest way to represent in RDF an equivalent TTree::Draw query. Doing it in C++ or Python would be equivalent, it's fine to start with C++. Probably the string-expressions are the shortest way of representing the TTree::Draw queries in RDF (being those queries string expressions in the first place). If in any case having a C++ lambda makes the code shorter, then it's also fine to use that.

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,211 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+// Get the tree and Draw a histogram of x for selected y values
+auto *tree = file->Get<TTree>("myTree");
+tree->Draw("x", "y > 2");
+~~~
+   </td>
+   <td>
+~~~{.cpp}
+ROOT::RDataFrame df("myTree", file);
+auto h = df.Filter("y > 2").Histo1D("x");
+h->Draw();
```
vepadulano: Here I would even go one step further
```suggestion
ROOT::RDataFrame df("myTree", file);
df.Filter("y > 2").Histo1D("x")->Draw();
```

My reasoning is: for TTree::Draw, we're showing two lines: one is to get the "executor" (a TTree), the second to draw a histogram. With my proposed change, we have the semantic equivalent in the first line and in the second line.

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,211 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+// Draw a histogram of "jet_eta" with the desired weight
+tree->Draw("jet_eta", "weight*(event == 1)");
+~~~
+   </td>
+   <td>
+~~~{.cpp}
+df.Filter("event == 1").Histo1D("jet_eta", "weight");
+
+// or the fully compiled version:
+df.Filter([] (ULong64_t e) { return e == 1; }, {"event"}).Histo1D<RVec<float>>("jet_eta", "weight");
```
vepadulano: I would not mix the two versions in this first table, possibly we'll make it in a separate column in a followup PR

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,211 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Draw a histogram using the method of the class that is being used. 
+tree->Draw("event.GetNtrack()");
+// note: this will not work in RDataFrame version
+// tree->Draw("GetNtrack()");
```
vepadulano: Here I'm losing a bit the context in the second example. What is "the class that is being used?". I am assuming that `event` is the name of a physical TTree branch on disk. In which case, I would understand that `GetNtrack()` alone wouldn't work in RDF (I'm surprised it would work in TTree in the first place, what if I have another branch called `event2` that also has `GetNtrack()`?)

vepadulano: If `event` is a name of a physical branch on disk, I would specify that both here and in the RDF example

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,211 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+
+~~~
+   </td>
+   <td>
+~~~{cpp}
+auto df1 = df.Define("NTrack","event.GetNtrack()");
+auto df2 = df1.Histo1D<int>("NTrack");
```
vepadulano: If we're going for the shortest amount of characters possible:
```suggestion
auto df2 = df1.Histo1D("NTrack");
```

And we also need to Draw
```suggestion
df1.Histo1D("NTrack")->Draw();
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,211 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+   <td>
+~~~{cpp}
+// object selection: for each event, fill histogram with array of selected pts
+tree->Draw('Muon_pt', 'Muon_pt > 100')
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// with RDF, arrays are read as ROOT::VecOps::RVec objects
+df.Define("good_pt", "Muon_pt[Muon_pt > 100]").Histo1D<float>("good_pt");
```
vepadulano:
```suggestion
df.Define("good_pt", "Muon_pt[Muon_pt > 100]").Histo1D("good_pt");
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,211 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+   <td>
+~~~{cpp}
+// Draw only every 10th event
+tree->Draw("fNtrack","fEvtHdr.fEvtNum%10 == 0");
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// Use the Filter operation together with the special RDF column: `rdfentry_`
+auto histo = df.Filter("rdfentry_ % 10 == 0").Histo1D<int>("fNtrack");
```
vepadulano:
```suggestion
auto histo = df.Filter("rdfentry_ % 10 == 0").Histo1D("fNtrack");
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,211 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+   <td>
+~~~{cpp}
+// Save the histogram to a file
+TFile f("myhistos.root","new");
+auto hist = df.Filter("y>0").Histo1D<float>({"myhisto","myhisto",10, 0, 10},"x");
+f.WriteTObject(hist.GetPtr());
+
+// Retrieve the histogram
+TFile f("myhistos.root");
+TH1D *h = (TH1D*)f.Get("myhisto");
```
vepadulano: I think here we're overcomplicating things. The original TTree::Draw example doesn't make any assumptions about writing the histogram to disk. In this case, the equivalent RDF code is much more concise and elegant (on top of being more correct, which is always true when compared to TTree::Draw).


```suggestion
// We pass histogram constructor arguments to the Histo1D operation, to easily give the histogram a name
auto hist = df.Filter("y>0").Histo1D({"hnew","hnew",10, 0, 10},"x");
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,211 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+tree->Draw("z:y:x","","prof");
+~~~
+   </td>
+   <td>
+~~~{cpp}
+// Draw a 1D Profile histogram
+auto profile1D = df.Profile1D<int, float>({"profName", "profTitle", 10, 0, 10}, "x", "y");
+
+// Draw a 2D Profile histogram
+auto profile2D = df.Profile2D<int, float, double>({"profName", "profTitle", 10u, 0, 10, 10, 0, 10}, "x", "y", "z");
```
vepadulano: Reducing to the shortest amount of characters

```suggestion
// Draw a 1D Profile histogram
auto profile1D = df.Profile1D("x", "y");

// Draw a 2D Profile histogram
auto profile2D = df.Profile2D("x", "y", "z");
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,211 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+// Draw the X() component of the 
+// ROOT::Math::DisplacementVector3D in vec_list
+tree->Draw("vec_list.X()");
+~~~
+   </td>
+   <td>
+~~~{cpp}
+auto getX = [](vector< ROOT::Math::DisplacementVector3D> &v){RVec<double> out; for( auto p: v) out.push_back(p.X()); return out;};
+auto histo = df.Define("X",getX,{"vec_list"}).Histo1D<double>("x");
+histo->Draw();
```
vepadulano: Shortest

```suggestion
auto histo = df.Define("x",R"(
ROOT::RVecD out;
for (const auto &el: vec_list)
    out.push_back(el.X());
return out;
)").Histo1D("x");
histo->Draw();
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,211 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+// and so on: 
+// fMatrix[1][0], fMatrix[2][0], fMatrix[3][0], ... , fMatrix[n][0]
+
+~~~
+   </td>
+   <td>
+~~~{cpp}
+Reading multi-dimensional arrays is not supported in RDF. 
+GH issue exists. If such feature is needed, please add a comment: "https://github.com/root-project/root/issues/15068"
+~~~
```
vepadulano: As discussed offline, this list of examples can be removed for this PR

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,211 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Count number of entries in a specific dataset branch
+tree->Draw("pt"); 
+TH1* myH {static_cast<TH1*>(gPad->GetPrimitive(\"htemp\"))}; 
+myH->GetEntries();
```
vepadulano: How can the number of entries in a specific dataset branch be any different from the number of entries in any other branch? There are only two ways I can imagine:

* This particular branch of the computation graph has seen some filtering. This option is not represented here, because there is no cut expression passed to TTree::Draw
* This particular branch `pt` is actually a collection of `pt` from various particles in the event. From this TTree::Draw string I'm also missing that, or maybe it's missing from the description.

Without a clearer explanation here I think we cannot make the RDataFrame example meaningful, what I see right now is the equivalent of
```
ROOT::RDataFrame df{};
auto count = *df.Count();
```

vepadulano (changes requested):
Thank you so much! This is great! I have left a few comments

tree/dataframe/src/RDataFrame.cxx
hageboeck: > what if I have another branch called `event2` that also has `GetNtrack()`?)

I guess that why we shouldn't show it 🙂. Let's maybe stick with `event.GetNtrack()`, and show the RDF equivalent?

tree/dataframe/src/RDataFrame.cxx
hageboeck: It's the latter option. Collections are implicitly unrolled. The RDF example therefore has to do that, too, but that's the case when histogramming the column.

tree/dataframe/src/RDataFrame.cxx
martamaja10: yes, you're right - I tried translating some of these examples: https://root.cern.ch/root/htmldoc/guides/users-guide/Trees.html#trees-in-analysis (TTree::Draw Examples). I will just remove the 2nd example, as it was there just to be very honest with the reference examples but without this context, it indeed is more confusing.

tree/dataframe/src/RDataFrame.cxx
vepadulano: Ok I see, then I think the comment describing what the query is doing should be more explicit, e.g. (both for TTree and RDF)
```
// Gather all values from a branch holding a collection per event, `pt`, and fill a histogram so that we can count the total number of values across all events
```

martamaja10: this actually comes directly from a recent forum post: https://root-forum.cern.ch/t/how-to-show-branch-histogram-entries-from-the-command-line/64295

hageboeck: That's a fair user request, but the "official" way to do it with TTree::Draw is to get it from `gDirectory->Get(<name>)`, so maybe we should stick to this and not use the pad.
See "Saving the result of Draw to a histogram" [here](https://root.cern.ch/doc/master/classTTree.html#a73450649dc6e54b5b94516c468523e45)

It gets trickier with TGraph, because those seem to be only in the pad. 😅

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,177 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+   <td>
+~~~{.cpp}
+df.Filter("event == 1").Histo1D("jet_eta", "weight")->Draw();
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Draw a histogram using the method of the class that is being used.
```
vepadulano: Here I would just phrase it as
```suggestion
// Draw a histogram filled with values resulting from calling a method of the class of the `event` branch in the TTree
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,177 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+   <td>
+~~~{cpp}
+// Draw the X() component of the 
+// ROOT::Math::DisplacementVector3D in vec_list
+tree->Draw("vec_list.X()");
+~~~
+   </td>
+   <td>
+~~~{cpp}
+auto histo = df.Define("x", R"(ROOT::RVecD out; for(const auto &el: vec_list) out.push_back(el.X()); return out;)").Histo1D("x");
```
vepadulano: Here the `R` prefix with the parentheses is only useful if the string is placed on multiple lines, otherwise you can use a simple string.

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,177 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+histo->Draw();
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Count number of entries in a specific dataset branch in case this branch has fewer entries than the full dataset
```
vepadulano: I'm sorry, the new phrasing is even more confusing, I don't know how this sentence connects either with the TTree::Draw example or with the RDataFrame example

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,177 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+tree->Draw("pt>>histo");
+TH1D histo = (TH1D)gDirectory->Get("histo");
```
vepadulano:
```suggestion
TH1D *histo = (TH1D *)gDirectory->Get("histo");
```

ferdymercury: Maybe nicer if one uses:
auto histo = gDirectory->Get<TH1D>("histo");

vepadulano: That  would unfortunately be wrong, the closest would be
```cpp
auto histo = gDirectory->Get<TH1D>("histo");
```

ferdymercury: I wrote it correctly, but... the brackets got eaten by HTML xD

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,178 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+</tr>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Draw the histogram and fill hnew with it
+tree->Draw("sqrt(x)>>hnew","y>0"); 
+
+// Retrieve hnew from the current directory
+TH1F *hnew = (TH1F*)gDirectory->Get("hnew");
```
ferdymercury, line 2009:
```suggestion
auto hnew = gDirectory->Get<TH1F>("hnew");
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1916,6 +1916,178 @@ Non-finite numbers can be suppressed using Filter(), e.g.:
...
+~~~
+   </td>
+</tr>
+<tr>
+   <td>
+~~~{cpp}
+// Gather all values from a branch holding a collection per event, `pt`, 
+// and fill a histogram so that we can count the total number of values across all events 
+tree->Draw("pt>>histo");
+TH1D *histo = (TH1D *)gDirectory->Get("histo");
```
ferdymercury, line 2079:
```suggestion
auto histo = gDirectory->Get<TH1D>("histo");
```

pcanal:
This is a great start!  Thanks.  For later PRs, there are several other cases (see for example [dt_DrawTest.C](https://github.com/root-project/root/blob/e4aef2dda6d0888cb7b4b5dccd3d3a32ffa483a0/roottest/root/treeformula/event/dt_DrawTest.C) that we should also tackle (in particular array element selection and indexing).

PR #20285, opened 2025-11-04
[tutorials] Return void instead of TCanvas* in histogram tutorials

siliataider:
# This Pull request:
Update the histogram tutorials to return `void` instead of a `TCanvas*`, which is also the convention used in other ROOT tutorials.

This is required for testing in the Python wheels workflow #19600 since running `root tutorial.C` in this context returns code `255` , causing `pytest` to flag the test as failed.

See failure message:
```bash
FAILED test/wheels/test_tutorials.py::test_cpp_tutorial[hist004_TH1_labels.C] - subprocess.CalledProcessError: Command '['/opt/hostedtoolcache/Python/3.11.14/x64/bin/root', '-b', '-q', '/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/ROOT/tutorials/hist/hist004_TH1_labels.C']' returned non-zero exit status 255.
```

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tutorials/hist/hist006_TH1_bar_charts.C
+++ b/tutorials/hist/hist006_TH1_bar_charts.C
@@ -9,7 +9,7 @@
 /// \date November 2024
 /// \author Rene Brun
 
-TCanvas *hist006_TH1_bar_charts()
+void hist006_TH1_bar_charts()
 {
    // Try to open first the file cernstaff.root in tutorials/io/tree directory
    TString filedir = gROOT->GetTutorialDir();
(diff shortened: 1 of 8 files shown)


tutorials/hist/hist006_TH1_bar_charts.C
```diff
@@ -28,14 +28,12 @@ TCanvas *hist006_TH1_bar_charts()
    auto file = std::unique_ptr<TFile>(TFile::Open(filename, "READ"));
    if (!file) {
       Error("hist006_TH1_bar_charts", "file cernstaff.root not found");
-      return nullptr;
```
hahnjo, line 31: The `return` should stay, here and in other places

linev:
See this list: https://github.com/root-project/root/blob/master/tutorials/CMakeLists.txt#L547-L562

linev:
Now looks better - but let see CI first

tutorials/hist/hist006_TH1_bar_charts.C
hahnjo: You'll probably need to update the line to `return;` because it's now a `void` function - sorry. (and CI is broken right now anyway)

hahnjo (approved):

PR #20288, opened 2025-11-05
Suppress deprecation warnings in rootcling, so `-Werror` builds can complete.

hageboeck:
- Define ROOT's R__DEPRECATED macro to empty in rootcling
- Put back two deprecated functions. They are scheduled for removal in v6.40


hageboeck:
The removal works as foreseen. Now testing the same in v6.38

PR #20289, opened 2025-11-05
Don't issue deprecation warnings in rootcling v6.38

hageboeck:
# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes #


hahnjo (changes requested):
I think this is wrong. The real problem is that `graf3d/gl/inc/TGLWSIncludes.h` has
```
#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 38, 00)
```
That is now the case for `v6-38-00-patches`. The correct condition (inversion from `RConfig.hxx`) would be
```
#if ROOT_VERSION_CODE > ROOT_VERSION(6, 39, 0)
```

hageboeck:
> I think this is wrong.

That's why there is #20290, because I also said that it shouldn't have to do with rootcling, and couldn't reproduce it locally.

PR #20290, opened 2025-11-05
[v6.38] Fix an upcoming error for v6.38

hageboeck:
The TGLWSIncludes.h header is being removed from the public interface
in v6.40. Due to using >= 6.38, it will cause a compiler error already
when the version number is incremented to 6.38, where it was supposed to raise only a warning.

Here, the version number is corrected, and the error message is written for users instead of ROOT developers.
On Windows, the unsupported `#warning` is suppressed.

### DIFF
--- a/graf3d/gl/inc/TGLWSIncludes.h
+++ b/graf3d/gl/inc/TGLWSIncludes.h
@@ -18,18 +18,11 @@
 
 // This header is deprecated according to
 // https://its.cern.ch/jira/browse/ROOT-9807
-// In the 6.38 release, our users will get a warning, and then in 6.40 we
-// remove this header. Conditional on the ROOT version, this header will give
-// an error on inclusion to remind us to move it. Remove also the
-// _ROOT_GL_BUILDS_ITSELF definition from the CMakeLists.txt then.
-#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 38, 00)
-#error "Header should be moved outside of the public ROOT interface now"
-#else
-#ifndef _ROOT_GL_BUILDS_ITSELF
+#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 39, 00)
+#error "The TGLWSIncludes.h header was removed from ROOT in v6.40"
+#elif !defined(_ROOT_GL_BUILDS_ITSELF) and !defined(_WIN32)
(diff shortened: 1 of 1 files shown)


hahnjo:
Maybe it would be better to force-push the branch, actually removing the two commits?

hageboeck:
> Maybe it would be better to force-push the branch, actually removing the two commits?

I would like to do that, but I will confirm with Danilo once he's back if it's OK.
*Edit: Done now*

graf3d/gl/CMakeLists.txt
hahnjo, line 1: The commit message says
> Due to using a >= instead of =, it was causing a compiler error already in v6.38, where it was supposed to raise only a warning.

How would `=` have helped?

graf3d/gl/inc/TGLWSIncludes.h
```diff
@@ -18,17 +18,9 @@
 
 // This header is deprecated according to
 // https://its.cern.ch/jira/browse/ROOT-9807
-// In the 6.38 release, our users will get a warning, and then in 6.40 we
-// remove this header. Conditional on the ROOT version, this header will give
-// an error on inclusion to remind us to move it. Remove also the
-// _ROOT_GL_BUILDS_ITSELF definition from the CMakeLists.txt then.
-#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 38, 00)
```
hahnjo, line 25: Instead of duplicating the header, can we only fix this line?
I proposed
```
#if ROOT_VERSION_CODE > ROOT_VERSION(6, 39, 0)
```
in https://github.com/root-project/root/pull/20289#pullrequestreview-3422007725

graf3d/gl/CMakeLists.txt
hageboeck: It's obviously a typo.

graf3d/gl/inc/TGLWSIncludes.h
hageboeck: I'm not sure.
Given that the header is installed, I would like to make it clearer for users that look into that header, so I would at least remove the irrelevant parts of the comments and the confusing reference to the removal of `_ROOT_GL_BUILDS_ITSELF`, because that has nothing to do with what the users might be wondering about when they see the warning.

PR #20286, opened 2025-11-05
[cmake][gl] Fix deprecation of glew

bellenot:

PR #20292, opened 2025-11-05
Fix GLEW and TH1K deprecation errors

bellenot:
- Move TGLWSIncludes.h in the src directory
- Don't install the glew headers
- Remove deprecated TH1K and its references
This PR supersedes https://github.com/root-project/root/pull/20286


linev:
The remaining point is `_uhi.py`. 
It refer non-existing classes `TH2K` and `TH3K`. 
Also comment for `_values_by_copy` function can be adjusted.

bellenot:
> The remaining point is `_uhi.py`. It refer non-existing classes `TH2K` and `TH3K`. Also comment for `_values_by_copy` function can be adjusted.

As I said, I'll let @siliataider take care of this

bellenot:
So I'll create a new PR for GLEW only and let anyone else take care of TH1K

PR #20294, opened 2025-11-05
Fix GLEW deprecation errors

bellenot:
- Move `TGLWSIncludes.h` from the `inc` to the `src` directory
- Don't install the GLEW headers


smuzaffar:
Hi @bellenot , while testing latest root mater changes , we get build errors in CMSSW [a]. The root change set which we are testing in cmssw is https://github.com/root-project/root/compare/4ee5e77eb33...59b8b01860d . I think this error might be a side effect of change here. Any idea why this header is now needed ?

[a]
```
In file included from src/Fireworks/Core/src/CmsAnnotation.cc:4:
.../lcg/root/6.39.1-f6b681b05d584b7e93413a4cff8a643d/include/TGLIncludes.h:21:10: fatal error: GL/glew.h: No such file or directory
   21 | #include <GL/glew.h>
```

smuzaffar:
`TGLIncludes.h` header in root has not changed in years , so may be something wrong on our side. Let me first check if this still works with previous commit of root

smuzaffar:
ah previously `GL/glew.h` header was part of root distribution but it is not any more. So are we suppose to get glew-devel from system ?

guitargeek:
Yes, I guess you don't build ROOT with `builtin_glew` on?

smuzaffar:
we do build ROOT with `builtin_glew` https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_16_0_X/rootmaster/root.spec#L91
```
  -Dbuiltin_zlib=OFF \
  -Dbuiltin_lzma=OFF \
  -Dbuiltin_gsl=OFF \
  -Dbuiltin_glew=ON \
  -Dbuiltin_ftgl=ON \
  -Dbuiltin_gl2ps=ON \
  -Dbuiltin_xxhash=ON \
```

smuzaffar:
`cmake` also confirms this
-- Enabled support for:  asimage asimage_tiff builtin_clang builtin_cling builtin_civetweb builtin_ftgl builtin_gl2ps **builtin_glew** builtin_llvm builtin_openui5 builtin_xxhash check_connection clad dataframe davix dcache fftw3 gdml geom http imt mathmore opengl pyroot roofit root7 runtime_cxxmodules shared spectrum ssl thisroot_scripts tmva tmva-cpu tmva-cudnn tmva-pymva tpython use_gsl_cblas webgui x11 xml xrootd

bellenot:
@smuzaffar Weird, it looks like `GLEW_INCLUDE_DIR` (or `GLEW_INCLUDE_DIRS`) is missing, somehow. I'll check

guitargeek:
Wait, isn't it just because this PR removed installing of the glew headers in `builtins/glew/CMakeLists.txt`?

So no matter what the include path is, the headers won't be found because ROOT doesn't ship them anymore.

guitargeek:
This PR did two things in one commit that seem unrelated to me:

> * Move `TGLWSIncludes.h` from the `inc` to the `src` directory
> 
>     * Don't install the GLEW headers

The `TGLWSIncludes.h` header was properly deprecated with an entry in the 6.38 release notes:
https://github.com/root-project/root/blob/master/README/ReleaseNotes/v638/index.md?plain=1#L44

Installing the builtin glew headers was not deprecated yet, so this change comes at a surprise. But I agree it's a good change, because we want to replace `glew` with glad anyway! But I think we are missing a few things in the release notes:

  * deprecate the installed glew headers in 6.38 with announcement in the release notes
  * mention in the 6.40 release notes that the builtin glew headers were removed (as done in this PR)

As a consequence, we also need to deprecate the `TGLIncludes.h` header, because this was a wrapper for the builtin `<GL/glew.h>`. I can take care of that stuff, because I started the whole chain with the deprecation of `TGLWSIncludes.h` :slightly_smiling_face: 

On the CMSSW side, glew is not used anyway, so I'll remove some includes there to make it compile. I'll open a PR to CMSSW.

bellenot:
@smuzaffar @guitargeek See https://github.com/root-project/root/pull/20323

guitargeek:
@smuzaffar, I think this PR should fix the situation for CMSSW:
  * https://github.com/cms-sw/cmssw/pull/49341

That gives us more time to think calmly what to do on the ROOT side.

PR #20295, opened 2025-11-05
[graphics] fix different SavePrimitve problems

linev:
1. Default values for fill attributes were not correct in many places. So one has to specify -1, -1 in SaveFillAttributes to correctly store them
2. For line attributes check if gStlye modified and force saving (1,1,1) attributes if this is a case
3. Correctly save frame border mode in canvas
4. For TMultiGraph store histogram like for TGraph - to avoid complex situation when `gPad->Update()` called many time during execution of saved macro
5. Update JSROOT where axis ticks error is fixed now.
6. Update stressGraphics_web.ref file after JSROOT fixes
7. Adjust hist tests where SavePrimitive is testes - fill attributes now forced.

### DIFF
--- a/hist/hist/test/test_MapCppName.cxx
+++ b/hist/hist/test/test_MapCppName.cxx
@@ -47,7 +47,7 @@ TEST(TH1, MapCppNameTest)
    if (!gSystem->GetPathInfo(CFile.Data(), fs))
       FileSize = (Int_t)fs.fSize;
 
-   EXPECT_NEAR(FileSize, 5867, 200);
+   EXPECT_NEAR(FileSize, 6350, 200);
 
    gSystem->Unlink(CFile.Data());
 }
--- a/hist/hist/test/test_TH1_SaveAs.cxx
+++ b/hist/hist/test/test_TH1_SaveAs.cxx
@@ -126,7 +126,7 @@ struct TestSaveAs {
       if (!infile) {
          return false;
       }
-      constexpr Int_t NC = 29; // lines in C file (excl. empty and commented out lines)
+      constexpr Int_t NC = 31; // lines in C file (excl. empty and commented out lines)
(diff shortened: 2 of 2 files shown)


couet (approved):

hist/hist/src/TMultiGraph.cxx
```diff
@@ -1602,21 +1604,22 @@ void TMultiGraph::SavePrimitive(std::ostream &out, Option_t *option)
...
    const char *l = strstr(option, "th2poly");
    if (l)
       out << "   " << l + 7 << "->AddBin(multigraph);\n";
    else
       SavePrimitiveDraw(out, "multigraph", option);
 
-   TAxis *xaxis = GetXaxis();
-   TAxis *yaxis = GetYaxis();
-
-   if (xaxis) {
```
pcanal, line 1614: This removal is not called out in the commit log/PR header.  Is it related/replaced-by storing the histogram?

js/build/jsroot.js
```diff
@@ -1,4 +1,4 @@
-// https://root.cern/js/ v7.10.0
+// https://root.cern/js/ v7.10.99
```
pcanal, line 1: The PR header does not mentioned replaced jsroot (it does mentioned adjusting a reference file though).
Is it an omission in the PR header or is it an unintended commit?

linev: I update JSROOT while it also involved with stressGraphics web tests.

hist/hist/src/TMultiGraph.cxx
linev: It is point 4 - "For TMultiGraph store histogram like for TGraph". 
Before TMultiGraph tried to store axes attributes - making canvas update when storing multigraph in the macro.
And this has different side effects.

PR #20297, opened 2025-11-05
[hist] remove TH1K

silverweed:
Removing TH1K as per the deprecation note.

Still todo: removing TH*K from _uhi.py (@siliataider)


ferdymercury:
Just cross-posting here the reasons for deprecation: https://github.com/root-project/root/issues/19761

siliataider:
removal from UHI:  #20299

PR #20305, opened 2025-11-06
[ci] Set `r=ON` where possible

guitargeek:
So far, the R bindings were only built on Ubuntu 22.04, which is becoming a bit dated already.

Since the R bindings usually don't cause any trouble, it would be good to build them on all platforms where we can easily do that.


ferdymercury:
Side note from:
https://its.cern.ch/jira/browse/ROOT-10756

`the R bindings of ROOT were not so popular, the development of these bindings has halted completely, and the "newest" platform that we still support them on right now is Ubuntu 22.04, which is also end-of-life in two years. Our focus is clearly on the Python bindings. So at this point, I don't think we should bother ensuring the right installation behavior for the R libraries.`

guitargeek:
That's a good point. Indeed I argued against continuing to develop these R bindings, but as they are already here we should at least test what we have I think. It doesn't cost much

dpiparo (approved):

PR #20306, opened 2025-11-06
[hist] Implement `RHistEngine::SetBinContent`

hahnjo:
This needs one indirection to construct the `std::array` of indices with the right (number of) elements.

### DIFF
--- a/hist/histv7/doc/DesignImplementation.md
+++ b/hist/histv7/doc/DesignImplementation.md
@@ -58,11 +58,12 @@ Special arguments are passed last.
 Examples include
 ```cpp
 template <typename... A> void Fill(const std::tuple<A...> &args, RWeight w);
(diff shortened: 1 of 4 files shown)


hahnjo:
After discussion with Jakob, we may decide that we don't want `RHistEngine::SetBinContent` in the public interface. We don't want it in `RHist` because the semantics are unclear together with global histogram statistics. That's not a problem for `RHistEngine`, but it is still questionable if users should call it. On the other hand, `RHistEngine::SetBinContent` is useful for testing, for example the conversion to `TH*`, but that we could also do with `Internal` functions that we be`friend`...

hahnjo:
We decided to not add this method to the public interface because it has unclear semantics for a number of cases, in particular related to global histogram statistics. While it might be fine to allow it in `RHistEngine`, it would also be confusing to then not have it in `RHist`. https://github.com/root-project/root/pull/20392 removes the two mentions we had so far.

Issue #8582, opened 2021-06-30
TThreadTimer behavior


ferdymercury:
### Explain what you would like to see improved
After scratching my head for several hours/days, I found out the reason for a weird behavior I was seing in my ROOT-based C++ application for a DAQ system.

Long story short: the DAQ acquires data in several chunks, and each chunk readout takes 50 ms. After each readout, a `TThread::Printf()` is called. This makes each iteration take much more than 50ms, sometimes 100ms, sometimes 60ms, etc.
(I observe a similar timing behaviour if I call `Emit("...")` to communicate with another thread, instead of `Printf`.)

I created a minimal reproducer:

```
//Run as root -l test_timing.cpp+
#include <chrono>
#include "TThread.h"

void test_timing()
{
	const std::chrono::steady_clock::time_point tbegin = std::chrono::steady_clock::now();
    for(int counter = 0; counter < 10; ++counter)
    {
        const std::chrono::steady_clock::time_point tend = std::chrono::steady_clock::now();
        usleep(50000);//simulate readout of 50ms
        TThread::Printf("Chunk: %u Time/ms: %u",counter, static_cast<unsigned int>(std::chrono::duration_cast<std::chrono::milliseconds>(tend - tbegin).count()));
        //std::cout << "Chunk: " << counter << " Time/ms: " << static_cast<unsigned int>(std::chrono::duration_cast<std::chrono::milliseconds>(tend - tbegin).count()) << std::endl;
    }
}
```

If I run this standalone example, I get as expected times in steps of 50 ms approx:
```
Chunk: 0 Time/ms: 0
Chunk: 1 Time/ms: 50
Chunk: 2 Time/ms: 100
Chunk: 3 Time/ms: 150
Chunk: 4 Time/ms: 200
Chunk: 5 Time/ms: 250
Chunk: 6 Time/ms: 301
Chunk: 7 Time/ms: 352
Chunk: 8 Time/ms: 402
Chunk: 9 Time/ms: 453
```
(A similar result is obtained if I use the `std::cout` statement instead of the `TThread::Printf`.)

Now the problem. If I run this same loop in an application with a GUI window and a `TThread` (full reproducer code is below), then I get quite weird results. The first step takes 107 ms ! And the later steps take 60 ms instead of 50 ms.
```
Chunk: 0 Time/ms: 0
Chunk: 1 Time/ms: 107
Chunk: 2 Time/ms: 167
Chunk: 3 Time/ms: 226
[...]

PR #20307, opened 2025-11-06
[core] improve by two the ttimer resolution

ferdymercury:
Fixes https://github.com/root-project/root/issues/8582

### DIFF
--- a/core/base/inc/Rtypes.h
+++ b/core/base/inc/Rtypes.h
@@ -60,7 +60,7 @@ enum ESysConstants {
    kMAXSIGNALS       = 16,
    kMAXPATHLEN       = 8192,
    kBUFFERSIZE       = 8192,
-   kItimerResolution = 10      // interval-timer resolution in ms
+   kItimerResolution = 5       ///< interval-timer resolution in ms (used for TThread, TTimer, Emit, Connect, etc.)
 };
 
 enum EColor { kWhite   =0,   kBlack    =1,   kGray   =920,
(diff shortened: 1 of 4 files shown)


pcanal:
Do we understand why the result of 'only' cutting in half as such a dramatic effects (in https://github.com/root-project/root/issues/8582 it seems to imply (but the change was seemingly slightly different) that it goes from 'wasting' 10ms per loop to wasting none?

What is the actually effect on the original reported problem on this PR?

ferdymercury:
> Do we understand why the result of 'only' cutting in half as such a dramatic effects (in #8582 it seems to imply (but the change was seemingly slightly different) that it goes from 'wasting' 10ms per loop to wasting none?
> 
> What is the actually effect on the original reported problem on this PR?

Well, I guess alignment plays a role, if you had bad luck you were always sleeping the 10ms until the next check. When removing that extra 10ms, sometimes it was exactly 50, sometimes 53, others 52.
So the idea is to reduce the general latency so that the Thread checks more often if there is stuff to do.
In any case, I switched to std::thread which does not have this problem, so I am also fine with closing the issue as won't fix.

pcanal:
>  When removing that extra 10ms,

This PR, does not remove the 10ms but reduce it to 5ms (and 5+5 in one of the case).  The question I have is whether 5ms is good enough or need to be lower ....

ferdymercury:
> > When removing that extra 10ms,
> 
> This PR, does not remove the 10ms but reduce it to 5ms (and 5+5 in one of the case). The question I have is whether 5ms is good enough or need to be lower ....

Tthreadtimer had a 10+10ms as rate. Back then I changed to 10+0 and it worked fine. Now  i changed to 5+5 and additionally changed def resolution to 5.
Whether it's enough depends on the application I guess...

pcanal (approved):
Thanks.

PR #20308, opened 2025-11-06
[RF][HS3] Remove redundant check in variable export

guitargeek:
The check for already added variables was redundant, because in the `exportObject()` function we already check if a given argument was already exported, tracking this in a standard map.

This drastically reduces performance and memory usage, because the lookup in the JSON itself was very slow and memory hungry, because of some caching that nlohmann-json tries to do when looking up in array collections.

Fixes a problem reported by @kratsg. To be backported to the affected previous releases.


kratsg:
Thanks! This is *not* obvious 😅

cburgard:
looks good to me, for what it's worth. impressive speed at which you found this @guitargeek !

dpiparo (approved):

PR #20311, opened 2025-11-06
CI: test minimal builds

pcanal:
The ctest for minimal build seems broken again (at least for me) ... adding it to the CI will at least make sure we do not easily break it.


.github/workflows/root-ci-config/buildconfig/global-minimal.txt
```diff
@@ -0,0 +1,80 @@
+alien=OFF
```
vepadulano: Just as curiosity, why isn't the config just `minimal=ON`, without specifying any other build option?

pcanal: That is a good question.  In this first approximation, I assumed there was a 'good' reason why those were explicitly off in the usual `global.txt` and keep all the setting that were `OFF` or were auxiliary but I may hare erred.

One other reason is that we wanted the options to be explicit rather than implicit and thus for the build to not depend on a change in the source code for the default to change.  In the case of `minimal` this might actually be a detriment.

dpiparo:
When #20711 that fixes tests for minimal builds is merged perhaps it can be worth rebasing and testing this PR again.

ferdymercury:
Rebased version: https://github.com/root-project/root/pull/20725

Issue #20313, opened 2025-11-06
roottest/root/io/newstl should preserves its data files.


pcanal:
As discussed [https://github.com/root-project/root/pull/19709#discussion_r2470311313](url) the pr https://github.com/root-project/root/pull/19709 remove (in practice) the automatic ability for the `roottest/root/io/newstl:run` test to preserve the `ROOT` files produced by each 'known' version of ROOT.   The test should be upgraded to preserve, at least once per ROOT release, those data files and to use those preserved files (at least in some [CI] builds).

linev:
I create PR #20319 addressing this issue

PR #20319, opened 2025-11-07
[roottest] Improve io/newstl tests, add reference ROOT files

linev:
With Makefile-based testing newstl tests were writing ROOT files into sub-directory with current version id.
So if same roottest was running with different ROOT version - it was creating different sub-dirs with such files.
And roottest was trying to read all these directories afterwards.

This functionality was never in use for CI - while never same roottest was running with different ROOT versions.

Now directly add ROOT files from previous ROOT versions and test reading of such files.
While automatic directory listing no longer required - simplify tests in this respect.
For the beginning add files generated with ROOT 6-32-19.

Fixes: https://github.com/root-project/root/issues/20313

### DIFF
--- a/roottest/root/io/newstl/CMakeLists.txt
+++ b/roottest/root/io/newstl/CMakeLists.txt
@@ -26,7 +26,7 @@ ROOTTEST_ADD_TEST(run
 
 ROOTTEST_ADD_TEST(read-6-32-19
                   MACRO srun.C
-                  MACROARG "\"${CMAKE_CURRENT_SOURCE_DIR}/\", \"${CMAKE_CURRENT_SOURCE_DIR}/6-32-19/\""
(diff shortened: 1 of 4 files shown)


pcanal:
@dpiparo This is an interesting option to preserve the legacy root files (i.e. store them in the repository).  The other option would be to store them (and retrieve them for the purpose of this test) in EOS.   For perspective, a clean complete clone of the ROOT source is 2.2GB, while each set of files is approximately 1.6MB.

dpiparo:
How much would we add in total in terms of size if we added the files to the repo?

linev:
@pcanal 

I do not think that we need to add many different ROOT versions. 
Something like from ROOT5 and several ROOT6. 
So all together about 10MB

dpiparo:
Let's add these 10 MB, it won't be a problem.

PR #20323, opened 2025-11-07
[cmake] Install the builtin GLEW headers

bellenot:
This fixes external use of `TGLIncludes.h`:
```
In file included from src/Fireworks/Core/src/CmsAnnotation.cc:4:
.../lcg/root/6.39.1-f6b681b05d584b7e93413a4cff8a643d/include/TGLIncludes.h:21:10: fatal error: GL/glew.h: No such file or directory
   21 | #include <GL/glew.h>
```


guitargeek (changes requested):
Hi! I think not installing the GLEW headers was actually a good idea, because we want to migrate away from GLEW anyway.

So I would suggest to properly deprecate them in 6.38, so that they can remain deleted in the 6.40 development branch: https://github.com/root-project/root/pull/20339

guitargeek:
Superseded by https://github.com/root-project/root/pull/20340

PR #20333, opened 2025-11-07
[BP][6.36][hist] Fix casts in TH[123]L::SetBinContent

ferdymercury:
BP of hahnjo's https://github.com/root-project/root/pull/20318


guitargeek:
Thank you!

PR #20337, opened 2025-11-07
[RF] Make RooEvaluatorWrapper public in experimental interface

guitargeek:
Enable checkpointing for loops in RooFit codegen and make `RooEvaluatorWrapper` class public for more convenient AD development in dependent frameworks like CMS Combine.


vgvassilev (approved):
Lgtm!

PR #20341, opened 2025-11-07
Remove dictionaries for raw file

jblomer:


pcanal (approved):
LGTM

hahnjo:
It seems this breaks the plugin mechanism for `RRawFileDavix` without `runtime_cxxmodules`, probably because ROOT cannot locate which library to load. Maybe we can get around it with `#pragma link C++ class ROOT::Internal::RRawFileDavix-;`, pretending to have a custom streamer?

jblomer:
> It seems this breaks the plugin mechanism for `RRawFileDavix` without `runtime_cxxmodules`, probably because ROOT cannot locate which library to load. Maybe we can get around it with `#pragma link C++ class ROOT::Internal::RRawFileDavix-;`, pretending to have a custom streamer?

Strange. In the plug-in description, we specify the library. Also, why does it work then for the XRootD raw file that doesn't have a dictionary.

hahnjo:
> Also, why does it work then for the XRootD raw file that doesn't have a dictionary.

Hm, are we testing the plugin support for the `root://` protocol? `net/netxng/test/RRawFileNetXNG.cxx` constructs `RRawFileNetXNG` directly...

hahnjo:
The motivation of the failing `mac-beta` build should be addressed by https://github.com/root-project/root/pull/20389

jblomer:
Since the motivation is gone and this PR causes problems, I'll close it.

PR #20343, opened 2025-11-08
Fix setting precision for matrix values in TGDMLWrite::CreateMatrixN

KoljaFrahm:
# This Pull request:

## Changes or fixes:
https://github.com/root-project/root/issues/20342

## Checklist:

- [x] tested changes locally
- [ ] updated the docs (if necessary)

This PR fixes https://github.com/root-project/root/issues/20342

PR #20345, opened 2025-11-08
[Geom] Fix setting precision for matrix values in TGDMLWrite::CreateM…

### LINKED ISSUE #20342: Precision not set for matrix values when writing gdml
### Check duplicate issues.

- [x] Checked for duplicates

### Description

When exporting a gdml, the matrix values are set with the default C++ ostream precision (6 significant digits) and not with the fFltPrecision. This is often not enough precision. (e.g. it fails to compute the correct group velocity in geant4 for me if it's just 6 significant digits).

### Reproducer

Use TGeoManager::Export with a matrix saved in the TGeoManager

### ROOT version

all

### Installation method

all

### Operating system

all

### Additional context

_No response_

ahmedbilal9:
Fixes #20342

## Changes or fixes:

When exporting GDML files, matrix values were being written with the default C++ ostream precision (6 significant digits) instead of using the configured fFltPrecision setting. This caused loss of precision in geometry data.

Applied the same precision formatting pattern used in CreateConstantN to ensure consistent precision across all GDML exports.

## Checklist:

- [x] Tested changes locally
- [ ] Updated the docs (if necessary)

### DIFF
--- a/geom/gdml/src/TGDMLWrite.cxx
+++ b/geom/gdml/src/TGDMLWrite.cxx
@@ -164,7 +164,6 @@ See that function for details.
 #include <ctime>
 #include <sstream>
 
-
 TGDMLWrite *TGDMLWrite::fgGDMLWrite = nullptr;
 
 namespace {
@@ -2040,9 +2039,10 @@ XMLNodePointer_t TGDMLWrite::CreateMatrixN(TGDMLMatrix const *matrix)
    XMLNodePointer_t mainN = fGdmlE->NewChild(nullptr, nullptr, "matrix", nullptr);
    fGdmlE->NewAttr(mainN, nullptr, "name", matrix->GetName());
    fGdmlE->NewAttr(mainN, nullptr, "coldim", TString::Format("%zu", cols));
+   const TString fltPrecision = TString::Format("%%.%dg", fFltPrecision);
    for (size_t i = 0; i < rows; ++i) {
       for (size_t j = 0; j < cols; ++j) {
-         vals << matrix->Get(i, j);
+         vals << TString::Format(fltPrecision.Data(), matrix->Get(i, j));
          if (j < cols - 1)
             vals << ' ';
       }


dpiparo:
Thanks! I started the tests waiting for @agheata to review.
Would it be feasible to reduce the setup that led you to identify this bug to a test?

ahmedbilal9:
> Thanks! I started the tests waiting for @agheata to review. Would it be feasible to reduce the setup that led you to identify this bug to a test?

Hi! I've added a test case for the precision fix in commit e3ea518ae1.

The test verifies that matrix values in GDML exports use the configured fFltPrecision setting instead of the default C++ ostream precision (6 digits).

Test details:
- Location: `roottest/root/io/geo/testMatrixPrecision.cxx`
- Creates a geometry with rotation matrices containing high-precision values
- Exports to GDML and verifies the output maintains sufficient decimal precision
- Only runs when GDML support is enabled

The test follows the existing pattern used in the geo test suite and integrates with the CMake build system.

KoljaFrahm:
This works as well in case you like this more than https://github.com/root-project/root/pull/20343

ahmedbilal9:
Hi! I've added a test case for the precision fix in commit e3ea518.

The test verifies that matrix values in GDML exports use the configured fFltPrecision setting instead of the default C++ ostream precision (6 digits).

Test details:
- Location: `roottest/root/io/geo/testMatrixPrecision.cxx`
- Creates a geometry with rotation matrices containing high-precision values
- Exports to GDML and verifies the output maintains sufficient decimal precision
- Only runs when GDML support is enabled

The test follows the existing pattern used in the geo test suite and integrates with the CMake build system.

I've also applied clang-format to fix the code style issues.

agheata (changes requested):
Thanks @ahmedbilal9 for catching this! The fix is OK, but not the unit test. The method actually does not apply (in TGeo) to rotation matrices but rather to the more general "matrix" GDML tag. ROOT only reads that tag with `TGDMLParse` if the GDML is produced externally and contains it; then, it stores the data temporarily in `TGeoManager` in case it needs to be exported as GDML. So your test does not find the "matrix" tag in the exported file, since ROOT generates "rotation" instead. Besides that, the feature is small and not likely to change after fixing,  so I would prefer not adding a unit test anyway. So, could you either rebase your branch and keep a single commit with the fix, or if you prefer make a new PR with that commit, closing the 2 PR's?

ahmedbilal9:
> Thanks @ahmedbilal9 for catching this! The fix is OK, but not the unit test. The method actually does not apply (in TGeo) to rotation matrices but rather to the more general "matrix" GDML tag. ROOT only reads that tag with `TGDMLParse` if the GDML is produced externally and contains it; then, it stores the data temporarily in `TGeoManager` in case it needs to be exported as GDML. So your test does not find the "matrix" tag in the exported file, since ROOT generates "rotation" instead. Besides that, the feature is small and not likely to change after fixing, so I would prefer not adding a unit test anyway. So, could you either rebase your branch and keep a single commit with the fix, or if you prefer make a new PR with that commit, closing the 2 PR's?

I've rebased the PR to a single commit and removed the test as requested. 
The PR now contains only the precision fix for matrix values in GDML exports.

ahmedbilal9:
> Thanks @ahmedbilal9 for the fix!

Hi @agheata, 

Thank you for approving the fix! I see some CI tests are failing, but they appear to be unrelated infrastructure issues (mac-beta ARM64, ubuntu2510). The core test suite shows 3,737 passing tests.

Should I do anything, or will you merge it as-is?

Thanks!

PR #15908, opened 2024-06-21
[RF] compiler-dependent RooNaNPacker tests

egpbos:
Like the tests added in https://github.com/root-project/root/pull/15499/commits/e00aded416c4d8b0c364035353dfb529602fd0e1, these are also sanity checks meant in part as code-as-documentation for users to see how things work.

It turns out in this case that GCC on the one hand and Clang and MSVC on the other hand have two different solutions for binary arithmetic operations on two NaNs. In both cases, one of the two NaNs is returned, but in one case it's the first, in the other is it's the second.

The tests added in this commit will hopefully warn RooNaNPacker users of this behavior.

PR #20346, opened 2025-11-08
[RF] additional RooNaNPacker tests

guitargeek:
Reboot of #15908.


guitargeek:
PR https://github.com/root-project/root/pull/15908 was not as trivial to resurrect as I thought. The compiler-dependent behavior doesn't seem to be stable over time...

PR #20349, opened 2025-11-08
[tutorials] Read file from EOS in RNTuple dimuon tutorial

vepadulano:
I've converted the CMS open dataset and stored it on our EOS public location.


vepadulano:
> What is the advantage in reading this through XRootD? Having an HTTP URL makes it easier for the user to download the file manually.

I see two main advantages:
* Using data read via XRootD is more representative of a realistic scenario
* Reading data via Davix in our own tests has had issues in the past

 @jblomer I acknowledge that none of the above is a deep, technical motivation to change the dataset read in this tutorial, so I can also just close this PR.

vepadulano:
Closing for lack of a strong technical reason

PR #20351, opened 2025-11-08
[RF][HS3] Remove unused constructor

guitargeek:
Follow up on the previous commit 4c545d7e116, removing a constructor of an internal class that is now unused.

Emergency PR because this fixes the CI failures on the Debians with `dev=ON`.

What is surprising is that the errors didn't appear in the PR that made the constructor unused:
https://github.com/root-project/root/pull/20276/checks

Maybe it's because of the incremental builds, but I'm not sure.

### DIFF
--- a/roofit/hs3/src/RooJSONFactoryWSTool.cxx
+++ b/roofit/hs3/src/RooJSONFactoryWSTool.cxx
@@ -160,12 +160,6 @@ struct Var {
     * @param n Number of bins.
     */
    Var(int n) : nbins(n), min(0), max(n) {}
-
-   /**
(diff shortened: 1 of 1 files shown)


pcanal:
> What is surprising is that the errors didn't appear in the PR that made the constructor unused:
[#20276 (checks)](https://github.com/root-project/root/pull/20276/checks)
> Maybe it's because of the incremental builds, but I'm not sure.

This vaguely deserve some investigation as it could uncover missing dependency or a blind spot in the CI.

pcanal (approved):
LGTM (beside the mystery of how it pass the CI before :( )

Issue #20320, opened 2025-11-07
[DF] Crash when applying Filter on Snapshot written with Vary'ed quantities


rbarrue:
Hello all ! I've been recently testing the functionality of writing Snapshot with Vary'ed quantities discussed in https://github.com/root-project/root/issues/11218 and related to the PR in https://github.com/root-project/root/pull/19725. 

When I try to apply a Filter on a quantity that's Vary'ed and save that quantity to a Snapshot **including its variations** the code crashes with a segmentation fault. I'm attaching the stack trace for the segmentation fault. I am running this in a EL9 container, with the dev3 LCG build of today (`/lcg/views/dev3/latest/x86_64-el9-gcc15-opt/`).

[stack_trace.txt](https://github.com/user-attachments/files/23414721/stack_trace.txt)

The crash doesn't happen if I:
- store the quantity and variations before applying the Filter
- store the quantity without variations after applying the Filter

I attach a MWE with the three scenarios, including three ways to apply `Vary()` so you can test it yourselves.

[mwe_snapshot_vary_filter_crash.py](https://github.com/user-attachments/files/23414753/mwe_snapshot_vary_filter_crash.py)

Tagging @hageboeck, @vepadulano and @martamaja10 as they were the ones tagged in the original issue.

vepadulano:
Thank you @rbarrue for testing this new feature! I was able to create the following C++ simplified reproducer, which already crashes:

```cpp
#include <ROOT/RDataFrame.hxx>
#include <ROOT/RVec.hxx>
#include <TRandom.h>
#include <ROOT/RSnapshotOptions.hxx>

int main(){

    ROOT::RDataFrame df{1};

    auto df_def = df.Define("val", [](){return gRandom->Gaus(10, 2); });

    auto df_var = df_def.Vary("val", [](double val) { return ROOT::RVecD(val * 0.9, val * 1.1); }, {"val"}, {"down","up"}, "var");

    // Jitted function breaks the following Snapshot
    auto df_fil = df_var.Filter("val > 11");
    // Compiled function works
    // auto df_fil = df_var.Filter([](double val) { return val > 11; }, {"val"});

    ROOT::RDF::RSnapshotOptions opts;
    opts.fIncludeVariations = true;
    df_fil.Snapshot("tree", "output_Filter_withVariations.root", ".*" , opts);

}
```

Looks like there is a faulty interaction between the Snapshot with variations and jitted filters, to be investigated.

vepadulano:
[...]

PR #20352, opened 2025-11-08
[df] Consider jitted nodes in Snapshot with variations

vepadulano:
This commit introduces two enhancements to the Snapshot with variations implementation:

- Snapshot with variations must call RLoopManager::Jit before trying to request varied filters from upstream nodes. This is necessary because the upstream node could be a jitted node itself.
- Snapshot with variations must deal with the possibility of having a different nominal upstream node type than the other varied upstream node types.

This aligns the implementation witht the one of RVariedAction for what concerns dealing with jitted nodes.

Fixes https://github.com/root-project/root/issues/20320

### DIFF
--- a/tree/dataframe/test/dataframe_snapshotWithVariations.cxx
+++ b/tree/dataframe/test/dataframe_snapshotWithVariations.cxx
@@ -494,3 +494,38 @@ TEST(RDFVarySnapshot, SnapshotVirtualClass)
       }
    }
 }
+
+// https://github.com/root-project/root/issues/20320
+TEST(RDFVarySnapshot, GH20320)
+{
+   const char *fileName{"dataframe_snapshot_with_variations_regression_gh20330.root"};
+   RemoveFileRAII(fileName);
+
+   ROOT::RDataFrame df{1};
+
+   auto df_def = df.Define("val", []() { return 2; });
+
+   auto df_var =
+      df_def.Vary("val", [](int val) { return ROOT::RVecI{val - 1, val + 1}; }, {"val"}, {"down", "up"}, "var");
+
+   // Jitted filters used to break the Snapshot because:
+   // - It did not JIT the RJittedFilter before requesting for the varied filters
+   // - It did not take into account that the previous nodes of the Snapshot could be of different types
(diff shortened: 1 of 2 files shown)


tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -494,3 +494,43 @@ TEST(RDFVarySnapshot, SnapshotVirtualClass)
...
+      df_def.Vary("val", [](int val) { return ROOT::RVecI{val - 1, val + 1}; }, {"val"}, {"down", "up"}, "var");
+
+   // Jitted filters used to break the Snapshot because:
+   // - It did not JIT the RJittedFilter before requesting for the varied filters
+   // - It did not take into account that the previous nodes of the Snapshot could be of different types
+   auto df_fil = df_var.Filter("val > 0");
+
+   ROOT::RDF::RSnapshotOptions opts;
+   opts.fIncludeVariations = true;
+   auto snap = df_fil.Snapshot("tree", outputGuard.fPath, {"val"}, opts);
```
hageboeck: For the other tests, we had added the `RemoveFileRAII` on your request, so maybe we could unify the file removal for this file. 🙂
```suggestion
   const auto filename = "dataframe_snapshot_with_variations_regression_gh20330.root";
   RemoveFileRAII(filename);

   ROOT::RDataFrame df{1};

   auto df_def = df.Define("val", []() { return 2; });

   auto df_var =
      df_def.Vary("val", [](int val) { return ROOT::RVecI{val - 1, val + 1}; }, {"val"}, {"down", "up"}, "var");

   // Jitted filters used to break the Snapshot because:
   // - It did not JIT the RJittedFilter before requesting for the varied filters
   // - It did not take into account that the previous nodes of the Snapshot could be of different types
   auto df_fil = df_var.Filter("val > 0");

   ROOT::RDF::RSnapshotOptions opts;
   opts.fIncludeVariations = true;
   auto snap = df_fil.Snapshot("tree", filename, {"val"}, opts);
```

vepadulano: Yes, absolutely! Good catch

Issue #20291, opened 2025-11-05
Wrong interatction of FromNumpy and Snapshot


martinfoell:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Taking a Snapshot of a dataframe with FromNumpy only works if a new column is created using Define where all the Numpy arrays passed to FromNumpy are used as inputs to the new column. Performing a Snapshot of the dataframe directly, without creating a new column, or creating a new column that does not use all the dataframe columns as inputs results in a segmentation fault. The python script in the reproducer illustrates the three cases described above.

### Reproducer

```python
import ROOT
import numpy

x, y = numpy.array([1, 2, 3]), numpy.array([4, 5, 6])
df = ROOT.RDF.FromNumpy({"x": x, "y": y})
 
# Uncomment one of the lines below to test the three different cases from the description

df.Define("z", "x + y").Snapshot("tree", "file.root") # this works
# df.Define("z", "2 * x").Snapshot("tree", "file.root") # gives a segmentation violation
# df.Snapshot("tree", "file.root") # gives a segmentation violation
```

### ROOT version

Any

### Installation method

Any

### Operating system

Any

PR #20353, opened 2025-11-08
Fix interaction between RVecDS and Snapshot

vepadulano:
This commit updates the implementation of the column reader retrieval in RVecDS to use GetColumnReaders with type_info instead of the older version. This in turn allows Snapshot to request the correct column reader when reading a Numpy-based dataset.

Fixes https://github.com/root-project/root/issues/20291

### DIFF
--- a/bindings/pyroot/pythonizations/test/rdataframe_misc.py
+++ b/bindings/pyroot/pythonizations/test/rdataframe_misc.py
@@ -3,19 +3,14 @@ import os
 import platform
 import unittest
 
-import ROOT
-
 import numpy
+import ROOT
 
 
 class DatasetContext:
     """A helper class to create the dataset for the tutorial below."""
 
-    filenames = [
-        "rdataframe_misc_1.root",
-        "rdataframe_misc_2.root",
-        "rdataframe_misc_3.root"
-    ]
+    filenames = ["rdataframe_misc_1.root", "rdataframe_misc_2.root", "rdataframe_misc_3.root"]
     treename = "dataset"
     nentries = 5
 
@@ -48,6 +43,7 @@ class DatasetContext:
         for filename in self.filenames:
             os.remove(filename)
 
+
 class RDataFrameMisc(unittest.TestCase):
     """Miscellaneous RDataFrame tests"""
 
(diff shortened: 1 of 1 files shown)


bindings/pyroot/pythonizations/test/.rootrc
hageboeck, line 1: Maybe this and the change in the CMakeLists.txt could be split into a second commit?

tree/dataframe/inc/ROOT/RVecDS.hxx
```diff
@@ -46,46 +54,18 @@ class RVecDS final : public ROOT::RDF::RDataSource {
...
-      const auto namesIt = std::find(colBegin, colEnd, colName);
-      const auto index = std::distance(colBegin, namesIt);
-
-      Record_t ret(fNSlots);
-      for (auto slot : ROOT::TSeqU(fNSlots)) {
-         ret[slot] = fPointerHolders[index][slot]->GetPointerAddr();
-      }
-      return ret;
-   }
+   Record_t GetColumnReadersImpl(std::string_view colName, const std::type_info &id) { return {}; }
```
hageboeck: To fix the warning:
```suggestion
   Record_t GetColumnReadersImpl(std::string_view, const std::type_info &) { return {}; }
```

hageboeck: But is this function needed at all?

hageboeck (approved):
Kind of LGTM, but two things could be touched up.

tree/dataframe/inc/ROOT/RVecDS.hxx
vepadulano: Unfortunately yes, I hope we'll be able to remove it at some point

bindings/pyroot/pythonizations/test/.rootrc
vepadulano: The Snapshot info message can be removed altogether, I uploaded this other PR https://github.com/root-project/root/pull/20375 . Once that is merged, we won't need these `.rootrc` files in the test directories anymore.

PR #20354, opened 2025-11-08
Help ROOT I/O of class in streamer field test

vepadulano:
Attempt at fixing failures such as seen on MacOS beta:

```
  Error in <CloseStreamerInfoROOTFile>: Unique pointer unique_ptr<PolyBase,default_delete<PolyBase> > has zero data members.
```

See e.g. https://github.com/root-project/root/actions/runs/19197912794/job/54881579525?pr=20353#step:6:5580

### DIFF
--- a/tree/ntuple/test/StreamerField.hxx
+++ b/tree/ntuple/test/StreamerField.hxx
@@ -67,20 +67,26 @@ class IgnoreUnsplitComment {
 // Test streamer field with polymorphic type
 
 struct PolyBase {
-   virtual ~PolyBase() {}
+   virtual ~PolyBase() = default;
    int x;
+   ClassDef(PolyBase, 3);
 };
 
 struct PolyA : public PolyBase {
+   ~PolyA() override = default;
    int a;
+   ClassDefOverride(PolyA, 3);
 };
 
 struct PolyB : public PolyBase {
+   ~PolyB() override = default;
    int b;
+   ClassDefOverride(PolyB, 3);
 };
 
 struct PolyContainer {
    std::unique_ptr<PolyBase> fPoly;
+   ClassDefNV(PolyContainer, 3);
 };
 
 template <typename T>


jblomer (approved):

tree/ntuple/test/StreamerField.hxx
```diff
@@ -67,20 +67,26 @@ class IgnoreUnsplitComment {
 // Test streamer field with polymorphic type
 
 struct PolyBase {
-   virtual ~PolyBase() {}
+   virtual ~PolyBase() = default;
    int x;
+   ClassDef(PolyBase, 3);
```
pcanal, line 72: I find surprising that this help and is likely indicative of some other (now obscured) problem (i.e. if this change actually makes a difference). ... 

One possibility is the anchoring (by the compiler) of the virtual function table (usually associated with the first virtual function).

However the issue seems to be reported from `CloseStreamerInfoROOTFile` which should not be related to the virtual function table ...

vepadulano: Opened https://github.com/root-project/root/pull/20373 https://github.com/root-project/root/pull/20374 to see which one of the two changes is necessary, if not both.

vepadulano: @pcanal apparently, each and every one line in this PR is **necessary** to fix the problem on MacOS, see https://github.com/root-project/root/actions/runs/19245732921/job/55019171178?pr=20374#step:6:5554 and https://github.com/root-project/root/actions/runs/19245699010/job/55019061912?pr=20373#step:6:5557

pcanal: This is probably a 'disaster' because it would mean that we would not be able to handle `unique_ptr` of class/struct that do not have a `ClassDef`.  We need to understand better the nature of the changes that lead to this message (We do not know yet if it is a spurrious error message or a serious issue).

Since this is introducing a know issue in the build, we should merge **temporarily** this PR while we work out a solution (i.e. open an issue and reverting PR).

pcanal (approved):
Let put this band-aid on for now, so the other PR are not distracted, but we need to resolved/understand this.

hahnjo:
Unfortunately this didn't work, the build in `master` still fails with the same error: https://github.com/root-project/root/actions/runs/19250798263/job/55035175377 The reason is that this PR landed on `macphsft34` which has
```
sftnight@macphsft34 ~ % sw_vers 
ProductName:		macOS
ProductVersion:		26.1
BuildVersion:		25B77
sftnight@macphsft34 ~ % clang --version
Apple clang version 17.0.0 (clang-1700.4.4.1)
Target: arm64-apple-darwin25.1.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

compared to
```
sftnight@macphsft35 ~ % sw_vers 
ProductName:		macOS
ProductVersion:		26.2
BuildVersion:		25C5031i
sftnight@macphsft35 ~ % clang --version
Apple clang version 17.0.0 (clang-1700.6.3.2)
Target: arm64-apple-darwin25.2.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

(btw the "correct" tag for commit messages is `[ntuple]`)

vepadulano:
I was mislead as well because today's nightly actually run well https://github.com/root-project/root/actions/runs/19251983040/job/55038655542 

This can be reverted then

PR #20355, opened 2025-11-09
[tmva][sofie] Disentangle PyMVA and SOFIE

guitargeek:
So far, the part of SOFIE that uses the C Python API was included in PyMVA for convenience, but the functionality is completely unrelated.

Moving this code to SOFIE itself means we can fully build and test SOFIE without enabling `tmva-pymva`.

The only subtelty is that we want to be able to disable these SOFIE features that require linking against `libpython`, because this is not always allowed (e.g. in the Python wheels). Therefore, this part is only built if we build also the other ROOT libraries that links against `libpython` besides PyMVA, wich is `TPython`.

### DIFF
--- a/tmva/sofie/test/CMakeLists.txt
+++ b/tmva/sofie/test/CMakeLists.txt
@@ -153,19 +153,29 @@ endif()
 
 # Any reatures that link against libpython are disabled if built with tpython=OFF
 if (tpython AND ROOT_KERAS_FOUND AND BLAS_FOUND)
-  configure_file(generateKerasModels.py generateKerasModels.py COPYONLY)
-  configure_file(scale_by_2_op.hxx scale_by_2_op.hxx COPYONLY)
 
-  ROOT_ADD_GTEST(TestRModelParserKeras TestRModelParserKeras.C
-     LIBRARIES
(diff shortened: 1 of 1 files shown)


lmoneta (commented):
Thank you Jonas for moving the Keras and PyTorch aprser away from PyMVA. A comment I have it is maybe to be more consistent, is to move them not in sofie but in the sofie_parsers. 
Something to discuss

guitargeek:
> Thank you Jonas for moving the Keras and PyTorch aprser away from PyMVA. A comment I have it is maybe to be more consistent, is to move them not in sofie but in the sofie_parsers. Something to discuss

Sure, this can be done! I agree it's better to move it into `sofie_parsers`.

guitargeek:
Hi @lmoneta, I have implemented your suggestion to moving the parser implementation to `tmva/sofie_parses` as you suggested. That means it's now conditional on the `tmva-sofie=ON` build flag, just like the ONNX parser.

I left the tests in `tmva/sofie/test` though, because also the tests for the ONNX parser is there for now (there is no `tmva/sofie_parsers/test` directory)

PR #20358, opened 2025-11-09
[Geom] Add auxiliary user-data export in TGDMLWrite

### LINKED ISSUE #20356: Auxiliaries of volumes are not exportet in gdml
### Check duplicate issues.

- [x] Checked for duplicates

### Description

I can import auxiliary values of a gdml file. They are saved as UserExtension in TGDMLParse but there is no way to export any auxiliary values with TGDMLWrite.

### Reproducer

- Import a gdml file with an auxiliary value, e.g. one containing
```
<volume name="test_volume">
      <materialref ref="test_mat"/>
      <solidref ref="test_solid"/>
      <auxiliary auxtype="test_type" auxvalue="test_value"/>
</volume>
```
with `gGeoManager->Import("testfile.gdml")`
- Export it with `gGeoManager->Export("testfile_export.gdml")` 

or add directly to vol and export (e.g. with pyroot)
```
ue = vol.GrabUserExtension()
if not ue:
    aux_map = r.TMap()
    vol.SetUserExtension(r.TGeoRCExtension(aux_map))
else:
    aux_map = ue.GetUserObject()
key = r.TObjString("test_type")
value = r.TObjString("test_value")
aux_map.Add(key, value)
```
and export with 
`r.gGeoManager.Export("test_file_export.gdml")`

In both cases no auxiliary is being exportet.

### ROOT version

all

### Installation method

all

### Operating system

all

### Additional context

_No response_

KoljaFrahm:
# This Pull request:
Adds the option of exporting auxiliary of TGeoVolumes in TGDMLWrite 
Uses same format as in https://github.com/root-project/root/blob/master/geom/gdml/src/TGDMLParse.cxx

## Changes or fixes:
Fixes #20356

## Checklist:

- [x] tested changes locally

This PR fixes #20356

### DIFF
--- a/geom/gdml/src/TGDMLWrite.cxx
+++ b/geom/gdml/src/TGDMLWrite.cxx
@@ -156,7 +156,10 @@ See that function for details.
 #include "TGeoElement.h"
 #include "TGeoShape.h"
 #include "TGeoCompositeShape.h"
+#include "TGeoExtension.h"
 #include "TGeoOpticalSurface.h"
+#include "TMap.h"
+#include "TObjString.h"
 #include <cstdlib>
 #include <string>
 #include <map>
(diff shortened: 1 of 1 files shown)


KoljaFrahm:
I reformatted my commit (clang), do I still have to do so something to start the tests again?

dpiparo:
Is there a way in which we could test this fix?

KoljaFrahm:
I described two ways in the issues (Reproducer).
Is this enough or should I describe something in more detail?

KoljaFrahm:
I am not sure why the ruff code analysis fails for files I haven't changed

geom/gdml/src/TGDMLWrite.cxx
```diff
@@ -759,6 +762,41 @@ void TGDMLWrite::ExtractVolumes(TGeoNode *node)
       fGdmlE->AddChild(volumeN, childN);
    }
 
+   // export auxiliary user-data if present (TMap of TObjString->TObjString)
+   {
+      TGeoRCExtension *rcext = (TGeoRCExtension *)volume->GrabUserExtension();
```
agheata: GrabUserExtension() is intended for a reference-counted use of the pointer. It must be followed by `rcext->Release();` when the pointer is no longer needed; otherwise, a leak occurs. For read-only access, it is better to use GetUserExtension() instead of the Grab version.

agheata:
> Is there a way in which we could test this fix?

Actually, this is not a fix but a feature, allowing the re-export of GDML auxiliary volume data imported from a GDML file. It could be tested eventually if we make a more general GDML import/export regression test

geom/gdml/src/TGDMLWrite.cxx
KoljaFrahm: Thanks for the review, changed this!

KoljaFrahm:
Adjusted and rebased :)

KoljaFrahm:
Is still something missing from my side? Would be nice to have this in the next root version :)

dpiparo:
I propose to have these changes integrated in the main branch as soon as the tests passes.
Then, if @agheata agrees, we could potentially think of having it integrated in the v6-38-00-patches branch to have it available in 6.38.02, likely available in a few weeks from now...

KoljaFrahm:
That would be great!

PR #20361, opened 2025-11-09
Fix rootssh connection issues on MacOS [skip-ci]

70rc:
# This Pull request:
Fixes root connection issues on MacOS.

## Changes or fixes:
Changed rootssh not to create the $listener_socket prior to calling nc.
nc BSD, which is used on MacOS, cannot connect to existing Unix sockets.

## Checklist:

- [x] tested changes locally

# This PR
Fixes rootssh connection issues on MacOS.

### DIFF
--- a/config/rootssh
+++ b/config/rootssh
@@ -55,13 +55,17 @@ elif [[ "$1" == "--as-listener--" ]] ; then
    #remember processid to be able kill it
    nc_procid=$!
 
-   # protect socket and log file from reading
-   chmod -f 0700 $listener_socket $listener_socket.log
+   # protect log file from reading
+   chmod -f 0700 $listener_socket.log
 
-   # remove netcat listening on socket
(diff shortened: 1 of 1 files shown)


linev (changes requested):
Probably it make sense.
I change this line in May 2025, but didn't test on mac.
But  change is not sufficient - in line 59 `chmod -f 0700 $listener_socket` fails for me too. So one need to adjust code here as well.

At the best one should check in the loop:
```
if [-f $listener_socket.log]; then
   chmod -f 0700 $listener_socket
fi
```
@70rc, will you add such change?

linev (approved):
I add extra commit to check socket before changing mode for it.

Thanks to reporting the problem!

70rc:
@linev Apologies for the slow reaction, I'm in the middle of a beamtime. Thanks for adding the missing check yourself!

Issue #20047, opened 2025-10-07
Code coverage shows incorrect missed lines


hahnjo:
After the recent merge of `histv7` PRs, Codecov started showing lines as missed that are definitely covered, for example in `RAxes::ComputeGlobalIndex`:

<img width="1123" height="590" alt="Image" src="https://github.com/user-attachments/assets/a1110b25-7363-4ab0-933b-70e4a32432f8" />

The entire function and all its code branches (except for the one annotated with `GCOVR_EXCL_LINE`) are exercised in [`hist/histv7/test/hist_axes.cxx`](https://github.com/root-project/root/blob/master/hist/histv7/test/hist_axes.cxx).

hahnjo:
I can reproduce this locally in a container. The coverage report looks suspicious (note the high number of lines for `RAxes.hxx` and `RHistEngine.hxx`, as if deduplication didn't work or something):
```
# gcovr --root ../root.src/hist/histv7/inc/ hist/histv7
(INFO) Reading coverage data...
(INFO) Writing coverage report...
------------------------------------------------------------------------------
                           GCC Code Coverage Report
Directory: ../root.src/hist/histv7/inc
------------------------------------------------------------------------------
File                                       Lines    Exec  Cover   Missing
------------------------------------------------------------------------------
[...]

PR #20362, opened 2025-11-10
[ci] Use alma10 for coverage build

hahnjo:
It has a newer compiler which should not show the incorrect missed lines documented in https://github.com/root-project/root/issues/20047.


hahnjo:
Many smaller changes, also fully resolving https://github.com/root-project/root/issues/20047#issuecomment-3501128989:
image

PR #20366, opened 2025-11-10
[hist] Conversion functions from `RHist` to `TH1*`

hahnjo:
Introduce a new package `ROOTHistUtil` that depends both on the "old" `TH1*` `Hist` library and the "new" `histv7` package. Conversion is offered via free functions in the newly added `Hist` namespace. In this initial implementation, only one-dimensional regular histograms are supported.

### DIFF
--- a/hist/histv7util/test/hist_convert_TH1.cxx
+++ b/hist/histv7util/test/hist_convert_TH1.cxx
@@ -3,6 +3,7 @@
 #include <TAxis.h>
 #include <TH1.h>
 
+#include <array>
 #include <cstddef>
 #include <memory>
 #include <stdexcept>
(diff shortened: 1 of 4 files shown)


hist/histv7util/inc/ROOT/Hist/ConvertToTH1.hxx
```diff
@@ -0,0 +1,117 @@
...
+/// Convert a one-dimensional histogram to TH1S.
+///
+/// \copydetails ConvertToTH1I(const RHist<int> &hist)
+std::unique_ptr<TH1S> ConvertToTH1S(const RHist<short> &hist);
+
+/// Convert a one-dimensional histogram to TH1I.
+///
+/// Throws an exception if the histogram has more than one dimension.
+///
+/// \param[in] engine the RHistEngine to convert
```
hageboeck:
```suggestion
/// \param[in] hist the RHist to convert
```

hist/histv7util/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+ROOT_STANDARD_LIBRARY_PACKAGE(ROOTHistUtil
+  HEADERS
+    ROOT/Hist/ConvertToTH1.hxx
```
hageboeck, line 3: I'm stuck at the name of the directory. Would `ROOT/HistUtil` make sense to be closer to the name of the package?

hist/histv7util/test/hist_convert_TH1.cxx
```diff
@@ -0,0 +1,99 @@
...
+
+TEST(ConvertToTH1I, RHistEngine)
+{
+   static constexpr std::size_t Bins = 20;
+   RHistEngine<int> engine(Bins, {0, Bins});
+
+   engine.Fill(-100);
+   for (std::size_t i = 0; i < Bins; i++) {
+      engine.Fill(i);
+   }
```
hageboeck, line 19: Maybe one bin could be filled twice to detect any accidental shifts.

hist/histv7util/CMakeLists.txt
hahnjo: The idea was that the directory layout matches the new `Hist` namespace (that I missed to point out in the commit message / PR summary; fixed now (PR summary) or with the next push for the commit message), but I see the point of matching `HistUtil`. I think ultimately it comes down to whether we also want to use the `Hist` namespace for free functions in the "main" `histv7`, then I think I would prefer to have all headers in one place...

hageboeck: Agreed on having only one `Hist`-related namespace, so if the main package will have free functions in this namespace, `Hist` is the better name.

jblomer (approved):
Nice!

Issue #14554, opened 2024-02-04
[ROOT-2927] Particular instance of data causes segfault in TDecompSparse


vepadulano:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description
From https://its.cern.ch/jira/browse/ROOT-2927

Dear ROOT,

I am attaching to this message a ROOT file that contains a sparse matrix, and a vector.

Trying to solve the equation formed by the two causes a segfault. The steps to reproduce the problem are the following:

```cpp
$ root \-l matrix_fail.root
root [0]
Attaching file matrix_fail.root as _file0...
root [1] M
(class TMatrixTSparse&amp;lt;double&amp;gt;*)0xbed850
root [2] x
(class TVectorT&amp;lt;double&amp;gt;*)0xbedcb0
root [3] TDecompSparse solver(*M,0)
root [4] Bool_t ok
root [5] TVectorD xx = solver.Solve(*x, ok)

*** Break *** segmentation violation
```

The problem is due to this particular instance of the data: if you change one of the values of the x vector for example, you won't observe the segfault.

I have tried this in two different linux distributions (SL5 and ubuntu karmic) and two different ROOT versions (5.24 and 5.18), both 64 bits.

### Reproducer

As per the description

### ROOT version

Any

### Installation method

Any

### Operating system

Any

### Additional context

Find input file in the Jira issue

ferdymercury:
The crash happens on line `2598` of `TDecompSparse.cxx`:

`w1 = w[ipiv]*a[apos];`

because `apos = 3493762`

but `fFact.N = 3493234`

Not sure what the best way is to solve this, so pinging @lmoneta 

The reproducer is:

```
#include "TFile.h"
#include "TDecompSparse.h"

void decomp() {
   auto f = TFile::Open("/tmp/matrix_fail.root");
   //f->ls();
   TMatrixTSparse<double>* M = (TMatrixTSparse<double>*)f->Get("M");
   //M->Print();
   TVectorT<double>* x = (TVectorT<double>*)f->Get("x");
   TDecompSparse solver(*M,0);
   Bool_t ok;
   TVectorD xx = solver.Solve(*x, ok);
}
```

This might help: https://www.hsl.rl.ac.uk/catalogue/index.html https://www.hsl.rl.ac.uk/archive-specs/ma47.pdf
[...]

PR #20367, opened 2025-11-10
[matrix] Fix for root-2927

mdessole:
# This Pull request:
fixes https://github.com/root-project/root/issues/14554

## Changes or fixes:
The matrix decomposition is not performed when the refactorization loop is executed more than once. This has been solved by moving the Decomposition check inside the loop.

## Checklist:

- [x] tested changes locally


guitargeek (approved):
Thank you very much for the precise fix!

PR #20370, opened 2025-11-10
[ntuple] deprecate small helpers in RFieldDescriptor

jblomer:
The query methods `IsCustomClass`, `IsCustomEnum`, and `IsStdAtomic` are rather arbitrary and should not have been added to the public interface. They only make sense in the internal context in which they are used.

Removed for v6.42.

### DIFF
--- a/tree/ntuple/src/RNTupleDescriptor.cxx
+++ b/tree/ntuple/src/RNTupleDescriptor.cxx
@@ -170,27 +170,12 @@ bool ROOT::RFieldDescriptor::IsCustomClass() const
 
 bool ROOT::RFieldDescriptor::IsCustomEnum(const RNTupleDescriptor &desc) const
 {
-   if (fStructure != ROOT::ENTupleStructure::kPlain)
-      return false;
-   if (fTypeName.rfind("std::", 0) == 0)
-      return false;
-
-   auto subFieldId = desc.FindFieldId("_0", fFieldId);
-   if (subFieldId == kInvalidDescriptorId)
-      return false;
-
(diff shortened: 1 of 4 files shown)


pcanal (approved):
LGTM.

tree/ntuple/src/RNTupleDescriptor.cxx
```diff
@@ -1419,7 +1390,7 @@ RNTupleSerializer::StreamerInfoMap_t ROOT::Internal::RNTupleDescriptorBuilder::B
 
    std::function<void(const RFieldDescriptor &)> fnWalkFieldTree;
    fnWalkFieldTree = [&desc, &streamerInfoMap, &fnWalkFieldTree](const RFieldDescriptor &fieldDesc) {
-      if (fieldDesc.IsCustomClass()) {
+      if (Internal::IsCustomClassFieldDesc(fieldDesc)) {
```
hahnjo: Given this goes away, do we even need `Internal::IsCustomClassFieldDesc`?

jblomer: As an implementation for the (deprecated) `IsCustomClass`, which (unfortunately) is/was part of the public API.

silverweed: Maybe we can add a comment on `IsCustomClass` to remind ourself to remove the internal helper when we remove the public API

jblomer: Oh, now I understand. Yes, for `IsCustomClass` we can leave the implementation where it is an not bother creating an internal function.

PR #20389, opened 2025-11-12
[rootcling] Remove check for `unique_ptr` members

### LINKED ISSUE #20377: Investigate I/O of `std::unique_ptr` on latest MacOS beta
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

As seen in https://github.com/root-project/root/pull/20354 , we don't yet fully understand why I/O of `std::unique_ptr` seems to misbehave on latest MacOS beta if the classes that it interacts with:
* Follow some dynamic polymorphism hierarchy
* Do not properly implement the destructors
* And do not include ClassDef macros (this last part especially seems to be a regression and contradicting what we say in our own docs)

### Reproducer

Not available yet

### ROOT version

Any

### Installation method

Any

### Operating system

MacOS Darwin Kernel Version 25.2.0

### Additional context

_No response_

hahnjo:
Since https://github.com/llvm/llvm-project/commit/f5e687d7bf49cd9fe38ba7acdeb52d4f30468dee libc++ implements `_LIBCPP_COMPRESSED_PAIR` with an anonymous struct (a Clang extension). This results in `unique_ptr` having a single `FieldDecl` with an empty name that does not appear as a data member in `TClass`. Instead, the AST has multiple `IndirectFieldDecl`s that "refer" into the anonymous struct.
    
In ROOT, we now have two options: We could support anonymous structs, either by allowing data members with empty names or by adding all indirect field declarations as members. However, I would argue that we actually do NOT want to support this extension for IO classes. Instead remove the (questionable) check in `rootcling` that verifies the number of data members in `unique_ptr`.

Closes #20377

### DIFF
--- a/core/metacling/src/TCling.cxx
+++ b/core/metacling/src/TCling.cxx
@@ -2764,15 +2764,6 @@ void TCling::InspectMembers(TMemberInspector& insp, const void* obj,
       return;
    }
 
-   if (TClassEdit::IsUniquePtr(cl->GetName())) {
-      // Ignore error caused by the inside of std::unique_ptr
-      // This is needed solely because of rootclingIO's IsUnsupportedUniquePointer
-      // which checks the number of elements in the GetListOfRealData.
-      // If this usage is removed, this can be replaced with a return statement.
-      // See https://github.com/root-project/root/issues/13574
-      isTransient = true;
-   }
-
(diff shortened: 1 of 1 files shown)


vgvassilev (approved):
Less is more! Nice, Lgtm!

PR #20402, opened 2025-11-12
Add more docs to the Snapshot with variations section

vepadulano:
Motivated by https://root-forum.cern.ch/t/documentation-on-bitmap-for-rdataframe-snapshot-with-variations/64397

### DIFF
--- a/tree/dataframe/src/RDataFrame.cxx
+++ b/tree/dataframe/src/RDataFrame.cxx
@@ -1264,11 +1264,30 @@ When Filters are employed, some variations might not pass the selection cuts (li
 In that case, RDataFrame will snapshot the filtered columns in a memory-efficient way by writing zero into the memory of fundamental types, or write a
 default-constructed object in case of classes. If none of the filters pass like in row 6, the entire event is omitted from the snapshot.
 
-To tell apart a genuine `0` (like `x` in row 0) from a variation that didn't pass the selection, RDataFrame writes a bitmask for each event, indicating which variations
-are valid (see last column). A mapping of column names to this bitmask is placed in the same file as the output dataset, and automatically loaded when
-RDataFrame opens a file that was snapshot with variations.
-Attempting to read such missing values with RDataFrame will produce an error, but RDataFrame can either skip these values or fill in defaults as
-described in the \ref missing-values "section on dealing with missing values".
+To tell apart a genuine `0` (like `x` in row 0) from a case where nominal or variation didn't pass a selection,
+RDataFrame writes a bitmask for each event, see last column of the table above. Every bit indicates whether its
+associated columns are valid. The bitmask is implemented as a 64-bit `std::bitset` in memory, written to the output
+dataset as a `std::uin64_t`. For every 64 columns, a new bitmask column is added to the output dataset.
+
+For each column that gets varied, the nominal and all variation columns are each assigned a bit to denote whether their
+entries are valid. A mapping of column names to the corresponding bitmask is placed in the same file as the output
+dataset, with a name that follows the pattern `"R_rdf_branchToBitmaskMapping_<NAME_OF_THE_DATASET>"`. It is of type
+`std::unordered_map<std::string, std::pair<std::string, unsigned int>>`, and maps a column name to the name of the
+bitmask column and the index of the relevant bit. For example, in the same file as the dataset "Events" there would be
+an object named `R_rdf_branchToBitmaskMapping_Events`. This object for example would describe a connection such as:
+
+~~~
+muon_pt --> (R_rdf_mask_Events_0, 42)
+~~~
+
+which means that the validity of the entries in `muon_pt` is established by the bit `42` in the bitmask found in the
+column `R_rdf_mask_Events_0`.
+
+When RDataFrame opens a file, it checks for the existence of this mapping between columns and bitmasks, and loads it automatically if found. As such,
+RDataFrame makes the treatment of the various bitmap maskings completely transparent to the user.
+
+In case certain values are labeled invalid by the corresponding bit, this will result in reading a missing value. The semantics of such a scenario follow the
+rules described in the \ref missing-values "section on dealing with missing values" and can be dealt with accordingly.
 
 \note Snapshot with variations is currently restricted to single-threaded TTree snapshots.
 
@@ -1780,6 +1799,9 @@ more of its entries. For example:
 - When joining different datasets horizontally according to some index value
   (e.g. the event number), if the index does not find a match in one or more
   other datasets for a certain entry.
+- If, for a certain event, a column is invalid because it results from a Snapshot
+  with systematic variations, and that variation didn't pass its filters. For
+  more details, see \ref snapshot-with-variations.
 
 For example, suppose that column "y" does not have a value for entry 42:
 


tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1780,6 +1795,9 @@ more of its entries. For example:
 - When joining different datasets horizontally according to some index value
   (e.g. the event number), if the index does not find a match in one or more
   other datasets for a certain entry.
+- If, for a certain event, the value of a certain column is invalid because
+  it results from a previous processing which involved systematic variations
+  and that value was removed by a selection. For more details, see \ref snapshot-with-variations.
```
```suggestion
- If, for a certain event, a column is invalid because it results from a snapshot with systematic variations, and that variation didn't pass its filters. For more details, see \ref snapshot-with-variations.
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1265,10 +1265,25 @@ In that case, RDataFrame will snapshot the filtered columns in a memory-efficien
 default-constructed object in case of classes. If none of the filters pass like in row 6, the entire event is omitted from the snapshot.
 
 To tell apart a genuine `0` (like `x` in row 0) from a variation that didn't pass the selection, RDataFrame writes a bitmask for each event, indicating which variations
-are valid (see last column). A mapping of column names to this bitmask is placed in the same file as the output dataset, and automatically loaded when
-RDataFrame opens a file that was snapshot with variations.
-Attempting to read such missing values with RDataFrame will produce an error, but RDataFrame can either skip these values or fill in defaults as
-described in the \ref missing-values "section on dealing with missing values".
+are valid (see last column). The bitmask is implemented as a 64-bit `std::bitset` in memory, written to the output dataset as a `std::uin64_t`.
+Thus, every 64 columns a new bitmask must be written to accommodate the bits for the next 64 columns.
```
```suggestion
For every 64 columns, a new bitmask column is added to the output dataset.
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1265,10 +1265,25 @@ In that case, RDataFrame will snapshot the filtered columns in a memory-efficien
...
+Each column stored in the output is connected to exactly one bit in one bitmask. A mapping of column names to the corresponding bitmask is placed in the
+same file as the output dataset, with a name that follows the pattern `"R_rdf_branchToBitmaskMapping_" + NAME_OF_THE_DATASET`. In this mapping, each
+column name is connected to one bitmask, and one particular bit in that bitmask. For example, in the same file as the dataset "Events" there would be
+an object named `R_rdf_branchToBitmaskMapping_Events`. This object for example would describe a connection such as:
+
+~~~
+muon_pt --> (R_rdf_mask_Events_0, 42)
+~~~
+
+which means that the validity of the column `muon_pt` is established by the bit `42` in the bitmask found in the column `R_rdf_mask_Events_0`.
```
```suggestion
Each column that might contain invalid values is connected to exactly one bit in one bitmask. A mapping of column names to the corresponding bitmask is placed in the
same file as the output dataset, with a name that follows the pattern `"R_rdf_branchToBitmaskMapping_<NAME_OF_THE_DATASET>"`. It is of type `std::unordered_map<std::string, std::pair<std::string, unsigned int>>`,
and maps a column name to the name of the bitmask column and the index of the relevant bit. For example, in the same file as the dataset "Events" there would be
an object named `R_rdf_branchToBitmaskMapping_Events`. This object for example would describe a connection such as:

~~~
muon_pt --> (R_rdf_mask_Events_0, 42)
~~~

which means that the validity of the entries in `muon_pt` is established by the bit `42` in the bitmask found in the column `R_rdf_mask_Events_0`.
```

rbarrue:
Hi team, looks good to me. To be completely clear, in the sentence
> To tell apart a genuine `0` (like `x` in row 0) from a variation that didn't pass the selection, RDataFrame writes a bitmask for each event, indicating which variations are valid (see last column)

 to

> (...) that didn't pass the selection (resulting in an invalid value) (...)

Also, in the sentence
> Each column that might contain invalid values is connected to exactly one bit in one bitmask.

It should read simply "each column is connected to exactly one bit in one bitmask", right?

hageboeck:
Hello @rbarrue, thanks for the feedback! See a question and an answer inline.


> Hi team, looks good to me. To be completely clear, in the sentence
> 
> > To tell apart a genuine `0` (like `x` in row 0) from a variation that didn't pass the selection, RDataFrame writes a bitmask for each event, indicating which variations are valid (see last column)
> 
> to
> 
> > (...) that didn't pass the selection (resulting in an invalid value) (...)
> 
I'm not fully sure if it's a request to update the text. Did you mean this?

```diff
- To tell apart a genuine `0` (like `x` in row 0) from a variation that didn't pass the selection, RDataFrame writes a bitmask for each event, indicating which variations are valid (see last column).
+ To tell apart a genuine `0` (like `x` in row 0) from a variation that didn't pass the selection (resulting in an invalid value), RDataFrame writes a bitmask for each event, indicating which variations are valid (see last column).
```

We can of course rephrase to make the sentence flow better, but now we're doubling the information, aren't we?

> Also, in the sentence
> 
> > Each column that might contain invalid values is connected to exactly one bit in one bitmask.
> 
> It should read simply "each column is connected to exactly one bit in one bitmask", right?

In this case the qualifier was added on purpose, because columns that are not being varied are always written. Therefore, we are not allocating any bit for these.

rbarrue:
Yes, it was a "request" for rephrasing. Yeah, we're doubling the information, one could finish instead with "variations passed the selection" ?

For the second point, if we're not allocating any bit for them, I'd actually change the example line:
> muon_pt --> (R_rdf_mask_Events_0, 42)

to 

> muon_pt_var1 --> (...)

At least for me the first line looks like we're also storing a bitmask for nominal.

hageboeck:
> Yes, it was a "request" for rephrasing. Yeah, we're doubling the information, one could finish instead with "variations passed the selection" ?
> 
Let's go to the actual sentence in the diff. I'll give it another try.

> For the second point, if we're not allocating any bit for them, I'd actually change the example line:
> 
> > muon_pt --> (R_rdf_mask_Events_0, 42)
> 
> to
> 
> > muon_pt_var1 --> (...)
> 
> At least for me the first line looks like we're also storing a bitmask for nominal.

This is indeed the case. If a column gets varied and filtered, all of its instances (nominal and variations) will be associated to a bit, because nominal might not pass while one of the variations doesn't.

That being said, columns that don't get varied won't have a bit. I'll try to clarify the text once more to point this out.

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1265,10 +1265,28 @@ In that case, RDataFrame will snapshot the filtered columns in a memory-efficien
...
 
 To tell apart a genuine `0` (like `x` in row 0) from a variation that didn't pass the selection, RDataFrame writes a bitmask for each event, indicating which variations
-are valid (see last column). A mapping of column names to this bitmask is placed in the same file as the output dataset, and automatically loaded when
-RDataFrame opens a file that was snapshot with variations.
-Attempting to read such missing values with RDataFrame will produce an error, but RDataFrame can either skip these values or fill in defaults as
-described in the \ref missing-values "section on dealing with missing values".
+are valid (see last column). The bitmask is implemented as a 64-bit `std::bitset` in memory, written to the output
+dataset as a `std::uin64_t`. For every 64 columns, a new bitmask column is added to the output dataset.
+
+Each column that might contain invalid values is connected to exactly one bit in one bitmask. A mapping of column names
```
```suggestion
For each column that gets varied, the nominal and all variation columns are each assigned a bit to denote whether their entries are valid. A mapping of column names
```

tree/dataframe/src/RDataFrame.cxx
```diff
@@ -1265,10 +1265,28 @@ In that case, RDataFrame will snapshot the filtered columns in a memory-efficien
 default-constructed object in case of classes. If none of the filters pass like in row 6, the entire event is omitted from the snapshot.
 
 To tell apart a genuine `0` (like `x` in row 0) from a variation that didn't pass the selection, RDataFrame writes a bitmask for each event, indicating which variations
-are valid (see last column). A mapping of column names to this bitmask is placed in the same file as the output dataset, and automatically loaded when
-RDataFrame opens a file that was snapshot with variations.
-Attempting to read such missing values with RDataFrame will produce an error, but RDataFrame can either skip these values or fill in defaults as
-described in the \ref missing-values "section on dealing with missing values".
+are valid (see last column). The bitmask is implemented as a 64-bit `std::bitset` in memory, written to the output
+dataset as a `std::uin64_t`. For every 64 columns, a new bitmask column is added to the output dataset.
```
```suggestion
To tell apart a genuine `0` (like `x` in row 0) from a case where nominal or variation didn't pass a selection, RDataFrame writes a bitmask for each event, see last column of the table above.
Every bit indicates whether its associated columns are valid. The bitmask is implemented as a 64-bit `std::bitset` in memory, written to the output
dataset as a `std::uin64_t`. For every 64 columns, a new bitmask column is added to the output dataset.
```

rbarrue:
Looks great to me, thanks.

Issue #20287, opened 2025-11-05
Can no longer python import libcppyy


andresailer:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Importing libcppyy no longer works in python in the ROOT master branch (at least when things are installed on CVMFS, the "x" bit was lost maybe when it was moved into the subfolder)?
```
ls -l /cvmfs/sft-nightlies.cern.ch/lcg/nightlies/dev3/Wed/ROOT/HEAD/x86_64-el9-gcc15-dbg/lib/cppyy/libcppyy.so  /cvmfs/sft.cern.ch/lcg/releases/LCG_108a/ROOT/6.36.04/x86_64-el9-gcc15-dbg/lib/libcppyy.so 
-rw-r--r--. 1 cvmfs cvmfs    54832 Nov  5 00:42 /cvmfs/sft-nightlies.cern.ch/lcg/nightlies/dev3/Wed/ROOT/HEAD/x86_64-el9-gcc15-dbg/lib/cppyy/libcppyy.so
-rwxr-xr-x. 1 cvmfs cvmfs 10823672 Nov  3 17:16 /cvmfs/sft.cern.ch/lcg/releases/LCG_108a/ROOT/6.36.04/x86_64-el9-gcc15-dbg/lib/libcppyy.so
```

Checking with strace the library is found, but  "readlink" fails
```
openat(AT_FDCWD, "/cvmfs/sft-nightlies.cern.ch/lcg/views/dev3/Wed/x86_64-el9-gcc15-dbg/lib/cppyy/libcppyy.so", O_RDONLY|O_CLOEXEC) = 3
...
readlink("/cvmfs/sft-nightlies.cern.ch/lcg/nightlies/dev3/Wed/ROOT/HEAD/x86_64-el9-gcc15-dbg/lib/cppyy/libcppyy.so", 0x7ffd7ab10eb0, 1023) = -1 EINVAL (Invalid argument)
```
Errors, I don't know enough if that is to be expected or not...

`ldd` also complains

```
ldd /cvmfs/sft-nightlies.cern.ch/lcg/nightlies/dev3/Wed/ROOT/HEAD/x86_64-el9-gcc15-dbg/lib/cppyy/libcppyy.so
/cvmfs/sft-nightlies.cern.ch/lcg/nightlies/dev3/Wed/ROOT/HEAD/x86_64-el9-gcc15-dbg/lib/cppyy/libcppyy.so:
[...]

PR #20404, opened 2025-11-12
[Python] Use `cppyy.types` module instead of private `cppyy._backend`

guitargeek:
Do what @wlav suggested in
https://github.com/root-project/root/issues/20287#issuecomment-3492420757.

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/__init__.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/__init__.py
@@ -8,9 +8,21 @@
 # For the list of contributors see $ROOTSYS/README/CREDITS.                    #
 ################################################################################
 
-import importlib
+import atexit
+import builtins
 import os
 import sys
+import types
+from importlib.abc import Loader, MetaPathFinder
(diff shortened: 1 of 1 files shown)


hahnjo:
This (probably the second commit) broke Python on Windows, with hundreds of failing tests:
```
ImportError: DLL load failed while importing libROOTPythonizations: The specified module could not be found.
```
It was already clearly visible in the PR builds and should not have been merged like this. Moreover a backport was already included in https://github.com/root-project/root/pull/20405 before this even landed. We seem to have been lucky and the backported commit (only the first one) seems to work, but I would strongly argue that this is the wrong procedure to follow. Ideally all backports should be applied with `git cherry-pick -x` to show the commit hash *after it has landed in `master`*.

hahnjo:
Partial revert in https://github.com/root-project/root/pull/20411

guitargeek:
Thanks, I didn't know about `cherry-pick -x`! I'll do that from now on, it's indeed useful to have the commit in master referenced

PR #20406, opened 2025-11-12
[CMake] Run TPython C++ test also on Windows

guitargeek:


guitargeek:
Even if the problem with `std::make_any` could be solved, there would be more problems in the Python tests because on Windows because MSVC’s standard library does not export the symbols needed for cling to JIT-compile code involving std::any, causing symbol resolution to fail.

So a more sustainable solution would be needed, but the priority is low given that theses tests never worked on Windows and I'm not sure if there are users of TPython on Windows.

PR #20407, opened 2025-11-12
[CMake] Don't set LD_LIBRARY_PATH in test macros

guitargeek:
This should not be needed anymore because of the runpath mechanism, and by leaving it out from the test environment we are making sure that this keeps working.

### DIFF
--- a/roottest/root/ntuple/makeproject/rntuple/CMakeLists.txt
+++ b/roottest/root/ntuple/makeproject/rntuple/CMakeLists.txt
@@ -53,17 +53,8 @@ target_link_directories(read_rntuple PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/librntu
 
 # Need to also explicitly set the LD_LIBRARY_PATH (at least on MacOS), so that
 # the generated library with the custom class can be loaded by the read test.
-# Setting the value of ${ld_library_path} in the ENVIRONMENT argument of the
-# ROOTTEST_ADD_TEST function will not work, because that function sets the same
-# variable internally, thus overriding our choice. The simplest thing to do is
-# to append our directory to the ${ROOTTEST_ENV_LIBRARYPATH} contents, which
-# are automatically used by ROOTTEST_ADD_TEST
-set(_roottest_env_librarypath ${ROOTTEST_ENV_LIBRARYPATH})
-set(ROOTTEST_ENV_LIBRARYPATH "${ROOTTEST_ENV_LIBRARYPATH}:${CMAKE_CURRENT_BINARY_DIR}/librntuplestltest")
 ROOTTEST_ADD_TEST(read_rntuple
                   EXEC ./read_rntuple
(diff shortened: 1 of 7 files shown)


hageboeck:
The failures are understandable, because the "main" libraries and executables don't contain runpaths to find libraries that are part of tests. For these, it's probably OK to continue setting LD_LIBRARY_PATH, but `$ROOTSYS/lib` should not be part of it.

guitargeek:
> The failures are understandable, because the "main" libraries and executables don't contain runpaths to find libraries that are part of tests. For these, it's probably OK to continue setting LD_LIBRARY_PATH, but `$ROOTSYS/lib` should not be part of it.

Ok makes sense! I'll see what can be done about that.

guitargeek:
Test failure is unrelated. It's because I had to remove chromium from the rawhide image, and there was no clean CMake re-configuration yet that recognized that there is no chrome browser available.

roottest/root/ntuple/makeproject/rntuple/CMakeLists.txt
```diff
@@ -53,17 +53,8 @@ target_link_directories(read_rntuple PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/librntu
...
-set(_roottest_env_librarypath ${ROOTTEST_ENV_LIBRARYPATH})
-set(ROOTTEST_ENV_LIBRARYPATH "${ROOTTEST_ENV_LIBRARYPATH}:${CMAKE_CURRENT_BINARY_DIR}/librntuplestltest")
 ROOTTEST_ADD_TEST(read_rntuple
                   EXEC ./read_rntuple
                   FIXTURES_REQUIRED read_rntuple_executable
                   # PATH is used on Windows to find libraries for loading
-                  ENVIRONMENT PATH=${CMAKE_CURRENT_BINARY_DIR}/librntuplestltest)
-set(ROOTTEST_ENV_LIBRARYPATH ${_roottest_env_librarypath})
-
+                  ENVIRONMENT ${ld_library_path}=${CMAKE_CURRENT_BINARY_DIR}/librntuplestltest)
```
hageboeck, line 60: Here it looks like the PATH disappeared. Was that intended?

guitargeek: Yes, that was intended, because it was redundant. As far as I can tell, the "PATH" is only needed for Windows because there is also functions as the library path. However, `${ld_library_path}` is defined as `PATH` on Windows inside RootMacros.cmake, so we don't need the hardcoded `PATH=`.

hageboeck: Indeed, there don't seem to be any executables in that directory in my build.

hageboeck (approved):

devajithvs:
This PR broke builds with `-Dclingtest=ON`, I'm investigating.

guitargeek:
Thanks!

PR #20414, opened 2025-11-13
[Python] Make the linter happy about `ROOT/__init__.py`

guitargeek:
This is done now because it prevents various distracting linter warnings in `ROOT/__init__.py` in the "Files changed" section in the PR view when this file is changed.

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/__init__.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/__init__.py
@@ -8,9 +8,21 @@
 # For the list of contributors see $ROOTSYS/README/CREDITS.                    #
 ################################################################################
 
-import importlib
+import atexit
+import builtins
 import os
 import sys
+import types
+from importlib.abc import Loader, MetaPathFinder
+from importlib.machinery import ModuleSpec
+from typing import Optional, Union
+
+import cppyy
+import cppyy.types
+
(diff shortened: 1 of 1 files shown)


bindings/pyroot/pythonizations/python/ROOT/__init__.py
```diff
@@ -178,11 +173,11 @@ def find_spec(self, fullname: str, path, target=None) -> ModuleSpec:
 
     ip = get_ipython()
     if hasattr(ip, "kernel"):
-        import JupyROOT
+        import JupyROOT  # noqa: F401  # imported the side effect of setting up JupyROOT
```
vepadulano, line 176: I don't understand the comment here. If it is about the lazy import at function scope, why don't we need the same noqa in other places?

And more in general, is the linter happy with lazy imports at function scope?

guitargeek: Yes, the linter is generally happy with lazy imports at function scope.

The warning suppression here is not about imports. It's about the unused module `JupyROOT`, which is fine because we import the module not to use is, but for the side effects in `JupyROOT/__init__.py`.

vepadulano: Thank you, I really missed the point of the comment, thanks for the explanation!

PR #20418, opened 2025-11-13
[hist] Implement `RHist{,Engine,Stats}::AddAtomic`

hahnjo:
... as building blocks for concurrent filling of `RHist`.

### DIFF
--- a/hist/histv7/test/hist_hist.cxx
+++ b/hist/histv7/test/hist_hist.cxx
@@ -69,6 +69,28 @@ TEST(RHist, AddAtomic)
    EXPECT_EQ(histA.GetBinContent(RBinIndex(9)), 1);
 }
 
+TEST(RHist, StressAddAtomic)
+{
+   static constexpr std::size_t NThreads = 4;
+   static constexpr std::size_t NAddsPerThread = 10000;
+   static constexpr std::size_t NAdds = NThreads * NAddsPerThread;
+
+   // Fill a single bin, to maximize contention.
+   const RRegularAxis axis(1, {0, 1});
+   RHist<int> histA({axis});
+   RHist<int> histB({axis});
+   histB.Fill(0.5);
+
+   StressInParallel(NThreads, [&] {
+      for (std::size_t i = 0; i < NAddsPerThread; i++) {
+         histA.AddAtomic(histB);
+      }
+   });
+
+   EXPECT_EQ(histA.GetNEntries(), NAdds);
+   EXPECT_EQ(histA.GetBinContent(0), NAdds);
+}
+
 TEST(RHist, Clear)
 {
    static constexpr std::size_t Bins = 20;
--- a/hist/histv7/test/hist_engine_atomic.cxx
+++ b/hist/histv7/test/hist_engine_atomic.cxx
@@ -1,5 +1,8 @@
 #include "hist_test.hxx"
 
+#include <atomic>
+#include <cstddef>
+
 TEST(RHistEngine, AddAtomic)
 {
    static constexpr std::size_t Bins = 20;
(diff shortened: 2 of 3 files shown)


hist/histv7/inc/ROOT/RHistEngine.hxx
```diff
@@ -196,6 +196,21 @@ public:
       }
    }
 
+   /// Add all bin contents of another histogram using atomic instructions.
+   ///
+   /// Throws an exception if the axes configurations are not identical.
+   ///
+   /// \param[in] other another histogram
+   void AddAtomic(const RHistEngine<BinContentType> &other)
```
hageboeck, line 204: One should maybe clarify that an undefined state of `other` ends up in `*this` if `other` continues to be written to while `AddAtomic` executes.

hist/histv7/inc/ROOT/RBinWithError.hxx
```diff
@@ -59,6 +59,12 @@ struct RBinWithError final {
       Internal::AtomicAdd(&fSum, w);
       Internal::AtomicAdd(&fSum2, w * w);
    }
+
+   void AtomicAdd(const RBinWithError &rhs)
+   {
+      Internal::AtomicAdd(&fSum, rhs.fSum);
+      Internal::AtomicAdd(&fSum2, rhs.fSum2);
+   }
```
hageboeck, line 70: Same problem as for `RHistEngine::AddAtomic`. One should not expect to be able to write to `rhs` while `AtomicAdd` runs.

hist/histv7/inc/ROOT/RHist.hxx
```diff
@@ -174,6 +174,17 @@ public:
       fStats.Add(other.fStats);
    }
 
+   /// Add all bin contents and statistics of another histogram using atomic instructions.
```
hageboeck, line 177: Add clarification that one should not write to `other` concurrently?

hist/histv7/inc/ROOT/RHistStats.hxx
```diff
@@ -125,6 +133,24 @@ public:
       }
    }
 
+   /// Add all entries from another statistics object using atomic instructions.
```
hageboeck, line 139: Idem

hist/histv7/test/hist_hist.cxx
```diff
@@ -52,6 +52,23 @@ TEST(RHist, Add)
...
+   histA.Fill(8.5);
+   histB.Fill(9.5);
+
+   histA.AddAtomic(histB);
+
+   EXPECT_EQ(histA.GetNEntries(), 2);
+   EXPECT_EQ(histA.GetBinContent(RBinIndex(8)), 1);
+   EXPECT_EQ(histA.GetBinContent(RBinIndex(9)), 1);
+}
+
```
hageboeck, line 71: Would it make sense to add a test that progressively fills A asynchronously, while B is added atomically?

Roughly:
```c++
auto future = std::async(std::launch::async, [&](){
   for (std::size_t i = 0; i < N; ++i) {
      histA.Fill(...);
      if (i == 5) // unblock other thread
   }
};

 while (blocked);
 histA.AddAtomic(histB);

 future.get();

 EXPECT_EQ(...);
```

hist/histv7/inc/ROOT/RHistEngine.hxx
hahnjo: Yes, that's correct, I added an even stronger "must not be modified during the operation" to the documentation because I don't think there is a meaningful way to make sense of a partially added histogram while it was filled.

FWIW I plan to have a dedicated section in `doc/ConcurrentFilling.md` about guarantees, in particular which functions can be combined with each other. In short, it is allowed to call the same `*Atomic` function concurrently on the same histogram but for other combinations we will have an explicit list. For example, `AddAtomic` and `FillAtomic` (on `this`, not `other`) is ok and we can give guarantees for `FillAtomic` and `SnapshotAtomic`, but I don't think I see a way to make `AddAtomic` and `SnapshotAtomic` work in a meaningful way (or at least it won't be there in the beginning).

hist/histv7/test/hist_hist.cxx
hahnjo: I added five `Stress` tests in the last commit pushed to the PR, three in `test/hist_engine_atomic.cxx` and one each in `test/hist_stats.cxx` and `test/hist_hist.cxx`. Note that `RHist` cannot be (directly) filled from multiple threads, but there is `RHistEngine.StressFillAddAtomicWeight` where half of the threads call `FillAtomic` with a weight and the other half calls `AddAtomic` of a "well-prepared" histogram.

hageboeck (approved):
:+1:

PR #20430, opened 2025-11-14
[CMake] Extend ROOT_FIND_PYTHON_MODULE to also save version info

guitargeek:
This is useful to steer testing depending on the version of Python modules, or give errors/warnings if a module has a version unsupported by ROOT.


vepadulano (approved):
I think it makes sense. I believe the Fedora failure is unrelated.

PR #20435, opened 2025-11-14
[ci] Also test DistRDF on Fedora 43

guitargeek:
It's better to test also on the newest Fedora release, to make sure it still works with the newest Python and other packages.


guitargeek:
Didn't work because of incompatible Java version.

Issue #20383, opened 2025-11-11
Broken rf501_simultaneouspdf.py  in dev3/nighlies


VanyaBelyaev:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Tutorial file rf501_imultaneouspdf.py inn dev3/latetx nightlies slot is broken.

central and right pltos  do not show the fit curves.

Reproducer is simple:
```
( source /cvmfs/sft.cern.ch/lcg/views/dev3/latest/x86_64-el9-clang19-opt/setup.sh ; 
ipython -i $ROOTSYS/tutorials/roofit/roofit/rf501_simultaneouspdf.py )

```

LCG_108 (and other) LCG slots are OK (all three plots show the fit curves) 

``` 
( source /cvmfs/sft.cern.ch/lcg/views/LCG_108/x86_64-el9-clang19-opt/setup.sh ; 
ipython -i $ROOTSYS/tutorials/roofit/roofit/rf501_simultaneouspdf.py )
```

### Reproducer

Reproducer is simple:
```
( source /cvmfs/sft.cern.ch/lcg/views/dev3/latest/x86_64-el9-clang19-opt/setup.sh ; 
ipython -i $ROOTSYS/tutorials/roofit/roofit/rf501_simultaneouspdf.py )

```

LCG_108 (and other) LCG slots are OK (all three plots show the fit curves) 

``` 
( source /cvmfs/sft.cern.ch/lcg/views/LCG_108/x86_64-el9-clang19-opt/setup.sh ; 
ipython -i $ROOTSYS/tutorials/roofit/roofit/rf501_simultaneouspdf.py )
```

### ROOT version

ROOT from dev3/latest LCG nightly slot 

### Installation method

cvmsf  LCG nighlied dev2/latet lot 

### Operating system

Linux, lxplus

guitargeek:
Thank you very much for the report! I identified a8ef8b03c84 as the origin of the problem. I'm investigating.

VanyaBelyaev:
Dear Jonas @guitargeek 
It is nice to know  the problem is promptly localized.
Hope to see  a fix soon.
cheers, Vanya

VanyaBelyaev:
One more issue with the same example.

With "new" ROOT there are many warning messages on duplicated arguments:

The comand : 
```
[...]

PR #20441, opened 2025-11-16
[RF] Correctly normalize RooSimultaneous when projecting over data

guitargeek:
Since https://github.com/root-project/root/commit/a8ef8b03c845b7ddd6aad266e24d745272b9ef82, any RooAbsPdf that makes a prediction for the expected number of events will use this prediction for the normalization when plotting.

This made is more meaningful to compare models with data because the model normalization was not automatically scaled to match the data.

However, it also broke plots with projections of RooSimultaneous pdfs over data, because the temporary RooAddPdf that is created in this case did not use coefficients that were scaled to the actual dataset, but to unity. Correctly normalizing to the sum of entries in the dataset fixes this problem.

A unit test is also implemented.

Closes https://github.com/root-project/root/issues/20383.

### DIFF
--- a/roofit/roofitcore/src/RooAbsPdf.cxx
+++ b/roofit/roofitcore/src/RooAbsPdf.cxx
@@ -210,6 +210,15 @@ inline double getLog(double prob, RooAbsReal const *caller)
    return std::log(prob);
 }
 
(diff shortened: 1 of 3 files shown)


lmoneta (approved):
LGTM!
Thank you Jonas to fix this imporatnt issue.

PR #20442, opened 2025-11-16
[RF] Return reference to actual RooAbsReal in `RooPolyVar::x()` getter

### LINKED ISSUE #20293: RooPolyVar.x() yields a default variable Name of "x"
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

When running RooPolyVar().x(), the proxy has a default name of "x" despite different variable name.

### Reproducer

```python
import ROOT

y = ROOT.RooRealVar("y", "y", 1, -5, 5)

a = ROOT.RooRealVar("a", "a", 2, -5, 5)
b = ROOT.RooRealVar("b", "b", 3, -5, 5)

func = ROOT.RooPolyVar("myFunc", "myFunc", y, [a, b])
func.Print("t")
print("Function Variable:", func.x().GetName())
```

### ROOT version

6.37.01

### Installation method

build from source

### Operating system

ArchLinux

### Additional context

_No response_

guitargeek:
Just like in any other getter for the inputs of RooFit functions, the `RooPolyVar::x()` getter should return the underlying RooAbsReal, instead of the internal proxy.

Closes #20293.

Thanks to @Phmonski for noticing this!

### DIFF
--- a/roofit/roofitcore/inc/RooPolyVar.h
+++ b/roofit/roofitcore/inc/RooPolyVar.h
@@ -34,7 +34,7 @@ public:
    Int_t getAnalyticalIntegral(RooArgSet &allVars, RooArgSet &analVars, const char *rangeName = nullptr) const override;
    double analyticalIntegral(Int_t code, const char *rangeName = nullptr) const override;
 
-   RooRealProxy const &x() const { return _x; }
+   RooAbsReal const &x() const { return _x.arg(); }
    RooArgList const &coefList() const { return _coefList; }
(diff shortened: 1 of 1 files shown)


lmoneta (approved):
LGTM!
I agree it is better being consistent. Can this be a backwaerd compatibility issue for somebody or is this function just used only internally?

guitargeek:
Thanks for the review! This interface was only introduced in ROOT 6.36 with 96e023b7ae0d4df3143457e606f10ff8370af9d9, and so far only used internally. So I think it's not too late to change it.

PR #20453, opened 2025-11-18
[RF][HF] Error out in HistFactory in case of duplicate channel names

guitargeek:
Duplicate channel names are not allowed, because it would result in name collisions for the channel pdfs.

Also make the logging more consistent in a separate commit.

FYI, @TomasDado


guitargeek:
Just fixed a typo in the commit message.

PR #20456, opened 2025-11-18
[roottest] Use nbconvert as a library in nbdiff.py

guitargeek:
Using nbconvert as a Python library is less fragile, because otherwise there is ambiguity how the command line util should be called (the name changed over the versions as it migrated from ipython to jupyter).

This will make it possible to run the notebook tests also on the Fedora Rawhide build that is linked against the Python debug version, and hopefully also fix the test failures on the `macOS` platforms when a Python interpreter other than the system Python is used.

### DIFF
--- a/roottest/python/JupyROOT/nbdiff.py
+++ b/roottest/python/JupyROOT/nbdiff.py
@@ -7,9 +7,9 @@ import tempfile
 
 
 # Replace the criterion according to which a line shall be skipped
-def customLineJunkFilter(line):
+def should_keep_line(line):
     # Skip the banner and empty lines
-    junkLines = [
+    skip_patterns = [
         "Info in <TUnixSystem::ACLiC",
         "Info in <TMacOSXSystem::ACLiC",
         "FAILED TO establish the default connection to the WindowServer",
@@ -18,8 +18,8 @@ def customLineJunkFilter(line):
         '     "execution_count":',
         "libclang_rt.asan-",
     ]
-    for junkLine in junkLines:
-        if junkLine in line:
+    for pattern in skip_patterns:
+        if pattern in line:
             return False
     return True
 
@@ -44,7 +44,7 @@ def removeCellMetadata(lines):
 
 def getFilteredLines(fileName):
     with open(fileName) as f:
-        filteredLines = list(filter(customLineJunkFilter, f.readlines()))
+        filteredLines = list(filter(should_keep_line, f.readlines()))
 
     # Sometimes the jupyter server adds a new line at the end of the notebook
     # and nbconvert does not.
@@ -136,10 +136,7 @@ def addEtcToEnvironment(inNBDirName):
 def getKernelName(inNBName):
     with open(inNBName) as f:
         nbj = json.load(f)
-    if nbj["metadata"]["kernelspec"]["language"] == "python":
-        return "python3"
-    else:  # we support only Python and C++
-        return "root"
+    return nbj["metadata"]["kernelspec"]["name"]
 
 
 def canReproduceNotebook(inNBName, needsCompare):
--- a/roottest/python/JupyROOT/importROOT.ipynb
+++ b/roottest/python/JupyROOT/importROOT.ipynb
@@ -14,21 +14,21 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 2",
+   "display_name": "Python 3",
    "language": "python",
-   "name": "python2"
+   "name": "python3"
   },
   "language_info": {
    "codemirror_mode": {
     "name": "ipython",
-    "version": 2
+    "version": 3
    },
    "file_extension": ".py",
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython2",
-   "version": "2.7.6"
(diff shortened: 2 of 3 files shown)


roottest/python/JupyROOT/nbdiff.py
```diff
@@ -2,94 +2,76 @@
...
-  "-f",
-  "{connection_file}"
- ]
-}
-''' %pythonInterpName
-
-
 
 # Replace the criterion according to which a line shall be skipped
 def customLineJunkFilter(line):
```
pcanal:
```suggestion
def customLineFilter(line):
```

roottest/python/JupyROOT/nbdiff.py
```diff
@@ -2,94 +2,76 @@
...
     # Skip the banner and empty lines
-    junkLines =['Info in <TUnixSystem::ACLiC',
-                'Info in <TMacOSXSystem::ACLiC',
-                'FAILED TO establish the default connection to the WindowServer',
-                '"version": ',
-                '"pygments_lexer": "ipython',
-                '     "execution_count":',
-                'libclang_rt.asan-']
+    junkLines = [
```
pcanal:
```suggestion
    filteredLines = [
```
Consider using a non-judgemental name :)

roottest/python/JupyROOT/nbdiff.py
```diff
@@ -149,39 +133,53 @@ def addEtcToEnvironment(inNBDirName):
...
-    ret = subprocess.call("type jupyter", shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    return "jupyter" if ret == 0 else "i%s" % pythonInterpName
-
-
 def getKernelName(inNBName):
     with open(inNBName) as f:
         nbj = json.load(f)
     if nbj["metadata"]["kernelspec"]["language"] == "python":
-        return pythonInterpName
+        return "python3"
```
pcanal: Is `python3` guaranteed to be available? i.e. is there any installation allowed to only install `python` rather than `pyhon3` (or both)?

roottest/python/JupyROOT/nbdiff.py
guitargeek: Will do!

roottest/python/JupyROOT/nbdiff.py
```diff
@@ -149,39 +133,53 @@ def addEtcToEnvironment(inNBDirName):
...
-        return exitStatus
+        return 0  # success
+
+    shutil.rmtree(tmpDir)
 
 
 def isInputNotebookFileName(filename):
     if not filename.endswith(".ipynb"):
-        print("Notebook files shall have the %s extension" % nbExtension)
+        print("Notebook files shall have the .ipynb extension")
```
pcanal: Out of curiosity, why is the extension no longer need to be 'flexible' (it is clear from the previous code that this was 'just in case' and never actually leveraged, the question is what (if anything) changed that make the 'just case in' no longer needed) ... and/or is the code really more readable overall by replacing the semantic variable `nbExtension` by its value throughout?

roottest/python/JupyROOT/nbdiff.py
```diff
@@ -200,13 +198,5 @@ def isInputNotebookFileName(filename):
...
-    try:
-        # If jupyter is there, ipython is too
-        import jupyter
-    except:
-        raise ImportError("Cannot import jupyter")
-
-    kernelName = getKernelName(nbFileName)
-
-    retCode = canReproduceNotebook(nbFileName, kernelName, needsCompare)
+    retCode = canReproduceNotebook(nbFileName, needsCompare)
```
pcanal, line 198: Is this a change in behavior or was the 'kernalName' unused or is the kernel name implied by the file name inside `canReproduceNotebook` ?

pcanal (approved):
LFTM (but see the set of questions)

roottest/python/JupyROOT/nbdiff.py
guitargeek: I saw it was hardcoded in some places, and taken from a variable name in other places, so it was never flexible. I just wanted to have it consistent.

roottest/python/JupyROOT/nbdiff.py
guitargeek: The `python3` kernel is always guaranteed to be available, because we only support Python 3 at this point. But you are right, this can be done better by updating the reference notebooks and taking the kernel name from the `kernelspec` directly.

roottest/python/JupyROOT/nbdiff.py
guitargeek: Yes, the `kernelName` was only used inside `canReproduceNotebook`, so there was no need for having that variable in the outer scope.

guitargeek:
Thanks for the review! I have added one commit that addresses your comments. In particular, the reference notebooks are not updated to Python 3, so we don't have to manually set the kernel name in the `nbdiff.py` script.

guitargeek:
The failures in Fedora Rawhide are unrelated: they come from an update of the system `gtest`, and because the builds are incremental it still tries to link against the old version.

Maybe it's possible to also "hash" the docker image somehow, and when it changed do a clean build. To be seen later if this should continue to annoy us.

PR #20458, opened 2025-11-18
[ci] Don't set `CMAKE_BUILD_RPATH` on macOS

guitargeek:
This was needed to find the Fortran libraries, but since 31613210d6 the RPATH is appended to in the CMake configuration code, so we don't need to pass "magic" CMAKE_* flags in the CI configuration on macOS anymore.


bellenot (approved):
I trust you, if it works...

PR #20468, opened 2025-11-20
Directly include OpenGL in gl2ps

guitargeek:
The OpenGL ROOT wrapper headers like TGLIncludes are deprecated, and not using them internally in ROOT makes the migration from Glew to Glad easier.


bellenot (approved):
LGTM

PR #20469, opened 2025-11-20
[RDF] Bring back a removed snapshot method

hageboeck:
As discussed on the ROOT Users workshop, entirely removing interfaces after one single ROOT release leaves users with broken code without any hints on how to solve the problem. Instead, a minimal placeholder should remain, which gives directions on how solve the error. In that spirit, the templated snapshot overload is brought back and marked deprecated.

This partly reverts commit 4da10a206fedc7c6b482ead3acd92f481c7fa35c.

### DIFF
--- a/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
@@ -1242,6 +1242,15 @@ public:
       return newInterface;
    }
 
+   template <typename... ColumnTypes>
+   [[deprecated("Snapshot is not any more a template. You can safely remove the template parameters.")]]
+   RResultPtr<RInterface<RLoopManager>>
+   Snapshot(std::string_view treename, std::string_view filename, const ColumnNames_t &columnList,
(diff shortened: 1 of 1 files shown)


tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -1242,6 +1242,15 @@ public:
       return newInterface;
    }
 
+   template <typename... ColumnTypes>
+   [[deprecated("Snapshot is not any more a template. It compiles faster when removing all template parameters.")]]
```
vepadulano: I think the previous wording that included "you can safely remove them from this function call." is much clearer and gives the user confidence about removing the template parameters.

hageboeck:
```suggestion
   [[deprecated("Snapshot is not any more a template. You can safely remove the template parameters.")]]
```

hageboeck: OK, added it now.

PR #20471, opened 2025-11-21
[tmva][sofie] Run Keras tests and tutorials for any Keras version

guitargeek:
This is to demo the failure mode.


guitargeek:
This PR has served its purpose in demonstrating that there are failures.

PR #20479, opened 2025-11-22
[hist] Implement calling TSpectrum2 and TSpectrum3 Background from TH2/3

ferdymercury:
# This Pull request:

## Changes or fixes:

Since it's implemented for the 1D version, and for the 2D/3D versions you can call it via raw pointers but not directly via TH2 or TH3.

Functions were marked before as to-do, so no one could be calling them directly since they were already doing nothing and raising error. Hence, function signature and name were adapted to match the 2D and 3D needed parameters. Since these are legacy classes, the code from the implemented 1D version was copy-pasted and adapted, but no code modernization was performed, such as avoiding new/delete etc or clang-formatting.

Fixes https://root-forum.cern.ch/t/tspectrum2-bac-error-function-not-yet-implemented/64427
I've been hit myself with this, too.

## Checklist:

- [x] tested changes locally


dpiparo:
Thanks for contributing to reduce the `Small Bug Density`

PR #20496, opened 2025-11-24
[CPyCppyy] Correctly handle static method calls via derived instance

### LINKED ISSUE #20463: pyROOT TH1::GetDefaultSumw2 throws error when called
### Check duplicate issues.

- [x] Checked for duplicates

### Description

When calling class method `TH1::GetDefaultSumw2()`, python3 throws the following error:
```
TypeError: static bool TH1::GetDefaultSumw2() =>
    TypeError: takes at most 0 arguments (1 given)
```
The method takes 0 arguments in the documentation. I assume the "1 given" here is the instance itself passed to the method of the class, but this doesn't happen with other methods of `TH1` that I have noticed.

### Reproducer

Fails when running in Python 3.9.14 (should be similar in any Python 3).
```python
import ROOT
hist = ROOT.TH1F("h", "histogram", 100, -3, 3)
print(hist.GetDefaultSumw2()) #throws error here
```

Works find in C++ (ran with `root temp.cc`)
```C++
//in temp.cc
#include <iostream>                                                                      
#include "TH1.h"                                                                         
                                                                                         
void temp(){                                                                             
  auto *hist = new TH1F("h", "histogram", 100, -3, 3);                                   
  std::cout << hist->GetDefaultSumw2() << std::endl;                                     
}
```

### ROOT version

ROOT Version: 6.36.04
Built for macosx64 on Aug 25 2025, 09:02:18
From tags/6-36-04@6-36-04

### Installation method

homebrew package manager

### Operating system

MacOS 15.5

### Additional context

Bug also found in older version (6.32.13) found in new CMSSW virtual environments (e.g. CMSSW_15_0_14)

guitargeek:
The mechanism to strip away the "self" argument from the call to a static method checks the type of "self", matching the exact type.

However, it should also remove the "self" if it is an instance of a derived type, because in C++ it is also possible to call static methods via instances of derived types.

A unit test was also added.

Closes #20463.

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/CPPClassMethod.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/CPPClassMethod.cxx
@@ -34,8 +34,8 @@ PyObject* CPyCppyy::CPPClassMethod::Call(CPPInstance*&
     int nargs = (int)CPyCppyy_PyArgs_GET_SIZE(args, nargsf);
     if ((!self || (PyObject*)self == Py_None) && nargs) {
         PyObject* arg0 = CPyCppyy_PyArgs_GET_ITEM(args, 0);
-        if ((CPPInstance_Check(arg0) && ((CPPInstance*)arg0)->ObjectIsA() == GetScope()) && \
-                (fArgsRequired <= nargs-1)) {
+        if (CPPInstance_Check(arg0) && fArgsRequired <= nargs - 1 &&
+            Cppyy::IsSubtype(reinterpret_cast<CPPInstance *>(arg0)->ObjectIsA(), GetScope())) {
             args   += 1;     // drops first argument
             nargsf -= 1;
         }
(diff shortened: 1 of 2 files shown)


guitargeek:
Tests passed (just fixed typo in the commit message with recent push)

wlav:
I'd recommend leaving `((CPPInstance*)arg0)->ObjectIsA() == GetScope())` in place (with an `or` for the subtype check). Will be faster in the most common cases.

guitargeek:
Good point, I was thinking that too before. But the equality shortcut check is already done in the implementation of `Cppyy::IsSubtype`:
https://github.com/root-project/root/blob/master/bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx#L1615

And I don't think the overhead of the function call matters here, or does it?

wlav:
No, you're quite right, that's good enough. It's also a better place for the short-cut.

PR #20500, opened 2025-11-24
[RF][HF] Move all classes for Measurement specification to Measurement.h

guitargeek:
The Measurement uses several other helper classes, which are scattered in different header and implementation files.

Merging these classes to one header and compilation unit has several advantages:

  * User only needs to include one header

  * Easier to get an overview on the structure of the Measurement class, which helps for refactoring and re-designing of HistFactor using the new ROOT histogram classes

  * Shorter compilation times

There is no breaking of backwards compatibility, as the old header files are all kept around, just with no content. However, the headers are clearly deprecated with a warning that they are going to be removed in ROOT 7.


lmoneta (approved):
LGTM!
Thanks Jonas for the improvement

Issue #14446, opened 2024-01-26
gtest-math-matrix-test-testMatrixTSparse fails on aarch64, ppc64le and s390x with gcc 14


ellert:
### Check duplicate issues.

- [X] Checked for duplicates

### Description

Fedora 40 updated gcc to version 14. After this. the test gtest-math-matrix-test-testMatrixTSparse started failing on aarch64, ppc64le and s390x. The test still succeeds on ix86 and x86_64.

The test failure:

```
          Start  207: gtest-math-matrix-test-testMatrixTSparse
 243/1265 Test  #207: gtest-math-matrix-test-testMatrixTSparse ..................................***Failed    0.03 sec
Running main() from /builddir/build/BUILD/googletest-1.14.0/googletest/src/gtest_main.cc
Note: Google Test filter = -RCsvDS.Remote:RRawFile.Remote:RSqliteDS.Davix:TChainParsing.RemoteGlob:TFile.ReadWithoutGlobalRegistrationNet:TFile.ReadWithoutGlobalRegistrationWeb:RNTuple.TClassEBO
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from testSparse
[ RUN      ] testSparse.LwbInit
/builddir/build/BUILD/root-6.30.02/math/matrix/test/testMatrixTSparse.cxx:47: Failure
Expected equality of these values:
  m1
    Which is: 80-byte object <00-00 02-AA 28-E9 07-70 00-00 00-00 02-00 00-00 00-00 00-04 00-00 00-05 00-00 00-01 00-00 00-00 00-00 00-14 00-00 00-05 3C-B0 00-00 00-00 00-00 01-00 03-FF 87-BA DF-98 00-00 02-AA 29-1E 84-50 00-00 02-AA 29-1E 65-B0 00-00 02-AA 29-1E 7F-00>
  m2
[...]

PR #20501, opened 2025-11-25
[ci] Use AlmaLinux 10 for ARM build

guitargeek:
This is to test the newer GCC 14 on ARM.

Should also add coverage for #14446.

Two commits with fixes to h2root and RVec tests are also added to avoid these test failures on alma10 ARM without any further changes:
```txt
TEST FAILURES:
317:gtest-math-vecops-vecops-rvec
1998:roottest-root-hist-h2root
```

### DIFF
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -403,7 +403,7 @@ jobs:
             is_special: true
             property: march_native
             overrides: ["CMAKE_BUILD_TYPE=RelWithDebInfo", "CMAKE_CXX_FLAGS=-march=native", "CMAKE_C_FLAGS=-march=native", "builtin_zlib=ON", "builtin_zstd=ON"]
-          - image: alma9
+          - image: alma10
             is_special: true
             property: arm64
             overrides: ["CMAKE_BUILD_TYPE=RelWithDebInfo", "builtin_zlib=ON", "builtin_zstd=ON"]


.github/workflows/root-ci.yml
```diff
@@ -403,7 +403,7 @@ jobs:
             is_special: true
             property: march_native
             overrides: ["CMAKE_BUILD_TYPE=RelWithDebInfo", "CMAKE_CXX_FLAGS=-march=native", "CMAKE_C_FLAGS=-march=native", "builtin_zlib=ON", "builtin_zstd=ON"]
-          - image: alma9
```
pcanal, line 406: What do the experiment use for arm64?  i.e. shouldn't we still test both alma9 and alma10.

guitargeek: I don't know, but I would prefer to not blow up the CI matrix too much for no reason. If you want, we can also keep Alma 9. In the end, the motivation for this PR was to demo that aarch64 works with gcc 14, which is shipped with Alma 10. If I can show that, #14446 can be closed.

In any case, if the fixes to `h2root` and the RVec tests successfully fix the problems with Alma 10 ARM, I will open independent PRs for these fixes.

pcanal: @dpiparo What coverage do we need for arm64/alma?

dpiparo: I think CMS uses alma8, while ATLAS alma9 - I am not sure about LHCb and ALICE. Alma10 is the release for Run4 though...

PR #20510, opened 2025-11-25
[RF] avoid unknown pragma warning when clad is disabled

ferdymercury:
```
/home/user/builds/build-root_src-Desktop-Debug/include/RooFit/Detail/MathFuncs.h:181:17: warning: expected 'ON' or 'OFF' or 'DEFAULT' in pragma [-Wunknown-pragmas]
   #pragma clad checkpoint loop
```

Question however:
Is ROOFIT_CLAD defined when this warning is happening? IE when

` Generating G__RooFitCore.cxx, ../../lib/RooFitCore.pcm`


guitargeek:
Hi @ferdymercury, thank you very much for this! The only problem was that `ROOFIT_CLAD` is only defined at build time of RooFit itself. I have changed it to a new macro, `R__HAS_CLAD`, that is defined whenever ROOT was built with Clad support (in the style of other macros like `R__HAS_COCOA`, etc).

guitargeek:
Fedora Rawhide failures are related to system upgrade and incremental build.

roofit/roofitcore/inc/RooFit/Detail/MathFuncs.h
```diff
@@ -367,9 +369,9 @@ double flexibleInterp(unsigned int code, ParamsArray params, unsigned int n, Dou
                       double boundary, double nominal, int doCutoff)
 {
    double total = nominal;
-   #ifdef __CLING__
-   #pragma clad checkpoint loop
-   #endif
+#if defined(__CLING__) && defined(R__HAS_CLAD)
```
vgvassilev, line 372: FYI, we define `__CLAD__` which probably is what we could use here.

PR #20517, opened 2025-11-25
[RDF] Add Skewness and Kurtosis actions to RDataFrame using Welford's algorithm

edish-github:
# This Pull Request:

Adds `.Skewness()` and `.Kurtosis()` actions to RDataFrame.

### Motivation
In high-energy physics, looking beyond the mean and standard deviation is usually critical.
* **Skewness** helps in detecting asymmetry in energy distributions (which potentially indicates parity violation).
* **Kurtosis** helps identify "heavy tails" where rare events (like new heavy particles or dark matter candidates) might hide.

But currently it requires users to either bin data into `TH1` (losing precision) or write manual loops. This action allows for exact calculation in a single pass.

### Implementation Details
I used **Welford's Online Algorithm**, which allows us to calculate mean, variance, skewness, and kurtosis simultaneously in one pass. This is numerically stable and fits the RDataFrame parallel map-reduce pattern.

* **Implementation Note:** I initially explored an explicit SIMD implementation using `ROOT::RVec`.However, benchmarks on my local machine (M3, Clang -O3) showed that the compiler auto-vectorizes the scalar Welford loop very effectively, so i went for the scalar implementation

## Changes or fixes:

- Added `SkewnessHelper` and `KurtosisHelper` in `tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx`.
- Exposed `.Skewness()` and `.Kurtosis()` in `tree/dataframe/inc/ROOT/RDF/RInterface.hxx`.
- Registered the new action tags in `tree/dataframe/inc/ROOT/RDF/InterfaceUtils.hxx`.

## Checklist:

- [x] tested changes locally (Verified against a reference Welford implementation via test macro)
- [x] updated the docs (Added Doxygen comments matching the StdDev style)

### DIFF
--- a/tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
@@ -44,8 +44,7 @@
 #include <utility> // std::index_sequence
 #include <vector>
 #include <numeric> // std::accumulate in MeanHelper
-#include <cmath>
-
+#include <cmath> // std::sqrt, std::pow
 class TCollection;
 class TStatistic;
 class TTreeReader;
@@ -1318,6 +1317,7 @@ public:
 };
 
 // Implements Welford's Online Algorithm for Skewness
+// Ref: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm
 class R__CLING_PTRCHECK(off) SkewnessHelper : public RActionImpl<SkewnessHelper> {
    unsigned int fNSlots;
    std::shared_ptr<double> fResultSkewness;
@@ -1339,8 +1339,12 @@ public:
    {
    }
 
-   SkewnessHelper(SkewnessHelper &&) = default;
-   SkewnessHelper(const SkewnessHelper &) = delete;
+   // Rule of Five:
+   ~SkewnessHelper() = default;                                   // Destructor
+   SkewnessHelper(const SkewnessHelper &) = delete;               // Copy Ctor
+   SkewnessHelper &operator=(const SkewnessHelper &) = delete;    // Copy Assignment
+   SkewnessHelper(SkewnessHelper &&) = default;                   // Move Ctor
+   SkewnessHelper &operator=(SkewnessHelper &&) = default;        // Move Assignment
 
    void InitTask(TTreeReader *, unsigned int) {}
 
@@ -1404,10 +1408,14 @@ public:
    }
 
    // Helper functions for RMergeableValue
-   std::unique_ptr<RMergeableValueBase> GetMergeableValue() const final { return nullptr; }
+   std::unique_ptr<RMergeableValueBase> GetMergeableValue() const final {
+      // TODO: Implement mergeable value to support distributed RDataFrame
+      return nullptr;
+   }
 };
 
 // Implements Welford's Online Algorithm for Kurtosis
+// Ref: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm
 class R__CLING_PTRCHECK(off) KurtosisHelper : public RActionImpl<KurtosisHelper> {
    unsigned int fNSlots;
    std::shared_ptr<double> fResultKurtosis;
@@ -1431,8 +1439,12 @@ public:
    {
    }
 
-   KurtosisHelper(KurtosisHelper &&) = default;
-   KurtosisHelper(const KurtosisHelper &) = delete;
+   // Rule of Five:
+   ~KurtosisHelper() = default;                                   // 1. Destructor
+   KurtosisHelper(const KurtosisHelper &) = delete;               // 2. Copy Constructor
+   KurtosisHelper &operator=(const KurtosisHelper &) = delete;    // 3. Copy Assignment
+   KurtosisHelper(KurtosisHelper &&) = default;                   // 4. Move Constructor
+   KurtosisHelper &operator=(KurtosisHelper &&) = default;        // 5. Move Assignment
 
    void InitTask(TTreeReader *, unsigned int) {}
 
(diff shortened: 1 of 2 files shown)


hahnjo:
Thanks for the pointers to the numerically stable algorithm, I think I will borrow this for the global statistics of ROOT's new histograms :wink:

edish-github:
@hahnjo Thanks, It’s really cool that the algorithm might be useful for the new histograms too. 🚀

I’m currently working to implement the weighted version (West's algorithm) to support weighted skewness/kurtosis in RDF. The parallel merge steps get complex for cubic terms, but I verified the prototype against calculations, it holds up perfectly.

If you end up needing the weighted merge derivations (adapted from West/Chan) for the histogram work, please let me know—happy to share my prototype if it saves you some digging! 😊

dpiparo:
Hello. A bit of context around this PR might be useful. Did you propose these changes to address an issue related to a task at hand or about which you learned? In other words, is this done to fix someone's problems?

edish-github:
@dpiparo  Hello, thanks for looking into this.

Motivation for this was primarily about the feature completeness. Since Mean and StdDev are already supported, adding Skewness and Kurtosis seemed like a logical step to bring RDataFrame closer to parity with other data analysis tools. I used Welford's algorithm to make sure the calculations are numerically stable, and I'm currently working on the statistical analysis for weighted distributions for these metrics as well.

I was also happy to see Jonas mention above that this approach might be helpful for the new histograms, so hopefully, this proves to be a useful addition to the codebase.

Best regards,
Edish.

hageboeck:
Hello @edish-github,

I could imagine extending RDF, but before that, I have a little request 😅:
Could we separate whitespace changes and code changes? This tremendously helps the review.

We typically use `git clang-format` to format only the lines that have been touched. Like this, we progressively format the code base, but we don't introduce noise in "older" lines that were unaffected.

edish-github:
Hello @hageboeck, thanks for reviewing!

I've cleaned up the diff by reverting the unrelated whitespace changes and applying clang-format strictly to the new code. I also added the example usage snippets to the documentation to match the style of the other actions.

On a side note, I am really enjoying working with the RDF internals. Currently I'm working on tackling **weighted skewness/kurtosis** next (e.g., `.Skewness(col, weight_col)`), as weighted events are standard in MC analysis. 
Would that be a welcome addition, or is there a higher priority area in RDF you would suggest?

Best regards,
Edish

edish-github:
Hello @hageboeck (and @vepadulano @martamaja10),

I've updated the PR  and ran the tests locally.
Could you please trigger the CI workflows when you have a moment? I'd like to ensure the tests pass on the official runners.

Best,
Edish

edish-github:
Hello @hageboeck  (@vepadulano and @martamaja10)

It's been a little while, and I completely respect your time and busy schedules. I just wanted to gently bump this PR to see if we can get the workflows approved or if you have any initial feedback on the implementation.

I am standing by for any feedback or changes required. Thanks!

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -44,6 +44,7 @@
 #include <utility> // std::index_sequence
 #include <vector>
 #include <numeric> // std::accumulate in MeanHelper
+#include <cmath>
```
vepadulano: Just to know that we need this header, please add a comment to the right signaling which function is used from this header

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1316,6 +1317,195 @@ public:
    }
 };
 
+// Implements Welford's Online Algorithm for Skewness
```
vepadulano, line 1319: Just for reference, I would add a link to the explanation of the algorithm, e.g. https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1316,6 +1317,195 @@ public:
...
+        fResultSkewness(resPtr),
+        fCounts(fNSlots, 0),
+        fMeans(fNSlots, 0.),
+        fM2(fNSlots, 0.),
+        fM3(fNSlots, 0.)
+   {
+   }
+
+   SkewnessHelper(SkewnessHelper &&) = default;
+   SkewnessHelper(const SkewnessHelper &) = delete;
```
vepadulano: Should apply rule of five here (missing assignment operators and destructor)

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -1316,6 +1317,195 @@ public:
...
+   std::string GetActionName() { return "Skewness"; }
+
+   SkewnessHelper MakeNew(void *newResult, std::string_view /*variation*/ = "nominal")
+   {
+      auto &result = *static_cast<std::shared_ptr<double> *>(newResult);
+      return SkewnessHelper(result, fCounts.size());
+   }
+
+   // Helper functions for RMergeableValue
+   std::unique_ptr<RMergeableValueBase> GetMergeableValue() const final { return nullptr; }
```
vepadulano: Please add a TODO here, without the mergeable value implementation we can't support these methods in distributed RDataFrame

edish-github:
Hi @vepadulano, really appreciate your detailed feedback! I have updated the PR to address the comments:

Documentation: Added the explanation for <cmath> and the Wikipedia references for Welford's algorithm.
Rule of Five: Explicitly handled the destructor, copy, and move constructors/assignments for both SkewnessHelper and KurtosisHelper.
Distributed Support: Added the GetMergeableValue implementation returning nullptr with the requested TODO.
Testing: Added a regression test in dataframe_simple.cxx covering both Skewness and Kurtosis.
Ready for another look when you have a moment!

Issue #20472, opened 2025-11-21
jsroot from pip install root not working in notebooks (no TWebCanvas?)


will-cern:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

Just reporting here that the pip install version of root in a jupyter notebook (e.g. google colab) the default drawing fails (just create any drawable object and try to draw the canvas):

```
    309     if TWebCanvasAvailable():
--> 310         return ROOT.TWebCanvas.CreateCanvasJSON(canvas, 23, True)
    311 
    312     # Add extra primitives to canvas with custom colors, palette, gStyle

AttributeError: <class cppyy.gbl.TWebCanvas at 0x3d259af0> has no attribute 'CreateCanvasJSON'. Full details:
  type object 'TWebCanvas' has no attribute 'CreateCanvasJSON'
  'TWebCanvas::CreateCanvasJSON' is not a known C++ class
  'CreateCanvasJSON' is not a known C++ template
  'CreateCanvasJSON' is not a known C++ enum
```

I note that the following is True:

```
hasattr(ROOT, "TWebCanvas")
```

But `ROOT.TWebCanvas` is just a cppyy shell object. Has the pythonization here incorrectly determined that class as being part of the build?

Anyway ... temporary workaround is to disable jsroot `%jsroot off`)

### Reproducer

```
!pip install root
import ROOT
c = ROOT.TCanvas()
c.Draw()
```
in a notebook

### ROOT version

6.37.01

### Installation method

pip install

### Operating system

linux (google colab)

ferdymercury:
Related PR https://github.com/root-project/root/pull/20461

linev:
After PR #20461 jsroot magic will be disabled via rootrc file.
But I agree - one should check that TWebCanvas is not dummy class.
I guess, with `hasattr(ROOT.TWebCanvas, "CreateCanvasJSON")`

linev:
Maybe https://github.com/root-project/root/pull/20474

linev:
Very soon 6.38 release will be published in the pip.

In the release `%jsroot on/off` magic will be disabled by default and methods of `TWebCanvas` will be checked properly - thanks to #20474.

Keep issue open to get confirmation from users once ROOT release is there.

ferdymercury:
Hi Will, I checked as of now (on Ubu22) and both options (jsroot on and jsroot off) seem to be working fine. Please reopen if you re-encounter the issue! Thanks for the report!

[...]

PR #20520, opened 2025-11-25
[build] Signal explicitly to users the PyPI distribution is in alpha

vepadulano:
By injecting a warning in the __init__.py file of the package before starting the build of the wheel.

This change is added to make the situation of the Python wheel distribution as clear as possible and to manage user expectations, in line with the discussions held during the ROOT Users Workshop 2025.

See some examples of questions and feedback already existing which could have benefited from a more clear description: https://github.com/root-project/root/issues/20472
https://root-forum.cern.ch/t/is-this-a-legitimate-release-of-root/64429

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,9 +33,34 @@ INSTALL_DIR = tempfile.mkdtemp()
 ROOT_BUILD_INTERNAL_DIRNAME = "mock_site_packages"
 
 
+def _patch_root_init():
+    warning_patch = '''
+import warnings
+import textwrap
+
+warnings.warn(
+textwrap.dedent("""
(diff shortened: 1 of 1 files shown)


vepadulano:
An example of the warning when installing the wheel (e.g. on lxplus)
```
(myenv) [vpadulan@lxplus9101 /tmp/vpadulan]$PYTHONPATH=/tmp/vpadulan/myenv/lib/python3.13/site-packages/:$PYTHONPATH python
Python 3.13.9 (main, Nov  6 2025, 00:07:41) [GCC 13.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ROOT
/tmp/vpadulan/myenv/lib/python3.13/site-packages/ROOT/__init__.py:5: UserWarning: 
This distribution of ROOT is in alpha stage. Feedback is welcome and appreciated. Feel free to reach out to the user forum for questions and general feedback at https://root-forum.cern.ch or to submit an issue at https://github.com/root-project/root/issues. Do not rely on this distribution for production purposes.
```

PR #20531, opened 2025-11-26
[io] TDirectoryFile: propagate returnExistingDirectory to sub-calls and properly apply title

silverweed:
Currently this doesn't work as expected:
```c++
auto dir1 = file->mkdir("a/b");
auto dir2 = file->mkdir("a/b", "", /* returnExisting = */ true); // should return the already-existing dir
EXPECT_EQ(dir1, dir2); // FAILS: dir2 is nullptr
```

because TDirectoryFile::mkdir doesn't propagate the value of `returnExistingDirectory` to recursive mkdir calls. This PR fixes it.

In addition, `title` was currently only applied to the *topmost* directory in case of hierarchies, which is counterintuitive as the returned TDirectory is actually the innermost. This PR changes the (previously undocumented) title logic to the opposite: every non-leaf directory gets created with a default title and only the innermost gets the user-defined one. In our codebase we have 0 cases of `mkdir` being called with a hierarchy and an explicit title, so this change shouldn't impact anything.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/io/io/src/TDirectoryFile.cxx
+++ b/io/io/src/TDirectoryFile.cxx
@@ -1258,7 +1258,9 @@ TFile *TDirectoryFile::OpenFile(const char *name, Option_t *option,const char *f
 /// Create a sub-directory "a" or a hierarchy of sub-directories "a/b/c/...".
 ///
 /// @param name the name or hierarchy of the subdirectory ("a" or "a/b/c")
-/// @param title the title
+/// @param title the title of the directory. For hierarchies, this is only applied
+/// to the innermost directory (so if `name == "a/b/c"` and `title == "my dir"`,
+/// only `c` will have the title `"my dir"`).
 /// @param returnExistingDirectory if key-name is already existing, the returned
 /// value points to preexisting sub-directory if true and to `nullptr` if false.
 /// @return a pointer to the created sub-directory, not to the top sub-directory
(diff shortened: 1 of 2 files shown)


io/io/src/TDirectoryFile.cxx
```diff
@@ -1284,10 +1286,11 @@ TDirectory *TDirectoryFile::mkdir(const char *name, const char *title, Bool_t re
       TDirectoryFile *tmpdir = nullptr;
       GetObject(workname.Data(), tmpdir);
       if (!tmpdir) {
-         tmpdir = (TDirectoryFile*)mkdir(workname.Data(),title);
+         // We give all intermediate directories a default title, as `title` is only given to the innermost dir.
+         tmpdir = (TDirectoryFile *)mkdir(workname.Data(), workname.Data());
```
jblomer: Shouldn't we rather use the empty string for the title of intermediate directories?

silverweed: Perhaps; the only reason I used the name as the title is because that's what the function does in the normal case when an empty title is passed (see line https://github.com/root-project/root/pull/20531/files#diff-56ce80e060312dead2b5c7965a71f7ddee4211abf448bee96240b3c908833e7eR1275).
But an empty title makes more sense to me as well.

silverweed: (but given the line linked above, even if I pass `""` the effective title will be `workname.Data()`, so if we wanted actually empty titles we would have to change that)

pcanal (approved):
Thanks.

PR #20532, opened 2025-11-26
[v628][cmake] Build clad with one core only

guitargeek:
Hopefully fixing the nightly failures.


bellenot:
The problem was fixed differently in master. See https://github.com/root-project/root/blob/master/interpreter/cling/tools/plugins/clad/CMakeLists.txt#L32

guitargeek:
Yes, indeed it was fixed in master, but this problem appeared no also for older branches after updating CMake on the Ubuntu 22.04 config. I don't plan to investigate this further, as I just wanted to make minimal changes to we can keep 6.28 around in the CI.

bellenot:
OK, fine with me, I just wanted to know if applying the same solution would work, but I guess it's not important.

PR #20533, opened 2025-11-26
[RF] Add RooFit headers to private CMake filesets

ferdymercury:
To improve IDE experience.
See also https://github.com/root-project/root/pull/18419 and https://github.com/root-project/root/pull/20515

Non-functional change.

Work towards https://its.cern.ch/jira/browse/ROOT-10915


roofit/codegen/CMakeLists.txt
```diff
@@ -21,3 +21,16 @@ ROOT_STANDARD_LIBRARY_PACKAGE(RooFitCodegen
...
+if(NOT CMAKE_VERSION VERSION_LESS "3.23.0")
+  target_sources(
+      RooFitCodegen
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/
+      FILES
+          # LinkDef.h # empty, not being used by ROOT_STL_PACKAGE call
+          inc/RooFit/CodegenImpl.h
```
pcanal, line 34: Is this really private? Since it is in 'inc', it should be meant to be public (if it was private, it would need to move).  Also see:
```
roofit/roofitcore/src/RooFit/CodegenContext.cxx
348:   gInterpreter->Declare("#include <RooFit/CodegenImpl.h>\n");
```
which requires the header file to be public and installed.

roofit/histfactory/CMakeLists.txt
```diff
@@ -109,4 +113,20 @@ install(FILES ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${prepareHistFactory_script}
...
+  target_sources(
+      HistFactory
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/
+      FILES
+          ${RELATIVE_INC_HEADERS}
+          inc/LinkDef.h # seemingly not being used by ROOT_STL_PACKAGE call
+          inc/HFMsgService.h # not part of list above
```
pcanal, line 112: Aren't those public?

roofit/hs3/CMakeLists.txt
```diff
@@ -29,4 +33,23 @@ ROOT_STANDARD_LIBRARY_PACKAGE(RooFitHS3
...
+list(TRANSFORM RELATIVE_INC_HEADERS PREPEND inc/)
+if(NOT CMAKE_VERSION VERSION_LESS "3.23.0")
+  target_sources(
+      RooFitHS3
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/ src/
+      FILES
+          ${RELATIVE_INC_HEADERS}
```
pcanal, line 50: Why do we need the `inc` headers here?

roofit/multiprocess/CMakeLists.txt
```diff
@@ -30,6 +30,31 @@ target_include_directories(RooFitMultiProcess
 
 target_link_libraries(RooFitMultiProcess PRIVATE nlohmann_json::nlohmann_json)
 
+if(NOT CMAKE_VERSION VERSION_LESS "3.23.0")
+  target_sources(
+      RooFitMultiProcess
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/ res/ src/
```
pcanal, line 39: Why is `inc` listed here?

roofit/roofit/CMakeLists.txt
```diff
@@ -166,4 +170,20 @@ if("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
...
+list(TRANSFORM RELATIVE_INC_HEADERS PREPEND inc/)
+if(NOT CMAKE_VERSION VERSION_LESS "3.23.0")
+  target_sources(
+      RooFit
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/ src/
+      FILES
+          ${RELATIVE_INC_HEADERS}
```
pcanal, line 185: Same question as above.

roofit/codegen/CMakeLists.txt
ferdymercury: All the explanation to these questions are in https://github.com/root-project/root/pull/20515

They are public but here we add them to a private file-set (that is not yet used for anything expect for IDE visualization). In a next step we make it public, remove target_include_directories and copy_to_build-dir steps.

roofit/multiprocess/CMakeLists.txt
ferdymercury: That's how CMake FileSets work (that was their design choice), one has to do:
```
BASE_DIRS
   myfolder1
   myfolder2
FILES
   myfolder1/myheader.h
   myfolder1/mysubfolder/myheader2.h
   myfolder2/myheader3.h
```

So `FILES` are not relative to `BASE_DIRS`.
Instead, `BASE_DIRS` is telling you from what point you start including in your CPP file.
In this case, you would include via:
```
#include "myheader.h"
#include "mysubfolder/myheader2.h"
#include "myheader3.h"
```

roofit/hs3/CMakeLists.txt
ferdymercury: It's so that they appear inside the IDE. It's not needed for the build system, that's why everything is private.

In a future step (PR), we can split the public from the private ones and change the mechanism in the ROOT macros.

roofit/histfactory/CMakeLists.txt
ferdymercury: It's just for the IDE. Future step is to make them public and remove `target_include_directories` statement as well as the 'copy-header-to-build-dir' step.

The current step tries to create a short-term solution to the IDE without breaking anything. Next PR can take care of the functional changes that are more in risk of breaking things.

roofit/roofit/CMakeLists.txt
```diff
@@ -14,74 +14,77 @@ if (runtime_cxxmodules AND WIN32)
...
+  RooDecay.h
+  RooDstD0BG.h
+  RooExponential.h
+  RooFunctor1DBinding.h
+  RooFunctorBinding.h
+  RooGExpModel.h
+  RooGamma.h
+  RooGaussExpTails.h
+  RooGaussModel.h
+  RooHistConstraint.h
```
guitargeek, line 48:
```suggestion
  RooGaussModel.h
  RooGaussian.h
  RooHistConstraint.h
```
The hello world class of RooFit got lost 😆

roofit/roostats/CMakeLists.txt
```diff
@@ -134,4 +138,19 @@ ROOT_STANDARD_LIBRARY_PACKAGE(RooStats
...
+if(NOT CMAKE_VERSION VERSION_LESS "3.23.0")
+  target_sources(
+      RooStats
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/
+      FILES
+          ${RELATIVE_INC_HEADERS}
+          inc/LinkDef.h # seemingly this is not being used in ROOT_STL_PACKAGE call
```
guitargeek:
```suggestion
          inc/LinkDef.h # seemingly this is not being used in ROOT_STANDARD_LIBRARY_PACKAGE call
```
We should spell the macro name out correctly (also in the other places)

roofit/xroofit/CMakeLists.txt
```diff
@@ -34,4 +38,22 @@ ROOT_STANDARD_LIBRARY_PACKAGE(RooFitXRooFit
...
+      RooFitXRooFit
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/ src/
+      FILES
+          ${RELATIVE_INC_HEADERS}
+          inc/LinkDef.h
+          inc/RooFit/xRooFit/Config.h # was not part of XROOFIT_HEADERS
+          src/coutCapture.h
```
guitargeek, line 55:
```suggestion
          inc/RooFit/xRooFit/Config.h # was not part of XROOFIT_HEADERS
          src/PythonInterface.h
          src/coutCapture.h
```
Missing a new file that was added in `master` since your last rebase

guitargeek (changes requested):
Thank you very much! A few comments inline.

Also, to make this work, you have to drop `src/RooXYChi2Var.h` from `roofit/roofitcore/CMakeLists.txt`, because that class doesn't exist anymore.

roofit/codegen/CMakeLists.txt
```diff
@@ -21,3 +21,16 @@ ROOT_STANDARD_LIBRARY_PACKAGE(RooFitCodegen
...
+
+if(NOT CMAKE_VERSION VERSION_LESS "3.23.0")
+  target_sources(
+      RooFitCodegen
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/
+      FILES
+          # LinkDef.h # empty, not being used by ROOT_STL_PACKAGE call
```
ferdymercury:
```suggestion
          # LinkDef.h # empty, not being used by ROOT_STANDARD_LIBRARY_PACKAGE call
```

roofit/histfactory/CMakeLists.txt
```diff
@@ -93,4 +97,20 @@ install(FILES ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${prepareHistFactory_script}
...
+if(NOT CMAKE_VERSION VERSION_LESS "3.23.0")
+  target_sources(
+      HistFactory
+      PRIVATE
+      FILE_SET private_header_files
+      TYPE HEADERS
+      BASE_DIRS inc/
+      FILES
+          ${RELATIVE_INC_HEADERS}
+          inc/LinkDef.h # seemingly not being used by ROOT_STL_PACKAGE call
```
ferdymercury:
```suggestion
          inc/LinkDef.h # seemingly not being used by ROOT_STANDARD_LIBRARY_PACKAGE call
```

roofit/hs3/CMakeLists.txt
```diff
@@ -30,6 +34,25 @@ ROOT_STANDARD_LIBRARY_PACKAGE(RooFitHS3
...
+          # LinkDef.h # seemingly not being used by ROOT_STL_PACKAGE call
```
ferdymercury:
```suggestion
          # LinkDef.h # seemingly not being used by ROOT_STANDARD_LIBRARY_PACKAGE call
```

roofit/roofitcore/CMakeLists.txt
```diff
@@ -24,224 +31,236 @@ endif()
...
     src/RooRealMPFE.cxx
   )
+  set(LegacyEvalBackendHeaders
+    src/BidirMMapPipe.h
+    src/RooAbsOptTestStatistic.h
+    src/RooAbsTestStatistic.h
+    src/RooChi2Var.h
+    src/RooNLLVar.h
+    src/RooRealMPFE.h
+    src/RooXYChi2Var.h
```
ferdymercury:
```suggestion
```

guitargeek (approved):
LGTM, thanks for this!

In general, I still don't think we should do this for ROOT at a larger scale until the minimum CMake version is raised or we figure out how to do this the right way with CMake macros or functions to we don't duplicate code all over the place.

But for RooFit in particular, I really like this change, because it also helps to harmonize the code with the CMake code in the standalone RooFit repository!

So for now, I would leave it at the existing precedents and come back to this later.

ferdymercury:
Thanks!

> we figure out how to do this the right way with CMake macros or functions

In my opinion, I think ROOT should think about this from another perspective. Rather than creating custom macros that outside users / developers find hard to understand, and inside users get often confused (LIBRARIES vs DEPENDENCIES, who chose those two names xD), we should make use of standard CMake nomenclature (PUBLIC vs PRIVATE), no one would get confused here.

Here is my proposal for the future, with standard CMake > 3.23 nomenclature:

```
add_library(RooFitCore)
target_sources(
      RooFitCore
      PRIVATE
          src/BatchModeDataHelpers.cxx
          src/ConstraintHelpers.cxx
          ...
          ...
          src/RooImproperIntegrator1D.h # could also go into a PRIVATE FILE_SET if preferred
          res/RooFitImplHelpers.h
          res/RooUnitTest.h
      PUBLIC
        FILE_SET HEADERS
          TYPE HEADERS
          BASE_DIRS inc/ res/
          FILES
            inc/Roo1DTable.h
            inc/RooAICRegistry.h
            ...
      PUBLIC
        FILE_SET dictionaryHeaders
          TYPE HEADERS
          BASE_DIRS inc/ res/
          FILES
            inc/LinkDef.h
            ...
)
target_link_libraries(RooFitCore
  PUBLIC ${EXTRA_LIBRARIES}
  PRIVATE
    Core
    Hist
    Graf
    Matrix
    Tree
    RIO
    MathCore
    Foam
    RooBatchCompute
    ${EXTRA_DEPENDENCIES}
)
NEW_ROOT_STL_PACKAGE(RooFitCore   DICTIONARY_OPTIONS "-writeEmptyRootPCM")
```

This way, all the information is 'stored' in a standard CMake target. We do not need the annoying "Transform" or "Preprend" macro.

Instead, you would need to create a NEW_ROOT_STL_PACKAGE macro that traverses the FILE_SET and generates the dictionaries and so on. No need to pass HEADERS and SOURCES or LINKDEF to that macro. Just the name of the CMake target and the dictionary options. Everything is parsed internally, targets must just correctly define Public or Private file_sets and a special Linkdef header fileset.

The old ROOT_STL_PACKAGE can stay of course there for backward compatibility.
But this approach would be much cleaner and help towards many things in installation etc and Debian porting etc.

guitargeek:
I like that direction, that's exactly where I would like to go!

So then the next step would be to prototype this new `NEW_ROOT_STL_PACKAGE` macro I guess, before branching out to all of ROOT? RooFit can continue to serve as an area to try things out here.

ferdymercury:
> to prototype this new `NEW_ROOT_STL_PACKAGE` macro I guess

Yes, indeed.

Retrieving from the defined CMake targets the different file sets is straightforward:

```
  get_target_property(type ${tgt} TYPE)
  if(NOT ${type} STREQUAL "INTERFACE_LIBRARY")
    get_target_property(SrcList ${tgt} SOURCES)
    get_target_property(SrcDir ${tgt} SOURCE_DIR)
    list(TRANSFORM SrcList PREPEND ${SrcDir}/)
   # do what ever one wants with the list of sources
  endif()
  get_target_property(HeaderList ${tgt} HEADER_SET)
  if(NOT "HeaderList-NOTFOUND" IN_LIST HeaderList)
    # do whatever one wants with the list of headers
  endif()
```

In the new ROOT_STL macro, one would no longer need the calls to
`ROOT_OBJECT_LIBRARY` or `ROOT_LINKER_LIBRARY` since that would be already taken care by standard CMake one layer above via add_library SHARED or OBJECT.

Main step would then be:
```
ROOT_GENERATE_DICTIONARY(G__${libname} HeaderListFromFileSet
                          ${NO_CXXMODULE_FLAG}
                          ${STAGE1_FLAG}
                          MODULE ${tgt}
                          LINKDEF LinkdefListFromFileSet
                          NODEPHEADERS ${ARG_NODEPHEADERS}
                          OPTIONS ${ARG_DICTIONARY_OPTIONS}
                          DEPENDENCIES PublicDepsFromTgt or maybe not needed and just link to target?
                          BUILTINS Probably no longer needed
                          )
```

and then instead of calling
`ROOT_INSTALL_HEADERS(${ARG_INSTALL_OPTIONS})`

one would rely instead on the standard CMake install step specific for file_sets

PR #20534, opened 2025-11-26
[VecOps] Remove outdated `IsSmall` helper function in tests

guitargeek:
The commit https://github.com/guitargeek/root/commit/2605710bea7b169dde75bb2f89172a31df3ad157 changed from using the `ROOT::Detail::VecOps::` namespace explicitly for `IsSmall` to a `using` statement. However, there was also a remnant `IsSmall` function with an outdated implementation in the test file itself, which we can now remove.


guitargeek:
While this is certainly a good change, it didn't fix the GCC 14 ARM 64 failures unfortunately. So I have to push again, updating the commit message

PR #20535, opened 2025-11-26
[h2root] Adapt to GCC 8 passing convention for string length

guitargeek:
According to the `gfortran` argument passing conventions, for any Fortran procedure, the compiler will automatically define a C prototype. This is what we use in `h2root`.

Note that for procedures like `HROPEN` that takes string arguments, the signature of the C prototype will have extra arguments for the string lengths, which we also have to include in our forward declaration and usage. However, the type of these was changed with GCC 8 to size_t, so we have to also account for that as recommended in [1]. Otherwise, we get undefined behavor, which causes the `h2root` test to fail on ARM64 with GCC 14.

[1] https://gcc.gnu.org/onlinedocs/gfortran/Argument-passing-conventions.html

See also this PR that tests this commit on Alma 10 ARM, with GCC 14:

### DIFF
--- a/main/src/h2root.cxx
+++ b/main/src/h2root.cxx
@@ -131,6 +131,15 @@ void MAIN__() {}
 # define type_of_call
 # define DEFCHAR  const char*
 # define PASSCHAR(string) string
+
+// As recommended in
+// https://gcc.gnu.org/onlinedocs/gfortran/Argument-passing-conventions.html
+#if __GNUC__ > 7
+typedef size_t fortran_charlen_t;
+#else
+typedef int fortran_charlen_t;
+#endif
+
 #else
 # define hlimit  HLIMIT
 # define hropen  HROPEN
@@ -169,7 +178,7 @@ void MAIN__() {}
 extern "C" void  type_of_call hlimit(const int&);
 #ifndef WIN32
 extern "C" void  type_of_call hropen(const int&,DEFCHAR,DEFCHAR,DEFCHAR,
-                        const int&,const int&,const int,const int,const int);
+                        const int&,const int&,fortran_charlen_t,fortran_charlen_t,fortran_charlen_t);
 #else
 extern "C" void  type_of_call hropen(const int&,DEFCHAR,DEFCHAR,DEFCHAR,
                         const int&,const int&);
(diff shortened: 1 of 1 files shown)


ferdymercury:
Thanks a lot for this!
Could you remove the `-O0` flag in 
`set_target_properties(minicern PROPERTIES COMPILE_FLAGS "-O0 -w")`

to see if it's now good or if there are still issues somewhere?

As well as locally add:

```
set_target_properties(minicern PROPERTIES COMPILE_FLAGS "-fsanitize=undefined -fsanitize=address")
target_link_options(minicern BEFORE PUBLIC -fsanitize=undefined PUBLIC -fsanitize=address)
```

guitargeek:
Ok! I'll try in a separate PR after this one is merged, so we can see what happens

ferdymercury:
Just tried your changes locally on Ubuntu 22, with sanitizer enabled, and I still get:

```
/opt/root_src/misc/minicern/src/zebra.f:6746:72: runtime error: load of address 0x555556df372c with insufficient space for an object of type 'integer(kind=4)'
0x555556df372c: note: pointer points here
  24 02 00 00 18 2f 0f 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00
              ^ 
/opt/root_src/misc/minicern/src/zebra.f:6747:72: runtime error: load of address 0x555556df372c with insufficient space for an object of type 'integer(kind=4)'
0x555556df372c: note: pointer points here
  24 02 00 00 18 2f 0f 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00
              ^ 
/opt/root_src/misc/minicern/src/zebra.f:6753:72: runtime error: store to address 0x555556df372c with insufficient space for an object of type 'integer(kind=4)'
0x555556df372c: note: pointer points here
  24 02 00 00 18 2f 0f 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00
              ^ 
 Converting directory //example
```

but I guess that's an unrelated issue.

main/src/h2root.cxx
```diff
@@ -169,7 +178,7 @@ void MAIN__() {}
 extern "C" void  type_of_call hlimit(const int&);
 #ifndef WIN32
 extern "C" void  type_of_call hropen(const int&,DEFCHAR,DEFCHAR,DEFCHAR,
-                        const int&,const int&,const int,const int,const int);
+                        const int&,const int&,fortran_charlen_t,fortran_charlen_t,fortran_charlen_t);
```
ferdymercury, line 181: Here the "const" is being removed but I guess it does not matter

guitargeek: I just did what the gcc docs suggested :)

PR #20536, opened 2025-11-26
[main] Rewrite rootcp in C++

silverweed:
# This Pull request:
Depends on #20531.

Replaces `rootcp.py` with a native version in C++. 
Like for `rootls` and `rootbrowse`, this makes `rootcp` available even without python and, as a bonus, makes the tool significantly faster (the python version takes about 500ms just to start).

The behaviour should be the same as the python version and it has the benefit of not crashing when trying to copy RNTuples (although RNTuple is not handled yet).

The new `rootcp` has been currently mostly tested with our own roottest suite, but probably we want some more thorough testing before merging, especially if rootcp is used in experiments workflows.

## Other significant changes
- Some code was put in common between `rootls` and `rootcp` (will also be useful for `rootmv` and likely `rootmkdir`): this is the code that basically converts a path like `"file.root:a/b/*"` to a tree of objects and directories.

## Checklist:

- [x] tested changes locally


main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+   if (args.fPrintHelp != RootCpArgs::EPrintUsage::kNo) {
+      std::cerr << kShortHelp;
+      if (args.fPrintHelp == RootCpArgs::EPrintUsage::kLong) {
+         std::cerr << kLongHelp;
+         return 0;
+      }
+      return 1;
+   }
+
+   // Get destination. In general it may be a string like "prefix://file.root:path/to/dir", so check if it refers to
```
vepadulano, line 322: Asking because I don't know the answer: can the prefix be a remote protocol? i.e. is it valid to copy from an xroot-based file to a local file (and especially vice-versa)?

main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+            Err() << "destination path \"" << destFname << ":" << destPath << "\" already exists (as an object of type "
+                  << destDirKey->GetClassName() << "). Use the --replace flag to overwrite it.\n";
+            return 1;
+         } else {
+            destDirKey = nullptr;
+         }
+      }
+   }
+
+   // If we are copying multiple objects the destination path must either be empty or a TDirectory.
```
vepadulano, line 388: What happens if the destination path is empty? A new TDirectory is created and all the objects are copied there? Or just the last object is copied with the same key name?

main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+      Err() << "multiple sources were specified, but destination path \"" << destFname << ":" << destPath
+            << "\" is not a directory.\n";
+      return 1;
+   }
+
+   if (args.fCompression)
+      destFile->SetCompressionSettings(*args.fCompression);
+
+   const std::uint32_t flags = args.fRecursive * EGetMatchingPathsFlags::kRecursive;
+   // const bool oneFile = args.fSources.size() == 1;
```
vepadulano: Commented code

main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+   for (const auto &srcFname : args.fSources) {
+      auto src = ROOT::CmdLine::ParseRootSource(srcFname, flags);
+      if (!src.fErrors.empty()) {
+         for (const auto &err : src.fErrors)
+            Err() << err << "\n";
+         return 1;
+      }
+
+      // If we are copying multiple objects the destination path must either be empty or a TDirectory.
```
vepadulano, line 416: Can this comment be extended to indicate the difference between this scenario and the one handled with the same code at lines 367-370?

main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+      if (destIsNewObject && src.fObjectTree.fLeafList.size() + src.fObjectTree.fDirList.size() > 1) {
+         Err() << "multiple sources were specified but destination path \"" << destFname << ":" << destPath
+               << "\" is not a directory.\n";
+         return 1;
+      }
+
+      // According to rootcp.py, this is necessary for fast copy. TODO: verify
+      gROOT->GetListOfFiles()->Remove(src.fObjectTree.fFile.get());
```
vepadulano: Maybe this is "fixed" by avoiding the global registration of the files? @pcanal

main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+      return {{}, path};
+
+   auto dirName = path.substr(0, lastSlashIdx);
+   auto pathName = path.substr(lastSlashIdx + 1);
+   return {dirName, pathName};
+}
+
+static void CopyNode(const RootSource &src, const RootCpDestination &dest, NodeIdx_t nodeIdx, const RootCpArgs &args)
+{
+   TFile *srcfile = src.fObjectTree.fFile.get();
```
vepadulano, line 180: Just to double check, we are guaranteed to get a valid pointer here right? Maybe add a comment as to where this guarantee comes from?

main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+   const RootObjNode &node = src.fObjectTree.fNodes[nodeIdx];
```
vepadulano, line 185: Similarly here, add a comment as to where we guarantee that `nodeIdx` is never going to be out of range.

main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+   const std::string srcFullPath = NodeFullPath(src.fObjectTree, nodeIdx);
+   // Directory path, excluding trailing '/' and without the "file.root:" prefix.
+   const std::string_view srcDirPath =
+      (node.fParent == 0) ? std::string_view{}
+                          : std::string_view{srcFullPath.data(), srcFullPath.size() - node.fName.size() - 1};
```
vepadulano, line 190: I don't see where the `file.root:` prefix is removed here

main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+      destFullPath = std::string(destDirPath) + "/" + node.fName;
+      destBaseName = node.fName;
+   }
+
+   if (src.fFileName == dest.fFname && srcFullPath == destFullPath) {
+      Err() << src.fFileName << ":" << srcFullPath << ": source and destination cannot be the same\n";
+      return;
+   }
+
+   Info(2) << "cp " << src.fFileName << ":" << srcFullPath << " -> " << dest.fFname << ":" << destFullPath << "\n";
```
vepadulano, line 214: Just to double check, we are just printing `cp` here, wouldn't it be better to print `rootcp`?

vepadulano: Maybe we already print `rootcp` somewhere else in the logs

main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+                                  /* returnPreExisting = */ true);
+   }
+
+   // Check if the destination already exists. There are 3 cases here:
+   // 1. it doesn't: just go on as normal;
+   // 2. it does and it is a directory: the copied object needs to be copied inside it, but this was already accounted
+   //    for outside CopyNode, so just go on as normal;
+   // 3. it does and it's not a directory: if we have the replace flag, replace it, otherwise error out.
+   const TKey *destKey = destDir->GetKey(std::string(destBaseName).c_str());
+   if (destKey && !TClass::GetClass(destKey->GetClassName())->InheritsFrom("TDirectory") && !args.fReplace) {
```
vepadulano, line 229: Similar code also appears at line 355, would be good to explain why in a comment

main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+
+      if (TTree *old = dynamic_cast<TTree *>(obj)) {
+         // special case for TTree
+         TDirectory::TContext ctx(gDirectory, destDir);
+         obj = old->CloneTree(-1, "fast");
+         if (dest.fIsNewObject) {
+            static_cast<TTree *>(obj)->SetName(std::string(destBaseName).c_str());
+         }
+         obj->Write();
+         old->Delete();
```
vepadulano, line 283: Why is `old->Delete` necessary here? Maybe @pcanal knows better

main/src/rootcp.cxx
```diff
@@ -0,0 +1,419 @@
...
+      }
+      obj->Delete();
+   } else if (cl == TClass::GetClass("ROOT::RNTuple")) {
+      // TODO: RNTuple
+      Warn() << "object '" << node.fName << "' of type '" << node.fClassName
+             << "' will not be copied, as its type is currently unsupported by rootcp.\n";
+   } else {
+      Warn() << "object '" << node.fName << "' of type '" << node.fClassName
+             << "' will not be copied, as its type is currently unsupported by rootcp.\n";
+   }
```
vepadulano, line 304: So we only support writing `TObject`s with `rootcp` right? e.g. a `std::vector` wouldn't be copied over.

vepadulano (changes requested):
Great work, thanks! Some questions from my side.

main/src/rootcp.cxx
pcanal: Yes, avoid the global registration in the first place would have the same effect.  (This is likely to have been used to speed up the case of files with large number of histograms).

main/src/rootcp.cxx
pcanal: We should support the copying independently of the file locations (i.e. support remote protocol here).

main/src/rootcp.cxx
pcanal: This is equivalent to (because the argument is empty - it would be different with the "all" parameter):
```
delete obj;
```
And thus we should switch to that syntax.

Calling it is not strictly necessary but helps reduce the average memory use.

silverweed: Looking at the code for e.g. `TTree::Delete` it looks like it does a bit more than just `delete this`, e.g. it removes the object from the Cling symbol table. Not sure if relevant (`TCling::DeleteGlobal` seems to do nothing anyway), but is this guaranteed to be the case for any TObject?

main/src/rootcp.cxx
silverweed: Yes, currently we only support TObject (like `rootcp.py`). I would like to also add RNTuple support at some point, but probably no more than that.

main/src/rootcp.cxx
silverweed: An empty destination (i.e. `out.root` with no path) is equivalent to specifying `out.root:.`, so it will copy all objects to the top-level directory of the out file.
If any of the source objects is a directory, it will be copied if and only if you had the `-r` flag (similarly to what `cp` would do).

main/src/rootcp.cxx
silverweed: It is indeed a bit hidden: `NodeFullPath` takes as a third argument an option whether to include or exclude the `file.root:` prefix, and it is "exclude" by default. So it's not that the string_view operation is removing it: it's not part of `srcFullPath` to begin with. I will change the comment to clarify it.

Edit: Actually, even better, I'll make the third argument explicit.

main/src/rootcp.cxx
silverweed: We do: the complete log is: `Info in <rootcp>: cp in.root:foo -> out.root:foo`.
This style of log was made to mimic `cp -v`

main/src/rootcp.cxx
silverweed: I added a test to read from a remote file (I don't think we can easily test writing to a remote file from roottest, but if I'm wrong please let me know how to do it)

main/src/RootObjTree.cxx
```diff
@@ -0,0 +1,261 @@
...
+
+ROOT::CmdLine::RootSource
+ROOT::CmdLine::GetMatchingPathsInFile(std::string_view fileName, std::string_view pattern, std::uint32_t flags)
+{
+   ROOT::CmdLine::RootSource source;
+   source.fFileName = fileName;
+   auto &nodeTree = source.fObjectTree;
+   nodeTree.fFile =
+      std::unique_ptr<TFile>(TFile::Open(std::string(fileName).c_str(), "READ_WITHOUT_GLOBALREGISTRATION"));
+   if (!nodeTree.fFile) {
```
vepadulano:
```suggestion
   if (!nodeTree.fFile || nodeTree.fFile->IsZombie()) {
```

vepadulano (approved):
Nice work, all clear for me!

main/src/RootObjTree.cxx
```diff
@@ -0,0 +1,261 @@
...
+               patternWasMatchedAtLeastOnce[cur->fNesting] = true;
+            else
+               continue;
+         }
+
+         if (namesFound.count(key->GetName()) > 0) {
+            std::cerr << "WARNING: Several versions of '" << key->GetName() << "' are present in '" << fileName
+                      << "'. Only the most recent will be considered.\n";
+            continue;
+         }
```
pcanal, line 92: Why is this an error?  Having backup keys is the common case?

main/src/RootObjTree.cxx
```diff
@@ -0,0 +1,261 @@
...
+      assert(cur->fDir);
+
+      // Gather all keys under this directory and sort them by name.
+      std::vector<TKey *> keys;
+      keys.reserve(cur->fDir->GetListOfKeys()->GetEntries());
+      for (TKey *key : ROOT::Detail::TRangeStaticCast<TKey>(cur->fDir->GetListOfKeys()))
+         keys.push_back(key);
+
+      std::sort(keys.begin(), keys.end(),
+                [](const auto *a, const auto *b) { return strcmp(a->GetName(), b->GetName()) < 0; });
```
pcanal, line 70: Why is this intermediary step needed?  (And it seems it might also shuffle the ordering of the keys that corresponding to the same name).

main/src/RootObjTree.cxx
silverweed: It's just a warning, and it was like this in `rootcp.py` so I kept it as-is

main/src/RootObjTree.cxx
silverweed: Because we want to show the keys sorted by name? I'm not sure I understand the question

PR #20538, opened 2025-11-26
[CI] fast H2rootcheck

ferdymercury:
Fast testing platform running only h2root for verifying failure rate.

to verify on several remote platforms at once multiple times

Since h2root might fail only 1 out of every 5 times, it's interesting to have a fast tool to rerun 5 or 10 times on multi-platform but consuming less resources.

fyi @guitargeek

TODO review Windows is still OFF


main/src/h2root.cxx
ferdymercury, line 1: @guitargeek maybe this change in h2root.cxx (i mean the last commit with DEFCHAR) also makes sense? If yes feel free to do a separate PR on master, but I guess we need ifdef win32 else ... On top.

Anyway it does not solve the crash

PR #20524, opened 2025-11-26
Mutable polynomial roots

meiyasan:
# This Pull request:

Hi @lmoneta @vepadulano ; we may benefit from introducing a `mutable` keyword to `fRoots` in the Polynomial.h methods. I would be grateful if you could consider such changes as described below.

## Changes or fixes:
I applied a mutable keywords to `fRoots` allowing to use FindRoots in const context. Such keyword was already applied along the same line to `fDerived_params` variable in `math/mathmore/inc/Math/Polynomial.h`.

## Checklist:

- [x] tested changes locally
- [x] updated the docs (if necessary)

## Tests:

I ran the ROOT battery of tests with couple of issues (related to some stress test, vectorization & simultions), but those also happened on my Mac M3 ARM before my modifications. If you have any concern, I believe it wouldn't be an issue to run it in a more standard dev context.
```
cmake -DCMAKE_BUILD_TYPE=Debug -Dtesting=ON -Droottest=ON -Dmathmore=ON ../polynomial-patch
make -j8
ctest -j8
```

Regarding the Polynomial class modifications, my use case output remains unchanged. (I use polynomials to compute digital filtering, biquad, etc..) I also paid attention in particular to those related to polynomials: `gtest-hist-hist-testTProfile2Poly`, etc..
```
1015/3593 Test  #155: gtest-graf2d-primitivesv7-graf2dprimitivesv7testUnit ..............................................   Passed    0.39 sec
          Start  158: gtest-hist-hist-testTProfile2Poly
1016/3593 Test  #156: gtest-graf3d-eve7-graf3deve7testUnit ..............................................................   Passed    0.46 sec
          Start  159: gtest-hist-hist-testTFractionFitter
1017/3593 Test  #159: gtest-hist-hist-testTFractionFitter ...............................................................   Passed    0.39 sec
          Start  160: gtest-hist-hist-testTH2PolyBinError
1018/3593 Test  #160: gtest-hist-hist-testTH2PolyBinError ...............................................................   Passed    0.30 sec
          Start  161: gtest-hist-hist-testTH2PolyAdd
1019/3593 Test   #84: pyunittests-bindings-pyroot-pythonizations-pyroot-pyz-rdataframe-histo-profile ....................   Passed   13.81 sec
          Start  162: gtest-hist-hist-testTH2PolyGetNumberOfBins
1020/3593 Test  #162: gtest-hist-hist-testTH2PolyGetNumberOfBins ........................................................   Passed    0.24 sec
[...]

PR #20539, opened 2025-11-26
Const polynomial roots 

meiyasan:
# This Pull request:

Hi @hageboeck ; This is a followup on my previous PR https://github.com/root-project/root/pull/20524 aiming to improve polynomial implementation. I just implemented your suggestions and deprecating internal `fRoot` property from polynomial header and defined it in each of the `Find*Root` methods .

## Changes or fixes:

Removed `fProp`, added const property to public `Find*Root` methods.
Feel free to as necessary.

## Checklist:

- [x] tested changes locally

## Tests:

I ran the ROOT battery of tests with couple of issues.
```
cmake -DCMAKE_BUILD_TYPE=Debug -Dtesting=ON -Droottest=ON -Dmathmore=ON ../polynomial-patch
make -j8
ctest -j8
```

Regarding the Polynomial class modifications, my use case output remains unchanged. (I use polynomials to compute digital filtering, biquad, etc..) I also paid attention in particular to those related to polynomials: `gtest-hist-hist-testTProfile2Poly`, etc..

### DIFF
--- a/main/src/h2root.cxx
+++ b/main/src/h2root.cxx
@@ -131,6 +131,15 @@ void MAIN__() {}
 # define type_of_call
 # define DEFCHAR  const char*
 # define PASSCHAR(string) string
+
+// As recommended in
+// https://gcc.gnu.org/onlinedocs/gfortran/Argument-passing-conventions.html
+#if __GNUC__ > 7
+typedef size_t fortran_charlen_t;
+#else
+typedef int fortran_charlen_t;
+#endif
+
 #else
 # define hlimit  HLIMIT
 # define hropen  HROPEN
@@ -169,7 +178,7 @@ void MAIN__() {}
 extern "C" void  type_of_call hlimit(const int&);
 #ifndef WIN32
 extern "C" void  type_of_call hropen(const int&,DEFCHAR,DEFCHAR,DEFCHAR,
-                        const int&,const int&,const int,const int,const int);
+                        const int&,const int&,fortran_charlen_t,fortran_charlen_t,fortran_charlen_t);
 #else
 extern "C" void  type_of_call hropen(const int&,DEFCHAR,DEFCHAR,DEFCHAR,
                         const int&,const int&);
@@ -179,7 +188,7 @@ extern "C" void  type_of_call hrin(const int&,const int&,const int&);
 extern "C" void  type_of_call hnoent(const int&,const int&);
 #ifndef WIN32
 extern "C" void  type_of_call hgive(const int&,DEFCHAR,const int&,const float&,const float&,
-   const int&,const float&,const float&,const int&,const int&,const int);
+   const int&,const float&,const float&,const int&,const int&,fortran_charlen_t);
 #else
 extern "C" void  type_of_call hgive(const int&,DEFCHAR,const int&,const float&,const float&,
    const int&,const float&,const float&,const int&,const int&);
@@ -187,20 +196,20 @@ extern "C" void  type_of_call hgive(const int&,DEFCHAR,const int&,const float&,c
 
 #ifndef WIN32
 extern "C" void  type_of_call hgiven(const int&,DEFCHAR,const int&,DEFCHAR,
-   const float&,const float&,const int,const int);
+   const float&,const float&,fortran_charlen_t,fortran_charlen_t);
 #else
 extern "C" void  type_of_call hgiven(const int&,DEFCHAR,const int&,DEFCHAR,
    const float&,const float&);
 #endif
 
 #ifndef WIN32
-extern "C" void  type_of_call hntvar2(const int&,const int&,DEFCHAR,DEFCHAR,DEFCHAR,int&,int&,int&,int&,int&,const int,const int, const int);
+extern "C" void  type_of_call hntvar2(const int&,const int&,DEFCHAR,DEFCHAR,DEFCHAR,int&,int&,int&,int&,int&,fortran_charlen_t,fortran_charlen_t,fortran_charlen_t);
 #else
 extern "C" void  type_of_call hntvar2(const int&,const int&,DEFCHAR,DEFCHAR,DEFCHAR,int&,int&,int&,int&,int&);
 #endif
 
 #ifndef WIN32
-extern "C" void  type_of_call hbnam(const int&,DEFCHAR,const int&,DEFCHAR,const int&,const int, const int);
+extern "C" void  type_of_call hbnam(const int&,DEFCHAR,const int&,DEFCHAR,const int&,fortran_charlen_t,fortran_charlen_t);
 #else
 extern "C" void  type_of_call hbnam(const int&,DEFCHAR,const int&,DEFCHAR,const int&);
 #endif
@@ -232,14 +241,14 @@ extern "C" double type_of_call hije(const int&,const int&,const int&);
 #endif
 
 #ifndef WIN32
-extern "C" void  type_of_call hcdir(DEFCHAR,DEFCHAR ,const int,const int);
+extern "C" void  type_of_call hcdir(DEFCHAR,DEFCHAR ,fortran_charlen_t,fortran_charlen_t);
 #else
 extern "C" void  type_of_call hcdir(DEFCHAR,DEFCHAR);
 #endif
 
 extern "C" void  type_of_call zitoh(const int&,const int&,const int&);
 #ifndef WIN32
-extern "C" void  type_of_call uhtoc(const int&,const int&,DEFCHAR,int&,const int);
+extern "C" void  type_of_call uhtoc(const int&,const int&,DEFCHAR,int&,fortran_charlen_t);
 #else
 extern "C" void  type_of_call uhtoc(const int&,const int&,DEFCHAR,int&);
 #endif
--- a/math/vecops/test/vecops_rvec.cxx
+++ b/math/vecops/test/vecops_rvec.cxx
@@ -1493,13 +1493,6 @@ TEST(VecOps, Construct)
    EXPECT_TRUE(fourVects[2] == ref2);
 }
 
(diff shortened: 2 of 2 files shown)


meiyasan:
sorry @hageboeck, I failed to put you as reviewer and cannot adjust.

math/mathmore/inc/Math/Polynomial.h
```diff
@@ -116,20 +116,20 @@ class Polynomial : public ParamFunction<IParamGradFunction>,
       equation is very small. In that case it might be more robust to use the numerical method, by calling directly FindNumRoots()
 
    */
-   const std::vector<std::complex <double> > & FindRoots();
+   const std::vector<std::complex <double> > FindRoots() const;
```
hageboeck: Now that we return by value, we don't need to impose constness on the user of this function, so we are actually making things easier to use! 🙂 
```suggestion
  std::vector<std::complex <double> > FindRoots() const;
```

math/mathmore/inc/Math/Polynomial.h
```diff
@@ -116,20 +116,20 @@ class Polynomial : public ParamFunction<IParamGradFunction>,
...
    */
-   std::vector<double > FindRealRoots();
+   std::vector<double > FindRealRoots() const;
 
    /**
       Find the polynomial roots using always an iterative numerical methods
       The numerical method used is from GSL (see <A HREF="https://www.gnu.org/software/gsl/doc/html/poly.html">documentation</A> )
    */
-   const std::vector<std::complex <double> > & FindNumRoots();
+   const std::vector<std::complex <double> > FindNumRoots() const;
```
hageboeck:
```suggestion
  std::vector<std::complex <double> > FindNumRoots() const;
```

math/mathmore/src/Polynomial.cxx
```diff
@@ -148,7 +148,7 @@ IGenFunction * Polynomial::Clone() const {
 }
 
 
-const std::vector< std::complex <double> > &  Polynomial::FindRoots(){
+const std::vector< std::complex <double> > Polynomial::FindRoots() const {
```
hageboeck:
```suggestion
std::vector< std::complex <double> > Polynomial::FindRoots() const {
```

math/mathmore/src/Polynomial.cxx
```diff
@@ -157,6 +157,7 @@ const std::vector< std::complex <double> > &  Polynomial::FindRoots(){
       n--;
     }
 
+    std::vector< std::complex < double > > fRoots;
```
hageboeck: I understand the intention, but maybe we should not use the `f` for "field" in this context, since it's not any more a field of the class.

math/mathmore/src/Polynomial.cxx
```diff
@@ -226,16 +227,17 @@ const std::vector< std::complex <double> > &  Polynomial::FindRoots(){
...
     return fRoots;
 
   }
 
 
-std::vector< double >  Polynomial::FindRealRoots(){
-  FindRoots();
+std::vector< double >  Polynomial::FindRealRoots() const {
+
+  std::vector< std::complex < double > > fRoots = FindRoots();
```
hageboeck: Same here

math/mathmore/src/Polynomial.cxx
```diff
@@ -244,14 +246,16 @@ std::vector< double >  Polynomial::FindRealRoots(){
   }
   return roots;
 }
-const std::vector< std::complex <double> > &  Polynomial::FindNumRoots(){
+const std::vector< std::complex <double> >  Polynomial::FindNumRoots() const {
```
hageboeck:
```suggestion
std::vector< std::complex <double> >  Polynomial::FindNumRoots() const {
```

hageboeck:
@meiyasan one more thing:
After committing (or staging in git), you can already run `git clang-format`, so your changes are auto-formatted. Otherwise, the formatting check in the CI will fail and you will have to do it anyway. 😄

meiyasan:
thanks! I just ran the formatter against the previous commit; removed the extra const you recommended & also renamed fRoots into `roots` (additionally, in GetRealRoots, I renamed `roots` into `real_roots` to avoid conflicts).
```
git clang-format b9d0ae5b4ee84d6d09fd9045c302b829c44edde8 -- math/mathmore/inc/Math/Polynomial.h     math/mathmore/src/Polynomial.cxx
```

hageboeck:
> thanks! I just ran the formatter against the previous commit; removed the extra const you recommended & also renamed fRoots into `roots` (additionally, in GetRealRoots, I renamed `roots` into `real_roots` to avoid conflicts).
> 

Hello, sounds good! There must have been a small error in git, though:
- I don't see `fRoots` being renamed.
- We don't allow merge commits in PRs to ROOT.  You will have to remove the two commits that merge master in your branch. That should be OK, because nobody will work on the same files. Once the merge commits are removed, you can apply the edits referring to `fRoots`, and then possibly `git clang-format HEAD~` or `git clang-format <root-project>/master`, where you insert the name of the remote that points to `root-project`.

math/mathmore/inc/Math/Polynomial.h
```diff
@@ -116,20 +116,20 @@ class Polynomial : public ParamFunction<IParamGradFunction>,
       equation is very small. In that case it might be more robust to use the numerical method, by calling directly FindNumRoots()
 
    */
-   const std::vector<std::complex <double> > & FindRoots();
+   const std::vector<std::complex<double>> FindRoots() const;
```
vepadulano, line 119:
```suggestion
  std::vector<std::complex<double>> FindRoots() const;
```

As per https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f49-dont-return-const-t

math/mathmore/inc/Math/Polynomial.h
```diff
@@ -116,20 +116,20 @@ class Polynomial : public ParamFunction<IParamGradFunction>,
...
    */
-   std::vector<double > FindRealRoots();
+   std::vector<double> FindRealRoots() const;
 
    /**
       Find the polynomial roots using always an iterative numerical methods
       The numerical method used is from GSL (see <A HREF="https://www.gnu.org/software/gsl/doc/html/poly.html">documentation</A> )
    */
-   const std::vector<std::complex <double> > & FindNumRoots();
+   const std::vector<std::complex<double>> FindNumRoots() const;
```
vepadulano, line 132:
```suggestion
  std::vector<std::complex<double>> FindNumRoots() const;
```

As per https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f49-dont-return-const-t

math/mathmore/src/Polynomial.cxx
```diff
@@ -226,32 +226,34 @@ const std::vector< std::complex <double> > &  Polynomial::FindRoots(){
...
-    while ( Parameters()[n] == 0 ) {
+   // check if order is correct
+   unsigned int n = fOrder;
+   while (Parameters()[n] == 0) {
       n--;
     }
+
+    std::vector<std::complex<double>> fRoots;
     fRoots.clear();
     fRoots.reserve(n);
```
vepadulano, line 258:
```suggestion
```

These are not needed now

math/mathmore/src/Polynomial.cxx
```diff
@@ -147,16 +147,16 @@ IGenFunction * Polynomial::Clone() const {
...
-    while ( Parameters()[n] == 0 ) {
+   // check if order is correct
+   unsigned int n = fOrder;
+   while (Parameters()[n] == 0) {
       n--;
     }
 
+    std::vector<std::complex<double>> fRoots;
     fRoots.clear();
     fRoots.reserve(n);
```
vepadulano, line 161:
```suggestion
```

These are not needed now

math/mathmore/src/Polynomial.cxx
```diff
@@ -226,32 +226,34 @@ const std::vector< std::complex <double> > &  Polynomial::FindRoots(){
...
-      roots.push_back( fRoots[i].real() );
-  }
-  return roots;
+   std::vector<std::complex<double>> fRoots = FindRoots();
+   std::vector<double> roots;
+   roots.reserve(fOrder);
+   for (unsigned int i = 0; i < fOrder; ++i) {
+      if (fRoots[i].imag() == 0)
+         roots.push_back(fRoots[i].real());
+   }
```
vepadulano, line 244: nitpick, consider the following perhaps a bit more idiomatic code
```cpp
auto roots = FindRoots();
std::vector<double> realRoots;
realRoots.reserve(roots.size());
for (const auto &r: roots)
   if (r.imag() == 0)
      realRoots.push_back(r.real());
```

vepadulano: Now that I look at this twice, I see the for loop is stopping after `fOrder` iterations. Although it's not clear from the code whether the vector returned by `FindRoots` will have a size greater (or even smaller which would be UB!) than `fOrder`

vepadulano (commented):
Thanks! A few comments with suggestions for minor improvements.

math/mathmore/src/Polynomial.cxx
meiyasan: well.. I agree the historical implementation sounds risky; what do you think about comparing double and strict 0 integers, shouldn't we implement an epsilon limit ? 
additionally, considering we are removing `std::vector::reverse()` in other cases, we could also drop `realRoots.reserve(roots.size());` here.

here would be my preference for simplicity purpose.
```cpp
auto roots = FindRoots();
std::vector<double> realRoots;
for (const auto &r: roots)
   if (std::abs(r.imag()) < std::numeric_limits<double>::epsilon())
      realRoots.push_back(r.real());
```

PR #20541, opened 2025-11-26
[vecops] Alignment-related fix for getting first element of small vector

guitargeek:
In the RVec code, we assumed that the location of the SmallVectorBase subobject inside the RVec is identical to the RVec itself.

This assumption is not always correct. For example, on AArch64 GCC 14, some extra padding seems to be inserted before the base class.

This commit generalized the implementation of getting the first element by first casting to the SmallVectorBase class, which is what is required to calculate the correct offset to the SmallVectorBase data members.

We can see in this PRs CI run if it actually works:


guitargeek:
Didn't help...

PR #20543, opened 2025-11-26
[vecops] Adaptive size of long RVec instances in RVec test

guitargeek:
For the tests to make sense, some vectors need to be longer than the
maximum small vector size. This maximum size is compiler and
architecture dependent. For `RVec<int>` on ARM64 with gcc 14, the
small vector capacity turns out to be 60, which is larger than the
current test vector size of 18.

Therefore, the test is refactored to adapt the large vector size
automatically, according to the maximum small vector size.

We can see in another PR how the commit in this PR would be doing on Alma 10 ARM 64 (gcc 14):

### DIFF
--- a/math/vecops/test/vecops_rvec.cxx
+++ b/math/vecops/test/vecops_rvec.cxx
@@ -1672,15 +1672,35 @@ TEST_P(VecOpsSwap, BothSmallVectors)
 
 TEST_P(VecOpsSwap, BothRegularVectors)
 {
-   RVec<int> fixed_vreg1{1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3};
-   RVec<int> fixed_vreg2{4, 5, 6, 4, 5, 6, 4, 5, 6, 4, 5, 6, 4, 5, 6, 4, 5, 6};
-   RVec<int> fixed_vreg3{7, 8, 9, 7, 8, 9, 7, 8, 9, 7, 8, 9, 7, 8, 9, 7, 8, 9, 7};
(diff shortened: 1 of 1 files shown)


math/vecops/test/vecops_rvec.cxx
```diff
@@ -1672,15 +1672,37 @@ TEST_P(VecOpsSwap, BothSmallVectors)
...
 TEST_P(VecOpsSwap, BothRegularVectors)
 {
-   RVec<int> fixed_vreg1{1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3, 1, 2, 3};
-   RVec<int> fixed_vreg2{4, 5, 6, 4, 5, 6, 4, 5, 6, 4, 5, 6, 4, 5, 6, 4, 5, 6};
-   RVec<int> fixed_vreg3{7, 8, 9, 7, 8, 9, 7, 8, 9, 7, 8, 9, 7, 8, 9, 7, 8, 9, 7};
+   constexpr std::size_t smallVecSize = ROOT::Internal::VecOps::RVecInlineStorageSize<int>::value;
+
+   constexpr std::size_t nElems = 72;
+   static_assert(nElems > smallVecSize,
+                 "The RVec instances in this test should be larger than the maximum small vector size!");
```
vepadulano: What about `smallVecSize + 1`?

guitargeek: Right, we can also make this adaptive. I have updated the logic to use the smallest multiple of three that is larger than the `smallVecSize`, so that we preserve this cyclic value pattern of the original test.

PR #20550, opened 2025-11-27
Support interactive TBrowser in python

linev:
Following approach from #17587

Implement alternative TBrowser::Draw() which runs event loop when block option is on.
So one can use TBrowser (web and classic) from any python session just doing:

```
br = ROOT.TBrowser()
br.Draw(block=True)
```
Or one can directly provide block parameter to the constructor:
```
br = ROOT.TBrowser(block=True)
```

Move `run_root_event_loop` function to `__init__.py` script - while it used from canvas and from browser

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/__init__.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/__init__.py
@@ -376,9 +376,8 @@ def _wait_press_posix():
    finally:
       termios.tcsetattr(sys.stdin, termios.TCSADRAIN, old_settings)
 
-# \endcond
 
-def run_root_event_loop():
+def _run_root_event_loop():
    from ROOT import gROOT
    import os
    import sys
@@ -397,4 +396,7 @@ def run_root_event_loop():
       _wait_press_windows()
    else:
       _wait_press_posix()
+
+
+# \endcond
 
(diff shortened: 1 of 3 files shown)


bindings/pyroot/pythonizations/python/ROOT/_pythonization/__init__.py
```diff
@@ -337,4 +337,64 @@ def _register_pythonizations():
...
+         if (c == '\x20'):
+            break
+         time.sleep(0.01)
+
+   finally:
+      termios.tcsetattr(sys.stdin, termios.TCSADRAIN, old_settings)
+
 # \endcond
+
+def run_root_event_loop():
```
vepadulano:
```suggestion
def _run_root_event_loop():
```

I think this function is not "public API", so we can signal it in the naming convention

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tbrowser.py
```diff
@@ -0,0 +1,72 @@
...
+
+r'''
+\pythondoc TBrowser
+
+Functionality of constructor and Draw() methods were extended to support interactive
+work in the python scripts. If extra block parameter is True, script execution
+will be suspended until <space> key pressed by user. Simple example:
+
+\code{.py}
+\endcode
```
vepadulano, line 19: I think this \endcode should be placed at the end of the Python snippet

vepadulano (approved):
Thanks! Some minor comments from my side

bindings/pyroot/pythonizations/python/ROOT/_pythonization/__init__.py
linev: Done

PR #20555, opened 2025-11-27
[hist] Implement initial `RHistConcurrentFiller`

hahnjo:
In order to support efficient concurrent filling of `RHist` with global histogram statistics, we keep one local `RHistStats` object per `RHistFillContext`.


hist/histv7/doc/CodeArchitecture.md
```diff
@@ -62,6 +62,17 @@ It can be used as a template argument to `RHistEngine` and `RHist`.
...
+## Classes for Concurrent Filling
+
+### `RHistConcurrentFiller`
+
+A class to orchestrate concurrent filling of `RHist` by creating (multiple) fill contexts.
+
+### `RHistFillContext`
+
+Parallel user code uses contexts to fill `RHist`s concurrently.
+To uses a local `RHistStats` object to avoid contention on the global histogram statistics.
```
hageboeck: This could be rephrased.

hist/histv7/inc/ROOT/RHistConcurrentFiller.hxx
```diff
@@ -0,0 +1,93 @@
...
+         }
+      }
+   }
+
+   const std::shared_ptr<RHist<BinContentType>> &GetHist() const { return fHist; }
+
+   /// Create a new context for concurrent filling.
+   std::shared_ptr<RHistFillContext<BinContentType>> CreateFillContext()
+   {
+      std::lock_guard g(fMutex);
```
hageboeck: One could maybe use the c++17 alternative:
```suggestion
      std::scoped_lock g(fMutex);
```

hist/histv7/inc/ROOT/RHistConcurrentFiller.hxx
```diff
@@ -0,0 +1,93 @@
...
+
+      // Cannot use std::make_shared because the constructor of RHistFillContext is private. Also it would mean that the
+      // (direct) memory of all contexts stays around until the vector of weak_ptr's is cleared.
+      std::shared_ptr<RHistFillContext<BinContentType>> context(new RHistFillContext<BinContentType>(*fHist));
+      fFillContexts.push_back(context);
```
hageboeck: Possibly move the lock after the creation of the `shared_ptr` to hold it for a shorter duration?

hist/histv7/test/hist_concurrent.cxx
```diff
@@ -0,0 +1,160 @@
...
+            if (i % FlushEveryNFills == 0) {
+               context->Flush();
+            }
+         }
+      });
+   }
+
+   EXPECT_EQ(hist->GetBinContent(0), NFills * Weight);
+   EXPECT_EQ(hist->GetNEntries(), NFills);
+   EXPECT_FLOAT_EQ(hist->ComputeNEffectiveEntries(), NFills);
```
hageboeck, line 140: ~~These two should not be the same if `Weight != 1` ?~~
(Only if the weights differ)

hist/histv7/inc/ROOT/RHistConcurrentFiller.hxx
hahnjo: As far as I understand, `std::lock_guard` is preferable for exactly one mutex.

hist/histv7/inc/ROOT/RHistConcurrentFiller.hxx
hahnjo: Yes, I think that's valid. There was actually a related issue where the `RHistFillContext` constructor would look at values of the `RHistStats` while other contexts could already be flushing. I tried to be future-proof here, but I think we will have to reconstruct the stats object from scratch...

hist/histv7/inc/ROOT/RHistFillContext.hxx
```diff
@@ -0,0 +1,116 @@
...
+\warning This is part of the %ROOT 7 prototype! It will change without notice. It might trigger earthquakes.
+Feedback is welcome!
+*/
+template <typename BinContentType>
+class RHistFillContext final {
+   friend class RHistConcurrentFiller<BinContentType>;
+
+private:
+   /// A pointer to the filled histogram
+   RHist<BinContentType> *fHist;
```
jblomer: Just in case we add/change constructor code later.

```suggestion
   RHist<BinContentType> *fHist = nullptr;
```

hist/histv7/inc/ROOT/RHistConcurrentFiller.hxx
```diff
@@ -0,0 +1,96 @@
...
+            // exit with an exception, it’s a bad design error and the program had better terminate".
+            std::terminate(); // GCOVR_EXCL_LINE
+         }
+      }
+   }
+
+   const std::shared_ptr<RHist<BinContentType>> &GetHist() const { return fHist; }
+
+   /// Create a new context for concurrent filling.
+   std::shared_ptr<RHistFillContext<BinContentType>> CreateFillContext()
```
jblomer, line 78: Since this is thread-safe, we could make `fFillContexts` mutable and this method `const`.

hahnjo: In principle we could. However, we would also have to make `fMutex` mutable for locking / unlocking, and the actual `fHist` is modified indirectly via the returned fill context. Given that the sole purpose of `RHistConcurrentFiller` is to create fill contexts, I think I would prefer to keep it non-`const`.

PR #20561, opened 2025-11-28
[CI] Add a github action linter.

hageboeck:
The linter checks all action configs in the .github/workflows folder.

Given that we have some work to do, it's for now set to `fail-on-error: false`. The step will pass, but generate warnings.

### DIFF
--- a/.github/workflows/lint-action-files.yml
+++ b/.github/workflows/lint-action-files.yml
@@ -1,0 +1,22 @@
+name: 'Lint Github Action Files'
+
+on:
+  pull_request:
+    branches:
+      - '**'
+    paths:
+      - '.github/workflows/*.yml'
+
+jobs:
+  lint-action-files:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-action-files.yml
```diff
@@ -0,0 +1,22 @@
...
+  lint-action-files:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: actionlint
+        uses: raven-actions/actionlint@v2
+        with:
+          files: '.github/workflows/*.yml'
```
pcanal, line 20: Does that ends up also testing the included yml files (i.e. the one in subdirectories)?

hageboeck: @pcanal, no, when these files were included, there were a lot more errors. I was busy enough with the top-level folder. 🙂
I suggest we progressively extend the coverage once we touch the subdirectories.

pcanal: Fair.  Consider adding a comment though, to clarify that the omission is intentional.

PR #20562, opened 2025-11-28
[cling] Drop isOpen() check in ClangInternalState

devajithvs:
# This Pull request:
Removing this distinction between `ParsedOpen` and `Parsed` will help us rid of the clang patch:
`Accessor to "is file opened"; this is crucial info for us.`

This is only used in cling-tests and in `TCling::IsLoaded`, but the latter doesn't care about the distinction between `ParsedOpen` and `Parsed`

After running all the clingtests, this codepath was also never taken in any of those tests.

Will update LLVM tag after review.

## Checklist:

- [ ] tested changes locally

This PR fixes #


vgvassilev (approved):
LGTM!

interpreter/cling/lib/Interpreter/ClangInternalState.cpp
hahnjo: Probably want to remove the variable entirely
```suggestion
    std::vector<std::string> Parsed, AST;
```

PR #20563, opened 2025-11-28
[CMake] Use output name of library target also for dictionary name

guitargeek:
Use output name of the library target also for the dictionary name, if the target is defined and the output name is different from the target name. This makes the dictionary and library file names more consistent.

Closes https://github.com/root-project/root/issues/15792.


guitargeek:
Thanks! No risk of breaking the users code, because this is only changing behavior for a case that did work before anyway: when the `OUTPUT_NAME` of the library was different from the target name. Before, the dictionaries could not be found because the filenames didn't match, and now this case also works.

bellenot:
> Thanks! No risk of breaking the users code, because this is only changing behavior for a case that did work before anyway: when the `OUTPUT_NAME` of the library was different from the target name. Before, the dictionaries could not be found because the filenames didn't match, and now this case also works.

Cool, thanks Jonas!

PR #20567, opened 2025-11-28
[build] Comply with PEP625

vepadulano:
PEP625 mandates that distribution names be normalised https://peps.python.org/pep-0625/. The full specification is available at https://packaging.python.org/en/latest/specifications/binary-distribution-format/#binary-distribution-format. For the ROOT package, this means that the wheel name must have the lowercase spelling of the project name.


vepadulano:
Before merging I will wait for the jobs at https://github.com/root-project/root/actions/runs/19771134407 to finish. Not sure why they don't appear in the github UI of this PR.

PR #20569, opened 2025-11-28
[Geom] Fix compiler warnings in TGDMLWrite

guitargeek:
This follows up on 74e607e6c08, which introduced some compiler warnings that cause the Debian CI to fail.

Same situation as in https://github.com/root-project/root/pull/20351, where the failures didn't appear in the CI run for the PR that made the changes yet...


pcanal (approved):
LGTM.  The git log message would been enhanced by copy/pasted the warning message.

guitargeek:
Sorry, I'm not with my laptop to update this now, and it would be nice to have green nightlies

PR #20574, opened 2025-11-29
io: Support 64 bits position/byte-count 

pcanal:


io/io/src/TBufferFile.cxx
```diff
@@ -374,6 +374,7 @@ void TBufferFile::SetByteCount(ULong64_t cntpos, Bool_t packInVersion)
       tobuf(buf, cnt | kByteCountMask);
 
    if (cnt >= kMaxMapCount) {
+      // NOTE: This should now be unreachable.
```
silverweed, line 396: Then maybe let's put an assert on it

io/io/inc/TBufferFile.h
```diff
@@ -52,6 +52,16 @@ class TBufferFile : public TBufferIO {
    TStreamerInfo  *fInfo{nullptr};  ///< Pointer to TStreamerInfo object writing/reading the buffer
    InfoList_t      fInfoStack;     ///< Stack of pointers to the TStreamerInfos
 
+   using ByteCountLocator_t = std::size_t; // This might become a pair<chunk_number, local_offset> if we implement chunked keys
+   using ByteCountStack_t = std::vector<ByteCountLocator_t>;
+   ByteCountStack_t fByteCountStack; ///<! Stack to keep track of byte count storage positions
+
+   using ByteCount_t = std::size_t; ///< Type used to store byte count values, can be changed to uint32_t if we implement chunked keys
```
jblomer: Shouldn't this be machine-independent, i.e. a `std::uint64_t`? I also don't see why this would become a `uint32_t`. Even if dispersed into multiple chunked keys, the _object_ is still large.

io/io/src/TBufferFile.cxx
```diff
@@ -3327,6 +3327,22 @@ UInt_t TBufferFile::CheckObject(UInt_t offset, const TClass *cl, Bool_t readClas
    return offset;
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// Reserve space for a leading byte count and return the position where to
+/// store the byte count value.
+///
+/// \param[in] mask The mask to apply to the placeholder value (default kByteCountMask)
```
jblomer: This parameter in fact doesn't exist

io/io/src/TBufferFile.cxx
```diff
@@ -3327,6 +3327,22 @@ UInt_t TBufferFile::CheckObject(UInt_t offset, const TClass *cl, Bool_t readClas
...
+/// \return The position (cntpos) where the byte count should be stored later,
+///         or 0 if the position exceeds kMaxInt
```
jblomer: The code returns `kMaxInt`, not zero, in case of a far position.

io/io/src/TBufferFile.cxx
```diff
@@ -360,10 +360,25 @@ void TBufferFile::SetByteCount(ULong64_t cntpos, Bool_t packInVersion)
 
 Long64_t TBufferFile::CheckByteCount(ULong64_t startpos, ULong64_t bcnt, const TClass *clss, const char *classname)
 {
-   if (!bcnt) return 0;
-   R__ASSERT(startpos <= kMaxUInt && bcnt <= kMaxUInt);
-   Long64_t offset = 0;
+   R__ASSERT(!fByteCountStack.empty());
+   if (startpos == kMaxInt) {
+      // The position is above 1GB and was cached using a 32 bit variable.
```
jblomer: I'm confused about the exact limit that separates a close position from a far position. Is it `kMaxInt` (2GB) or `kByteCountMask` (1GB)?

io/io/src/TBufferFile.cxx
```diff
@@ -360,10 +360,25 @@ void TBufferFile::SetByteCount(ULong64_t cntpos, Bool_t packInVersion)
 
 Long64_t TBufferFile::CheckByteCount(ULong64_t startpos, ULong64_t bcnt, const TClass *clss, const char *classname)
 {
-   if (!bcnt) return 0;
-   R__ASSERT(startpos <= kMaxUInt && bcnt <= kMaxUInt);
-   Long64_t offset = 0;
+   R__ASSERT(!fByteCountStack.empty());
+   if (startpos == kMaxInt) {
```
jblomer: Maybe also assert that `startpos` is never larger than the limit.

jblomer: Also: we unconditionally push the position onto the stack. So we could also take it from there unconditionally.

io/io/src/TBufferFile.cxx
```diff
@@ -360,10 +360,25 @@ void TBufferFile::SetByteCount(ULong64_t cntpos, Bool_t packInVersion)
...
+      // The position is above 1GB and was cached using a 32 bit variable.
+      startpos = fByteCountStack.back();
+   }
+   if (bcnt == kMaxInt) {
+      // in case we are checking a byte count for which we postponed
+      // the writing because it was too large, we retrieve it now
+      auto it = fByteCounts.find(startpos);
+      if (it != fByteCounts.end()) {
+         bcnt = it->second;
+      }
```
jblomer, line 430: else assert?

io/io/src/TBufferFile.cxx
```diff
@@ -360,10 +360,25 @@ void TBufferFile::SetByteCount(ULong64_t cntpos, Bool_t packInVersion)
...
+   }
+   fByteCountStack.pop_back();
 
+   if (!bcnt || startpos == kMaxInt)
```
jblomer: Why are we checking for `startpos == kMaxInt`?

io/io/src/TBufferFile.cxx
```diff
@@ -2749,6 +2764,8 @@ TClass *TBufferFile::ReadClass(const TClass *clReq, UInt_t *objTag)
       bcnt = 0;
    } else {
       fVersion = 1;
+      if (objTag)
```
jblomer, line 2826: I'd need a comment here why we have a byte count (only) in this case

io/io/src/TBufferFile.cxx
```diff
@@ -2956,9 +2976,29 @@ Version_t TBufferFile::ReadVersion(UInt_t *startpos, UInt_t *bcnt, const TClass
       fBufCur -= sizeof(UInt_t);
       v.cnt = 0;
    }
-   if (bcnt) *bcnt = (v.cnt & ~kByteCountMask);
+   if (bcnt) {
+      if (!v.cnt) {
+         // no byte count stored
+         *bcnt = 0;
+         fByteCountStack.pop_back();
```
jblomer, line 3047: Should we check for `startpos` before removing from the stack?

io/io/src/TBufferFile.cxx
```diff
@@ -2956,9 +2976,29 @@ Version_t TBufferFile::ReadVersion(UInt_t *startpos, UInt_t *bcnt, const TClass
...
+      }
+   }
    frombuf(this->fBufCur,&version);
 
+   // NOTE: The code above is not straightforward to refactor by a call
+   // to ReadVersionNoCheckSum because of the following code need the
+   // 'raw' value in `v`.
+
    if (version<=1) {
       if (version <= 0)  {
```
jblomer, line 3066: Side note: may be easier to read by returning early if `version > 0`.  The remainder of the function can then be back-indented by two levels.

io/io/src/TBufferFile.cxx
jblomer, line 3147: We don't use this method anywhere in our code. Can it be removed?

jblomer, line 3008: If both `startpos` and `bcnt` are provided, the code _may_ add the position to `fByteCountStack` (it doesn't if the serialized object doesn't have a byte count). That means that the caller of `ReadVersion` needs to pop from the stack, correct? This looks a bit brittle and should at least be documented.

jblomer: Adjust assert (I think now `cntpos` can never by larger than `kMaxInt`.

io/io/src/TBufferFile.cxx
```diff
@@ -323,7 +323,26 @@ void TBufferFile::SetByteCount(ULong64_t cntpos, Bool_t packInVersion)
...
+   // routines.
+   R__ASSERT( !fByteCountStack.empty() );
+   if (cntpos == kMaxInt) {
+      cntpos = fByteCountStack.back();
+   }
+   fByteCountStack.pop_back();
+   // if we are not in the same TKey chunk or if the cntpos is too large to fit in UInt_t
+   // let's postpone the writing of the byte count
+   const ULong64_t full_cnt = ULong64_t(fBufCur - fBuffer) - cntpos - sizeof(UInt_t);
+   if (full_cnt >= kMaxMapCount) {
```
jblomer: I think this also should converge to one limit constant (`kMaxInt` ?)

io/io/src/TBufferFile.cxx
```diff
@@ -323,7 +323,26 @@ void TBufferFile::SetByteCount(ULong64_t cntpos, Bool_t packInVersion)
...
+      fByteCounts[cntpos] = full_cnt;
+      return;
+   }
+   UInt_t cnt = static_cast<UInt_t>(full_cnt);
```
jblomer, line 374: Can't this overflow?

io/io/test/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ ROOT_ADD_GTEST(TBufferFile TBufferFileTests.cxx LIBRARIES RIO)
 ROOT_ADD_GTEST(TBufferMerger TBufferMerger.cxx LIBRARIES RIO Imt Tree)
 ROOT_ADD_GTEST(TBufferJSON TBufferJSONTests.cxx LIBRARIES RIO)
 ROOT_ADD_GTEST(TFileMerger TFileMergerTests.cxx LIBRARIES RIO Tree Hist)
+ROOT_ADD_GTEST(TBufferFileInnerReferences InnerReferencesTests.cxx LIBRARIES RIO)
```
jblomer: Seems missing

io/io/src/TBufferFile.cxx
pcanal: I don't think so because of the early return on line 351-353 (and `kMaxMapCount < std::numeric_limits<UInt_t>::max()`

io/io/inc/TBufferFile.h
pcanal: I guess that I was assuming you could not deal with large buffers on 32bits platforms ... but you might be able to wiggle to slightly more than 1GB, so indeed I will make the change.

io/io/src/TBufferFile.cxx
pcanal: Instead I diverged in using distinct semantically named constant.   Here the limit is 'byte count that can be represented on file in 32 bits after the mask is applied, so 1GB).

io/io/src/TBufferFile.cxx
pcanal: It is now 'kMaxCountPosition' (i.e. `kMaxUInt-1`)

io/io/src/TBufferFile.cxx
pcanal: One idea is to (in a future update) add new overloads that detects that the cache variables are 64 bits and skip the stacking in those cases.  See comment on line 337 in `void TBufferFile::SetByteCount(ULong64_t cntpos, Bool_t packInVersion)`

In regard to the current code, it exchanges, in the common case, a comparison and jump for a dereference and assignment.

io/io/src/TBufferFile.cxx
pcanal: No reason.  It was actually working as the long range start position could be equal to the overflow value (now `kMaxUInt`).

io/io/src/TBufferFile.cxx
pcanal: I added a comment:
```
      // When objTag is not used, the caller is not interested in the byte 
      // count and will not (can not) call CheckByteCount.
```
This pattern is actually used in the IO code to 'peek' at the class name.

io/io/src/TBufferFile.cxx
pcanal: If the user does not provide the address of their bytecount variable, they will not be given the value of the byte count and thus they can not and will not call `CheckByteCount`.
If the user has provided the address, we assume they are using it in a call to `CheckByteCount`.
I will add to the doc string that the CheckByteCount call is (now) required.

io/io/test/CMakeLists.txt
pcanal: This line is a stray line not meant for this PR.

io/io/src/TBufferFile.cxx
pcanal: Yep, stray line left from a previous implementation.

io/io/src/TBufferFile.cxx
pcanal: This is a public interface (that is missing a test :( ).  It was introduced with this git log (which I added as a documentation):
```
    This (new) routine can be used from custom streamers
    when it is known that the class was always versioned
    (and thus never stored a checksum within the buffer).
    This allows to disambiguate the case where the class
    used to have a version number equal to zero and did
    not save a byte count and is now versioned.
    
    When reading the version number being zero, without
    the byte count we have no way to guess whether the
    class was version un-versioned or had previously a
    version number equal to zero.
```

pcanal:
> We may want to swap the reading and writing commits so that the code always works correctly on small objects.

Currently the reading commit is first and should be able to read small buffers ... or am I missing something?

pcanal:
> I have difficulties parsing the first sentence of the commit message of the first commit.

New wording:

```
io: Add data structure for 64 bit byte count.
    
In order to support 64 bits byte counts, which do not fit in the space reserved
for them in the stream (4 bytes minus 2 control bits) and to work around the
fact that the variables that holds the position and the byte count information in
Streamer functions are only 32 bits (see arguments to TBufferFile::WriteVersion
and TBufferFile::ReadVersion), we need to pass them indirectly when they needs 64 bits

Since the streaming is inherently serial, we can leverage the
sequence of calls and cache the 64 bits values in a queue.

The bytecount that can not be stored in place inside the io stream will be held
in a collection (fByteCounts) that need to be stored externally to the
buffer.
```

> Also, this sentence mentions three control bits. Isn't it just two (up to kByteCountMask, i.e. 0x40000000)?

Correct.

pcanal:
@Jakob This PR is kept open but put on Draft mode to keep the conversation going.  However because of the branch naming limitation, I have to move the testing (and merging) to a new PR: https://github.com/root-project/root/pull/21183 which has a cleaned up (final?) list of commits.

Issue #20572, opened 2025-11-29
A public field with a "private" type in the TFileMergeInfo class


grasph:
### Explain what you would like to see improved and how.

The field [`fIOFeatures`](https://github.com/root-project/root/blob/1df6d55224023993c59adf911135f13e6bd99074/core/meta/inc/TFileMergeInfo.h#L54) of `TFileMergeInfo` is public while its type is defined as a private alias.

In my opinion for consistency, if `fIOFeatures` needs to be public, then  the explicit public type `ROOT::TIOFeatures` should be use or the alias must be made public. 

I've found this, while running [WrapIt](http://github.com/grasph/wrapit) that generated a function returning `ROOT::TIOFeatures`. The generated code is not valid as this alias is private.

### ROOT version

E.g. 6.32.08 or github master HEAD

### Installation method

N/A

### Operating system

N/A

PR #20576, opened 2025-11-30
[core] publish alias since it appears in public function ret type or arg

ferdymercury:
Fixes https://github.com/root-project/root/issues/20572

@grasph would this change solve your issue?

### DIFF
--- a/core/rint/test/TTabComTests.cxx
+++ b/core/rint/test/TTabComTests.cxx
@@ -95,8 +95,13 @@ TEST(TTabComTests, CompleteTProfile)
 
 TEST(TTabComTests, CompleteTObj)
 {
+   #ifdef R__USE_CXXMODULES
    std::string expected = "TObjArray TObjArrayIter TObjLink TObjLinkPtr_t TObjOptLink"
       " TObjString TObject TObjectRefSpy TObjectSpy TObjectTable";
+   #else
+   std::string expected = "TObjArray TObjArrayIter TObjLink TObjOptLink"
+      " TObjString TObject TObjectRefSpy TObjectSpy TObjectTable";
+   #endif
    // FIXME: See ROOT-10989
    ASSERT_STREQ(expected.c_str(), GetCompletions("TObj",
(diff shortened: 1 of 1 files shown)


grasph:
Yes, it does. Thanks. There few other similar cases:

- [`TList::Remove(const TObjLinkPtr_t &lnk)`](https://github.com/root-project/root/blob/8a4a8cbc31ce72e8254f08949c6af9ff7ceee7a8/core/cont/inc/TList.h#L94C22-L94C54),  [`TObjLink::NextSP()`](https://github.com/root-project/root/blob/8a4a8cbc31ce72e8254f08949c6af9ff7ceee7a8/core/cont/inc/TList.h#L151), and [ TObjOptLink::PrevSP](https://github.com/root-project/root/blob/8a4a8cbc31ce72e8254f08949c6af9ff7ceee7a8/core/cont/inc/TList.h#L152)  require the protected alias `TObjLinkPtr_t`
- [`TBranch::GetIOFeatures`](https://github.com/root-project/root/blob/8a4a8cbc31ce72e8254f08949c6af9ff7ceee7a8/tree/tree/inc/TBranch.h#L244) and `TBranch::SetIOFeatures` require the private `TIOFeature` alias
- [`TLeaf::GetLeafCountValues`](https://github.com/root-project/root/blob/8a4a8cbc31ce72e8254f08949c6af9ff7ceee7a8/tree/tree/inc/TLeaf.h#L124) requires the protected alias `Counts_t`
- [`TGeoFacet::CompactFacet`](https://github.com/root-project/root/blob/8a4a8cbc31ce72e8254f08949c6af9ff7ceee7a8/geom/geom/inc/TGeoTessellated.h#L44) requires the private `Vertex_t` alias

vepadulano:
@pcanal what do you think about this one?

ferdymercury:
> Thanks.

It seems TTabCom test is failing on some platforms. Not sure if it's a real issue, or if it's just a clean-build label missing?

EDIT: Danilo found out: it's because Windows has no pcms and the failing alma9-build is with modules=OFF

See also: https://github.com/root-project/root/pull/20716

PR #20577, opened 2025-11-30
[tmva][pymva] Exercise removal of deprecated TMVA `PyKeras` method

guitargeek:
The new Keras 3 API broke the existing TMVA `PyKeras` method.

Also exercise the removal in some temporary commits, to check if removing `PyKeras` is sufficient to fix `tmva-pymva`.

Alternative to:
  * https://github.com/root-project/root/pull/15790


guitargeek:
Draft PR can be closed, because it was only to show that removal of the PyKeras method would be easy, should be decide to deprecate it.

Issue #15537, opened 2024-05-16
[cling] Crash when non-void function does not return a value


ferdymercury:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

If you forget a return value in a non-void function, ROOT (cling) crashes. This does not happen if you compile via AcLIC.

### Reproducer

```
int badcast()
{
}
```

root -l badcast.cpp

```
===========================================================
#11 0x00007d4323de7008 in ?? ()
#12 0x00007fffe3dd2010 in ?? ()
#13 0x00007d4323dea02d in ?? ()
#14 0x00007fffe3dd26a0 in ?? ()
#15 0x00005ab8d4e3b860 in ?? ()
#16 0x00007fffe3dd26a0 in ?? ()
#17 0x00007d431babdda8 in cling::IncrementalExecutor::executeWrapper(llvm::StringRef, cling::Value*) const () from /home/user/build/build-root-Desktop-Debug/lib/libCling.so
#18 0x00007d431ba40ddc in cling::Interpreter::RunFunction(clang::FunctionDecl const*, cling::Value*) () from /home/user/build/build-root-Desktop-Debug/lib/libCling.so
#19 0x00007d431ba41527 in cling::Interpreter::EvaluateInternal(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, cling::CompilationOptions, cling::Value*, cling::Transaction**, unsigned long) () from /home/user/build/build-root-Desktop-Debug/lib/libCling.so
#20 0x00007d431bb2eeef in cling::MetaSema::actOnxCommand(llvm::StringRef, llvm::StringRef, cling::Value*) () from /home/user/build/build-root-Desktop-Debug/lib/libCling.so
#21 0x00007d431bb3c767 in cling::MetaParser::isXCommand(cling::MetaSema::ActionResult&, cling::Value*) () from /home/user/build/build-root-Desktop-Debug/lib/libCling.so
#22 0x00007d431bb3e1d9 in cling::MetaParser::isCommand(cling::MetaSema::ActionResult&, cling::Value*) () from /home/user/build/build-root-Desktop-Debug/lib/libCling.so
#23 0x00007d431bb27790 in cling::MetaProcessor::process(llvm::StringRef, cling::Interpreter::CompilationResult&, cling::Value*, bool) () from /home/user/build/build-root-Desktop-Debug/lib/libCling.so
[...]

PR #20582, opened 2025-12-01
[cling] Treat `-Wreturn-type` as an error

devajithvs:
Cling currently crashes when a non-void function is missing a return, upgrade this warning to an error.

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes https://github.com/root-project/root/issues/15537

### DIFF
--- a/interpreter/cling/lib/Interpreter/CIFactory.cpp
+++ b/interpreter/cling/lib/Interpreter/CIFactory.cpp
@@ -1422,6 +1422,11 @@ namespace {
       argvCompile.push_back("-fno-omit-frame-pointer");
     }
 
+    // Promote -Wreturn-type to an error. A missing return in a non-void
+    // function is a warning by default, and cling cannot safely continue and
+    // crashes later.
+    argvCompile.push_back("-Werror=return-type");
+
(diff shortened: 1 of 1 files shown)


interpreter/cling/lib/Interpreter/CIFactory.cpp
```diff
@@ -1508,6 +1508,10 @@ namespace {
     // to the output and are rather confusing.
     CI->getDiagnosticOpts().SnippetLineLimit = 1;
     CI->getDiagnosticOpts().ShowLineNumbers = 0;
+    // Promote -Wreturn-type to an error. A missing return in a non-void
+    // function is a warning by default, and cling cannot safely continue and
+    // crashes later.
+    CI->getDiagnosticOpts().Warnings.push_back("error=return-type");
```
vgvassilev: Why not passing a cc1 argument `-Werror=return-type` instead?

devajithvs: I decided to put it there because we already modify some `DiagnosticOptions` for errors and warnings. But yeah, passing `-Werror=return-type` is better.

vgvassilev (approved):
LGTM!

PR #20588, opened 2025-12-01
[tmva][pymva] Deprecate the `PyKeras` method

guitargeek:
The new Keras 3 API broke the existing TMVA `PyKeras` method.

Re-implementing the method for Keras 3 turned out to be difficult, as it is not possible anymore to disable eager execution and have a decent speed in evaluating the model for single events. So large-scale refactoring would be necessary to implement `PyKeras` again with good performance (see also https://github.com/root-project/root/pull/15790).

Nowadays, we also have the RBatchGenerator to train Keras models directly with batches that are provided by ROOT. Therefore, the TMVA `PyKeras` method is not the only way anymore to train a Keras model with ROOT data without 3rd party libraries for the IO. That means it's not an essential feature anymore, and deprecating it will even make the situation clearer for the user, as there are not two different ways anymore to train Keras models on ROOT data.

Note that it doesn't help us in this case to use the deprecation macros, because `tmva-pymva` is always disabled anyway. So we would not benefit from the automatic removal-reminder errors that one would get in the next development cycle.

I also opened a separate PR that exercises the eventual removal in the next development cycle, also showing that after `PyKeras` is gone, we can enable `tmva-pymva` again, as the other methods it provides work just fine (except for for corner cases like PyTorch on Linux ARM):
  * https://github.com/root-project/root/pull/20577


tmva/tmva/inc/TMVA/Types.h
```diff
@@ -100,7 +100,7 @@ namespace TMVA {
          kPyRandomForest ,
          kPyAdaBoost     ,
          kPyGTB          ,
-         kPyKeras        ,
+         kPyKeras, // deprecated, will be removed in ROOT 6.42
```
silverweed: Should we mark this with `R__DEPRECATED`?

guitargeek: I don't think you can mark individual enum elements as deprecated with `R__DEPRECATED` (it's only for functions and classes).

silverweed: Actually [you can](https://en.cppreference.com/w/cpp/language/attributes/deprecated) since C++17! (and we're doing it already, see RNTupleTypes.hxx)

lmoneta:
I would wait to deprecate the PyKeras interface.  Once we merge https://github.com/root-project/root/pull/15790, the work in refactoring is already done there, we can easily support Keras 3.

guitargeek:
Hi @lmoneta, thanks for the comment! I have recently tried your #15790 and it didn't work, which is why I thought it might be unreasonably difficult to support Keras 3 in PyMVA. If you update that PR, we can of course re-evaluate the situation!

guitargeek:
Let's give some more time to @lmoneta to fix PyMVA.

guitargeek:
Reopened for now because I wanted to try out the suggestion by @silverweed to mark enum values as deprecated.

dpiparo:
@lmoneta is there an ETA?

guitargeek:
PyKeras works again!

  * https://github.com/root-project/root/pull/15790

PR #20594, opened 2025-12-01
[cppyy] Remove now irrelevant `load_cpp_backend()` call

guitargeek:
As we don't have a separate `cppyy_backend` anymore (neither as a Python package, nor as a shared library), trying to load the backend library will always give an exception that will be caught as an `ImportError`.

It would be cleaner if this code branch is not even taken.


guitargeek:
Not the best idea to follow `ruff` here, because it will increase the divergence with upstream cppyy even more :(

PR #20597, opened 2025-12-01
[Python][Hist] Extend vectorized Fill pythonization to TH1 and TH2 with generic array-like support

siliataider:
# This Pull request:

## Changes:
This PR extends the vectorized `Fill` Pythonization behavior, we now:
- support **any array-like input** (lists, ranges, NumPy arrays…) --> instead of only NumPy arrays
- provide vectorized filling for **TH2 histograms** in addition to TH1 --> previously only 1D histograms supported

## Checklist:

- [x] tested changes locally

### DIFF
--- a/bindings/pyroot/pythonizations/test/th2_fillN.py
+++ b/bindings/pyroot/pythonizations/test/th2_fillN.py
@@ -28,10 +28,7 @@ class FillWithArrayLike(unittest.TestCase):
 
         for i in range(nbins):
             for j in range(nbins):
-                self.assertAlmostEqual(
-                    verbose_hist.GetBinContent(i, j),
-                    simple_hist.GetBinContent(i, j)
-                )
+                self.assertAlmostEqual(verbose_hist.GetBinContent(i, j), simple_hist.GetBinContent(i, j))
 
         # Test filling with weights
         weights_np = np.linspace(0.1, 0.9, n)
@@ -42,10 +39,7 @@ class FillWithArrayLike(unittest.TestCase):
 
         for i in range(nbins):
             for j in range(nbins):
-                self.assertAlmostEqual(
-                    verbose_hist.GetBinContent(i, j),
-                    simple_hist.GetBinContent(i, j)
-                )
+                self.assertAlmostEqual(verbose_hist.GetBinContent(i, j), simple_hist.GetBinContent(i, j))
 
         # Test mismatched weight size
         with self.assertRaises(ValueError):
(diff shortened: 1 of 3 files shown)


bindings/pyroot/pythonizations/test/th1_fillN.py
```diff
@@ -3,39 +3,59 @@
...
             self.assertAlmostEqual(verbose_hist.GetBinContent(i), simple_hist.GetBinContent(i))
-        # Test filling with weights with a different length
-        weights = np.array([0.1, 0.2, 0.3, 0.4])
+
+        # Test mismatched weight size
         with self.assertRaises(ValueError):
-            simple_hist.Fill(data, weights)
+            simple_hist.Fill(data, [0.1, 0.2, 0.3])  # too short
+
+    # Run with différent inputs
```
vepadulano: small typo in `different`

bindings/pyroot/pythonizations/test/th2_fillN.py
```diff
@@ -0,0 +1,71 @@
...
+                self.assertAlmostEqual(
+                    verbose_hist.GetBinContent(i, j),
+                    simple_hist.GetBinContent(i, j)
+                )
+
+        # Test mismatched weight size
+        with self.assertRaises(ValueError):
+            simple_hist.Fill(x_data, y_data, [0.1, 0.2, 0.3])  # too short
+
+    # Run with différent inputs
```
vepadulano: Also here small typo

bindings/pyroot/pythonizations/test/th1_fillN.py
siliataider: thanks

fwinkl:
Since this has been merged we are seeing the following type of failures in the LCG dev3 nightlies:
```Python
>>> from ROOT import TH1D
>>> h = TH1D("h", "", 5, 0, 5)
>>> h.Fill("a", 1)
  File "<python-input-3>", line 1, in <module>
    h.Fill("a", 1)
    ~~~~~~^^^^^^^^
  File "/cvmfs/sft-nightlies.cern.ch/lcg/nightlies/dev3/Sat/ROOT/HEAD/x86_64-el9-gcc14-opt/lib/ROOT/_pythonization/_th1.py", line 218, in _FillWithArrayTH1
    data = np.asanyarray(args[0], dtype=np.float64)
ValueError: could not convert string to float: 'a'
```
Probably you want to extend your unit tests for this case (unless this syntax is deprecated?).

scott-snyder:
No, that should be perfectly valid, since TH1::Fill has an overload taking a const char* as the first argument.

Previously, the python TH1.Fill was filling an array if the first argument is a numpy array; now it does that if the first argument has len() defined.  But that also selects some things that shouldn't be considered arrays, like strings.

And now that i think about it, TH1.Fill tends to get called a lot --- so it may not be such a good idea to have more python code on top of it in the first place.  Maybe this pythonization should really be done for FillN, not Fill.

PR #20605, opened 2025-12-02
[core] Move FileRaii to TestSupport.hxx

silverweed:
This is a general-purpose utility that can be used by many unit tests so it makes sense to deduplicate it and share it.

Note that for now this doesn't touch duplicated instances of `FileRaii` in roottest, to avoid introducing additional dependencies (e.g. to the dataframe roottest suite)

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/testsupport/inc/ROOT/TestSupport.hxx
+++ b/core/testsupport/inc/ROOT/TestSupport.hxx
@@ -152,6 +152,33 @@ private:
    static CheckDiagsRAII *sActiveInstance; /// Instance that will receive ROOT's callbacks.
 };
 
+///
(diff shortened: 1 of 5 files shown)


core/testsupport/inc/ROOT/TestSupport.hxx
```diff
@@ -152,6 +152,33 @@ private:
    static CheckDiagsRAII *sActiveInstance; /// Instance that will receive ROOT's callbacks.
 };
 
+/**
+ * An RAII wrapper around an open temporary file on disk. It cleans up the guarded file when the wrapper object
+ * goes out of scope.
+ */
```
jblomer: I think we commonly use the `///` style comments.

jblomer (approved):

PR #20608, opened 2025-12-02
[io] add test bed for streaming large objects

jblomer:


tree/ntuple/test/rfield_streamer_beyond.cxx
```diff
@@ -0,0 +1,70 @@
...
+#include <cstdio>
+#include <string>
+#include <utility>
+
+#include "StreamerBeyond.hxx"
+#include "gtest/gtest.h"
+
+namespace {
+
+class FileRaii {
```
silverweed, line 15: This can be pulled from `ntuple_test.hxx`, no?

io/io/src/TBufferFile.cxx
```diff
@@ -323,8 +323,13 @@ void TBufferFile::SetByteCount(ULong64_t cntpos, Bool_t packInVersion)
         && (fBufCur >= fBuffer)
         && static_cast<ULong64_t>(fBufCur - fBuffer) <= std::numeric_limits<UInt_t>::max()
         && "Byte count position is after the end of the buffer");
-   const UInt_t cnt = UInt_t(fBufCur - fBuffer) - UInt_t(cntpos) - sizeof(UInt_t);
+
    char  *buf = (char *)(fBuffer + cntpos);
+   if ((fBufCur - fBuffer - cntpos - sizeof(UInt_t)) >= kMaxMapCount) {
+      tobuf(buf, 0 | kByteCountMask);
```
silverweed: The reason to write `0 | kByteCountMask` is clear for us (today) but it might be confusing to anyone not familiar with the idea behind the plan; maybe a comment would help here.

tree/ntuple/test/rfield_streamer_beyond.cxx
jblomer: Yes, but perhaps just from the test support when the other PR is merged.

jblomer: I wanted to avoid including this large header here.

io/io/src/TBufferFile.cxx
pcanal: On the other hand, this code is already almost obsolete, i.e. replaced by https://github.com/root-project/root/pull/20574 (which is still missing a few wrinkles: decision on Reset behavior, getter and setter)

pcanal: It is now obsolete and has been removed (since support for large byteCount has been merged in the feature branch).

pcanal:
The next step for this PR is the implementation of storage of the storage and retrieval of the large byte count array.

jblomer:
Replaced by #22430

PR #20612, opened 2025-12-02
[roottest][cling/python] Rename CINT to CLING and MAKECINT to ROOTCLING [part 3]

ferdymercury:
A third subpart of https://github.com/root-project/root/pull/19825 as requested by @pcanal


guitargeek (approved):
Thank you! And great you could avoid a `WILLFAIL` in one of the tests :+1:

PR #20613, opened 2025-12-02
[roottest][io] Rename CINT to CLING and MAKECINT to ROOTCLING [part 4]

ferdymercury:
A fourth subpart of https://github.com/root-project/root/pull/19825 as requested by @pcanal


roottest/root/io/customStreamer/header.h
```diff
@@ -2,9 +2,7 @@
 #include "TClass.h"
 class Hard2Stream {
 private:
-#ifndef __CINT__
```
pcanal, line 5: The hiding was intentional and not related to a CINT limitation.  This is an 'important' part of the test as it ensures that the object is not streaming via the default mechanisms.

pcanal: @ferdymercury Can you readd (with `__CLING__`)?

PR #20616, opened 2025-12-02
[CMake] Never add current binary dir to command in `ROOT_ADD_TEST`

guitargeek:
If a test added with `ROOT_ADD_TEST` uses a command that can't be found on the system and that is not an absolute path, the `ROOT_ADD_TEST` macro will implicitly prefix `$CMAKE_CURRENT_BINARY_DIR/`, assimum that the executable is supposed to be found in the current binary dir.

However, that is not always a reasonable fallback, and can be even wrong if the executable was supposed to be found in the `PATH` that is set in the `ENVIRONMENT` for that test.

So I think it's better to remove that fallback, requiring the tests to be explicit about which executable should be called.

### DIFF
--- a/roottest/root/rint/CMakeLists.txt
+++ b/roottest/root/rint/CMakeLists.txt
@@ -21,7 +21,7 @@ ROOTTEST_ADD_TEST(missingSymbol
 if (NOT MSVC)
 # ROOT-5843
 ROOTTEST_ADD_TEST(dotLibraries
-                  COMMAND root -b -q -l -e ".libraries"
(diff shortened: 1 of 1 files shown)


hageboeck:
As discussed in person, only removing the path won't be enough.
```
root <xxx>
```
Would use a system-installed ROOT if present. We either need to retain the path, rewrite the tests to use `$<TARGET_FILE:root>` or alter `PATH` for the tests to succeed. The generator expression seems the safest and most easy to read.

guitargeek:
Thanks! The first commit message is now updated.

PR #20618, opened 2025-12-02
[RF] Don't double-scale pdfs plotted by RooSimultaneous::plotOn()

guitargeek:
After https://github.com/root-project/root/commit/a8ef8b03c845b7ddd6aad266e24d745272b9ef82, when plotting an extended pdf it automatically scales
itself to the number of expected events.

However, when plotting slices of a RooSimultaneous, the normalization is
already precomputed in `RooSimultanous::plotOn()` and should not be
overridden again. To prevent this, flag the calculated normatization as
the final number of events, and not some relative factor, which it is by
default.

Closes https://github.com/root-project/root/issues/20383.

The tutorials are also updated to use less confusing statements to plot slices of the RooSimultaneous.

@VanyaBelyaev, is that okay for you now?

Some unrelated RooFit code modernization commits were added to this PR to keep the net number of added lines of code negative.

### DIFF
--- a/roofit/histfactory/inc/RooStats/HistFactory/Measurement.h
+++ b/roofit/histfactory/inc/RooStats/HistFactory/Measurement.h
@@ -650,9 +650,9 @@ extern Channel BadChannel;
 class Measurement : public TNamed {
 
 public:
-   Measurement();
-   ///  Measurement( const Measurement& other ); // Copy
-   Measurement(const char *Name, const char *Title = "");
+   Measurement() = default;
+   /// Constructor specifying name and title of measurement
+   Measurement(const char *Name, const char *Title = "") : TNamed(Name, Title) {}
 
    ///  set output prefix
    void SetOutputFilePrefix(const std::string &prefix) { fOutputFilePrefix = prefix; }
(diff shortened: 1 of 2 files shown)


VanyaBelyaev:
Dear @guitargeek 
Thank you for the fix! 
As soon as it propagates to the LCG-de3 nightly slot, I'll be happy to test it in my working environment and to report the results. 

And I like a lot *significantly* simplified "interface" without `Slice&ProjWData`.
It always was rather non-intuitive and partly confusing
It is nice to know better alternative exists.  It will take some time to 
update all  my code for new model, but definitely it is a step in right direction.

ferdymercury:
@guitargeek does this PR also clarify/close https://its.cern.ch/jira/browse/ROOT-7499 ?

guitargeek:
> @guitargeek does this PR also clarify/close https://its.cern.ch/jira/browse/ROOT-7499 ?

No. But I have created another PR for that:
  * https://github.com/root-project/root/pull/20625

guitargeek:
> And I like a lot significantly simplified "interface" without Slice&ProjWData.

Just to be clear: the simplified interface is not new. Plotting the slices as I suggest now has always worked in RooFit, and I'm just changing the tutorials to use simpler patterns :slightly_smiling_face: Why the plot command in the tutorial was so convoluted is not clear to me.

Issue #20619, opened 2025-12-03
fRC increased twice in TGDMLParse


KoljaFrahm:
### Explain what you would like to see improved and how.

One minor thing came up when we saw the code in TGDMLParse:
 https://github.com/root-project/root/blob/ad22457003c8c39069cf2a4d98a0cfbb8cfb431e/geom/gdml/src/TGDMLParse.cxx#L2245
With this code, fRC is increased twice, so the UserExtension isn't deleted when the volume is deleted. This seems unintuitive for me, because if someone would want to use the UserExtension without the volume, they would use GrabUserExtension and therefore increase fRC anyway.
Shouldn't you use Release() once here? Or change SetUserExtension so it doesn't increase fRC?

### ROOT version

all

### Installation method

all

### Operating system

all

ferdymercury:
Maybe relevant:
https://root.cern/doc/master/classTGeoRCExtension.html

```
The following usage is not correct:
some_TGeoVolume->SetUserExtension(new TGeoRCExtension())

since the producer code does not release the extension. One cannot call directly "delete ext" nor allocate an extension on the stack, since the destructor is protected. Use Release instead.
```

PR #20620, opened 2025-12-03
[geom] prevent memleak when SetUserExtension

ferdymercury:
Fixes https://github.com/root-project/root/issues/20619

Follow the docu of SetUserExtension:
`The volume "grabs" a copy, so the original object can be released by the producer. Release the previously connected extension if any.`

Also in the TGeoRCExtension docu:
```
The following usage is not correct:
some_TGeoVolume->SetUserExtension(new TGeoRCExtension()) since the producer code does not release the extension. One cannot call directly "delete ext" nor allocate an extension on the stack, since the destructor is protected. Use Release instead.
```

Found out by @KoljaFrahm


KoljaFrahm:
Thanks, lgtm!

guitargeek (approved):
Thank you!

PR #20622, opened 2025-12-03
makes sure bash executes .bash_profile

couet:
on Mac makes sure bash executes .bash_profile


.github/workflows/root-ci.yml
```diff
@@ -465,7 +465,7 @@ jobs:
         env:
           OVERRIDES: ${{ join( matrix.overrides, ' ') }}
           CONFIGFILE: '.github/workflows/root-ci-config/buildconfig/${{ matrix.image }}.txt'
-        shell: bash
+        shell: bash -leo pipefail {0}
```
dpiparo: Is this really needed? linux builds do work...

couet:
This PR and the other are just tries to understand which shell is executed. It is not meant to be merged yet

PR #20628, opened 2025-12-03
[io] Add ROOT::Internal::DumpBin(TMemFile&) for debugging purposes

silverweed:
# This Pull request:
adds a method to `TMemFile` that allows to dump the current contents of the memory blocks to file. Differently from `Cp`, it does not modify the contents in any way, so it's useful/meant to be used for debugging.

The doc comment clarifies that `Cp` should still be preferred for regular use cases.

### DIFF
--- a/io/io/src/TMemFile.cxx
+++ b/io/io/src/TMemFile.cxx
@@ -313,6 +313,26 @@ void TMemFile::Print(Option_t *option /* = "" */) const
 }
 
 ////////////////////////////////////////////////////////////////////////////////
+/// Writes the contents of the TMemFile to the given file. This is meant to be
+/// used mostly for debugging, as it dumps the current file's content as-is with
+/// no massaging (meaning the content might not be a valid ROOT file): for regular use cases, use Cp().
+/// Example usage:
+/// ~~~ {.cpp}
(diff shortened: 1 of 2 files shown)


io/io/src/TMemFile.cxx
```diff
@@ -312,6 +312,26 @@ void TMemFile::Print(Option_t *option /* = "" */) const
...
 
+////////////////////////////////////////////////////////////////////////////////
+/// Writes the contents of the TMemFile to the given file. This is meant to be
+/// used mostly for debugging, as it dumps the current file's content as-is with
+/// no massaging (meaning the content might not be a valid ROOT file): for regular use cases, use Cp().
+/// Example usage:
+/// ~~~ {.cpp}
+/// FILE *out = fopen("memfile_dump.root", "wb");
+/// ROOT::Internal::DumpBin(memFile, out);
+/// fclose(out);
```
pcanal, line 323: What is the advantage of requiring the 3 steps instead of folding them into the function?

silverweed: Mostly the ability to pass `stdout` or similar, which would be more awkward to do with a string argument.
Also one may potentially want to dump multiple MemFiles to the same output, though that's probably rare.

PR #20629, opened 2025-12-03
[ntuple] fixes to subfield name handling

jblomer:
The following two fixes to the treatment of subfield names:
  - Ensure that subfields of compound types with numbered fields actually carry the field number as a name
  - Prevent duplicate field names in RFieldZero and RRecordField

### DIFF
--- a/tree/ntuple/inc/ROOT/RField.hxx
+++ b/tree/ntuple/inc/ROOT/RField.hxx
@@ -69,6 +69,8 @@ class RFieldZero final : public RFieldBase {
    /// This flag is reset on Clone().
    bool fAllowFieldSubstitutions = false;
 
+   std::unordered_set<std::string> fSubFieldNames; ///< Efficient detection of duplicate field names
+
 protected:
    std::unique_ptr<RFieldBase> CloneImpl(std::string_view newName) const final;
    void ConstructValue(void *) const final {}
@@ -76,7 +78,9 @@ protected:
 public:
    RFieldZero() : RFieldBase("", "", ROOT::ENTupleStructure::kRecord, false /* isSimple */) {}
 
-   using RFieldBase::Attach;
+   /// A public version of the Attach method that allows piece-wise construction of the zero field.
+   /// Will throw on duplicate subfield names.
+   void Attach(std::unique_ptr<RFieldBase> child);
    size_t GetValueSize() const final { return 0; }
    size_t GetAlignment() const final { return 0; }
 
--- a/tree/ntuple/src/RField.cxx
+++ b/tree/ntuple/src/RField.cxx
@@ -38,11 +38,22 @@ void ROOT::Internal::SetAllowFieldSubstitutions(RFieldZero &fieldZero, bool val)
    fieldZero.fAllowFieldSubstitutions = val;
 }
 
+void ROOT::RFieldZero::Attach(std::unique_ptr<RFieldBase> child)
+{
+   const std::string childName = child->GetFieldName();
+   if (fSubFieldNames.count(childName) > 0)
+      throw RException(R__FAIL("duplicate field name: " + childName));
+   RFieldBase::Attach(std::move(child), "");
+   fSubFieldNames.insert(childName);
+}
+
 std::unique_ptr<ROOT::RFieldBase> ROOT::RFieldZero::CloneImpl(std::string_view /*newName*/) const
 {
    auto result = std::make_unique<RFieldZero>();
-   for (auto &f : fSubfields)
+   for (auto &f : fSubfields) {
       result->Attach(f->Clone(f->GetFieldName()));
+      result->fSubFieldNames.insert(f->GetFieldName());
+   }
    return result;
 }
 
@@ -603,7 +614,12 @@ ROOT::RRecordField::RRecordField(std::string_view fieldName, std::vector<std::un
 {
    fTraits |= kTraitTrivialType;
    fOffsets.reserve(itemFields.size());
+   std::unordered_set<std::string_view> fieldNames;
    for (auto &item : itemFields) {
+      const auto &itemName = item->GetFieldName();
+      if (!fieldNames.insert(itemName).second) {
+         throw RException(R__FAIL("duplicate field name: " + itemName));
+      }
       fSize += GetItemPadding(fSize, item->GetAlignment());
       fOffsets.push_back(fSize);
       fMaxAlignment = std::max(fMaxAlignment, item->GetAlignment());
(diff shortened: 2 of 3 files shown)


tree/ntuple/test/ntuple_print.cxx
```diff
@@ -126,7 +126,7 @@ TEST(RNtuplePrint, ArrayAsRVec)
    testField.AcceptVisitor(visitor);
    std::string expected{std::string("") +
                         "$ Field 1       : arrayasrvecfield (ROOT::VecOps::RVec<float>)                 $\n" +
-                        "$   Field 1.1   : myfloat (float)                                              $\n"};
+                        "$   Field 1.1   : _0 (float)                                                   $\n"};
```
pcanal, line 129: This seems to be a degradation.  The user used on line 121 the name `myfloat` which is now no longer represented in the schema ... is that the intent?

hahnjo: Yes, because the binary format specification says that the item field has name `_0`

tree/ntuple/inc/ROOT/RField.hxx
```diff
@@ -69,14 +69,18 @@ class RFieldZero final : public RFieldBase {
...
    std::unique_ptr<RFieldBase> CloneImpl(std::string_view newName) const final;
    void ConstructValue(void *) const final {}
 
 public:
    RFieldZero() : RFieldBase("", "", ROOT::ENTupleStructure::kRecord, false /* isSimple */) {}
 
-   using RFieldBase::Attach;
+   /// A public version of the Attach method that allows piece-wise construction of the zero field.
+   /// Will throw on duplicate subfield names.
+   void Attach(std::unique_ptr<RFieldBase> child);
```
hahnjo, line 83: Hm, `RFieldBase::Attach` is non-`virtual` so "overriding" it in the base class only works if we statically know the type of the `RFieldZero` - is that the case?

tree/ntuple/src/RField.cxx
```diff
@@ -38,11 +38,22 @@ void ROOT::Internal::SetAllowFieldSubstitutions(RFieldZero &fieldZero, bool val)
    fieldZero.fAllowFieldSubstitutions = val;
 }
 
+void ROOT::RFieldZero::Attach(std::unique_ptr<RFieldBase> child)
+{
+   const std::string childName = child->GetFieldName();
```
hahnjo, line 43:
```suggestion
   const std::string &childName = child->GetFieldName();
```
to avoid the copy?

tree/ntuple/src/RField.cxx
```diff
@@ -38,11 +38,22 @@ void ROOT::Internal::SetAllowFieldSubstitutions(RFieldZero &fieldZero, bool val)
...
+   if (fSubFieldNames.count(childName) > 0)
+      throw RException(R__FAIL("duplicate field name: " + childName));
+   RFieldBase::Attach(std::move(child), "");
+   fSubFieldNames.insert(childName);
```
hahnjo, line 47: Can do this in one go:
```suggestion
   if (!fSubFieldNames.insert(childName).second)
      throw RException(R__FAIL("duplicate field name: " + childName));
   RFieldBase::Attach(std::move(child), "");
```

tree/ntuple/src/RField.cxx
```diff
@@ -602,13 +614,19 @@ ROOT::RRecordField::RRecordField(std::string_view fieldName, std::vector<std::un
 {
    fTraits |= kTraitTrivialType;
    fOffsets.reserve(itemFields.size());
+   std::unordered_set<std::string> fieldNames;
```
hahnjo: For the local `std::unordered_set`, we probably only need `std::string_view` as the item type and avoid the string copies

tree/ntuple/src/RField.cxx
```diff
@@ -602,13 +614,19 @@ ROOT::RRecordField::RRecordField(std::string_view fieldName, std::vector<std::un
...
    for (auto &item : itemFields) {
+      const auto itemName = item->GetFieldName();
```
hahnjo: This will copy the field name

tree/ntuple/src/RField.cxx
```diff
@@ -602,13 +614,19 @@ ROOT::RRecordField::RRecordField(std::string_view fieldName, std::vector<std::un
...
+      if (fieldNames.count(itemName) > 0) {
+         throw RException(R__FAIL("duplicate field name: " + itemName));
+      }
```
hahnjo:
```suggestion
      if (!fieldNames.insert(itemName).second) {
         throw RException(R__FAIL("duplicate field name: " + itemName));
      }
```

tree/ntuple/test/ntuple_print.cxx
pcanal: Humm .... when reading back how will the user properly retrieve `myfloat` vs `myotherfloat` if they are both turned into `_0` and the name are completely forgotten?
i.e. Is the name really gone (likely a blocker of sort) or is it just an artefact of the printing?

jblomer: The name is really gone. It is not foreseen to have a custom field name for the collection item. It should be quite uncommon to assemble collections manually (and not through a collection type). We could warn of fail if we detect a name mismatch.

tree/ntuple/inc/ROOT/RField.hxx
jblomer: I think we are ok: the `Attach()` method is protected in the base class, so we anyway need to know that we have an `RFieldZero` to invoke this version of `Attach()` outside the RField class hierarchy.

tree/ntuple/src/RField.cxx
jblomer: I think we need the copy, because we move `child` and only then insert the name into `fSubFieldNames`. We could swap these two operations but then it gets more complicated if `RFieldBase::Attach()` throws an exception...

tree/ntuple/src/RField.cxx
jblomer: Same here, `RFieldBase::Attach()` may throw.

tree/ntuple/src/RField.cxx
jblomer: Fixed

tree/ntuple/test/ntuple_print.cxx
jblomer: As discussed, the code now checks and throws if the subfield has a wrong name.

PR #20630, opened 2025-12-03
[roottest] Remove unreachable or redundant code from CMakeLists.txt

guitargeek:
Some configuration code branches are unreachable, now that `roottest` has been merged into ROOT.

This also applies to the code that finds the Python interpreter and libraries, as Python is unconditionally found as a dependency already in the main `SearchIntalledSoftware.cmake`, even if `pyroot=OFF`.

Another example of redundant code was re-setting the compression default.

Specifying a separate project name and minimum CMake version is also not necessary anymore.

### DIFF
--- a/roottest/CMakeLists.txt
+++ b/roottest/CMakeLists.txt
@@ -1,14 +1,3 @@
-# Use ${ROOTSYS} CMake variable to search for ROOT. This is set when roottest is
-# built as part of ROOT, i.e., when ROOT is configured with -Droottest=ON. It
-# may also be set by passing -DROOTSYS=<PATH> to CMake when configuring a
-# standalone roottest against an installation of ROOT. If this is set, it takes
-# precedence over the environment to ensure that when roottest is being built in
-# a ROOT build configured with -Droottest=ON, it will find the ROOT being built,
-# and not a ROOT installation sourced with thisroot.sh or found in $PATH.
-if(DEFINED ROOTSYS)
-  list(INSERT CMAKE_PREFIX_PATH 0 ${ROOTSYS})
-endif()
-
 find_package(ROOT REQUIRED CONFIG)
 
 if(MSVC)


roottest/CMakeLists.txt
```diff
@@ -1,14 +1,3 @@
...
-# If no explicit ROOTSYS is set, use ${ROOTSYS} environment variable to search
-# for ROOT. This may be set by sourcing thisroot.sh from an installation.
-if(NOT DEFINED ROOTSYS AND DEFINED ENV{ROOTSYS})
-  cmake_path(CONVERT "$ENV{ROOTSYS}" TO_CMAKE_PATH_LIST ROOTSYS)
-  list(INSERT CMAKE_PREFIX_PATH 0 ${ROOTSYS})
-endif()
-
 # Use ${ROOTSYS} CMake variable to search for ROOT. This is set when roottest is
 # built as part of ROOT, i.e., when ROOT is configured with -Droottest=ON. It
 # may also be set by passing -DROOTSYS=<PATH> to CMake when configuring a
```
ferdymercury: This paragraph should also be updated/removed

roottest/CMakeLists.txt
```diff
@@ -20,12 +9,6 @@ if(DEFINED ROOTSYS)
   list(INSERT CMAKE_PREFIX_PATH 0 ${ROOTSYS})
```
ferdymercury: is it needed?

guitargeek: I don't know, maybe not. I'd prefer to only remove things in this commit where I understand why, and the rest can be done some other time.

roottest/CMakeLists.txt
guitargeek: Done

roottest/CMakeLists.txt
guitargeek: Actually I didn't see why it was needed. Removed in one additional commit in this PR.

PR #20633, opened 2025-12-03
[ntuple] add support for leaf count arrays in classes

jblomer:
Transparently translates a leaf count array in a custom class in an RVec (read & write). Classes with leaf count arrays are used, e.g., in the ALICE RAW EDM but most likely there are other relevant cases, too.

EDIT: also needed for LHCb DST files.

Since this approach diverges the in-memory representation from the on-disk representation, we should discuss if we actually want this.

In any case, we probably want the first commit that detects leaf count arrays in classes and fails gracefully in such a case.

### DIFF
--- a/tree/ntuple/inc/ROOT/RField/RFieldSequenceContainer.hxx
+++ b/tree/ntuple/inc/ROOT/RField/RFieldSequenceContainer.hxx
@@ -417,6 +417,51 @@ public:
    void AcceptVisitor(ROOT::Detail::RFieldVisitor &visitor) const final;
 };
 
+/**
+\class ROOT::RLeafCountArrayField
+\brief A field for an RVec field as a class member on disk that is represented as leaf count array` in memory.
+\ingroup NTuple
+*/
+class RLeafCountArrayField final : public RFieldBase {
+   friend class RClassField; // only class fields are able to construct this
+
+private:
+   std::size_t fItemSize{0};                  ///< The size of a child field's item
+   ROOT::Internal::RColumnIndex fNWritten{0}; ///< Number of items written in the current cluster
+   std::ptrdiff_t fCountLeafDelta{0};         ///< Offset delta of the count leaf int member relative to the array
+   bool fHasPersistentCountLeaf{true};        ///< The count leaf may be only transient in RNTuple
+
+   RLeafCountArrayField(std::string_view fieldName, std::unique_ptr<RFieldBase> itemField,
+                        std::ptrdiff_t countLeafDelta, bool hasPersistentCountLeaf);
+
+protected:
+   std::unique_ptr<RFieldBase> CloneImpl(std::string_view newName) const final;
+
+   const RColumnRepresentations &GetColumnRepresentations() const final;
+   void GenerateColumns() final;
+   void GenerateColumns(const ROOT::RNTupleDescriptor &desc) final;
+
+   void ConstructValue(void *where) const final { *reinterpret_cast<void **>(where) = nullptr; }
+
+   std::size_t AppendImpl(const void *from) final;
+   void ReadGlobalImpl(ROOT::NTupleSize_t globalIndex, void *to) final;
+
+   void ReconcileOnDiskField(const RNTupleDescriptor &desc) final;
+
+   void CommitClusterImpl() final { fNWritten = 0; }
+
+public:
+   RLeafCountArrayField(RLeafCountArrayField &&other) = default;
+   RLeafCountArrayField &operator=(RLeafCountArrayField &&other) = default;
+   ~RLeafCountArrayField() final = default;
+
+   std::vector<RValue> SplitValue(const RValue &value) const final;
+
+   size_t GetValueSize() const final { return sizeof(void *); }
+   size_t GetAlignment() const final { return std::alignment_of<void *>(); }
+   void AcceptVisitor(ROOT::Detail::RFieldVisitor &visitor) const final;
+};
+
 } // namespace ROOT
 
 #endif
--- a/tree/ntuple/src/RFieldMeta.cxx
+++ b/tree/ntuple/src/RFieldMeta.cxx
@@ -239,12 +239,19 @@ ROOT::RClassField::RClassField(std::string_view fieldName, TClass *classp)
          }
       }
 
-      if (IsLeafCountArray(*dataMember)) {
-         throw RException(R__FAIL(std::string("leaf count arrays are currently unsupported: ") + GetTypeName() + "." +
-                                  dataMember->GetName()));
-      }
+      const auto memberName = dataMember->GetName();
+      std::unique_ptr<RFieldBase> subField;
 
-      auto subField = RFieldBase::Create(dataMember->GetName(), typeName).Unwrap();
+      if (auto realMember = IsLeafCountArray(*dataMember)) {
+         assert(typeName.length() > 0);
+         assert(*typeName.rbegin() == '*');
+         auto itemField = RFieldBase::Create("_0", typeName.substr(0, typeName.length() - 1)).Unwrap();
+         const std::ptrdiff_t offsetDelta = realMember->GetThisOffset() - dataMember->GetOffset();
+         subField = std::unique_ptr<RLeafCountArrayField>(new RLeafCountArrayField(
+            memberName, std::move(itemField), offsetDelta, true /* fHasPersistentCountLeaf */));
+      } else {
+         subField = RFieldBase::Create(memberName, typeName).Unwrap();
+      }
 
       const auto normTypeName = ROOT::Internal::GetNormalizedUnresolvedTypeName(origTypeName);
       if (normTypeName == subField->GetTypeName()) {
--- a/tree/ntuple/src/RFieldMeta.cxx
+++ b/tree/ntuple/src/RFieldMeta.cxx
@@ -239,6 +239,11 @@ ROOT::RClassField::RClassField(std::string_view fieldName, TClass *classp)
          }
       }
 
+      if (IsLeafCountArray(*dataMember)) {
+         throw RException(R__FAIL(std::string("leaf count arrays are currently unsupported: ") + GetTypeName() + "." +
+                                  dataMember->GetName()));
+      }
+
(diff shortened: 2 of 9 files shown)


tree/ntuple/inc/ROOT/RField/RFieldSequenceContainer.hxx
```diff
@@ -417,6 +417,51 @@ public:
    void AcceptVisitor(ROOT::Detail::RFieldVisitor &visitor) const final;
 };
 
+/**
+\class ROOT::RLeafCountArrayField
+\brief A field for an RVec field as a class member on disk that is represented as leaf count array` in memory.
+\ingroup NTuple
+*/
+class RLeafCountArrayField final : public RFieldBase {
```
hahnjo, line 425: The commit message says it's `Internal`, but the declaration doesn't seem to be?

tree/ntuple/test/CustomStruct.hxx
```diff
@@ -460,4 +460,28 @@ struct ExampleMC {
 };
 } // namespace v2
 
+struct LeafCountInClassBase {
+   Int_t fSize = 0;
+};
+
+struct LeafCountInClass : public LeafCountInClassBase {
+   unsigned char *fPayload1 = nullptr; //[fSize];
+   unsigned char *fPayload2 = nullptr; //[fSize];
```
vepadulano, line 469: I'm probably missing the context behind this PR, but looking at this class makes me wonder if we're subtly introducing support for raw pointers in RNTuple on-disk format.

silverweed: One could "emulate" a raw pointer by using `fSize = 1`, but since this only works for owned payloads it's not conceptually different from supporting `unique_ptr` (I believe the problematic part of raw pointers is the fact that they may not be owning and reference data outside the parent class)

jblomer: Also note that on-disk this is stored as an `RVec<char>`. This PR adds support for a peculiar memory layout of a vector, if you will.

tree/ntuple/inc/ROOT/RField/RFieldSequenceContainer.hxx
```diff
@@ -417,6 +417,51 @@ public:
...
+   friend class RClassField; // only class fields are able to construct this
+
+private:
+   std::size_t fItemSize{0};                  ///< The size of a child field's item
```
pcanal, line 429: For better or worse, the leaf count is currently limited (hard coded) to 32bits.

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -239,6 +251,43 @@ void ROOT::RClassField::Attach(std::unique_ptr<RFieldBase> child, RSubFieldInfo
...
+
+   std::string title = dm.GetTitle();
+   if (title.length() == 0 || title[0] != '[')
+      return nullptr;
+
+   auto idxRight = title.find_first_of("]", 1);
+   if (idxRight == std::string::npos)
+      return nullptr;
+
+   auto countLeaf = title.substr(1, idxRight - 1);
```
pcanal:
```suggestion
   auto nameCountLeaf = title.substr(1, idxRight - 1);
```
(and subsequent change)

pcanal (commented):
LGTM code wise.  Inclusion shall be discussed at next ROOT I/O workshop.

tree/ntuple/inc/ROOT/RField/RFieldSequenceContainer.hxx
jblomer: This member does not store the count (size) of the collection but the size of the collection item (as in `sizeof(T)` for `std::vector<T>`).

tree/ntuple/src/RFieldSequenceContainer.cxx
```diff
@@ -1051,3 +1051,129 @@ void ROOT::RArrayAsVectorField::AcceptVisitor(ROOT::Detail::RFieldVisitor &visit
...
+
+   if (fHasPersistentCountLeaf) {
+      if (nItems != *countPtr) {
+         throw RException(R__FAIL("count leaf value different from collection size on disk"));
+      }
+   } else {
+      *countPtr = nItems;
+   }
+
+   operator delete(*typedValue);
```
hahnjo, line 1143: I think this only works for trivially-destructible types. If we (want to) support non-simple types, we will probably need to go via the item deleter...

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -239,7 +239,19 @@ ROOT::RClassField::RClassField(std::string_view fieldName, TClass *classp)
...
+      const auto memberName = dataMember->GetName();
+      std::unique_ptr<RFieldBase> subField;
+
+      if (auto realMember = IsLeafCountArray(*dataMember)) {
+         assert(typeName.length() > 0);
+         assert(*typeName.rbegin() == '*');
+         auto itemField = RFieldBase::Create("_0", typeName.substr(0, typeName.length() - 1)).Unwrap();
+         const std::ptrdiff_t offsetDelta = realMember->GetThisOffset() - dataMember->GetOffset();
+         subField = std::unique_ptr<RLeafCountArrayField>(new RLeafCountArrayField(
+            memberName, std::move(itemField), offsetDelta, true /* fHasPersistentCountLeaf */));
```
hahnjo, line 251: Is this missing support for detecting if `realMember` is transient?

tree/ntuple/src/RFieldSequenceContainer.cxx
hahnjo: Never mind, I already forgot that the constructor of `RLeafCountArrayField` checks that the item type has `kTraitTriviallyDestructible`...

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -273,6 +285,43 @@ void ROOT::RClassField::Attach(std::unique_ptr<RFieldBase> child, RSubFieldInfo
...
+
+   std::string title = dm.GetTitle();
+   if (title.length() == 0 || title[0] != '[')
+      return nullptr;
+
+   auto idxRight = title.find_first_of("]", 1);
+   if (idxRight == std::string::npos)
+      return nullptr;
+
+   auto nameCountLeaf = title.substr(1, idxRight - 1);
```
pcanal, line 301: This is similar to
```
std::string nameCountLeaf = dm.GetArrayIndex();
if (nameCountLeaf.empty())
	return nullptr;
```

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -273,6 +285,43 @@ void ROOT::RClassField::Attach(std::unique_ptr<RFieldBase> child, RSubFieldInfo
...
+   for (auto realMember : ROOT::Detail::TRangeStaticCast<TRealData>(*fClass->GetListOfRealData())) {
```
pcanal, line 302: I think one can also replace the loop with:
```
auto realMember = (TRealData*) fClass->GetListOfRealData()->FindObject(nameCountLeaf.c_str());
```

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -273,6 +285,43 @@ void ROOT::RClassField::Attach(std::unique_ptr<RFieldBase> child, RSubFieldInfo
...
+      if (nameCountLeaf != realMember->GetName())
+         continue;
+
+      const auto dmCountLeaf = realMember->GetDataMember();
+
+      const auto dtCountLeaf = dmCountLeaf->GetDataType();
+      if (!dtCountLeaf || ((dtCountLeaf->GetType() != kInt_t) && (dtCountLeaf->GetType() != kUInt_t))) {
+         throw ROOT::RException(
+            R__FAIL(std::string("invalid count leaf type: ") + GetTypeName() + "." + realMember->GetName()));
```
pcanal, line 311:
```suggestion
            R__FAIL(std::string("Unsupported count leaf type: ") + GetTypeName() + "." + realMember->GetName()));
```
may be slightly more accurate.

tree/ntuple/src/RFieldMeta.cxx
```diff
@@ -273,6 +285,43 @@ void ROOT::RClassField::Attach(std::unique_ptr<RFieldBase> child, RSubFieldInfo
    RFieldBase::Attach(std::move(child));
 }
 
+TRealData *ROOT::RClassField::IsLeafCountArray(const TDataMember &dm) const
```
pcanal, line 288: `IsLeafCountArray` is usually returning a `bool`.  Did you consider `GetLeafCountArrayIndex`? ( which is admittedly differently odd as it is actually `GetLeafCountArrayIndexIfAny`)

tree/ntuple/src/RFieldSequenceContainer.cxx
```diff
@@ -1051,3 +1051,129 @@ void ROOT::RArrayAsVectorField::AcceptVisitor(ROOT::Detail::RFieldVisitor &visit
...
+     fCountLeafDelta(countLeafDelta),
+     fHasPersistentCountLeaf(hasPersistentCountLeaf)
+{
+   static_assert(sizeof(int) == sizeof(std::int32_t)); // implicit assumption in this field
+   assert(fCountLeafDelta < 0);                        // the count leaf must be defined before the array
+
+   if (!(itemField->GetTraits() & kTraitTriviallyDestructible)) {
+      throw RException(R__FAIL("RLeafCountArrayField only supports trivially destructible item types"));
+   }
+   // Note that we expect the array pointer to be deleted by the user.
```
pcanal, line 1071: Do we document somewhere why it is useful to have this limitation?

tree/ntuple/test/CustomStruct.hxx
```diff
@@ -460,4 +460,28 @@ struct ExampleMC {
...
+};
```
pcanal, line 470: Those would usually have a custom destructor to delete the payload memory.

tree/ntuple/src/RFieldMeta.cxx
silverweed: Maybe `TryGetLeafCountArrayIndex`?

PR #20640, opened 2025-12-04
[CMake] Don't use find_package(ROOT REQUIRED) in `test/` subdirectory

guitargeek:
As Sergey said in the comment in `test/CMakeLists.txt`, this test
project stopped working standalone. But, it still uses
`find_package(ROOT)` to find ROOT, even though it's not installed yet as
we are building ROOT itself at the same time.

Apparently that works, and sets the flags like `ROOT_opengl_FOUND`, but
we should not rely on this and use the build-time configuration flags
instead.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -3,40 +3,11 @@
 #
 # For the licensing terms see $ROOTSYS/LICENSE.
 # For the list of contributors see $ROOTSYS/README/CREDITS.
-
-# CMakeLists.txt for the ROOT test programs.
-# This shows nicely how to compile and link applications
-# using the ROOT libraries on all supported platforms.
 #
 # Author: Pere Mato, 25/10/2010
-cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
 
-project(test)
-
-# Sergey: make no sense while CMakeLists.txt file cannot be used separately from ROOT
-# but variables like ROOT_asimage_FOUND used here and produced in ROOTConfig.cmake
-find_package(ROOT REQUIRED)
-
-#---Copy the CTestCustom.cmake file into the build directory---------------------------------
-configure_file(${CMAKE_CURRENT_SOURCE_DIR}/CTestCustom.cmake ${CMAKE_CURRENT_BINARY_DIR} COPYONLY)
-enable_testing()
-
-if(CMAKE_SYSTEM_NAME MATCHES Darwin)
-  # To avoid to set ld_Library_path to locate the test libraries
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath,@loader_path/.")
-endif()
-
-if(DEFINED ROOT_SOURCE_DIR)
-  # Testing using the binary tree
-  set(ROOT_root_CMD ${ROOT_BINDIR}/root.exe)
-  include_directories(${ROOT_SOURCE_DIR}/tutorials)
-else()
-  # Testing using an installation (assuming access to ROOT CMake modules)
-  include_directories(${ROOT_INCLUDE_DIRS})
-  include_directories(${ROOT_INCLUDE_DIRS}/../tutorials)
-  add_definitions(${ROOT_DEFINITIONS})
-  include(RootMacros)
-endif()
+# Testing using the binary tree
+set(root_exe $<TARGET_FILE:root.exe>)
 
 #---environment-------------------------------------------------------------------------------
 ROOT_ADD_TEST(show-environment COMMAND ${CMAKE_COMMAND} -E environment)
@@ -79,12 +50,12 @@ endif()
 #---invalid pointer checks---------------------------------------------------------------------
 if((NOT (APPLE AND CMAKE_SYSTEM_PROCESSOR MATCHES arm64) OR M1_BROKEN_TESTS) AND NOT experimental_adaptivecpp)
   ROOT_ADD_TEST(test-check-nullptr
-    COMMAND ${ROOT_root_CMD} -b -q -l --ptrcheck -e "int*ptr=(int*)0" -e "*ptr"
+    COMMAND ${root_exe} -b -q -l --ptrcheck -e "int*ptr=(int*)0" -e "*ptr"
     PASSREGEX "dereference null pointer")
 
   if(NOT MSVC OR win_broken_tests)
     ROOT_ADD_TEST(test-check-invalidptr
-      COMMAND ${ROOT_root_CMD} -b -q -l --ptrcheck -e "int*ptr=(int*)1" -e "*ptr"
+      COMMAND ${root_exe} -b -q -l --ptrcheck -e "int*ptr=(int*)1" -e "*ptr"
       PASSREGEX "invalid memory address")
   endif()
 endif()
@@ -95,7 +66,6 @@ ROOT_ADD_TEST(test-minexam COMMAND minexam)
 
 #--tstring-------------------------------------------------------------------------------------
 ROOT_EXECUTABLE(tstring tstring.cxx LIBRARIES Core)
-configure_file(tstring.cxx tstring.cxx COPYONLY)
 ROOT_ADD_TEST(test-tstring COMMAND tstring)
 
 #--tcollex-------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


cmake/modules/RootNewMacros.cmake
vepadulano, line 1: Is it necessary to remove this file? Could it be possible that someone in the world is still using it? Should we perhaps change the deprecation message to specify a ROOT release when this will be removed?

guitargeek: Not strictly necessary, but it's a bit confusing to keep as it's called `RootNewMacros.cmake`, with "New" in the name, but deprecated and replaced by `RootMacros.cmake` without "New". I would like to keep confusing situations like this behind for good.

Users will have seen the deprecation message for 6 years now. I think people will understand if we remove it now.

guitargeek: I you think this is controversial, I can bring it forward in a separate PR!

bellenot: Geant4 still uses it: https://github.com/Geant4/geant4/blob/master/examples/extended/persistency/P01/CMakeLists.txt#L39
https://github.com/Geant4/geant4/blob/master/examples/advanced/CaTS/CMakeLists.txt#L53

guitargeek: ...only if it exists :slightly_smiling_face: Otherwise, they automatically get the new `RootMacros.cmake` via the `find_package(ROOT REQUIRED)` before in the file. So they actually to the right thing, to even be backwards compatible with ROOT before 6.20.

bellenot: Oh right, sorry, I overlooked at it...

test/CMakeLists.txt
```diff
@@ -3,40 +3,11 @@
 #
 # For the licensing terms see $ROOTSYS/LICENSE.
 # For the list of contributors see $ROOTSYS/README/CREDITS.
-
-# CMakeLists.txt for the ROOT test programs.
-# This shows nicely how to compile and link applications
-# using the ROOT libraries on all supported platforms.
```
pcanal, line 9: Can we update the PR description and commit log on where we now document and/or demonstrate the modern way to build you application that uses ROOT?

test/CMakeLists.txt
```diff
@@ -252,37 +222,40 @@ if(MSVC AND NOT CMAKE_GENERATOR MATCHES Ninja)
...
+ROOT_ADD_TEST(test-stressfit-interpreted COMMAND ${root_exe} -b -q -l ${CMAKE_CURRENT_SOURCE_DIR}/stressFit.cxx
               FAILREGEX "FAILED|Error in" DEPENDS test-stressfit)
 
 #--stressHistoFit-----------------------------------------------------------------------------
-if(ROOT_unuran_FOUND)
-  ROOT_EXECUTABLE(stressHistoFit stressHistoFit.cxx LIBRARIES MathCore Matrix Unuran Tree Gpad)
+if(unuran)
+  if(mathmore)
+    list(APPEND stress_hist_fit_extra_libs MathMore)
+  endif()
```
pcanal: Why is this now needed?

test/CMakeLists.txt
```diff
@@ -129,7 +99,7 @@ ROOT_LINKER_LIBRARY(Aclock Aclock.cxx G__Aclock.cxx LIBRARIES Graf Gpad)
 #--bench------------------------------------------------------------------------------------
 ROOT_GENERATE_DICTIONARY(G__TBench ${CMAKE_CURRENT_SOURCE_DIR}/TBench.h MODULE TBench LINKDEF benchLinkDef.h DEPENDENCIES MathCore Tree)
 ROOT_LINKER_LIBRARY(TBench TBench.cxx G__TBench.cxx LIBRARIES Core MathCore RIO Tree)
-ROOT_EXECUTABLE(bench bench.cxx LIBRARIES Core TBench)
+ROOT_EXECUTABLE(bench bench.cxx LIBRARIES Core TBench RIO)
```
pcanal, line 102: Why is `RIO` needed/wanted when it wasn't before?

test/CMakeLists.txt
guitargeek: Well, anywhere where we use `ROOT_EXECUTABLE` or `ROOT_STANDARD_LIBRARY_PACKAGE` I guess? I don't see what made this `test/CMakeLists.txt`  a good place "to show nicely how to compile and link applications using the ROOT libraries on all supported platforms". That comment seemed pretty random to me.

test/CMakeLists.txt
guitargeek: Before, we implicitly linked all main ROOT libraries via `find_package(ROOT REQUIRED)`. But I want to avoid calling that when building ROOT itself, as that seems pretty weird to me. So now, we have to link all required libraries explicitly.

test/CMakeLists.txt
guitargeek: Because of this conditional include here:
https://github.com/root-project/root/blob/master/math/mathcore/inc/Math/DistFunc.h#L39

But let me try to replace this include

test/CMakeLists.txt
pcanal: Rather than random, it is historical.  Once upon a time,  this was a tidy smallish standalone (Makefile then CMake) example that build dictionaries, libraries and executables that use ROOT.   I can see that over time we drifted away from that (seemingly breaking the standalone part and getting a bit complex).   However unless (but I assume we actually do have one) we have a replacement documentation, then this would be the closest we have to a documentation.  (If we don't have the documentation and nice examples, then this PR can still proceed but we need to open an issue to remind us to write that documentation).

> Well, anywhere where we use ROOT_EXECUTABLE or ROOT_STANDARD_LIBRARY_PACKAGE

Humm .. that does not sound enough, it also needs at least the call to `find_package(ROOT)` and a good set of usage of the CMake function we recommend using. i.e. enough for a `CMake` beginner to get started.

test/CMakeLists.txt
pcanal: That include was already there before this PR, so why don't we need the dependency before ...

test/CMakeLists.txt
pcanal: >  But I want to avoid calling that when building ROOT itself

+1

> we implicitly linked all main ROOT libraries via find_package(ROOT REQUIRED)

humm ... what? ... If I read that correctly (probably didn't), this would be mean that for any user of `find_package(ROOT)` "all" their libraries would linked against `libCore` and friends?   i.e. Can you remind what is the side effect of `find_package` that make the dependency on RIO (and apparently `mathMore`) optional/redundant?

guitargeek: You are right, that sounds like it can't be true. I looked into it again, and I think I found the real reason:
https://github.com/root-project/root/blob/master/cmake/modules/RootMacros.cmake#L1420

```cmake
  if(NOT (PROJECT_NAME STREQUAL "ROOT"))
    # only for non-ROOT executable use $ROOTSYS/include
    include_directories(BEFORE ${CMAKE_BINARY_DIR}/include)
```
So it was the removal of `project(test)` in this PR that resulted `${CMAKE_BINARY_DIR}/include` in not being appended anymore, and this is why the include directories need to be explicitly added per package via the CMake `target_link_library` mechanism.

test/CMakeLists.txt
guitargeek: Same reason as https://github.com/root-project/root/pull/20640#discussion_r2591879379

test/CMakeLists.txt
guitargeek: Okay, I see now! The entry-point documentation for using ROOT via CMake is here now:
https://root.cern/manual/integrate_root_into_my_cmake_project/#full-example-event-project

Do you see anything missing from there? I think it's good that we advocate the custom macros like `ROOT_EXECTUABLE` too much, and instead show how to use standard CMake, just like the example on the website does.

guitargeek:
@pcanal, I addressed all your inline comments

test/CMakeLists.txt
pcanal: That looks fine enough (although it 'requires' the information provide by Henry Schreiner.  It is linked from our documentation but we depends on that linked page staying up.

test/CMakeLists.txt
pcanal: Fair enough.  Thanks for looking into this and for the clarification.

test/CMakeLists.txt
guitargeek: Ah good that you point that out, I didn't notice! We should revamp this documentation at some point.

If I remember correctly, one of the conclusion from the ROOT users workshop was that we should have example of what makes a good package in the ROOT ecosystem, so good CMake code should be part of it. @dpiparo

PR #20646, opened 2025-12-04
[roottest][io] Bring back `__CLING__` guard in Hard2Stream members

guitargeek:
Follows up on 5727e1d6a65, bringing back a `__CINT__` preprocessor guard that should have stayed and rename it to `__CLING__`.

Address https://github.com/root-project/root/pull/20613/files#r2590556753

### DIFF
--- a/roottest/root/io/customStreamer/header.h
+++ b/roottest/root/io/customStreamer/header.h
@@ -2,7 +2,9 @@
 #include "TClass.h"
 class Hard2Stream {
 private:
+#ifndef __CLING__
    double val;
+#endif
 public:
    Hard2Stream() : val(-1) {};
    Hard2Stream(double v) : val(v) {};


pcanal (approved):

pcanal:
Apparently we are missing something :(

guitargeek:
Seems like it! You want to bring back the `__CINT__` (and a comment) for now and try to understand the problem later? 

But isn't `__CINT__` a no-op now anyway? I don't understand what is going on, or why exactly the macro is required here :slightly_frowning_face:

pcanal:
> Seems like it! You want to bring back the __CINT__ (and a comment) for now and try to understand the problem later?

Nah ... but it looks like it was indeed no longer doing anything for that test and that the suggestion I made is the wrong thing (does not work) to accomplish the intent (ensuring that the I/O cannot/will not use the interpreter information for streaming - i.e. we don't want the test to succeed even though the Streamer function was not used).

So it looks like I need to find a better semantic replacement (and/or maybe I am lucky there is already part of the current test that does it some other way).

guitargeek:
Ok! I leave it to you then :slightly_smiling_face:

PR #20648, opened 2025-12-05
[CI] Remove the homebrew mac update job from pull requests.

hageboeck:
Due to on outdated "on:" clause, the update job ran as part of every pull request on master.


guitargeek (approved):
Thanks!

PR #20658, opened 2025-12-06
[cppyy] Remove code related to finding CPyCppyy API header path

guitargeek:
This is not needed for ROOT, because the CPyCppyy API headers are installed with the other ROOT headers anyway, and ROOT knows where to find them already.


vepadulano (approved):
Not against the change per se, but is it convenient to increase divergence with standalone cppyy for this?

guitargeek:
I don't think worries about code divergence should be a concern at this point. The development of Wims version is slow, and we can easily sync with it by cherry picking his few commits if we want to.

In addition, it will have completely diverged early next year with the rebase on CppInterOp anyway, so we might as well start tolerating increasing divergence already now.

PR #20659, opened 2025-12-07
[CMake] Fix vetoes for tutorials if numpy or xrootd are not available

guitargeek:
Follows up on https://github.com/root-project/root/commit/45bef341f2058c5a3ff605d56d8181fcdd50964e, adding a veto if NumPy is not found. Also, move vetoes for tutorials when xrootd is not available to the right place.

### DIFF
--- a/tutorials/CMakeLists.txt
+++ b/tutorials/CMakeLists.txt
@@ -216,6 +216,7 @@ if(NOT xrootd)
                   roofit/roofit/rf618_mixture_models.py # depends on df106_HiggsToFourLeptons.py
                   visualisation/rcanvas/df104.py
                   visualisation/rcanvas/df105.py
+                  io/tree/imt_parTreeProcessing.C
                   )
 endif()
 
@@ -281,8 +282,6 @@ endif()
 if(MSVC)
   #---Multiproc is not supported on Windows
   set(imt_veto ${imt_veto} analysis/parallel/mp*.C io/tree/mp*.C legacy/multicore/mp*.C  multicore/mp*.C ./analysis/parallel/mtbb_parallelHistoFill.C)
-  #---XRootD is not supported on Windows
-  set(imt_veto ${imt_veto} io/tree/imt_parTreeProcessing.C)
 endif()
 
 if(ROOT_CLASSIC_BUILD)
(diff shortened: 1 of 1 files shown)


tutorials/CMakeLists.txt
```diff
@@ -281,8 +282,6 @@ endif()
 if(MSVC)
   #---Multiproc is not supported on Windows
   set(imt_veto ${imt_veto} analysis/parallel/mp*.C io/tree/mp*.C legacy/multicore/mp*.C  multicore/mp*.C ./analysis/parallel/mtbb_parallelHistoFill.C)
-  #---XRootD is not supported on Windows
-  set(imt_veto ${imt_veto} io/tree/imt_parTreeProcessing.C)
```
vepadulano, line 285: Why was this removed? Is the absence of xrootd support for Windows handled elsewhere?

tutorials/CMakeLists.txt
```diff
@@ -589,17 +588,6 @@ set(math-foam-foam_demopers-depends tutorial-math-foam-foam_demo)
...
-                       tutorial-math-fit-fit1
-                       tutorial-math-fit-myfit
-                       tutorial-hist-hist015_TH1_read_and_draw
-                       tutorial-hist-hist056_TPolyMarker_contour
-                       tutorial-legacy-g3d-na49view
-                       tutorial-io-tree-tree120_ntuple.C
-                       tutorial-io-tree-spider
-                       tutorial-io-hadd
-                       tutorial-io-loopdir
-                       tutorial-io-copyFiles)
```
vepadulano, line 602: Why were these lines removed? Similarly for `pyroot-benchmarks-depends`

vepadulano (changes requested):

tutorials/CMakeLists.txt
vepadulano: Ah yes, I understand, on Windows we also fall in the `NOT xrootd` case I guess

tutorials/CMakeLists.txt
guitargeek: I thought this dependency was not relevant anymore, but now I see that this was only because the `benchmarks` tutorial is always disabled:
https://github.com/root-project/root/blob/master/tutorials/CMakeLists.txt#L446

I'll bring this `benchmarks-depends` variable back for now.

vepadulano (approved):
LGTM thanks! Please merge once CI is green.

tutorials/CMakeLists.txt
guitargeek: Correct!

hahnjo:
I think this accidentally broke testing of the Python wheels (that were probably relying on successfully exiting even if the input file is missing, which btw was not mentioned anywhere in the PR!): https://github.com/root-project/root/actions/runs/20119030998

FYI @siliataider

vepadulano:
@hahnjo Thanks for checking so quickly! I am glad to see the action is working as intended!

siliataider:
> that were probably relying on successfully exiting even if the input file is missing, which btw was not mentioned anywhere in the PR!

Well the wheels tests were not relying on successfully exiting even if the input file is missing, they were relying on successfully exiting if no error is raised, and since the tutorial was detecting a missing file and still exiting with 0, the wheels tests cannot detect that a file is missing in this case and mark it failed
also thanks for catching this quickly!

PR #20664, opened 2025-12-08
[df] First integration of `RHist` filling

hahnjo:
As the classes are experimental, using the method will print a warning to inform the user.

Note that this is only a first version, there are a couple of advanced functionalities and better error handling that will come in future PRs.

### DIFF
--- a/tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
@@ -36,6 +36,7 @@
 #ifdef R__HAS_ROOT7
 #include <ROOT/RHist.hxx>
 #include <ROOT/RHistConcurrentFiller.hxx>
+#include <ROOT/RWeight.hxx>
 #endif
 
 #include <algorithm>
@@ -46,6 +47,7 @@
 #include <stdexcept>
 #include <string>
 #include <string_view>
+#include <tuple>
 #include <type_traits>
 #include <utility> // std::index_sequence
 #include <vector>
@@ -476,8 +478,9 @@ public:
 };
 
 #ifdef R__HAS_ROOT7
-template <typename BinContentType>
-class R__CLING_PTRCHECK(off) RHistFillHelper : public ROOT::Detail::RDF::RActionImpl<RHistFillHelper<BinContentType>> {
+template <typename BinContentType, bool WithWeight = false>
+class R__CLING_PTRCHECK(off) RHistFillHelper
+   : public ROOT::Detail::RDF::RActionImpl<RHistFillHelper<BinContentType, WithWeight>> {
 public:
    using Result_t = ROOT::Experimental::RHist<BinContentType>;
 
@@ -504,10 +507,25 @@ public:
    void Initialize() {}
    void InitTask(TTreeReader *, unsigned int) {}
 
+   template <typename... ColumnTypes, const std::size_t... I>
+   void
+   ExecWithWeight(unsigned int slot, const std::tuple<const ColumnTypes &...> &columnValues, std::index_sequence<I...>)
+   {
+      // Build a tuple of const references with the actual arguments, stripping the weight and avoiding copies.
+      std::tuple<const std::tuple_element_t<I, std::tuple<ColumnTypes...>> &...> args(std::get<I>(columnValues)...);
+      ROOT::Experimental::RWeight weight(std::get<sizeof...(ColumnTypes) - 1>(columnValues));
+      fContexts[slot]->Fill(args, weight);
+   }
+
    template <typename... ColumnTypes>
    void Exec(unsigned int slot, const ColumnTypes &...columnValues)
    {
-      fContexts[slot]->Fill(columnValues...);
+      if constexpr (WithWeight) {
+         auto t = std::forward_as_tuple(columnValues...);
+         ExecWithWeight(slot, t, std::make_index_sequence<sizeof...(ColumnTypes) - 1>());
+      } else {
+         fContexts[slot]->Fill(columnValues...);
+      }
    }
 
    void Finalize()
--- a/tree/dataframe/inc/ROOT/RDF/InterfaceUtils.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/InterfaceUtils.hxx
@@ -92,6 +92,7 @@ struct Histo3D{};
 struct HistoND{};
 struct HistoNSparseD{};
 struct Hist{};
+struct HistWithWeight{};
 struct Graph{};
 struct GraphAsymmErrors{};
 struct Profile1D{};
@@ -182,6 +183,18 @@ BuildAction(const ColumnNames_t &columnList, const std::shared_ptr<ROOT::Experim
             const RColumnRegister &colRegister)
 {
    using Helper_t = RHistFillHelper<BinContentType>;
+   using Action_t = RAction<Helper_t, PrevNodeType, TTraits::TypeList<ColTypes...>>;
+   return std::make_unique<Action_t>(Helper_t(h, nSlots), columnList, std::move(prevNode), colRegister);
+}
+
+// Action for RHist using RHistConcurrentFiller with weights
+template <typename... ColTypes, typename BinContentType, typename PrevNodeType>
+std::unique_ptr<RActionBase>
+BuildAction(const ColumnNames_t &columnList, const std::shared_ptr<ROOT::Experimental::RHist<BinContentType>> &h,
+            const unsigned int nSlots, std::shared_ptr<PrevNodeType> prevNode, ActionTags::HistWithWeight,
+            const RColumnRegister &colRegister)
+{
+   using Helper_t = RHistFillHelper<BinContentType, true>;
    using Action_t = RAction<Helper_t, PrevNodeType, TTraits::TypeList<ColTypes...>>;
    return std::make_unique<Action_t>(Helper_t(h, nSlots), columnList, std::move(prevNode), colRegister);
 }
--- a/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
@@ -47,7 +47,9 @@
 
 #include "RConfigure.h" // for R__HAS_ROOT7
 #ifdef R__HAS_ROOT7
+#include <ROOT/RBinWithError.hxx>
 #include <ROOT/RHist.hxx>
+#include <ROOT/RHistEngine.hxx>
 #endif
 
 #include <algorithm>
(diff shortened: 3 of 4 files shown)


tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -2357,6 +2364,142 @@ public:
...
+   /// allowed during concurrent filling.
+   ///
+   /// ### Example usage:
+   /// ~~~{.cpp}
+   /// auto h = std::make_shared<ROOT::Experimental::RHist<double>>(10, {5.0, 15.0});
+   /// auto myHist = myDf.Hist(h, {"col0"});
+   /// ~~~
+   template <typename ColumnType = RDFDetail::RInferredType, typename... ColumnTypes, typename BinContentType>
+   RResultPtr<ROOT::Experimental::RHist<BinContentType>>
+   Hist(std::shared_ptr<ROOT::Experimental::RHist<BinContentType>> h, const ColumnNames_t &columnList)
```
hahnjo, line 2417: I believe we should market this overload as "bring your own histogram" 😅 

Jokes aside, this will enable some nice use cases, for example filling a shared histogram from multiple computation graphs, as demonstrated by the `RDFHist.PtrRunGraphs` unit test.

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -2357,6 +2362,66 @@ public:
...
+   /// ROOT::Experimental::RRegularAxis axis(10, {5.0, 15.0});
+   /// auto myHist = myDf.Hist({axis}, {"col0"});
+   /// ~~~
+   template <typename BinContentType = double, typename ColumnType = RDFDetail::RInferredType, typename... ColumnTypes>
+   RResultPtr<ROOT::Experimental::RHist<BinContentType>>
+   Hist(std::vector<ROOT::Experimental::RAxisVariant> axes, const ColumnNames_t &columnList)
+   {
+      std::shared_ptr h = std::make_shared<ROOT::Experimental::RHist<BinContentType>>(std::move(axes));
+      if (h->GetNDimensions() != columnList.size()) {
+         throw std::runtime_error("Wrong number of columns for the specified number of histogram axes.");
```
hahnjo: In `histv7`, we are consistently throwing `std::invalid_argument` while RDF prefers `std::runtime_error`. I think we have to decide which "consistency" to follow...

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -2357,6 +2362,66 @@ public:
...
+   ///
+   /// This action is *lazy*: upon invocation of this method the calculation is
+   /// booked but not executed. Also see RResultPtr.
+   ///
+   /// ### Example usage:
+   /// ~~~{.cpp}
+   /// ROOT::Experimental::RRegularAxis axis(10, {5.0, 15.0});
+   /// auto myHist = myDf.Hist({axis}, {"col0"});
+   /// ~~~
+   template <typename BinContentType = double, typename ColumnType = RDFDetail::RInferredType, typename... ColumnTypes>
```
hahnjo, line 2383: One could also argue for `BinContentType = int` or `long` as default since it's unweighted filling. However, then a number of "post-processing steps" require conversion, for example scaling...

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
hahnjo, line 2228: Reminder to ourselves that we want to change this, and backport to 6.38 a warning message

tree/dataframe/inc/ROOT/RDF/InterfaceUtils.hxx
```diff
@@ -92,6 +92,7 @@ struct Histo3D{};
 struct HistoND{};
 struct HistoNSparseD{};
 struct Hist{};
+struct HistWithWeight{};
```
hahnjo, line 95: Having a second tag felt like the easiest solution, together with just a template argument `bool WithWeight = false` for `RHistFillHelper`. Other approaches are certainly possible...

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
```diff
@@ -469,6 +477,66 @@ public:
...
+
+   void Initialize() {}
+   void InitTask(TTreeReader *, unsigned int) {}
+
+   template <typename... ColumnTypes, const std::size_t... I>
+   void ExecWithWeight(unsigned int slot, const std::tuple<ColumnTypes...> &columnValues, std::index_sequence<I...>)
+   {
+      std::tuple args{std::get<I>(columnValues)...};
+      ROOT::Experimental::RWeight weight(std::get<sizeof...(ColumnTypes) - 1>(columnValues));
+      fContexts[slot]->Fill(args, weight);
```
vepadulano: Is the extra `args` tuple necessary or can we directly pass `columnValues` to `Fill`?

tree/dataframe/inc/ROOT/RDF/InterfaceUtils.hxx
vepadulano: I believe this is valid and given that everything is internal anyway I don't see a reason to spend more time on this.

tree/dataframe/inc/ROOT/RDF/InterfaceUtils.hxx
```diff
@@ -171,6 +174,32 @@ BuildAction(const ColumnNames_t &bl, const std::shared_ptr<ActionResultType> &h,
    }
 }
 
+#ifdef R__HAS_ROOT7
+// Action for RHist using RHistConcurrentFiller
```
vepadulano:
```suggestion
// Action for RHist using RHistConcurrentFiller without weights
```

tree/dataframe/inc/ROOT/RDF/InterfaceUtils.hxx
```diff
@@ -171,6 +174,32 @@ BuildAction(const ColumnNames_t &bl, const std::shared_ptr<ActionResultType> &h,
...
+BuildAction(const ColumnNames_t &columnList, const std::shared_ptr<ROOT::Experimental::RHist<BinContentType>> &h,
+            const unsigned int nSlots, std::shared_ptr<PrevNodeType> prevNode, ActionTags::Hist,
+            const RColumnRegister &colRegister)
+{
+   using Helper_t = RHistFillHelper<BinContentType>;
+   using Action_t = RAction<Helper_t, PrevNodeType, TTraits::TypeList<ColTypes...>>;
+   return std::make_unique<Action_t>(Helper_t(h, nSlots), columnList, std::move(prevNode), colRegister);
+}
+
+// Action for RHist using RHistConcurrentFiller
```
vepadulano:
```suggestion
// Action for RHist using RHistConcurrentFiller with weights
```

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
vepadulano: Opened https://github.com/root-project/root/issues/20816 to keep track of this comment 👍

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
vepadulano: I like this approach: least number of operations by default.

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
vepadulano: According to https://en.cppreference.com/w/cpp/error/runtime_error.html and https://en.cppreference.com/w/cpp/error/invalid_argument.html :

* `std::runtime_error`: It reports errors that are due to events beyond the scope of the program and cannot be easily predicted.
* `std::invalid_argument`: It reports errors that arise because an argument value has not been accepted.

I infer from this that any type of error that arises from breaking a pre-condition of a function, such as this case, should be an `invalid_argument`. In RDataFrame, we are just using `std::runtime_error` everywhere (for simplicity/consistency). If we were to apply the two definitions above, we should actually use both type of errors: `runtime_error` for things like errors in JITting, `invalid_argument` for this type of cases

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -2357,6 +2364,142 @@ public:
...
+   /// \brief Fill the provided RHist (*lazy action*).
+   /// \param[in] h The histogram that should be filled.
+   /// \param[in] columnList A list containing the names of the columns that will be passed when calling `Fill`
+   /// \return the histogram wrapped in a RResultPtr.
+   ///
+   /// This action is *lazy*: upon invocation of this method the calculation is
+   /// booked but not executed. Also see RResultPtr.
+   ///
+   /// During execution of the computation graph, the passed histogram must only be accessed with methods that are
+   /// allowed during concurrent filling.
```
vepadulano, line 2408: This is always a tricky subject to communicate to users, so I wonder if we can be even more explicit here?

```suggestion
   /// During execution of the computation graph, the passed histogram must only be accessed with methods that are
   /// allowed during concurrent filling, i.e. that are thread-safe
```

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
vepadulano: This is great!

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -2357,6 +2364,142 @@ public:
...
+             typename ColumnType = RDFDetail::RInferredType, typename... ColumnTypes>
+   RResultPtr<ROOT::Experimental::RHist<BinContentType>>
+   Hist(std::vector<ROOT::Experimental::RAxisVariant> axes, const ColumnNames_t &columnList, std::string_view wName)
+   {
+      static_assert(ROOT::Experimental::RHistEngine<BinContentType>::SupportsWeightedFilling,
+                    "weighted filling is not supported for integral bin content types");
+
+      std::shared_ptr h = std::make_shared<ROOT::Experimental::RHist<BinContentType>>(std::move(axes));
+      if (h->GetNDimensions() != columnList.size()) {
+         throw std::runtime_error("Wrong number of columns for the specified number of histogram axes.");
```
vepadulano: Since we know both numbers, we could think about making the error a bit more friendly:

```suggestion
         throw std::runtime_error("Wrong number of columns for the specified number of histogram axes: need '" + std::to_string(h->GetNDimensions()) + "', got '" + std::to_string(columnList.size()) + "'.");
```

Applies similarly in other places.

tree/dataframe/test/dataframe_hist.cxx
```diff
@@ -0,0 +1,274 @@
...
+   for (auto index : axis.GetNormalRange()) {
+      EXPECT_EQ(hist->GetBinContent(index), 1.0);
+   }
+}
+
+TEST_P(RDFHist, RegularJit)
+{
+   RDataFrame df(10);
+   const RRegularAxis axis(10, {5.0, 15.0});
+   auto hist = df.Define("x", "rdfentry_ + 5.5").Hist({axis}, {"x"});
```
vepadulano, line 58: Thinking out loud, outside the scope of this PR, about how much JITting we strictly need. In principle, the bin content type defaults to double, so what we're JITting is the column type, I wonder if we could find a way to ask the type-system to just read the contents of the column truly as double, then we'd have at least one default scenario without JITting and without the need for template arguments.

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
hahnjo: Ok, I will change the exception type accordingly.

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
hahnjo: The problem is that "thread-safe" alone may be too weak: It is thread-safe to call `const` methods on the same object, but they are not automatically allowed during concurrent filling. Even when reads of individual bins are atomic, for example a naive `Clone()` will be biased towards "later" bins (that's where we need the future `SnapshotAtomic` instead). When implemented, I will add a list of "allowed methods" to https://github.com/root-project/root/blob/master/hist/histv7/doc/ConcurrentFilling.md that we should eventually link.

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
hahnjo: Done (with slightly different wording, and also now throwing `std::invalid_argument`)

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
hahnjo: We cannot pass `columnValues` directly because we need to strip off the last element (the weight). We could get rid of the additional temporary by writing
```c++
fContexts[slot]->Fill(std::get<I>(columnValues)..., weight);
```
but that actually doesn't gain much because then the variadic `Fill` overload recreates the `tuple` :shrug:

tree/dataframe/test/dataframe_hist.cxx
hahnjo: If everything was a `double`, we could do that but we will probably also want to support strings and (discrete) integers for the categorical axis types...

vepadulano: Yes, good point, it's not so trivial, but something to think about for the long term perhaps.

tree/dataframe/inc/ROOT/RDF/ActionHelpers.hxx
hahnjo: Wait, I discovered that the `tuple` would copy the arguments. Avoiding that requires a bit more magic.

> but that actually doesn't gain much because then the variadic `Fill` overload recreates the `tuple`

Correction: the variadic `Fill` overload would recreate a `tuple` including the converted `RWeight` argument, only to then extract it again. So that's not really nice (even though it would have more easily solved the problem of copying).

hahnjo:
> I believe as a followup we should introduce at some point one (or more) tutorials, specifically with examples that compare using `Histo[1,2,3,N]D` and `Hist` to help users in the future transition (or not, depending on what they will want to do).

Yes, tutorials are on my list. We probably want tutorials for:
 * Concurrent filling (without RDF, only with `RHistConcurrentFiller`)
 * Conversion to `TH*`
 * Usage in RDataFrame (as added by this PR)
 * Translations of (some) `df1*` tutorials to show the steps working together

PR #20675, opened 2025-12-09
[RF] Fix RooLognormal analytical integral for xMin < zero

guitargeek:
Just like for other pdfs that are strictly zero for negative observable values, the integral for xMin < zero should still evaluate correctly (compare for example with the RooLandau).

Addresses this forum report:
https://root-forum.cern.ch/t/roofit-pdfs-with-different-domain-in-the-same-model-e-g-gauss-lognormal/64505/3

Little demo that it works now:
```c++
RooRealVar x("x", "x", 0.0, 10.0);
RooRealVar mu("mu", "mu", 1.0);
RooRealVar sigma("sigma", "sigma", 0.5);
RooLognormal logn("logn", "lognormal pdf", x, mu, sigma);

double a = -1.0;
double b = 5.0;

x.setRange("intRange", -1.0, 5.0);

RooAbsReal* integral = logn.createIntegral(x, RooFit::NormSet(x),
                                           RooFit::Range("intRange"));

double result = integral->getVal();

std::cout << "Integral of Lognormal from " << a << " to " << b
          << " = " << result << std::endl;
```

### DIFF
--- a/roofit/roofit/src/RooLognormal.cxx
+++ b/roofit/roofit/src/RooLognormal.cxx
@@ -27,7 +27,6 @@ The parameterization here is physics driven and differs from the ROOT::Math::log
 
 #include "RooLognormal.h"
 #include "RooRandom.h"
-#include "RooMath.h"
 #include "RooHelpers.h"
 #include "RooBatchCompute.h"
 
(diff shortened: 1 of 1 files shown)


guitargeek:
That is true. It's debatable whether a lognormal over a negative domain makes sense. But here we had no choice if we want to be consistent. Evaluating a RooLognormal for x < 0 gives zero and not NaN, so the integral needs to be consistent with that if we don't want to confuse users

PR #20689, opened 2025-12-11
Remove tutorials requiring py-hsimple.root from wheel tests

vepadulano:


siliataider:
So these 2 tutorials are failing because the file needed is created by [`tutorials/hsimple.py`](https://github.com/root-project/root/blob/master/tutorials/hsimple.py) which is not run, maybe we should just run it instead of disabling them?

vepadulano:
> So these 2 tutorials are failing because the file needed is created by [tutorials/hsimple.py](https://github.com/root-project/root/blob/master/tutorials/hsimple.py) which is not run, maybe we should just run it instead of disabling them?

Even after running the hsimple.py tutorial *before* the other tutorials, the file is still not found. In the interest of returning the CI to a working state, I will go back to removing the tutorials and then we can tackle this another time.

Issue #10058, opened 2022-03-05
CINT still around


ferdymercury:
### Explain what you would like to see improved
grep on the source code still shows many places where CINT is mentioned.

### Optional: share how it could be improved
Change to equivalent variables in Cling, or remove if no longer needed.

For example:

`__CINT__` --> `__CLING__`

### To Reproduce
grep -r CINT root-src

### Setup
```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.27/01                        https://root.cern |
  | (c) 1995-2021, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Jan 12 2022, 10:17:19                 |
  | From heads/master@v6-25-01-2870-gdac9b6398d                      |
  | With c++ (Ubuntu 8.4.0-1ubuntu1~18.04) 8.4.0                     |
  | Try '.help', '.demo', '.license', '.credits', '.quit'/'.q'       |
   ------------------------------------------------------------------
```

ferdymercury:
@vgvassilev just addressed part of this issue with a couple of commits

dpiparo:
Maybe a catalogue of the places where __CINT__ can be found could help? It could be easier to track progress with the migration...

ferdymercury:
Here it is:

bindings:

- [x] bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx:        return (intptr_t)m->GetOffsetCint();    // yes, CINT (GetOffset() is both wrong
- [x] bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/cppyy.h:struct CPPYY_G__DUMMY_FOR_CINT7 {
- [x] bindings/pyroot/cppyy/cppyy/doc/source/philosophy.rst:The `CINT`_ interpreter was already well established in HEP, however, and
- [x] bindings/pyroot/cppyy/CPyCppyy/src/PyResult.cxx:// you have to use the void* converter, as CINT will not call any other).
- [x] bindings/pyroot/pythonizations/src/TPyDispatcher.cxx:// to install callbacks from CINT.
- [x] bindings/tpython/src/TPyReturn.cxx:/// you have to use the void* converter, as CINT will not call any other).

cmake:

- [x] cmake/modules/RootConfiguration.cmake:if(IS_ABSOLUTE ${CMAKE_INSTALL_CINTINCDIR})
[...]

PR #20694, opened 2025-12-11
[gui,graf,hist,io,tree,tmva] rename CINT to Cling

ferdymercury:
Subpart of work towards https://github.com/root-project/root/issues/10058

### DIFF
--- a/tree/treeplayer/src/TSelectorDraw.cxx
+++ b/tree/treeplayer/src/TSelectorDraw.cxx
@@ -1527,7 +1527,7 @@ void TSelectorDraw::TakeAction()
    else if (fAction == 6 || fAction == 7) {
       TakeEstimate();
       bool candle = (fAction == 7);
-      // Using Cling to avoid a dependency in TParallelCoord
+      // Using Interpreter to avoid a dependency in TParallelCoord
       if (!fOption.Contains("goff"))
          gROOT->ProcessLine(TString::Format("TParallelCoord::BuildParallelCoord((TSelectorDraw*)0x%zx,0x%zx)",
                                 (size_t)this, (size_t)candle));


graf2d/cocoa/inc/TGQuartz.h
```diff
@@ -96,7 +96,7 @@ class TGQuartz : public TGCocoa {
    void AlignTTFString();
    Bool_t IsTTFStringVisible(Int_t x, Int_t y, UInt_t w, UInt_t h);
    void RenderTTFString(Int_t x, Int_t y, ETextMode mode);
-   //I have to use void * instead of QuartzPixmap * because of CINT :(
+   //TODO: use void * instead of QuartzPixmap *
```
guitargeek:
```suggestion
   //TODO: move internal headers like TGQuartz.h out of the public ROOT user interface. Then we can also use XQuartz types like QuartzPixmap* instead of void * in functions like these
```
Right now, GCocoa is a public `ROOT_STANDARD_LIBRARY_PACKAGE`, where the headers get installed for the user, which doesn't make sense to me. Until we change that, we should stick with `void *` to not require users to include the XQuartz header, should they have the crazy idea to include `TGQuartz.h`.

graf2d/fitsio/src/TFITS.cxx
```diff
@@ -42,7 +42,7 @@
 ///
 /// Please have a look to the tutorials ($ROOTSYS/tutorials/fitsio/) to see
 /// some examples. IMPORTANT: to run tutorials it is required that
-/// you change the current working directory of ROOT (CINT) shell to the
+/// you change the current working directory of ROOT (Cling) shell to the
```
guitargeek:
```suggestion
/// you change the current working directory of the ROOT shell to the
```
No need to mention implementation details here.

graf2d/win32gdk/inc/TGWin32ProxyDefs.h
```diff
@@ -13,7 +13,7 @@
 #ifndef ROOT_TGWin32ProxyDefs
 #define ROOT_TGWin32ProxyDefs
 
-#include "Rtypes.h"  // CINT headers
+#include "Rtypes.h"  // Cling headers
```
guitargeek:
```suggestion
#include "Rtypes.h"
```
Seems wrong here.

graf3d/eve/src/TEveBrowser.cxx
```diff
@@ -463,7 +463,7 @@ Specialization of TRootBrowser for Eve.
 
 
 ////////////////////////////////////////////////////////////////////////////////
-/// Add "Export to CINT" into context-menu for class cl.
+/// Add "Export to CINT" (now Cling) into context-menu for class cl.
```
guitargeek:
```suggestion
/// Add "Export to CINT" into context-menu for class cl.
```
I would not make that change now. I think it's better to change the name of the menu item "Export the CINT" in the code below and the docs at the same time, in a smaller PR that can be exclusively reviewed by @alja

graf3d/eve7/src/REveElement.cxx
```diff
@@ -601,7 +601,7 @@ TClass *REveElement::IsA() const
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-/// Export render-element to CINT with variable name var_name.
+/// Export render-element to Cling with variable name var_name.
```
guitargeek: Can we maybe spin off all the eve changes into a separate PR? This code is not owned by us folks at CERN, to it would be better to keep the changes for a different PR reviewed by @alja

graf3d/gl/inc/TGLParametric.h
```diff
@@ -49,7 +49,7 @@ class TGLParametricEquation : public TNamed {
    Bool_t               fModified;
 
    typedef std::unique_ptr<TGLHistPainter> Painter_t;
-   //C++ compiler do not need TGLhistPainter definition here, but I'm not sure about CINT,
+   //C++ compiler do not need TGLhHstPainter definition here, but I'm not sure about Cling,
```
guitargeek:
```suggestion
```
You can remove that comment. The definition is needed, because the destructor of `TGLParametricEquation` is implicitly defined in this header file, and it will have to know how to delete the `TGLHistPainter` because of the `unique_ptr`. So the definition is needed.

gui/ged/src/TStyleManager.cxx
```diff
@@ -75,10 +75,10 @@ Delete a style:
 The selected style can be deleted from the list when you use the
 Style menu/Delete or the corresponding tool bar button. The selected
 style is removed from the list of all available styles for the
-current ROOT session. WARRNING: it will be lost if you didn't saved
+current ROOT session. WARNING: it will be lost if you didn't saved
 it in a C++ macro file before its deletion. Also, you cannot delete
 the selected style if it is set to gStyle. A message 'Can not delete
-gStyle' will be displayed on the CINT prompt.
+gStyle' will be displayed on the Cling prompt.
```
guitargeek:
```suggestion
gStyle' will be displayed on the ROOT prompt.
```

io/io/src/TStreamerInfo.cxx
```diff
@@ -1108,7 +1108,7 @@ void TStreamerInfo::BuildCheck(TFile *file /* = 0 */, Bool_t load /* = kTRUE */)
                // match in the 'current' in memory TStreamerInfo for
                // a non foreign class (we can not get here if this is
                // a foreign class so we do not need to test it),
-               // we need to add this one more test since the CINT behaviour
+               // we need to add this one more test since the Cling behaviour
```
guitargeek:
```suggestion
               // we need to add this one more test since the interpreter behaviour
```
That would not be a correct statement if the behavior was stable since the migration to Cling. Maybe generalize to "interpreter" here.

io/io/src/TStreamerInfo.cxx
```diff
@@ -1158,7 +1158,7 @@ void TStreamerInfo::BuildCheck(TFile *file /* = 0 */, Bool_t load /* = kTRUE */)
...
```
guitargeek:
```suggestion
               // we need to add this one more test since the interpreter behaviour
```

io/xml/src/TXMLFile.cxx
```diff
@@ -24,7 +24,7 @@
 // The XML file is built in memory before being dumped to disk.
 //
 // Like for normal ROOT files, XML files use the same I/O mechanism
-// exploiting the ROOT/CINT dictionary. Any class having a dictionary
+// exploiting the ROOT/Cling dictionary. Any class having a dictionary
```
guitargeek:
```suggestion
// exploiting the I/O dictionary. Any class having a dictionary
```

tree/treeplayer/src/TTreeProxyGenerator.cxx
```diff
@@ -29,9 +29,9 @@
 
   When using in interpreted mode understand why the reloading reloads the calling script and then crashes :(
 
-  CINT does not properly call the custom operators when doing return fNtrack.
+  Cling does not properly call the custom operators when doing return fNtrack.
```
guitargeek: Does that really still apply to Cling? Is that easy to figure out by reading the source code in this file? I would be more on the careful side and keep CINT here for now.

tree/treeplayer/src/TTreeProxyGenerator.cxx
```diff
@@ -29,9 +29,9 @@
...
 
-  CINT does not handle fMatrix[2][1] well.
+  Cling does not handle fMatrix[2][1] well.
```
guitargeek: Also here, I would maybe keep CINT, unless you can confirm Cling also doesn't handle that well.

guitargeek (changes requested):
Thank you! A few comments inline, and thanks for adding the `TODO`s

graf3d/eve/src/TEveElement.cxx
```diff
@@ -512,7 +512,7 @@ void TEveElement::SaveVizParams(std::ostream& out, const TString& tag, const TSt
 /// Write-out visual parameters for this object.
 /// This is a virtual function and all sub-classes are required to
 /// first call the base-element version.
-/// The name of the element pointer is 'x%03d', due to cint limitations.
+/// The name of the element pointer is 'x%03d', due to historial CINT limitations.
```
ferdymercury:
```suggestion
/// The name of the element pointer is 'x%03d', due to historical CINT limitations.
```

tree/treeplayer/inc/TBranchProxyDirector.h
```diff
@@ -47,7 +47,7 @@ namespace Internal{
    public:
 
       TBranchProxyDirector(TTree* tree, Long64_t i);
-      TBranchProxyDirector(TTree* tree, Int_t i);     // cint has (had?) a problem casting int to long long
+      TBranchProxyDirector(TTree* tree, Int_t i);     // Historial: CINT had a problem casting int to long long
```
ferdymercury:
```suggestion
      TBranchProxyDirector(TTree* tree, Int_t i);     // Historical: CINT had a problem casting int to long long
```

tree/treeplayer/src/TTreeProxyGenerator.cxx
ferdymercury:
```suggestion
  TODO check now with Cling: CINT did not properly call the custom operators when doing return fNtrack.
```

tree/treeplayer/src/TTreeProxyGenerator.cxx
ferdymercury:
```suggestion
  TODO check with Cling: CINT did not handle fMatrix[2][1] well.
```

graf3d/gl/inc/TGLParametric.h
```diff
@@ -49,7 +49,6 @@ class TGLParametricEquation : public TNamed {
    Bool_t               fModified;
 
    typedef std::unique_ptr<TGLHistPainter> Painter_t;
-   //C++ compiler do not need TGLhistPainter definition here, but I'm not sure about CINT,
    //so I've included TGLHistPainter definition.
```
ferdymercury:
```suggestion
```

graf3d/eve7/src/REveElement.cxx
ferdymercury: Sure, done. :)

tree/treeplayer/src/TSelectorDraw.cxx
```diff
@@ -1527,7 +1527,7 @@ void TSelectorDraw::TakeAction()
    else if (fAction == 6 || fAction == 7) {
       TakeEstimate();
       bool candle = (fAction == 7);
-      // Using CINT to avoid a dependency in TParallelCoord
+      // Using Cling to avoid a dependency in TParallelCoord
```
pcanal: Minor

```suggestion
      // Using Interpreter to avoid a dependency in TParallelCoord
```

pcanal (approved):
Thanks.

Issue #20687, opened 2025-12-10
Zero-parameter `TF1` created from C++ functor crashes `TCanvas::SaveAs()`


bgrube:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Creating a `TF1` from a C++ functor with the number of parameters set to zero and drawing this function into a `TCanvas`, causes a Python `TypeError` when attempting to save the canvas as a PDF file.

### Reproducer

Running the script `tf1bug.py`
```python
#!/usr/bin/env python3

import ROOT

# function without parameters
def f1(args, pars):
  return 4.0 * args[0]

# functor without parameters
F2_CPP = """
class Functor {
public:

	Functor()
	{ }

	double
	operator () (
		double* args,
		double*
	) {
		return 4.0 * args[0];
	}
};
"""

if __name__ == "__main__":
  ROOT.gROOT.SetBatch(True)
  fcn1 = ROOT.TF1("f1", f1, 0, 1, 0)  # using a Python function works fine
  ROOT.gInterpreter.Declare(F2_CPP)
  functor = ROOT.Functor()
  fcn2 = ROOT.TF1("f2", functor, 0, 1, 1)  # setting the number of parameters to 1 instead of 0 makes the bug go away
  fcn3 = ROOT.TF1("f3", functor, 0, 1, 0)  # <- setting the number of parameters to 0 causes the bug
  for fcn in (fcn1, fcn2, fcn3):
    canv = ROOT.TCanvas()
    fcn.Draw()
    canv.SaveAs(f"{fcn.GetName()}.pdf")
```
in ROOT 6.38.00 yields
```
Info in <TCanvas::Print>: pdf file f1.pdf has been created
Info in <TCanvas::Print>: pdf file f2.pdf has been created
Traceback (most recent call last):
  File "./tf1bug.py", line 35, in <module>
    canv.SaveAs(f"{fcn.GetName()}.pdf")
TypeError: could not convert argument to buffer or nullptr
```
With ROOT 6.32.20, I get
```
Info in <TCanvas::Print>: pdf file f1.pdf has been created
Info in <TCanvas::Print>: pdf file f2.pdf has been created
Traceback (most recent call last):
  File "/u/home/bgrube/./tf1bug.py", line 35, in <module>
    canv.SaveAs(f"{fcn.GetName()}.pdf")
TypeError: void TPad::SaveAs(const char* filename = "", Option_t* option = "") =>
    TypeError: double Functor::operator()(double* args, double*) =>
[...]

PR #20696, opened 2025-12-11
[cppyy] Disallow conversion from C++ instance to function pointers

guitargeek:
Implicit conversion from C++ instances to function pointers is not allowed in C++, and it can result in confusing overload resolutions if we allow it in cppyy.

See:
https://github.com/root-project/root/issues/20687#issuecomment-3641579159

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
@@ -2655,7 +2655,7 @@ static PyMethodDef gWrapperCacheEraserMethodDef = {
 };
 
 static void* PyFunction_AsCPointer(PyObject* pyobject,
-    const std::string& rettype, const std::string& signature)
+    const std::string& rettype, const std::string& signature, bool allowCppInstance)
 {
 // Convert a bound C++ function pointer or callable python object to a C-style
 // function pointer. The former is direct, the latter involves a JIT-ed wrapper.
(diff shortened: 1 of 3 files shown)


bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
```diff
@@ -2702,8 +2702,12 @@ static void* PyFunction_AsCPointer(PyObject* pyobject,
         return fptr;
     }
 
-    if (PyCallable_Check(pyobject)) {
+
+    if (PyCallable_Check(pyobject) && (allowCppInstance || !CPPInstance_Check(pyobject))) {
     // generic python callable: create a C++ wrapper function
+    // Sometimes we don't want to take this branch if the object is a C++
+    // instance, because C++ doesn't allow converting functor obejcts to
```
ferdymercury:
```suggestion
    // instance, because C++ doesn't allow converting functor objects to
```

bindings/pyroot/cppyy/cppyy/test/test_overloads.py
```diff
@@ -405,5 +405,51 @@ class Derived : public Base {
 
         assert result_instance == result_direct
 
+
+    def test14_disallow_functor_to_function_pointer(self):
+        """Make sure we're no allowing to convert C++ functors to funciton
```
ferdymercury:
```suggestion
        """Make sure we're no allowing to convert C++ functors to function
```

PR #20702, opened 2025-12-11
[hist] Remove CINT compatibility constructors from TF1, THF2, and TF3

guitargeek:
The `TF1`, `TF2`, and `TF3` constructors for CINT compatibility can be removed. This concerns the templated constructors that additionally took the name of the used functor class and member function. With ROOT 6, these names can be omitted.

See also commit 4e4cdfb from 2015, where these constructors were introduced.

The idea is to clear a bit more the jungle of `TH1` constructors, of which there are too many. That makes it confusing for our uses and for cppyy overload resolution.

### DIFF
--- a/hist/hist/inc/TF1.h
+++ b/hist/hist/inc/TF1.h
@@ -352,15 +352,6 @@ public:
       ROOT::Internal::TF1Builder<Func>::Build(this, f);
    }
 
-   /// backward compatible interface
-   template <typename Func>
-   TF1(const char *name, Func f, Double_t xmin, Double_t xmax, Int_t npar, const char *, EAddToList addToGlobList = EAddToList::kDefault) :
-      TF1(EFType::kTemplScalar, name, xmin, xmax, npar, 1, addToGlobList, new TF1Parameters(npar))
-   {
-      ROOT::Internal::TF1Builder<Func>::Build(this, f);
-   }
-
-
    /// Template constructors from a pointer to any C++ class of type PtrObj with a specific member function of type
    /// MemFn.
    /// The member function must have the signature of  (double * , double *) and returning a double.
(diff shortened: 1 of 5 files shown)


hist/hist/inc/TF1.h
```diff
@@ -352,15 +352,6 @@ class TF1 : public TNamed, public TAttLine, public TAttFill, public TAttMarker {
       ROOT::Internal::TF1Builder<Func>::Build(this, f);
    }
 
-   // backward compatible interface
-   template <typename Func>
-   TF1(const char *name, Func f, Double_t xmin, Double_t xmax, Int_t npar, const char *, EAddToList addToGlobList = EAddToList::kDefault) :
```
pcanal, line 357: Should we deprecate them first or are we sure they were only use by/via CINT?

guitargeek: I don't see why anyone would use these constructors otherwise, as it's not needed in C++ to pass the name of the class  instance `Func` as a separate argument.

The reason why I think we can remove it without deprecation is that the replacement is trivial and backwards compatible: just drop these `const char*` inputs at the end. But if you're more comfortable with a deprecation period, I have no problem.

Maybe @lmoneta also has a opinion there?

lmoneta: I think we can just remove them, were already used as backward compatible for 10 years, but if you prefer to deprecate them is fine with me. 
We just need to update the documentation in TF1.cxx, see 
https://root.cern.ch/doc/master/classTF1.html#F6
where the names are passed in the constructor. 
We should check if in other places they are documented

lmoneta (approved):
LGTM, but please update the doc in 
https://root.cern.ch/doc/master/classTF1.html#F6

guitargeek:
Related discussion on the ROOT forum:

https://root-forum.cern.ch/t/tf1-constructor-broken-in-root-6-40-02/64903/7

PR #20711, opened 2025-12-14
[test] Make all tests green when using a minimal build

### LINKED ISSUE #20315: Some test fails with a minimal build
As seen at: https://github.com/root-project/root/pull/20311/checks?check_run_id=54747021200
The following tests fail with a minimal build:
```
30:gtest-core-rint-TTabComTests
1535:roottest-root-meta-execOptionList-auto
1881:roottest-root-multicore-processExecutorH1Test
1920:pyunittests-roottest-root-ntuple-atlas-datavector-rentry-getptr
1921:pyunittests-roottest-root-ntuple-atlas-datavector-template-instantiation
```
I.e. they should probably be disabled or adapted when some of the feature they (indirectly) rely on is disabled.

dpiparo:
Adapt the tests in order to make them run properly with minimal builds or disable them if they do not make sense for that configuration.

Fixes #20315

### DIFF
--- a/core/rint/test/CMakeLists.txt
+++ b/core/rint/test/CMakeLists.txt
@@ -4,6 +4,8 @@
 # For the licensing terms see $ROOTSYS/LICENSE.
 # For the list of contributors see $ROOTSYS/README/CREDITS.
 
-ROOT_ADD_GTEST(TTabComTests TTabComTests.cxx LIBRARIES Rint)
+if(NOT minimal)
+    ROOT_ADD_GTEST(TTabComTests TTabComTests.cxx LIBRARIES Rint)
+endif()
 ROOT_ADD_GTEST(TRintTests TRintTests.cxx LIBRARIES Rint)
 


dpiparo:
We'll see the tests, however the formatting errors reflect the original formatting and might not be really worth fixing (at the cost of changing the formatting of stable tests)

ferdymercury:
A bit related: maybe we should also add a cmake error if minimal=ON but X11/asimage=ON, see https://root-forum.cern.ch/t/cant-built-minimal-root-6-38-00-with-x11-and-asimage/64532

core/rint/test/TTabComTests.cxx
```diff
@@ -71,7 +71,10 @@ TEST(TTabComTests, CompleteTH1)
       // FIXME: See ROOT-10989
       " TH1DModel"
 #endif
-      " TH1Editor TH1F TH1I TH1L TH1S";
+#ifndef R__IS_MINIMAL
```
guitargeek:
```suggestion
#if not defined(ROOT_IS_MINIMAL)
```
I would prefer to not add a ROOT runtime preprocessor macro `R__IS_MINIMAL` that is also visible to the user if there is no strong reason for it. Because then we need to keep it in principle forever for backwards compatibility, locking ourselves in without a strong reason. Instead, one can add a
```cmake
if (minimal)
    target_compile_definitions(TTabComTests PRIVATE ROOT_IS_MINIMAL)
endif()
```
to the [CMakeLists.txt](https://github.com/root-project/root/blob/master/core/rint/test/CMakeLists.txt). The runtime `R__` flags only need to be defined if we need to do these conditionals also in the header files included by the users, but here it's just for a test.

dpiparo: This is a good suggestion. I stopped the builds and will change that part of the PR.

dpiparo:
Thanks for the advice!

core/rint/test/TTabComTests.cxx
pcanal: > I would just suggest to not introduce a public R__IS_MINIMAL macro at this point.

I disagree.  The fact that the test are failing under the minimal build is only a secondary symptoms.  The real issue is that the tests were failing because a feature they depend on was turned off and they were not reacting properly to that turning off.  The problem would still have been there if the users explicitly disable 'just' those feature.

core/rint/test/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@
 # For the licensing terms see $ROOTSYS/LICENSE.
 # For the list of contributors see $ROOTSYS/README/CREDITS.
 
-ROOT_ADD_GTEST(TTabComTests TTabComTests.cxx LIBRARIES Rint)
+if(NOT minimal)
```
pcanal, line 7: Is that really the best test?  What 'feature(s)' is missing that make the test not working?

roottest/root/meta/CMakeLists.txt
```diff
@@ -89,4 +89,9 @@ ROOTTEST_ADD_TEST(hello_ROOT_C
                   OUTREF hello_ROOT_C_${libsuf}.ref
                   FIXTURES_REQUIRED root-meta-hello_ROOT-fixture)
 
+if(NOT minimal)
```
pcanal, line 92: Is that really the best test?  What 'feature(s)' is missing that make the test not working?

tutorials/CMakeLists.txt
```diff
@@ -440,6 +423,12 @@ else()
   list(APPEND root7_veto io/ntuple/ntpl011_global_temperatures.C)
 endif()
 
+#---These do not run with minimal
+if(minimal)
```
pcanal, line 427: Is that really the best test?  What 'feature' is missing that make the test not working?

core/rint/test/TTabComTests.cxx
guitargeek: I don't understand how the `R__IS_MINIMAL` macro would sustainably address the issue you describe.

If it's useful for the user to check if ROOT has a certain feature, we should introduce `R__HAS_FEATURE` macros like we always did.

The `R__IS_MINIMAL` would be mute, because one can toggle any an feature on top of a minimal build. Why would you use it then?

PR #20716, opened 2025-12-15
[core] Remove private type alias

### LINKED ISSUE #20572: A public field with a "private" type in the TFileMergeInfo class

dpiparo:
fixes #20572

### DIFF
--- a/core/meta/inc/TFileMergeInfo.h
+++ b/core/meta/inc/TFileMergeInfo.h
@@ -8,25 +8,12 @@
  * For the list of contributors see $ROOTSYS/README/CREDITS.             *
  *************************************************************************/
 
+/** \class TFileMergeInfo
+A class to pass information from the TFileMerger to the objects being merged.
+*/
+
 #ifndef ROOT_TFileMergeInfo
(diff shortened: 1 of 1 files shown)


ferdymercury:
Thanks. Does this supersede https://github.com/root-project/root/pull/20576 ?

see relevant missing things here:
https://github.com/root-project/root/pull/20576#issuecomment-3593181259

PR #20718, opened 2025-12-15
[win] Fix XMLReader on Windows

bellenot:
- Use `std::ios::binary` flag to fix `LF` vs `CRLF` issue on Windows
- Enable related failing tests


dpiparo (approved):
Thanks for these changes! I am sorry for not having thought about this aspect before myself in the first place.

pcanal (approved):

Issue #20719, opened 2025-12-15
[math] TH2 FillRandom xygaus does not work


ferdymercury:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

This works:
```
TH1F* h = new TH1F("h", "" , 10, 0, 10);
h->FillRandom("gaus");
h->Draw();
```

This fails:
```
TH2F* h = new TH2F("h", "" , 10, 0, 10, 10, 0, 10);
h->FillRandom("xygaus");
h->Draw();
```

`Error in <TH2F::FillRandom>: Unknown function: xygaus`

### Reproducer

```
TH2F* h = new TH2F("h", "" , 10, 0, 10, 10, 0, 10);
h->FillRandom("xygaus");
h->Draw();
```

### ROOT version

```
ROOT v6.39.01
Built for linuxx8664gcc on Dec 13 2025, 17:24:54
From heads/tspec2back@v6-23-01-RF-binSampling-25185-g080daba378
With g++ (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0 std201703
Binary directory: /home/user/build/build-root_src-Desktop-Debug/bin
```

### Installation method

Source

### Operating system

Ubuntu22.04

### Additional context

@guitargeek is this maybe a side effect of ef2209c7b198e4338f31a702108063dc15c475a7 ?

guitargeek:
I don't see the connection to https://github.com/root-project/root/commit/ef2209c7b198e4338f31a702108063dc15c475a7. Is this a regression with respect to some previously released ROOT version? If yes, which was the last one that worked/first that fails?

ferdymercury:
You're right, sorry. I think it's rather an incomplete feature.

PR #20720, opened 2025-12-15
[math] allow TH2::FillRandom xygaus

ferdymercury:
Fixes https://github.com/root-project/root/issues/20719

Fix Depends on: https://github.com/root-project/root/pull/20734


dpiparo:
Thanks for the changes :) I am trying to figure out the issue with `xyexpo`. That is a real bug, uncovered while completing this feature.

PR #20723, opened 2025-12-15
[RF][HF] Support setting ShapeFactor value and range

### LINKED ISSUE #20697: [RF] Add support to set the nominal value and limits for the ShapeFactor parameters in HistFactory

Currently, the code allows to pass the nominal value and the limits for the NormFactors via:
```c++
void RooStats::HistFactory::Sample::AddNormFactor	(	std::string const &	Name,
double	Val,
double	Low,
double	High )
```

But there is no similar function for ShapeFactors, only:
```c++
void RooStats::HistFactory::Sample::AddShapeFactor	(	std::string	Name	)	
```

is available. This has the disadvantage that in some cases the limits hardcoded in the code might be very sub-optimal resulting in convergence issues. 

Please note, the ShapeFactors are important e.g. for ABCD esimate, see page 23 here (and replace NormFactor with ShapeFactor): https://cds.cern.ch/record/1456844/files/CERN-OPEN-2012-016.pdf

### ROOT version

any

### Installation method

any

### Operating system

any

### Additional context

_No response_

guitargeek:
Adds a Sample::AddShapeFactor overload that takes the initial value and the range of the ShapeFactor gammas, analogous to AddNormFactor. This is important e.g. for ABCD estimates, where the hard-coded default range can cause convergence problems.

The value and range are persisted to and read back from both XML and ROOT files, and a dedicated test covers the full round trips as well as the resulting workspace parameters. The ShapeFactor element in the XML schema (HistFactorySchema.dtd) is updated accordingly. The value and range use the same 'Val' / 'Low' / 'High' attribute and accessor names as NormFactor for consistency.

Closes https://github.com/root-project/root/issues/20697.

### DIFF
--- a/roofit/etc/HistFactorySchema.dtd
+++ b/roofit/etc/HistFactorySchema.dtd
@@ -156,5 +156,12 @@ For this element there is no sublemenents so the setting will only have local ef
 <!ELEMENT ShapeFactor EMPTY>
 <!ATTLIST ShapeFactor
           Name              CDATA            #REQUIRED
+          Val               CDATA            #IMPLIED
+          Low               CDATA            #IMPLIED
+          High              CDATA            #IMPLIED
+          Const             CDATA            #IMPLIED
+          InputFile         CDATA            #IMPLIED
+          HistoName         CDATA            #IMPLIED
+          HistoPath         CDATA            #IMPLIED
 >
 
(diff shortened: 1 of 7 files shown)


lmoneta (approved):
LGTM!

guitargeek:
Backporting also to 6.40 on request by ATLAS (@TomasDado)

PR #20725, opened 2025-12-15
CI: test minimal builds

ferdymercury:
Just rebased version of pcanal's https://github.com/root-project/root/pull/20311


guitargeek:
Thank you for this! Looks like we need a check here, around the Python tests:
https://github.com/root-project/root/blob/master/test/PostInstall/CMakeLists.txt#L21
```CMake
if(BUILD_TESTING AND ROOT_pyroot_FOUND)
```
Can you try to add that?

.github/workflows/root-ci.yml
```diff
@@ -385,6 +385,10 @@ jobs:
           - image: alma9
             overrides: ["CMAKE_BUILD_TYPE=Debug"]
           - image: alma10
+          # Minimal build
+          - image: alma10
+            minimal: true
+            property: "Minimal build"
```
guitargeek:
```suggestion
            property: "minimal build"
```
All the other properties are lower case.

.github/workflows/root-ci.yml
```diff
@@ -490,6 +494,19 @@ jobs:
...
+          GLOBAL_CONFIG_DIR: '.github/workflows/root-ci-config/buildconfig'
+          CONFIGFILE_STEM: '.github/workflows/root-ci-config/buildconfig/${{ matrix.image }}'
+        run: |
+          echo "Applying minimal request options"
+          # Add commands to apply minimal request options here
+          set -x
+          cp "$GLOBAL_CONFIG_DIR/global-minimal.txt" "$GLOBAL_CONFIG_DIR/global.txt"
+          if [ -f "${CONFIGFILE_STEM}-minimal.txt" ]; then
+            cp "${CONFIGFILE_STEM}-minimal.txt" "${CONFIGFILE_STEM}.txt"
+          fi
```
guitargeek, line 509: IMHO this mechanism only makes sense if we plan to have minimal builds on multiple platforms (which I'd argue is not worth it). Otherwise we gain nothing by splitting `global-minimal.txt` and `alma10-minimal.txt`.

Why not just merge `global-minimal.txt` into `alma10-minimal.txt`, making sure that all flags set by `global.txt` are overridden if necessary?

ferdymercury: Thanks. Letting @pcanal comment on that (I just did a rebase of his work to see where we were standing.)

guitargeek:
Thanks for fixing the post-install tests! Next, we need @pcanal to comment on https://github.com/root-project/root/pull/20725#discussion_r2622226434 to proceed.

pcanal:
> Just rebased version of pcanal's https://github.com/root-project/root/pull/20311

I am a bit confused.  The list of commits is different and at the very least https://github.com/root-project/root/pull/20311/changes/63ddaea3de02911eb9c5e7817137d2c36b889ce0 is missing.

update: some of the commits merged or moot but the above commit was indeed missing.

.github/workflows/root-ci.yml
pcanal: > making sure that all flags set by global.txt are overridden if necessary?

I think that would be defeating the purpose, see the new version of `global.txt` which actually match the semantic of minimal: nothing set except `minimal` and `test` ... with only 2 debatable addition: `ccache` to reduce turn around and ` fail-on-missing` which may or may not be meaningful here.

The options set in `alma10` are to only account for platform dependent behavior (and I guess `vdt` probably should not be there - we need to try to remove it before merging).

pcanal:
I pushed on this PR the missing commit (reducing the size of `global.txt` closer to what it should really be).

PR #20726, opened 2025-12-15
[6.36] CI binaries: include deprecated odbc pgsql mysql

ferdymercury:
A compromise suggested by https://root-forum.cern.ch/t/odbc-and-mysql-connectivity-with-root-6-36-and-beyond/64065/6 would be to leave 6.38/6.40 with completely removed SQL support, but continue providing binaries of 6.36 that use pgsql/mysql for the next 5 years.

This explores what the CI looks like when reenabling those options.


.github/workflows/root-ci-config/buildconfig/mac-beta.txt
```diff
@@ -24,6 +24,8 @@ builtin_zstd=ON
 ccache=ON
 cocoa=ON
 fortran=OFF
+mysql=OFF
+pgsql=OFF
```
guitargeek, line 28: We need this also for `mac26`.

PR #4141, opened 2019-08-01
[TMVA][Preprocessing]

josephmckenna:
Add scaling VarTransform functionality to TMVA preproccessing (like normalisation it linearly scales the data but the sign of the input and output data is retained).

I have added to the functionality of the VariableNormalizeTransform class in the style of the VariableGaussTransform class to transform data such that it remains in the range of [-1,1], there is no offset, so the sign of the input data is unchanged by the transformation. 

This is proving essential for my neural network analyses that treat a detector hit data like an image classification problem and use ReLU activation functions at the beginning of my network.

I have also added a description to the TMVA documentation

PR #20735, opened 2025-12-16
[tmva] Check for existence of xml node before reading it

tmadlener:
# This Pull request:

Adds a check for the existence of `UseOffsetOrNot` fields in TMVA files before trying to read from it.

## Changes or fixes:

This avoids breaking the reading of pre-existing TMVA weight files that have been written without this tag. See also https://github.com/root-project/root/pull/4141#discussion_r2623310131

## Checklist:

- [x] tested changes locally
- [x] updated the docs (not necessary)

I have a small-ish (13K) TMVA weight file that triggered the issue for me originally. Should I add that to some test to ensure that it remains readable? If yes, could you point me to where it should go?

### DIFF
--- a/tmva/tmva/src/VariableNormalizeTransform.cxx
+++ b/tmva/tmva/src/VariableNormalizeTransform.cxx
@@ -377,13 +377,17 @@ void TMVA::VariableNormalizeTransform::AttachXMLTo(void* parent)
 
 void TMVA::VariableNormalizeTransform::ReadFromXML( void* trfnode )
 {
-   TString UseOffsetOrNot;
+   fNoOffset = kFALSE;
+   if (gTools().HasAttr(trfnode, "UseOffsetOrNot")) {
+      TString UseOffsetOrNot;
 
-   gTools().ReadAttr(trfnode, "UseOffsetOrNot", UseOffsetOrNot );
+      gTools().ReadAttr(trfnode, "UseOffsetOrNot", UseOffsetOrNot);
 
-   if (UseOffsetOrNot == "NoOffset") fNoOffset = kTRUE;
-   else                              fNoOffset = kFALSE;
-
+      if (UseOffsetOrNot == "NoOffset")
+         fNoOffset = kTRUE;
+      else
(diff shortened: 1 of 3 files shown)


guitargeek:
Thanks for the PR!

> I have a small-ish (13K) TMVA weight file that triggered the issue for me originally. Should I add that to some test to ensure that it remains readable? If yes, could you point me to where it should go?

Absolutely welcome! The tests go to `tmva/tmva/test`, and are implemented using googletest and declared in CMake with the `ROOT_ADD_GTEST` macro. An example are the unit tests in `TestRandomGenerator.cxx`:
https://github.com/root-project/root/blob/master/tmva/tmva/test/CMakeLists.txt#L13

Maybe you can implement a new regression test, like `TestRegression.cxx` that collects the tests for the cases that regressed at some point, like yours?

guitargeek:
There are some failures on the CI:
```txt
 /github/home/ROOT-CI/src/tmva/tmva/src/VariableNormalizeTransform.cxx: In member function ‘virtual void TMVA::VariableNormalizeTransform::ReadFromXML(void*)’:
 ##[error]/github/home/ROOT-CI/src/tmva/tmva/src/VariableNormalizeTransform.cxx:381:15: error: request for member ‘HasAttr’ in ‘TMVA::gTools’, which is of non-class type ‘TMVA::Tools&()’
     if (gTools.HasAttr(trfnode, "UseOffsetOrNot")) {
   
```

tmadlener:
Fixed the small typo I introduced via the web editor and things should compile now. I have added a very minimal regression test that I confirmed to fail without the fixes. Let me know if you think this is enough, or if the tests should be adapted somehow.

guitargeek:
Thanks a lot! Squash and merge maybe, with the commit message of the second original commit?

andresailer:
I think this needs to be backported to 6.38?

guitargeek:
You're right. Thanks!

guitargeek:
Here is the backport, which will make it into 6.38.02:

  * https://github.com/root-project/root/pull/20867

Issue #20673, opened 2025-12-09
ROOT 6.36.04 does a factor 2 more stats when loading a library than 6.32.08


ktf:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

We see a large number of (unneeded?) stats when trying to load a library. This results in greatly reduced performance on startup to the point it prevents us to start our datataking workflows.

In particular doing:

```
strace -f root -q -e 'gSystem->Load("libO2DataFormatFV0s.so")' 2>&1 | grep BUILD
```

results in 15843 matches while with the old release we have 6049. Notice in both cases none of the files actually exists. So this is really ROOT looking up in some unrelocated path, which itself is wrong.

Moreover it looks like the same stat is done multiple times. In both logs I have patterns like:

```
→ cat stats-new.txt | head -n 100
 newfstatat(AT_FDCWD, "/home/alibuild/sw/BUILD/<build-hash>/ROOT/lib/libc.pcm", 0x7ffd02cf75e0, 0) = -1 ENOENT (No such file or directory)
 newfstatat(AT_FDCWD, "/data/sw/BUILD/<build-hash>/ROOT/etc/cling/lib/clang/18/include", 0x7ffd02cf5fc0, 0) = -1 ENOENT (No such file or directory)
 newfstatat(AT_FDCWD, "/data/sw/BUILD/<build-hash>/ROOT/etc/cling/lib/clang/18/include", 0x7ffd02cf6430, 0) = -1 ENOENT (No such file or directory)
 newfstatat(AT_FDCWD, "/data/sw/BUILD/<build-hash>/ROOT/etc/cling/lib/clang/18/include/stddef.h", 0x7ffd02cf61c0, 0) = -1 ENOENT (No such file or directory)
 newfstatat(AT_FDCWD, "/data/sw/BUILD/<build-hash>/ROOT/etc/cling/lib/clang/18/include/stddef.h", 0x7ffd02cf6630, 0) = -1 ENOENT (No such file or directory)
 newfstatat(AT_FDCWD, "/data/sw/BUILD/<build-hash>/ROOT/etc/cling/lib/clang/18/include/__stddef_size_t.h", 0x7ffd02cf6130, 0) = -1 ENOENT (No such file or directory)
[...]

PR #20736, opened 2025-12-16
[core] Add TGenericClassInfo::AdoptMemberStreamer()

silverweed:
# This Pull request:
adds a new function to `TGenericClassInfo` to allow deferring the registration of a MemberStreamer during static initialization. Performs a similar function to `TGenericClassInfo::AdoptStreamer` et al. but for `TMemberStreamer`s.

## Reason
Currently it's impossible to register a MemberStreamer of a class during static initialization without invoking `TClass::GetClass`, which should never be done at library loading time, as it causes [issues](https://github.com/root-project/root/issues/20673#issuecomment-3640963838).

Specifically, ALICE was hit by this after [adding a GetClass in a static initializer](https://github.com/AliceO2Group/AliceO2/commit/2439bbec037ff4b4c40361cbeaec97fafbaae132#diff-553ae85251c34492bd9adbecd77a43e390e740d98be2e13a4f2b2939a02d256eR68) and as a result observed huge performance regressions due to Cling being initialized earlier than intended (which in turn causes modules to be registered too early).

cc @ktf

### DIFF
--- a/core/meta/inc/TClass.h
+++ b/core/meta/inc/TClass.h
@@ -584,10 +584,6 @@ public:
 
    void               AdoptReferenceProxy(TVirtualRefProxy* proxy);
    void               AdoptStreamer(TClassStreamer *strm);
-   // Returns false if the member streamer could not be adopted (which happens if this class had its StreamerInfo
-   // compiled already).
-   // This function transfers ownership of the `strm` pointer to the TClass, so it should not be used anymore on the
-   // caller side. In particular, if `AdoptMemberStreamer` returns false `strm` has been deleted and becomes invalid.
    bool               AdoptMemberStreamer(const char *name, TMemberStreamer *strm);
    void               SetMemberStreamer(const char *name, MemberStreamerFunc_t strm);
    void               SetStreamerFunc(ClassStreamerFunc_t strm);
--- a/core/meta/inc/TGenericClassInfo.h
+++ b/core/meta/inc/TGenericClassInfo.h
@@ -120,8 +120,6 @@ namespace ROOT {
       TClass                           *IsA(const void *obj);
 
       void                              AdoptAlternate(ROOT::TClassAlt *alt);
-      // Returns false if the member streamer could not be adopted (this happens if the underlying TClass had its
-      // StreamerInfo compiled already).
       bool                              AdoptMemberStreamer(const char *name, TMemberStreamer *strm);
       Short_t                           AdoptStreamer(TClassStreamer*);
       Short_t                           AdoptCollectionProxy(TVirtualCollectionProxy*);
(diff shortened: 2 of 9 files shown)


core/meta/src/TGenericClassInfo.cxx
```diff
@@ -434,6 +438,15 @@ namespace Internal {
       fAlternate.push_back(alt);
    }
 
+   void TGenericClassInfo::AdoptMemberStreamer(const char *name, TMemberStreamer *strm)
+   {
+      auto [it, inserted] = fAdoptedMemberStreamers.emplace(name, strm);
+      if (!inserted) {
+         delete it->second;
+         it->second = strm;
+      }
```
pcanal, line 454:
```suggestion
      if (fClass) {
          assert(fAdoptedMemberStreamers.empty());
          fClass->AdoptMemberStreamer(name.c_str(), strm);
          return;
      }
      auto [it, inserted] = fAdoptedMemberStreamers.emplace(name, strm);
      if (!inserted) {
         delete it->second;
         it->second = strm;
      }
```

See TGenericClassInfo::AdoptStreamer

core/meta/inc/TClass.h
```diff
@@ -584,7 +584,11 @@ friend class TStreamerInfo;
 
    void               AdoptReferenceProxy(TVirtualRefProxy* proxy);
    void               AdoptStreamer(TClassStreamer *strm);
-   void               AdoptMemberStreamer(const char *name, TMemberStreamer *strm);
+   // Returns false if the member streamer could not be adopted (which happens if this class had its StreamerInfo
```
pcanal: This documentation needs to be moved to the source file.

core/meta/inc/TGenericClassInfo.h
```diff
@@ -118,6 +120,9 @@ namespace ROOT {
       TClass                           *IsA(const void *obj);
 
       void                              AdoptAlternate(ROOT::TClassAlt *alt);
+      // Returns false if the member streamer could not be adopted (this happens if the underlying TClass had its
```
pcanal: Move doc to source.

roottest/root/io/customStreamer/testMemberStreamer2.C
```diff
@@ -0,0 +1,74 @@
...
+}
+
+bool checkMemberStreamer(int expectedDataValue)
+{
+   {
+      MemberStreamerContainer container{};
+      TFile file("testMemberStreamer.root", "RECREATE");
+      file.WriteObject(&container, "container");
+      // custom streamer should have modified the data
+      if (container.fMember->fData != expectedDataValue) {
```
pcanal, line 79: From my reading, it seems that this would not detect if the streamer function was not used but the default behavior was used instead, or am I missing something?

silverweed: `fData` can only be equal to `expectedDataValue` if the custom streamer was called, otherwise it'd always be initialized to 0

PR #20739, opened 2025-12-16
[docu,math,RF] more CINT to cling renames

ferdymercury:
subpart of work towards https://github.com/root-project/root/issues/10058

### DIFF
--- a/roofit/roofitcore/inc/RooAbsData.h
+++ b/roofit/roofitcore/inc/RooAbsData.h
@@ -176,7 +176,7 @@ public:
 
   virtual RooPlot* plotOn(RooPlot* frame, const RooLinkedList& cmdList) const ;
 
-  // WVE --- This needs to be public to avoid CINT problems
+  // WVE --- This needs to be public to avoid Cling problems
   struct PlotOpt {
    const char* cuts = "";
    Option_t* drawOptions = "P";
--- a/roofit/roofitcore/inc/RooFactoryWSTool.h
+++ b/roofit/roofitcore/inc/RooFactoryWSTool.h
@@ -82,9 +82,9 @@ public:
   std::vector<std::string> splitFunctionArgs(const char* funcExpr) ;
 
 
-  // --- Internal stuff that must be public so that CINT can access it ---
+  // --- Internal stuff that must be public so that Cling can access it ---
 
-  // CINT constructor interface
+  // Cling constructor interface
   static RooAbsArg& as_ARG(UInt_t idx) { checkIndex(idx) ; return of()->of()->asARG(of()->_args[idx].c_str()) ; }
 
   static RooAbsPdf& as_PDF(UInt_t idx) { checkIndex(idx) ; return of()->asPDF(of()->_args[idx].c_str()) ; }
@@ -178,7 +178,7 @@ protected:
 
   TClass* resolveClassName(const char* className) ;
 
-  // CINT constructor interface back end
+  // Cling constructor interface back end
   static RooFactoryWSTool* of() ;
   static RooFactoryWSTool* _of ;
   std::vector<std::string> _args ;
--- a/roofit/roofitcore/inc/RooWorkspace.h
+++ b/roofit/roofitcore/inc/RooWorkspace.h
@@ -271,8 +271,8 @@ public:
 
     std::unique_ptr<RooFactoryWSTool> _factory; ///<! Factory tool associated with workspace
 
-    bool _doExport;          ///<! Export contents of workspace to CINT?
-    std::string _exportNSName; ///<! Name of CINT namespace to which contents are exported
+    bool _doExport;          ///<! Export contents of workspace to Cling?
+    std::string _exportNSName; ///<! Name of Cling namespace to which contents are exported
 
     bool _openTrans = false; ///<! Is there a transaction open?
     RooArgSet _sandboxNodes; ///<! Sandbox for incoming objects in a transaction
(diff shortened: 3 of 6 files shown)


math/mathcore/inc/LinkDef1.h
```diff
@@ -35,7 +35,7 @@
 #pragma link C++ function TMath::Limits<Int_t>::Max();
 #pragma link C++ function TMath::Limits<Int_t>::Min();
 
-// exclude these since they do not work in CINT
+// exclude these since they do not work in Cling
```
pcanal, line 38: That is probably an inaccurate statement (i.e. those functions are very likely to work with Cling.).
On the other hand, currently all the `#pragma link C++ function` are ignored (if I recall correctly).

math/mathmore/src/GSLNLSMinimizer.cxx
```diff
@@ -134,7 +134,7 @@ class FitTransformFunction : public FMFunc {
 template <class Func>
 class LSResidualFunc : public IMultiGradFunction {
 public:
-   // default ctor (required by CINT)
+   // default ctor (required by Cling)
```
pcanal, line 137: It is unlikely to be required by Cling (it would have to mean it is required per the C++ standard).    And it does not appear to be a persistent type, so it would not be required by the I/O either ... i.e. I am not sure if it is still required ...

math/mathmore/test/testGSLRootFinder.cxx
```diff
@@ -153,7 +153,7 @@ int testGSLRootFinder() {
 
 
 
-  // the following two examples won't work when interpreted CINT
+  // the following two examples won't work when interpreted Cling
```
pcanal, line 156: This is unlikely to be true for Cling.  i.e. I bet the example works fine in Cling ...

math/matrix/doc/index.md
```diff
@@ -491,7 +491,7 @@ class MakeHilbert is declared *within* a function and local to
 that function. It means one can define another MakeHilbert class
 (within another function or outside of any function, that is, in
 the global scope), and it still will be OK. Note, this currently
-is not yet supported by the interpreter CINT.
+is not yet supported by the interpreter Cling.
```
pcanal, line 494: This is likely to be supported by Cling.

math/smatrix/doc/Functions.md
```diff
@@ -105,5 +105,5 @@ std::cout << v << std::endl;
 ~~~
 
 
-In the ROOT distribution, the CINT dictionary is generated for SMatrix and SVector for double types and sizes up to 5\. This allows the storage of them in a ROOT file.
+In the ROOT distribution, the Cling dictionary is generated for SMatrix and SVector for double types and sizes up to 5\. This allows the storage of them in a ROOT file.
```
pcanal, line 108:
```suggestion
In the ROOT distribution, the I/O dictionary is generated for SMatrix and SVector for double types and sizes up to 5\. This allows the storage of them in a ROOT file.
```

roofit/roofitcore/inc/RooAbsData.h
```diff
@@ -176,7 +176,7 @@ class RooAbsData : public TNamed, public RooPrintable {
 
   virtual RooPlot* plotOn(RooPlot* frame, const RooLinkedList& cmdList) const ;
 
-  // WVE --- This needs to be public to avoid CINT problems
+  // WVE --- This needs to be public to avoid Cling problems
```
pcanal, line 179: More likely than not (to be verified though), this what is now meant.
```suggestion
  // WVE --- This needs to be public to allow I/O
```
i.e. To be storable a class need to be public. (but is `PlotOpt` meant to be stored?)

roofit/roofitcore/inc/RooFactoryWSTool.h
```diff
@@ -82,9 +82,9 @@ class RooFactoryWSTool : public TNamed, public RooPrintable {
   std::vector<std::string> splitFunctionArgs(const char* funcExpr) ;
 
 
-  // --- Internal stuff that must be public so that CINT can access it ---
+  // --- Internal stuff that must be public so that Cling can access it ---
```
pcanal, line 85: This is probably imprecise.  (I.e. Cling does not need access to it per se ... but something else in the interactive layer might).

guitargeek:
Thanks for your comments @pcanal!

You're right, there are some inconsistencies in some internal comments now, but there are so many of them already that I didn't really want to bother @ferdymercury with them in this PR, which does the more automatic replacements. Also I don't think we gain much by fixing up those.

But as you took already the time to make concrete suggestions, we will certainly address them.

roofit/roofitcore/inc/RooAbsData.h
guitargeek: Doesn't need to be stored. It's only used as argument for `protected` methods. Probably that was the reason.

guitargeek:
Follow-up PR: https://github.com/root-project/root/pull/20745

pcanal:
> but there are so many of them already that I didn't really want to bother @ferdymercury with them in this PR, which does the more automatic replacements. 

We intentionally did *not* automatically replace all those CINT leftover when we did the v6 migration because they needed to be looked at to make sense.  If we replace them automatically we will no longer be able to find/detect the mention of CINT that actually need to be reviewed (eg. as noted in this PR a few simplification can be done and in a few case the resulting text is inaccurate).

Issue #20731, opened 2025-12-16
Investigate/Remove potentially unnecessary explicit linking


guitargeek:
**Was [ROOT-9874](https://its.cern.ch/jira/browse/ROOT-9874) on JIRA, but it's good to make this issue public, as it's a good first issue for new contributors. No ROOT-specific knowledge is required, just some CMake.**

According to `ldd -u`/`ldd --unused` we are (might be) sometimes explicitly linking too many libraries (see the list below for examples of how the output can look like for some libraries).

To address this issue for good, one would first have to check for unused dependecies in each of the shared libraries in the build directory. This is best done by writing a script. If there are unused dependencies, there are two general cases that one has to address:
  1. The unused dependency is linked directly in the library itself (via `target_link_libraries` or the `LIBRARIES` or `DEPENDENCIES` arguments if the library is declared with ROOT macros like `ROOT_STANDARD_LIBRARY_PACKAGE`). Then it just has to be removed
  2. The unused dependency comes indirectly from another dependency, which unnecessarily linked some of its own dependencies with the `PUBLIC` property. In that case, one has to identify where this indirect dependency comes from, and then make sure that the dependency is liked privately there (via the `PRIVATE` property of `target_link_libraries`, or by appending to the `LIBRARIES` instead of `DEPENDENCIES` argument when a ROOT macro is used. The mapping is `PUBLIC` -> `DEPENDENCIES` and `PRIVATE` -> `LIBRARIES`).

```txt
libCling.so:
Unused direct dependencies:
libcomplexDict.so:
Unused direct dependencies:
libCore.so:
Unused direct dependencies:
libdequeDict.so:
Unused direct dependencies:
libEG.so:
Unused direct dependencies:
	/home/pcanal/root_builds/master/gcc8/lib/libGraf3d.so
	/home/pcanal/root_builds/master/gcc8/lib/libGpad.so
	/home/pcanal/root_builds/master/gcc8/lib/libGraf.so
	/home/pcanal/root_builds/master/gcc8/lib/libHist.so
	/home/pcanal/root_builds/master/gcc8/lib/libMatrix.so
	/home/pcanal/root_builds/master/gcc8/lib/libRIO.so
	/home/pcanal/root_builds/master/gcc8/lib/libThread.so
libFitPanel.so:
Unused direct dependencies:
	/home/pcanal/root_builds/master/gcc8/lib/libGraf.so
	/home/pcanal/root_builds/master/gcc8/lib/libMatrix.so
[...]

PR #20740, opened 2025-12-16
[AI SLOP] Add script to detect unused library dependencies

ahmedbilal9:
This PR adds a Python script to help identify unused library dependencies in ROOT's shared libraries. 

## What does this script do?
- Scans all `.so` files in the ROOT build directory
- Uses `ldd -u` to detect libraries that are linked but not actually used
- Generates a report showing which libraries have unused dependencies

## Why is this useful?
ROOT has many shared libraries with complex dependencies. Over time, some libraries may link to dependencies they don't actually use, which wastes resources and increases build times.  This script helps identify these cases so they can be fixed. 

## Usage
```bash
python3 check_unused_dependencies.py /path/to/root/build

### DIFF
--- a/UNUSED_DEPENDENCIES_CHECK.md
+++ b/UNUSED_DEPENDENCIES_CHECK.md
@@ -1,0 +1,20 @@
+# Checking for Unused Library Dependencies in ROOT
+
+This document describes how to identify and fix unused library dependencies in ROOT's CMake build system. 
+
+## Problem
+
+ROOT's shared libraries sometimes link to other libraries they don't actually use. This happens when:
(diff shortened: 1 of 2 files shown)


guitargeek:
Hi @ahmedbilal9, thanks for the PR :slightly_smiling_face: I guess you want to address #20731?

We don't need to add scripts to check for unused libraries to the repository. It's trivial to write such a script (or let AI do it...). The contribution we are looking for is actually removing the unused link libraries from our CMake code.

guitargeek (changes requested):
Goes in the wrong direction

PR #20745, opened 2025-12-17
[RF][Math][TMVA] Follow up on CINT -> Cling replacement in comments

guitargeek:
Follows up on #20739 with some improvements that became apparent now.


ferdymercury (approved):
Thanks a lot!

There were two more places recently where I had doubts in case you wanna to take a look:

`bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx: return (intptr_t)m->GetOffsetCint(); // yes, CINT (GetOffset() is both wrong`
`bindings/tpython/src/TPyReturn.cxx:/// you have to use the void* converter, as CINT will not call any other).`

guitargeek:
Sorry, I don't know yet what is right in the cases that you point out :slightly_frowning_face: Better to deal with them in separate PRs

pcanal (approved):
Thanks.

PR #20748, opened 2025-12-17
[main] optparse: allow multiple repeated options

silverweed:
# This Pull request:
Allows specifying that an option may appear multiple times in `optparse.hxx`, and allows retrieving all its values.

### DIFF
--- a/main/src/optparse.hxx
+++ b/main/src/optparse.hxx
@@ -41,7 +41,13 @@
 /// grouped flags like "-abc" as equivalent to "-a -b -c". The last flag in the group may also accept an argument, in
 /// which case "-abc foo" will count as "-a -b -c foo" where "foo" is the argument to "-c".
 ///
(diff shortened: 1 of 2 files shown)


dpiparo:
Thanks. Could you perhaps give some details relative to the use case that this feature will unblock?

silverweed:
> Thanks. Could you perhaps give some details relative to the use case that this feature will unblock?

It's to investigate if we can use this option parsing to replace rootcling's current one, which has some issues related to the LLVM option parsing code (@hahnjo and @devajithvs have more details, but it's something we're gonna look at next year - no need to merge it now)

devajithvs (approved):
LGTM. Thanks.

As disussed offline, to make `rootcling` behave the same way as it is now, we would also need support for `Prefix options` (eg, -I/something, -compilerI/usr/include/c++/13). 

But we can also decide not to support them in `rootcling` and simply require a space after the option.

PR #20749, opened 2025-12-17
[ci] PRs: Check for fork branch name

### LINKED ISSUE #20743: [ci] PR branch not checked out in CI tests if it's called `master`
I see it quite often that first-time contributors or people who contribute rarely use the `master` branch for development and opening their PR.

The problem is that the branch is not correctly checked out for the CI tests then, so the CI just runs and succeeds with the current `root-project:master`! That's bad, because one thinks that everything is fine, the PR is merged, and then one sees compilation or test failures only in the incremental of the nightly builds.

One can tell that this is the case from the log, e.g. in #20708:
```txt
git fetch origin 94d9c9d7c9c6215bfe72fab4c14db10ecb07bbe0:master
git checkout master
git rebase 2f1d165234bc718600f6f3a590573fc6bb727ac8
�[0m
�[90mfatal: refusing to fetch into branch 'refs/heads/master' checked out at '/github/home/ROOT-CI/src'
Already on 'master'
Your branch is up to date with 'origin/master'.
Current branch master is up to date.
�[0m
** Elapsed time for group "Rebase" 0:00:00.4
```

This came to bite us three times recently, breaking the builds because the CI tests were not run with the actual commit that was supposed to be tested:
  * https://github.com/root-project/root/pull/20708
  * https://github.com/root-project/root/pull/20358
  * https://github.com/root-project/root/pull/20276

Every time, the name of the PR branch was `master`.

dpiparo:
if identical to the name of the base-ref, error out.

Fixes #20743

This is a redo of https://github.com/root-project/root/pull/20747, that demonstrates that the protection mechanism works.

### DIFF
--- a/.github/workflows/root-ci-config/build_root.py
+++ b/.github/workflows/root-ci-config/build_root.py
@@ -54,7 +54,18 @@ def main():
 
     pull_request = args.head_ref and args.head_ref != args.base_ref
 
-    if not pull_request:
+    if pull_request:
+        # We check whether the PR is done from a branch that has the same name of
+        # the target branch in the ROOT repository using base_ref and head_ref:
+        # - The base_ref is e.g. "master" or "v6-38-00-patches" 
+        # - The head_ref for PRs is formatted as <PR branch>:<Fork branch> e.g. "refs/pull/20742/head:cppyy_fixup"
+        if not ":" in args.head_ref:
+            build_utils.print_error(f"This has been identified as a PR build. However, the head-ref is {args.head_ref}.")
+        branch_in_fork = args.head_ref.split(":")[-1]
+        if branch_in_fork == args.base_ref:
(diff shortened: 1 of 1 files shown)


dpiparo:
we do not need all builds... one is enough

.github/workflows/root-ci-config/build_root.py
```diff
@@ -54,7 +54,18 @@ def main():
 
     pull_request = args.head_ref and args.head_ref != args.base_ref
 
-    if not pull_request:
+    if pull_request:
+        # We check whether the PR is done from a branch that has the same name of
+        # the target branch in the ROOT repository using base_ref and head_ref:
+        # - The base_ref is e.g. "master" or "v6-38-00-patches" 
+        # - The head_ref for PRs is formatted as <PR branch>:<Fork branch> e.g. "refs/pull/20742/head:cppyy_fixup"
+        if not ":" in args.head_ref:
```
pcanal, line 62: Ruff recommend 
```suggestion
        if ":" not in args.head_ref:
```
as more readable.

PR #20751, opened 2025-12-18
Fixes for use of internal, low-level compression functions

### LINKED ISSUE #20750: Insufficient checks around R__unzip_header
### Check duplicate issues.

- [x] Checked for duplicates

### Description

ROOT compression blocks are typically unzipped in a loop where every iteration parses the compression block header followed by a call to R__unzip.  Generally, the following checks are currently missing

 - Before calling `R__unzip_header`, check that the input buffer size is at least as along as the header (9 bytes)
 - After calling `R__unzip_header` and before calling `R__unzip`, check that the reported compressed and decompressed block size fit into the remaining enclosing input and output buffer sizes.

Many thanks to Pavel Kohout from AISLE for the original report!

### Reproducer

Reproducible through corrupted / crafted source and target sizes in the compression header.

### ROOT version

master and previous

### Installation method

n/a

### Operating system

n/a

### Additional context

_No response_

jblomer:
Fixes #20750 

PR Status:

- [x] unit test infrastructure to test zip header corruption
- [x] fix unzipping in RNTuple
- [x] fix unzipping in TKey
- [x] fix unzipping in TBasket
- [x] fix unzipping in TTreeCache
- [x] fix unzipping in TMessage
- [x] fix unzipping in TBufferXML


jblomer:
@pcanal: Regarding `TBufferXML`:  do you by chance know how `XmlWriteBlock` and `XmlReadBlock` is reached?  Trying this with a `TXMLFile`, `XmlWriteBlock` backs out immediately because `Length()` is zero.

pcanal:
> Regarding TBufferXML: do you by chance know how XmlWriteBlock and XmlReadBlock is reached? Trying this with a TXMLFile, XmlWriteBlock backs out immediately because Length() is zero.

@linev might be able to confirm but it looks like it is currently not used per se.  (In our tests, it is being called but, as you mentioned, it returns immediately).

linev:
`XmlWriteBlock` and `XmlReadBlock` are invoked only when binary data directly written into the buffer.
Such data may appear in custom class streamers.
For the moment there are no such classes in ROOT itself.

pcanal:
> XmlWriteBlock and XmlReadBlock are invoked only when binary data directly written into the buffer.
> Such data may appear in custom class streamers.
> For the moment there are no such classes in ROOT itself.

Would it make sense to add a test for this use case? (and/or deprecate it?)

jblomer:
> > XmlWriteBlock and XmlReadBlock are invoked only when binary data directly written into the buffer.
> > Such data may appear in custom class streamers.
> > For the moment there are no such classes in ROOT itself.
> 
> Would it make sense to add a test for this use case? (and/or deprecate it?)

To me it looks safe to remove these functions: the documentation of `XmlWriteBlock()` says "This data can be produced only by direct call of TBuffer::WriteBuf() functions.". The `WriteBuf()` abstract method is implemented in `TBufferText` and emits an `Error()`.

jblomer:
@pcanal @linev this is ready for review.

linev:
@jblomer 

You completely delete code which writes binary data in TBufferXML.

> For the moment there are no such classes in ROOT itself.

My remark was about ROOT code itself, but not about users code.
If by chance users have `WriteBuf()/ReadBuf()` in there custom streamers - than streaming will no longer works for them.

I must admit - such binary data support is not implemented in JSON. 

So I am not sure if we need to keep support of such custom binary data in XML output.

linev (commented):
I let Philippe approve PR while main changes are in the unzipping of TKey data,
Probably I would name method `TKey::UnzipPayload()`

io/io/inc/TKey.h
```diff
@@ -35,6 +35,8 @@ class TKey : public TNamed {
    TKey(const TKey&) = delete;            // TKey objects are not copiable.
    TKey& operator=(const TKey&) = delete; // TKey objects are not copiable.
 
+   Int_t UnzipObject(char *targetBuffer, const char *compressedBuffer) const;
```
pcanal:
```suggestion
   Int_t UnzipBuffer(char *targetBuffer, const char *compressedBuffer) const;
```
The buffer may contain an object or multiple objects or just numbers, etc.

io/io/src/TKey.cxx
```diff
@@ -382,6 +382,33 @@ TKey::TKey(const void *obj, const TClass *cl, const char *name, Int_t bufsize, T
    }
 }
 
+/// Core of uncompressing the key payload. Returns number of bytes uncompressed.
```
pcanal, line 385: We should document the mechanism (currently done by the caller) by which we 'guarantee' that we don't overflow the `targetBuffer`.

PR #20753, opened 2025-12-18
[PyROOT] Don't install the python modules twice

ellert:
# This Pull request:

## Changes or fixes:

The binary python modules were moved one directory level down into the ROOT and cppyy directories, so that they reside alongside the python files in the same directories. The install rule for the python files must therefore be extended to exclude the binary modules (`PATTERN *.so EXCLUDE`) so that these are not installed a second time overwriting the files installed by the install rule for the binary modules with another copy with the wrong file permissions.

Also use the no-version option for the cppyy binary module as is already done for the pythonization binary module, since it is now installed in the python directory. Before it was installed in the library directory with an unversioned symlink in the python directory.

## Checklist:

- [x] tested changes locally


guitargeek (approved):
Thank you, very nice!

PR #20757, opened 2025-12-18
Address some minor TODOs after CINT to Cling migration

guitargeek:
Follows up on the observations made in #20694.

I also understand now that the `TGQuartz.h` header is actually needed by the ROOT plugin system.


pcanal (approved):
Thanks.

PR #20762, opened 2025-12-19
[roottest] Skip `rootcling` instead of `rootcint` from output

guitargeek:
Now that we use `rootcling` instead of `rootcint` in Make project, this
needs to be updated.

Follows up on https://github.com/guitargeek/root/commit/e421de683b2f31b6bdf11378c4c0097322d8dbab.


bellenot (approved):
Thanks Jonas!

Issue #20602, opened 2025-12-02
TGTextEntry spills text out of the text box on macOS but not on Ubuntu


zoglauer:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

The problem can easily reproduced with the guitest.C tutorial:
Open Test --> Dialog and enter a long string. Under macOS it will overflow out of the text box (bug), in Linux it will stay contained in the box.

<img width="291" height="245" alt="Image" src="https://github.com/user-attachments/assets/0a49e7fe-2cf6-4873-927b-48c4d216f5ef" />

### Reproducer

See description.

### ROOT version

  | Welcome to ROOT 6.36.04                        https://root.cern |
  | (c) 1995-2025, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for macosxarm64 on Nov 19 2025, 22:35:48                   |
  | From tags/6-36-04@6-36-04                                        |
  | With Apple clang version 17.0.0 (clang-1700.4.4.1)               |

### Installation method

built from source

### Operating system

macOS Sequoia, Ubuntu 24.04

couet:
We asked @TimurP (the author of this code) to look at it. He can reproduce the issue and will let us know if he finds something.

dpiparo:
Thanks @TimurP !

PR #20764, opened 2025-12-19
Fix a TGTextEntry issue with Cocoa

couet:
This PR fixes https://github.com/root-project/root/issues/20602
Thanks @TimurP for the fix !


dpiparo:
Thanks a lot for this fix, @TimurP . I already backported to the release series Core Support.

PR #20767, opened 2025-12-20
[ci] Make minimal alma10 build a special build and use at least C++20 for all special builds

guitargeek:
Follows up on #20725.

By raising the C++ standard of all special builds to at least C++20, we can test also ROOT features exclusive to C++ 20 if applicable. This could include for example usage of `std::experimental::simd`. There is no effect on the binaries provided to our users, as the special builds are not used to produce binaries.

### DIFF
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -398,20 +398,20 @@ jobs:
           - image: alma9
             is_special: true
             property: modules_off
-            overrides: ["runtime_cxxmodules=Off"]
+            overrides: ["runtime_cxxmodules=Off", "CMAKE_CXX_STANDARD=20"]
           - image: alma9
             is_special: true
             property: march_native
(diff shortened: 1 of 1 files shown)


guitargeek:
Yes, I think that's a very good idea! My suggestion would be to have some mechanism to test different platforms in the nightlies and in the PR CI builds, so that we can slim down the PR builds a bit without losing platform support.

Then, the first candidate to drop in the PR builds would be for me the previous-to-last Fedora version now **Fedora 42**, because it is between Alma 10 and the newer Fedora 43. It's pretty difficult to break only these sandwiched versions with some changes in ROOT. Either, you break something before a certain version, or after, but it's rare to break only a specific version in between. The same logic would kick out **macOS 15** and **Ubuntu 24.04** from the PR builds, if we want to follow that. And technically also Alma 9, but the Alma Linuxes are so far apart and our main supported platforms. So I would not go that far.

PR #20781, opened 2025-12-27
[tmva][sofie][tutorials]Fix model parameters and cleanup

shambhvi2006:
This PR aligns CreateModel parameters with the tutorial model names (50/200/500)
and makes the cleanup step more robust by checking file existence before removal.


guitargeek:
Thanks! The update with the file removal where absorbed into c3dae66022fcef1b9d, and the "wrong" small model sizes are actually on purpose to keep the tutorial runtime under control, so let's not change that (the tutorial are also executed as unit tests in every CI run).

PR #20787, opened 2025-12-30
FANG event Generator

itay-space:
# This Pull request:

Adds the Focused Angular N-body event Generator (FANG) to the `montecarlo` directory.

## Changes or fixes:

* **New Feature:** Implemented FANG, a Monte Carlo tool for efficient event generation in restricted (or full) Lorentz-Invariant Phase Space (LIPS).
* Unlike conventional approaches that always sample the full 4pi solid angle, FANG can directly generate events in which selected final-state particles are constrained to fixed directions or finite angular regions in the laboratory frame.
* **Files Added:**
    * Source code in `montecarlo/fang/src` and headers in `montecarlo/fang/inc`.
    * Tests in `montecarlo/fang/test`.
    * Tutorials in `tutorials/evegen` (`runFANG.C`, `runParallelFANG.C`).
* **Authors:** Arik Kreisel and Itay Horin.
* **Reference:** Horin, I., Kreisel, A. & Alon, O. *Focused angular N-body event generator (FANG)*. J. High Energ. Phys. 2025, 137 (2025). https://doi.org/10.1007/JHEP12(2025)137
## Checklist:

- [x] tested changes locally (Test suite included in `montecarlo/fang/test/testFANG.cxx`)
- [x] tutorials added

### DIFF
--- a/tutorials/evegen/runTFANG_parallel.C
+++ b/tutorials/evegen/runTFANG_parallel.C
@@ -648,7 +648,7 @@ void runTFANG()
       v3.SetXYZ(detPosX[i], detPosY[i], detPosZ[i]);
       v3Det.push_back(v3);
       radius = TMath::Sqrt(v3.Mag2() + detRadius[i] * detRadius[i]);
-      omega[i] = kTwoPi * radius * (radius - v3.R());
+      omega[i] = kTwoPi * (1-v3.R()/radius);  
       shape[i] = 0.0;
       totalOmega *= omega[i];
    }


hahnjo:
Hi @itay-space, thanks for opening the PR. Can you maybe share a bit the motivation behind implementing this event generator? In particular can you share what is the advantage of integrating this code in ROOT directly, since it seems that `FANG.h` consists only of free functions. I'm asking because in the past years we've been gradually deprecating and removing `montecarlo` subpackages, so I think we should understand if this new code is something we should commit to maintain...

arikk40:
Dear @hahnjo,

We believe that a tool for kinematic calculations of the outgoing four momenta from a particle decay or a reaction was and is one of the basic fundamental tools for particle and nuclear physicists to do their analysis or MC calculations with root.

FANG aims to replace the currently available tool, TGenPhaseSpace, which has become a legacy code. 

FANG improves the basic functions done by TGenPhaseSpace, and adds new and novel possibilities that will be available for the first time.
In its basic use FANG does exactly what TGenPhaseSpace did but with a very useful improvement, FANG sum of weights gives the exact value of the  Lorentz-Invariant Phase Space.

What FANG can do that no other code does, is allow the user to constrain the outgoing particles to eject either to a point or a restricted solid angle in the Lab frame. Again giving as output the value of the constrained phase space.

Integrating this code directly to root will enable root users an easy access and use of an important and fundamental tool for their analysis.

It will not be a problem for us to change the code to be written as a class rather than free functions. We can also move FANG to the Math directory or any other location you think is a more appropriate.

Best regards,
                          Arik & Itay @itay-space

ferdymercury:
> TGenPhaseSpace, which has become a legacy code.

in your PR, could you replace in https://root.cern.ch/doc/master/TGenPhaseSpace_8cxx_source.html#l00005

"No alternative classes are currently available."
with
"Consider using FANG instead"
or sth like that.

itay-space:
Hi @ferdymercury,

Thanks for the suggestion! I have updated the `\legacy` note in `math/physics/src/TGenPhaseSpace.cxx`.

arikk40:
Hello, @hahnjo, @bellenot, @dpiparo @couet, @ferdymercury ,

We changed the FANG code to run with class TFANG rather than free functions.
We moved the code from montecarlo to math/physics near TGenPhaseSpace .
We tested that the test and tutorials work correctly.

Arik & Itay @itay-space

itay-space:
Hi @dpiparo @bellenot @couet , just checking in on this. We believe the recent changes (moving to math/physics and class-based implementation) address the architectural concerns. Is there anything else needed before we can proceed? Thanks!

PR #20791, opened 2026-01-02
[tmva][sofie] Guard old Keras parser against Keras-3

### LINKED ISSUE #20591: [tmva][sofie] SOFIE Keras parser doesn't support Keras 3
[Keras 3](https://pypi.org/project/keras/3.0.0/) was released in November 2023, and is until now unsupported by TMVA SOFIE. It would be good to support it, because Keras 3 is also part of TensorFlow since [TensorFlow 2.16.1](https://pypi.org/project/tensorflow/2.16.1/) from March 2024.

In our CI, we don't see any failures because we only activate the SOFIE Keras tests and tutorials when Keras 2 is found. But our users will get segfaults when running the SOFIE Keras tutorials with TensorFlow 2.16.1+. This is shown in PR https://github.com/root-project/root/pull/20471, which removes the Keras version check to disable failing tests and tutorials for a trial CI run. The following failures appear:

```txt
TEST FAILURES:
498:gtest-tmva-sofie-TestRModelParserKeras
977:tutorial-machine_learning-TMVA_SOFIE_Keras
1424:tutorial-machine_learning-TMVA_SOFIE_Models-py
```
All three tests/tutorials fail with a segmentation violation with `TMVA::Experimental::SOFIE::PyKeras::INTERNAL::AddKerasLayer()` at the end of the stack trace.

This issue can be closed when the tests pass for all Keras versions on our CI, and no tests and tutorials have to be disabled conditionally on the Keras version.

Possibly fixed by:
  * https://github.com/root-project/root/pull/19692

Shlok-Saxena:
# This Pull request
Adds a defensive runtime guard to the TMVA SOFIE Keras parser to prevent segmentation faults when used with Keras 3 (enabled by default in TensorFlow ≥ 2.16).

The current parser is implemented against the Keras 2 API. When Keras 3 objects are encountered, undefined behavior may occur, leading to runtime crashes. This PR does not add Keras 3 support; it only detects the unsupported configuration early and fails cleanly with a clear error message.

This provides a temporary safety measure while proper Keras 3 support is being developed.

## Changes or fixes:
- Add an explicit runtime check for the Keras major version
- Disable the old SOFIE Keras parser when Keras ≥ 3 is detected
- Replace segmentation faults with a clear, user-facing error message
- Do not change behavior for Keras 2 users

## Checklist:
- [x] tested changes locally

This PR fixes #20591

### DIFF
--- a/tmva/sofie_parsers/src/RModelParser_Keras.cxx
+++ b/tmva/sofie_parsers/src/RModelParser_Keras.cxx
@@ -873,8 +873,16 @@ RModel Parse(std::string filename, int batch_size){
    // For each layer: type,name,activation,dtype,input tensor's name,
    // output tensor's name, kernel's name, bias's name
    // None object is returned for if property doesn't belong to layer
-   PyRunString("import tensorflow",fGlobalNS,fLocalNS);
-   PyRunString("import tensorflow.keras as keras",fGlobalNS,fLocalNS);
+   PyRunString("import tensorflow\n", fGlobalNS, fLocalNS);
+   PyRunString("import tensorflow.keras as keras\n"
+               "version = keras.__version__\n"
+               "major = int(version.split('.')[0])\n"
+               "if major >= 3:\n"
+               "    raise RuntimeError(\n"
+               "        'TMVA SOFIE Keras parser supports Keras 2 only.\\n'\n"
+               "        'Keras 3 detected. Please export the model to ONNX.\\n'\n"
+               "    )\n",
+               fGlobalNS, fLocalNS);
    PyRunString("from tensorflow.keras.models import load_model",fGlobalNS,fLocalNS);
    PyRunString("print('TF/Keras Version: '+ tensorflow.__version__)",fGlobalNS,fLocalNS);
    PyRunString(TString::Format("model=load_model('%s')",filename.c_str()),fGlobalNS,fLocalNS);


tmva/sofie_parsers/src/RModelParser_Keras.cxx
```diff
@@ -832,157 +835,171 @@ std::unique_ptr<ROperator> MakeKerasIdentity(PyObject* fLayer)
...
-   PyRunString("from tensorflow.keras.models import load_model",fGlobalNS,fLocalNS);
-   PyRunString("print('TF/Keras Version: '+ tensorflow.__version__)",fGlobalNS,fLocalNS);
-   PyRunString(TString::Format("model=load_model('%s')",filename.c_str()),fGlobalNS,fLocalNS);
-   PyRunString(TString::Format("model.load_weights('%s')",filename.c_str()),fGlobalNS,fLocalNS);
-   PyRunString("globals().update(locals())",fGlobalNS,fLocalNS);
-   PyRunString("modelData=[]",fGlobalNS,fLocalNS);
+   PyRunString("import tensorflow", fGlobalNS, fLocalNS);
+   PyRunString("import tensorflow.keras as keras", fGlobalNS, fLocalNS);
+   PyRunString("import keras\n"
+               "version = keras.__version__\n"
```
guitargeek:
```suggestion
   PyRunString("version = keras.__version__\n"
```
`keras` is already imported before as `tensorflow.keras`, so you don't need to import it again I think.

Shlok-Saxena: Thanks a lot for the review!

I’ve removed the clang-format commit, dropped the unrelated initial commit, and removed the redundant `import keras`, keeping only `tensorflow.keras`.

The PR now contains only the intended Keras 3 runtime guard logic. Please let me know if any further changes are needed.

PR #20794, opened 2026-01-04
[modules] Remove headers deprecated in C++17 and non-standard in C++20 from the module maps

guitargeek:
Remove the headers `ccomplex`, `cstdalign`, `cstdbool`, and `ctgmath` from the module maps. They are deprecated since C++17 and should not be included anymore by ROOT or its users, as C++17 is the minimum supported C++ standard.

This fixes build warnings that can be seen in the CI when building with GCC 15, for example on Fedora 42:

```txt
[ 47%] Generating G__Core.cxx, ../lib/Core.pcm
While building module 'Core':
While building module 'std' imported from input_line_1:1:
In file included from <module-includes>:5:
/usr/lib/gcc/x86_64-redhat-linux/15/../../../../include/c++/15/ccomplex:51:4: warning: "<ccomplex> is deprecated in C++17, use <complex> instead" [-W#warnings]
 # warning "<ccomplex> is deprecated in C++17, use <complex> instead"
   ^
While building module 'Core':
While building module 'std' imported from input_line_1:1:
In file included from <module-includes>:23:
/usr/lib/gcc/x86_64-redhat-linux/15/../../../../include/c++/15/cstdalign:50:6: warning: "<cstdalign> is deprecated in C++17, remove the #include" [-W#warnings]
 #   warning "<cstdalign> is deprecated in C++17, remove the #include"
     ^
While building module 'Core':
While building module 'std' imported from input_line_1:1:
In file included from <module-includes>:25:
/usr/lib/gcc/x86_64-redhat-linux/15/../../../../include/c++/15/cstdbool:50:6: warning: "<cstdbool> is deprecated in C++17, remove the #include" [-W#warnings]
 #   warning "<cstdbool> is deprecated in C++17, remove the #include"
     ^
While building module 'Core':
While building module 'std' imported from input_line_1:1:
In file included from <module-includes>:30:
/usr/lib/gcc/x86_64-redhat-linux/15/../../../../include/c++/15/ctgmath:50:6: warning: "<ctgmath> is deprecated in C++17, use <complex> or <cmath> instead" [-W#warnings]
 #   warning "<ctgmath> is deprecated in C++17, use <complex> or <cmath> instead"
```

### DIFF
--- a/interpreter/cling/include/cling/std.modulemap
+++ b/interpreter/cling/include/cling/std.modulemap
@@ -44,6 +44,7 @@ module "std" [system] {
     header "cinttypes"
   }
   module "ciso646" {
+    requires !cplusplus20
     export *
     header "ciso646"
   }
--- a/interpreter/cling/include/cling/std_msvc.modulemap
+++ b/interpreter/cling/include/cling/std_msvc.modulemap
@@ -68,6 +68,7 @@ module "std" [system] {
     header "cinttypes"
   }
   module "ciso646" {
+    requires !cplusplus20
     export *
     header "ciso646"
   }


dpiparo:
I added @hahnjo and @devajithvs to the reviewers.

guitargeek:
Thanks! Note that there is one more warning about a non-standard header in C++20, but I don't know what we can do about that since still have to support C++17:
```txt
While building module 'Core':
While building module 'std' imported from input_line_1:1:
In file included from <module-includes>:12:
/usr/lib/gcc/x86_64-redhat-linux/15/../../../../include/c++/15/ciso646:49:6: warning: "<ciso646> is not a standard header since C++20, use <version> to detect implementation-specific macros" [-W#warnings]
#    warning "<ciso646> is not a standard header since C++20, use <version> to detect implementation-specific macros"
```
Can we also do something about that?

cmake/unix/makepchinput.py
```diff
@@ -52,113 +52,116 @@ def getGuardedStlInclude(headerName):
...
+    # Special case for regex
+    allHeadersPartContent += (
+        "// treat regex separately\n"
+        + '#if __has_include("regex") && !defined __APPLE__\n'
+        + "#include <regex>\n"
+        + "#endif\n"
+    )
+
+    # treat this deprecated headers in a special way
+    stlDeprecatedHeadersList = ["strstream"]
```
ferdymercury, line 151:
```suggestion
    stlDeprecatedHeadersList = ["strstream", "ciso646"]
```
maybe this could help?
(Not sure if strstream can be removed.)

hahnjo:
> Thanks! Note that there is one more warning about a non-standard header in C++20, but I don't know what we can do about that since still have to support C++17:
> 
> [...]
> 
> Can we also do something about that?

I guess you can try `requires !cplusplus20`, that should only build the module with C++17 because later standards always define all previous versions.

interpreter/cling/include/cling/std_darwin.MacOSX14.2.sdk.modulemap
```diff
@@ -119,10 +119,6 @@ module std [system] {
       // <cassert>'s use of NDEBUG requires textual inclusion.
       textual header "cassert"
     }
-    module ccomplex {
-      header "ccomplex"
-      export *
-    }
```
hahnjo, line 125: The upstream https://github.com/llvm/llvm-project/blob/main/libcxx/include/module.modulemap.in still has these, so I think we shouldn't be modifying this copied file from the macOS SDK.

guitargeek: Ok, sure! I was only seeing the warnings on Linux anyway.

cmake/unix/makepchinput.py
guitargeek: That doesn't help, the warnings come from compiling the modules :(

guitargeek:
> > Thanks! Note that there is one more warning about a non-standard header in C++20, but I don't know what we can do about that since still have to support C++17:
> > [...]
> > Can we also do something about that?
> 
> I guess you can try `requires !cplusplus20`, that should only build the module with C++17 because later standards always define all previous versions.

Nice, that works!

Issue #20786, opened 2025-12-30
thisroot.sh can set incorrect ROOTSYS


ClarkMcGrew:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

`thisroot.sh` doesn't check that the found `SOURCE` actually contains `thisroot.sh` before using it to set `ROOTSYS`.  I've checked and the relevant lines on "master" don't seem to have changed.

### Reproducer

This can be reproduced from the command line using
```bash
cd  # don't be in the root source
cat [location of root]/bin/thisroot.sh | /usr/bin/dash
```
(substitute the path to root).  On my system the exact command is:
```bash
cd
cat work/t2k/software/nd280-v14.38/ROOT_6.32.02.03/Linux-Debian_13-gcc_14-x86_64/bin/thisroot.sh | /usr/bin/dash
```

### ROOT version

```
ROOT v6.32.02
Built for linuxx8664gcc on Dec 23 2025, 19:07:20
From heads/master@tags/v6-32-02
With c++ (Debian 14.2.0-19) 14.2.0
Binary directory: /home/mcgrew/work/t2k/software/nd280-v14.38/ROOT_6.32.02.03/Linux-Debian_13-gcc_14-x86_64/bin
```

### Installation method

Built from source using T2K build wrapper

### Operating system

Linux Debian 13.2

### Additional context

Using thisroot.sh inside a Makefile (generated by cmake) on debian 13 where "/usr/bin/sh" is linked to "dash" causes 
```shell
   x=$(lsof -p $$ -Fn0 2>/dev/null | tail -1); # Paul Brannan https://stackoverflow.com/a/42815001/7471760
   SOURCE=${x#*n}
```
to return `/dev/pts/0` (expected since Brannan's fix isn't completely portable, and there's probably no way to make it work everyplace), leaving SOURCE pointing to /dev.

This can be handled by checking that SOURCE contains `thisroot.sh`.  For instance, add:
```shell
   # Check to see if the right directory was found (near thisroot.sh line 238)
   if [ ! -f ${thisroot}/../bin/thisroot.sh ]; then
       echo SOURCE IS ${SOURCE}
       echo DID NOT FIND thisroot.sh at ${thisroot}/../bin
       return 1
   fi
```
before using `${thisroot}` to set ROOTSYS.  The thisroot.sh script will fail, but it doesn't leave ROOTSYS pointing to the wrong location, which is probably the best that can be hoped for on a pure posix shell and an odd setup.

[...]

PR #20796, opened 2026-01-05
Fix thisroot.sh: Added validation for SOURCE directory

ahmedbilal9:
## Description
Fixes root-project/root#20786

This PR adds validation to config/thisroot.sh to check that the detected SOURCE directory actually contains thisroot.sh before using it to set ROOTSYS.

## Problem
When lsof fails to detect the correct source location (particularly with dash shell on Debian), SOURCE can resolve to invalid paths like /dev/pts/0, causing ROOTSYS to be incorrectly set to /dev.

## Solution
- Added thisroot variable to store the calculated ROOT directory
- Added validation check to verify ${thisroot}/bin/thisroot.sh exists before setting ROOTSYS
- Returns error with informative message if validation fails
- Only sets ROOTSYS if validation passes

## Changes
- Modified config/thisroot.sh around line 265
- Check happens before ROOTSYS is set (prevents invalid state)
- Proper error handling with return 1 2>/dev/null || exit 1

## Testing
- Tested failure case: cat config/thisroot.sh | dash shows error message
- Verified fix prevents ROOTSYS from being set to /dev
- Logic validated on source repository

### DIFF
--- a/config/thisroot.sh
+++ b/config/thisroot.sh
@@ -262,7 +262,15 @@ if [ -z "${SOURCE}" ]; then
    fi
 else
    thisrootdir=$(dirname $(realpath "${SOURCE}"))
-   export ROOTSYS=${thisrootdir%/*}
+   thisroot=${thisrootdir%/*}
+   
+   if [ ! -f "${thisroot}/bin/thisroot.sh" ]; then
+      echo "ERROR: Could not determine correct ROOT installation directory" >&2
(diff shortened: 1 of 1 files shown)


ferdymercury:
Thanks for the proposed PR.
However, this addition is redundant since a quite similar check is already done a few lines below, (was fixed in the meantime in 6.38, was not there in 6.32)
The only remaining question is whether that fix present in 6.38 needs to be backported to a specific older version, depending on what user says.

PR #20798, opened 2026-01-05
Fix Chi2Test: Support residuals for 2D/3D histograms

### LINKED ISSUE #20761: Chi2Test: Report residuals also for higher dimensions
I wanted to perform a chi^2 comparison between two TH2 histograms (`h1` and `h2`), making use of the `TH1::Chi2Test` method. I hoped to get access to the individual residuals as follows:
```
TH2D h_res {*h1};
h_res.Reset();

h1->Chi2Test(h2, "OF UF" , h_res.GetArray());

//h_res should now contain the residuals
``` 

It really took me some time to understand why this is not working. It turns out this is a bug(?) in the code. The expression that fills the residuals only considers one dimension (x, lookp variable `i`) and simply ignores y (`j`) and z (`k`):

https://github.com/root-project/root/blob/9c4a8d221ad20873c70c22f17b9bbf9d1ea6af4e/hist/hist/src/TH1.cxx#L2256

I understand that a higher dimensional `res` array layout is more complicated, especially if custom limits are involved. Nevertheless I would expect at least the full 2D/3D case (as sketched above) to work.

ahmedbilal9:
## Description
Fixes #20761

This PR fixes a bug in `TH1::Chi2TestX` where residuals were only correctly stored for 1D histograms. For 2D and 3D histograms, the residual array indexing only used the `i` loop index and ignored `j` and `k`, causing incorrect storage.

## Changes
- Added `resIndex` counter for each of the three comparison modes (UU, UW, WW)
- Replaced `res[i - i_start]` with `res[resIndex]` in all three sections
- Added `resIndex++` increment after each residual is stored
- Each comparison mode now properly handles multi-dimensional histogram bins

## Testing
The fix ensures that residuals are stored sequentially for all bins in 2D/3D histograms, matching the order returned by `GetBin(i, j, k)`.

## Impact
Users can now retrieve residuals from `Chi2Test` for 2D and 3D histograms as originally intended by passing a residuals array.

### DIFF
--- a/hist/hist/src/TH1.cxx
+++ b/hist/hist/src/TH1.cxx
@@ -2389,7 +2389,7 @@ Double_t TH1::Chi2TestX(const TH1* h2,  Double_t &chi2, Int_t &ndf, Int_t &igood
                   }
                   else
                      res[resIndex] = delta1 / TMath::Sqrt(nexp1);
-                  if (res) resIndex++;
+                  resIndex++;
                }
             }
          }


hist/hist/src/TH1.cxx
hageboeck: Given that this block is enclosed in `if (res) {`, we could remove the second `if (res)`. Otherwise, people might suspect that it has a special purpose.
```suggestion
                  resIndex++;
```

hageboeck: Same simplification here:
```suggestion
                  resIndex++;
```

PR #20799, opened 2026-01-05
[math] missing overloads for addToGlobList

ferdymercury:
noticed by @makortel in https://github.com/root-project/root/pull/19645#discussion_r2651245873

name,formula should have an overload not using option, for symmetry with current TF1:

 ```
TF1(const char *name, const char *formula, Double_t xmin = 0, Double_t xmax = 1, EAddToList addToGlobList = EAddToList::kDefault, bool vectorize = false);
 TF1(const char *name, const char *formula, Double_t xmin, Double_t xmax, Option_t * option);  // same as above but using a string for option

```
These were missed when adding the different overloads on TF2 and TF3

### DIFF
--- a/hist/hist/src/TF2.cxx
+++ b/hist/hist/src/TF2.cxx
@@ -92,8 +92,9 @@ TF2::TF2(): fYmin(0),fYmax(0),fNpy(100)
 /// the formula string is "fffffff" and "xxxx" and "yyyy" are the
 /// titles for the X and Y axis respectively.
 
-TF2::TF2(const char *name, const char *formula, Double_t xmin, Double_t xmax, Double_t ymin, Double_t ymax, Option_t * opt)
-   :TF1(name, formula, xmax, xmin, opt) // purposely swapped xmax, xmin to signal that TFormula may be 1D or 2D
+TF2::TF2(const char *name, const char *formula, Double_t xmin, Double_t xmax, Double_t ymin, Double_t ymax,
+         Option_t *opt)
+   : TF1(name, formula, xmax, xmin, opt) // purposely swapped xmax, xmin to signal that TFormula may be 1D or 2D
 {
    if (ymin < ymax) {
       fYmin   = ymin;
@@ -126,7 +127,8 @@ TF2::TF2(const char *name, const char *formula, Double_t xmin, Double_t xmax, Do
 
 TF2::TF2(const char *name, const char *formula, Double_t xmin, Double_t xmax, Double_t ymin, Double_t ymax,
          EAddToList addToGlobList, bool vectorize)
-   : TF1(name, formula, xmax, xmin, addToGlobList, vectorize) // purposely swapped xmax, xmin to signal that TFormula may be 1D or 2D
+   : TF1(name, formula, xmax, xmin, addToGlobList,
+         vectorize) // purposely swapped xmax, xmin to signal that TFormula may be 1D or 2D
 {
    if (ymin < ymax) {
       fYmin = ymin;
@@ -178,7 +180,6 @@ TF2::TF2(const char *name, Double_t xmin, Double_t xmax, Double_t ymin, Double_t
    fNpy    = 30;
    fContour.Set(0);
 }
-
 
 ////////////////////////////////////////////////////////////////////////////////
 /// TF2 constructor using a pointer to a compiled function
--- a/hist/hist/src/TF3.cxx
+++ b/hist/hist/src/TF3.cxx
@@ -64,8 +64,10 @@ TF3::TF3()
 ///
 /// See TFormula constructor for explanation of the formula syntax.
 
-TF3::TF3(const char *name,const char *formula, Double_t xmin, Double_t xmax, Double_t ymin, Double_t ymax, Double_t zmin, Double_t zmax, Option_t * opt)
-   :TF2(name,formula,xmin,xmax,ymax,ymin,opt) // purposely swapped ymax, ymin to signal that TFormula may be 1D or 2D or 3D
+TF3::TF3(const char *name, const char *formula, Double_t xmin, Double_t xmax, Double_t ymin, Double_t ymax,
+         Double_t zmin, Double_t zmax, Option_t *opt)
+   : TF2(name, formula, xmin, xmax, ymax, ymin,
+         opt) // purposely swapped ymax, ymin to signal that TFormula may be 1D or 2D or 3D
 {
    fZmin   = zmin;
    fZmax   = zmax;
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by makortel** [comment]
Thanks!

File: hist/hist/src/TF2.cxx [hunk 1]
```diff
@@ -116,6 +115,39 @@ TF2::TF2(const char *name, const char *formula, Double_t xmin, Double_t xmax, Do
    }
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// F2 constructor using a formula definition and explicit option args
```
hageboeck:
```suggestion
/// TF2 constructor using a formula definition and explicit option args
```

hist/hist/src/TF2.cxx
```diff
@@ -116,6 +115,39 @@ TF2::TF2(const char *name, const char *formula, Double_t xmin, Double_t xmax, Do
...
+///
+/// See TFormula constructor for explanation of the formula syntax.
+///
+/// If formula has the form "fffffff;xxxx;yyyy", it is assumed that
+/// the formula string is "fffffff" and "xxxx" and "yyyy" are the
+/// titles for the X and Y axis respectively.
+
+TF2::TF2(const char *name, const char *formula, Double_t xmin, Double_t xmax, Double_t ymin, Double_t ymax,
+         EAddToList addToGlobList, bool vectorize)
+   : TF1(name, formula, xmax, xmin, addToGlobList, vectorize)
```
hageboeck: The arguments are listed as `xmin, xmax` in the [TF1 doxygen](https://root.cern.ch/doc/master/classTF1.html#a3522b7b13ba77e270257fafff233c009). Bug?

hist/hist/src/TF3.cxx
```diff
@@ -80,6 +79,28 @@ TF3::TF3(const char *name,const char *formula, Double_t xmin, Double_t xmax, Dou
...
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// F3 constructor using a formula definition and explicit option args
+///
+/// See TFormula constructor for explanation of the formula syntax.
+
+TF3::TF3(const char *name, const char *formula, Double_t xmin, Double_t xmax, Double_t ymin, Double_t ymax,
+         Double_t zmin, Double_t zmax, EAddToList addToGlobList, bool vectorize)
+   : TF2(name, formula, xmin, xmax, ymax, ymin, addToGlobList, vectorize)
```
hageboeck:
```suggestion
   : TF2(name, formula, xmin, xmax, ymin, ymax, addToGlobList, vectorize)
```

hist/hist/src/TF3.cxx
```diff
@@ -80,6 +79,28 @@ TF3::TF3(const char *name,const char *formula, Double_t xmin, Double_t xmax, Dou
    }
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// F3 constructor using a formula definition and explicit option args
```
hageboeck:
```suggestion
/// TF3 constructor using a formula definition and explicit option args
```

hist/hist/src/TF3.cxx
```diff
@@ -80,6 +79,28 @@ TF3::TF3(const char *name,const char *formula, Double_t xmin, Double_t xmax, Dou
...
+///
+/// See TFormula constructor for explanation of the formula syntax.
+
+TF3::TF3(const char *name, const char *formula, Double_t xmin, Double_t xmax, Double_t ymin, Double_t ymax,
+         Double_t zmin, Double_t zmax, EAddToList addToGlobList, bool vectorize)
+   : TF2(name, formula, xmin, xmax, ymax, ymin, addToGlobList, vectorize)
+{
+   fZmin = zmin;
+   fZmax = zmax;
```
hageboeck, line 95: Why do you commutate the y arguments in the TF2 part, but not these ones? And why are some arguments mixed up while others are not?

This looks like a source of problems.

hageboeck (commented):
Hello @ferdymercury,

please have a look at the suggestions below. I think we should make sure that the min max arguments are placed and used consistently.

hist/hist/src/TF2.cxx
ferdymercury: I copied this from preexisting line 96.  Shall I change both?

hist/hist/src/TF3.cxx
ferdymercury: I copied this from line 68.
I am suspecting that this wrong order is a "marker" for TF(n-1) to say, hey, this is not the last dimension.
Both for TF2 and TF3 follow the same pattern, so unlikely a bug?

hist/hist/src/TF3.cxx
ferdymercury: This was all copy-pasted from preexisting constructor.

hist/hist/src/TF3.cxx
```diff
@@ -80,6 +79,28 @@ TF3::TF3(const char *name,const char *formula, Double_t xmin, Double_t xmax, Dou
...
+   fNpz = 30;
+   Int_t ndim = GetNdim();
+   // accept 1-d or 2-d formula
+   if (ndim < 3)
+      fNdim = 3;
+   if (ndim > 3 && xmin < xmax && ymin < ymax && zmin < zmax) {
```
ferdymercury, line 101: This is what makes me suspect that swapping min and max has some kind of meaning for extendable dimensionality or so. see https://github.com/root-project/root/commit/b58da6b2271c8af184913499fa3d342b78bff2b5

hist/hist/src/TF2.cxx
ferdymercury: See commit where this was introduced: https://github.com/root-project/root/commit/112ed92e0bc4e39a1cbef9e1649dac9b1d73b578

ferdymercury: actually earlier: https://github.com/root-project/root/commit/b58da6b2271c8af184913499fa3d342b78bff2b5

hist/hist/src/TF3.cxx
ferdymercury: see https://github.com/root-project/root/commit/b58da6b2271c8af184913499fa3d342b78bff2b5

hageboeck: You are right. They didn't think about what code reviewers would see ...
Any chance we can make all of this look consistent, and still check the dimensions?

hageboeck: I had a look at the code, but nothing other than changing arguments or members comes to mind.
Maybe we should leave it as it is, but add comments for the next person who will stumble over this code.

What do you think?

ferdymercury: Seems a good idea! Thks.

hageboeck (approved):

pcanal (approved):
Thanks.

PR #20800, opened 2026-01-05
[Python] Raise minimum required Python version from 3.9 to 3.10

guitargeek:
Python 3.8 has been end-of-life since the release of Python 3.14 in October 2025.

The macOS runners are now using Python 3.11 and also our `alma9` images were upgraded to Python 3.12:
https://github.com/root-project/root-ci-images/pull/90

That means we can now follow the official Python end-of-life cycle and support only Python 3.10+ in the upcoming ROOT 6.40.

Note that for ROOT 6.40, we already raised the minimum supported version from 3.8 to 3.9 in 599cbfd279ff.

The next minimum Python version raise will be done in April 2027 when Ubuntu 22.04 is end-of-life, because that Ubuntu version uses Python 3.10.

See also:
  * https://devguide.python.org/versions/

### DIFF
--- a/.github/workflows/python_wheel_build.yml
+++ b/.github/workflows/python_wheel_build.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


siliataider (approved):
Thanks, looks great!

PR #20807, opened 2026-01-06
[test] Avoid using `GetRootSys()` in unit tests

guitargeek:
We want to rely less on environment variables like `ROOTSYS`.

See also the discussion in https://github.com/root-project/root/pull/20006#discussion_r2392531370


dpiparo (approved):
thanks.

PR #20809, opened 2026-01-06
Documentation updates for #20792

### LINKED ISSUE #20792: TGraph.Fit() usage with EnableImplicitMT() produces inconsistent results
### Check duplicate issues.

- [x] Checked for duplicates

This is a duplicate of [my root forum post](https://root-forum.cern.ch/t/tgraph-fit-usage-with-enableimplicitmt-is-tgraph-fit-thread-safe/64541) that got closed due to inactivity.

### Description

Running `TGraph.Fit()` with `EnableImplicitMT()` produces results that are not reproducible run-to-run.

### Reproducer

1. Create dummy data. Run as `root -l create.cpp`

```cpp
// create.cpp
using ROOT::RDataFrame;
using namespace ROOT::VecOps;
// modified df016_vecOps
int create(){
   auto unifGen = [](double) { return gRandom->Uniform(-1.0, 1.0); };

   auto vGen = [&](int len) {
      RVec<double> v(len);
      std::transform(v.begin(), v.end(), v.begin(), unifGen);
      return v;
   };

   // note that effect is very small and 200k events might be necessary to reproduce it.
   // I couldn't reproduce it with only 1024 events. So big N events is the key parameter.
   RDataFrame d(200000);
   auto d0 = d.Define("len", []() { return (int)gRandom->Uniform(5, 32); })
      .Define("x", vGen, {"len"})
      .Define("y", vGen, {"len"})
      .Redefine("y", "x+y")
      .Snapshot("events", "example.root");
      
   return 0;
}
```

2. Run `root -l read.cpp` several times with multi-threading disabled.

```cpp
// read.cpp
using ROOT::RDataFrame;
using namespace ROOT::VecOps;
// ROOT::EnableImplicitMT(4);

int read(){
   auto fitLine = [](RVec<double> x, RVec<double> y){
    // fit simple line a return a slope
    auto gr = TGraph();

    for (int i=0; i<x.size(); i++) gr.SetPoint(i, x[i], y[i]);

    // Solution
    gr.Fit("pol1", "Q");
    auto reco_slope = gr.GetFunction("pol1")->GetParameter(1);
    return reco_slope;
  };

   auto d = RDataFrame("events", "example.root");
   auto d0 = d.Define("slope", fitLine, {"x", "y"});

    auto h_x = d0.Histo1D({"hx", "hx", 200, -5, 5}, "x");
    auto h_y = d0.Histo1D({"hy", "hy", 200, -5, 5}, "y");
    auto h_slope = d0.Histo1D({"hslope", "hslope", 200, -5, 5}, "slope");
 
   std::cout<<std::fixed<<std::setprecision(3)<<"X: "<<h_x->GetMean()<<" +- "<<h_x->GetStdDev()<<std::endl;
   std::cout<<std::fixed<<std::setprecision(3)<<"Y: "<<h_y->GetMean()<<" +- "<<h_y->GetStdDev()<<std::endl;
   std::cout<<std::fixed<<std::setprecision(3)<<"Slope: "<<h_slope->GetMean()<<" +- "<<h_slope->GetStdDev()<<std::endl;
   return 0;
}
```

Output on my machine:

```bash
X: 0.000 +- 0.577
Y: 0.000 +- 0.817
Slope: 1.000 +- 0.309
```
I can run it as many times as I want and the number will remain identical (reproducible) from run to run.

3. Run `root -l read.cpp` several times with multithreading **enabled**.

Output on my machine:

```bash
# X, Y stay always the same
# Run with ROOT::EnableImplicitMT(2);
Slope: 1.000 +- 0.311
# Run with ROOT::EnableImplicitMT(4);
Slope: 1.001 +- 0.311
# Run with ROOT::EnableImplicitMT(6);
Slope: 1.001 +- 0.312
# Run with ROOT::EnableImplicitMT(8);
Slope: 1.001 +- 0.312
# Run with ROOT::EnableImplicitMT(16);
Slope: 1.000 +- 0.313
```

The standard deviation of the fitted slope is different from the run w/o multi-threading and moreover it changes based on the number of cores I use.

### ROOT version

ROOT v6.36.04
Built for linuxx8664gcc on Sep 19 2025, 13:34:39
From tags/6-36-04@6-36-04
With g++ (Spack GCC) 14.2.0
Binary directory: /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2025-09-21/x86_64-almalinux9-gcc14.2.0-opt/root/6.36.04-6qfi76/bin

### Installation method

source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh -r 2025-12-15

### Operating system

Almalinux9

### Additional context

If this is expected behavior due to hidden race conditions, I would like to know, how can I avoid that.

[I have already observed a very similar behavior](https://root-forum.cern.ch/t/using-random-with-rdataframe/58577), when creating random variables from a single engine caused race conditions, and the solution was to create N random engines, one per slot.
However, in this example the data is static, there is no randomly generated variables, (except maybe initial fit parameters), so I would expect the fit results to be reproducible.

There is also [an old forum post from 2012](https://root-forum.cern.ch/t/multi-threaded-fitting/16313/3) discussing that TH1::Fit() is not thread-safe, but [the mentioned bug report there](https://its.cern.ch/jira/browse/ROOT-5335) is long time closed.

What is the current up-to-date information on this?

hageboeck:
Clarifications regarding thread safety as a followup to #20792.

Fix #20792.

### DIFF
--- a/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
@@ -404,7 +404,7 @@ public:
    ////////////////////////////////////////////////////////////////////////////
    /// \brief Define a new column.
    /// \param[in] name The name of the defined column.
-   /// \param[in] expression Function, lambda expression, functor class or any other callable object producing the defined value. Returns the value that will be assigned to the defined column.
+   /// \param[in] expression Function, lambda expression, functor class or any other callable object producing the defined value. Returns the value that will be assigned to the defined column. This callable must be thread safe when used with multiple threads.
    /// \param[in] columns Names of the columns/branches in input to the producer function.
    /// \return the first node of the computation graph for which the new quantity is defined.
    ///
@@ -419,6 +419,7 @@ public:
    /// * extraction of quantities of interest from complex objects
    ///
    /// An exception is thrown if the name of the new column is already in use in this branch of the computation graph.
+   /// Note that the callable must be thread safe when called from multiple threads. Use DefineSlot() if needed.
    ///
    /// ### Example usage:
    /// ~~~{.cpp}


tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -404,7 +404,7 @@ public:
    ////////////////////////////////////////////////////////////////////////////
    /// \brief Define a new column.
    /// \param[in] name The name of the defined column.
-   /// \param[in] expression Function, lambda expression, functor class or any other callable object producing the defined value. Returns the value that will be assigned to the defined column.
+   /// \param[in] expression Function, lambda expression, functor class or any other callable object producing the defined value. Returns the value that will be assigned to the defined column. This callable must be reentrant when used with multiple threads.
```
guitargeek: What does "reentrant" mean here? Maybe it's not clear to all users. At least I don't know it :smile:

hageboeck: https://en.wikipedia.org/wiki/Reentrancy_(computing) 😄

I think we can say "thread safe", but that's not exactly the same.

guitargeek: Ok thanks! I didn't know that was an actual technical term :slightly_smiling_face:

hist/hist/src/TGraph.cxx
```diff
@@ -1267,14 +1267,16 @@ TFitResultPtr TGraph::Fit(TF1 *f1, Option_t *option, Option_t *goption, Axis_t r
...
 ////////////////////////////////////////////////////////////////////////////////
-/// Fit this graph with function with name `fname`.
+/// Fit this graph with the global function named `fname`.
 ///
-/// This is a different interface to TGraph fitting using TGraph::Fit(TF1 *f1,Option_t *, Option_t *, Axis_t, Axis_t)
-/// See there for the details about fitting a TGraph.
+/// This will retrieve the function with name `fname` from ROOT's global list of functions, and use it to
+/// fit the data in the TGraph.
+/// TF1 or TF2 functions that have been created in the same ROOT session can be accessed using `fname`.
+/// Predefined functions such as gaus, expo and poln are automatically created by ROOT.
```
ferdymercury:
```suggestion
/// Predefined functions such as gaus, expo poln, etc. are automatically created by ROOT. @see TF1::InitStandardFunctions, TF2::InitStandardFunctions, TF3::InitStandardFunctions
```

PR #20810, opened 2026-01-06
[core] Generalize TROOT::GetShardLibDir() also on Windows

guitargeek:
This follows up on 47caba598a15, also generalizing the function on Windows by dynamically finding `libCore.dll` (or `Core.dll` depending on the toolchain).

The goal is to avoid using `TROOT::GetRootSys()` in the implementation of this method, with the overarching goal of not having to rely on environment variables like `ROOTSYS` one day, which might not be set in some environments like the Python wheels.

**I disclose that ChatGPT was used for this mechanical task of translating code that already existed for macOS and Linux to Windows.**

### DIFF
--- a/core/base/src/TROOT.cxx
+++ b/core/base/src/TROOT.cxx
@@ -3018,6 +3018,33 @@ const TString& TROOT::GetBinDir() {
 ////////////////////////////////////////////////////////////////////////////////
 /// Get the library directory in the installation. Static utility function.
 ///
+/// By default, this is just an alias for TROOT::GetSharedLibDir(), which
+/// returns the directory containing the ROOT shared libraries.
+///
+/// On Windows, the behavior is different. In that case, this function doesn't
+/// return the directory of the **shared libraries** (like `libCore.dll`), but
+/// the **import libraries**, which are used at link time (like `libCore.lib`).
+
+const TString &TROOT::GetLibDir()
+{
+#if defined(R__WIN32)
+   static bool initialized = false;
+   static TString rootlibdir;
+   if (initialized)
+      return rootlibdir;
+
+   initialized = true;
+   rootlibdir = "lib";
+   gSystem->PrependPathName(GetRootSys(), rootlibdir);
+   return rootlibdir;
+#else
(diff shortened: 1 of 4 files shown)


core/base/src/TROOT.cxx
```diff
@@ -3057,9 +3057,35 @@ const TString &TROOT::GetLibDir()
...
 #elif defined(_WIN32)
 
-   // Or Windows, the original hardcoded path is kept for now.
-   rootlibdir = "lib";
-   gSystem->PrependPathName(GetRootSys(), rootlibdir);
+   HMODULE modules[1024];
+   DWORD needed = 0;
+
+   HANDLE process = GetCurrentProcess();
+   if (EnumProcessModules(process, modules, sizeof(modules), &needed)) {
```
pcanal: Should we handle and/or warn about the failure case (eg. if there are more than 1024 libraries or any other error)?  Right now, it seems that in the failure case the return value is 'empty'.

core/base/src/TROOT.cxx
```diff
@@ -3057,9 +3057,35 @@ const TString &TROOT::GetLibDir()
...
+      const unsigned int count = needed / sizeof(HMODULE);
+
+      for (unsigned int i = 0; i < count; ++i) {
+         wchar_t wpath[MAX_PATH];
+         DWORD len = GetModuleFileNameW(modules[i], wpath, MAX_PATH);
+         if (!len)
```
pcanal, line 3112: It is probably okay to ignore arbitrary library (for which the path is too long) ... but one of them might be `libCore` resulting in return an empty string.  Can we do better and/or can we warn the user in the case libCore was not found?

core/base/src/TROOT.cxx
guitargeek: I have now done what is recommended in the Windows API documentation to deal with this: call `EnumProcessModules` with a larger output array, if the original one was too small.

core/base/src/TROOT.cxx
guitargeek: It is now printing errors if the path is too long

PR #20812, opened 2026-01-06
[core] Deprecate TROOT::GetSourceDir()

guitargeek:
The `TROOT::GetSourceDir()` method is deprecated and will be removed in ROOT 6.42. It stopped making sense because the ROOT source is generally not shipped alongside ROOT in the `src/` subdirectory anymore.

This is done with the overarching goal of avoiding the need for `ROOTSYS`, which is used to infer the source directory `$ROOTSYS/src` (which doesn't exist anyway).

### DIFF
--- a/core/base/inc/TROOT.h
+++ b/core/base/inc/TROOT.h
@@ -366,7 +366,10 @@ public:
    static const TString& GetDocDir();
    static const TString& GetMacroDir();
    static const TString& GetTutorialDir();
-   static const TString& GetSourceDir();
+   static const TString &GetSourceDir()
+      R__DEPRECATED(6, 42,
+                    "This function is removed because it made only sense in the corner case where the ROOT source is "
+                    "copied inside the ROOT installation, which is never the case unless the user does it by hand.");
    static const TString& GetIconPath();
    static const TString& GetTTFFontDir();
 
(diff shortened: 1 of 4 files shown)


core/base/inc/TROOT.h
```diff
@@ -366,7 +366,10 @@ friend TROOT *ROOT::Internal::GetROOT2();
    static const TString& GetDocDir();
    static const TString& GetMacroDir();
    static const TString& GetTutorialDir();
-   static const TString& GetSourceDir();
+   static const TString &GetSourceDir()
+      R__DEPRECATED(6, 42,
+                    "This function is removed because it made only sense in the corner case where the ROOT source is "
+                    "manually copied inside the ROOT installation, which is typically not the case.");
```
pcanal: I am not yet finding in our `CMake` any way to install the source .. so technically the end of the sentence might be 'which is never the case unless the user does it by hand' ?

guitargeek: Good idea! By the way, I suspect that maybe there used to be a way way to install the source with the old Makefile-based build system. Otherwise, I can't explain myself why `TROOT::GetSourceDir()` exists.

guitargeek:
These spurious `Core.pcm' is out of date and needs to be rebuilt: module file out of date` just don't go away on Debian :slightly_frowning_face:

They seem unrelated though, so I'll merge.

PR #20814, opened 2026-01-07
[ci] Do not clone all the history of the repo

dpiparo:
Should shave off 1-2 minutes from every build.

### DIFF
--- a/.github/workflows/root-ci-config/build_root.py
+++ b/.github/workflows/root-ci-config/build_root.py
@@ -270,7 +270,7 @@ def git_pull(directory: str, repository: str, branch: str):
             """)
         else:
             returncode = subprocess_with_log(f"""
-                git clone --branch {branch} --single-branch {repository} "{targetdir}"
+                git clone --depth 1 --branch {branch} --single-branch {repository} "{targetdir}"
             """)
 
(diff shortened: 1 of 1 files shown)


guitargeek (approved):
Nice!

.github/workflows/root-ci-config/build_root.py
```diff
@@ -270,7 +270,7 @@ def git_pull(directory: str, repository: str, branch: str):
             """)
         else:
             returncode = subprocess_with_log(f"""
-                git clone --branch {branch} --single-branch {repository} "{targetdir}"
+                git clone --depth 1 --branch {branch} --single-branch {repository} "{targetdir}"
```
pcanal, line 273: I ponder if `1` is the right number.  If I remember correctly, we need some history to find the merging point between PR and the master in some circumstances.

ferdymercury: maybe as a two-step process? If the search fails, then we clone fully?

dpiparo: it is indeed what one can see in the incremental builds. I wonder whether it's an optimisation worth the complication in the code...

PR #20815, opened 2026-01-07
[win] Fix ACLiC compilation warnings on Windows

bellenot:
Fix several compilation warnings like the following:
```
execWriteCustomCollection.cxx(55): warning C4477: 'printf' : format string '%lu' requires an argument of type 'unsigned long', but variadic argument 1 has type 'size_t'
execWriteCustomCollection.cxx(55): note: consider using '%zu' in the format string
```


pcanal:
Thanks.

PR #20817, opened 2026-01-07
[Math] Explicitly define `FitUtil::Evaluate<Double_v>`

guitargeek:
This commit suggests to explicitly define `FitUtil::Evaluate<Double_v>` in `FitUtil.cxx`, as this is the only template instantiation that will be triggered besides `Evaluate<double>`, which is already explicitly defined in `FitUtil.h`.

This change changes IMT from a **public** to a **private** MathCore dependency, which drastically reduces unused linking.

In partticular, almost all ROOT libraries that depend on MathCore unnecessarily linked against `libIMT` because of that, and also its dependencies `libMultiProc` and `libNet`.

This reduces unnecessary linking by a lot, working towards #20731.

For example for `ldd --unused lib/libHist.so`, before it was:
```txt
Unused direct dependencies:
	/home/rembserj/code/root/root_build/lib/libImt.so
	/home/rembserj/code/root/root_build/lib/libMultiProc.so
	/home/rembserj/code/root/root_build/lib/libNet.so
	/home/rembserj/code/root/root_build/lib/libThread.so
	/nix/store/j193mfi0f921y0kfs8vjc1znnr45ispv-glibc-2.40-66/lib/libpthread.so.0
```
With this commit:
```txt
Unused direct dependencies:
	/home/rembserj/code/root/root_install/lib/libThread.so
	/nix/store/j193mfi0f921y0kfs8vjc1znnr45ispv-glibc-2.40-66/lib/libpthread.so.0
```


guitargeek:
These specializations are used in the end only in `Fitter.cxx` via the wrapper classes like `LogLikelihoodFCN` and `Chi2FCN`, so we're safe :+1:

PR #20818, opened 2026-01-07
[roottest] Fix roottest trying to invoke /usr/bin/root.exe in tests.

hageboeck:
When ROOT is installed in the system, roottest would try invoking /usr/bin/root.exe, leading to failures when trying to combine .pcms from two different ROOT builds.

This was caused by a find_package(ROOT) and find_program(root-config) at the start of the roottest config, which find the system ROOT before the ROOT being configured. With roottest being part of the ROOT build, any find_package(ROOT) and discovery of root-config and ROOTSYS could be removed, since these variables are anyway known already in the running CMake configuration.

This entailed a move from `if(ROOT_<component>_FOUND)` to `if(<component>)`, since the former is only set by `find_package(ROOT)` whereas the latter always available. As a side effect of this move, 4 tests enabled themselves, because the `ROOT_xxx_FOUND` apparently didn't propagate correctly to all locations in roottest. These are:
- roottest-python-JupyROOT-cppcompleter_doctest
- roottest-python-JupyROOT-handlers_doctest
- roottest-python-JupyROOT-utils_doctest
- gtest-roofit-roofit-vectorisedPDFs-testLegendre

- [x] Wait for #20830 to be merged

### DIFF
--- a/roottest/CMakeLists.txt
+++ b/roottest/CMakeLists.txt
@@ -1,5 +1,3 @@
-find_package(ROOT REQUIRED CONFIG)
-
 if(MSVC)
   set(CMAKE_SKIP_TEST_ALL_DEPENDENCY TRUE)
   set(CMAKE_SUPPRESS_REGENERATION TRUE)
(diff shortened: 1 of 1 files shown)


roottest/CMakeLists.txt
```diff
@@ -53,12 +51,13 @@ else()
 
   # root-config --prefix helps to retrieve ROOTSYS env variable.
   # ROOT_BINDIR is set by ROOTConfig.cmake
-  find_program(ROOT_CONFIG_EXECUTABLE root-config PATHS ${ROOT_BINDIR})
-  if(NOT ROOT_CONFIG_EXECUTABLE)
-    message(ERROR "root-config is not found, please rerun configuration step.")
+  set(ROOT_CONFIG_EXECUTABLE ${ROOT_BINDIR}/root-config)
+  if(NOT EXISTS ${ROOT_CONFIG_EXECUTABLE})
+    message(ERROR "root-config was not found in ${ROOT_CONFIG_EXECUTABLE}.")
```
ferdymercury:
```suggestion
    message(FATAL_ERROR "root-config was not found in ${ROOT_CONFIG_EXECUTABLE}.")
```

ERROR does not exist? https://cmake.org/cmake/help/v3.5/command/message.html

There are three more occurrences of this bug:

```
graf2d/asimage/src/libAfterImage/CMakeLists.txt:30:  message(ERROR "Can't find ft2build.h in ${FREETYPE_INCLUDE_DIR}")
graf2d/asimage/src/libAfterImage/CMakeLists.txt:34:  message(ERROR "Can't find zlib.h in ${ZLIB_INCLUDE_DIR}")
interpreter/llvm-project/llvm/cmake/modules/HandleLLVMOptions.cmake:578:      message(ERROR "LLVM_WINSYSROOT requires clang-cl")
```

hageboeck: 👍, I will work on those.

I also found the cause of (most of the) test failures. Will update tomorrow.

roottest/python/distrdf/backends/CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
-if(NOT ROOT_test_distrdf_dask_FOUND AND NOT ROOT_test_distrdf_pyspark_FOUND)
+if(NOT test_distrdf_dask_FOUND AND NOT ROOT_test_distrdf_pyspark)
```
guitargeek:
```suggestion
if(NOT test_distrdf_dask AND NOT test_distrdf_pyspark)
```
Automatic replacement didn't do the right thing here :)

hageboeck:
@guitargeek 
> Now that these variables are not set, the features are not detected and many tests are skipped, like the linked `roottest-python-pythonizations-pythonizations`. You can compare your CI runs with the nightlies and count the number of tests, and then the extent of this gets clear. For example, in this PR, 3706 test ran on `alma10`, and in the last nightlies there were `3742`.

I raise to 3746. I noticed this also locally, since there were subfolders where the (previously plain and local) CMake variables didn't get propagated correctly. Now, since the ifs in rootest rely on global CACHE variables, they are visible in all subdirectories.

roottest/python/distrdf/backends/CMakeLists.txt
hageboeck: Thanks, will fix!

hageboeck:
> Nice, thanks for this! Looks good now

OK, needs one more iteration, though. The following ifs don't do what we expected on Windows:
https://github.com/root-project/root/blob/6ecf2d7c58e231169154f68b85eb3b5542014950/roottest/python/cling/CMakeLists.txt#L11-L18


It's understood now. The logic will be updated in #20830

PR #20822, opened 2026-01-08
XMLParser: drop unused libm dependency

### LINKED ISSUE #20731: Investigate/Remove potentially unnecessary explicit linking

jainoshika:
This PR removes an unused direct dependency on `libm` from `libXMLParser`.

The unused dependency was identified via `ldd -u` on the official ROOT 6.38.00 binary build for AlmaLinux 9.7.
## Checklist:

- [x] tested changes locally

This PR fixes #20731

### DIFF
--- a/io/xmlparser/CMakeLists.txt
+++ b/io/xmlparser/CMakeLists.txt
@@ -24,7 +24,7 @@ ROOT_STANDARD_LIBRARY_PACKAGE(XMLParser
      src/TXMLDocument.cxx
      src/TXMLNode.cxx
      src/TXMLParser.cxx
-  DEPENDENCIES
+  LIBRARIES
     Core
 )
 


io/xmlparser/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 1995-2019, Rene Brun and Fons Rademakers.
+# CopyrightO (C) 1995-2019, Rene Brun and Fons Rademakers.
```
guitargeek:
```suggestion
# Copyright (C) 1995-2019, Rene Brun and Fons Rademakers.
```

io/xmlparser/CMakeLists.txt
```diff
@@ -39,3 +39,4 @@ target_link_libraries(XMLParser
     ${LIBXML2_LIBRARIES}
     $<$<PLATFORM_ID:Windows>:ws2_32.lib>
 )
+
```
guitargeek:
```suggestion
```

guitargeek (changes requested):
Thank you very much! Just try to keep unrelated changes out of the commit please

jainoshika:
Fixed — thanks for catching that. I’ve amended the commit. I can follow up with separate PRs for other libraries if you think that would be useful, keeping each change focused.

guitargeek (approved):
Thanks!

guitargeek (changes requested):
Wait that can't be right. `Core` is a public dependency of **XMLParser**, because `"TObject.h"` is included in the public `TXMLAttr.h`. So the Core dependency needs to be public.

jainoshika:
Thanks for the review and explanation. I will close this PR.
For the remaining cases in this issue, are there specific libraries you’d recommend I focus on first - ones that are safe to clean up?
I’m happy to follow your preference.

PR #20823, opened 2026-01-08
[core] Avoid `ROOTSYS` in TSystem include path

guitargeek:
Using environment variables like `ROOTSYS` in the include paths stored in TSystem - which are used in ACLiC - is problematic if we want to move away from such environment variables.

This commit suggests to figure out instead the actual `ROOTSYS` path with `gROOT->GetRootSys()`, and then fill it into the TSystem include path. This makes ALiC work reliably even if `ROOTSYS` is not defined as an environment variable.

The code is further simplified by using private preprocessor macros defined at compile time to pass around the `CMAKE_INSTALL_INCLUDEDIR`.

For illustration, here is the output of `root_build/bin/root -b -q -e "gSystem->GetIncludePath()"` on my system before and after this PR. Before:
```txt
(const char *) "-I$ROOTSYS/include"
```
After:
```txt
(const char *) "-I/home/rembserj/code/root/root_install/include"
```
Or, if running `bin/root` in the build tree:
```txt
(const char *) "-I/home/rembserj/code/root/root_build/include"
```

Spinoff from this PR, where I want to use ACLiC for some dictionary generation at test time, without `ROOTSYS` in the environment:
  * https://github.com/root-project/root/pull/18872

### DIFF
--- a/core/base/src/TROOT.cxx
+++ b/core/base/src/TROOT.cxx
@@ -1993,9 +1993,7 @@ void TROOT::InitSystem()
       if (gSystem->Init())
          fprintf(stderr, "Fatal in <TROOT::InitSystem>: can't init operating system layer\n");
 
-#define TO_LITERAL(string) _QUOTE_(string)
-
-      TString rootincludedir = TO_LITERAL(CMAKE_INSTALL_INCLUDEDIR);
+      TString rootincludedir = _R_QUOTEVAL_(CMAKE_INSTALL_INCLUDEDIR);
       gSystem->PrependPathName(GetRootSys(), rootincludedir);
       gSystem->SetIncludePath(("-I" + rootincludedir).Data());
 
@@ -3052,7 +3050,7 @@ const TString &TROOT::GetLibDir()
          continue;
 
       fs::path p(path);
-      if (p.filename() == TO_LITERAL(LIB_CORE_NAME)) {
+      if (p.filename() == _R_QUOTEVAL_(LIB_CORE_NAME)) {
          rootlibdir = p.parent_path().c_str();
          break;
       }
@@ -3072,7 +3070,7 @@ const TString &TROOT::GetLibDir()
          return 0;
 
       fs::path p = info->dlpi_name;
-      if (p.filename() == TO_LITERAL(LIB_CORE_NAME)) {
+      if (p.filename() == _R_QUOTEVAL_(LIB_CORE_NAME)) {
          libdir = p.parent_path().c_str();
          return 1; // stop iteration
       }
--- a/core/foundation/inc/ROOT/RConfig.hxx
+++ b/core/foundation/inc/ROOT/RConfig.hxx
@@ -447,6 +447,8 @@
 
 #endif
 
+#define _R_QUOTEVAL_(string) _QUOTE_(string)
+
 /* produce an identifier that is almost unique inside a file */
 #   define _R__JOIN_(X,Y) _NAME2_(X,Y)
 #   define _R__JOIN3_(F,X,Y) _NAME3_(F,X,Y)
--- a/roofit/batchcompute/src/RooBatchCompute.cxx
+++ b/roofit/batchcompute/src/RooBatchCompute.cxx
@@ -92,12 +92,7 @@ public:
    std::string architectureName() const override
    {
       // transform to lower case to match the original architecture name passed to the compiler
-#ifdef _QUOTEVAL_ // to quote the value of the preprocessor macro instead of the name
-#error "It's unexpected that _QUOTEVAL_ is defined at this point!"
-#endif
-#define _QUOTEVAL_(x) _QUOTE_(x)
-      std::string out = _QUOTEVAL_(RF_ARCH);
-#undef _QUOTEVAL_
+      std::string out = _R_QUOTEVAL_(RF_ARCH);
       std::transform(out.begin(), out.end(), out.begin(), [](unsigned char c) { return std::tolower(c); });
       return out;
    };
--- a/tree/tree/test/TTreeRegressions.cxx
+++ b/tree/tree/test/TTreeRegressions.cxx
@@ -238,12 +238,11 @@ TEST(TTreeRegressions, EmptyLeafObject)
 #define MYCLASS struct MyClass { std::vector<MySubClass> sub; MySubClass *Get(int id) { for (size_t i = 0; i < sub.size(); ++i) if (sub[i].id == id) return &sub[i]; return nullptr; } };
 MYSUBCLASS
 MYCLASS
-#define TO_LITERAL(string) _QUOTE_(string)
 
 TEST(TTreeRegressions, TTreeFormulaMemberIndex)
 {
-   gInterpreter->Declare(TO_LITERAL(MYSUBCLASS));
-   gInterpreter->Declare(TO_LITERAL(MYCLASS));
+   gInterpreter->Declare(_R_QUOTEVAL_(MYSUBCLASS));
+   gInterpreter->Declare(_R_QUOTEVAL_(MYCLASS));
 
    TTree tree("tree", "tree");
    MyClass mc;
--- a/tree/treeplayer/test/regressions.cxx
+++ b/tree/treeplayer/test/regressions.cxx
@@ -216,10 +216,9 @@ TEST(TTreeFormulaRegressions, ConstantAlias)
 // ROOT-8577 (JIRA)
 #define MYSTRUCT struct MyS { int x; };
 MYSTRUCT
-#define TO_LITERAL(string) _QUOTE_(string)
 TEST(TTreeFormulaRegressions, WrongName)
 {
-   gInterpreter->Declare(TO_LITERAL(MYSTRUCT));
+   gInterpreter->Declare(_R_QUOTEVAL_(MYSTRUCT));
    MyS s;
    TLorentzVector v(1, 2, 3, 4);
    TTree t("t", "t");


pcanal:
Were you able to test this with a relocated install? (i.e. installed in `$mylocal_install/root` and then copied/moved to `$myusable_install/root` (of course with `$mylocal_install/root` being deleted)

guitargeek:
Yes, that works without problem (checked both with `root` and `root.exe`, where `ROOTSYS` is never set at all). The `gROOT->GetRootSys()` method figures it out.

core/base/src/TROOT.cxx
```diff
@@ -1993,6 +1993,12 @@ void TROOT::InitSystem()
       if (gSystem->Init())
          fprintf(stderr, "Fatal in <TROOT::InitSystem>: can't init operating system layer\n");
 
+#define TO_LITERAL(string) _QUOTE_(string)
```
pcanal: This is the second place that we introduce this synonym for `_QUOTE_` (from `ROOT/RConfig.hxx`).  This might arguably be more readable but using both (since `_QUOTE_` is used in few other places) seems inconsistent.  Should we simply use `_QUOTE_` directly (or get rid of `TO_LITERAL`) or should we add `R__TO_LITERAL` to `RConfig.hxx` (and use it every place we use `_QUOTE_`)?

pcanal: correction: it is still declared only once but used twice, I just noticed it is being moved from the other place we introduced it).

core/base/src/TROOT.cxx
```diff
@@ -1993,6 +1993,12 @@ void TROOT::InitSystem()
...
+
+      TString rootincludedir = TO_LITERAL(CMAKE_INSTALL_INCLUDEDIR);
+      gSystem->PrependPathName(GetRootSys(), rootincludedir);
```
pcanal: Either this can be moved to the `TSystem` constructor or we should  consider using `AddIncludePath`.  (i.e. If we can not move in the `TSystem` constructor then at some later date a developer trying to set some value in `fIncludePath` in the constructor will be surprised that it is being ignored).

core/base/src/TROOT.cxx
guitargeek: And it's still necessary because here, we want to `_QUOTE_` a preprocessor macro itself. So it only works with this two-step macro somehow. I don't know the technical details, but it's a standard solution. I also don't know how to properly call it if we would make it a public `R__` macro, so I'd prefer to keep this private in TROOT.cxx

core/base/src/TROOT.cxx
guitargeek: It can't be moved to the `TSystem` constructor, because `gROOT` is initializing `gSystem` just a few lines before:
https://github.com/root-project/root/blob/master/core/base/src/TROOT.cxx#L1981

So using `gROOT->GetRootSys()` in the `TSystem` constructor would be illegal.

Using `AddIncludePath` would make no practical difference here, as I'm suggesting to call `SetIncludeDir` right after `gSystem->Init()`. So there is no chance for the user to be confused by some intermediate state where the include path is not set yet. For a developer, things will also get clear very quickly when reading this code.

pcanal: > For a developer, things will also get clear very quickly when reading this code.

Except for finding it :).  Adding one line of doc would solve that problem :)

core/base/src/TSystem.cxx
```diff
@@ -200,7 +200,6 @@ Bool_t TSystem::Init()
    fBuildNode     = BUILD_NODE;
    fFlagsDebug    = CXXDEBUG;
    fFlagsOpt      = CXXOPT;
-   fIncludePath   = INCLUDEPATH;
```
pcanal, line 203: Suggestion:
```
   // fIncludePath is set in TROOT::InitSystem()
```

core/base/src/TROOT.cxx
pcanal: > so I'd prefer to keep this private in TROOT.cxx

Except that we have this macro is now defined (i was inaccurate in my first post) 4 times:
```
roofit/batchcompute/src/RooBatchCompute.cxx:#define _QUOTEVAL_(x) _QUOTE_(x)
core/base/src/TROOT.cxx:#define TO_LITERAL(string) _QUOTE_(string)
tree/treeplayer/test/regressions.cxx:#define TO_LITERAL(string) _QUOTE_(string)
tree/tree/test/TTreeRegressions.cxx:#define TO_LITERAL(string) _QUOTE_(string)
```

I recommend to define `_QUOTEVAL_` (or `R__QUOTE_EVAL`) next to `_QUOTE_` as the `QUOTE EVAL` seems to me to be the clearest to make the distinction of behavior.

guitargeek: Ok, good idea! Thanks, I forgot that this was already used in other places too. I'll go with `_QUOTEVAL_`.

guitargeek:
@pcanal, thanks for your suggestions. I have implemented them.

core/foundation/inc/ROOT/RConfig.hxx
```diff
@@ -447,6 +447,8 @@
 
 #endif
 
+#define _QUOTEVAL_(string) _QUOTE_(string)
```
pcanal: Why not
```suggestion
#define _R_QUOTEVAL_(string) _QUOTE_(string)
```
to reduce the risk of clash (too late for the existing ones :) ).

pcanal (approved):
LGTM. Thanks.

ellert:
This breaks when CMAKE_INSTALL_INCLUDEDIR is an absolute path, since it always prepends the rootsys directory, even if the path is not a relative path.
This uses the install include path also when ROOTIGNOREPREFIX is set in order to run tests in the install tree where the include path is always "include" even if the configured install path is something else, e.g. "include/root".
This breaks ROOT::FoundationUtils::GetIncludeDir(), since it relies on the macro ROOTINCDIR:
https://github.com/root-project/root/blob/6ecf2d7c58e231169154f68b85eb3b5542014950/core/foundation/src/FoundationUtils.cxx#L181-L194
and ROOTINCDIR in turn is defined by incdir:
https://github.com/root-project/root/blob/6ecf2d7c58e231169154f68b85eb3b5542014950/config/RConfigure.in#L12
and the definition of incdir was removed from cmake/modules/RootConfiguration.cmake by this PR.
This also means that TROOT::GetIncludeDir() is broken, since it calls ROOT::FoundationUtils::GetIncludeDir():
https://github.com/root-project/root/blob/6ecf2d7c58e231169154f68b85eb3b5542014950/core/base/src/TROOT.cxx#L3101-L3106
So please revert this.
Then change the initialization for fInclude to "-I" + TROOT::GetIncludeDir().
This I think will achieve thw desired result without breaking a lot of other things.

guitargeek:
> This breaks when CMAKE_INSTALL_INCLUDEDIR is an absolute path

I was aware of this limitation, but tolerated it because the path is conventionally relative. Having `CMAKE_INSTALL_INCLUDEDIR` as an absolute path is not good practice as far as I remember. It should be a relative path inside the install prefix. Do you have a usecase where the path *needs* to be absolute?

I agree with your other observations, but I think the conclusion should be different. I think `TROOT::GetIncludeDir()` should also be updated to rely on `CMAKE_INSTALL_INCLUDEDIR` directly. Then we can also remove the hardcoded `include/`:
https://github.com/root-project/root/blob/6ecf2d7c58e231169154f68b85eb3b5542014950/core/foundation/src/FoundationUtils.cxx#L191

I'll make a PR to show what I mean

ellert:
Using absolute paths is the standard practice for package builds in distributions, so this not working will be disruptive. The current cmake files allow both - prepending the prefix already in cmake in case the paths are relative:
https://github.com/root-project/root/blob/6ecf2d7c58e231169154f68b85eb3b5542014950/cmake/modules/RootConfiguration.cmake#L42-L51
and following lines. (The definition of incdir was here before too). These paths are only used for gnu-install builds. The non-gnuinstall builds assume that the relative install path are the same as the relative build paths.
Not using TROOT::GetIncludeDir() to initialize fInclude would result in code duplication and repeating the intricate logic for handling relative and absolute paths and different settings of ROOTIGNOREPREFIX in more than one place.
Granted, the current TROOT::GetIncludeDir() doesn't allow using another value for CMAKE_INSTALL_INCLUDEDIR than "include" for a non-gnuinstall build. Up to this point non-gnuistall builds never allowed changing the default relative paths and the GetXxxDir() functions always used the same relative paths as in the build directory for non-gnuinstall builds. So the ROOTIGNOREPREFIX did not have to be checked for non-gnuinstall builds, since the relative paths are the same for the build and install trees in this case. If you start allowing changing the install paths for non-gnuinstall builds you also need to start checking the ROOTIGNOREPREFIX for them as is currently doen for the gnuinstall builds.

On the other hand the gnuinstall builds have a hardcoded install directories and can not be relocated.

In my opinion, fInclude shou be initialized using GetIncludeDir() as every other place that requires the include directory in the code does it. Any improvements, like allowing non-default values for paths in non-gnuinstall builds should be implemented in the GetXxxDir() functions, so that such changes only have to be implemented in one place.

guitargeek:
> Using absolute paths is the standard practice for package builds in distributions, so this not working will be disruptive. The current cmake files allow both - prepending the prefix already in cmake in case the paths are relative:

Okay fair enough, I don't see why but if that's widely done then we should account for it. That's easy: we can just infer the relative path, from `CMAKE_INSTALL_PREFIX` from `CMAKE_INSTALL_INCLUDEDIR`. But note that this is even [discouraged in the CMake documentation](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables).

> In my opinion, fInclude shou be initialized using GetIncludeDir() as every other place that requires the include directory in the code does it. Any improvements, like allowing non-default values for paths in non-gnuinstall builds should be implemented in the GetXxxDir() functions, so that such changes only have to be implemented in one place.

Yes, absolutely. Sorry I was not clear enough, I meant to imply that `GetIncludeDir()` will be rewritten but then used to set the TSystem include path, just like you suggested. Like that, there won't be any code duplication.

I'm suggesting this followup PR:
  * https://github.com/root-project/root/pull/20828

>  If you start allowing changing the install paths for non-gnuinstall builds you also need to start checking the ROOTIGNOREPREFIX for them as is currently doen for the gnuinstall builds.

We have started to allow that, and explicitly do that in the [build of the Python wheels](https://github.com/root-project/root/blob/master/setup.py#L94). However, I don't see why we need to check the `ROOTIGNOREPREFIX` what's the point of it anyway?

I think the only use of `ROOTIGNOREPREFIX` is for running tests in the build tree when the header install directory is different from the header directory in the build tree. In that case, setting `ROOTIGNOREPREFIX` will make `GetIncludeDir()` fall through to `$ROOTSYS/include`, which is always correct for the build tree. But then, I'd argue that this mechanism is also a bit opaque.

But the possible differences in the relative header/library/etc dirs between install and build tree are indeed a problem that needs to be addressed, while keeping both build and install tree relocatable. My goal for this is to introduce some other mechanism to auto-detect that, so that tests don't need to explicitly set `ROOTIGNOREPREFIX`. A simple marker file in the `lib` dir of *only* the build tree would be sufficient and AFAIK robust, which is what I have in mind right now to solve this problem longterm.

akritkbehera:
Hello
I guess post this PR root-config generates an empty -I flag as incdir is removed in root-config.
Just opened this issue [#20854](https://github.com/root-project/root/issues/20854) regarding it.

PR #20828, opened 2026-01-09
[core] Generalize TROOT::GetIncludeDir() using build tree marker

guitargeek:
Introducing a new marker file generalizes `TROOT::GetIncludeDir()` to
work in all cases for both relocated install and build trees, even if
the `CMAKE_INSTALL_INCLUDEDIR` is different from `include`.

This follows up on https://github.com/root-project/root/commit/43ec084a466458e09ed2cb7ad018418a739a8082, which was incomplete and left ROOT
configuration in a broken state, because `ROOTINCDIR` was not defined
anymore but still used.

The `RConfigure.h` header still defines a few other internal macros with
hardcoded paths, and if the suggested mechanism proves to be stable for
the include directory, we can also migrate the other macros. They are
fragile anyway. For example, they contain relative paths for
`gnuinstall=OFF` and absolute paths for `gnuinstall=ON`, which is
nowhere documented and breaks relocatable builds. Also, they use
`ROOTSYS` in the case of `gnuinstall=OFF`, which we want to migrate away
from.

The `ROOT::FoundationUtils::GetIncludeDir()` function was also removed
and completely absorbed into `TROOT::GetIncludeDir()`, as it was not
used in any other place.

Also, allow absolute `CMAKE_INSTALL_INCLUDEDIR` paths.

Follows up on:
  * https://github.com/root-project/root/pull/20823#issuecomment-3728343702

### DIFF
--- a/core/base/src/TROOT.cxx
+++ b/core/base/src/TROOT.cxx
@@ -1993,9 +1993,7 @@ void TROOT::InitSystem()
       if (gSystem->Init())
          fprintf(stderr, "Fatal in <TROOT::InitSystem>: can't init operating system layer\n");
 
-      TString rootincludedir = _R_QUOTEVAL_(CMAKE_INSTALL_INCLUDEDIR);
-      gSystem->PrependPathName(GetRootSys(), rootincludedir);
-      gSystem->SetIncludePath(("-I" + rootincludedir).Data());
+      gSystem->SetIncludePath(("-I" + GetIncludeDir()).Data());
 
       if (!gSystem->HomeDirectory()) {
          fprintf(stderr, "Fatal in <TROOT::InitSystem>: HOME directory not set\n");
@@ -3098,11 +3096,29 @@ const TString& TROOT::GetSharedLibDir() {
 ////////////////////////////////////////////////////////////////////////////////
 /// Get the include directory in the installation. Static utility function.
 
-const TString& TROOT::GetIncludeDir() {
-   // Avoid returning a reference to a temporary because of the conversion
-   // between std::string and TString.
-   const static TString includedir = ROOT::FoundationUtils::GetIncludeDir();
-   return includedir;
+const TString &TROOT::GetIncludeDir()
+{
+   static TString rootincdir;
+
+   if (!rootincdir.IsNull())
+      return rootincdir;
+
+   const std::string &sep = ROOT::FoundationUtils::GetPathSeparator();
+
+   // Check if we are in the build tree using the build tree marker file
+   const bool isBuildTree = std::filesystem::exists((GetSharedLibDir() + sep + "root-build-tree-marker").Data());
+
+   if (isBuildTree) {
+      rootincdir = GetRootSys() + sep + "include";
+   } else {
+#if INSTALL_INCLUDEDIR_IS_ABSOLUTE
+      rootincdir = _R_QUOTEVAL_(INSTALL_INCLUDEDIR);
+#else
+      rootincdir = GetRootSys() + sep + _R_QUOTEVAL_(INSTALL_INCLUDEDIR);
+#endif
+   }
+
+   return rootincdir;
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/cmake/modules/RootConfiguration.cmake
+++ b/cmake/modules/RootConfiguration.cmake
@@ -548,6 +548,19 @@ endif()
 string(REGEX MATCH "__cplusplus[=| ]([0-9]+)" __cplusplus "${__cplusplus_PPout}")
 set(__cplusplus ${CMAKE_MATCH_1}L)
 
+# To mark the build tree. Important for automatic resolution of relative paths,
+# for example to the include directory. Use custom target to ensure re-creation
+# when someone deletes the marker.
+set(build_tree_marker "${localruntimedir}/root-build-tree-marker")
+add_custom_command(
+  OUTPUT "${build_tree_marker}"
+  COMMAND ${CMAKE_COMMAND} -E touch "${build_tree_marker}"
+  COMMENT "Ensuring that \"${build_tree_marker}\" exists"
+)
+add_custom_target(ensure_build_tree_marker ALL
+  DEPENDS "${build_tree_marker}"
+)
+
 configure_file(${PROJECT_SOURCE_DIR}/config/RConfigure.in ginclude/RConfigure.h NEWLINE_STYLE UNIX)
 install(FILES ${CMAKE_BINARY_DIR}/ginclude/RConfigure.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 
--- a/config/RConfigure.in
+++ b/config/RConfigure.in
@@ -9,7 +9,6 @@
 #define ROOTPREFIX    "@prefix@"
 #define ROOTBINDIR    "@bindir@"
 #define ROOTLIBDIR    "@libdir@"
-#define ROOTINCDIR    "@incdir@"
 #define ROOTETCDIR    "@etcdir@"
 #define ROOTDATADIR   "@datadir@"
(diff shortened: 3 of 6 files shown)


guitargeek:
Hi @ellert, I can't think of a better solution. What do you think?

core/base/src/TROOT.cxx
```diff
@@ -3098,11 +3096,24 @@ const TString& TROOT::GetSharedLibDir() {
...
+
+   const std::string &sep = ROOT::FoundationUtils::GetPathSeparator();
+
+   // Check if we are in the build tree using the build tree marker file
+   const bool isBuildTree = std::filesystem::exists((GetSharedLibDir() + sep + "root-build-tree-marker").Data());
+
+   if (isBuildTree)
+      rootincdir = GetRootSys() + sep + "include";
+   else
+      rootincdir = GetRootSys() + sep + _R_QUOTEVAL_(INSTALL_INCLUDEDIR_REL);
```
pcanal: I wonder if, in the case where the include dir is an absolute path, rather than calculating the relative path in the CMake, we should just use here:

```suggestion
      if constexpr ( INSTALL_INCLUDEDIR is absolute path ) {
         rootincdir = INSTALL_INCLUDEDIR
      else 
         rootincdir = GetRootSys() + sep + _R_QUOTEVAL_(INSTALL_INCLUDEDIR);
```
The reason is that I don't see how the proposed solution 'helps' for the 'general' case.  In weird but 'plausible' cases, it does not allow relocatibility (which is likely not desired in this case), unless I am mistaken.   Concrete example:
- CMAKE_INSTALL_PREFIX = /opt/generic
- CMAKE_INSTALL_INCLUDEDIR = /opt/only_the_headers_please/root/include
- INSTALL_INCLUDEDIR_REL = ../only_the_headers_please/root/include

If in addition we have:
- library installed in /opt/generic/root/lib
- thus "ROOTSYS" is /opt/generic/root

I am guessing that the current state of the PR would not find the header because
- rootincdir = /opt/generic/root/../only_the_headers_please/root/include ... which is wrong: /opt/generic/only_the_headers/root/include (unwanted 'generic' part)

guitargeek: Good point. That would also simplify the CMake code, and indeed if one specifies absolute paths for the `CMAKE_INSTALL_<dirs>`, relocatability is arguably not intended (and we should not forget that CMake even discourages [in the documentation](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables)).

I'll make an alternative suggestion.

guitargeek: I now changed it to use the `CMAKE_INSTALL_INCLUDEDIR` as-is if it is absolute.

core/base/src/TROOT.cxx
```diff
@@ -3098,11 +3096,29 @@ const TString& TROOT::GetSharedLibDir() {
 ////////////////////////////////////////////////////////////////////////////////
 /// Get the include directory in the installation. Static utility function.
 
-const TString& TROOT::GetIncludeDir() {
-   // Avoid returning a reference to a temporary because of the conversion
-   // between std::string and TString.
-   const static TString includedir = ROOT::FoundationUtils::GetIncludeDir();
```
pcanal, line 3104: Could you extend the commit log to explain why the structure is changed to remove `ROOT::FoundationUtils::GetIncludeDir` (instead of putting the new implementation there).

core/base/src/TROOT.cxx
pcanal: Nice.  Thanks.

core/base/src/TROOT.cxx
```diff
@@ -3098,11 +3096,29 @@ const TString& TROOT::GetSharedLibDir() {
...
+{
+   static TString rootincdir;
+
+   if (!rootincdir.IsNull())
+      return rootincdir;
+
+   const std::string &sep = ROOT::FoundationUtils::GetPathSeparator();
+
+   // Check if we are in the build tree using the build tree marker file
+   const bool isBuildTree = std::filesystem::exists((GetSharedLibDir() + sep + "root-build-tree-marker").Data());
```
pcanal, line 3109: Side note that this means that each invocation of `root` will have one more (intentionally) failing file operation at startup.

core/base/src/TROOT.cxx
guitargeek: Sure! The explanation is simple: there was not other use of `ROOT::FoundationUtils::GetIncludeDir`, so the indirection was  not needed.

core/base/src/TROOT.cxx
guitargeek: I though we have so many file operations when starting ROOT already, because of all the library and symbol lookups etc. that this will not be significant at all. Or you have another intuition?

pcanal: An alternative (beyond the scope of this PR) could be to use 2 slightly different libCore for the inBuildTree and for the install version. Eg. there could be an extra file with a symbol set with a different value for both case.  In the buildTree, the `libCoreInBuildTree` would be copied to `libCore.so` and in the install step (using for example the `install(CODE ... COMPONENT Core)` form that let you execute CMake fragment at install time) to copy `libCoreForInstallOnly` onto `libCore.so`).

core/base/src/TROOT.cxx
guitargeek: Done

pcanal: > there was not other use of ROOT::FoundationUtils::GetIncludeDir

There used to be ... See the completely related commit 99b31eee58d3503c064eac7e954ca1f42ca14c59.
(ie. Does `gDriverConfig->fTROOT__GetIncludeDir()` relate to the current work?)

core/base/src/TROOT.cxx
pcanal: > I though we have so many file operations when starting ROOT already, because of all the library and symbol lookups etc. 

That is true, hence this is not holding this PR but it is a step in the wrong direction as we want/need to reduce this number *if* possible.

guitargeek: Okay, thanks for not holding off the PR for that. And so far, the build tree detection is also only used in one place, but I plan to use this more. I can think about a better mechanism on the next occasion, in the direction you described

guitargeek: Okay, thanks for not holding off the PR for that. And so far, the build tree detection is also only used in one place, but I plan to use this more. I can think about a better mechanism on the next occasion, in the direction you described

core/base/src/TROOT.cxx
guitargeek: For `rootcling`, the `fTROOT_GetIncludeDir()` already points to `TROOT::GetIncludeDir()`:
https://github.com/root-project/root/blob/master/core/metacling/src/rootclingTCling.cxx#L40

For `rootcling_stage1`, there is a separate implementation, which is AFAIK correct and has to only work for the build tree:
https://github.com/root-project/root/blob/master/core/rootcling_stage1/src/rootcling_stage1.cxx#L26

Thanks for pointing out https://github.com/root-project/root/commit/99b31eee58d3503c064eac7e954ca1f42ca14c59. This commit explains why `FoundationUtils::GetIncludeDir()` formerly existed but is not relevant anymore.

pcanal: ```rootcling_stage1.cxx:GetIncludeDir``` still uses `ROOTSYS` whereas it could probably be hard-coded (since stage1 is only used in the build tree) - However this could be address in another PR (if we decide to address it).

guitargeek:
@pcanal, thanks a lot for your review and also suggestions for future PRs!

Is this PR as it is okay for you?

pcanal (approved):
LGTM.  Thanks.

PR #20829, opened 2026-01-09
Add benchmark build

dpiparo:


ferdymercury:
Cool thanks. A-bit-related-potentially-useful feature: https://github.com/root-project/rootbench/pull/192 (https://its.cern.ch/jira/browse/ROOT-10181)

Fixes https://its.cern.ch/jira/browse/ROOT-10185

dpiparo:
Thanks for the comment, @ferdymercury !
You are right. This is the very first step towards the ambitious goals you list. We need first to revive rootbench, building and running it regularly. Some discussion will be needed about how we intend to address all the other requirements of a system for continuous performance monitoring, but the direction is clear :)

ferdymercury:
Thanks for the info and the great initiative!
One related question from my side: do you have plans in the future to merge rootbench with root as you did with roottest? I guess this will be much easier here since it's a very small git repo, and would potentially help for keeping it maintained.

dpiparo:
This is a smart question. We are discussing how to proceed. In principle you are right, however some components of rootbench, for example the cmake code or the way in which workflows are invoked from within C++, could be difficult to port as they are into ROOT as it is today. However, the value of the individual benchmarks and the idea of rootbench is still high for the project, we'll have to pragmatically converge towards a way to preserve it in the modern ROOT.

PR #20832, opened 2026-01-09
A few minor fixes in roottest

hageboeck:
While tracing the problems addressed in #20818, I stumbled over a few small errors / warnings, which are fixed here.

What's a bit worrying is the test `roottest-root-io-newstl-nolib`:

When it was ported and re-enabled by @linev, it was already failing, see 1795466620c24dd4404e8e73b16991b240b0b7f2.
Afterwards, a non-existing function call was introduced in https://github.com/root-project/root/pull/18859, https://github.com/root-project/root/commit/e60640c52a48dc0ab43d9e07270422ad1a532e90, such that the macro couldn't even be interpreted, any more. Due to the WILLFAIL annotation, this wasn't noticed.

Now we are back to the macro being interpreted, and failing during TTree::Scan with:
```
2336: ********************************************************
2336: *    Row   * Instance *       split99.fPairFlInt.first *
2336: ********************************************************
2336: *        0 *        0 *                              2 *
2336: *        1 *        0 *                              1 *
2336: *        1 *        1 *                              3 *
2336: *        2 *free(): invalid pointer
```

What should be the best action? Create an issue or disable/remove the test? @dpiparo @pcanal maybe?

### DIFF
--- a/roottest/root/io/newstl/readNoLib.C
+++ b/roottest/root/io/newstl/readNoLib.C
@@ -5,7 +5,7 @@ int readNoLib(const char *fname = "vector.root")
    dirname.ReplaceAll("/","-");
 
    TString filename = fname;
-   gSystem->PrependFilePath(dirname, filename);
+   gSystem->PrependPathName(dirname, filename);
 
    auto file0 = TFile::Open(filename);
    if (!file0)


pcanal:
> What should be the best action? Create an issue or disable/remove the test?

The 'best' action is to solve the problem of course :).  Short of this we should make sure that there is an issue tracking the underlying problem (creating if not present yet, it could also be in JIRA) and keep the test as WILL FAIL (the intent there is that is a cheap way to be notified when we 'unintentionally' fix the issue. (The downside is indeed, the test decay where it can not tell whether we are seeing the original issue or something else :( ).

PR #20833, opened 2026-01-09
Introduce a RUNPATH for compiled macros.

hageboeck:
This follows up on [3161321](https://github.com/root-project/root/commit/31613210d6c298cda056f3c47eb5b73722a75348), where LD_LIBRARY_PATH was removed from roottest.

It cannot be removed without replacement for tests that compile macros. Either LD_LIBRARY_PATH or RUNPATH is needed for those: When ROOT is installed in the system, or when an LD_LIBRARY_PATH is set such that a different ROOT installation is visible, compiled macros can fail. The sequence of events is as follows:
- roottest compiles a macro into an `.so`
- In a subsequent test, that `.so` is loaded.
- The dynamic loader will try to load the dependencies of that `.so`, e.g. `libROOTVecOps.so`, `libCore.so`, ... Given that the library being opened (the `.so` resulting from the macro) doesn't have any RUNPATHs set, the ROOT libraries it depends on are searched for first in LD_LIBRARY_PATH and then in the system.
- If a library of another ROOT is found in one of those two locations, the whole chain of library loads starts, opening all ROOT libraries that the last library found depends on, loading e.g. `/usr/lib64/libCore.so` and similar. This will provoke a lot of warnings, because TClassTable is already populated, and potentially crashes the program.

Therefore, in this PR, a RUNPATH is added for ACLiC libraries on Linux and Mac. `TSystem::GetMakeSharedLib()` by default now contains `$RPath`, which gets expanded to `-Wl,-rpath,<ROOT library path>`.

### DIFF
--- a/cmake/unix/compiledata.sh
+++ b/cmake/unix/compiledata.sh
@@ -43,10 +43,6 @@ if [ "$ARCH" = "macosx" ] || [ "$ARCH" = "macosx64" ] || \
       # If install_name is specified, remove it.
       SOFLAGS="$OPT -dynamiclib -single_module -Wl,-dead_strip_dylibs"
    fi
-   # Add rpath to the compiler options on MacOS, reproducing the same
-   # behaviour of manually creating the shared library by using the flags
-   # output by `root-config --libs`
-   SOFLAGS="$SOFLAGS -Wl,-rpath,$LIBDIR"
 elif [ "x`echo $SOFLAGS | grep -- '-soname,$'`" != "x" ]; then
     # If soname is specified, add the library name.
     SOFLAGS=$SOFLAGS\$LibName.$SOEXT


cmake/modules/RootMacros.cmake
```diff
@@ -3186,12 +3191,23 @@ function(ROOTTEST_ADD_TEST testname)
     string(REPLACE ";" ":" _path "${ROOTTEST_ENV_PATH}")
     string(REPLACE ";" ":" _pythonpath "${ROOTTEST_ENV_PYTHONPATH}")
 
+    if(ARG_MACRO)
+      # Help ".so"s from compiled macros locate the ROOT libraries being tested.
+      # The .so resulting from compiled macros don't contain RUNPATHs. Not setting LD_LIBRARY_PATH, root.exe loads
+      # the library, which will itself load all its dependencies (e.g. libCore, libHist, etc). However,
+      # since the macro library doesn't contain RUNPATHs, this might load /usr/lib64/libCore.so if ROOT is installed,
+      # or load another ROOT from LD_LIBRARY_PATH. This leads to two libCore being loaded, messing up all TClassTables,
+      # and possibly leading to ABI breakages.
```
guitargeek: Can we also write in the comment that a possible alternative would be to embed the RUNPATHS in the macro libraries compiled with ACLiC, maybe even as a `TODO`? I think that would be the more sustainable solution once effort for that is available. I can already see the problems ahead with people doing `pip install root` on lxplus, where there is also ROOT installed on the system (that's also a big reason why I try to get rid of relying on environment variables where we can).

hageboeck: OK, maybe I found a way to have a working RUNPATH in all ACLiC .so. The tests are running locally ...

core/base/src/TSystem.cxx
```diff
@@ -3736,6 +3736,7 @@ int TSystem::CompileMacro(const char *filename, Option_t *opt,
    cmd.ReplaceAll("\"$BuildDir","$BuildDir");
    cmd.ReplaceAll("$BuildDir","\"$BuildDir\"");
    cmd.ReplaceAll("$BuildDir",build_loc);
+   cmd.ReplaceAll("$RPath", "-Wl,-rpath=" + gROOT->GetSharedLibDir());
```
pcanal:
```
   cmd.ReplaceAll("$RPath", "-Wl,-rpath=" + gROOT->GetSharedLibDir());
```
This syntax is likely to be platform dependent (eg. different on Windows) and thus the `-Wl,-rpath=` may belong to the `compiledata.sh`.

pcanal: humm .. on the other hand for the Windows case, just not having `$PATH` on the produced line would have the same effect ... still is `-Wl,-rpath=` for all other platforms?

hageboeck: The Mac linker indeed didn't support `-rpath=x`, it was `-rpath x`. Since Linux supports this, too, I changed it to use spaces.
For Windows, there's no rpath as far as I understand, so `compiledata` for Windows doesn't contain the `$RPath` placeholder.

hageboeck:
> LGTM! It would just be nice to add some comment avoid the a possible more future-proof solution as a TODO, so we maybe attract effort to implement RUNPATH embedding in ACLiC.

In fact, that what I did now. 🙂

vepadulano (approved):
LGTM, remember to squash the second commit before merging.

PR #20836, opened 2026-01-09
[hist] Assert on zero arguments to `Fill`

hahnjo:
... and avoid follow-up errors from called functions or `LastType` by wrapping the remaining body into `if constexpr`.


dpiparo:
I like the clarity of the code and resulting error for the users.

PR #20847, opened 2026-01-12
[hist] Address some sign conversion warnings

hahnjo:


hist/histv7/inc/ROOT/RBinIndex.hxx
```diff
@@ -21,9 +21,9 @@ Objects of this type should be passed by value.
 Feedback is welcome!
 */
 class RBinIndex final {
-   static constexpr std::uint64_t UnderflowIndex = -3;
-   static constexpr std::uint64_t OverflowIndex = -2;
-   static constexpr std::uint64_t InvalidIndex = -1;
+   static constexpr auto UnderflowIndex = static_cast<std::uint64_t>(-3);
+   static constexpr auto OverflowIndex = static_cast<std::uint64_t>(-2);
+   static constexpr auto InvalidIndex = static_cast<std::uint64_t>(-1);
```
jblomer: Style: `kUnderflowIndex` etc.

(Not related to this PR)

hahnjo: On the other hand, the constants are only internal so let's just do it here in this PR :wink:

PR #20848, opened 2026-01-12
[638] Fix gcc16 638

dpiparo:
BP of #20841, #20826


guitargeek:
Thanks! Can https://github.com/root-project/root/pull/20826 also be backported in this PR? It's also for gcc16 support.

dpiparo:
added, thanks.

PR #20857, opened 2026-01-12
[ci] Simplify build configuration and pick up specific options

dpiparo:
This PR decouples the concept of plarform and platform configuration. If no platform configuration is specified, the default platform configuration is picked. Otherwise, a special one, `<platform>-<config label>.txt` is picked. 
This has multiple benefits, at least:
- Simpler yml files for steering builds
- Same PR, scheduled and workflow dispatch builds (modulo inlined overrides in the yml, as usual) *.
- Possibility to use the same platform (container) for many different configuration without specifying more parameters in the yml, e.g.: minimal builds, benchmark builds, march native, clang and in the future asan, tsan, ubsan, valgrind, helgrind...

* So far, for scheduled builds and workflow dispatch builds, the override parameters specified in the yml were simply ignored.

### DIFF
--- a/.github/workflows/root-ci-config/buildconfig/alma9-march_native.txt
+++ b/.github/workflows/root-ci-config/buildconfig/alma9-march_native.txt
@@ -1,0 +1,12 @@
+CMAKE_CXX_FLAGS=-march=native
+CMAKE_C_FLAGS=-march=native
+CMAKE_CXX_STANDARD=20
+builtin_vc=ON
+builtin_vdt=ON
+builtin_veccore=ON
+builtin_zlib=ON
+builtin_zstd=ON
+pythia8=ON
+tmva-sofie=ON
+veccore=ON
+vc=ON
--- a/.github/workflows/root-ci.yml
+++ b/.github/workflows/root-ci.yml
@@ -398,11 +398,13 @@ jobs:
           - image: alma9
             is_special: true
             property: modules_off
-            overrides: ["runtime_cxxmodules=Off", "CMAKE_CXX_STANDARD=20"]
+            platform_config: "alma9-modules_off"
+            overrides: ["CMAKE_CXX_STANDARD=20"]
           - image: alma9
             is_special: true
             property: march_native
-            overrides: ["CMAKE_BUILD_TYPE=RelWithDebInfo", "CMAKE_CXX_FLAGS=-march=native", "CMAKE_C_FLAGS=-march=native", "builtin_zlib=ON", "builtin_zstd=ON", "CMAKE_CXX_STANDARD=20", "builtin_vc=ON", "builtin_veccore=ON", "vc=ON", "veccore=ON"]
+            platform_config: "alma9-march_native"
+            overrides: ["CMAKE_BUILD_TYPE=RelWithDebInfo"]
           - image: alma10
             is_special: true
             property: arm64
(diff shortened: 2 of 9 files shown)


.github/workflows/root-ci.yml
```diff
@@ -411,8 +413,8 @@ jobs:
           - image: alma10
             is_special: true
             property: "clang Ninja"
-            overrides: ["CMAKE_C_COMPILER=clang", "CMAKE_CXX_COMPILER=clang++", "CMAKE_CXX_STANDARD=20"]
-            cmake_generator: Ninja
+            platform_config: "alma10-clang_ninja"
+            overrides: ["CMAKE_CXX_STANDARD=20"]
```
guitargeek, line 417: The `overrides` only make it into the PR builds, and not the nightlies. With this PR, you are actually fixing the long-standing problem that the special builds in the nightlies were not so special after all, because the `overrides` don't make it into the final CMake flags passed in the nightlies.

The overrides mechanism makes sense for non-special builds, where we want the nighlies to be different from the RP and incremental runs because they have different purposes. The PR runs are only for testing, and the nighlies are to test the artifact building.

For any special build, I don't see why the `overrides` make sense, also in this case where we would still use `CMAKE_CXX_STANDARD=20` for the CI builds but not the nightlies. Is there a motivation for that, or just an oversight?

It it's the latter, I would aim for the general guideline that no special build should use `overrdies` also, maybe that variable should be renamed to make clear that these overrides are only applied to the incrementals and pull request builds, but *not* the nightlies.

.github/workflows/root-ci-config/buildconfig/alma9-march_native.txt
```diff
@@ -0,0 +1,12 @@
...
+CMAKE_CXX_STANDARD=20
+builtin_vc=ON
+builtin_vdt=ON
+builtin_veccore=ON
+builtin_zlib=ON
+builtin_zstd=ON
+pythia8=ON
+tmva-sofie=ON
+veccore=ON
+vc=ON
```
guitargeek, line 12:
```suggestion
```
These deprecated flags can be removed (I forgot to remove them in `root-ci.yml` before...)

.github/workflows/root-ci-config/buildconfig/alma9-march_native.txt
```diff
@@ -0,0 +1,12 @@
+CMAKE_CXX_FLAGS=-march=native
+CMAKE_C_FLAGS=-march=native
+CMAKE_CXX_STANDARD=20
+builtin_vc=ON
+builtin_vdt=ON
+builtin_veccore=ON
```
guitargeek, line 6:
```suggestion
builtin_vc=ON
builtin_veccore=ON
```
Also here, two deprecated flags can be removed.

.github/workflows/root-ci.yml
dpiparo: > For any special build, I don't see why the overrides make sense, also in this case where we would still use CMAKE_CXX_STANDARD=20 for the CI builds but not the nightlies. Is there a motivation for that, or just an oversight?

The intent was to, as for the non-special builds, to increase coverage. We can decide that we do not want to specify overrides for special builds as a general rule. If we do so, we should also decide what to do about the global overrides.

Issue #20860, opened 2026-01-12
Inconsistent information for histogram contours


bgrube:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

When setting the number of contours of 2D histogram via `TH1::SetContour(n)`, only `n - 1` contours are created. In contrast, `TH1::GetContour` returns `n` levels. Also the list of contours, obtained by calling `(TObjArray*)gROOT->GetListOfSpecials()->FindObject("contours");` after drawing the histogram using the `"CONT0 LIST"` option, contains `n` elements, but the last entry does not contain any graphs. The index of the list of contours is shifted by 1 with respect to the level values.

### Reproducer

The issue is illustrated by the script below, which compares the results obtained by explicitly defining contour levels and by using `TH1::SetContour(n)`:
```cpp
void test(TH1* hist)
{
	// plot histogram with contours
	TCanvas* canv = new TCanvas();
	hist->Draw("CONT0 Z LIST");
	canv->Update();
	// loop over contours and copy graphs
	TObjArray* contours = (TObjArray*)gROOT->GetListOfSpecials()->FindObject("contours");
	std::cout << "Found " << contours->GetSize() << " contour levels." << std::endl;
	double contourLevels[contours->GetSize()];
	hist->GetContour(contourLevels);
	std::vector<TGraph*> graphs;
	for (int contourLevelIndex = 0; contourLevelIndex < contours->GetSize(); ++contourLevelIndex) {
		TList* contoursAtLevel = (TList*)contours->At(contourLevelIndex);
		std::cout << "Contour with index " << contourLevelIndex << " and level " << contourLevels[contourLevelIndex]
		          << " has " << contoursAtLevel->GetSize() << " graph(s)." <<  std::endl;
		for (TObject* graph: *contoursAtLevel) {
			graphs.push_back((TGraph*)graph->Clone());
		}
	}
	// overlay histogram with contours
	hist->SetContour(100);
	hist->Draw("COLZ");
	for (TGraph* graph : graphs) {
		graph->SetLineColor(kRed);
		graph->Draw("C");
	}
}


void testContours()
{
	TF2* fcn = new TF2("fcn", "x", -1, 1, -1, 1);
	TH1* hist = fcn->GetHistogram();
	// this produced the expected result
	hist->SetContour(2, std::vector<double>{-0.5, 0.5}.data());
	test(hist);
	// this does not work as expected
	hist->SetContour(2);
	test(hist);
}
```

### ROOT version

[...]

PR #20863, opened 2026-01-13
[hist] center contour levels between zmin and zmax

ferdymercury:
# This Pull request:

## Changes or fixes:

@bgrube fyi

Before: levels go from zmin till zmax - dz <--- this is weird, they are not centered in interval, and first contour is invisible since it's at the exact min.
After: levels go from zmin + 0.5 * dz till zmax - 0.5 * dz

Related to https://github.com/root-project/root/issues/20860

@linev what is the procedure to update the SVG? And why does stressSVG work on some alma? shouldn't it fail everywhere?

## Checklist:

- [x] tested changes locally


hageboeck:
> @linev what is the procedure to update the SVG? And why does stressSVG work on some alma? shouldn't it fail everywhere?

I just did it two weeks ago:
- Run stressgraphics with `-k` for keeping the outputs and with `-svg` to run in svg mode.
- Move the svg output files into `test/svg_ref` and commit

ferdymercury:
Thanks.

As can be seen in the SVG diff, the drawback of my solution is that  values between zmin and zmin + dz/2 appear now as white when using surface plots rather than contour plots.

Maybe a different solution could be to make them start in zmin and stop at zmax ? what are your thoughts?

ferdymercury:
Closing in favor of https://github.com/root-project/root/pull/20910
which seems cleaner and more backward compatible.

linev:
> @linev what is the procedure to update the SVG? And why does stressSVG work on some alma? shouldn't it fail everywhere?

Also problem is resolved - small readme is here:

https://github.com/root-project/root/blob/master/test/svg_ref/Readme.md

PR #20865, opened 2026-01-13
[windows] Fix GetWindowsVersion()

bellenot:
Add support for Windows 7, 8, 8.1, 10 + code formatting

### DIFF
--- a/core/winnt/src/TWinNTSystem.cxx
+++ b/core/winnt/src/TWinNTSystem.cxx
@@ -5646,201 +5646,196 @@ static const char *GetWindowsVersion()
 
    osvi.dwOSVersionInfoSize = sizeof(OSVERSIONINFOEX);
 
-   if( !(bOsVersionInfoEx = GetVersionEx ((OSVERSIONINFO *) &osvi)) )
-   {
+   if (!(bOsVersionInfoEx = GetVersionEx ((OSVERSIONINFO *) &osvi))) {
       osvi.dwOSVersionInfoSize = sizeof (OSVERSIONINFO);
-      if (! GetVersionEx ( (OSVERSIONINFO *) &osvi) )
+      if (!GetVersionEx((OSVERSIONINFO *) &osvi))
          return "";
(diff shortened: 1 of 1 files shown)


core/winnt/src/TWinNTSystem.cxx
```diff
@@ -5646,201 +5646,196 @@ static const char *GetWindowsVersion()
...
       // Test for the Windows NT product family.
       case VER_PLATFORM_WIN32_NT:
-
          // Test for the specific product.
-         if ( osvi.dwMajorVersion == 6 && osvi.dwMinorVersion == 0 )
-         {
-            if( osvi.wProductType == VER_NT_WORKSTATION )
-                strlcpy(strReturn, "Microsoft Windows Vista ",2048);
-            else strlcpy(strReturn, "Windows Server \"Longhorn\" " ,2048);
+         if (osInfo.dwMajorVersion == 10 && osInfo.dwMinorVersion == 0) {
```
pcanal, line 5674: Should we add a comment here and/or before new line 5655 giving a brief explanation of why 'osvi' is not enough here for Windows 7 through 10?

bellenot: Maybe. But I think I should review/rewrite several parts of `TWinNTSystem`, using new/documented API instead of undocumented one, to cleanup a bit the code (and adding comments where needed). Will do it later

bellenot: As complement of information: Since we don't use manifest, we hit this case with `GetVersionEx(LPOSVERSIONINFOA lpVersionInformation);`: Applications not manifested for Windows 8.1 or Windows 10 will return the Windows 8 OS version value (6.2). While `RtlGetVersion(PRTL_OSVERSIONINFOW lpVersionInformation);` fills the correct version without the need of manifest.

pcanal: thanks

PR #20868, opened 2026-01-13
[ci] Improve the CI in various respects

dpiparo:
- Allow to specify CMAKE_GENERATOR in the ci config files
- Use internally only uppercase ON and OFF for better readability
- Introduce platform configurations

### DIFF
--- a/.github/workflows/root-ci-config/buildconfig/global-minimal.txt
+++ b/.github/workflows/root-ci-config/buildconfig/global-minimal.txt
@@ -1,5 +1,0 @@
-ccache=ON
-fail-on-missing=ON
-minimal=ON
-roottest=ON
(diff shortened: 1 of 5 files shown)


dpiparo:
superseeding https://github.com/root-project/root/pull/20857

.github/workflows/root-ci-config/buildconfig/global-minimal.txt
```diff
@@ -1,5 +0,0 @@
-ccache=ON
```
pcanal, line 1: Minor disadvantage of not have a 'global' set of minimal setting is that we (sort of) lose the set that is meant to be used on all platforms we would test minimal on.

PR #13593, opened 2023-09-04
[PyROOT] Set memory policy to "strict"

guitargeek:
**Corresponding PR in roottest:**
  * https://github.com/root-project/roottest/pull/1011

PyROOT has many memory leaks, which is a major pain point for people using it for long-running scripts in batch jobs.

One source of memory leaks was indentified to be the "heuristic memory policy" of PyROOT. This means that PyROOT assumes that every non-const pointer member function argument was interpreted as the object taking ownership if the argument.

For examle, take the non-owning RooLinkedList container. It has a `RooLinkedList::Add(RooAbsArg *arg)` method. PyROOT wrongly assumes that this means the RooLinkedList takes ownership of arg, and it drops the PyROOT overship. Nobody feels responsible for deleting the object anymore, and there is a memory leak or `arg`.

That particular leak was reported in this forum post: https://root-forum.cern.ch/t/memory-leak-in-fits/56249

Function parameters of type `T *` are very common in ROOT, and only rarely do they imply ownership transfer. So changing the memory policy to "strict" would surely fix also many other memory leaks that are not reported so far. In fact, upstream cppyy doesn't even have this heuristic memory policy anymore! So moving PyROOT also to the strict memory policy closes the gap between PyROOT and cppyy.

The potential drawback of this change are crashes in usercode if memory is not properly managed. But these problems should either be fixed by:

  * the user

  * dedicated pythonizations for these methods to manage shared ownership via Python reference counters (i.e., setting the parameter as an attribute of the object that the member function was called on)

This follows up on PR #4294, in particular it reverts 3a12063116.

PR #20869, opened 2026-01-13
[hist] Delete `TMultiGraph::Add(TMultiGraph*, Option_t*)` overload

guitargeek:
The `TMultiGraph::Add(TMultiGraph*, Option_t*)` overload that adds the graphs in another **TMultiGraph** to a **TMultiGraph** is removed without deprecation. It was inconsistent from a memory ownership standpoint. A **TMultiGraph** always owns all the added graphs, so adding the same graph instances to two **TMultiGraphs** forcibly led to double-deletes. The ownership behavior could also not be changed, which gets clear from the [TMultiGraph destructor](https://github.com/root-project/root/blob/master/hist/hist/src/TMultiGraph.cxx#L389), which always calls `TList::Delete()` on the collection that holds the TGraphs.

No deprecation period is defined, because as the overload unavoidibly leads to double-deletes, it is likely not used. Or at least rarely used, because it can work if one of the two TMultiGraphs is leaked on purpose.

If you want to add all graphs from `otherMultiGraph` to `multiGraph`, one should use a for-loop and clone the graphs instead:

```c++
for (TObject *gr : *otherMultiGraph) {
   multiGraph->Add(static_cast<TGraph*>(gr->Clone()));
}
```

Some examle code with the double-delete:
```c++
{
   auto mg1 = new TMultiGraph();
   auto mg2 = new TMultiGraph();

   auto gr1 = new TGraph();

   mg1->Add(gr1);

   auto gr2 = new TGraph();

   mg1->Add(gr2);

   mg2->Add(mg1); // This is the Add(TMultiGraph* overload)

   delete mg1;
   delete mg2;
}
```

This is a spinoff from https://github.com/root-project/root/pull/13593,  triggered by a question from @vepadulano (https://github.com/root-project/root/pull/13593#discussion_r2681538582).

### DIFF
--- a/README/ReleaseNotes/v640/index.md
+++ b/README/ReleaseNotes/v640/index.md
@@ -44,6 +44,15 @@ The following people have contributed to this new version:
 * The `TGLIncludes.h` and `TGLWSIncludes.h` that were deprecated in ROOT 6.38 and scheduled for removal are gone now. Please include your required headers like `<GL/gl.h>` or `<GL/glu.h>` directly.
 * The GLEW headers (`GL/eglew.h`, `GL/glew.h`, `GL/glxew.h`, and `GL/wglew.h`) that were installed when building ROOT with `builtin_glew=ON` are no longer installed. This is done because ROOT is moving away from GLEW for loading OpenGL extensions.
 * The `TF1`, `TF2`, and `TF3` constructors for CINT compatibility were removed. This concerns the templated constructors that additionally took the name of the used functor class and member function. With ROOT 6, these names can be omitted.
+* The `TMultiGraph::Add(TMultiGraph*, Option_t*)` overload that adds the graphs in another **TMultiGraph** to a **TMultiGraph** is removed without deprecation.
+  It was inconsistent from a memory ownership standpoint.
+  A **TMultiGraph** always owns all the added graphs, so adding the same graph instances to two **TMultiGraphs** forcibly led to double-deletes.
+  If you want to add all graphs from `otherMultiGraph` to `multiGraph`, please use a for-loop and clone the graphs instead:
+  ```c++
+  for (TObject *gr : *otherMultiGraph) {
+     multiGraph->Add(static_cast<TGraph*>(gr->Clone()));
+  }
+  ```
 
 ## Build System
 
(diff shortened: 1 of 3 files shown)


guitargeek:
Note that this method was introduced 20 years ago in 701fbd2b51 with no added tutorial or other demo, so likely it is not used much. And indeed, the demo in the commit description was letting `mg1` leak, to avoid the double-delete. @couet do you maybe remember the original motivation, or have other suggestions to fix this ownership inconsistency? Or is just removing the method fine for you?

README/ReleaseNotes/v640/index.md
```diff
@@ -44,6 +44,15 @@ The following people have contributed to this new version:
...
 * The GLEW headers (`GL/eglew.h`, `GL/glew.h`, `GL/glxew.h`, and `GL/wglew.h`) that were installed when building ROOT with `builtin_glew=ON` are no longer installed. This is done because ROOT is moving away from GLEW for loading OpenGL extensions.
 * The `TF1`, `TF2`, and `TF3` constructors for CINT compatibility were removed. This concerns the templated constructors that additionally took the name of the used functor class and member function. With ROOT 6, these names can be omitted.
+* The `TMultiGraph::Add(TMultiGraph*, Option_t*)` overload that adds the graphs in another **TMultiGraph** to a **TMultiGraph** is removed without deprecation.
+  It was inconsistent from a memory ownership standpoint.
+  A **TMultiGraph** always owns all the added graphs, so adding the same graph instances to two **TMultiGraphs** forcibly led to double-deletes.
+  If you want to add all graphs from `otherMultiGraph` to `multiGraph`, please use a for-loop and clone the graphs instead:
+  ```c++
+  for (TObject *gr : *otherMultiGraph) {
```
vepadulano, line 52: I don't think `TMultiGraph` itself is an iterator, so I guess this should rather be
```suggestion
  for (TObject *gr : *(otherMultiGraph->GetListOfGraphs())) {
```

vepadulano (approved):
I believe this is a good choice in this case, because the existing API is shown to be a cause of immediate and verifiable bugs. The fact this is communicated clearly in the release notes is enough imho.

README/ReleaseNotes/v640/index.md
guitargeek: It can be used directly, because it implements the `begin()` and `end()` methods:
https://github.com/root-project/root/blob/master/hist/hist/inc/TMultiGraph.h#L69

Issue #9354, opened 2021-11-30
[Win] Warnings about missing PCMs


hageboeck:
Cling warns about missing PCMs on windows.

1. Revert e3f2ed26b7903ff540a37e77f6c026b8502db747
2. Fix the warnings in the failing tests.

dpiparo:
Is this still an issue?

guitargeek:
Let's find it out! #15186.

guitargeek:
It is still an issue on both Windows 64 bit and 32 bit as of April 2024.

Here is the relevant output:
```
2206/2226 Test #2197: gtestC--ROOT-CI-roottest-root-tree-readrules-gh-14462-gh14462regression ...........................***Failed    1.06 sec
  Running main() from C:\ROOT-CI\build\googletest-prefix\src\googletest\googletest\src\gtest_main.cc
  [==========] Running 1 test from 1 test suite.
  [----------] Global test environment set-up.
  [----------] 1 test from ReadRules
  [ RUN      ] ReadRules.CustomReadRuleWithSpace
  C:\ROOT-CI\src\core\testsupport\src\TestSupport.cxx(76): error: Failed
  Received unexpected diagnostic of severity 3000 at 'TCling::LoadPCM' reading 'ROOT PCM C:\ROOT-CI\build\roottest\root\tree\readrules\gh-14462\Release\gh14462regressionDict_rdict.pcm file does not exist'.
  Suppress those using ROOT/TestSupport.hxx
  [  FAILED  ] ReadRules.CustomReadRuleWithSpace (883 ms)
  [----------] 1 test from ReadRules (883 ms total)
  
  [----------] Global test environment tear-down
  [==========] 1 test from 1 test suite ran. (883 ms total)
  [  PASSED  ] 0 tests.
  [  FAILED  ] 1 test, listed below:
  [  FAILED  ] ReadRules.CustomReadRuleWithSpace
  
   1 FAILED TEST
  CMake Error at C:/ROOT-CI/build/RootTestDriver.cmake:232 (message):
    error code: 1
```

bellenot:
Here is the current status:
```
C:\root-dev\build\x64\relwithdebinfo>ctest -VV -C RelWithDebInfo -R gh-14462-gh14462regression
UpdateCTestConfiguration  from :C:/root-dev/build/x64/relwithdebinfo/DartConfiguration.tcl
Parse Config file:C:/root-dev/build/x64/relwithdebinfo/DartConfiguration.tcl
[...]

PR #18315, opened 2025-04-08
[cmake][win] Fix issue #9354

bellenot:
Don't use the Debug, Release, ... subdirs for gtest
This PR fixes https://github.com/root-project/root/issues/9354

PR #20870, opened 2026-01-13
[cmake][win] Fix issue #9354

bellenot:
Don't use the Debug, Release, ... subdirs for gtest
- supersedes https://github.com/root-project/root/pull/18315
- fixes https://github.com/root-project/root/issues/9354

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -36,6 +36,7 @@ if(MSVC AND NOT CMAKE_GENERATOR MATCHES Ninja)
   endif()
 endif()
 ROOT_EXECUTABLE(eventexe MainEvent.cxx LIBRARIES Event RIO Tree TreePlayer Hist Net)
+ROOT_SET_OUTPUT_DIRECTORIES(eventexe)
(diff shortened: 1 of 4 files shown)


test/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ if(MSVC AND NOT CMAKE_GENERATOR MATCHES Ninja)
   endif()
 endif()
 ROOT_EXECUTABLE(eventexe MainEvent.cxx LIBRARIES Event RIO Tree TreePlayer Hist Net)
+ROOT_SET_OUTPUT_DIRECTORIES(eventexe)
```
pcanal, line 39: Should this actually be inside `ROOT_EXECUTABLE` and `ROOT_STANDARD_LIBRARY_PACKAGE` or is there (any) cases where it should not be applied?

pcanal (approved):
Thanks.

test/CMakeLists.txt
bellenot: No, it's only needed for the tests. ROOT itself is just fine, the executables are already going into the `bin` directory and the libraries into the `lib` one

PR #20871, opened 2026-01-13
[tmva][sofie] New Keras Parser from SOFIE

lmoneta:
This PR superseeds #19692  fixing the conflicts and add some bug fixes

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/batchnorm.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/batchnorm.py
@@ -1,5 +1,7 @@
 from cppyy import gbl as gbl_namespace
+
 from .. import get_keras_version
+
 
 def MakeKerasBatchNorm(layer): 
     """
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/binary.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/binary.py
@@ -1,5 +1,6 @@
 from cppyy import gbl as gbl_namespace
 
+
 def MakeKerasBinary(layer):
     input = layer['layerInput']
     output = layer['layerOutput']
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/concat.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/concat.py
@@ -1,5 +1,6 @@
 from cppyy import gbl as gbl_namespace
 
+
 def MakeKerasConcat(layer):
     finput = layer['layerInput']
     foutput = layer['layerOutput']
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/conv.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/conv.py
@@ -1,6 +1,9 @@
-from cppyy import gbl as gbl_namespace
 import math
+
+from cppyy import gbl as gbl_namespace
+
 from .. import get_keras_version
+
 
 def MakeKerasConv(layer): 
     """
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/dense.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/dense.py
@@ -1,5 +1,6 @@
 from cppyy import gbl as gbl_namespace
 
+
 def MakeKerasDense(layer):
     """
     Create a Keras-compatible dense (fully connected) layer operation using SOFIE framework.
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/elu.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/elu.py
@@ -1,5 +1,6 @@
 from cppyy import gbl as gbl_namespace
 
+
 def MakeKerasELU(layer):
     """
     Create a Keras-compatible exponential linear Unit (ELU) activation operation using SOFIE framework.
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/flatten.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/flatten.py
@@ -1,5 +1,7 @@
 from cppyy import gbl as gbl_namespace
+
 from .. import get_keras_version
+
 
 def MakeKerasFlatten(layer):
     """
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/identity.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/identity.py
@@ -1,5 +1,6 @@
 from cppyy import gbl as gbl_namespace
 
+
 def MakeKerasIdentity(layer):
     input = layer['layerInput']
     output = layer['layerOutput']
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/layernorm.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/layernorm.py
@@ -1,5 +1,7 @@
 from cppyy import gbl as gbl_namespace
+
 from .. import get_keras_version
+
 
 def MakeKerasLayerNorm(layer):
     """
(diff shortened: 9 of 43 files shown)


guitargeek:
Thanks for the PR! One two things to do before a more detailed review is to merge #15790 and fix the test failures in this PR. But I also have a general question, the same as in https://github.com/root-project/root/pull/19692#issuecomment-3207115069: why not remove the old C++ parser at the same time, so we don't risk code duplication and user confusion?

guitargeek:
@lmoneta, it seems the new Python files are missing in the relevant `CMakeLists.txt` now, which is my fault because I have moved that `CMakeLists.txt` file in 3281aa99212b to [bindings/pyroot/pythonizations/python/CMakeLists.txt](https://github.com/root-project/root/blob/master/bindings/pyroot/pythonizations/python/CMakeLists.txt). Can you please add the Python files there now? Thanks!
```diff
diff --git a/bindings/pyroot/pythonizations/python/CMakeLists.txt b/bindings/pyroot/pythonizations/python/CMakeLists.txt
index 4c96c7556c7..e9bbed07628 100644
--- a/bindings/pyroot/pythonizations/python/CMakeLists.txt
+++ b/bindings/pyroot/pythonizations/python/CMakeLists.txt
@@ -58,7 +58,33 @@ if(tmva)
         ROOT/_pythonization/_tmva/_rtensor.py
         ROOT/_pythonization/_tmva/_tree_inference.py
         ROOT/_pythonization/_tmva/_utils.py
-        ROOT/_pythonization/_tmva/_gnn.py)
+        ROOT/_pythonization/_tmva/_gnn.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/__init__.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/generate_keras_functional.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/generate_keras_sequential.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/parser.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/parser_test_function.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/__init__.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/batchnorm.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/binary.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/concat.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/conv.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/dense.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/elu.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/flatten.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/identity.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/layernorm.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/leaky_relu.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/permute.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/pooling.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/reshape.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/relu.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/rnn.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/selu.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/sigmoid.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/softmax.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/swish.py
+        ROOT/_pythonization/_tmva/_sofie/_parser/_keras/layers/tanh.py)
     if(dataframe)
       list(APPEND PYROOT_EXTRA_PYTHON_SOURCES
           ROOT/_pythonization/_tmva/_batchgenerator.py)
```

guitargeek:
Hi @lmoneta, about these `ruff` linter errors: I know it's annoying to fix all of them, but many can actually be fixed automatically by running:
```txt
ruff check bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tmva/_sofie/_parser/_keras --fix
```
Ruff can just be installed with pip:
https://github.com/astral-sh/ruff

I think it would be worth it to add a commit that does does these automatic fixes! Then we don't have to update the code again later just for that.

tmva/sofie/test/CMakeLists.txt
```diff
@@ -185,9 +185,9 @@ if (tpython AND ROOT_KERAS_FOUND AND BLAS_FOUND)
        SYSTEM
        ${CMAKE_CURRENT_BINARY_DIR}
     )
-  else()
```
guitargeek, line 188: Since the ROOT Python module is used in the parser, you'd have to add this like to he `ROOT_ADD_GTEST` call above:
```CMake
       ENVIRONMENT PYTHONPATH=${localruntimedir}
```
So it becomes:
```CMake
    ROOT_ADD_GTEST(TestRModelParserKeras TestRModelParserKeras.C
       LIBRARIES
       ROOTTMVASofie
       Python3::NumPy
       Python3::Python
       BLAS::BLAS
       INCLUDE_DIRS
       SYSTEM
       ${CMAKE_CURRENT_BINARY_DIR}
       ENVIRONMENT PYTHONPATH=${localruntimedir}
    )
```

lmoneta: Thank you Jonas for the suggestion. I was just looking at how to set the PYTHONPATH in the C++ test

tutorials/CMakeLists.txt
```diff
@@ -363,10 +363,10 @@ else()
     list(APPEND tmva_veto machine_learning/TMVA_SOFIE_RSofieReader.C)
   endif()
   # These SOFIE tutorials take models trained via PyMVA-PyKeras as input
-  if (NOT tmva-pymva OR NOT tmva-sofie OR NOT ROOT_KERAS_FOUND OR keras_unsupported)
-    list(APPEND tmva_veto machine_learning/TMVA_SOFIE_Keras.C)
+  if (NOT tmva-sofie OR NOT ROOT_KERAS_FOUND OR keras_unsupported)
```
guitargeek:
```suggestion
  if (NOT tmva-sofie OR NOT ROOT_KERAS_FOUND)
```
I would just remove the whole mechanism with the unsupported Keras version at this point, which also means deleting the code branch above:
```cmake
  set(unsupported_keras_version "4.0.0")
  if (ROOT_KERAS_FOUND AND NOT DEFINED ROOT_KERAS_VERSION)
    message(WARNING "Keras found, but version unknown — cannot verify compatibility.")
  elseif (ROOT_KERAS_FOUND AND NOT ROOT_KERAS_VERSION VERSION_LESS ${unsupported_keras_version})
      message(WARNING "Keras version ${ROOT_KERAS_VERSION} is too new for the SOFIE Keras parser (only supports < ${unsupported_keras_version}). Corresponding tutorials will not be tested.")
    set(keras_unsupported TRUE)
  endif()
```

tmva/sofie/test/CMakeLists.txt
```diff
@@ -166,14 +166,14 @@ endif()
...
 
   # TODO: make sure we also support the newest Keras
-  if (NOT DEFINED ROOT_KERAS_VERSION)
-    message(WARNING "Keras found, but version unknown — cannot verify compatibility.")
-  elseif (ROOT_KERAS_VERSION VERSION_LESS ${unsupported_keras_version})
-    configure_file(generateKerasModels.py generateKerasModels.py COPYONLY)
-    configure_file(scale_by_2_op.hxx scale_by_2_op.hxx COPYONLY)
+  #if (NOT DEFINED ROOT_KERAS_VERSION)
+  #  message(WARNING "Keras found, but version unknown — cannot verify compatibility.")
+  #elseif (ROOT_KERAS_VERSION VERSION_LESS ${unsupported_keras_version})
```
guitargeek: I think this unused code can be deleted, including the `TODO` one line above

tmva/sofie/test/CMakeLists.txt
```diff
@@ -184,10 +184,11 @@ if (tpython AND ROOT_KERAS_FOUND AND BLAS_FOUND)
...
        SYSTEM
        ${CMAKE_CURRENT_BINARY_DIR}
+       ENVIRONMENT PYTHONPATH=${localruntimedir}
     )
-  else()
-    message(WARNING "Keras version ${ROOT_KERAS_VERSION} is too new for the SOFIE Keras parser (only supports < ${unsupported_keras_version})")
-  endif()
+#  else()
+#    message(WARNING "Keras version ${ROOT_KERAS_VERSION} is too new for the SOFIE Keras parser (only supports < ${unsupported_keras_version})")
+#  endif()
```
guitargeek: I'd suggest to delete unused code

bindings/pyroot/pythonizations/test/sofie_keras_parser.py
```diff
@@ -0,0 +1,96 @@
...
+   if backend.backend() == "tensorflow" :
+      import tensorflow as tf
+      if len(tf.config.list_physical_devices("GPU")) == 0:
+         return False
+
+   return True
+
+from ROOT._pythonization._tmva._sofie._parser._keras.parser_test_function import generate_and_test_inference
+from ROOT._pythonization._tmva._sofie._parser._keras.generate_keras_functional import generate_keras_functional
+from ROOT._pythonization._tmva._sofie._parser._keras.generate_keras_sequential import generate_keras_sequential
```
guitargeek: These functions are used only in the tests. Can they just be moved here?

It makes no sense to have unused functions in the Pythonization library, because it's not meant to be used by the user. So if something is in there that is not used by ROOT itself, we don't need it.

guitargeek:
There is also a `mac14` specific error:
```txt
Node: 'model_9/average_pooling2d/AvgPool'
Default AvgPoolingOp only supports NHWC on device type CPU
 	 [[{{node model_9/average_pooling2d/AvgPool}}]] [Op:__inference_train_function_2459]
```
I guess it's there because it's the only Intel mac platform we test. Maybe it's not necessarily worth for SOFIE to support it? I think setting `tmva-sofie=OFF` for mac14 would be an option:
https://github.com/root-project/root/blob/master/.github/workflows/root-ci-config/buildconfig/mac14.txt#L33

Issue #20790, opened 2026-01-01
Applications compiled on MacOS with -Wshadow report warnings from 14 ROOT headers


kelseymh:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

As reported in the ROOT forums in 2019, https://root-forum.cern.ch/t/shadow-warnings-when-compiling-against-root, when a user application is compiled against ROOT on MacOS (reported under 10.15.7/Catalina, still present with Tahoe 26.2), there are fourteen ROOT files which generate compiler warnings via `-Wshadow`, because they define local enumerator names which duplicate global namespace enums.  For example:
```
<ROOTSYS>/include/root/TBranch.h:108:7: warning: declaration shadows a variable in the global namespace [-Wshadow]
  108 |       kBranchAny    = ::kBranchAny,    ///< Branch is an object*
      |       ^
<ROOTSYS>/include/root/TBranch.h:59:13: note: previous declaration is here
   59 | const Int_t kBranchAny    = BIT(17); // branch is an object*
      |             ^
```

Following the guidance from @pcanal in that post, I was able to suppress the warnings (but not fix the underlying potential conflict) by adding `#pragma` directives at the top and bottom of the affected files:

	core/base/inc/TApplication.h
	core/base/inc/TAttMarker.h
	core/base/inc/TSystem.h
	core/base/inc/TVirtualX.h
	core/meta/inc/TClass.h
	core/meta/inc/TInterpreter.h
	gui/gui/inc/TGFrame.h
	io/io/inc/TFile.h
	math/matrix/inc/TMatrixT.h
	math/matrix/inc/TMatrixTSparse.h
	math/matrix/inc/TMatrixTSym.h
	net/net/inc/TSocket.h
	tree/tree/inc/TBranch.h

I have copied over from Forums user **vaubee** a minimal example to demonstrate the problem using `TBranch.h`.

### Reproducer

#include <iostream>
#include "TROOT.h"
#include "TBranch.h"

int main(int, char**) {
        std::cout<<TROOT::GetMacroPath()<<std::endl;
        return 0;
}

Comple this on MacOS with `clang++ -o test test.cc $(root-config --cflags --libs) -Wshadow`

### ROOT version

All ROOT versions since _at least_ 6.12/06.  Reproduced most recently with 6.36.06.

### Installation method

Homebrew (`brew install root`)

### Operating system

MacOS Catalina through Tahoe

aryansri05:
[...]

PR #20793, opened 2026-01-02
Fix: Suppress -Wshadow warnings in headers on macOS (Fixes #20790)

aryansri05:
## Changes or fixes:
This PR suppresses `-Wshadow` warnings in 13 header files when compiling with Clang on macOS.

The warnings were caused by local enumerator names (e.g., `kBranchAny`) shadowing global namespace enums. Since these are public API constants, they cannot be easily renamed without breaking backward compatibility.

Following the suggestion in the issue, I wrapped the affected code blocks with:
```cpp
#if defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wshadow"
#endif
// ... code ...
#if defined(__clang__)
#pragma clang diagnostic pop
#endif
Checklist:
[x] tested changes locally

[ ] updated the docs (N/A)

This PR fixes #20790

PR #20873, opened 2026-01-13
silence enum -Wshadow warnings by clang

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/20790

Tested locally with clang15, using: [test.txt](https://github.com/user-attachments/files/24595689/test.txt)

I did not go into modifying ROOT/....hxx headers, they give more warnings/errors, so to run the example above, comment all include ROOT/...:

This test.txt script could maybe become part of the static analyzer we mentioned some time ago (based on https://github.com/olifre/rootStaticAnalyzer)

On the other hand, one could say this is all a bug in clang and the issue is a Won'tFix, since g++ does not shown these warnings. So I'm happy to close this PR, too.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/tree/inc/TTreeCloner.h
+++ b/tree/tree/inc/TTreeCloner.h
@@ -96,12 +96,6 @@ class TTreeCloner {
    UInt_t FillCache(UInt_t from);
    void RestoreCache();
 
-private:
-   TTreeCloner(const TTreeCloner&) = delete;
-   TTreeCloner &operator=(const TTreeCloner&) = delete;
-
-   TTreeCloner(TTree *from, TTree *to, TDirectory *newdirectory, Option_t *method, UInt_t options = EClonerOptions::kNone);
-
 public:
    enum EClonerOptions {
 // clang++ <v20 (-Wshadow) complains about shadowing GuiTypes.h global variable kNone. Let's silence warning:
@@ -118,8 +112,16 @@ public:
       kNoFileCache = BIT(3)
    };
 
-   TTreeCloner(TTree *from, TTree *to, Option_t *method, UInt_t options = EClonerOptions::kNone);
-   TTreeCloner(TTree *from, TDirectory *newdirectory, Option_t *method, UInt_t options = EClonerOptions::kNone);
+private:
+   TTreeCloner(const TTreeCloner&) = delete;
+   TTreeCloner &operator=(const TTreeCloner&) = delete;
+
+   TTreeCloner(TTree *from, TTree *to, TDirectory *newdirectory, Option_t *method, UInt_t options = kNone);
+
+public:
+
+   TTreeCloner(TTree *from, TTree *to, Option_t *method, UInt_t options = kNone);
+   TTreeCloner(TTree *from, TDirectory *newdirectory, Option_t *method, UInt_t options = kNone);
    virtual ~TTreeCloner();
 
    void   CloseOutWriteBaskets();


core/base/inc/TParameter.h
```diff
@@ -36,12 +36,20 @@ class TParameter : public TObject {
...
 public:
    // Defines options / status while merging:
-   enum EStatusBits { kMultiply   = BIT(16),    // Use multiplication
-                      kMax        = BIT(17),    // Take max value
-                      kMin        = BIT(18),    // Take min value
-                      kFirst      = BIT(19),    // Take the first value
-                      kLast       = BIT(20),    // Take the last value
-                      kIsConst    = BIT(21)     // Set if all values are equal
+   enum EStatusBits {
+#if defined(__clang__)
```
ferdymercury:
```suggestion
// clang++ (-Wshadow) complains about shadowing TMatrixT::EMatrixCreatorsOp2. Let's silence warning:
#if defined(__clang__)
```

pcanal:
Is this meant to replace/extend https://github.com/root-project/root/pull/20793 ?

ferdymercury:
> Is this meant to replace/extend #20793 ?

Yes, it's a replacement since the other one was incorrect.

core/base/inc/TParameter.h
```diff
@@ -36,12 +36,21 @@ class TParameter : public TObject {
...
 public:
    // Defines options / status while merging:
-   enum EStatusBits { kMultiply   = BIT(16),    // Use multiplication
-                      kMax        = BIT(17),    // Take max value
-                      kMin        = BIT(18),    // Take min value
-                      kFirst      = BIT(19),    // Take the first value
-                      kLast       = BIT(20),    // Take the last value
-                      kIsConst    = BIT(21)     // Set if all values are equal
+   enum EStatusBits {
+// clang++ (-Wshadow) complains about shadowing TMatrixT::EMatrixCreatorsOp2. Let's silence warning:
```
pcanal: I am confused.  Is `clang` really complaining about `TParameter::kMultiply` shadowing `TMatrixT::kMultiply`?

ferdymercury: Yep. At least clang 15.

It's the same issue as https://issues.chromium.org/issues/41420763#comment6

Maybe related: https://stackoverflow.com/a/53844780

ferdymercury: Or should we rather close the issue as won't fix since clang is so nitpicky?

pcanal: @aaronj0 Is there a related `Clang` bug report and/or discussion?

ferdymercury: sorry, I was mistaken. Clang actually complains about: TParameter::kMultiply (enum within a class) shadowing EMarkerStyle { kMultiply} which is a global enum in TAttMarker.h

Here the reproducer outside of ROOT:

```
// clang++ -o test test.cc -Wshadow

/*
test.cc:7:28: warning: declaration shadows a variable in the global namespace [-Wshadow]
   enum MyClassLocalEnum { kMultiply = 4, kDivide = 6 };
                           ^
test.cc:3:21: note: previous declaration is here
enum MyGlobalEnum { kMultiply = 5 };
*/

enum MyGlobalEnum { kMultiply = 5 };

class MyClass {
public:
   enum MyClassLocalEnum { kMultiply = 4, kDivide = 6 };
};

class MyOtherClass {
public:
   enum MyOtherClassLocalEnum { kDivide = 7 };
};

int main(int, char**)
{
    return 0;
}
```

I have fixed the explanations in the comments to be clearer.

core/base/inc/TBuffer3D.h
```diff
@@ -46,7 +46,16 @@ class TBuffer3D : public TObject
    static void IncCSLevel();
    static UInt_t DecCSLevel();
 
-   enum ESection { kNone            = BIT(0),
+   enum ESection {
+// clang++ (-Wshadow) complains about shadowing GuiTypes.h global variable kNone. Let's silence warning:
```
pcanal: When investigating whether the 'fix' could be set on the one global enums instead of the class local enums, I discovered that the warning message is gone in clang 20 and later (see [godbolt code ](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:20,endLineNumber:9,positionColumn:20,positionLineNumber:9,selectionStartColumn:20,selectionStartLineNumber:9,startColumn:20,startLineNumber:9),source:'enum+GlobalEnums+%7B+kNone+%7D%3B%0A%0Astruct+DataClass+%7B%0A++++enum+InnerEnums+%7B+kNone+%7D%3B%0A++++%0A++++void+func()+%7B%0A++++++++auto+e+%3D+kNone%3B%0A++++++++if+(e+%3D%3D+::kNone)%0A++++++++++++return%3B%0A++++%7D%0A%7D%3B%0A%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:53.80150899593732,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:output,i:(compilerName:'x86-64+gcc+13.3',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+20.1.0+(assertions)+(Compiler+%231)',t:'0')),k:46.19849100406268,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:clang201assert,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!((name:fmt,ver:'1200')),options:'-std%3Dc%2B%2B20+-O3+-Wshadow+-Wall',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+clang+20.1.0+(assertions)+(Editor+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:46.19849100406268,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)) 

It is unclear whether it is intentional or not.  There is a related mention in the [release notes ](https://releases.llvm.org/20.1.0/tools/clang/docs/ReleaseNotes.html) that point to a related but different [issue](https://github.com/llvm/llvm-project/issues/62588) about nested classes.

So the conclusion of my investigation:
1. If we are going to suppress the warning, this PR is the (only) way to go.
2. This work-around wont be needed once Clang 20 and later are the only once used.

@dpiparo Any option on whether to merge this or close it?

dpiparo: Thanks for this investigation, it adds useful information to decide the way forward. What about silencing only if the Clang version is <20?

ferdymercury: Can be easily added, just let me know first your decision one way or the other, since there are a bunch of files

ferdymercury: added

ferdymercury:
@bellenot do you think the Windows failures are related to this PR?

bellenot:
> @bellenot do you think the Windows failures are related to this PR?

Yes, most probably. Here is the error:
```
C:\ROOT-CI\src\tree\tree\inc\TTreeCloner.h(103,101): error C2653: 'EClonerOptions': is not a class or namespace name
```
And I see https://github.com/root-project/root/pull/20873/changes#diff-7bf391e202dc2231e22184f39652173cf310f2eff835d1e38ee8caf1603a4e82L95-R103

ferdymercury:
> And I see https://github.com/root-project/root/pull/20873/changes#diff-7bf391e202dc2231e22184f39652173cf310f2eff835d1e38ee8caf1603a4e82L95-R103

Thanks. One question about shadowing, if we have:

```
UInt_t kNone = 3; // in a global namespace of an included header

class Whatever {
AnotherFunction(UInt_t options = kNone);
};
enum EClonerOptions {
      kNone       = 1
}
Whatever(UInt_t options = kNone);
};
```

which one will be used by Windows? kNone from the enum or the global one?
Or will it use different ones depending on if it was declared before or after the enum?

bellenot:
> > And I see https://github.com/root-project/root/pull/20873/changes#diff-7bf391e202dc2231e22184f39652173cf310f2eff835d1e38ee8caf1603a4e82L95-R103
> 
> Thanks. One question about shadowing, if we have:
> 
> ```
> UInt_t kNone = 3; // in a global namespace of an included header
> 
> class Whatever {
> AnotherFunction(UInt_t options = kNone);
> };
> enum EClonerOptions {
>       kNone       = 1
> }
> Whatever(UInt_t options = kNone);
> };
> ```
> 
> which one will be used by Windows? kNone from the enum or the global one? Or will it use different ones depending on if it was declared before or after the enum?

Good question... I'll have to try

ferdymercury:
> Good question... I'll have to try

Thanks. In any case, I did a reordering to be on the safe side. Now compilation on Windows works :)

PR #20877, opened 2026-01-13
[tmva][sofie] Fix Python C-API reference handling in Keras parser

Shlok-Saxena:
# This pull request
Fixes Python C-API reference handling issues in the SOFIE Keras parser (`tmva/sofie_parsers/src/RModelParser_Keras.cxx`) to avoid memory leaks during model parsing.

The Keras parser relies on Python C-API calls (dictionary lookups, attribute access, string conversions). A few utilities and attribute extraction paths were creating temporary Python objects (new references) without releasing them, which could **lead to increasing memory usage** when parsing multiple models or repeatedly parsing within the same ROOT session.

This PR ensures temporary Python objects created during parsing are properly released, improving stability when parsing multiple models in the same session.

## Changes or fixes:
- Fixed memory leak in `GetValueFromDict()` by avoiding leaking temporary `PyUnicode` key objects created via `PyUnicode_FromString()`
- Simplified and corrected `PyStringAsString()` implementation by using `PyUnicode_AsUTF8()` for safe UTF-8 extraction (avoids leaking temporary objects and unsafe lifetime issues)
- Fixed Python reference leaks in `MakeKerasBatchNorm()` by properly releasing `PyObject*` returned from `PyObject_GetAttrString()` when retrieving BatchNorm weight names (`gamma`, `beta`, `moving_mean`, `moving_variance`)

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tmva/sofie_parsers/src/RModelParser_Keras.cxx
+++ b/tmva/sofie_parsers/src/RModelParser_Keras.cxx
@@ -43,9 +43,7 @@ void PyRunString(TString code, PyObject *globalNS, PyObject *localNS)
 
 const char *PyStringAsString(PyObject *string)
 {
-   PyObject *encodedString = PyUnicode_AsUTF8String(string);
-   const char *cstring = PyBytes_AsString(encodedString);
-   return cstring;
+   return PyUnicode_AsUTF8(string);
 }
 
 std::vector<size_t> GetDataFromTuple(PyObject *tupleObject)
@@ -63,7 +61,10 @@ std::vector<size_t> GetDataFromTuple(PyObject *tupleObject)
 
 PyObject *GetValueFromDict(PyObject *dict, const char *key)
 {
-   return PyDict_GetItemWithError(dict, PyUnicode_FromString(key));
+   PyObject *pyKey = PyUnicode_FromString(key);
+   PyObject *val = PyDict_GetItemWithError(dict, pyKey);
+   Py_DECREF(pyKey);
+   return val;
 }
 
 } // namespace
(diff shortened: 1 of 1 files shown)


tmva/sofie_parsers/src/RModelParser_Keras.cxx
```diff
@@ -63,7 +61,10 @@ std::vector<size_t> GetDataFromTuple(PyObject *tupleObject)
 
 PyObject *GetValueFromDict(PyObject *dict, const char *key)
 {
-   return PyDict_GetItemWithError(dict, PyUnicode_FromString(key));
+   PyObject *pyKey = PyUnicode_FromString(key);
+   PyObject *val = PyDict_GetItemWithError(dict, pyKey);
+   Py_DECREF(pyKey);
+   return val;
```
sanjibansg, line 67: Why do we need Py_DECREF here and the changed implementation?

tmva/sofie_parsers/src/RModelParser_Keras.cxx
```diff
@@ -666,10 +667,18 @@ std::unique_ptr<ROperator> MakeKerasBatchNorm(PyObject* fLayer)
...
+      PyObject *meanName  = PyObject_GetAttrString(fMoving_Mean, "name");
+      PyObject *varName   = PyObject_GetAttrString(fMoving_Var, "name");
+      std::string fNScale = PyStringAsString(gammaName);
+      std::string fNB     = PyStringAsString(betaName);
+      std::string fNMean  = PyStringAsString(meanName);
+      std::string fNVar   = PyStringAsString(varName);
+      Py_DECREF(gammaName);
+      Py_DECREF(betaName);
+      Py_DECREF(meanName);
+      Py_DECREF(varName);
```
sanjibansg, line 681: same question on benefit of using this different implementation and Py_DECREF here.

guitargeek:
Thanks for the PR!

Unfortunately the timing was not good, as the Keras parser has been re-implemented in native Python:

  * https://github.com/root-project/root/pull/20871

So there is no need anymore to use the Python C API in the Keras parser.

PR #20883, opened 2026-01-14
[roottest] Fix build races

hahnjo:


roottest/root/ntuple/atlas-datavector/CMakeLists.txt
```diff
@@ -35,7 +35,11 @@ add_dependencies(${ATLASLIKEDATAVECTOR_LIB} ${ATLASLIKEDATAVECTOR_DICTNAME})
 
 add_test(NAME ${GENERATE_DICTIONARY_TEST}
   COMMAND ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR}
-                           --target ${ATLASLIKEDATAVECTOR_DICTNAME} ${ATLASLIKEDATAVECTOR_LIB})
+                           --target ${ATLASLIKEDATAVECTOR_DICTNAME}${fast} ${ATLASLIKEDATAVECTOR_LIB}${fast} -- ${always-make})
```
vepadulano: About this specific line, IIUC this PR https://github.com/root-project/root/pull/19344 should practically do the same thing by moving to using `ROOTTEST_GENERATE_DICTIONARY` which by default uses both `${fast}` and `${always_make}`. Since that PR is now working (just needs an approval), we could wait to merge that one and you could then rebase this one, or viceversa.

hahnjo: Ok, let's do that. I will then rebase this PR.

roottest/cling/dict/ROOT-8096/CMakeLists.txt
```diff
@@ -20,6 +20,10 @@ add_test(NAME roottest-cling-dict-ROOT-8096-build
                                   -- ${always-make})
 #                                  --target  ${targetname_libgen}${fast})
 set_property(TEST roottest-cling-dict-ROOT-8096-build PROPERTY ENVIRONMENT ${ROOTTEST_ENVIRONMENT})
+if(CMAKE_GENERATOR MATCHES Ninja AND NOT MSVC)
```
bellenot, line 23: It's not clear to me if the `AND NOT MSVC` part is really needed. I guess I'll have to try

hahnjo: Right, I just blindly copied from @hageboeck's changes in commit f5fca066fd6ef7f4c6957348918cf7cbcc468072. Yes, would be great if you could try if `RESOURCE_LOCK` also works on Windows because then I think we also want it there. In general, we may always wish to run `cmake --build` before any (compilation) test, even for GNU Make and other generators.

hageboeck: The `AND NOT MSVC` originates from Pere's commits trying to address the ninja-related failures, but it might indeed be a good thing to employ the RESOURCE_LOCK on Windows, because two concurrent invocations of ninja are unlikely to work.

hahnjo: @bellenot is this ok to merge in the current state and we follow up on the Windows side separately? I can create an issue to track, if you prefer.

bellenot: > @bellenot is this ok to merge in the current state and we follow up on the Windows side separately? I can create an issue to track, if you prefer.

Sure! Please go ahead! And no need to create an issue, I'll take care of it once it's merged

bellenot:
@hahnjo  FYI: https://github.com/root-project/root/pull/20917

PR #20886, opened 2026-01-14
Updates to stressgraphics

hageboeck:
## 1. Add a few small improvements for stressgraphics
- Use non-zero exit code when a test fails.
- When a test fails, print the file name of the generated graphics.
- Label all stressGraphics runs as longtest.
- Print differences in SVGs similar to how a unified diff would be
  printed.

## 2. Needed for #18083: Always use a new histogram for the tests
This makes the test future proof, working with or without implicit ownership of histograms.
Otherwise, due to TFile possibly returning a histogram that has been drawn in another test, tests can influence each other. This manifests in the stats box being drawn twice (once in the stats draw function of THistPainter, and a second time when the list of functions is drawn). This leads to a test failure in the svg tests, because here it's visible whether drawing happens once or twice, even if the output looks the same.
With this PR, the test will always use a fresh histogram, making the output independent of whether the same histogram is used in another test.

(I also tried a bit to remove the double drawing of the stats box, but this breaks other features like adding a custom stats box to the list of functions or enabling / disabling the stats box using the stats option. Always cloning the original histogram is the most robust option.)

### Note:
My clang-format did not have any issues, so the failure looks to be due to version differences in clang-format


test/stressGraphics.cxx
```diff
@@ -355,9 +356,19 @@ Int_t CompareSVGFiles(const TString &filename1, const TString &filename2, int te
       return 0;
    }
 
+   std::stringstream diffOld;
```
ferdymercury, line 360: `#include <sstream>` missing on top

dpiparo (approved):

PR #20888, opened 2026-01-14
[core] Make TDirectory::Append tolerant to identical objects.

hageboeck:
When calling TDirectory::Append(obj, replace=true), and obj is already in the directory, return without doing anything.
If obj has the same name as an existing object, but is physically different, the usual warning is raised.

This will enable workflows such as:
```c++
   auto h = new TH1D(...);
   directory->Append(h, true);
```

After this change, the above lines work both with and without auto registration to gDirectory. This is needed to prepare ROOT for working without auto registration.

### DIFF
--- a/hist/hist/test/test_TH1.cxx
+++ b/hist/hist/test/test_TH1.cxx
@@ -6,6 +6,9 @@
 #include "TH1F.h"
 #include "THLimitsFinder.h"
 
+#include "TDirectory.h"
+#include "TList.h"
+
 #include <cmath>
 #include <cstddef>
 #include <random>
@@ -332,4 +335,29 @@ TEST(TH1L, SetBinContent)
    static constexpr long long Large = 1LL << 42;
    h.SetBinContent(1, Large);
    EXPECT_EQ(h.GetBinContent(1), Large);
+}
+
+TEST(TH1, InteractionWithTDirectoryAdd)
+{
+   auto histo = std::make_unique<TH1F>("h", "h", 2, 0, 1);
+   auto tNamed = std::make_unique<TNamed>("h", "name clash");
+   auto tNamed2 = std::make_unique<TNamed>("h", "name clash 2");
+
+   auto dir = std::make_unique<TDirectory>("dir", "dir");
+   histo->SetDirectory(dir.get());
+   dir->Add(histo.get());
+   dir->Add(histo.get());
+   dir->Add(tNamed.get());
+   dir->Add(tNamed.get());
+   dir->Add(tNamed2.get());
+
+   EXPECT_EQ(histo->GetDirectory(), dir.get());
+
+   ROOT_EXPECT_WARNING_PARTIAL(dir->Add(histo.get(), /*replace=*/true), "Append", "Replacing existing TNamed");
+   EXPECT_EQ(histo->GetDirectory(), dir.get());
+   EXPECT_EQ(dir->GetList()->size(), 1);
+   EXPECT_EQ(dir->GetList()->First(), histo.get());
+
+   // Leave the ownership to the unique_ptrs
+   histo->SetDirectory(nullptr);
 }


core/base/src/TDirectory.cxx
```diff
@@ -205,6 +205,7 @@ void TDirectory::Append(TObject *obj, Bool_t replace /* = kFALSE */)
    if (replace && obj->GetName() && obj->GetName()[0]) {
       TObject *old;
       while (nullptr != (old = GetList()->FindObject(obj->GetName()))) {
+         if (old == obj) return;
```
dpiparo: Can there be a case in which the operator == is not well defined and would make this operation dangerous?

dpiparo: my apologies. These are pointers :( The comment does not apply.

dpiparo (approved):

core/base/src/TDirectory.cxx
pcanal: Minor white space.
```suggestion
         if (old == obj)
            return;
```

core/base/src/TDirectory.cxx
```diff
@@ -205,6 +205,7 @@ void TDirectory::Append(TObject *obj, Bool_t replace /* = kFALSE */)
    if (replace && obj->GetName() && obj->GetName()[0]) {
       TObject *old;
       while (nullptr != (old = GetList()->FindObject(obj->GetName()))) {
```
pcanal: The while loop implies that the list could contains multiple times the same names when this is called.  In this case the net result was to remove all duplicates and add only the one passed.
With the new code it seems that if the list contains multiple time the same name *and* already contains the item being passed we can have:
* before this PR: list contains only the argument
* after this PR: list contains the argument and possibly many duplicates.

hageboeck: Thanks, this should be addressed now!

hageboeck: This should be addressed now.

core/base/src/TDirectory.cxx
```diff
@@ -202,9 +202,16 @@ void TDirectory::Append(TObject *obj, Bool_t replace /* = kFALSE */)
 {
    if (!obj || !fList) return;
 
+   bool objInList = false;
    if (replace && obj->GetName() && obj->GetName()[0]) {
       TObject *old;
       while (nullptr != (old = GetList()->FindObject(obj->GetName()))) {
+         if (old == obj && !objInList) {
```
pcanal: 'Fun' fact.  The existing implementation is broken (and thus the new one is probably also, maybe differently, broken).  
This first one is fixed by the current state of the PR (because the 'old' is not removed).
```
root [0] auto d = new TDirectory("somewhere", "somehow")
(TDirectory *) 0x3603040
root [1] d->cd()
(bool) true
root [2] auto h = new TH1F("h","h",10,0,10);
root [3] h->GetDirectory()
(TDirectory *) 0x3603040
root [4] d->Append(h, true)
Warning in <TDirectory::Append>: Replacing existing TH1F: h (Potential memory leak).
root [5] h->GetDirectory()
(TDirectory *) nullptr
```
Where the fact that the histogram 'forgets' its attached to the directory is due to the call `func(old,nullptr);` but the adding done via `fList->Add(obj);` instead of `func(obj, this)`.

I am not sure about this second one (but it likely works 'better' but return to the histogram forgetting what directory it is attached to.  In the current code (without PR), this prints the warning an infinite amount of times:
```
root [0] auto d = new TDirectory("somewhere", "somehow")
(TDirectory *) 0x18886f0
root [1] d->cd()
(bool) true
root [2] auto h = new TH1F("h","h",10,0,10);
[Detaching after vfork from child process 4178911]
root [3] h->GetDirectory()
(TDirectory *) 0x18886f0
root [4] d->Append(h)
root [5] d->GetList()->ls()
OBJ: THashList  THashList       Doubly linked list with hashtable for lookup : 0
 OBJ: TH1F      h       h : 0 at: 0x11acca0
 OBJ: TH1F      h       h : 0 at: 0x11acca0
root [6] d->Append(h, true)
Warning in <TDirectory::Append>: Replacing existing TH1F: h (Potential memory leak).
Warning in <TDirectory::Append>: Replacing existing TH1F: h (Potential memory leak).
....
```

pcanal: As of the update to commit ae2f59bcf62e6f646629d2254c47f7ba823343e6, the 2 behaviors above are back (which the second now possibly looping/recursing silently but infinitely).

core/base/src/TDirectory.cxx
```diff
@@ -202,20 +202,37 @@ void TDirectory::Append(TObject *obj, Bool_t replace /* = kFALSE */)
 {
    if (!obj || !fList) return;
 
+   unsigned int nExistingInstances = 0;
```
pcanal: This number needs to be per instance.  I.e.:
```
auto ptr1 = new TNamed("abc", "abc");
auto ptr2 = new TNamed("abc", "abc");
...
list->Append(ptr1, false);
list->Append(ptr1, false);
list->Append(ptr2, false);
list->Append(ptr2, true);
```
If I am not mistaken the code as of commit af54f14cecad91ec9df87acdf299b163e62f5320 would only remove 'one' of the ptr1.

hageboeck: I updated the test, and it actually removes all instances. That's because they get pushed multiple times into the list of things to be removed.

hageboeck: Your comment did make me review the counter, though, and it's not needed. Now, all duplicates are removed using the `toRemoveList`.

core/base/src/TDirectory.cxx
```diff
@@ -202,20 +202,35 @@ void TDirectory::Append(TObject *obj, Bool_t replace /* = kFALSE */)
...
+            toRemoveList.push_back(old);
+            Warning("Append", "Replacing existing %s: %s (Potential memory leak).", old->IsA()->GetName(),
+                    old->GetName());
+         }
+      }
+
+      for (auto toRemove : toRemoveList) {
+         ROOT::DirAutoAdd_t func = toRemove->IsA()->GetDirectoryAutoAdd();
+         if (func && toRemove != obj) {
+            func(toRemove, nullptr);
```
pcanal, line 224: Last problem (which I don't think we can fix) is that if an histogram (that is not `obj` but has the same name as `obj`) is both in the current directory **and** is attached to a completely distinct `TDirectory`, this will result in that second histogram being removed from the other directory but not this one .... 

On the other hand this would mean the user did:
```
TFile f1(...);
auto h1 = new TH1F("histo", ....);
TFile f2(...);
auto h2 = new TH1F("histo", ....);
f2.Append(h1);
f2.Append(h2, true);
```
If I read the code correctly, this would result in `f1` being empty and `f2` containing both histograms.

pcanal: Also related, if `TH1::AddDirectoryStatus()` is false, the function is a no-op :(
i.e. this 'fails':
```
TH1::AddDirectory(false);
TFile f1(...);
auto h1 = new TH1F("histo", ....);
h1.SetDirectory(&f1);
auto h2 = new TH1F("histo", ....);
f1.Append(h2, true);
```

pcanal:
Why is the proposed addition to the user code:
```
   auto h = new TH1D(...);
   directory->Append(h, true);
```
and not
```
   auto h = new TH1D(...);
   h->SetDirectory(directory);
```
?
(`SetDirectory` is a no-op when it is called multiple time with the same argument)

hageboeck:
Superseded by #21648

PR #20889, opened 2026-01-14
[RF] Prepare RooFit and HistFactory for ROOT without implicit ownership

hageboeck:
This takes a few RooFit-related commits out of #18083.
- ~~The backend-related commit fixes the tests in case root has been configured with `-Dclad=off`.~~ Went to #20890 
- The other two move over to explicitly registering/writing histograms instead of relying on implicit ownership. This way, the tests are independent of whether implicit registration is used or not.

Note:
- [x] This contains the commit from #20888, so let's merge that one first.

### DIFF
--- a/roofit/roofitcore/test/testRooDataSet.cxx
+++ b/roofit/roofitcore/test/testRooDataSet.cxx
@@ -170,6 +170,7 @@ TEST(RooDataSet, ReducingData)
    for (int i = 0; i < 3; ++i) {
       // Check with root:
       TH1F test_hist(("h" + std::to_string(i)).c_str(), "histo", 10, massmin, massmax);
+      test_hist.SetDirectory(gDirectory); // TTree::Draw needs to find the histogram
       chi2cutval += 0.5;
 
       std::stringstream cutString;
@@ -439,6 +440,7 @@ TEST(RooDataSet, SplitDataSetWithWeightErrors)
 TEST(RooDataSet, ReadDataSetWithErrors626)
 {
    std::unique_ptr<TFile> file{TFile::Open("dataSet_with_errors_6_26_10.root", "READ")};
+   ASSERT_NE(file, nullptr);
 
    auto data = file->Get<RooDataSet>("data");
 


roofit/roofitcore/test/gtest_wrapper.h
```diff
@@ -32,8 +32,12 @@
 
 #define ROOFIT_EVAL_BACKENDS ROOFIT_EVAL_BACKEND_LEGACY ROOFIT_EVAL_BACKEND_CUDA RooFit::EvalBackend::Cpu()
 
+#ifdef ROOFIT_CLAD
```
hageboeck: The failure is:
In `gtest-roofit-histfactory-testHistFactory`
```
 [#0] ERROR:InputArguments -- RooFit was built without clad. Codegen backends are unavailable. Falling back to default.
314: 
314: [ FATAL ] /usr/include/gtest/internal/gtest-param-util.h:589:: Condition test_param_names.count(param_name) == 0 failed. Duplicate parameterized test name 'EquidistantBins_OverallSyst_Backend_cpu', in /home/stephan/code/root/roofit/histfactory/test/testHistFactory.cxx line 706
```

And in `gtest-roofit-roofitcore-testRooMinimizer`:
```
 [#0] ERROR:InputArguments -- RooFit was built without clad. Codegen backends are unavailable. Falling back to default.
387: 
387: [ FATAL ] /usr/include/gtest/internal/gtest-param-util.h:589:: Condition test_param_names.count(param_name) == 0 failed. Duplicate parameterized test name 'EvalBackendcpu', in /home/stephan/code/root/roofit/roofitcore/test/testRooMinimizer.cxx line 163
```

Running the preprocessor on the latter, I see that `ROOFIT_EVAL_BACKENDS_WITH_CODEGEN` expands to
`RooFit::EvalBackend::Legacy(), RooFit::EvalBackend::Cpu(), RooFit::EvalBackend::CodegenNoGrad()`

The return values of these functions are (when clad is off):
`Legacy, Cpu, Cpu`. The doubling of `Cpu` is why gtest aborts the test.

I fixed this by not adding `CodegenNoGrad()`, so the macro would expand to
`Legacy, Cpu`

guitargeek: Ok, then the problem is that `CodegenNoGrad()` unnecessarily falls back to `Cpu`. I'll see how I can fix this.

hageboeck: So to sum up:
The problem is that CodeNoGrad also expands to Cpu. That's why it helped to remove it.

guitargeek: PR opened:
  * https://github.com/root-project/root/pull/20890

hageboeck: OK, commit removed now.

hageboeck:
Github won't let me merge, because there's a supposedly unresolved comment that cannot be resolved.

Issue #20108, opened 2025-10-14
ROOT dictionary cannot be generated for classes with `static thread_local` members?


taehyounpark:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

I have a class that generates random weight variations. The toys are generated once per-event, where the toys are used by multiple histogram instances and event processing happens across multiple threads. This is a done using a ATLAS-wide used packaged called [BootstrapGenerator](https://gitlab.cern.ch/atlas-physics/sm/StandardModelTools_BootstrapGenerator). The full wrapper source code can be found [here](https://github.com/taehyounpark/queryosity/blob/master/extensions/queryosity/ROOT/hist_with_toys.hpp)

The most straightforward C++11 way to handle this is to have a `static thread_local` RNG to be used by all the thread-local histograms. Upon generating dictionary around the (wrapper) of a class containing such member, however, this instance cannot be instantiated by cling within an interactive root session, even though the dictionary is seemingly generated with the message.

```
runStaticInitializersOnce]: Failed to materialize symbols: { (main, { _ZTISt5_BindIFZN10 ...
```

Only when the `thread_local` specifier is removed as a dummy test, the class can be instantiated. This is, of course, not thread-local and not usable within the multithreaded workflow. A workaround would be to manually manage a `std::vector` of generators where each slot is accessed and called once per-event, but before I go about this more-verbose workaround, I wanted to ask if (lack of) support for classes with `thread_local` variables in ROOT dictionary generation was expected, or if it's a different mistake I might be making? If it's the former, would there be an easier way?

### Reproducer

(Please see above)

### ROOT version
```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.36.02                        https://root.cern |
  | (c) 1995-2025, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Aug 07 2025, 16:25:09                 |
  | From tags/v6-36-02@v6-36-02                                      |
  | With g++ (GCC) 13.1.0                                            |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------
```

### Installation method

LCG environment

### Operating system

Linux

ferdymercury:
Mm maybe this example with a simple class helps:

```
// MyClass.h

#ifndef MyClass_h
#define MyClass_h

#include "TRandom2.h"
#include <memory>
#include <vector>

class MyClass
{
public:
[...]

Issue #18014, opened 2025-03-16
Inconsistent and random behaviour in AsNumpy method from RDataFrame


acampove:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

It seems that `AsNumpy` (and prossibly other functions) process data in random and inconsistent orders when using multithreading.

### Reproducer

```python
from ROOT import EnableImplicitMT, RDataFrame

EnableImplicitMT(10)

rdf=RDataFrame(100_000)
rdf=rdf.Define('a', 'rdfentry_')
arr_dat = rdf.AsNumpy(['a'])['a']

print(arr_dat[:10])
print(arr_dat[10:])
```

The code above shows me:

```
(rx_data) [acampove@thinkbook enableMT]$ python test.py
[65000 65001 65002 65003 65004 65005 65006 65007 65008 65009]
[65010 65011 65012 ... 99997 99998 99999]
(rx_data) [acampove@thinkbook enableMT]$
(rx_data) [acampove@thinkbook enableMT]$
(rx_data) [acampove@thinkbook enableMT]$ python test.py
[90000 90001 90002 90003 90004 90005 90006 90007 90008 90009]
[90010 90011 90012 ... 24997 24998 24999]
```

I expect to see the first 10and the last 10 digits of the 0-100K sequence _every time_

### ROOT version

   ------------------------------------------------------------------
  | Welcome to ROOT 6.32.10                        https://root.cern |
  | (c) 1995-2024, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Feb 12 2025, 01:47:45                 |
  | From tags/6-32-10@6-32-10                                        |
  | With                                                             |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------

### Installation method

micromamba

### Operating system

Almalinux 9

### Additional context

This is a very dangerous bug and needs to be fixed ASAP.

dpiparo:
Thanks for reaching out. If I understand correctly, you see the order of the output events scrambled when going MT. Please correct me if I am wrong.

RDF relies on a multithreaded runtime for parallelism. As such, the order in which the ranges of events read (or generated) is processed is intrinsically random and not reproducible. This is not a feature of RDF, but in general of task parallelised applications, such as data processing frameworks of HEP experiments.

guitargeek:
[...]

Issue #12988, opened 2023-06-12
module.idx creation can easily cause missing symbols


Axel-Naumann:
### Check duplicate issues.

- [ ] Checked for duplicates

### Describe the bug

In an ongoing development we have an external pulled into a header (SYCL related); that header has a static variable. JITting that header triggers static initialization which need the library to be loaded to resolve the symbol. That library doesn't get loaded when generating module.idx causing a missing symbol. (We had a similar case with Vc in the past, IIRC.)

### What is the expected behaviour?

No error message when generating `module.idx`.

### How to reproduce?

This can be reproduced with a modified ROOT:
```diff
diff --git a/tree/dataframe/inc/ROOT/RDataFrame.hxx b/tree/dataframe/inc/ROOT/RDataFrame.hxx
index 7a094cc01f..d74c69aa2b 100644
--- a/tree/dataframe/inc/ROOT/RDataFrame.hxx
+++ b/tree/dataframe/inc/ROOT/RDataFrame.hxx
@@ -73,4 +73,6 @@ namespace cling {
 std::string printValue(ROOT::RDataFrame *tdf);
 } // ns cling
 
+static ROOT::RDataFrame thisWillBeBadForTheModulesIdx(17);
+
 #endif // ROOT_RDATAFRAME
```

This causes
```
[522/527] Generating lib/modules.idx
[...]

Issue #12190, opened 2023-01-31
Fix rdfentry_: it should represent the global entry number in the TChain even in MT runs


eguiraud:

eguiraud:
Currently `rdfentry_` does not correspond to the global TChain entry number in MT runs (see also [the relevant docs](https://root.cern/doc/master/classROOT_1_1RDataFrame.html#helper-cols)).

This is surprising for users (hence the big warning in the docs linked above).

We could instead make it match if only each MT task knew the offset of the current tree w.r.t. all other trees in the chain.

pcanal:
I suppose the proposed can work but require a synchronization step of sort.  The thread handling file number n, must wait until the file number [0,n-1] have been open before starting to process entries ; the file opening can be in parallel (somewhat) but still if an arbitrary file inside the chain is much smaller than the other, the thread processing it will have to wait a bit.

eguiraud:
In the proposed solution if a task needs an entry number that's not there yet it goes and retrieves it itself (subsequent tasks that need that entry number will then find it in the shared list).

pcanal:
If you are not carefully, this might be that at start up for n-core/n-tasks, we might issue (n(n+1) / 2) files opens (eg. the first file being requested to be open n times).

pcanal:
> we could do nothing: rdfentry_ would be unstable and it could not be relied upon to e.g. index into manually added "friend columns" 

Indeed, the global number is needed to load the proper friend.  For example we could have a friend which is a chain which contains files that have different lengths (number of entries)( (but same total lengths) than the files in the main chain (consequently a single file in the main chain maybe have to use/open 2 or more files from the friend chain).

Axel-Naumann:
Sounds good! What might help - asp during startup where I agree with @pcanal things can get a bit wild - is to tell the workers: "and report your tree's entries back, and I will then - at some point in the near future - let you know when I know your global offset". I.e. the sync step @pcanal was referring to. It makes sense to parallelize that!

eguiraud:
> If you are not carefully, this might lead to the case where at start up for n-core/n-tasks, we might issue (n(n+1) / 2) files opens (eg. at the very least the first file being requested to be open n times).

Mmmh that's right...we'll have to be careful.

> the global number is needed to load the proper friend

[...]

PR #20898, opened 2026-01-15
Add RDF example on thread-safe usage of RNGs

### LINKED ISSUE #20839: Examples of DefineSlot and RedefineSlot for python

I am having again problems with multithreading in python because I seem to be using `Define` and `Redefine`, which are not thread safe. There seem to be thread safe versions of these methods called `DefineSlot` and `RedefineSlot`. However I am struggling to make it work and I cannot find any example of their usage for python.

### ROOT version

6.38

### Installation method

mamba

### Operating system

almalinux9

### Additional context

_No response_

dudarboh:
# This Pull request:

## Changes or fixes:
   Add example on thread-safe usage of RNGs

## Checklist:

- [ ] tested changes locally

## Notes:

As discussed in the [forum post](https://root-forum.cern.ch/t/random-number-generation-with-rdataframe/64583) it might be a good idea to have an example of a thread-safe usage of random number generators (RNGs).

It is an easy pitfall to fall into, as quite a few people have struggled with thread-safety in MT mode judging by forum posts [here](https://root-forum.cern.ch/t/random-number-generation-with-rdataframe/64583), [here](https://root-forum.cern.ch/t/problem-with-rdataframe-multithreading/57723/3), [here](https://root-forum.cern.ch/t/using-random-with-rdataframe/58577/3), and [here](https://root-forum.cern.ch/t/tgraph-fit-usage-with-enableimplicitmt-is-tgraph-fit-thread-safe/64541/3).

It is also quite important to have an example usage for that because there is no error for existing race condition. Most of the forum discovered artifacts in their random distributions, which easily can be unnoticeable by eye.

### DIFF
--- a/tutorials/analysis/dataframe/index.md
+++ b/tutorials/analysis/dataframe/index.md
@@ -64,6 +64,7 @@ A collection of building block examples for your analysis.
 | df036_missingBranches.C | df036_missingBranches.py | Deal with missing values due to a missing branch when switching to a new file in a chain. |
 | df037_TTreeEventMatching.C | df037_TTreeEventMatching.py | Deal with missing values due to not finding a matching event in an auxiliary dataset. |
 | df040_RResultPtr_lifetimeManagement.C | | Lifetime management of RResultPtr and the underlying objects. |
+| df042_ThreadSafeRNG.C | df042_ThreadSafeRNG.py | Thread-safe usage of random number generators in multithreading. |
 
 
 \anchor readwrite
--- a/tutorials/analysis/dataframe/df042_ThreadSafeRNG.C
+++ b/tutorials/analysis/dataframe/df042_ThreadSafeRNG.C
@@ -1,0 +1,86 @@
+/// \file
+/// \ingroup tutorial_dataframe
+/// \notebook -nodraw
+/// Usage of multithreading mode with random generators.
+///
+/// This example illustrates how to define functions that generate random numbers and use them in an RDataFrame
+/// computation graph in a thread-safe way.
+///
+/// Using only one random number generator in an application running with ROOT::EnableImplicitMT() is a common pitfall.
+/// This pitfall creates race conditions resulting in a distorted random distribution. In the example, this issue is
+/// solved by creating one random number generator per RDataFrame processing slot, thus allowing for parallel and
+/// thread-safe access. The example also illustrates the difference between non-deterministic and deterministic random
+/// number generation.
+///
+/// \macro_code
+/// \macro_image
+/// \macro_output
+///
+/// \date February 2026
+/// \author Bohdan Dudar (JGU Mainz), Fernando Hueso-González (IFIC, CSIC-UV), Vincenzo Eduardo Padulano (CERN)
+
+#include <algorithm>
+#include <iostream>
+#include <memory>
+#include <thread>
+#include <TCanvas.h>
+#include <ROOT/RDataFrame.hxx>
+
+#include "df042_ThreadSafeRNG.hxx"
+
+// Canvas that should survive the running of this macro
+auto myCanvas = std::make_unique<TCanvas>("myCanvas", "myCanvas", 1000, 500);
+
+void df042_ThreadSafeRNG()
+{
+   myCanvas->Divide(3, 1);
+
+   unsigned int nEntries{10000000};
+
+   // 1. Single thread for reference
+   auto df1 = ROOT::RDataFrame(nEntries).Define("x", GetNormallyDistributedNumberFromGlobalGenerator);
+   auto h1 = df1.Histo1D({"h1", "Single thread (no MT)", 1000, -4, 4}, {"x"});
+   myCanvas->cd(1);
+   h1->Draw();
+
+   // 2. One generator per RDataFrame slot, with random_device seeding
+   // Notes and Caveats:
+   // - How many numbers are drawn from each generator is not deterministic
+   //   and the result is not deterministic between runs.
+   unsigned int nSlots{std::max(2u, std::thread::hardware_concurrency() / 4u)};
+   ROOT::EnableImplicitMT(nSlots);
+   // Before running the RDataFrame computation graph, we reinitialize the generators (one per slot), so they can
+   // be used accordingly during the execution.
+   ReinitializeGenerators(nSlots);
+   auto df2 = ROOT::RDataFrame(nEntries).DefineSlot("x", GetNormallyDistributedNumberPerSlotGenerator);
+   auto h2 = df2.Histo1D({"h2", "Thread-safe (MT, non-deterministic)", 1000, -4, 4}, {"x"});
+   myCanvas->cd(2);
+   h2->Draw();
+
+   // 3. One generator per RDataFrame slot, with entry seeding
+   // Notes and Caveats:
+   // - With RDataFrame(INTEGER_NUMBER) constructor (as in the example),
+   //   the result is deterministic and identical on every run
+   // - With RDataFrame(TTree) constructor, the result is not guaranteed to be deterministic.
+   //   To make it deterministic, use something from the dataset to act as the event identifier
+   //   instead of rdfentry_, and use it as a seed.
+
+   // Before running the RDataFrame computation graph, we reinitialize the generators (one per slot), so they can
+   // be used accordingly during the execution.
+   ReinitializeGenerators(nSlots);
+   auto df3 = ROOT::RDataFrame(nEntries).DefineSlotEntry("x", GetNormallyDistributedNumberPerSlotGeneratorForEntry);
+   auto h3 = df3.Histo1D({"h3", "Thread-safe (MT, deterministic)", 1000, -4, 4}, {"x"});
+   myCanvas->cd(3);
+   h3->Draw();
+
+   std::cout << std::fixed << std::setprecision(3) << "Final distributions                : " << "Mean " << " +- "
+             << "StdDev" << std::endl;
+   std::cout << std::fixed << std::setprecision(3) << "Theoretical                        : " << "0.000" << " +- "
+             << "1.000" << std::endl;
+   std::cout << std::fixed << std::setprecision(3) << "Single thread (no MT)              : " << h1->GetMean() << " +- "
+             << h1->GetStdDev() << std::endl;
+   std::cout << std::fixed << std::setprecision(3) << "Thread-safe (MT, non-deterministic): " << h2->GetMean() << " +- "
+             << h2->GetStdDev() << std::endl;
+   std::cout << std::fixed << std::setprecision(3) << "Thread-safe (MT, deterministic)    : " << h3->GetMean() << " +- "
+             << h3->GetStdDev() << std::endl;
+}
--- a/tutorials/analysis/dataframe/df042_ThreadSafeRNG.hxx
+++ b/tutorials/analysis/dataframe/df042_ThreadSafeRNG.hxx
@@ -1,0 +1,55 @@
+#ifndef ROOT_TUTORIALS_ANALYSIS_DATAFRAME_DF042
+#define ROOT_TUTORIALS_ANALYSIS_DATAFRAME_DF042
+
+#include <random>
+
+// NOTE: these globals are intentionally NOT protected by a mutex.
+// This function is only safe to call from a single thread (used as reference).
+inline std::random_device globalRandomDevice{};
+inline std::mt19937 globalGenerator(globalRandomDevice());
+inline std::normal_distribution<double> globalGaus(0., 1.);
+
+double GetNormallyDistributedNumberFromGlobalGenerator()
+{
+   return globalGaus(globalGenerator);
+}
+
+// One generator per slot — initialized once before the event loop
+// An alternative to these global vectors could be to have thread_local
+// variables within a function scope and then call that function from RDataFrame
+inline std::vector<std::mt19937> generators;
+inline std::vector<std::normal_distribution<double>> gaussians;
+
+void ReinitializeGenerators(unsigned int nSlots)
+{
+   std::random_device rd;
+   generators.resize(nSlots);
+   for (auto &gen : generators)
+      gen.seed(rd());
+   gaussians.resize(nSlots, std::normal_distribution<double>(0., 1.));
+}
+
+double GetNormallyDistributedNumberPerSlotGenerator(unsigned int slot)
+{
+   return gaussians[slot](generators[slot]);
+}
+
+double GetNormallyDistributedNumberPerSlotGeneratorForEntry(unsigned int slot, unsigned long long entry)
+{
+   // We want to generate a random number distributed according to a normal distribution in a thread-safe way and such
+   // that it is reproducible across different RDataFrame runs, i.e. given the same input to the generator it will
+   // produce the same value. This is one way to do it. It assumes that the input argument represents a unique entry ID,
+   // such that any thread processing an RDataFrame task will see this number once throughout the entire execution of
+   // the computation graph
+   // Calling both `reset` and `seed` methods is fundamental here to ensure reproducibility: without them the same
+   // generator could be seeded by a different entry (depending on which is the first entry ID seen by a thread) or
+   // could be at a different step of the sequence (depending how many entries this particular thread is processing).
+   // Alternatively, if both the generator and the distribution objects were recreated from scratch at every function
+   // call (i.e. by removing the global std::vector stores), then the next two method calls would not be necessary, at
+   // the cost of a possible performance degradation.
+   gaussians[slot].reset();
+   generators[slot].seed(entry);
+   return gaussians[slot](generators[slot]);
+}
+
+#endif // ROOT_TUTORIALS_ANALYSIS_DATAFRAME_DF041
--- a/tutorials/analysis/dataframe/df042_ThreadSafeRNG.py
+++ b/tutorials/analysis/dataframe/df042_ThreadSafeRNG.py
@@ -1,0 +1,89 @@
+# \file
+# \ingroup tutorial_dataframe
+# \notebook -nodraw
+# Usage of multithreading mode with random generators.
+#
+# This example illustrates how to define functions that generate random numbers and use them in an RDataFrame
+# computation graph in a thread-safe way.
+#
+# Using only one random number generator in an application running with ROOT.EnableImplicitMT() is a common pitfall.
+# This pitfall creates race conditions resulting in a distorted random distribution. In the example, this issue is
+# solved by creating one random number generator per RDataFrame processing slot, thus allowing for parallel and
+# thread-safe access. The example also illustrates the difference between non-deterministic and deterministic random
+# number generation.
+#
+# \macro_code
+# \macro_image
+# \macro_output
+#
+# \date February 2026
+# \author Bohdan Dudar (JGU Mainz), Fernando Hueso-González (IFIC, CSIC-UV), Vincenzo Eduardo Padulano (CERN)
+
+import os
+
+import ROOT
+
+
+def df042_ThreadSafeRNG():
+
+    header_path = os.path.join(str(ROOT.gROOT.GetTutorialDir()), "analysis", "dataframe", "df042_ThreadSafeRNG.hxx")
+    if not os.path.exists(header_path):
+        raise RuntimeError(f'Could not find required header file "{header_path}".')
+
+    # First, we declare the functions needed by the RDataFrame computation graph to the interpreter
+    if not ROOT.gInterpreter.Declare(f'#include "{header_path}"'):
+        raise RuntimeError("Failed to declare the functions needed by the RDataFrame computation graph.")
+
+    myCanvas = ROOT.TCanvas("myCanvas", "myCanvas", 1000, 500)
+    myCanvas.Divide(3, 1)
+
+    nEntries = 10000000
+
+    # 1. Single thread for reference
+    df1 = ROOT.RDataFrame(nEntries).Define("x", ROOT.GetNormallyDistributedNumberFromGlobalGenerator)
+    h1 = df1.Histo1D(("h1", "Single thread (no MT)", 1000, -4, 4), "x")
+    myCanvas.cd(1)
+    h1.Draw()
+
+    # 2. One generator per RDataFrame slot, with random_device seeding
+    # Notes and Caveats:
+    # - How many numbers are drawn from each generator is not deterministic
+    #   and the result is not deterministic between runs.
+    nSlots = max(2, os.cpu_count() // 4)
+    ROOT.EnableImplicitMT(nSlots)
+    # Before running the RDataFrame computation graph, we reinitialize the generators (one per slot), so they can
+    # be used accordingly during the execution.
+    ROOT.ReinitializeGenerators(nSlots)
+    df2 = ROOT.RDataFrame(nEntries).DefineSlot("x", ROOT.GetNormallyDistributedNumberPerSlotGenerator)
+    h2 = df2.Histo1D(("h2", "Thread-safe (MT, non-deterministic)", 1000, -4, 4), "x")
+    myCanvas.cd(2)
+    h2.Draw()
+
+    # 3. One generator per RDataFrame slot, with entry seeding
+    # Notes and Caveats:
+    # - With RDataFrame(INTEGER_NUMBER) constructor (as in the example),
+    #   the result is deterministic and identical on every run
+    # - With RDataFrame(TTree) constructor, the result is not guaranteed to be deterministic.
+    #   To make it deterministic, use something from the dataset to act as the event identifier
+    #   instead of rdfentry_, and use it as a seed.
+    # Before running the RDataFrame computation graph, we reinitialize the generators (one per slot), so they can
+    # be used accordingly during the execution.
+    ROOT.ReinitializeGenerators(nSlots)
+    df3 = ROOT.RDataFrame(nEntries).DefineSlotEntry("x", ROOT.GetNormallyDistributedNumberPerSlotGeneratorForEntry)
+    h3 = df3.Histo1D(("h3", "Thread-safe (MT, deterministic)", 1000, -4, 4), "x")
+    myCanvas.cd(3)
+    h3.Draw()
+
+    print(f"{'{:<40}'.format('Final distributions')}: Mean +- StdDev")
+    print(f"{'{:<40}'.format('Theoretical')}: 0.000 +- 1.000")
+    print(f"{'{:<40}'.format('Single thread (no MT)')}: {h1.GetMean():.3f} +- {h1.GetStdDev():.3f}")
+    print(f"{'{:<40}'.format('Thread-safe (MT, non-deterministic)')}: {h2.GetMean():.3f} +- {h2.GetStdDev():.3f}")
+    print(f"{'{:<40}'.format('Thread-safe (MT, deterministic)')}: {h3.GetMean():.3f} +- {h3.GetStdDev():.3f}")
+
+    # We draw the canvas with block=True to stop the execution before end of the
+    # function and to be able to interact with the canvas until necessary
+    myCanvas.Draw(block=True)
+
+
+if __name__ == "__main__":
+    df042_ThreadSafeRNG()


dudarboh:
Possible output:

image

```
Warning in <ROOT_TImplicitMT_DisableImplicitMT>: Implicit multi-threading is already disabled
Final distributions        : Mean  +- STD  
Theoretical                : 0.000 +- 1.000
Single thread       (no MT): 0.000 +- 0.999
Race condition         (MT): 0.000 +- 1.013
Thread-safe            (MT): -0.000 +- 1.000
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,82 @@
...
+   h1->DrawClone();
+
+   // 2. Enable MT and generate random variables with global generator
+   // NOTE: Causes race conditions! Don't do it in your code!
+   // More cores - more noticeable effect
+   ROOT::EnableImplicitMT(32);
+   auto df2 = ROOT::RDataFrame(10000000).Define("x", GetRNG_global);
+   auto h2 = df2.Histo1D({"h3", "Global generator (MT)", 1000, -4, 4}, {"x"});
+   c1->cd(2);
+   h2->DrawClone();
```
vepadulano: I am not a fan of showing wrong approaches in tutorial, I'd rather remove this part and just show the right thing to do.

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,82 @@
...
+   return gaus(generator);
+}
+
+void df041_ThreadSafeRNG()
+{
+   c1 = new TCanvas("c1", "c1", 1500, 500);
+   c1->Divide(3, 1);
+
+   // 1. Single thread for reference
+   ROOT::DisableImplicitMT();
```
vepadulano:
```suggestion
```

This can be removed as there is no previous call to EnableImplicitMT()

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,82 @@
...
+   // NOTE: Causes race conditions! Don't do it in your code!
+   // More cores - more noticeable effect
+   ROOT::EnableImplicitMT(32);
+   auto df2 = ROOT::RDataFrame(10000000).Define("x", GetRNG_global);
+   auto h2 = df2.Histo1D({"h3", "Global generator (MT)", 1000, -4, 4}, {"x"});
+   c1->cd(2);
+   h2->DrawClone();
+
+   // 2. Generate random variables with several per-thread generators
+   // NOTE: Do this!
```
vepadulano:
```suggestion
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,82 @@
...
+   auto df3 = ROOT::RDataFrame(10000000).DefineSlot("x", GetRNG_thread_safe);
```
vepadulano: Re-add the call to `EnableImplicitMT()` here after removing the lines above

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,82 @@
...
+{
+   return global_gaus(global_generator);
+}
+
+double GetRNG_thread_safe(unsigned int slot)
+{
+   static thread_local std::mt19937 generator(rand());
+   static thread_local std::normal_distribution<double> gaus(0., 1.);
+   return gaus(generator);
+}
```
vepadulano: nitpick: there is a bit of an inconsistent usage of CamelCase and snake_case here, probably we should choose the first for consistency with the rest of ROOT
```suggestion
std::mt19937 globalGenerator(rand());
std::normal_distribution<double> globalGaus(0., 1.);

double GetGlobalRNG()
{
   return globalGaus(globalGenerator);
}

double GetThreadSafeRNG(unsigned int slot)
{
   static thread_local std::mt19937 generator(rand());
   static thread_local std::normal_distribution<double> gaus(0., 1.);
   return gaus(generator);
}
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,82 @@
...
+/// \macro_image
+/// \macro_output
+///
+/// \date 2025
+/// \author Bohdan Dudar (JGU Mainz) and Fernando Hueso-González (IFIC)
+
+#include <iostream>
+#include <thread>
+#include <random>
+#include "TRandom3.h"
```
vepadulano:
```suggestion
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,82 @@
+/// \file
+/// \ingroup tutorial_dataframe
+/// \notebook -nodraw
+/// Usage of multithreading mode with random generators.
+///
+/// This tutorial illustrates how to make a thread-safe program with thread-local random number engines.
+/// Using only one random number generator in a ROOT::EnableImplicitMT() mode is a common pitfal.
```
vepadulano:
```suggestion
/// Using only one random number generator in a ROOT::EnableImplicitMT() mode is a common pitfall.
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,82 @@
...
+/// This pitfal creates race conditions resulting in a distorded random distribution.
```
vepadulano:
```suggestion
/// This pitfall creates race conditions resulting in a distorted random distribution.
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,82 @@
...
+/// This example illustrates how to define random number generators thread-safely avoid race condition.
```
vepadulano:
```suggestion
/// This example illustrates how to define thread-safe functions that generate random numbers and use them in an RDataFrame computation graph.
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,82 @@
...
+/// This example also illustrates the difference between wrong and correct distributions.
```
vepadulano: I don't think this is a good practice, we should only show the good approach in the tutorial
```suggestion
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
vepadulano: The thread-safety of `std::rand` [is implementation-defined](https://en.cppreference.com/w/cpp/numeric/random/rand.html), I suggest this other implementation for this function

```suggestion
inline double GetThreadSafeRNG()
{
   thread_local std::random_device rd{};
   thread_local std::mt19937 gen{rd()};
   thread_local std::normal_distribution<double> gaus{0., 1.};
   return gaus(gen);
}
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,82 @@
...
+#include <thread>
+#include <random>
+#include "TRandom3.h"
+#include "TCanvas.h"
+#include "ROOT/RDataFrame.hxx"
+
+// Canvas that should survive the running of this macro:
+TCanvas *c1;
+
+std::mt19937 global_generator(rand());
```
vepadulano: Use `random_device` for consistency with the proposed improvement to the thread-safe version
```suggestion
std::random_device globalRandomDevice{};
std::mt19937 globalGenerator{globalRandomDevice()};
```

vepadulano (changes requested):
Thank you, this is a nice tutorial to add! I have a few suggestions for improvement.

tutorials/analysis/dataframe/index.md
```diff
@@ -64,6 +64,7 @@ A collection of building block examples for your analysis.
 | df036_missingBranches.C | df036_missingBranches.py | Deal with missing values due to a missing branch when switching to a new file in a chain. |
 | df037_TTreeEventMatching.C | df037_TTreeEventMatching.py | Deal with missing values due to not finding a matching event in an auxiliary dataset. |
 | df040_RResultPtr_lifetimeManagement.C | | Lifetime management of RResultPtr and the underlying objects. |
+| df041_ThreadSafeRNG.C | | Thread-safe usage of random number generators in multithreading. |
```
ferdymercury: Thanks for this! It would be great to have the same tutorial in Python:

```
import ROOT
ROOT.gInterpreter.Declare('double myfunctor(UInt_t rdfslot) { thread_local TRandom3 r3(rdfslot); return r3.Gaus();}') # or replace with std::random_device ...
rdf = ROOT.RDataFrame(90000000)
rdf_x = rdf.Define("x", "myfunctor(rdfslot_)")
h = rdf_x.Histo1D("x")
h.Draw()
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
ferdymercury: I guess

`thread_local std::mt19937 gen{slot};`

could be an alternative way if you'd want reproducible seeding.

tutorials/analysis/dataframe/index.md
vepadulano: Agreed, I'd rather have it in a separate PR though :+1:

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
vepadulano: Yes, that is a valid alternative

dudarboh: Shouldn't we use `static` here to avoid re-initialization on every function call?

ferdymercury: It's equivalent: you can use static if you want to be more vebose / explicity (which might be good for a tutorial).

(Note: thread_local becomes automatically static (it's implicit). Just static does not become thread_local

https://stackoverflow.com/questions/22794382/are-c11-thread-local-variables-automatically-static

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,71 @@
...
+std::random_device globalRandomDevice{};
+std::mt19937 globalGenerator(globalRandomDevice());
+std::normal_distribution<double> globalGaus(0., 1.);
+
+double GetGlobalRNG()
+{
+   return globalGaus(globalGenerator);
+}
+
+double GetThreadSafeRNG(unsigned int slot)
```
vepadulano: The argument should not be named if it's not used in the function to avoid compiler warnings
```suggestion
double GetThreadSafeRNG(unsigned int)
```

vepadulano (changes requested):
Just a small change

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,71 @@
...
+std::random_device globalRandomDevice{};
+std::mt19937 globalGenerator(globalRandomDevice());
+std::normal_distribution<double> globalGaus(0., 1.);
+
+double GetGlobalRNG()
+{
+   return globalGaus(globalGenerator);
+}
+
+double GetThreadSafeRNG(unsigned int)
```
ferdymercury: I guess that if you don't use rdfslot as seed, then you can fully remove the argument and call Define instead of DefineSlot

Alternatively, define GetThreadSafeRNGFixedSeeds(uint) to have both variants, one with Define and another with DefineSlot (or mention as comment)

ferdymercury (approved):
Thanks a lot!

hahnjo:
I think this tutorial is not ideal: By using `std::random_device` the results are not reproducible between runs. Even worse, there is no guarantee for the scheduling of multithreaded RDF, so it is not deterministic how many numbers are drawn from which generator (also because the user can change the number of threads). In a real application, I would expect this to be even more problematic where it's probably required to get the same random numbers for each RDF entry - in particular if data is loaded from an input file.

tldr I'm not going to veto the tutorial, but I would strongly suggest to think about reproducibility and whether we want to show something that may be even more confusing for users.

vepadulano:
> In a real application, I would expect this to be even more problematic where it's probably required to get the same random numbers for each RDF entry - in particular if data is loaded from an input file.

I disagree that this is the only correct representation of a realistic scenario.

>  I would strongly suggest to think about reproducibility

In principle, we don't have to. It's a perfectly valid example to show thread-safety.

I agree that we *could* show something more reproducible by using the slot of each thread, but this still wouldn't be 100% reproducible because as you said there is no guarantee on the scheduling and thus association between threads and slots.

ferdymercury:
> I think this tutorial is not ideal: By using `std::random_device` the results are not reproducible between runs.

Do you think the proposed alternative using DefineSlot would be more convenient? https://github.com/root-project/root/pull/20898#discussion_r2694151136

> where it's probably required to get the same random numbers for each RDF entry - in particular if data is loaded from an input file.

Would an additional example for entry-wise reproducibility using DefineSlotEntry
and using rdfentry_ as seed instead of rdfslot ? ie using a threadlocal object but calling setSeed each time. 

However, this might be problematic too due to rdfentry issue: https://github.com/root-project/root/issues/12190 / https://github.com/root-project/root/issues/18014

vepadulano:
> Would an additional example for entry-wise reproducibility using DefineSlotEntry
> and using rdfentry_ as seed instead of rdfslot ? ie using a threadlocal object but calling setSeed each time.

> However, this might be problematic too due to rdfentry issue: https://github.com/root-project/root/issues/12190 / https://> github.com/root-project/root/issues/18014

Yes, see my comment https://github.com/root-project/root/pull/20898#issuecomment-3759481168 . Also, that is not an issue, it's a feature request.

hahnjo:
> In a real application, I would expect this to be even more problematic where it's probably required to get the same random numbers for each RDF entry - in particular if data is loaded from an input file.

Clarification because I now realize it could be ambiguous how I wrote it: I didn't mean to say "the same random numbers for *every* RDF entry", but rather that given an entry / "event" a common requirement is that they "see" the same sequence of random numbers. For example event 1 has numbers `a, b, c` and event 2 has `d, e, f`, no matter how they are scheduled. AFAICT that's only possible to achieve with seeding per event or block of events (that have to be identical on every run).

> I disagree that this is the only correct representation of a realistic scenario.

Sure, but it's a very common requirement. Debugging a non-deterministic workflow is a nightmare.

vepadulano:
> AFAICT that's only possible to achieve with seeding per event or block of events (that have to be identical on every run).

I agree, that's also what I meant in my previous comment, and that is just not possible in RDataFrame, see https://github.com/root-project/root/issues/12190

Further clarification: it's not possible using RDataFrame-generated information. It's possible if the dataset itself has information that can either give a seed per event (e.g. an event ID), or that can act as a possible proxy for a pseudo random number generation. For an example of the latter, see what we do in the AGC https://github.com/root-project/analysis-grand-challenge/blob/ee7946440d27a730d67a366ee495ca5c0735a57e/analyses/cms-open-data-ttbar/helpers.h#L16C35-L23

vepadulano:
This slightly tweaked example shows something along the lines of what described by @hahnjo 

```
double GetNormallyDistributedNumberForEntry(unsigned long long entry)
{
   std::mt19937 generator(entry);
   std::normal_distribution<double> gaus(0., 1.);
   return gaus(generator);
}

void run(){
    ROOT::EnableImplicitMT(NSLOTS);
    auto df = ROOT::RDataFrame(NENTRIES).Define("x", GetNormallyDistributedNumberForEntry, {"rdfentry_"});
    auto h = df2.Histo1D({"h", "Generate same random number for the same entry value", 1000, -4, 4}, {"x"});
```

Caveats:

* Cannot use `thread_local` generator anymore, we need to ensure each entry is used to seed a different generator, meaning we'll have many more generators created (one per entry).
* The example above only works for the constructor `RDataFrame(INTEGER_NUMBER)` for which every processing slot will have a distinct and non-overlapping range of entries to process (i.e. no two slots will ever process the same entry number), unlike what happens for every other data source, as already described by https://github.com/root-project/root/issues/12190


I will let the author @dudarboh decide whether the example above, is reasonable enough to end up in this tutorial. In which case, my suggestion would  be to include this example together with the other two examples in the tutorial and explain the difference, with the caveats from above.

ferdymercury:
> * Cannot use `thread_local` generator anymore, we need to ensure each entry is used to seed a different generator, meaning we'll have many more generators created (one per entry).

This mixed approach might avoid creation/destruction of the mt19937?

```
double GetNormallyDistributedNumberForEntry(unsigned long long entry)
{
   thread_local std::mt19937 generator;
   thread_local std::normal_distribution<double> gaus(0., 1.);
   generator.seed(entry);
   return gaus(generator);
}
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,98 @@
...
+   thread_local std::random_device rd{};
+   thread_local std::mt19937 generator(rd());
+   thread_local std::normal_distribution<double> gaus(0., 1.);
+   return gaus(generator);
+}
+
+double GetNormallyDistributedNumberForEntry(unsigned long long entry)
+{
+   std::mt19937 generator(entry);
+   std::normal_distribution<double> gaus(0., 1.);
```
ferdymercury:
```suggestion
   thread_local std::mt19937 generator;
   thread_local std::normal_distribution<double> gaus(0., 1.);
   generator.seed(entry);
   gaus.reset();
```
Maybe this is more efficient (didn't test though).

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,98 @@
...
+   c1->Divide(3, 1);
+
+   // 1. Single thread for reference
+   auto df1 = ROOT::RDataFrame(10000000).Define("x", GetGlobalNormallyDistributedNumber);
+   auto h1 = df1.Histo1D({"h1", "Single thread (no MT)", 1000, -4, 4}, {"x"});
+   c1->cd(1);
+   h1->DrawClone();
+
+   // 2. thread_local generators with random_device seeding
+   // Notes and Caveates:
```
ferdymercury:
```suggestion
   // Notes and Caveats:
```

dudarboh:
Thanks everybody for the discussion. It is, indeed, more nuanced  than I initially thought 😅.
I have incorporated @vepadulano example, and added some explanations, as far as I understood them.
Let me know if you have any additional comments.

> This mixed approach might avoid creation/destruction of the mt19937?

I think this still has a problem of being non-deterministic between runs, because each individual generator is not guaranteed to be called exactly the same number of times between runs and for exactly the same subset of events.
Creating a new `std::mt19937` and (!) `std::normal_distribution` for each entry makes it reproducible.

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,98 @@
...
+   // - how many numbers are drawn from each generator is not deterministic
+   //   and the result is not deterministic between runs
+   ROOT::EnableImplicitMT(32);
+   auto df2 = ROOT::RDataFrame(10000000).Define("x", GetNormallyDistributedNumber);
+   auto h2 = df2.Histo1D({"h2", "Thread-safe (MT, non-deterministic)", 1000, -4, 4}, {"x"});
+   c1->cd(2);
+   h2->DrawClone();
+
+   // 3. thread_local generators with entry seeding
+   // Notes and Caveates:
```
ferdymercury:
```suggestion
   // Notes and Caveats:
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,98 @@
...
+   // - With RDataFrame(INTEGER_NUMBER) constructor (as in the example),
+   //   the result is deterministic and identical on every run
+   // - With RDataFrame(TTree) constructor, the result is not guaranteed to be deterministic.
+   //   To make it deterministic, use something from the dataset to act as the event identifier
+   //   instead of rdfentry_, and use it as a seed.
+   // - requires constructing a new generator for each entry, which may have performance
```
ferdymercury:
```suggestion
   // - requires reseeding generator for each entry, which may have performance
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
dudarboh: I think this still has a problem of being non-deterministic between runs, because each individual generator is not guaranteed to be called exactly the same number of times between runs and for exactly the same subset of events.
Creating a new std::mt19937 and (!) std::normal_distribution for each entry makes it reproducible.

ferdymercury: > and (!) std::normal_distribution

Hmm I see. Could you try with gaus.reset() ?

dudarboh:
@ferdymercury suggestion with `gaus.reset()` seems to work 👀 and gives reproducible results, regardless of scheduling.

```cpp
double GetNormallyDistributedNumberForEntry(unsigned long long entry)
{
   thread_local std::mt19937 generator;
   thread_local std::normal_distribution<double> gaus(0., 1.);
   gaus.reset();
   generator.seed(entry);
   return gaus(generator);
}

void test(){
    ROOT::EnableImplicitMT(3);
    auto df = ROOT::RDataFrame(7).Define("x", GetNormallyDistributedNumberForEntry, {"rdfentry_"});
    auto print = [](double x, unsigned long long entry, unsigned int slot){
        auto str = TString();
        str.Form("Slot: %d   Seed: %llu  Value: %f\n", slot, entry, x);
        std::cout << str.Data();
    };
    df.Foreach(print, {"x", "rdfentry_", "rdfslot_"});
}
```

Example of two runs:

```
# RUN 1
Slot: 0   Seed: 0  Value: 1.122795
Slot: 1   Seed: 4  Value: 0.318870  
Slot: 0   Seed: 1  Value: -0.549746
Slot: 2   Seed: 2  Value: -0.022565
Slot: 1   Seed: 6  Value: -0.176084
Slot: 0   Seed: 5  Value: 0.107794  
Slot: 1   Seed: 3  Value: 0.184128  # <- 3rd call from thread 1 for entry 3
# RUN 2
Slot: 0   Seed: 4  Value: 0.318870 
Slot: 1   Seed: 0  Value: 1.122795
Slot: 2   Seed: 2  Value: -0.022565
Slot: 1   Seed: 1  Value: -0.549746
Slot: 0   Seed: 5  Value: 0.107794  
Slot: 2   Seed: 3  Value: 0.184128   # <- 2nd call from thread 2 for entry 3 
Slot: 1   Seed: 6  Value: -0.176084

```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,100 @@
...
+   // 1. Single thread for reference
+   auto df1 = ROOT::RDataFrame(10000000).Define("x", GetGlobalNormallyDistributedNumber);
+   auto h1 = df1.Histo1D({"h1", "Single thread (no MT)", 1000, -4, 4}, {"x"});
+   c1->cd(1);
+   h1->DrawClone();
+
+   // 2. thread_local generators with random_device seeding
+   // Notes and Caveats:
+   // - how many numbers are drawn from each generator is not deterministic
+   //   and the result is not deterministic between runs
```
ferdymercury:
```suggestion
   //   and the result is not deterministic between runs
   //   even if one seeded each generator with its rdfslot index.
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,100 @@
...
+/// This pitfall creates race conditions resulting in a distorted random distribution.
+/// This example illustrates how to define thread-safe functions that generate random numbers and use them in an
+/// RDataFrame computation graph.
+///
+/// \macro_code
+/// \macro_image
+/// \macro_output
+///
+/// \date 2025
+/// \author Bohdan Dudar (JGU Mainz) and Fernando Hueso-González (IFIC)
```
ferdymercury:
```suggestion
/// \author Bohdan Dudar (JGU Mainz) and Fernando Hueso-González (IFIC, CSIC-UV)
```
thks!

ferdymercury (approved):
Thanks a lot!
Just some tiny (optional) suggestions from my side

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,101 @@
...
+   return gaus(generator);
+}
+
+double GetNormallyDistributedNumberForEntry(unsigned long long entry)
+{
+   thread_local std::mt19937 generator;
+   thread_local std::normal_distribution<double> gaus(0., 1.);
+   gaus.reset();
+   generator.seed(entry);
+   return gaus(generator);
```
vepadulano:
```suggestion
   // We want to generate a random number distributed according to a normal distribution in a thread-safe way and such that it is reproducible across different RDataFrame runs, i.e. given the same input to the generator it will produce the same value.
   // This is one way to do it. It assumes that the input argument represents a unique entry ID, such that any thread processing an RDataFrame task will see this number once throughout the entire execution of the computation graph
   thread_local std::mt19937 generator;
   thread_local std::normal_distribution<double> gaus(0., 1.);
   // Calling both `reset` and `seed` methods is fundamental here to ensure reproducibility: without them the same generator could be seeded by a different entry (depending on which is the first entry ID seen by a thread) or could be at a different step of the sequence (depending how many entries this particular thread is processing).
   // Alternatively, if both the generator and the distribution objects were recreated from scratch at every function call (i.e. by removing the `thread_local` attribute), then the next two method calls would not be necessary, at the cost of a possible performance degradation.
   gaus.reset();
   generator.seed(entry);
   return gaus(generator);
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.C
```diff
@@ -0,0 +1,101 @@
...
+   h2->DrawClone();
+
+   // 3. thread_local generators with entry seeding
+   // Notes and Caveats:
+   // - With RDataFrame(INTEGER_NUMBER) constructor (as in the example),
+   //   the result is deterministic and identical on every run
+   // - With RDataFrame(TTree) constructor, the result is not guaranteed to be deterministic.
+   //   To make it deterministic, use something from the dataset to act as the event identifier
+   //   instead of rdfentry_, and use it as a seed.
+   // - requires reseeding generator for each entry, which may have performance
```
vepadulano: I would leave the performance implication part local to the function scope, see other comment
```suggestion
```

vepadulano (changes requested):

vepadulano:
Weirdly enough, the errors appearing in the mac-beta CI runs
```
[runStaticInitializersOnce]: Failed to materialize symbols: { (main, { __ZNSt3__112__destroy_atB8ne200100IN4ROOT6Detail3RDF7RDefineIPFdyENS3_18ExtraArgsForDefine4NoneEEETnNS_9enable_ifIXntsr8is_arrayIT_EE5valueEiE4typeELi0EEEvPSB_, ...
```

Are not reproducible **on the same mac machine with the same CMake configuration and the same environment variables**.

I am trying to address this inconsistency, until then please don't push more commits to this branch so I can rerun the CI when necessary.

ferdymercury:
Maybe it only appears in interpreted vs compiled? A bit related: https://github.com/root-project/root/issues/20108 https://github.com/root-project/root/issues/12988

vepadulano:
> Maybe it only appears in interpreted vs compiled? A bit related: https://github.com/root-project/root/issues/20108 https://github.com/root-project/root/issues/12988

Could be, but I'm just doing `ctest -j 8 -R tutorial-analysis-dataframe` and when I manually run on the CI machine everything passes. I'm not touching anything in the source files.

ferdymercury:
tagging @bellenot since it's also failing in Windows (with a different message)

vepadulano:
@dudarboh @ferdymercury I have updated the tutorial with code that makes it work. I also took the chance to refactor a few things, moving the RNG functionality to a header and add a Python equivalent tutorial that makes use of that header as well. I am about to run the CI, please do not push further changes to the branch. Feel free to review/comment meanwhile though.

tutorials/analysis/dataframe/index.md
ferdymercury:
```suggestion
| df041_ThreadSafeRNG.C | df041_ThreadSafeRNG.py | Thread-safe usage of random number generators in multithreading. |
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.hxx
```diff
@@ -0,0 +1,53 @@
...
+   // Calling both `reset` and `seed` methods is fundamental here to ensure reproducibility: without them the same
+   // generator could be seeded by a different entry (depending on which is the first entry ID seen by a thread) or
+   // could be at a different step of the sequence (depending how many entries this particular thread is processing).
+   // Alternatively, if both the generator and the distribution objects were recreated from scratch at every function
+   // call (i.e. by removing the `thread_local` attribute), then the next two method calls would not be necessary, at
+   // the cost of a possible performance degradation.
+   gaussians[slot].reset();
+   generators[slot].seed(entry);
+   return gaussians[slot](generators[slot]);
+}
```
ferdymercury, line 53:
```suggestion
double GetNormallyDistributedNumberPerSlotGeneratorForEntry(unsigned int slot, unsigned long long entry)
{
   // We want to generate a random number distributed according to a normal distribution in a thread-safe way and such
   // that it is reproducible across different RDataFrame runs, i.e. given the same input to the generator it will
   // produce the same value. This is one way to do it. It assumes that the input argument represents a unique entry ID,
   // such that any thread processing an RDataFrame task will see this number once throughout the entire execution of
   // the computation graph
   // Calling both `reset` and `seed` methods is fundamental here to ensure reproducibility: without them the same
   // generator could be seeded by a different entry (depending on which is the first entry ID seen by a thread) or
   // could be at a different step of the sequence (depending how many entries this particular thread is processing).
   // Alternatively, if both the generator and the distribution objects were recreated from scratch at every function
   // call (i.e. by removing the global std::vector stores), then the next two method calls would not be necessary, at
   // the cost of a possible performance degradation.
   gaussians[slot].reset();
   generators[slot].seed(entry);
   return gaussians[slot](generators[slot]);
}
```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.hxx
```diff
@@ -0,0 +1,53 @@
...
+
+#endif // ROOT_TUTORIALS_ANALYSIS_DATAFRAME_DF041
```
ferdymercury:
```suggestion
#endif // ROOT_TUTORIALS_ANALYSIS_DATAFRAME_DF041

```

tutorials/analysis/dataframe/df041_ThreadSafeRNG.hxx
```diff
@@ -0,0 +1,53 @@
...
+inline std::random_device globalRandomDevice{};
+inline std::mt19937 globalGenerator(globalRandomDevice());
+inline std::normal_distribution<double> globalGaus(0., 1.);
+
+double GetNormallyDistributedNumberFromGlobalGenerator()
+{
+   return globalGaus(globalGenerator);
+}
+
+// One generator per slot — initialized once before the event loop
```
ferdymercury, line 17:
```suggestion
// One generator per slot — initialized once before the event loop
// An alternative to these global vectors could be to have thread_local variables inside a class or function
```

ferdymercury (changes requested):
Thanks a lot!! A few suggestions from my side

tutorials/analysis/dataframe/df041_ThreadSafeRNG.hxx
vepadulano: I don't want to write implementation details or alternatives that are not implemented as part of the tutorial description. This example is already complicated enough as it is.

tutorials/analysis/dataframe/df041_ThreadSafeRNG.hxx
vepadulano: Nice catch, thanks!

tutorials/analysis/dataframe/index.md
vepadulano: Nice catch, thanks!

tutorials/analysis/dataframe/df041_ThreadSafeRNG.hxx
vepadulano: On second thoughts, we are already doing this in other places, so once more won't hurt I guess.

ferdymercury (approved):
Thanks a lot!

vepadulano:
I wanted to record in writing that I have attempted reproducing the Windows failures multiple times without success and it's been hard to move this PR further. I think the tutorial is already good as it is and shouldn't be modified but until the CI is green unfortunately this can't be merged. Tangentially, this PR would also fix this issue https://github.com/root-project/root/issues/20839, I will link this there.

vepadulano (approved):
Thank you so much for this very interesting addition to the RDataFrame tutorials! All builds are passing now so we can go ahead.

PR #20908, opened 2026-01-16
[ntuple] Throw exception if appended file is not writable

silverweed:
instead of crashing.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/test/ntuple_basics.cxx
+++ b/tree/ntuple/test/ntuple_basics.cxx
@@ -217,6 +217,26 @@ TEST(RNTuple, WriteReadSubdir)
    EXPECT_FLOAT_EQ(137.0, *reader->GetModel().GetDefaultEntry().GetPtr<float>("pt"));
 }
 
+TEST(RNTuple, AppendReadOnly)
+{
+   FileRaii fileGuard("test_ntuple_append_readonly.root");
+
+   {
+      auto model = RNTupleModel::Create();
+      model->MakeField<int>("a");
+      auto writer = RNTupleWriter::Recreate(std::move(model), "A", fileGuard.GetPath());
+   }
+
+   auto model = RNTupleModel::Create();
+   auto file = std::unique_ptr<TFile>(TFile::Open(fileGuard.GetPath().c_str(), "READ"));
+   try {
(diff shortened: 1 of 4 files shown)


hahnjo (changes requested):
Can you please also fix this for `RNTupleParallelWriter`?

tree/ntuple/test/ntuple_basics.cxx
```diff
@@ -217,6 +217,30 @@ TEST(RNTuple, WriteReadSubdir)
...
+
+   {
+      auto model = RNTupleModel::Create();
+      auto pA = model->MakeField<int>("a");
+      auto writer = RNTupleWriter::Recreate(std::move(model), "A", fileGuard.GetPath());
+      for (int i = 0; i < 100; ++i) {
+         *pA = i;
+         writer->Fill();
+      }
+   }
```
jblomer, line 228: We don't need that, it should be enough to create an empty file.

tree/ntuple/test/ntuple_parallel_writer.cxx
```diff
@@ -80,6 +80,30 @@ TEST(RNTupleParallelWriter, Basics)
...
+
+   {
+      auto model = RNTupleModel::Create();
+      auto pA = model->MakeField<int>("a");
+      auto writer = RNTupleWriter::Recreate(std::move(model), "A", fileGuard.GetPath());
+      for (int i = 0; i < 100; ++i) {
+         *pA = i;
+         writer->Fill();
+      }
+   }
```
jblomer: See above

jblomer (approved):
LGTM

tree/ntuple/test/ntuple_parallel_writer.cxx
```diff
@@ -80,6 +80,26 @@ TEST(RNTupleParallelWriter, Basics)
    }
 }
 
+TEST(RNTupleParallelWriter, AppendReadOnly)
```
hahnjo: Maybe move down after `ForbidNonRootTFiles`? Was scanning the context for similar tests (which I was sure we had) and couldn't easily find it...

hahnjo (approved):

PR #20909, opened 2026-01-16
[ntuple][rfile] Add support for RFile in RNTupleWriter

silverweed:
# This Pull request:
allows RMiniFileWriter to handle RFile and exposes an experimental way to create an RNTupleWriter using RFile.
Currently the semantics are the same as the `TDirectory` overload: you pass in a reference to `RFile` and are responsible to keep the file alive until the RNTupleWriter is done.

Main question to answer: is this the semantics we want? It seems to me that it's the most straightforward API in C++, but it requires some extra work in python to avoid use-after-free cases like:
```
def CreateWriter():
  file = ROOT.RFile.Open(...)
  writer = ROOT.Experimental.RNTupleWriter_Append(..., file)
  return writer  # <- file may be GC'd from here on
```

This should be solveable by pythonizing `Append` to hold on to the file's proxy on the python side, but there might be additional complications I'm not considering. I'm open for discussion.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/inc/ROOT/RNTupleWriter.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleWriter.hxx
@@ -39,7 +39,19 @@ class RNTupleWriteOptions;
 
 namespace Experimental {
 class RFile;
-}
+
+/// Creates an RNTupleWriter that writes into the given `file`, appending to it. The RNTuple is written under the
+/// path `ntuplePath`.
+/// `ntuplePath` may have the form `"path/to/ntuple"`, in which case the ntuple's name will be `"ntuple"` and it will
+/// be stored under the given `ntuplePath` in the RFile.
+/// Throws an exception if the model is null.
+/// NOTE: this is a temporary, experimental API that will be replaced by an overload of RNTupleWriter::Append in the
+/// future.
+std::unique_ptr<RNTupleWriter>
+RNTupleWriter_Append(std::unique_ptr<ROOT::RNTupleModel> model, std::string_view ntuplePath,
+                     ROOT::Experimental::RFile &file,
+                     const ROOT::RNTupleWriteOptions &options = ROOT::RNTupleWriteOptions());
+} // namespace Experimental
 
 namespace Internal {
 // Non-public factory method for an RNTuple writer that uses an already constructed page sink
@@ -106,6 +118,9 @@ class RNTupleWriter {
    friend ROOT::RNTupleModel::RUpdater;
    friend std::unique_ptr<RNTupleWriter>
       Internal::CreateRNTupleWriter(std::unique_ptr<ROOT::RNTupleModel>, std::unique_ptr<Internal::RPageSink>);
+   friend std::unique_ptr<RNTupleWriter>
+   Experimental::RNTupleWriter_Append(std::unique_ptr<ROOT::RNTupleModel> model, std::string_view ntuplePath,
+                                      ROOT::Experimental::RFile &file, const ROOT::RNTupleWriteOptions &options);
 
 private:
    RNTupleFillContext fFillContext;
@@ -155,12 +170,6 @@ public:
    /// \see Recreate()
    static std::unique_ptr<RNTupleWriter> Append(std::unique_ptr<ROOT::RNTupleModel> model, std::string_view ntupleName,
                                                 TDirectory &fileOrDirectory,
-                                                const ROOT::RNTupleWriteOptions &options = ROOT::RNTupleWriteOptions());
-   /// Throws an exception if the model is null.
-   /// `ntuplePath` may have the form `"path/to/ntuple"`, in which case the ntuple's name will be `"ntuple"` and it will
-   /// be stored under the given `ntuplePath` in the RFile.
-   static std::unique_ptr<RNTupleWriter> Append(std::unique_ptr<ROOT::RNTupleModel> model, std::string_view ntuplePath,
-                                                ROOT::Experimental::RFile &file,
                                                 const ROOT::RNTupleWriteOptions &options = ROOT::RNTupleWriteOptions());
 
    RNTupleWriter(const RNTupleWriter &) = delete;
--- a/tree/ntuple/src/RNTupleWriter.cxx
+++ b/tree/ntuple/src/RNTupleWriter.cxx
@@ -119,15 +119,6 @@ ROOT::RNTupleWriter::Append(std::unique_ptr<ROOT::RNTupleModel> model, std::stri
    return Create(std::move(model), std::move(sink), options);
 }
 
-std::unique_ptr<ROOT::RNTupleWriter>
-ROOT::RNTupleWriter::Append(std::unique_ptr<ROOT::RNTupleModel> model, std::string_view ntupleName,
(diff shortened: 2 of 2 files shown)


tree/ntuple/src/RMiniFile.cxx
```diff
@@ -1177,6 +1178,47 @@ std::uint64_t ROOT::Internal::RNTupleFileWriter::RFileProper::ReserveBlobKey(siz
...
+{
+   auto *file = ROOT::Experimental::Internal::GetRFileTFile(*fFile);
+   file->Seek(offset);
+   bool rv = file->WriteBuffer((char *)(buffer), nbytes);
+   if (rv)
+      throw RException(R__FAIL("WriteBuffer failed."));
+}
+
+std::uint64_t ROOT::Internal::RNTupleFileWriter::RFileRFile::ReserveBlobKey(size_t nbytes, size_t len,
+                                                                            unsigned char keyBuffer[kBlobKeyLen])
```
hahnjo, line 1216: This looks like an (almost) identical cop&paste from `RFileProper`. Can this be refactored to share the implementation?

tree/ntuple/src/RMiniFile.cxx
```diff
@@ -1290,6 +1344,18 @@ void ROOT::Internal::RNTupleFileWriter::Commit(int compression)
...
+      fileRFile->fFile->Put(fileRFile->fDir + fNTupleName, fNTupleAnchor);
+
+      // Make sure the streamer info records used in the RNTuple are written to the file
+      TBufferFile buf(TBuffer::kWrite);
+      buf.SetParent(ROOT::Experimental::Internal::GetRFileTFile(*fileRFile->fFile));
+      for (auto [_, info] : fStreamerInfoMap)
+         buf.TagStreamerInfo(info);
+
+      fileRFile->fFile->Flush();
+      return;
```
hahnjo, line 1357: This also looks almost identical to the code path above, maybe worth to deduplicate.

silverweed: I'm not finding good ways to deduplicate it effectively (I can reduce the number of lines but at the price of obfuscating the code quite a bit); if you have suggestions feel free to propose

tree/ntuple/src/RMiniFile.cxx
```diff
@@ -1138,19 +1139,12 @@ std::uint64_t ROOT::Internal::RNTupleFileWriter::RFileSimple::ReserveBlobKey(std
...
-   fDirectory->GetFile()->Seek(offset);
-   bool rv = fDirectory->GetFile()->WriteBuffer((char *)(buffer), nbytes);
-   if (rv)
-      throw RException(R__FAIL("WriteBuffer failed."));
-}
-
-std::uint64_t ROOT::Internal::RNTupleFileWriter::RFileProper::ReserveBlobKey(size_t nbytes, size_t len,
-                                                                             unsigned char keyBuffer[kBlobKeyLen])
+template <typename T>
+std::uint64_t ROOT::Internal::RNTupleFileWriter::ReserveBlobKey(T &caller, TFile &file, std::size_t nbytes,
```
hahnjo, line 1160: Isn't `caller` always a `TFile *`? Also not sure if we need to expose this additional function via the `RNTupleFileWriter`, it should not need members...

tree/ntuple/src/RMiniFile.cxx
hahnjo: Ok, I was mostly looking at the `TBufferFIle` part and `TagStreamerInfo`. That can probably go into a local lambda that takes a `TFile *` argument.

tree/ntuple/src/RMiniFile.cxx
silverweed: `caller` is either `RFileProper` or `RFileRFile`; this could be a free function but since it calls `RNTupleFileWriter::PrepareBlobKey` and uses `RNTupleFileWriter::kBlobKeyLen` I made it a private static function for consistency (otherwise we have a mix of private static functions and free functions - alternatively I could maybe extract `PrepareBlobKey` into a free function as well...)

hahnjo: Ah sorry, I misread the code, makes perfectly sense now.

io/io/test/rfile.cxx
```diff
@@ -694,3 +697,39 @@ TEST(RFile, GetKeyInfo)
       EXPECT_EQ(key->GetCycle(), 1);
    }
 }
+
+TEST(RFile, RNTuple)
```
jblomer, line 701: I think that can be slightly simpler. We could test empty RNTuples or ones with a single field and entry.

tree/ntuple/inc/ROOT/RMiniFile.hxx
```diff
@@ -126,6 +132,19 @@ private:
       operator bool() const { return fDirectory; }
    };
 
+   struct RFileRFile {
+      /// A sub directory in fFile or nullptr if the data is stored in the root directory of the file
```
jblomer: Is this comment still accurate?

tree/ntuple/inc/ROOT/RMiniFile.hxx
```diff
@@ -177,9 +196,16 @@ private:
...
 
+   template <typename T>
+   static std::uint64_t
+   ReserveBlobKey(T &caller, TFile &file, std::size_t nbytes, std::size_t len, unsigned char keyBuffer[kBlobKeyLen]);
+
    /// RFileSimple: for simple use cases, survives without libRIO dependency
    /// RFileProper: for updating existing files and for storing more than just an RNTuple in the file
-   std::variant<RFileSimple, RFileProper> fFile;
+   /// RFileRFile: like RFileProper but using RFile instead of TFile.
+   using FileType = std::variant<RFileSimple, RFileProper, RFileRFile>;
```
jblomer:
```suggestion
   using FileType_t = std::variant<RFileSimple, RFileProper, RFileRFile>;
```

silverweed:
Added a python test (and a small pythonization of the RFile-based Append)

silverweed:
Since the python test and pythonization are blocked by https://github.com/root-project/root/pull/21030, for this PR I'll only add the C++ part and will do the python side on a separate one.

tree/ntuple/inc/ROOT/RMiniFile.hxx
```diff
@@ -128,6 +134,18 @@ private:
       operator bool() const { return fDirectory; }
    };
 
+   struct RFileRFile {
```
pcanal, line 137: This class would benefit from a short introduction to its raison d'être as the name itself is not self explanatory.

silverweed:
> LGTM.
> 
> I think we should have a follow-up PR to rename the internal file structs in the mini file writer, e.g. `RImplSimple`, `RImplTFile`, `RImplRFile`... or something better. It just seems that currently the code overuses the word `File`.

#21100

tree/ntuple/inc/ROOT/RMiniFile.hxx
silverweed: See #21100

PR #20917, opened 2026-01-16
[cmake] Unify Ninja builds

bellenot:
- Don't exclude MSVC for the `Ninja` `RESOURCE_LOCK`
- Fix a warning on Windows about unknown `-fPIC` compiler flag

### DIFF
--- a/bindings/pyroot/cppyy/cppyy-backend/CMakeLists.txt
+++ b/bindings/pyroot/cppyy/cppyy-backend/CMakeLists.txt
@@ -6,6 +6,4 @@
 
 add_library(cppyy_backend STATIC clingwrapper/src/clingwrapper.cxx)
 target_link_libraries(cppyy_backend Core)
-if(NOT MSVC)
-  target_compile_options(cppyy_backend PRIVATE -fPIC)
-endif()
+target_compile_options(cppyy_backend PRIVATE -fPIC)


bindings/pyroot/cppyy/cppyy-backend/CMakeLists.txt
```diff
@@ -6,4 +6,6 @@
 
 add_library(cppyy_backend STATIC clingwrapper/src/clingwrapper.cxx)
 target_link_libraries(cppyy_backend Core)
-target_compile_options(cppyy_backend PRIVATE -fPIC)
+if(NOT MSVC)
+  target_compile_options(cppyy_backend PRIVATE -fPIC)
+endif()
```
hahnjo: This one is in principle there since commit 3f0145a1e17b66d16e1ab4cbec4cb22c24a6558b last July, and looks in principle unrelated to Ninja - maybe you can split it into a separate commit? And why do we only see this now?

hahnjo (approved):
The Ninja changes LGTM

bindings/pyroot/cppyy/cppyy-backend/CMakeLists.txt
bellenot: Well, right, it's there since a while, but it's harmless, and I don't thinks it deserves its own PR

hahnjo: That's fine, but I think a separate commit (in the same PR) wouldn't hurt, even if just for future understanding.

bellenot: > That's fine, but I think a separate commit (in the same PR) wouldn't hurt, even if just for future understanding.

See https://github.com/root-project/root/pull/20951

PR #20919, opened 2026-01-16
[roottest] Move hist ownership test from diffs to programmatic test.

hageboeck:
Diffs of the outputs are fragile, so the ownership of histograms is now tested programmatically.

### DIFF
--- a/roottest/root/hist/misc/CMakeLists.txt
+++ b/roottest/root/hist/misc/CMakeLists.txt
@@ -1,6 +1,4 @@
-ROOTTEST_ADD_TEST(runownership
-                  MACRO runownership.C+
-                  OUTREF ownership.ref)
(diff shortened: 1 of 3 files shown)


guitargeek (approved):
Thanks! That's better indeed

hageboeck:
The recurring mac14 failure was related to the cppyy leakcheck test, which has been disabled now.

pcanal:
> Diffs of the outputs are fragile

Is this a general statement or did this test have unexpected behavior recently?

hageboeck:
> > Diffs of the outputs are fragile
> 
> Is this a general statement or did this test have unexpected behavior recently?

I broke it when testing some ownership work, and I find it harder to read reference files + correlated diffs to the code than having an `if (xxx) fail` directly in the code.
So it's a general statement, but the change to make the test logic more local was triggered by failures I was investigating.

PR #20920, opened 2026-01-16
[hist] Make two tests independent of gDirectory

hageboeck:
In view of ROOT 7, these test can be run independent of what `gDirectory` is.

### DIFF
--- a/hist/hist/test/test_TH1_SaveAs.cxx
+++ b/hist/hist/test/test_TH1_SaveAs.cxx
@@ -162,6 +162,9 @@ struct TestSaveAs {
          if (line != "{" && line != "}" && (line.rfind("//", 0) == 0 || line.length() < 6)) {
             continue;
          }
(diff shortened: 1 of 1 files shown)


hist/hist/test/test_TH1_SaveAs.cxx
```diff
@@ -162,6 +162,8 @@ struct TestSaveAs {
          if (line != "{" && line != "}" && (line.rfind("//", 0) == 0 || line.length() < 6)) {
             continue;
          }
+         if (line.find("SetDirectory") != std::string::npos)
```
vepadulano, line 166: Maybe just a small explanation here
```suggestion
         // Skip lines that detach the histogram from the gDirectory
         // making them work irrespectively of whether implicit
         // memory management is turned on
         if (line.find("SetDirectory") != std::string::npos)
```

vepadulano (approved):

PR #20921, opened 2026-01-16
[PyROOT] Remove a test that worked just by chance.

hageboeck:
The test broke when implicit object ownership was switched off in ROOT. However, this test can also break in other circumstances, because
   del histogram
generally does not destroy the C++ object immediately. It used to work because only a single reference to the object was alive (which changed when trying to prepare the test for implicit object ownership off), but one cannot base a test on undocumented details of a Python implementation.

For this reason, it does not make sense to test if gDirectory (C++) contains an entry for the object that was just "del"-ed.

### DIFF
--- a/roottest/python/memory/PyROOT_memorytests.py
+++ b/roottest/python/memory/PyROOT_memorytests.py
@@ -60,21 +60,6 @@ class Memory1TestCase( MyTestCase ):
       del b, c
       self.assertEqual( MemTester.counter, 0 )
 
-   def test2ObjectDestructionCallback( self ):
-      """Test ROOT notification on object destruction"""
-
-    # create ROOT traced object
-      a = TH1F( 'memtest_th1f', 'title', 100, -1., 1. )
-
-    # locate it
-      self.assertTrue( a is gROOT.FindObject( 'memtest_th1f' ) )
-
-    # destroy it
-      del a
-
-    # should no longer be accessible
-      self.assertTrue( not gROOT.FindObject( 'memtest_th1f' ) )
-
    def test3ObjectCallHeuristics( self ):
       """Test memory mgmt heuristics for object calls"""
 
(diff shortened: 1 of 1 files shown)


guitargeek (approved):
Makes sense, thank you very much!

vepadulano (approved):

pcanal:
> For this reason, it does not make sense to test if gDirectory (C++) contains an entry for the object that was just "del"-ed.

If I understand correctly this also means that with the implicit ownership, the following does not work as expected (minus python typos)
```
file = TFile.Open(filename, "RECREATE")
a = TH1F( 'memtest_th1f', 'title', 100, -1., 1. )
del a
file.Write()
```
might or might not actually write the histogram.  
Similarly, even with explict registration (obviously in real use case, the line would not be literally consecutive :) ).
```
file = TFile.Open(filename, "RECREATE")
a = TH1F( 'memtest_th1f', 'title', 100, -1., 1. )
a.SetDirectory(a)
del a
file.Write()
```
might or might not actually write the histogram.

hageboeck:
> > For this reason, it does not make sense to test if gDirectory (C++) contains an entry for the object that was just "del"-ed.
> 
> If I understand correctly this also means that with the implicit ownership, the following does not work as expected (minus python typos)
> 
> ```
> file = TFile.Open(filename, "RECREATE")
> a = TH1F( 'memtest_th1f', 'title', 100, -1., 1. )
> del a
> file.Write()
> ```
> 
> might or might not actually write the histogram. Similarly, even with explict registration (obviously in real use case, the line would not be literally consecutive :) ).
> 
> ```
> file = TFile.Open(filename, "RECREATE")
> a = TH1F( 'memtest_th1f', 'title', 100, -1., 1. )
> a.SetDirectory(a)
> del a
> file.Write()
> ```
> 
> might or might not actually write the histogram.

This is correct. It is unclear when the garbage collector runs, so the histograms might or might not be written.
For us, these tests mostly worked due to a special case that instances with only one ref are deleted immediately instead of waiting for the garbage collector in CPython. This seems to be a feature of CPython only, though, so we shouldn't rely on it.

PR #20922, opened 2026-01-16
[HF] Prepare HistFactory for ROOT without implicit ownership

hageboeck:
- Go from implicit to explicit writing in `makeExample.C` tutorial
- Also explicitly write histograms in testHistFactoryPlotting.cxx

By making the writes explicit, the tests become independent of gDirectory and whether or not histograms register themselves.

### DIFF
--- a/roofit/histfactory/test/testHistFactoryPlotting.cxx
+++ b/roofit/histfactory/test/testHistFactoryPlotting.cxx
@@ -57,7 +57,10 @@ void createToyHistos1D()
       histoData->Fill(x_sig);
    }
 
-   histoFile->Write();
+   for (auto histo : {histoBG, histoSig, histoData}) {
+      histoFile->WriteTObject(histo);
+      delete histo;
+   }
 }
 
 /// Test that plotting HistFactory components works correctly. Covers the


tutorials/roofit/histfactory/makeExample.C
```diff
@@ -5,32 +5,38 @@ void makeDataDriven()
   TFile* file = new TFile("dataDriven.root", "RECREATE");
 
   TH1F* FlatHist = new TH1F("FlatHist","FlatHist", 2,0,2);
+  FlatHist->SetDirectory(file);
```
guitargeek: I don't get it yet :slightly_smiling_face: If you later explicitly write the histograms with `file->WriteTObject(histo)`, why do you need to do `SetDirectory(file)` before?

hageboeck: You indeed don't need it to write, but you do need it to destroy the histogram. Alternatively, you can of course take the ownership "manually".

hageboeck: Let me do a solution that doesn't need `SetDirectory`

PR #20923, opened 2026-01-16
[main] optparse: add support for prefix flags

silverweed:
TL;DR: adds support for treating `-Dfoo` as `-D foo`. See header documentation for details.

EDIT: superseded by https://github.com/root-project/root/pull/20947

## Checklist:

- [x] tested changes locally

### DIFF
--- a/main/src/optparse.hxx
+++ b/main/src/optparse.hxx
@@ -37,11 +37,19 @@
 /// ~~~
 ///
 /// ## Additional Notes
-/// If all the short flags you pass (those starting with a single `-`) are 1 character long, the parser will accept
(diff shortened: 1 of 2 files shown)


devajithvs:
I feel like this approach is a bit too limiting. If you have a short-flag starting with a certain character, you can't define any other option that starts with the same character. 

In LLVM, they have a specific argument you can pass when creating a prefix option, something like: 

```
gOptPPDefines("D", cl::Prefix, ...);
```

Could we support something similar? For example:
```
opts.AddFlag({"-D"}, ROOT::RCmdLineOpts::EFlagType::kPrefixArg, ...);
```

silverweed:
@devajithvs see https://github.com/root-project/root/pull/20947

PR #20924, opened 2026-01-16
[core] Use shared library dir as anchor in `GetIncludeDir()`

guitargeek:
Using the shared library directory as the anchor to resolve the resource directories like `GetIncludeDir()` is more direct and robust than using `GetRootSys()`, which loses its meaning in some installations (e.g. what should it be when ROOT is installed by the package manager? Just `/`?).

Most importantly, this change lifts some more constraint from how the install tree can be reorganized after the fact. One can for example move the include and library directory around as wanted, as long as their relative paths don't change. This fixes the Python wheels, which were broken by commit a5b1ed9080aca63347d3156540864f774ac49d4d.

As of commit a5b1ed9080aca63347d3156540864f774ac49d4d, ROOT became smarter about finding the include path in the install tree, inferring the correct relative path from $ROOTSYS and the CMAKE_INSTALL_INCLUDEDIR variable at build time. Before, that only happened for gnuinstall=ON, but now it always happens.

However, for the Python wheel, we are breaking the assumptions that ROOT makes by moving around directories in the install tree. This commit fixes that situation.

I tested this with three build configurations:
  * no `CMAKE_INSTALL_INCLUDEDIR` set
  * `CMAKE_INSTALL_INCLUDEDIR` relative path
  * `CMAKE_INSTALL_INCLUDEDIR` absolute path

In all cases, `gROOT->GetIncludeDir()` returns the correct path, both via the `root` executable in the build and install tree.

### DIFF
--- a/core/base/CMakeLists.txt
+++ b/core/base/CMakeLists.txt
@@ -233,15 +233,19 @@ set(full_core_filename $<TARGET_FILE_NAME:Core>)
 # Absolue CMAKE_INSTALL_<dir> paths are discouraged in CMake, but some
 # packagers use them anyway. So we support it.
 if(IS_ABSOLUTE ${CMAKE_INSTALL_INCLUDEDIR})
-   set(install_includedir_is_absolute 1)
+  set(install_lib_to_include "${CMAKE_INSTALL_INCLUDEDIR}")
 else()
-   set(install_includedir_is_absolute 0)
+  if(IS_ABSOLUTE ${CMAKE_INSTALL_LIBDIR})
+    set(libdir "${CMAKE_INSTALL_LIBDIR}")
+  else()
+    set(libdir "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
+  endif()
+  file(RELATIVE_PATH install_lib_to_include "${libdir}" "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_INCLUDEDIR}")
+  unset(libdir)
 endif()
-file(TO_NATIVE_PATH "${CMAKE_INSTALL_INCLUDEDIR}" install_includedir_native)
 
 target_compile_options(Core PRIVATE -DLIB_CORE_NAME=${full_core_filename}
-                                    -DINSTALL_INCLUDEDIR=${install_includedir_native}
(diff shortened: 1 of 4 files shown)


guitargeek:
Hi @pcanal, once the tests pass, can you maybe review this PR since you are already familiar with my previous PR https://github.com/root-project/root/pull/20828? This is a followup that is important to fix the Python wheels. Thanks!

guitargeek:
Also, here are my next plans in this `ROOTSYS` business:
  * Also implement `GetRootSys()` by using `GetSharedLibDir()` as an anchor
  * At that point we don't need `ROOTSYS` anymore
  * https://github.com/root-project/root/pull/20006 can proceed

pcanal:
> At that point we don't need ROOTSYS anymore

However, we should warn if `ROOTSYS` *is set* but different from the calculated one ...

pcanal:
> However, for the Python wheel, we are breaking the assumptions that ROOT makes by moving around directories in the install tree. 

What kind of movement is done?  For Wheels is `CMAKE_INSTALL_INCLUDEDIR` absolute or relative?

guitargeek:
> > At that point we don't need ROOTSYS anymore
> 
> However, we should warn if `ROOTSYS` _is set_ but different from the calculated one ...

Good idea :+1: I'll keep that in mind when implementing it.


> > However, for the Python wheel, we are breaking the assumptions that ROOT makes by moving around directories in the install tree.
> 
> What kind of movement is done? For Wheels is `CMAKE_INSTALL_INCLUDEDIR` absolute or relative?

It is [set to be relative](https://github.com/root-project/root/blob/master/setup.py#L94). The moving is [done here in the setup.py](https://github.com/root-project/root/blob/master/setup.py#L133).

The reason why this is done is because for the Python wheels, we want only a subset of the artifacts in the install prefix. Most of the subset gets installed to a `mock_site_packages` subdirectory in the install tree, and then moved out to the actual Python package directory. That was however a hacky solution to make it work initially. On the long term, we should make sure that ROOT can be configured in such a way that the install tree is exactly what we need for the Python package, and then we just move it to the Python package directory (FYI @vepadulano)

That being said, this PR is not only to make the hacky setup for the Python wheels for again, but mainly to make the `ROOTSYS`-less ROOT work. So the design of the `setup.py` for the Python wheels should not be a criterion when reviewing this PR.

Issue #17697, opened 2025-02-11
pyroot can trigger clang assertion failure on enum with value too big to fit in int64_t


scott-snyder:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

In root 6.34.02, trying to use via pyroot a C++ class containing an enum with a value that is too big to fit into a (signed) int64_t will trip an assertion in the dbg build:

```
python: /build/jenkins/workspace/lcg_release_pipeline/build/projects/ROOT-6.34.02/src/ROOT/6.34.02/interpreter/llvm-project/llvm/include/llvm/ADT/APSInt.h:99: int64_t llvm::APSInt::getExtValue() const: Assertion `isRepresentableByInt64() && "Too many bits for int64_t"' failed.
```

See the reproducer below.

Here is a partial stack trace:

```
#4  0x00007ffff76373c6 in __assert_fail () from /lib64/libc.so.6
#5  0x00007fffedeaa1fe in llvm::APSInt::getExtValue (this=0x7fffffff9a20)
    at /build/jenkins/workspace/lcg_release_pipeline/build/projects/ROOT-6.34.02/src/ROOT/6.34.02/interpreter/llvm-project/llvm/include/llvm/ADT/APSInt.h:99
#6  0x00007fffedea92e7 in TClingDataMemberInfo::Offset (this=0x6a278a0)
    at /build/jenkins/workspace/lcg_release_pipeline/build/projects/ROOT-6.34.02/src/ROOT/6.34.02/core/metacling/src/TClingDataMemberInfo.cxx:408
#7  0x00007fffedcf0d28 in TCling::DataMemberInfo_Offset (this=0x5b0c20, 
    dminfo=0x6a278a0)
    at /build/jenkins/workspace/lcg_release_pipeline/build/projects/ROOT-6.34.02/src/ROOT/6.34.02/core/metacling/src/TCling.cxx:8653
#8  0x00007ffff6806e53 in TDataMember::GetOffsetCint (this=0x6ac11a0)
    at /build/jenkins/workspace/lcg_release_pipeline/build/projects/ROOT-6.34.02/src/ROOT/6.34.02/core/meta/src/TDataMember.cxx:499
#9  0x00007ffff6cd84a6 in Cppyy::GetDatamemberOffset (scope=17, idata=0)
[...]

PR #20925, opened 2026-01-16
[core] do not assert when casting huge enum uint64_t into int64_t

ferdymercury:
# This Pull request:

## Changes or fixes:

even if it becomes negative due to the overflow.

Fixes https://github.com/root-project/root/issues/17697

Tested that the Python reproducer in https://github.com/root-project/root/issues/17697 no longer crashes

```
code = """
struct Foo
{
  enum E {
    BIG = ((unsigned long long)1)<<63
  };
};
"""

import ROOT
ROOT.gInterpreter.Declare (code)

f = ROOT.Foo()
f.BIG
(Foo::E::Foo::E::BIG) : (internal_enum_type_t) -9223372036854775808
```

## Checklist:

- [x] tested changes locally

### DIFF
--- a/core/metacling/test/MetaClingTests.py
+++ b/core/metacling/test/MetaClingTests.py
@@ -32,3 +32,18 @@ struct GH_14425_Required {
         self.assertEqual(ROOT.GH_14425_Default().fMember, 1)
         self.assertEqual(ROOT.GH_14425_Default(ROOT.GH_14425(2)).fMember, 2)
         self.assertEqual(ROOT.GH_14425_Required(ROOT.GH_14425(3)).fMember, 3)
+
+class TClingTest(unittest.TestCase):
+    """Tests related to TCling and federation of classes usage from Python"""
+
+    def test_GH_20925(self):
+      """Do not assert when casting huge enum uint64_t into int64_t"""
+   
+      ROOT.gInterpreter.Declare(r"""
+struct Foo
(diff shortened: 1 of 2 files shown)


dpiparo:
Do we maybe need a unit test for this?

dpiparo:
I added a test and restarted the builds. Let's see how builds go and then we can merge.

dpiparo (approved):

ferdymercury:
test seems to fail on Windows. Maybe `internal_enum_type_t` has a different size on Windows?

core/metacling/test/MetaClingTests.py
```diff
@@ -32,3 +32,18 @@ def test_GH_14425(self):
...
+class TClingTest(unittest.TestCase):
+    """Tests related to TCling and federation of classes usage from Python"""
+
+    def test_GH_20925(self):
+      """Do not assert when casting huge enum uint64_t into int64_t"""
+   
+      ROOT.gInterpreter.Declare(r"""
+struct Foo
+{
+enum E { BIG = ((unsigned long long)1)<<63};
```
bellenot:
```suggestion
enum E: unsigned long long { BIG = ((unsigned long long)1)<<63};
```
To fix the error on Windows

dpiparo:
2 commits were fused to ease future bisects.

PR #20927, opened 2026-01-17
[ci] Add Ubuntu 26.04 and Fedora 44

guitargeek:
Ubuntu 26.04 and Fedora 44 will be released in April and May just before ROOT 6.40, so it's good to already test these distributions with ROOT `master`.

Even though these are still only previews, the added builds are not marked as "special", because by the time we'll have the next release, these distribution will already be out. So no need to make them "special", which would indicate that the artifacts don't have to be provided as binary releases.

In turn, remove Ubuntu 25.10 and Fedora 42, which will be end-of-life around the time the new ROOT release 6.40 comes out.


.github/workflows/root-ci-config/buildconfig/fedora44.txt
```diff
@@ -1,6 +1,8 @@
+CMAKE_CXX_STANDARD=23
 builtin_zlib=ON
 builtin_zstd=ON
 experimental_adaptivecpp=ON
-experimental_genvectorx=ON
```
dpiparo, line 4: We need to continue to test genvectorx. Is it possible?

guitargeek: I'm not sure. I just copy-pasted the config from Fedora 43 with no changes. I don't know why only Fedora 42 had `experimental_genvectorx=ON` and not Fedora 43. But for an initial addition of Fedora 44, I didn't want to risk failing builds by experimenting too much. We can change the config later once we have a working baseline :+1:

It's still tested on `ubuntu2404` though, so we don't lose test coverage for now.

PR #20906, opened 2026-01-15
[cppyy] Review xfail'ed tests

guitargeek:
All xfail'ed cppyy tests have been reviewed, and the ones that are
actually failing are now marked as `xfail(strict=True)`, so the test
fails if it unexpectedly passes.

This gives us a very useful baseline for what works now in ROOT and what
doesn't, and we'll learn when our developments like the cppyy upgrade on
the CppInterOp-based version will fix some tests.

There are some tests remaining that can't be run because they crash, and
therefore the `strict` mode doesn't make sense there. But the number of
these tests is not high, and now we always give a reason for why they
crash:

```txt
git grep xfail | grep -v "strict"
```
```txt
test_concurrent.py:    @mark.xfail(run=False, reason="Crashes because TClingCallFunc generates wrong code")
test_concurrent.py:    @mark.xfail(run=False, reason="Crashes because the interpreter emits too many warnings")
test_concurrent.py:    @mark.xfail(run=False, reason="segmentation violation")
test_cpp11features.py:    @mark.xfail(run=False, reason = "Crashes")
test_datatypes.py:    @mark.xfail(run=False, reason="segmentation violation")
test_datatypes.py:    @mark.xfail(run=False, reason="error code: Subprocess aborted")
test_doc_features.py:    @mark.xfail(run=False, condition=WINDOWS_BITS == 64, reason = "Crashes on Windows 64 bit")
test_fragile.py:    @mark.xfail(run=False, condition=has_asserts(),
test_fragile.py:    @mark.xfail(run=False, reason="Fatal Python error: Aborted")
test_fragile.py:    @mark.xfail(run=False, reason="Fatal Python error: Aborted")
test_fragile.py:    @mark.xfail(run=False, condition=is_modules_off(), reason="Crashes on build with modules off: Fatal Python error: Segmentation fault")
test_lowlevel.py:    @mark.xfail(run=False, condition=IS_WINDOWS, reason="Windows fatal exception: access violation")
test_regression.py:    @mark.xfail(run=WINDOWS_BITS != 64, condition=IS_MAC_ARM | WINDOWS_BITS == 64, reason = "Crashes on Windows 64 bit and fails macOS ARM with" \
test_stltypes.py:    @mark.xfail(run=False, reason="Fatal Python error: Segmentation fault")
test_stltypes.py:    @mark.xfail(run=False, condition=WINDOWS_BITS == 64, reason="Crashes on Windows 64 bit")
[...]

PR #20930, opened 2026-01-17
[ci] Speedup mac14 builds by not running tutorials

dpiparo:
This will compensate for the slowness of the runners.


guitargeek:
Interesting idea! I have two questions:

* Can this PR maybe wait for https://github.com/root-project/root/pull/20906? I'm doing a large-scale review of which cppyy tests are actually passing on which platforms, so it would be good to keep the full mac14 configuration until this is done. I *could* revert your proposed change in that PR for testing, but then I still would not have a complete `mac14` reference in the nightlies and might not be able to tell apart failures in my PR from potential regressions in `master`.
* What's the strategy for the release branches, including the upcoming 6.40 branch? We enable the full build there?

PR #20932, opened 2026-01-18
[math] Turn regression test for ROOT-7739 into `static_assert`

guitargeek:
The `roottest/root/math/matrix/ops.C` unit test can be easily converted to a `static_assert` when compiling ROOT itself.


dpiparo:
chapeau!

PR #20935, opened 2026-01-18
[cppyy] Integrate upstream cppyy developments from last months

guitargeek:
This is done to make sure that the cppyy versions don't diverge too much.

Also, reverts a small `cppyy` documentation change from 60cb356cbab, which was not significant enough to merit a patch. The `cppyy` docs are not used for ROOT at all.


vepadulano:
We should wait for https://github.com/root-project/root/pull/20960 to be merged, then restart this PR builds with clean builds and also build and test the Python wheels (this also needs https://github.com/root-project/root/pull/20962 first)

guitargeek:
Sound good!

Issue #20913, opened 2026-01-16
MIGRAD can call G2 even if HasG2 is set to false


AdrianDBS:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Minuit2's gradient-based minimizers are occasionally calling the hessian diagonal method `G2` of a cost function even if the `HasG2` property is set to false. 
In other words, `HasG2`may just be ignored, at least during minimizer seeding.

See #20665

### Reproducer

Example:

1. Create a `FCNBase` subclass for the Rosenbrock problem:
```
function = Pow(1 - x, 2) + 100*Pow(-Pow(x, 2) + y, 2);

gradient[x] = 400*x*(Pow(x, 2) - y) + 2*x - 2;
gradient[y] = -200*Pow(x, 2) + 200*y;

hessian[x, x] = 1200*Pow(x, 2) - 400*y + 2;
hessian[x, y] = -400*x;
hessian[y, x] = -400*x;
hessian[y, y] = 200;
```
which
- returns true for `HasGradient` and `HasHessian`
- returns the above-mentioned values for the cost function value, gradient and hessian
- returns false for `HasG2`
- throws an exception when `G2` is called (or passively monitors a call to it) 

2. Minimize that problem for the initial values (x=0, y=1) using MnMigrad
3. See `G2` getting called although it shouldn't (at least to my mind)

### ROOT version

ROOT v6.36.06

### Installation method

Build from source

### Operating system

Windows

ferdymercury:
Thanks a lot for the report. Do you have a working reproducer (small ROOT script that we can run to reproduce) ?

guitargeek:
No need to provide a reproducer. I have created one myself and I see the problem. I'm on it.

```c++
// root repro.C
#include <Minuit2/FCNBase.h>
#include <Minuit2/MnMigrad.h>
#include <Minuit2/MnUserParameters.h>
#include <Minuit2/FunctionMinimum.h>

#include <vector>
#include <iostream>
#include <cmath>

class RosenbrockFCN : public ROOT::Minuit2::FCNBase {

   static int gG2CallCount;

public:
   static int StaticG2CallCount() { return gG2CallCount; }

   double Up() const override { return 1.0; }

   // Rosenbrock: (1-x)^2 + 100*(y-x^2)^2
   double operator()(std::vector<double> const &p) const override
   {
      const double x = p.at(0);
      const double y = p.at(1);
      const double a = (1.0 - x);
      const double b = (y - x * x);
[...]

PR #20936, opened 2026-01-18
[Minuit2] Fix external index usage in analytical Hessian

### LINKED ISSUE #20954: [Minuit2] Analytical Hessian indexing mismatch
When using analytic external Hessians in Minuit 2 (possible since https://github.com/root-project/root/commit/888a7677f7b4e7acea6e3ec8ddffd32a0e8e8597), the index mapping between Minuit-internal floating parameters and all the "external" parameters of the function is not applied correctly.

Reproducer:
```c++
#include <Minuit2/FCNBase.h>
#include <Minuit2/MnMigrad.h>
#include <Minuit2/MnUserParameters.h>
#include <Minuit2/FunctionMinimum.h>
#include <Minuit2/MnPrint.h>

#include <cassert>
#include <vector>
#include <cmath>
#include <iostream>

using namespace ROOT::Minuit2;

class QuadraticFCN : public FCNBase {
public:
   double operator()(const std::vector<double>& p) const override {
      const double x = p[0];
      const double y = p[1];
      const double z = p[2];
      return x*x + 10*y*y + 100*z*z + 2*x*y + 4*x*z + 8*y*z;
   }

   bool HasGradient() const override { return true; }
   bool HasHessian()  const override { return true; }
   bool HasG2()       const override { return false; }

   std::vector<double> Gradient(const std::vector<double>& p) const override {
      return {
         2*p[0] + 2*p[1] + 4*p[2],
         20*p[1] + 2*p[0] + 8*p[2],
         200*p[2] + 4*p[0] + 8*p[1]
      };
   }

   std::vector<double> Hessian(const std::vector<double>&) const override {
      // Row-major 3x3
      return {
          2,  2,   4,
          2, 20,   8,
          4,  8, 200
      };
   }

   double Up() const override { return 1.0; }
};

void test_Hessian_external_indexing_bug()
{
   QuadraticFCN fcn;

   MnUserParameters upar;
   upar.Add("x", 1.0, 0.1);
   upar.Add("y", 2.0, 0.1);
   upar.Add("z", 3.0, 0.1);

   // FIX x to break the internal/external identity mapping
   upar.Fix("x");

   MnMigrad migrad(fcn, upar);
   FunctionMinimum min = migrad();

   const auto& h = min.Error().Hessian();

   // Internal parameters are (y, z)
   // Expected Hessian:
   // [20  8]
   // [ 8 200]
   //
   // Packed: [20, 8, 200]

   std::cout << "Internal Hessian:\n";
   std::cout << "h(0,0) = " << h(0,0) << "\n";
   std::cout << "h(0,1) = " << h(0,1) << "\n";
   std::cout << "h(1,1) = " << h(1,1) << "\n";

   std::cout << "Expected Hessian:\n";
   std::cout << "h(0,0) = 20\n";
   std::cout << "h(0,1) = 8\n";
   std::cout << "h(1,1) = 200\n";
}
```
Output:
```txt
Internal Hessian:
h(0,0) = 2
h(0,1) = 2
h(1,1) = 20
Expected Hessian:
h(0,0) = 20
h(0,1) = 8
h(1,1) = 200
```

guitargeek:
Fix external index usage in analytical Hessian and also add a unit test to cover this former bug.

This problem exists since the introduction of external Hessians to Minuit2 in 888a7677f7b4e7a in the ROOT 6.28 development cycle.

This bug was discovered while investigating #20913.

Closes #20954.

### DIFF
--- a/math/minuit2/test/testMinuit2.cxx
+++ b/math/minuit2/test/testMinuit2.cxx
@@ -1,0 +1,121 @@
+// Tests for Minuit2, covering former bugs and regressions
+//
+// Author: Jonas Rembser, CERN 01/2026
+
+#include <Minuit2/FCNBase.h>
+#include <Minuit2/MnMigrad.h>
+#include <Minuit2/MnUserParameters.h>
+#include <Minuit2/FunctionMinimum.h>
+#include <Minuit2/MnPrint.h>
+
+#include <gtest/gtest.h>
+
+class QuadraticFCNBase : public ROOT::Minuit2::FCNBase {
+public:
+   double operator()(const std::vector<double> &p) const override
+   {
+      const double x = p[0];
+      const double y = p[1];
+      const double z = p[2];
+      // clang-format off
+      return x*x + 10*y*y + 100*z*z + 2*x*y + 4*x*z + 8*y*z;
+      // clang-format on
+   }
+
+   bool HasGradient() const override { return true; }
+   bool HasG2() const override { return false; }
+
+   std::vector<double> Gradient(const std::vector<double> &p) const override
+   {
+      // clang-format off
+      return {
+         2 * p[0] + 2 * p[1] + 4 * p[2],
+         20 * p[1] + 2 * p[0] + 8 * p[2],
+         200 * p[2] + 4 * p[0] + 8 * p[1]
+      };
+      // clang-format on
+   }
+
+   double Up() const override { return 1.0; }
+};
+
+class QuadraticFCNWithHessian : public QuadraticFCNBase {
+public:
+   bool HasHessian() const override { return true; }
+
+   std::vector<double> Hessian(const std::vector<double> &) const override
+   {
+      // Row-major 3x3 Hessian in *external* parameter space
+      // clang-format off
+      return {
+         2,   2,   4,
+         2,  20,   8,
+         4,   8, 200
+      };
+      // clang-format on
+   }
+};
+
+class QuadraticFCNNoHessian : public QuadraticFCNBase {
+public:
+   bool HasHessian() const override { return false; }
+};
+
+// ----------------------------------------------------------------------
+// Shared test helper
+// ----------------------------------------------------------------------
+
+template <typename FCN>
+void RunHessianExternalIndexingTest(const FCN &fcn, double tolerance)
+{
+   using namespace ROOT::Minuit2;
+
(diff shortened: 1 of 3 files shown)


math/minuit2/test/testMinuit2.cxx
```diff
@@ -0,0 +1,121 @@
...
+   ASSERT_EQ(hessian.Nrow(), 2);
+
+   // Internal parameters are (y, z)
+   //
+   // Expected Hessian:
+   // [20   8]
+   // [ 8 200]
+
+   EXPECT_NEAR(hessian(0, 0), 20.0, tolerance);
+   EXPECT_NEAR(hessian(0, 1), 8.0, tolerance);
```
hageboeck, line 96: Does one need to test the other diagonal element here or is it clear that it cannot be different?

guitargeek:
Ah, okay I get now what you mean! In another PR you made a similar comment, and there I thought you meant "disentangle fix+test from unrelated changes". Sorry for the misunderstanding.

Just to be clear for the future: you prefer test commit first or fix commit first? fix first, so that tests always pass, right?

hageboeck:
> Just to be clear for the future: you prefer test commit first or fix commit first? fix first, so that tests always pass, right?

Yes, precisely! And if the message says "Add test for xxx", you already know what to expect in the review. By splitting, you could even revert the fix and replace with a new one, and you would expect the test to stay green.

math/minuit2/test/testMinuit2.cxx
guitargeek: The `hessian` is a `MnAlgebraicSymMatrix`, so that is indeed redundant. I can change the `const auto &` to make that type more clear!

guitargeek:
Damn, sorry it was not split up in two commits in the end! I have done this in my local branch, and realized too late that the PR branch was not synced :slightly_frowning_face: Sorry, I really didn't mean to ignore your request! I juggled too many open PRs and lost the overview....

AdrianDBS:
Will this fix be backported to 6.36?

hageboeck:
> Damn, sorry it was not split up in two commits in the end! I have done this in my local branch, and realized too late that the PR branch was not synced 🙁 Sorry, I really didn't mean to ignore your request! I juggled too many open PRs and lost the overview....

No problem. 🙂

PR #20937, opened 2026-01-18
[RF] Group together many small tests in one executable 

guitargeek:
Building RooFit during development feels a bit sluggish, even with
`ccache` on. That is because lots of time is spent in linking many small
test executables, which `ccache` doesn't make faster.

This can be improved by grouping some tests execuables in RooFitCore and
RooFit, suggested in this commit.

Some comparisons (for a build with `roofit_multiprocess=OFF`):

  * With a hot cache, building `roofit/rooficore/test` and
    `roofit/roofit/test` after deleting these directories takes 10
    seconds instead of 30 on a single core with my Clang setup.

  * The number of tests in the same two directories is reduced from 76
    to 21

  * The time it takes to run all tests in the same directories on a
    single core reduces from 48 to 34 seconds

  * The new bundles tests are still relatively small:
    ```txt
    1/21 gtest-roofit-roofit-testRooFit ..........................   Passed    3.56 sec
    2/21 gtest-roofit-roofit-vectorisedPDFs-testVectorizedPDFs ...   Passed   11.64 sec
    4/21 gtest-roofit-roofitcore-testRooFitCore ..................   Passed    4.09 sec
    ```
  * In case of a failure on the CI, it will still be easy to see which
    test failed exactly, because the combined logs are still clean and
    not overwhelming (the longest is `testVectorizedPDFs` with 909
    lines)

### DIFF
--- a/roofit/roofit/test/CMakeLists.txt
+++ b/roofit/roofit/test/CMakeLists.txt
@@ -6,29 +6,35 @@
 
 # @author Stephan Hageboeck, CERN, 2019
 
-ROOT_ADD_GTEST(stabilityTests stabilityTests.cxx LIBRARIES RooFit)
-ROOT_ADD_GTEST(testRooBernstein testRooBernstein.cxx LIBRARIES RooFit)
-ROOT_ADD_GTEST(testRooBifurGauss testRooBifurGauss.cxx LIBRARIES RooFit)
-ROOT_ADD_GTEST(testRooCrystalBall testRooCrystalBall.cxx LIBRARIES Gpad RooFit)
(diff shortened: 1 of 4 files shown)


hageboeck (approved):
I counted the .cxx files in the other directories, but in RooFitCore I gave up. 🙂 
Did you verify that none was missed?

(+ `git clang-format`)

guitargeek:
Thanks for the review, good idea to check that all files made it over! Indeed, I forgot `loadOldWorkspace.cxx` in RooFitCore. I'll add this.

About the `clang-format`: I'd prefer not to do it, because it suggests to change the indentation of a whole test just because I added one line.

PR #20945, opened 2026-01-18
[v632][ci] Add Ubuntu 26.04

guitargeek:
To validate the older ROOT releases on the upcoming Ubuntu LTS.


guitargeek:
Adding Ubuntu 26.04 didn't work with minimal effort, and it's not strictly required because 6.32 is out of of the core support window.

PR #12449, opened 2023-03-08
[WIP] Improved error handling for template instantiation

bendavid:
Two substantive changes:
1) Explicitly catch errors in pyroot when the wrapper function fails to compile
2) Make sure template instantiation fails by catching clang errors within LookupHelper (not entirely sure this is exactly the right fix, @Axel-Naumann please take a look)

This PR partially fixes https://github.com/root-project/root/issues/11854

This PR does NOT address automatically providing the user with the underlying clang errors in case all template instantiation attempts from pyroot/cppyy ultimately fail.

There are also still problems with transactionality.

Consider the following test case:

test.h:
```cpp
template <typename T>
class Helper {

public:

  Helper() {}

  std::size_t operator() () const {
    const std::size_t res = 0;
    res = T{0, 0}.size();
    return res;
  }

};

template <typename H>
std::size_t call_helper(const H &helper) {
  return helper();
}
```

test.py (here setting gDebug=6 so that we see all the clang output):
```python
import ROOT

ret = ROOT.gInterpreter.Declare('#include "test.h"')
print("declare ret", ret)

print("creating helper")
helper = ROOT.Helper[ROOT.std.vector["double"]]()


print("calling helper")

ROOT.gDebug = 6

for i in range(2):
   print(f"call attempt {i}")
   try:
      res = ROOT.call_helper(helper)
      print("helper call succeeded:", res)
   except Exception as e:
      print("helper call failed")
      print(e)
```

output (note in particular that the "error: cannot assign to variable" occurs only on the first attempt):
```
Info in <TInterpreter::TCling::AutoParse>: Trying to autoparse for call_helper
Info in <TInterpreter::TCling::AutoLoad>: Trying to autoload for call_helper
Info in <TInterpreter::TCling::AutoParse>: Starting autoparse for call_helper

Info in <TInterpreter::TCling::AutoParse>: Trying to autoparse for call_helper
Info in <TInterpreter::TCling::AutoLoad>: Trying to autoload for call_helper
Info in <TInterpreter::TCling::AutoParse>: Starting autoparse for call_helper

Info in <TInterpreter::TCling::AutoParse>: Trying to autoparse for call_helper
Info in <TInterpreter::TCling::AutoLoad>: Trying to autoload for call_helper
Info in <TInterpreter::TCling::AutoParse>: Starting autoparse for call_helper

Info in <TInterpreter::TCling::AutoParse>: Trying to autoparse for call_helper
[...]

PR #20946, opened 2026-01-19
[cling] Do not suppress all diagnostics during function lookup

vepadulano:
This is the first part of https://github.com/root-project/root/pull/12449

I'm still trying to understand how to create a test. I used the original reproducer

```
template <typename T>
class Helper {

public:

  Helper() {}

  std::size_t operator() () const {
    const std::size_t res = 0;
    res = T{0, 0}.size();
    return res;
  }

};

template <typename H>
std::size_t call_helper(const H &helper) {
  return helper();
}
```

And ran it with `root`, but I see the same errors printed both after this changes and with ROOT master
```
Processing t.C...
In file included from input_line_8:1:
/Users/vpadulan/Programs/rootproject/t.C:10:9: error: cannot assign to variable 'res' with const-qualified type 'const std::size_t' (aka 'const unsigned long')
    res = T{0, 0}.size();
    ~~~ ^
/Users/vpadulan/Programs/rootproject/t.C:18:10: note: in instantiation of member function 'Helper<std::vector<double, std::allocator<double> > >::operator()' requested here
  return helper();
         ^
/Users/vpadulan/Programs/rootproject/t.C:24:5: note: in instantiation of function template specialization 'call_helper<Helper<std::vector<double, std::allocator<double> > > >' requested here
    call_helper(h);
    ^
/Users/vpadulan/Programs/rootproject/t.C:9:23: note: variable 'res' declared const here
    const std::size_t res = 0;
    ~~~~~~~~~~~~~~~~~~^~~~~~~
Assertion failed: (!m_Unloading && "Must not nest within unloading transaction"), function addNestedTransaction, file Transaction.cpp, line 98.
```

### DIFF
--- a/roofit/roofitcore/inc/RooAbsDataHelper.h
+++ b/roofit/roofitcore/inc/RooAbsDataHelper.h
@@ -143,9 +143,8 @@ public:
    void Exec(unsigned int slot, ColumnTypes... values)
    {
       std::vector<double> &vector = _events[slot];
-      for (auto &&val : {static_cast<double>(values)...}) {
-         vector.push_back(val);
-      }
+      std::array<double, sizeof...(ColumnTypes)> valuesArr{static_cast<double>(values)...};
+      vector.insert(vector.end(), valuesArr.begin(), valuesArr.end());
 
       ExecImpl(sizeof...(values), vector);
    }


vepadulano:
This PR is currently blocked by https://github.com/root-project/root/issues/20952

PR #20947, opened 2026-01-19
[main] optparse: add support for prefix flags (per-flag version)

silverweed:
Alternative version of https://github.com/root-project/root/pull/20923, that allows to define prefix flags per-flag rather than globally.

### DIFF
--- a/main/test/optparse_test.cxx
+++ b/main/test/optparse_test.cxx
@@ -664,7 +664,7 @@ TEST(OptParse, PrefixMultiple)
    opts.Parse(args, std::size(args));
 
    EXPECT_TRUE(opts.GetErrors().empty());
-   EXPECT_EQ(opts.GetFlagValues("D"), std::vector<std::string_view>({"name", "name", "name"}));
+   EXPECT_EQ(opts.GetFlagValues("D"), std::vector<std::string_view>({"name", "=name", "name"}));
 }
 
 TEST(OptParse, PrefixWithEqual)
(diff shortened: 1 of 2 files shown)


devajithvs:
Thanks @silverweed. I've updated the draft PR: https://github.com/root-project/root/pull/20903 to use this.

main/test/optparse_test.cxx
```diff
@@ -589,3 +589,141 @@ TEST(OptParse, MultipleFlagsAsIntError)
...
+TEST(OptParse, PrefixWithEqual)
+{
+   ROOT::RCmdLineOpts opts;
+   opts.AddFlag({"--D"}, ROOT::RCmdLineOpts::EFlagType::kWithArg, "",
+                ROOT::RCmdLineOpts::kFlagAllowMultiple | ROOT::RCmdLineOpts::kFlagPrefixArg);
+
+   const char *args[] = {"somename", "--Df=a"};
+   opts.Parse(args, std::size(args));
+
+   EXPECT_EQ(opts.GetErrors().size(), 1);
```
devajithvs: Why should this test fail?

I think since the CI for `rootcling` is happy we don't do this anywhere. But I can think of a case where we might need to pass a compile definition: `-Dsomething=ON` 

If we register `--D` as a `kFlagPrefixArg`, we already have a check that prevents another flag that starts with `D` prefix, can't we make this valid?

silverweed: I guess we can; I just found the pattern to be confusing but indeed I didn't think of the compiler definition case, where you do want something like that.
I'll change it

PR #20953, opened 2026-01-19
[cppyy] Gracefully fail in `Cppyy::CallO()` if TypeInfo not available to avoid crash in boost test

guitargeek:
Instead of trying to use a `nullptr` TypeInfo in `ClassInfo_Size`, the `CallO()` clingwrapper function should just return NULL and let CPyCppyy handle the failed call - which it does, by either throwing a Python exception or returning a default-constructed return value in the case of strings. See `Executors.cxx` in CPyCppyy.

This avoids a crash in the failing cppyy boost type erasure test, which can now be marked as `xfail(strict=True)`.

### DIFF
--- a/bindings/pyroot/cppyy/cppyy/test/test_boost.py
+++ b/bindings/pyroot/cppyy/cppyy/test/test_boost.py
@@ -141,7 +141,7 @@ class TestBOOSTERASURE:
         cppyy.include("boost/type_erasure/member.hpp")
         cppyy.include("boost/mpl/vector.hpp")
 
-    @mark.skipif(noboost, reason="boost not found")
+    @mark.xfail(strict=True, reason="TClassInfo for 'Lengths' is missing")
     def test01_erasure_usage(self):
         """boost::type_erasure usage"""
 


bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx
```diff
@@ -1113,7 +1113,12 @@ Cppyy::TCppObject_t Cppyy::CallO(TCppMethod_t method,
     TCppObject_t self, size_t nargs, void* args, TCppType_t result_type)
 {
     TClassRef& cr = type_from_handle(result_type);
-    void* obj = ::operator new(gInterpreter->ClassInfo_Size(cr->GetClassInfo()));
+    auto *classInfo = cr->GetClassInfo();
+    // If the calss info is missing, we better return null and let cppyy
```
vepadulano:
```suggestion
    // If the class info is missing, we better return null and let cppyy
```

bindings/pyroot/cppyy/cppyy-backend/clingwrapper/src/clingwrapper.cxx
```diff
@@ -1113,7 +1113,12 @@ Cppyy::TCppObject_t Cppyy::CallO(TCppMethod_t method,
...
+    // handle the error, before we step into undefined behavior
+    if(!classInfo)
+       return (TCppObject_t)0;
+    void* obj = ::operator new(gInterpreter->ClassInfo_Size(classInfo));
```
vepadulano: Since we're adding the checks, what about a check on the value returned by ClassInfo_Size?



```suggestion
    auto classSize = gInterpreter->ClassInfo_Size(classInfo);
    if (classSize < 0) // ClassInfo_Size returns -1 in case of invalid info
        return (TCppObject_t)0;
    void* obj = ::operator new(classSize);
```

vepadulano (changes requested):

vepadulano (approved):
LGTM, feel free to merge once the CI is green.

Issue #20665, opened 2025-12-08
Analytical Hessian is not made positive-definite during minimizer seeding in Minuit2


AdrianDBS:
### Explain what you would like to see improved and how.

When I provide my cost function with an analytical Hessian, I would expect the same/equivalent behavior as for the "numerical path" (just with fewer function calls).

While this seems to be the case for `MnHesse`, where both analytical and numerically approximated Hessians are made positive-definite, the behavior at minimizer seeding is different (inconsistent to my mind): 
When an analytical Hessian is provided and it is not positive-definite for the given initial parameter values, it is left non-positive-definite which leads to the minimizer initially stepping away from the optimal solution and can even lead to unexpected minimization failures where the "numerical path" for the same scenario succeeds. 

I think the analytical Hessian should be made positive-definite for the minimizer seeding as well. But maybe I am missing something?!

Simple example:
Use MIGRAD to minimize the sum of squared residuals between the following exponential decay model and toy data (generated from the model using amplitude p[0] = 3, rate p[1] = 2, offset p[2] = 1 and adding random normal noise with a standard deviation of 0.01):

modelFunc(x, p) = p[0] * exp(-p[1] * x) + p[2]
modelGrad(x, p) = [exp(-p[1] * x), -p[0] * x * exp(-p[1] * x), 1]
modelHess(x, p) = [0,  -x * exp(-p[1] * x), 0,  -x * exp(-p[1] * x), p[0] * x * x * exp(-p[1] * x), 0, 0, 0, 0]

x = [0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9, 2, 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, 2.8, 2.9]
y = [3.99, 3.45, 3.02, 2.65, 2.36, 2.11, 1.9, 1.75, 1.6, 1.49, 1.41, 1.32, 1.28, 1.22, 1.19, 1.15, 1.14, 1.09, 1.08, 1.07, 1.03, 1.05, 1.05, 1.04, 1.03, 1.02, 1.03, 1.02, 1.02, 1.01]

[...]

PR #20957, opened 2026-01-19
[Minuit2] Fallback to full Hessian in AnalyticalGradientCalculator if no G2 

guitargeek:
If the user function doesn't claim to implement the diagonal of the
Hessian in an efficient way by overriding `HasG2()` to return `true`,
the AnalyticalGradientCalculator should not call `G2()`, but instead
fall back to extract the diagonal from the full Hessian.

Closes https://github.com/root-project/root/issues/20913.

A test is implemented in a separate commit, and there is also a third commit with some code simplification.

### DIFF
--- a/math/minuit2/src/AnalyticalGradientCalculator.cxx
+++ b/math/minuit2/src/AnalyticalGradientCalculator.cxx
@@ -30,12 +30,8 @@ FunctionGradient AnalyticalGradientCalculator::operator()(const MinimumParameter
    MnAlgebraicVector v(par.Vec().size());
    for (unsigned int i = 0; i < par.Vec().size(); i++) {
       unsigned int ext = fTransformation.ExtOfInt(i);
-      if (fTransformation.Parameter(ext).HasLimits()) {
-         double dd = fTransformation.DInt2Ext(i, par.Vec()(i));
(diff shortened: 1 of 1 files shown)


guitargeek:
@AdrianDBS, to which release branches do you need this backported?

And does this also fix #20665?

AdrianDBS:
> @AdrianDBS, to which release branches do you need this backported?

Would be great if this could be backported to 6.36.

> And does this also fix #20665?

I will test this today. Yet, while it will most certainly fix the inconsistency described in the first comment in the ticket, I don't think it will solve the main issue described there.

dpiparo:
the mac26 failure seems spurious.

guitargeek:
> the mac26 failure seems spurious.

Indeed it's a sporadic failure that we have suppressed in the past (with https://github.com/root-project/root/pull/19043), but in commit 50ae29d43431e61d I have re-enabled such sporadically failing tests again so that we see if they are still there and if yes address them. The problem with disabling these tests is that we won't know if we regress with the cppyy upgrade.

guitargeek:
@AdrianDBS, thanks for trying it out, I'm looking forward to your report! The backport to 6.36 won't be a problem.

PR #20958, opened 2026-01-19
[CMake] Fix problems in handling Python code in cppyy and ROOT

guitargeek:
The CMake code for managing the ROOT Pythonization Python sources had many flaws:

  1. Overly verbose, because every Python source file had its own copy target

  2. Lots of time wasted in starting up the Python interpreter for compiling Python sources to byte code. It is several seconds faster to call `python -m py_compile` once, with all the sources passed as arguments. Fixing this makes the build step when changing only Python sources feel instant again.

  3. Wrong dependency graph (introduced with commit b824594), because the Python sources were compiled to byte code as a POST_BUILD step for the CPython extension library, which is not strictly related. As a result, the Python code was not re-compiled when *only* Python sources were updated, and unnecessarily re-compiled when the CPython extension was changed.

The `cppyy` package also has the first two problems, but instead of the third problem with the wrong dependency graph, it didn't have targets for copying and compiling the Python sources at all. It used `copy(FILE ..)` and `install(CODE ..)` to do the copying and configuration time and the compiling at install time, which makes incremental builds mute and developing the `cppyy` Python package without manual copying impossible.

This commit re-implements the handling of the Python code in CMake for these two Python packages to solve all of these problems.

As a result, things are now clean, correct, and fast. To illustrate, here is the Ninja output for building `bindings/pyroot/cppyy` and `bindings/pyroot/pythonizations`:

```txt
[3/9] Copying cppyy Python sources
[4/9] Linking CXX shared library lib/cppyy/libcppyy.so
[5/9] Compiling cppyy Python sources
[6/9] Copying ROOT Pythonization Python sources
[7/9] Compiling ROOT Pythonization Python sources
[8/9] Linking CXX shared library lib/ROOT/libROOTPythonizations.so
```

The C++ translation units are missing, because they were c-cached in that case.

As a followup, one might consider writing CMake functions in `RootMacros.cmake` to avoid code repetition and then also use it for JupyROOT and DistRDF, which have similar problems. But I thought that was a too big of a change to do and review at once. My suggestion would be to first establish the new pattern in cppyy and ROOTPythonizations, and then abstract and expand from there.

### DIFF
--- a/bindings/pyroot/cppyy/cppyy/CMakeLists.txt
+++ b/bindings/pyroot/cppyy/cppyy/CMakeLists.txt
@@ -19,19 +19,60 @@ set(py_sources
   cppyy/types.py
 )
 
-set(cppyyPySrcDir python/cppyy)
-file(COPY ${cppyyPySrcDir} DESTINATION ${localruntimedir})
+# Ensure output directory exists
+file(MAKE_DIRECTORY ${localruntimedir}/cppyy)
+
+set(cppyy_copy_commands)
+set(cppyy_py_sources_in_source_dir)
+set(cppyy_py_sources_in_localruntimedir)
+
+foreach(py_source ${py_sources})
+  list(APPEND cppyy_copy_commands
+       COMMAND ${CMAKE_COMMAND} -E copy_if_different
+               ${CMAKE_CURRENT_SOURCE_DIR}/python/${py_source}
+               ${localruntimedir}/${py_source})
+
+  list(APPEND cppyy_py_sources_in_source_dir
+       ${CMAKE_CURRENT_SOURCE_DIR}/python/${py_source})
+
+  list(APPEND cppyy_py_sources_in_localruntimedir
+       ${localruntimedir}/${py_source})
+endforeach()
+
+add_custom_command(
+  OUTPUT ${cppyy_py_sources_in_localruntimedir}
+  ${cppyy_copy_commands}
+  DEPENDS ${cppyy_py_sources_in_source_dir}
+  COMMENT "Copying cppyy Python sources"
+)
+
+add_custom_target(cppyyPySources ALL DEPENDS ${cppyy_py_sources_in_localruntimedir})
 
 # Compile .py files
-foreach(py_source ${py_sources})
-  install(CODE "execute_process(COMMAND ${Python3_EXECUTABLE} -m py_compile ${localruntimedir}/${py_source})")
-  install(CODE "execute_process(COMMAND ${Python3_EXECUTABLE} -O -m py_compile ${localruntimedir}/${py_source})")
-endforeach()
+
+# Stamp file so CMake knows it doesn't need to re-compile. We can't set the
+# actual bytecode files as the OUTPUT of the custom command, because their
+# names are CPython implementation dependent and therefore not reliable.
+set(cppyy_bytecode_stamp ${localruntimedir}/cppyy/bytecode.stamp)
+
+# It's 10x faster to compile all in one go than in single invocations
+add_custom_command(
+  OUTPUT ${cppyy_bytecode_stamp}
+  COMMAND ${Python3_EXECUTABLE} -m py_compile ${cppyy_py_sources_in_localruntimedir}
+  COMMAND ${Python3_EXECUTABLE} -O -m py_compile ${cppyy_py_sources_in_localruntimedir}
+  COMMAND ${CMAKE_COMMAND} -E touch ${cppyy_bytecode_stamp}
+  DEPENDS ${cppyy_py_sources_in_localruntimedir}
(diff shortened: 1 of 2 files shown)


bindings/pyroot/cppyy/cppyy/CMakeLists.txt
```diff
@@ -19,19 +19,69 @@ set(py_sources
...
+
+  list(APPEND cppyy_py_sources_in_localruntimedir
+       ${localruntimedir}/${py_source})
 endforeach()
 
+set(cppyy_py_sources_copy_stamp ${localruntimedir}/cppyy/copy.stamp)
+
+add_custom_command(
+  OUTPUT ${cppyy_py_sources_copy_stamp}
+  ${cppyy_copy_commands}
```
hageboeck: I would set the `OUTPUT` directly to `cppyy_py_sources_in_localruntimedir`. Otherwise, if I deleted one (or all) of the `.py` files, but the stamp file survives (rm *.py), no copy will be triggered.
For this step, the stamp file wouldn't be needed.

bindings/pyroot/cppyy/cppyy/CMakeLists.txt
```diff
@@ -19,19 +19,69 @@ set(py_sources
...
+  COMMAND ${CMAKE_COMMAND} -E touch ${cppyy_py_sources_copy_stamp}
+  DEPENDS ${cppyy_py_sources_in_source_dir}
+  COMMENT "Copying cppyy Python sources"
+)
+
+add_custom_target(cppyyPySources ALL DEPENDS ${cppyy_py_sources_copy_stamp})
```
hageboeck: Set the DEPENDS to `cppyy_py_sources_in_localruntimedir`

bindings/pyroot/cppyy/cppyy/CMakeLists.txt
```diff
@@ -19,19 +19,69 @@ set(py_sources
...
+# but that backfires: the most time spent in compiling the sources is spent in
+# firing up the Python interpreter, so it's about 10x faster to compile all
+# sources at once, and the overall runtime is insignificant.
+# Therefore we intentionally compile all sources in a single invocation.
+add_custom_command(
+  OUTPUT ${cppyy_bytecode_stamp}
+  COMMAND ${Python3_EXECUTABLE} -m py_compile ${cppyy_py_sources_in_localruntimedir}
+  COMMAND ${Python3_EXECUTABLE} -O -m py_compile ${cppyy_py_sources_in_localruntimedir}
+  COMMAND ${CMAKE_COMMAND} -E touch ${cppyy_bytecode_stamp}
+  DEPENDS ${cppyy_py_sources_copy_stamp}
```
hageboeck: Use `cppyy_py_sources_in_localruntimedir` directly. Any change will trigger a rebuild

bindings/pyroot/cppyy/cppyy/CMakeLists.txt
```diff
@@ -19,19 +19,69 @@ set(py_sources
...
+  COMMENT "Compiling cppyy Python sources"
+)
+
+add_custom_target(cppyyPyBytecode ALL DEPENDS ${cppyy_bytecode_stamp})
+
+# To trigger re-compilation if Python sources are updated in the build tree
+add_dependencies(cppyyPyBytecode cppyyPySources)
```
hageboeck: This should be unnecessary if the dependency to the copy is done directly using the generated files.

bindings/pyroot/cppyy/cppyy/CMakeLists.txt
```diff
@@ -19,19 +19,69 @@ set(py_sources
...
+# Stamp file so CMake knows it doesn't need to re-compile. We can't set the
+# actual bytecode files as the OUTPUT of the custom command, because their
+# names are CPython implementation dependent and therefore not reliable.
+set(cppyy_bytecode_stamp ${localruntimedir}/cppyy/bytecode.stamp)
+
+# One could imagine to have one custom command per file as an "optimization",
+# but that backfires: the most time spent in compiling the sources is spent in
+# firing up the Python interpreter, so it's about 10x faster to compile all
+# sources at once, and the overall runtime is insignificant.
+# Therefore we intentionally compile all sources in a single invocation.
```
hageboeck: Less is more. Why not just?
```
It's 10x faster to compile all in one go than in single invocations
```

guitargeek: Fair! No need for brain dump comments like this :laughing:

guitargeek:
Thanks for the great review! I implemented all your suggestions 1-to-1

guitargeek:
Python wheel test failures are unrelated and will be fixed by: https://github.com/root-project/root/pull/20962

PR #20961, opened 2026-01-20
[RF] Relax test thresholds for a likelihood test.

hageboeck:
When built with -march=native -O3 on Mac26, the tests started failing due to numerical differences at the relative 1.E-12 level. These should be acceptable for a numerical algorithm.


hageboeck:
After offline discussion, I'll close this. It seems that the desire was to be numerically exact, at least in `-O2`, and that `-O3 -march=native` wasn't targeted. One could of course try to write conditions for this, but that does not seem to be worth the effort at the moment.

guitargeek:
Good call, thank you! FYI, @egpbos

PR #20962, opened 2026-01-20
[build] Skip all histv7 tutorials in Python wheels tests

vepadulano:


test/wheels/test_tutorials.py
```diff
@@ -23,6 +23,7 @@
     "h1analysisProxy.C",  # helper macro, not meant to run standalone
     "hist001_RHist_basics.C",  # required RHist, not supported in ROOT wheels
     "hist002_RHist_weighted.C",  # required RHist, not supported in ROOT wheels
+    "hist003_RHist_multi.C",  # required RHist, not supported in ROOT wheels
```
hahnjo: Can we exclude the entire subdirectory? I will forget again when adding other tutorials, and none of them will work...

PR #20963, opened 2026-01-20
Parallel overlap checking with staged workflow

agheata:
# This Pull request:
Refactors and accelerates geometry overlap checking by splitting the procedure into well-defined stages, introducing candidate filtering and parallel execution, and fixing long-standing meshing and caching issues affecting correctness and performance. Preserves backward-compatibility of interfaces.

## Changes or fixes:
* Three-stage overlap checking workflow
image

  1. Candidate enumeration
      - Traverses the geometry tree.
      - Uses **oriented bounding box (OBB) tests** to reject non-overlapping volume pairs early. This gives in general a **sizeable** candidate reduction compared to the legacy version, and a **massive** one for assembly-dominated setups (a factor of ~1000 for ALICE O² geometry)
  2. Surface point generation and caching
      - Generates surface sampling points per shape using the values set via `TGeoManager::SetPoints()` (default 20) and `TGeoManager::SetNmeshVertices()` (default 1000).
      - Introduces a thread-safe per-shape mesh cache inside `TGeoChecker`.
      - Cache invalidation is correctly handled when the number of mesh points or mesh segments changes
      - Fixes incorrect behavior caused by:
         - reuse of static `TBuffer3D` buffers
         - stale mesh caches in `TGeoCompositeShape / TGeoBoolNode`
         - inconsistent `GetPointsOnSegments()` contracts in several shapes
  3. Exact overlap checking following check points density
     - Performs navigation-based overlap tests using cached surface points.
     - This stage is now **parallelized** using ROOT Implicit MT.
 
* Parallel execution
  - Only the **final overlap checking stage** is parallelized (safe and scalable).
  - Uses `ROOT::TThreadExecutor` and respects `ROOT::EnableImplicitMT(N)`.
  - Demonstrates good strong scaling on realistic detector geometries.

* Correctness fixes in meshing
  - Reworked `GetPointsOnSegments()` for several shapes to:
     - avoid dependence on TBuffer3D static storage
     - generate points only on valid surface generators
     - respect a strict contract: return `true` _only if exactly_ `npoints` _are filled_
  - Fixed incorrect assumptions in legacy implementations (e.g. `TGeoBBox`, `TGeoTube`, `TGeoCone`, `TGeoPcon`, `TGeoCtub`).

* Performance impact
  - The check is **more thorough** (points more uniformly spread on solid surfaces) and typically much faster, because it reduces the number of candidates for the expensive check, and parallelizes this most expensive part.
  - On typical setups (e.g. cmsRun4D116), the gain using 32 threads is significant 
 
c_strong_scaling_cms

  - On assembly-rich setups (e.g. o2sim_geometry) the gain is **massive** (from ~4 hours to ~15 sec on the same machine)
  - Below is a strong scaling plot showing both the total time and the time for the parallelized checking stage for ALICE. Note that this is **not** the typical profile, it is shown mainly to show the scaling of the parallelized stage. For most setups, the weight of the sequential stages 1 and 2 is much smaller that the weight of the parallel stage 3.
c_strong_scaling

## Checklist:

- [x] tested changes locally

This PR was done at the request of ALICE experiment, but provides setup-independent enhancements of the overlap checking feature

### DIFF
--- a/geom/geom/inc/TGeoMatrix.h
+++ b/geom/geom/inc/TGeoMatrix.h
@@ -464,7 +464,7 @@ private:
 
 public:
    TGeoHMatrix();
-   TGeoHMatrix(const TGeoHMatrix &other) : TGeoHMatrix((TGeoMatrix &)other) {}
+   TGeoHMatrix(const TGeoHMatrix &other);
    TGeoHMatrix(const TGeoMatrix &matrix);
    TGeoHMatrix(const char *name);
    ~TGeoHMatrix() override;
--- a/geom/geom/src/TGeoMatrix.cxx
+++ b/geom/geom/src/TGeoMatrix.cxx
@@ -2409,7 +2409,17 @@ TGeoHMatrix::TGeoHMatrix(const char *name) : TGeoMatrix(name)
 }
 
 ////////////////////////////////////////////////////////////////////////////////
-/// assignment
+/// copy constructor
+
+TGeoHMatrix::TGeoHMatrix(const TGeoHMatrix &other) : TGeoMatrix(other)
+{
+   memcpy(fTranslation, other.fTranslation, kN3);
+   memcpy(fRotationMatrix, other.fRotationMatrix, kN9);
+   memcpy(fScale, other.fScale, kN3);
+}
+
+////////////////////////////////////////////////////////////////////////////////
+/// copy constructor from generic matrix
 
 TGeoHMatrix::TGeoHMatrix(const TGeoMatrix &matrix) : TGeoMatrix(matrix)
 {


agheata:
Working on fixing no-IMT mode

agheata:
The failure in CI is on Windows only, it seems to throw an exception in TThreadExecutor::Foreach during overlap checking, not because the test gtest-geom-boolean-extrusion produces a wrong result, because I added a print that would appear in case of a bad result and it doesn't. I don't know if this is a known feature on Windows. The PR is now ready for review.

geom/geom/inc/TGeoBBox.h
```diff
@@ -65,6 +66,7 @@ class TGeoBBox : public TGeoShape {
    void GetBoundingCylinder(Double_t *param) const override;
    const TBuffer3D &GetBuffer3D(Int_t reqSections, Bool_t localFrame) const override;
    Int_t GetByteCount() const override { return 36; }
+   inline TVector3 GetDimensions() const { return TVector3(fDX, fDY, fDZ); }
```
dpiparo, line 69: we try to avoid these classes: TVectorX is "legacy": https://root.cern.ch/doc/master/classTVector3.html. Could ROOT::Math::XYZVector be used instead?

geom/geom/inc/TGeoManager.h
```diff
@@ -221,12 +224,14 @@ class TGeoManager : public TNamed {
...
    Int_t GetVisLevel() const;
    Int_t GetVisOption() const;
+   void InvalidateMeshCaches();
    Bool_t IsInPhiRange() const;
    Bool_t IsDrawingExtra() const { return fDrawExtra; }
    Bool_t IsNodeSelectable() const { return fIsNodeSelectable; }
    Bool_t IsVisLeaves() const { return fVisOption == 1; }
    void ModifiedPad() const;
-   void OptimizeVoxels(const char *filename = "tgeovox.C");                                     // *MENU*
+   void OptimizeVoxels(const char *filename = "tgeovox.C"); // *MENU*
```
dpiparo, line 233: from where is this name coming?

geom/geom/inc/TGeoMatrix.h
```diff
@@ -19,6 +19,7 @@
  *************************************************************************/
 
 #include "TNamed.h"
+#include "TVector3.h"
```
dpiparo, line 22: See comment above, that applies to all usages of TVector3.

geom/geom/inc/TGeoTube.h
```diff
@@ -72,10 +72,10 @@ class TGeoTube : public TGeoBBox {
    virtual Double_t GetRmin() const { return fRmin; }
    virtual Double_t GetRmax() const { return fRmax; }
    virtual Double_t GetDz() const { return fDz; }
-   Bool_t HasRmin() const { return (fRmin > 0) ? kTRUE : kFALSE; }
+   inline Bool_t HasRmin() const { return (fRmin > 0) ? kTRUE : kFALSE; }
```
dpiparo, line 75: `return fRmin > 0;` ?

geom/geom/inc/TGeoTube.h
```diff
@@ -153,8 +153,9 @@ class TGeoTubeSeg : public TGeoTube {
    void GetMeshNumbers(Int_t &nvert, Int_t &nsegs, Int_t &npols) const override;
    Double_t GetPhi1() const { return fPhi1; }
    Double_t GetPhi2() const { return fPhi2; }
+   inline Bool_t HasDphi() const { return (fPhi2 - fPhi1) < 360.; }
```
dpiparo, line 156: Just to be on the safe side: is this subtraction ok in all circumstances or is some more detailed checking needed? (it's not in general, but it could be here!)

geom/geom/inc/TVirtualGeoChecker.h
```diff
@@ -9,7 +9,10 @@
 #ifndef ROOT_TVirtualGeoChecker
 #define ROOT_TVirtualGeoChecker
 
+#include <vector>
```
dpiparo, line 12: Is it perhaps more standard to include stl includes after ROOT's?

geom/geom/inc/TVirtualGeoChecker.h
```diff
@@ -34,19 +38,24 @@ class TVirtualGeoChecker : public TObject {
    virtual void CheckGeometryFull(Bool_t checkoverlaps = kTRUE, Bool_t checkcrossings = kTRUE, Int_t nrays = 10000,
                                   const Double_t *vertex = nullptr) = 0;
    virtual void CheckGeometry(Int_t nrays, Double_t startx, Double_t starty, Double_t startz) const = 0;
-   virtual void CheckOverlaps(const TGeoVolume *vol, Double_t ovlp = 0.1, Option_t *option = "") = 0;
+   virtual void CheckOverlapsBySampling(const TGeoVolume *vol, Double_t ovlp = 0.1, Int_t npoints = 1000000) const = 0;
```
dpiparo, line 41: There is some variety of pointers and references in the signatures: could we adopt only references?

geom/geom/inc/TVirtualGeoChecker.h
```diff
@@ -21,11 +24,12 @@ class TStopwatch;
 class TVirtualGeoChecker : public TObject {
 protected:
    static TVirtualGeoChecker *fgGeoChecker; // Pointer to checker instance
-
+   Int_t fNmeshPoints{1000};                // Number of mesh points per shape
 public:
    TVirtualGeoChecker();
    ~TVirtualGeoChecker() override;
 
+   virtual void BuildMeshPointsCache(const std::vector<TGeoOverlapCandidate> &candidates) = 0;
```
dpiparo, line 32: Are the new methods perhaps a bit light in terms of doxygen documentation?

geom/geom/src/TGeoBBox.cxx
```diff
@@ -702,38 +701,58 @@ Bool_t TGeoBBox::GetPointsOnFacet(Int_t index, Int_t npoints, Double_t *array) c
...
-         x = p0[0] + (j + 1) * dx;
-         y = p0[1] + (j + 1) * dy;
-         z = p0[2] + (j + 1) * dz;
-         array[icrt++] = x;
-         array[icrt++] = y;
-         array[icrt++] = z;
-         ipoints--;
+         nperseg = remaining;
+
+      const Double_t dx = (p1[0] - p0[0]) / (nperseg + 1);
```
dpiparo, line 739: if this is performance critical, we could avoid 3 divisions and just have one

geom/geom/src/TGeoBBox.cxx
```diff
@@ -816,19 +851,159 @@ void TGeoBBox::InspectShape() const
    printf("    origin: x=%11.5f y=%11.5f z=%11.5f\n", fOrigin[0], fOrigin[1], fOrigin[2]);
 }
 
+////////////////////////////////////////////////////////////////////////////////
+/// @brief Test whether a given axis is a separating axis for two oriented boxes.
```
dpiparo, line 855: Thanks for this *great* doc!

geom/geom/src/TGeoBBox.cxx
```diff
@@ -816,19 +851,159 @@ void TGeoBBox::InspectShape() const
...
+   const Double_t rA = dA.Mag();
+   const Double_t rB = dB.Mag();
+   const Double_t dmax = rA + rB - tol;
+
+   // Touching within tolerance is treated as non-intersecting
+   if (D.Mag2() >= dmax * dmax)
+      return kFALSE;
+
+   // World-space box axes
+   TVector3 Ax, Ay, Az;
```
dpiparo, line 973: not too important for *implementations* but TVector is legacy: it is highly preferable to use ROOT::Math::XYZVector in the implementations and it is basically compulsory to do so in the interface.

geom/geom/src/TGeoCone.cxx
```diff
@@ -754,50 +753,89 @@ TGeoShape *TGeoCone::GetMakeRuntimeShape(TGeoShape *mother, TGeoMatrix * /*mat*/
...
+      const Int_t q = nSurfPoints / nGen;
+      const Int_t r = nSurfPoints % nGen;
+
+      for (Int_t ig = 0; ig < nGen; ++ig) {
+         const Int_t m = q + (ig < r ? 1 : 0); // points on this generator
+         if (m <= 0)
+            continue;
+
+         const Double_t phi = ig * dphi;
+         const Double_t c = TMath::Cos(phi);
```
dpiparo, line 799: opportunity to save cycles with sincos.

geom/geom/src/TGeoCone.cxx
```diff
@@ -2568,41 +2564,104 @@ const TBuffer3D &TGeoConeSeg::GetBuffer3D(Int_t reqSections, Bool_t localFrame)
...
+      const Double_t dphi = dphiSpan / (Double_t)nGen;
+
+      for (Int_t ig = 0; ig < nGen; ++ig) {
+         const Int_t m = q + (ig < r ? 1 : 0); // points on this generator
+         if (m <= 0)
+            continue;
+
+         // Center of the φ bin => never exactly on the cut planes
+         const Double_t phi = phi1 + (ig + 0.5) * dphi;
+         const Double_t c = TMath::Cos(phi);
```
dpiparo, line 2627: sincos opportunity

geom/geom/src/TGeoMatrix.cxx
```diff
@@ -335,6 +334,32 @@ void TGeoMatrix::GetHomogenousMatrix(Double_t *hmat) const
...
+///   [ r3 r4 r5 ]
+///   [ r6 r7 r8 ]
+///
+/// The columns of this matrix represent the transformed local axes.
+///
+/// @param[out] ax World-space direction of the local X axis.
+/// @param[out] ay World-space direction of the local Y axis.
+/// @param[out] az World-space direction of the local Z axis.
+
+void TGeoMatrix::GetWorldAxes(TVector3 &ax, TVector3 &ay, TVector3 &az) const
```
dpiparo, line 355: TVector3 is legacy and XYZVector should be used instead in the interface.

geom/geom/src/TGeoNode.cxx
```diff
@@ -69,26 +69,31 @@ painting a node on a pad will take the corresponding volume attributes.
 */
 
 #include <iostream>
```
dpiparo, line 71: stl includes after ROOT's?

geom/geom/src/TGeoNode.cxx
```diff
@@ -499,7 +627,7 @@ void TGeoNode::SaveAttributes(std::ostream &out)
 
 void TGeoNode::SetUserExtension(TGeoExtension *ext)
 {
-   TGeoExtension* tmp = fUserExtension;
+   TGeoExtension *tmp = fUserExtension;
```
dpiparo, line 630: what is the goal of this line?

geom/geom/src/TGeoNode.cxx
```diff
@@ -517,7 +645,7 @@ void TGeoNode::SetUserExtension(TGeoExtension *ext)
 
 void TGeoNode::SetFWExtension(TGeoExtension *ext)
 {
-   TGeoExtension* tmp = fFWExtension;
+   TGeoExtension *tmp = fFWExtension;
```
dpiparo, line 648: what is the goal of this line?

geom/geom/src/TGeoPcon.cxx
```diff
@@ -1520,6 +1519,168 @@ Int_t TGeoPcon::GetNmeshVertices() const
...
+{
+   if (!array || npoints <= 0)
+      return kFALSE;
+
+   if (fNz < 2)
+      return kFALSE;
+
+   // Angular span (degrees). In SetPoints it is fDphi/(n-1) with n = Nsegments+1.
+   // Here we sample generator centers, so we use nGen generators and step = fDphi/nGen.
+   const Double_t phi1Deg = fPhi1;
```
dpiparo, line 1537: should this line be moved after the check below?

geom/geom/src/TGeoPcon.cxx
```diff
@@ -1520,6 +1519,168 @@ Int_t TGeoPcon::GetNmeshVertices() const
...
+
+   Double_t sumLen = 0.0;
+   for (Int_t i = 0; i < fNz - 1; ++i) {
+      const Double_t z0 = fZ[i];
+      const Double_t z1 = fZ[i + 1];
+      const Double_t dz = z1 - z0;
+      if (dz == 0.0)
+         continue;
+      const Double_t len = std::abs(dz);
+      zsegs.push_back({i, z0, z1, len});
```
dpiparo, line 1561: emplace?

geom/geom/src/TGeoPcon.cxx
```diff
@@ -1520,6 +1519,168 @@ Int_t TGeoPcon::GetNmeshVertices() const
...
+         const Double_t dphi = dPhiDeg / (Double_t)nGen;
+
+         for (Int_t ig = 0; ig < nGen; ++ig) {
+            const Int_t m = q + (ig < rem ? 1 : 0);
+            if (m <= 0)
+               continue;
+
+            // Generator center angle: avoids phi endpoints
+            const Double_t phi = (phi1Deg + (ig + 0.5) * dphi) * TMath::DegToRad();
+            const Double_t c = TMath::Cos(phi);
```
dpiparo, line 1646: sincos opportunity?

geom/geom/src/TGeoTube.cxx
```diff
@@ -2350,38 +2336,89 @@ void TGeoTubeSeg::SetDimensions(Double_t *param)
...
+   if (!(dPhiTot > 0.0))
+      return kFALSE;
+
+   const Bool_t hasInner = (fRmin > 0.0);
+
+   // Split budget: outer gets the extra point if odd
+   Int_t outPoints = npoints;
+   Int_t inPoints = 0;
+   if (hasInner) {
+      outPoints = (npoints + 1) / 2;
```
dpiparo, line 2360: opportunity for one division less?

geom/geom/src/TGeoTube.cxx
```diff
@@ -2350,38 +2336,89 @@ void TGeoTubeSeg::SetDimensions(Double_t *param)
...
+      if (nGen < 1)
+         nGen = 1;
+      if (nGen > nSurfPoints)
+         nGen = nSurfPoints;
+
+      const Int_t q = nSurfPoints / nGen;
+      const Int_t rem = nSurfPoints % nGen;
+
+      // Use generator centers: never exactly phi1 or phi2
+      const Double_t dphi = dPhiTot / (Double_t)nGen;
```
dpiparo, line 2381: no need to cast to double

geom/geom/src/TGeoTube.cxx
```diff
@@ -2350,38 +2336,89 @@ void TGeoTubeSeg::SetDimensions(Double_t *param)
...
+
+      for (Int_t ig = 0; ig < nGen; ++ig) {
+         const Int_t m = q + (ig < rem ? 1 : 0);
+         if (m <= 0)
+            continue;
+
+         const Double_t phi = phi1 + (ig + 0.5) * dphi; // avoid endpoints
+         const Double_t c = TMath::Cos(phi);
```
dpiparo, line 2389: sincos

geom/geom/src/TGeoTube.cxx
```diff
@@ -3252,9 +3288,103 @@ void TGeoCtub::SetDimensions(Double_t *param)
...
+         const Double_t s = TMath::Sin(phi);
```
dpiparo, line 3345: sincos

geom/geom/src/TGeoVolume.cxx
```diff
@@ -380,16 +380,20 @@ volumes (or volume assemblies) as content.
...
-#include "TBuffer.h"
-#include "TBrowser.h"
-#include "TStyle.h"
-#include "TH2F.h"
-#include "TROOT.h"
-#include "TEnv.h"
-#include "TMap.h"
-#include "TFile.h"
-#include "TKey.h"
+#include <TString.h>
```
dpiparo, line 383: stl and root's include order

geom/geom/src/TGeoVolume.cxx
```diff
@@ -1479,7 +1599,7 @@ void TGeoVolume::SaveAs(const char *filename, Option_t *option) const
 
 void TGeoVolume::SetUserExtension(TGeoExtension *ext)
 {
-   TGeoExtension* tmp = fUserExtension;
+   TGeoExtension *tmp = fUserExtension;
```
dpiparo, line 1602: use of this line?

geom/geom/src/TGeoVolume.cxx
```diff
@@ -1497,7 +1617,7 @@ void TGeoVolume::SetUserExtension(TGeoExtension *ext)
 
 void TGeoVolume::SetFWExtension(TGeoExtension *ext)
 {
-   TGeoExtension* tmp = fFWExtension;
+   TGeoExtension *tmp = fFWExtension;
```
dpiparo, line 1620: use of this line?

geom/geomchecker/inc/TGeoChecker.h
```diff
@@ -13,6 +13,10 @@
 #define ROOT_TGeoChecker
 
 #include "TVirtualGeoChecker.h"
+#include "TGeoOverlapCandidate.h"
```
dpiparo, line 16: stl - ROOT includes order

geom/geomchecker/src/TGeoChecker.cxx
```diff
@@ -1418,56 +1510,162 @@ Int_t TGeoChecker::NChecksPerVolume(TGeoVolume *vol)
...
+
+   // Strict mode: caller forgot to include this shape in BuildMeshPointsCache
+   ::Error("TGeoChecker::GetMeshPoints",
+           "Mesh cache miss for shape %s. BuildMeshPointsCache did not cover all candidates.", shape->ClassName());
+   return kEmpty;
+}
+
+void TGeoChecker::BuildMeshPointsCache(const std::vector<TGeoOverlapCandidate> &candidates)
+{
+   std::unique_lock lock(fMeshCacheMutex);
```
dpiparo, line 1532: Just triple checking here: is this a mutex different from ROOT's global mutex? If yes, are we dead sure we cannot trigger a deadlock?

geom/geomchecker/src/TGeoChecker.cxx
```diff
@@ -1418,56 +1510,162 @@ Int_t TGeoChecker::NChecksPerVolume(TGeoVolume *vol)
...
-   // first, test if daughters extrude their container
-   TGeoNode *node, *nodecheck;
-   TGeoChecker *checker = (TGeoChecker *)this;
+   fFullCheck = opt.Contains("f");
 
-   //   TGeoOverlap *nodeovlp = 0;
-   UInt_t id;
-   Int_t level;
-   // Check extrusion only for daughters of a non-assembly volume
+   Int_t ncand = 0;
```
dpiparo, line 1649: please solve these warnings

geom/test/test_boolean_extrusion.cxx
```diff
@@ -1,6 +1,7 @@
 #include <gtest/gtest.h>
```
dpiparo, line 1: is this test enough? Maybe there are more cases to check and that would be very useful!

geom/geom/inc/TGeoManager.h
agheata: This is just one of the whitespace modifications done by vscode, that I should have avoided. Just an older checking function that I have not touched.

geom/geom/inc/TGeoTube.h
agheata: Agreed, I wanted to inline and did not pay attention - there are many things like this in TGeo, but this requires a much deeper cleanup.

geom/geom/inc/TGeoBBox.h
agheata: Sure, I did not know, I should rather change this fast because it is a new interface in TGeoBBox...

agheata: Done in https://github.com/root-project/root/pull/21057

geom/geom/inc/TGeoTube.h
agheata: It is fine, fPhi1 and fPhi2 are checked and normalized in `TGeoTubeSeg::SetTubsDimensions`

geom/geom/inc/TVirtualGeoChecker.h
agheata: Yes, thanks for reminding me the "local to global" order rule, I will enforce it from now on - currently most of the geom files are not following it

geom/geom/inc/TVirtualGeoChecker.h
agheata: We could, but most of these are legacy interfaces (or helpers of those), and I'm reluctant to change them. I hear what you are saying and totally agree in general.

geom/geom/inc/TVirtualGeoChecker.h
agheata: True, but I plan to do a bigger effort and doxygenify all geom headers at some point. For these particular ones it is not urgent, they are mainly implementation helpers, the pretentious pure virtuality is there mainly because of the geom<->painter/checker library insulation

geom/geom/src/TGeoBBox.cxx
agheata: It is not, but I agree it can be optimized

geom/geom/src/TGeoCone.cxx
agheata: Yes, there are many places where this can be done

agheata: I wonder why sincos is not available in TMath?

geom/geom/src/TGeoNode.cxx
agheata: TBD in a much larger PR doing also clang-format for all geom classes

geom/geom/src/TGeoNode.cxx
agheata: Just clang-format for the full file

geom/geom/src/TGeoPcon.cxx
agheata: Not really needed, with any decent optimization, the compiler will eliminate those loads as unused values via dead code elimination

geom/geom/src/TGeoPcon.cxx
agheata: Agreed in general for complex/heavy constructors, but nowadays for trivial structs `push_back` and `emplace_back` generate the same code, the compiler moves the data from the temporary to the vector memory (POD copy), inlining and eliding aggressively, so no temporaries survive codegen

geom/geom/src/TGeoTube.cxx
agheata: you mean: `outPoints = (npoints + 1)  >> 1` ? If yes, I was doing this agressively at some point, until I discovered that it is worse: it hides the intention (looks more like a hack, not math), on signed integers it is wrong, and the compiler eventually generates the same code. So `n / 2` lets the compiler take the decision on the implementation

geom/geom/src/TGeoTube.cxx
agheata: Agreed, I'm a bit maniac due to the work on mixed precision in VecGeom. Harmless here.

geom/geom/src/TGeoTube.cxx
agheata: The problem is that `sincos` is a non-standard extension that exists on most platforms, but not on all (such as mac/windows flavors), so IMHO `TMath` should (it does not) provide a safe wrapper for it that falls back on regilar Sin/Cos where not available). Or maybe Iam wrong?

geom/geom/src/TGeoVolume.cxx
agheata: formatting only

geom/geomchecker/src/TGeoChecker.cxx
agheata: of course, this is a private mutex

geom/geomchecker/src/TGeoChecker.cxx
agheata: Did I miss some warnings?

geom/test/test_boolean_extrusion.cxx
agheata: Indeed, we can have a more extended check, but the thing with overlap checking is that the results may change if for some reason, the points generated on the shape meshes change. Overlap checking only strives not to give false positives, but may miss. So this case is exactly doing this: knows that the shapes mother/daughter have some common faces, but this is correct, so no extrusion should be reported. Actually this unit test triggered the error on Windows, that was a real bug, and also other errors when changing some global static and having 2 geometry managers in the same process.

PR #20966, opened 2026-01-20
[CMake] Introduce ROOT_PYTHON_PACKAGE function

guitargeek:
As anticipated in the commit message of 656ceb83f0370, the CMake code that is used to copy Python sources to the build directory, compile to byte code, and do the install is consolidated in a ROOT_PYTHON_PACKAGE CMake function. This function is then used for all the four Python packages that the ship: `cppyy`, `ROOT`, `JupyROOT`, and `DistRDF`.

Three little improvements are done in the new macro, compared to the prototype hardcoded for `cppyy` and `ROOT`:

  * Use whitelist pattern instead of blacklist pattern in install step,
    to ensure no unwanted artifacts are installed (or files built by
    other targets, like the shared libraries)

  * Register the byte code directories to be cleaned up by `make clean`
    or `ninja clean`

  * There is just one target that depends on the final stamp file,
    implying the execution of the copy and the compile commands in
    succession

This fixes remaining incremental build problems with `JupyROOT` and `DistRDF`.


cmake/modules/RootMacros.cmake
```diff
@@ -1493,6 +1493,83 @@ function(REFLEX_BUILD_DICTIONARY dictionary headerfiles selectionfile )
...
+  # Ensure output directory exists
+  file(MAKE_DIRECTORY ${localruntimedir}/${pkgname})
+
+  set(copy_commands)
+  set(py_sources_in_source_dir)
+  set(py_sources_in_localruntimedir)
+  set(bytecode_dirs)
+
+  foreach(py_source ${ARG_SOURCES})
+    set(src ${CMAKE_CURRENT_SOURCE_DIR}/python/${py_source})
```
hageboeck: I would suggest configuring it with
```
SOURCES
  python/ABC.py
  python/DEF.py
)
```

```suggestion
    set(src ${CMAKE_CURRENT_SOURCE_DIR}/${py_source})
```

cmake/modules/RootMacros.cmake
```diff
@@ -1493,6 +1493,83 @@ function(REFLEX_BUILD_DICTIONARY dictionary headerfiles selectionfile )
...
+  )
+
+  add_custom_target(${pkgname}PySources ALL DEPENDS ${py_sources_in_localruntimedir})
+
+  # Compile .py files
+
+  # Stamp file so CMake knows it doesn't need to re-compile. We can't set the
+  # actual bytecode files as the OUTPUT of the custom command, because their
+  # names are CPython implementation dependent and therefore not reliable.
+  set(bytecode_stamp ${localruntimedir}/${pkgname}/pybytecode.stamp)
```
hageboeck, line 1546: The folder `${localruntimedir}/${pkgname}` might not exist yet. Will it be created during `py_compile`?

cmake/modules/RootMacros.cmake
```diff
@@ -1493,6 +1493,83 @@ function(REFLEX_BUILD_DICTIONARY dictionary headerfiles selectionfile )
...
+  list(REMOVE_DUPLICATES bytecode_dirs)
+
+  add_custom_command(
+    OUTPUT ${py_sources_in_localruntimedir}
+    ${copy_commands}
+    DEPENDS ${py_sources_in_source_dir}
+    COMMENT "Copying ${pkgname} Python sources"
+  )
+
+  add_custom_target(${pkgname}PySources ALL DEPENDS ${py_sources_in_localruntimedir})
```
hageboeck: This might not be needed. Naming the generated files as dependency of the command following below should be sufficient.

```suggestion
```

cmake/modules/RootMacros.cmake
```diff
@@ -1493,6 +1493,83 @@ function(REFLEX_BUILD_DICTIONARY dictionary headerfiles selectionfile )
...
+    OUTPUT ${bytecode_stamp}
+    COMMAND ${Python3_EXECUTABLE} -m py_compile ${py_sources_in_localruntimedir}
+    COMMAND ${Python3_EXECUTABLE} -O -m py_compile ${py_sources_in_localruntimedir}
+    COMMAND ${CMAKE_COMMAND} -E touch ${bytecode_stamp}
+    DEPENDS ${py_sources_in_localruntimedir}
+    COMMENT "Compiling ${pkgname} Python sources"
+  )
+
+  add_custom_target(${pkgname}PyBytecode ALL DEPENDS ${bytecode_stamp})
+  add_dependencies(${pkgname}PyBytecode ${pkgname}PySources)
```
hageboeck: I think this won't be needed either, since the dependencies go as:
stamp --> all python files in runtimedir --> all python files in source dir

In the end, there should be only one target per package, and this target consists of a succession of two custom commands.

```suggestion
```

cmake/modules/RootMacros.cmake
guitargeek: It is created at the top of the function with `file(MAKE_DIRECTORY ${localruntimedir}/${pkgname})`

cmake/modules/RootMacros.cmake
guitargeek: But then it's not possible anymore to easily figure out the subdirectory structure in the build tree for the `ADDITIONAL_CLEAN_FILES` :slightly_frowning_face:

What I can also suggest is to move the `CMakeLists.txt` files themselves into the `python/` subdirectories, like how it's done with `test/` already. Then we don't need to make assumptions on subdirectories.

cmake/modules/RootMacros.cmake
guitargeek: Perfect! So in the end there is just one target that depends on the stamp and done! Thanks

cmake/modules/RootMacros.cmake
```diff
@@ -1493,6 +1493,80 @@ function(REFLEX_BUILD_DICTIONARY dictionary headerfiles selectionfile )
   install(CODE "EXECUTE_PROCESS(COMMAND ${merge_rootmap_cmd} --do-merge --input-file ${srcRootMap} --merged-file ${mergedRootMap})")
 endfunction()
 
+#---------------------------------------------------------------------------------------------------
+#---ROOT_PYTHON_PACKAGE(pkgname
```
hageboeck, line 1497: Maybe add one line about what it does here.

e.g. copy python scripts and precompile them.

hageboeck (approved):
LGTM!

PR #20976, opened 2026-01-21
[CPyCppyy] Correctly propagate template argument name errors

guitargeek:
In case the `Utility::ConstructTemplateArgs()` method fails to create the C++ template names, it sets a useful Python exception that is unfortunately not propagated, because the calling code in `TemplateProxy.cxx` is ignoring it. This is fixed in this commit.

The place where the error is set is also moved from `Utility::ConstructTemplateArgs()` to the actual implementation: `AddTypeName()`. This means the error is set by the function that actually has the context of the error, which is conceptually cleaner and will allow us also to set more detailed errors for different failure modes of `AddTypeName()`.

Demo code:
```python
import cppyy

cppyy.cppdef("""

template<class T>
class Foo {};

template<class T, class Y=void>
auto make_foo () { return Foo<T>{}; }

template<class T, class Y=void>
auto make_foo (T x) { return Foo<T>{}; }

""")

cppyy.gbl.make_foo[set(), list()]()

```

Before:
```
    cppyy.gbl.make_foo[set(), list()]()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
TypeError: Could not find "make_foo" (set cppyy.set_debug() for C++ errors):
  Template method resolution failed:
  Failed to instantiate "make_foo()"
  Failed to instantiate "make_foo()"
```

After:
```
    cppyy.gbl.make_foo[set(), list()]()
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
TypeError: could not construct C++ name from template argument set()
```

Note: this work is orthogonal to the rebase of `cppyy` on CppInterOp.

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/TemplateProxy.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/TemplateProxy.cxx
@@ -173,6 +173,13 @@ PyObject* TemplateProxy::Instantiate(const std::string& fname,
 
         Py_DECREF(pyargs);
         Py_DECREF(tpArgs);
+
+    // Propagate the error that occurs if we can't construct the C++ name
+    // from the provided template argument
+        if (PyErr_Occurred()) {
+            return nullptr;
+        }
+
         if (name_v1.size())
             proto = name_v1.substr(1, name_v1.size()-2);
     }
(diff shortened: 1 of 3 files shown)


vepadulano (approved):
LGTM, feel free to merge once the CI is green

PR #20977, opened 2026-01-21
[cppyy] Avoid JIT-compiling make_unique/make_shared wrappers for builtins

### LINKED ISSUE #19122: [cppyy] C++11 tests fail on macOS26-Beta+XCode26 Beta1
### Check duplicate issues.

- [x] Checked for duplicates

### Description

A subset of these tests fail when calling `std::make_unique` (`test15_unique_ptr_template_deduction ` and `test16_unique_ptr_moves `). This stems from incorrect type info collected in TClingCallFunc, that goes through `cling::utils::Transform::GetPartiallyDesugaredType`, which explicitly adds a 0 to the arg pack of `make_unique` during codegen, which is illegal:

```
2025-06-21T02:05:14.3073430Z ../../../../../../src/bindings/pyroot/cppyy/cppyy/test/test_cpp11features.py::TestCPP11FEATURES::test15_unique_ptr_template_deduction input_line_132:7:124: error: address of overloaded function 'make_unique' does not match required type 'std::unique_ptr<int, std::default_delete<int> > ()'
2025-06-21T02:05:14.3074880Z       new (ret) (std::unique_ptr<int, std::default_delete<int> >) (((std::unique_ptr<int, std::default_delete<int> > (&)())std::__1::make_unique<int, 0>)());
2025-06-21T02:05:14.3075500Z                                                                                                                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2025-06-21T02:05:14.3076450Z /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/__memory/unique_ptr.h:766:76: note: candidate template ignored: invalid explicitly-specified argument for template parameter '_Args'
2025-06-21T02:05:14.3077580Z inline _LIBCPP_HIDE_FROM_ABI _LIBCPP_CONSTEXPR_SINCE_CXX23 unique_ptr<_Tp> make_unique(_Args&&... __args) {

2025-06-21T02:05:14.3096070Z ../../../../../../src/bindings/pyroot/cppyy/cppyy/test/test_cpp11features.py::TestCPP11FEATURES::test16_unique_ptr_moves input_line_134:7:130: error: address of overloaded function 'make_unique' does not match required type 'std::unique_ptr<int, std::default_delete<int> > (int &&)'
2025-06-21T02:05:14.3097560Z       new (ret) (std::unique_ptr<int, std::default_delete<int> >) (((std::unique_ptr<int, std::default_delete<int> > (&)(int &&))std::__1::make_unique<int, int, 0>)((int&&)*(int*)args[0]));

2025-06-21T02:05:14.3109120Z input_line_134:11:76: error: address of overloaded function 'make_unique' does not match required type 'std::unique_ptr<int, std::default_delete<int> > (int &&)'
2025-06-21T02:05:14.3113010Z       (void)(((std::unique_ptr<int, std::default_delete<int> > (&)(int &&))std::__1::make_unique<int, int, 0>)((int&&)*(int*)args[0]));
```

See https://github.com/root-project/root/pull/19119 and backports.

### Reproducer

Run the `pyunittests-bindings-pyroot-cppyy-cppyy-test-cpp11features`

### ROOT version

macOS26-Beta+XCode26 Beta1

### Installation method

sources

### Operating system

macOS26-Beta+XCode26 Beta1

### Additional context

_No response_

guitargeek:
std::make_unique/make_shared for a fundamental element type forces the interpreter to JIT a call wrapper whose variadic Args... template pack is mis-desugared on some platforms (macOS with C++ modules) into invalid code such as `make_unique<int, 0>`, breaking e.g. make_unique<int>.

Route arithmetic element types through dedicated, non-variadic helpers (__cppyy_internal::make_unique_scalar / make_shared_scalar) that build the smart pointer from a heap-allocated scalar (new T{} / new T(v)), side-stepping the make_unique/make_shared wrapper entirely. The helpers static_assert that the element type is arithmetic, mirroring the Python-side dispatch. Class types keep the existing pure-introspection adopt path, which never instantiated make_unique either.

Arithmetic element types are detected by querying std::is_arithmetic through the interpreter, letting C++ be the source of truth: canonical spellings and typedefs (size_t, Int_t, ...) resolve correctly instead of silently yielding a null smart pointer. is_arithmetic (rather than is_fundamental) also excludes void and std::nullptr_t, which cannot back a heap scalar.

This also fixes make_unique/make_shared for fundamental spellings other than int/float (e.g. 'double', 'long'), which previously errored.

Re-enables the cppyy test15/test16 unit tests that were xfailed on macOS/Windows.

Closes https://github.com/root-project/root/issues/19122

### DIFF
--- a/bindings/pyroot/cppyy/cppyy/python/cppyy/__init__.py
+++ b/bindings/pyroot/cppyy/cppyy/python/cppyy/__init__.py
@@ -145,25 +145,93 @@ class py_make_smartptr(object):
             obj = self.cls(*args)
         return self.ptrcls[self.cls](obj)   # C++ takes ownership
 
+# The introspection trick above only works for class types the smart pointer can
+# adopt from a Python-owned object. An arithmetic element type (int, double, ...)
+# has no such object, so it needs a heap-allocated scalar. We test for *arithmetic*
+# rather than *fundamental* types because std::is_fundamental also matches 'void'
+# and 'std::nullptr_t', neither of which can back a 'new T{}' scalar.
+#
+# _py_arithmetic maps the Python numeric builtins onto a C++ spelling by convention
+# (Python's int is arbitrary-precision, bool subclasses int); they carry no
+# __cpp_name__ to query, so the mapping has to be explicit.
+_py_arithmetic = {int: 'int', float: 'float', bool: 'bool'}
+
+def _as_arithmetic(cls):
+    """Return the C++ type name if 'cls' denotes an arithmetic type, else None.
+
+    'cls' may be a Python builtin type (int/float/bool), a string naming a C++
+    type, or a cppyy type proxy (e.g. cppyy.gbl.int)."""
+    name = _py_arithmetic.get(cls)
+    if name is not None:
+        return name
+    name = cls if isinstance(cls, str) else getattr(cls, '__cpp_name__', None)
+    if name is None:
+        return None
+    # Let C++ be the source of truth: std::is_arithmetic resolves canonical
+    # spellings and typedefs (size_t, Int_t, ...) alike, with no hand-kept list to
+    # drift out of sync. cppyy caches the trait instantiation itself.
+    #
+    # A name that is no resolvable type (e.g. a typo) makes the trait lookup raise;
+    # treat that as "not arithmetic" so it falls through to the class path, which
+    # reports the bad name itself instead of leaking std::is_arithmetic<...>.
+    try:
+        is_arith = bool(gbl.std.is_arithmetic[name].value)
+    except Exception:
+        is_arith = False
+    return name if is_arith else None
+
+# Helpers that build a smart pointer to a heap-allocated scalar. They are
+# deliberately *non-variadic*: instantiating std::make_unique/make_shared for an
+# arithmetic type makes the interpreter JIT a call wrapper whose variadic Args...
+# pack is mis-desugared on some platforms (macOS with C++ modules) into invalid
+# code like 'make_unique<int, 0>'; these single-argument templates avoid it.
+# See https://github.com/root-project/root/issues/19122.
+#
+# Each body static_asserts T is arithmetic, mirroring the Python-side
+# _as_arithmetic() gate (class types take the adopt path instead). The assert is
+# in the body, not the signature: an extra template parameter is exactly what
+# triggers the mis-desugaring above.
+_scalar_makers = {}
+def _scalar_maker(name):
+    maker = _scalar_makers.get(name)
+    if maker is None:
+        if not _scalar_makers:
+            cppdef("""namespace __cppyy_internal {
+            template <class T> std::unique_ptr<T> make_unique_scalar()    { static_assert(std::is_arithmetic_v<T>, "scalar maker requires an arithmetic type"); return std::unique_ptr<T>(new T{}); }
+            template <class T> std::unique_ptr<T> make_unique_scalar(T v) { static_assert(std::is_arithmetic_v<T>, "scalar maker requires an arithmetic type"); return std::unique_ptr<T>(new T(v)); }
+            template <class T> std::shared_ptr<T> make_shared_scalar()    { static_assert(std::is_arithmetic_v<T>, "scalar maker requires an arithmetic type"); return std::shared_ptr<T>(new T{}); }
+            template <class T> std::shared_ptr<T> make_shared_scalar(T v) { static_assert(std::is_arithmetic_v<T>, "scalar maker requires an arithmetic type"); return std::shared_ptr<T>(new T(v)); }
+            }""")
+        maker = getattr(gbl.__cppyy_internal, name)
+        _scalar_makers[name] = maker
+    return maker
+
 class make_smartptr(object):
-    __slots__ = ['ptrcls', 'maker']
-    def __init__(self, ptrcls, maker):
-        self.ptrcls = ptrcls
-        self.maker  = maker
+    __slots__ = ['ptrcls', 'maker', 'scalarname']
+    def __init__(self, ptrcls, maker, scalarname):
+        self.ptrcls     = ptrcls
+        self.maker      = maker
+        self.scalarname = scalarname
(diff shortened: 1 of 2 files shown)


bindings/pyroot/cppyy/cppyy/python/cppyy/__init__.py
```diff
@@ -145,25 +145,88 @@ def __call__(self, *args):
...
 
+# The introspection trick above only works for class types the smart pointer can
+# adopt from a Python-owned object. A fundamental element type (int, double, ...)
+# has no such object: it needs a heap-allocated scalar.
+_cpp_fundamentals = frozenset([
+    'bool', 'char', 'signed char', 'unsigned char', 'wchar_t', 'char16_t', 'char32_t',
+    'short', 'unsigned short', 'int', 'unsigned int', 'unsigned',
+    'long', 'unsigned long', 'long long', 'unsigned long long',
+    'float', 'double', 'long double',
+])
```
vepadulano: For the values of the frozenset I would use the official C++ standard fundamental types, also used e.g. in https://github.com/root-project/root/blob/ba4e9086ade4933457e24775b04ca68c7174bc71/tree/dataframe/src/RDFUtils.cxx#L88

bindings/pyroot/cppyy/cppyy/python/cppyy/__init__.py
```diff
@@ -145,25 +145,88 @@ def __call__(self, *args):
...
+# Python builtin types that map onto a C++ fundamental type.
+_py_fundamentals = {int: 'int', float: 'float', bool: 'bool'}
```
vepadulano: Do these types really map cleanly? Or is it more like our own internal convention? I would add a clarification in the comment here

bindings/pyroot/cppyy/cppyy/python/cppyy/__init__.py
```diff
@@ -145,25 +145,88 @@ def __call__(self, *args):
...
+        return name
+    name = cls if isinstance(cls, str) else getattr(cls, '__cpp_name__', None)
+    if name is None:
+        return None
+    if name in _cpp_fundamentals:            # canonical spelling, no reflection needed
+        return name
+    is_fund = _fundamental_cache.get(name)   # else resolve typedefs (size_t, Int_t, ...)
+    if is_fund is None:
+        try:
+            is_fund = bool(gbl.std.is_fundamental[name].value)
```
vepadulano: `std::is_fundamental` includes `void` and `std::nullptr_t`. Do we want those included, or perhaps would it be better to use `std::is_arithmetic`?

bindings/pyroot/cppyy/cppyy/python/cppyy/__init__.py
```diff
@@ -145,25 +145,88 @@ def __call__(self, *args):
...
+def _scalar_maker(name):
+    maker = _scalar_makers.get(name)
+    if maker is None:
+        if not _scalar_makers:
+            cppdef("""namespace __cppyy_internal {
+            template <class T> std::unique_ptr<T> make_unique_scalar()    { return std::unique_ptr<T>(new T{}); }
+            template <class T> std::unique_ptr<T> make_unique_scalar(T v) { return std::unique_ptr<T>(new T(v)); }
+            template <class T> std::shared_ptr<T> make_shared_scalar()    { return std::shared_ptr<T>(new T{}); }
+            template <class T> std::shared_ptr<T> make_shared_scalar(T v) { return std::shared_ptr<T>(new T(v)); }
+            }""")
```
vepadulano, line 204: Reading this code I am wondering if we couldn't bypass the Python-side checks by constraining the templates of these functions with e.g. `static_assert` on they type being arithmetic. Or making them truly generic with `if constexpr` statements.

vepadulano (changes requested):

bindings/pyroot/cppyy/cppyy/python/cppyy/__init__.py
guitargeek: The `static_assert`s are a good idea, I have implemented that! But they don't need to be truly generic with an `if contexpr`, because they are only used in the case where we don't have arithmetic types.

bindings/pyroot/cppyy/cppyy/python/cppyy/__init__.py
guitargeek: We don't need this frozenset anymore, because thanks to your comment I remembered `std::is_arithmetic` is an option here, so the new version of the PR just uses that.

Note that this implies a template instantiation, but it's fine IMHO. `std::make_unique` and `std::make_shared` from Python are far from the hot path :slightly_smiling_face:

bindings/pyroot/cppyy/cppyy/python/cppyy/__init__.py
guitargeek: Very good point! Moved to `std::is_arithmetic`.

bindings/pyroot/cppyy/cppyy/python/cppyy/__init__.py
guitargeek: No, they don't map cleanly, since the Python int is arbitrary precision and Python `float` is 64 bit. This is all convention set by the rest of cppyy. I have added a comment to note this.

By the way, I'm not exactly happy with this convention, but changing it would require wider discussion that is outside the mandate of this PR.

guitargeek:
@vepadulano Thanks for the excellent suggestions! I have rewritten the logic around `std::is_arithmetic`, let's see what the CI says!

guitargeek:
Failures like the one of `gtest-tree-ntuple-ntuple-attributes` on mac15 are unrelated.

vepadulano (approved):
Clever solution given the constraints! Ideally I would have preferred a C++ only solution for all types using `if constexpr` to differentiate between arithmetic and not. But I see the non-arithmetic types need Python-side treatment anyway, so it's probably out of scope for this PR. Thanks for the effort!

PR #20978, opened 2026-01-21
[RF] Prepare RooFit for ROOT without implicit ownership.

hageboeck:
Also add an assertion to stop the test when executed in the wrong directory. Previously, the test would crash.

Repeat PR #20889, which couldn't be merged due to a github bug.


roofit/roofitcore/test/testRooDataSet.cxx
```diff
@@ -439,6 +440,7 @@ TEST(RooDataSet, SplitDataSetWithWeightErrors)
 TEST(RooDataSet, ReadDataSetWithErrors626)
 {
    std::unique_ptr<TFile> file{TFile::Open("dataSet_with_errors_6_26_10.root", "READ")};
+   ASSERT_NE(file, nullptr);
```
pcanal:
```suggestion
   ASSERT_TRUE(file && !file->IsZombie());
```

PR #20980, opened 2026-01-21
[ci] Don't override C++ standard for modules=off PR builds

guitargeek:
Trial commit to see if we can reproduce the failures in the nightlies also in the CI.


dpiparo:
just for info, one can specify the standard in the build config file with a line like `CMAKE_CXX_STANDARD=20` if needed

guitargeek:
Not needed anymore. This was just a one-time check.

PR #20983, opened 2026-01-21
[hist] Introduce dispatching `RAxisVariant` class

hahnjo:
It encapsulates the `std::variant` of axis types and can provide dispatching methods for common accessors.

### DIFF
--- a/hist/histv7/inc/ROOT/RAxisVariant.hxx
+++ b/hist/histv7/inc/ROOT/RAxisVariant.hxx
@@ -9,6 +9,7 @@
 #include "RRegularAxis.hxx"
 #include "RVariableBinAxis.hxx"
 
+#include <cstdint>
 #include <stdexcept>
 #include <utility>
 #include <variant>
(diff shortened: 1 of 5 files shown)


hist/histv7/inc/ROOT/RAxisVariant.hxx
```diff
@@ -0,0 +1,60 @@
...
+   using VariantType = std::variant<RRegularAxis, RVariableBinAxis, RCategoricalAxis>;
+
+private:
+   VariantType fVariant;
+
+public:
+   RAxisVariant(VariantType axis) : fVariant(std::move(axis)) {}
+   RAxisVariant(RRegularAxis axis) : fVariant(std::move(axis)) {}
+   RAxisVariant(RVariableBinAxis axis) : fVariant(std::move(axis)) {}
+   RAxisVariant(RCategoricalAxis axis) : fVariant(std::move(axis)) {}
```
hageboeck, line 40: Would it make sense to use a single template? If a wrong type is passed, the variant can't be constructed.

hahnjo: My hope was that non-templated constructors would make the job easier for cppyy and the Python interface. Unfortunately that doesn't seem to work out-of-the-box, but I think I still prefer explicitly stating which argument types are accepted over a template...

PR #20984, opened 2026-01-22
[hist] Implement conversion of `RVariableBinAxis` and `RCategoricalAxis`

hahnjo:


hist/histv7util/src/ConversionUtils.cxx
```diff
@@ -0,0 +1,28 @@
...
+#include <ROOT/Hist/ConversionUtils.hxx>
+#include <ROOT/RAxisVariant.hxx>
+#include <ROOT/RRegularAxis.hxx>
+
+#include <TAxis.h>
+
+namespace ROOT {
+namespace Experimental {
+namespace Hist {
+namespace Internal {
```
hageboeck, line 16: `ROOT::Experimental::[...]` ?

hist/histv7util/src/ConversionUtils.cxx
```diff
@@ -0,0 +1,28 @@
...
+
+void ConvertAxis(TAxis &dst, const RAxisVariant &src)
+{
+   if (auto *regular = src.GetRegularAxis()) {
+      dst.Set(regular->GetNNormalBins(), regular->GetLow(), regular->GetHigh());
+   } else {
+      throw std::logic_error("unimplemented axis type"); // GCOVR_EXCL_LINE
```
hageboeck: It's hard to trigger this exception, but maybe the message could  give more context.

hageboeck (approved):

hist/histv7util/src/ConversionUtils.cxx
hahnjo: I'm not a fan of this C++17 syntax... Having separate `namespace` lines also makes it easier to move out of `Experimental`, once we are at that point.

PR #20985, opened 2026-01-22
[ci] Explicitly set Ruff version latest to avoid version warning

siliataider:
# This Pull request:

Explicitly sets the Ruff version in CI to avoid the warning:
`Warning: Could not parse version from …/pyproject.toml. Using latest version.`

### DIFF
--- a/.github/workflows/code_analysis.yml
+++ b/.github/workflows/code_analysis.yml
@@ -69,7 +69,7 @@ jobs:
 
     - name: Lint code
       run: |
-        files=$(cat changed_files.txt | grep '\.py$' || echo "")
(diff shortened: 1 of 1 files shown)


.github/workflows/code_analysis.yml
hageboeck:
```suggestion
        files=$(grep '\.py$' changed_files.txt || echo "")
```

hageboeck:
```suggestion
        files=$(grep '\.py$' changed_files.txt || echo "")
```

hageboeck (approved):
LGTM!

Maybe we can also make shellcheck happy in a second commit?

PR #20986, opened 2026-01-22
[ntuple] Throw instead of asserting in RMiniFileReader::ReadBuffer

silverweed:
Corrupted files may cause `ReadBuffer` to fail, so the proper way of handling it is by returning a `RResult` rather than asserting.
In the interest of not changing the many places using `ReadBuffer`, I made it wrap a `TryReadBuffer` function that returns `RResult` and simply call `ThrowOnError` on it.
This gives similar semantics as before, except `ReadBuffer` will throw rather than asserting.

### DIFF
--- a/tree/ntuple/src/RMiniFile.cxx
+++ b/tree/ntuple/src/RMiniFile.cxx
@@ -887,6 +887,16 @@ ROOT::RResult<ROOT::RNTuple> ROOT::Internal::RMiniFileReader::GetNTupleBare(std:
 
 void ROOT::Internal::RMiniFileReader::ReadBuffer(void *buffer, size_t nbytes, std::uint64_t offset)
 {
+   TryReadBuffer(buffer, nbytes, offset).ThrowOnError();
+}
+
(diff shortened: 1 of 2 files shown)


tree/ntuple/src/RMiniFile.cxx
```diff
@@ -900,7 +910,9 @@ void ROOT::Internal::RMiniFileReader::ReadBuffer(void *buffer, size_t nbytes, st
 
       // Read first chunk
       nread = fRawFile->ReadAt(bufCur, fMaxKeySize, offset);
-      R__ASSERT(nread == fMaxKeySize);
+      if (R__unlikely(nread != fMaxKeySize))
```
jblomer: I think we should use `R__(un)likley` only on performance critical paths.
```suggestion
      if (nread != fMaxKeySize)
```

silverweed: I used it as a "safe default" since it's also there in `R__ASSERT`, but it's probably not really impactful in this case.

PR #20987, opened 2026-01-22
[cmake][win] Fix compilation errors with /std:c++20

bellenot:


cmake/modules/RootMacros.cmake
```diff
@@ -941,6 +941,9 @@ function(ROOT_LINKER_LIBRARY library)
     endif()
   endif()
   set_target_properties(${library} PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
+  if(MSVC)
+    target_compile_options(${library} PRIVATE -Zc:__cplusplus -std:c++${CMAKE_CXX_STANDARD})
```
pcanal: Why is `CMake` not passing those option based on the value of `${CMAKE_CXX_STANDARD}`?

bellenot: > Why is `CMake` not passing those option based on the value of `${CMAKE_CXX_STANDARD}`?

Right, thanks, I think I found the origin of the problem...

roottest/CMakeLists.txt
```diff
@@ -3,7 +3,6 @@ if(MSVC)
   set(CMAKE_SUPPRESS_REGENERATION TRUE)
   set(libsuffix .dll)
   set(exesuffix .exe)
-  set(CMAKE_CXX_FLAGS "/nologo /Zc:__cplusplus /MD /GR /EHsc- /W3 /D_WIN32")
```
pcanal, line 6: Could you extend the git log to explain why this is no longer needed?

bellenot: Done

PR #20988, opened 2026-01-22
[hist] Extend the region in which TH3::Interpolate can be used

### LINKED ISSUE #10678: TH3::Interpolate min/max value condition behaves differently for TH1/TH2
### Describe the bug
Calling Interpolate(x, y, z) with three values within axis ranges yields a "cannot interpolate outside histogram domain" error. 

### Expected behavior
I expected to get a value different than 0 and no error. 

### To Reproduce
```c++
#include <iostream> 
#include "TH3.h"

int main() {
    auto m_hist = new TH3F("3D", "boring histo", 50, 0, 1000, 50, 0, 1000, 50, 0, 1000);
    
    for(int i=1; i <= m_hist->GetNbinsX(); i++)
        for(int j=1; j <= m_hist->GetNbinsY(); j++)
            for(int k=1; k <= m_hist->GetNbinsZ(); k++)
                m_hist->SetBinContent(i, j, k, i + j + k);

    double x = 575.474;
    double y = 0.002;
    double z = 0.002;

    std::cout << "bins = " << m_hist->GetXaxis()->FindBin(x) << " " << m_hist->GetYaxis()->FindBin(y) << " " << m_hist->GetZaxis()->FindBin(z) << "\n"; 
    std::cout << "value = " << m_hist->Interpolate(x, y, z) << "\n";
    std::cout << "value = " << m_hist->Interpolate(10, 10, 10) << "\n";
    
    return 0;
}
```
This outputs the following on my machine,
```
bins = 29 1 1
Error in <UnknownClass::Interpolate>: Cannot interpolate outside histogram domain.
value = 0
value = 3
```

### Setup
Built for linuxx8664gcc on Apr 12 2022, 16:28:03
From tags/v6-26-02@v6-26-02
downloaded the ROOT pre-compiled binaries from your website.

hageboeck:
When a user asks to interpolate a value close to the edges of an axis,
the function would generate an error. This is inconsistent with the TH1
version of Interpolate.
Here, the valid range is extended by assuming that the histogram is
constant at the axis limit (i.e. the imaginary next bin outside the axis
limits has the same bin content as the last bin inside the limits). From
here, the trilinear interpolation proceeds as usual.

Fix #10678

### DIFF
--- a/hist/hist/test/test_TH3.cxx
+++ b/hist/hist/test/test_TH3.cxx
@@ -74,3 +74,25 @@ TEST(TH3D, FillAtomicNoSumW2)
 }
 
 #endif
+
+// ROOT-10678
+TEST(TH3D, InterpolateCloseToEdge)
+{
+   TH3F hist("3D", "boring histo", 20, 0, 20, 20, 0, 20, 20, 0, 20);
+
+   for (int i = 0; i < hist.GetNbinsX(); i++)
+      for (int j = 0; j < hist.GetNbinsY(); j++)
+         for (int k = 0; k < hist.GetNbinsZ(); k++)
+            hist.SetBinContent(i + 1, j + 1, k + 1, i + 100. * j + 10000. * k);
+
+   EXPECT_DOUBLE_EQ(hist.Interpolate(hist.GetXaxis()->GetBinCenter(2), hist.GetYaxis()->GetBinCenter(3),
+                                     hist.GetZaxis()->GetBinCenter(4)),
+                    1. + 100. * 2. + 10000. * 3);
+   EXPECT_DOUBLE_EQ(hist.Interpolate(hist.GetXaxis()->GetBinCenter(2) + 0.5, hist.GetYaxis()->GetBinCenter(3) + 0.4,
+                                     hist.GetZaxis()->GetBinCenter(4) + 0.3),
+                    1. + 0.5 + 100. * 2.4 + 10000. * 3.3);
+
+   EXPECT_DOUBLE_EQ(hist.Interpolate(0., 0., 5.), 10000. * 4.5);
+   EXPECT_DOUBLE_EQ(hist.Interpolate(0.3, 19.9, 5.), 100. * 19 + 10000. * 4.5);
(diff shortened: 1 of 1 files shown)


hist/hist/src/TH3.cxx
```diff
@@ -1496,29 +1496,38 @@ Double_t TH3::Interpolate(Double_t x, Double_t y, Double_t z) const
    if ( z < fZaxis.GetBinCenter(ubz) ) ubz -= 1;
    Int_t obz = ubz + 1;
 
+   auto GetBinCenterExtrapolate = [](TAxis const &axis, Int_t index) {
+      if (index == 0) {
+         return axis.GetBinCenter(1) - axis.GetBinWidth(1);
+      } else if (index == axis.GetNbins() + 1) {
+         return axis.GetBinCenter(axis.GetNbins() + 1) + axis.GetBinWidth(axis.GetNbins() + 1);
```
guitargeek: I got a bit tripped up here thinking about if it's okay to "mirror" the bin centers like this, but then I realized it doesn't matter because you are using it to calculate this `xd` variable, which is used to to morph between `w1` and `w2`, which are the same anyway if you are close to the boundary and the histogram is assumed to be constant beyond the limits.

Maybe consider adding a comment that the `- axis.GetBinWidth(1)` and `+ axis.GetBinWidth(axis.GetNbins() + 1)` are arbitrary, but necessary to avoid division by zero in the calculation of `xd`, `yd`, and `zd` (as the code is now)?

hageboeck: Actually, I think you found a bug here. It should read `GetBinCenter( GetNbins() )`, because otherwise it would be the (*not defined*) centre of the overflow bin. It just happened to work because the regular axis ignored the limits.

Fixed now, and comment added as well.

guitargeek: Right! It's better if the number we make up as the imaginary bin center is at least well defined. Thanks for updating the code and adding a comment. I'm fine with that now.

ferdymercury:
Thanks!!
"funny  note", JIRA thinks you fixed issue https://its.cern.ch/jira/projects/ROOT/issues/ROOT-10678

PR #20991, opened 2026-01-22
[build] Unify retrieval of MacOS SDK path

vepadulano:
Use `xcrun --show-sdk-path` only once to set the CMAKE_OSX_SYSROOT variable and then reuse it also for the modulemap fix on Darwin.

Since the modulemap fix needed the `xcrun` command to work anyway, turn the warning previously issued if CMake couldn't find this command to an error.

### DIFF
--- a/cmake/scripts/std_modulemap_darwin_fix.py
+++ b/cmake/scripts/std_modulemap_darwin_fix.py
@@ -4,14 +4,6 @@ import re
 import subprocess
 import sys
 
-
-def get_sdk_path():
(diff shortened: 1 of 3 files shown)


devajithvs (approved):
LGTM!

cmake/scripts/std_modulemap_darwin_fix.py
hahnjo, line 26: Actually, after this change, should the Conda detection move to the CMake side? It seems the script could be agnostic of the details if we "just" pass `$BUILD_PREFIX`...

PR #20992, opened 2026-01-22
Fix edge case in analytical integral of exponential TF1

swetank18:
Motivation:
The analytical integral for exponential TF1 functions does not handle the
p1 -> 0 limit and may result in division by zero or numerical instability.

Changes:
- Handle the p1 == 0 case using the correct analytical limit
- Preserve existing behavior for non-zero slopes

Impact:
- Improves numerical robustness of TF1 analytical integrals
- No API changes


swetank18:
> Hello, thanks for the contribution!
> 
> The change looks reasonable. Could you run something like `git clang-format HEAD~` on the commit? Otherwise, the CI will fail on the formatting checks.

Thank you for the review.
I’ve applied git clang-format to the commit and pushed the updated changes.

hageboeck:
The work here will be included in #21060

PR #20993, opened 2026-01-22
Add parameter domain checks to analytical integrals

swetank18:
Depends on #20992

Motivation:
Several analytical integrals assume valid parameter domains (e.g. sigma > 0)
without enforcing them, which may lead to undefined numerical behavior.

Changes:
- Add explicit parameter-domain checks for Gaussian, Landau, and Crystal Ball integrals
- Return QuietNaN for invalid parameter configurations

Impact:
- Improves robustness and predictability of analytical integrations
- No API changes


hageboeck:
This work is now part of #21060

PR #20994, opened 2026-01-22
[RNTuple] Implement I/O Performance Metrics: Sparseness, Randomness, and Transactions

### LINKED ISSUE #20853: [ntuple] add several useful performance metrics

There has been a proposal to include the following useful metrics:

1) Sparseness: Total bytes read divided by the total file size.
2) Randomness: Sum of the skip distances, divided by the total bytes read.
3) Transactions: Number of interactions with the I/O layer per file.

Moved from https://its.cern.ch/jira/browse/ROOT-5180

### ROOT version

master

### Installation method

from sources

### Operating system

n/a

### Additional context

_No response_

JasMehta08:
## This Pull request: 
improves the existing Performance Metrics in RNTuple.

## Changes or fixes:
This PR addresses the issue that had been raised for the improvements of metrics. It implements,

* GetSparseness(): Calculates the ratio of explicit payload bytes read to the total file size.
* GetRandomness(): Calculates the ratio of seek distance to bytes read (identifies inefficient access patterns).
* GetTransactions(): Tracks the total number of physical I/O operations (now using a dedicated fTransactions atomic counter).

### Implementation Details: 
These have been implemented through the introduction of transient, thread-safe members in RNTupleMetrics,

* std::atomic<std::uint64_t> fSumSkip: To track total seek distance.

* std::atomic<std::uint64_t> fExplicitBytesRead: To track payload bytes.

* std::atomic<std::uint64_t> fTransactions: To track I/O operations.

I have changed RPageStorageFile to update these counters during ReadBuffer, ReadV, and LoadClusters operations. The use of std::atomic ensures thread safety with negligible performance overhead on the I/O path.

*Reset Functionality & Accuracy:* 
I added a Reset() method to RNTupleMetrics. This is needed for obtaining accurate Randomness metrics for the analysis loop.

* Without Reset(): The metric is dominated by the initial file seek (Header → Footer → Header), resulting in a Randomness score > 2.0.

* With Reset(): Users can clear the counters after initialization, isolating the steady-state performance of their event loop.

### Verification:

Added a new unit test TEST(Metrics, IOMetrics) in tree/ntuple/test/ntuple_metrics.cxx to verify the logic and the Reset() behavior.

Also tested locally using this code 
<details> <summary><b>Click to view Manual Verification Code & Output</b></summary>

```cpp
#include <ROOT/RNTupleMetrics.hxx>
#include <iostream>

int main()
{
   using namespace ROOT::Experimental;
   using namespace ROOT;


   std::string ntupleName = "ntuple";
   std::string fileName = "verify_metrics.root";

   // Create a dummy RNTuple
   std::cout << "[INFO] Creating dummy RNTuple..." << std::endl;
   {
      auto field = model->MakeField<int>("val");
      auto writer = RNTupleWriter::Recreate(std::move(model), ntupleName, fileName);
      for (int i = 0; i < 1000; ++i) {
         *field = i;
         writer->Fill();
      }
   }

   // Read it back
   std::cout << "[INFO] Reading RNTuple and enabling metrics..." << std::endl;
   auto reader = RNTupleReader::Open(ntupleName, fileName);
   reader->EnableMetrics();

   // Read everything to force I/O
   for (auto entry : *reader) {
      (void)entry;
   }

   // Print Metrics
   auto& metrics = reader->GetMetrics();
   
   std::cout << "\n===== RNTuple Performance Metrics Report =====" << std::endl;
   std::cout << "Transactions: " << metrics.GetTransactions() << " (Expected > 0)" << std::endl;
   std::cout << "Sparseness:   " << metrics.GetSparseness() << " (Expected > 0.0)" << std::endl;
   std::cout << "Randomness:   " << metrics.GetRandomness() << " (Expected > 0.0)" << std::endl;
   std::cout << "==============================================\n" << std::endl;

   // Verification Logic
   bool passed = true;
   if (metrics.GetTransactions() <= 0) {
       std::cerr << "[FAIL] Transactions should be > 0" << std::endl;
       passed = false;
   }
   if (metrics.GetSparseness() <= 0.0 || metrics.GetSparseness() > 1.0) {
       std::cerr << "[FAIL] Sparseness should be in (0.0, 1.0]" << std::endl;
       passed = false;
   }
   if (metrics.GetRandomness() <= 0.0) {
       std::cerr << "[FAIL] Randomness should be > 0.0" << std::endl;
       passed = false;
   }

   // Verify Reset functionality
   std::cout << "[INFO] Testing Reset()..." << std::endl;
   reader->GetMetrics().Reset();
   
   // We expect Randomness to go from ~2.13 down to 0.0
   if (metrics.GetSparseness() == 0.0 && metrics.GetRandomness() == 0.0) {
       std::cout << "[SUCCESS] Reset() worked! Randomness dropped to 0.0" << std::endl;
   } else {
       std::cerr << "[FAIL] Reset() failed! Counters are not zero." << std::endl;
       passed = false;
   }
   
   if (passed) {
       std::cout << "[SUCCESS] All metrics logic verified!" << std::endl;
       return 0;
   } else {
       std::cout << "[FAILURE] Some metrics checks failed." << std::endl;
       return 1;
   }
}
```

```bash
[INFO] Reading RNTuple and enabling metrics...

===== RNTuple Performance Metrics Report =====
Transactions: 2 (Expected > 0)
Sparseness:   0.181648 (Expected > 0.0)
Randomness:   2.13746 (Expected > 0.0)
==============================================

[INFO] Testing Reset()...
[SUCCESS] Reset() worked! Randomness dropped to 0.0
[SUCCESS] All metrics logic verified!
```
```cpp
#include <ROOT/RNTupleMetrics.hxx>
#include <iostream>
#include <cstdlib>

int main()
{
   using namespace ROOT::Experimental;
   using namespace ROOT;

   std::string ntupleName = "ntuple";
   std::string fileName = "verify_metrics_v2.root";

   // Create Data (50MB analysis dataset)
   {
      auto fieldPt = model->MakeField<float>("pt");
      auto fieldEta = model->MakeField<float>("eta");
      auto fieldPhi = model->MakeField<float>("phi");

      // Realistic cluster size for analysis (1MB)
      RNTupleWriteOptions options;
      options.SetApproxZippedClusterSize(1024 * 1024); 
      
      auto writer = RNTupleWriter::Recreate(std::move(model), ntupleName, fileName, options);
      
      std::cout << "[INFO] Creating 50MB analysis dataset..." << std::endl;
      // ~4.2M events × 12 bytes/event = ~50MB
      for (int i = 0; i < 4200000; ++i) {
         *fieldPt = 20.0f + static_cast<float>(std::rand() % 100);
         *fieldEta = -2.5f + static_cast<float>(std::rand() % 500) / 100.0f;
         *fieldPhi = -3.14f + static_cast<float>(std::rand() % 628) / 100.0f;
         writer->Fill();
         
         if (i % 500000 == 0) {
            std::cout << "  Written " << (i / 1000) << "k events..." << std::endl;
         }
      }
      std::cout << "[INFO] Dataset created successfully." << std::endl;
   }

   // Open and IMMEDIATE RESET
   std::cout << "[INFO] Opening RNTuple..." << std::endl;
   auto reader = RNTupleReader::Open(ntupleName, fileName);
   reader->EnableMetrics();
   
   // FORCE the first seek (Tail -> Head) so we can clear it
   try {
        // LoadEntry is void but might throw if out of range, 0 is safe here
        reader->LoadEntry(0); 
   } catch (...) {}

   auto& m_pre = reader->GetMetrics();
   std::cout << "\n[DEBUG] Pre-Reset Metrics:" << std::endl;
   std::cout << "  Transactions: " << m_pre.GetTransactions() << std::endl;
   std::cout << "  Randomness:   " << m_pre.GetRandomness() << std::endl;
   std::cout << "  Total File Size: " << m_pre.GetTotalFileSize() << " bytes" << std::endl;
   std::cout << "  N Entries: " << reader->GetNEntries() << std::endl;

   std::cout << "[INFO] Resetting Metrics to isolate analysis phase..." << std::endl;
   reader->GetMetrics().Reset();

   // Analysis Loop (actually access the data)
   std::cout << "[INFO] Running analysis on " << reader->GetNEntries() << " events..." << std::endl;
   
   auto viewPt = reader->GetView<float>("pt");
   auto viewEta = reader->GetView<float>("eta");
   auto viewPhi = reader->GetView<float>("phi");
   
   long selectedEvents = 0;
   double sumPt = 0.0;
   
   for (auto i : reader->GetEntryRange()) {
      float pt = viewPt(i);
      float eta = viewEta(i);
      float phi = viewPhi(i);
      
      if (pt > 25.0 && std::abs(eta) < 2.4) {
         selectedEvents++;
         sumPt += pt;
      }
      
      if (i % 500000 == 0 && i > 0) {
         std::cout << "  Processed " << (i / 1000) << "k events..." << std::endl;
      }
   }
   
   std::cout << "[INFO] Analysis complete. Selected " << selectedEvents 
             << " events, <pT> = " << (sumPt / selectedEvents) << " GeV" << std::endl;

   auto& metrics = reader->GetMetrics();

   // Report Analysis Metrics
   std::cout << "\n===== Analysis Performance Report =====" << std::endl;
   std::cout << "Transactions: " << metrics.GetTransactions() << " (Expected > 0)" << std::endl;
   std::cout << "Sparseness:   " << metrics.GetSparseness() << std::endl;
   std::cout << "Randomness:   " << metrics.GetRandomness() << " (Expected low for sequential)" << std::endl;
   std::cout << "======================================\n" << std::endl;

   if (metrics.GetTransactions() > 0 && metrics.GetRandomness() < 1.0) {
       std::cout << "[SUCCESS] Analysis metrics look good!" << std::endl;
       std::cout << "          Sequential reading achieved (low randomness)." << std::endl;
       return 0;
   } else {
       std::cerr << "[FAIL] Unexpected metrics pattern." << std::endl;
       std::cerr << "       Transactions: " << metrics.GetTransactions() << std::endl;
       std::cerr << "       Randomness: " << metrics.GetRandomness() << std::endl;
       return 1;
   }
}

```
```bash
[INFO] Creating 50MB analysis dataset...
  Written 0k events...
  Written 500k events...
  Written 1000k events...
  Written 1500k events...
  Written 2000k events...
  Written 2500k events...
  Written 3000k events...
  Written 3500k events...
  Written 4000k events...
[INFO] Dataset created successfully.
[INFO] Opening RNTuple...

[DEBUG] Pre-Reset Metrics:
  Transactions: 4
  Randomness:   24.7884
  Total File Size: 0 bytes
  N Entries: 4200000
[INFO] Resetting Metrics to isolate analysis phase...
[INFO] Running analysis on 4200000 events...
  Processed 500k events...
  Processed 1000k events...
  Processed 1500k events...
  Processed 2000k events...
  Processed 2500k events...
  Processed 3000k events...
  Processed 3500k events...
  Processed 4000k events...
[INFO] Analysis complete. Selected 3782202 events, <pT> = 72.4999 GeV

===== Analysis Performance Report =====
Transactions: 24 (Expected > 0)
Sparseness:   0.919237
Randomness:   4.08388e-05 (Expected low for sequential)
======================================

[SUCCESS] Analysis metrics look good!
          Sequential reading achieved (low randomness).
```
</details>
## Checklist:

- [x] tested changes locally

This PR fixes #20853

### DIFF
--- a/tree/ntuple/inc/ROOT/RPageStorageFile.hxx
+++ b/tree/ntuple/inc/ROOT/RPageStorageFile.hxx
@@ -152,6 +152,19 @@ private:
    RNTupleDescriptorBuilder fDescriptorBuilder;
    /// Populated by LoadStructureImpl(), reset at the end of Attach()
    RStructureBuffer fStructureBuffer;
+   /// Tracks the last read offset for seek distance calculation
+   std::uint64_t fLastOffset = 0;
+
+   /// File-specific I/O performance counters
+   struct RFileCounters {
+      ROOT::Experimental::Detail::RNTupleAtomicCounter &fSzSkip;
+      ROOT::Experimental::Detail::RNTupleCalcPerf &fSzFile;
+      ROOT::Experimental::Detail::RNTupleCalcPerf &fRandomness;
+      ROOT::Experimental::Detail::RNTupleCalcPerf &fSparseness;
+   };
+   std::unique_ptr<RFileCounters> fFileCounters;
+   /// Total file size, set once in AttachImpl()
+   std::int64_t fFileSize = 0;
 
    RPageSourceFile(std::string_view ntupleName, const ROOT::RNTupleReadOptions &options);
 
--- a/tree/ntuple/src/RPageStorageFile.cxx
+++ b/tree/ntuple/src/RPageStorageFile.cxx
@@ -42,7 +42,20 @@
 #include <functional>
 #include <mutex>
 
+using ROOT::Experimental::Detail::RNTupleAtomicCounter;
 using ROOT::Experimental::Detail::RNTupleAtomicTimer;
+using ROOT::Experimental::Detail::RNTupleCalcPerf;
+using ROOT::Experimental::Detail::RNTupleMetrics;
+using ROOT::Internal::MakeUninitArray;
+using ROOT::Internal::RCluster;
+using ROOT::Internal::RClusterPool;
+using ROOT::Internal::RNTupleCompressor;
+using ROOT::Internal::RNTupleDecompressor;
+using ROOT::Internal::RNTupleFileWriter;
+using ROOT::Internal::RNTupleSerializer;
+using ROOT::Internal::ROnDiskPage;
+using ROOT::Internal::ROnDiskPageMap;
+using ROOT::Internal::RPagePool;
 
 ROOT::Internal::RPageSinkFile::RPageSinkFile(std::string_view ntupleName, const ROOT::RNTupleWriteOptions &options)
    : RPagePersistentSink(ntupleName, options)
@@ -315,6 +328,46 @@ ROOT::Internal::RPageSourceFile::RPageSourceFile(std::string_view ntupleName, co
    : RPageSource(ntupleName, opts)
 {
    EnableDefaultMetrics("RPageSourceFile");
+   fFileCounters = std::make_unique<RFileCounters>(RFileCounters{
+      *fMetrics.MakeCounter<RNTupleAtomicCounter *>("szSkip", "B",
+                                                    "cumulative seek distance (excluding header/footer reads)"),
+      *fMetrics.MakeCounter<RNTupleCalcPerf *>(
+         "szFile", "B", "total file size", fMetrics,
+         [this](const RNTupleMetrics &) -> std::pair<bool, double> {
+            if (fFileSize > 0)
+               return {true, static_cast<double>(fFileSize)};
+            return {false, -1.};
+         }),
+      *fMetrics.MakeCounter<RNTupleCalcPerf *>(
+         "randomness", "",
+         "ratio of seek distance to bytes read (excluding file structure reads)", fMetrics,
+         [](const RNTupleMetrics &metrics) -> std::pair<bool, double> {
+            if (const auto szSkip = metrics.GetLocalCounter("szSkip")) {
+               if (const auto szReadPayload = metrics.GetLocalCounter("szReadPayload")) {
+                  if (const auto szReadOverhead = metrics.GetLocalCounter("szReadOverhead")) {
+                     auto totalRead = szReadPayload->GetValueAsInt() + szReadOverhead->GetValueAsInt();
+                     if (totalRead > 0) {
+                        return {true, (1. * szSkip->GetValueAsInt()) / totalRead};
+                     }
+                  }
+               }
+            }
+            return {false, -1.};
+         }),
+      *fMetrics.MakeCounter<RNTupleCalcPerf *>(
+         "sparseness", "",
+         "ratio of bytes read to total file size (excluding file structure reads)", fMetrics,
+         [this](const RNTupleMetrics &metrics) -> std::pair<bool, double> {
+            if (fFileSize > 0) {
+               if (const auto szReadPayload = metrics.GetLocalCounter("szReadPayload")) {
+                  if (const auto szReadOverhead = metrics.GetLocalCounter("szReadOverhead")) {
+                     auto totalRead = szReadPayload->GetValueAsInt() + szReadOverhead->GetValueAsInt();
+                     return {true, (1. * totalRead) / fFileSize};
+                  }
+               }
+            }
+            return {false, -1.};
+         })});
 }
 
 ROOT::Internal::RPageSourceFile::RPageSourceFile(std::string_view ntupleName,
@@ -460,6 +513,9 @@ ROOT::RNTupleDescriptor ROOT::Internal::RPageSourceFile::AttachImpl(RNTupleSeria
    // For the page reads, we rely on the I/O scheduler to define the read requests
    fFile->SetBuffering(false);
 
+   // Set file size once after buffering is turned off
+   fFileSize = fFile->GetSize();
+
    return desc;
 }
 
@@ -523,8 +579,16 @@ ROOT::Internal::RPageRef ROOT::Internal::RPageSourceFile::LoadPageImpl(ColumnHan
       directReadBuffer = MakeUninitArray<unsigned char>(sealedPage.GetBufferSize());
       {
          RNTupleAtomicTimer timer(fCounters->fTimeWallRead, fCounters->fTimeCpuRead);
-         fReader.ReadBuffer(directReadBuffer.get(), sealedPage.GetBufferSize(),
-                            pageInfo.GetLocator().GetPosition<std::uint64_t>());
+         const auto offset = pageInfo.GetLocator().GetPosition<std::uint64_t>();
+         // Track seek distance (excluding file structure reads)
+         R__ASSERT(fFileCounters);
+         if (fLastOffset != 0) {
+            const auto distance = static_cast<std::uint64_t>(std::abs(
+               static_cast<std::int64_t>(offset) - static_cast<std::int64_t>(fLastOffset)));
+            fFileCounters->fSzSkip.Add(distance);
+         }
+         fReader.ReadBuffer(directReadBuffer.get(), sealedPage.GetBufferSize(), offset);
+         fLastOffset = offset + sealedPage.GetBufferSize();
       }
       fCounters->fNPageRead.Inc();
       fCounters->fNRead.Inc();
@@ -728,6 +792,18 @@ ROOT::Internal::RPageSourceFile::LoadClusters(std::span<RCluster::RKey> clusterK
                break;
             }
          }
+      }
+
+      // Track seek distance for each read request (excluding file structure reads)
+      R__ASSERT(fFileCounters);
+      for (std::size_t i = 0; i < nBatch; ++i) {
+         const auto offset = readRequests[iReq + i].fOffset;
+         if (fLastOffset != 0) {
+            const auto distance = static_cast<std::uint64_t>(std::abs(
+               static_cast<std::int64_t>(offset) - static_cast<std::int64_t>(fLastOffset)));
+            fFileCounters->fSzSkip.Add(distance);
+         }
+         fLastOffset = offset + readRequests[iReq + i].fSize;
       }
 
       if (nBatch <= 1) {
--- a/tree/ntuple/test/ntuple_metrics.cxx
+++ b/tree/ntuple/test/ntuple_metrics.cxx
@@ -112,3 +112,45 @@ TEST(Metrics, RNTupleWriter)
    // one page for the int field, one for the float field
    EXPECT_EQ(2, page_counter->GetValueAsInt());
 }
+
+TEST(Metrics, IOMetrics)
+{
+   FileRaii fileGuard("test_ntuple_io_metrics.root");
+
+   {
+      auto model = RNTupleModel::Create();
+      auto int_field = model->MakeField<int>("ints");
+      auto ntuple = RNTupleWriter::Recreate(std::move(model), "ntuple", fileGuard.GetPath());
+      for (int i = 0; i < 1000; ++i) {
+         *int_field = i;
+         ntuple->Fill();
+      }
+      ntuple->CommitCluster();
+   }
+
+   {
+      auto ntupleReader = RNTupleReader::Open("ntuple", fileGuard.GetPath());
+      EXPECT_FALSE(ntupleReader->GetMetrics().IsEnabled());
+      ntupleReader->EnableMetrics();
+      EXPECT_TRUE(ntupleReader->GetMetrics().IsEnabled());
+
+      auto view = ntupleReader->GetView<int>("ints");
+      for (auto i : *ntupleReader) {
+         (void)view(i);
+      }
+
+      const auto &metrics = ntupleReader->GetMetrics();
+      auto *randomness = metrics.GetCounter("RNTupleReader.RPageSourceFile.randomness");
+      auto *sparseness = metrics.GetCounter("RNTupleReader.RPageSourceFile.sparseness");
+      auto *szSkip = metrics.GetCounter("RNTupleReader.RPageSourceFile.szSkip");
+      auto *szFile = metrics.GetCounter("RNTupleReader.RPageSourceFile.szFile");
+
+      ASSERT_NE(randomness, nullptr);
+      ASSERT_NE(sparseness, nullptr);
+      ASSERT_NE(szSkip, nullptr);
+      ASSERT_NE(szFile, nullptr);
+
+      EXPECT_GE(szSkip->GetValueAsInt(), 0);
+      EXPECT_GT(szFile->GetValueAsInt(), 0);
+   }
+}


jblomer (changes requested):
Thank you!

Before going into a detailed review, I see a few larger items that need to be addressed.

- The metrics should not be added directly as atomic integers to the RNTupleMetrics object. We have `RNTuplePerfCounter` for this.
- Nice observation on the randomness metric. However, I would prefer doing this internally, e.g. excluding header and footer or having two metrics (with and without).
- I think it makes more sense to do the accounting in the `fReader`; should save us to chase all the invocations of `ReadBuffer`.

JasMehta08:
@jblomer  Thanks for the review. I have updated the PR to address your three main points:

Counters: I replaced the `std::atomic` members with `RNTupleAtomicCounter` (using the standard `RNTuplePerfCounter`).

Accounting: I moved all I/O accounting into `RMiniFileReader` (specifically `ReadBuffer` and `ReadV`). `RPageStorageFile` now just passes the metrics pointer to the reader.

Randomness Logic: As suggested, I handled the exclusion internally rather than via `Reset()`.

I added an internal `fSumSkipAnalysis` counter.

The reader automatically switches to the "Analysis Phase" at the end of `Open()`.

`GetRandomness()` calculates the ratio using only the analysis-phase seeks, effectively ignoring header/footer overhead without user intervention.

JasMehta08:
While verifying the changes, I noticed a pre-existing issue with `RPageSourceFile`. When the source object is moved (e.g. into the Reader), the internal `fReader` still points to the old `fMetrics` address.

This results in 0 metrics during verification tests that snapshot the reader across a move. It likely doesn't affect production (where the reader owns the source), but it breaks the "Snapshot & Subtract" verification logic.

I have not added a custom Move Constructor to fix this yet, as it felt like a separate architectural fix outside the scope of this metrics refactor. Let me know if you would prefer I include that fix here.

JasMehta08:
@jblomer 

I pushed a polish commit addressing 3 specific things that i noticed while reviewing:

- Object Lifetimes (`CloneImpl`): Explicitly documented the raw pointer dependency with a // SAFETY comment. I kept the raw pointer pattern to match the existing `RNTuple` design (Reader owns Views).

  - Question: Is this "Contract-Based Safety" acceptable here, or would `std::weak_ptr` be better for strict lifetime enforcement?

- Concurrency Docs: Added comments clarifying that `fLastOffset` is thread-local (per Clone), while `fMetrics` uses shared atomics.

- UB Safety: seek distance logic to use purely unsigned arithmetic (`offset >= fLastOffset ? `) to eliminate theoretical signed integer overflow (even though unlikely).

jblomer:
I'm afraid this isn't going quite in the right direction. Take a look at `ROOT::Internal::RPageSource::EnableDefaultMetrics`, we already have some computed metrics that you can use as a blueprint.

JasMehta08:
hey @jblomer, Thanks for the review, sorry about miss understanding the previous review.

I moved the implementation to follow the `EnableDefaultMetrics` blueprint.

please let me know if they are in the right directions, or should i look a bit deeper in the library. 

On a side note RNTuple seems very interesting to me, I'm interested in the low-level systems engineering side of this and learning about columnar data structures in general, could you let me know if there is anyway to deep dive and understand it even better (i did go through the root reference guide i could only find an introduction page. And i also went through the youtube video on the HSF youtube channel but it did not have a lot of details).

please have a review and let me know if these changes are as desired.

Thanks!

jblomer:
Thank you!

That looks better. One issue here is still that the concrete metrics are added to the `RNTupleMetrics` class. This is not the right place. The `RNTupleMetrics` is a generic binder for metrics. It is the user of the `RNTupleMetrics` class (e.g. `RPageSource`) that implements the metric-specific logic .

As an introduction documentation, I recommend the write-ups in the doc folder of the RNTuple sources. Start with the `architecture.md` (the architectural overview).

JasMehta08:
@jblomer,

I removed the concrete methods from RNTupleMetrics class. Please have a look and let me know if I should change anything.

Note: I changed the `GetCounter()` method so that, Users can access metrics via `GetCounter()` with either short names or full paths:
- `metrics.GetCounter("randomness")`
- `metrics.GetCounter("RNTupleReader.RPageSourceFile.randomness")`

as having to write the whole name for randomness would be tedious.

Both work identically. Once confirmed by you that everything has been implemented properly, I will add a tutorial and unit test (assuming that would be desired).

Thanks!

jblomer:
Thank you!

I think we should only address metrics by their full name. I'm not convinced by the short name as it may result in name collisions. In any case, I think that particular functionality should go in a separate PR.

jblomer:
Also, please rebase/squash your commits for a clean history.

JasMehta08:
@jblomer 
Thank you for the review, yes i agree about the short name ones, i had actually realised that it should be a separate PR, if even implemented, just missed removing it from the PR, sorry about that, so will do that. 

Also other than that i will add a test for this as well as please let me know if i would have to change some documentation for it or not. 

I hope the other changes are good. I will ask for a review after making the changes suggested and also adding tests.

Thanks!

JasMehta08:
@jblomer 
I have removed the short name update, i added a simple test to make sure that the metrics work, and also squashed the commits into one.

Please let me know if i should change anything else.

Thanks!

tree/ntuple/inc/ROOT/RPageStorage.hxx
```diff
@@ -650,6 +646,10 @@ protected:
       ROOT::Experimental::Detail::RNTupleCalcPerf &fBandwidthUnzip;
       ROOT::Experimental::Detail::RNTupleCalcPerf &fFractionReadOverhead;
       ROOT::Experimental::Detail::RNTupleCalcPerf &fCompressionRatio;
+      ROOT::Experimental::Detail::RNTupleAtomicCounter &fSumSkip;
+      ROOT::Experimental::Detail::RNTupleAtomicCounter &fTotalFileSize;
+      ROOT::Experimental::Detail::RNTupleCalcPerf &fRandomness;
+      ROOT::Experimental::Detail::RNTupleCalcPerf &fSparseness;
```
jblomer: These counters only make sense for files, so should move to `RPageSourceFile`

tree/ntuple/inc/ROOT/RPageStorageFile.hxx
```diff
@@ -152,8 +141,14 @@ private:
...
    /// The descriptor is created from the header and footer either in AttachImpl or in CreateFromAnchor
    RNTupleDescriptorBuilder fDescriptorBuilder;
+   /// The cluster pool asynchronously preloads the next few clusters
+   std::unique_ptr<ROOT::Internal::RClusterPool> fClusterPool;
    /// Populated by LoadStructureImpl(), reset at the end of Attach()
    RStructureBuffer fStructureBuffer;
+   /// Tracks the last read offset for seek distance calculation
+   std::uint64_t fLastOffset = 0;
+   /// Flag to track if total file size has been set (set lazily on first page load when metrics are enabled)
+   bool fFileSizeSet = false;
```
jblomer: I think we can use a negative file size as a flag and don't need the additional member.

tree/ntuple/src/RPageStorage.cxx
```diff
@@ -555,6 +542,30 @@ void ROOT::Internal::RPageSource::EnableDefaultMetrics(const std::string &prefix
                                                      }
                                                   }
                                                   return {false, -1.};
+                                               }),
+      *fMetrics.MakeCounter<RNTupleAtomicCounter *>("sumSkip", "B", "cumulative seek distance"),
```
jblomer:
```suggestion
      *fMetrics.MakeCounter<RNTupleAtomicCounter *>("szSkip", "B", "cumulative seek distance"),
```

tree/ntuple/src/RPageStorage.cxx
```diff
@@ -555,6 +542,30 @@ void ROOT::Internal::RPageSource::EnableDefaultMetrics(const std::string &prefix
...
+      *fMetrics.MakeCounter<RNTupleAtomicCounter *>("totalFileSize", "B", "total file size"),
```
jblomer:
```suggestion
      *fMetrics.MakeCounter<RNTupleAtomicCounter *>("szFile", "B", "total file size"),
```

tree/ntuple/inc/ROOT/RPageStorage.hxx
```diff
@@ -650,6 +646,10 @@ protected:
       ROOT::Experimental::Detail::RNTupleCalcPerf &fBandwidthUnzip;
       ROOT::Experimental::Detail::RNTupleCalcPerf &fFractionReadOverhead;
       ROOT::Experimental::Detail::RNTupleCalcPerf &fCompressionRatio;
+      ROOT::Experimental::Detail::RNTupleAtomicCounter &fSumSkip;
```
jblomer:
```suggestion
      ROOT::Experimental::Detail::RNTupleAtomicCounter &fSzSkip;
```

tree/ntuple/inc/ROOT/RPageStorage.hxx
```diff
@@ -650,6 +646,10 @@ protected:
...
+      ROOT::Experimental::Detail::RNTupleAtomicCounter &fTotalFileSize;
```
jblomer:
```suggestion
      ROOT::Experimental::Detail::RNTupleAtomicCounter &fSzFile;
```

tree/ntuple/src/RPageStorage.cxx
```diff
@@ -555,6 +542,30 @@ void ROOT::Internal::RPageSource::EnableDefaultMetrics(const std::string &prefix
...
+      *fMetrics.MakeCounter<RNTupleCalcPerf *>("randomness", "", "ratio of seek distance to bytes read", fMetrics,
+                                               [](const RNTupleMetrics &metrics) -> std::pair<bool, double> {
+                                                  if (const auto sumSkip = metrics.GetLocalCounter("sumSkip")) {
+                                                     if (const auto szRead = metrics.GetLocalCounter("szReadPayload")) {
```
jblomer: Here I think we want to use the actual read bytes (payload + overhead)

tree/ntuple/src/RPageStorage.cxx
```diff
@@ -555,6 +542,30 @@ void ROOT::Internal::RPageSource::EnableDefaultMetrics(const std::string &prefix
...
+                                                        if (auto payload = szRead->GetValueAsInt()) {
+                                                           return {true, (1. * sumSkip->GetValueAsInt()) / payload};
+                                                        }
+                                                     }
+                                                  }
+                                                  return {false, -1.};
+                                               }),
+      *fMetrics.MakeCounter<RNTupleCalcPerf *>("sparseness", "", "ratio of bytes read to total file size", fMetrics,
+                                               [](const RNTupleMetrics &metrics) -> std::pair<bool, double> {
+                                                  if (const auto szRead = metrics.GetLocalCounter("szReadPayload")) {
```
jblomer: Same here, I think we want to use payload + overhead

tree/ntuple/src/RPageStorageFile.cxx
```diff
@@ -459,6 +437,12 @@ void ROOT::Internal::RPageSourceFile::LoadSealedPage(ROOT::DescriptorId_t physic
                                                      RNTupleLocalIndex localIndex, RSealedPage &sealedPage)
 {
    const auto clusterId = localIndex.GetClusterId();
+   //  Set total file size on first page load (when metrics are enabled)
+   if (!fFileSizeSet && fCounters) {
+      fCounters->fTotalFileSize.SetValue(fFile->GetSize());
+      fFileSizeSet = true;
+   }
```
jblomer: Setting the file size should only be done once, I think `AttachImpl()` is the right place, after buffering is turned off.

tree/ntuple/src/RPageStorage.cxx
```diff
@@ -555,6 +542,30 @@ void ROOT::Internal::RPageSource::EnableDefaultMetrics(const std::string &prefix
                                                      }
                                                   }
                                                   return {false, -1.};
+                                               }),
+      *fMetrics.MakeCounter<RNTupleAtomicCounter *>("sumSkip", "B", "cumulative seek distance"),
+      *fMetrics.MakeCounter<RNTupleAtomicCounter *>("totalFileSize", "B", "total file size"),
+      *fMetrics.MakeCounter<RNTupleCalcPerf *>("randomness", "", "ratio of seek distance to bytes read", fMetrics,
```
jblomer: Here and for sparseness: we should clarify that the metrics excludes reading the file structure (anchor, header, footer).

tree/ntuple/test/ntuple_metrics.cxx
```diff
@@ -112,3 +112,48 @@ TEST(Metrics, RNTupleWriter)
...
+
+TEST(Metrics, IOMetrics)
+{
+   std::string rootFileName{"test_ntuple_io_metrics.root"};
+   FileRaii fileGuard(rootFileName);
+
+   {
+      auto model = RNTupleModel::Create();
+      auto int_field = model->MakeField<int>("ints");
+      auto ntuple = RNTupleWriter::Recreate(std::move(model), "ntuple", rootFileName);
```
jblomer:
```suggestion
      auto ntuple = RNTupleWriter::Recreate(std::move(model), "ntuple", fileGuard.GetPath());
```

tree/ntuple/test/ntuple_metrics.cxx
```diff
@@ -112,3 +112,48 @@ TEST(Metrics, RNTupleWriter)
    // one page for the int field, one for the float field
    EXPECT_EQ(2, page_counter->GetValueAsInt());
 }
+
+TEST(Metrics, IOMetrics)
+{
+   std::string rootFileName{"test_ntuple_io_metrics.root"};
```
jblomer: Not needed, you can ask the file guard for the path

tree/ntuple/src/RPageStorageFile.cxx
```diff
@@ -515,16 +505,23 @@ ROOT::Internal::RPageRef ROOT::Internal::RPageSourceFile::LoadPageImpl(ColumnHan
       directReadBuffer = MakeUninitArray<unsigned char>(sealedPage.GetBufferSize());
       {
          RNTupleAtomicTimer timer(fCounters->fTimeWallRead, fCounters->fTimeCpuRead);
-         fReader.ReadBuffer(directReadBuffer.get(), sealedPage.GetBufferSize(),
-                            pageInfo.GetLocator().GetPosition<std::uint64_t>());
+         auto offset = pageInfo.GetLocator().GetPosition<std::uint64_t>();
```
jblomer:
```suggestion
         const auto offset = pageInfo.GetLocator().GetPosition<std::uint64_t>();
```

tree/ntuple/src/RPageStorageFile.cxx
```diff
@@ -515,16 +505,23 @@ ROOT::Internal::RPageRef ROOT::Internal::RPageSourceFile::LoadPageImpl(ColumnHan
...
+         // Track seek distance
+         if (fLastOffset != 0) {
+            std::int64_t diff = static_cast<std::int64_t>(offset) - static_cast<std::int64_t>(fLastOffset);
```
jblomer:
```suggestion
            const std::int64_t diff = static_cast<std::int64_t>(offset) - static_cast<std::int64_t>(fLastOffset);
```

tree/ntuple/src/RPageStorageFile.cxx
```diff
@@ -515,16 +505,23 @@ ROOT::Internal::RPageRef ROOT::Internal::RPageSourceFile::LoadPageImpl(ColumnHan
...
+            std::uint64_t distance = (diff >= 0) ? diff : -diff;
+            fCounters->fSumSkip.Add(distance);
```
jblomer:
```suggestion
            fCounters->fSumSkip.Add(std::abs(diff));
```

tree/ntuple/src/RPageStorageFile.cxx
```diff
@@ -535,6 +532,16 @@ ROOT::Internal::RPageRef ROOT::Internal::RPageSourceFile::LoadPageImpl(ColumnHan
...
       R__ASSERT(onDiskPage && (sealedPage.GetBufferSize() == onDiskPage->GetSize()));
+      
+      // Track seek distance based on page location even when using cluster pool
+      auto offset = pageInfo.GetLocator().GetPosition<std::uint64_t>();
+      if (fLastOffset != 0) {
+         std::int64_t diff = static_cast<std::int64_t>(offset) - static_cast<std::int64_t>(fLastOffset);
+         std::uint64_t distance = (diff >= 0) ? diff : -diff;
+         fCounters->fSumSkip.Add(distance);
+      }
+      fLastOffset = offset + sealedPage.GetBufferSize();
```
jblomer: This is wrong, we need to track the actual read requests in `LoadClusters()`

jblomer (changes requested):

JasMehta08:
@jblomer Thank you for the review,

I have made all the changes as suggested, 
The only additional change that was needed was for moving the setting of the file size to `AttachImpl()` we would have to track the size using another variable as the metrics are not enabled by that point,

I have used a member variable to store the file size `fFileSize`(let me know if i should change the name), and now the metric `fSzFile` is a computed metric rather than it storing the size it returns `fFileSIze`.

Thank you!

aashirvad08:
I'm still exploring the codebase, but I was wondering -would these metrics (randomness, sparseness) also be useful for remote backends like S3, where reads are coalesced into ranged HTTP requests rather than local seeks?

JasMehta08:
hey @jblomer ,

can you let me know if there are any further improvements that i should make on the PR.

Thanks!

tree/ntuple/src/RPageStorageFile.cxx
```diff
@@ -722,6 +754,19 @@ ROOT::Internal::RPageSourceFile::LoadClusters(std::span<RCluster::RKey> clusterK
          }
       }
 
+      // Track seek distance for each read request (excluding file structure reads)
+      if (fFileCounters) {
```
jblomer: fFileCounters should always be there, so I think we can just `assert()` that it exists.

tree/ntuple/src/RPageStorageFile.cxx
```diff
@@ -515,16 +539,23 @@ ROOT::Internal::RPageRef ROOT::Internal::RPageSourceFile::LoadPageImpl(ColumnHan
       directReadBuffer = MakeUninitArray<unsigned char>(sealedPage.GetBufferSize());
       {
          RNTupleAtomicTimer timer(fCounters->fTimeWallRead, fCounters->fTimeCpuRead);
-         fReader.ReadBuffer(directReadBuffer.get(), sealedPage.GetBufferSize(),
-                            pageInfo.GetLocator().GetPosition<std::uint64_t>());
+         const auto offset = pageInfo.GetLocator().GetPosition<std::uint64_t>();
+         // Track seek distance (excluding file structure reads)
+         if (fLastOffset != 0 && fFileCounters) {
```
jblomer: just `assert(fFileCounters)`

jblomer (approved):
Very nice! Thank you.

Apologies for having this left unhandled for so long.

After rebasing and the two tiny comments, we'll merge.

JasMehta08:
@jblomer 

I have made the same changes and also rebased, please let me know if there is anything left to do.

Thanks!

PR #20996, opened 2026-01-22
[CMake] Don't use find_package(ROOT) in `tutorials/` subdirectory

guitargeek:
As Sergey said in the comment in `tutorials/CMakeLists.txt`, this tutorials project stopped working standalone. But, it still uses `find_package(ROOT)` to find ROOT, even though it's not installed yet as we are building ROOT itself at the same time.

This should better not be done. This commit suggest to remove it, and also replaces the variables defined by `find_package(ROOT)` with the built time variables.

This follows up on 9d2dc6e1641b5e c21bc3ffc, and 458803ee4f, which did the same for the `test/` and `roottest/` subdirectories.

### DIFF
--- a/tutorials/CMakeLists.txt
+++ b/tutorials/CMakeLists.txt
@@ -4,34 +4,14 @@
 # For the licensing terms see $ROOTSYS/LICENSE.
 # For the list of contributors see $ROOTSYS/README/CREDITS.
 
-# CMakeLists.txt for the ROOT tutorials programs.
-# Author: Pere Mato, 25/10/2010
-cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
-
-project(tutorials)
-
-# Sergey: make no sense while CMakeLists.txt file cannot be used separately from ROOT
-# but variables like ROOT_asimage_FOUND used here and produced in ROOTConfig.cmake
-find_package(ROOT REQUIRED)
-
-if(DEFINED ROOT_SOURCE_DIR)  # Testing using the binary tree
-  set(ROOT_root_CMD root.exe)
-  if(NOT MSVC)  # Ignore environment on Windows
-    set(ROOT_environ PATH=${CMAKE_BINARY_DIR}/bin:$ENV{PATH}
-                     ROOT_INCLUDE_PATH=${CMAKE_BINARY_DIR}/tutorials/io/tree:${DEFAULT_ROOT_INCLUDE_PATH}
-                     PYTHONPATH=${CMAKE_BINARY_DIR}/lib:$ENV{PYTHONPATH})
-  else()
-    set(ROOT_environ ROOT_INCLUDE_PATH=${CMAKE_BINARY_DIR}/tutorials/io/tree:${DEFAULT_ROOT_INCLUDE_PATH}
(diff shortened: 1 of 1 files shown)


tutorials/CMakeLists.txt
```diff
@@ -4,34 +4,14 @@
...
-  endif()
-else()                       # testing using an installation
-  include(${ROOT_USE_FILE})
-  if(DEFINED ROOT_CONFIG_EXECUTABLE) #---If ROOT was built with the classic configure/make---
-    set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/../cmake/modules)
-    include(RootMacros)
-    set(ROOT_root_CMD root.exe)
-  endif()
-  enable_testing()
+set(ROOT_root_CMD root.exe)
```
hageboeck, line 7: This will call `/usr/bin/root.exe` or similar if installed.

```suggestion
set(ROOT_root_CMD $<TARGET_FILE:root>)
```

tutorials/CMakeLists.txt
```diff
@@ -4,34 +4,14 @@
...
+if(NOT MSVC)  # Ignore environment on Windows
+  set(ROOT_environ PATH=${CMAKE_BINARY_DIR}/bin:$ENV{PATH}
```
hageboeck, line 9: This PATH will theoretically fix up the `root.exe` path issue above, but I would still invoke it with the full path. It makes things clearer when you debug.

If you think whether the `PATH` can be removed now, I don't think so, because there might be some things like `hist2workspace` that might need to be invoked during the tutorials. 🙂

guitargeek: Good point! In the end, I want to make everything independent of environment variables, for better portability. If there is a tutorial that starts root binaries, they should prepend `gROOT->GetBinaryDir()`. Would you agree?

guitargeek:
Thanks for the review! Your comments are perfectly valid and I will consider them, but as they are independent from this PR I'll do that in a followup PR.

tutorials/CMakeLists.txt
hageboeck: > If there is a tutorial that starts root binaries, they should prepend gROOT->GetBinaryDir(). Would you agree?

It's a laudable goal, but I doubt that we are doing it consistently. One sign of non-compliance is `ProcessLine ... .!`:
```
$ git grep "ProcessLine.*\.\!"
tutorials/roofit/histfactory/hf001_example.C:      gROOT->ProcessLine(".! prepareHistFactory .");
tutorials/roofit/histfactory/hf001_example.py:    ROOT.gROOT.ProcessLine(".! prepareHistFactory .")
[...]
```
This alone generated 56 hits. There might be more things like subprocesses, but I didn't check for those.

hageboeck: Tagging @guitargeek for info. See the reply above. 😅

guitargeek: Thanks! That's a good signature to look for. Ok you are right, we are far from not relying on `PATH` :slightly_smiling_face:

PR #20998, opened 2026-01-23
Make RBatchLoader class more general and introduce class for tensor operators

martinfoell:
# This Pull request:
- Remove the training and validation distinction in the functions in RBatchLoader to make it more general
- Add number of entries and drop remainder bool as input parameters to RBatchLoader to have a cleaner interaction between RBatchGenerator and RBatchLoader
- Introduce class for tensor operators that simplifies RChunkLoader

### DIFF
--- a/tmva/tmva/inc/TMVA/BatchGenerator/RBatchLoader.hxx
+++ b/tmva/tmva/inc/TMVA/BatchGenerator/RBatchLoader.hxx
@@ -118,9 +118,9 @@ public:
 
    /// \brief Return a batch of data as a unique pointer.
    /// After the batch has been processed, it should be destroyed.
-   /// \param[in] chunkTensor RTensor with the data from the chunk
+   /// \param[in] chunkTensor Tensor with the data from the chunk
    /// \param[in] idxs Index of batch in the chunk
-   /// \return Training batch
+   /// \return Batch
    std::unique_ptr<RFlat2DMatrix> CreateBatch(RFlat2DMatrix &chunTensor, std::size_t idxs)
    {
       auto batch = std::make_unique<RFlat2DMatrix>(fBatchSize, fNumColumns);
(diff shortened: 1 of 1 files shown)


martinfoell:
> The code LGTM, nice work! Let's first make sure the CI is green, then we can merge.

Thank you!

tmva/tmva/inc/TMVA/BatchGenerator/RFlat2DMatrixOperators.hxx
```diff
@@ -0,0 +1,74 @@
...
+      : fShuffle(shuffle),
+        fSetSeed(setSeed)
+   {
+   
+   }
+
+  void ShuffleTensor(RFlat2DMatrix &ShuffledTensor, RFlat2DMatrix &Tensor ) {
+    std::random_device rd;
+    std::mt19937 g;
+
```
guitargeek, line 43: Just a fly-by comment to be considered in a followup PR.

I think we should prefer our own TRandom classes over features from `<random>`, because the latter are compiler-dependent. I think it would add value in terms of reproducibility across platforms if we use our compiler-independet TRandom algorithms instead.

See also https://github.com/root-project/root/commit/e1f42658629ad67ce7cb224a1379d133f61f2e35

vepadulano: Thanks for the comment. I see your point, I think we have to discuss this better because it's one thing doing it just internally in our code but if this should become some kind of recommendation for users I believe it's quite tricky to communicate.

In any case, we'll leave this discussion for some time in the future, this PR can move forward meanwhile.

guitargeek: Okay! Yes we can discuss later. I think it's quite easy to communicate in the TRrandom docs or tutorials:

> To ensure analysis reproducibility, please consider using TRandom instead the C++ standard library to ensure reproducible pseudorandom numbers across compiler versions and platforms.

PR #21001, opened 2026-01-23
[io] Add bounds checking to TMemFile

silverweed:
Currently we just crash (if we're lucky) when we try to read past the TMemFile's size due to TMemFile issuing a memcpy with negative length.

Assert that len >= 0 to avoid reading out of bounds.

### DIFF
--- a/io/io/src/TMemFile.cxx
+++ b/io/io/src/TMemFile.cxx
@@ -454,6 +454,8 @@ Int_t TMemFile::SysReadImpl(Int_t, void *buf, Long64_t len)
          len = fSize - fSysOffset;
       }
 
+      R__ASSERT(len >= 0);
+
       if (fBlockOffset+len <= fBlockSeek->fSize) {
          // 'len' does not go past the end of the current block,
          // so let's make a simple copy.


io/io/src/TMemFile.cxx
```diff
@@ -454,6 +454,8 @@ Int_t TMemFile::SysReadImpl(Int_t, void *buf, Long64_t len)
          len = fSize - fSysOffset;
       }
 
+      R__ASSERT(len >= 0);
```
jblomer, line 457: That looks reasonable but why is the check in the lines before not effective?

silverweed: The only check is that `fSysOffset + len > fSize`, which says nothing about the sign of `len`. Even if the `if` is entered it's not guaranteed that `fSysOffset <= fSize` so you can end up with a negative `len`.

Examples:
- fSysOffset = 10000, len = 100, fSize = 200 ----> (`if` is entered) --> len = -9800
- fSysOffset = 0, len = -100, fSize = 100 ---> (`if` is not entered) --> len = -100

pcanal (approved):

PR #21003, opened 2026-01-23
[CMake] Don't set `PATH` for running tutorials and explicitly get ROOT executable from target

guitargeek:
Follows up on https://github.com/root-project/root/pull/20996, implementing suggestions by @hageboeck.

### DIFF
--- a/tutorials/CMakeLists.txt
+++ b/tutorials/CMakeLists.txt
@@ -4,7 +4,7 @@
 # For the licensing terms see $ROOTSYS/LICENSE.
 # For the list of contributors see $ROOTSYS/README/CREDITS.
 
-set(ROOT_root_CMD root.exe)
+set(ROOT_root_CMD $<TARGET_FILE:root.exe>)
 
 set(pythonpaths ${localruntimedir} $ENV{PYTHONPATH})
 cmake_path(CONVERT "${pythonpaths}" TO_NATIVE_PATH_LIST pythonpaths_native)


tutorials/CMakeLists.txt
```diff
@@ -5,22 +5,24 @@
 # For the list of contributors see $ROOTSYS/README/CREDITS.
 
 set(ROOT_root_CMD root.exe)
-if(NOT MSVC)  # Ignore environment on Windows
-  set(ROOT_environ PATH=${CMAKE_BINARY_DIR}/bin:$ENV{PATH}
-                   ROOT_INCLUDE_PATH=${CMAKE_BINARY_DIR}/tutorials/io/tree:${DEFAULT_ROOT_INCLUDE_PATH}
```
hageboeck, line 10: We should maybe check with @LukasBreitwieser if the default include path should be retained here.

guitargeek: Okay, we decided to keep it just to be safe.

tutorials/CMakeLists.txt
```diff
@@ -5,22 +5,27 @@
...
-endif()
 
-# - Set the environment for the tutorials, which is the eventual ROOT_environ
-# plus some environment variables related to limiting the number of threads
-# used by NumPy and enabling batch mode.
+set(pythonpaths ${localruntimedir} $ENV{PYTHONPATH})
+cmake_path(CONVERT "${pythonpaths}" TO_NATIVE_PATH_LIST pythonpaths_native)
+
+set(root_includepaths ${CMAKE_BINARY_DIR}/tutorials/io/tree ${DEFAULT_ROOT_INCLUDE_PATH})
+cmake_path(CONVERT "${root_includepaths}" TO_NATIVE_PATH_LIST root_includepaths_native)
```
hageboeck, line 13: Does this do the right thing if the default path list is more than one element? I *think* the first command would generate a list such as `A;B:C:D`, and now the question is what the second command does when the second element is a list separated by `:`.

guitargeek:
Thanks! I studied the behavior of `cmake_path` locally with several printouts for some cases, and it works as intended to build these colon-separated paths (I trust it that it will correctly use semicolons on Windows :slightly_smiling_face: ).

tutorials/CMakeLists.txt
guitargeek: Good question! I was also thinking about that, but the good news is is that the right separator is already used in `DEFAULT_ROOT_INCLUDE_PATH`: https://github.com/root-project/root/blob/master/cmake/modules/RootMacros.cmake#L2273

So fortunately it is correct in any case.

PR #21004, opened 2026-01-23
[cppyy] Skip crashing test using `boost::type_erasure`

hahnjo:
It was enabled in commit 50ae29d434 and marked as `xfail(strict=True)` in commit a7131ffd3d, but is still crashing with assertions enabled on AlmaLinux 8: "Must not nest within unloading transaction"

### DIFF
--- a/bindings/pyroot/cppyy/cppyy/test/test_boost.py
+++ b/bindings/pyroot/cppyy/cppyy/test/test_boost.py
@@ -141,7 +141,7 @@ class TestBOOSTERASURE:
         cppyy.include("boost/type_erasure/member.hpp")
         cppyy.include("boost/mpl/vector.hpp")
 
-    @mark.xfail(strict=True, reason="TClassInfo for 'Lengths' is missing")
+    @mark.xfail(run=False, reason="TClassInfo for 'Lengths' is missing & triggering Cling assertions during unloading")
     def test01_erasure_usage(self):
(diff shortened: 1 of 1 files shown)


bindings/pyroot/cppyy/cppyy/test/test_boost.py
```diff
@@ -141,7 +141,7 @@ def setup_class(cls):
         cppyy.include("boost/type_erasure/member.hpp")
         cppyy.include("boost/mpl/vector.hpp")
 
-    @mark.xfail(strict=True, reason="TClassInfo for 'Lengths' is missing")
+    @mark.skip(reason="TClassInfo for 'Lengths' is missing & triggering Cling assertions during unloading")
```
guitargeek:
```suggestion
    @mark.xfail(run=False, reason="TClassInfo for 'Lengths' is missing & triggering Cling assertions during unloading")
```
Thank for the PR! Can we go for `xfail(run=False)` here? "skip" is used to mark test as skipped for "good" reasons, like some missing dependency or because the cppyy test was not well designed (stochastic and fails sometimes) or makes no sense in the context of ROOT.

For tests that we unfortunately expect to fail because of more or less known problems, we use `xfail(strict=True)`. If they are expected to crash and kill the process because of unsolved problems, we can't run them and they get a `xfail(run=False)`.

PR #21008, opened 2026-01-23
[hist] Implement conversion of bin content types

hahnjo:


hist/histv7/test/hist_engine.cxx
```diff
@@ -207,6 +207,29 @@ TEST(RHistEngine, Clone)
...
+
+   RHistEngine<float> engineF = engineI.Convert<float>();
+   ASSERT_EQ(engineF.GetNDimensions(), 1);
+   ASSERT_EQ(engineF.GetTotalNBins(), Bins + 2);
+
+   EXPECT_EQ(engineF.GetBinContent(RBinIndex::Underflow()), 1);
+   for (auto index : axis.GetNormalRange()) {
+      EXPECT_EQ(engineF.GetBinContent(index), 1);
+   }
+   EXPECT_EQ(engineF.GetBinContent(RBinIndex::Overflow()), 1);
```
hageboeck: I can see that this works now, but to be safe in the future, one could choose a different bin content than 1 everywhere.

hahnjo: I added some additional `Fill`s for every second bin.

PR #21012, opened 2026-01-23
[ci] Move arm build to alma10

dpiparo:
This also removes some pressure from the nodes and the infrastructure in general, e.g. for caching images.


guitargeek:
Good idea! But we need to figure out before merging if the `test-root-hist-h2root` failure was just the usual spurious failure and you were just very unlucky, or if it's a real problem on ARM with newer compilers that was fixed in later releases (I vaguely remember doing something like that...)

dpiparo:
it looks like a real issue!

guitargeek:
This issue should be fixed now with this backport: https://github.com/root-project/root/pull/21022

PR #21015, opened 2026-01-23
[tmva][sofie] Add a new SOFIE tutorial to show full pipeline from ONNX

lmoneta:


tutorials/machine_learning/TMVA_SOFIE_ONNX.py
```diff
@@ -0,0 +1,156 @@
...
+      y_pred = model(x)
+      loss = criterion(y_pred,y)
+      optimizer.zero_grad()
+      loss.backward()
+      optimizer.step()
+
+   model.eval()
+   #export the model to ONNX
+   modelFile = modelName + ".onnx"
+   dummy_x = torch.randn(1,32)
```
sanjibansg, line 55: Should we add a comment here as to why we need a `dummy_x` for the ONNX export method?

PR #21017, opened 2026-01-24
Remove stray line break when running `root -q` with input files

guitargeek:
ROOT first prints a newline before going over the inputs to separate printouts from the initial "root [0]" prompt. But if root is run with "-q", there is no prompt and we don't need the newline.

Removing this spurious newline completely avoids all output when building ROOT, where the newline used to happen when running `hsimple.C`.

This is commit is similar to 8e8b6bbe79594f0, which removed another stray linebreak in the opposite case of running `root` with no input files.


ferdymercury:
Thanks!

I guess this fixes a part of https://its.cern.ch/jira/browse/ROOT-8344 ?

PR #21030, opened 2026-01-26
[ntuple] Add RPageStorage::Close

silverweed:
# This Pull request:
adds a virtual `Close` method to `RPageStorage`, to be used by pythonizations of `RNTupleReader` and `RNTupleWriter`.
Also updates those pythonizations to use it.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
@@ -141,6 +141,12 @@ def _RNTupleReader_LoadEntry(self, *args):
     return self._LoadEntry(*args)
 
 
+def _RNTupleReader_exit(self, *args):
+    import ROOT
+    ROOT.Internal.CloseRNTupleReader(self)
+    return False
+
+
 @pythonization("RNTupleReader", ns="ROOT")
 def pythonize_RNTupleReader(klass):
     klass._Open = klass.Open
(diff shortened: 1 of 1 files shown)


tree/ntuple/inc/ROOT/RPageStorage.hxx
```diff
@@ -191,6 +191,9 @@ public:
    /// overriding this.
    virtual ROOT::Experimental::Detail::RNTupleMetrics &GetMetrics() { return fMetrics; }
 
+   // Called only by the Pythonization. Closes the source/sink without destroying it.
+   virtual void Close() {}
```
vepadulano, line 195: Any reason not to make it `= 0` ?

silverweed: Just to avoid implementing it in all our mock classes

silverweed:
Rethinking about this, I now think this is not the best solution. 

As far as I see it, the `Close` method is not really needed on the C++ side and it's just a side effect of Python needing a "valid but invalid" object after the context manager exits. If we go and add this to the C++ side we will end up with additional unwarranted code complexity stemming from the fact that the Reader and Writer lose the invariant that the source/sink is valid as long as they are valid, which will result in adding validity checks around which would not be needed in principle.

I think the proper solution would be to do this purely on the Python side: in the `__exit__` method we reset the internal pointer to the reader/writer (assuming this is possible) and hijack the proxy object to raise an exception whenever you try to access a method of the underlying object.

@vepadulano @jblomer opinions on this?

jblomer:
Thank you, @silverweed!

I agree, I think I'd prefer an pure Python approach in this case.

silverweed:
Alternative PR: #21131

PR #21043, opened 2026-01-27
[cling] Unload implicit instantiations also from ASTFile

### LINKED ISSUE #20638: [cling] Able to call function with failing `static_assert` on second try
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

`ROOT::Experimental::RHistEngine<BinContentType>::Scale(double)` must not be called for integral bin content types. This is enforced with a `static_assert`, but it is still possible to call the function when retrying.

### Reproducer

```
root [0] ROOT::Experimental::RHistEngine<int> hist(10, {5.0, 15.0});
root [1] hist.Scale(1);
In module 'ROOTHist':
/home/jhahnfel/ROOT/build-asserts-clang/include/ROOT/RHistEngine.hxx:486:21: error: static assertion failed due to requirement '!std::is_integral_v<int>': scaling is not supported for integral bin content types
      static_assert(!std::is_integral_v<BinContentType>, "scaling is not supported for integral bin content types");
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ROOT_prompt_1:1:6: note: in instantiation of member function 'ROOT::Experimental::RHistEngine<int>::Scale' requested here
hist.Scale(1);
     ^
root [2] hist.Scale(1);
root [3]
```

### ROOT version

`master`

### Installation method

from source

### Operating system

Linux

### Additional context

The (possible) reason is that `Scale` is a non-templated method in the templated class `RHistEngine`. During unloading, we probably only take care of entire templated classes and templated methods (?).

hahnjo:
The assumption is that serialized declarations must be valid, however that doesn't hold anymore in the presence of static_assert in templated classes.

Fixes #20638

### DIFF
--- a/interpreter/cling/lib/Interpreter/DeclUnloader.cpp
+++ b/interpreter/cling/lib/Interpreter/DeclUnloader.cpp
@@ -511,6 +511,22 @@ namespace cling {
       m_FilesToUncache.insert(FID);
   }
 
+  bool DeclUnloader::UnloadDecl(Decl* D) {
+    if (D->isFromASTFile()) {
+      // We cannot ignore instantiated CXXMethodDecl's in templated classes:
+      // They may have failing static_assert.
+      if (auto* M = dyn_cast<CXXMethodDecl>(D)) {
+        if (M->getTemplateSpecializationKind() !=
+            TemplateSpecializationKind::TSK_ImplicitInstantiation) {
+          return true;
+        }
+      } else {
+        return true;
+      }
+    }
+    return Visit(D);
+  }
+
   bool DeclUnloader::VisitDecl(Decl* D) {
     assert(D && "The Decl is null");
     CollectFilesToUncache(D->getBeginLoc());
--- a/interpreter/cling/lib/Interpreter/DeclUnloader.h
+++ b/interpreter/cling/lib/Interpreter/DeclUnloader.h
@@ -59,11 +59,7 @@ namespace cling {
     ///\param[in] D - The declaration to unload
     ///\returns true on success.
     ///
-    bool UnloadDecl(clang::Decl* D) {
-      if (D->isFromASTFile())
-        return true;
-      return Visit(D);
-    }
+    bool UnloadDecl(clang::Decl* D);
 
     ///\brief If it falls back in the base class just remove the declaration
     /// only from the declaration context.
--- a/interpreter/cling/test/CodeUnloading/PCH/Inputs/TemplateAssert.h
+++ b/interpreter/cling/test/CodeUnloading/PCH/Inputs/TemplateAssert.h
@@ -1,0 +1,18 @@
+#ifndef TemplateAssert_h
+#define TemplateAssert_h
+
+template <typename T> struct Trait {
+  static constexpr bool Value = false;
+};
+
+template <> struct Trait<double> {
+  static constexpr bool Value = true;
+};
+
+template <typename T> struct RequiresTrait {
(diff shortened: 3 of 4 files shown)


interpreter/cling/lib/Interpreter/DeclUnloader.cpp
```diff
@@ -511,6 +511,22 @@ namespace cling {
       m_FilesToUncache.insert(FID);
   }
 
+  bool DeclUnloader::UnloadDecl(Decl* D) {
+    if (D->isFromASTFile()) {
+      // We cannot ignore instantiated CXXMethodDecl's in templated classes:
+      // They may have failing static_assert.
+      if (auto* M = dyn_cast<CXXMethodDecl>(D)) {
```
vgvassilev, line 518: Is that change sufficient for these things to be reinstantiated? In particular, non-static_assert method decls coming from ast files? We should probably add some tests.

dpiparo: This could be relatively easy, right? If I understand correctly we need a template broghut into memory from a pcm and then 2 instantiations in a row - please correct me if I am wrong.
If the above is correct, and this cannot be put in cling since we need a dictionary, we can comfortably put the test in roottest, maybe even profiting from ROOT helpers that at runtime allow us to generate dictionaries on the fly

hahnjo: Yes, Dev has a test case (in roottest) for the problem I'm trying to solve. We don't have a (new) test for an unloaded method that does not have a `static_assert` (I believe that's what Vassil is asking about). It's probably covered by other tests, but I need to check. Moreover there are two failing tests (`roottest-cling-reload-reloadvector` and `roottest-root-tree-selector-selector`) that I don't have locally...

vgvassilev: Yes, having tests in cling is better because when we break this it's far easier to debug. The test does not need a dictionary. We need a module file with a static assert declaration.

hahnjo: We don't have a single test using modules / pcm's in `interpreter/cling/test` and I'm not going to invent the infrastructure for this PR. The test is going to go into roottest.

vgvassilev: Thank you for being constructive. In that case please hold off merging this PR until I provide you with the infrastructure (which is 3 lines).

hahnjo: Ok.

vgvassilev: In fact the infrastructure seems to be there already. Please take a look at `cling/test/CodeUnloading/PCH/`. We should be able to extend some of these tests to cover that case, too.

hahnjo: I added a Cling test that is fixed by the change. I also had a quick look at `roottest-root-tree-selector-selector` and it seems that one is failing because it cannot resolve `_ZNKSt13__atomic_baseIhEcvhEv = std::__atomic_base<unsigned char>::operator unsigned char()` and some related methods. It's probably now unloaded because of previous errors, but not correctly re-instantiated?

vgvassilev: Yes, because the ASTReader has a cache of deserialized declarations. Maybe some parts are in that cache?

PR #21044, opened 2026-01-27
[cppyy] Disable tests that fail with recent MSVC

bellenot:
Disable tests that fail with recent versions of Visual Studio
To be reviewed after the migration to the new Windows 11 runners

### DIFF
--- a/bindings/pyroot/cppyy/cppyy/test/test_advancedcpp.py
+++ b/bindings/pyroot/cppyy/cppyy/test/test_advancedcpp.py
@@ -714,7 +714,7 @@ class TestADVANCEDCPP:
         assert len(cppyy.gbl.gtestv1) == 1
         assert len(cppyy.gbl.gtestv2) == 1
 
-    @mark.skipif(IS_WINDOWS, reason="Crashes with some versions of Visual Studio")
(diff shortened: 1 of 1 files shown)


bindings/pyroot/cppyy/cppyy/test/test_advancedcpp.py
```diff
@@ -714,6 +714,7 @@ def test21_access_to_global_variables(self):
         assert len(cppyy.gbl.gtestv1) == 1
         assert len(cppyy.gbl.gtestv2) == 1
 
+    @mark.skipif(IS_WINDOWS, reason="Crashes with some versions of Visual Studio")
```
guitargeek:
```suggestion
    @mark.xfail(run=False, condition=IS_WINDOWS, reason="Crashes with some versions of Visual Studio")
```
Can we stick with the convention that `skipif()` is only used for "good" reasons, like missing dependencies etc., and tests that crash for no good reason are marked as `xfail(run=False)`? Thanks!

bellenot: No problem, I wasn't aware of such convention...

PR #21045, opened 2026-01-27
Add BVH-based navigation functions to TGeoTessellated

sawenzel:
This commit adds implementations of the navigation interfaces required by TGeoNavigator to TGeoTessellated.

As a result, TGeoTessellated can now be used in applications that rely on this navigator, such as VMC-based GEANT simulations.

The implementation is based on fast Bounding Volume Hierarchy (BVH) queries introduced in a recent TGeoParallelWorld rewrite.

The BVH is only exposed to the internal implementation. It is not streamed to disk for now and therefore needs to be reinitialized when reading from TBuffer. For this reason, a custom streamer for TGeoTessellated is proposed.

The implementation shows very good performance, being multiple times faster than comparable implementations in Geant4 or VecGeom. The latter were also used to validate the algorithms. Navigation functionality has additionally been tested at a high level using VecGeom’s XRayNavigationBenchmarker.

Note that this is an initial implementation and should be considered beta quality. The following limitations and remarks apply:

- DistFromInside and DistFromOutside do not implement all `iact=` cases. Only `iact == 3` is supported, which is the case actually used by TGeoNavigator.

- The algorithms have not yet been validated in large-scale production environments.

- Further optimization seem possible and will be done in future commits.

This work is motivated by the need to GEANT-simulate detector prototypes from CAD models in ALICE. ALICE is using TGeo.

Finally, the commit also fixes some minor bugs in the TGeoChecker with respect to bounding box origins, previously not taken into account.


sawenzel:
Sharing code for initial discussion. Adding a table to get idea about performance for a 20K triangular faceted sphere (testobject):
image

The simple BVH performs well in comparison to the Voxelization implementation of G4 and the SIMD-based structure of VecGeom.

sawenzel:
Further, validation tests have been performed in the level of the global navigation system (G4Navigator, TGeoNavigator, VecGeomNavigator) within the XRayBenchmark of VecGeom. All implementations yield the same XRay image which gives confidence that the present implementation is ok.

image


Overall, the present PR should present a significant step forward for the TGeo navigation system to handle tessellated shapes from CAD models.

geom/geom/inc/LinkDef1.h
```diff
@@ -77,7 +77,7 @@
 #pragma link C++ class TGeoXtru + ;
 #pragma link C++ class ROOT::Geom::Vertex_t + ;
 #pragma link C++ class TGeoFacet + ;
-#pragma link C++ class TGeoTessellated + ;
+#pragma link C++ class TGeoTessellated - ;
```
dpiparo, line 80: Do we need the dictionary at all?

geom/geom/inc/TGeoTessellated.h
```diff
@@ -129,6 +139,27 @@ class TGeoTessellated : public TGeoBBox {
...
+   bool Contains_Loop(const Double_t *point) const;
+
+   Double_t Capacity() const override;
+
+private:
+   // a safety kernel used in multiple implementations
+   template <bool closest_facet = false>
+   Double_t SafetyKernel(const Double_t *point, bool in, int *closest_facet_id = nullptr) const;
+
    ClassDefOverride(TGeoTessellated, 1) // tessellated shape class
```
dpiparo: we probably can use version 0 as we will never do IO of this class...

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -174,8 +186,8 @@ bool TGeoTessellated::AddFacet(const Vertex_t &pt0, const Vertex_t &pt1, const V
    fNseg += 3;
    fFacets.emplace_back(ind[0], ind[1], ind[2]);
 
-   if (fNfacets > 0 && GetNfacets() == fNfacets)
-      CloseShape();
+   // if (fNfacets > 0 && GetNfacets() == fNfacets)
```
dpiparo, line 190: I would remove the dead code if possible

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+   auto q = cross(tvec, e1);
+   auto v = dot(dir, q) * invDet;
+   if (v < 0.0 || u + v > 1.0) {
+      return INF;
+   }
+   auto t = dot(e2, q) * invDet;
+   return (t > rayEPS) ? t : INF;
+}
+
+template <typename T = float>
```
dpiparo, line 798: Would ROOT::Math::XYZVector be usable here instead?

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+      bbox.min[2] = std::min(std::min(v1[2], v2[2]), v3[2]) - 0.001f;
+      bbox.max[0] = std::max(std::max(v1[0], v2[0]), v3[0]) + 0.001f;
+      bbox.max[1] = std::max(std::max(v1[1], v2[1]), v3[1]) + 0.001f;
+      bbox.max[2] = std::max(std::max(v1[2], v2[2]), v3[2]) + 0.001f;
+      return bbox;
+   };
+
+   // we need bounding boxes enclosing the primitives and centers of primitives
+   // (replaced here by centers of bounding boxes) to build the bvh
+   auto bboxes_ptr = new std::vector<BBox>();
```
dpiparo, line 1129: I do not understand this allocation. Why? Who deletes this memory?

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+   // fBoundingBoxes = (void *)bboxes_ptr;
```
dpiparo, line 1130: If we don't need this line, perhaps we can remove it instead of commenting?

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+   };
+
+   // let's do very quick checks against the top node
+   // is this useful for inside?
+   if (!TGeoBBox::Contains(point)) {
+      return false;
+   }
+
+   // doesn't need to be normalized and probes all normals
+   double test_dir[3] = {1.0, 1.41421356237, 1.73205080757};
```
dpiparo, line 1193: maybe a comment about these numbers is in order?

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+   double vol = 0.0;
+   for (int i = 0; i < fFacets.size(); ++i) {
+      auto &facet = fFacets[i];
+      auto a = fVertices[facet[0]];
+      auto b = fVertices[facet[1]];
+      auto c = fVertices[facet[2]];
+      vol +=
+         a[0] * (b[1] * c[2] - b[2] * c[1]) + b[0] * (c[1] * a[2] - c[2] * a[1]) + c[0] * (a[1] * b[2] - a[2] * b[1]);
+   }
+   return vol / 6.0;
```
dpiparo, line 1094: oppportunity to get rid of a division?

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+// structure keeping cost (value) for a BVH index
+struct BVHPrioElement {
+   size_t bvh_node_id;
+   float value;
+};
+
+// A priority queue for BVHPrioElement with an additional clear method
+// for quick reset
+template <typename Comparator>
+class BVHPrioQueue : public std::priority_queue<BVHPrioElement, std::vector<BVHPrioElement>, Comparator> {
```
dpiparo, line 1293: Can we find an alternative to subclassing stl containers?

dpiparo:
Thanks a lot for this PR. I left a few comments. In general, I would like to ask whether you think it's possible to add some test coverage for this new nice functionality. The coverage in the geometry package in general can be improved, and here we have an opportunity!

geom/geom/inc/LinkDef1.h
agheata: Yes, otherwise we won't be able to store this on disk. I'll check later, but @sawenzel mentioned a custom streamer that likely replaces the default one before and creates the bvh.

geom/geom/inc/TGeoTessellated.h
agheata: But of course, we already do I/O, and we must; it is just that the optimization structure is built on-the-fly. Ideally, this would also be streamed on disk. I guess this is not done yet because bvh v2 is not yet instrumented

geom/geom/inc/TGeoTessellated.h
```diff
@@ -129,6 +139,27 @@ class TGeoTessellated : public TGeoBBox {
...
+   Double_t DistFromInside(const Double_t *point, const Double_t *dir, Int_t iact = 1, Double_t step = TGeoShape::Big(),
+                           Double_t *safe = nullptr) const override;
+   bool Contains(const Double_t *point) const override;
+   Double_t Safety(const Double_t *point, Bool_t in = kTRUE) const override;
+   void ComputeNormal(const Double_t *point, const Double_t *dir, Double_t *norm) const override;
+
+   // these are trivial implementations, just for debugging
+   Double_t DistFromInside_Loop(const Double_t *point, const Double_t *dir) const;
+   Double_t DistFromOutside_Loop(const Double_t *point, const Double_t *dir) const;
+   bool Contains_Loop(const Double_t *point) const;
```
agheata, line 156: Since these are for testing purposes only, leaving them in the public interface may not be the best choice. The alternative could be to move these to a test utility, where we can add also some regression tests in future. What about `Safety_Loop` ?

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
    tsl->Print();
    return tsl;
 }
+
+// implementation of some geometry helper functions in anonymous namespace
+namespace {
```
agheata, line 731: `Tessellated::Vertex_t is eventually `TGeoVector3`, which already has these functions, so no need to re-implement them here

geom/geom/src/TGeoTessellated.cxx
agheata: It could, if we replace the `typedef Tessellated::Vertex_t` point to `ROOT::Math::XYZVector`, with probably some tweaks, because the interfaces are not identical. Probably this is desirable for future maintainability, but IMO can/should be done in a separate PR. And XYZVector could benefit from some simple additions, such as magnitude (now only squared available), or indexed access.

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+   auto v = dot(dir, q) * invDet;
+   if (v < 0.0 || u + v > 1.0) {
+      return INF;
+   }
+   auto t = dot(e2, q) * invDet;
+   return (t > rayEPS) ? t : INF;
+}
+
+template <typename T = float>
+struct Vec3f {
```
agheata, line 799: Why this helper, it duplicates stuff, can we use Vertex_t?

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+{
+   T len2 = dot(v, v);
+   if (len2 == T(0.0f)) {
+      std::cerr << "Degnerate triangle. Cannot determine normal";
+      return {0, 0, 0};
+   }
+   T invLen = T(1.0f) / std::sqrt(len2);
+   return {v.x * invLen, v.y * invLen, v.z * invLen};
+}
+
```
agheata, line 921: Vertex_t ?

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+/// DistFromOutside
+
+Double_t TGeoTessellated::DistFromOutside(const Double_t *point, const Double_t *dir, Int_t iact, Double_t stepmax,
+                                          Double_t *safe) const
+{
+   // use the BVH intersector in combination with leaf ray-triangle testing
+   double local_step = Big(); // we need this otherwise the lambda get's confused
+
+   using Scalar = float;
+   using Vec3 = bvh::v2::Vec<Scalar, 3>;
```
agheata, line 942: We will eventually need to make `bvh::v2::Vec` point to `XYZVector` as well; we now have a proliferation of vectors. I would go for making bvh::v2 use as much as possible ROOT native constructs, and enable persistency for it. But this in a separate PR.

geom/geom/inc/TGeoTessellated.h
```diff
@@ -65,13 +65,23 @@ class TGeoTessellated : public TGeoBBox {
...
    bool fClosedBody = false;        // The faces are making a closed body
+
+   // for now separate vectors but might be better to group per face
    std::vector<Vertex_t> fVertices; // List of vertices
    std::vector<TGeoFacet> fFacets;  // List of facets
+   std::vector<Vertex_t> fOutwardNormals; //! Vector of outward-facing normals
+
    std::multimap<long, int> fVerticesMap; //! Temporary map used to deduplicate vertices
+   bool fIsClosed = false;                //! to know if shape still needs closure/initialization
+   void *fBVH = nullptr;                  //! BVH acceleration structure for safety and navigation
```
agheata, line 77: `void *` is really bad for any future persistence of this structure, and I/O may become **very** relevant to avoid re-creating it for very large number of triangles. IMO we need to put here the real type, even if it makes the header directly dependent on bvh::v2.

The destructor of the class should delete fBVH

geom/geom/src/TGeoTessellated.cxx
agheata: It is assigned to fBVH, but not deleted in the dtor

geom/geom/src/TGeoTessellated.cxx
agheata: Right, we should strive to document better the implementation so that it will be easier to maintain. To start with, we can do it in a doxygen `@details` banner for the methods containing sensibly-complex algorithms

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+      return false;
+   }
+
+   // doesn't need to be normalized and probes all normals
+   double test_dir[3] = {1.0, 1.41421356237, 1.73205080757};
+
+   double local_step = Big();
+   // the ray used for bvh interaction
+   Ray ray(Vec3(point[0], point[1], point[2]),          // origin
+           Vec3(test_dir[0], test_dir[1], test_dir[2]), // direction
```
agheata, line 1198: Is this normalized anywhere?

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+
+// determines if a point is inside the bounding box
+template <typename T>
+bool contains(bvh::v2::BBox<T, 3> const &box, bvh::v2::Vec<T, 3> const &p)
+{
+   auto min = box.min;
+   auto max = box.max;
+   return (p[0] >= min[0] && p[0] <= max[0]) && (p[1] >= min[1] && p[1] <= max[1]) &&
+          (p[2] >= min[2] && p[2] <= max[2]);
+}
```
agheata, line 1236: It makes sense to put this in `bvh::v2`?

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+      const auto &v1 = fVertices[facet[1]];
+      const auto &v2 = fVertices[facet[2]];
+
+      const double t = rayTriangle(p, test_dir, v0, v1, v2, 0.);
+      if (t != std::numeric_limits<double>::infinity()) {
+         ++crossings;
+      }
+   }
+   return (crossings & 1);
+}
```
agheata, line 1527: Candidates for a test utility, no need to keep these here

sawenzel:
@dpiparo @agheata :Thanks for initial review. I'll take a look and see what can be addressed with reasonable effort.

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +721,818 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+   using Scalar = float;
+   using BBox = bvh::v2::BBox<Scalar, 3>;
+   using Vec3 = bvh::v2::Vec<Scalar, 3>;
+   using Node = bvh::v2::Node<Scalar, 3>;
+   using Bvh = bvh::v2::Bvh<Node>;
+
+   // helper determining axis aligned bounding box from a facet;
+   auto GetBoundingBox = [this](TGeoFacet const &facet) {
+      const auto nvertices = facet.GetNvert();
+      assert(nvertices == 3); // for now only triangles
```
agheata: We need to support both 3 and 4-vertex facets.

geom/geom/inc/TGeoTessellated.h
agheata: Now the BVH is streamable to disk

geom/geom/inc/TGeoTessellated.h
silverweed: If we put the real type, can we also use a `unique_ptr` while we're at it?

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +726,867 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+   // we need bounding boxes enclosing the primitives and centers of primitives
+   // (replaced here by centers of bounding boxes) to build the bvh
+   auto bboxes_ptr = new std::vector<BBox>();
+   // fBoundingBoxes = (void *)bboxes_ptr;
+   auto &bboxes = *bboxes_ptr;
+   std::vector<Vec3> centers;
+
+   // loop over all the triangles/Facets;
+   int nd = fFacets.size();
+   for (int i = 0; i < nd; ++i) {
```
silverweed, line 1136: Since we know how many elements we're gonna push in:

```suggestion
   centers.reserve(nd);
   bboxes.reserve(nd);
   for (int i = 0; i < nd; ++i) {
```

geom/geom/src/TGeoTessellated.cxx
```diff
@@ -703,3 +726,867 @@ TGeoTessellated *TGeoTessellated::ImportFromObjFormat(const char *objfile, bool
...
+
+   // create the bvh
+   typename bvh::v2::DefaultBuilder<Node>::Config config;
+   config.quality = bvh::v2::DefaultBuilder<Node>::Quality::High;
+   auto bvh = bvh::v2::DefaultBuilder<Node>::build(bboxes, centers, config);
+   auto bvhptr = new Bvh;
+   *bvhptr = std::move(bvh); // copy structure
+   fBVH = (void *)(bvhptr);
+
+   return;
```
silverweed, line 1158:
```suggestion
```

geom/geom/inc/bvh/v2/bbox.h
pcanal, line 1: If this file is only for internal use (used only by our sources and not by user code), it should be moved to `geom/geom/src/`.  If it needs to be used from multiple ROOT packages, it should move to `geom/geom/res`

geom/geom/inc/bvh/v2/bbox.h
```diff
@@ -10,38 +10,39 @@ namespace bvh::v2 {
 
 template <typename T, size_t N>
 struct BBox {
-    Vec<T, N> min, max;
-
-    BBox() = default;
-    BVH_ALWAYS_INLINE BBox(const Vec<T, N>& min, const Vec<T, N>& max) : min(min), max(max) {}
```
pcanal, line 16: If this file is coming from a 3rd party package/provider, we should:
- avoid to make NFC changes (include no white changes) so that we can import new version easier
- have the provider include into the file a version number so that it is easier to keep track of the irregular imports.

sawenzel:
Just a quick note: since additional commits have been pushed on top of my original PR (including additional features but also formatting and whitespace changes), I’ll pause on making further updates for now until we align on the best way forward.

From my perspective, there are two reasonable options: (a) I can try to integrate everything and we squash all into a single clean commit and consolidate the changes, or (b) We separate my original PR from the additional features introduced by @agheata, and integrate them stepwise.

In either case, I would be in favour of minimizing changes to the external BVH files and keeping reformatting to a minimum to make the review process easier. We could also control compiler warnings for the BVH files by introducing an additional single-entry header `bvh_external.h` which bundles headers together and may control compiler warnings for this external entity.

agheata:
> From my perspective, there are two reasonable options: (a) I can try to integrate everything and we squash all into a single clean commit and consolidate the changes, or (b) We separate my original PR from the additional features introduced by @agheata, and integrate them stepwise.

I would rather go for (b), probably the easiest will be to make a new branch with your original PR, adding bvh_external.h and the changes needed to avoid the compilation error/warnings, then I can make a PR on top of that, excluding the changes done to bvh::v2 and whitespace changes?
> 
> In either case, I would be in favour of minimizing changes to the external BVH files and keeping reformatting to a minimum to make the review process easier. We could also control compiler warnings for the BVH files by introducing an additional single-entry header `bvh_external.h` which bundles headers together and may control compiler warnings for this external entity.

Note that keeping this package as "external" and not touching it also prevents doing native ROOT I/O (or at least I don't know how to do it). Making the BVH data structure persistent IS important for performance in complex cases, so I added a simple mechanism to do it in one of the commits, which should also deal with schema changes in `bvh::v2`. I am not sure, however, that this is the best solution, @pcanal can you have a look at my commit adding blob persistence and comment?

pcanal:
> Note that keeping this package as "external" and not touching it also prevents doing native ROOT I/O 

It should not be a show stopper.  In most case you can just generate the dictionary for un-instrustmented classes.  If the transient member can not be marked as such in the header files (by just adding a comment) then there is still some way to register them as such in the dictionary request file (eg. selection.xml and maybe linkdef)

geom/geom/inc/bvh/v2/blob_persistence.h
```diff
@@ -0,0 +1,212 @@
...
+#include <cstring>
+#include <type_traits>
+#include <vector>
+
+namespace bvh::v2 {
+
+// BVH blob persistence helpers.
+//
+// Rationale:
+// - "Blob" means a compact byte buffer storing serialized BVH data.
```
pcanal, line 17: Is the resulting of this hand streaming 'better' than just using regular I/O?  If yes in what 'dimensions', if not what prevented the use of the regular I/O?

agheata: I don't know how to enable the regular I/O on this foreign data structure (the `void *fBVH` data member of `TGeoTessellated`) and this is the only reason for the hand streaming. Any suggestions?

pcanal: > (the void *fBVH data member of TGeoTessellated)

I see.  It is not so much that it is 'foreign' but rather that it is 'opaque'.  Indeed in order for the automatic I/O one would need to provided the actual type being pointed (and the dictionary would have to requested for those types).

geom/geom/inc/TGeoTessellated.h
sawenzel: I have removed them in the updated PR #21409.

geom/geom/src/TGeoTessellated.cxx
sawenzel: removed in updated version

geom/geom/src/TGeoTessellated.cxx
sawenzel: cleaned up in updated version.

geom/geom/src/TGeoTessellated.cxx
sawenzel: I would leave this as is. It is readable and this function is not performance critical.

geom/geom/src/TGeoTessellated.cxx
sawenzel: allocation has been removed

geom/geom/src/TGeoTessellated.cxx
sawenzel: suggestion taken in updated version

geom/geom/src/TGeoTessellated.cxx
sawenzel: added a comment in updated version

geom/geom/src/TGeoTessellated.cxx
sawenzel: doesn't need to be normalized for Contains, since never calculating a distance (just number of crossings).

geom/geom/src/TGeoTessellated.cxx
sawenzel: yes, I factored these functions into header for modular use in the updated version

geom/geom/src/TGeoTessellated.cxx
sawenzel: I did not find a more convenient/shorter implementation. I would like to keep it as is. In any case, this is never publicly exposed and an internal detail of the implementation in the source file (in the same way as done in TGeoParallelWorld). I added the following statement to the code in the updated version:
"
We intentionally derive from std::priority_queue here to expose a
// clear() convenience method via access to the protected container `c`.
// This is internal, non-polymorphic code and relies on standard-library
// implementation details that are stable across supported platforms.
"

agheata:
Superseded by https://github.com/root-project/root/pull/21409

PR #21047, opened 2026-01-27
Geom color scheme

agheata:
# This Pull request:
Introduces an extensible color scheme mechanism for geometry visualization, improving the default coloring of imported geometries (notably GDML, which does not store color information).

## Changes or fixes:
TGeoManager::DefaultColors() historically assigns colors based only on material effective Z, resulting in:
- many volumes sharing the same gray color,
- limited visual distinction,
- little resemblance to “natural” material appearance in detector geometries.
tpc_none

Modern detector descriptions contain rich material names (e.g. Copper, FR4, Kapton, Steel, Carbon fiber, Gases), which can be exploited to produce clearer and more meaningful visualizations.

## What’s new
 - A new **runtime-only strategy class** `TGeoColorScheme` that encapsulates color and transparency policies.
 - A default **“natural”** color scheme based on:
   - name-based material classification (metals, polymers, composites, gases, ceramics, etc.),
   - fallback to a Z-binned lookup when no name-based rule applies.
   
tpc_natural

 - Two additional built-in schemes:
   - **Flashy** – high-contrast, presentation-friendly colors.
 
tpc_flashy

   - **High-contrast** – darker, saturated colors suited for light backgrounds.
tpc_high_contrast

 - **User extensibility:**
   - users can inject custom logic via hooks (`std::function`) to override colors or transparency,
   - or subclass `TGeoColorScheme` for full control.

 - **Backward compatibility:**
   - calling `TGeoManager::DefaultColors()` with no arguments still works and now defaults to the natural scheme.

## API overview
```cpp
void TGeoManager::DefaultColors(const TGeoColorScheme *scheme = nullptr);
```

if `scheme == nullptr` a default `TGeoColorScheme(EGeoColorSet::kNatural)` is used internally.
The color set is selected via a global enum:
```cpp
enum class EGeoColorSet {
   kNatural,
   kFlashy,
   kHighContrast
};
```

## Examples
- Default behavior (no code changes required):
```cpp
gGeoManager->DefaultColors();   // uses natural, name-based scheme
```

- Select a predefined scheme
```cpp
TGeoColorScheme cs(EGeoColorSet::kFlashy);
gGeoManager->DefaultColors(&cs);
```

- User override via hook (no subclassing)
```cpp
TGeoColorScheme cs(EGeoColorSet::kNatural);
cs.SetColorHook([](const TGeoVolume *v) -> Int_t {
   const TGeoMaterial *m = TGeoColorScheme::GetMaterial(v);
   if (!m || !m->GetName()) return -1;
   if (std::string(m->GetName()).find("Copper") != std::string::npos)
      return TColor::GetColor(0.9f, 0.55f, 0.3f);
   return -1; // fallback to defaults
});
gGeoManager->DefaultColors(&cs);
```

- Override Z-based fallback
```cpp
cs.SetZFallbackHook([](Int_t Z, EGeoColorSet) -> Int_t {
   float g = std::min(1.f, Z / 100.f);
   return TColor::GetColor(g, g, g);
});
```

## Checklist:

- [x] tested changes locally

This PR fixes #

### DIFF
--- a/README/ReleaseNotes/v640/index.md
+++ b/README/ReleaseNotes/v640/index.md
@@ -17,6 +17,7 @@ The following people have contributed to this new version:
  Olivier Couet, CERN/EP-SFT,\
  Marta Czurylo, CERN/EP-SFT,\
  Florine de Geus, CERN/EP-SFT and University of Twente,\
+ Andrei Gheata, CERN/EP-SFT,\
  Jonas Hahnfeld, CERN/EP-SFT and Goethe University Frankfurt,\
  Fernando Hueso Gonzalez, IFIC (CSIC-University of Valencia),\
  Stephan Hageboeck, CERN/EP-SFT,\
@@ -61,6 +62,40 @@ The following people have contributed to this new version:
 - ROOT now adds a RUNPATH to compiled macros. This ensures that when compiled macros are loaded, they load the libraries that belong to the ROOT installation
   that compiled the macro. See [TSystem::SetMakeSharedLib()](https://root.cern.ch/doc/master/classTSystem.html#a80cd12e064e2285b35e9f39b5111d20e) for
   customising or disabling the RUNPATH.
+
+## Geometry
+
+### Extensible color schemes for geometry visualization
+ROOT now provides an extensible mechanism to assign colors and transparency to geometry volumes via the new `TGeoColorScheme` strategy class, used by `TGeoManager::DefaultColors()`.
+
+This improves the readability of geometries imported from formats such as GDML that do not store volume colors. The default behavior now uses a name-based material classification (e.g. metals, polymers, composites, gases) with a Z-binned fallback. Three predefined color sets are provided:
+* `EGeoColorSet::kNatural` (default): material-inspired colors
+* `EGeoColorSet::kFlashy`: high-contrast, presentation-friendly colors
+* `EGeoColorSet::kHighContrast`: darker, saturated colors suited for light backgrounds
+
+Users can customize the behavior at runtime by providing hooks (std::function) to override the computed color, transparency, and/or the Z-based fallback mapping.
+
+**Usage examples:**
+```cpp
+gGeoManager->DefaultColors(); // default (natural) scheme
+
+TGeoColorScheme cs(EGeoColorSet::kFlashy);
+gGeoManager->DefaultColors(&cs); // select a predefined scheme
+```
+
+**Override examples (hooks):**
+```cpp
+TGeoColorScheme cs(EGeoColorSet::kNatural);
+cs.SetZFallbackHook([](Int_t Z, EGeoColorSet) -> Int_t {
+   float g = std::min(1.f, Z / 100.f);
+   return TColor::GetColor(g, g, g); // grayscale fallback
+});
+gGeoManager->DefaultColors(&cs);
+```
+
+A new tutorial macro demonstrates the feature and customization options: `tutorials/visualization/geom/geomColors.C`.
+
+See: https://github.com/root-project/root/pull/21047 for more details
 
 ## I/O
 
--- a/tutorials/visualisation/geom/geomColors.C
+++ b/tutorials/visualisation/geom/geomColors.C
@@ -1,0 +1,180 @@
+/**
+ * \file geomColors.C
+ * \ingroup tutorial_geom
+ * Script demonstrating geometry color schemes.
+ *
+ *
+ * \macro_code
+ *
+ * \author andrei.gheata@cern.ch
+ */
+
+#include <string>
+#include <algorithm>
+
+#include "TROOT.h"
+#include "TColor.h"
+#include "TControlBar.h"
+#include "TGeoManager.h"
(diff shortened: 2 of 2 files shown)


geom/geom/inc/LinkDef1.h
```diff
@@ -101,6 +101,7 @@
 #pragma link C++ class TGeoNavigator + ;
 #pragma link C++ class TGeoNavigatorArray;
 #pragma link C++ class TGDMLMatrix + ;
+#pragma link C++ class TGeoColorScheme;
```
dpiparo: Do we need to select the class or can we work without the dictionary? Do we need IO?

geom/geom/inc/TGeoColorScheme.h
```diff
@@ -0,0 +1,68 @@
...
+   /// Return >=0 to apply a ROOT color index (TColor index), <0 for "no decision".
+   virtual Int_t Color(const TGeoVolume *vol) const;
+
+   /// Return [0..100] to apply transparency, <0 to leave unchanged.
+   virtual Int_t Transparency(const TGeoVolume *vol) const;
+
+   /// Z-binned fallback. Users may override via SetZFallbackHook() or subclassing.
+   virtual Int_t ColorForZ(Int_t Z, EGeoColorSet set) const;
+
+   /// Hooks setters. Set to nullptr to disable.
```
dpiparo: The comment will not propagate well in the Doxygen doc. Can it be perhaps specialised for each method?

geom/geom/src/TGeoColorScheme.cxx
```diff
@@ -0,0 +1,529 @@
+//==============================================================================
```
dpiparo, line 1: This documentation looks *great*. I propose we keep a close eye to it to check how this is rendered and if something was not formatted properly.

geom/geom/src/TGeoColorScheme.cxx
```diff
@@ -0,0 +1,529 @@
...
+  - The default name token rules are heuristic and aimed at typical HEP
+    detector material naming conventions; users can refine them locally via
+    hooks or subclassing.
+
+------------------------------------------------------------------------------
+*/
+
+#include "TGeoColorScheme.h"
+
+#include <string>
```
dpiparo, line 96: Do we prefer standard includes after ROOT's?

geom/geom/src/TGeoManager.cxx
```diff
@@ -2258,42 +2259,30 @@ Int_t TGeoManager::GetSafeLevel() const
...
-   col[10] = col[11] = kRed - 10;
-   col[12] = kGray + 1;
-   col[13] = kBlue - 10;
-   col[14] = kOrange + 7;
-   col[16] = kYellow + 1;
-   col[20] = kYellow - 10;
-   col[24] = col[25] = col[26] = kBlue - 8;
-   col[29] = kOrange + 9;
-   col[79] = kOrange - 2;
+void TGeoManager::DefaultColors(const TGeoColorScheme *cs /*=nullptr*/)
```
dpiparo, line 2266: Can't we pass by reference or something more modern than the bare ptr?

geom/geom/inc/LinkDef1.h
agheata: No I/O needed, just needed that the interpreter "sees" the class, which I realized it does w/o creating the dictionary, so this is a leftover of my attempt to make the interpreter autocomplete the enum defined in the header file. I also realized that ROOT cannot do that yet. Removing.

geom/geom/inc/TGeoColorScheme.h
agheata: Hmm, are standard doxygen comments OK (like @brief, @param, ...) ?- I was just following the general style in TGeo which was so far **NOT** doxygen.

geom/geom/src/TGeoManager.cxx
agheata: We need backward compatibility for the method, so the parameter must take a default value. const pointer is fine here as compromise, as it implies no lifetime ownership

geom/geom/inc/TGeoColorScheme.h
agheata: Done

geom/geom/src/TGeoColorScheme.cxx
agheata: I converted to doxygen style as well

agheata:
Adding a tutorial for this, and release notes

agheata:
> Thanks for this solid work.
> 
> Besides the minor comments on the code, I see an opportunity for improvement, i.e. the possibility to add an example in the geometry tutorials and tests. Presently, the test coverage of the geometry package can be improved, and this seems like a perfect opportunity to kick off this process!
> 
> These changes are impactful and would also deserve some words (and a beautiful image, like "before and after") in the Release Notes.

I added a nice tutorial and the entry in the release notes, w/o a picture (don't know where to put it) but a link to the PR. A coverage test for this is difficult to implement because it is pure visualization, but a working tutorial is more meaningful.

PR #21048, opened 2026-01-27
[tmva][sofie] Add support for parsing external ONNX data file for weights

lmoneta:
Support external_data when parsing the initialized tensor from an ONNX file To test the feture, run tutorial RMVA_SOFIE_ONNX.py changing when exporting to the ONNX file from external_data=False to external_data=True

### DIFF
--- a/tmva/sofie_parsers/src/RModelParser_ONNX.cxx
+++ b/tmva/sofie_parsers/src/RModelParser_ONNX.cxx
@@ -120,61 +120,129 @@ struct ExtractDataFromTP {
 // trait function to extract data from TensorProto
 template<>
 struct ExtractDataFromTP<float> {
-   static void Copy(onnx::TensorProto * tensor, void * data) {
+   static void Copy(onnx::TensorProto * tensor, void * data, int length) {
+      if (tensor->float_data_size() != length)
+         throw std::runtime_error("TMVA::SOFIE - Failed to read float initialized tensor - actual size is " + std::to_string(tensor->float_data_size()));
       tensor->mutable_float_data()->ExtractSubrange(0, tensor->float_data_size(),
                                                             static_cast<float *>(data));
    }
 };
 template<>
 struct ExtractDataFromTP<double> {
-   static void Copy(onnx::TensorProto * tensor, void * data) {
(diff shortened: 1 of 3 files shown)


tmva/sofie_parsers/src/RModelParser_ONNX.cxx
```diff
@@ -541,56 +616,41 @@ void RModelParser_ONNX::ParseONNXGraph(RModel & rmodel, const onnx::GraphProto &
...
+
+      std::shared_ptr<void> data = GetInitializedTensorData(tensorproto, tensor_length * GetTypeSize(tensor_type), tensor_type);
+      rmodel.AddInitializedTensor(tensor_name, tensor_type, shape, data);
+      allInitializedTensors[tensor_name] = i;
+
+      if (verbose) {
+         std::cout << "add initialized tensor " << tensor_name << "with shape " << ConvertShapeToString(shape) << "and  ";
+         if (tensor_type == ETensorType::FLOAT) {
+            std::cout << " float data: ";
+            for (int i = 0; i < std::min(int(tensor_length),3); i++) std::cout << static_cast<float*>(data.get())[0] << "  ";
```
sanjibansg: As the warning says, probably better to use something else than `i` here for the for loop variable, since this shadows the previous one which might cause errors in the future.

tmva/sofie_parsers/src/RModelParser_ONNX.cxx
```diff
@@ -541,56 +616,41 @@ void RModelParser_ONNX::ParseONNXGraph(RModel & rmodel, const onnx::GraphProto &
...
+         }
+         else if (tensor_type == ETensorType::INT64) {
+            std::cout << " int64 data: ";
+            for (int i = 0; i < std::min(int(tensor_length),3); i++) std::cout << static_cast<int64_t*>(data.get())[0] << "  ";
```
sanjibansg: Same comment on using a different loop variable here.

guitargeek:
This is great! It tripped me up several times that you get silently wrong results with SOFIE when you use external ONNX data files.

Is there anything blocking this PR?

sanjibansg (approved):
LGTM! Thanks for this implementation.

PR #21051, opened 2026-01-27
[ntuple] More preparatory changes for RNTupleAttributes

silverweed:
As per the PR title, some small changes that are needed for the RNTupleAttribute writing (full reference PR: #19153).

See the individual commits for the changes.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/inc/ROOT/RNTupleFillContext.hxx
+++ b/tree/ntuple/inc/ROOT/RNTupleFillContext.hxx
@@ -100,7 +100,7 @@ private:
    std::size_t FillImpl(Entry &entry)
    {
       ROOT::RNTupleFillStatus status;
-      FillNoFlush(entry, status);
+      FillNoFlushImpl(entry, status);
       if (status.ShouldFlushCluster())
          FlushCluster();
       return status.GetLastEntrySize();


tree/ntuple/inc/ROOT/RNTupleFillContext.hxx
```diff
@@ -109,6 +109,8 @@ private:
    RNTupleFillContext(std::unique_ptr<ROOT::RNTupleModel> model, std::unique_ptr<ROOT::Internal::RPageSink> sink);
    RNTupleFillContext(const RNTupleFillContext &) = delete;
    RNTupleFillContext &operator=(const RNTupleFillContext &) = delete;
+   RNTupleFillContext(RNTupleFillContext &&) = default;
+   RNTupleFillContext &operator=(RNTupleFillContext &&) = default;
```
hahnjo: Why is this needed? I believe we should rather forbid moving the `RNTupleFillContext` out of its `std::unique_ptr` shell, and the same for `RNTupleWriter` and `RNTupleReader`...

silverweed: Neither `RNTupleWriter` nor the upcoming `RNTupleAttrSetWriter` wrap `RNTupleFillContext` in a unique_ptr, and if it's allowed to have it by value it stands to reason to me that you should be able to move it (unless there are good reason not to).
At the moment it's not strictly needed (I could probably work around it by emplacing it rather than moving it) but I'm not sure why we should prevent it.

hahnjo: You cannot move `RNTupleWriter`, currently the constructor is implicitly deleted (and I think we should make it explicit). Otherwise you could do things like
```c++
std::unique_ptr<ROOT::RNTupleWriter> writer = /* ... */;
ROOT::RNTupleWriter stack = std::move(*writer.release());
```

silverweed: I understand this is a problem for the Writer, but why is it also for the fill context? And if it is, shouldn't we prevent constructing it on the stack in the first place?

hahnjo: It's part of the classes where we expect objects to only appear in smart pointers. For the fill context, only the `RNTupleWriter` has a direct member and `RNTupleParallelWriter` can create `std::shared_ptr<RNTupleFillContext>`. This is required for the warning in `~RNTupleParallelWriter` to work. The user is unable to create fill contexts on the stack.

PR #21053, opened 2026-01-28
[cppyy] Check early for conversion to initializer list

hahnjo:
It requires to construct default objects. Before it would blindly call `Cppyy::Construct` and check the return value. This leads to a confusing error message from `TClass::New` even though conversion to `std::vector` subsequently succeeds (which doesn't need to construct default objects).

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
@@ -16,6 +16,7 @@
 #include "Utility.h"
 
 // Standard
+#include <cassert>
 #include <complex>
 #include <limits.h>
 #include <stddef.h>      // for ptrdiff_t
@@ -3111,7 +3112,13 @@ bool CPyCppyy::InitializerListConverter::SetArg(
     // so either.
         PyErr_Clear();
 
-    // can only construct empty lists, so use a fake initializer list
+    // Can only construct empty lists, so use a fake initializer list. For that we
+    // need to construct default objects. Fail early if that cannot work.
+        if (fValueType && !Cppyy::IsDefaultConstructable(fValueType)) {
+            PyErr_SetString(PyExc_TypeError, "default constructor needed for initializer list of objects");
+            return false;
+        }
+
         size_t len = (size_t)PySequence_Size(pyobject);
         fake = (faux_initlist*)malloc(sizeof(faux_initlist)+fValueSize*len);
         fBuffer = (void*)fake;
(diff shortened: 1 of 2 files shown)


hahnjo:
Obvious follow-up question: why do we need to default construct objects for conversion to initializer list, but not for vector? But since I'm ok with the vector overload for my case in the histograms, I'm not going to look further into this...

bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
```diff
@@ -3111,7 +3112,13 @@ bool CPyCppyy::InitializerListConverter::SetArg(
     // so either.
         PyErr_Clear();
 
-    // can only construct empty lists, so use a fake initializer list
+        // Can only construct empty lists, so use a fake initializer list. For that we need to construct default
+        // objects. Fail early if that cannot work.
```
guitargeek:
```suggestion
    // Can only construct empty lists, so use a fake initializer list. For that we need to construct default
    // objects. Fail early if that cannot work.
```
Sorry for the nitpick: so far we tried to respect Wims indentation convention to minimize divergence in case our changes also get picked up by Wim. Could we also do that in this PR?

bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
```diff
@@ -3142,11 +3149,9 @@ bool CPyCppyy::InitializerListConverter::SetArg(
                     // clunky, but the use of a copy constructor isn't much better as the Python object
                     // need not be a C++ object
                         memloc = (void*)Cppyy::Construct(fValueType, memloc);
-                        if (memloc) entries += 1;
-                        else {
-                           PyErr_SetString(PyExc_TypeError,
-                              "default ctor needed for initializer list of objects");
-                        }
+                        // We checked above that we are able to construct default objects of fValueType.
```
guitargeek:
```suggestion
                    // We checked above that we are able to construct default objects of fValueType.
```
Dito

bindings/pyroot/cppyy/CPyCppyy/src/Converters.cxx
```diff
@@ -3111,7 +3112,13 @@ bool CPyCppyy::InitializerListConverter::SetArg(
     // so either.
         PyErr_Clear();
 
-    // can only construct empty lists, so use a fake initializer list
+        // Can only construct empty lists, so use a fake initializer list. For that we need to construct default
+        // objects. Fail early if that cannot work.
+        if (fValueType && !Cppyy::IsDefaultConstructable(fValueType)) {
+            PyErr_SetString(PyExc_TypeError, "default ctor needed for initializer list of objects");
```
guitargeek:
```suggestion
            PyErr_SetString(PyExc_TypeError, "default constructor needed for initializer list of objects");
```
If we are already moving the code, this could be an opportunity to avoid some jargon.

guitargeek (approved):
Thanks! I only have minor formatting comments, but the changes look perfect to me!

> Obvious follow-up question: why do we need to default construct objects for conversion to initializer list, but not for vector?

I don't know, but since the changes in this PR make sense no matter what the answer to that question is, I can look into this later.

PR #21054, opened 2026-01-28
[hist] Test Python interface of core classes

hahnjo:
Construction from a list of axes will currently print a confusing error:
```
Error in <TClass::New with placement>: cannot create object of class ROOT::Experimental::RAxisVariant version 1 at address 0x55c093cb7000
```

This is harmless since the `std::vector` overload is chosen afterwards, see https://github.com/root-project/root/pull/21053.

### DIFF
--- a/hist/histv7/test/hist.py
+++ b/hist/histv7/test/hist.py
@@ -1,0 +1,467 @@
+import math
+import unittest
+
+import ROOT
+
+RAxisVariant = ROOT.Experimental.RAxisVariant
+RBinIndex = ROOT.Experimental.RBinIndex
+RCategoricalAxis = ROOT.Experimental.RCategoricalAxis
+RHist = ROOT.Experimental.RHist
+RHistEngine = ROOT.Experimental.RHistEngine
+RHistStats = ROOT.Experimental.RHistStats
+RRegularAxis = ROOT.Experimental.RRegularAxis
+RVariableBinAxis = ROOT.Experimental.RVariableBinAxis
+RWeight = ROOT.Experimental.RWeight
+
+
+class RegularAxis(unittest.TestCase):
+    def test_init(self):
+        Bins = 20
+        axis = RRegularAxis(Bins, (0, Bins))
+        self.assertEqual(axis.GetNNormalBins(), Bins)
+        self.assertEqual(axis.GetTotalNBins(), Bins + 2)
+        self.assertEqual(axis.GetLow(), 0)
+        self.assertEqual(axis.GetHigh(), Bins)
+        self.assertTrue(axis.HasFlowBins())
+
+        axis = RRegularAxis(Bins, (0, Bins), enableFlowBins=False)
+        self.assertEqual(axis.GetNNormalBins(), Bins)
+        self.assertEqual(axis.GetTotalNBins(), Bins)
+        self.assertFalse(axis.HasFlowBins())
+
+    def test_GetNormalRange(self):
+        Bins = 20
+        axis = RRegularAxis(Bins, (0, Bins))
+        self.assertEqual(len(list(axis.GetNormalRange())), Bins)
+        self.assertEqual(len(list(axis.GetNormalRange(1, 5))), 4)
+
+    def test_GetFullRange(self):
+        Bins = 20
+        axis = RRegularAxis(Bins, (0, Bins))
+        self.assertEqual(len(list(axis.GetFullRange())), Bins + 2)
+
+
+class VariableBinAxis(unittest.TestCase):
+    def test_init(self):
+        Bins = 20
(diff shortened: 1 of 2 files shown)


hist/histv7/test/hist.py
```diff
@@ -0,0 +1,455 @@
...
+    def test_init(self):
+        Bins = 20
+        axis = RRegularAxis(Bins, (0, Bins))
+        self.assertEqual(axis.GetNNormalBins(), Bins)
+        self.assertEqual(axis.GetTotalNBins(), Bins + 2)
+        self.assertEqual(axis.GetLow(), 0)
+        self.assertEqual(axis.GetHigh(), Bins)
+        self.assertTrue(axis.HasFlowBins())
+
+        axis = RRegularAxis(Bins, (0, Bins), enableFlowBins=False)
```
hageboeck, line 27: Is (and should) a list be supported, too?
```diff
- axis = RRegularAxis(Bins, (0, Bins), enableFlowBins=False)
+ axis = RRegularAxis(Bins, [0, Bins], enableFlowBins=False)
```

hist/histv7/test/hist.py
```diff
@@ -0,0 +1,455 @@
...
+        self.assertTrue(axis.HasFlowBins())
+
+        axis = RVariableBinAxis(bins, enableFlowBins=False)
+        self.assertEqual(axis.GetNNormalBins(), Bins)
+        self.assertEqual(axis.GetTotalNBins(), Bins)
+        self.assertFalse(axis.HasFlowBins())
+
+    def test_GetNormalRange(self):
+        Bins = 20
+        bins = [i for i in range(0, Bins + 1)]
```
hageboeck, line 72: I don't expect any problems, but could we maybe construct it from something like `i + 0.5` or so, so we can test that the floating-point values survive until they reach the C++ API?

hist/histv7/test/hist.py
```diff
@@ -0,0 +1,455 @@
...
+    def test_GetFullRange(self):
+        Bins = 20
+        bins = [i for i in range(0, Bins + 1)]
+        axis = RVariableBinAxis(bins)
+        self.assertEqual(len(list(axis.GetFullRange())), Bins + 2)
+
+
+class CategoricalAxis(unittest.TestCase):
+    def test_init(self):
+        categories = ["a", "b", "c"]
```
hageboeck, line 86: What happens when we pass:
```py
categories = [3, 2, 1]
```
?

hist/histv7/test/hist.py
```diff
@@ -0,0 +1,455 @@
...
+
+class HistStats(unittest.TestCase):
+    def test_init(self):
+        stats = RHistStats(1)
+        self.assertEqual(stats.GetNDimensions(), 1)
+
+    def test_Fill(self):
+        stats = RHistStats(3)
+        self.assertEqual(stats.GetNEntries(), 0)
+        self.assertTrue(math.isnan(stats.ComputeNEffectiveEntries()))
```
hageboeck, line 282: This could return `0.` without being inconsistent. Maybe we should keep the code safe and simple, though, so if it's making things complicated, please ignore the comment.

hist/histv7/test/hist.py
hahnjo: I would have guessed that it's supported, but I'm wrong: cppyy doesn't convert `list` into `std::pair`. I would anyway argue that a `tuple` feels more natural. As a user, would you expect it to work with a `list`?

hist/histv7/test/hist.py
hahnjo: Yes, good point 👍 I added another constructor in `test_init`, including checks for the bin edge values that I was missing so far

hist/histv7/test/hist.py
hahnjo: cppyy will complain that it cannot convert the first argument to `vector<string>`

hist/histv7/test/hist.py
hahnjo: True, I changed this in https://github.com/root-project/root/pull/20851 even though the denominator `fSumW2` doesn't actually suffer from problems with negative weights. We could consider returning `0.` again, I'll note it down as a question.

hist/histv7/test/hist.py
hageboeck: > As a user, would you expect it to work with a list?

No, I just wanted to make sure that it doesn't do "wrong" things. If it generates an error, we are on the safe side. 👍

hist/histv7/test/hist.py
hageboeck: OK, that's something the user can work with. Resolving

PR #21056, opened 2026-01-28
 [tutorials][test] Specify the include path for SOFIE tutorials in CMake

guitargeek:
We don't want to rely on environment variables like `ROOTSYS` in the tutorials, and we should also not leak details of the test setup like `$ROOTSYS/runtutorials` in the user-facing tutorials.

To make tutorial environment variables also work on Windows when they include paths with multiple elements, the first commit avoids the use of the `ENVIRONMENT` argument for the ROOT test macros, because it doesn't support multiple path elements on Windows.

This is a spinoff of https://github.com/root-project/root/pull/21018.


tutorials/machine_learning/TMVA_SOFIE_RDataFrame.C
```diff
@@ -25,7 +25,6 @@ using namespace TMVA::Experimental;
 // need to add the current directory (from where we are running this macro)
 // to the include path for Cling
 R__ADD_INCLUDE_PATH($PWD)
-R__ADD_INCLUDE_PATH($ROOTSYS/runtutorials)
```
pcanal, line 28: If the user wants to run the tutorials by hand (not through ctest), what do they need to do? (I.e. it seems that the question would also has been relevant with the current code)

tutorials/CMakeLists.txt
```diff
@@ -1035,11 +1044,11 @@ if(pyroot)
     endif()
 
     ROOT_ADD_TEST(${tutorial_name}
-                COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/${t}
+                COMMAND ${CMAKE_COMMAND} -E env ${TUTORIAL_ENV} ${ROOT_MAX_THREADS}
+                        ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/${t}
                 PASSRC ${rc} FAILREGEX "Error in" ": error:" "segmentation violation"
                 LABELS ${labels}
                 DEPENDS ${tutorial_dependency}
-                ENVIRONMENT ${TUTORIAL_ENV} ${ROOT_MAX_THREADS}
```
pcanal, line 1042: There are other use of `ENVIRONMENT` elsewhere, most of them look fine but the one in `roottest/root/io/transient/base/CMakeLists.txt`, `roottest/root/ntuple/makeproject/rntuple/CMakeLists.txt` and `roottest/root/ntuple/makeproject/ttree/CMakeLists.txt` seem potentially troublesome (they deal with paths).

tutorials/machine_learning/TMVA_SOFIE_RDataFrame.C
guitargeek: The user needs to first run the `TMVA_Higgs_Classification.C` and `TMVA_SOFIE_Keras_HiggsModel.C` tutorials in the same directory, to get the header file produced by TMVA SOFIE. That's also explained in the beginning of the tutorial.

tutorials/CMakeLists.txt
guitargeek: Good remark! In the end I went back to using the `ENVIRONMENT` argument, because also in the `CMD`, the semicolons get replaced when serializing the RooTestDriver arguments :( So I might as well stick with `ENVIRONMENT` for now and do some aggressive escaping so that the semicolons survive all the steps of string parsing that are going on in `ROOT_ADD_TEST` and `RootTestDriver.cmake`.

We can think of how to solve this more elegantly later, also for the cases you point out.

For now we were just lucky that path lists with multiple entries were never required for the test environments on Windows.

pcanal: Thanks.

PR #21059, opened 2026-01-28
Add tests for analytical TF1 integrals

swetank18:
### Motivation
Recent changes improved robustness of analytical TF1 integrals for edge cases
such as vanishing exponential slopes and invalid parameter domains.

### Changes
- Add test covering analytical exponential integrals with p1 == 0
- Add test verifying invalid sigma handling for Gaussian integrals

### Impact
- Prevents regressions in analytical integration behavior
- Improves confidence in TF1 fitting machinery

### DIFF
--- a/hist/hist/test/test_tf1.cxx
+++ b/hist/hist/test/test_tf1.cxx
@@ -91,7 +91,10 @@ TEST(TF1AnalyticalIntegral, ExponentialP1Zero)
    const double a = 0.3;
    const double b = 1.7;
 
+
    const double result = f.AnalyticalIntegral(a, b);
+   const double result = f.Integral(a, b);
+
    const double expected = std::exp(1.2) * (b - a);
 
    EXPECT_NEAR(result, expected, 1e-12);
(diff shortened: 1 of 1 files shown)


hist/hist/test/test_tf1.cxx
```diff
@@ -81,6 +82,31 @@ void test_normalization() {
...
+// Test analytical exponential integral when p1 == 0
+TEST(TF1AnalyticalIntegral, ExponentialP1Zero)
+{
+   TF1 f("f_expo_zero", "expo", 0.0, 2.0);
+   f.SetParameters(1.2, 0.0); // p0 = 1.2, p1 = 0
+
+   const double a = 0.3;
+   const double b = 1.7;
+
+   const double result = f.AnalyticalIntegral(a, b);
```
hageboeck, line 95: This does not compile. You need to include AnalyticalIntegral.h and construct an instance of the class. It would be best to test in a local build if the test compiles and passes.

hageboeck (changes requested):

hist/hist/test/test_tf1.cxx
swetank18: Thanks for the clarification.
I’ve updated the tests to use the public TF1::Integral() interface and pushed the fix.

hageboeck:
This work continues in #21060

Issue #7052, opened 2021-01-15
rootcp --replace option ineffective


pcanal:
rootcp helps says:
```
  --replace             replace object if already existing
 ```
but:
```
rm copy.root
rootcp -r --replace tutorials/hsimple.root copy.root;
root.exe -b -l copy.root -e 'gFile->ls()' -q
rootcp -r --replace tutorials/hsimple.root copy.root; 
root.exe -b -l copy.root -e 'gFile->ls()' -q
```
leads to:
```
pcanal@Viroflay:~/root_working/build/master-debug$ rm copy.root 
pcanal@Viroflay:~/root_working/build/master-debug$ rootcp -r --replace tutorials/hsimple.root copy.root; root.exe -b -l copy.root -e 'gFile->ls()' -q

Attaching file copy.root as _file0...
(TFile *) 0x5645f4423530
TFile**         copy.root
 TFile*         copy.root
  KEY: TH1F     hpx;1   This is the px distribution
  KEY: TH2F     hpxpy;1 py vs px
  KEY: TProfile hprof;1 Profile of pz versus px
  KEY: TNtuple  ntuple;1        Demo ntuple
pcanal@Viroflay:~/root_working/build/master-debug$ rootcp -r --replace tutorials/hsimple.root copy.root; root.exe -b -l copy.root -e 'gFile->ls()' -q

Attaching file copy.root as _file0...
(TFile *) 0x55f09c1bbc50
TFile**         copy.root
 TFile*         copy.root
  KEY: TH1F     hpx;2   This is the px distribution [current cycle]
  KEY: TH1F     hpx;1   This is the px distribution [backup cycle]
  KEY: TH2F     hpxpy;2 py vs px [current cycle]
  KEY: TH2F     hpxpy;1 py vs px [backup cycle]
  KEY: TProfile hprof;2 Profile of pz versus px [current cycle]
  KEY: TProfile hprof;1 Profile of pz versus px [backup cycle]
  KEY: TNtuple  ntuple;1        Demo ntuple
```

seemingly addition rather than replacement.

**Comment by guitargeek** [comment]
This should be easy to fix. It is not generally ineffective, but only if you use the recursive option and not specify object names at the same time. This already works:
```txt
rootcp -r --replace tutorials/hsimple.root:hpx copy.root:hpx;
```

silverweed:
This was fixed by https://github.com/root-project/root/pull/20536

PR #21061, opened 2026-01-29
[main] Add Rootcp tests for --replace with entire file

silverweed:
Based on https://github.com/root-project/root/pull/21067 (only last commit relevant).

Adds a roottest for `rootcp --replace a.root b.root` (where both `a.root` and `b.root` exist).

Serves as a test for #7052, which can be closed.


roottest/main/CMakeLists.txt
```diff
@@ -414,6 +414,26 @@ ROOTTEST_ADD_TEST(SimpleRootcpInexistent
...
+ROOTTEST_ADD_TEST(RootcpReplaceEntireFilePrepareInput
+                  COMMAND ${TOOLS_PREFIX}/rootcp${exeext} --recreate -r test.root copy_repl.root
+                  FIXTURES_SETUP main-RootcpReplaceEntireFilePrepareInput-fixture)
+
+ROOTTEST_ADD_TEST(RootcpReplaceEntireFile
+                  COMMAND ${TOOLS_PREFIX}/rootcp${exeext} --replace test.root copy_repl.root
+                  FIXTURES_REQUIRED main-RootcpReplaceEntireFilePrepareInput-fixture
+                  FIXTURES_SETUP main-RootcpReplaceEntireFile-fixture)
+
+ROOTTEST_ADD_TEST(RootcpReplaceEntireFileCheckOutput
```
pcanal, line 428: Is this test sufficient to distinguish between `rootcp --replace` doing the 'right' thing and doing (inadvertently) 'nothing'?

silverweed: I changed the CheckOutput to actually check that the object is different (TH1D that replaces a TH1F)

roottest/main/RootcpReplaceEntireFile.ref
```diff
@@ -0,0 +1,3 @@
+hprof;1
```
pcanal: Is this file still used?

PR #21063, opened 2026-01-29
[ntuple] Add RPageSinkFile::CloneWithDifferentName

silverweed:
Last part of #19904.

Adds a writing analogue to RPageSourceFile::OpenWithDifferentAnchor, which allows to derive a new Sink from an existing one using the same underlying directory but writing to a different RNTuple.
Was previously called `DeriveFor` but I found the name too obscure and I think this is clearer.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/tree/ntuple/inc/ROOT/RMiniFile.hxx
+++ b/tree/ntuple/inc/ROOT/RMiniFile.hxx
@@ -261,6 +261,12 @@ public:
    RNTupleFileWriter &operator=(RNTupleFileWriter &&other) = delete;
    ~RNTupleFileWriter();
 
+   /// Creates a new RNTupleFileWriter with the same underlying TDirectory as this but writing to a different
+   /// RNTuple named `ntupleName`. Onle one of the two writers can safely write to the file at the same time.
+   /// This method is currently only supported for TFile-based Writers and will throw an exception if that's not the
+   /// case.
+   std::unique_ptr<RNTupleFileWriter> CloneWithDifferentName(std::string_view ntupleName) const;
+
    /// Seek a simple writer to offset. Note that previous data is not flushed immediately, but only by the next write
    /// (if necessary).
    void Seek(std::uint64_t offset);
@@ -286,6 +292,8 @@ public:
    void UpdateStreamerInfos(const ROOT::Internal::RNTupleSerializer::StreamerInfoMap_t &streamerInfos);
    /// Writes the RNTuple key to the file so that the header and footer keys can be found
    void Commit(int compression = RCompressionSetting::EDefaults::kUseGeneralPurpose);
+
+   std::string_view GetNTupleName() const { return fNTupleName; }
 };
 
 } // namespace Internal
--- a/tree/ntuple/inc/ROOT/RPageStorageFile.hxx
+++ b/tree/ntuple/inc/ROOT/RPageStorageFile.hxx
@@ -73,6 +73,7 @@ private:
    ROOT::Internal::RNTupleSerializer::StreamerInfoMap_t fInfosOfClassFields;
 
    RPageSinkFile(std::string_view ntupleName, const ROOT::RNTupleWriteOptions &options);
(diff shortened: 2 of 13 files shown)


pcanal:
> which allows to derive a new Sink from an existing one using the same underlying directory but writing to a different RNTuple.

Do we also need to support the case where the user would like to clone in a different directory (and/or is it mostly for internal use)?

silverweed:
> > which allows to derive a new Sink from an existing one using the same underlying directory but writing to a different RNTuple.
> 
> Do we also need to support the case where the user would like to clone in a different directory (and/or is it mostly for internal use)?

It's for internal use only (for the Attributes) and we always want to write to the same directory

tree/ntuple/inc/ROOT/RPageStorageFile.hxx
```diff
@@ -109,6 +110,11 @@ public:
    ~RPageSinkFile() override;
 
    void UpdateSchema(const ROOT::Internal::RNTupleModelChangeset &changeset, ROOT::NTupleSize_t firstEntry) final;
+
+   /// Creates a new sink that uses the same underlying file/directory but writes to a different RNTuple with the
+   /// given `name`.
+   std::unique_ptr<RPageSinkFile>
+   CloneWithDifferentName(std::string_view name, const ROOT::RNTupleWriteOptions &opts) const;
```
jblomer: I think we should make this a virtual function. Object-store based sinks need to implement this, too.

tree/ntuple/src/RPageStorageFile.cxx
```diff
@@ -291,6 +298,15 @@ void ROOT::Internal::RPageSinkFile::CommitDatasetImpl(unsigned char *serializedF
    fWriter->Commit(GetWriteOptions().GetCompression());
 }
 
+std::unique_ptr<ROOT::Internal::RPageSinkFile>
+ROOT::Internal::RPageSinkFile::CloneWithDifferentName(std::string_view name,
+                                                      const ROOT::RNTupleWriteOptions &opts) const
+{
+   auto writer = fWriter->CloneWithDifferentName(name);
+   auto cloned = std::unique_ptr<RPageSinkFile>(new RPageSinkFile(std::move(writer), opts));
+   return cloned;
```
jblomer, line 307:
```suggestion
   return std::unique_ptr<RPageSinkFile>(new RPageSinkFile(std::move(writer), opts));
```

tree/ntuple/inc/ROOT/RMiniFile.hxx
```diff
@@ -261,6 +261,13 @@ public:
    RNTupleFileWriter &operator=(RNTupleFileWriter &&other) = delete;
    ~RNTupleFileWriter();
 
+   /// Creates a new RNTupleFileWriter with the same underlying TDirectory as this but writing to a different
+   /// RNTuple named `ntupleName`. This class is not thread safe, so one of the two writers may be writing to the file
```
jblomer: I think the point is not so much that the class is not thread-safe (because we create to separate objects). It's more that both classes operate on the same underlying file resource.

jblomer (approved):

tree/ntuple/src/RPageStorageFile.cxx
silverweed: I prefer the two-line version for debuggability, but if you insist I can change it.

tree/ntuple/inc/ROOT/RMiniFile.hxx
silverweed: But writing to the same file simultaneously is only a problem is you do it in parallel, no?

tree/ntuple/inc/ROOT/RPageStorageFile.hxx
silverweed: Should this be in `RPageSink` directly or in `RPagePersistentSink`? (does it make sense to clone a sink that has no actual underlying storage?)

tree/ntuple/inc/ROOT/RMiniFile.hxx
jblomer: Yes, that's right. It's a nit anyway, feel free to resolve this.

PR #21064, opened 2026-01-29
[main] rootls: properly handle multiple cycles

silverweed:
With the regular view we still deduplicate cycles, but with detailed view be show all of them.

### DIFF
--- a/main/src/RootObjTree.cxx
+++ b/main/src/RootObjTree.cxx
@@ -50,9 +50,6 @@ ROOT::CmdLine::GetMatchingPathsInFile(std::string_view fileName, std::string_vie
    }
    std::deque<NodeIdx_t> nodesToVisit{0};
 
-   // Keep track of the object names found at every nesting level and only add the first one.
-   std::unordered_set<std::string> namesFound;
-
    const bool isRecursive = flags & EGetMatchingPathsFlags::kRecursive;
    do {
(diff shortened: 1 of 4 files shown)


vepadulano (approved):

main/src/RootObjTree.cxx
```diff
@@ -50,26 +50,24 @@ ROOT::CmdLine::GetMatchingPathsInFile(std::string_view fileName, std::string_vie
...
       std::vector<TKey *> keys;
       keys.reserve(cur->fDir->GetListOfKeys()->GetEntries());
       for (TKey *key : ROOT::Detail::TRangeStaticCast<TKey>(cur->fDir->GetListOfKeys()))
          keys.push_back(key);
 
-      std::sort(keys.begin(), keys.end(),
-                [](const auto *a, const auto *b) { return strcmp(a->GetName(), b->GetName()) < 0; });
-
-      namesFound.clear();
+      std::sort(keys.begin(), keys.end(), [](const TKey *a, const TKey *b) {
```
pcanal, line 66: The operation is usually fast enough but did you verify that it at least in some case does 'something'?  I.e. is there cases (maybe all but maybe none) where the list is not already sorted?

silverweed: Yes, I verified that the keys must be explicitly sorted, otherwise sometimes they aren't

PR #21067, opened 2026-01-29
[main] rootls: add -c to display all cycles

silverweed:
Based on https://github.com/root-project/root/pull/21064 (only last commit is relevant).

Adds a `-c` / `--allCycles` flag to `rootls` that forces it to display all object cycles explicitly. 
`-l` also displays all cycles, but it also shows other information and this flag helps with testing, among other things.


main/src/rootls.cxx
```diff
@@ -71,16 +71,16 @@ static const char *const kLongHelp = R"(
 Display ROOT files contents in the terminal.
 
 positional arguments:
-  FILE                  Input file
+  FILE                    Input file
```
vepadulano, line 74: I guess the extra two characters are to align with `recursiveListing`?

main/src/rootls.cxx
```diff
@@ -379,22 +380,41 @@ static void PrintNodesInColumns(std::ostream &stream, const RootObjTree &tree,
...
+      const auto &node = tree.fNodes[nodeIdx];
+      // De-duplicate node names (we only print one name even if the node has multiple cycles - unless we have the
+      // kForceAllCycles flag active)
+      if (!displayAllCycles && node.fName == prevNodeName)
+         continue;
+      uniqueNodes.push_back(nodeIdx);
+      auto elemWidth = node.fName.length();
+      if (displayAllCycles) {
+         // If we're displaying the cycles we need to keep that in mind for calculating the column widths.
+         const auto cycleLen = std::to_string(node.fCycle).length();
```
pcanal, line 401:
```suggestion
         const auto cycleLen = std::to_string(node.fCycle).length() + 1;
```
i.e. do we also need space for the separator?

main/src/rootls.cxx
```diff
@@ -416,7 +436,12 @@ static void PrintNodesInColumns(std::ostream &stream, const RootObjTree &tree,
                if ((j % nCols) == colIdx) {
                   NodeIdx_t childIdx = nodesBegin[j];
                   const RootObjNode &child = tree.fNodes[childIdx];
-                  width = std::max<int>(width, child.fName.length() + minCharsBetween);
+                  auto nameLen = child.fName.length();
+                  if (displayAllCycles) {
+                     const auto cycleLen = std::to_string(child.fCycle).length();
```
pcanal, line 441:
```suggestion
                     const auto cycleLen = std::to_string(child.fCycle).length() + 1;
```
same as https://github.com/root-project/root/pull/21067/changes#r2742870135

Issue #21058, opened 2026-01-28
ROOT interpreter crashes when using `nlohmann::json`


linev:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Described on forum: https://root-forum.cern.ch/t/64550,

### Reproducer

root interpreter crashes when type:
```
root [0] nlohmann::json j;
root [1] j["3"] = 3
```

If put semicolon - than everything fine.

### ROOT version

master

### Installation method

Build from source

### Operating system

Linux 64bit

dpiparo:
This also crashes
```
root [0] nlohmann::json j;
root [1] j["3"] = 3;
root [2] j["3"]
```
actually on macos, root simply quits, no message.

dpiparo:
It looks like an infinite recursion in the value printing, this is why with the ";" everything seems to work fine. A stacktrace, obtained on macOS with the debugger (note the frame numbers!) can be found below. To trigger the problem directly from the commandline: `root-b -q -l -e "nlohmann::json j;j["3"] = 3"`
```
Process 79908 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x16f603f78)
    frame #0: 0x00000001150d54c8 libCling.so`cling_runtime_internal_throwIfInvalidPointer(Interp=<unavailable>, Expr=<unavailable>, Arg=<unavailable>) at Exception.cpp:33
   30  	/// T* -> const void* -> const_cast<void*> -> T* round trip.
   31  	CLING_LIB_EXPORT
   32  	void* cling_runtime_internal_throwIfInvalidPointer(void* Interp, void* Expr,
-> 33  	                                                   const void* Arg) {
   34  	
   35  	#if defined(__APPLE__) && defined(__arm64__)
   36  	  // See https://github.com/root-project/root/issues/7541 and
Target 0: (root.exe) stopped.
(lldb) bt
warning: could not execute support code to read Objective-C class data in the process. This may reduce the quality of type information available.

* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x16f603f78)
  * frame #0: 0x00000001150d54c8 libCling.so`cling_runtime_internal_throwIfInvalidPointer(Interp=<unavailable>, Expr=<unavailable>, Arg=<unavailable>) at Exception.cpp:33
    frame #1: 0x000000010090c088
    frame #2: 0x000000010090c158
....
[...]

Issue #10140, opened 2022-03-16
[cling] crash with .Class


ferdymercury:
### Describe the bug
A crash is produced if you type .Class in the ROOT prompt.

```
===========================================================================
struct std::atomic<_Tp *>
SIZE: (NA) FILE: atomic LINE: -1
List of member variables --------------------------------------------------
root.exe: /opt/root_src/interpreter/llvm/src/tools/clang/include/clang/AST/TypeNodes.def:100: clang::TypeInfo clang::ASTContext::getTypeInfoImpl(const clang::Type*) const: Assertion `!T->isDependentType() && "should not see dependent types here"' failed.
```

### Expected behavior
No crash.

Also, if you type `.Class TTree` it should only print TTree's class info, not all the list of classes. In analogy to .class TTree vs .class

### To Reproduce
`.Class
`
or `.Class TTree`

### Setup
```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.27/01                        https://root.cern |
  | (c) 1995-2021, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Jan 12 2022, 10:17:19                 |
  | From heads/master@v6-25-01-2870-gdac9b6398d                      |
  | With c++ (Ubuntu 8.4.0-1ubuntu1~18.04) 8.4.0                     |
  | Try '.help', '.demo', '.license', '.credits', '.quit'/'.q'       |
   ------------------------------------------------------------------
```

### Additional context
https://github.com/root-project/root/pull/10121

couet:
`.class` is fine. Should `.Class` be transformed in `.class` ?

ferdymercury:
> `.class` is fine. Should `.Class` be transformed in `.class` ?

.Class is the 'extra verbose' option of .class, thus it is printing more information.

ferdymercury:
Hi @Axel-Naumann this does not seem fixed yet on Ubuntu 22 at least, I tried with:

```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.29/01                        https://root.cern |
  | (c) 1995-2022, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Dec 24 2022, 01:58:00                 |
  | From remotes/origin/master@v6-29-01-47-g37ac665419               |
  | With c++ (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0                   |
[...]

PR #21072, opened 2026-01-29
[cling] Fix infinite recursion when printing self-referencing containers

devajithvs:
Some container types (e.g., nlohmann::json scalar values) have iterators that dereference to the container itself, causing infinite recursion in `printValue_impl`.

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes https://github.com/root-project/root/issues/21058

### DIFF
--- a/interpreter/cling/include/cling/Interpreter/RuntimePrintValue.h
+++ b/interpreter/cling/include/cling/Interpreter/RuntimePrintValue.h
@@ -210,10 +210,20 @@ namespace cling {
       const void* M = TypeTest::isMap(obj);
 
       std::string str("{ ");
-      str += printValue(&(*iter), M);
+
+      // If the dereferenced iterator points to the container itself, we have
+      // infinite recursion. This occurs with scalar values in nlohmann::json
+      auto printWithRecursionGuard = [obj](const auto* ptr,
+                                           const void* M) -> std::string {
+        if (static_cast<const void*>(ptr) == static_cast<const void*>(obj))
+          return "<recursion detected>";
+        return printValue(ptr, M);
+      };
+
+      str += printWithRecursionGuard(&(*iter), M);
       while (++iter != iterEnd) {
         str += ", ";
-        str += printValue(&(*iter), M);
+        str += printWithRecursionGuard(&(*iter), M);
       }
       return str + " }";
     }
--- a/interpreter/cling/test/Prompt/ValuePrinter/Recursion.C
+++ b/interpreter/cling/test/Prompt/ValuePrinter/Recursion.C
@@ -1,0 +1,30 @@
+//------------------------------------------------------------------------------
+// CLING - the C++ LLVM-based InterpreterG :)
+//
+// This file is dual-licensed: you can choose to license it under the University
+// of Illinois Open Source License or the GNU Lesser General Public License. See
+// LICENSE.TXT for details.
+
+//------------------------------------------------------------------------------
+
+// RUN: cat %s | %cling -Xclang -verify 2>&1 | FileCheck %s
+
+.rawInput 1
+// When begin() != end() but *begin() points to the container itself
+// (in nlohmann::json), printValue_impl infinitely recurses without the
+// self-reference check.
+
+class RecursionTest {
+public:
+    RecursionTest() = default;
+    auto begin() const { return this; } // iterate over self
+    auto end() const { return this + 1; } // just to make sure begin() != end()
+};
+.rawInput 0
+
+RecursionTest j;
+j
+// CHECK: (RecursionTest &) { <recursion detected> }
+
+// expected-no-diagnostics
+.q


interpreter/cling/test/Prompt/ValuePrinter/Recursion.C
```diff
@@ -0,0 +1,42 @@
+//------------------------------------------------------------------------------
+// CLING - the C++ LLVM-based InterpreterG :)
```
pcanal, line 2: (I forgot) why is there a trailing `G` ?

interpreter/cling/include/cling/Interpreter/RuntimePrintValue.h
```diff
@@ -210,10 +210,20 @@ namespace cling {
       const void* M = TypeTest::isMap(obj);
 
       std::string str("{ ");
-      str += printValue(&(*iter), M);
+
+      // If the dereferenced iterator points to the container itself, we have
+      // infinite recursion. This occurs with scalar values in nlohmann::json
```
pcanal, line 215: I am confused.  In the example, the 'points to the container itself' happens only if the container is empty and thus this line should not be reached (see line 208).

interpreter/cling/test/Prompt/ValuePrinter/Recursion.C
```diff
@@ -0,0 +1,42 @@
...
+
+    // Assignment to int (scalar)
+    RecursionTest& operator=(int) {
+        data.clear();
+        return *this;
+    }
+
+    // When empty (scalar), iterate over self
+    auto begin() const { return data.empty() ? this : &data[0]; }
+    auto end() const { return data.empty() ? this + 1 : &data[0] + data.size(); }
```
pcanal:
```suggestion
    auto end() const { return data.empty() ? this : &data[0] + data.size(); }
```
Otherwise we don't seem to have the invariant
```
object->empty() == ( object->begin() == object->end() );
```

interpreter/cling/include/cling/Interpreter/RuntimePrintValue.h
pcanal: This is probably related to the odd implementation mentioned at https://github.com/root-project/root/pull/21072/changes#r2742975180

interpreter/cling/test/Prompt/ValuePrinter/Recursion.C
dpiparo: To make the acronym sound better than CLIN?

dpiparo (approved):
lgtm once [this comment ](https://github.com/root-project/root/pull/21072/changes#r2742975180) is addressed.

interpreter/cling/test/Prompt/ValuePrinter/Recursion.C
vepadulano: cling/clang

interpreter/cling/include/cling/Interpreter/RuntimePrintValue.h
devajithvs: This is a problem in `nlohmann::json`, scalar values are iterable and can dereference to themselves. 

```bash
root [0] nlohmann::json j;
root [1] j[0] = 1;
root [2] auto it = j.begin();
root [3] bool(it == j.end())
(bool) false
root [4] auto& elem = *it;
root [5] auto elem_it = elem.begin();
root [6] bool(elem_it == elem.end()) // line 208 check
(bool) false
root [7] bool(&elem == &(*elem_it)) // iterator dereferences to itself and infinite recursion
(bool) true
```

devajithvs: I was trying to somewhat mimic what scalar values does in `nlohmann::json`, I have simplified the test case to avoid this confusion.

interpreter/cling/test/Prompt/ValuePrinter/Recursion.C
```diff
@@ -0,0 +1,30 @@
...
+.rawInput 1
+// When begin() != end() but *begin() points to the container itself
+// (in nlohmann::json), printValue_impl infinitely recurses without the
+// self-reference check.
+
+class RecursionTest {
+public:
+    RecursionTest() = default;
+    auto begin() const { return this; } // iterate over self
+    auto end() const { return this + 1; } // just to make sure begin() != end()
```
pcanal, line 21:
```suggestion
    auto end() const { return this + 1; } // Hard code the size of the collection to be exactly one and it 'just' contains this object.
```

ferdymercury:
Thanks!! Question: does this PR also fix by chance the infinite loop seen in https://github.com/root-project/root/issues/10140#issuecomment-1091806544 ?

pcanal:
> Question: does this PR also fix by chance the infinite loop seen in https://github.com/root-project/root/issues/10140#issuecomment-1091806544 ?

That is very unlikely.

PR #21073, opened 2026-01-29
[tmva][sofie] arguments check to avoid mismatches with dynamic parameters

sanjibansg:
This PR implements checks on the arguments of the infer method to avoid any mismatches with the dynamic parameters.

When SOFIE is used to infer models with dynamic parameters, it is essential to check if they are in the allowed range. Otherwise this may cause an overflow and subsequently a segmentation fault. This check verifies is the passed
parameters is lesser or equal to their limits set during instantion of the Session object, otherwise throws a runtime error.

### DIFF
--- a/tmva/sofie/src/RModel.cxx
+++ b/tmva/sofie/src/RModel.cxx
@@ -1016,18 +1016,13 @@ void RModel::GenerateOutput()
                }
                dynamic_parameters_check += d.param + " > f" + cap + " || ";
                input_params_checked.insert(pName);
+               fGC += SP + "if (" + d.param + " > f" + cap + ") {\n";
+               fGC += SP + SP + "throw std::runtime_error(\"TMVA-SOFIE: dynamic input tensor shape parameter " +
+                      d.param + " exceeds the initialized maximum allowed shape.\");\n";
+               fGC += SP + "}\n";
             }
          }
       }
-   }
-   // remove last && from fGC
-   if (input_params_checked.size() > 0) {
-      dynamic_parameters_check = dynamic_parameters_check.substr(0, dynamic_parameters_check.size() - 4);
-      fGC += "\n" + SP + "if (" + dynamic_parameters_check + ") {\n";
-      fGC += SP + SP +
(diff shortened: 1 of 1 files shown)


dpiparo (changes requested):
Besides the failing tests, could you please follow the contribution guide here https://github.com/root-project/root/blob/master/CONTRIBUTING.md ? This applies also to the commit message.

tmva/sofie/src/RModel.cxx
```diff
@@ -1001,6 +1001,35 @@ void RModel::GenerateOutput()
    if (!doInferArgs.empty())
       doInferArgs.back() = ' ';
 
+   // verifying if the dynamic parameters are within allowed range
+   std::unordered_set<std::string> input_params_checked;
```
lmoneta, line 1005: Why using an unorered set? Since later we sort the names , we should also keep the values sorted there for consistency

tmva/sofie/src/RModel.cxx
```diff
@@ -1001,6 +1001,35 @@ void RModel::GenerateOutput()
...
+                  cap[0] = std::toupper(static_cast<unsigned char>(cap[0]));
+               }
+               dynamic_parameters_check += d.param + " > f" + cap + " || ";
+               input_params_checked.insert(pName);
+            }
+         }
+      }
+   }
+   // remove last && from fGC
+   if (input_params_checked.size() > 0) {
```
lmoneta: I would include this in the loop above and generate an error condition separately for each parameter, so we know which parameter causes the failure. Since we generate the code it is not more work

tmva/sofie/src/RModel.cxx
sanjibansg: in this case, we just need a set to keep a check so that we do not generate the checks repeatedly, so probably that will not affect much? and a set works fine?

tmva/sofie/src/RModel.cxx
sanjibansg: Thanks for the suggestion, I have added the change.

PR #21074, opened 2026-01-29
[CMake] Deal with semicolons in `ENV` passed to RootTestDriver

guitargeek:
The processing of the environment variables in the RootTestDriver used excessive CMake string-to-list parsing, and the semicolons would have to be escaped for each of these parsings. That means the RootTestDriver command needed to include a magic number amount of escape characters to propagate semicolons to the final environment variables.

The fix is to escape semicolons once in the beginning, and then re-escape them after each step that involves list parsing.

Like this, the command can contain just bare semicolons without having to escape anything.

In a next step, one can consider giving the same treatment to the `ROOT_ADD_TEST` macro, which also does excessive list parsing. Then we don't need tricks like the `$<SEMICOLON>` generator expression anymore to propagate semicolons into the final command.

Follows up on a comment by @pcanal: https://github.com/root-project/root/pull/21056#discussion_r2742193422

### DIFF
--- a/roottest/root/io/transient/base/CMakeLists.txt
+++ b/roottest/root/io/transient/base/CMakeLists.txt
@@ -30,8 +30,18 @@ else()
   set(outref_suffix "ZLIB")
 endif()
 
+set(root_includepaths
+    ${CMAKE_CURRENT_SOURCE_DIR}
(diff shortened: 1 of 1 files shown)


pcanal (approved):
LGTM.  Thanks.

PR #21082, opened 2026-01-30
[df] Make Report action mergeable and available in distributed RDF

martamaja10:
The report action has not been mergeable so far and hence it was not possible to use it in distributed RDF. This PR makes it possible. In order to make the testing easier, basic AsString method was added to the RCutFlowReport class. 

Note: 
* distributed RDF tests to be added --> done 
* for now, merged report is tested within the test_report instead of merge_results since it requires a ref file - to be decided where is best to keep this test. --> changed to test merged report in dataframe_merge_results

### DIFF
--- a/bindings/distrdf/python/DistRDF/Operation.py
+++ b/bindings/distrdf/python/DistRDF/Operation.py
@@ -125,6 +125,7 @@ SUPPORTED_OPERATIONS: Dict[str, Union[Action, InstantAction, Transformation]] = 
     "Stats": Action,
     "StdDev": Action,
     "Sum": Action,
+    "Report": Action,
     "VariationsFor": VariationsFor,
     "Vary": Transformation,
 }
--- a/roottest/python/distrdf/backends/check_reducer_merge.py
+++ b/roottest/python/distrdf/backends/check_reducer_merge.py
@@ -422,6 +422,27 @@ class TestReducerMerge:
         assert sum_before.GetValue() == 10.0
         assert sum_after.GetValue() == 20.0
 
+    def test_report(self, payload):
+        """Test the report action (outputting cut flow reports)."""
+
+        import textwrap
+
+        connection, _ = payload
+        df = ROOT.RDataFrame(100, executor=connection)
+        def1 = df.Define("x", "rdfentry_")
+
+        filtered1 = def1.Filter("x>25.", "Cut1")
+        filtered2 = filtered1.Filter("x>50.", "Cut2")
+
+        report = filtered2.Report()
+
+        expected = textwrap.dedent("""\
+            Cut1                : pass=74         all=100        -- eff=74.00 % cumulative eff=74.00 %
+            Cut2                : pass=49         all=74         -- eff=66.22 % cumulative eff=49.00 %
+        """)
+
+        assert report.AsString() == expected
+
     @pytest.mark.parametrize("datasource", ["ttree", "rntuple"])
     def test_distributed_stddev(self, payload, datasource):
         """Test support for the StdDev action."""
--- a/tree/dataframe/test/dataframe_merge_results.py
+++ b/tree/dataframe/test/dataframe_merge_results.py
@@ -54,6 +54,42 @@ class RDataFrameMergeResults(unittest.TestCase):
 
         self.assertAlmostEqual(deser_deser1.GetValue(), mergedmean)
 
+    def test_pickle_report(self):
+        """Merge two reports, pickle and merge again."""
+
+        import textwrap
+
+        df1 = ROOT.RDataFrame(100)
+        df2 = ROOT.RDataFrame(100)
+
+        def1 = df1.Define("x", "rdfentry_")
+        def2 = df2.Define("x", "2*rdfentry_")
+
+        filtered1 = def1.Filter("x>25.", "Cut1")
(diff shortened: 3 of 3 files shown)


tree/dataframe/src/RCutFlowReport.cxx
```diff
@@ -30,6 +31,29 @@ void RCutFlowReport::Print()
...
+{
+   std::stringstream stringRepresentation;
+
+   const auto allEntries = fCutInfos.empty() ? 0ULL : fCutInfos.begin()->GetAll();
+   for (auto &&ci : fCutInfos) {
+      const auto &name = ci.GetName();
+      const auto pass = ci.GetPass();
+      const auto all = ci.GetAll();
+      const auto eff = ci.GetEff();
+      const auto cumulativeEff = 100.f * float(pass) / float(allEntries);
```
hageboeck: Maybe this is better run in `double`. With `float`, one could arrive earlier at 0 or 100%, even if the true values were 99.9999999 or so.

tree/dataframe/src/RCutFlowReport.cxx
```diff
@@ -30,6 +31,29 @@ void RCutFlowReport::Print()
...
+
+      std::string stringtodisplay = Form("%-10s: pass=%-10lld all=%-10lld -- eff=%3.2f %% cumulative eff=%3.2f %%",
```
hageboeck: I wonder if one should allow a bit more space for the cut names? The number is of course arbitrary, but having a bit more space than for a number seems reasonable to me.

tree/dataframe/inc/ROOT/RDF/RMergeableValue.hxx
```diff
@@ -557,6 +560,58 @@ public:
    RMergeableSum &operator=(RMergeableSum &&) = delete;
 };
 
+/**
+\class ROOT::Detail::RDF::RMergeableReport
```
hageboeck: To make things easier, I would omit this line, since it's auto-generated as long as the block stays before the class.
In this way, a future renaming or moving it to another namespace is reflected automatically.

tree/dataframe/inc/ROOT/RDF/RMergeableValue.hxx
```diff
@@ -557,6 +560,58 @@ public:
...
+      }
+      this->fValue = report;
+   }
+
+public:
+   /////////////////////////////////////////////////////////////////////////////
+   /// \brief Constructor that initializes data members.
+   /// \param[in]
+   /// \param[in]
+   /// \param[in]
```
hageboeck: If you put the `\param` tags, I think we should fill them in the same go, or remove them.

hageboeck (approved):
LGTM, mostly cosmetics below.

I have just one question:
What do we do, if by some accident or future extension the two cutflow reports don't contain the same cuts? Should we signal some kind of error or proceed with the match cuts by name and merge what can be merged, and leave untouched what cannot be merged?

tree/dataframe/inc/ROOT/RDF/RMergeableValue.hxx
martamaja10: good point, I think I left it empty to complete later but forgot

tree/dataframe/src/RCutFlowReport.cxx
martamaja10: I followed the choice that was there before (for Print() ), but it is true, we could leave more space, should I double it? or would idk 15 be enough?

tree/dataframe/src/RCutFlowReport.cxx
martamaja10: ok, I will change it, I'll change it here and in the Print function in a separate commit (maybe together with changing the space for the cut name)

tree/dataframe/inc/ROOT/RDF/RMergeableValue.hxx
```diff
@@ -557,6 +561,65 @@ public:
...
+   std::vector<ROOT::RDF::TCutInfo> fCutInfoVec;
+   void Merge(const RMergeableValue<ROOT::RDF::RCutFlowReport> &other) final
+   {
+      ROOT::RDF::RCutFlowReport report;
+      const auto &othercast = dynamic_cast<const RMergeableReport &>(other);
+
+      for (unsigned long i = 0; i < fCutInfoVec.size(); i++) {
+
+         assert(strcmp(this->fCutInfoVec[i].GetName().c_str(), othercast.fCutInfoVec[i].GetName().c_str()) == 0 &&
+                "Mergeable reports have different cut names.");
```
vepadulano: Isn't this assert redundant with the `runtime_error` being thrown a few lines below?

tree/dataframe/src/RCutFlowReport.cxx
```diff
@@ -12,24 +12,49 @@
...
 #include <algorithm>
+#include <sstream>
 #include <stdexcept>
 
 namespace ROOT {
 
 namespace RDF {
 
-void RCutFlowReport::Print()
+void RCutFlowReport::Print() const
```
vepadulano, line 23: Ideally now that we have `AsString` this method should just forward to it.

martamaja10:
Thank you! The requested changes have been implemented and the test has been added.

Issue #8446, opened 2021-06-16
root executable :: remove (or disable by default) any not requested printouts


adriansev:
### Explain what you would like to see improved
root executable emits a lot of non-requested output which:
1. force the users to do workarounds just to make root print only what was instructed by user to do
2. makes it harder to be used as interpreter (to be used in shebang)

### Optional: share how it could be improved
there should be NO output if it was not requested ... 
moreover, IMHO, it's kind of silly that there is a need for `root` and `root.exe`
and users that need to remove annoyances need to `alias root='root.exe -l'`
so, i would say, that all differences between non exe and exe should be removed and hidden under a flag
and all banners or any other text should be by default disabled (so `-l` in fact should enable the banner if requested)
for compatibility the .exe files should be replaced with symlinks to the actual main executables

### To Reproduce
just use it

### Setup
any root version

### Additional context
so, my argument is that, given the _interpreter_ profile of the root executable (the fact that it can execute a human text input make is an interpreter in the usage sense), there should be no actions performed that were not instructed by the user/developer; if so, there is a breach of contract like: `i told you to print a new line and this information and i got another newline and some other text beside what i wanted`

ferdymercury:
Seems related to https://its.cern.ch/jira/browse/ROOT-8344

guitargeek:
Hi! These are all very good points, that we are addressing:

  * The difference between `root` and `root.exe` will be gone soon, and `roo.exe` is kept as a copy of `root` for backwards compatibility for now: https://github.com/root-project/root/pull/20006
  * Spurious new lines in the interpreter will be avoided: https://github.com/root-project/root/pull/21017
  * The splash screen is gone since 2024

About disabling the other printouts: I think we have an excellent precedent with the `python` interpreter. It *only* prints out its introduction banner if you start it up without any input scripts or commands:
```txt
[I] omen nix-shell ~/c/r/root_src rint_line_break > python
Python 3.13.11 (main, Dec  5 2025, 16:06:33) [GCC 15.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
[I] omen nix-shell ~/c/r/root_src rint_line_break > python script.py
Hello world
[I] omen nix-shell ~/c/r/root_src rint_line_break > python -c "print('Hello world')"
Hello world
[I] omen nix-shell ~/c/r/root_src rint_line_break > python -i script.py
Hello world
>>>
[...]

PR #21094, opened 2026-02-01
Don't print ROOT logo in interpreter when input arguments are present

guitargeek:
If one uses the ROOT interpreter to process some scripts, the ROOT logo is always a nuisance, because the output contains lines that are not related to what the user asks. For example, the output can't be piped directly to other commands, or compared to reference files.

Inspired by the behavior of the Python interpreter, this commit suggests to not print the ROOT logo if inputs are passed to the interpreter.

This addresses part of #8446.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -590,7 +590,7 @@ add_custom_command(OUTPUT tutorials/hsimple.root
                      ROOT_INCLUDE_PATH=${DEFAULT_ROOT_INCLUDE_PATH}
                      ROOTIGNOREPREFIX=1
                      ROOT_HIST=0
-                     $<TARGET_FILE:root.exe> -l -q -b -n
+                     $<TARGET_FILE:root.exe> -q -b -n
                        -e ".L ${CMAKE_SOURCE_DIR}/tutorials/hsimple.C"
                        -e "hsimple();"
                        -e return
--- a/bindings/jupyroot/python/JupyROOT/helpers/utils.py
+++ b/bindings/jupyroot/python/JupyROOT/helpers/utils.py
@@ -253,7 +253,7 @@ def _invokeAclicMac(fileName):
     to exclude these libraries, so we launch a second root session to compile
     the library, which we then load.
     """
-    command = 'root -l -q -b -e gSystem->CompileMacro("%s","k")*0' % fileName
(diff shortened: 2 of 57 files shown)


linev:
Why removing `-l` argument from all tests/tutorials?
It is still valid option.

guitargeek:
It is a valid option indeed, but has no effect anymore, because all tests and tutorials use input files.

silverweed:
Can you split this into a commit that removes the `-l` from all the scripts and one that does the functionality change in the interpreter?

pcanal:
I very often use:
```
root.exe -b -q
```
to get the ROOT version information.  Does this still works or is it now 'empty'?  What is the replacement?

Related, this made `-l` the default under 'some' circumstances, is there a way for the user to explicit undo this default (`--print-the-logo-really-I-want-it` :) )

guitargeek:
> I very often use:
> 
> ```
> root.exe -b -q
> ```
> 
> to get the ROOT version information. Does this still works or is it now 'empty'? What is the replacement?

That still works! The logo is only suppressed when you're also passing some inputs to the interpreter, like scripts or commands with `-e`.

> Related, this made `-l` the default under 'some' circumstances, is there a way for the user to explicit undo this default (`--print-the-logo-really-I-want-it` :) )

Yes, that exists :smile: Demo:
```txt
root -b -q -e "dynamic_cast<TRint*>(gApplication)->PrintLogo()" -e "std::cout << 42 << std::endl;"
```
```txt
  | Welcome to ROOT 6.39.01                        https://root.cern |
  | Built for linuxx8664gcc on Jan 01 1980, 00:00:00                 |
  | From heads/master@v6-39-01-887-g3ab1d5c62a0                      |
  | With clang version 21.1.8 std202302                              |

42
```

pcanal:
```
root -b -q -e "d...."
```
Yuck :) .. this is an inflation of 630%!
Please consider re-adding that `root -b -q` with no files or -e arguments still prints the logo (with the justification that other wise that command is a no-op)

guitargeek:
> Please consider re-adding that `root -b -q` with no files or -e arguments still prints the logo (with the justification that other wise that command is a no-op)

Sorry there must be some misunderstanding: `root -b -q` is still printing the logo. With this PR, the only change is that there is only no logo anymore when you pass input files or `-e` commands. So your bare `root -b -q` to get the ROOT version is still fine!

PR #21095, opened 2026-02-01
Implementation of the TGeoTessellated navigation functionality  

BSali:
# This Pull request:
Introduce navigation functionality for TGeoTessellated using triangle-ray-intersection tests.

This is the navigation functionality as it is implemented for the historically grown, non-ROOT-implementation [TGeoArbN](https://doi.org/10.1051/epjconf/202533701022).
To improve usability in VMC based simulations of meshes with high triangle count (>500), partitioning structures in the form of an Octree, as well as a BVH (using the bvh library included in root), are added. We found so far that the choice of partitioning structures seems to be mesh dependent (often BVH is best, however, for some particularly large meshes (~250k triangles) the Octree seems better). However, our BVH usage may be suboptimal.

As Dr. Sandro Wenzel opened a pull-request for his navigation implementation  [Add BVH-based navigation functions to TGeoTessellated #21045](https://github.com/root-project/root/pull/21045), we wanted to "offer" our implementation in parallel. Maybe something can be learned from it.

Tested using TGeoChecker CheckShape functionality (on a box mesh, the teddy.obj in tutorials/visualisation/geom/ contains triangle-intersections and hence is faulty), as well as by computing navigation information for 100 million random points for a box mesh and comparing results with TGeoBBox as quick test that no errors were introduced while translating our TGeoArbN to ROOT's TGeoTessellated.

Dr. Sandro Wenzel's 'rayTriangle' implementation was used to improve own implementation of Triangle ray intersection in TGeoTriangle. 

## Still required changes:
- TBVH not optimally implemented, may learn from Dr. Sandro Wenzel's implementation approach (see PR  [Add BVH-based navigation functions to TGeoTessellated #21045](https://github.com/root-project/root/pull/21045))
- root/geom/geom/inc/bvh/v2/top_down_sah_builder.h required commenting out 
```c++
// #if __has_include(<span>)
// #include <span>
// #else
// Falling back to ROOT span
#include "ROOT/span.hxx"
// #endif
``` 
  to compile for me, otherwise issues with std.span and modules
- I had to reintroduce 'Physics' to geom's dependencies to be able to use TVector3.h again. Change to Math::ROOT::XYZVector may be required.
- Squash commits

### DIFF
--- a/geom/geom/src/TGeoTessellated.cxx
+++ b/geom/geom/src/TGeoTessellated.cxx
@@ -254,9 +254,13 @@ Double_t TGeoTessellated::DistFromOutside(const Double_t *pointa, const Double_t
       return result;
    }
 
+   ROOT::Math::XYZVector tmpoffset{dir * -1};
+   tmpoffset *= (2 * TGeoShape::Tolerance()) / TMath::Sqrt(tmpoffset.Mag2());
+   ROOT::Math::XYZVector newpoint = point + tmpoffset;
+
    std::vector<TGeoTriangleMesh::IntersectedTriangle_t> indir{};
    std::vector<TGeoTriangleMesh::IntersectedTriangle_t> oppdir{};
-   fMesh->FindClosestIntersectedTriangles(point, dir, fUsedTriangles, indir, oppdir);
+   fMesh->FindClosestIntersectedTriangles(newpoint, dir, fUsedTriangles, indir, oppdir);
    size_t size = indir.size();
    size_t counter = 0;
 
--- a/geom/geom/src/Tessellated/TOctree.cxx
+++ b/geom/geom/src/Tessellated/TOctree.cxx
@@ -406,14 +406,10 @@ Double_t TOctree::DistFromOutside(const ROOT::Math::XYZVector &origin, const ROO
    size_t counter = 0;
 
    while (counter < size) {
-      // If you sit on the triangle, ignore it
-      if (triangleIntersections[counter].fDistance < TGeoShape::Tolerance()) {
-         ++counter;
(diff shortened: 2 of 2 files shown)


dpiparo:
Thanks for the proposal @BSali ! Even if this is a draft, let me comment that TVector3 is a `legacy` class (see [the doc](https://root.cern.ch/doc/master/classTVector3.html)) and, as such, should not be used and has to be replaced.

agheata:
Thanks a lot for the PR @BSali ! We are evaluating Sandro's request now and will have a look at this soon. We will likely integrate indeed his version based on `bvh::v2` to start with, but certainly evaluate also your octree-based implementation to compare the performance. I'll be out of the office for a couple of weeks, but maybe @sawenzel will find some time meanwhile to have a look. I'm sure that thanks to these contributions, we will have a very efficient version filling the gap of missing navigation for the tessellated shapes by the time of the release of ROOT this spring.

BSali:
Sorry for the late reply. I had  played around more with the BVH and it is more performant than the Octree implementation. So, with Sandro Wenzel's PR merged, this Draft PR can be closed. 

It may be beneficial to add some more options of loading triangle meshes into TGeoTessellated instances. Loading triangle meshes from .stl files is (from my  experience) a prominent use case for TGeoTessellated in regards to automatically converting CAD/STEP to ROOT, and dedicated functionality to load .stls may be slightly more user friendly. Especially, since TGeoTessellated already provides .obj loading functionality.

agheata:
Thanks @BSali ! The main current limitation is the inability to load facets with more than 4 edges, needing a triangulator. Support for .stl can also be added.

Issue #21076, opened 2026-01-30
Tests: roottest-root-ntuple-makeproject-ttree-read_ttree,  roottest-root-ntuple-makeproject-rntuple-read_rntuple  failing to find libtbb.so


andresailer:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

E.g.: https://lcgapp-services.cern.ch/cdash/tests/76346683
```
-- TEST COMMAND -- 
cd /build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-HEAD/src/ROOT-HEAD-build/roottest/root/ntuple/makeproject/rntuple
/usr/bin/timeout -s USR2 270s ./read_rntuple
-- BEGIN TEST OUTPUT --
./read_rntuple: error while loading shared libraries: libtbb.so.12: cannot open shared object file: No such file or directory

-- END TEST OUTPUT --
CMake Error at /build/jenkins/workspace/lcg_nightly_pipeline/build/projects/ROOT-HEAD/src/ROOT/HEAD/cmake/modules/RootTestDriver.cmake:186 (message):
  got exit code 127 but expected 0
```

### Reproducer

build from source, run tests with non-system tbb
```
source /cvmfs/sft.cern.ch/lcg/releases/LCG_108a/Python/3.12.11/x86_64-el9-gcc15-dbg/Python-env.sh
source /cvmfs/sft.cern.ch/lcg/releases/LCG_108a/jsonmcpp/3.11.3/x86_64-el9-gcc15-dbg/jsonmcpp-env.sh
source /cvmfs/sft.cern.ch/lcg/releases/LCG_108a/tiff/4.7.0/x86_64-el9-gcc15-dbg/tiff-env.sh
source /cvmfs/sft.cern.ch/lcg/releases/LCG_108a/GSL/2.8/x86_64-el9-gcc15-dbg/GSL-env.sh
source /cvmfs/sft.cern.ch/lcg/releases/LCG_108a/gl2ps/1.4.2/x86_64-el9-gcc15-dbg/gl2ps-env.sh
source /cvmfs/sft.cern.ch/lcg/releases/LCG_108a/fftw3/3.3.10/x86_64-el9-gcc15-dbg/fftw-env.sh
source /cvmfs/sft.cern.ch/lcg/releases/LCG_108a/cfitsio/4.6.2/x86_64-el9-gcc15-dbg/cfitsio-env.sh
[...]

PR #21097, opened 2026-02-02
Allow proper RPATH usage on MacOS for TFile::MakeProject and TSystem::CompileMacro

vepadulano:
Read each commit description for more details.

In particular, see the following before and after

ROOT master
```
$: otool -l librntuplestltest/librntuplestltest.so | grep -A2 LC_ID_DYLIB
          cmd LC_ID_DYLIB
      cmdsize 56
         name librntuplestltest.so (offset 24)
```
Patch
```
$: otool -l librntuplestltest/librntuplestltest.so | grep -A2 LC_ID_DYLIB
          cmd LC_ID_DYLIB
      cmdsize 56
         name @rpath/librntuplestltest.so (offset 24)
```

And for reference also one with TSystem::CompileMacro

```
# myclass.h
#ifndef MYCLASS
#define MYCLASS

class MyClass{};

#endif

$: root -l -b -q -e "gSystem->CompileMacro(\"myclass.h\", \"k\")"
Info in <TUnixSystem::ACLiC>: creating shared library /Users/vpadulan/Programs/rootproject/myclass_h.so
(int) 1

$: otool -l myclass_h.so | grep -A2 LC_ID_DYLIB
          cmd LC_ID_DYLIB
      cmdsize 48
         name @rpath/myclass_h.so (offset 24)
```

Fixes https://github.com/root-project/root/issues/21076

### DIFF
--- a/cmake/modules/RootMacros.cmake
+++ b/cmake/modules/RootMacros.cmake
@@ -2471,7 +2471,8 @@ endfunction(ROOTTEST_ADD_AUTOMACROS)
 #-------------------------------------------------------------------------------
 #
 # macro ROOTTEST_COMPILE_MACRO(<filename> [BUILDOBJ object] [BUILDLIB lib]
-#                                         [FIXTURES_SETUP ...] [FIXTURES_CLEANUP ...] [FIXTURES_REQUIRED ...])
+#                                         [FIXTURES_SETUP ...] [FIXTURES_CLEANUP ...] [FIXTURES_REQUIRED ...]
+#                                         [EXTRA_RPATHS ...])
 #
 # This macro creates and loads a shared library containing the code from
 # the file <filename>. A test that performs the compilation is created.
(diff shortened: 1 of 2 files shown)


cmake/modules/RootMacros.cmake
```diff
@@ -2496,11 +2497,18 @@ macro(ROOTTEST_COMPILE_MACRO filename)
   endforeach()
 
   cmake_path(CONVERT "${CMAKE_CURRENT_BINARY_DIR}" TO_NATIVE_PATH_LIST NATIVE_BINARY_DIR)
+  if(ARG_EXTRA_RPATHS)
+    set(extra_rpaths)
+    foreach(rpath ${ARG_EXTRA_RPATHS})
+      list(APPEND extra_rpaths -e "gSystem->AddExtraRPath(\"${rpath}\")")
```
```suggestion
      list(APPEND extra_rpaths -e "gSystem->SetMakeSharedLib(TString{gSystem->GetMakeSharedLib()}.ReplaceAll(\"$RPath\", \"$RPath -Wl,-rpath,${rpath}\"))")
```

You could in principle achieve the same outcome without extending the public interface of TSystem.

Personally I'd prefer that for now, because it's good if the public interfaces don't grow too much unless there is explicit user demand.

vepadulano: Thanks! That's a good idea indeed, I'll implement it.

Issue #21093, opened 2026-02-01
Fixed malformed path with double slash in CMakeLists.txt


x12301450:
### Explain what you would like to see improved and how.

In line 343 of [interpreter/CMakeLists.txt](https://github.com/root-project/root/tree/master/interpreter/CMakeLists.txt), you defined `ROOT_LLVM_MAIN_SRC_DIR` with trailing slash.

`set(ROOT_LLVM_MAIN_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/llvm-project/llvm/")`

In the following codes, this path is used to construct new paths with other paths with leading slashes, which leads to malformed paths with double slash:

```
if(EXISTS ${ROOT_LLVM_MAIN_SRC_DIR}/utils/lit/lit.py)
set(LLVM_LIT ${ROOT_LLVM_MAIN_SRC_DIR}/utils/lit/lit.py CACHE PATH "The location of the lit test runner.")
add_subdirectory(${ROOT_LLVM_MAIN_SRC_DIR}/utils/FileCheck utils/FileCheck)
add_subdirectory(${ROOT_LLVM_MAIN_SRC_DIR}/utils/count utils/count)
add_subdirectory(${ROOT_LLVM_MAIN_SRC_DIR}/utils/not utils/not)
...
```

I think it's better to remove the trailing slash at the time of definition:

`set(ROOT_LLVM_MAIN_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/llvm-project/llvm")`

### ROOT version

N/A

### Installation method

N/A

### Operating system

N/A

dpiparo:
Hi @x12301450 . Would you like to propose a PR for this change?

x12301450:
Sure. I have commit pr #21096 to address this.

dpiparo:
Thanks the PR was tried, and some comments are waiting to be addressed.

x12301450:
The origin one was closed and I submit the pull request #21102 again! Thank you!

ferdymercury:
Closing since CMake seems to deal fine with repeated slashes for all the cases used by ROOT, see: https://cmake.org/cmake/help/latest/command/cmake_path.html
Please reopen if you find a case where this leads to a compilation failures in your setup that can be reproduced.
Thanks!

PR #21102, opened 2026-02-02
[cmake] fix malformed path with double slash in CMakeLists.txt

x12301450:
Remove the trailing slash of ROOT_LLVM_MAIN_SRC_DIR in CMakeLists.txt.

This PR fixes is #21093

### DIFF
--- a/interpreter/CMakeLists.txt
+++ b/interpreter/CMakeLists.txt
@@ -340,7 +340,7 @@ else()
         AND EXISTS ${LLVM_TOOLS_BINARY_DIR}/not${CMAKE_EXECUTABLE_SUFFIX})
       set(LLVM_UTILS_PROVIDED ON)
     endif()
-    set(ROOT_LLVM_MAIN_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/llvm-project/llvm/")
+    set(ROOT_LLVM_MAIN_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/llvm-project/llvm")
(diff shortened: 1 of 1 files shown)


guitargeek:
I'm curious: what's the benefit of this change? Can CMake not deal with double slashes? Normally that doesn't make the path invalid.

x12301450:
I think this is an irregular path that could lead to confusion with path separators in future development. Moreover, some CMake commands handle "//" while some don't, some may delegate handling (or not handling) of such paths to the different operating systems. Whether it will cause bugs depends on how you will use it in the fureture. Take this following fix as [example](https://github.com/arduino/ArduinoCore-zephyr/commit/c696ec304df8ebc8c3343db3df639fff5830), the irregular path with "//" caused an error in string manipulation. So I think this is a possible enhancement and commit this pr.

ferdymercury:
> Whether it will cause bugs depends on how you will use it in the future

Some related questions:

Does this mean that there is currently no bug / actual issue due to this double slash? How did you find out this potential problem, by chance or via a static analyzer? Is this the only place throughout the codebase where this happens, or are there more occurrences of double slash?

If this fixes just a potential-future-bug, maybe it's more work to fix it now than to fix it whenever the issue may potentially appear in the future.

> Take this following fix as [example](https://github.com/arduino/ArduinoCore-zephyr/commit/c696ec304df8ebc8c3343db3df639fff5830), the irregular path with "//" caused an error in string manipulation.

The error seems to come from `-iwithprefixbefore` which is not about CMake. As far as I can see, ROOT-specific code does not use this flag directly.

guitargeek:
No reply by the creator of the PR: closing.

PR #21105, opened 2026-02-02
[hist] Ensure consistency of `RBinWithError::AtomicAdd `

hahnjo:
Instead of using separate atomic operations on the two members, use the sign bit of `fSum2` to implement a cheap spin lock. This allows another thread to detect inconsistent reads and implement a retry loop.

According to the microbenchmarks, this is even faster with a single thread / for the uncontended case. A possible explanation is that the code needs only one compare-exchange operation instead of two. With multiple threads and heavy contention, the implementation is slightly slower.

An alternative would be double-width compare-and-swap instructions, as found on x86 (`cmpxchg16b`). Setting aside the portability issues, GCC generates a library call to `__atomic_compare_exchange_16` which is slower than two separate compare-exchange instructions. If a direct instruction is emitted (for example with Clang), it beats the separate compare-exchange instructions in the uncontended scenario and is tied in case of contention.

### DIFF
--- a/hist/histv7/inc/ROOT/RBinWithError.hxx
+++ b/hist/histv7/inc/ROOT/RBinWithError.hxx
@@ -7,6 +7,8 @@
 
 #include "RHistUtils.hxx"
 
+#include <cmath>
+
 namespace ROOT {
 namespace Experimental {
 
@@ -61,8 +63,29 @@ struct RBinWithError final {
 private:
    void AtomicAdd(double a, double a2)
    {
-      Internal::AtomicAdd(&fSum, a);
-      Internal::AtomicAdd(&fSum2, a2);
+      // The sum of squares is always non-negative. Use the sign bit to implement a cheap spin lock.
+      double origSum2;
+      Internal::AtomicLoad(&fSum2, &origSum2);
+
+      while (true) {
+         // Repeat loads from memory until we see a non-negative value.
+         // NB: do not use origSum2 < 0, it does not work for -0.0!
+         while (std::signbit(origSum2)) {
+            Internal::AtomicLoad(&fSum2, &origSum2);
+         }
+
+         // The variable appears to be unlocked, confirm with a compare-exchange.
+         double negated = std::copysign(origSum2, -1.0);
+         if (Internal::AtomicCompareExchangeAcquire(&fSum2, &origSum2, &negated)) {
+            break;
+         }
+      }
+
+      // By using a spin lock, we do not need atomic operations for fSum.
+      fSum += a;
+
+      double sum2 = origSum2 + a2;
+      Internal::AtomicStoreRelease(&fSum2, &sum2);
    }
 
 public:
--- a/hist/histv7/inc/ROOT/RHistUtils.hxx
+++ b/hist/histv7/inc/ROOT/RHistUtils.hxx
@@ -5,6 +5,7 @@
 #ifndef ROOT_RHistUtils
 #define ROOT_RHistUtils
 
+#include <atomic>
 #include <type_traits>
 
 #ifdef _MSC_VER
@@ -33,6 +34,10 @@ struct AtomicOps<1> {
    {
       *static_cast<char *>(ret) = __iso_volatile_load8(static_cast<const char *>(ptr));
    }
+   static void Store(void *ptr, void *val)
+   {
+      __iso_volatile_store8(static_cast<char *>(ptr), *static_cast<char *>(val));
+   }
    static void Add(void *ptr, const void *val)
    {
       _InterlockedExchangeAdd8(static_cast<char *>(ptr), *static_cast<const char *>(val));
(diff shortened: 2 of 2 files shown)


hist/histv7/inc/ROOT/RBinWithError.hxx
```diff
@@ -58,26 +60,43 @@ struct RBinWithError final {
       return *this;
    }
 
-   void AtomicInc()
+private:
+   void AtomicAdd(double a, double a2)
```
pcanal, line 64:
```suggestion
   void SumAtomicAdd(double a, double a2)
```
As this is specifically to atomic at the sum/sum2.
or
```suggestion
   void AtomicAddToSum(double a, double a2)
```

hahnjo: I think I would prefer to keep the current naming scheme: User-facing functions, such as `FillAtomic`, are implemented in terms of increasingly lower-level functions `AtomicAdd`. Given that it's a `private` function that is used to implemented other `AtomicAdd` functions just below, I think the current name is sufficient.

hist/histv7/inc/ROOT/RBinWithError.hxx
```diff
@@ -61,8 +63,29 @@ struct RBinWithError final {
...
+
+      while (true) {
+         // Repeat loads from memory until we see a non-negative value.
+         // NB: do not use origSum2 < 0, it does not work for -0.0!
+         while (std::signbit(origSum2)) {
+            Internal::AtomicLoad(&fSum2, &origSum2);
+         }
+
+         // The variable appears to be unlocked, confirm with a compare-exchange.
+         double negated = std::copysign(origSum2, -1.0);
```
hageboeck, line 78: One could make this a `const double` to prevent accidental modifications (see related comment on that function).

hist/histv7/inc/ROOT/RHistUtils.hxx
```diff
@@ -142,6 +171,19 @@ bool AtomicCompareExchange(T *ptr, T *expected, T *desired)
 #endif
 }
 
+template <typename T>
+bool AtomicCompareExchangeAcquire(T *ptr, T *expected, T *desired)
```
hageboeck, line 175: Could one protect `desired` from undesired modifications, similar to `std::atomic::compare_exchange`?
```suggestion
bool AtomicCompareExchangeAcquire(T *ptr, T *expected, T desired)
```

hist/histv7/inc/ROOT/RHistUtils.hxx
```diff
@@ -132,6 +149,18 @@ void AtomicLoad(const T *ptr, T *ret)
 #endif
 }
 
+template <typename T>
+void AtomicStoreRelease(T *ptr, T *val)
```
hageboeck, line 153: It's probably too late, but could one have disallowed modifications of the read-only arguments?

```suggestion
void AtomicStoreRelease(T *ptr, T const *val)
```
Or by value?

hageboeck (commented):
LGTM in general, but for the future, should one think about preventing undesired modifications of the read-only arguments?

hist/histv7/inc/ROOT/RHistUtils.hxx
hahnjo: It's not clear to me if it's better to take `desired` by value or by pointer. I think I chose this design because the "generic" version of the `__atomic` functions take a pointer. FWIW, we cannot take it via `const T *` because Clang complains:
```
RHistUtils.hxx:178:52: error: cannot initialize a parameter of type 'double *' with an lvalue of type 'const double *'                               
  178 |    return __atomic_compare_exchange(ptr, expected, desired, /*weak=*/false, __ATOMIC_ACQUIRE, __ATOMIC_RELAXED);                                                                               
      |                                                    ^~~~~~~
```

hist/histv7/inc/ROOT/RBinWithError.hxx
hahnjo: On the other hand, the variable lives only for two lines...

hist/histv7/inc/ROOT/RHistUtils.hxx
hahnjo: Same, Clang complains for `const T *`:
```
RHistUtils.hxx:156:24: error: cannot initialize a parameter of type 'double *' with an lvalue of type 'const double *'                               
  156 |    __atomic_store(ptr, val, __ATOMIC_RELEASE);
      |                        ^~~
```

Regarding "by value", see the other comment...

hist/histv7/inc/ROOT/RHistUtils.hxx
hahnjo: > It's not clear to me if it's better to take desired by value or by pointer. I think I chose this design because the "generic" version of the `__atomic` functions take a pointer.

Answer from the compiler:
```
RHistUtils.hxx:168:11: error: address argument to atomic operation must be a pointer to integer or pointer ('double *' invalid)
  168 |    return __atomic_compare_exchange_n(ptr, expected, desired, /*weak=*/false, __ATOMIC_RELAXED, __ATOMIC_RELAXED);
      |           ^                           ~~~
```

So unless we want to reinterpret memory as an integer, we have to keep using the pointer version, and then it must not be `const` unfortunately.

PR #21106, opened 2026-02-02
[Python] Register Pythonizations only in `ROOTFacade._finalSetup()`

guitargeek:
Delaying the registering of the Pythonizations to the final setup of the ROOT module means we can `import ROOT` anywhere in the pythonization modules without causing a circular import.

Right now this is not a strict requirement, because the circular import is circumvented by using `cppyy` directly in the Pythonization module, but that workaround relies on `cppyy` and ROOT being separate packages. We want to move `cppyy` to be a private submodule of ROOT one day (`ROOT._cppyy`), so we have to find another solution.

Any import of `cppyy` can now be completely avoided in the pythonization modules, showing that the circular import problem is solved.


vepadulano (approved):
Clever trick, very nice! The changes LGTM, but before merging I will launch a clean build of the CI and also the build of the Python wheels, let's wait for those

Issue #21104, opened 2026-02-02
Out-of-bounds access in TFormula constructor


lmoureaux:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Doing `TFormula("f", "[")` triggers out-of-bounds `TString` reads.

### Reproducer

```cpp
TFormula f("f", "[");
```
Output in the interpreter:
```
root [0] TFormula f("f", "[")
Error in <TString::AssertElement>: out of bounds: i = 2, Length = 1
Error in <TString::AssertElement>: out of bounds: i = 3, Length = 1
Error in <TString::AssertElement>: out of bounds: i = 4, Length = 1
Error in <TString::AssertElement>: out of bounds: i = 5, Length = 1
Error in <TString::AssertElement>: out of bounds: i = 6, Length = 1
Error in <TString::AssertElement>: out of bounds: i = 7, Length = 1
Error in <TString::AssertElement>: out of bounds: i = 8, Length = 1
```
In compiled code this can trigger a SEGV instead.

The correct behaviour is to safely create a `TFormula` for which `IsValid()` returns `false`.

### ROOT version

```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.32.02                        https://root.cern |
  | (c) 1995-2024, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Sep 18 2024, 20:27:53                 |
  | From heads/master@tags/v6-32-02                                  |
  | With                                                             |
  | Try '.help'/'.?', '.demo', '.license', '.credits', '.quit'/'.q'  |
   ------------------------------------------------------------------
```

### Installation method

conda

### Operating system

Linux (Ubuntu 24.04)

### Additional context

Also on macOS with a more recent ROOT, see https://gitlab.cern.ch/Proto/Darwin/-/merge_requests/146#note_10873492

PR #860, opened 2017-08-15
Formul params3

arthur-tsang:
This pull request allows one to use user-defined (and some predefined) functions with different variables and/or parameters from what was given in the original definition. For example, a function originally defined on variable `x` with parameter `[0]` can now easily be applied to variable `y` with parameter `[alpha]` as part of another function definition. The syntax for the `TFormula` is as follows:
* `f(var0, var1, var2, ..., param0, param1, param2, ...)`

Furthermore, this PR introduces a few shortcuts:
* The syntax `[n..m]` for ranges of numbered parameters (for example, `[1..3]` expands to `[1],[2],[3]`)
* To simply replace/rename parameters, the variables may be omitted
* Otherwise, if redefining variables, the parameters may be omitted (except, of course, if substituting them all for parameters if the numbers of parameters and variables are equal)

This PR supports the nesting of functions (but still not recursive definitions!).

PR #21110, opened 2026-02-02
[hist] prevent TFormula out of bounds access

ferdymercury:
# This Pull request:

## Changes or fixes:

Fixes https://github.com/root-project/root/issues/21104

## Checklist:

- [x] tested changes locally

### DIFF
--- a/hist/hist/src/TFormula.cxx
+++ b/hist/hist/src/TFormula.cxx
@@ -1845,13 +1845,15 @@ void TFormula::HandleLinear(TString &formula)
    TString expandedFormula = "";
    int delimeterPos = 0;
    for(std::size_t iTerm = 0; iTerm < terms.size(); ++iTerm) {
-      // determine the position of the "@" operator in the formula
-      delimeterPos += terms[iTerm].size() + (iTerm == 0);
-      if(IsAParameterName(formula, delimeterPos)) {
-         // append the current term and the remaining formula unchanged to the expanded formula
-         expandedFormula += terms[iTerm];
-         expandedFormula += formula(delimeterPos, formula.Length() - (delimeterPos + 1));
-         break;
+      if (iTerm < terms.size() - 1) { // N terms, N - 1 @
+         // determine the position of the "@" operator in the formula
+         delimeterPos += terms[iTerm].size() + iTerm;
+         if(IsAParameterName(formula, delimeterPos)) {
+            // append the current term and the remaining formula unchanged to the expanded formula
+            expandedFormula += terms[iTerm];
+            expandedFormula += formula(delimeterPos, formula.Length() - delimeterPos);
+            break;
+         }
       }
       SetBit(kLinear, true);
       auto termName = std::string("__linear") + std::to_string(iTerm+1);


hist/hist/test/test_TFormula.cxx
```diff
@@ -98,3 +98,10 @@ TEST(TFormulaPolTest, CompoundExpressions)
    TFormula f1("f1", "pol1(x,0) + pol1(y,2)");
    EXPECT_EQ(f1.GetExpFormula(), TString("([p0]+[p1]*x)+([p2]+[p3]*y)"));
 }
+
+// https://github.com/root-project/root/issues/21104
+TEST(TFormula, SingleOpeningBracket)
+{
+   TFormula f1("f1", "[");
+   EXPECT_EQ(f1.GetNdim(), 0);
```
vepadulano, line 109: Let's maybe also check for `IsValid() == false`?

ferdymercury: I checked and IsValid() is true.

Supposedly, TFormula works ignoring anything that starts with square brackets. Quoting:
`// ignore things that start with square brackets`
(there is no mention about it being forced to end with square brackets)

I guess that if someone relied on this, if we now return false for IsValid, that might break compatibility?

vepadulano: @lmoureaux @pcanal FYI. I'm unsure how to proceed, `"["` shouldn't be a valid formula imho but breaking behaviour could be dangerous at the same time.

ferdymercury: Note that the same happens with quotes or hex
```
// ignore strings
// ignore x in hexadecimal number
```
there is only check that it starts, but no checks whether it correctly ends.

lmoureaux: If it helps the discussion, Patrick got a similar crash with `"("`. Semantically I'd expect `IsValid()` to be false if the formula can't return anything, which is definitely the case of an empty formula.

pcanal: For reference some/most of this code was introduce in https://github.com/root-project/root/pull/860

pcanal: I am guessing the 'ignore things that start with square brackets' in `TFormula::HandleFunctionArguments` is because some other part of the code is suppose to handle the leading brackets (they are expected in the syntax for selecting parameters, eg.`[1]` (see https://github.com/root-project/root/pull/860).

I think that there is missing code (possibly somewhere else) to mark both `"["` and `"("` as `invalid`.

> but breaking behaviour could be dangerous at the same time.

In this case, I think it should be safe to 'end up' with marking them invalid but it is not clear 'where' in the control flow this need to be checked for.

ferdymercury: check added

ferdymercury:
For the test failure, I think it's due to the logic applied here (https://github.com/root-project/root/commit/dcf920904564bbdca08a6acfa70019bbc24b3897#r176468371)

hist/hist/src/TFormula.cxx
```diff
@@ -49,6 +49,43 @@ std::string doubleToString(double val)
...
+            i.push(s[c]);
+         else if (areMatching(i.top(), s[c]))
+            i.pop();
+         else {
+            i.push(s[c]);
+         }
+      }
+   }
+   if (!i.empty()) {
+      Error("TFormula", "String %s with %zu unbalanced chars.", s.Data(), i.size());
```
vepadulano:
```suggestion
      Error("TFormula", "String '%s' with %zu unbalanced chars.", s.Data(), i.size());
```

hist/hist/src/TFormula.cxx
```diff
@@ -49,6 +49,43 @@ std::string doubleToString(double val)
...
+bool isBalanced(const TString &s)
+{
+   std::stack<char> i;
+   auto sLength = s.Length();
+   for (Ssiz_t c = 0; c < sLength; ++c) {
+      if (s[c] == '[' || s[c] == '{' || s[c] == '(') {
+         i.push(s[c]);
+      } else if (s[c] == ']' || s[c] == '}' || s[c] == ')') {
+         if (i.empty() || !areMatching(i.top(), s[c])) {
+            Error("TFormula", "Found unbalanced char %c (expected closing of %c) at index %d of %s", s[c], i.top(), c,
```
vepadulano:
```suggestion
            Error("TFormula", "Found unbalanced char '%c' (expected closing of '%c') at index %d of '%s'", s[c], i.top(), c,
```

hist/hist/src/TFormula.cxx
```diff
@@ -1844,7 +1881,7 @@ void TFormula::PreProcessFormula(TString &formula)
    HandleParamRanges(formula);
    HandleFunctionArguments(formula);
    HandleExponentiation(formula);
-   // "++" wil be dealt with Handle Linear
+   // "++" (now @) wil be dealt with Handle Linear
```
vepadulano: Why "now"? I don't see where the substitution happens. In any case, this is a comment that will age quickly with this formulation.

ferdymercury: In line 1840 this is done. There are no ++ left in the formula.

hist/hist/test/test_TFormula.cxx
```diff
@@ -98,3 +98,17 @@ TEST(TFormulaPolTest, CompoundExpressions)
    TFormula f1("f1", "pol1(x,0) + pol1(y,2)");
    EXPECT_EQ(f1.GetExpFormula(), TString("([p0]+[p1]*x)+([p2]+[p3]*y)"));
 }
+
+// https://github.com/root-project/root/issues/21104
+TEST(TFormula, SingleOpeningBracket)
+{
+   ROOT::TestSupport::CheckDiagsRAII diags;
+   diags.requiredDiag(kError, "TFormula", "String [ with 1 unbalanced chars.");
```
ferdymercury:
```suggestion
   diags.requiredDiag(kError, "TFormula", "String '[' with 1 unbalanced chars.");
```

hist/hist/test/test_TFormula.cxx
```diff
@@ -98,3 +98,17 @@ TEST(TFormulaPolTest, CompoundExpressions)
...
+   TFormula f1("f1", "[");
+   EXPECT_FALSE(f1.IsValid());
+   EXPECT_EQ(f1.GetNdim(), 0);
+   diags.requiredDiag(kError, "TFormula", "String ( with 1 unbalanced chars.");
```
ferdymercury:
```suggestion
   diags.requiredDiag(kError, "TFormula", "String '(' with 1 unbalanced chars.");
```

hist/hist/src/TFormula.cxx
```diff
@@ -1829,7 +1866,7 @@ void TFormula::HandleLinear(TString &formula)
 
 ////////////////////////////////////////////////////////////////////////////////
 ///    Preprocessing of formula
-///    Replace all ** by ^, and removes spaces.
+///    Replace all ** by ^, all ++ by  @, and removes spaces.
```
ferdymercury:
```suggestion
///    Replace all ** by ^, all ++ by @, and removes spaces.
```

hist/hist/src/TFormula.cxx
ferdymercury:
```suggestion
   // "++" (now @ since they we just called ReplaceAll a few lines above) wil be dealt with Handle Linear
```

ferdymercury:
Could someone restart alma9 glitch?
Thks

ferdymercury:
Thanks, looking green now.

dpiparo:
@vepadulano , do you think we can merge this PR?

ferdymercury:
gentle ping also to @lmoneta

hist/hist/src/TFormula.cxx
```diff
@@ -1844,7 +1883,7 @@ void TFormula::PreProcessFormula(TString &formula)
    HandleParamRanges(formula);
    HandleFunctionArguments(formula);
    HandleExponentiation(formula);
-   // "++" wil be dealt with Handle Linear
+   // "++" (now @ since they we just called ReplaceAll a few lines above) wil be dealt with Handle Linear
```
vepadulano:
```suggestion
   // "++" (now @ since we just called ReplaceAll a few lines above) wil be dealt with Handle Linear
```

lmoureaux: One more typo ;)
```suggestion
   // "++" (now @ since we just called ReplaceAll a few lines above) will be dealt with Handle Linear
```

hist/hist/src/TFormula.cxx
```diff
@@ -1844,7 +1883,7 @@ void TFormula::PreProcessFormula(TString &formula)
    HandleParamRanges(formula);
    HandleFunctionArguments(formula);
    HandleExponentiation(formula);
-   // "++" wil be dealt with Handle Linear
+   // "++" (now @ since we just called ReplaceAll a few lines above) will be dealt with Handle Linear
```
ferdymercury:
```suggestion
   // "++" (now @ since we just called ReplaceAll a few lines above) will be dealt with HandleLinear
```

ferdymercury:
> LGTM for the changes, I just noticed a small typo. Also, I think the commit history could be squashed and adjusted a bit, I will let you decide what's the best way.

Thanks.
I fixed commit history. But it's fine for me if you decide it's better to squash all on merge.
Mac26 failure seems unrelated.
So ready to merge from my side

vepadulano (approved):

PR #21127, opened 2026-02-03
[NFC] Apply linter/formatter fixes to data loader files

vepadulano:


guitargeek (approved):
LGTM, it's good to make sure the linter checks are green!

I only have a concern about the Python code formatting, unrelated to this PR: the check in the CI doesn't seem to work correctly. @siliataider, maybe you can understand what's happening there? If I remember correctly, you introduced these checks.

Image

siliataider:
> LGTM, it's good to make sure the linter checks are green!
> 
> I only have a concern about the Python code formatting, unrelated to this PR: the check in the CI doesn't seem to work correctly. @siliataider, maybe you can understand what's happening there? If I remember correctly, you introduced these checks.
> Image

Thanks @guitargeek for catching this! Indeed there was a recent change in the logic of computing the diff, this PR should take care of the issue # #21130

Issue #21080, opened 2026-01-30
Bad interference for TH1.Fit and EnableImplicitMT


VanyaBelyaev:
### Check duplicate issues.

- [x] Checked for duplicates

### Description

Dear Experst 
 
I've observed a and interference between `TH1.Fit` and `EnableImplicitMT`  in PyROOT

ROOT from LCG dev3 cvmfs nightly build 

Reproducer is trivial: 

```
import ROOT, random

xmin = 0
xmax = 1 

h1 = ROOT.TH1F ( 'h1' , '' , 20 , xmin , xmax ) 

for i in range ( 1000 ) : h1.Fill ( random.gauss ( 0 , 1 ) )

def func ( x , pars ) :
    x0  = x [ 0 ]    
    return pars [ 0 ] + pars [ 1 ] * x0 

my_func = ROOT.TF1( 'f1' , func, xmin , xmax , npar = 2 , ndim = 1 )

my_func.SetParNames  ( 'A'  , 'B'  , )
my_func.SetParameter ( 0 ,  1 )
my_func.SetParameter ( 1 , -1 )


## If commented OUT: ok 
ROOT.ROOT.EnableImplicitMT()   ## THIS LINE

r = h1.Fit ( my_func , 'SE0Q' , '' , xmin , xmax ) 

print ( r ) 

```

In case `EnableImplicitMT` is commented out it works nicely:

```
( source /cvmfs/sft.cern.ch/lcg/views/dev3/latest/x86_64-el9-clang19-opt/setup.sh ; python ./hfit.py ) 
````

But if this line is uncommented, it hangs....

 

### Reproducer

see above 

### ROOT version

ROOT from LCG dev3 nighly slot i cvmfs 

### Installation method

LCG nightly builds at cvmfs, dev3 slot  

### Operating system

linux, lxplus

**Comment by ferdymercury** [comment]
Seems a duplicate of https://github.com/root-project/root/issues/20792
(https://root-forum.cern.ch/t/tgraph-fit-usage-with-enableimplicitmt-is-tgraph-fit-thread-safe/64541/4) see comment by @guitargeek

**Comment by VanyaBelyaev** [comment]
surely it migth be somehow related...
    but in my case I do not have results at all...  
    It just hangs forever... and no results produced 
    ...  and the reproducer is much simpler - just trivial.

**Comment by ferdymercury** [comment]
It seems that it works fine in C++ so it might be a Python-GIL issue:

```
#include <TROOT.h>
[...]

PR #21129, opened 2026-02-03
[PyROOT] Fix a GIL deadlock in THx.Fit

hageboeck:
TH1.Fit might call into Python functions, which would require the GIL to
be held. If TH1.Fit holds it, though, the process deadlocks.
Here, the GIL is released before TH1.Fit, which runs fully in C++
anyway.

A test was added with a 30s timeout. It completes in about 1s, unless the
GIL deadlock is maintained.

Fix https://github.com/root-project/root/issues/21080.

### DIFF
--- a/bindings/distrdf/python/DistRDF/Backends/Dask/Backend.py
+++ b/bindings/distrdf/python/DistRDF/Backends/Dask/Backend.py
@@ -323,10 +323,11 @@ class DaskBackend(Base.BaseBackend):
         cumulative_plots: Dict[int, Any] = {}
 
         # Collect all futures in batches that had arrived since the last iteration
-        for batch in as_completed(future_tasks, with_results=True).batches():
-            for future, result in batch:
-               merged_results = reducer(merged_results, result) if merged_results else result
-            
+        for batch in as_completed(future_tasks, with_results=False).batches():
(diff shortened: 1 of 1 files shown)


guitargeek:
Nice, and compute intense functions like these is exactly when the GIL should be released to enable Python multithreading. Did you check the Pythonization also works for TH1-derived classes?

hageboeck:
> Nice, and compute intense functions like these is exactly when the GIL should be released to enable Python multithreading. Did you check the Pythonization also works for TH1-derived classes?

We checked in the interpreter:
```py
>>> import ROOT
>>> ROOT.TH2.Fit.__release_gil__
True
```

vepadulano:
I've taken a look at the MacOS failures. There seems to be some other kind of interference on that platform for two different types of functions that release the GIL, namely the distributed RDataFrame execution and now TH1.Fit. Needs investigation

guitargeek:
Some of the `ruff` suggestions are probably irrelevant now that https://github.com/root-project/root/commit/647bf1ac737e4feef188c6e907e92e509dde3e64 was merged.

dpiparo:
Thanks for this fix! maybe something to `/backport to 6.38` after merging?

PR #21131, opened 2026-02-03
[ntuple] Add ctx manager to RNTupleReader and update RNTupleWriter's

silverweed:
Now both the Reader and Writer's context objects will close the underlying file upon `__exit__`. The object will raise an exception if accessed after the `with` statement.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
@@ -126,13 +126,58 @@ def pythonize_RNTupleModel(klass):
     klass.MakeField = MethodTemplateGetter(klass.MakeField, _RNTupleModel_MakeField)
 
 
+# Wrapper class used for RNTupleReader and RNTupleWriter.
+# It deletes the underlying smart pointer on context manager exit and ensures that the inner object becomes
+# inaccessible by raising an error every time an attribute of the object is accessed.
+# It also raises an error if `with` statements using the same object are nested.
+# This is a generic class and can in principle be used with any class that needs this behavior.
+class RNTupleContextWrapper:
+    def __init__(self, inner, pretty_name, on_ctx_enter = None, on_ctx_exit = None):
+        self._inner = inner
+        self._pretty_name = pretty_name
+        self._closed = False
+        self._in_context = False
+        self._on_ctx_enter = on_ctx_enter
+        self._on_ctx_exit = on_ctx_exit
+
+    def __getattribute__(self, name):
+        if name.startswith('_'):
+            return super().__getattribute__(name)
+
+        if super().__getattribute__("_closed"):
+            raise RuntimeError(
+                f"cannot access {super().__getattribute__('_pretty_name')} after the `with` statement is exited"
+            )
+        return super().__getattribute__("_inner").__getattribute__(name)
+
+    def __enter__(self, *args):
+        if self._on_ctx_enter:
+            self._on_ctx_enter(self._inner)
+        if self._closed:
+            raise RuntimeError(f"cannot reuse {self._pretty_name} in multiple `with` statements")
+        if self._in_context:
+            raise RuntimeError(f"cannot nest `with` statements using the same {self._pretty_name}")
+
+        self._in_context = True
+        return self
+
+    def __exit__(self, *args):
+        assert self._in_context and not self._closed
+        if self._on_ctx_exit:
+            self._on_ctx_exit(self._inner)
+        self._in_context = False
+        self._closed = True
+        self._inner.__smartptr__().reset()
+        return False
+   
+
 def _RNTupleReader_Open(maybe_model, *args):
     if hasattr(type(maybe_model), "__cpp_name__") and type(maybe_model).__cpp_name__ == "ROOT::RNTupleModel":
         # In Python, the user cannot create REntries directly from a model, so we can safely clone it and avoid destructively passing the user argument.
         maybe_model = maybe_model.Clone()
     import ROOT
 
-    return ROOT.RNTupleReader._Open(maybe_model, *args)
+    return RNTupleContextWrapper(ROOT.RNTupleReader._Open(maybe_model, *args), "RNTupleReader")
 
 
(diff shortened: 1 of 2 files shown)


bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
```diff
@@ -141,6 +155,17 @@ def _RNTupleReader_LoadEntry(self, *args):
     return self._LoadEntry(*args)
 
 
+def _RNTupleReader_enter(reader):
+    reader._active_context = CloseableContext(reader, "RNTupleReader")
```
vepadulano: As an extra safe guard, we should check that this reader does not already have an active context. The language allows for syntax like
```python
reader = RNTupleReader.Open(...)
with reader as f:
    with reader as f2:
        ....
```

silverweed: What's the expectation in this case? Should the inner `__enter__` and `__exit__` be a no-op?

vepadulano: There is no standard practice, it's up to the implementer. I can imagine we could either make it no-op, warn or raise an error.

tree/ntuple/test/ntuple_basics.py
```diff
@@ -22,12 +22,20 @@ def test_write_read(self):
...
-        reader.LoadEntry(0, entry)
-        self.assertEqual(entry["f"], 42)
-        self.assertEqual(entry["mystr"], "string stored in RNTuple")
+        # Accessing the writer after the context manager is an error
+        with self.assertRaisesRegex(RuntimeError, "cannot access RNTupleWriter after"):
+            writer.GetNEntries()
+
+        with ROOT.RNTupleReader.Open("ntpl", "test_ntuple_py_write_read.root") as reader:
+            self.assertEqual(reader.GetNEntries(), 1)
+            entry = reader.CreateEntry()
```
vepadulano, line 31: What is the behaviour of the `entry` variable after the `reader` has been made inactive? That should be tested for both reading and writing.

tree/ntuple/test/ntuple_basics.py
```diff
@@ -100,3 +108,34 @@ class WrongClass: ...
             entry = writer.CreateEntry()
             with self.assertRaises(TypeError):
                 entry["mystr"] = WrongClass()
+
+    def test_nested_ctxmanager(self):
+        """Nesting context managers of the same object is a no-op and gives a warning"""
```
vepadulano: > gives a warning

This does not actually happen, right?

tree/ntuple/test/ntuple_basics.py
```diff
@@ -100,3 +108,34 @@ class WrongClass: ...
...
+
+        try:
+            fileName = "test_ntuple_nested_ctxmanager_py.root"
+            model = ROOT.RNTupleModel.Create()
+            model.MakeField["int"]("f")
+            writer = ROOT.RNTupleWriter.Recreate(model, "ntpl", fileName)
```
vepadulano, line 127: In this scenario, `writer` and `reader` are not made inactive after the `with` context, I just wanted to make sure that this is what we want and it doesn't go unnoticed.

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_rntuple.py
```diff
@@ -126,6 +126,43 @@ def pythonize_RNTupleModel(klass):
...
+            context = owner._active_context = CloseableContext(owner, pretty_name)
+        else:
+            context._nesting += 1
+        return context
+        
+    @staticmethod
+    def exit(owner):
+        context = owner._active_context
+        if context._nesting == 0:
+            context._closed = True
```
vepadulano: Just out of curiosity, why did you go for the no-op approach? I would have instead chosen to go for the option of raising an error when nesting contexts. This doesn't mean you should change this choice, I think it's still ok.

silverweed: Yeah, thinking again about it raising an error is probably the better idea. I think I'll change it

tree/ntuple/test/ntuple_basics.py
silverweed: I think it's fine since this test is only testing that nesting `with` statements works as intended and, since I'm changing it to raise an exception, we don't care much about the state of the objects after it

vepadulano: Indeed, for this test it won't matter. I guess the bigger question is what happens in a scenario such as:
```python
writer = ROOT.RNTupleWriter.Recreate(model, "ntpl", fileName)
with writer as ctxt:
    # do stuff

# use writer afterwards
```

Here we have the two variables `writer` and `ctxt`, both available to the user. While `ctxt` will be unusable thanks to the `ClosableContext` logic, `writer` IIUC will just be an empty `unique_ptr`, right?

tree/ntuple/test/ntuple_basics.py
silverweed: All values that were already loaded in the entry stay alive and readable (they're stored in shared_ptr internally) but you can't load or write new entries (since the reader/writer is gone).
I'll add an assert for that.

tree/ntuple/test/ntuple_basics.py
silverweed: Good point; that's probably what would happen...not sure how to handle that case.
But isn't it a "wrong" pattern in python?

silverweed: I changed the code to return a `CloseableWrapper` directly from the reader/writer's factory methods. This wrapper forwards everything to the wrapped reader/writer but it also implements `__enter__` and `__exit__` to make it sure that 
a) we cannot nest `with` statements with the same object, and
b) we properly close the wrapped object and cannot access it after the `with` statement ends

(should I squash the commits?)

vepadulano (approved):
Thank you! LGTM once tests pass.

Issue #11004, opened 2022-07-20
TCanvas size is inconsistent in batch vs non-batch mode


ferdymercury:
- [x] Checked for duplicates

### Describe the bug
I see one small issue when saving a 600x600 TCanvas window as png.

I get:
- Interactive mode: 598x571 pixels
- Batch mode: 596x572 pixels

### Expected behavior
No pixel difference is found between running batch or not. Right now there is an offset of 1 up to 2 pixels.

### To Reproduce

```cpp
void test() {
    TCanvas *c = new TCanvas("c1","c1",600,600);
    c->SaveAs("c1.png");
}
//root -l test.C -b -q
//root -l test.C -q
```

### Setup
```
   ------------------------------------------------------------------
  | Welcome to ROOT 6.27/01                        https://root.cern |
  | (c) 1995-2021, The ROOT Team; conception: R. Brun, F. Rademakers |
  | Built for linuxx8664gcc on Jan 11 2022, 18:39:08                 |
  | From heads/master@v6-25-01-2870-gdac9b6398d                      |
  | With c++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0                     |
  | Try '.help', '.demo', '.license', '.credits', '.quit'/'.q'       |
   ------------------------------------------------------------------
```

couet:
I do not think there is a lot we can do. The png file is not a screen copy and some rouning issues in completely different code might be possible. I propose to close this issue.

ferdymercury:
In my other computer, I get now a difference of 4 pixels using latest ROOT master. So it seems to me that this is not a small rounding error.

- Interactive mode: 598x576 pixels
- Batch mode: 596x572 pixels


Could it be just an outdated calibration from many years ago? See this https://github.com/root-project/root/commit/c3de92f4d1ad4372febc81955a877345d4de3a9c

Check out this code below. It seems there is always the same offset on my current computer. Whenever I run in batch mode, I get 4 pixels difference wrt batch mode. My hypothesis is that this could solved if those numbers are adapted, potentially depending on Windows / Mac with an if. It could be even added as a roottest.

```cpp
// Run either as root -l test_pixels.cpp -b -q
// Or as xvfb-run root -l test_pixels.cpp -q
#include <TCanvas.h>
#include <TROOT.h>
#include <iostream>
void test_pixels() {
    for(Int_t w=500;w<=900;w++) {
        for(Int_t h=500;h<=600;h++) {
[...]

PR #21136, opened 2026-02-03
[graf2d] Fix Setting window size in batch mode

rlalik:
The current implementation doesn't make sense at all? Was it some accidentally committed code?

@linev It was your commit? Was it mistake?

```
771c44692bcb graf2d/gpad/src/TCanvas.cxx (Sergey Linev      2023-07-06 16:42:42 +0200 2151)    if (fBatch && !IsWeb())
d023aecbdda4 graf2d/gpad/src/TCanvas.cxx (Sergey Linev      2020-05-04 19:06:00 +0200 2152)       SetCanvasSize((ww + fCw) / 2, (wh + fCh) / 2);
d023aecbdda4 graf2d/gpad/src/TCanvas.cxx (Sergey Linev      2020-05-04 19:06:00 +0200 2153)    else if (fCanvasImp)
d023aecbdda4 graf2d/gpad/src/TCanvas.cxx (Sergey Linev      2020-05-04 19:06:00 +0200 2154)       fCanvasImp->SetWindowSize(ww, wh);
```
- [x] tested changes locally

### DIFF
--- a/graf2d/gpad/src/TCanvas.cxx
+++ b/graf2d/gpad/src/TCanvas.cxx
@@ -2149,7 +2149,7 @@ void TCanvas::SetWindowPosition(Int_t x, Int_t y)
 void TCanvas::SetWindowSize(UInt_t ww, UInt_t wh)
 {
    if (fBatch && !IsWeb())
-      SetCanvasSize((ww + fCw) / 2, (wh + fCh) / 2);
+      SetCanvasSize(ww, wh);
    else if (fCanvasImp)
       fCanvasImp->SetWindowSize(ww, wh);
 }


linev:
@rlalik 

No, this kind of calculation was there since a while. See:

https://github.com/root-project/root/blob/v6-24-00-patches/graf2d/gpad/src/TCanvas.cxx#L2201

I only add special handling for web case

linev:
I have no idea why code looks like this but I prefer not to touch it.
SetWindowSize has no real meaning for the batch - one should use SetCanvasSize instead.

rlalik:
Well, in the batch mode there is indeed no window but this code changes the canvas size. And I would say, for compatibility reasons, should work properly. Otherwise macro run in batch mode will produce wrongly sized canvases.

ferdymercury:
related: https://github.com/root-project/root/issues/11004 and https://github.com/root-project/root/commit/c3de92f4d1ad4372febc81955a877345d4de3a9c

ferdymercury:
> @linev It was your commit? Was it mistake?

It was this commit by @couet suggested by Wile in 2017

https://github.com/root-project/root/commit/281c67087ce00514a95aafa13a9484d6a437bffe
https://root-forum.cern.ch/t/tcanvas-setwindowsize-extension-proposal/26226

According to Wile, it's because:

[TCanvas](https://root.cern.ch/doc/master/classTCanvas.html) shows the typical usage “c->SetWindowSize(w + (w - c->GetWw()), h + (h - c->GetWh()));”, so we need to “recalculate” parameters’ values (in order to get the requested original “w” and “h”).

rlalik:
I do not understand the logic behind it. In the batch mode `SetCanvasSize((ww + fCw) / 2, (wh + fCh) / 2);` causes that the new canvas size has size of half window. I don't know how it explains the message in commit https://github.com/root-project/root/commit/281c67087ce00514a95aafa13a9484d6a437bffe. Perhaps it worked with some other logic which was changed in the meantime?

Did Wile test his suggestion? Is there any trace in the issue or in the forum about that? Perhaps there was miscommunication and Wile's suggestion as implemented incorrectly?

rlalik:
Ok, now I indeed understand, because `c->SetWindowSize(c->GetWindowWidth() + (c->GetWindowWidth() - c->GetWw()), c->GetWindowHeight() + (c->GetWindowHeight() - c->GetWh()));` the Ww() is 0 so we set double of window width and we need to divide by half later.

But you see that this function will work properly only in this specific case, someone need to call it in this particular way, otherwise it will give wrong result. This is so wrong in my opinion. The logic is flaved here and should be handled different way.

ferdymercury:
> But you see that this function will work properly only in this specific case, someone need to call it in this particular way, otherwise it will give wrong result. This is so wrong in my opinion. The logic is flaved here and should be handled different way.

Yes, I agree all this is really messy.

PR #21137, opened 2026-02-03
Move ROOT ML data loader to tree

vepadulano:
The native ROOT data loading for ML training functionality is related to IO and
processing of structured data. It only depends on the ROOTDataFrame component.
As such, the following decisions have been made for this part of the codebase:

- C++ code
  - It is meant for internal use only
  - It is hosted in tree/ml in the repository
  - It is written in the ROOT[::Experimental]::Internal::ML namespace
  - The headers will appear in the include/ROOT/ML directory in the install tree

- Python code
  - It contains the public API
  - Public functions appear under ROOT[.Experimental].ML module

All public references to the name 'batch generator' are removed to avoid mixing
responsibilities: RBatchGenerator is just the name of an internal class, which
is never exposed to the user directly.

Update tests and tutorials accordingly.

Apply formatting changes to C++ and Python files.

Add code owners to new directory.


io/ml/inc/ROOT/DataLoader/RBatchLoader.hxx
```diff
@@ -12,8 +12,8 @@
  * For the list of contributors see $ROOTSYS/README/CREDITS.             *
  *************************************************************************/
 
-#ifndef TMVA_RBATCHLOADER
-#define TMVA_RBATCHLOADER
+#ifndef IO_ML_INTERNAL_RBATCHLOADER
+#define IO_ML_INTERNAL_RBATCHLOADER
```
pcanal: `inc/Internal` is a new convention.  Are those files installed or only used during builds (in which case they should be in `io/ml/res`).  Since they are `.hxx` and in the `ROOT` namespace, if they are installed, they should probably go into the `include/ROOT` subdirectory (which currently only has one subdirectory (`RField`).

vepadulano: After offline discussion we concluded that:

* All the headers actually need to be installed so that the Python side can find the classes
* It seems like `include/ROOT/DataLoader` is a sensible target directory

bindings/pyroot/pythonizations/test/CMakeLists.txt
```diff
@@ -180,9 +180,9 @@ ROOT_ADD_PYUNITTEST(pyroot_tcomplex tcomplex_operators.py)
 # Tests with memory usage
 ROOT_ADD_PYUNITTEST(pyroot_memory memory.py)
 
-# rbatchgenerator tests
-if (tmva)
-  ROOT_ADD_PYUNITTEST(batchgen rbatchgenerator_completeness.py PYTHON_DEPS numpy tensorflow torch)
+# ML dataloader tests
+if (dataframe)
+  ROOT_ADD_PYUNITTEST(io_ml_dataloader io_ml_dataloader.py PYTHON_DEPS numpy tensorflow torch)
```
siliataider:
```suggestion
  ROOT_ADD_PYUNITTEST(ml_dataloader ml_dataloader.py PYTHON_DEPS numpy tensorflow torch)
```

PR #21141, opened 2026-02-04
[interop] Avoid usage of `<filesystem>`

hahnjo:
It requires linking `libstdc++fs` for older versions of the standard library, and the same function is provided by LLVM.


hahnjo:
Upstream PR: https://github.com/compiler-research/CppInterOp/pull/797 (confirmed to work locally with Clang and `libstdc++` from GCC 8.5.0)

hahnjo:
As discussed, this was only to test the upstream PR in ROOT.

PR #21150, opened 2026-02-04
[Python] Move JupyROOT to be private submodule of `ROOT`

guitargeek:
JupyROOT should not be accessed by the user, and not having it as a separate package makes packaging cleaner. In the end, ROOT should only ship the ROOT package.

The PR also contains some commits to defer the import of cppyy and the initialization of the C++ runtime to the last possible moment, so we're not (partially) initializing it when importing the `ROOT` module. This is what makes it possible to have clean Python-only `ROOT` submodules like `_jupyroot` without the unwanted sideeffects of initializing the C++ runtime on import.


roottest/python/JupyROOT/ipythondir/kernels/root/kernel.json
```diff
@@ -4,7 +4,7 @@
  "argv": [
   "python",
   "-m",
-  "JupyROOT.kernel.rootkernel",
+  "ROOT._jupyroot.kernel.rootkernel",
```
pcanal: I am confused.  Since this is part of the testing, it looks like the users might be doing the same thing.  If that is correct was is the backward compatibility plan?  If that is not correct, what is the indirect way that user take advantage of `JupyROOT`?

guitargeek: The user should not write his own kernel file: we ship it via `etc/notebook/kernels/root/kernel.json.in`.

This kernel file in `roottest` is probably a remnant of the JupyROOT development process and I'll see if I can remove it.

bindings/pyroot/pythonizations/python/ROOT/__init__.py
```diff
@@ -100,12 +95,14 @@ def _can_be_module(obj) -> bool:
     return False
 
 
-def _lookup_root_module(fullname: str) -> Optional[Union[types.ModuleType, cppyy.types.Scope]]:
+def _lookup_root_module(fullname: str):
```
vepadulano: The return type hint in principle could stay there if we had a `from __future__ import annotations` at the top of the module

PR #21157, opened 2026-02-04
[CPyCppyy] Do not decref a constructor result on Windows

guitargeek:
CPPMethod::ExecuteFast() ends with a Windows-only block that turns a
result returned with a pending Python exception into a failure, to cover
the PyException throw not propagating there:

    #ifdef _WIN32
        if (PyErr_Occurred()) {
            Py_XDECREF(result);
            result = nullptr;
        }
    #endif

ExecuteFast() is shared by every CPPMethod subclass, including
CPPConstructor, whose executor is ConstructorExecutor. That one does not
return a PyObject* at all: it returns the address of the newly
constructed C++ object, cast to PyObject* for CPPConstructor::Call() to
unpack. Decref'ing that address treats arbitrary C++ object memory as a
PyObject and corrupts the heap.

Ask the method whether its executor hands back a real PyObject* before
dropping a reference. The object is leaked instead, which is acceptable
on the error path of a call that is about to be reported as failed.

🤖 Done with the help of AI

### DIFF
--- a/bindings/pyroot/cppyy/CPyCppyy/src/CPPConstructor.h
+++ b/bindings/pyroot/cppyy/CPyCppyy/src/CPPConstructor.h
@@ -22,6 +22,7 @@ public:
 
 protected:
     bool InitExecutor_(Executor*&, CallContext* ctxt = nullptr) override;
+    bool ResultIsPyObject() const override { return false; }
 };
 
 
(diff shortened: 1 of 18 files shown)


aaronj0 (approved):
Changes on the cppyy side LGTM!

guitargeek:
Thanks for the review! I removed all non-cppyy commits from this PR, so this part can already been merged.

The part about using `-W error` for Python tests is anyway still blocked by deprecation warnings in the SOFIE parsers, which will go away soon:

   * https://github.com/root-project/root/pull/22734

Once they are gone, I'll re-start the effort with the Python development module.

Issue #21159, opened 2026-02-04
[RF] Unexpected result with RooFFTConvPDF after ROOT 6.32


guitargeek:
Another corner case where the new RooFit evaluation backend fails to give the right results was reported on the ROOT forum:

https://root-forum.cern.ch/t/unexpected-result-with-roofftconvpdf-after-root-6-32/64574/4

Reproducer:
```c++
using namespace RooFit;

RooRealVar obs{"mass", "", 5500, 6100};
RooRealVar mean{"mean", "", 5800.0, 5795, 5805};
RooRealVar sigma{"sigma", "", 8.1, 1, 30};

RooGaussian sigpdf{"sigpdf", "", obs, mean, sigma};

RooRealVar argus_par{"argus_par", "", -55.0};
RooRealVar argus_m0{"argus_m0", "", 5800 - 135};
RooArgusBG argus{"argus", "", obs, argus_m0, argus_par};
RooGaussian resolution{"resolution", "", obs, RooFit::RooConst(0), sigma};
RooFFTConvPdf conv{"conv", "", obs, argus, resolution};

RooAddPdf model{"model", "", {sigpdf, conv}, {RooConst(0.3)}};

std::unique_ptr<RooDataSet> data{model.generate(obs, 1000)};

std::unique_ptr<RooFitResult> fit_result{model.fitTo(*data, Save(true))};

// This works!
// std::unique_ptr<RooFitResult> fit_result{model->fitTo(*data, Save(true), EvalBackend("legacy"))};

fit_result->Print("v");
```
The output is:
```txt
  RooFitResult: minimized FCN value: 1.1609e-12, estimated distance to minimum: 5.82024e-16
                covariance matrix quality: Not calculated at all
                Status : MINIMIZE=-1 HESSE=302

    Constant Parameter    Value
  --------------------  ------------
              argus_m0    5.6650e+03
             argus_par   -5.5000e+01

    Floating Parameter  InitialValue    FinalValue +/-  Error     GblCorr.
  --------------------  ------------  --------------------------  --------
                  mean    5.8000e+03    5.8000e+03 +/-  0.00e+00  <none>
                 sigma    8.1000e+00    8.0336e+00 +/-  0.00e+00  <none>
```
With the `"legacy"` evaluation backend, the fit succeeds and both the `MINIMIZE` and `HESSE` status codes are zero, indicating that there is no error.

PR #21161, opened 2026-02-04
[RF] Ensure RooHistPdf values are positive in `RooHistPdf::doEval()`

guitargeek:
This is to be consistent with the old `RooHistPdf::evaluate()` code
path.

Closes https://github.com/root-project/root/issues/21159, where floating point precision effects made the value
fluctuate slightly below zero, resulting in NaN values when the
RooFFTConvPdf (which internally uses a `RooHistPdf` to cache values) is
used in a `RooAbsPdf`.

Closes https://github.com/root-project/root/issues/21159.


guitargeek:
backport to 6.38, 6.36

PR #21168, opened 2026-02-05
[hist] Move underflow bin before normal bins

hahnjo:
This is a bit more complicated in the code because the shift only applies if the underflow bin is enabled. It is also slower in the existing microbenchmarks, in particular with Clang. However, we need to make this change in order to support UHI's `values(flow=True)` with zero-copy buffers in the future and have the underflow bin at "the right place." Otherwise the design currently treats the memory layout as an internal implementation detail that users will not notice when using `RBinIndex`.

### DIFF
--- a/hist/histv7/inc/ROOT/RRegularAxis.hxx
+++ b/hist/histv7/inc/ROOT/RRegularAxis.hxx
@@ -83,9 +83,10 @@ public:
 
    /// Compute the linarized index for a single argument.
    ///
-   /// The normal bins have indices \f$0\f$ to \f$fNNormalBins - 1\f$, the underflow bin has index
-   /// \f$fNNormalBins\f$, and the overflow bin has index \f$fNNormalBins + 1\f$. If the argument is outside the
-   /// interval \f$[fLow, fHigh)\f$ and the flow bins are disabled, the return value is invalid.
+   /// If flow bins are disabled, the normal bins have indices \f$0\f$ to \f$fNNormalBins - 1\f$. Otherwise the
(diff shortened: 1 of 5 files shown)


hahnjo:
gentle ping - I start to have a couple of changes that build on this PR, particularly related to slicing etc. :smiley:

hist/histv7/inc/ROOT/RRegularAxis.hxx
```diff
@@ -104,28 +105,41 @@ public:
...
-      return {bin, bin < fNNormalBins};
+      if (bin >= fNNormalBins) {
+         // Index is out of range and invalid.
+         return {bin, false};
+      }
+      // If the underflow bin is enabled, shift the normal bins by one.
+      if (fEnableFlowBins) {
+         bin += 1;
+      }
+      return {bin, true};
```
jblomer, line 142: I wonder where the performance is lost in the micro-benchmarks. Is the code sensitive to this optimization (perhaps done by the compiler anyway...)

```
   return {bin + fEnableFlowBins, bin < fNNormalBins};
```

hahnjo: Note that the overload accepting `RBinIndex` is not the performance-critical one, during filling the overload with `double x` is used. It's possible (and I think correct) that this code could be optimized, it wasn't a target yet.

For the other overload, I had tried locally to just add `fEnableFlowBins`, but it doesn't change performance (the compiler already does it). My guess would be that the performance loss is just the loading of `fEnableFlowBins` and executing dependent code, which wasn't needed before in the code path for normal bins...

Issue #14611, opened 2024-02-07
[DF] ProgressBar destroys formatted output streams


martamaja10:
### Explain what you would like to see improved and how.

As reported by a user on the forum: https://root-forum.cern.ch/t/rdf-progress-bar-used-together-with-formatted-output-streams/58064, when ProgressBar is being used, the custom formatting of output streams is affected. 

### ROOT version

6.30.04

### Installation method

any

### Operating system

any

PR #21170, opened 2026-02-05
[RDF] Restore stream precision and width when ProgressBar in use.

hageboeck:
As a followup to #14611, restore also precision and width of the streams.

This was reported in:
https://root-forum.cern.ch/t/rdf-progress-bar-changes-printout-precision/


martamaja10 (approved):
Thanks Stephan!

hageboeck:
> I was also working on this locally and I propose a different approach, see https://github.com/root-project/root/pull/21171

Much nicer, thanks!

Issue #21165, opened 2026-02-05
RDataFrame progress bar changes stream output precision


vepadulano:
### Check duplicate issues.

- [ ] Checked for duplicates

### Description

RDataFrame progress bar sets an output precision and then leaves the output stream in a bad state, see reproducer below.

### Reproducer

```cpp
#include <ROOT/RDataFrame.hxx>
#include <ROOT/RDFHelpers.hxx>
#include <iostream>

int ret42 () {return 42;}
int ret1 () {return 1;}

int main()
{ 
   std::vector<double> values{13922869, 13599, 277, 40003186, 68.5966, 999.008, 40107.1};
   std::cout << "Before:\n{";
   for(auto v: values)
	   std::cout << v << ", ";
   std::cout << "}\n";

   ROOT::RDF::RNode d = ROOT::RDataFrame(10000);
   ROOT::RDF::Experimental::AddProgressBar(d);
   d.Count().GetValue();

   std::cout << "After:\n{";
   for(auto v: values)
           std::cout << v << ", ";
   std::cout << "}\n";
}
```

```
Before:
{1.39229e+07, 13599, 277, 4.00032e+07, 68.5966, 999.008, 40107.1, }
[Total elapsed time: 0:00m  processed files: 0 / 0  processed evts: 10000 / 10000]                                                                             
After:
{1.4e+07, 1.4e+04, 2.8e+02, 4e+07, 69, 1e+03, 4e+04, }
```

### ROOT version

Any

### Installation method

Any

### Operating system

Any

PR #21171, opened 2026-02-05
[df] Restore all attributes of std::cout after printing progress

vepadulano:
Fixes https://github.com/root-project/root/issues/21165


dpiparo:
/backport to 6.36
as a test to see if labels are transferred

PR #21181, opened 2026-02-06
[RDF] Improvements for the progress bar

hageboeck:
I realised that we have enough information to ensure that the progress bar doesn't jump around as new files are opened:
So far, when a file was split in ranges, RDF was using the upper end of each range to update the estimated number of events, so the completion times are wrong until the last cluster is opened. Furthermore, unopened files weren't taken into account to estimate the completion.

With this PR, by getting the number of entries from the TTree or RNT datasources, and taking into account how many files haven't been opened yet, one can significantly improve this. Now, the bar doesn't jump backwards (but it might jump a bit faster/slower when the next files opened have significantly less/more events than the previous ones).

Furthermore, I did a lot of refactoring, trying to outline as much as possible into the `.cxx`, and reducing the times that locks are taken etc. (see list at the end). Lastly, when the output doesn't go to a tty, the width of the line and frequency of updates is reduced to not unnecessarily clutter the files.

### New look
```
|=>          |   [Elapsed: 0:10m  files: 3 / 4  events: 42886000 / (184621239 + x)  4.29e+06 evt/s  remaining ca.: 0:47m]
|===>        |   [Elapsed: 0:20m  files: 3 / 4  events: 94453000 / (184621239 + x)  4.72e+06 evt/s  remaining ca.: 0:32m]
|======>     |   [Elapsed: 0:30m  files: 3 / 4  events: 144333000 / (184621239 + x)  4.81e+06 evt/s  remaining ca.: 0:21m]
|========>   |   [Elapsed: 0:40m  files: 4 / 4  events: 189735000 / 246161652  4.74e+06 evt/s  remaining ca.: 0:11m]
|==========> |   [Elapsed: 0:50m  files: 4 / 4  events: 226096000 / 246161652  4.52e+06 evt/s  remaining ca.: 0:04m]
[Total elapsed time: 0:56m  processed files: 4 / 4  processed evts: 246161652 / 246161652]
```
Check the times. The first two estimates are 57s and 52s, and it completes in 56s.
It also says `+ x` for the total number of events until all files have been opened.

<img width="1603" height="25" alt="image" src="https://github.com/user-attachments/assets/386830d0-c50c-4ce5-8c7e-394d4940c288" />

### Old look
```
|>                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |   [Elapsed time: 0:01m  processing file: 3 / 4  processed evts: 1000 / 24650850  6.03e+02 evt/s 11:20:54h  remaining time (per file being processed)]   
|=======================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================>                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |   [Elapsed time: 0:02m  processing file: 3 / 4  processed evts: 4529000 / 24650850  2.26e+06 evt/s 0:08m  remaining time (per file being processed)]   
|======================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================================>                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |   [Elapsed time: 0:03m  processing file: 3 / 4  processed evts: 10675000 / 24650850  3.56e+06 evt/s 0:03m  remaining time (per file being processed)]   
```

Check again the times: For a workflow of about 1 minute, the estimated times in the first 3 updates are: 11h, 10s and 6s. I understand that this is per file, but the above seems to do a better job.
And the width when written to a file seems to be somehow arbitrary. 🙂 

<img width="2092" height="26" alt="image" src="https://github.com/user-attachments/assets/66bc1134-f090-42d3-9c7b-5aaa585e71dd" />


## More details of the refactoring
- Handle locking logic for updates with a single RAII, update locks to C++17.
- Remove a lock and mutex that didn't have any effect.
- Reduce repeated function calls to functions that hold locks.
- Simplify computation of average number of events.
- Relax memory order of the atomics to what's necessary.
- Outline as many functions as possible. Since RDFHelpers.hxx goes
  through JITting and pcms, it gets compiled frequently, so outlining is
  probably beneficial.
- Make a lot of members const to avoid unintenional modifications. This
  might be interesting because it's an MT context.
- Collect helper functions in one anonymous namespace.
- Remove a constructor argument that was ignored.
- Bring back evts/s to the final line that's printed.

### DIFF
--- a/tree/dataframe/inc/ROOT/RDFHelpers.hxx
+++ b/tree/dataframe/inc/ROOT/RDFHelpers.hxx
@@ -310,8 +310,6 @@ void AddProgressBar(ROOT::RDataFrame df);
 /// @param nThread Number of threads that share a TH3D.
 void ThreadsPerTH3(unsigned int nThread = 1);
 
-class ProgressBarAction;
-
 /// RDF progress helper.
 /// This class provides callback functions to the RDataFrame. The event statistics
 /// (including elapsed time, currently processed file, currently processed events, the rate of event processing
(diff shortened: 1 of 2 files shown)


tree/dataframe/inc/ROOT/RDF/RSampleInfo.hxx
```diff
@@ -37,6 +37,7 @@ class RSampleInfo {
    std::pair<ULong64_t, ULong64_t> fEntryRange;
 
    const ROOT::RDF::Experimental::RSample *fSample = nullptr; // non-owning
+   ULong64_t fTotalEntries = 0;
```
vepadulano: By lookin at the changes in `RNTupleDS.cxx` and `RTTreeDS.cxx` I see that this is the total number of entries in the current file, I would add a comment to explain it.

tree/dataframe/inc/ROOT/RDF/RSampleInfo.hxx
```diff
@@ -121,9 +122,13 @@ public:
    /// Multiple multi-threading tasks might process different entry ranges of the same sample.
    std::pair<ULong64_t, ULong64_t> EntryRange() const { return fEntryRange; }
 
-   /// @brief Return the number of entries of this sample that is being taken into consideration.
+   /// @brief Return the number of entries of this range of the sample.
    ULong64_t NEntries() const { return fEntryRange.second - fEntryRange.first; }
 
+   /// Return the total number of entries in the underlying dataset. This can be larger than NEntries() if
+   /// the sample is split in multiple ranges.
+   ULong64_t NEntriesTotal() const { return std::max(fTotalEntries, fEntryRange.second); }
```
vepadulano, line 131: In which situation can it be that `fTotalEntries < fEntryRange.second`?

tree/dataframe/inc/ROOT/RDFHelpers.hxx
```diff
@@ -333,67 +333,49 @@ class ProgressBarAction;
...
+   bool const fUseShellColours;
 
    std::atomic<std::size_t> fProcessedEvents{0};
    std::size_t fLastProcessedEvents{0};
-   std::size_t fIncrement;
-
-   mutable std::mutex fSampleNameToEventEntriesMutex;
-   std::map<std::string, ULong64_t> fSampleNameToEventEntries; // Filename, events in the file
+   std::size_t const fIncrement;
+   unsigned int const fNColumns;
```
vepadulano, line 347: There is a core guideline to avoid `const` data members in case the class is copyable/moveable in any way https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c12-dont-make-data-members-const-or-references-in-a-copyable-or-movable-type . Given that this is not well defined in this class, we should do something about it. If you want to keep the `const`ness (although I don't grasp the full reason behind it), I would suggest to implement rule of five and make the class non copyable and non movable

tree/dataframe/inc/ROOT/RDFHelpers.hxx
```diff
@@ -333,67 +333,49 @@ class ProgressBarAction;
...
+   std::chrono::time_point<std::chrono::system_clock> fLastPrintTime = fBeginTime;
+   std::chrono::seconds const fPrintInterval;
 
-   std::mutex fPrintMutex;
-   bool fIsTTY;
-   bool fUseShellColours;
+   std::mutex fUpdateMutex; // Mutex to ensure that only one thread updates the progress bar
 
-   std::shared_ptr<TTree> fTree{nullptr};
+   mutable std::mutex fSampleNameToEventEntriesMutex;          // Mutex to protect access to the below map
```
vepadulano, line 361:
```suggestion
   mutable std::mutex fSampleNameToEventEntriesMutex;          // Mutex to protect access to the map below
```

tree/dataframe/inc/ROOT/RDFHelpers.hxx
vepadulano: Outside of scope for this PR, but I don't remember why this operator is templated.

TomasDado:
Thanks for these updates, this is great! One comemnt that we discussed at some point, would it be possible to add an option for the users to pass the total number of events (if they know it)? We often need to open each file before we run RDF to build the file lists and the total sum of weights for normalisation. We could also trivially collect the total number of events.

tree/dataframe/inc/ROOT/RDF/RSampleInfo.hxx
hageboeck: When a data source doesn't set the value. So it's not the case for TTree / RNT, but there might be data sources that don't know where they end.

tree/dataframe/inc/ROOT/RDFHelpers.hxx
hageboeck: The class is not copyable / movable because of the mutexes and atomics, but we could of course make it explicit if you would like.

So why are the members `const` now? I'm going by the statement that `const == thread safe` in c++11. Given that for this class there's two mutexes that refer to different members, any future observer would have to check which mutex they need to lock. The `const` members, however, can always be used without any locks.

tree/dataframe/inc/ROOT/RDFHelpers.hxx
hageboeck: Me neither. I'll try to remove it.

hageboeck:
> Thanks for these updates, this is great! One comemnt that we discussed at some point, would it be possible to add an option for the users to pass the total number of events (if they know it)? We often need to open each file before we run RDF to build the file lists and the total sum of weights for normalisation. We could also trivially collect the total number of events.

Hello @TomasDado, yes, it hasn't been forgotten. 🙂
Let's first merge this one, and then we can use some of the infrastructure here to deal with hints by the users.

tree/dataframe/inc/ROOT/RDFHelpers.hxx
hageboeck: Both are correct in English, so I won't go through the effort of pushing another change now. 🙂

tree/dataframe/inc/ROOT/RDFHelpers.hxx
vepadulano: All clear thank you! Yes ideally let's add the special member methods to the class.

tree/dataframe/inc/ROOT/RDFHelpers.hxx
hageboeck: I managed to remove the template and outline the function. In addition, I was also able to remove a friend declaration and another call to an update function that updated internal statistics under lock.

PR #21183, opened 2026-02-06
io: Support 64 bits position/byte-count

pcanal:
Continuation/Replacement of https://github.com/root-project/root/pull/20574 with a new incoming branch name to enable proper testing.

### DIFF
--- a/io/io/src/TBufferFile.cxx
+++ b/io/io/src/TBufferFile.cxx
@@ -3007,14 +3007,15 @@ void TBufferFile::SkipVersion(const TClass *cl)
 
 Version_t TBufferFile::ReadVersion(UInt_t *startpos, UInt_t *bcnt, const TClass *cl)
 {
+   assert((!startpos && !bcnt) || (startpos && bcnt)); // both or none should be set
+
    Version_t version;
 
    if (startpos) {
       // before reading object save start position
       auto full_startpos = fBufCur - fBuffer;
       *startpos = full_startpos <= kMaxCountPosition ? UInt_t(full_startpos) : kOverflowPosition;
-      if (bcnt)
-         fByteCountStack.push_back(full_startpos);
+      fByteCountStack.push_back(full_startpos);
    }
 
    // read byte count (older files don't have byte count)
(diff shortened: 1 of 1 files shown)


jblomer:
The first commit message didn't get reworded, I think. Cf. https://github.com/root-project/root/pull/20574#issuecomment-3856358083

io/io/src/TBufferFile.cxx
```diff
@@ -3327,6 +3461,21 @@ UInt_t TBufferFile::CheckObject(UInt_t offset, const TClass *cl, Bool_t readClas
...
+/// \return The position (cntpos) where the byte count should be stored later,
+///         or kOverflowPosition if the position exceeds kMaxCountPosition
+
+UInt_t TBufferFile::ReserveByteCount()
+{
+   // reserve space for leading byte count
+   auto full_cntpos = fBufCur - fBuffer;
+   fByteCountStack.push_back(full_cntpos);
+   *this << (UInt_t)kByteCountMask; // placeholder for byte count
+   return full_cntpos < kMaxCountPosition ? full_cntpos : kOverflowPosition;
```
jblomer:
```suggestion
   return full_cntpos <= kMaxCountPosition ? full_cntpos : kOverflowPosition;
```

io/io/src/TBufferFile.cxx
```diff
@@ -2925,14 +3001,20 @@ void TBufferFile::SkipVersion(const TClass *cl)
 
 ////////////////////////////////////////////////////////////////////////////////
 /// Read class version from I/O buffer.
+/// If passing startpos and bcnt, CheckByteCount must be called later with
```
jblomer, line 3004: Is it valid to set only one of them? If not, we can assert both are set or both are nullptr and simplify a bit the logic.

io/io/src/TBufferFile.cxx
```diff
@@ -3242,8 +3372,12 @@ void TBufferFile::StreamObject(TObject *obj)
...
 void TBufferFile::CheckCount(UInt_t offset)
 {
-   if (IsWriting()) {
-      if (offset >= kMaxMapCount) {
+   // FIXME:
+   // We could continue to issue the error for cases where we don't want to
+   // support storing more than 1GB in a single TBufferFile.
+   const bool bufferLimitedToMaxMapCount = false;
+   if (bufferLimitedToMaxMapCount && IsWriting()) {
+      if (offset >= kMaxRecordByteCount) {
```
jblomer:
```suggestion
      if (offset > kMaxRecordByteCount) {
```

pcanal:
> The first commit message didn't get reworded, I think.

Correct.  I have now pushed the revised version.

io/io/src/TBufferFile.cxx
pcanal: It does not make much sense to only pass one.

pcanal: So I updated to use an assert and less conditionals.

pcanal:
Review continues at https://github.com/root-project/root/pull/21192

PR #21184, opened 2026-02-06
[tmva][sofie] Parse generated code at test runtime

guitargeek:
TMVA SOFIE development is challenging sometimes, because of how the tests are structured.

The tests that covers many possible models imported from ONNX or ROOT have the issue that they includes **all** emitted code in the compiled executables. This means that one gets a build failure on the first model that generated invalid code, and that was it. Therefore, it's difficult to debug what is going wrong.

This commit suggests include the generated code with the interpreter instead. Then, one can check for each individual model if the code was valid, and if not, skip over to the next test a print the emitted code that failed to compile.

It has some performance overhead, but the tests still only take about 6 seconds. The drastically improved debugging experience justifies these few extra seconds spent on testing. But it total, testing is still faster, because compilation of
the test executable takes about 12 seconds, and now it doesn't need to be rebuilt at test time.

This was motivated by the effort to refactor the SOFIE-emitted code to make it differentiable with Clad.

### DIFF
--- a/tmva/sofie/test/CMakeLists.txt
+++ b/tmva/sofie/test/CMakeLists.txt
@@ -47,22 +47,15 @@ ROOTTEST_ADD_TEST(SofieCompileModels_ONNX
 
 # Creating a Google Test
 if (BLAS_FOUND)  # we need BLAS for compiling the models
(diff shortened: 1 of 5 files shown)


guitargeek:
Thanks for the review! Good point with the RSofieReader :+1: That would be a good followup, but from a practical point of view I'm happy with how it is right now for the development that I want to do.

PR #21185, opened 2026-02-07
[eve7] Correct main color change update in REveDigitSet

alja:
This is correction for the issue reported in https://github.com/alja/FireworksWeb/issues/135.


alja:
@linev Please check.

PR #21191, opened 2026-02-08
[df] Deprecate confusing signatures of HistoN[Sparse]D

### LINKED ISSUE #20816: Improve `RInterface::HistoND` signature with an extra argument for weight

According to the `HistoND` documentation, we are currently accepting a list of columns where the size matters:

https://github.com/root-project/root/blob/f91f2775cb776ce7cecab207ce4e26feadfc0860/tree/dataframe/inc/ROOT/RDF/RInterface.hxx#L2222

This should instead be aligned with other histogram actions signature, where the weight is a separate argument, e.g. for Histo3D

https://github.com/root-project/root/blob/f91f2775cb776ce7cecab207ce4e26feadfc0860/tree/dataframe/inc/ROOT/RDF/RInterface.hxx#L2189-L2190

Extracting this comment into a separate issue to keep better track of it https://github.com/root-project/root/pull/20664#discussion_r2613262855

### ROOT version

Any

### Installation method

Any

### Operating system

Any

### Additional context

_No response_

vepadulano:
The signatures of HistoN[Sparse]D operations allow passing the column that will provide the weights as the last element in the list of input columns, practically allowing for a size which is one more than the number of dimensions of the histogram to be filled. This was done to align these signatures with the logic of THnBase::Fill that follows the same schema.

But these signatures are also different than all other Histo* signatures that accept weight columns as separate arguments. Thus, they are aligned to the other signatures by adding a new optional argument for the name of the column that will provide the weights.

The previous usage of the operations is deprecated with a warning.

Add tests.

Fixes #20816

### DIFF
--- a/README/ReleaseNotes/v640/index.md
+++ b/README/ReleaseNotes/v640/index.md
@@ -192,6 +192,7 @@ As part of this migration, the following build options are deprecated. From ROOT
 ## RDataFrame
 
 - The message shown in ROOT 6.38 to inform users about change of default compression setting used by Snapshot (was 101 before 6.38, became 505 in 6.38) is now removed.
+- Signatures of the HistoND and HistoNSparseD operations have been changed. Previously, the list of input column names was allowed to contain an extra column for events weights. This was done to align the logic with the THnBase::Fill method. But this signature was inconsistent with all other Histo* operations, which have a separate function argument that represents the column to get the weights from. Thus, HistoND and HistoNSparseD both now have a separate function argument for the weights. The previous signature is still supported, but deprecated: a warning will be raised if the user passes the column name of the weights as an extra element of the list of input column names. In a future version of ROOT this functionality will be removed. From now on, creating a (sparse) N-dim histogram with weights should be done by calling `HistoN[Sparse]D(histoModel, inputColumns, weightColumn)`.
 
 ## Python Interface
 
--- a/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
+++ b/tree/dataframe/inc/ROOT/RDF/RInterface.hxx
@@ -2216,7 +2216,7 @@ public:
    /// \param[in] model The returned histogram will be constructed using this as a model.
    /// \param[in] columnList
    /// A list containing the names of the columns that will be passed when calling `Fill`.
-   ///  (N columns for unweighted filling, or N+1 columns for weighted filling)
+   /// \param[in] wName The name of the column that will provide the weights.
    /// \return the N-dimensional histogram wrapped in a RResultPtr.
    ///
    /// This action is *lazy*: upon invocation of this method the calculation is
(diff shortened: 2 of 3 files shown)


tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -2216,7 +2216,7 @@ public:
    /// \param[in] model The returned histogram will be constructed using this as a model.
    /// \param[in] columnList
    /// A list containing the names of the columns that will be passed when calling `Fill`.
-   ///  (N columns for unweighted filling, or N+1 columns for weighted filling)
+   /// \param[in] wName The name of the column that will provide the weights.
```
ferdymercury, line 2219:
```suggestion
   /// \param[in] wName The name of the column that will provide the weights.
   /// \deprecated Passing the column with the weights as the last column in the list is deprecated.
   /// Instead, pass it as a separate argument, e.g. `HistoND(model, cols, weightCol)`.
```

same suggestions below

dpiparo:
A candidate for `/backport to 6.38`?

vepadulano:
> A candidate for /backport to 6.38?

I'm not opposed to this per se. But I wanted to highlight that this would set a precedent that we need to be ready to confirm also for other similar situations in the future.

hahnjo:
My two cents: If we are deprecating method overloads that have been there for a long time, this should be part of ROOT 6.40. My previous push to have this in ROOT 6.38.04 was based on the understanding that we want to remove an interface that was only introduced recently, which would be done better sooner than later.

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
ferdymercury: Also: potentially add mention in ReleaseNotes

vepadulano:
> If we are deprecating method overloads that have been there for a long time, this should be part of ROOT 6.40.

I agree. Since the two methods are similar, I would propose to have both HistoND and HistoNSparseD deprecated in 6.40, would you agree?

dpiparo:
Yes. When proposing the backport, I was also thinking it was what was discussed. As far as I am concerned, this PR can be merged to `master` at your earliest convenience.

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
vepadulano: Done in latest commit 👍

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -2216,7 +2216,9 @@ public:
...
+   /// \deprecated Passing the column with the weights as the last column in the list is deprecated.
```
hageboeck: The placement of this `\deprecated` looks like `wName` is deprecated, but I think that's misleading.

hageboeck: The \deprecated tag might actually be interpreted as belonging to the entire function.

Instead, I would put a `\note` explaining the change in behaviour below the list of arguments, e.g.:
```c++
/// \note Starting in ROOT v6.40, a column with event weights should not be passed as part of `columnList`, but instead be passed in the new argument `wName`:
/// `HistoND(model, cols, weightCol)`.
```

tree/dataframe/inc/ROOT/RDF/RInterface.hxx
```diff
@@ -2230,18 +2232,38 @@ public:
...
+            Warning("HistoND", "Passing the column with the weights as the last column in the list is deprecated. "
+                               "Instead, pass it as a separate argument, e.g. 'HistoND(model, cols, weightCol)'.");
+
+         if (!wName.empty() || nCols == hDims + 1)
             h->Sumw2();
-         } else if (int(columnList.size()) != h->GetNdimensions()) {
-            throw std::runtime_error("Wrong number of columns for the specified number of histogram axes.");
-         }
+
+         if (!(nCols == hDims + 1) && nCols != hDims)
```
hageboeck:
```suggestion
         if (nCols != hDims + 1 && nCols != hDims)
```

README/ReleaseNotes/v640/index.md
```diff
@@ -192,6 +192,7 @@ As part of this migration, the following build options are deprecated. From ROOT
 ## RDataFrame
 
 - The message shown in ROOT 6.38 to inform users about change of default compression setting used by Snapshot (was 101 before 6.38, became 505 in 6.38) is now removed.
+- Signatures of the HistoND and HistoNSparseD operations have been changed. Previously, the list of input column names for the values to fill the bins width could also have an extra column name which represented the column to read values to be used as weights from. This was done to align the logic with the THnBase::Fill method. But this signature was inconsistent with all other Histo* operations, which have a separate function parameter that represents the column(s) to get the weights from. Thus, HistoND and HistoSparseND both now have a separate function argument for the weights. The previous signature is still supported, but deprecated: a warning will be raised if the user passes the column name of the weights as an extra element of the list of input column names. In a future version of ROOT this functionality will be removed. From now on, creating a (sparse) N-dim histogram with weights should be done by calling HistoN[Sparse]D(histoModel, inputColumns, weightColumn).
```
hageboeck:
```suggestion
- Signatures of the HistoND and HistoNSparseD operations have been changed. Previously, the list of input column names was allowed to contain an extra column for events weights. This was done to align the logic with the THnBase::Fill method. But this signature was inconsistent with all other Histo* operations, which have a separate function argument that represents the column(s) to get the weights from. Thus, HistoND and HistoSparseND both now have a separate function argument for the weights. The previous signature is still supported, but deprecated: a warning will be raised if the user passes the column name of the weights as an extra element of the list of input column names. In a future version of ROOT this functionality will be removed. From now on, creating a (sparse) N-dim histogram with weights should be done by calling `HistoN[Sparse]D(histoModel, inputColumns, weightColumn)`.
```

hahnjo (commented):
Looks good from my point of view.

PR #20498, opened 2025-11-24
[io] further conversions from Int_t to Long64_t in interfaces

ferdymercury:
@pcanal Exploring followup of https://github.com/root-project/root/pull/19606

with a pragma I tried to avoid some warnings when declaring the final-deprecated version.

But still some there.

So not sure what else to try.

PR #21192, opened 2026-02-08
io: Correct handling of byte count pos in TBufferFile::ReadObjectAny

pcanal:
This is a missing commit for https://github.com/root-project/root/pull/21183 which was merge a bit too soon.


pcanal:
Note that https://github.com/root-project/root/pull/21183/changes/588b29e6f2ed80bd37eaafd9ea2e14336420b1b2 was also not reviewed.

ferdymercury:
Closing https://github.com/root-project/root/pull/20498 since it seems this is an alternative / better approach

io/io/src/TBufferFile.cxx
```diff
@@ -2593,7 +2593,8 @@ void *TBufferFile::ReadObjectAny(const TClass *clCast)
    InitMap();
 
    // before reading object save start position
-   UInt_t startpos = UInt_t(fBufCur-fBuffer);
+   ULong64_t startpos = static_cast<ULong64_t>(fBufCur-fBuffer);
+   ULong64_t cntpos = startpos <= kMaxCountPosition ? startpos : kOverflowPosition;
```
jblomer, line 2659: Don't we need to push the startpos on the stack if it's large?

jblomer (approved):

io/io/src/TBufferFile.cxx
pcanal: yes we do ... but not always :( .. the tag can be either a reference or a byte count, we don't know yet which one it is going to be ... Eitherway the push is indeed missing in some cases (and/or always pushd and explicit pop in the other cases)

pcanal: Actual we already did actually push the byte count position ... indirectly in `ReadClass`.   I added the relevant code comment.

pcanal:
@jblomer This is ready for a final review.  Note that the last commit (733ac652ab96d7caa9c60719fe61cbeac4abd3d6) is not really part of this PR but is necessary to make the test work as is (That last commit will be replaced on the master by a more general solution).

core/base/inc/TBuffer.h
```diff
@@ -69,6 +69,8 @@ class TBuffer : public TObject {
    Int_t Write(const char *name, Int_t opt, Long64_t bufsize) const override
                               { return TObject::Write(name, opt, bufsize); }
 
+   virtual UInt_t ReserveByteCount(const TClass *) = 0;
```
jblomer, line 79: Add a comment that explains what is returned (direct byte count pos or special value).

core/base/inc/TBuffer.h
```diff
@@ -125,6 +127,22 @@ class TBuffer : public TObject {
    virtual Long64_t   CheckByteCount(ULong64_t startpos, ULong64_t bcnt, const TClass *clss) = 0;
    virtual Long64_t   CheckByteCount(ULong64_t startpos, ULong64_t bcnt, const char *classname) = 0;
    virtual void       SetByteCount(ULong64_t cntpos, Bool_t packInVersion = kFALSE)= 0;
+   class ByteCountWriter {
```
jblomer, line 167: I think we should delete the copy constructor/assignment operator of this class.

jblomer, line 167: Add documentation

core/base/inc/TBuffer.h
```diff
@@ -135,6 +153,7 @@ class TBuffer : public TObject {
 
    virtual void      *ReadObjectAny(const TClass* cast) = 0;
    virtual void       SkipObjectAny() = 0;
+   virtual void       SkipObjectAny(Long64_t start, UInt_t bytecount) = 0;
```
jblomer, line 215: It would be useful to add additional documentation that explains why this overload is needed. We seem to need this only once in the code base. Is it worth (the only option) to add a new virtual method in a fundamental class?

io/io/src/TBufferFile.cxx
```diff
@@ -2569,7 +2612,26 @@ void TBufferFile::SkipObjectAny()
...
+   // actually a byte count.
+   if (count)
+      fByteCountStack.pop_back();
+}
+
+////////////////////////////////////////////////////////////////////////////////
+/// Skip any kind of object from buffer
+
+void TBufferFile::SkipObjectAny(Long64_t start, UInt_t count)
+{
```
jblomer, line 2629: assert count > 0?

io/io/test/CMakeLists.txt
```diff
@@ -41,6 +41,21 @@ endif()
...
+
+if(NOT _32BIT)
+  ROOTTEST_COMPILE_MACRO(testByteCount.cxx
+                        FIXTURES_SETUP io-io-tobj-fixture)
+
+  ROOTTEST_ADD_TEST(testLargeByteCounts
+                    MACRO testByteCount.cxx+
+                    OUTREF testByteCount.ref
+                    FIXTURES_REQUIRED io-io-tobj-fixture)
+endif()
```
jblomer: Missing trailing newline

cmake/modules/RootCTest.cmake
```diff
@@ -51,13 +51,14 @@ foreach(d ${test_list})
...
-# If multiple of these are invoked in parallel, ninja will suffer from race conditions.
+# When ninja or the Microsoft generator are in use, tests that compile an executable might try
+# to rebuild the entire build tree. If multiple of these are invoked in parallel, ninja will
+# suffer from race conditions.
 # To solve this, do the following:
 # - Add a test that updates the build tree (equivalent to "ninja all"). This one will run in complete isolation.
 # - Make all tests that require a ninja build depend on the above test.
 # - Use a RESOURCE_LOCK on all tests that invoke ninja, so no two tests will invoke ninja in parallel
-if(CMAKE_GENERATOR MATCHES Ninja)
+if(CMAKE_GENERATOR MATCHES Ninja OR CMAKE_GENERATOR MATCHES "Visual Studio")
```
bellenot: Does that mean ROOT is build twice? Once with MSVC and once with Ninja? At least it looks like it's what is happening in this PR...

pcanal:
> Closing https://github.com/root-project/root/pull/20498 since it seems this is an alternative / better approach

No per se :).   I think that part of #20498 is already merged into either master or the Arlesienne branch and what is left may or may not be needed.

cmake/modules/RootCTest.cmake
pcanal: > Does that mean ROOT is build twice?

Indeed.  With Ninja and the Visual Studio generator when `ctest` indirectly ends calling `cmake -build` because of some silly timestamp shenanigans, it ends up needing to rebuild large fraction of ROOT.

pcanal: I had missed 9 more places to update :( ... see https://github.com/root-project/root/pull/21192/changes/fbaed6e1da5eecb68bbf18e9eab7f6a44dc9e826

Hopefully, this goes better!

io/io/src/TBufferFile.cxx
jblomer, line 2765: Aren't we missing a `CheckByteCount()` here?

io/io/src/TBufferFile.cxx
```diff
@@ -2824,7 +2886,8 @@ TClass *TBufferFile::ReadClass(const TClass *clReq, UInt_t *objTag)
       // When objTag is not used, the caller is not interested in the byte
       // count and will not (can not) call CheckByteCount.
       if (objTag)
-         fByteCountStack.push_back(fBufCur - fBuffer);
+         // Note: the actual class is set later on.
+         fByteCountStack.push_back({fBufCur - fBuffer - sizeof(UInt_t), clReq, nullptr});
```
jblomer: Maybe enclose if block in curly braces.

io/io/inc/TBufferFile.h
```diff
@@ -53,7 +53,12 @@ class TBufferFile : public TBufferIO {
    InfoList_t      fInfoStack;     ///< Stack of pointers to the TStreamerInfos
 
    using ByteCountLocator_t = std::size_t; // This might become a pair<chunk_number, local_offset> if we implement chunked keys
-   using ByteCountStack_t = std::vector<ByteCountLocator_t>;
+   struct ByteCountLocationInfo {
+      ByteCountLocator_t locator; ///< Position where the byte count value is stored
+      const TClass* cl;         ///< Class for which the byte count is reserved
+      const TClass* alt;
```
jblomer, line 64: Add member comment

cmake/modules/RootCTest.cmake
pcanal: But now I see what you mean, it is actually rebuilding *everything * rather than only a little bit :(

pcanal: The `ninja-build-all` (now renamed here to `cmake-build-all`) was missing `${build_config}` (see https://github.com/root-project/root/pull/21192/changes/677a60982c65d95992a942e913b39cfd46e18d3d)

pcanal: Let's continue the discussion at the dedicated PR: https://github.com/root-project/root/pull/21358

io/io/src/TBufferFile.cxx
pcanal: Wouldn't that always be true?  (`count` is an `unsigned int` here)
Or Do you mean to require strictly positive?

core/base/inc/TBuffer.h
pcanal: This is used for error handling within a Streamer.

The use case is a streamer with a flow like:
- Read version and bytecount
- Start reading the date
- Detect an error (in the use case, missing some information for a `CollectionProxy`)
- Properly set the cursor to the end of the object to allow to continue without problem.

Because the actual `ByteCount` information for large byte count is only in the stack, I could only see two ways to support this use case: add the new `SkipObjectAny` overload or make the stack `public` (via a Getter).

io/io/src/TBufferFile.cxx
pcanal: No.  In that code path we handle the reading of an object written multiple times in the stream (once for 'real' and many times just as a reference), the `tag` is the location the object was first written to.

core/base/inc/TBuffer.h
jblomer: Thank you for the explanation! Can you add this as a method documentation.

io/io/src/TBufferFile.cxx
jblomer: Yes, strictly positive.

pcanal: I don't think that is needed/wanted.  Apriori the object could indeed be 'empty' (all data transient for example).

PR #21193, opened 2026-02-08
[io] Add support for long range references in TBufferFile.cxx

pcanal:
References for class names and objects are stored in 64bits whenever they are
    located past 1GB.  This is the 'only' option as there are no control bits left
    to distinguish between a 32bits and a 64bits reference.

    Reference in the longRange section (>1GB) are now tagged so that we can differentiate nullptr
    (stored as only 32 bits) from short reference stored in the longRange section (i.e. those are stored
    in 64 bits but have only zeros in the high bit).


pcanal:
Wrong target branch.

PR #21194, opened 2026-02-08
io: Add support for long range references in TBufferFile

pcanal:
References for class names and objects are stored in 64bits whenever they are
    located past 1GB.  This is the 'only' option as there are no control bits left
    to distinguish between a 32bits and a 64bits reference.

    Reference in the longRange section (>1GB) are now tagged so that we can differentiate nullptr
    (stored as only 32 bits) from short reference stored in the longRange section (i.e. those are stored
    in 64 bits but have only zeros in the high bit).

### DIFF
--- a/io/io/src/TBufferFile.cxx
+++ b/io/io/src/TBufferFile.cxx
@@ -2820,7 +2820,6 @@ void TBufferFile::WriteObjectClass(const void *actualObjectStart, const TClass *
          const bool shortRange = (fBufCur - fBuffer) <= kMaxMapCount;
 
          // save index of already stored object
-         // FIXME/TRUNCATION: potential truncation from 64 to 32 bits
          if (R__likely(shortRange)) {
             // truncation is OK the value we did put in the map is an 30-bit offset
             // and not a pointer
@@ -2832,7 +2831,6 @@ void TBufferFile::WriteObjectClass(const void *actualObjectStart, const TClass *
             // This is needed so that the reader can distinguish between references,
             // bytecounts, and new class definitions.
             ULong64_t objIdx = static_cast<ULong64_t>(idx);
-            // FIXME: verify that objIdx is guaranteed to fit in 60-bits, i.e. objIdx <= kMaxLongRange
             assert(objIdx <= kMaxLongRange);
             *this << (objIdx | kLongRangeRefMask);
          }
@@ -2900,7 +2898,6 @@ TClass *TBufferFile::ReadClass(const TClass *clReq, ULong64_t *objTag)
    const bool shortRange = (fBufCur - fBuffer) <= kMaxMapCount;
    bool isNewClassTag = false;
 
-   // FIXME/TRUNCATION: potential truncation from 64 to 32 bits
    *this >> bcnt;
    if (bcnt == kNewClassTag) {
       isNewClassTag = true;
@@ -2959,7 +2956,6 @@ TClass *TBufferFile::ReadClass(const TClass *clReq, ULong64_t *objTag)
    // in case tag is object tag return tag
    // NOTE: if we return early for reference for longRange, this would be only for shortRange
    if (!isClassTag && !isNewClassTag) {
-      // FIXME/TRUNCATION: potential truncation from 64 to 32 bits
       if (objTag)
          *objTag = tag64;
       return 0;
@@ -3045,9 +3041,6 @@ void TBufferFile::WriteClass(const TClass *cl)
          UInt_t clIdx = UInt_t(idx);
 
          // save index of already stored class
-         // FIXME/TRUNCATION: potential truncation from 64 to 32 bits
-         // FIXME/INCORRECTNESS: if clIdx > 0x3FFFFFFF the control bit (2nd highest bit) will be wrong
-         // FIXME/INCORRECTNESS: similarly if clIdx > kClassMask (2GB) the code will be wrong
          *this << (clIdx | kClassMask);
       } else {
          // The 64-bit value is stored highest bytes first in the buffer,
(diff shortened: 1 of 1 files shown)


core/base/inc/TBuffer.h
```diff
@@ -150,7 +150,7 @@ class TBuffer : public TObject {
    virtual TVirtualArray *PopDataCache();
    virtual void           PushDataCache(TVirtualArray *);
 
-   virtual TClass    *ReadClass(const TClass *cl = nullptr, UInt_t *objTag = nullptr) = 0;
+   virtual TClass    *ReadClass(const TClass *cl = nullptr, ULong64_t *objTag = nullptr) = 0;
```
jblomer, line 231: We could keep the old signature, forward to the new one and throw if the returned `objTag` is too large.

core/base/src/TString.cxx
```diff
@@ -1375,7 +1375,7 @@ TString *TString::ReadString(TBuffer &b, const TClass *clReq)
    // Before reading object save start position
    UInt_t startpos = UInt_t(b.Length());
```
jblomer: Not subject to this PR but should also be 64bit. Same issue for the write counterpart.

core/cont/src/TArray.cxx
```diff
@@ -47,7 +47,7 @@ TArray *TArray::ReadArray(TBuffer &b, const TClass *clReq)
    // Before reading object save start position
    UInt_t startpos = UInt_t(b.Length());
```
jblomer: Should be 64bit. Same issue for the write counterpart.

io/io/src/TContainerConverters.cxx
```diff
@@ -105,7 +105,7 @@ void TConvertClonesArrayToProxy::operator()(TBuffer &b, void *pmember, Int_t siz
          UInt_t startpos = b.Length();
```
jblomer: 64bit? To be checked for writing.

io/io/src/TBufferFile.cxx
```diff
@@ -74,6 +74,10 @@ constexpr Version_t kByteCountVMask  = 0x4000;       // OR the version byte coun
 constexpr Version_t kMaxVersion      = 0x3FFF;       // highest possible version number
 constexpr Int_t  kMapOffset          = 2;   // first 2 map entries are taken by null obj and self obj
 
+constexpr ULong64_t kMaxLongRange = 0x0FFFFFFFFFFFFFFE; // We reserve the 4 highest bits for flags, currently only 2 are in use.
+constexpr ULong64_t kLongRangeClassMask = 0x8000000000000000; // OR the class index with this
+// constexpr ULong64_t kLongRangeByteCountMask = 0x4000000000000000; // OR the byte count with this
```
jblomer, line 82: Would be useful to add a comment why this is commented. Because it conflicts with the regular `kByteCountMask`?

io/io/src/TBufferFile.cxx
```diff
@@ -2746,13 +2751,25 @@ void TBufferFile::WriteObjectClass(const void *actualObjectStart, const TClass *
...
+            // and not a pointer
+            UInt_t objIdx = UInt_t(idx);
+            *this << objIdx;
+         } else {
+            // The 64-bit value is stored highest bytes first in the buffer,
+            // so when reading just the first 32-bits we get the control bits in place.
+            // This is needed so that the reader can distinguish between references,
+            // bytecounts, and new class definitions.
+            ULong64_t objIdx = static_cast<ULong64_t>(idx);
+            // FIXME: verify that objIdx is guaranteed to fit in 60-bits, i.e. objIdx <= kMaxLongRange
```
jblomer: This comment can be removed, I think

io/io/src/TBufferFile.cxx
```diff
@@ -2901,19 +2967,31 @@ void TBufferFile::WriteClass(const TClass *cl)
...
+         // FIXME/INCORRECTNESS: if clIdx > 0x3FFFFFFF the control bit (2nd highest bit) will be wrong
+         // FIXME/INCORRECTNESS: similarly if clIdx > kClassMask (2GB) the code will be wrong
+         *this << (clIdx | kClassMask);
+      } else {
+         // The 64-bit value is stored highest bytes first in the buffer,
+         // so when reading just the first 32-bits we get the control bits in place.
+         // This is needed so that the reader can distinguish between references,
+         // bytecounts, and new class definitions.
+         ULong64_t clIdx = static_cast<ULong64_t>(idx);
+         // FIXME: verify that clIdx is guaranteed to fit in 60-bits, i.e. clIdx <= kMaxLongRange
```
jblomer: Can be removed?

core/base/src/TString.cxx
pcanal: Good catch.  The writer part needs to use `TBuffer::ByteCountWriter`.

Both `TString::ReadString` and `TArray::ReadArray` need an update.

io/io/src/TContainerConverters.cxx
pcanal: Since this is for conversion, this is only for reading.

io/io/src/TBufferFile.cxx
pcanal: I added:
```
// kLongRangeByteCountMask is currently unused as the byte count are always 32 bits long
// and thus we use kByteCountMask for them, but we keep it here for future use if we want
// to support 64 bits byte counts in the long range format
```

core/base/inc/TBuffer.h
pcanal: I am fairly hesitant but add it in its own commit.

jblomer (approved):
LGTM

PR #21195, opened 2026-02-08
[CMake] Use OUTPUT_NAME if defined in ROOT_GENERATE_DICTIONARY

aphecetche:
# This Pull request:

Fixes the logic used to infer the base name used to generate the `libXXX.dylib|so`, `libXXX_rdict.pcm` and `libXXX.rootmap` when using the `MODULE` option of the `ROOT_GENERATE_DICTIONARY` CMake macro, in the case where the `OUTPUT_NAME` of XXX has been set prior to the call to ROOT_GENERATE_DICTIONARY. 

## Checklist:

- [X ] tested changes locally
- [ ] updated the docs (if necessary) => not necessary (?)

### DIFF
--- a/cmake/modules/RootMacros.cmake
+++ b/cmake/modules/RootMacros.cmake
@@ -528,7 +528,13 @@ function(ROOT_GENERATE_DICTIONARY dictionary)
     if (NOT ${ARG_MODULE} STREQUAL ${library_target_name})
 #      message(AUTHOR_WARNING "The MODULE argument ${ARG_MODULE} and the deduced library name "
 #        "${library_target_name} mismatch. Deduction stem: ${dictionary}.")
-      set(library_target_name ${ARG_MODULE})
+      get_target_property(custom_name ${ARG_MODULE} OUTPUT_NAME)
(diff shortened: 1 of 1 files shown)


guitargeek:
Hi, thanks for the PR! Let's see if it works.

However, I think this goes in the opposite direction of https://github.com/root-project/root/commit/498165dc9c88d8e84c3d753beb0af45bb7bcc174, which was fixing https://github.com/root-project/root/issues/15792.

Do you maybe have a simple CMake reproducer for the problem that you try to fix?

aphecetche:
@guitargeek actually it's trying to fix the same bug as https://github.com/root-project/root/issues/15792. I did not do my homework correctly and a) did not check past issues thoroughly b) forgot I'm using Root 6.32 daily and not the latest master (so I did try my fix in my current setup and "pasted-in" in master to prepare the PR w/o too much thinking).  Sorry about that.

guitargeek:
No problem! Why would you assume that a niche problem like this is fixed in `master` and check in the closed issues before starting to code. I also wouldn't have done that :slightly_smiling_face:

PR #21196, opened 2026-02-08
[Python] Assert ABI-compatible Python version on `import ROOT`

guitargeek:
We should give the user a helpful error message if the ROOT and Python interpreter versions don't match. Otherwise, one will get confusing `undefined symbol` error or other unclear behavior at runtime.

This came up sometimes in forum posts or GitHub issues, so the clear error would have helped.


dpiparo (approved):
thanks for this useful improvement.

vepadulano (approved):
For the record, the proper/more established approach would be to develop CPython extensions that only use Python limited API and thus can be built to be ABI compatible across all Python versions supported. After discussion with @guitargeek , it's clear that this is currently not possible for cppyy, so this approach is the next best thing 👍

PR #21197, opened 2026-02-09
[RF][HS3] bugfix for external constraint handling

cburgard:
# This Pull request:

Makes sure the external constraints exported are re-imported successfully.

## Checklist:

- [x] tested changes locally

This PR fixes #


guitargeek:
Hi @cburgard! Thanks for the PR. Do you maybe have a small example with a problematic workspace that can be turned into a unit test? I think adding more tests would help a lot to prevent future regressions!

cburgard:
Hi @guitargeek ,
thanks for the quick reply. I'm currently working on a paper that uses this. I'll happily provide the workspace for the paper once it's out, but that will need a few more weeks.

PR #21199, opened 2026-02-09
[test] Test that import ROOT does not implicitly import cppyy

vepadulano:
Adds a test after the recent improvements of the ROOT Python package.

### DIFF
--- a/bindings/pyroot/pythonizations/test/root_no_implicit_cppyy.py
+++ b/bindings/pyroot/pythonizations/test/root_no_implicit_cppyy.py
@@ -1,0 +1,26 @@
+import unittest
+
+
+class ROOTNoImplicitCppyy(unittest.TestCase):
+    """
+    Test that import ROOT does not implicitly import cppyy.
+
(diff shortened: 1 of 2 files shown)


bindings/pyroot/pythonizations/test/root_no_implicit_cppyy.py
```diff
@@ -0,0 +1,21 @@
...
+    parts of ROOT which involve initialization of the C++ runtime.
+    """
+
+    def test_no_implicit_cppyy_import(self):
+        import sys
+
+        import ROOT  # noqa: F401 We just need to import the package without using it
+
+        self.assertFalse("cppyy" in sys.modules)
+
```
guitargeek, line 18:
```suggestion

ROOT.gTInterpreter

self.assertTrue("cppyy" in sys.modules)
```

Just as a sanity check (imagine we rename `cppyy` at some point, then the existing test would be mute).

vepadulano: Good point, thanks!

PR #21204, opened 2026-02-09
[net] Make TSocket::Recv slightly more readable

silverweed:
- set out pointer to `nullptr` instead of `0`
- use `while` instead of `goto`
- add a couple of explaining comments


pcanal (approved):
Thanks.

PR #21206, opened 2026-02-09
[test][cppyy] Add test for string_view to Python str

aaronj0:
Adds a cppyy test for https://github.com/root-project/root/pull/19620. This simple test ensures we don't regress and see the old behaviour of string_view: `str(s) == '"Hello, World!"[13]'`


guitargeek (approved):
Thank you very much!

PR #21207, opened 2026-02-09
[net] testParallelMergingFile: enable thread safety

silverweed:
Tentative fix for the sporadic crash we get in this test. On my machine it seems to solve the issue.


net/net/test/testParallelMergingFile.cxx
```diff
@@ -90,6 +90,8 @@ static void Server(std::unique_ptr<TServerSocket> ss, const std::string &outFile
 
       delete msg;
    }
+
+   std::scoped_lock<std::mutex> lock(gMutex);
```
pcanal: If this solves the problem, we ought to understand better the underlying cause.

PR #21210, opened 2026-02-09
[hist] Add tutorials for concurrent filling and conversion to `TH1`

hahnjo:


hahnjo:
gentle ping :smiley:

siliataider (approved):
They look good thanks!

pcanal:
> Just a reminder, we need to make a decision if it is acceptable to create a TH* histogram without a name.

It would be sub-optimal to have an empty name both for code (see next point) and for clarity when the histogram is stored in a file or shown in the browser (and when drawn).  For reminder the lists this histogram may/will end up in are using hash list based on the name.

hahnjo:
> Since we now have the conversion to TH1, should we use it to draw the histogram in the concurrent filler tutorial?

Yes, that's a point of discussion. I'm not sure how much we want to "advertise" the fallback path to `TH*` in all tutorials. For the moment, it's separated between "needs only `histv7`" and "relies also on `histv7util`"...

> Just a reminder, we need to make a decision if it is acceptable to create a TH* histogram without a name.

Yes, I took note.

> It would be sub-optimal to have an empty name both for code (see next point) and for clarity when the histogram is stored in a file or shown in the browser (and when drawn). For reminder the lists this histogram may/will end up in are using hash list based on the name.

At the moment, we use the default `TH1()` constructor which doesn't add the histogram anywhere. From that point of view, a name is completely optional and I'm not sure I see the advantage of forcing the user to choose a name...

PR #21211, opened 2026-02-09
Support Home/End keys in ROOT prompt on rxvt-unicode

guitargeek:
Closes JIRA ticket https://its.cern.ch/jira/browse/ROOT-10988

Thanks to Joao Pedro Athayde Marcondes De Andre for the patch, which was tested on rxvt-unicode (aka. urxvt) version 9.31.


pcanal (approved):
Thanks.

PR #21212, opened 2026-02-09
[RF] No verbose teststat printouts

cburgard:
# This Pull request:

remove some very verbose printouts from the ProfileLikelihoodTestStat

I guess the printouts could also be changed to use one of the RooFit printout mechanisms, but as it stands, they cannot be silenced, which is an issue since the test can be invoked from the ToyMCSampler inside the toy loop, which can easily generate O(10000) lines of output.

## Checklist:

- [x] tested changes locally

This PR fixes #

### DIFF
--- a/roofit/roostats/src/ProfileLikelihoodTestStat.cxx
+++ b/roofit/roostats/src/ProfileLikelihoodTestStat.cxx
@@ -303,25 +303,26 @@ std::unique_ptr<RooFitResult> RooStats::ProfileLikelihoodTestStat::GetMinNLL() {
    TString algorithm = ROOT::Math::MinimizerOptions::DefaultMinimizerAlgo();
    if (algorithm == "Migrad") algorithm = "Minimize"; // prefer to use Minimize instead of Migrad
    int status;
+   auto &msg = RooMsgService::instance();
    for (int tries = 1, maxtries = 4; tries <= maxtries; ++tries) {
       status = minim.minimize(minimizer,algorithm);
       if (status%1000 == 0) {  // ignore errors from Improve
          break;
       } else if (tries < maxtries) {
-	//         std::cout << "    ----> Doing a re-scan first" << std::endl;
+	msg.log(nullptr, RooFit::DEBUG, RooFit::Minimization) << " ----> Doing a re-scan first\n";
          minim.minimize(minimizer,"Scan");
          if (tries == 2) {
             if (fStrategy == 0 ) {
-	      //               std::cout << "    ----> trying with strategy = 1" << std::endl;
-               minim.setStrategy(1);
+	      msg.log(nullptr, RooFit::DEBUG, RooFit::Minimization) << "    ----> trying with strategy = 1\n";
+	      minim.setStrategy(1);
             }
             else
-               tries++; // skip this trial if strategy is already 1
+	      tries++; // skip this trial if strategy is already 1
          }
          if (tries == 3) {
-	   //            std::cout << "    ----> trying with improve" << std::endl;
-            minimizer = "Minuit";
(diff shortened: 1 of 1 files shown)


roofit/roostats/src/ProfileLikelihoodTestStat.cxx
```diff
@@ -308,18 +308,18 @@ std::unique_ptr<RooFitResult> RooStats::ProfileLikelihoodTestStat::GetMinNLL() {
       if (status%1000 == 0) {  // ignore errors from Improve
          break;
       } else if (tries < maxtries) {
-         std::cout << "    ----> Doing a re-scan first" << std::endl;
+	//         std::cout << "    ----> Doing a re-scan first" << std::endl;
```
silverweed: @guitargeek can we use `R__LOG_DEBUG` to silence these by default but still have them available if one sets the debug level appropriately?

roofit/hs3/src/RooJSONFactoryWSTool.cxx
```diff
@@ -581,9 +581,10 @@ void importAnalysis(const JSONNode &rootnode, const JSONNode &analysisNode, cons
    if (!nllNode->has_child("data")) {
       throw std::runtime_error("likelihood node has no data attached!");
    }
+
    std::vector<std::string> nllDistNames = valsToStringVec((*nllNode)["distributions"]);
```
silverweed, line 584: Should this also be:
```suggestion
   std::vector<std::string> nllDistNames = valsToStringVec(analysisNode["distributions"]);
```
?

roofit/roostats/src/ProfileLikelihoodTestStat.cxx
cburgard: I've just committed a change to this effect, I hope I used it correctly.

silverweed: That's not how `R__LOG_DEBUG` is used: see `core/foundation/inc/ROOT/RLogger.hxx` for the definition and e.g. `core/foundation/test/testLogger.cxx` for the usage.
However I don't see it used anywhere in roofit so maybe rootfit uses something different, that's why I asked @guitargeek.

Also, as a general rule please always test that your code at the very least compiles before pushing, otherwise the CI can waste a lot of CPU time for something that's easily caught locally...

cburgard: Ah ok, sorry! I'm currently working on several PRs at the same time locally and didn't find it very easy to go back to this one to test (I did the edit via the webinterface), but you are correct, of course!
I'll roll back my change.

roofit/hs3/src/RooJSONFactoryWSTool.cxx
cburgard: No, this is correct as-is. Distributions are stored as part of the NLL, wheras the external constraints are stored as part of the analysis node (=ModelConfig).
But thanks for checking!

roofit/roostats/src/ProfileLikelihoodTestStat.cxx
cburgard: I have now reverted the change and instead used the `RooMsgService` debug logging, which is the default for most parts of roofit AFAIK.

Please let me know if you find this more agreeable.

roofit/hs3/src/RooJSONFactoryWSTool.cxx
cburgard: This commit does not actually belong in this PR but is supposed to be merged seprately here:

https://github.com/root-project/root/pull/21197

I've removed it from the history to keep the log clean.

PR #21214, opened 2026-02-09
[ML][Python] Clean up dataloader API and avoid redundant RDF graph triggers

siliataider:
# This Pull request:

## Changes or fixes:
This PR summarizes some optimizations and cleanup, main points here:
- Removed the `RDataFrame::Count()` calls from the `RDatasetLoader` and the `RChunkLoader` to reduce the number of times we trigger the RDF computation graph.
- Removed mutable default arguments, replaced with `None` and explicit normalization.
- Updated docstrings and type hints to be consistent.

## Checklist:

- [x] tested changes locally

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_ml_dataloader.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_ml_dataloader.py
@@ -29,26 +29,30 @@ class BaseGenerator:
     def get_template(
         self,
         x_rdf: ROOT.RDF.RNode,
-        columns: list[str] = list(),
-        max_vec_sizes: dict[str, int] = dict(),
+        columns: list[str] | None = None,
+        max_vec_sizes: dict[str, int] | None = None,
     ) -> Tuple[str, list[int]]:
         """
         Generate a template for the DataLoader based on the given
         RDataFrame and columns.
 
         Args:
-            rdataframe (RNode): RDataFrame or RNode object.
-            columns (list[str]): Columns that should be loaded.
+            x_rdf (RNode): RDataFrame or RNode object.
+            columns (list[str], optional): Columns that should be loaded.
                                  Defaults to loading all columns
                                  in the given RDataFrame
-            max_vec_sizes (list[int]): The length of each vector based column.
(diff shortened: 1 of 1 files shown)


bindings/pyroot/pythonizations/python/ROOT/_pythonization/_ml_dataloader.py
```diff
@@ -83,14 +87,14 @@ def get_template(
...
         batch_size: int = 0,
         chunk_size: int = 0,
         block_size: int = 0,
-        columns: list[str] = list(),
-        max_vec_sizes: dict[str, int] = dict(),
+        columns: list[str] | None = None,
+        max_vec_sizes: dict[str, int] | None = None,
         vec_padding: int = 0,
-        target: str | list[str] = list(),
+        target: str | list[str] | None = None,
```
vepadulano, line 97: Doesn't have to be done for this PR, but something to discuss for the future is what is the meaning of having `rdataframes`, `columns`, `target` be optional arguments? That means that we are supporting calls such as `CreateNumpyGenerators()`, the meaning of which is confusing to me.

siliataider: I agree that calling `CreateNumpyGenerators()` with no arguments does not represent a meaningful use case, here is my take on the arguments you listed:
- In practice, a non-empty `rdataframe` argument is required, and the code will fail later if it's not provided.
    - I agree it should be made explicit, I suggest a follow up PR to make it mandatory.
- For `columns`, the intention is that `None` explicitly means "use all columns", which I think could be convenient if you want to use all columns without having to provide a long list 🤷🏻‍♀️ .
- For `target`, `None` or an empty value is a supported case, [it is handled via the `base_generator.target_given` flag](https://github.com/root-project/root/pull/21214/changes#diff-1169c4753afae431a51add0393c785d0bd19870beaf46e7c053f1292dce0eea2R955) and when no target is provided, rhe generator yields feature only batches (e.g use-cases: unsupervised workflows, validation, etc.).

vepadulano: Agree on all points!

PR #21225, opened 2026-02-10
Prepare roottest for ROOT 7's ownership model

hageboeck:
In essence, add `histo.SetDirectory(xxx)` to all places where a histogram should show up in a TDirectory independent of whether it has been registered automatically.

### DIFF
--- a/hist/hist/test/test_TH1.cxx
+++ b/hist/hist/test/test_TH1.cxx
@@ -5,6 +5,9 @@
 #include "TH3.h"
 #include "TH1F.h"
 #include "THLimitsFinder.h"
+#include "TDirectory.h"
+#include "TList.h"
+#include "TROOT.h"
 
 #include <cmath>
 #include <cstddef>
@@ -165,6 +168,51 @@ TEST(TH1, Normalize)
    EXPECT_FLOAT_EQ(v2.Integral(), 14.399998);
    EXPECT_FLOAT_EQ(v2.Integral("width"), 1.);
    EXPECT_FLOAT_EQ(v2.GetMaximum(), 7.9999990);
+}
+
+TEST(TH1, RegistrationToTDirectory_ImplicitOwnershipOn)
+{
+   TH1D histo1("histo1", "Test Histogram", 10, 0, 10);
+   auto histo2 = std::make_unique<TH1D>("histo2", "Test Histogram", 10, 0, 10);
+   TH1D *histo3 = new TH1D("histo3", "Test Histogram", 10, 0, 10);
+
+   {
+      TDirectory dir("dir", "Test Directory");
+      histo3->SetDirectory(&dir);
+
+      dir.cd();
+
+      TH1D histo4("histo4", "Test Histogram", 10, 0, 10);
+      auto histo5 = std::make_unique<TH1D>("histo5", "Test Histogram", 10, 0, 10);
+
+      EXPECT_EQ(dir.GetList()->GetSize(), 3);
+      EXPECT_EQ(dir.Get<TH1D>("histo1"), nullptr);
+      EXPECT_EQ(dir.Get<TH1D>("histo2"), nullptr);
+      EXPECT_EQ(dir.Get<TH1D>("histo3"), histo3);
+      EXPECT_EQ(dir.Get<TH1D>("histo4"), &histo4);
+      EXPECT_EQ(dir.Get<TH1D>("histo5"), histo5.get());
+
+      EXPECT_EQ(histo1.GetDirectory(), gROOT);
+      EXPECT_EQ(histo2->GetDirectory(), gROOT);
+      EXPECT_EQ(histo3->GetDirectory(), &dir);
+      EXPECT_EQ(histo4.GetDirectory(), &dir);
+      EXPECT_EQ(histo5->GetDirectory(), &dir);
+
+      histo5.reset();
+
+      EXPECT_EQ(dir.GetList()->GetSize(), 2);
+      EXPECT_EQ(dir.Get<TH1D>("histo1"), nullptr);
+      EXPECT_EQ(dir.Get<TH1D>("histo2"), nullptr);
+      EXPECT_EQ(dir.Get<TH1D>("histo3"), histo3);
+      EXPECT_EQ(dir.Get<TH1D>("histo4"), &histo4);
+      EXPECT_EQ(dir.Get<TH1D>("histo5"), nullptr);
+   }
+
+   EXPECT_STREQ(histo1.GetName(), "histo1");
+   EXPECT_STREQ(histo2->GetName(), "histo2");
+
+   EXPECT_EQ(histo1.GetDirectory(), gROOT);
+   EXPECT_EQ(histo2->GetDirectory(), gROOT);
 }
 
 TEST(TAxis, BinComputation_FPAccuracy)


hageboeck:
> We should enhanced this PR by adding a new test that checks the current memory ownership explicitly (Otherwise we no longer test the current/main-stream usages)

That's a very good idea, and it's now implemented in the ["master" PR](https://github.com/root-project/root/pull/18083) for this feature. I'll cherry-pick the commit that will work already in ROOT 6 mode into this PR next.

*Edit*: Done

hist/hist/test/test_TH1.cxx
```diff
@@ -167,6 +167,40 @@ TEST(TH1, Normalize)
    EXPECT_FLOAT_EQ(v2.GetMaximum(), 7.9999990);
 }
 
+TEST(TH1, RegistrationToTDirectory_ImplicitOwnershipOn)
+{
+   TH1D histo1("histo1", "Test Histogram", 10, 0, 10);
```
pcanal, line 175: Maybe (or maybe not) for the PR, we should also have similar for the other self Registering classes ....

hist/hist/test/test_TH1.cxx
```diff
@@ -167,6 +167,40 @@ TEST(TH1, Normalize)
...
+      histo3->SetDirectory(&dir);
+
+      dir.cd();
+
+      TH1D histo4("histo4", "Test Histogram", 10, 0, 10);
+      auto histo5 = std::make_unique<TH1D>("histo5", "Test Histogram", 10, 0, 10);
+
+      EXPECT_EQ(dir.Get<TH1D>("histo1"), nullptr);
+      EXPECT_EQ(dir.Get<TH1D>("histo2"), nullptr);
+      EXPECT_EQ(dir.Get<TH1D>("histo3"), histo3);
```
pcanal, line 191:
```suggestion
      EXPECT_EQ(dir.Get<TH1D>("histo3"), histo3);
      EXPECT_EQ(&dir, histo3->GetDirectory());
```

hist/hist/test/test_TH1.cxx
```diff
@@ -167,6 +167,40 @@ TEST(TH1, Normalize)
...
+      EXPECT_EQ(dir.Get<TH1D>("histo4"), &histo4);
+      EXPECT_EQ(dir.Get<TH1D>("histo5"), histo5.get());
```
pcanal, line 193:
```suggestion
      EXPECT_EQ(dir.Get<TH1D>("histo4"), &histo4);
      EXPECT_EQ(&dir, histo4->GetDirectory());
      EXPECT_EQ(dir.Get<TH1D>("histo5"), histo5.get());
      EXPECT_EQ(&dir, histo5->GetDirectory());
```

hist/hist/test/test_TH1.cxx
hageboeck: Duly noted. For TH1-derived classes, we do it in this PR. For other cases, I added a TODO to #18083

roottest/root/multicore/commands1.txt
```diff
@@ -3,5 +3,5 @@ TH1F h1("h","",100,0,1);
 std::vector<std::list<TGraph>> v;
 THtml h;
 //gInterpreter->SetClassAutoparsing(false);
-h.LoadAllLibs();
+h1.LoadAllLibs();
```
pcanal: What is the relation between this change and the intent of the PR? (i.e. what is `h1` in this context?)

roottest/root/multicore/commands2.txt
```diff
@@ -4,4 +4,4 @@ std::vector<std::list<TGraph>> v;
 std::set<TString> stringset;
 THtml h;
 //gInterpreter->SetClassAutoparsing(false);
-h.LoadAllLibs();
+h1.LoadAllLibs();
```
pcanal: See https://github.com/root-project/root/pull/21225/changes#r2829655657

roottest/root/multicore/commands1.txt
hageboeck: I'll bring in @dpiparo, the original author of this test. To give some context:
- I discovered now due to @pcanal's question that the multicore/fork test is broken since THTML got removed from ROOT.
- The breakage never got detected because of how we grep for test errors.
- And when histograms don't get registered globally, the output changed, and the breakage got detected due to the presence of `Error in`.

I think I know a way how to proceed now:
- Remove this commit from this PR and open another one doing as follows:
- Detect errors in `gInterpreter->ProcessLine`. So far, they are ignored.
- Wait for the forked child and check its exit status.
- Return non-zero exit status if any of the above failed.
- And finally: Inject interpreter commands that are supposed to succeed.

Give me a 👍 / 👎 if you approve of the above.

All of this has nothing to do with ROOT's ownership model, but I can explain now how the ownership model revealed the broken test. 🙂 


### How did we get here?
I have to admit that I didn't at all understand the intention of this test when I traced down failures due to not-registered histograms. At the time, I just looked for `THxx` and fixed problems related to that. I saw a TH1 named `h1`, and saw the interpreter trying to give a helpful suggestion related to that, so I changed the command to read `h1.` instead of `h.`. The test failure went away, and I proceeded with another one. 😅

Now, looking in more detail, the entire test seems to be very broken since `THTML` got removed from ROOT. My change just hid a symptom by making the interpreter return another error, but it didn't fix any of the underlying issues. (For how to do that, see the list above. )

### Why did it break now?
When `h1` is not registered implicitly, the test fails because the output changes:
```diff
--- /tmp/on.txt	2026-02-20 11:09:57.507412706 +0100
+++ /tmp/off.txt	2026-02-20 11:10:06.524176161 +0100
@@ -1,29 +1,25 @@
 Starting
 Filtering out RooFit banner
 Filtering out RooFit banner
-Info in <TROOT>: Implicit object ownership switched on using ROOT_IMPLICIT_OWNERSHIP
+Info in <TROOT>: Implicit object ownership switched off using ROOT_IMPLICIT_OWNERSHIP
 
-Info in <TROOT>: Implicit object ownership switched on using ROOT_IMPLICIT_OWNERSHIP
+Info in <TROOT>: Implicit object ownership switched off using ROOT_IMPLICIT_OWNERSHIP
 
 input_line_13:2:2: error: unknown type name 'THtml'
  THtml h;
  ^
-input_line_15:2:3: error: member reference type 'TH1F *' is a pointer; did you mean to use '->'?
- h.LoadAllLibs();
- ~^
-  ->
-input_line_15:2:4: error: no member named 'LoadAllLibs' in 'TH1F'
- h.LoadAllLibs();
- ~ ^
+input_line_16:2:3: error: use of undeclared identifier 'h'
+ (h.LoadAllLibs())
+  ^
+Error in <HandleInterpreterException>: Error evaluating expression (h.LoadAllLibs())
+Execution of your code was aborted.
 Program with finished
 input_line_14:2:2: error: unknown type name 'THtml'
  THtml h;
  ^
-input_line_16:2:3: error: member reference type 'TH1F *' is a pointer; did you mean to use '->'?
- h.LoadAllLibs();
- ~^
-  ->
-input_line_16:2:4: error: no member named 'LoadAllLibs' in 'TH1F'
- h.LoadAllLibs();
- ~ ^
+input_line_17:2:3: error: use of undeclared identifier 'h'
+ (h.LoadAllLibs())
+  ^
+Error in <HandleInterpreterException>: Error evaluating expression (h.LoadAllLibs())
+Execution of your code was aborted.
 Program with finished
```

The reason it fails is that `Error in` is detected in the output. Before, it's `input_line_xxx error`, which is accepted without failure. This happens because `h1` is suggested as an alternative when it's registered globally, and that's because `h1`'s TNamed-name is `"h"`. So the error message changes based on whether or not `h1` is registered globally.

### What to do now?
I think the most reasonable way to proceed is to remove this commit from this PR, and fix the entire test as outlined above.

@pcanal thanks for having such a keen eye!

dpiparo: I approve the proposed strategy to move on, provided that it does not require too much work on your side. Otherwise, we can re-think the test, no problem about that.

hageboeck: @pcanal @dpiparo an attempt at fixing the test is now in #21341.
Afterwards, I'll remove the commit from this PR.

pcanal: Thanks Stefan for the thorough investigation!

hageboeck: The commit has now been removed.

hageboeck:
Only the known failures, merging ...

PR #21226, opened 2026-02-10
Prepare tutorials for the ROOT 7 object ownership model

hageboeck:
The code transformations are:
- Write an object explicitly into a file
- Or use `SetDirectory` to add it to a directory


martamaja10 (approved):
Thanks!

PR #21228, opened 2026-02-10
Prepare the "stress" tests for ROOT 7 ownership model.

hageboeck:
Assign histograms explicitly to a directory. This ensures that the tests work independent of whether an implicit registration is done.

### DIFF
--- a/test/stressGraphics.cxx
+++ b/test/stressGraphics.cxx
@@ -3813,6 +3813,8 @@ void zoomfit()
 void hbars()
 {
    TCanvas *C = StartTest(700,800);
+   TDirectory dir("hbars_dir", "Directory for the hbars test");
+   TDirectory::TContext dirCtx{&dir};
 
    TTree *T = (TTree*)gCernstaff->Get("T");
    T->SetFillColor(45);
@@ -3826,9 +3828,10 @@ void hbars()
    //vertical bar chart
    C->cd(2); gPad->SetGrid(); gPad->SetFrameFillColor(33);
    T->Draw("Division>>hDiv","","goff");
-   TH1F *hDiv   = (TH1F*)gDirectory->Get("hDiv");
+   TH1F *hDiv = (TH1F *)dir.Get("hDiv");
    hDiv->SetStats(0);
    TH1F *hDivFR = (TH1F*)hDiv->Clone("hDivFR");
+   hDivFR->SetDirectory(&dir);
    T->Draw("Division>>hDivFR","Nation==\"FR\"","goff");
    hDiv->SetBarWidth(0.45);
    hDiv->SetBarOffset(0.1);


hageboeck:
@pcanal could you have a look when you have a moment?

test/stressGraphics.cxx
```diff
@@ -3772,6 +3772,7 @@ void hbars()
    TH1F *hDiv   = (TH1F*)gDirectory->Get("hDiv");
    hDiv->SetStats(0);
    TH1F *hDivFR = (TH1F*)hDiv->Clone("hDivFR");
+   hDivFR->SetDirectory(gDirectory);
```
pcanal: Is it know what `gDirectory` is point to here? i.e. Can we be more explicit?

hageboeck: It's not directly known. It just happens to be the last-active directory.

What we could do is to explicitly go to gROOT or a hand-made directory for the TTree::Draw calls, and use that same directory for the histogram. How does that sound? @pcanal

hageboeck: I replaced the last commit by something along that line now.

test/stressGraphics.cxx
```diff
@@ -3812,6 +3812,8 @@ void zoomfit()
 void hbars()
 {
    TCanvas *C = StartTest(700,800);
+   TDirectory dir("hbars_dir", "Directory for the hbars test");
+   TDirectory::TContext dirCtx{&dir};
```
pcanal, line 3817: This call to `TContext` is a NOP. After the creation of the `TDirectory`, that directory become the current directory and thus this call to `TContext` has no practical effect.  Maybe we meant:
```suggestion
   TDirectory::TContext dirCtx{};  // Record current directory.
   TDirectory dir("hbars_dir", "Directory for the hbars test");
```

test/stressGraphics.cxx
```diff
@@ -3812,6 +3812,8 @@ void zoomfit()
 void hbars()
 {
    TCanvas *C = StartTest(700,800);
+   TDirectory dir("hbars_dir", "Directory for the hbars test");
```
pcanal, line 3816: Side note: (For better or worse), this changes the lifetime of the histograms.

test/stressGraphics.cxx
hageboeck: That's not what I saw, I believe the `cd()` only happens for TFiles, but not when you just create a TDirectory.

Here is an instrumented version:
```diff
  ////////////////////////////////////////////////////////////////////////////////
  /// Ntuple drawing with alphanumeric variables
+ #include <dbg.h>
  
  void hbars()
  {
     TCanvas *C = StartTest(700,800);
+    dbg(gDirectory->GetName());
     TDirectory dir("hbars_dir", "Directory for the hbars test");
+    dbg(gDirectory->GetName());
     TDirectory::TContext dirCtx{&dir};
+    dbg(gDirectory->GetName());

     TTree *T = (TTree*)gCernstaff->Get("T");
```

This outputs:
```
**********************************************************************
[..t/stressGraphics.cxx:3817 (hbars)] (::ROOT::Internal::TDirectoryAtomicAdapter{})->GetName() = "ntuple_parallelcoord_entries.root" (const char*)
[..t/stressGraphics.cxx:3819 (hbars)] (::ROOT::Internal::TDirectoryAtomicAdapter{})->GetName() = "ntuple_parallelcoord_entries.root" (const char*)
[..t/stressGraphics.cxx:3821 (hbars)] (::ROOT::Internal::TDirectoryAtomicAdapter{})->GetName() = "hbars_dir" (const char*)
```

So creating a `TDirectory` doesn't change `gDirectory`. It changes when you `dir->cd()`, when you open a file, or when you use the context. In this case, it makes the test more local, because a histogram is used for TTree::Draw, and subsequently it's drawn, but there is no need for this histogram to leak out of the test.

test/stressGraphics.cxx
hageboeck: For better in this case. 🙂 
The histogram is used for TTree::Draw, and drawn on a canvas. Once that's done, the histogram can be retired.

PR #21239, opened 2026-02-11
[ci] Fix ROOT build after Visual Studio update

bellenot:


.github/workflows/root-ci-config/build_root.py
```diff
@@ -231,6 +231,7 @@ def parse_args():
     parser.add_argument("--repository",      default="https://github.com/root-project/root.git",
                         help="url to repository")
     parser.add_argument("--overrides",       default=None,      help="Override build options using a syntax like 'A=1 B=2'", nargs="*")
+    parser.add_argument("--vs_update",       default="false",   help="Build after update of Visual Studio")
```
dpiparo: given that the switch is only used to decide whether to upload or not tarballs, why not calling it such and make it True by default?

bellenot: Right, thanks. now done.

Issue #7167, opened 2021-02-10
[TTreeProcessorMP] Warn if Process is called from a multi-thread program


eguiraud:
`fork()`-ing a multi-thread program typically ends up badly. In ROOT, TBB spawns threads when RTaskArena is constructed, and Cocoa spawns threads, on Mac, when graphics are activated.

We would like to employ a similar mechanism to what is used to loop over threads to print their stacktraces on a crash in order to detect whether `TTreeProcessorMP::Process` is called from a multi-thread program, and warn users if so.

PR #21240, opened 2026-02-11
[tree] warn when forking multiprocess

ferdymercury:
Fixes https://github.com/root-project/root/issues/7167

### DIFF
--- a/tree/treeplayer/src/TTreeProcessorMP.cxx
+++ b/tree/treeplayer/src/TTreeProcessorMP.cxx
@@ -14,6 +14,9 @@
 #include "ROOT/TTreeProcessorMP.hxx"
 #include "TMPWorkerTree.h"
 
+#include <thread>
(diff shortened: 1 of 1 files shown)


tree/treeplayer/src/TTreeProcessorMT.cxx
dpiparo, line 1: Why adding a warning here? The issue is about warning if in a multi process context. I think we can tolerate MT within MT, in any case TBB will correctly (hopefully) do the right thing underneath. 
Please correct me if I misunderstand something :)

ferdymercury: Thanks for the review!

Do you suggest to remove the warning from tree/treeplayer/src/TTreeProcessorMT.cxx but keep it on tree/treeplayer/src/TTreeProcessorMP.cxx ?

PR #21243, opened 2026-02-11
[python] Provide better examples of graphics in a Python script

vepadulano:
Show two relevant use cases:

- User wants to plot static objects to a canvas and the canvas should survive and be displayed to screen.
- User wants to plot live objects updating the canvas in a loop and show the canvas at the end of the loop.

### DIFF
--- a/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
+++ b/bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
@@ -8,69 +8,106 @@
 # For the list of contributors see $ROOTSYS/README/CREDITS.                    #
 ################################################################################
 
-r'''
+r"""
 \pythondoc TCanvas
 
-Functionality of TCanvas::Update() method was extended to support interactive
-graphics in the python scripts. If extra block parameter is True, script execution
-will be suspended until <space> key pressed by user. Simple example:
+The TCanvas class is used to create the canvas on which graphical objects such as histograms can be drawn. The following
+is a simple example of typical usage:
 
 \code{.py}
 import ROOT
+import numpy
 
-c = ROOT.TCanvas()
-h = ROOT.TH1I("h1", "h1", 100, -5, 5)
-h.FillRandom("gaus", 10000)
-h.Draw("")
 
-# block here until space is pressed
-c.Update(True)
+def plot():
+    h = ROOT.TH1D("h", "h", 100, -5, 5)
+    h.Fill(numpy.random.normal(size=1000))
 
-# continues after <space> is pressed
-c.SaveAs("canvas.root")
+    c = ROOT.TCanvas()
+    h.Draw()
+    c.Draw(block=True)
+
+
+if __name__ == "__main__":
+    print("Before plot function")
+    plot()
+    print("After plot function")
 \endcode
+
+Note the optional argument `block` passed to the `Draw` method of the canvas. If set to `True`, it will block the script
+execution and run the ROOT graphics event loop until the <space> key is pressed. This allows interacting with the
+canvas and its content until necessary, then move on with the rest of the script.
+
+Another relevant use case is drawing live updates on a canvas, shown in the example below. In this case, the canvas is
+setup by first creating the object to be drawn and drawing it once. Then, the object is updated in a for loop, which
+could represent for example an incoming stream of data with which the histogram should be filled. Each time the plot
+should be updated, the `ModifiedUpdate` function should be called. This will immediately show the new contents on the
+plot. Finally, the canvas is drawn again with `Draw(block=True)` at the end of the loop so that it stays visible and can
+be interacted with.
+
+\code{.py}
+import ROOT
+import numpy
+
+
+def live_update():
+    c = ROOT.TCanvas()
+    h = ROOT.TH1D("h", "h", 100, -5, 5)
+
+    h.Draw()
+    for _ in range(100):
+        h.Fill(numpy.random.normal(size=10))
+        c.ModifiedUpdate()
+    c.Draw(block=True)
+
+if __name__ == "__main__":
+    print("Before plot function")
+    live_update()
+    print("After plot function")
+\endcode
+
 \endpythondoc
-'''
+"""
 
 from . import _run_root_event_loop, pythonization
 
 
-def _TCanvas_Update(self, block = False):
-   """
-   Updates the canvas.
-   Also blocks script execution and runs the ROOT graphics event loop until the <space> keyword is pressed,
-   but only if the following conditions are met:
-   * The `block` optional argument is set to `True`.
-   * ROOT graphics are enabled, i.e. `ROOT.gROOT.IsBatch() == False`.
-   * The script is running not in ipython notebooks.
-   """
+def _TCanvas_Update(self, block=False):
+    """
+    Updates the canvas.
+    Also blocks script execution and runs the ROOT graphics event loop until the <space> key is pressed,
+    but only if the following conditions are met:
+    * The `block` optional argument is set to `True`.
+    * ROOT graphics are enabled, i.e. `ROOT.gROOT.IsBatch() == False`.
+    * The script is not running in ipython notebooks.
+    """
 
-   self._Update()
+    self._Update()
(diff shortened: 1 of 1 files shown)


bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
```diff
@@ -8,69 +8,104 @@
...
+
+
+if __name__ == "__main__":
+    print("Before plot function")
+    plot()
+    print("After plot function")
 \endcode
+
+Note the optional argument `block` passed to the `Draw` method of the canvas. If set to `True`, it will block the script
+execution and run the ROOT graphics event loop until the <space> keyword is pressed. This allows interacting with the
```
hageboeck:
```suggestion
execution and run the ROOT graphics event loop until the <space> key is pressed. This allows interacting with the
```

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
```diff
@@ -8,69 +8,104 @@
...
+canvas and its content until necessary, then move on with the rest of the script.
+
+Another relevant use case is drawing graphics to a canvas as they are being update live. In this case, the `Update`
```
hageboeck:
```suggestion
Another relevant use case is drawing live updates on a canvas. In this case, the `Update`
```

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
```diff
@@ -8,69 +8,104 @@
...
+
+def plot_live():
+    h = ROOT.TH1D("h", "h", 100, -5, 5)
+
+    c = ROOT.TCanvas()
+    for _ in range(100):
+        h.Fill(numpy.random.normal(size=10))
+        h.Draw()
+        c.Update()
+    c.Update(block=True)
```
hageboeck: According to the `TH1::Draw()` documentation, 99 Draw calls can be saved here. I'm not sure if the example becomes easier to understand, but this would be the code:

```suggestion
    c = ROOT.TCanvas()
    h.Draw()
    for _ in range(100):
        h.Fill(numpy.random.normal(size=10))
        c.Update()
    c.Update(block=True)
```

hageboeck: Interestingly, (on the Mac where I tested), none of the two codes results in a live update.
I tested with `time.sleep(1)`, but also this doesn't show any canvas except for when `c.Update(block=True)` is invoked.

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
```diff
@@ -8,69 +8,104 @@
...
-   Also blocks script execution and runs the ROOT graphics event loop until the <space> keyword is pressed,
-   but only if the following conditions are met:
-   * The `block` optional argument is set to `True`.
-   * ROOT graphics are enabled, i.e. `ROOT.gROOT.IsBatch() == False`.
-   * The script is running not in ipython notebooks.
-   """
+def _TCanvas_Update(self, block=False):
+    """
+    Updates the canvas.
+    Also blocks script execution and runs the ROOT graphics event loop until the <space> keyword is pressed,
```
hageboeck:
```suggestion
    Also blocks script execution and runs the ROOT graphics event loop until the <space> key is pressed,
```

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
```diff
@@ -8,69 +8,104 @@
...
+    but only if the following conditions are met:
+    * The `block` optional argument is set to `True`.
+    * ROOT graphics are enabled, i.e. `ROOT.gROOT.IsBatch() == False`.
+    * The script is running not in ipython notebooks.
```
hageboeck:
```suggestion
    * The script is not running in ipython notebooks.
```

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
```diff
@@ -8,69 +8,104 @@
...
-   Draw the canvas.
-   Also blocks script execution and runs the ROOT graphics event loop until the <space> keyword is pressed,
-   but only if the following conditions are met:
-   * The `block` optional argument is set to `True`.
-   * ROOT graphics are enabled, i.e. `ROOT.gROOT.IsBatch() == False`.
-   * The script is running not in ipython notebooks.
-   """
+    """
+    Draw the canvas.
+    Also blocks script execution and runs the ROOT graphics event loop until the <space> keyword is pressed,
```
hageboeck:
```suggestion
    Also blocks script execution and runs the ROOT graphics event loop until the <space> key is pressed,
```

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
```diff
@@ -8,69 +8,104 @@
...
+    but only if the following conditions are met:
+    * The `block` optional argument is set to `True`.
+    * ROOT graphics are enabled, i.e. `ROOT.gROOT.IsBatch() == False`.
+    * The script is running not in ipython notebooks.
```
hageboeck:
```suggestion
    * The script is not running in ipython notebooks.
```

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
hageboeck: What works is this, but the code isn't simpler:
```py
    c = ROOT.TCanvas()                                                                                                     
    h.Draw()                                                                                                               
    for i in range(100):                                                                                                   
        h.Fill(numpy.random.normal(size=10))                                                                                                                                                                                                                                                                                                                                                                                          
        c.Modified()                                                                                                       
        c.Update()                                                                                                                                                                                                     
    c.Update(block=True)  

```

bindings/pyroot/pythonizations/python/ROOT/_pythonization/_tcanvas.py
```diff
@@ -8,69 +8,104 @@
...
+    print("After plot function")
 \endcode
+
+Note the optional argument `block` passed to the `Draw` method of the canvas. If set to `True`, it will block the script
+execution and run the ROOT graphics event loop until the <space> keyword is pressed. This allows interacting with the
+canvas and its content until necessary, then move on with the rest of the script.
+
+Another relevant use case is drawing graphics to a canvas as they are being update live. In this case, the `Update`
+method should be used to refresh the canvas after the contents have been updated. If it is desired to show the canvas on
+screen, the `block` argument can be used to the same effect as described above:
```
hageboeck: This works for me on Linux, but the canvas doesn't show on Mac until the blocking `Update()` is invoked.

vepadulano:
@linev , @hageboeck is right, this simple script doesn't work on MacOS (but it works on Linux):

```python
import ROOT
import numpy


def plot_live():
    h = ROOT.TH1D("h", "h", 100, -5, 5)

    c = ROOT.TCanvas()
    for _ in range(100):
        h.Fill(numpy.random.normal(size=10))
        h.Draw()
        c.Update()
    c.Update(block=True)

if __name__ == "__main__":
    print("Before plot function")
    plot_live()
    print("After plot function")
```

The canvas should show up immediately on screen at the beginning of the for loop and then at the end of the for loop it should stay on screen thanks to the use of `block=True`. But on MacOS the canvas shows up on screen only at the end.

vepadulano:
@hageboeck I have addressed your comments. I removed the second example shown because of https://github.com/root-project/root/issues/21362

vepadulano:
@hageboeck The use case of live update has been re-introduced with updated code that works on both MacOS and Linux. Feel free to double check.

hageboeck:
> @hageboeck The use case of live update has been re-introduced with updated code that works on both MacOS and Linux. Feel free to double check.

It works perfectly, good work!

PR #21246, opened 2026-02-11
[RDF] Always write nominal columns in Snapshot with variations.

hageboeck:
Before, the nominal columns were omitted from the output if their filter
didn't pass. This is, however, insufficient, because if at least two
variations are run, (some of) the nominal columns might still be in use.

One example is this setup:
```c++
df.Vary("x", [...], "var0")
  .Vary("y", [...], "var1")
  .Filter([...], {"x", "y"})
  .Snapshot([...], {"x", "y"})
```

In every variation universe, a Filter on `x` and `y` is used. If one writes down
what the value of the nominal and varied columns should be in the snapshot,
one gets conflicting requirements for the nominal columns if one keeps zeroeing
the invalid columns to save space as was done before.

Imagine a dataset where only the `var0` variation passes the filter. One gets:
|  | x | y | x_var0 | y_var1 |
|---|---|---|---|---|
| Filter | fail | fail | pass | fail |
| Should the columns be valid? | No, because nominal didn't pass | No, because nominal didn't pass, but yes because `var0` passed | Yes | No |
| What's written before this PR | 0 | 0 |x_var0|0|
| What's written now | x | y | x_var0 | 0 (invalid) |

The problem is the column `y`:
In the nominal universe, it shouldn't be present, but since it's used as an unvaried column also in the `var0` universe, it must be in the output dataset. The same argument can be made for `x` if `var1` were to pass. Therefore, the nominal columns must remain visible in the output dataset as soon as any filter in any variation passes.

The tests have been updated to reflect the fact that the nominal columns are not zeroed out.

### DIFF
--- a/tree/dataframe/test/dataframe_snapshotWithVariations.cxx
+++ b/tree/dataframe/test/dataframe_snapshotWithVariations.cxx
@@ -53,7 +53,8 @@ void checkOutput(TTree &tree, std::vector<std::string> const &systematics, F &&a
          ASSERT_GT(tree.GetEntry(i), 0);
 
          EXPECT_EQ(x, -1 * y);
-         if (!activeCuts(x, y)) {
+         if (!activeCuts(x, y) && !sysName.empty()) {
+            // Branches with systematics should be zeroed when cuts don't pass
             EXPECT_EQ(x, X_t{});
             EXPECT_EQ(y, Y_t{});
          }
@@ -98,7 +99,7 @@ TEST(RDFVarySnapshot, SimpleRDFWithFilters)
       for (const auto branchName : {"x", "y", "x__xVar_0", "x__xVar_1", "y__xVar_0", "y__xVar_0"})
          EXPECT_NE(tree->FindBranch(branchName), nullptr) << branchName;
 
-      checkOutput<float, double>(*tree, std::vector<std::string>{"__xVar_0", "__xVar_1"}, cuts);
+      checkOutput<float, double>(*tree, std::vector<std::string>{"", "__xVar_0", "__xVar_1"}, cuts);
 
       if (HasFailure()) {
          tree->Print();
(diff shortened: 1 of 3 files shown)


tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -98,7 +99,7 @@ TEST(RDFVarySnapshot, SimpleRDFWithFilters)
       for (const auto branchName : {"x", "y", "x__xVar_0", "x__xVar_1", "y__xVar_0", "y__xVar_0"})
          EXPECT_NE(tree->FindBranch(branchName), nullptr) << branchName;
 
-      checkOutput<float, double>(*tree, std::vector<std::string>{"__xVar_0", "__xVar_1"}, cuts);
+      checkOutput<float, double>(*tree, std::vector<std::string>{"", "__xVar_0", "__xVar_1"}, cuts);
```
vepadulano, line 103: What is the empty string here?

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -264,50 +265,59 @@ TEST(RDFVarySnapshot, Bitmask)
...
+         ASSERT_EQ(count_sys.GetValue(), nVal) << "systematic: " << systematic;
+         EXPECT_EQ(meanX_sys.GetValue(), sum / nVal) << "systematic: " << systematic;
+         EXPECT_EQ(meanY_sys.GetValue(), -1. * sum / nVal) << "systematic: " << systematic;
       }
+   }
 
-      if (HasFailure()) {
-         tree->Scan("entry:x:y:x__xVar_0:y__xVar_0:x__xVar_1:y__xVar_1:x__xVar_2:y__xVar_2");
+   if (HasFailure()) {
+      TFile file(filename, "READ");
```
vepadulano: Many times in the past we've seen weird things happening when one or both of the file or the tree were not allocated on the heap, so I would suggest
```suggestion
      auto file = std::make_unique<TFile>(filename);
```

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -264,50 +265,59 @@ TEST(RDFVarySnapshot, Bitmask)
...
+      std::unique_ptr<TTree> tree{file.Get<TTree>(treename.data())};
+      ASSERT_NE(tree, nullptr);
+      tree->Scan("entry:R_rdf_mask_testTree_0:x:y:x__xVar_0:y__xVar_0:x__xVar_1:y__xVar_1:x__xVar_2:y__xVar_2");
+
+      auto map = file.Get<std::unordered_map<std::string, std::pair<std::string, unsigned int>>>(
+         ("R_rdf_column_to_bitmask_mapping_" + treename).c_str());
+      for (auto const &[name, mapping] : *map) {
+         std::cout << std::setw(20) << name << " --> " << mapping.first << "  " << mapping.second << "\n";
```
vepadulano, line 322: missing `#include <iomanip>`

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -264,50 +265,59 @@ TEST(RDFVarySnapshot, Bitmask)
...
-      if (HasFailure()) {
-         tree->Scan("entry:x:y:x__xVar_0:y__xVar_0:x__xVar_1:y__xVar_1:x__xVar_2:y__xVar_2");
+   if (HasFailure()) {
+      TFile file(filename, "READ");
+      std::unique_ptr<TTree> tree{file.Get<TTree>(treename.data())};
+      ASSERT_NE(tree, nullptr);
+      tree->Scan("entry:R_rdf_mask_testTree_0:x:y:x__xVar_0:y__xVar_0:x__xVar_1:y__xVar_1:x__xVar_2:y__xVar_2");
+
+      auto map = file.Get<std::unordered_map<std::string, std::pair<std::string, unsigned int>>>(
+         ("R_rdf_column_to_bitmask_mapping_" + treename).c_str());
```
vepadulano: For good measure I would wrap this in a `unique_ptr`.

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -577,3 +589,75 @@ TEST(RDFVarySnapshot, IncludeDependentColumns_JIT)
...
    }
 }
+
+TEST(RDFVarySnapshot, TwoVaryExpressions)
+{
+   constexpr auto filename = "VarySnapshot_TwoVaryExpressions.root";
+   RemoveFileRAII(filename);
+   constexpr unsigned int N = 10;
+   ROOT::RDF::RSnapshotOptions options;
+   options.fOverwriteIfExists = true;
```
vepadulano, line 601: Is `options.fOverwriteIfExists` needed for the test here?

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -577,3 +589,75 @@ TEST(RDFVarySnapshot, IncludeDependentColumns_JIT)
...
+                  .Snapshot("t", filename, {"x", "y"}, options);
+
+   {
+      std::unique_ptr<TFile> file{TFile::Open(filename)};
+      auto tree = file->Get<TTree>("t");
+
+      EXPECT_EQ(tree->GetEntries(), 10);
+      EXPECT_EQ(tree->GetNbranches(), 7); // 6 branches for x/y with variations, bitmask
+      for (const auto branchName : {"x", "y", "x__xVar_down", "x__xVar_up", "y__yVar_down", "y__yVar_up"})
+         EXPECT_NE(tree->FindBranch(branchName), nullptr) << branchName;
```
vepadulano: In this particular test `tree->GetBranch` should be enough

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
```diff
@@ -577,3 +589,75 @@ TEST(RDFVarySnapshot, IncludeDependentColumns_JIT)
...
+               EXPECT_EQ(x, expectedX) << "entry:" << i << " (" << xName << "=" << expectedX << ", " << yName << "="
+                                       << expectedY << ")";
+               EXPECT_EQ(y, expectedY) << "entry:" << i << " (" << xName << "=" << expectedX << ", " << yName << "="
+                                       << expectedY << ")";
+               EXPECT_TRUE(cuts(x, y)) << "entry:" << i << " (" << xName << "=" << expectedX << ", " << yName << "="
+                                       << expectedY << ")";
+            }
+         }
+
+         tree->ResetBranchAddresses();
```
vepadulano, line 652: Why do we need to call this here?

vepadulano (approved):
Thanks for the fix! This will enable more use cases for the Snapshot with variations 👍 Changes LGTM, I left a couple of comments in the test only

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
hageboeck: It's the names of the systematics to check. `"" == "nominal"`, but it's not called nominal in the outputs.

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
hageboeck: With the `RemoveFileRAII`, it's probably needed less, but if the program crashes (or is killed with Ctrl+C), you might have a forever-failing test until you remove the broken file.

tree/dataframe/test/dataframe_snapshotWithVariations.cxx
hageboeck: Because the branches are bound to stack variables which are now going out of scope. If you don't release them here, things can crash in weird ways.

PR #21248, opened 2026-02-11
[core] Deprecate TDirectory::AddDirectoryStatus()

hageboeck:
Contrary to what the function documentation states, `TDirectory::AddDirectoryStatus` has absolutely no effect on ROOT.
The only situation where this might have an effect is users setting the status bit and checking it in another location. Therefore, the functions are deprecated now and marked for removal in ROOT 7.

To document this change, the docstrings in `TDirectory.h` were fixed, so doxygen assigns them to the correct member.

*Note*: clang-format is expected to fail, since the classic ROOT headers are not formatted correctly.

### DIFF
--- a/README/ReleaseNotes/v640/index.md
+++ b/README/ReleaseNotes/v640/index.md
@@ -37,6 +37,9 @@ The following people have contributed to this new version:
 * The headers in `RooStats/HistFactory` for data classes related to the measurement definition were merged into the `RooStats/HistFactory/Measurement.h` header to simplify usage and development. For now, the whole set of header files is kept for backwards compatibility, but the empty headers will be removed in ROOT 7.
 * The `TROOT::GetSourceDir()` method is deprecated and will be removed in ROOT 6.42. It stopped making sense because the ROOT source is generally not shipped alongside ROOT in the `src/` subdirectory anymore.
 * Using the `rpath` build option - deprecated and without effect since ROOT 6.38 - is now scheduled to give configuration errors starting from ROOT 6.42.
+* `TDirectory::AddDirectoryStatus()` and `TDirectory::AddDirectory()` have been deprecated. These functions were meant to replace TH1::AddDirectoryStatus(), but
+  never had any effect on ROOT. The associated bit TDirectory::fgAddDirectory was deprecated as well. Although users can set and read the bit, its usage should be
+  stopped completely to avoid any confusion. The bit and functions will be removed in ROOT 7.
 
 ## Removals
 


core/base/inc/TDirectory.h
```diff
@@ -88,11 +88,12 @@ can be replaced with the simpler and exception safe:
...
-      std::atomic<bool> fActiveDestructor{false};   //! Set to true during the destructor execution
-      std::atomic<bool> fDirectoryWait{false};      //! Set to true if a TDirectory might still access this object.
-      TContext   *fPrevious{nullptr};               //! Pointer to the next TContext in the implied list of context pointing to fPrevious.
-      TContext   *fNext{nullptr};                   //! Pointer to the next TContext in the implied list of context pointing to fPrevious.
+      std::atomic<TDirectory *> fDirectory{nullptr}; ///< Pointer to the previous current directory.
+      std::atomic<bool> fActiveDestructor{false};    ///< Set to true during the destructor execution
+      std::atomic<bool> fDirectoryWait{false};       ///< Set to true if a TDirectory might still access this object.
+      TContext *fPrevious{
+         nullptr};              ///< Pointer to the next TContext in the implied list of context pointing to fPrevious.
+      TContext *fNext{nullptr}; ///< Pointer to the next TContext in the implied list of context pointing to fPrevious.
```
ferdymercury: I think this is not a Doxygen comment but rather a meta-mark for I/O to be transient member?

If you want Doxygen commenting, then we would need sth like:

```
/// Transient pointer to the previous current directory. \deprecated
std::atomic<TDirectory*> fDirectory{nullptr}; //! Pointer to the previous current directory.
```

hageboeck: > I think this is not a Doxygen comment but rather a meta-mark for I/O to be transient member?

Ouch, I forgot about this ... 🥲

hageboeck: And we need to tell doxygen to ignore `//!` ...

ferdymercury: Maybe even better: we could try to modify

`documentation/doxygen/filter.cxx`

so that each line containing "//!" is transformed on the fly into "///<"

hageboeck: OK, I had to refresh my memory ...
The comment to use is:
```
T a; ///<!
```
This is understood both by doxygen and ROOT's IO.

ferdymercury: ahh nice. thks.

Maybe it's worth to use "sed" to replace all 1916 occurrences of the same issue.

core/base/src/TDirectory.cxx
```diff
@@ -164,31 +164,24 @@ TDirectory::TContext::~TContext()
...
-/// ~~~ {.cpp}
-///     TDirectory::AddDirectory
-/// ~~~
+/// Set the value returned by TDirectory::AddDirectoryStatus().
+/// \deprecated This function is not used in ROOT.
 
 void TDirectory::AddDirectory(Bool_t add)
 {
+   // When removing this function, mark the bit as deprecated.
+   // Derived classes out there might be using it.
```
guitargeek: Why is that the plan? Can we not deprecate the `fgAddDirectory` at the same time as the getter and setter function explicitly with `R__DEPRECATED`?

I know that would require some gymnastics, which we have done before too. To avoid deprecation warnings when ROOT builds itself, you need to guard the deprecation macro of `fgAddDirectory` with a private preprocessor define, similar to:
https://github.com/root-project/root/commit/59da6b1808dc6c7801ed0108d70b0c441e460c3a

hageboeck:
> Thanks for this initiative and cleanup! But can we explicitly deprecate `fgAddDirectory` at the same time if it's unused?

Not yet. We would ourselves receive deprecation warnings when accessing it in the deprecated functions.

core/base/src/TDirectory.cxx
```diff
@@ -163,35 +161,43 @@ TDirectory::TContext::~TContext()
...
 }
 
+// Mask deprecation warnings to allow for deprecating the fgAddDirectory bit.
+#ifdef _MSC_VER
+#pragma warning( push )
+#pragma warning( disable : C4996 )
+#else
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
```
ferdymercury, line 171:
```suggestion
#elif defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
#elif defined(__GNUC__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
#endif
```

core/base/src/TDirectory.cxx
```diff
@@ -163,35 +161,43 @@ TDirectory::TContext::~TContext()
...
+   // Derived classes out there might be using it.
    return fgAddDirectory;
 }
 
+#ifdef _MSC_VER
+#pragma warning( pop )
+#else
+#pragma GCC diagnostic pop
+#endif
+
```
ferdymercury, line 196:
```suggestion
#ifdef _MSC_VER
#pragma warning( pop )
#elif defined(__clang__)
#pragma clang diagnostic pop
#elif defined(__GNUC__)
#pragma GCC diagnostic pop
#endif

```

core/base/src/TDirectory.cxx
hageboeck: @guitargeek Gymnastics have been done, I believe. Have a look if that's what you had in mind.

guitargeek:
> > Thanks for this initiative and cleanup! But can we explicitly deprecate `fgAddDirectory` at the same time if it's unused?
> 
> Not yet. We would ourselves receive deprecation warnings when accessing it in the deprecated functions.

You can avoid this with some build-time macros (see my earlier comment). Sorry I only edited that later into the comment because I realized the solution is not obvious

hageboeck:
> You can avoid this with some build-time macros (see my earlier comment). Sorry I only edited that later into the comment because I realized the solution is not obvious

I did it in the last commit (albeit only for MSVC, gcc and clang, but I think that should be sufficient). My reply [here](https://github.com/root-project/root/pull/21248#discussion_r2797832482) was probably easy to overlook, though.

README/ReleaseNotes/v640/index.md
```diff
@@ -37,6 +37,8 @@ The following people have contributed to this new version:
 * The headers in `RooStats/HistFactory` for data classes related to the measurement definition were merged into the `RooStats/HistFactory/Measurement.h` header to simplify usage and development. For now, the whole set of header files is kept for backwards compatibility, but the empty headers will be removed in ROOT 7.
 * The `TROOT::GetSourceDir()` method is deprecated and will be removed in ROOT 6.42. It stopped making sense because the ROOT source is generally not shipped alongside ROOT in the `src/` subdirectory anymore.
 * Using the `rpath` build option - deprecated and without effect since ROOT 6.38 - is now scheduled to give configuration errors starting from ROOT 6.42.
+* `TDirectory::AddDirectoryStatus()` and `TDirectory::AddDirectory()` are deprecated. These functions have long been replaced by TH1::AddDirectoryStatus(), and
```
guitargeek: It's a protected member that probably nobody uses, but maybe it's still worth to mention the deprecation of `TDirectory::fgAddDirectory` in the release notes?

guitargeek:
> > You can avoid this with some build-time macros (see my earlier comment). Sorry I only edited that later into the comment because I realized the solution is not obvious
> 
> I did it in the last commit (albeit only for MSVC, gcc and clang, but I think that should be sufficient). My reply [here](https://github.com/root-project/root/pull/21248#discussion_r2797832482) was probably easy to overlook, though.

Indeed I overlooked your comment :slightly_smiling_face:

README/ReleaseNotes/v640/index.md
hageboeck: OK, good idea! I'll add that after the :coffee:. 🙂

guitargeek:
Thanks!

PR #21250, opened 2026-02-11
[cling] Ensure HeaderFileInfo exists during incremental parsing 

devajithvs:
When registering incremental inputs as virtual files, we need to ensure that the `HeaderSearch` is aware of the `FileEntry`.

This prevents assertions in `HeaderSearch::LookupFile` where it expects file info to exist.

This will help revert one of our LLVM patches that was added when upgrading to LLVM20.

# This Pull request:

## Checklist:

- [ ] tested changes locally

This PR fixes #


devajithvs:
Upstreaming fixes: https://github.com/llvm/llvm-project/pull/182044, https://github.com/llvm/llvm-project/pull/182070

devajithvs:
> > Upstreaming fixes: [llvm/llvm-project#182044](https://github.com/llvm/llvm-project/pull/182044), [llvm/llvm-project#182070](https://github.com/llvm/llvm-project/pull/182070)
> 
> To confirm: these are not needed for this PR, but align upstream `clang-repl` to what Cling does / needs to do?

Yes, these are not needed for this PR, just aligning to what cling does now. That will help get rid of one more downstream cling patch.

PR #21257, opened 2026-02-12
[docu] Fix 2000 off-by-one mistakes

ferdymercury:
# This Pull request:

## Changes or fixes:

As found out by @hageboeck in https://github.com/root-project/root/commit/5f614f8edd6eb4cc7e659b2d82567e1f549bd3e4

there were almost 2000 errors in the online documentation where the members were assigned the documentation of their neighbors. See for example:

image

where the docu is off-by-one.

This fixes it throughout the whole codebase, without disturbing the transient I/O mark, copying what was done by Stephan for TDirectory.h in the linked commit.

## Checklist:

- [ ] tested changes locally

### DIFF
--- a/bindings/tpython/inc/TPyArg.h
+++ b/bindings/tpython/inc/TPyArg.h
@@ -49,7 +49,7 @@ public:
    ClassDef(TPyArg, 1) // Python morphing argument type
 
 private:
-   mutable PyObject *fPyObject; //! converted C++ value as python object
+   mutable PyObject *fPyObject; ///<! converted C++ value as python object
 };
 
 #endif
--- a/bindings/tpython/inc/TPyReturn.h
+++ b/bindings/tpython/inc/TPyReturn.h
@@ -59,7 +59,7 @@ public:
    ClassDef(TPyReturn, 1) // Python morphing return object
 
 private:
-   PyObject *fPyObject; //! actual python object
+   PyObject *fPyObject; ///<! actual python object
 };
 
 #endif


hist/histv7/inc/ROOT/RRegularAxis.hxx
```diff
@@ -46,7 +46,7 @@ private:
    /// The upper end of the axis interval
    double fHigh;
    /// The cached inverse of the bin width to speed up ComputeLinearizedIndex
-   double fInvBinWidth; //!
+   double fInvBinWidth;
```
hahnjo: This comment is intentional, it marks the data member as transient (for future IO support). As you can see, this is no problem in the generated Doxygen: https://root.cern/doc/master/classROOT_1_1Experimental_1_1RRegularAxis.html#a71ef532bba156a2d5851fe29bcf0ce8b please remove this change from the PR, and I would strongly suggest to check other instances.

pcanal: This instance is different from most others as the comment after the data member is meant to be only the I/O decoration and is not the doxygen comment (which is the line before).    Most others are meant to be both the I/O decoration and the doxygen comment.

So indeed, in this case the right choice is to keep it as is.

hageboeck:
I agree with @hahnjo that this introduces some bugs. Besides that, I would not touch (and review) 300+ files. I think it makes sense to fix these errors when you have to touch a member anyway, but a global change like this is akin to when the universe became transparent: When you look back in git, you only see the CMB.

hist/histv7/inc/ROOT/RRegularAxis.hxx
ferdymercury: Would you be ok to have 

`double fInvBinWidth; ///<! The cached inverse of the bin width to speed up ComputeLinearizedIndex`

?

> As you can see, this is no problem in the generated Doxygen

This is just by chance, but imagine in 10 years, someone adds something to this code.
`double fInvBinWidth; //! transient`

Then the bug will be silently back.

By having it as `///<! The cached inverse of the bin width to speed up ComputeLinearizedIndex`
you prevent this from happening in the future. Since the member is transient it does not matter if this is comment is also used as I/O decoration on top of being a Doxygen comment?

ferdymercury:
> I think it makes sense to fix these errors when you have to touch a member anyway, but a global change like this is akin to when the universe became transparent: When you look back in git, you only see the CMB.

This can be partly solved using https://dev.to/ramunarasinga-11/git-blame-ignore-revs-to-ignore-bulk-formatting-changes-5fee

I don't agree here with Stephan: it's much worse to have 2000 (well maybe just 500, if 1500 were just empty ///! with no extra doc) wrongly documented members in our online documentation.

Wrt 300+ files, I can split them in different PR per commit, but then it's worse for the CI

Not doing anything because it clutters github is a developer-centered perspective, that I understand. But also think of the user-centered perspective: "I found yet another wrongly documented thing in ROOT docu"

And also think of the domino effect: if I create another class, I will copy paste one as example that is still wrong, increasing the bugginess size

hist/histv7/inc/ROOT/RRegularAxis.hxx
hahnjo: No, Doxygen comments in new code goes before the member.

PR #21258, opened 2026-02-12
[ntuple] Optimize memory layout of cluster descriptor

jblomer:
More space-efficient memory layout of locator and page range. As a result, a typical cluster descriptor consumes about 1/3 less space in memory.


tree/ntuple/src/RNTupleTypes.cxx
```diff
@@ -0,0 +1,102 @@
...
+
+#include <TError.h>
+
+#include <string>
+
+/// See GetReserved(): we ignore the reserved flag since we don't use it anywhere currently.
+void ROOT::RNTupleLocator::SetReserved(std::uint8_t reserved)
+{
+   if (reserved > 1) {
+      R__LOG_WARNING(ROOT::Internal::NTupleLog()) << "Overflow of the locator reserved field";
```
hahnjo: Should this throw instead of warning?

